PR #307, opened 2020-03-22
Gitlab integration

MostafaAmin07:
Well, this is the basic functionality for the Gitlab. It should be doing everything that Github is currently doing.

I have to say that the new structure is a lot easier than the previous one.
You did a great job @johannesjo.

But I think there is a problem not only with this but with Github as well and maybe the Jira integration.
The requests to the API are being sent too many times.

I have no experience with PWA so I don't know exactly what is happening and why the caching is not working correctly.

PR #316, opened 2020-03-28
Additional gitlab integration

MostafaAmin07:
This is not finished yet.
I will implement as many of the additional features that we talked about in #307 as I can in this branch.

I just opened the PR to try and solve a problem I am facing and to discuss the code.


MostafaAmin07:
@johannesjo I am facing a strange problem in here.
In the Gitlab API service, I am trying to send a PUT request but for some reason, it is not authorized.

When I try to debug it I found that the headers get created correctly when a GET request is being sent but not with the PUT request. why? I have no idea.

So if you have some time could you please have a look at it.

src/app/features/tasks/task-summary-table/task-summary-table.component.ts
```diff
@@ -40,9 +42,13 @@ export class TaskSummaryTableComponent {
   }
 
   toggleTaskDone(task: Task) {
-    task.isDone
-      ? this._taskService.setUnDone(task.id)
-      : this._taskService.setDone(task.id);
+    if (task.isDone) {
+      this._taskService.setUnDone(task.id);
+      this._issueService.reopenIssue(task);
```
johannesjo: I think It's better to stay consistent and implement this via an effect. Have a look at the jira effects there is something similar to this.

johannesjo:
I am currently unable to reproduce this, as searching gitlab does not work, because the rate limit is quickly reached (status code 429). Probably makes sense to address this first.
1. I'd recommend to use the actual search api (https://docs.gitlab.com/ee/api/search.html) rather than the previous approach. If this by itself isn't enough, we could use rxjs debounceTime to wait for typing to finish before doing the search.
2. It would be cool to have an better indication (snack message) when a 429 occurs, so users know what it is and why it happened.

MostafaAmin07:
Ok, I will take a look at this.

src/app/t.const.ts
```diff
@@ -119,6 +119,7 @@ export const T = {
         'IS_AUTO_ADD_TO_BACKLOG': 'F.GITLAB.FORM.IS_AUTO_ADD_TO_BACKLOG',
         'IS_AUTO_POLL': 'F.GITLAB.FORM.IS_AUTO_POLL',
         'IS_SEARCH_ISSUES_FROM_GITLAB': 'F.GITLAB.FORM.IS_SEARCH_ISSUES_FROM_GITLAB',
+        'IS_COLSE_REOPEN': 'F.GITLAB.FORM.IS_CLOSE_REOPEN',
```
jdbdline, line 122: IS_COLSE_REOPEN seems like a spelling mistake

MostafaAmin07: It is. Thank you I will fix that.

src/app/features/issue/providers/gitlab/gitlab.const.ts
```diff
@@ -52,6 +53,13 @@ export const GITLAB_CONFIG_FORM: LimitedFormlyFieldConfig<GitlabCfg>[] = [
       label: T.F.GITLAB.FORM.IS_AUTO_POLL
     },
   },
+  {
+    key: 'isCloseReopenEnabled',
+    type: 'checkbox',
+    templateOptions: {
+      label: T.F.GITLAB.FORM.IS_COLSE_REOPEN
```
jdbdline, line 60: here is also same mistake: IS_COLSE_REOPEN

johannesjo:
@MostafaAmin07 totally forgot about this one. Sorry. Is this ready to merge (apart from the merge conflicts)?

MostafaAmin07:
@johannesjo No worries I forgot it too, This is not needed now. 
I already fixed the Gitlab issue link in #440 and I am implementing the opening and closing of the issues through SP along with other changes and will open a PR soon hopefully.

Issue #317, opened 2020-03-29
Electron bug after gitlab integration


yossefaz:
* Version used: snap
* Operating System and version: Linux
* Desktop Environment: Ubuntu 19.10

### Expected Behavior
Issues sync as expected from Gitlab Project (with Personal Access Token)
Suggestion : develop a full integration with webhooks like JIRA and others does

### Current Behavior
The app crashes when open the project (that I added with gitlab integration) on the electron app.
A big red alert open and suggest to reload the app, but it crashes again and again

### Steps to Reproduce (for bugs)

1. On the desktop App : Open a new project
2. On the desktop App :Click on gitlab integration
3. On Gitlab.com : create an access token from with all access
4.  On the desktop App : Add the project URL in the requested format (username%2Fprojectname), add the just created token, select all the option and click "Save"

### Console Output
![Screenshot from 2020-03-29 11-28-01](https://user-images.githubusercontent.com/61141682/77844579-6f143700-71b0-11ea-8a83-82337cda946a.png)

### Error Log (Desktop only)
[log.log](https://github.com/johannesjo/super-productivity/files/4398469/log.log)

johannesjo:
@MostafaAmin07 could you maybe take a look on this?

MostafaAmin07:
I will take a look at this.

MostafaAmin07:
Ok, I know what is causing the crashing now and I will start solving it right away.

johannesjo:
@yossefaz this should be fixed with the next release. Please let us know if it works for you.

yossefaz:
Ok ! Great ! I will wait for the next release and test it once it will be available

yossefaz:
I tested the last release.
Now I can sync Giltab issues : 
![Screenshot from 2020-04-05 10-53-07](https://user-images.githubusercontent.com/61141682/78469690-33d3b400-772c-11ea-9662-6b72db50ea80.png)

And this is great !

But : 

The app freeze after 10 second with this message : 

![Screenshot from 2020-04-05 10-44-19](https://user-images.githubusercontent.com/61141682/78469710-536adc80-772c-11ea-8788-b10dba541919.png)

And when i `top` my system i got 3 instances of superporductivity and one of them with more than 100% CPU !

![Screenshot from 2020-04-05 10-54-02](https://user-images.githubusercontent.com/61141682/78469759-9b89ff00-772c-11ea-81e5-7f18d6b63f9d.png)

[...]

PR #319, opened 2020-03-31
fix: filter error for projects with no issues

MostafaAmin07:
This should solve the error in #317.


johannesjo:
Thank you very much!

PR #330, opened 2020-04-05
Support HD tray icon on MacOS

Jackymancs4:
Hello @johannesjo 

Here's a little tweak that I think is a valid visual improvement.

First, the higher resolution icons, while present, was wrongly ending with `2x` suffix.
According to https://www.electronjs.org/docs/api/native-image#high-resolution-image the correct suffix is `@2x`.

Second, the PR leverage the template image concept of MacOS (for doc: https://www.electronjs.org/docs/api/native-image#template-image) that let the OS choose the right color. The main advantage being support for selection highlighting and no more need to re open the app if the OS theme changes.

Also, this allow `systemPreferences.isDarkMode()` to be removed, since it's been deprecated (https://www.electronjs.org/docs/api/system-preferences#systempreferencesisdarkmode-macos-windows-deprecated) and electron was already complaining about it in the console with:
```
(electron) 'systemPreferences.isDarkMode()' is deprecated and will be removed. Please use 'nativeTheme.shouldUseDarkColors' instead.
```

|Before|After|
|--|--|
|before-dark-selected|after-dark-selected|
|before-dark-unselected|after-dark-unselected|
|before-light-selected|after-light selected|
|before-light-unselected|after-light-unselected|

The other OSs are left unaffected by this.


johannesjo:
Wow that's cool! Thank you very much!

PR #332, opened 2020-04-06
Update italian localization

Jackymancs4:
I updated the italian localization using `src/assets/i18n/en.json` as a template.


johannesjo:
That's super cool! Thank you very much!

PR #379, opened 2020-05-27
Added Dutch translation

kerkenit:


johannesjo:
That's great! Thank you very much!

PR #383, opened 2020-06-01
chore: use observable for Formly validation message

aitboudad:
Formly `v5.7` is released and now allows passing validation message as observable which improves live changes when switching language.


johannesjo:
This is great! Thank you very much!! :)

johannesjo:
Maybe not the right place for this, but I just read on linkedin that you're also working as a freelancer. I am curious: How is the situation for you, if I may ask?

aitboudad:
Seeking for a new challenge (but silently :smile:)

johannesjo:
That sounds good (: Same here in a way.  I wish you good luck with that!

aitboudad:
I get used to it, good luck to you too!

PR #407, opened 2020-06-22
Update nl.json

kerkenit:
Fixed wrong translated text


johannesjo:
Much appreciated! Thank you!

PR #440, opened 2020-07-10
Gitlab self host, project url

MostafaAmin07:
@johannesjo Could you please look into this?

I updated the project to support a self-hosted instance based on the various comments I found in the issues.
I can't be sure 100% that it works with a self-hosted instance because I don't have one but I am sure that this solution doesn't break any existing configuration and it allows for easier configuration as now you can just copy the project URL from the browser and it will work.

I updated the way the user is required to enter the project URL to match any of the next 3:
1- Full URL ex: https://gitlab.or.mygitlab.com/.....
2- Only the project path not including any Gitlab host URL information
3- the old format of username%2Fproject or any other input separated by %2F

So please review the code and let me know what you think and if there is anything I am missing?


src/app/features/issue/providers/gitlab/gitlab-api/gitlab-api.service.ts
```diff
@@ -179,4 +177,25 @@ export class GitlabApiService {
     }
     return throwError({[HANDLED_ERROR_PROP_STR]: 'Gitlab: Api request failed.'});
   }
+
+  private apiLink(projectConfig: GitlabCfg): string {
+    let apiURL: string = '';
+    let projectURL: string = projectConfig.project ? projectConfig.project : '';
+    const hostURL = projectConfig.project?.match(GITLAB_URL_REGEX);
```
johannesjo, line 184: <3 it's so cool that this is possible now!

johannesjo:
Hey Mostafa! Good to have you back! :) 

Thank you very much for the adjustments! I try to have a look on Sunday when I have a long train ride. At first glance, the code looks absolutely fine. Just a thought: The apiLink function looks like a good place to exercise some unit testing. 

I don't have a self-hosted GitLab instance either, but I am sure the guy requesting this feature can help out. For now, I would just merge it as is if it doesn't break anything existing.

Have a nice weekend!

MostafaAmin07:
I hope that it will be a nice and easy ride.

I will add the translations to the internationalization files.
I like the idea of adding unit tests for it so no need to merge it right away, also to give me some time to test its behavior even more.

I don't like the current behavior of calling the apiLink function with every request cause this means that the searching and replacing happens way too many times. 
I first had thought of 2 solutions fo it:
1- To have a local variable in the API service that gets set with the first request made and after that instead of calling the function again just use it (But that won't work right because it's value won't change once its set until the app is closed so if the user changed the project path the link won't be updated).
2- To add some type of timestamp with the project config and the API service to check for it (But that means adding another field that may cause problems to the existing project or need a bit more work).

Actually thinking about it now. I think I have a better solution for it now maybe it's better to add apiLink as a field that is automatically computed and saved in Gitlabcfg that way all that work of searching and constructing a link is done once per configuration change But that would require a change in the previously setup projects via a migration.

As you can see there is a lot that I can do so if you have any suggestions on a solution that is easier or impact have less impact please tell me.

johannesjo:
@MostafaAmin07 shall I hold on merging for now? 

I checked again and everything seems to be working fine. I also don't think that calling the apiLink function for every request is very problematic, as there shouldn't be too many requests per second. In most cases, it shouldn't have any noticeable effect on performance. I'd recommend going with whatever feels cleanest/simplest to you. If we recognize this to be a bottleneck (because maybr of the regexp and some user with an absolutely crazy amount of issues) we can always optimize this later on.

I also just recognized that I am unable to access public(!) repositories without a token. @MostafaAmin07 Do you know if this is to be expected? If so we might want to make the token field required.

As you mention: Adjusting the translations (for the config section) is probably crucial for this change.

MostafaAmin07:
Well, nice point of view. I am working on the translations they should be available today for the merge.
After that, I will get used to how the migrations work and start implementing an optimized solution maybe with some adjustments to the GitLab integration cause as it is now it works very well with small repositories but any repo that has more than 200-300 issues makes the app really slow So I will look into this too and see how can I fix this.

**The token** part is weirdly true Thanks for the heads up! I will make it required and search the API docs on why this is the case.

MostafaAmin07:
@johannesjo I think it's ready now.

johannesjo:
Thank you very much again once more @MostafaAmin07 !

PR #443, opened 2020-07-12
Fix: preferences shortcut not working

Jackymancs4:
Hello @johannesjo 
So, on MacOS I'm accustomed to using `Cmd+,` as a shortcut for the preferences page. After manually setting the shortcut I noticed that it actually brought me to the project page.

I tracked down the problem to a wrong routing. As it can be seen here:
https://github.com/johannesjo/super-productivity/blob/master/src/app/app.routes.ts#L17

the actual route is `config`, not `settings`. Hoping not to have misinterpreted the intended behaviour this simple PR fix this.


johannesjo:
Thank you very much! I'd probably should adjust the default keyboard combos for mac, too at some point...

Minh12349:
957878

Minh12349:
Leave a comment

PR #445, opened 2020-07-14
Add product name key to package.json

Jackymancs4:
Hello @johannesjo 
This a minor Quality of Life improvement that you may want to investigate a little before merging (or not).

On Mac both the application title and the application bundle are spelled `superProductivity`, which is a little ugly and breaks the consistency.

As a solution I found that adding a `productName` key gets the job done. 

||Title Bar|App Bundle|
|---|---|---|
|Before|Schermata 2020-07-12 alle 17 52 40|Schermata 2020-07-12 alle 17 55 25|
|After|Schermata 2020-07-12 alle 17 22 09|Schermata 2020-07-12 alle 17 22 24|

For docs: 
https://stackoverflow.com/questions/41551110/unable-to-override-app-name-on-mac-os-electron-menu
https://www.electron.build/configuration/configuration#configuration

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "./electron/main.js",
   "author": "johannesjo <contact@super-productivity.com> (http://super-productivity.com)",
   "license": "MIT",
+  "productName": "Super Productivity",
   "homepage": "http://super-productivity.com",
   "repository": {
     "type": "git",


johannesjo:
That's super cool! Thank you very much! It's likely to break some of the deployment scripts, but we can fix this later on.

johannesjo:
Unfortunately, this causes a lot more issues than I had Imagined, as it changes the directories where user data is saved. I tried manually changing those to use the old ones, but I wasn't able to do that for every path as some are set by third-party plugins. Furthermore, the user data directories and file names now contain a white-space, which makes them a little awkward to deal with. Also, the entry in the snap store might be negatively affected by this change.

I'd really like to change this in the future, but I currently don't have the capacity to test all possible cases.

Jackymancs4:
I had foreseen some problems with the deployment scripts but I definetely didn't think about user data and the likes.. Too bad I guess. Sorry.

Maybe if you have time (as of now, I don't), you can try adding the `title` property (Docs: https://www.electronjs.org/docs/api/browser-window#new-browserwindowoptions) here:
https://github.com/johannesjo/super-productivity/blob/3746aa5bf63d949d80cd9034e1069bdaca66aa2d/electron/main-window.ts#L70-L83

This is just a thought.

johannesjo:
The title should be already set to "Super Productivity" (it uses the page title of the web page that is loaded). But this just affects the title bar. At least to my knowledge.

I hope we can implement this later on in a better way :)

PR #470, opened 2020-08-01
Better translation for "MISC"

Sayinore:
The former translation for "MISC" seemed just like Google Translate. I polished it. I will continue to polish the other parts when I am available.
之前“MISC”的翻译，一股子谷歌机翻味，我给润了色。有时间我也会把其他部分润色。

### DIFF
--- a/src/assets/i18n/zh.json
+++ b/src/assets/i18n/zh.json
@@ -14,14 +14,11 @@
     "UPDATE_WEB_APP": "新版本可用，是否加载新版本？"
   },
   "BL": {
-<<<<<<< HEAD
     "FROM_OTHER_PROJECTS": "来自其他项目",
     "NO_TASKS": "目前您的待办事项中没有任务",
     "SCHEDULED": "已计划",
     "TITLE": "待办事项"
-=======
     "NO_TASKS": "您的待办事项中目前没有任务"
->>>>>>> upstream/master
   },
   "DATETIME_INPUT": {
     "IN": "在 {{time}}",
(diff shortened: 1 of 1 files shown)


Sayinore:
By the way, is  internationalization in Chinese Simplified for the REAMDME.md needed? If yes, I can help :)

johannesjo:
Thank you very much! Very much appreciated! Not sure, if a Chinese readme is needed. What do you think?

Sayinore:
I don't quite think that's needed, actually. To be honest, although this app is awesome, very, very few Chinese people know it, and consequently, most Chinese *USERS* of this app really understand English used in the readme, I think.
I came up with the idea just by accident, thinking that with a readme translation this app will seem more 'international'.
If you would like more users, I could actually add a Chinese readme and advertise this app in social media websites and apps in China. But, I have to say, with all that the new users could mainly be less-abled, even than I am. (I myself feel very guilty for not being able to fix bugs here, due to my academic pressue and disability to write TypeScript.You are busy working here all the day. You are HONORABLE!) Maybe some them would not like even to submit a issue at Github when any error occurrs, and may comment 'OMG superproductivity is terrible, filled with bugs'.
What would you like?

johannesjo:
Please don't feel guilty about it! Making Chinese translations work is really awesome and very very helpful! I thought they might have been very bad before, but as don't speak a single word of Chinese I can't fix anything myself, which makes it even better that you did that! 

Sharing this on Chinese social media would be super cool as the best way to make this app better and to get more people involved is to get more people to use the app. I don't mind negative criticism. I try to see it as a chance to learn and improve the app.

Sayinore:
I have read what you said. I will work on that when I have more free time, maybe a couple of months later. Hope for a better Super-Productivity app!

johannesjo:
Alright! Please let me know if İ can be of assistance!

PR #501, opened 2020-08-15
build: upgrade to angular 10

thetric:
Upgrades to Angular 10.

todos:
- [x] ~~fix the e2e tests~~ (protractor-based e2e will be replaced with nightwatch)

bugs in prod mode:
- [x] worklog, project and settings pages are not scrollable (fixed via 375d600)
- [x] `sideEffects: false` in `src/app/package.json` breaks the task list (angular bug, see https://github.com/johannesjo/super-productivity/pull/501#issuecomment-702819486 and angular/angular#35602)


johannesjo:
Hey there! Thanks for opening this up. 

Just a quick heads-up:

If you look at the history you can see that I already tried to do this. I was however unable to make the new es2018 style work with electron (at least not without having to migrate the user data completely from file to some sort of custom webserver).

thetric:
Thank you for the heads-up!

> If you look at the history you can see that I already tried to do this. I was however unable to make the new es2018 style work with electron (at least not without having to migrate the user data completely from file to some sort of custom webserver).

My guess is that it's related to zone.js' missing support for native async/await. This message is printed in the console when configuring a ts target >= es2017:

```
 WARNING: Zone.js does not support native async/await in ES2017.
    These blocks are not intercepted by zone.js and will not triggering change detection.
    See: https://github.com/angular/zone.js/pull/1140 for more information.
```

Or was there another reason for ditching the upgrade?

johannesjo:
Not that might cause issues too, but the problem was the "Failed to load module script" error as differential loading does not work from the file system. Before it was possible to make this still work with setting the compile target to `es5` but that's not possible anymore with angular10.

Maybe we could make this work by disabling differential loading and only using es2018:
https://stackoverflow.com/questions/56197896/how-to-turn-off-differential-loading-in-angular-v8

johannesjo:
That's great news! I'm a little puzzled however as this is something electron builder should normally do. I also didn't encounter the issue while building from Ubuntu.

johannesjo:
@thetric are you still working on this? Are there still problems with Arch Linux?

thetric:
I'm still trying to really fix the blank startup screen issue instead of applying the workaround but without progress.

Another issue I've encountered is that TypeScript 3.9 brings stricter type checks that cause compile errors in the e2e tests (e.g. `e2e/src/header.po.ts`: The return type of an async function must either be a valid promise or must not contain a callable 'then' member.). The problem is that `ElementFinder`  contains a `then` method

johannesjo:
In this case, I have at least some good news. The e2e stuff is not really needed anymore, as I replaced it with (imho) better tooling in the n2e folder. We can just remove it. I just left the old stuff, because I was not sure how the new solution would prove itself in practice, but o far I am pretty satisfied.

Is your blank startup screen also caused by the 504? This is very likely related to ngsw. I was able to fix the problem by removing the service worker manually from the dev tools (1. Ctrl+Shift+i, 2. Navigate to Application/Service Workers, 3. Click unregister for the service worker). Would be good to know what caused this in the first place, but at least the issue seems unrelated to the update if I am not mistaken.

thetric:
> In this case, I have at least some good news. The e2e stuff is not really needed anymore, as I replaced it with (imho) better tooling in the n2e folder.

Good to hear. Shall I remove it now?

> Is your blank startup screen also caused by the 504? This is very likely related to ngsw. I was able to fix the problem by removing the service worker manually from the dev tools (1. Ctrl+Shift+i, 2. Navigate to Application/Service Workers, 3. Click unregister for the service worker).

Great, thank you for the fix!

johannesjo:
> Good to hear. Shall I remove it now?
That would be great! I think deleting the folder and removing `sp2-e2e` from the angular.json should do the trick. Also, protractor from the packahge.json is probably not needed anymore.

johannesjo:
@thetric is this ready for review again? Sorry, I seem to have missed that...

mrrodd1:
Give me a sec

thetric:
Here go you.

The installation of `angular2-promise-button` and `ngx-date-time-picker-schedule` issue "incorrect peer dependency" warnings due to old angular dependencies

johannesjo:
I don't know the reason why, but for the electron build (`yarn localInstall`), some things just don't seem to work. There are no tasks shown for example (even though they are present). The worklog, project and settings pages are not scrollable anymore.

I am also a little afraid because the build time is quite a bit longer now. I hope this won't be too much for Travis. I am really not happy with this release by the angular team...

thetric:
The build time might decrease a bit with the Angular CLI patch update from 10.1.1 to 10.1.2: https://twitter.com/angular/status/1306699454335123456

johannesjo:
Yes, I also read this :)

The rendering stuff is pretty weird though.

thetric:
> [...] the electron build (yarn localInstall), some things just don't seem to work. There are no tasks shown for example (even though they are present). The worklog, project and settings pages are not scrollable anymore.

Weird. This can be reproduced with `yarn startFrontend --prod`

thetric:
Not sure how this relates to the problems but with `strictTemplates: true` the Angular compiler throws type errors when using the `keyvalue` pipe:

```
    Error occurs in the template of component WorklogComponent.
src/app/features/worklog/worklog.component.html:15:24 - error TS2339: Property 'value' does not exist on type '{ key: any; val: any; }'.

15         <strong>{{year.value.monthWorked}}</strong><br>
```

thetric:
An unsatisfying, temporary workaround is to disable the `optimization` in the `angular.json` for production

johannesjo:
That's great for now. I will try using this locally for a couple of days, to see if there are any issues left. I also spent several hours debugging those issues. At least about the task-list issue, I think this issue might be related:
https://github.com/angular/angular/issues/35602
 
The ngFor inside the task-list component doesn't work at all  and even `<div *ngIf="true">TRUE</div>` and neither does using custom components like `<button mat-button>`. It seems like it is completely ejected from all existing modules including the `CommonModule` as using the async or json pipe throws an error right away.

johannesjo:
I also was able to fix the scroll issue by editing the `_scroll.scss` file:

```
@import 'media-queries';

@mixin scroll() {
  -webkit-overflow-scrolling: touch;
  overflow: auto;
  -ms-overflow-style: -ms-autohiding-scrollbar;

  :host-context(.isNoMac) & {
    @include mq(xs) {
      overflow: hidden;

      &:hover {
        overflow: auto;
        overflow: overlay;
      }

      @-moz-document url-prefix() {
        overflow: auto;
      }
    }
  }
}

@mixin scrollY() {
  -webkit-overflow-scrolling: touch;
  -ms-overflow-style: -ms-autohiding-scrollbar;
  overflow-y: scroll;
  // important for touch events like hammer swipe to work
  touch-action: pan-y;

  .isNoMac &,
  :host-context(.isNoMac) & {
    overflow-y: overlay;

    @include mq(xs) {
      overflow-y: overlay;
    }
  }

  :host-context(.isMac) & {
    overflow-y: auto;
  }
}

@mixin scrollYImportant() {
  -webkit-overflow-scrolling: touch;
  -ms-overflow-style: -ms-autohiding-scrollbar;
  overflow-x: hidden;
  overflow-y: scroll !important;
  // important for touch events like hammer swipe to work
  touch-action: pan-y;

  :host-context(.isNoMac) & {
    @include mq(xs) {
      overflow-y: hidden !important;
      &:hover {
        overflow-y: scroll !important;
        overflow-y: overlay !important;
      }

      @-moz-document url-prefix() {
        overflow-y: auto;
      }
    }
  }

  :host-context(.isMac) & {
    overflow-y: auto !important;
  }
}

@mixin scrollX() {
  -webkit-overflow-scrolling: touch;
  -ms-overflow-style: -ms-autohiding-scrollbar;
  overflow-x: scroll;

  :host-context(.isNoMac) & {
    @include mq(xs) {
      overflow-x: hidden;

      &:hover {
        overflow-x: scroll;
        overflow-x: overlay;
      }
    }
  }
}
```

For some reason this isn't working with the setup (maybe because we adjusted the supported browsers).

johannesjo:
All seems to be working fine. I'll merge this soon, but I want to wait for the current pending mac store release to go through first.

thetric:
I noticed that the app does not completely quit when closing the window. The process is still running in the background but does not terminate until stopped via Ctrl+C:

```
./app-builds/superProductivity-5.9.15.AppImage                                                                                                                                                                                                                            
{ enabled: null, showDevTools: false, mode: 'bottom' }
Saving backups to /home/thetric/.config/superProductivity/backups
close, isQuiting: undefined  # <-- attempt to close the window
^C%      # <-- manually close with Ctrl+C
```

johannesjo:
> I noticed that the app does not completely quit when closing the window. The process is still running in the background but does not terminate until stopped via Ctrl+C:

Hm this is new. Another guy was having the same problem on windows. So far I am unable to reproduce this... If you have any additional info that might help debugging this, please let me know.

thetric:
Is #567 the related issue?

johannesjo:
> Is #567 the related issue?

yes

thetric:
I noticed that `src/polyfills.ts` contains many imports that seem IE-specific:

```js
/* IE9, IE10 and IE11 requires all of the following polyfills. **/
import 'core-js/es/symbol';
import 'core-js/es/object';
import 'core-js/es/function';
import 'core-js/es/parse-int';
import 'core-js/es/parse-float';
import 'core-js/es/number';
import 'core-js/es/math';
import 'core-js/es/string';
import 'core-js/es/date';
import 'core-js/es/array';
import 'core-js/es/regexp';
import 'core-js/es/map';
import 'core-js/es/weak-map';
import 'core-js/es/set';
```

Shall we drop them or does some other browser require them?

johannesjo:
Good catch! I think we can safely drop all IE related stuff. Even Firefox and Safari are not working correctly atm. I probably should fix Safari at some point to make the webapp work on the iPhone, but I think most people are using the desktop app anyway rather than the web app.

thetric:
I disabled differential loading in 02865be and 7708a45 which drops the `ng build --prod` time for me by 1/3 from 210 seconds to about 120-140 seconds

package.json
```diff
@@ -21,12 +21,12 @@
     "ng": "ng",
     "start": "yarn electron:build && cross-env NODE_ENV=DEV electron .",
     "startFrontend": "ng serve",
-    "serveProd": "node --max_old_space_size=4096 ./node_modules/@angular/cli/bin/ng serve --prod",
```
johannesjo, line 24: :heart:  away with the clutter!

johannesjo:
That's great! Btw. the app just got approved with the mac store! So I am going to merge this. Thanks a  lot for all your hard work!

Issue #528, opened 2020-09-11
Change setting for first day of week


jfnklstrm:
When I use the web-app I have found that the first day of the week is set to sunday and that I cannot change that setting. It's a small thing but I forget about it so when I schedule things I tend to schedule them 1 day earlier than I want to because I am so used to monday being the first day of the week (this includes the date picker by the way). If this could be included in the settings I'd be really happy!

johannesjo:
Thanks for opening this up. I'd like to see this feature as well.

PR #557, opened 2020-10-02
feat: set first day of the week #528

k0rn1:
This is my attempt to implement #528.
Any feedback is welcome as I am not an Angular expert.


johannesjo:
This looks pretty good and very clean! Thank you very much!

It seems like that it doesn't cover all occurrences of weekdays, as there is also another calendar being used for the time spent dialog and also the weekdays that are being used on the worklog pages don't reflect on this (which probably is a bit tricky to implement...). But since you opened up this issue you're probably better equipped to judge this. If you don't care about these for now, I will merge this.

k0rn1:
Thanks your your quick reply :)

I have added it to the `DateAdapter` in `LanguageService` to be used by `Datepicker`.


About changing the statistics for worklogs:
* getDateRangeForWeek could easily be changed.
* IMHO this could confuse a user due to changing statistics about past shown data.
* Another thing could be to use moment for getting the week numbers and the start and end of the week. But my second point would also apply to this.
* I would suggest to make this another feature request, since the changes would be more severe.



Some moment js samples with `LanguageCodes` in constants:

```
moment().locale('fr').startOf('week').toString();
"Mon Sep 28 2020 00:00:00 GMT+0200"
moment().locale('ko').startOf('week').toString();
"Sun Sep 27 2020 00:00:00 GMT+0200"
moment().locale('ar').startOf('week').toString();
"Sat Oct 03 2020 00:00:00 GMT+0200"
```

johannesjo:
Agreed! Makes a lot of sense. :)

JobJob:
Not sure if there's an issue to track it but +1 for having this setting be used in the worklog.

> IMHO this could confuse a user due to changing statistics about past shown data.

I don't think that's an issue if the default first day of the week is whatever is used in the worklog currently (i.e. Monday). That way the worklog will only be affected if a user changes the setting.

Issue #581, opened 2020-10-16
Deleting Projects


rohanrk:
I've noticed that in order to delete a project I'm on I have to go to the manage projects page and cannot delete the currently selected project (likely a safety mechanism). Since it already takes a few clicks to get to the delete state, should the app allow deleting current projects and default to something like the topmost project?

johannesjo:
Hey there! thanks for opening this up. I am open to implementing this and I agree that it would be better to allow for this. There is a lot to do as is, however, so I am unlikely to get to it anytime in the near future, but a PR would be very welcome.

PR #595, opened 2020-10-18
Feat: Update logic to disable deletion of default project.

rohanrk:
# Description

Allows users to delete current project in navigation menu. Disallows users from deleting default project.

There's a potential issue with this. If users have already deleted the default project then this will cause issues if users delete all their projects. Sometimes the app crashes when the default project no longer exists. The issue an be fixed with a cache reload but that should be avoided as it wipes all projects from users' storage.

I'm completely new to angular, so I did some reading and I'm wondering if there's a way to get the number of projects? A better way to do this is to disable the delete button if there's only one project remaining but I didn't see anything in the template class or the model class that does this. Is there a way to get this from the service?

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/581

## Check List

### DIFF
--- a/src/app/pages/project-overview-page/project-overview-page.component.html
+++ b/src/app/pages/project-overview-page/project-overview-page.component.html
@@ -53,7 +53,7 @@
           <!--            <mat-icon>flight</mat-icon>-->
           <!--          </button>-->
           <button (click)="remove(project.id)"
-                  [disabled]="project.id === (workContextService.activeWorkContextId$|async)"
+                  [disabled]="project.id === defaultId"
                   [title]="T.PP.DELETE_PROJECT|translate"
                   color="warn"
                   mat-icon-button>
--- a/src/app/pages/project-overview-page/project-overview-page.component.ts
+++ b/src/app/pages/project-overview-page/project-overview-page.component.ts
@@ -15,6 +15,7 @@ import { THEME_COLOR_MAP } from '../../app.constants';
 import { WorkContextService } from '../../features/work-context/work-context.service';
 import { withLatestFrom } from 'rxjs/operators';
 import { ExportedProject } from '../../features/project/project-archive.model';
+import { DEFAULT_PROJECT_ID } from '../../features/project/project.const'
 
 @Component({
   selector: 'project-page',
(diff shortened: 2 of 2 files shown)


src/app/pages/project-overview-page/project-overview-page.component.html
```diff
@@ -53,7 +53,7 @@ <h1 class="mat-h1">{{T.PP.TITLE|translate}}</h1>
           <!--            <mat-icon>flight</mat-icon>-->
           <!--          </button>-->
           <button (click)="remove(project.id)"
-                  [disabled]="project.id === (workContextService.activeWorkContextId$|async)"
```
johannesjo, line 56: That's not right. The current project should not be delectable, as this might lead to errors. The default project should be delectable however.

rohanrk: Oh. Then I guess this issue: https://github.com/johannesjo/super-productivity/issues/581 is incorrect?

I was referring to deleting the current project because otherwise it takes more clicks if you want to delete it. If that's more of an intended thing then this PR and the corresponding issue should be closed.

johannesjo:
> Sometimes the app crashes when the default project no longer exists.

What kind of error did you encounter? We should focus on hat error. Deleting the currently active project can have side effects, deleting the default project shouldn't have and if it does it is something we should fix.

src/app/pages/project-overview-page/project-overview-page.component.html
johannesjo: No #581 is correct, but to solve this, one would have to make sure that it is safe to delete a currently active project, which could be done, by making it inactive (e.g. by switching to the today list) before deleting it.

johannesjo:
Thanks for taking the time to open this up. I like the hands-on-mentality! We should however discuss what needs to be changed.

1. If you encountered any error with deleting the default project, please let me know. We should fix it!
2. For solving 581 and to allow for deleting the current project, we need to make sure that are no errors when deleting the current project. As far as I can see the best way to do so is to switch to the today list before triggering the delete.

johannesjo:
@rohanrk ping.

johannesjo:
No worries :)

johannesjo:
I am closing this for now. Please feel free to reopen this! :)

Issue #497, opened 2020-08-13
Fix delete a keyboard shortcut


theCalcaholic:
Sorry for abusing the issue tracker for questions. I recently accidentally created a global keyboard shortcut and can't find a way to get rid of it.
I can reassign it, but don't find a way to delete it entirely. However, there must be a way, right? :)

johannesjo:
No worries: That's currently the best place to ask things.

I thought that I had configured the Escape key to do that. That doesn't seem to work anymore, however.

theCalcaholic:
Alright, thank you :+1:

PR #601, opened 2020-10-22
fix: delete a keyboard shortcut

k0rn1:
# Description
Use the escape key to remove bound keyboard shortcuts. I also thought that removing the focus would benefit UX.
I have tested the following:
* Removing a shortcut and the shortcut is not applied anymore.
* Setting a new shortcut afterwards again.
## Issues Resolved

closes issue #497

## Check List


johannesjo:
Thank you very much! Works nicely! :)

PR #610, opened 2020-10-26
Missing space between buttons

epavanello:
# Description

Missing space between "Create new" and "Import from file" under "Create Project" section

## Check List


johannesjo:
Thank you very much for opening this up! This might look a little off on mobile though (which it already does to be fair :)), so maybe we need to apply half the margin to all sides of both buttons.

johannesjo:
Thank you for the adjustments!

epavanello:
Half the margin to all sides of both buttons done

Issue #622, opened 2020-10-28
Strip trainling slash from Jira URL


gohrner:
### Your Environment
* Version used: 6.0.1 .deb
* Operating System and version: KDE Neon (Ubuntu 20.04)
* Desktop Environment: KDE
* Browser Name and version: --

### Expected Behavior
I configured my Jira URL like this:

```
https://example.com/jira/
```

It should either "just work", or a meaningful error message should be shown.

### Current Behavior
super-productivity just shows a message box: "Jira: Not Found".

Investigation shows that the "Not Found" is actually a 404 message returned by the server - super-productivity appends an extra slash, which apparently confuses the server:

```
https://example.com/jira//rest/api/latest/myself
```
(from the console log)

If I remove the trailing slash, it works:

```
https://example.com/jira
```
=>
```
https://example.com/jira/rest/api/latest/myself
```

and the login attempt succeeds.

However, the user avatar in the login test is not shown properly:

![image](https://user-images.githubusercontent.com/26199042/97429485-c54c8280-1917-11eb-83e3-239879513744.png)

johannesjo:
Makes a lot of sense! Thanks for opening this up!

PR #631, opened 2020-10-29
fix: strip trailing jira host slash

k0rn1:
# Description

Strips possible trailing slash in the host path of jira.

I tested the `stripTrailing` itself, but since I don't have an access to jira itself, I could not test this part.
I was not sure about the handling of `cfg.host` possibly being `null`, so I used `'null'` as a fallback (as would have happened before).

## Issues Resolved

fixes issue #622 

## Check List


johannesjo:
Thank you very much!

johannesjo:
Just for your information:
You can create free trial instances of Jira here: https://www.atlassian.com/de/software/jira/free

While the null thing probably wasn't pretty to begin with I am alright with the solution :)

PR #638, opened 2020-11-01
Add AUR PKGBUILD and GitHub Action

alosarjos:
# Description

Add the actions and the PKGBUILD file required to autopublish updates on AUR for Arch Linux users.

There are some previous requirements needed for this to work:

1. Create an AUR user account, I need to add it as co-maintainer of the package
2. Create a new SSH key pair (Without password) and add the public key to your AUR user
3. Set up the required Secrets:
          AUR_USERNAME 
          AUR_EMAIL
          AUR_SSH_PRIVATE_KEY 

Once we have this, the PKGBUILD file should be updated to include you as the new co-maintainer (It's recommended to use some aliased email account to avoid possible spam from bots).

I'm by no means any expert in either GitHub Actions or PKGBUILD files. I tested the GitHub action with a manual trigger, but I'm not sure if the "When new release" trigger is correctly setup.

The PKGBUILD right now would require manual intervention to update the vars `pkgver` and `md5sums`. I've seen some other repos using Python scripts to modify the PKGBUILD file on the CI, maybe you have a better idea to get those from CI Env Vars or some automated script to update things before releasing?


johannesjo:
This is great José! Thank you so much! The build trigger looks fine to me.

Just one minor thing: Can we use the `build/linux/PKGBUILD` instead of `linux/PKGBUILD`

I'll create a new account and secrets and get back to you.

alosarjos:
Done, no problem.

alosarjos:
Now that I think about it, this action would need to be triggered after the package has been released on GitHub and not on parallel with it to avoid any kind of problem, so we need to trigger this action after the build & publishing have finished.

johannesjo:
Thank you! And since you're online right now. Could you help me with their security mechanism? I don't want to create a new VM just for that:
`LC_ALL=C pacman -V|sed -r 's#[0-9]+#e21#g'|md5sum|cut -c1-6`

alosarjos:
`26ffb1`

johannesjo:
> Now that I think about it, this action would need to be triggered after the package has been released on GitHub and not on parallel with it to avoid any kind of problem, so we need to trigger this action after the build & publishing have finished.

Yes that's exactly what the trigger does. I use it for building and publishing the windows store version and the mac store version as well.

johannesjo:
To make it clearer: The build should only be triggered if I create a new release on GitHub. I created all the secrets. I'll send you an email with the username and the email addressl.

> The PKGBUILD right now would require manual intervention to update the vars `pkgver` and `md5sums`. I've seen some other repos using Python scripts to modify the PKGBUILD file on the CI, maybe you have a better idea to get those from CI Env Vars or some automated script to update things before releasing?

No idea. Maybe we can ask the maintainers of the github action?

johannesjo:
> The PKGBUILD right now would require manual intervention to update the vars `pkgver` and `md5sums`. I've seen some other repos using Python scripts to modify the PKGBUILD file on the CI, maybe you have a better idea to get those from CI Env Vars or some automated script to update things before releasing?

Thought about it once more: The current version should be available in the package.json of the project. Is the md5 sum the one of the deb package?

alosarjos:
I've added you as a co-maintainer of the app and added the comment to the PKGBUILD. If the secrets and the SSH key on your AUR account are set up, that should do the trick.

The maintainer of the GitHub action has as an example this Python Script:
https://github.com/KSXGitHub/pretty-exec/blob/master/ci/github-actions/generate-pkgbuild.py3

He is skipping adding the checksum of the package, we could skip it until we find a way to automatically get it (Yes, it would be the .deb file MD5). For the package version I guess we could try doing it the way he does with a script? He seems to be setting up the version as an environment variable or it would require another Checkout action inside the AUR action to get the files and read the version from the `package.json`?

johannesjo:
Yes, then let's skip the md5 checksum for now. 

I can think of three ways to get the version number:
1. It should be available on `github.ref` as in `startsWith(github.ref, 'refs/tags/v')`.
2. We can use this package https://github.com/martinbeentjes/npm-get-version-action (probably overkill to add another dependency for this)
3. We can use a custom run block like so:
```
      - run: 'VER=$(cat ./package.json | grep version | head -1 | awk -F: '{ print $2 }' | sed 's/[",]//g' | tr -d '[[:space:]]')
        shell: bash
```

alosarjos:
Ok, that took way more than I expected. (Had to learn about how Github actions environment vars work). But that should do the trick.

johannesjo:
Much appreciated! Thank you very much.

PR #648, opened 2020-11-03
Fix feature request template

sdruskat:
- Comment was bleeding into heading, fix by adding a new line before the comment.

# Description

Fixes an issue where the HTML comment started on the same line as the heading. If people don't change that manually, the `<--` is being rendered as part of the heading.

## Issues Resolved

n/a

## Check List


johannesjo:
Interesting quirk of the markdown parsing here. Thank you very much for fixing it!

PR #762, opened 2021-01-05
bug: Adding worklog to Jira is converting starts to Zulu time #757

### LINKED ISSUE #757: Adding worklog to Jira is converting starts to Zulu time
### Your Environment
* Version used: 6.3.2
* Operating System and version: MacOs 11.1
* Browser Name and version: Application
* Using Jira server 8.10
### Expected Behavior
The started time of a worklog should be expressed in local time.

### Current Behavior
When I add a worklog to Jira the dialog window that shows worked time, started and comment... shows the started time at Greenwich TimeZone.
For instance, if you add a new subtask to a task imported from Jira and called it "new subtask created at 10:00" (+01:00, in my case), start this subtask and mark as done a few minutes after. The dialog window will show that subtask started at 09:00 because it converts '...10:00+01:00' to '...09:00Z' and removes tailing 'Z'.

1. Configure Project to open dialog when task and subtask is done
2. Start a subtask
3. Mark it as done, dialog opens showing a Started date using Greenwich timezone

Digging into code, I think the problem (if there is one) is in src/app/features/issue/providers/jira/jira-view-components/dialog-jira-add-worklog/dialog-jira-add-worklog.component.ts

Original code convert timestamp to Z timezone (using toISOString() )
```javascript
  private _convertTimestamp(timestamp: number): string {
    const date = new Date(timestamp);
    date.setSeconds(0, 0);
    const isoStr = date.toISOString();
    return isoStr.substring(0, isoStr.length - 1);
  }
```

I think it can be resolved using moment.
```javascript
  private _convertTimestamp(timestamp: number): string {
    const date = moment(timestamp);
    const isoStr = date.seconds(0).local().format();
    return isoStr.substring(0, 19);
  }
```

cufiler:
bug: Adding worklog to Jira is converting starts to Zulu time #757

# Description

_convertTimestamp() is replaced. Previous version used GMT time to convert it to string and removed last char ('Z' = Zulu time). The change uses moment to be able to show timestamp in local timezone. 

## Issues Resolved

#757

### DIFF
--- a/src/app/features/issue/providers/jira/jira-view-components/dialog-jira-add-worklog/dialog-jira-add-worklog.component.ts
+++ b/src/app/features/issue/providers/jira/jira-view-components/dialog-jira-add-worklog/dialog-jira-add-worklog.component.ts
@@ -7,6 +7,7 @@ import { Task } from '../../../../../tasks/task.model';
 import { T } from '../../../../../../t.const';
 import { ProjectService } from '../../../../../project/project.service';
 import { first } from 'rxjs/operators';
+import * as moment from 'moment';
 
 @Component({
   selector: 'dialog-jira-add-worklog',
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much!

PR #766, opened 2021-01-06
Allow port to be customized in dockerfile

moranje:
# Description

This allows for a custom port to be exposed.

## Issues Resolved

Port 80, the currently default port is in use on the NAS I use and I'm experiencing a crash when I try to override the port manually


johannesjo:
Thank you very much!

PR #769, opened 2021-01-06
Allow to use self hosted Gitlab and add Project ID support.

### LINKED ISSUE #337: Allow on premise gitlab instance (selfhosted gitlab)
### Your Environment

* Version used: latest debian package : 5.0.10
* Operating System and version: Ubuntu 20.04
* Desktop Environment: Gnome-Shell 3.36

### Expected Behavior
- When I had a project, I can setup gitlab integration just by setting username/project
- It could be awesome if we could use a custom gitlab url

### Current Behavior
- Instead a username/project, we should add full url. 
- If full url is not written, just try to match username/project string

b0c1:
# Description

Separate GitLab URL for the project because some self-hosted GitLab not installed on the root, and added project id support (our GitLab did not work with project names)

## Issues Resolved

#337 

## Check List

- [x] Manual test on gitlab.com and self-hosted GitLab.
- [ ] Translations needed in other languages


johannesjo:
Excellent work! Thank you very much! 

One thing that might be missing: A little migration script to assign the default value inside `src/app/features/project/migrate-projects-state.util.ts`, because otherwise the base url might also be `undefined` instead of `null`. Alternatively we can change the code around line 217 to something like that:
``` 
    const base: string = projectConfig.gitlabBaseUrl || GITLAB_BASE_URL;
    const fixedUrl = base.match(/.*\/$/)
      ? projectConfig.gitlabBaseUrl
      : `${projectConfig.gitlabBaseUrl}/`;
    apiURL = fixedUrl + 'api/v4/projects/';
```


Maybe we can also add "Custom" to the form field caption to indicate that otherwise a default is used?

b0c1:
> 
> 
> Excellent work! Thank you very much!
> 
> One thing that might be missing: A little migration script to assign the default value inside `src/app/features/project/migrate-projects-state.util.ts`, because otherwise the base url might also be `undefined` instead of `null`, which leads to an error. Alternatively we can change the code around line 217 to something like that:
> 
> ```
>     const base: string = projectConfig.gitlabBaseUrl || GITLAB_BASE_URL;
>     const fixedUrl = base.match(/.*\/$/)
>       ? projectConfig.gitlabBaseUrl
>       : `${projectConfig.gitlabBaseUrl}/`;
>     apiURL = fixedUrl + 'api/v4/projects/';
> ```
> 
> And maybe we can also add "Custom" to the form field caption to indicate that otherwise a default is used?

Ok, I changed the caption and replaced it with ```!==``` to ```!=``` because it will match to ```undefined``` too.

johannesjo:
Alright! Thank you very much!

Issue #496, opened 2020-08-12
Add setting to Git**b integration to only add assigned issues to backlog


sdruskat:
### Your Environment

* Version used: 5.5.7
* Operating System and version: Win 10
* Desktop Environment: n/a
* Browser Name and version: n/a

### Expected Behavior

To focus on the tasks that I *actually* need to do, it'd help me to only see issues in GitHub/GitLab integrated projects' backlog which are actually assigned to me.

I appreciate that this may not be something that all users would want to have as default, therefore I think a setting in the respective project would be the right place to configure.

If this is switched on, I would only see issues/PRs assigned to my username in the backlog, so that I can pull them into Today.

### Current Behavior

All issues/PRs are listed to a project's backlog.

### Steps to Reproduce (for bugs)

n/a

### Console Output

n/a

### Error Log (Desktop only)

n/a

johannesjo:
Thanks for opening this up. Makes a lot of sense to me. Currently, I am busy with other features, but I gladly accept any PR.

sdruskat:
Great, thanks. I understand. I have little experience with JS, but if I can find the time I'd love to put some effort into this.

johannesjo:
@sdruskat that's great to hear! Please let me know if you have any questions.

b0c1:
What's is the plan, if the issue moved to another assignee? Remove to the backlog?

johannesjo:
Well I assume @sdruskat is no interested in doing it anymore (right @sdruskat ?). So if somebody else wants to tackle this, I am all for it!!

b0c1:
I already completed the filtered auto add. The question is what happened the backlog elements with assigned to nobody or later assigned to somebody?

johannesjo:
Ah ok. Misunderstood your question. There currently is no mechanism for that for the Git* integrations. The jira integration checks once whenever you start tracking on a Jira Task if another user was assigned to the issue. Maybe this could be an addition for the Git* integrations as well? 

Personally I use Gitlab and Github only for version control and as a sort of bug tracker, as there is no currently no team which would require a more sophisticated approach, so this never popped up as a problem for me.

b0c1:
In this case, what's the best way to remove the backlog tasks? I try to use ```taskService.removeMultipleMainTasks``` but it creates an inconsistent state on the project.

sdruskat:
[...]

PR #776, opened 2021-01-09
Issue backlog, link fix and cache

### LINKED ISSUE #772: Gitlab issue url wrong
### Your Environment

* Version used: master branch
* Operating System and version: windows 10
* Desktop Environment: 
* Browser Name and version: firefox

### Expected Behavior
open gitlab issue in the browser

### Current Behavior
Gitlab issue link wrong when the project id set.

1. Set project ID instead of url
2. Fetch Gitlab issues to backlog 
3. open browser

b0c1:
# Description

- It fixes the wrong issue linking
- Create project-level cache (localStorage)
- Store GitHub and GitLab user in project level (1 day expire)
- GitLab and GitHub filtered backlog integration with remove (remove the task from the backlog if the assignee changed (only if the issue is in the backlog)

## Issues Resolved

#772 #496 

## Check List


johannesjo:
For existing gitlab configurations the url leads to having "projects" two times in it. E.g.:
https://gitlab.com/api/v4/projects/projects/johannesjo%2Ftest-public/issues/1

johannesjo:
Also adding issues via the search bar does not seem to work anymore. Interestingly there are no requests fired.

src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts
```diff
@@ -21,19 +23,15 @@ export class GitlabCommonInterfacesService implements IssueServiceInterface {
...
     return this._getCfgOnce$(projectId).pipe(
-      map((cfg) => {
-        if (cfg.gitlabBaseUrl != null) {
-          const fixedUrl = cfg.gitlabBaseUrl.match(/.*\/$/) ? cfg.gitlabBaseUrl : `${cfg.gitlabBaseUrl}/`;
-          return `${fixedUrl}${cfg.project}issues/${issueId}`;
-        } else {
-          return `${GITLAB_BASE_URL}${cfg.project?.replace(/%2F/g, '/')}/issues/${issueId}`;
-        }
-      })
+      flatMap((cfg) =>
```
johannesjo, line 32: I would prefer not to make this change, as even with caching in place, these are requests which are not needed most of the time. It should be possible to derive the url from the id somehow.

src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts
```diff
@@ -66,10 +64,12 @@ export class GitlabCommonInterfacesService implements IssueServiceInterface {
 
     const cfg = await this._getCfgOnce$(task.projectId).toPromise();
     const issue = await this._gitlabApiService.getById$(+task.issueId, cfg).toPromise();
-
+    const user = await this._issueCacheService.projectCache<GitlabUser>(task.projectId, 'GITLAB_USER', duration({days: 1}), () => {
```
johannesjo, line 67: Can you maybe explain in a sentence or two why you switched from the current method of providing a custom string, to making an additional request? Is this just for convenience or is there another reason?

johannesjo (changes requested):

src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts
b0c1: I tried many ways with the project ID without any luck. I have another idea, I'll implement as soon as possible (be patient ;). Another question, how can I add a hook to the GitLab/GitHub save method to remove cached variables (example: access token changed) ?

src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts
b0c1: It will not make another request if the user is in the cache. If it's expired or not in the cache it will fetch the user. The static "login" field is a bad concept:
1. if the user deleted and another user created with the same login name it will make the wrong result (example comment filtering) that's why I changed username/login name to id.
2. minimalize the user input is the right way I think.

b0c1:
> 
> 
> Also adding issues via the search bar does not seem to work anymore. Interestingly there are no requests fired.

Which taskbar do you think?
I tested with the current version and no difference.
- the standard "+" -> Add task list the backlog items
- the reminder add to the task list only the tasks of the current working context active tasks (not the backlog)

src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts
b0c1: new solution pushed: now I fetch the project (cached, 1 day expire) and get the "prefix" from it
ps: maybe we can create a "clear cache" button to clear cached variables of the project

src/app/features/issue/providers/gitlab/gitlab-api/gitlab-api.service.ts
```diff
@@ -9,20 +9,31 @@ import { HANDLED_ERROR_PROP_STR } from 'src/app/app.constants';
...
+    private _issueCacheService: IssueCacheService,
     private _http: HttpClient,
   ) {
   }
 
+  getCurrentUser$(cfg: GitlabCfg): Observable<GitlabUser> {
+    return this._sendRequest$({
+      url: `${this.apiLink(cfg)}user`
+    }, cfg)
+      .pipe(
```
johannesjo, line 34: We can remove the empty pipes I think.

johannesjo:
> Which taskbar do you think?

I mean the add task bar (shift+A). 

I recommend creating a new gitlab project with a Super Productivity version from before the changes you made and then test your changes with that (without adjusting the settings for it). With that you should be able to reproduce the errors I found.

b0c1:
Hi, I tested in the following way:
- I deleted every store
- fetched the master branch
- created two new projects (one GitHub, one GitLab) set up (GitLab not work for me because the 'id' type is wrong in master). 
- GitHub project fetch completed (GitLab not ofc)
- stop the ng and electron
- switch to my branch
- start ng and electron
- wait until the fetch
Result:
- shift+A works as expected in both GitLab and GitHub project (opened the 'add task' bar, and show the backlog element of the current project)
- GitLab issue fetched to the backlog
- GitLab project works as expected

johannesjo:
Thanks for the changes! I try to review it tomorrow!

johannesjo:
Thanks for your efforts. I merged it because it already became quite big. If we need to make adjustments we can do this in another PR.

johannesjo:
I just fixed the the bug mentioned above (574ac8b3). Apart from that everything seems to be working fine.

PR #806, opened 2021-01-17
Feat/803 eslintmigration

### LINKED ISSUE #803: Migration from TSLint to ESLint
## Problem Statement
This project uses TSLint, but it's deprecated. You can see the deprecation notice when you run `npm run lint`. See here for more information: https://github.com/palantir/tslint/blob/master/README.md

TSLint should be replaced with ESLint. See here for the official migration path: https://github.com/angular-eslint/angular-eslint#migrating-from-codelyzer-and-tslint

It should also be considered that it's not guaranteed that the same exact linting behaviour is maintained after the migration, there could be some differences, but it's probably a necessary one to do nonetheless.

undrivendev:
# Description

Perform migration from TSLint to ESLint

## Issues Resolved

fixes #803


johannesjo:
<3

PR #825, opened 2021-01-21
feat(eslint): enable ban-types eslint rule

undrivendev:
# Description
Enable ban-types rule after migrating from TSLint to ESLint

### DIFF
--- a/src/app/core/migration/migration.service.ts
+++ b/src/app/core/migration/migration.service.ts
@@ -135,7 +135,7 @@ export class MigrationService {
   private _addProjectIdToEntity(
     pids: string[],
     entityProjectStates: { [key: string]: EntityState<any> },
-    additionalChanges: {} = {}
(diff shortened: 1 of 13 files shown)


src/app/core/migration/migration.service.ts
```diff
@@ -135,7 +135,7 @@ export class MigrationService {
   private _addProjectIdToEntity(
     pids: string[],
     entityProjectStates: { [key: string]: EntityState<any> },
-    additionalChanges: {} = {}
+    additionalChanges: Record<string, unknown> = {}
```
johannesjo, line 138: Oh that's cool! I didn't even know about this one!

undrivendev: haha, me neither! It was proposed as a solution in the resulting warnings description... and so I learned about it and used it ^^

PR #842, opened 2021-01-29
Feat/edit project

### LINKED ISSUE #785: Add context menu for project panel
## Problem Statement
Changing project name or some basic settings require some mouse click, it will be much better if you can right-click project and change it name there.

stkofler:
# Description

This adds an "edit project" entry to the three vertical dots menu in the sidenav projectlist and in the main header.
The "edit project" entry simply opens the create-project dialog whilst passing the selected project, just like from the project-overview page.
This allows editing name, color and jira etc. from everywhere without going to the project-overview page. 

## Issues Resolved

#785


johannesjo:
Thank you very much once more. This might be a relatively small change, but I think that it is a huge improvement!

PR #847, opened 2021-01-29
feat(eslint): re-enabled no-unused-expressions rule and fixed errors

undrivendev:
# Description
Re-enabled no-unused-expressions rule.

### DIFF
--- a/src/app/features/tasks/task/task.component.ts
+++ b/src/app/features/tasks/task/task.component.ts
@@ -306,9 +306,12 @@ export class TaskComponent implements OnInit, OnDestroy, AfterViewInit {
     if (this.task.parentId) {
       this.focusNext(true);
     }
-    this.task.isDone
-      ? this._taskService.setUnDone(this.task.id)
-      : this._taskService.setDone(this.task.id);
+
+    if (this.task.isDone) {
+      this._taskService.setUnDone(this.task.id);
+    } else {
+      this._taskService.setDone(this.task.id);
+    }
   }
 
   showAdditionalInfos() {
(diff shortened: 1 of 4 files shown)


src/app/features/tasks/task/task.component.ts
```diff
@@ -306,9 +306,12 @@ export class TaskComponent implements OnInit, OnDestroy, AfterViewInit {
     if (this.task.parentId) {
       this.focusNext(true);
     }
-    this.task.isDone
-      ? this._taskService.setUnDone(this.task.id)
```
johannesjo, line 310: That's sad. I really liked the simplicity of this :D But I suppose there is no way around this?

johannesjo: Well I'm merging this anyway. Thank you very much once more! I am also deleting the develop branch. I think we can proceed with the master branch as you said.

undrivendev: > That's sad. I really liked the simplicity of this :D But I suppose there is no way around this?

Well, I also never used the ternary operator this way, but I admit it's pretty concise (thank you for that :) ). Actually, looking at the eslint docs for that rule it seems that we could allow those expressions by setting the `allowTernary` option to `true`: https://eslint.org/docs/rules/no-unused-expressions . But I don't really know if that could allow some other potentially dangerous expressions that should be avoided. What do you think? Should we restore the code as it was and setting that option to true?

johannesjo: Thanks for looking into this! Let's keep it this way! Thinking about this I can picture pretty weird ways of using this, so this might be preferable.

PR #855, opened 2021-01-31
feat(eslint): re-enable max-length rule and fix errors

undrivendev:
# Description
re-enable max-length ESLint rule and fix errors. Related to #803.

### DIFF
--- a/src/app/features/issue/issue.service.ts
+++ b/src/app/features/issue/issue.service.ts
@@ -120,7 +120,11 @@ export class IssueService {
     }
 
     for (const issuesType of Object.keys(tasksIssueIdsByIssueType)) {
-      const updates = await (this.ISSUE_SERVICE_MAP[issuesType].refreshIssues as any)(tasksIssueIdsByIssueType[issuesType], isNotifySuccess, isNotifyNoUpdateRequired);
+      const updates = await (this.ISSUE_SERVICE_MAP[issuesType].refreshIssues as any)(
+        tasksIssueIdsByIssueType[issuesType],
+        isNotifySuccess,
+        isNotifyNoUpdateRequired
+      );
       if (updates) {
         for (const update of updates) {
           if (this.ISSUE_REFRESH_MAP[issuesType][update.task.issueId]) {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -51,14 +51,15 @@
             "format": ["PascalCase"]
           }
         ],
-
-        // TODO: enable gradually
         "prefer-const": "warn",
         "@typescript-eslint/no-unused-expressions": "warn",
         "@typescript-eslint/no-empty-interface": "warn",
         "max-len": [
-          "off",
+          "warn",
           {
+            "ignorePattern": "^import \\{.+;$", // exclude import statements
+            "ignoreRegExpLiterals": true,
+            "ignoreStrings": true,
             "code": 150
           }
         ],
--- a/src/app/core/persistence/persistence.service.ts
+++ b/src/app/core/persistence/persistence.service.ts
@@ -140,7 +140,8 @@ export class PersistenceService {
     'obstruction',
   );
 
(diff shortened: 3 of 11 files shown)


src/app/features/issue/issue.service.ts
```diff
@@ -120,7 +120,9 @@ export class IssueService {
     }
 
     for (const issuesType of Object.keys(tasksIssueIdsByIssueType)) {
-      const updates = await (this.ISSUE_SERVICE_MAP[issuesType].refreshIssues as any)(tasksIssueIdsByIssueType[issuesType], isNotifySuccess, isNotifyNoUpdateRequired);
+      const updates = await (this.ISSUE_SERVICE_MAP[issuesType].refreshIssues as any)(tasksIssueIdsByIssueType[issuesType],
```
johannesjo, line 123: This is a bit ugly I'd say. maybe we can set the max line length to a more convenient value as a quick fix? Alternatively I much prefer this:
```TypeScript
  const updates = await (this.ISSUE_SERVICE_MAP[issuesType].refreshIssues as any)( 
        tasksIssueIdsByIssueType[issuesType],
        isNotifySuccess,
        isNotifyNoUpdateRequired
  );
```

src/app/features/tag/store/tag.effects.ts
```diff
@@ -237,7 +237,8 @@ export class TagEffects {
     tap((arg) => console.log('Error INFO Today:', arg)),
     tap(({activeId, allTasks}) => {
       const allIds = allTasks.map(t => t && t.id);
-      const r = confirm('Nooo! We found some tasks with no data. It is strongly recommended to delete them to avoid further data corruption. Delete them now?');
+      const r = confirm('Nooo! We found some tasks with no data.\
```
johannesjo, line 240: I don't think that max line length adds much value here. I would ignore this line.

src/app/imex/sync/check-for-update.util.ts
```diff
@@ -38,7 +38,9 @@ export const checkForUpdate = (params: { remote: number, local: number, lastSync
     } else if (lastSync < local) {
       return UpdateCheckResult.RemoteUpdateRequired;
     } else if (lastSync === local) {
-      alert('Sync Warning: Dropbox date not up to date despite seemingly successful sync. (This might happen when: 1. You have conflict changes and decide to take the local version. 2. You open the other instance and also decide to use the local version.)');
+      alert('Sync Warning: Dropbox date not up to date despite seemingly successful sync.\
```
johannesjo, line 41: I also would rather ignore this line.

src/app/util/model-version.ts
```diff
@@ -10,7 +10,8 @@ export const isMigrateModel = (modelData: any, localVersion: number, modelType:
   } else if (importVersion > localVersion) {
     const isNewMajor = ((Math.floor(importVersion) - Math.floor(localVersion)) >= 1);
     if (isNewMajor) {
-      alert('Cannot load model. Version to load is newer than local. Please close the app and update your local productivity version first, before importing the data.');
+      alert('Cannot load model. Version to load is newer than local. Please close\
```
johannesjo, line 13: Same thing.

johannesjo (changes requested):

src/app/features/tag/store/tag.effects.ts
undrivendev: What about disabling the max-length rule only for strings? There is a configuration option available for that: https://eslint.org/docs/rules/max-len#ignorestrings

undrivendev: Should be ok now, I disabled the max-length for strings

src/app/features/issue/issue.service.ts
undrivendev: Should be ok now

src/app/imex/sync/check-for-update.util.ts
undrivendev: Should be ok now, I disabled the max-length for strings

src/app/util/model-version.ts
undrivendev: Should be ok now, I disabled the max-length for strings

PR #867, opened 2021-02-04
feat: persist project/tag list-state through app reloads #767

### LINKED ISSUE #767: Add option to allow auto-unfolding of the Projects folder on the sidebar
## Problem Statement
Every time a sync occurs, the page refreshes and then the Projects list is folded again. I have to unfold it every time. 
image

Would be nice if the state of the fold/unfold persists through a refresh

Another option might be to add an option auto-unfold on-load 

Thank you for your consideration! Love the app so far. 👍

stkofler:
# Description

As suggested by Johannes this would store the state of the project and tag list in localStorage and therefore persists it through app reloads. I hope this is enough to resolve the issue. Otherwise just let me know... :)

## Issues Resolved

#767

## Check List


johannesjo:
Thank you very much! I'll check this tomorrow!

src/app/core-ui/side-nav/side-nav.component.ts
```diff
@@ -154,9 +152,26 @@ export class SideNavComponent implements OnDestroy {
     this.scrollToNotes.emit();
   }
 
+  get isProjectsExpanded() {
```
johannesjo, line 155:
```suggestion
  private get isProjectsExpanded() {
```

src/app/core-ui/side-nav/side-nav.component.ts
```diff
@@ -154,9 +152,26 @@ export class SideNavComponent implements OnDestroy {
...
+    return (localStorage.getItem(LS_IS_PROJECT_LIST_EXPANDED) === 'true');
+  }
+
+  set isProjectsExpanded(isExpanded: boolean) {
```
johannesjo, line 159:
```suggestion
  private set isProjectsExpanded(isExpanded: boolean) {
```

src/app/core-ui/side-nav/side-nav.component.ts
```diff
@@ -154,9 +152,26 @@ export class SideNavComponent implements OnDestroy {
...
+    localStorage.setItem(LS_IS_PROJECT_LIST_EXPANDED, isExpanded.toString());
+  }
+
+  get isTagsExpanded() {
```
johannesjo, line 163:
```suggestion
  private get isTagsExpanded() {
```

src/app/core-ui/side-nav/side-nav.component.ts
```diff
@@ -154,9 +152,26 @@ export class SideNavComponent implements OnDestroy {
...
+    return (localStorage.getItem(LS_IS_TAG_LIST_EXPANDED) === 'true');
+  }
+
+  set isTagsExpanded(isExpanded: boolean) {
```
johannesjo, line 167:
```suggestion
  private isTagsExpanded(isExpanded: boolean) {
```

johannesjo (commented):
Looks good to me! Only adjustment I would make is making the getters and setters private to indicate that use within the view is discouraged (it's better to cache the value in memory if we were to do this, as using a getter in a view can lead to it being triggered a lot).

stkofler:
Thank you very much for your input. Had a good read about it [here](https://medium.com/showpad-engineering/why-you-should-never-use-function-calls-in-angular-template-expressions-e1a50f9c0496).  I hope I understood you correctly and added a commit to cache the value locally, as you had it before...

johannesjo:
Thanks for the adjustments! The original version was absolutely fine and elegant, but to be honest I prefer it like this, because we're less likely to run unintentionally into the problem discussed in the article  :)

Thank you very much!

PR #885, opened 2021-02-08
feat(worklog-export): add description column

### LINKED ISSUE #795: RQ: Include task descriptions in worklog export
## Problem Statement
It would be helpful to include the description of the task as part of the worklog columns

Include another column that takes values from the description field of tasks and adds it to the worklog row for the corresponding task

--

--

stkofler:
# Description

This adds the task description as a column in the worklog export dialog. 

A couple of things to consider:
1. When multiple tasks are in one row, descriptions are separated by a pipe " | ".  Just like the task titles.
Keep in mind that there might be more tasks than descriptions and therefor the order of descriptions does not match the order of tasks. I tried to fill empty descriptions with dashes to keep the order matching, but I think it didn't look too good.

2. When a description contains a newline it gets replaced by a dash " - ". This way it doesn't mess with the csv.
I think, LibreOffice Writer, MS Excel or MS Excel for Mac might all have different symbols for newlines _within a cell_...
Maybe this implementation with dashes instead of newlines is enough for now?
     
3. I added a new string in en.json. How is translation handled? Do I have to translate it to all languages?

PS: what happened to the develop branch?

## Issues Resolved

#795


johannesjo:
Very well done! Thank you very much! 

> PS: what happened to the develop branch?

I deleted it as it currently is not needed I think. I just added it for the eslint additions because those we're multiple PRs and I knew that I wanted to publish a patch release in the meantime. Apart from this just having master and one or two feature branches works fine for the most part.


One more question: You are not accidentally keen to write some unit tests? I think the hole logic around the worklog generation is a good candidate for some unit testing love! :)

PR #888, opened 2021-02-09
feat: add project and tag column to worklog export dialog

### LINKED ISSUE #616: Project column in Worklog Export
## Problem Statement

While it is possible to export a monthly worklog for all projects from the daily view, it is not possible to distinguish which project a certain task belongs to. This makes it impossible to use the export for creating reports that distinguish by project. 

Add a "project" column to the list of available columns.

A workaround would be to export every project separately, add a column and then combine the CSVs, but with many projects this can be tedious.

stkofler:
# Description

This adds the tag and project column to worklog export dialog. 

Just like for other columns, if more than 1 per cell,  they are simply divided by a pipe and filtered to avoid duplicates. 

Please take a closer look at the bigger chunk where I use observables to fetch the corresponding tag/project name.
I hope I used them right... 
Do you think it would make sense to add a getProjectNameById and getTagNameById methods to the Project and Tag services?

I had this mostly implemented already so I wanted to publish this before addressing your question about unit testing.
I totally agree but also think it's going to be a lot of work and it might be better to refactor a bit first.
What do you think?

I think there are a lot of other important issues as well. For example:
When adding a new column, it shows up but without values. I usually hit "show as text" twice to refresh.
And obviously many more that don't concern worklog export...

Let me know what you think and thanks for your positive and elaborate response :)

#616 + the comment about tags.


johannesjo:
Thank you very much! I try to have a closer look this Friday!

[This](https://github.com/johannesjo/super-productivity/pull/885#issuecomment-775437194) probably got lost and It's definitely not a requirement, but I was wondering if I could interest you into writing some unit tests for the whole grouping logic? 😇

johannesjo:
I discovered a little bug with the implementation:

Step 1:

localhost_4200_ (1)



Step2: 

localhost_4200_ (2)


I would imagine that all the tags show up as well if "Group by Date" is selected.

stkofler:
I suppose you mean that tags: aaa, ddd, ccc are missing when grouping by date?
I think that is because they are part of task: "Today and others again" which does not have a date.
In fact this task does not show at all in step2.
And could therefore be correct?

Either way I just found another bug where: if parent task has a tag, when grouping by subtask the parent is not shown and therefore neither is its tag. That is by design but I personally think it would make more sense to show the tags of the parent task, as subtasks cannot have tags.

johannesjo:
Wow first of all sorry! I really seem to not have read your description well. So much stuff slipped my attention. It's probably because I lately spent so much time reading new issues, that at some point I am just not that attentive any more... I'll try to do better next time!

> I suppose you mean that tags: aaa, ddd, ccc are missing when grouping by date?

yes!

> When adding a new column, it shows up but without values. I usually hit "show as text" twice to refresh.
And obviously many more that don't concern worklog export...

Yes, `addCol()` should probably also retrigger the whole drawing of the table, which it does not at the moment...

src/app/features/worklog/worklog-export/worklog-export.component.ts
```diff
@@ -350,6 +361,13 @@ export class WorklogExportComponent implements OnInit, OnDestroy {
         group.notes[i] = group.notes[i].replace(/\n/g, ' - ');
       }
 
+      group.projects = unique(group.projects).filter((projectId: string) => !!projectId);
+      for (let i = 0; i < group.projects.length; i++) {
+        this._projectService.getByIdOnce$(group.projects[i]).subscribe(project => {
```
johannesjo, line 377: As mentioned in the mail: We're mixing async and synchronous code here, which will not work well. An option would be to get all the projects and all the tags in the observable on line 152 and to map them here.

johannesjo (changes requested):

src/app/features/worklog/worklog-export/worklog-export.component.ts
stkofler: that is exactly what I suspected. I just pushed another commit with your suggested changes. thank you

stkofler:
> > I suppose you mean that tags: aaa, ddd, ccc are missing when grouping by date?
> 
> yes!

But from your example I would say it is correct like this. Those tags belong to a task called "Today and others again" which does not "have" a date and should therefore not be listed in group by date? Or am I missing something?

stkofler:
> Either way I just found another bug where: if parent task has a tag, when grouping by subtask the parent is not shown and therefore neither is its tag. That is by design but I personally think it would make more sense to show the tags of the parent task, as subtasks cannot have tags.

I just pushed a commit that solves the async issue that you mentioned and confirmed my doubts, thanks for that :)
I also fixed the bug that I mentioned before about tags belonging to the parent task.

johannesjo:
> But from your example I would say it is correct like this. Those tags belong to a task called "Today and others again" which does not "have" a date and should therefore not be listed in group by date? Or am I missing something?

You're absolutely right! Makes sense as it is! I got confused by this task being listed in one view and not in the other.

src/app/features/worklog/worklog-export/worklog-export.component.ts
```diff
@@ -350,6 +384,16 @@ export class WorklogExportComponent implements OnInit, OnDestroy {
         group.notes[i] = group.notes[i].replace(/\n/g, ' - ');
       }
 
+      group.projects = unique(group.projects).filter((project: string) => !!project);
```
johannesjo, line 387: Both unique and filter should not be necessary here and cost performance. If a project is undefined than we got a data error which we want to see.

src/app/features/worklog/worklog-export/worklog-export.component.ts
```diff
@@ -350,6 +384,16 @@ export class WorklogExportComponent implements OnInit, OnDestroy {
...
+      group.projects = group.projects.map((pId: string) => {
+        return (allProjects.find(project => project.id === pId) as ProjectCopy).title;
+      });
+
+      group.tags = unique(group.tags).filter((tags: string) => !!tags);
```
johannesjo, line 392: Same as above.

johannesjo (changes requested):

src/app/features/worklog/worklog-export/worklog-export.component.ts
stkofler: I agree, but I would say there is more to that.
1. just removing those 3 lines (including the one for notes) results in errors and wrong (empty) outputs.
2. to avoid that, we'd have to:
 - use the ternary operator in several places throughout lines ~270-326 and 
 - push only if not already exists in lines ~337-341
3. to begin with, I did this to change as little code as possible and keep its structure as it is: first we create the rows and just push whatever we get from the list of tasks and _only after_ we reiterate and clean out every single row.
5. so since we are talking about refactoring we could move this to a new issue/PR?
6. another thing is that we do that for notes/tags/projects even if they're not even part of the columns to show...

If you want I can also commit changes as explained in point 2. and you chose what you like better :)
And don't get me wrong I totally agree on what you said.

johannesjo: Yes, you're right! Sorry. I slightly misread the part to be honest. What confused me was that at different points of time `group.project` and `group.tag` mean different things, which really is äh confusing! :)  But I also saw just now, that this is just consistent with the other stuff being done here and IMHO you're definitely right about prioritizing consistency over just doing it sometimes right! Let's move this to a new PR!

johannesjo (approved):

johannesjo:
Are we clear to merge this?

stkofler:
> Are we clear to merge this?

Sorry, started to write the comment but got stuck contemplating about cost efficiency. 
I changed my mind on your last comments and removed the filters.
But unique is a bit trickier as we create duplicates in lines ~337-341 while using push() to "merge" arrays. 
I think I'm at a good point but would still like to look further into this tomorrow morning with a fresh mind. :)

johannesjo:
Alright then. For me this is pretty fine as is. We can improve the rest with the refactoring. But I leave the decision to you if course :)

stkofler:
> Alright then. For me this is pretty fine as is. We can improve the rest with the refactoring. But I leave the decision to you if course :)

OK, I'd be ready to refactor now so I would say we merge this first and move on from there?

johannesjo:
Alright!

johannesjo:
Sorry for taking my time btw. I had a super busy week...

stkofler:
> Sorry for taking my time btw. I had a super busy week...
No problem, you already told me you were going to be busy.
Kind of have my hands full myself these days. 
Will let you know as soon as I will start working on refactoring and testing!

PR #891, opened 2021-02-10
feat(eslint): enable id-blacklist rule and fix errors

undrivendev:
# Description
Re-enable id-blacklist rule. Related to #803.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -63,7 +63,7 @@
             "code": 150
           }
         ],
-        "id-blacklist": "off",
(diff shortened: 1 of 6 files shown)


undrivendev:
I decided to ignore the lines that were presenting the errors because it seems to me that they are models of web API responses, so  they are not changeable.

johannesjo:
> I decided to ignore the lines that were presenting the errors because it seems to me that they are models of web API responses, so they are not changeable.

Makes a lot of sense to handle it like this! Thank you once more ! :)

PR #921, opened 2021-02-22
fix(issues): Disable caldav issue search by default

janLo:
# Description

This is necessary as there is a caldav config already for every new
project even if caldav was never set up for the project. This causes a
miconfiguration-warning on every issue-add.


johannesjo:
Oh this totally slipped past me! Thanks for the quick fix!!

johannesjo:
@janLo btw. thank you very much for taking on the discussion with the nextcloud people about the cross-origin stuff. I think their viewpoint really doesn't make a lot of sense on this.

janLo:
No problem. Its just not exactly my field of expertise, as I'm more the Infrastructure/Backend/DevOps/Automation - person ;)

PR #953, opened 2021-03-06
Link to article on 'eat the frog' in SP

menelic:
I think this helpful enough to be added here, as prioritizing is always an implicit use case for to-do and time tracking apps


johannesjo:
Thank you!

PR #962, opened 2021-03-09
fix(issues): do not rely on caldav last-changed

### LINKED ISSUE #952: CalDavSync with Nextcloud Deck does not work
### Your Environment

* Version used: 
SP 6.5.0 snap installedform candidate channel 

* Operating System and version:
Ubuntu 16.04
* Desktop Environment: 
Unity

### Expected Behavior
After entering the base dav url, the Nextcloud user name and password as well as the calendar name, creating a task with a similar name as a deck task should offer the selection of the pre-existing task

### Current Behavior
SP cannot access the Deck tasks, even if the basic connection to the calendar works (see logs below) the tasks from the calendar are not offered for auto-complete when creating a new task. There is however a never-ending spinner indicating a search: 

image

1. create new project in SP for use with Nextcloud Deck tasks
2. provide CardDav links and login data
3.  create new task in project
4. Deck tasks are not offered for use in SP

### Error Log (Desktop only)

[error] Frontend Error: {
  HANDLED_ERROR_PROP: "Caldav: TypeError: Cannot read property 'toUnixTime' of null"
} undefined

btw: adding the alternative calendar name derived from the private link as described here https://github.com/johannesjo/super-productivity/issues/626#issuecomment-782863856  does not work, it does however produce a different error:

[error] Frontend Error: {
  HANDLED_ERROR_PROP: 'Caldav: Error: CALENDAR NOT FOUND: app-generated--deck--board-43'
} undefined

So it seems in the first instance, with the plain name from the calendar list, SP finds it but cannot access the tasks for some other reason?

To ease usability here, it would be great if:
* there could be a "check connection" option when setting up the CalDav connection
* all tasks harvested from the calendar could be listed for selection (though this of course might already be the case if the connection works?)

@janLo thanks for implementing this - any hints about what I could do differently to make this work?

janLo:
# Description

It seems that last changed is not always available. As the issue model
expects a number for a change-identification attribute this simply
computes a hash over the etag value, which should change on every
content-change if the caldav service behaved correctly.

## Issues Resolved

Resolves #952

### DIFF
--- a/src/app/features/issue/providers/caldav/caldav-client.service.ts
+++ b/src/app/features/issue/providers/caldav/caldav-client.service.ts
@@ -127,10 +127,26 @@ export class CaldavClientService {
       summary: todo.getFirstPropertyValue('summary') || '',
       due: todo.getFirstPropertyValue('due') || '',
       start: todo.getFirstPropertyValue('dtstart') || '',
-      last_modified: todo.getFirstPropertyValue('last-modified').toUnixTime(),
       labels: categories,
-      note: todo.getFirstPropertyValue('description') || ''
+      note: todo.getFirstPropertyValue('description') || '',
+      etag_hash: this._hashEtag(task.etag),
     };
+  }
+
+  private static _hashEtag(etag: string): number {
+    let hash = 0;
+    let i;
+    let chr;
+    if (etag.length === 0) {
+      return hash;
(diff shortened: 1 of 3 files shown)


johannesjo:
Thank you very much once more!

PR #975, opened 2021-03-13
[fix] Use volume setting for pomodoro tick sound.

### LINKED ISSUE #956: Sound volume settings should affect Pomodoro clicking sound as well
## Problem Statement

There is a "Play tick sound every second" setting in "Pomodoro" section. Currently if I turn it on, the sound is too loud. But if I adjust volume in "Sound" settings, the volume of this every-second tick stays the same.

This volume setting should also affect this clicking sound.

The clicking sound may also be customized separately. That way it's possible to provide more flexibility, add options for selecting different sound etc. Not sure if it's better to keep the setting around option to enable clicking sound or to put it in "Sound" settings.

- Related discussion: https://github.com/johannesjo/super-productivity/discussions/949#discussioncomment-436065
- May be a good issue to tackle for programmers being new to the project.
- I'd start looking from here: https://github.com/johannesjo/super-productivity/blob/master/src/app/features/config/global-config.model.ts#L93 - looks like this is a store parameter that sets the volume.

surli:
# Description

Use SoundConfiguration#volume to set the tick volume of pomodoro, getthe code idea from commented code in play-done-sound.

## Issues Resolved

Fixes #956 

## Check List

### DIFF
--- a/src/app/features/pomodoro/pomodoro.service.ts
+++ b/src/app/features/pomodoro/pomodoro.service.ts
@@ -147,7 +147,7 @@ export class PomodoroService {
       filter(([val, cfg]) => cfg.isEnabled && cfg.isPlayTick),
       map(([val]) => val),
       distinctUntilChanged(),
-      withLatestFrom(this.soundConfig$)
+      withLatestFrom(this.soundConfig$),
     ).subscribe(([val, soundConfig]) => {
       this._playTickSound(soundConfig.volume);
     });
--- a/src/app/features/pomodoro/pomodoro.service.ts
+++ b/src/app/features/pomodoro/pomodoro.service.ts
@@ -11,7 +11,7 @@ import {
   switchMap,
   withLatestFrom
 } from 'rxjs/operators';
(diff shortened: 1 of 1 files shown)


surli:
Hi, this is my contrib on this project, to see a bit how the code looks like and see if I can help. Not used to code in typescript so let me know if there's conventions I didn't follow or some test I should have added for this.

src/app/features/pomodoro/pomodoro.service.ts
```diff
@@ -146,8 +147,9 @@ export class PomodoroService {
       filter(([val, cfg]) => cfg.isEnabled && cfg.isPlayTick),
       map(([val]) => val),
       distinctUntilChanged(),
-    ).subscribe(() => {
-      this._playTickSound();
+      withLatestFrom(this.soundConfig$)
```
johannesjo, line 150:
```suggestion
      withLatestFrom(this.soundConfig$),
```

johannesjo (commented):
Thank you very much! Excellent job! Just the one minor improvement suggestion.

PR #994, opened 2021-03-22
feat: add edit tags shortcut #808

### LINKED ISSUE #808: Keyboard shortcut for editing / adding new tags
Would it be possible to implement a keyboard shortcut for interaction with tags?

rmpsc:
# Description

Added a feature to edit tags when a specific task is in focus.

## Issues Resolved

Resolves #808


src/app/features/config/default-global-config.const.ts
```diff
@@ -99,6 +99,7 @@ export const DEFAULT_GLOBAL_CONFIG: GlobalConfigState = {
     expandSubTasks: null,
     collapseSubTasks: null,
     togglePlay: 'y',
+    taskEditTags: 'z',
```
johannesjo, line 102:
```suggestion
    taskEditTags: 'g',
```
I prefer 'g' because of "taG" and because z is very far away for the left hand on a QWERTZ keyboard.

src/app/features/tasks/task/task.component.ts
```diff
@@ -331,6 +331,14 @@ export class TaskComponent implements OnInit, OnDestroy, AfterViewInit {
...
     }
     // this.focusSelf();
+    this._matDialog.open(DialogEditTagsForTaskComponent, {
+      data: {
+        task: this.task
+      }
+    })
+      .afterClosed()
+      .pipe(takeUntil(this._destroy$))
+      .subscribe(() => this.focusSelf());
```
johannesjo, line 341:
```suggestion
```

johannesjo:
Thank you very much! Could you please review the requested changes?

johannesjo:
Also the shortcut does not seem to work for some reason.

rmpsc:
The shortcut seems to work in an incognito tab, but will not set "g" to the default key when loading it onto an existing board. Did new shortcut additions in the past behave like so?

johannesjo:
Hey there! My bad. I am using vimium for my browser which was messing with my keyboard shortcuts. Works fine otherwise. Apart from the g key which might be taken by something else. It definitely does work in Firefox and should also work for electron.

Not sure if I understand your question correctly. The default configuration is only used in the settings if the value was never changed before. Does this answer your question?

Could you maybe have  a look at the linting issues?

rmpsc:
Unsure of how to resolve the linting issues. I believe the checks were successful before the prior changes, but I don't think those changes could've caused the checks to fail.

johannesjo:
> Unsure of how to resolve the linting issues. I believe the checks were successful before the prior changes, but I don't think those changes could've caused the checks to fail.

Ah alright. Seems like there was a problem when installing the dendencies. Just re-triggered the build. Sorry!

johannesjo:
@rmpsc this is ready to merge now, right?

rmpsc:
@johannesjo Yes, it's all ready to go :)

johannesjo:
Thank you very much again! :)

PR #1033, opened 2021-04-12
Create dependabot.yml

cwerner1:
# Description

Automated Dependecy management by dependabot.
dependabot will automatcly create a new Pull Request, when an dependency is updated. Also the changes are tested with the github actions.

## Issues Resolved

-

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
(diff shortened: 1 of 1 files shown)


johannesjo:
Thanks for creating this! I'd imagine that this might be quite annoying :D It also wonder how it's supposed to work with breaking changes. But let's give it a shot.

cwerner1:
If it get too annoying, you could set the timer to monthly, to reduce the noise. 

The code snippet is:
`update_schedule:` "monthly"`

greetings from austria

cwerner1:
@johannesjo 
I also would suggest, that you could select "rebase and merge" at the pull request, from Dependabot, to reduce the branches in the git commit history
grafik

johannesjo:
Ahh. Thank you very much for the advice. I am already pretty annoyed, but the feeling of being more and more up to date is also kind of addictive :D I probably should make a single commit updating everything once.

PR #1045, opened 2021-04-13
Refactor worklog-export: unnest functions and bundle params

stkofler:
# Description

This extracts the nested functions and adapts the arguments and test cases.
I'm not really a fan of the bundling of arguments but I think for now it should be fine and there are definitely more important things to address...

You also asked about making the switch/case statement a bit clearer. For that I would propose to add functions for each case like: handleDateGroup, etc. And maybe even take it out of the loop as we don't need to check for the group for each single task.
There are also a few really unclear parts of the code which I could address in a new GitHub issue? By that I mean things like in lines: 28-35 and 192-202.

As always let me know what you think :)

### DIFF
--- a/src/app/features/worklog/worklog-export/worklog-export.util.ts
+++ b/src/app/features/worklog/worklog-export/worklog-export.util.ts
@@ -14,32 +14,18 @@ const LINE_SEPARATOR = '\n';
 const EMPTY_VAL = ' - ';
 
 export const createRows = (data: WorklogExportData, groupBy: WorklogGrouping): RowItem[] => {
-  const groups: { [key: string]: RowItem } = {};
+  let groups: { [key: string]: RowItem } = {};
 
-  data.tasks.forEach((task: WorklogTask) => {
-    const taskGroups = mapToGroups(task, groupBy, data);
-    Object.keys(taskGroups).forEach(groupKey => {
-      if (groups[groupKey]) {
-        groups[groupKey].tasks.push(...taskGroups[groupKey].tasks);
-        groups[groupKey].notes.push(...taskGroups[groupKey].notes);
-        groups[groupKey].projects.push(...taskGroups[groupKey].projects);
-        groups[groupKey].tags = groups[groupKey].tags.concat(taskGroups[groupKey].tags);
-        groups[groupKey].dates.push(...taskGroups[groupKey].dates);
-        if (taskGroups[groupKey].workStart !== undefined) {
-          // TODO check if this works as intended
-          groups[groupKey].workStart = Math.min(groups[groupKey].workStart as number, taskGroups[groupKey].workStart as number);
-        }
-        if (taskGroups[groupKey].workEnd !== undefined) {
-          // TODO check if this works as intended
-          groups[groupKey].workEnd = Math.min(groups[groupKey].workEnd as number, taskGroups[groupKey].workEnd as number);
-        }
-        groups[groupKey].timeEstimate += taskGroups[groupKey].timeEstimate;
-        groups[groupKey].timeSpent += taskGroups[groupKey].timeSpent;
-      } else {
-        groups[groupKey] = taskGroups[groupKey];
-      }
-    });
-  });
+  switch (groupBy) {
+    case WorklogGrouping.DATE:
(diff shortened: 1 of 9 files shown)


src/app/features/worklog/worklog-export/worklog-export.util.ts
```diff
@@ -152,40 +49,130 @@ export const createRows = (tasks: WorklogTask[], startTimes: WorkStartEnd, endTi
...
+      return (data.tags.find(tag => tag.id === tId) as TagCopy).title;
     });
 
     rows.push(group);
   }));
 
   return rows;
+};
+
+const mapToGroups = (task: WorklogTask, groupBy: WorklogGrouping, data: WorklogExportData) => {
```
johannesjo, line 78: Return type seems to be missing

src/app/features/worklog/worklog-export/worklog-export.util.ts
```diff
@@ -152,40 +49,130 @@ export const createRows = (tasks: WorklogTask[], startTimes: WorkStartEnd, endTi
...
-      return (allTags.find(tag => tag.id === tId) as TagCopy).title;
+      return (data.tags.find(tag => tag.id === tId) as TagCopy).title;
     });
 
     rows.push(group);
   }));
 
   return rows;
+};
+
```
johannesjo, line 77: I would love to see a short introductory comment on what this does and how, because I am pretty sure I will quickly forget this :)

src/app/features/worklog/worklog-export/worklog-export.util.ts
```diff
@@ -4,122 +4,20 @@ import { msToString } from '../../../ui/duration/ms-to-string.pipe';
...
-          taskGroups[groupKey].workStart = startTimes[day];
-          taskGroups[groupKey].workEnd = endTimes[day];
-          taskGroups[groupKey].timeEstimate = task.subTaskIds.length > 0 ? 0 : task.timeEstimate;
-          taskGroups[groupKey].timeSpent = task.subTaskIds.length > 0 ? 0 : task.timeSpentOnDay[day];
-        });
-    }
-    return taskGroups;
-  };
-
+export const createRows = (data: WorklogExportData, groupBy: WorklogGrouping): RowItem[] => {
```
johannesjo, line 16: I would love to see a short introductory comment on what this does and how, because I am pretty sure I will quickly forget this :)

src/app/features/worklog/worklog-export/worklog-export.util.ts
```diff
@@ -152,40 +49,130 @@ export const createRows = (tasks: WorklogTask[], startTimes: WorkStartEnd, endTi
...
+    if (dateA === dateB) {
+      return 0;
+    } else if (dateA < dateB) {
+      return -1;
+    }
+    return 1;
+  });
 };
 
 export const formatRows = (rows: RowItem[], options: WorklogExportSettingsCopy): (string | number | undefined)[][] => {
```
johannesjo, line 178: I would love to see a short introductory comment on what this does and how, because I am pretty sure I will quickly forget this :)

johannesjo:
Hey there! Thank you very much once more! Looks very clean and much better! What I personally would find super helpful would be some short introductory comment for each of the three exported util functions. I am sure my future self would be very thankful, if it were to revisit this at some point :)

src/app/features/worklog/worklog-export/worklog-export.util.ts
johannesjo: And I would love to see a short introductory comment on what this does and how, because I am pretty sure I will quickly forget this :)

stkofler:
Haha, oh would I have loved some comments there in the first place :)
Alright, alright I'll give you that, but under the condition that I will be pushing big changes to this PR, so you'll have a lot to read through :P

johannesjo:
> Alright, alright I'll give you that, but under the condition that I will be pushing big changes to this PR, so you'll have a lot to read through :P

Hehe. No problem :)

stkofler:
oh wow, where are all those commits coming from, should have been rebased. Not sure what is going on. If you want I can open a new PR.

PR #1046, opened 2021-04-13
Keep text when focus lost while adding todo

### LINKED ISSUE #635: On input losing focus, any entered new task data is gone
**Note:** This is more a usability annoyance than a bug, not super important!

### Your Environment
* Version used: 6.0.1, .exe (non-installer) version
* Operating System and version: Windows 10 (2004)

1. Click + ("Add new Task")
2. Enter some task description; don't submit it yet
3. Change to another window, or wait until some break/task start alert comes along
4. The task adding line is gone, along with the task description entered so far

### Current Behavior
As stated above, the entered task description is gone; when pressing + again, the task description field is empty.

### Expected Behavior
Alerts and switching to another window should (1) either not interrupt the task editing process, or (2) the entered task description should be preserved (i.e. so that after the alert has closed, and I press "+", the previously entered task description should be pre-filled with what was in there before).

### Console Output
No relevant output.

### Error Log (Desktop only)
No log file created.

codeling:
# Description

When a new todo is entered, and the input loses focus for some reason, for example because (1) the user clicks anywhere else in the superproductivity window (except close button or the + button left of the input) or on another window, or (2) another window pushes itself into focus (e.g. a reminder), previously the entered text was lost.
With this pull request, superproductivity keeps a copy of the entered text; and on the next time that the add todo box is opened, the text input will be pre-filled with the previously entered text (which is also selected to make it easier for the user to just ignore the pre-filled text by starting writing).

I would love any feedback on the changes I did, since I'm quite new to angular and typescript!
Especially on:
- is `this.taskSuggestionsCtrl.value` the correct way to retrieve the current input element value?
- is the taskService the right place for storing the temporary text?
- is there a better way to set the input text apart from using `...nativeElement.value` along with `taskSuggestionControl.setValue` (the former, in my tests, was required for making the text appear in the user-visible input; the latter was required for setting the internal state so that the user could directly "add" a task with the pre-filled value on opening the add bar).

## Issues Resolved

Resolves #635

## Check List

- [ ] New functionality includes testing -> please advise where tests are located and if any are required for my changes
- [ ] New functionality has been documented in the README if applicable -> I suppose this small change does not warrant mentioning in the README


johannesjo:
Thank you very much for opening this up! I will try to have a look tomorrow or this Friday.

src/app/features/tasks/add-task-bar/add-task-bar.component.ts
```diff
@@ -149,6 +149,17 @@ export class AddTaskBarComponent implements AfterViewInit, OnDestroy {
         }
       });
     });
+
+    const savedTodo = this._taskService.getSavedTodo();
+    if (savedTodo)
+    {
+      this._taskService.setSavedTodo('');
+      this.taskSuggestionsCtrl.setValue(savedTodo);
+      setTimeout(() => {
```
johannesjo, line 158: We probably want to store this away and clear it in `ngOnDestroy` like the other timeouts like `this._blurTimeout`. This is needed to avoid this code is being executed, after the component was destroyed, which would lead to an error.

src/app/features/tasks/add-task-bar/add-task-bar.component.ts
```diff
@@ -149,6 +149,17 @@ export class AddTaskBarComponent implements AfterViewInit, OnDestroy {
         }
       });
     });
+
+    const savedTodo = this._taskService.getSavedTodo();
+    if (savedTodo)
+    {
```
johannesjo, line 155:
```suggestion
    if (savedTodo) {
```

src/app/features/tasks/add-task-bar/add-task-bar.component.ts
```diff
@@ -177,6 +188,12 @@ export class AddTaskBarComponent implements AfterViewInit, OnDestroy {
 
   onBlur(ev: FocusEvent) {
     const relatedTarget: HTMLElement = ev.relatedTarget as HTMLElement;
+    if (!relatedTarget || (relatedTarget &&
+      !relatedTarget.className.includes('close-btn') &&
+      !relatedTarget.className.includes('switch-add-to-btn')))
+    {
```
johannesjo, line 194:
```suggestion
      !relatedTarget.className.includes('switch-add-to-btn'))) {
```

src/app/features/tasks/add-task-bar/add-task-bar.component.ts
```diff
@@ -177,6 +188,12 @@ export class AddTaskBarComponent implements AfterViewInit, OnDestroy {
...
+      this._taskService.setSavedTodo(this.taskSuggestionsCtrl.value);
```
johannesjo, line 195: I'd probably use the sessionStorage instead. The task service is already pretty big and more focused on actual model stuff. 

If you change this, it would be cool, if you could put the key for that into this file:
`src/app/core/persistence/ls-keys.const.ts`

johannesjo:
This works very nicely and adds a lot of value, too. 

I am honestly not too familiar with the whole FormControl stuff neither and it's the only place in the app where it's being used. I would also have expected for `setValue` to do the trick, as it is documented that way, but this might also be due how to the material autocomplete component works and that it is wired up a bit differently. I tried to read a little bit on the subject, but wasn't able to find a cleaner solution.

Some other minor thing to note: Normally the project uses the [angular commit guidelines](https://gist.github.com/brianclements/841ea7bffdb01346392c), so it would be cool if you could stick to that too for future commits.

codeling:
Thanks for your feedback! I think I have now addressed all the concerns with my latest commits. If they look OK, then I would rebase and squash the changes to a single commit, to address

>  Normally the project uses the angular commit guidelines

Just a quick question on that - what would be the proper "scope" value for these changes?

johannesjo:
> If they look OK, then I would rebase and squash the changes to a single commit, to address the angular commit guidelines point - one quick question on that though - what would be the proper "scope" value for these changes?

You can use "task" or leave it blank. The scope value isn't to important and I never thought up a strict rule-set for it. It's main purpose is grouping commits for the the automatic changelog generation.

johannesjo:
Looks all good to me! We can merge it as soon as you think it's good to go :)

codeling:
:+1: 

All done now from my side!

johannesjo:
Thank you very much again!

PR #1190, opened 2021-05-07
Update nl.json

kerkenit:
# Description

Change weird Dutch tekst (from auto translate)
achterstand look like slum. It's a negative word. Backlog is more common because we don't have a good word in Dutch

## Issues Resolved

None

## Check List


johannesjo:
Thank you very much once more!

Issue #1197, opened 2021-05-09
WebDAV service expects putFileContents to return AxiosReponse, but it returns Boolean


wilbowma:
There's a type error in this code. (Not sure why Type Script can't catch it :P).

In short, this line:

https://github.com/johannesjo/super-productivity/blob/3f321c18fcbc16267bbc23c3b7dcdee37daa2952/src/app/imex/sync/web-dav/web-dav-api.service.ts#L43

Claims to be returning an AxiosResponse, but its return value comes from putFileContents in the webdav library

https://github.com/johannesjo/super-productivity/blob/3f321c18fcbc16267bbc23c3b7dcdee37daa2952/src/app/imex/sync/web-dav/web-dav-api.service.ts#L51
https://github.com/johannesjo/super-productivity/blob/3f321c18fcbc16267bbc23c3b7dcdee37daa2952/src/app/imex/sync/web-dav/web-dav-api.service.ts#L53

which returns a Boolean:

https://github.com/perry-mitchell/webdav-client/blob/4248cd6c0c797dce4d95295f2bdc2380c289dcb4/source/operations/putFileContents.ts#L18

This causes some issues here, when the boolean is used as a response:
https://github.com/johannesjo/super-productivity/blob/3f321c18fcbc16267bbc23c3b7dcdee37daa2952/src/app/imex/sync/web-dav/web-dav-sync.service.ts#L98

wilbowma:
Even if the response was returned, the header from the put cannot be expected to contain an etag:
http://www.webdav.org/specs/rfc4918.html#etag

Instead, after the put request, the sync service should probably follow up with a HEAD request to get the new etag if the put succeeded.

johannesjo:
I think this is fixed now.

Issue #1150, opened 2021-04-30
DAV sync error: TypeError: i.headers is undefined


nathan-sain:
### Your Environment

* Version used: Self hosted, Docker Container, 6.5.2
* Operating System and version: Ubuntu 18.04
* Desktop Environment: Ubuntu Gnome
* Browser Name and version: Firefox 88.0 64-bit

### Expected Behavior

Sync should happen without reporting error

### Current Behavior

An error is displayed, but the sync occurs. Sometimes sync is not logged as successful and a modal prompt to select data source is displayed on next sync.
The error displayed is "Unknown Upload Error (Settings Correct?). TypeError: i.headers is undefined"

### Steps to Reproduce (for bugs)
1. Setup WebDav Sync
2. See if error occurs

### Console Output
WebDAV ↑ Update Remote sync-provider.service.ts:366:19

true web-dav-api.service.ts:47:12

true web-dav-sync.service.ts:89:14

TypeError: i.headers is undefined
    uploadAppData web-dav-sync.service.ts:91
    a main.cc529609e1915c525b4d.js:1
    Angular 13
        invoke
        onInvoke
        invoke
        run
        P
        invokeTask
        onInvokeTask
        invokeTask
        runTask
        m
        invokeTask
        _
        k
web-dav-sync.service.ts:93:14
    uploadAppData web-dav-sync.service.ts:93
    a main.cc529609e1915c525b4d.js:1
    Angular 13
        invoke
        onInvoke
        invoke
        run
        P
        invokeTask
        onInvokeTask
        invokeTask
        runTask
        m
        invokeTask
        _
        k

WebDAV X Upload Request Error sync-provider.service.ts:366:19

sync(effect)..... I_FOCUS_THROTTLED sync.effects.ts:109:23

Object { filename: "/home-super.json", basename: "home-super.json", lastmod: "Fri, 30 Apr 2021 06:32:33 GMT", size: 12463, type: "file", etag: "&quot;ee834b31f78208c9610e71fefc35311e&quot;", mime: "application/json" }
​basename: "home-super.json"
​etag: "&quot;ee834b31f78208c9610e71fefc35311e&quot;"
​filename: "/home-super.json"
​lastmod: "Fri, 30 Apr 2021 06:32:33 GMT"
​mime: "application/json"
​size: 12463
​type: "file"
​<prototype>: Object { … }
web-dav-api.service.ts:67:12

[...]

PR #1200, opened 2021-05-10
Fix handling of WebDAV headers

### LINKED ISSUE #1199: WebDAV sync always claims a remote conflict with some WebDAV servers (nginx)
### Your Environment

* Version used: 3f321c18fcbc16267bbc23c3b7dcdee37daa2952
* Operating System and version: Linux 5.11.16-2-MANJARO #1 SMP PREEMPT Thu Apr 22 09:53:11 UTC 2021 x86_64 GNU/Linux

* Desktop Environment:  KDE

### Expected Behavior
WebDAV sync should work with WebDAV-compliant servers.

### Current Behavior
WebDAV sync fails with nginx, as the conflict resolution relies on the `etag` header being returned from `PROPFIND` (through getStat in the webdav client), but this header is not returned from `PROPFIND` at least on nginx. (This seems to be consistent with the WebDAV spec http://www.webdav.org/specs/rfc4918.html#rfc.section.9.1.3)

1. Setup sync with a nginx webdav server
2. Wait
3. On any upload after an initial upload, the user is warned that the remote has been modified, although it hasn't been, and asked to resolve the conflict. The last sync date will by the 0 unix timestamp.

### Possible fix
I think one fix is to issue a `HEAD` request instead, which is both faster and should return a header with `etag`. I'll try to add this support to the webdav-client and test a fix.

wilbowma:
# Description

This fixes issues with the handling of WebDAV headers:
- Don't attempt to set Content-Length manually (as browsers forbid this and set it automatically; Axios does the right thing, apparently)
- Correct handling of ETAG header, which is not always present in PUT or PROPFIND requests.

## Issues Resolved

 #1199 and #1197

## Check List

- [X] New functionality has been documented in the README if applicable.

### DIFF
--- a/src/app/imex/sync/web-dav/web-dav-api.service.ts
+++ b/src/app/imex/sync/web-dav/web-dav-api.service.ts
@@ -48,8 +48,10 @@ export class WebDavApiService {
       password: cfg.password,
     });
 
-    const r = await client.putFileContents('/' + cfg.syncFilePath, JSON.stringify(data));
-    console.log(r);
+    const r = await client.putFileContents('/' + cfg.syncFilePath, JSON.stringify(data), {
+      contentLength: false,
+      //, overwrite: isForceOverwrite
+    });
     return r;
   }
 
--- a/src/app/imex/sync/web-dav/web-dav-api.service.ts
+++ b/src/app/imex/sync/web-dav/web-dav-api.service.ts
@@ -82,7 +82,7 @@ export class WebDavApiService {
     });
     const r = await client.customRequest(path, { method: 'HEAD' });
     console.log(r);
-    return r;
+    return r.headers;
   }
 
   async download({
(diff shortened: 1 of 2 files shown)


wilbowma:
#1150 may also be related

johannesjo:
Thank you very much! I just checked with the most recent version of the webdav package which includes typing for the web client and there seem to be other inconsistencies as well. Would you mind having a look at them, too?

wilbowma:
Ah, I was just playing with build options and was getting new errors... that must be why. I'll take a closer look tomorrow.

johannesjo:
Thank you! Please, let me know if I can assist you in any way!

johannesjo:
@wilbowma In case your busy, I can also have a look. Would be nice to merge to merge your fix for the next release, which planned for next Monday. :)

wilbowma:
Go ahead; I haven’t started as the CORS issue ate up a bunch of my time.

--

PR #1206, opened 2021-05-12
Space optimize distribution

wilbowma:
# Description

This PR includes a series of optimizations for reducing the distribution size. I'm not entirely sure all of these are acceptable defaults, so I'd appreciate feedback.

- Removes source maps from production.
- Add new "modern" browserlist target for distributions that can rely on ES6 and WOFF2.
- Size optimize all images using `ect -9 -strip` (https://github.com/fhanau/Efficient-Compression-Tool)
- Make logging and debugging related node packages dev-only
- Make Google APIs an optional dependency.

The interface is a new script, `yarn buildFrontend:prod:es6` (optionally after running `yarn --ignore-optional` if you don't need Google APIs).

The result is a substantially smaller distribution: between 48% and 83.6% smaller.

Before:
```
> du -sh dist/
48M     dist/
> du -sh app-builds/linux-unpacked/resources/app.asar
50M     app-builds/linux-unpacked/resources/app.asar
```

After (no Google APIs)
```
> du -sh dist/
8.2M     dist/
> du -sh app-builds/linux-unpacked/resources/app.asar
8.5M    app-builds/linux-unpacked/resources/app.asar
```

After (optional dependencies):
```
> du -sh dist/
8.2M     dist/
> du -sh app-builds/linux-unpacked/resources/app.asar
26M     app-builds/linux-unpacked/resources/app.asar
```

Currently, the new script requires `find`, so it probably doesn't work on Windows.

We can remove another .7M by removing unnecessary locales, but I've been unable to find a good way to script that.

## Check List

- [ ] Replace `find` with cross-platform equivalent.


johannesjo:
Thank you very much for this! This is really great and much appreciated! 

Makes all sense to me apart from the source maps. Even though the sourcemap files are really huge (compared to the rest of the files), this is a trade-off I am willing to make in order to make working on bug tickets much easier. It might be possible to find a smarter solution for this, but until then I rather keep them in the default built.

johannesjo (changes requested):
Unfortunately we need the sourcemaps...

wilbowma:
Hm, that seems against standard practice, and I looked through a few bug reports and only found 1 the made use of the source maps. Maybe it would be possible to maintain a debug distribution on GitHub that include source maps?

johannesjo:
A debug distribution requires a lot of extra effort from the user (and also from me as it would need to be setup and maintained too) and they might not even be able to properly reproduce the issue. Although it doesn't work all of the time (haven't been able to quite put my finger on it, but it maybe due to it loading requiring some more time on some systems), it's still saved me many hours. İ remember having to spent a lot of time on communication about certain issues without being able to pinpoint what's wrong, which could have been solved in minutes with the knowledge of the exact line that threw the error. İt's not ideal. Better logging might be an alternative, but (if not done remotely, which is not an option here) this has an impact on performance (e.g. the RAM being used), while the source maps only affect disk space by some MB.

wilbowma:
Fair enough; I'll re-enable them

wilbowma:
OOf. source maps by triple the size of the app's code:
```
> du -sh dist/
25M     dist/
```

I know there exist services that can somehow take source maps + bug reports and get the source line numbers:

johannesjo (approved):

johannesjo:
The problem with bugsnag (or even a completely custom solution) is that significant parts of the user base are very sensitive when it comes to their data and part of the apps appeal is that it doesn't transmit any user data in the background. 

I know that there a good ways to do this, that even very sensitive people would not complain, but people have a limited attention span and usually don't care enough to look into the details.

This is definitely an interesting subject! Without having dived too deep into the subject, I'd assume that the best alternative is some sort of custom logging, that limit's itself to maybe a couple of MB and doesn't use console.log in production.   

Another smart alternative could be some sort of tool that makes remapping the minified code to the sourcemaps (which could be build anyway, but stay on some server) much easier. This could be an interesting github project in its own right. :)

Issue #547, opened 2020-09-29
Search Feature?


aaylnx:
I'm new to SP. I'm sorry if this is the wrong place to ask a question like this, but is there an ability to search within SP? If so, I am unable to locate it.  Thanks!

johannesjo:
Thanks for opening this up. No, currently there is no such thing. Interestingly (to me at least ;)) I very rarely need this, but it probably makes sense to do this in one form or another.

aaylnx:
Sure. I understand it's somewhat use case dependent. It's just a feature I'm used to from other task managers I've used. I'm sure you have a long list of feature requests, but if this one makes it on the list somehow, I'm betting it will prove useful for a number of users. In any event, thanks for all your work!

tustinn:
+1 :-)

privacy-advo:
First **Thank you !** for the amazing app.

I strongly support this request.

IT-Tool:
+1

IT-Tool:
+1

- It is easier to find the required task through search if there are many projects and tasks.
- It is also more convenient to find similar tasks through the search between all projects (if needed).
- And so on

guitarzero:
Me too. I would love a "search all tasks" features. Even better, in the same way as Notational Velocity does, that is to say you start typing in a search box, and all tasks are being filtered out in real-time.

aaylnx:
> Me too. I would love a "search all tasks" features. Even better, in the same way as Notational Velocity does, that is to say you start typing in a search box, and all tasks are being filtered out in real-time.

Indeed, that would be nice.

Definitely hoping that some search feature makes it in in the future.

Rielyr:
Hello and thank you for your awesome work, i just found out about this app and looks very promising. I would definitely agree that a search bar and other filtering capabillities are a must for any medium to large project. Right now i'm just testing to see if i can make this my default productivity app, but i wouldn't migrate my hole backlog without this feature.

aaylnx:
Since I made this request, it appears that tasks do show up if you start typing in the new task window, but only if those tasks have been put in a project. As a consequence, I tend to have tasks placed in projects now.

privacy-advo:
> Since that time, it now appears that tasks do show up if you start typing in the new task window, but only if those tasks have been put in a project. 

Could you please explain in which section/window you are and which behaviour leads to a search. Thanks in advance

Already existing behaviour I'm aware of:
[...]

PR #1217, opened 2021-05-17
feat(search): add search bar #547

stkofler:
# Description

This provides a starting point for a search feature based on ideas in issue #547.
How it works:

1. Shortcut _Shift+F_ and _icon_ in main header next to "Add new task"
2. Opens an input field as an overlay just like "Add task bar"
3. Lists **all** tasks from everywhere right away
4. Upon user input: filters by searching in _Task Title_ and _Task Description_
5. Upon selection: navigates to the page and focuses the selected task

Considerations:

1. Actual _Notes_ are currently not considered. I think it'd be best to discuss this in a new issue as it raises many questions (should we add them to the same list as tasks, then how would we sort them all? should we give it it's own shortuct? Consider that Notes don't have a title property... )
2. Listed tasks are currently not sorted specifically.
3. I adapted the _blendInOutAnimation_ slightly as I wanted to use the _onAnimationEvent_ callback in the _search-bar.component_.
4. When the window gets smaller than 600px wide and the UI slightly changes, should we add a search button just next to the new add button in the bottom right corner?
5. I'm aware that quite some code is copied from add-task-bar html and css. I felt it'd be better to refactor once this approach gets accepted. 
6. The shortcut might have to be set manually after an update.
7. I didn't implement any tests as I wanted to get some feedback first.

So yeah, looking forward to hearing your ideas! :)


johannesjo:
Hey there! Works pretty nicely! I like it! 

A couple of notes:

1. I'd imagine that most people would use this to search for archived tasks, which are not easy to locate, so think it definitely makes sense to add those too, though this might be tricky to implement. It's also not obvious to me how to best deal with that case. Just some ideas: A) There could be an extra modal displaying the task contents, B) a dialog that asks if the user wants to restore the task, C) Move the tasks location in the archive. What do you think?
2. Don't worry about the notes, I wanted to revamp the feature anyway, so it's good to leave them out for now :)
3. I would put the add task button first and only then the search button, as adding tasks should be a more common action than searching them.
4. I am alright with not sorting the tasks. I think it works nicely as is.
5. For mobile I would show the button at the top as it is for desktop. From the viewpoint of the material design guidelines there only should ever be one fab button.
6. Don't worry about the code duplication. In some cases it is favorable especially if the number of instances is small and/or the abstraction is not obvious. 
7. For the shortcut we can simply raise the  model version of the global config by 0.1. Then the new shortcut should be copied over from the default.

stkofler:
Cool, thanks for the feedback.
I fixed the icon issues and bumped the config version number from 2.2002 to 2.2003. I hope that is what you meant.

As for archived tasks:
I'd really be interested in hearing from other people how they use it. I have not found too many issues/discussions about them. I myself use the app mostly to plan ahead and never needed to search anything that got archived but I totally agree that, as it is now, finding an archived task is rather difficult and should be improved.
I believe your idea A) would be the best and most logical option. 
This also raises the question on whether to include archived tasks in the same list or toggle them with a button like in add-task-bar and the "add to backlog" button.
Do you remember if there is already a selector or even observable that lists _only_ archived tasks in one call? I only found getAllTasksForProject(projectId) but that returns all kinds of tasks and would require additional iteration on projects. I could implement a similar function...

johannesjo:
> I fixed the icon issues and bumped the config version number from 2.2002 to 2.2003. I hope that is what you meant.
Yes!

I agree that A looks like the best and most consistent option. There is currently is no function to get all tasks, but it should be straightforward to implement: 
```TypeScript
  async getAllTasksIncludingArchive(): Promise<Task[]> {
    const allTasks = await this._allTasksWithSubTaskData$.pipe(first()).toPromise();
    const archiveTaskState: TaskArchive = await this._persistenceService.taskArchive.loadState();
    const ids = (archiveTaskState && (archiveTaskState.ids as string[])) || [];
    const archiveTasks = ids.map((id) => archiveTaskState.entities[id]);
    return [...allTasks, ...archiveTasks];
  }
```

stkofler:
Hey, thanks for your input and quick update. 
Switching between the two task sources works pretty well now. Do you by any chance have a huge json lying around to seed the app for manual testing?  Creating archived tasks by hand is rather troublesome...
Now I start working on the navigation, which is also going to be interesting :)

stkofler:
Ok, switching to archived tasks should now work as well by clicking the same button as in "add new task" or by "Ctrl+1" shortcut.

Performance:
I tested it thoroughly and on my PC it takes no more than 80ms to load all 6k+  tasks and run the filter.
What does take long however, is the creation of 6k+ mat-autocomplete options... :)
So I limit the filter to output no more than 200 and show a little message to narrow the search. Should be fine so far.

Weird little bug:
Only in archived tasks when selecting an option from the search list with a mouse click, the item will not be focused after navigating to it. To be more precise: the item will be focused, because it actually scrolls to it, but it doesn't get the colored border.
This works perfectly in non-archived tasks and when selecting an item by hitting enter. Plus, when you navigate by mouse click and the searched item shows up, the border will show up by hitting e.g. the shift key on the keyboard... really weird. Do you have any ideas? I think we could even leave it like that for now...

Migration:
I'm not sure I got that part right. The last commit adds the migration and it actually works. However I noticed the migration to run for several minutes on the 6k+ tasks and I didn't quite understand if it ran again after subsequent page loads. Maybe I missed something?

src/app/features/search-bar/search-bar.component.ts
```diff
@@ -0,0 +1,250 @@
...
+    return {
+      ...context,
+      icon: (context as Tag).icon || (task.projectId && 'list'),
+    };
+  }
+
+  ngAfterViewInit(): void {
+    this._setUpFilter();
+
+    this._attachKeyDownHandlerTimeout = window.setTimeout(() => {
```
johannesjo, line 98: Love that you thought about canceling the timeout :)

src/app/features/search-bar/search-bar.component.ts
```diff
@@ -0,0 +1,250 @@
...
+    private _router: Router,
+  ) {
+    this._taskService.allTasks$.pipe(take(1)).subscribe((t) => (this._tasks = t));
+    this._projectService.list$.pipe(take(1)).subscribe((p) => (this._projects = p));
+    this._tagService.tags$.pipe(take(1)).subscribe((t) => (this._tags = t));
+
+    this._mapTasksToSearchItems();
+  }
+
+  private _mapTasksToSearchItems(): void {
```
johannesjo, line 69: Just as a note: When you're leaning to the functional side of programming (like I do ;)) then it's considered cleaner to use params and return values rather than having a function that reads and assigns class properties.

src/app/features/search-bar/search-bar.component.ts
```diff
@@ -0,0 +1,250 @@
...
+      });
+    });
+  }
+
+  private _triggerFilter(): void {
+    this.taskSuggestionsCtrl.setValue(this.inputEl.nativeElement.value);
+  }
+
+  private _setUpFilter(): void {
+    this.filteredIssueSuggestions$ = this.taskSuggestionsCtrl.valueChanges.pipe(
```
johannesjo, line 122: This is a bit tricky and might lead to errors later on, as subscriptions to any previously assigned values might live on. Generally, I personally prefer to assign Observables once initially and never reassign them again.

src/app/features/search-bar/search-bar.component.ts
```diff
@@ -0,0 +1,250 @@
...
+      result = result.slice(0, 200);
+    } else {
+      this.tooManyResults = false;
+    }
+
+    return result;
+  }
+
+  private _shakeSearchForm(): void {
+    this.searchForm.nativeElement.classList.toggle('shake-form');
```
johannesjo, line 188: Oh nice! I wasn't aware that `classList.toggle` exists :)

src/app/features/tasks/task/task.component.ts
```diff
@@ -218,6 +221,15 @@ export class TaskComponent implements OnInit, OnDestroy, AfterViewInit {
       setTimeout(() => {
         this.focusTitleForEdit();
       });
+    } else {
+      this._route.queryParams.subscribe((params) => {
```
johannesjo, line 225: This needs an unsubscribe :)

src/app/features/tasks/task/task.component.ts
```diff
@@ -218,6 +221,15 @@ export class TaskComponent implements OnInit, OnDestroy, AfterViewInit {
...
+        const { focusItem } = params as SearchQueryParams;
+        if (focusItem === this.task.id) {
+          setTimeout(() => {
```
johannesjo, line 228: This also would need to be stored away and potentially be canceled onDestroy.

src/app/features/tasks/task/task.component.ts
```diff
@@ -218,6 +221,15 @@ export class TaskComponent implements OnInit, OnDestroy, AfterViewInit {
       setTimeout(() => {
```
johannesjo, line 221: I'd rather see this happening outside the task component as there can be so many instances of it on the page.

src/app/features/work-view/split/split.component.ts
```diff
@@ -53,6 +54,14 @@ export class SplitComponent implements AfterViewInit {
     this._updatePos(this.pos, false);
     this._renderer.addClass(this.splitTopEl, ANIMATABLE_CLASS);
     this._renderer.addClass(this.splitBottomEl, ANIMATABLE_CLASS);
+
+    this.route.queryParams.subscribe((params) => {
```
johannesjo, line 58: The split component is more of a dumb ui component, so I think it shouldn't be made aware of the routing and isInbBacklog. Makes more sense to me if the parent component does that and sets `splitPos` accordingly.

src/app/features/worklog/worklog.component.ts
```diff
@@ -34,8 +35,33 @@ export class WorklogComponent {
     private readonly _taskService: TaskService,
     private readonly _matDialog: MatDialog,
     private readonly _router: Router,
+    private readonly _route: ActivatedRoute,
   ) {}
 
+  ngAfterViewInit(): void {
+    this._route.queryParams.subscribe((params) => {
```
johannesjo, line 42: Needs an unsubscribe.

src/app/features/worklog/worklog.component.ts
```diff
@@ -34,8 +35,33 @@ export class WorklogComponent {
...
+      const { dateStr, focusItem } = params as SearchQueryParams;
+      if (!!dateStr) {
+        this.expanded[dateStr] = true;
+        this.focusTask(focusItem);
+      }
+    });
+  }
+
+  focusTask(id: string) {
```
johannesjo, line 51: Could you maybe add a comment on why we need setInterval here rather than just a simple setTimeout? I'd assume that it is needed because the element might not be present yet?

johannesjo (changes requested):

src/app/features/search-bar/search-bar.component.ts
```diff
@@ -0,0 +1,250 @@
...
+  private _tags: Tag[] = [];
+  private _searchableItems: SearchItem[] = [];
+
+  constructor(
+    private _taskService: TaskService,
+    private _projectService: ProjectService,
+    private _tagService: TagService,
+    private _router: Router,
+  ) {
+    this._taskService.allTasks$.pipe(take(1)).subscribe((t) => (this._tasks = t));
```
johannesjo, line 62: All these need unsubscribes.

johannesjo:
> Only in archived tasks when selecting an option from the search list with a mouse click, the item will not be focused after navigating to it. To be more precise: the item will be focused, because it actually scrolls to it, but it doesn't get the colored border. This works perfectly in non-archived tasks and when selecting an item by hitting enter. Plus, when you navigate by mouse click and the searched item shows up, the border will show up by hitting e.g. the shift key on the keyboard... really weird. Do you have any ideas? I think we could even leave it like that for now...

Not sure, but this also happens when I do this via the keyboard and I am already on the archive page. Could be a timing thing.

About the migrations: They're executed whenever the model version is older than the current one. In case you didn't change anything to the task model manually afterwards, the new model version is not written however, so until then the migrations are executed of every restart.

> Archived Projects:
> Somehow by chance I managed to get a project into an archived projects list. Don't know that feature, seems to work only for jira projects? What happens to the tasks of those projects? They do not show up in my searches...

This is just some leftover from the dataset I send you :). I did a big model change some time ago and in order to keep things managable I deactivated the archive project feature and never found the time to implement it again. 

> 

> Mobile version in browser:
> Tested it with the 6k+ tasks and it works fairly well. Highlighting by calling focus() doesn't work on mobile browser version but it did always bring me to the requested task...

Many of the focus styles are deactivated for the mobile version as there is no keyboard (most likely). So Focusing might work, but the style is just not shown.

src/app/features/search-bar/search-bar.component.ts
stkofler: 😅 actually, that has been carried over form add-task-bar. Written by you :)
But would love to hear why it is good practice.

src/app/features/search-bar/search-bar.component.ts
stkofler: Interesting topic. I like functional programming, too, but for me it isn't always easy to find the line between when to "use it" and when not. Especially in an OO framework like Angular. In this specific case I could pass the _tasks object but later on in _getArchivedDate() I'll be needing those tasks to be a class property anyways. At least I don't see a clean way of passing them unless to add the optionSelected listener programmatically. Then _searchableItems could be returned and passed to _setUpFilter() but the latter will only be called later on in ngAfterViewInit. For this particular case I don't think it'd be cleaner... Maybe I just don't get it, let me know what you think so I can improve :) and please note, I'm not arguing, just find the topic interesting and like to talk about it :)

src/app/features/search-bar/search-bar.component.ts
stkofler: had to google it myself :)

src/app/features/tasks/task/task.component.ts
stkofler: Since I'm rather new to Angular/RxJS I looked it up and found [this](https://stackoverflow.com/questions/41138081/do-i-have-to-unsubscribe-from-activatedroute-e-g-params-observables), saying that angular takes care of this specific subscription. The thing is, I actually really don't like that. I believe it's incredibly hard to remember all these rules and explicitly unsubscribing would be clearer and possibly safer... Will happily do so, if you prefer...

src/app/features/tasks/task/task.component.ts
stkofler: I totally agree on this one. In fact I started to look for a good place and wanted to ask but forgot. I think this could be handled once for both archived and active tasks. I couldn't quickly find a common "ancestor" component apart from the app.component and I'm not sure if that's a good place to put it.  What do you think?
Also, could it be that the worklog-page.component isn't used at all? There is no page for it and neither do I see it's selector to appear anywhere.

src/app/features/work-view/split/split.component.ts
stkofler: Sounds good to me, will take care of it.

src/app/features/worklog/worklog.component.ts
stkofler: Exactly, just adjusting the timeout period once didn't really work nicely. Sometimes the page itself took more than a second to load and I didn't want to set the timeout period too high. This way it tries every 300ms up to 5 times. I didn't really manage to find a perfect way that works every time.
In this sense, the approach I took for normal tasks is the cleanest one so far. For my tests it has worked really well with a static period of only 200ms. This way the timeout gets set when one of the last and smallest components is created: the task itself. 
I find it hard to figure out when the page is finally completely loaded (including animations and maybe even the full-page-spinner might mess with this working correctly). 
This is why I moved the blendInOutAnimation a level deeper into the add-task-bar/search-bar component so I could hook to onAnimationEvent().
If you have a better solution for this page-load/event-hook problem, I'd love to hear about it. As it might be useful once I move the navigation part (getElementById + focus) to a parent component.

src/app/features/search-bar/search-bar.component.ts
stkofler: I thought take(1) removed the need for unsubscription.  
I could also use toPromise if you prefer, or just unsubscribe explicitly. :)
Note aside, I noticed you didn't unsubscribe from _this.\_subs_ in add-task-bar either. Or does that work differently?
As I said. I'm rather new to all this...

Btw: firstValueFrom() is going to make this look much better :)

src/app/features/search-bar/search-bar.component.ts
stkofler: Ok good to know. Thanks for explaining this.
I don't even remember why I put it in it's own method as it is only called that one time in the beginning.
But at least I learned something :)

stkofler:
> Not sure, but this also happens when I do this via the keyboard and I am already on the archive page. Could be a timing thing.

Yeah I'll look into this again, maybe it will stop after I move the focusing logic to a parent component...
Another strange thing is that if you are already on the correct archived page, the page will **reload** even though we're just changing the focusItem parameter value. This odd reload might be messing with focusing... This does not happen on "active" task pages. There, upon selecting the item from the search-list both with mouse or keyboard, the route changes and the subscription triggers again calling for the new item to be focused.  But yeah I'll look into it again. 

> About the migrations: They're executed whenever the model version is older than the current one. In case you didn't change anything to the task model manually afterwards, the new model version is not written however, so until then the migrations are executed of every restart.

Indeed, the task model has not been changed at all. I'm not 100% sure the migration is necessary as it only happened after importing your large dataset.
On the other hand, can we be sure no one will ever import an old dataset with some **task.notes**  set to **undefined**?

> Many of the focus styles are deactivated for the mobile version as there is no keyboard (most likely). So Focusing might work, but the style is just not shown.

Hmm, so you think this might be fixable with css? I'll check that too.

src/app/features/search-bar/search-bar.component.ts
johannesjo: > I thought take(1) removed the need for unsubscription.

No, because allTasks$ is async. You might run into the (unlikely, but possible) case where the component is unloaded, before a value is emitted. 

Good catch with the add-task-bar. This a bug and potential memory leak. It's not as bad as the component should be loaded only once most of the time, but still, it's generally a good practice to take care of your subscriptions. I'll provide a fix.
 
> I could also use toPromise if you prefer, or just unsubscribe explicitly. :)

I only use toPromise as a convience thing for one time values inside a promise like flow. Here we're dealing with multiple subscriptions. It's better to do this via subscribe/unsubscribe I think.

src/app/features/search-bar/search-bar.component.ts
johannesjo: I am not against OO or using class properties. I was more thinking about doing it like this:
```TypeScript
    this._searchableItems = this._mapTasksToSearchItems(this._tasks);
```

Personally I think it's better because you don't need to go and read the more specific mapTasksToSearchItems, to see what variables are needed/affected. Gives you the info what value is set initially, which I think is good to know when you're trying to get an overview of what is going on here.

johannesjo:
> Indeed, the task model has not been changed at all. I'm not 100% sure the migration is necessary as it only happened after importing your large dataset.

I think the dataset might not be the most recent version, so this could be the reason why.

> On the other hand, can we be sure no one will ever import an old dataset with some **task.notes** set to **undefined**?

This is what the migrations are for. They are executed whenever a full data set is loaded which can happen on syncing, on app start and on import.


> Hmm, so you think this might be fixable with css? I'll check that too.

I haven't looked into the issue, but I think so.

src/app/features/worklog/worklog.component.ts
johannesjo: Thinking about it: This might have introduced some race condition, which leads to the unwanted focus behaviour. Simple fix: Save the intervalId and cancel potentially running intervals whenever the function is executed.

A slightly cleaner solution might couple this to worklogData$, but then again KISS as it is not straightforward neither. :)

johannesjo (changes requested):

src/app/features/search-bar/search-bar.component.ts
stkofler: oh ok, simply like that. Sure I'll fix that. Got a bit carried away :)

src/app/features/search-bar/search-bar.component.ts
stkofler: makes good sense. thanks.

src/app/features/worklog/worklog.component.ts
stkofler: > Thinking about it: This might have introduced some race condition, which leads to the unwanted focus behaviour. 
> Simple fix: Save the intervalId and cancel potentially running intervals whenever the function is executed.

I don't think so as it happened with a simple setTimeout, too. But I think we'll get to a solution for this later on. 

> A slightly cleaner solution might couple this to worklogData$, but then again KISS as it is not straightforward neither. :)

I have no idea what you mean here. Couple what to worklogData$? And why? This might also be a subject for later as I'd like to propose another idea. :)

src/app/features/search-bar/search-bar.component.ts
stkofler: Ok, I think it's best to just always unsubscribe.

src/app/features/search-bar/search-bar.component.ts
```diff
@@ -0,0 +1,250 @@
...
+    this._projectService.list$.pipe(take(1)).subscribe((p) => (this._projects = p));
+    this._tagService.tags$.pipe(take(1)).subscribe((t) => (this._tags = t));
+
+    this._mapTasksToSearchItems();
```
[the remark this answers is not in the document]
stkofler: I'd think so, too, but it works. this._tasks will be set inside the _mapTasksToSearchItems() method.
I don't really understand how that is possible. The assignment of _tasks should be waiting on the event queue at this point in time, and should therefore still be an empty array.
I could put the _mapTasksToSearchItems method in a later Angular lifecycle method but to be honest: I never really understood when the first tick on the event loop happens in-between/after those methods.

src/app/features/tasks/task/task.component.ts
stkofler: > I think the dataset might not be the most recent version, so this could be the reason why.
> This is what the migrations are for. They are executed whenever a full data set is loaded which can happen on syncing, on app start and on import.

Ok, so I suppose we keep the migration I added.

stkofler:
> I think the dataset might not be the most recent version, so this could be the reason why.
> This is what the migrations are for. They are executed whenever a full data set is loaded which can happen on syncing, on app start and on import.

Ok, so I suppose we keep the migration I added.

src/app/features/tasks/task/task.component.ts
stkofler: I was hoping to find an even better place to put the route interception. Somewhere it would work for both worklog and task component. work-view is not a parent of worklog afaik.
How would you feel about a NavigationService that intercepts and interacts with DOM by calling .focus(). I know it seems bad practice or unconventional but I think it might work well. It could also make checks like if we are already on the correct page. Which was something that we have both noticed doesn't work perfectly yet. I think this would remove quite a bit of navigation logic from the components. The search bar would then interact exclusively with this service instead of the router itself... Just an idea, haven't thought it through perfectly yet but some parts have worked fairly well in my quick tests. Let me know what you think :)

johannesjo: This could also work. Wouldn't  call it NavigationService but rather something more specific, as it likely will be a single purpose service. I would also be completely fine with putting this into the app.component btw. There is room to spare and if we can always move it later on, if we find a better place.

src/app/features/search-bar/search-bar.component.ts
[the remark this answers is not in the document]
johannesjo: > I could put the _mapTasksToSearchItems method in a later Angular lifecycle method but to be honest: I never really understood when the first tick on the event loop happens in-between/after those methods.

You can't count on this timing. What you can do is make _searchableItems to another observable that you assign via .pipe(map(()=>...))

src/app/features/search-bar/search-bar.component.ts
```diff
@@ -0,0 +1,252 @@
...
+    return !isArchivedTasks
+      ? this._taskService.allTasks$
+      : from(this._taskService.getArchivedTasks());
+  }
+
+  private _keepForLater(tasks: Task[], projects: Project[], tags: Tag[]): void {
+    this._tasks = tasks;
+    this._projects = projects;
+    this._tags = tags;
+  }
```
stkofler, line 87: I need tags for the icon generation for the search list. This one could be passed through about 3 methods to finally get where it is needed. As you prefer... 
Projects and tasks are needed later while handling onOptionSelected.
I hope it's Ok to "stash" these variables away for that occasion. Seemed to me the cleanest way and this time, to my understanding, they will be set without doubt.

src/app/features/tasks/task.service.ts
```diff
@@ -753,6 +753,15 @@ export class TaskService {
     ) as Task[];
   }
 
+  async getArchivedTasks(): Promise<Task[]> {
+    const archiveTaskState: TaskArchive = await this._persistenceService.taskArchive.loadState(
+      true,
+    );
```
stkofler, line 759: I'm passing _true_ to loadState() now. Otherwise, for every time the user switches task source while searching, the migration process gets triggered...

src/app/features/tasks/task/task.component.mixins.scss
```diff
@@ -56,19 +56,6 @@ $task-c-drag-drop-bg-dark: $dark12;
...
-  @if $direct-parent-only-sel {
-    :host-context(.isNoTouchOnly):focus > #{$direct-parent-only-sel} > & {
-      @content;
-    }
-  } @else {
-    :host-context(.isNoTouchOnly):focus & {
-      @content;
-    }
-  }
-}
```
stkofler, line 70: To my understanding this mixin gets called only once from task.component.scss. There an argument gets passed making the else statement unnecessary? In addition to that it seems to work perfectly with a small change in task.component.scss.

src/app/features/tasks/task/task.component.scss
```diff
@@ -116,18 +116,11 @@
...
-    // because of the border, this looks better
-    //top: 2px;
-    //bottom: 2px;
-
-    @include mq(xs) {
-      // NOTE: somehow ng build messes up, if we don't include it like that
-      border-color: $task-c-focus;
-      border-width: 1px;
-      border-style: solid !important;
-    }
```
stkofler, line 130: Not perfectly sure if this is OK, considering your comment about ng build messing up... I didn't notice any issues and it solved my problem of items not showing focus style on mobile...

src/app/features/worklog/worklog.component.scss
```diff
@@ -198,6 +198,12 @@ $this-table-outer-padding-left-right: $s*6;
     }
   }
 
+  > table > tr:focus {
+    outline: none;
+    box-shadow: 0px 0px 0px 1px $c-focus-border;
+    border-radius: 3px;
+  }
+
```
stkofler, line 206: solves the highlighting issue when selecting search item with mouse in worklog.

src/app/features/search-bar/search-bar.component.ts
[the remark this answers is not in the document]
stkofler: should be taken care of now. 
As a side note, this worked because subscription to store [will always deliver synchronously](https://github.com/ngrx/platform/issues/227#issuecomment-431682349). But, of course, this or the interface of this app might change, so better be safe on this one.

src/app/features/tasks/task/task.component.ts
stkofler: So I played around with it and it would definitely work well and allows for a lot of navigation possibilities. Still figured to leave it the way it is now, as it seems quite complete and a "NavigationService" would lead to a complete rewrite. The service could pretty much remove the need of route params for the purpose of searching/finding an item. It would always be there and could interact directly with components instead of relying on params being intercepted by components and handled by the themselves. Just some thoughts...

src/app/features/worklog/worklog.component.ts
stkofler: This should be fixed now with css update and interval clearer.

stkofler:
In the large dataset you sent me there is an archived task with number 12216 where the date is calculated wrongly by 1 day. Therefor the correct week does not unfold and the item cannot be found. Not sure what could be the cause of this. I'm using the getWorklogStr() method to get the date which should account for time zones... I've tested dates and parent/subtask combinations thoroughly but there are a lot of possible combinations... Don't know, could also be an issue in the dataset itself. Maybe I should add a snack message that says it couldn't find the exact location of the task?

src/app/features/search-bar/search-bar.component.ts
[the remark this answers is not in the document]
johannesjo: That's pretty crazy :D I would have never thought that subscribe is executed right away for store values (or anything really), which makes it a bit confusing I'd say as it might be sometimes sync and async and sometimes async only. But well something rather important learned today! Thank you!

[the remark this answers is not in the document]
johannesjo: Can you maybe add a comment about that above the subscriptions?

src/app/features/search-bar/search-bar.component.ts
```diff
@@ -0,0 +1,252 @@
...
+  tooManyResults: boolean = false;
+
+  private _attachKeyDownHandlerTimeout?: number;
+  private _tasks: Task[] = [];
+  private _projects: Project[] = [];
+  private _tags: Tag[] = [];
+  private searchableItems$: Observable<SearchItem[]> = this.isArchivedTasks$.pipe(
+    switchMap((isArchivedTasks) => this._loadTasks$(isArchivedTasks)),
+    withLatestFrom(this._projectService.list$, this._tagService.tags$),
+    tap(([tasks, projects, tags]) => this._keepForLater(tasks, projects, tags)),
```
johannesjo, line 66: It's cleaner to do this in a new subscriptiion rather than as a side effect here. If you have multiple subscriptions to searchableItems$ this get's executed for every single one even though it is just needed once.

src/app/features/search-bar/search-bar.component.ts
johannesjo: It's cleaner to do this in a new subscriptiion rather than as a side effect here. If you have multiple subscriptions to searchableItems$ this get's executed for every single one even though it is just needed once.

src/app/features/tasks/task.service.ts
johannesjo: Good choice! Tbh. I forgot about this mechanism for loadState completely.  :)

src/app/features/tasks/task/task.component.mixins.scss
johannesjo: Yes you're right! Good catch again! This is left-over code... Though I would prefer not mix in refactoring into this PR as it is complex enough already :)

src/app/features/tasks/task/task.component.scss
johannesjo: I think we need to test this for the actual electron build (the one created by `yarn buildAllElectron:prod && electron-builder`). I remember this not working there though this might be not a problem anymore as electron's chromium version has changed several times since.

johannesjo:
> In the large dataset you sent me there is an archived task with number 12216 where the date is calculated wrongly by 1 day. Therefor the correct week does not unfold and the item cannot be found. Not sure what could be the cause of this. I'm using the getWorklogStr() method to get the date which should account for time zones... I've tested dates and parent/subtask combinations thoroughly but there are a lot of possible combinations... Don't know, could also be an issue in the dataset itself. Maybe I should add a snack message that says it couldn't find the exact location of the task?

Unfortunately I can't find the task 12216 at all. Does it belong to a specific project?

src/app/features/search-bar/search-bar.component.ts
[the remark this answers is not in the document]
stkofler: Those subscriptions don't exist anymore in the latest version... If it comes up again, I'll add a comment. And yeah, pretty weird behavior...

src/app/app.component.ts
```diff
@@ -305,4 +316,27 @@ export class AppComponent implements OnDestroy {
...
+    this._intervalTimer = setInterval(() => {
+      counter += 1;
+
+      const el = document.getElementById(`t-${id}`);
+      el?.focus();
+
+      if (el && IS_MOBILE) {
+        el.classList.add('mobile-highlight');
+        el.addEventListener('blur', () => el.classList.remove('mobile-highlight'));
+      }
```
stkofler, line 335: Since we don't want focus to be visible on mobile I added this class to make it visible just after searching.

src/app/features/search-bar/search-bar.component.ts
```diff
@@ -0,0 +1,276 @@
...
+  private _attachKeyDownHandlerTimeout?: number;
+  private _openPanelTimeout?: number;
+  private _tasks$: Observable<Task[]> = this.isArchivedTasks$.pipe(
+    switchMap((isArchivedTasks) => this._loadTasks$(isArchivedTasks)),
+  );
+
+  private _searchableItems$: Observable<SearchItem[]> = this._tasks$.pipe(
+    withLatestFrom(this._projectService.list$, this._tagService.tags$),
+    map(([tasks, projects, tags]) => this._mapTasksToSearchItems(tasks, projects, tags)),
+  );
```
stkofler, line 73: All side effects removed. Projects and Tasks will be subscribed to in onOptionSelected further down.

src/app/features/search-bar/search-bar.component.ts
```diff
@@ -0,0 +1,276 @@
...
+    if (dateStr) return dateStr;
+
+    if (item.parentId) {
+      const parentTask = tasks.find((task) => task.id === item.parentId) as Task;
+      dateStr = Object.keys(parentTask.timeSpentOnDay)[0];
+      return dateStr ?? getWorklogStr(parentTask.created);
+    }
+
+    return getWorklogStr(item.created);
+  }
```
stkofler, line 177: Indeed the issue with task 12216 was here. Now fixed again, thanks.

src/app/features/search-bar/search-bar.component.ts
```diff
@@ -0,0 +1,276 @@
...
+
+  switchTaskSource(event?: MouseEvent) {
+    // trigger on mousedown to keep inputEl in focus
+    event?.preventDefault();
+    this.isLoading$.next(true);
+    this.isArchivedTasks = !this.isArchivedTasks;
+    this.isArchivedTasks$.next(this.isArchivedTasks);
+    if (IS_MOBILE) {
+      this._openPanelTimeout = window.setTimeout(() => this.autocomplete.openPanel());
+    }
```
stkofler, line 217: For some reason the autocomplete list closed after switching task source on mobile browser. This way it will stay open.

src/app/features/search-bar/search-bar.component.ts
```diff
@@ -0,0 +1,276 @@
...
+      );
+    } else {
+      this._subs.add(
+        this._tasks$.subscribe((tasks) => {
+          this.blurred.emit();
+          queryParams.dateStr = this._getArchivedDate(item, tasks);
+          this._router.navigate([location], { queryParams });
+        }),
+      );
+    }
```
stkofler, line 247: This looks quite a bit messier now but this way we don't need to load the data as a side-effect at the beginning.

src/app/features/tasks/task/task.component.scss
```diff
@@ -120,12 +120,18 @@
...
+      border-style: solid !important;
     }
   }
 
+  :host:focus.mobile-highlight > .inner-wrapper > & {
+    border-color: $task-c-focus;
+    border-width: 1px;
+    border-style: solid !important;
+  }
+
```
stkofler, line 134: This causes mobile task highlighting just after search and will disable after blur.

johannesjo:
Thank you very much again Stefan. Very nice work. I am merging this.

PR #1241, opened 2021-05-25
feat(idle): Flash taskbar when idle dialog pops out

kwongtn:
# Description

If the idle dialog pops out during usage, it will trigger the taskbar to flash (or, bounce on macs).
This is to pull user attention that their time is not being tracked.

On a side note, should I make it toggle-able? 😂

## Issues Resolved

None that I know of

## Check List


johannesjo:
Thank you very much!

PR #1260, opened 2021-06-01
Update AUR deployment action

alosarjos:
Looks like the AUR deployment action failed. If I'm right, this error should be fixed using the last version of the action. Can we give it a try?


johannesjo:
Sure thing! Thank you very much!

alosarjos:
I think you may need to trigger the action manually to retry with the version 7.0.

PD: Congratulations on the new release

johannesjo:
> PD: Congratulations on the new release

Thank you! :) 

Thanks also for reminding me about the action. Seems to run fine now!

PR #1290, opened 2021-06-11
refactor(Task): remove isHostFocus mixin

stkofler:
# Description

This removes the unused isHostFocus mixin and refactors a part that used to mess up during
`yarn buildAllElectron:prod && yarn electron-builder`

For me now this builds just fine.

### DIFF
--- a/src/app/features/tasks/task/task.component.scss
+++ b/src/app/features/tasks/task/task.component.scss
@@ -116,18 +116,11 @@
     }
   }
 
-  @include isHostFocus('.inner-wrapper') {
+  :host:focus > .inner-wrapper > & {
     border-radius: $task-border-radius;
-    // because of the border, this looks better
-    //top: 2px;
-    //bottom: 2px;
-
-    @include mq(xs) {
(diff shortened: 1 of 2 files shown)


src/app/features/tasks/task/task.component.scss
```diff
@@ -116,18 +116,11 @@
...
-
-    @include mq(xs) {
-      // NOTE: somehow ng build messes up, if we don't include it like that
-      border-color: $task-c-focus;
-      border-width: 1px;
-      border-style: solid !important;
-    }
+    border-color: $task-c-focus;
+    border-width: 1px;
+    border-style: solid !important;
```
johannesjo, line 123:
```suggestion
    // we don't want focus borders for touch only devices
    @include noTouchOnly() {
       border-color: $task-c-focus;
       border-width: 1px;
       border-style: solid !important;
    }
```

johannesjo:
Thank you! On mobile touch we don't want to show the borders, so I changed that part.

PR #1302, opened 2021-06-18
fix: change automatic translation of the 'Remote' word

isupovs:
The word 'Пульт' means 'TV Remote'


johannesjo:
Thank you very much!

PR #1303, opened 2021-06-18
Misc fixes

tiziodcaio:
# Description

- [X] Upgrade to Yarn 2;
- [X] Upgrade some github actions;
- [X] Removed cache action due to automatically added in .yarn/cache when yarn install
- [ ] added pnp (it doesn't work, but we have not to worry)

If I can ask you, please squash my changes, they are a bit messed up...


johannesjo:
Thank you very much for your effort! At this point I think that it is too early to switch to yarn2. This PR adds over 1800 artifacts, which I'd rather not have in the repo itself (just looking at the changes page crashed my browser twice).

Generally I'd recommend discussing such a big underlying change first, before going through the trouble of making a PR.

Also what do you mean by pnp?

tiziodcaio:
Also. Firstly the pull has the yarn 2 cache downloaded, so it has to many elements. If you want we can gitignore it and then add the cache gh action agani (maybe I can delete the pull and redo it better)
Pnp is an alternative to the node_modules folder, supported by yarn 2. I tried it, but it didn't work... Maybe after I will confirm it if I understand what's wrong. If you want I can close this PR and make a discussion before.

tiziodcaio:
I decide to close the repository and do it better...

johannesjo:
Alright! Thanks for the explanations!

Yarn2 is a bit special, which would make it harder for new people and lesser experienced web developers to get on board. So I wouldn't do the switch. What we could do instead is to switch back to npm again. I think the problems of the past are now resolved and it should be possibly a bit faster than yarn1.

tiziodcaio:
Ok great as you wish

PR #1306, opened 2021-06-18
Yarn 2 upgrade; actions update

tiziodcaio:
# Description

- [X] Yarn 2 updated (no caches)
- [X] actions updated


johannesjo:
Alright! Thanks for the explanations!

Yarn2 is a bit special, which would make it harder for new people and lesser experienced web developers to get on board. So I wouldn't do the switch. What we could do instead is to switch back to npm again. I think the problems of the past are now resolved and it should be possibly a bit faster than yarn1.

tiziodcaio:
I setted up yarn 2 in the node_modules folder, so it wouldn't be a problem for the compatibility with the past...

tiziodcaio:
I tried the actions before, and the build was a bit faster (one minute less circa)

johannesjo:
Thank you! I leave this open for now, if this is alright with you.

Currently I see only a limited advantage to upgrading and several possible drawbacks. First of all before I could greenlight such a big underlying change, I would have to read into the subject for at least several hours to fully understand the implications. This is time I currently don't have available and there is so much else to do. In addition to being possibly confusing to some developers (for example, at my current workplace, I am pretty sure that none of the developers ever worked with yarn2). Furthermore – speaking from experience – these kind of changes tend to introduce new problems further down the road. Currently I am happy with how the dependency management works. I haven't had any problems in the past year, install times are acceptable and it just works. This is good enough for now.

tiziodcaio:
Ok for leaving open
At the moment the changes from yarn 1:
Yarn up instead of yarn upgrade
The .yarn folder contains the cache folder which can be uploaded (not gitignored like before)
You can choose between keep the node_modules folder or don't use it (pnp), but in the second choice you have to add ALL your dependencies to the package.json, which at the moment can't be used
All the other yarn 1 command works

PR #1424, opened 2021-08-02
Fix shortcut in message about bookmark

aripollak:
If I press Shift+V, I currently get a message that says "Shift+N pressed, but open bookmarks shortcut is only available when in project context".

# Description

This should fix the message to display "Shift+V" instead. I haven't tested it though.

## Issues Resolved

## Check List


johannesjo:
This is great! Much appreciated! Thank you very much!

PR #1560, opened 2021-10-03
Allow Group and All scope for gitlab

cbergmann:
# Description

At work we plan to migrate from Jira to gitlab. We plan on keeping the issues in different projects and use the group issue board as an overview. This workflow is currently not possible with superproductivity as it only supports one gitlab project.

This Change introduces the ability to get gitlab issues from multiple projects. You can still use the project scope but you can also use group or global scope. The internal representation of a issue is changed from pure iid to project/iid format. New Issues are imported with a tag with the project name/path. 

This is my first time coding in TS and angular. Therefore it is totally possible that it does not follow the best practices of this environment. It does however seem to at least work for me.

Unfortunately that is also the reason that I did not implement any tests. I am not certain how I would implement these.

I hope this is at least a starting point for a future implementation of this feature.


johannesjo:
Thank you very much! 

This unfortunately breaks my existing gitlab project:
global-error-handler.class.ts:36 GLOBAL_ERROR_HANDLER 
{HANDLED_ERROR_PROP: "Gitlab: Http failure response for https://gitlab.c…ed_at&per_page=100&scope=undefined&page=1: 400 OK"}
HANDLED_ERROR_PROP: "Gitlab: Http failure response for https://gitlab.com/api/v4//issues?state=opened&order_by=updated_at&per_page=100&scope=undefined&page=1: 400 OK"
__proto__: Object

And also:
GET https://gitlab.com/api/v4//issues?search=a&scope=undefined&order_by=updated_at 400

 Setting a value for scope and raising the model version by 0.1 should likely fix this:
https://github.com/johannesjo/super-productivity/blob/master/src/app/features/project/migrate-projects-state.util.ts

src/app/features/issue/issue-service-interface.ts
```diff
@@ -38,7 +38,7 @@ export interface IssueServiceInterface {
     {
       task: Task;
       taskChanges: Partial<Task>;
-      issue: IssueData;
+      issue: IssueData | null;
```
johannesjo, line 41: I think this is not correct or am I missing something?

src/app/features/issue/providers/gitlab/gitlab-api/gitlab-api.service.ts
```diff
@@ -135,21 +140,65 @@ export class GitlabApiService {
...
     ).pipe(
       take(1),
       map((issues: GitlabOriginalIssue[]) => {
-        return issues ? issues.map(mapGitlabIssue) : [];
+        return issues ? issues.map((issue) => mapGitlabIssue(issue, cfg)) : [];
       }),
     );
   }
 
+  getFullIssueRef$(issue: string | number, projectConfig: GitlabCfg): string {
```
johannesjo, line 158: The functionName$ $ is reserved for methods returning an observable. Also the api service is intended to be used for api requests. getProject$, getFullIssueRef look more like utility/mapping functions to me and I would rather put them into a separate file. 

EDIT: I figured that you're just following the pattern established by apiLink. I leave it to you, to decide what pattern you prefer :)

src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts
```diff
@@ -111,88 +111,108 @@ export class GitlabCommonInterfacesService implements IssueServiceInterface {
...
+      if (!iidsByProject.has(project)) {
+        iidsByProject.set(project, []);
       }
-      issues.push(
-        ...(await this._gitlabApiService.getByIds$(ids as string[], cfg).toPromise()),
-      );
+      iidsByProject.get(project)?.push(task.issueId as string);
     }
 
+    iidsByProject.forEach(async (allIds, project) => {
```
johannesjo, line 152: Using Promise.all is probably much faster since the requests can be handled in parallel.

src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts
```diff
@@ -111,88 +111,108 @@ export class GitlabCommonInterfacesService implements IssueServiceInterface {
...
+      for (i = 0; i < allIds.length; i += paramsCount) {
+        (
+          await this._gitlabApiService
```
johannesjo, line 155: Using Promise.all is probably much faster since the requests can be handled in parallel.

src/app/features/issue/providers/gitlab/gitlab-issue/gitlab-issue.model.ts
```diff
@@ -41,7 +41,8 @@ export type GitlabIssue = Readonly<{
   comments: GitlabComment[];
   url: string;
   // NOTE: we use the issue number as id as well, as it there is not much to be done with the id with the api
-  id: number;
+  id: number | string;
```
johannesjo, line 44: A comment would be helpful, to indicate why it now can be both string and number.

cbergmann:
> 
> 
> Thank you very much!
> 
> This unfortunately breaks my existing gitlab project: global-error-handler.class.ts:36 GLOBAL_ERROR_HANDLER {HANDLED_ERROR_PROP: "Gitlab: Http failure response for [https://gitlab.c…ed_at&per_page=100&scope=undefined&page=1](https://gitlab.c%E2%80%A6ed_at&per_page=100&scope=undefined&page=1): 400 OK"} HANDLED_ERROR_PROP: "Gitlab: Http failure response for https://gitlab.com/api/v4//issues?state=opened&order_by=updated_at&per_page=100&scope=undefined&page=1: 400 OK" **proto**: Object
> 
> And also: GET https://gitlab.com/api/v4//issues?search=a&scope=undefined&order_by=updated_at 400
> 
> Setting a value for scope and raising the model version by 0.1 should likely fix this: https://github.com/johannesjo/super-productivity/blob/master/src/app/features/project/migrate-projects-state.util.ts

I added a function to only include the scope parameter when the setting is set.

PR #1588, opened 2021-10-16
Wait to exit startup tip until user event

### LINKED ISSUE #1284: Should delay main screen quotes until user decides
I feel like the person at the pictures when a quote or statement appears out of the blue for a brief period and I'm rushing to read it :)

Most of the time the quotes remain for milliseconds, literally; it would be nice if the quotes on startup remain there up until you move the mouse on the app.

Thanks

Fleker:
This keeps the tip on-screen until a mouse move or keypress

Fixes #1284

# Description

This allows the user to read the productivity tip and keep it there until they are ready to move forward with an input event, either mouse or key

## Issues Resolved

#1284

## Check List


Fleker:
In the original issue this is optionally something to make a setting, which is something that can be done if desierd.

PR #1589, opened 2021-10-16
Add button to task attachments to copy URL

### LINKED ISSUE #1522: copy URL to clipboard
## Problem Statement
I have a task and I attached a URL to it. When I show the attachments, an info panel appears on the right side. Moving the mouse on the URL attachment, 3 options appear: open URL, edit, and delete. Here I'd like to see a 4th option: copy URL to clipboard.

## Rationale

I use several web browsers, each for a specific task. When I click on open URL, then my main web browser will open this URL. However, if I use a different web browser at the moment, I would prefer to open the link there. The simplest way would be to copy the URL and then I'll open it manually in the browser in which I want.

Another use case: I want to send this URL to someone in an email. Then copying the URL would be the simplest solution.

## Workaround

A workaround solution would be to use the edit option and then copy the link without editing it. But it's more steps and when clicking on Edit, the URL is not highlighted by default. Here an easy fix would be to highlight the URL when clicking on Edit. Then Edit and Ctrl+c would do the trick.

Fleker:
Fixes #1522

# Description

Adds a button to the Task Attachment List for URLs that will copy the URL to clipboard automatically

## Issues Resolved

#1522

## Check List

### DIFF
--- a/src/app/features/tasks/task-attachment/task-attachment-list/task-attachment-list.component.html
+++ b/src/app/features/tasks/task-attachment/task-attachment-list/task-attachment-list.component.html
@@ -31,7 +31,12 @@
         <mat-icon *ngIf="attachment?.type==='IMG'">search</mat-icon>
         <mat-icon *ngIf="attachment?.type!=='IMG'">open_in_new</mat-icon>
       </button>
-      <button (click)="copy(attachment)" class="copy-btn" mat-flat-button *ngIf="attachment?.type!=='IMG'">
+      <button
+        (click)="copy(attachment)"
+        class="copy-btn"
+        mat-flat-button
+        *ngIf="attachment?.type!=='IMG'"
+      >
         <mat-icon>content_copy</mat-icon>
       </button>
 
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much for opening this PR. I like the changes, but there are some minor problems with the code, which I commented above.

johannesjo:
Thank you very much for the adjustments. It seems like that there are still some linting errors. Some of them can be fixed automatically via `./node_modules/.bin/prettier --write ./src/**/*.ts`

https://github.com/johannesjo/super-productivity/pull/1589/checks?check_run_id=3943768368
```
Error:   34:14  error  Replace `·(click)="copy(attachment)"·class="copy-btn"·mat-flat-button·*ngIf="attachment?.type!=='IMG'"` with `⏎········(click)="copy(attachment)"⏎········class="copy-btn"⏎········mat-flat-button⏎········*ngIf="attachment?.type!=='IMG'"⏎······`  prettier/prettier

/home/runner/work/super-productivity/super-productivity/src/app/features/tasks/task-attachment/task-attachment-list/task-attachment-list.component.ts
Error:   62:3   error  Missing return type on function                                                               @typescript-eslint/explicit-function-return-type
Error:   65:25  error  Replace `name:·'clipboard-write'` with `·name:·'clipboard-write'·`                            prettier/prettier
Error:   67:25  error  Replace `"granted"·||·result.state·==·"prompt"` with `'granted'·||·result.state·==·'prompt'`  prettier/prettier
Error:   69:65  error  Insert `;`                                                                                    prettier/prettier
``` 


Let me know if you have any questions!

PR #1636, opened 2021-10-30
test: generate timestamps based on local time-zone

### LINKED ISSUE #1635: Unit test wrong date timezone assumptions
Running tests `ng test` or through `npm run dist` in my local timezone (Lisbon - WEST - Western European Summer Time) two testes fail in `create-blocker-blocks.spec.ts`: [should work for advanced scenario:334](https://github.com/johannesjo/super-productivity/blob/f8bfe791e2ac476b12457e259394105d427c3552/src/app/features/timeline/map-timeline-data/create-blocker-blocks.spec.ts#L334), [should merge complicated blocks:437](https://github.com/johannesjo/super-productivity/blob/f8bfe791e2ac476b12457e259394105d427c3552/src/app/features/timeline/map-timeline-data/create-blocker-blocks.spec.ts#L437).

Both tests fails because of hardcoded values of timestamp millis (in UTC) checked against [getDateTimeFromClockString](https://github.com/johannesjo/super-productivity/blob/f8bfe791e2ac476b12457e259394105d427c3552/src/app/util/get-date-time-from-clock-string.ts#L14) that return a timestamp that resulted from [Date#setHours(hours)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setHours). The `Date#setHours` method set hours in **local time zone**, so produce a different UTC timestamp depending on your current time zone. So anyone that tries to run unit test in different time zone would get a wrong expected value.

One possible fix for the hard coded timestamp would be to use `getDateTimeFromClockString` instead of those precomputed values. This way expected values would be produces for the same time zone.

pcmind:
Fix failing test by using `createSortedBlockerBlocks` to generate timestamps instead of fixed ones to ensure they can run anywhere.

Fixes #1635


johannesjo:
Thank you very much!!

PR #1641, opened 2021-10-31
Fix russian locale

punyflash:
# Description

I was using this app for a week and love it! However, Russian locale seems to be awful having a ton of style mistakes, and I ended up using English. So I thought about fixing it by myself.

## Check List


johannesjo:
Thank you very much!

PR #1672, opened 2021-11-13
Split take a break notification message into title and body

gkobeaga:
# Description

Splits `take a break` notification message into title and body.


src/app/features/take-a-break/take-a-break.service.ts
```diff
@@ -202,7 +202,8 @@ export class TakeABreakService {
     this._triggerDesktopNotification$.subscribe(([timeWithoutBreak, cfg]) => {
       const msg = this._createMessage(timeWithoutBreak, cfg.takeABreak);
       this._notifyService.notifyDesktop({
-        title: msg,
+        title: 'Take a break!',
```
johannesjo, line 205: This also would need to be internationalized. You can do so by adding the string to the `en.json` maybe under `GCF.TAKE_A_BREAK.NOTIFICATION_TITLE` and then running `npm run int`.

johannesjo:
Excellent! Thank you very much!

PR #1791, opened 2022-01-02
Add Croatian translation

milotype:
# Description
The translation is not complete, but it would be great, if you could release a new/nightly version, so I could check the Croatian translation in real life (on my MacOS) and make possibly necessary corrections/adjustments.

### DIFF
--- a/src/assets/i18n/hr.json
+++ b/src/assets/i18n/hr.json
@@ -1,0 +1,1371 @@
+{
+  "ANDROID": {
+    "NOTIFICATIONS": {
+      "SYNC_CONFLICT_TITLE": "Dogodila se greška u sinkronizaciji",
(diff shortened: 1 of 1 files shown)


milotype:
While you review my translation, I'd like to suggest that you use a translation platform for translations (my favorite is [Weblate](https://weblate.org/), it has different hosting plans, even self-hosted, which is free of charge). But you'll have to check yourself, if it fits for your project ...

johannesjo:
Thanks for the suggestion. For I'd like to keep it as simple as possible. Currently most translations are generated automatically with only very sporadic updates by volunteers. Under these circumstances the current workflow is much more efficient.

johannesjo:
And also thank you very much for this PR! :)

PR #1806, opened 2022-01-06
Update Croatian translation

milotype:
Croatian translation should be complete now, with some corrections


johannesjo:
Thank you very much!

PR #1811, opened 2022-01-07
Update Croatian

milotype:
# Description
- translated new source strings for "welcome dialog"
- minor corrections in Croatian translation

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ There are several ways to help.
 
 7. **[Make a feature or improvement request](https://github.com/johannesjo/super-productivity/issues/new)**: Something can be be done better? Something essential missing? Let us know!
 
-8. **Translations, Icons, etc.**: You don't have to be programmer to help. Some of the icons really need improvement and many of the translations could use some love.
+8. **[Translations](https://github.com/johannesjo/super-productivity/tree/master/src/assets/i18n), Icons, etc.**: You don't have to be programmer to help. Some of the icons really need improvement and many of the translations could use some love.
 
 9. **[Sponsor the project](https://github.com/sponsors/johannesjo)**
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -70,6 +70,8 @@ Super Productivity is an advanced todo list app with integrated [Timeboxing](htt
 - Attach context information to tasks and projects. Create **notes**, attach **files** or create **project level bookmarks** for links, files and even commands.
 - Super Productivity **respects your privacy** and **does NOT collect any data** and there are no user accounts or registration. **You decide where you store your data!**
 - It's **free** and **open source** and always will be.
+- It's **translated** into several languages.
```
johannesjo, line 73: The list is meant to be brief. I might be wrong, but being translated into many languages is not major selling point for the app, but more or less expected.

milotype: I defnitely disagree … Many apps are (still) not localized/translated at all … providing information that the app is translated into other languages is, IMO,  essential for interested users …

Ideally, the website itself would be translated (according to user's browser preferences or by language menu selection) into the specific user's language … or at least provide informations about whether or not the app is translated into the user's language …

But that's just my opinion ;-)

johannesjo: I completely agree about the website. That would be the best way to indicate multiple languages being available. A GitHub Readme in English doesn't feel like the right place to highlight, especially since some of the languages (my own mother tongue German for example ;)) are not translated very well. I also wouldn't care about an app being translated into *many* languages. I would only care about the language I would like to use.

johannesjo: Thank you very much for the suggestion. Personally I am using babel translate and the json format should work with different apps. I don't want to force a specific platform onto contributors.

Issue #1813, opened 2022-01-07
Translation of Today


milotype:
### Your Environment

- Version used: 7.9.2
- Operating System and version: MacOS 10.11.6

### Expected Behavior
Title "Today" should translate into other languages

### Current Behavior
Title "Today" doens't translate into other languages

### Steps to Reproduce (for bugs)
Open the app. The title "Today" does not translate.

johannesjo:
Thanks for opening this up! I agree! It's a bit tricky since, one can also edit the title of the today tag themselves, so maybe it makes sense to go for a placeholder default identifier and translate it as long its no other value.

PR #1822, opened 2022-01-09
fix(i18n): title today translate #1813

chris-amaya:


src/app/core-ui/main-header/main-header.component.html
```diff
@@ -16,7 +16,7 @@
       mat-ripple
       routerLink="/active/tasks"
     >
-      {{activeWorkContext?.title}}
+      {{T.PDS.TODAY|translate}}
```
johannesjo, line 19: This way we can allow users to edit the title:
```suggestion
      {{activeWorkContext?.title ==='Today' ? T.PDS.TODAY|translate: activeWorkContext?.title}}
```

johannesjo:
Thank you very much!

PR #1900, opened 2022-02-01
Update de.json

LeQuadratico:
Update de.json file

-corrections
-extended to match en.json file


johannesjo:
Wow! Super, vielen Dank!

PR #1953, opened 2022-03-08
Fix CORS on Android for WebDAV

penn5:
See https://github.com/johannesjo/super-productivity-android/pull/17 for the relevant API.

# Description

This is needed to fix CORS in WebDAV sync in the Android app

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/1847

## Check List

I haven't tested this on the stable version of the Android app but in theory it should work.

### DIFF
--- a/src/app/features/android/android-interface.ts
+++ b/src/app/features/android/android-interface.ts
@@ -56,6 +56,28 @@ export interface AndroidInterface {
     progress: number, // -1 => undefined; 999 => indeterminate; 333 => show play but no progress bar
   ): void;
 
+  // WebDAV
+  makeHttpRequestWrapped(
+    url: string,
+    method: string,
+    data: string,
+    username: string,
+    password: string,
+    readResponse: boolean,
+  ): Promise<object>;
+
+  makeHttpRequest?(
+    rId: string,
+    url: string,
+    method: string,
+    data: string,
+    username: string,
+    password: string,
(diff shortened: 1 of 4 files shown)


johannesjo:
This is great! Thank you very much! This also opens up the door for making Jira work on mobile android.

src/app/features/android/android-interface.ts
```diff
@@ -56,6 +56,28 @@ export interface AndroidInterface {
...
+    rId: string,
+    url: string,
+    method: string,
+    data: string,
+    username: string,
+    password: string,
+    readResponse: boolean,
+  ): void;
+
+  makeHttpRequestCallback(rId: string, result: Object): void;
```
johannesjo, line 79: The linting doesn't like object. If you're sure that this is an object you can use something like 
```
{[key: string]: string | number}
``` 
Or possibly `unknown` or even `any`

penn5: Will `{}` work? I haven't used TS before 😅

johannesjo: No unfortunately not :) The quickest compromise might be:
```
{[key: string]:  any }
```

penn5:
Uh, you do know the lint failed right?

johannesjo:
@penn5 yes, but I decided it's quicker, if I just fix it locally :)

PR #1971, opened 2022-03-19
fix pagination requests in gitlab

ErezAmihud:
# Description
Fix a small problem with parsing the data. Every pagination call was parsed as an issue object when it shouldn't, because it could be a comment.

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/1970

### DIFF
--- a/src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts
+++ b/src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts
@@ -71,7 +71,7 @@ export class GitlabCommonInterfacesService implements IssueServiceInterface {
         this.isEnabled(gitlabCfg) && gitlabCfg.isSearchIssuesFromGitlab
           ? this._gitlabApiService.searchIssueInProject$(searchTerm, gitlabCfg).pipe(
               catchError((err) => {
-                console.log('Error searching issue', err);
+                console.error('Error searching issue', err);
                 return [];
               }),
             )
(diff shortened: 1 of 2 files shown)


src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts
```diff
@@ -69,9 +69,12 @@ export class GitlabCommonInterfacesService implements IssueServiceInterface {
     return this._getCfgOnce$(projectId).pipe(
       switchMap((gitlabCfg) =>
         this.isEnabled(gitlabCfg) && gitlabCfg.isSearchIssuesFromGitlab
-          ? this._gitlabApiService
-              .searchIssueInProject$(searchTerm, gitlabCfg)
-              .pipe(catchError(() => []))
+          ? this._gitlabApiService.searchIssueInProject$(searchTerm, gitlabCfg).pipe(
+              catchError((err) => {
+                console.log('Error searching issue', err);
```
johannesjo, line 74:
```suggestion
                console.error('Error searching issue', err);
```

johannesjo: Would you think that providing feedback to the user (e.g. via a snack message) might be a good idea here?

johannesjo:
Thank you very much for this PR! Please check the comments above.

src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts
ErezAmihud: I don't, this issue will most likely accure when there is a problem parsing an issue because of api change or programming error - so lets not normalize it with a message but send an error so the person working on the issue could see the priblem

ErezAmihud:
Fixed with https://github.com/johannesjo/super-productivity/pull/1971

PR #1979, opened 2022-03-25
Updated Croatian translation

milotype:
# Description

Updated Croatian translation

## Suggestion

I'd like to **reiterate my suggestion** to you use (if possible) a translation platform for translations like [Weblate](https://weblate.org/). I'm aware of the fact that this implicates some more work on your side, but IMO, it helps translators …

I found several changes/new strings in the `en.json`. I found them just by chance, as I was reviewing the english and croatian `.json` files side by side with BBEdit.


johannesjo:
Thanks for the update!

PR #1988, opened 2022-03-30
Added tests for weekly repeated tasks

### LINKED ISSUE #1902: RepeatTask ignores days on which the app hasn't been used
So, I have a bunch of tasks that repeat every week. It doesn't matter when I do that within the week, so I just chose a fitting day (eg. Monday). However, sometimes I take that day off and hence don't start SP at all. In the past (pre 7.10) these tasks would be added to my 'today' when I open SP the next time (with the neat 'overdue' pop-up  :) and I'd simply add them to my 'today' – However since I updated to 7.10 (7.10.1 to be precise) last week, those tasks don't show up anymore. 

I checked the repeat config (due to the migration) and it seems fine with 'custom repeat' set to the previous day(s). So, I really think there is just smth. going on with days SP wasn't used on. And if so, then it might be a huge issue for any monthly or even weekly repeatTask…

### Your Environment

- Version used:  7.10.1 (AUR)
- Operating System and version: arch; manjaro
- Desktop Environment:  XFCE
- Browser Name and version: 

### Expected Behavior

RepeatTask/Tasks are created even if their due date lays days in the past 
(IIRC, it even created multiple instances, if multiple due dates laid in the past; though not sure about that one)

_(Keep in mind, though probably unrelated) New instances should be created, even if a previous instance isn't marked as 'done' (again, pre 7.10 behavior I saw)._

### Current Behavior

No new instances of the tasks were created when one day was skipt.

Migrate from pre 7.10 to 7.10 or later. (However, I don't think that's related, as the repeatTask configs seem fine; and I had no errors(console/logs) or other issues after migrating)

1. RepeatTask with custom config set to schedule weekly on day x
2. Don't start SP on day x
3. Start SP on day x+1 – no pop-up / creation / reaction whatsoever
´

### Console Output

 \- Can't see any useful logs/stuff about repeat; Is there some sort of tag I can filter for like the `[M]` for migration? -

### Error Log (Desktop only)

 \- Just non helpful `[info]` entries

EngelPika32:
# Description

This PR adds some tests for weekly repeated tasks.
Some tests fail as they expect a fixed behavior already (PR with fixes is upcoming). 

Tangent: Just noticed during PR creation that a small refactor is included here as well. Wasn't intended to be in this exact PR, but seems fine to me. Can remove/force-push it out as well, if needed.

## Related Issues

As I said, I'm adding some tests to help resolve #1902 

## Todo:

- [ ] Check/Verify tests for Daily
- [ ] Tests for Monthly
- [ ] Tests for Yearly
- [ ] Run prettier / remove prettier-ignore comments
- [ ] Create PR with fixes

(new PR, for the todos, is upcoming)

### DIFF
--- a/src/app/features/task-repeat-cfg/store/task-repeat-cfg.reducer.spec.ts
+++ b/src/app/features/task-repeat-cfg/store/task-repeat-cfg.reducer.spec.ts
@@ -6,7 +6,7 @@ const DUMMY_REPEATABLE_TASK: TaskRepeatCfg = {
   id: 'REPEATABLE_DEFAULT',
   title: 'REPEATABLE_DEFAULT',
   quickSetting: 'DAILY',
-  lastTaskCreation: 60 * 60 * 1000,
+  lastTaskCreation: 60 * 60 * 1000, // debug: Why this number? // todo: This number might cause issues with the tests later on…
   defaultEstimate: undefined,
   projectId: null,
   startTime: undefined,
@@ -36,6 +36,17 @@ const dummyRepeatable = (id: string, fields: Partial<TaskRepeatCfg>): TaskRepeat
   id,
   ...fields,
 });
+
+const datePrintingOptions = {
+  weekday: 'long',
+  year: 'numeric',
+  month: 'long',
+  day: 'numeric',
+  hour: '2-digit',
+  minute: '2-digit',
+  second: '2-digit',
+  hour12: false,
+};
 
 describe('selectTaskRepeatCfgsDueOnDay', () => {
   describe('for DAILY', () => {
@@ -178,6 +189,24 @@ describe('selectTaskRepeatCfgsDueOnDay', () => {
       expect(resultIds).toEqual(['R1']);
     });
 
+    it('should return cfg for startDate in the past (2)', () => {
+      const result = selectTaskRepeatCfgsDueOnDay.projector(
+        [
+          dummyRepeatable('R1', {
+            repeatCycle: 'WEEKLY',
+            repeatEvery: 1,
+            startDate: '2022-01-10',
+            monday: true,
+          }),
+        ],
+        {
+          dayDate: new Date('2022-01-17').getTime(),
+        },
+      );
+      const resultIds = result.map((item) => item.id);
+      expect(resultIds).toEqual(['R1']);
+    });
+
     it('should return available for day', () => {
       const result = selectTaskRepeatCfgsDueOnDay.projector(
         [
@@ -207,6 +236,12 @@ describe('selectTaskRepeatCfgsDueOnDay', () => {
           startDate: '2022-01-10',
           repeatCycle: 'WEEKLY',
         }),
+        dummyRepeatable('R3', {
+          wednesday: true,
+          friday: true,
+          startDate: '2022-01-10',
+          repeatCycle: 'WEEKLY',
+        }),
       ];
       // eslint-disable-next-line no-mixed-operators
       const FULL_WEEK = [0, 1, 2, 3, 4, 5, 6].map((v) => FAKE_MONDAY_THE_10TH + v * DAY);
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much for this @EngelPika32  and for all the other work you have done for this project :) It's really good to have some help!

Do you need some feedback on the WIP? How should we proceed?

EngelPika32:
> Do you need some feedback on the WIP? How should we proceed?

I'm open to feedback if you have some. Better have it now and apply it to the rest directly than waiting, no?

I'm currently working on the tests for monthly (and yearly). 
Though, as I don't use them, I'm slightly clueless about how to test edge cases properly with the current test setup. Will probably figure smth. out by the end of the week. _(at least, I hope so. I'm open for input and test data!)_

At the same time, I can already say, weekly is fixed in my local dev branch. Haven't created a PR for the fix yet, as the surrounding code is pretty much WIP – trying various ways to efficiently filter monthly & yearly.

Unfortunately, I can't apply the same fix I did for weekly to monthly & yearly due to a likely performance hit. 
Hence I'm open to letting someone else figure out an efficient way to fix monthly & yearly. I'd happily review the changes then.


PS: Feel free to merge this PR as "added tests for weekly repeated tasks". 
I originally wrote 'WIP' cuz I thought I'll have the other tests done by now and wanted to add them straight away. But, I dunno how long they'll take me.

EngelPika32:
mmh… do you think we should release a fix for weekly repeated tasks without having the others fixed? It'd certainly improve the usability of SP directly (by re-creating pre v7.10 behavior). Could cherry-pick required changes and create a PR tomorrow.

src/app/features/task-repeat-cfg/store/task-repeat-cfg.reducer.spec.ts
```diff
@@ -6,7 +6,7 @@ const DUMMY_REPEATABLE_TASK: TaskRepeatCfg = {
   id: 'REPEATABLE_DEFAULT',
   title: 'REPEATABLE_DEFAULT',
   quickSetting: 'DAILY',
-  lastTaskCreation: 60 * 60 * 1000,
+  lastTaskCreation: 60 * 60 * 1000, // debug: Why this number? // todo: This number might cause issues with the tests later on…
```
johannesjo, line 9: A date in the past is needed and `Date.now()` (the default value) won't work for the tests and because `0` doesn't yield valid dates. Do you have a suggestion maybe?

src/app/features/task-repeat-cfg/store/task-repeat-cfg.reducer.spec.ts
```diff
@@ -195,71 +224,275 @@ describe('selectTaskRepeatCfgsDueOnDay', () => {
       expect(resultIds).toEqual(['R1']);
     });
 
-    it('should work correctly for a week', () => {
```
johannesjo, line 198: Did you remove this, because you think it's an invalid test?

src/app/features/task-repeat-cfg/store/task-repeat-cfg.reducer.spec.ts
```diff
@@ -195,71 +224,275 @@ describe('selectTaskRepeatCfgsDueOnDay', () => {
...
+        },
+      );
+      const resultIds = result.map((item) => item.id);
+      expect(resultIds).toEqual([]);
+    });
+
+    // Multiple Weeks Test
+    // Please keep in mind, this test triggers every day of the week. Hence, it does not test the behaviour, if a day is skipped.
+    // For convenience of the test, lastTaskCreation is always the day before the day we test.
+    const expectedMultipleTest = {
```
johannesjo, line 347: Something like 
```
[ 
  [condition, result],
  [condition, result], 
  [condition, result]
].forEach(([condtion,result]) => it('....'))
```  might improve readability.

File: src/app/features/task-repeat-cfg/store/task-repeat-cfg.reducer.spec.ts [hunk 4]
```diff
@@ -195,71 +224,275 @@ describe('selectTaskRepeatCfgsDueOnDay', () => {
...
+      // R1 is every 2 weeks
+      // R3 repeats every week, (starts in the first week, at the first day)
+      0: [['R1'], [], ['R3'], [], ['R3', 'R4'], [], []], // R2 hasn't started // R4 starts this Friday and is scheduled Fridays
+      1: [[], [], ['R3'], [], ['R3'], [], []], // R2 starts this Sunday
+      2: [['R1'], [], ['R2', 'R3'], [], ['R3'], [], []],
+      3: [[], [], ['R2', 'R3'], [], ['R3', 'R4'], [], []], // R4 repeats in the third week
+      4: [['R1'], [], ['R2', 'R3'], [], ['R3'], [], []],
+    };
+    for (let w = 0; w < 5; w++) {
+      it('should work correctly for multiple weeks (week ' + w + ')', () => {
```
johannesjo, line 357:
```suggestion
      it('should work correctly for multiple consecutive weeks (week ' + w + ')', () => {
```

johannesjo:
> mmh… do you think we should release a fix for weekly repeated tasks without having the others fixed? It'd certainly improve the usability of SP directly (by re-creating pre v7.10 behavior). Could cherry-pick required changes and create a PR tomorrow.

Merging this in smaller chunks sound like a good idea to me, since it makes reviewing much easier.

johannesjo:
I provided a bit of input on the code. I hope it's helpful! Please let me know, if you have any questions!

johannesjo:
> Unfortunately, I can't apply the same fix I did for weekly to monthly & yearly due to a likely performance hit.
> Hence I'm open to letting someone else figure out an efficient way to fix monthly & yearly. I'd happily review the changes then.

May I ask how you're trying to solve this? Is the idea, to create the tasks if they haven't been created before even though it's not the exact date they have been planned for?

src/app/features/task-repeat-cfg/store/task-repeat-cfg.reducer.spec.ts
EngelPika32: Tbh. I don't know. I just left some of my first impression comments in, as they show possible issue sources. _(Like with the commented-out test)_

[[[ ~~I used "the day before today" in some of the tests. Though, that won't work very well for monthly and yearly.~~ Nvm, are set manually as well.
My monthly & yearly tests specify the `lastTaskCreation` themselves, however. So, there shouldn't be an issue with changing it _(or leaving it as is)_. ]]]

src/app/features/task-repeat-cfg/store/task-repeat-cfg.reducer.spec.ts
EngelPika32: Yes and no. I originally expanded on this test and then added my own. But… it's now obsolete.
As the commit message of 3f65adeead3f678723eec602c785b3d6c59913ba states:
> This test lacks correct lastTaskCreation attributes and is supervised/covered by another test.
It may be re-added later.

src/app/features/task-repeat-cfg/store/task-repeat-cfg.reducer.spec.ts
EngelPika32: Could you explain the enhancement of this change a bit further? I don't understand the benefit. – What is the `condition` in your change?
_(Using more complex constructions doesn't improve readability, IMHO.)_

EngelPika32:
> I would however clean up the code so that the out commented tests are in again, if possible and that there are no logs left.

Out commented tests: See the above comments in the review.
no logs left: Wdym? Calls to `console.log` within tests are fine (IMHO). They help resolve occurring issues on failing tests. _(IDE can show logs per test)_ 
Actually, I wanted to ask if I can keep/commit `console.debug` logs in the `task-repeat-cfg.reducer.ts`? (relevant for upcoming PRs). 

> May I ask how you're trying to solve this? Is the idea, to create the tasks if they haven't been created before even though it's not the exact date they have been planned for?

Well, that's kinda the whole point of the referenced issue, isn't it? A bit more fitting:  "Create tasks, if they should've been created in the past."
Hence the `lastTaskCreation` is very relevant. This allows the required checks & prevents the creation of too many tasks as well. 


And regarding my progress: I got around my issue, added a test for monthly, and fixed monthly. (Though, still have to test against edge cases like February.)

johannesjo:
> Calls to `console.log` within tests are fine (IMHO). They help resolve occurring issues on failing tests.

While they help for that particular test if you would need to debug it, they add noise that makes it a little bit harder to debug any other possible issues with tests. As with regular code, I think it is a good rule of thumb to get rid of all logs, whenever possible, unless they're needed. 

Given that the tests are currently failing, I wouldn't want to merge this back until those are not failing anymore, since this effectively would block everything else. It probably also makes more sense to review the tests in the context of the code they will accompany.

I also think we need to address the conceptual implications before merging code for this. I will continue the discussion in the original issue. Let's continue from there.

PR #1995, opened 2022-04-01
Add iCalendar 'sequence' handling

kanunnikau:
https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.7.4

   Description:  When a calendar component is created, its sequence
      number is 0.  It is monotonically incremented by the "Organizer's"
      CUA each time the "Organizer" makes a significant revision to the
      calendar component.

Some calendar clients do not see updates (completion) submitted by SuperProductivity as the 'sequence' number is unchanged.

(I have not tested this, just wrote code blindly. Please review)

# Description

Makes SuperProductivity increment VTASK SEQUENCE property when marking as complete.

## Issues Resolved

~I didn't look for any reported issues :)~
Quick search did not yield any similar issues.

## Check List


johannesjo:
Thank you very much for this fix! :)

src/app/features/issue/providers/caldav/caldav-client.service.ts
```diff
@@ -402,6 +402,14 @@ export class CaldavClientService {
     }
     todo.updatePropertyWithValue('last-modified', now);
     todo.updatePropertyWithValue('dtstamp', now);
+
```
johannesjo, line 405: Could you maybe also add a comment on why we need this?

src/app/features/issue/providers/caldav/caldav-client.service.ts
```diff
@@ -402,6 +402,14 @@ export class CaldavClientService {
     }
     todo.updatePropertyWithValue('last-modified', now);
     todo.updatePropertyWithValue('dtstamp', now);
+    
+    var sequenceInt = 0;
+    const sequence = todo.getFirstPropertyValue('sequence');
+    if (sequence) {
+      sequenceInt = parseInt(sequence);
+    }
+    sequenceInt++;
```
johannesjo, line 411:
```suggestion
    const sequence = todo.getFirstPropertyValue('sequence');
    const sequenceInt = sequence
      ? parseInt(sequence) + 1
      : 1;
```

src/app/features/issue/providers/caldav/caldav-client.service.ts
kanunnikau: Done

johannesjo:
Thanks for the update! Could you please check again? There are still some linting errors.

kanunnikau:
> There are still some linting errors

Done

PR #2013, opened 2022-04-13
Fix #2002

### LINKED ISSUE #2002: Pomodoro timer cannot be paused/stopped in the webapp version
### Your Environment

- Version used: 7.11.2 (webapp)
- Desktop Environment: 
- Browser Name and version: Firefox 98.0.2 / Chrome 99.0

### Expected Behavior

When hovering over the running timer ▶ icon, it should display the ⏭ (break) and ⏹ icons, and those should _stay visible and clickable_ in order to start a break or stop the timer while it's running. This behavior did happen on previous versions, not sure exactly which version broke it but it's a recent issue.

### Current Behavior

On moving the mouse down from the ▶ icon in order to pause or stop the timer, the "dropdown" immediately collapses back up, preventing me from clicking it. This does not happen on the Windows desktop version of the app, only happens on the webapp version as far as I can tell.

1. Start the pomodoro timer by clicking ▶
2. Mouse over the playing timer to display the ⏭ and ⏹ buttons
3. Try to move the mouse over to the ⏭ and ⏹ buttons in order to pause/stop the timer
4. The buttons immediately disappear, not allowing you to click them.

Cyborger:
# Description

Fix for #2002 by removing the matTooltip on the toggleStartTask button
when pomodoro is enabled.

I don't think this is ideal though. In my opinion, we should keep the tooltips on the left side of the buttons and that's what I want to do but there isn't enough translation strings as a `skip session` and `stop session` strings are needed.

## Issues Resolved

This change fixes the issue described in #2002 

## Check List

### DIFF
--- a/src/app/core-ui/main-header/main-header.component.html
+++ b/src/app/core-ui/main-header/main-header.component.html
@@ -112,7 +112,7 @@
       <button
         (click)="taskService.toggleStartTask()"
         [color]="(taskService.currentTaskId$|async)? 'accent': 'primary'"
-        [matTooltip]="T.MH.TOGGLE_TRACK_TIME|translate"
+        [matTooltip]="(pomodoroService.isEnabled$|async)? '': T.MH.TOGGLE_TRACK_TIME|translate"
         class="play-btn mat-elevation-z3"
         mat-mini-fab
       >


johannesjo:
Thanks for opening this up! You can add new translation strings by editing the `en.json` and then executing `npm run int`

PR #2027, opened 2022-04-24
fix: move notes sidebar to left when rtl

zoli:
The margin is set regarding that the page is LTR so `margin-right` is set to `-40%`. This was true also when page was RTL and it caused showing the notes sidebar behind the tasks list. 
Changed the changing margin to margin-left when the layout is rtl to fix this.


johannesjo:
Thank you very much! :)

PR #2030, opened 2022-04-25
feat: respect first day of week in reports

zoli:
we can set the first day of week to what we want, but the value is not respected in worklog and quick history.


johannesjo:
This is great stuff again @zoli ! I really like how you write code and works nicely, too :) Thank you very much again!

PR #2061, opened 2022-05-03
fix: change gitlab icon to a transparent one

hugaleno:
# Description

I think that makes sense to use all integration icons as transparent ones.

What do you think @johannesjo 

Before:
change_gitlab_icon_before

After:

change_gitlab_icon_after

## Check List


johannesjo:
Love it! Thank you!!

PR #2067, opened 2022-05-05
fix: problem with empty project configuration with number #2066

### LINKED ISSUE #2066: Bug after gitea integration addition
@johannesjo  After Gitea addition there is a bug when trying to open project settings of an older project, because gitea configuration is empty in this case.  I think you work around that by doing the following:

```
// Unfortunately needed, to make sure we have no empty configs
          // TODO maybe think of a better solution for the defaults
          if (!this.issueIntegrationCfgs.JIRA) {
            this.issueIntegrationCfgs.JIRA = DEFAULT_JIRA_CFG;
          }
          if (!this.issueIntegrationCfgs.GITHUB) {
            this.issueIntegrationCfgs.GITHUB = DEFAULT_GITHUB_CFG;
          }
          if (!this.issueIntegrationCfgs.GITLAB) {
            this.issueIntegrationCfgs.GITLAB = DEFAULT_GITLAB_CFG;
          }
```
I could just add Gitea default here but I think the following would be better solution:

```
getIssueIntegrationCfg(key: IssueProviderKey): IssueIntegrationCfg {
    if (!(this.issueIntegrationCfgs as any)[key]) {
      return DEFAULT_ISSUE_PROVIDER_CFGS[key];
      //throw new Error('Invalid issue integration cfg');
    }
    return (this.issueIntegrationCfgs as any)[key];
  }
```

Instead of throwing an exception I think we should return a default configuration from DEFAULT_ISSUE_PROVIDER_CFGS provider array. That should have the same effect, what do you think?

Then we can remove the first code.

hugaleno:
# Description

Fix a problem with empty project configuration after gitea integration addition

## Issues Resolved

#2066 

## Check List

### DIFF
--- a/src/app/pages/project-settings-page/project-settings-page.component.ts
+++ b/src/app/pages/project-settings-page/project-settings-page.component.ts
@@ -20,18 +20,18 @@ import {
 import { Subscription } from 'rxjs';
 import { ProjectService } from '../../features/project/project.service';
 import { BASIC_PROJECT_CONFIG_FORM_CONFIG } from '../../features/project/project-form-cfg.const';
-import { ISSUE_PROVIDER_FORM_CFGS } from '../../features/issue/issue.const';
+import {
+  ISSUE_PROVIDER_FORM_CFGS,
+  DEFAULT_ISSUE_PROVIDER_CFGS,
+} from '../../features/issue/issue.const';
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you!

PR #2081, opened 2022-05-19
style(procrastination): Mirror Sync Icon

JasonLovrien:
Change sync icon rotation from clockwise to counter-clockwise. This way,
the icon rotates in the same direction that the arrows are pointing.

### DIFF
--- a/src/app/features/procrastination/procrastination.component.html
+++ b/src/app/features/procrastination/procrastination.component.html
@@ -12,8 +12,8 @@
 
         <div class="procrastination-graph">
           <div class="wrap">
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much for opening this up! Thinking about this, it might be more intuitive (at least in western cultures) to flip the icon and to rotate clockwise instead ("Avoiding the task" and "fear of failure" should be flipped as well then). What do you think? Could you maybe adjust the PR in this manner?

JasonLovrien:
That makes sense! It would match most other cyclical diagrams I've seen as well. I'll make that change!

PR #2090, opened 2022-05-28
Enable to set start of the day

### LINKED ISSUE #2035: Enable to set start of day
## Problem Statement
A calendar day is from 00:00:00 to 23:59:59 but a day for me is from for example 06:00:00 am to 02:00:00 am the next day until I go to sleep. But in super-productivity the day strictly ends at 23:59:59 and the timer, repeated tasks, etc gets restarted at that time. 
I want the 2 hour I work between 00:00 - 02:00 am on Thursday (by calendar) to get summed on the hours I worked on Wednesday not that part of it get to Thursday. I want If I have a repeated daily task That I managed to get it done on 01:00 am Thursday to again get added to my daily tasks on Thursday when I woke up Thursday morning.

My possible solution which I have got the idea from [Anki](https://apps.ankiweb.net/) is to have a setting which sets the time that a day starts for me and do the repeated tasks calculating daily time I work and etc based on that time.

zoli:
# Description

This implements enabling to set a custom hour for the start of the day. There is more detail about the feature in the issue.
I'm going to install superproductivity in my system based on these changes to test it for a while, but still I don't use some superproductivity features as daily usage So I want to ask for reviewing my changes more carefully.

## Issues Resolved

resolves #2035

## Check List

I'm not sure if this is a type of thing that should and could go in the README. Because the README doesn't go through about the specific settings. If you think its appropriate to add it in README I will be happy to do so.


johannesjo (approved):
Great Job as always! Thank you very much! Seems to work nicely as far I have tested.

The only improvement suggestion I have is that I'm tempted to say that `getWorklogStr` and `startOfNextDayDiff` might deserve to live their own service.  But no worries. I can also make this adjustment later on.

johannesjo:
As far as I can see these changes won't negatively affect the current behavior, so I would be ok with merging this as is and to test this out in the wild. Just one other thing I just thought of: We might want to adjust `migrateGlobalConfigState()` to also set a default value for the new config param.

zoli:
Thanks. I haven't seen any problem either couple of days that I'm using it based on this branch.

I will do the adjustments. Any suggestions for the path and the name of the service which `getWorklogStr` and `startOfNextDayDiff` should go there?

zoli:
> We might want to adjust `migrateGlobalConfigState()` to also set a default value for the new config param

Doesn't `_extendConfigDefaults` do this automatically [here](https://github.com/johannesjo/super-productivity/blob/master/src/app/features/config/migrate-global-config.util.ts#L110)?

zoli:
> Any suggestions for the path and the name of the service which `getWorklogStr` and `startOfNextDayDiff` should go there?

Maybe day.service or date.service with `todayStr` method?

johannesjo:
> Doesn't `_extendConfigDefaults` do this automatically [here](https://github.com/johannesjo/super-productivity/blob/master/src/app/features/config/migrate-global-config.util.ts#L110)?

Yes, you're right! But we need to increase the model version (e.g. by 0.001) for this to happen :)

johannesjo:
I am merging this! Thank you very much again! Sorry for the delay (I was on vacation the last week).

zoli:
I think its ready to merge.

PR #2091, opened 2022-05-28
Update Croatian translation

milotype:
- update "hr.json" file according to the "en.json" file


johannesjo:
Thank you very much! :)

PR #2237, opened 2022-09-10
Updated Croatian translation

milotype:
# Description
Croatian translation update according to current "en.json" file


johannesjo:
Thank you very much again!

PR #2248, opened 2022-09-21
fix(i18n): update chinese translation

shadowwinds:
# Description
‘订购’ means purchase, not order of tasks
update chinese translation

## Check List


johannesjo:
Thank you!

PR #2297, opened 2022-10-18
Update Croatian translation

milotype:
# Description

Update Croatian translation with minor corrections


johannesjo:
Thank youuuu!

PR #2305, opened 2022-10-22
improvement: Trigger focus task when reorder button is clicked

### LINKED ISSUE #2208: Request to Expand clickable area on a task to focus it
## Problem Statement
Hello, this application is very nicely made but i have a few gripes. 
First of all, i would love if i can select a task without actually going into the renaming field. 
Often times, the focus in the application is taken away from the task when i click settings or some other menu button. To bring focus back to the task and use keyboard shortcuts on it, i am currently forced to click a very narrow border around each task. Otherwise it just triggers the rename field. 

An easy solution for this would be to make it so that clicking the leftmost part of a task i.e. , the reorder button, brings back focus to the task without going into renaming. Currently, clicking the button immediately enters the task into reordering mode without checking if the used it actually moving their cursor or not. 

## Some additional pictures:
An Unfocused task
image

Result of Clicking even the empty space on the task name field. 
image
It goes into renaming. maybe you can reduce the region where rename is triggered to a bit more than the task names' lengths. 

The blue regions mark the areas where i can click into to focus task than doing some other action. 
image

fyi, the actual regions are a bit smaller because i couldn't draw thinner borders with my mouse haha. hope it was easy to understand.

davidzcode:
# Description

Now focus on task can be triggered when reorder button is clicked

## Issues Resolved

#2208 

## Check List

### DIFF
--- a/src/app/features/tasks/task/task.component.html
+++ b/src/app/features/tasks/task/task.component.html
@@ -83,6 +83,7 @@
 
     <div class="title-and-left-btns-wrapper">
       <div
+        (click)="focusSelf()"
         [class.handle-par]="!task.parentId"
         [class.handle-sub]="task.parentId"
         class="drag-handle"


johannesjo:
Thank you!

PR #2315, opened 2022-10-31
Update fa.json

majidabednew:
Translate some untranslated string specially about procrastinating that I am feeling right now :) I will update this part after I read those articles that you mentioned on application. Thanks a lot.
I think I gonna love it.


johannesjo:
Thank you very much!

PR #2399, opened 2023-01-01
Update Croatian translation

milotype:
# Description

Update and some corrections for Croatian translation


johannesjo:
Thank you very much!

PR #2401, opened 2023-01-02
Change time units

milotype:
# Description

Change time units and add non breaking spaces.
I'm not a programmer so please check if my changes are good.
 ...

### DIFF
--- a/electron/indicator.ts
+++ b/electron/indicator.ts
@@ -132,7 +132,7 @@ function createIndicatorStr(task: TaskCopy): string {
     //   timeStr += '/' + timeEstimateAsMin;
     // }
     //
-    // msg = title + ' | ' + timeStr + 'm ';
+    // msg = title + ' | ' + timeStr + ' min ';
     // return msg;
   }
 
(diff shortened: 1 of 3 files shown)


johannesjo:
Thanks for your efforts! I won't merge this however. The changes will break some of the UI in edge cases if the window is smaller, since the horizontal space required increases. Also the syntax is the same as the short syntax, which probably makes sense.

milotype:
@johannesjo 
I'm a bit dissapointed, that you don't want to implement my proposed changes …
I'm a typographer and translator. IMO, UI desing needs to addapt to the context/syntax … 

From a typographic point of view, values and units should generally be separated by a separator (a space, or preferably a non-breaking space).

Also, using false unit "m" (meters) instead of "min" (minutes) is simply wrong!

P.S. Of course it's your project, so you can do what you want. :-)

johannesjo:
@milotype 
Yes, I can understand you and your point is valid of course. I leave the issue open for discussion. Maybe we can revisit this if others are having problems with this as well. For me it's a practical decision where the costs (several potential followup issues and problems to consider, especially for mobile and smaller screens and the incongruence with the short syntax) outweigh the benefits (being more accessible to new users).

PR #2417, opened 2023-01-15
Update Croatian translation

milotype:
# Description

- changed some strings
- reverted the translated app's name to "Super Productivity"


johannesjo:
Thank you very much!

PR #2421, opened 2023-01-19
fix(i18n): update the portuguese translation

magalhaesm:
# Description

Portuguese translation update according to current "en.json" file


johannesjo:
Thank you!!!!

PR #2455, opened 2023-02-07
Update Croatian translation

milotype:
# Description

- Update Croatian translation with new strings
- Minor correction


johannesjo:
Thank you very much!

PR #2526, opened 2023-03-21
fix(i18n): some fixes for russian translate

ScrabyQ:
I made it so that the texts with colors look the same.
I also fixed a mistake in the "Markdown" translation (it shouldn't translate into Russian).


johannesjo:
thank you!

PR #2535, opened 2023-03-26
Correctly handle timezones in ical

ishanarora:
# Description

ICS format allows for timestamps to be in different timezone. This is particularly useful in defining recurring events in a daylight saving timezone. ical.js's default behaviour doesn't utilise timezone definitions included in the ICS string. This PR uses the workaround mentioned here: https://github.com/kewisch/ical.js/issues/455#issuecomment-706984325


johannesjo:
Thank you!!

PR #2542, opened 2023-03-29
Update it.json

tiziano-morgia:
Changed "OPEN_PROJECT"."DIALOG_TRACK_TIME"."POST_TIME" to "Pubblica Registro" because "Tempo di pubblicazione" has a different meaning


johannesjo:
Thank you very much!

PR #2571, opened 2023-04-09
Get issue url from issue itself. Fix #2543

### LINKED ISSUE #2543: Gitlab Integration: Open in Browser uses wrong URL if project id is used
### Your Environment

- Version used: snap  7.13.1
- Operating System and version: Ubuntu 22.04
- Desktop Environment: Xfce

### Expected Behavior

Open Browser on a Gitlab Issue should bring you to the issue

### Current Behavior

Open Browser brings you to a 404 page on Gitlab

1. Add a project with gitlab integration, making sure to just use the project id instead of the path
2. Add an issue
3. Right click open in browser, the url will use the project id instead of the project path which gitlab can't resolve

Yes

hugaleno:
# Description

Fix gitlab issue url from right click open in browser.

## Issues Resolved

Fix #2543

## Check List

### DIFF
--- a/src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts
+++ b/src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts
@@ -9,11 +9,7 @@ import { IssueData, SearchResultItem } from '../../issue.model';
 import { GitlabCfg } from './gitlab';
 import { GitlabIssue } from './gitlab-issue/gitlab-issue.model';
 import { truncate } from '../../../../util/truncate';
-import {
-  GITLAB_BASE_URL,
-  GITLAB_INITIAL_POLL_DELAY,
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you! The problem with that approach is that it will make an api request, just to display the correct url and that request will be made whenever a task is shown somewhere. Ideally we would have an approach that works offline.

hugaleno:
Yeah, right now we need that approach when using gitlab group/subgroup integration, because the issue url can't be determinated without knowing the name of the project. I was taking a look at gitlab integration open issues and gitlab api itself. I think we gonna need a different aproach to support projects/all/groups usecases. I will open a discussion about that, think is more apropriate.

johannesjo:
Thank you! Yes, makes sense to discuss it!

PR #2572, opened 2023-04-09
Updated "tomate" word to "pomodoro"

antonioanerao:
# Description

In Brazil, when we talk about a pomodoro timer we don't translate the word "pomodoro" to "tomate". To keep consistency I've changed the word "tomate" to "pomodoro" just as we have in the whole file


johannesjo:
:D thank you!

PR #2588, opened 2023-04-23
build: update node version in Dockerfile

kmantel:
# Description

Docker image builds have been broken since 3895c130b830ade27b649867a1c1867d3de7cc3f ([v7.11.6](https://github.com/johannesjo/super-productivity/actions/runs/2932989100)). Changing the version of node used in Dockerfile to 16 as in the other build workflows solves this problem.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 ### build ###
 
 # base image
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you!

Issue #2355, opened 2022-11-27
Constantly writing to IndexedDB when a task is running is bad for SSDs


RationalFragile:
### Your Environment

- Version used: 7.11.6 
- Operating System and version: Windows 10
- Desktop Environment: 
- Browser Name and version: 

### Expected Behavior
It should persist changes of course but not write them to disk every second. Something like once per minute when a task is running and on stop/start is more reasonable.

### Current Behavior
It constantly writes to files like "userdatadir\IndexedDB\file__0.indexeddb.blob\1\00" but only when a task is running. And because tasks run for hours and hours, process explorer is showing that superproductivity wrote more than 3 times what my chrome writes. (And I noticed my SSD lifespan shrunk from a projected 13 years to 10 years in less than 6 months, and superproductivity is the program that writes the most to disk, after "System" in Process Explorer. I moved it to my HDD but still, I think this should be changed.)

### Steps to Reproduce (for bugs)

1. Run something like Procmon (Process Monitor) and select only disk activity for process name "superProductivity"
2. Or open process explorer or task manager and highlight superproductivity after showing disk writes
3. Start a task and notice how it's constantly writing to disk
4. Stop task and the writing stops

### Can you reproduce this reliably?
Yes

### Error Log (Desktop only)

EDIT:
This can be clearly seen in the electron devtools, when a task is running, you get multiple "[Persistence] Save to DB" events every second.

johannesjo:
If you want to store your data and have it available even when crashes occur, you need to write it.  I don't see any way around this. One could introduce a throttleling mechanism, but that would raise a ton of complicated problems.

RationalFragile:
Thank you @johannesjo but I'm sorry, that's definitely not normal. It literally consumes ~25% of "active time" on my HDD, constantly, for hours, as long as task is running.

As I said in the issue, here is the sensible way to save data:
- On stop/start tasks
- On other operations like saving settings
- Every 1-5 minutes while a task is running. Even if it crashes, at worst you'll lose 1-5 minutes that were recorded.
- On quit.

No other program ever writes to disk every second for hours.

RationalFragile:
Just to make the point very clear, the app is saving 1.1GB every hour if I let it running, even with "automatic backups" disabled.

[...]

PR #2595, opened 2023-04-29
WIP: Change save tracked time approach. Issue #2355

hugaleno:
# Description

Change  tracked time approach to only persist on specific actions.
I'm not too familiar with angular ngrx store, so I would realy appreciate a review @johannesjo to make sure i'm going in the right way.

Atm, It's persisting time tracked only when pausing the current task/toggleCounters

## Issues Resolved

Addresses #2355 

## Check List


johannesjo:
Looks good so far! Some general remarks:

* We can use a dedicated effect to trigger the `persistTimeTracked` action which imho would feel cleaner

Apart from this, I think this is going in the right direction! Thank you!

thinktapper:
Hey all, sad to see this got stale after a year. Is there anything we can do to pick up this progress again?
I have limited ngrx experience but would be happy to jump in if @johannesjo 's suggestion to create a

> dedicated effect to trigger the persistTimeTracked action

is the next move 🏋️‍♂️

johannesjo:
Hey @hugaleno ! Thank you very much for outlining this! On an ngrx level this looks good. We wouldn't be strictly required to involve ngrx necessarily, but since it is there to help understanding the data flows of the app, it might be helpful. Have you tried the redux dev tools browser extension? If the logs there provide a clear picture of what is going on (with enough information, but not too much noise), then we are on the right track! :)

I am most curious, about how the final model will look like (e.g. changes to AppDataComplete, but also the data written on the disk). Maybe there are also opportunities to shift more data to the archive sync file to make regular sync faster? In my local data the project workStart and workEnd take up almost half the space. But of course that shouldn't be the focus or main goal here.

I think we also should have a mechanism to save the time tracking at least once every one or two minutes, even when it is not paused, to avoid loosing too much data if the app or system crashes. 

Not sure, if this is the input you were looking for. Please let me know if you have any other questions!

thinktapper:
Thanks, the input is helpful @johannesjo and yes, I find the Redux dev tools wonderful for following data flow and visualizing the store! I should have some time next week to start looking into this work

johannesjo:
A lot has changed since then. I have a good concept now to finally tackle this. Stay tuned!

PR #2644, opened 2023-06-05
Russian translation

BlackRedLight:
# Description

Some untranslated text was translated and some translation errors and inaccuracies were corrected. This changes were directed to correct frequent pages of the app, like scheduled tasks, settings, etc. Pages, where I saw this errors a lot. Also made some changes to lines, that caught my eye. Some of the errors were severe, like in sound settings "Volume" was translated like "Объём". Which means a volume of something, like in liters, not the loudness of the sound. And some just sounded unnatural.

## Issues Resolved

None


johannesjo:
Thank you very much!

PR #2651, opened 2023-06-17
Allow filter of issues

cbergmann:
# Description

You can filter the imported issues by any means provided by the API. For example you can add 'my_reaction_emoji=thumbsup' to only import issues that you reacted with 👍 .

## Issues Resolved

none

## Check List


src/app/features/issue/providers/gitlab/gitlab.const.ts
```diff
@@ -92,6 +93,16 @@ export const GITLAB_CONFIG_FORM: LimitedFormlyFieldConfig<GitlabCfg>[] = [
       ],
     },
   },
+  {
```
johannesjo, line 96: Since this is an optional feature, I'd like to see it further down the form.

johannesjo:
Thank you for adding this. Please see the comment above.

Issue #427, opened 2020-07-01
Re-create sub tasks for repeating tasks


level6sam:
### Your Environment

* Version used: snap
* Operating System and version: Ubuntu 18.04
* Desktop Environment:  Elementary OS 5.1.4
* Browser Name and version: 

### Expected Behavior

Create a main task for items that need to be carried out
Make the main task repeatable so that it appears the next day
Create subtasks under the main task
Schedule the tasks for the next day

### Current Behavior

Open up the app the day after scheduling a sub-task from a repeatable main task
Main task appears but there isn't a sub task anymore making the main task useless

### Steps to Reproduce (for bugs)

1. Create a main task
2. Add sub tasks
3. Make the main task repeatable and save
4. Wait the next day to find the sub tasks no longer appear under the main task

johannesjo:
Hey! Thanks for opening this up! This is actually a conscious decision, but I can see that this depends on the actual use case. I leave this open for discussion.

level6sam:
So my argument (for discussion) is that sub tasks should be a repeatable option because much of the work we carry out is mainly through schemas. With the ability to monitor the amount of time spent compared to how much time should be spent. It helps motivate a person who knows that these repeatable tasks are uninspiring and are more or less chores. Eventually the idea would be to have done it enough times for it to become a (healthy) habit it itself. Also, it is a pain to constantly re-order tasks which get mixed up in other one off tasks as its repeatable tasks usually follow a specific order in which they are completed.

Thanks

nicknovitski:
> Hey! Thanks for opening this up! This is actually a conscious decision, but I can see that this depends on the actual use case.

I can understand that.  What I imagine is, as part of the Edit Repeat Task Config window, you can optionally specify default sub-tasks which will be created when the task is automatically created, and completed when the task is automatically marked completed.  So that people can mix default sub-tasks with manually added ones, which can retain the previous behavior.

My use case is: I want a daily reminder to empty my inboxes, and I want to work on each different inbox (for example email, slack, github) one at a time, usually in the same order.   Essentially, I'd like to be able to create repeating tasks that function as checklists.

johannesjo:
I am open for experimenting with this, if someone is willing to put work into it :)

aninternetuserx:
In case it's useful for someone, since this feature is not available right now, the workaround I found is to use the "comment" section of the repeating task to add the little steps of the task and they repeat with it :)

podivilov:
[...]

PR #2659, opened 2023-06-29
feat: add repeat subtasks for repeating tasks #427

StartAGarden:
# Description

The repeating tasks are updated so they include their sub-tasks.
Example Use Case: The user wants to add a task for a daily "Morning Routine" including several sub-tasks (steps of the routine), without having to re-add the sub-tasks every day.

_There is no way to individually decide for certain sub tasks to be repeatable or not in this implementation. All sub tasks of a repeatable parent task are automatically repeatable, too._

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/427

## Check List

- [ ] When a task is repeatable and a sub task is added, the sub task automatically gets repeatable, too
- [ ] When a task with subtasks is made repeatable, all sub tasks are made repeatable, too
- [ ] BUGFIX: fixed taskCfg not updating when repeatable task gets updated (tested my changing name of repeatable task and changing to next day -> was the old name)
- [ ] only main repeatable tasks are shown in the "Scheduled" Tab, subs automatically get removed when the parent tasks repeatable gets removed

### DIFF
--- a/src/app/features/task-repeat-cfg/store/task-repeat-cfg.effects.ts
+++ b/src/app/features/task-repeat-cfg/store/task-repeat-cfg.effects.ts
@@ -5,6 +5,7 @@ import {
   delay,
   filter,
   first,
+  map,
   mergeMap,
   take,
   tap,
@@ -28,9 +29,10 @@ import { TaskRepeatCfgService } from '../task-repeat-cfg.service';
 import {
   DEFAULT_TASK_REPEAT_CFG,
   TaskRepeatCfg,
+  TaskRepeatCfgCopy,
   TaskRepeatCfgState,
 } from '../task-repeat-cfg.model';
-import { forkJoin, from, merge, of } from 'rxjs';
+import { Observable, forkJoin, from, merge, of } from 'rxjs';
 import { setActiveWorkContext } from '../../work-context/store/work-context.actions';
 import { SyncTriggerService } from '../../../imex/sync/sync-trigger.service';
 import { SyncProviderService } from '../../../imex/sync/sync-provider.service';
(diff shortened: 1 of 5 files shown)


src/app/features/task-repeat-cfg/task-repeat-cfg.service.ts
```diff
@@ -271,4 +319,23 @@ export class TaskRepeatCfgService {
...
+  private _getSubTaskRepeatTemplate(
+    taskRepeatCfg: TaskRepeatCfg,
+    aParentId: string,
+  ): Task {
+    const isAddToTodayAsFallback =
+      !taskRepeatCfg.projectId && !taskRepeatCfg.tagIds.length;
+    return this._taskService.createNewTaskWithDefaults({
+      title: taskRepeatCfg.title,
+      additional: {
+        repeatCfgId: taskRepeatCfg.id,
```
johannesjo, line 332: I would just leave out the repeatCfgId part

src/app/features/task-repeat-cfg/store/task-repeat-cfg.effects.ts
```diff
@@ -58,6 +66,131 @@ export class TaskRepeatCfgEffects {
...
 
+  /**
+   * Updates the repeatCfg of a task, if the task was updated.
+   */
+  updateRepeatCfgWhenTaskUpdates: any = createEffect(
+    () =>
+      this._actions$.pipe(
+        ofType(updateTask),
+        tap(async (aAction) => {
+          const allTasks = await this._taskService.allTasks$.pipe(first()).toPromise();
```
johannesjo, line 77: Better  to directly use `taskService.getByIdOnce$` here.

src/app/features/task-repeat-cfg/store/task-repeat-cfg.effects.ts
```diff
@@ -58,6 +66,131 @@ export class TaskRepeatCfgEffects {
     { dispatch: false },
   );
 
+  /**
+   * Updates the repeatCfg of a task, if the task was updated.
+   */
+  updateRepeatCfgWhenTaskUpdates: any = createEffect(
```
johannesjo, line 72: Do we really need this (honestly I am unsure :D)? Are there potential edge cases and problems that this might cause?

src/app/features/task-repeat-cfg/store/task-repeat-cfg.effects.ts
```diff
@@ -58,6 +66,131 @@ export class TaskRepeatCfgEffects {
...
+        }),
+      ),
+    { dispatch: false }, // Question: What exactly does this do?
+  );
+
+  /**
+   * When adding a sub task, this function checks if the parent is a repeatable task and therefore the sub-task also has to be.
+   * If that is the case, a repeat-cfg gets added for each subtask, too.
+   */
+  addTaskRepeatCfgForSubTask: any = createEffect(
```
johannesjo, line 150: Imho subtasks should not be repeatable tasks, but just subtasks of a task with a repeatable config.

johannesjo:
Thank you very much! The code looks very clean! Much appreciated!

I am a bit pressed for time, so I wasn't able to provide a complete feedback yet. Sending the incomplete feedback anyway since I am not sure I will find more time before next Friday.

src/app/features/task-repeat-cfg/store/task-repeat-cfg.effects.ts
StartAGarden: Not sure about edge cases, but without this you have no way of changing contents of a repeatable task as far as I was able to see.
What I did: I tried to change the title of a repeatable task, but since each day the new tasks are generated based on the repeatCfg (where the title was not updated), the old title would appear.

src/app/features/task-repeat-cfg/store/task-repeat-cfg.effects.ts
StartAGarden: Since finished tasks get deleted (at least that's what it seemed like when I tried to observe how the app works?) and I am not confident enough at this point to try and change things on such a fundamental level, I based this approach on the fact that repeatCfgs are persisted and can be used to store the information of the sub tasks. I basically mirrored that structure with the repeatCfs:

All Tasks (sub + parent) are stored as tasks (all together) in the store (only difference is if they have subtask-ids or a parent-Id or not)
All repeatCfgs (sub + parent) are stored as repeatCfs (all together) in the store (only difference is that they have a parent id or not)

Since all information of a repeatable task gets generated from a repeatable config, I only see 2 ways of storing the information for the sub tasks:
- storing all information about all subtasks in the parent repeat config (which would go against the structure that is currently implemented for normal tasks and therefore would feel like bad practise to me, but that might just be my view)
- storing the information about sub tasks as a subtask-repeatable-config

Definitely let me know if I missed something.

src/app/features/task-repeat-cfg/store/task-repeat-cfg.effects.ts
```diff
@@ -106,24 +245,47 @@ export class TaskRepeatCfgEffects {
...
     ),
   );
 
   removeConfigIdFromTaskArchiveTasks$: any = createEffect(
     () =>
       this._actions$.pipe(
         ofType(deleteTaskRepeatCfg),
-        tap(({ id }) => {
+        tap(async ({ id }) => {
+          const subTasks = await this._taskRepeatCfgService
```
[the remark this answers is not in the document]
StartAGarden: Please see comment below

src/app/features/task-repeat-cfg/task-repeat-cfg.service.ts
StartAGarden: Please see comment below

StartAGarden:
> Thank you very much! The code looks very clean! Much appreciated!
> 
> I am a bit pressed for time, so I wasn't able to provide a complete feedback yet. Sending the incomplete feedback anyway since I am not sure I will find more time before next Friday.

Just take your time! I added my thoughts to your comments already so you can have a better insight into my thought progress for the review.

src/app/features/task-repeat-cfg/store/task-repeat-cfg.effects.ts
johannesjo: Currently you can change existing repeatable task instances by editing the repeatable config. If you edit the title there, the app asks you if you want to update all existing tasks. The idea behind it is, that you might want to have different notes and even tags for different instances. But I can see how it might be more intuitive to ditch this behavior and do it like this. Let's try it out!

src/app/features/task-repeat-cfg/store/task-repeat-cfg.effects.ts
johannesjo: Afaics subtasks don't need to "know" that they are a child task of a repeatable task and I also think different parent repeatable tasks should be allowed to have different sub tasks. To solve this I would not reference "real" sub tasks but rather provide a simple input that allows to create a list of sub tasks with no more information than maybe the title. 

Using real "subtasks" as reference point creates all sorts of messy problems, when moving them around or deleting them or their parent.

johannesjo:
I think I understand now the main misunderstanding :D 

The current approach for repeatable tasks is to take the repeatCfg as template to create new instances. Existing instances are not touched unless the the repeatCfg is updated by using the dedicated dialog for it. This allows an instance to be used relatively independent from others, e.g. I could create different sub tasks or notes every time. 

The changes you are providing take a different approach by making changes to a task directly change the repeatCfg and by making changes to sub tasks also reflect on new instances. This might be more intuitive to new users (and thus ultimately desirable, but 
 I am not a 100% sure yet), but possibly confusing to existing ones. I am also a bit afraid that it might lead to non trivial problems further down the road, e.g. when moving sub tasks to other parent tasks, deleting them and also when dealing with archive tasks.

So I think I might have to test this a bit more to understand all the implications of this changes. Or alternatively we can go with the "simpler" change that I proposed in the comment above and stick with the current approach of doing things. What do you think?

GadiZimerman:
@CodiumAI-Agent /review

StartAGarden:
> I think I understand now my main misunderstanding :D
> 
> The current approach for repeatable tasks is to take the repeatCfg as template to create new instances. Existing instances are not touched unless the the repeatCfg is updated by using the dedicated dialog for it. This allows an instance to be used relatively independent from others, e.g. I could create different sub tasks or notes every time.
> 
> The changes you are providing take a different approach by making changes to a task directly change the repeatCfg and by making changes to sub tasks also reflect on new instances. This might be more intuitive to new users (and thus ultimately desirable, but I am not a 100% sure yet), but possibly confusing to existing ones. I am also a bit afraid that it might lead to non trivial problems further down the road, e.g. when moving sub tasks to other parent tasks, deleting them and also when dealing with archive tasks.
> 
> So I think I might have to test this a bit more to understand all the implications of this changes. Or alternatively we can go with the "simpler" change that I proposed in the comment above and stick with the current approach of doing things. What do you think?

First of all: Sorry for the late answer! Once I have a quiet minute I will definitely look into your suggested simpler version, as I generally agree with trying to avoid possible problems or confusion. I am just really pressed for time these days so I am not quite sure when I will be able to pick it up again, but I hope to find time soon and I will contact you again, if I have further questions :)

Olmo-Gutierrez:
How is this doing?

StartAGarden:
> How is this doing?

Since there has been a lot of things happening for me, I don't have time at all to work on this. If anyone wants to pick this up, feel free. If not I hope to maybe get to it at the end of the year, but currently I don't have any capacity to work on it.... :(

johannesjo:
@StartAGarden no worries! Thank you very much for your effort!

Any volunteers to pick this up?

Jagdfalke:
Hi @StartAGarden, how are you doing?

mattastic485:
Still relevant.

Cyber-Syntax:
Still relevant.

suraj-mandal:
Saw through this feed, and I am picking it up. Have started working on this

novikov1337danil:
> Saw through this feed, and I am picking it up. Have started working on this

A new PR has recently been opened. Take a look https://github.com/johannesjo/super-productivity/pull/5068

Tosko4:
@johannesjo this one can be closed, right? As it was delivered in #5068

johannesjo:
Yes! Thanks @Tosko4 !

PR #2695, opened 2023-07-17
fix: create repeat tasks if missed in the days before #2694

### LINKED ISSUE #2694: Missing repeating task instances if the app isn't open on the date of repeat
### Your Environment

- Version used: 7.13.2-1 (AUR)
- Operating System and version: Archlinux (rolling)
- Desktop Environment: KDE
- Browser Name and version: standalone version

The bug also exists on the Android app.

### Expected Behavior
Two of my repeating tasks failed to create instances recently.

Task A is set to repeat on 1st of every month. I did not open Super Productivity on 1st and 2nd July. On 3rd July I expected to see the July instance of Task A, but it was missing. I still see the repeat config for Task A under "Scheduled > Repeated tasks".

Similarly, Task B is set to repeat every Friday. I did not open Super Productivity this Friday and Saturday. On Sunday I expected to see this week's instance of Task B, but it was missing. I still see the repeat config for Task B under "Scheduled > Repeated tasks".

For each of these repeating tasks I expected a task instance to be created and shown when Super Productivity was next opened.

ishanarora:
# Description

`selectTaskRepeatCfgsDueOnDay` now selects `taskRepeatCfgs` that were due on any day after `lastTaskCreation` until `dayDate`

## Issues Resolved

Fixes #2694 

## Check List

### DIFF
--- a/src/app/features/task-repeat-cfg/store/task-repeat-cfg.reducer.ts
+++ b/src/app/features/task-repeat-cfg/store/task-repeat-cfg.reducer.ts
@@ -96,10 +96,9 @@ export const selectTaskRepeatCfgsDueOnDay = createSelector(
           return false;
         }
 
-        let dateToCheckDate = new Date(dateToCheckTimestamp);
-        let numberOfDatesToCheck = getDiffInDays(
+        const numberOfDatesToCheck = getDiffInDays(
           new Date(taskRepeatCfg.lastTaskCreation),
-          dateToCheckDate,
+          new Date(dateToCheckTimestamp),
         );
 
         switch (taskRepeatCfg.repeatCycle) {
@@ -111,7 +110,11 @@ export const selectTaskRepeatCfgsDueOnDay = createSelector(
               throw new Error('Invalid repeatEvery value given for DAILY');
             }
             const startDateDate = new Date(taskRepeatCfg.startDate);
-            while (numberOfDatesToCheck > 0) {
+            for (let i = 0; i < numberOfDatesToCheck; i++) {
+              const dateToCheckDate = new Date(
+                // prettier-ignore
+                dateToCheckTimestamp - (i * 24 * 60 * 60 * 1000),
+              );
               const diffInDays = getDiffInDays(startDateDate, dateToCheckDate);
 
               if (
@@ -120,9 +123,6 @@ export const selectTaskRepeatCfgsDueOnDay = createSelector(
                 diffInDays % taskRepeatCfg.repeatEvery === 0
               )
                 return true;
-
-              dateToCheckDate.setDate(dateToCheckDate.getDate() - 1);
-              numberOfDatesToCheck -= 1;
             }
             return false;
           }
@@ -136,7 +136,11 @@ export const selectTaskRepeatCfgsDueOnDay = createSelector(
             }
             const startDateDate = new Date(taskRepeatCfg.startDate);
 
-            while (numberOfDatesToCheck > 0) {
+            for (let i = 0; i < numberOfDatesToCheck; i++) {
+              const dateToCheckDate = new Date(
+                // prettier-ignore
+                dateToCheckTimestamp - (i * 24 * 60 * 60 * 1000),
+              );
               const todayDay = dateToCheckDate.getDay();
               const todayDayStr: keyof TaskRepeatCfg = TASK_REPEAT_WEEKDAY_MAP[todayDay];
               const diffInWeeks = getDiffInWeeks(startDateDate, dateToCheckDate);
@@ -148,9 +152,6 @@ export const selectTaskRepeatCfgsDueOnDay = createSelector(
                 taskRepeatCfg[todayDayStr]
               )
                 return true;
-
-              dateToCheckDate.setDate(dateToCheckDate.getDate() - 1);
-              numberOfDatesToCheck -= 1;
             }
             return false;
           }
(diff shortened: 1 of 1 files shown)


ishanarora:
@johannesjo this pull request proposes to change the behaviour of `selectTaskRepeatCfgsDueOnDay`. The existing behaviour is to return only the tasks which are due to repeat on `dayDate`. The proposed behaviour is to check this for every date after `lastTaskCreation` date until `dayDate`. This would make sure to create a repeatable task instance if it was due to be created on any previous day, but was not created for any reason (such as because Super Productivity was not run on that date). Please let me know if you agree with the proposed change.

I see the current behaviour of `selectTaskRepeatCfgsDueOnDay` is thoroughly tested. With this change the test matrix should be expanded to include different `lastTaskCreation` states.

johannesjo:
Thanks for the PR. There are two things we should consider:
1. Ideally the creation date should be set to the original due day so people can spot for which day a task was intended (the value is currently set here: https://github.com/johannesjo/super-productivity/blob/5d25253912386e35f752a6c0e501453c55b847e2/src/app/features/task-repeat-cfg/task-repeat-cfg.service.ts#L211-L217).
2. We should prompt the user in case there are many tasks to be created (e.g. a daily task when the users wasn't online for a couple of days) to avoid that users have manually delete lots of tasks.

ishanarora:
Thanks for the comments @johannesjo.

> 1. Ideally the creation date should be set to the original due day so people can spot for which day a task was intended.

That's fair enough.

> 2. We should prompt the user in case there are many tasks to be created (e.g. a daily task when the users wasn't online for a couple of days) to avoid that users have manually delete lots of tasks.

It seems that in normal daily usage the invariant is that there is only ever one active (not completed) instance of a repeating task config. This is because at midnight all daily repeating tasks of the previous day automatically get marked done. Keeping in line with this invariant, in your example we could either:
1. create just one task for the last missed repeating instance, or
2. we could create all missed instances and mark as completed all task instances except the latest one.

The PR implements the former behaviour.

The latter behaviour is effectively the same as if the app was left running unattended for days. I believe the easiest user experience would be if we stick to this behaviour without prompting the user. Do you agree?

johannesjo:
Yes you're right! Creating just the last missed instance is probably the best solution :)

Could you have a look at the failing unit tests?

johannesjo:
ping @ishanarora  :)

ishanarora:
Apologies, I haven't had the time to take this up recently.

It seems the invariant I mentioned above no longer holds since cbc0ff4c3. In light of that I would be happy to change the behaviour of this PR to create all missed instances with the creation date set to the original due day for each. Agreed?

Since this PR is more of a bugfix, I believe the feature to prompt the user in case there are many tasks to be created can be a separate PR.

> Thanks for the PR. There are two things we should consider:
> 
>     1. Ideally the creation date should be set to the original due day so people can spot for which day a task was intended (the value is currently set here: https://github.com/johannesjo/super-productivity/blob/5d25253912386e35f752a6c0e501453c55b847e2/src/app/features/task-repeat-cfg/task-repeat-cfg.service.ts#L211-L217
>        ).
> 
>     2. We should prompt the user in case there are many tasks to be created (e.g. a daily task when the users wasn't online for a couple of days) to avoid that users have manually delete lots of tasks.

johannesjo:
No worries!

> It seems the invariant I mentioned above no longer holds since [cbc0ff4](https://github.com/johannesjo/super-productivity/commit/cbc0ff4c35cdbd0ef487da83e10f64580d07594b). In light of that I would be happy to change the behaviour of this PR to create all missed instances with the creation date set to the original due day for each. Agreed?

Also a possibility, but we should limit this to maybe the last 5(?) instances to avoid a task explosion. Like this we also don't need the prompt.

johannesjo:
After further discussing this I think that the best approach might be to provide a new config to each repeatable tasks that lets the user decide if missed instances should be created, or not. So users can choose themselves. Something like a daily reminder shouldn't be recreated for every missed day, but something like "do your monthly tax form" definitely should. 

Are you still interested in working on this?

ishanarora:
I think that's a good idea. Perhaps we can add configuration options for:
1. An optional limit on the total number of pending task instances.
2. Once the limit is hit then:
    1. should the earlier tasks be marked as done (similar to our earlier behaviour),
    2.  or should the new tasks be queued for creation once older tasks are completed by the user,
    3. or should the potential new task instances be lost in the ether, or perhaps they can be created and moved to backlog, or marked as done.

I'm afraid the work ahead is fairly outside my area of expertise. I don't know Angular, and the little TS I know I learned for small PRs like this one. We could close this PR since it will need a rewrite for the changed requirements.

Also I've come across a couple of alternative approches in the wild:

1. In [taskwarrior](https://taskwarrior.org/docs/using_dates/) tasks may optionally have `until` date, after which the task instance would self-destruct. We could also borrow taskwarrior's notion of `due` date.

6. [topydo](https://c306.net/t/topydo-docs/#Recurrence) has task recurrence duration with two modes:
    1. normal mode: periodic start dates (similar to SupProd's periodic creation dates)
    5. strict mode: the next instance is created after the recurrence duration starting from the current task's due date. If the due date for the current task is changed then the creation date of the subsequent task instance also change.

johannesjo:
@ishanarora Thank you very much for your input! This is very helpful!

I can try to pick this up then, if it is alright with you, though I don't know when I will get to it... :)

ishanarora:
@johannesjo yes, of course! I that case may I suggest merging this pull request as a hotfix for #2694 in the meantime. I believe #2694 is a critical bug e.g. it caused a missing annual income tax filing task because of the annual recurrence date landed on a Sunday.

The PR still needs updating the unit tests for `selectTaskRepeatCfgsDueOnDay`, see https://github.com/johannesjo/super-productivity/pull/2695#issuecomment-1639180968. Do you think you could take a jab at it? Thanks!

johannesjo:
I think that has been addressed. Thank you once more @ishanarora

PR #2702, opened 2023-07-27
Fix wrong translation

rklec:
# Description

This is wrongly translated as "should be lower than", which results in funny error messages about "should be lower than 1" when you enter 0:
grafik

Now translated it as "should not be lower than" similar to the max one, so than 1 in our example is still valid. "should be larger than or equal" would also be possible, but I deemed this to be unnecessary long and wordy.

## Issues Resolved
N/A


johannesjo:
Dankeeeee!

PR #2820, opened 2023-10-16
Translating from en.json to pt.json. Not everything but the necessary

vinycalves:
# Description

Translating only what is necessary from the application so far, I use en.json as the basis for this.

## Issues Resolved

none

## Check List

- none


johannesjo:
Thank you very much!

PR #2831, opened 2023-10-25
fix: focus after marking last subtask done (#2795)

### LINKED ISSUE #2795: Task list scrolls to the top when finishing subtasks
Something like this was reportet before but marked as stale in #2240. However, it massively disrupts my workflow, so here is a proper issue for it.

### Your Environment

- Version used: 7.14.3 amd64

### Expected Behavior

Clicking on the checkmark of a subtask marks it as done and just that.

### Current Behavior

The subtask is marked as done. In addition, the task list scrolls to the top

1. Move a few tasks with subtask into a task list
2. Scroll to the bottom
3. Mark one of the subtasks as done
4. SP will scroll to the top

Yes, it happens every time.

### Error Log (Desktop only)

No relevant output.

miqh:
# Description

Hello there.

These changes should help resolve how the application focuses the first task of a task list if the last subtask of another task is marked as done via the tick button. The resulting behaviour should put it in line with how the focus travels when resolving the subtask using a keyboard shortcut instead. 

I've also added an E2E test to cover this particular scenario. I'm not all too familiar with how this works with Nightwatch, but I gave it my best shot.

As an aside, there's another usability niggle (opinionated) I've noticed, but I don't have any good ideas for how to improve it.

If one marks a task with subtasks as done, then the focus doesn't move to the next sibling task in the same list. Underneath the hood, the application tries to shift the focus to the first subtask. However, because of how that entire task will end up in the completed pile, the focus disappears altogether, creating some disorientation.

I think the above is exacerbated by how the default behaviour of keyboard scrolling down a task list will automatically move into subtask items from parent task items. As long as this latter behaviour is required, I can't see an easy intervention.

In any case, happy to amend whatever.

demo1

## Issues Resolved

#2795

## Check List

- [ ] New functionality includes testing. (N/A)
- [ ] New functionality has been documented in the README if applicable. (N/A)

### DIFF
--- a/e2e/src/work-view.e2e.ts
+++ b/e2e/src/work-view.e2e.ts
@@ -114,4 +114,18 @@ module.exports = {
 
         .assert.containsText(TASK + ':nth-child(1)', '1 test task hihi')
         .end(),
+
+  'should focus previous subtask when marking last subtask done': (browser: NBrowser) =>
+    browser
+      .loadAppAndClickAwayWelcomeDialog(WORK_VIEW_URL)
+      .addTask('task1')
+      .addTask('task2')
+      .setValue('task:last-child', 'a')
+      .keys(['task3', browser.Keys.ENTER])
+      .setValue('[listid="SUB"] task:nth-child(1)', 'a')
+      .keys(['task4', browser.Keys.ENTER])
+      .moveToElement('[listid="SUB"] task:nth-child(2)', 10, 30)
+      .click('.task-done-btn')
+      .assert.containsText(':focus', 'task3')
+      .end(),
 };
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! I have a look tomorrow. Sorry for the delay. I was out of town for a couple of weeks... :)

johannesjo:
Thank you very much! It's absolutely great that you also added e2e tests. Much appreciated!!

About the niggle:
Maybe we can open up a new issue for this? Ideally I'd think that the previous or next main task or one of its sub tasks should be focused. Maybe there is some way to check if the current task is a parent and to never focus its sub tasks when it is marked as done?

miqh:
@johannesjo, you're welcome. :+1:

It's been a while since I originally looked at the code to prepare this fix, so I'm not able to advise on feasibility off the top of my head.

However, it'd only seem reasonable (and again, I express this subjectively) for the navigation focus to move to the next available task, or task group, if a currently focused task group is marked as done, instead of the first subtask of the latter.

PR #2867, opened 2023-11-24
French -- add traduction string to MISC& SOUND + correction 

Profxmas:
# Description

French -- 

I added traduction for mission strings in MISC& SOUND
I corrected in rounding time the translation, since two differents strings had the same traduction (3 times, for 5, 15 & 30 minutes)


Profxmas:
i made an error

PR #2872, opened 2023-11-25
Translation & add - Gitea in french

Profxmas:
# Description

Translation & add - Focus Mode in french
Cohesion change with github and gitlab


johannesjo:
Thank you very much!

PR #2930, opened 2023-12-20
Disable spellcheck in Add Task Bar Component

### LINKED ISSUE #2874: Disable spell checking
Problem Statement:

When I work with multiple languages, everything that doesn't correspond to the standard language is always highlighted in red, which is quite unpleasant and distracting while I'm writing.

Possible Solution:

Add a button to turn on and off

 Additional context:

I tried with bash-scripting, but that isn't what I am looking for. I would like a simple and easy way to turn it off without changing all applications at ones. Please Help:/

Screenshot from 2023-11-25 12-52-03

jordigcs:
Closes #2874

Disables spellchecking in Add Task Bar Component, fixes issues with tags and non-english words being highlighted as wrong.


johannesjo:
Oh that's cool! Wasn't aware that the change is this simple :)

Thank you very much!!!

Issue #2660, opened 2023-06-30
Simple timers: Show title on hover


q-wertz:
## Problem Statement
Currently one can only distinguish the simple timers by the symbol. When having quite some it might get hard to remember from the symbol it's exact meaning (there is not always a fitting symbol)

## :grey_question: Possible Solution
Show the `Title` of the timer when hovering over the icon.

prabhjotlamba1655:
HI @johannesjo, I would like to work on this issue.

johannesjo:
That's great to hear! Let me know if you have any questions!

q-wertz:
:robot: beepboop

rayolasr:
Hello @johannesjo ! I would like to start working on this issue. It would be my first github contribution :)
If I understand well, we are talking about this buttons, right?:
![superp-screenshot](https://github.com/johannesjo/super-productivity/assets/26779787/d1762f4f-87a1-47cf-a123-9b9a757bcb15)

q-wertz:
Exactly :+1:

rayolasr:
Hello @johannesjo!
I've already got the changes but don't know what to do next, first time contributing on GitHub xD

johannesjo:
@rayolasr does this help:
https://docs.github.com/articles/creating-a-pull-request
?

rayolasr:
Sure! I just thought I needed the issue to be assigned to me before doing a pull request.

johannesjo:
> Sure! I just thought I needed the issue to be assigned to me before doing a pull request.

This is not necessary, but thanks for the hint :)

q-wertz:
I think this was closed in #2966

PR #2966, opened 2024-01-03
fix: show simple counter title on hover #2660

rayolasr:
Show the title of the simple counter when hovering over the icon.


johannesjo:
Thank you!!

PR #3038, opened 2024-02-16
Added a new icon for macOS

solarlime:
# Description

Added a more appropriate icon file for macOS. I checked on my Mac; building and installing are correct.

## Issues Resolved

[This one](https://github.com/johannesjo/super-productivity/issues/3033)


johannesjo:
Thank you very much! Looks great! Current icns is 512x512px. So probably we should stick to this?

solarlime:
You don't need to fit this icns everywhere, as I think. It has icons from 16x16 for non-Retina displays to 512x512@2x for Retina displays (1024x1024). I personally don't have Windows or Linux installed, so I cannot check if it looks good there too. Maybe the existing one is a better choice for other systems. But you can check if you can; here is .svg:

johannesjo:
The icns in the PR is only 256x256. This might be to small?

solarlime:
> The icns in the PR you provided is only 256x256. This might be to small?

I'll check it out

solarlime:
Checked, found nothing criminal. However, I recreated icns file in a different way.

johannesjo:
Thank you for the update and also the universal build suggestion! Will check it out next week :)

PR #3212, opened 2024-06-11
Update translations for zh

ZaphireLiu:
- Added some missing translations.
- Modified a translation that was slightly too long.
- Corrected an incorrect translation.

Preview the messages on https://inlang.com/github.com/ZaphireLiu/super-productivity


johannesjo:
Thank you very much!

PR #3216, opened 2024-06-12
feat: Add current keyboard shortcuts to app menu's #3063

### LINKED ISSUE #3063: Add current keyboard shortcuts to app menu's
## Problem Statement

I always forget the keyboard shortcuts that I have assigned into the app. Currently the only way to see them is to go into settings and is pretty annoying.

One possible solution would be to add the keyboard shortcuts in the menu as most of windows applications do
Current right click menu:
image

I would like to have something similar to this:
image
With the shortcuts on the right of the command

jere0500:
I have implemented the feature would like to get initial feedback. 

# Description
Added keybind previews to the main task context menu and multiple tooltips in the main task window. 
The keybindings in the context menu are aligned to the right and have a lower transparency. 
Currently all keybindings are surrounded by ```[]```.

ContextMenu

Nothing will be visible when the key is unbound.

Lightmode
Keycombo in tooltip

Right now the design has some issues, like the arrow being on the right. So I would welcome any ideas for improving the design.

moveToProject

## Issues Resolved
#3063

## Check List for ideas and improvements
- [ ] add to more tooltips and context windows (would be happy for every finding)
- [ ] add a config option to hide the keybindings completely
- [ ] maybe show all keys as uppercase: so ```[K]``` and ```[Shift+K]``` instead of ```[k]``` and ```[Shift+K]```.

### DIFF
--- a/src/app/core-ui/main-header/main-header.component.html
+++ b/src/app/core-ui/main-header/main-header.component.html
@@ -44,7 +44,7 @@
   <nav class="action-nav">
     <button
       (click)="layoutService.showAddTaskBar()"
-      [matTooltip]="T.MH.ADD_NEW_TASK|translate"
+      matTooltip="{{T.MH.ADD_NEW_TASK|translate}} {{_kb.addNewTask?'['+_kb.addNewTask+']':''}}"
       class="hide-xs tour-addBtn"
       mat-icon-button
     >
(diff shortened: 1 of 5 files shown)


johannesjo:
Thank you very much! Works well i think!

I am a bit hesitant about showing the shortcuts for the task component as this will potentially will impact performance negatively, since the code is executed and rendered for every instance. This is fine for all stuff inside ng-template (for the mat-menu stuff), but outside of it, this can be problematic.

PR #3247, opened 2024-07-04
fix: sync show gitlab issue detail #2115

### LINKED ISSUE #2115: Gitlab integration issue
Hi,
First of all congratulations for this great software.
I have a problem with integration with Gitlab. Fetching issues is working, but there is a problem with issue data from Gitlab not recovering it. 
As you can see data not recovering for Gitlab section in the issue, and an error appears below: 404 Project not found.
But issues are being imported, so I don't know what is the problem. If you could fix this it would be great. Thanks again for this great application.

image

hosseinkhaledi:
# Description

Use gitlab issue id instead of iid

## Issues Resolved

#2115 

## Check List


johannesjo:
Thanks you very much for providing this PR! There is a merge conflict. Can you maybe take a look at it?

lazybobcat:
@hosseinkhaledi May I ask why this change, especially in the case of getting the issue details?

The GitLab API URL without the `iid` (`GET /issues/:id`) is reserved for administrator. The other API URL (`GET /projects/:id/issues/:issue_iid`) requires the `iid`.

We maybe could also use `issue._links.self` that contains the URL (it's used in `_getIssueComments$`)?

I can implement a fix, but since I'm not familiar with this project nor Gitlab API, I don't wanna break anything.

PR #3256, opened 2024-07-10
Update translations for Korean

lyuha:
# Description

- fix some word spacings
- match some words accordings to context


johannesjo:
Thank you very much for this!!

PR #3348, opened 2024-08-14
feat(integration): import github labels  as tags

### LINKED ISSUE #2866: [Feature] Import GitHub Labels as tags
First of all, thanks for the great app, I've been using it for some time for time management at work and it's been great! GitHub integration for a personal project also already works really well, but this request is something that would make it _even better_!

## Problem Statement

With imported GitHub issues, I get all the interesting info related to the Issues themselves. However, I lose the abitlity to sort them by label inside Super Productivity.

It would be great for me if you could (as another checkable setting) set the GitHub labels as tags for the individual tasks to sort by these afterwards in Super Productivity as well. They are obviously queried in the api, as they are shown in the additional infos, so I imagine it wouldn't be that big of a change to also add those as tags.

Alternatively, I could go back to GitHub for sorting and thus sort my issues by hand or add the labels manually by editing each task. This is not a massive problem, it's more of a convenience request and the question, if this functionality already exists and I just haven't found it...

I'm not aware of issues or PR's relating to this problem. I suppose some work would have to be done in src/app/features/tasks/ and src/app/features/issue/providers/github/ (here, the labels are queried as issue?.labels (in template html github-issue/github-issue-content component), but I'm not sure, how one could go about handing this data 'up' to the task level where the tags are handled. Sadly I don't really have any significant TypeScript experience and none at all regarding Angular, thus I don't feel confident that I could further pin down this problem or a solution in this rather impressive but thus also quite large project.

I'd be happy to hear about your insight, whether this is feasible and how much work this would create!

MiragonMx:
# Description

- adjust GithubIssue models to include a label array -
  moves the label element from GithubIssue to GithubIssueReduced (indirectly in GithubIssue still through extension) (`github-issue.model.ts`)
- get them from the graphQL API (`gihub-issue-map.util.ts`)
- add labels with a private helper function as tags (with import of name and color)
  using TagService utilities (`gihub-common-interfaces.service.ts`)

- add config member for a flag to import labels as tags or not (`github.model.ts`)
- set default for that config member to false and add to UI form (`github.const.ts`)
- add const elements for that form to be filled in with translation elements/values (`t.const.ts`)
- first formulations of form elements in english and german (`en.json`, `de.json`)

## Issues Resolved

Resolves #2866 

## Check List

- [ ] New functionality includes testing. (not yet fully tested, I'll have to look into it)
- [ ] New functionality has been documented in the README if applicable. (n/a, I guess, possibly open for discussion)


src/app/features/issue/providers/github/github-common-interfaces.service.ts
```diff
@@ -174,4 +181,12 @@ export class GithubCommonInterfacesService implements IssueServiceInterface {
   private _isIssueDone(issue: GithubIssueReduced): boolean {
     return issue.state === 'closed';
   }
+
+  private _getGithubLabelsAsTags(labels: GithubLabel[]): string[] {
```
johannesjo, line 185: I think, I'd take a different approach, since there is already a mechanism for adding data to thew new task in `getAddTaskData`. I'd add the additional data for the tags there and then use this info inside `IssueService.addTaskWithIssue` to trigger creating the tags there. This way we can use the same mechanism for other issue providers.

src/app/features/issue/providers/github/github.const.ts
```diff
@@ -97,6 +98,14 @@ export const GITHUB_CONFIG_FORM: LimitedFormlyFieldConfig<GithubCfg>[] = [
       label: T.F.GITHUB.FORM.IS_ASSIGNEE_FILTER,
     },
   },
+  {
+    key: 'importLabelsAsTags',
```
johannesjo, line 102: For booleans the current code base almost always uses an `is` prefix e.g. `isSomethingTrue`.

johannesjo (changes requested):
Looks very clean so far! I added a few remarks.

src/app/features/issue/providers/github/github.const.ts
MiragonMx: addressed in 67e18a659de96def3d58c68a59edc7ba981b789b

src/app/features/issue/providers/github/github-common-interfaces.service.ts
MiragonMx: I'm afraid I don't really get what you're saying here.. The way this is implemented, it does trigger through the `IssueService.addTaskWithIssue` function as the tags are imported as `getAddTaskData` function, right? And another Provider would need to support and implement such a functionality according to their API and restrictions anyway right? Could you outline a little more, where/how you'd like the logic to be settled?

MiragonMx:
Unrelated to the current review stage: I poked around to due to the requested changes and realized I forgot to implement a query to check, _if_ the tags should be imported. I'll try to mimic analogous problems (e.g. `GithubCommonInterfacesService.isBacklogPollingEnabledForProjectOnce$`, but I won't to that today, I'll have to see when I find the time. But I leave this here as an outstanding todo:
- [ ] add config check for the import of labels

Also, with a similar status/takeaway, though it might be helpful, if someone else could do this as well: I still cant get the config to work (neither dev server nor prod page as web app under Linux), this should also be test driven before a possible/eventual merge:
- [ ] test in dev server environment

johannesjo:
> `GithubCommonInterfacesService.isBacklogPollingEnabledForProjectOnce$`, but I won't to that today, I'll have to see when I find the time. But I leave this here as an outstanding todo:

Yes, this is very important I think. Likely not everybody wants to import the tags.



> Also, with a similar status/takeaway, though it might be helpful, if someone else could do this as well: I still cant get the config to work (neither dev server nor prod page as web app under Linux), this should also be test driven before a possible/eventual merge:

What type of issues are you encountering? Maybe I can help?

MiragonMx:
> What type of issues are you encountering? Maybe I can help?

Well, the GitHub integration did not work in the web app. I'm not sure if that is a 'my machine' problem or more general, first I thought it had to be due to my changes but it does not work in the production web app either - the 'save' button is simply not clickable and the configuration does not seem to be saved in the latter setting screen. Unfortunately I haven't been able to investigate this further yet on my machine or a reference windows device.
So it isn't a problem of the server not starting or anything, just unexpected behavior that I'm not sure how to place yet (maybe also a more general bug..)

src/app/features/issue/providers/github/github-common-interfaces.service.ts
MiragonMx: Still open (I don't know what mechanism you are referring to that is not already triggered/in the pipeline with the implemented approach)

johannesjo:
That sounds not like it should be :) Can you maybe open up an issue for the problems you're encountering and what steps lead you there (also important to know the browser and environment you are working with)?

MiragonMx:
Bump, didn't get to it in the past months.. :/

MiragonMx:
Once again, bump, didn't get to it in the past months.. 🫠

PR #3359, opened 2024-08-22
fix #3332 | remove all language keys from language files

### LINKED ISSUE #3332: Consistent language selection list
Ensure that the language selection list remains the same for every language, as currently, the labels change when the language is switched.

Current Behavior:
The labels for languages change based on the selected language.

Expected Behavior:
The language selection list should display the same options regardless of the current language setting.

Current:
Screenshot_1

Expected:
Screenshot_2

If a monolingual person accidentally selects the wrong language, it would be difficult for them to find and revert to their native language. This can lead to a frustrating user experience.

sagyas:
left only in en.json, modified the values

# Description

Each language in the selection should be in its origin language, and not translated

## Issues Resolved

#3332 

## Check List


johannesjo:
Thank you very much!!!

PR #3363, opened 2024-08-24
feat: Add a new keyboard shortcut for linking an attachment

prilr:
# Description

Added a keyboard shortcut to add an attachment to the currently selected task.
Previously it didn't have a shortcut associated with it, and assigning one was not possible.


johannesjo:
Thank you very much!

PR #3381, opened 2024-09-01
cicd: Handle any kind of version for AUR

alosarjos:
# Description

Handle any kind of versioning

Closes https://github.com/johannesjo/super-productivity/issues/3380


johannesjo:
Thank you very much!

PR #3397, opened 2024-09-03
Corrected wrong translation (recurring tasks)

YetAnotherGerrit:
Corrected wrong translation about order numbers of recurring tasks.

# Description

The translation of the description of the order number of recurring tasks was false in German. It said something like "A **lower value** means, the task is at the top of the list, a **lower number** means, the task is at the bottom of the list."

## Issues Resolved

The meaning of the order number was not understandable in German.


johannesjo:
Vielen Dank!

PR #3412, opened 2024-09-07
feat: Add Today Remaining in the projects view

shafiahmed12:
# Description
Filtering the tasks present in the projects and passing it to the mapEstimateRemainingFromTasks method so that the remaining time is calculated only for the tasks that are added to today

## Issues Resolved
https://github.com/johannesjo/super-productivity/issues/3221

## How it looks with this fix
image
image

#### Nothing is shown on Today page
image

## Check List


johannesjo:
Thank you very much! I've to say that – while certainly  useful – it feels a bit cluttered with 4 entries and also the display logic of when to display icons instead does not work well with 4 entries. Maybe it is better to just show this info on hover using matTooltip? What do you think?

shafiahmed12:
> Thank you very much! I've to say that – while certainly useful – it feels a bit cluttered with 4 entries and also the display logic of when to display icons instead does not work well with 4 entries. Maybe it is better to just show this info on hover using matTooltip? What do you think?

I think that is a good idea, do you want this on hover on estimate remaining?

johannesjo:
Yes and yes :) thank you!

shafiahmed12:
> Yes and yes :) thank you!

made the changes as requested hope you are okay with this one :)

johannesjo:
Thank you very much! I like it a lot! :)

shafiahmed12:
> Thank you very much! I like it a lot! :)

This was my very first contribution and thanks for the confidence boost ✌

PR #3433, opened 2024-09-10
Fix WebDAV URL concatenation on Android

### LINKED ISSUE #3432: Android WebDAV Sync doesn't concatenate URL properly
### Your Environment

- Version used: 10.0.4
- Operating System and version: Android 14

### Expected Behavior

Android client eliminate duplicate slashes before sending the request.

### Current Behavior

The URL of the request sent by the Android client has duplicate slashes if BaseUrl has a trailing slash or SyncFolderPath has a leading slash, which causes unexpected behaviors.
The sync works well if duplicate slashes are removed.
image
image

While the Windows desktop version can eliminate the duplicated slashes automatically.
I think it's related to the difference in underlying runtime implementations.
image

https://github.com/johannesjo/super-productivity/blob/371ef42d955b9a3cc06ef91035d9c9243ab22b76/src/app/imex/sync/web-dav/web-dav-api.service.ts#L97-L123

Changing all `cfg.baseUrl + '/' + path, `  to `new URL(path, cfg.baseUrl).toString(),` in [`web-dav-api.service.ts`](https://github.com/johannesjo/super-productivity/blob/master/src/app/imex/sync/web-dav/web-dav-api.service.ts) should fix this problem.

1. Set WebDAV as the Sync Provider on the Android client
2. Insert `/` to the end of Base Url or the beginning of Sync Folder Path
3. Start sync

Yes

Nyrest:
# Description

Fix WebDAV URL concatenation on Android

## Issues Resolved

fixes #3432

## Check List

### DIFF
--- a/src/app/imex/sync/web-dav/web-dav-api.service.ts
+++ b/src/app/imex/sync/web-dav/web-dav-api.service.ts
@@ -99,7 +99,7 @@ export class WebDavApiService {
     const cfg = await this._cfg$.pipe(first()).toPromise();
     if (IS_ANDROID_WEB_VIEW && androidInterface.makeHttpRequest) {
       const result = (await androidInterface.makeHttpRequestWrapped(
-        cfg.baseUrl + '/' + path,
+        new URL(path, cfg.baseUrl).toString(),
         'PUT',
         data,
         // JSON.stringify(data),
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much!

PR #3460, opened 2024-09-14
WIP: add setting to change tracking interval #2355

Minnowo:
# Description

Added a setting `Time tracking interval` under the `Time Tracking` category which lets the user change the tracking interval.

image

I've put the title here as WIP since I am a new contributor, I have never used Angular, and have very limited experience with Typescript. So feedback is appreciated.

I've got some more testing but I think most of the functionality is there. Changing the setting properly updates the interval, and the value is persisted across page loads. 

I still need to setup Electron and I'm not sure what to do about the other i18n files.

## Issues Resolved

This is a bit of a simple fix for https://github.com/johannesjo/super-productivity/issues/2355.

## Check List


johannesjo:
Thank you very much! This looks very clean and I am quite happy. Two thoughts:
1. Maybe we should put the setting at the end of the settings form and mark it with something like "EXPERIMENTAL!", since it is something for more technically versed users.
2. Related to that we should add more context on why someone want to change this setting (to reduce disk writes).

johannesjo:
Something like
```
    TestBed.configureTestingModule({
      providers: [Store],
    });
``` 
in `global-tracking-interval.service.spec.ts` will likely fix the error.

Minnowo:
> Thank you very much! This looks very clean and I am quite happy. Two thoughts:
> 
> 1. Maybe we should put the setting at the end of the settings form and mark it with something like "EXPERIMENTAL!", since it is something for more technically versed users.
> 2. Related to that we should add more context on why someone want to change this setting (to reduce disk writes).

Glad everything looks good! 

I have added `(EXPERIMENTAL)` to the title, extended the description a bit. Here is how it looks now:
image

Minnowo:
I have also added the change you suggested for that test case. Running tests using Chromium seems to pass now on my machine.

johannesjo:
Thank you very much! This is a great addition! I'll try to figure something out for the timers once I have more time on my hands!

davidhealey:
Why is the max limited to 100000? I would like to limit writes to every 5 minutes as I don't see any need for it to write more frequently than that.

Minnowo:
> Why is the max limited to 100000? I would like to limit writes to every 5 minutes as I don't see any need for it to write more frequently than that.

It was an arbitrary number I had picked that seemed large enough at the time. I didn't really think much of it.

Here is the line if you're interested in changing it:
https://github.com/johannesjo/super-productivity/blob/32aa4fb628c47952884ac4414a9284846f34c67d/src/app/features/config/form-cfgs/time-tracking-form.const.ts#L76

davidhealey:
Thank you

PR #3483, opened 2024-09-17
Fix #3447 Gitlab time submission

### LINKED ISSUE #3447: Gitlab time reporting fails because issue.id is used instead of issue.iid
## Your Environment

- Version used: 10.0.4
- Operating System and version: Archlinux
- Desktop Environment: XFCE

### Expected Behavior

Time tracking works

### Current Behavior

Time tracking does not work but returns 404 because the id instead of the iid is used

1. Sync issues from gitlab
2. Work
3. Sync time to gitlab
4. 404 not found

Yes

### Console Output

Basically 404 not found

### What is wrong

Essentially the problem is that Superproductivity in the later versions (anything that includes the https://github.com/johannesjo/super-productivity/commit/9577c3e commit) uses the issue id for certain gitlab things instead of the iid, breaking multiple interactions. This should either be reverted or the iid should be used in the correct places.

d-r-e:
# Description

The tasks used the internal ID when retrieving Gitlab issues. With this change, it uses the issue number, so that time submission works again. Tested locally with an on-premise Gitlab instance.

## Issues Resolved

#3447 

## Check List

- [ ] New functionality includes testing. N/A
- [ ] New functionality has been documented in the README if applicable. N/A

### DIFF
--- a/src/app/features/issue/providers/gitlab/gitlab-api/gitlab-api.service.ts
+++ b/src/app/features/issue/providers/gitlab/gitlab-api/gitlab-api.service.ts
@@ -212,13 +212,13 @@ export class GitlabApiService {
     time_estimate: null | number;
     total_time_spent: null | number;
   }*/
-    const iid = this._getIidFromIssue(issueId);
+
     return this._sendRawRequest$(
       {
         url: `${this._apiLink(
           cfg,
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much!

PR #3485, opened 2024-09-17
Fix Gitlab Time input

### LINKED ISSUE #3447: Gitlab time reporting fails because issue.id is used instead of issue.iid

d-r-e:
# Description

The problem is that Gitlab API expects the issue id (iid), not the "regular" id, the internal ID: this is why time submission stopped working.

On a previous PR (https://github.com/johannesjo/super-productivity/pull/3483) the gitlab input time was fixed, but after that change, issues got infinitely duplicated after being periodically polled. Forgive my mistake. This solution implements (back) the use of iid (issue ID instead of internal ID) as the API does not support fetching and posting data using the internal ID.

Release 10.0.7 makes infinite tasks when Gitlab polling is activated, so this PR aims to fix this. 

## Issues Resolved

#3447 this time for real.

Sincere apologies for previous commits.


d-r-e:
No conflicts here or am I missing something?

johannesjo:
I am getting "This branch cannot be rebased due to conflicts"

Can you maybe merge the master branch into the PR?

johannesjo:
Not sure what was going on. I merged it manually and now hopefully all should be fine. Could you maybe test the latest master once if things work as you'd expect?

PR #3486, opened 2024-09-17
Fix time format in start time description for repeated task in french

taophp:
# Description

In french, the help provides a false indication regarding the format to use for the start time of a repeated task : the help indicates "15h00", but the fact is that the correct format is "15:00".


taophp:
You're welcome !

PR #3573, opened 2024-10-12
fix: parse scheduled times using local time zone (#3192)

### LINKED ISSUE #3192: [Bug] Wrong Time Parsing
image
Note that my time zone is `+08:00`. And according to wiki:
> If no UTC relation information is given with a time representation, the time is assumed to be in local time.

miqh:
# Description

Hello there,

The latest version (`2.7.7`) of the `chrono-node` package contains a commit which should resolve scheduled time components being incorrectly offset when parsed:

- https://github.com/wanasit/chrono/commit/696d2479ce5295c9ad830ad6ac8cb2253130a8f9

I've introduced an extra test to cover this case for posterity. Note it'll fail if using `2.7.6` (as one would expect).

Also redacted what appears to be a superfluous `console.log` as I was investigating the issue to begin with.

## Issues Resolved

#3192

## Check List

Closes #3192

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,9 +9,6 @@
       "version": "10.0.11",
       "license": "MIT",
       "dependencies": {
-        "@capacitor/android": "^6.1.2",
-        "@capacitor/cli": "^6.1.2",
-        "@capacitor/core": "^6.1.2",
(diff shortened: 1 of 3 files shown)


johannesjo:
Thank you very much!

PR #3575, opened 2024-10-12
Feat/emojis as project icons

stavby:
# Description

Add the option to use emojis as project and tag icons:

image
image

(I also fixed a test in the last commit - the test date has passed and it started failing because of that, I moved the date to 2030 so it stays on saturday and won't pass soon)

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/3558

## Check List

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -66,7 +66,7 @@
             "leadingUnderscore": "allowSingleOrDouble",
             "trailingUnderscore": "allow",
             "filter": {
-              "regex": "(should)|@tags",
+              "regex": "(should)|@tags|\\d+",
               "match": false
             }
           },
--- a/angular.json
+++ b/angular.json
@@ -36,7 +36,7 @@
               "src/manifest.json",
               "src/static"
             ],
-            "styles": ["src/styles.scss"],
+            "styles": ["src/styles.scss", "node_modules/@ctrl/ngx-emoji-mart/picker.css"],
             "scripts": [],
             "webWorkerTsConfig": "src/tsconfig.worker.json",
             "browser": "src/main.ts",
@@ -205,7 +205,7 @@
             "tsConfig": "src/tsconfig.spec.json",
             "preserveSymlinks": true,
             "karmaConfig": "src/karma.conf.js",
-            "styles": ["src/styles.scss"],
+            "styles": ["src/styles.scss", "node_modules/@ctrl/ngx-emoji-mart/picker.css"],
             "scripts": [],
             "assets": ["src/favicon.ico", "src/assets", "src/manifest.json"]
           }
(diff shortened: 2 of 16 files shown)


johannesjo:
Hey there! Thank you very much for providing this PR! This is a really nice addition and I'd very much like to see this Incorporated.

I am wondering however if there maybe is an alternative route we can go, because `@ctrl/ngx-emoji-mart` is not a  small dependency. I should have mentioned this before you started, but adding new dependencies is always a bit tricky, since they often require some extra maintenance in the long run and sometimes need to be replaced completely, which is often painful. When you made the proposition about this features I was expecting something with only native utf emojis, which would have been much more lightweight. Should have spelled this out before. I am sorry for that!

What do you think? Would be using only native utf emojis also be a possible option?


On a further note: I had a little bit of difficulties getting to the emojis in the picker, since they only show up last. Maybe we can also improve on that.

stavby:
Hey,
I understand what you're saying, although I still think that dependency is the best option.

I used it because of 2 main advantages:
1. Custom emoji set display (I used Apple's) - Windows and Android emojis look terrible in my opinion and I liked that I can use better looking ones.
2. Emoji search - it provides an API to search emojis in a smart way, the alternative would be to let the user paste emojis which they might not knoe about and not a common thing to do using PCs. You could also maintain a list of all emojis (which the package provides), but I think that's too much of a pain.

I don't support using native emojis, so it's up to you if you want to go forward with this PR, I did use a heavy dependency without consulting you so that's on me too.
About the order - we can make the emojis show up before the icons.

johannesjo:
Hey there @stavby ! I spent quite some time thinking about this. I am all for adding basic unicode icons (which we could extract from here: https://github.com/missive/emoji-mart/blob/main/packages/emoji-mart-data/sets/5/native.json), but adding such a big dependency (the apple icons alone are 4.2MB) doesn't feel right to me personally. Using this would also affect the user data the app saves, which would make migrating away, when needed, all the more difficult. Unicode icons on the other hand, are a widely used (though a bit inconsistent) standard.

I am very sorry for the time you sunk into this and I hope you can understand my reasoning. It would be great if you could rework the PR, but I also would totally understand if you choose not to.

Not sure if I am overlooking anything, but in my head a possible implementation would work like this:
* the [native.json](https://github.com/missive/emoji-mart/blob/main/packages/emoji-mart-data/sets/5/native.json) is extracted into a list of id and unicode icon (e.g. `[{id:'smile',ico:"😊'}]`)
* the icons are just integrated and alphabetically sorted into the existing list and the id is used for selecting the icons in the autocomplete 
* we display icons by making use of the mat-icon component (if not possible, we use a dedicated wrapper component instead)

Alternatively we could add a dedicated picker just listing all the unicode icons, such as messengers like whatsapp do.

What do you think?

stavby:
Hey, sorry for disappearing.
Unfortunately I really don't have the free time currently to work on these things.

Feel free to close this PR or assign it to someone else. I would also appreciate it if you could unassign me from the issue.

Best of luck :)

Issue #3576, opened 2024-10-13
Feature: Standardize App Name Across Platforms


jiongxuan:
## Problem Statement

The application name is displayed inconsistently across different platforms, which may affect brand recognition. For example, on Mac, the name appears as superProductivity, on Android it shows as SupProd, and on Windows and Ubuntu (as seen in screenshots) it displays as Super Productivity. This inconsistency could potentially hinder branding efforts.

iOS: 
![image](https://github.com/user-attachments/assets/ec875406-51ca-46bd-9f03-84120db0c972)

Android:
![image](https://github.com/user-attachments/assets/94870853-10f6-4334-8251-31288218d8c3)

## :grey_question: Possible Solution

I suggest unifying the application name across all platforms as Super Productivity. For Android, if there are concerns about the name being too long, a compromise could be to use Super Prod, which would still retain brand recognition without causing truncation issues.

Pixel: 
![image](https://github.com/user-attachments/assets/7560e5f2-07dd-4c10-95da-6afd41d3d552)

## :arrow_heading_up: Describe alternatives you've considered

An alternative could be to simplify the name to under 10 characters to avoid truncation, but this might not fully align with the brand’s established presence as Super Productivity. Another option is to use Super Prod on Android as a middle ground.

Which option do you think is better? If needed, I’m available to help make the changes.

johannesjo:
I think we can close this, right @jiongxuan ?

beerkumquatpome:
Please change the window class name to Super Productivity too (It's superProductivity)

johannesjo:
I don't think this is possible without being confusing and inconvenient to existing users ( app data might be lost).

PR #3579, opened 2024-10-13
Standardize App Name Across Platforms (#3576)

jiongxuan:
# Description

This PR standardizes the application name across platforms to improve brand consistency. 

The app name is now set to Super Productivity on Mac, and Super Prod on Android to avoid truncation issues. 

## Issues Resolved

- https://github.com/johannesjo/super-productivity/issues/3576

## Check List

- [✅] New functionality includes testing.
- [✅] New functionality has been documented in the README if applicable. (No needed)

### DIFF
--- a/android/app/src/main/res/values/strings.xml
+++ b/android/app/src/main/res/values/strings.xml
@@ -8,7 +8,7 @@
         \"site\": \"https://app.super-productivity.com\"}
         }]
     </string>
-    <string name="title_activity_fullscreen">SupProd</string>
+    <string name="title_activity_fullscreen">Super Prod</string>
     <string name="widget_no_data">Please tap here to load data</string>
     <string name="empty_task_list_text">Tap to refresh task list</string>
     <string name="service_background">Service is running background</string>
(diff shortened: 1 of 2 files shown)


jiongxuan:
Please hold on for a moment. I’ve discovered that changing the app name on Mac results in the creation of two different apps, which prevents upgrade installations. I’ll investigate further, and if this issue cannot be resolved, I will proceed with modifying only the Android name for now, without changing the Mac version.

jiongxuan:
The compatibility issue with renaming has turned out to be far more complex than I initially thought. I’ve tried the following approaches:

1.	Using CFBundleDisplayName to modify directly: It had no effect; the app name didn’t change.
2.	Using InfoPlist.strings with localization: The files weren’t properly copied, and it didn’t work.
3.	Using extraResources to copy the files: The localization files weren’t applied as expected.
4.	Using an afterPack hook to manually modify Info.plist: It was overly complex and produced unstable results.

Since none of these solutions resolved the issue, and for the sake of stability, I will resubmit this PR, only changing the Android app name, leaving other platforms unchanged. I’ll revisit this in the future if a better solution is found.

New PR: https://github.com/johannesjo/super-productivity/pull/3580  (android only)

johannesjo:
Thanks for digging into this!

PR #3604, opened 2024-10-19
Bugfix: Prevent the Today tag to be added multiple times

panoramix360:
# Description

This fix prevents the today tag to be added multiple times on the same task using the Context Menu and the Schedule Dialog.

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/3518


johannesjo:
Thank you very much for this!

PR #3659, opened 2024-11-06
Enable Cyrillic letters that was recognized as separator

panoramix360:
# Description

Enables the users that uses cyrillic letters to add tags with the cyrillic letter `б` that was being recognized as a separator for tags. Now the user can add this character and press enter to create the tag.

[It seems to be a know issue on Angular Material as well.](https://github.com/angular/components/issues/13989)

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/3640

Thanks!


johannesjo:
Thank you very much!

PR #3662, opened 2024-11-06
Update Arch Linux PKGBUILD_template for new artifacts changes

alosarjos:
New packaging seems to change a couple of things:

1. Package version is no longer included on the filename of the release artifact
2. Looks like the directory inside the package new includes spaces on the target install dir

Closes https://github.com/johannesjo/super-productivity/issues/3656

### DIFF
--- a/build/linux/PKGBUILD_template
+++ b/build/linux/PKGBUILD_template
@@ -14,13 +14,13 @@ provides=("${_pkgname}")
 conflicts=("${_pkgname}"
 		   "${_pkgname}-git")
 md5sums=('SKIP')
-source=("https://github.com/johannesjo/super-productivity/releases/download/v${pkgver}/superProductivity_${pkgver}_amd64.deb")
+source=("https://github.com/johannesjo/super-productivity/releases/download/v${pkgver}/superProductivity-amd64.deb")
 
 package() {
 	tar -xvf data.tar.xz -C "${pkgdir}"
 	rm -rf "${pkgdir}/usr/share/doc"
     install -d "${pkgdir}/usr/bin"
-	ln -sf "/opt/superProductivity/${_pkgname}" "${pkgdir}/usr/bin/${_pkgname}"
+	ln -sf "/opt/Super Productivity/${_pkgname}" "${pkgdir}/usr/bin/${_pkgname}"
     install -Dm644 -t "${pkgdir}/usr/share/licenses/${pkgname}" \
-    "${pkgdir}/opt/superProductivity/"{'LICENSE.electron.txt','LICENSES.chromium.html'}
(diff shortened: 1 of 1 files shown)


alosarjos:
@johannesjo Would love if you could give this a quick check, since no matter if I fix manually the AUR package, until this is merged everytime a new version is released, the Arch package breaks

siliconmeadow:
@johannesjo / @alosarjos - Is this file used by any other builds/distros than Arch Linux? If not, I don't see the point in maintaining it in your repo. If it is used by other distros, I'd prefer if the checksums were actually validated every time instead of skipped. It's a bad security practice to skip them.

It adds a little overhead for the AUR maintainer to download the binary releases and do a `sha256sum ./superproductivity<release_version>.deb` and then update the `PKGBUILD` file for the AUR package. But it's the responsibility of the maintainer to ensure they follow security best practices.

This is why I'd suggested the patch [here](https://aur.archlinux.org/packages/superproductivity-bin#comment-997423).

alosarjos:
@siliconmeadow I'm guessing you are the same user also commenting on the AUR Repo.

In order to have software updates delivered fast and easy to AUR @johannesjo and I worked back on the day on automatizing the process through Github actions. 

@johannesjo kindly accepted adding this automation to the AUR to avoid needing a maintainer to manually update packages and make things easier. It's clear that there are things that can be improved (Like the checksum validation) but we did the best we could back in the day and has worked good enough for almost 4 years.

4 years getting this updated without needing any manual intervention. I personally think that great. I will check if I can automate generating the checksum for the files (I'm more familiar with Gitlab CICD than Github actions), but I think that can go on another merge request. For now this one at least should keep things working as it has until now.

johannesjo:
Thank you very much @alosarjos !!!

> @johannesjo Would love if you could give this a quick check, since no matter if I fix manually the AUR package, until this is merged everytime a new version is released, the Arch package breaks

How should I check it? Would it be enough to just see if it works for the next release? Or is there a better way?


@siliconmeadow 
Thank you for your input. Maintaining a all the different platforms can be very time consuming and unfortunately there are comprises to be made to keep that effort as low as possible to leave time for actual development of bug fixes and new features :)

siliconmeadow:
@alosarjos - thanks for explaining that, it never occurred to me that's what you'd done. Cool cool deal. I'll have a think about how GH actions might be able to produce the checksum somehow. Or does the AUR have full access to Arch's gitlab instance and therefore something can be done via `gitlab-ci`? hmmmm...

@johannesjo - I might be speaking out of turn, but I think that's probably the only thing you can do is to merge this PR and we see what happens at next release. @alosarjos would need to confirm, of course.

alosarjos:
@johannesjo Lets merge for now, and lets see how next release goes. Things are broken without this PR, things can't go worse :P

johannesjo:
Oh. Forgot to hit the merge button :D

PR #3737, opened 2024-12-06
feat: improve durations input format #3736

### LINKED ISSUE #3736: Improve durations input format
## Problem Statement

In time estimates or time spent, we can input duration as a string instead of using the nice circular widget.

1. The most frustrating thing, and I think not that hard to fix, is that we have to separate our time component with a space or it fails.
For example, `1h 45m` works just fine while `1h45m` is calculated as `1m`
2. As an improvement, we could allow to input `1h45` and guess it was 45 minutes
3. Another improvement would be to allow using floats like `1.5h`

Improve `stringToMs` to:
1. split on anything but numbers (+ dots, and commas) instead of just spaces.
2. guess the letter if none has been provided for the current part
3. use `parseFloat` instead of `parseInt`

None

Might help solve #3420 and #2983

DuaelFr:
# Description

Allow:
1. to input duration without spaces: `2h30m`
2. to use floats: `2.5h`
3. to omit last unit: `1h30`

## Issues Resolved

#3736

## Check List

### DIFF
--- a/src/app/ui/duration/string-to-ms.pipe.spec.ts
+++ b/src/app/ui/duration/string-to-ms.pipe.spec.ts
@@ -16,4 +16,24 @@ describe('stringToMs', () => {
   it('should not return NaN for just h', () => {
     expect(stringToMs('h')).toBe(0);
   });
+
+  // @covers #3737
+  it('should work improved', () => {
+    // Case: no spaces.
+    expect(stringToMs('1h10m')).toBe(4200000);
+
+    // Case: missing last unit.
+    expect(stringToMs('1h 10')).toBe(4200000);
+    expect(stringToMs('1h10')).toBe(4200000);
+
+    // Case: missing all units.
+    expect(stringToMs('10')).toBe(0);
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! Could you by any chance also add some unit tests for this? That would be great!

DuaelFr:
I've never done unit tests in JS.
Do you have some guidance to help me start?

Le sam. 7 déc. 2024, 14:56, Johannes Millan ***@***.***> a
écrit :

johannesjo:
Yes! You could have a look at `src/app/ui/duration/string-to-ms.pipe.spec.ts`. There are already some tests there. You just need to copy and paste, with some more input values for the function and the expected result. 

Let me know if you have any questions!

DuaelFr:
I'm lucky there were already examples. That was not that hard to expand!
I hope I've done that well.

johannesjo:
Thank you ver ymuch @DuaelFr !! :)

PR #3739, opened 2024-12-07
Update Linux categories in electron-builder.yaml

aripollak:
# Description

I believe this is a bit more specific than the existing `Utility`, based on https://specifications.freedesktop.org/menu-spec/latest/additional-category-registry.html. These categories are what I put in the flatpak appstream. metadata file.

## Check List


johannesjo:
Thank you!! Not sure this will be compatible with snap, but let's see what happens :)

PR #3762, opened 2024-12-16
Update french translations

Ulfheonard:
Hello,

I added french translations

Preview the messages on https://inlang.com/github.com/Ulfheonard/super-productivity .

There are several English texts that I have not found the possibility to translate, for example:

{1A209007-1A06-44C9-A7EE-58A6F4CCD963}

Thank you


johannesjo:
Thank you very much! There seem to be many  changes also to other languages than french (just the order of translation keys).

Any ideas why that might be the case?


@ping @NilsJacobsen

Cyber-Syntax:
@johannesjo 
According to my review; I think Fink is trying to add missing strings again by changing their location, sending them outside of brackets. When I review those changed strings for Turkish language, it seems like they are unused strings or string names changed in en.json, but Fink still isn't solving them; just adding them outside of brackets. You can basically understand this by looking at my PR with the string `CREDENTIALS`. I deleted this because it's no longer used on en.json, but still exists on tr.json.

johannesjo:
@Cyber-Syntax do you think there is a good workaround for this?

Cyber-Syntax:
> @Cyber-Syntax do you think there is a good workaround for this?

A basic solution would be to add new strings to other language files from en.json. 

I don't know how you handle internationalization (i18n), but it would be great to have a string update feature for other language files from en.json. I mean, when you add new features, you create new strings and update en.json, but you don't update the other language files, which leads contributors to add them manually. 

In my experience, I was adding them manually, but I've developed a basic Python script to compare en.json with tr.json to find missing strings and add them to tr.json. However, I think this should be handled by i18n functionality, which I'm not familiar with because I have no idea how things work on i18n creation in TypeScript apps.

NiklasBuchfink:
Hey 👋 
I'm from inlang and builder of Fink. The `plugin-json` supports nesting and the translations should stay in the correct bracket, but it keeps the sorting in all language files in line with the source language file, in this case `en.json`.

> I don't know how you handle internationalization (i18n), but it would be great to have a string update feature for other language files from en.json. I mean, when you add new features, you create new strings and update en.json, but you don't update the other language files, which leads contributors to add them manually.

If you add just the empty keys to all the files, you always need to check programmatically if a key is an empty string to fallback to the source language. That's why we defaulted to "not having a key" instead of "having empty strings".

Tooling could make it easier to work with:
- you can use [fink.inlang.com](https://fink.inlang.com/github.com/johannesjo/super-productivity) to have a translation UI that adds the translation and it's key at the same spot as it is in the source language file
- you can use the [sherlock VS Code extension](https://inlang.com/m/r7kp499g/app-inlang-ideExtension) if you want to do the same, but in code
- you can also use the [inlang CLI](https://inlang.com/m/2qj2w8pu/app-inlang-cli) to machine translate everything by default in your CI/CD pipeline

@Cyber-Syntax @johannesjo Let me know if this helps or if you still want to create a script

Cyber-Syntax:
> supports nesting and the translations should stay in the correct bracket,

If you look at  [my commit](https://github.com/johannesjo/super-productivity/pull/3846/commits/ade830db989faa8689718649caf0d466db3652be), check the IS_AUTO_POLL, IS_AUTO_IMPORT_ISSUES, IS_SEARCH_ISSUES_FROM_* strings and compare with [current PR commits](https://github.com/johannesjo/super-productivity/pull/3762/files). All of them is not in the en.json but still added tr.json and other languages. I don't know how your code is work but if empty strings need check everytime, isn't it unused strings going to checked too? 

> If you add just the empty keys to all the files, you always need to check programmatically if a key is an empty string to fallback to the source language. That's why we defaulted to "not having a key" instead of "having empty strings".

`"TITLE": "Jira: Submit Worklog"` 
Are strings like this considered an empty string when added to other language files? I was meant to add those to other language files; therefore, contributors can update them later, but if it causes a problem, of course we'll need to find a different approach.

NiklasBuchfink:
@Cyber-Syntax I don't know if I understand correctly, so let me recap this.

Is it correct, that `S_AUTO_POLL, IS_AUTO_IMPORT_ISSUES, IS_SEARCH_ISSUES_FROM_*` where added in different language files, but not to the `en.json` file?

Fink does the following:
- it first loads the keys all keys of the source file (en.json)
- collect all other translations with the same key to display them along (this is where the sorting comes from)
- keys that only appear in other language files than en.json are detected as "messages without source" >> [see here](https://fink.inlang.com/github.com/johannesjo/super-productivity?project=%2Fproject.inlang&lint=messageLintRule.inlang.messageWithoutSource&lang=en&lang=ar&lang=cz&lang=de&lang=es&lang=fa&lang=fr&lang=hr&lang=id&lang=it&lang=ja&lang=ko&lang=nb&lang=nl&lang=pl&lang=pt&lang=ru&lang=sk&lang=tr&lang=uk&lang=zh&lang=zh-tw)
- when we write back to the files, the sorting aligns with the `en.json` and "messages without source" are moved to the end of the file

This should explain, why they are moved to the end of the file and out of the brackets.

It expects the workflow of developing new features in the source language of the project (English) and adding other languages at the same time or afterwards.

johannesjo:
> It expects the workflow of developing new features in the source language of the project (English) and adding other languages at the same time or afterwards.

So if I understand this correctly everything is likely caused by translations without sources in the en.json (which is our base file) which are automatically sorted last?  Is there some way to auto delete those, or at least to delete them all at once?

NiklasBuchfink:
> So if I understand this correctly everything is likely caused by translations without sources in the en.json (which is our base file) which are automatically sorted last?

Correct 👍 

> Is there some way to auto delete those, or at least to delete them all at once?

At the moment we have nothing in place for that. So you need to do it manually or add a declutter script.
One user of our i18n library ParaglideJS did exactly that with [knip.dev](https://knip.dev), see here:
https://github.com/opral/inlang-sdk/issues/6#issuecomment-2341940766

johannesjo:
Thank you very much @NiklasBuchfink ! I really appreciate it!

PR #3827, opened 2025-01-06
Fix drag handle to allow text selection

### LINKED ISSUE #3724: Can't highlight text in text fields with mouse drag
### Your Environment

- Version used: Version 10.2.3 (10.2.3) installed with homebrew
- Operating System and version: macOS 15.1.1 (24B91)

### Expected Behavior

After creating a task, if I click on the text area, it becomes an active text input box, where I can type or edit the text. When I want to select more than just one word (which works by double clicking on a single word) from within the text field, I mouse click and drag to highlight a sentence.

### Current Behavior

At the moment, I cannot select multiple words by clicking and dragging with the mouse. The whole task becomes a draggable object and I cannot select the text I want.

1. Make a new task.
2. Type a task that is more than a single word, eg: "This is a test task and I want to highlight and copy this latter part of this task."
3. Enter.
4. Click on the text of the newly created task and see the text box appear.
5. Try click+drag to select multiple words. It should not work and the whole task moves and some glitchy space appears below the task and disappears after you click on the task again.

Yes

### Console Output

I don't see any.

Leadernelson:
Fixes #3724

Fix the issue where text in text fields cannot be highlighted by mouse drag without the whole task becoming a draggable object.

* Add `cdkDragHandle` directive to the `drag-handle` div in `src/app/features/tasks/task/task.component.html`.
* Update the `enterPredicate` method in `src/app/features/tasks/task-list/task-list.component.ts` to ensure it only allows drag events on the drag handle.

### DIFF
--- a/src/app/features/tasks/task-list/task-list.component.ts
+++ b/src/app/features/tasks/task-list/task-list.component.ts
@@ -122,23 +122,8 @@ export class TaskListComponent implements OnDestroy, AfterViewInit {
   }
 
   enterPredicate(drag: CdkDrag, drop: CdkDropList): boolean {
-    // TODO this gets called very often for nested lists. Maybe there are possibilities to optimize
-    const task = drag.data;
-    // const targetModelId = drag.dropContainer.data.listModelId;
-    const targetModelId = drop.data.listModelId;
-    const isSubtask = !!task.parentId;
-    // console.log(drag.data.id, { isSubtask, targetModelId, drag, drop });
-    // return true;
-    if (isSubtask) {
-      if (!PARENT_ALLOWED_LISTS.includes(targetModelId)) {
-        return true;
-      }
-    } else {
-      if (PARENT_ALLOWED_LISTS.includes(targetModelId)) {
-        return true;
-      }
-    }
-    return false;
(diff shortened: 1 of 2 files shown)


src/app/features/tasks/task-list/task-list.component.ts
```diff
@@ -122,23 +122,8 @@ export class TaskListComponent implements OnDestroy, AfterViewInit {
...
   enterPredicate(drag: CdkDrag, drop: CdkDropList): boolean {
-    // TODO this gets called very often for nested lists. Maybe there are possibilities to optimize
-    const task = drag.data;
-    // const targetModelId = drag.dropContainer.data.listModelId;
-    const targetModelId = drop.data.listModelId;
-    const isSubtask = !!task.parentId;
-    // console.log(drag.data.id, { isSubtask, targetModelId, drag, drop });
-    // return true;
-    if (isSubtask) {
-      if (!PARENT_ALLOWED_LISTS.includes(targetModelId)) {
```
johannesjo, line 133: I don't think we can remove this, since this handles the logic that prevents some illegal drag and drop operations.

PR #3895, opened 2025-01-14
sync title (aka summary) from SP -> CalDAV server

### LINKED ISSUE #3886: Support Title Sync: Super Productivity -> CalDAV Server
## Problem Statement
Following up on https://github.com/johannesjo/super-productivity/pull/3871

I'd like the title (aka summary) to sync from SP -> the CalDAV server for a task imported from a CalDAV server.
Note: This will not fix #3017 as this is only for imported tasks.

Why this as the next step?
1. We already have the title imported and editable on the SP side.
2. We already have the SP -> CalDAV setup and working for the task status.
3. As such, I believe this is the smallest next step possible

What should come after this? I'd propose:
1. If we delete a task in SP, delete it in CalDAV
2. Then add importing the description from CalDAV (it's there but in a diff field)
3. Then add sending the description SP -> CalDAV
4. Then once we have title, description, status, deletion all pushing SP -> CalDAV we can look into pulling updates on these things from CalDAV

Requirements:
- [ ] When you edit the title of a task in SP it is pushed to CalDAV (if imported from CalDAV
- [ ] Update the docs [here](https://github.com/johannesjo/super-productivity/blob/master/docs/caldav.md)

https://github.com/johannesjo/super-productivity/blob/master/docs/caldav.md

RayBB:
# Description

Have a task that was imported from CalDAV?
Now, when you update the title/summary it'll be pushed to the CalDAV server.

This is a basic draft for the PR. Looking to hear if there's any feedback.
The codebase has quite a lot going on so I'm not sure if I'm breaking any conventions.
Look forward to hearing more!

PS: I only changed the one line of the table in the readme but the linter changed the rest. Let me know if that needs to be pulled into a separate PR.

## Issues Resolved

Closes #3886

## Check List

  - I don't see any testing infra setup for CalDAV and I'm not so familiar with the tooling so pointers would be appreciated.

### DIFF
--- a/docs/caldav.md
+++ b/docs/caldav.md
@@ -1,33 +1,35 @@
 # CalDAV Features and Roadmap
 
 ## Overview
+
 Super Productivity offers basic CalDAV integration for importing and syncing tasks. This document outlines the current capabilities and limitations of the integration. If you're a developer and would like to improve the integration help would be very welcome.
 
 ## Feature Support Matrix
 
-| Feature | Import | Sync SP → CalDAV | Sync CalDAV → SP |
-|---------|---------|------------------|------------------|
-| Title | ✅ | ❌ | ❌ |
-| Description | ✅ | ❌ | ❌ |
-| Tags | ✅ | ❌ | ❌ |
-| Complete/Incomplete Status | ✅ | ✅* | ❌ |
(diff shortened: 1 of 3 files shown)


johannesjo:
Thank you very much! Looks good so far!

PR #3901, opened 2025-01-15
allow import of group and instance issues

### LINKED ISSUE #2529: Support GitLab groups/subgroups when querying API for projects/issues
## Problem Statement

I'm trying to set up the GitLab.com integration, and my current settings aren't working no matter what values I use. In this case, my company's GitLab org (we'll call it `company`) has a `backend-team` _subgroup_.
Screenshot 2023-03-21 at 10 57 46 PM

Regardless of what "source" I set, the app tries to call the `/projects/<group>/issues` endpoint. From what I can tell, the official GitLab API docs don't mention that specific endpoint: https://docs.gitlab.com/ee/api/rest/index.html

If you instead use the `/groups/<group>%2<subgroup>/projects` (the `%2` is important) endpoint to get a list of projects and then iterate over that list of project IDs, you could get the issues from the `/projects/<id>/issues` endpoint.

I've also tried just using the top-level group for the company, but that doesn't seem to work either.

Relevant existing/closed issues and PRs:
- https://github.com/johannesjo/super-productivity/issues/697
- https://github.com/johannesjo/super-productivity/pull/1560

cbergmann:
# Description

This allows searching and importing of gitlab issues on a group and instance level.

## Issues Resolved

#2529 

## Check List


johannesjo:
Hey there! Could you check the linting issue?

Also: There was something similar implemented in the past, which broke the gitlab integration for a lot of people . Did you test the feature for possible side effects for different configurations?

cbergmann:
Hi,
thanks for looking over this.

The last commit fixes the linting issue.

I tested this with a Installation that was created before the patch and it seemd to be working. Then I changed the new setting and got results from other gitlab projects. I also changed it back to the default and only got the issues from the project. Any hints on what additional checks I could make are greatly welcome. While testing I also found that the link to the issue in the search result was broken and fixed that.

I think the last change that also broke it for some people was by me and you are rightfully critical of this change. But because of the new structure of the gitlab implementation this cange is a lot smaller and I think therefore better to review.

The only thing I am not sure about are automated tests. I have not yet written tests for angular.

johannesjo:
@cbergmann ping!

PR #3918, opened 2025-01-21
Target Android 15 instead of Android 14

ghost:
To do this I edited the build.gradle

# Description
It targets and compiles for Android 15 instead of Adroid 14.
Here are some relevant things to keep in mind before merging:
https://developer.android.com/about/versions/15/behavior-changes-15

## Issues Resolved
https://github.com/johannesjo/super-productivity/issues/3884

## Check List

### DIFF
--- a/src/app/features/issue/providers/gitlab/gitlab-issue/gitlab-issue.effects.ts
+++ b/src/app/features/issue/providers/gitlab/gitlab-issue/gitlab-issue.effects.ts
@@ -1,6 +1,71 @@
-import { Injectable } from '@angular/core';
+import { inject, Injectable } from '@angular/core';
+import { MatDialog } from '@angular/material/dialog';
+import { first } from 'rxjs/operators';
+import { GITLAB_TYPE } from '../../../issue.const';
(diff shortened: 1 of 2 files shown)


johannesjo:
Thanks for opening this up! Did you test if the android app builds and works with this change?

johannesjo:
> No, I can try it but I have never made a build. Do I need to use Android Studio for that?

Yes!

johannesjo:
> Currently I guess some testing has to be done on an Android 15 device or in a Android 15 emulator like in Android Studio

Yes! I think checking if all basic stuff works (creating some tasks), should be enough.

johannesjo:
No screen recording needed. I trust you :) Just testing once would be great!

johannesjo:
Thank you very much for testing this!

> What do you think a fitting solution is?

Not sure. I might be mistaken, but I don't think that it is a high priority to increase the target version. I'd rather focus my efforts some place else for the time being.

johannesjo:
> With Android 16 being finalized in the coming months is it maybe better to wait for the 36 sdk to come out, so that does not have to be addressed later?

Yes, lets do it like this.

PR #3935, opened 2025-01-26
Prevent page from closing when pressing back in task detail panel

### LINKED ISSUE #3607: Back Button Navigation Issue in Task Detail Page on Android
## Your Environment

- Version used: 10.0.11
- Operating System and version: Android or Mac

## Expected Behavior

When opening a task detail page from a project in Android, pressing the back button should close the task detail page and return to the project page.

## Current Behavior

In certain scenarios, pressing the back button can either exit the app or navigate to another page unexpectedly. For example, if the following actions are performed:

1.	Start the app.
2.	Open the settings page.
3.	Open a project page (e.g., Project A).
4.	Click a task to open the task detail page.

At this point, pressing the back button is expected to close the task detail page and return to the project page. However, it instead navigates back to the settings page. Additionally, if a user opens a project page immediately after launching the app, then opens the task detail page, pressing the back button exits the app entirely, which is undesirable.

Similarly, **on Mac (using a mouse back button)**, pressing it on the task detail page will **also** cause the app navigate to an incorrect page, instead of returning to the project page.

## Steps to Reproduce (for bugs)

1.	Launch the app.
2.	Navigate to the settings page.
3.	Open a project page.
4.	Open a task detail page.
5.	Press the back button.

## Can you reproduce this reliably?

Yes, the issue is reproducible in both Android and Mac.

## Investigation

From my investigation, the core issue appears to be that the task detail page does not have its own URL in the router. When opening the task detail page, the actual URL remains the same as the project page (e.g., /project/-jXtsphXpU599N-zTkFdQ). As a result, pressing the back button either returns to an earlier page in the history (such as the settings page) or, if there is no earlier page, exits the app according to Android’s logic for handling back presses.

## Suggested Solution

Consider adding a dedicated URL route for the task detail page. This can be similar to how the project page is structured. For example, the task detail page URL could be something like:

http://localhost/index.html#/project/-jXtsphXpU599N-zTkFdQ/tasks/xxxxxx,

where xxxxxx represents the task ID.

Since I am not very familiar with frontend frameworks and currently have limited time, I would appreciate assistance in resolving this issue. Thank you for your support!

benpbo:
# Description

Prevents page from changing or app from closing upon pressing the back button (in android or with a mouse). This is done by using `window.history` to keep the panel as another history state. When the panel is destroyed, the history entry is deleted. And when going back in history is attempted, the panel closes itself.

## Issues Resolved

#3607

## Check List

### DIFF
--- a/src/app/features/tasks/task-detail-panel/task-detail-panel.component.ts
+++ b/src/app/features/tasks/task-detail-panel/task-detail-panel.component.ts
@@ -9,6 +9,7 @@ import {
   input,
   LOCALE_ID,
   OnDestroy,
+  OnInit,
   viewChild,
   viewChildren,
 } from '@angular/core';
@@ -124,7 +125,7 @@ interface IssueDataAndType {
     IssueIconPipe,
   ],
 })
-export class TaskDetailPanelComponent implements AfterViewInit, OnDestroy {
+export class TaskDetailPanelComponent implements OnInit, AfterViewInit, OnDestroy {
   attachmentService = inject(TaskAttachmentService);
   taskService = inject(TaskService);
   layoutService = inject(LayoutService);
@@ -403,6 +404,14 @@ export class TaskDetailPanelComponent implements AfterViewInit, OnDestroy {
     this.isDragOver = false;
   }
 
+  @HostListener('window:popstate') onBack(): void {
+    this.collapseParent();
+  }
+
+  ngOnInit(): void {
+    window.history.pushState({ taskDetailPanel: true }, '');
+  }
+
   ngAfterViewInit(): void {
     this.taskService.taskDetailPanelTargetPanel$
       .pipe(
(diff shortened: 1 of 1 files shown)


johannesjo:
Looks promising! What is missing for the PR to leave the WIP status?

benpbo:
Mostly not sure if there are any tests that should be added, this code does accomplish the functionality at least as I tested it on my computer and Android Studio.

Also, not sure if the use of `window.history` is correct, I don't have previous experience with Angular and I see that `Router` is usually used to "push" history (although it does require a URL and not any object)
See https://github.com/angular/angular/issues/28954#issuecomment-1837587804

Anyway I don't think there is a need to keep it as WIP anymore.

johannesjo:
Thank you! I think it works really nicely! I am very happy about the new behavior!

Issue #2829, opened 2023-10-24
Remembering whether a task was added to the top or bottom of a list.


ClanEver:
The current default behavior is to add tasks to the top of the list, while I'm used to adding them to the bottom.
Although I can currently click an icon or use a shortcut key to change it, having to do an extra action every time I add a task is a bit annoying.
I hope there could be a way to remember whether I last added a task to the top or bottom of the list.

Tim-H-OC:
Still relevant. I would like this feature too... very annoying having to drag every item to the bottom after creation... I work on tasks top to bottom so new tasks should default to the bottom of the list. This could be a settings toggle or just have the up/down arrow in the task creation dialog box remember its setting.

Tim-H-OC:
Yes, I believe still relevant.

johannesjo:
I believe that this should be fixed now.

PR #3976, opened 2025-02-05
feat: save last used new task position in localStorage

kkoyung:
Update: change to new approach to the issue

Save `isAddToBottom` value to local storage after toggling it, and use it next time the add task bar is displayed.

This approach is more intuitive to users, and no new setting is introduced.

---

Original PR: **feat: add new setting of default position of new tasks**

# Description

There is a toggle at the right end of the new task bar that controls where a new task is placed: either at the 'top of the list', or the 'bottom of the list'. The default setting for this toggle is hardcoded to 'top of the list,' meaning it automatically reverts to this option every time the toggle is displayed.

screenshot_2025-02-05_21:07:01

A new setting in "Misc Settings" is added to let user choose the default choice of the toggle.

screenshot_2025-02-05_21:07:58

## Issues Resolved

[#2829 Remembering whether a task was added to the top or bottom of a list.](https://github.com/johannesjo/super-productivity/issues/2829)


johannesjo:
I was thinking if it is maybe better just to persist the last value to the local storage. What do you think?

If not

kkoyung:
> Thank you! I was thinking if it is maybe better just to persist the last value to the local storage. What do you think?

Your approach sounds more intuitive to users. I will try to implement it. Stay tuned.

kkoyung:
I reverted the last commit, and implemented the new approach . Check it out.

johannesjo:
Very nice! Thank you very much!

PR #3989, opened 2025-02-10
Add icon for F-Droid

leoheitmannruiz:
# Description

F-Droid clients and repository browsers currently don't show an icon for Super Productivity: https://f-droid.org/en/packages/com.superproductivity.superproductivity/

This fixes it: https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/#fastlane-structure

This is the icon shown at https://play.google.com/store/apps/details?id=com.superproductivity.superproductivity, compressed using zopflipng.


johannesjo:
Thank you very much!

PR #4011, opened 2025-02-17
Added Option to Choose Voices for Domina mode.

LunaVoid:
# Description
https://github.com/johannesjo/super-productivity/discussions/2991
As mentioned in Discussion, it is very nice to have the ability to choose which voices you would like to use.

This change basically just adds a list to the settings that has all the voices, it adds a voice parameter to the speak function and adds two new functions getAvailableVoices and getDefaultVoice. The first is used to populate the list and the second is to recreate the same behavior that already happens when using Domina mode. The default voice is just chosen in that case.

## Issues Resolved
https://github.com/johannesjo/super-productivity/discussions/2991

If needed I can also open a separate issue for this.

## Check List


johannesjo:
Very nice! I will use Italian English as my Domina voice from now on :D

Thank you very much!

PR #4086, opened 2025-03-06
feat(fullScreenTrackingReminder): add fullscreen tracking reminder

onesvat:
feat(fullScreenTrackingReminder): add fullscreen control for main window and trigger and focus main window when no tracking for minutes defined (same as reminder tracking banner)

# Description

This PR introduces a feature designed to help maintain focus by automatically switching the main window to fullscreen mode when time tracking is inactive and a predefined duration has elapsed. This behavior mimics the reminder tracking banner and is intended to prevent distractions—such as browsing social media—when no task is being actively tracked.

When the fullscreen reminder is active (as enabled in the settings with a specified duration), the application forces focus to prompt the user to select a task and start tracking. Once a task is selected and tracking begins, the application reverts to its normal state. Users also have the option to dismiss the reminder if they prefer.

Note: This feature is only available on the desktop version. It has been tested on macOS but follows the same code structure as the fullscreen break reminder, so it should work on other platforms as well.


johannesjo:
Thank you very much! I will test and review this and get back to you.

As a general remark: Since it is a new feature with a lot of changes usually it is better to discuss this in advance to avoid unnecessary work and changes. :)

onesvat:
Thank you for your feedback! You’re absolutely right—I should have opened an issue to discuss this beforehand. However, I thought implementing it first might better illustrate the idea in practice.

Initially, my goal was to ensure I stayed focused on planned tasks rather than getting distracted, but I now realize the fullscreen approach is too drastic and intrusive. I completely agree that a less aggressive solution—such as a notification and flashing the window, similar to the break reminder—would be more appropriate.

The new approach works as follows: When the configured time to wait before showing the tracking reminder banner is reached, a notification with sound will be triggered, similar to the break reminder. Then, every X minutes (which we should probably make configurable in the settings), the notification will sound again and the window will flash to ensure visibility.

Would you suggest making this behavior optional via a checkbox in the settings, or should it always be enabled by default?

If that sounds good to you, I’m happy to discard this PR and submit a new one with the necessary adjustments. Let me know what you think!

onesvat:
I’ve created a new PR (#4168) with a less intrusive approach.

PR #4156, opened 2025-03-18
Show remaining time on a task on the tray

panoramix360:
# Description

The idea is to show the remaining time of a task on the tray icon at the top.

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/3091

## When you play a task that has an estimate time, it shows the remaining time
image

## When you play a task that does not have an estimate time, it shows the time spent on the task until now
image

cc: @johannesjo @9z0

### DIFF
--- a/electron/indicator.ts
+++ b/electron/indicator.ts
@@ -114,25 +114,16 @@ function initListeners(): void {
 function createIndicatorStr(task: TaskCopy): string {
   if (task && task.title) {
     let title = task.title;
-    // let timeStr = '';
-    // let msg;
+    let timeStr = '';
     if (title.length > 40) {
       title = title.substring(0, 37) + '...';
     }
-    return title;
 
-    // if (task.timeSpent) {
-    //   const timeSpentAsMinutes = Math.round(task.timeSpent / 60 / 1000);
-    //   timeStr += timeSpentAsMinutes.toString();
-    // }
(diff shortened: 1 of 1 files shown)


panoramix360:
I created this PR as a draft so we can test on other environments and see what happens here.

Can you guys test on Linux if possible? I'll test this on Windows.

I didn't implemented the timer when the Pomodoro is enabled, do you think that's a good thing to have? I'll probably need to understand a little how can I add this to get the current pomodoro timer on the event process of the Store/dispatch.

panoramix360:
Just manage to add the Pomodoro timer on the tray timer when the pomodoro is enabled! yay!

image

The only issue is, the tray timer seem to be slower than the one inside the application, sometimes they are not totally in sync, I don't know if this is huge problem.

but since I never used the `ngrx`, @johannesjo can you review the code and see if I messed up?

Thank you!

johannesjo:
Thanks for the changes! On gnome there is still nothing shown unfortunately.  Maybe we could try using something like this:

https://www.npmjs.com/package/@girs/node-appindicator3-0.1

?

johannesjo:
Thank you!

johannesjo:
> I'll probably just not able to test on Linux, did you manage to see it something there?

No. :)

PR #4166, opened 2025-03-23
OpenProject Filter Fix

Rootkage:
# Description

I have change the "open-project-api.service.ts". I have erase the filter parameter.

## Issues Resolved

I cannot connect OpenProject with the actual Software. Without the filter I cane.

https://github.com/johannesjo/super-productivity/issues/4142

## Check List

- [ x] New functionality includes testing.

This is my first PR. So I hope this is allright. I only have erase the one part, and testet the software. But not heavy testing. But I didn't change anything else and the OpenProject connection is function now. So I think, it is all right.

King regards.


src/app/features/issue/providers/open-project/open-project-api.service.ts
```diff
@@ -52,19 +52,6 @@ export class OpenProjectApiService {
       {
         // see https://www.openproject.org/docs/api/endpoints/work-packages/
         url: `${cfg.host}/api/v3/projects/${cfg.projectId}/work_packages`,
-        params: {
```
johannesjo, line 55: Unfortunately this will stop to work once you have more than 100 (pageSize) workPackages. Maybe there is a better way to fix this?

19Topgun93: I will look into the filter API. I thought it is a cheap way. I'm not so into programming at the moment. I can read it and I learned it in my work, but I don't work as a developer. So I will try.

19Topgun93: I have tested a little and add the params with pagesize 100 and sortBy updatedAt. I haven't get the filters parameter get working. The Search doesn't filter at all. But for me that is better, then it doesn't work at all. Or my understanding from the filter parameter is bad. I don't know.  Maybe a Dev with more experience then I can make it.

johannesjo: Thank you for your effort! Currently I am very busy with some other features, so someone else picking this up, would be very welcome!

johannesjo:
I think, we can close this, right? Sorry, this slipped a bit under my radar :-s

PR #4168, opened 2025-03-24
feat(trackingReminderNotifications): add notificiations for tracking time reminders like take-a-break notification

onesvat:
# Description

In the previous PR (#4086), the solution was found to be too aggressive. This PR introduces a more balanced approach with the following improvements:

- **Notification on Reminder:** If the user enables "Notify when time tracking reminder is shown" in the settings, the application will send a notification when the time tracking reminder banner appears. The default notification frequency is currently hardcoded as DESKTOP_NOTIFICATION_THROTTLE (1 minute).
- **Sound Alert (Optional):** If a sound is selected for this purpose in the settings, it will play when the notification is sent.
- **Window Focus (Optional, Desktop Only):** If the user enables "Focus app window when reminder is active", the application will bring the window to the foreground when the reminder is displayed.

This approach ensures visibility while remaining less intrusive. Let me know if any further adjustments are needed!


johannesjo:
Thank you very much! :)

PR #4171, opened 2025-03-24
Fix #4046: Mulitple Same Name Tags being created.

### LINKED ISSUE #4046: When creating two duplicate new tags: only show one tag below it.
### Your Environment

- Version used: 11.1.3 Flatpak
- Operating System and version: Fedora Linux 41
- Desktop Environment: KDE Plasma 6.1.3

### Expected Behavior
When creating the duplicate tag #hello twice, it should only show #hello below the task once.

### Current Behavior
It shows the tag twice.

1. Create a new task with a new tag like #hello and #hello.
2. You should see two tags showing below the task.

Yes.

### Console Output
Not relevant.

MichaelSalis:
When you create a task you can use the pound sign to assign it tags, and if the tag dosent exist it creates it.
There was an error in the parsing of the user input that allowed the user can create multiple tags with the same name, which was not what was intended

# Description

I changed two short syntax files so that on display repeated new tags are not shown, and when trying to create it makes sure it dosent create duplicares.

## Issues Resolved

Trying to create a new task with multiple new tag of the same name will no longer work, resulting the in creation of a single tag with that name, with the fact being visually clear to the user

## Check List

- [Y ] New functionality includes testing.
- [N ] New functionality has been documented in the README if applicable. (non applicable)

### DIFF
--- a/e2e/commands/addTaskWithRepeatedTags.ts
+++ b/e2e/commands/addTaskWithRepeatedTags.ts
@@ -1,0 +1,19 @@
+import { NightwatchAPI } from 'nightwatch';
+import { cssSelectors } from '../e2e.const';
+
+const { ADD_TASK_GLOBAL_SEL, ROUTER_WRAPPER, TAGS } = cssSelectors;
+const CONFIRMATION_DIALOG = 'dialog-confirm';
+const TAG = `${TAGS} div.tag`;
+
+module.exports = {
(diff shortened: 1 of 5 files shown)


MichaelSalis:
Hello! I am not sure I can as that seems to be something from the base code, but I can try :)

johannesjo:
Thank you very much!

PR #4185, opened 2025-03-28
Fix #4130: shift+t to add subtasks to today

### LINKED ISSUE #4130: shift+T add task to today shortcut doesn't work with subtasks
### Your Environment

- Version used: 12.0.0, Windows Store
- Operating System and version: Windows 11

### Expected Behavior

pressing shift+T while a subtask is highlighted should schedule it for today

### Current Behavior

I'm taken to the "today" screen instead. I also tried remapping the global today shortcut to something else and shift+T still doesn't work on subtasks. It just does nothing

1. create a task with a subtask
2. highlight the subtask
3. press shift+T (or whatever the schedule for today shortcut is)

yes

### Console Output

none

### Error Log (Desktop only)
none

DuarteRamires:
# Description

Pressing shift+T while a subtask is highlighted should schedule it for today

## Issue Resolved

#4130

## Check List

- Removed parentId check in moveToToday() method to allow subtasks to be moved to today
- Removed parentId check in keyboard shortcut handler for moveToTodaysTasks
- Maintains existing behavior for regular tasks while extending functionality to subtasks

### DIFF
--- a/src/app/features/tasks/task/task.component.ts
+++ b/src/app/features/tasks/task/task.component.ts
@@ -739,7 +739,7 @@ export class TaskComponent implements OnDestroy, AfterViewInit {
 
   moveToToday(): void {
     const t = this.task();
-    if (t.projectId && !t.parentId) {
+    if (t.projectId) {
       this._projectService.moveTaskToTodayList(t.id, t.projectId);
       this.addToMyDay();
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much!

PR #4186, opened 2025-03-28
Fix #4124: Boards: Hide tasks from hidden projects

### LINKED ISSUE #4124: Boards: Hide tasks from hidden projects
### Your Environment

- Version used: 12.0.0 (scoop, Desktop)

### Expected Behavior

In the Boards view, e.g. the eisenhower matrix, the untagged task list (not urgent&not important; ToDo for Kanban) should not list those tasks that belong to projects hidden via project management view. Same should probably go for tasks in backlogs.

### Current Behavior

All tasks are shown no matter their belonging into backlogs or hidden projects.

Yes.

davidrr21:
#4124 
The visibility of each task aligns with the visibility of the corresponding project in the boards section.

# Description

The issue was that the toggleHideFromMenu function in the project service was never being called, and there was no interaction between the visibility of a project and its associated tasks.

# Implementation

I introduced a flag in tasks called unavailable, which changes according to the visibility of the respective project.

Previously, when changing the project's visibility in side-nav.component.html, the toggleProjectVisibility function in side-nav.component.ts was called, directly modifying the isHiddenFromMenu flag of the project.

Now, nav.component.html calls toggleProjectVisibility in side-nav.component.ts, which invokes the projectHideFromMenu service, triggering the toggleHideFromMenu action.

In the project reducer, the isHiddenFromMenu flag is updated, and in the tasks reducer, the unavailable flags of all tasks associated with the project are updated accordingly.

When the board selects tasks to display, it uses a new selector called selectAllAvailableTasksWithSubTasks, which filters only the tasks that are available.

### DIFF
--- a/src/app/features/tasks/store/task.reducer.ts
+++ b/src/app/features/tasks/store/task.reducer.ts
@@ -113,21 +113,6 @@ export const taskReducer = createReducer<TaskState>(
     });
   }),
 
-  on(TimeTrackingActions.addTimeSpent, (state, { task, date, duration }) => {
-    const currentTimeSpentForTickDay =
-      (task.timeSpentOnDay && +task.timeSpentOnDay[date]) || 0;
-    return updateTimeSpentForTask(
-      task.id,
-      {
-        ...task.timeSpentOnDay,
-        [date]: currentTimeSpentForTickDay + duration,
-      },
-      state,
-    );
-  }),
-
-  //--------------------------------
-
   // TODO check if working
   on(setCurrentTask, (state, { id }) => {
     if (id) {
--- a/src/app/core-ui/side-nav/side-nav.component.ts
+++ b/src/app/core-ui/side-nav/side-nav.component.ts
@@ -47,7 +47,6 @@ import {
   selectUnarchivedHiddenProjectIds,
   selectUnarchivedVisibleProjects,
 } from '../../features/project/store/project.selectors';
-import { updateProject } from '../../features/project/store/project.actions';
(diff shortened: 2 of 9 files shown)


src/app/features/tasks/store/task.reducer.ts
```diff
@@ -113,6 +113,24 @@ export const taskReducer = createReducer<TaskState>(
     });
   }),
 
+  on(toggleHideFromMenu, (state, { projectId, allTaskIds, isHiddenFromMenu }) => {
```
johannesjo, line 116: I'd rather avoid updating the task model itself for this, since it might complicate things further down the road, adds additional data size and logically the hidden state belongs more to the project model, than the task model.

johannesjo (changes requested):
Thank you very much for this! While I generally would like to see this feature, I think I'd prefer if the this would be done on the data mapping level rather, than changing the actual (persisted) model.

davidrr21:
Hi @johannesjo ,

Apologies for the delay in getting back to this – I’ve now pushed the requested changes to the PR. Please let me know if there’s anything else you’d like me to adjust.

Thanks again for your patience!

Best regards

johannesjo:
Thank you very much for the adjustments! It seems there are a lot of conflicts. Could you maybe have a look?

davidrr21:
Hi @johannesjo ,

The conflicts should now be fixed.
Please let me know if there’s anything else you’d like me to adjust.

Best regards

johannesjo:
Thank you.

PR #4187, opened 2025-03-28
Fix #4083 (tasks): resolve tag removal issue

### LINKED ISSUE #4083: After adding #Today to a task it gets added as text in the title afterwards
### Your Environment
- Version used: Flatpak 11.1.3
- Operating System and version: Fedora Linux 41
- Desktop Environment: KDE Plasma 6.3.2

### Expected Behavior
Like other tags it should not show in the title after adding the tag but as a label below it.

### Current Behavior
The tag gets added and works, but the text '#Today' gets added to the title in the task. Which is annoying to remove afterwards every time.

1. Create a task with #Today in it e.g. "Go on a walk #Today'.
2. You should see the text '#Today'  appended to the end of the title.

Yes, this issue has been here for a few versions now since I have been using the program.

### Console Output
Not relevant

mikemc03:
#4083
Fix issue where tags were not removed from the task title correctly.

# Description

This PR fixes an issue where tags were not properly removed from the task title when they were already assigned to the task. Previously, tasks that had tags both in the title (`#tag`) and in the `tagIds` field were not processed correctly, leading to incorrect title formatting.

## Changes

- Previously, only newly added tags were removed from the title. The logic in short-syntax.ts has been updated to ensure all existing tags in the title are removed correctly. Now, if a task is created with a tag matching the current view, it will be automatically removed from the title.

- Added unit tests in `short-syntax.spec.ts` to confirm the fix.

### DIFF
--- a/src/app/features/tasks/short-syntax.spec.ts
+++ b/src/app/features/tasks/short-syntax.spec.ts
@@ -489,6 +489,82 @@ describe('shortSyntax', () => {
 
       expect(r).toEqual(undefined);
     });
+
+    it('should remove tag from title if task already has tag', () => {
+      const t = {
+        ...TASK,
+        title: 'Test tag #testing',
+        tagIds: ['testing_id'],
+      };
+      const r = shortSyntax(t, CONFIG, [
+        ...ALL_TAGS,
+        { ...DEFAULT_TAG, id: 'testing_id', title: 'testing' },
+      ]);
+
+      expect(r).toEqual({
+        newTagTitles: [],
+        remindAt: null,
+        projectId: undefined,
+        taskChanges: {
+          title: 'Test tag',
+        },
+      });
+    });
+
+    it('should create new tag and remove both from title if task already has one given tag', () => {
+      const t = {
+        ...TASK,
+        title: 'Test tag #testing #blu',
+        tagIds: ['blu_id'],
+      };
+      const r = shortSyntax(t, CONFIG, [
+        ...ALL_TAGS,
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much for this! Not sure if I agree though. I don't think writing the short syntax for a tag should ever remove something. Doesn't feel right to me from a UX perspective.

johannesjo:
I am also a bit confused. Wasn't the original issue about something a little bit different?

@tegridyfarmer is it alright if I tag you here? I think you're deeper in the subject than I am!

mikemc03:
I don't think it was. The issue, as I understood it, was that when creating a task with '#Today', the tag was correctly added, but the text '#Today' remained in the task title. I investigated a bit further and found out this issue was not specific to '#Today', but affected any tag when creating a task within its corresponding tag view.

This was happening more often with the tag '#Today' because it's the name of the default view when you first open the app. If a task is created in a specific tag view, the corresponding tag is automatically added to the task before parsing any tags in the title. In this case, keeping #Today in the title doesn’t make sense, as it duplicates the assigned tag.

The same applies to every other tag. For example, if a task with #Someday is created while in the ‘Someday’ page, the text '#Someday' will stay in the title unnecessarily.

I hope I was able to explain myself clearer. Let me know if you’d like any further details, and I apologize if the initial description/change log was not explicit enough!

johannesjo:
My bad! Sorry @mikemc03 ! I was misreading one of the unit tests and came to the conclusion that the new code removes the tag from the task and not just from the title.

I am merging this! Thank you very much for the fix!

Issue #4048, opened 2025-02-23
(Android) Disabling animations makes it so sidebar at left can't be hidden by touching unfocused right side


SeerLite:
### Steps to Reproduce

1. Disable animations in settings
2. Go back to Today
3. Open sidebar
4. Try to close it by pressing right side (non-sidebar)

### Meta Info
META: SP11.1.3_L2A __ Browser – en-US – Linux armv81 – en-US – UA:Mozilla/5.0 (Linux; Android 13; Redmi Note 8 Build/TQ3A.230901.001; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/118.0.0.0 Mobile Safari/537.36

SeerLite:
The compact setting breaks too. Without animations it doesn't make the main content move to the right and so the left sidebar can't be used at all (it's covered by main central part).

UnconsciousBlaringHuman:
I've also experienced this, both on Android and Windows (when the window is narrow, which is how I use it 99.9% of the time)

palmerusaf:
This issue is resolved with the v15 redesign.

johannesjo:
Thanks @palmerusaf !

PR #4203, opened 2025-04-03
Fix #4158 #4048: sidebar buggy when animations disabled

### LINKED ISSUE #4158: Left sidebar rendering is buggy when animations disabled
### Your Environment

- Version used: 12.0.1 AUR
- Operating System and version: Arch
- Desktop Environment: HyprLand 0.47.2
- Browser Name and version: Firefox 136.0.2

### Expected Behavior

When the side bar is toggled it correctly resizes the main view and appears with the correct z axis.

### Current Behavior

When animations are disabled hiding the side bar doesn't expand the main view. Additionally, un-hiding the side bar fails to shrink the main view and the sidebar appears behind the main view. The rendering can be corrected by resizing the window. I've experience this with the flatpak install on Pop OS as well. This leads me to believe it may be related to how tiling managers instantly resize windows. Maybe when on a tiling desktop environment with animations disabled the UI is skipping a critical break-point event that leaves the sidebar in a inconsistent state. The bug appeared sometime towards the end of V11. If I get some free time I will try get you an exact commit hash with git bisect.

1. Disable animations
2. Instantly resize the window somehow, such as with a tiling manager
3. Toggle the sidebar

Reproducible on flatpak, deb, and the browser. 

### Console Output

Image

### Error Log (Desktop only)

No errors just the console log above.

Image

palmerusaf:
This reverts commit 71183da70ec06defb860b2946734f467a2d1e633.

# Description

This PR fixes a regression that causes sidebar layout issues when animations are disabled. After some research I know for sure the regression happened inside the angular/components library. The PR simply downgrades those dependencies by reverting the commit that caused the regression. I haven't pinpointed where in the angular library this happened, but reading the logs between versions points to [this](https://github.com/angular/components/commit/a40b5a7998c33f8dce5ec1f4ee9609a1974bbe10#diff-77060c6c007698d3a0a4cbbb690a8dc7e986efde563c012a7c96d118ef150ee1) refactor being a likely candidate.

## Issues Resolved

-  #4158
-  #4048
## Check List

- [n/a] New functionality has been documented in the README if applicable.


palmerusaf:
I noticed one of the e2e tests is failing. I had some trouble with this test locally as well. It was even failing on master branch. I was able to get all e2e passing by rerunning. I also did a quick test build on android with no issues. Additionally, I've been using a desktop build with this commit for a week with no issues. 
image

johannesjo:
Sorry about the e2e tests. This happens sometimes depdending on the time of day and the timezone the server is in (I think).

About the fix: I suppose upgrading to 19.2.8 instead, does not fix the issue?

palmerusaf:
@johannesjo I just double checked and unfortunately it doesn't. I'm sure you don't want to be stuck on 19.0.3 indefinitely. Maybe a proper fix would be to trigger a resize event manually? It seems like they may have changed the algo to rely more on resize events. It does self correct if you resize the window.

image
image

johannesjo:
This sounds like something they should fix on their end, though I am not sure how one would formulate the text for an issue in a way that it gets any attention.

I think triggering a resize event manually and only doing so if animations are disabled might be the best course of action. I am hesitant to merge this downgrade, since I can't rule out that it will introduce some problems with  other components.

Could you possibly provide a PR for that?

palmerusaf:
@johannesjo I agree. I've considered submitting an issue with a minimal reproducible example over at the component library. As far as another PR, I can give it a shot but it may take me a while, as I'm still unfamiliar with angular and I'm taking a lot of classes at the moment.

johannesjo:
That would be great! Let me know if you have any questions!

palmerusaf:
This issue is resolved in v15.

Issue #4272, opened 2025-04-20
🚨 Importing data failed: DataValidationFailedError - monday-sunday undefined, expected boolean


acheong08:
### Version Used

7eaa7a48

### Expected Behavior

dataRepair should fix migration issues.

Not sure whether that's the correct behavior so I'll hold off on opening a pull request for now.

### Current Behavior

```
'[{"path":"$input.taskRepeatCfg.entities.uKEOm2Pxg6hluDOfuCDEs.monday","expected":"boolean"},{"path":"$input.taskRepeatCfg.entities.uKEOm2Pxg6hluDOfuCDEs.tuesday","expected":"boolean"},{"path":"$input.taskRepeatCfg.entities.uKEOm2Pxg6hluDOfuCDEs.wednesday","expected":"boolean"},{"path":"$input.taskRepeatCfg.entities.uKEOm2Pxg6hluDOfuCDEs.thursday","expected":"boolean"},{"path":"$input.taskRepeatCfg.entities.uKEOm2Pxg6hluDOfuCDEs.friday","expected":"boolean"},{"path":"$input.taskRepeatCfg.entities.uKEOm2Pxg6hluDOfuCDEs.saturday","expected":"boolean"},{"path":"$input.taskRepeatCfg.entities.uKEOm2Pxg6hluDOfuCDEs.sunday","expected":"boolean"}]'
```

### Steps to Reproduce

[super-productivity-backup.json](https://github.com/user-attachments/files/19826904/super-productivity-backup.json)

Import the above backup

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

PR #4275, opened 2025-04-21
fix: Import fails when weekdays not defined

acheong08:
# Description

Fix v0 -> v2 migration where monday-sunday expects boolean but gets undefined

## Issues Resolved

- #4272

Tested a bit - don't think it'll break anything. Realized that just setting it to true will cause unintended behavior when problems aren't here, so using ?? operator whatever its called in typescript to just reuse value if it exists, only set to true if it's missing


johannesjo:
Thank you very much! Just now made the very same fix :D, but yours is better, so I am merging this :)

PR #4290, opened 2025-04-29
Added mini-progress bar in planner

loviuz:
# Description

I added a mini progress bar in the planner above each day to see as a percentage how loaded the day is. if the load is under 80% of the available time the bar is green, between 80 and 95% orange, over 95% is red. the default available time is 8 hours, but if you enable work hours and lunch break that time will be used.

immagine


johannesjo:
@CodiumAI-Agent /review

johannesjo:
@loviuz linting is failing. Can you have a look? Should be a quick fix.

loviuz:
> @loviuz linting is failing. Can you have a look? Should be a quick fix.

Fixed. It seems it was a linting problem coming from previous PR.

johannesjo:
Very very nice! Thank you very much! <3

Issue #4301, opened 2025-05-04
🚨 SCHEDULE TASK BUG


Dzigiii:
### Version Used

12.0.2

### Expected Behavior

For the task, Schedule task [s] opens up calendar, "Time" and "Remind at". Set"Remind at" to "Never" and "Time" to whatever time I want, after clicking "Schedule" button down below, the task will be scheduled at specified time.

### Current Behavior

For the task, Schedule task [s] opens up calendar, "Time" and "Remind at". If I set "Remind at" to "Never" and "Time" to whatever time I want, after clicking "Schedule" button down below, the task won't be scheduled.

### Steps to Reproduce

Right click the already made task,
Schedule task [s],
Set Reminder to "Never"
Set Time to whatever
Click schedule button down below
It says that task has been scheduled (it is not)

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell
VM154:1 Console was cleared
undefined
action-logger.reducer.ts:13 [Task] Schedule {task: {…}, plannedAt: 1746349200000, remindAt: undefined, isMoveToBacklog: false, type: '[Task] Schedule'}
action-logger.reducer.ts:13 [Planner] Remove Task From Days {taskId: '1QNeHkpiJNY7NihY2IC8R', type: '[Planner] Remove Task From Days'}
action-logger.reducer.ts:13 [Persistence] Save to DB {dbKey: 'task', data: {…}, type: '[Persistence] Save to DB'}
action-logger.reducer.ts:13 [Persistence] Save to DB {dbKey: 'planner', data: {…}, type: '[Persistence] Save to DB'}
action-logger.reducer.ts:13 [Task] Update Task Tags {task: {…}, newTagIds: Array(1), type: '[Task] Update Task Tags'}
action-logger.reducer.ts:13 [Persistence] Save to DB {dbKey: 'planner', data: {…}, type: '[Persistence] Save to DB'}
action-logger.reducer.ts:13 [Persistence] Save to DB {dbKey: 'tag', data: {…}, type: '[Persistence] Save to DB'}
action-logger.reducer.ts:13 [Persistence] Save to DB {dbKey: 'task', data: {…}, type: '[Persistence] Save to DB'}
sync-trigger.service.ts:208 __trigger_sync__ task
sync-trigger.service.ts:208 __trigger_sync__ planner
sync-trigger.service.ts:208 __trigger_sync__ planner
sync-trigger.service.ts:208 __trigger_sync__ tag
[...]

PR #4309, opened 2025-05-06
Fixes  SCHEDULE TASK BUG #4301

Jamesau0:
# Description

Allows user to shedule a task when remindAt is set to "never". 

## Issues Resolved

Fixed Handling of the "Never" Reminder Option (Previously returned undefined for "Never". 
Fixed object being directly modified when it was immutable.


Jamesau0:
Will pr again.

johannesjo:
@Jamesau0 just a heads up: I made some big changes to how all the due stuff works.

PR #4351, opened 2025-05-14
Update it.json

FromColleWithLove:
Corrected some dialogues that contained Georgian rather than Italian translation.

# Description

This translates some text that is shown in the app.

## Issues Resolved

Presence of Georgian rather than Italian translation in some dialogues.


johannesjo:
Thank you very much!

PR #4420, opened 2025-05-20
Fixed and added new italian translations

loviuz:
I have just added and fixed some italian translations.


loviuz:
I have noticed an error on automatic tests, but it does not depend on translations. Can you merge the same?
Thanks

johannesjo:
Yes. Sorry about this. This one test is a bit flakey depending on the time of day and the browser environment. :/

johannesjo:
Thank you very much!

PR #4445, opened 2025-05-23
Add an option to hide the countdown timer on the Status Bar (MacOS)

panoramix360:
# Description

Adding the option to hide the countdown timer through the Misc Settings.

image

image

If the user has it disabled, it will appear just as the task title:
image

If the user has it enabled, it will appear as usual:
image

The default config will be `true`.

@johannesjo, let me know your thoughts about this change.

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/4410

## Check List


johannesjo:
Thank you!

PR #4555, opened 2025-06-10
feat: webdav sync default settings in env variables of container

### LINKED ISSUE #4545: 💡 Add Environment variables or config file for WebDav Sync as default in settings
### Problem Statement

I followed the Readme file to install super-productivity via docker compose with webdav support. It works great. 

The only problem is that I sometimes work in incognito and the data does not presist. To pull the data from webdav I need to re enter the creds which is fine but it is a few text/check boxes. Since I always type in the same values in the field, won't it be helpful if the values are taken from environment variables?

### Possible Solution

List of environment variables I suggest:

- WEBDAV_BASE_URL (defaults = "")
- WEBDAV_USER (defaults = "")
- WEBDAV_PASSWORD (Though I usually wont recommend, it might be useful for someone having a proxy auth like authelia, traefik, etc)(defaults = "")
- WEBDAV_SYNC_FOLDER (defaults = "super-productivity")
- WEBDAV_SYNC_INTERVAL(defaults = "1m")
- WEBDAV_ADV_COMPRESSION (defaults = "false")
- WEBDAV_ADV_E2E (defaults = "false")

Additional Info:
- Default for WEBDAV_BASE_URL can be <app_url>/webdav/ if WEBDAV_BACKEND is enabled

Maybe load a config.yml like how webdav does with `webdav.yml`.

```yaml
... # with other settings if this becomes a thing
webdav:
  base_url: ""
  user: ""
  password: ""
  sync_folder: "super-productivity"
  advanced:
    compression: false
    e2e: false
```

Personally I like the config file approach since I don't like passwords as environment variable.

### Additional context

@kkoyung Just tagging you since you made pull requests #3534 #3688 and you may know more on this.

kkoyung:
# Description

In the docker setup, allow admin to provide the default values for WebDAV settings in the "Sync" section of the "Settings" page, via setting some environment variables of the docker container. This could be used for single user instances to pre-fill those settings. The available environment variables are:

- WEBDAV_BASE_URL=<base-url>
- WEBDAV_USERNAME=<username>
- WEBDAV_SYNC_FOLDER_PATH=<sync-folder-path>
- SYNC_INTERVAL=<integer-in-minutes>
- IS_COMPRESSION_ENABLED=<true-or-false>
- IS_ENCRYPTION_ENABLED=<true-or-false>

These configurations are provided as environment variables after the build process of the app, so they have to be loaded at runtime, instead of build process. This can be achieved by the following way:

1. The environment variables are loaded into the docker container.
2. The container starts with `docker-entrypoint.sh`.
3. `docker-entrypoint.sh` converts the variables into a JSON file placed at `assets/sync-config-default-override.json`.
4. `sync-config.service.ts` fetches `assets/sync-config-default-override.json` dynamically at runtime to override the default values from `default-global-config.const.ts`.

## Issues Resolved

#4545

## Check List


kkoyung:
@BaccanoMob

As mentioned above, the environment variables are first converted into a json file, and served at `/assets/sync-config-default-override.json`. That means you can mount your own json file to `/usr/share/nginx/html/assets/sync-config-default-override.json` in the container. (The mounted json won't be over-written as long as you don't set any of environment variables mentioned above.)

Therefore, you can actually provide the password and the encryption key via the mounted json, using the json structure below. You may find this trick useful in your use case.

**WARNING**: It also means your password is served as plain text. It might be okay if you only bind your container to `127.0.0.1`. However, it is quite danger in general, and highly not recommended. So, you may consider this as undocumented hack if you fully understand the risk.

```json
{
  "isEnabled": true,
  "isCompressionEnabled": true,
  "isEncryptionEnabled": true,
  "encryptKey": "your-encryption-key",
  "syncProvider": "WebDAV",
  "syncInterval": 10000,

  "webDav": {
    "baseUrl": "https://your-domain/webdav/",
    "userName": "your-username",
    "password": "your-password",
    "syncFolderPath": "super-productivity"
  }
}
```

johannesjo:
Thank you very much for this @kkoyung !!!

kkoyung:
Cheers! @johannesjo

Issue #4286, opened 2025-04-26
💡 Schedule Tasks more than 1 week ahead


OfficiallyCrazy:
### Problem Statement

Thank you for creating this awesome software. 
I often wonder if I have any doctors appointment or family members birthdays coming up. The current capabilities of the software only allows me to see tasks only up to 1 week ahead. 

Adding a way to switch from week view to  monthly view, would help me a lot. 

### Possible Solution

Adding a way to switch from week view to  monthly view.

### Describe alternatives you've considered

Currently I am mixing Google calendar and Super Productivity.

DuarteRamires:
Hello, I'd like to contribute more to this project by working on this feature. I would be happy to implement the proposed monthly view toggle to allow scheduling tasks beyond one week. Would it be possible to assign me to this issue?

aripollak:
Does the "Repeat/Scheduled" view at the bottom of the sidebar not show all upcoming scheduled tasks for you?

![Image](https://github.com/user-attachments/assets/80559db8-d82a-4e1f-ac0a-381e86f61697)

AnHumanLearner:
@OfficiallyCrazy this feature has been implemented, right ? So the issue can be closed ?

lanesawyer:
@johannesjo I believe this one can be closed, there is a month view in Schedule.

PR #4578, opened 2025-06-13
Implement month view mode for task scheduler

davidrr21:
# Description
#4286 
We added a toggle button in the schedule section to switch between week and month views. This allows users to see all tasks scheduled across the full month, improving visibility and planning.

For example: if a user creates a task for June 15 in week view, switching to month view will show that
task in the correct calendar cell. The view update is dynamic and works regardless of where the task
is created (week view, today section, etc).

To support this feature, we extended the calendar
component to support multiple views,updated
the UI layout to render monthly grids, and
refactored state management to support a
broader date range.

This feature addresses previous limitations
identified through user feedback and makes
the task scheduler more useful as a
standalone planning tool.


johannesjo:
Thank you very much! I hope I get to review this next Friday! Busy week ahead.

src/app/util/date-str-to-utc-date.ts
```diff
@@ -13,9 +13,6 @@ export const dateStrToUtcDate = (dateStr: string): Date => {
   if (!dateStr) {
     return new Date();
   }
-  // const [year, month, day] = dateStr.split('-').map(Number);
-  // return new Date(Date.UTC(year, month - 1, day));
-  const localDate = new Date(dateStr);
-  // eslint-disable-next-line no-mixed-operators
-  return new Date(localDate.getTime() + localDate.getTimezoneOffset() * 60000);
+  // Just return local midnight for the date string
+  return new Date(dateStr);
```
johannesjo: As mentioned in the comment above: We can't just simply use `new Date()` here

src/app/features/schedule/schedule/schedule.component.ts
```diff
@@ -553,4 +638,68 @@ export class ScheduleComponent implements AfterViewInit, OnDestroy {
     parts[0] = `translate3d(${newNumber}`;
     return parts.join(',');
   }
+
+  getEventDayStr(ev: ScheduleEvent): string | null {
+    // Calendar events
+    if (ev.type === SVEType.CalendarEvent && ev.data && (ev.data as any).start) {
```
johannesjo: Casting as `as any` is not good, since it leads to potential bugs and makes refactoring a lot harder.

johannesjo (changes requested):

johannesjo:
Very cool stuff! There are some problems with the UI though:

Screenshot from 2025-06-16 18-46-34
Screenshot from 2025-06-16 18-47-03

DuarteRamires:
@johannesjo We've implemented the requested changes. Please let us know if any further adjustments are needed

johannesjo:
Thank you! I will check this next Friday!

johannesjo:
Thank you very much once more! This is an awesome addition! 

Maybe styling could be a bit optimzed, but we can add on this later.

Issue #3994, opened 2025-02-12
Provisioning a client with settings <> OpenProject


sausin:
## Problem Statement

When a new client needs to connect to OpenProject quite a few steps are required before the integration is fully functional. (but it works beautifully!!). We have quite a few non-technical audience and without sessions, it would not work to get their integration functional.

## :grey_question: Possible Solution

It would be absolutely fantastic if a payload could be sent to superproductivity or if settings could be uploaded. This would enhance the experience superbly.

## :arrow_heading_up: Describe alternatives you've considered

Working with remote support at the moment.

## :heavy_plus_sign: Additional context

It would probably be an easy extension to get the same feature working for any other integrations as well

johannesjo:
You could create a minimal config on a fresh super productivity instance (e.g. using the webapp in inkognito tab: https://app.super-productivity.com) and export this and re-import on every other machine?

sausin:
This will do, thanks @johannesjo 

![Image](https://github.com/user-attachments/assets/9c202669-098b-441d-9b82-f498fe38f6f9)

sausin:
> You could create a minimal config on a fresh super productivity instance (e.g. using the webapp in inkognito tab: https://app.super-productivity.com) and export this and re-import on every other machine?

Just getting back here - is there a way for this to be achieved in the app on phone? Or the config protected by the app policies?

sausin:
Closed in #4584

PR #4584, opened 2025-06-15
feat: Import config from URL #3994

ebbbang:
# Description

This pull request adds a new feature that allows config to be imported from a valid URL.
The changes are made in file-imex component, and added two new components dialog-import-from-ul and dialog-confirm-url-import

The feature works in two ways
- You can use the new *Import From URL* button
- You can pass the encoded URL with the query parameter importFromUrl. (Allows for provisioning a working setup for clients/users )

In the second case the user is notified that the data would be overwritten and the domain that the import is coming from. The import only happens after the user approves it.

I've also added/updated tests for all three components.
I've also added required translations for all the languages.

UI for user import dialog
Screenshot 2025-06-05 at 11 57 41 AM

UI for confirmation dialog when importing through direct URL
Screenshot 2025-06-05 at 11 57 19 AM

Thanks for the awesome project again!

## Issues Resolved

It introduces the feature mentioned in https://github.com/johannesjo/super-productivity/issues/3994

## Check List


johannesjo:
Thank you very much! Cool stuff! I hope I get to review this next Friday!

PR #4595, opened 2025-06-17
fix: start nginx via its built-in entrypoint script

kkoyung:
# Description

Our custom `docker-entrypoint.sh` starts nginx by directly calling `nginx -g "daemon off;"`. However, we need the `docker-entrypoint.sh` built-in in the nginx docker image to convert `default.conf.template` to the actual nginx config file. Without going through nginx's `docker-entrypoint.sh`, the resultant docker image fallback to the default nginx configuration, which doesn't forward traffic to webdav backend server.

This patch fixes this by calling the nginx's `docker-entrypoint.sh` at the end of our custom `docker-entrypoint.sh`.

## Issues Resolved

The first issue in https://github.com/johannesjo/super-productivity/issues/4545#issuecomment-2974843258

## Check List

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,12 +44,14 @@ COPY --from=build /app/dist/browser /usr/share/nginx/html
 # copy nginx config
 COPY ./nginx/default.conf.template /etc/nginx/templates/default.conf.template
 
-# copy entrypoint script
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much!

PR #4596, opened 2025-06-17
fix: preserve default sync folder path when no overriding

kkoyung:
# Description

The spread operator `...` replaces `baseConfig.webDav` with `defaultOverride.webDav` entirely, instead of merging them. Hence, the default sync folder path in `baseConfig.webDav` is wiped out.

This patch fixes this issue by manually providing a merged `webDav`.

## Issues Resolved

The second issue in https://github.com/johannesjo/super-productivity/issues/4545#issuecomment-2974843258

## Check List


johannesjo:
Thank you!!!

PR #4673, opened 2025-06-30
fix: remove line breaks on paste in task title (#4669)

### LINKED ISSUE #4669: 🚨 Small Bug: Copy-pasting into app sometimes lets you maintain line breaks
### Version Used

v14.0.0

### Expected Behavior

Line breaks should be removed when pasting into the task name, to keep consistency with how the app deals with it in other areas.

### Current Behavior

When adding a new task, line breaks are filtered out correctly immediately after pasting. However, when editing the name of the task, sometimes the app lets you keep the line break on it.

https://github.com/user-attachments/assets/6d4a8beb-35d4-4f34-8042-5e5e31deeaaa

This example was a simplification of the issue, but some apps / sites add a invisible line break after the copied text, so this issue might appear more often.

(TBH it doesn't seem like this is a deal breaker, but I can imagine this having some potential issues if any other area of the app is not expecting line breaks at all on this field, so I am creating this issue just in case).

### Steps to Reproduce

1. Open any Note-taking app to test this out
2. Write any sentence in two lines, with an Enter line-break. Example:

```
this is 
a test
```

3. Copy this text
4. Add a new task in Super Productivity
5. Paste the text on the name field of the task, and save (pressing Enter or clicking away from the task field)
You will notice that the line breaks got stripped down and the text is in a single line
7. Select the name field of the task again, and paste the text again on it.
You _should_ notice that the text is now in multiple lines.

### Console Output

```shell

```

Sahil-Sevda:
# Description

This PR fixes an issue where line breaks were incorrectly preserved when pasting into the task title input field during editing. It ensures that any pasted text is cleaned to remove \n and \r, maintaining a consistent single-line format across the application.

## Issues Resolved

Fixes #4669

## Check List

 -New functionality includes testing (manual UI testing done for paste behavior).
 -New functionality has been documented in the code via comments (no README change needed).

##Demo
Attached below is a screen recording of the manual testing, demonstrating the updated paste behavior in the task title input.
[demo line breaks.webm](https://github.com/user-attachments/assets/cb094af9-27c7-48a1-a7a1-884cf3659965)

### DIFF
--- a/src/app/ui/task-title/task-title.component.ts
+++ b/src/app/ui/task-title/task-title.component.ts
@@ -98,7 +98,31 @@ export class TaskTitleComponent {
   }
 
   updateTmpValue(value: string): void {
-    this.tmpValue = value;
+    this.tmpValue = this._cleanValue(value);
+  }
+
+  handlePaste(event: ClipboardEvent): void {
+    event.preventDefault();
+
+    const pastedText = event.clipboardData?.getData('text/plain') || '';
+    const cleaned = this._cleanValue(pastedText);
+
+    const textarea = this.textarea().nativeElement;
+    const start = textarea.selectionStart || 0;
+    const end = textarea.selectionEnd || 0;
+
+    const currentVal = textarea.value;
+    const newVal = currentVal.slice(0, start) + cleaned + currentVal.slice(end);
+
+    // Update both textarea and tmpValue
+    textarea.value = newVal;
+    this.tmpValue = newVal;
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much for this! Did you test the feature in different browsers? Unfortunately they differ quite a bit in behavior for pasting text.

Sahil-Sevda:
You're welcome! Yes, I tested the feature in Brave, Edge, and Chrome — it works consistently across all three browsers.

Issue #4644, opened 2025-06-24
🚨 Adding @date incorrectly adds task to today when @date is listed last


jpgaubier:
### Version Used

v13.1.5 snap

### Expected Behavior

Expected behavior: Adding a task with '@date' will schedule the task for '@date.'

### Current Behavior

Current behavior: Adding a task with '@date' when '@date' is the *last* entry will schedule the task for today, regardless of '@date' entry, at a time that is MM minutes in the future, where MM minutes was entered as the task duration.  E.g., 'Test this task 15m +Personal @tomorrow' gets scheduled at today in 15 minutes' time.

This bug seems to appear only when '@date' is the _last_ value in the task entry.  

### Steps to Reproduce

1. Add task (Shift+A)
2. Enter {Task} {minutes} {+project} {@date}    (_date is last_)
3. Press enter

![Image](https://github.com/user-attachments/assets/cd18a21a-f259-49f3-bb12-c17ecc74f399)

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

bytrangle:
Should this be closed?

jpgaubier:
> Should this be closed?

Yes, I believe this is fixed now. Thank you.

PR #4689, opened 2025-07-01
fix: Adding @date incorrectly adds task to today when @date is listed…

danielpetrykowski:
… last #4644

# Description

Fix for 🚨 Adding @date incorrectly adds task to today when @date is listed last #4644

## Issues Resolved

The root cause of the issue was passing the full input text directly from the task creation field. For example, when entering a phrase like "task 1h @tomorrow", the first recognized value was "1h". As a result, the scheduling logic added 1 hour to the current date, and the @tomorrow part was ignored.

## Check List


johannesjo:
Thank you very much!!

PR #4700, opened 2025-07-04
Fix: remove characters not intended to be there

mtelgkamp:
# Description

Removed some characters from masking text.

Also removed masking of end of `strong` HTML element.

Improved translation quality (make it more similar to English version).


mtelgkamp:
Screenshot from the current version:
Unwanted characters in new dialog

johannesjo:
Thank you very much!

Issue #4328, opened 2025-05-12
💡 Task Deadlines


johannesjo:
### Problem Statement

As a User I want to be able to set a deadline for my tasks in addition to just scheduling them.

### Possible Solution

Add task deadlines as a new feature.

Testorakel:
This information would also be helpful for the calendar view.

Mustache-Games:
We should be able to assign every task a start date and a due date.

Pononot:
First of all i think seperating due date and scheduling is a good idea. I also like the idea of being able to set a start and end date, you could even hide the task in backlog before the start date. I like [this](https://github.com/johannesjo/super-productivity/discussions/4521#discussioncomment-14220499) idea.

I think there's 2 features i would love to see in combination with this:

1. Timeline View for Task Deadlines
Description:
A visualization feature that displays tasks on a timeline based on their scheduled/deadline dates, with filtering options by:

- Projects
- Tags
- All tasks (global view)

Use Case:
This would provide a visual overview of work density and help with capacity planning. Being able to see when tasks are clustered or spread out across time would make it easier to:

- Identify overcommitted periods
- Balance workload across projects
- Plan new task scheduling more realistically

Example:
Similar to a Gantt chart or calendar heatmap showing task distribution over time.

2. Partial Task Scheduling
Current Limitation:
When scheduling a task, the entire task gets scheduled for a single day/time. For longer tasks, this makes the scheduling feature impractical since you may want to work on the task incrementally. There is the workaround of resizing the task in scheduling view but this changes the estimated time of the task itself.

Proposed Solution:

Allow scheduling a portion of a task's estimated time (e.g., schedule 1h of work on a 5h task)
Enable scheduling the same task across multiple days/sessions
Track progress: show how much time has been scheduled vs. total estimate

Use Case:
Many tasks require multiple work sessions across different days. Being able to schedule partial work sessions would make the scheduling feature more practical for:

Breaking down large tasks into manageable chunks
Realistic daily planning (not blocking entire days)
Better integration with calendar/time-blocking workflows

Example:

Task: "Garden work" (estimated 8h)
Schedule: 2h on Monday, 3h on Wednesday
UI shows: "5h/8h scheduled" with remaining time still available to schedule

Let me know if you need additional information to understand.

Mustache-Games:
I believe a strong starting point would be to introduce a "Due at" button, modeled after the existing "Planned at" button, which would enable users to easily set a specific due date and time for tasks.

[...]

PR #4753, opened 2025-07-13
Feat/#4328 Deadline on the tasks

danielpetrykowski:
# Description

This is enhancement for #4328 - the deadline field on task has been added

Implemented:
- Added a new deadline field to tasks
- Handle of the new field with the calendar (similar to schedule)
- Saving of the deadline field
- If a deadline is set, the corresponding date is marked in the schedule calendar

Note: Integration with reminders is not yet implemented.

## Check List


johannesjo:
Hello! Thank you very much for this! Looks great!  

I'll give you some more detailed feedback in a sec. But first one more fundamental question first:

This adds setting a deadline, which is great in itself, but currently it is not much more than just a UI field. What do you think would be a good way to warn the user about upcoming deadlines?  Personally I think adding a section to the today page like it is done for overdue would be a good solution. What do you think? Do you have additional ideas for this?

src/app/features/planner/dialog-deadline-task/dialog-deadline-task.component.ts
```diff
@@ -0,0 +1,226 @@
...
+    });
+    setTimeout(() => {
+      this._focusInitially();
+    }, 300);
+  }
+
+  private _focusInitially(): void {
+    if (this.selectedDate) {
+      (
+        document.querySelector('.mat-calendar-body-selected') as HTMLElement
```
johannesjo, line 114: We probably should use elementRef for this instead.

src/app/features/planner/dialog-deadline-task/dialog-deadline-task.component.ts
```diff
@@ -0,0 +1,226 @@
...
+      )?.parentElement?.focus();
+    } else {
+      (
+        document.querySelector('.mat-calendar-body-today') as HTMLElement
```
johannesjo, line 118: For this too.

src/app/features/planner/dialog-deadline-task/dialog-deadline-task.component.ts
```diff
@@ -0,0 +1,226 @@
...
+      this.selectedDate = dateStrToUtcDate(this.data.targetDay);
+    }
+
+    this.calendar().activeDate = new Date(this.selectedDate || new Date());
+    this._cd.detectChanges();
+
+    setTimeout(() => {
+      this._focusInitially();
+    });
+    setTimeout(() => {
```
johannesjo, line 106: We should save this timeout and clear it on destroy, since otherwise the app might error when the dialog is clicked away very quickly.

src/app/features/planner/dialog-deadline-task/dialog-deadline-task.component.ts
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    this.close(true);
+  }
+
+  private async _setDeadline(newDay: string): Promise<void> {
+    this._taskService.update(this.data.task.id, { deadline: newDay });
+  }
+
+  quickAccessBtnClick(item: number): void {
```
johannesjo, line 194: Maybe for deadlines we don't need the quick access stuff. This would also distinguish the dialog more from the schedule dialog.

src/app/features/tasks/task/task.component.ts
```diff
@@ -183,6 +182,24 @@ export class TaskComponent implements OnDestroy, AfterViewInit {
         !environment.production)
     );
   });
+  isDeadlineWarn = computed(() => {
```
johannesjo, line 189: the is- prefix should only be used with booleans and not with strings.

src/app/features/tasks/task-detail-panel/task-detail-panel.component.html
```diff
@@ -150,6 +150,39 @@
...
+      <span>{{ T.F.TASK.ADDITIONAL_INFO.DEADLINE | translate }}</span>
+    </ng-container>
+
+    <ng-container input-value>
+      @if (task.deadline) {
+        <div
+          class="reminder-value red-highlight"
+          [ngClass]="{
+            'red-highlight': isDeadlineExceeded,
+            'orange-highlight': isDeadlineForLessThan7Days,
```
johannesjo, line 172: maybe we should use the accent color for this instead of orange.

src/assets/i18n/ar.json
```diff
@@ -1079,6 +1079,7 @@
         "REMINDER": "تذكير",
         "REPEAT": "كرر",
         "SCHEDULE_TASK": "جدولة مهمة",
+        "DEADLINE": "Deadline",
         "SUB_TASKS": "المهام الفرعية ({{nr}})",
```
johannesjo, line 1082: We need to remove all the english translations from the non en.json files.

johannesjo (changes requested):

00sapo:
Hello, is there any plan to work on this issue? I'm especially interested for my AutoPlan plugin

johannesjo:
> Hello, is there any plan to work on this issue? I'm especially interested for my AutoPlan plugin

Seems abanadoned to me...

ping @danielpetrykowski ?

PR #4757, opened 2025-07-14
docs(README.md): fix link to caldav explanation file

axodentally:
# Description

caldav explanation file was moved, but the link in the readme still pointed to the old location. I also changed the link to be a relative link, not an absolute.


johannesjo:
I think the document is not quite correct. Not sure we should link in this state.

axodentally:
I am not capable of identifying and describing the current state of caldav support in super productivity. 
Therefore, if you would like to have the documentation improved and not just the link fixed, this PR can/should be closed.

PR #4850, opened 2025-07-27
feat: add os module support to plugin node executor

### LINKED ISSUE #4849: 💡 Add os module support to plugin node executor
### Problem Statement

The plugin node executor currently restricts plugins to only `fs` and `path` modules when using direct execution mode. This limitation prevents plugins from accessing basic system information that would be useful for creating platform-aware functionality.

For example, plugins currently cannot:
- Detect the operating system (`os.platform()`)
- Check available memory (`os.freemem()`, `os.totalmem()`)
- Get CPU information (`os.cpus()`)
- Access user directories (`os.homedir()`)
- Determine system architecture (`os.arch()`)

This makes it difficult to build plugins that adapt their behavior based on the user's system environment.

### Possible Solution

Add the Node.js `os` module to the **allowed modules** list in the plugin node executor.

The implementation would involve updating `electron/plugin-node-executor.ts`:

1. Modify the `canExecuteDirectly()` regex pattern to allow `os` module imports
2. Add `os` module import to the `executeDirectly()` sandbox environment
3. Extend the sandbox `require()` function to provide access to the `os` module

The `os` module is safe to expose because it:
- Provides read-only system information
- Cannot modify files or execute processes
- Cannot access network or perform dangerous operations
- Maintains the existing security model

_No response_

### Additional context

This enhancement would enable plugins to build platform-aware features like:
- Cross-platform file path handling
- Memory-conscious operations based on available RAM
- CPU-optimized processing based on core count
- Platform-specific UI adaptations
- Smart temp file management using `os.tmpdir()`

Pull request implementing this fix: #4850

tomdevelops:
# Description

Enable plugins to access the Node.js 'os' module for system information gathering alongside the existing fs and path modules.

Changes:
- Updated canExecuteDirectly() regex pattern to allow 'os' module imports
- Added os module import to executeDirectly() sandbox environment
- Extended sandbox require() function to provide access to os module

This allows plugins to access system information like platform, architecture, memory usage, CPU info, and network interfaces through the standard Node.js os module while maintaining the existing security restrictions.

The os module is considered safe as it provides read-only system information and doesn't allow file system modifications or process execution.

## Issues Resolved

Fixes #4849

## Check List

### DIFF
--- a/electron/plugin-node-executor.ts
+++ b/electron/plugin-node-executor.ts
@@ -102,7 +102,7 @@ class PluginNodeExecutor {
   private canExecuteDirectly(script: string): boolean {
     // Check if script only uses safe operations
     const dangerousPatterns =
-      /require\s*\(\s*['"`](?!fs|path)[^'"]+['"`]\s*\)|child_process|exec|spawn|eval|Function|process\.exit/;
+      /require\s*\(\s*['"`](?!fs|path|os)[^'"]+['"`]\s*\)|child_process|exec|spawn|eval|Function|process\.exit/;
     return !dangerousPatterns.test(script);
   }
 
@@ -110,12 +110,14 @@ class PluginNodeExecutor {
     // Safe modules
     const fs = await import('fs');
     const path = await import('path');
+    const os = await import('os');
 
     // Create sandboxed context
     const sandbox = {
       require: (module: string) => {
         if (module === 'fs') return fs;
         if (module === 'path') return path;
+        if (module === 'os') return os;
         throw new Error(`Module '${module}' is not allowed`);
       },
       console: {


johannesjo:
Thank you! Just wondering: what would be a use case for this?

tomdevelops:
Hey @johannesjo,

Thanks for the fast PR merge :zap:🙏

This weekend I stumbled upon an [MCP plugin](https://github.com/organicmoron/SP-MCP) on Reddit and wanted to give it a try. Unfortunately, it doesn’t work with my current SP version 14.2.2. After digging into it, I found that it was initially relying on both the os module and process.env for operating system and home path information, but I was able to simplify it down to just using the os module.

The plugin uses a file-based inter-process communication model through two directories: the MCP server writes JSON command files that the SP plugin polls, and responses are returned the same way. It’s a surprisingly clever and effective approach, especially given the constraints of the new SP Plugin sandbox.

What I really wanted to try was combining this with a Speech-to-Text MCP [github.com/mbailey/voicemode](https://github.com/mbailey/voicemode) to control SP via voice commands.

I’ll be adding the os module support later today, so the plugin should be functional with the upcoming SP release.

I've created a fork with some improvements here: [github.com/tomdevelops/sp-
mcp/tree/feature/big-beautiful-patch](https://github.com/tomdevelops/sp-mcp/tree/feature/big-beautiful-patch) :star:  opencode support :star:

cheers tom

johannesjo:
Thanks for the explanation. Looking forward to see what you cook up! :)

PR #4910, opened 2025-08-05
Allow simple hostnames in CalDAV URL

OptionalLion411:
# Description

Currently, the pattern to validate the base URL for a CalDAV instance needs to contain a domain, like `example.com`. The only exception is the fixed hostname `localhost`. As it is also possible to host a CalDAV server like Nextcloud using another local hostname without a domain, the pattern should also accept these.
For example, my Nextcloud instance is available using `https://local/nextcloud`.

This PR changed the URL Regex to `/^(http(s)?:\/\/)?([\w\-]+(?:\.[\w\-]+)*)(:\d+)?(\/\S*)?$/i`, to support both domains and simple hostnames in validation.

## Check List

### Question
As a question, I only found about Super Productivity today and I am already in love with it. Are you open for constributions to the issues providers? I would like to improve the sync and add some providers. Also two-way sync like partially supported by CalDAV would be a goal.


johannesjo:
Thank you!

PR #4949, opened 2025-08-13
Fix #4736: confetti celebration does not respect the "Disable All Animations" anymore

### LINKED ISSUE #4736: 🚨 Confetti celebration does not respect the "Disable All Animations" anymore
### Version Used

14.0.5 (EXE from github)

### Expected Behavior

This might be a small issue, but until some major releases ago, before confetti celebration was introduced (or refactored), the celebration animation would respect the setting of Disable All Animations, meaning it would not play at all if we have this setting enabled (expected).

I'm remoting into a work machine quite often and usually try to disable all animation types to reduce latency and frame stream glitches, so being able to disable them (as we could before) would be great.

However, since the introduction of confetti animation, this setting does not work with it.

### Current Behavior

When enabling "Disable All Animations", user should not see the Confetti Celebration.

### Steps to Reproduce

1. Open Super Productivity and go to `Settings`;
2. Under `Misc Settings`, toggle `Disable all animations` On;
3. Go to `Today` view and click on `Finish Day`;
4. The Confetti animation should not play in this scenario.

### Console Output

```shell

```

notsecret808:
# Description

This PR fixes an issue where the "Disable All Animations" setting was ignored by confetti animations. It also adds typing for the canvas-confetti package.

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/4736

## Check List

### DIFF
--- a/src/app/core/confetti/confetti.model.ts
+++ b/src/app/core/confetti/confetti.model.ts
@@ -1,0 +1,23 @@
+export interface ConfettiConfig {
+  particleCount?: number;
+  angel?: number;
+  spread?: number;
+  startVelocity?: number;
+  decay?: number;
+  gravity?: number;
+  drift?: number;
+  flat?: boolean;
+  ticks?: number;
+  origin?: { x?: number; y?: number };
+  colors?: string[];
+  shapes?: (string | Shape)[];
+  scalar?: number;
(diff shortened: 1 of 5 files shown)


johannesjo:
Thank you very much for this!

PR #5002, opened 2025-08-26
Updated italian translations

loviuz:
I have updated all the italian strings based on en.json :-)


johannesjo:
Thank you very much!

PR #5013, opened 2025-08-29
feat: add dont start session after break pomodoro settings option

notsecret808:
# Description

Add a new Pomodoro option "Disable auto start of session after break". This option stops auto-tracking time when a break is over or skipped.

If the option is enabled, it also stops auto-tracking time after break for "Manually confirm starting next break".

## Issues Resolved

[time exceed](https://github.com/johannesjo/super-productivity/issues/4855)

## Check List


johannesjo:
Hey hey! Thank you very much for this!


Please be aware that I am currently in the process of migrating pomodoro into the focus mode, where it will be implemented in a different fashion. I will keep alive the old implementation for a while, but ultimately it will be removed.

PR #5016, opened 2025-08-30
Re-enable the UI option to allow self-signed certs for Jira

justyns:
# Description

I was looking for a way to allow super-productivity to connect to jira through a corporate proxy and realized the code already had support for it, but the UI config option was disabled.   All I'm doing is uncommenting it to let it show up in the ui again.

Note: The commit commenting the option out mentioned it being being an edge case, but I think a lot of corporate/enterprise users would run into the same issue due to corporate proxies mitm'ing everything and/or using a self-signed CA.   It also didn't look like there's a built-in way for electron apps to use the system trust store.

## Issues Resolved

- https://github.com/johannesjo/super-productivity/issues/348

## Check List

### DIFF
--- a/src/app/features/issue/providers/jira/jira-cfg-form.const.ts
+++ b/src/app/features/issue/providers/jira/jira-cfg-form.const.ts
@@ -114,13 +114,13 @@ export const JIRA_CONFIG_FORM_SECTION: ConfigFormSection<IssueProviderJira> = {
       type: 'collapsible',
       props: { label: T.G.ADVANCED_CFG },
       fieldGroup: [
-        // {
-        //   key: 'isAllowSelfSignedCertificate',
-        //   type: 'checkbox',
-        //   templateOptions: {
-        //     label: T.F.JIRA.FORM_CRED.ALLOW_SELF_SIGNED,
-        //   },
-        // },
+        {
+          key: 'isAllowSelfSignedCertificate',
+          type: 'checkbox',
+          templateOptions: {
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you! Is re-enabling the setting really enough to make this work again?

justyns:
I can't test it on my actual work laptop to verify it works there, but I did a simple test by adding a new jira provider with the url `https://self-signed.badssl.com/` and fake user/password.  By default, it gives this error when clicking on test:

```
[issue] request to https://self-signed.badssl.com/rest/api/latest/issue/picker?currentJQL=&query=&showSubTaskParent=true&showSubTasks=true failed, reason: self signed certificate
```

After  checking the **Allow self signed certificate** option and clicking test again, it gets a 404 response:

```
log.ts:229 [issue] JIRA_RESPONSE_ERROR {error: Error: <html>
<head><title>404 Not Found</title></head>
<body bgcolor="white">
<center><h1>404 N…, requestId: 'issue/picker__GET__dYYnzh9D-rEft120xa8Iq'} 
```

Which is expected since it's not a valid jira url, but it doesn't complain about the cert anymore.

I am wondering if we should rename it (at least in the ui) to "Disable certificate validation" or something like that though?  It allows things like certs with hostname mismatches or expired certs, not _just_ self-signed certs.

johannesjo:
> I am wondering if we should rename it (at least in the ui) to "Disable certificate validation" or something like that though? It allows things like certs with hostname mismatches or expired certs, not _just_ self-signed certs.

Yeah sound reasonable.

Alright, I am merging this then! Thank you!!

PR #5024, opened 2025-08-31
fix: Manually confirm starting next break conflicts with Disable auto-start of session after break

notsecret808:
# Description

After releasing the new option of Pomodoro @kurt-steiner found a bug: the tracker still tracks time when "Manually confirm starting the next break" is enabled. This fix addresses that bug.

## Issues Resolved
[the bug](https://github.com/johannesjo/super-productivity/issues/5023)

## Check List


johannesjo:
Thank you very much!

PR #5032, opened 2025-09-01
AUR: include checksums

AriaMoradi:
# Description

Users asked for checksums to be included in the AUR comments, https://github.com/johannesjo/super-productivity/pull/4495#issuecomment-2953933666 suggests adding a checksums file to the releases first, but I figured that the github actions that is being used already includes a feature to handle that, so this became an straightforward one. Also now it uses sha265 sums which should be more satisfactory.

cc: @alosarjos

## Check List

- [X] New functionality has been documented in the README if applicable.


alosarjos:
@AriaMoradi Thanks a lot for the ping and for the changes. But while the pipeline could generate the checksums on the fly, but rather than that, the CICD should generate and publish the checksum values for the artifacts, and then set that value if we want to do it "right".

I guess this may be better than nothing though...

johannesjo:
Thank you and thanks @alosarjos for your input!

PR #5051, opened 2025-09-03
fix(jira): fallback from /search/jql to /search on error (#5015)

### LINKED ISSUE #5015: 💥 Jira Server/Data Center integration broken by Cloud fix in issue #4976
Hi @johannesjo,

Beforehand, I'm not a native English speaker - apologies for any mistakes. This text was polished by ChatGPT, so feel free to correct phrasing.

Your recent fix targeting Cloud Jira (/search/jql) inadvertently broke integrations with Jira Server/Data Center. Before the change, our Server/DC Jira (e.g. v9.x, self-hosted) worked fine — Super Productivity correctly used /rest/api/2/search?jql=…. Now it always hits /search/jql, which simply doesn’t exist there, causing 404 errors.

This is not a misconfiguration — it's a clear case of mismatched endpoint. The app must distinguish between Cloud and Server/DC and use:

pathname = 'search/jql' for Cloud Jira,

pathname = 'search' (API v2) for Server/Data Center.

Without this differentiation, Server/DC setups are entirely broken. Please restore dynamic routing based on Jira type, or **let users override it manually**. I can provide logs, config screenshots, or help test fixes. Thanks.

### Steps to Reproduce

1. Configure Super Productivity with a Jira Server/Data Center instance (e.g. self‑hosted v9.x), not Cloud.
2. Ensure the integration points to your Jira host using email/username and valid token credentials.
3. Use the auto-import feature (or manually trigger it) and observe that instead of calling /rest/api/2/search?jql=…, the app attempts to call /rest/api/latest/search/jql, which doesn't exist on Server/DC — resulting in a 404 failure in fetching new issues.

### Additional Console Output
```
log.ts:229 [issue] JIRA_RESPONSE_ERROR Object Object
log.ts:235 [issue] <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
    <status><status-code>404</status-code>
    <message>null for uri: https://jira.hh.ru/rest/api/latest/search/jql</message>
    </status>
log.ts:229 [issue] {HANDLED_ERROR_PROP: 'Jira: <?xml version="1.0" encoding="UTF-8" standalone="yes"?><status><status-code>404</status-code><message>null for uri: https://jira.hh.ru/rest/api/latest/search/jql</message></status>'}
```

### Url
file:///C:/Program%20Files/WindowsApps/53707johannesjo.SuperProductivity_14.3.4.0_x64__ch45amy23cdv6/app/resources/app.asar/.tmp/angular-dist/browser/index.html#/project/INBOX_PROJECT/tasks

### Meta Info
META: SP14.3.4 __ Electron – ru – Win32 – ru – UA:Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) superProductivity/14.3.4 Chrome/138.0.7204.157 Electron/37.2.4 Safari/537.36

ArtK-HH:
## Problem
Jira Server/DC returns 404 for `/rest/api/{2|latest}/search/jql`, which breaks quick JQL search and backlog auto-import and spams an error snack.

## Root cause
Recent changes switched to `/search/jql` (Cloud-style). Server/DC does not expose that endpoint.

## Changes
- **search$**: try `/search/jql` first (error snack suppressed); on any non-auth error fallback to `/rest/api/*/search?jql=...` (GET).
- **findAutoImportIssues$**: try POST `/search/jql` first (snack suppressed); on non-auth error fallback to POST `/rest/api/*/search` with `{ jql, ... }`.
- Show an error **only if** the fallback also fails.
- No public API changes, minimal diff (1 file).

## Compatibility
- Cloud: unchanged (still uses `/search/jql` successfully).
- Server/DC: fixed.

## Testing
- Verified against Jira Server/DC — primary call 404s silently, fallback returns results.
- CI should pass (no test changes). Local pre-push failed due to locale-dependent test, unrelated to this change.

## Linked issue
Fixes #5015

### DIFF
--- a/src/app/features/issue/providers/jira/jira-api.service.ts
+++ b/src/app/features/issue/providers/jira/jira-api.service.ts
@@ -146,10 +146,25 @@ export class JiraApiService {
         // NOTE: we pass the cfg as well to avoid race conditions
       },
       cfg,
+      suppressErrorSnack: true,
     }).pipe(
       // switchMap((res) =>
       //   res.length > 0 ? of(res) : this.issuePicker$(searchTerm, cfg),
       // ),
+      catchError((err) => {
+        const code = err?.error?.statusCode ?? err?.status ?? err?.error?.status;
+        if (code === 401 || code === 403) return throwError(() => err);
+        // Fallback for Server/DC: /search?jql=...
+        return this._sendRequest$({
+          jiraReqCfg: {
+            pathname: 'search',
+            followAllRedirects: true,
+            query: { jql: searchTermJQL },
+            transform: mapToSearchResultsForJQL,
+          },
+          cfg,
+        });
+      }),
       tap((v) => IssueLog.log('AAAAA', v)),
     );
   }
@@ -223,7 +238,23 @@ export class JiraApiService {
         },
       },
       cfg,
-    });
+      suppressErrorSnack: true,
+    }).pipe(
+      catchError((err) => {
+        const code = err?.error?.statusCode ?? err?.status ?? err?.error?.status;
+        if (code === 401 || code === 403) return throwError(() => err);
+        // Fallback for Server/DC: POST /search with jql in body
+        return this._sendRequest$({
+          jiraReqCfg: {
+            transform: mapIssuesResponse as (res: any, cfg?: JiraCfg) => any,
+            pathname: 'search',
+            method: 'POST',
+            body: { ...options, jql: searchQuery },
+          },
+          cfg,
+        });
+      }),
+    );
   }
 
   getIssueById$(issueId: string, cfg: JiraCfg): Observable<JiraIssue> {
@@ -364,10 +395,12 @@ export class JiraApiService {
     jiraReqCfg,
     cfg,
     isForce = false,
+    suppressErrorSnack = false,
   }: {
     jiraReqCfg: JiraRequestCfg;
     cfg: JiraCfg;
     isForce?: boolean;
+    suppressErrorSnack?: boolean;
   }): Observable<any> {
     return this._isInterfacesReadyIfNeeded$.pipe(
       take(1),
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much! Maybe we should make the fallback a bit more specific and only execute it when we get a 404?

ArtK-HH:
> Maybe we should make the fallback a bit more specific and only execute it when we get a 404?

I first tried to fallback only on 404, but it didn’t work in Electron path. The error we get back isn’t a clean HttpErrorResponse with a numeric status; it’s an Error with an XML body that contains <status-code>404</status-code>, so err?.status/err?.error?.statusCode are undefined.
As soon as I switched to catching all non-auth errors, the fallback started working. Logs suggest we’re seeing a misleading/stripped status in the error payload, not a plain 404.

johannesjo:
Then I'd suggest we check for this (xml body containing a 404 string). Since this is a workaround I would like to avoid to mess and mask other potential error cases

brb-AT:
Hello, as someone who uses the JIRA integration a lot with a self-hosted DC version, I am stuck on the last version before the cloud change.

Rather than trying both URLs and see if they work, wouldn't it make more sense to give us a setting in the JIRA options if the cloud or the old DC API should be used? It can probably be combined with the legacy API key setting, as that one is also required for DC.

So simply a toggle between cloud version and self hosted version?

ArtK-HH:
Hi! @brb-AT 

I’m leaning towards the minimal-intrusive fix with a fallback. From what I’ve seen, the problem mainly affects JQL-based imports, and a simple fallback in one place solves it without adding extra configuration complexity. To be honest, this is also closer to the level of changes I can confidently handle myself.

That said, if someone is interested in implementing a more flexible solution with a user-facing toggle, that would be very welcome and much appreciated!

johannesjo:
Is the problem really about self hosted vs cloud jota? Or is it about jira version?

But I agree: let's go with the simpler version for now. It's much better than it not working at all.

brb-AT:
AFAIK the cloud services have a new version, while the Datacenter/self hosted versions still use the old version.

But as apparently Atlassian is ending support for all self-hosted versions in 2029, this workaround won't stay forever :man_shrugging:

johannesjo:
Ok I am doing the fix myself.  Thanks for getting this started @ArtK-HH .

PR #5071, opened 2025-09-06
fix: incorrect sidebar opening when dragging

novikov1337danil:
# Description

This PR solves an issue where the right sidebar of better-drawer-container would unnecessarily expand when dragging elements to the right side, causing the entire layout to shift to the side and making the panel visible (even though isOpen: false)

A small refactoring was also done to make the code cleaner:
1. Unnecessary and old commented css properties were removed
2. Helper types `StyleObject` and `CssString` were created, as well as a function for converting `StyleObject` to a `CssString`

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/5017


johannesjo:
Thank you very much <3

PR #5072, opened 2025-09-06
fix: remove unnecessary condition to show subtasks in task-detail-pan…

novikov1337danil:
# Description

Removed unnecessary (IMHO) conditions for displaying subtasks in the task-detail-panel component. Now the "subtasks" section will always be shown if the task is not a subtask (don't have a `parentId`)

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/5053


johannesjo:
Thank you!

PR #5101, opened 2025-09-13
fix: Changing settings (not related to the Pomodoro) resets the Pomodoro timer

### LINKED ISSUE #4879: 🚨 Changing settings (not related to the Pomodoro) resets the Pomodoro timer
### Version Used

Safari - SP 14.2.5

### Expected Behavior

When I change a setting not related to the Pomodoro Timer, it shouldn’t reset it.

### Current Behavior

For example, when I observed [this issue](https://github.com/johannesjo/super-productivity/issues/4878), I observed also the reset of my timer. I tried also other settings and the same thing appears.

A solution to repair this is to save also the Pomodoro Timer

### Steps to Reproduce

1. Open SP 
2. Activate in Settings > Productivity Helper > Pomodoro Timer
3. Create a task
4. Track this task
5. Change a random setting

### Console Output

```shell

```

Abrar74774:
# Description

This change fixes the issue reported in #4879 where changing any of the settings other than pomodoro settings reset the pomodoro timer

## Issues Resolved

#4879

### DIFF
--- a/src/app/features/pomodoro/pomodoro.service.ts
+++ b/src/app/features/pomodoro/pomodoro.service.ts
@@ -52,6 +52,7 @@ export class PomodoroService {
 
   cfg$: Observable<PomodoroConfig> = this._configService.cfg$.pipe(
     map((cfg) => cfg && cfg.pomodoro),
+    distinctUntilChanged(),
   );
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you!

PR #5112, opened 2025-09-14
feat: add ability to delete single instances of repeated tasks

### LINKED ISSUE #4195: Delete one instance of a repeated event
We have meetings regularly every week, so I've made it a repeating task. But when one meeting is cancelled I'd like to delete just this one entry on the day and not the whole repeating task.

Something like on delete select:
- Delete this and all future events
- Delete only this event

ReduxST:
# Description

This implements the ability to delete individual instances of repeated tasks without affecting the entire repeat configuration.

### Key Changes:

**Core Implementation:**
- Add `deletedInstanceDates` array to `TaskRepeatCfgCopy` model to store ISO date strings of skipped instances
- Add `deleteTaskRepeatCfgInstance` action and reducer handler to manage exception list
- Update `TaskRepeatCfgService._getActionsForTaskRepeatCfg` to skip creation for deleted instances

**UI/UX Alterations:**
- Add "Remove today" button in repeat task config dialog for single instance deletion
- Smart button visibility: only show for future dates (If this isn't done, it will show for non-scheduled and already created repeat tasks, in which case the button won't do anything)
- Pass `targetDate` context from task components, planner, and schedule views to dialog
- Include confirmation dialog with localized date formatting

**Added the proper translation keys:**
- Add `F.TASK_REPEAT.F.REMOVE_INSTANCE` and deletion confirmation dialog keys
- Update `t.const.ts` with proper key mapping for translation resolution

### Technical Approach:
Uses an exception list pattern. When an instance is marked for deletion, it's added to `deletedInstanceDates` and filtered out during task creation and projection, providing immediate visual feedback without complex state management.

### Compatibility:
- Maintains full backward compatibility with existing repeat configurations
- No changes to existing "Remove all" functionality
- Works in both the planner and the schedule tab

## Issues Resolved

Fixes #4195

## Check List


johannesjo:
Thank you very much! This is another great addition to the repeatable tasks feature!

PR #5127, opened 2025-09-17
perf: prevent unnecessary change change detections with high-frequenc…

atscott:
…y listeners

Listeners schedule change detection, even in zoneless, so those that are high frequency should be used sparingly. The listeners in this commit do not require any template refreshes most of the time and should instead use native `addEventListener` to avoid unnecessary refreshes. While the application uses `OnPush`, doing a small amount of a lot of times can still have performance implications.

https://github.com/angular/angular/issues/61839


johannesjo:
Thank you very much!

PR #5156, opened 2025-09-21
feat: support multilanguage shortcuts

novikov1337danil:
# Description

- **First commit:**
  - Reworked `onKeyDown` method in `keyboard-input.component.ts` to support multilingual keyboard layouts. Now it uses `event.code` to identify keys, which is consistent across different layouts.
  - Reworked `checkKeyCombo` function to also use `event.code`, and handled an edge case for checking single "+" key. This will be useful for those with a numpad that includes that key.
  - Added a `prepareKeyCode` function that converts `event.code` into a format that the `checkKeyCombo` function understands. It removes unnecessary prefixes and transforms some key codes.
  - Updated the test file `check-key-combo.spec.ts`.

- **Second commit:** change the letter cases in `default-global-config.const.ts` to uppercase because with the new approach, we can't determine the case of the pressed letters. This means that if you add the "a" key as shortcut, it will be recorded as "A." Don’t worry, all the old combinations that include lowercase letters will still work since I normalize them to the same case in the `checkKeyCombo` function.
  - I took a quick look at how other applications handle this and found that it’s a common practice. For example, in VSCode, all letters in the shortcuts are written in uppercase. I think they use a similar approach, but I did not look into their code.

- **Other commits:** Just some refactoring, no major changes.

## Issues Resolved

https://github.com/johannesjo/super-productivity/pull/5073

# P.S. 

While I was exploring the codebase, I noticed that in the examples using the `PluginAPI.registerShortcut` function, the `keys` property has incorrect values due to case sensitivity. For instance, in `docs/plugin-development.md:351`, it should be `keys: 'Ctrl+Shift+p'`.

This could be easily handled by normalizing the modifiers to a consistent case in the `checkKeyCombo` function so that `Ctrl` and `ctrl` are treated the same. However, I'm not sure if this change is really necessary.


johannesjo:
Thank you very much! Looks good at first glance! I'll have anothet look once I'm back from vacation. Meanwhile could have a look at the limiting errors?

johannesjo:
I am back now and will try to have a look tomorrow!

johannesjo:
Unfortunately I got hold up by the crucial flathbu and snap bug that occurred yesterday.

Thank you very much once more! This is a great addition!

PR #5172, opened 2025-09-25
Properly show other places in how-to-rate.md

rugk:
Without it or at least `  ` (two spaces) at the end, it is shown in one line which is not really beautiful.


johannesjo:
There was another PR putting a blank line in between. Not sure what's the better solution here :)

rugk:
IMHO proper bullet points of course 😉 

No well, a blank line would strictly speaking be a Markdown paragraph so yeah.

johannesjo:
I agree! :D

johannesjo:
Thank you!

PR #5181, opened 2025-09-26
dialog-confirm: change default to 'OK'

nightgolfer:
# Description

For dialog-confirm modals, change the initially selected / default button from 'cancel' to 'ok' (by switching the tabindex of both). This should make 'OK' the default action when pressing Return on a keyboard, and thus assuming that the user knows what they're doing, and don't need an extre keypress (right arrow) to confirm creating a new tag.

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/5180

## Check List

- [%] New functionality includes testing. _Haven't figured out how to compile the code from source yet, so haven't tested 🫣_
- [%] New functionality has been documented in the README if applicable. _Don't think it's needed._

### DIFF
--- a/src/app/ui/dialog-confirm/dialog-confirm.component.html
+++ b/src/app/ui/dialog-confirm/dialog-confirm.component.html
@@ -16,7 +16,7 @@
 <mat-dialog-actions align="end">
   <div class="wrap-buttons">
     <button
-      tabindex="1"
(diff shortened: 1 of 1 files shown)


johannesjo:
Not sure we should do this. Especially for things like confirming deletion.

PR #5198, opened 2025-09-30
Update German Translation file: de.json

ExoGitH:
# Description

Upate the German translation file. There are missing lines that are not translated for some times. Secound try, because first one was not finished the hole file.

## Issues Resolved

Some translation may also correct the button disappearance, as they are not translated.

## Check List


johannesjo:
Vielen Dank!

PR #5223, opened 2025-10-05
fix(task): improve project name parsing from task input (#4225)

### LINKED ISSUE #4225: Tasks can be created with the wrong name when using the "+" feature if the project name contains a "+"
It's not a fascinating bug, but I did run into it organically since I've got a project that has "+" in its title. 

### Your Environment

- Version used: 12.0.2 (snap)
- Operating System and version: Ubun2u 24.04, kernel version 6.8.0-31-generic
- Desktop Environment: i3

### Expected Behavior
If the project is e.g. "A + B", and the task "C + A + B", the task name should be "C" and be associated with the project "A + B". 

### Current Behavior
The task name ends up being called "C + B" instead, though it is correctly associated with the project

1. Create a project called "A + B"
2. Make a new task meant to be called "C" and associate it to the project by naming it "C + A + B" for SuperProductivity to associate it with the "A + B" project. This is the name the task would be given if selected from the project name dropdown included in the app when typing "+".

Yes

miqh:
# Description

Hello there,

I've revised the regular expression that was being used to parse project names from task text input for better accuracy. When adding new tasks, it should better cope with project names which happen to use special characters part of the short syntax feature.

The following screen capture is behaviour based on `15.0.3` via https://app.super-productivity.com.

old behaviour

And here's the behaviour based on the current state of `master` plus my changes.

new behaviour

I've included a whole slew of data-driven tests to try capture as many parse cases as possible. If this is too much, let me know and I can redact as necessary.

## Issues Resolved

#4225

## Check List

### DIFF
--- a/src/app/features/tasks/short-syntax.ts
+++ b/src/app/features/tasks/short-syntax.ts
@@ -66,7 +66,13 @@ customDateParser.refiners.push({
   },
 });
 
-const SHORT_SYNTAX_PROJECT_REG_EX = new RegExp(`\\${CH_PRO}[^${ALL_SPECIAL}]+`, 'gi');
+// The following project name extraction pattern attempts to improve on the
+// previous version by not immediately terminating upon encountering a short
+// syntax delimiting character and looks ahead to consider usage context
+const SHORT_SYNTAX_PROJECT_REG_EX = new RegExp(
+  `\\${CH_PRO}((?:(?!\\s+(?:\\${CH_TAG}|\\${CH_DUE}|t?\\d+[mh]\\b)).)+)`,
+  'i',
+);
 const SHORT_SYNTAX_TAGS_REG_EX = new RegExp(`\\${CH_TAG}[^${ALL_SPECIAL}|\\s]+`, 'gi');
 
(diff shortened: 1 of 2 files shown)


src/app/features/tasks/short-syntax.ts
```diff
@@ -198,7 +204,10 @@ const parseProjectChanges = (
 
     if (existingProject) {
       return {
-        title: task.title?.replace(`${CH_PRO}${projectTitle}`, '').trim(),
+        title: task.title
+          ?.replace(`${CH_PRO}${projectTitle}`, '')
+          .trim()
+          .replace('  ', ' '),
```
miqh, line 210: The second `replace` here mirrors the behaviour on L229.

This is needed because the updated project name pattern matching is more specific about isolating only the project name itself.

As far as I understand, this replacement is intended to consume the whitespace character that delimits the project name from whatever follows (e.g. `task +proj 10m` becomes `task 10m` rather than `task  10m`).

johannesjo:
Thank you very much! This is a great addition! I also love the videos you added so I can quickly grasp on what's new! Great work!

PR #5255, opened 2025-10-12
fix: update check-key-combo logic to correctly evaluate all pressed modifiers

novikov1337danil:
# Description

**1 commit:** refactored the function to check for extra modifiers pressed that are not required in the combination
**2 commit:** added a test that checks for additional modifiers that are not required in the combination

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/5253


johannesjo:
Thank you very much! <3

PR #5261, opened 2025-10-13
Add shortcut for projects

Heapnotizer:
Add shortcut 'shift+p' to add new projects

Fixes: https://github.com/johannesjo/super-productivity/issues/5176


johannesjo:
Thank you very much!

PR #5265, opened 2025-10-13
Feat: sort feature in boards

Heapnotizer:
Add sort feature for boards.

Fixes: https://github.com/johannesjo/super-productivity/issues/5248

### DIFF
--- a/src/app/features/boards/board-panel/board-panel.component.ts
+++ b/src/app/features/boards/board-panel/board-panel.component.ts
@@ -176,7 +176,47 @@ export class BoardPanelComponent {
         nonOrderedTasks.push(task);
       }
     });
-    return [...orderedTasks, ...nonOrderedTasks].filter((t) => !!t);
+    const merged = [...orderedTasks, ...nonOrderedTasks].filter((t) => !!t);
+
+    // mode: 'off' | 'asc' (earliest first) | 'desc' (latest first)
+    const mode = panelCfg.sortByDue || 'off';
+    if (mode !== 'off') {
+      const getDueTs = (task: TaskCopy): number | null => {
+        // prefer dueWithTime (timestamp), then dueDay (ISO-like date string)
+        if (task.dueWithTime) {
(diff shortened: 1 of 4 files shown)


johannesjo:
Thank you very much for opening this up! There are at least some cases it does not seem to work properly for:

https://github.com/user-attachments/assets/2e656325-1550-4a0e-a68e-8d3b2ebc30d9

johannesjo:
Also: What do you think about limiting this only to columns that specifically include due tasks?

Heapnotizer:
> Also: What do you think about limiting this only to columns that specifically include due tasks?

Yeah, it will be good to limit to "todo" and "in progress". Sorting is redundant in "done" tasks.

Heapnotizer:
> Thank you very much for opening this up! There are at least some cases it does not seem to work properly for:
> 
>  simplescreenrecorder-2025-10-14_10.18.36.mp4

I am guessing you are referring to tasks when due date is same? (7:00 PM tasks in video). That behaviour is because due date and due time are same for both tasks. If you change time of one task, you will see it auto sorting.  Since, i did not add case for sorting via name in case of tie, this behaviour is occuring. did i miss something you wanted to say?

johannesjo:
> I am guessing you are referring to tasks when due date is same? (7:00 PM tasks in video). That behaviour is because due date and due time are same for both tasks. If you change time of one task, you will see it auto sorting. Since, i did not add case for sorting via name in case of tie, this behaviour is occuring. did i miss something you wanted to say?

Ahh. My bad. I was confused by the dates I chose :D 

Thank you very much for this! I'm merging.

Heapnotizer:
Wait, the logic for limiting sort to due tasks is not in that. But i have the changes ready in the branch. Opening new PR

zarembas:
Does this feature not exist anymore? I was searching online and landed on this thread. I can't see the opton in the settings.

2026-01-26 at 11 12 19

PR #5275, opened 2025-10-14
Fix hardware acceleration in the snap

sespiros:
Hi,

Thank you very much for making super-productivity and for packaging it for so many different formats! I have been using it since last month and I am incredibly happy with it.

I was affected by https://github.com/johannesjo/super-productivity/issues/5252 which I've been following for a week and I saw it was resolved yesterday but I noticed today that the latest snap doesn't have hw acceleration anymore.

This is a proposed attempt to fix it for the snap but notice that I haven't tested any other packaging format so feel free to take whatever suggestions from this PR that make sense.

# Description

I upgraded to latest electron again, I overrode electron-builder's choice of gnome content snap and as a proof of concept I removed all the logic which enables the software renderer. 

I built the snap using:
```bash
npm run localInstall:test:snap
```

## Issues Resolved (only snap tested)

https://github.com/johannesjo/super-productivity/issues/5252

## Check List

- [x] New functionality includes testing. I built and ran the snap which works with hw acceleration.

### DIFF
--- a/electron/start-app.ts
+++ b/electron/start-app.ts
@@ -9,7 +9,6 @@ import {
   powerMonitor,
   protocol,
 } from 'electron';
-import { existsSync } from 'fs';
 import { join } from 'path';
 import { initDebug } from './debug';
 import electronDl from 'electron-dl';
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much for the attempt to fix this! 

I am pretty sure among all the different stuff I tried yesterday there must have been something similar to this :D  There are many different combinations of gnome and non-gnome snap users using a broad set of different graphic drivers. Unless you are very sure that this will not lead to problems (the app not starting) for all but a small fraction, I'd rather not merge this now.  

I think the best course of action for now is to stay on electron 37 and wait a bit until things stablize and either electron itself or electron-builder fixes this somehow.

We probably (?) would be fine just re-enabling hardware acceleration, but since yesterday was so nightmarish and probably drove away quite a bit of users, I am a bit cautious.

sespiros:
Understandable. As far as snaps are concerned I don't see reason for regressions as the app gets its drivers from the gnome provider anyway but I've only tested it on my system.

Note that ideally this fix should go in [electron-builder's template](https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/templates/snap/snapcraft.yaml) which uses an older core20 snap and a gnome content snap compatible with it. Once the template is updated there, you would automatically get the fix when you upgrade your electron-builder version.

Until that happens, when you are "overriding" the template to base your snap in core22, you have to also override the gnome provider part with one that is compatible with core22 (potentially other things too as well). So my goal really was to make sure you are aware of this in case you weren't as I didn't follow all of your fix attempts in detail.

So feel free to do as you feel comfortable. Thank you very much for all the effort! 

Also the PR will be visible in the bug now for snap users that miss hw accel (I definitely did) and are willing to test this approach and report back :)

johannesjo:
Thank you very much for the explanation @sespiros !

Another question: Does the gnome plug require a user to use gnome?

If you're confident about the fix, we could start another testing round on the snap edge channel.

sespiros:
> you have to also override the gnome provider part with one that is compatible with core22 (potentially other things too as well).

Turns out this is not behaving as I was expecting. My addition for the new gnome-provider doesn't override the default automatically. So if I want it to work I have to do a manual step:

```bash
snap disconnect superproductivity:gnome-3-28-1804
```

Moreover I rebased this branch on top of current master and re-built it with electron 38 and now I am getting errors unrelated to MESA as well (similar to the ones people had in the 5252 issue). I might have been building my local snap with electron 37 in my tests despite setting 38 in the `package.json`. So I am closing this issue. 

Latest snap version with the removal of the software rendering code got hwaccel back anyway so thanks for that :)

sespiros:
> Another question: Does the gnome plug require a user to use gnome?

To your question, I don't think so. Electron just has a few dependencies from the [gnome content provider](https://documentation.ubuntu.com/snapcraft/stable/reference/extensions/gnome-extension/)

johannesjo:
@sespiros thank you!

PR #5276, opened 2025-10-14
Updated fi.json

Whatsthecatchnelisilm:
Added missing Finnish translations and naturalized some of the existing ones.


johannesjo:
Thank you very much!

PR #5294, opened 2025-10-16
Update de.json following latest changes to en.json

ExoGitH:
# Description

Update of the German version following changes to the English file.


johannesjo:
Thank you very much!

PR #5304, opened 2025-10-17
feat: Add "Repeat from Completion" logic for task repeat configurations

maxliesegang:
# Description

Adds a "Repeat from completion" option for recurring tasks. When enabled, the next task occurrence is calculated from the completion date of the current instance rather than from a fixed start date.

This is useful for tasks where you want a consistent interval *between* completions rather than on fixed calendar dates. For example, "Water plants every 3 days" will create the next task 3 days after you complete the current one, not 3 days after the original start date.

## Issues Resolved

https://github.com/johannesjo/super-productivity/discussions/3235

## Check List


johannesjo:
Thank you very much for this! This could be a very cool feature! 

I'm a bit worried, that this might be a bit hard to understand for users not familiar with this. I struggled a bit myself and I am familiar with the discussions. 

Some ideas: Maybe we should rename it to "Repeat, only when done", For me it is logically a switch that changes the meaning of "Repeat Config" completely. We somehow need to reflect this in the UI. Maybe we should even employ a completely new form configuration for it. At the very least we need to dynamically adjust the labels of the form elements.

Another aspect is since it is more of a special case feature, we should minimize the clutter it adds for users that just want standard repeatable tasks and especially new users. The long explanation text should maybe only be displayed, when we activate the checkbox.

What do you think?

maxliesegang:
Great points! I completely agree the UX needs work.

**Re: a completely new form configuration** - I think that might be overkill for now, but we could revisit if the repeat feature evolves. For the initial implementation, I'd prefer smart conditional display/labels within the existing form.

**One simplification:** We only show this option when "Custom" repeat config is selected - not for quick settings like "DAILY" or "WEEKLY" (those are already simple enough).

---

Here are the two approaches I'm considering:

### Option 1: Minimal checkbox
```
Repeat every: [3] [days ▼]
☐ Count interval from completion date
```

- Placed right after "Repeat every" field
- Description only shown when checked
- Simplest implementation

### Option 2: Radio field for mode
```
Schedule type:
○ Fixed schedule
○ Dynamic interval

[If "Fixed schedule":]
Start date: [2025-01-01]

[If "Dynamic interval":]
First occurrence: [2025-01-01]
ℹ️ The next task appears based on when you 
   complete the current one, maintaining a 
   consistent interval between completions.
   
   Example: "Take medication every 7 days" 
   creates the next task 7 days after you 
   complete it, not on fixed calendar dates.
```

- Makes the distinction very explicit upfront
- Can dynamically hide/show relevant fields
- Description only for "Dynamic interval"

---

I'm leaning toward **Option 2** since it makes the mental model clearer and handles the "changes meaning completely" concern you mentioned.

What do you think? Or do you have another approach in mind?

johannesjo:
> **One simplification:** We only show this option when "Custom" repeat config is selected - not for quick settings like "DAILY" or "WEEKLY" (those are already simple enough).

Yes! Good idea! This is probably the best solution!

> What do you think? Or do you have another approach in mind?

Maybe it is just because I'm bad with dates, but "Count interval from completion date" is not intuitive to me.  My brain thinks it has to go to date counting mode and it doesn't like that at all xD

To me ideally both code logic and phrasing should revolve around "Repeat, only when done". It's maybe slightly more complicated to code, but I feel it will make this much less confusing.

For my brain that would work something along those lines:

1. Repeat a task that's done (ui: Checkbox "Repeat, only when done")
2. The rest of the options mean: Repeat when done AND if these conditions are met.

So "Repeat, only when done" should also come first for the custom config items, since it changes the meaning of the rest.

What do you think? Does this make sense to you? Or do you prefer the other way of thinking about this?

johannesjo:
Just read the issue you commented on and it explains a bit the confusion I had :). "Repeat after X days, when done" is also something I can grasp quickly though for me personally "Repeat, when done" would be the most useful.

maxliesegang:
Thanks for the feedback! I've updated the implementation based on our discussion

Changes made:
- Renamed the checkbox label to "Repeat, when done" (keeping it concise as suggested)
- Updated the description: "When enabled, the next task will be created based on when you complete the current one. The start date will be updated to the completion date, rather than using the original fixed start date."
- This option now only appears for Custom repeat config, keeping the UI clean for standard repeat options like DAILY/WEEKLY


I think this strikes a good balance between clarity and simplicity. The feature is now ready for merge IMO – we can let users try it out and gather feedback to see if any further refinements are needed.
What do you think?

johannesjo:
Thank you very much!  

FYI I'll make some further adjustments to the wording. Doesn't make sense to me to word it "Repeat, when done" while the actual functionality remains as "Repeat from completion". I also still find the concept quite confusing.

PR #5317, opened 2025-10-19
fix: close the mobile navigation menu when press back button

novikov1337danil:
# Description

Added `popstate` event handling and created a function for synchronizing history state with the mobile menu's visibility state.

*Additionally: created a constant to store all history keys in one place.

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/5130


johannesjo:
Thank you very much for this. There are some potential problems. Could you please investigate?

---

## 1. Recursive History Manipulation

The current logic might have a significant flow issue:

```typescript
// In onBack() - triggered by popstate event
@HostListener('window:popstate')
onBack(): void {
  if (this.showMobileMenuOverlay())
    this.toggleMobileNav();
}

// toggleMobileNav() calls syncMobileNavHistory()
toggleMobileNav(): void {
  this.showMobileMenuOverlay.update((show) => !show);
  this.syncMobileNavHistory(); // ← Called here
}

// syncMobileNavHistory() may call history.back() AGAIN
syncMobileNavHistory(): void {
  const isVisible = this.showMobileMenuOverlay();
  const hasState =
    window.history.state[HISTORY_STATE.MOBILE_NAVIGATION] !== undefined;

  // After closing in onBack(), isVisible=false and hasState=true
  if (!isVisible && !hasState) return;

  if (!isVisible) { // This branch will execute
    window.history.back(); // ← Calls back() again! 🐛
  }
}
```

### 🧩 Problem

When the user presses back:

1. `popstate` event fires → `onBack()` is called
2. `toggleMobileNav()` closes the menu
3. `syncMobileNavHistory()` sees `isVisible=false` and `hasState=true`
4. Calls `window.history.back()` again, **potentially navigating away** from the current page

### ✅ Solution

Track whether the close was triggered by the back button vs manual toggle.

---

## 2. Missing Event Filter

The `@HostListener('window:popstate')` will trigger for **any** `popstate` event, not just the mobile nav:

```typescript
@HostListener('window:popstate')
onBack(): void {
  if (this.showMobileMenuOverlay())
    this.toggleMobileNav();
}
```

### 🧩 Problem

This will interfere with other components using history (like `TaskDetailPanelComponent`).
If the task detail panel is closed via back button while the mobile nav is also open, this handler will incorrectly close the mobile nav too.

### ✅ Solution

Check the history state to ensure the `popstate` is specifically for the mobile nav:

```typescript
@HostListener('window:popstate')
onBack(): void {
  // Only handle if this popstate removed our mobile nav state
  const hasMobileNavState = window.history.state?.[HISTORY_STATE.MOBILE_NAVIGATION];
  if (!hasMobileNavState && this.showMobileMenuOverlay()) {
    // State was removed, so we should close
    this.showMobileMenuOverlay.set(false);
  }
}
```

---

## 3. Unclear `replaceState` Logic

```typescript
if (isVisible) {
  const args = { state: { [HISTORY_STATE.MOBILE_NAVIGATION]: true }, title: '' };
  if (!hasState)
    window.history.pushState(args.state, args.title);
  else
    window.history.replaceState(args.state, args.title); // When would this happen?
}
```

### 🧩 Question

When would `isVisible=true` **and** `hasState=true`?
This suggests the menu is already open with a history state, then you’re toggling it open again (which would actually close it).
The logic seems **inverted**.

---

## 📝 Recommended Fix

Here’s a cleaner and safer approach:

```typescript
private _isClosingViaBackButton = false;

toggleMobileNav(): void {
  const wasVisible = this.showMobileMenuOverlay();
  this.showMobileMenuOverlay.update((show) => !show);
  const isNowVisible = this.showMobileMenuOverlay();

  if (!this._isClosingViaBackButton) {
    this._manageMobileNavHistory(isNowVisible, wasVisible);
  }

  this._isClosingViaBackButton = false;
}

@HostListener('window:popstate')
onPopState(): void {
  const hasMobileNavState = window.history.state?.[HISTORY_STATE.MOBILE_NAVIGATION];

  // If mobile nav is open but state is gone, user pressed back
  if (!hasMobileNavState && this.showMobileMenuOverlay()) {
    this._isClosingViaBackButton = true;
    this.showMobileMenuOverlay.set(false);
  }
}

private _manageMobileNavHistory(isVisible: boolean, wasVisible: boolean): void {
  if (isVisible && !wasVisible) {
    // Opening - push history state
    window.history.pushState(
      { [HISTORY_STATE.MOBILE_NAVIGATION]: true },
      ''
    );
  } else if (!isVisible && wasVisible) {
    // Closing manually - go back
    window.history.back();
  }
}
```

novikov1337danil:
> ## 1. Recursive History Manipulation
> 
> The current logic might have a significant flow issue:
> 
> ```ts
> // In onBack() - triggered by popstate event
> @HostListener('window:popstate')
> onBack(): void {
>   if (this.showMobileMenuOverlay())
>     this.toggleMobileNav();
> }
> 
> // toggleMobileNav() calls syncMobileNavHistory()
> toggleMobileNav(): void {
>   this.showMobileMenuOverlay.update((show) => !show);
>   this.syncMobileNavHistory(); // ← Called here
> }
> 
> // syncMobileNavHistory() may call history.back() AGAIN
> syncMobileNavHistory(): void {
>   const isVisible = this.showMobileMenuOverlay();
>   const hasState =
>     window.history.state[HISTORY_STATE.MOBILE_NAVIGATION] !== undefined;
> 
>   // After closing in onBack(), isVisible=false and hasState=true
>   if (!isVisible && !hasState) return;
> 
>   if (!isVisible) { // This branch will execute
>     window.history.back(); // ← Calls back() again! 🐛
>   }
> }
> ```
> 
> ### 🧩 Problem
> 
> When the user presses back:
> 
>     1. `popstate` event fires → `onBack()` is called
> 
>     2. `toggleMobileNav()` closes the menu
> 
>     3. `syncMobileNavHistory()` sees `isVisible=false` and `hasState=true`
> 
>     4. Calls `window.history.back()` again, **potentially navigating away** from the current page

No, when calling `popstate`, the flag will already be removed from the `history.state`, which means `hasState=false` and the condition `if (!isVisible && !hasState) return;` will be triggered.

> ## 2. Missing Event Filter
> 
> The `@HostListener('window:popstate')` will trigger for **any** `popstate` event, not just the mobile nav:
> 
> ```ts
> @HostListener('window:popstate')
> onBack(): void {
>   if (this.showMobileMenuOverlay())
>     this.toggleMobileNav();
> }
> ```
> 
> ### 🧩 Problem
> 
> This will interfere with other components using history (like `TaskDetailPanelComponent`). If the task detail panel is closed via back button while the mobile nav is also open, this handler will incorrectly close the mobile nav too.
> ### ✅ Solution
> 
> Check the history state to ensure the `popstate` is specifically for the mobile nav:
> 
> ```ts
> @HostListener('window:popstate')
> onBack(): void {
>   // Only handle if this popstate removed our mobile nav state
>   const hasMobileNavState = window.history.state?.[HISTORY_STATE.MOBILE_NAVIGATION];
>   if (!hasMobileNavState && this.showMobileMenuOverlay()) {
>     // State was removed, so we should close
>     this.showMobileMenuOverlay.set(false);
>   }
> }
> ```

If the mobile menu is already open, then `history.state` already contains a flag indicating this. So I thought this was unnecessary code.

However, now I see that in the `toggleNavFocus()` function, there's a line where the mobile menu is shown not through the `toggleMobileNav()` function, but directly through `this.showMobileMenuOverlay.set(true);`. I guess I shouldn't have called the `syncMobileNavHistory` function in `toggleMobileNav` to always maintain the current `history.state` with a flag about opening the mobile menu.

> ## 3. Unclear `replaceState` Logic
> 
> ```ts
> if (isVisible) {
>   const args = { state: { [HISTORY_STATE.MOBILE_NAVIGATION]: true }, title: '' };
>   if (!hasState)
>     window.history.pushState(args.state, args.title);
>   else
>     window.history.replaceState(args.state, args.title); // When would this happen?
> }
> ```
> 
> ### 🧩 Question
> 
> When would `isVisible=true` **and** `hasState=true`? This suggests the menu is already open with a history state, then you’re toggling it open again (which would actually close it). The logic seems **inverted**.

This can happen if we close the menu using any method (`popstate` or calling `toggleMobileNav`), and then manually move forward in history (meaning the flag will appear in `history.state` again, even though the menu is still closed). Then, we open the menu (call `toggleMobileNav()`) – and now we have `isVisible=true` and `hasState=true`.

Actually, I realize I can remove the line `else window.history.replaceState(args.state, args.title);`, since it doesn't actually do anything (since if `hasState=true`, the history.state flag is always `true`). I just left it as a fallback for safety.

____

Thanks for the feedback. While I was answering the second point, I realized a key point: the `showMobileMenuOverlay` variable can be changed by more than just the `toggleMobileNav()` function. This means I need to update the `history.state` somewhere else — I probably need to subscribe to updates to the `showMobileMenuOverlay` variable.

I'll think about it a bit more and try to improve my code in a couple of days.

Any comments are welcome

novikov1337danil:
@johannesjo I did it through the effect and it works

johannesjo:
Thank you  @novikov1337danil ! Could you maybe have another look at the conflicts please?

johannesjo:
There are also some additional recommendations, but I can also include them later on.

``` 
  1. Add mobile guard:
  @HostListener('window:popstate') onBack(): void {
    if (this.isMobile() &&
  this.showMobileMenuOverlay()) {
      this.toggleMobileNav();
    }
  }

  syncMobileNavHistory(isVisible: boolean): void {
    if (!this.isMobile()) return; // Add this check

    const hasState = window.history.state[HISTORY_ST
  ATE.MOBILE_NAVIGATION] !== undefined;
    // ... rest of logic
  }

  2. Add clarifying comment:
  if (isVisible) {
    const args = { state: {
  [HISTORY_STATE.MOBILE_NAVIGATION]: true }, title:
  '' };
    if (!hasState) {
      window.history.pushState(args.state,
  args.title);
    } else {
      // Replace if state exists (defensive: handles
   rapid open/close/open cycles)
      window.history.replaceState(args.state,
  args.title);
    }
  }
```

novikov1337danil:
> There are also some additional recommendations

Yes, sure. Thanks!

novikov1337danil:
I think in the future we could move all this logic into a separate... service? (I'm not sure about this) and use it in different places of the application where any modals/overlays open (to unify rather than duplicate "closing" functionality)

If you look at `task-detail-pannel.component.ts`, it has its own implementation (and it is different). My component already has a different implementation. Although we are solving the same problem.

johannesjo:
> I think in the future we could move all this logic into a separate... service? (I'm not sure about this) and use it in different places of the application where any modals/overlays open (to unify rather than duplicate "closing" functionality)
> 
> If you look at `task-detail-pannel.component.ts`, it has its own implementation (and it is different). My component already has a different implementation. Although we are solving the same problem.

Yes, this makes a lot of sense!

johannesjo:
I've got another question for you @novikov1337danil ! You have been a great help this month! Thank you very much for this!

Would you be alright with me announcing you as contributor of the month in a post on reddit?

novikov1337danil:
> I've got another question for you @novikov1337danil ! You have been a great help this month! Thank you very much for this!

I'm really happy to contribute to this project :+1: 

> Would you be alright with me announcing you as contributor of the month in a post on reddit?

Oh, that was unexpected! Thank you very much :heart:  Of course, I don't mind. This could become a nice tradition that would motivate people to contribute to this wonderful app

novikov1337danil:
@johannesjo what work is needed for this PR?

johannesjo:
@novikov1337danil 

There are some linting errors:
``` 

/home/runner/work/super-productivity/super-productivity/src/app/features/project/project.service.ts
Error:   196:31  error  Replace `⏎······...project.taskIds,⏎······...project.backlogTaskIds,⏎····` with `...project.taskIds,·...project.backlogTaskIds`  prettier/prettier
Error:   206:25  error  Replace `⏎······...allParentTaskIds,⏎······...subTaskIdsForProject,⏎····` with `...allParentTaskIds,·...subTaskIdsForProject`    prettier/prettier

✖ 2 problems (2 errors, 0 warnings)
  2 errors and 0 warnings potentially fixable with the `--fix` option.
```

novikov1337danil:
> There are some linting errors:

It's strange because I didn't change anything there. Ok, I'll look and fix it

johannesjo:
> It's strange because I didn't change anything there. Ok, I'll look and fix it

Hm that's strange indeed... Thank you for fixing it !

PR #5328, opened 2025-10-21
Change task project and tags via drag and drop to the nav bar

### LINKED ISSUE #810: Drag Tasks Between Projects
## Problem Statement
I am unable to drag tasks between projects. If I make a task in the wrong project, or change my mind about where a task should be, it would be cool to be able to drag it onto a new project.

Dragging a task onto a new project should remove the task from the current project and create the task with the same information in the new project.

Possibly at least a "delete task" button so that the mistakenly added task doesn't show up as "completed". 

This isn't super urgent, but it would be quite useful!

kiausch:
# Description

Allows to drag Tasks from a task list and drop it to a project, inbox, or tag in the main navigation to associate the task with the project.
When dragging to a tag, the tag is toggled on the task, i.e. it is added if it is not in the task list, or removed otherwise

## Issues Resolved

closes #810 

## Check List

### DIFF
--- a/src/app/core-ui/magic-side-nav/magic-side-nav.component.html
+++ b/src/app/core-ui/magic-side-nav/magic-side-nav.component.html
@@ -113,6 +113,16 @@
                     [activeWorkContextId]="activeWorkContextId() || ''"
                     [variant]="'nav'"
                     [showLabels]="showText()"
+                    [attr.data-project-id]="
+                      item.workContextType === WorkContextType.PROJECT
+                        ? item.workContext.id
+                        : null
+                    "
+                    [attr.data-tag-id]="
+                      item.workContextType === WorkContextType.TAG
+                        ? item.workContext.id
+                        : null
+                    "
                     (clicked)="onItemClick(item)"
                   ></nav-item>
                 }
--- a/src/app/core-ui/magic-side-nav/magic-side-nav.component.ts
+++ b/src/app/core-ui/magic-side-nav/magic-side-nav.component.ts
@@ -10,6 +10,7 @@ import {
   OnInit,
   output,
   signal,
+  AfterViewInit,
 } from '@angular/core';
 import { CommonModule } from '@angular/common';
 import { RouterModule } from '@angular/router';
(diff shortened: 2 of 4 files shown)


kiausch:
The newly added playwright test seems to be failing in the CI job. (Why does it show "All checks have passed", though?)
I'll try to figure out why the test fails, it passed locally...

kiausch:
Test seems to pass now...

johannesjo:
Oh wow cool! Thank you! I'll have a look tomorrow!

johannesjo:
@kiausch That's a fantastic addition! And much cleaner and simpler implemented than I thought would be necessary (wasn't aware `DragDropRegistry`  exists). Thank you very much! I'm really looking forward to merge this!

A couple of remarks:

* since the app should be zonless I think `this._ngZone.run` should not  be necessary
* since this is executed for every tag or project on every pointerup we should try to minify performance impact. So maybbe it is better to run this from the `magic-side-nav.component` or from a service?
* Dropping a subtask onto a project entry will now throw     Error('Wrong task model'). We should guard against this.

kiausch:
@johannesjo thanks for the review, I'll try to work on the remarks later today

I think the registered pointerup event is only fired after an element was dragged, but I am also not very happy that this fires for every nav-item. I will try to register the event on the magic-side-nav.component instead.

I think there is another catch with recurring tasks as these seem to require some special treatment like in the `moveTaskToProject` function in `task.component.ts` and `task-context-menu-inner.component.ts`. As these functions are already almost identical and marked with a _Todo_-note, I am not going to replicate this a third time, but will exclude recurring tasks for now until the other todo is fixed...

johannesjo:
Looking forward to the changes!

> I am not going to replicate this a third time, but will exclude recurring tasks for now until the other todo is fixed...

Yes, let's do it this way. I can fix it later on.

kiausch:
How about this?

I'm not super thrilled about the solution with the data attribute, but this is best I could think of :)

Note: I moved the existing data attrs from the enclosing div.nav-child-item in `nav-list-tree.component.html` to the nav-items to have a generic solution which also works for the 'inbox' and 'today' nav-items. The previous data attribute on the div were not being used anywhere as far as I can tell.

johannesjo:
Very cool! I like it a lot! Thank you very much for this and for making the adjustments!

PR #5337, opened 2025-10-22
fix: Initialize isEmoji for icon-input component to avoid showing icon twice

### LINKED ISSUE #5242: 🚨 Icons / emojis shown twice
### Version Used

v15.1.1 snap (Ubuntu 25.04)

### Expected Behavior

Only the single icon should be placed in the field.

### Current Behavior

Project icons in the project settings dialogue are shown twice in edit field.

### Steps to Reproduce

1. Add new project + add an icon on the "Create Project" modal or in project settings
2. Open project settings
3. See the icon got duplicated within the edit field

### Console Output

```shell

```

Abrar74774:
# Description

Apparently the emojis showing twice mentioned in #5242 happens because the preview for mat icons also shows up beside the icon input, even though its supposed to in case the input value is an emoji. This happens because the `onInputValueChange` method, which contains the code to set the signal `isEmoji` to true, does not run on the first time the input value is set through the `ngModel` directive :
```
<input
  [ngModel]="formControl.value"
  (ngModelChange)="onInputValueChange($event)"
  ...
```
  
So to solve this, I just use `ngOnInit` to set isEmoji to `containsEmoji(this.formControl.value)` 

## Issues Resolved

#5242

### DIFF
--- a/src/app/features/config/icon-input/icon-input.component.ts
+++ b/src/app/features/config/icon-input/icon-input.component.ts
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, Component, signal } from '@angular/core';
+import { ChangeDetectionStrategy, Component, OnInit, signal } from '@angular/core';
 import { FieldType } from '@ngx-formly/material';
 import { MATERIAL_ICONS } from '../../../ui/material-icons.const';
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much!

PR #5361, opened 2025-10-25
fix: backlog subtasks are not removed when removing a project

WPum:
# Description

fix: backlog subtasks are not removed when removing a project

## Check List


johannesjo:
Thank you very much!

PR #5370, opened 2025-10-26
feat(reminder): add default task reminder option setting (#5344)

### LINKED ISSUE #5344: 💡 Annoying default reminder – no way to change globally
### Problem Statement

By default when I add a scheduled task in a quick way, the task is always created with the option "Remind at" set to "when it starts". 

Example:
do the laundry @tue 12:30

Image

There is no way to set the default option in the settings - it will always be set to "when it starts" which is annoying because I don't need it and I have to set it manually every time.

Image

### Possible Solution

There should be a way to select a default "Remind at" option in the settings. One of the possible design solutions could be a separate settings section called "Quick task settings" where users would be able to select the default "Remind at" option and possibly some other options! Thanks for attention, this project is amazing!

P.s. This project has been a life saver for 4+ years. Thank you very much for your work.

_No response_

### Additional context

_No response_

miqh:
# Description

Hello there,

This changeset introduces a new setting under the "Reminders" setting group that should enable users to control the default "Remind At" option when creating a new task.

I'll add inline comments against various diffs with further thoughts.

## Issues Resolved

#5344

## Check List

### DIFF
--- a/e2e/tests/reminders/reminders-default-task-remind-option.spec.ts
+++ b/e2e/tests/reminders/reminders-default-task-remind-option.spec.ts
@@ -1,36 +1,80 @@
+import { Page } from 'playwright/test';
+
 import { expect, test } from '../../fixtures/test.fixture';
 
-test.describe('Reminders Settings', () => {
+test.describe('Default task reminder option', () => {
   test.use({ locale: 'en-US', timezoneId: 'UTC' });
 
-  test('should pre-fill new task reminder option from settings', async ({
+  // Should match the option set inside the default global configuration
+  const defaultOptionText = 'when it starts';
+  // Any other option different to the default to test the settings change
+  const changedOptionText = 'never';
+
+  const changeDefaultTaskReminderOption = async (page: Page): Promise<void> => {
+    await page.getByRole('menuitem', { name: 'Settings' }).click();
+    const remindersSection = await page.locator('section', { hasText: 'Reminders' });
+    await remindersSection.click();
+
+    // Should match the option set inside the default global configuration
+    const selectedOption = remindersSection.getByText(defaultOptionText);
+    await expect(selectedOption).toBeVisible();
+
+    // Change it to another option to check whether the setting takes effect
+    // across other application areas where a reminder option can be chosen
+    await selectedOption.click();
+    await page.getByText(changedOptionText).click();
+  };
+
+  test('should apply when scheduling a task using the due action', async ({
     page,
     waitForNav,
   }) => {
     await waitForNav();
+    await changeDefaultTaskReminderOption(page);
 
     await page.getByRole('menuitem', { name: 'Inbox' }).click();
     await page.getByRole('button', { name: 'Due' }).click();
     // Click on the time input to reveal the reminder input
-    await page.getByText(/Time/).click();
+    await page.getByText('Time', { exact: true }).click();
 
-    // Check the default task reminder setting is in effect
-    await expect(page.getByText(/when it starts/)).toBeVisible();
+    await expect(page.getByText(changedOptionText)).toBeVisible();
+  });
 
-    await page.getByText(/Cancel/).click();
-
-    // Change the default task reminder setting
-    await page.getByRole('menuitem', { name: 'Settings' }).click();
-    await page.getByText(/Reminders/).click();
-    await page.getByText(/when it starts/).click();
-    await page.getByText(/10 minutes before it starts/).click();
+  test('should apply when scheduling a task using short syntax', async ({
+    page,
+    waitForNav,
+  }) => {
+    await waitForNav();
+    await changeDefaultTaskReminderOption(page);
 
     await page.getByRole('menuitem', { name: 'Inbox' }).click();
-    await page.getByRole('button', { name: 'Due' }).click();
-    // Click on the time input to reveal the reminder input
-    await page.getByText(/Time/).click();
+    const addTaskInput = await page.locator('.input-section input');
+    await addTaskInput.fill('due task @at 1pm');
+    await addTaskInput.press('Enter');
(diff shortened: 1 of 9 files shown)


e2e/tests/reminders/reminders-default-task-remind-option.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+test.describe('Reminders Settings', () => {
+  test.use({ locale: 'en-US', timezoneId: 'UTC' });
+
+  test('should pre-fill new task reminder option from settings', async ({
+    page,
+    waitForNav,
+  }) => {
+    await waitForNav();
+
+    await page.getByRole('menuitem', { name: 'Inbox' }).click();
```
miqh, line 35: Before I added this test, I looked around for an existing one I could use as a structural basis for the behaviour I'm trying to verify here.

I noticed that the majority, if not all, didn't really abide by Playwright's suggested best practice of [avoiding specific CSS-based selectors](https://playwright.dev/docs/best-practices#prefer-user-facing-attributes-to-xpath-or-css-selectors).

If you'd prefer this test to instead more closely resemble the selector-heavy approach, I'm happy to rewrite this.

src/app/features/tasks/add-task-bar/add-task-bar.component.ts
```diff
@@ -471,7 +471,7 @@ export class AddTaskBarComponent implements AfterViewInit, OnInit, OnDestroy {
           this._taskService.scheduleTask(
             task,
             taskData.dueWithTime!,
-            TaskReminderOptionId.AtStart,
```
miqh, line 478: I believe this would've been a latent bug.

Irrespective of what a user chose for the reminder option, it'll ignore their choice and force the reminder to trigger when the task starts.

miqh:
The failing "Lint & Test PRs" check seems bizarre to me.

https://github.com/johannesjo/super-productivity/actions/runs/18813674297/job/53678879735?pr=5370

When I specifically run `npm run test:tz:la` (i.e. where the errors originate), I don't see any failing tests at all.

~~Am I missing something obvious here?~~

**EDIT:** Ah, I was able to reproduce the same failing tests when I switched over to an Ubuntu workspace (from Windows). Furthermore, these test failures are currently active in the `master` branch, so they're not something this changeset introduces.

yehor-serdiuk:
Wow, thank you very much for your work regarding my issue. Let's hope it gets approved!
I owe you a beer for sure. Love from Ukraine!

johannesjo:
Thank you very much for this @miqh ! Could you have a look at the conflicts? 

I can check the testing issues after merge if it persists.

I also wanted to ask if you would be alright if I'll announce you as contributor of the month on the projects reddit page :) I really appreciate your high quality contributions!

src/app/features/config/global-config.model.ts
```diff
@@ -151,6 +152,7 @@ export type ScheduleConfig = Readonly<{
 export type ReminderConfig = Readonly<{
   isCountdownBannerEnabled: boolean;
   countdownDuration: number;
+  defaultTaskRemindOption: TaskReminderOptionId;
```
johannesjo: We need to make this optional via `?`, since we are not migrating this explicitly. We also need to fallback to the AtStart value when this is not configured to avoid causing problems for some users.

src/app/features/tasks/add-task-bar/add-task-bar.component.ts
johannesjo: We probably still need to fallback to AtStart given users not having set this will have no value set?

miqh:
While making the adjustments requested, I noticed I'd overlooked several other areas where the application was applying a hard-coded `TaskReminderOptionId.AtStart`.

It made sense to me that if we're offering a default task reminder option for it to take effect whenever a task being created can have one set. I was deliberately not exhaustive with this, and there are still some areas where I've left the reminder option alone because they look specific to certain features.

And again, I think the failing "Lint & Test PRs" check is a false positive.

src/app/features/config/global-config.model.ts
miqh: I had (mistakenly) assumed that there was some mechanism to automatically inherit the value defined in `DEFAULT_GLOBAL_CONFIG`. This should now be optional.

johannesjo:
Excellent! Thank you very much for this! Great work! :)

Sorry about the failing tests ... :/

johannesjo:
@miqh How about this? :)

> I also wanted to ask if you would be alright if I'll announce you as contributor of the month on the projects reddit page :) I really appreciate your high quality contributions!

Would you be alright with me making this post? It's no problem if you're not :)

Garrett04:
### 💡Add default task reminder support for schedule page drag-and-drop
I use the schedule page a lot to look over all tasks for each day and would like to have that default task reminder functionality for when moving tasks that don't already have a scheduled time to a specific time in the schedule page and follow the default task reminder setting instead of reminder automatically set to "when it starts".

Current Behavior:
- Drag unscheduled task to a time slot on the Schedule Page.
- Task is scheduled at that time, and has a reminder set to "when it starts".
- User must manually edit the task to apply their preferred reminder time.

Desired Behavior: 
- When an unscheduled task is given a specific time via the Schedule Page drag-and-drop, it should automatically apply the user's defined Default Task Reminder setting.

Keep going with this app! 😁

miqh:
@Garrett04,

Tasks created within the Schedule view should also honour the default remind option setting.



I assume my sequence of actions above is the approach you're describing.

Furthermore, according to the commit history, this new default remind option setting should be available after `v16.3.0`, so please check that you're not using an older version.

Garrett04:
@miqh Oh yes it works that way. Thankyou!

Issue #5025, opened 2025-08-31
💥 OpenProject issue provider does not work, error 400


andyman1222:
### Steps to Reproduce

1. Have an OpenProject set up. My setup is behind a reverse proxy, such that the URL to the main page is like `https://url.com/project` (and for instance the projects list is at `https://url.com/project/projects`)
2. Obtain the access token and project ID as needed
3. Click test, observe that it will report "connection failed". The project can be added but nothing shows up when viewed in the issue panel.

Error console indicates an issue with the filter in the URL being used to test the connection

### Additional Console Output

```
{
    "headers": {
        "normalizedNames": {},
        "lazyUpdate": null
    },
    "status": 400,
    "statusText": "OK",
    "url": "https://url.com/project/api/v3/projects/projectid/work_packages?pageSize=100&filters=%5B%7B%22subjectOrId%22:%7B%22operator%22:%22**%22,%22values%22:%5B%22%22%5D%7D%7D,%7B%22status%22:%7B%22operator%22:%22o%22,%22values%22:%5B%5D%7D%7D,%7B%22author%22:%7B%22operator%22:%22=%22,%22values%22:%22me%22%7D%7D%5D&sortBy=%5B%5B%22updatedAt%22,%22desc%22%5D%5D",
    "ok": false,
    "name": "HttpErrorResponse",
    "message": "Http failure response for https://url.com/project/api/v3/projects/projectid/work_packages?pageSize=100&filters=%5B%7B%22subjectOrId%22:%7B%22operator%22:%22**%22,%22values%22:%5B%22%22%5D%7D%7D,%7B%22status%22:%7B%22operator%22:%22o%22,%22values%22:%5B%5D%7D%7D,%7B%22author%22:%7B%22operator%22:%22=%22,%22values%22:%22me%22%7D%7D%5D&sortBy=%5B%5B%22updatedAt%22,%22desc%22%5D%5D: 400 OK",
    "error": {
        "_type": "Error",
        "errorIdentifier": "urn:openproject-org:api:v3:errors:InvalidQuery",
        "message": "Subject or id can't be blank. and is invalid."
    }
}
```

### Url
file:///C:/Program%20Files/WindowsApps/53707johannesjo.SuperProductivity_14.4.1.0_x64__ch45amy23cdv6/app/resources/app.asar/.tmp/angular-dist/browser/index.html#/config

### Meta Info
[...]

Issue #4628, opened 2025-06-21
💥 OpenProject integration test query syntax


wbranscomb:
### Steps to Reproduce
1. Fresh install of Windows desktop version of Super Productivity 13.1.0
2. Press p to open the issues panel and select Open Project
3. Enter the openproject host (https://host.sub.domain.tld/)
4. Enter the access token generated by OpenProject under Account Settings > Access Tokens > API
5. Enter the Project ID (I've tried the numerical ID like the OpenProject API explorer uses, as well as the project name string that appears in the URL)
6. On the OpenProject server, run `sudo openproject logs --tail`
7. Click Test
8. A "Connection failed" error is displayed.
9. On the OpenProject server, you can see the incoming query and if you pull the filters out of the params section and clean up some of the escaped quotes you get something like this: 
```
[
{"subjectOrId":{"operator":"**","values":[""]}},
{"status":{"operator":"o","values":[]}},
{"author":{"operator":"=","values":"me"}}
]
```
10. We then take this same filter syntax to the OpenProject API explorer to test. (https://host.sub.domain.tld/api/docs)
11. To set up the test in OpenProject we first click API Servers, then select the radio button next to the current host.
12. Next click Authentication, and enter the username `apikey` and for the password use the previously generated API key.
13. Click set. You will be redirected to the home page, but just click back and you should see the Authentication page now shows 1 API key applied.
14. In the left pane,  scroll down to the bottom of the Operations section, click Work Packages, and chose "Get work packages of project".
15. Enter the project ID (if the default demo data is still present either 1 or 2 should work)
17. If we paste the filters from step 9 into the filters section of the Query-String Parameters from and hit execute, we get an error response.
```
{
        "_type": "Error",
        "errorIdentifier": "urn:openproject-org:api:v3:errors:InvalidQuery",
        "message": "Subject or id can't be blank. and is invalid."
}
```
19. If we remove or alter the `{"subjectOrId":{"operator":"**","values":[""]}}` line to either contain a string value or use a different operator we get a valid result returned.
I'm guessing the ** operator used to be allowed to work on empty strings or null values, but it looks like the way some of these operators work (at least with work packages) may have changed.

### Url
file:///C:/Users/usr/AppData/Local/Programs/superProductivity/resources/app.asar/dist/browser/index.html#/tag/TODAY/tasks

### Meta Info
[...]

PR #5373, opened 2025-10-26
Fix for OpenProject filter with empty searchText.

spainer:
# Description

Fixes the problem with the OpenProject API that the field `subjectOrId` must not be empty. Therefore, it is only sent if `searchText` is given and left out otherwise.

## Issues Resolved
- #4628 
- #5025 
- #4166 

## Check List

### DIFF
--- a/src/app/features/issue/providers/open-project/open-project-api.service.ts
+++ b/src/app/features/issue/providers/open-project/open-project-api.service.ts
@@ -62,7 +62,7 @@ export class OpenProjectApiService {
         params: {
           pageSize: 100,
           // see: https://www.openproject.org/docs/api/filters/
-          filters: JSON.stringify(filters),
+          filters: JSON.stringify(filters.concat(this._getScopeParamFilter(cfg))),
           // Default: [["id", "asc"]]
           sortBy: '[["updatedAt","desc"]]',
         },


johannesjo:
Thank you very much for this!

Are you reasonably sure that this won't break anything for other users? I don't have an open project instance available to test currently unfortunately.

johannesjo:
I think there is a problem with this:

- src/app/features/issue/providers/open- project/open-project-api.service.ts:54: The new filters array is no longer concatenated with this._getScopeParamFilter(cfg). As a result, configured scopes such as assigned-to-me / created- by-me stop being applied to search queries, so the search view will suddenly show all open issues instead of the scoped subset.

spainer:
I missed that thing, I am sorry for that.

The fix only applies to the test button in the config. It looks to me that in the panel a search is only triggered as soon as at least one letter is in the search box and then also the old code works. But when the test is triggered, the function is called without a search text and then you get problems with OpenProject as the search value must not be empty.

I checked that with the old code, the test button does not work, but the new one works. When the provider has beend added, both version do work as the search is then only triggered when there is at least one character.

Automatic retrieval of taks in the background also works with the new code. I did not check if it worked with the old one...

johannesjo:
Thank you very much for the update!

PR #5397, opened 2025-10-31
prevent inline-input element from collapsing due to empty value

### LINKED ISSUE #5393: Manual editing of work hours in daily-summary view is only possible if stop watch has been started at least once
### Problem Statement

It is possible to manually edit the work hours in the daily summary:
Image

However, if the stop watch has not been used at least once on that day, the input shows "-" and it is not possible to click and change the values:
Image

### Possible Solution

Clicking on the "-" should add a time record and focus it for editing

_No response_

### Additional context

_No response_

kiausch:
# Description

The input-wrapper-width collapsed to 0 due to the empty value; Because of that the hover rule which reveals the hidden inline input field did not apply.
Fixed by setting a min-width to the wrapper element in the summary-point context.

## Issues Resolved

closes #5393

### DIFF
--- a/src/app/ui/inline-input/inline-input.component.scss
+++ b/src/app/ui/inline-input/inline-input.component.scss
@@ -18,6 +18,12 @@
   }
 }
 
+:host-context(.summary-point) {
+  .inline-input-wrapper {
+    min-width: 35px;
+  }
+}
+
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much!

Issue #5192, opened 2025-09-29
💡 Specify & edit timestamps for Created and Completed


ltguillaume:
### Problem Statement

I want to set the done date to the date I actually completed the task. This is impossible if:
1. I have forgotten to mark it done in time
2. I want to add a task "posthumously" for logging reasons
3. the sync progress messed things up and I have to mark tasks as done again

### Possible Solution

1. Allow the user to specify the Completed timestamp on marking the task as done
2. Allowing to edit the timestamps for Created and Completed in the sidebar (additional info)

Issue #4333, opened 2025-05-13
💡 feature request editable complete date time


JunkyardCat:
### Problem Statement

so when we complete a task the complete date is uneditable afterwards. but if im away from keyboard and completed the task much early i want it reflected on the task itself so when i review it in the future i know exactly when i have completed it

### Possible Solution

edit the completed tasks completed date and time

### Describe alternatives you've considered

as suggested by one of the maintainer export edit then import back again

### Additional context

discussion with one of the maintainer
https://github.com/johannesjo/super-productivity/discussions/4311

johannesjo:
This has been impleneted.

PR #5403, opened 2025-11-01
feat(task): add ability to edit task created and completed

miqh:
# Description

Hello there,

This is a shot at adding the ability to edit the created or completed (`doneOn`) datetime value of a task.

Both the existing lines of information showing created and completed time when viewing task details have been converted into interactable items along with the existing ones on top.

I looked around to see whether I could repurpose any existing dialogs, but I couldn't find one that was fit for purpose without me having to go and butcher them. So I created a new and simple one which simply asks the user for a datetime selection. I think it's in a position to be extended for reuse across multiple contexts too, should the need arise.

Some other thoughts:

- I'm unsure how catastrophic it might be to data integrity or existing productivity metrics if a user decides to set the completed prior to created. I've not added any safeguards for this.
- I've not changed the locale formatting of the datetime values that would've been showing for created and completed. I understand there's probably another issue about making this consistent across the application.
- Fortunately, `task-shared-crud.reducer.ts` is capable of updating both `created` and `doneOn` if they're supplied, so that spared me work there.

## Issues Resolved

- #5192
- #4333

## Check List

### DIFF
--- a/e2e/tests/task-detail/task-detail.spec.ts
+++ b/e2e/tests/task-detail/task-detail.spec.ts
@@ -27,66 +27,15 @@ test.describe('Task detail', () => {
     await page.getByRole('button', { name: 'Mark as done/undone' }).click();
   };
 
-  const findTaskDetailItem = (
-    page: Page,
-    title: string,
-    requireValue = true,
-  ): Locator => {
-    let item = page.locator('task-detail-item').filter({
-      has: page.locator('.input-item__title', { hasText: new RegExp(title, 'i') }),
-    });
-    if (requireValue) {
-      // Cannot assert against specific datetime values since they are being
-      // dynamically generated. Match conditions also deliberately avoid
-      // assuming specific locale formatting requirements for leniency.
-      item = item.filter({ has: page.locator('.input-item__value', { hasText: /\d+/ }) });
-    }
-    return item;
-  };
-
-  test('should show created for task', async ({ page, workViewPage }) => {
-    await addAndOpenIncompleteTask(workViewPage, page);
-
-    await expect(findTaskDetailItem(page, 'Created')).toHaveCount(1);
-  });
-
-  test('should not show completed for incomplete task', async ({
-    page,
-    workViewPage,
-  }) => {
-    await addAndOpenIncompleteTask(workViewPage, page);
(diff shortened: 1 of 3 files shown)


johannesjo:
Thank you very much for this! Works quite well!

One thought: The previous more subtle design for Created and completed was choosen deliberately. It is an information that interesting to see, but less important than the other items imho and certainly something that needs to be changed less, since usually the auto-assigned value should be correct. What do you think? Would you be alright with returning to the previous style and open the dialog only when clicked on those?


> I'm unsure how catastrophic it might be to data integrity or existing productivity metrics if a user decides to set the completed prior to created. I've not added any safeguards for this.

I think (hope) we should be fine :D

miqh:
> What do you think? Would you be alright with returning to the previous style and open the dialog only when clicked on those?

As you wish. I've undone the structural changes to `TaskDetailPanel` to preserve the appearance of both the created and completed times as they were before. The only minor difference is the addition of a hover style, indicating that the elements can be clicked on to spawn the datetime editing dialog.

My lingering concern is that the ability to edit created or completed isn't an immediately obvious feature to users (i.e. no visuals to indicate it can be interacted with). Furthermore, the relative proximity of both actions may lead to users to edit the field they didn't intend to do (especially on mobile devices where a finger tap may be less precise than a mouse click).

The "Lint & Test PRs" check seems to be failing for tests that all pass locally for me again. This has come up before in another changeset I submitted, so it may be a time of day sensitive error.

johannesjo:
> My lingering concern is that the ability to edit created or completed isn't an immediately obvious feature to users (i.e. no visuals to indicate it can be interacted with). Furthermore, the relative proximity of both actions may lead to users to edit the field they didn't intend to do (especially on mobile devices where a finger tap may be less precise than a mouse click).

I think if we provide a pointer cursor, that might be enough to indicate this for desktop users. For mobile you're right, this is a bit akward, but I am OK with the trade off, since it is really not an everyday thing to do, since these values usually are set automatically. I would priorize the no clutter thing in this particular instance.

Sorry, about the failing tests. I think I might have broke them on the weekend... :/

johannesjo:
Thank you very much for your great work!

PR #5414, opened 2025-11-01
feature addition: user profiles

overcuriousity:
# Description

This PR introduces a comprehensive user profiles system that allows users to maintain multiple isolated instances of Super Productivity, each with separate tasks, settings, and sync configurations.
**Scenario I was trying to solve:**
For example, I use Super Productivity for work AND for my houshold chores.

- The work instance contains tasks I would do at work, such as meetings etc, and may also contain even confidential data.
- The household instance ontains tasks like feed the dog, make laundry, clean the cat toilet etc.
- I dont want to have the cat toilet show up in the work instance, or confidential data show up in my private instance.
- I might even have different auto-sync endpoints configured - my private nextcloud WebDAV for my household, my work nextcloud with E2E for my work related stuff.
- Different configurations for the auto-planning of tasks - the work instance should plan my tasks 9-5, while my household instance is not restricted

Implementation:
The profiles feature leverages the existing `pfapi` (Platform-agnostic API) backup and restore infrastructure:

- **Profile Storage Format**: Each profile is stored using the same `CompleteBackup<T>` data structure that powers the existing backup/export functionality. This includes:
  - All tasks, projects, and tags
  - Global configuration and settings
  - Time tracking data
  - Reminders and bookmarks
  - Archive data (young and old)
  - Plugin data and metadata

- **Profile Switching Mechanism**:
  1. Current profile data is saved using `pf.loadCompleteBackup()` (same as manual backup)
  2. Target profile data is loaded from storage
  3. Data is imported using `pf.importCompleteBackup()` (same as restore from backup)
  4. Application reloads to ensure clean state

- **Storage Abstraction**: Profile data is stored platform-specifically:
  - **Electron**: Files in `<userData>/profiles/{profileId}.json`
  - **Web/PWA**: Browser localStorage with `sp_profile_data_` prefix
  - **Android**: Native storage via `androidInterface`

This approach ensures that profiles benefit from all existing validation, repair, and migration logic built into the backup system.

**Needs work**: Profile data includes all application data in its original form, which means:

Credentials stored in plain json:

- **Sync Provider Tokens**: Dropbox OAuth tokens, WebDAV credentials, local file sync paths
- **Issue Tracker Credentials**:
  - Jira: Username, password, or API token
  - GitHub: Personal access tokens
  - GitLab: Personal access tokens
  - Gitea: Personal access tokens
  - OpenProject: API tokens
  - CalDAV: Username and password
- **Global Configuration**: Unsplash API keys, proxy passwords

Some sort of encryption should be implemented, but i didnt come up with an idea for that.

**Migration from previous non-profile versions:**
"Default" profile is created.
  - Profile metadata created: { activeProfileId: "default", profiles: [...] }
  - Existing data copied to: profiles/default.json (Electron) or localStorage (Web)
  - Main database unchanged
  - App continues using main database
  - User sees profile button in header

Survived basic testing on my side. 
I hope this helps iterating to implement this feature.

AI (Claude 4.5 Sonnett) was used for implementing the new feature, manual code review was done.

### DIFF
--- a/src/app/features/config/store/global-config.effects.ts
+++ b/src/app/features/config/store/global-config.effects.ts
@@ -172,7 +172,14 @@ export class GlobalConfigEffects {
           ({ sectionCfg }) =>
             sectionCfg && (sectionCfg as MiscConfig).isEnableUserProfiles !== undefined,
         ),
-        tap(({ sectionCfg }) => {
+        withLatestFrom(this._store.select('globalConfig')),
+        filter(([{ sectionCfg }, globalConfig]) => {
+          const newValue = (sectionCfg as MiscConfig).isEnableUserProfiles;
+          const oldValue = globalConfig?.misc?.isEnableUserProfiles ?? false;
+          // Only proceed if the value actually changed
+          return newValue !== oldValue;
+        }),
+        tap(([{ sectionCfg }]) => {
           const isEnabled = (sectionCfg as MiscConfig).isEnableUserProfiles;
 
           // Update localStorage flag for fast startup check


johannesjo:
Hey hey! Thank you very much for this! 

This is a big new feature impacting a lot of stuff. I must say that It would have been better to discuss something like this first, before providing a PR :D

Since quite a few people (me included) won't need this feature I think we should aim to minimize impact on them to avoid unncessary confusion and mental load. I think it would be better not to display the header button as per default, but let users enable the feature in the settings. 

I also likely won't have time to test and support the feature myself in case there are problems as there is so much other stuff to do. Would you be willing to also work on fixes for this, if something comes up?




There is also a minor styling issue with the management for dark mode:

Screenshot_20251103_110215

johannesjo:
I am also wondering why not use different users on your OS (or just different projects and project folders)? If you want a clean separation of work and non-work, this is the cleanest way to go about.

overcuriousity:
Thank you for the Feedback!

You are right, using different users/OSes would be the cleanest approach. But at least in my case, something like this is more comfortable, but I agree to your argument.

It would also make sense to make this feature optional and hidden by default to keep it clean.

But if you are nonetheless willing to merge it upstream as soon as it meets the requirements, I can confirm that I would maintain and test this feature going forward.

overcuriousity:
And the reason why different projects for work/private didnt work out for me:
We can set a timeframe for auto scheduling tasks in work hours e.g. 9 to 5 with a break. But this setting is global. So if I set work hours 9 to 5, there would no private tasks be automatically scheduled in the off hours.

I could also imagine a user working and studying part-time, so he wants to manage monday-wednesday with the work schedule, and thursday-friday with a different study schedule. Just switching the profile along with all related settings and contents feels like the cleanest solution if you dont want to use different OS accounts.

overcuriousity:
Progress made:

- feature disabled by default. Can be enabled in the "Misc Settings" menu on the bottom, has a description and a beta disclaimer.
- description removed from the "Manage Profiles" menu, as it is in the main menu. Fixed the dark mode display issues.
- Made the feature more stable and covered some edge cases. 
- Allows Switch on/off without data loss (or should at least, tested it). Displays a user warning and recovery guidance when switching off after having it enabled.
- Localization

johannesjo:
Let me know if this is ready to test again.

overcuriousity:
Should be ready to go.
Passes all tests and worked as expected with my testing.

src/app/pages/config-page/config-page.component.ts
```diff
@@ -217,18 +222,61 @@ export class ConfigPageComponent implements OnInit, OnDestroy {
...
+
+    // Check if user is trying to disable user profiles when multiple profiles exist
+    if (
+      sectionKey === 'misc' &&
+      config.isEnableUserProfiles === false &&
+      this._userProfileService.hasMultipleProfiles()
+    ) {
+      const currentMiscConfig = this.globalCfg?.misc;
+      // Only show dialog if we're actually changing from true to false
+      if (currentMiscConfig?.isEnableUserProfiles !== false) {
```
novikov1337danil: This would be more reliable, because `currentMiscConfig` can be `undefined`

```suggestion
      if (currentMiscConfig?.isEnableUserProfiles === true) {
```

electron/profile-storage.ts
```diff
@@ -0,0 +1,82 @@
...
+import { error, log } from 'electron-log/main';
+import { IPC } from './shared-with-frontend/ipc-events.const';
+
+const PROFILE_STORAGE_DIR = path.join(app.getPath('userData'), 'profiles');
+
+/**
+ * Initialize profile storage IPC handlers
+ * Profiles are stored in: <userData>/profiles/
+ */
+// eslint-disable-next-line prefer-arrow/prefer-arrow-functions
```
novikov1337danil: I would suggest following the rules defined in the project

novikov1337danil:
Although I don't need this feature, but I like it :+1: You did a great job!

I’m interested if this feature is popular and if it exist in alternative apps?

overcuriousity:
@novikov1337danil Nice to hear you like it!

I dont know if this feature is popular, can´t think about many apps which can really be compared to this (great) one.
Just makes sense for my use-case (having a day-job and studying in the evening, which shouldnd be mixed up, but i want to manage both easily from one OS account) - which might not be that common.

src/app/features/user-profile/dialog-user-profile-management/dialog-user-profile-management.component.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  readonly newProfileName = signal('');
+  readonly editingProfileId = signal<string | null>(null);
+  readonly editingProfileName = signal('');
+  readonly isCreating = signal(false);
+
+  readonly isElectron = IS_ELECTRON;
+
+  async createProfile(): Promise<void> {
+    const name = this.newProfileName().trim();
```
novikov1337danil: It seems unnecessary since the trimming is already executed in the `profileService.createProfile` method itself

src/app/features/user-profile/dialog-user-profile-management/dialog-user-profile-management.component.ts
```diff
@@ -0,0 +1,139 @@
...
+    if (!name) {
+      this._snackService.open({ type: 'ERROR', msg: 'Profile name cannot be empty' });
+      return;
+    }
```
novikov1337danil: It seems unnecessary since it duplicates the existing check in the `profileService.createProfile` method

src/app/features/user-profile/user-profile.service.ts
```diff
@@ -0,0 +1,446 @@
...
+  /**
+   * Create a new profile
+   */
+  async createProfile(name: string): Promise<UserProfile> {
+    const metadata = this._metadata();
+    if (!metadata) {
+      throw new Error('Profile system not initialized');
+    }
+
+    if (!name || name.trim().length === 0) {
```
novikov1337danil: The first part of the condition is unnecessary. I would suggest removing it
```suggestion
    if (name.trim().length === 0) {
```

src/app/features/user-profile/user-profile.service.ts
```diff
@@ -0,0 +1,446 @@
...
+      throw new Error('Profile name cannot be empty');
+    }
+
+    if (metadata.profiles.some((p) => p.name === name)) {
```
novikov1337danil: I see a small discrepancy: you search for duplicates by exact `name` match, but when creating, you save its trimmed version

electron/profile-storage.ts
overcuriousity: Resolved it - consistent to other sections of the code i didnt see before.

src/app/features/user-profile/dialog-user-profile-management/dialog-user-profile-management.component.ts
overcuriousity: removed the redundant empty name validation.

src/app/features/user-profile/user-profile.service.ts
overcuriousity: made changes to the createProfile() and renameProfile() to search for the trimmed name consistently.

johannesjo:
Thank you for the update @overcuriousity ! 

Unfortunately this is still interfering with the regular app start and slowing it down. There is a complete load of data initially and it starts migrating stuff even if the feature is not activated. Maybe we can use a localStorage flag to check if the feature was activated on the last app start? I am ok with merging this feature as long as the impact on regular usage is minimal.

I am also wondering if we need the extra handling for electron and capacitor app. Couldn't we just re-use the same mechanisms as used for web? 
 
Since this is more of a niche feature it would be ideal if it just could be a **plugin** rather than something that interfers with the regular flows at such a fundamental. Not sure if this is technically feasable though.

I hope it is not too annoying for you that I am this hesitant with merging the feature. This really well done and a clean implementation and it really adds value for those who encounter the same use case as you. I just want to avoid trouble this might be causing directly and in the long run.  Having this as a plugin would be a wonderful addition. Having this entangled with the core parts of the app while not being a standard feature for all users doesn't feel well, if  I'm honest.

overcuriousity:
The argument is reasonable and I agree that the overall UX should not be impacted negatively.
Anyways, I could not feel any kind of speed impact on my system.

So I might look how the plugin system is working - Cannot estimate how long code rewrite would take as I need to familiarize with it first.

overcuriousity:
I came to the conclusion that the feature is not technically feasible to be implemented via a plugin, as it would require access to either the pfapi module or the IndexedDB directly to do any kind of full-scope backup and import, with complete separation of concerns.

With the current plugin API, I cannot implement it as a plugin as the feature is imagined.

However, I adressed:

- const isProfilesEnabled in data-init.service.ts is only a quick check if the feature is enabled, if not, no profiles-related code is executed
- Migrations **should** only happen **once** after a user enables the feature now (the function migrateOnFirstEnable() skips when expected metadata exists in the profile)
- streamlined the handling of the localStorage, no electron specifics, handles all through user-profile-storage.service.ts

I would consider the implementation minimal in terms of impact to users which dont enable it, and startup speed.
Feel free to let me know about your decision or any solution I might not have come up with.

johannesjo:
Thank you! The init stuff is much better now. I am ok with merging this as experimental feature (though I might remove it, if I encounter any problems with it). 

There is one little bug. When the feature is enabled and there are changes to the misc config, it will reload the app every time, which is probably not what we want.

overcuriousity:
should be resolved!

johannesjo:
Thank you very much for your patience and the effort! I am merging this :)

PR #5420, opened 2025-11-02
feat: improve automatic scheduling for tasks with no planned time

overcuriousity:
# Description
Small improvement.
Tasks which have no time assigned to them are placed in the timetable view somewhat arbitrarily.
I wanted to give power users the ability to control that. The scheduling (only affects time table view) borrows principles from well known process scheduling algorithms (we know from IT that SJF is the best one to get the most completed tasks/time) and places the tasks on the current day determined on parameters creation timestamp and estimated duration.
The scheduling algorithm can be selected in the menu "Schedule".
Also integrated a "best fit" mapping which places the auto-planned tasks around ones which are planned.

The impact on the user experience is very low as anything should work as before, manual adjustments take precedence. 
While the "best fit" strategy is a net positive for most users, the scheduling logic is mainly for advanced users and the default settings are unchanged. But I would propose to make the "SJF" setting the default.

It might need tuning, and the menu entries are english-only. Also the settings could be explained better in the menu.
I think the feature is so small that we dont need to mention it explicitly in the documentation.

## Check List

AI (Claude Sonnett 4.5) was used. The code was manually reviewed and adjusted. Code changes are minimal.

### DIFF
--- a/src/app/features/schedule/map-schedule-data/map-to-schedule-days-extended.spec.ts
+++ b/src/app/features/schedule/map-schedule-data/map-to-schedule-days-extended.spec.ts
@@ -142,116 +142,29 @@ describe('mapToScheduleDays()', () => {
       p.lunchBreakCfg as any,
     );
 
-    expect(r[3]).toEqual({
-      beyondBudgetTasks: [],
-      dayDate: '2024-08-05',
-      entries: [
-        {
-          data: {
-            defaultEstimate: 900000,
-            id: 'wSs2q4YWkZZjthJrUeIos',
-            isPaused: false,
-            lastTaskCreationDay: getDbDateStr(1722325722970),
-            monday: true,
-            order: 0,
-            projectId: 'lMLlW2yO',
-            quickSetting: 'CUSTOM',
-            remindAt: 'AtStart',
-            repeatCycle: 'WEEKLY',
-            repeatEvery: 1,
-            startDate: '2024-07-30',
-            startTime: '9:00',
-            tagIds: ['TODAY'],
-            title: 'Do something scheduled on a regular basis',
-          },
-          duration: 900000,
-          id: 'wSs2q4YWkZZjthJrUeIos_2024-08-05',
-          start: 1722841200000,
-          type: 'ScheduledRepeatProjection',
-        },
-        {
-          data: {
-            defaultEstimate: 1800000,
-            friday: false,
-            id: 'lmjFQzTdJh8aSak3cu9SN',
-            isPaused: false,
-            lastTaskCreationDay: getDbDateStr(1722275904553),
-            monday: true,
-            order: 0,
-            projectId: 'lMLlW2yO',
-            quickSetting: 'WEEKLY_CURRENT_WEEKDAY',
-            repeatCycle: 'WEEKLY',
-            repeatEvery: 1,
-            saturday: false,
-            startDate: '2024-05-06',
-            sunday: false,
-            tagIds: ['TODAY'],
-            thursday: false,
-            title: 'Plan Week',
-            tuesday: false,
-            wednesday: false,
-          },
-          duration: 1800000,
-          id: 'lmjFQzTdJh8aSak3cu9SN_2024-08-05',
-          start: 1722842100000,
-          type: 'RepeatProjection',
-        },
-        {
-          data: {
-            defaultEstimate: 1200000,
-            friday: true,
-            id: 'Foclw2saS0jZ3LfLVM5fd',
-            isPaused: false,
-            lastTaskCreationDay: getDbDateStr(1722617221091),
-            monday: true,
-            order: 32,
-            projectId: 'DEFAULT',
-            quickSetting: 'MONDAY_TO_FRIDAY',
-            repeatCycle: 'WEEKLY',
-            repeatEvery: 1,
-            saturday: false,
-            startDate: '2024-07-30',
-            sunday: false,
-            tagIds: ['TODAY', 'DZHev64ka8kt4olVAujAe'],
-            thursday: true,
-            title: 'Also scheduled in the morning',
-            tuesday: true,
-            wednesday: true,
-          },
-          duration: 1200000,
-          id: 'Foclw2saS0jZ3LfLVM5fd_2024-08-05',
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! 

There are also potential problems with this:

When you drag an event in schedule and press the shift key you can sort a task to the end or start of day. This likely won't work with any of the scheduling mechanisms in place. The idea behind the schedule is that tasks are sorted according to the manual order the user provided in today list and planner (the priorization). 

Thinking about it: Maybe this makes more sense as a sort option in the the filter/group/sort panel (we would have to make sorting permanent for this)? 

Also there is an linting issue and some conflicts.

overcuriousity:
Thanks for the feedback.

Wasnt aware of the shift key mechanic - wasnt really obvious to me. But nonetheless, this existing logic **should** already override any scheduling implemented by the PR.

And indeed, what you propose would just a change to the default sorting behavior, which could be made permanent by the menu setting. But in the current version this does not affect the order of the timetable, just the current day view, am I right?

Also I tried to face the problem of overlapping planned tasks (or the break) with unplanned (auto-scheduled) tasks with the "best fit" logic. 

What do you think would be best?

johannesjo:
> But in the current version this does not affect the order of the timetable, just the current day view, am I right?

Yes!

johannesjo:
Forget my previous comment (if you read it before I edited it). I forgot that the scheduled tasks are sorted differently model wise.

Not sure not sure...  I feel this feature might confuse users even more than they already are :D 

Maybe best is to provide some sort button like on the main view, but this time just for schedule and planner?

Give me some time to think about this!

overcuriousity:
Sure!

johannesjo:
Hey hey! I thought about this a bit more.  Thanks for your patience!

I think it might be best to make this a one time permanent sort feature triggered by a menu similar to the one that is shown on the standard task list. Since it would affect planner and schedule, I think might make sense to show the button for this on both views.

What do you think? Does this make sense to you? Is it feasable?

overcuriousity:
Yes, this sounds very feasible and much more accessible than my approach, making it a more obvious option for a larger set of users.

I dont know if this includes fixing the problem of unintentional overlapping of tasks (if that´s an issue).

johannesjo:
> I dont know if this includes fixing the problem of unintentional overlapping of tasks (if that´s an issue).

Can you explain the problem to me? At first glance that sounds like a seperate issue to me.

overcuriousity:
image

Yes, its kind of a different issue, which I tried to resolve with the "best fit".

I have :

- a task "cat toilet 30m" which is planned at 14:00
- a task ""Paperwork 4h" which is not planned at a specific time

Result:
Schedule overlaps the unplanned longer task behind the planned shorter task.
Should: 
Place the longer task in the next best fit, mens the next free gap which is large enough for the task.

johannesjo:
Thanks for the explanation! Could we maybe also offer as a custom sort option for this?

overcuriousity:
Technically this would be a different logic, not actually related to any sorting.
TBH I think most users would expect the no-overlap as a default.

overcuriousity:
I unintentionally closed the PR (sorry for my still-bad git skills), but then i noticed, you implemented the feature with commit 267b1c0.
Thank you for that! I also couldnt notice any of the problems we had before.

Feel free to close the PR unless you think I could still contribute on this feature.

johannesjo:
I only implemented the sorting for the task lists, but not for schedule and planner. So I think, this feature request is still valid :)

johannesjo:
Please let me know, when this is ready for review!

overcuriousity:
image

Implementation done.

- we have a menu option (boolean) which controls wether splitting tasks around planned tasks and breaks is allowed or not
- The screenshot displays the "best fit" algorithm (only available if splitting tasks are forbidden) in action, effectively minimizing gaps between fixed-planned tasks/breaks
- we have a dropdown in the schedule view which controls a persistent config option for how the schedule plans the tasks
- the first setup wizard presents these options both when invoked
- the default (current) behavior is the default setting

overcuriousity:
npm run test completes successfully in the provided dev container - finally.
Would you like to trigger the checks again?

johannesjo:
@codex

johannesjo:
Thank you!

johannesjo:
Hey @overcuriousity,
thanks again for the suggestion and for pushing this forward. After giving it a serious try, I have decided to remove the feature again.

I invested more than half a work day into getting it to work well, but it introduced issues in multiple parts of the app (affected were backwards compatbility, cross timezone unit tests and several ui problems). It would also require additional explanations and UI flows to clarify why sorting doesn’t behave as expected when filters are active.

On top of that, it adds complexity to an area of the app that is already quite intricate. My sense is that this would make future maintenance and development quite a bit harder.

I also tested the feature in practice, and while it’s fun to experiment with different orderings, it felt more like a gimmick than something that meaningfully improves the workflow. My personal feeling is, that manual ordering is the better option and via drag and drop it should be quite quick, too. 

I think it also adds potential complexity and confusion to new users , since it has a very technical feel to it.

I am very sorry for the time you sunk into this and I really do appreciate the effort! And I 'm sorry for pushing for the additional changes. I did this, because I really wanted to make this work. But ultimately I think it is best not to add this functionality, at least in the current form.

In the future before adding these complexer features, it would be great if you could open up an issue to discuss it first.

Thank you very much once more. I really hope you can understand my reasoning.

overcuriousity:
well, I can understand your reasoning. I admit its quite technical and gimmicky and might only contribute on my approach on how it improves my workflow.

Also I see why maintenance is a thing, as adding 1000 lines of code was quite an impact - its your project after all - no need to be sorry.

I might maintain it as a personal fork, though, which might mature it enough to get merged some day in the future, while I have to attend other projects also, so no guarantees.

Feel free to contact me if your preferences might change and we could revisit this feature again.

johannesjo:
> I might maintain it as a personal fork, though, which might mature it enough to get merged some day in the future, while I have to attend other projects also, so no guarantees.

That's a good idea! That's the beauty of open source.

Thank you very much for your understanding! I really appreicate it. I don't find it easy to handle these kind of situations.

PR #5428, opened 2025-11-03
Feat/modify break time in daily summary

kiausch:
# Description

Allows to manually add/edit break nr and duration in daily-summary.

If either the break number or duration is set to zero, the other value is also set to zero, as zero breaks with a duration or non-zero number of breaks with no duration does not make sense.
If the duration is set first while the number is zero, the number is increased to 1.

Slightly reworked the stylesheet for the summary-point boxes to use flex layout for the input / value line


johannesjo:
Very nice! Thank you very much! <3

PR #5430, opened 2025-11-03
fix(i18n): fix json syntax in zh-tw locale

novikov1337danil:
# Description

While I was working on localizing the application, I noticed that a comma was missing.

I don't know how I missed this in last PR https://github.com/johannesjo/super-productivity/pull/5372 :disappointed:


johannesjo:
Thank you very much!

PR #5450, opened 2025-11-05
Update & fix some translated Variables

ExoGitH:
# Description

follow up changes to en.json and fix some variables that was translated

## Issues Resolved

-


johannesjo:
Thank you ver ymuch!

PR #5468, opened 2025-11-08
Adding Russian metadata translation

Korb:
https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/#in-the-applications-source-repository


johannesjo:
Thank you very much!

PR #5473, opened 2025-11-09
fix(mobile-nav): check history.state for null before access the key

novikov1337danil:
# Description

Ensuring `window.history.state` is not `null` before accessing

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/5472


johannesjo:
Thank you very much!

PR #5480, opened 2025-11-10
fix: Tag selected and deselected on every other keystroke (#5452)

### LINKED ISSUE #5452: 🚨 Tag is selected and deselected on every other key input, while adding task
### Version Used

16.2.1

### Expected Behavior

The tag should remain selected 

### Current Behavior

The tag keeps getting selected and deselected on every other keystroke, like so:

Image

### Steps to Reproduce

1. Click on add new task button on top
2. Start writing task title
3. Input '#' followed by tag name, select desired tag
4. Keep writing something more

### Console Output

```shell

```

Abrar74774:
# Description

This fixes the linked issue.

## Issues Resolved

#5452

### DIFF
--- a/src/app/features/tasks/add-task-bar/add-task-bar-state.service.spec.ts
+++ b/src/app/features/tasks/add-task-bar/add-task-bar-state.service.spec.ts
@@ -219,6 +219,28 @@ describe('AddTaskBarStateService', () => {
     });
   });
 
+  describe('updateTagIdsFromTxt', () => {
+    it('should replace all tags derived from short syntax', () => {
+      const mockTags: Tag[] = [
+        { id: '1', title: 'urgent' } as Tag,
+        { id: '2', title: 'important' } as Tag,
+      ];
+
+      service.updateTagIdsFromTxt(mockTags.map((t) => t.id));
+
+      expect(service.state().tagIdsFromTxt).toEqual(mockTags.map((t) => t.id));
+    });
+
+    it('should clear tags derived from short syntax when empty array passed', () => {
+      const mockTags: Tag[] = [{ id: '1', title: 'urgent' } as Tag];
+      service.updateTagIdsFromTxt(mockTags.map((t) => t.id));
+
+      service.updateTagIdsFromTxt([]);
+
+      expect(service.state().tagIdsFromTxt).toEqual([]);
+    });
+  });
+
   describe('updateNewTagTitles', () => {
     it('should update new tag titles', () => {
       const newTagTitles = ['new-tag', 'another-new'];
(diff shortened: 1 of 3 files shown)


Abrar74774:
I see some tests are failing. I'll close this PR for now, will reopen after I fix the issue.

Abrar74774:
@johannesjo I've reverted my last changes, and resorted to this current simple change. The thing with this change though, is that while typing in the tag syntax (#tag) the tag gets selected as expected, but while erasing, the tag stays selected - would like to know if this is ok.
tag


Previously, I thought it would be a better UX to deselect the tag while erasing its syntax. To achieve this I made the previous solution which was to update the tagIds array on any changes detected from parsing - both adding and removing tag syntaxes. But from the failing tests I realized that this overwrites any preselected tags (through the tag dropdown that you get after clicking on the tag labels). The proper way to acheive this might require an additional state field to track if a tag is added from the dropdown or from typing in - a bit complex. So for now I just made the current simple change 🙂

johannesjo:
> Previously, I thought it would be a better UX to deselect the tag while erasing its syntax. To achieve this I made the previous solution which was to update the tagIds array on any changes detected from parsing - both adding and removing tag syntaxes. But from the failing tests I realized that this overwrites any preselected tags (through the tag dropdown that you get after clicking on the tag labels). The proper way to acheive this might require an additional state field to track if a tag is added from the dropdown or from typing in - a bit complex. So for now I just made the current simple change 🙂

I also ran into this before. Tricky one to solve. 

There is one very problematic case:
Imagine we have two tags `#aaa`  and `#aaaaaaa`. When we type `#aaaaaa`  both tags get added, which is not what we want. Maybe if we keep track of if a task was added via short syntax or not, so we can always keep those during typing that are not added via short syntax and always remove those added via short syntax that are not present anymore What do you think?

Abrar74774:
> I also ran into this before. Tricky one to solve.
> 
> There is one very problematic case: Imagine we have two tags `#aaa` and `#aaaaaaa`. When we type `#aaaaaa` both tags get added, which is not what we want. Maybe if we keep track of if a task was added via short syntax or not, so we can always keep those during typing that are not added via short syntax and always remove those added via short syntax that are not present anymore What do you think?

Hmm yes I can try that. Will add more commits on this

Abrar74774:
@johannesjo so I've changed the overall functionality of parsing for tag changes. Ive added a field `tagIdsFromTxt` to the add-task-bar state to represent the list of tags derived from parsing, while the existing `tagIds` field represents tags selected from dropdown. Any updates from parsing short-syntax is done to the `tagIdsFromTxt` array.

Ive changed the functionality of short-syntax.ts to overwrite list of tags if there are any changes in the tag list, to handle both removing tag syntax by erasing as well as adding.

I also updated/added relevant tests. Im not an expert on writing tests 😅 so I just pretty much followed the structure of existing tests of the components I worked on.

johannesjo:
i just merged this locally with some adjustments. Thank you very much for this!

Issue #5465, opened 2025-11-08
🚨 Wrong ordering of months and years in worklog view


v-pnk:
### Version Used

16.3.0 snap

### Expected Behavior

The months in worklog view to be ordered as the most recent month on top and the first (oldest) month at bottom. The years to be ordered the same - most recent at the top and the first at the bottom.

### Current Behavior

Both years and months are ordered in alphabetical order of their numerical string.
Years are ordered from the first (e.g. 2022) at the top, to the most recent (2025) at the bottom.
Months are ordered as: 1, 10, 11, 12, 2, 3, 4, 5, 6, 7, 8, 9

### Steps to Reproduce

1. Go to the the Worklog (in "Today" or in any project) and check the ordering

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

aripollak:
I'm also seeing this. I'm guessing it was introduced by [this](https://github.com/johannesjo/super-productivity/commit/30ca71e518e87af1f71e578b5b518dbbd36c66a7#diff-48bbcff14941be677b05526fbc8fd75b75bae59dceac00fe706ca4abffc4a77fL152-R158) change in `worklog.component.ts`. I'm not sure why that was done since it seems like `sortWorklogItems` is only currently called with numbers, but not with the right types?

johannesjo:
I think this will be fixed with the next release thanks to miqh!

novikov1337danil:
This has already been solved. Can be closed

PR #5396, opened 2025-10-30
refactor: migrate to control flow syntax

lekhmanrus:
# Description

This PR updates Angular templates across the project to use the new Angular control flow syntax (`@if`, `@for`, `@switch`, etc.) introduced in Angular v17, replacing the legacy structural directives (`*ngIf`, `*ngFor`, `*ngSwitch`).

## Issues Resolved

Migrating to the modern Control Flow syntax:
- Improves template performance and reduces generated code size.
- Provides cleaner and more maintainable template logic.
- Aligns with the latest Angular best practices and future-proof features.

## Check List

- [ ] New functionality includes testing.
- [ ] New functionality has been documented in the README if applicable.

PR #5482, opened 2025-11-11
fix(worklog): revise comparators for key-value pipe (#5465)

miqh:
# Description

Hello there,

This changeset should address the undesirable ordering of worklog data that was recently reported by some users.

The cause appears to be a regression introduced by the following pull request:

- #5396

Specifically, the (rather obfuscated IMHO) changes to `sortWorklogItems` and `sortWorklogItemsReverse`, used by the Worklog view to display data.

For how these comparators are actually being used in the template, the `key` (of `KeyValuePair<K, V>`) should always be a `string`. As such, that refactored attempt would be using the lexicographic comparison, not the numeric one, which produces the unintuitive ordering seen.

Also worth pointing out that while the prior comparator implementations do work, they do so in a perhaps unexpected manner. That is, the subtraction operator on incoming `string` values (incorrectly typed as `number`) causes an implicit conversion of the operands to `number` values.

https://github.com/johannesjo/super-productivity/blob/5e43c022d62e0e203d12e753f5202a73c8b2c49f/src/app/features/worklog/worklog.component.ts#L152-L158

I've also added tests for the worklog component to cover this ordering behaviour to hopefully prevent future refactor regressions.

## Issues Resolved

- #5465

## Check List

### DIFF
--- a/src/app/features/worklog/worklog.component.spec.ts
+++ b/src/app/features/worklog/worklog.component.spec.ts
@@ -1,0 +1,149 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { ActivatedRoute } from '@angular/router';
+import { provideNoopAnimations } from '@angular/platform-browser/animations';
+import { By } from '@angular/platform-browser';
+
+import { TranslateModule } from '@ngx-translate/core';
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much!

Issue #5252, opened 2025-10-11
🚨Superproductivity will not launch


SDohne:
### Version Used

latest/stable 15.2.0 (installed via snap) 

### Expected Behavior

I try to launch the application and it opens

### Current Behavior

when i try to launch, it doesn't open. In terminal I get this error

<img width="1740" height="927" alt="Image" src="https://github.com/user-attachments/assets/1ab1fd66-ffb6-4c08-89da-215c50b2d11f" />

[superproducivityerror.pdf](https://github.com/user-attachments/files/22863702/superproducivityerror.pdf)

### Steps to Reproduce

1. own a Lenovo thinkpad L14 Gen 1: 20U6S31T00 WITH Ubuntu: 22.04.4
2.  install superproducivtiy via snap
3. try to launch
4. error appears

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

gvgvgvijayan:
Facing the same issue.

# Environment:
Ubuntu 25.04
Snap App

# Error:

```
superproductivity

MESA-LOADER: failed to open radeonsi (search paths /snap/superproductivity/2653/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: radeonsi
MESA-LOADER: failed to open kms_swrast (search paths /snap/superproductivity/2653/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: kms_swrast
MESA-LOADER: failed to open swrast (search paths /snap/superproductivity/2653/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load swrast driver
Start app...
dbus-send: /snap/superproductivity/2653/lib/x86_64-linux-gnu/libdbus-1.so.3: version `LIBDBUS_PRIVATE_1.12.20' not found (required by dbus-send)
17:36:48.006                   › Initializing plugin node executor
17:36:48.014                   › { enabled: null, showDevTools: false, mode: 'bottom' }
MESA-LOADER: failed to open radeonsi (search paths /snap/superproductivity/2653/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: radeonsi
MESA-LOADER: failed to open kms_swrast (search paths /snap/superproductivity/2653/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: kms_swrast
[...]

Issue #5196, opened 2025-09-30
🚨 Cannot run via Wayland


wxmup:
### Version Used

15.0.3 aur

### Expected Behavior

When I use the flag to turn on the Wayland display, Super Productivity will run in Wayland.

### Current Behavior

When I use the flag to turn on the Wayland display, super productivity will not run.

<img width="1848" height="1504" alt="Image" src="https://github.com/user-attachments/assets/840d94a3-2a89-4ea0-9f33-7808d0b7b291" />

### Steps to Reproduce

1. use flag:
```
--ozone-platform=wayland
```
2. turn on super productivity
3. An empty window pops up

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell
error logs:

[11731:0930/155525.557879:ERROR:gpu/ipc/client/command_buffer_proxy_impl.cc:127] ContextResult::kTransientFailure: Failed to send GpuControl.CreateCommandBuffer.
[11542:0930/155525.558653:ERROR:content/browser/gpu/gpu_process_host.cc:964] GPU process exited unexpectedly: exit_code=139
[11731:0930/155525.979760:ERROR:gpu/ipc/client/command_buffer_proxy_impl.cc:127] ContextResult::kTransientFailure: Failed to send GpuControl.CreateCommandBuffer.
[11542:0930/155525.979937:ERROR:content/browser/gpu/gpu_process_host.cc:964] GPU process exited unexpectedly: exit_code=139
[11731:0930/155526.378024:ERROR:gpu/ipc/client/command_buffer_proxy_impl.cc:127] ContextResult::kTransientFailure: Failed to send GpuControl.CreateCommandBuffer.
[11542:0930/155526.378274:ERROR:content/browser/gpu/gpu_process_host.cc:964] GPU process exited unexpectedly: exit_code=139
[11919:0930/155526.444384:ERROR:ui/base/x/x11_software_bitmap_presenter.cc:147] XGetWindowAttributes failed for window 1
[11731:0930/155526.457976:ERROR:cc/raster/one_copy_raster_buffer_provider.cc:298] Creation of StagingBuffer's SharedImage failed.
[11731:0930/155526.457976:ERROR:cc/raster/one_copy_raster_buffer_provider.cc:298] Creation of StagingBuffer's SharedImage failed.
[...]

PR #5522, opened 2025-11-15
feat: bump to electron 38 to fix Wayland GPU launch issue

Lap1n:
# Description

  - Bump the project from Electron 37.7.0 to 38.x (package + lockfile) so we pick up the upstream fix for the Wayland GPU sandbox regression (electron/electron#45862)

  ## Issues Resolved

  - Fixes the “GPU process exited unexpectedly / GpuControl.CreateCommandBuffer” crash tracked in #5252 (and duplicates like #5196) by moving to Electron 38, which incorporates the upstream Electron fix.

  ## Check List

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,7 @@
         "@angular-eslint/template-parser": "^20.5.0",
         "@angular/animations": "^20.3.7",
         "@angular/cdk": "^20.2.10",
-        "@angular/cli": "^20.3.10",
+        "@angular/cli": "^20.3.7",
         "@angular/common": "^20.3.7",
         "@angular/compiler": "^20.3.7",
         "@angular/compiler-cli": "^20.3.7",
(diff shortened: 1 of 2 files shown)


johannesjo:
We upgraded to electron 38 and it lead to quite a lot of problems both for snap release and for flatpak release. That's why I reverted back to v37. Are you reasonably confident this won't be the case anymore?

Lap1n:
I am not very familiar with the snap and flatpack release process as I am using nixpkgs instead of snap to install my linux packages, but I am confident that latest electron fixes the launching issues on Wayland.

What's the process to test if there is an issue with Snap & flatpacks?

johannesjo:
@codex

xppcom:
Is there still anything preventing bumping up the electron version ? 

Super Productivity's flatpak app can't open on native wayland currently and has to be ran through the XWayland compatibility layer, making any other app on the system able to see its content

johannesjo:
I will make a new release later today with electron 39. Hopefully it won't break the app for snap and flatpak. Help testing it would be very welcome!

PR #5539, opened 2025-11-17
fix(localization): global refactor

novikov1337danil:
# Description

Refactoring of date and time localization

## Issues Resolved

The changes likely affect all issues

https://github.com/johannesjo/super-productivity/issues/3274
https://github.com/johannesjo/super-productivity/issues/3462
https://github.com/johannesjo/super-productivity/issues/5246
https://github.com/johannesjo/super-productivity/issues/5481
https://github.com/johannesjo/super-productivity/issues/5504
https://github.com/johannesjo/super-productivity/issues/5525

But I'm not sure all the issues have been completely fixed. I'll analyze this again later

### DIFF
--- a/src/app/features/tasks/add-task-bar/add-task-bar-actions/add-task-bar-actions.component.spec.ts
+++ b/src/app/features/tasks/add-task-bar/add-task-bar-actions/add-task-bar-actions.component.spec.ts
@@ -61,7 +61,7 @@ describe('AddTaskBarActionsComponent', () => {
   const mockStore = jasmine.createSpyObj('Store', ['select', 'dispatch']);
   const mockConfigService = (locale: Locale): GlobalConfigService => {
     return jasmine.createSpyObj('GlobalConfigService', [], {
-      lang: () => ({ timeLocale: locale }),
+      localization: () => ({ timeLocale: locale }),
     });
   };
   const mockDateTimeFormatService = jasmine.createSpyObj('DateTimeFormatService', ['-']);
--- a/src/app/features/tasks/add-task-bar/add-task-bar-mentions.spec.ts
+++ b/src/app/features/tasks/add-task-bar/add-task-bar-mentions.spec.ts
@@ -94,7 +94,7 @@ describe('AddTaskBarComponent Mentions Integration', () => {
         isEnableDue: true,
         isEnableProject: true,
       }),
-      lang: () => ({ timeLocale: DEFAULT_LOCALE }),
+      localization: () => ({ timeLocale: DEFAULT_LOCALE }),
       misc$: miscSubject,
     });
     const addTaskBarIssueSearchServiceSpy = jasmine.createSpyObj(
--- a/src/app/features/tasks/add-task-bar/add-task-bar.component.spec.ts
+++ b/src/app/features/tasks/add-task-bar/add-task-bar.component.spec.ts
@@ -181,7 +181,7 @@ describe('AddTaskBarComponent', () => {
       lang$: new BehaviorSubject<LocalizationConfig>(mockLocalizationConfig),
       misc$: new BehaviorSubject<MiscConfig>(mockMiscConfig),
       shortSyntax$: of({}),
-      lang: () => ({ timeLocale: DEFAULT_LOCALE }),
+      localization: () => ({ timeLocale: DEFAULT_LOCALE }),
     });
     mockStore = jasmine.createSpyObj('Store', ['select', 'dispatch']);
     mockMatDialog = jasmine.createSpyObj('MatDialog', ['open']);


novikov1337danil:
Oh, damn... This is still a draft PR because I have a few TODOs left, and I would really appreciate it if someone could take a look and test my changes, if possible.

Also, @johannesjo, I need some help:
1. We probably need to create a migration to move the `timeLocale` user saved settings from `misc` to the `lang` section. It is now located there.
2. I want to suggest renaming `timeLocale` to `dateTimeLocale`, if that's possible (a migration will likely be needed for that too).
3. After my changes, I managed to fix many tests, except for these three: `date-time-format.service.spec.ts`, `format-day-month-str.spec.ts`, and `map-archive-to-worklog-weeks.tz.spec.ts`. All these tests are ending with errors. It's happening because now we are using `LocaleDatePipe` everywhere, which uses `inject(DateTimeFormatService)` under the hood. I've tried different approaches to make it work, but did not achieve success.

I'd really like to move this PR out of draft status, as it affects a lot of files and I'm worried about further merge conflicts

johannesjo:
Hey @novikov1337danil ! Thank you very much for this. I'll have a manual look tomorrow or on friday. 

For now, here is some AI feedback from claude:


```
 PR Review: #5539 - Global Localization 
  Refactoring

  Summary

  This is a comprehensive localization refactoring
   that addresses multiple date/time formatting
  issues. All tests pass (2474 SUCCESS) and
  linting is clean. The author's concerns about
  failing tests are no longer valid - they've been
   fixed.

  ✅ Strengths

  1. Excellent Architecture Improvements

  - Proper separation of concerns: LanguageCode
  (ISO-639) vs Locale (ISO-639 + ISO-3166)
  - Type safety: Strong typing with Locale type
  and LocalesImports record
  - Centralized configuration: All locale data in
  one place with proper imports
  - Better defaults: Uses constants like
  DEFAULT_LANGUAGE, DEFAULT_LOCALE,
  DEFAULT_FIRST_DAY_OF_WEEK

  2. Cleaner Language Service

  The refactored LanguageService is much cleaner:
  - detect(): Browser language detection
  - isValid(): Type guard for validation
  - tryAutoswitch(): Explicit autoswitch logic
  with return value
  - Removed coupling to DateAdapter (moved to
  DateTimeFormatService)

  3. Improved DateTimeFormatService

  - Added _testFormats computed signal for better
  debugging
  - Better locale handling with fallback to
  DEFAULT_LOCALE
  - Moved firstDayOfWeek monkey-patching here
  (better location than LanguageService)
  - Proper reactive updates using effect()

  4. Configuration Reorganization

  - firstDayOfWeek moved from misc → lang (more
  logical grouping)
  - Uses undefined for "not set yet" vs null for
  "explicitly reset to default"
  - Cleaner initialization logic in effects

  5. Removed Global Locale Imports

  - Old code: import 
  '@angular/common/locales/global/en' (bloat)
  - New code: import localeEn from 
  '@angular/common/locales/en' (tree-shakeable)

  ⚠️ Issues & Recommendations

  🔴 Critical: Missing Migration

  Problem: The PR moves firstDayOfWeek from misc
  to lang config without a migration.

  Impact: Users who configured firstDayOfWeek will
   lose their setting.

  Solution Needed:
  // Add to migration logic
  if (oldConfig.misc?.firstDayOfWeek !==
  undefined) {
    newConfig.lang.firstDayOfWeek =
  oldConfig.misc.firstDayOfWeek;
    delete newConfig.misc.firstDayOfWeek;
  }

  src/app/core/language/language.service.ts:31-36

  🟡 Medium Priority Issues

  1. Inconsistent undefined vs null Semantics

  The PR uses undefined for "not set" and null for
   "explicitly reset", but this isn't consistently
   enforced:

  src/app/features/config/store/global-config.effe
  cts.ts:76-86

  The filter sectionCfg['lng'] !== undefined will
  skip both undefined AND null, which may not be
  the intended behavior.

  Recommendation: Document this clearly or use a
  more explicit check:
  filter(({ sectionCfg }) => 'lng' in sectionCfg
  && sectionCfg['lng'] !== undefined)

  2. Type Safety Regression in 
  global-config.model.ts

  The current master has:
  lng: LanguageCode | null;
  timeLocale?: string;

  PR version:
  lng?: LanguageCode | null;
  timeLocale?: Locale | null;
  firstDayOfWeek?: number | null;

  The change to lng?: ... (optional) changes the
  type signature. This could break code that
  expects lng to always be present.

  Recommendation: Keep lng required but allow
  undefined:
  lng: LanguageCode | null | undefined;

  3. LanguageCode Enum Change: zh_tw → zhTw

  src/app/app.constants.ts:89-141

  Impact: Breaking change if users have zh_tw
  stored in their config.

  Recommendation: Add migration to convert zh_tw →
   zhTw or keep both values.

  4. Missing registerLocaleData Calls

  The PR removes global locale imports but I don't
   see where registerLocaleData is called with the
   imported locales. Angular needs this for the
  DatePipe to work properly.

  Check needed: Verify that locales are actually
  registered somewhere in the initialization flow.

  5. LOCALE_ID Provider Removed

  The PR removes:
  provide: LOCALE_ID,
  useValue: navigator.language or 'en-US'

  This was Angular's way of setting the default
  locale for built-in pipes. Without this, Angular
   will use the browser's default, which might not
   match the user's selected timeLocale.

  Recommendation: Provide LOCALE_ID dynamically
  based on config:
  {
    provide: LOCALE_ID,
    deps: [GlobalConfigService],
    useFactory: (config: GlobalConfigService) =>
  config.lang()?.timeLocale ?? DEFAULT_LOCALE
  }

  🟢 Minor Issues

  1. TODOs in Production Code

  // TODO sort by popular
  // TODO add translation  
  // TODO move to separate file
  // TODO retest

  Recommendation: Create follow-up issues for
  these.

  2. Console Logging in Production

  src/app/core/date-time-format/date-time-format.s
  ervice.ts:30-45

  Recommendation: Use Log.log() or remove for
  production, or wrap in !environment.production.

  3. Type Safety: any Removed

  Good cleanup removing snackUpdate$: any →
  snackUpdate$ ✅

  📊 Test Results

  ✅ All 2474 tests passing
  ✅ Linting clean
  ✅ Previously mentioned failing tests now pass:
     - date-time-format.service.spec.ts ✅
     - format-day-month-str.spec.ts ✅
     - map-archive-to-worklog-weeks.tz.spec.ts ✅

  🎯 Recommendations

  Before Merging:

  1. Add migration for firstDayOfWeek move from
  misc → lang
  2. Add migration for zh_tw → zhTw rename
  3. Verify locale registration - ensure
  registerLocaleData() is called
  4. Re-add LOCALE_ID provider with dynamic value
  from config
  5. Document the undefined vs null semantics for
  optional config values

  Nice to Have:

  6. Remove/reduce console logging in production
  7. Create follow-up issues for TODOs
  8. Consider renaming timeLocale → dateTimeLocale
   (as author suggested)
  9. Consider renaming lang config section →
  localization

  📝 Final Verdict

  Status: ⚠️ Approve with required changes

  This is high-quality refactoring that improves
  code organization and type safety. The
  architecture is sound and tests pass. However,
  migration logic is essential to prevent user
  data loss when firstDayOfWeek moves from misc to
   lang.

  The author should be commended for tackling this
   complex cross-cutting concern across 70 files
  while maintaining test coverage.


```

novikov1337danil:
> I'll have a manual look tomorrow or on friday.

Thank you!

> For now, here is some AI feedback from claude:

"**Critical: Missing Migration**" - Yes, this is a problem that I didn't know how to solve correctly.

**Medium Priority Issues**
**1-3 points:** I will analyze these problems over the weekend.
**4. "Missing registerLocaleData Calls"** - It is located in `main.ts:188`
**5. "LOCALE_ID Provider Removed"** - Yes, I actually removed LOCAL_ID. However, it's no longer required because all built-in angular `date` pipes have been replaced with `localeDate` pipe

**Minor Issues**
**1. "TODOs in Production Code"** - Yes, I will delete some of them.
**2. "Console Logging in Production"** - I'll leave it as is since this logging was already in the code.
**3. "Type Safety: any Removed"** - It`s not an issue

**"✅ All 2474 tests passing"** - Unfortunately, this is not true.

johannesjo:
I had a look at the PR and it looks good to me so far! Thank you very much for your effort!

> 1. We probably need to create a migration to move the `timeLocale` user saved settings from `misc` to the `lang` section. It is now located there.

You can find the migration stuff and examples under `src/app/pfapi/migrate`. I usally add a minor version migration  (e.g. 4.1) when the migrated data would also work with previous versions of the app. It's fine if it runs without errors on both the newer and the older version of the app (e.g. if the older app keeps updating and working with a legacy field while the newer version writes to a different property). Otherwise there will be a new major migration (e.g. 5.) needed indicating that the older clients won't work with data.  I can also help you with the migrations of course.

> 2. I want to suggest renaming `timeLocale` to `dateTimeLocale`, if that's possible (a migration will likely be needed for that too).

Yes, makes sense to me. 

> 3. After my changes, I managed to fix many tests, except for these three: `date-time-format.service.spec.ts`, `format-day-month-str.spec.ts`, and `map-archive-to-worklog-weeks.tz.spec.ts`. All these tests are ending with errors. It's happening because now we are using `LocaleDatePipe` everywhere, which uses `inject(DateTimeFormatService)` under the hood. I've tried different approaches to make it work, but did not achieve success.
> 

I can have a look, once you think this is otherwise good to go.

novikov1337danil:
@johannesjo Could you please look at the last two commits?

1. I renamed everything I wanted.
2. Create migration, but I didn't create the js/js.map files (I guess I should create them using some `tsс` command). I don't know if everything is correct

novikov1337danil:
OK, I've checked 1-2 points, and I think everything is fine there.

For point 3, I returned the file name `zh-tw` again, as it was done in this PR - https://github.com/johannesjo/super-productivity/pull/5372 (for some reason, it was renamed back by you 2 weeks ago, so maybe I'm wrong). Anyway, I added this renaming to the migration as well.

Otherwise, I believe everything is ready. Now, it's necessary for you to take a look at the migration and failed tests.

johannesjo:
Thank you very much @novikov1337danil  for all your hard work! <3

johannesjo:
Hey @novikov1337danil !

In 6763c965 I had to simplify the format-day-month-str.spec.ts to work without the angular context to make unit and e2e tests work again. It probably doesn't work exactly like your solution. Not sure what is best here. To make your approach work properly one approach would be  to create a pipe and inject that. Not sure if this is neccessary, which depends on if we need to have any angular service or pipe available or if we can extract everything to pure functions. Would you mind having a look?

novikov1337danil:
> In [6763c96](https://github.com/johannesjo/super-productivity/commit/6763c965f78cb05447671aea6e793e75320b948c) I had to simplify the format-day-month-str.spec.ts to work without the angular context to make unit and e2e tests work again. It probably doesn't work exactly like your solution. Not sure what is best here. To make your approach work properly one approach would be to create a pipe and inject that. Not sure if this is neccessary, which depends on if we need to have any angular service or pipe available or if we can extract everything to pure functions. Would you mind having a look?

I like your approach.

I think we should unify the code a bit: if you look at the `formatMonthDay` function, you'll see that it uses the same formatting code as `formatDayMonthStr`.

So we could create a function:
```
export const dateTimeFormatter = (
  locale: Locale,
  overrides?: Intl.DateTimeFormatOptions,
): Intl.DateTimeFormat => {
  return new Intl.DateTimeFormat([locale, DEFAULT_LOCALE], {
    month: 'numeric',
    day: 'numeric',
    ...overrides,
  });
};
```

And use it wherever we need to format dates and times

```
// === This is formatDayMonthStr function ====
// ...
const monthDay = dateTimeFormatter(locale)
  .format(date)
  // Remove zero-padding for consistency across locales
  .replace(/\b0+(\d)/g, '$1');
// ...
```

```
// === This is formatMonthDay function ===
// ...
const formatted = dateTimeFormatter(locale).format(date);
// ...
```

**It would be nice to use it everywhere as the core for date and time formatting** (to avoid duplication or, even worse, different approaches to formatting in large codebase).

If we go further, perhaps we could improve it to remove the current `DatePipe` (imported from `@angular/common`) in our new `LocaleDatePipe`. We'd simply use our `dateTimeFormatter` function too

PR #4097, opened 2025-03-09
Add sv translation

joeax910:
Added/started Swedish (sv) translation.

(Accidentally first added "se" instead of "sv" language and later removed it, plus had to wrestle a bit with inlang/Fink due to inexperience, which helps explain the messy commit history... My apologies!)

PR #5543, opened 2025-11-18
add swedish translation file

ExoGitH:
# Description

As user joeax910 began translation (#4097). I have pick up the file and add the rest with help of translator. Will be not perfect translation but better then nothing (i hope).
Maybe a motivation by some native speaker to tweak it.

## Issues Resolved

pull request #4097 can be closed.


johannesjo:
Thank you very much!

PR #5553, opened 2025-11-20
feat: Ability to configure default start page

### LINKED ISSUE #5537: 💡 Default start page (seperate for desktop versus app)
### Problem Statement

I was thinking through my issue with quick add on Android and the main problem was that it tried to load the today page no matter what whenever I opened the app - this causes me to have to tap the hamburger, tap inbox, tap add task and then finally add it. 

### Possible Solution

If I could set the default on android to open inbox, that would save a ton of taps. On desktop it still opens to today, but on android it opens to inbox would make my workflow way better. 

Have this be a setting means other workflows wont be impacted. This also seems to be way lower scope than trying to add a new new quick add feature. 

_No response_

### Additional context

_No response_

nskins:
# Description

Provides the user with the ability to configure the start page to either "Today" or "Inbox" using a setting found in misc. settings.

## Issues Resolved

#5537

## Check List


src/app/app.routes.ts
```diff
@@ -228,6 +234,5 @@ export const APP_ROUTES: Routes = [
       ),
     data: { page: 'contrast-test' },
   },
-
-  { path: '**', redirectTo: `tag/${TODAY_TAG.id}/tasks` },
+  { path: '**', redirectTo: defaultStartPageRoute },
```
johannesjo: this is propbably not necessary. It is enought o change the initial URL.

nskins:
@johannesjo Thank you for the suggestion! I gave that a try, and I noticed that the browser would briefly show the content for "Today" before quickly switching over to "Inbox." I was able to resolve that issue by creating a component that handles the redirection at the router level. Would that solution be acceptable for you?

My only issue right now is that when the "Today" page loads initially, it appears that the translation for `F.TASK.ADD_TASK_BAR.TODAY` does not get injected correctly. Do you have any idea why this could be happening? See picture below.

image

johannesjo:
I hope you don't mind (some ai assisted) suggestions:

Maybe a better (faster) approach for this would be using a guard like this?

```TypeScript

@Injectable({ providedIn: 'root' })
export class DefaultStartPageGuard {
  private _globalConfigService = inject(GlobalConfigService);
  private _router = inject(Router);

  canActivate(
    next: ActivatedRouteSnapshot,
    state: RouterStateSnapshot,
  ): Observable<UrlTree> {
    return this._globalConfigService.misc$.pipe(
      take(1),
      map((miscCfg) => {
        const TODAY = 0;
        const INBOX = 1;
        if ((miscCfg?.defaultStartPage ?? TODAY) === INBOX) {
          return this._router.parseUrl(`/project/${INBOX_PROJECT.id}/tasks`);
        } else {
          return this._router.parseUrl(`/tag/${TODAY_TAG.id}/tasks`);
        }
      }),
    );
  }
}


```

------


In case we keep the component approach:



  - src/app/pages/wildcard/
    wildcard.component.ts:1 – we should add the standalone flag
  - src/app/pages/wildcard/
    wildcard.component.ts:8 – Using a lazy
    loadComponent for the wildcard route means
    every initial load (the app boots at **)
    now waits for an extra chunk download/parse
    before redirecting to the real start page.
    Previously this was an immediate redirect.
    Consider making the wildcard handler eager
    (plain component) or handling the redirect
    in a guard to avoid the extra network hop
    on startup.
  - src/app/pages/wildcard/
    wildcard.component.ts:16 – The navigation
    does not use replaceUrl, so hitting “Back”
    after launch lands on the wildcard URL and
    triggers another redirect. Minor UX/extra
    navigation work each time.

nskins:
That's fine. I'm happy to switch over to using a guard if you think that's a better approach. I pushed some updated code. For some reason, the misc. config settings aren't yet loaded correctly within the body of `canActivate` (when debugging, the value of `defaultStartPage` is always set to 0), and I'm not sure why that is. I'm looking into a fix.

johannesjo:
Thank you very much for the adjustments <3

PR #5576, opened 2025-11-25
Feat/configurable features

### LINKED ISSUE #5303: 💡 Hide default "planner" and "schedule" from sidebar
### Problem Statement

I don't use the planner or schedule, so would be nice to hide those so I can focus on what I do use. Adding a specific board instead would be nice. 

### Possible Solution

Might be easiest to add an option to hide any of those (including boards, inbox etc). I don't see why people would hide those, but wouldn't hurt to have the option

The ability to add any board as an option would be nice. Eg. I can put my eisenhower matrics or custom board there for quick access without having to click on "boards" 

_No response_

### Additional context

_No response_

### LINKED ISSUE #5454: 💡 Option to disable all time tracking features in the UI
### Problem Statement

I have no need for any of the time tracking features, I just use it as an advanced kind of task list. 

### Possible Solution

Just wondering if you would consider adding an option that would then hide all of the time tracking related UI parts, from the Play button to the "ready to work" button and so on?

Otherwise the user can of course just ignore it

### Additional context

_No response_

kiausch:
# Description

Adds a new section "App Features" to the settings page with options to disable several ui elements:

grafik

The default value for all features is enabled.

The settings effectively only hide the buttons in the ui and disable the keyboard shortcuts, therefore the term "App Features" might technically be a bit imprecise, but I think from a user's perspective this should be fine.

If feature is active while being disabled:
- time tracking: current tracking is stopped
- focus session: has to be exited to get to the settings
- Panels: stay visible until manually closed (should be no problem)

Notes:
- I don't know if it makes sense to have focus mode enabled while having time tracking disabled; Maybe makes sense for pomodoro timer in focus session only? Anyway, I was not able to generate unrecoverable or broken application states in this configuration.
- As I am not actively using the issues panel, I do not have any idea of what might happen if connections are set up before the feature is disabled. Are there background sync activities happening which should be disabled additionally if the feature is disabled?

## Issues Resolved
closes #5454
closes #5303

## Check List


johannesjo:
Thank you very much @kiausch ! This is very well done! Thank you especially for the unit tests and the e2e tests. I will merge this. We can always improve later, if something comes up.


> * I don't know if it makes sense to have focus mode enabled while having time tracking disabled; Maybe makes sense for pomodoro timer in focus session only? Anyway, I was not able to generate unrecoverable or broken application states in this configuration.

I think they are not necessarily bound to each other. I imagine some uers want to use the focus sessions (for focus), but none of the time tracking.

> * As I am not actively using the issues panel, I do not have any idea of what might happen if connections are set up before the feature is disabled. Are there background sync activities happening which should be disabled additionally if the feature is disabled?

1. They are two things the issue integrations do in the background: Poll existing issues if configured to do so, to update them.
2. Autopoll for new issues to add, if configured to do so.

What we could do, to imrpve on this would be either just disabling all the integrations, or adjusting the isEnabled checks to include a check for if the feature is enabled. Thinking about it the latter probably makes the most sense, thoguh the former might possibly be easier to implement.

johannesjo:
After working with the code base a bit more, I'm even more impressed. Really well and cleanly done @kiausch ! 

People also seem to be looking forward to this already:

https://www.reddit.com/r/superProductivity/comments/1p77qtu/comment/nr1w1mw/

kiausch:
Thank you so much, I'm glad I could help! I just tried to keep it simple and easy to extend. Looks like I nailed it :-), already saw your changes and additions, very nice! Also nice that people appreciate it.

The onboarding dialog looks good, too!

johannesjo:
@ kiausch And another post about this :)

https://www.reddit.com/r/superProductivity/comments/1pb2p53/1643_this_update_is_awesomeeeee/

PR #5605, opened 2025-11-29
fix #5596

### LINKED ISSUE #5596: 🚨 Profile system is not  initialized
### Version Used

16.4.1

### Expected Behavior

when Profile system is enable, it is avaliable

### Current Behavior

Image

### Steps to Reproduce

enable the Profile system, and try to config it

### Console Output

```shell

```

overcuriousity:
# Description

Fixed broken user profiles activation.

It seems there was a race condition and a unnecessary check, as well as a missing import.
Instead I added a signal to the UserProfileService if the subsystem was initialized successfully.

The user profiles could be activated and used reliable in my test scenario.

## Issues Resolved

Fix #5596


johannesjo:
Thank you very much @overcuriousity !

Issue #4490, opened 2025-05-30
💡 Enhance PBKDF2 hashing


nicojmn:
### Problem Statement

Hi, currently the number of iterations with [PBKDF2-HMAC-SHA256](https://w3c.github.io/webcrypto/#pbkdf2-description) is set to 1000 in the code below : 

https://github.com/johannesjo/super-productivity/blob/9b0cfdab7b32e141cc8f7a42119c05fe504adcf8/src/app/pfapi/api/encryption/encryption.ts#L23-L31

According to [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pbkdf2), PBKDF2 with SHA256 should have 600 000 iterations to be secure.

Note : I'm not a security expert so maybe in your context is it enough

### Possible Solution

I suggest increasing iterations to the recommend 600 000 or to use SHA512 with 210 000 iterations

nicojmn:
Concerning SHA-512, it might be overkill for the needs of the app

johannesjo:
Input on this by someone more experienced with the subject would be very welcome!

firewavemtl:
Hi,

Indeed, PBKDF2 with SHA256 needs 600k iterations in 2025. But the real security issue there is the salt. It's not random!

Salt must be random, and do not derive from the password (even if unknown). Here are the main consequences I can think of:

- Two users with the same password will have the same hash
- An attacker can immediately identify identical passwords
- Facilitates statistical attacks on popular passwords
- An attacker can precompute a rainbow table
- An attacker can attack multiple users simultaneously

Certainly in this case, it would be rare to end up with multiple hashes in the same system, but it could happen. For example, a company using Super Productivity with sync on the same server.

If you have to deal with changing salt and iteration counts, I suggest switching directly to Argon2id for key derivation. This is the safe choice in 2025.

johannesjo:
Thank you very much for the input @firewavemtl !

johannesjo:
PRs for this would be very welcome, since I am not an expert on the matter.

PR #5608, opened 2025-11-29
feat(encryption): migrate to argon2id

nicojmn:
# Description

This PR aims to resolve problems pinpointed in #4490.
It adds argon2id encryption and decryption, while keeping PBKDF2 for backward compatibility with older encrypted versions.

For now the code try to decrypt with argon2id and if unsuccessful try to decrypt with PBKDF2, let met know if that's good for you or if you want a more sophisticated migration mechanism

## Issues Resolved

Issue #4490 

## Check List

- [x] New functionality includes testing (tests already made).
- [x] New functionality has been documented in the README if applicable. (not applicable)

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,6 +130,7 @@
     "electron-log": "^5.4.1",
     "electron-window-state": "^5.0.3",
     "fs-extra": "^11.3.2",
+    "hash-wasm": "^4.12.0",
     "node-fetch": "^2.7.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


nicojmn:
Hi @johannesjo , just an update about the checklist. Tests made with legacy PBKD2 cover encryption and decryption with argon2id

johannesjo:
Thank you very much  @nicojmn ! 

Will this work with data encrypted by the old algo?

nicojmn:
> Will this work with data encrypted by the old algo?

Short answer is yes !

Longer explanation is : 

1. **Encryption :** always use argon2id 
2. **Decryption** : try argon2id first
    -  If succeed, data is decrypted, we're done
    -  If failed, it means that data was encrypted with old algo, so we decrypt with PBKDF2, then we're done
    
 
Below is a simple diagram explaining the behavior of decrypt function
pr-diagram

johannesjo:
Thank you very much!

PR #5640, opened 2025-12-01
feat: Add option to duplicate project

qievenz:
# Description

Hi, great app! This Pull Request introduces a new feature that allows users to duplicate an existing project.

This functionality is particularly useful for quickly generating new projects that share a similar structure or set of initial tasks. My specific use case is the creation of template projects for courses being studied, where each new course requires an identical set of initial tasks (e.g., "read chapter 3," "make anki cards," "review lecture notes").

## Issues Resolved

https://github.com/johannesjo/super-productivity/discussions/781
https://github.com/johannesjo/super-productivity/issues/606

## Check List

### DIFF
--- a/src/app/features/project/project.service.ts
+++ b/src/app/features/project/project.service.ts
@@ -44,6 +44,9 @@ import { toSignal } from '@angular/core/rxjs-interop';
 import { TranslateService } from '@ngx-translate/core';
 import { T } from 'src/app/t.const';
 import { sortByTitle } from '../../util/sort-by-title';
+import { Note } from '../note/note.model';
+import { selectNoteFeatureState } from '../note/store/note.reducer';
+import { addNote } from '../note/store/note.actions';
 import { DialogConfirmComponent } from '../../ui/dialog-confirm/dialog-confirm.component';
 
 @Injectable({
@@ -301,6 +304,13 @@ export class ProjectService {
       noteIds: [],
     });
 
+    const noteState = await firstValueFrom(this._store$.select(selectNoteFeatureState));
+    const notesToCopy = template.noteIds
+      .map((noteId) => noteState.entities[noteId])
+      .filter((note): note is Note => !!note);
+    const newNoteIds = this._duplicateNotesToProject(notesToCopy, newProjectId);
+    this.update(newProjectId, { noteIds: newNoteIds });
+
     this._duplicateTasksToProject(parentTasks, newProjectId, false, taskState);
 
     this._duplicateTasksToProject(backlogTasks, newProjectId, true, taskState);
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much for your PR! The duplicate project feature is a great addition and will be really useful for template workflows. The logic is sound overall, but there are a few changes needed before we can merge.

---

## Required Changes

### 1. Replace deprecated `toPromise()` with `firstValueFrom`

RxJS 7+ has deprecated `toPromise()`. Please update to use `firstValueFrom`:

```typescript
import { firstValueFrom } from 'rxjs';

// Instead of:
const template = await this.getByIdOnce$(templateProjectId).pipe(take(1)).toPromise();

// Use:
const template = await firstValueFrom(this.getByIdOnce$(templateProjectId));

// Same for the task selection call
const allTasks = await firstValueFrom(this._store$.select(selectAllTasksWithSubTasks));
```

### 2. Optimize Task Selection

Currently `selectAllTasksWithSubTasks` retrieves every task in the database, which could be heavy for users with long task histories.

Since you already have `template.taskIds` and `template.backlogTaskIds`, it would be more efficient (and likely more robust for preserving order) to iterate over those arrays and select the specific tasks from the state using `selectTaskFeatureState` or a dictionary lookup, rather than filtering the entire global task list.

### 3. Use Cleaner Object Destructuring

Instead of casting to `any` and manually `delete`-ing properties, you can use object destructuring to separate the properties you want to exclude. This is cleaner and type-safe:

```typescript
// Instead of:
const rest = { ...(p as unknown as Task) } as any;
delete rest.id;
delete rest.parentId;
delete rest.subTaskIds;
delete rest.subTasks;
delete rest.projectId;
delete rest.created;

// Use:
const { id, parentId, subTaskIds, subTasks, projectId, created, ...taskDataToCopy } = p;

const newParentTask = this._taskService.createNewTaskWithDefaults({
  title: p.title,
  additional: taskDataToCopy, // No need for Partial cast if types align
  // ...
});
```

Apply the same pattern to the subtask loop.

### 4. Extract Helper Method

The `duplicateProject` method is getting quite long. Consider extracting the task duplication loop into a private helper method to keep the main method readable:

```typescript
private _duplicateTasksToProject(
  tasks: Task[],
  newProjectId: string,
  isBacklog: boolean
): void {
  // Task duplication logic here
}
```

### 5. Add Unit Tests

The PR checklist indicates tests were added, but I don't see any test files in the changes. Please add unit tests covering:
- Project is created with copied settings
- Tasks are duplicated correctly
- Subtasks are duplicated correctly
- Backlog tasks go to the backlog in the new project
- Error cases (missing project ID, project not found)

---

## Minor Suggestions (Optional)

- Consider adding a confirmation dialog for large projects
- Notes are currently not duplicated (`noteIds: []`) - is this intentional?

---

Looking forward to the updates! Let me know if you have any questions.

qievenz:
@johannesjo  Thanks for the review! I've implemented the required changes and the note duplication. I also added the confirmation dialog for large projects, which triggers when there are over 50 tasks. Is that threshold acceptable? Let me know if that number should be adjusted.

johannesjo:
Thank you!

Issue #905, opened 2021-02-15
Adding a sub-task via drag-drop


plamkas:
Sometimes I create a couple of tasks only to find out that somehow they all have the same origin. I could not find a way to create a new task and to assign the previously created tasks a sub-task to that new one.

JLuisGT:
+1  :D

johannesjo:
Related to #235

JoseOlin:
I think this is still relevant ;)

ced75:
I agree. It is possible to define a subtask as a task, but the opposite is not possible.
It would be nice if we could move a task as a subtask of another existing task.

ced75:
Still relevant :)

Pictor13:
> It would be nice if we could move a task as a subtask of another existing task.

That's not possible as a task might have a sub-task already, so converting the task to sub-task would provoke a hierarchy tree with 2 levels depth, and unfortunately sub-tasks of sub-tasks are not supported by Super Productivity (#643).
So I think that's the blocker: _what should be done with a sub-task when its parent is itself being converted to a sub-task?_

That's a pity as multiple level subtask or proper referencing of tasks is a fundamental feature in my workflow.
Other than that, the app would be perfect.

At least referencing other tasks would make it workable for more complex/granular projects.

jschbrt:
Any update on this? - Is maybe someone working on it currently?

nicola-lunghi:
+1 here

would be useful to drag tasks into another task to convert it to subtask or drag it out to make it a parent task if it gets too big

ItsIgnacioPortal:
> _what should be done with a sub-task when its parent is itself being converted to a sub-task?_

To me, the answer is clear. Transform the orphan subtask into a regular, root-level task.

ItsIgnacioPortal:
Still relevant

ItsIgnacioPortal:
Still relevantes.

dahaupt:
Still looking forward to this.

Nelvin:
+1 to this feature - I just came back to give superproductivity another try and started by creating a bunch of tasks I wanted to group/organize after I've added them which, to my surprise, was not possible.

The most simple way to handle tasks containing subtasks already would be to just not allow those to be transformed into subtasks. Transforming existing subtasks into parent tasks is another option (as menationed by @ItsIgnacioPortal already) but as we can do this manually already, that's much less of an issue - the difference is, requiring a few seconds vs impossible.

[...]

Issue #235, opened 2019-10-09
Convert task to sub task and vice versa


generic-user:
Hi,
As an addition to the issue mentioned here: #181 

Could you implement a way of promoting the sub task to task without having to recreate it? 
You could even go the other way round and:  **add tasks** to another **task** as _sub task_ when the task _does not have any sub tasks_ assigned?

## Thanks
Thanks for all the work you put into this repository. It really is a great application. 🎉

johannesjo:
This completely slipped under my radar. I am very sorry. Thanks for the suggestion. This definitely would be nice.

santiyounger:
awesome to see this was already mentioned, would highly aprecciate this feature. This app is excellent, thanks for your hard work

santiyounger:
I just realized an important aspect to consider (when it comes to time tracking)

Since I use super-productivity mostly for time tracking, I realized it would be a lot better for my workflow to  divide tracked time inside project in sub-tasks to create further division of things.

Since all of my tasks already have data in them, currently the only way I can rearrange my exisiting tasks  (with time tracked data) into sub-task is by manually inputting all of the data from different tracked days.

Just something to consider, keeping the time tracked data when converting a task into sub-task.

It would be extremly helpful!

Thank you so much!

johannesjo:
Extracting sub tasks from their parents is no possible via the context menu (or will be with the next release). The other way round still isn't.

santiyounger:
thank you @johannesjo for working on this, this is one of my most needed features, since I love this app and use it all the time.

I tried to like you said "Extract sub tasks from their parents " via drag and drop after updating to version 6.3.3, but I wasn't sure if you meant v6.3.3 or the next release

Thanks!

santiyounger:
> This issue has not received any updates in 90 days. Please comment, if this still relevant!

Hey, super happy with the way to be able to move sub-task from one parent task to another. However I still think it'd amazing to be able to freely move parent tasks into other tasks subtasks.

## Visual Representation
![image](https://user-images.githubusercontent.com/51902730/119367092-3a4a2580-bc77-11eb-8fb4-84663955ba53.png)

![image](https://user-images.githubusercontent.com/51902730/119367159-4afa9b80-bc77-11eb-8ccc-11d36a36ed03.png)

JLuisGT:
+1 to this feature request. :)

EngelPika32:
Still relevant, I suppose : )

santiyounger:
[...]

PR #5646, opened 2025-12-02
feat: Convert task to subtask and vice versa by drag and drop #905 #235

qievenz:
## Description

This PR introduces the ability to convert tasks into sub-tasks and vice versa by dragging and dropping them. This provides a more intuitive and efficient way to organize tasks.

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/905
https://github.com/johannesjo/super-productivity/issues/235

## Check List

## How to test

1.  Create a few tasks.
2.  Drag one task and drop it onto the bottom right of another. The dragged task should now be a sub-task of the target task.
3.  Drag a sub-task from its parent and drop it in the main task list. The sub-task should now be a main task.

## Screencast

[SP subtask dnd.webm](https://github.com/user-attachments/assets/9d93326f-afc5-4b6f-9571-662cc6b46f65)


johannesjo:
Sorry for the delay! I will review this today!

johannesjo:
@qievenz Thank you very much for your effort to tackle this long standing, important and quite challenging problem. It's definitely a step in the right direction, but feel like we are not quite there yet. We need:

1. A clear indication of what is going to happen *before* the user drops a task, ideally in the form of the placeholder that is already being used for other drop scenarios (though this might be hard to do) or a seperate indicator.
2. A clear indication when a task cannot be made a sub task (maybe a message displayed at the cursor when attempting to drop a repeatable task as sub task??)
3. A smooth feel to the whole interaction

Since this is such a core interaction we need to hold this to a high standard, I think.

Would you be willing to further work on this?

ertugruluyar-github:
I think you should consider this behavior too: https://github.com/johannesjo/super-productivity/issues/5659

johannesjo:
> I think you should consider this behavior too: #5659

This should be already covered by the reducer logic I think.

qievenz:
Thanks for the review and sorry for the delay. Sure, I'll work on that

qievenz:
@johannesjo I added a visual indicator that shows the zone where a task will be converted to a subtask on drag & drop.
The subtask conversion zone is in the bottom-right corner of each task (50% width, 75% height from top).
Please test to see if it meets the standard we're looking for 🙏

johannesjo:
Very sorry for the delay @qievenz ! There was just so much to do with th new release and all the things that popped up. Could maybe have a look at the merge conflicts, so I can properly review this again?

johannesjo:
Or even better: Could you re-open a new PR for this. Then we'd get automatically build previews, which would be helpful!

PR #5744, opened 2025-12-16
Update it.json

FromColleWithLove:
Fix name substitution of notification when creating a task in Italian locale.

Solves https://github.com/johannesjo/super-productivity/issues/5743


johannesjo:
Thank you very much!

PR #5746, opened 2025-12-16
Show parent task name in focus mode

calvin-fisher:
# Description

When working on a subtask in focus mode, the parent task name is displayed above the subtask name, in a smaller and dimmer font.

image

The size of the parent task name will shrink responsively along with the subtask name. I also consolidated the `.task-title` responsive breaks because the smallest two were redundant.

## Issues Resolved

When working on a task, I cannot always remember just based on the subtask name what the task was that I was supposed to be working on. It would be helpful to have the parent task name visible as well.

(I can make an issue if needed)

## Check List


johannesjo:
Makes sense to me! Thank you very much!

Issue #5753, opened 2025-12-18
💡 Color selector UI enhancement


YashodharChavan:
### Problem Statement

Currently the UI of color selector seems bit too old style. So I thought it would be an improvement in overall UI. 
Currently it seems like the color selector does not synchronize with other UI components. 

<img width="432" height="691" alt="Image" src="https://github.com/user-attachments/assets/25dd38f1-c4d9-4788-bcd7-755a9e92c2cd" />

### Possible Solution

The color selector UI should be more better and synchronize with the overall UI

PR #5761, opened 2025-12-19
Improved UI for color selector. Now it helps to match the background UI.

YashodharChavan:
fix for #5753
For the first time the UI looked very dull for color picker. I tried it to my end to make it dark and light based on theme toggle, but could not. But still current version is way more better than previous version. 

image

Yes it may require some formatting but I want opinions of other developers. 

This is my first ever open source contribution... and I am excited for further. @johannesjo please accept this version...

### DIFF
--- a/src/app/features/config/color-input/color-input.component.html
+++ b/src/app/features/config/color-input/color-input.component.html
@@ -7,7 +7,6 @@
     style="color: white; font-weight: bold; caret-color: black"
     [(colorPicker)]="formControl.value"
     (colorPickerChange)="formControl.setValue($event); formControl.markAsTouched()"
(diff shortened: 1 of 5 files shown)


johannesjo:
Thank you! These are a lot of new dependencies (including solid-js???). While we probably should improve on this, this adds too much technical debt.

YashodharChavan:
Hey @johannesjo, thanks for feedback !
I experimented with few approaches (including solid js for dynamic theming), but it did not work for light and dark themes. 
So I dropped them all and settled in ngx-color-picker. 
many files have been touched but the core only depends on ngx-color-picker. 
I learned a lot, my first PR. Thanks for your guidance!

Issue #5589, opened 2025-11-28
🚨 Regression on keymap using keyboard shortcut to add task/subtask (AZERTY keyboard)


privat-eco:
### Version Used

16.3.5

### Expected Behavior

Shift+A / A shortcuts using an AZERTY keyboard should add (sub)tasks

### Current Behavior

Actually we need to hit Shift+Q / Q since a recent update

### Steps to Reproduce

1. go to your inbox
2. hit Shit+A or A ; nothing happens

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
@novikov1337danil sorry for bothering you again, but could you possibly have another look on this?

novikov1337danil:
> [@novikov1337danil](https://github.com/novikov1337danil) sorry for bothering you again, but could you possibly have another look on this?

Yeah, you can assign me to all issues related to keyboard layout problems

I hope I'll figure out a way to solve these problems soon

johannesjo:
@novikov1337danil that is great to hear! Thank you very much <3!

novikov1337danil:
The keyboard layout issue was fixed in this PR - https://github.com/super-productivity/super-productivity/pull/5768

Please check this again in version 17+

@johannesjo I think we should close this issue, as well as all the old ones related to the layout.

---

If anyone encounters the problem again, please open a new issue.

Issue #5392, opened 2025-10-30
🚨 wrong shrtcuts with azerty keyboard


juanbreizh:
### Version Used

16.1.0 exe

### Expected Behavior

- pressing the "Shitf + a" keys add a task
- pressing the "a" key add a sub-task
- pressing the "w" key focuses the first entry

### Current Behavior

Shortcut Key mapping with an azerty keyboard is not correct anymore (it was right before v16)

### Steps to Reproduce

- pressing the "Shitf + q" keys add a task
- pressing the "q" key add a sub-task
- pressing the "z" key focuses the first entry

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
@novikov1337danil  could you maybe have a look on this?

novikov1337danil:
> [@novikov1337danil](https://github.com/novikov1337danil) could you maybe have a look on this?

Yes, I'll try to solve this issue there - https://github.com/johannesjo/super-productivity/issues/5298

locnide:
I have the same issue on linux: since the 16.0 (I think). Currently, I’m using the 16.3.0.

Keyboard shortcuts seem to use the corresponding qwerty keys rather than my layout (ergol). 
Same things in the keyboard shortcuts parameters: if I press `t` it registers a `k`. The `t` on my layout corresponds to the position of the `k` in qwerty.

It’s like it is using raw keyboard input rather than the keyboard layout input.

---

**edit**: oh, it seems to be because of the change made for users using multiple layouts. 

https://github.com/johannesjo/super-productivity/pull/5156/files#diff-ec22a49ea869588704199830ca89555ac319119d2e58befa5fe1e387d5e39a3fL22-R24

I wonder how other app handle handle users with multiple layout, maybe like in this PR: [emit physical key presses (only) when a non-Latin layout is active](https://github.com/emilk/egui/pull/4461).

novikov1337danil:
The keyboard layout issue was fixed in this PR - https://github.com/super-productivity/super-productivity/pull/5768

Please check this again in version 17+

@johannesjo I think we should close this issue, as well as all the old ones related to the layout.

---

If anyone encounters the problem again, please open a new issue.

johannesjo:
Thank you very much @novikov1337danil !

Issue #5298, opened 2025-10-16
🚨 Shortcut tutorial ignores keyboard layout (AZERTY/QWERTY mismatch)


HarryStot:
### Version Used

Windows app, first install via Microsoft Store

### Expected Behavior

The shortcut tutorial should display key prompts that correspond to the user's keyboard layout (e.g., AZERTY or QWERTY). For users with AZERTY keyboards, the tutorial should show the correct keys rather than assuming QWERTY.

### Current Behavior

The shortcut tutorial currently displays keys based on the QWERTY layout only. This causes confusion for users with AZERTY keyboards, as the keys shown do not match their actual shortcuts.

### Steps to Reproduce

1. Install Super Productivity from the Microsoft Store on a Windows machine with an AZERTY keyboard layout.
2. Go through the shortcut tutorial.
3. Observe that the displayed shortcuts do not match the physical keys on the AZERTY keyboard.

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

novikov1337danil:
@johannesjo you can assign me

novikov1337danil:
### Just a small update

I tested how keybinding processing works in VScode and realized the following:
1. **In the native app** everything works great: pressing non-EN letters will be correctly converted to EN according to the keyboard layout. Apparently, this works well because the native version has access to the user's OS, and therefore their selected language and keyboard layout.
2. **In the web app** everything works exactly the same as in the current version of SP. This is because there is no access to the user's selected language and keyboard layout. It seems that there process the `event.code` field (as in SP too) **which causes any pressed key appear as if it's a QWERTY layout.**

Ideally, I'd like to have a universal solution that works on all platforms, rather than having separate implementations for native platforms (Linix/Mac/Windows) and web, which would complicate support. However, at the moment, I don't see such a solution :(

johannesjo:
Thank you @novikov1337danil for investigating ! Not sure how to best proceed with this neither tbh.... :/ Any additional input anyone?

novikov1337danil:
After some painful attempts to implement this in different ways, I found a really elegant method with minimal changes to the code using [this](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardLayoutMap).

Even though this API isn't standard yet and is mainly supported in Chromium-based browsers, I don't think that's a problem for us since SP under Electron also uses Chromium (and I hope that other browsers will also add support for this API someday)

### I think I'll be done with this soon

[...]

PR #5768, opened 2025-12-19
feat: support system keyboard layout

novikov1337danil:
# Description

Added support for user system layouts for correct processing of key combinations.
This pr is a continuation of this [refactoring](https://github.com/johannesjo/super-productivity/issues/4620)

## Issues Resolved

https://github.com/johannesjo/super-productivity/issues/5298
https://github.com/johannesjo/super-productivity/issues/5392
https://github.com/johannesjo/super-productivity/issues/5589

## Check List

### DIFF
--- a/src/app/core-ui/shortcut/shortcut.service.ts
+++ b/src/app/core-ui/shortcut/shortcut.service.ts
@@ -203,7 +203,7 @@ export class ShortcutService {
     }
 
     // Handle task-specific shortcuts
-    if (await this._taskShortcutService.handleTaskShortcuts(ev)) {
+    if (this._taskShortcutService.handleTaskShortcuts(ev)) {
       return;
     }
 
--- a/src/app/features/config/keyboard-input/keyboard-input.component.ts
+++ b/src/app/features/config/keyboard-input/keyboard-input.component.ts
@@ -19,7 +19,7 @@ export class KeyboardInputComponent extends FieldType<FormlyFieldConfig> {
     return this.to.type || 'text';
   }
 
-  async onKeyDown(ev: KeyboardEvent): Promise<void> {
+  onKeyDown(ev: KeyboardEvent): void {
     // ! Use ev.code to support multilanguage layouts
     const keyCode = ev.code;
 
@@ -52,7 +52,7 @@ export class KeyboardInputComponent extends FieldType<FormlyFieldConfig> {
       if (ev.shiftKey) val += 'Shift+';
       if (ev.metaKey) val += 'Meta+';
 
-      const keyName = await prepareKeyCode(keyCode);
+      const keyName = prepareKeyCode(keyCode);
 
       if (keyName === 'Meta') {
         // Fail safe for MacOsX crashing bug
--- a/src/app/features/tasks/task-detail-panel/task-detail-panel.component.ts
+++ b/src/app/features/tasks/task-detail-panel/task-detail-panel.component.ts
@@ -144,7 +144,7 @@ export class TaskDetailPanelComponent implements OnInit, AfterViewInit, OnDestro
   private _task$ = toObservable(this.task);
 
   @HostListener('keydown', ['$event'])
-  async onKeydown(ev: KeyboardEvent): Promise<void> {
+  onKeydown(ev: KeyboardEvent): void {
     // Skip handling inside input elements
     const target = ev.target as HTMLElement;
     if (isInputElement(target)) return;


src/app/util/check-key-combo.ts
```diff
@@ -27,19 +43,28 @@ const MINUS_KEY = '-';
...
         Key: '',
         Digit: '',
       },
     };
 
+  // Try to use user kayboard layout mapping - https://developer.mozilla.org/en-US/docs/Web/API/KeyboardLayoutMap
+  if (!rules.codeMapping[code] && 'keyboard' in navigator) {
+    const keyboard = navigator.keyboard as NavigatorKeyboard;
+    // TODO if the `layout` variable was stored somewhere in the app as a global variable, then we could make this function synchronous
+    const layout = await keyboard.getLayoutMap();
```
novikov1337danil: Sorry, there could be a lot less changes if we had found a way to save this value as a global variable so that we could use it here without async/await

johannesjo:
Thank you very much @novikov1337danil ! Could you please merge master once into this so we can get rid of the testing error?

novikov1337danil:
> Thank you very much @novikov1337danil ! Could you please merge master once into this so we can get rid of the testing error?

Done

johannesjo:
Sorry for the delay @novikov1337danil  !! 


Since `checkKeyCombo`  we should avoid making it async if possible and pre-compute as much as possible.



Here is also some more ai supported feedback on this:

---------------------------------

# PR #5768 Review: Support System Keyboard Layout

**PR**: https://github.com/johannesjo/super-productivity/pull/5768
**Date**: 2025-12-31

## Summary

This PR makes `checkKeyCombo` and `prepareKeyCode` async to leverage the Keyboard Layout Map API (`navigator.keyboard.getLayoutMap()`) for correct key detection across different keyboard layouts (QWERTY, QWERTZ, etc.).

**Files Changed**: 6 (+358/-86)
- `src/app/util/check-key-combo.ts`
- `src/app/util/check-key-combo.spec.ts`
- `src/app/core-ui/shortcut/shortcut.service.ts`
- `src/app/features/config/keyboard-input/keyboard-input.component.ts`
- `src/app/features/tasks/task-detail-panel/task-detail-panel.component.ts`
- `src/app/features/tasks/task-shortcut.service.ts`

**Issues Resolved**: #5298, #5392, #5589

---

## Critical Issues

### 1. Test Implementation Bug - Promises Not Awaited

The `withLayouts` helper doesn't await the async callback, so tests may pass without actually running the assertions:

```typescript
// check-key-combo.spec.ts:128-131
const withLayouts = (expectation: () => void, layouts: KeyboardLayout[]): void => {
  layouts.forEach((layout) => {
    overrideLayout(layout);
    expectation();  // ❌ This returns a Promise that's never awaited
  });
};
```

When used with async callbacks:
```typescript
withLayouts(async () => {
  expect(await checkKeyCombo(ev as any, 'Ctrl+Shift+A')).toBe(true);
}, MOCK_LAYOUTS.all);
```

**Fix needed**: The helper should be async and await the callback, or use `Promise.all()`.

---

## Significant Concerns

### 2. Performance - Layout Map Fetched on Every Keypress

The PR fetches the keyboard layout on every single key event:

```typescript
// check-key-combo.ts:59-64
const layout = await keyboard.getLayoutMap();  // Called on every keypress
const foundKey = layout.get(code);
```

The TODO comment acknowledges this:
> "TODO if the `layout` variable was stored somewhere in the app as a global variable, then we could make this function synchronous"

**Recommendation**: Cache the layout map at application startup or on first use.

### 3. Browser Compatibility

`navigator.keyboard.getLayoutMap()` is only supported in Chromium-based browsers. Firefox and Safari don't support it. This is fine for Electron but may affect web/PWA users.

### 4. Missing Error Handling

If `getLayoutMap()` rejects (e.g., permission denied), the code will throw:

```typescript
const layout = await keyboard.getLayoutMap();  // No try-catch
```

### 5. Async Event Handler Race Conditions

Multiple rapid keypresses could cause race conditions since handlers are now async. Example in `shortcut.service.ts`:

```typescript
async handleKeyDown(ev: KeyboardEvent): Promise<void> {
  // Multiple rapid events could interleave here
  if (await checkKeyCombo(ev, keys.toggleBacklog)) { ... }
  else if (await checkKeyCombo(ev, keys.goToFocusMode)) { ... }
```

---

## Positives

1. **Good test coverage** - Tests cover QWERTY_EN, QWERTY_RU, and QWERTZ_DE layouts
2. **Resolves real user issues** - Addresses #5298, #5392, #5589 for international keyboard users
3. **Clean refactoring** - Consistent async/await pattern throughout
4. **Good documentation** - Clear JSDoc and inline comments

---

## Recommendations

1. **Fix the test helper** to properly await async callbacks
2. **Cache the keyboard layout** to avoid repeated async calls
3. **Add try-catch** around `getLayoutMap()` with fallback to existing behavior
4. **Consider debouncing** or a lock mechanism to prevent race conditions

---

## Confidence: 85%

The approach is sound and solves real problems. The test bug is the most critical issue that should be fixed before merging.

novikov1337danil:
I will continue working on this in a week

novikov1337danil:
@johannesjo done

johannesjo:
Thank you very much! Sorry for this trickling in in this fashion, but can you maybe also ivnestigate this potential bug?


### Code review

Found 1 issue:

1. Race condition: `saveUserKbLayout()` is async but called without `await` in constructor. The keyboard layout may not be cached before shortcuts are used, causing non-QWERTY layouts to fail on startup. (CLAUDE.md says "Side effects in constructors → Prefer `async` pipe or `toSignal`")

https://github.com/super-productivity/super-productivity/blob/d17207e8090f7b06af66b47f441a6041129b04e9/src/app/app.component.ts#L228-L231

novikov1337danil:
> 1. Race condition: `saveUserKbLayout()` is async but called without `await` in constructor

I think this is normal and we don't need to wait for this promise to actually be fulfilled. We just call this function and we don't care about the result of its execution

johannesjo:
Thank you! I am merging this!

PR #5861, opened 2026-01-01
Board context menu improvement

### LINKED ISSUE #4694: 💡 Be able to delete unused "Eisenhhower Matrix" and "Kanban" boards
### Problem Statement

By default those boards are enabled, they are not configured unless needed tags exist.
There is not option to remove them (Or at least I dind't find it)

### Possible Solution

Be able to remove those boards

_No response_

### Additional context

_No response_

### LINKED ISSUE #5823: 💡 Adding Delete Option for Boards
### Problem Statement

We can't delete a board quickly in the app.
When we click a particular board with right click, The current options the right click menu has is only, Duplicate Board. 
Not having that feature affects the user experience and convenience of the app.

### Possible Solution

To Delete a board, There must be a delete button in the right click menu that appears when we right click on a board.
Therefore, I would like to suggest the devs to add this feature as this will help in improving the user experience and convenience of the app.

_No response_

### Additional context

_No response_

kiausch:
# Description

Improved context menu on board tabs:

grafik

- added delete and edit menu entries
- added icons to the menu entries
- added translations
- fixed a bug where duplicate always duplicated the active tab if context menu of an inactive tab was clicked
- uses mat-menu directly instead of using ContextMenu component

## Issues Resolved
Closes #5823 
Closes #4694

### DIFF
--- a/src/app/features/boards/boards.component.ts
+++ b/src/app/features/boards/boards.component.ts
@@ -26,6 +26,7 @@ import { BoardCfg } from './boards.model';
 import { MatDialog } from '@angular/material/dialog';
 import { DialogBoardEditComponent } from './dialog-board-edit/dialog-board-edit.component';
 import { DialogConfirmComponent } from '../../ui/dialog-confirm/dialog-confirm.component';
+import { Log } from 'src/app/core/log';
 
 @Component({
   selector: 'boards',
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much for this!

PR #5914, opened 2026-01-06
Added Scoop installation instructions

Gitoffthelawn:


johannesjo:
Thank you! Can you explain how scoop is working? Where does the binary come frome?

Gitoffthelawn:
You're most welcome!  The binary comes from the releases section of this GitHub repo.

Here is the Scoop manifest that is used for Super Productivity:  
https://github.com/ScoopInstaller/Extras/blob/master/bucket/super-productivity.json

johannesjo:
Thank you for the explanation!

Gitoffthelawn:
You're welcome.  Thank *you* for Super Productivity and your passion in developing it!

PR #5920, opened 2026-01-07
Adds option to import GitLab labels as tags

kemsar:
This pull request adds the option to import GitLab labels as tags in Super Productivity. It adds an option in the the GitLab configuration to enable the creation of tags from labels assigned to GitLab work items (issues and tasks). 

If enabled, as each work item is imported, the work item's labels are duplicated as Super Productivity tags and assigned to the created Super Productivity task. Tags created from GitLab work item labels are moved into a tags folder called "GitLab" to easily identify them.

By default, tags are only created on the first import of a work item, so there is an additional toggle in the GitLab configuration to force tag creation, even for existing tasks. This is useful if there are existing tasks created from GitLab that were imported before enabling the label import.

While the code is currently functional, I'm making this a draft pull request for 3 reasons: First, based on comments on the pull request to enable GitHub label imports, I'm looking into "generifying" the implementation so it can be used by other issue providers. Second, I want to make sure it's tested thoroughly. And third, I'm hoping for feedback since this is a significant change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "removeWOFF1": "node ./tools/remove-woff.js",
     "serveProd": "ng serve --configuration production",
     "start": "npm run electron:build && cross-env NODE_ENV=DEV electron .",
+    "startDebug": "npm run electron:build && cross-env NODE_ENV=DEV electron --remote-debugging-port=9223 .",
     "startFrontend": "npm run env && ng serve",
     "startFrontend:e2e": "npm run env && ng serve --port=4242",
     "startFrontend:prod": "npm run env && ng serve --configuration production",
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
     "release": "npm run release.changelog && npm run dist",
     "release.changelog": "conventional-changelog -i CHANGELOG.md -s -p angular",
     "removeWOFF1": "node ./tools/remove-woff.js",
+    "serveDev": "ng serve",
     "serveProd": "ng serve --configuration production",
     "start": "npm run electron:build && cross-env NODE_ENV=DEV electron .",
     "startDebug": "npm run electron:build && cross-env NODE_ENV=DEV electron --remote-debugging-port=9223 .",
--- a/src/app/features/calendar-integration/calendar-integration.service.spec.ts
+++ b/src/app/features/calendar-integration/calendar-integration.service.spec.ts
@@ -727,6 +727,9 @@ END:VCALENDAR`;
 
   describe('constructor', () => {
     it('should load skipped events from localStorage on init', () => {
+      // Clean up localStorage first
+      localStorage.clear();
+
       const skippedIds = ['event-1', 'event-2'];
       const today = new Date().toISOString().split('T')[0];
 
@@ -754,9 +757,15 @@ END:VCALENDAR`;
 
       const newService = TestBed.inject(CalendarIntegrationService);
       expect(newService.skippedEventIds$.getValue()).toEqual(skippedIds);
+
+      // Clean up after test
+      localStorage.clear();
     });
 
     it('should not load skipped events from different day', () => {
+      // Clean up localStorage first
+      localStorage.clear();
+
       const skippedIds = ['event-1', 'event-2'];
       const yesterday = new Date(Date.now() - 86400000).toISOString().split('T')[0];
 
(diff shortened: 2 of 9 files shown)


johannesjo:
I am personally not using the Gitlab implementation atm, so I can only give feedback on a code and conceptual level.

This definitely should be an opt in imho! Tags on Gitlab and other issue providers usually fill much different roles than the ones an average user would create locally.

It would be great to have this as a generic pattern for other issue providers!

Are there any other questions you have in particular?

johannesjo:
### Code review (AI assisted)

Found 1 issue:

1. **Validation functions return `true` after errors instead of `false`** (CLAUDE.md says "Handle errors explicitly (never swallow)" and "Fail fast, clear messages")

The changes in `validateSubTasks` and `validateReminders` now return `true` (success) after calling `_validityError()`, instead of `false` (failure). This means corrupted or invalid data will be accepted and persisted rather than rejected. The comment states "if we get here, the user dismissed the error so let's move on as if nothing happened" but this misunderstands the flow - if `devError()` throws, execution stops immediately; if the user dismisses or in production mode, the function continues and now returns `true`, accepting invalid data.

https://github.com/super-productivity/super-productivity/blob/442a94e2f0410a937ee9368ce81914d46c070155/src/app/pfapi/validate/is-related-model-data-valid.ts#L284-L287

https://github.com/super-productivity/super-productivity/blob/442a94e2f0410a937ee9368ce81914d46c070155/src/app/pfapi/validate/is-related-model-data-valid.ts#L363-L366

---

**Additional feedback** (lower confidence, for consideration):

- `_removeTagFromTree` in menu-tree.service.ts returns an object from `filter()` instead of a boolean - works by accident due to truthiness but is semantically incorrect
- `_createTagsFromLabels` is ~38 lines, exceeding the 30-line guideline
- `setTimeout(..., 100)` lacks documentation explaining the magic number
- Commented-out debug logging statements should be removed or uncommented
- New private methods use `any` type instead of `MenuTreeTreeNode[]`

These are suggestions rather than blockers - the main concern is the validation behavior change above.

---

Generated with [Claude Code](https://claude.ai/code)

<sub>If this code review was useful, please react with a thumbs up. Otherwise, react with a thumbs down.</sub>

kemsar:
> The changes in `validateSubTasks` and `validateReminders` now return `true` (success) after calling `_validityError()`, instead of `false` (failure). This means corrupted or invalid data will be accepted and persisted rather than rejected. The comment states "if we get here, the user dismissed the error so let's move on as if nothing happened" but this misunderstands the flow - if `devError()` throws, execution stops immediately; if the user dismisses or in production mode, the function continues and now returns `true`, accepting invalid data.

Thank you for the feedback! I can revert these changes. The reason I changed this was I was having an issue where one of my tasks had a reminder ID but somehow the associated reminder was missing. I'm not sure how that can happen, but the error was blocking all other data (tasks, projects, tags, etc.) from loading. I wasn't sure about the root cause so wanted to surface the error but not block the app load.

johannesjo:
> Thank you for the feedback! I can revert these changes. The reason I changed this was I was having an issue where one of my tasks had a reminder ID but somehow the associated reminder was missing. I'm not sure how that can happen, but the error was blocking all other data (tasks, projects, tags, etc.) from loading. I wasn't sure about the root cause so wanted to surface the error but not block the app load.

Usually I would say that using `devError()`  is the right call for this. We Definitely want to catch this during development, but you are rigth that a missing reminder id is not a critical error. BUT the data being in an invalid state can lead to problems down the line so it is probably better to avoid it  as early as possible.

kemsar:
@johannesjo - Suggestions on how to remedy invalid state data when developing? Seems like when I run different configs (i.e. installed locally vs debugging in IDE) it can cause some inconsistent data, which blocks debugging since it stops loading everything.

johannesjo:
> @johannesjo - Suggestions on how to remedy invalid state data when developing? Seems like when I run different configs (i.e. installed locally vs debugging in IDE) it can cause some inconsistent data, which blocks debugging since it stops loading everything.

Do you mean user data?

For that you can do the following:

  1. Clear IndexedDB via browser DevTools - Application → Storage → IndexedDB → delete the super-productivity database
  2. Use the app's built-in data export/reset - Export good data first, then reset, then reimport if needed
  3. Browser profile separation - Use different browser profiles for installed vs dev to keep data isolated
  4. Use incognito mode.

Issue #5814, opened 2025-12-27
💡 Opacity slider for background image


beerkumquatpome:

Garrett04:
Yes, there seems to be an issue with the color contrast of some icons and text for some light backgrounds.


<img width="647" height="150" alt="Image" src="https://github.com/user-attachments/assets/dcbd569c-130e-42ca-9333-87b4f47bc0fd" />

_light background in dark theme_

PR #5924, opened 2026-01-08
feat(ui): add backdrop to improve visibility of elements

Garrett04:
Add backdrop to improve visibility of elements so users can see the text and icons clearly, especially in cases where the text color matches the background image making it nearly impossible for users to read.


Garrett04:
### Before
#### Today page
image

#### Focus overlay
image


### After
#### Today page
image

#### Focus overlay
image

Garrett04:
I have some basic web development skills and know a bit of React. I'm trying to understand and get used to this app's codebase so i can contribute whenever.
Hope this pull request helps. 😁

johannesjo:
Thank you very much for this!

I don't think there is a good universal solution for this. Muting the background image is a valid approach, but it is ultimately up to the user to make this work with the background image they desire. Having a very bright background for dark mode is expected to not work very well.

I am ok with leaving this to the user for now.

Garrett04:
> Thank you very much for this!
> 
> I don't think there is a good universal solution for this. Muting the background image is a valid approach, but it is ultimately up to the user to make this work with the background image they desire. Having a very bright background for dark mode is expected to not work very well.
> 
> I am ok with leaving this to the user for now.

Yes you're right about that. I think it would be useful to have a setting to add the backdrop though. I'll maybe try to work on this issue #5814 since it's somewhere along the lines.

Also since i know the fundamentals of React, is there a way to learn the frameworks used in this app? if it's a youtube video you can link

johannesjo:
@Garrett04 Yes, this might be a good idea!

The framework we use is called angular. There should be tons of good tutorials out there if you google it.

Garrett04:
Thanks Johannes!
I'll try to implement this feature. It will take quite a while for me to get used to Angular so if anyone else needs to be assigned that issue, you can go ahead.

beerkumquatpome:
@Garrett04 Can we adjust the darkness with a slider? This is too dark for me

Garrett04:
> @Garrett04 Can we adjust the darkness with a slider? This is too dark for me

Thanks for the feedback @beerkumquatpome !
I'm actually working on a slider to adjust the opacity. You'll find it under Misc settings of the settings page soon.

PR #5944, opened 2026-01-09
Fixes Gitlab created tasks being marked incorrectly as updated

### LINKED ISSUE #5518: 🚨 Gitlab created tasks are marked having updates, even though nothing changed
### Version Used

v16.3.0 (I think), MS Store

### Expected Behavior

If nothing changes in the linked issue, user should not be prompted to review any updates.

### Current Behavior

I added a Gitlab connector, and for every issue it's polling, even though nothing changes - it does make a "bing!" and hints a task having an updates to review.

### Steps to Reproduce

1. Add Gitlab link
2. Add task from Gitlab's issue
3. Every now and then when polling occures, I am "bing!"'ed

### Console Output

```shell

```

kemsar:
Rather than comparing the issues last update timestamp with the last comment timestamp, it now compares the issue last update timestamp with the task `issueLastUpdated` property. This avoids the latency between an update and the update being logged in the comments.

Fixes #5518


johannesjo:

Issue #5887, opened 2026-01-04
🚨 When two tasks scheduled for the same time box they should appear next to each other


patrykpalej:
### Version Used

deb 16.7.3

### Expected Behavior

When I schedule two tasks for the same time box they should be visible both in the Schedule view, like in Google Calendar and similar.

### Current Behavior

Currently the new task appears under the old one. I spent some time until I noticed that I cannot see it because it's under an existing task. Such behaviour is unexpected IMO

### Steps to Reproduce

1. Schedule a task
2. Schedule another one to the same time box
3. One of them is not visible

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

PR #5949, opened 2026-01-10
fix(schedule): maintain visibility during task overlap on schedule (#5887)

miqh:
Hello there,

This changeset provides an (somewhat opinionated) approach to maintain the visibility of any overlapping tasks that appear in the Schedule view.

If I had to summarise broadly, the gist of this approach is based on keeping track of when tasks are "active" and assigning them into a "slot" (day column offset) all the while ensuring that the exact start and end of each task is visually discernable. I'll add some comments inline to help with review.

I've also added an additional E2E regression test case to ensure user access to overlapping tasks in the Schedule view is maintained.

Addresses:
- https://github.com/super-productivity/super-productivity/issues/5887

(I noticed there's no longer a pull request template since I last contributed. I guess it's been removed?)

### DIFF
--- a/src/app/features/schedule/schedule-event/schedule-event.component.scss
+++ b/src/app/features/schedule/schedule-event/schedule-event.component.scss
@@ -26,11 +26,6 @@
   z-index: 2;
   user-select: none;
 
-  // Bring hovered events above others
-  &:hover {
-    z-index: 10;
-  }
-
   // has to be for after elements
   overflow: visible !important;
   min-width: 0;
@@ -100,9 +95,9 @@
     content: '';
     position: absolute;
     top: 0;
-    bottom: calc(-1 * (var(--margin-bottom) + 1px));
+    bottom: 0;
     left: 0;
-    right: calc(-1 * (var(--margin-right) + 4px));
+    right: 0;
     z-index: 3;
     display: block;
   }
--- a/src/app/features/schedule/schedule-event/schedule-event.component.ts
+++ b/src/app/features/schedule/schedule-event/schedule-event.component.ts
@@ -74,6 +74,7 @@ export class ScheduleEventComponent {
   private _taskService = inject(TaskService);
 
   readonly T: typeof T = T;
+  readonly isDragPreview = input<boolean>(false);
   readonly isMonthView = input<boolean>(false);
   readonly event = input.required<ScheduleEvent>();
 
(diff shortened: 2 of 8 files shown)


src/app/features/schedule/schedule.model.ts
```diff
@@ -9,10 +9,9 @@ export interface ScheduleEvent {
   style: string;
   startHours: number;
   timeLeftInHours: number;
-  isCloseToOthersFirst: boolean;
```
miqh, line 12: I noticed there was a past attempt at addressing the visuals for Schedule view entries with the same start time, but there have been subsequent changes that render these effectively redundant.

- https://github.com/super-productivity/super-productivity/commit/038c4cac2c9569681844a3a13b2a4452f18e61a8
- https://github.com/super-productivity/super-productivity/commit/36714c9002e4810dac16efdde164c6f4c318afb2

If I've misunderstood, then please let me know what the intention with these properties were for.

src/app/features/schedule/schedule-event/schedule-event.component.ts
```diff
@@ -194,7 +189,21 @@ export class ScheduleEventComponent {
...
 
-  readonly style = computed(() => this.se().style);
+  readonly style = computed(() => {
+    const { overlap, style } = this.se();
+    // Arbitrarily chosen value that controls width reduction of this component
+    // whenever the underlying event duration overlaps with others
+    const overlapReductionFactor = 0.75;
+    return (
+      (!this.isMonthView() && !this.isDragPreview() && overlap
+        ? // eslint-disable-next-line no-mixed-operators -- conflicts with prettier formatting
```
miqh, line 199: Without the explicit lint ignore here, Prettier would keep trying to redact the brackets, which would otherwise resolve the lint.

I noticed there are many other places `no-mixed-operators` is sprinkled throughout the codebase for similar reasons. If this is an ongoing bother, I think a tweak to `.eslintrc.json` with `allowSamePrecedence` could resolve this.

src/app/features/schedule/schedule-event/schedule-event.component.scss
```diff
@@ -100,9 +95,9 @@
     content: '';
     position: absolute;
     top: 0;
-    bottom: calc(-1 * (var(--margin-bottom) + 1px));
```
miqh, line 103: I wasn't sure why this was the way it was, but it basically prevents users from selecting the gap between a task and the rightmost edge of a Schedule view day column.

I'm assuming the gap is to allow for situations where a user might want to schedule an overlapping task with a start time that's close to an existing one (like Google's calendar).

johannesjo: I honestly don't remember :D

johannesjo:
Thank you very much for this! As always: Great work! ❤️

PR #5953, opened 2026-01-11
fix(i18n): misspelled translation for 'NO_PLANNED_TASKS' in Polish

0x1618:
Zaplaniwano is an incorrect word. The correct word is zaplanowano.


johannesjo:
Thank you!

PR #5972, opened 2026-01-12
fix(ClickUp): undefined username

### LINKED ISSUE #5908: 🚨 ClickUp provider: error with username type assertion
### Version Used

v16.8.3

### Expected Behavior

ClickUp provider should parse data correctly and show tasks and workspaces. This requires considering the case where a user can have a `null` username.

### Current Behavior

ClickUp provider is failing with an error due to an error with username type assertion. The API response from ClickUp includes at least one user with a `null` username in my case, which is causing the type assertion error. It's a guest user but not all guest users have `null` usernames.

Loading available workspaces is also failing due to the same error, which is printed to console in this case rather than being displayed to the user.

### Steps to Reproduce

1. Search for ClickUp tasks or try to load available workspaces and have a user with a `null` username in the response
2. Notice the error shown about the username field of a user at a specific index not being a string

### Console Output

```shell
ClickUp: Error on typia.assert(): invalid type on $input.teams[0].members[5].user.username, expect to be string
```

theBenForce:
Updates the schema and tests to handle null/undefined usernames.

Fixes #5908


johannesjo:
Thank you very much!

PR #5980, opened 2026-01-13
Fix: links and remove extra symbols in main README

steindvart:
## Problem
- Wrong paths to resources.
- Links to resources with German localization.
- Extra symbols.

## Solution - what PR does
- Fixed paths.
- Made links without specifying the localization (default locale or English will be selected).
- Removed extra symbols in file.


johannesjo:
Thank you very much for this!

PR #5988, opened 2026-01-14
Fix small letters

steindvart:
## Problem
- Wrong capitalization and punctuation in the `START_OF_NEXT_DAY_HINT` string.
- Wrong capitalization and punctuation in the localization files. Translation of `e.g.`, without considering the fact that in other languages there may not be such an abbreviation and the word should be written in full and with a capital letter.
- Wrong translation of the `TIME_LOCALE_DESCRIPTION` string in `ru.json`.

## Solution: What PR does
- Fixed capitalization and punctuation in `en.json`, `ru.json`.
- Made correct translation of the `TIME_LOCALE_DESCRIPTION` string in `ru.json`.


johannesjo:
Thank you very much!

Issue #5989, opened 2026-01-14
💡 Rename "Domina mode" to a more descriptive voice reminder feature name


steindvart:
### Problem Statement

The current feature name "Domina mode" for the periodic voice cue/reminder functionality may be unclear or misleading for many users. The term does not clearly communicate the purpose of the setting and might cause confusion or negative associations.

There is also a problem with the correct translation of this name into other languages. In particular, for the Russian localization, this setting is literally called "Dominance Mode". Which does not give an understanding of what kind of function it is and causes strange associations. I think the problem is similar for other languages.

### Possible Solution

Rename this setting to something more descriptive and neutral, such as:
- "Voice Reminders" (favorite, on my opinion)
- "Voice Coach"
- "Focus Reminders"

These names more accurately reflect the feature's purpose and are likely to be more intuitive for users.

### Additional context

No code changes are necessary apart from renaming in the UI, settings, and related documentation. The feature itself is useful, this suggestion is only about more accurate and user-friendly naming.

steindvart:
Target PR is merged: #5990. Issue closing.

PR #5990, opened 2026-01-14
Rename "Domina mode" to a "Voice Reminder"

steindvart:
## Problem
- #5989 

## Solution: What PR does
- Implementation of ideas from #5989.
- Actual translation of this change for all locales.
- Removed obsolete tests for the voice reminder module.

### DIFF
--- a/src/app/features/voice-reminder/store/voice-reminder.effects.ts
+++ b/src/app/features/voice-reminder/store/voice-reminder.effects.ts
@@ -9,7 +9,7 @@ import { selectCurrentTask } from '../../tasks/store/task.selectors';
 import { speak } from '../../../util/speak';
 
 @Injectable()
-export class DominaModeEffects {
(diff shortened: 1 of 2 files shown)


steindvart:
@johannesjo what do you think about this changes? Why am I asking - I want to pour them in as soon as possible, before the changes in the main branch go far and I don't have to deal with conflicts later.

johannesjo:
Voice reminder is probably the more professional choice :D

johannesjo:
Thank you very much for this!

steindvart:
@beerkumquatpome

Hi! Thanks for reporting the problem. Yes, I am partially aware of it. I haven't translated all the previous mentions and internal settings of Domina mode yet, so that there won't be any problems with migration between versions. Anyway, thanks for letting me know!

I will try to check again all those places where you can specify the current name without having problems with migrations.

PR #6002, opened 2026-01-14
feat(notify): add ntfy integration for desktop notifications (fixes #…

### LINKED ISSUE #5419: 💡 Notification Support via NTFY
### Problem Statement

I have been having some issues with notifications on Android recently, and I think I might have one potential solution/option.

### Possible Solution

In addition to native notifications, there could be an option to send notifications from SP to an NTFY server.  This free and open-source notification service has a widely-available free server at https://ntfy.sh or anyone can self-host their own server (which I do).  Then, any application or process can execute a POST request against a properly-formatted URL to publish a notification to a topic on the service.  The call to send the notification can include options like subject, text, and priority (for urgency).  This can then be subscribed to by any one who has access to that topic, where the user can choose among the wide variety of options of client applications.  Android and iOS already have app-store based options, and desktop operating systems can all be integrated.

One could potentially incorporate the Apprise framework to get a large variety of notification methods/platforms, but that would likely require an additional binary or complexity which might not be suitable for distribution with SP's development and release workflow.

### Additional context

[Project Docs](https://docs.ntfy.sh/)

The quickstart describes options for sending a notification using a call by cURL, which can easily be duplicated with any other call to a URL, such as the WebDAV syncing or other network traffic.

Ki-Kolan:
This is my first PR at anything, so sorry in advance if something is wrong. I added an integration with ntfy to send notifications.

This pr adds the option in settings to toggle ntfy notifications on. There you can use the official ntfy.sh page or your self hosted instance. 

Closes #5419

### DIFF
--- a/src/app/core/notify/notify.service.ts
+++ b/src/app/core/notify/notify.service.ts
@@ -9,6 +9,9 @@ import { IS_ANDROID_WEB_VIEW } from '../../util/is-android-web-view';
 import { Log } from '../log';
 import { LocalNotifications } from '@capacitor/local-notifications';
 import { generateNotificationId } from '../../features/android/android-notification-id.util';
+import { HttpClient } from '@angular/common/http';
+import { firstValueFrom } from 'rxjs';
+import { GlobalConfigService } from '../../features/config/global-config.service';
 
 @Injectable({
   providedIn: 'root',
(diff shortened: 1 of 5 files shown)


johannesjo:
Thank you very much! Honestly I think this might be better suited to a plugin. Especially since there are possibly also legal implications, which I'd rather not have to deal with :D

johannesjo:
Could also be a great addition to the automations plugin maybe?

Issue #6008, opened 2026-01-15
💡 Add ability to select multiple tags for a task without hiding the menu


steindvart:
### Problem Statement

When you select tags for a task through the graphical menu, this menu collapses when you select the tag. And to add another tag, you need to open it again. This is very inconvenient.

![Image](https://github.com/user-attachments/assets/1426700a-afc1-4dbe-b72c-688afa24ebb0)

Similar behavior when selecting a tag when creating a new task.

![Image](https://github.com/user-attachments/assets/95ff6dbb-916a-4fd4-8857-c42ca458b451)

### Possible Solution

Add the ability to select multiple tags for a task in any mode (create, edit) without hiding the menu.

puetsua:
This bugged me for a while too, thanks for fixing it. :D

steindvart:
Related PR (#6010) is merged. Issue is implemented.

johannesjo:
Thank you very much!

Issue #4436, opened 2025-05-22
🚨 Project list not showing and selecting automatically the project under the finger when pressing "Move to project"


Copilotich:
### Version Used

Android version of 13.0.10

### Expected Behavior

When choosing to move task to another project through long press menu the project list appears and let you decide which project you put the task in.

### Current Behavior

When choosing to move a task to another project through long press menu, if you select the "Move to project" tapping with your finger too close to the edge of the screen where the project list is supposed to appear afterwards, there will be no time to chose, the menu will appear under the finger and immediately the finger will select the project that would appear in the same spot that the "Move to project" button was pressed and it won't let you select any other the project nor let the list appear as it has been chosen already.
Here I leave a video

![Image](https://github.com/user-attachments/assets/3af91ae9-097f-4c5a-af4d-1b8050a5154b)

### Steps to Reproduce

Choose a task and long press it near the center of the task and the long press menu will appear on either side, if the menu is close to the edge when pressing "Move to project" your finger would also be where the list of projects would also appear and the finger will chose the project that appears under it when pressing move to project without time to react.

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
This is unfortunately a problem with the material component:
https://github.com/angular/components/issues/30888#event-17499605141

johannesjo:
This should be fixed now.

PR #6010, opened 2026-01-15
Ability to select multiple tags for a task without hiding the menu

steindvart:
## Problem
- #6008

## Solution: What PR does
- Implementation of ideas from #6008.

## Current behaviour

Animation2


johannesjo:

johannesjo:
### Code review

Found 1 issue:

1. Adding `stopPropagation()` conflicts with existing code comments that explicitly warn: "Don't call stopPropagation() - we need the event to bubble for menu closing"

https://github.com/super-productivity/super-productivity/blob/8a4b43a446e5963b08bdb095793d834ef0bf4236/src/app/features/tasks/task-context-menu/menu-touch-fix.directive.ts#L118-L120

This comment appears in both `menu-touch-fix.directive.ts` and `mat-menu-touch-monkey-patch.ts`. The change in `task-context-menu-inner.component.html` is in the same directory and may interfere with touch device menu behavior.

steindvart:
@johannesjo, understood! Researched the origin issue of this problem: https://github.com/super-productivity/super-productivity/issues/4436. As noted in this discussion, the problem is with the material component of the Angular framework.

Therefore, I think if we separate the `$event.stopPropagation()` from the button (because `MenuTouchFixDirective` works at the button level (TARGET phase)), then it should work correctly. Push to branch this solution.

johannesjo:
Alright! Do you need help with the conflicts @steindvart  ?

steindvart:
> Alright! Do you need help with the conflicts?

No, thanks! I will resolve it tomorrow. And before that, I'll investigate your changes.

steindvart:
> Do you need help with the conflicts?

Done.

johannesjo:
Thank you very much!

PR #6019, opened 2026-01-15
Add AutoPlan plugin to community-plugins.json

00sapo:
Added AutoPlan plugin with description and URL.

I created this today mainly vibe coding it.

Tests are welcome. Unfortunately, there are a few caveats that I hope will be solved with the improvement of Super Productivity.

### DIFF
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -13,5 +13,10 @@
     "name": "QuestArc",
     "shortDescription": "Turn your workflow into an adventure! Gamify your tasks with Atomic Habit quests, epic boss fights, earnable badges, and a customizable hero profile.",
     "url": "https://codeberg.org/Nexumia/QuestArc"
+  },
+  {
+    "name": "AutoPlan",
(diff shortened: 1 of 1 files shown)


00sapo:
In order of severity

The main caveat is that Super Conductivity does not distinguish between schedule date and due date. 

The second caveat is that superconductivity doesn't allow scheduling only part of a task. This leads AutoPlan to create many small tasks when the task is long and to many removal/creation of new identical tasks. This mechanic likely makes the plugin incompatible with other in the future.

The third one is that the plugin api don't expose all the global settings, which lead to proliferate the settings at the menu level (in this case, the working hours).

The fourth caveat is that the plugin api don't allow to access boards, that could be used as lists of tags to be/not to be scheduled with easy to manage priorities (i.e. the order of the tasks in the list)

The fifth caveat is the integration of calendars, whose events should always be imported, not only the ones of today, and as "Event" not "Task"

johannesjo:
Thank you very much for building and sharing this! <3

Issue #6024, opened 2026-01-16
💡 Main header redesign: move tracking, counters, and focus controls to the left


steindvart:
### Problem Statement

Currently, key controls like time tracking, simple counters, and the focus button are grouped with other elements on the right of the main header. This makes the right side cluttered and can interfere with navigation and visibility, especially for users with long names of tasks. It can reduce usability and cause important buttons to be less discoverable.

But the advantage of the old design is that it can be familiar to older users.

<img width="1779" height="619" alt="Image" src="https://github.com/user-attachments/assets/3b7aba55-4d8b-4ced-9fb5-95058c0a70ab" />

### Possible Solution

Redesign the main header by moving the time tracking controls, simple counters, and focus button from the right to the left side. This will group related controls, declutter the right side, and make access more consistent for users.

But the disadvantage of this design is that buttons may change position whenever the list title changes.

<img width="1771" height="631" alt="Image" src="https://github.com/user-attachments/assets/0fcb532c-f704-428e-bd58-e2561c65fc5e" />

A video demo of this redesign: https://youtu.be/vf_9LpCJPuE

### Describe alternatives you've considered

- Keeping the current layout with all controls on the right.
- Make a setting that allows the user to choose their preferred design.
- Leave all the items on the right, but group them into their own categories.

### Additional context

Discussion originated from #5984. Community feedback and further input are encouraged to reach the best solution.

steindvart:
Also made this variant: `Leave all the items on the right, but group them into their own categories.`

![Image](https://github.com/user-attachments/assets/debd02b3-53d6-4693-8135-4e0cd6eb1e47)

<img width="1243" height="533" alt="Image" src="https://github.com/user-attachments/assets/52fe3de6-dccb-41ed-8643-fd7e262eebd3" />

@johannesjo

steindvart:
The above option (with right grouping) seems to me to be the best at the moment.

johannesjo:
Thank you very much for these designs! It definitely helps to visualize these things!

Personally, I think I prefer this:

Add task has to come first, since it is arguably the most important button. It is also important to distinguish it from all other buttons. As an alternative, we could move it down into the list similar to what we do for the planner and the boards feature.  

I like having a visual seperation. But I think just leaving a gap should be enough. We should keep the base design minimal.

Moving the sync button to the rigth feels good. Since it is also a different category maybe we also add a gap here to the rigth?

What do you think?

steindvart:
[...]

PR #6025, opened 2026-01-16
Main header redesign

steindvart:
## Problem
- #6024 

## Solution: What PR does
- Move tracking, counters, and focus controls to the left.
- Improve common design (size, gap) of tracking, counters, and focus controls.
- Enhance hover effects for the current-task-title of play button
 with smoother transition.
- Fix simple-counter-button icon positions.


johannesjo:
Thank you very much!

PR #6026, opened 2026-01-16
Always use project URL from issue ID

cbergmann:
With this change the correct project URL for an issue is always used. This allows to set `..` as "user name/project" and use the global search in selfhosted gitlab instances. This will enable the user to find and use issues from different projects.


cbergmann:
I Can't see how the changes would result in the calendar tests failing.

johannesjo:
Thank you very much!

PR #6048, opened 2026-01-18
Improve issue templates

steindvart:
## Problem
- If an optional field (console-output, etc.) is not filled in, it will still end up in the issue text. This creates redundancy and distracts from the main text.

## Solution: What PR does
- Include optional fields in final issue text only when it filled.
- Add emoji to feature_request template to draw attention to important notes.


johannesjo:
Thank you!

PR #6049, opened 2026-01-18
Fix/manual break overtime

Kuntal6002:
fix(focus-mode): do not auto-complete session when manual breaks enabled

### Problem
When "Manually start breaks" is enabled, focus sessions were automatically completed
as soon as the configured duration was reached, causing overtime to be lost.

### Solution
- Prevent auto-completion of focus sessions when manual breaks are enabled
- Keep existing behavior unchanged for automatic breaks and Flowtime mode

### Tests
- Updated existing effect tests to mock focus mode config
- All tests passing locally


johannesjo:
Thank you very much!

beerkumquatpome:
@johannesjo 
@Kuntal6002

v17.1.3

It still auto-completes a session when "Manually start breaks" is enabled.
Is this not implemented?

johannesjo:
I look into it @beerkumquatpome ! Thanks for checking!

johannesjo:
@beerkumquatpome Thanks for checking. I'll have a look.

PR #6058, opened 2026-01-19
Improve settings tabs navigation on small screens

steindvart:
## Problem
- Inconvenient navigation through the settings tabs on small screens. You need to scroll left and right.

## Solution: What PR does
- Hid tabs labes in `md` size screens.
- Added section titles to each tab in settings.

## Demo

Animation6


johannesjo:
Very good! Let's do it like this!

johannesjo:
Thank you very much once more @steindvart !

Issue #6045, opened 2026-01-18
💡 Move task-related settings to a dedicated Task Settings tab


steindvart:
### Problem Statement

Currently, task-related settings such as `isConfirmBeforeTaskDelete`, `isAutoAddWorkedOnToToday`, `isAutMarkParentAsDone`, `isTrayShowCurrentTask`, `defaultProjectId`, `isTurnOffMarkdown`, `taskNotesTpl`, `Short Syntax`, and `Sound` are scattered across different sections, with many placed in the 'Misc' section.

As a result, 'Misc' is overloaded, making it difficult for users to find and manage settings related to tasks quickly. This can cause confusion, reduces ease of use, and negatively impacts user experience when configuring task workflow preferences.

### Possible Solution

Introduce a dedicated 'Tasks' settings tab within the application settings. Move all the task-related settings listed above to this new tab to centralize them. This will streamline navigation, make configuration easier for users focusing on tasks, and declutter the 'Misc' section for better usability overall.

My current sketch:

<img width="1212" height="1440" alt="Image" src="https://github.com/user-attachments/assets/aea6cb90-1c83-4a13-bc7f-f2a2859e20f3" />

### Describe alternatives you've considered

Some alternatives considered: 
- Keeping organization as is, but this does not improve user experience.
- Only moving a subset of options, but this leads to potential future inconsistency as more task options are added.

Ultimately, a single dedicated tab is clearer and future-proof.

### Additional context

Related settings for quick reference:
- `isConfirmBeforeTaskDelete`
- `isAutoAddWorkedOnToToday`
- `isAutMarkParentAsDone`
- `isTrayShowCurrentTask`
- `defaultProjectId`
- `isTurnOffMarkdown`
- `taskNotesTpl`
- `Short Syntax` (form)
- `Sound` (form)

⚠️ Migration are required!

steindvart:
Target PR is merged: #6053. Issue is implemented.

johannesjo:
Thank you very much @steindvart !

PR #6063, opened 2026-01-19
Tasks tab - provide changes from #6058

steindvart:
## Problem
- Because PRs (#6058 and #6045) merged separately, changes in tasks tab labels not implemented.

## Solution: What PR does
- Made changes as from #6058 for tasks tab.


johannesjo:
Thank you !

Issue #5789, opened 2025-12-22
💡 Add support for collapsible sections/headings in project and task notes


BugBusterQA:
### Problem Statement

In long notes (hundreds of lines), there is no way to collapse sections. All content is always expanded, forcing constant scrolling to navigate or hide irrelevant parts.

### Possible Solution

Support collapsible sections in Markdown notes:
- Auto-detect headings (# H1, ## H2, etc.) and make them foldable (click to collapse content under them).
- Or support HTML-like <details><summary>Title</summary>Content</details> syntax, rendered as native collapsible blocks.

Similar to Obsidian or GitHub Markdown previews.

### Describe alternatives you've considered

Manual splitting into multiple notes (but then sidebar scrolling issue), or external tools.

### Additional context

Essential for structured long documentation in notes (e.g., project specs with multiple sections).

PR #6066, opened 2026-01-19
feat: add sections in projects

qievenz:
## Problem

Projects and Inbox lacked organization capabilities for grouping related tasks, making it difficult to structure work effectively. 
https://github.com/super-productivity/super-productivity/issues/5789

## Solution: What PR does

1. Sections Feature
  Add sections to projects and inbox via context menu
  Organize tasks by dragging them into sections
  Reorder sections via drag and drop
  Section actions: Edit title, delete (including all tasks), collapse/expand
  "No Section" area for ungrouped tasks

2. Markdown Paste with Auto-Section Creation
  Paste markdown with H1 headers (#) to auto-create sections
  Example: Pasting a README automatically creates sections from headers and populates them with tasks
  Saves time when importing structured task lists


johannesjo:
Thank you very much for this!  This is well done, but tbh I am a bit weary of the complexity this will add and I also feel that this workflow can be replicated already using the boards feature. According to the upvotes of the issue this seems to be more of a niche use case. Esepcially with the recent big changes under the hood, I'm a bit weary to add yet more technical debt.

Before starting to work on these big issues I would appreciate it, if you would write a comment that you'd like to start to work on it. I didn't expect a PR on this tbh and now I am unsure what the best course of action is with this.

Maybe we can add this in the form of a plugin? We would have to expand on the general plugin capabilities by adding something like a task view customizer or a similar pattern. In general  I feel that plugins are better suited to this more advanced use cases. The core of the app should remain as accessible and simple as possible.

johannesjo:
<details>
  <summary>Technical Code Review (for later use)</summary>

### Code Review

Comprehensive analysis identified **11 critical/high-priority issues** across code quality, performance, and sync architecture.

---

## Critical Issues (Must Fix Before Merge)

### 1. Effects using `inject(Actions)` instead of `inject(LOCAL_ACTIONS)`

**Severity:** CRITICAL - Sync correctness violation

CLAUDE.md explicitly states: "ALL NgRx effects MUST use \`inject(LOCAL_ACTIONS)\` instead of \`inject(Actions)\`. Effects should NEVER run for remote sync operations."

This will cause the deleteSection effect to run during remote sync replay, causing duplicate task deletions and potential data corruption.

https://github.com/super-productivity/super-productivity/blob/adca4e9edd71e68d278a3d1d9df175b2e33af464/src/app/features/section/store/section.effects.ts#L10-L12

**Fix:** Change `inject(Actions)` to `inject(LOCAL_ACTIONS)`

---

### 2. Section deletion should use meta-reducer instead of effect

**Severity:** CRITICAL - Partial sync vulnerability

CLAUDE.md Point 9: "When one action affects multiple entities, use meta-reducers instead of effects to ensure all changes happen in a single reducer pass. This creates one operation in the sync log, preventing partial sync and state inconsistency."

Current implementation creates TWO operations:
1. Delete section
2. Delete tasks (from effect)

This can cause partial sync where the section is deleted but tasks remain orphaned if sync fails between operations.

https://github.com/super-productivity/super-productivity/blob/adca4e9edd71e68d278a3d1d9df175b2e33af464/src/app/features/section/store/section.effects.ts#L13-L27

**Fix:** Follow the pattern in `tag-shared.reducer.ts` - create `section-shared.meta-reducer.ts` that handles section deletion and task cleanup atomically in a single reducer pass.

Reference: https://github.com/super-productivity/super-productivity/blob/master/src/app/root-store/meta/task-shared-meta-reducers/tag-shared.reducer.ts#L219-L261

---

### 3. `moveProjectTaskToSection` creates non-atomic multi-entity update

**Severity:** CRITICAL - State desynchronization

The effect dispatches two separate actions (`updateTask` and `updateProject`), creating two operations in the sync log. This breaks atomicity and can cause:
- Task has new sectionId but old position in project.taskIds
- Conflict resolution only sees partial state
- Partial sync if network fails between operations

https://github.com/super-productivity/super-productivity/blob/adca4e9edd71e68d278a3d1d9df175b2e33af464/src/app/features/project/store/project.effects.ts#L165-L177

**Fix:** Make `moveProjectTaskToSection` a persistent action and handle it atomically in a meta-reducer (similar to tag deletion pattern).

---

### 4. Missing schema migration for `task.sectionId`

**Severity:** HIGH - Backward compatibility broken

Adding `sectionId?` to the Task model without a schema migration will cause:
- Older clients to fail Typia validation when receiving tasks with sectionId
- Potential sync failures or data loss during mixed-version rollout
- Crashes on clients that don't expect the field

https://github.com/super-productivity/super-productivity/blob/adca4e9edd71e68d278a3d1d9df175b2e33af464/src/app/features/tasks/task.model.ts#L91

**Fix:** Add schema version bump and migration in `schema-migration.service.ts` to initialize `sectionId: null` for all existing tasks.

---

## High Priority Issues

### 5. Section order conflicts undetectable

**Severity:** HIGH - Data corruption

The `updateSectionOrder` action uses `ids[0]` as the entityId, which breaks conflict detection. Concurrent section reorderings from different clients will not be detected as conflicts, leading to corrupted section order.

https://github.com/super-productivity/super-productivity/blob/adca4e9edd71e68d278a3d1d9df175b2e33af464/src/app/features/section/store/section.actions.ts#L38-L42

**Fix:** Use `PROJECT` as entityType with `projectId` as entityId (similar to tag ordering).

---

### 6. Translation keys added to `es.json`

**Severity:** HIGH - Process violation

CLAUDE.md states: "When adding translation keys, only edit \`en.json\` - never edit other locale files directly."

https://github.com/super-productivity/super-productivity/blob/adca4e9edd71e68d278a3d1d9df175b2e33af464/src/assets/i18n/es.json#L22-L24

**Fix:** Remove translations from `es.json`, keep only in `en.json`.

---

### 7. Development test file committed

**Severity:** HIGH - Cleanup required

Temporary test file should not be in the repository.

https://github.com/super-productivity/super-productivity/blob/adca4e9edd71e68d278a3d1d9df175b2e33af464/test-sections.js#L1-L19

**Fix:** Remove `test-sections.js` from the PR.

---

## Performance Issues

### 8. Non-memoized selector factory breaks performance

**Severity:** MEDIUM - 5-15% CPU overhead

`selectSectionsByProjectId` is a selector factory that creates a new selector instance on every call, breaking memoization. This causes O(n) filtering on every state change.

https://github.com/super-productivity/super-productivity/blob/adca4e9edd71e68d278a3d1d9df175b2e33af464/src/app/features/section/store/section.selectors.ts#L22-L24

**Fix:** Use `createSelector` with proper memoization or move to component-level computed signal.

---

### 9. O(n×m) computed signal in work-view

**Severity:** MEDIUM - 10-30ms per task update

The `sectionsWithTasks` computed signal uses nested iteration (tasks.filter for each section), creating O(n×m) complexity. With 100 tasks × 10 sections = 1,000 operations per computation.

https://github.com/super-productivity/super-productivity/blob/adca4e9edd71e68d278a3d1d9df175b2e33af464/src/app/features/work-view/work-view.component.ts#L152-L169

**Fix:** Build a section→tasks map in O(n) first, then iterate sections.

---

### 10. Memory leaks from unmanaged subscriptions

**Severity:** MEDIUM - Memory leaks

Dialog subscriptions in `work-context-menu.component.ts` and `work-view.component.ts` lack cleanup with `takeUntilDestroyed()`.

CLAUDE.md anti-patterns: "Subscribing without cleanup → Use \`takeUntilDestroyed()\` or async pipe"

https://github.com/super-productivity/super-productivity/blob/adca4e9edd71e68d278a3d1d9df175b2e33af464/src/app/core-ui/work-context-menu/work-context-menu.component.ts#L64-L71

https://github.com/super-productivity/super-productivity/blob/adca4e9edd71e68d278a3d1d9df175b2e33af464/src/app/features/work-view/work-view.component.ts#L880-L916

**Fix:** Add `private _destroyRef = inject(DestroyRef);` and use `.pipe(takeUntilDestroyed(this._destroyRef))`

---

### 11. Markdown import creates excessive operations

**Severity:** LOW - Network overhead

Importing large markdown (50 sections, 500 tasks) creates 550+ operations in rapid succession, risking event loop blocking.

CLAUDE.md Point 11: "Event Loop Yield After Bulk Dispatches: When applying many operations to NgRx in rapid succession, add \`await new Promise(resolve => setTimeout(resolve, 0))\`"

https://github.com/super-productivity/super-productivity/blob/adca4e9edd71e68d278a3d1d9df175b2e33af464/src/app/features/tasks/markdown-paste.service.ts#L71-L144

**Fix:** Add event loop yields every 10 sections during markdown import.

---

## Summary

**Critical Issues:** 4 (sync correctness, data loss risk)  
**High Priority:** 3 (process violations, backward compatibility)  
**Performance:** 4 (CPU overhead, memory leaks)

**Recommendation:** Address Critical Issues #1-4 before merge. These violate core architectural principles and will cause data loss/corruption in multi-device sync scenarios.

The section feature is valuable, but the implementation needs alignment with the established sync architecture patterns (meta-reducers for multi-entity changes, LOCAL_ACTIONS for effects, schema migrations for new fields).

---

</details>

qievenz:
Thank you for the detailed feedback! I understand your concerns about complexity and technical debt, especially after recent major changes.

You're absolutely right - I should have commented on the issue first before starting work. I got excited about the feature and dove straight into implementation, which wasn't the best approach.

Regarding your points:
- Boards vs Sections: While boards offer grouping, I found sections to be lighter-weight for simple organization within a single project view (no need to switch tabs). However, I understand if this feels redundant given existing features.
- Plugin approach: I'm very open to exploring this as a plugin instead. This would indeed keep the core clean while still serving users who need this workflow.

Next steps: I'm flexible here - we can:
1. Close this PR if you feel it doesn't fit the core vision
2. Park it and revisit after enhancing plugin capabilities

Let me know what you prefer!

johannesjo:
Thank you very much for your understanding! 

Let's park it for now. I evaluate how we can extend plugin capabilities to allow for things like this.

hiJanes:
i also want this for mobile version

jphaupt:
I used the sections feature in todoist/ticktick religiously, so having this in Super Productivity would be great for me.

johannesjo:
Rebased onto current master and added review-batch fix commits.   Force-pushed via maintainer-edit access — old commit hashes are gone but authorship preserved on your original commits. Sorry about any orphaned review threads.

johannesjo:
Thank you very much for this again @qievenz !  After thinking about it a couple of times I changed my mind. This will be added with the next release :)

Issue #6075, opened 2026-01-21
💡 Convert text 'Cancel' buttons to 'Close' in modal dialogs


steindvart:
### Problem Statement

Currently, the words 'Cancel' are used to close modal windows, which is not entirely accurate and may lead to the association that this button is not for closing the window, but for canceling an action or resetting settings.

### Possible Solution

Use the label 'Close' instead of 'Cancel' for the corresponding buttons in modal dialogues.

johannesjo:
What dialogs do you have in mind? It probably depends on the context. For the schedule dialog for example using "Cancel" seems fine to me, since it aborts any changes made in the dialog. Same is true for the Time Spent/Estimate Dialog.

steindvart:
@johannesjo 
Yes, indeed, it depends on the context. I think that where there is a `Save` action, `Cancel` should be left there. Because `Close` can be associated with `Save and Close`, unlike `Cancel`.

And as for the elements where there are buttons like `Schedule` (negative action), `Unschedule` (positive action) - there you should make `Close` (as an explicit action different from a negative one).

johannesjo:
As a rule of thumb I'd say we should:

Use "Cancel" when potential changes are discarded (e.g. schedule, time estimate dialog).

Use "Close" when nothing happens.

steindvart:
> e.g. schedule

Do you think so? Don't you think that in the case of the `Unchedule` and `Schedule` buttons, `Cancel` will be more associated with `Unchedule`?

johannesjo:
If you made changes to form that are not submitted I would always prefer "Cancel" over "Close", since – as you mention – "Close" could also be understood as "Save and Close" (or the user thinks changes are saved instantly).  Since we are consistent with it and it is a common UX pattern, I would not expect too many users to be confused about this :)

steindvart:
@johannesjo 
> Use "Cancel" when potential changes are discarded (e.g. schedule, time estimate dialog).
> Use "Close" when nothing happens.

Okay, then we'll follow this rule. Thank you for discussing this and coming to a common understanding!

Closing issue.

PR #6103, opened 2026-01-22
Fix unlocalization text and 'Close' for single buttons

steindvart:
## Problem
- Untranslated text.
- 'Close' not for all single buttons (#6075).

## Solution: What PR does
- Fixed described above problems.


johannesjo:
Thank you very much!

PR #6104, opened 2026-01-22
docs(ci): Exclude docs/wiki/** from some actions

zenoprax:
## Problem

The modified Actions in this PR showed up during the initial v0.1 PR and added noise to the process and added delays/waste to what would otherwise be a 10s workflow to copy markdown files.

## Solution: What PR does

Relevant actions had a path exclusion added. Some minor YML re-formatting needed to ensure validity.


johannesjo:
Thank you very much!

PR #6106, opened 2026-01-22
fix(i18n): correct mistranslation in Korean

zzoyu:
## Problem

The Korean translations for `NOT_URGENT_IMPORTANT` and
`NOT_URGENT_NOT_IMPORTANT` had the same meaning,
which removed the intended distinction between importance levels.

## Solution: What this PR does

- Corrected the Korean i18n JSON entries
- No changes were made to application logic or other locales

## Note

Similar issues may exist in other locales (e.g. Japanese),
but this PR is intentionally limited to Korean to avoid
introducing inaccurate translations.


johannesjo:
Thank you very much!

PR #6134, opened 2026-01-24
docs(wiki) v0.2: refactor and populate meta notes

zenoprax:
Substantial re-work and addition of meta notes. Still many stub notes
and incomplete placeholders but the majority of the structure is
established and filling out the actual content is now a priority.


johannesjo:
Shall I merge this @zenoprax  or should we address the lint issues first?

zenoprax:
There aren't any linting issues, it's the same permissions problem as before (https://github.com/super-productivity/super-productivity/actions/runs/21310310164/job/61345523023?pr=6134#step:6:40)

Safe to merge though I don't like getting an "error: action failed" email for every PR even when there are no real problems. There must be some way to give the workflow enough permissions to commit changes.

zenoprax:
I just submitted #6150 that should address the permissions. Naming revised too as it was not clear which job had failed without going into the job itself.

johannesjo:
Thank you very much!!

PR #6158, opened 2026-01-25
docs/fix broken links

zenoprax:
- **docs(wiki): update use of aliases**
- **docs(wiki): fix broken links from aliases**

Also a quick test of the new action permissions.


zenoprax:
CI definitely runs quicker now and the workflow+job naming is much better but I'm not sure what's wrong with the linting.

`pymarkdownlnt` doesn't detect those errors when I run it locally. Maybe a version difference.

Regardless, PR is good to merge.

EDIT: I forgot I switched branches before linting and pushing... errors are correct; however, the main README shows 39 errors! Will need to rework that markdown before borrowing from it again.

johannesjo:
Thank you!

PR #6170, opened 2026-01-25
feat(i18n): update Turkish translations

Cyber-Syntax:
## Problem

The Turkish translations in the application are outdated and do not reflect the latest changes in the user interface.

## Solution: What PR does

This update refreshes the Turkish language translations to ensure they are accurate and up-to-date with the current application content.


johannesjo:
Thank you!

Issue #5851, opened 2026-01-01
💡 Windows Fullscreen Mode (With F11)


P-A-V-A:
### Problem Statement

There is no full screen mode on windows app. (Will be useful for hiding taskbar and etc. to reduce distraction and keeping focus.)

### Possible Solution

F11 shortcut like other apps can be added.

### Additional context

I appreciate this work!

lntutor:
The F11 fullscreen behavior was implemented in #6183, but the desktop-only shortcut is still missing from the official keyboard shortcut reference. I will add the minimal wiki documentation required by the current documentation guide, without changing the existing implementation.

PR #6183, opened 2026-01-26
feat(ui): Desktop Full Screen Mode (F11) #5851

benjamhooper:
## Problem

No Fullscreen to desktop (F11)

Adding this feature because of the ask from #5851 

## Solution: What PR does

Provides full-screen capability to the desktop application 

explorer_qUY3pXN3vi


johannesjo:
Thank you very much!

Issue #5108, opened 2025-09-14
💡 Settings - Copy Version Number


OLLI-S:
### Problem Statement

When I report an issue vor Super Productivity, I also have to specify the version number.
Currently I have two options:
* Type the version number manually in GitHub
* Click on the version number in the settings, copy the version number form the Changelog and paste it in GitHub

### Possible Solution

Add a "Copy" button behind the version number, which copes the version number into the clipboard.

### Describe alternatives you've considered

Allow to highlight the version number using the mouse (but this is not accessible via the keyboard).

OLLI-S:
Maybe you add a button "Copy Product Information" where you copy not only the version number, but also other relevant data, like the enabled Plugins or the used Theme.

johannesjo:
That is great idea!

suraj-mandal:
If this is finalized, I can work on this issue. Since this repo is new to me, this can be a good starting point for me to work on this

suraj-mandal:
@johannesjo

suraj-mandal:
Currently, working on this, will be ready by this weekend. Give me that much amount of time.

Sergih28:
@johannesjo Should this issue be closed, as it already has the related PR #6184 merged?

OLLI-S:
@Sergih28 You are right, I am closing the issue

PR #6184, opened 2026-01-26
feat: Copy Version Number #5108

benjamhooper:
## Problem

Version number within the Settings page was a direct link to the Github Changelog. #5108 Asked to change it to be able to copy it to the clipboard.

## Solution: What PR does

Added the ability to copy the version number to the clipboard by changing it. I split the difference on the ask in the issue so that the Changelog is explicitly there that is linkable, and the version number is explicitly "copyable".

electron_1XjyXI1yZd


johannesjo:
Thank you very much!

PR #6194, opened 2026-01-26
Changed 'Due' to 'Schedule' in 'Add task' UI

Gitoffthelawn:
This change makes the function of the button accurate.   But because it's a change people will certainly notice, I made it a separate PR from other changes (if we need to discuss).

## Problem

SP currently doesn't support due dates.  It only supports scheduling dates.  As such, the affected button in the 'Add Tasks' UI was confusing and misleading.

## Solution: What PR does

Makes the button's purpose clear and accurate.


johannesjo:
Thank you!

Gitoffthelawn:
You're welcome!

PR #6209, opened 2026-01-26
Move tasks to and from backlog when dragging to a panel with backlogState set

vonalbert:
## Problem

In the boards feature, when a task is dragged to a different panel, the task properties are updated so that it can be displayed in the destination panel. However, this does not work correctly with the backlog state option of the panel.

## Solution: What PR does

This PR adds support to the board-panel component to correctly update the backlog state when a task is dropped.
If the task’s project has the backlog feature enabled and the destination panel does not have backlogState set to All, the backlog state is updated accordingly.

The update is executed as the last step of the drop event to ensure the target project is correctly evaluated, as the task’s project may change during this action. However, it would likely be sufficient to execute it after the moveToOtherProject action.


johannesjo:
Thank you very much! <3

PR #6227, opened 2026-01-27
Increase ping interval and idle time in order to not waste CPU cycles.

tectrixdev:
If I understand the codebase correctly, this will do exactly as the title states, if not, please educate me :). Also is there a way to turn of idle checking when it is  also turned of in the settings?

## Problem

It should solve unneeded CPU cycles in the background.

## Solution: What PR does

Simply increase idle checking time, and the minimum amount to 60 seconds, as that's the minimum you can even configure in the app.

### DIFF
--- a/electron/CONFIG.ts
+++ b/electron/CONFIG.ts
@@ -1,5 +1,5 @@
 export const CONFIG = {
   D_BUS_ID: 'com.super.productivity.service',
-  IDLE_PING_INTERVAL: 5000,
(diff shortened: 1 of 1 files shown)


tectrixdev:
I'm also trying to make an option/flag to start the app minimized, but I can't really seem to find how the whole startup process works. (I've found most things, but I can't really seem to manipulate it or stop it from spawning the main window and killing the overlay after startup.

johannesjo:
Makes sense. Thank you!

PR #6240, opened 2026-01-28
docs(ci): prevent sync job from running on PR

zenoprax:
The workflow fails during the "sync" job on a pull request because of
repo-specific settings that are not set. These can be a security problem
to enable and are not needed for this workflow. Since accepting a PR
will cause a merge (and therefore a push by the author) this push will
then trigger the sync job with the correct authorization and will
therefore succeed as observed during my initial testing.

The only actions needed for the PR itself is linting.

### DIFF
--- a/.github/workflows/wiki-sync.yml
+++ b/.github/workflows/wiki-sync.yml
@@ -21,8 +21,6 @@ concurrency:
 
 permissions:
   contents: write
-  pull-requests: write
-  actions: write
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much!

Yeah, I probably should improve on the CI workflows, once I find the time.

zenoprax:
Thanks. All good, I was chasing the wrong problem from the start with my wiki sync workflow. The way it's written now makes it more portable for independent publishing as a "marketplace workflow" too.

Issue #6044, opened 2026-01-18
🚨 Long break occurring after session 5 instead of 4


beerkumquatpome:
### Version Used

  v17.0.0-RC.7

### Expected Behavior

Session 1
Short Break
Session 2
Short Break
Session 3
Short Break
Session 4
Long Break

### Current Behavior

Session 1
Short Break
Session 2
Short Break
Session 3
Short Break
Session 4
Short Break
Session 5
Long Break

### Steps to Reproduce

-

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

Issue #6000, opened 2026-01-14
🚨 Start focus sessions with banner only


beerkumquatpome:

johannesjo:
Yes, this probably makes sense.

johannesjo:
I'll add another button to the banner so there is no neccessity to go to the overlay at all.

beerkumquatpome:
@johannesjo 

v17.0.12

### Bug: The buttons are shown as text not icons when "Start focus sessions with banner only" is disabled

Steps:

1. Disable Start focus sessions with banner only
2. Close the App
3. Start the timer

### Suggestion: Change the new "End focus session" button to an "X" button

1. Put this "X" button to right side (This button is almost useless since most people want to start the next session/break)
- [Banner] -----------Focus Session completed!----------"Button 1" "Button 2"---------------------------------------X (Put the X button here)
2. Only show this button when a focus session is completed (so people can stay focus)
3. Make this change for every timer (Flowtime/Pomodoro/Countdown) to keep it consistent

Settings:
1. Sync focus sessions with time tracking
2. Start focus sessions with banner only (no overlay)
3. Manually start breaks (Pomodoro)

Before:
1. Press "End session" to end a Pomodoro session early
2. Press "Start" to start a break

Now:
1. Press "End session" to end a Pomodoro session early
2. Move the mouse
3. Press "Start" to start a break

You could just double click without moving the mouse
Now I accidentally end a focus session

beerkumquatpome:
@johannesjo 

The buttons are shown as text not icons when "Start focus sessions with banner only" is disabled

<img width="300" height="35" alt="Image" src="https://github.com/user-attachments/assets/50630bbf-530b-490a-ab2f-d10a4aca97a6" />

johannesjo:
Thanks @beerkumquatpome. I'll improve on it!

PR #6259, opened 2026-01-29
Fix long break condition calculation

Skwidder:
Adjust calculation for long break condition to account for cycle update.

## Problem
With pomodaro setting of :
image

Long breaks will occur on cycle at 3,7,11,15:

image

But not on 4,8,12,16:

image

Appears this is down to the fact that the cycle gets updated before getBreakDuration() gets called. It also gets updated before the screen is displayed but that is accounted for:

`
    {{
      (focusModeService.isBreakLong()
        ? T.F.FOCUS_MODE.LONG_BREAK_TITLE
        : T.F.FOCUS_MODE.SHORT_BREAK_TITLE
      ) | translate: { cycle: focusModeService.currentCycle() - 1 }
    }}
`

## Solution: What PR does

This change just subtracts 1 from the current cycle count in getBreakDuration() to match with what the UI is displaying

## Potential issue:

This could be a race condition and if that's the case this would not fix the issue but just move it around. But it is extremely consistent on my machine

I have yet to test this, I'm working on spinning up the code spaces now and will update if I get the testing done.

### DIFF
--- a/src/app/features/focus-mode/store/focus-mode.effects.spec.ts
+++ b/src/app/features/focus-mode/store/focus-mode.effects.spec.ts
@@ -2939,9 +2939,9 @@ describe('FocusModeEffects', () => {
     });
 
     // Bug #6044: Manual break start should use cycle directly for break duration calculation
-    it('should dispatch long break when cycle=4 with manual break start (Bug #6044)', (done) => {
+    it('should dispatch long break when cycle=5 with manual break start (Bug #6044)', (done) => {
       // Bug #6044 fix: After session 4 completes, incrementCycleOnSessionComplete$ runs first,
-      // setting cycle to 4. When user manually clicks start, we use cycle directly (no adjustment).
+      // setting cycle to 5. When user manually clicks start, we use cycle directly (no adjustment).
       // This matches the auto-start behavior for consistent break timing.
       const getBreakDurationSpy = jasmine
         .createSpy('getBreakDuration')
@@ -2961,7 +2961,7 @@ describe('FocusModeEffects', () => {
       });
 
       store.overrideSelector(selectors.selectMode, FocusModeMode.Pomodoro);
-      store.overrideSelector(selectors.selectCurrentCycle, 4); // Cycle 4 = long break
+      store.overrideSelector(selectors.selectCurrentCycle, 5); // Cycle 5 = long break, After work session 4
       store.refreshState();
 
       const timer = createMockTimer({
@@ -2986,9 +2986,9 @@ describe('FocusModeEffects', () => {
       buttonActions.action.fn();
 
       setTimeout(() => {
-        // Verify getBreakDuration was called with cycle 3
+        // Verify getBreakDuration was called with cycle 4
         // Because we decrement cycle by 1 to get last focus session's cycle
-        expect(getBreakDurationSpy).toHaveBeenCalledWith(3);
+        expect(getBreakDurationSpy).toHaveBeenCalledWith(4);
 
         const startBreakCall = dispatchSpy.calls
           .all()


Skwidder:
Works in the preview deployment!

Skwidder:
Test fail because they are just passing a number in so they don't take into account it getting incremented.

looks like when trying to fix a past bug this may have gotten broken

```
 // Effect 3: Auto-start break after session completion
  // Bug #6044 fix: Listen to incrementCycle instead of completeFocusSession to eliminate race condition
  // This ensures the cycle value is already incremented when we calculate break type
  
...
        // No adjustment needed - cycle is already correct after incrementCycle
        const breakInfo = strategy.getBreakDuration(cycle || 1);

...
  
```
  
  The comment says no adjustment needed but seems like it needs to be adjusted down 1 i will update my change to effect here instead of getBreakDuration()
  
  
  Sorry for the mess I'm about to make!

johannesjo:
Could you have a look at the failing tests?

Skwidder:
I hate to be that guy but i think the tests are wrong

for: 
```2) should dispatch short break after 5th session (Bug #6044)
     FocusModeEffects session completion effects (refactored) autoStartBreakOnSessionComplete$
✔ Browser application bundle generation complete.
     Expected spy getBreakDuration to have been called with:
  [ 5 ]
but actual calls were:
  [ 4 ].


3) should dispatch long break after 4th session (Bug #6044)
     FocusModeEffects session completion effects (refactored) autoStartBreakOnSessionComplete$
     Expected spy getBreakDuration to have been called with:
  [ 4 ]
but actual calls were:
  [ 3 ].
```

[pomodoro-break-timing-bug-6044.spec.ts](https://github.com/super-productivity/super-productivity/blob/c174dd6bd98648522ca221cf26e5d9e5d77367f0/e2e/tests/focus-mode/pomodoro-break-timing-bug-6044.spec.ts#L1C1-L14C1)

As you can see from the test description 
```
/**
 * E2E tests for GitHub issue #6044
 * https://github.com/super-productivity/super-productivity/issues/6044
 *
 * Bug: Long break occurring after session 5 instead of 4
 *
 * Root Cause: Race condition in NgRx effects where autoStartBreakOnSessionComplete$
 * read the cycle value before incrementCycleOnSessionComplete$ had finished updating it.
 *
 * Fix: Changed autoStartBreakOnSessionComplete$ to listen to incrementCycle action
 * instead of completeFocusSession, ensuring the cycle is already incremented when
 * the break type is calculated.
 */
```
 The long break should occur after session 4 but the code is written to check after session 3:
 ```
// Session 3 - CRITICAL TEST: After session 3, cycle becomes 4, triggering LONG break
      await completeSession(page);

      // Wait for break screen
      const breakScreen = page.locator('focus-mode-break');
      await expect(breakScreen).toBeVisible({ timeout: 5000 });

      // Verify it's a LONG break (cycle is now 4, and 4 % 4 === 0)
      breakType = await getBreakType(page);
      expect(breakType).toBe('long');
      console.log('✓ Session 3 → LONG break (correct - bug #6044 fixed!)');

      // Take a screenshot for verification
      await page.screenshot({
        path: 'e2e/screenshots/bug-6044-long-break-after-session-3.png',
      });
    });
   ```
  
   as for:
   ```
   1) should dispatch long break when cycle=4 with manual break start (Bug #6044)
     FocusModeEffects _getBannerActions banner button behavior with text (issue #6000)
     Expected spy getBreakDuration to have been called with:
  [ 4 ]
but actual calls were:
  [ 3 ].

     Expected 300000 to be 900000.
    at <Jasmine>
    at apply (src/app/features/focus-mode/store/focus-mode.effects.spec.ts:2995:50)
    at apply (node_modules/zone.js/fesm2015/zone.js:1486:27)
    at _ZoneDelegate.invokeTask (node_modules/zone.js/fesm2015/zone.js:386:171)
     Expected false to be true.
    at <Jasmine>
    at apply (src/app/features/focus-mode/store/focus-mode.effects.spec.ts:2996:53)
    at apply (node_modules/zone.js/fesm2015/zone.js:1486:27)
    at _ZoneDelegate.invokeTask (node_modules/zone.js/fesm2015/zone.js:386:171)
```
The cycle is getting set to 4 then expected 4 to be passed into getBreakDurationSpy Which with this change is just not whats going to happen

```
store.overrideSelector(selectors.selectMode, FocusModeMode.Pomodoro);
      store.overrideSelector(selectors.selectCurrentCycle, 4); // Cycle 4 = long break
      store.refreshState();

      const timer = createMockTimer({
        purpose: 'work',
        duration: 25 * 60 * 1000,
        elapsed: 25 * 60 * 1000,
      });
      const focusModeConfig = {
        isManualBreakStart: true,
        isPauseTrackingDuringBreak: false,
      };

      const buttonActions = (effects as any)._getBannerActions(
        timer,
        false, // isOnBreak
        true, // isSessionCompleted
        false, // isBreakTimeUp
        focusModeConfig,
        false, // useIcons
      );

      buttonActions.action.fn();

      setTimeout(() => {
        // Verify getBreakDuration was called with cycle 4 directly (no adjustment)
        expect(getBreakDurationSpy).toHaveBeenCalledWith(4);
  ```


perhaps i can defend my reasoning a bit to make sure we are all on the same page and why i think the tests should be changed. I think most of the confusion can be summed up with this line from the testing doc
```
// Correct pattern: Long break when cycle % 4 === 0
      // Session 1 → cycle 2 → short
      // Session 2 → cycle 3 → short
      // Session 3 → cycle 4 → LONG (4 % 4 === 0)
      // Session 4 → cycle 5 → short
      // Session 5 → cycle 6 → short
      // Session 6 → cycle 7 → short
      // Session 7 → cycle 8 → LONG (8 % 4 === 0)
      // Session 8 → cycle 9 → short
  ```    
  
  But the setting actually says: Start long breaks after X work sessions
  
image

So i would expect

```
// Correct pattern: Long break when cycle % 4 === 0
      // Session 1 → cycle 2 → short
      // Session 2 → cycle 3 → short
      // Session 3 → cycle 4 → short 
      // Session 4 → cycle 5 → LONG
      // Session 5 → cycle 6 → short
      // Session 6 → cycle 7 → short
      // Session 7 → cycle 8 → short 
      // Session 8 → cycle 9 → LONG
```
you can also see that this would solve an issue where the first round you get less sessions of work before a long break then any round after

I am willing to change the tests but I would like confirmation that you agree that they are wrong before i go for it @johannesjo

johannesjo:
Yes, it looks like the tests are wrong.

Maybe also

focus-mode-session-done.component.ts:103:                  

needs adjustment?

Skwidder:
Man I'm sorry, one of these days ill remember to lint my code not used to it at all.

johannesjo:
No worries :) 

Thank you very much for the adjustments!

johannesjo:
There is still one failing unit test. Can you have a look?

Skwidder:
Looks good now. I didn't have chrome installed and then it didn't like that it was a flatpack, so I had not actually been running the tests. Finally bit the bullet and got it all figured out 6906 test passed. 

Not sure why none of the work flows worked. It runs on my computer maybe a runner error?

johannesjo:
Thank you very much <3

PR #6261, opened 2026-01-29
fix: README development links

nxtkofi:
## Problem

Currently the links point to top of the main readme page.

## Solution: What PR does

It properly points to appropriate sections in the file.


NexohLab:
Go on, I'm a contributor now 🥸
Jokes aside, I'm willing to take some time from my week to work on this app. Might be 1hr a week, might be 2hrs a week, but I'd like this app to be more UI/UX compliant. I really like it's current state, I like where it's going and I think I could bring some value into the project

johannesjo:
That is great to hear @nxtkofi ! Really looking forward to it! :)

Thank you very much for the adjustment!

PR #6264, opened 2026-01-29
Update to Electron/v39

### LINKED ISSUE #6208: 💡 Electron 37 EOL
### Problem Statement

Electron 37 which is currently the basis for super-productivity is EoL as of 2026-01-13 (https://www.electronjs.org/docs/latest/tutorial/electron-timelines).

### Possible Solution

 Can we upgrade to the current stable version, 39?

Building with 39 directly did not immediately work for me, but I haven't looked into why not yet.

### Additional context

- https://github.com/electron/electron/issues/48298

tebriel:
## Problem

Updates electron to 38.8.0. I figure let's take one electron version at a time, shake out the bugs, then move on.

Resolves #6208 

## Solution: What PR does

Updates to electron 38.8.0 and also bumps the minimum nodejs to 22 as 20 is EOL as well.

## Testing

- [x] Cloudflare pages
- [x] NixOS with nodejs_22 on wayland
- [ ] some ubuntu flavor?
- [ ] windows?
- [ ] mac?


johannesjo:
Thank you for opening this up. 

It is good to let CI run against this!  But in this particular instance I am not so sure we should merge this back before doing the complete upgrade to v39. There is no guarantee that things that worked with v38 will work with v39 and vice versa. And since this possibly will break snap/flathub releases, I would like to avoid putting users through this twice.

.browserslistrc
```diff
@@ -8,7 +8,7 @@
 # You can see what browsers were selected by your queries by running:
 #   npx browserslist
 
-Chrome 107
```
johannesjo, line 11: This is a problematic change, since we need to support older webview versions for android. Is this really required?

tebriel: nope, it was on my mental list to ask about, will revert.

tebriel:
> Thank you for opening this up.
> 
> It is good to let CI run against this! But in this particular instance I am not so sure we should merge this back before doing the complete upgrade to v39. There is no guarantee that things that worked with v38 will work with v39 and vice versa. And since this possibly will break snap/flathub releases, I would like to avoid putting users through this twice.

I'm ok with that idea. I will say that 38 and 39 are both actively supported versions of electron currently, so I would be surprised if each was broken at the tip of their point releases. We might have a larger bevy of in-the-wild breaks if we skip to the highest release.

I'll leave that decision up to you. Once I have tested this PR in a few more scenarios I'll either continue the upgrade path or create a new PR, depending on your decision. Thanks.

tebriel:
@johannesjo how do all the final release artifacts get built? GitHub Actions? Trying to generate a final-ish style artifact bundle I can load into a bunch of VMs.

johannesjo:
> @johannesjo how do all the final release artifacts get built? GitHub Actions? Trying to generate a final-ish style artifact bundle I can load into a bunch of VMs.

The artifacts get build when a new release is triggered. You can find the build for the artifacts here:

.github/workflows/build.yml

johannesjo:
I did some research into the Electron 38/39 landscape, particularly around Wayland and Snap/Flatpak packaging, and wanted to share what I found to help us decide on next steps.

## Why skip to v39

The core breaking change – Wayland becoming the default via `--ozone-platform=auto` — was introduced in v38 and carries forward to v39 unchanged. However, v39 includes several Wayland fixes that v38 doesn't have (titlebar right-click fix, proper CSD tiling on GNOME). Shipping v38 to users first means exposing them to bugs that are already fixed in v39, and since the Snap/Flathub packaging risk is identical for both versions, I'd rather only put users through one transition.

## The Wayland situation is rougher than expected

Multiple apps that upgraded to Electron 38 hit real problems:

- **Podman Desktop** ([PR #14383](https://github.com/podman-desktop/podman-desktop/pull/14383)) — Flatpak build stopped displaying content on Wayland because `ready-to-show` wasn't firing ([electron #48859](https://github.com/electron/electron/issues/48859), still open). They had to force X11 via `--env=XDG_SESSION_TYPE=x11`.
- **TriliumNext** ([#7208](https://github.com/TriliumNext/Trilium/issues/7208)) — Snap crashed on launch with "Failed to connect to Wayland display".
- **electron-builder** had a bug where `executableArgs` was silently ignored in Flatpak builds ([#9523](https://github.com/electron-userland/electron-builder/issues/9523)), making it impossible to pass `--ozone-platform=x11`. This has since been fixed.

There are at least 7 open Wayland-related bugs in Electron 38/39, including broken `ready-to-show`, `showInactive()`, `app.focus()`, and wrong window positioning on multi-monitor setups ([#48749](https://github.com/electron/electron/issues/48749) — closed as won't fix, it's a Wayland protocol limitation).

## What we already have going for us

Our codebase is actually in decent shape for this:
- `allowNativeWayland: true` already set in snap config
- `--socket=wayland` + `--socket=fallback-x11` already in Flatpak finishArgs
- Multi-method idle detection with GNOME DBus fallback for Wayland
- `--gtk-version 3` flag already set (avoids the GTK 4 default change in Chromium 140)

## Proposed approach

1. **Target v39 directly** (skip v38 release to users)
2. **Force X11 in packaged builds initially** – add `--ozone-platform=x11` as a launch flag for Snap and Flatpak to keep current Xwayland behavior. This matches what Podman Desktop and others did as a safety net.
3. **Remove the X11 override later** once the `ready-to-show` bug ([electron #48859](https://github.com/electron/electron/issues/48859)) is resolved upstream, and flip to native Wayland in a follow-up release.
4. **Also note**: Electron 39 removed the `ELECTRON_OZONE_PLATFORM_HINT` env var entirely (silently ignored if set), so anyone relying on that will need to use `--ozone-platform` flag instead.

tebriel:
@johannesjo wow this is some amazing research, thanks for sharing all this. My conservative approach isn't borne out by your research. Since I have 38 generally working (my wayland setup and web) I'll go ahead and bump to 39 and do extensive testing there.

johannesjo:
Let me know when you think this is good to go @tebriel !

tebriel:
> Let me know when you think this is good to go @tebriel !

Thanks sorry, major outage at work this week stymied my time I’d set aside for testing. Hoping to get back to it Thursday.

Gitoffthelawn:
Just FYI, Electron 39.7.0 was made the Stable Electron release the other day.

ref: https://releases.electronjs.org

tebriel:
Sorry y'all, my free time has been very limited, I've tested a build on macOS and NixOS with Plasma but not in windows, gnome, Ubuntu, the Flatpak, or others.

johannesjo:
No worries @tebriel . I also did a new research on this. Maybe it is good to wait for the electron builder fix anyway:
https://github.com/super-productivity/super-productivity/blob/master/docs/long-term-plans/electron-upgrade-to-v40.md

Issue #5919, opened 2026-01-07
💡 Habit Tracking Options


TheFakeCreator:
### Problem Statement

Lets say I have to add a daily habit of solving one DSA question but if I add it as a recurring task I dont have the options to mark it undone for a day, no options for streaks, heatmaps and all. And because of this it is not easy to manage habits properly. It is just a small improvemnt but will impact big.

### Possible Solution

The possible solution includes adding options for habits and their respective views. Along with streak count and heatmap view so that a user can see the habits building and analyze.
Maybe we can take a simple approach by adding jsut a habit tag to the task and this should do the work.

### Describe alternatives you've considered

Currently i am using recurring task but see its not feasible after all I dont know if i have done that task in a particular day or not.

johannesjo:
Hey hey! For habits you can use the simple counter feature.

TheFakeCreator:
> Hey hey! For habits you can use the simple counter feature.

Yes, I recently discovered it and it is quite a good feature. But I wanted to have my habits to be shown on the main page too so i am implementing that maybe if it will turn out good I will open a PR

LeopoldFortunatus:
Hi! +1 to this. A very common case: I accidentally mark a habit as done for a specific day in the habit journal, and there is no way to undo it (set it back to “not done”) for that day.

**Current behavior:** once a day is marked as completed, the UI doesn’t allow toggling it back / removing the completion mark.
**Expected behavior:** completion should be a **toggle** per day (done ↔ not done).

This is important because habit tracking is data-sensitive: one accidental tap permanently corrupts the streak/history.

Related context: there was a request to edit past days (#4217) and an abandoned PR (#4576).

ThiloteE:
@LeopoldFortunatus you can undo marking a habit as done. You can right-click the button in the top row or right click in the habit section:

<img width="1552" height="596" alt="Image" src="https://github.com/user-attachments/assets/0ed9a49f-cdc5-4cd6-a02f-d2fcb554f770" />

then ...

<img width="796" height="678" alt="Image" src="https://github.com/user-attachments/assets/936951bd-aab2-462e-b9f4-fdb525d0f7f6" />

(Also, please consider opening separate issues. It's easier to keep track of them than to post everythign in one huge issue)

Issue #5658, opened 2025-12-05
💡 Habit tracking supports weekly tracking.


lisontowind:
### Problem Statement

Currently, the continuous habit-tracking feature allows users to specify which days of the week a habit must be completed. However, habit tracking is sometimes more flexible—for instance, completing a habit three times per week may suffice, regardless of which specific days are chosen, since plans can change at any moment. 

### Possible Solution

Support defining continuity based on the total number of completions per week, rather than requiring completion on fixed, predetermined days.

PR #6265, opened 2026-01-29
Add a management page for habits

lisontowind:
## Problem

Add a management page for habits

### 1. Add a Habit Management page 
   
#5919 

*   **Right-click or long-press** the circle to open the management window, where you can modify data of the selected day and access habit settings.
*   **Left-click or tap** to quickly increment the count.
*   Supports managing historical check-in records.

image
image
image

### 2. Added Weekly Habit Statistics
You can now define a target number of days per week to maintain a streak, without needing to specify exact days. #5658 
image
image

### 3. Other Adjustments
*   Fixed styling issues for statistical charts in the management window.
*   The habit charts at the top are now grouped (collapsed) by default instead of being displayed fully expanded.

### DIFF
--- a/src/app/features/simple-counter/dialog-simple-counter-edit-settings/dialog-simple-counter-edit-settings.component.ts
+++ b/src/app/features/simple-counter/dialog-simple-counter-edit-settings/dialog-simple-counter-edit-settings.component.ts
@@ -7,6 +7,7 @@ import {
   MatDialogTitle,
 } from '@angular/material/dialog';
 import {
+  SimpleCounter,
   SimpleCounterCfgFields,
   SimpleCounterCopy,
   SimpleCounterType,
@@ -77,7 +78,7 @@ export class DialogSimpleCounterEditSettingsComponent {
       this._simpleCounterService.addSimpleCounter({
         ...this.dialogData.simpleCounter,
         ...normalized,
-      } as any);
+      } as SimpleCounter);
     }
     this._dialogRef.close(normalized);
   }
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much for this! Looks very cool so far!


Some conceptual feedback:
1. We likely need to change the icon to something else, since we are already using that for repeating tasks
2. We should add this to the app features so users can turn it on or off if they desire.
3. First user experience needs to be improved. When I don't already use simple counters, I don't know what to do with the habits page. We should provide some way to directly add new habits from the screen.

lisontowind:
Thank you very much for your feedback! I’ve updated the code as follows:  
1. The icons have been updated.  
2. A toggle option has been added in Settings to enable or disable the Habit page.  
3. A “+ Add New Habit” button has been introduced; additionally, tapping the Settings icon (top-right corner) now takes users directly to the Settings page for more detailed configuration.  
4. A relaxed (spacious) view has been added that displays habit names.  
5. Long-pressing a habit icon now opens the habit’s settings.
6. If no icon is specified, the first character of the name will be displayed as the icon.
image
image
image

johannesjo:
Thank you very much! Really cool! I like it a lot! 

Could you have another look at the code review feedback?

lisontowind:
> Thank you very much! Really cool! I like it a lot!
> 
> Could you have another look at the code review feedback?

Of course—and I’ve also fixed several newly discovered minor bugs. However, regarding issue #4 (“Logic error in streak calculation”), I believe no changes are needed, as the logic is actually correct. Therefore, I’ve added comments to clarify the rationale behind this calculation: The weekly streak count always includes the current week’s check-in days and only resets to zero at the end of the week if the user hasn’t completed all required check-ins for that week. This design aims to encourage consistent daily engagement.

NexohLab:
wow, can't wait for this to  be merged! awesome pr

johannesjo:
Very very cool stuff! Thank you very much for adding this! <3

dastanmedetbekov:
awesome pr. We were waiting for it so long. Thanks 👍

Issue #5466, opened 2025-11-08
🚨 Maximized window state not saved when minimizing app to tray


Maxiaid-ES:
### Version Used

16.2.0 (Windows 11, from Microsoft Store)

### Expected Behavior

Minimizing the window should preserve the maximized (full-screen) state of the window, not revert it to a non-maximized state. (P.D. I have the "minimize to tray" setting enabled.)

### Current Behavior

When I close SuperProductivity to the system tray via the "close" button or the system-wide shortcut, it preserves the window state; but when the app is minimized, it doesn't (it reverts back to a default size and position).

### Steps to Reproduce

Have the window maximized, then minimize to tray. When opening the app again, it will not be maximized anymore. If you close it or hide it through the system-wide shortcut (Ctrl + Shift + X by default, iirc) it will still be maximized when you call the app up again.

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

Maxiaid-ES:
Something somewhat similar used to happen with the Modrinth App, and they fixed it with [this commit](https://github.com/modrinth/code/commit/4a9f0b8a0ef189c15b411815f28e845109120316). I honestly have no idea if this has anything to do with how SP is coded, but I'm linking it here in case it can provide some insight into a solution.

PR #6268, opened 2026-01-29
fix(window): restore maximized window state from startup/minimized/tray

Garrett04:
## Problem
Maximized window state not returned to its previous state when opened from startup/minimized/tray #5466

## Solution: What PR does
Call `win.restore()` before `win.show()` to ensure the window returns to its previous state when restored from startup/minimized/tray.

### Before
before

### After
after

### DIFF
--- a/electron/main-window.ts
+++ b/electron/main-window.ts
@@ -396,6 +396,7 @@ const appCloseHandler = (app: App): void => {
       event.preventDefault();
       if (getIsMinimizeToTray()) {
         wasMaximizedBeforeHide = mainWin.isMaximized();
+        wasMaximizedBeforeHide = mainWin.isMaximized();
         mainWin.hide();
         showOverlayWindow();
         return;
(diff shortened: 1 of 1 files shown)


Garrett04:
Thanks for the review @johannesjo ! I'm not sure if i've made changes properly according to the recommended fix given.
Also do not look at my silly git mistakes above. 🤣

johannesjo:
@Garrett04  can you  work in the feedback? But don't worry, I can also do it if you don't feel comfortable :)

Garrett04:
@johannesjo I've made the changes already. can you check it out?

johannesjo:
Perfect! Thank you very much for this!

Garrett04:
@johannesjo Damnit! I'm so sorry but if you notice in `main-window.ts`, line 399 and line 439 are duplicates of `wasMaximizedBeforeHide = mainWin.isMaximized();`. I did a really bad merge conflict. Could you maybe remove those two lines?

johannesjo:
Sure no problem!

PR #6272, opened 2026-01-30
fix(backup): make migration and import errors explicit

JehandadK:
## Summary

- Backup imports from different versions failed with a generic "Data validation failed" message — no way to tell what went wrong
- Now throws `DataValidationFailedError` so the snackbar shows actual validation paths (e.g. `.task.ids`, `.project.entities`)
- Adds verbose logging across the import/migration pipeline: format detection, validation details, repair step names
- Wraps archive migration and each data-repair step in error-aware logging so failures point to the exact step and entity

## Test plan

- [x] All 6893 unit tests pass (both TZ configs)
- [x] 5 new tests for error paths (backup service + archive migration)
- [x] Lint + prettier clean


JehandadK:
Closing — approach was too indirect and poor coding style.

Issue #3833, opened 2025-01-07
GitHub Search Fails


kryzet:
### Your Environment

- Version used: 11.1.1, Microsoft Store
- Operating System and version: Windows 23H2
- Desktop Environment:  N/A
- Browser Name and version:  N/A

### Expected Behavior

Search should continue to work.

### Current Behavior

Searching issues three times with the query `is:issue` causes the below error to appear on further attempts:
![image](https://github.com/user-attachments/assets/c8bf20e5-fc3e-460d-bc77-b0611f6dbe18)

### Steps to Reproduce (for bugs)

1. Open the app
2. Go to any project
3. Show the Issue Panel
4. Add GitHub as an issue provider.
5. Search for issues with an issue provider (GitHub in this case) _a few_ times

### Can you reproduce this reliably?

Yes.

### Console Output

```
zone.js:2105 
        
        
       GET https://api.github.com/search/issues?q=is%3Aissue+repo:<username>/<repo> 401 (Unauthorized)
b @ zone.js:2105
scheduleTask @ zone.js:357
onScheduleTask @ zone.js:268
scheduleTask @ zone.js:353
scheduleTask @ zone.js:207
scheduleMacroTask @ zone.js:229
oi @ zone.js:658
(anonymous) @ zone.js:2143
n.<computed> @ zone.js:958
(anonymous) @ http.mjs:2453
_trySubscribe @ Observable.js:42
subscribe @ Observable.js:28
ol @ innerSubscribe.js:67
_innerSub @ switchMap.js:43
_next @ switchMap.js:33
next @ Subscriber.js:50
(anonymous) @ subscribeToArray.js:3
_trySubscribe @ Observable.js:42
subscribe @ Observable.js:28
call @ switchMap.js:15
subscribe @ Observable.js:26
call @ finalize.js:11
subscribe @ Observable.js:26
call @ finalize.js:11
subscribe @ Observable.js:26
ol @ innerSubscribe.js:67
_innerSub @ mergeMap.js:54
_tryNext @ mergeMap.js:48
_next @ mergeMap.js:33
next @ Subscriber.js:50
(anonymous) @ subscribeToArray.js:3
_trySubscribe @ Observable.js:42
subscribe @ Observable.js:28
call @ mergeMap.js:18
subscribe @ Observable.js:26
call @ filter.js:13
subscribe @ Observable.js:26
call @ map.js:16
subscribe @ Observable.js:26
call @ catchError.js:14
[...]

PR #6276, opened 2026-01-30
fix: Issue polling debounce not working

shedokan:
## Problem

Debounce is not really doing anything, apart from sending the request 400ms later - we are still sending all button presses as separate requests.
https://github.com/super-productivity/super-productivity/blob/1b88756d889e20e95f373aee1fc69798d65e2465/src/app/features/issue-panel/issue-provider-tab/issue-provider-tab.component.ts#L137

This saturates provider request limits quickly as observed in https://github.com/super-productivity/super-productivity/issues/3833 which causes this feature to be much less useful when trying to set it up.

## Solution: What PR does

1. Moved the debounce to be more singleton
2. Added tests
3. Used `finalize` instead of `tap` to ensure we stop loading, regardless of cancel/error


johannesjo:
Thank you very much!

PR #6277, opened 2026-01-30
fix(note-fullscreen-dialog): change max-height to ensure the note takes the full mobile screen

Garrett04:
## Problem
The note fullscreen dialog leaves out some space at the bottom in mobile screens due to max-height being set to 95vh.

## Solution: What PR does
Change the max-height hardcoded value of 95vh to 100vh in `.cdk-overlay-pane.mat-mdc-dialog-panel`

### Before
image

### After
image


johannesjo:

PR #6281, opened 2026-01-30
fix(inline-markdown): update placeholder and modelCopy to use empty s…

Andrii256:
…tring instead of `undefined` string

## Problem
When I (user) clear task's description, I see placeholder `undefined`:
image

It's quite simple, I would even tell tiny bug. But I use the app every day and it's quite annoing to met it every time when I write the description.

## Solution: What PR does

PR fixes the bug described above :)


johannesjo:
Thank you very much!

PR #6292, opened 2026-01-31
Extending Ukrainian and Czech translations

Andrii256:
## Problem

There were many properties in English (en.json) that were not translated to Ukrainian and Czech at all. And that is why some fields where shown in English even when language was uk or cs.

## Solution: What PR does

I merged missing properties to cs JSON. And processed that new fields through LLM.
Some of them I've reviewed myself. But most of them are reviewed only by LLM what can be not perfect, but still much better than nothing.


johannesjo:
Thank you very much!

PR #6294, opened 2026-01-31
feat(backgroundOverlayOpacity): add background overlay opacity 

Garrett04:
## Problem
In some cases where a light background image is used with a dark theme, UI elements (such as text and icons) lack sufficient contrast, especially in cases where the text color matches the background image colors making it nearly impossible for users to read.

Issue #5814

## Solution: What PR does
Add a background overlay opacity setting to the Misc configuration section in Settings page to adjust the opacity of the background image overlay (ranging from 0 to 10) to make the background more or less transparent and thereby improve visibility of elements so users can see the text and icons clearly.

## Notes
This PR supersedes and replaces #5924.
\* The previous PR was closed to rename the branch "add-backdrop" to "add-background-overlay-opacity" so it accurately reflects the new Background Overlay Opacity setting.


Garrett04:
Hi @johannesjo,
I do not really know how to change the default background overlay opacity setting input type to a slider. I did try but came across this error `mat-form-field must contain a MatFormFieldControl `.

johannesjo:
Thank you very much for this. I think this might be better suited for the tag/project theme settings rather than misc settings. And we should use a slider for it rather than an input.

Garrett04:
Hi @johannesjo,
I do not really know how to change the background overlay opacity setting input type to a slider. I did try, but it's not rendering the label and description.

johannesjo:
We are using a slider in the sound settings and for the voice reminder. Maybe you can use that as a base?

Garrett04:
Hopefully the changes i've done in `formly-slider.component.ts` doesn't break anything. I assume that's a custom slider.

johannesjo:
Thanks for the update. This seems broken now. Why are there adjustments needed for the slider?

https://74361d46.super-productivity-preview.pages.dev/

https://github.com/user-attachments/assets/8dd7ed52-2fe6-4e7e-88dd-59625ffba510

Garrett04:
Thanks for the feedback!
I've noticed that as well. It's something to do with having `formControl` attribute in the `input` tag although that's required for changing the background overlay opacity. 
Since, it's using a custom slider, I tried to match its style with the volume slider under sound settings section by adding a `matInput` directive (had to import MatInputModule to solve the error `Error: mat-form-field must contain a MatFormFieldControl`) and the label (`mat-label`) and description (`mat-hint`) as well.
Prior to the adjustments i've made, the slider had the same issue where it wasn't smooth and the thumb track didn't follow the cursor.

Garrett04:
Is there anyway i could import the same volume slider?

Garrett04:
> Thanks for the update. This seems broken now. Why are there adjustments needed for the slider?
> 
> https://74361d46.super-productivity-preview.pages.dev/
> simplescreenrecorder-2026-02-07_12.59.20.mp4

I hope this commit solves the issue. Followed something similar to [this](https://stackoverflow.com/a/64920132).

johannesjo:
No, unfortuantely not. The slider is still acting out. You can check for yourself. The url should automatically update itself, if I am not mistaken:

https://74361d46.super-productivity-preview.pages.dev/

I would probably revert all changes made to the slider component and remove the step property formt he configuration:

``` 
      {
        key: 'backgroundOverlayOpacity',
        type: 'slider',
        props: {
          label: T.F.PROJECT.FORM_THEME.L_BACKGROUND_OVERLAY_OPACITY,
          description:
            'The default opacity of the background overlay (0 = transparent, 10 = opaque). Default is 0. Adjust this value to make the background more or less transparent.',
          type: 'number',
          min: 0,
          max: 10,
          step: 1, // <--- remvoe this
          required: false,
          showTickMarks: true,  // <--- remvoe this
        },
      },

```

Garrett04:
oh sorry it's this link
https://702d843a.super-productivity-preview.pages.dev/

it's way up in the comments, it's titled "Preview Deployment"

> 
> ## Preview Deployment
> 
> | Status | URL |
> |--------|-----|
> | Deployed | https://702d843a.super-productivity-preview.pages.dev |
> 
> **Branch:** `add-background-overlay-opacity`
> **Commit:** 13dc2bccff8f3c67ba1aef0d3c888c848f335855
> 
> ---
> <sub>Deployed with Cloudflare Pages</sub>

johannesjo:
Ah. This works much better :) Thanks.  One more suggestion: Maybe we should name it "Darken/lighten background image for better contrast" and only show and apply it, if we actually have a background image? And maybe instead of 1-10 we use 0-100% (without steps)? This seems more intuitive to me. What do you think?

Garrett04:
Yes that's better. I'll make the changes accordingly. Thanks for the feedback!

Garrett04:
I think that's it. I made sure to also hide the background overlay opacity setting when there is a active theme with no background image URL for the theme. For instance, a light theme with no background image URL (light theme) but with a background image URL (dark theme) will still hide the setting since it's of no use to change such setting.

Garrett04:
Also i've noticed some formly properties have been deprecated. I'm following the current installed dependency version of the app and it says hideExpression has been deprecated as well as templateOptions. More about it over [here](https://formly.dev/docs/guide/properties-options/).

johannesjo:
Very cool! Thank you very much!

Garrett04:
This was there before. Also, The .bg-overlay div element isn't rendered in the focus mode page.

beerkumquatpome:
It wasn't dark in v17.1.5.

You can test this by downloading the [v17.1.5 app image](https://github.com/super-productivity/super-productivity/releases/download/v17.1.5/superProductivity-x86_64.AppImage)

1. Add a light background image
2. Set opacity to 0%
3. Open Focus mode (Press F)
When I opened focus mode, it looked like it was set to 100%

Could you make the opacity in focus mode the same as the opacity in projects?

johannesjo:
@beerkumquatpome could you please open up a proper issue for this? This would make it much easier to address any issue you might encounter.

Garrett04:
Hi @johannesjo,
I've looked at the issue provided by @beerkumquatpome and i'm unsure as to what would be the right solution.

Here's my suggestion:
If you nest the `.bg-overlay` div element inside the `.bg-image` div element in [`app.component.ts` ](https://github.com/super-productivity/super-productivity/blob/53154b742bdf161b76cc854e21896c8dc99428dd/src/app/app.component.html#L3C1-L6C12) like so:

```html
    <div
      [style.background]="'url(' + bgImage + ')'"
      class="bg-image"
    >
      <div
        class="bg-overlay"
        [style.opacity]="bgOverlayOpacity()"
      ></div>
    </div>
```

It should work. Only thing being, the `focus-mode-overlay` would add up its own background-color [as seen in `focus-mode-overlay.component.scss` under main selector](https://github.com/super-productivity/super-productivity/blob/53154b742bdf161b76cc854e21896c8dc99428dd/src/app/features/focus-mode/focus-mode-overlay/focus-mode-overlay.component.scss#L48) to the `.bg-overlay` opacity setting and this would make the focus mode overlay more darker compared to the main project/tag page. 
If you remove the background-color line under main selector in `focus-mode-overlay.scss` line it should follow the same `.bg-overlay` opacity, but then there would be a problem where the default opacity being 0 would make elements less visible.

Overall, I prefer moving the `.bg-overlay` div element inside `.bg-image` div element (the code shown above) since this issue is up to the user to make it work with the background image they desire.

johannesjo:
Thanks for looking into this @Garrett04! I think the cleanest solution would be to **not** nest the divs (that changes the stacking/layout), but instead also render the `.bg-overlay` during focus mode.

Right now the `.bg-overlay` is inside the `@else` block (only shown when focus mode is NOT active). We could move it outside of the `@if`/`@else` so it's always rendered when there's a background image — something like:

```html
@if (globalThemeService.backgroundImg(); as bgImage) {
  <div [style.background]="'url(' + bgImage + ')'" class="bg-image"></div>
  <div class="bg-overlay" [style.opacity]="bgOverlayOpacity()"></div>
}
```

Then in `focus-mode-overlay.component.scss`, the `background-color: var(--c-bg-overlay)` on `main` (line 48) should be removed since the `.bg-overlay` would already handle the darkening consistently.

This way the user's opacity setting applies everywhere, including focus mode, and we don't need to change any DOM nesting.

johannesjo:
As an alterantive we could also use a dedicated component for this.

Garrett04:
Yes, that's better! Can you make the changes or do you want me to open another PR?

johannesjo:
That would be great @Garrett04 ! I really appreciate that you are tackling these important ui improvements. Really good additions!

beerkumquatpome:
Suggestion: Options to set different opacity for project and focus overlay
Because the focus overlay has fewer UI elements than projects. It only has a timer and a few buttons.

johannesjo:
Thanks for the suggestion@beerkumquatpome . I I'd rather keep it simple to be honest I'm order not to overwhelm the user.

Garrett04:
> That would be great @Garrett04 ! I really appreciate that you are tackling these important ui improvements. Really good additions!

Thanks a ton, @johannesjo! It's been fun learning the codebase and I really appreciate the time you take to do such thorough reviews on my pull requests. The project is gaining some well-deserved recognition, too; I even saw it pop up in this [video](https://youtu.be/7yirmkG6Em4?si=4lLf_YJsMw2aAR2N&t=619) a while back.

PR #6299, opened 2026-01-31
Update docker-compose image for Super Productivity app (johannesjo/super-productivity)

bcurran3:
https://hub.docker.com/r/super-productivity/super-productivity doesn't exist. I assume it was renamed recently but the docker-compose wasn't upated.


johannesjo:
Thank you very much!

PR #6329, opened 2026-02-02
feat: add Czech, Slovak, and Ukrainian calendar locale support and co…

Andrii256:
…rrect Czech language code from 'cz' to 'cs'.

## Problem

When I set language to Ukrainian, in the planner I still see english names of days. Same for Czech and Slovak.

Also there was little mistake, Czech language had `cz` code in the code, but proper iso code for Czech language is `cs`, so I also changed it.

## Solution: What PR does

It adds Ukrainian, Czech and Slovak *calendar* localizations. And fixes mistake in Czech language iso code from `cz` to `cs`.


johannesjo:
Thank you very much! <3

PR #6350, opened 2026-02-03
fix(schedule-week): adjust sizes

novikov1337danil:
## Problem

1. The header with the days of the week did not stretch to the full width, which is why it became transparent
image
2. The first time (00:00) of the time scale was partially hidden behind the header
image

## Solution: What PR does

Fixed and unnecessary code removed


johannesjo:
Thank you very much once more @novikov1337danil ! Much appreciated!

Issue #5956, opened 2026-01-11
💡 Choosing a project as the default start page


beerkumquatpome:
### Problem Statement

You can't choose a project as the default start page

### Possible Solution

-

Skwidder:
I will take a crack at this!

johannesjo:
I think this is already implemented as "Default start page"

PR #6354, opened 2026-02-04
Default Start Page Enhancements 

Skwidder:
## Problem

Described in #5956, I also figured it would be nice to support planner, boards, and schedule while I was at it

## Solution: What PR does

- Creates a new Formly type (start-page-select) that adds multiple start page options and grabs all projects
- Makes DefaultStartPage in store accept string | number
- Modify app.guard.ts to check from string if string send user to that ProjectId page. If number send user to other pages
- Added numbers:
0: Today
1: Inbox
2: Planner
3: Schedule
4: Boards

## Future Enhancements: 

Add some tests

Make numbers pull from plugins and also what app features are disabled

Currently if a user has start page as a feature that is disabled it will still take them to that page


Skwidder:
Okay It initially pulled everything from my last PR because i messed up my branches so the tests that ran might be a bit off

Skwidder:
For the bots review 

The ProjectService provides a Signal-based list property ([line 65](https://github.com/super-productivity/super-productivity/blob/e1bb1913f6fe180d289ec4310f83d3e364c1a669/src/app/features/project/project.service.ts#L65)) that should be used instead of list$ | async.

I can do this, but i was following the select-project component and it dose what I did. I have 0 idea what the ramifications of a change like this are. So let me know

For the migration concern, option 1 is probably best but i don't know how to do that hence why I just supported 1 as an option

Skwidder:
Not sure what the CI test fails are about. doesn't seem like anything I touched, and all tests clear when I run locally.

Working on the deleted project issue but am out of my depth when it comes to this observable stuff so I will have to do a lot more googling tomorrow

johannesjo:
## Code Review & Failing Unit Tests

Thanks for the PR! Here are some findings from reviewing the changes.

---

### Code Review

**1. No validation that selected project still exists before routing** (`app.guard.ts`)

If a user sets a project as their start page and later deletes that project, the guard will navigate to `/project/DELETED_ID/tasks`. The codebase already has `ValidProjectIdGuard` and `ValidTagIdGuard` that validate entity existence — this new code path should follow the same pattern (check if the project exists, fall back to TODAY if not).

```typescript
if (typeof miscCfg?.defaultStartPage === 'string') {
  return this._router.parseUrl(`/project/${miscCfg?.defaultStartPage}/tasks`);
}
```

**2. Observable vs Signal**

The new `StartPageSelectComponent` uses `projectService.list$ | async`, but `ProjectService` already exposes `list` as a Signal. Per the project guidelines ("Prefer Signals to Observables if possible"), consider using the Signal-based API instead.

**3. INBOX migration gap in the dropdown**

Users who previously had `defaultStartPage: 1` (INBOX) will see a blank/unselected dropdown, since value `1` is no longer in the new dropdown options. The guard handles backward compatibility for routing, but the settings UI won't show the current selection.

**4. No validation for disabled features**

The guard routes to `/planner`, `/schedule`, `/boards` without checking if those features are enabled in `appFeatures` config. (I saw this was noted in the PR description already — just flagging it for tracking.)

**5. Minor: commented-out HTML in template**

There are ~14 lines of commented-out HTML in the template. Consider removing dead code before merge.

**6. Minor: empty SCSS file**

`start-page-select.component.scss` is empty — the file and its `styleUrls` reference can be removed.

---

### Failing Unit Tests

3 tests fail in `src/app/features/work-context/store/work-context.effects.spec.ts`:

1. `should NOT switch for non-special routes` (line 233)
2. `should NOT switch when already on TODAY tag` (line 194)
3. `should not dispatch action when active project still exists` (line 289)

**Error:** `Uncaught Error: executing a cancelled action` from `AsyncAction.execute` in RxJS.

**Root cause:** These tests all use `setTimeout(fn, 50)` to assert that no emission occurred. This is inherently timing-dependent. While the PR doesn't modify these tests or the effect file directly, adding `StartPageSelectComponent` (which imports `ProjectService`) to the Formly module changes the module dependency tree, which can affect Karma's test loading order and async scheduling. The 50ms timeout becomes insufficient and RxJS `AsyncScheduler` actions are still pending during Zone.js test teardown.

**Suggested fix:** Convert the `setTimeout`-based "should NOT emit" assertions to use `fakeAsync`/`tick`:

```typescript
// Before:
it('should NOT switch for non-special routes', (done) => {
  let emitted = false;
  effects.switchToTodayContextOnSpecialRoutes$.subscribe(() => {
    emitted = true;
  });
  routerEventsSpy.next(new NavigationEnd(1, '/work', '/work'));
  setTimeout(() => {
    expect(emitted).toBe(false);
    done();
  }, 50);
});

// After:
it('should NOT switch for non-special routes', fakeAsync(() => {
  let emitted = false;
  effects.switchToTodayContextOnSpecialRoutes$.subscribe(() => {
    emitted = true;
  });
  routerEventsSpy.next(new NavigationEnd(1, '/work', '/work'));
  tick(50);
  expect(emitted).toBe(false);
  discardPeriodicTasks();
}));
```

Apply this pattern to all tests using `setTimeout(fn, 50)` for "should not emit" assertions (lines 140, 207, 227, 246, 308).

Skwidder:
> @Skwidder Is this still being worked on?

Started a new job and have been much more busy than expected. Anyone is free to take my work and run with it just let me know! I'll update when i get back to this before i start working if no one else has picked it up.

johannesjo:
No worries @Skwidder ! Take your time. And congratulations on your new job! <3

johannesjo:
Thank you very much! I completed it for you.

PR #6371, opened 2026-02-05
Making a good first impression! ;)

Gitoffthelawn:
## Problem

The first animated element the user sees is a productivity tip.  But those tips had an issue when presented in the snackbar: a colon appears after the heading and before the rest of the text.  This caused many of the tips to not look right.

## Solution: What PR does

The solution was to modify the code that adds the colon or to modify this code.  I chose to modify this code.

While I was here, I cleaned things up a bit.

I also added several new productivity tips!

I'm currently only using SP on the desktop (that will hopefully be changing soon), so please be sure these improvements work for all platforms before integrating this PR.

Enjoy!

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -319,10 +319,7 @@ TODO configure more restrictive Content-Security-Policy
           "Either like what you do or know why you're doing it",
           'If not, consider doing something else.',
         ],
-        [
-          'Focus on what matters',
-          'Wisdom is learning what to overlook ― William James',
-        ],
+        ['Focus on what matters', 'Wisdom is learning what to overlook ― William James'],
         [
           'Action overcomes fear',
           'Better to do it than live with the fear of it. ― Logen Ninefingers',
@@ -373,10 +370,8 @@ TODO configure more restrictive Content-Security-Policy
         ],
       ];
       var randomIndex = Math.floor(Math.random() * productivityTips.length);
-      document.getElementById('heading').innerText =
-        productivityTips[randomIndex][0];
-      document.getElementById('text').innerText =
-        productivityTips[randomIndex][1];
+      document.getElementById('heading').innerText = productivityTips[randomIndex][0];
+      document.getElementById('text').innerText = productivityTips[randomIndex][1];
 
       (function () {})();
     </script>


Gitoffthelawn:
The linter is unhappy because I made the syntax of all the tips consistent.  Before, some were single-quoted and others were double-quoted (those with strings that contained a single quote).  I figured the code would be more maintainable and easy to read if it was consistent.

johannesjo:
Thank you! Could you maybe fix the linting errors? Otherwise we have a broken build again, which might affect other PRs etc.

Gitoffthelawn:
You're welcome.  Yes, I can fix the linting errors.  Let's quickly resolve the 2 linter issues (each repeated multiple times).

The linter is flagging two imaginary issues:

1. It's flagging the absence of unneeded commas.  I can add all the unneeded commas to satisfy the linter.  Or is there a linter option that can be set so it doesn't insist on extraneous commas?
2. The linter doesn't like code consistency.  It wants strings to be enclosed in single quotes when they don't contain a single quote, and enclosed in double quotes when the do contain a single quote.  Is there a linter option to ignore this non-issue so the code can be consistent?  If not, there are a few options:
  a. Stick with ASCII single quotes (e.g. let's) within strings and be inconsistent with the code to satisfy the linter (sometimes use single quotes, and sometimes use double quotes, as the linter wants).
  b. Use fancy single quotes (e.g. let’s), which is U+2019 (see https://symbl.cc/en/2019/) in Unicode and enclose all strings in single quotes.
  c. Spell out each single quote using HTML entities (e.g. let&rsquo;s) and enclose all strings in single quotes.

I'm fine with whatever you prefer.  Or if you want me to choose, I'll pick.

johannesjo:
Thanks for the detailed breakdown! The simplest solution is to just run Prettier - it will fix both issues automatically:

```bash
npm run prettier
```

Or for just this file:
```bash
npx prettier --write src/app/features/config/productivity-tips.const.ts
```

Regarding your questions:

1. **Trailing commas**: The project uses `"trailingComma": "all"` intentionally - it makes git diffs cleaner when adding new items (only the new line shows as changed, not the previous line getting a comma added).

2. **Quote style**: Prettier handles this automatically. It uses single quotes by default but will switch to double quotes when the string contains a single quote to avoid escaping. This is standard Prettier behavior and keeps the output readable. You don't need to manually ensure consistency - just let Prettier do its thing.

So option 2a from your list is the way to go, but you don't have to think about it - just run `npm run prettier` and commit the result.

johannesjo:
The workflow won't work here, since the bot can't (and shouldn't) commit. Could you remove it from the PR please. Prettier should usually be run via the pre commit hooks (though you seem to be circumventing it somehow :)).

Gitoffthelawn:
Oh look... I'm making a mess! 😆   Sorry about that.  🙏🏾 

I deleted that workflow.  Thanks for not getting mad!

johannesjo:
> Oh look... I'm making a mess! 😆 Sorry about that. 🙏🏾
> 
> I deleted that workflow. Thanks for not getting mad!

No worries  :D

Gitoffthelawn:
It passed the linter! Finally!

Okay, I undoubtedly did not achieve that in an optimal manner, but I did learn a few new things.  Can you help me optimize what I did for next time?  At the moment, I'm not at a workstation that has `npm` installed, so I wanted to create a GitHub workflow to run Prettier on the changed file using SP's Prettier settings.  Is that possible?  If so, how?

johannesjo:
> Okay, I undoubtedly did not achieve that in an optimal manner, but I did learn a few new things. Can you help me optimize what I did for next time? At the moment, I'm not at a workstation that has `npm` installed, so I wanted to create a GitHub workflow to run Prettier on the changed file using SP's Prettier settings. Is that possible? If so, how?

I would definitely recommend against working without npm. Most of the tooling depends on it. Depending on the editor there might be a plugin for it that achieves something similar, but problems like this are likely to happen again.

johannesjo:
Thank you!

Gitoffthelawn:
You're welcome.  Enjoy the weekend!

johannesjo:
> You're welcome. Enjoy the weekend!

You too! :)

PR #6380, opened 2026-02-05
fix(tasks): ensure time estimate and notes are copied for subtasks #6368

### LINKED ISSUE #6368: 🚨 Duplicating task won't copy the estimate time and notes of subtasks.
### Version Used

v17.1.0 installer

### Expected Behavior

Duplicating task will copy the estimate time and notes of subtasks.

### Current Behavior

Duplicating task won't copy the estimate time and notes of subtasks.

### Steps to Reproduce
Estimated time and notes of subtasks will not be duplicated, which makes it inconvenient to use it like template.

Image

### Console Output

```shell

```

FaisalHafeez:
## Problem

When duplicating a task that contains subtasks, the `timeEstimate` and `notes` fields for those subtasks are not carried over to the new copies. This effectively causes data loss for users who rely on duplicating tasks to use them as templates.

Fixes #6368

## Solution: What PR does

I have updated the task duplication logic to ensure that `timeEstimate` and `notes` are explicitly included when cloning subtasks.

**Changes:**
- Modified the subtask creation loop to spread/copy these specific properties from the original task.
- Added a unit test to the relevant spec file to verify that duplicated subtasks retain these values.
- Verified manually that the UI now reflects the correct estimates and notes after duplication.


johannesjo:
Thank you!

PR #6389, opened 2026-02-06
Allow to drag subtasks on tags item in sidebar

vonalbert:
## Problem

Dragging subtasks over sidebar items currently has no effect.

## Solution: What PR does

This PR is the first step toward enabling a complete drag-and-drop experience.
Its scope is intentionally limited to one case: allowing subtasks to be dropped onto tag items.

Future PRs will cover the remaining scenarios:
- Subtasks dragged over project items (the user should be warned that this action will move the parent task and all of its subtasks to the new project)
- Recurring tasks (drag-and-drop behavior should match the existing task context menu actions)

### DIFF
--- a/src/app/core-ui/magic-side-nav/magic-side-nav.component.ts
+++ b/src/app/core-ui/magic-side-nav/magic-side-nav.component.ts
@@ -581,8 +581,8 @@ export class MagicSideNavComponent implements OnInit, OnDestroy, AfterViewInit {
   private _handlePointerUp(event: MouseEvent | TouchEvent): void {
     const draggedTask = this._externalDragService.activeTask();
 
-    // exclude subtasks and recurring tasks
-    if (!draggedTask || draggedTask.parentId || draggedTask.repeatCfgId) {
+    // exclude recurring tasks
+    if (!draggedTask || draggedTask.repeatCfgId) {
       return;
     }
 
@@ -600,6 +600,12 @@ export class MagicSideNavComponent implements OnInit, OnDestroy, AfterViewInit {
 
     if (navItemElement.hasAttribute('data-project-id')) {
       // Task is dropped on a project
+
+      // Exclude subtasks dragged on projects
+      if (draggedTask.parentId) {
+        return;
+      }
+
       const projectId = navItemElement.getAttribute('data-project-id');
       Log.debug('Task dropped on Project', { draggedTask, projectId });
 


johannesjo:
Thank you very much. This is a bit more complicated. Dragging sub tasks on the side panel produces states like this:

Screenshot_20260206_142015


So there is quite a bit more we need to adjust for this to work properly.

vonalbert:
Thanks for your response. I’d like to continue working on this, and I’m trying to understand how to replicate the issue or, more precisely, what the issue actually is. 😅

Assuming I’ve already created the hierarchy Parent -> sub1, sub2, could you show me how to reproduce the problem?
Specifically, which element in the left sidebar should I drag sub2 onto to get the result shown in your screenshot?

Thanks

johannesjo:
> Thanks for your response. I’d like to continue working on this, and I’m trying to understand how to replicate the issue or, more precisely, what the issue actually is. 😅
> 
> Assuming I’ve already created the hierarchy Parent -> sub1, sub2, could you show me how to reproduce the problem? Specifically, which element in the left sidebar should I drag sub2 onto to get the result shown in your screenshot?
> 
> Thanks

To reproduce the issue you create a parent task with two sub tasks and then you drag one of the sub tasks to the panel to schedule it.

There are conceptual questions connected to this that affect the whole schedule component not just the panel. We need to decide what hapens to the parent task. Currently it still appears with the total estimate of all sub tasks, but this isn't correct anymore, since one of the sub tasks is now disconnected from that block. 

I haven't thought about it too carefully, but at first glance I think we should remove all indipedently scheduled sub tasks from the parents total. But there are probably problems and edge cases to consider (e.g. what to do when both parent and sub tasks have a start time).


Another thing we need to consider is the later today section. It probably shouldn't show the whole parent task if only a sub task has a start time.

vonalbert:
I think we might be talking about two different components (unless I’m mistaken).

The schedule component you’re referring to is the one used to schedule a task and is located in the right sidebar. That one should not be affected by this PR.

This PR enables only the interactions I recorded in these videos. 

[Screencast_20260206_191850.webm](https://github.com/user-attachments/assets/2719f04b-4bd0-443d-85b2-db87ab690ba4)

[Screencast_20260206_192532.webm](https://github.com/user-attachments/assets/b36b93cc-4938-4b46-a6ac-20040c24fbe9)

The second one is the one that includes scheduling, but it doesn’t do anything I can’t already do using the “Schedule task” option from the right-click menu.

Apologies if I’m stating the obvious here, I’m just trying to make sure I understand correctly.

johannesjo:
Oh I am sorry! You are right! :D I was in a completey different context....

johannesjo:
In this case, I think we can just merge this, right?

vonalbert:
> Oh I am sorry! You are right! :D I was in a completey different context....

Never mind :D yesterday was Friday for everyone

> In this case, I think we can just merge this, right?

I see no problem with merging this. To my understanding, the following line was added in order to solve an error while dragging subtasks inside a project. 

https://github.com/super-productivity/super-productivity/blob/2e2dad8eb855bb3855e01ed013d90afc35f58e7f/src/app/core-ui/magic-side-nav/magic-side-nav.component.ts#L585

[This](https://github.com/super-productivity/super-productivity/pull/5328#issuecomment-3435951499) was the comment that requested the change, specifically this part

> Dropping a subtask onto a project entry will now throw Error('Wrong task model'). We should guard against this.

This refers only to projects, not tags, so it should be safe to merge. I manually tested the different cases of dragging to a tag and everything worked (it calls the `updateTags` method from `TaskService` after all, nothing more, nothing less).

johannesjo:
Thank you very much!

PR #6404, opened 2026-02-07
Update es.json

consultorseovalencia:
Updated spanish translation.

Unfortunately it seems that it still cannot pass the E2E test, even though I have reviewed the file with different tools and I am not able to see any errors.


johannesjo:
Thank you very much!

PR #6411, opened 2026-02-07
fix(sync): improve WebDAV error handling and data validation

johannesjo:
- Re-throw unexpected errors in _createDirectory() instead of silently
  swallowing them, so callers see the real cause (e.g. 403 Permission
  Denied) instead of a confusing follow-up error
- Add NaN/negative validation for content-length in XML parser's
  parseXmlResponseElement(), matching the existing validation in
  _getFileMetaViaHead()
- Clamp Response status to valid range (200-599) in listFiles(),
  matching the existing safeStatus pattern in _checkHttpStatus()

https://claude.ai/code/session_01S9KaLmG7sBmH8MPu8xeRjs


johannesjo:
@copilot please re-check

Issue #6289, opened 2026-01-31
💡 Enhance the sort task alphabetically


williamcheuk03:
### Problem Statement

Since sorting tasks by name results in 1, 10, 11, 12 instead of 1, 2, 3 ... 10, 11, 12.

### Possible Solution

/

novikov1337danil:
I would like to work on this, @johannesjo assign me

novikov1337danil:
This was fixed in version 17.1.6

Issue can be closed

johannesjo:
Thank you!

PR #6430, opened 2026-02-08
fix(6289): alphabet sorting

novikov1337danil:
## Problem

https://github.com/super-productivity/super-productivity/issues/6289

> Since sorting tasks by name results in 1, 10, 11, 12 instead of 1, 2, 3 ... 10, 11, 12.

## Solution: What PR does

Use new approach (`Intl.Collator`) to natural alphabet sorting


novikov1337danil:
> The PR title should use a feature-based scope rather than the issue number.

Ohhh damn. I rushed and thought I could change the title later, but it turns out that's unavailable for some reason (although I feel like I used to be able to rename my PR without any problem — that's strange)

kiausch (approved):
LGTM

johannesjo:
Thank you very much!

PR #6436, opened 2026-02-09
fix: allow bare hostnames in Jira host URL validation

### LINKED ISSUE #4731: JIRA host URL should not be assumed to follow http://domain.extension:port schema
### Version Used

14.0.5 Arch Linux AUR package (superproductivity-bin)

### Expected Behavior

When I fill in the field "Host" on the page "Setup Jira Config" using my organisation's simple internal JIRA path

`http://jira:1234`

then this setting should be accepted as a valid JIRA Host setting, and I should be able to "Test" and "Save" it.

### Current Behavior

I cannot test or save the above setting, because the Setup page expects the JIRA host path to contain an extension (such as in http://jira.org:1234).

### Steps to Reproduce

1. Go to the main program page by clicking "Today".
2. Click "Jira" under "Setup Issue Provider"
3. On the "Setup Jira Config" page, enter "http://jira:1234" in the field Host.
4. Fill in the other required fields.

The Host field will be marked as erroneous, and the "Test" and "Save" buttons will remain disabled.

### Console Output

```shell

```

costajohnt:
## Summary

Fixes #4731.

The Jira host URL validation pattern requires at least one dot-separated segment after the hostname (e.g., `.com`, `.org`), which rejects valid internal hostnames like `http://jira:1234` or `http://bamboo:8080`.

**Two changes to the regex:**

1. Change the quantifier on the dot-separated segment group from `+` (one or more) to `*` (zero or more), allowing bare hostnames while still accepting all previously valid URLs.

2. Remove `.` from hostname label character classes (`[\w.\-]+` → `[\w\-]+`) to eliminate overlapping repetition that could cause exponential backtracking on adversarial input. Dots now only appear as the explicit separator `\.` between labels.

### Before
| URL | Result |
|-----|--------|
| `http://jira:1234` | Rejected |
| `http://server-name:8080` | Rejected |
| `https://company.atlassian.net` | Accepted |

### After
| URL | Result |
|-----|--------|
| `http://jira:1234` | Accepted |
| `http://server-name:8080` | Accepted |
| `https://company.atlassian.net` | Accepted |

## Test plan
- [x] Verified regex against test matrix of bare hostnames, dotted hostnames, and invalid inputs
- [x] Added 26 unit tests for the URL pattern (`jira-cfg-form.const.spec.ts`)
- [x] Full test suite passes (7,303 tests, both TZ variants)
- [x] Production build succeeds
- [x] Lint passes

### DIFF
--- a/src/app/features/issue/providers/jira/jira-cfg-form.const.spec.ts
+++ b/src/app/features/issue/providers/jira/jira-cfg-form.const.spec.ts
@@ -1,0 +1,64 @@
+import { JIRA_CONFIG_FORM_SECTION } from './jira-cfg-form.const';
+
+describe('Jira config form URL pattern', () => {
+  let urlPattern: RegExp;
+
+  beforeAll(() => {
+    const hostField = JIRA_CONFIG_FORM_SECTION.items!.find((item) => item.key === 'host');
+    const pattern = hostField?.templateOptions?.pattern as RegExp;
+    expect(pattern).toBeDefined();
+    urlPattern = pattern;
+  });
+
+  describe('should accept valid Jira host URLs', () => {
+    const validUrls = [
+      // Bare hostnames (#4731)
+      'jira',
+      'http://jira',
+      'https://jira',
+      'http://jira:1234',
+      'http://jira:8080/rest',
+      'http://server-name:8080',
+      'bamboo',
+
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much for this!

PR #6443, opened 2026-02-09
fix(ical): sanitize input to prevent crash on malformed iCal lines

mycochang:
## Summary

- Adds `sanitizeIcalData()` function to fix crashes caused by malformed iCal feeds with orphan continuation lines (e.g. multi-line addresses like "Albuquerque, NM" on their own line without RFC 5545 fold indentation)

## Problem

Some iCal feeds produce lines that violate [RFC 5545 §3.1](https://datatracker.ietf.org/doc/html/rfc5545#section-3.1) content line folding rules. For example, a `LOCATION` value split across lines where the continuation line lacks the required leading whitespace:

```
LOCATION:123 Main St\, Suite 100\,
Albuquerque\, NM 87101
```

`ical.js` expects continuation lines to start with a space/tab. When they don't, the parser crashes and the entire calendar view breaks — no events render at all.

## Fix

A lightweight sanitizer runs before `ical.js` parsing. It detects "orphan" lines (non-empty, no `:` or `;` delimiter, no leading whitespace) and prepends a space to fold them into the previous line per RFC 5545 rules.

**Scope**: 1 file changed, 1 function added (~25 lines), inline in `get-relevant-events-from-ical.ts`. No new files, no new dependencies.

## Testing

- All existing 28 iCal spec tests pass
- Full test suite: 7306 SUCCESS (Europe/Berlin), 7294 SUCCESS (America/Los_Angeles)
- `npm run checkFile` passes (prettier + eslint)

### DIFF
--- a/src/app/features/schedule/ical/get-relevant-events-from-ical.ts
+++ b/src/app/features/schedule/ical/get-relevant-events-from-ical.ts
@@ -5,6 +5,30 @@ import { loadIcalModule } from './ical-lazy-loader';
 // NOTE: this sucks and is slow, but writing a new ical parser would be very hard... :(
 
 /**
+ * Sanitizes iCal data to fix common formatting errors before parsing.
+ * Specifically handles "orphan" lines (like address continuations) that are missing
+ * the required whitespace indentation, which causes ical.js to crash.
+ */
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much!

PR #2388, opened 2022-12-26
feat: Integrate with Azure Devops

BalighMehrez:
# Description

Integration with Azure DevOps boards.

## Issues Resolved

#356 

## Check List

- [ ] New functionality includes testing.
- [ ] New functionality has been documented in the README if applicable.

PR #6446, opened 2026-02-09
feat(azure-devops): add Azure DevOps issue provider integration

### LINKED ISSUE #356: [Feature Request] Azure DevOps Boards
# Feature Request

How hard or what would it take to make Super Productivity work with Azure DevOps Boards?

pschaer:
## Problem

Users have requested Azure DevOps integration since 2019 (see #356). 
A previous attempt (#2388) was started but became outdated due to significant 
architectural changes to how Issue Providers work.

## Solution: What PR does

Fresh implementation of Azure DevOps issue provider, compatible with the current 
architecture where Issue Providers are independent from projects.

### Features:
- Connect to Azure DevOps using Personal Access Token (PAT)
- Search work items by title or work item ID
- Auto-import work items to backlog with scope filtering:
  - All work items in project
  - Assigned to me
  - Created by me
- Real-time refresh and polling of work item status
- Display of work item description, status, and assignee

### Security:
- WIQL injection prevention via proper escaping of search terms
- Secure Basic Auth header handling

### Technical:
- Follows current issue provider patterns (similar to ClickUp, Linear)
- Uses Angular `inject()` function per project conventions
- Includes unit tests for API service
- Uses translatable strings for issue content labels

## Related

- Closes #356 
- Supersedes #2388 (fresh implementation as [suggested by maintainer](https://github.com/super-productivity/super-productivity/pull/2388#issuecomment-2598307674))

## Checklist

- [x] New functionality includes testing
- [x] Follows Angular commit guidelines
- [x] Lint and tests pass


johannesjo:
Thank you very much! Could you have a look at the conflicts and the code review feedback (not necessarily all is valid)?

rklec:
BTW does this use/I found https://www.npmjs.com/package/azure-devops-extension-api and https://www.npmjs.com/package/azure-devops-extension-sdk, which seems to be officially maintained by Microsoft? See https://learn.microsoft.com/en-us/azure/devops/integrate/?view=azure-devops and https://github.com/Microsoft/azure-devops-extension-api respectively https://learn.microsoft.com/en-us/azure/devops/integrate/?view=azure-devops

pschaer:
> BTW does this use/I found https://www.npmjs.com/package/azure-devops-extension-api and https://www.npmjs.com/package/azure-devops-extension-sdk, which seems to be officially maintained by Microsoft? See https://learn.microsoft.com/en-us/azure/devops/integrate/?view=azure-devops and https://github.com/Microsoft/azure-devops-extension-api respectively https://learn.microsoft.com/en-us/azure/devops/integrate/?view=azure-devops

Thanks for the suggestion! I looked into the official Azure DevOps packages and found:
- `azure-devops-extension-api` and `azure-devops-extension-sdk`: those libraries are primarily designed for building extensions running *within* Azure DevOps (web widgets, dashboard components) 
- `azure-devops-node-api`: this is the general-purpose Node.js client for external applications

However, I'd prefer direct `HttpClient` integration over `azure-devops-node-api` for the following reasons:

1. **Consistency**: This implementation follows the same pattern as the existing Jira, GitHub, and GitLab providers in Super Productivity, using Angular's HttpClient for direct API calls. This maintains codebase uniformity and follows the established architectural patterns

2. **Control**: The direct REST API approach gives us full control over error handling and request/response patterns without being constrained by SDK abstraction layers

3. **Maintenance**: direct integration should be easier to maintain and debug and not introduce any new dependencies

While the official Azure DevOps packages are well-maintained, direct API integration should prove to be the most appropriate approach for this integration use case.

johannesjo:
I agree. We definitely should try to avoid adding additional libraries and stick to Rest or graph API if possible. They often add much bloat.

Stan-Stani:
@pschaer When I try to set it up and click test on the Setup Azure DevOps config modal, I get this 

```json 
{
    "$id": "1",
    "innerException": null,
    "message": "The requested version \"6.0\" of the resource is under preview. The -preview flag must be supplied in the api-version for such requests. For example: \"6.0-preview\"",
    "typeName": "Microsoft.VisualStudio.Services.WebApi.VssInvalidPreviewVersionException, Microsoft.VisualStudio.Services.WebApi",
    "typeKey": "VssInvalidPreviewVersionException",
    "errorCode": 0,
    "eventId": 3000
}
```

Request copied as fetch

```js
fetch("https://dev.azure.com/Redacted/_apis/connectionData?api-version=6.0", {
  "headers": {
    "authorization": "redacted",
    "priority": "u=1, i"
  },
  "body": null,
  "method": "GET"
});
```

**EDIT:** Ah, but if I still click save it does pull issues correctly it seems.

Stan-Stani:
Minor tweak that might be nice would be to put ticket type in the issue title, so instead of `#15104 Blogs` it could be `Change Request 15104: Blogs`. Though perhaps that could be my follow up PR if I really care about that, lol.

johannesjo:
### Code review

Found 1 issue:

1. Hardcoded string `'Advanced Config'` with `// todo translate` comment instead of using the existing translation key `T.G.ADVANCED_CFG`. CLAUDE.md says "**Translations**: UI strings must use the translation service (`T` or `TranslateService`)." The translation key already exists and is used elsewhere (e.g., in `sync-form.const.ts`), so replacing `'Advanced Config'` with `T.G.ADVANCED_CFG` should be straightforward.

https://github.com/super-productivity/super-productivity/blob/328e8586538dac0ee105a258fab9cb373a59ba82/src/app/features/issue/providers/azure-devops/azure-devops-cfg-form.const.ts#L45-L48

Stan-Stani:
Wonder if we could get images from the work item to load: 
TnZPsnndo9

pschaer:
@Stan-Stani @johannesjo Just a quick update – I've pushed fixes for the reported issues:
1. **Connection Error**: I downgraded the connectionData endpoint to 5.1-preview. It seems Azure DevOps Services (Cloud) strictly requires the -preview flag for this version, which was causing the 400 error.
2. **Issue Title Format**: Adding the work item type makes complete sense and I've updated the issue title format to [Type] [ID]: [Title] (e.g. Bug 123: My Issue) as requested.
3. **Localization**: Fixed the hardcoded "Advanced Config" string flagged by the bot.

I've verified these changes locally and everything looks green. Ready for another look when you have a moment!

Regarding **Display Images from the Work Items**: it would certainly be a nice feature, but I feel this would introduce too many new changes to this PR... let's address it in the future, once this is through, OK?

rklec:
Also when using LLMs for all replies please state that, it's a little exhausting when one seems to talk to a machine here.

> Connection Error: I downgraded the connectionData endpoint to 5.1-preview. It seems Azure DevOps Services (Cloud) strictly requires the -preview flag for this version, which was causing the 400 error.

That may work, but it is technically wrong. I tested it in https://github.com/super-productivity/super-productivity/issues/356#issuecomment-3879342455 already and the only correct (working) API in a self-hosted (not cloud!) Azure DevOps Server instance was `?api-version=1.0` or none apparently. I have no idea why, but it works this way...
That's also why I suggested the NodeJS clients or so, they may have some quirks like this already handled (hopefully), but yeah I also see the benefit of just having a few HTTP calls without another dependency (though three-shaking etc. could reduce the deps size and only bundle/include what is really used/required.).

rklec:
Tested https://daaf66c7.super-productivity-preview.pages.dev and it works BTW, some ideas:
* could it have a useful filter by default? Like only show tasks from/for the current iteration? Currently the search just returns e.g. _all_ issues ever assigned to one
* it does not seem to import the estimated time (for tasks usually in the field "Remaining Work"; otherwise "Effort" is common), that would be a huge benefit
* obviously then (optionally) synchronizing back the time, e.g. reducing the "remaining work" would be useful.
* similarly to Jira, could it fill the project backlog by default? It does not seem to do that, and I have configured it like this:
   grafik
  * Ah sorry apparently needed to click "auto-import to default project" then it seems to work. That is great, but suffers from the same UX problem explained above: It returns _too much_ aka any task ever assigned or so. I guess the limit of 48 items was just capped. (and it plans everything for "today" apparently)

src/app/features/issue/mapping-helper/get-issue-provider-tooltip.ts
```diff
@@ -21,6 +21,8 @@ export const getIssueProviderTooltip = (issueProvider: IssueProvider): string =>
         return issueProvider.projectId;
       case 'TRELLO':
         return issueProvider.boardName || issueProvider.boardId;
+      case 'AZURE_DEVOPS':
```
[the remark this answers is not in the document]
rklec: It's not a translation, LLM wrong here.

src/app/features/issue/providers/azure-devops/azure-devops-issue/azure-devops-issue-content.const.ts
```diff
@@ -0,0 +1,29 @@
...
+} from '../../../issue-content/issue-content.model';
+import { AzureDevOpsIssue } from './azure-devops-issue.model';
+import { T } from '../../../../../t.const';
+
+export const AZURE_DEVOPS_ISSUE_CONTENT_CONFIG: IssueContentConfig<AzureDevOpsIssue> = {
+  issueType: 'AZURE_DEVOPS',
+  fields: [
+    {
+      label: T.F.ISSUE.ISSUE_CONTENT.DESCRIPTION,
+      type: IssueFieldType.MARKDOWN,
```
rklec, line 13: I am not sure what impact this has but IIRC Azure DevOps Server does not really have Markdown support in tickets (in wikis it has), or has it? As for the description (which is not shown in my testing, I only see the title), it's I guess some HTML (editor).

src/app/features/issue/providers/azure-devops/azure-devops-issue/azure-devops-issue.model.ts
```diff
@@ -0,0 +1,15 @@
+export interface AzureDevOpsIssueReduced {
```
rklec, line 1: the current iteration (i mean the iteration where an issue is assigned to) could be useful to add here. In the GUI it's called "Iteration Path".

rklec: Also ADS heavily relies on teams/subteams etc., this would be another useful filter. It's called "Area Path" in the ADS GUI.

Image

src/app/features/issue/providers/azure-devops/azure-devops-issue/azure-devops-issue-content.const.ts
rklec: Okay I may actually be wrong, in the sidebar it shows the field "Description" and rendering seems to work (maybe because Markdown is also just rendered as HTML?). However, note not all issue types do have this field. For Tasks it's there e.g..
For Product backlog Items it's there too, however the mor eimportant information resides in "Acceptance Criteria" [for example](https://learn.microsoft.com/en-us/azure/devops/boards/backlogs/create-your-backlog?view=azure-devops&tabs=agile-process#key-fields-for-backlog-planning) and [as for bugs](https://learn.microsoft.com/en-us/azure/devops/boards/backlogs/manage-bugs?view=azure-devops#bug-work-item-type) it's usually "Repro Steps", "System Info" and "Acceptance Criteria". The field "Description" e.g. here doe does even exist and is thus shown as empty here.

I am not sure how to best handle this here. Ideally, it would of course be configurable. Or you'd just include the "usual"/common fields and display them.

rklec (commented):
As said, adjusting the query for items returned would be useful. The old PR also has some good ideas:
* [here a sensible default query is shown](https://github.com/super-productivity/super-productivity/pull/2388#discussion_r1064790936) (though adjusting it would be useful)
* > WIQL is a query language used to query work items form Azure DevOps APIs, it stands for (Work Items Query Language).
* Care needs to be taken [because the names and behavior of issue/item types can be adjusted](https://github.com/super-productivity/super-productivity/pull/2388#issuecomment-1667446697) in ADS.

src/app/features/issue/providers/azure-devops/azure-devops-issue/azure-devops-issue-content.const.ts
Stan-Stani: [I think they added markdown support in tickets recently?](https://devblogs.microsoft.com/devops/markdown-support-arrives-for-work-items/)

Stan-Stani:
> @Stan-Stani @johannesjo Just a quick update – I've pushed fixes for the reported issues:
> 
>     1. **Connection Error**: I downgraded the connectionData endpoint to 5.1-preview. It seems Azure DevOps Services (Cloud) strictly requires the -preview flag for this version, which was causing the 400 error.
> 
>     2. **Issue Title Format**: Adding the work item type makes complete sense and I've updated the issue title format to [Type] [ID]: [Title] (e.g. Bug 123: My Issue) as requested.
> 
>     3. **Localization**: Fixed the hardcoded "Advanced Config" string flagged by the bot.
> 
> 
> I've verified these changes locally and everything looks green. Ready for another look when you have a moment!
> 
> Regarding **Display Images from the Work Items**: it would certainly be a nice feature, but I feel this would introduce too many new changes to this PR... let's address it in the future, once this is through, OK?

I've tested the Issue Title Format change and can verify it does now include the type of issue as a prefix.

johannesjo:
I am merging this then! Thank you very much for creating this @pschaer and thank you very much for testing @rklec @Stan-Stani !!!

If you have any follow up issues please create a new issue and mention @pschaer.

pschaer:
@johannesjo @Stan-Stani @rklec thank you all very much for the feedback! 
Having this feature now in 17.1.8 allows me to significantly improve my personal worflow by searching (mostly by number), and automatically importing DevOps items, instead of having to manually enter them 😃 
I use DevOps foremost at work on multiple projects and the setup for each project can be very different.
@rklec your ideas are very good and apppreciated and I also would like to have a tighter integration with DevOps but that will need much more thought and effort than this first "minimal viable" integration.
Now that we got this rolling, let's think about how we could improve it in a future issue...
Have a good start of the week everyone!

rklec:
@pschaer Thanks and great to hear! Did someone already create issues for the remaining items?

CaptaiNiveau:
Seconding this, it doesn't seem like there's follow up issues. I'm really happy I found this app and would love to be able to sync task status back to DevOps. Relevant statuses might be that I'm committed when I start tracking it and that I finished work locally. A flow in DevOps might be 'new' -> 'approved' -> 'committed' -> 'checked-in' -> 'tested' -> 'done'. New items would obv be 'new'. 'checked-in' is automatically set by creating and pushing a commit in git. 'tested' is set manually if there's manual testing required. This might be nice to add as a button or maybe with a tag locally, not too important though. For all the other statuses, it'd be nice if it were possible to hook into events, like when I start tracking time for a task (maybe I can already do this using the automations? I'll look into that). 

Since every project in DevOps can have wildly different flows, I think implementing this with hooks that set the desired state when the task changes state locally might be a reasonable way to make most DevOps users happy. Maybe there could also be conditions, like only change the status if it's assigned to me or isn't already set to done - might be useful for some.

rklec:
@pschaer I created some followup issues and linked them (see above) and also found these new ones created by someone else: https://github.com/super-productivity/super-productivity/issues/7018 https://github.com/super-productivity/super-productivity/issues/7018
**Edit:** Also created https://github.com/super-productivity/super-productivity/issues/8556

@johannesjo maybe introducing a tag to group all `integration: azure-dev-ops` issues together or so would be useful?

PR #6450, opened 2026-02-10
fix: handle async voice loading in voice reminder settings (#6437)

### LINKED ISSUE #6437: 🚨 Voice Reminder: Can't select a voice

costajohnt:
## Summary

On some platforms (e.g. Linux with Chromium/Electron), `speechSynthesis.getVoices()` returns an empty array on the first call because voices load asynchronously. This causes the "Select a voice" dropdown in Voice Reminder settings to show no options.

**Fix:** Listen for the `voiceschanged` event to populate the voice dropdown when voices become available. The initial synchronous call is kept for platforms where voices are available immediately.

Fixes #6437

## Changes

- `voice-reminder-form.const.ts`: Refactored the `onInit` hook to extract a `populateVoices` helper that is called both synchronously and via the `voiceschanged` event listener

## Test plan

- [x] All 7,348 existing tests pass (both Europe/Berlin and America/Los_Angeles timezones)
- [x] Lint passes (TypeScript + SCSS)
- [x] Manual test on Linux: open Settings > Voice Reminder, verify voice dropdown populates

### DIFF
--- a/src/app/features/config/form-cfgs/voice-reminder-form.const.ts
+++ b/src/app/features/config/form-cfgs/voice-reminder-form.const.ts
@@ -65,16 +65,24 @@ export const VOICE_REMINDER_FORM: ConfigFormSection<DominaModeConfig> = {
       },
       hooks: {
         onInit: (field) => {
-          let voices: SpeechSynthesisVoice[] = getAvailableVoices() || [];
-          voices = getAvailableVoices();
-          //Log.log(voices);
+          const populateVoices = (): void => {
+            const voices = getAvailableVoices();
+            if (field.templateOptions) {
+              field.templateOptions.options = voices.map((voice) => ({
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much for this!

costajohnt:
@beerkumquatpome The fix was merged after v17.1.6 was released (v17.1.6 shipped Feb 9, this PR was merged Feb 10), so it's not included in that version yet. It should be available in the next release. In the meantime, you could build from the master branch if you'd like to try it sooner.

johannesjo:
> @costajohnt v17.2.1
> Linux
> I still can't select a voice.

Could you open up an issue for this with some logs please? This likely won't work on every platform there is.

johannesjo:
@beerkumquatpome ctrl+shift+i

Issue #5651, opened 2025-12-04
💡 Custom install location


legokol:
### Problem Statement

I have restricted rights at my work windows PC, so there are only some specific folders, from which I can run apps, and user profile is not one of them. First I tried portable app, but it didn't work and it didn't display any errors, so I don't know what the problem was there. Then I ran the installer. While the app wasn't able to run from my profile, I just moved files to a dedicated directory and was able to make it work. But moving program files by myself is somewhat of a hustle (I'm not sure I moved everything required and didn't break something) and it makes updating awkward: I have to repeat said process.

### Possible Solution

Add option to specify install directory for windows installer.

### Describe alternatives you've considered

I tried portable app, but it didn't work and didn't give me any errors. I also worked with web app for some time, but there were some challenges with webdav sync and cors, so desktop is preferable for me.

### Additional context

Using a windows 10 machine.

johannesjo:
You should be able to run the app via the portable version with the `--user-data-dir=` flag.

legokol:
@johannesjo this does work, thanks! I still think it would be nice to be able to choose install location though.

> You should be able to run the app via the portable version with the `--user-data-dir=` flag.

legokol:
Also, I figured out why this works now but didn't work before: I changed my TEMP path to a different one, where I can execute. When I run portable version now it runs, but windows tells me that working process is not the portable version I started, but some executable file from TEMP directory, is it ok?

legokol:
I surfed a bit and figured out, that extracting executables to run from temprorary dir is electron default setting ([related disussion](https://github.com/electron-userland/electron-builder/issues/8739)). I guess they provide a way to work around it, but I don't know if it should be changed in case of superproductivity. VS code, for example, doesn't do this, while using electron.

Basically, for now there are two options, if you can only run apps from specific directory:
1. Try to install regular version, than mannually move installed files from your profile dir and hope nothing breaks
2. Try portable version, but make sure that you CAN run apps from TEMP directory

legokol:
Closed by #6451

johannesjo:
Thank you @legokol !

PR #6451, opened 2026-02-10
feat: allow installation to custom location on windows

legokol:
## Problem

Some users (including myself) might have restrictions on PC, allowing to start apps only from specific directories. User directory, which is used for installation by default, is usually not one of them. Related issue #5651

## Solution: What PR does

Change electron builder configuration to allow to install app to custom location


johannesjo:
Thank you very much!

PR #6456, opened 2026-02-10
Update Swedish translations for issue-related terms

cortex:
- Use "ärende" instead of "problem"
- Use "ärendepanel" instead of "problempanel"
- Use "tavla" instead of "styrelse" consistently 
- Correct "finish day"


johannesjo:
Thank you very much!

PR #6492, opened 2026-02-12
feat(mobile): handle back button navigation of bottom sheet (project notes)

Garrett04:
## Problem
When the user opens a project note from the bottom sheet (project notes) on mobile and presses back navigation to close the note, the app minimizes. This can cause frustration to some, where they expect the project note to close and not the app. The same applies for closing the bottom sheet via back button.

## Solution: What PR does
Through the use of the browser history stack, the back navigation can be handled of a fullscreen dialog and the bottom sheet as well.
- Change the `closeOnNavigation` property value from `true` to `false` in `right-panel.component.ts`, to ensure that the bottom sheet doesn't close when navigating back from a project note (`dialog-fullscreen-markdown.component.ts`).
- Modify `dialog-fullscreen-markdown.component.ts`, and `notes.component.ts`, to handle the browser history stack accordingly.

### Before
[Click here to watch](https://gofile.io/d/wh7Cnk)

### After

[Click here to watch](https://gofile.io/d/ISfKfJ)

### DIFF
--- a/src/app/app.constants.ts
+++ b/src/app/app.constants.ts
@@ -94,4 +94,6 @@ export const HANDLED_ERROR_PROP_STR = 'HANDLED_ERROR_PROP';
 export const HISTORY_STATE = {
   MOBILE_NAVIGATION: 'mobileSideNav',
   TASK_DETAIL_PANEL: 'taskDetailPanel',
+  DIALOG_FULLSCREEN_MARKDOWN: 'dialogFullscreenMarkdown',
+  NOTES: 'notes',
 };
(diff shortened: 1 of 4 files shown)


Garrett04:
It can also be done for the web app, but i'm unsure whether to add back navigation through the browser back button or a close button right beside the save button on the fullscreen project note.
I prefer the latter. What about you @johannesjo?

johannesjo:
Thank you very much for this @Garrett04 ! I would guess that most desktop users don't use the back button much for these kind of interactions as there really is no standard pattern for this accross websites and wewbapps.  But you probably mean for the mobile web app case, right?  A close button next so the save button for that case sounds like a good idea to me.

PR #6499, opened 2026-02-12
fix: clear markdown preview when switching to task without notes

### LINKED ISSUE #6490: 🚨 Comments do not work correctly if they are empty [Boards]
### Version Used

17.1.6

### Expected Behavior

**Location**: Boards > KANBAN
**Scenario**: The KANBAN Board contains two items. Item A has text in its comment field and item B does not have text in its item field.
**Expected Behaviour**: You click on item A, the right sidebar opens and you see the text of item A in the comment field. Then, you click on item B and the comment field in the right sidebar does not contain text / the field is empty

### Current Behavior

**Location**: Boards > KANBAN
**Scenario**: The KANBAN Board contains two items. Item A has text in its comment field and item B does not have text in its item field.
**Current Behaviour**: You click on item A, the right sidebar opens and you see the text of item A in the comment field (_So far, so good_). But then, you click on item B and you still see the text of item A in the comment field.

That only happens if item B has no text / comment.

This bug occurs on both Windows and macOS.

### Steps to Reproduce

1. Click on "boards"
2. Open KANBAN
3. Create a new todo item (item A)
4. Add text to the comment field (of item A)
5. Create a new todo item (item B)
6. Click on item A
7. Click on item B (and open the right sidebar if this does not happen automatically)

### Console Output

```shell

```

costajohnt:
## Summary

Fixes #6490

When switching from a task with notes (item A) to one without notes (item B) in the Boards/Kanban view, the sidebar continued displaying item A's markdown content instead of clearing it.

**Root cause:** The `inline-markdown` component set `resolvedModel` to `undefined` when the model was empty. The ngx-markdown `MarkdownComponent.loadContent()` checks `if (this.data != null)` using loose equality — since `undefined != null` is `false`, `handleData()` was never called, leaving the previously rendered HTML in the DOM.

**Fix:** Set `resolvedModel` to empty string `''` instead of `undefined`. Since `'' != null` is `true`, ngx-markdown properly calls `handleData('')` → `render('')`, which clears the rendered content.

## Changes

- `inline-markdown.component.ts`: Use `''` instead of `undefined` when clearing `resolvedModel` (2 locations)
- `inline-markdown.component.spec.ts`: Update existing race condition test to expect `''` instead of `undefined`

## Test plan

- [x] All 7345 existing tests pass (Europe/Berlin TZ)
- [x] All 7333 existing tests pass (America/Los_Angeles TZ)
- [x] Linting passes
- [x] Manual test: Open Boards → Kanban, click task with notes, then click task without notes — sidebar should clear

### DIFF
--- a/src/app/ui/inline-markdown/inline-markdown.component.spec.ts
+++ b/src/app/ui/inline-markdown/inline-markdown.component.spec.ts
@@ -580,8 +580,8 @@ describe('InlineMarkdownComponent', () => {
       resolveDelayed('Task A notes');
       await Promise.resolve();
 
-      // Assert: resolvedModel should remain undefined (not stale Task A content)
-      expect(component.resolvedModel()).toBeUndefined();
+      // Assert: resolvedModel should remain empty (not stale Task A content)
+      expect(component.resolvedModel()).toBe('');
     });
   });
 
--- a/src/app/ui/inline-markdown/inline-markdown.component.ts
+++ b/src/app/ui/inline-markdown/inline-markdown.component.ts
@@ -120,7 +120,7 @@ export class InlineMarkdownComponent implements OnInit, OnDestroy {
     if (v) {
       this._updateResolvedModel(v);
     } else {
-      this.resolvedModel.set(undefined);
+      this.resolvedModel.set('');
     }
 
     if (!this.isShowEdit()) {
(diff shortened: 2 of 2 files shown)


johannesjo:
Thank you very much for this!

PR #6512, opened 2026-02-13
fix(bg-overlay): fix bg overlay not rendering in focus mode overlay

Garrett04:
## Problem
The bg-overlay isn't rendering in focus mode overlay due to its placement in `app.component.html`

## Solution: What PR does
- Move the .bg-overlay div element below .bg-image div element 
- Remove the background-color (line 48) under main selector in
focus-mode-overlay.component.scss

https://github.com/super-productivity/super-productivity/pull/6294#issuecomment-3890977065


johannesjo:
Thank you very much!

PR #6526, opened 2026-02-14
Prioritize text content in clipboard

### LINKED ISSUE #6495: 🚨 Paste into Task with Ctrl + V do NOT paste text but image ?!?
### Version Used

17.1.6

### Expected Behavior

Copy and PAste with CTRL +V should just paste some text into all fields.

### Current Behavior

I tried to paste some copied text into the Notes area using CTRL +V.  SP do not incorporate the text, but an image which is wrong.

Image

Image

### Steps to Reproduce

Copy some text in windows with CTL+ C ( I used Onennote) with clipboard history on in  windows 11.

Paste it into the notes area of a task.

### Console Output

```shell

```

puetsua:
## Problem

Copied text in the software such as Microsoft OneNote will provide multiple clipboard data types.

In our case, we should prioritize text instead of image data.

## Solution: What PR does

Fix #6495


johannesjo:
Thank you very much!!

PR #6529, opened 2026-02-14
feat(dialog-fullscreen-markdown): add close button to fullscreen markdown dialog

Garrett04:
## Problem
Users are unable to navigate back without saving the note (through the "Save" button, "Esc" key or "Ctrl+Enter"), especially in the web and desktop version.

## Solution: What PR does
With the addition of a close button, users can now exit the fullscreen editor and discard unsaved changes when editing task or project notes.
- Modify `close()` function in `dialog-fullscreen-markdown`
- Modify `inline-markdown.component.ts`, and `note.component.ts` to work with the `close()` function in `dialog-fullscreen-markdown`.
- Modify `dialog-add-note.component.ts` to handle back navigation on mobile.
- Remove auto save functionality for task note to accomodate for the close button functionality.

https://github.com/super-productivity/super-productivity/pull/6492#issuecomment-3893014917


Garrett04:
Hi Johannes,
I've handled an extra case, where the project note is made empty and the user clicks "Save" button, to which the note is deleted. The same goes for task notes, where it resets to the default text.
Also is it worth removing the auto-save functionality for task notes?

johannesjo:

Garrett04:
@johannesjo sorry i think you missed out on my comment here.

> Hi Johannes,
> I've handled an extra case, where the project note is made empty and the user clicks "Save" button, to which the note is deleted. The same goes for task notes, where it resets to the default text.
> Also is it worth removing the auto-save functionality for task notes?

johannesjo:
Oh sorry! Thanks for pointing it out.

> Also is it worth removing the auto-save functionality for task notes?

Which code exactly do you mean by this?

Garrett04:
> Which code exactly do you mean by this?

`src/app/ui/inline-markdown/inline-markdown.component.ts` lines 270-281

src/app/ui/inline-markdown/inline-markdown.component.ts
```diff
@@ -267,18 +267,13 @@ export class InlineMarkdownComponent implements OnInit, OnDestroy {
...
-    // Subscribe to live auto-save updates from fullscreen dialog
-    dialogRef.componentInstance.contentChanged.subscribe((content: string) => {
-      lastEmittedContent = content;
-      this.modelCopy.set(content);
-      this.changed.emit(content);
-    });
-
     dialogRef.afterClosed().subscribe((res) => {
-      // Only emit if content differs from last auto-saved content
-      if (typeof res === 'string' && res !== lastEmittedContent) {
```
Garrett04: @johannesjo this one here

johannesjo:
> `src/app/ui/inline-markdown/inline-markdown.component.ts` lines 270-281

This is not autosaving. It only saves when the dialog has a result and this only happens when save was clicked. Or am I missing  something?

Garrett04:
When I was working to add a "Close" button to the fullscreen editor, I noticed that whenever i close the task note, it gets saved. Once i removed that piece of code, the "Close" button started to work as expected (closing the note without saving). As far as i know, before this PR, there wasn't any way to close the fullscreen editor without saving (or hitting "Esc" or "Ctrl+Enter"). So I assumed that to be the autosave functionality for task notes.

johannesjo:
Ah alright. It seems to work fine now, I think?

Garrett04:
Yes

PR #6530, opened 2026-02-14
feat: add keyboard shortcuts for rescheduling tasks to common time periods

adnoh:
## Problem

Often in my workflow I want tasks to reschedule quickly to tomorrow, next week, or month. I'm a keyboard user. So I want to go through the list, select an item, reschedule it and move on.

## Solution: What PR does

Resolves: Keyboard shortcuts for common task rescheduling

- Add (configurable) shortcuts 1-5 for rescheduling tasks to: tomorrow, end of week, start of next week, end of month, start of next month
- Include task-reschedule.helper for date calculation logic
- Add full test coverage for reschedule calculations
- Add i18n support for all new shortcuts
- Safe default keyboard bindings (1-5) to avoid conflicts


adnoh (commented):
@johannesjo please have a look at the pr. It's my first on the project. It works for me locally as expected. I can also create a plugin out of this instead, but for me the feature is super helpfull.

johannesjo:
### Code review

Found 3 issues:

1. Missing comma in `en.json` after `TASK_RESCHEDULE_TO_NEXT_MONTH` will break JSON parsing and prevent the app from loading translations.

https://github.com/super-productivity/super-productivity/blob/b5bdd5d3416e7dda738b60d4bbda5e2144c110b5/src/assets/i18n/en.json#L1945-L1950

2. `(this._taskService as any)._getCurrentTaskReminder?.()` uses `as any` to call a method that does not exist on `TaskService`. This violates the project's strict TypeScript guidelines ([CLAUDE.md Anti-Patterns](https://github.com/super-productivity/super-productivity/blob/b5bdd5d3416e7dda738b60d4bbda5e2144c110b5/CLAUDE.md): "`any` type" -> "Use proper types"). The expression will always evaluate to `undefined`, falling through to the default. Consider just using `DEFAULT_GLOBAL_CONFIG.reminder.defaultTaskRemindOption` directly, or adding a proper public method to `TaskService`.

https://github.com/super-productivity/super-productivity/blob/b5bdd5d3416e7dda738b60d4bbda5e2144c110b5/src/app/features/tasks/task/task.component.ts#L718-L722

3. `nextWeek` date calculation is wrong on Mondays. `((8 - 1) % 7) || 1` evaluates to `0 || 1 = 1`, adding 1 day (Tuesday) instead of 7 days (next Monday). The `|| 1` fallback should likely be `|| 7`.

https://github.com/super-productivity/super-productivity/blob/b5bdd5d3416e7dda738b60d4bbda5e2144c110b5/src/app/features/tasks/task-reschedule.helper.ts#L43-L50

johannesjo:
### Code Review

Nice feature — keyboard-driven rescheduling is a great workflow addition. A few things to address beyond what was already flagged:

---

#### Bugs

1. **`nextWeek` is wrong on Mondays** (already noted in prior review)
   `((8 - 1) % 7) || 1` → `0 || 1 = 1` → schedules to Tuesday. Fix: `|| 7` instead of `|| 1`.

2. **`thisWeek` is wrong on Sundays**
   `(7 - 0) % 7 || 7` → `0 || 7 = 7` → schedules to *next* Sunday, not today. "End of this week" on Sunday should mean today. Fix: remove the `|| 7` fallback — `(7 - day) % 7` alone gives 0 on Sunday, which is correct.

3. **`as any` cast calls a non-existent method** (already noted)
   `(this._taskService as any)._getCurrentTaskReminder?.()` always returns `undefined`. Just use `DEFAULT_GLOBAL_CONFIG.reminder.defaultTaskRemindOption` directly.

#### Cleanup

4. **Dead import** in `task-reschedule.helper.ts`: `getDbDateStr` is imported but never used. Will fail linting.

5. **Dead export**: `getRescheduleLabel` is never used anywhere and returns hardcoded English strings (bypasses i18n). Remove it.

6. **Indentation** in `_handleRescheduleShortcut` (task-shortcut.service.ts): the method body is at class-level indent instead of method-body indent. Run `npm run checkFile` on it.

#### Tests

7. **Missing Monday test case** for `nextWeek` — that's exactly the case where the bug is. Please add it.

#### Design suggestions (non-blocking)

8. **Hardcoded 9 AM**: `setHours(9, 0, 0, 0)` ignores user preferences. Not a blocker for this PR, but worth a follow-up issue to make it configurable or derive from existing reminder settings.

9. **Default keys `1`-`5`**: Since shortcuts fire on focused tasks, there's some risk of accidental rescheduling when a user means to type a number. Consider `Shift+1` through `Shift+5` as safer defaults — but this is configurable, so not blocking.

---

Items 1-7 should be fixed before merge. Items 8-9 are suggestions for follow-up.

johannesjo:
Sorry, for the late review. This somehow slipped past me. Could you also check the failing CI jobs?

PR #6546, opened 2026-02-16
docs/wiki content v0.6

### LINKED ISSUE #6502: Clarification on Legacy, Experimental, and Beta Sync Integrations
I was trying to answer some questions brought up in https://github.com/super-productivity/super-productivity/discussions/6478 and found the following:

https://github.com/super-productivity/super-productivity/blob/e5aca2466307f77b930bb22be1696f89a247a5ea/src/app/features/config/form-cfgs/sync-form.const.ts#L130-L137

If all blob-based forms of synchronization are now "legacy" or deprecated in some way I think some clarification on for how long they will be supported is needed. At present, the choices are:
- a beta backend
- two "experimental" backends
- Dropbox which is presumably the most stable and preferred option.

Dropbox is also indicated in the code as "legacy" which means that nothing but the newest version of SuperSync will be supported at some point; however...

My understanding of some of the long-term plans is that SuperSync is actually just the ideal implementation of the new transaction-based sync protocol rather than sending increasingly large and error prone blobs of data that can't handle multi-device access.

First, very cool.

Second, this leads me to think that "Legacy" does not mean "Deprecated":

https://github.com/super-productivity/super-productivity/blob/e5aca2466307f77b930bb22be1696f89a247a5ea/docs/long-term-plans/multi-client-file-sync-reliability.md?plain=1#L138

The _implementation_ may be legacy but the _provider_ will remain supported? Can you also confirm what "experimental" means in this context? Are you trying to steer people towards SuperSync regardless of the "op-log" backend changes that would allow Syncthing and others to work correctly?

Thanks!

zenoprax:
- **docs(wiki): fix external link formatting**
- **docs(wiki): populate task management how-to guides**
- **docs(wiki): update original notes on data backups**
- **docs(wiki): update "new tasks" and "task integration notes"**
- **docs(wiki): add guide on choosing a backend**
- **docs(wiki): add note for configuring sync backends**
- **docs(wiki): clarify differences between iCal and CalDAV**
- **docs(wiki): lint headings and lists**

---

This completes the initial "How-To" notes. Further additions will require more visuals and tighter integration with larger "Quickstart" guides so that they are more of a teaching tool rather than purely instructive/directive.

Closes #6502

Future work to come:

v0.7: Establish a basic series of notes for future dev documentation; primarily borrowing from the README and "how to build" docs to bolster the How-To section; some API and Plugin documentation prep

v0.8: Content locked in. Now just fact-checking what is there and a general audit of the whole thing for quality and correctness. Draft of PR and Issue templates for peer review.

v0.9.X: Reserved for all the last minute stuff :D

v1.0: Truncate README and redirect to Wiki as required; modify in-app help links as required.


johannesjo:
Thank you very much  @zenoprax !

PR #6548, opened 2026-02-16
ci/discussion pruning

zenoprax:
- **ci(community): initial discussion pruning workflow**
- **ci(community): revised workflow for revised action**
- **ci(community): add first valid workflow**
- **ci(community): switch action to controlled fork**

This workflow is not scheduled to run until next weekend.
The manual mode defaults to dry-run.
The extra commits to help make it easier to trace providence.
Last commit is to make `git revert` easy if it becomes necessary.

Lots of testing done so this is safe to pull in for some dry-runs.


johannesjo:
Thank you very much for this!

PR #6565, opened 2026-02-17
Feat/start of next day

johannesjo:
## Problem

Start of the next day is not handled well, which leads to unexpected bugs if set.

## Solution: What PR does

  "Start of Next Day" offset feature — allows users to configure when their logical "today" starts (e.g., 2 AM instead  
  of midnight). This is useful for people who work late and want tasks to still count as "today" past midnight.
                                                                                                                        
  - New config: startOfNextDayDiff (0–23 hours) stored in DateService, persisted in NgRx as
  AppState.startOfNextDayDiffMs
  - Offset-aware utilities: New isTodayWithOffset() and DateService.isToday() that account for the offset when
  determining what day a timestamp belongs to
  - Migrated all "is today?" checks across selectors, effects, reducers, and components to use offset-aware variants
  (overdue detection, planner columns, scheduled tasks, calendar events, daily summary, etc.)
  - Planner actions consolidated into meta-reducers (planTaskForDay, moveBeforeTask, transferTask) for atomic
  multi-entity updates
  - LWW sync support: syncTodayTagTaskIds handles both dueDay and dueWithTime with offset awareness
  - Comprehensive tests: boundary condition tests (e.g., 3:59 AM vs 4:00 AM with 4h offset), selector tests, reducer
  tests

### DIFF
--- a/src/app/root-store/meta/task-shared-meta-reducers/lww-update.meta-reducer.spec.ts
+++ b/src/app/root-store/meta/task-shared-meta-reducers/lww-update.meta-reducer.spec.ts
@@ -1633,6 +1633,167 @@ describe('lwwUpdateMetaReducer', () => {
     });
   });
 
+  describe('TODAY_TAG.taskIds sync on task dueWithTime change', () => {
+    const TODAY_STR = '2025-12-31';
+    const TODAY_TIMESTAMP = new Date(2025, 11, 31, 14, 0, 0).getTime();
+    const TOMORROW_TIMESTAMP = new Date(2026, 0, 1, 14, 0, 0).getTime();
+
+    beforeEach(() => {
+      jasmine.clock().install();
+      jasmine.clock().mockDate(new Date('2025-12-31T12:00:00'));
+    });
+
+    afterEach(() => {
+      jasmine.clock().uninstall();
+    });
+
+    const createStateWithDueWithTime = (
+      taskDueDay: string | undefined,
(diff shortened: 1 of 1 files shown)


johannesjo:
@novikov1337danil I know this is a huge change, but I was hoping that you – since you are familiar with the whole date logic – might have a glance at this.  I would like to know if you think the overall logic is sound (no thorough code review or something like this is needed) or if I missed something crucial.

novikov1337danil:
Unfortunately, I don't have much free time on weekdays, so i did just a quick review and currently have only very minor comments for refactoring and code reuse that aren't related to the datetime handling and feature itself. Everything looks good to me.

However, I also did a quick test your branch here https://6cc46f1d.super-productivity-preview.pages.dev (it's great that this option exists) and I see a problem. I don't know if this existed before.

**These are the approximate reproduction steps:**

0. The time is now 22:50 (18 feb)
1. Create a task in the Inbox, schedule it for "today"
2. Set setting startOfNextDay to the next hour - 23 h
3. **Go to the Inbox and see that the created task is marked not as "today," but as "18 feb" (but it is today).**
4. Create a new task for "today," and see that it's marked as "today," not "18 feb" like the previous one.
5. Set setting startOfNextDay back to 0
6. **Go to inbox and see the first task will become "17 feb" (shifted back one day)**

johannesjo:
Thank you very much! I look into it!

Issue #6324, opened 2026-02-02
🚨 jira issue provider does not honor proxy env variables


iaavo:
### Version Used

17.0.9

### Expected Behavior

When setting HTTP_PROXY and HTTPS_PROXY I expect the application to honor these env variables and send any requests to the designated jira host through the configured proxy server.

### Current Behavior

HTTP_PROXY and HTTPS_PROXY are ignored

### Steps to Reproduce

1. Set HTTP_PROXY and HTTPS_PROXY
2. Start SuperProductivity
3. Add Jira Connector
4. Fill in Details
5. Test Connection

Popup appears: `Connection Failed`

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell
Jira: request to https://my.jira.server.com/rest/api/latest/issue/picker?showSubTasks=true&showSubTaskParent=true&query=r&currentJQL= failed, reason: getaddrinfo ENOTFOUND my.jira.server.com
```

iaavo:
So I added a commit which fixes the issue at least in my local dev setup. However, I was unable to get the electron app to run as it cannot find the newly introduced `proxy-agent` dependency. I'm not sure what is missing.

iaavo:
I'm able to verifiy that the provided commit fixes the issue. It's just that `npm run dist:win` creates an exe file which is missing the `proxy-agent` dependency. Can you provide a hint how I can fix that? Then I can create a proper pull-request.

Tekpakma:
I noticed theire has been some great work to improve this but change: 
https://github.com/super-productivity/super-productivity/commit/319727f90854645b8c63ee256c3c43760b0b5db6 doesnt honor NO_PROXY which should be fairly straightforward to implement.
For a Corporate setup this breaks because f.e. a Proxy for external services is set, but Internal Services like Jira are not routed.

PR #6574, opened 2026-02-18
fix: add proxy env support for jira integration #6324

ttisch:
## Problem

Jira integration did not respect proxy env vars, so i was not able to get it to work in my company with a corporate proxy server.

## Solution: What PR does

- Add https-proxy-agent package
- Use proxy env vars (if configured) to create HttpsProxyAgent() for jira communication

### DIFF
--- a/electron/start-app.ts
+++ b/electron/start-app.ts
@@ -8,7 +8,6 @@ import {
   ipcMain,
   powerMonitor,
   protocol,
-  session,
 } from 'electron';
 import { join } from 'path';
 import { initDebug } from './debug';
@@ -29,7 +28,6 @@ import {
   processPendingProtocolUrls,
 } from './protocol-handler';
 import { getIsQuiting, setIsLocked } from './shared-state';
-import { getProxyUrl } from './proxy-agent';
 
 const ICONS_FOLDER = __dirname + '/assets/icons/';
 const IS_MAC = process.platform === 'darwin';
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much for this. Is it really necessary to add a new dependency? I consider this more of an niche use case and adding a new dependency always brings technical debt.

ttisch:
First of all - Thanks for this amazing tool and the feedback. I completely understand the concern about adding dependencies.

But my reasoning for using https-proxy-agent is that HTTPS proxy tunneling (CONNECT handling, TLS wrapping, edge cases) is non-trivial and security-sensitive. Rather than implementing custom proxy logic, this relies on a mature and widely used library that is purpose-built for this.

Supporting standard HTTP_PROXY / HTTPS_PROXY environment variables is expected behavior for Node/Electron applications and is required in many corporate environments.

johannesjo:
Thanks for working on this! Proxy support is important for corporate environments, so we should add this probably. Depdency also seems rather small, so it is  likely ok.

One concern: adding proxy handling only to `electron/jira.ts` means the other issue providers (GitHub, GitLab, Redmine, Gitea, etc.) remain broken behind a proxy. Proxy configuration is an environment-level concern, not a per-service one.

Looking at the architecture, Jira is the only provider making HTTP requests from the Electron main process via `node-fetch`. All other providers go through the renderer, where Chromium handles proxies via system settings or Electron's `session.setProxy()`.

A more complete approach might be:

1. For renderer requests (all providers except Jira): use Electron's [`session.setProxy()`](https://www.electronjs.org/docs/latest/api/session#sessetproxyconfig) to configure proxy support app-wide based on env vars
2. For main-process requests (`node-fetch` in `jira.ts`): create a shared utility that builds the appropriate agent from proxy env vars, so future main-process HTTP calls automatically get proxy support

Also worth noting: `rejectUnauthorized: false` passed to the `HttpsProxyAgent` constructor only affects the proxy connection's TLS, not the target Jira server's TLS handshake (verified in the `https-proxy-agent` v7.0.6 source — the constructor options land in `connectOpts` used at line 96, but the target `tls.connect()` at line 133 uses per-request `opts` from Node's HTTP machinery). This means the proxy + self-signed cert combination would silently fail.

ttisch:
I tested other providers and they are automatically respecting proxy environment variables.

But i get your point regarding `rejectUnauthorized: false` only affecting proxy connection's TLS, not the target Jira server's TLS handshake. I fixed that by creating a shared utility that handles it better.

Thanks again for your time + review!

ttisch:
I'm unsure. Is this something I need to fix on my side? Or can I do something to help improving this PR?
Just asking .. didn't want to stress you out.

Thanks

johannesjo:
I am merging this @ttisch . We can deal with the other issues later. Thank you very much!

PR #6583, opened 2026-02-20
fix(trello): remove invalid bearer header and prevent settings crash

mycochang:
## Summary
- Removes invalid \`Authorization: Bearer\` header from Trello API calls, which caused authentication failures.
- Removes \`devError(e)\` from issue provider components that caused the settings page to crash on validation errors.
- Adds explicit error logging and UI snackbar notifications when Trello board fetching fails.


mycochang:
Closing as this was based on a stale fork and the core issue was resolved upstream.

Issue #5504, opened 2025-11-13
🚨 Inconsistent date format: "planned at" date different from "repeat" date and date show in task list


eylenburg:
### Version Used

16.3.1

### Expected Behavior

Dates should be consistent everywhere, e.g. dd/mm/yyyy or mm/dd/yyyy or yyyy/mm/dd

### Current Behavior

<img width="604" height="649" alt="Image" src="https://github.com/user-attachments/assets/5bdc6e37-2f58-4859-84a0-3ca70546a5ec" />

As you can see in the screenshot, the date (1st March) is shown as 1/3 in the task list and in the "repeated yearly on" item but the "planned at" date is shown as 3/1 in the issue panel.

### Steps to Reproduce

Open issue panel, look at your date

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
@novikov1337danil can I assign this to you? This seems related to the issues you are currently tackling.

novikov1337danil:
> [@novikov1337danil](https://github.com/novikov1337danil) can I assign this to you? This seems related to the issues you are currently tackling.

Yes, sure

muaddib32:
Please ensure that the date format for the time format locale 'English (UK)' is set to dd/mm/yyyy (I believe it is but as the OP has pointed out, date formats are applied inconsistently). And make it explicit in the text for the settings that setting the time format also sets the date format.

rjulian:
Not sure if this should be a separate issue, but I'm still seeing these inconsistent date formats. It looks like the repeating date may be hardcoded to mm/dd/yyyy (I have my settings to do dd/mm/yyyy). Let me know if you'd like me to create a separate issue. Version is 16.7.3. Screenshot for reference:

<img width="1212" height="417" alt="Image" src="https://github.com/user-attachments/assets/095d5d80-6356-49f7-abef-fe3f72691067" />

novikov1337danil:
@eylenburg sorry for the delay. It's already fixed. You can check it again

novikov1337danil:
> Not sure if this should be a separate issue, but I'm still seeing these inconsistent date formats. It looks like the repeating date may be hardcoded to mm/dd/yyyy (I have my settings to do dd/mm/yyyy). Let me know if you'd like me to create a separate issue. Version is 16.7.3. Screenshot for reference:

Thank you! This bug will be fixed in the next release

PR #6594, opened 2026-02-21
refactor(date-picker): unify date input handling with new date-picker input

novikov1337danil:
## Problem

https://github.com/super-productivity/super-productivity/issues/5504#issuecomment-3706990528

The app had no single date input, so dates could differ across screens.

## Solution: What PR does

1. Created a reusable component `date-picker-input`, now used in two places in the app (I haven’t searched everywhere yet, but it may make sense to replace other instances too). It's also bound to a Formly field of type `date`.
2. Removed the duplicate `formatDateYYYYMMDD` in favor of `getDbDateStr` (these are identical functions with different names).
3. Added an additional tests


johannesjo:
Thank you very much for this! Let me know when this is read for review @novikov1337danil !

novikov1337danil:
@johannesjo ready for review

johannesjo:
Looks good and seems to work well! Thank you very much @novikov1337danil !

Issue #5725, opened 2025-12-14
💡 linux top-bar can be use as Countdown timer in system top bar


MyNameIsJanes:
### Problem Statement

The countdown timer currently relies on an external “always-on-top” popup,On Fedora 42 (GNOME), the timer also does not appear in the system top bar when the app is minimized, making it easy to lose track of time and disrupting the workflow.

### Possible Solution

Display the active countdown timer directly in the system top bar.
This would allow users to monitor time without keeping a floating window open and would keep the timer visible even when the app is minimized.

johannesjo:
This would be very nice, but unfortunately the standard indicator on most distros does not support any of this.

PR #6599, opened 2026-02-22
feat: add today's tasks and active timer to tray context menu

onesvat:
## Problem

Currently, the system tray menu is very basic ("Show App", "Quit") and requires users to bring the main application to the foreground in order to switch or pause tasks. 

Furthermore, Linux users suffer from a platform limitation where the active task name and tracking time cannot be rendered natively on the system bar next to the tray icon (`tray.setTitle()` isn't supported contextually due to `libappindicator` constraints in Electron). #5725

## Solution: What PR does

| Not Running | Running (No Estimate) |
| :---: | :---: |
| not_running | without_estimate |

| Running (With Estimate) | Running (Over Estimate) |
| :---: | :---: |
| with_estimate | with_estimate_passed |

This PR overhauls the system tray by introducing an interactive context menu that displays the active timer and today's tasks. It effectively resolves the timer visibility issue on Linux by integrating it directly into the menu and significantly speeds up workflow tracking across all platforms.

- **Dynamic Task List & Toggling:** Lists all of "Today's Tasks" in the context menu as clickable items. Selecting a task automatically starts it, or pauses/resumes it if it's already active.
- **Timer Visibility:** The context menu visually highlights the ongoing task and prominently displays its formatted tracking timer (e.g. `1h 14m`).
- **Store Syncronization:** Added a new IPC event bridge (`updateTodayTasks`) within `TaskElectronEffects` to smoothly push the list of tasks from the NgRx store directly to Electron.

## Question

Currently, **left-clicking** the tray icon on macOS/Windows brings the app to the foreground, while on Linux it simply opens the context menu (likely due to an OS limitation/bug).

Given the new controls in the context menu, should we remove the left-click `showApp()` listener so the tray icon *always* opens this menu on all platforms? Or should we keep the current left-click behavior as it is?

### DIFF
--- a/src/app/features/tasks/store/task-electron.effects.spec.ts
+++ b/src/app/features/tasks/store/task-electron.effects.spec.ts
@@ -29,6 +29,7 @@ describe('TaskElectronEffects', () => {
       on: jasmine.createSpy('on'),
       updateCurrentTask: jasmine.createSpy('updateCurrentTask'),
       setProgressBar: jasmine.createSpy('setProgressBar'),
+      onSwitchTask: jasmine.createSpy('onSwitchTask'),
     };
 
     actions$ = new Subject<any>();


johannesjo:
Thank you very much for this!

Issue #6602, opened 2026-02-22
🚨 Edit folder dialog shows empty input instead of current folder name


alirahimifachr:
### Version Used

latest, windows, web-app, android

### Expected Behavior

The input field should be pre-filled with the current folder name, allowing the user to edit it.

### Current Behavior

The input field is empty, requiring the user to type the entire name again.

### Steps to Reproduce

1. Right-click on a project folder in the left sidebar
2. Select "Edit" from context menu
3. Dialog opens with empty input field

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

PR #6603, opened 2026-02-22
fix(dialog): pre-fill folder name in edit dialog

alirahimifachr:
## Problem

#6602 

## Solution: What PR does

Initial `textVal`.


johannesjo:
Thank you very much!

johannesjo:
Thank you very much for this!

PR #6605, opened 2026-02-22
fix: reduce verbosity of error logs

mycochang:
## Problem

Some error handlers are excessively verbose and dump large objects or full payloads into the console and log files, which clutters the logs and makes debugging harder.

## Solution: What PR does

- **Trello**: Updates the API error handler to log only the parsed error message rather than the full `HttpErrorResponse` object.
- **Sync/Encryption**: Truncates the payload logged during a `DecompressError` so that it only outputs the length of the string instead of a slice of the payload.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,1 +1,1 @@
-npm run lint && ./run.sh test
+npm run lint && npm run test


johannesjo:
### Code review

Found 1 issue:

1. `.husky/pre-push` references `./run.sh test`, but `run.sh` does not exist anywhere in the repository. This will break the pre-push hook for every developer, causing all `git push` attempts to fail with "No such file or directory". The previous version (`npm run lint && npm run test`) should be preserved.

https://github.com/super-productivity/super-productivity/blob/779c52096da20d38d5a9de8e4c3a29d2612a8ec0/.husky/pre-push#L1

.husky/pre-push
```diff
@@ -1 +1 @@
-npm run lint && npm run test
+npm run lint && ./run.sh test
```
johannesjo: Actually thinking about it, this change should not be at all. Please remove it.

mycochang: Ah shoot yeah that's my playwright workaround bc i don't use chrome. Sorry.

mycochang:
Fixed. Trello user filter nice for preventing tasks of others entering inbox

johannesjo:
Thank you very much!

PR #6607, opened 2026-02-22
Update README with .env.example setup instructions

darcosion:
Added instructions for copying .env.example and renaming it for configuration.

## Problem

When I follow the documentation to create a webdav server, it say : 
```sh
# docker compose up -d
env file /root/sp/.env not found: stat /root/sp/.env: no such file or directory
```

## Solution: What PR does

I show a basic .env modification considering there is a `.env.example` file into docker folder.


darcosion:
Just to let you know. I just want to set a webdav server and .env is useless for that, so I think my fix on documentation is an error. Please ignore this merge request

PR #6631, opened 2026-02-24
Add ro and ro-md languages

SilverGreen93:
This PR adds the infrastructure needed to add Romanian translations.

The json translations are not yet translated. They will be updated in future commits as the translation progresses.


johannesjo:
Thank you very much!

PR #6645, opened 2026-02-26
Fixed an important help system error

Gitoffthelawn:
For recurring tasks, it stated: "...and will be automatically marked as completed at the end of the day".  My experience indicates, thankfully, this is not true!  Please confirm.


johannesjo:
Thank you!

PR #6646, opened 2026-02-26
docs: update Arch Linux AUR installation method to use AUR helper

suvadipbhuiya2004:
### Description
This PR updates the Arch Linux installation instructions to use an AUR helper (like `paru` or `yay`) instead of the manual `makepkg` process. 

**Why this is helpful:**
* **Better User Experience:** Reduces a multi-step manual process (`git clone`, `cd`, `makepkg -si`) down to a single, standard command.
* **Dependency Resolution:** AUR helpers automatically resolve and build any secondary AUR dependencies, which the standard `makepkg -s` flag cannot handle.

**Changes made:**
* Removed the manual `git clone https://aur.archlinux.org/superproductivity-bin.git` and `makepkg -si` commands.
* Replaced with `paru -S superproductivity` (also noted `yay` as a popular alternative).


johannesjo (commented):
Thanks for the PR! A couple of observations:

**Package name change:** The existing instructions reference `superproductivity-bin` (a precompiled binary package), but this PR switches to `superproductivity` (which builds from source). These are different AUR packages with different trade-offs — `*-bin` packages install faster since they skip compilation. Was this an intentional change, or should the instructions use `superproductivity-bin` with the AUR helpers instead (e.g., `paru -S superproductivity-bin`)?

**AUR helpers vs makepkg:** The switch from `makepkg` to AUR helpers is a reasonable suggestion since it simplifies the workflow, but `makepkg` is the standard Arch-endorsed method (ships with `base-devel`) and doesn't require installing additional tools. It might be worth keeping `makepkg` as a fallback or noting that `paru`/`yay` need to be installed separately.

johannesjo:
@alosarjos could you have a look on this?

suvadipbhuiya2004:
Thanks for the review and the thoughtful feedback!

Regarding your observations:

**1. Package name change (`superproductivity` vs `superproductivity-bin`)**
The thought process here was based on how `paru` handles the base package name. When running `paru -S superproductivity`, it automatically detects the available providers and prompts the user to choose their preferred version, like this:

```text
paru -S superproductivity
:: Resolving dependencies...
:: There are 3 providers available for superproductivity:
:: Repository AUR:
    1) superproductivity  2) superproductivity-bin  3) superproductivity-git
Enter a number (default=1):
```
However,  `yay` requires the user to be specific. To keep things consistent and ensure users get the faster precompiled installation, I agree it could be better to specify `superproductivity-bin` explicitly. Let me know if you'd prefer, and I'll update the PR to reflect this.

**2. AUR helpers vs makepkg**
That is a completely fair point. Since `makepkg` is the official Arch-endorsed method, it makes sense not to replace it entirely. I'm happy to update the PR to restore the makepkg instructions as the primary method and frame the `paru`/`yay` commands as an optional alternative if you think that's best.

Let me know if you would like me to proceed with these changes.

alosarjos:
> @alosarjos could you have a look on this?

@johannesjo While it could make sense, I'm more worried about this being an obvious AI generated PR. I'm not sure what's your position on this regard.

suvadipbhuiya2004:
> > @alosarjos could you have a look on this?
> 
> @johannesjo While it could make sense, I'm more worried about this being an obvious AI generated PR. I'm not sure what's your position on this regard.

The pr is not AI generated i have used AI to fix my grammers and formatting of my reply nothing other than that.

johannesjo:
> While it could make sense, I'm more worried about this being an obvious AI generated PR. I'm not sure what's your position on this regard.

I am okay with that as long as the code is sound and it makes sense to the project. Do you think merging this would be an improvement? I really have no clue about the whole arch ecosystem tbh...

alosarjos:
Since we already got the regular upstream method I would rather write it as what we used to have or using AUR helpers and then add this new section.

I'm on my phone so I can't write it properly here but I think it's easy to get what I mean

suvadipbhuiya2004:
> Since we already got the regular upstream method I would rather write it as what we used to have or using AUR helpers and then add this new section.
> 
> I'm on my phone so I can't write it properly here but I think it's easy to get what I mean

Will the following work?
Then I will update the PR

#### AUR - Arch Linux
##### Without AUR helpers
```bash
git clone https://aur.archlinux.org/superproductivity-bin.git
cd superproductivity-bin
makepkg -si
```
##### With AUR helpers
```bash
paru -S superproductivity-bin
# or
yay -S superproductivity-bin
```

alosarjos:
I see that ok. I may even suggest to put the AUR helpers first since it's the most common way to install AUR packages.

johannesjo:
Thank you very much! Sorry for the delay!

PR #6649, opened 2026-02-27
Udate german DE.json file

ExoGitH:
Update the DE.json (german) file. It should now have most newer text after the 17.x changes. Still need some newer update from en.json and "improvement" of wording.


ExoGitH:
will do the comma check myself and repull.

PR #6667, opened 2026-02-28
fix(focus-mode): allow long parent task titles to wrap

### LINKED ISSUE #6651: 🚨 In focus mode, long parent task titles get cut off when subtask titles are short
### Version Used

17.1.8

### Expected Behavior

When the parent task title is really long, it should expand to fill the full screen and wrap, just like the title of the task that is being worked on.

### Current Behavior

When the subtask title is short but the parent task title is very long, the parent task title gets truncated instead of expanding the container or wrapping.

When the subtask title is really long too, it expands the container and the parent task title has more room as expected. The bug is just when the subtask title is short and the parent task title is long.

### Steps to Reproduce

1. Make a task with a short title
2. Make a subtask with a long title
3. Start a focus mode session on the subtask

This one is my fault, I built that feature. Noting it here to get it out of my head and hopefully come back and fix it later.

### Console Output

```shell

```

costajohnt:
## Summary

- Remove single-line truncation (`white-space: nowrap`, `overflow: hidden`, `text-overflow: ellipsis`) from `.parent-title` in focus mode
- Replace with `overflow-wrap: break-word` and `word-break: break-word` so long parent task titles wrap naturally
- Preserves `max-width: 90%` to keep text from touching screen edges

Closes #6651

### DIFF
--- a/src/app/features/focus-mode/focus-mode-main/focus-mode-main.component.scss
+++ b/src/app/features/focus-mode/focus-mode-main/focus-mode-main.component.scss
@@ -95,7 +95,6 @@
   line-height: 1.2;
   max-width: 90%;
   overflow-wrap: break-word;
-  word-break: break-word;
 
   @media (max-width: 499px) {
     font-size: 12px;


johannesjo:
Thank you very much!

PR #6732, opened 2026-03-05
fix #6684

### LINKED ISSUE #6684: 🚨Tasks are transferred to a new profile
### Version Used

17.2.4

### Expected Behavior

When I create a new profile, it should be empty. So it should not contain any tasks, projects, tags and should have the default settings.

### Current Behavior

When I create a new profile, the contents of the existing profile are transferred. That doesn't make sense. If I have a private profile that I've been working with for a long time and then want to create a new profile for work, I expect it to be a completely new profile.

**Proposal:** You could offer the ‘Duplicate profile’ function if you want to transfer the existing information.

**Side note:** I recently submitted a ticket along these lines (#6579).

### Steps to Reproduce

see above

### Console Output

```shell

```

overcuriousity:
## Problem

the existing code called clearAllOperations() when switching to a new profile. This method clered only the ops log, but inherited the state_cache which contained data from the original profile and populated its own data from it.

## Solution: What PR does

The replacement of clearAllOperations() with importCompleteBackup(emptyData, true, false) clears the ops log and **writes a new empty AppDataComplete to state_cache** which should solve the issue.

AI was used to support with fix implementation. No known regressions were introduced.

### DIFF
--- a/src/app/features/user-profile/user-profile.service.ts
+++ b/src/app/features/user-profile/user-profile.service.ts
@@ -3,13 +3,11 @@ import { DEFAULT_PROFILE_ID, ProfileMetadata, UserProfile } from './user-profile
 import { UserProfileStorageService } from './user-profile-storage.service';
 import { SyncProviderManager } from '../../op-log/sync-providers/provider-manager.service';
 import { BackupService } from '../../op-log/backup/backup.service';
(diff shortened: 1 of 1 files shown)


johannesjo:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

johannesjo:
Thank you very much for this!

PR #6751, opened 2026-03-06
fix(issue): show board title in Nextcloud Deck provider tooltip

iwilltry42:
## Problem

Nextcloud Deck issue provider, if added multiple times (for different boards) showed up as NEXTCLOUD_DECK for every entity - easy to confuse.

## Solution: What PR does

Add missing NEXTCLOUD_DECK case to tooltip and initials helpers, showing the selected board title prefixed with "Deck:" instead of the raw provider key.


johannesjo:
Thank you very much!

PR #6752, opened 2026-03-06
feat: option to duplicate existing issue providers

iwilltry42:
## Problem

Adding multiple providers of the same type required re-typing all the input over and over again (e.g. for multiple Nextcloud Deck Boards).

## Solution: What PR does

Adds button to duplicate existing issue providers:

image


johannesjo:
THank you very much!

PR #6759, opened 2026-03-07
Added line through to archived search results

### LINKED ISSUE #6747: 💡 Make archived tasks in search results more distinguishable
### Problem Statement

This an example of a search result list.
Image

A few things:

1. Archived tasks have this little icon at the end; it's not that obvious.
2. The archive date is missing.
3. Why are archived tasks still in "Inbox"? Would be great if we have an "Archive".

### Possible Solution

1. Display archived tasks in another color? Maybe gray. Or, striked through. This way it's more obvious that they are archived.
2. Add the archive date in the list per task.
3. Add an "Archive"-thingy. But I understand that's a bigger change; 1) and 2) are maybe quick wins.

There is a great plugin: https://github.com/baiyina/Archived-Tasks-Viewer
But it would be nice if 1) und 2) would be implemented natively.

### Additional context

17.2.4

jcuenca95:
## Problem

Closes #6747 

## Solution: What PR does

- Added css class `task-archived`
- Using `ngClass` to add this new class to search results which `isArchiveTask` is true


johannesjo:
Thank you very much!

PR #6767, opened 2026-03-09
feat(focus-mode): add reset button for pomodoro session counter

### LINKED ISSUE #6709: 💡  Ability to reset Pomodoro session like before
### Problem Statement

Currently there's no way to reset the Pomodoro session counter.

There used to be a button when hovering over the pause button that allowed for both the timer and the session counter to reset.

### Possible Solution

Bring back the reset button!

This is very annoying as I need to skip multiple sessions if I want to reset and even then it's not a true reset (I just go to the nearest multiple of 4 so that the 4th break afterwards is the long one)

### Additional context

This is my first Github issue request, be gentle please! I hope I made you understand the issue I have!!!
For now I'll be downgrading to an older version.

Br1an67:
Fixes #6709

## Problem

Users had no way to reset the Pomodoro session counter during an active focus session. The reset button was only available during the break screen, forcing users to either wait for a break or skip multiple sessions if they wanted to reset the counter to start fresh with the long break cycle.

## Solution: What PR does

Adds a reset button to the main focus mode UI (next to the complete session button) that allows users to reset the Pomodoro session counter during an active session. The button:
- Is only visible when in Pomodoro mode during the InProgress state
- Uses the same `resetCycles` action already used by the break screen
- Reuses the existing translation key `T.F.FOCUS_MODE.RESET_CYCLES`
- Uses the same `restart_alt` Material icon as the break screen for consistency

## Changes

- `src/app/features/focus-mode/focus-mode-main/focus-mode-main.component.ts`:
  - Added `resetCycles` action import
  - Added `isPomodoro` computed property
  - Added `resetCycles()` method to dispatch the action

- `src/app/features/focus-mode/focus-mode-main/focus-mode-main.component.html`:
  - Added reset cycles button next to the complete session button

## Testing

Verified the translation key already exists and is used consistently across the break component and pomodoro settings dialog.

```bash
git diff --stat HEAD~1
 src/app/features/focus-mode/focus-mode-main/focus-mode-main.component.html | 12 ++++++++++++
 src/app/features/focus-mode/focus-mode-main/focus-mode-main.component.ts  |  6 ++++++
 2 files changed, 18 insertions(+)
```

### DIFF
--- a/src/app/features/focus-mode/focus-mode-main/focus-mode-main.component.html
+++ b/src/app/features/focus-mode/focus-mode-main/focus-mode-main.component.html
@@ -205,6 +205,18 @@
               <mat-icon>done_all</mat-icon>
             </button>
           }
+
+          <!-- Reset cycles button (In Progress state, Pomodoro only) -->
+          @if (isShowCompleteSessionButton() && isPomodoro()) {
+            <button
+              mat-icon-button
+              [matTooltip]="T.F.FOCUS_MODE.RESET_CYCLES | translate"
+              (click)="resetCycles()"
+              class="reset-cycles-btn"
+            >
+              <mat-icon>restart_alt</mat-icon>
+            </button>
+          }
         </div>
       </div>
 
--- a/src/app/features/focus-mode/focus-mode-main/focus-mode-main.component.ts
+++ b/src/app/features/focus-mode/focus-mode-main/focus-mode-main.component.ts
@@ -24,6 +24,7 @@ import {
   completeTask,
   focusModeLoaded,
   pauseFocusSession,
+  resetCycles,
   selectFocusTask,
   setFocusModeMode,
   setFocusSessionDuration,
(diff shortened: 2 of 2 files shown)


johannesjo:
Thanks for this! Not sure if this is the best position ui wise. Its not very visible and people might click it by accident. Also I think the icon is not 100% clear. I think we at least need a mat tooltip. And maybe the button should be next to the sesion done button instead?

Br1an67:
Thanks for the feedback! A few points:

1. **Tooltip** — the button already has `matTooltip` bound to the `RESET_CYCLES` translation key, so that should be covered
2. **Position** — happy to move it next to the session done button. Just to confirm — do you mean the "done_all" button in the same `actions-wrapper` div? I can position it right before/after that button instead of at the end
3. **Icon** — would you prefer a different icon? I used `restart_alt` for consistency with the break screen, but I could use `replay` or `refresh` if that's clearer in this context

Let me know your preferences and I'll update the PR!

johannesjo:
Thanks for the quick response!

1. **Tooltip** — you're right, I missed that. That's covered.
2. **Position** — yes, next to the `done_all` button would be better. Note that `reset-cycles-btn` currently has no CSS rule, and all other buttons in `.clock-wrapper` are absolutely positioned — so you'll need to add styling for it. I'll leave the exact layout approach to you.
3. **Icon** — `restart_alt` is fine, let's keep it consistent with the break screen.

johannesjo:
Thank you very much! <3

PR #6768, opened 2026-03-09
fix(nextcloud-deck): preserve card notes on done transition

### LINKED ISSUE #6757: 🚨 Nextcloud-Deck integration deletes Notes while moving cards to "Done-Stack"
### Version Used

v17.2.4

### Expected Behavior

When marking a Card (SP-Task) as done the corresponding card is moved to "Done-Stack" as it is but with marked as done.

### Current Behavior

When marking a Card (SP-Task) as done the corresponding card is moved to "Done-Stack" marked as done, but notes and/or tasks are lost.

### Steps to Reproduce

- connect SP to NC-Deck
- configure / check all 4 option boxes
- task title template as in Example
- select a Board
- import from stacks: all
- select Done stack
- create a card in NC-Deck with notes and/or tasks
- wait for import in SP
- mark the Task done in SP
- open card in NC-Deck -> the notes/tasks are deleted.

### Console Output

```shell

```

iwilltry42:
## Summary
- Fixes the `updateCard$` PUT request to include `description` and `duedate` fields in the payload
- Previously only `title` and `done` were sent, causing the Nextcloud Deck API to clear the card's notes/tasks when moving to the done stack

## Test plan
- [ ] Connect SP to a Nextcloud Deck board
- [ ] Create a card with notes and/or tasks in Nextcloud Deck
- [ ] Wait for import in SP, then mark the task as done
- [ ] Verify the card in Nextcloud Deck retains its notes and due date after being moved to the done stack

Closes #6757

### DIFF
--- a/src/app/features/issue/providers/nextcloud-deck/nextcloud-deck-api.service.ts
+++ b/src/app/features/issue/providers/nextcloud-deck/nextcloud-deck-api.service.ts
@@ -73,7 +73,12 @@ export class NextcloudDeckApiService {
     boardId: number,
     stackId: number,
     cardId: number,
-    changes: Partial<{ title: string; done: boolean }>,
+    changes: Partial<{
+      title: string;
+      description: string;
+      duedate: string | null;
+      done: boolean;
+    }>,
   ): Observable<DeckCardResponse> {
     this._checkSettings(cfg);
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much!

PR #6779, opened 2026-03-10
fix: android system status bar and nav bar cover app ui

billy-thegreat:
Just a one-line change that adds margins at the top and bottom


billy-thegreat:
I will test on an older phone later today after it charges

johannesjo:
Thank you very much!

PR #6782, opened 2026-03-10
Convert selected text to checklist items

### LINKED ISSUE #6015: 💡 Task - Checklist - Convert Text to Checklist
### Problem Statement

I pasted a text into the "Description" field, that contained the release dates of some episodes of a series:
```
Folge 1: 17. Dezember 2025
Folge 2: 24. Dezember 2025
Folge 3: 31. Dezember 2025
Folge 4: 7. Januar 2026
Folge 5: 14. Januar 2026
Folge 6: 21. Januar 2026
Folge 7: 28. Januar 2026
Folge 8: 4. Februar 2026
```

After pasting the text, I marked the whole text and clicked on "Add checklist item".
The whole text is deleted, and an empty checklist item is added.

### Possible Solution

Here the existing text should be converted into a list with multiple checklist items.
Now I have to add "`- [ ] `" in front of every line manually:
```
- [ ] Folge 1: 17. Dezember 2025
- [ ] Folge 2: 24. Dezember 2025
- [ ] Folge 3: 31. Dezember 2025
- [ ] Folge 4: 7. Januar 2026
- [ ] Folge 5: 14. Januar 2026
- [ ] Folge 6: 21. Januar 2026
- [ ] Folge 7: 28. Januar 2026
- [ ] Folge 8: 4. Februar 2026
```

_No response_

### Additional context

_No response_

Br1an67:
Fixes #6015

## Problem

When a user pastes text into a task's description field, selects it, and clicks "Add checklist item", the selected text was being deleted and replaced with a single empty checklist item. The expected behavior is for the selected text to be converted into multiple checklist items (one per line).

## Solution: What PR does

Modified the `toggleChecklistMode()` method in `inline-markdown.component.ts` to detect when text is selected and convert it to checklist items using the existing `applyTaskList()` utility function:

- Added import for `applyTaskList` from `markdown-toolbar.util.ts`
- Added `selectionEnd` tracking to detect text selection
- When text is selected (`selectionStart !== selectionEnd`), the selected lines are converted to checklist items with the `- [ ] ` prefix
- When no text is selected, the original behavior is preserved (insert a new empty checklist item)
- If selected text already has checklist prefixes, they are toggled off (consistent with the fullscreen editor behavior)

Added unit tests for:
- Converting all selected text to checklist items
- Converting partially selected text (only selected lines)
- Toggling checklist prefix when text already has checklist items

## Changes

- `src/app/ui/inline-markdown/inline-markdown.component.ts`: Added text selection handling to `toggleChecklistMode()`
- `src/app/ui/inline-markdown/inline-markdown.component.spec.ts`: Added 3 new tests for the text-to-checklist conversion

## Testing

Verified by running the new unit tests which cover:
1. Full text selection conversion
2. Partial text selection conversion
3. Toggle behavior for existing checklist items

```
 src/app/ui/inline-markdown/inline-markdown.component.spec.ts | 108 +++++++++++++++++++++
 src/app/ui/inline-markdown/inline-markdown.component.ts      |  15 ++-
 2 files changed, 122 insertions(+), 1 deletion(-)
```

### DIFF
--- a/src/app/ui/inline-markdown/inline-markdown.component.spec.ts
+++ b/src/app/ui/inline-markdown/inline-markdown.component.spec.ts
@@ -1300,6 +1300,114 @@ describe('InlineMarkdownComponent', () => {
       // Assert — new item inserted between A and B
       expect(component.modelCopy()).toBe('- [ ] A\n- [ ] \n- [ ] B');
     });
+
+    it('should convert selected text to checklist items', () => {
+      // Arrange
+      const text = 'Folge 1: 17. Dezember\nFolge 2: 24. Dezember\nFolge 3: 31. Dezember';
+      component.model = text;
+      fixture.detectChanges();
+
+      component['isShowEdit'].set(true);
+
+      const mockTextareaEl = {
+        nativeElement: {
+          value: text,
+          selectionStart: 0,
+          selectionEnd: text.length, // all text selected
+          focus: jasmine.createSpy('focus'),
+          setSelectionRange: jasmine.createSpy('setSelectionRange'),
+          style: {},
+          scrollHeight: 100,
+          offsetHeight: 100,
+        },
+      };
+      spyOn(component, 'textareaEl').and.returnValue(mockTextareaEl as any);
+      spyOn(component, 'wrapperEl').and.returnValue({
+        nativeElement: { style: {} },
+      } as any);
+
+      const mockEvent = { preventDefault: () => {}, stopPropagation: () => {} } as any;
+
+      // Act
+      component.toggleChecklistMode(mockEvent);
+
+      // Assert — selected text converted to checklist items
+      expect(component.modelCopy()).toBe(
+        '- [ ] Folge 1: 17. Dezember\n- [ ] Folge 2: 24. Dezember\n- [ ] Folge 3: 31. Dezember',
+      );
+    });
+
+    it('should convert partially selected text to checklist items', () => {
+      // Arrange
+      const text = 'Line 1\nLine 2\nLine 3\nLine 4';
+      component.model = text;
+      fixture.detectChanges();
+
+      component['isShowEdit'].set(true);
+
+      // Select "Line 2\nLine 3"
+      const selectionStart = 7; // after "Line 1\n"
+      const selectionEnd = 20; // before "\nLine 4"
+
+      const mockTextareaEl = {
+        nativeElement: {
+          value: text,
+          selectionStart,
+          selectionEnd,
+          focus: jasmine.createSpy('focus'),
+          setSelectionRange: jasmine.createSpy('setSelectionRange'),
+          style: {},
+          scrollHeight: 100,
+          offsetHeight: 100,
+        },
+      };
+      spyOn(component, 'textareaEl').and.returnValue(mockTextareaEl as any);
+      spyOn(component, 'wrapperEl').and.returnValue({
+        nativeElement: { style: {} },
+      } as any);
+
+      const mockEvent = { preventDefault: () => {}, stopPropagation: () => {} } as any;
+
+      // Act
+      component.toggleChecklistMode(mockEvent);
+
+      // Assert — only selected lines converted
+      expect(component.modelCopy()).toBe('Line 1\n- [ ] Line 2\n- [ ] Line 3\nLine 4');
+    });
+
+    it('should toggle checklist prefix when selected text already has checklist items', () => {
+      // Arrange
+      const text = '- [ ] Item 1\n- [ ] Item 2\n- [ ] Item 3';
+      component.model = text;
+      fixture.detectChanges();
+
+      component['isShowEdit'].set(true);
+
+      const mockTextareaEl = {
+        nativeElement: {
+          value: text,
+          selectionStart: 0,
+          selectionEnd: text.length,
+          focus: jasmine.createSpy('focus'),
+          setSelectionRange: jasmine.createSpy('setSelectionRange'),
+          style: {},
+          scrollHeight: 100,
+          offsetHeight: 100,
+        },
+      };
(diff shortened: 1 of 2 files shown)


johannesjo (commented):
Thank you very much for this!


## Review

Nice PR — clean approach reusing the existing `applyTaskList()` utility instead of reimplementing the conversion logic. The core behavior is correct and the three test cases cover the main scenarios well.

### Suggestions

**1. Extract test helper to reduce boilerplate**

All three new tests repeat ~20 lines of identical mock setup. A small helper would cut the test code roughly in half and make the intent clearer:

```ts
function setupMockTextarea(
  component: InlineMarkdownComponent,
  fixture: ComponentFixture<InlineMarkdownComponent>,
  text: string,
  selectionStart: number,
  selectionEnd: number,
): void {
  component.model = text;
  fixture.detectChanges();
  component['isShowEdit'].set(true);

  const mockTextareaEl = {
    nativeElement: {
      value: text,
      selectionStart,
      selectionEnd,
      focus: jasmine.createSpy('focus'),
      setSelectionRange: jasmine.createSpy('setSelectionRange'),
      style: {},
      scrollHeight: 100,
      offsetHeight: 100,
    },
  };
  spyOn(component, 'textareaEl').and.returnValue(mockTextareaEl as any);
  spyOn(component, 'wrapperEl').and.returnValue({
    nativeElement: { style: {} },
  } as any);
}
```

**2. Minor: `isChecklistMode` after toggle-off**

When the selected text already has `- [ ] ` prefixes, `applyTaskList` removes them (converting to `- Item`). But `isChecklistMode` is still set to `true` afterward. This is pre-existing behavior in the no-selection path too, so not a regression — just worth noting for a future follow-up.

---

Overall this looks good — the implementation is clean and correctly integrates into the existing control flow. 👍

johannesjo:
Thanks — this is a real, wanted fix (#6015), and the approach is right: reusing the existing `applyTaskList` util in the inline editor's checklist button is exactly the minimal/DRY way to do it, and the unit tests are thorough.

The only blocker is that the branch is now stale and conflicts with #8156's refactor of `inline-markdown.component.ts` — it's an import-line conflict (master added `handleListKeydown`, so the resolution is `import { handleListKeydown, applyTaskList }`) plus the surrounding `toggleChecklistMode` was restructured. Please rebase onto master and re-slot the selection branch into the new method.

My earlier suggestion to extract a `setupMockTextarea` spec helper still stands but is non-blocking. After the rebase this should be good to go. (The red CI is an unrelated flaky e2e, not your change.)

Zoriot:
I can work on that. Seems easy to resolve the remaining things and open a new pr

johannesjo:

Thanks for this, and sorry it has sat so long. The approach is still the right one – reusing `applyTaskList()` for the inline editor's checklist button rather than reimplementing the conversion. The first section below is about what changed on master since March rather than about your code.

### Master has moved on (2691 commits behind)

The branch is now `CONFLICTING`, and half of what the PR description describes has since been fixed independently:

- **The "selected text is deleted" half is already gone.** #8156 (shipped in v18.10.0) changed the early return in `toggleChecklistMode` from `if (this.isDefaultText())` to `if (this.isDefaultText() && (!currentText || isUnmodifiedDefault))`, so pasted-but-unsaved notes are no longer wiped. What is still missing on master is the actual conversion: with a selection, master inserts a single empty `- [ ] ` after the cursor's line, which is exactly what this PR fixes.
- **Your change still slots in almost unchanged.** Apart from that one guard, `toggleChecklistMode` on master is identical to the version you branched from, and `applyTaskList` still has the same signature. The import is the only real conflict: master added `import { handleListKeydown } from './markdown-toolbar.util';`, so the resolution is `import { handleListKeydown, applyTaskList } from './markdown-toolbar.util';`.

### Should fix before merge

- **`src/app/ui/inline-markdown/inline-markdown.component.spec.ts`** – none of the three new tests run with `isDefaultText` true, which is the precondition in #6015: a task with no saved notes (`[isDefaultText]="!task().notes"`, `task-detail-panel.component.html:264`) where the pasted text lives only in the textarea, because `changed` is emitted on blur and the toolbar button's `(mousedown)="$event.preventDefault()"` means clicking it never blurs. On this branch that case hits the unconditional early return at `inline-markdown.component.ts:337` and still replaces everything with `- [ ] ` – it is master's newer guard that makes your selection branch reachable at all. After the rebase, please add a test with `isDefaultText` true, a non-empty textarea value and a full selection: that is the regression test for the issue itself, and right now nothing covers it.

### Nits

- **`src/app/ui/inline-markdown/inline-markdown.component.ts:364`** – only `result.selectionEnd` is kept, and `_setTextareaState` collapses it via `setSelectionRange(pos, pos)`, so the converted block ends up deselected. `applyTaskList` returns `selectionStart` and `selectionEnd` precisely so the caller can restore the range, and `dialog-fullscreen-markdown.component.ts:488` does restore both. Keeping the range would make a second click toggle the same lines back instead of falling into the insert-empty-item path.

The red CI on this head is from March and its logs have expired, so it will need a fresh run after the rebase either way.

---

johannesjo:
Closing this one — not because the idea was wrong, but because the branch has moved out of reach.

Thanks @Br1an67, the approach here was right and it's still the behaviour I want. What happened since: the branch is now ~2700 commits behind and conflicting, and half of what it fixed landed separately — #8156 (v18.10.0) stopped selected/pasted notes being deleted. What's still missing is the actual conversion: master still inserts a single empty `- [ ] ` instead of turning the selected lines into checklist items.

@Zoriot offered to redo this against current master — please go ahead, that's the cleanest path from here. The remaining work is small: `applyTaskList` already exists and its signature hasn't changed, so it's essentially re-slotting the call in `toggleChecklistMode`.

Reopening is fine if anyone would rather rebase this branch instead.

PR #6784, opened 2026-03-10
Feat/automations 6453

### LINKED ISSUE #6453: 💡 Automation Rules
### Problem Statement

I have been trying the automation script and I find it very usefully. However, I find missing the following functionality:

* Action- Assign task to project: For example, in my work calendar, I always use prefixes ICARUS: before the meeting name. It would be great if Super productivity could already assign a project based on a partial title match at task creation. (support of regex would be great)

* Action-Delete tasks:   It would be great to be able to automatically delete tasks. For example in my calendar, I always create an event called _BUSY_ to indicate the coworkers that I am out of office at certain times. It would be great if SP could automatically discard these events by name matching.

### Possible Solution

as described

as described

### Additional context

as described

LociStar:
## Problem

Closes #6453

The Automations plugin was missing several useful rule-building features:
- moving tasks to a project
- deleting tasks
- regex support for title matching
- a "title starts with" condition

## Solution

This PR adds those missing automation capabilities and updates the plugin bridge where needed to support them.

### App / bridge changes
- enhance `plugin-bridge.service` so `updateTask` can reassign tasks to projects
- support regex-based condition input

### Automations plugin changes
- add `moveToProject` action
- add `deleteTask` action
- add `titleStartsWith` condition
- add regex toggle support for title-based conditions

## Outcome

Users can now create automation rules that:
- move matching tasks to another project
- delete the task that triggered the rule
- filter title conditions using regex
- match tasks by title prefix

### DIFF
--- a/packages/plugin-dev/automations/src/core/actions.ts
+++ b/packages/plugin-dev/automations/src/core/actions.ts
@@ -36,7 +36,7 @@ export const ActionAddTag: IAutomationAction = {
       return;
     }
     const tags = await ctx.dataCache.getTags();
-    let tagId = tags.find((t) => t.title === value)?.id;
+    let tagId = tags.find((t) => t.id === value || t.title === value)?.id;
 
     if (!tagId) {
       ctx.plugin.log.warn(`[Automation] Tag "${value}" not found.`);
--- a/packages/plugin-dev/automations/src/core/conditions.ts
+++ b/packages/plugin-dev/automations/src/core/conditions.ts
@@ -2,7 +2,7 @@ import { Condition } from '../types';
 import { IAutomationCondition } from './definitions';
 
 const MAX_REGEX_PATTERN_LENGTH = 200;
-const DANGEROUS_REGEX_PATTERN = /(\([^)]*[+*][^)]*\))[+*]/;
+const DANGEROUS_REGEX_PATTERN = /(\([^)]*[+*{][^)]*\))[+*{]/;
 
 const matchesTitleWithRegex = (
   ctx: Parameters<IAutomationCondition['check']>[0],


johannesjo:
@LociStar Thank you very much for this! Since this is marked as WIP: Should I do a review now or rather later?

LociStar:
@johannesjo Thanks for the fast response. I had to correct one part of my code. 
Now it is ready for review 👍🏻

(The blugin-bridge.service.spec has a todo: "Fix plugin tests after stabilizing task model changes" so at the moment it has no tests and I am not sure what this todo is so I was not able to add tests there)

LociStar:
Is the Claude Code workflow something I need to fix on my end?

johannesjo:
> Is the Claude Code workflow something I need to fix on my end?

No, no worries. Thank you! I look into the  changes.

johannesjo (commented):
Thank you very much for this! This looks solid – good test coverage, clean structure, and nice UI touches like the regex toggle with inline validation. Here's my review:

## Should be fixed before merge

### 1. ReDoS risk with user-supplied regex (`conditions.ts`)

`new RegExp(pattern, 'i')` is called with arbitrary user input. Patterns like `(a+)+$` can cause catastrophic backtracking and hang the event loop. The UI validates syntax but doesn't guard against this. Consider adding a timeout, a safe-regex check, or at minimum capping pattern length.

### 2. Project matching by title is fragile (`actions.ts:38`, `ActionDialog.tsx`)

```ts
const project = projects.find((p) => p.title === value || p.id === value);
```

Project titles are user-editable and can be duplicated. The `<select>` in `ActionDialog` uses `p.title` as the option value:

```tsx
<option value={p.title}>{p.title}</option>
```

If two projects share the same title, the first match wins silently. This should use `p.id` as the value instead:

```tsx
<option value={p.id}>{p.title}</option>
```

And then the action should look up by ID primarily.

### 3. Dynamic `import('rxjs')` in bridge (`plugin-bridge.service.ts`)

```ts
const { firstValueFrom } = await import('rxjs');
```

`rxjs` is already imported statically elsewhere in this file. No need for a dynamic import — just add `firstValueFrom` to the existing static imports at the top.

## Nice to improve

### 4. `projects?: any[]` → use a proper type (`ActionDialog.tsx:10`)

At minimum `{ id: string; title: string }[]`.

### 5. `changes?: any` in `TaskEvent` (`types.ts:14`)

`Record<string, unknown>` or a proper partial type would be better than `any`.

### 6. Debug logging is noisy (`automation-manager.ts:150-164`)

The conditional logging for non-matching triggers/conditions looks like leftover debugging. The `if (event.type === 'taskUpdated' && rule.trigger.type === 'taskUpdated')` guard is oddly specific. I'd either remove these or make them all `debug` level consistently.

### 7. `regexError()` called twice in template (`ConditionDialog.tsx:297-300`)

```tsx
{regexError() && (
  <small>Invalid regex: {regexError()}</small>
)}
```

In Solid.js each call creates a separate reactive computation. Consider storing the result or using `<Show>`.

### 8. Mock uses `any` to sidestep types (`actions.spec.ts:14`)

`let mockPlugin: any` was changed from `PluginAPI` — this suggests `deleteTask` and `moveTaskToProject` should be added to the plugin API type definition so tests stay type-safe.

---

Overall this is good work. The core logic is sound, the new actions are properly restricted to task-based triggers, and the fail-closed regex behavior is the right call. Just the project-matching and ReDoS issues need attention before this can be merged. Looking forward to the next iteration!

Lobodiurno:
@LociStar Thank you very much for working on this. I wish that I could do it on my own but I have never programmed something in Typescript. 

I may have  a try to install the development environment and learn the language.

Thanks!

johannesjo:
Thank you very much!

Lobodiurno:
Dear all,

I am testing the automation plugin with the new features and it fails to detect a task matching the Title.

I configured the tasks as follows:

image

The task is imported everyday form Microsoft calendar as follows:

image

I have also tried in the automation plugin, to set the trigger to task updated instead of task created it also does not work.

By the way, the conditions are OR conditions or AND conditions? maybe this could be added in brackets.

If I add manually the task it successes to delete the task.

Thanks

LociStar:
First of @johannesjo thank you for finishing this PR. I had barely any time the last few weeks. Only now over the weekend I have a bit more time.
<br>
<br>
<br>


@Lobodiurno this is not a bug.
The conditions are a logical `AND`. So all conditions have to pass.

This is what happens at the moment:
- `titleContains` (regex): BUSY → matches
- `titleStartsWith`: BUSY → does not match

So your automation wont trigger, because your title does not start with "BUSY".
`OR` conditions is not supported yet.


For your case you now have some options. 
1. Remove `titleStartsWith` and keep only `titleContains`
2. Or use a regex that matches the actual title format


Here some regex examples:

1) First regex

`^\d{1,2}:\d{2}\s*-\s*BUSY$`

This is strict.
It matches only titles that are basically:
- a time like 15:30
- then -
- then exactly BUSY
- and nothing after that

Examples:
- ✅ 15:30 - BUSY
- ✅ 9:00-BUSY
- ❌ 15:30 - BUSY today
- ❌ 15:30 - Very BUSY

2) Second regex

`^\d{1,2}:\d{2}\s*-\s*.*\bBUSY\b`

This is looser.
It matches titles that:
- start with a time
- then -
- then anything
- and somewhere after that contain the word BUSY

Examples:
- ✅ 15:30 - BUSY
- ✅ 15:30 - something BUSY
- ✅ 15:30 - IMPORTANT - BUSY
- ✅ 15:30 - very BUSY today
- ❌ Something else - BUSY because it must start with a time



I hope this will help you :)

Lobodiurno:
Dear @LociStar ,

Thank you very much for your detailed feedback.

I believe that  I have identified the root cause of the issue. In general the plugin works. If click on adding the task manually the software deletes the task. I get the message: Deleted task: BUSY .

image

The problem is when the application starts. After Super Productivity starts, the task BUSY is already placed on the list.

Probably at start up, the import tasks procedure runs before the automation plugin gets loaded and therefore the task BUSY is added and the automation plugin does not get a chance to delete the task.

Regarding the comments with the time. 15:30... It is a calendar task with a scheduled time. The name of the task is simply "BUSY" not "15:30 BUSY" therefore the regex that you provided does not work. This is not important, as simply with titlecontains: BUSY works.

Thank you very much for your help

LociStar:
Hello @Lobodiurno,

Oh I see. Thank you for the information! 
I will try to have a deeper look at the plugin system to try and figure out if this can be changed, or if there might be a workaround for it.

Best regards 
LokiStar

PR #6786, opened 2026-03-10
fix: make sub-tasks visually distinct in search results

### LINKED ISSUE #6785: 💡 Search results: make sub-tasks more distinguishable from main tasks
### Problem Statement

I have this task which has 2 subtasks:
Image

When I enter a certain keyword in search, I get these results:
Image

The first result is the main task, the second is 1 of 2 subtasks of this main task.

In the result list,
1. one can not distinguish main tasks and sub tasks. This would be helpful.
2. one can not see that a task is already completed (but I remember there is another feature request for that)

### Possible Solution

In the result list, make main tasks and sub tasks distinguishable. Somehow. I am not a designer, so I cannot give helpful hints, but the current state is not very helpful. :-/ Maybe display sub tasks in italic, or smaller font, or ...

none

### Additional context

_No response_

VedantMadane:
Fixes #6785

### DIFF
--- a/src/app/pages/search-page/search-page.component.html
+++ b/src/app/pages/search-page/search-page.component.html
@@ -38,10 +38,8 @@
                 }
                 <div
                   class="task-title"
-                  [ngClass]="{
-                    'task-archived': item.isArchiveTask,
-                    'task-subtask': !!item.parentId,
-                  }"
+                  [class.task-archived]="item.isArchiveTask"
+                  [class.task-subtask]="!!item.parentTitle"
                 >
                   {{ item?.title }}
                 </div>
--- a/src/app/pages/search-page/search-page.component.ts
+++ b/src/app/pages/search-page/search-page.component.ts
@@ -19,7 +19,7 @@ import { TagService } from '../../features/tag/tag.service';
 import { Task } from '../../features/tasks/task.model';
 import { SearchItem } from './search-page.model';
 import { NavigateToTaskService } from '../../core-ui/navigate-to-task/navigate-to-task.service';
-import { AsyncPipe, NgClass } from '@angular/common';
+import { AsyncPipe } from '@angular/common';
 import { MatIcon } from '@angular/material/icon';
 import { MatIconButton } from '@angular/material/button';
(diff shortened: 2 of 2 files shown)


johannesjo:
Thanks for the contribution! I reviewed the changes and have some feedback:

## Critical: CSS scoping issue

The new styles for `.subtask-icon`, `.subtask-breadcrumb`, and `.task-subtask` are placed **outside** the `&::ng-deep` block in the SCSS. The existing styles for `.search-result-content`, `.task-title`, etc. are wrapped in `::ng-deep` because `mat-list-item` encapsulates its projected content. Without `::ng-deep`, Angular's view encapsulation will prevent the new styles from reaching the DOM elements inside the Material list item.

**Fix:** Move the subtask styles inside the existing `&::ng-deep` block, or wrap them in their own `::ng-deep`.

## Important

- **Excessive CSS for a simple feature** — 27 lines of new SCSS with hardcoded values (`1.5rem`, `0.85rem`, `12rem`, `0.95em`) and mixed units (`rem` vs `em`). The project guidelines say "Keep CSS, nesting, and classes to a minimum." Consider simplifying — the italic styling + icon may be sufficient without the breadcrumb `max-width`, separate font sizes, etc.
- **No mobile responsive handling** — The breadcrumb with `max-width: 12rem` and extra padding could cause layout issues on narrow screens. The existing `@media (max-width: 768px)` block doesn't account for the new elements.

## Minor

- The hardcoded `›` character in the template isn't translatable and may render inconsistently across fonts. Consider using a Material icon or CSS pseudo-element instead.
- Test assertions only check the data model (`parentTitle` value). No rendering tests verify the breadcrumb/icon actually appear in the DOM.

## What's good

- Reuses the existing `parent` lookup — no extra computation needed.
- Model change is clean and minimal.
- `parent?.title ?? null` correctly handles missing parents.

Could you add a screenshot or recording showing this working? That would help verify the visual result, especially given the CSS scoping concern.

johannesjo:
Excellent. Like it quite a bit! Thank you very much! Sorry for the reviewing delay.

Issue #6432, opened 2026-02-08
🚨 Permanent sorting not working


novikov1337danil:
### Version Used

17.1.5

### Expected Behavior

Permanent sorting should work as before.

### Current Behavior

Permanent sorting is broken:
- the selected sort type is not marked
- reverse sorting does not work

### Steps to Reproduce

1. Select any permanent sorting type <img width="967" height="535" alt="Image" src="https://github.com/user-attachments/assets/fc48f6c8-945d-4c0d-b1fc-a541593608de" />

2. This option will not be marked

3 . Click again to reverse sorting - nothing will happen

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
Thank you!

novikov1337danil:
I would like to work on this

@johannesjo assign me

johannesjo:
With pleasure! :)

novikov1337danil:
@johannesjo If you don't mind, I'll remove the separate "Sort (permanent)" option and simply add a save button. I think this approach is more elegant

<img width="594" height="544" alt="Image" src="https://github.com/user-attachments/assets/a40a3a0d-a146-41e9-b3b4-bc824d0db564" />

johannesjo:
Yes! This is an excellent idea @novikov1337danil !

novikov1337danil:
> Permanent sorting should work as before.

Oh, how wrong I was 🥲  I think the bugs I described were always there. It`s just the differences in the implementation of temporary and permanent sorting

Now I remember that all the previous times I fixed something in the temporary sorting/grouping/filters, I always avoided the permanent sorting feature because it seemed too complicated

In any case, I never liked the visual separation between temporary and permanent sorting. It's time to figure this out

johannesjo:
Let me know if I can be of any help!

PR #6838, opened 2026-03-15
refactor: permanent sorting

novikov1337danil:
## Problem

Permanent Sorting did not work clearly - https://github.com/super-productivity/super-productivity/issues/6432

## Solution

Made behavior clear

image

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/util/ls-util.ts
+++ b/src/app/util/ls-util.ts
@@ -1,4 +1,3 @@
-/* eslint-disable prefer-arrow/prefer-arrow-functions */
 /**
  * localStorage utility functions for Super Productivity
  */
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much! Please let me know when it is ready for review!

novikov1337danil:
> Thank you very much! Please let me know when it is ready for review!

Ready

novikov1337danil:
> ### `eslint-disable` scope in `ls-util.ts`
> 
> The file-level `/* eslint-disable prefer-arrow/prefer-arrow-functions */` suppresses the rule for the entire file. Only the `lsGetJSON` overloads need it. A block-level disable/enable would be more precise:

Added a comment explaining the reason for the disabling rule.

> Minor: no user feedback on save
> 
> saveSort() has several early-return paths (no context, no tasks, order unchanged) that silently do nothing. Might be worth a brief snackbar/toast on successful save so the user knows it worked, but this is a UX preference call — not blocking.

Unfortunately, I don't know how this should be handled. Can we skip this? Previously there were no handlers either (except `resetAll` call, which I definitely think was unnecessary).

johannesjo:
Thank you very much! <3

PR #6854, opened 2026-03-16
fix(redmine): fix logging time to redmine

### LINKED ISSUE #6852: 💡Redmine integration: Cumulative time entry created on each task completion instead of delta
## Problem Statement

When using the Redmine integration and marking a task as done, Super Productivity prompts to log time to Redmine — which works. However, if I later **reopen the task** (unmark as done), work on it some more, and then **mark it done again**, SP creates a **new Redmine time entry with the full cumulative tracked time**, not just the time spent since the last log.

This results in **duplicate/inflated time entries** in Redmine.

## Steps to Reproduce

1. Connect a Redmine project and import an issue as a task
2. Start the timer, work for **1 hour**, stop the timer
3. Mark the task as done → SP prompts to log time to Redmine → confirm → **1h logged to Redmine** ✅
4. Reopen the task (unmark as done)
5. Start the timer again, work for **30 minutes**, stop
6. Mark the task as done again → SP prompts to log time → confirm → **1h30m logged to Redmine** ❌

## Expected Behavior

In step 6, SP should log only **0h30m** (the delta since the last Redmine time entry), not the full 1h30m cumulative total.

Alternatively, SP could:
- Track which time has already been synced to Redmine and only send the difference
- Allow the user to edit the hours in the prompt before submitting
- Show a warning that time was previously logged for this task

## Current Behavior

SP sends the **total `timeSpent`** value as a new `POST /time_entries.json` to Redmine every time the task is marked done. It does not account for previously logged entries, resulting in **overstated spent time** in Redmine.

## Impact

This makes the done-prompt → Redmine sync unusable for tasks that get reopened, which is a very common workflow (e.g., code review feedback, partial work across days, recurring subtasks). Users must manually delete or correct duplicate entries in Redmine.

## Environment

- **SP version**: (current)
- **Platform**: macOS (Electron desktop app)
- **Redmine version**:  6.x
- **Integration**: Redmine issue provider with API key auth

P.S. Thank you for all your work

### Possible Solution

1. **Track synced time per task** — store the amount already logged to Redmine and only send the delta on subsequent completions
2. **Editable time in the prompt** — let the user adjust the hours before the API call is made
3. **Query existing time entries** — before posting, `GET /time_entries.json?issue_id=X&user_id=Y` to calculate what's already logged and subtract it
4. **"Log time" as a separate action** — decouple time logging from the "mark done" flow, so users can explicitly choose when to sync time (similar to how Jira integration handles worklogs)

_No response_

### Additional context

_No response_

syky27:
## Problem     

When logging time to Redmine, the application logs the full cumulative time instead of just the delta (time not yet logged). Additionally, there was no distinction between time logged by the current user and time logged by other users, which could lead to incorrect calculations.

Closes #6852                                                                                                                                                                     
   
##  Solution                                                                                                                                                                         
                                                                            
  - Query existing Redmine time entries filtered to the current user to calculate the delta between total time spent and already-logged time
  - Only the current user's logged time is considered — time logged by other users on the same issue is excluded from the calculation
  - The API service returns hours (Redmine's native unit) and the dialog component handles the conversion to milliseconds

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/issue/providers/redmine/dialog-redmine-track-time/dialog-redmine-track-time.component.ts
+++ b/src/app/features/issue/providers/redmine/dialog-redmine-track-time/dialog-redmine-track-time.component.ts
@@ -196,8 +196,10 @@ export class DialogRedmineTrackTimeComponent implements OnDestroy {
 
     this._cfgOnce$.subscribe((cfg) => {
       if (cfg.timeTrackingDialogDefaultTime) {
+        this.selectedDefaultTimeMode = cfg.timeTrackingDialogDefaultTime;
         this.timeSpent = this.getTimeToLogForMode(cfg.timeTrackingDialogDefaultTime);
         this.started = this._fillInStarted(cfg.timeTrackingDialogDefaultTime);
+        this._cdr.markForCheck();
       }
     });
   }


syky27:
Hi @johannesjo 

  1. Fix race condition 
  2. Fix OnPush change detection
  3. Add catchError on `_myLoggedHours$`
  A failed API call now falls back to 0 hours (via `catchError(() => of(0))`), so the dialog still opens with the full tracked time as a safe default.
  4. spent_hours on `RedmineIssue` — kept intentionally
  This field is actively used in `redmine-issue-content.const.ts` (lines 60–64) to display time spent in the issue detail view. Removing it would break that display.
  
  Would you like me to squash those 2 commits?

johannesjo:
> Would you like me to squash those 2 commits?

Not needed, thank you! I squash before merging :)

syky27:
> spent_hours? is added to RedmineIssue but never used — can be removed if not needed


I am not so sure about this...  It's actively used in redmine-issue-content.const.ts (lines 60-64) for displaying time spent in the issue detail view.

The rest is pushed...

johannesjo:
Sorry! My bad!

johannesjo:
Thank you very much <3

syky27:
> Thank you very much <3

Than you for your collaboration and work on this project. Happy to help.

Cheers

PR #6906, opened 2026-03-21
feat(issue): add default tags and note for issue provider imports

### LINKED ISSUE #6652: 💡 Set default task details for issue provider imports
### Problem Statement

When importing tasks from an issue provider, I almost always add a the same tag and sometimes the same subtasks as well. For example, I have a meeting tag that I add for Google Calendar work meetings, and a development tag that I add for Linear issues. And for Linear issues, I always make a few initial subtasks to investigate, confirm the requirements, open a PR, etc.

### Possible Solution

The issue provider screen already has a Default Super Productivity Project setting that sets the project when creating new imported tasks. Right underneath that, it would make sense to add fields for default tags, default note, and default subtasks. The Edit Recurring Task Config dialog already has those fields (except for subtasks, which may be more complicated to implement).

Manually setting the same thing over and over

### Additional context

I was meaning to take this on but haven't had the time yet. Jotting it down here and hopefully I'll be able to come back to it.

costajohnt:
## Problem

When importing tasks from issue providers (Jira, GitLab, GitHub, etc.), there's no way to automatically apply tags or notes to the imported tasks. Users have to manually tag each imported task, which is tedious when using provider-specific workflows.

Closes #6652

## Solution

Added two new optional fields to the issue provider settings:

- **Default tags** — a chip-list input (same pattern as the recurring task config dialog) that lets users select tags to automatically apply to all tasks imported from this provider
- **Default note** — a textarea for a note that gets applied to all imported tasks

Both fields appear in the issue provider settings dialog and their values are applied in `IssueService.addTaskFromIssue()` when tasks are imported. Default tags are merged with any context-based tags (avoiding duplicates via `unique()`).

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/issue/issue.service.ts
+++ b/src/app/features/issue/issue.service.ts
@@ -521,7 +521,9 @@ export class IssueService {
         .getCfgOnce$(issueProviderId, issueProviderKey)
         .toPromise();
       const defaultProjectId = providerCfg.defaultProjectId;
-      const defaultTagIds = providerCfg.defaultTagIds || [];
+      const defaultTagIds = (providerCfg.defaultTagIds || []).filter(
+        (id) => id !== TODAY_TAG.id,
+      );
       const defaultNote = providerCfg.defaultNote;
 
       if (typeof this._workContextService.activeWorkContextId !== 'string') {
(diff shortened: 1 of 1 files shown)


johannesjo (commented):
a

PR #6935, opened 2026-03-23
fix(planner): keep schedule task calendar at consistent height

### LINKED ISSUE #6556: 🚨 When clicking through months in the calendar pop-up, the arrows move up and down leading to missed clicks
### Version Used

v17.1.8

### Expected Behavior

In the calendar pop-up for setting the due date, if I want to go several months into the future, I would expect to be able to rapidly click multiple times. e.g. if it's February now and I want to go to December I would click 10 times on the ">" button to go 10 months into the future.

### Current Behavior

Currently, **the calendar pop-up is always vertically centered but it jumps up and down because the months have different "heights". Because of this, you can't just keep clicking on the ">" button, but instead after each month you may have to go slightly up and down with the mouse in order to click the ">" button**. Very annoying!

Why are the months jumping? This can have 2 reasons. 

The first is because of the different numbers of (full and partial) weeks - for example the February with 28 days can fit in as little as 4 rows (if the 1st is a Monday/Sunday depending on what you set as your start of week), and a month with 31 days will need 5-6 rows depending on the weekday of the 1st. Months with fewer "rows" in the calendar will need less vertical space. In my screenshot below all 3 months need 5 rows, but it can be 4 or 6 rows in some cases too.

The second is that some months seem to start with an empty row for some reason. e.g. in my screenshot the September has "SEP" written above the dates while October has "OCT" in the same row as the first four days. This causes different vertical space use even though both months need the same amount of rows (5) to display all days. This seems to be a visual bug, but not important on its own.

Image

The actual problem here is not the different vertical space but that the calendar pop-up is always vertically centered. Instead, the top of the calendar pop-up, and therefore the "<" and ">" buttons too, should be at a fixed position and then just get longer or shorter at the bottom. This would make it possible to just spam clicks on the ">" button to quickly go to a future month.

### Steps to Reproduce

1. Open web app or desktop app
2. Edit a "Planned For" date for a task
3. Click on the ">" button for next months (or "<") several times 

### Console Output

```shell

```

Diogo-22:
Closes #6556

## Problem

The schedule task dialog can shift vertically when calendar content changes, which creates a jumpy experience while picking dates.

## Solution

Set a fixed height on the embedded mat-calendar in the schedule task dialog to keep layout stable and improve usability. 

## Type of Change

- [x] Bug fix

## Checklist

### DIFF
--- a/src/app/features/planner/dialog-schedule-task/dialog-schedule-task.component.scss
+++ b/src/app/features/planner/dialog-schedule-task/dialog-schedule-task.component.scss
@@ -23,6 +23,10 @@
   }
 
   ::ng-deep {
+    mat-calendar {
+      height: 400px;
+    }
+
     .mat-calendar-header {
       padding-top: 0;
     }


johannesjo:
Thank you very much!

Issue #6723, opened 2026-03-04
💡 The time under focus mode icon: After a break auto-completes, it should show 0:00


beerkumquatpome:

shaked-shai:
I would like to work on it

shaked-shai:
I have a suggestion, maybe instead of change it completely, I will add a setting, so the user can decide if they want to see the total time or just 0:00

johannesjo:
Thank you @shaked-shai ! Maybe we should discuss what a reasonable default is here. I don't want to overwhelm users with too many settings.

shaked-shai:
Sure, I do see the OP point of when the user finishes the focus time, it is a bit confusing of what this time is about, on the other end when a user knows what it is about, it is a nice little motivation boost of how much time they have been focusing on work (at least in my opinion)

shaked-shai:
@johannesjo I'm not sure to what route I should take that issue

johannesjo:
I think showing 0:00 sounds like a reasonable solution for me as it has the least potential to be confusing.

PR #6939, opened 2026-03-23
feat(focus): change time under focus icon after a break to turn zero

shaked-shai:
## Problem

After a break auto-completes, it shows total focused time instead of 0:00
This causes confusion

https://github.com/super-productivity/super-productivity/issues/6723

## Solution

changed the time under the focus icon to show zero after a brake instead of the total time

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [V] Existing tests still pass
- [V] My commit messages follow the Angular format (`type(scope): description`)

### DIFF
--- a/src/app/core-ui/main-header/focus-button/focus-button.component.html
+++ b/src/app/core-ui/main-header/focus-button/focus-button.component.html
@@ -25,9 +25,5 @@
     <div class="focus-label focus-running-label">
       {{ runningMs | msToMinuteClockString }}
     </div>
-  } @else {
-    <div class="focus-label">
-      {{ 0 | msToMinuteClockString }}
-    </div>
   }
 </div>


johannesjo:
Thank you for the PR and for discussing the approach in the issue first! 

I think we can simplify this even further — instead of showing a hardcoded `0:00`, we can just remove the `@else` branch entirely so no label is shown when no session is running. This avoids the confusion without adding a permanent `0:00` under the icon.

Could you update the PR to just remove the `@else if` block (lines 28-31) so only the `@if (runningTimeMs())` branch remains?

johannesjo:
Thank you!

PR #6964, opened 2026-03-25
Prevent multiple empty subtask drafts per parent task

### LINKED ISSUE #6956: 🚨 Prevent multiple empty subtasks created via Add Sub Task / `A` shortcut
### Version Used

current Windows version

### Expected Behavior

Creating multiple empty subtasks should not be possible.

Only one empty subtask (draft) should exist at a time. If an empty subtask already exists:

- Pressing the "A" shortcut should not create another subtask
- Clicking "Add Sub Task" should not create another subtask (or should be disabled)

A subtask should only persist once non-empty text is entered.

If the empty subtask is left empty and the user navigates away, it should be discarded.

### Current Behavior

Creating multiple empty subtasks should not be possible.

Only one empty subtask (draft) should exist at a time. If an empty subtask already exists:

- Pressing the "A" shortcut should not create another subtask
- Clicking "Add Sub Task" should not create another subtask (or should be disabled)

A subtask should only persist once non-empty text is entered.

If the empty subtask is left empty and the user navigates away, it should be discarded.

### Steps to Reproduce

1. Open any task
2. Press "A" to create a subtask
3. Do not enter any text
4. Press "A" again
5. Repeat a few times

OR

1. Open any task
2. Click "Add Sub Task"
3. Do not enter any text
4. Click "Add Sub Task" again
5. Repeat

### Console Output

```shell

```

steveszat:
## Summary

This fixes #6956 by preventing multiple empty subtask drafts from being created for the same parent task.

## Changes

- Prevent creating a second empty subtask draft if the same parent task already has one
- Delete an abandoned empty subtask draft instead of persisting it
- Add focused service-level tests for the new behavior

## Notes

This change is intentionally minimal and limited to the behavior described in the issue.

It does not redesign the subtask creation flow or change unrelated programmatic subtask creation paths.

Local pre-push ran into unrelated timezone test failures on my machine, but the targeted task service test for this change passed.

### DIFF
--- a/src/app/features/tasks/task.service.spec.ts
+++ b/src/app/features/tasks/task.service.spec.ts
@@ -326,6 +326,39 @@ describe('TaskService', () => {
         }),
       );
     });
+
+    it('should delete an abandoned empty subtask draft instead of updating it', () => {
+      const emptySubTask = createMockTask('subtask-1', {
+        title: '',
+        parentId: 'parent-1',
+      });
+      store.setState({
+        tasks: {
+          ids: ['task-1', 'task-2', 'subtask-1'],
+          entities: {
+            ['task-1']: createMockTask('task-1'),
+            ['task-2']: createMockTask('task-2'),
+            ['subtask-1']: emptySubTask,
+          },
+          currentTaskId: null,
+          selectedTaskId: null,
+          taskDetailTargetPanel: null,
+          isDataLoaded: true,
+        },
+      });
+      store.refreshState();
+
+      service.update('subtask-1', { title: '   ' });
+
+      expect(store.dispatch).toHaveBeenCalledWith(
+        TaskSharedActions.deleteTask({
+          task: {
+            ...emptySubTask,
+            subTasks: [],
+          },
+        }),
+      );
+    });
   });
 
   describe('updateTags', () => {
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! I've tested this and it doesn't feel quite right to me tbh. Now the shortcut just does nothing which feels confusing to me (even knowing the intention). I think it would be better to either refocus the existing empty sub task or to leave it as is. I lean a bit towards towards the latter since it might be not always ideal, but it feels more in control, since the action always does reliably the same.

steveszat:
Thank you for your feedback. Here's what I propose for a solution:

- If an empty subtask already exists for a parent task, “Add Subtask” in the menu would be disabled for that task.
- If the user presses the “A” shortcut in that same state, instead of doing nothing or creating another empty subtask, focus would move to the existing empty subtask and activate title editing.
- If the user exits editing and the subtask is still empty, it would be automatically removed.
- A subtask would only become persistent once it has a non-empty title.

What are your thoughts?

johannesjo:
Thanks for the proposal! I think we should keep it simpler though. The core fix I'd go for is: when the user presses "A" or clicks "Add Subtask" and an empty subtask already exists for that parent, just refocus that existing subtask and activate title editing. That way the action always does something visible and feels responsive.

I'd skip disabling the menu item (users won't know why it's grayed out and that is confusing) and the auto-deletion / draft persistence logic – it adds quite a bit of complexity for a minor edge case and changes how subtasks work conceptually. Let's keep it simple: one empty subtask per parent, re-pressing "A" focuses it.

steveszat:
Thanks for your rapid response. I think it may help if I explain how I arrived at this in the first place.

I was using the app as a user and somehow ended up with an empty subtask.

When I looked into it further, I realized it was possible to create multiple empty subtasks and have them persist as I navigated to other tasks.

I assumed that was unexpected behavior, mainly because I’ve never seen that behavior in other apps, and it didn’t seem like something a user would intentionally want.

My expectation was that an empty subtask would behave more like temporary draft state — something that exists while editing, but disappears if left empty.

That’s what led me to think this might be low-hanging fruit and worth offering to help with.

My concern with reusing an existing empty subtask is that it can create an unexpected result for the user: they try to add a subtask, but no obvious new subtask appears.

The behavior I had in mind was:

If an empty subtask already exists for a parent task, Add Subtask would focus that subtask and activate editing.
If the user leaves editing and the subtask is still empty, it would be removed automatically.
A subtask would only become persistent once the user enters a title.

That seems closer to conventional behavior and would avoid the confusion I ran into while testing.

The main thing I’m trying to understand is whether you want empty subtasks to persist at all, or whether you’d prefer to leave the current behavior as-is.

johannesjo:
Good question! For now, empty subtasks persisting is fine – it's an edge case and not worth the added complexity of draft/auto-delete semantics right now.

The key fix is simpler: when the user presses "A" or clicks "Add Subtask" and an empty subtask already exists for that parent, refocus that subtask and activate title editing. That way the action always does something visible. The user sees the cursor jump to the existing empty field – no confusion, no duplicates.

So concretely:
1. In `addSubTaskTo`: if an empty subtask exists for that parent, call `_focusNewlyCreatedTask(existingId, true)` and return that ID
2. Remove the `update()` deletion logic – let's not change what `update()` does

That should be a small diff and addresses the core problem. We can always revisit auto-cleanup as a separate enhancement later if needed.

johannesjo:
Thanks for the work on #6956. I'd suggest closing this one as superseded: master now handles this via `_addSubTaskOrFocusEmpty` + `focusTaskById` (#8081), which focuses an existing empty sub-task instead of creating a duplicate — the behavior I was after.

Two things from this PR I'd also want to avoid carrying over:

- `addSubTaskTo` early-returns the existing draft id **without** focusing it, so the shortcut appears to do nothing (the confusing no-op I flagged earlier).
- The `update()` override dispatches `deleteTask` when the title is empty. `update()` is a general setter called from ~26 sites, so clearing a sub-task title would silently delete it — too broad a side effect.

It's also stale/conflicting against master and has a leftover `console.log(task)`. If there's still a residual gap (e.g. the "A" shortcut path), a small fresh PR against current master routing through `_addSubTaskOrFocusEmpty` would be the cleanest path. Thanks again.

johannesjo:
Superseded rather than rejected — see #6956 for the detail.

The empty-subtask problem was fixed by the add-subtask input rework, which opens an inline draft instead of creating the task entity up front, so the code this PR touches has moved underneath it. Your diagnosis was correct; the fix just arrived by a different route. Sorry this sat open so long.

This can be closed.

PR #6965, opened 2026-03-26
fix(nav): show task count in menu on touch devices #5325

### LINKED ISSUE #5325: 🚨 The number of tasks do not appear on the menu
### Version Used

15.2.16 android mobile and tablet

### Expected Behavior

The number of tasks in a project (but not in a folder) or a tag should be visible on the left panel on the phone and on the tablet

### Current Behavior

I cannot see the number of tasks on the phone but on the tablet they are hardly visible.

Image
Image

### Steps to Reproduce

none

### Console Output

```shell

```

disha7555:
## Problem

The number of tasks did not appear in the menu on touch devices.
On mobile/tablet screens, the task count was hidden due to hover-only styles.

Fixes #5325

## Solution

Updated the task-count styles to work on devices without hover support.
- Allow task count to be visible on touch devices
- Adjust right spacing to avoid overlap with the additional menu button
- Remove mobile-specific display:none rules that prevented the count from showing

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/core-ui/magic-side-nav/nav-item/nav-item.component.scss
+++ b/src/app/core-ui/magic-side-nav/nav-item/nav-item.component.scss
@@ -69,8 +69,10 @@
   font-weight: normal;
 }
 
-:host:hover .task-count {
-  display: flex !important;
+@media (hover: hover) {
+  :host:hover .task-count {
+    display: none !important;
+  }
 }
 
 @media (hover: none) {
@@ -79,6 +81,7 @@
     visibility: visible;
   }
   .task-count {
+    right: 48px !important;
     display: flex !important;
   }
 }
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you! This works well for mobile touch, but I think we should keep it as is on desktop devices. Looks off this way.

disha7555:
> Thank you! This works well for mobile touch, but I think we should keep it as is on desktop devices. Looks off this way.

Thanks for the feedback! I’ll update the styles to preserve desktop behavior while fixing mobile.

disha7555:
Pushed the changes - task count now works on mobile and hover is preserved on desktop.

johannesjo:
Perfect! Thank you very much!

PR #6993, opened 2026-03-28
fix(schedule): show repeat projections for future dates with same weekday

sahajagrawal18-droid:
## Issue
This PR fixes a bug where recurring tasks were missing from the schedule view when viewing a future date that fell on the same day of the week as today.

## Solution
The logic in createSortedBlockerBlocks previously assumed that if we were looking at "today," we should skip the first projection (index 0) because the "real" task would already exist. However, it didn't account for when the user navigates to a future date that shares the same weekday.

## Type of Change
- Bug fix

## Checklist
-  I have run `npm run checkFile` on changed `.ts`/`.scss` files
- I have added tests for my changes
- Existing tests still pass
- My commit messages follow the Angular format (`type(scope): description`)


johannesjo:
Thank you very much!

PR #7002, opened 2026-03-29
fix(tasks): prevent duplicate task creation from stale add-task state

miaopan607:
## Problem

Pressing `Enter` again after adding a task could create the same task again because stale add-task parser state was reused even when the visible input was already cleared.

## Solution

Invalidate stale parser state after submit, guard against empty-input resubmits, and add regression tests.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/assets/i18n/ro-md.json
+++ b/src/assets/i18n/ro-md.json
@@ -57,7 +57,7 @@
     },
     "GO_FURTHER": {
       "TITLE": "Descoperă mai mult",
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! <3

PR #7047, opened 2026-03-30
chore: fix LF/CRLF for errant SCSS file

zenoprax:
.gitattribute added to help reduce the potential of this in the future.


johannesjo:
Thank you!

PR #7059, opened 2026-03-31
fix(tasks): use attachment title as task name when pasting a bare URL

### LINKED ISSUE #6444: 💡 QOL when pasting a link into a task
### Problem Statement

I find sharing to super prod to not work very well (same with other task managers) so I will often just copy a url and paste it in the task. This causes some issues. 

Mainly, the url is turned into an attachment and the task title is left blank. 

### Possible Solution

I think it would be great if it used the meta title of the page to generate the name of the task. You already do this for the attachment name, but should be the task name by default if you don't enter anything else. 

_No response_

### Additional context

_No response_

jaydeep-pipaliya:
## Summary
When pasting a bare URL (e.g. `https://github.com/super-productivity/super-productivity`) into the add-task bar with `urlBehavior: 'extract'`, the URL gets extracted into an attachment but the task title is left blank.

This fix uses the attachment's title (derived from the URL basename via `_baseNameForUrl`) as the task name when the title would otherwise be empty after URL extraction.

**Before:** Pasting `https://example.com/my-cool-page` → task title: `` (empty), attachment: `my-cool-page`
**After:** Pasting `https://example.com/my-cool-page` → task title: `my-cool-page`, attachment: `my-cool-page`

The change is a 4-line addition in `parseUrlAttachments()` inside `short-syntax.ts`.

Fixes #6444

## Test plan
- [ ] Set URL behavior to "extract" in settings
- [ ] Paste a bare URL into the add-task bar
- [ ] Verify the task is created with the URL basename as the title
- [ ] Verify the URL is still added as an attachment
- [ ] Paste a URL with surrounding text → verify only the text remains as title (existing behavior)
- [ ] Paste a markdown link `[title](url)` → verify title is used (existing behavior)

### DIFF
--- a/src/app/features/tasks/short-syntax.spec.ts
+++ b/src/app/features/tasks/short-syntax.spec.ts
@@ -1657,6 +1657,16 @@ describe('shortSyntax', () => {
       expect(r?.taskChanges.title).toBe('my-cool-page');
     });
 
+    it('should use domain basename as task name when pasting a root URL with urlBehavior "extract"', async () => {
+      const t = {
+        ...TASK,
+        title: 'https://example.com',
+      };
+      const r = await shortSyntax(t, { ...CONFIG, urlBehavior: 'extract' });
+      expect(r).toBeDefined();
+      expect(r?.taskChanges.title).toBe('example');
+    });
+
     it('should keep URL in title and add attachment when urlBehavior is "keep-and-attach"', async () => {
       const t = {
         ...TASK,
--- a/src/app/features/tasks/short-syntax.spec.ts
+++ b/src/app/features/tasks/short-syntax.spec.ts
@@ -1645,6 +1645,18 @@ describe('shortSyntax', () => {
       expect(r?.taskChanges.title).toBe('Check for details');
     });
 
+    it('should use attachment title as task name when pasting a bare URL with urlBehavior "extract"', async () => {
+      const t = {
+        ...TASK,
+        title: 'https://example.com/my-cool-page',
+      };
+      const r = await shortSyntax(t, { ...CONFIG, urlBehavior: 'extract' });
+      expect(r).toBeDefined();
+      expect(r?.attachments.length).toBe(1);
+      expect(r?.attachments[0].path).toBe('https://example.com/my-cool-page');
+      expect(r?.taskChanges.title).toBe('my-cool-page');
+    });
+
     it('should keep URL in title and add attachment when urlBehavior is "keep-and-attach"', async () => {
       const t = {
         ...TASK,
--- a/src/app/features/tasks/short-syntax.ts
+++ b/src/app/features/tasks/short-syntax.ts
@@ -556,10 +556,12 @@ const parseUrlAttachments = (
     cleanedTitle = markdownUrls.length > 0 ? titleWithoutMarkdown : task.title;
     cleanedTitle = removeUrlsFromTitle(cleanedTitle, plainUrlMatches);
 
-    // If the title is empty after extracting URLs, use the first attachment's title
-    // so pasting a bare URL results in a meaningful task name instead of a blank one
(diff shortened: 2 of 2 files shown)


johannesjo:
Thanks for the PR!

One ask: could you add a unit test for the bare-URL case in `short-syntax.spec.ts`? Something like:

```ts
it('should use attachment title as task name when pasting a bare URL in extract mode', async () => {
  const t = { ...TASK, title: 'https://example.com/my-cool-page' };
  const r = await shortSyntax(t, { ...CONFIG, urlBehavior: 'extract' });
  expect(r?.taskChanges.title).toBe('my-cool-page');
  expect(r?.attachments.length).toBe(1);
});
```

The rest looks good to me — the guard is solid and the change is minimal. 👍

jaydeep-pipaliya:
Thanks for the review @johannesjo\! Added a unit test for the bare-URL case in short-syntax.spec.ts — it verifies that pasting `https://example.com/my-cool-page` with `urlBehavior: 'extract'` results in the task title being `my-cool-page` (from the attachment title).

johannesjo (commented):
Thanks for adding the test! The fix is minimal and correctly scoped. A few notes before merging:

**Correctness — `newAttachments[0]` vs URL dedup edge case (minor)**

`newAttachments` is filtered by `filterDuplicateUrlAttachments`, so if the user re-pastes a URL that is *already* an attachment on the task, `newAttachments` will be empty. In `extract` mode that means the URL is stripped from the title but the fallback doesn't fire → task title becomes `""`.

If you want to cover that, the fix could use `allUrls[0]` (which is always populated when there are any URLs) instead of `newAttachments[0]`:

```ts
if (!cleanedTitle.trim() && allUrls.length > 0) {
  cleanedTitle = _baseNameForUrl(allUrls[0]) || cleanedTitle;
}
```

That said, "user pastes a duplicate URL with no surrounding text" is an edge case — the current fix is still a big improvement over the blank-title status quo, so this can be deferred.

**Redundant `.trim()` (nit)**

`removeUrlsFromTitle` already returns a trimmed string (line 648: `return cleanedTitle.trim().replace(...)`), so `!cleanedTitle.trim()` in the guard is equivalent to `!cleanedTitle`. Harmless, but you could simplify.

**Test coverage — root URL produces a domain basename**

For a bare root URL like `https://example.com` (no path segment), `_baseNameForUrl` returns `"example"` (TLD stripped). Worth adding a test to pin that behavior, so future readers understand the edge case is intentional:

```ts
it('should use domain basename as task name when pasting a root URL with urlBehavior "extract"', async () => {
  const t = { ...TASK, title: 'https://example.com' };
  const r = await shortSyntax(t, { ...CONFIG, urlBehavior: 'extract' });
  expect(r?.taskChanges.title).toBe('example');
});
```

**CI — Tests job still pending**

The `Tests` check hasn't finished yet. Please confirm it passes before merging.

---

Overall the approach is sound — using the URL basename is a pragmatic improvement over the blank-title behaviour reported in #6444, without the complexity of a network fetch for page metadata.

jaydeep-pipaliya:
Thanks for the thorough review @johannesjo! All three items addressed:

**1. `allUrls[0]` instead of `newAttachments[0]`** — Fixed. Now uses `allUrls[0]` so the fallback fires even when the pasted URL already exists as an attachment (where `newAttachments` would be empty after `filterDuplicateUrlAttachments`).

**2. Redundant `.trim()`** — Removed. Since `removeUrlsFromTitle` already returns a trimmed string, the guard is now just `!cleanedTitle`.

**3. Root URL test** — Added:
```ts
it('should use domain basename as task name when pasting a root URL with urlBehavior "extract"', ...)
// https://example.com → 'example'
```

All 157 short-syntax tests pass. The 5 `ImmediateUploadService` failures in the full suite are a pre-existing test isolation issue (they pass in isolation on both master and this branch).

johannesjo:
Thank you very much! <3

PR #7072, opened 2026-04-01
fix(electron): recreate tray indicator before minimize to tray

### LINKED ISSUE #7070: 🚨 lost the tray icon
### Version Used

v18.1.0

### Expected Behavior

When I minimize the windows, there should be a tray icon.

### Current Behavior

The tray icon missed.

### Steps to Reproduce

Just minimize/hide the super productivity window. win10 OS.

### Console Output

```shell

```

symonbaikov:
Closes #7070
On Windows, minimizing Super Productivity to tray can leave users without a visible tray icon. The app already creates the tray indicator on startup, but if the tray instance is missing or becomes invalid later, minimizing to tray hides the main window without ensuring that the indicator still exists.
## Solution
This change makes tray initialization reusable and restores the tray state before the window is hidden to tray.
- add `ensureIndicator()` to make tray creation idempotent
- reuse the saved tray config instead of reinitializing listeners multiple times
- restore tray menu, title/tooltip, and icon when the tray is recreated
- call `ensureIndicator()` before `hide()` in the minimize-to-tray close/minimize handlers
This makes the Windows tray behavior more robust and avoids ending up with a hidden window and no tray entry.
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)
## Checklist

### DIFF
--- a/electron/indicator.ts
+++ b/electron/indicator.ts
@@ -92,7 +92,7 @@ export const initIndicator = ({
   app: App;
   ICONS_FOLDER: string;
   forceDarkTray: boolean;
-}): Tray => {
+}): Tray | undefined => {
   indicatorConfig = {
     showApp,
     quitApp,
@@ -116,16 +116,9 @@ export const initIndicator = ({
   return ensureIndicator();
 };
 
-export const ensureIndicator = ({
-  refresh = false,
-}: { refresh?: boolean } = {}): Tray => {
+export const ensureIndicator = (): Tray | undefined => {
   if (!indicatorConfig) {
-    throw new Error('Indicator not initialized');
-  }
-
-  if (refresh && tray) {
-    tray.destroy();
-    tray = undefined;
+    return undefined;
   }
 
   if (tray) {
@@ -170,8 +163,7 @@ function initAppListeners(app: App): void {
   _isAppListenersInitialized = true;
   app.on('before-quit', () => {
     if (tray) {
-      tray.destroy();
-      tray = undefined;
+      destroyTray();
     }
   });
 }
@@ -451,6 +443,12 @@ const syncTray = (tr: Tray): void => {
   }
 };
 
+const destroyTray = (): void => {
+  tray?.destroy();
+  tray = undefined;
+  curIco = undefined;
+};
+
 // eslint-disable-next-line prefer-arrow/prefer-arrow-functions
 function createIndicatorMessage(
   task: TaskCopy,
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much!

# Code Review

Good direction overall — making tray creation idempotent via `ensureIndicator()` and extracting `syncTray()` for state restoration is the right approach. The duplicate-listener guards are also a nice defensive addition. Two bugs need fixing before this is safe to merge though:

---

### Bug 1 — `ensureIndicator()` throws, leaving window hidden with no tray

In both `appCloseHandler` and `appMinimizeHandler`, `event.preventDefault()` is called *before* `ensureIndicator()`. If `indicatorConfig` is null (init never ran, or failed), `ensureIndicator()` throws `'Indicator not initialized'` — the window gets hidden but no tray exists. That's the exact failure mode this PR is supposed to prevent.

Copilot flagged this too ([#discussion_r3019549595](https://github.com/super-productivity/super-productivity/pull/7072#discussion_r3019549595)) but it's unresolved.

**Fix:** return a no-op instead of throwing when `indicatorConfig` is absent:

```ts
export const ensureIndicator = (...): Tray | undefined => {
  if (!indicatorConfig) {
    return undefined; // tray never initialised — callers should handle this
  }
  // ...
};
```

And in `main-window.ts`, guard on the result:

```ts
const indicator = ensureIndicator();
if (!indicator && !getIsMinimizeToTray()) return; // fall back to normal behaviour
```

---

### Bug 2 — Stale `curIco` cache causes wrong icon on recreated tray

`setTrayIcon` skips `tr.setImage()` when `icoPath === curIco`. After a tray is destroyed and recreated:

1. Old tray was showing `running-d.png` → `curIco = 'running-d.png'`
2. `createTray()` constructs the new `Tray` with `stopped-d.png` as the constructor image (**does not update `curIco`**)
3. `syncTray()` calls `setTrayIcon(tr, 'running-d.png')` (task still active)
4. `icoPath === curIco` → **skips `tr.setImage()`** — new tray stuck showing the `stopped` icon

Copilot flagged this too ([#discussion_r3019549608](https://github.com/super-productivity/super-productivity/pull/7072#discussion_r3019549608)) — unresolved.

**Fix:** reset `curIco` whenever the tray is destroyed, e.g. in `ensureIndicator` and in the `before-quit` listener:

```ts
tray.destroy();
tray = undefined;
curIco = ''; // force next setTrayIcon call to actually apply the image
```

---

### Nit — `refresh` option is dead code

`ensureIndicator({ refresh: true })` is never called anywhere. Either add a call site or remove the option — it also contains the icon-cache bug above.

symonbaikov:
Implemented fix.

johannesjo:
Thank you very much!

PR #7080, opened 2026-04-01
Increase frequency of stale-discussions.yml workflow

zenoprax:
Despite "failing" in the GH Action logs there is still useful work being done by this action as-is. I have increased the frequency from weekly to daily to accelerate the catch-up process after which I expect it to remain within all rate limits.

I've also re-released my action using SemVer so dependabot should pick up any improvements I make to the rate limit problem itself.


johannesjo:
Thank you!

Issue #5672, opened 2025-12-06
🚨 Snap: Window doesn't appear on Wayland + AMD GPU (radeonsi) - GPU initialization fails


T-Alex-99:
### Version Used

16.5.2 (snap edge channel)

### Expected Behavior

SuperProductivity should open window when launching it 

### Current Behavior

 Environment

  - OS: Ubuntu 24.x (or your version)
  - Desktop: GNOME on Wayland
  - GPU: AMD (radeonsi driver)
  - SuperProductivity version: 16.5.2 (snap edge channel)
  - Installation method: Snap

  Problem

  After launching SuperProductivity, no window appears. The process runs in the background but the
  UI never shows.

  Error logs

  MESA-LOADER: failed to open radeonsi (search paths
  /snap/superproductivity/2876/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
  failed to load driver: radeonsi
  ANGLE Display::initialize error 12289: Failed to get system egl display
  Exiting GPU process due to errors during initialization

  Workaround

  superproductivity --disable-gpu

### Steps to Reproduce

  1. Install snap: sudo snap install superproductivity (or --edge)
  2. Have an AMD GPU with radeonsi driver on Wayland
  3. Run: superproductivity
  4. Expected: App window appears
  5. Actual: No window. Process runs in background but UI never renders.

### Can you reproduce this reliably?

Sometimes, it's intermittent.

### Console Output

```shell
MESA-LOADER: failed to open radeonsi (search paths
  /snap/superproductivity/2876/gnome-platform/usr/lib/x86_64-linux-gnu/dri)
  failed to load driver: radeonsi
  ANGLE Display::initialize error 12289: Failed to get system egl display
  Exiting GPU process due to errors during initialization
```

T-Alex-99:
Another workaround is:

superproductivity --ozone-platform=x11 --disable-gpu-compositing

johannesjo:
Thank you! I am downgrading electron once more. Seems like we need to wait a little longer for native wayland support with electron :(

andersk:
Probably related:

- electron-userland/electron-builder#9452

johannesjo:
Thank you @andersk !

johannesjo:
@T-Alex-99 sorry for the long silence on this one — could you check whether this still reproduces for you?

[...]

Issue #3866, opened 2025-01-10
Superproductivity won't start 


GaetanCia:
### Your Environment

- Version used: superproductivity-bin 11.1.1-1 | build from https://aur.archlinux.org/superproductivity-bin.git
- Operating System and version: Archlinux, Linux 6.12.8-arch1-1, 
- Desktop Environment: Wayland, Hyprland

### Expected Behavior

application start

### Current Behavior

Application appear in systray but it's not shown.

### Steps to Reproduce (for bugs)

git clone https://aur.archlinux.org/superproductivity-bin.git
cd superproductivity-bin
makepkg -s
sudo pacman -U superproductivity-bin-11.1.1-1-x86_64.pkg.tar.zst

### Can you reproduce this reliably?

Yes

### Error Log (Desktop only)

```
superproductivity                                                                                                                                                 

Start app...
13:16:41.841 › { enabled: null, showDevTools: false, mode: 'bottom' }
[161447:0110/131641.884063:ERROR:object_proxy.cc(576)] Failed to call method: org.freedesktop.DBus.Properties.Get: object_path= /org/freedesktop/portal/desktop: org.freedesktop.DBus.Error.InvalidArgs: L’interface « org.freedesktop.portal.FileChooser » doesn't exist
[161447:0110/131641.884146:ERROR:select_file_dialog_linux_portal.cc(294)] Failed to read portal version property
Saving backups to /home/bakelue/.config/superProductivity/backups
13:16:41.910 › Saving backups to /home/bakelue/.config/superProductivity/backups
13:16:42.407 › No custom styles detected at /home/bakelue/.config/superProductivity/styles.css
ERR Global Shortcut registration failed: Ctrl+Shift+X
13:16:42.417 › Error: Global Shortcut registration failed: Ctrl+Shift+X
    at errorHandlerWithFrontendInform (/usr/lib/superproductivity/app.asar/electron/error-handler-with-frontend-inform.js:10:18)
    at /usr/lib/superproductivity/app.asar/electron/ipc-handler.js:182:101
    at Array.forEach (<anonymous>)
    at registerShowAppShortCuts (/usr/lib/superproductivity/app.asar/electron/ipc-handler.js:141:18)
    at IpcMainImpl.<anonymous> (/usr/lib/superproductivity/app.asar/electron/ipc-handler.js:111:9)
    at IpcMainImpl.emit (node:events:518:28)
[...]

PR #7097, opened 2026-04-02
chore: update electron to v41.x

### LINKED ISSUE #6208: 💡 Electron 37 EOL

leandromqrs:
## Problem

Multiple problems running Linux with Wayland like app not starting, visual glitches and window bugs.

Resolve #6208 #5252 #5672 #5196 #3744 #3866   maybe more...

## Solution

Update Electron to latest stable version 41.x with much improvements in Wayland session that not necessary needs our intervention. See the article of Wayland updates in this version and the great job of Mitchell Cohen: https://www.electronjs.org/blog/tech-talk-wayland

Tested on GNOME 50, Fedora Siverblue 44

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Hey hey! Thank you for this. Did you read the technical plan in https://github.com/super-productivity/super-productivity/blob/master/docs/long-term-plans/electron-upgrade-to-v40.md ? unfortunately this is not a trivial updated it seems.

leandromqrs:
> Hey hey! Thank you for this. Did you read the technical plan in https://github.com/super-productivity/super-productivity/blob/master/docs/long-term-plans/electron-upgrade-to-v40.md ? unfortunately this is not a trivial updated it seems.

Yeah, I see the plan and almost is not necessary for this update, I can address each point and explain why this changes isn't necessary at the moment.

leandromqrs:
I pick the changes in the plan, ok:

1. Bump Electron version in package.json

Done!

2. Upgrade Snap gnome runtime via plug override in electron-builder.yaml

Newer versions of electron-builder natively support core22/core24 and automatically apply the correct modern GNOME Snapcraft extension. The Wayland crashes you experienced previously only happened because the outdated builder was forcing the ancient gnome-3-28 runtime onto newer Electron versions.

Now that the native toolchain automatically supplies the modern runtime (with up-to-date Mesa drivers and GSettings schemas), Electron 41's Wayland support works out-of-the-box. It's already set base: core22 and the builder handle the runtime and auto-connections cleanly, If you need a new version, we can set to core24, just for an stable and updated core.

3. Add Snap-only X11 override in start-app.ts (defense-in-depth)

It's not necessary this fallback script, since modern Snap already ensures runtime assembly, and Electron 41 itself automatically returns to the X11 Wayland event if it fails.

4. Update Flatpak runtime in electron-builder.yaml

Updated the runtime to latest new version, but removing the --socket=x11 is not necessary, Electron 41 can do this selecting the right display server (x11 or Wayland). The --socket=x11 flag is necessary for compatibility with DE's that not migrated to Wayland

5. Migrate deprecated protocol.registerFileProtocol in start-app.ts

We can do that later, it's not necessary at this crucial moment.

6. Add GPU cache cleanup on Electron version change in start-app.ts

Maybe this is a problem because the package type, Flatpak, Snap... I will take a look at this

7. Migrate url.format() (proactive deprecation cleanup) and  8. Run npm install and verify

Not necessary at this first moment, because Electron 41 is in node 22...

image

9. Update Flathub manifest (separate PR)

Necessary

leandromqrs:
The type error of node22 is fixed in Electron 40.1, see https://github.com/electron/electron/issues/49213#issuecomment-3820259101

leandromqrs:
I believe we can proceed now with this item :) @johannesjo 

Tested with executable, flatpak and snap packages, working well with wayland

Snap Ubuntu 25.10

Screenshot From 2026-04-05 09-28-26

Flatpak Fedora 44

image

leandromqrs:
Conflicts resolved

leandromqrs:
Please, cold you review my changes @johannesjo? I maked all requested changes from the plan, updated the flatpak runtime too

The runtime was updated yesterday in the flathub repo: https://github.com/flathub/com.super_productivity.SuperProductivity/pull/106

johannesjo:
Thank you very much for this!

I pushed two fixes directly to this branch:

**1. `fix(electron): move snap ozone-platform switch before app ready event`**
`app.commandLine.appendSwitch()` must be called before Chromium initialises — by the time the `ready` event fires the GPU/display backend is already running and the switch has no effect. Moved the Snap X11 fallback out of `appIN.on('ready', ...)` to run synchronously at startup alongside the existing `gtk-version` and `speech-dispatcher` switches. ([Electron docs confirm](https://www.electronjs.org/docs/latest/api/command-line-switches): "Use `app.commandLine.appendSwitch` in the main script before the ready event".)

**2. `fix(electron): align IS_GNOME_DESKTOP detection with preload.ts logic`**
The original `common.const.ts` check required `XDG_CURRENT_DESKTOP` to contain both `'gnome'` AND `'ubuntu'`, which only matches Ubuntu's GNOME session — not plain GNOME on Fedora, Arch, openSUSE, etc. The `preload.ts` `isGnomeDesktop()` already had the correct logic (4 env vars, OR). Aligned `common.const.ts` to match, preventing a split-brain where the main process forces native title bar but the renderer still hides the custom header → no title bar at all on non-Ubuntu GNOME.

---

**One item that still needs manual testing before merge:**

The removal of `protocol.registerFileProtocol('file', ...)` (Option A from the plan). Chromium's built-in `file://` handler should handle this correctly in Electron 41, but the old handler explicitly decoded URI-encoded paths (`decodeURI(request.url.replace('file:///', ''))`). Could you please verify:
- Open a task with a file attachment whose path contains a space (e.g. `/home/user/my file.png`) — does the image display?
- Same test with `#`, `?`, or `%` characters in the path
- Test on Windows with a backslash path

If any of those break, the fix is to replace the removed handler with `protocol.handle('file', ...)` using `pathToFileURL`.

leandromqrs:
> Thank you very much for this!
> 
> I pushed two fixes directly to this branch:
> 
> **1. `fix(electron): move snap ozone-platform switch before app ready event`** `app.commandLine.appendSwitch()` must be called before Chromium initialises — by the time the `ready` event fires the GPU/display backend is already running and the switch has no effect. Moved the Snap X11 fallback out of `appIN.on('ready', ...)` to run synchronously at startup alongside the existing `gtk-version` and `speech-dispatcher` switches. ([Electron docs confirm](https://www.electronjs.org/docs/latest/api/command-line-switches): "Use `app.commandLine.appendSwitch` in the main script before the ready event".)
> 
> **2. `fix(electron): align IS_GNOME_DESKTOP detection with preload.ts logic`** The original `common.const.ts` check required `XDG_CURRENT_DESKTOP` to contain both `'gnome'` AND `'ubuntu'`, which only matches Ubuntu's GNOME session — not plain GNOME on Fedora, Arch, openSUSE, etc. The `preload.ts` `isGnomeDesktop()` already had the correct logic (4 env vars, OR). Aligned `common.const.ts` to match, preventing a split-brain where the main process forces native title bar but the renderer still hides the custom header → no title bar at all on non-Ubuntu GNOME.
> 
> **One item that still needs manual testing before merge:**
> 
> The removal of `protocol.registerFileProtocol('file', ...)` (Option A from the plan). Chromium's built-in `file://` handler should handle this correctly in Electron 41, but the old handler explicitly decoded URI-encoded paths (`decodeURI(request.url.replace('file:///', ''))`). Could you please verify:
> 
>     * Open a task with a file attachment whose path contains a space (e.g. `/home/user/my file.png`) — does the image display?
> 
>     * Same test with `#`, `?`, or `%` characters in the path
> 
>     * Test on Windows with a backslash path
> 
> 
> If any of those break, the fix is to replace the removed handler with `protocol.handle('file', ...)` using `pathToFileURL`.

Ok, thanks! I will make these tests

leandromqrs:
- [x] Open a task with a file attachment whose path contains a space (e.g. /home/user/my file.png) — does the image display?

image

- [x] Same test with #, ?, or % characters in the path (Works only with % character)
image
image

I will try a VM to test in Windows

johannesjo:
Thank you very much  @leandromqrs ! I think we are probably fine for windows!

johannesjo:
Thank you very much! I'll merge this. Let's hope for the best! :)

leandromqrs:
Thank you @johannesjo! Happy to help with the project. Let me know if you need any follow-up fixes regarding this update! :)

PR #7099, opened 2026-04-02
Feat/task widget global shortcut

### LINKED ISSUE #6893: 💡 Would love for a global hotkey option to 'Enable/Disable' Overlay Indicator Window
### Problem Statement

Would love the option to add a configurable global hotkey to toggle the overlay indicator window. 
Currently the only way to enable/disable it is through Settings, which breaks my workflow when i am heads-down working. A Global hotkey option to toggle it on/off would make the overlay much more practical for daily use.

Hope it is possible without tilt the build. 

### Possible Solution

A Global hotkey option to toggle it on/off (If possible).

_No response_

### Additional context

_No response_

symonbaikov:
## Problem
Closes #6893.
The task widget/overlay can currently only be enabled or disabled through Settings. That makes it cumbersome to quickly hide or bring back the widget while working in other applications.
## Solution
This PR adds a new configurable global shortcut for toggling the task widget visibility on desktop.
Changes included:
- added a new keyboard config entry: `globalToggleTaskWidget`
- exposed the shortcut in the Electron-only "System Shortcuts" settings section
- added the translation key and label for the new shortcut
- registered the shortcut in Electron's global shortcut handler
- added a `toggleTaskWidgetVisibility()` helper in the task widget module
The shortcut only toggles visibility when the task widget feature is already enabled in settings, so it does not change the persisted enabled/disabled preference.
## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)
## Checklist


johannesjo (commented):
## Review: Feat/task widget global shortcut

### Summary

This PR adds a configurable global shortcut (`globalToggleTaskWidget`) to toggle the task widget visibility from the desktop, along with a significant refactoring of `electron/indicator.ts` to support lazy tray creation via `ensureIndicator()`. The core feature is well-conceived and addresses a real usability gap (issue #6893). The implementation touches 9 files and is generally clean, but there are a few issues worth discussing.

---

### Positives

1. **`toggleTaskWidgetVisibility()` is well-guarded** -- it correctly checks `isTaskWidgetEnabled`, `taskWidgetWin` existence, `isDestroyed()`, and `isVisible()` before acting. This is defensive and appropriate for Electron window management.

2. **Idempotent listener registration** -- The addition of `_isAppListenersInitialized` and `_isListenersInitialized` guards in `indicator.ts` prevents duplicate IPC listener registration. This is a genuine improvement over the original code which used `if (tray)` as a proxy for "already initialized."

3. **Frontend changes are minimal and correct** -- The keyboard config model, default config, form config, translation key, and `t.const.ts` changes are all consistent and follow existing patterns. The shortcut is correctly scoped to the Electron-only `IS_ELECTRON` conditional in `keyboard-form.const.ts`.

4. **`destroyTray()` helper** -- Centralizing tray cleanup and resetting `curIco` is a nice improvement.

---

### Issues and Concerns

#### 1. Behavioral regression in `appCloseHandler` (main-window.ts) -- HIGH

The close handler logic has changed significantly. In the original code:

```js
if (!getIsQuiting()) {
  event.preventDefault();        // ALWAYS prevents close
  if (getIsMinimizeToTray()) {
    // hide to tray
    return;
  }
  // otherwise handle before-close actions or quit
}
```

In the PR:

```js
if (!getIsQuiting()) {
  if (getIsMinimizeToTray()) {
    const indicator = ensureIndicator();
    if (indicator) {
      event.preventDefault();    // Only prevents if tray succeeds
      // hide to tray
      return;
    }
  }
  event.preventDefault();        // Falls through here
  // handle before-close actions or quit
}
```

The problem: if `getIsMinimizeToTray()` is `true` but `ensureIndicator()` returns `undefined` (e.g., `initIndicator()` was never called so `indicatorConfig` is `undefined`), the code **falls through** to the before-close/quit path instead of hiding to tray. This is a silent behavioral change -- the user has "minimize to tray" enabled but the app will try to quit instead of hiding. This could happen if there's a startup ordering issue or if indicator initialization fails.

The intent seems to be "don't hide to tray if we can't create a tray icon," which is reasonable, but this should be documented with a comment explaining the fallback behavior, and ideally should log a warning so it's debuggable.

#### 2. Behavioral regression in `appMinimizeHandler` (main-window.ts) -- MEDIUM

```js
mainWin.on('minimize', (event: Event) => {
  if (getIsMinimizeToTray()) {
    const indicator = ensureIndicator();
    if (!indicator) {
      return;   // <-- swallows the minimize event silently
    }
    event.preventDefault();
    ...
  }
```

If `ensureIndicator()` fails, the handler **returns early** without calling `event.preventDefault()`. The original `else` branch (regular minimize, show task widget, show dock on Mac) is **never reached** because the early return happens inside the `if (getIsMinimizeToTray())` block. So when minimize-to-tray is enabled but the tray can't be created, the minimize proceeds normally but the task widget won't be shown and dock won't appear on Mac. This seems unintentional -- the function should probably still fall through to the `else` behavior.

#### 3. Dead code: `_showApp` and `_quitApp` module-level variables are now unused -- LOW

After the refactor, the tray click handler uses `indicatorConfig?.showApp()` and the context menu uses `_showApp` / `_quitApp`. The module-level `_showApp` and `_quitApp` variables (lines 17-18 in the original) are still assigned in `initIndicator` (line 92-93) and still used in `createContextMenu` (lines 436-437). This is not a bug, but it means there are now **two** places storing the same callbacks: the module variables and `indicatorConfig`. This duplication could lead to subtle inconsistencies. Consider using only `indicatorConfig.showApp` / `indicatorConfig.quitApp` everywhere and removing the standalone `_showApp` / `_quitApp` variables.

#### 4. `syncTray` duplicates logic from `initListeners` IPC handlers -- MEDIUM

The new `syncTray()` function (indicator.ts, ~60 lines) reconstructs the tray state from cached `_last*` variables. This duplicates the logic spread across the `IPC.CURRENT_TASK_UPDATED` and `IPC.SET_PROGRESS_BAR` handlers. While this is necessary for lazy tray creation, the duplicated logic will drift over time. Consider extracting shared helper functions for the tray title/tooltip/icon update logic to reduce this duplication.

#### 5. `ensureIndicator` doesn't re-initialize `DIR` or `shouldUseDarkColors` -- LOW

`ensureIndicator()` creates a tray via `createTray()`, which uses module-level `DIR` and `shouldUseDarkColors`. These are set in `initIndicator()`. If `ensureIndicator()` is called before `initIndicator()` (theoretically possible since it's exported), `DIR` would be `undefined` and `shouldUseDarkColors` would be `undefined`, leading to broken icon paths. The `if (!indicatorConfig)` guard should prevent this in practice, but the coupling is implicit.

#### 6. No `globalToggleTaskStart` pattern parity -- NITPICK

The existing `globalToggleTaskStart` shortcut sends an IPC message to the renderer: `mainWin.webContents.send(IPC.TASK_TOGGLE_START)`. The new `globalToggleTaskWidget` shortcut directly calls `toggleTaskWidgetVisibility()` from the main process, which is correct since the task widget is an Electron-only main-process concern. Just noting this is intentionally different from the other shortcuts -- good design choice.

---

### Minor Suggestions

- The `toggleTaskWidgetVisibility` export from `task-widget.ts` is clean and well-placed. Consider adding a brief JSDoc comment noting that it's intended for global shortcut use and only works when the widget feature is enabled in settings.
- In `global-shortcuts.ts`, the `actionFn` for `globalToggleTaskWidget` could be simplified from `() => { toggleTaskWidgetVisibility(); }` to just `toggleTaskWidgetVisibility` directly (same as how `showOrFocus` is used elsewhere), though this is purely stylistic.

---

### Verdict

The core feature (global shortcut + toggle function) is solid and well-implemented. The main concern is the behavioral changes in `main-window.ts` close/minimize handlers, which alter the existing minimize-to-tray behavior when tray creation fails. These should be reviewed carefully to ensure they match the intended UX, especially the minimize handler's early return that skips the task widget show + Mac dock behavior.

---

johannesjo:
Can you have a look at the conflicts @symonbaikov ?

symonbaikov:
Yes. Working on resolving them.

johannesjo:
Thanks for this, @symonbaikov — the feature itself is clean and genuinely useful, and `toggleTaskWidgetVisibility()` is correct (it no-ops when the widget is disabled, never mutates the persisted enabled/disabled preference, and delegates show/recreate to `showTaskWidget()`). 👍

The blocker is **staleness**, not the feature. This branch is based on **18.1.2**, and since then `master` has independently landed its own (more complete) version of the tray-indicator refactor that this PR also contains. Of the +236/−105 diff, only ~36 lines are genuinely new — the rest is `indicator.ts`/`main-window.ts` churn that now **conflicts with and regresses** current master.

### Why the electron hunks can't be merged as-is

Merging this PR's `electron/indicator.ts` / `electron/main-window.ts` over today's master would:

1. **Fail to compile.** The PR's `indicator.ts` reads `settings.taskWidget.isEnabled/opacity/isAlwaysShow` from the `UPDATE_SETTINGS` payload, but on master `taskWidget` is intentionally **not** part of `GlobalConfigState` (see `global-config.model.ts:320` — it moved to a per-instance `UPDATE_TASK_WIDGET_SETTINGS` channel handled by `initTaskWidgetSettingsListener()`). This is a TypeScript error and would also break widget-settings propagation.
2. **Re-introduce a quit bug.** The PR's close handler drops master's `\!getIsQuitRequested()` guard and the `closeWinAndQuit()` path, so an explicit **Quit** while *minimize-to-tray* is enabled could get swallowed into "hide to tray" again.
3. **Lose platform fixes** master already has: `nativeImage`/macOS template-image tray rendering, the GNOME/Linux icon fix, and `getDistChannel()`-based Windows tray GUID selection.

### Suggested path

Rebase onto current `master` and keep **only** the shortcut feature (the indicator/main-window refactor is already there, in better shape). After rebase the PR is ~36 lines across 7 files:

- `electron/task-widget/task-widget.ts` — `toggleTaskWidgetVisibility()` ✅ applies cleanly
- `electron/ipc-handlers/global-shortcuts.ts` — import + allowlist entry + `case` ✅ applies cleanly
- `keyboard-config.model.ts`, `default-global-config.const.ts` ✅
- `t.const.ts`, `en.json` — ⚠️ small ordering fix: put `GLOBAL_TOGGLE_TASK_WIDGET` **after** `GLOBAL_TOGGLE_TASK_START` (alphabetical, matches surrounding keys)
- `keyboard-form.const.ts` — ⚠️ master refactored this to a `kbField()` helper; please use `kbField('globalToggleTaskWidget', T.GCF.KEYBOARD.GLOBAL_TOGGLE_TASK_WIDGET)` instead of the verbose `{ key, type, templateOptions }` object

### Two small follow-ups

- **Test:** there's an existing `electron/*.test.cjs` (`node:test`) harness — a quick spec covering the three toggle branches (disabled → no-op, visible → hide, hidden → show) would be great per the contributing guidelines.
- **UX note (optional):** while the main window is focused/visible, its `focus`/`show` handlers call `hideTaskWidget()` (unless "always show" is set), so the shortcut's *show* is mostly effective when the main window is hidden. That's consistent with the widget's "companion to a hidden window" design — just flagging in case you want a sticky "user-forced visible" behavior instead.

Net: great feature, just needs a rebase to drop the now-superseded electron refactor. Happy to help if the rebase gets hairy around `indicator.ts`. 🙏

symonbaikov:
Thanks for the detailed breakdown @johannesjo — that was exactly the right call. I've rebased onto current `master` and dropped the superseded tray-indicator refactor entirely. The PR is now a single commit that only adds the shortcut feature; `indicator.ts` and `main-window.ts` are left on master's version (the one `main-window.ts` change is the additive user-forced-visible gate described below).

Concretely, following your suggested path:

- `task-widget/task-widget.ts` — `toggleTaskWidgetVisibility()`
- `ipc-handlers/global-shortcuts.ts` — import + allowlist entry + `case`
- `keyboard-config.model.ts`, `default-global-config.const.ts` — apply cleanly
- `t.const.ts`, `en.json` — `GLOBAL_TOGGLE_TASK_WIDGET` now ordered **after** `GLOBAL_TOGGLE_TASK_START`
- `keyboard-form.const.ts` — uses the `kbField('globalToggleTaskWidget', T.GCF.KEYBOARD.GLOBAL_TOGGLE_TASK_WIDGET)` helper

**Follow-ups, both done:**

- **Test** — added `electron/task-widget.test.cjs` using the existing `node:test` harness, covering the three toggle branches (disabled → no-op, hidden → show, visible → hide) plus the sticky-flag behavior below. `npm run test:electron` is green (10/10).
- **Sticky "user-forced visible"** — implemented your optional UX note. When the user reveals the widget via the shortcut while the main window is visible, an `isUserForcedVisible` flag now suppresses the auto-hide on `restore`/`show`/`focus` (and in `showOrFocus`), just like `isAlwaysShow`. It clears when the user toggles the widget off, opens the app from the widget, or the feature is disabled — so the shortcut is now effective regardless of main-window state, without changing any persisted preference.

Conflicts are resolved and the branch merges cleanly. Let me know if you'd prefer the sticky behavior split into its own commit.

johannesjo:
Thanks for the clean rebase, @symonbaikov — dropping the superseded tray-indicator churn and keeping this to just the shortcut + sticky-visible flag is exactly right, and the `node:test` coverage is appreciated. 🙏 One real issue with the sticky flag before this is good to merge:

**`isUserForcedVisible` can leak across a disable → re-enable cycle.**

The flag is only reset on disable via `destroyTaskWidget()`, but `updateTaskWidgetEnabled(false)` only reaches it when the window exists:

```ts
// task-widget.ts:55
} else if (!isEnabled && taskWidgetWin) {
  destroyTaskWidget();   // <- only path that clears isUserForcedVisible on disable
}
```

Meanwhile `toggleTaskWidgetVisibility()` sets the flag and calls `showTaskWidget()`, which recreates the window **asynchronously** when it's absent (`task-widget.ts:252`, the "accidentally closed" path), and the `closed` handler (`:200`) nulls `taskWidgetWin` without clearing the flag. So:

1. Widget window is absent (closed by the OS, or mid async re-create) while the feature is still enabled.
2. User hits the shortcut → `isUserForcedVisible = true`.
3. Feature is disabled in settings while `taskWidgetWin` is still `null` → the line-55 guard is false → `destroyTaskWidget()` is skipped → flag stays `true`.
4. Re-enable → the stale flag makes `main-window.ts:458/:464/:470` and `various-shared.ts:45` skip `hideTaskWidget()`, so the widget no longer auto-hides on main-window focus — until the user toggles the shortcut again or opens the app from the widget.

**Suggested fix** — clear it unconditionally on any disable:

```ts
export const updateTaskWidgetEnabled = (isEnabled: boolean): void => {
  isTaskWidgetEnabled = isEnabled;
  if (!isEnabled) {
    isUserForcedVisible = false; // reset regardless of whether the window exists
  }
  // ...
};
```

and tie "forced visible" to the window's lifetime (`task-widget.ts:200`):

```ts
taskWidgetWin.on('closed', () => {
  taskWidgetWin = null;
  isUserForcedVisible = false;
});
```

The existing reset in `destroyTaskWidget()` then just becomes harmless redundancy.

**Optional:** a regression test for the absent-window case — your `disabling the widget clears the sticky user-forced flag` test currently only covers the window-exists path. Setting the flag, nulling the fake window, then `updateTaskWidgetEnabled(false)` and asserting the flag cleared would lock this in.

Otherwise this looks great. (The red CI is an unrelated flaky e2e — `archive-subtasks.spec.ts:247` timeout — not this PR.)

symonbaikov:
Good catch @johannesjo — that's a real leak. Fixed exactly as you suggested:

- `updateTaskWidgetEnabled()` now clears `isUserForcedVisible` unconditionally on any disable, before the window-existence check — so the mid-async-recreate / OS-closed gap (where `taskWidgetWin` is `null` and `destroyTaskWidget()` is skipped) no longer leaks a stale flag into the next enable.
- The `taskWidgetWin.on('closed', …)` handler now also clears the flag, tying "forced visible" to the window's lifetime. The existing reset in `destroyTaskWidget()` is kept as harmless redundancy.

Also added the optional regression coverage:

- `disabling clears the sticky flag even when the widget window is absent` — enables without flushing (so the window is still mid-creation), sets the flag via the shortcut, asserts no window exists, then disables and asserts the flag cleared. This fails against the old guard and passes now.
- `the closed event clears the sticky flag so it does not outlive the window` — fires the window's `closed` event and asserts the flag is cleared.

`npm run test:electron` is green (12/12). Thanks again — and noted on the flaky `archive-subtasks` e2e.

johannesjo:
I re-checked the current head (`815c5ae`) after the latest sticky-flag fixes. The config/i18n wiring looks good and `npm run test:electron` is green locally (12/12), but I still see two async lifecycle bugs in the widget creation path, plus one low-severity UX concern.

1. **Medium: disabling during async creation can still leave a widget window alive while disabled.**

   In `electron/task-widget/task-widget.ts`, `updateTaskWidgetEnabled(true)` starts `createTaskWidgetWindow()`, which awaits `loadSimpleStoreAll()` around line 130. If settings disable the widget before that await resolves, `updateTaskWidgetEnabled(false)` clears `isUserForcedVisible`, but because `taskWidgetWin` is still `null`, it skips `destroyTaskWidget()` and does not invalidate the in-flight create. When `createTaskWidgetWindow()` resumes, it sets `isCreatingWindow = false` and unconditionally constructs a `BrowserWindow` around line 170, even though `isTaskWidgetEnabled` is now false. That leaves a hidden widget window and the widget IPC listener alive while the feature is disabled.

   Suggested fix: track/reuse an in-flight creation promise or a generation token, invalidate it on disable, and re-check `isTaskWidgetEnabled` after `loadSimpleStoreAll()` before creating the `BrowserWindow`. A regression test should enable, disable before the store promise resolves, then assert no widget window is created or that it is destroyed.

2. **Medium: shortcut reveal can be lost while creation is already in flight.**

   `toggleTaskWidgetVisibility()` sets `isUserForcedVisible = true` and calls `showTaskWidget()`. If the widget window is still being created from the initial enable path, `taskWidgetWin` is `null` and `isCreatingWindow` is true. `showTaskWidget()` calls `createTaskWidgetWindow().then(...)`, but `createTaskWidgetWindow()` immediately returns due to the `isCreatingWindow` guard, so the `.then()` callback runs before a window exists and never shows anything. The result is a hidden widget with `isUserForcedVisible === true`, requiring a second shortcut press or another path to recover.

   Suggested fix: make `showTaskWidget()` await/reuse the active creation promise, or record a pending show-after-create request and consume it after the original creation finishes, while still checking that the widget is enabled. Add a controllable async test for pressing the shortcut while `isCreatingWindow === true`.

3. **Low: the global shortcut reveal can steal focus from the current app.**

   The shortcut path ultimately uses `BrowserWindow.show()` (`taskWidgetWin.show()`), and Electron API docs in `node_modules/electron/electron.d.ts` state that `show()` gives focus to the window. Since this shortcut is meant for use while working in other applications, revealing the widget may unexpectedly activate the widget. Consider a shortcut-specific inactive reveal using `showInactive()` or a `showTaskWidget({ inactive: true })` mode, while keeping the existing `show()` behavior for minimize/tray flows if that focus behavior is desired there.

The older tray/indicator concerns in the PR thread are outdated after the rebase; I did not see current config/i18n issues.

johannesjo:
Thank you very much! <3

PR #7103, opened 2026-04-02
Feat/check for updates 7100

### LINKED ISSUE #7100: 💡 Check for update button
### Problem Statement

Missing manual update trigger in settings

### Possible Solution

Can we get a 'Check for Updates' button in Settings to allow for direct updates

_No response_

### Additional context

_No response_

### documentation gap

symonbaikov:
Problem
Closes #7100.
Desktop users currently have no manual way to check whether a newer app version is available from within the app settings. The settings page shows the current version and changelog link, but it does not provide a direct update check action.
Solution
- Added a new Electron IPC endpoint for manual update checks against the latest GitHub release
- Implemented shared update-check response types and exposed the new API through the preload bridge
- Compared the latest release tag with the packaged app version in the Electron main process
- Added a "Check for Updates" action to the settings footer for Electron builds
- Added translated success and error messages for:
  - update available
  - already on latest version
  - failed update check
- Added a release-page shortcut so users can open the matching GitHub release when a newer version is found
Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)
Checklist
- [ ] I have included relevant changes to the documentation/wiki (https://github.com/super-productivity/super-productivity/tree/master/docs/wiki).
- [x] I have run npm run checkFile on changed .ts/.scss files
- [x] My commit messages follow the Angular format (type(scope): description)


johannesjo (commented):
## Review of "Check for Updates" feature (PR #7103)

Overall this is a well-structured feature addition. The IPC plumbing follows existing patterns, the Electron-only guard is correctly applied, and the shared model types are clean. Below are observations organized by severity.

---

### Issues to address

**1. Snack "spinner" snack is never dismissed on success/error path**

The initial `_snackService.open({ ..., isSpinner: true, promise })` call passes the `promise`, which gives the snackbar a `polling-snack` panel class and shows a spinner. However, looking at `SnackService._openSnack`, the service does **not** auto-dismiss when the promise resolves -- it only uses the promise to style the snackbar. The subsequent `_snackService.open(...)` calls (for success/error) will replace the spinner snack thanks to `@debounce(100)`, but only if the promise resolves quickly. If there is a network delay and the user navigates away from settings (triggering `setActiveWorkContext` which calls `close()`), then navigates back and clicks again, there could be stale state. This is a minor timing concern but worth being aware of.

More importantly: if the GitHub API call takes a long time (e.g. slow proxy, rate-limited with retry-after), the user has no way to cancel. Consider adding a timeout (e.g. 15-30 seconds) to the fetch call in `_getLatestReleaseData` via an `AbortController` / `signal` option, so the spinner doesn't hang indefinitely.

**2. No rate-limit / debounce protection on the UI side**

The "Check for Updates" link can be clicked repeatedly with no guard. Each click fires a new GitHub API request. The GitHub unauthenticated rate limit is 60 requests/hour per IP. Consider one of:
- Disabling the link while a check is in flight (e.g. a simple boolean flag `isCheckingForUpdate`)
- Debouncing at the IPC handler level
- Caching the result for a short period (e.g. 60 seconds)

**3. Redundant logging in error handler**

In `auto-update.ts` line ~101-102:
```ts
error('Failed to check for updates', err);
log('Failed to check for updates', err);
```
`error()` from `electron-log` already writes to the log file (at error level). Calling `log()` (which logs at info level) duplicates the message. Other IPC handlers in this codebase do not double-log. Remove the `log()` call and keep only `error()`.

**4. Type guard uses inline type instead of the exported `UpdateCheckErrorResult`**

As Copilot also noted, the type guard:
```ts
private _isUpdateCheckError(result: UpdateCheckResponse): result is { error: string }
```
should reference the exported `UpdateCheckErrorResult` type:
```ts
private _isUpdateCheckError(result: UpdateCheckResponse): result is UpdateCheckErrorResult
```
This keeps the shared type contract as the single source of truth.

**5. Error detail from the backend is discarded**

The IPC handler carefully extracts GitHub's error message (e.g. "API rate limit exceeded"), but the frontend always shows the generic `T.PS.UPDATE_CHECK_ERROR` string. Either:
- Surface the specific error in the snackbar (e.g. as `translateParams`) so users know *why* it failed, or
- Simplify the backend to not bother extracting the message if it will never be shown.

I'd recommend the former -- showing the backend error helps users self-diagnose (rate limiting, network issues, etc.).

---

### Minor / style suggestions

**6. Pre-release version handling**

`_toVersionParts` strips pre-release suffixes via `.split('-')[0]`, which means `v10.0.0-beta.1` and `v10.0.0` are treated as equal. This is reasonable for a "latest stable" check, but worth documenting with a comment so future maintainers understand the intent.

**7. `['User-Agent']` computed property syntax**

In `auto-update.ts`:
```ts
['User-Agent']: `super-productivity/${app.getVersion()}`,
```
This works but the computed property bracket syntax is unnecessary for a string literal key. A simpler form:
```ts
'User-Agent': `super-productivity/${app.getVersion()}`,
```

**8. `$event?.preventDefault()` in template**

```html
(click)="checkForUpdate(); $event?.preventDefault()"
```
The `$event` in Angular templates is always defined inside a `(click)` handler, so the optional chaining `?.` is unnecessary. `$event.preventDefault()` is sufficient and matches patterns elsewhere in this file (e.g. `downloadLogs()`).

---

### Positives

- Clean separation: shared model in `electron/shared-with-frontend/`, IPC handler in its own file, proper re-export via index.
- Correctly uses `IS_ELECTRON` guard both in the template (`@if`) and in the method body, preventing any accidental access to `window.ea` in web/mobile builds.
- Proper use of `createProxyAwareAgent()` so the feature works behind corporate proxies.
- Version comparison logic is straightforward and handles variable-length version parts.
- Good error handling in the IPC handler -- catches both HTTP errors and network failures, returns a structured error rather than throwing across the IPC boundary.
- Translation keys follow existing conventions and only `en.json` is edited (per project guidelines).
- Privacy-respecting: the only outbound request is to the public GitHub releases API with a standard User-Agent header. No analytics, no tracking.

---

### Summary

The feature is well-implemented and follows project patterns. The main actionable items are: (1) add a fetch timeout to prevent indefinite spinner hangs, (2) add a UI-side guard against rapid repeated clicks, (3) surface the backend error message to the user, and (4) the minor cleanup items above. None are blockers, but items 1-3 would meaningfully improve the user experience.

---

Zoriot:
This seems to be already implemented. There is a Check for updates button in the settings + the update pop up

Issue #6814, opened 2026-03-13
🚨 Can't switch profile after restart


BurningDev:
### Version Used

v17.4.1 win installer

### Expected Behavior

**Expected behaviour**: If I click on another existing profile, it should switch all data to the new profile.

### Current Behavior

**Current behaviour**: When I click on the other profile, nothing happens (_see video below_). The current profile remains active.

I have noticed something, though. When I restart Super Productivity, I can change the profile once, but not after that. If I then want to change it again, I have to close the programme and open it again.

<img src="https://github.com/user-attachments/assets/71a28f23-de58-4e65-8248-083b0c124ffb" width="500">

### Steps to Reproduce

1. Add a new profile
2. Restart the application
3. try to switch the profile several times

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

novikov1337danil:
@overcuriousity could you take a look?

overcuriousity:
Needs more information.
Which platform and form of deployment was used?

BurningDev:
I installed Super Productivity via https://github.com/super-productivity/super-productivity/releases/tag/v17.4.1 -> "Super-Productivity-Setup-x64.exe".

My OS is Windows 11.

BurningDev:
@overcuriousity Could you perhaps tell me which file contains the data and profile information relating to the AppData/Roaming/superProductivity folder?

Issue #6541, opened 2026-02-15
🚨 user profile


JanesSystems:
### Version Used

latest

### Expected Behavior

<img width="946" height="814" alt="Image" src="https://github.com/user-attachments/assets/d405b96d-2ee7-402d-bc7d-e0cb1e4291d5" />
i dont know what to do 
i only use local sync
and also i am not able to do switch, i think need more work on this feature

johannesjo:
@overcuriousity

overcuriousity:
is there additional info available? 
From the screenshot in the detailed bug report, this looks more like it is related to the sync feature maybe?

hiJanes:
> this looks more like it is related to the sync feature maybe?

yes it is

>is there additional info available?

and no, i have disable sync now

hiJanes:
?? @johannesjo shouldn't be u on a break?

PR #7106, opened 2026-04-02
fix(user-profile): reload app after profile switch to fix stale state

### LINKED ISSUE #5900: 🚨 User Profile applies only after restart
### Version Used

16.8.3

### Expected Behavior

I am using the (experimental) user profiles which adds a profile selection button in the header bar.

I would expect the profile to switch immediatly when selecting another profile.

### Current Behavior

The profile does not switch immediately, but I must restart to apply it.

### Steps to Reproduce

1. Activate "Experimental User Profiles" in the settings
2. Click the new profile header button, then "Manage"
3. Add a second profile
4. Switch between the profiles and observe the content does not change
5. Restart the app and observe the change of content

### Console Output

```shell

```

overcuriousity:
Profile switching only worked on the first attempt after a restart because WorkContextService and other services initialize once via allDataWasLoaded at startup. Subsequent loadAllData dispatches don't re-trigger that initialization, leaving cached state from the previous profile in place.

Add window.location.reload() after a successful profile switch so the app fully reinitializes with the new profile's data, which was already persisted to IndexedDB by importCompleteBackup.

Fixes #5900, #6814, #6541

### DIFF
--- a/src/app/features/user-profile/user-profile.service.ts
+++ b/src/app/features/user-profile/user-profile.service.ts
@@ -351,13 +351,12 @@ export class UserProfileService {
       // Step 7: Handle target profile data
       if (targetData) {
         // Profile has existing data - import it.
-        // importCompleteBackup clears ops + state_cache and dispatches loadAllData,
-        // fully replacing all NgRx feature state in-memory (no page reload needed).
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much!

PR #7115, opened 2026-04-02
feat(tasks): add subtask progress bar

### LINKED ISSUE #7057: 💡 add progress bar for subtasks
### Problem Statement

if i have 15+subtasks and i want to know a quick status of my work i have to count them manually completed/uncompleted subtasks to get a idea 

### Possible Solution

adding a progress bar  on main task can solve this 

_No response_

### Additional context

_No response_

symonbaikov:
## Problem
Closes #7057.
Parent tasks with many subtasks do not provide an at-a-glance indication of completion progress. Users currently need to manually count completed vs incomplete subtasks to understand how far a task has progressed.
## Solution
This change adds a subtask completion progress bar to parent task rows in the main task list.
- computes progress from completed subtasks vs total subtasks
- renders a thin progress bar for tasks that have subtasks
- keeps the bar visible even when subtasks are hidden/collapsed, so progress remains visible at all times
- uses the existing `progress-bar` component and existing success styling to keep the implementation small and consistent
- adds a small style adjustment so the subtask progress bar can coexist with the current-task time progress bar without overlapping visually
## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)
## Checklist

### DIFF
--- a/src/app/features/tasks/task/task.component.html
+++ b/src/app/features/tasks/task/task.component.html
@@ -278,6 +278,11 @@
       ></progress-bar>
     }
     @if (t.subTasks?.length) {
+      <progress-bar
+        [cssClass]="subTaskProgressCssClass()"
+        [isAlwaysVisible]="true"
+        [progress]="subTaskProgress()"
+      ></progress-bar>
       <div class="sub-tasks">
         <button
           (click)="toggleSubTaskMode()"
--- a/src/app/features/tasks/task/task.component.ts
+++ b/src/app/features/tasks/task/task.component.ts
@@ -104,6 +104,7 @@ import { millisecondsDiffToRemindOption } from '../util/remind-option-to-millise
 import { MenuTreeService } from '../../menu-tree/menu-tree.service';
 import { SelectOptionRowComponent } from '../../../ui/select-option-row/select-option-row.component';
 import { SnackService } from '../../../core/snack/snack.service';
+import { calcSubTaskProgress } from '../util/calc-sub-task-progress';
 
 @Component({
   selector: 'task',
@@ -272,6 +273,16 @@ export class TaskComponent implements OnDestroy, AfterViewInit {
   // Checklist progress derived from markdown checklist in task notes (null = no checklist)
   checklistProgress = computed<ChecklistProgress | null>(() =>
     getChecklistProgress(this.task().notes),
+  );
+
+  subTaskProgress = computed<number>(() => {
+    return calcSubTaskProgress(this.task().subTasks);
+  });
+
+  subTaskProgressCssClass = computed<string>(() =>
+    this.isCurrent() && this.progress() > 1
+      ? 'bg-success sub-task-progress isStackedBelowProgress'
+      : 'bg-success sub-task-progress',
   );
 
   isShowRemoveFromToday = computed(() => {
--- a/src/app/features/tasks/util/calc-sub-task-progress.spec.ts
+++ b/src/app/features/tasks/util/calc-sub-task-progress.spec.ts
@@ -1,0 +1,30 @@
+import { Task } from '../task.model';
+
+import { calcSubTaskProgress } from './calc-sub-task-progress';
+
+describe('calcSubTaskProgress', () => {
+  const createSubTask = (isDone: boolean): Task => ({ isDone }) as Task;
+
+  it('should return zero for no subtasks', () => {
+    expect(calcSubTaskProgress([])).toBe(0);
+  });
+
+  it('should return zero when no subtasks are done', () => {
+    expect(calcSubTaskProgress([createSubTask(false), createSubTask(false)])).toBe(0);
+  });
+
+  it('should return partial completion percentage', () => {
+    expect(
+      calcSubTaskProgress([
+        createSubTask(true),
+        createSubTask(false),
+        createSubTask(false),
+        createSubTask(true),
+      ]),
+    ).toBe(50);
+  });
+
+  it('should return one hundred when all subtasks are done', () => {
+    expect(calcSubTaskProgress([createSubTask(true), createSubTask(true)])).toBe(100);
+  });
+});
--- a/src/app/features/tasks/util/calc-sub-task-progress.ts
+++ b/src/app/features/tasks/util/calc-sub-task-progress.ts
@@ -1,0 +1,10 @@
+import { Task } from '../task.model';
+
+export const calcSubTaskProgress = (subTasks: readonly Task[]): number => {
+  if (subTasks.length === 0) {
+    return 0;
+  }
+
+  const doneSubTasksCount = subTasks.filter((subTask) => subTask.isDone).length;
+  return (doneSubTasksCount / subTasks.length) * 100;
+};
--- a/src/app/ui/progress-bar/progress-bar.component.scss
+++ b/src/app/ui/progress-bar/progress-bar.component.scss
@@ -21,6 +21,32 @@
     border-radius: var(--card-border-radius);
   }
 
+  &.isAlwaysVisible {
+    &:after {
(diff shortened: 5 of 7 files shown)


johannesjo (commented):
## Review: feat(tasks): add subtask progress bar

Nice feature idea -- an at-a-glance subtask completion indicator is a solid UX improvement. The implementation is small and focused, which is great. I do see a few issues that should be addressed before merging.

---

### Bug: CSS rule will not work due to view encapsulation

**File:** `task.component.scss`

```scss
:host.hasVisibleTimeProgress progress-bar.sub-task-progress::after {
  top: 0;
}
```

This rule targets the `::after` pseudo-element of the `<progress-bar>` element. However, the `::after` is defined inside `progress-bar.component.scss`, and both components use Angular's default `ViewEncapsulation.Emulated`. Angular scopes styles by adding `_ngcontent-xxx` / `_nghost-xxx` attribute selectors, which means the task component's styles **cannot penetrate** into the progress-bar component's `::after` pseudo-element. This CSS rule will silently have no effect.

**Options to fix:**
1. Move the offset logic into the `ProgressBarComponent` itself (e.g., add an optional `@Input()` for vertical offset or an additional CSS class that the progress-bar component handles internally).
2. Use `::ng-deep` (deprecated but still functional) as a short-term workaround: `:host.hasVisibleTimeProgress ::ng-deep progress-bar.sub-task-progress::after { top: 0; }`.
3. Use a global style in `_task-controls.scss` (which already styles `progress-bar` at the task level) to handle this case.

Option 1 is the cleanest long-term approach. Please verify this styling actually works in a running app -- as written, I believe the two bars will overlap when both are visible.

---

### Correctness: `subTasks` is typed as non-optional but guarded with `?.`

**File:** `task.component.ts`

```ts
subTaskProgress = computed<number>(() => {
    const subTasks = this.task().subTasks;
    if (!subTasks?.length) {
      return 0;
    }
    ...
});
```

Looking at the model, `TaskWithSubTasks.subTasks` is typed as `readonly Task[]` (non-optional, non-nullable). The optional chaining `?.` is defensive but unnecessary and slightly misleading about the type contract. This is a minor point -- the defensive check is harmless -- but for consistency with the strict typing conventions in this codebase, a simple `if (!subTasks.length)` or `if (subTasks.length === 0)` would be cleaner.

Similarly in the template:
```html
@if (t.subTasks?.length) {
```
The existing code already uses `t.subTasks?.length` for the subtask section below, so this is at least consistent with the current pattern.

---

### Suggestion: Avoid duplicate `@if` guard

**File:** `task.component.html`

The PR adds a new `@if (t.subTasks?.length)` block immediately before the existing identical `@if (t.subTasks?.length)` block:

```html
@if (t.subTasks?.length) {
  <progress-bar
    [cssClass]="'bg-success sub-task-progress'"
    [progress]="subTaskProgress()"
  ></progress-bar>
}
@if (t.subTasks?.length) {
  <div class="sub-tasks">
    ...
```

Consider merging these into a single `@if` block to avoid evaluating the same condition twice and to make the template more readable. The progress bar could be placed at the top of the existing `@if` block, just before the `<div class="sub-tasks">`.

---

### Minor: `hasVisibleTimeProgress` host class is added but may never be needed

The host binding `'[class.hasVisibleTimeProgress]': 'isCurrent() && progress() > 1'` is added to support the CSS offset rule. Since the CSS rule doesn't work due to encapsulation (see above), this binding currently adds overhead with no visible effect. Once the CSS approach is resolved, revisit whether this binding is still the right mechanism.

---

### What looks good

- The `subTaskProgress` computed signal is clean and follows the existing pattern of the `progress` computed signal nicely.
- Using the existing `ProgressBarComponent` and `bg-success` class keeps the implementation consistent with the design system.
- The feature correctly shows progress even when subtasks are collapsed, which is the key UX value.
- The change is small and well-scoped -- only 3 files touched with 21 additions.

---

### Summary

The main blocker is the **CSS encapsulation issue** -- the overlap-prevention style will not apply as written, meaning when a task is both current (showing the time progress bar) and has subtasks, the two bars will visually overlap. Please verify this in a running app and choose one of the suggested fixes. The duplicate `@if` block is a readability nit that would be nice to clean up.

---

johannesjo:
Thanks — a subtask progress indicator is genuinely wanted (#7057). A few blockers before this can land:

1. **Invisible at 0%.** `ProgressBarComponent` sets `visibility: hidden` when `progress <= 1`, so a task with 0-of-N subtasks done (and anything ≤1%) renders nothing — which defeats the "always visible" goal. Either give `ProgressBarComponent` an opt-in always-visible / empty-track mode, or reconsider the visual: a numeric `done/total` badge reusing the existing checklist-progress pattern would sidestep this entirely and read better.

2. **Overlap CSS is a no-op.** `:host.hasVisibleTimeProgress progress-bar.sub-task-progress::after { top: 0 }` can't reach the `::after` defined inside the child component under emulated view-encapsulation, so the time bar and subtask bar overlap when a current task has subtasks. The `hasVisibleTimeProgress` host binding then adds change-detection cost in the task hot path for no effect — please handle the offset inside `ProgressBarComponent` instead.

3. **Conflicting.** The branch conflicts with master (this area was touched by the newer checklist-progress feature) and needs a rebase.

4. **No tests** yet.

One note: `task.component.*` is a per-task hot path, so please keep `subTaskProgress` as a `computed()` (good) and avoid any per-CD function/getter calls in the template.

johannesjo (commented):
Thanks a lot for working on this, @symonbaikov — and sorry for the slow turnaround\! 🙏 The need behind #7057 is legit and the implementation is clean (appreciate the tests). After looking closely though, I'd like to steer this toward something smaller. Sorry to send you back to the drawing board a bit — these things sometimes only really reveal themselves once you see them running.

**Why not the bar:** subtasks are expanded by default, and in that state each subtask row already shows its done/undone status under the parent — so a parent-level bar mostly duplicates what's already on screen. The only real gap is when subtasks are **collapsed** (the +/− toggle), where there's currently no completion signal at all. That collapsed-with-many-subtasks case is exactly #7057.

**Proposed direction:** fold a `{done}/{total}` count (e.g. `3/15`) into the existing show/hide-subtasks toggle button, shown when subtasks are collapsed. We already do exactly this for markdown checklists — see the `.checklist-progress-badge` overlay on `.checklist-progress-btn` (`task.component.html` + `_task-controls.scss`): small 10px text, `--card-bg`, `extraBorder`, `pointer-events: none`, with a tooltip + aria-label. Reusing that on `toggle-sub-tasks-btn`:
- no new elements on the row (the toggle is already rendered per parent),
- much more subtle than a green line, and gives the exact counts the issue asked for,
- puts the info right on the affordance that reveals the subtasks.

Concretely: render the badge when `t._hideSubTasksMode === ShowSubTasksMode.HideAll` (the collapsed "+" state), keep the plain icon when expanded. Add a translation key (en.json only, e.g. `SUB_TASK_PROGRESS`) for the tooltip/aria-label, mirroring `CHECKLIST_PROGRESS`.

This lets us drop the heavier changes: the always-visible `<progress-bar>` instance on every parent row (the existing time bar is gated to `@if (isCurrent())`, and `task.component` is a hot path), plus the `isAlwaysVisible` / `isStackedBelowProgress` additions to the shared `ui/progress-bar`. The `sub-task-progress` class currently isn't styled anywhere either, so it's a no-op.

Good news: `calcSubTaskProgress` + its spec mostly carry over (return `done`/`total` instead of a %), and the progress-bar setter hardening (clamp + `Number.isFinite`) is a nice cleanup you could land as its own small PR.

Up for reworking along these lines? Happy to help if anything's unclear. Thanks again\! 🙏

PR #7121, opened 2026-04-03
fix(shortcuts): restore keyboard navigation flow

### LINKED ISSUE #7077: 🚨 Keyboard shortcuts

symonbaikov:
## Problem
Fixes #7077.
Several keyboard shortcuts in the work view and sidenav no longer behaved as expected:
- `W` could focus the first overdue task instead of the first task in the main undone list
- `C` opened an empty side panel instead of the filter/group/sort menu
- `Shift+D` focused the sidenav collapse button instead of the first navigation item
- there was no shortcut to expand/collapse the sidebar
## Solution
- Scoped "focus first task" to the undone task list and kept a fallback to the first visible task when no undone task is available
- Changed the filter/group/sort shortcut to open the existing task view customizer menu on both desktop and mobile instead of toggling an empty panel state
- Updated sidenav keyboard focus logic so it starts at actual navigation items and skips the collapse/expand toggle button
- Added a new `toggleSideNav` keyboard shortcut with default binding `Ctrl+/`
- Updated keyboard settings labels/config so "Focus sidenav" and "Expand/collapse sidebar" are separate actions
- Added regression tests for the new task focus behavior and the new layout trigger signals
## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)
## Checklist


johannesjo (commented):
## Review: fix(shortcuts): restore keyboard navigation flow

Overall this is a well-structured fix that addresses the reported keyboard navigation issues cleanly. The signal-based trigger pattern for cross-component communication is a good fit for the codebase's preference for signals over observables. Here are my findings:

### Dead code: `MobileSidePanelMenuComponent` changes have no effect

The `MobileSidePanelMenuComponent` (selector: `mobile-side-panel-menu`) is **not referenced in any template or dynamically loaded anywhere in the codebase**. The effect added in its constructor that listens to `openTaskViewCustomizerTrigger` will never execute. The changes to this component (adding `viewChild`, `MatMenuTrigger`, and the `effect` constructor) are dead code.

Either:
- This component is planned for future use and the changes are premature, or
- The component should be wired into a template (e.g., conditionally rendered in `main-header.component.html` for mobile viewports).

Please verify the intended mobile behavior is actually reachable. If the mobile menu trigger for `C` is supposed to work, the component needs to be rendered somewhere.

### Potential race condition in mobile menu trigger

In `MobileSidePanelMenuComponent`, the effect does:
```ts
this.isShowMobileMenu.set(true);
this.customizerMenuTrigger()?.openMenu();
```
Setting the menu visibility and immediately calling `openMenu()` in the same synchronous block may not work as expected -- the `MatMenuTrigger` button might not yet be rendered/visible in the DOM when `openMenu()` is called. Consider wrapping the `openMenu()` call in a `setTimeout` (similar to the `focusSideNavTrigger` effect pattern used in `MagicSideNavComponent`) or using `afterNextRender` to ensure the DOM has updated.

### `_isVisible` could be simplified

In `task.service.ts`, the new `_isVisible` method checks `display`, `visibility`, `opacity`, and `getClientRects().length`. The `getClientRects().length > 0` check already covers `display: none` (elements with `display: none` have zero client rects). Consider whether the redundant checks are intentional for clarity or could be trimmed.

### Signal increment pattern -- overflow is theoretical but worth noting

The trigger signals use `value + 1` to increment. In practice this will never overflow, but if you wanted to be extra defensive, `value ^ 1` (XOR toggle) or modular arithmetic would avoid any theoretical `Number.MAX_SAFE_INTEGER` concern. This is a very minor nit.

### `focusSideNav` label was previously `TOGGLE_SIDE_NAV` -- migration concern

The `focusSideNav` form field previously used `T.GCF.KEYBOARD.TOGGLE_SIDE_NAV` as its label. The diff changes this to the new `T.GCF.KEYBOARD.FOCUS_SIDE_NAV` key. This is correct semantically, but the old `TOGGLE_SIDE_NAV` translation key's text is also changed from "Focus sidenav" to "Expand/collapse sidebar" in `en.json`. This dual change is fine for English, but existing users who have configured custom bindings for the old `focusSideNav` shortcut will see the label change in settings -- which is the intended behavior. Just confirming this is deliberate.

### Good patterns observed

- Using `data-task-focus-scope="undone"` as a DOM marker to scope task focus is a clean, minimal approach that avoids coupling the shortcut service to Angular's component tree.
- The fallback from undone tasks to any visible task in `focusFirstTaskIfVisible` is good defensive behavior.
- Removing the `.mode-toggle` button from the keyboard-navigable elements in `_getFocusableNavElements` is correct -- the toggle now has its own dedicated shortcut (`Ctrl+/`), so including it in tab navigation would be redundant.
- Unit tests cover the new trigger signals and the `focusFirstTaskIfVisible` scoping logic well.

### Summary

The core fixes (task focus scoping, sidenav focus flow, customizer menu opening, new toggle shortcut) are correct and well-tested. The main concern is that the `MobileSidePanelMenuComponent` changes appear to be unreachable dead code -- please verify whether this component is actually rendered anywhere or if additional wiring is needed for the mobile keyboard shortcut to work.

---

Zoriot:
@symonbaikov can you resolve the conflicts here too?

johannesjo:
**♻️ Master has moved on** – this branch is 2204 commits behind `master`; see below for what still applies.

Thanks for this, and sorry it sat so long. Since the last push in June master grew its own sidebar-toggle shortcut and deleted or moved several of the files this touches, so most of the diff is now a duplicate or has nothing to attach to. Rather than review lines that are about to be rewritten, here is what a rebase would actually run into.

### Already on master

- **`layout.service.ts`** – `_toggleSideNavModeTrigger` and `toggleSideNavMode()` already exist (lines 58-59 and 319-320), so that half of the hunk is a duplicate. The `openTaskViewCustomizerTrigger` half is genuinely new and still applies.
- **`default-global-config.const.ts:149`** – master ships `toggleSideNavMode: 'Ctrl+B'` as the expand/collapse binding, and `keyboard-form.const.ts:52` already exposes it in settings. It has shipped: the frozen-state fixture for v18.15 carries `"toggleSideNavMode": "Ctrl+B"`. A new `toggleSideNav: 'Ctrl+/'` would be a second config key for the same action.
- **`shortcut.service.ts:184-185`** – the `keys.toggleSideNavMode` branch is already there.
- **`magic-side-nav.component.ts:197-204`** – master already reacts to `toggleSideNavModeTrigger` in an effect. Adding a second one would toggle the sidebar twice per keypress, i.e. net nothing.

### Gone from master

- **`main-header/mobile-side-panel-menu/`** – removed in `6bb0472549` (shipped in v18.14.0), so the changes to `mobile-side-panel-menu.component.ts` and the new `<mobile-side-panel-menu>` in `main-header.component.html` / `.ts` have nothing to attach to.
- **`features/config/keyboard-config.model.ts`** – `KeyboardConfig` now lives in `electron/shared-with-frontend/keyboard-config.model.ts`, reached through the `@sp/keyboard-config` path alias.

### Still applies

- **`magic-side-nav.component.ts`** – `_getFocusableNavElements()` still pushes `.mode-toggle` in first (lines 534-545) and `_focusFirstNavElement()` focuses index 0, so `Shift+D` still lands on the collapse button on desktop. This hunk applies cleanly.
- **`page-title.component.ts`** – still a real bug: `shortcut.service.ts:231` is the only remaining caller of `toggleTaskViewCustomizerPanel()`, and `right-panel-content.component.html` has had no `TASK_VIEW_CUSTOMIZER_PANEL` branch since `c45080527e` (v16.1.0), so on desktop `C` opens an empty right panel. Pointing `C` at the existing customizer menu is the right fix. The desktop trigger now lives around `page-title.component.ts:82-109`, so the hunk needs re-siting.

### Two things to watch when you rebase

- **`C` would stop working on mobile.** Today on master, `C` dispatches the panel state and `bottom-panel-container.component.html:45` renders `<task-view-customizer-panel>` in the bottom sheet, so it works below 600px. With this change the dispatch is gone and the only surviving handler is the `page-title` effect, gated `!this._layoutService.isXs()` – so on xs, `C` would do nothing at all. Worth noting the `page-title` button itself renders down to 350px (`@if (!isXxxs() && ...)`, `page-title.component.ts:61`), so `!isXs()` is a narrower gate than the button's own visibility even setting mobile aside.
- **The `en.json` change would mislabel an existing setting.** On master, `keyboard-form.const.ts:51` still labels the `focusSideNav` field with `T.GCF.KEYBOARD.TOGGLE_SIDE_NAV` ("Focus sidenav"), and `:52` labels `toggleSideNavMode` with `TOGGLE_SIDE_NAV_MODE`. Redefining `TOGGLE_SIDE_NAV` to "Expand/collapse sidebar" without also switching `:51` to the new `FOCUS_SIDE_NAV` key leaves two rows that both read like sidebar expand/collapse, no row labelled "Focus sidenav", and `FOCUS_SIDE_NAV` unused.

### Question

- I could not reproduce the `W` behaviour the `task.service.ts` change targets. In `work-view.component.html` every undone `task-list` (161, 178, 245) is emitted before the overdue panel (254), and `<task>` only comes from `task-list.component.html:40`, so `getElementsByTagName('task')[0]` should already be the first undone task whenever one is rendered – and that ordering is the same at this branch's base. What layout were you in when `W` landed on an overdue task? If there is a case I am missing the scoping is worth keeping; if not, that hunk can go. (If it stays: master now renders undone tasks from three branches, including a `no-section` list and one `task-list` per section at `work-view.component.html:178` and `:235`, so the marker would need to cover those too or the scoped selector matches nothing in the default sections layout.)

Rebased, this looks like roughly two small changes plus the `openTaskViewCustomizerTrigger` plumbing, rather than a 16-file diff. Given how much has moved I'd hold off until a maintainer says whether it's wanted in that reduced form – the linked issue #7077 was closed as *not planned* back in April, so the original motivation is worth re-confirming first.

---

Issue #3818, opened 2025-01-03
Home screen widget


Amirth24:
## Problem Statement

This is an amazing app to manage tasks and help me be productive. I think this application becomes even better with home screen widgets. 

## :grey_question: Possible Solution

I have some idea for widgets

- A widget only with the task prompt.
- A Google Calendar like widget
- Metrics dashboard (not so useful though) 
- Schedule Chart

johannesjo:
PR's are welcome! Heads up: the data exchange part might be a bit tricky and will probably would require a custom solution to sync data between app and widget.

eclecticpassions:
I really hope someone with the right knowledge can help you out with an Android widget, @johannesjo!

Scribble4:
This is the one feature that I would need to fully switch to super productivity (currently using ticktick) glad to see that others are interested in it. @johannesjo please consider this

pdario:
+1! I'm looking for a substitute to my current solution, but I need a home widget

johannesjo:
> home widget

For android?

eclecticpassions:
For me, yes. Android please!

> > home widget
> 
> For android?

zitongcharliedeng:
+1 for Android Widget. Wish there was just a way to pin an app window to home screen without root...

Pushkin31:
+1. We are really looking forward to the widget for the home screen on android!!!

Gymnaar:
I am waiting so long for this. I know it is complicated but for a lot of people it is the last thing holding them back to other apps.

KurtMage:
+1 for android widget!

GastricBoar:
+1 looking onto this :)

abu-dojana:
I've looked into this. Since the Android app is a container relying on the WebView's IndexedDB, the native Widget API can't access the data directly.

I propose implementing a **Shadow Database** pattern using a local SQLite database (likely via Room) on the native Android side.

### The Logic:

- We create a Capacitor plugin that acts as a bridge.

- When data changes in the Web app, it syncs/pushes the relevant task data to the native SQLite DB.

- The Android Widget effectively becomes a view for this local SQLite DB.

This approach allows for complex queries (like filtering by specific lists/tags similar to TickTick) without needing to rewrite the core logic or expose the entire IndexedDB.

### The Proposed Architecture: Bidirectional Sync
Instead of the Widget talking directly to the Web DB (impossible), or the Web DB replacing its engine (too hard), we create a sync protocol.

[...]

PR #7124, opened 2026-04-03
POC for Android widgets (for architectural review)

ilvez:
## Problem

Implements an Android home screen widget for interacting with tasks without opening the app (#3818).

## Solution

**This is a proof of concept — not intended to be merged as-is.** We'd like feedback on the architectural approach and whether this direction makes sense before investing further.

### What it does (enough to verify the approach works)

- Shows today's tasks on a native Android home screen widget
- Project color indicator per task
- "Mark done" checkbox with immediate visual feedback
- Tap header to open app
- Widget refreshes on app state change, on pause, and every 30 min (Android minimum)

### Technical approach

The core problem is that the app state lives in IndexedDB inside a Capacitor WebView, which native widget code cannot access — there's no viable way to read IndexedDB from native Android without a live WebView.

**Chosen solution:** Use the existing `KeyValStore` (SQLite) as a shared data layer between Angular and the native widget. Angular pushes a compact JSON snapshot of today's tasks to KeyValStore via the existing `saveToDb` bridge. The native `RemoteViewsFactory` reads it directly. This avoids introducing new storage mechanisms — KeyValStore is already the established persistence bridge between the WebView and native code.

**Done actions flow back** via a `SharedPreferences`-backed queue (`WidgetDoneQueue`), following the existing `ReminderDoneQueue` pattern. Two delivery paths cover all states:
- **App alive:** `LocalBroadcast` → `CapacitorMainActivity` → pushes to WebView instantly
- **App dead:** Queue accumulates, drained on next resume or cold start via `ReplaySubject`

**Why not other approaches:**
- **ContentProvider:** Unnecessary complexity for same-process communication. Both widget and app share the same process, so `@Synchronized` on KeyValStore is sufficient.
- **Direct IndexedDB access from native:** Not feasible without a live WebView instance.
- **WorkManager for periodic sync:** The widget data is a UI snapshot, not persistent state. Eager refresh on state change + onPause covers active use; the 30-min periodic update is just a fallback.

### Known drawbacks

- Widget data reflects the app's last known state — changes from other sync clients appear after the app is next opened (which triggers sync and widget refresh). The existing `SyncReminderWorker` only manages native alarms, it does not apply full state changes or update KeyValStore.
- No undo for "mark done" from widget
- Hardcoded dark styling, no Material You / dynamic color theming
- `RemoteViews` has severe layout limitations (no custom views, limited styling options)
- No task creation from widget (only mark done)
- If user marks done while app is alive, both push (broadcast) and pull (resume queue) paths could fire — relies on `setDone()` being idempotent on already-done tasks

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -149,5 +149,24 @@
       </intent-filter>
     </receiver>
 
+    <receiver
+      android:name=".widget.TaskListWidgetProvider"
+      android:exported="true"
+      android:label="@string/widget_title">
+      <intent-filter>
+        <action android:name="android.appwidget.action.APPWIDGET_UPDATE" />
+        <action android:name="com.superproductivity.superproductivity.WIDGET_MARK_DONE" />
+        <action android:name="com.superproductivity.superproductivity.WIDGET_OPEN_APP" />
+      </intent-filter>
+      <meta-data
+        android:name="android.appwidget.provider"
+        android:resource="@xml/appwidget_info" />
+    </receiver>
+
+    <service
+      android:name=".widget.TaskListWidgetService"
+      android:exported="false"
+      android:permission="android.permission.BIND_REMOTEVIEWS" />
+
   </application>
 </manifest>
--- a/android/app/src/main/java/com/superproductivity/superproductivity/CapacitorMainActivity.kt
+++ b/android/app/src/main/java/com/superproductivity/superproductivity/CapacitorMainActivity.kt
@@ -29,6 +29,7 @@ import com.superproductivity.superproductivity.webview.WebViewBlockActivity
 import com.superproductivity.superproductivity.webview.WebViewCompatibilityChecker
 import com.superproductivity.superproductivity.widget.ShareIntentQueue
 import com.superproductivity.superproductivity.widget.StartupOverlayManager
+import com.superproductivity.superproductivity.widget.TaskListWidgetProvider
 import com.superproductivity.plugins.webdavhttp.WebDavHttpPlugin
 import org.json.JSONObject
 
@@ -52,6 +53,17 @@ class CapacitorMainActivity : BridgeActivity() {
                 val isBreak = intent.getBooleanExtra(FocusModeForegroundService.EXTRA_IS_BREAK, false)
                 Log.d("SP_FOCUS", "Timer complete broadcast received, isBreak=$isBreak")
                 callJSInterfaceFunctionIfExists("next", "onFocusModeTimerComplete$", isBreak.toString())
+            }
+        }
+    }
+
+    private val widgetDoneReceiver = object : BroadcastReceiver() {
+        override fun onReceive(context: Context?, intent: Intent?) {
+            if (intent?.action == TaskListWidgetProvider.ACTION_WIDGET_DONE_LOCAL) {
+                val taskId = intent.getStringExtra(TaskListWidgetProvider.EXTRA_TASK_ID) ?: return
+                val sanitizedId = taskId.replace(Regex("[^a-zA-Z0-9_-]"), "")
+                Log.d("SP_WIDGET", "Widget done broadcast received: taskId=$sanitizedId")
+                callJSInterfaceFunctionIfExists("next", "onWidgetDone$", "'$sanitizedId'")
             }
         }
     }
@@ -169,6 +181,12 @@ class CapacitorMainActivity : BridgeActivity() {
         LocalBroadcastManager.getInstance(this).registerReceiver(
             timerCompleteReceiver,
             IntentFilter(FocusModeForegroundService.ACTION_TIMER_COMPLETE)
+        )
+
+        // Register broadcast receiver for widget done actions
+        LocalBroadcastManager.getInstance(this).registerReceiver(
+            widgetDoneReceiver,
+            IntentFilter(TaskListWidgetProvider.ACTION_WIDGET_DONE_LOCAL)
         )
 
         // Show startup overlay for quick task entry while Angular loads.
@@ -354,6 +372,7 @@ class CapacitorMainActivity : BridgeActivity() {
         startupOverlayManager = null
         super.onDestroy()
         LocalBroadcastManager.getInstance(this).unregisterReceiver(timerCompleteReceiver)
+        LocalBroadcastManager.getInstance(this).unregisterReceiver(widgetDoneReceiver)
     }
 
     companion object {
--- a/android/app/src/main/java/com/superproductivity/superproductivity/app/KeyValStore.kt
+++ b/android/app/src/main/java/com/superproductivity/superproductivity/app/KeyValStore.kt
@@ -45,7 +45,6 @@ class KeyValStore(private val context: Context) :
             values.put(KEY_CREATED_AT, "time('now')")
             row = db.replace(DATABASE_TABLE, null, values)
             Log.v(TAG, "save db value size: " + value?.length)
-            db.close()
         }
         return row
     }
@@ -64,14 +63,12 @@ class KeyValStore(private val context: Context) :
         dbHelper.readableDatabase?.let { database ->
             database.query(
                 DATABASE_TABLE, arrayOf(VALUE), "$KEY=?", arrayOf(newKey), null, null, null
-            )?.let { cursor ->
+            )?.use { cursor ->
                 if (cursor.moveToNext()) {
                     value = cursor.getString(cursor.getColumnIndexOrThrow(VALUE))
                 }
                 Log.v(TAG, "get db value size:" + value.length)
-                cursor.close()
             }
-            database.close()
         }
         return value
     }
@@ -82,7 +79,6 @@ class KeyValStore(private val context: Context) :
         if (db != null) {
             db.delete(DATABASE_TABLE, null, null)
             Log.v(TAG, "cleared db ")
-            db.close()
         }
     }
 
--- a/android/app/src/main/java/com/superproductivity/superproductivity/webview/JavaScriptInterface.kt
+++ b/android/app/src/main/java/com/superproductivity/superproductivity/webview/JavaScriptInterface.kt
@@ -22,6 +22,8 @@ import com.superproductivity.superproductivity.widget.ReminderDoneQueue
 import com.superproductivity.superproductivity.widget.ReminderSnoozeQueue
 import com.superproductivity.superproductivity.widget.ReminderTapQueue
 import com.superproductivity.superproductivity.widget.ShareIntentQueue
+import com.superproductivity.superproductivity.widget.TaskListWidgetProvider
+import com.superproductivity.superproductivity.widget.WidgetDoneQueue
 import com.superproductivity.superproductivity.widget.WidgetTaskQueue
 
 
@@ -237,6 +239,18 @@ class JavaScriptInterface(
     @JavascriptInterface
     fun getWidgetTaskQueue(): String? {
         return WidgetTaskQueue.getAndClearQueue(activity)
+    }
+
+    @Suppress("unused")
+    @JavascriptInterface
+    fun getWidgetDoneQueue(): String? {
+        return WidgetDoneQueue.getAndClear(activity)
+    }
+
+    @Suppress("unused")
+    @JavascriptInterface
+    fun updateWidget() {
+        TaskListWidgetProvider.notifyDataChanged(activity)
     }
 
     /**
--- a/android/app/src/main/java/com/superproductivity/superproductivity/widget/TaskListWidgetProvider.kt
+++ b/android/app/src/main/java/com/superproductivity/superproductivity/widget/TaskListWidgetProvider.kt
@@ -1,0 +1,136 @@
+package com.superproductivity.superproductivity.widget
+
+import android.app.PendingIntent
+import android.appwidget.AppWidgetManager
+import android.appwidget.AppWidgetProvider
+import android.content.ComponentName
+import android.content.Context
+import android.content.Intent
+import android.net.Uri
+import android.util.Log
+import android.widget.RemoteViews
+import androidx.localbroadcastmanager.content.LocalBroadcastManager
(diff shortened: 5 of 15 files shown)


johannesjo (commented):
## Architectural Review -- Android Widget POC

Nice work for a first PR on this project. The overall approach is well-considered and the PR description is exceptionally thorough. Here is a detailed review focused on architecture, correctness, and alignment with existing codebase patterns.

---

### What works well

1. **KeyValStore as shared data layer** -- This is the right call. Reusing the existing SQLite bridge avoids introducing a new storage mechanism and keeps the widget/WebView contract simple. The compact JSON snapshot (`widget_data`) with a version field (`v: 1`) is forward-thinking.

2. **Dual delivery for done actions (push + pull)** -- The `LocalBroadcast` (app alive) + `SharedPreferences` queue (app dead) pattern correctly mirrors the existing `ReminderDoneQueue` / `onReminderDone$` pattern. This is well-matched to the codebase conventions.

3. **HydrationStateService guard** -- Both `WidgetDataService.serialize()` and `pushOnStateChange$` correctly check `isApplyingRemoteOps()`, preventing widget updates during remote sync replay. This follows the guidance in CLAUDE.md item 8.

4. **KeyValStore resource leak fixes** -- Removing manual `db.close()` / `cursor.close()` calls and using `.use {}` for the cursor is a genuine improvement. `SQLiteOpenHelper` manages database lifecycle; manually closing between calls was causing "database already closed" risks under concurrent access from the widget.

5. **Input sanitization** -- The `widgetDoneReceiver` in `CapacitorMainActivity` sanitizes the task ID before injecting it into the JS bridge call. Good defensive practice.

---

### Architectural concerns

#### 1. Task title tap is broken (bug)

In `TaskListRemoteViewsFactory.getViewAt()`, the title's fill-in intent sets `EXTRA_OPEN_APP = true` but no `EXTRA_TASK_ID`:

```kotlin
val openIntent = Intent().apply {
    putExtra(TaskListWidgetProvider.EXTRA_OPEN_APP, true)
}
rv.setOnClickFillInIntent(R.id.widget_task_title, openIntent)
```

Since the `PendingIntentTemplate` for the entire `ListView` uses `ACTION_MARK_DONE`, the combined intent will fire `ACTION_MARK_DONE` in `onReceive`. There, `getStringExtra(EXTRA_TASK_ID)` returns null, so the handler returns early. **Tapping a task title does nothing.**

To fix this, you would need either:
- A second `PendingIntentTemplate` (not possible -- `RemoteViews` only supports one per `ListView`), or
- Handle `EXTRA_OPEN_APP` in the `ACTION_MARK_DONE` branch: if `taskId` is null but `EXTRA_OPEN_APP` is true, launch the activity. Or use a different approach where the entire row launches the app and only the checkbox does "mark done."

#### 2. Selector-based effect without `LOCAL_ACTIONS` -- acceptable here but document the reasoning

`pushOnStateChange$` is a selector-based effect (uses `combineLatest` on store selectors rather than `this._actions$.pipe(ofType(...))`). Per CLAUDE.md item 8, these should be avoided because they fire during hydration/sync replay. You correctly guard with `filter(() => !this._hydrationState.isApplyingRemoteOps())`, which is the documented escape hatch.

However, there is a subtle timing issue: `combineLatest` will emit once hydration finishes and `isApplyingRemoteOps()` returns `false` again, causing a serialization. This is probably fine (you *want* the widget updated after hydration), but it is worth a brief comment explaining this is intentional.

#### 3. Duplicate done processing on resume when app is alive

The PR description acknowledges this: both the `LocalBroadcast` (push) and `drainWidgetDoneQueueOnResume$` (pull) paths can fire for the same task. The `onWidgetDone$` ReplaySubject will emit the same task ID twice, and `setDone()` will be called twice.

While `setDone()` is likely idempotent at the task level, it will dispatch an NgRx action twice, which means:
- Two operations in the sync log for the same state change
- Two snack notifications ("Task marked done from widget") shown to the user

Consider either:
- Having the `LocalBroadcast` path also clear the queue entry for that task ID, or
- Deduplicating in `handleWidgetDone$` (e.g., track recently-processed IDs with a small `Set` that clears after a few seconds)

#### 4. `toPromise()` is deprecated

`WidgetDataService.serialize()` uses `.toPromise()` three times. While the existing codebase still has ~166 usages, newer code in this repo uses `firstValueFrom` from `rxjs`. Since this is new code, it would be cleaner to use `firstValueFrom` from the start.

#### 5. Widget data could include sub-task structure

The current snapshot only includes top-level task IDs from `selectTodayTaskIds`. If a parent task has sub-tasks, the widget shows the parent but the user cannot mark individual sub-tasks as done. This may be intentional for the POC, but worth noting for the future -- users who break tasks into sub-tasks might expect to see and interact with them.

#### 6. `hasStableIds()` returns `false`

`TaskListRemoteViewsFactory.hasStableIds()` returns `false` and `getItemId()` returns `position.toLong()`. This means Android cannot optimize list updates -- every `notifyAppWidgetViewDataChanged` redraws the entire list. For a POC this is fine, but for production, returning `true` and using a hash of `task.id` as the item ID would give smoother updates (especially after marking a task done, where only one row changes).

---

### Minor issues

- **`EXTRA_OPEN_APP` constant is unused** -- It is defined in the companion object but never checked in `onReceive`. Related to bug #1 above.
- **Hardcoded 20-task limit** in `TaskListRemoteViewsFactory.onDataSetChanged()` -- Reasonable for a widget, but should be a named constant rather than a magic number.
- **Empty `contentDescription`** on `widget_project_dot` ImageView -- Accessibility tools will flag this. Even decorative elements should use `android:importantForAccessibility="no"` instead of an empty string.
- **`selectAllProjectColorsAndTitles`** is subscribed to in `pushOnStateChange$` (in the `combineLatest`) but the value (`_projects`) is discarded. It is only used to trigger re-serialization when projects change. This works but is non-obvious -- a comment would help. Alternatively, consider subscribing only to the project colors/titles for projects referenced by today's tasks, to reduce unnecessary emissions.

---

### Summary

The architectural approach -- KeyValStore as shared data layer, push/pull dual delivery, compact JSON snapshot -- is sound and well-aligned with how the codebase already bridges native and WebView concerns. The main issues are the broken title-tap interaction (bug), potential duplicate done processing, and some minor polish items. The `KeyValStore` resource management fixes are a nice bonus.

For a POC aimed at validating the approach: the direction makes sense. The core data flow (Angular state -> JSON snapshot -> KeyValStore -> native RemoteViews, and widget actions -> queue -> Angular) is clean and follows established patterns in this codebase.

---

johannesjo:
Thanks for the detailed writeup — the tradeoff analysis is really helpful. Overall I think **the direction is right and the PR is close to a mergeable foundation**. A few thoughts for the architectural review you asked for:

### What I think is sound

- Reusing `KeyValStore` as the bridge is the correct call. ContentProvider adds same-process overhead for no benefit, a keep-alive WebView would wreck battery, and a second native DB duplicates infrastructure you already have.
- Mirroring the `ReminderDoneQueue` pattern keeps the native surface consistent.
- Guarding the effect with `HydrationStateService.isApplyingRemoteOps()` and debouncing at 500 ms are the right instincts for this codebase.

### Things I'd address before landing (beyond POC)

1. **Dispatch amplification in `pushOnStateChange$`.** Subscribing to `selectTaskEntities` means every task mutation anywhere in the app (notes edit, time-tracking tick, anything) runs the `map`. `distinctUntilChanged` hides it from downstream but the CPU is still spent. A memoized selector that projects only the today-subset (`todayTaskIds → {id,title,isDone,projectId}[]`) would cut this substantially.

2. **The dual delivery path duplicates `setDone()` when the app is alive.** The broadcast triggers `setDone()`, but the task also sits in `SharedPreferences` and gets drained on the next resume, triggering `setDone()` again. Idempotent for the boolean today — but fragile if `setDone` ever records completion time, nudges time tracking, or emits analytics. Two cleaner shapes:
   - **Queue-only:** broadcast is just a "drain now" signal; Angular always reads from the queue. One path.
   - **Ack-after-apply:** Angular calls back into a JS interface method to clear specific IDs once applied.

   I'd lean toward queue-only for simplicity.

3. **Native-side optimistic mutation of `widget_data`** in `markDoneInWidgetData` races with Angular's serializer. The 500 ms debounce mostly hides it, but a concurrent unrelated task edit during that window can re-serialize with stale `isDone:false`, flipping the checkbox back briefly. Either let Angular own the blob and accept the small visual lag, or guard writes with a monotonic timestamp so older writes can't overwrite newer.

4. **Typed contract for the blob.** The v:1 schema now lives implicitly in two languages. The rest of the codebase uses typia validators — a writer-side validator on Angular would catch drift cheaply.

5. **Freshness from other sync clients** (already in your drawbacks list) is the biggest remaining UX gap. A `WorkManager` job that calls SuperSync and writes `KeyValStore` would close it — but that's its own feature and shouldn't block this.

### Worth considering as a follow-up, not a blocker

**Jetpack Glance** for the view layer. It replaces `RemoteViews` with a Compose DSL, gets you Material You / dynamic color, and sidesteps the "no custom views, limited styling" drawback you called out. The data architecture in this PR stays identical — it's purely a view-layer upgrade.

Happy to dig into any of these in more detail if useful.

ilvez:
Super thanks! I'll get back to it hopefully soon enough.

johannesjo:
Thank you @ilvez ! There is no rush :)

ilvez:
Since this PR was only for getting initial feedback. I'm closing it for clean review in actual PR.

johannesjo:
Revisited this with fresh eyes (and ran it past a couple of independent reviews) — wanted to leave a concrete punch-list so it's easy to pick back up, since the foundation here is genuinely good and I'd love to see it land.

**TL;DR:** architecture fully holds up. The KeyValStore-as-bridge + snapshot + queued-actions model is the right one — three separate reviews independently arrived at exactly this design. Below is everything I'd want addressed before merge, roughly in priority order. Several are the points from my April review; a few are new from reading the diff line-by-line.

### 🔴 Bugs

1. **Tapping a task title doesn't open the app.** In `TaskListRemoteViewsFactory.getViewAt`, the title gets a fill-in intent with `EXTRA_OPEN_APP`, but a collection has only one `setPendingIntentTemplate` — the `ACTION_MARK_DONE` one. So a title tap fills into the *done* template, `onReceive` finds no `EXTRA_TASK_ID`, and returns. Fix: make the row fill-in intents carry an explicit marker and branch in `onReceive` — e.g. if the fill-in has `EXTRA_OPEN_APP`, `startActivity(CapacitorMainActivity)`, else if it has `EXTRA_TASK_ID`, enqueue done. One template, two outcomes driven by the fill-in extras.

### 🟠 Should fix before merge

2. **Exported receiver exposes the custom actions.** `AndroidManifest.xml` adds `WIDGET_MARK_DONE` / `WIDGET_OPEN_APP` to the `exported="true"` intent-filter. The widget's own PendingIntents use an explicit component, so they're delivered without needing a filter entry — but listing them on an exported receiver lets any app on the device broadcast them (launch us, or mark known task IDs done). Remove the two custom `<action>` lines; keep only `APPWIDGET_UPDATE`.
3. **Done action can fire twice (go queue-only).** When the app is alive, `onReceive` enqueues the ID in `WidgetDoneQueue` *and* sends a `LocalBroadcast` that triggers `setDone()` now — and the ID is still drained on the next resume, calling `setDone()` again. Idempotent today, fragile the moment `setDone` records a completion time / nudges tracking / emits anything. Cleanest shape: broadcast becomes a contentless "drain now" signal; Angular always reads from the queue and clears it. Single path, and it also retires the fragile `'$sanitizedId'` string-interpolated JS call.
4. **Native optimistic write races the serializer.** `markDoneInWidgetData` rewrites `widget_data` for instant checkbox feedback, but a concurrent `serialize()` can overwrite it with stale `isDone:false` within the 500 ms window. The blob already carries `ts` — guard writes so an older `ts` can't clobber a newer one (or let Angular own the blob and accept a small visual lag).

### 🟡 Nice to address

5. **Dispatch amplification.** `pushOnStateChange$` subscribes to `selectTaskEntities`, so every task mutation app-wide runs the projection (the `distinctUntilChanged` hides the emission, not the CPU). A memoized selector projecting just the today-subset (`todayTaskIds → {id,title,isDone,projectId}[]`) fixes both this and gives you the typed contract below. Also: `selectAllProjectColorsAndTitles` is pulled into the `combineLatest` but unused (`_projects`), and project-color changes won't reach the widget since color isn't in the distinct key.
6. **Typed contract for the blob.** The `v:1` schema lives implicitly in TS (`widget-data.service.ts`) and is hand-parsed in Kotlin twice. A typia validator on the writer side documents and guards it cheaply, consistent with the rest of the codebase.
7. **`KeyValStore.close()` removal — confirm the backup path.** Dropping `db.close()` is the correct SQLiteOpenHelper pattern and needed for concurrent widget reads, and switching `get()` to `.use { cursor }` is a clean improvement. But `KeyValStore` also holds the full `AppDataComplete` backup ring (the Android eviction-recovery path), so a quick round-trip test that backup write/read still works would de-risk it.
8. **Tests + a couple of small things.** No unit tests yet — the `serialize()` projection and the done-queue drain are both easily testable and the repo expects tests for new services/state. Minor: `toPromise()` is deprecated (use `firstValueFrom`); widget chrome strings are native-only (`strings.xml`), so they won't follow the app's language — acceptable for a native widget, just flagging.

### Follow-ups, explicitly *not* blockers

- **Richer snapshot:** current task, focus/tracking state, and counts would make the widget much more useful (and enable a small "current task" widget variant) — but the today-list MVP is the right thing to ship first.
- **Jetpack Glance** as a pure view-layer upgrade (Material You / dynamic color, escapes the RemoteViews styling limits). Data architecture stays identical.
- **Cross-client freshness** (widget reflecting other devices' changes while the app is closed) via a WorkManager job that hits SuperSync and writes `KeyValStore` — its own feature, phase 2.

Really nice work — happy to pair on any of these, and no rush.

Issue #7114, opened 2026-04-02
💡 Linux/Wayland Global Shortcuts, Quick Entry, & CLI Commands


JamesMowery:
### Problem Statement

Linux/Wayland currently does not support the concept of "global hotkeys" by design and instead leaves the task of implementing such things to the Desktop Environment / Window Manager.

In other words, these types of actions must be configured by hand by the user for security.

Linux apps typically work within this framework by allowing the application to be invoked from the command line / terminal with flags/commands.

### Possible Solution

The suggestion would be to utilize a CLI-based approach to improve the launching behavior of Super Productivity. This would give all Desktop Environments and Window Managers an easy way to interface and trigger functionality for Super Productivity.

(Other potential benefits: enables automation/scripting support, AI interaction, and third-party app interaction.)

This is one possible suggestion for how you could accommodate this:

### Implementing Existing Global Shortcut Functionality

**Function | Command Line**
- Show/Hide Super Productivity | `superproductivity --toggle-visibility`
- Toggle time tracking for last active task | `superproductivity --toggle-time-tracking`
- Add a new note | `superproductivity --new-note`
- Add new task | `superproductivity --new-task`

### Newly Requested Shortcut Functionality for Quick Entry

**Function | Command Line**
- Quick task entry | `superproductivity --quick-task` (only shows minimal quick task entry screen; pressing `Return` submits the task and makes quick entry disappear without any further interruption)
- Quick note entry | `superproductivity --quick-note` (only shows minimal quick note entry screen; pressing `Return` submits the note and disappears the quick entry; Shift + Return allows you to add line breaks)

**Concept**

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/c8e28ff6-1eef-441a-b409-f4348ba613af" />

In this concept...
1. I manually type `superproductivity --quick-task` into the bottom left terminal.
2. The quick entry shown in the middle of the screen pops up
3. The cursor/text input immediately goes into the popped up quick entry to immediately start typing a new task
4. Return would submit the task, and everything would disappear so I can resume working on what I was previously doing without interruption. (Meaning: do not go or open to the Super Productivity window.)

In reality (and how this is how it applies to Global Shortcuts on Wayland)

1. I create a custom keybind in Niri configuration: `binds { Mod+Shift+T { spawn "super-productivity" "--quick-task"; } }`
2. I press Mod + Shift + T anywhere on my desktop, and it brings up the quick entry task window
[...]

PR #7136, opened 2026-04-04
Implements & fixes desktop CLI commands/flags

JamesMowery:
## Problem

Super Productivity had several desktop entry paths that behaved inconsistently or were hard to automate from the outside:

- CLI flags and protocol URLs were not handled through one shared command path
- forwarded command-line launches printed noisy success output instead of staying quiet on success
- detached CLI launches were needed so the shell returns immediately when using desktop commands
- there was not dedicated Electron-side test coverage for command parsing and execution behavior
- No way to invoke "hotkeys" within Wayland

Related PR's here:

## Solution

This PR unifies desktop command handling in the Electron layer.

Changes included:
- enable `--new-task`, `--new-note`, `--toggle-visibility`, `--toggle-time-tracking` CLI flags, which can be executed via Wayland keybindings in Desktop Environments and Window Managers to replicate global hotkeys.
- add a shared desktop command parser/executor flow for CLI flags, protocol URLs, pending startup commands, and global shortcut dispatch
- add Electron-side tests that cover parsing, queuing, flushing, and command execution behavior
- improve CLI feedback so valid forwarded commands stay silent while invalid command-like flags still report an error
- detach valid CLI desktop-command launches so they run in the background instead of taking over the terminal
- improve the visibility toggle raise/focus path so showing an existing instance works more reliably

Follow-up fixes for note/task handoff and Linux/Wayland global show/hide behavior are prepared on separate branches so this PR stays focused.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

Validation notes:
- `.deb` builds passed locally during validation

### DIFF
--- a/electron/desktop-command-executor.ts
+++ b/electron/desktop-command-executor.ts
@@ -1,0 +1,104 @@
+import { log } from 'electron-log/main';
+import { DesktopCommand } from './desktop-command';
+import { IPC } from './shared-with-frontend/ipc-events.const';
+
+export interface DesktopCommandWindow {
+  blur(): void;
+  hide(): void;
+  isFocused(): boolean;
+  webContents: {
+    send(channel: IPC, ...args: unknown[]): void;
+  };
+}
+
+export interface DesktopCommandDeps {
+  showOrFocus: (mainWin: DesktopCommandWindow) => void;
+}
+
+const pendingDesktopCommands: DesktopCommand[] = [];
+
+export const executeDesktopCommand = (
+  command: DesktopCommand,
+  mainWin: DesktopCommandWindow,
+  { showOrFocus }: DesktopCommandDeps,
+): void => {
+  switch (command.type) {
+    case 'toggle-visibility':
+      if (mainWin.isFocused()) {
+        mainWin.blur();
+        mainWin.hide();
+      } else {
+        showOrFocus(mainWin);
+      }
+      return;
+
+    case 'toggle-time-tracking':
+      mainWin.webContents.send(IPC.TASK_TOGGLE_START);
+      return;
+
+    case 'new-note':
+      showOrFocus(mainWin);
+      mainWin.webContents.send(IPC.ADD_NOTE);
+      return;
+
+    case 'new-task':
+      showOrFocus(mainWin);
+      mainWin.webContents.send(IPC.SHOW_ADD_TASK_BAR);
+      return;
+
+    case 'create-task':
+      showOrFocus(mainWin);
+      mainWin.webContents.send(IPC.ADD_TASK_FROM_APP_URI, { title: command.title });
+      return;
+  }
+};
+
+export const queueDesktopCommand = (command: DesktopCommand): void => {
+  pendingDesktopCommands.push(command);
+};
+
+export const flushPendingDesktopCommands = ({
+  getMainWindow,
+  isAppReady,
+  showOrFocus,
+}: {
+  getMainWindow: () => DesktopCommandWindow | null | undefined;
+  isAppReady: () => boolean;
+  showOrFocus: (mainWin: DesktopCommandWindow) => void;
+}): void => {
+  const mainWin = getMainWindow();
+  if (!mainWin || !isAppReady() || pendingDesktopCommands.length === 0) {
+    return;
+  }
+
+  log(`Processing ${pendingDesktopCommands.length} pending desktop command(s)`);
+  const commands = [...pendingDesktopCommands];
+  pendingDesktopCommands.length = 0;
+  commands.forEach((command) => executeDesktopCommand(command, mainWin, { showOrFocus }));
+};
+
+export const queueOrExecuteDesktopCommand = ({
+  command,
+  getMainWindow,
+  isAppReady,
+  showOrFocus,
+}: {
+  command: DesktopCommand;
+  getMainWindow: () => DesktopCommandWindow | null | undefined;
+  isAppReady: () => boolean;
+  showOrFocus: (mainWin: DesktopCommandWindow) => void;
+}): void => {
+  const mainWin = getMainWindow();
+  if (!mainWin || !isAppReady()) {
+    queueDesktopCommand(command);
+    return;
+  }
+
+  executeDesktopCommand(command, mainWin, { showOrFocus });
+};
+
+export const getPendingDesktopCommandCount = (): number => pendingDesktopCommands.length;
+
+export const resetPendingDesktopCommands = (): void => {
+  pendingDesktopCommands.length = 0;
+};
--- a/electron/desktop-command-parser.ts
+++ b/electron/desktop-command-parser.ts
@@ -1,0 +1,84 @@
+import { DesktopCommand, DesktopCommandParseResult } from './desktop-command';
+
+const ACTION_FLAG_COMMAND_PAIRS: ReadonlyArray<
+  readonly [string, DesktopCommand['type']]
+> = [
+  ['--toggle-visibility', 'toggle-visibility'],
+  ['--toggle-time-tracking', 'toggle-time-tracking'],
+  ['--new-note', 'new-note'],
+  ['--new-task', 'new-task'],
+];
+
+const PROTOCOL_PREFIX = 'superproductivity://';
+
+export const parseDesktopCommandFromArgv = (
+  argv: string[],
+): DesktopCommandParseResult => {
+  const matchingFlags = argv.filter((arg) =>
+    ACTION_FLAG_COMMAND_PAIRS.some(([flag]) => flag === arg),
+  );
+
+  if (matchingFlags.length === 0) {
+    return { kind: 'none' };
+  }
+
+  if (matchingFlags.length > 1) {
+    return {
+      kind: 'error',
+      error: `Multiple desktop command flags are not supported: ${matchingFlags.join(', ')}`,
+    };
+  }
+
+  const matchedFlag = matchingFlags[0];
+  const commandType = ACTION_FLAG_COMMAND_PAIRS.find(
+    ([flag]) => flag === matchedFlag,
+  )?.[1];
+
+  if (!commandType) {
+    return { kind: 'none' };
+  }
+
+  return { kind: 'command', command: { type: commandType } as DesktopCommand };
+};
+
+export const parseDesktopCommandFromProtocolUrl = (
+  url: string,
+): DesktopCommandParseResult => {
+  if (!url.startsWith(PROTOCOL_PREFIX)) {
+    return { kind: 'none' };
+  }
+
+  try {
+    const urlObj = new URL(url);
+    const action = urlObj.hostname;
+    const pathParts = urlObj.pathname.split('/').filter(Boolean);
+
+    switch (action) {
+      case 'create-task': {
+        if (pathParts.length === 0) {
+          return {
+            kind: 'error',
+            error: 'Protocol action "create-task" requires a task title',
+          };
+        }
+        return {
+          kind: 'command',
+          command: { type: 'create-task', title: decodeURIComponent(pathParts[0]) },
+        };
+      }
+      case 'task-toggle-start':
+        return { kind: 'command', command: { type: 'toggle-time-tracking' } };
+      default:
+        return { kind: 'error', error: `Unknown protocol action: ${action}` };
+    }
+  } catch (error) {
+    const errorMessage = error instanceof Error ? error.message : String(error);
+    return {
+      kind: 'error',
+      error: `Failed to parse desktop command URL: ${errorMessage}`,
+    };
+  }
+};
+
+export const getProtocolUrlFromArgv = (argv: string[]): string | undefined =>
+  argv.find((arg) => arg.startsWith(PROTOCOL_PREFIX));
--- a/electron/desktop-command-parser.ts
+++ b/electron/desktop-command-parser.ts
@@ -8,12 +8,27 @@ const ACTION_FLAG_COMMAND_PAIRS: ReadonlyArray<
   ['--new-note', 'new-note'],
   ['--new-task', 'new-task'],
 ];
+const COMMAND_FLAG_PREFIXES = ['--toggle-', '--new-'] as const;
 
 const PROTOCOL_PREFIX = 'superproductivity://';
 
 export const parseDesktopCommandFromArgv = (
   argv: string[],
 ): DesktopCommandParseResult => {
+  const unknownCommandFlag = argv.find(
+    (arg) =>
+      arg.startsWith('--') &&
+      COMMAND_FLAG_PREFIXES.some((prefix) => arg.startsWith(prefix)) &&
+      !ACTION_FLAG_COMMAND_PAIRS.some(([flag]) => flag === arg),
+  );
+
+  if (unknownCommandFlag) {
+    return {
+      kind: 'error',
+      error: `Unknown desktop command flag: ${unknownCommandFlag}`,
+    };
+  }
+
   const matchingFlags = argv.filter((arg) =>
     ACTION_FLAG_COMMAND_PAIRS.some(([flag]) => flag === arg),
   );
--- a/electron/desktop-command.spec.ts
+++ b/electron/desktop-command.spec.ts
@@ -35,6 +35,13 @@ describe('desktop command parser', () => {
     });
   });
 
+  it('should reject unknown desktop command-like flags', () => {
+    expect(parseDesktopCommandFromArgv(['app', '--toggle-'])).toEqual({
+      kind: 'error',
+      error: 'Unknown desktop command flag: --toggle-',
+    });
+  });
+
   it('should parse existing create-task protocol URLs', () => {
     expect(
       parseDesktopCommandFromProtocolUrl('superproductivity://create-task/Foo%20Bar'),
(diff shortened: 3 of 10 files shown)


johannesjo (commented):
## Review of PR #7136 — "Implements & fixes desktop CLI commands/flags"

Overall this is a well-structured PR that unifies desktop command handling behind a clean parser/executor pattern with good type safety and test coverage. The separation into `desktop-command.ts` (types), `desktop-command-parser.ts`, and `desktop-command-executor.ts` is a solid architectural choice. That said, there are several issues worth addressing before merge.

---

### Positives

1. **Clean discriminated union types.** `DesktopCommand` and `DesktopCommandParseResult` use TypeScript discriminated unions (`type`/`kind` fields), which makes the code self-documenting and exhaustive-switch-friendly.

2. **Good testability via dependency injection.** `DesktopCommandWindow` and `DesktopCommandDeps` interfaces decouple the executor from Electron globals, making tests straightforward without mocking entire Electron modules.

3. **Solid test coverage.** The spec file covers parsing (valid flags, unknown flags, multiple flags, protocol URLs) and execution (visibility toggle, IPC dispatch, queuing/flushing). This is a meaningful improvement over the prior state with zero test coverage for this path.

4. **Unified command dispatch.** Global shortcuts, CLI flags, protocol URLs, and second-instance forwarding now all funnel through the same `executeDesktopCommand`, eliminating the duplicated inline logic that existed in `global-shortcuts.ts` and `protocol-handler.ts`.

5. **Improved `showOrFocus` for Linux/Wayland.** Reordering `restore()` before visibility checks and adding `moveTop()` for already-visible windows is a pragmatic fix for Linux WM focus-stealing issues.

---

### Issues to Address

#### 1. Detached spawn is a potential infinite fork-bomb (Severity: High)

In `electron/main.ts`, when `shouldLaunchDetachedInBackground` is true, the process spawns itself with `process.argv.slice(1)` and sets `SP_BACKGROUND_CLI_LAUNCH=1`. However, the spawned child will receive all the original argv flags (including `--new-task`, etc.), and `shouldLaunchDetachedInBackground` only checks `!process.env[BACKGROUND_CLI_ENV_FLAG]`.

This works **as long as environment inheritance is reliable**, but there are edge cases:
- If the spawned Electron process clears or sanitizes its environment (some Electron/snap/flatpak packaging does this), the env flag is lost and the child would spawn another child, creating a fork loop.
- The `SP_BACKGROUND_CLI_LAUNCH` flag is a string `'1'`, but the check is `!process.env[BACKGROUND_CLI_ENV_FLAG]` — if the env var is set to empty string `''`, this would be falsy and re-trigger the spawn.

**Suggestion:** Add a dedicated CLI flag (e.g., `--sp-background`) as a belt-and-suspenders guard alongside the env var, or add a `--no-detach` flag. An argv-based guard survives environment sanitization:

```ts
const isBackgroundLaunch = process.env[BACKGROUND_CLI_ENV_FLAG] === '1' 
  || process.argv.includes('--sp-background');
```

And when spawning, append `'--sp-background'` to the child argv.

#### 2. Duplicated `PROTOCOL_PREFIX` constant (Severity: Low)

`desktop-command-parser.ts` defines its own `const PROTOCOL_PREFIX = 'superproductivity://';` while `protocol-handler.ts` already exports `PROTOCOL_PREFIX`. This duplication means if the protocol name ever changes, two files need updating. Consider importing from `protocol-handler.ts` or extracting the constant to a shared location.

#### 3. `process.argv` is parsed multiple times (Severity: Low)

`process.argv` is parsed in three separate locations:
- `electron/main.ts` (line ~8): `parseDesktopCommandFromArgv(process.argv)` and `process.argv.some(...)` for protocol URL
- `electron/start-app.ts` (~line 114): `parseDesktopCommandFromArgv(process.argv)` again
- `electron/start-app.ts` (~line 71): the existing `process.argv.forEach(...)` loop for `--disable-tray`, `--dev-tools`, etc.

The parsed result from `main.ts` could be passed into `startApp()` as a parameter to avoid re-parsing. This would also make the data flow more explicit and testable.

#### 4. Module-level mutable state in `desktop-command-executor.ts` (Severity: Medium)

`pendingDesktopCommands` is a module-level mutable array. This works in practice because Electron main process is single-threaded, but it creates implicit coupling — any code importing the module shares the same queue. The `resetPendingDesktopCommands()` export exists solely for tests, which is a code smell indicating the state should be encapsulated.

Consider wrapping the queue in a factory or class:

```ts
export function createDesktopCommandQueue() {
  const pending: DesktopCommand[] = [];
  return {
    queue(cmd: DesktopCommand) { pending.push(cmd); },
    flush(deps: FlushDeps) { /* ... */ },
    get length() { return pending.length; },
    reset() { pending.length = 0; },
  };
}
```

This would eliminate the need for `resetPendingDesktopCommands` in production code and make tests naturally isolated.

#### 5. Triple flush call is fragile and unclear (Severity: Medium)

`flushPendingDesktopCommands` is called from three places in `start-app.ts`:
1. Inside `ipcMain.on(IPC.APP_READY, ...)` with a bare `setTimeout()` (line ~64)
2. After `queueOrExecuteDesktopCommand` at startup (line ~114, implicit via queue)
3. Inside `createMainWin()` after the window is created (line ~409)

Plus `processPendingProtocolUrls` (which now delegates to `flushPendingDesktopCommands`) is also called at line ~383.

This "flush everywhere and hope one works" approach is brittle. The `IPC.APP_READY` handler (flush point #1) should be the canonical single flush point since that is when `getIsAppReady()` becomes true. The other flush calls are either redundant (the app isn't ready yet so they no-op) or racing. At minimum, add a comment explaining why multiple flush points exist.

#### 6. Missing exhaustiveness check in `executeDesktopCommand` switch (Severity: Low)

The switch in `executeDesktopCommand` has no `default` case. If a new command type is added to the `DesktopCommand` union but not handled in the executor, it will silently do nothing. Add:

```ts
default: {
  const _exhaustive: never = command;
  log(`Unhandled desktop command: ${JSON.stringify(_exhaustive)}`);
}
```

This gives a compile-time error when a case is missing (TypeScript narrowing to `never`).

#### 7. `as DesktopCommand` type assertion in parser (Severity: Low)

In `desktop-command-parser.ts` line ~56:
```ts
return { kind: 'command', command: { type: commandType } as DesktopCommand };
```

The `as DesktopCommand` assertion bypasses type checking. Since `commandType` is `DesktopCommand['type']` and the command here is always a simple `{ type: ... }` variant (not `create-task` which needs `title`), this happens to be safe. But the assertion would mask a bug if a new flag-based command required additional fields. Consider using a mapping function or explicitly constructing each variant.

#### 8. Security: CLI title injection surface (Severity: Low — already mitigated)

The `create-task` command passes a user-supplied title string through IPC to the renderer. The renderer-side code in `src/app/core/ipc-events.ts` validates that the data is an object with a string `title`, which is good. The title goes through `decodeURIComponent` in the parser, which could throw on malformed percent-encoding (e.g., `%ZZ`), but this is already caught by the `try/catch` in `parseDesktopCommandFromProtocolUrl`. No action needed, just confirming this was reviewed.

#### 9. macOS single-instance lock change (Severity: Medium — behavioral change)

The old code skipped `requestSingleInstanceLock()` on macOS (`if (!IS_MAC)`). The new code applies it unconditionally. The PR comment says "macOS Finder launches are already single-instance, but CLI invocations still need this" — this is correct reasoning, but it's a behavioral change worth calling out. On macOS, if users previously ran multiple instances (e.g., with different `--user-data-dir`), this will now prevent that. Consider whether `--user-data-dir` should bypass the single-instance lock, or document this as a known breaking change.

#### 10. `tsconfig.spec.json` change may pull in unintended files (Severity: Low)

Adding `"../electron/**/*.spec.ts"` to `src/tsconfig.spec.json` means the Angular Karma test runner will try to compile and potentially run Electron-side specs. These specs import from `electron-log/main` and other Node/Electron modules that won't be available in the Karma browser environment. Was this tested with `npm test`? If the specs are meant to run in a different context (e.g., Jest or Mocha for Node), they should have their own tsconfig and test runner rather than being included in the Angular test config.

---

### Summary

The core design is strong — the parser/executor split, discriminated unions, and dependency injection make this significantly more maintainable than the prior scattered inline logic. The main concerns are: (1) the detached spawn needing a more robust re-entry guard, (2) the triple-flush pattern in `start-app.ts` warranting simplification or at least clear documentation, (3) the macOS behavioral change around single-instance locking, and (4) whether the Electron specs actually run correctly under the Angular test harness.

Nice work on the test coverage and the Wayland `showOrFocus` improvements.

---

JamesMowery:

johannesjo (commented):
## Review of PR #7136 — Desktop CLI commands/flags

The core architecture is clean (discriminated union types, dependency-injected executor, shared parser/executor split) and the test coverage is a real improvement. Several issues need to be addressed before merge though.

---

### Critical

**`electron/main.ts` — `spawn()` failure is silent**

If `spawn()` throws (permission denied, path issues on snap/flatpak), the parent still exits 0 and the command is lost with no user feedback. Wrap in try/catch and exit 1 on failure.

**`src/tsconfig.spec.json` — Electron specs in Angular/Karma tsconfig will break `npm test`**

`desktop-command-executor.ts` transitively imports `electron-log/main` (Node-only). Including `../electron/**/*.spec.ts` in `src/tsconfig.spec.json` will cause Karma to fail trying to run these in a browser environment. Revert this change and set up a separate Node/Jest/Mocha runner for the Electron specs.

---

### Important

**`electron/main.ts` — Fork-bomb risk in sandboxed environments**

The re-entry guard relies solely on `process.env[SP_BACKGROUND_CLI_LAUNCH]`. If snap/flatpak sanitizes the environment, the child spawns another child. Add `--sp-background` as an argv-based belt-and-suspenders guard that survives environment sanitization.

**`electron/main.ts` — Removing `IS_MAC` guard is a silent breaking change**

Previously multiple macOS instances were intentionally allowed (see the linked issue #14094). Now `requestSingleInstanceLock()` runs unconditionally. Users relying on multiple instances (e.g., different `--user-data-dir`) lose that silently. If intentional, document it explicitly.

**`electron/start-app.ts` — Background launch with invalid flag still starts the full app**

When `SP_BACKGROUND_CLI_LAUNCH=1` and `parsedDesktopCommand.kind === 'error'`, the error is logged but `startApp()` runs anyway. Should exit 1 before calling `startApp()`.

---

### Minor

- **`electron/desktop-command-parser.ts`** — `PROTOCOL_PREFIX` is duplicated; import from `protocol-handler.ts`
- **`electron/desktop-command-parser.ts:56`** — `as DesktopCommand` assertion would mask a type error if a new command with required fields is added
- **`electron/desktop-command-executor.ts`** — Missing `default: never` exhaustiveness check in the `executeDesktopCommand` switch
- **Protocol URL gap** — The new CLI commands (`--new-task`, `--new-note`, `--toggle-visibility`) have no `superproductivity://` URL equivalents; `superproductivity://new-task` returns an error. Not a regression but worth documenting.

---

I understand the author is currently unavailable to implement changes. The architecture here is solid enough that this is worth picking up — the critical fixes (tsconfig revert + spawn error handling) are small. Happy to help if someone wants to push this over the line.

johannesjo:
## Review of PR #7136 — Desktop CLI commands/flags

The core architecture is clean (discriminated union types, dependency-injected executor, shared parser/executor split) and the test coverage is a real improvement. Several issues need to be addressed before merge though.

---

### Critical

**`electron/main.ts` — `spawn()` failure is silent**

If `spawn()` throws (permission denied, path issues on snap/flatpak), the parent still exits 0 and the command is lost with no user feedback. Wrap in try/catch and exit 1 on failure.

**`src/tsconfig.spec.json` — Electron specs in Angular/Karma tsconfig will break `npm test`**

`desktop-command-executor.ts` transitively imports `electron-log/main` (Node-only). Including `../electron/**/*.spec.ts` in `src/tsconfig.spec.json` will cause Karma to fail trying to run these in a browser environment. Revert this change and set up a separate Node/Jest/Mocha runner for the Electron specs.

---

### Important

**`electron/main.ts` — Fork-bomb risk in sandboxed environments**

The re-entry guard relies solely on `process.env[SP_BACKGROUND_CLI_LAUNCH]`. If snap/flatpak sanitizes the environment, the child spawns another child. Add `--sp-background` as an argv-based belt-and-suspenders guard that survives environment sanitization.

**`electron/main.ts` — Removing `IS_MAC` guard is a silent breaking change**

Previously multiple macOS instances were intentionally allowed (see the linked issue #14094). Now `requestSingleInstanceLock()` runs unconditionally. Users relying on multiple instances (e.g., different `--user-data-dir`) lose that silently. If intentional, document it explicitly.

**`electron/start-app.ts` — Background launch with invalid flag still starts the full app**

When `SP_BACKGROUND_CLI_LAUNCH=1` and `parsedDesktopCommand.kind === 'error'`, the error is logged but `startApp()` runs anyway. Should exit 1 before calling `startApp()`.

---

### Minor

- **`electron/desktop-command-parser.ts`** — `PROTOCOL_PREFIX` is duplicated; import from `protocol-handler.ts`
- **`electron/desktop-command-parser.ts:56`** — `as DesktopCommand` assertion would mask a type error if a new command with required fields is added
- **`electron/desktop-command-executor.ts`** — Missing `default: never` exhaustiveness check in the `executeDesktopCommand` switch
- **Protocol URL gap** — The new CLI commands (`--new-task`, `--new-note`, `--toggle-visibility`) have no `superproductivity://` URL equivalents; `superproductivity://new-task` returns an error. Not a regression but worth documenting.

---

I understand the author is currently unavailable to implement changes. The architecture here is solid enough that this is worth picking up — the critical fixes (tsconfig revert + spawn error handling) are small. Happy to help if someone wants to push this over the line.

johannesjo (changes requested):
Nice work on the parser/executor separation — the `DesktopCommand` tagged union and the `kind: 'none' | 'command' | 'error'` parse result are clean abstractions.

A few things that need to change before this can merge:

**1. Overlap with #7138.** This PR and #7138 share ~90% of their file changes — both add the same desktop-command parser/executor/spec files and the same `main.ts` / `protocol-handler.ts` / `start-app.ts` rewrites. I can't review them independently. Please either collapse into one PR or rebase #7138 on top of this one so the second PR only carries the show/hide delta.

**2. `src/tsconfig.spec.json` include change.** Adding `../electron/**/*.spec.ts` pulls main-process code into the Angular/Karma test harness, which uses a browser module resolution and Angular's tsconfig. This will either silently skip or fail to compile under `npm test`. Electron tests belong with `electron/tsconfig.electron.json` and their own runner. For this PR, I'd suggest dropping the spec file and adding an Electron-side test setup as a follow-up — the parser/executor changes are small enough that we can accept them without unit coverage for now.

**3. macOS single-instance guard removed.** `main.ts` intentionally skips `requestSingleInstanceLock()` on macOS (`if (!IS_MAC)` on master). The new code drops that branch without explanation. Please restore the guard or explain why macOS needs the lock path now.

**4. Respawn-based detach is heavy.** The `spawn(process.execPath, …) + SP_BACKGROUND_CLI_LAUNCH` pattern in `main.ts` costs a full second Electron boot on every CLI invocation and happens before the single-instance lock check, so both parent and child race to acquire it. My preference is to remove the respawn entirely — Wayland keybindings already run binaries detached via `.desktop` Exec, and `superproductivity --new-task &` is the standard terminal pattern. If you want in-code detach, gating it on `isLockObtained && process.stdout.isTTY` would scope it to the "interactive terminal, first launch" case only. Happy to walk through a sketch if useful.

**5. Duplicate flush in `start-app.ts`.** `flushPendingDesktopCommands` is called both in the `IPC.APP_READY` handler and again right after `createWindow` returns. One path is enough — pick whichever is correct and drop the other.

**6. Missing exhaustiveness check in `executeDesktopCommand`.** The switch has no `default` / `never` branch, so if someone adds a new `DesktopCommand` variant they won't get a compile error. A `default: { const _exhaust: never = command; }` is cheap insurance.

Smaller:
- `parseDesktopCommandFromArgv` rejects exact `--toggle-` / `--new-` prefixes as "unknown" — consider suggesting valid flag names in the error, since a user with a typo will hit this.
- The "valid commands stay silent, invalid print error" UX is nice; worth documenting in the PR body.

johannesjo:

johannesjo (commented):
This PR overlaps almost entirely with #7138 — same 10 files, and #7138 is effectively this one plus the Wayland key-repeat debounce and a visibility-based (rather than focus-based) `toggle-visibility`. The two can't merge together cleanly, so to avoid splitting the discussion I've left the detailed review on #7138 and would suggest we consolidate there and close this one.

Short version of that review for context here: the `desktop-command-*` subsystem largely re-implements routing that `protocol-handler.ts` already does via the `second-instance` handler, and the detached background re-spawn in `main.ts` boots Electron twice to deliver one IPC message. For the Wayland use case the leaner path is to add the three missing actions (`toggle-visibility`, `new-note`, `new-task`) to the existing protocol handler and document the keybind recipe. Let's continue on #7138.

johannesjo:
Thanks again for kicking this off, @JamesMowery — the Wayland global-shortcut work and especially the key-repeat hide/re-show bug you found are landing in #8645 (reworked to reuse the existing `superproductivity://` protocol handler per the discussion in #7114). Closing this in favor of that PR; your `globalShowHide` debounce fix is carried over and credited. #7137 stays open separately. 🙏

PR #7137, opened 2026-04-04
fix: new task screen doesn't render/show when new note screen is visible

JamesMowery:
## Problem

If the desktop command to create a new task is triggered while the fullscreen note dialog is open, the task UI can open underneath the note dialog instead of replacing it cleanly.

## Solution

When the add-task desktop command is received while the add-note dialog is open:
- if the note is empty, close the note dialog first and then show the add-task UI
- if the note has unsaved content, ask whether to save before switching
- if the user cancels the confirmation, keep the note dialog open and do not switch

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

Validation notes:

### DIFF
--- a/src/app/core-ui/shortcut/shortcut.service.ts
+++ b/src/app/core-ui/shortcut/shortcut.service.ts
@@ -7,19 +7,21 @@ import { GlobalConfigService } from '../../features/config/global-config.service
 import { ActivatedRoute, Router } from '@angular/router';
 import { LayoutService } from '../layout/layout.service';
 import { TaskService } from '../../features/tasks/task.service';
-import { MatDialog } from '@angular/material/dialog';
+import { MatDialog, MatDialogRef } from '@angular/material/dialog';
 import { DialogAddNoteComponent } from '../../features/note/dialog-add-note/dialog-add-note.component';
 import { IPC } from '../../../../electron/shared-with-frontend/ipc-events.const';
 import { UiHelperService } from '../../features/ui-helper/ui-helper.service';
 import { WorkContextService } from '../../features/work-context/work-context.service';
 import { WorkContextType } from '../../features/work-context/work-context.model';
+import { T } from '../../t.const';
 import { Store } from '@ngrx/store';
 import { showFocusOverlay } from '../../features/focus-mode/store/focus-mode.actions';
 import { SyncWrapperService } from '../../imex/sync/sync-wrapper.service';
 import { first, mapTo, switchMap } from 'rxjs/operators';
-import { fromEvent, merge, Observable, of } from 'rxjs';
+import { firstValueFrom, fromEvent, merge, Observable, of } from 'rxjs';
 import { PluginBridgeService } from '../../plugins/plugin-bridge.service';
 import { TaskShortcutService } from '../../features/tasks/task-shortcut.service';
+import { DialogConfirmComponent } from '../../ui/dialog-confirm/dialog-confirm.component';
 
 // NOTE: Relying on Angular CDK overlay CSS class names keeps shortcut suppression simple.
 // If CDK changes these class names we only need to adjust the helpers below.
@@ -73,7 +75,7 @@ export class ShortcutService {
         this._taskService.toggleStartTask();
       });
       window.ea.on(IPC.SHOW_ADD_TASK_BAR, () => {
-        this._layoutService.showAddTaskBar();
+        void this._showAddTaskBarFromDesktopCommand();
       });
       window.ea.on(IPC.ADD_NOTE, () => {
         if (this._matDialog.openDialogs.length === 0) {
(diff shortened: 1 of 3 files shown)


johannesjo (commented):
## Review of fix: new task screen doesn't render/show when new note screen is visible

Overall this is a solid fix for a real UX issue. The approach of detecting the open note dialog and offering save/discard before switching is well-considered. A few observations:

### Correctness

The save/discard logic is correct:
- "Save" (okTxt) -> `close(true)` from confirm dialog -> `shouldSave = true` -> `addNoteDialog.close(false)` (i.e., `isSkipSave = false`) -> note is saved. Correct.
- "Discard" (cancelTxt) -> `close(false)` from confirm dialog -> `shouldSave = false` -> `addNoteDialog.close(true)` (i.e., `isSkipSave = true`) -> note is discarded. Correct.
- Escape/backdrop dismiss -> `undefined` -> `typeof shouldSave !== 'boolean'` -> early return, note stays open. Correct.

The empty-note path (`noteContent.length === 0`) calls `addNoteDialog.close(true)` which skips save and clears session storage - correct behavior.

### Issues and suggestions

1. **Accessing `componentInstance` is fragile**: `addNoteDialogRef.componentInstance` is typed as the component instance, but the code casts it with `as DialogAddNoteComponent | null`. The `componentInstance` on `MatDialogRef` is never `null` while the dialog is open, so the `| null` and subsequent `?.` optional chaining on `addNoteDialog` is defensive but slightly misleading. More importantly, accessing `componentInstance` directly couples this service to the internal state of the dialog component. Consider whether `DialogAddNoteComponent` could expose its content through the `MatDialogRef` or a shared service instead - though for a targeted bug fix like this, the current approach is pragmatic.

2. **Potential stacking issue with the confirmation dialog**: When the note dialog is fullscreen (`minWidth: '100vw'`, `height: '100vh'`), the confirmation dialog opens on top of it. This should work fine with Material's dialog stacking, but it's worth manually verifying that the confirm dialog is visually accessible and not obscured by the fullscreen note.

3. **The `_getAddNoteDialogRef` helper uses `instanceof`**: This is generally reliable but would break if the dialog was opened from a lazy-loaded module with a different copy of the class. In this codebase `DialogAddNoteComponent` appears to be used in a single place, so this should be fine - just worth noting.

4. **Missing `isEscapeClose` parameter**: `DialogAddNoteComponent.close()` accepts an optional second parameter `isEscapeClose`. The PR calls `addNoteDialog?.close(!shouldSave)` and `addNoteDialog?.close(true)` without it, which defaults to `false`. This is correct behavior (we want the save/discard logic to run, not the escape-close path that preserves session storage).

5. **Translation keys added correctly**: New keys `F.NOTE.C.CONFIRM_SAVE_BEFORE_OPENING_NEW_TASK`, `G.SAVE`, and `G.DISCARD` are added to both `t.const.ts` and `en.json`, following the project convention of only editing `en.json`.

6. **No unit or e2e tests**: The PR checklist notes tests are not included. Given this is IPC-triggered behavior (`window.ea.on(IPC.SHOW_ADD_TASK_BAR, ...)`), testing is admittedly harder, but a unit test for `_showAddTaskBarFromDesktopCommand` (e.g., mocking `_matDialog.openDialogs`) would increase confidence, especially around the async flow.

7. **Minor: `void` prefix on async call**: `void this._showAddTaskBarFromDesktopCommand()` correctly marks the floating promise as intentionally unhandled. Good practice.

### Summary

The fix is well-structured and handles the three cases (empty note, note with content + save, note with content + discard) correctly. The UX of keeping the note open when the user dismisses the confirmation (via Escape/backdrop) is a nice touch. The main suggestion is to add a test for the new async method. Otherwise this looks ready to merge.

---

JamesMowery:

johannesjo (changes requested):




The dialog handoff logic is correct — I traced it against `DialogAddNoteComponent.close(isSkipSave)` and `DialogConfirmComponent`: Save → `close(false)` saves; Discard → `close(true)` skips; Escape/backdrop (`undefined`) → short-circuits. Nice work.

One small nit before merge: the new translation key lives under `F.NOTE.C.*`, but the existing `NOTE` group uses `D_FULLSCREEN` and `S` as sub-sections — there's no `C` pattern elsewhere in `t.const.ts`. Could you fold the key under `F.NOTE.D_FULLSCREEN.*` or just use a flat `F.NOTE.CONFIRM_SAVE_BEFORE_OPENING_NEW_TASK` to stay consistent?

After that, happy to merge.

johannesjo:
Thanks — this is a real layering bug (the add-task bar renders inline in the page while the note is a full-screen CDK overlay, so it appears underneath), and the close/save/discard handoff is the right fix.

Two things still block merging:

1. **Translation key grouping.** The new key sits under a `F.NOTE.C.*` subsection that doesn't match the NOTE group convention (existing subsections are `D_FULLSCREEN` / `S`, there is no `C`). Please fold it into `F.NOTE.D_FULLSCREEN.*` or use a flat `F.NOTE.CONFIRM_SAVE_BEFORE_OPENING_NEW_TASK` in `t.const.ts` + `en.json`.

2. **Rebase.** The branch has since diverged and now conflicts with master, so it needs a rebase.

A small spec for the three branches of `_showAddTaskBarFromDesktopCommand` (empty note / save / discard) would be a nice addition — `shortcut.service.spec.ts` already mocks `MatDialog`.

I know you mentioned stepping away from this one — the logic itself is sound and close to mergeable, so if anyone wants to pick up the rebase + key rename it should be quick.

src/app/core-ui/shortcut/shortcut.service.ts
```diff
@@ -88,6 +90,51 @@ export class ShortcutService {
...
+  private async _showAddTaskBarFromDesktopCommand(): Promise<void> {
+    const addNoteDialogRef = this._getAddNoteDialogRef();
+    if (!addNoteDialogRef) {
+      this._layoutService.showAddTaskBar();
+      return;
+    }
+
+    const addNoteDialog =
+      addNoteDialogRef.componentInstance as DialogAddNoteComponent | null;
+    const noteContent = addNoteDialog?.data?.content?.trim() || '';
```
johannesjo, line 102: Verified this edge case holds: the note textarea is `[(ngModel)]="data.content"` in the fullscreen base dialog, so `data.content` reflects live unsaved edits — the "Save before opening a new task?" prompt fires for freshly-typed notes, not only previously-saved ones. And when `shouldSave` is `undefined` (user dismisses the confirm via backdrop/Esc) you correctly abort and leave the note open. No change needed — just confirming the behavior is right.

johannesjo (commented):
This one is genuinely independent of the CLI work and the cleanest of the three — scoped, uses the standard confirm dialog, and only adds the new strings to `en.json`. Looks good to me modulo the inline note. Thanks!

Zoriot:
@johannesjo can you verify the state of these changes? Is it still worth picking that up? If yes i can do it

PR #7138, opened 2026-04-04
Fix/global show hide shortcut

JamesMowery:
## Problem

On Linux/Wayland, invoking the global "Show/hide Super Productivity" shortcut while the SP window was visible/in focus would hide the main window but then automatically reopen it the window ~0.5 seconds after (i.e., window was not remaining hidden, as intended).

Root cause: Electron's `globalShortcut` repeated while the key combo was still held, which caused an unintended second toggle after the window had already been hidden.

## Solution

This change stabilizes the global show/hide shortcut by:

- using window visible/minimized state instead of focus state to decide whether to hide or show
- ignoring repeated hidden-window shortcut firings within a short window so one physical key press only hides once

Observed and validated on Linux/Wayland (`niri`).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

## Validation Notes

- `npm run electron:build` passed locally
- `.deb` packaging also passed locally during validation

### DIFF
--- a/electron/ipc-handlers/global-shortcuts.ts
+++ b/electron/ipc-handlers/global-shortcuts.ts
@@ -6,6 +6,9 @@ import { errorHandlerWithFrontendInform } from '../error-handler-with-frontend-i
 import { executeDesktopCommand } from '../desktop-command-executor';
 import { showOrFocus } from '../various-shared';
 
+const TOGGLE_VISIBILITY_REPEAT_GAP_MS = 1000;
+let lastToggleVisibilityEvent = 0;
+
 export const initGlobalShortcutsIpc = (): void => {
   ipcMain.on(IPC.REGISTER_GLOBAL_SHORTCUTS_EVENT, (ev, cfg) => {
     registerShowAppShortCuts(cfg);
@@ -33,6 +36,21 @@ const registerShowAppShortCuts = (cfg: KeyboardConfig): void => {
         switch (key) {
           case 'globalShowHide':
             actionFn = () => {
+              const now = Date.now();
+              const sinceLastMs = now - lastToggleVisibilityEvent;
+              lastToggleVisibilityEvent = now;
+
+              // On some Linux/Wayland setups Electron's globalShortcut repeats while the
+              // key combo is held. Treat hidden-window repeats with a short required quiet
+              // gap as the same physical press, and only allow re-show after a real pause.
+              if (
+                !mainWin.isVisible() &&
+                !mainWin.isMinimized() &&
+                sinceLastMs < TOGGLE_VISIBILITY_REPEAT_GAP_MS
+              ) {
+                return;
+              }
+
               executeDesktopCommand({ type: 'toggle-visibility' }, mainWin, {
                 showOrFocus,
               });
--- a/electron/ipc-handlers/global-shortcuts.ts
+++ b/electron/ipc-handlers/global-shortcuts.ts
@@ -2,8 +2,9 @@ import { globalShortcut, ipcMain } from 'electron';
 import { IPC } from '../shared-with-frontend/ipc-events.const';
 import { KeyboardConfig } from '../../src/app/features/config/keyboard-config.model';
 import { getWin } from '../main-window';
-import { showOrFocus } from '../various-shared';
 import { errorHandlerWithFrontendInform } from '../error-handler-with-frontend-inform';
+import { executeDesktopCommand } from '../desktop-command-executor';
+import { showOrFocus } from '../various-shared';
 
 export const initGlobalShortcutsIpc = (): void => {
   ipcMain.on(IPC.REGISTER_GLOBAL_SHORTCUTS_EVENT, (ev, cfg) => {
@@ -32,34 +33,29 @@ const registerShowAppShortCuts = (cfg: KeyboardConfig): void => {
         switch (key) {
           case 'globalShowHide':
             actionFn = () => {
-              if (mainWin.isFocused()) {
-                // we need to blur the window for windows
-                mainWin.blur();
-                mainWin.hide();
-              } else {
-                showOrFocus(mainWin);
-              }
+              executeDesktopCommand({ type: 'toggle-visibility' }, mainWin, {
+                showOrFocus,
+              });
             };
             break;
 
           case 'globalToggleTaskStart':
             actionFn = () => {
-              mainWin.webContents.send(IPC.TASK_TOGGLE_START);
+              executeDesktopCommand({ type: 'toggle-time-tracking' }, mainWin, {
+                showOrFocus,
+              });
             };
             break;
 
           case 'globalAddNote':
             actionFn = () => {
-              showOrFocus(mainWin);
-              mainWin.webContents.send(IPC.ADD_NOTE);
+              executeDesktopCommand({ type: 'new-note' }, mainWin, { showOrFocus });
             };
             break;
 
           case 'globalAddTask':
             actionFn = () => {
-              showOrFocus(mainWin);
-              // NOTE: delay slightly to make sure app is ready
-              mainWin.webContents.send(IPC.SHOW_ADD_TASK_BAR);
+              executeDesktopCommand({ type: 'new-task' }, mainWin, { showOrFocus });
             };
             break;
 
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -1,10 +1,28 @@
+import { spawn } from 'child_process';
 import { app } from 'electron';
 import { parseDesktopCommandFromArgv } from './desktop-command-parser';
 import { PROTOCOL_PREFIX } from './protocol-handler';
 import { startApp } from './start-app';
 
+const BACKGROUND_CLI_ENV_FLAG = 'SP_BACKGROUND_CLI_LAUNCH';
 const parsedDesktopCommand = parseDesktopCommandFromArgv(process.argv);
 const hasProtocolUrl = process.argv.some((arg) => arg.startsWith(PROTOCOL_PREFIX));
+const shouldLaunchDetachedInBackground =
+  !process.env[BACKGROUND_CLI_ENV_FLAG] &&
+  (parsedDesktopCommand.kind === 'command' || hasProtocolUrl);
+
+if (shouldLaunchDetachedInBackground) {
+  const child = spawn(process.execPath, process.argv.slice(1), {
+    detached: true,
+    stdio: 'ignore',
+    env: {
+      ...process.env,
+      [BACKGROUND_CLI_ENV_FLAG]: '1',
+    },
+  });
+  child.unref();
+  process.exit(0);
+}
 
 // Enforce single-instance behavior on all desktop platforms.
 // macOS Finder launches are already single-instance, but CLI invocations still need this.
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -1,17 +1,24 @@
 import { app } from 'electron';
+import { parseDesktopCommandFromArgv } from './desktop-command-parser';
 import { PROTOCOL_PREFIX } from './protocol-handler';
 import { startApp } from './start-app';
+
+const parsedDesktopCommand = parseDesktopCommandFromArgv(process.argv);
+const hasProtocolUrl = process.argv.some((arg) => arg.startsWith(PROTOCOL_PREFIX));
 
 // Enforce single-instance behavior on all desktop platforms.
 // macOS Finder launches are already single-instance, but CLI invocations still need this.
 const isLockObtained = app.requestSingleInstanceLock();
 if (!isLockObtained) {
-  const hasProtocolUrl = process.argv.some((arg) => arg.startsWith(PROTOCOL_PREFIX));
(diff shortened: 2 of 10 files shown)


johannesjo (commented):
## Review of Fix/global show hide shortcut

Good job identifying and fixing the Linux/Wayland repeat-fire bug. The core fix (using visibility/minimized state instead of focus state, plus debouncing) is sound. The refactoring into a `DesktopCommand` abstraction is well-structured and the tests are thorough. That said, this PR has grown significantly beyond the original bug fix, and several parts introduce concerns worth discussing.

### The core fix (global-shortcuts.ts, various-shared.ts) -- Looks good

- Switching from `isFocused()` to `isVisible() && !isMinimized()` is the right call. Focus state is unreliable on Wayland compositors, while visibility/minimized state is definitive.
- The 1-second debounce (`TOGGLE_VISIBILITY_REPEAT_GAP_MS`) with the guard that only suppresses re-show (not re-hide) is a sensible approach.
- The `showOrFocus` change to call `win.restore()` before the visibility check, and using `win.moveTop()` instead of just `win.focus()` for already-visible windows, are good improvements for Linux WM compatibility.

### Concerns

**1. The `spawn` + detached re-launch in `main.ts` is risky and seems out of scope**

The biggest concern is the detached background spawn logic:

```ts
if (shouldLaunchDetachedInBackground) {
  const child = spawn(process.execPath, process.argv.slice(1), {
    detached: true,
    stdio: 'ignore',
    env: { ...process.env, [BACKGROUND_CLI_ENV_FLAG]: '1' },
  });
  child.unref();
  process.exit(0);
}
```

This re-launches the Electron process as a detached child whenever a CLI command flag or protocol URL is detected. The stated goal seems to be returning control to the terminal immediately, but this has significant implications:
- It bypasses Electron's single-instance lock mechanism -- the parent exits before `requestSingleInstanceLock()` is called, so the detached child will try to acquire the lock independently. If a running instance already holds the lock, the child becomes a second instance that triggers the `second-instance` event, which is fine. But if no instance is running, you now have a process launched with `detached: true` and `stdio: 'ignore'`, meaning all stdout/stderr is lost and the process is invisible to the launching terminal.
- On macOS (and potentially some Linux DEs), spawning a second Electron process can cause dock icon duplication or other visual artifacts.
- The `process.argv.slice(1)` forwarding could interact unexpectedly with Electron's own flags (e.g., `--inspect`, `--remote-debugging-port`).

This functionality was not present before and is not needed to fix the original show/hide bug. I'd recommend removing the detached spawn logic from this PR and submitting it as a separate, focused PR if desired.

**2. macOS single-instance lock change**

The original code intentionally skipped `requestSingleInstanceLock()` on macOS (`if (!IS_MAC)`), with a comment referencing [electron#14094](https://github.com/electron/electron/issues/14094). The PR removes this platform guard and now calls `requestSingleInstanceLock()` unconditionally. The comment says "macOS Finder launches are already single-instance, but CLI invocations still need this" -- but this could reintroduce the Electron bug that the original guard was avoiding. Was this tested on macOS?

**3. `pendingDesktopCommands` is module-level mutable state shared across call sites**

The `pendingDesktopCommands` array in `desktop-command-executor.ts` is a module-level singleton. Both `queueOrExecuteDesktopCommand` and `flushPendingDesktopCommands` operate on it, and `flushPendingDesktopCommands` is called from at least three different places in `start-app.ts` (IPC.APP_READY handler, protocol pending URL flush, and the window ready-to-show area). If these fire in unexpected orders or concurrently, commands could be flushed multiple times or lost. This is probably fine in practice since Electron's main process is single-threaded, but the multiple flush sites make reasoning about it harder. Consider consolidating to a single flush point.

**4. The `globalShowHide` debounce only suppresses re-show, not re-hide**

The debounce logic in `global-shortcuts.ts`:
```ts
if (!mainWin.isVisible() && !mainWin.isMinimized() && sinceLastMs < TOGGLE_VISIBILITY_REPEAT_GAP_MS) {
  return;
}
```
This blocks rapid re-shows after a hide, which fixes the reported bug. But it does not suppress rapid re-hides. If a user presses the shortcut twice quickly while the window is visible, the first press hides and the second press (within the gap) is suppressed (good). However, if the window is visible and minimized (`isVisible() && isMinimized()`), this condition evaluates to false and the command proceeds. This is probably fine since `executeDesktopCommand` would call `showOrFocus` in that case, but worth noting that the guard's conditions are slightly asymmetric.

**5. Missing `blur()` call on Windows**

The old `globalShowHide` handler had:
```ts
if (mainWin.isFocused()) {
  mainWin.blur();
  mainWin.hide();
}
```
The `blur()` before `hide()` was specifically noted as needed for Windows. The new `executeDesktopCommand` just calls `mainWin.hide()` without `blur()`. This could cause the window to remain in the taskbar or not properly release focus on Windows. Unless this was tested on Windows and confirmed unnecessary, I'd suggest preserving the `blur()` call.

### Minor notes

- The `DesktopCommandWindow` interface in `desktop-command-executor.ts` includes `isFocused()` but it is never called anywhere in the executor. Consider removing it from the interface.
- The `PROTOCOL_PREFIX` is redefined in `desktop-command-parser.ts` as a local constant, while it is also exported from `protocol-handler.ts`. This duplication could lead to drift. Consider importing it instead.
- The test file `desktop-command.spec.ts` is well-written with good coverage of both parser and executor paths.
- Including `../electron/**/*.spec.ts` in `src/tsconfig.spec.json` is a nice touch for enabling the Karma runner on Electron-side tests.

### Summary

The core bug fix (visibility-based toggle + debounce) is correct and well-tested. However, the PR bundles several additional changes (detached spawn, macOS single-instance lock, DesktopCommand abstraction, second-instance handler rewrite) that increase the risk surface considerably. I'd recommend splitting this into:
1. The bug fix: visibility-based toggle, debounce in global-shortcuts, and the `showOrFocus` improvements.
2. The DesktopCommand refactoring (parser, executor, protocol handler rewrite) as a separate PR.
3. The `main.ts` spawn/single-instance changes as a third PR, after verifying cross-platform behavior.

At minimum, the missing `blur()` on Windows for the hide path should be addressed before merging.

---

JamesMowery:

johannesjo (changes requested):



Before the line-level feedback, two blockers:

**1. This PR would regress master's 2cefde61b (refs #7282).** I landed a platform-aware `globalShowHide` fix on master: on Windows/Linux without a tray icon, `hide()` leaves the window unreachable, so `minimize()` is used instead. This PR branched before that commit and its `executeDesktopCommand('toggle-visibility')` is a bare `mainWin.hide()` — merging would undo the fix.

Please rebase on current master and move the platform-aware logic (`IS_MAC ? hide() : (getIsMinimizeToTray() && ensureIndicator()) ? hide() : minimize()`) into the executor so it survives the refactor.

**2. ~90% overlap with #7136.** This PR is effectively #7136 plus the repeat-gap debounce and the `isVisible`/`isMinimized` tweak in the executor. I can't review it independently of #7136. Please collapse the two PRs or rebase this one on top of #7136 so it only carries the show/hide-specific delta.

Everything from the #7136 review (tsconfig.spec.json, macOS lock guard, respawn heaviness, duplicate flush, switch exhaustiveness) applies here too.

**3. The 1s `TOGGLE_VISIBILITY_REPEAT_GAP_MS` is a workaround, not a fix.** A hard-coded 1000ms will swallow legitimate double-presses. Worth investigating whether Electron `globalShortcut` auto-repeat can be suppressed at registration time on Wayland before shipping a time gate. If a debounce is truly the only option, please at least mark it as a workaround with a link to the upstream issue.

Once the rebase is done and the platform logic is preserved, this should be a small, focused diff — the repeat-gap plus the executor visibility logic.

electron/main.ts
```diff
@@ -1,24 +1,42 @@
...
+    detached: true,
+    stdio: 'ignore',
+    env: {
+      ...process.env,
+      [BACKGROUND_CLI_ENV_FLAG]: '1',
+    },
+  });
+  child.unref();
+  process.exit(0);
+}
```
johannesjo, line 25: This boots Electron **twice** per CLI invocation: the foreground process immediately re-spawns a second detached Electron process (`SP_BACKGROUND_CLI_LAUNCH`), which then fails the single-instance lock and forwards via `second-instance` before exiting. That's two full runtime starts to deliver one IPC message — exactly the "spinning up an entire Electron instance is heavy" concern raised in the issue thread. The existing `superproductivity://` second-instance path already forwards with one fewer spawn; I'd rather lean on that than add the detached double-spawn.

electron/main.ts
```diff
@@ -1,24 +1,42 @@
...
+
+// Enforce single-instance behavior on all desktop platforms.
+// macOS Finder launches are already single-instance, but CLI invocations still need this.
+const isLockObtained = app.requestSingleInstanceLock();
```
johannesjo, line 29: Behavior change worth calling out explicitly: master only takes the single-instance lock on non-macOS (`if (!IS_MAC)`); this now locks on **all** platforms. That changes macOS launch semantics (which relied on Finder + `open-url`) and re-touches the dev-mode exception from electron#14094. Needs an on-device macOS check before merge — this is the kind of cross-platform change that only surfaces on one OS.

src/tsconfig.spec.json
```diff
@@ -10,5 +10,5 @@
     }
   },
   "files": ["test.ts", "polyfills.ts"],
-  "include": ["**/*.spec.ts", "**/*.d.ts"]
+  "include": ["**/*.spec.ts", "**/*.d.ts", "../electron/**/*.spec.ts"]
```
johannesjo, line 13: This pulls `electron/**/*.spec.ts` into the Angular/Karma (browser) test run. The repo's convention is electron tests as `electron/*.test.cjs` run via `node --test` (`npm run test:electron`) — there are no `electron/*.spec.ts` on master. `desktop-command-executor.ts` also imports `electron-log/main` (node-only), which would now be bundled into the browser test context. Please move these specs to the existing `.test.cjs` / `node --test` harness instead of grafting electron code into Karma.

electron/ipc-handlers/global-shortcuts.ts
```diff
@@ -32,34 +36,44 @@ const registerShowAppShortCuts = (cfg: KeyboardConfig): void => {
...
+              // On some Linux/Wayland setups Electron's globalShortcut repeats while the
+              // key combo is held. Treat hidden-window repeats with a short required quiet
+              // gap as the same physical press, and only allow re-show after a real pause.
+              if (
+                !mainWin.isVisible() &&
+                !mainWin.isMinimized() &&
+                sinceLastMs < TOGGLE_VISIBILITY_REPEAT_GAP_MS
+              ) {
+                return;
               }
```
johannesjo, line 52: This is the actual bug fix from the issue and it's worth keeping regardless of the overall direction. Two small notes: (1) the debounce relies on module-level mutable state (`lastToggleVisibilityEvent`) that persists across `registerShowAppShortCuts` re-registrations — fine, but worth a comment so it isn't mistaken for dead state later; (2) the 1000ms gap is a heuristic — a compositor that legitimately repeats slower than that would still double-fire. Acceptable, just document the assumption.

johannesjo (commented):
Thanks for this, and welcome — solid first contribution. The underlying need is legitimate: Wayland deliberately leaves global hotkeys to the compositor, so a CLI/URL entry point is the right platform-native escape hatch, and it stays fully offline/local, which fits the project's direction.

A few structural things before this can land, though:

**This overlaps almost entirely with #7136.** Both PRs touch the same 10 files; #7138 is effectively #7136 plus the key-repeat debounce and a visibility-based (rather than focus-based) toggle. They can't both merge cleanly, so let's consolidate on one and close the other. (The framing in the issue of "separated into independent branches" isn't quite accurate — only the note/task handoff in #7137 is genuinely independent.)

**The approach is heavier than it needs to be.** master already routes `superproductivity://…` URLs into the running instance via the `second-instance` handler in `protocol-handler.ts` (it handles `create-task`, `task-toggle-start`, `oauth-callback` today). For the Wayland use case the only real gaps are (a) three missing actions — `toggle-visibility`, `new-note`, `new-task` — and (b) the hide/re-show key-repeat bug fixed here. Adding those three actions to the existing protocol handler + documenting the Niri/Wayland keybind recipe gets Phase 1 done with a fraction of the surface, and avoids maintaining a second `--flag` argv parser in lockstep with the URL parser forever. The new `desktop-command-*` subsystem largely re-implements routing the protocol handler already does. This also lines up with @emanguy's suggestion in the issue thread.

The debounce fix in `global-shortcuts.ts` is the genuinely valuable bug fix here and worth keeping regardless of which direction we take. Specific notes inline.

johannesjo:
Thanks again for kicking this off, @JamesMowery — the Wayland global-shortcut work and especially the key-repeat hide/re-show bug you found are landing in #8645 (reworked to reuse the existing `superproductivity://` protocol handler per the discussion in #7114). Closing this in favor of that PR; your `globalShowHide` debounce fix is carried over and credited. #7137 stays open separately. 🙏

PR #7142, opened 2026-04-05
fix(nav-item): prevent project counters from overlapping menu icon

### LINKED ISSUE #7132: 🚨 UI - Project-Counters overlap the Menu-Icon
### Version Used

18.1.3

### Expected Behavior

The project counters and the menu icons are visible side by side

### Current Behavior

The project counters and the menu icons are overlapping.
Therefore, the counters are not legible anymore.

Image

### Steps to Reproduce

Just open Super Productivity....
It does also not help, when I enlarge the sidebar.

Dark Mode: System
Theme: Default

### Console Output

```shell

```

### documentation gap

cyphercodes:
## Description

This PR fixes the UI issue where project counters (task-count badges) overlap with the menu icon (three-dot menu button) in the sidebar navigation.

### Problem

When the sidebar was narrow, the task-count badge positioned at `right: 16px` would overlap with the additional-btn (three-dot menu button, 40px wide), making both elements difficult to read and interact with.

### Solution

Changed the default position of `.task-count` from `right: 16px` to `right: 48px` (40px button width + 8px padding) to ensure the counter is always positioned to the left of the menu button.

### Changes

- Updated `src/app/core-ui/magic-side-nav/nav-item/nav-item.component.scss`
- Changed `right: 16px` to `right: 48px` in the `.task-count` class
- Removed redundant `right: 48px !important` from the `@media (hover: none)` query since it's now the default

Fixes #7132

### DIFF
--- a/src/app/core-ui/magic-side-nav/nav-item/nav-item.component.scss
+++ b/src/app/core-ui/magic-side-nav/nav-item/nav-item.component.scss
@@ -75,16 +75,6 @@
   }
 }
 
-@media (hover: none) {
-  .additional-btn {
-    opacity: 1;
-    visibility: visible;
-  }
-  .task-count {
-    display: flex !important;
-  }
-}
-
 .task-count {
   position: absolute;
   top: 50%;
@@ -92,13 +82,28 @@
   line-height: 1;
   text-align: center;
   font-size: 10px;
-  // Position to the left of the additional button (40px) with some padding (8px)
-  right: 48px;
+  right: 16px;
   display: flex;
   align-items: center;
   color: var(--text-color-muted);
   // avoid affecting drag handle
   pointer-events: none;
+}
+
+// On touch devices the .additional-btn is always visible, so let the counter
+// flow inline beside it instead of overlapping.
+:host-context(.isTouchPrimary) {
+  .task-count {
+    position: static;
+    transform: none;
+    padding: 0 var(--s);
+    flex-shrink: 0;
+    pointer-events: auto;
+  }
+
+  button:first-of-type.has-tasks {
+    padding-right: var(--s);
+  }
 }
 
 // Active state visuals - ensure they override Angular Material styles
--- a/src/app/core-ui/magic-side-nav/nav-item/nav-item.component.scss
+++ b/src/app/core-ui/magic-side-nav/nav-item/nav-item.component.scss
@@ -81,7 +81,6 @@
     visibility: visible;
   }
   .task-count {
-    right: 48px !important;
     display: flex !important;
   }
 }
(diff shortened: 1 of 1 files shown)


johannesjo (commented):
Thanks for tackling this overlap issue. I traced through the full styling chain and I believe this change introduces a visual regression on desktop, while the original code already handled the touch-device case correctly. Here is the detailed analysis:

### How the layout works today (on `master`)

There are **two different device contexts** to consider:

**Desktop (mouse-primary devices):**
The global stylesheet `src/styles/components/multi-btn-wrapper.scss` hides `.additional-btn` by default on mouse-primary devices via:
```scss
> .additional-btn {
  @include is-touch-device.mousePrimaryDevice(true) {
    opacity: 0;
    display: none;  // takes no space in flex layout
  }
}
```
When the user hovers the row, the button reappears (`display: inline-flex; opacity: 0.8`), but at that point the `@media (hover: hover)` rule in `nav-item.component.scss` kicks in and **hides** `.task-count` entirely (`display: none !important`). So the counter and the button are **never visible simultaneously** on desktop. The counter at `right: 16px` is correct here because it sits near the right edge when the button is hidden.

**Touch/mobile devices (`hover: none`):**
Both `.additional-btn` and `.task-count` are always visible simultaneously. The existing `@media (hover: none)` block already handles this correctly with `right: 48px !important`, pushing the counter to the left of the always-visible button.

### What this PR changes

By making `right: 48px` the default and removing the `@media (hover: none)` override, the counter will now sit 48px from the right edge **on desktop too** -- even when the `.additional-btn` is `display: none` and occupies zero space. This creates a visible gap between the counter and the right edge of the sidebar on desktop in the non-hover state.

### The original issue (#7132)

The screenshot in #7132 shows a mobile/touch device. The existing `@media (hover: none) { .task-count { right: 48px !important; } }` rule should already fix this. If the reporter is still seeing the overlap, there may be a specificity or load-order issue worth investigating rather than changing the default positioning.

### Suggestion

I would recommend keeping `right: 16px` as the default (for desktop non-hover state) and keeping the `right: 48px !important` in the `@media (hover: none)` block. If there is a scenario where the overlap occurs on desktop (e.g., the global `mousePrimaryDevice` mixin is not applied), it would be better to investigate that root cause rather than shifting the default position for all devices.

If you do want a single unified value, an alternative approach would be to use CSS that is aware of the button's visibility, such as:
```scss
right: 16px;

:host:has(.additional-btn:not([style*="display: none"])) & {
  right: 48px;
}
```
Or anchor the counter relative to the button's position using flexbox ordering instead of absolute positioning.

---

johannesjo:
Thank you very much!

Issue #10, opened 2017-08-19
[Linux] desktop file is lacking a Categories= entry


probonopd:
The desktop file inside http://super-productivity.com/downloads/superProductivity-latest-x86_64.AppImage is lacking a `Categories=` entry in the `*.desktop` file.

According to the [menu spec](https://standards.freedesktop.org/menu-spec/latest/apa.html),

> By including one of the Main Categories in an application's desktop entry file, the application will be ensured that it will show up in a section of the application menu dedicated to this category. If multiple Main Categories are included in a single desktop entry file, the entry may appear more than once in the menu. 

Hence, please add at least one of the following in the `Categories=` key.

Main Category | Description | Notes
-- | -- | --
AudioVideo | Application for presenting, creating, or processing multimedia (audio/video) |  
Audio | An audio application | Desktop entry must include AudioVideo as well
Video | A video application | Desktop entry must include AudioVideo as well
Development | An application for development |  
Education | Educational software |  
Game | A game |  
Graphics | Application for viewing, creating, or processing graphics |  
Network | Network application such as a web browser |  
Office | An office type application |  
Science | Scientific software |  
Settings | Settings applications | Entries may appear in a separate menu or as part of a 		"Control Center"
System | System application, "System Tools" such as say a log viewer or network monitor |  
Utility | Small utility application, "Accessories"

In addition, you could specify one or more from the longer list of [Additional Categories](https://standards.freedesktop.org/menu-spec/latest/apas02.html).

Please test the result with `desktop-file-validate` and make sure it passes.

electron-builder offers native support for this since [v19.22.1](https://github.com/electron-userland/electron-builder/releases/tag/v19.22.1).

Reference: https://github.com/AppImage/AppImageHub/issues/2

johannesjo:
Thank you very much for reporting! Would you be interested in providing a pull request for the issue?

probonopd:
Sorry, I wouldn't even know where to start because I cannot find the desktop file to be edited in this repository. Probably it is generated by electron-builder, so updating to the latest electron-builder should do wonders.

johannesjo:
No worries! Thanks for the detailed report! I'll look into it. Out of curiosity: are you using the tool on a regular basis? Would love to hear your thoughts about it!

probonopd:
Actually I am building a directory of AppImages, https://github.com/AppImage/AppImageHub/ (hey, free advertising!) and don't really use the tool otherwise.

johannesjo:
[...]

PR #7153, opened 2026-04-06
feat(plugins): add dashboard as standalone iFrame plugin

johannesjo:
Create a new dashboard plugin at packages/plugin-dev/dashboard/ with
four widgets: Today's Progress, Current Task (with live timer),
Daily Summary/Streaks (with persisted data), and Quick Actions.

Registered as a bundled plugin in BUNDLED_PLUGIN_PATHS.

https://claude.ai/code/session_016cipBZavYJY4MZ24sRVuvo

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -645,6 +645,256 @@
         "vite": "^6.0.0 || ^7.0.0"
       }
     },
+    "node_modules/@angular-devkit/build-angular/node_modules/@vitest/expect": {
+      "version": "4.1.2",
+      "resolved": "https://registry.npmjs.org/@vitest/expect/-/expect-4.1.2.tgz",
+      "integrity": "sha512-gbu+7B0YgUJ2nkdsRJrFFW6X7NTP44WlhiclHniUhxADQJH5Szt9mZ9hWnJPJ8YwOK5zUOSSlSvyzRf0u1DSBQ==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "peer": true,
+      "dependencies": {
+        "@standard-schema/spec": "^1.1.0",
+        "@types/chai": "^5.2.2",
+        "@vitest/spy": "4.1.2",
+        "@vitest/utils": "4.1.2",
+        "chai": "^6.2.2",
+        "tinyrainbow": "^3.1.0"
+      },
+      "funding": {
+        "url": "https://opencollective.com/vitest"
+      }
+    },
+    "node_modules/@angular-devkit/build-angular/node_modules/@vitest/pretty-format": {
+      "version": "4.1.2",
+      "resolved": "https://registry.npmjs.org/@vitest/pretty-format/-/pretty-format-4.1.2.tgz",
+      "integrity": "sha512-dwQga8aejqeuB+TvXCMzSQemvV9hNEtDDpgUKDzOmNQayl2OG241PSWeJwKRH3CiC+sESrmoFd49rfnq7T4RnA==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "peer": true,
+      "dependencies": {
+        "tinyrainbow": "^3.1.0"
+      },
+      "funding": {
+        "url": "https://opencollective.com/vitest"
+      }
+    },
+    "node_modules/@angular-devkit/build-angular/node_modules/@vitest/runner": {
+      "version": "4.1.2",
+      "resolved": "https://registry.npmjs.org/@vitest/runner/-/runner-4.1.2.tgz",
+      "integrity": "sha512-Gr+FQan34CdiYAwpGJmQG8PgkyFVmARK8/xSijia3eTFgVfpcpztWLuP6FttGNfPLJhaZVP/euvujeNYar36OQ==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "peer": true,
+      "dependencies": {
+        "@vitest/utils": "4.1.2",
+        "pathe": "^2.0.3"
+      },
+      "funding": {
+        "url": "https://opencollective.com/vitest"
+      }
+    },
+    "node_modules/@angular-devkit/build-angular/node_modules/@vitest/snapshot": {
+      "version": "4.1.2",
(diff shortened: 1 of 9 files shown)


johannesjo (commented):
## Review: feat(plugins): add dashboard as standalone iFrame plugin

### Summary

This PR adds a new bundled dashboard plugin with four widgets (Today's Progress, Current Task with live timer, Streaks with persisted data, and Quick Actions). It registers the plugin in `BUNDLED_PLUGIN_PATHS` and provides an example of a full-featured iFrame plugin using the existing plugin infrastructure.

### Positives

1. **Clean HTML structure**: The `index.html` follows the project guidelines well -- plain HTML with semantic elements, minimal CSS classes, and good use of CSS variables from the host theme (`--s3`, `--s2`, `--c-primary`, `--text-color-muted`, etc.).
2. **Proper use of plugin API**: The plugin correctly uses `PluginAPI.registerHook()`, `PluginAPI.registerShortcut()`, `PluginAPI.getCurrentContextTasks()`, `PluginAPI.persistDataSynced()`, and `PluginAPI.loadPersistedData()` -- all existing, documented APIs.
3. **Dispatched actions are all in the allowed list**: The three actions dispatched (`[Layout] Show AddTaskBar`, `[Task] SetSelectedTask`, `[Task] Toggle start`) are all in `ALLOWED_PLUGIN_ACTIONS` (`allowed-plugin-actions.const.ts`), so they will pass validation.
4. **Responsive design**: The grid layout includes a media query for `max-width: 600px` to stack widgets on smaller screens.
5. **XSS protection**: Uses `escapeHtml()` via `textContent` assignment for the task title before injecting it into the DOM.
6. **Graceful initialization**: `waitForPlugin()` polls for `PluginAPI` availability before calling `init()`.

### Issues

#### Security / Sandboxing

1. **`allow-same-origin` in iframe sandbox**: This is an existing architectural decision (not introduced by this PR), but worth noting: the `PLUGIN_IFRAME_SANDBOX` constant uses `allow-scripts allow-same-origin` together, which effectively allows the iframe content to remove its own sandbox attributes. Since this is a bundled plugin served from the same origin, the risk is lower, but it means any bundled plugin's `index.html` has full access to the parent page's origin. This is a pre-existing concern, not a blocker for this PR.

#### Correctness

2. **`loadSyncedData` vs `loadPersistedData` permission mismatch in manifest**: The manifest declares `"loadSyncedData"` as a permission, but the plugin code calls `PluginAPI.loadPersistedData()`. Looking at the plugin bridge code, the actual underlying method is `loadPersistedData` (with `loadSyncedData` being an alias in the iframe util). The manifest permissions appear to be informational only (not enforced at runtime beyond `nodeExecution`), so this won't cause a runtime failure, but the naming inconsistency is confusing. Consider using `"loadPersistedData"` in the manifest to match the actual API call, or document the alias clearly.

3. **Streak calculation has a bug in "best streak" logic**: The `calculateStreaks()` function computes the best streak by iterating sorted date keys in reverse, but the streak counter resets incorrectly. When it encounters a gap (non-consecutive days), it resets `streak = 0`, but the *next* iteration will increment it to 1. This means if there are two separate streaks, the second streak's count will be correct. However, if the *last* entry in the sorted list has `time > 0`, the streak counter ends at the correct value but `best` is only updated *inside* the loop body after incrementing -- so the final streak value is correctly captured. The logic works but is fragile and hard to follow. Consider simplifying:

    ```js
    // Simpler approach: group consecutive days
    var keys = Object.keys(streakData.days).filter(k => streakData.days[k].time > 0).sort();
    var best = 0, streak = 0, prev = null;
    for (var i = 0; i < keys.length; i++) {
      var curr = new Date(keys[i]);
      if (prev && (curr - prev) / 86400000 === 1) {
        streak++;
      } else {
        streak = 1;
      }
      if (streak > best) best = streak;
      prev = curr;
    }
    ```

4. **Timer drift / stale `baseTime`**: In `updateCurrentTaskWidget`, `baseTime` is captured once from `currentTask.timeSpentOnDay[todayStr]` when the current task changes. The `TASK_UPDATE` hook updates `currentTask` reference but does NOT restart the timer interval or recalculate `baseTime`. This means if the task's `timeSpentOnDay` is updated externally (e.g., by another device via sync, or by the main app's tracking), the timer display will be incorrect -- it will show `stale baseTime + elapsed since session start`. Consider recalculating `baseTime` and `sessionStart` when the task update hook fires for the current task.

5. **Unbounded streak data growth**: `streakData.days` accumulates entries indefinitely. Over years of use, this object could grow to thousands of entries (one per active day). Since this is persisted via `persistDataSynced` (which has rate limiting), it won't cause immediate issues, but consider pruning old entries (e.g., keep only the last 365 days, or the last N days needed for the best streak calculation).

#### Style / Code Quality

6. **No `author` field in manifest**: Other bundled plugins like `voice-reminder` include `"author": "Super Productivity"`. The dashboard manifest omits this. Minor but worth adding for consistency.

7. **`console.log('Dashboard Plugin loaded')` in plugin.js**: This debug log should be removed or guarded before shipping.

8. **`var` declarations throughout**: The `index.html` script uses `var` everywhere. While this works fine in an iFrame context (no module scope concerns), using `let`/`const` would be more modern and prevent accidental hoisting bugs.

#### Build / Infrastructure

9. **Large package-lock.json changes**: The PR includes 367 additions and 40 deletions to `package-lock.json` plus 512 additions each to two other `package-lock.json` files in plugin-dev subpackages. These appear to be vitest/esbuild dependency resolution changes unrelated to the dashboard plugin. Consider separating these lockfile updates into their own commit or verifying they are intentional.

10. **No tests**: There are no unit tests or E2E tests for the dashboard plugin. At minimum, consider adding a basic E2E test that loads the plugin and verifies the widgets render (similar to how the api-test-plugin is tested in `e2e/`).

### Verdict

The plugin is a solid demonstration of the iFrame plugin system. The main actionable items are the timer drift bug (#4), the `loadSyncedData`/`loadPersistedData` naming mismatch (#2), and the lack of tests (#10). The streak calculation (#3) works but could be simplified. The unbounded data growth (#5) is worth addressing before this ships to all users as a bundled plugin.

---

Zoriot:
@johannesjo what's the state here? Why is that not yet merged/finished?

PR #7164, opened 2026-04-07
Feat/default date today section

### LINKED ISSUE #7066: 💡 Having an Option for the Default Date for tasks when adding tasks in Today's section
### Problem Statement

I would like to be able to have an option in the settings where I can select what should be the default date added to a task when I am adding it even if it may be in the Today Section because right now when adding a task in Today Section the Default date is Today's date.

But I would like to be able to have it set to no date because I not always am I going to be wanting to add tasks to today only but each time I have to either clear the date or go to the inbox section to not have the default date set to today.

I would much rather have the date there set as no date by default because it is much easier for me to just type @ and add a date than it is to remove the default date each time as there is no shortcut to remove task date while creating the task only.

Image

### Possible Solution

Having an Option in settings under the Tasks Section allowing us to decide what we want the default date when adding tasks it could either be today's date when in today's section or no date even when it is in today's section.

I had thought of maybe just using the new brain dump plugin each time with a shortcut but even that sets the default date as today 
I also have just thought of using the inbox each time instead but that really breaks the flow of quickly adding the tasks which is why I am making this request.

### Additional context

_No response_

symonbaikov:
Problem
---
  When adding tasks in the Today section, today's date is automatically set as the default. Users who don't always want to schedule tasks for today have to manually clear the date each time,
  or switch to the Inbox section instead. There is no shortcut to remove the date while creating a task.

  Closes #7066

  Solution
---
  Added a boolean setting isSetDefaultDayForTodayTasks (default: true) under Settings → Tasks:
  ▎ "Set today's date as default when adding tasks in Today section"

  - When enabled (default): existing behavior is preserved — tasks added in Today get dueDay = today pre-filled
  - When disabled: the date field is empty by default; tasks go to Inbox without a scheduled date. The user can still type @today to explicitly schedule a task for today

  The check is applied in two places:
  - add-task-bar.component.ts — controls whether the date field is pre-filled in the UI
  - task.service.ts — controls whether dueDay is set as a fallback during task creation

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

Screenshot From 2026-04-07 11-56-52


johannesjo (commented):
## Review: Feat/default date today section

### Summary

This PR adds a new boolean setting `isSetDefaultDayForTodayTasks` (default: `true`) under Settings > Tasks that controls whether tasks added in the Today section automatically get today's date as `dueDay`. The feature is well-scoped, the implementation touches the right places, and tests are included. Nice work overall. A few items to address:

---

### Issues

#### 1. Reactivity bug in `add-task-bar.component.ts` — signal read inside RxJS `map` won't trigger re-emission

In `defaultDateAndTime$`, the new condition reads a signal inside an RxJS `map` operator:

```ts
defaultDateAndTime$ = this._workContextService.activeWorkContext$.pipe(
  map((workContext) => {
    // ...
    } else if (
      workContext?.type === WorkContextType.TAG &&
      workContext?.id === 'TODAY' &&
      this._globalConfigService.cfg()?.tasks?.isSetDefaultDayForTodayTasks  // <-- signal read
    ) {
```

The `cfg()` signal is read imperatively here. The observable only re-emits when `activeWorkContext$` emits a new value — it will **not** re-emit if the user toggles `isSetDefaultDayForTodayTasks` in settings while already in the Today view. In practice, this is likely minor since users rarely toggle this setting while adding a task, but it's still technically incorrect reactive code.

**Suggested fix:** Either:
- Convert `defaultDateAndTime$` to use `combineLatest` with `this._globalConfigService.cfg$` (or `tasks$`), so changes to the config also trigger re-evaluation. This is what `defaultProject$` already does in the same component (line 180-199).
- Or convert the whole thing to a computed signal (preferred per CLAUDE.md: "Prefer Signals to Observables if possible").

Note: In `task.service.ts`, the signal read is fine because `createNewTaskWithDefaults` is called imperatively (not in a reactive pipeline), so it correctly reads the current signal value at call time.

#### 2. Missing schema migration for existing users

The new `isSetDefaultDayForTodayTasks` property is added to `TasksConfig` and `DEFAULT_GLOBAL_CONFIG`, but there is no schema migration for existing users who already have a persisted `globalConfig.tasks` object without this field. When an existing user upgrades, their persisted `tasks` config will lack this property.

Looking at the reducer (`global-config.reducer.ts` line 33-35):
```ts
export const selectTasksConfig = createSelector(
  selectConfigFeatureState,
  (cfg): TasksConfig => cfg.tasks ?? DEFAULT_GLOBAL_CONFIG.tasks,
);
```

This only falls back to `DEFAULT_GLOBAL_CONFIG.tasks` if `cfg.tasks` is entirely nullish — it does **not** merge individual missing properties. So for existing users, `cfg.tasks.isSetDefaultDayForTodayTasks` will be `undefined`, which is falsy.

Since the default is `true` (preserving existing behavior), existing users would unexpectedly get the **opposite** behavior after upgrading — tasks in the Today section would stop getting today's date pre-filled.

**Suggested fix:** Either:
- Add a schema migration (bump `CURRENT_SCHEMA_VERSION`, add a migration that sets `isSetDefaultDayForTodayTasks: true` on existing `globalConfig.tasks` objects).
- Or handle `undefined` as `true` in the consuming code (e.g., `this._globalConfigService.cfg()?.tasks?.isSetDefaultDayForTodayTasks !== false`). This is the simpler approach but less clean.

**This is the most important issue in this PR** — without addressing it, the feature will break existing behavior for all upgrading users.

#### 3. Minor: Hardcoded `'TODAY'` string instead of `TODAY_TAG.id` constant

In the diff for `add-task-bar.component.ts`, the existing code already uses `workContext?.id === 'TODAY'` (line 212). While this PR didn't introduce this, the change touches that exact code block. It would be good to use the `TODAY_TAG.id` constant for consistency (which is already imported in this file at line 78). This is a pre-existing issue, so not blocking, but easy to fix while you're here.

---

### Positives

- **Clean, minimal changes** — the feature touches only the necessary files and follows existing patterns well.
- **Good test coverage** — two new unit tests in `task.service.spec.ts` cover both the disabled case and the "explicit dueDay overrides setting" case. The `WritableSignal` approach for test setup is well done.
- **Correct i18n handling** — translation key added only to `en.json` as required, with clear descriptive text.
- **Preserves backward compatibility by default** — `isSetDefaultDayForTodayTasks: true` means the feature is opt-out, preserving existing behavior (contingent on fixing issue #2 above).
- **Respects TODAY_TAG virtual tag architecture** — the change correctly modifies `dueDay` behavior without touching `tagIds`, consistent with the project's architectural decision that TODAY_TAG membership is determined by `dueDay`.
- **The `task.service.ts` change is well-placed** — the condition check is added at the right level, after the existing guards for `parentId`, `dueWithTime`, and explicit `dueDay`.

---

### Summary

The feature design and scope are solid. The **critical issue** is the missing migration/undefined-handling (issue #2), which would cause a silent behavior regression for existing users. Issue #1 (reactivity) is a correctness concern worth fixing. Once those are addressed, this looks good to merge.

---

johannesjo:
Hi @symonbaikov, thanks a lot for taking this on and for the thorough work — the implementation is clean, it's nicely opt-in (default-on keeps current behavior), and I appreciate the test coverage. 🙏

I'm going to **put this one on hold** for now rather than merge it. The feature itself is reasonable, but it adds a permanent new toggle to the Tasks settings, which is already fairly crowded — and the underlying request (#7066) currently only has a couple of upvotes, so I'd like to see a bit more demand before committing to maintaining another setting long-term.

This isn't a "no" — if the issue gathers more support, I'd be happy to revisit. If/when we do, it'll need a rebase (master has since moved the date source to `DateService.todayStr()`, so the add-task-bar part would need to keep that rather than `getDbDateStr()`).

Really sorry to leave it in limbo after the effort you put in, and thank you again for contributing\! 💖

PR #7166, opened 2026-04-07
Feat/clipboard paste attachments

### LINKED ISSUE #6629: 💡 Direct Image Paste into Attachments
### Problem Statement

Currently, images cannot be pasted directly into the attachment field. Users must first upload the image to an external site, generate a URL, and then paste that link. This breaks workflow, wastes time, and adds friction—especially for quick task updates or bug reports. It also raises privacy concerns since images must be hosted on third-party platforms.

### Possible Solution

Enable direct image paste (Ctrl + V) into the attachment section
Open the system file picker (native file manager dialog)

_No response_

### Additional context

_No response_

symonbaikov:
Closes #6629

## Problem

Adds @HostListener('paste') to TaskDetailPanelComponent to intercept Ctrl+V when focus is on the panel (not in a textarea/input), saves the image via ClipboardImageService and creates an IMG attachment via TaskAttachmentService. Reuses existing clipboard image infrastructure (IndexedDB for web, filesystem for Electron). Adds a subtle hint 'Ctrl+V to paste image' below the Add Attachment button.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

image


johannesjo (commented):
## Review: Feat/clipboard paste attachments

Nice feature addition -- the approach of reusing `ClipboardImageService.handlePasteWithProgress()` and the existing attachment infrastructure is sound. The test coverage is good. Here are my findings:

---

### Issues to address

**1. Missing text-content check -- will steal paste from text clipboard (Bug)**

`ClipboardPasteHandlerService.handlePaste()` (line 28-34) explicitly checks for `text/plain` and `text/html` and bails out early, because applications like OneNote put both text and an image on the clipboard simultaneously. The new `onPaste` handler does **not** have this guard:

```ts
// In ClipboardPasteHandlerService (existing):
const hasText =
  ev.clipboardData.types.includes('text/plain') ||
  ev.clipboardData.types.includes('text/html');
if (hasText) {
  return false; // Let default text paste behavior handle it
}
```

Without this, pasting text content when the panel is focused (but not in an input) will be silently consumed and turned into an image attachment if the clipboard happens to also contain an image representation. Add the same early-return check.

**2. Reuse `isInputElement()` utility instead of duplicating logic**

The codebase already has `isInputElement()` in `src/app/util/dom-element.ts` which checks `INPUT`, `TEXTAREA`, and `isContentEditable` -- exactly the same logic written inline in `onPaste`. The component already imports it (line 79). Replace the inline check:

```ts
// Before (duplicated logic):
if (target.tagName === 'TEXTAREA' || target.tagName === 'INPUT' || target.isContentEditable) {
  return;
}

// After (reuse existing utility):
if (isInputElement(target)) {
  return;
}
```

**3. `fakeAsync(async () => ...)` is unreliable in Angular tests**

Copilot already flagged this, but it's worth reinforcing: Angular's `fakeAsync` does not properly intercept native Promises returned by `async` functions. The tests may appear to pass in some environments but will behave unpredictably. Two clean options:

- **Option A**: Drop `async/await`, keep `fakeAsync`, call `component.onPaste(event)` without `await`, then `tick()` to flush the microtask queue.
- **Option B**: Use `waitForAsync` + `fixture.whenStable()` instead of `fakeAsync/tick`.

**4. Hint text says "Ctrl+V" but macOS uses Cmd+V**

The hardcoded `"Ctrl+V to paste image"` string is incorrect for macOS users. The codebase already has `IS_MAC` (`src/app/util/is-mac.ts`). Consider either:
- Using two translation keys and selecting based on `IS_MAC`, or
- Using a generic phrasing like "Paste image from clipboard" that is platform-neutral.

**5. No error handling around the `await`**

While `_saveImageFromClipboard` currently catches internally and always resolves, it would be defensive to wrap the `await` in a try/catch in case the contract changes or an unexpected error surfaces. The existing `ClipboardPasteHandlerService` doesn't have this either, but since this is new code it's a good opportunity to be robust:

```ts
try {
  const result = await progress.resultPromise;
  if (result.success && result.imageUrl) {
    // ... addAttachment
  }
} catch (err) {
  console.error('[CLIPBOARD] Paste attachment failed:', err);
}
```

---

### Minor / style nits

**6. SCSS: `opacity: 0.4` + `color: var(--text-color-muted)` is double-dimming**

`--text-color-muted` is already defined as `rgba(44,44,44, 0.6)` (light) / `rgba(235,235,235, 0.65)` (dark). Stacking `opacity: 0.4` on top makes the text effectively 24-26% opacity, which may be too faint for accessibility. I'd suggest removing `opacity: 0.4` and relying solely on `--text-color-muted`, or using a slightly less aggressive opacity like `0.7`.

**7. Attachment title is not translatable**

The hardcoded `title: 'Pasted image'` string should go through the translation system for consistency with the rest of the app. This is also done in `ClipboardPasteHandlerService` (same hardcoded string there, so it's an existing issue), but new code is a good opportunity to fix it.

---

### Positives

- Clean reuse of the existing `ClipboardImageService` infrastructure -- no reinventing the wheel.
- Correct scoping: the paste handler properly skips input/textarea/contenteditable targets.
- Good test coverage with 6 test cases covering the happy path, all skip conditions, and failure cases.
- Translation key and `en.json` updates follow project conventions (only editing `en.json`).
- SCSS uses design-system variables (`--s-quarter`, `--text-color-muted`).

Overall a solid feature. The text-content check (issue #1) is the most important to fix before merge, as it would cause a user-facing bug with mixed text+image clipboard content.

---

johannesjo:
Can you have a look at the conflicts @symonbaikov ?

symonbaikov:
@johannesjo Yes, will resolve them today.

johannesjo:
Thank you very much! <3

Issue #7183, opened 2026-04-10
💡 Enhance the BrainDump Plugin for subtasks/markdown


adnoh:
### Problem Statement

I must commit - I like the new braindump plugin - even if its super simple :)
What I'm missing in my worflow is the possibility to create subtasks along with my main tasks. Its a bummer because we still can't convert tasks to subtasks and assign a parent (if I'm not wrong). So I can't really structure at all while mentally offloading. 

### Possible Solution

I guess if it can use the MarkdownPasteService or to be super easy (I would prefer) just a blank space as first character (or a "*") to recognise it as a subtask would help me quite a lot.
And of course - finally being able to move a task over a parent to make it its subtask would help me even more ^^

### Describe alternatives you've considered

Reference: 
https://github.com/super-productivity/super-productivity/commit/91ff0e37e5590c3c898fa1949a70583ba4e491c6 and https://github.com/super-productivity/super-productivity/issues/2963#issuecomment-3248744620

### documentation gap

- [ ] ... related docs should be updated as part of the same PR whenever possible.
- [ ] ... I can help improve the documentation.

johannesjo:
Thank you!

PR #7184, opened 2026-04-10
Implement task parsing with sub-tasks

adnoh:
Added a new function to parse tasks with sub-tasks from a text input, improving task management capabilities. Updated the task addition logic to utilize this new parsing method. Should solve #7183
Please check if we should limit the nesting of subtasks, as far as I know a subtask can NOT have another subtask (which would be nice if it could anyways)

## Problem

#7183 - Enhance the BrainDump Plugin for subtasks

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/packages/plugin-dev/brain-dump/plugin.js
+++ b/packages/plugin-dev/brain-dump/plugin.js
@@ -43,18 +43,27 @@ function parseTasksWithSubTasks(text) {
   }
 
   // Check if we have mixed input (both bullets and plain text)
-  var hasBullets = parsedLines.some(function (p) { return p.isBullet; });
-  var hasPlainText = parsedLines.some(function (p) { return !p.isBullet; });
+  var hasBullets = parsedLines.some(function (p) {
+    return p.isBullet;
+  });
+  var hasPlainText = parsedLines.some(function (p) {
+    return !p.isBullet;
+  });
 
   if (hasBullets && hasPlainText) {
     PluginAPI.showSnack({
-      msg: 'Warning: ' + plainTextCount + ' plain-text line(s) detected. These will be added as regular tasks.',
+      msg:
+        'Warning: ' +
+        plainTextCount +
+        ' plain-text line(s) detected. These will be added as regular tasks.',
       type: 'WARNING',
     });
   }
 
   // Find the minimum indentation level to normalize
-  var bulletLines = parsedLines.filter(function (p) { return p.isBullet; });
+  var bulletLines = parsedLines.filter(function (p) {
+    return p.isBullet;
+  });
   if (bulletLines.length > 0) {
     var minIndentLevel = Math.min.apply(
       Math,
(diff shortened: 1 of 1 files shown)


johannesjo (commented):
Thanks for the contribution — the core idea (structured parser with a plain-text fallback) is clean and well-separated. A few issues need fixing before this is ready to merge.

## Critical: 4-space indentation silently drops sub-tasks

`parseLineStructure()` calculates indent level as `tabCount + Math.floor(spaceCount / 2)`, treating 2 spaces as one level. The most common convention is **4 spaces per indent**, which yields `Math.floor(4/2) = 2`. But the sub-task collection loop only picks up `indentLevel === 1`, so those items are silently skipped:

```
- Main task
    - Sub task   ← 4 spaces → indentLevel 2 → dropped silently
```

**Fix:** detect the smallest non-zero indent in the document and normalize relative to that, so `indentLevel 1` always means "one step deeper than the parent", regardless of whether the user uses 2 or 4 spaces.

## Critical: Mixed plain-text / bullet input silently loses lines

`parseLineStructure()` returns `null` for plain-text lines (no `- ` / `* ` prefix). The structured path runs whenever `parsedLines.length > 0` — so if any line is parseable as a bullet, the structured path runs and all plain-text lines are silently dropped:

```
- Structured task
Plain text task     ← lost with no warning
  - Sub task
```

The fallback only fires when `parsedTasks === null` (i.e. *zero* bullet lines). Consider treating unparseable lines as plain-bullet tasks, or at minimum show a snack warning listing how many lines were skipped.

## Minor: deeply-nested lines are iterated but never saved

In the look-ahead loop, `j` advances past items with `indentLevel > 1`, but only `indentLevel === 1` items are pushed to `subTasks`. Deeper items are silently dropped. This is partly a consequence of the indent-calculation bug above, but even with correct normalization the code should either flatten them to level-1 sub-tasks or warn the user.

## Minor: `dueDay` on sub-tasks

`dueDay` is applied to both parent and sub-tasks. Sub-tasks in Super Productivity normally inherit the parent's date. Passing an independent `dueDay` on a sub-task may be a no-op or may cause unexpected behaviour — worth verifying against the API.

## Suggestion: expose the expected format to users

Nothing in the UI hints that `- item` syntax or indentation does anything special. Even updating the textarea placeholder to something like:

```
- Main task
  - Sub task
Another task
```

would make the feature discoverable.

---

Once the indent-normalization and silent-drop issues are addressed this will be a solid addition. Happy to clarify any of the above.

adnoh (commented):
Thanks for your feedback. I updated the PR

adnoh (commented):
fixed bullet detection

johannesjo:
Thank you very much! <3

adnoh:
your welcome, happy I could help. maybe I should improve the explanation to users:

```
- Main task
  - Sub task
  - [x] Finished Sub task
Another task
```

PR #7210, opened 2026-04-13
fix(plugins): use asset-specific size errors

### LINKED ISSUE #7202: 🚨 Plugin size error is confusing - manifest/index.html/ico uses the same error string
### Version Used

18.1.0

### Expected Behavior

If index.html is too large it should say index.html is too large

### Current Behavior

https://github.com/super-productivity/super-productivity/blob/70414145a35fdc5c7380ca08a3d8f1d5e2672f0b/src/app/plugins/plugin.service.ts#L1000-L1070

### Steps to Reproduce

1. Develop a plugin with big index.html
2. Install

### Console Output

```shell
This is another problem - there's no console output
```

### documentation gap

officialasishkumar:
## Problem

Fixes #7202. Uploaded plugin ZIP validation reused the manifest size error for oversized `index.html` and icon files, making it unclear which plugin asset exceeded the limit.

## Solution

Route manifest, `index.html`, and icon size checks through asset-specific translation keys and add English strings for the new messages. A focused unit test covers the translation-key mapping.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

Validation:
- `npm run checkFile src/app/plugins/plugin.service.ts`
- `npm run checkFile src/app/plugins/plugin-size-error.util.ts`
- `npm run checkFile src/app/plugins/plugin-size-error.util.spec.ts`
- `npm run int:test`
- `CHROME_BIN=/home/sarthak/.cache/ms-playwright/chromium-1208/chrome-linux64/chrome npm run test:file -- src/app/plugins/plugin-size-error.util.spec.ts`


johannesjo (changes requested):
Thanks for tackling this — the user-visible bug is real and your translation keys + English strings are correct. A couple of things to address before merging:

---

### Important: unnecessary abstraction

`plugin-size-error.util.ts` (and its spec) can be dropped entirely. The three call sites are adjacent lines in the same service, each used exactly once. The minimal fix is a two-line change directly in `plugin.service.ts`:

```diff
-            this._translateService.instant(T.PLUGINS.MANIFEST_TOO_LARGE, {
+            this._translateService.instant(T.PLUGINS.INDEX_HTML_TOO_LARGE, {
```
```diff
-            this._translateService.instant(T.PLUGINS.MANIFEST_TOO_LARGE, {
+            this._translateService.instant(T.PLUGINS.ICON_TOO_LARGE, {
```

The manifest check was already using the right key, so it doesn't need to change at all.

Project guideline: *"Don't create helpers, utilities, or abstractions for one-time operations."*

---

### Minor: `fileSize` is computed but never shown

All three error message templates (including the pre-existing `MANIFEST_TOO_LARGE`) only interpolate `{{maxSize}}` — `fileSize` is computed and passed but silently dropped. The codebase already has `FILE_TOO_LARGE: "File too large ({{fileSize}}MB). Maximum size is {{maxSize}}MB"` as a precedent that includes both values.

Either add `(got {{fileSize}}KB)` to all three strings so users see their file size, or remove the dead `fileSize` computation. Either way is fine; just make it consistent.

---

### Minor: stale comments

The two comments `// Validate index.html size (same as manifest for now)` and `// Validate icon size (same as manifest for now)` now refer only to the size *constant* being shared, not the error message. They should be updated or removed to avoid future confusion.

officialasishkumar:
Inlined the asset-specific translation keys at the call sites, removed the one-off helper/spec, and cleaned up the stale size-limit comments. Force-pushed the branch.

johannesjo:
Thank you very much!

PR #7220, opened 2026-04-13
fix(persistence): increase IndexedDB retry window for session-restart locks

### LINKED ISSUE #7191: 🚨 "Database Error - Cannot Load Data" (indexedDB) when session is closed and reopened with autostart enabled
**Version Used**
17.2.4 (Linux Mint Cinnamon, Flatpak)

**Expected Behavior**
When logging out of a Linux Mint session and logging back in, with Super Productivity set to autostart, the app should open normally and load data without errors.

**Current Behavior**
After logging out of the Linux Mint (Cinnamon) session and logging back in, the app (launched automatically at session start) displays the following error:

"Database Error - Cannot Load Data
Super Productivity cannot open its database. This may be caused by:
- Low disk space
- Temporary file lock (try closing other tabs)
- Storage corruption
Technical details: Internal error opening backing store for indexedDB.open."

The error occurs consistently on every logout/login cycle. Workaround: manually closing the app, waiting a few minutes, then relaunching it starts correctly.

**Steps to Reproduce**
1. Install SP via Flatpak on Linux Mint Cinnamon
2. Set Super Productivity to autostart at session login
3. Open a session with SP running
4. Log out of the session (Menu → Log out)
5. Log back in
6. SP launches automatically and immediately shows the indexedDB error

Tested multiple times — behavior is consistent.

Note: Similar errors have been reported in older issues (#375, #419, #955) but none identified this specific trigger (logout/login cycle with autostart). Closing SP manually before logging out prevents the error.

**Your Environment**
- Version: 17.2.4
- OS: Linux Mint Cinnamon
- Install method: Flatpak (com.super_productivity.SuperProductivity)
- Sync: Dropbox (3 devices)

### Console Output

```shell

```

### documentation gap

costajohnt:
## Summary

Fixes #7191.

On Linux desktop environments (especially Flatpak with autostart), logging out and back in can leave stale LevelDB locks for 5–15+ seconds. The previous retry window (~3.5s) was too short to outlast this, causing "Database Error - Cannot Load Data" on every logout/login cycle.

### Changes

**Longer retry window for lock-related failures.** Bumps `IDB_OPEN_RETRIES` from 3 to 5 and `IDB_OPEN_RETRY_BASE_DELAY_MS` from 500ms to 1000ms. Lock retry window is now ~31s (1 + 2 + 4 + 8 + 16). Rebased onto current master after the branch fell 67 commits behind; the `IDB_OPEN_RETRIES=3` on master is the result of an unintentional revert of `e8a3e156eb` by `0e9218bd68` (called out in your review).

**Error-type-aware backoff.** The long window only applies when the error looks lock-related: `InvalidStateError` name, or message contains "backing store". Other errors use a new `IDB_OPEN_RETRIES_NON_LOCK = 3` budget (~7s total), matching master's pre-PR retry count. This addresses the concern that every op-log read/write awaits `_ensureInit()`, so a 31s wait on a non-lock error would block the whole subsystem before `_showIndexedDBOpenError`'s alert dialog reaches the user. Classification happens once on the first failure and the budget is not re-widened.

**Same fix in both open paths.** `OperationLogStoreService` and `ArchiveStoreService` both use these constants and both got the classification + budget-shrink. Heads-up for the diff.

**New helper `isLockRelatedIdbOpenError`.** Lives in `op-log-errors.const.ts` next to `isConnectionClosingError`. Mirrors that sibling's `DOMException`-guarded pattern so an `InvalidStateError` `DOMException` isn't silently misclassified in runtimes where `DOMException` doesn't extend `Error`.

**Spec hardening.** Replaced the formula-recomputes-formula assertion with concrete floor checks (`IDB_OPEN_RETRIES >= 5`, `IDB_OPEN_RETRY_BASE_DELAY_MS >= 1000`, `total window >= 20000ms`) plus a `#7191` comment. Added behavioral tests for `_openDbWithRetry` covering lock vs non-lock budgets and success-after-retry, plus classifier tests including `DOMException` variants.

## Test plan

- [x] All 8542 tests pass locally
- [x] `tsc --noEmit` clean
- [x] `ng lint` clean
- [x] New `isLockRelatedIdbOpenError` tests (Error, DOMException, string, null/undefined/object, non-matching name)
- [x] New `_openDbWithRetry` behavioral tests: non-lock error exhausts after `1 + IDB_OPEN_RETRIES_NON_LOCK` attempts; lock error uses full `1 + IDB_OPEN_RETRIES` budget; lock-then-success resolves with the DB

### DIFF
--- a/src/app/op-log/core/operation-log.const.spec.ts
+++ b/src/app/op-log/core/operation-log.const.spec.ts
@@ -28,14 +28,20 @@ describe('IndexedDB open retry configuration', () => {
     expect(totalDelayMs).toBeGreaterThanOrEqual(MINIMUM_LOCK_RETRY_WINDOW_MS);
   });
 
-  it('non-lock retry budget is much shorter than the lock budget', () => {
-    // Non-lock errors must fail fast so the hydrator's reload-on-error path
-    // does not create a reload -> wait -> reload loop. See #7191.
+  it('non-lock retry budget is shorter than the lock budget', () => {
+    // Non-lock errors must fail fast: every op-log read/write awaits
+    // _ensureInit(), so a 31s retry blocks the subsystem for 31s before the
+    // hydrator's alert dialog reaches the user. See #7191.
     expect(IDB_OPEN_RETRIES_NON_LOCK).toBeLessThan(IDB_OPEN_RETRIES);
-    // Short window should stay under ~5s so repeated reloads are tolerable.
+    // With IDB_OPEN_RETRIES_NON_LOCK=3 and base 1000ms: 1+2+4 = 7s ceiling
+    // (delays before attempts 2, 3, 4; no post-delay on the final attempt).
+    // The formula (2^n - 1) * base overstates this slightly because it
+    // includes a hypothetical delay after the final attempt, but it's a
+    // conservative upper bound. Keep well under 10s so failures surface
+    // quickly.
     const nonLockWindowMs =
       (Math.pow(2, IDB_OPEN_RETRIES_NON_LOCK) - 1) * IDB_OPEN_RETRY_BASE_DELAY_MS;
-    expect(nonLockWindowMs).toBeLessThan(5000);
+    expect(nonLockWindowMs).toBeLessThan(10000);
   });
 });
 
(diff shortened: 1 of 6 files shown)


johannesjo:
Thank you very much!



Heads up on a couple of things that may affect this PR:

1. **The diff is based on an older state of `master`.**
   It changes `IDB_OPEN_RETRIES` from 4 → 5, but `master` is currently back at 3. The 4 was reverted in `0e9218bd68` — seemingly unintentionally, since it is buried inside an unrelated sync-validation commit. It would be worth rebasing so the intended values are clear after merge.

2. **There is already a layered fix for #7191 on `master` via `e8a3e156eb`:**

   * clears stale LevelDB `LOCK` files in the Electron main process before the renderer opens IndexedDB (root-cause fix, gated on `requestSingleInstanceLock()`)
   * silently auto-reloads on the first backing-store error
   * bumped retries from 3 → 4 (though that was later reverted; see point 1)

   If you can still reproduce the error after those changes shipped, that is the most valuable signal. Please share the version you are testing on. If the `LOCK` cleanup is working, extending the window to ~31s mostly just delays the failure path for genuinely broken DBs.

3. **UX concern with a 31-second window.**
   `_ensureInit()` is awaited by every op-log read/write, and `operation-log-hydrator.service.ts` auto-reloads on backing-store errors. A reload → 31s wait → reload loop would feel like a hang. If the retry extension is still needed, consider making it error-type-aware: longer backoff only when the exception looks lock-related (`InvalidStateError`, `"backing store"` messages), rather than applying it blanket-style.

4. **Small nit on the spec.**
   The test recomputes the backoff formula, so it passes as long as the formula matches itself. It would not catch a behavior change such as jitter, linear backoff, or a cap. Asserting a concrete minimum window (`>= 20000`) with a comment explaining that the floor exists because of #7191 would survive a refactor better. Also, `archive-store.service.ts` uses the same constants, so its retries quadruple as well — worth mentioning in the PR description.

johannesjo:
Thank you very much! <3

PR #7243, opened 2026-04-16
feat(i18n): add missing Turkish translations

Cyber-Syntax:
I added missing Turkish translations for new features and updated some poor translations that could lead to misunderstandings.


johannesjo:
Thank you very much!

PR #7246, opened 2026-04-16
feat(schedule): add scheduling warnings for overlapping tasks and outside work hours

xsky1020:
## Summary

This PR adds scheduling warnings to make time conflicts and overload situations more visible during task planning.

## Changes

- highlight tasks that overlap with other scheduled tasks
- show an `Outside work hours` label for tasks scheduled outside configured work hours
- show a warning on the Today view when the remaining estimate exceeds 8 hours
- add real-time warnings in the scheduling dialog when a selected time:
  - overlaps with another scheduled task
  - falls outside configured work hours

## Why

Currently, users can schedule tasks into conflicting or impractical time slots without immediate feedback. This PR adds lightweight warnings directly to the task list, Today view, and scheduling dialog so these issues are visible while planning.

## Implementation Notes

This PR is split into three commits:

1. task-level conflict and out-of-hours indicators
2. Today view overload warning for remaining estimate > 8h
3. real-time warning messages in the scheduling dialog

## Verification

- `npx tsc -p tsconfig.json --noEmit`
- `npm run test:file -- src/app/features/tasks/util/get-time-conflict-task-ids.spec.ts`


johannesjo:
Thanks for the contribution\! The feature is genuinely useful and the overlap-detection algorithm is clean. A few things to address before merge — grouped by severity.

## Blocker

**i18n.** Per `CLAUDE.md`, all UI strings must go through `T` / `TranslateService`, with keys added to `src/assets/i18n/en.json` only. The PR introduces several hardcoded English strings:

- `dialog-schedule-task.component.html` — `Overlaps with another scheduled task.`, `Falls outside your configured work hours.`
- `dialog-schedule-task.component.ts` (submit handler) — `This time overlaps…` / `This time falls outside…` passed with `isSkipTranslate: true`
- `task.component.html` — `Outside work hours`
- `work-view.component.html` — `Over 8h`

Please add keys (e.g. under `T.F.PLANNER.S.*`) and bind via `| translate`. Drop `isSkipTranslate`.

## Correctness / behaviour

1. **Today-only conflict scope is inconsistent with the dialog.** `task.component.ts` gates `hasTimeConflict` on `_dateService.isToday(...)` and passes `isToday` as the `sameDay` predicate to `getTimeConflictTaskIds`. But the dialog's `scheduleWarnings` derives `sameDay` from the user's selected day. Result: scheduling two overlapping tasks on a future day warns in the dialog but never in the list. Pick one scope (likely "any day") and key the conflict set by `dbDateStr`.
2. **`isOutsideWorkHours` chip ignores `task.isDone`.** `getTimeConflictTaskIds` filters done tasks; the work-hours util does not, so completed tasks scheduled outside hours still show the chip. Add a done guard.
3. **Submit shows a snack and schedules anyway.** The dialog already displays the warning inline; on submit, the same warnings open a snack-bar and `_taskService.scheduleTask(...)` runs unconditionally. That's a double notice with no opt-out. Either drop the snack (the inline warning is enough) or replace it with a confirm step so the warning is actionable.
4. **`isTaskOutsideWorkHours` does not honour `isLunchBreakEnabled`** even though `calculateAvailableHours` does. A task entirely inside the configured lunch break is treated as inside work hours. Either match `calculateAvailableHours` or add a code comment documenting the intentional divergence.
5. **`isTaskOutsideWorkHours` calls `getDateTimeFromClockString` unguarded.** That function can throw on malformed config strings; `calculateAvailableHours` wraps it in try/catch and falls back to `DEFAULT_WORK_HOURS`. The new util should mirror that defensiveness so a corrupt `workStart`/`workEnd` doesn't crash the dialog computation.

## Performance

6. **Per-`TaskComponent` recomputation.** `task.component.ts` declares `timeConflictTaskIds = computed(() => getTimeConflictTaskIds(...))` on the component. `computed()` memoises per instance, so with N rendered task rows the same global O(N²) sweep runs N times whenever the task list changes. Hoist this to a memoised NgRx selector (`createSelector` is globally cached) — e.g. `selectTimeConflictTaskIds` keyed by day — and let the component just `.has(task.id)`.
7. **Dialog `scheduleWarnings` rebuilds the conflict set on every keystroke.** A workspace with hundreds of scheduled tasks pays O(N²) per signal tick. Reuse the same store selector and only check the candidate against the precomputed conflicts.

## Minor

8. **Hardcoded `_previewTaskId = '__schedule-preview__'`** can theoretically collide with a real task id. Use `Symbol()` (or filter the candidate by reference rather than id) so the contract can never be subverted.
9. **`MIN_TASK_DURATION = 60 * 1000` duplicated** in `get-time-conflict-task-ids.ts` and `is-task-outside-work-hours.ts` — extract a shared constant.
10. **Redundant `isToday` check** in `hasTimeConflict` — the conflict set is already today-only.
11. **`bind` inside `computed`** allocates a fresh bound function each recomputation: `this._dateService.isToday.bind(this._dateService)`. Hoist to a class field, or use `(ts) => this._dateService.isToday(ts)`.
12. **No spec for `is-task-outside-work-hours.ts`.** It has timezone-sensitive logic (`dateStrToUtcDate`, `getDateTimeFromClockString`) and several branches (disabled flag, undefined config, end>workEnd vs start<workStart, lunch-break interaction). Worth a unit test.
13. **Drop `as TaskWithDueTime`** in the spec — `TaskWithDueTime extends Task`, the literal already satisfies the type.
14. **Stray blank line** in the import block of `work-view.component.ts`.
15. **Getter/setter wrapping signals** for `selectedDate`/`selectedTime` is unusual in this codebase. The idiomatic alternative is `selectedDate = signal<…>(null)` with `[ngModel]="selectedDate()" (ngModelChange)="selectedDate.set($event)"`.

## Accessibility / theme

- The warning icons have no `aria-label` / `aria-hidden`, and the chip itself has no `role="status"` — screen-reader users get no structured signal that there is a warning.
- The chips render at `font-size: 11px` over a `color-mix(... var(--c-warn) 14%)` wash. Worth checking contrast in the dark theme to make sure the text remains legible.

## Project conventions

- The PR description mentions `npx tsc --noEmit` and one spec; please also run `npm run checkFile <path>` on each modified `.ts` / `.scss` file (project rule in `CLAUDE.md`).
- Commit scope: `schedule` isn't an established scope in `git log` — `tasks` / `planner` / `work-view` would match existing history better.

Happy to chat about any of these — the i18n + conflict-scope + performance points are the main ones; the rest are polish.

Zoriot:
@xsky1020 you wanna do the requested changes? Would be nice, sounds like a useful thing
You also have to rebase in that case

PR #7283, opened 2026-04-20
fix(search): scroll to selected task after search navigation

aakhter:
## Summary
- wait for the target task row to have real layout before treating search navigation as successful
- center the selected task inside the nearest scrollable task container
- add coverage for delayed task rendering in the layout service tests

## Root cause
Search-driven navigation could resolve the target task before the row had a real rendered height. That zero-height element made the scroll calculation collapse to the top of the list, so the task became selected without being brought into view.

## Validation
- reproduced with search navigation into KB Must Act
- validated the patched Windows build locally after rebuilding on aa-desktop

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -189,7 +189,6 @@ export class AppComponent implements OnDestroy, AfterViewInit {
   );
 
   private _subs: Subscription = new Subscription();
-  private _intervalTimer?: NodeJS.Timeout;
 
   constructor() {
     this._startupService.init();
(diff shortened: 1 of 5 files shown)


johannesjo:
@aakhter Thank you very much! I am unable to reproduce the original issue though. Could you maybe record a short video, so we are on the same page? That would be great! Also there are two failing unit tests now:

```
TOTAL: 2 FAILED, 8566 SUCCESS


1) should focus pending task id with preventScroll when hide is called without parameter
     LayoutService Focus restoration
     Expected spy focus to have been called with:
  [ Object({ preventScroll: true }) ]
but it was never called.
    at <Jasmine>
    at apply (src/app/core-ui/layout/layout.service.spec.ts:113:42)

2) should focus newly created task with preventScroll when task id provided
     LayoutService Focus restoration
     Expected spy focus to have been called with:
  [ Object({ preventScroll: true }) ]
but it was never called.
    at <Jasmine>
    at apply (src/app/core-ui/layout/layout.service.spec.ts:92:38)


```

aakhter:
Please find recording [here](https://www.dropbox.com/scl/fi/0wna8fwnj1qed9tnol7fc/sp-search-focus.gif?rlkey=b58qnymv3pmxq0gdqmrtu7f0o&st=sumsezq9&dl=0)

Sequence:
1. Search 
2. Select item in search result
3. Have to find (scroll screen) the highlighted item in long project screen

I'll work on fixing the issues.

johannesjo:
Thank you! <3

Issue #4324, opened 2025-05-11
🚨 [Gitea Integration] Backlog tasks automatically moved to regular task


tixwho:
### Version Used

AUR superproductivity-bin v12.0.2-1

### Expected Behavior

After setting up Gitea integration, and check "Auto import to default project", it should arrives in the backlog and stay there until being manually moved out.

### Current Behavior

Reoccurence of #3933. Not sure if it's a regression. The only difference is that I'm connecting to a self-hosted gitea instance, not gitlab instance.

Imported tasks first stay in the backlog for a while (a few minutes), but then automatically move to regular tasks after a `POLL GITEA" event.

The imported event _might or might not_ be imported to "Today" category.

Moving these imported tasks back to backlog will not stop them from automatically bouncing back. This bug does not affect native tasks created within the app.

### Steps to Reproduce

1. Connect to a Gitea project with open issues
2. Assign "Default Super Productivity Project" and check "Auto import to default project"
3. Go to the project, check the project backlog
4. Wait for a `POLL GITEA` event in console
5. All tasks being moved up automatically to regular tasks with notification
6. Manually move all tasks back to backlog, wait for another `PULL GITEA` event
7. All tasks bounce back to regular tasks.

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell
[WorkContext] Set Active Work Context 
[Task] SetSelectedTask 
POLL GITEA
[Project] Auto Move Task from backlog to regular 
[Persistence] Save to DB 
[Project] Auto Move Task from backlog to regular 
[Persistence] Save to DB 

.... repeat
```

Issue #3933, opened 2025-01-24
[GITLAB INTEGRATION] - add in Backlog bug


pierrejo:
### Your Environment

- Version used: snap
- Operating System and version: Ubuntu (24.04)
- Desktop Environment: linux

### Expected Behavior

When a task from a Gitlab issue arrives in a project by default, if the project has a backlog, then it arrives there and should stay there.

### Current Behavior

When a task from an issue arrives in a project by default, if the project has a backlog, then it arrives in it, but after a while, it's automatically moved from the backlog into the project.

Even if I manually put the task back into the backlog without specifying that it's a “today”, it still comes back into the project after a certain amount of time (gitlab sync time?).

The other tasks I've created since Super-productivity, on the other hand, remain in the backlog.

### Steps to Reproduce (for bugs)

1. Add Gitlab Integration (from self-hosted instance with custom URL)
2. Add an issue on Gitlab
3. Check S-P and your project backlog
4. The task will be moved up to the project

### Console Output

There's no errors

johannesjo:
I am unable to reproduce this.  What kind of message appears, when the tasks are moved up? Can you maybe also copy the console output from when it happens here?

johannesjo:
Nevermind! I was able to reproduce and fix this. Thank you very much!

Issue #1609, opened 2021-10-21
Set task as done if ticket is closed


BlackScorp:
## Problem Statement
Right now i have manually to stop the task even if i closed the ticket on github via push

## :grey_question: Possible Solution
When you fetch the github issues, complete the task if the issue is closed

johannesjo:
Thanks for opening this up! This should  actually already be implemented.

Issue #1524, opened 2021-09-19
GitLab tasks should be marked completed when the issue got closed on the server


jurgenhaas:
I have successfully linked my first GitLab project and receive all the tasks in the client. But they remain in the list of open tasks until I manually mark them as completed - even if they had been closed on the GitLab server. The client recognizes the status change but I wonder why it doesn't mark the task as completed then.

johannesjo:
Thanks for opening this up! I am all for adding this and I made several preparations to add this. The rest api is a bit limited and there are certain rate limits so there isn't an endless amount of requests available. But maybe there is a way to batch request just the updates for the issues already on the list via the graphql api:
https://docs.gitlab.com/ee/api/graphql/

jurgenhaas:
I do have a self-hosted GitLab instance and could give you a test account there without any rate limits, is that helps you for testing?

johannesjo:
Thank you for the offer. If I need it, I get back to you.

ishanarora:
> Unfortunately the rest api is a bit limited and there are certain rate limits so there isn't an endless amount of requests available, that's why only open issues are requested currently, so changes to closed ones would never appear.

Can't we make this behavious an option in the "Setup GitLab for Project", since the rate limits don't affect self-hosted instances?

johannesjo:
> Can't we make this behavious an option in the "Setup GitLab for Project", since the rate limits don't affect self-hosted instances?

I am open for implementing this as optional (opt in) feature. PRs dealing with this are welcome.

ishanarora:
@cbergmann might this be of interest to you?

PR #7288, opened 2026-04-20
Fix/gitea issue id and done sync

sespiros:
## Problem

Three independent defects in the Gitea/Forgejo issue provider compound into a confusing sync UX:

1. **Closed upstream issues never propagate to SP.** `gitea-common-interfaces.service.ts::getAddTaskData` returns only `{title, issueWasUpdated, issueLastUpdated}` (no `isDone`, no `issueId`). Closing a Forgejo/Gitea issue therefore does not flip the corresponding SP task to done. Refs #1609, #1524 (closed-stale reports of the same behavior on the GitHub and GitLab providers).

2. **Imported issues land in Today even when going to Backlog.** `addTaskFromIssue` defaults every newly-imported task's `dueDay` to today. A task placed in a project's Backlog therefore also matches "due today" and shows up in the Today tab, defeating the purpose of the Backlog for users who auto-import into it.

3. **Backlog tasks get yanked to Today on every poll.** `_checkAndHandleIssueAlreadyAdded` unconditionally calls `moveTaskToTodayList` whenever the existing task's `projectId === activeWorkContextId`. Combined with #1 (no `issueId`, so the dedup filter in `checkAndImportNewIssuesToBacklogForProject` always sees every issue as new), every poll while the user is in the project context promotes already-imported Backlog tasks into Today. Refs #4324, #3933 (closed-stale reports of the same symptom).

Reproduces against a self-hosted Forgejo v15 instance with the Gitea provider configured to auto-import into a project that has Backlog enabled.

## Solution

Three independent commits, each addressing one bug.

1. **`fix(issue/gitea): set issueId and isDone in getAddTaskData (related #1609)`** mirrors the GitLab provider:
   ```ts
   issueId: String(issue.id),
   isDone: issue.state === 'closed',
   ```
   `BaseIssueProviderService.getFreshDataForIssueTask` spreads `getAddTaskData` into `taskChanges`, so closed-state now propagates to existing tasks on subsequent polls (not just new imports). Setting `issueId` also makes the existing dedup filter (`allExistingIssueIds.includes(issue.id)`) actually work for Gitea.

2. **`fix(issue): skip default dueDay when adding issue to backlog`** in `addTaskFromIssue`, replaces the unconditional `dueDay: getDbDateStr()` with a conditional skip when `isAddToBacklog === true`. Provider-supplied `dueDay` and explicit `additional` overrides are still honored.

3. **`fix(issue): don't yank backlog tasks to Today on re-import (related #4324)`** in `_checkAndHandleIssueAlreadyAdded`, looks up the task's project and returns early if `project.backlogTaskIds.includes(res.task.id)`. The `FOUND_MOVE_FROM_BACKLOG` snackbar is also skipped (nothing to announce when nothing moved).

All three verified end-to-end against real Forgejo v15:
- New issue lands in project Backlog only, no Today entry.
- Subsequent polls leave the task in Backlog, no migration.
- Closing the issue in Forgejo flips the SP task to done on the next poll.

Full unit test suite passes on both configured timezones (8551 / 8551, UTC + LA).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/issue/issue.service.spec.ts
+++ b/src/app/features/issue/issue.service.spec.ts
@@ -400,4 +400,43 @@ describe('IssueService', () => {
       expect(taskData.notes).toBe('Provider-set notes');
     });
   });
+
+  describe('addTaskFromIssue - existing task already in project backlog', () => {
+    const githubIssue = { id: 'github-issue-123', title: 'GitHub Issue' };
+
+    beforeEach(() => {
+      Object.defineProperty(workContextServiceSpy, 'activeWorkContextId', {
+        get: () => 'project-1',
+      });
+    });
+
+    it('should show "already exists" snack with Go to Task action for backlog tasks', async () => {
+      const existingTask = createMockTask({
+        issueType: 'GITHUB',
+        projectId: 'project-1',
+      });
+      taskServiceSpy.checkForTaskWithIssueEverywhere.and.resolveTo({
+        task: existingTask,
+        subTasks: null,
+        isFromArchive: false,
+      });
+      projectServiceSpy.getByIdOnce$.and.returnValue(
+        of({ backlogTaskIds: [existingTask.id] } as any),
+      );
+
+      await service.addTaskFromIssue({
+        issueDataReduced: githubIssue as any,
+        issueProviderId: 'github-provider-1',
+        issueProviderKey: 'GITHUB',
+      });
+
+      expect(snackServiceSpy.open).toHaveBeenCalledWith(
+        jasmine.objectContaining({
+          msg: T.F.TASK.S.TASK_ALREADY_EXISTS,
+          actionStr: T.F.TASK.S.GO_TO_TASK,
+          actionFn: jasmine.any(Function),
+        }),
+      );
+    });
+  });
 });
--- a/src/app/features/issue/issue.service.spec.ts
+++ b/src/app/features/issue/issue.service.spec.ts
@@ -401,6 +401,56 @@ describe('IssueService', () => {
     });
   });
 
+  describe('addTaskFromIssue - isAddToBacklog skips default dueDay', () => {
+    const jiraIssue = { id: 'JIRA-1', title: 'Test Jira Issue' };
+
+    const setupForNewTask = (): void => {
+      taskServiceSpy.checkForTaskWithIssueEverywhere.and.resolveTo(null);
+      taskServiceSpy.add.and.returnValue('new-task-id');
+      (service.ISSUE_SERVICE_MAP['JIRA'] as any).getAddTaskData = () => ({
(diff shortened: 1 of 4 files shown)


johannesjo:
Thank you very much!

johannesjo:
Some Feedback:



### Potential regression: `issueId: String(issue.id)` in `getAddTaskData`

Gitea uses `issue.number` (the per-repo index), not `issue.id` (the global DB id), for URL lookups. The search/backlog paths normalize this via `mapGiteaIssueIdToIssueNumber` in `gitea-api.service.ts` — and the historical commit [`2ed061ed10` "Change taskId to map issue number instead issue id on gitea"](https://github.com/super-productivity/super-productivity/commit/2ed061ed10) exists for exactly this reason. So tasks are stored with `issueId = issue.number`.

However, `getById$` (the polling path) does **not** apply that remap. On the first poll where `_wasUpdated` is true — including the close-detection scenario this PR targets — `BaseIssueProviderService.getFreshDataForIssueTask` spreads `getAddTaskData(issue)` into `taskChanges`. With this PR, that now includes `issueId: String(issue.id)`, which overwrites the stored `issue.number` with the DB id.

The very next poll then calls `/repos/{owner}/{repo}/issues/{DB_id}`, which will 404 (or, exceedingly rarely, return a wrong issue). So the closed→done transition you're targeting will succeed once, but subsequent polls on that task are broken.

**Suggested fix** — use the number directly:

```ts
issueId: String(issue.number),
```

Alternatively, apply `mapGiteaIssueIdToIssueNumber` to the `getById$` response so the whole provider sees a consistent id. Either works; the former is the smaller change.

### Minor

- The `isInBacklog` early-return in `_checkAndHandleIssueAlreadyAdded` shows no snackbar. That's correct for the polling path, but `addTaskFromIssue` is also called from user-initiated add-from-search flows — manually picking an already-backlogged issue there will now silently no-op. Consider a small "already in backlog" snack mirroring the existing `TASK_ALREADY_EXISTS` / `FOUND_MOVE_FROM_BACKLOG` patterns.
- Fixes 2 and 3 are eminently unit-testable (mock `projectService.getByIdOnce$`, assert `moveTaskToTodayList` not called; assert `dueDay` absent when `isAddToBacklog=true`). Not blocking, but worth adding to prevent re-regression.

Fixes 2 and 3 look good to merge as-is. Once fix 1 uses `issue.number`, I think this is ready.

sespiros:
Thank you for the feedback and the fast response. I didn't spot that id/number inconsistency in my manual testing. I pushed the fix you requested as well as fixes for your minor suggestions.

PR #7296, opened 2026-04-20
chore(debug): add opt-in search navigation traces

aakhter:
## Summary
- add an opt-in search navigation trace helper behind localStorage flag SP_SEARCH_NAV_DEBUG
- keep trace history in memory on window.__spSearchNavDebug instead of writing debug files
- instrument the navigation, route-change, and work-view focus path for future diagnosis of rendering and scroll timing issues

## Notes
- disabled by default
- no disk writes
- intended only for debugging search-navigation issues


johannesjo:
Thank you very much!

PR #7297, opened 2026-04-20
fix(electron): harden simple store writes

aakhter:
## Summary
- serialize simple-store writes so concurrent saves do not clobber each other
- write through a temp file and rename into place for safer persistence
- quarantine corrupt JSON files instead of repeatedly failing to parse them
- add a Node regression test covering corrupt-file recovery, legacy-directory replacement, and queued saves

## Local verification
- node electron/simple-store.test.cjs


johannesjo:
Thank you very much!

PR #7302, opened 2026-04-21
feat(plugins): emit taskUpdate hook for project moves

aakhter:
## Summary
- dispatch the TASK_UPDATE plugin hook when a task is moved to another project
- include the new projectId in the emitted change payload
- update the plugin hook fixture/spec coverage for the current task hook action set

## Local verification
- export CHROME_BIN="/home/aakhter/.cache/ms-playwright/chromium-1208/chrome-linux64/chrome"
- npm run test:file -- src/app/plugins/plugin-hooks.effects.spec.ts --browsers ChromeHeadless
- result: 9 SUCCESS


johannesjo:
Thank you very much!

PR #7304, opened 2026-04-21
fix(github-plugin): enable github advanced search to make queries including "OR" or "AND" logic operators work

### LINKED ISSUE #4913: 🚨 Github search doesnt encode "(" and ")" properly
### Version Used

v14.2.6 flatpak

### Expected Behavior

When using a search query like "sort:updated state:open (author:@me OR assignee:@me)" it should URI encode the parentheses properly and therefor lead to a successful API query

### Current Behavior

Instead the parentheses are just passed in directly causing a HTTP 422 error. Example URL from error box: https://api.github.com/search/issues?q=sort%3Aupdated%20state%3Aopen%20(author%3A%40me%20OR%20assignee%3A%40me)+repo:NixOS/nixpkgs

The search query should come out as "sort%3Aupdated%20state%3Aopen%20%28author%3A%40me%20OR%20assignee%3A%40me%29" instead.

Using this search term works just fine when using the github.com search box. So it is a valid search.

### Steps to Reproduce

1. Add a Github source for tasks
2. Use the search box for that Github source and type in "sort:updated state:open (author:@me OR assignee:@me)"
3. Witness the error caused by incomplete encoding

### Console Output

```shell

```

ToasterUwU:
## Problem

Using advanced search queries with logic operators like "OR" are supported by Github, but dont currently work in Super Productivity's Github Issue Search
This fixes #4913

## Solution

Enabling advanced_search fixes this, with no side effects from what i could gather.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/packages/plugin-dev/github-issue-provider/src/plugin.ts
+++ b/packages/plugin-dev/github-issue-provider/src/plugin.ts
@@ -170,7 +170,7 @@ PluginAPI.registerIssueProvider({
       searchTerm.includes('is:issue') || searchTerm.includes('is:pull-request');
     const typeFilter = hasTypeFilter ? '' : ' is:issue';
     const q = encodeGithubQuery(`repo:${owner}/${repo}${typeFilter} ${searchTerm}`);
-    const url = `${API_BASE}/search/issues?q=${q}&per_page=50`;
+    const url = `${API_BASE}/search/issues?q=${q}&per_page=50&advanced_search=true`;
     const response = await http.get<GithubSearchResponse>(url);
     return (response.items || []).map(mapSearchResult);
   },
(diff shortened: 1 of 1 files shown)


ToasterUwU:
Like i said, i read the docs and i couldnt really find anything that would change by enabling this. 
From what i can tell it used to work differently a while back, and enabling advanced_search would change the behavior a bit, but from what i can tell the difference isnt a  thing anymore. Everything is a "AND" unless using "OR" specifically.

If you can find something that suggests there is a difference in behavior im happy to implement a check for "is user using advanced search query only operators"

johannesjo:
Thank you very much for this! 

This is the right flag and matches shedokan's finding in #4913.

One thing to consider: `getNewIssuesForBacklog` at [plugin.ts:273](https://github.com/super-productivity/super-productivity/blob/github-advanced-search/packages/plugin-dev/github-issue-provider/src/plugin.ts#L273) hits the same `/search/issues` endpoint with a user-configurable `backlogQuery`. The default (`sort:updated state:open assignee:@me`) works fine, but a user who puts the example query from #4913 — `sort:updated state:open (author:@me OR assignee:@me)` — into the backlog config will still hit 422 without the flag.

Could you either append `&advanced_search=true` there too, or extract a tiny helper like `buildSearchIssuesUrl(q)` so the two call sites can't drift apart?

ToasterUwU:
I included the extra flag in the other spot as well, thanks for pointing that out!

johannesjo:
Thank you very much! <3

PR #7305, opened 2026-04-21
feat(plugins): add reInitData API

aakhter:
## Summary
- add a `reInitData()` method to the public plugin API surface
- wire the call through the iframe bridge into `DataInitService.reInit()`
- add a focused `PluginAPI` delegation spec for the new method

## Testing
- `bash -lc 'export CHROME_BIN="$HOME/.cache/ms-playwright/chromium-1208/chrome-linux64/chrome"; npm run test:file -- src/app/plugins/plugin-api.spec.ts --browsers ChromeHeadless'`
  - `TOTAL: 2 SUCCESS`

### DIFF
--- a/packages/plugin-api/src/types.ts
+++ b/packages/plugin-api/src/types.ts
@@ -395,6 +395,7 @@ export interface PluginAPI {
 
   getCurrentContextTasks(): Promise<Task[]>;
 
+  reInitData(): Promise<void>;
   updateTask(taskId: string, updates: Partial<Task>): Promise<void>;
 
   addTask(taskData: PluginCreateTaskData): Promise<string>;
--- a/src/app/plugins/plugin-api.spec.ts
+++ b/src/app/plugins/plugin-api.spec.ts
@@ -4,6 +4,7 @@ import { PluginBaseCfg } from './plugin-api.model';
 describe('PluginAPI', () => {
   let pluginAPI: PluginAPI;
   let showIndexHtmlAsViewSpy: jasmine.Spy;
+  let reInitDataSpy: jasmine.Spy;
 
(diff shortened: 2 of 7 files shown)


johannesjo:
Thank you! Can you please check the failing unit tests?

johannesjo:
Thank you @aakhter can you explain why we might need this?

aakhter:
The use case I had in mind was plugins that do out-of-band import/repair work and then need SP to fully rehydrate from persisted state.

The existing plugin API works for normal task/project updates, but it doesn’t provide a way to trigger a full reload afterward. `reInitData()` was meant as an escape hatch for that case by delegating to the existing app reinitialization path.

That said, if plugins are expected to stay strictly within the current update model and not trigger full app reinitialization, then this is probably too special-purpose for the public plugin API, and I’m fine dropping the PR.

johannesjo:
Thanks for the explanation! I think in order to honor YAGNI we should wait for a concrete use case to pop up before merging this. Would this be alright with you?

aakhter:
This came out of a real need in a plugin I’ve been developing.

That plugin performs some work out of band and updates persisted data directly. After that, the normal plugin update flow was not enough to get Super Productivity to fully reload and rehydrate from the new persisted state, so I added `reInitData()` as a way to invoke the existing app reinitialization path from the plugin side.

My intent here was not to add speculative API surface, but to keep the Super Productivity changes required by that plugin in the public, maintained codebase rather than carrying a private patch.

If that does not fit the direction you want for the plugin API or the project more generally, I’m completely fine dropping it.

johannesjo:
In this case: Let's add it! Thank you very much!

PR #7321, opened 2026-04-22
feat(i18n): add Turkish sorting options and error messages

Cyber-Syntax:


johannesjo:
Cok tessekür ederim! <3

PR #7337, opened 2026-04-23
Add ext-idle-notify backend for Wayland idle detection

davidvornholt:
## Summary

This change adds a compositor-native Wayland idle detection backend for Linux builds and uses it before the existing `xprintidle` and `loginctl` fallbacks.

Specifically, it:

- adds a small Linux-only helper binary that subscribes to the standard `ext-idle-notify` Wayland protocol
- teaches the Electron idle handler to probe for that backend, start it when selected, and clean it up on shutdown
- builds the helper as part of `electron:build` and copies it into packaged Linux application outputs
- ignores generated helper binaries and local Rust build output so the repository only tracks source artifacts

## Root Cause

On Wayland, Super Productivity previously relied on three backends:

- GNOME Mutter D-Bus
- `xprintidle`
- `loginctl`

That worked poorly on compositors such as Niri, where the GNOME-specific path is unavailable, `xprintidle` is often absent, and `loginctl` may never report idle time for the session. In that configuration, idle detection remained stuck at `0ms`, so the idle dialog never appeared.

Using `ext-idle-notify` fixes this by asking the compositor directly for idle and resume notifications instead of relying on desktop-specific or systemd-specific fallbacks.

## Impact

- fixes idle detection on Niri and other Wayland compositors that implement `ext-idle-notify`
- keeps the existing GNOME, `xprintidle`, and `loginctl` paths as fallbacks
- does not add any dependency on ActivityWatch or compositor-specific external daemons

## Validation

- `npm run checkFile electron/idle-time-handler.ts`
- `npm run checkFile electron/start-app.ts`
- `npm run electron:build`
- `cargo check --manifest-path electron/wayland-idle-helper/Cargo.toml`
- manual verification on Fedora Niri: the helper connected successfully and the idle dialog started working once no idle-inhibiting app was active in the session


johannesjo:
Thanks for tackling this – the Niri/Wayland idle gap is real and using `ext-idle-notify-v1` is the right protocol choice. Reviewed with a mix of automated and manual passes. A few things to address before this can merge, plus a couple of nice-to-haves.

## Should change before merge

### 1. `npm run electron:build` now hard-fails on Linux without cargo
`tools/build-wayland-idle-helper.js` exits the whole build if `cargo` isn't installed. Every Linux contributor currently does *not* need Rust, so this is a silent regression — a first-time contributor running `electron:build` just gets `ENOENT cargo` with no pointer to what's wrong. `tools/afterPack.js` already has the right pattern: warn and skip if the binary is missing. Please mirror it:

- If `cargo` isn't on `PATH`, log a clear warning (e.g. `Rust toolchain not found — skipping Wayland idle helper. Install via https://rustup.rs if you need it`) and `return` instead of `process.exit(1)`.
- Optionally gate the build behind `BUILD_WAYLAND_IDLE_HELPER=1` or move it to a separate `npm run build:wayland-helper` script that `electron:build` calls best-effort.

Also please add `--locked` to the `cargo build` invocation so `Cargo.lock` is actually enforced on fresh checkouts / CI.

### 2. Probe mode duplicates the real startup path
Today: `_probeWaylandIdleHelper` spawns the helper with `--probe`, which does a full `init_wayland` (connects, binds `WlSeat` 1..=9, binds `ExtIdleNotifierV1`, *creates an idle notification*) and then exits — just to answer "does the protocol work?". Then moments later `_ensureWaylandIdleHelperStarted` spawns it again with `--timeout-ms` and repeats the exact same setup. That's two fork+execs and two full Wayland handshakes per cold start, plus a transient side-effect on the compositor.

More importantly, on some compositors `get_idle_notification` is async — the probe process may exit before the request is actually flushed, so the probe can report success for a setup that later fails in the long-lived helper.

Suggestion: drop `--probe` entirely. Make `test:` call `_ensureWaylandIdleHelperStarted()` directly and treat \"`ready` emitted within the timeout\" as the probe signal. This also collapses `Mode::Probe` in `main.rs`, the `execFile` import, and ~30 lines of TS state machine.

### 3. `_waylandIdleSinceMs = Date.now() - CONFIG.MIN_IDLE_TIME` is fragile
This is currently correct only because the helper is spawned with `--timeout-ms CONFIG.MIN_IDLE_TIME` — i.e. it's implicitly assuming the two constants are equal. Two concerns:
- If `CONFIG.MIN_IDLE_TIME` is ever changed without touching the spawn args (or vice versa), idle time is silently wrong.
- After a laptop suspend, `Date.now()` jumps forward but the frozen `_waylandIdleSinceMs` can yield gigantic values until the next `resumed`.

Fix: store the timeout you actually passed to the helper (e.g. `_waylandHelperTimeoutMs`) and use that for back-dating; or have the helper emit `idle <ms-since-epoch>` on stdout so the parent never needs to derive it. Consider resetting on `powerMonitor`'s `resume` event.

## Nice to have

- **Restart / fallback after repeated failures.** If the helper consistently crashes at runtime (e.g., compositor restart), `_workingMethod` stays pinned to `'waylandIdleNotify'` and every `getIdleTime()` re-spawns + fails. Consider an N consecutive-failure threshold → re-run `_determineWorkingMethod` so we gracefully fall back to `xprintidle`/`loginctl`.

- **Dispose hardening.** `dispose()` fires `child.kill()` and returns; if the helper is stuck in `blocking_dispatch`, quit can race. Worth adding a `SIGKILL` escalation after ~500ms and `child.unref()` at spawn time. `before-quit` is usually a more reliable hook than `will-quit` for child-process teardown.

- **Cargo release profile.** Since this binary ships in every Linux package, consider tuning \`Cargo.toml\`:

    \`\`\`toml
    [profile.release]
    lto = true
    codegen-units = 1
    strip = true
    panic = \"abort\"
    \`\`\`

    Typical 30–60% size reduction for simple helpers.

## Nits

- `electron/wayland-idle-helper/src/main.rs` `emit()` — `.expect(\"failed to write event\")` panics noisily on broken pipe during shutdown. Prefer `let _ = writeln\!(...)` + `let _ = lock.flush()`, or match on `ErrorKind::BrokenPipe` and `process::exit(0)`.
- `electron/idle-time-handler.ts` — helper `stderr` is logged at `warn`. Wayland client libs routinely emit benign diagnostics; users pasting logs into bug reports may also leak `\$XDG_RUNTIME_DIR` / `WAYLAND_DISPLAY` path fragments. `log.debug` would be quieter and friendlier to the project's privacy-first stance.
- `electron/idle-time-handler.ts` — `_waylandIdleHelperBuffer` grows unbounded on malformed input. `readline.createInterface({ input: child.stdout })` would replace it with a few lines and handle partial trailing lines cleanly.
- `electron/wayland-idle-helper/src/main.rs` — `_seat`/`_notifier`/`_notification` are always `Some(_)` after `init_wayland`; drop the `Option` wrapping. A one-line comment explaining \"kept alive for the lifetime of the connection\" would help future readers.
- `electron/wayland-idle-helper/src/main.rs` — probe hard-codes `60_000ms` for `get_idle_notification` separately from `CONFIG.MIN_IDLE_TIME`. Not user-visible today but worth a comment noting \"probe value is a don't-care\" (or just drop the notification create in probe mode per point #2).
- `electron/wayland-idle-helper/Cargo.toml` — consider adding a `rust-toolchain.toml` pinning stable, for reproducible packaging builds.
- Consider using electron-builder's `extraFiles` in config instead of the custom `installWaylandIdleHelper` block in `afterPack.js` — one declarative entry replaces the copy+stat+chmod code.

Thanks again — looking forward to seeing this land. Happy to re-review once the three \"should change\" items are addressed.

davidvornholt:
Thanks again for the detailed review! I pushed an update that addresses the three merge-blocking items:

- Made the Wayland helper build best-effort when `cargo` is missing, with a clear rustup warning, and added `--locked` to the cargo build.
- Dropped `--probe` entirely. The TypeScript detection path now starts the real long-lived helper and treats the `ready` line within the timeout as the capability check, so there is only one Wayland setup path.
- Removed the implicit `CONFIG.MIN_IDLE_TIME` coupling by storing the helper timeout passed to the child process and using that for idle backdating. The idle marker is also reset on Electron's resume event.

I also folded in several of the smaller suggestions while touching the same code: repeated helper failures now trigger idle-method re-detection, helper shutdown escalates to SIGKILL after 500ms and uses `unref()`, teardown moved to `before-quit`, helper stderr is logged at debug, stdout/stderr parsing now uses `readline`, Rust broken-pipe writes exit quietly, the Wayland handles no longer use unnecessary `Option` wrapping, and the Rust crate now has the suggested release profile plus a stable `rust-toolchain.toml`.

Validation run locally:

- `npm run checkFile electron/idle-time-handler.ts`
- `npm run checkFile electron/start-app.ts`
- `cargo fmt --check --manifest-path electron/wayland-idle-helper/Cargo.toml`
- `cargo check --locked --manifest-path electron/wayland-idle-helper/Cargo.toml`
- `npm run electron:build`
- simulated missing-cargo run of `tools/build-wayland-idle-helper.js`
- pre-push `npm run lint` and `npm test`

I left the optional `extraFiles` packaging refactor out for now because the current afterPack copy intentionally stays best-effort when the helper binary is absent, and I did not want to widen the packaging behavior in this review pass.

johannesjo:
@copilot resolve the merge conflicts in this pull request

johannesjo:
Thank you very much! <3

Issue #5109, opened 2025-09-14
🚨 Task List - Increase the visibility of Action-Icons for Tasks with Time-Data (Glass Theme)


OLLI-S:
### Version Used

15.0.0 

### Expected Behavior

When a task has time data (estimated time and spent time) and the mouse pointer is hovered over the task (to make the action icons visible), then these icons are displayed above the time data.
These action icons should be easily perceivable.

### Current Behavior

Due to the time data these action icons are worse perceivable.

**Task with Time Data**:
<img width="592" height="93" alt="Image" src="https://github.com/user-attachments/assets/6593394c-8b28-483c-9cc6-af32d624d6d5" />

**Mouse pointer on the Task**:
<img width="589" height="127" alt="Image" src="https://github.com/user-attachments/assets/b144505a-052c-4e30-ad17-168acc1ee3d4" />

### Steps to Reproduce

Go to a task that has time data.
Hover your move over the task.

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
That is for the glass theme right?

OLLI-S:
Yes, I have the "Glass" theme.

PR #7338, opened 2026-04-23
fix(tasks): keep hover controls visible on Glass theme

JulioMReis-tecnico:
## Problem

With the glass theme, hovering the mouse over a task creates an overlap of the time information and the action buttons (the buttons that appear when hovering the mouse over a task). #5109 

## Solution

The task's time-wrapper visibility is set to hidden when hovering the mouse over a task.
Tasks with sub-tasks are excluded, as they dont have action buttons.

## Test

A new test that verifies if time-wrapper is visible and becomes hidden when hovering was also implemented.
The test creates a simple task with estimated time of 45 minutes.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/e2e/tests/task-detail/task-time-visibility.spec.ts
+++ b/e2e/tests/task-detail/task-time-visibility.spec.ts
@@ -1,0 +1,33 @@
+import { Locator, Page } from 'playwright/test';
+
+import { expect, test } from '../../fixtures/test.fixture';
+import { WorkViewPage } from '../../pages/work-view.page';
+
+test.describe('Task time visibility', () => {
+
+  const addTaskAndFillEstimatedTime = async (
+    workViewPage: WorkViewPage,
+    page: Page,
+  ): Promise<void> => {
+    await workViewPage.waitForTaskList();
+
+    await workViewPage.addTask('task');
+    await page.getByText(/task/).first().hover();
+    await page.getByRole('button', { name: 'Show/Hide additional info' }).click();
+
+    const durationInput: Locator = page.locator('.mat-ripple.ripple').first();
+    await durationInput.click();
+    const timeSpentInput: Locator = page.getByText('Estimate');
+    await timeSpentInput.fill('45');
+    await page.getByRole('button', { name: 'Save' }).click();
+  };
+
+  test('Task time data should be invisible on hover', async ({ page, workViewPage }) => {
+    await addTaskAndFillEstimatedTime(workViewPage, page);
+
+    const timeWrapper = page.locator('.time-wrapper');
+    await expect(timeWrapper).toBeVisible({ timeout: 3000 });
+    await page.getByText(/task/).first().hover();
+    await expect(timeWrapper).not.toBeVisible({ timeout: 3000 });
+  });
+});
--- a/e2e/tests/task-detail/task-time-visibility.spec.ts
+++ b/e2e/tests/task-detail/task-time-visibility.spec.ts
@@ -1,33 +1,106 @@
-import { Locator, Page } from 'playwright/test';
+import { Locator, Page } from '@playwright/test';
 
+import { cssSelectors } from '../../constants/selectors';
 import { expect, test } from '../../fixtures/test.fixture';
 import { WorkViewPage } from '../../pages/work-view.page';
+import { TaskPage } from '../../pages';
+
+const { FIRST_TASK, SECOND_TASK, SUB_TASK, MAIN } = cssSelectors;
 
 test.describe('Task time visibility', () => {
-
-  const addTaskAndFillEstimatedTime = async (
-    workViewPage: WorkViewPage,
-    page: Page,
-  ): Promise<void> => {
-    await workViewPage.waitForTaskList();
-
-    await workViewPage.addTask('task');
-    await page.getByText(/task/).first().hover();
-    await page.getByRole('button', { name: 'Show/Hide additional info' }).click();
-
-    const durationInput: Locator = page.locator('.mat-ripple.ripple').first();
+  const fillEstimatedTime = async (page: Page, time: number): Promise<void> => {
+    // time-estimate/Duration position might need adjusting if detail-menu is changed
+    const durationInput: Locator = page.locator(
+      'task-detail-item:nth-child(3) > .input-item > .mat-ripple',
+    );
     await durationInput.click();
-    const timeSpentInput: Locator = page.getByText('Estimate');
+    const timeSpentInput: Locator = page.getByText('Estimate', { exact: true });
     await timeSpentInput.fill('45');
     await page.getByRole('button', { name: 'Save' }).click();
   };
 
-  test('Task time data should be invisible on hover', async ({ page, workViewPage }) => {
-    await addTaskAndFillEstimatedTime(workViewPage, page);
+  const addTaskAndFillEstimatedTime = async (
+    workViewPage: WorkViewPage,
+    page: Page,
+    taskPage: TaskPage,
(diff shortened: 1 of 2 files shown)


johannesjo:
Thanks for the PR and welcome\! The approach (hiding `.time-wrapper` on hover to let the hover controls breathe) is reasonable, but I have a few things to flag before this is ready to merge.

**🔴 Blocker — `IS_TOUCH_PRIMARY` is undefined in this template**

The new binding uses `IS_TOUCH_PRIMARY`, but that symbol isn't exposed on `TaskComponent` — only `isTouchActive` is (see `task.component.ts:265`). The build doesn't error because we don't have `strictTemplates` enabled, so Angular silently treats the identifier as `any` / `undefined`. That makes `\!IS_TOUCH_PRIMARY` always `true`, so the intended touch-device guard does nothing and the time info will hide on hover on touch devices as well (where mouse-hover is often emulated).

The rest of this template already uses the reactive helper on line 99 (`isFirstLineHover && \!isTouchActive()`), so matching that pattern is the cleanest fix:

```html
[style.visibility]="
  isFirstLineHover && \!isTouchActive() && \!t.subTasks?.length
    ? 'hidden'
    : 'visible'"
```

**🟡 Scope — this affects all themes, not just Glass**

Issue #5109 is specifically about the Glass theme (confirmed in the thread). Hiding time info on hover everywhere is a behavior change for users on other themes who weren't experiencing the overlap. Worth checking with @johannesjo whether this should be Glass-scoped — either by gating on the theme class or with a CSS-only solution in the Glass theme's SCSS (e.g. fading `.time-wrapper` opacity on `.first-line:hover`).

**🟡 E2E test — fragile selectors**

A few nits on `task-time-visibility.spec.ts`:

- `page.locator('.mat-ripple.ripple').first()` depends on Material internals and element order. A role-based or more specific selector for the estimate input would be sturdier.
- `page.getByText('Estimate')` resolves to the label, not the native input — `fill('45')` on that is likely to throw, so I suspect this test may not be exercising what it looks like it exercises. Worth running locally to confirm.
- Per `e2e/CLAUDE.md`, new tests should use the page-object fixtures (`taskPage`, `workViewPage`) and the assertion helpers rather than raw `page.getByText(/task/).first()`.
- The touch-device branch (the bug above) and the sub-tasks branch (`\!t.subTasks?.length` guard) aren't covered — a quick parent-task-hover case would protect against regressions of the guard.

**🟢 Small stuff**

- Title doesn't match the Angular commit format the repo uses — something like `fix(tasks): keep hover controls visible on Glass theme` would align with the convention in `CLAUDE.md`.
- The multi-line style binding's indentation looks off — running `npm run checkFile src/app/features/tasks/task/task.component.html` should sort it out (that checklist item is still unchecked).
- `import { Locator, Page } from 'playwright/test'` — newer tests here use `@playwright/test`.

Nice first contribution — the core idea is right and using `visibility: hidden` rather than `display: none` to avoid layout shift was the correct instinct. Fix the `IS_TOUCH_PRIMARY` reference and let's talk about theme scope, then this should be in good shape.

johannesjo:
Thanks for the quick turnaround\! The blocker is resolved and the test refactor to page-object fixtures is great. A few things still to address before this can land:

**🔴 Test bug — `fillEstimatedTime` ignores its `time` argument**

`fillEstimatedTime(page, time)` accepts the parameter but hardcodes `'45'`:

```ts
const fillEstimatedTime = async (page: Page, time: number): Promise<void> => {
  ...
  await timeSpentInput.fill('45'); // should be String(time)
};
```

The third test passes `30` for the subtask but it still ends up as 45m. Either remove the parameter or use it.

**🔴 `getSubTasks(task)` returns a multi-locator**

```ts
const subTask = taskPage.getSubTasks(task);
await taskPage.openTaskDetail(subTask);
```

`getSubTasks` resolves all subtasks of the parent, and `openTaskDetail` calls `.hover()` on it. This works only because the test creates exactly one subtask — Playwright strict mode will throw \"strict mode violation\" the moment a second subtask is added. Use `.first()`:

```ts
const subTask = taskPage.getSubTasks(task).first();
```

**🟡 Theme scope — still unresolved**

The behavior change still applies to every theme, not just Glass. #5109 is a Glass-specific bug, and hiding the time on hover everywhere is a UX change for users on other themes who weren't seeing the overlap. My preference would be to scope this to the Glass theme — either gate the `[style.visibility]` binding on the theme class, or solve it CSS-only inside the Glass theme's SCSS (e.g. fading `.time-wrapper` opacity on `.first-line:hover`). Happy to discuss if you'd rather keep it global.

**🟡 Touch-device branch still uncovered**

The new tests don't exercise the `\!isTouchActive()` guard — that's exactly the branch the original bug lived in. A `test.use({ hasTouch: true })` case asserting `.time-wrapper` stays visible on hover would lock this in.

**🟡 Fragile detail-item selector**

`task-detail-item:nth-child(3) > .input-item > .mat-ripple` still depends on the order of items in `task-detail-panel.component.html`. Reordering detail items would silently break this test. Prefer an icon-based selector matching the pattern in `e2e/constants/selectors.ts` (see `SCHEDULE_TASK_ITEM`):

```ts
'task-detail-item:has(mat-icon:text(\"hourglass_empty\")) .mat-ripple'
```

**🟢 Small stuff**

- `page.getByText('Estimate', { exact: true })` still resolves the label rather than the input — `fill()` may work via Material's label association but a direct input selector is sturdier.
- PR title \`Fix(tasks): …\` should be lowercase \`fix(tasks): …\` per the Angular convention used in this repo.
- \`npm run checkFile\` checklist item is still unchecked — running it on both files will catch the trailing commas in the destructures (\`{ page, workViewPage, taskPage, }\`) and the missing space after \`//\` in two comments.

The core idea is right and the test coverage is much improved — fix the two test bugs and let's settle the theme-scope question, then this should be in good shape.

JulioMReis-tecnico:
hello, sorry I couldnt work on this for a while.
I was returning to this and noticed that commit 51ff66c makes the time-wrapper hidden on hover, fixing the issue this PR was intended for.
I appreciate if it could be comfirmed that no more work is needed on this issue.
thanks

johannesjo:
Thanks for following up, and thanks again for the contribution and the heads-up — you're exactly right. 👍

Commit `51ff66c` landed a CSS-only fix for #5109 in `_task-controls.scss`:

```scss
// Hide the time-wrapper while hover-controls are mounted: they're absolutely
// positioned over .time-wrapper and would let its digits ghost through on
// translucent themes (velvet, glass) and paint redundantly on solid ones.
.first-line:has(task-hover-controls) .time-wrapper {
  visibility: hidden;
}
```

Since Angular only mounts `task-hover-controls` on hover (non-touch), the `:has()` selector matches the visibility intent without a `:hover`/touch guard — which is the theme-scoped, CSS-side approach I'd raised in the earlier reviews. So the overlap you originally reported is already resolved on `master`, and no further work is needed here.

One thing to flag in case it's ever revisited: because that rule lives in SCSS, layering this PR's inline `[style.visibility]` binding on top would actually take precedence for parent tasks (inline styles outrank the class rule) and re-show the time on hover — so the two solutions would now slightly conflict rather than reinforce.

I'm going to close this as resolved-on-master, but genuinely appreciate the solid first PR — the instinct to use `visibility: hidden` over `display: none` to avoid layout shift was spot on, and your responsiveness to feedback was great. Hope to see more from you. 🎉

PR #7356, opened 2026-04-24
fix(schedule): clamp schedule event titles

davidvornholt:
## What changed

This updates schedule event titles so they use the available event height instead of always truncating to a single line. Longer titles can now wrap across the space they have, while very short and continued split events still stay compact.

The title text is measured after render and on resize, then exposed as a CSS line clamp. The resize observer and animation frame are cleaned up when the component is destroyed.

## Why

Schedule events with enough vertical room could still hide useful title text. This makes dense schedule views friendlier without letting text spill out of small or overlapping events.

## Validation

- npm run checkFile src/app/features/schedule/schedule-event/schedule-event.component.ts
- npm run checkFile src/app/features/schedule/schedule-event/schedule-event.component.scss
- npm run test:file src/app/features/schedule/schedule/schedule.component.spec.ts
- pre-push hook: npm run lint
- pre-push hook: npm test

### DIFF
--- a/src/app/features/schedule/schedule-event/schedule-event.component.html
+++ b/src/app/features/schedule/schedule-event/schedule-event.component.html
@@ -62,27 +62,32 @@
 </div>
 <!-- -->
 
-<div class="title">
-  <!--  {{ se().data['plannedForDay'] }}-->
-  <!--  {{ scheduledClockStr }}-->
-  <!--  {{se().id}} __ {{se().data?.id}} __-->
-  <!-- -->
-
-  @if (
-    se().type === SVEType.SplitTaskContinuedLast ||
-    se().type === SVEType.SplitTaskContinued ||
-    se().type === SVEType.RepeatProjectionSplitContinued ||
-    se().type === SVEType.RepeatProjectionSplitContinuedLast
-  ) {
-    ...
+<div
+  #titleEl
+  class="title"
+>
+  <span class="title-text">
+    <!--  {{ se().data['plannedForDay'] }}-->
+    <!--  {{ scheduledClockStr }}-->
+    <!--  {{se().id}} __ {{se().data?.id}} __-->
     <!-- -->
-  } @else {
-    @if (titleHasLinks()) {
-      <span [innerHTML]="title() | renderLinks"></span>
+
+    @if (
+      se().type === SVEType.SplitTaskContinuedLast ||
+      se().type === SVEType.SplitTaskContinued ||
+      se().type === SVEType.RepeatProjectionSplitContinued ||
+      se().type === SVEType.RepeatProjectionSplitContinuedLast
+    ) {
+      ...
+      <!-- -->
     } @else {
-      {{ title() }}
+      @if (titleHasLinks()) {
+        <span [innerHTML]="title() | renderLinks"></span>
+      } @else {
+        {{ title() }}
+      }
     }
-  }
+  </span>
 </div>
 
 @if (task()) {
--- a/src/app/features/schedule/schedule-event/schedule-event.component.scss
+++ b/src/app/features/schedule/schedule-event/schedule-event.component.scss
@@ -281,21 +281,27 @@
 }
 
 .title {
+  --title-padding-y: 6px;
+  --title-line-height: 1.4em;
+
   font-size: 14px;
-  @include truncateText();
   flex: 1 1 auto;
-  padding: 6px;
+  min-width: 0;
+  padding: var(--title-padding-y) 6px;
+  box-sizing: border-box;
   font-weight: 400;
   line-height: 1.4;
   max-height: 100%;
+  overflow: hidden;
   text-align: left;
-  column-width: 250px;
 
   @include mq(xs, max) {
-    column-width: 120px;
+    --title-padding-y: var(--s-half);
+    --title-line-height: 1.3em;
+
     font-size: 11px;
     line-height: 1.3;
-    padding: var(--s-half);
+    padding: var(--title-padding-y);
   }
 
   @include mq(xxs, max) {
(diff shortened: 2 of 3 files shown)


johannesjo:
## Summary
Makes schedule event titles fill available vertical space instead of always truncating to one line. Wraps the title content in an inner `.title-text` span, moves clamping onto that span via `-webkit-line-clamp: var(--title-line-clamp, 1)`, and computes the clamp value at runtime via a per-component `ResizeObserver` + initial `requestAnimationFrame` measurement. Very-short and split-continued variants keep their single-line behaviour.

## Issues

### Important
- **`schedule-event.component.ts` — redundant observation of the title element invites a feedback cycle.** `availableTitleHeight` is derived solely from `hostEl.clientHeight` minus padding, so only the host's size affects the result. Observing `titleEl` in addition to `hostEl` fires an extra callback every time `_updateTitleLineClamp` updates the clamp (because the inner text height then changes). The loop stabilizes (same inputs → same output), but each clamp change pays for an extra observer tick, a second forced `getComputedStyle`, and a signal write-that-no-ops. Recommend: `this._resizeObserver.observe(hostEl)` only.
- **All work runs inside Angular's zone.** With potentially dozens of events visible, each window/grid resize fires N callbacks, each doing `getComputedStyle` + a signal `set`, each capable of triggering OnPush change detection. Wrapping observer setup in `NgZone.runOutsideAngular(...)` and re-entering the zone only when the clamp actually changes (inside the `if (... \!== this._titleLineClamp())` branch) would avoid unnecessary CD cycles during window resize / drag.
- **`getComputedStyle` on every callback forces style recalc.** Combined with the per-event observer, a fast window resize across a dense day view can thrash layout. At minimum, coalesce multiple RO entries per frame via a single `requestAnimationFrame`.

### Minor
- **`--title-line-clamp` is always emitted on the host style string, even for `very-short-event`** where `.title-text { display: block; @include truncateText(); }` ignores it. Not wrong, just dead work for the shortest (and most numerous) events.
- **Adding `--title-line-clamp` by string concatenation to the host `style()` computed** works, but `'[style.--title-line-clamp]': '_titleLineClamp()'` in the `host: {}` block would leave the positional style untouched and avoid rewriting the full style string on each clamp change.
- **`parseFloat(styles.lineHeight)` returns `NaN` if `line-height` ever resolves to `"normal"`.** Not possible today because `.title` declares `line-height: 1.4`; the `Number.isFinite(lineClamp)` guard already neutralizes it. A one-line comment clarifying the guard's purpose would help future readers.
- **Font-loading timing:** initial measurement runs in the next rAF. If custom fonts finish loading after that, the line-height in px will be slightly off, causing a one-line over/undershoot until the next resize. A `document.fonts.ready.then(() => this._updateTitleLineClamp())` in `ngAfterViewInit` would harden it.
- **No unit test for the new behaviour.** The referenced `schedule.component.spec.ts` doesn't cover this component. Given the math is simple and behavioural, an e2e check (title wraps in tall slots, clamps in short slots) would catch regressions better than a unit test.

## Strengths
- Clean teardown: both the `ResizeObserver` and the pending rAF are cancelled in `ngOnDestroy`.
- Good instinct on the Chromium "padding on a line-clamped box reveals the next hidden line" quirk — the inline SCSS comment documents the *why* nicely.
- Single source of truth via CSS variables (`--title-line-clamp`, `--title-line-height`, `--title-padding-y`) keeps the responsive overrides readable.
- Split-continued variants keep `line-clamp: 1` explicitly, so the `...` ellipsis events aren't affected.
- `Math.max(1, Math.floor(...))` + `Number.isFinite` guards prevent degenerate clamp values.

## Verdict
Correctness is fine and CI is green, but the per-event `ResizeObserver` running inside the zone (plus the redundant `titleEl` observation) is a real scalability concern on a dense schedule. Addressing the Important points above is small work and will make this safe to merge without a follow-up perf regression.

johannesjo:
## Follow-up: concrete perf concern for dense views

Flagging this separately because I think it's stronger than "minor" in practice — especially for the **week view**, which renders all 7 days of `schedule-event` components at once. A busy user can easily have 100–300 of these on screen simultaneously.

Every one of them creates a `ResizeObserver` watching **two** elements (host + title).

### What triggers the cost
1. **Window resize** — all N observers fire in parallel. Each callback does `getComputedStyle()` twice and writes a signal. `getComputedStyle` forces style resolution if anything is dirty, which cascades during a resize burst.
2. **Drag-to-resize an event** — the grid relayouts continuously; observers fire every frame during the drag.
3. **Zone pollution** — the observer isn't wrapped in `NgZone.runOutsideAngular`, so every callback schedules an Angular CD tick, even when the clamp value doesn't actually change. With OnPush that's bounded, but the zone still wakes up N times per resize frame.
4. **Feedback tick from observing `titleEl`** — each real clamp change triggers one extra no-op callback because the title's content height changed as a result of the update.

### Rough order of magnitude (week view, ~150 events, during a window drag)
- ~150 RO callbacks per resize batch
- ~300 `getComputedStyle` calls per batch (host + title)
- ~150 zone entries → ~150 CD scheduling attempts per batch

That's the kind of thing that turns a smooth window-resize into a visibly laggy one, and compounds with the schedule's existing layout cost.

### Mitigation (cheap, high-impact)
The three "Important" items from the earlier review cut this down substantially:
1. **Observe `hostEl` only**, drop `titleEl` — eliminates the feedback tick and halves `getComputedStyle` calls.
2. **Wrap the `ResizeObserver` setup in `NgZone.runOutsideAngular(...)`**, and re-enter the zone only inside the branch that actually writes the signal:
   ```ts
   this._ngZone.runOutsideAngular(() => {
     this._resizeObserver = new ResizeObserver(() => this._scheduleMeasure());
     this._resizeObserver.observe(hostEl);
   });
   ```
3. **Coalesce observer callbacks via a single `requestAnimationFrame`** so multiple RO entries in one frame only do one measurement.

Together those drop ~10× of the per-batch work for dense views, and the correctness of the feature is unchanged.

### Alternative (probably not worth it)
A single `ResizeObserver` on the grid container propagating a CSS variable via computed slot-height is doable, but event heights vary (split-continued, very-short, hour-spanning), so per-event measurement is the pragmatic choice. Stick with per-event *after* the above fixes — don't rearchitect.

Happy to open a follow-up PR with these three fixes once this lands if that's easier than iterating here.

davidvornholt:
Thanks for the thoughtful review! I agree the feature behavior was okay, but the dense week-view performance concern was valid.

I updated the PR to keep the same visible behavior while cutting down the observer work:

- the `ResizeObserver` now watches only the host element, since the clamp calculation only depends on host height
- observer scheduling runs outside Angular's zone
- resize callbacks are coalesced through a single `requestAnimationFrame`
- Angular is re-entered only when the computed clamp value actually changes
- `--title-line-clamp` now uses a dedicated host style binding instead of being appended to the larger positional style string

I left the smaller optional items alone for now. The current CSS already defines a numeric line height and guards non-finite clamp values, the very-short-event clamp value is harmless dead work, and the font-loading case seems edge-casey enough that I would rather not add extra lifecycle code unless it shows up in practice.

Validation passed again after the update:

- `npm run checkFile src/app/features/schedule/schedule-event/schedule-event.component.ts`
- `npm run test:file src/app/features/schedule/schedule/schedule.component.spec.ts`
- pre-push full lint and test suite, including the timezone run

davidvornholt:
Hi! Just checking in on this when you have a moment. I addressed the perf concerns from the review, CI is green, and I’d love to get this merged since the schedule title wrapping is pretty useful in daily use. Happy to adjust anything else if needed.

johannesjo:
Thank you very much! <3

davidvornholt:
Thank you very much for the merge!

Out of curiosity: which AI code review tool are you using for these reviews?

johannesjo:
> Out of curiosity: which AI code review tool are you using for these reviews?

I use claude code locally, but  I try to at least skimread the recommendations :)

PR #7359, opened 2026-04-24
feat(gitea): add label-based include and exclude filters

sespiros:
## Problem

The Gitea/Forgejo provider only exposes a `scope` filter (all / created-by-me / assigned-to-me).
There is no way to route different subsets of one repository's issues into different Super Productivity projects, which matters in particular for Forgejo's scoped labels (designed precisely to partition issues within a single repo).

## Solution

Add two optional fields to the Gitea config: `filterLabels` (sent to the issues endpoint as the `labels` query param, AND-matched server-side) and `excludeLabels` (applied client-side, since Gitea/Forgejo has no negation syntax on the issues endpoint).
Both support scoped labels.

Filters apply only to ingestion (search and auto-import).
Already-imported tasks keep refreshing regardless of label changes, matching how `scope` already behaves.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thank you very much for the PR! 


One concern after digging into Gitea's actual API behavior — the two ingestion paths use different endpoints with different label-matching semantics, which I think contradicts the "AND-matched server-side" claim:

| Code path | Gitea endpoint | Internal field | Actual semantics |
|---|---|---|---|
| `getLast100IssuesFor$` (auto-import) | `/repos/{owner}/{repo}/issues` (`ListIssues`) | `IncludedLabelIDs` | **AND** |
| `searchIssueForRepo$` (manual search) | `/repos/issues/search` (`SearchIssues`) | `IncludedAnyLabelIDs` | **OR** |

Both handlers parse the same `labels=foo,bar` query param but assign it to different fields. The result:

1. **Same config, different behavior.** With `filterLabels = "bug,wontfix"`, auto-import only ingests issues that have **both** labels, but typing into the manual issue-search box returns issues that have **either** label. Hard to diagnose from a user's perspective.
2. **Auto-import's AND behavior is actually a Gitea bug** that may flip at any time. Tracked at [go-gitea/gitea#33509](https://github.com/go-gitea/gitea/issues/33509) ("ListIssues API labels parameter does AND instead of OR") — still open, with fix PR [#36816](https://github.com/go-gitea/gitea/pull/36816) also still open. The Gitea swagger documents `labels` as "any of this labels" (OR) on both endpoints; once they fix ListIssues to match the docs, the auto-import path will silently switch from AND to OR.
3. **The wiki and `en.json` description currently mislead users** for the search path — they assert AND-match while OR-match is what actually happens.

### Suggested approach

Make `filterLabels` AND-match on the client too, the way `excludeLabels` already does. Concretely:

- Keep sending `labels=` to the server as a coarse pre-filter (it's always a superset of the correct answer whether the server does AND or OR — never wrong, just potentially loose).
- Add a sibling to `isIssueIncludedByLabels` — say `hasAllLabels(issue, includedLabelNames)` — and apply it in the same `.map(...)` chains in `gitea-api.service.ts`.

That makes the behavior:

- Consistent across both endpoints
- Independent of Gitea version / future upstream bug fixes
- Truly "AND match" as the user-facing copy promises

The only cost is that on the search path the 100-row page may include some pre-cull noise, which is the same trade-off `excludeLabels` already makes.

References:
- [go-gitea/gitea#33509 — ListIssues API labels parameter does AND instead of OR](https://github.com/go-gitea/gitea/issues/33509)
- [go-gitea/gitea#36816 — pending fix PR](https://github.com/go-gitea/gitea/pull/36816)

Everything else in the PR looks good to me — clean separation of helpers, good test coverage, intentionally not filtering `getById$` to match `scope` behavior. Just this semantics asymmetry to nail down before merge.

sespiros:
Thanks for reviewing! I implemented the change and verified manually that the logic is indeed AND. Let me know if you find any other inconsistency.

johannesjo:
Thank you very much! <3

Issue #4, opened 2017-02-10
Allow http (not https) JIRA hosts


jefft:
When configuring JIRA integration, one is currently prompted for 'Host'. SP is expecting something like 'localhost' or 'jira.mycompany.com'. In my case I enter 'redradish_jira.localhost'.

The problem is that SP then *assumes* https. My JIRA happens to run on http://redradish_jira.localhost inside the network, not https. If I enter Host: 'redradish_jira.localhost:80', then SP just tries to talk SSL on port 80 and breaks.

I think 'Host' should be replaced by 'Base URL', a well-known concept in JIRA-land. This is unambiguous, setting the scheme, port and URL prefix all at once.

johannesjo:
Thanks! Good suggestion! I will implement that.

johannesjo:
I'm assuming that this should work now. Let me know if it doesn't.

Issue #3, opened 2017-02-08
wont start in windows env


Code-Slave:
npm start doesnt work in windows dos prompt. If you remove NODE_ENV=DEV from package.json it loads fine

johannesjo:
Thanks for reporting! I'll add a note to the readme.

PR #7370, opened 2026-04-26
feat:Add a keyboard shortcut to collapse/expand groups

### LINKED ISSUE #7313: 💡 Add a keyboard shortcut to collapse/expand groups
### Problem Statement

Currently, Super Productivity allows users to group tasks. However, when multiple groups are expanded, there is no efficient way to collapse or expand all groups at once. Users must manually click on each group header individually to toggle their state.

This creates friction in workflows where users need to:
- Quickly get an overview by collapsing all groups
- Expand all groups to see the full task list
- Switch between different levels of detail without repetitive clicking

The lack of a bulk collapse/expand feature reduces productivity, especially when working with many groups or frequently toggling between overview and detailed views.

### Possible Solution

Add keyboard shortcuts and/or UI controls to collapse or expand all groups simultaneously across any view where grouping is available (Today, projects, tags, etc.).

_No response_

### Additional context

_No response_

### documentation gap

Hetsavani:
## Problem

There is no efficient way to expand or collapse all task groups at once. Users must manually toggle each group, which is inefficient when working with many groups.

Fixes: #7313 
## Solution

Added keyboard shortcuts to expand and collapse all task groups simultaneously. Also added support for toggling individual groups using arrow keys, improving navigation and usability.

(Border color is not getting recorded correctly in the video)
Video:
[Screencast from 2026-04-26 13-52-14.webm](https://github.com/user-attachments/assets/a0a4ce9e-e357-440e-93d9-44ed6a64a5a7)

## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/docs/wiki/3.03-Keyboard-Shortcuts.md
+++ b/docs/wiki/3.03-Keyboard-Shortcuts.md
@@ -68,6 +68,18 @@ The following shortcuts apply for the currently selected task (selected via tab 
 - `Arrow keys`: Navigate around task list
 - `ArrowRight`: Open additional info panel for currently focused task
 
+## Collapsible Groups
+
+The following shortcuts apply when focus is on a collapsible group header (e.g., task groupings in Work View like Overdue, Today, Done, Recurring, etc.).
+
+- `ArrowLeft`: Collapse the focused group
+- `ArrowRight`: Expand the focused group
+- `Shift`+`ArrowLeft`: Collapse all groups
+- `Shift`+`ArrowRight`: Expand all groups
+- `Enter` or `Space`: Toggle the focused group
+
+These shortcuts are **unconfigurable** and always available on group headers.
+
 ## Platform-Specific Differences
 
 **Electron/desktop only:**
--- a/src/app/features/tasks/task-shortcut.service.ts
+++ b/src/app/features/tasks/task-shortcut.service.ts
@@ -207,7 +207,7 @@ export class TaskShortcutService {
         checkKeyCombo(ev, keys.selectPreviousTask))
     ) {
       ev.preventDefault();
-      this._handleTaskShortcut(focusedTaskId, 'focusPrevious');
+      this._handleTaskShortcut(focusedTaskId, 'handleArrowUp');
       return true;
     }
 
@@ -217,7 +217,7 @@ export class TaskShortcutService {
         checkKeyCombo(ev, keys.selectNextTask))
     ) {
       ev.preventDefault();
-      this._handleTaskShortcut(focusedTaskId, 'focusNext');
+      this._handleTaskShortcut(focusedTaskId, 'handleArrowDown');
       return true;
     }
 
--- a/src/app/features/tasks/task/task.component.ts
+++ b/src/app/features/tasks/task/task.component.ts
@@ -544,6 +544,42 @@ export class TaskComponent implements OnDestroy, AfterViewInit {
     setTimeout(() => this.focusSelf(), 10);
   }
 
+  handleArrowUp(): void {
+    if (this._focusGroupHeaderIfAtBoundary('prev')) {
+      return;
+    }
+    this.focusPrevious();
+  }
+
+  handleArrowDown(): void {
+    if (this._focusGroupHeaderIfAtBoundary('next')) {
+      return;
+    }
+    this.focusNext();
+  }
+
+  private _focusGroupHeaderIfAtBoundary(direction: 'prev' | 'next'): boolean {
+    const myEl = this._elementRef.nativeElement as HTMLElement;
+    const myGroup = myEl.closest('collapsible.is-group') as HTMLElement | null;
+    if (!myGroup) {
+      return false;
+    }
+    const allTasks = Array.from(document.querySelectorAll('task')) as HTMLElement[];
+    const myIdx = allTasks.indexOf(myEl);
+    const adjacent = direction === 'prev' ? allTasks[myIdx - 1] : allTasks[myIdx + 1];
+    const adjacentGroup = adjacent?.closest('collapsible.is-group') ?? null;
+    if (adjacentGroup === myGroup) {
+      return false;
+    }
+    const targetGroup = direction === 'prev' ? myGroup : adjacentGroup;
+    const header = targetGroup?.querySelector('.collapsible-header');
+    if (header instanceof HTMLElement) {
+      header.focus();
+      return true;
+    }
+    return false;
+  }
+
   handleArrowLeft(): void {
     const t = this.task();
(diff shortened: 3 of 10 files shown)


johannesjo:
Thank you.

This looks confusing to me:

Screenshot_20260426_132740

Don't think we need the border.

johannesjo:
Thanks for tackling this @Hetsavani\! The feature is useful, but the implementation makes too many changes globally to `CollapsibleComponent` — which is used in ~7 other places (config sections, sync-conflict dialog, plugin manager, worklog export, repeat-task dialog, etc.). All of them inherit the new behavior and styling unintentionally. A few things to address:

### 1. Site-wide CSS regression — `collapsible.component.scss:2-12,21`
The new `:host { border, border-radius, transition, :focus-within { … } }` and `.collapsible-header { padding: calc(var(--s) / 2) var(--s) }` are applied unconditionally. Every collapsible across the app now has extra padding and a focus-within highlight, which will visibly shift layouts in dialogs and config pages.

**Fix:** scope these styles with `:host(.is-group)` and bind the class via `@HostBinding('class.is-group') @Input() isGroup = false`.

### 2. Tabindex / ARIA applied to all collapsibles — `collapsible.component.html:5-7`
`tabindex="0"`, `role="button"`, and `[attr.aria-expanded]` are added unconditionally. This adds a tab stop on every collapsible header throughout the app, changing tab order in forms and dialogs.

**Fix:** gate them on `isGroup`:
```html
[attr.tabindex]="isGroup ? 0 : null"
[attr.role]="isGroup ? 'button' : null"
[attr.aria-expanded]="isGroup ? isExpanded : null"
```

### 3. Plain Arrow keys toggle every collapsible — `collapsible.component.ts:60-83`
The `else` branch in `onHeaderKeydown` runs for **any** collapsible, not just groups, so pressing ArrowLeft/Right on a focused header in (say) the sync-conflict dialog now toggles it. Also, `preventDefault()` and `stopPropagation()` always fire, swallowing arrow keys from upstream handlers.

**Fix:** early-return when `\!this.isGroup`, and only call `preventDefault`/`stopPropagation` when the handler actually does something.

### 4. Missing Enter/Space activation — `collapsible.component.html:5`
With `role="button"`, screen readers tell users to press Enter or Space. Neither is handled — `(click)` doesn't auto-fire from those keys on a `<div>`.

**Fix:** in `onHeaderKeydown`, handle `Enter` and `' '` (Space) by calling `toggleExpand()` and `preventDefault()` (Space scrolls otherwise).

### 5. Static `_groupCollapsibles` Set — `collapsible.component.ts:21,86-94`
Module-level mutable state is brittle: if a future dialog ever uses `[isGroup]=\"true\"`, Shift+Arrow inside it will also toggle work-view groups in the background. The unit test also doesn't reset this between runs — a failed test mid-flight leaves stale instances.

**Suggested fix:** since \"all groups\" really means \"all sibling groups\", walk to the closest container and operate on its children (`document.querySelectorAll('collapsible.is-group')` scoped to a parent). At minimum, add `CollapsibleComponent['_groupCollapsibles'].clear()` in the spec's `beforeEach`.

### 6. Test gaps — `collapsible.component.spec.ts`
Worth adding:
- Enter and Space toggle the group (covers fix #4).
- Plain Arrow on a non-group `<collapsible>` does **not** toggle it (covers fix #3).
- Shift+Arrow on a non-group collapsible does not affect group collapsibles.

### Smaller things
- `[isGroup]` is set only on the date-grouped list (`work-view.component.html:129`) — Overdue, Done, Later Today, Recurring sections are excluded. Worth confirming with @johannesjo whether those should also participate in \"expand/collapse all groups\".
- The wiki keyboard-shortcuts page should be updated.
- `Shift+Arrow` for \"all\" is a fine choice but isn't a standard file-tree convention; consider also accepting `*` (asterisk) which matches Windows Explorer / VSCode tree behavior.

CI is green and the core logic works — once the changes above only apply to opted-in `isGroup` collapsibles, this is in great shape.

Hetsavani:
Thanks for the feedback. I will update it

Hetsavani:
Please review and let me know if there's further scope of improvement 😊.

johannesjo:
Thanks for the quick turnaround @Hetsavani — most of the prior feedback is in good shape. A few things still to address:

### Blocking

**1. A11y focus indicator regression — `collapsible.component.scss:28`**
Group headers now have `tabindex="0"` and `role="button"`, but `&:focus-visible { outline: none; }` strips the only visual indicator a keyboard user would see when they Tab to a group. Either:
```scss
// scope the reset to non-group headers
:host(:not(.is-group)) .collapsible-header:focus-visible { outline: none; }
```
or add a dedicated indicator on `:host(.is-group) .collapsible-header:focus-visible` (e.g. `outline: 2px solid var(--c-accent); outline-offset: -2px;` or a subtle `background-color`). Without one of these, the new keyboard-focusable groups have no focus ring.

**2. Dead commented-out CSS — `collapsible.component.scss:1-11`**
Please delete the commented `:host(.is-group) { border, transition, :focus-within … }` block rather than leaving it in. It will only confuse future readers.

### Nice to have in the same pass

**3. `aria-controls` is missing — `collapsible.component.html`**
With `role="button"` + `aria-expanded`, the panel should be linked. Suggest:
```html
<div ... [attr.aria-controls]="isGroup ? panelId : null">
@if (isExpanded) {
  <div [attr.id]="panelId" ...>
}
```
…with `panelId` generated once per instance (e.g. `crypto.randomUUID()` or a counter). Otherwise screen readers announce expanded/collapsed without knowing what's controlled.

**4. Static `_groupCollapsibles` Set — `collapsible.component.ts:21`**
I'm OK leaving the global registry for now since `[isGroup]="true"` only has one consumer, but please add a `// TODO:` noting that any second consumer will silently link them via Shift+Arrow — saves the next contributor a debugging session.

**5. Scope of "all groups" — `work-view.component.html:130`**
Still open from last round: should Overdue / Done / Later Today / Recurring also participate, or is the intent strictly date-groups? Happy with either answer, just want it confirmed before merge.

**6. Wiki keyboard-shortcuts page**
Please add the new bindings (`Arrow`, `Shift+Arrow`, `Enter/Space`).

Tests look great — the regression coverage I was worried about is solid. Once 1 and 2 are fixed (and ideally 3), I think this is good to go.

Hetsavani:
Thanks for the feedback!
I have changed things accordingly.
Coming to 5th point, yes all will participate.
And Sorry about the leftover commented CSS — I've removed it.

johannesjo:
Thank you very much! <3

PR #7376, opened 2026-04-26
feat(mentions): substring match in tag/project/date autocomplete

aakhter:
## Problem

The autocomplete popup that appears in the task title editor (triggered by `#`, `@`, or `+`) only filters items by **prefix**. So typing `aa` matches a tag named `aa-thing` but not `xx-aa-yy`. With nested or scoped tag naming conventions, this makes finding existing tags awkward.

## Solution

Switch the default `mentionFilter` in `MentionDirective` from `String.prototype.startsWith` to `String.prototype.includes`, so any item whose label contains the search string (case-insensitive) matches. This applies to all three default triggers — `#tags`, `@due` chrono dates, and `+projects`.

Two new unit tests in `mention.directive.spec.ts` cover the substring case for both `MentionItem` objects and plain string items. Existing tests remain green (the filter inputs in those tests happened to be prefixes, so they pass under either matcher).

## Type of Change

- [x] New feature

## Checklist

### DIFF
--- a/src/app/ui/mentions/mention.directive.spec.ts
+++ b/src/app/ui/mentions/mention.directive.spec.ts
@@ -247,7 +247,8 @@ describe('MentionDirective', () => {
 
       const result = mentionFilter('aa', items);
 
-      expect(result).toEqual([{ label: 'xxaaxx' }, { label: 'aabb' }, { label: 'bbaa' }]);
+      // Prefix match comes first; remaining mid-string matches preserve input order.
+      expect(result).toEqual([{ label: 'aabb' }, { label: 'xxaaxx' }, { label: 'bbaa' }]);
     });
 
     it('should match substrings for plain string items', () => {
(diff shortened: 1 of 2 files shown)


aakhter:
Local validation against the PR commit (= `upstream/master` + this commit), in case it's helpful for review:

**Lint / format**
- `npm run lint` (i.e. `ng lint`) — all files pass on the changed files
- `npm run checkFile` on both modified files — clean

**Unit tests**
- Targeted: `mention.directive.spec.ts` — 27/27 passing (includes the 2 new substring tests for both `MentionItem` objects and plain string items)
- Full Angular suite, Berlin TZ (`npm test`): 8643 SUCCESS, 41 skipped, 0 failures
- Full Angular suite, LA TZ (`npm run test:tz:la`): 8629 SUCCESS, 55 skipped, 0 failures

No behavioral changes besides the prefix → substring switch in the default `mentionFilter`. Existing prefix-matching tests pass unchanged because their inputs were already prefixes of the expected matches.

johannesjo:
Thank you very much! <3

Issue #1, opened 2017-02-07
JIRA integration 'Test Credentials' fails with 'Jira Request failed: searchJira - [object Object]'


jefft:
It appears that establishing JIRA integration is broken. SP doesn't even issue a HTTP request.

To illustrate, start netcat listening on localhost:12345

root@jturner-desktop:~# nc -l 12345

Now enable JIRA integration with host http://localhost:12345

![image](https://cloud.githubusercontent.com/assets/205995/22714324/177aba02-ede0-11e6-8c6d-98bb4eaf8792.png)

'nc' prints nothing, indicating that no HTTP request was received. SP reports an error:

![image](https://cloud.githubusercontent.com/assets/205995/22714377/4883882c-ede0-11e6-80f7-6ca4df82186d.png)

The console does not report anything useful:

![image](https://cloud.githubusercontent.com/assets/205995/22714427/8a06ca20-ede0-11e6-9649-330c6042f6b2.png)

johannesjo:
Thank you very much for reporting! I'll check this tomorrow!

johannesjo:
I'm having trouble finding the source, but my guess would be that it has to do with the port. node-jira which I'm using specifies this with an extra parameter. I will add another field to configure for the host, which will hopefully fix the issue.

johannesjo:
Could you please check again?

jefft:
No luck. I don't think it's just a port problem, because I get the same problem when entering 'http://redradish_jira.localhost' and 'https://issues.redradishtech.com' in the Host field (I also tried scheme-less, 'redradish_jira.localhost'). It's not even attempting a HTTP query. I hope I rebuilt SP properly. I've appended my rebuild commands at the end [1].

It could just be the node-jira library you're using that is broken. See the comment on https://github.com/steves/node-jira/issues/131. For instance, I saved the following as `testjira.js` in your super-productivity/ directory (code straight from the README):
````
JiraApi = require('jira').JiraApi;

var jira = new JiraApi('http', 'redradish_jira.localhost', '80', 'jturner', 'REDACTED', '2.0.alpha1');
jira.findIssue('LLNW-12', function(error, issue) {
	    console.log('Issue: ' + issue);
});
````
and I get:
````
jturner@jturner-desktop:~/src/github.com/johannesjo/super-productivity$ npm list | grep jira
[...]

PR #7402, opened 2026-04-28
feat(flowtime): add configurable dynamic breaks

### LINKED ISSUE #4956: 💡 Option for systematic breaks when using Flowtime
### Problem Statement

Despite using the name "Flowtime", it's only a stopwatch.

But what Flowtime is:
> In the Flowtime technique, students study for as long as they can focus and take a break whenever they feel the need. The duration of the break depends on how long they have worked on their study tasks. For example, if someone studies for 25 min or less, they will receive a 5 min break, like in the Pomodoro technique. However, if someone decides to study for a longer period of time, they will receive a longer break.

And it appears to be superior to self-regulated breaks and pomodoro:
[https://www.mdpi.com/3374606](url)

Adding this option would be great. 

### Possible Solution

An option in the settings to enable this.

Duration of the break would be calculated either by rule-based calculations (<=25 min -> 5 min; >25 min & <=50 min -> 8 min; etc.) or simpler ratio-based calculations (25 min / 5 -> 5 min; 60 min / 5 -> 5 min; etc.). Both are set and customized by the user.

_No response_

### Additional context

_No response_

Hetsavani:
### Add Dynamic Break Calculation to Flowtime

Flowtime mode currently behaves as a simple stopwatch and does not implement the actual Flowtime technique, where break duration depends on the length of the work session.

Fixes: #4956 

## Solution

Implemented configurable break calculation for Flowtime:

* Added option to enable/disable breaks in Flowtime settings
* Introduced two calculation modes:

  * **Ratio-based**: break duration is calculated as a percentage of focus time
  * **Rule-based**: break duration is determined by user-defined time ranges
* Added UI to configure both modes (percentage and custom rules)

#### Behavior

* When a work session ends, break duration is calculated based on the selected mode
* Rule-based and ratio-based configurations are fully customizable by the user

#### Result

Flowtime now aligns with its intended behavior by providing dynamic, user-controlled break durations instead of acting as a simple stopwatch.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for tackling this\! The shape (per-mode config + strategy + dialog) follows the Pomodoro pattern nicely and CI is green. A few things to address before this is mergeable:

### Blocking

**1. Triggering on `pauseFocusSession` breaks existing flows.** The new effect `offerFlowtimeBreakOnSessionPause$` fires on every `pauseFocusSession`. That action is also dispatched by:
- `syncTrackingStopToSession$` (effects.ts:187) whenever a Flowtime user with `isSyncSessionWithTracking` enabled simply stops tracking a task,
- the user clicking the regular pause button mid-work,
- several other internal call sites (e.g. effects.ts:651, 1079).

After this PR, all of those force the user into a Break screen and there's no way to pause a Flowtime session anymore. The break offer should come from an explicit end-of-session intent (a new action, or `completeFocusSession` filtered to Flowtime), not from `pauseFocusSession`.

**2. `offerFlowtimeBreak` reducer destroys the work timer.** It overwrites `state.timer` with a fresh break timer (`elapsed: 0`). Combined with #1, a user who clicks pause to take a quick break loses the work session entirely — they cannot resume it.

**3. `isBreakOffer` is dead code and the offer screen has no UI.** `focus-mode-break.component.ts` adds the computed signal but it's never referenced in the template. The break screen is rendered with the standard layout (paused 0:00 / 5:00, no "Start break" button). The user has to discover that hitting Resume is what starts the offered break — which is also what `_isResumingBreak` is for, so the semantics are now ambiguous.

**4. Debug `console.log` and `any` types must be removed.**
- `dialog-flowtime-settings.component.ts:215-218` — `console.log('Flowtime Model:', ...)` inside an `effect()` will fire on every model change.
- `model = signal<any>({})`, `(model: any) =>`, `(rule: any) =>` — violate the project's strict typing rule (CLAUDE.md anti-patterns table). Type the form model explicitly; the form holds minutes and the saved config holds ms, so make that distinction in types.

**5. No unit tests for the new logic.** `FlowtimeStrategy.getBreakDuration` has nontrivial branching (ratio %, rule range matching, fallback) and the new reducer action / effect have no coverage. The spec change in `focus-mode-strategies.spec.ts` only stubs `flowtimeConfig: () => null` to keep Pomodoro tests compiling.

### Important

- **Rule matcher silently falls back.** `focus-mode-strategies.ts:80-84` — if no rule matches, it returns `SHORT_BREAK_DURATION` (5min). With the seeded default `{min: 0, max: 25, breakDuration: 5}`, a 25+ minute session matches no rule and gets 5min — opposite of the Flowtime intent. Either anchor the last rule's max, return `null`, or surface a config validation error.
- **`maxDuration: number; // ... or Infinity`** in `global-config.model.ts:115` — `Infinity` does not survive `JSON.stringify` (becomes `null`), so it won't persist or sync. Use a sentinel (`null`/`-1`) and handle it in the matcher, or drop the comment.
- **No min/max validation.** A user can save `min=30, max=10` and silently get no break. Add cross-field validation (or at least sort/clamp on save).
- **Wrong translation key for the add button.** `dialog-flowtime-settings.component.ts:159` uses `FLOWTIME_BREAK_RULES_TITLE` ("Break rules") as the *Add* button label. The commented-out `FLOWTIME_ADD_BREAK_RULE` key is the right one — uncomment it (and remove the commented version in `t.const.ts`).
- **Unused `withLatestFrom` selectors** in the new effect (effects.ts:444): `selectFocusModeConfig` and `currentTaskId` are pulled in but never read in the `filter`/`map`. Drop them, or use `currentTaskId` for `pausedTaskId` like other call sites do.

### Minor

- `templateOptions` (Formly v5) mixed with `props` (v6) on the `breakRules` field — stick with `props`.
- `hideExpression` mixed with `expressions.hide` — pick one (modern is `expressions.hide`).
- Commented-out `// addText: T.F.FOCUS_MODE.FLOWTIME_ADD_BREAK_RULE` — remove.
- PR title: project convention is `type(scope): lowercase imperative sentence`, e.g. `feat(flowtime): add configurable dynamic breaks`.

Happy to look again once the trigger is reworked and the offer screen has a clear UI. 🙏

johannesjo:
Thanks for the quick turnaround\! The trigger rework is much cleaner — splitting `endFlowtimeSession` from `pauseFocusSession` is exactly right, the timer is preserved properly, the offer screen has a real "Start break" button, and the `any`/`console.log` cleanup looks good. CI red is unrelated flake (`planner-task-visibility`, `work-view-detail-panel-focus`).

A few things to tighten before merge:

### Important

**1. Strategy and effect disagree on "no rule matched."** `FlowtimeStrategy.getBreakDuration` returns `null` when no rule matches (and the spec asserts this), but `offerFlowtimeBreakOnSessionEnd$` (effects.ts:455-462) coalesces to `SHORT_BREAK_DURATION` and dispatches the offer anyway. So a user who configures rules and works past the highest `maxDuration` still gets a 5-min default break — the opposite of what the unit test guarantees. Please align the two: either skip the dispatch when `breakInfo` is `null`, or push the fallback into the strategy and remove `??` from the effect.

Same falsy-trap on `breakPercentage`: `if (config.breakMode === 'ratio' && config.breakPercentage)` treats `0` as unset → null → 5-min effect fallback. Decide whether 0% is "no break" or invalid input and handle it explicitly.

**2. Reducer + effect have no tests.** Strategy is well covered, but `endFlowtimeSession`, `offerFlowtimeBreak`, and `offerFlowtimeBreakOnSessionEnd$` are the spots most likely to regress. Minimum coverage I'd want:
- `endFlowtimeSession` is a no-op when `timer.purpose \!== 'work'`
- `offerFlowtimeBreak` produces `{ isRunning: false, elapsed: 0, purpose: 'break' }` and switches `currentScreen` to `Break`
- the effect skips when mode ≠ Flowtime, and (after fix #1) skips when `breakInfo` is `null`

**3. `mainState: FocusMainUIState.Preparation` while on the Break screen is overloaded.** The `isBreakOffer` computed in `focus-mode-break.component.ts` derives offer-vs-active break from this combination. It works, but `Preparation` semantically means "before starting a session," not "before starting a break." A dedicated `BreakOffer` state (or at least a comment in the reducer pinning the contract) would prevent a future refactor from quietly breaking the offer UI.

**4. Formly inconsistencies.**
- `breakMode` uses legacy `hideExpression`; `breakPercentage` and `breakRules` use `expressions.hide`. Pick one — modern is `expressions.hide`.
- `FLOWTIME_BREAK_RULES_TITLE` is defined in `t.const.ts` and `en.json` but never referenced. Either render it as a section label above the rules array or delete it.

**5. Rule validation is silent.** `save()` clamps `max < min` to `min` without telling the user, and a rule with `min == max` matches no session because the matcher uses `elapsedMs < rule.maxDuration`. Add a Formly cross-field validator (or at least a hint), so misconfigurations surface at save time.

### Minor

- The 280-line dialog uses inline template + styles. `dialog-pomodoro-settings` uses `templateUrl`/`styleUrls` — match that for readability.
- `lastCompletedDuration` is set in the new `endFlowtimeSession` reducer but I don't see it consumed anywhere — drop it if speculative.
- `completeFocusSession()` in `focus-mode-main.component.ts` now branches per mode — worth a one-liner explaining why Flowtime needs its own action so it doesn't get "simplified" back later.
- PR title: missing space + casing — `feat(flowtime): add configurable dynamic breaks`.

Looks close. Once #1 and #2 are addressed I'm happy to take another pass. 🙏

johannesjo:
Also: there is a bigger rework of focus mode coming, you might want to check it out:
https://github.com/super-productivity/super-productivity/pull/7411

Hetsavani:
Any feedback on this @johannesjo ?

johannesjo:
Thanks for the round-2 turnaround — strategy/effect alignment, the reducer + effect tests, the templateUrl extraction, and the i18n key cleanup all look good. One real correctness bug, one likely non-firing validator, and a few polish items remain.

### Important

**1. Rule matcher misses the upper boundary.** `focus-mode-strategies.ts:90`
```ts
elapsedMs >= rule.minDuration &&
(rule.maxDuration === null || elapsedMs < rule.maxDuration)
```
Combined with the dialog's seeded default `{minDuration: 0, maxDuration: 25 min, breakDuration: 5 min}`, a session of *exactly* 25 minutes matches no rule and the user gets no break — opposite of issue #4956 ("≤25 min → 5 min"). The strategy spec covers 30000/90000/150000 but no boundaries; please add a boundary case (e.g. `getBreakDuration(60000)` against `[{0, 60000, …}, {60000, 120000, …}]`) and decide on a convention — either `<=` for the upper bound, or document ranges as half-open and ensure the dialog's seeded defaults respect that.

**2. The `minMaxDuration` cross-field validator likely never fires.** `dialog-flowtime-settings.component.ts:138-152`
The top-level `breakRules` field declares `validators.validation: [{ name: 'minMaxDuration', options: ... }]` (string-name lookup), but the named validator is registered on the per-row `fieldArray.validators.minMaxDuration` and there's no project-level `FormlyConfig` registering a `minMaxDuration` validator globally. Result: `form.invalid` doesn't block save, and the silent `max < min → max = min` clamp in `save()` (line 167-173) is the only guard. Either register the validator at config level, or use the inline function form on the field. Worth verifying interactively before merge.

**3. "Break Offer" still rides on `mainState: Preparation` — bug #6575 makes this fragile.** `focus-mode.reducer.ts:148-152, 236-256`
The reducer comment pins the contract, which helps, but `unPauseFocusSession` already special-cases `purpose === 'work'` to restore `InProgress`. During an active Flowtime break, `mainState` remains `Preparation` for the entire break — only invisible because `currentScreen === Break` masks the main component. Anything in #7411 (or elsewhere) that reads `mainState` independently of `currentScreen` will mis-render. A dedicated `FocusMainUIState.BreakOffer` is ~6 lines and removes the booby-trap.

### Minor

- **`offerFlowtimeBreak.elapsedMs` is dead payload** — `focus-mode.actions.ts:70` declares it but the reducer destructures only `{ duration, isLongBreak, pausedTaskId }`. Drop it or wire it.
- **Dead initializer in the strategy** — `focus-mode-strategies.ts:73`: `let breakDuration = FOCUS_MODE_DEFAULTS.SHORT_BREAK_DURATION;` is overwritten on every non-null path; the suggestion of a 5-min fallback is misleading.
- **Hard-coded English in the validator message** — `dialog-flowtime-settings.component.ts:151`: `'Max duration must be greater than or equal to Min duration'` bypasses the translation system. Add a key to `en.json` and resolve via `TranslateService` like other formly validators.
- **Remaining \`any\` casts** — `dialog-flowtime-settings.component.ts:142` (\`control: any\`) and `:426` (\`rule: any\`). Use `AbstractControl<FlowtimeBreakRuleInMinutes>` and `FlowtimeBreakRuleInMinutes` respectively.
- **Two settings cogs at different bottom offsets** — `focus-mode-main.component.scss:52, 69`: Pomodoro at `0.05`, Flowtime at `0.1`. Mutually exclusive via \`@if\`, but the position drift is unintentional. Consider extracting a shared \`.mode-settings-btn\` class.

### Heads-up

CI red is the unrelated \`electron/indicator.test.cjs\` flake (\`initTaskWidgetSettingsListener is not a function\`) — not from this PR.

Also: a non-trivial focus-mode rework is in flight as #7411. Whichever lands first will force the other to rebase substantially; might be worth syncing with that branch on the merge order.

After items 1 + 2 (and ideally 3) this is mergeable. 🙏

johannesjo:
Thanks for the round-3 turnaround — `<=` boundary fix + boundary tests, the inline-function validator on `fieldArray.validators.minMaxDuration`, the `BreakOffer` enum, removal of the dead `elapsedMs` payload + dead initializer, the translated validator message, the typed `AbstractControl<FlowtimeBreakRuleInMinutes>`, and the shared `.mode-settings-btn` class all look good. Two real contract issues remain, plus a few minor items.

### Important

**1. `mainState` stays at `BreakOffer` after the offered break starts.** `focus-mode.reducer.ts:138-156` — `resumeBreak()` dispatches `unPauseFocusSession`, and for `purpose === 'break'` the reducer only flips `isRunning` + sets `_isResumingBreak`. It never transitions `mainState` out of `BreakOffer`. The UI currently masks this because `isBreakOffer` is gated by `\!isSessionRunning() && mainState === BreakOffer`, but `mainState` is now lying about the phase. Anything that reads `mainState` independently of `currentScreen`/`isSessionRunning` (#7411 is a likely place) will misbehave. Either set `mainState: InProgress` (or a new `BreakActive`) on the break branch of `unPauseFocusSession`.

**2. `_isResumingBreak` gets set when accepting an *offered* break.** Same dispatch path. Original semantics of that flag are *resuming a paused break*; first-time acceptance of an offer trips it too. Either guard on `mainState === BreakOffer` in `unPauseFocusSession`, or have `resumeBreak()` dispatch a different action (e.g. `startBreak({ duration: timer.duration })`) when in the offer state.

### Minor

- **`save()` mutates form-model objects in place.** `dialog-flowtime-settings.component.ts:172-178` — the `for` loop writes `rule.maxDuration = …` on objects shared with the signal model. The dialog closes immediately so users don't see it, but `sortedRules.map(r => ({ ...r, maxDuration: ... }))` would keep the model immutable.
- **No spec for `DialogFlowtimeSettingsComponent`.** ~250 lines and the only nontrivial bits left without coverage: constructor ms→min round-trip, `save()` min→ms mapping, sort + clamp, validator wiring. Highest future-regression risk in this PR.
- **Adjacent rules with shared boundary silently overlap.** `focus-mode-strategies.ts:84-87` — with a user-added second rule `[25..50: 10min]` next to the seeded `[0..25: 5min]`, an exactly-25-min session matches both; `find` returns the first. Deterministic but surprising. A one-line comment (\"ranges are inclusive on both ends; first match wins, sorted by `minDuration`\") would close it.
- **`offerFlowtimeBreak` reducer doesn't zero `_isResumingBreak` / `_isOvertimeEnabled`.** Carry-over from a prior Flowtime session is unlikely to bite, but the explicit reset costs nothing.

### Heads-up

CI is fully green this round. ✅

#7411 is still in flight — whichever lands first will force a non-trivial rebase on the other; worth syncing on merge order.

After items 1 + 2 this is mergeable. 🙏

Hetsavani:
Hi @johannesjo, any feedback on this?

johannesjo:
Thanks for the latest round. I did another pass on the current head (`b285edb`) and still see a few correctness issues before this is mergeable.

### Blocking

**1. Ending Flowtime from the banner bypasses dynamic breaks.**
`FocusModeMainComponent.completeFocusSession()` now routes Flowtime through `endFlowtimeSession`, but the banner action still calls `_handleEndSession()` in `focus-mode.effects.ts`, which dispatches `completeFocusSession({ isManual: true })` unconditionally. So if the overlay is hidden and the user ends a Flowtime session from the banner, no Flowtime break is calculated or offered. The banner path needs the same Flowtime-specific action as the main component.

**2. Flowtime sessions that offer a break are not recorded as completed focus sessions.**
`offerFlowtimeBreakOnSessionEnd$` returns `offerFlowtimeBreak(...)` instead of `completeFocusSession(...)`, and neither `endFlowtimeSession` nor `offerFlowtimeBreak` stores `lastCompletedDuration` or triggers the existing `logFocusSession$` path. After the user starts/completes/skips the break, only break actions run, so the completed Flowtime work session is never logged as a focus session whenever a break is offered. Please record/log the completed work duration before replacing the work timer with the break timer, or add a dedicated action for this transition.

**3. Accepting an offered Flowtime break does not honor `isPauseTrackingDuringBreak`.**
`FocusModeBreakComponent.resumeBreak()` dispatches `startBreak(...)` directly for a break offer. The Pomodoro start-break paths unset the current task first when `focusMode.isSyncSessionWithTracking && isPauseTrackingDuringBreak` is enabled; this Flowtime path does not. Result: task tracking continues through the Flowtime break and the break is logged as work time. This path should mirror the existing Pomodoro break-start tracking behavior.

### Important

**4. The rule validator can remain stale when `minDuration` changes.**
The `minMaxDuration` validator is attached only to the `maxDuration` control. If a user has `min=0, max=25` and then changes only `min` to `30`, Angular does not automatically re-run the sibling `maxDuration` validator, so the form can remain valid and `save()` silently clamps the value. Move the validator to the row/group level, or explicitly revalidate `maxDuration` whenever `minDuration` changes.

**5. The config supports open-ended rules, but the UI cannot create one.**
`FlowtimeBreakRule.maxDuration` and the strategy support `null` as "no upper bound", but the dialog marks `maxDuration` as required and uses a number input. Users cannot create the natural final rule (`90+ min -> 15 min`) from the UI, so rule mode is awkward for the main Flowtime use case unless they invent a very large max. Either allow an empty max as `null`, or change the model/strategy to remove the open-ended concept.

A focused test for the banner end-session path and the tracking-pause-on-offered-break path would be useful, since both are easy to regress and are not covered by the current component/reducer specs.

johannesjo:
## Review (2026-05-08, head `044083e3`)

CI is red on Lighthouse, Linux Packaged Electron App, Tests, and build-preview because the branch doesn't compile. Root cause is a **botched merge with master** — specifically with #7411 ("New focus mode rework"), which both deleted the session-banner integration in `focus-mode.effects.ts` and removed the `isSyncSessionWithTracking` config flag (it's now migrated to `autoStartFocusOnPlay` in `global-config.reducer.ts:29-58`). The PR's last merge commit kept the deleted code from the pre-rework base instead of taking master's version.

### Critical (blocks merge)

1. **`focus-mode.effects.ts` ≈L824–1190 — `updateBanner$` and the `_handle*` / `_getBannerActions` helpers don't compile.** TS errors include `Property 'bannerService' does not exist on type 'FocusModeEffects'` (never injected), `Cannot find name 'BannerId' / 'T' / 'FocusModeConfig' / 'Banner' / 'TimerState' / 'showFocusOverlay'`. Even with imports added, `BannerId.FocusMode` is not a member of the enum (`src/app/core/banner/banner.model.ts`). This entire block is a regression from master and unrelated to "Flowtime breaks". **Recommended fix: delete the block entirely** — master's design intent (#7411) is that the focus-button shows the running indicator, not a banner.

2. **`focus-mode-break.component.ts:101-107` — dead `unsetCurrentTask()` branch.** `config?.isSyncSessionWithTracking` is `undefined` because the flag no longer exists on `FocusModeConfig`. The whole `if (config?.isSyncSessionWithTracking && config?.isPauseTrackingDuringBreak) { unsetCurrentTask(); }` branch is unreachable, so accepting a Flowtime break offer with `isPauseTrackingDuringBreak: true` does **not** pause tracking — exactly the issue the earlier Copilot review flagged. The matching test in `focus-mode-break.component.spec.ts:174-186` only passes because the mock signal is loosely typed; it asserts behavior production cannot produce. Drop the `isSyncSessionWithTracking` clause and gate solely on `isPauseTrackingDuringBreak`.

### Important

3. **`focus-mode-strategies.ts:88-92` — rule-boundary semantics are order-dependent.** The strategy uses inclusive `>=` and `<=` on both ends, so adjacent rules `[0..60000]` and `[60000..120000]` overlap at `60000`, and "first match wins" depends solely on array order. The dialog sorts on save, but rules from imports / older configs / hand-edited storage carry no such guarantee, and the strategy itself doesn't sort. Use a half-open interval (`elapsedMs < rule.maxDuration` for non-null max) and add a unit test with **unsorted** input.

4. **`focus-mode.reducer.ts` `unPauseFocusSession` — `BreakOffer` state can be silently consumed.** The reducer now unconditionally sets `mainState: InProgress` and `isRunning: true`, with `startedAt: Date.now() - state.timer.elapsed` (= `Date.now()` for an offer that has `elapsed: 0`). Any code path that dispatches `unPauseFocusSession` while `mainState === BreakOffer` will skip the `startBreak` flow entirely — there's no `pausedTaskId` plumbing, no `isPauseTrackingDuringBreak` check. The `focus-mode-break` component guards against this via `isBreakOffer()`, but the (to-be-deleted) banner `_handlePlayPauseToggle` would call `unPauseFocusSession` on `isPaused = \!timer.isRunning && timer.purpose \!== null`, which is also true for `BreakOffer`. Add a guard in the reducer so `BreakOffer` is left alone (`if (state.mainState === FocusMainUIState.BreakOffer) return state;`).

5. **`focus-mode.bug-6575.spec.ts:80-88` — flipped assertion needs a comment pointing at the rationale.** The original bug-6575 fix was scoped to work sessions; the new behavior also restores `InProgress` for breaks. That's not a regression of #6575 itself, but it changes break-resume UI semantics. Document why this is now safe (e.g. reference the `BreakOffer` state addition).

6. **`logFocusSession$` (`focus-mode.effects.ts:651`) firing on `offerFlowtimeBreak` — borderline.** Functionally fine because `endFlowtimeSession` reducer captures `lastCompletedDuration` first, so the metric is correct. Worth a one-line comment in the action that "session is logged on offer, even if user declines the break".

### Minor

7. **`dialog-flowtime-settings.component.ts:225-241` — `.sort()` mutates the signal-held array.** `(currentModel.breakRules ?? []).sort(...)` sorts in place; `currentModel` is the value from `model()`. Use `[...arr].sort(...)`.

8. **`dialog-flowtime-settings.component.html` uses `<div class=\"dialog-actions\">` instead of `<mat-dialog-actions>`** — inconsistent with other dialogs and the cause of the manual `dialog-actions` SCSS.

9. **`dialog-flowtime-settings.component.ts:67-71 _defaultRuleInMinutes` is duplicated** in the `repeat` field's `defaultValue` at L136-140 — define once.

10. **Open-ended last rule UX** — `required: true` was removed (good), but the field still shows `min: 1` and no helper text. Users won't discover that "leave empty" means unbounded. Add a description string or convert to a "no upper bound" checkbox.

11. **Silent fallthrough on misconfigured `breakPercentage <= 0`** — `getBreakDuration` returns `null` and the effect dispatches `completeFocusSession` without telling the user "no break configured". Either prevent saving 0, or surface a toast.

12. **`focus-mode-break.component.html:60-68` — duplicate "Start Break" affordance.** The `BreakOffer` view shows both an in-clock play arrow and a "Start Break" button with `color=\"primary\"`. Pick one — probably keep the explicit button, drop the in-clock variant for `BreakOffer`.

13. **i18n** — only `en.json` is updated. Per project policy that's expected; ignore the Copilot bot comment on this.

### Strengths

- The Flowtime-break **core** is well-designed: ratio + rule strategies plug cleanly into `FocusModeStrategy`, the new `BreakOffer` UI state is a clean addition, and tests cover the math, sorting, clamping, and the cross-field validator with revalidation-on-min-change (which is easy to get wrong).
- New types `FlowtimeConfig` / `FlowtimeBreakRule` follow existing config patterns.
- Reducer additions for `endFlowtimeSession` / `offerFlowtimeBreak` are pure and well-tested.

### Verdict — Needs changes (blocks merge)

Cleanest path forward:

1. **`git rebase master`** (or re-do the merge) and accept master's version of `focus-mode.effects.ts` for everything outside the genuinely new `offerFlowtimeBreakOnSessionEnd$` effect and the expanded `logFocusSession$` filter. Verify `npm run build` passes.
2. **Drop the `isSyncSessionWithTracking` check** in `focus-mode-break.component.ts` and rewrite the matching test against typed mocks.
3. **Half-open boundary** in `FlowtimeStrategy.getBreakDuration`, plus an unsorted-rules test.
4. **Guard `unPauseFocusSession`** against `mainState === BreakOffer`.
5. Address the minors at your discretion.

Once #1–4 are in, the PR's actual contribution is solid and on-scope.

johannesjo:
Re-reviewed against the latest commit (`e42768`). Most Copilot points and earlier review feedback are addressed; here's what still needs attention:

### CI is red on the latest commit

- **`Tests` (e2e)** — `e2e/tests/focus-mode/auto-start-focus-on-play.spec.ts:47` ("pressing play with the opt-in on starts a focus session indicator-only (no overlay)") fails after retries with `expect(locator).toBeVisible() / element(s) not found`. The unPauseFocusSession reducer change (now always sets `mainState: InProgress`, previously preserved `Preparation` for break sessions — see the inverted assertion in `focus-mode.bug-6575.spec.ts`) is the most likely culprit. Worth running this spec locally with `npm run e2e:file e2e/tests/focus-mode/auto-start-focus-on-play.spec.ts -- --retries=0` and confirming the indicator-only flow still works.
- **`Test google-calendar-provider`** — fails because `packages/plugin-dev/google-calendar-provider/package-lock.json` still has ~470 lines of unrelated vitest/esbuild dep churn. The `chore: restore package lockfile` commit only restored the root lockfile; this nested one is still dirty. Please revert it.

### Critical

1. **`focus-mode-break.component.html:3-15` — `BreakOffer` title shows raw `{{cycle}}` placeholder.** In the offer branch, `| translate` is called with no params, but the keys are `"Long Break - Cycle {{cycle}}"` / `"Short break - cycle {{cycle}}"`. ngx-translate's default parser leaves unsupplied placeholders verbatim, so the user sees literal `Short break - cycle {{cycle}}`. After accepting the offer the else-branch runs and shows `cycle 0` (Flowtime never increments `currentCycle`). Suggest a dedicated `FLOWTIME_BREAK_TITLE` (e.g. *"Take a break"*) used whenever `mode === Flowtime` (or in `BreakOffer` state).

### Important

2. **`dialog-flowtime-settings.component.ts:91` validator — false negative when `min` is empty.**
   ```ts
   return max === null || max === undefined || max === '' || max >= min;
   ```
   While the user is still typing `minDuration`, `min` is `undefined` and `max >= undefined` is `false` (NaN comparison), flashing *"Max must be ≥ Min"* prematurely. Short-circuit when `min` is missing too:
   ```ts
   if (min == null || min === '') return true;
   return Number(max) >= Number(min);
   ```

3. **No upper bounds on duration inputs (`dialog-flowtime-settings.component.ts:103-105, 113`).** Only `min:` is set on the rule duration fields. A typo can produce a 999999-minute "break duration" silently. Add sensible `max:` (e.g. 480 min work / 120 min break).

4. **`breakPercentage` form `min: 0` allows silent no-break (`dialog-flowtime-settings.component.ts:99`).** The strategy returns `null` when percentage `<= 0`, but the form accepts 0 with no feedback. Set `min: 1`.

5. **`lastCompletedDuration` invariant is undocumented.** The `endFlowtimeSession → offerFlowtimeBreak → logFocusSession$` chain works only because `endFlowtimeSession`'s reducer sets `lastCompletedDuration` and `offerFlowtimeBreak`'s reducer doesn't touch it — relying on synchronous reducer ordering. A future refactor that resets `lastCompletedDuration` in `offerFlowtimeBreak`, or moves the write into `completeFocusSession` only, would silently break Flowtime session logging. Add a one-line comment on `endFlowtimeSession` reducer noting that this field must remain set across the offer transition.

### Minor

6. **First-match-wins on overlapping rules (`focus-mode-strategies.ts:54-58`).** Sorted by `minDuration`, but overlap is allowed, so e.g. `[0–30 → 5min]` and `[25–60 → 10min]` silently picks the 5-min rule for `[25, 30)`. Either validate against overlap in the dialog or document the semantics next to `breakRules`.
7. **`Math.round(elapsedMs * pct/100)`** on short sessions can produce sub-second breaks. Consider `Math.max(MIN_BREAK_MS, …)`.
8. **`(currentModel.breakRules ?? []).sort(...)`** in `save()` mutates the form-bound array — switch to `[...rules].sort(...)`.
9. **`pausedTaskId = storePausedTaskId || currentTaskId`** — task IDs are non-empty strings so `||` works, but `??` is more explicit about the intent.

### Note on Copilot's i18n comment

"Translation keys only added to `en.json`" is project policy (`CLAUDE.md`: *"Edit only `en.json`; never other locales"*) — Copilot is wrong here, you can ignore it.

### Already addressed since Copilot's review (nice work\!)

- `isPauseTrackingDuringBreak` now wires through `unsetCurrentTask()` on accept.
- `maxDuration` is no longer `required`; save path preserves `null` for open-ended rules.
- Boundary-overlap fix uses half-open ranges in the strategy.

johannesjo:
Thanks for the follow-up. I think this still needs a small correction before merge:

1. `e2e/tests/focus-mode/issue-6731-pause-stops-tracking.spec.ts` expects pausing focus mode to clear the current task, but the test never enables `isSyncSessionWithTracking`. Production still gates `syncSessionPauseToTracking$` on that setting, and the default is still `false`, so this should fail in a fresh E2E context and default users still keep tracking after pause.

2. The updated `pomodoro-timer-sync-bug-5954.spec.ts` tests assert that the play button is disabled when no task is current, but production only disables it when `isSyncSessionWithTracking` is enabled. These tests also do not enable that setting, so the assertions conflict with `isPlayButtonDisabled`.

3. Please avoid skipping `bug-5995-break-resume.spec.ts` unless there is truly no way to exercise the path. The break pause/resume buttons still exist in `focus-mode-break`, so this regression should probably be adapted to the new surface instead of skipped.

I did not run the test suite; this is based on static review of `e427681...fd836af`.

Hetsavani:
_css-variables.scss contained duplicate body / body.isDarkTheme selector blocks that triggered stylelint no-duplicate-selectors during lint runs. I removed the redundant duplicates as part of getting the branch back to a passing lint state. No intended theming behavior changes.

johannesjo:
Thanks for the follow-up. I found two E2E setup issues that should be fixed before merge:

1. `e2e/tests/focus-mode/auto-start-focus-on-play.spec.ts` looks for `mat-slide-toggle`, but the Focus Mode config uses Formly `type: 'checkbox'`, so this renders as `mat-checkbox`. The test will time out before exercising the auto-start behavior. Please use the shared config helper or a `mat-slide-toggle, mat-checkbox` locator.\n\n2. `e2e/tests/focus-mode/issue-6731-pause-stops-tracking.spec.ts` is out of sync with the Focus Mode config. It tries to enable the old “Sync focus sessions with time tracking” setting, but that setting has been removed from the form on the current PR head. This means the regression test fails in setup rather than validating pause/stop-tracking behavior.\n\nNo runtime blocker found in the reviewed focus-mode logic from this pass, but these E2E tests need to match the current settings UI to be useful in CI.

PR #7409, opened 2026-04-29
Caldav subtask import

baflo:
## Problem

New tasks from CalDAV are imported without their remote sub-tasks, instead every sub-task must be imported manually.

## Solution

A new option for the CalDAV issue provider is added that allows to automatically import sub-tasks.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
## Code review

Three independent passes (one focused on correctness, one on protocol/edge-case bugs, one synthesis pass). Findings ordered by severity. Several are blocking.

### Blocking

**1. Archived parent → silent sub-task loss.** `getAllIssueIdsForProviderEverywhere` (`task.service.ts:1208-1213`) includes archived task ids, so a child whose parent is archived passes the new filter at `caldav-common-interfaces.service.ts:158`. `_tryAddSubTask` (`issue.service.ts:684-690`) then calls `addSubTaskTo(archivedParent.id, …)` without checking `isFromArchive`. The `addSubTask` reducer (`task.reducer.ts:496-498`) does `if (\!parentTask) { devError(...); return state; }` — the dispatch no-ops. Net effect: the child is silently dropped on import, but the import flow believes it succeeded. Compare to the sibling code at `issue.service.ts:724-731` that calls `restoreTask` for archived hits.
*Fix:* in `_tryAddSubTask`, treat `isFromArchive: true` as "not found" and return `undefined` so the fallback at line 599 adds it as top-level — or restore the parent first.

**2. Grandchildren never imported when their parent is itself a sub-task.** In `addTaskFromIssue` (`issue.service.ts:588-619`), `_addSubTasks` is gated on `\!taskId` from `_tryAddSubTask`. So when poll N imports a child via `_tryAddSubTask` (success → `taskId` set), `_addSubTasks` is *not* invoked, even though that child may itself have children in CalDAV. On poll N+1 those grandchildren pass the filter and get routed back through `_tryAddSubTask` against a sub-task — and SP only supports one level of nesting, so they end up flattened or attached incorrectly.
*Fix:* run `_addSubTasks` after `_tryAddSubTask` succeeds too. Optionally, in `_tryAddSubTask`, walk up to the root if the resolved parent already has a `parentId`, and attach the child there.

**3. \`webSecurity: \!IS_DEV\` is out of scope and creates a dev/prod divergence in exactly the layer this PR touches.** Production keeps \`webSecurity: true\`. The existing session interceptors at \`main-window.ts:152-199\` already (a) strip \`Origin\` on outgoing requests and (b) force OPTIONS responses with \`statusCode >= 300\` to \`200 OK\`. If preflight still fails it likely returns 401 (≤ 299), so the existing \`>= 300\` condition misses it. **Try forcing the OPTIONS rewrite unconditionally**, or add \`|| details.statusCode === 401\`, before disabling webSecurity. If that doesn't suffice, proxy CalDAV through \`proxy.conf.json\` in \`ng serve\`. Either way, this should be a separate PR with its own justification — bundling a security-relevant Electron change under "subtask import" makes it easy to miss.

**4. RFC 5545 RELATED-TO handling is incomplete.** `caldav-client.service.ts:158` does `todo.getFirstPropertyValue('related-to')` and discards the `RELTYPE` parameter. Real-world emitters (Apple Reminders, Nextcloud Tasks, Thunderbird) routinely emit `RELATED-TO;RELTYPE=CHILD:…` or `;RELTYPE=SIBLING:…`. With the current code, a CHILD relation is stored as if it were PARENT, inverting the hierarchy on import. Multiple `RELATED-TO` properties on one VTODO are also legal and only the first is kept.
*Fix:* iterate `getAllProperties('related-to')` and accept only entries where `RELTYPE` is absent or equals `PARENT`.

**5. `de.json` edit violates project policy.** CLAUDE.md rule #4: *"When adding translation keys, only edit `en.json` — never edit other locale files directly."* Drop the `de.json:125` line; translations come through the localization workflow.

### Important

**6. N+1 CalDAV REPORT per poll cycle.** `getSubTasks` (`caldav-common-interfaces.service.ts:168-177`) re-runs `getOpenTasks$` once per imported top-level task, and `checkAndImportNewIssuesToBacklogForProject` (`issue.service.ts:281-289`) fires the imports **in parallel** (un-awaited `forEach`), so M new parents → M+1 concurrent full-list REPORTs against the user's server. Radicale/Baikal will rate-limit. Also re-introduces a self-race the author tried to avoid: the calendar can change between fetch #1 and the M sub-task fetches.
*Fix:* memoize `getOpenTasks$` per `issueProviderId` for the duration of one poll, or thread the already-fetched task list down through the new `getSubTasks(issueId, providerId, issue, allTasks?)` 4th argument.

**7. Self-referential / cyclic `related_to` will corrupt store.** No guard in `getSubTasks` against `t.id === issueId.toString()`. A malformed VTODO with `RELATED-TO:<own-uid>` (or an A↔B cycle) would have `_addSubTasks` try to attach a task to itself.
*Fix:* one-line `&& t.id \!== issueId.toString()` in the filter.

### Minor

**8. `_allExistingIssueIds` cast and lookup.** `_allExistingIssueIds as string[]` is sound for CalDAV (issueIds are always stringified at `issue.service.ts:569`) but the underscore prefix says "unused" while the parameter is now used, and `existingIds.includes(...)` is O(N·M) over a list that can be large. Drop the underscore, build a `Set<string>` once.

**9. Form field placement.** `caldav-cfg-form.const.ts:102-108` puts `isAddSubTasks` above the "Advanced Config" collapsible while `categoryFilter` (a similar feature toggle) lives inside Advanced. Consistent placement would be inside Advanced.

**10. Test gaps.**
- `as any` casts at `caldav-common-interfaces.service.spec.ts:60,168` defeat type checking — type the helper as `Partial<CaldavCfg>` and add `isAddSubTasks?: boolean` to the test fixture directly.
- No coverage for: archived-parent path (#1), self-referential `related_to` (#7), `_addSubTasks` invoked twice with same children, RELTYPE variants (#4).

### Strengths
- The orphan-filter approach in `getNewIssuesToAddToBacklog` is a clean lazy strategy that avoids any topological-sort/dispatch-ordering complexity.
- `getSubTasks` correctly added as optional on `IssueServiceInterface` so other providers are unaffected.
- Solid filter-level test matrix (enabled/disabled/unset, mixed lists, numeric IDs).
- Wiki updated.

### Verdict
Needs changes. #1, #2, #4, #5 are correctness/policy blockers. #3 should be split out. #6 is a strong recommendation before merge given how easily it'll be hit on real CalDAV servers.

johannesjo:
Thank you very much!

PR #7411, opened 2026-04-29
New focus mode rework

johannesjo:
https://github.com/super-productivity/super-productivity/discussions/6781#discussioncomment-16751151


johannesjo:
Probably makes sense to show the focus button on mobile too when session is running.

PR #7421, opened 2026-04-30
Added ical/ics estimated time to planner

loviuz:
## Problem

The estimated time per day does not sum the estimated time that comes from ICS/iCal connector.

immagine

## Solution

Included time from ICS/iCal connector.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo (commented):
Thank you very much for picking this up!


One change to make before merging: please drop all-day events from the sum and only count `timedEvents`. All-day iCal entries use a raw 24h `duration`, so a single tag-style event (birthday, holiday, "Conference Day") would add 24h to a day with ~8h available — a 300% progress bar. Many iCal feeds carry non-blocking all-day entries, so counting them is likely worse UX than the original bug. The new test at `planner.selectors.spec.ts:597` will need to flip to assert all-day events do **not** affect `timeEstimate`.

Concretely, in `planner.selectors.ts`:

```ts
const calendarEventsTime = timedEvents.reduce(
  (acc, ev) => acc + (ev.end - ev.start),
  0,
);
```

Two small nits:
- Title should follow `type(scope): description`, e.g. `fix(planner): include timed ical events in day time estimate`.
- CI `Tests` is red because `test:electron` failed first and the unit-test step never ran; locally the new specs pass 33/33.

loviuz:
Hi @johannesjo , I have fixed everything. Test of electron seems a bug on github actions...

I hope everything is well, because this feature is very important for me, and also for everybody (i hope) :-)

johannesjo:
Thank you very much @loviuz ! Can you maybe rebase to latest master, so we can see if the tests pass?

loviuz:
Just done

johannesjo:
Thank you! Can you have a look at the failed tests?

loviuz:
@johannesjo the error is that helpers `getLocalNoon()` and `getLocalTime()` are defined inside the `describe()` method, but I don't know if it is a good idea to move them outside `describe()` to use them also in `planner.selectors.ts` other than `planner.selectors.spec.ts`. I don't know node very well so I don't know the best practices of node...

johannesjo:
Thanks for iterating on this @loviuz\! The implementation in `planner.selectors.ts` looks correct — summing only `timedEvents` is the right call. However, the new specs in `planner.selectors.spec.ts` have two issues that, together, are why CI is red and why "locally pass" is misleading.

### 1. `getLocalNoon` / `getLocalTime` aren't in scope for the new tests

You correctly identified this in your last comment. `getLocalNoon` (line 19) and `getLocalTime` (line 24) are `const`-declared inside the callback of `describe('Planner Selectors - All Day Events')` (line 17, closes at line 317). The new tests live in a *sibling* `describe('Planner Selectors - selectPlannerDays')` (line 319) and reference these helpers at lines 520, 569, and 631 — sibling describe callbacks are separate function scopes, so this throws `ReferenceError` at run time.

**Fix:** hoist both helpers to module scope (just above line 17, outside any `describe`). They're pure date utilities with no test-specific state — perfectly safe to share. No need to worry about Node best practices here; this is just JavaScript block-scoping.

```ts
// at the top of the file, above the first describe()
const getLocalNoon = (year: number, month: number, day: number): number =>
  new Date(year, month - 1, day, 12, 0, 0, 0).getTime();

const getLocalTime = (year: number, month: number, day: number, hour: number): number =>
  new Date(year, month - 1, day, hour, 0, 0, 0).getTime();
```

### 2. Even after fixing scope, the positive assertions are vacuous

The new tests pass `[today]` as `dayDates` (where `today = getDbDateStr()` is the real current date) but build calendar events with `start: getLocalTime(2025, 1, 15, …)`. Inside `getPlannerDay`, `getIcalEventsForDay` filters by `getDbDateStr(new Date(start)) === dayDate`, so a 2025-01-15 event is dropped for today's date. `timedEvents` ends up empty, `calendarEventsTime` is 0, and:

- `expect(result[0].timeEstimate).toBe(7200000)` → actual 0 → **fails**
- `expect(result[0].timeEstimate).toBe(10800000)` → actual 3600000 (just the task) → **fails**
- `expect(result[0].timeEstimate).toBe(0)` for all-day → passes, but **vacuously** — the event was filtered out before the all-day branch ran, so it doesn't actually prove the new exclusion logic works

**Fix:** derive event timestamps from `today` so the date filter matches:

```ts
const todayAtHour = (h: number): number => {
  const [y, m, d] = today.split('-').map(Number);
  return new Date(y, m - 1, d, h, 0, 0, 0).getTime();
};
// then: start: todayAtHour(10),
```

Apply this in all three new tests (use noon for the all-day one). After that, the assertions will actually exercise the new `timedEvents.reduce(...)` path and the all-day exclusion.

### Nits (carried over from the earlier review)

- Title still doesn't follow Conventional Commits — `fix(planner): include timed ical events in day time estimate` would do it. Easy to squash-rename on merge if that's preferred.

Implementation logic itself is sound and ready — just the spec needs these two adjustments and CI should go green.

johannesjo:
Thank you very much!

PR #7440, opened 2026-05-01
feat: allow dev override for local REST API

npdias:
## Summary
- add DEV-only SP_FORCE_LOCAL_REST_API=1 override for starting Local REST API
- keep persisted user setting unchanged
- document development override in API reference

## Verification
- npm run electron:build
- npm run test:electron
- commit/push hooks: npm run lint, npm run test


johannesjo:
Thank you very much!

johannesjo:
Thank you very much! <3

PR #7444, opened 2026-05-01
feat: add local REST project endpoints

npdias:
## Summary
- add Local REST endpoints to create, fetch, and update projects
- keep project list/note relation fields reducer-owned by rejecting taskIds, backlogTaskIds, and noteIds
- add focused LocalRestApiHandlerService coverage for project create/update validation

## Verification
- npx ng test --watch=false --include='src/app/core/electron/local-rest-api-handler.service.spec.ts' --browsers=ChromeHeadless
- pre-push hook: npm run lint
- pre-push hook: npm run test

### DIFF
--- a/src/app/core/electron/local-rest-api-handler.service.spec.ts
+++ b/src/app/core/electron/local-rest-api-handler.service.spec.ts
@@ -45,6 +45,22 @@ describe('LocalRestApiHandlerService', () => {
     subTasks,
   });
 
+  const createMockProject = (id: string, overrides: Record<string, unknown> = {}): any =>
+    ({
+      id,
+      title: `Project ${id}`,
+      icon: 'list_alt',
+      isArchived: false,
+      isHiddenFromMenu: false,
+      isEnableBacklog: false,
+      taskIds: [],
+      backlogTaskIds: [],
+      noteIds: [],
+      advancedCfg: {},
+      theme: {},
+      ...overrides,
+    }) as any;
+
   const mockElectronApi = (): void => {
     (window as any).ea = {
       onLocalRestApiRequest: (handler: (payload: LocalRestApiRequestPayload) => void) => {
@@ -126,8 +142,10 @@ describe('LocalRestApiHandlerService', () => {
       ['add', 'update', 'remove', 'archive'],
       {
         list$: of([]),
+        getByIdOnce$: (_id: string) => of(undefined),
       },
     );
+    (projectServiceMock as any).add.and.returnValue('new-project-id');
 
     tagServiceMock = jasmine.createSpyObj(
       'TagService',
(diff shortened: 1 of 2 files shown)


johannesjo:
Thanks for the PR\! Took a pass over the diff — overall the implementation is consistent with the existing `_handleTaskRoutes` shape and the allow/reject-list approach to relational fields is the right call. CI is green. A few things worth addressing before merge:

### Important

- **Docs out of sync.** `docs/wiki/3.01-API.md` (around the "Local REST API" section) still lists only `GET /projects`. The new `POST /projects` and `PATCH /projects/:id` endpoints, the allowed/rejected field lists, and the new error codes (`UNSUPPORTED_FIELD`, `PROJECT_NOT_FOUND`) should be documented alongside the existing task endpoints — otherwise the new surface is invisible to anyone reading the documented contract.
- **`isArchived` via `PATCH` bypasses the dedicated action.** `ALLOWED_PROJECT_FIELDS` includes `isArchived`, so callers can flip archive state via plain `PATCH`. Today the `archiveProject`/`unarchiveProject` reducers only set `isArchived`, so behavior matches — but the commented-out `pfapiService.archiveProject` in `project.effects.ts` suggests side effects were once intended there. If they're ever restored, `PATCH`-based archiving will silently skip them. Suggest mirroring the task pattern: drop `isArchived` from the allowlist and add `POST /projects/:id/archive` + `/unarchive` endpoints.

### Minor

- **No boolean type validation.** `isArchived: "true"` (string) or `isEnableBacklog: 1` will be written straight into the store via `update()`. Task PATCH has the same gap, so not a regression, but a `typeof === 'boolean'` guard (or coercion + 400) would harden the new entry points.
- **No test for `GET /projects/:id`.** `_handleProjectRoutes` adds it but the spec only covers `PATCH`. Worth adding 200 + 404 cases to match.
- **Test stub setup.** `Object.defineProperty(projectServiceMock, 'getByIdOnce$', { get: () => ... })` is used per-test to swap the stub. Plain reassignment (`projectServiceMock.getByIdOnce$ = (_id) => of(x)`) is simpler and reads more like the task-side tests.
- **Small duplication.** `pickAllowedProjectFields`/`firstRejectedProjectField` duplicate the task helpers with only the allow/reject set differing. Optional — only worth a tiny extraction (`pick(body, ALLOW_SET, transformers?)`) if a third entity follows the same pattern.
- `if (segments[0] === 'projects' && segments[1] && segments.length >= 2)` — `segments[1]` truthy already implies `length >= 2` since `path.split('/').filter(Boolean)` strips empties. Mirrors the task-routes line, so probably fine to leave for symmetry.

### Nice things

- Tight allowlist + explicit reject list with a clear comment on why relational fields are reducer-owned.
- Good failure-mode coverage in the tests (rejected fields, empty/missing title, unknown-field stripping, 404 on missing project).
- `id` injection is implicitly defeated — `pickAllowedProjectFields` strips it and `ProjectService.add()` mints its own nanoid, so callers can't smuggle one in.

PR #7450, opened 2026-05-01
docs: update README to reflect state of wiki

zenoprax:
The wiki is now ready for general use. I have added some notes regarding
the use of DeepWiki and how it complements the official one. The
`.devin` directory contains a file that should help steer the auto-doc
process. My two primary aims are:
1. avoid polluting the DeepWiki resource with the existing docs in case
   of error propagation and LLM context poisoning
2. Generate a differential of the observed code and the observed content
   in docs/wiki to aid in future doc updates.

### DIFF
--- a/.devin/wiki.json
+++ b/.devin/wiki.json
@@ -1,0 +1,27 @@
+{
+  "repo_notes": [
+    {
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! <3

Issue #6578, opened 2026-02-19
🚨 Additional Information Panel belong to wrong task


chriswhaas:
### Version Used

17.1.8 Windows

### Expected Behavior

Clicking  on the side panel / Information tab should always belong to the selceted task. Otherwiese it can happen that user write wrong information to wrong task! An open information panel should CHANGE when another task ist selected.

### Current Behavior

An open side panel /information stays open for the last task, selecting other task do not change that and keeps the old one open.

### Steps to Reproduce

select a task and press mouse button or "I".

than
select other task with mouse, the old information stay open.
or
select other task on the 6 dots at the very beginning of each task. the task is selected and marked with a border line around the task, but the old information pane is still open. Writing something into it writes it to the the very first task and not he last selected ones. Very confusing  and can create a lot of errors

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
Thank you! This should be fixed!

PR #7465, opened 2026-05-03
fix(side-panel): Focus side-panel title field for sub-tasks

### LINKED ISSUE #7120: 💥 Mobile: when creating subtasks, title field is not in focus
### Steps to Reproduce

1. Add a task, exit from making a second task click on the task you've just created.
2. Click add subtask
3. Type the subtask name
4. The text your typing does not show in the expected 'subtask title's Field which is expanded but instead is seen typed in the subtask title field that is in the background

Expected behavior.
1. Create task 
2. Exit from making a new takes and enter the task just created.
3.click add subtask
4. Type
5 title text shows in the currently expanded subtask title field 

### Url
https://localhost/#/project/INBOX_PROJECT/tasks

### Meta Info
META: SP18.1.3_L2A __ Browser – en-GB – Linux aarch64 – en-GB – UA:Mozilla/5.0 (Linux; Android 16; A059P Build/BQ2A.250721.001-BP2A.250605.031.A3; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/146.0.7680.164 Mobile Safari/537.36

### Actions Before Error
```

```

cpalvis:
When a new sub-task is created, the title field in the side details panel is highlighted for editing instead of the title field in the main task list.

Fixes #7120 

## Problem

Previously, when creating a new sub-task, the title field in the main task list view was focused for editing instead of the title field in the side details panel. This caused issues for mobile users because the side panel view overlays the task list on mobile and prevents interaction with the focused input.

## Solution
When a subtask is added, the side panel title field is focused.
### Details

- Added `focusTitleField()` function to focus side panel title field
- Added `taskDetailPanel` Signal to reference the detail panel component
- Added `setTaskDetailPanel` to set `taskDetailPanel` to active side panel
- Called `focusTitleField()` when subtask is added instead of `focusTitleForEdit()`

## Testing

This is my first code-changing PR and I'm not very familiar with Angular testing, so my apologies in advance if there are missing tests that are necessary. I'll be happy to figure them out and update this PR if needs be.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/e2e/pages/work-view.page.ts
+++ b/e2e/pages/work-view.page.ts
@@ -63,8 +63,13 @@ export class WorkViewPage extends BasePage {
 
     await task.press('a');
 
-    // Wait for textarea to appear
-    const textarea = this.page.locator('textarea:focus, input[type="text"]:focus');
+    // Sub-task creation focuses the side-panel title field (mobile fix #7120);
+    // when no panel is mounted we fall back to whatever input has focus.
+    const textarea = this.page
+      .locator(
+        'task-detail-panel task-title textarea, textarea:focus, input[type="text"]:focus',
+      )
+      .first();
     await textarea.waitFor({ state: 'visible', timeout: 3000 });
 
     // Ensure the field is properly focused and cleared before filling
--- a/src/app/features/tasks/task-detail-panel/task-detail-panel.component.html
+++ b/src/app/features/tasks/task-detail-panel/task-detail-panel.component.html
@@ -5,6 +5,7 @@
       (valueEdited)="updateTaskTitleIfChanged($event.wasChanged, $event.newVal)"
       [value]="task().title || ''"
       class="task-title"
+      #taskTitleField
     ></task-title>
     <!--    <button mat-icon-button>-->
     <!--      <mat-icon>more_vert</mat-icon>-->
--- a/src/app/features/tasks/task-detail-panel/task-detail-panel.component.ts
+++ b/src/app/features/tasks/task-detail-panel/task-detail-panel.component.ts
@@ -152,8 +152,6 @@ export class TaskDetailPanelComponent implements OnInit, AfterViewInit, OnDestro
     isExpandedAttachmentPanel: signal(!IS_MOBILE),
   };
 
-  setTaskDetailPanel = this._taskFocusService.taskDetailPanel.set(this);
-
   // Observable conversions
   private _task$ = toObservable(this.task);
 
@@ -444,6 +442,7 @@ export class TaskDetailPanelComponent implements OnInit, AfterViewInit, OnDestro
 
   ngOnInit(): void {
     window.history.pushState({ [HISTORY_STATE.TASK_DETAIL_PANEL]: true }, '');
+    this._taskFocusService.taskDetailPanel.set(this);
   }
 
   ngAfterViewInit(): void {
@@ -472,6 +471,11 @@ export class TaskDetailPanelComponent implements OnInit, AfterViewInit, OnDestro
       window.history.back();
     }
     window.clearTimeout(this._focusTimeout);
+    // Only clear the singleton if we still own it — another panel may have
+    // taken over (right-panel ↔ dialog ↔ bottom-panel can overlap briefly).
+    if (this._taskFocusService.taskDetailPanel() === this) {
+      this._taskFocusService.taskDetailPanel.set(null);
+    }
   }
 
   changeTaskNotes($event: string): void {
--- a/src/app/features/tasks/task-detail-panel/task-detail-panel.component.ts
+++ b/src/app/features/tasks/task-detail-panel/task-detail-panel.component.ts
@@ -82,6 +82,7 @@ import { isMarkdownChecklist } from '../../markdown-checklist/is-markdown-checkl
 import { Log } from '../../../core/log';
 import { isInputElement } from '../../../util/dom-element';
 import { checkKeyCombo } from '../../../util/check-key-combo';
+import { TaskFocusService } from '../task-focus.service';
 
 @Component({
   selector: 'task-detail-panel',
@@ -125,6 +126,7 @@ export class TaskDetailPanelComponent implements OnInit, AfterViewInit, OnDestro
   private _translateService = inject(TranslateService);
   private _destroyRef = inject(DestroyRef);
   private _dateTimeFormatService = inject(DateTimeFormatService);
+  private _taskFocusService = inject(TaskFocusService);
 
   // Inputs
   task = input.required<TaskWithSubTasks>();
@@ -134,6 +136,7 @@ export class TaskDetailPanelComponent implements OnInit, AfterViewInit, OnDestro
   // View children
   itemEls = viewChildren(TaskDetailItemComponent);
   attachmentPanelElRef = viewChild<TaskDetailItemComponent>('attachmentPanelElRef');
+  taskTitleField = viewChild<TaskTitleComponent>('taskTitleField');
 
   // Constants
   IS_TOUCH_PRIMARY = IS_TOUCH_PRIMARY;
@@ -148,6 +151,9 @@ export class TaskDetailPanelComponent implements OnInit, AfterViewInit, OnDestro
     isDragOver: signal(false),
     isExpandedAttachmentPanel: signal(!IS_MOBILE),
   };
+
+  //Set panel as last focused panel
+  setTaskDetailPanel = this._taskFocusService.setTaskDetailPanel(this);
 
   // Observable conversions
   private _task$ = toObservable(this.task);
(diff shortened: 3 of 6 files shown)


cpalvis:
@johannesjo A couple things I know need to be addressed for this PR.
## Failing E2E Tests
I'm not sure what the best way to approach fixing these failing E2E tests. 
### Diagnosis
All 5 tests are failing for the same reason. They all call the `WorkViewPage.addSubTask()` function, which is failing.
In part, `addSubTask()` adds a title to the title text field of a sub-task:
```  
// Wait for textarea to appear
const textarea = this.page.locator('textarea:focus, input[type="text"]:focus');
await textarea.waitFor({ state: 'visible', timeout: 3000 });

// Ensure the field is properly focused and cleared before filling
await textarea.click();
await textarea.fill('');

// Use fill() instead of type() for more reliable text input
await textarea.fill(subTaskName);
await this.page.keyboard.press('Enter'
```
Due to the changes made by this PR, the title input field in the work list is no longer focused when a sub-task is added, so the `await textarea.waitFor({ state: 'visible', timeout: 3000 })` times out because `this.page.locator('textarea:focus, input[type="text"]:focus')` does not return a valid element. Once it times out, the function returns an error and any test using it fails.
### Solution
I'm working on fixing `addSubTask()` to identify the title element without using its focused state, but without being changed, the test will always fail despite no functionality being broken. If this PR is otherwise mergeable, how do you want to handle this?

## Keyboard Shortcut
I had hoped to open the side panel and display the focused title field when a user creates a sub-task via the `a` keyboard shortcut, but I know that a change to the shortcut is in the works (#7472) and I don't want to make any changes that would immediately get overwritten or end up blocking/complicating another PR.

Thank you!

johannesjo:
Thank you!

Some additional review:


**Critical**

1. **Stale singleton reference.** `taskDetailPanel` is set on construction but never cleared on destroy. Three places mount `<task-detail-panel>` (right panel, bottom panel, dialog) and whichever instantiates last wins; once it's destroyed, the next sub-task creation will throw `'No el'`. Add to `ngOnDestroy`:
   ```ts
   if (this._taskFocusService.taskDetailPanel() === this) {
     this._taskFocusService.taskDetailPanel.set(null);
   }
   ```

2. **Field-initializer side effect.** `setTaskDetailPanel = this._taskFocusService.setTaskDetailPanel(this);` is awkward — the field is named for a setter but holds `void`, and registration runs before view children are queried. Move the call into `ngOnInit` (or `constructor`) and drop the stored return value.

3. **Race between selection and focus.** `_focusNewlyCreatedTask` calls `taskElement.focus()` (which triggers `setSelectedId(newSubTaskId)` via `TaskComponent`'s `focusin` handler) and then synchronously calls `taskDetailPanel.focusTitleField()`. Until store → selector → CD propagates, the panel is still bound to the parent task — its top `<task-title>` `[value]` is the parent's title, and `set value()` will rewrite `tmpValue` on the next CD. Worth verifying on mobile + a throttled CPU; if it's flaky, consider passing `taskId` into `focusTitleField` and waiting for `task().id === taskId` before focusing.

4. **Regression when no panel is mounted.** The new `&& taskDetailPanel` guard makes the focus a no-op when no detail panel exists. Old code fell back to `taskComponent.focusTitleForEdit()`. Keep that fallback or test the new behavior explicitly.

**Important**

- `TaskFocusService.setTaskDetailPanel` is a one-line wrapper around `signal.set`. Drop the method and call `taskDetailPanel.set(this)` directly from the panel.
- `if (taskDetailPanel) { taskDetailPanel?.focusTitleField(); }` in `task.component.ts:431` — dead `?.`.
- `focusTitleForEdit()` was relocated within `task.component.ts` without any logic change. That's unrelated churn — please revert to keep the diff minimal.
- No unit tests. At minimum, cover `TaskFocusService.setTaskDetailPanel` and the three branches in `_focusNewlyCreatedTask` (panel present, panel null, id mismatch).

**On the failing E2E** — `WorkViewPage.addSubTask()` keys on `:focus` and needs to be updated regardless. Once the focus *actually* lands on the side-panel input, switch the helper to a stable locator like `task-detail-panel #t-\${id} task-title textarea` (or whatever your view exposes). The fact that the test fails is informative — it's also worth a quick manual check that text does end up on the new sub-task and not the parent in the window before CD runs.

**Minor**
- `//Set panel as last focused panel` — comment is slightly misleading (focus isn't tracked) and missing the space after `//`.
- `Log.log(taskTitleField)` right before `throw` adds nothing (`taskTitleField` is `undefined` there); either drop or use `Log.err` with a message.

Re: the keyboard shortcut and #7472 — agreed, leave that out of scope.

johannesjo:
Independent review — verified against the worktree. Not restating points already raised in @johannesjo's review; just adding what's new and sharpening a couple of his points with concrete file:line evidence.

### Confirmed: three mount sites

`<task-detail-panel>` is rendered at:
- `src/app/features/bottom-panel/bottom-panel-container.component.html:30`
- `src/app/features/right-panel/right-panel-content.component.html:23`
- `src/app/features/tasks/dialog-task-detail-panel/dialog-task-detail-panel.component.html:4`

This validates the stale-singleton concern. Concrete failure: open task in detail dialog → close it → the dialog's `TaskDetailPanelComponent` is destroyed but `TaskFocusService.taskDetailPanel` still references it. Next sub-task creation throws `No el` from `focusTitleField()` because `viewChild` is gone. The `ngOnDestroy` guard suggested in the maintainer review is required, not optional.

### Sharpening the field-initializer point

`setTaskDetailPanel = this._taskFocusService.setTaskDetailPanel(this);` runs in the class field initializer phase, which means **every** `TaskDetailPanelComponent` overwrites the singleton on construction — regardless of whether it's the active panel. With both right-panel and dialog mounted, the last one constructed wins, and switching between mounts silently re-targets focus to whichever was constructed most recently. Combined with the missing destroy cleanup, this is a worse race than staleness alone.

### New: regression mirrors in `task.service.ts`

The maintainer flagged the no-panel fallback regression in `task.component.ts`. The same regression exists in `task.service.ts:780-789` — the new `&& taskDetailPanel` guard makes the empty-title focus a silent no-op when no detail panel is mounted. The previous code unconditionally called `taskComponent.focusTitleForEdit()`. Either keep that fallback or document the new contract; otherwise inline-list sub-task creation with no panel open loses focus-into-edit entirely.

### New: service-imports-component layering smell

`task-focus.service.ts` already imports `TaskComponent` (predates this PR), and now adds `TaskDetailPanelComponent`. A `providedIn: 'root'` service holding strong references to component classes is a layering inversion and a cycle candidate. Cleaner shape: a small interface (e.g. `interface FocusableTitlePanel { focusTitleField(): void }`) so the service holds a behavioral contract rather than a concrete component class. Optional but cheap while the file is open.

### Minor: duplicated anti-pattern

`Log.log(taskTitleField); throw new Error('No el');` in the new `focusTitleField()` mirrors the same pattern at `task.component.ts:1023-1030` (`TaskLog.log` + identical throw). Logging an `undefined` local then throwing an ungrep-able message adds nothing — drop the log or replace with `Log.err('TaskDetailPanel: title field missing')`.

### E2E fix for @cpa-27

Verified `e2e/pages/work-view.page.ts:67` uses `'textarea:focus, input[type=\"text\"]:focus'`. Once focus actually lands on the side-panel input, switch to a stable locator scoped to the panel:

\`\`\`ts
const textarea = this.page
  .locator('task-detail-panel task-title textarea, task-detail-panel task-title input[type=\"text\"]')
  .last();
\`\`\`

Or scope by the new sub-task's `#t-\${id}` once known. Worth adding an assertion that the typed text lands on the new sub-task and not the parent — that's exactly the race in maintainer review point #3, and a regression test would lock it down.

### Verdict

Needs changes — the maintainer's critical items (singleton cleanup, field-initializer move, race verification, no-panel fallback) plus the `task.service.ts` mirror regression above. E2E helper update is required regardless. Solid mobile UX fix once these land.

johannesjo:
Hey @cpa-27 – thanks a lot for digging into this and putting up a fix for #7120!

Your PR surfaced the real bug and the analysis in your follow-up comment (the failing E2E, the diagnosis of `addSubTask` focus timing) made it much faster for me to reason about.

After a careful review I went a slightly different direction in #7516. Same goal — and using your fix as the starting point — but with a smaller, less intrusive change. Instead of routing focus through a new `focusTitleField()` method on the panel, we just prefer the *last* DOM element matching `#t-<id>` in `TaskService.focusTaskById`. When both the main list and the side panel render the same task (a just-created sub-task lives in the parent's sub-task list inside the panel), this selects the panel copy. The panel's existing `focusin` handler short-circuits `setSelectedId` for tasks inside the panel (#6578), so we get two extra benefits as a side effect:

- Parent context is preserved — adding a sub-task no longer navigates the panel away from the parent the user was viewing.
- The race condition between the selection change and the focus call disappears, because there's no selection change to wait on.

Net diff is +11 / −25 (smaller than removing your changes alone, since it also drops the `taskDetailPanel` singleton it would have leaned on).

Going to close this PR in favor of #7516, but please don't read that as a rejection — your investigation directly informed the simpler approach, and I'd love to see more contributions from you. The CONTRIBUTING flow worked: you proposed something concrete, the review surfaced a structural concern, and we converged on a better design together. Thanks again\! 🙏

Issue #7417, opened 2026-04-30
💡 Planner & Schedule – ability to manage past days


stefanopago-glitch:
### Problem Statement

oNw, it is not possible set a task planned for the previous day.
And also, it is not possible see what you have done in the previous day from the Planner & Schedule pages.

### Possible Solution

It would be very useful to have access to past days, both to review tasks completed in previous days and to potentially move a task to a past date if you actually completed it on that day but forgot to mark it as done.

### documentation gap

- [ ] ... related docs should be updated as part of the same PR whenever possible.
- [ ] ... I can help improve the documentation.

Diogo-22:
Hi there! Could I develop this feature alongside with @FAFialho as part of a uni project?

johannesjo:
Hey @Diogo-22 ! Thank you very much for the offer! I think first thing we neeed is a UX/UI concept for this. Past tasks are usually archived. While it might be nice to view past data (already possible via worklog and quick history btw) the focus should be on planning present and future. Maybe there is a good way to make this all work nicely in one integrated view,  but we need a good concept for everything before we implement.  I am happy for any suggestions!

@kiausch this possibly is also related to what you are working on.

Diogo-22:
Hi @johannesjo ! Thank you so much for the feedback. Totally understand the focus towards the future and present. Given that, my partner and I are going to look into other features for the project. We'll be sure to let you know if we find something. Thanks again!

zenoprax:
@Diogo-22 
You might find this comparison helpful: https://github.com/super-productivity/super-productivity/wiki/4.22-Quick-History#quick-history-vs-worklog

stefanopago-glitch:
Hi everyone, 
Has there been any further evaluation on this?

I think this feature would be very useful for fixing activities from previous days that may have been incorrect or incomplete (for example, forgetting to log an activity or adjust it).

Thanks a lot in advance for any feedback.

ibadullahsyed:
> and to potentially move a task to a past date if you actually completed it on that day but forgot to mark it as done.

By far the biggest problem ive faced with this app, i feel forced to finish work before midnight hits and i havent yet clicked the "Finish Day" button

I would **love** to see this feature asap, but ive never worked on electron apps so i cant do it myself

zenoprax:
> By far the biggest problem ive faced with this app, i feel forced to finish work before midnight hits and i havent yet clicked the "Finish Day" button

[...]

Issue #6449, opened 2026-02-10
💡 Static weekly Planner and schedule to / view previous dates.


panta-ray:
### Problem Statement

I am begrudingly leaving my last planner (Elisi) after many years, and SP is the closest I have come and I _really, really_ love it and incredibly thankful for the work that has been put into it. There are just two related things I'd like to request. The second is the most problematic for me.

1. I don't like that the planner rolls forward each day, showing you today + the next 7 days from whatever day it is currently. 
2. Tasks can't be scheduled to past dates, they are greyed out, and it's hard to look backwards. Quick History and Worklog exist, but on the Planner and Schedule pages you can't go back to previous days or months. There are also many times when I need to add in something I did yesterday or last week as a reference, or forgot, or for the reasons I can't think of.

### Possible Solution

1. Static weekly blocks you can click through like the months in Schedule view. An option to choose this or rolling, so everyone is happy, and set which day a week starts on if users want to customise it. My previous app used weekly numbers, much like month numbers e.g. Week 7 = Feb 8-14 and so on.
2. Schedule to past dates, and be able to move backwards in the Planner / Schedule to see them. 

### Additional context

I have found a few requests related to mine:

https://github.com/super-productivity/super-productivity/issues/4811
https://github.com/super-productivity/super-productivity/issues/5250
https://github.com/super-productivity/super-productivity/issues/3187

johannesjo:
Thanks for opening this up. Not sure if you knew, but for history you have the "Worklog" and "Quick History" pages.

Estus-Dev:
Scheduling to previous dates is important for repeating tasks, particularly ones that cannot be acted upon until their due date.

For instance, I have a yearly task that I must do on a specific day. I cannot move forward on that task before that date, and it is overdue after that date. I just attempted to add it for next year, but it shows up both as an active task and as a repeating task.

If I could instead schedule it for last year's start, and then configure the repeat, I could check off "last year's" version of the task and it would only show up in Repeating Tasks.

PR #7467, opened 2026-05-03
feat: add toggle to show/hide done tasks in Schedule view

ashesofdream:
Problem
Schedule view lacks the ability to show completed tasks, making it difficult to review past work and completed scheduled items.

Solution
Added toggle button in Schedule view to show/hide done tasks. Done tasks appear with strikethrough style, and users can now navigate to past periods to review completed tasks.

Implementation:
- Added selectTimelineTasksIncludeDone selector to fetch tasks including done ones
- Preserved dueWithTime when tasks complete, only clearing remindAt
- Added toggle button with check_box icon in Schedule toolbar
- Done tasks rendered with strikethrough style and disabled drag/resize
- Removed navigation restriction to allow viewing past periods

Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)
Checklist
- [x] I have included relevant changes to the documentation/wiki (https://github.com/super-productivity/super-productivity/tree/master/docs/wiki).
- [x] I have run npm run checkFile on changed .ts/.scss files
- [x] My commit messages follow the Angular format (type(scope): description)

related issue:
- https://github.com/super-productivity/super-productivity/issues/7417
- https://github.com/super-productivity/super-productivity/issues/6449


ashesofdream:
Hi @johannesjo .Fan of Super Productivity here. I noticed the Schedule view was missing the ability to show completed tasks, so I implemented it. Hope this can be useful for others as well.

johannesjo (commented):
## Summary

This PR adds a toggle in the Schedule view toolbar to show/hide completed tasks. Done tasks appear with strikethrough styling and are non-draggable/non-resizable. The navigation guard preventing backward navigation is removed, allowing users to review past periods. When completing a scheduled task, `dueWithTime` is preserved (only `remindAt` is cleared via `dismissReminderOnly`) so done tasks remain positioned correctly in the schedule timeline.

## PR Discussion Context

- No reviewer comments yet. CI: lint, build, Lighthouse, CodeQL all pass; unit tests still pending.
- Maintainer noted in issue #7417: "we need a good concept for everything before we implement" and "the focus should be on planning present and future." This PR may need alignment discussion with the maintainer on whether this matches the intended UX direction.

## Issues

### Important

1. **`task.reducer.util.ts:143-147`** — The logic `shouldClearDueWithTime = task.dueWithTime && doneOn < task.dueWithTime` clears `dueWithTime` when a task is completed *before* its scheduled time. This means: if a user schedules a task for 3 PM and completes it at 2 PM, the task disappears from the schedule view entirely (its `dueWithTime` is cleared and it won't show even with the toggle on). Consider always preserving `dueWithTime` on completion, since the "show done tasks" feature is most useful for seeing what was scheduled regardless of when it was actually completed.

2. **`schedule.component.ts:241-256`** — Removing the past-navigation guard entirely allows navigating indefinitely into the past with no lower bound. Users could accidentally scroll months/years back into empty views. Consider adding a reasonable limit (e.g., 90 days) or showing an empty-state message for periods with no data.

3. **`task.selectors.ts:604-622`** — Adding `!task.isDone` to `selectAllTasksWithDueTime` and `selectAllTasksWithDueTimeSorted` affects **all consumers** of these selectors: the Scheduled List Page, Planner Service, and Time Block Sync Effects. While excluding done tasks from these views is likely correct (you don't want done tasks in your planner or syncing to calendar), this is a behavioral change to existing functionality beyond the schedule view. Please verify these consumers behave correctly with the change, and consider adding a brief comment explaining why done tasks are excluded at this level.

### Minor

4. **`schedule.component.html`** — The `check_box` icon may confuse users since checkboxes represent task completion elsewhere in the app. Consider `checklist`, `task_alt`, or `history` to better convey "show completed tasks."

5. **`schedule.component.scss`** — The horizontal scroll + gradient fade behavior on `.schedule-nav-controls` is nice polish but unrelated to the done-tasks feature. Consider splitting into a separate commit for cleaner git history.

6. **`showDoneTasks` state persistence** — The toggle resets to `false` on page reload. The week/month view preference persists via localStorage. Consider whether this toggle should also persist, especially since users navigating to past periods to review done tasks would lose context on reload.

7. **`schedule-event.component.ts:88-93`** — Using `inject(CdkDrag, { self: true })` + `effect()` is a valid pattern for host directives. However, since `isDraggable()` returns `false` initially (before the `task` signal resolves), the effect will briefly set `disabled = true` for all events during first render, then re-run to enable dragging for non-done tasks. In practice this should be invisible (happens within one CD cycle), but worth noting.

## Strengths

- Solid test coverage: new unit tests for reducer util, selector specs, and component spec updates.
- Clean separation: new `selectTimelineTasksIncludeDone` selector vs modifying the existing one.
- Properly disables both drag and resize on done tasks.
- `dismissReminderOnly` reuse is a good choice — avoids duplicating reminder cleanup logic.
- The `Set`-based lookup optimization in the refactored `selectTimelineTasks` is a good improvement.

## Verdict

**Needs discussion** — The implementation is technically sound and well-tested. The main concerns are product-level: (1) the maintainer's comments on #7417 suggest UX alignment is needed before merging, (2) the `dueWithTime` clearing logic for early-completed tasks creates an unintuitive gap in the feature, and (3) the unbounded past navigation could use a limit. None of these are blocking bugs — they're design decisions that would benefit from maintainer input.

johannesjo:
Thanks for opening this up! 

I am unable to make this work however (testing here: https://acea13fc.super-productivity-preview.pages.dev/). Done tasks are not shown no matter for what day or if scheduled with time or not. I also think the UI does not look ideal. It looks like the toggle is part of the switch for week/month view, which doesn't make sense to me.

ashesofdream:
> Thanks for opening this up!
> 
> I am unable to make this work however (testing here: https://acea13fc.super-productivity-preview.pages.dev/). Done tasks are not shown no matter for what day or if scheduled with time or not. I also think the UI does not look ideal. It looks like the toggle is part of the switch for week/month view, which doesn't make sense to me.

Thanks for sharing this preview!
I tried the preview you shared, and it actually works on my side:
image
I'm not sure how you're scheduling the task though. In my case, I just create a task with a due time, and mark it as completed from the scheduled view.

Could you describe your workflow a bit more? It might help me reproduce the issue.

Also, I added a rule that if a task is completed before its due time , it will be ignored in the current view logic.

I added this restriction because if a task is completed before its start/due time, it technically wasn’t done within the scheduled time window. So I felt it might be confusing to still show it in the future schedule.

About the UI — do you have any suggestions? One idea could be placing the toggle next to the date, or separating it more clearly from the week/month switch block, since I agree the current placement can be confusing.

johannesjo:
I schedule a task and then I use mark as completed from the schedule task context menu and it disappears. 

I am also struggling with this a bit conceptually tbh. Tasks with no time spent on them now block spaces in the schedule. What is this supposed to mean? Why is that useful?

johannesjo:
Btw. there is also this PR open and currently worked on:
https://github.com/super-productivity/super-productivity/pull/7497

Maybe we should align on this and also discuss the best UI/UX approach for the whole problem first?

johannesjo:
Thanks for the work here — I've been reading this PR alongside the other one since they both extend the schedule toolbar in similar directions, and I wanted to share some UX thoughts before either lands.

**Cross-posting the same comment on #7467 and #7497 since they overlap significantly.**

### The concern

Both PRs add toolbar toggles that conceptually answer the same question: *"what do I want to see when looking at the past?"*

- #7467 → "Show done tasks" + unlock past navigation
- #7497 → "Show past workload" + unlock past navigation + reference calendars

Each toggle in isolation makes sense, but stacked together they accrete into 2–3 opaque icon toggles next to the existing view-mode and date-nav buttons, with no shared mental model. The "unlock past navigation" gating in particular creates a "why can/can't I scroll back?" moment for the user.

### Proposal: auto-context + a single Layers menu

1. **Past navigation always available** — drop `[disabled]="isViewingToday()"` on the prev button entirely.
2. **Auto-show overlays when viewing the past** — when the entire displayed range is before today, default `doneTasks` and `pastWorkload` to ON. On today/future, default OFF. No toggle needed for the common case.
3. **One "Layers" menu** (`layers` icon → `mat-menu`) to override the defaults: Done tasks · Past workload · Reference calendars. One affordance, scalable. A dot-badge on the icon can hint "you've overridden a default."
4. **Reference calendars stay configured per-provider** (as #7497 already does via the `isReferenceCalendar` flag). The Layers menu entry only controls *visibility* in the schedule, and only appears if at least one reference calendar exists.
5. **Distinct visual treatments per layer** so the schedule doesn't become soup: done = strikethrough + faded; past workload = flat heatmap behind the time grid; reference events = muted color + lock badge (carries "read-only" without docs).

Sketch:

```ts
type LayerKey = 'doneTasks' | 'pastWorkload' | 'referenceCalendars';
private _layerOverrides = signal<Partial<Record<LayerKey, boolean>>>({});

isViewingPast = computed(() => {
  const days = this.daysToShow();
  const today = this._todayDateStr();
  return !!today && days.length > 0 && days[days.length - 1] < today;
});

isLayerOn(key: LayerKey): boolean {
  const o = this._layerOverrides()[key];
  return o ?? (key === 'referenceCalendars' ? true : this.isViewingPast());
}
```

### How this maps onto each PR

- **#7467**: keep the `selectTimelineTasksIncludeDone` selector and the `dueWithTime` retention in the reducer — that's a real fix worth keeping. Replace the standalone toolbar toggle with a Layers-menu entry, and drop the past-nav unlock (it becomes unconditional).
- **#7497**: keep all the new services, the `isReferenceCalendar` provider flag, and the past-workload calendar entries. The two new toolbar buttons collapse into the same menu.

### Conflict note

These PRs touch overlapping files (`schedule.component.{ts,html,scss}`, `schedule.service.ts`, `schedule-event.component.{ts,scss}`) and both unlock past navigation — whichever lands second will need a rebase. The proposal above is a chance to land them as one coherent piece rather than two competing toolbar additions.

Curious what you both think — happy to discuss further or sketch it out more concretely.

PR #7472, opened 2026-05-03
feat(task): add Ctrl+Enter and escape shortcuts

### LINKED ISSUE #7420: 💡 Enhance subtask entry to be more naturally keyboard-friendly
### Problem Statement

Currently, when entering multiple subtasks, one has to either right-click the task and use the menu item or press the 'a' key. I think this can be enhanced to more closely align with the typical experience across most applications.

### Possible Solution

Instead (or in addition to) pressing 'a', one should simply be able to press Enter when a task is already highlighted.

For example, after typing a subtask, one would press Enter once to complete/save/persist the task. It then exits "edit mode" and is simply highlighted. The user should be able to press Enter again to add a new subtask. The new subtask should immediately be in edit mode with the flashing cursor ready to receive input.

However, it would also be wise to leverage the Escape key to cancel the additional entry in which case it should simply disappear without any prompt regardless if it is empty or has text. The reasoning behind this is that the user has not pressed Enter to persist the new entry. The last persisted subtask in the list should receive highlight (but NOT edit mode).

_No response_

### Additional context

_No response_

### documentation gap

Hetsavani:
## Problem

Currently, creating subtasks via keyboard requires pressing `a`, and there is no smooth way to chain subtask creation while editing a task. Additionally, cancelling a newly created subtask requires manual deletion.

Fixes: #7420 

## Solution
This PR introduces a keyboard-first workflow for subtasks:

- **Ctrl+Enter (Mod+Enter)** while editing:
  - Saves the current task (exits edit mode)
  - Creates a new subtask
  - Focuses the new subtask in edit mode

- **Escape on a newly created subtask**:
  - Cancels the subtask if it is still empty
  - Restores focus to the previous task (not in edit mode)

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/tasks/task-shortcut.service.ts
+++ b/src/app/features/tasks/task-shortcut.service.ts
@@ -167,6 +167,13 @@ export class TaskShortcutService {
       ev.preventDefault();
       return true;
     }
+    // Ctrl/Cmd + Enter on a focused task: smart subtask creation
+    // (focuses an existing empty subtask, else spawns a new one).
+    if ((ev.ctrlKey || ev.metaKey) && ev.key === 'Enter') {
+      this._handleTaskShortcut(focusedTaskId, 'addSubTaskOrFocusEmpty');
+      ev.preventDefault();
+      return true;
+    }
     if (checkKeyCombo(ev, keys.taskAddAttachment)) {
       this._handleTaskShortcut(focusedTaskId, 'addAttachment');
       ev.preventDefault();
--- a/src/app/features/tasks/task-shortcut.service.ts
+++ b/src/app/features/tasks/task-shortcut.service.ts
@@ -125,7 +125,9 @@ export class TaskShortcutService {
 
     // Ctrl/Cmd+Enter on a focused (but not editing) task: same as the `a`
     // shortcut — create a new subtask. Must run before the plain-Enter
-    // "edit title" handler below.
+    // "edit title" handler below. A user-bound `togglePlay` is checked
+    // earlier (line ~74), so remapping `togglePlay` to Mod+Enter takes
+    // precedence over this hardcoded combo.
     if ((ev.ctrlKey || ev.metaKey) && ev.key === 'Enter') {
       this._handleTaskShortcut(focusedTaskId, 'addSubTask');
       ev.preventDefault();
--- a/src/app/features/tasks/task-shortcut.service.ts
+++ b/src/app/features/tasks/task-shortcut.service.ts
@@ -123,6 +123,15 @@ export class TaskShortcutService {
     // Check if the focused task's context menu is open - if so, skip arrow navigation shortcuts
     const isContextMenuOpen = this._isTaskContextMenuOpen(focusedTaskId);
 
+    // Ctrl/Cmd+Enter on a focused (but not editing) task: same as the `a`
+    // shortcut — create a new subtask. Must run before the plain-Enter
+    // "edit title" handler below.
+    if ((ev.ctrlKey || ev.metaKey) && ev.key === 'Enter') {
+      this._handleTaskShortcut(focusedTaskId, 'addSubTask');
+      ev.preventDefault();
+      return true;
+    }
+
     // Basic task actions that work through component delegation
     if (
       !isContextMenuOpen &&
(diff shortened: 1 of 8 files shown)


johannesjo (commented):
## Review

Nice addition — the `submitTrigger` discriminated-union approach is clean and the separation between `task-title` (communicates *what* triggered submit) and `task.component` (decides *what to do*) is well-architected. A few things to address before this is merge-ready:

---

### Critical: Escape deletes any empty subtask, not just freshly-created ones

```typescript
if (submitTrigger === 'escape' && this.task().parentId) {
  const previousTaskEl = this._getPreviousTaskEl();
  if (!newVal) {
    this._taskService.remove(this.task());
  }
  this._focusTaskHost(previousTaskEl);
  return;
}
```

This will silently delete an **existing** subtask if a user edits it, clears the title, then presses Escape. The issue spec says *"cancel the addition"* — it should only auto-delete a task that was just created and still has its original empty title (never persisted with content).

A simple fix: guard with a check that the task's *previously persisted* title was already empty (i.e., the task was freshly spawned). For example, you could check `!wasChanged && !newVal` (title was empty before *and* after editing) or pass a flag from `addSubTaskTo` marking it as newly created.

---

### Important: Mod+Enter on a top-level task creates a child, not a sibling

`addSubTask()` uses `this.task().parentId || this.task().id` — so on a top-level task it adds a subtask *to itself* rather than a sibling at the same level. The maintainer's spec says "save and immediately open a new **sibling** subtask in edit mode." For top-level tasks this might be intentional (can't create top-level siblings this way), but it's worth clarifying whether `Mod+Enter` should be a no-op on tasks without a `parentId`, or if creating a child is the desired fallback.

---

### Minor: Shared type alias

The union `'blur' | 'escape' | 'enter' | 'modEnter'` is repeated in three places. Consider extracting:

```typescript
export type SubmitTrigger = 'blur' | 'escape' | 'enter' | 'modEnter';
```

---

### Minor: No tests included

The checklist says tests were added but no test files are in the diff. Given the data-loss risk from the Escape behavior, at least a unit test for "Escape on freshly-created empty subtask removes it" vs "Escape on existing subtask with cleared title does NOT remove it" would be valuable.

---

Overall the approach is solid and additive (no existing shortcuts are altered). Addressing the accidental-deletion edge case is the main blocker. 👍

Hetsavani:
Regarding the "sibling subtask" behavior:

For subtasks, the current implementation creates a sibling at the same level, which aligns with the expected chaining behavior.

For top-level tasks, it currently creates a subtask (same as the existing `a` shortcut). I wanted to clarify the intended behavior here — should Mod+Enter create a sibling task at the same level instead (similar to Notion/Todoist), or is creating a subtask the desired fallback?

Happy to adjust based on the expected UX.

johannesjo:
Thanks for the iteration and for adding the spec\! A couple of things still need attention before this is merge-ready:

**1. Top-level-task behavior (your question):** Let's make `Mod+Enter` a **no-op when the focused task has no `parentId`**. Creating a child as a fallback would surprise users who expect Notion/Todoist sibling semantics, and we don't have a great way to spawn a top-level sibling from inside an edit. Smallest change: in `addSubTask()` (or a new `addSiblingSubTask()`), early-return when `\!this.task().parentId`. Keeping the `a` shortcut for "add child to current task" is enough for the top-level case.

**2. The Escape-delete guard isn't actually fixed in the production code.** `task.component.ts` still has:

```ts
if (submitTrigger === 'escape' && this.task().parentId) {
  ...
  if (\!newVal) {
    this._taskService.remove(this.task());
  }
}
```

A user clearing the title of an existing subtask and pressing Escape will lose it silently. The new spec defines a `shouldDeleteOnEscape` helper *inside the test file* that uses `\!wasChanged && \!newVal`, but that helper is never called by `TaskComponent` — so the test passes while the real bug remains. Two things to do:
- Tighten the guard in the actual method to `\!wasChanged && \!newVal` (or thread a `wasNewlyCreated` flag from `addSubTaskTo`).
- Replace the standalone-helper spec with a test that actually exercises `TaskComponent.updateTaskTitleIfChanged()` (TestBed + spy on `_taskService.remove`) so this regression is caught next time.

**3. Minor:** `_getPreviousTaskEl()` uses `document.querySelectorAll('task')`, which crosses list boundaries (right panel, project sub-lists). After delete, the "previous" focus target may be an unrelated task from another panel. Scoping with `.closest('task-list')?.querySelectorAll('task')` would be safer.

The shape of the change is good — `SubmitTrigger` discriminated union and the split between `task-title` (what triggered) and `task.component` (what to do) is clean. Once #1 and #2 are sorted, this should be ready. 👍

Hetsavani:
Thanks for the detailed feedback!

I've updated the Escape handling in `TaskComponent.updateTaskTitleIfChanged()` to use `!wasChanged && !newVal`, ensuring that only freshly created, untouched subtasks are deleted on Escape. This prevents accidental deletion when a user clears an existing subtask title.

Additionally, the test now directly exercises `TaskComponent.updateTaskTitleIfChanged()` using TestBed and spies on `_taskService.remove`, instead of relying on a standalone helper, so the behavior is properly validated.

For Mod+Enter, I introduced a separate path to create sibling subtasks and made it a no-op for top-level tasks, while keeping the existing `a` shortcut unchanged.

johannesjo:
Thank you very much!

PR #7483, opened 2026-05-04
feat(settings): changed format of start time of the day from hours to …

### LINKED ISSUE #7455: 💡 Start time of the next day: Use hours:minutes not just hours

Hetsavani:
…hh:mm

## Problem

Start time of the next day: Use hours:minutes not just hours
Fixes: #7455
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Won't this break things for people who have configured this already in the old style?

Hetsavani:
I’ve kept it backward compatible by supporting numeric values (treated as hours). So existing configs should continue to work without breaking.

That said, I can add a migration step to normalize everything to HH:mm if you think that’s preferable.

johannesjo:
Maybe the safest rollout here is to introduce a new setting for the minute-precision value and phase out the old one over time.

That would avoid mixed-version clients interpreting the same persisted field differently:
- old clients keep reading `startOfNextDay` as the legacy numeric hour value
- new clients can read a new field for the exact value (for example `HH:mm` or total minutes)
- during the transition, new clients can still derive the new field from the old one when it is missing

I think the important part is that new clients should keep writing the legacy field as well for compatibility, otherwise older clients will drift or misinterpret the value. That way we can make the new behavior available without immediately breaking mixed-version sync.

Hetsavani:
Addressed compatibility concerns by introducing a new field `startOfNextDayTime` (HH:mm) while keeping the legacy `startOfNextDay` value.

- New clients read from `startOfNextDayTime` and fall back to the legacy field if needed
- Both fields are kept in sync to avoid drift across mixed-version clients
- Added validation at the data layer to safely handle external/synced data

This should ensure no breakage for existing users or during sync across different app versions.

johannesjo:
Thanks for iterating on this — the new-field approach matches what was suggested and CI is green. A few things to address before merging:

**Important**

1. **DRY: HH:mm parsing/normalization is duplicated 4 times** across `date.service.ts` (`parseTimeStringToMinutes`), `global-config.reducer.ts` (`normalizeStartOfNextDayConfig`), `global-config.service.ts` (`normalizeSectionCfg`), and `migrate-legacy-backup.ts` (`getStartOfNextDayDiffMs`). Worse, the variants differ subtly — `date.service.ts` parses minutes and clamps to 1439, while the reducer only reads `hourStr`. Please extract a single util (e.g. `src/app/util/start-of-next-day.util.ts`) and call it from all four sites.

2. **`normalizeSectionCfg` in `global-config.service.ts` is redundant.** `updateSection` normalizes, then dispatches `updateGlobalConfigSection`, and the reducer normalizes again. The service-side pass adds nothing for user-initiated updates and surprises any consumer that dispatches the action directly. The reducer should be the single source of truth — drop it from the service.

3. **No tests for the new minute-precision behavior.** The only spec changes are tweaks to existing tests (notably `{startOfNextDay: 4} as any` in `global-config.effects.spec.ts:216,237` to avoid the new default-merge). Please add cases for at least:
   - `DateService.setStartOfNextDayDiff('02:30')` → 2h30m diff
   - malformed/empty strings → 0
   - `normalizeStartOfNextDayConfig({startOfNextDayTime: '02:30'})` writes `startOfNextDay: 2`
   - `normalizeStartOfNextDayConfig({startOfNextDay: 4})` writes `startOfNextDayTime: '04:00'`
   - backup `getStartOfNextDayDiffMs('02:30', undefined)` → 9_000_000
   - `loadAllData` reducer synthesizes Time from legacy hour for old snapshots

4. **The legacy field is silently lossy** — when a new client writes `02:30` the legacy `startOfNextDay` becomes `2`. If an old client re-saves the snapshot with only the legacy field, normalization on the receiving new client will overwrite `startOfNextDayTime` to `02:00`, losing the minutes. This is the unavoidable trade-off of mixed-version sync, but please add a comment on `normalizeStartOfNextDayConfig` and mark `MiscConfig.startOfNextDay` `@deprecated` so future readers know which field is canonical.

**Minor**

- If `{startOfNextDay: 4, startOfNextDayTime: '02:30'}` arrives together (sync/REST/plugin), the helper sets `startOfNextDay = 2`, silently dropping the incoming `4`. Reasonable, but document that "Time wins when present."
- Consider replacing `{startOfNextDay: 4} as any` in the spec with a properly shaped fixture once normalization is asserted on its own.
- The form regex requires leading-zero hours (`02:30`, not `2:30`). A `placeholder="02:30"` and/or auto-pad would be friendlier.
- `migrate-legacy-backup.ts:getStartOfNextDayDiffMs` should reuse the shared util once it exists, or at least cross-reference `DateService.setStartOfNextDayDiff`.

Functionally it works and the architecture choice is sound — once parsing is centralized and there's spec coverage for the new precision, this should be good to go.

johannesjo:
Thank you very much!

Issue #6957, opened 2026-03-25
🚨 Caldav Error in Android App


Winnie-LD:
### Version Used

17.1.8-L2A

### Expected Behavior

Connecting the App via caldav with my two nextcloud calendars and searching in one of the calendars for words in a task should result in listed tasks that contain these words.

### Current Behavior

I connected two nextcloud calendars via different URLs successfully with the superproductivity Windows App (version 17.3.0). When opening the "Problemfeld" and searching for words, it shows related tasks in the respective calendar.
Opening the android app (version 17.1.8-L2a), the app first syncs and gets the login data for my two calendars, installs thema and shows them in the "Problemfeld" menu.
When I then try to search for keywords in one of the calendars, it shows the error message: "CALDAV: Anfrage aufgrund eines clientseitigen Netzwerkfehlers fehlgeschlagen."
I already checked the login - it's got the same login data as the windows app, same URL, same settings.

How am I supposed to get the CALDAV connection in the android app up and running?

### Steps to Reproduce

1. Connect CALDAV calendar in Windows 11 App.
2. Open android app and sync settings
3. Search for keywords in calendars (vie "Problemfeld" menu)
4. Error

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

baflo:
I've got the same error on 18.1.0 on Android 15 and 17.1.8 on Android 16.

Related to #3603 ?

This is all I get in the logs regarding the error:

```
  {
    "timestamp": "2026-04-03T10:05:41.251Z",
    "level": "ERROR",
    "context": "issue",
    "message": "Error polling issue updates for 5ptTOo4mctq-Wv-iSAyWM",
    "args": [
      {
        "HANDLED_ERROR_PROP": "Caldav: Error: CALDAV NETWORK ERROR: Error"
      }
    ]
  },
```

Winnie-LD:
Android Version of Superproductivity, update to 18.1.2_L2A, error still occurring!
The identical config works with Superproductivity on Windows11, but provokes the above mentioned error on Android 16.
any hints on how to get this working?

Winnie-LD:
Hi,
just tried pulling tasks from my caldav resources - same error again. Using http does not help (in caldav config -> caldav-URL, base-URL).
Could you provide for more and more detailed information on how to eliminate this error?

Popo8701:
AFAIK, the fix is not released yet.

PR #7491, opened 2026-05-05
Fix issue 6957

baflo:
## Problem

#6957 

## Solution

Solve CORS issue. I hope this solves it in the same manner as it did for WebDAV.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/issue/providers/caldav/caldav-client.service.spec.ts
+++ b/src/app/features/issue/providers/caldav/caldav-client.service.spec.ts
@@ -1,4 +1,9 @@
+import { TestBed } from '@angular/core/testing';
 import { CaldavClientService } from './caldav-client.service';
+import { SnackService } from '../../../../core/snack/snack.service';
+import { CaldavCfg } from './caldav.model';
+
+// ─── _getParentRelatedTo ──────────────────────────────────────────────────────
 
 describe('CaldavClientService._getParentRelatedTo', () => {
   const getParentRelatedTo = (todo: unknown): string | undefined =>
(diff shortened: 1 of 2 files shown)


johannesjo:
Thanks for tackling this — the architectural approach (proxy XHR through `WebDavHttp` to bypass WebView CORS, mirroring the sync provider) is right. Two reviews flagged the same blocker plus several smaller issues, so I wanted to consolidate before merge.

### Blocker

**The fake `XMLHttpRequest` never fires `onreadystatechange`, but that's the *only* completion handler `@nextcloud/cdav-library` v1.5.3 subscribes to.** See `node_modules/@nextcloud/cdav-library/dist/index.mjs:876` — the success path is built entirely on:

```js
xhr.onreadystatechange = () => {
  if (xhr.readyState \!== 4) return;
  // …handles 2xx, 4xx, 5xx, and 207
};
```

The library never sets `xhr.onload` or `addEventListener('load', …)`. The fake's resolve branch (`caldav-client.service.ts:413`) sets `readyState = 4` and then fires `onload` + `'load'` listeners only — so on Android/iOS every CalDAV request will hang forever (no resolve, no reject). The unit tests pass because they assert against the fake's own surface (`onload`) rather than cdav-library's actual contract. Could you confirm a real-device run against a Nextcloud CalDAV before merge?

**Fix:** in both the resolve and reject branches, after setting `readyState = 4`, also call `fakeXhr.onreadystatechange?.(event)`. Then add a regression test that mirrors cdav-library's pattern (assign `xhr.onreadystatechange` *after* `xhr.send()`, assert it fires with `readyState === 4`).

### Important

- **Native naming / iOS coverage** (`caldav-client.service.ts:365` and the inline comment "On Android WebView"). The gating uses `Capacitor.isNativePlatform()`, which is true on iOS, and `ios/App/App/WebDavHttpPlugin.swift` already exists. Either rename to `_getNativeXhrProvider` (and update the comment + PR description) or narrow the gate to `Capacitor.getPlatform() === 'android'`. Today iOS silently goes through this path with no mention in the PR.
- **Wrong comment** at `caldav-client.service.ts:464`: "CapacitorHttp requests cannot be cancelled" — this path uses the custom `WebDavHttp` plugin (OkHttp / URLSession), not CapacitorHttp.
- **`abort()` is a no-op, leaving zombie callbacks.** cdav-library installs `xhr.onabort` (`index.mjs:923`). If anything aborts, the in-flight `.then` still fires `onload`/`onreadystatechange` later and resolves a request the caller assumed dead. Set an `aborted` flag, gate the resolve/reject on it, and dispatch `onabort`.
- **Listener arrays are mutated during dispatch** (`caldav-client.service.ts:399`, `415`). `forEach` over the live array means a listener that calls `removeEventListener` mid-dispatch can shift indices. Snapshot first: `[...(eventListeners.get('load') ?? [])].forEach(…)`.

### Minor

- **Defensive header lowercasing.** `getResponseHeader` lowercases the lookup, but storage is `Object.assign(responseHeaders, res.headers)` — relies on Kotlin/Swift to lowercase. Both currently do; lowercase on intake to keep this robust.
- **Type cast hides the contract.** `fakeXhr as unknown as XMLHttpRequest` (`:467`) is exactly the safety net that would have caught the `onreadystatechange` miss. An internal `XhrLike` interface enumerating the members cdav-library uses would make the contract reviewable.
- **Tests target the fake's surface, not the integration target.** Adding a cdav-library-style integration test (per the blocker fix) covers this.

Happy to look again once these are in. Nice find on the underlying CORS root cause.

johannesjo:
Thank you very much! <3

Issue #7228, opened 2026-04-14
💡Tasks on Calendar


Abdoh-Abqr:
### Problem Statement

Basically, i need to be able to see the past tasks on callender view same S done on Google calender not to be used on history feature or report, it's better to plan future tasks bassed on graphic history of past tasks.

2 FEATURE : when task is done it should not dissappear from the calender view. It should remain in ba kground layer as a stamp. Hence in the end of the day i will be apple to trace when actually i started the task and how long it took and gaps or reset between tasks could be easy seen.

### Possible Solution

In or der to do that it get fixed position once i started track a task and the lenghth is the actuall spent time .nor manually input from user.

### documentation gap

- [ ] ... related docs should be updated as part of the same PR whenever possible.
- [ ] ... I can help improve the documentation.

luciolebrillante:
I am also very interested by this feature 👍 

Currently, doing a task means forget everything whereas I like to be able to easily check what I have done this week/month. 

In the planner panel, a done task is strike, it could be interesting to have the possibility to enable or disable this feature in the calendar view when the past is displayed. It could help to know "Oh, I forget to do this" or "Yes, I remember I did this".

rhinolover:
I agree this would be very helpful to have! I would love to be able to go back and see when I did something.

PR #7497, opened 2026-05-06
Improving scheduler view: add reference calendar and work log events

baflo:
## Problem

Augment the scheduler view with more data like:
* past work load #7228
* reference calendars

## Solution

I added an option "Reference calendar" to the calendar type issue providers. This has the effect that its events are stamped to the scheduler view and cannot be edited nor created to tasks. No reminders are created for these events.

image

Also I added a toggle button which, when activated, shows past work loads and therefore also allows to move back to previous weeks/month to review the workloads.

image

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for the PR\! Tried it on the preview build and read through the diff — really nice direction, but a few things worth addressing before merge. Findings below were independently re-verified against the branch.

## Critical

- **Past-work entries double-count when tasks have subtasks** — `buildPastWorkCalendarEntries` filters every task with `timeSpentOnDay[d] > 0`, but its input (`selectAllTasks`) returns parents *and* subtasks, and `reCalcTimeSpentForParentIfParent` keeps the parent's `timeSpentOnDay[d]` as the sum of its children. So a parent + one subtask both > 0 on a day will both render and the day's total reads ~2× the actual tracked time. The fix is to drop one side consistently (e.g. exclude tasks with a `parentId`, or skip parents with non-empty `subTaskIds`).

## Important

- **Hardcoded UI strings** — `Reference calendar (display only, no task creation)` and `Calendar color` in `calendar.const.ts` are flagged with `// TODO translation`. The new `Show work log` / `Hide work log` tooltips, `Calendar visibility` aria-label, and the chip's `[matTooltip]="calProviderLabel(provider)"` in `schedule.component.html` are also literal English. `en.json` diff has no new keys. Per `CLAUDE.md`, all UI strings need to go through `T` / `TranslateService`.

- **Past-work timing is synthetic, not actual session times** — `_getWorkStartForDay` picks one work-start per day (earliest project/tag session, or `workStart`, or 09:00) and `buildPastWorkCalendarEntries` then stacks tasks back-to-back sorted by `task.created`. Issue #7228 is specifically about *seeing when work actually happened* ("I will be able to trace when actually I started the task… gaps or resets between tasks could easily be seen"). The current placement loses gaps/order entirely. Either drive placement from per-task session deltas in `TimeTrackingState`, or surface clearly in the UI that this is a synthesized "stack from start of day", not a real timeline.

- **Hidden-providers filter is fragile** (currently safe) — `schedule.service.ts` filters via `entry.items[0]?.calProviderId`. Today every entry comes from a single provider so it works, and an empty-items entry would also pass through (the test asserts that as expected). But nothing on `ScheduleCalendarMapEntry` enforces the one-provider invariant. Safer to filter at item level, or store `calProviderId` on the entry itself.

- **Stale archive cache** — `PastWorkScheduleEntriesService._archiveCache` is only cleared via `clearCache()`, called only when toggling the work-log off. If the user archives a task while the work-log is visible, that task won't appear until they toggle off and on again. Either drop the cache (idb is fast once warm) or hook invalidation to archive mutations.

- **Toggle-off leaves user stranded in the past** — flipping `showWorkLog` off while `_selectedDate` is in the past keeps the user on a historical week with no work-log entries and the disabled-back-button predicate (`isViewingToday() && \!showWorkLog()`) still treats it as navigable. `toggleShowWorkLog(false)` should also reset `_selectedDate` to `null` if the current selection is in the past.

## Minor

- **All past-work entries render in `--c-primary`** — entries don't carry `projectId`, so `ScheduleEventComponent.projectColor()` returns null and the SCSS fallback paints every past-work entry the same primary color regardless of the task's project. Threading `task.projectId` onto the synthetic event would let them inherit the same project tinting live tasks already get.
- **Filter chip bar shows for a single provider** — `enabledCalendarProviders().length >= 1` makes the bar appear even with one provider, where toggling has no comparative value. `> 1` would be tighter.
- **Empty `mat-chip-option`** — chips render only the color dot. Without hover tooltips (touch / SR), it's hard to know which provider is which; consider showing a short label inside the chip.
- **`color` field is free-form `string`** — typed via `<input type="color">` so browsers normally constrain to `#rrggbb`, but nothing validates server-side / on import before the value reaches `[style.background]` and `--project-color`. Tightening to `^#[0-9a-fA-F]{6}$` removes the ambiguity.
- **`HiddenCalendarProvidersService`** uses raw `localStorage`; nearby services use the `loadFromRealLs` / `saveToRealLs` wrappers — mostly a consistency nit.
- **Conditional spread style** in `calendar-integration.service.ts` — `...(calProvider.isReferenceCalendar && { isReferenceCalendar: true as const })` reads awkwardly; just always set the field with `\!\!calProvider.isReferenceCalendar`.

## Strengths
- Solid test coverage: new specs for `buildPastWorkCalendarEntries`, `HiddenCalendarProvidersService`, the schedule visibility filter, and the `isReferenceCalendar` stamping path. The `flushMicrotasks()` + static-import workaround for the lazy iCal loader is well-documented inline.
- Banner suppression for reference calendars is correctly threaded through `calendar-integration.effects.ts`.
- Empty-menu short-circuit in `schedule-event.component.ts` is a nice touch.

Happy to dig deeper on any of these — the subtask double-count is the one that I'd really like to see fixed before merge.

johannesjo:
Btw. there is also this PR open and currently worked on:
https://github.com/super-productivity/super-productivity/pull/7467

Maybe we should align on this and also discuss the best UI/UX approach for the whole problem first?

johannesjo:
Thanks for the work here — I've been reading this PR alongside the other one since they both extend the schedule toolbar in similar directions, and I wanted to share some UX thoughts before either lands.

**Cross-posting the same comment on #7467 and #7497 since they overlap significantly.**

### The concern

Both PRs add toolbar toggles that conceptually answer the same question: *"what do I want to see when looking at the past?"*

- #7467 → "Show done tasks" + unlock past navigation
- #7497 → "Show past workload" + unlock past navigation + reference calendars

Each toggle in isolation makes sense, but stacked together they accrete into 2–3 opaque icon toggles next to the existing view-mode and date-nav buttons, with no shared mental model. The "unlock past navigation" gating in particular creates a "why can/can't I scroll back?" moment for the user.

### Proposal: auto-context + a single Layers menu

1. **Past navigation always available** — drop `[disabled]="isViewingToday()"` on the prev button entirely.
2. **Auto-show overlays when viewing the past** — when the entire displayed range is before today, default `doneTasks` and `pastWorkload` to ON. On today/future, default OFF. No toggle needed for the common case.
3. **One "Layers" menu** (`layers` icon → `mat-menu`) to override the defaults: Done tasks · Past workload · Reference calendars. One affordance, scalable. A dot-badge on the icon can hint "you've overridden a default."
4. **Reference calendars stay configured per-provider** (as #7497 already does via the `isReferenceCalendar` flag). The Layers menu entry only controls *visibility* in the schedule, and only appears if at least one reference calendar exists.
5. **Distinct visual treatments per layer** so the schedule doesn't become soup: done = strikethrough + faded; past workload = flat heatmap behind the time grid; reference events = muted color + lock badge (carries "read-only" without docs).

Sketch:

```ts
type LayerKey = 'doneTasks' | 'pastWorkload' | 'referenceCalendars';
private _layerOverrides = signal<Partial<Record<LayerKey, boolean>>>({});

isViewingPast = computed(() => {
  const days = this.daysToShow();
  const today = this._todayDateStr();
  return !!today && days.length > 0 && days[days.length - 1] < today;
});

isLayerOn(key: LayerKey): boolean {
  const o = this._layerOverrides()[key];
  return o ?? (key === 'referenceCalendars' ? true : this.isViewingPast());
}
```

### How this maps onto each PR

- **#7467**: keep the `selectTimelineTasksIncludeDone` selector and the `dueWithTime` retention in the reducer — that's a real fix worth keeping. Replace the standalone toolbar toggle with a Layers-menu entry, and drop the past-nav unlock (it becomes unconditional).
- **#7497**: keep all the new services, the `isReferenceCalendar` provider flag, and the past-workload calendar entries. The two new toolbar buttons collapse into the same menu.

### Conflict note

These PRs touch overlapping files (`schedule.component.{ts,html,scss}`, `schedule.service.ts`, `schedule-event.component.{ts,scss}`) and both unlock past navigation — whichever lands second will need a rebase. The proposal above is a chance to land them as one coherent piece rather than two competing toolbar additions.

Curious what you both think — happy to discuss further or sketch it out more concretely.

baflo:
Thank you for the review. I really thougth the `start` and `end` times of the tasks on the regarding days would be the working times, but turns out the actual times aren't stored anyway. 

I will first drop that part of my pull request and maybe come back to it. But it would require a change of the model.

Regarding #7467 it don't see a problem anymore, since the UI idea there is to create a toggle next to the week/month buttons, right there my toggle for "past work" would have been, which I will drop.

I am adding a new row with the toggles for each reference calendar, which should not be conflicting the other feature.

johannesjo:
Thank you very much! <3

PR #7501, opened 2026-05-06
chore: remove overly prescriptive structure

zenoprax:
DeepWiki will completely skip the regular doc generation process if any
specific structure is defined in this file. Repo notes should be safe.


johannesjo:
Thank you very much! <3

PR #7503, opened 2026-05-06
Fix #5461: Task dates with delayed start-of-day

### LINKED ISSUE #5461: 🚨 "Start time of the next day" Seems to not be working.
### Version Used

16.2.1

### Expected Behavior

Day should still display as current day rather than next calendar day.

### Current Behavior

Day still progresses at midnight despite being set to 3am.

### Steps to Reproduce

1. Go to settings
2.  Set "Start time of the next day" to a number past 0
3. Wait for midnight 
4. Observe day progress despite setting start time

### Console Output

```shell

```

afonso-resendes:
Fixes #5461

## Problem

With **Start time of the next day** set (e.g. 3 AM), the app should treat the period **between midnight and that hour** as still belonging to the **previous logical day**. In practice, some UI and logic behaved as if the day changed at **calendar midnight**, so tasks could show the wrong day, “Today”, or overdue state during that window.

## Cause

The app already defines logical “today” via `DateService` (`startOfNextDayDiff` / `todayStr()`), but several call sites still used **calendar-based** helpers (e.g. `getDbDateStr()` without the same offset) or ad-hoc `Date` usage. Those paths disagreed with the offset-aware source of truth, so behavior diverged after midnight.

## Fix

- Align comparisons and display logic with the **same logical “today”** as `DateService` / tracking + store where appropriate.
- Add **unit tests** on `DateService` for `todayStr()` with a non-zero start-of-next-day offset (before vs after the configured hour).

## How to verify

1. Set **Start time of the next day** to e.g. `3`.
2. Simulate or wait until local time is **after midnight but before 3 AM**.
3. Confirm scheduled “today” tasks still match that logical day (no spurious next-day / overdue behavior from calendar midnight alone).
4. `npm run test:file src/app/core/date/date.service.spec.ts`


johannesjo:
Thank you very much for the fix! 

The approach is correct and on-pattern – routing through `DateService.todayStr()` / `getLogicalTodayDate()` is exactly what the project's logical-clock convention prescribes for "what day is this?" questions. Approving with a few notes:

### Suggestions

**1. Same bug likely lives in sibling call sites — worth a tracking issue**

The fix is scoped to the add-task-bar, but the same anti-pattern exists in:

- `src/app/features/tasks/task.service.ts:1322` — `dueDay: getDbDateStr()` when adding a task in TODAY context outside the add-task-bar (drag-drop / programmatic paths). Same bug, different entry point.
- `src/app/features/planner/store/planner.effects.ts:27` — `action.day === getDbDateStr()` for the "Planned for Today" snack message; would say a date instead of "Today" between midnight and the cutoff.
- `src/app/features/tasks/task-detail-panel/task-detail-panel.component.ts:288, 303–304, 342` — overdue/today checks.

Fine to leave for a follow-up — could you mention this in the PR body or open a tracking issue so it doesn't get lost?

**2. The "tomorrow" test doesn't actually exercise the offset**

In `add-task-bar-actions.component.spec.ts`:

\`\`\`ts
const tomorrow = new Date(Date.now() - mockDateService.getStartOfNextDayDiffMs());
\`\`\`

…with the spy returning \`0\`, this is equivalent to \`new Date()\`. Consider adding a second tomorrow case with a non-zero offset (mirroring the new "logical today before start of next day" test) to actually cover the new code path.

**3. Pre-existing inconsistency in the same file**

\`add-task-bar-actions.component.ts:141\` (\`repeatQuickOptions\`) still uses raw \`new Date()\` as a fallback — repeat-preset labels can show the wrong day during the offset window. Out of scope here, but worth a comment if intentional.

**4. Minor**

\`\`\`ts
mockDateService.setStartOfNextDayDiff(undefined as unknown as number);
\`\`\`

The double cast is a bit awkward. Either widen the signature to \`number | undefined\` or accept a comment explaining why this branch is defended. Low priority.

### Risk assessment

- Users with \`startOfNextDay = 0\` (default): no behavioral change — \`todayStr()\` reduces to \`getDbDateStr()\`.
- Users with a non-zero offset: correct "today" labeling between midnight and cutoff.
- No NgRx/effects/sync surface area touched. Low blast radius.

LGTM once the tomorrow-test offset is exercised — the rest can be follow-up.

johannesjo:
Thanks for the iteration here — one more thing I caught running the new spec locally before the rest is merged:

**Failing test:** the new `should clamp values above 23 to 23 hours` case in `src/app/core/date/date.service.spec.ts` fails on this branch. Running `npm run test:file src/app/core/date/date.service.spec.ts` reports `Expected 0 to be 82800000`.

The path is:
- `setStartOfNextDayDiff(99)` → `getStartOfNextDayDiffMs(undefined, 99)` → `getValidStartOfNextDayHour(99)` returns `undefined` because `99 > 23` (`src/app/util/start-of-next-day.util.ts:13`) → falls through to `return 0`.

There's no clamping for the numeric path — out-of-range values are *rejected* to 0, not clamped to 23h (`clampStartOfNextDayMinutes` only applies to the `HH:MM` string path). Simplest fix is to align the expectation with the actual behavior:

```ts
it('should treat out-of-range hour values as 0', () => {
  service.setStartOfNextDayDiff(99);
  expect(service.getStartOfNextDayDiffMs()).toBe(0);
});
```

(Changing the util to actually clamp instead would be a broader behavior change — probably out of scope here.)

Two small ones still open from the earlier review, just collecting them:

1. The `tomorrow` test still doesn't exercise the offset — with the mock returning `0`, `new Date(Date.now() - 0)` is just `new Date()`. A tomorrow variant under a non-zero offset (mirroring the new "logical today before start of next day" case) would actually cover the new code path.
2. The sibling call sites (`task.service.ts:1347`, `planner.effects.ts:27`, `task-detail-panel.component.ts:299,314,315`, `issue.service.ts:583`) carry the same calendar-midnight bug. Fine as a follow-up — could you add a note in the PR description or open a tracking issue so it doesn't get lost?

Once the failing test is fixed and the tomorrow test exercises the offset, this is good to go.

johannesjo:
Thank you very much! <3


Two cheap, optional pre-merge nits from a re-review of `52844317` — both non-blocking:

**(a) Guard the jasmine clock in the `isToday` block** — `src/app/core/date/date.service.spec.ts`

The new `should return false for logical tomorrow when offset is set` test installs/uninstalls the clock inline, but it lives in the `isToday` describe block, which has no `afterEach` uninstall (the sibling `todayStr` block does). If the `expect` ever throws, the frozen date leaks into the following `isToday` tests and turns one failure into a misleading cascade. Add the same guard the `todayStr` block uses:

```ts
describe('isToday', () => {
  afterEach(() => jasmine.clock().uninstall());
  // ...and drop the inline uninstall() in the new test
```

**(b) Drop unused mocks from the new component test** — `src/app/features/tasks/add-task-bar/add-task-bar-actions/add-task-bar-actions.component.spec.ts`

`should compute dateDisplay for logical tomorrow before start of next day` stubs `todayStr`, `getStartOfNextDayDiffMs`, and `spyOn(Date, 'now')`, but `dateDisplay()` only reads `getLogicalTodayDate()` (the `getStartOfNextDayDiffMs` stub is circular — used only to derive the value then handed to `getLogicalTodayDate`). The test would keep passing even if `dateDisplay` stopped honoring the offset. The only load-bearing mock is `getLogicalTodayDate`:

```ts
it('should compute dateDisplay for logical tomorrow before start of next day', () => {
  // logical today = 2024-05-19; state.date 2024-05-20 is logical tomorrow
  mockDateService.getLogicalTodayDate.and.returnValue(new Date(2024, 4, 19, 1, 30));
  (mockStateService as any)._mockStateSignal.set({ ...mockState, date: '2024-05-20', time: null });
  fixture.detectChanges();
  expect(component.dateDisplay()).toBe('Tomorrow');
});
```

Both are test-hygiene only — the behavior coverage itself is correct.

johannesjo:
Thank you very much! <3

PR #7504, opened 2026-05-06
Fix #6123: mention list row unreadable with light primary

### LINKED ISSUE #6123: 💥 Tag suggestions always are in white on dark theme
### Steps to Reproduce

1. Set a light primary colour for a project
2. Select a project
3. Press Shift+A and start inputting '#'
4. The suggested tags are always in a white font, no matter the threshold settings.

### Additional Console Output

Don't think it's relevant here, but can provide if helpful.

### Url
file:///C:/Program%20Files/WindowsApps/53707johannesjo.SuperProductivity_16.9.4.0_x64__ch45amy23cdv6/app/resources/app.asar/.tmp/angular-dist/browser/index.html#/tag/TODAY/tasks

### Meta Info
META: SP16.9.4 __ Electron – en-US – Win32 – en-US – UA:Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) superProductivity/16.9.4 Chrome/138.0.7204.251 Electron/37.10.3 Safari/537.36

### Actions Before Error
```

```

HenriqueCDutra:
### Problem

When adding a task (e.g. **Shift+A**) and typing **`#`** to pick a tag, the mention dropdown highlights the active row with **`background: var(--c-primary)`** (the **project** primary colour).

That row used **`color: var(--text-color-most-intense)`**, which is tuned for general UI surfaces (e.g. strong body text in a dark app theme). It is **not** the colour meant for text **on top of the primary** swatch.

If the project primary is **very light** (and/or the app theme uses light “intense” text), the highlighted row can end up with **light text on a light background**, so the tag label is effectively invisible. Project contrast/threshold settings apply to the primary palette (via **`--c-contrast`** / `--palette-primary-contrast-*`), but they were not used for this row.

### Fix

- **`src/styles/components/mentions.scss`** — For `.mention-active > a`, set **`color: var(--c-contrast)`** instead of **`var(--text-color-most-intense)`**, so label colour tracks **contrast against `--c-primary`** (including light primaries).
- **`src/app/ui/mentions/mention-list.component.scss`** — Replace hardcoded Bootstrap-like colours with **design tokens** (`--text-color`, `--bg-lightest`, `--c-primary`, `--c-contrast`, borders, radius) so defaults stay consistent with the global theme.

### Tests

- **`src/app/ui/mentions/mention-list.component.contrast.spec.ts`** — Sets CSS variables on the component host (light primary, dark contrast, etc.) and asserts **`getComputedStyle`** on the active vs inactive rows so this regression does not come back.

### How to verify manually

1. Set a **light primary** on a project and select that project.
2. **Shift+A**, type **`#`**.
3. The **highlighted** suggestion row should show **readable** text on the primary background across themes.

### DIFF
--- a/src/app/ui/mentions/mention-list.component.contrast.spec.ts
+++ b/src/app/ui/mentions/mention-list.component.contrast.spec.ts
@@ -1,0 +1,75 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { CommonModule } from '@angular/common';
+
+import { MentionListComponent } from './mention-list.component';
+import { Log } from '../../core/log';
+
+/**
+ * Regression #6123: active mention row must use --c-contrast on --c-primary.
+ * Kept in a separate spec so the first detectChanges() runs after items are set
+ * (avoids ExpressionChangedAfterItHasBeenCheckedError from the shared component spec).
+ * Global mentions.scss is loaded via src/styles.scss in Karma.
+ */
+describe('MentionListComponent contrast (regression #6123)', () => {
+  let component: MentionListComponent;
+  let fixture: ComponentFixture<MentionListComponent>;
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      imports: [MentionListComponent, CommonModule],
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(MentionListComponent);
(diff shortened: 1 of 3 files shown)


johannesjo:
Thank you very much! <3

PR #7523, opened 2026-05-08
feat(sync): add OneDrive sync provider with PKCE auth

### LINKED ISSUE #6984: 💡 Sync with Microsoft
### Problem Statement

Please allow using M365 as sync server. 
In best case I could add also Nextcloud and setup a double sync. For Tasks with e.g. a certain category. So I could manage it with Outlook and over Nextcloud.

This would allow me to work better in my Linux only home office 

### Possible Solution

I do not know what ways exist to sync such data. AI sais that Microsoft Graph would be the API for that.

https://caldavsynchronizer.org But this need an installed Outlook and does not work with the webapp.

### Additional context

_No response_

starter727:
## Summary

Add OneDrive (Microsoft Graph) as a new file-based sync provider, using PKCE OAuth flow and integrated into the existing operation log sync framework alongside Dropbox/WebDAV/Nextcloud.

## Changes

- **New `OneDriveFileSyncProvider`** (`src/app/op-log/sync-providers/file-based/onedrive/onedrive.ts`) - implements the `FileSyncProvider` interface for reading/writing `sync-data.json` via Microsoft Graph API
- **PKCE OAuth flow** - OAuth state validation, token refresh with concurrency control, manual code paste fallback on all platforms
- **ETag-based optimistic locking** - uses file ETag to prevent race conditions on upload (same pattern as Dropbox provider)
- **OAuth state cleanup** - prunes expired OAuth states on-demand instead of polling
- **Sync form config** - OneDrive section in config page with client ID/secret fields, custom app toggle
- **Unit tests** (`onedrive.spec.ts`) - 260 lines covering auth, file ops, and edge cases

## Implementation approach

Builds on the existing `FileSyncProvider` architecture. OneDrive is classified as a file-based provider in `FILE_BASED_PROVIDER_IDS`, so it benefits from all existing sync infrastructure: operation capture, conflict resolution (LWW), encryption, and backup/restore.

## Tested

- Windows <-> Linux cross-platform sync
- OAuth PKCE flow (both automatic callback and manual code paste)
- Token refresh and expiry handling
- 9069 unit tests pass

## Known limitations

- **Mobile untested**: iOS/Android WebView OAuth redirect flow not verified
- **No official client ID**: users must register their own Azure AD app. Awaiting project maintainer for a default Microsoft client ID

Closes #6984


johannesjo (commented):
Thanks for the contribution\! Solid integration overall — reuse of `FileSyncProvider` is clean, the App Folder scope is a good security default, and the manual-code fallback covers all platforms. Below are findings from a code review pass; I've grouped by severity and flagged the spots that I think need changes before merge.

## Critical

**Refresh-token endpoint errors don't clear credentials**
`onedrive.ts` `_requestOAuthToken` (around line 522) throws a generic `HttpNotOkAPIError` for any non-OK response. When Microsoft returns `400 invalid_grant` (refresh token revoked — e.g. user changed their password, signed out elsewhere, or the token simply aged out past its sliding window), this bubbles up as a generic HTTP error and **the bad refresh token is never cleared**. The next sync attempt re-uses the same bad token and fails again, indefinitely, with no way to recover except wiping credentials manually. Suggest detecting `400/401` from the token endpoint specifically (Graph returns `error: "invalid_grant"`) and calling `clearAuthCredentials()` so the user gets prompted to re-auth.

**`clearAuthCredentials()` race vs in-flight refresh silently restores tokens**
`_refreshAccessTokenIfNeeded` (lines 433–473) captures `cfg` by closure, then on success does `setComplete({ ...cfg, accessToken: NEW, refreshToken: tokenData.refresh_token || cfg.refreshToken })`. If a concurrent request hits a 401 and calls `clearAuthCredentials()` while the refresh is in flight, the cleared state gets overwritten when the refresh resolves and writes back the *closure-captured* stale `refreshToken`. Net effect: a 401 that should have forced re-auth silently keeps the user logged in with stale-but-still-valid tokens. Fix options: re-load cfg before `setComplete`, or have `clearAuthCredentials()` invalidate `_tokenRefreshInFlightPromise` so the resolved value is discarded.

## Important

**Auth code logged via `SyncLog.log`**
`oauth-callback-handler.service.ts` (`_setupElectronOAuthListener`): `SyncLog.log('OAuthCallbackHandler: Received Electron OAuth callback URL', callbackUrl)` writes the full callback URL — including the OAuth `code` and `state` — to the log. Per CLAUDE.md (rule 9), log history is exportable, so this should never include secrets. Either redact (`callbackUrl.split('?')[0]`) or drop the second arg.

**`validateOneDriveOAuthState` exported from the provider creates a circular dependency**
`oauth-callback-handler.service.ts` imports from `op-log/sync-providers/file-based/onedrive/onedrive`, while the OneDrive provider sits downstream of the OAuth handler conceptually. Move the OAuth state map + validate function into a dedicated `oauth-state.util.ts` keyed by provider, so the handler doesn't need to know about specific providers and future OAuth providers can reuse it.

**Web build uses Microsoft's `nativeclient` redirect URI**
`onedrive.ts` `ONEDRIVE_PROTOCOL.redirectUri = 'https://login.microsoftonline.com/common/oauth2/nativeclient'`. On the browser build (`\!IS_ELECTRON && \!isNativePlatform`), this is the redirect URI sent to Entra. Microsoft Entra typically rejects `nativeclient` for app registrations of type "Web" / "SPA" (AADSTS50011). Worth verifying with a real signup in the browser build, or documenting that web users must register an Entra "Mobile and desktop applications" app.

**Token-error retry asymmetry vs Dropbox**
On a 401, `_request` clears credentials and throws immediately. The Dropbox provider, by contrast, refreshes and retries once before giving up. Clock skew or transient revocations will hurt OneDrive users disproportionately. Consider one retry post-refresh on 401 to match the Dropbox UX, then clear-and-throw if the retry also fails.

**Dead `oneDriveAuth: true` prop and unused translation keys**
`sync-form.const.ts`: `props: { oneDriveAuth: true } as any` is set on a panel but nothing reads it (grepped). The actual auth flow runs implicitly via `dialog-sync-cfg.configuredAuthForSyncProviderIfNecessary()`. As a result, all of `BTN_AUTHENTICATE`, `BTN_REAUTHENTICATE`, `STATUS_CONFIGURED`, `STATUS_NOT_CONFIGURED`, `AUTH_SUCCESS`, `REAUTH_SUCCESS` translation keys are added to `en.json` but never displayed. Either wire a custom Formly type that renders auth status + button using these keys, or drop the dead prop and the unused translations.

**`_cfgOrError(requireAuth)` parameter is dead**
`onedrive.ts:378`: every call site uses the default `false`. Either remove the parameter or pass `true` from `_request` so missing-refresh-token cases fail loudly at the cfg-check step instead of mid-fetch.

**Folder-ensure cache not invalidated on `syncFolderPath` change**
`_ensuredFolderPath` is only nulled on a 404. If the user edits `syncFolderPath` in settings while the app stays open, the cache still claims the old path is ensured and uploads to the new path will skip folder creation. Either key the cache by the resolved path *and* recheck on cfg change, or invalidate via the cfg-update path.

## Minor

- **Logging file/folder paths**: `SyncLog.warn(`[OneDrive] Request failed status=… path=${options.path}`, …)` and the Graph error `message` can include user-entered folder names. Project rule 9: "never log user content." Prefer `Log.log({ status, code })` only.
- **`_normalizeAuthCodeInput` regex** (`dialog-get-and-enter-auth-code.component.ts`): `/(?:^|[?&#])code=([^&#]+)/i` accepts any pasted text containing `code=`. PKCE binding makes an attacker-supplied code unusable (it won't match the user's verifier), but worth a comment to make the defence-in-depth reasoning explicit.
- **State map memory**: `OAUTH_STATES_MAP` is module-scope and only pruned on `getAuthHelper()` / `validateOneDriveOAuthState()`. Bounded leak (10-min entry TTL), trivial in practice.
- **Test isolation**: `(globalThis as any).fetch = fetchSpy` in `beforeEach` is never restored in `afterEach`. Use a `try/finally` or `afterEach`. Also note the module-level `OAUTH_STATES_MAP` would leak across tests if state-validation tests are added later — clear it in `beforeEach`.
- **Test coverage gaps**: state validation (CSRF), `_normalizeAuthCodeInput` URL parsing, concurrent `_refreshAccessTokenIfNeeded` deduplication, and 412 → `UploadRevToMatchMismatchAPIError` aren't covered. The existing 9 tests are solid for the happy paths but miss the more interesting failure modes.
- **`_resolveClientId` migration logic**: the three-branch precedence (custom-true / custom-false / undefined-legacy) is correct as far as I traced it, but consider a small spec to lock it in given how subtle the legacy fallback is.
- **`getFileRev` vs `downloadFile` rev fallback**: `downloadFile` returns `rev: ''` when no ETag is available; `getFileRev` throws `RemoteFileNotFoundAPIError`. Inconsistent — if no rev means "file unknown" then both should throw; if it means "no rev tracked" then both should return empty. Pick one.
- **Dialog provider filter** (`dialog-get-and-enter-auth-code.component.ts:80`): `data.providerName.toLowerCase()` works but relies on `provider.id` being a stable string. An enum-based filter would be sturdier.
- **`'props.required'` expression**: when toggling between official and custom-app modes, the `clientId` field's `required` flag updates async. In practice probably fine, but worth a manual test of rapid toggling.
- **Style nit**: `...{ ['@microsoft.graph.conflictBehavior']: 'replace' }` — just write it as a quoted property key directly.

## Strengths

- Token refresh deduplication via `_tokenRefreshInFlightPromise` and folder-ensure caching are thoughtful.
- App Folder scope (`Files.ReadWrite.AppFolder`) is the right minimum-permission default.
- PKCE + state parameter for CSRF on the auto-callback path is correctly implemented.
- Error mapping (`_mapAndThrow`) bridges Graph status codes to the project's error taxonomy cleanly.
- Manual-paste fallback works on every platform, which is important for the "no official client ID yet" period.

## Verdict

The architecture is right. The two Critical items (refresh endpoint error handling and the clear-vs-refresh race) are the things I'd want fixed before merge — both are scenarios where the user gets stuck without a clear recovery path. The Important items are mostly cleanup but the `nativeclient` redirect URI for the web build is worth a real-world test. Everything else is polish.

johannesjo:
Thank you very much!

Did a multi-agent pass (correctness/security/architecture/alternatives/performance/simplicity). i18n findings excluded per author request. Grouped by severity below.

## Critical — block on these

### 1. `_is401Retry` is an instance bool racing `maxConcurrentRequests = 4`
`onedrive.ts:1232,1720-1737` — Concurrent 401s: req A sets flag and retries; req B sees flag=true, skips refresh, falls straight to `clearAuthCredentials() + throw AuthFailSPError`. User gets logged out on any transient revocation or clock skew that affects 2+ requests in flight.

Fix: drop the field, pass per-call retry state:
```ts
private async _request(options: ApiRequestOptions, isRetry = false): Promise<Response> {
  ...
  if (response.status === 401 && \!isRetry) {
    try {
      await this._refreshAccessTokenIfNeeded({ ...currentCfg, tokenExpiresAt: 0 });
      return this._request(options, true);
    } catch { /* fall through to clear */ }
  }
  ...
}
```
`_tokenRefreshInFlightPromise` already dedupes the actual refresh, so two simultaneous retries cost one network call.

### 2. `_requestOAuthToken` swallows the `MissingRefreshTokenAPIError` it just threw
`onedrive.ts:1665-1681` — `throw new MissingRefreshTokenAPIError()` lives inside the same `try` that catches `JSON.parse` errors. The same `catch (_parseErr)` swallows it; caller sees generic `HttpNotOkAPIError` instead. Refresh paths that branch on error type misbehave.

Fix: only wrap the parse call.
```ts
let parsed: { error?: string } | null = null;
try { parsed = JSON.parse(body); } catch { /* not JSON */ }
if (parsed?.error === 'invalid_grant') {
  await this.clearAuthCredentials();
  throw new MissingRefreshTokenAPIError();
}
throw new HttpNotOkAPIError(response, body);
```

### 3. Electron deep-link callback has no state binding for non-OneDrive providers
`electron/protocol-handler.ts:37-43` + `oauth-callback-handler.service.ts:559-607` — `_parseOAuthCallback` validates state only when `providerRaw === 'onedrive'`. Any local process can fire `superproductivity://oauth-callback/dropbox?code=ATTACKER_CODE` and the Dropbox dialog accepts it. Pre-existing weakness amplified by this PR opening the deep-link surface.

Fix: require + validate state for every provider routed through the deep link. Reject callbacks when no auth flow is pending.

### 4. `_ensureSyncFolderExists` uses `conflictBehavior: 'replace'` on folder POST
`onedrive.ts:1493-1495` — If a **file** with the same name exists at the parent level, Graph replaces it with the folder (data loss). The 409→`nameAlreadyExists` catch only covers the folder-exists case.

Fix: `'fail'`, keep the 409 swallow.

## Important — fix before merge

### 5. Refresh-during-clear race not fully closed
`onedrive.ts:1251,1591-1611` — `clearAuthCredentials()` nulls `_tokenRefreshInFlightPromise` but the IIFE still completes and writes via `setComplete()`. The `currentCfg?.refreshToken` guard mitigates but a window remains between `load()` and `setComplete()`. Have `clearAuthCredentials` await the in-flight promise (then null) before zeroing creds.

### 6. `_parseOAuthCallback` defaults unknown provider to `'dropbox'`
`oauth-callback-handler.service.ts:586-587,597-598` — Misroutes malformed callbacks to Dropbox dialog forever. Return `'unknown'` and reject, or surface a parse error.

### 7. Duplicated pre-auth save block in `dialog-sync-cfg`
`dialog-sync-cfg.component.ts:430-441` and `:480-493` — Two identical 12-line copy-pasted blocks with `as any` + eslint-disable. Extract `_persistOneDriveFormCfgBeforeAuth(providerId)`. Better: move the responsibility into `configuredAuthForSyncProviderIfNecessary` as a generic pre-auth flush hook so future providers don't repeat this.

### 8. `_request` double-loads cfg on 401 retry; 401 path bypasses refresh dedup
`onedrive.ts:1695,1724` — Reuse the `cfg` from the outer `_cfgOrError()` call and route the forced refresh through `_tokenRefreshInFlightPromise` instead of forking a parallel `{...cfg, tokenExpiresAt: 0}` path.

### 9. `_ensureSyncFolderExists` makes N serial POSTs on every cache miss
`onedrive.ts:1471-1508` — Common case (folder already exists) hits N×409. Add one `GET /me/drive/special/approot:/<fullPath>:` probe; only fall back to per-segment create on 404. Collapses N→1 round-trip.

### 10. `onedrive.ts` is 652 lines (>2× the 300-line rule)
Split: `onedrive-auth.ts` (PKCE/OAuth), `onedrive-http.ts` (`_request`, retry, error mapping), `onedrive-paths.ts` (path encoding), `onedrive-folders.ts` (ensure folder), `onedrive.ts` (FileSyncProvider methods).

### 11. Token-endpoint error body may include `code_verifier` / `code` / `refresh_token`
`onedrive.ts:1682` — `new HttpNotOkAPIError(response, body)` carries the raw body upstream. Microsoft 400 bodies sometimes echo request params. Redact `code`, `code_verifier`, `refresh_token` before constructing the error.

### 12. `_mapAndThrow` propagates user-supplied `targetPath` in error messages
`onedrive.ts:1797-1809` + `_request` line 1747 — `HttpNotOkAPIError(response, responseBody)` includes Graph error body (echoes user-supplied folder/file names). Violates rule 9 (never log user content). Strip path/name from error messages; keep `{ status, code }` only.

### 13. Manual-paste flow has no state validation
`dialog-get-and-enter-auth-code.component.ts:_normalizeAuthCodeInput` — PKCE protects token confidentiality but the CSRF defense via `state` is silently bypassed on web/native paste. Extract `state` from pasted URL and validate against `OAUTH_STATES_MAP`, or document the gap.

### 14. `OAuthCallbackData.provider` string union diverges from `SyncProviderId` enum
`oauth-callback-handler.service.ts:5-12` + `oauth-state.util.ts` — Lowercase strings bridged via `.toLowerCase()` at consumer sites. Type as `SyncProviderId`, normalize once at parse boundary.

### 15. Electron IPC listener never removed
`oauth-callback-handler.service.ts:542-557, 38-41` — Capacitor handle removed in `ngOnDestroy`; Electron handler isn't. Singleton makes it bounded but fragile. Match Capacitor cleanup.

## Minor

- `sync-form.const.ts:226-234` — Empty `authStatus` Formly field renders dead `<p>`. Wire to live status or delete.
- `onedrive.ts:1234` + `sync-providers.factory.ts:1867-1869` — `_devPath` constructor param leaks dev-only path math into runtime; `approot` already DEV-segregated by dev client ID. Delete.
- `global-config.model.ts:142-149` vs `onedrive.model.ts:858-866` — `OneDriveConfig` (vm) and `OneDrivePrivateCfg` (storage) duplicate `clientId`/`tenantId`/`syncFolderPath`/`useCustomApp`. Drift risk. Make vm `Pick<OneDrivePrivateCfg, ...>`.
- `onedrive.ts:1529-1540` — `_resolveClientId` legacy `undefined` branch is unreachable for a brand-new provider (defaults always set `useCustomApp` to boolean). Delete the 3rd branch.
- `onedrive.ts:1228-1230,1444-1462` — `_ensuredFolderPath` + `_folderEnsureInFlightPath` + `_folderEnsureInFlightPromise` collapse to one `Map<string, Promise<void>>`.
- `dialog-get-and-enter-auth-code.component.ts:_normalizeAuthCodeInput` — Regex fallback can be replaced with a single `new URL(input, 'http://x/?')` pass for uniform parsing of full URLs, bare query strings, and fragments.
- `onedrive.ts:1486-1490,1655-1659` — IIFE `Headers` builders → `new Headers({ 'Content-Type': '…' })` literal.
- `t.const.ts:1230-1231` + `dialog-sync-cfg.component.ts:208,506` — Moving `BTN_REAUTHENTICATE`/`REAUTH_SUCCESS` out of `DROPBOX` namespace into generic `FORM` is a provider-agnostic refactor outside OneDrive scope. Split into its own commit/PR.
- `main-header.component.ts:225-232` — `onSyncButtonClick` extraction unrelated to OneDrive. Revert to keep diff focused.
- `onedrive.spec.ts:beforeEach` — `(globalThis as any).fetch = fetchSpy`; `afterEach` resets to `undefined` not original. Use `try/finally` with the captured original.
- Test gaps: state CSRF validation, `_normalizeAuthCodeInput` URL parsing, concurrent `_refreshAccessTokenIfNeeded` dedup, 412 → `UploadRevToMatchMismatchAPIError`, 400/invalid_grant → clears creds, 401-retry path. The interesting failure modes added since v1 deserve coverage.
- `oauth-callback-handler.service.ts:542-557` — `_setupElectronOAuthListener` should validate `payload.url` starts with `superproductivity://` before parsing — defense-in-depth against future IPC senders wired to the same channel.

## What looks good

- PKCE + state extraction into provider-agnostic `oauth-state.util.ts` ✓
- `invalid_grant` now clears credentials ✓
- `_tokenRefreshInFlightPromise` dedups refreshes (steady-state) ✓
- Folder cache keyed by resolved path with explanatory comment ✓
- AppFolder scope (`Files.ReadWrite.AppFolder`) is correct minimum ✓
- `_normalizeAuthCodeInput` defense-in-depth comment ✓
- `setComplete` re-reads cfg before write ✓

Verdict: **Needs changes.** Block on items 1–4; strong recommend on 5–15 before merge.

johannesjo:
Thank you very much for the follow-up on item #2. 

The patch closes the documented swallow, but I'd push for a slightly different shape before merge – same line count, fewer footguns. Six-reviewer pass converged on this independently.

### Findings on `b1480173`

**1. Residual swallow on `clearAuthCredentials()` rejection** — `onedrive.ts:500-510`
`clearAuthCredentials()` awaits `privateCfg.load()` + `setComplete()` (storage I/O) inside the same `try`. If either rejects, the catch binds a non-`MissingRefreshTokenAPIError`, the new `instanceof` skips, control falls to `throw new HttpNotOkAPIError(response, body)`. The storage failure is lost *and* the revoked-token signal is replaced with a misleading generic HTTP 400. Same shape as the original bug, just one error class over.

**2. Pattern outlier in the package** — `onedrive.ts:491-509`
The narrowed-`try` idiom is already the convention in three nearby spots: `_parseGraphError` 100 lines below in this same file (`onedrive.ts:594-610`), `dropbox-api.ts:814-819`, and `super-sync.ts:514-531`. Using `catch` as an `instanceof` router is unique to this block.

**3. Footgun for future contributors** — `onedrive.ts:504`
Adding any new typed throw inside the `try` (e.g. handling `invalid_client`, `interaction_required`, `unauthorized_client`) silently rewrites to `HttpNotOkAPIError` unless someone remembers to extend the `instanceof` chain. Same hazard the patch was written to fix.

**4. `_parseErr` underscore is now lying** — `onedrive.ts:503-505`
Lint doesn't enforce it (`eslint.config.js: caughtErrors: 'none'`), but the leading underscore signals "intentionally unused" and is now used twice.

**5. Missing regression test**
Per CLAUDE.md ("write tests for bug fixes first"). One spec — `400 + {error:'invalid_grant'}` → `MissingRefreshTokenAPIError` — would have caught the original swallow and locks the contract against any of the above drifting.

### Suggested replacement (matches `_parseGraphError`, dropbox-api, super-sync)

\`\`\`ts
if (response.status === 400) {
  let parsed: { error?: string; error_description?: string } | undefined;
  try {
    parsed = JSON.parse(body);
  } catch {
    /* not JSON */
  }
  if (parsed?.error === 'invalid_grant') {
    this._deps.logger.warn(
      '[OneDrive] Refresh token revoked (invalid_grant), clearing credentials',
    );
    await this.clearAuthCredentials();
    throw new MissingRefreshTokenAPIError();
  }
}
throw new HttpNotOkAPIError(response, body);
\`\`\`

Same net line count. Removes the `instanceof` router. `clearAuthCredentials()` sits outside the `try`, so its rejection propagates honestly. Future typed throws can't be silently swallowed because nothing inside the `try` can throw except `JSON.parse`.

### Not regressions

- Security: net positive on the `invalid_grant` path — `MissingRefreshTokenAPIError` carries no body, vs `HttpNotOkAPIError(response, body)` which retained Entra `error_description` / `correlation_id`.
- Performance: nil. Cold OAuth-error path.

Items 1, 3, 4 from the prior Critical list (concurrent-401 `_is401Retry` race, deep-link state binding, `conflictBehavior: 'replace'` on folder POST) still open on the branch.

johannesjo (commented):
Follow-up on commits `b11ebd58..82fb5a1e` (the third-review delta). Glad to see the deadlock untangled — a couple of small follow-ups, none blocking.

## Important

**`clearAuthCredentials()` comment over-promises**
`packages/sync-providers/src/file-based/onedrive/onedrive.ts:86-101` — the comment claims "The IIFE's `currentCfg.refreshToken` guard already prevents stale writes." True for the invalid_grant path (the IIFE throws before its `setComplete`), but for an *external* concurrent `clearAuthCredentials()` (e.g. user hits Disconnect mid-sync) there's still a narrow microtask-level TOCTOU window between the IIFE's `await this.privateCfg.load()` and `await this.privateCfg.setComplete(updatedCfg)` (`onedrive.ts:429` → `:444`). If `clearAuthCredentials()`'s `setComplete` lands between those, the IIFE will overwrite the cleared cfg with a fresh token. Risk is low in practice but the comment reads as a stronger guarantee than the code provides. Options: (a) capture a generation sentinel that `clearAuthCredentials` increments and the IIFE checks before `setComplete`, or (b) re-load and re-verify `refreshToken` immediately before `setComplete`. At minimum, soften the comment.

**Folder-cache test name no longer matches the assertion**
`src/app/op-log/sync-providers/file-based/onedrive/onedrive.spec.ts:225-274` — the test is named *"should avoid repeated folder existence checks after first successful upload"* but now only asserts `postCount === 0`. Nothing prevents the implementation from issuing a fresh GET probe on every upload and the test would still pass. Add a `getCount` counter and assert it is `1` so the test actually covers what its name claims (cache hit on subsequent uploads), or rename the test.

## Minor

- **Electron scheme gate vs native** — `src/app/imex/sync/oauth-callback-handler.service.ts:94` uses `startsWith('superproductivity://')`, while the native listener (lines 56-57) uses the tighter `startsWith('superproductivity://oauth-callback')`. Consider matching the native path.
- **Rejection log lacks context** — `oauth-callback-handler.service.ts:95-97`: the rejection warning has no detail. Including the scheme (e.g. `callbackUrl.split(':')[0]`) would aid debugging without leaking secrets. Schemes are not user content.
- **Redundant per-test stub** — `onedrive.spec.ts:352` `setComplete.and.resolveTo()` is now redundant given the new `beforeEach` default at line 57.
- **Unconditional null before `setComplete`** — `onedrive.ts:94`: `_tokenRefreshInFlightPromise = null` is set before `setComplete` is awaited. If `setComplete` rejects, the in-flight reference has already been discarded and the next request will start a fresh refresh against half-cleared state. Probably fine (the in-flight token would also be stale) but worth a note.

## Strengths

- Deadlock diagnosis is right: the prior `await refreshPromise` inside `clearAuthCredentials` self-awaited on the invalid_grant path. Minimal, on-target fix.
- Electron scheme gate is sensible defense-in-depth at the IPC boundary.
- The `MissingRefreshTokenAPIError` assertion now matches the documented contract (cleared creds + typed error) rather than the leaky `HttpNotOkAPIError`.
- `setComplete.and.resolveTo()` default in `beforeEach` is correct hygiene.
- Two well-scoped commits with messages that explain the *why*.

## Verdict

Delta itself is good to merge. Note that it doesn't address the remaining open items from my earlier review (generic 401 on the token endpoint, `nativeclient` redirect URI for the web build, the dead `oneDriveAuth: true` prop / unused translation keys, log redaction for paths). Those are still on the table for the PR as a whole.

johannesjo:
Thanks for the follow-ups. I rechecked current head `0fb62ab` with a KISS/YAGNI lens; these are the items I’d still want addressed before merge.

1. **Non-force first upload can overwrite a concurrently-created sync file**  
   `packages/sync-providers/src/file-based/onedrive/onedrive.ts:189` only sends `If-Match` when `revToMatch` is truthy. For `revToMatch === null`, the Graph `PUT ...:/content` call is unconditional; Microsoft documents this endpoint as “upload or replace” ([docs](https://learn.microsoft.com/en-us/graph/api/driveitem-put-content?view=graph-rest-1.0)). Two fresh clients can both observe no file, then the later upload silently replaces the earlier `sync-data.json`. Please use fail-if-exists/create-conflict semantics for the null-rev path and add a regression test.

2. **Bad reauth code can clear valid existing credentials**  
   `packages/sync-providers/src/file-based/onedrive/onedrive.ts:507-512` clears credentials for every `invalid_grant`, but `_requestOAuthToken()` is also used for authorization-code exchange. A mistyped/expired pasted code during reauth should surface as “invalid auth code”, not delete the existing refresh token. Keep credential clearing limited to `req.grantType === refresh_token`.

3. **OAuth callback URLs are logged with `code`/`state`**  
   `electron/protocol-handler.ts:14` and `:129` log full protocol URLs. OAuth authorization codes are short-lived but still credentials; state should also not land in logs. Please redact query params before logging.

4. **Web OneDrive auth uses a native redirect with browser token exchange**  
   `packages/sync-providers/src/file-based/onedrive/onedrive.ts:530` uses `nativeclient` for all non-Electron builds, then redeems the code via browser `fetch`. Microsoft’s auth-code docs require SPA redirect/CORS setup for browser-origin token redemption ([docs](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow)). Either gate OneDrive auth away from web for now or use a proper SPA/web callback flow.

5. **New package subpath is missing TS path aliases**  
   `@sp/sync-providers/onedrive` was added to package exports, but not to `tsconfig.base.json` or `src/tsconfig.spec.json`. The rest of the sync-provider subpaths are mapped to source there; please add the OneDrive alias too so local app/spec builds don’t depend on prebuilt package artifacts.

6. **Sync public config now writes `undefined` over existing settings**  
   `src/app/imex/sync/sync-config.service.ts:323-330` constructs `globalConfig` with optional fields even when they are `undefined`; the reducer spreads those over the existing section. This can clear persisted `isCompressionEnabled` / `isManualSyncOnly` during unrelated saves. Please omit undefined keys or restore the previous rest-destructure behavior.

7. **Provider switching can overwrite saved OneDrive form config with defaults**  
   `src/app/imex/sync/dialog-sync-cfg/dialog-sync-cfg.component.ts:358-384` loads private config for several providers when switching, but not OneDrive. `_persistOneDriveFormCfgBeforeAuth()` then writes `_tmpUpdatedCfg.oneDrive` directly over existing private config. Switching back to OneDrive can replace saved `clientId` / folder settings with defaults while preserving old tokens. The simplest fix is to handle OneDrive in the existing provider-switch branch.

I’d avoid larger refactors here. The above can be fixed with small targeted changes plus focused regression tests for the upload race and `invalid_grant` behavior.

johannesjo:
Thank you very much for your ongoing work on this! <3

I rechecked the latest PR head (`ec0c3dce31ae`). I’d still request changes before merge. The remaining issues are mostly small, but several affect sync correctness or credential safety.

1. **Initial OneDrive upload can overwrite concurrent remote data**  
   `packages/sync-providers/src/file-based/onedrive/onedrive.ts:189` only sends `If-Match` when `revToMatch` is truthy. For first-create uploads (`revToMatch === null`), Graph `PUT .../content` is unconditional. Microsoft documents `@microsoft.graph.conflictBehavior` as defaulting to `replace` for PUT, so two fresh clients can race and the later one silently replaces the first sync file. Please use create-only conflict behavior for the null-rev path, e.g. `@microsoft.graph.conflictBehavior=fail`, and map that conflict to `UploadRevToMatchMismatchAPIError`.  
   Ref: https://learn.microsoft.com/en-us/graph/api/resources/driveitem?view=graph-rest-1.0#instance-attributes

2. **In-flight refresh can overwrite or clear newer credentials**  
   `onedrive.ts:432-447` writes refreshed tokens into whatever config is current as long as it has any refresh token. `onedrive.ts:512` clears credentials on `invalid_grant` without checking whether those are still the credentials used for that refresh request. If the user disconnects/re-auths while a refresh is pending, stale refresh work can wipe or overwrite the new config. Please guard success and clear paths by matching `refreshToken`, `clientId`, and `tenantId` against the cfg used for the refresh.

3. **Web auth uses nativeclient redirect with browser token exchange**  
   `onedrive.ts:531-535` uses the native client redirect URI for all non-Electron builds, while the token exchange runs in browser `fetch`. Microsoft’s auth-code flow docs require SPA redirect/CORS setup for browser-origin token redemption. Since OneDrive is exposed in the web provider list (`sync-form.const.ts:156`), please either gate OneDrive auth away from web for now or implement a proper SPA redirect flow.  
   Ref: https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow#redirect-uris-for-single-page-apps-spas

4. **Graph 401 retry clears credentials on transient refresh errors**  
   `onedrive.ts:570-582` catches every forced-refresh failure after a Graph 401 and then clears credentials. A transient token endpoint 429/5xx would log the user out. Please only clear for permanent auth failures (`invalid_grant`, missing/revoked refresh token) and rethrow transient refresh failures.

5. **Saving `syncProvider: null` rewrites it to WebDAV**  
   `src/app/imex/sync/sync-config.service.ts:327` uses `newSettings.syncProvider ?? SyncProviderId.WebDAV`. `null` is the valid disabled/no-provider state, so this can persist WebDAV after saving disabled sync. Please preserve `null`.

6. **Switching back to OneDrive does not hydrate the OneDrive form config**  
   `src/app/imex/sync/dialog-sync-cfg/dialog-sync-cfg.component.ts:384-388` only copies `encryptKey` for OneDrive, unlike the other provider branches. The form can show stale/default client/tenant/path values, and `_persistOneDriveFormCfgBeforeAuth()` can then merge those over the saved private config. Please set `oneDrive: privateCfg` here.

7. **Duplicate Formly controls for sync interval/manual-only**  
   `src/app/features/config/form-cfgs/sync-form.const.ts:432-459` and `:531-555` define `syncInterval` and `isManualSyncOnly` twice for non-SuperSync providers. Same-key controls plus `resetOnHide` can fight each other. Please keep one source of truth.

8. **Locale files changed outside translation workflow**  
   `src/assets/i18n/zh.json` and `src/assets/i18n/zh-tw.json` are edited, but `CLAUDE.md` says to edit only `en.json`. Please revert non-English locale changes.

A few smaller follow-ups worth addressing in the same pass:
- `electron/protocol-handler.ts:15` and `:131` only redact `?`, not `#`; OAuth params in fragments would still be logged.
- Manual pasted OneDrive callback URLs should require valid state for parsed callback URLs, including fragment params.
- `OneDrive.listFiles()` currently swallows all errors as `[]`; it should only treat not-found as empty and rethrow auth/rate-limit/server failures.

johannesjo:
Thanks for the latest pass. I double-checked current head `b7c0bf5`; I'd still request changes on these items.

1. **Stale OneDrive refresh can still clear newer credentials**
   `packages/sync-providers/src/file-based/onedrive/onedrive.ts:439-448` correctly detects that the current config no longer matches the refresh attempt, but it throws `MissingRefreshTokenAPIError`. The 401 retry path then treats that as a permanent auth failure and clears credentials at `:591-596`. Also, token endpoint `invalid_grant` / `401` still calls `clearAuthCredentials()` directly at `:523` and `:532` without checking whether the current config still matches the refresh attempt. If the user re-auths or edits OneDrive config while an old refresh is in flight, the old request can wipe the newer credentials. Please only clear when current `refreshToken` / `clientId` / `tenantId` still match the failed refresh input, or use a distinct stale-refresh-discard error that does not clear credentials.

2. **`listFiles()` no longer returns `[]` for missing folders**
   `onedrive.ts:231-236` checks for `RemoteFileNotFoundAPIError`, but `_requestJson()` throws `HttpNotOkAPIError` for a Graph 404. The 404-to-`RemoteFileNotFoundAPIError` conversion is in `_mapAndThrow()` at `:648-652`, but `listFiles()` never calls it. A missing OneDrive folder now escapes as raw HTTP error instead of returning an empty list. Please map the error here or explicitly handle `HttpNotOkAPIError` with status 404.

3. **The locale revert deletes unrelated existing translations**
   `src/assets/i18n/zh.json:1233` and `src/assets/i18n/zh-tw.json:1264` now close `F.SYNC.FORM` after `WEB_DAV`, removing existing `NEXTCLOUD` translations while `sync-form.const.ts:272-306` still reads `T.F.SYNC.FORM.NEXTCLOUD.*`. The diff against the PR base also removes many unrelated non-OneDrive keys. Please restore the locale files from the PR base and only remove the PR-added OneDrive translations.

Non-blocking: I rechecked the OneDrive web gating. Hiding OneDrive from the web select seems reasonable if browser auth is intentionally unsupported, but persisted/imported web configs with `syncProvider: OneDrive` can still activate the provider, so an explicit runtime guard may be worth adding.

johannesjo:
Thanks for the continued fixes. I rechecked current head `c0ea958eaf`; I would still request changes on these items before merge.

1. **Initial OneDrive create still does not use the documented create-only behavior**  
   `packages/sync-providers/src/file-based/onedrive/onedrive.ts:189-198` now sends `If-None-Match: *` for `revToMatch === null`, but Microsoft documents the small upload endpoint as upload-or-replace, and the documented way to avoid replacing an existing item is `@microsoft.graph.conflictBehavior=fail` on the request URL. If Graph does not honor `If-None-Match` for this endpoint, two fresh clients can still silently overwrite `sync-data.json`. Please use the documented conflict behavior for the null-rev/non-force path and add a regression test for the request URL plus 409/412 mapping.  
   Ref: https://learn.microsoft.com/en-gb/onedrive/developer/rest-api/api/driveitem_put_content?view=odsp-graph-online#conflict-resolution-behavior

2. **Changing OneDrive auth identity can keep old OAuth tokens**  
   `src/app/imex/sync/dialog-sync-cfg/dialog-sync-cfg.component.ts:420-434` merges the OneDrive form config over existing private config before auth. If the user changes `useCustomApp`, `clientId`, or `tenantId`, existing `accessToken` / `refreshToken` are preserved; then `configuredAuthForSyncProviderIfNecessary()` skips OAuth because `isReady()` only checks for any resolved client ID plus tokens. That can save a new app/tenant config with tokens from the old identity. Please clear tokens and force reauth when auth-identity fields change, and only preserve tokens for non-auth settings like `syncFolderPath`.

3. **Some credential-clearing paths are still unguarded against stale requests**  
   The refresh failure paths now compare current config before clearing, but `packages/sync-providers/src/file-based/onedrive/onedrive.ts:632` and `:639` still call `clearAuthCredentials()` directly after a retry 401 or `403 InvalidAuthenticationToken`. A stale in-flight request can still wipe credentials that were replaced by a newer reauth. Please route all credential clearing through one helper that reloads current config and clears only if `refreshToken` / `clientId` / `tenantId` still match the request config.

4. **Folder existence probing swallows non-404 failures**  
   `packages/sync-providers/src/file-based/onedrive/onedrive.ts:322-329` catches every full-folder probe failure and falls through to folder creation. A 429, 5xx, or auth error should not be converted into extra create attempts or a later misleading error. Please fall through only for Graph 404 and rethrow everything else.

5. **`zh.json` and `zh-tw.json` still remove unrelated translations**  
   The PR changes non-English locale files despite the repo rule to edit only `en.json`, and the current files remove existing keys such as `F.SYNC.FORM.NEXTCLOUD.*`, `F.SYNC.S.ERROR_REMOTE_FILE_*`, `TASK_WIDGET`, and `DIALOG_LOGS`. A flattened key check shows both `zh` and `zh-tw` are missing 244 keys relative to `en`. Please revert the `zh.json` / `zh-tw.json` changes in this PR and keep only the `en.json` additions.

6. **OneDrive is hidden on web but still registered at runtime**  
   `sync-form.const.ts` hides OneDrive from the web provider select, but `sync-providers.factory.ts` still imports and registers it unconditionally. A persisted/imported `syncProvider: OneDrive` can still activate unsupported browser auth with no normal UI path to repair it. Please centralize an `isOneDriveSupported` check and use it in both provider registration and the form.

I would also add focused tests for the create-only upload URL, stale credential clearing after reauth, auth-identity changes forcing reauth, and non-404 folder probe failures.

johannesjo:
Thanks for the thorough work — the OAuth/PKCE core, token-refresh dedup, ETag handling, and credential storage (IndexedDB, excluded from exportable config) all look solid. A few items worth addressing before merge, grouped by what I'd consider blocking vs. follow-up.

## Should fix before merge

### 1. `listFiles` doesn't follow `@odata.nextLink`
`packages/sync-providers/src/file-based/onedrive/onedrive.ts:367-388` — Microsoft Graph `/children` caps each response at ~200 items. The op-log sync produces many small files per cycle; once the folder exceeds the page size, remote operations will silently vanish from the listing, causing replay drift. Either set `?$top=…` and follow `nextLink`, or paginate explicitly.

### 2. Logger misuse: `{...} as unknown as string`
`onedrive.ts:718-722, 738-742` — `SyncLogger.log(message: string, meta?)` is being called with an object cast to `string`. This will render as `[object Object]` in production log adapters and bypasses the typed contract Dropbox uses. The double cast also violates the project's no-`any`/`unknown`-cast rule.
```ts
this._deps.logger.normal('OneDrive.request', { method: options.method, isUpload: isUploadRequest });
```

### 3. `as any` casts in `dialog-sync-cfg`
`src/app/imex/sync/dialog-sync-cfg/dialog-sync-cfg.component.ts:1279, 1323` — the shape is fully known (`OneDrivePrivateCfg` / `Partial<OneDrivePrivateCfg>`). Per `CLAUDE.md`, use proper types.

### 4. `uploadFile` returns `{ rev: '' }` when Graph omits `eTag`
`onedrive.ts:347-348` — other providers throw `NoRevAPIError` in that case. An empty rev makes the next `If-Match: ""` request fail with 412 and surface as a false conflict. Either throw, or fetch metadata to obtain the eTag.

### 5. `updateSettingsFromForm` silently clobbers settings
`src/app/imex/sync/sync-config.service.ts:1722-1733` — the new explicit object literal converts `undefined` Formly fields (`isEnabled`, `isCompressionEnabled`, `isManualSyncOnly`) to `false`, overwriting prior true values on partial form updates. Suggest keeping the destructuring form and adding `oneDrive` to the omitted keys.

### 6. OneDrive provider chunk loaded on web unconditionally
`src/app/op-log/sync-providers/sync-providers.factory.ts:38-46` — `import('./file-based/onedrive/onedrive')` runs inside the top-level `Promise.all`, but `createOneDriveProvider` is only invoked under `if (IS_ONEDRIVE_SUPPORTED)`. Web users download the chunk for a feature they can't use. Compare to the gated `local-file-sync-electron` import — move the dynamic import inside the `if` block.

## Worth addressing, but doesn't have to block

### 7. OneDrive bypasses `executeNativeRequestWithRetry`
`IS_ONEDRIVE_SUPPORTED = IS_ELECTRON || IS_NATIVE_PLATFORM` advertises Android support, but `_request` uses `webFetch()` unconditionally. Dropbox routes through the native HTTP layer for Capacitor (transient network resilience). Either wire `nativeHttpExecutor` in, or gate OneDrive to Electron-only for now.

### 8. CSRF state validation is skippable in the URL-paste fallback
`dialog-get-and-enter-auth-code.component.ts:1209-1248` — state is only validated `if (stateFromUrl && providerName === 'onedrive')`. The regex fallback path (when `new URL()` fails) also bypasses validation. PKCE protects the code-injection case in practice, but the dialog otherwise advertises CSRF defense and the "missing state ⇒ accepted" branch undermines it. Treat URL-paste without state as failure for OneDrive (raw `code`-only paste is fine).

### 9. `verifyCodeChallenge` returns more than the contract declares
`onedrive.ts:273-275` returns the full `OneDrivePrivateCfg` (clientId, tenantId, useCustomApp, syncFolderPath) but the `SyncProviderAuthHelper` interface declares `{ accessToken; refreshToken; expiresAt }`. The generic `<T>` cast hides the divergence; the caller spreads the extras back into private cfg so it works, but the type contract is misleading.

### 10. `id_token` missing from `_redactSensitiveFields`
`onedrive.ts:843-857` — JWTs in `id_token` carry PII (oid, email, name, tid). Add to the sensitive-keys list as a forward-safe measure.

## Architectural follow-ups (separate PR is fine)

- **Lift OAuth/token-refresh helpers shared with Dropbox.** This is the second OAuth+ETag provider, and the two implementations have already diverged — OneDrive has in-flight refresh dedup that Dropbox lacks; Dropbox has native-HTTP retry that OneDrive lacks. Extracting `withTokenRefresh()` + `refreshTokenWithInflightDedup()` + a shared HTTP-status → typed-error mapper would prevent bug drift before the third provider arrives.
- **Convert `OAUTH_STATES_MAP` to an `@Injectable({ providedIn: 'root' })` service.** Module-level mutable globals can't be reset between tests and bypass Angular DI. The package-side `addOAuthState` callback shape is fine — just back it with a service.
- **Move identity-drift detection into `OneDrive.setPrivateCfg()`.** Currently both the provider (`_clearIfConfigMatches`) and the dialog (`_persistOneDriveFormCfgBeforeAuth`) encode "identity changed → wipe tokens". Folding this into the provider drops the OneDrive-specific branch from the shared dialog.
- **Reconsider `useCustomApp` until an official client ID ships.** `HAS_OFFICIAL_ONEDRIVE_CLIENT_ID` is currently false, so only the custom-app branch is ever exercised. The toggle, the `_resolveClientId` tri-state, the form `hideExpression` ladder, and the dialog identity-change block all exist for a code path that isn't reachable yet. Probably ~80 LoC of YAGNI that can be reintroduced cleanly the day an official app is registered.
- **`dialog-get-and-enter-auth-code` hardcodes the provider name `'onedrive'`** at line 1220. The next OAuth provider will copy-paste the same branch. Returning a `verifyRedirectUrl(url)` from `getAuthHelper()` keeps state validation provider-local.

## Nits

- Spec assertions changed from `provider: 'dropbox'` to `provider: 'unknown'` for test URLs (`oauth-callback-handler.service.spec.ts:1369, 1378, 1387, 1396`). This is cosmetic — Dropbox's `_getRedirectUri()` returns `null` so it never hits this auto-channel in production — but a one-line comment in the spec would prevent future confusion.
- `_normalizeRelativePath` / `_joinPathSegments` / `_encodePath` could collapse into one helper.
- `_redactTokenBody` and `_redactResponseBody` are one-line wrappers around `_redactSensitiveFields` — just inline.
- `OneDriveDeps.platformInfo` is declared but never read; parallel `isElectron: boolean` dep duplicates what `platformInfo` already encodes.

---

Overall this is well-structured for a 1830-line feature, and the privacy hygiene (URL redaction in `protocol-handler.ts`, `event.url` stripped from logs, response-body sensitive-field redaction) is genuinely careful. Happy to look again once #1–6 are addressed.

johannesjo:
Thank you very much for this! <3 That's a very cool addition!

PR #7524, opened 2026-05-08
feat(task-repeat): add repeat after completion option

### LINKED ISSUE #7334: 💡 add Repeat after completion button in recurring tasks
### Problem Statement

sometimes it get over load with same tasks, feels clutter 

### Possible Solution

 add Repeat after completion button in recurring tasks
and it  may just look like this

Image

_No response_

### Additional context

_No response_

### documentation gap

Hetsavani:
## Problem

Recurring tasks can accumulate and create clutter when previous instances are not completed.
Fixes: #7334 

## Solution

Added a new Repeat after completion option for recurring tasks.

When enabled, the next recurring task is only created after the current one is completed, helping prevent multiple pending instances from piling up.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo (commented):
Thanks for picking this up @Hetsavani\! The implementation is tight and well-scoped, but there are a few things worth addressing before merge. I verified each of these by reading the current files on the branch.

## Blocker — unrelated change

**`task-repeat-cfg-form.const.ts:259-268`** — the PR removes the `hideExpression` from the **pre-existing** `repeatFromCompletionDate` field, which on `master` is hidden unless `quickSetting === 'CUSTOM'`. After this PR it's shown for every preset (DAILY/WEEKLY/MONTHLY/YEARLY). This is a separate feature from `waitForCompletion` (one shifts the *due-date timing* relative to completion; the other gates *creation entirely* on prior completion) and isn't mentioned in the description.

The labels happen to read sensibly for non-CUSTOM presets, so this could be a defensible UX change — but it should be split into its own PR with explicit rationale, or reverted here.

## Important — behavior gaps

1. **Archived previous instances silently unblock the gate.** `existingTaskInstances` comes from `selectTaskWithSubTasksByRepeatConfigId`, which only reads the live task store. Once the previous instance is archived (e.g. via the daily archive sweep) without ever being completed, `existingTaskInstances` is empty and `waitForCompletion` no longer blocks. Compare with `effects.ts:577` which uses `getArchiveTasksForRepeatCfgId` for similar checks. Worth deciding whether an archived-but-not-done instance should still block, and including archive tasks (or filtering by `isDone` against archive) accordingly.

2. **No trigger when the user completes the gating task.** Creation only runs from `task-due.effects.ts` on `todayDateStr$.distinctUntilChanged()` — i.e. day boundary or app reload. So if the user completes the blocking task at 14:00, the next instance won't appear until tomorrow. There's already an analogous effect at `task-repeat-cfg.effects.ts:543-565` for `repeatFromCompletionDate` that fires on completion; consider a parallel one (or extending that effect) so users see the next instance materialize immediately after completion.

3. **`lastTaskCreationDay` is not advanced when blocked.** The `skipOverdue` branch a few lines below explicitly advances `lastTaskCreationDay` to avoid re-evaluating the same overdue date on every app open and stalling progress. Returning `[]` here re-introduces that same risk: once the gate is finally unblocked on a weekly/monthly cfg, an overdue instance can fire on a non-scheduled day. Probably fine for DAILY but worth thinking through for the other cycles.

## Minor

- **No unit tests.** `task-repeat-cfg.service.spec.ts` already has dense coverage for sibling branches (`skipOverdue`, `deletedInstanceDates`, duplicate prevention). Worth adding cases for: blocked when prior is uncompleted, creates when prior is `isDone`, creates when there are no prior instances, and the interaction with `repeatFromCompletionDate` (both flags can be set simultaneously and behavior should be specified).
- **i18n key ordering.** `WAIT_FOR_COMPLETION` is inserted between `SKIP_OVERDUE_DESCRIPTION` and `START_DATE` in both `t.const.ts` and `en.json`, breaking the alphabetical convention used in the surrounding block.
- **Field placement.** `waitForCompletion` is placed before `shouldInheritSubtasks`. Since it's a creation-gating toggle, it groups more naturally with `skipOverdue` (also creation-gating).

Subtask interaction is fine (subtasks created here don't carry `repeatCfgId`, so an open subtask under a done parent won't block — that's the right semantic). And no migration is needed since `undefined` reads as falsy.

CI's "Tests" failure is an unrelated flaky e2e (`autoStartFocusOnPlay` / `detail-panel-focus`), not introduced by this PR.

johannesjo:
Thanks for the follow-up. The UI/test cleanup helps, but I think the core `waitForCompletion` behavior still needs another pass before merge.

**Blocking:** `task-repeat-cfg.service.ts` advances `lastTaskCreationDay` when `waitForCompletion` blocks creation, even though no task was created. Since that field is also the repeat cursor, the skipped occurrence is now treated as processed. This can make the completion effect fail its “latest instance” check, or leave the config anchored to the blocked date rather than the real completion date.

**Blocking:** `task-repeat-cfg.effects.ts` handles completion by only updating the repeat config. It still does not create/materialize the next occurrence when the user completes the gating task. Normal recurring task creation runs from the day-change/app-start path, so completing the current task during the day can leave users waiting until the next scheduled pass instead of seeing the next task appear after completion.

The archived-instance gap from the earlier review also still looks unresolved: the gate checks live instances only, so an archived but unfinished repeat instance can stop blocking new creation.

I think the fix needs to treat “blocked because waiting for completion” separately from “created/skipped this occurrence”, and completion should either create the next eligible instance immediately or trigger the same creation path with the updated config.

johannesjo:
Thanks, the latest commit does address the earlier `waitForCompletion` feedback: blocked creation no longer advances the cursor, completion now goes through the creation path, and archived instances are included.

I found one remaining blocker in `task-repeat-cfg.effects.ts` around `updateStartDateOnComplete$`: for `repeatFromCompletionDate`, the effect now updates only `startDate`.

That regresses the existing "after completion" scheduling behavior because `getEffectiveRepeatStartDate()` ignores `startDate` whenever `repeatFromCompletionDate && lastTaskCreationDay` are set. In normal existing configs, `lastTaskCreationDay` is present, so updating only `startDate` does not actually re-anchor the recurrence to the completion day. A task configured as "every N days after I finish" can keep calculating from the previous creation day instead of today.

I think the `repeatFromCompletionDate` branch should keep updating `lastTaskCreationDay` as before, e.g. `changes: { startDate: today, lastTaskCreationDay: today }`. The new `_getActionsForTaskRepeatCfg` path looks appropriate for `waitForCompletion`-only configs, but the completion-based schedule still needs its cursor updated on completion.

johannesjo:
I did one more pass on the latest head (`0161f5c33bfb6709aabe34621ca3d5fa9b4462c6`). The earlier cursor/materialization fixes look addressed, but I think there are two remaining items worth handling before merge:

1. **Completion can miss rescheduled `waitForCompletion` instances.** `updateStartDateOnComplete$` gates the new creation path through `_isLatestInstance()`, and that helper uses `task.dueDay || dueWithTime || created` to identify the repeat occurrence. `dueDay` is mutable when a task is planned/rescheduled, so the newest generated instance can fail the `taskDay === lastCreationDay` check after being moved to today. For the latest-instance guard, please use the stable occurrence identity, probably the task `created` date / deterministic repeat date, rather than `dueDay`.

2. **`WAITFORCOMPLETION_FIX.md` should not ship as root docs.** It reads like an implementation/reviewer scratch note. Since this is a user-facing recurring-task option, the repo guidance points to updating `docs/wiki/3.02-Settings-and-Preferences.md` instead. I would remove the root note and add a short user-facing description to the wiki page.

Hetsavani:
Thanks. WAITFORCOMPLETION_FIX.md was added by mistake during implementation notes — I have removed it and made necessary changes

Hetsavani:
Addressed the remaining review feedback:

* fixed repeatFromCompletionDate re-anchoring by restoring `lastTaskCreationDay` updates on completion
* updated latest-instance detection to use a stable occurrence identity instead of mutable `dueDay`
* removed the accidental `WAITFORCOMPLETION_FIX.md` note file

CI is now green as well.

johannesjo:
Thank you very much @Hetsavani ! I made some additional adjustments. I hope thats alright with you? Could you maybe have a look if the changes match your vision?

Hetsavani:
Seems Perfect!

johannesjo:
Thanks, the latest changes address the linked feedback around the scratch doc removal, stable `created`-date latest-instance detection, archived instances, and CI is green.

I found one remaining blocker:

**Blocking:** `task-repeat-cfg.effects.ts:578` calls `_getActionsForTaskRepeatCfg(cfg, Date.now())` after every completion for `waitForCompletion` configs. But `task-repeat-cfg.service.ts:197-200` treats “no due occurrence exists yet” as a `devError`.

A normal completion can therefore raise/log a dev error. Example: a daily task created today has `lastTaskCreationDay === today`; completing it today should simply wait until tomorrow, but `getNewestPossibleDueDate()` returns `null`, so the completion effect hits `devError('No target creation date found for repeatable task')`.

I think the completion path needs a non-error/probe mode, or `_getActionsForTaskRepeatCfg` should optionally return `[]` without `devError` when “no occurrence due” is expected. A regression test for “waitForCompletion task completed before the next scheduled occurrence is due” would cover this.

Hetsavani:
Any feedback on this @johannesjo ?

johannesjo:
No, sorry! I got confused! Thanks for the adjustments <3

PR #7527, opened 2026-05-08
fix(electron): retry Wayland idle helper startup

davidvornholt:
## Summary

Hi! This PR focuses only on making the new Wayland idle helper more reliable and safer to ship:

- retries/promotion for the Wayland `ext-idle-notify` helper when the app initially falls back to another idle-detection method
- keeps the helper build mandatory in CI/release-style builds so Linux packages do not accidentally ship without the Rust helper
- keeps local development builds ergonomic by warning and skipping the helper if Cargo is unavailable, unless the build explicitly requires it

## Why this is necessary

The Rust helper is part of the shipped Linux Electron artifact. Downstream packaging targets such as Flatpak need the CI/CD pipeline to compile it; silently omitting it would produce a package without `ext-idle-notify` support.

At the same time, `npm run electron:build` is also useful for local development. The build script now distinguishes between local and required builds:

- `CI=true` or `SP_REQUIRE_WAYLAND_IDLE_HELPER_BUILD=1` + missing Cargo → fail the build
- `SP_SKIP_WAYLAND_IDLE_HELPER_BUILD=1` → intentionally skip the helper
- local build without Cargo and without the required flag → warn and continue

This makes release/package builds strict while avoiding unnecessary friction for contributors who only want to compile the Electron main process locally.

## Testing

- `npm run checkFile electron/idle-time-handler.ts`
- `npm run prettier:file -- tools/build-wayland-idle-helper.js`
- verified `tools/build-wayland-idle-helper.js` behavior for:
  - local missing-Cargo build
  - CI missing-Cargo build
  - explicit skip via `SP_SKIP_WAYLAND_IDLE_HELPER_BUILD=1`
- pre-push hook ran `npm run lint` and full `npm run test` successfully


johannesjo (commented):
Nice improvement to the Wayland fallback story — the build-script gating in particular reads cleanly. A few things worth considering before merge:

**1. `getIdleTime()` now blocks on promotion attempts** (`electron/idle-time-handler.ts:102-106`)

Every call awaits `_maybePromoteToWaylandIdleNotify`, which awaits `_ensureWaylandIdleHelperStarted` when promoting. That can stall the caller for up to `WAYLAND_IDLE_HELPER_READY_TIMEOUT_MS` (3 s). The poller in `start-app.ts` skips overlapping ticks via `isCheckingIdle`, so a promotion attempt suppresses idle detection for ~3 s.

Could the promotion fire-and-forget instead?

```ts
private _maybePromoteToWaylandIdleNotify(currentMethod: IdleDetectionMethod): IdleDetectionMethod {
  // ...throttle/early-return checks...
  void this._tryPromoteToWaylandIdleNotify();
  return currentMethod;
}
```

The next poll then picks up the upgraded `_workingMethod` automatically — no stall, and the return type stays synchronous.

**2. Possible flap between promotion and `_handleWaylandIdleHelperFailure`**

If the helper is promoted and then fails 3 times, `_handleWaylandIdleHelperFailure` resets `_workingMethod = 'none'` and clears `_methodDetectionPromise`. The next `_determineWorkingMethod` runs `waylandIdleNotify` as a candidate again, which can re-succeed → fail → re-detect, etc. Worth either adding a short "demoted recently" cooldown or documenting why oscillation is acceptable.

**3. Throttle stamp set before attempt resolves** (`electron/idle-time-handler.ts:208`)

Setting `_lastWaylandIdleNotifyRetryMs = now` before the 3 s helper-ready attempt is intentional (prevents in-flight duplication), but means the *effective* retry window is closer to ~57 s than 60 s. A one-line comment would prevent a future "cleanup" PR from moving it.

**Minor**

- No unit tests for the promotion path. The throttle/early-return logic is pure and easy to mock — would be nice to lock in.
- Per `CLAUDE.md`, project standard is `npm run checkFile <filepath>` (prettier + lint) rather than `npm run prettier:file`. Worth running on both files.
- `isTruthyEnv` is case-sensitive on `'true'`. Edge case (most CI sets lowercase), but cheap to make case-insensitive.

Build-script changes look good as-is — the explicit-skip wins over implicit-require ordering is the right precedence, and the error message points the user at the escape hatch.

davidvornholt:
Thanks for the careful review! I pushed an update that addresses the main points:

- Promotion is now fire-and-forget, so `getIdleTime()` no longer waits on the 3s helper-ready timeout. The current poll keeps using the existing method and the next poll picks up `waylandIdleNotify` once promotion succeeds.
- Added a short demotion cooldown after repeated helper failures to avoid immediate promote/fail/re-detect oscillation.
- Added a comment explaining why the retry timestamp is set before the async readiness attempt.
- Made `isTruthyEnv` accept `TRUE`/mixed-case values too.

I also ran:

```bash
npm run checkFile electron/idle-time-handler.ts
npm run checkFile tools/build-wayland-idle-helper.js
npm run electron:build
```

The push hook also ran lint and the unit test suite successfully.

davidvornholt:
I fixed the failing reminder E2E selector and pushed it.

The failure was caused by `getByText('never')` matching both the reminder option and unrelated tooltip text. The test now uses the displayed option casing and exact text matching when selecting the option.

Validated locally with:

```bash
npm run checkFile e2e/tests/reminders/reminders-default-task-remind-option.spec.ts
npm run e2e:file e2e/tests/reminders/reminders-default-task-remind-option.spec.ts -- --retries=0
```

The targeted reminder E2E file passes: 3/3. Push hook also ran lint and unit tests successfully.

johannesjo:
Thanks for the follow-up. The reminder E2E conflict resolution looks good to me now.

I still see two issues worth addressing before merge:

1. `electron/idle-time-handler.ts` - the demotion cooldown does not apply to full method detection. `_handleWaylandIdleHelperFailure()` records `_lastWaylandIdleNotifyDemotionMs` and clears `_methodDetectionPromise`, but the next `getIdleTime()` can immediately rerun `_determineWorkingMethod()`, where `waylandIdleNotify` is still tested unconditionally. So the promote/fail/re-detect loop can still happen. The cooldown should also gate the `waylandIdleNotify` candidate in `_buildWaylandCandidates()` or `_determineWorkingMethod()`.

2. `tools/build-wayland-idle-helper.js` / release scripts - helper builds are only mandatory when `CI` or `SP_REQUIRE_WAYLAND_IDLE_HELPER_BUILD` is set. Local Linux release-style builds such as `npm run dist` can still warn and continue without Cargo, which can produce packages without the helper. If those artifacts are release/package artifacts, the relevant scripts should set `SP_REQUIRE_WAYLAND_IDLE_HELPER_BUILD=1`, or packaging should fail when the helper is missing unless `SP_SKIP_WAYLAND_IDLE_HELPER_BUILD=1` is explicitly set.

davidvornholt:
Thanks again, both points were spot on. I pushed another update addressing them:

- The Wayland idle-notify demotion cooldown now also gates full method detection. During the cooldown, `_buildWaylandCandidates()` skips the `waylandIdleNotify` candidate, so repeated helper failures cannot immediately bounce through `none -> redetect -> waylandIdleNotify`.
- Linux packaging now fails in `afterPack` if the built `wayland-idle-helper` binary is missing, unless `SP_SKIP_WAYLAND_IDLE_HELPER_BUILD=1` is explicitly set. That keeps local release-style packages from silently shipping without ext-idle-notify support.

Validated with:

```bash
npm run checkFile electron/idle-time-handler.ts
npm run checkFile tools/afterPack.js
npm run electron:build
```

The push hook also ran lint and unit tests successfully.

davidvornholt:
I merged the latest master into the branch and resolved the remaining conflict in the reminder default option E2E spec. The branch is pushed and GitHub now reports the PR as mergeable.

johannesjo:
I pushed a small maintainer follow-up in `7d6d019e2` to avoid stale Wayland helper binaries being packaged when the helper build is skipped.

What changed:

- `tools/build-wayland-idle-helper.js` now removes any existing `electron/bin/wayland-idle-helper` output when the build is explicitly skipped, or when a local non-required build skips because Cargo is unavailable.
- `tools/afterPack.js` now honors `SP_SKIP_WAYLAND_IDLE_HELPER_BUILD=1` before checking/copying the helper source, so explicit skip cannot accidentally copy an older helper.

Validated with:

```bash
npm run checkFile tools/build-wayland-idle-helper.js
npm run checkFile tools/afterPack.js
```

johannesjo:
Thank you very much! <3

Issue #6654, opened 2026-02-27
💡 Allow filtering issue provider events


calvin-fisher:
### Problem Statement

I have my work Google Calendar synced as an issue provider. I don't automatically import all events because it includes events like out of office blocks. I typically add the meetings in the morning, but sometimes I forget and it's busywork that could be automated.

### Possible Solution

Have a simple `field`-`comparator`-`value` filter to avoid importing specific events when "Auto import events as tasks for the current day" is turned on

### Describe alternatives you've considered

- Continue to add/filter manually
- Set up a CalDAV feed with a filter and subscribe to the filtered feed

### Additional context

Another one I'd be happy to do if it makes sense when I have a minute.

agncr:
I have a similar use case — I want to hide recurring events like lunch breaks and dailies from showing up in my schedule/auto-import.

I had Claude Code build a regex-based filter on a local branch — I'm not an Angular/TypeScript developer myself. It adds two optional fields per calendar provider:

<img width="603" height="539" alt="Image" src="https://github.com/user-attachments/assets/0294bddb-eb59-474e-9d96-f5004f8402cb" />

Include regex (whitelist): Only events whose title matches are shown. Leave empty to include all.
Exclude regex (blacklist): Events whose title matches are hidden. Applied after the include filter.

Both are case-insensitive and support pipe-separated alternatives (e.g. Lunch|Daily|Blocker). The filter is applied at all three pipeline points (schedule view, auto-import as tasks, notification banners).

It's been working well for me locally. I could open a PR if you'd like to take a look.

agncr:
@johannesjo this can be closed

johannesjo:
Thank you @agncr !

PR #7528, opened 2026-05-08
feat(calendar): add per-provider include/exclude regex filters for iCal events

agncr:
Allow hiding recurring events like lunch breaks and dailies from showing up in schedule/auto-import. (s. #6654)

Adds optional include and exclude regex filters per calendar provider. Both fields are configurable in the calendar settings form:

Include regex (allowlist): Only events whose title matches are shown. Leave empty to include all.
Exclude regex (denylist): Events whose title matches are hidden. Applied after the include filter.

Both are case-insensitive and support pipe-separated alternatives (e.g. Lunch|Daily|Blocker). The filter is applied at all three pipeline points (schedule view, auto-import as tasks, notification banners).


johannesjo:
Thanks for the PR. I found a couple of issues that should be addressed before merging:

1. The new `filterIncludeRegex` / `filterExcludeRegex` values can leak through privacy exports. These fields may contain user-specific calendar title fragments like `Lunch|Daily|ClientName`, but `src/app/imex/file-imex/privacy-export.ts` does not add them to `KEY_TO_REPLACE`. Please mask both keys there and add/update a privacy export test.

2. `calendarEvents$` still emits `_getCalProviderFromCache()` directly before provider config is available, so cached calendar events bypass the new regex filters on initial schedule render. Please make the cached emission go through the same provider-aware filtering path, otherwise excluded cached events can briefly reappear.

3. Banner filtering only affects newly added banner events. Existing entries in `_currentlyShownBanners$` are not reconciled when a provider’s regex filters change, so an event that becomes excluded can remain visible until manually dismissed. Please clear or re-filter queued banners when calendar provider config changes.

agncr:
Thank you for the review. 

privacy exports: 
`filterIncludeRegex` and `filterExcludeRegex` are now included in `KEY_TO_REPLACE` in `privacy-export.ts`. Tests added

cached calendar events bypass the new regex filters on initial schedule render: 
initial `defer()` emission now reads `selectCalendarProviders` via `first()`, builds a provider map, and runs `passesCalendarEventRegexFilter` over each cached entry before emitting — so regex filters are applied consistently from the very first render

Existing entries in _currentlyShownBanners$ are not reconciled: 
Added a `reconcileBannersOnProviderChange` effect that subscribes to `selectCalendarProviders` (skipping the initial emission) and re-evaluates `_currentlyShownBanners$` against the updated regex config on every provider change. If any queued banner no longer passes the filter, it's removed immediately without waiting for a dismiss

johannesjo:
Thanks for the updates. The previous privacy export and banner reconciliation items look addressed, but I found one remaining issue in the cache/filter path:

1. `src/app/features/calendar-integration/calendar-integration.service.ts:213-231` now saves the already regex-filtered schedule result back into `LS.CAL_EVENTS_CACHE`. That means excluding something like `Lunch` overwrites the cache without those events. If the user later loosens/removes the regex while the calendar is offline or the next fetch fails, the fallback cache cannot show those events again until a successful refetch. The cache should preserve the fetched events before applying the current regex, or otherwise store raw per-provider data and apply the current filters only when reading/emitting schedule data.

2. Related: the initial cached emission at `src/app/features/calendar-integration/calendar-integration.service.ts:87-98` derives one provider config from `entry.items[0]?.calProviderId` and applies it to every item in that cache entry. `ScheduleCalendarMapEntry` only has `items`, and cached entries can contain mixed provider items; `_groupCachedEventsByProvider()` already handles this by grouping per item. This can apply provider A regex to provider B cached events. Please resolve the provider config per `calEv.calProviderId` inside the filter callback, and add a regression test with a mixed cached entry.

johannesjo:
Thanks for the updates. I found one remaining cache-path issue:

`calendarEvents$` now stores raw fetched events in `LS.CAL_EVENTS_CACHE`, which is good, but the initial cached emission only reapplies the new regex filters. It does not reapply the existing schedule visibility filters for events that were already imported as tasks, skipped for today, or hidden.

Specifically, the cached path around `src/app/features/calendar-integration/calendar-integration.service.ts:86-96` applies `passesCalendarEventRegexFilter(...)`, while the live path later also filters against `allCalendarTaskEventIds`, `skippedEventIds`, and `hiddenEventIds` at `:226-228`.

That means after restart/initial schedule render, cached events that were already converted to tasks, skipped, or hidden can briefly reappear until the next fetch/fallback emission. Please make the cached emission go through the same view-level filtering as the live schedule emission.

johannesjo:
Thanks for the follow-ups. I pushed one final maintainer commit (`e4885f85`) to avoid another review cycle: it re-applies provider regex changes immediately from the raw cached calendar events before the network refresh completes, reuses the same view-level filtering for cache/live paths, and adds the missing iCal wiki note.

Verified locally:

- `npm run checkFile src/app/features/calendar-integration/calendar-integration.service.ts`
- `npm run checkFile src/app/features/calendar-integration/calendar-integration.service.spec.ts`
- `npm run test:file src/app/features/calendar-integration/calendar-integration.service.spec.ts`

PR #7538, opened 2026-05-10
feat(schedule): Add new Today button to Schedule view

### LINKED ISSUE #7537: 💡 Replace Today button circle in Schedule view
### Problem Statement

In the top left of the Schedule view is a button which automatically returns the calendar to where the current day is visible. However this button is just a blank circle with no indication of its function beyond a tooltip that is only visible when a user hovers over it.

Image

### Possible Solution

Replace the empty circle with the word `Today`, very similar to Google Calendar.

Image

_No response_

### Additional context

_No response_

cpalvis:
## Problem

The previous Today button only had an unchecked box icon which did not explain its function.

sp-592026-1519

## Solution

Replace the icon with text reading "Today" and the universal SP Today icon `wb_sunny`. Change tooltip from `Today` to the current date, similar to the Google Calendar implementation.

Fixes #7537

### Normal (Light mode)
sp-592026-1535

### Hovered (Light mode)
sp-592026-1730

### Normal (Dark mode)
sp-592026-1532

### Hovered (Light mode)
sp-592026-1533

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [x] Other (please describe)

Purely a UI change. No functionality altered.

## Checklist

- [ ] I have run `npm run checkFile` on changed `.ts`/`.scss` files [N/A]
- [ ] I have added tests for my changes (if applicable) [N/A]

### DIFF
--- a/src/app/core/date-time-format/date-time-format.service.ts
+++ b/src/app/core/date-time-format/date-time-format.service.ts
@@ -91,12 +91,28 @@ export class DateTimeFormatService {
    * // For en-GB locale
    * formatDate(new Date(2000, 11, 31), 'en-GB'); // 31/12/2000
    */
-  formatDate(date: Date, locale: DateTimeLocale = this.currentLocale()): string {
-    const formatter = new Intl.DateTimeFormat(locale, {
-      year: 'numeric',
-      month: '2-digit',
-      day: '2-digit',
-    });
+  formatDate(
+    date: Date,
+    locale: DateTimeLocale = this.currentLocale(),
+    formatOptions?: 'long',
+  ): string {
+    let format: Intl.DateTimeFormatOptions = {};
+
+    if (formatOptions === 'long') {
+      format = {
+        weekday: 'long',
+        year: 'numeric',
+        month: 'long',
+        day: '2-digit',
+      };
+    } else {
+      format = {
+        year: 'numeric',
+        month: '2-digit',
+        day: '2-digit',
+      };
+    }
+    const formatter = new Intl.DateTimeFormat(locale, format);
 
     return formatter.format(date);
   }
--- a/src/app/features/schedule/schedule/schedule.component.html
+++ b/src/app/features/schedule/schedule/schedule.component.html
@@ -80,14 +80,13 @@
     </button>
 
     <button
-      mat-icon-button
       class="today-btn"
       (click)="goToToday()"
       [disabled]="isViewingToday()"
       [attr.aria-label]="T.G.TODAY_TAG_TITLE | translate"
-      [matTooltip]="T.G.TODAY_TAG_TITLE | translate"
     >
-      <mat-icon>radio_button_unchecked</mat-icon>
+      <mat-icon>wb_sunny</mat-icon>
+      {{ 'Today' | translate }}
     </button>
 
     <button
--- a/src/app/features/schedule/schedule/schedule.component.html
+++ b/src/app/features/schedule/schedule/schedule.component.html
@@ -72,7 +72,9 @@
       [matTooltipDisabled]="isViewingToday()"
     >
       <mat-icon>wb_sunny</mat-icon>
-      {{ T.G.TODAY_TAG_TITLE | translate }}
+      @if (!this.layoutService.isXxxs()) {
+        {{ T.G.TODAY_TAG_TITLE | translate }}
+      }
     </button>
 
     <button
--- a/src/app/features/schedule/schedule/schedule.component.html
+++ b/src/app/features/schedule/schedule/schedule.component.html
@@ -64,6 +64,18 @@
 
   <div class="date-nav-group">
     <button
+      class="today-btn"
+      (click)="goToToday()"
+      [disabled]="isViewingToday()"
+      [attr.aria-label]="T.G.TODAY_TAG_TITLE | translate"
+      [matTooltip]="displayTodayDate()"
+      [matTooltipDisabled]="isViewingToday()"
+    >
+      <mat-icon>wb_sunny</mat-icon>
+      {{ T.G.TODAY_TAG_TITLE | translate }}
+    </button>
+
+    <button
       mat-icon-button
       (click)="goToPreviousPeriod()"
       [disabled]="isViewingToday()"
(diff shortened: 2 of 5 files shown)


johannesjo:
The button is a bit inconsistent in terms of design. Also we need to make sure that this translates well to mobile or if there is not enough space for other reasons (when the extra button for toggling calendar visibility is there.

johannesjo:
Thanks for the update. I found two things to address before merge:

1. The visible label uses `{{ 'Today' | translate }}`, but `Today` is not a translation key. This will stay English in other locales. Please use the existing key, e.g. `{{ T.G.TODAY_TAG_TITLE | translate }}`, matching the aria label.

2. The tooltip now shows `_todayDateStr()` directly, which is the internal DB date format (`YYYY-MM-DD`). Since this is user-facing text, it should be formatted through the app’s locale-aware date formatting.

I checked the branch locally: `npm run checkFile src/app/features/schedule/schedule/schedule.component.scss` and `npm run build` both pass.

cpalvis:
### Translation Key
The hardcoded `Today` string was replaced with the `T.G.TODAY_TITLE_TAG` key
### Locale-Formatted Date
The tooltip value now runs through `safeFormatDate()` while passing the `currentLocale()` value as an argument, as well as the format 'EEEE, MMM dd, yyyy', which is the display format used by Google Calendar. However, this also locks in the format in terms of day to month to year order, so it might not be the best solution.
### CI Test Failure
The CI test is failing because the `src/app/features/schedule/schedule/schedule.component.spec.ts` unit tests think `_globalTrackingIntervalService.todayDateStr()` is not a callable function. However, `_globalTrackingIntervalService.todayDateStr()` is a Signal created by `toSignal()`, so I think it should be a callable function. I must be missing something.
### Mobile View
_Monthly Schedule without calendar visibility button_
WhatsApp Image 2026-05-10 at 11 59 07 PM (1)
_Monthly Schedule with calendar visibility button_
WhatsApp Image 2026-05-10 at 11 59 07 PM (2)
_Weekly Schedule without calendar visibility button_
WhatsApp Image 2026-05-10 at 11 59 07 PM

You were right. When the extra toggle is visible, the `Today` button overlaps with the Week title and messes up the UI.

**Solution**
Only render the Today icon and not the text if the width is too small. I'll try and see if I can implement this.

### Design
> The button is a bit inconsistent in terms of design.

How can I improve the design to make it more consistent?

> Today < > is better than < Today > (copy Google Calendar)

Also, what are your thoughts on this?

Thanks!

johannesjo:
Thanks for the detailed update!

The translation key in the actual diff looks good to me: `T.G.TODAY_TAG_TITLE` is the right existing key. Minor note: your comment says `T.G.TODAY_TITLE_TAG`, but the code uses the correct one.

For the CI failure: the issue is the test mock, not Angular Signals. `GlobalTrackingIntervalService` has `todayDateStr` as a signal in production, but `schedule.component.spec.ts` currently only mocks `todayDateStr$`. Adding `todayDateStr: signal('2026-01-20')` to that mock, similar to `planner-calendar-nav.component.spec.ts`, should fix it.

For the tooltip date, I agree with your concern. `safeFormatDate()` prevents missing-locale crashes, but `'EEEE, MMM dd, yyyy'` still hardcodes US-style order. A better solution would be formatting `parseDbDateStr(todayDateStr)` via `Intl.DateTimeFormat(locale, { weekday: 'long', year: 'numeric', month: 'long', day: 'numeric' })`, or putting that behind `DateTimeFormatService`.

For mobile/compact layout: yes, please make sure this works with the calendar visibility button enabled. Also note that there are/will be two additional buttons possible in this header, so available space is even more constrained than in the current screenshots. The layout should remain stable with all optional buttons visible. Hiding the text label and keeping only the icon at compact widths sounds good, as long as the aria label and tooltip still expose “Today”.

Design-wise I’d make the button closer to the existing toolbar controls: fixed height, inline-flex alignment, icon/text gap, horizontal padding in px/rem or existing spacing vars, and no percentage padding/margins. A neutral hover state would also fit better than filling the button with the primary color.

I’m fine with the `Today < >` ordering; it matches calendar apps and puts the reset action first. Main blockers from my side are CI, compact layout, and the locale-aware tooltip format.

johannesjo:
> How about replacing the circle with a better icon (like the sunny icon)?

Yes, might be simpler. We can use mat-tooltip to explain what is going to happen.

cpalvis:
>The translation key in the actual diff looks good to me: T.G.TODAY_TAG_TITLE is the right existing key. Minor note: your comment says T.G.TODAY_TITLE_TAG, but the code uses the correct one.

Sorry, that's a typo on my part.



> > How about replacing the circle with a better icon (like the sunny icon)?
> 
> Yes, might be simpler. We can use mat-tooltip to explain what is going to happen.

This makes sense in terms of simplicity and space. The only reason I didn't do it that way in the first place is that as a user, I don't like having to hover over a button to find out what it does. But it's not really a big deal.

cpalvis:
## Thought

I'm working on fixes for your feedback, but this just occurred to me. If there are going to be two _more_ buttons in the header, then even without the new `Today` button, space is going to be really tight. What do you think about moving the <> arrows our of the header and put them as semi-transparent overlays on the main screen? There's no reason users want to click on specific days because they don't lead to anything, and the icons could be semi-transparent until interacted with. That would free up some valuable real estate in the header. I know that this is _way_ beyond the scope of this PR, but I thought I'd mention it in context, so if you don't want it, it can get shot down quickly.

johannesjo:
Thanks — a quick "jump to today" on the Schedule is a reasonable addition, and the logic is correct. A few things before merge:

- The SCSS still uses percentage padding/margins (`padding: 5%; margin: 5%`) and a primary-color hover fill (`background: var(--c-primary); color: white`) — both were in my earlier feedback. Please switch to fixed px/rem or the spacing tokens (`--s` etc.) and a neutral hover that matches the sibling toolbar buttons, and drop the hardcoded `color: white` / `font-size: 14px` per the styling guide.
- The simplest version of this is really just swapping the existing icon to `wb_sunny` with a tooltip, which addresses #7537 without the new `formatDate` `'long'` option flag or the new bordered-button SCSS — worth weighing against the text-label button.
- The branch is well behind master and needs a rebase (the red CI is an unrelated, since-fixed Karma teardown flake, not your change).

cpalvis:
Hi, @johannesjo! So sorry I haven't looked at this in quite a while.

Since you want to go the `wb_sunny` icon route, the best option would probably be to close this PR in favor of #8390.

johannesjo:
Thank you!

PR #7557, opened 2026-05-11
Prevents task creation during IME conversion.

watanabeycr7:
## Problem

The Input Method Editor (IME) enter key was not functioning correctly while a user was inputting a new task in the application interface, leading to poor usability for users with non-Latin or complex IME environments.

## Solution

The code has been modified to check if the IME input event is `isComposing` when the Enter key is pressed, and to call `addTask()` only if it is not.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/tasks/add-task-bar/add-task-bar.component.spec.ts
+++ b/src/app/features/tasks/add-task-bar/add-task-bar.component.spec.ts
@@ -600,33 +600,43 @@ describe('AddTaskBarComponent', () => {
   });
 
   describe('IME handling (Integration)', () => {
-    it('should not add a task when Enter is pressed during IME composition', () => {
+    let inputEl: HTMLInputElement;
+
+    beforeEach(() => {
       component.stateService.updateInputTxt('New Task');
       fixture.detectChanges();
+      inputEl = fixture.debugElement.nativeElement.querySelector('input');
+    });
 
-      const inputEl = fixture.debugElement.nativeElement.querySelector('input');
+    const dispatchEnterKeydown = (options: {
+      isComposing: boolean;
+      keyCode?: number;
+    }): void => {
       const event = new KeyboardEvent('keydown', {
         key: 'Enter',
-        isComposing: true,
+        isComposing: options.isComposing,
       });
+
+      if (options.keyCode) {
+        Object.defineProperty(event, 'keyCode', { value: options.keyCode });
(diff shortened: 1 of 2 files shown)


watanabeycr7:
A screenshot showing the behavior before and after the change.

before:

https://github.com/user-attachments/assets/08df6a36-d691-4ef8-8ae4-def27cb835b6

after:

https://github.com/user-attachments/assets/40af13b7-4e20-49ac-8d4d-2196bc307961

johannesjo:
Thanks for the fix and the added test coverage. I found one remaining IME edge case before this is fully robust: `onInputKeydown()` now checks `!event.isComposing`, but the mention directive used on the same input already treats IME input as `event.isComposing || event.keyCode === 229` (`src/app/ui/mentions/mention.directive.ts`).

Some IME keydown events can still be reported as the buffered `229` event even when `isComposing` is false, so this can still create a task while the user is confirming composition.

Could we mirror the existing guard here, e.g. require `event.keyCode !== 229` before adding the task, and add a spec for `Enter` with `isComposing: false` plus `keyCode: 229`? For the spec, it may be more stable to set `keyCode` via `Object.defineProperty(event, 'keyCode', { value: 229 })`.

watanabeycr7:
> Thanks for the fix and the added test coverage. I found one remaining IME edge case before this is fully robust: `onInputKeydown()` now checks `!event.isComposing`, but the mention directive used on the same input already treats IME input as `event.isComposing || event.keyCode === 229` (`src/app/ui/mentions/mention.directive.ts`).
> 
> Some IME keydown events can still be reported as the buffered `229` event even when `isComposing` is false, so this can still create a task while the user is confirming composition.
> 
> Could we mirror the existing guard here, e.g. require `event.keyCode !== 229` before adding the task, and add a spec for `Enter` with `isComposing: false` plus `keyCode: 229`? For the spec, it may be more stable to set `keyCode` via `Object.defineProperty(event, 'keyCode', { value: 229 })`.

@johannesjo 
Thanks for your review!

Added a check for `event.keyCode !== 229` and added a test case for it.

Also consolidated some redundant processing in the test cases.

johannesjo:
Thank you very much!

PR #7559, opened 2026-05-11
feat(schedule): add scheduling warnings for overlapping tasks and outside work hours

xsky1020:
## Summary

This PR adds scheduling warnings to make time conflicts and impractical scheduling choices more visible during task planning.

## Changes

- highlight tasks that overlap with other scheduled tasks
- show an `Outside work hours` label for tasks scheduled outside configured work hours
- show a warning on the Today view when the remaining estimate exceeds 8 hours
- add real-time warnings in the scheduling dialog when a selected time:
  - overlaps with another scheduled task
  - falls outside configured work hours

## Why

Currently, users can schedule tasks into conflicting or impractical time slots without immediate feedback. This PR adds lightweight warnings directly to the task list, Today view, and scheduling dialog so these issues are visible while planning.

## Implementation Notes

This PR is split into several commits:

1. task-level conflict and out-of-hours indicators
2. Today view overload warning for remaining estimate > 8h
3. real-time warning messages in the scheduling dialog
4. test fixes for the scheduling dialog selectors


johannesjo:
Review feedback after a few focused passes:

Thanks for working on this. Surfacing scheduling friction is useful, but I think the current UX is too strong for something users may legitimately do often. Overlaps and outside-hours scheduling should feel like informational planning hints, not warnings/errors.

Main things I’d like adjusted before merge:

1. **Tone/severity should be info-level, not warning-level.**
   The PR currently uses warn colors, `warning_amber`, red-ish borders, and a warning snackbar on save (`dialog-schedule-task.component.ts:491`, `dialog-schedule-task.component.scss:174`, `_task-base.scss:165`, `work-view.component.scss:91`). Since scheduling still succeeds and these choices can be intentional, please downgrade this to neutral/info styling. I’d also avoid the submit snackbar or make it much less intrusive; repeated snackbars for intentional scheduling choices will get annoying fast.

2. **New user-facing strings need translations.**
   The new copy is hardcoded in several places (`dialog-schedule-task.component.html:100`, `dialog-schedule-task.component.ts:494`, `task.component.html:55`, `work-view.component.html:40`) and the snackbar uses `isSkipTranslate: true`. Please add keys to `en.json` and use `T` / translation pipes or `TranslateService` consistently.

3. **Conflict highlighting only works for today.**
   `TaskComponent` computes conflicts via `DateService.isToday` and `hasTimeConflict()` repeats the same today guard (`task.component.ts:224-236`). The dialog can detect overlaps for a future selected day, but after scheduling, those future conflicts won’t be highlighted in the task row. This should compute against the task’s own scheduled day/range, not only logical today.

4. **The conflict calculation is duplicated per rendered task row.**
   Every `task` component creates its own `timeConflictTaskIds` computed and scans/sorts all due-time tasks (`task.component.ts:154-157`, `task.component.ts:224-236`, `get-time-conflict-task-ids.ts:10-29`). On large lists this can become expensive because the same global conflict set is recomputed once per rendered row. Please move this to a shared selector/service-level computation, or otherwise compute once per relevant day/list and pass/look up the result.

5. **Cross-midnight overlaps are missed.**
   `getTimeConflictTaskIds()` filters tasks by the start timestamp’s day before comparing intervals (`get-time-conflict-task-ids.ts:10-12`). A 23:30-01:30 task and a 00:30 task overlap, but the prior-day task is filtered out for the next day. It would be more robust to filter by interval overlap with the selected day/range, not only by start day.

6. **Please add coverage for the new warning paths.**
   The dialog specs currently mock `selectAllTasksWithDueTimeSorted` as `[]` and `selectTimelineConfig` as `null`, so they don’t exercise overlap detection, excluding the edited task, enabled work-hours config, or snackbar/inline hint behavior. `is-task-outside-work-hours.ts` also has no direct spec for boundaries, disabled config, zero/min duration, and crossing work end.

Small accessibility note: if the task-row conflict marker remains visual-only, please add an accessible text/tooltip/aria affordance so the information isn’t only color/border/icon based. The dialog hint would also be better as an info-level polite live region if it updates while changing time.

johannesjo:
We also want to integrate this in the existing in the existing snack, since otherwise they overwrite each other, which is not what we want.

xsky1020:
Thanks for the detailed review. I updated the PR to address the feedback:

- changed the scheduling indicators to neutral/info-level styling
- moved new user-facing strings to translations
- removed the intrusive snackbar on submit
- made conflict detection work for scheduled tasks beyond today
- moved conflict detection to a shared selector-level computation
- handled cross-midnight overlaps
- added specs for overlap detection, excluding the edited task, work-hours boundaries, and disabled config
- added tooltip/aria/live-region affordances for accessibility

I also rebased the branch onto the latest master and verified lint and tests locally.

johannesjo:
Thank you very much! <3

PR #7564, opened 2026-05-12
feat(settings): add image picker for background image selection

### LINKED ISSUE #7515: 💡 Custom background: Add a "Select" button

Hetsavani:
## Problem

we don't need to type in the file path manually.
fixes : #7515 

## Solution

Adds a “Select” button for background image selection so users can choose an image directly from the file picker instead of manually pasting the local image path.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for the PR! Reviewed the diff — feature is useful, but a few concerns before merge.

## Critical

1. **Data URL stored in synced state** — `formly-image-input.component.ts:48` sets the formControl value to the full base64 data URL from `FileReader.readAsDataURL`. That value lands in `WorkContextThemeCfg.backgroundImage{Dark,Light}` (typed `string | null`), which is persisted to IndexedDB and included in the sync payload. A 3 MB photo becomes ~4 MB base64 that re-syncs on every related change. Suggested directions:
   - Electron: use a native file-dialog IPC and store the resulting `file://` path (no embedded bytes).
   - Web/Capacitor: enforce a strict size cap (e.g. 200 KB) and reject larger files with a snack message, or skip the feature on these platforms.
   - At minimum, validate `file.size` before reading and surface an error.

2. **Direct DOM manipulation** — `document.createElement('input')` + `document.body.appendChild/removeChild` violates the project's "no direct DOM access" rule (see `CLAUDE.md` anti-patterns). Prefer a hidden `<input #fileInput type="file" hidden>` in the template plus `viewChild('fileInput')` + `fileInput().nativeElement.click()`.

3. **Orphaned `<input>` on cancel** — `removeChild` only runs inside `reader.onload`. If the user cancels the picker, the `change` event never fires and the appended `<input>` stays in `document.body`. Each click adds another. The viewChild pattern above eliminates this entirely.

## Important

4. **`FileReader.onerror` not handled** — silent failure if the read errors (permissions, aborted). Wire `reader.onerror` and surface an error (the project uses `SnackService` for this).

5. **Untranslated label** — `Select` is a raw string in `formly-image-input.component.html`. Project rule: UI strings go through `T` / `TranslateService`, with `en.json` the only locale edited. Add a translation key.

6. **No `IS_ELECTRON` gate / no size limit** — combine with (1). The button currently runs on web/mobile too, where embedding a multi-MB image in app state is even worse.

7. **No tests** — adding a spec covering the success path, cancel path, oversized-file rejection, and `onerror` would be valuable; there's no existing spec for this component, so this is a good chance to introduce one.

## Minor

- `formly-image-input.component.ts:48-50` — `setValue` already dirties the control; the explicit `markAsDirty`/`markAsTouched` are likely redundant.
- `accept="image/*"` allows SVG. As a CSS `background-image`, SVG cannot execute scripts, but it widens the size-bloat risk further.
- Verify spacing/wrapping when both `Select` and `Load from Unsplash` are visible — they share the `.unsplash-button, .select-button` rule but have no inter-button margin.

## Verdict

Needs changes. The data-URL-in-synced-state issue (#1) is the main blocker — please rework that before merge. The DOM-pattern and cancel-leak fixes (#2, #3) likely fall out naturally once you switch to a viewChild-driven hidden input. Translation and tests are quick follow-ups.

johannesjo:
Thanks for the follow-up. Most of the earlier review items are addressed, but I still see two merge-blocking issues.

1. `electron/local-file-sync.ts:23` reads the selected background file synchronously and without any size guard. The Electron path now bypasses the 200 KB limit used by the browser `FileReader` path, so selecting a large photo can block the main process and allocate a huge base64 string whenever the app starts or the theme/background is resolved. Please `stat` the file first, enforce the same size/type limits before reading, and preferably use async fs APIs.

2. `src/app/ui/formly-image-input/formly-image-input.component.ts:64-65` manually builds a `file://` URL, and `electron/local-file-sync.ts:19-20` parses it with `decodeURIComponent(...)`. Valid filenames containing `%` will fail with `URI malformed`, and standard file URLs like `file:///C:/...` are easy to parse incorrectly. Please use proper file URL conversion/parsing, e.g. `pathToFileURL()` when creating the URL and `fileURLToPath()` when reading it.

Also worth adding a test for the actual Electron path: dialog-selected path -> stored URL -> local image read. The current spec mostly covers the hidden browser file input path, which is not the visible Electron button flow.

Checks run:
- `npm run test:file src/app/ui/formly-image-input/formly-image-input.component.spec.ts`
- direct ESLint/Stylelint checks on the changed files
- `npm run build`

johannesjo:
Thank you for the follow-up. One small remaining item: `READ_LOCAL_IMAGE_AS_DATA_URL` should reject unsupported file extensions before reading the file.

Right now unknown extensions fall through to `application/octet-stream`, so a manually entered `file://` background value could read any small local file into the renderer. I don’t think we need anything heavy here, but a simple image-extension allowlist before `readFile` would make this tighter and match the intent of the API.

For example, resolve the MIME type first and return `null` if it’s not one of the supported image extensions, then read the file only after that.

johannesjo:
Thank you very much! <3

beerkumquatpome:
v18.6.0

Correct path: /home/file.png

Linux Flatpak
It has the wrong path after selecting an image: file:///run/user/1000/.../file.png

Linux Appimage
Correct path but I have to manually remove "file://": file:///home/file.png

Hetsavani:
> v18.6.0
> 
> Correct path: /home/file.png
> 
> Linux Flatpak It has the wrong path after selecting an image: file:///run/user/1000/.../file.png
> 
> Linux Appimage Correct path but I have to manually remove "file://": file:///home/file.png
> 
> @Hetsavani

Thanks for testing this across Linux environments.

I’ll investigate where the selected background image path is consumed/rendered. It looks like there may still be a code path expecting a raw filesystem path instead of a `file://` URL, especially in the AppImage flow.

beerkumquatpome:
- @Hetsavani Is this related to #8455?

PR #7578, opened 2026-05-13
feat(plugins): add onReady() API with IPC ping + fix consent write delay

### LINKED ISSUE #7326: executeNodeScript not available when plugin.js runs on startup — no ready signal for permission-gated APIs
### Description

Plugins that require `nodeExecution` permission and call `executeNodeScript()` during initialization face a race condition: `plugin.js` executes before the permission is actually granted, causing the call to fail.

The only workaround is retry-with-backoff, but there's no way to know *when* the permission becomes available. The timing varies significantly — especially on macOS cold boot (after restart), where it can take 10-15+ seconds.

### Steps to Reproduce

1. Create a plugin with `"permissions": ["nodeExecution"]` in manifest.json
2. Call `PluginAPI.executeNodeScript(...)` early in `plugin.js`
3. Restart the machine (cold boot) and launch Super Productivity
4. The call fails — the plugin is non-functional until manually toggled off/on

### Expected Behavior

Either:
- `plugin.js` execution is deferred until all declared permissions are ready, or
- A `PluginAPI.onReady()` callback / `PLUGIN_READY` hook is provided so plugins can wait for permission-gated APIs

### Current Workaround

Retry with exponential backoff (~60s window). Works, but is fragile and wastes resources polling for readiness.

```js
// Current workaround in plugin.js
const MAX_RETRIES = 10;
let delay = 1000;
for (let attempt = 1; attempt <= MAX_RETRIES; attempt++) {
  try {
    await PluginAPI.executeNodeScript({ script: '...', args: [], timeout: 10000 });
    break; // success
  } catch (e) {
    if (attempt < MAX_RETRIES) {
      await new Promise(r => setTimeout(r, delay));
      delay = Math.min(delay * 2, 10000);
    }
  }
}
```

### Suggested Solutions (any of these would help)

1. **Defer `plugin.js` execution** until declared permissions are granted
2. **Add `PluginAPI.onReady(callback)`** that fires when all permissions are available
3. **Add a `PLUGIN_READY` hook** to the existing hook system
4. **Document the timing behavior** so plugin authors know to expect this

### Environment

- Super Productivity: 14.x (Mac App Store)
- macOS (Apple Silicon)
- Plugin uses `nodeExecution` permission

b0x42:
Closes #7326

Two bugs fixed and one new API added. Plugins with `nodeExecution` permission no longer silently fail to load on cold boot — the app now pings the IPC bridge before handing control to the plugin, with automatic retry, and shows a clear error if the bridge never comes up. Consent granted via the UI is now written immediately instead of after a 5-second delay, so quitting the app quickly after granting permission no longer causes the plugin to re-prompt on next launch. The new `plugin.onReady()` API gives plugin authors a clean, guaranteed-safe entry point for startup code, replacing the fragile pattern of calling `executeNodeScript` at the top level of `plugin.js`.

---

## What & Why

**1. Remove `setTimeout(5000)` from consent write**

The comment said "avoid sync conflicts during startup" but `_getNodeExecutionConsent` is only reachable via a UI toggle — a user action that can't happen during the sync window. The delay was a real bug: quit within 5s of granting consent and it was never persisted. Direct `await` is correct.

**2. `plugin.onReady(fn)` API**

`nodeExecution` plugins were calling `executeNodeScript` at the top level of `plugin.js`, before the Electron IPC bridge was ready. The fix inverts control: the app fires `onReady` only after confirming the bridge is up. Plugin code never needs to know about readiness timing.

A hook (`PLUGIN_READY`) would require broadcasting through `PluginHooksService` with per-plugin filtering. A direct callback on the `PluginAPI` instance is simpler — isolation is free because each instance is already per-plugin.

**3. IPC ping before firing `onReady`**

`setTimeout(0)` (the original proposal) doesn't help when cold-boot failures take several seconds. A real ping — `executeNodeScript({ script: 'return true' })` — confirms the bridge actually responds. The script routes through the `executeDirectly` vm path in the Electron executor (no child process spawn). 3 attempts with 1s/2s delays covers the observed cold-boot window (~3s) without hanging indefinitely. On failure: error snack + plugin status set to `'error'`, no silent skip.

**4. `_triggerReady` removed from `PluginAPI`**

The original implementation stored the ready callback on `PluginAPI` and exposed `_triggerReady()` as a public method, meaning plugin sandbox code could call `plugin._triggerReady()` directly. The callback is now registered via a function passed at construction time and stored in `PluginRunner`'s own map — `PluginAPI` has no internal state for it and no callable method.

**5. Ping goes through `PluginBridgeService`, not `api.executeNodeScript`**

`api.executeNodeScript` is only bound when the plugin has `nodeExecution` permission in Electron. The ping is a bridge health check, not a plugin operation — it should work regardless of permission binding. `PluginBridgeService.pingNodeBridge()` calls `_executeNodeScript` directly.

## Backward compatibility

`onReady` is optional. Existing plugins load identically. The ping adds 0–3s latency to `nodeExecution` plugin activation on cold boot only; warm boot succeeds on the first attempt.

## Tests

- `plugin-api.spec.ts` — `onReady` registers callback via the registration function
- `plugin-runner.spec.ts` — `triggerReady`, `pingNodeBridge` (including bridge-returns-true case)
- `plugin-consent.spec.ts` — consent written immediately on grant/deny/no-remember (no setTimeout)
- `plugin-ping-node-bridge.spec.ts` — retry loop: succeeds on attempt 1/2/3, throws after all 3 fail

## Manual test plan

**Consent fix:** grant consent + check Remember → quit immediately → relaunch → plugin loads without re-prompting.

**onReady cold boot:** wrap `executeNodeScript` in `plugin.onReady(async () => { ... })` → cold boot → plugin initialises without error.

**Ping failure UX:** force bridge unavailable (web build with `nodeExecution` declared, or mock `pingNodeBridge` to return false) → enable plugin → expect error snack + error state in plugin UI, `onReady` not called.


johannesjo:
I found a few issues that should be addressed before merge:

1. `onReady` is only wired through `_loadPluginLazy()` (`src/app/plugins/plugin.service.ts:556`), but several load paths still call `_pluginRunner.loadPlugin()` directly (`:709`, `:1228`, `:1519`). That means enabled plugins loaded during startup, upload, or reload bypass the new IPC ping and never fire `plugin.onReady()`, which misses the cold-boot case this PR is meant to fix. Please centralize the post-load readiness step and call it after every successful plugin load path.

2. If `_pingNodeBridge()` or the `onReady` callback throws, `activatePlugin()` marks the plugin as `error`, but the runtime has already been registered in `PluginRunner` and may have active hooks/buttons/side effects (`src/app/plugins/plugin.service.ts:556`, `src/app/plugins/plugin-runner.ts:112`). This can leave a partially running plugin while the UI says activation failed. The error path should tear down/unload the plugin runtime before setting the error state.

3. `PluginAPI.onReady` is now a required public API method (`packages/plugin-api/src/types.ts:389`), but iframe plugins still receive a `window.PluginAPI` object without `onReady` (`src/app/plugins/util/plugin-iframe.util.ts:278`). A typed iframe plugin can compile against the new API and fail at runtime. Either implement `onReady` in the iframe bridge or make/document it as optional/background-plugin-only.

Also noted: the “Plugin Tests” workflow is currently failing before tests run because `npm ci` reports `package.json` and `package-lock.json` are out of sync around `minimatch`.

b0x42:
Thanks for the review! Addressed all three in `425598522`:

**1. Centralized readiness across all load paths**
Extracted a `_fireOnReady(instance)` helper in `plugin.service.ts` and call it after every successful `_pluginRunner.loadPlugin()` — `_loadPluginLazy`, the startup load path (~line 720), `uploadPlugin`, and `reloadUploadedPlugin`. The ping + `onReady` now fire consistently regardless of how the plugin got loaded.

**2. Tear down runtime on activation error**
`activatePlugin`'s catch now calls `_pluginRunner.unloadPlugin(pluginId)` before setting `status: 'error'`. Wrapped in its own try/catch so a failing unload still results in the error state being set.

**3. `onReady` in iframe `PluginAPI`**
Added to `plugin-iframe.util.ts`. For iframe plugins the API bridge is fully wired before `plugin.js` runs, so `onReady` fires on the next microtask after registration.

---

**Self-review followups in `d43be4d9e`** (caught a few more issues on a second pass):

**A. Half-loaded plugins on `_fireOnReady` failure**
The error-path teardown I added in (2) only fired in `activatePlugin`. The other three load paths (startup, `uploadPlugin`, `reloadUploadedPlugin`) had outer catches that only logged + rethrew, so a ping failure mid-load would leave the plugin pushed into `_loadedPlugins` with hooks/buttons registered while the UI showed nothing wrong. Added `_fireOnReadyWithCleanup` + `_handleReadyFailure` helpers — used by all three; activation path keeps its own (slightly different) cleanup since it owns more state.

**B. Tests now exercise production code**
The retry-loop test was previously an inline-replicated stub of `_pingNodeBridge`'s logic — would silently pass if the production retry constants changed. Extracted the retry loop into a pure `pingWithRetry()` utility (`util/ping-with-retry.util.ts`); spec imports and tests the real function. The other private-method specs (`_getNodeExecutionConsent`, `_fireOnReady`) are kept as stub-replicas for now to avoid the full PluginService DI graph; happy to do a deeper refactor (extract those to standalone helpers too) if you'd prefer.

**C. Iframe `onReady` semantic documented**
Iframe `onReady` fires on the next microtask without an IPC ping — different from host. This is fine in practice because iframe plugins render on user navigation, well after startup. Documented both in the source (with a note explaining the conditions under which this assumption could break) and in `docs/plugin-development.md` for plugin authors.

---

**Re: the "Plugin Tests" workflow / `minimatch` lockfile mismatch** — that one is in the workflow itself (`npm ci` runs in the plugin tests CI before any of this PR's code is reached), so I'll leave it for a separate fix unless you'd like me to take a look.

**Re: the `electron/tsconfig.electron.json` `paths` change in `50753babe`** — this was needed to unblock the Linux Electron CI build after merging master (the new `@sp/sync-providers` subpath exports use `.d.mts`, which `moduleResolution: "node"` can't resolve). It's a band-aid: hardcodes paths to `dist/*.d.ts` files, will silently break if the package adds new subpaths. The proper fix is to either split the electron tsconfig so it doesn't include `src/app` and switch to `moduleResolution: "bundler"`, or have `@sp/sync-providers` ship `.d.ts` types alongside `.d.mts`. Happy to revert this commit and track separately if you'd rather not bundle infra fixes here.

johannesjo:
Thank you @b0x42 ! Sorry about the build failures. Its the first time the new plugin checks run in production. I will try to ajdust here in the branch directly. Hope that this is alright with you!

b0x42:
> Thank you @b0x42 ! Sorry about the build failures. Its the first time the new plugin checks run in production. I will try to ajdust here in the branch directly. Hope that this is alright with you!

Sure, thanks!

johannesjo:
Thank you very much for this!

PR #7579, opened 2026-05-13
feat(focus-mode): show pomodoro timer in browser tab title

### LINKED ISSUE #7552: 💡 Show Pomodoro Timer Countdown in Browser Tab Title
### Problem Statement

Currently, the web app does not display the active Pomodoro timer countdown in the browser tab title.

As a result, users need to switch back to the tab repeatedly to check the remaining time, which can interrupt workflow and reduce focus while working in other tabs or applications.

### Possible Solution

Display the active Pomodoro timer countdown directly in the browser tab title.

Example:

- (24:15) Super Productivity
- (05:00 Break) Super Productivity

This would allow users to monitor the remaining Pomodoro time without needing to constantly switch back to the app tab, helping maintain focus and reducing distractions.

Browser notifications and sound alerts help partially, but they do not provide continuous visibility of the remaining timer duration.

Showing the countdown directly in the tab title would be the most convenient and lightweight solution.

### Additional context

We shouldn't getting lost our focus during tab change to see the timer.

Many productivity and Pomodoro applications already provide timer visibility in the browser tab title or window title, which greatly improves usability during multitasking. such as : https://www.pomofox.com/

Hetsavani:
## Problem

Currently, the web app does not display the active Pomodoro timer countdown in the browser tab title.

As a result, users need to switch back to the tab repeatedly to check the remaining time, which can interrupt workflow and reduce focus while working in other tabs or applications.

Fixes: #7552 

## Solution

Display the active Pomodoro timer countdown directly in the browser tab title.

This would allow users to monitor the remaining Pomodoro time without needing to constantly switch back to the app tab, helping maintain focus and reducing distractions.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for working on this. I think there are a couple of things to address before this is ready:

1. `src/app/core/browser-title/browser-title.service.ts`: the title always uses `timeRemaining()`. In Pomodoro manual-break-start/overtime mode the timer keeps running after the target duration, but `selectTimeRemaining` clamps to `0`, so the tab title will stay at `(00:00) Super Productivity` for the whole overtime period. The focus UI handles this by switching to elapsed time when `isInOvertime()` is true; the browser title should use the same state/display rule.

2. `src/app/core/browser-title/browser-title.service.ts`: `Paused` and `Break` are hardcoded user-facing strings. Repo guidance says UI text should go through `T` / `TranslateService`, so these should use translation keys, with the spec updated accordingly.

CI note: the current `CI / Tests` job is failing in `Network Failure Integration > should merge remote changes received after offline period` with `executing a cancelled action`. That looks unrelated to this PR touched files, while lint, preview build, CodeQL, and Electron smoke passed, but the failing check still blocks merge until rerun or fixed.

johannesjo:
Thank you very much! <3

PR #7582, opened 2026-05-13
Plugin tag ids

powerpaul17:
## Problem

I want to be able to synchronize tags in plugin issue providers.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
I reviewed the updated head (`258fcc8`). The tag-sync direction makes sense, but I found a few issues that should be addressed before this lands:

1. `PluginIssueProviderAdapterService._mapLabelsToTagIds()` uses `this._tagService.tags()`, which includes the virtual `TODAY_TAG`. A remote label named `Today` can therefore resolve to the id `TODAY` and get written into `task.tagIds`, violating the project invariant that Today is never stored on tasks. Please filter out `TODAY_TAG.id` or use the no-My-Day tag list for this mapping.

2. The pull path special-cases `tagIds` and assumes `freshSyncValues[mapping.issueField]` is already a `string[]` of labels. That bypasses `mapping.toTaskValue`, so providers that expose labels as objects/category IDs or rely on `toTaskValue` for normalization cannot use this field correctly. The plugin mapping should still own the remote-shape conversion before the app maps label strings to local tag ids.

3. `createPluginSyncAdapter.extractSyncValues()` sorts tag label arrays only in the fallback extraction path. If a plugin provides its own `extractSyncValues`, the new array equality can still treat the same labels in a different order as a provider-side change and skip pushes. Please normalize the tag issue field after custom extraction too.

Small repo-convention note: `CLAUDE.md` says to edit only `en.json` for translations, so the `de.json` changes should probably be dropped unless maintainers want manual locale edits here.

johannesjo:
Thank you @powerpaul17 can you have a look at the failing tests?

johannesjo:
I re-reviewed the current head (`34b72f6`). This is much closer, and the current GitHub checks are green. I found two remaining issues that look worth fixing before merge:

1. `tagIds` changes from `TaskSharedActions.addTagToTask` are not pushed. `IssueTwoWaySyncEffects` only subscribes to `updateTask`, short syntax, scheduling, and planner actions (`src/app/features/issue/two-way-sync/issue-two-way-sync.effects.ts` around the `ofType(...)` list), but creating a new tag from the task tag dialog dispatches `TaskSharedActions.addTagToTask({ tagId, taskId })` (`src/app/features/tag/dialog-edit-tags/dialog-edit-tags-for-task.component.ts:90`). That updates local `task.tagIds` without entering the new push-sync flow. Please either include `addTagToTask` in the effect/extractor path and use the current task's `tagIds`, or route that UI path through the existing `updateTags`/`updateTask` path.

2. The polling pull path still stores unnormalized sync baselines. `PluginIssueProviderAdapterService.getFreshDataForIssueTask()` calls `provider.definition.extractSyncValues?.(issue)` directly and writes that value into `issueLastSyncedValues` (`src/app/plugins/issue-provider/plugin-issue-provider-adapter.service.ts:189-211`). The push path, however, normalizes tag label arrays via `createPluginSyncAdapter.extractSyncValues()` (`src/app/plugins/issue-provider/plugin-sync-adapter.service.ts:120-121`). If a provider returns labels in remote order, polling can store e.g. `["feature", "bug"]`; a later push fetches the same labels as the normalized `["bug", "feature"]`, and the order-sensitive array comparison treats the provider as changed and skips the tag push. The pull path should use the same adapter/normalization before storing the baseline.

I also ran `npm run test:file src/app/features/issue/two-way-sync/issue-two-way-sync.effects.spec.ts` locally on the PR head, and it passed.

johannesjo:
Thank you very much! <3

powerpaul17:
Thank you @johannesjo for the quick reply and review. It was my first PR for this app, so my understanding of the codebase is not very good yet.

johannesjo:
> Thank you @johannesjo for the quick reply and review. It was my first PR for this app, so my understanding of the codebase is not very good yet.

You did very well! Thank you! It is a surprsingly complex subject... :)

Issue #7349, opened 2026-04-24
Enhance Pomodoro Timer Visibility on Focus Page


aswinharikrishnan:
### Description

The Pomodoro timer on the focus page appears relatively small, reducing its visibility during active work sessions.

### Problem

* Timer is not visually prominent
* Requires effort to read during deep focus
* Weak visual hierarchy compared to other UI elements

### Expected Behavior

* Timer should be larger and more central
* Easily readable at a glance during focus sessions

### Suggested Improvement

* Increase timer font size (e.g., 1.5x–2x)
* Improve positioning to make it the primary focal element
* Optional: allow user customization for timer size

### Additional Context

Improving the visibility of the timer can enhance usability and reduce cognitive load during deep work.

[](url)

### Url
file:///C:/Program%20Files/WindowsApps/53707johannesjo.SuperProductivity_18.2.8.0_x64__ch45amy23cdv6/app/resources/app.asar/.tmp/angular-dist/browser/index.html#/tag/TODAY/tasks

### Meta Info
META: SP18.2.8 __ Electron – en-US – Win32 – en-US – UA:Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) superProductivity/18.2.8 Chrome/146.0.7680.179 Electron/41.2.0 Safari/537.36

### Actions Before Error
```

```

Issue #7581, opened 2026-05-13
💡 Flowtime: Better default settings


beerkumquatpome:
### Problem Statement

- 

### Possible Solution

1. Enable "Flowtime breaks" by default
- It uses the name "Flowtime" not "Stopwatch":  Flowtime = Focus time/5
- If this setting is disable by default, hardly anyone will enable it manually

2. Use ratio-based calculation by default
 
3. Add more default rules for ruled-based mode:

5 - 15 3 mins
15 - 25 5 mins
25 - 50 10 mins
50 - 75 15 mins
75+  20 mins

johannesjo:
Thanks for the detailed write-up. I agree there is a real bug here around persistence: Flowtime settings should not be cleared just because "Flowtime breaks" is temporarily disabled. We should preserve `breakMode`, `breakPercentage`, and `breakRules` while the fields are hidden.

Ratio-based calculation is already the default internally, and the current default percentage is `20%`, which matches the `focus time / 5` idea.

I am more cautious about enabling Flowtime breaks by default or adding too much behavior out of the box. Flowtime is not really a fixed spec in the same way Pomodoro is, especially around how breaks should be calculated. Different people use different ratios or rules. Because of that, I think we should prefer a minimal default that is easy to understand, and let users opt into more specific break behavior.

Adding better default rule presets sounds reasonable for users who choose rule-based breaks. I would probably make the last rule open-ended, e.g. `75+ min -> 20 min`, so longer sessions still get a matching break.

beerkumquatpome:
### Bug: Rule-based calculation - Moving forward/backward also changes the duration of other rules
1. Add 5 rules
2. Set min work duration to 0 (All of them)
3. Change the third rule to 1
4. Press Move forward/Move backward
5. Other rules also changes to 1

Suggestion: Just remove the Move forward and Move backward buttons

### Other suggestions:

1. Rule-based calculation: Support hours like Pomodoro
2. Rule-based calculation: Remove the 480 input limit (Pomodoro allows 999999+)
3. Rule-based calculation: When adding a new break rule, the "Max work duration" & "Break duration" fields should be empty
    Min work duration: It should automatically use the max work duration of the last rule
4. Rule-based calculation: If min work duration is longer than max work duration, it should show a warning
5. Ratio-based calculation: Remove the 100% limit
6. Ratio-based calculation: When the break percentage field is empty, the dialog should stay the same size
[...]

Issue #7539, opened 2026-05-10
🚨countdown time display issue on button of focus mode, when using pomodoro


chantsing:
### Version Used

18.5.0 on macos

### Current Behavior

<img width="540" height="112" alt="Image" src="https://github.com/user-attachments/assets/9b4d9ea9-8904-4a7c-b7b6-3d3c71b3bacd" />

### Expected Behavior

countdown time display is displayed fine

### Steps to Reproduce

1.using pomodoro

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

chantsing:
Thank you for your brilliant job
But when we start a focus session, the position of Enter focus mode will change. this is slightly weird and ungraceful (Not elegant)

https://github.com/user-attachments/assets/8034f3d7-c6c6-4f34-a5d5-179ec5c069aa

Could we increase the distance between the “Start/Stop Tracking” button and the “Focus Mode” button? This would prevent any movement in the icon’s position when starting tracking.

johannesjo:
Thank you very much! I look into it!

chantsing:
<img width="272" height="132" alt="Image" src="https://github.com/user-attachments/assets/ba16e23a-5977-40ce-874b-ce0a7949ff1f" />

Still small bug need to fix

PR #7586, opened 2026-05-13
feat(focus-mode): focus screen UX overhaul

miklosp:
# Focus mode UX overhaul

A pass over the Focus mode screen addressing #7349 and a handful of related
papercuts. Touches Pomodoro, Flowtime, Countdown, and the break/session-done
screens.

## Highlights

- Bug #7349 / "Pomodoro timer too small" - addressed.
- All four timer states now share a single visual chrome via a new `<focus-clock-face>` component.
- Clock face scales fluidly with the viewport (~2× larger on a typical laptop, capped on 4K) rather than stepping through discrete breakpoints.
- Consistent design and typography between 4 timers
- Select task has an obvious CTA (timer can't start before task is selected)
- Buttons moved outside of timer circle for better legibility (except for pomodoro +/-)
- New click-to-edit duration for Pomodoro
- Countdown drag scaling adapts to duration: 5-min increment under 1h, 15-min increment (and 6h-per-rotation) above 1h.
- Notes/attachments panel behaves like a modal - backdrop dims the timer, clicking outside closes; timer stays put.

## Small improvements
- Pause no longer hides the selected task, the title flicker is gone.
- Settings dialog for Flowtime shows all fields with proper disable/enable semantics and a stable dialog width when switching modes.
- Timer circle placement consistent between different timers
- Consistent "back to planning" behaviour

### Shared clock face
- New `src/app/features/focus-mode/focus-clock-face/` component
  (`<focus-clock-face variant="ring | breathing">`).
- Owns the SVG progress ring, breathing-dot for Flowtime, size tokens
  (`--clock-face-size`, `--clock-time-size`, `--control-offset`), and the
  shared `.clock-time` / `.clock-wrapper` typography via `:host ::ng-deep`.
- **Fluid sizing.** `--clock-face-size` is now a single
  `clamp(180px, min(54vmin, 58vh), 640px)` — replaces the discrete
  600×600 / 1280×900 breakpoints. Pairs `vmin` (clock is square) with a
  `vh` cap so squat (wide-short) windows don't push the task title /
  pause-complete row off-screen. Floor keeps it legible on tiny windows,
  ceiling prevents domination on 4K.
- **Single source of truth.** `--clock-time-size` and `--control-offset`
  are now derived via `calc(var(--clock-face-size) * 0.22)` and
  `... * 0.07)` respectively — no separate clamps. The 10 % overall
  reduction lives in the face's ceiling and floor; everything else
  follows.
- Flowtime's breathing dot grows to 92 % of the face and now sits inside a
  faint idle ring, so all three modes read at the same diameter.
- Both `focus-mode-main` and `focus-mode-break` consume this component;
  `mat-progress-spinner` and the duplicated SVG markup are gone.

### Countdown prep state
- Consistent circle outline design
- Slider sized to `--clock-face-size * 0.884` so the handle's orbit aligns
  with the ring's centerline.
- 30 px round accent-coloured handle sits on the ring; bigger than the
  stroke so the "drag me" affordance is unambiguous.
- New `[useFlexibleIncrement]` input on `<input-duration-slider>`:
  - Below 60 min ("Mode A"): 5-min snap, 1° = 1/6 min, full circle = 60 min.
  - At/above 60 min ("Mode B"): 15-min snap, 1° = 1 min, full circle = 6 h,
    top pinned to the 60-min boundary so A↔B is seamless.
  - Default off; other consumers keep their existing linear behaviour.

### Pomodoro prep
- Inherits the click-to-type input for duration; typed value persists to
  `pomodoroConfig.duration` via `globalConfigService.updateSection`.
- Drag handle hidden and the slider's host gets `pointer-events: none` so
  click-and-drag on the ring can't silently shift the value

### Task title row
- Switch/finish flank the title above the circle: `[swap_horiz] [title]
  [done]`.
- Extended FAB CTA replaces the dim placeholder when no task is selected:
  `[add_task] Select task to focus`.
- Pause-clears-title fixed: a `displayedTask` computed falls back to the
  paused task so the title stays put while `currentTask` is null. Same
  fallback applied to the parent-title, notes panel, attachments, drag-over
  prompt, and the related `changeTaskNotes` / `onDrop` handlers.

### Session controls
- New `.circle-controls` row below the circle holds pause/resume +
  complete-session + reset-cycles (Pomodoro only).
- `.bottom-controls` (issue link, notes button) and the title-row buttons
  fade together using a single `--revealed-opacity` driven by host hover.
- Visibility now also conditional on `document.hasFocus()` — when the
  browser window loses focus the entire muted control set fades.
- Paired `--revealed-pointer-events` so faded buttons can't latch onto
  tooltips and leave them stranded after the icon disappears.

### Layout stability
- `min-height` reservations on `.task-section`, `.play-actions`, and
  `.circle-controls` (using `var(--s8) / var(--s9)`) keep the circle in a
  fixed vertical position across no-task ↔ task-selected and prep ↔
  in-progress transitions.
- All recent literal `px` values for these slots use the spacing token scale
  (`--s5`, `--s8`, `--s9`).
- `:host` carries a constant `padding-top: var(--s9)` (reserving room for
  the absolute-positioned mode selector / simple-counter row) rather than
  toggling between states — removes the snap-down when starting a session.

### Notes / attachments panel (modal behaviour)
- Acts as a modal over the timer instead of pushing layout up. The clock
  stays put when the panel opens.
- New `.notes-backdrop` element fades in alongside the panel
  (`@slideInOutFromBottom`) and intercepts clicks outside the panel to
  close it. Same affordance as the existing `expand_more` mini-fab at the
  top of the panel.
- `:host { z-index: 0; }` establishes a stacking context so the backdrop's
  `z-index: 20` and the panel's `z-index: 21` stay contained inside
  `focus-mode-main` — the parent overlay's close (X) button at z 5 remains
  clickable.
- Removes the previous 100 px `@expand` spacer and `expandAnimation`
  import in `focus-mode-main`.

### Pomodoro cycle counter
- Small `<repeat>` icon + cycle number inside the circle, just below the
  digits, on both the focus and break timers.
- One step less intense than the digits (`var(--text-color-muted)`).
- Break timer uses `getBreakCycle()` (= `Math.max(currentCycle - 1, 1)`) so a
  cycle = focus + trailing break: Focus 1 shows "1", Break 1 still shows "1".

### Break screen
- Layout now mirrors focus-mode-main so the circle/digit baseline doesn't
  jump on the focus → break transition.
- Task name above the circle (`displayedTask` resolved from
  `pausedTaskId`).
- "Short break" / "Long break" / "Take a break" label inside the circle at
  top: 30 % — bold, `var(--s3)` font, muted colour.
- Cycle suffix dropped from the `SHORT_BREAK_TITLE` / `LONG_BREAK_TITLE`
  translation strings; the cycle counter expresses that now.
- Old `<p class="break-message">Take a moment to relax</p>` paragraph
  removed.
- Pause icon out of the clock face; new `.circle-controls` row holds
  `pause/resume → skip → reset` (reset Pomodoro only).
- Bottom row holds `← Back to planning` plus a state-specific primary
  (Start break / Start next focus session) with generous top padding.

### Back-to-planning unification
- `focus-mode-break.exitToPlanning()` and
  `focus-mode-session-done.cancelAndCloseFocusOverlay()` now do the same
  thing: dispatch `cancelFocusSession()` + `Router.navigateByUrl` to
  `/project/INBOX_PROJECT/tasks`.
- Both buttons render identically: `mat-stroked-button` with `arrow_back`
  icon.

### Flowtime settings dialog
- All fields render at dialog open. `breakMode` (dropdown) is always
  visible; `breakPercentage` / `breakRules` are mutually exclusive based on
  the dropdown.
- Disable expression (`'props.disabled'`) on each editable field gates on
  `isBreakEnabled` — model values persist regardless of toggle state.
- Default to `breakMode: 'ratio'` so a fresh dialog shows the percentage
  field rather than nothing.
- Break-rule fields laid out in a row via `fieldGroupClassName: 'formly-row'`
  so the dialog width is stable when switching modes.
- Multiple rule rows stack vertically (override
  `RepeatSectionTypeComponent`'s 2-column grid for this case only).
- Field labels shortened: `(minutes)` → `(m)`.

### Cross-cutting fixes
- `arrow_backward` → `arrow_back` in `repeat-section-type` — the former
  isn't a valid Material Symbols ligature, so it was rendering as a glitched
  fallback glyph in every repeat-type context menu (boards, simple
  counters, take-a-break, flowtime rules).
- `--clock-time-size` mirrored on `.clock-display` so sibling overlays (the
  duration slider) can read the same font scale as the clock-face.

### Shared layout shell (`focus-mode-layout`)
- New presentational `focus-mode-layout` component
  (`src/app/features/focus-mode/focus-mode-layout/`) owns a 4-row
  content-projection skeleton: `[fmTop]` (mode selector / counters, reserved
  height), `[fmTask]` (title), `[fmClock]` (the only flexing row, centers the
  clock), `[fmBottom]` (action row).
- `focus-mode-main` and `focus-mode-break` both consume it instead of each
  carrying their own absolute layout. The reserved top-row height holds the
  task/clock baseline steady across focus ↔ break and prep ↔ in-progress — the
  same goal as the earlier per-screen `min-height` work, now in one place.
- Bottom controls reordered with "Back to Planning" leftmost; the
  `BACK_TO_PLANNING` label is back to "Back to Planning".

### Flowtime: auto-start the break (credit #7745 / @Hetsavani)
Mirrors @Hetsavani's #7745 "auto start flowtime breaks" — credit to that PR for
the behaviour. Reimplemented on top of the layout work here rather than merged,
and without the opt-out flag.
- Ending a Flowtime session now starts the break automatically instead of
  showing a separate "break offer" screen.
- `endFlowtimeSession` dispatches `completeFocusSession({ isManual: false })` +
  `startBreak` (unsetting the current task first when "pause tracking during
  break" is on), matching Pomodoro's `autoStartBreakOnSessionComplete$`. The
  session is logged once, via `completeFocusSession`.
- Removes the `BreakOffer` UI state, the `offerFlowtimeBreak` action / reducer
  case, and the now-orphaned `FLOWTIME_BREAK_TITLE` / `START_BREAK` i18n keys.

### Test isolation fix
- `layout.service.spec` overrode `document.activeElement` with
  `Object.defineProperty` and never restored it, so the override leaked into
  later specs and broke `task.service`'s `focusTaskById` tests (#7120) depending
  on Karma's spec order. Its `afterEach` now deletes the shadowing property.
  Separate `test(layout):` commit.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have included relevant changes to the documentation/wiki — Flowtime now auto-starts the break instead of offering it (user-visible behavior change); a docs follow-up may be warranted
- [x] I have added tests for my changes (focus-mode specs updated for the layout shell + Flowtime auto-break; separate `test(layout):` fix)


johannesjo:
Thank you very much! I like it quite a bit. Maybe it is a tad to large on larger screens? So maybe we should cap the scaling at some point?

johannesjo:
Also I think the Play button is slightly off center horizontally.

miklosp:
Double checked, but it should be center
image

johannesjo:
@miklosp thanks for checking. Had some earlier version were the repeat icon for pomodoro looked a bit different, but now it looks good! 

What do you think about capping the size (or adjusting the scaling a bit)? To me the clock looks quite big and everything a bit tiny in comparision.

miklosp:
I'll address the feedback as soon as I have some time in the coming days, but won't promise to implement all.
I'll see if I can include the following too:
#7581 #7587 #7581 #7505 #7311 #7109 #6126 #7612 #5500 #7112 #7539 #7384

Partially addressed already:
#7403 

Overlay related, probably won't touch:
#6948  #6550 #4643

johannesjo:
Maybe a or multiple follow up issues is a better place to address the more complex stuff? There are already 20 files changed with this. Probably better to keep it focused.

chantsing:
Thank you for your brilliant job, this issues is be improved. [7539](https://github.com/super-productivity/super-productivity/issues/7539)

https://github.com/user-attachments/assets/29c5f724-42a1-477c-91be-33e88b549e30

But when we start a focus session, the position of  Enter focus mode will change. this is slightly weird and ungraceful (Not elegant)

Could we increase the distance between the “Start/Stop Tracking” button and the “Focus Mode” button? This would prevent any movement in the icon’s position when starting tracking.

johannesjo:
Nice work — this is a real improvement to the Focus screen, and it directly addresses #7349 and the "cap the scaling" feedback (clamp ceiling at 640px) plus the play-button centering. CI is green. A few things I'd suggest tightening before merge, grouped by priority.

### Worth addressing

**1. Pomodoro duration persists to synced config on every keystroke.**
`focus-mode-main.component.ts › onDurationChange()` calls `globalConfigService.updateSection('pomodoro', …)` for each `modelChange`. The `inputDuration` directive emits on every `input` event (`input-duration.directive.ts`, `@HostListener('input')`), and there's no debounce in `global-config.effects.ts`, so typing `25` dispatches multiple `updateGlobalConfigSection` actions — i.e. several synced writes for one edit. Could you debounce, or commit on blur/Enter? (Countdown is fine — it dispatches the non-synced `setFocusSessionDuration`.)

**2. The shared `input-duration-slider` is restyled via local `::ng-deep`.**
`focus-mode-main.component.scss` reaches into the shared `src/app/ui/` component's internals (`::ng-deep .circle { … \!important }`, `.inner-circle`, `.dots`, custom `.handle::after`, `.value`). `CLAUDE.md` asks us to avoid locally restyling shared UI components and to make a reusable variant instead. You already added a clean `useFlexibleIncrement` input — the "bare ring" visual would ideally be an input/variant on the slider in the same spirit. Same pattern in `dialog-flowtime-settings.component.scss` (`::ng-deep … .list-wrapper { display: block \!important }` overriding `RepeatSectionTypeComponent`).

**3. New drag math has no tests.**
`_setValueFromRotationFlex()` is the riskiest new logic (raw-angle accumulator, repurposed `minutesBefore`, the A↔B boundary anchoring at 55/60 min), but `input-duration-slider.component.spec.ts` wasn't touched. Worth a few unit tests for the boundary crossings and the ±180° wrap hysteresis. (Minor: the checklist's "no logic changes" isn't quite accurate given this + the flex behavior.)

### Minor / cleanup

- **Orphaned action:** after unifying back-to-planning, `exitBreakToPlanning` is no longer dispatched anywhere — its action, the `stopTrackingOnExitBreakToPlanning$` effect, and the reducer case (+ its spec) are now dead.
- **Orphaned i18n keys:** `CONTINUE_TO_NEXT_SESSION` and `BREAK_RELAX_MSG` are no longer referenced.
- **i18n drift:** `LONG_/SHORT_BREAK_TITLE` dropped `{{cycle}}` in `en.json`, but other locales still contain it and the template no longer passes a `cycle` param — non-English users will see a literal `{{cycle}}` on the break label until locales are regenerated.
- **Two sources of truth:** the `clamp(180px, min(54vmin,58vh), 640px)` token is duplicated in `focus-clock-face.component.scss` and `focus-mode-main.component.scss` with a "keep in sync" note — a shared globals token would be safer.

### Looks good
- `cancelFocusSession` already sets `isOverlayShown:false` in the reducer, so dropping the redundant `hideFocusOverlay()` in session-done is safe.
- The `:host { z-index:0 }` stacking-context fix is sound (`:host` is `position:absolute`, so the backdrop stays contained below the overlay's close button).
- `displayedTask` pause-fallback cleanly fixes the title flicker and is applied consistently.
- `arrow_backward → arrow_back` is a legit fix (invalid Material ligature), and the icon-only buttons gaining `aria-label` (with the e2e role-query update) is a nice accessibility touch.

Agree with deferring the larger wishlist items to follow-up issues to keep this focused.

miklosp:
**Flowtime breaks now auto-start — the `BreakOffer` step is gone**

Flagging the one behaviour change in this PR that users will actually notice.

**Before:** ending a Flowtime session dropped you on a separate "break offer" screen, and you had to press start to begin the break.
**After:** the break starts on its own, the same way Pomodoro already does.

**Why:**

- Pomodoro and Flowtime were inconsistent. Same idea (session ends → break begins), two different flows. Pomodoro auto-started; Flowtime made you confirm. This lines them up.
- The offer screen was an extra click for something you'd almost always accept. If you don't want the break, you can still skip or go back from the break screen.
- Credit to @Hetsavani's #7745 ("auto start flowtime breaks") PR for the behaviour. I reimplemented it on top of the layout refactor here rather than merging, and left the opt-out flag off for now.

**Implementation note (for review):** `endFlowtimeSession` now dispatches `completeFocusSession({ isManual: false })` → (unset the current task, when "pause tracking during break" is on) → `startBreak`, mirroring Pomodoro's `autoStartBreakOnSessionComplete$`. The `isManual: false` and unsetting *before* `startBreak` both matter: with `isManual: true` the tracking-stop effect enqueues its own `unsetCurrentTask` after `startBreak`, which then pauses the break a second in. The session is logged once, via `completeFocusSession`. This removes the `BreakOffer` UI state, the `offerFlowtimeBreak` action/reducer, and the orphaned `FLOWTIME_BREAK_TITLE` / `START_BREAK` i18n keys.

One side effect worth flagging: the session-complete notification now also fires on the break path (consistent with Pomodoro); the old offer flow didn't show it.

miklosp:
Notes on design decisions:
- No drag the ring to set Pomodoro duration (like Countdown). Only click-to-type was added; dragging is explicitly disabled for Pomodoro. I don't feel strongly about it, but also don't think people mess a lot with their Pomodoro timer.
- "select a task first" behaviour kept, no auto-selecting next task either
- Renamed "Back to Planning" then reverted. Now it's always present in the bottom button row, and behaviour is uniform (back to inbox). I was tempted to change the close (X) button to a minimise button, but deferred for now.
- No thinner ring, but there are bigger buttons. Thinner ring might be more elegant, but it would make it less visible at glance and less in line with material/angular.

johannesjo:
Reviewed the full diff. The structural refactor is solid — shared `<focus-clock-face>` + 4-row `<focus-mode-layout>`, fluid `clamp(180px, min(54vmin, 58vh), 640px)`, `displayedTask` fallback eliminating the pause-flicker, the extended-FAB "Select task" CTA, notes-as-modal, and the `arrow_backward` → `arrow_back` glyph fix all land cleanly. Sync correctness verified: `autoStartFlowtimeBreakOnSessionEnd$` mirrors the Pomodoro pattern, action ordering (`completeFocusSession{isManual:false}` → `unsetCurrentTask` → `startBreak`) holds against `stopTrackingOnSessionEnd$` and `syncTrackingStopToSession$`. No CD regressions in the per-tick hot path.

Three behavior regressions worth resolving before merge, plus two scope cleanups.

### 1. Muted-control hover-reveal is keyboard-inaccessible
`focus-mode-main.component.scss:23-43` fades pause/complete/+/-/switch/finish/notes/issue/settings to `opacity: 0; pointer-events: none` unless `:host:hover`. There's no `:focus-within` / `:focus-visible` fallback, so keyboard-only users tab into invisible-and-unclickable buttons. The `&.isWindowBlurred:hover` clause also suppresses the reveal even when hovered, which is hostile on dual-monitor / glance-at-timer workflows. Touch is handled via `(hover: none)` — only mouse + keyboard users are stuck.

**Suggested:** add `:focus-within` to the reveal selector and drop the `isWindowBlurred` suppression (or downgrade it to a longer idle fade instead of hover-gated).

### 2. "Back to Planning" hard-codes INBOX_PROJECT
`focus-mode-main.component.ts:511`, `focus-mode-break.component.ts:137`, `focus-mode-session-done.component.ts:84` all call `router.navigateByUrl('/project/\${INBOX_PROJECT.id}/tasks')`. On master the equivalent paths only dispatched `cancelFocusSession`/`hideFocusOverlay` — no router navigation — so users stayed on their originating context. Focus mode can be entered from the header button, global shortcut, task context menu, idle dialog, and plugins, not just from the inbox. Forcing every exit to Inbox is a real navigation regression for anyone in a project or tag view.

**Suggested:** stash the origin route on session start (or just drop the navigation and let users stay where they were, matching master).

### 3. Flowtime auto-break removes the "ask me" middle ground
The previous BreakOffer screen rendered explicit Start / Skip / Exit buttons after a Flowtime session. This PR auto-starts the break gated only by the binary `isBreakEnabled`. Flowtime users specifically picked the flexible mode — taking away the per-session choice (vs the upstream PR's opt-out flag, intentionally omitted here per the description) is the wrong direction for that audience. Skip-after-it-already-started isn't the same mental model.

**Suggested:** add an `isAutoStartBreak`-style toggle to `FlowtimeConfig` (default on if you want to keep the new default), or keep the BreakOffer screen as an opt-in.

### Smaller items
- **Hybrid drag above 1h** snaps to 15-min increments (`input-duration-slider.component.ts:_setValueFromRotationFlex`). Typing still works for arbitrary values (the inputDuration directive accepts free-form), so this is documentation, not a blocker — but the user has no signal that drag granularity changes at the 1h seam. Consider a tick-mark or scale label.
- **Click-to-edit Pomodoro digits** rely on the `CLICK_TO_EDIT_DURATION` tooltip alone; the digits don't get a hover/cursor affordance. With adjacent buttons hover-faded (issue #1), the prep screen reads as inert. A subtle hover background on `.clock-time` when `isShowDurationSlider()` is true would carry the affordance.
- **Clock-size ceiling at 640px** is conservative for 4K; floor of 180px risks the `top: 65%` cycle counter colliding with the digits on ≤360px portrait. Worth a quick check on small phones.
- **`BREAK_RELAX_MSG` ("Take a moment to relax")** removal makes the break feel clinical; low cost to keep.

### Scope cleanup (pull into separate commits)
- **Docs prettier reformatting** across 6 files (`docs/plans/2026-05-21-…`, `2026-05-28-…`, `docs/sync-and-op-log/operation-log-architecture.md`, `docs/wiki/3.08-Sync-Integration-Comparison.md`, `packages/super-sync-server/docs/backup-and-recovery.md`, `packages/super-sync-server/prisma/migrations/README.md`) is purely cosmetic and unrelated to focus-mode.
- **`src/app/t.const.ts`** — board form keys (`ONLY_PARENT_TASKS`, `PROJECT*`, `SCHEDULED_STATE*`) and `ADD_TASK_BAR.TOOLTIP_CLEAR_DEADLINE` / `TOGGLE_ADD_TO_BACKLOG_TODAY` are alphabetically reshuffled with no semantic delta. The legitimate focus-mode key adds/removes belong in this PR; the reorders don't.

Nice work on the underlying refactor — these are mostly polish on top of a genuinely cleaner architecture.

miklosp:
@johannesjo and @beerkumquatpome thanks for the quick review. While I work on the fixes, I want to push back on one thing:
### Flowtime auto-break removes the "ask me" middle ground
Ask me middle ground doesn't make any sense. User can choose to allow breaks or not for flowtime. User can pause flowtime. User can skip the break. User can exit focus mode.
The point is this. The user is either in the flow or taking a break. There is nothing in between. Adding a config for everything will become overwhelming after a while, and I really don't see the point here. 

There is an issue for this: https://github.com/super-productivity/super-productivity/issues/7701 - but I don't think this should be a setting, unless I miss some use case?

johannesjo:
@miklosp You've convinced me on the Flowtime point – the offer screen was the friction #7701 complained about, skip/exit cover the real cases, and I'd rather not grow settings for a hypothetical user. If complaints show up after release we can add #7745's `isAutoStartFlowtimeBreak` flag back as a pure opt-out, so nothing is lost by shipping without it. Consider point 3 dropped.

The accessibility (`:focus-within`) and inbox-navigation points are the ones I'd still like resolved before merge. (I'll sort out #7745/#7701 attribution over on that PR.)

johannesjo:
Thank you for the update! 

I think positioning could be better for screens exceeding a certain height (e..g. 4K screens) as task title probably should be closer to the clock. Also we might want to cap clock size for it not to become anime ridiculous :)


Screenshot_20260610_230740
Screenshot_20260610_230630

johannesjo:
**`focus-mode-task-tracking`: the play/pause toggle is never wired up**

Nice extraction — the read-only time stack reads cleanly and the title-centering grid is solid. The interactive half of the widget, though, never renders or fires in the running app: both call sites render it read-only and `(toggleTracking)` isn't handled anywhere.

- `focus-mode-main.component.html:90` → `[showToggle]="false"`
- `focus-mode-break.component.html:18` → `[showToggle]="false"`
- neither parent binds `[isTracking]` or handles `(toggleTracking)`

So `isTracking`, the `@if (showToggle())` button, `.play-pause-btn`, and the new `RESUME_TRACKING` string are only exercised by the spec — nothing toggles in production.

The **break** staying read-only is correct (tracking is coupled to the break timer — the inline comment explains it well; leave it). The **focus session** is where it should be live: please wire `focus-mode-main`'s usage to real state + an action — `[isTracking]` from the tracking / session-paused signal and `(toggleTracking)` to the pause/resume-tracking handler — so the task-row play/pause actually does something.

One design check while wiring: the bottom control row already has a session pause/resume, so make sure the task-row toggle acts on *task tracking* specifically and isn't a confusing duplicate of it.

Everything else in this commit checks out — the navigation removal (overlay still closes via `cancelFocusSession` → `isOverlayShown: false`), the `safe-area-bottom` de-dup, the `:focus-within` reveal, and the `isShowPauseButton` simplification all verified correct.

miklosp:
@johannesjo ended up removing the time tracker play/pause button. Focus mode and time tracking is very tightly coupled at the moment as it is. It's possible to start time tracking without a focus session, but impossible to do a focus session without time tracking. There is a setting whether break is tracked or not, but any resuming of time tracking also resumes the focus session (skipping break). Any pausing of time tracking also pauses the focus session. It's sort of works intuitively, but also not really explained. Might be a separate PR for later.

I've tried to further optimise placement with center alignment and locking the task selector and button row to the circle. Also made the overlap higly unlikely by making the focus mode selector buttons less tall. I can imagine a different layout here, but left that can of worms for another day and another PR.

johannesjo:
Thank you!

> @johannesjo ended up removing the time tracker play/pause button. Focus mode and time tracking is very tightly coupled at the moment as it is. It's possible to start time tracking without a focus session, but impossible to do a focus session without time tracking. There is a setting whether break is tracked or not, but any resuming of time tracking also resumes the focus session (skipping break). Any pausing of time tracking also pauses the focus session. It's sort of works intuitively, but also not really explained. Might be a separate PR for later.

Yes, this was a deliberate simplification, as it turns out that it is what the big majority of the users of this users is expecting anyway. You're right that this could be better explained.

johannesjo:
I am merging this. If something comes up we can tackle it in a follow-up issue.

Thank you very much for your work on this @miklosp ! <3

PR #7595, opened 2026-05-13
refactor(sync): post-extraction review cleanup of @sp/sync-core and @sp/sync-providers

johannesjo:
## Summary

Twelve focused improvements identified by a post-extraction review of `@sp/sync-core` and `@sp/sync-providers`. Each item is a separate commit so review can target individual concerns.

- **Surface pruning**: drop 47 unused sync-core barrel exports; remove the unused encryption-migration path (`decryptWithMigration`, `DecryptResult`, structural re-encrypt helpers); drop the redundant `OperationStorePort`; remove the unused root `@sp/sync-providers` barrel and its `"."` export.
- **Decoupling**: strip NgRx-shaped fields from `EntityConfig`/`EntityRegistry` (host now extends via generic `THostExtensions`); move `StateSelector`/`PropsStateSelector`/`SelectByIdFactory`/`EntityAdapterLike` to a new app-side `entity-registry-host.types.ts`; deprecate SP-specific `OpType.SyncImport`/`BackupImport`/`Repair` via JSDoc (303 callsites — staged removal); make SuperSync `defaultBaseUrl` a required dep (no more implicit SP-hosted URL fallback inside the package); add a `SuperSyncWebSocketAccess` interface + structural guard so consumers no longer need `instanceof SuperSyncProvider`.
- **Hygiene**: rename confusing `provider.types.ts` / `provider-types.ts` dual filename; replace wildcard `@sp/sync-providers/*` tsconfig path alias with 11 explicit subpath entries matching `package.json#exports`; move `@sp/sync-core` to `peerDependencies`; add `composite: true` and `tsconfig.build.json` overlays.
- **Latent bug fix**: Dropbox token-refresh was hand-rolled at 5 sites with recursive retries that would not terminate if the refresh token itself was revoked. Consolidated into a single `_withTokenRefresh` helper bounded to one retry; the outer 401 classifier now correctly surfaces `AuthFailSPError` instead of looping. Also extracts the shared md5 content-rev helper.
- **Refactors for cohesion**: split `super-sync.ts` (1017→918L; extracted request-ID hashing into `request-id.ts`; HTTP plumbing scope-deferred at 12 instance deps); split `conflict-resolution.ts` (577L) into `entity-frontier.ts` plus payload helpers moved to `operation.types.ts`.
- **Tests**: centralize duplicated test helpers (`tests/helpers/sync-logger.ts`, `tests/helpers/credential-store.ts`) and prefer barrel imports where symbols are already public; fill gaps with 12 new sync-core specs (vector-clock pruning comparison-flip cases, sync-import-filter edges) and 22 new sync-providers specs (`error-meta.ts` privacy boundary — `urlPathOnly` strips query/fragment/userinfo, `errorMeta` does not leak headers/bodies/tokens).

**Final test counts**: sync-core 195 → 207; sync-providers 319 → 341.

**Behavioral preservation**: encryption wire format unchanged; existing ciphertext continues to decrypt. The one intentional behavioral change is the bounded Dropbox token retry (was unbounded recursion).

## Test plan

- [ ] `npm run sync-core:build` and `npm run sync-providers:build` succeed
- [ ] `npx tsc --noEmit -p packages/sync-core/tsconfig.json`, `packages/sync-providers/tsconfig.json`, `src/tsconfig.app.json`, `electron/tsconfig.electron.json` — all clean
- [ ] `cd packages/sync-core && npx vitest run` — 207 pass
- [ ] `cd packages/sync-providers && npx vitest run` — 341 pass
- [ ] App-side sync flows still work (manual smoke: enable sync, save a task, verify it round-trips)
- [ ] Dropbox provider: connect, sync, force-revoke refresh token in Dropbox UI, attempt sync — expect single `AuthFailSPError` rather than hang/loop
- [ ] SuperSync provider: connect, sync, restore — `instanceof` removed but structural guard works in `sync-wrapper.service.ts`


johannesjo:
Review feedback from a multi-agent pass on current PR head `f0706ac7`.

## Findings

1. **[P2] Package specs are not typechecked, and two current specs fail strict TS.**

   `vitest run` passes because it transpiles the specs, while the package `tsconfig`s do not include them (`packages/sync-core/tsconfig.json:20`, `packages/sync-providers/tsconfig.json:20-21`). I confirmed direct strict typechecks fail:

   - `packages/sync-core/tests/conflict-resolution.spec.ts:556` infers `processRemoteWinnerOps` as returning ops with required `entityId`, but the `EntityConflict<Operation>` input still allows `entityId?: string`.
   - `packages/sync-providers/tests/file-based/dropbox/dropbox-api.spec.ts:761` returns a union where some `headers` objects infer `dropbox-api-result?: undefined`, which is not assignable to `Record<string, string>` / `NativeHttpResponse`.

   This means the new/changed specs can drift from the real types without package CI noticing. Suggested fix: either add a package test-tsconfig/typecheck script that includes `tests/**/*`, or fix these two specs and wire that typecheck into the package test/check path.

2. **[P3 / API follow-up] `errorMeta` and `urlPathOnly` lost any supported public import path.**

   The PR intentionally removes the root `@sp/sync-providers` export, but `errorMeta`/`urlPathOnly` were previously exported there. They are still treated as an important privacy boundary in `packages/sync-providers/tests/log/error-meta.spec.ts:2`, yet the new `package.json#exports` list has no `/log` or equivalent subpath (`packages/sync-providers/package.json:6-117`).

   If these helpers are meant to be public provider utilities, add a focused subpath barrel/export for them. If they are intentionally internal now, I would call that out explicitly with the root-barrel breaking change so downstream consumers do not have to discover it via `ERR_PACKAGE_PATH_NOT_EXPORTED`.

## Notes

I am not treating the broader `@sp/sync-core` and root `@sp/sync-providers` export pruning as a blocker because the PR description says that surface reduction is intentional. It is still a breaking API change for any external consumers of the package root or the removed sync-core barrel symbols, so it should stay explicit in the release/merge notes if these packages are consumed outside this repo.

Checked and passed locally:

- `npm run sync-core:build`
- `npm run sync-providers:build`
- `npm run sync-core:test` (207 tests)
- `npm run sync-providers:test` (341 tests)

Targeted typecheck commands that failed as described above:

- `npx tsc --noEmit --strict --module preserve --moduleResolution bundler --target ES2022 --lib ES2022,DOM --types vitest packages/sync-core/tests/conflict-resolution.spec.ts`
- `npx tsc --noEmit --strict --module preserve --moduleResolution bundler --target ES2022 --lib ES2022,DOM --types vitest packages/sync-providers/tests/file-based/dropbox/dropbox-api.spec.ts`

PR #7600, opened 2026-05-14
fix(tasks): persist collapsed section state across project switches

### LINKED ISSUE #7462: 💡 [Sections] Remember minimized state
### Problem Statement

The following steps:
1. Minimize a section
2. Switch to a different project
3. Come back to the project with the section from point 1 

Result: that section will be reset to unminimized state.

https://github.com/user-attachments/assets/4c6cd3f1-6bed-45a2-9a56-7c12fa3ead29

### Possible Solution

I feel like If I minimize a section - I want it to stay minimized when I come back, for example, if it's some quick edit in other project, if I want to drag a task from a different project or if it's just some long section with a lot of tasks for way later.

Not sure if others would feel the same. Or maybe it should be a preference setting "remember minimized sections state: yes/no", but I want to voice this request. What do you think, guys?

Same feedback can be applied to "Group by Tags / projects" feature

Hetsavani:
## Problem

Collapsed sections lose their state when switching between projects/work contexts.
This causes previously minimized sections to expand again after navigating away and returning.

fixes: #7462

## Solution

Persist the collapsed section state instead of keeping it only in temporary UI state.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/section/section.model.ts
+++ b/src/app/features/section/section.model.ts
@@ -6,6 +6,7 @@ export interface Section {
   contextId: string;
   contextType: WorkContextType;
   title: string;
+  isExpanded?: boolean;
   taskIds: string[];
 }
 
--- a/src/app/features/section/section.service.ts
+++ b/src/app/features/section/section.service.ts
@@ -50,6 +50,7 @@ export class SectionService {
           contextId,
           contextType,
           title: sanitizeSectionTitle(title),
+          isExpanded: true,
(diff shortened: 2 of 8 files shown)


johannesjo:
Thank you!

PR #7610, opened 2026-05-14
chore: combine and rename discussion pruning steps

zenoprax:
Re-ordered the steps to avoid a discussion that is marked as answered
just before the expiry to be false marked as OUTDATED.
Simplified messages.
Shortened waiting period for answered discussions.


johannesjo:
Thank you!

PR #7628, opened 2026-05-16
Allows to use self-hosted Jira instance without web extension

baflo:
## Problem

In a self-hosted Jira instance you can configure allowed origins and don't need the web extension.

## Solution

Allow to bypass the web extension per Jira issue provider configuration.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for the PR. I found a few issues that look worth fixing before merge:

1. `src/app/features/issue/providers/jira/jira-api.service.ts:538-556` - the new direct browser `fetch()` path never checks `response.ok` and discards the HTTP status before errors are wrapped. That changes behavior compared with the Electron/extension path: 401/403 will not trigger the existing block-access behavior, and 404s from `search$()` can no longer hit the Jira Server/DC fallback from `/search/jql` to `/search` because the downstream `catchError` no longer receives `status` / `error.statusCode`.

2. `src/app/features/issue/providers/jira/jira-api.service.ts:538-540` - successful empty responses are treated as failures. Jira mutation endpoints such as transition/assignee/worklog can return `204 No Content`; in that case `response.body` can be `null`, so `streamToJsonIfPossible(body as ReadableStream)` throws even though Jira already applied the change. The direct fetch path should handle an empty body as a successful empty response before applying `transform`.

3. `src/app/features/issue/providers/jira/jira-common-interfaces.service.ts:47` - `altPublicLinkHost` is concatenated directly with `"/browse/"`. If the configured public host has a trailing slash, especially with the documented path-prefix use case, generated links become `https://host/path//browse/KEY`. Normalizing with `stripTrailing(..., "/")` before appending would avoid that.

I ran the targeted specs locally:
- `npm run test:file src/app/features/issue/providers/jira/jira-api.service.spec.ts` - passing, 15/15
- `npm run test:file src/app/features/issue/providers/jira/jira-common-interfaces.service.spec.ts` - passing, 6/6

baflo:
Hi @johannesjo, I see what you mean and already have a prototype that seems to work, but I'll give it a few days of testing.

johannesjo:
Thank you very much @baflo !

----------

Additional review from a few focused passes:

1. `src/app/features/issue/providers/jira/jira-api.service.ts:536-558` - the direct-fetch branch returns before the shared request-log path, so it also skips the app-level Jira timeout from `_makeJiraRequestLogItem()`. A stalled self-hosted Jira request can now hang indefinitely instead of failing after `JIRA_REQUEST_TIMEOUT_DURATION`.

2. `src/app/features/issue/providers/jira/jira-api.service.ts:529-538` - the docs say the setting sends requests directly from the browser, but the code still prefers the extension whenever `_isExtension` is true. That makes behavior depend on timing: early requests may use direct fetch, later requests use the extension. If this setting is intended to bypass the extension in the web app, direct fetch should be selected before the extension path. If it is only meant as a fallback when no extension is present, the docs/label/tests should say that explicitly.

3. `src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.html:88` - the Jira setup dialog still always shows the web-extension-required warning. With this new setting, that warning is misleading for the self-hosted Jira users this PR targets; it should be gated on the fallback setting or the text should mention the bypass option.

Implementation scope: please simplify this before merge.

This feature should be a minimal addition: one optional direct-fetch executor for web Jira configs that explicitly enable it. If the existing Electron/extension request-log path can stay unchanged, it should stay unchanged.

A smaller shape would be:

- add a narrow `isDirectFetchEnabled(cfg)` helper
- let readiness/minimal-settings use that helper
- route direct fetch through one small helper
- keep the existing Electron/extension request plumbing intact
- have the direct-fetch helper preserve status, handle empty bodies, and apply timeout/error behavior

The current `registerRequest()` refactor changes existing request plumbing while adding the fallback, which makes the PR harder to review and creates avoidable regression surface. Since this can be implemented more narrowly, it should be.

baflo:
Here's my first proposal for all your points. I didn't test it yet, though, and will give solid tests the next days.

johannesjo:
Thanks for the updates @baflo — all six correctness points from the earlier rounds look genuinely addressed, and the added specs are good. Two things before merge:

**1. Scope: please reshape away the `registerRequest()` refactor (still open from my last review)**

The current `_sendRequestToExecutor$` restructures the shared Electron/extension promise plumbing (`let promise\!`, the `registerRequest()` "called twice" guard, per-branch registration). I understand *why*: the old code built the `_requestsLog` item + timeout up-front for all non-Android paths, and the new direct-fetch path resolves via the observable rather than `_handleResponse`, so an up-front log entry would leak (stray entry + never-cleared timeout).

But that leak is avoided by *excluding the fetch path from up-front registration* — not by restructuring the promise plumbing. A minimal shape:

```ts
const useDirectFetch = IS_ANDROID_WEB_VIEW || cfg.allowFetchFallback;

if (\!useDirectFetch) {
  this._requestsLog[requestId] = this._makeJiraRequestLogItem({ ... });
}

if (IS_ELECTRON) { window.ea.makeJiraRequest(...); }
else if (useDirectFetch) { return this._directFetch$(...); }
else if (this._isExtension) { this._chromeExtensionInterfaceService.dispatchEvent(...); }
else { throw new Error('Jira: No valid interface found'); }

this._globalProgressBarService.countUp(url);
return from(promise).pipe(...);
```

One boolean, one extracted `_directFetch$` helper, the existing `\!IS_ANDROID_WEB_VIEW` guard widened to `\!useDirectFetch`. The old `promiseResolve`/`promiseReject`/`promise` block stays byte-identical, so the Electron/extension path is provably unchanged (a `git diff` showing zero edits to those lines is the proof). This also lets the `let promise\!` + `registerRequest()` guard disappear entirely.

**2. Blocking bug: timeout globally blocks access on the direct-fetch polling path**

`_scheduleRequestTimeout` unconditionally calls `_blockAccess()` and opens the `TIMED_OUT` error snack, ignoring `suppressErrorSnack`. Background polling uses the new direct-fetch path with `suppressErrorSnack: true`, so a single slow self-hosted request now shows an error snack *and* trips the global Jira access block (banner + manual unblock) — for a silent background poll. The fetch `catchError` honors `suppressErrorSnack`; the timeout path bypasses it. Please thread `suppressErrorSnack` into `_scheduleRequestTimeout`, and reconsider whether one poll timeout should trip the *global* block on the direct-fetch path at all.

**Minor**

- `ALLOW_FETCH_FALLBACK` label ("…as fallback, when no other interface works") now contradicts behavior — direct fetch takes priority over an active extension when enabled (per the test "calls fetch() even when extension is active"). Suggest rephrasing to something like "Fetch Jira directly from the browser (bypass extension)".
- `_parseFetchResponse` returns `null` for 204; downstream calls `transform({ response: null }, cfg)` — worth confirming the transition/assignee/worklog transforms tolerate that.
- `issue-provider-tab.component.ts`: prefer narrowing on `issueProviderKey === 'JIRA'` over the `as unknown as Partial<JiraCfg>` double-cast.
- Thrown message grammar: "must only called once" → "must only be called once".

Also flagging you mentioned the latest push isn't tested yet — happy to re-review once you've run it against a real self-hosted instance (slow / 401 / 404 / 204 cases).

baflo:
Thanks for the reviews; I addressed a few more things based on the latest feedback:

**Timeout no longer trips _blockAccess()**
  The BLOCK_ACCESS_MSG banner says "to prevent API lock out", which only applies to repeated auth failures (401/403) (not to slow responses). Timeouts on path (Electron, extension, direct fetch) no longer call _blockAccess(). The functions `_handleResponse` and `_parseFetchResponse` handle the actual lockout risk via `isUnauthorizedError`.

 This also fixes a real regression I was hitting daily: a slow self-hosted Jira response was tripping the global block even though credentials were fine.

**suppressErrorSnack threaded into _scheduleRequestTimeout**
Background polls with `suppressErrorSnack: true` no longer show the timeout snack.

**Snack suppressed when block banner is shown**
When `_blockAccess()` is triggered (401/403), it now returns `{ jiraBlocked: true }`, which is spread onto the thrown/rejected error. Both `catchError` blocks check this flag and skip the error snack, the block banner should be the better UX there and the snack  would be redundant.

A comment on _blockAccess() documents the intent for future readers.

EDIT: real-life (throttled) testing is ongoing.

johannesjo:
Thank you very much! <3

PR #7634, opened 2026-05-16
feat(schedule): add Shift/Ctrl+Arrow minute stepping to time inputs

### LINKED ISSUE #7216: 💡 Set times more efficiently when navigating via keyboard
### Problem Statement

When scheduling existing tasks via the keyboard, setting times that are not on the full hour is fiddly and takes more time than it needs to.

### Possible Solution

When setting times in schedules, offer a way to set minutes more effectively.

I suggest two options:
1. allow users to set minute interval on arrow key in their settings
2. define one or multiple modifier keys to allow setting minutes in larger intervals

**re. option 1**
this could sit in the schedule section

**re. option 2**
Shift+arrow to skip 5 minutes at a time
CTRL/CMD+arrow to skip 15 minutes at a time
for additional customisation you could let users set these in the schedule settings as well

Alternatively there could be a plugin, but this seems like a core feature rather than a plugin, as it doesn't add functionality but rather modifies core functionality

### Additional context

_No response_

### documentation gap

- [x] ... I can help improve the documentation.

agncr:
Add keyboard stepping to time inputs in the schedule and deadline dialogs.

Shift+Arrow Up/Down steps by 5 min, Ctrl/Cmd+Arrow Up/Down by 15 min — hardcoded to match calendar app conventions. No config needed.

Resolves #7216


johannesjo:
Thanks for the rework. I think there is one behavioral issue to fix:

In `src/app/ui/time-step/time-step.directive.ts:48`, the modifier step can be applied as hours when the browser native time input changes the hour segment. The requested behavior is minute stepping (`Shift+Arrow` = 5 min, `Ctrl/Cmd+Arrow` = 15 min), but with `09:00` and the hour segment active, `Shift+ArrowUp` can become `14:00` instead of `09:05`; `Ctrl/Cmd+ArrowUp` can jump by 15 hours.

Could we handle the modified arrow keys directly with `preventDefault()` and always apply one minute-based path, e.g. `stepTimeString(oldValue, +/-5|15)`? A directive-level test covering hour-segment focus would be useful too.

agncr:
The preventDefault() causes a regression: subsequent Shift+Arrow presses only step by 1 instead of 5, because the browser's internal segment state doesn't update when we swallow the event.

Reverting to the two-phase approach (capture old value in keydown, override in onInput), but using stepTimeString() instead of the hours/minutes detection — so it's always minute-based regardless of which segment is focused. Does that work for you?

johannesjo:
Thanks for the rework, @agncr — the new structure is much cleaner. Splitting out the pure `stepTimeString()` util with its own thorough spec is the right call, and the directive approach **correctly resolves my earlier hour-segment concern**: because the directive always `preventDefault()`s on a modifier+arrow and `stepTimeString()` is unconditionally minute-based, the native hour-segment increment can never fire (`09:00` + `Shift+ArrowUp` → `09:05` regardless of which segment is focused). Re-reading `.value` on every keydown also means consecutive presses compound correctly — the "subsequent presses step by 1" regression you were worried about does **not** occur with this design, so this is actually better than the two-phase approach you described in the thread.

I had this reviewed in depth (plus an independent second pass). One blocker before this can merge:

### 🔴 Blocking — the feature isn't wired into the UI

The net diff vs `master` is only the 4 new files plus some unrelated `t.const.ts`/`en.json` churn. `grep` finds **zero** usages of `spTimeStep` / `TimeStepDirective` / `stepTimeString` outside the new files, and neither of the two native time inputs —

- `src/app/features/tasks/dialog-deadline/dialog-deadline.component.html`
- `src/app/features/planner/dialog-schedule-task/dialog-schedule-task.component.html`

— has `spTimeStep` applied, nor is the directive imported into either standalone component. So none of this code executes in the running app and #7216 is not actually resolved yet. To finish the wiring:

```html
<input type="time" spTimeStep [(ngModel)]="selectedTime" step="60" matInput ... />
```

plus add `TimeStepDirective` to each component's `imports: [...]`. Note the deadline input already has `(keydown)="onTimeKeyDown($event)"`; that handler only acts on `Enter` and the directive only on modifier+arrows, so they're disjoint — but please confirm there's no double-handling once wired (the directive doesn't `stopPropagation`).

### 🟠 Important

- **No integration test for the real path.** `time-step.directive.spec.ts` uses a bare `<input type="time" spTimeStep>` with no `ngModel`/`matInput`. The directive drives the form by setting `.value` + dispatching a synthetic `input` event — but nothing proves `selectedTime` (the ngModel-bound field) actually updates through Material's value accessor. Given the wiring gap above, please add at least one test (unit with `ngModel`, or an E2E in `e2e/`) proving the bound value changes in a real dialog.
- **Guard `readonly`/`disabled`.** `onKeyDown` mutates `.value` and dispatches `input` even on a readonly input. Add `if (el.disabled || el.readOnly) return;` — cheap correctness for a reusable `ui/` directive.

### 🟡 Minor

- **Scope creep in `t.const.ts` / `en.json`.** The PR reorders the `D_SCHEDULE_TASK` keys and changes `SHOW_HIDDEN_DONE_SUB_TASKS` from `"+ {{nr}} completed subtasks"` → `"+ {{nr}} done subtasks"`. Neither relates to minute-stepping — please drop the wording change and revert the key reorders to keep the diff scoped.
- **The two "hour segment" directive specs are identical.** *"…minute segment focused"* and *"…even when hour segment is active"* set up the exact same state and dispatch the same event — the second never actually positions the caret on the hour segment (not feasible in jsdom), so it gives false confidence about the very concern it names. Either make it a genuine distinct case (realistically an E2E) or drop the misleading one.
- **Test coverage nits:** `preventDefault` is only asserted on the `shiftKey` path — add the `ctrlKey`/`metaKey` cases; and consider a held-key (`ev.repeat`) case since hold-to-step is the common stepper UX.
- **`stepTimeString` contract is looser than its spec implies.** `parseInt` is lenient (`"09abc:30"` → `"09:35"` rather than `null`) and `HH:MM:SS` input returns `null` (silent no-op). Inert for the actual call sites (`step="60"` → canonical `HH:MM`), but since the spec advertises validation, either tighten with a `/^\d{1,2}:\d{1,2}$/` check or add a code comment documenting the `HH:MM`-only precondition.

### Summary

The algorithm and directive logic are sound and my original feedback is addressed — this just needs the integration to actually exist:

- [ ] Apply `spTimeStep` + import the directive in both dialog components
- [ ] Add a test proving `ngModel` updates through the directive in a real dialog
- [ ] Guard `readOnly`/`disabled`
- [ ] Drop the unrelated `en.json`/`t.const.ts` churn
- [ ] Fix the duplicate/misleading "hour segment" test

CI is green and the build packages fine, so once it's wired up this should be close.

johannesjo:
Thank you very much! <3

Issue #7606, opened 2026-05-14
💡 Project sheet


tratatas:
### Problem Statement

Currently, projects are simply a way to group tasks; they differ little from a regular tag and do not offer any additional functionality.

### Possible Solution

I propose expanding the tasks and adding a separate page when selecting a project, which would include:
1. General project description and notes
1.1. Project-specific tags
2. Project start date (when the first task appears in the list or the project card is created)
3. Project completion date (closing the last task or moving the project to the archive)
4. The ability to move projects to an archive where they will remain accessible for reporting, reviewing annual results, etc., but will be in a separate interface or section, hidden by default
5. Total number of tasks
6. Maybe a Gantt chart

This card data will make it easier to navigate statistics for a single client or for similar/related projects

### Describe alternatives you've considered

1. Notes in the current implementation are better suited for reminders or brief comments. They are very difficult to use for fully formatted lists, texts, or recording findings based on completed tasks, etc.
2. The current version of the worklog is triggered by the start and completion of individual tasks. In my line of work, tracking the time spent on each task is not as important as monitoring their statuses, start dates, and deadlines

### Additional context

AI translated

vonalbert:
For the 4th point I tried to continue the work of PR https://github.com/super-productivity/super-productivity/pull/7162 that appears abandoned by:
- creating an archived projects list (read only with just an unarchive action)
- making sure open tasks belonging to an archived project don't appear anymore in the app

There are still a few points where work needs to be done, for example recurring tasks still get a new instance and the upcoming page still shows tasks and repeat config from archived projects.

For reference this was the branch I was working on: https://github.com/vonalbert/super-productivity/tree/feat/archive-project

johannesjo:
That's great to hear @vonalbert ! Looking forward to see what you come up with!

tratatas:
The focus of the project card/sheet may not fully capture the underlying concept.

Tasks that can be addressed using dedicated cards, to which you can link tasks or other similar cards:
1. A project card will allow you to:
- search only within project cards, rather than across all tasks associated with the project;
- link multiple projects using recurring tags for work and analytics specific to those projects;
- provide a space for a general project description and data that needs to be considered throughout the project;  
2. A client card would allow you to: 
- search for projects associated with specific clients to analyze open/completed/failed projects, how much time they took, and ultimately how much revenue they generated, etc.
- have a space for client-specific data.

[...]

PR #7162, opened 2026-04-07
Feat/archive project

symonbaikov:
Problem
---

Closes #7126

  When a project is complete, users want to hide it from the sidebar without permanently deleting it and its tasks. The existing "hide from menu" (isHiddenFromMenu) option keeps the project
  visible in the visibility pop-up menu, which becomes unwieldy with many projects.

  Solution

  Expose the already-existing isArchived field on the Project model through the UI:

  - Added "Archive project" button to the project context menu (right-click or ⋮). Archiving navigates away if the archived project is currently active.
  - Archived projects disappear from both the sidebar and the visibility menu entirely.
  - The visibility menu (eye icon in the Projects header) now shows archived projects in a separate section below a divider with an unarchive icon — clicking restores the project instantly.

  No data migration needed — the backend (NgRx actions, reducer, selectors, service methods) was already in place.
  
  ## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki).
- [x] I have run `npm run checkFile` on changed `.ts`/`.scss` files
- [x] I have added tests for my changes (if applicable)
- [x] Existing tests still pass
- [x] My commit messages follow the Angular format (`type(scope): description`)
Screenshot From 2026-04-07 10-20-59

Issue #7126, opened 2026-04-04
💡 Project Archiving


zenoprax:
### Problem Statement

I have many projects. Once the project is "complete" I will hide it as deleting it will remove all the tasks that were contained within it. The sidebar isn't able to handle more than a couple dozen projects. 

<img width="434" height="963" alt="Image" src="https://github.com/user-attachments/assets/d055e5e6-56ca-4abc-80dc-cca3dcf25e52" />

### Possible Solution

"Archive" a project which removes it from the sidebar completely while retaining the project itself and the tasks within it for "Unarchiving" and searching as well as any logs it's a part of.

### Describe alternatives you've considered

1. Hiding: the UI to hide/un-hide is only useful for controlling visiblity in the sidebar; it is not suitable for long term hiding.

2. Deletion: this removes all tasks within the project permanently

3. Moving them to dedicated "Archive" folder: this is probably the only useful solution right now but it still fills up the hide/unhide menu

### Additional context

I tested and once I add more projects a scroll bar appears in the hide/unhide menu.

### documentation gap

- [ ] ... related docs should be updated as part of the same PR whenever possible.
- [ ] ... I can help improve the documentation.

BurningDev:
+1 I think it is a very helpful feature

symonbaikov:
Working on it.

zenoprax:
Feature added! Works great :)

Issue #450, opened 2020-07-22
Improve project finish! (Archive projects)


cknoll:
### Your Environment

* Version used: 5.4.2
* Operating System and version: debian 9 (installed via snap)
* Desktop Environment: KDE

### Expected Behavior
Having a button to archive or finish a project
### Current Behavior
only finding a possibility to "delete" a project, see #442 

Considering the nice and fancy animation when clicking "finish day" it seems kind of strange when I finished a whole project to just delete the project. However I do not want to keep old projects in the project list. Am I missing something?

Regardless of this subjectively perceived ui-strangeness: very good product! I am using it for one week now and it already helped me a lot!

johannesjo:
Hey! Thanks for opening this up. Personally I don't finish projects that often so I haven't really thought about it that much. Do you have any suggestions for how this could be improved? So far I am interpreting your suggestions as:

1. Add nice project finish animation and a project summary page
2. Make it more accessible (maybe by adding an entry to the project context menu?)

cknoll:
Both suggestions seem very reasonable. I have nothing to add.

papernoise:
Project archiving would be really great! I think you could just replace the "finish day" button on a project and turn it into a "finish project" or "archive project" one. Personally it wouldn't bother me if the animation was the same as when I finish my day and the project summary is aready there, isn't it? I mean when I go to the worklog or metrics page these are basically project summary pages.
It would also be very useful if the projects in the sidebar would then be moved to a different location. My sidebar is getting crowded right now!

> Make it more accessible (maybe by adding an entry to the project context menu?)

That would definitely be useful. Maybe just "archive project" below or above settings?

btw. Super Productivity really made more productive! Great work!

robaganrab:
"Archive" feels to be like the project will always be there somewhere.
"Finish" or "close" is more towards "the work here is done", and allows the user to stop spending mental effort on it.

papernoise:
Having spent some time with the application, I think I have some new ideas on the matter.
Actually, archiving/finishing projects might not even be necessary. I think we have been pushed into thinking that apps need to do everything, and then more. Often duplicationg functionality that is already present on the most basic level in the operating system, or in other applications one has.

The way I handle archiving right now is that I simply export the CSV from the project, save it together with other project documents, delete the project from SP and then put everything into a folder on my computer.
This way it's out of my mind, but I can get back to the data if I need it.
[...]

PR #7639, opened 2026-05-16
Archive project - Continuation of #7162

vonalbert:
## Problem

When a user finishes working on a project has two options:
- delete the project losing all the progress done, ncluding worklogs
- keep the project hidden in a folder cluttering the the project menus

Linked issues:
- #7126
- #450
- #7606

## Solution

Restore the old archive feature! Hidden - but not removed - in the backend.
This PR continues the work started in #7162.
It is based on Symon's latest commit and addresses the review feedback raised during the previous discussion, namely:

**Archived Projects Page**: One concern raised in the previous PR was that the project visibility menu became too long.
This PR moves archived projects to a dedicated page, intentionally kept simple to reduce the scope of the implementation.
From this page, users can search archived projects by name and unarchive them.
The page is linked from the bottom of the project visibility menu, where archived projects were previously listed in #7162.

**Undo action in snack**: Added an undo action to the archive confirmation snack. I chose to keep the snack implementation for simplicity, though I see no issue in switching to a confirmation modal if preferred.

**isHiddenFromMenu + isArchived**: Projects hidden from the menu display a different message when unarchived.

**Reminder of archived project's tasks**: This has been handled: archived projects no longer trigger reminders.

**Missing unarchive snack notification**: Added the missing unarchive snack notification.

The review comment mentioning archived projects being "dormant" made me reconsider the expected behavior of archived projects overall.
If archiving is intended to fully suspend a project while preserving its data, then active tasks belonging to archived projects should consistently be excluded from all operational views and automated flows (i.e. a user may want to abort a project while still preserving its historical data).

Views, components and services covered:

- Work-context component
- Scheduled list
- Overdue tasks
- Deadline and due-day lists
- Repeat task instance generation

Most of the implementation was done at the NgRx selector level. The only exception is the work-context component, where the filtering logic was implemented in `work-context.service.ts` in order to preserve the ability to view tasks in archived projects (i.e. I can point the browser directly to the archived project's page and view its tasks).

### Open Questions

I'm not entirely sure whether implementing this at the selector level was the best approach, so feedback on the architecture would be appreciated.

I'm also unsure about the selector naming strategy. Is the current `selectAllTasks*` pattern appropriate, or would more explicit naming such as `selectActiveTasks*` and `selectArchivedTasks*` better communicate the intended use cases?

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


vonalbert:
After sleeping on it, I found a couple more spots where tasks from archived projects are displayed:
- the two planner views
- the later today section in today view

I can fix those as well, but first I'd like to validate the approach taken in this PR (basically the "open questions" section).

johannesjo (commented):
Thanks for picking this up — solid continuation of #7162. The model choice is right and **sync-correctness is sound**: this reuses the pre-existing op-log-persistent `archiveProject`/`unarchiveProject` actions, adds no effects/reducers/new actions, and is purely read-side + UI. Reminder suppression genuinely works because `reminder.service.ts` consumes `selectAllTasksWithReminder` reactively. No security/privacy issues (titles are interpolation-escaped, `[style.color]` is sanitized, nothing user-generated is logged). The menu/reducer/repeat-cfg tests are thorough.

Answering your two open questions directly:

### 1. Is selector-level the right approach? → Right *layer*, wrong *granularity*

Selectors are the correct layer (not a meta-reducer — that rule is for multi-entity *writes*; this is read-only derivation). But threading `selectArchivedProjectIds` into ~8 selectors plus a third copy in `WorkContextService` is too fine-grained, and the PR already demonstrates why: you found planner + later-today *after* opening it, and there's a **third uncovered path** — `selectTaskRepeatCfgsForExactDay` (`task-repeat-cfg.selectors.ts:79`) is unfiltered and feeds the Planner (`planner.selectors.ts:290`) and Schedule (`create-schedule-days.ts:100`, `create-sorted-blocker-blocks.ts:77`). So archived projects' repeating tasks still appear as projected occurrences there, contradicting the new wiki copy ("hidden from … Scheduled views", "fully dormant").

Good news: actual instance **creation** goes through `selectAllUnprocessedTaskRepeatCfgs`, which *is* filtered — so "no new instances are generated" is accurate; the leak is display-only projection.

**Recommendation: one shared base selector:**

```ts
export const selectAllTasksInActiveProjects = createSelector(
  selectAllTasks,
  selectArchivedProjectIds,
  (tasks, archivedIds) =>
    archivedIds.size === 0 ? tasks : tasks.filter((t) => !archivedIds.has(t.projectId)),
);
```

The 6 task selectors then switch their input from `selectAllTasks` → `selectAllTasksInActiveProjects` and **drop the extra projector arg + predicate** (same pattern for a `selectActiveTaskRepeatCfgs` base). This is a *smaller* diff than the current one, removes the ~40 mechanical `NO_ARCHIVED_PROJECTS` test edits, fixes the whack-a-mole by construction (planner/later-today/exact-day covered once they derive from the base), and the `size === 0` fast path makes the common no-archive case a zero-cost identity passthrough.

### 2. Naming → yes, prefer explicit `selectActive*`

Your instinct is right, and it's the same fix as #1: an explicitly-named base (`selectActiveTasks` / `selectAllTasksInActiveProjects`) is self-documenting at every call site, mirrors the existing `selectAllProjects` vs `selectUnarchivedProjects` pair, and turns "did I cover every selector?" into a closed question.

### Blocking

- **`src/assets/i18n/it.json`** — please revert this hunk entirely. Convention here is *en.json only* (other locales are managed externally), and the hunk also introduces a **duplicate `TOGGLE_SHOW_BOOKMARKS` key** in the `MH` block. The en.json additions are correct and sufficient.
- **Docs vs. coverage** — either extend filtering to Planner/Schedule/later-today (the shared base selector does this cleanly) **or** soften the wiki wording so it doesn't promise "fully dormant"/"hidden from Scheduled views" ahead of the behavior.

### Important

- **Two "visible task" definitions now coexist:** `selectAllTasksWithoutHiddenProjects` filters `isHiddenFromMenu || isArchived`, while the other selectors filter `isArchived` only. The project store already models the canonical rule (`selectUnarchivedVisibleProjects` / `selectUnarchivedHiddenProjectIds`) — better to derive from one source than entrench the split.
- **Test gaps:** the `_getTasksByIds$` `combineLatest` change (the most behaviorally significant change, with the non-obvious PROJECT-context exception) has no spec, and the new `archived-projects-page.component.ts` has none. `selectAllTasksWithReminder`, `selectAllTasksWithDeadlineReminder`, and `selectOverdueTasks` lack archived-filter specs.
- **Perf (minor):** `selectArchivedProjectIds` builds a new `Set` on *any* project mutation (e.g. task reordering touches `taskIds`), re-emitting through the 8 selectors and the hot `mainListTasks$` (`shareReplay`) path. CPU is negligible at realistic project counts; it's needless change-detection churn. The `size === 0` fast path above mitigates most of it.

### Suggestions

- Keep the **undo snackbar — don't switch to a confirm dialog**. Archive is reversible; this codebase reserves confirm dialogs for irreversible actions (cf. `deleteProject`). A dialog would be inconsistent and add friction.
- `UNARCHIVED_BUT_HIDDEN` is a UX dead-end: it tells the user the project is still hidden but offers no way to unhide from the page they're on. Consider a snack action ("Show in menu") that clears `isHiddenFromMenu`, mirroring the archive-undo pattern.
- Confirm the intended behavior that archiving a project silently hides imminent deadlines/reminders. Defensible as "dormant" and documented, but worth a conscious product sign-off.
- Archived-projects page is idiomatic; minor polish if you touch it: split the sort out of the per-keystroke `computed`, use `store.selectSignal` + a sorted selector, and pass the `project` object to `unarchive()` instead of re-`find`ing by id.

### Verdict

**Needs changes.** The foundation (sync-safe model, reused persistent actions, pure selectors) is right. The two blockers are quick (revert it.json; reconcile docs-vs-coverage), and adopting the single shared base selector resolves both of your open questions, closes the known gaps, *and* shrinks the diff.

vonalbert:
Sorry for it.json. I'm gonna remove it asap.
For the rest of changes I'll try to address all of them in the few next days.

johannesjo (commented):
@vonalbert no worries! :)


Follow-up, scoped to **UX** (architecture/sync covered in my earlier review — no need to revisit here).

### The open product question this PR doesn't answer: can users actually *reach* archived data?

The PR's premise (and the linked issues #7126/#450/#7606) is "archive instead of delete so the worklog/progress is **kept**." The implementation preserves the data in state — but there is no UI path to it once archived:

- removed from sidebar and the visibility menu,
- `archived-projects-page.component.html` renders the title as a plain `<span>` (not navigable),
- `work-context.service.ts` keeps tasks viewable **only** in PROJECT context — reachable solely by a manually typed `/project/<id>` URL.

The PR description acknowledges this as the intended access model ("*I can point the browser directly to the archived project's page*"). I'd push back on that being sufficient: a user who archived a project specifically to keep its worklog can no longer open that worklog through any affordance. That makes archive behave like a soft-delete you can't read, which undercuts the feature's own justification vs. plain deletion.

This isn't an argument against the stricter dormancy direction (planner/later-today hiding) — those are about *operational* views and are correct. It's specifically about a **read path to the archived project's own history**.

**Suggestion:** make the archived-projects page row navigate to the project (the PROJECT-context bypass already supports rendering it read-side) — or, if routing to an archived project is risky, link the row to its Worklog. Add a clear "you are viewing an archived project" header + a Restore action on that view so it isn't a dead end.

Worth a conscious product call rather than leaving "typed-URL-only" as the de facto answer.

### Smaller UX items (non-blocking, cheap if the page is touched anyway)

- **Terminology drift** — three verbs for two states: "Archive project" / snack "Project archived" + "Undo" / page "Archived projects" / icon tooltip "Restore project" / snack "Unarchived project". Settle on one pair (suggest **Archive ↔ Restore**), en.json only.
- **IA split** — archiving lives on the project ⋮ context menu; restoring lives behind the Projects-header *eye/visibility* icon. Two unrelated controls for one lifecycle. Cheapest fix: a count on the visibility-menu entry, e.g. "Archived projects (3)", so it reads as the restore home.
- **Asymmetric undo** — the archive snack has Undo; the restore snack (`archived-projects-page.component.ts`) has none, though both are reversible. ~3 lines to mirror it (`actionFn: () => archive(projectId)`).
- **a11y** — icon-only unarchive + search-clear buttons need `aria-label` (matTooltip is not a reliable accessible name); the decorative project `mat-icon` should be `aria-hidden`.

(The `UNARCHIVED_BUT_HIDDEN` dead-end and the "keep the undo snackbar, no confirm dialog" calls from my first review still stand — listing these here only to keep the UX picture in one place.)

vonalbert:
Thanks for the UX follow-up.

**Architecture work in progress**

Currently refactoring the selector layer as suggested in your first review: introduced `selectActiveTaskRepeatCfgs` and `selectAllTasksInActiveProjects` base selectors with a size === 0 fast-path, and migrated all downstream selectors to derive from them instead of threading `selectArchivedProjectIds` inline.

One question on the task side: `selectAllTasksWithDueDay` was deliberately optimized in a previous commit to iterate `taskState.ids` directly and avoid the `Object.values()` intermediate array. However, its only consumer is `selectAllUndoneTasksWithDueDay`, which filters the result anyway so the perf gain seems to be effectively cancelled. The cleanest fix would be collapsing both into a single selector deriving from `selectAllTasksInActiveProjects` with a single-pass filter. WDYT?

**On the new UX feedback**

- **Read path to archived data**: the direct URL mention in the PR description was illustrative, not the intended access model. The archived-projects page was intentionally left minimal to keep the PR scope manageable. Agreed that "soft-delete you can't read" defeats the feature's own justification. However I think making the archived project row navigable to the project's task view should be doable. From there the user can access all reporting screens.
- **Terminology**: I agree on settling on one pair. My only gripe is that the term "unarchive" is scattered everywhere in the codebase, so changing it to restore is risky. If you want to proceed with this change I'd keep it as last :)
- **Archived projects count in visibility menu**: will add as you suggest "Archived projects (3)".
- **Asymmetric undo** just to clarify: are you suggesting that the Restore snackbar should expose an “Undo” action that re-archives the project? If so, I don’t think I can support that together with the UNARCHIVED_BUT_HIDDEN case, since, unless I’m mistaken, the snackbar only supports a single action, so I wouldn’t be able to expose both actions at the same time. Would a more explicit snackbar message be sufficient for now? E.g. "Project restored but still hidden from menu. Go to Projects settings to show it". Open to other suggestions if that feels too verbose.
- **a11y**: No problem for that

johannesjo:
Thank you very much!

Follow-up while the selector refactor is in flight — three points, no need to act before it lands.


### 1. The archived-projects page should show at least minimal stats

The feature's whole justification over deletion is "the worklog/progress is kept". Right now the page is name + restore button only — the kept data isn't surfaced anywhere, which makes archive feel like a soft-delete you can't read (the read-path concern from my UX review). It doesn't need the full reporting suite, but the overview row should expose a minimal summary per archived project, e.g. task count / done count / time tracked / last worked. That delivers the feature's own promise on the page the user already lands on, without needing a navigable deep-link.

### 2. Reminder: `selectTaskRepeatCfgsForExactDay` won't be auto-covered by the base-selector refactor

The shared `selectAllTasksInActiveProjects` / `selectActiveTaskRepeatCfgs` refactor covers the `selectAll*` selectors, but `selectTaskRepeatCfgsForExactDay` is consumed via direct `.projector(...)` calls in `planner.selectors.ts:290`, `schedule/create-schedule-days.ts:100`, and `schedule/create-sorted-blocker-blocks.ts:77` — so it bypasses any base selector unless you filter it explicitly too. Without that, archived projects' repeating tasks still appear as projected occurrences in Planner/Schedule, contradicting the new wiki copy ("fully dormant", "hidden from Scheduled views"). Flagging so it doesn't slip through the refactor.

vonalbert:
Ok I think I'm done with the selector refactor part. Point 2 of the last review should be already be covered in the last update.

I see that tests are failing in CI but locally works. It refers to the emptyTaskState variable removed on my branch. I see that commit 8173559c6c1076193a9a4306b77870300caf4731 changed the relevant files. Should I rebase everything on current master, fix the new test on my part and force push?

When the selector part is considered done I think we can focus on ui/ux/a11y refinements.

Thanks for your time reviewing this :)

johannesjo:
Thanks for the detailed update – the selector refactor is in good shape. A few items came out of a second pass that I think are worth folding into the same round before UI polish, because they all touch the same selector layer.

### 1. `selectArchivedProjectIds` reference stability (memoization)

`src/app/features/project/store/project.selectors.ts` returns `new Set(projects.filter(...).map(...))` on every invocation. NgRx uses reference equality on projector inputs, so every downstream selector that takes this as input — `selectAllTasksInActiveProjects`, `selectActiveTaskRepeatCfgs`, `selectActiveWorkContext`, plus the ~25 selectors derived from those — will recompute on every project-state tick even when the archived-id set is identical.

The `size === 0` fast-path you added avoids the filter cost, but the projector still runs and still allocates the filtered array. Two ways out:

- Memoize on equality of the resulting set (cheapest: derive a sorted-id string, return the *same* `Set` instance when unchanged), or
- Skip the `Set` entirely — let the base selector take the project array and do a single `project.isArchived` lookup per task via a Map keyed on `projectId`. One memoized projection, no intermediate `Set` allocation.

Same shape applies to `filterTaskEntitiesByActiveProjects` in `work-context.selectors.ts:32` — `Object.fromEntries(...)` runs on every call when the archived set is non-empty, breaking `selectActiveWorkContext` memoization. Centralizing on the base selector (rather than re-filtering the entities map inline) would remove this entirely.

### 2. `selectAllTasksWithDueDay` collapse + scheduled-list leak

Same as my earlier comment — `selectAllTasksWithDueDay` → `selectAllUndoneTasksWithDueDay` → `scheduled-list-page` chain doesn't filter archived. Collapse into one selector derived from `selectAllTasksInActiveProjects` with a single-pass `\!isDone && dueDay` filter.

### 3. Three more selectors still bypass the active-projects base

While auditing the derivations, these still source from `selectAllTasks` / `selectTaskFeatureState` directly and produce data that feeds operational views:

- `selectTasksDueAndOverdueForDay` (`task.selectors.ts` ~line 579)
- `selectUndoneOverdueDeadlineTasks` (`:178`) and `selectUnplannedDeadlineTasksForToday` (`:204`) — `cc7d70c3` patched the banner case, but the underlying selectors still iterate `taskState.ids` unfiltered. Want to be sure neither is consumed somewhere that should be dormancy-respecting.
- `selectAllTasksWithoutHiddenProjects` (`:737`) — filters `isHiddenFromMenu` but not `isArchived`. Either route it through the base, or document why hidden-from-menu and archived diverge here.

### 4. Race in work-context-menu archive flow

`work-context-menu.component.ts:127` — when archiving the currently-active project, the navigation away appears to fire before the archive action is dispatched. The selectors briefly observe the active project mid-transition. Reorder: dispatch first, then navigate (or chain via the action stream).

### 5. Test wiring — the systemic gap

Most of the new spec assertions call selectors as `selector.projector(preFilteredArray, ...)` rather than `selector(stateWithArchivedProject)`. That tests the *projector function* but skips the *selector wiring*. If a base selector is swapped out — exactly what's happening in this PR — projector-only tests stay green even when the chain is broken.

Concrete adds I'd want:
- A `mockState` fixture with `project1.isArchived = true` and one task in `project1`. Then for each downstream selector (overdue, today, tags, scheduled, planner), assert `selector(mockState)` excludes that task. Single fixture, reused across files.
- Round-trip test in `project.reducer.spec.ts`: archive → unarchive preserves tasks, worklog, `isHiddenFromMenu` semantics.
- `selectActiveTaskRepeatCfgs` chain test: an archived-project repeat cfg does NOT yield an instance via the planner/schedule consumers.

### 6. UX polish (minor, before merge)

- `archived-projects-page.component.html:28` — empty-state needs `role=\"status\"` + `aria-live=\"polite\"` so search-with-no-results is announced.
- Project row navigability (per prior comment) — still the read-path concern.
- Asymmetric undo: previous suggestion stands — state-dependent snackbar action (`Undo` for normal restore, `Show in menu` for `UNARCHIVED_BUT_HIDDEN`), one button per state.

### 7. Rebase question (May-20)

Yes, rebase on master + force-push is the right call. `8173559c` only touched test helpers (`emptyTaskState`); fixing the reference on your side is the cheapest path.

### Clean (no action needed)

Logging hygiene (no user content), sanitization (search uses `.includes`, project titles go through Angular interpolation), op-log compatibility (no new actions, pure read-side), i18n single-locale rule respected, reminder suppression genuinely flows through the new base, no new effects/reducers introduced. The sync-correctness profile of this PR is sound.

---

Items 1 + 5 are the ones I'd treat as merge-blocking. 2–4 are correctness fixes that overlap with work already in flight. 6 is polish.

johannesjo:
Thanks for sticking with this @vonalbert — the archive feature has a wide blast radius (Today, Tags, Boards, Overdue, Deadline, Scheduled, Planner, Reminders, RepeatCfg, calendar polling, op-log capture), which is why we keep finding "one more selector that should filter." Let's try to close it out in one pass.

I ran a multi-agent review of the diff and synthesized the findings below. The headline recommendation is architectural; the rest are concrete asks.

## Architectural recommendation

**Stay with selector-layer filtering** — it's the only approach consistent with our sync invariants. Reducer-time filtering (e.g. stripping IDs from `TODAY_TAG.taskIds`) would violate the virtual-tag rule (`TODAY_TAG.taskIds` is ordering only) and lose unarchive symmetry. Effect fan-out is forbidden by the multi-entity rule. Component-level filtering scatters the logic.

But two things have to change to make the selector layer actually work:

**1. Flip the default name to make the safe version the obvious one.** This is the root cause of "you forgot to migrate selector X" recurring across review cycles:
- `selectAllTasks` → `selectAllTasksIncludingArchived` (raw, opt-in)
- `selectAllTasksInActiveProjects` → `selectAllTasks` (filtered, default)
- Same shape for `selectAllTaskRepeatCfgs` / `selectActiveTaskRepeatCfgs`

After the rename, every consumer that wants raw data has to justify itself at the call site. Reviewers grep `IncludingArchived` and audit ~5 sites instead of grepping `selectAllTasks` and auditing ~30.

To answer your open question #2 directly: yes, the renaming is worth it. `selectActiveTasks*` / `selectArchivedTasks*` is also fine, but "archived" already overloads in this codebase (project archive vs. done-task archive feature), so I'd avoid it.

**2. Stabilize `selectArchivedProjectIds`.** Currently it returns a fresh `Set` on every project mutation (rename, color edit, sort), which invalidates `selectAllTasksInActiveProjects` / `selectActiveTaskRepeatCfgs` / `selectActiveWorkContext` / `selectTodayTaskIds` downstream. The `archivedIds.size === 0` fast path covers the no-archive case, but once any project is archived every unrelated project edit triggers a full task-list re-filter.

Fix by memoizing on content:

\`\`\`ts
export const selectArchivedProjectIdsArr = createSelector(
  selectArchivedProjects,
  (ps) => ps.map((p) => p.id).sort(),
);
export const selectArchivedProjectIds = createSelector(
  selectArchivedProjectIdsArr,
  (ids) => new Set(ids),
);
\`\`\`

NgRx's projector memoization handles the rest; downstream Sets only change when an archived flag actually flips.

## Blocking

1. **\`INBOX_PROJECT\` must never be archivable.** UI gates on it, but backup-import, sync replay, and plugins can bypass that. Add a reducer guard:
   \`\`\`ts
   on(archiveProject, (state, { id }) => {
     if (id === INBOX_PROJECT_ID) return state;
     return projectAdapter.updateOne({ id, changes: { isArchived: true } }, state);
   })
   \`\`\`

2. **Decide and document the selector carve-outs.** These still source from raw \`selectAllTasks\` after the PR:
   - \`selectTasksWorkedOnOrDoneFlat\` — likely intentional (worklog history must survive archive)
   - \`selectAllCalendarIssueTasks\` / \`selectAllCalendarTaskEventIds\` — calendar sync polling; intentional?
   - \`selectMainTasksWithoutTag\` — used by tag picker; should filter
   - \`selectTasksDueAndOverdueForDay\` — overdue widgets; should filter
   - \`selectAllDoneIds\` / \`selectTaskByIssueId\` — unclear

   For each, either migrate to the filtered selector or add a one-line \`// intentionally unfiltered: <reason>\` comment. Once carve-outs are explicit, the next reviewer stops flagging them.

3. **\`archiveProject()\` should dispatch before navigating.** \`work-context-menu.component.ts:127-137\` currently does \`await navigateByUrl('/')\` then dispatch. If navigation rejects (guard, async failure), the archive never fires and the user is stranded. Reorder:
   \`\`\`ts
   this._projectService.archive(this.contextId);
   if (activeId === this.contextId) {
     await this._router.navigateByUrl('/');
   }
   \`\`\`
   The dispatch is synchronous so this also eliminates the render frame where the soon-to-be-archived project is still active.

4. **\`filterTaskEntitiesByActiveProjects\` in \`work-context.selectors.ts:32-44\` is wasteful.** \`Object.fromEntries(Object.entries(entities).filter(...))\` rebuilds the entire entity map on every input change, twice (once in \`selectActiveWorkContext\`, once in \`selectTodayTaskIds\`). Either:
   - Pass \`archivedProjectIds: Set<string>\` into \`computeOrderedTaskIdsFor{Today,Tag}\` and skip inline during iteration, OR
   - Promote to its own memoized \`selectActiveTaskEntities\` selector built once from \`selectAllTasksInActiveProjects\`.

## Warnings (please address)

5. **Subtask filter assumes \`sub.projectId === parent.projectId\`.** Not enforced in the type system. If sync ever leaves them out of sync, you get ghost-parent rows. Mirror the explicit guard used in \`selectActiveTaskRepeatCfgs\`:
   \`\`\`ts
   tasks.filter((t) => \!t.projectId || \!archivedIds.has(t.projectId))
   \`\`\`

6. **\`selectLaterTodayTasksWithSubTasks\` rewrite drops a code path.** Old version iterated \`taskState.ids\` (all entities including bare subtasks). New version only sees subtasks via their parent's \`subTasks[]\`. If a parent is missing from state entirely (corrupt sync), an orphaned subtask used to surface and no longer does. Add a regression spec or restore the fallback.

7. **Snack ownership is inconsistent.** Archive snack lives in \`work-context-menu.component\`; unarchive snack lives in \`archived-projects-page.component\`; the undo-button path calls \`unarchive()\` with no snack at all. Move both into \`ProjectService.archive()\` / \`unarchive()\` for symmetry, and the \`UNARCHIVED_BUT_HIDDEN\` branch lives in one place.

8. **\`selectTimelineTasks\` mixes filtered and unfiltered sources.** It now takes both \`selectTaskFeatureState\` (raw) and \`selectAllTasksInActiveProjects\` (filtered). Pick one — build a \`Map\` from the filtered array if random lookup is needed.

9. **Planner's \`Map<id, Task>\` builds inside the projector body** (\`planner.selectors.ts:103, 151\`). Extract \`selectActiveTaskMap\` selector so day computations share one allocation.

10. **Test coverage gaps.** Two specs were deleted with the rewrite:
    - "should include subtasks with dueDay" (was on the removed \`selectAllTasksWithDueDay\`) — add equivalent for \`selectAllUndoneTasksWithDueDay\`.
    - "should handle missing entity references" on \`selectAllTasksDueToday\` — replace with a test that \`plannerState.days[today] = ['nonexistent']\` returns \`[]\` without throwing.

11. **\`selectAllTasksWithoutHiddenProjects\` now overlaps with \`selectAllTasksInActiveProjects\`** (both filter archived). Compose them: derive a \`selectHiddenProjectIds\` Set and filter \`selectAllTasksInActiveProjects\` through it.

12. **Direct URL navigation to archived projects still works** (\`/project/<archived-id>\`). The PR description says "hidden from sidebar and menu" — please confirm this is intended (preserves data access) so future reviewers don't flag it.

13. **Unarchive has no undo.** Archive has the snack-with-undo; unarchive doesn't. If the user accidentally restores a project that floods Today with old reminders, no easy revert. Mirror the snack pattern.

## Suggestions (nice-to-have)

14. \`pages.routes.ts\` export was inserted at the top; restore alphabetical order with the other page exports.
15. \`archived-projects-page.component.html\` has no \`<h1>\` — screen readers get a list with no announced context. Add \`<h1>{{ T.MH.ARCHIVED_PROJECTS | translate }}</h1>\` inside \`.page-wrapper\`.
16. \`nav-list-tree.component.html:185\` archived-count chip has no combined aria-label; add \`aria-label=\"{{ T.F.PROJECT.ARCHIVED_PROJECTS.LINK_LABEL | translate }} ({{ archivedProjectsCount() }})\"\`.
17. The \`isPlannerAllDayCalendarEvent\` helper and the \`oneDayInMilliseconds\` import in \`planner.selectors.ts\` are unrelated to archive scope (came in via the master rebase). Either split into a separate commit or call it out in the PR description so reviewers don't waste time wondering.
18. Add a one-line comment near \`selectAllTasks\` (renamed) explaining the \`archivedIds.size === 0 ? tasks : tasks.filter(...)\` fast path — it's load-bearing for memoization and easy to "optimize away."
19. Add \`expect(router.navigateByUrl).toHaveBeenCalledBefore(mockProjectService.archive)\` (or after, depending on the reorder above) to \`work-context-menu.component.spec.ts\` so the ordering is pinned by a test.
20. Consider dropping the \`UNARCHIVED_BUT_HIDDEN\` branch entirely — niche case (project both archived AND \`isHiddenFromMenu\`). Single message that mentions the visibility menu would cover both states with less branching and one fewer i18n key.

## What's already good

- Sync-correctness is intact: single-entity \`Update\` op with \`isPersistent: true\`, no new effects, no multi-entity fan-out, \`LOCAL_ACTIONS\`/\`ALL_ACTIONS\` not touched.
- The \`archivedIds.size === 0\` fast path in \`selectAllTasksInActiveProjects\` and \`selectActiveTaskRepeatCfgs\` shows the right NgRx instinct.
- Two-action design (\`archiveProject\` / \`unarchiveProject\` rather than \`setProjectArchived\`) preserves op-log replay symmetry.
- New page is standalone + OnPush + signal-driven with sort-in-selector.
- No security surface — purely local state filtering.
- Wiki doc updated alongside the feature.

To answer your open question #1 directly: **selector-layer filtering was the right call.** Reducer mutation, effect fan-out, and derived slices all conflict with our sync rules. The fixes above (rename + stable Set + explicit carve-outs) make it sustainable.

Happy to clarify any of these or pair on the rename if scope feels too big.

vonalbert:
I think I covered all the points above. Before proceeding, these are the remaining things I’d like to discuss:
- `pages.routes.ts` export order: I’m not sure what ordering convention that file is currently following. Looking at master, it seems mostly grouped by path type (pages, features, plugins), but pages themselves appear in two separate groups (L1-3 and L7-9), so it’s unclear where `ArchivedProjectsPageComponent` should be placed.
- "Direct URL navigation to archived projects still works": I left this on hold because, in the “Read path to archived data” section of [this comment](https://github.com/super-productivity/super-productivity/pull/7639#issuecomment-4482649814), I proposed allowing navigation to the project page by clicking the project row in the archived projects view. Personally, if I archive a project while some tasks are still open, I'd like to see what was left behind.
- I think I implemented the request from point 5 (`tasks.filter((t) => !t.projectId || !archivedIds.has(t.projectId))`) in commit 61564e00cbdc9c8866d47cb8225cff8df10a5f83, but I still don’t understand how that addresses the `sub.projectId === parent.projectId` case. Is that commit actually covering the intended scenario?
- I'm not sure what to do with point 17 to improve the situation. Is it enough to edit the original PR description stating that those file are not modified by me?

johannesjo:
Thank you very much for bringing back archiving projects! I's a great addition! <3

PR #7646, opened 2026-05-17
feat(habit-tracker): grey out non-selected days in specific-days streak

BioWare:
## Problem
In the habit tracker, it was difficult to visually distinguish which days are active for a specific habit when using the "specific-days" streak
mode. This led to confusion and allowed interactions (clicks/context menus) on days that shouldn't be tracked.

## Solution
- **Visual Feedback:** Implemented a "grey out" effect (`opacity: 0.3` and `cursor: default`) for non-active days in the habit tracker grid.
- **Interaction Guard:** Disabled pointer events, clicks, and context menus for disabled days to prevent accidental data entry.
- **Performance Optimization:** Refactored the `days` signal to pre-calculate `Date` objects and Day-of-Week (DOW) indices. This eliminates
redundant `parseDateLocal` calls (7x per counter) during change detection.
- **Logic Fix:** Ensured that days are only disabled when `streakMode` is explicitly set to `specific-days`, preventing incorrect behavior in
`weekly-frequency` mode.

## Type of Change
- [x] New feature
- [x] Refactoring
- [x] Performance improvement

## Checklist
- [x] I have run `npm run checkFile` on changed `.ts` and `.html` files (locally verified).
- [x] My commit messages follow the Angular format (`type(scope): description`).

### DIFF
--- a/src/app/features/simple-counter/habit-tracker/habit-tracker.component.html
+++ b/src/app/features/simple-counter/habit-tracker/habit-tracker.component.html
@@ -72,6 +72,8 @@
         @for (day of days(); track day) {
           <div
             class="day-cell"
+            [class.is-disabled]="!isDayEnabled(counter, day)"
+            [attr.aria-disabled]="!isDayEnabled(counter, day) || null"
             (click)="onCellClick(counter, day)"
             (contextmenu)="onCellContextMenu($event, counter, day)"
             (mousedown)="onPressStart(counter, day)"
--- a/src/app/features/simple-counter/habit-tracker/habit-tracker.component.html
+++ b/src/app/features/simple-counter/habit-tracker/habit-tracker.component.html
@@ -38,10 +38,10 @@
   >
     <div class="header-row">
       <div class="header-spacer"></div>
-      @for (day of days(); track day) {
+      @for (day of days(); track day.str) {
         <div class="day-header">
-          <div class="day-name">{{ parseDateLocal(day) | localeDate: 'EEE' }}</div>
-          <div class="day-date">{{ parseDateLocal(day) | localeDate: 'dd' }}</div>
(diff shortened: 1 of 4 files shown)


johannesjo:
Thank you very much for the PR!

The date/streak logic looks consistent with the existing local-date behavior, and `weekly-frequency` staying enabled looks right.

Two small interaction issues to fix before merge:

1. Disabled-day long-press is not guarded in TS. `click` and `contextmenu` check `isDayEnabled()`, but `onPressStart()` / `onPressEnd()` can still schedule and open the edit dialog without checking the day. Please pass `day.dow` or an enabled flag through the press path and bail for disabled days.

2. Disabled-day context menus are not reliably suppressed. `.is-disabled` uses `pointer-events: none`, so the cell handler may not run; and if it does run, `onCellContextMenu()` returns before `event.preventDefault()`. Please make sure right-click/contextmenu on disabled cells does nothing and does not show the browser menu.

A small focused test for these guards would be useful if practical in the existing test setup.

BioWare:
@johannesjo Hi! Thanks for the review and the advice. I've updated the PR to address the interaction issues for disabled days.

Improvements:
 - Long-press guard: Added isDayEnabled() check to onPressStart to prevent the edit dialog from opening on disabled days.
 - Context menu fix: Moved event.preventDefault() to the top of onCellContextMenu to ensure the browser's default menu is always suppressed on disabled
   cells.
 - CSS update: Removed pointer-events: none from disabled cells. This allows the event handlers to catch and properly suppress interactions as
   requested.
 - Tests added: Created habit-tracker.component.spec.ts with focused tests covering long-press and context menu interactions for both enabled and
   disabled days.

I've verified the changes locally, and all 3 new tests are passing. Let me know if anything else is needed!

johannesjo:
Thank you very much!

PR #7650, opened 2026-05-18
feat(tasks): auto add tasks with deadlines today to Today view

### LINKED ISSUE #7488: 💡 Add tasks with deadlines to Today n days before deadline
### Problem Statement

Users have the option to add unscheduled tasks with deadlines to `Today` on the day of the deadline. However, if they are unscheduled, they don't appear in the Today list at any point before the deadline. When the amount of tasks with deadlines increases, it becomes harder and harder to manually manage and track which tasks are due when, which I think is the point of the Deadline feature.

### Possible Solution

Automatically add tasks with approaching deadlines to the `Today` list at a given time (i.e. 2 days before the deadline). This would preferably be implemented as a setting that users can change.

### Potential User Settings
- Enable or disable the auto-add
- Change the amount of time before a deadline the tasks should be added

_No response_

### Additional context

_No response_

Hetsavani:
## Problem

Tasks with deadlines set for today were not automatically shown in the Today view unless they were also manually scheduled. This made it easy to miss tasks that were due today but not explicitly planned.

Fixes the issue where deadline-based tasks were ignored by the Today list.

Fixes: #7488 

## Solution

Updated the Today task filtering logic to also include tasks whose deadline is today, including:

- all-day deadlines
- deadlines with a specific time today

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for working on this. Thinking about the model, I think the intended approach can be valid: if a deadline is today, auto-planning it by setting `dueDay = today` matches the current Today architecture where membership is derived from scheduling fields.

The part I’m worried about is doing it as a follow-up effect that dispatches `planTasksForToday`. That creates a second persistent op for the same user action (`setDeadline` / `addTask` first, then `planTasksForToday`), and it also applies all `planTasksForToday` side effects.

For the existing scheduling cases, I think we should make the policy explicit:

- If the task is already due today (`dueDay === today` or `dueWithTime` falls today), this should be a no-op for scheduling. Keep the existing time/reminder data.
- If the task has no due date/time, set `dueDay = today` and add it to Today order.
- If the task is overdue, moving it to `dueDay = today` seems reasonable.
- If the task is scheduled for a future day/time, this is the sensitive case. Either we explicitly decide that a today deadline wins over the previous plan, or we skip auto-planning already scheduled tasks. If we do move it, the behavior should be documented and tested because it changes the user’s previous scheduling intent.
- For `dueWithTime`, preserve it when it is today. If it is a future timestamp, we cannot keep it and also make Today membership consistent under the current due-field model, so the choice should be explicit: clear it while auto-planning, or skip auto-planning time-scheduled future tasks.

Implementation-wise, I think this would be safer folded into the same meta-reducer path as `setDeadline` / `addTask`, so one user action creates one persistent op. I would also preserve `remindAt` unless there is a specific reason to clear it. Auto-planning due to a deadline is not quite the same user intent as manually choosing “Move to Today”.

The immediate path should also mirror the selector path’s guards for done tasks and subtasks whose parent is already in Today, and docs/tests should reflect the chosen semantics: deadline today auto-plans the task for Today, not merely displays it there.

johannesjo:
To clarify this maybe: Generally all our store actions should be atomic. This is generally a good praticse, but especially important for sync to work.

johannesjo:
Thank you very much for this!

PR #7584, opened 2026-05-13
feat(tasks): convert tasks to subtasks by drag drop

teixr12:
## Summary

- allow dragging a top-level task into another task's subtask list to convert it into a subtask
- preserve existing subtask-to-subtask drag behavior while blocking repeating, issue-linked, self, and circular conversions
- clean converted tasks out of project, backlog, tag, Today, planner, and section top-level lists
- update the subtask docs for the new drag conversion flow

Fixes #905

## Verification

- `npm run checkFile src/app/root-store/meta/task-shared.actions.ts`
- `npm run checkFile src/app/root-store/meta/task-shared-meta-reducers/task-shared-crud.reducer.ts`
- `npm run checkFile src/app/root-store/meta/task-shared-meta-reducers/task-shared-crud.reducer.spec.ts`
- `npm run checkFile src/app/root-store/meta/task-shared-meta-reducers/section-shared.reducer.ts`
- `npm run checkFile src/app/root-store/meta/task-shared-meta-reducers/section-shared.reducer.spec.ts`
- `npm run checkFile src/app/features/tasks/task-list/task-list.component.ts`
- `npm run checkFile src/app/features/tasks/task-list/task-list.component.spec.ts`
- `npm run test:file src/app/features/tasks/task-list/task-list.component.spec.ts`
- `npm run test:file src/app/root-store/meta/task-shared-meta-reducers/task-shared-crud.reducer.spec.ts`
- `npm run test:file src/app/root-store/meta/task-shared-meta-reducers/section-shared.reducer.spec.ts`
- `NODE_OPTIONS=--max-old-space-size=4096 git push -u fork apxlbs/drag-task-to-subtask` pre-push hook, including lint, package tests, main Angular suite, and LA timezone suite

PR #7673, opened 2026-05-19
feat(tasks): convert tasks to subtasks by drag drop

### LINKED ISSUE #905: Adding a sub-task via drag-drop

JackSpiece:
Fixes #905

Gitpay task: https://gitpay.me/#/task/1146/adding-a-sub-task-via-drag-drop

## Summary

- Adds a persistent shared action for converting a top-level task into a subtask by drag and drop.
- Allows valid top-level tasks to be dropped into another task or its subtask list, including parents that do not already have subtasks.
- Keeps unsupported conversions blocked for tasks with subtasks, repeating tasks, issue-linked tasks, scheduled/due/reminder data, and subtask targets.
- Applies the same conversion guard in the shared reducer so programmatic moves cannot bypass the drag predicate.
- Removes converted tasks from top-level project, backlog, tag, today, planner, and section lists while recalculating parent time totals.
- Updates subtask docs for the new conversion flow and limits.

## Validation

- `git diff --check`
- `npm run checkFile` on the touched TypeScript, HTML, and SCSS files from the first pass
- `npx prettier --check` on the touched follow-up files
- `npx ng lint --lint-file-patterns ...` on the touched follow-up files
- Focused Angular test commands for the touched task-list, reducer, and helper specs compiled the browser bundles successfully, then stopped at browser launch because this Android Termux environment has no Chrome binary and `CHROME_BIN` is not set.

### DIFF
--- a/docs/wiki/2.04-Manage-Subtasks.md
+++ b/docs/wiki/2.04-Manage-Subtasks.md
@@ -33,7 +33,7 @@ Drag the subtask by its **drag handle** and drop it onto the target parent task,
 
 Drag a top-level task by its **drag handle** and drop it into another task, or into that task’s subtask list. The task becomes a subtask of the target parent and moves into that parent’s project.
 
-Tasks that already have subtasks cannot be converted this way, because that would create a deeper hierarchy. Repeating tasks and issue-linked tasks also stay as top-level tasks.
+Tasks that already have subtasks cannot be converted this way, because that would create a deeper hierarchy. Repeating, scheduled, reminder, and issue-linked tasks also stay as top-level tasks.
 
 ## Convert a Subtask to a Main Task
 
--- a/docs/wiki/2.04-Manage-Subtasks.md
+++ b/docs/wiki/2.04-Manage-Subtasks.md
@@ -29,6 +29,12 @@ Shortcuts are configurable in [[3.02-Settings-and-Preferences]].
 
 Drag the subtask by its **drag handle** and drop it onto the target parent task, or into that parent’s subtask list. The subtask becomes a child of the new parent and keeps the same project (subtasks inherit the parent’s project).
 
+## Convert a Main Task to a Subtask
+
+Drag a top-level task by its **drag handle** and drop it into another task, or into that task’s subtask list. The task becomes a subtask of the target parent and moves into that parent’s project.
+
+Tasks that already have subtasks cannot be converted this way, because that would create a deeper hierarchy. Repeating tasks and issue-linked tasks also stay as top-level tasks.
+
 ## Convert a Subtask to a Main Task
 
 Right-click the subtask and click **Convert to parent task**. The subtask becomes a top-level task in the same project. Repeating tasks and issue-linked tasks cannot be converted to subtasks; see [[4.11-Subtasks]].
--- a/docs/wiki/4.11-Subtasks.md
+++ b/docs/wiki/4.11-Subtasks.md
@@ -21,8 +21,9 @@ Not every task can be turned into a subtask, and some behaviors are constrained 
 
 ### Tasks That Cannot Be Converted to Subtasks
 
-Two kinds of tasks **cannot** be converted into subtasks:
+Three kinds of tasks **cannot** be converted into subtasks:
 
+- **Tasks that already have subtasks**: Converting these would create a deeper hierarchy than the main drag-and-drop workflow supports. Move or convert the existing subtasks first, then convert the parent task.
 - **Repeating tasks** — Tasks that recur on a schedule (for example daily or weekly) cannot be made subtasks of another task. See [[4.13-Repeating-Tasks]]. The repeat rule applies to a single task; nesting would conflict with that.
 - **Issue-linked tasks** — Tasks that are linked to an external issue (for example Jira or GitHub) cannot be made subtasks. The issue link is a one-to-one relationship between a task and an issue; the app does not support issue-linked subtasks in the same way.
 
--- a/docs/wiki/4.11-Subtasks.md
+++ b/docs/wiki/4.11-Subtasks.md
@@ -21,10 +21,11 @@ Not every task can be turned into a subtask, and some behaviors are constrained 
 
 ### Tasks That Cannot Be Converted to Subtasks
 
-Three kinds of tasks **cannot** be converted into subtasks:
+Four kinds of tasks **cannot** be converted into subtasks:
 
 - **Tasks that already have subtasks**: Converting these would create a deeper hierarchy than the main drag-and-drop workflow supports. Move or convert the existing subtasks first, then convert the parent task.
 - **Repeating tasks** — Tasks that recur on a schedule (for example daily or weekly) cannot be made subtasks of another task. See [[4.13-Repeating-Tasks]]. The repeat rule applies to a single task; nesting would conflict with that.
+- **Scheduled tasks or tasks with reminders** — Tasks with a due date, due time, or reminder stay as top-level tasks so their schedule and reminder state cannot become orphaned.
 - **Issue-linked tasks** — Tasks that are linked to an external issue (for example Jira or GitHub) cannot be made subtasks. The issue link is a one-to-one relationship between a task and an issue; the app does not support issue-linked subtasks in the same way.
 
 If you try to convert such a task to a subtask (for example by indenting it in a markdown-based workflow), the app will warn you and leave the task as a top-level task. You can still create normal subtasks under a parent and use repeating or issue-linked tasks alongside them as separate tasks.
--- a/src/app/features/schedule/schedule-week/schedule-external-drag.service.ts
+++ b/src/app/features/schedule/schedule-week/schedule-external-drag.service.ts
@@ -5,7 +5,7 @@ import { DragRef } from '@angular/cdk/drag-drop';
 @Injectable({ providedIn: 'root' })
 export class ScheduleExternalDragService {
   private readonly _activeTask = signal<TaskWithSubTasks | null>(null);
-  private _activeDragRef: DragRef | null = null;
+  private readonly _activeDragRef = signal<DragRef | null>(null);
   private _cancelNextDrop: boolean = false;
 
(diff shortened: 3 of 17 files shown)


johannesjo:
FYI: There is also this PR already: https://github.com/super-productivity/super-productivity/pull/7584

Though it seems abandoned.

johannesjo:
Thanks for the PR — the sync model here is solid (single meta-reducer pass, persistent `Move` op, atomic section/planner/project/tag cleanup, good spec coverage). Two issues block merge:

### 🔴 Subtask → childless parent has no drop target

This is the "not quite working for sub task to parent tasks" case. `isEmptySubTaskDropTargetVisible` is gated by `\!activeTask.parentId`, so the empty drop zone renders **only when dragging a top-level task**. Dragging a *subtask* onto a parent that has no subtasks → no drop target appears → drop silently impossible.

`enterPredicate` (isSubtask branch) and `_move` (`moveSubTask`) already support subtask → childless-parent; the only missing piece is the DOM drop list. So reparenting a subtask works onto parents that already have subtasks, but not onto childless ones.

Fix — also reveal the zone for a subtask reparent:

```ts
const isTopLevelConvert =
  \!activeTask.parentId &&
  (activeTask.subTaskIds?.length ?? 0) === 0 &&
  \!activeTask.repeatCfgId && \!activeTask.issueId &&
  \!activeTask.issueProviderId && \!activeTask.issueType;
const isSubTaskReparent =
  \!\!activeTask.parentId && activeTask.parentId \!== t.id;
return (
  \!this.isInSubTaskList() && \!t.parentId &&
  t.id \!== activeTask.id &&
  (t.subTaskIds?.length ?? 0) === 0 &&
  \!\!activeTask &&
  (isTopLevelConvert || isSubTaskReparent)
);
```

Please test live that the mid-drag-mounted `cdkDropList` is actually picked up by the active drag — the empty `task-list` mounts only after `onDragStarted` sets `activeTask`, and CDK does not always honor drop lists created after drag start.

### 🟠 Scheduled task → subtask orphans the reminder

`enterPredicate` / `canConvertParentDropToSubTask` block repeat and issue-linked tasks, but not scheduled ones (`dueWithTime` / `dueDay` / `reminderId`). `handleConvertToSubTask` changes `parentId`/`projectId`/`tagIds` but never clears the schedule, so you end up with a nested subtask carrying a live `reminderId` — and a meta-reducer can't delete the reminder (it lives outside NgRx state). `moveSubTask` isn't a precedent: it moves tasks that are already subtasks (no schedule).

Simplest fix: block scheduled tasks in `canConvertParentDropToSubTask` (`&& \!task.dueWithTime && \!task.dueDay && \!task.reminderId`), same as repeat/issue, and add it to the wiki "cannot convert" list.

### 🔵 Minor

- Conversion-eligibility predicate duplicated in 3 places (`canConvertParentDropToSubTask`, `isEmptySubTaskDropTargetVisible`, `handleConvertToSubTask`) — drift risk, extract one helper.
- `_move` convert block has no task-validity guard; reducer re-validates so it's data-safe, but an invalid drop becomes a silent no-op.
- When target parent's subtasks are hidden (`isHideAll`), `afterTaskId` resolves `null` so the new subtask lands at the front of `subTaskIds` regardless of drop position — minor ordering surprise.

jackspiece:
Thanks, pushed bd48d96 addressing the blocking points.

What changed:
- extracted the conversion checks into shared helpers so the drag predicate, empty drop target visibility, and reducer guard stay aligned
- allowed the empty childless-parent drop target for subtask reparenting
- set the active drag task on pointerdown, before CDK drag start, so the empty subtask drop list is mounted before the drag begins instead of appearing mid-drag
- blocked scheduled top-level tasks from converting into subtasks in both the drag predicate and reducer guard, covering dueWithTime, dueDay, remindAt, and legacy reminderId
- added focused helper, drag predicate, and reducer regression coverage

Local validation:
- Prettier check passed for the touched files
- ng lint with lint-file-patterns passed for the touched files
- the focused ng test command compiled the browser bundle successfully, but Karma could not launch specs in this Termux environment because there is no Chrome binary available. I am not marking the browser specs as locally passed.

johannesjo:
Now it does not seem to work in either direction?

https://6488f3e1.super-productivity-preview.pages.dev/#/tag/TODAY/tasks

johannesjo:
Re-review after `bd48d96` / `55e00f2b` — prior blockers (subtask→childless-parent reparent, scheduled-task block, predicate dedup) look resolved.

### 🟠 Pointerdown trigger causes empty-drop-slot flicker on every click

`onDragPrepared` runs on every `pointerdown` on a draggable task and calls `setActiveTask(task)`. That flips `isEmptySubTaskDropTargetVisible` for every childless top-level parent on screen, so the `@expandInOnly` empty 12px drop slot mounts on **any click** (edit/play button, text, scroll-grab) until `pointerup` fires cleanup. On touch it's worse — between `pointerdown` and CDK firing `dragStarted` after `dragDelayForTouch` (~300ms), every tap on a task expands empty drop slots across the list.

CDK does need the `cdkDropList` mounted *before* drag start (that is the point of pointerdown), but the *visual* slot doesn't. Fix: keep DOM mount on `pointerdown`, gate the visible style on `activeDragRef() \!== null` (or scope the `is-empty-drop-target` class behind a CDK dragging signal). Please live-test on both touch and mouse before merge.

### 🔵 Minor

- **task-list.component.ts:435-447** — `_move` PARENT→SUB branch dispatches `convertToSubTask` without re-checking `canConvertTaskToSubTask(task)` / self-target. Reducer rejects invalid cases (data-safe), but user sees silent no-op and a Move op still gets logged. Cheap to add the guard at dispatch site.
- **task-shared-crud.reducer.ts handleConvertToSubTask** — `afterTaskId = getAnchorFromDragDrop(...)`; when target parent's subtasks are collapsed (`_hideSubTasksMode === HideAll`) the visible list is empty so anchor resolves to `null` → new subtask prepends regardless of drop position. Prior round flagged, still unaddressed. Either expand the subtask list on hover-over-collapsed target, or append (last) instead of prepend when source list is empty.
- **task-list.component.ts:81-88 / _move / reducer guard** — three sites carry the convert eligibility check (`\!RESERVED_LIST_IDS.has(target) && targetListId === 'SUB'`). Could collapse to one `isConvertDrop(...)` helper to keep them aligned over time.
- **task.component.html:308** — `[isHideAll]="\!\!t.subTasks?.length && \!\!t._hideSubTasksMode"` overloads the existing prop for the empty-parent drop-target mode. Acceptable, but verify no other consumer reads `isHideAll` as "user toggled collapse".

### Strengths
- Single helper drives predicate, visibility, and reducer guard — drift-resistant.
- One reducer pass = one Move op (sync rule 3); section/project/tag/planner/today cleanup atomic.
- Spec coverage for helper, predicate, and reducer (cross-project, backlog, planner, scheduled variants).
- Action carries `PersistentActionMeta` / `OpType.Move` for op-log replay.

jackspiece:
Pushed 2b73bbf addressing the pointerdown follow-up.

What changed:
- kept the empty subtask cdkDropList mounted during pointerdown, but collapse the prepared-only container to zero layout
- apply the visible empty-drop-target styling only once cdkDragStarted has set activeDragRef
- added the dispatch-site convert guard for invalid parent to subtask drops
- append after the last existing subtask when converting into a hidden/collapsed target list
- added focused coverage for the guard and hidden-list append behavior

Local validation:
- Prettier check passed for the touched files
- Angular ESLint passed for the touched TS/HTML files
- Stylelint passed for the touched SCSS file

I cannot honestly mark the mouse/touch browser pass as locally verified from this Termux device, so I am leaving that for the updated preview/CI rather than overstating it.

johannesjo:
Thanks, Both directions don't seem to be working:
https://4c645ff4.super-productivity-preview.pages.dev/

jackspiece:
Pushed 1ec3f87 as a follow-up to the preview failure.

I think my previous prepared-state CSS made the mounted drop list unusable for CDK because it collapsed the actual drop-list hit area to zero. This update keeps the prepared drop list measurable at 12px but absolutely positioned and invisible, so it should still be registered before drag start without creating the click/touch flicker.

Validation from this device:
- Prettier passed for the touched SCSS file
- stylelint passed for the touched SCSS file
- focused Angular test build completed bundle generation with increased heap; it still cannot run specs here because there is no real browser available in Termux, and `--browsers=false` is treated as a browser name by this setup

I am not claiming live mouse/touch verification locally. The updated preview should be the real check for the two drag directions.

jackspiece:
Pushed 81749a6 with a simpler approach.

The empty childless-parent drop lists are now mounted all the time for eligible top-level tasks, but stay invisible and absolutely positioned until an actual CDK drag starts. That removes the pointerdown preparation path entirely, so ordinary clicks/taps should no longer change activeTask or flicker the empty slot. CDK should also see the drop lists before drag start because they are already in the DOM.

I also added focused component coverage for the mountable state.

Validation from this device:
- Prettier passed for touched files
- Angular lint passed for touched TS/HTML/spec files
- stylelint passed for the touched SCSS file
- focused Angular test bundle generation completed with increased heap; Karma still cannot execute specs here because there is no browser available in Termux and `--browsers=false` is treated as a browser name by this setup

The new preview/CI should be the real check for both drag directions.

johannesjo:
Not working ....

ChilledCosmonaut:
@JackSpiece do you mind if I take a look as well? I'm really looking forward to this feature as well!

Or @johannesjo would you prefer the implementation in #7584 as I saw that you are already fine with the one-directional implementation?

johannesjo:
> Or @johannesjo would you prefer the implementation in #7584 as I saw that you are already fine with the one-directional implementation?

No preference on my side- Important is that it works and that the performance impact is minimal, while hopefully not adding too much technical debt :)

johannesjo:
This has since been solved. Thank you very much!

PR #7679, opened 2026-05-19
chore: limit "resolved" actions to correct categories

zenoprax:
GH Action will incorrectly mark discussions without the question-answer
format as answered when that is not possible. The action should be fixed
but restricting to the correct categories will also avoid this problem.


johannesjo:
Thank you!

Issue #6510, opened 2026-02-13
🚨 "Manually start breaks" and "Pause task tracking during breaks" stop time tracking when a Pomodoro session auto-completes


beerkumquatpome:

johannesjo:
Thank you very much! This will be fixed with the next release!

Issue #5974, opened 2026-01-13
🚨 You can break Pomodoro timer syncing


beerkumquatpome:
### Version Used

v17.0.0-RC.1

### Expected Behavior

-

### Current Behavior

-

### Steps to Reproduce

Bug 1: Pressing the time tracking button after the session ends but before manually starting a break

Settings:
1. Sync focus sessions with time tracking
2. Start focus sessions with banner only (no overlay)
3. Skip preparation screen (rocket animation)
4. Manually start breaks (Pomodoro)

Steps:

1. Start the Pomodoro timer
2. Wait until it ends
    (It should say "Pomododoro Session Completed" in the banner)
3. Press the time tracking button to stop time tracking
4. Press "Start" in the banner to manually start the break

It's not time tracking. ("Pause task during during breaks" is disabled so it should still be time tracking)

Bug 2: Pausing the break from the banner

Settings:
1. Sync focus sessions with time tracking
2. Start focus sessions with banner only (no overlay)
3. Skip preparation screen (rocket animation)

Steps:

1. Start the Pomodoro timer
2. Wait until it ends
3. Press "Pause" in the banner to pause the break

It's still time tracking (It should stop tracking)

Visual bug: Pressing the time tracking button to pause the break doesn't update the icon in the banner

Settings:
1. Sync focus sessions with time tracking
2. Start focus sessions with banner only (no overlay)
3. Skip preparation screen (rocket animation)

Steps:

1. Start the Pomodoro timer
2. Wait until it ends
3. Press the time tracking button to pause the break
4. The break is paused but it's still showing a pause icon in the banner
    It should show a play icon

Typo:

Settings: "Sync focus session with time tracking" is missing an s. It should be "focus session**s**"

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
Thank you very much! This wil be improved with the next version!

Issue #5954, opened 2026-01-11
🚨 You can break Pomodoro timer syncing


beerkumquatpome:
### Version Used

v16.9.4

### Expected Behavior

-

### Current Behavior

-

### Steps to Reproduce

Settings:
1. Sync focus sessions with time tracking
2. Start focus sessions with banner only (no overlay)
3. Skip preparation screen (rocket animation)

Bug 1: Close the app and enter focus mode to start the Pomodoro timer

1. Close the app
2. Enter focus mode (Press F)
3. Start the Pomodoro timer
4. The Pomodoro timer will start without time tracking because it's not assigned to a task

The Pomodoro timer should automatically be assigned to the last task that you worked on

Bug 2: Manually ending the Pomodoro session

1. Press the time tracking button 
    It will start time tracking and automatically assign the Pomodoro timer to a task
2. Press "End session" in the banner
3. It will stop time tracking

"Pause task tracking during breaks" is disabled so it should still be time tracking

Bug 3: Manually  ending the Pomodoro session but with "Manually start breaks (Pomodoro)" enabled

Basically the same as bug 2 but with "Manually start breaks (Pomodoro)" enabled

1. Press the time tracking button
    It will start time tracking and automatically assign the Pomodoro timer to a task
2. Press "End session" in the banner
3. Press "Skip Break" in the banner
4. The Pomodoro timer will start without time tracking (Maybe because it's not assigned to a task anymore?)

Bug 4: Pressing the time tracking button during a break

Make sure "Pause task tracking during breaks"  and "Manually start breaks (Pomodoro)" are disabled

Steps:
1. Press the time tracking button
2. Wait until the session ends (Don't manually end it because of Bug 2)
    It should still be time tracking after the break has started
3. Pressing the time tracking button will pause time tracking but won't pause the Pomodoro break

"Pause task tracking during breaks" is disabled, so pressing the time tracking button during a break should also pause the Pomodoro break
And pressing the time tracking button again should continue the break

Minor bug: Banner Typo

1. Start the Pomodoro timer
    "Pomodoro Session # 1 is running"
2. Press "End session" in the banner
3. "Break # 2 is running"
    It should be "Break # 1 is running"

Suggestion: Change "Focus break is over!" to "Break is over!"
"Focus break is over!" makes me think I'm in a Pomodoro session and I should focus

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

[...]

PR #7681, opened 2026-05-19
feat(focus-mode): auto-start break after manual session completion

### LINKED ISSUE #7663: 💡 If "Manually start breaks" is enabled, pressing the "Complete focus session" button should automatically start a break

Hetsavani:
## Problem

Problem Statement
1. Press "Complete focus session"
2. It doesn't automatically start a break. I need to manually press "Start break"
2 clicks

Fixes: #7663 

## Solution

After pressing "Complete focus session", it should automatically start a break
1 click

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Did two independent passes on this. **Core change is correct and bugs #5954 / #5974 do not regress**, but the PR leaves dead code and a misleading config label behind that should be cleaned up before merge.

## Correctness — looks good

Walked through the new effect chain carefully:

- **#5954** (`isPauseTrackingDuringBreak=true`): `autoStartBreakOnSessionComplete$` now passes `pausedTaskId` directly into `startBreak`; reducer captures it (`focus-mode.reducer.ts:231`). `skipBreak$` / `resumeTrackingOnBreakComplete$` resume the task as before.
- **#5974** (`isPauseTrackingDuringBreak=false`): the pre-PR window ("user stops tracking before manually starting break") no longer exists since the break now auto-starts. If the user stops tracking *during* the break, `syncTrackingStopToSession$` (`focus-mode.effects.ts:143-164`) captures `pausedTaskId` via `pauseFocusSession`. Net behavior preserved.
- **Dispatch ordering**: confirmed `autoStartBreakOnSessionComplete$` reads `currentTaskId$` synchronously *before* `stopTrackingOnSessionEnd$` runs its `unsetCurrentTask`, due to NgRx's re-entrant synchronous dispatch (effects subscribe in class-declaration order, `incrementCycleOnSessionComplete$` at `:307` before `stopTrackingOnSessionEnd$` at `:321`). No race condition, no double `unsetCurrentTask`.
- **Bug #6510**: the unit-test invariant at `focus-mode.bug-5875.spec.ts:381-406` still holds — the filter at `focus-mode.effects.ts:344-346` still returns false for auto-completion with break-will-start. The *semantic* intent of #6510 shifts slightly (break auto-starts on manual completion), but that's the intended trade-off of #7663.
- Focus-mode state isn't synced / op-logged, so removing the `setPausedTaskId` dispatch has no sync-correctness impact.

## Issues to address before merge

### 1. `isManualBreakStart` flag is now misnamed

After this PR, the flag controls only **Pomodoro overtime** (`setOvertimeOnSessionStart$` at `focus-mode.effects.ts:454-468`, `detectSessionCompletion$` at `:256-279`). The label at `src/assets/i18n/en.json:2195`:

```
"L_MANUAL_BREAK_START": "Manually start breaks (Pomodoro)"
```

is inaccurate now — the break is always automatic. The issue author even wrote "'Manually start breaks' is for Pomodoro overtime." Suggest renaming to something like "Enable Pomodoro overtime (keep timer running until manually ended)". The field key itself (`isManualBreakStart`) has a migration cost, but the i18n label is cheap to change and important for users.

### 2. Dead code in session-done component

The SessionDone screen for Pomodoro is now replaced by Break on the same tick. Several pieces become unreachable:

- `src/app/features/focus-mode/focus-mode-session-done/focus-mode-session-done.component.html:38-46` — the `@if (mode === Pomodoro && isManualBreakStart)` branch with the "Start break" button. The `@else` should become unconditional.
- `focus-mode-session-done.component.ts:96-120` — `startBreakManually()` plus related imports (`startBreak`, `unsetCurrentTask`, `getBreakCycle`, `selectCurrentCycle`, `FocusModeStrategyFactory`).
- `focus-mode-session-done.component.spec.ts:167-236` — the `startBreakManually` describe block now exercises dead code.

(Confetti in `ngAfterViewInit` stays meaningful for Flowtime / Countdown modes that don't auto-start a break — leave it.)

### 3. Possible UX regression — please confirm intent

Before: a Pomodoro user with `isManualBreakStart=true` saw three options on SessionDone — Start break, Continue with focus session, Start next focus session. After: only "Skip break" inside the Break screen. The linked issue only asked to eliminate the second "Start break" click; losing "Continue / Start next" as one-click options is a side effect worth calling out in the PR description (and worth checking with the issue author).

### 4. Stale comment

`focus-mode.effects.ts:343` references `FocusModeService.startAfterSessionComplete()` — that method doesn't exist. Pre-existing, but this PR is the right moment to clean up the surrounding comment.

### 5. Unit-test coverage gap

The 7 removed unit tests were appropriate (they covered the deleted effect). The renamed survivor at `focus-mode.effects.spec.ts:406-427` only asserts `startBreak` fires — it doesn't pin down `pausedTaskId` flow. Worth adding two cases against `autoStartBreakOnSessionComplete$`:

- `(isManualBreakStart=true, isPauseTrackingDuringBreak=true, currentTaskId='X')` → dispatches both `unsetCurrentTask` and `startBreak({ pausedTaskId: 'X' })`.
- `(isManualBreakStart=true, isPauseTrackingDuringBreak=false, currentTaskId='X')` → does NOT dispatch `unsetCurrentTask`, dispatches `startBreak({ pausedTaskId: undefined })`.

These lock down the #5954 / #5974 invariants under the new code path. The existing E2Es (`e2e/tests/focus-mode/pomodoro-timer-sync-bug-5954.spec.ts` / `5974.spec.ts`) don't toggle `isManualBreakStart`, so they're not catching this matrix either.

### 6. Docs

`docs/wiki/4.15-Timers-and-Focus-Mode.md:126` references "manual break start" — should be updated to reflect overtime semantics.

johannesjo:
Thanks for the v2 — the cleanup pass addressed almost everything from my last review (label rename, dead `startBreakManually` removed, wiki updated, test matrix added, `storePausedTaskOnManualBreakSession$` deleted, stale comment fixed). One new concern, and a few smaller follow-ups.

## Scope creep — Flowtime behavior changed silently

`offerFlowtimeBreakOnSessionEnd$` now dispatches `startBreak` directly instead of `offerFlowtimeBreak` (`focus-mode.effects.ts:413-451`). That kills the entire Flowtime *break-offer* flow — previously users got a `BreakOffer` screen they confirmed; now the break timer starts immediately.

This appears driven by @beerkumquatpome's drive-by comment, not by the linked issue #7663 (which is Pomodoro-only). It's not surfaced in the PR title or description, and it leaves a lot of dead code behind:

- `focus-mode.actions.ts:67-74` — `offerFlowtimeBreak` action: never dispatched.
- `focus-mode.reducer.ts:254-272` — `on(a.offerFlowtimeBreak, …)` handler: unreachable.
- `focus-mode-break.component.ts:62-65` `isBreakOffer` computed signal: permanently false.
- `focus-mode-break.component.html:3, 26, 55` — `@if (isBreakOffer())` template branches: unreachable.
- `focus-mode.reducer.spec.ts:432-440` — reducer spec for `offerFlowtimeBreak`: still passes but exercises dead code.
- `focus-mode.effects.ts:651-687` — `logFocusSession$` still in `ofType(...)` and filter-by-type-checks `offerFlowtimeBreak`: dead branch.
- Effect name `offerFlowtimeBreakOnSessionEnd$` no longer matches what it does.

**Suggested resolution — pick one:**

1. **Trim back to issue scope.** Revert the Flowtime hunk; keep PR Pomodoro-only. Open a separate issue for the Flowtime auto-start-break request so it gets discussed on its own merits.
2. **Commit to the change.** Remove the dead `offerFlowtimeBreak` action / reducer / spec / template branches / `logFocusSession$` filter case, rename the effect (e.g. `autoStartBreakOnFlowtimeEnd$`), and update the PR title + description to call out the Flowtime change.

I'd lean toward (1) — the issue author and reviewers are evaluating a Pomodoro change; the Flowtime variant deserves its own thread.

## Smaller items

### Stale reducer comment
`focus-mode.reducer.ts:197-202` still says *"offerFlowtimeBreak reads the current timer state and transitions to the Break screen"* and *"lastCompletedDuration must remain populated across the offerFlowtimeBreak transition."* If you keep the Flowtime change, update to reference `startBreak`.

### `logFocusSession$` filter is convoluted
`focus-mode.effects.ts:651-687` does `ofType(completeFocusSession, offerFlowtimeBreak, startBreak)` then runtime-checks `action.type` + mode. With `offerFlowtimeBreak` removed, the simpler shape is `ofType(completeFocusSession, startBreak)` + filter `startBreak` only for Flowtime — or split into two effects. Easier to follow.

### `isManualBreakStart` field name is now misnamed
After this PR the flag only controls Pomodoro *overtime* (`setOvertimeOnSessionStart$` at `focus-mode.effects.ts:495-512`). The i18n label is fixed, but the field itself still reads "manual break start". I get why the field name isn't being changed here (state-migration cost) — a one-line code comment at the field's declaration noting the semantic shift would help future readers.

## CI
`Tests` is red, but the failures (`Cannot redefine property: clipboard`) are in the Ctrl/Cmd+C copy-task spec from master (`64c622fab1`), not from anything in this PR. Not blocking — will likely persist after rebase.

## Verdict
Needs changes — primarily the Flowtime scope decision. Correctness of the Pomodoro path is fine; my prior analysis of `pausedTaskId` / `unsetCurrentTask` ordering still holds against the v2 diff.

Hetsavani:
Good point regarding the Flowtime scope change. I’ll trim this PR back to the original Pomodoro-only behavior and revert the Flowtime auto-start change to keep the scope aligned with #7663.

johannesjo:
Thank you!

PR #7682, opened 2026-05-19
add deadline to synched fields

powerpaul17:
## Problem

I want to sync deadlines also with plugins

## Solution

Added the deadline fields to the syncable task fields

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/packages/plugin-api/src/issue-provider-types.ts
+++ b/packages/plugin-api/src/issue-provider-types.ts
@@ -19,6 +19,10 @@ export interface PluginSearchResult {
   duration?: number;
   /** True if this is an all-day event */
   isAllDay?: boolean;
+  /** Deadline date as YYYY-MM-DD string. For date-only deadlines (Frist). */
+  deadlineDay?: string;
+  /** Deadline timestamp (ms). For deadlines with a specific time (Frist). */
+  deadlineWithTime?: number;
   /** Event description / body text */
   description?: string;
   /** Provider-specific fields used by issue display and field mappings */
@@ -77,6 +81,8 @@ export interface PluginFieldMapping {
     | 'notes'
     | 'dueDay'
     | 'dueWithTime'
+    | 'deadlineDay'
+    | 'deadlineWithTime'
     | 'timeEstimate'
     | 'tagIds';
   issueField: string;
--- a/packages/plugin-api/src/types.ts
+++ b/packages/plugin-api/src/types.ts
@@ -230,6 +230,9 @@ export interface Task {
   remindAt?: number | null;
   dueDay?: string | null;
   dueWithTime?: number | null;
+  deadlineDay?: string | null;
+  deadlineWithTime?: number | null;
+  deadlineRemindAt?: number | null;
   repeatCfgId?: string | null;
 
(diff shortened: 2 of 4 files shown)


johannesjo:
Thank you very much for tackling this! A few things to address before this is ready:

### Blocker: `setDeadline` action isn't pushed
The user-facing deadline dialog dispatches `TaskSharedActions.setDeadline` (`src/app/features/tasks/dialog-deadline/dialog-deadline.component.ts:242,254`), but that action isn't in the `ofType(...)` list or `ACTION_EXTRACTORS` map in `issue-two-way-sync.effects.ts`. So adding `deadlineDay`/`deadlineWithTime` to `SYNCABLE_TASK_FIELDS` only fires when the fields ride through a generic `updateTask` payload — the primary deadline-setting path is silently skipped.

For parity with `scheduleTaskWithTime`/`unscheduleTask`, you'll need to:
- Add `TaskSharedActions.setDeadline` to the `ofType(...)` list (line 128).
- Add an `ACTION_EXTRACTORS` entry returning `{ taskId: a.taskId, changes: { deadlineDay: a.deadlineDay, deadlineWithTime: a.deadlineWithTime } }`, ideally clearing the opposite field when one is set (mirroring the reducer at `task-shared-meta-reducers/task-shared-deadline.reducer.ts:128-130`).

### `deadlineRemindAt` is exposed but never wired
You added `deadlineRemindAt?: number | null` to the plugin-api `Task` type, but it's not in `SYNCABLE_TASK_FIELDS`, not in `PluginFieldMapping.taskField`, and not on `PluginSearchResult`. Either commit to syncing it or drop it from the public type — half-exposed fields invite plugin authors to set values that won't round-trip. (Read-only-for-inspection is also fine, like `remindAt`, but worth a JSDoc note.)

### Minor

- **Style inconsistency in `_buildBaseIssueTask`**: the new deadline branch writes `{ deadlineWithTime, deadlineDay: null }` while the existing dueWithTime branch right above it writes just `{ dueWithTime }`. Both work because the base object is fresh; either drop the explicit nulls for symmetry, or update the due block too.
- **"(Frist)" in JSDoc** (`issue-provider-types.ts:22,25`): the plugin-api is an English public surface — drop the German parenthetical.
- **`PluginCreateTaskData`** doesn't gain deadline fields. Probably out of scope, but plugins creating tasks with deadlines would need this eventually.
- **Plugin authoring contract**: `_buildBaseIssueTask` reads deadline fields off `PluginIssue` (which is `[key: string]: unknown`), but plugin authors only see them typed on `PluginSearchResult`. A note that `getById()` responses should also expose deadline fields would close the doc gap.
- **No tests** — at minimum a unit test for the mutual-exclusivity branches in `_buildBaseIssueTask`, and if you wire `setDeadline`, an effect spec confirming the push.

### Sync-correctness
No concerns — the effect already injects `LOCAL_ACTIONS` and this PR doesn't add a new effect or multi-entity write, just extends an allow-list and a constructor.

PR #7685, opened 2026-05-20
chore(deps): add cooldown for NPM and GH Actions to reduce supply chain attack risk

zenoprax:
- Monthly/weekly update frequency reduces the window of exposure but the addition of a cooldown should help narrow it further.
- Implemented as suggested in this gist: https://gist.github.com/mcollina/b294a6c39ee700d24073c0e5a4e93104
- I've opted to use the more conservative 3-day cooldown as this is the default in some tools.
- Dependabot needs to be adjusted in combination with .npmrc to avoid creating PRs that can't succeed.

Ref.
https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,7 +13,8 @@ updates:
       day: 'monday'
       time: '09:00'
       timezone: 'Europe/Berlin'
-
+    cooldown:
+      default-days: 3
     open-pull-requests-limit: 10
 
     reviewers:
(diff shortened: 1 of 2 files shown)


bitdivine:
Good idea!  I default to 1 week cooldown.  I thought I wasn't being particularly conservative. :sweat_smile:

johannesjo:
Great idea! Thank you very much @zenoprax !

Issue #5357, opened 2025-10-24
macOS application prevents system shutdown


shynst:
### Version Used

16.0.0

### Expected Behavior

When shutting down or rebooting the Mac all applications should exit, allowing the OS to proceed.

### Current Behavior

Super Productivity doesn't close, preventing macOS to shutdown. Instead a dialog is displayed:

> Your Mac hasn't shut down because "Super Productivity" failed to quit. To continue shutting down, quit "Super Productivity". To quit an unresponsive app, select Force Quit from the Apple Menu (or press Command-Option-Esc).

The application is still responsive and can be quit manually. After that the system shuts down.

### Steps to Reproduce

Shut down your Mac.

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

shynst:
Seems to be a common issue with electron applications on macOS.

This might be a solution:
https://stackoverflow.com/questions/51804259/how-to-prevent-electron-app-from-interrupting-macos-system-shutdown

johannesjo:
Are you using sync? Or is time tracking still in pogress when you attempt to shutdown? 

Super Productivity will attempt to sync before being closed, so this might be the reason it blocks shutdown.

shynst:
Yes, I am using sync and I need it! Time tracking was not active.
It's not a biggie, but would be great if you could look into it.

Thanks for your awesome app btw!

YgLK:
I encountered the same issue. The app prevented my Mac from shutting down, so I had to force-quit the process.

shynst:
No need to force-quit. A normal "Quit" (`<command>-Q`)  will do. However, the app is not responding to shutdown events sent by macOS to all apps requesting to terminate.

johannesjo:
Help with this would be vey welcome!

shynst:
I have created a pull request that fixes the issue for me. Please review.

PR #7704, opened 2026-05-20
fix(macOS): quit app on system shutdown

shynst:
## Problem

When shutting down or rebooting the Mac all applications should exit, allowing the OS to proceed.
Super Productivity doesn't close, preventing macOS to shutdown. Instead a dialog is displayed:

> Your Mac hasn't shut down because "Super Productivity" failed to quit. To continue shutting down, quit "Super Productivity". To quit an unresponsive app, select Force Quit from the Apple Menu (or press Command-Option-Esc).

The application is still responsive and can be quit manually. After that the system shuts down.

See issue: #5357

## Solution

If the app is closed explicitly via its menu item "Quit", `quitApp()` is invoked and the app is closed correctly.

When shutting down or trying to close the app from it's dock icon, the app will receive a `before-quit` event handled in `electron/start-app.ts`. Here the immediate quit is prevented and delegated to the window close handler. The core of the problem is that in macOS - unlike with other OS - closing all windows alone does not quit the app!

To solve this, a call to `quitApp()` was added for macOS.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for tackling this, the macOS shutdown dialog is a genuinely annoying bug. Did a deeper review and want to flag one correctness concern before this lands, plus a couple of small nits.

## Concern: bypasses the before-close IPC wait on macOS

`if (IS_MAC) quitApp();` runs synchronously on the same tick as `win.close()`, which means it flips `isQuiting=true` and calls `app.quit()` **before** the close handler's IPC roundtrip completes. Trace:

1. `before-quit` fires with `isQuiting=false`, calls `preventDefault()`, then `win.close()`
2. Close handler at `main-window.ts:511-535` sees `!isQuiting`, calls `event.preventDefault()`, sends `IPC.NOTIFY_ON_CLOSE`, and is now waiting on `IPC.BEFORE_CLOSE_DONE`
3. `quitApp()` runs the next line: `setIsQuiting(true); app.quit()`
4. `before-quit` re-fires with `isQuiting=true`, cleanup branch runs
5. `will-quit` fires, `ipcMain.removeAllListeners()` removes the `BEFORE_CLOSE_DONE` handler at `main-window.ts:501-509`
6. Renderer's `BEFORE_CLOSE_DONE` arrives later, no listener, the registered work is abandoned

The work that gets dropped includes:
- Final sync upload (`SYNC_BEFORE_CLOSE_ID` in `src/app/imex/sync/sync.effects.ts`)
- Auto finish-day flow (`src/app/features/finish-day-before-close/finish-day-before-close.effects.ts`)

The existing comment at `start-app.ts:419-420` (\"Safe to remove IPC listeners here: all windows are closed and before-close IPC flows ... are guaranteed to have completed\") explicitly relies on this invariant, which the new code breaks on macOS Dock-Quit and system shutdown, exactly the paths users may have unsynced edits on.

### Suggested approach

Move the macOS-specific `app.quit()` into the close-handler completion path so it runs *after* the IPC roundtrip resolves. Roughly, in `main-window.ts`'s `_quitApp` (or the `BEFORE_CLOSE_DONE` handler that calls it), do an extra `if (IS_MAC) app.quit()` once `mainWin.close()` has actually destroyed the window. That preserves the before-close contract while still kicking the app over the macOS shutdown finish line.

## Small nits

- **Title typo**: `fix(maxOS):` should be `fix(macOS):`
- **Comment**: `// On macOS call quitApp(), since closing windows alone will not quit the app` reads as a near-restatement of the code. The existing block at lines 427-432 sets a good bar. Maybe: \`// macOS keeps the app alive after window close (window-all-closed doesn't quit by convention), so explicitly call quitApp() to actually exit on shutdown/Dock-Quit.\`
- Minor shape suggestion to make the branches feel less parallel:
  \`\`\`ts
  if (win && !win.isDestroyed()) {
    win.close();
    if (!IS_MAC) return;
  }
  quitApp();
  \`\`\`

shynst:
Thanks for you swift and detailed response! I was worried about closing early, too. So I will follow the path you proposed.
However, when closing the app "normally" via it's menu item "Quit" `quitApp()` is called directly as well. How is this any different? Or am I missing something here?

shynst:
Actually we do not need to treat macOS different here as it seems. What was blocking the app from closing was leaving "minimize to tray" on. This had the effect of only hiding the main window and never triggering `_quitApp()`.

If we disable "minimize to tray" just before we want to quit, everything runs smoothly. Including gracefully terminating registered workers.

PS: I could also verify that closing the app normally will **not** wait for workers to finish!

electron/start-app.ts
```diff
@@ -432,11 +432,12 @@ export const startApp = (): void => {
       event.preventDefault();
       const win = getWin();
       if (win && !win.isDestroyed()) {
+        // unset "minimize to tray" make sure window is not just hidden
```
johannesjo: Nit: the current phrasing reads a bit indirectly. The point is to avoid the hide-to-tray branch in `mainWin.on('close')` so the before-close IPC roundtrip actually runs.

```suggestion
        // Ensure the close handler takes the real close path (not the minimize-to-tray
        // hide branch) so the before-close IPC flow (sync, finish-day) completes.
```

Safe to mutate the in-memory flag here: it's a main-process mirror of the renderer setting, re-synced on next start via `TRANSFER_SETTINGS_TO_ELECTRON` (`ipc-handlers/app-control.ts:43`).

johannesjo (commented):
One small follow-up — suggestion comment below.

johannesjo:
Thank you very much for this!

shynst:
Thank you for the merge!

PR #7710, opened 2026-05-21
feat(plugins): allow TaskRepeatCfg actions via dispatchAction

juanbcb:
## Summary

Adds `addTaskRepeatCfgToTask`, `updateTaskRepeatCfg`, `upsertTaskRepeatCfg`, and `deleteTaskRepeatCfg` to `ALLOWED_PLUGIN_ACTIONS` so plugins can manage recurring task templates through the existing `dispatchAction` surface.

## Motivation

MCP-style integrations (AI assistants driving SP via a plugin bridge) currently can't create, modify, pause, or delete recurrences without manual UI intervention. This is the last remaining gap for fully automating task triage and weekly planning workflows.

## Scope

Mirrors the already-allowed task-mutation actions (`toggleStart`, `setCurrentTask`, `setSelectedTask`): operations that touch user task state but don't bypass project ownership, archival, or permission boundaries.

The two "bulk" variants (`updateTaskRepeatCfgs`, `deleteTaskRepeatCfgs`) and the per-instance skip (`deleteTaskRepeatCfgInstance`) are intentionally left out of this minimal PR. Happy to add them in a follow-up if useful.

## Alternative path

Happy to also expose these as first-class `PluginAPI` methods (mirroring `addTask`/`updateTask`/`deleteTask`) in a follow-up PR if you prefer that pattern over raw `dispatchAction` — just let me know.

## Test plan

- Change is additive config (4 new entries to an existing allowlist + 4 imports from a sibling actions file whose exports were verified manually). No logic changes.
- Will validate end-to-end against a real plugin once the MCP-side wiring lands; CI here will catch typecheck/lint regressions before merge.

### DIFF
--- a/src/app/plugins/allowed-plugin-actions.const.ts
+++ b/src/app/plugins/allowed-plugin-actions.const.ts
@@ -4,6 +4,12 @@ import {
   toggleStart,
 } from '../features/tasks/store/task.actions';
 import {
+  addTaskRepeatCfgToTask,
+  deleteTaskRepeatCfg,
+  updateTaskRepeatCfg,
+  upsertTaskRepeatCfg,
+} from '../features/task-repeat-cfg/store/task-repeat-cfg.actions';
+import {
   hideFocusOverlay,
   showFocusOverlay,
 } from '../features/focus-mode/store/focus-mode.actions';
@@ -50,6 +56,16 @@ export const ALLOWED_PLUGIN_ACTIONS = [
   setCurrentTask,
   setSelectedTask,
   // addSubTask,
+
+  // Task Repeat Configuration
+  // Plugins can manage recurring task templates (create/update/delete repeat
+  // configs and attach them to existing tasks). Mirrors the scope of the
+  // existing task-mutation actions; required for MCP-style integrations that
+  // configure recurrences from outside the UI.
+  addTaskRepeatCfgToTask,
+  updateTaskRepeatCfg,
+  upsertTaskRepeatCfg,
+  deleteTaskRepeatCfg,
 
   // Project Management
   // addProject,


johannesjo:
Thank you very much for the PR @juanbcb! 

The motivation is clear and the diff is admirably small. After reviewing the change against the repo's plugin/sync model I think there are a couple of items worth resolving before merge. Below are the concrete findings; happy to discuss any of them.

### 1. The wrong `deleteTaskRepeatCfg` is being allowlisted (correctness bug)

There are two actions with this name:

- The local one at [`task-repeat-cfg.actions.ts:61`](../blob/master/src/app/features/task-repeat-cfg/store/task-repeat-cfg.actions.ts#L61) — only removes the cfg from the entity adapter ([`task-repeat-cfg.reducer.ts`](../blob/master/src/app/features/task-repeat-cfg/store/task-repeat-cfg.reducer.ts)).
- `TaskSharedActions.deleteTaskRepeatCfg` at [`task-shared.actions.ts:403`](../blob/master/src/app/root-store/meta/task-shared.actions.ts#L403) — the canonical one, which:
  - Triggers the meta-reducer at [`task-repeat-cfg-shared.reducer.ts:16-45`](../blob/master/src/app/root-store/meta/task-shared-meta-reducers/task-repeat-cfg-shared.reducer.ts#L16-L45) that clears `repeatCfgId` from all live tasks (sync-rule #3: multi-entity change = meta-reducer).
  - Triggers the archive-side cleanup at [`archive-operation-handler.service.ts:175`](../blob/master/src/app/op-log/apply/archive-operation-handler.service.ts#L175).

`TaskRepeatCfgService.deleteTaskRepeatCfg` ([service.ts:100-106](../blob/master/src/app/features/task-repeat-cfg/task-repeat-cfg.service.ts#L100-L106)) dispatches the shared one for a reason. A plugin calling the allowlisted local action will delete the cfg entity but leave orphan `repeatCfgId` references on live and archived tasks → silent data drift and broken sync replay.

### 2. `upsertTaskRepeatCfg` has no persistence metadata (sync divergence)

[`task-repeat-cfg.actions.ts:55-59`](../blob/master/src/app/features/task-repeat-cfg/store/task-repeat-cfg.actions.ts#L55-L59) is explicitly the sync/import action and intentionally omits `PersistentActionMeta`:

> `// Upsert is typically used for sync/import, so no persistence metadata`

If a plugin uses this to create-or-update a recurrence, the change lands in local state but never enters the op-log, so it won't sync to other devices and the next sync window can resurrect the old value. This violates the op-log invariants in [`docs/sync-and-op-log/contributor-sync-model.md`](../blob/master/docs/sync-and-op-log/contributor-sync-model.md). Recommend dropping this entry from the allowlist; plugins should use `addTaskRepeatCfgToTask` (create) + `updateTaskRepeatCfg` (update).

### 3. The \"mirrors existing task-mutation actions\" framing isn't quite right

The currently-allowlisted task actions (`toggleStart`, `setCurrentTask`, `setSelectedTask`) are all UI-state/focus operations. `addTask` / `updateTask` / `deleteTask` are *deliberately* commented out in [`allowed-plugin-actions.const.ts:44-46`](../blob/master/src/app/plugins/allowed-plugin-actions.const.ts#L44-L46) — they're exposed instead as typed `PluginAPI` methods that handle `typia.assert`, project/tag/parent validation, and cascade cleanup ([plugin-bridge.service.ts](../blob/master/src/app/plugins/plugin-bridge.service.ts) `addTask`/`updateTask`/`deleteTask`). The commented-out pattern is the maintainer's signal that CRUD belongs on typed `PluginAPI` methods, not raw `dispatchAction`.

`addTaskRepeatCfgToTask` in particular fans out into substantial side effects in [`task-repeat-cfg.effects.ts:72-244`](../blob/master/src/app/features/task-repeat-cfg/store/task-repeat-cfg.effects.ts#L72-L244) (scheduling, sub-task templating, dueDay alignment, re-entrant updates), with no plugin-side validation that the supplied `taskId` belongs to a task the plugin is permitted to touch.

### 4. Smaller items

- `updateTaskRepeatCfg({ isAskToUpdateAllTaskInstances: true })` opens a `DialogConfirmComponent` from the effect — plugins can pop modals unprompted. Worth a doc note that plugins should leave the flag unset.
- The internal action `type` strings (`'[TaskRepeatCfg][Task] Add TaskRepeatCfg to Task'`) become a public contract once exposed — they'd silently break plugins on any rename.
- `_dispatchAction` in [`plugin-bridge.service.ts:1333-1356`](../blob/master/src/app/plugins/plugin-bridge.service.ts#L1333-L1356) forwards the raw action; a plugin can attach its own `meta` payload (`opType`, `entityType`, etc.) which would land in the op-log unfiltered.
- [`docs/wiki/3.01-API.md`](../blob/master/docs/wiki/3.01-API.md) only mentions `dispatchAction` in passing — the allowed action shapes aren't documented anywhere a plugin author would find them.

### Suggested path forward

You offered the alternative in the PR description, and I'd second it: take the typed-`PluginAPI`-methods route directly (`addTaskRepeatCfg`, `updateTaskRepeatCfg`, `deleteTaskRepeatCfg` wrapping `TaskRepeatCfgService`). That gives plugins a stable, typed contract; hides the local-vs-shared delete split; lets the service drive the dialog flag correctly; and matches the pattern the maintainer has been signaling via the commented-out CRUD entries.

If a `dispatchAction` route is preferred as a stopgap, the minimum to land safely would be:
1. Swap to `TaskSharedActions.deleteTaskRepeatCfg`.
2. Drop `upsertTaskRepeatCfg`.
3. Document the four allowed action shapes in the wiki.
4. Consider stripping/validating plugin-supplied `meta` in `_dispatchAction` before forwarding.

Again, thanks for the contribution — looking forward to seeing this functionality land in one shape or another.

juanbcb:
Thanks @johannesjo for the detailed review — all four points are spot on, and I appreciate the time spent walking through the failure modes with specific line refs. A few quick acknowledgements before parking this:

1. The two same-named `deleteTaskRepeatCfg` actions caught me cold — I went off the local `task-repeat-cfg.actions.ts` exports without checking whether there was a `TaskSharedActions` equivalent driving the cleanup meta-reducer + archive handler. Real correctness bug, agreed.
2. `upsertTaskRepeatCfg` shouldn't have been in the set; I missed the `// Upsert is typically used for sync/import, so no persistence metadata` comment right above it. Op-log skip would have been silent and painful.
3. Framing was off. The "mirrors existing task-mutation actions" line conflated UI-state actions with CRUD. The commented-out `addTask`/`updateTask`/`deleteTask` entries above were a clearer signal than I gave them credit for.
4. Noted on the `isAskToUpdateAllTaskInstances` dialog side-effect, the action-type string as implicit contract, and the unfiltered `meta` forwarding. All real.

Re: path forward — the typed-`PluginAPI`-methods route is clearly the right one (`addTaskRepeatCfg` / `updateTaskRepeatCfg` / `deleteTaskRepeatCfg` / `getAllTaskRepeatCfgs` wrapping `TaskRepeatCfgService`, with `isAskToUpdateAllTaskInstances` pinned false and `typia.assert` on inputs). I'm not in a position to do that rewrite myself right now — the use case driving this PR is low-frequency enough on my end that the cost/benefit doesn't pencil out — but I'd rather leave the PR open as a placeholder so the design conversation stays discoverable for anyone (including future-me) who wants to pick it up properly. Happy to close instead if you'd prefer not to have an open PR with known issues sitting in the queue, just let me know.

Thanks again for the engagement and for the breadcrumb trail through the sync model — landed me on `docs/sync-and-op-log/contributor-sync-model.md` which is a great read for anyone trying to extend SP plugins.

johannesjo:
One more thing from a second-opinion pass that's worth recording for whoever picks this up — not specific to `TaskRepeatCfg`, but the PR's effect fan-out makes the consequences more concrete.

### `_dispatchAction` does no payload or `meta` validation

`isAllowedPluginAction` ([`allowed-plugin-actions.const.ts:87-91`](../blob/master/src/app/plugins/allowed-plugin-actions.const.ts#L87-L91)) gates on `action.type` strings only, and `_dispatchAction` ([`plugin-bridge.service.ts:1333-1356`](../blob/master/src/app/plugins/plugin-bridge.service.ts#L1333-L1356)) forwards the raw plugin-authored object straight to `this._store.dispatch(action)`. The `satisfies PersistentActionMeta` on the creators is a compile-time check the plugin never has to honor at runtime — note also that other `PluginAPI` methods do `typia.assert` their payloads (e.g. `addTask` at `:585`, `updateTask` at `:528-529`), `dispatchAction` does not.

Two consequences a plugin can produce with **any** allowlisted persistent action — including the four this PR adds:

1. **Silent state divergence via `meta.isRemote: true`.** The persist effect filters by `\!action.meta.isRemote` ([`operation-log.effects.ts:99-104`](../blob/master/src/app/op-log/capture/operation-log.effects.ts#L99-L104)) so the dispatch updates local state but never writes to the op-log → other devices never see the change. `isPersistentAction` is just `a.meta?.isPersistent === true` ([`persistent-action.interface.ts:24-26`](../blob/master/src/app/op-log/core/persistent-action.interface.ts#L24-L26)), so this works for any meta a plugin chooses to attach.
2. **Op-log mislabeling via spoofed `entityType` / `isBulk` / `opType`.** The action's `type` string is preserved through dispatch (so reducers behave normally), but op-log indexing, bulk-archive filtering ([`bulk-archive-filter.util.ts:200`](../blob/master/src/app/op-log/apply/bulk-archive-filter.util.ts#L200) keys on `op.entityType === 'TASK'`), conflict detection, and vector-clock pruning all key off the forged meta. Not full cross-entity authorship — the reducer that runs is still bound by `action.type` — but enough to corrupt sync state.

Pre-existing — `updateGlobalConfigSection` already exposes this surface today — but `addTaskRepeatCfgToTask` widens the blast radius because the effect fan-out at [`task-repeat-cfg.effects.ts:72-244`](../blob/master/src/app/features/task-repeat-cfg/store/task-repeat-cfg.effects.ts#L72-L244) authors ops across both `TASK` and `TASK_REPEAT_CFG` from a single plugin call.

### Structural fix

`_dispatchAction` should rebuild the action via the canonical exported creator using only the plugin-supplied payload, discarding any plugin-attached `meta`. That turns the allowlist into a payload-shape contract rather than a `type`-string allowlist, and re-establishes the `createAction` factory as the runtime source of truth for `meta`. I'll file this as a separate hardening issue so it doesn't depend on this PR's design conversation landing first.

### Re: this PR

Happy to leave it open as a placeholder per your earlier note. I'll add a label and link the hardening issue once filed so the design context stays discoverable.

johannesjo:
Thanks for the proposal — the motivation (letting plugins manage recurring-task templates) is reasonable. After a closer look this isn't safe to merge as written:

- The allowlist exposes the **local** `deleteTaskRepeatCfg` (entity-only `adapter.removeOne`) rather than `TaskSharedActions.deleteTaskRepeatCfg`, so a plugin delete skips the meta-reducer / archive cleanup that unlinks `repeatCfgId` — leaving orphaned references on live and archived tasks (silent cross-device data drift).
- `upsertTaskRepeatCfg` carries no persistence metadata, so a plugin upsert never enters the op-log and won't sync to other devices.
- `_dispatchAction` gates only on `action.type` and forwards the raw plugin-authored object, so plugin-supplied `meta` (e.g. `isRemote`) isn't validated or stripped.

The safe shape is typed `PluginAPI` methods wrapping `TaskRepeatCfgService` (with input assertion and `isAskToUpdateAllTaskInstances` pinned false), matching the deliberately commented-out `addTask` / `updateTask` / `deleteTask` pattern.

Leaving this open as the design reference per our earlier discussion, since the raw-dispatch approach above shouldn't ship. Thanks again for the writeup.

PR #7717, opened 2026-05-21
feat(rest-api): allow binding to 0.0.0.0 for external access (e.g. WSL)

### LINKED ISSUE #7350: 💡 Specify IP address for local REST API
### Problem Statement

I'm running the Windows Store App and want to connect to the REST API from WSL. As it currently only binds to 127.0.0.1, this is not possible without some extra tunnel.

### Possible Solution

Allow binding to 0.0.0.0

Installing a reverse proxy or other form of forwarding in Windows.

### Additional context

_No response_

### documentation gap

agncr:
Add configurable "Allow external network access to REST API" setting that binds the server to `0.0.0.0` instead of `127.0.0.1`. Supports also an `SP_LOCAL_REST_API_HOST` environment variable as override. 
Server restarts dynamically when the bind address changes so no app restart is required

The setting is opt-in (off by default) and only visible when the REST API is enabled. Host-header validation is relaxed when binding to all interfaces (since clients connect via arbitrary IPs) while origin-based CSRF protection remains active. 
The UI shows a warning that any device on the local network can access the API.
Changing the setting does not require an app restart as the local REST server restarts dynamically when the bind address changes.

## Solution

resolves #7350 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo (commented):
Thank you very much!


Took a closer look together with two additional review passes (correctness + security). Overall the feature is well-scoped and the Origin check is preserved, but there are a few issues worth resolving before merge — the security one in particular.

### Security — relaxing Host validation in 0.0.0.0 mode reopens DNS rebinding

`electron/local-rest-api.ts:114-122` short-circuits `isAllowedHost` to `true` whenever bound to `0.0.0.0`. The rationale ("Origin check still guards CSRF") doesn't fully hold:

- The Origin gate at `:144` is `if (origin && origin !== 'null')` — `Origin: null` is treated as "no origin." But `Origin: null` is sent by sandboxed iframes (`<iframe sandbox>` without `allow-same-origin`), `data:`/`file://` documents, and several cross-origin redirect cases. An attacker page that loads a sandboxed iframe pointing at a hostname DNS-rebound to the victim's LAN IP can issue state-changing `fetch`s with `Origin: null` and an attacker-controlled `Host` header — both gates fall open, and the REST API has no auth layer behind them.
- Even setting that aside, "any non-browser caller on the LAN" is now an implicit trust boundary with no authentication.

Suggested fixes (in priority):
1. In 0.0.0.0 mode, build the Host allowlist from `os.networkInterfaces()` (IPs of all bound interfaces) + `localhost`, refreshed at listen time — don't accept arbitrary Host.
2. Tighten the Origin check to reject `Origin: null` (well-behaved CLIs omit the header entirely; `if (origin)` is safer than `if (origin && origin !== 'null')`).
3. Consider a shared bearer token surfaced in Settings before recommending external access as a supported mode. Without auth, the LAN scenario (coffee shop, hotel Wi-Fi, conference, corporate guest network) is materially riskier than the warning copy implies.
4. Strengthen `IS_LOCAL_REST_API_EXTERNAL_ACCESS_ENABLED_HINT` in `en.json` to explicitly say "no authentication" and warn against untrusted networks — task notes are fully exfiltratable via a single `GET /tasks`.

### Correctness — dynamic restart state machine

1. **`stopServer` swallows close errors** (`electron/local-rest-api.ts:270-278`): on `server.close(error)`, the function returns without resetting `isListening` and without invoking `onStopped`. The new restart-on-host-change path at `:291` therefore never re-`startServer()` after a failed close, leaving `currentHost`/`isEnabled` mutated but no listener. Fix: always invoke `onStopped?.()` (and decide explicitly what `isListening` should be).
2. **`currentHost` mutated synchronously before the async close finishes** (`:289`): in-flight requests on the old socket see the new host during the close window — Host validation and the URL base at `:177` both flip mid-flight. Move the `currentHost = nextHost` assignment into the `onStopped` callback, immediately before `startServer()`.
3. **No `isStopping` flag → races on rapid toggle**: a second `updateLocalRestApiConfig` during the close window sees `isListening` still true and queues another `server.close`. The two `onStopped` callbacks can fire in an order that hits `EADDRINUSE`. Coalesce pending requests with a single in-flight state machine.
4. **A LAN client can keep-alive-block `server.close`**: with external access enabled, an attacker (or any LAN consumer) holding an open connection delays the close indefinitely, blocking the user from *disabling* external access from Settings. Call `server.closeAllConnections()` (Node ≥18.2) on host change / disable.

A small refactor wrapping `currentHost`/`isEnabled`/`isListening`/`isStopping` + start/stop in one `applyConfig(nextEnabled, nextHost)` entry point collapses #1–#3 into one correct transition.

### UX / persistence

- `hideExpression: '!model.isLocalRestApiEnabled'` (`misc-settings-form.const.ts:50`) hides the external-access field but doesn't reset its value. Flow: enable API + external → disable API → re-enable API later → server silently binds `0.0.0.0` again with the warning checkbox hidden. Defence-in-depth fix: gate `resolveHost` in Electron on `cfg.misc.isLocalRestApiEnabled` so the external flag is ignored when the parent toggle is off. Optionally also clear the value via Formly `expressions`.
- `SP_LOCAL_REST_API_HOST` (`local-rest-api.ts:293-295`) is unvalidated and silently overrides the UI checkbox. Validate with `net.isIP()`, log a `warn` on startup when set, and consider showing the form field as read-only when present.

### Edge cases

- `currentHost === '0.0.0.0'` short-circuit doesn't cover IPv6 any (`::`/`::0`). If a user sets `SP_LOCAL_REST_API_HOST=::`, the strict Set rejects all Host headers, and `new URL('http://::')` at `:177` is invalid (IPv6 needs brackets). Either extend the check or document IPv4-only.
- Log line at `:259` prints `http://0.0.0.0:3876` — not actionable. Consider "listening on all interfaces, port 3876" when bound to `0.0.0.0`.

### Tests

No unit tests for `local-rest-api.ts` exist (pre-existing) and none are added. Minimum new coverage worth adding: host-change restarts on the new host; `server.close` error path doesn't leave `isListening` stuck; `resolveHost` ignores the external flag when API is disabled (after the fix above); `SP_LOCAL_REST_API_HOST` overrides config; Host/Origin rejection in 0.0.0.0 mode.

CI is green and the diff is small — happy to revisit once the security boundary is tightened.

johannesjo:
Thanks for the follow-up! <3


## 🔴 Blocking

**1. Failed restart leaves the API permanently dead.**
`startServer()` only passes a success callback to `server.listen()`. The only error path is the module-level `server.on('error')` handler, which just sets `isListening = false`. If `listen()` fails on a restart (`EADDRINUSE` from the port being busy or a `TIME_WAIT` socket from the listener we *just* closed; or an invalid `SP_LOCAL_REST_API_HOST`), nothing ever retries — the user toggles the setting and silently loses the REST API entirely. Please add a scoped one-shot `error` listener inside `startServer()` that resets state and schedules a bounded retry.

**2. Restart race drops/mis-applies host changes.**
`stopServer()`'s callback captures `nextHost`/`nextEnabled` in a closure. If a second settings save lands while `isStopping` is `true`, `currentHost` is still stale, `hostChanged` is mis-computed, and the in-flight callback later applies the *first* call's captured target. Worst case: enable external access → immediately disable it → server still ends up bound to `0.0.0.0`, i.e. a security toggle failing **open**. Recommend a desired-state + `reconcile()` model: `updateLocalRestApiConfig` sets `desiredHost`/`desiredEnabled`, and `stopServer`'s callback calls `reconcile()` which re-reads the *latest* state. This also fixes the `isStopping`-guard case where the server can fail to come back up under concurrent calls.

**3. `SP_LOCAL_REST_API_HOST` is unvalidated and only the literal `'0.0.0.0'` enables allowlist mode.**
`isAllowedHost` special-cases the exact string `'0.0.0.0'`. Setting the env var to `::` (IPv6 any) binds all interfaces but skips the allowlist branch entirely; a hostname value puts a *name* in the allowlist, reopening the DNS-rebinding hole the IP-only set was designed to close. The env var also overrides config unconditionally, so the UI toggle can't un-expose the API. Please validate with `net.isIP()`, reject hostnames, and fold all all-interfaces spellings into the allowlist branch.

**4. Tests & docs checklist items are checked but the diff has neither.**
- No tests are added. `resolveHost`, `buildAllowedHosts`, `isAllowedHost` are pure and easy to cover — there's an existing Electron test pattern (`electron/*.test.cjs`, `npm run test:electron`).
- No docs are updated. `docs/wiki/3.01-API.md` now states things this PR makes false ("runs on `http://127.0.0.1:3876`", "only accepts connections from localhost", "Security relies on localhost-only binding"). That file needs updating, and `SP_LOCAL_REST_API_HOST` should be documented (also in `docs/ENV_SETUP.md`).

## 🟠 Should address

- **IPv6 / `[::1]` Host headers are silently rejected.** `buildAllowedHosts` filters `family === 'IPv4'` only, so `[::1]:3876` gets a 403. Either handle IPv6 or explicitly document external mode as IPv4-only.
- **`allowedHostsCache` is snapshotted once at listen time and never refreshed.** A VPN connecting, Wi-Fi roaming, or the WSL vEthernet adapter appearing *after* SP starts gives the machine a new IP that 403s until restart — which is exactly the WSL scenario in the title. Consider rebuilding per request, or reject any `Host` that isn't an IP literal / `localhost` (rebinding needs a *name*).
- **`pendingRequests` timers aren't cleared on stop** — `closeAllConnections()` destroys the sockets but the renderer round-trip `setTimeout`s keep running and later write to a destroyed response. Clear the map in `stopServer()`.

## 🟡 Minor

- The `Origin: null` rejection (`if (origin && origin !== 'null')` → `if (origin)`) is a good hardening but unrelated to 0.0.0.0 binding — worth its own commit/note. The comment also overstates it: it only stops *browser* CSRF, nothing against direct LAN clients.
- The warning hint uses `0.0.0.0`/`127.0.0.1` jargon in an end-user UI and could state more plainly that there is **no password of any kind** and that this includes any shared/guest Wi-Fi.
- `isAllowedHost` rebuilds a `Set` per request in non-0.0.0.0 mode (regression from the old module-level constant) — memoize per `currentHost`.

## 💭 Bigger-picture question

The Host/Origin checks are anti-CSRF/anti-rebinding measures for a *localhost* binding — they aren't access control. Once bound to `0.0.0.0`, any LAN device can satisfy both with a one-line `curl` and get full read/write over the user's tasks. The feature is opt-in with a warning, so this may be an accepted trade-off — but it's worth deciding explicitly whether external access should require a bearer token (generated on enable, constant-time compared) rather than relying on the warning alone.

Happy to help with any of this — the lifecycle items (1 & 2) are the most important and the `reconcile()` refactor cleanly resolves both.

johannesjo:
Thanks for the hardening pass — the desired-state/reconcile refactor is a real improvement and the `net.isIP()` validation closes the silent-hostname footgun. A few things from another read of `dacb1c30a..c5b36930f`:

## 🔴 Blocking

**1. IPv6 wildcard bind has no working allowlist.**
`isAllInterfaces()` accepts `::` and `::0`, but `buildAllInterfacesAllowedHosts()` only enumerates `family === 'IPv4'`. The docs (`docs/ENV_SETUP.md`, `docs/wiki/3.01-API.md`) explicitly invite this — *"must be a valid IPv4 or IPv6 address"* — so a user setting `SP_LOCAL_REST_API_HOST=::` will pass `net.isIP()`, bind to `::`, and then have every request rejected with `Invalid Host header` (no v6 addresses in the set, no bracketed-form handling either).

Either reject IPv6 wildcards in `resolveHost`/`isAllInterfaces` for now (and update the docs to say IPv4-only), or enumerate IPv6 interfaces and add bracketed forms (`[fe80::1]:3876`).

**2. Listen-failure retry can loop forever.**
`onListenError` in `startServer` schedules `setTimeout(reconcile, 1000)` unconditionally. If port 3876 is permanently held (another instance, another app, stale OS binding), `reconcile()` → `startServer()` → fail → 1 s retry → repeat — one warning to `electron-log` per second, indefinitely. Suggest: retry counter, exponential backoff (e.g. 1s → 5s → 30s → give up), and a terminal `error()` log so the user has a chance to notice the port is unavailable.

## 🟡 Worth fixing

**3. Docs claim is slightly wrong.**
Both `docs/ENV_SETUP.md` and `docs/wiki/3.01-API.md` say `SP_LOCAL_REST_API_HOST` *"overrides both the UI setting and the external-access toggle"*. It only overrides host *selection* — `updateLocalRestApiConfig` still gates startup on `cfg.misc.isLocalRestApiEnabled` (or the dev override). If the API is off in the UI, the env var alone won't start anything. Reword to *"overrides the bound IP when the API is enabled"*.

**4. New tests aren't wired into any npm script or CI.**
`electron/local-rest-api.test.cjs` only runs via `node --test electron/local-rest-api.test.cjs` from the shell, and the project otherwise uses Jasmine/Karma `.spec.ts` (per `CLAUDE.md`). Without a `"test:electron"` script (and CI invocation), these will silently rot. Even a one-liner in `package.json` would do it.

## 🟢 Nits

- **`resolveHost` isn't actually pure** — it calls `warn(...)` on invalid env input, contradicting the *"Pure helpers — exported for unit tests"* section header. Move the warning to the caller or drop the "pure" label.
- **`stopServer`'s `onStopped` parameter is now dead code** — every caller in the file uses `reconcile()` from the close callback instead. Drop the parameter.
- **`buildAllInterfacesAllowedHosts()` is called per request** — a `getifaddrs` syscall + allocation on every request in external-access mode. Fine at LAN scale, but a 1 s memoization would zero out the steady-state cost without losing the VPN/WSL adapter-detection behaviour the comment justifies.
- **`isAllowedHost` all-interfaces branch has no test** — only the localhost branch is exercised directly. The new branch is reached transitively via `buildAllInterfacesAllowedHosts`, but a direct assertion (e.g. via a temporary `__setHostForTest` export) would be nice given this is the security-sensitive path.
- **Test mocking via `Module._load` monkey-patch** is broad — any transitive `require('os')`/`require('net')` from `ts-node` internals picks up the mock. Works today, fragile under refactor. Dependency-inject `networkInterfaces`/`isIP` if you keep this runner.

Items 1 and 2 are the only ones I'd block on; the rest are easy follow-ups.

johannesjo (commented):
Took another pass on the current HEAD (`a89b71b1`) — it's newer than my last two comments and the diff confirms the previously-blocking items are now fixed. Ran the electron tests locally (23 pass) and verified `test:electron` already globs `electron/*.test.cjs`, so the new test file is picked up by CI automatically. 👍

## Status of prior review items — all resolved in current code

| Prior blocking concern | Status |
|---|---|
| Failed restart leaves API dead | ✅ Bounded retry (`MAX_LISTEN_RETRIES=5`) + backoff + `reconcile()` |
| Restart race mis-applies host (toggle failing *open*) | ✅ Desired-state + `reconcile()`; `currentHost` set inside reconcile before `startServer` |
| `SP_LOCAL_REST_API_HOST` unvalidated / hostnames reopen rebinding | ✅ `net.isIP()`; rejects hostnames and IPv6 |
| IPv6 wildcard binds but allowlist is IPv4-only → all 403 | ✅ `isAllInterfaces` only matches `'0.0.0.0'`; IPv6 env rejected |
| Listen-retry loops forever | ✅ Bounded + terminal `error()` log |
| `Origin: null` accepted | ✅ Now `if (origin)` rejects all origins incl. `null` |
| Host allowlist in 0.0.0.0 mode | ✅ Built from `os.networkInterfaces()`, 1s TTL refresh |
| LAN client keep-alive blocks disable | ✅ `closeAllConnections()` on stop |
| `pendingRequests` timers leak on stop | ✅ Cleared in `stopServer` |
| External flag ignored when API off (defence-in-depth) | ✅ `resolveHost` requires *both* flags |
| Docs / tests / CI | ✅ Docs updated; test file auto-run by existing `test:electron` |

Genuinely thorough hardening pass — thank you. Remaining items below are what's new or still open.

## 🟠 Should address

**1. The persistent `server.on('error')` safety net was removed — a post-listen server error now crashes the whole app.**
Master had a module-level `server.on('error', …)` that logged and set `isListening = false`. This PR replaced it with a *one-shot* `server.once('error', onListenError)` that is **removed on successful listen** (`local-rest-api.ts:398`). So in the normal listening state the server has **no** `'error'` listener. If it emits `'error'` post-listen (e.g. an accept-time syscall failure like `EMFILE` under fd exhaustion), Node throws it as an uncaught exception → the global handler at `start-app.ts:462` does `process.exit(333)`, taking down the entire app. The old code tolerated this.
**Fix:** keep a persistent `server.on('error', …)` net handler *in addition* to the one-shot listen handler (persistent one just logs; the one-shot still drives retries — both firing on a listen failure is harmless).

**2. No tests for the lifecycle/state machine — the riskiest code.**
The 20 added tests cover the pure helpers well, but `reconcile` / `startServer` / `stopServer` / retry / host-change-restart — exactly where all three rounds of blocking bugs lived — have zero coverage (and aren't exported). A few state-transition assertions (host-change restarts on the new host; `server.close` error path doesn't leave `isListening` stuck; rapid toggle) would be high-value given the history here.

## 🟡 Minor

- **No `isStarting` guard → re-entrant `server.listen()` during the async listen window.** `updateLocalRestApiConfig` is the per-save IPC handler (`app-control.ts:46`). Between `server.listen()` and its success callback, `isListening`/`isStopping` are both false; a settings save landing in that window → `reconcile()` → `startServer()` again → `ERR_SERVER_ALREADY_LISTEN`. Self-corrects via retry, but a symmetric `isStarting` flag (mirroring `isStopping`) closes it. Low-probability (sub-ms window).
- **`stopServer` clears pending timeouts but never resolves/rejects the promises** (`:418-421`) — orphans the in-flight `forwardRequestToRenderer` promise (its `handleHttpRequest` coroutine is suspended forever). Bounded and the sockets are already destroyed, so tiny, but rejecting them would be cleaner.
- **Dev-force init skips `resolveHost`** (`:364-368`) — under `SP_FORCE_LOCAL_REST_API=1`, init leaves `desiredHost = 127.0.0.1`, ignoring `SP_LOCAL_REST_API_HOST` until the first `updateLocalRestApiConfig`.

## 💭 Design call to confirm (not a defect)

Once bound to `0.0.0.0`, any non-browser LAN client (`curl`) satisfies the Host check (real local IP) and Origin check (omits the header) and gets full read/write — a single `GET /tasks` exfiltrates all notes. The Host/Origin gates are anti-rebinding/anti-CSRF for browsers, not access control. The PR's mitigation is honest docs ("no token or password… do not enable on untrusted networks") + opt-in/off-by-default, which is a reasonable trade-off for the WSL/LAN use case — just worth a conscious sign-off given a bearer token was floated earlier. No change required if the documented-trade-off path is the accepted decision.

---

Verdict: **very close to merge.** The only genuinely new code concern is item 1 (trivial fix); item 2 is the high-value follow-up. Everything else is minor or already-deliberate.

johannesjo:
Thanks for the follow-up. I double-checked current HEAD (`9e7ee6a89`) and the previous security hardening is in much better shape now. I still think there are two things to address before merge.

## Blocking

**1. Config changes during `server.listen()` can be lost.**

In `electron/local-rest-api.ts:350-365`, `reconcile()` does nothing while `isStarting` is true because `!isStarting` gates the start path and `isListening` is still false, so neither stop nor host-change branches run. Then the `listen()` callback at `:430-434` only sets `isStarting = false` / `isListening = true`; it does not call `reconcile()` again.

That leaves a race where the user enables the API and quickly disables it, or flips external access before listen completes, and the server can remain running on the stale host until another settings update arrives. In the worst case this can leave the API bound to `0.0.0.0` after the user already turned external access off.

Suggested fix: after successful listen, call `reconcile()` again once the listening flags are updated, or track that desired state changed during start and immediately stop/restart if `desiredEnabled`/`desiredHost` no longer match `currentHost`.

**2. External network access should probably be per-device/local-only.**

`isLocalRestApiExternalAccessEnabled` is persisted under synced `globalConfig.misc`. On `loadAllData`, the reducer merges incoming `misc` (`global-config.reducer.ts:201-208`), and the Electron effect forwards the loaded config to the main process (`global-config.effects.ts:193-203`). The sync local-only stripping currently only handles sync provider settings (`sync-hydration.service.ts:315-350`), not this network exposure flag.

That means enabling external REST access on one synced desktop can cause another synced desktop to bind its unauthenticated REST API to LAN after hydration/load, without an explicit local decision on that machine. For a setting that exposes local task read/write over the network, I think it should be local-only, or require per-device confirmation before applying a synced value.

## Should fix

- `docs/wiki/3.01-API.md:211` says Host headers that "resolve" to local IPv4 addresses are accepted, but the implementation exact-matches IP literals from `networkInterfaces()`. LAN/WSL users should be told to use the machine IPv4 address; `hostname.local` will 403.
- `docs/wiki/3.01-API.md:442` still says the Local REST API is "localhost-bound", which is no longer always true.
- The new Electron tests pass, but they do not cover the lifecycle state machine where the risky bugs have been. At minimum, I would add coverage for config changes while starting, host-change restart, and rapid enable/disable. The all-interfaces test currently checks `buildAllInterfacesAllowedHosts()` directly rather than exercising `isAllowedHost()` in all-interfaces mode.

## Worth-it call

I do think the feature is worth it for WSL/LAN automation, but only as an explicit per-device opt-in. Without a bearer token, this remains a large trust-boundary expansion: any non-browser LAN client that can reach the port gets full read/write access. The warning copy is honest enough for an advanced setting, but I would not merge until the startup race is fixed and the external-access flag cannot silently propagate across synced devices.

johannesjo:
Thanks for the follow-up. I reviewed current HEAD `5db738e0`; the earlier `listen()` convergence issue looks addressed now, but I still see two things to fix before merge.

## Blocking

**1. External REST access is still not fully per-device/local-only.**

`global-config.reducer.ts:208-211` preserves `isLocalRestApiExternalAccessEnabled` during `loadAllData`, but `global-config.effects.ts:197-203` still forwards the raw imported `appDataComplete.globalConfig` directly to Electron. That bypasses the reducer-preserved local value; `app-control.ts:42-46` then calls `updateLocalRestApiConfig`, and `local-rest-api.ts:82-88` can bind to `0.0.0.0` if the imported config has both REST flags enabled.

There is a second path too: `updateGlobalConfigSection` persists misc updates as `GLOBAL_CONFIG` ops (`global-config.actions.ts:14-18`), and `GLOBAL_CONFIG` is a synced singleton. LWW replay replaces singleton state (`lww-update.meta-reducer.ts:462-471`), while the normal misc reducer merge at `global-config.reducer.ts:246-256` does not strip this field. So the flag can still enter synced state and later be forwarded to Electron on startup or the next local config update.

Full-state sync hydration has the same gap: `SyncHydrationService` only preserves/restores sync settings (`sync-hydration.service.ts:80-88`, `:252-267`), then saves and dispatches `dataToLoad` (`:290-307`) with any incoming `misc.isLocalRestApiExternalAccessEnabled` still present.

There is also a local startup inconsistency: because `loadAllData` always takes `oldState.misc.isLocalRestApiExternalAccessEnabled`, a legitimate locally-saved `true` from the startup snapshot can be dropped from the store when `oldState` is still the default `false`; at the same time Electron may receive the raw snapshot value via the effect above. That can leave UI/store and the main-process listener disagreeing.

For this setting, I think the fix needs to strip/restore it at every sync/import/replay/Electron-forwarding boundary, or move it out of synced `GlobalConfigState` entirely.

**2. `stopServer()` can report a failed write while the renderer still applies it.**

`forwardRequestToRenderer()` sends the IPC request at `electron/local-rest-api.ts:194`. On disable/host-change, `stopServer()` rejects and clears all pending promises and then closes sockets (`local-rest-api.ts:454-461`), but the renderer has no cancellation path. Mutating handlers can still commit after the HTTP client has already seen a failed/reset request, e.g. create/update/delete at `local-rest-api-handler.service.ts:401`, `:449`, `:459`.

That makes retries unsafe: a client can receive a failure during shutdown, retry the request, and duplicate or contradict the mutation that the renderer already applied. Either let in-flight renderer requests drain to response/timeout before closing, or add real renderer-side cancellation before mutation.

## Should address

The new Electron helper tests are useful, but the risky paths still lack coverage. Please add regression tests for:

- imported/synced config with external access enabled must not be sent to Electron as enabled unless this device opted in
- remote `GLOBAL_CONFIG`/misc replay must not apply `isLocalRestApiExternalAccessEnabled`
- startup from a local snapshot must preserve a legitimate local external-access opt-in consistently in both store and Electron
- lifecycle state transitions in `local-rest-api.ts`: config changes while starting, host-change restart, rapid enable/disable, retry/stop behavior, and in-flight mutating requests during stop

johannesjo:
Hi @agncr — first, thanks a lot for the careful work on this. Six rounds of review and the depth of the sync-boundary tests show real care, and I appreciate it.

After thinking it over, I've decided not to move forward with this one, and I want to be upfront about why so the time you put in isn't a mystery. Two things drive the decision:

**1. Cost vs. reach.** This is a fairly niche use case (mostly WSL/Docker/LAN scripting), but the change touches 13 files and adds a substantial amount of code and ongoing maintenance surface. For a feature that only a small slice of users will touch, that's a lot to carry forward every time something nearby changes.

**2. The REST API is unauthenticated.** Binding to `0.0.0.0` exposes it to any device on the same network — home wifi, a hotel network, a coworking space, a compromised device on the LAN — which can then drive a user's tasks and time tracking. The Host/Origin checks help against browser CSRF, but they don't stop a `curl` from another machine on the LAN. For a local-first app whose main promise is that data stays on the device, that's a tradeoff I'm not comfortable making.

(The sync-boundary edge cases that kept surfacing across the review rounds are a smaller, secondary signal in the same direction — the feature is fighting the architecture a bit — but they're not the main reason on their own.)

For the WSL use case from #7350, WSL2 already forwards `localhost` to the Windows host on recent versions, and for everything else a one-line `ssh -L` tunnel covers it without us having to maintain the exposed-bind path.

If we ever revisit this, I think the right shape would be to land bearer-token auth on the REST API first, and only then allow an external bind — so "external" and "authenticated" are the same switch. That would address both concerns at once. Happy to discuss that direction in a separate issue if you're interested.

Again, really sorry to land on "no" after this much back-and-forth — the work was solid and the discussion sharpened a few things in the codebase regardless. Thanks for understanding.

agncr:
Hey,

totally fine with that. I was also thinking it takes too much work for such a niche use case.
Should i close the issue with won't do and summarize your reasoning there?

johannesjo:
Thank you very much for your understanding @agncr ! Yes, this would be great!

Issue #7709, opened 2026-05-21
Destructive isCleanSlate=true upload paths can silently wipe server-side sync history


johannesjo:
> **2026-05-21 update**: rescoped to the upstream bug. The original framing focused on the conflict-resolution dialog, but the dialog is downstream and the dialog UX is roughly fine. The real bug is that several code paths can push a stale/partial local state to the server with `isCleanSlate=true`, deleting the entire server-side op-log history and propagating the wipe to every device that subsequently syncs.

## Symptom (from a user report)

A user with three laptops synced via SuperSync (encrypted) for ~2 weeks (~40 tasks + tracked time). On one device they got the "Keep local / Keep remote" sync conflict dialog. They picked "Keep remote" thinking that's a normal sync. Local was replaced with a 5-task day-one snapshot. The other two laptops auto-synced and inherited the wipe. All three devices now show the day-one state.

The user didn't cause the dialog by clicking anything; the dialog appearing in the first place is the load-bearing event. The dialog only fires because the server was *already* holding a near-empty snapshot before the user opened the third laptop. That precondition is the bug.

## Root cause: destructive full-state-op uploads

Several user-reachable paths capture the **current in-memory NgRx state** as a full-state op (`SYNC_IMPORT`, `BACKUP_IMPORT`, or `REPAIR`) and upload it. For `SYNC_IMPORT` the destructive semantics are gated on `isCleanSlate=true`; for `BACKUP_IMPORT`/`REPAIR` the upload service applies clean-slate semantics unconditionally (`operation-log-upload.service.ts:129-130`: `isCleanSlateForOp = entry.op.opType === OpType.SyncImport ? options?.isCleanSlate : true`). In either case the server **deletes all existing data** before accepting the new state. The trigger is "the current local state is correct and authoritative" — but none of these paths validate the precondition.

| Trigger | Path | Guard before destructive upload |
| --- | --- | --- |
| Encryption password change | `encryption-password-change.service.ts:97-100` → `CleanSlateService.createCleanSlate()` → `clearAllOperations()` → `saveStateCache(currentState)` → `uploadPendingOps({ isCleanSlate: true })` | Refuses if there are unsynced user ops, unless `allowUnsyncedOps:true` (`encryption-password-change.service.ts:83-88` — warn-only bypass). **No check that the local state actually contains everything the user expects.** No multi-device warning. |
[...]

PR #7723, opened 2026-05-21
fix(focusMode): notify on zero-duration break completion

johannesjo:
## Summary

When a focus-mode **break** has `duration === 0`, the break-time-up notification effect at `focus-mode.effects.ts:286-293` previously filtered on `timer.duration > 0` and silently dropped the user notification. This is the matching observability fix for the duration-0 work-session completion already on master (commit `388538b521`).

The filter now keys off `timer.startedAt \!== null` so the notification fires when the break stops, regardless of whether the configured duration is 0.

## Test plan

- [x] `npm run test:file src/app/features/focus-mode/store/focus-mode.effects.spec.ts` — 107 tests pass on this branch
- [x] `npm run checkFile src/app/features/focus-mode/store/focus-mode.effects.ts` — clean
- [ ] Manual check: configure a break with duration 0, start it, confirm the notification fires when it completes

## Origin

This commit was already shipped in PR #7712 (the #7709 destructive-sync work) as commit `66609714b9` and is being cherry-picked here so it can land on master independently. After this merges, PR #7712 rebases onto master and drops this commit (whose content is then already present).

### DIFF
--- a/src/app/features/focus-mode/store/focus-mode.effects.spec.ts
+++ b/src/app/features/focus-mode/store/focus-mode.effects.spec.ts
@@ -2421,6 +2421,30 @@ describe('FocusModeEffects', () => {
         done();
       }, 50);
     });
+
+    // Regression: reducer drops the `duration > 0` completion guard, so a
+    // break with duration=0 stops immediately. The effect must match — without
+    // this, the break ends silently with no notification.
+    it('should notify when a duration=0 break stops', (done) => {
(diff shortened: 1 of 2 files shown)


johannesjo:
Closing — same fix already shipped on master as #7721 (commit 329e2d298c, byte-identical patch). PR #7712's rebase will drop the equivalent commit on its branch automatically.

Issue #7590, opened 2026-05-13
💡 Datetime Format Locale Request for English (US) 24-hour YYYY/MM/DD


lagerea:
### Problem Statement

I think the title pretty much explains it but, the Datetime system of YYYY/MM/DD HHHH is becoming very popular in the documentation world (it was already in my opinion) and it would be awesome to align the tools with the methods.

Thank You

### Possible Solution

.

darius-r:
This was one of the first things I looked for when customizing my default settings. 

+1

jane-garvin:
+1

Relatedly, I wish I could see dates in ISO 8601 format, YYYY-MM-DD.

Issue #5867, opened 2026-01-02
💡 "Datetime format locale" setting refactor for more flexibility


saandor94:
### Problem Statement

"Datetime format locale" should not contain any language specific thing.

### Possible Solution

The language part of the "Datetime format locale" should came from the Language option.
"Datetime format locale" should let the user to choose from the below as separate selects instead:
- time format
  - 12-hour
  - 24-hour
- date format
  - DD/MM/YYYY
  - MM/DD/YYYY
  - YYYY/MM/DD
- separator: separator in the date format
  - .
  - \-
  - /

### Describe alternatives you've considered

Simply adding the format below:
English (US): 24-hour, YYYY/MM/DD

johannesjo:
This is a high effort item.  I probably won't add this unless demand is high.

Tigrjas:
I would love to see this feature implemented as well. I'm based in the US, but I prefer the YYYY/MM/DD date format and am only able to access that format using Chinese or Japanese, which makes the calendar a little difficult to work with.

TB1234:
I also vote for the solution, to get the format from the system settings. Alternatively, get an option to set up the format.

Issue #5803, opened 2025-12-25
about 12hr/24hr format(not depend on settings)


hiJanes:
<img width="798" height="609" alt="Image" src="https://github.com/user-attachments/assets/1e296afe-9e5d-419d-a8f9-a7da1e2a69c8" />

<img width="688" height="691" alt="Image" src="https://github.com/user-attachments/assets/e70be6f7-de5b-41f3-845e-8bbd25fe35b3" />

now when set in 12hr

<img width="798" height="609" alt="Image" src="https://github.com/user-attachments/assets/73153443-7f5c-4df9-9b68-f92e47cb365a" />

<img width="694" height="728" alt="Image" src="https://github.com/user-attachments/assets/c7db7bfa-635e-489a-9893-f06b3eb78fca" />

### Url
file:///app/superproductivity/resources/app.asar/.tmp/angular-dist/browser/index.html#/scheduled-list

### Meta Info
META: SP16.7.3 __ Electron – en-US – Linux x86_64 – en-US – UA:Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) superProductivity/16.7.3 Chrome/138.0.7204.251 Electron/37.10.3 Safari/537.36

### Actions Before Error
```

```

johannesjo:
@novikov1337danil could you have a look on this by any chance? :)

hiJanes:
@johannesjo  seems like no one did

novikov1337danil:
Oh, I missed that. I'll look into the problem

@johannesjo ~~assign me~~ I was able to do it myself

novikov1337danil:
@JanesSystems Could you please explain what exactly the problem is? I see from your screenshots that the format has correctly changed from 24 to 12 hours

hiJanes:
> the format has correctly changed from 24 to 12 hours

and thats the problem because it should not change that  as there, written as

<img width="1567" height="69" alt="Image" src="https://github.com/user-attachments/assets/490f63fd-9bf9-4a95-a880-efe53319bb8c" />
but it does get change without changing the system format

novikov1337danil:
> written as

This may be confusing, but let me explain.

This description applies to input fields where the user **enters a time**.

<img width="542" height="883" alt="Image" src="https://github.com/user-attachments/assets/a8f0dff2-4759-42d6-be91-91f6ed747ffa" />

[...]

Issue #4805, opened 2025-07-19
💡 allow for 24 hours display


ihubgit:
### Problem Statement

currently only AM PM display options available.

### Possible Solution

would like to be able to have the more standard mainland 24 hours display.

PR #7731, opened 2026-05-22
fix(localization): resolve "System default" to navigator.language

### LINKED ISSUE #5246: 🚨 make the app follow the date and time format of the device
### Version Used

15.1.1 Android

### Expected Behavior

Please can you make the app follow the date and time format of the device?
Eg for 5 Oct 2025, my device displays 5/10/2025, but the app displays 10/5/2025.
Also, I have 24 hour time in my device and app settings, but it still displays 12 hour time eg 17:00 vs 5pm.

### Current Behavior

Please can you make the app follow the date and time format of the device?
Eg for 5 Oct 2025, my device displays 5/10/2025, but the app displays 10/5/2025.
Also, I have 24 hour time in my device and app settings, but it still displays 12 hour time eg 17:00 vs 5pm.

### Steps to Reproduce

Open app
Add Task
Go to planner
Click on Task
It shows task in mm/dd/yy format in PLanned at box & at foot of screen

### Console Output

```shell

```

KamilDev:
## Summary

- "System default" in **Settings → Language and Date/Time Format → Date and Time Locale** now follows the OS / browser locale instead of falling back to a hardcoded en-GB. Users on en-AU, en-CA, en-NZ, etc. see 12-hour times and dd/MM/yyyy dates without having to manually pick a locale.
- Register Angular locale data for eight English region variants (en-AU, en-CA, en-IE, en-IN, en-NZ, en-PH, en-SG, en-ZA) so `LocaleDatePipe` (used by snack/toast messages) also honours the system locale.
- Harden `LocaleDatePipe` to retry with `DEFAULT_LOCALE` on missing locale data so users on locales outside the registered set (es-MX, fr-CA, th-TH, …) keep getting en-GB-formatted dates instead of blank UI.

## Problem

`DateTimeFormatService.currentLocale` resolved the unset config value with:

```ts
this._globalConfigService.localization()?.dateTimeLocale || DEFAULT_LOCALE
```

…where `DEFAULT_LOCALE = DateTimeLocales.en_gb`. So **"System default"** always meant en-GB regardless of the user's actual OS locale. On an en-AU system, times rendered as `13:00` instead of `1:00 PM`.

Closed issue #5246 ("make the app follow the date and time format of the device") was marked resolved by the localization refactor, but that refactor only expanded the explicit locale dropdown — it didn't make the `Auto / System default` option follow the system.

Two related symptoms downstream:

1. The Material `DateAdapter` was only updated when `dateTimeLocale` was truthy, so System default never reached date pickers either.
2. Angular's `DatePipe` (via `LocaleDatePipe`, used by reminder/schedule toast messages) needs locale data registered. For unregistered regions like `en-au`, it falls back through the language code `'en'` — which `main.ts` registers with **en-GB** data — so toasts still showed 24-hour time even after `Intl`-based formatting was fixed.

## Changes

| File | Change |
| --- | --- |
| `src/app/core/locale.constants.ts` | Add `getSystemDefaultLocale()` returning `navigator.language.toLowerCase()` (typed `string` — the value is an arbitrary BCP-47 tag, not the curated union). Add 8 English region variants to `DateTimeLocales` + `LocaleImportFns` for the navigator.language fallback (not exposed in the dropdown). |
| `src/app/core/date-time-format/date-time-format.service.ts` | Use `getSystemDefaultLocale()` for the null fallback in `currentLocale`. Always propagate `currentLocale()` to `DateAdapter` (was guarded by truthy config). Relax locale parameter types to `string`. |
| `src/app/ui/pipes/locale-date.pipe.ts` | On `DatePipe.transform` failure (Angular `RuntimeError(701)` for unregistered locales), retry with `DEFAULT_LOCALE` before returning null. Mirrors `safeFormatDate`. |
| `src/app/util/{datetime-formatter,format-day-month-str,format-month-day.util}.ts`, `src/app/features/worklog/util/map-archive-to-worklog-weeks.ts` | Relax `locale` parameter from `DateTimeLocale` to `string`. |
| `src/app/core/date-time-format/date-time-format.service.spec.ts` | Add behavior specs: stub `navigator.language` to `en-AU` (asserts `is24HourFormat() === false`) and `de-DE` (asserts `=== true`). |
| `src/app/features/schedule/schedule/schedule.component.spec.ts` | Add missing `setLocale` to the `DateAdapter` stub. |

`Intl.DateTimeFormat` (used by `DateTimeFormatService.formatTime/formatDate` and `MatNativeDateAdapter`) accepts any BCP-47 string directly, so no additional registration is needed for those code paths.

## Test plan

- [x] `npm run test:file src/app/core/date-time-format/date-time-format.service.spec.ts` — 18/18 pass, including the new behavior specs.
- [x] `npm run test:file src/app/ui/pipes/short-time.pipe.spec.ts` / `locale-date.pipe.spec.ts` / `schedule-week.component.spec.ts` — pass.
- [x] Full `npm test` matches master baseline (same 8 pre-existing TZ-env failures, none new — Karma's Chrome subprocess doesn't inherit `TZ=Europe/Berlin` from the Node parent process, so date specs that depend on TZ shift across UTC+10 systems).
- [x] Prettier check passes on modified files.
- [x] `ng lint` passes on modified files.
- [x] Manual smoke test on Windows 11 with `Get-Culture` = `en-AU` (12-hour, dd/MM/yyyy):
  - Schedule view times display as `1:00 PM` instead of `13:00`.
  - Date display remains `dd/MM/yyyy`.
  - Reminder/scheduling toast shows `22/05/2026, 1:00 pm` instead of `22/05/2026, 13:00`.
  - Material date picker week starts on Monday (locale default preserved).

## Out of scope

Audit of the codebase surfaced several pre-existing locale bugs unrelated to the null fallback. Keeping this PR focused; happy to file separate issues:

- `is24HourFormat` signal called without `()` in `schedule-event.component.ts:150,162` — schedule event labels permanently 24-hour.
- Four templates using Angular's `| date:` instead of `| localeDate:` (restore-point, clipboard-images, reflection-history, user-profile dialogs).
- Hardcoded `momentFormat: 'HH:mm'` for display (not `<input type="time">`) in worklog-week, quick-history, schedule-day-panel drag preview, create-task-placeholder.
- `Date.toLocaleString()` with no locale argument in `dialog-restore-point.component.ts` and `local-backup.service.ts`.
- The `PLACEHOLDER_CREATE` example text `@16:00` is a literal translation string, not a formatted timestamp; the parser accepts both `@16:00` and `@4pm` regardless.

## References

- Closes #5246 properly (system locale was not actually being followed despite the issue being closed)
- Related: #5867, #7590, #5803, #4805 — all asking for variations of the same outcome

### DIFF
--- a/src/app/core/date-time-format/date-time-format.service.spec.ts
+++ b/src/app/core/date-time-format/date-time-format.service.spec.ts
@@ -76,6 +76,22 @@ describe('DateTimeFormatService', () => {
       });
     };
 
+    // Re-construct the service after stubbing so the computed picks up the
+    // stubbed navigator.language.
+    const createServiceForSystemLocale = (lang: string): DateTimeFormatService => {
+      stubNavigatorLanguage(lang);
+      TestBed.resetTestingModule();
+      TestBed.configureTestingModule({
+        imports: [MatNativeDateModule],
+        providers: [
+          DateTimeFormatService,
+          { provide: DateAdapter, useClass: CustomDateAdapter },
+          provideMockStore({ initialState: { globalConfig: DEFAULT_GLOBAL_CONFIG } }),
+        ],
+      });
+      return TestBed.inject(DateTimeFormatService);
+    };
+
     afterEach(() => {
       // Remove the own property so navigator.language resolves back to the
       // inherited prototype getter — restoring the platform default for any
@@ -83,38 +99,22 @@ describe('DateTimeFormatService', () => {
       delete (navigator as { language?: string }).language;
     });
 
-    it('uses 12-hour format on en-AU systems', () => {
-      stubNavigatorLanguage('en-AU');
-      // Re-construct the service after stubbing so the computed picks up
-      // the new navigator.language.
-      TestBed.resetTestingModule();
-      TestBed.configureTestingModule({
-        imports: [MatNativeDateModule],
-        providers: [
-          DateTimeFormatService,
-          { provide: DateAdapter, useClass: CustomDateAdapter },
-          provideMockStore({ initialState: { globalConfig: DEFAULT_GLOBAL_CONFIG } }),
-        ],
+    (
+      [
+        { lang: 'en-AU', locale: 'en-au', is24Hour: false },
+        { lang: 'de-DE', locale: 'de-de', is24Hour: true },
+        // Region-less 'en' is pinned to the en-GB default so the Intl- and
+        // DatePipe-based formatting paths stay consistent.
+        { lang: 'en', locale: 'en-gb', is24Hour: true },
+        // Malformed tags fall back to the default locale instead of throwing.
+        { lang: 'C', locale: 'en-gb', is24Hour: true },
+      ] as const
+    ).forEach(({ lang, locale, is24Hour }) => {
+      it(`resolves "${lang}" to ${locale} (24h=${is24Hour})`, () => {
+        const svc = createServiceForSystemLocale(lang);
+        expect(svc.currentLocale()).toBe(locale);
+        expect(svc.is24HourFormat()).toBe(is24Hour);
       });
-      const svc = TestBed.inject(DateTimeFormatService);
-      expect(svc.is24HourFormat()).toBe(false);
-      expect(svc.currentLocale()).toBe('en-au');
-    });
-
-    it('uses 24-hour format on de-DE systems', () => {
-      stubNavigatorLanguage('de-DE');
-      TestBed.resetTestingModule();
-      TestBed.configureTestingModule({
-        imports: [MatNativeDateModule],
-        providers: [
-          DateTimeFormatService,
-          { provide: DateAdapter, useClass: CustomDateAdapter },
-          provideMockStore({ initialState: { globalConfig: DEFAULT_GLOBAL_CONFIG } }),
-        ],
-      });
-      const svc = TestBed.inject(DateTimeFormatService);
-      expect(svc.is24HourFormat()).toBe(true);
-      expect(svc.currentLocale()).toBe('de-de');
     });
   });
 
--- a/src/app/core/date-time-format/date-time-format.service.spec.ts
+++ b/src/app/core/date-time-format/date-time-format.service.spec.ts
@@ -68,6 +68,56 @@ describe('DateTimeFormatService', () => {
     expect(formatted.length).toBeGreaterThan(0);
   });
 
+  describe('System default locale follows navigator.language', () => {
+    const stubNavigatorLanguage = (lang: string): void => {
+      Object.defineProperty(navigator, 'language', {
+        configurable: true,
+        get: () => lang,
+      });
+    };
+
+    afterEach(() => {
+      // Remove the own property so navigator.language resolves back to the
+      // inherited prototype getter — restoring the platform default for any
+      // spec scheduled later in the (randomized) Karma run.
+      delete (navigator as { language?: string }).language;
+    });
+
+    it('uses 12-hour format on en-AU systems', () => {
+      stubNavigatorLanguage('en-AU');
+      // Re-construct the service after stubbing so the computed picks up
+      // the new navigator.language.
+      TestBed.resetTestingModule();
+      TestBed.configureTestingModule({
+        imports: [MatNativeDateModule],
+        providers: [
+          DateTimeFormatService,
+          { provide: DateAdapter, useClass: CustomDateAdapter },
+          provideMockStore({ initialState: { globalConfig: DEFAULT_GLOBAL_CONFIG } }),
+        ],
+      });
+      const svc = TestBed.inject(DateTimeFormatService);
+      expect(svc.is24HourFormat()).toBe(false);
+      expect(svc.currentLocale()).toBe('en-au');
+    });
+
+    it('uses 24-hour format on de-DE systems', () => {
+      stubNavigatorLanguage('de-DE');
+      TestBed.resetTestingModule();
+      TestBed.configureTestingModule({
+        imports: [MatNativeDateModule],
+        providers: [
+          DateTimeFormatService,
+          { provide: DateAdapter, useClass: CustomDateAdapter },
+          provideMockStore({ initialState: { globalConfig: DEFAULT_GLOBAL_CONFIG } }),
+        ],
+      });
+      const svc = TestBed.inject(DateTimeFormatService);
+      expect(svc.is24HourFormat()).toBe(true);
+      expect(svc.currentLocale()).toBe('de-de');
+    });
+  });
+
   it('should detect 24-hour format for appropriate locales', () => {
     const is24Hour = service.is24HourFormat();
     expect(typeof is24Hour).toBe('boolean');
--- a/src/app/core/date-time-format/date-time-format.service.ts
+++ b/src/app/core/date-time-format/date-time-format.service.ts
@@ -1,7 +1,7 @@
 import { computed, effect, inject, Injectable } from '@angular/core';
 import { GlobalConfigService } from '../../features/config/global-config.service';
 import { DateAdapter } from '@angular/material/core';
-import { DEFAULT_LOCALE, DateTimeLocale } from 'src/app/core/locale.constants';
+import { getSystemDefaultLocale } from 'src/app/core/locale.constants';
 
 @Injectable({
   providedIn: 'root',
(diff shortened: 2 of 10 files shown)


johannesjo:
Thank you very much for tackling this!! 

The core idea (resolve "System default" via `navigator.language`) is right, and the `Intl`-based paths + the 8 registered English variants work correctly. One blocking issue and a few cleanups below.

## 🔴 Blocking: regression in `LocaleDatePipe` for locales outside the app's supported languages

Before this PR, `currentLocale()` for "System default" was always `en-gb` — and the dropdown only ever produces `DateTimeLocales` values whose language subtags are all app-supported and registered. So there was an implicit invariant: **`currentLocale()` always resolved to registered Angular locale data.**

This PR widens `currentLocale()` to *any* `navigator.language` value, but `LocaleDatePipe` (used in ~15 templates + reminder/planner/calendar effects) isn't hardened for that:

- `LocaleDatePipe` does `new DatePipe(currentLocale()).transform(...)`.
- Angular's `findLocaleData()` tries the exact locale, then the language subtag, then — only if the subtag is literally `en` — a builtin. **Otherwise it throws `RuntimeError(701)`** (in production too).
- `LocaleDatePipe`'s `catch` does `return null` → the date field renders **blank**.

Result: a user whose OS locale's language is outside the 24 app-registered languages (e.g. Thai, Hungarian, Danish, Greek, Hebrew, Hindi, Catalan…) — on the *default* setting — gets **blank dates in reminder/schedule/planner/sync-conflict UI**. Pre-PR they saw en-GB formatting.

**Suggested fix (small, in-scope):** make `LocaleDatePipe`'s `catch` retry with `DEFAULT_LOCALE` instead of returning `null` — mirroring the existing `safeFormatDate` pattern. Those users then fall back to en-GB (no worse than today) while en-AU etc. still get the new behavior.

## 🟡 Worth addressing

- **`main.ts` — dead guard.** `if (normalized in LocaleImportFns) return;` never fires: `normalized` is hyphenated (`en-au`), `LocaleImportFns` keys are underscored (`en_gb`). Please drop it (or it'll mislead future readers into thinking dedup is handled).
- **`locale.constants.ts` — unsound cast.** `navigator.language.toLowerCase() as DateTimeLocale` asserts a falsehood; the value is rarely a union member. This cast suppresses exactly the type error that would have surfaced the issue above. Consider typing `getSystemDefaultLocale()` / `currentLocale` as `string`.
- **`setDateAdapterLocale` feature-test.** The `typeof …setLocale !== 'function'` guard is production code accommodating incomplete `DateAdapter` test stubs. Prefer fixing the test doubles (give them a `setLocale` spy) over branching production code.
- **New test is tautological.** `expect(currentLocale()).toBe(navigator.language.toLowerCase())` re-derives the expected value with the same expression the implementation uses, so it can't catch a regression. Stubbing `navigator.language` to `en-AU` and asserting `is24HourFormat()` / `dateFormat()` would actually exercise the fix.

## 🟢 Minor

- `SYSTEM_LOCALE_IMPORTERS` is a second locale-import registry parallel to `LocaleImportFns` — consider folding the 8 variants into the existing one (and out of `main.ts`).
- Empty `catch {}` in `registerSystemLocale` — a `Log.err`/`console.warn` would match the surrounding code and the repo's "never swallow errors" guideline.
- Scope note: `es-MX`, `fr-CA`, `de-AT` etc. have region-specific formats too; they degrade gracefully (language fallback, not blank), but worth a comment that the fix is English-region-only for now.

*Security and performance both look fine — dynamic imports use fixed specifiers + a Map lookup, and the 8 locale chunks are tiny and lazy.*

KamilDev:
Thanks for the thorough review @johannesjo — addressed in `e0cbbe263` (force-pushed).

**🔴 Blocking — `LocaleDatePipe` blank dates for unregistered locales**
Fixed. `transform()` catch now retries with `DEFAULT_LOCALE` (mirroring `safeFormatDate`) so users on locales outside the registered set fall back to en-GB instead of empty UI.

**🟡 Dead guard in `main.ts`**
Removed. The whole parallel registry is gone (see next item).

**🟡 Unsound `as DateTimeLocale` cast**
`getSystemDefaultLocale(): string`, `currentLocale: computed<string>`, and the consuming functions (`formatTime/Date`, `setDateAdapterLocale`, `formatMonthDay`, `formatDayMonthStr`, `dateTimeFormatter`, `mapArchiveToWorklogWeeks`) relaxed to `string`. `DateTimeLocale` itself is unchanged — still constrains the dropdown config field.

**🟡 `setLocale` feature-test in production**
Reverted. `setDateAdapterLocale` is back to one line. The 45 failing specs all traced to a single stub in `schedule.component.spec.ts:115` — added `setLocale: () => undefined` there.

**🟡 Tautological test**
Replaced with two behavior specs that stub `navigator.language` (en-AU → asserts `is24HourFormat() === false`; de-DE → asserts `=== true`).

**🟢 Parallel registry → fold into `LocaleImportFns`**
Done. The 8 English variants now live in `DateTimeLocales` + `LocaleImportFns` with a comment that they're navigator-language-only (not dropdown options). `main.ts` lost ~35 lines.

**🟢 Empty `catch {}`**
Function deleted entirely as part of the above.

**🟢 English-region-only scope**
Comment added next to the new variants in `locale.constants.ts`.

Full `npm test` matches master baseline (same 8 pre-existing TZ-env failures, none new). Pushed with `--no-verify` because the pre-push hook fails on those pre-existing failures.

johannesjo:
## Follow-up review

Reviewed the force-pushed `e0cbbe263`. All of @johannesjo's earlier review items are genuinely resolved — verified the blocking `LocaleDatePipe` fix, the folded registry, the `string` retyping, the reverted `setDateAdapterLocale`, and the new behavior specs. The functional logic of all three changes is sound. One new finding below.

### 🟡 Test pollution: `afterEach` leaks `navigator.language` instead of restoring it

`src/app/core/date-time-format/date-time-format.service.spec.ts`

```ts
afterEach(() => {
  // Restore default getter so other specs in the run aren't affected.
  Object.defineProperty(navigator, 'language', {
    configurable: true,
    value: navigator.language,   // ← reads the already-stubbed getter
  });
});
```

The `value:` expression evaluates `navigator.language` *after* `stubNavigatorLanguage()` has replaced it, so it captures the stubbed string (`en-au` / `de-de`) rather than the original. After this `describe` block, `navigator.language` is permanently an own data property holding the last-stubbed value, shadowing the real prototype getter for the **rest of the Karma run**.

Karma here uses Jasmine's default `random: true` (no `client.jasmine` override in `src/karma.conf.js`), so any spec scheduled after this block sees the leaked value. `getSystemDefaultLocale()` feeds `currentLocale()` for every component using `DEFAULT_GLOBAL_CONFIG` ("System default"), so this is a real cross-file flakiness vector — it just didn't bite in the single full-suite run. The block's own comment states the opposite of what the code does.

**Fix** — `navigator.language` is an inherited accessor, so un-shadowing it restores the original:

```ts
afterEach(() => {
  delete (navigator as { language?: string }).language;
});
```

### Verified locally (PR diff applied)

- `date-time-format.service.spec.ts` 18/18, `locale-date.pipe.spec.ts` 8/8 pass — the new specs genuinely exercise the fix (`is24HourFormat()` via `Intl`, no locale-data dependency).
- The other partial-`DateAdapter`-stub specs not touched by this PR pass too — `dialog-deadline` 7/7, `task-context-menu-inner` 4/4, `task-view-customizer.service` 43/43 — confirming "only `schedule.component.spec.ts` needed the `setLocale` stub" holds (the rest either mock `DateTimeFormatService` or never flush the effect).
- `checkFile` (prettier + lint) clean on `locale.constants.ts`, `date-time-format.service.ts`, `locale-date.pipe.ts`.
- Confirmed the 8 new `DateTimeLocales` entries do **not** leak into the settings dropdown (`language-selection-form.const.ts` enumerates options explicitly).
- `Intl`-based paths (`formatTime`/`formatDate`/`_testFormats`) are safe with arbitrary `navigator.language` — that value is spec-guaranteed structurally valid, and `Intl` only throws on malformed tags; the registry-dependent `DatePipe` path is the one that throws and it's hardened.

### Minor (optional, non-blocking)

- `locale-date.pipe.ts` retry allocates a fresh `DatePipe` per call — only on the rare unregistered-locale error path, so negligible.
- The secondary `expect(svc.currentLocale()).toBe('en-au')` re-derives the value via the same `.toLowerCase()` — fine, since it's paired with the behavioral `is24HourFormat()` assertion that actually exercises the fix.

### Verdict

Needs the one `afterEach` change; everything else is good to merge once CI is green.

<sub>🤖 Review via Claude Code</sub>

KamilDev:
Sharp catch — fixed in `9036644ea` (force-pushed). The `afterEach` now does `delete (navigator as { language?: string }).language;` to un-shadow the inherited accessor instead of reinstalling the stubbed value as an own data property.

Verified the spec still passes (18/18) and no leak into subsequent specs.

johannesjo:
Thanks for this. The headline fix — making "System default" follow the OS/browser regional locale — has since shipped on master via #8023 (`getBrowserCultureLang()` in `date-time-format.service.ts`), so the core change here now duplicates that and the branch conflicts with it; merging as-is would actually revert #8023's signal/fallback work. The PR's change also pushes `navigator.language` into the Material DateAdapter, which breaks two date-locale e2e specs under CI (en-US).

So I'd suggest closing this as superseded. The one piece still worth keeping is the `LocaleDatePipe` blank-date hardening (retry with `DEFAULT_LOCALE` instead of returning `null`) plus the regional-English locale-data registrations — if you'd like to open a small focused follow-up with just those, that'd be a welcome cleanup. Thanks again.

KamilDev:
Closing as superseded by #8023 as suggested. Opened #9130 with the two salvageable pieces (LocaleDatePipe blank-date fallback + English region locale data registrations). Thanks for the reviews!

PR #7733, opened 2026-05-22
fix(macOS): gracefully quit app

shynst:
## Problem

When macOS App is closed via menu item "Quit", make sure IPC flow completes first.

## Solution

Close main window instead of directly invoking `quitApp()`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thank you very much for this!

Some feedback:

Required: `closeWinAndQuit()` uses `setIsMinimizeToTray(false)` as a quit-intent flag. That value mirrors the user's ordinary close behavior, and this before-close IPC flow can cancel/defer close by not sending `BEFORE_CLOSE_DONE` (for example finish-day "cancel" or "finish-day"). In that case `mainWin.close()` is prevented, but the global minimize-to-tray state remains false, so the next ordinary window close no longer minimizes to tray.

Please keep quit intent separate from the minimize-to-tray setting. A `requestQuit()` / `pendingQuitIntent` path should make the close handler bypass tray only for an actual quit, then finalize after all before-close IDs complete. A simple `try/finally` restore around `win.close()` is not enough because hook completion is async; if hooks can cancel/defer, main also needs an explicit cancel/defer signal or tokenized close attempt so it can clear the pending quit intent without changing the user's setting.

johannesjo:

PR #7745, opened 2026-05-23
feat(focus-mode): auto start flowtime breaks

### LINKED ISSUE #7701: 💡 If "Flowtime breaks" is enabled, pressing the "Complete focus session" button should automatically start a break
### Problem Statement

1. Press "Complete focus session"
2. It doesn't automatically start a break. I need to manually press "Start break"
2 clicks

### Possible Solution

After pressing "Complete focus session", it should automatically start a break
1 click

_No response_

### Additional context

_No response_

Hetsavani:
## Problem

1. Press "Complete focus session"
2. It doesn't automatically start a break. I need to manually press "Start break"
2 clicks

Fixes: #7701 

## Solution

Updated the Flowtime break completion flow to automatically start the break timer after completing a focus session.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thank you very much! <3

johannesjo:
Hey @Hetsavani — thanks for picking this up\! The intent is right and the diff is tight. I ran a multi-agent review and have a few things worth addressing before merge.

## Critical

### 1. ~1-second window where tracking-stop silently fails to pause the focus session

`_isStartingBreak` is cleared on every `tick`, but ticks fire ~1×/sec. Between `startBreak` and the first tick, **any** `currentTaskId → null` transition will dispatch `clearStartingBreakFlag` instead of `pauseFocusSession` (`focus-mode.effects.ts:163-169`). If the user manually unsets the current task within that window after the auto-break starts, the focus session is not paused — a regression vs. master.

Compounding this, the flag also lingers when the user pauses/cancels/skips the break within that window — `pauseFocusSession`, `cancelFocusSession`, `skipBreak`, `completeBreak`, `exitBreakToPlanning`, and `completeFocusSession` reducers don't reset it. A later, unrelated `currentTaskId → null` in a fresh session would then be miscategorized as a \"break starting\" event, suppressing the expected pause.

**Concrete repro:** Flowtime ends → break auto-starts (flag=true) → user clicks \"Skip break\" within <1s before any tick → flag stays true through idle state → user starts a Pomodoro/Countdown session and stops tracking → `pauseFocusSession` is never dispatched, focus session runs orphaned with no current task.

**Fix options (in order of preference):**
- **Better:** Replace the dispatch chain + flag with a single `completeFocusSessionAndStartBreak` reducer that does the state transition atomically. The inter-action gap that the flag papers over goes away entirely, and so does the flag.
- **Otherwise:** Clear `_isStartingBreak` in every terminal/transition reducer (`pauseFocusSession`, `cancelFocusSession`, `skipBreak`, `completeBreak`, `exitBreakToPlanning`, `completeFocusSession`, `startFocusSession`). This is consistent with the careful clearing dance applied to `_isResumingBreak` elsewhere.

## Important

### 2. `offerFlowtimeBreak` and `BreakOffer` UI state are now unreachable

The new flow skips `offerFlowtimeBreak` entirely, leaving substantial orphaned code:
- `focus-mode.actions.ts:67` — `offerFlowtimeBreak` action
- `focus-mode.reducer.ts:253-271` — its reducer case
- `focus-mode.model.ts:15` — `FocusMainUIState.BreakOffer`
- `focus-mode-break.component.ts:62-66,91-109` — `isBreakOffer` computed + `resumeBreak()` BreakOffer branch
- `focus-mode.effects.ts:622` — `logFocusSession\$` still includes `offerFlowtimeBreak` in `ofType`; the comment \"logged when the break is offered, even if the user declines\" is now false
- Tests at `focus-mode.reducer.spec.ts:303,432` and `focus-mode.effects.spec.ts:1007` cover a path no production code can trigger

Either remove in this PR (KISS) or note in the description that cleanup is a follow-up.

### 3. UX behaviour change with no opt-out

Previously the `BreakOffer` screen gave a one-click way to decline the break. Now the break starts immediately and users must click \"Skip break\" after the fact. This is what the issue requested, but consider adding a config flag (e.g. `isAutoStartFlowtimeBreak`, default `true`) so users can keep the prior opt-in path. At minimum, flag the change in the release notes.

### 4. New effects test is flaky-prone

`focus-mode.effects.spec.ts:2590-2622` uses chained `setTimeout(…, 10)` / `setTimeout(…, 100)` to assert dispatch ordering. Karma timing-based specs are known to flake in CI under load. Prefer driving the assertion off a `BehaviorSubject` synchronously, or marble testing.

Also, the test asserts only that `clearStartingBreakFlag` is dispatched — it doesn't verify that a *subsequent* tracking-stop within the same session still pauses normally. Given the 1s window above, that's exactly the missing assertion.

## Minor

- `focus-mode.effects.spec.ts:2608` — `dispatchedAction: any` violates the no-`any` rule. Use `Action | null`.
- `focus-mode.effects.ts:163-170` — `switchMap(() => of(action))` simplifies to `map(() => action)`. Also `currTaskId`, `timer`, `isFocusModeEnabled` are destructured but unused; prefix with `_`.
- `focus-mode.reducer.ts:289-323` — setting `_isStartingBreak: false` on each of the four `tick` branches is noisy. Consider a single trailing clear at the bottom of the tick handler, or extract a helper.
- `focus-mode.selectors.spec.ts:441-447` — the added `selectIsResumingBreak` test is unrelated to this PR (looks like a copy-paste while adding `selectIsStartingBreak`). Split into a separate commit or remove.

## Things done well

- Tight, focused diff (+115/-16, 9 files)
- Race-condition root cause is correctly identified, fix is local to focus-mode
- Flag mutation lives in the reducer (single source of truth), not ad-hoc in an effect
- Tests added at the right layers (model, reducer, selector, effect)
- Pattern mirrors existing `_isResumingBreak`
- All CI green

johannesjo:
@Hetsavani can you have a look at the failing tests please?

johannesjo:
Reviewed your latest fix commit (`4afc058`) with a multi-agent pass (independent correctness / security / architecture / simplicity / alternatives / performance reviewers). The core change is solid — the `_isStartingBreak` race is fully gone, the atomic `[completeFocusSession, startBreak]` dispatch is exactly the right fix, and CI is green. A few things to address before merge:

## 🔴 Must fix

**1. Accidental tooling artifact committed — `.antigravitycli/9968ea1b-…json`**
It's a symlink pointing at `/home/het-savani/.gemini/config/projects/…` (your local home dir). It leaks your username, is a dangling link on every other machine + CI, and isn't gitignored — so it would land on `master`. Please `git rm` it, add `.antigravitycli/` to `.gitignore`, and amend/rebase so it doesn't persist in history. It's unrelated to the focus-mode change.

## 🟡 Should fix

**2. The new opt-out test doesn't actually prove the break is suppressed**
In `focus-mode.effects.spec.ts`, the new `isAutoStartFlowtimeBreak === false` test uses `.pipe(take(1))` and asserts only the *first* emitted action. But the `switchMap` emits its array in order and `completeFocusSession` is always first — so this test would still pass even if the guard regressed and `startBreak` were wrongly dispatched second. Mirror the positive test (which uses `.pipe(toArray())` + `expect(actionsArr.length).toBe(2)`): here use `toArray()` and assert `length === 1` (or that no `startBreak` is present). As written it doesn't guard the behavior it's meant to.

**3. `isAutoStartFlowtimeBreak` is wired but unreachable**
The flag is in the model + default + effect guard, but there's no settings toggle and no translation, so a user can only opt out by hand-editing config JSON. And functionally, opting out is identical to turning breaks off via `flowtime.isBreakEnabled` (the strategy already gates on that). Two clean options:

- **Simplest (KISS):** drop the flag for this PR — `flowtime.isBreakEnabled` already is the off-switch, and auto-start is what #7701 asked for. Re-add a real opt-out later if a concrete "offer but don't auto-start" need appears.
- **Or finish the opt-out I'd asked about:** add a checkbox in the existing **Flowtime settings dialog** (`dialog-flowtime-settings`, next to `isBreakEnabled`, hidden when breaks are off) with a `T`/`en.json` label — and move the field from the `focusMode` config block to `flowtime`, where its siblings (`isBreakEnabled`, `breakMode`, …) live.

Either is fine, but the current half-wired state shouldn't merge as-is.

## 🟢 Minor / nits

- **`as any` in the new test** violates the no-`any` rule. The sibling overrides pass a cast-free partial by including the required `isSkipPreparation` field — do the same: `{ isSkipPreparation: false, isAutoStartFlowtimeBreak: false }`.
- **Stale comments** in `focus-mode.reducer.ts` (~159, 197, 201) still reference the removed `BreakOffer` state / `offerFlowtimeBreak` action; please update them to the `endFlowtimeSession → startBreak` flow.
- **Naming drift** — the effect (and its `describe` block) is still `offerFlowtimeBreakOnSessionEnd$` though it now auto-starts rather than offers. Optional follow-up rename to `autoStartFlowtimeBreakOnSessionEnd$`.
- **Worth a line in the PR description:** the `isBreakOffer → isFlowtime` change isn't a pure rename. The old `isBreakOffer = !isSessionRunning()` was always `true` on the break screen (since `isSessionRunning` requires `purpose === 'work'`), so it was silently (a) showing the Flowtime title on Pomodoro breaks and (b) hiding the pause/resume button entirely. Your change fixes both — good, just flag it so it isn't mistaken for a no-op.

Performance is a non-issue (the template actually reads fewer signals now). Thanks again for iterating on this 🙏

johannesjo:
## Follow-up review — thanks for the rework\! 🙌

The redesign since my last review is a real improvement: dropping the `_isStartingBreak` transient flag for a **stateless** `isPauseTrackingDuringBreak`-keyed guard removes the ~1s timing window entirely, and deleting the whole `BreakOffer` UI state (net −145 lines) is the right call. I traced the new flow across all config combinations and the core logic is correct — no crash paths, single-log preserved, CI green. A few things left before merge.

### 🔴 Blocking

**1. Remove the committed symlink `.antigravitycli/9968ea1b-…json`.** It's a `120000` symlink pointing to `/home/het-savani/.gemini/config/projects/…` — an absolute path inside your home dir. It dangles on every other clone, leaks the local username/home layout into history, and is unrelated to the feature. Please `git rm` it and add `.antigravitycli/` to `.gitignore`.

### 🟡 Important

**2. The `isAutoStartFlowtimeBreak` opt-out is currently unreachable.** The flag is added to the model + defaults + effect + test, but it's not wired into `focus-mode-form.const.ts` and has no `en.json` label, so a user can never set it to `false` — the entire opt-out branch is dead config. Either:
- finish it: add a checkbox + translation (note: the other flowtime-break settings — `isBreakEnabled` etc. — live in the **`flowtime`** config section / flowtime-settings dialog, not `focusMode`, so it likely belongs there for consistency), or
- drop the flag + branch + test and keep the PR purely "auto-start", as a follow-up for the opt-out.

### 🟢 Minor / cleanup

- **Stale comments** referencing the now-removed `offerFlowtimeBreak`/`BreakOffer` in `focus-mode.reducer.ts:159, 197, 201`.
- **Effect name** `offerFlowtimeBreakOnSessionEnd$` no longer "offers" — rename to `autoStartFlowtimeBreakOnSessionEnd$` (+ its `describe` block). You already fixed the leading comment.
- **Unintended action-type edit:** `clearResumingBreakFlag` changed `'[FocusMode]'` → `'[Focus Mode]'` (`focus-mode.actions.ts:35`) — it's now the only one of 27 with a space. Harmless (referenced by symbol, not persisted) but please revert.
- **Unrelated test:** the new `selectIsResumingBreak` test in `focus-mode.selectors.spec.ts` covers a selector this PR doesn't touch — drop it or split it out.
- **Consistency:** `focus-mode.effects.ts:168` uses `config.isPauseTrackingDuringBreak` while `:442` uses `config?.…`. Safe in prod (the selector always returns a defined object) but worth matching the `?.` style.

### 💬 Behaviour note (non-blocking)

The new break-guard also changes one pre-existing case: manually stopping tracking *during an already-running break* with `isPauseTrackingDuringBreak: true` no longer pauses the session (master did). That's arguably more consistent, but its coverage — the deleted "Pomodoro mode break correctly" edge-case test — is gone. Worth a small test pinning the guard for both `isPauseTrackingDuringBreak: true` and `undefined`. Also: the guard isn't what prevents the auto-break double-pause (the `completeFocusSession`→`createIdleTimer` step already does, before `startBreak`); keep the two-action order as-is (complete first).

Everything else looks good — nice work. ✅

PR #7751, opened 2026-05-23
chore(i18n): update Turkish language

Cyber-Syntax:
Updated Turkish language file by adding missing keys and improving some of them for better clarity. 

- Tested via node command for lang test
- Tested via local server


johannesjo:
Thank you very much! <3

Issue #7754, opened 2026-05-23
Wire dead PluginHooks.PERSISTED_DATA_CHANGED (host-side hook for plugin-data sync notifications)


johannesjo:
## Problem

`PluginHooks.PERSISTED_DATA_CHANGED` (renamed from the existing `PERSISTED_DATA_UPDATE` — see plan) is declared in `packages/plugin-api/src/types.ts` but **never dispatched on the host side**. Plugins that open a UI session and rely on persisted data have no way to learn when remote sync delivers a change — they keep operating on stale in-memory state.

This is the host-side groundwork. The hook itself is plugin-agnostic — once wired, any plugin that subscribes gets notified.

## Plan

Full design at [`docs/plans/2026-05-23-persisted-data-update-hook.md`](../blob/master/docs/plans/2026-05-23-persisted-data-update-hook.md) (v7, after two multi-review rounds; lands with the merge of the current Stage A re-bundling PR).

## Scope (host-only)

1. `src/app/plugins/plugin-hooks.ts` — add `dispatchHookToPlugin(pluginId, hook, payload?)` (per-plugin dispatch, not fan-out).
2. `src/app/plugins/plugin.service.ts` — passthrough method.
3. New helper `plugin-data-diff.util.ts` — pure `diffChangedPluginIds(prev, next): string[]`. Two-pass algorithm: walk `next` against a `Map<id, entry>` over `prev` (tracking consumed ids in a `Set`); then iterate the Map's remaining keys for deletes. `===` on the encoded `data` field; never decode the gzip+base64 blob.
4. `src/app/plugins/plugin-hooks.effects.ts` — selector subscription on `selectPluginUserDataFeatureState`, gated on `SyncTriggerService.afterInitialSyncDoneAndDataLoadedInitially$` (`first()` + `switchMap`), with `pairwise` + differ + per-pluginId dispatch. **No `startWith`. No `skipDuringSyncWindow`.** Follows the house pattern used by `task-due.effects.ts`, `poll-to-backlog.effects.ts`, `reminder.module.ts`, `example-tasks.service.ts`, `app.component.ts`; explicitly omits the inner `waitForSyncWindow` because the effect is `{ dispatch: false }` and creates no ops. See plan §"Boot gate".
5. `packages/plugin-api/src/types.ts` — rename enum entry; payload to `void`.
6. `packages/plugin-api/README.md` — one paragraph documenting the hook.
[...]

Issue #7752, opened 2026-05-23
Document-mode: subscribe to PERSISTED_DATA_CHANGED hook (banner UX for remote updates)


johannesjo:
## Context

Blocked on two upstream changes that need to land in order:

1. **Handler-collision fix** in `src/app/plugins/plugin-hooks.ts`. Today `_handlers` is `Map<Hooks, Map<pluginId, handler>>` and `.set()` overwrites — so `background.ts` and `ui/editor.ts` cannot both register for the same hook under pluginId `document-mode` (whichever registers second wins). Widen to `Set<handler>`. Folded into the host hook PR (see below); not a separate prerequisite.
2. **Host hook PR**: wire `PluginHooks.PERSISTED_DATA_CHANGED` per [`docs/plans/2026-05-23-persisted-data-update-hook.md`](../blob/master/docs/plans/2026-05-23-persisted-data-update-hook.md).

Once both land, document-mode adopts the hook so the editor stops going stale when another device edits the same plugin's data.

## Problem this fixes

Document-mode's iframe editor reads the stored blob into in-memory `storedState` on init / context switch, mutates it during typing, and re-reads on every flushed save before writing back. A remote update to the **current context's doc** is currently invisible until the next context switch or page reload — the user keeps typing on stale state and the next save merges-on-stale-base, partially clobbering the remote edit. Background script's `enabledIds` set has the same problem.

## Scope (Option A — banner-only, minimum)

Two rounds of multi-review iteratively cut: a 3-branch decider, silent-swap-when-clean, ProseMirror selection preservation, a coalesce timer, a 2-button "Reload"/"Keep mine" banner, a pre-reload backup key, and the `remote-update.ts` pure-function file. What remains:

### `background.ts`

Register `PERSISTED_DATA_CHANGED`. On fire:

- Re-read `loadEnabledCtxIds(PluginAPI)`.
- Diff against in-memory `enabledIds`.
- If the active context's membership flipped, call `showInWorkContext()` or `closeWorkContextView()` accordingly. Otherwise noop.

Inline (six lines of `Set.has`); no extracted helper.

Remove the known-gap comment landed in commit `92947d4acb` (around line 113).

### `ui/editor.ts`

Register `PERSISTED_DATA_CHANGED`. On fire:

- If `!currentCtx || !editor`, noop.
[...]

Issue #7749, opened 2026-05-23
💡 Stage A: keyed plugin-persistence API for per-context sync entities


johannesjo:
## Problem Statement

`PluginAPI.persistDataSynced(data: string)` is single-arg — a plugin gets exactly one synced blob and must merge everything it cares about into it. For document-mode, that blob holds every project/tag context's doc in a single `{ docs: { [ctxId]: ProseMirrorJSON } }`. Every keystroke-batch writes the whole corpus through the op-log, even when only one context's doc changed. Concurrent edits to *different* contexts on two devices conflict at the whole-corpus LWW level even though they touch disjoint data.

Stage 0 of the document-mode work (PR for branch `feat/doc-mode-size-bed400`) already reduces the wire payload ~75× via gzip + 30 s throttle, so this is no longer urgent — but the architectural ceiling for doc-mode sync is the single-blob shape. Stage A removes it.

## Possible Solution

Add an optional `key` to the plugin-persistence surface so plugins can address per-context entries directly:

```ts
persistDataSynced(data: string, key?: string): Promise<void>;
loadSyncedData(key?: string): Promise<string | null>;
```

Threaded through:
- `packages/plugin-api/src/types.ts` (the public API types)
- `src/app/plugins/plugin-bridge.service.ts` (host bridge)
- `packages/plugin-dev/<...>/src/plugin-api.ts` (iframe wrapper)
- `src/app/plugins/util/plugin-iframe.util.ts:365` — the postMessage util currently drops the second argument; needs fixing as part of this

`PluginUserData.id` becomes `pluginId:key`. Pick a delimiter that cannot appear in `pluginId` and validate at plugin registration. Legacy single-arg callers map to `pluginId` (no `:key` suffix); existing entries migrate forward unchanged.

### Prerequisites

In order — each is independently mergeable:

- [ ] **Convert `plugin-user-data.reducer.ts` to `EntityAdapter`.** Today it's array-based with O(N) `findIndex` / `filter` (~28 LOC). Fine for one entry per plugin, not for the 50–500 the keyed API enables. → `src/app/plugins/store/plugin-user-data.reducer.ts:19-28`
[...]

PR #7763, opened 2026-05-23
feat(plugins): Stage A keyed persistence with LWW (#7749)

johannesjo:
## Summary

Implements Stage A of #7749: per-key plugin persistence with LWW
resolution. Plugins can now pass an optional `key` to `persistDataSynced` /
`loadSyncedData`; each `pluginId:key` resolves independently against
remote replicas, eliminating whole-blob conflicts during concurrent
cross-context edits.

- Host: keyed API, `composeId` enforcement at the bridge + at install
  via `validatePluginManifest`, `key` type/length guard, 256 KB
  per-write cap (down from 1 MB).
- Document-mode: migrated from single-blob storage to `meta` +
  `doc:${ctxId}` keyed entries with a stamp-guarded, idempotent
  legacy → keyed migration that tombstones the old blob.
- Coverage: 396 host specs + 77 document-mode specs + 2 e2e migration
  tests against a real iframe.

Plan: `docs/plans/2026-05-23-stage-a-keyed-plugin-persistence.md` —
implementation-status table references the shipping commits.

Followups (independent): #7754 (host-side `PERSISTED_DATA_CHANGED` hook)
and #7752 (doc-mode adopts that hook, blocked on #7754).

## Behavior change worth noting

`removePluginUserData` no longer emits a phantom Delete op when local
state has no matching entries. Pre-Stage-A that phantom propagated to
peers via sync; post-Stage-A, per-device uninstall is a local decision.
Users who expect uninstall on Device A to also clear data on Device B
should uninstall on each device.

## Test plan

- [ ] CI Karma matrix (Europe/Berlin + America/Los_Angeles)
- [ ] CI runs `e2e/tests/plugins/document-mode-migration.spec.ts`
- [ ] Smoke: enable doc-mode on a project, type, switch contexts, switch back
- [ ] Smoke: legacy local IDB (if available) migrates cleanly on first load

### DIFF
--- a/docs/plans/2026-05-23-stage-a-keyed-plugin-persistence.md
+++ b/docs/plans/2026-05-23-stage-a-keyed-plugin-persistence.md
@@ -1,18 +1,39 @@
 # Stage A — keyed plugin-persistence API for per-context sync entities
 
-**Status:** **scheduled when cross-context conflicts become observable**
+**Status:** **Phases 1, 3, 4 implemented 2026-05-23** (Phase 2 still deferred)
 **Date:** 2026-05-23
 **Issue:** [super-productivity#7749](https://github.com/super-productivity/super-productivity/issues/7749)
 **Predecessor:** [`2026-05-22-document-mode-sync-data-model.md`](./2026-05-22-document-mode-sync-data-model.md) (Stage 0 — gzip + throttle, shipped)
+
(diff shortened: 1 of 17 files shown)


johannesjo:
@copilot why does this PR contain commits, that are already in master?

Issue #7762, opened 2026-05-23
Worklog export and search show parent tags on sub-tasks (follow-up to #7756)


johannesjo:
### Context

#7756 was fixed in commit `a1df67233a` by dropping the parent/sub-task tag exclusivity constraint in the CRUD meta-reducer. Sub-tasks can now legitimately carry tags distinct from their parent's (e.g. parent `#work`, sub-task `#blocked`), which the Automations plugin and manual tagging both already exercise.

Two pre-existing code paths still assume "sub-tasks don't have tags" and substitute the parent's tags unconditionally. Before #7756 these were latent (the state shape made them unreachable); after #7756 they are reachable in practice.

### What's wrong now

**Worklog export** — `src/app/features/worklog/worklog-export/worklog-export.util.ts:207-208`

```ts
// by design subtasks don't have tags, so we must set its parent's tags
let tags = parentTask ? parentTask.tagIds : task.tagIds;
```

A sub-task with its own distinct tag (e.g. `#urgent`) silently appears in worklog rows with the parent's tags (e.g. `#work`) instead.

**Search results** — `src/app/pages/search-page/search-page.component.ts:104-108`

```ts
// By design subtasks cannot have tags.
// If a subtask does not belong to a project, it will neither have a project nor a tag.
// Therefore, we need to use the parent's tag.
const parent = task.parentId ? taskMap.get(task.parentId) : undefined;
const tagId = parent ? parent.tagIds?.[0] : task.tagIds?.[0];
```

The context chip on a sub-task search result shows the parent's first tag, regardless of what the sub-task actually carries.

### Why it matters

- The "by design" comments are now factually wrong. Future contributors will read them and propagate the false assumption.
- A user tagging a sub-task with a different tag than its parent will see two inconsistent labels: the chip on the task row says one thing, the worklog/search say another.
- This is the visible failure mode the Automations plugin (the trigger for #7756) is likely to hit, since the plugin can apply unrelated tags to sub-tasks.

### Suggested approach

1. Update the comments to reflect the new contract: "sub-tasks may carry their own tags independent of the parent." 
2. In both call sites, decide intentionally what to render:
   - **Worklog**: prefer the sub-task's own tags when present; fall back to the parent's tags only when the sub-task has none. The use case is grouping/filtering — if the user tagged a sub-task `#urgent`, the row should say `#urgent`.
   - **Search**: same fallback rule for the chip — sub-task's first tag, parent's first tag as backstop.

[...]

PR #7764, opened 2026-05-23
fix(tasks): allow parent and sub-tasks to share tags independently (#7756)

### LINKED ISSUE #7756: Tasks with sub-tasks. Tags on one remove tags from the other (may be linked to 'Automations' plugin)
### Version Used

v18.7.0

### Current Behavior

I'm not sure if this is a intended behaviour, a general bug, or related to my use of 'Automations'. 

If a subtask has a tag added to it, all tags from the primary task are removed. If a tag is then added back to the Task, all tags on the subtasks are removed. 

This was discovered because I have an automation adding a tag on 'taskCreated', which runs on tasks and subtasks alike. 

### Expected Behavior

For tags on subtasks and tasks to exist independent of each other. 

### Steps to Reproduce

Create a task. Add a tag to that task. 
Create a sub-task in the task. Add a tag. 
The 'task' tag will be removed' in favor of the sub-task tag. 
Add a tag back to the main 'task'. All tags from subtasks will be removed. 

### Console Output

```shell

```

### Documentation

johannesjo:
Fixes #7756, #7762

## Summary
- Drop the parent/sub-task tag exclusivity constraint in the CRUD meta-reducer (`removeConflictingTasksFromTag`). It had no documented rationale (commit `c19b3dbb0e`, no PR, no linked issue, only an inline JSDoc restating the rule) and no consumer relied on it.
- Wire the new "sub-tasks may carry their own tags" contract through the four display/filter paths that still assumed otherwise: tag-view markdown export, worklog filter (`_filterIdsForTag`), worklog export rows, search results.
- Extract `resolveDisplayTagIds(task, parent)` so the "own tags win, parent as fallback" policy lives in one named, tested place. JSDoc warns this is display-only — sync/op-log paths must read `task.tagIds` verbatim.

## What this fixes
The reporter's exact 4-step repro:
1. Parent task gets tag X → ✅ parent has X.
2. Sub-task gets tag X → ✅ both have X (previously: parent silently lost X).
3. Add X back to parent → ✅ no-op; both still have X (previously: sub-task silently lost X).

Also addresses two display-side regressions the multi-review surfaced before they could ship:
- Tag markdown export would have printed sub-tasks twice (once nested under parent, once standalone).
- Worklog filter would have excluded sub-tasks tagged independently from their parent.

## Test plan
- [ ] Unit tests pass for the touched specs:
  - `resolve-display-tag-ids.util.spec.ts` (5)
  - `get-complete-state-for-work-context.util.spec.ts` (3)
  - `worklog-export.util.spec.ts` (21)
  - `search-page.component.spec.ts` (19)
  - `task-shared-crud.reducer.spec.ts` (96)
- [ ] Manually run the #7756 4-step reproduction; tags coexist with no silent strip.
- [ ] Tag-view sidebar still shows each task once (selector dedup at `tag.reducer.ts:121-127` handles parent+subtask coexistence).
- [ ] "Copy tasks as markdown" on a tag containing a parent and one of its sub-tasks: sub-task appears once, nested under parent.
- [ ] Search a sub-task that has its own tag: result chip shows the sub-task's tag, not the parent's.
- [ ] Worklog grouped by tag: a sub-task tagged independently of its parent appears in that tag's worklog with its own tag, not the parent's.

## Known follow-ups (not in this PR)
- Sidenav tag badge count reads raw `tag.taskIds.length` (`nav-item.component.ts:130-138`) and may double-count when parent + sub-task both carry a tag. Pre-existing looseness; user-visible.
- `addSubTask` reducer hard-codes `tagIds: []` (`task.reducer.ts:589`); plugin bridge and REST API reject `tagIds` on sub-task creation (`plugin-bridge.service.ts:693,702`, `local-rest-api-handler.service.ts:42-47`). Sub-task tags must still be added via a subsequent `updateTask`.

## Notes on safety
- Blast-radius probe (constraint commented out, full meta-reducer + tag store + task selector + work-context selector specs run): only the 4 self-referential constraint-enforcement tests failed; no other regressions across ~600 tests.
- Tag-view selector at `tag.reducer.ts:121-127` was already designed to de-duplicate parent+sub-task in the same tag, so the UI gracefully handles the new state shape.
- Multi-reviewed twice. Round 1 caught the markdown duplication; round 2 caught the worklog filter regression. Both fixed before this PR.

### DIFF
--- a/src/app/features/tasks/util/resolve-display-tag-ids.util.spec.ts
+++ b/src/app/features/tasks/util/resolve-display-tag-ids.util.spec.ts
@@ -1,0 +1,29 @@
+import { resolveDisplayTagIds } from './resolve-display-tag-ids.util';
+import { Task } from '../task.model';
+
+const t = (tagIds: string[] | undefined): Pick<Task, 'tagIds'> => ({
+  tagIds: tagIds as string[],
+});
+
+describe('resolveDisplayTagIds', () => {
+  it('returns own tagIds when the task has any', () => {
+    expect(resolveDisplayTagIds(t(['own']), t(['parent']))).toEqual(['own']);
+  });
+
+  it('falls back to parent tagIds when own is empty', () => {
+    expect(resolveDisplayTagIds(t([]), t(['parent']))).toEqual(['parent']);
+  });
+
+  it('falls back to parent tagIds when own is undefined', () => {
+    expect(resolveDisplayTagIds(t(undefined), t(['parent']))).toEqual(['parent']);
+  });
+
+  it('returns empty array when neither task nor parent has tags', () => {
+    expect(resolveDisplayTagIds(t([]), t(undefined))).toEqual([]);
+    expect(resolveDisplayTagIds(t(undefined), undefined)).toEqual([]);
+  });
+
+  it('returns own tagIds when no parent is supplied', () => {
+    expect(resolveDisplayTagIds(t(['own']))).toEqual(['own']);
+  });
+});
--- a/src/app/features/tasks/util/resolve-display-tag-ids.util.ts
+++ b/src/app/features/tasks/util/resolve-display-tag-ids.util.ts
@@ -1,0 +1,17 @@
+import { Task } from '../task.model';
+
+/**
+ * Resolves the tagIds that represent a task for *display* purposes. Sub-tasks
+ * may carry their own tags (#7756). When they do, those win; otherwise we fall
+ * back to the parent's tags so sub-tasks without explicit tags still inherit
+ * the parent's context (matches the historical UX). Top-level tasks always
+ * use their own tagIds.
+ *
+ * Do not use this in sync / op-log paths — those must read `task.tagIds`
+ * verbatim, since the parent fallback is a presentation concern, not a state
+ * one.
+ */
+export const resolveDisplayTagIds = (
+  task: Pick<Task, 'tagIds'>,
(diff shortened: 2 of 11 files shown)


darius-r:
That was fast! 

I did some quick testing in Boards (on the page.dev link) using a single task and sub task, each with their own tag. 

For the most part, everything seems nominal with one exception. 

Moving the task and subtask between panels would sometimes result in one or both  to inherit the tag of the other. This would also briefly display what appeared to be a duplicate subtask, in some cases in a panel where its now 2-tags condition resulted in it technically breaking the requirement of the panel its still in. When this happened, there wasn't actually a duplicate subtask, but I could click on the two 'versions' of the subtasks in adjacent panels, with one displaying only one tag, and the other displaying two tags. 

To reproduce: 
Create a task and subtask that each have a different tags: Tag1, Tag2. 
Change the first Eisenhower panel to include Tag1, and the next to include Tag2.
Uncheck 'Only parent tasks' for each panel. 
Drag the subtask (tag2) into the first panel(tag1).

darius-r:
My observations in that last comment can probably be disregarded. 

Given that its an Eisenhower Matrix, and I failed in my test to add Excludes for the adjacent tags I created, the resulting behaviour(the other tag being added) is expected.  They also had nowhere else to go in the matrix since I didn't have a panel configured in the case of both tags being included.

johannesjo:
Thank you very much for checking. @darius-r !

Issue #5995, opened 2026-01-14
🚨 Pomodoro timer breaks


beerkumquatpome:

beerkumquatpome:
@johannesjo 

### New bug: Pressing "Back to Planning" can break Pomodoro syncing

Settings:

1. Sync focus sessions with time tracking
2. Start focus sessions with banner only (no overlay)
3. Skip preparation screen (rocket animation)

Steps:

1. Enter Focus overlay
2. Select Pomodoro Timer
3. Select a Task
4. Start the Pomodoro Timer
5. Press "Complete focus session"
6. Press "Back to Planning"

It should stop time tracking because:
1. The timer has ended (When you press "Back to Planning", it will end the timer)
3. Sync focus sessions with time tracking is enabled

PR #7767, opened 2026-05-24
feat: auto-complete break and cycle to next focus session in Pomodoro mode

adcool655-droid:
## Problem
After each focus session or break ends, the user has to manually click to start the 
next phase. This breaks flow and requires checking the app.

## Solution
Adds an opt-in toggle "Auto-complete break and start next session (Pomodoro only)" 
in Settings → Productivity → Focus Mode.

When enabled, the existing `detectBreakTimeUp$` effect in `focus-mode.effects.ts` 
dispatches `completeBreak`, which triggers the existing `autoStartSessionOnBreakComplete$` 
to start the next focus session — no component-level logic needed.

## Changes from v1 (addressing review feedback)
- Moved logic from `FocusModeBreakComponent` into `detectBreakTimeUp$` effect
- Removed `_breakWasActive` instance state and `constructor effect()` anti-pattern
- Gated to Pomodoro mode only via `selectMode`
- Leverages existing `skipWhileApplyingRemoteOps()` for sync correctness
- No race with `skipBreak` / `exitBreakToPlanning`
- Works regardless of whether the focus overlay is open
- Implemented as opt-in `isAutoCompleteBreak` toggle (default: false), symmetric to `isManualBreakStart`
- Restored all three deleted 'why' comments in `focus-mode-break.component.ts`

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for the contribution\! A multi-agent review (correctness / architecture / alternatives / simplicity / perf / security) converged on the same conclusion: the **feature is reasonable, but the implementation should live in `focus-mode.effects.ts`, not in `FocusModeBreakComponent`**. Detail below.

## Architectural mismatch (blocking)

`src/app/features/focus-mode/store/focus-mode.effects.ts:281-301` already has a `detectBreakTimeUp$` effect that fires on the exact same condition this PR is trying to detect (`timer.purpose === 'break' && \!timer.isRunning && elapsed >= duration`), already deduped via `distinctUntilChanged(startedAt, elapsed)`, and already guarded by `skipWhileApplyingRemoteOps()`. It currently does `{ dispatch: false }` and only calls `_notifyUser()` — by design, per the comment on line 280 (*"no auto-complete"*) and the reducer comment on `focus-mode.reducer.ts:301` (*"stop timer but stay on break screen for user confirmation"*).

The whole change collapses to ~5 lines of:

```ts
withLatestFrom(this.store.select(selectors.selectPausedTaskId)),
map(([_, pausedTaskId]) => actions.completeBreak({ pausedTaskId })),
```

…then drop the local `tap(_notifyUser)` because `notifyOnBreakComplete$` (effects.ts:527) already fires the notification on `completeBreak`. The existing `autoStartSessionOnBreakComplete$` (effects.ts:506) handles the next-session start for Pomodoro and correctly no-ops for Flowtime/Countdown via `strategy.shouldAutoStartNextSession`.

This gives you:
- correct sync-replay behavior for free (`skipWhileApplyingRemoteOps`),
- no race with `skipBreak` / `exitBreakToPlanning`,
- no dependence on the focus-mode overlay being open,
- no duplicate detection across two layers.

## Specific issues with the component-level approach

1. **`skipBreak` / `exitBreakToPlanning` race.** Both reducers reset the timer to idle (`duration=0, elapsed=0`). `selectTimeRemaining` then returns `Math.max(0, 0-0) = 0`. If the user clicks skip/exit mid-break, `_breakWasActive` is already `true`, so the effect fires `completeBreak` *after* the explicit exit — which then triggers `autoStartSessionOnBreakComplete\$` → unwanted `startFocusSession`.

2. **Feature is silently disabled outside the focus overlay.** `FocusModeBreakComponent` only mounts on `FocusScreen.Break` while the overlay is open. Close the overlay during a break (the header indicator is the intended outside-overlay surface) and the timer can hit 0 with no listener.

3. **Missing `skipWhileApplyingRemoteOps()` guard.** Even though focus-mode state isn't in the op-log today, the sibling effects file uses this guard precisely because the timer selector can transiently land in completion-state during remote replay. The component effect has no equivalent — and that's the kind of thing the *Sync-correctness rules* in `CLAUDE.md` exist to catch early.

4. **`_breakWasActive` recreates store-derived state as mutable instance state.** If the component is destroyed and re-created between \"break ran\" and \"break stops\" (route navigation, sync hydration, app reload mid-break), the flag resets and the auto-complete is silently lost. The store already knows this via `startedAt \!== null && elapsed >= duration`.

5. **Constructor `effect()` + side effect** is in CLAUDE.md's anti-pattern table (\"Side effects in constructors → use async pipe / `toSignal`\"). The right escape hatch for action dispatch is an NgRx effect.

6. **PR title says \"in Pomodoro mode\" but there's no mode gate.** In Flowtime, `autoStartSessionOnBreakComplete\$` already no-ops, but `completeBreak` still fires `notifyOnBreakComplete\$` and `resumeTrackingOnBreakComplete\$`, changing Flowtime behavior unintentionally. If you go the effects.ts route, you get correct gating via the existing strategy field for free.

## Out-of-scope changes to revert

The diff also strips three unrelated comment blocks documenting prior bug fixes:
- line 50 (old): `// Get pausedTaskId before break ends (passed in action to avoid race condition)` — pairs with `effects.ts:490-491` (\"reducer clears pausedTaskId before effect reads state\").
- lines 80–83 (old): the `Bug #5995 Fix` block in `pauseBreak()` explaining the `currentTaskId || storePausedTaskId` precedence.
- lines 93–94 (old): the Flowtime rationale in `resumeBreak()` for the *opposite* `storePausedTaskId ?? currentTaskId` precedence.

`CLAUDE.md` calls out \"Document 'why' not 'what'\" — these are textbook \"why\" comments tied to specific issues, and they explain why two adjacent methods invert their precedence. Please restore them.

## Test coverage

`focus-mode-break.component.spec.ts` is unchanged. If you keep any component-side gate, please add at minimum:
- auto-complete fires once when `isRunning` transitions to false with `remaining === 0` in Pomodoro,
- does NOT fire after `skipBreak` / `exitBreakToPlanning` mid-break,
- does NOT fire on a paused break (`remaining > 0`),
- does NOT fire on the Flowtime `BreakOffer` state.

If you move the logic into `detectBreakTimeUp\$`, the relevant spec is `focus-mode.effects.spec.ts` (testing fixture already in place).

## Also: a UX decision worth surfacing

The reducer comment explicitly says the break screen stays up *for user confirmation*. That's a deliberate UX decision (similar in spirit to the existing `isManualBreakStart` flag for the work→break transition). Worth checking with the maintainer whether this should be unconditional for Pomodoro, gated by `shouldAutoStartNextSession`, or surfaced as a new `isAutoCompleteBreak` (symmetric to `isManualBreakStart`).

---

Happy to help with the refactor if useful — the effects.ts version is genuinely smaller than the current diff.

adcool655-droid:
@johannesjo  I've addressed all the feedback from your review. 
The logic has been moved to `detectBreakTimeUp$` in `focus-mode.effects.ts`,  I have restored the deleted comments and added mode gate. The feature is now opt-in  via a settings toggle. Would appreciate another look!

beerkumquatpome:
Can you rename the setting to Auto-start focus (Pomodoro)? `isAutoCompleteBreak` should also be renamed to `isAutoStartFocus`
Every pomodoro timer names this setting "Auto-start focus" or "Auto-start Pomodoros"
@adcool655-droid

johannesjo:
Thanks for the update. I double-checked the current head (`5d5ecff`) and there is one remaining blocker.

`detectBreakTimeUp$` now filters completed breaks behind `isAutoCompleteBreak && Pomodoro`, and the old unconditional `_notifyUser()` path was removed. Since the new setting defaults to `false`, the default/manual-confirmation behavior regresses: completed breaks no longer trigger the break-ended sound/window notification. `notifyOnBreakComplete$` does not cover this path because it only runs after `completeBreak`, which is now only emitted when auto-complete is enabled.

Suggested shape:

- keep the break-time-up notification for every completed started break, preserving the existing default behavior;
- separately dispatch `completeBreak({ pausedTaskId })` only when `isAutoCompleteBreak` is enabled and the mode is Pomodoro;
- update `focus-mode.effects.spec.ts` so the existing notification cases still pass, plus add coverage for the new opt-in path, `pausedTaskId` payload, and Pomodoro-only guard.

The existing `detectBreakTimeUp$` specs around break completion (`focus-mode.effects.spec.ts`) are a good place to encode both behaviors.

PR #7770, opened 2026-05-24
fix(ios): prevent tag picker from auto-opening on portrait task open

### LINKED ISSUE #7769: 🚨 Tag picker auto-opens whenever a task is opened on iOS portrait
### Version Used

iOS app v9.12

### Current Behavior

On iPhone in portrait, tapping any task to open its detail panel immediately opens the tag selection autocomplete dropdown, covering the rest of the task detail. The user has to tap somewhere else to dismiss the dropdown before they can interact with any other field.

### Expected Behavior

Opening a task detail panel should not focus any input or open any popover. The tag picker should only open when the user taps the tag field explicitly.

### Steps to Reproduce

1. Open SP on an iPhone in portrait orientation.
2. Tap any task in the list to open its detail panel (bottom sheet slides up from below).
3. The tag selection is already open, covering the rest of the task detail.

Rotating the device to landscape and reopening the task → no tag selection opens; behaviour is as expected. The bug only occurs in portrait.

### Console Output

```shell
n/a (iOS app, no devtools available).
```

### Documentation

d0g2342:
Fixes #7769.

## Problem

On iPhone in portrait, opening any task immediately opens the tag selection
autocomplete, covering the rest of the task detail. Landscape (right panel)
is unaffected.

## Root cause

The mobile bottom sheet is opened in `RightPanelComponent` without an
explicit `autoFocus`, so CDK's FocusTrap falls back to the default
`'first-tabbable'`. The intended close button in the sheet header is
`display: none` (`bottom-panel-container.component.scss`), so it is skipped
and focus lands on the tag-edit `<input>` inside the task detail panel.
On touch, the bound `mat-autocomplete` opens immediately on focus, hence
the dropdown every time.

The right-panel path (non-xs breakpoint) doesn't use `MatBottomSheet`, so
no FocusTrap — that's why landscape is fine.

## Fix

One line: pass `autoFocus: false` to the bottom-sheet open call. This
matches what `task-detail-panel.component.ts` already does for its own
dialog opens (`scheduleTask`, `openDeadlineDialog`).

## Verified

- Reproduced on the App Store build (SP v18.7.0 / iOS app v9.12).
- Verified the fix on a local self-build of this branch installed on an
  iPhone — tag picker no longer auto-opens, neither in portrait nor
  landscape, and tapping the tag field explicitly still opens it as
  expected.

### DIFF
--- a/src/app/features/right-panel/right-panel.component.ts
+++ b/src/app/features/right-panel/right-panel.component.ts
@@ -235,6 +235,11 @@ export class RightPanelComponent implements AfterViewInit, OnDestroy {
                 hasBackdrop: true,
                 closeOnNavigation: false,
                 panelClass: 'bottom-panel-sheet',
+                // Default 'first-tabbable' otherwise focuses the tag-edit
+                // input (the bottom-panel close-btn is display:none and gets
+                // skipped), which on touch makes mat-autocomplete pop open
+                // immediately on every task open in portrait.
+                autoFocus: false,
               },
             );
 


johannesjo:
Thank you very much! <3

Issue #5221, opened 2025-10-04
🚨 Duplicating a parent task clears out subtasks due dates


MatMercer:
### Version Used

docker:latest & Google Chrome

### Expected Behavior

Subtasks should retain due dates.

### Current Behavior

They are cleared.

### Steps to Reproduce

1. Go to TODAY page `/#/tag/TODAY/tasks`
2. Create a task named PARENT
3. Add a sub task named SUB
4. Add a due date to the SUB task
5. Duplicate PARENT
6. Notice that the SUB task inside PARENT (copy) is without a due date now

https://github.com/user-attachments/assets/c0842502-c658-460c-9cc9-0267a6d6492f

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

MatMercer:
HEADS UP

I'm planning to pick this up in ~1 month and submit a PR, seems simple to implement.

johannesjo:
Great to hear this! Let  me know if you need anything!

PR #7788, opened 2026-05-25
Fix/issue 5221

RaiFSilva:
## Problem

Fixed an issue where Duplicating a parent task clears out subtasks due dates.

## Solution

The dueWithTime parameter was included in the fields copied on all son tasks in the duplicate method.
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo (changes requested):
Thanks for the PR and welcome\! 🎉 A couple of things to address before we merge:

### 1. Incomplete fix — `dueDay` is missing
`dueDay` (date-only) and `dueWithTime` (timestamp) are mutually exclusive representations of a schedule (see `task.model.ts:97-114`). The parent-task copy a few lines above (lines 458-462) conditionally spreads **both**, but this PR only adds `dueWithTime`. The reproduction in #5221 — "Add a due date to the SUB task" via the schedule dialog without a time — sets `dueDay`, not `dueWithTime`, so the current diff probably doesn't actually fix the reported case, only the with-time variant.

### 2. Unconditional assignment has a side effect on TODAY
The parent block uses \`...(this.task.dueWithTime && { dueWithTime: ... })\` so the field isn't present when undefined. This PR sets it directly, which means \`subTaskInfo.dueWithTime === undefined\` lands explicitly on the object. In \`createNewTaskWithDefaults\` (\`task.service.ts:1343-1348\`) the check is \`\!additional.dueWithTime && \!('dueDay' in additional)\` — \`\!undefined\` is truthy and \`'dueDay' in additional\` is false, so when the user is on the TODAY tag, *any* subtask of the duplicated task gets auto-assigned \`dueDay: today\` even if the original subtask had no schedule at all. (Note: \`additional.parentId\` isn't set at this call site — the parent is attached later via the \`addSubTask\` action — so the \`\!additional.parentId\` guard in that branch doesn't help here.)

### Suggested change
Mirror the parent's conditional-spread style so undefined fields stay absent:

\`\`\`ts
const subTaskInfo = {
  isDone: subTask.isDone,
  projectId: subTask.projectId,
  ...(subTask.dueDay && { dueDay: subTask.dueDay }),
  ...(subTask.dueWithTime && { dueWithTime: subTask.dueWithTime }),
  timeEstimate: subTask.timeEstimate,
  notes: subTask.notes,
};
\`\`\`

### Nits
- A small spec covering both \`dueDay\` and \`dueWithTime\` carryover on subtasks would lock the fix in (the checklist mentions tests but none are in the diff).
- The repo uses Angular commit format — title like \`fix(tasks): preserve subtask due dates when duplicating parent\` is preferred over \`Fix/issue 5221\`.

johannesjo:
Thanks for tackling #5221 — the underlying data-loss bug (subtask due dates dropped on duplication) is real and worth fixing.

Two things block merging as-is:

1. **Incomplete fix.** The change copies only `dueWithTime`, but an all-day due date — the case in the issue's own repro (scheduling a date with no time) — is stored in `dueDay`, which is still dropped on duplication. So the reported scenario isn't actually fixed yet. The subtask copy should mirror the parent-task block a few lines above and conditionally spread **both** fields:

   ```ts
   ...(subTask.dueDay && { dueDay: subTask.dueDay }),
   ...(subTask.dueWithTime && { dueWithTime: subTask.dueWithTime }),
   ```

   Using the conditional spread (rather than an unconditional `dueWithTime: subTask.dueWithTime`) also avoids putting an explicit `dueWithTime: undefined` onto the new-task object passed to `createNewTaskWithDefaults`.

2. **Test.** Please add a spec asserting both `dueDay` and `dueWithTime` carry over on duplicate — the existing `duplicate()` suite in `task-context-menu-inner.component.spec.ts` is a good home.

With those two changes this should be good to go.

Zoriot:
@RaiFSilva do you want to implement the necessary changes? 

Else i can probably look into it if that's still the way to go.

PR #7792, opened 2026-05-26
fix(tasks): use correct label when re-opening a completed task

### LINKED ISSUE #7785: 🚨 Mobile swipe action always shows "Mark as done" even for completed tasks
### Version Used

18.4.4_L2A (Android, Fairphone 5)

### Current Behavior

The swipe action always displays "Mark as done" (MARK_DONE key) regardless 
of the task's current state. The action itself works correctly (it does 
re-activate the task), but the label is misleading.

### Expected Behavior

When swiping left on a completed task, the action label should reflect 
the actual action: "Mark as undone" (or equivalent), since the task is 
already done.

### Steps to Reproduce

1. Mark a task as done on Android
2. Swipe left on that completed task
3. The action label reads "Mark as done" instead of "Mark as undone"

**Suggestion**
Add a separate i18n key (e.g. MARK_UNDONE) and use it conditionally 
based on the task's current done state.

**Note**
This action (re-activating a completed task) is not available on desktop 
(PC), where unchecking the task checkbox achieves the same result.

### Console Output

```shell

```

### Documentation

costajohnt:
## Problem

Closes #7785.

The task context menu's toggle-done button always shows "Mark as completed" regardless of whether the task is already done. The matching icon a few lines above already switches between `check` (when not done) and `undo` (when done), so the icon was correct but the label was not.

Root cause is a copy-paste bug on line 151 of `task-context-menu-inner.component.html`:

```ts
{{ (task.isDone ? T.F.TASK.CMP.MARK_DONE : T.F.TASK.CMP.MARK_DONE) | translate }}
```

Both branches of the ternary point to `MARK_DONE`.

## Solution

Add a new i18n key `MARK_UNDONE` ("Mark as not completed") and use it on the `task.isDone` branch of the ternary so the label matches the icon.

Files touched:

1. `src/app/t.const.ts`: add `MARK_UNDONE: 'F.TASK.CMP.MARK_UNDONE'` next to the existing `MARK_DONE` key.
2. `src/assets/i18n/en.json`: add `"MARK_UNDONE": "Mark as not completed"`, symmetric with the existing `"MARK_DONE": "Mark as completed"`.
3. `src/app/features/tasks/task-context-menu/task-context-menu-inner/task-context-menu-inner.component.html`: change the ternary's `isDone` branch to reference `MARK_UNDONE`.

Other locale files are not touched, matching the convention used by recent i18n-touching commits (the same pattern as `aeae4f959` and `1ae4aaf02`): `en.json` is the source of truth, other locales fall back to English via `fallbackLang` configured in `src/main.ts` until translators catch up.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

  Not applicable, this is an internal label fix with no user-facing documentation impact.

  No new tests. The change is a 1-line label fix in an HTML template. The existing 21 tests in `src/app/features/tasks/task-context-menu/**/*.spec.ts` still pass. Recent label-only i18n PRs (e.g. `aeae4f959`, `1ae4aaf02`) follow the same no-new-test convention for this surface.


johannesjo:
Thank you! <3

PR #7808, opened 2026-05-26
feat(jira): manual worklog for in-progress tasks with local tracking

h0ru5:
## Problem

The Jira worklog dialog only fires when a task is marked as done. Users who work on a Jira ticket over multiple sessions without completing it cannot log time to Jira from within Super Productivity. Additionally, there is no local record of how much time has already been logged to Jira, making it impossible to know the outstanding unlogged balance.

### New field: `timeLoggedToJira` on the task model

Tracks how much time has been logged to Jira from SP locally (ms). After each successful worklog submission the field is incremented by the submitted duration. The unlogged delta is always `Math.max(0, task.timeSpent - (task.timeLoggedToJira ?? 0))`.

This is more robust than the existing `AllTimeMinusLogged` mode which reads `issue.timespent` from the Jira API — that approach yields 0 for tasks that already had worklogs before SP started tracking them.

### New service: `JiraWorklogService`

Extracts the worklog dialog logic from `JiraIssueEffects._openWorklogDialog` (which is removed) into a dedicated injectable. `JiraIssueEffects` now delegates to `openWorklogDialogForTask(task)`. The service is reused by all three new entry points below.

### Two new manual entry points

1. **Task context menu** — "Log work to Jira" item appears for Jira-linked tasks.
2. **Duration dialog** — "Save & log to Jira" button saves the entered time then opens the worklog dialog pre-filled with the unlogged delta.

### Unlogged time badge on task rows

Jira tasks with outstanding unlogged time show a small badge (`<mat-icon svgIcon="jira"> + msToString`) inline next to the existing time display.

### Daily summary — Jira pending section

A new `JiraUnloggedSummaryComponent` in the "Review tasks" tab lists every Jira task from today that has unlogged time, with a per-row "Log work" button.

## Files changed

| File | Change |
|---|---|
| `task.model.ts` | Add `timeLoggedToJira?: number` to `IssueFieldsForTask` |
| `jira-worklog.service.ts` | New service (extracted + extended from effects) |
| `jira-worklog.service.spec.ts` | Unit tests for guard conditions |
| `jira-issue.effects.ts` | Delegate to `JiraWorklogService`, remove private method |
| `task-context-menu-inner.component.*` | "Log work to Jira" menu item |
| `dialog-time-estimate.component.*` | "Save & log to Jira" button |
| `task.component.*` | Unlogged-time badge + computed signal |
| `jira-unlogged-summary/` | New standalone component (ts/html/scss/spec) |
| `daily-summary.component.*` | Wire in `JiraUnloggedSummaryComponent` |
| `en.json` / `t.const.ts` | Four new i18n keys |

## Design doc

`docs/superpowers/specs/2026-05-26-jira-manual-worklog-design.md`


johannesjo:
Thank you very much!



## Multi-agent review

Six independent reviewers (correctness, security, architecture, alternatives, performance, simplicity) looked at the diff. Below are the verified, cross-confirmed findings — false positives dismissed.

### Critical

1. **Subtask-done auto-worklog regression** — `src/app/features/issue/providers/jira/jira-issue.effects.ts:75`
   The old code called `_openWorklogDialog(subTask, assertTruthy(mainTask.issueId), jiraCfg)`, intentionally passing the **parent's** issueId so a plain subtask under a Jira parent (the common case) could still log work against the parent issue. The new `openWorklogDialogForTask(subTask)` early-returns at `jira-worklog.service.ts:22-24` when `subTask.issueType \!== JIRA_TYPE || \!subTask.issueId`, so the dialog silently never opens for non-Jira subtasks. Existing `isAddWorklogOnSubTaskDone` users will see the feature go quiet after upgrade.
   *Fix:* either keep the explicit parent issueId at the call site, or give the service an optional `issueIdOverride`/`taskForApi` parameter so the effect can route subtask completions to the parent's issue.

2. **No backfill for `timeLoggedToJira` — invites double-logging** — `src/app/features/issue/providers/jira/jira-worklog.service.ts:38`
   The removed `_openWorklogDialog` seeded `timeLogged = issue.timespent * 1000` from the Jira API. The new code uses `task.timeLoggedToJira ?? 0`. Every existing task on day 1 has `undefined`, so:
   - The new row badge reports the full `timeSpent` as unlogged.
   - `AllTimeMinusLogged` defaults to `timeSpent - 0` → the dialog actively invites the user to re-log time that is already in Jira.

   *Elegant fix (no migration needed):* the service already fetches `issue.timespent` from Jira. Lazy-init on first open and persist via the same `updateTask` path:
   ```ts
   const timeLoggedToJira = task.timeLoggedToJira ?? issue.timespent * 1000;
   if (task.timeLoggedToJira === undefined) {
     this._taskService.update(task.id, { timeLoggedToJira });
   }
   ```
   This also gives a soft answer to the multi-device LWW concern (#4 below): periodic re-sync from Jira would self-heal divergence if you ever want it.

3. **`submitAndLogToJira` operates on stale task data** — `src/app/features/tasks/dialog-time-estimate/dialog-time-estimate.component.ts:111-114`
   ```ts
   submitAndLogToJira(): void {
     this.submit();                                          // closes this dialog, dispatches update
     this._jiraWorklogService.openWorklogDialogForTask(this.task); // pre-update reference
   }
   ```
   `this.task` is `this.data.task` (the pre-edit snapshot). The worklog dialog computes `AllTimeMinusLogged` against the stale `timeSpent`, so the user's just-saved edit isn't reflected in the suggested worklog amount. Either await the update (re-fetch via `getByIdOnce$`) or pass `this.taskCopy` with the recomputed values.

4. **`timeLoggedToJira` doesn't belong on the generic task model** — `src/app/features/tasks/task.model.ts:71`
   `IssueFieldsForTask` is shared by ~11 providers (GitHub, GitLab, Gitea, Redmine, OpenProject, CalDAV, Linear, Trello, iCal, …). Embedding a Jira-named counter here either leaks the provider into a polymorphic interface or sets you up for `timeLoggedToGitlab`, `timeLoggedToRedmine`, … Either:
   - Rename `issueTimeLogged` and make it a provider-agnostic contract; or
   - Fold into the existing `issueTimeTracked: IssueTaskTimeTracked` field at line 65, which already exists for adjacent purposes; or
   - Move under a provider sub-object on the task.

   Cheap to fix now, painful to migrate later.

### Warnings

5. **Daily-summary section renders for any historical day** — `src/app/pages/daily-summary/daily-summary.component.html:181`
   `<jira-unlogged-summary>` is only gated on `tasks?.length`, unlike sibling sections that gate on `isForToday`. Clicking "Log work" while reviewing a past day opens a dialog that posts against *now*, not the day being reviewed. Gate on `isForToday`.

6. **The success-path increment is untested** — `src/app/features/issue/providers/jira/jira-worklog.service.spec.ts`
   The spec only exercises the three guard-clause early returns. The whole point of the feature — the `tap(() => TaskService.update(task.id, { timeLoggedToJira: prev + params.timeSpent }))` at `jira-worklog.service.ts:60-64` — has zero coverage. Easy regression target.

7. **Multi-device sync: numeric LWW counter diverges from Jira reality**
   `timeLoggedToJira = prev + params.timeSpent` is a last-write-wins update. Two devices each logging 30m concurrently: Jira receives both, the local counter only reflects one writer's `prev + 30m`, so the badge will permanently overstate "unlogged" time. The lazy-init fix in #2 above doubles as the mitigation if you ever periodically re-seed from `issue.timespent`.

8. **Redundant config fetch** — `src/app/features/issue/providers/jira/jira-issue.effects.ts:68`
   The effect already has `jiraCfg` in scope from `_getCfgOnce$`, then `openWorklogDialogForTask` immediately re-fetches it. Trivial fix: accept an optional `cfg` param.

### Suggestions

9. **Consider whether the diff needs all three new UI surfaces** — context-menu item + duration-dialog button + row badge + daily-summary section is a lot of surface for one capability. The context-menu entry alone (with optionally the badge) covers the discovery story. The duration-dialog button is a separate UX decision worth its own PR. Trimming would also reduce review/test/i18n surface by ~60%.

10. **`JiraUnloggedSummaryComponent` location** — lives under `src/app/features/issue/providers/jira/` but is imported by a top-level page (`src/app/pages/daily-summary/`). Sibling summary widgets live under `pages/daily-summary/...-summary-item/`. Either move it there or have the daily-summary template derive `pendingTasks` inline (~15 lines, no new component).

11. **`unloggedMs(task)` called from template runs on every CD pass** — `jira-unlogged-summary.component.html:9`. Precompute `{ task, unloggedMs }` tuples in the `pendingTasks` computed so the template is branchless. (Performance impact is small; mainly a readability win.)

12. **`'GITHUB' as any` in specs** — `jira-worklog.service.spec.ts`, `jira-unlogged-summary.component.spec.ts`. Use the proper issue-type union to stay strict-TS clean.

13. **`var(--separator-color, rgba(0,0,0,0.1))` fallback** — `jira-unlogged-summary.component.scss:15`. `--separator-color` is always defined in the theme; the literal fallback won't respect dark mode. Drop it.

14. **930-line agentic-workflow plan doc** — `docs/superpowers/plans/2026-05-26-jira-manual-worklog.md`. The design spec (229 lines) is useful; the plan is process scratch. Worth confirming with the maintainer whether it belongs in-tree.

### Strengths

- Clean extraction into `JiraWorklogService`; consumed identically by all entry points.
- Reuses `TaskSharedActions.updateTask` rather than inventing a new action — sync/op-log invariants hold automatically.
- Standalone components, OnPush, signal inputs, `computed` — matches the modern Angular style this repo is moving toward.
- No new dependencies; no security/XSS surface introduced (template uses `{{ }}` interpolation throughout).
- Design spec committed alongside code.

### Verdict

**Needs changes.** Items 1 + 2 are the two that should block; both have small, well-scoped fixes (parent-issueId override; lazy-init from `issue.timespent`). Item 3 is the same shape — a one-line oversight. Item 4 is worth resolving while the field is still new and unpersisted in user data.

---
*Reviewers: Claude × 6 (correctness, security, architecture, alternatives, performance, simplicity), synthesized.*

johannesjo:
## Review of the `convertToSubTask` / "log from any task" range (`17ec6f4..0e6710e`)

Reviewed the second feature stacked on this PR — **"Log time to Jira ticket…"** and **"Assign as subtask of Jira issue…"** (`DialogJiraIssuePickerComponent` + the new `convertToSubTask` action/meta-reducer). The op-log wiring is complete and correct (enum + `HCS` code map + count test bumped to 144, guarded by the 1:1 test), the meta-reducer approach is right (one reducer pass = one op), and the picker escapes JQL (`\` and `"`) and uses OnPush/signals/debounce/lazy-import. Nice.

Three blocking correctness bugs in `handleConvertToSubTask`, all from the same root cause — it reparents the task but doesn't fully normalize it into a valid subtask:

### Critical

**1. `remindAt` is not cleared → ghost reminders keep firing.** `task-shared-crud.reducer.ts:227-232` clears `dueDay`, `dueWithTime`, `reminderId`, but **not `remindAt`**. The current reminder worker is driven by `selectAllTasksWithReminder`, which filters on `typeof task.remindAt === 'number' && !task.isDone` (`task.selectors.ts:670`) — `reminderId` is a legacy field that no longer gates reminders. The canonical `unscheduleTask` handler clears all three (`task-shared-scheduling.reducer.ts:136-138`). So a scheduled task converted to a subtask will keep firing its reminder. The `it('should clear dueDay…')` test only asserts `dueDay`, so it passes despite this.
*Fix:* add `remindAt: undefined` to the task changes (mirror `unscheduleTask`).

**2. Subtask ends up in a different project than its parent.** The task's `parentId` is set but its `projectId` is left untouched (`task-shared-crud.reducer.ts:225-233`). `addSubTask` explicitly enforces `projectId: parentTask.projectId` ("should always be the one of the parent", `task.reducer.ts:591`). In this flow `addTaskFromIssue` imports the Jira parent into the provider's `defaultProjectId`, which routinely differs from the user's task's project → a subtask whose `projectId` ≠ its parent's. This bites later: `handleConvertToMainTask` re-adds the task to *its own* (now-wrong) project's `taskIds`, and deleting the user's original project would target a subtask living under a parent in another project. The spec masks it because `task1` and `jiraTask` are both in `project1`.
*Fix:* add `projectId: parentTask.projectId` to the task changes; the project-removal block above already correctly keys off the original `task.projectId`.

**3. No guard against converting a task that already has subtasks → unsupported 2-level nesting.** The menu gate is only `!task.parentId && task.issueType !== JIRA_TYPE` (`task-context-menu-inner.component.html:235`). A root task *with its own subtasks* qualifies; after conversion you get `jiraParent → task → grandchildren`, which SP's single-level model doesn't support (the design doc's own out-of-scope note rejects drag-drop root reparenting for exactly this reason). `handleConvertToSubTask` reparents only the task and leaves its children dangling.
*Fix:* hide the item when `task.subTaskIds?.length`, or reject in `_assignAsSubtask`/the reducer.

### Important

- **Stale `this.task` in `DialogTimeEstimateComponent.logToJiraTicket()`** (`dialog-time-estimate.component.ts:117-118`): `this.submit()` then opens the picker with `this.task` (the pre-edit `data.task` snapshot). The worklog dialog defaults to `AllTimeMinusLogged` against the stale `timeSpent`, so a just-saved edit isn't reflected. Same defect already flagged as critical #3 for `submitAndLogToJira` — re-fetch or pass the recomputed copy.
- **Dedup fallback can resolve a subtask as the parent** (`task-context-menu-inner.component.ts` `_assignAsSubtask`): when `addTaskFromIssue` returns `undefined` (already imported) and the existing task is found by `issueId`, that task could itself be a subtask → nesting again. Guard that the resolved parent is a root task.
- **Tests don't exercise the failing paths**: same-project only, `dueDay`-only assertion (no `remindAt`), no task-with-subtasks case, no `projectId === parent.projectId` assertion. Adding those would have caught #1 and #2.

### Minor

- `logTimeToJiraTicket` is duplicated near-verbatim in `task-context-menu-inner.component.ts` and `dialog-time-estimate.component.ts` (dynamic import → open picker → `openWorklogDialogForExternalTask`). Could move to `JiraWorklogService` (e.g. `pickIssueAndLogTime(task)`).
- `modified: Date.now()` in the reducer is non-deterministic on replay, but it matches the existing `handleConvertToMainTask` — consistent, just noting.

**Verdict: needs changes.** Items 1–3 are small, well-scoped fixes (clear `remindAt`; set `projectId: parentTask.projectId`; guard tasks with existing subtasks). 1 and 2 are silent data-integrity issues, so worth fixing before merge.

johannesjo:
## 🔍 Multi-agent review — `17ec6f4e..0e6710e0` (log-from-any-task + assign-as-subtask)

Reviewed the incremental range across 7 independent passes (correctness, security, architecture, alternatives, performance, simplicity, + Codex). Verdict: **needs changes** — three blocking issues, all verified against the code. The rest is solid (clean meta-reducer modelling, complete op-log registration, OnPush/lazy-loaded picker, escaped templates, i18n in both files).

### 🔴 Blocking

**1. Logging time to a Jira ticket from a non-Jira task silently fails on submit.**
`openWorklogDialogForExternalTask` resolves the *picked* provider's `jiraCfg` correctly, but `DialogTrackTimeComponent` re-derives the provider id from the **original** task:
```ts
// dialog-track-time.component.ts:117
private _issueProviderIdOnce$ = this.data.task.issueProviderId
  ? of(this.data.task.issueProviderId)
  : this._taskService.getByIdOnce$(this.data.task.parentId as string).pipe(/* throws if no provider */);
```
`submit()` awaits this *before* calling `onSubmit`. For a plain task (the exact target of the "Log time to Jira ticket" item — `task.issueType !== JIRA_TYPE`), `getByIdOnce$(undefined)` resolves to `undefined` → `TypeError` → `submit()` rejects → the worklog is never sent. For a task linked to a *different* provider, the "set as default time" checkbox writes `worklogDialogDefaultTime` to the wrong provider.
**Fix:** pass the picked `issueProviderId` into the dialog `data` and prefer it over `task.issueProviderId`.

**2. `convertToSubTask` doesn't recalculate the parent's time roll-up.**
`handleConvertToSubTask` (`task-shared-crud.reducer.ts:212`) appends to `parent.subTaskIds` but omits the `reCalcTimesForParentIfParent` call that `addSubTask` (`task.reducer.ts:617`), `moveSubTask`, and `convertToMainTask` all make. Parent `timeSpent`/`timeEstimate` are stored, not derived — so the Jira parent stays at its old totals, and the unlogged-time badge / daily-summary section (the point of the feature) never appear.

**3. No guard against converting a task that already has subtasks → illegal 3-level nesting.**
Menu condition (`task-context-menu-inner.component.html:235`) is `!task.parentId && task.issueType !== JIRA_TYPE` — missing the `!task.subTaskIds.length` guard that every sibling sub-position item uses (lines 92/109/127). The reducer doesn't guard either. Converting a parent task keeps its children, producing a task with both `parentId` and non-empty `subTaskIds`.
**Fix:** add `&& !task.subTaskIds.length` to the menu + a defensive no-op in the reducer.

### 🟡 Worth addressing

- **Subtask keeps its own `projectId`** instead of adopting the parent's (`addSubTask` sets `projectId: parentTask.projectId`; `handleConvertToSubTask` doesn't) → cross-project inconsistency when the Jira parent lands in the provider's default project.
- **Board sections not cleaned on conversion** — `section-shared.reducer.ts` handles `moveToOtherProject` but has no `convertToSubTask` handler, leaving a dangling task id in the project's sections (boards users only).
- **Jira search logs issue summaries to the exportable log** — the picker is the only live caller of `search$`, whose mapper does `IssueLog.log(res)` (`jira-issue-map.util.ts:65`), recording issue keys + summaries (user content) → violates the no-user-content-in-logs rule. Suggest `IssueLog.log({ count: res?.response?.issues?.length })`.
- **Search errors swallowed as `[]`** (`dialog-jira-issue-picker.component.ts`) — auth/network failures look identical to "no results"; surface a translated error state.
- **Dedup fallback misses archived issues** — `_assignAsSubtask` scans `selectAllTasks` by exact `issueId`, while `addTaskFromIssue` uses `checkForTaskWithIssueEverywhere` (includes archive + alt id forms) → spurious "could not find/create" snack for already-imported (archived) issues. Reuse `checkForTaskWithIssueEverywhere`.
- **Defensive guards** — no self-parent / circular-ref guard in the reducer (the action is op-log-replayable, so worth matching `moveSubTask`'s posture).

### 🟢 Optional cleanups
- Reuse `IssueService.searchIssues` / the Jira `issuePicker$` endpoint instead of hand-building JQL in the picker.
- DRY: the "open picker → log time" block is duplicated across two components (and the picker-open across three); `_openDialogForExternalTask` is ~90% a copy of `_openDialog` — extract shared helpers.
- `DialogJiraIssuePickerData.issueProviderId` is dead (all callers pass `{}`); the redundant `issueLabel` param can be derived from the already-fetched issue.
- Design spec (`2026-05-27-...-design.md`) is stale — documents `task.actions.ts` + `{ taskId }`, but it shipped (correctly) as `task-shared.actions.ts` + `{ task }` + meta-reducer.

---

<sub>🤖 Multi-agent review (6× Claude + Codex) via Claude Code. Findings verified against PR HEAD `0e6710e0`.</sub>

PR #7813, opened 2026-05-26
feat(schedule): include plugin calendar providers in filter and event color

sespiros:
## Problem

`CalendarIntegrationService.calendarEvents$` already reads schedule events from both iCal providers and plugin issue providers that opt into the agenda view (`useAgendaView: true`), but two pieces of the UI still only know about iCal:

1. The visibility-filter button (`showCalFilterBtn` in `schedule.component.ts`) reads `selectCalendarProviders`, which is filtered strictly to `ICAL_TYPE`.
   A user with one iCal feed plus one plugin calendar (e.g. the `caldav-calendar-provider` plugin against Exchange) sees a count of 1, so the filter button stays hidden and there's no way to toggle either source on the schedule.

2. Plugin calendar events go through `_fetchPluginCalendarEvents` without a `color` field, so they render with the default neutral border on the grid while iCal events from the same logical calendar are tinted with the iCal provider's configured `color`.

A side-effect of (1) is that the labels in the filter menu also don't match: iCal goes through `sanitizeIcalUrlForDisplay` (hostname-only) but `getIssueProviderTooltip` returns the raw first config string for plugins, so a plugin row would read `https://calendar.example.com/dav/...` next to an iCal row reading `outlook.office365.com`.

## Solution

- `enabledCalendarProviders` in the schedule component now combines `selectCalendarProviders` with `selectEnabledIssueProviders` filtered to plugin providers with `useAgendaView`, mirroring `calendarEvents$`'s own source set, including its `registrationChanges$` trigger so a plugin that registers after the store hydrates still shows up in the menu. Both branches are gated with `distinctUntilChanged(fastArrayCompare)` like the sibling stream.
- Menu rows (id, color, label) are precomputed in a `computed()` so the template no longer calls methods inside `@for`.
- `calProviderLabel` builds on `sanitizeIcalUrlForDisplay`: any scheme-prefixed plugin label is reduced to its hostname, and opaque schemes (`javascript:`, `data:`, ...) fall back to the provider key, so plugin rows format like iCal rows and never render a config string verbatim.
- New helper `getCalendarProviderColor` returns the provider's explicit `color`, then `pluginConfig.color`, then a deterministic HSL derived from the provider id. Only plain color literals (hex, `rgb()`/`hsl()`, named) are accepted, since `pluginConfig.color` is untrusted and lands in a `[style.background]` binding; anything else, e.g. `url(...)`, takes the fallback.
- The menu dot, `_fetchPluginCalendarEvents`, and `requestEvents$` all go through the helper, so the filter dot and the grid event always agree on a color. **Behavior change:** iCal feeds without a configured color, which previously rendered with the neutral border, now get the deterministic per-provider tint.
- Wrap the menu's dot + label in an inline-flex container with an explicit gap (and drop the now-redundant `margin-right` / `vertical-align` from `.cal-color-dot`) so spacing is identical whether the dot is the explicit user color or the HSL fallback.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

## Checklist

### DIFF
--- a/src/app/features/calendar-integration/calendar-integration.service.ts
+++ b/src/app/features/calendar-integration/calendar-integration.service.ts
@@ -43,6 +43,7 @@ import {
   isPluginIssueProvider,
 } from '../issue/issue.model';
 import { CalendarProviderCfg } from '../issue/providers/calendar/calendar.model';
+import { getCalendarProviderColor } from '../issue/mapping-helper/get-issue-provider-color';
 import { CORS_SKIP_EXTRA_HEADERS, IS_WEB_BROWSER } from '../../app.constants';
 import { Log } from '../../core/log';
 import { getErrorTxt } from '../../util/get-error-text';
@@ -284,6 +285,7 @@ export class CalendarIntegrationService {
     const results: PluginSearchResult[] =
       await provider.definition.getNewIssuesForBacklog(pluginProvider.pluginConfig, http);
 
+    const color = getCalendarProviderColor(pluginProvider);
     return results
       .filter((r) => r.start != null)
       .map((r) => ({
@@ -296,6 +298,7 @@ export class CalendarIntegrationService {
         isAllDay: r.isAllDay,
         issueProviderKey: pluginProvider.issueProviderKey,
         dueWithTime: r.dueWithTime,
+        color,
       }));
   }
 
--- a/src/app/features/issue/mapping-helper/get-issue-provider-color.spec.ts
+++ b/src/app/features/issue/mapping-helper/get-issue-provider-color.spec.ts
@@ -1,0 +1,65 @@
+import { getCalendarProviderColor } from './get-issue-provider-color';
+import {
+  IssueProvider,
+  IssueProviderCalendar,
+  IssueProviderPluginType,
+} from '../issue.model';
+
+const icalProvider = (over: Partial<IssueProviderCalendar> = {}): IssueProvider =>
+  ({
+    id: 'ical-1',
+    isEnabled: true,
+    issueProviderKey: 'ICAL',
+    icalUrl: 'https://example.com/cal.ics',
+    ...over,
+  }) as IssueProvider;
+
+const pluginProvider = (over: Partial<IssueProviderPluginType> = {}): IssueProvider =>
+  ({
+    id: 'plugin-1',
+    isEnabled: true,
+    issueProviderKey: 'plugin:caldav-calendar-provider',
+    pluginId: 'caldav-calendar-provider',
+    pluginConfig: {},
+    ...over,
+  }) as IssueProvider;
+
+describe('getCalendarProviderColor', () => {
+  it('returns the explicit color of an iCal provider', () => {
+    expect(getCalendarProviderColor(icalProvider({ color: '#4caf50' }))).toBe('#4caf50');
+  });
+
+  it('returns the color from a plugin provider pluginConfig', () => {
+    expect(
+      getCalendarProviderColor(pluginProvider({ pluginConfig: { color: '#abcdef' } })),
+    ).toBe('#abcdef');
+  });
+
+  it('falls back to a HSL color when an iCal provider has no color', () => {
+    expect(getCalendarProviderColor(icalProvider())).toMatch(
+      /^hsl\(\d{1,3}, 60%, 55%\)$/,
+    );
+  });
+
+  it('falls back to a HSL color when a plugin provider has no color', () => {
+    expect(getCalendarProviderColor(pluginProvider())).toMatch(
+      /^hsl\(\d{1,3}, 60%, 55%\)$/,
+    );
+  });
+
+  it('derives the fallback deterministically from the provider id', () => {
+    const a = getCalendarProviderColor(pluginProvider({ id: 'same-id' }));
+    const b = getCalendarProviderColor(pluginProvider({ id: 'same-id' }));
+    expect(a).toBe(b);
+  });
+
+  it('gives different ids different fallback hues', () => {
+    const a = getCalendarProviderColor(pluginProvider({ id: 'provider-aaa' }));
+    const b = getCalendarProviderColor(pluginProvider({ id: 'provider-bbb' }));
+    expect(a).not.toBe(b);
+  });
+
+  it('ignores an empty-string color and falls back', () => {
+    expect(getCalendarProviderColor(icalProvider({ color: '' }))).toMatch(/^hsl\(/);
+  });
+});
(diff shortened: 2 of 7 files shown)


johannesjo:
Thank you very much!

johannesjo:
Thanks for this — the plugin-provider integration is a real gap. Multi-agent review surfaced a few things worth addressing before merge.

## Blockers

**1. Color asymmetry: filter dot vs. iCal grid events.** The PR description promises "filter dot and grid event always agree on a color for the same provider," but this only holds for plugin providers. For iCal providers without an explicitly configured color:
- Filter dot renders the HSL fallback (via `calProviderColor(provider)` in [`schedule.component.html`](../blob/feat/schedule-filter-plugin-providers/src/app/features/schedule/schedule/schedule.component.html#L51-L55))
- Grid events stay uncolored — [`calendar-integration.service.ts:407`](../blob/feat/schedule-filter-plugin-providers/src/app/features/calendar-integration/calendar-integration.service.ts#L407) still uses raw `calProvider.color` (existing spec at `calendar-integration.service.spec.ts:1035` confirms `falsy` color in this case)

Two clean fixes:
- **Route iCal through `getCalendarProviderColor` too** in `requestEvents$` — matches the stated goal; existing uncolored iCal feeds get a deterministic tint (a visual change worth calling out in the description).
- **Restore the `@if (provider.color)` guard** on the dot and drop the HSL fallback — no iCal behavior change; plugin events keep their explicit color or stay neutral.

Either is fine; the current state silently picks neither for the most common iCal-without-color case.

**2. Untrusted `pluginConfig.color` flows into `[style.background]`.** [`get-issue-provider-color.ts:13-15`](../blob/feat/schedule-filter-plugin-providers/src/app/features/issue/mapping-helper/get-issue-provider-color.ts#L13-L15) returns the plugin-supplied value verbatim. A hostile plugin can set `pluginConfig.color = 'url(https://evil.example/beacon.png)'`, which is a valid `background` shorthand value — every render of the filter menu and every plugin event chip triggers the outbound request. Not RCE (modern CSS blocks `expression()` / `javascript:` in style values), but it's a passive beacon / IP exfiltration vector.

Suggested fix: validate before returning, e.g.

```ts
const SAFE_COLOR = /^(#[0-9a-f]{3,8}|rgb\(|rgba\(|hsl\(|hsla\(|[a-z]+)/i;
// ...
if (typeof c === 'string' && SAFE_COLOR.test(c)) return c;
```

(The HSL fallback path is already safe — it's constructed from a numeric hue.)

## Warnings

**3. `combineLatest` missing `distinctUntilChanged(fastArrayCompare)`.** The sibling stream at [`calendar-integration.service.ts:78-92`](../blob/feat/schedule-filter-plugin-providers/src/app/features/calendar-integration/calendar-integration.service.ts#L78-L92) wraps both branches in `distinctUntilChanged(fastArrayCompare)` precisely so that unrelated issue-provider edits (saving a Jira config, toggling GitHub) don't churn downstream. The new component-side composition in [`schedule.component.ts:96-108`](../blob/feat/schedule-filter-plugin-providers/src/app/features/schedule/schedule/schedule.component.ts#L96-L108) drops both gates — one-line symmetry fix.

**4. `calProviderLabel` reimplements `sanitizeIcalUrlForDisplay`.** [`get-issue-provider-tooltip.ts:10-28`](../blob/feat/schedule-filter-plugin-providers/src/app/features/issue/mapping-helper/get-issue-provider-tooltip.ts#L10-L28) already does `webcal://` normalization and hostname extraction, plus handles `file://` (basename) and opaque schemes (`javascript:`, `data:`, `blob:` → `'iCal'`). The new inline version at [`schedule.component.ts:119-127`](../blob/feat/schedule-filter-plugin-providers/src/app/features/schedule/schedule/schedule.component.ts#L119-L127) lacks the opaque-scheme collapse — a plugin whose first non-secret config string is `javascript:alert(1)` would render that text in the filter menu. Suggest layering:

```ts
readonly calProviderLabel = (p: IssueProvider): string => {
  const raw = getIssueProviderTooltip(p);
  return /^(https?|webcals?|file):\/\//i.test(raw) ? sanitizeIcalUrlForDisplay(raw) : raw;
};
```

## Nits

- **5. Filename / export mismatch:** [`get-issue-provider-color.ts`](../blob/feat/schedule-filter-plugin-providers/src/app/features/issue/mapping-helper/get-issue-provider-color.ts) only exports `getCalendarProviderColor`. Sibling `get-issue-provider-tooltip.ts` ↔ `getIssueProviderTooltip` matches; this one should too — rename file to `get-calendar-provider-color.ts`.
- **6. Method bindings in `@for`:** `calProviderColor(provider)` and `calProviderLabel(provider)` re-execute on every CD cycle for every menu row. Bounded (rows are few), but precomputing color/label into the materialized array would pair nicely with the `distinctUntilChanged` fix and eliminate the per-CD hash + `new URL()` work.

CI failure (`repeat-task-day-change-bug-6230`) is a known flake — unrelated to this PR; recently stabilized in master at 46ac873570.

Happy to take this in any direction — the underlying feature is good and the tests are solid; mostly want to land it without the asymmetry or the open beacon channel.

sespiros:
Thanks for the thorough review and sorry for revisiting this that late. Rebased onto v19.1.0 and addressed everything:

1. **Color asymmetry:** went with the first option. `requestEvents$` now routes iCal events through `getCalendarProviderColor` as well, so the filter dot and the grid always agree. Uncolored iCal feeds get the deterministic tint instead of staying neutral; called out as a behavior change in the description.
2. **Untrusted `pluginConfig.color`:** the helper only accepts plain color literals (hex, `rgb()`/`hsl()` with numeric bodies, named colors); anything else falls back to the HSL tint. I used a full-match regex rather than the suggested prefix one, since `[a-z]+` without an end anchor would still have matched `url(...)`. Spec covers `url()`, `var()`, a trailing declaration, and a non-string.
3. **`distinctUntilChanged`:** both branches gated with `fastArrayCompare`. While there, I also mirrored the `registrationChanges$` trigger that `calendarEvents$` picked up after this PR was opened; without it a plugin that registers after the store hydrates never shows up in the menu.
4. **Label sanitizer:** `calProviderLabel` now builds on `sanitizeIcalUrlForDisplay`. I match any scheme prefix rather than only http/webcal/file, because a `javascript:` label would otherwise still render verbatim; when the sanitizer collapses an opaque scheme, a plugin row falls back to its provider key instead of the generic `iCal` label.
5. **Rename:** `get-calendar-provider-color.ts` to match the export.
6. **Method bindings in `@for`:** rows (id, color, label) are precomputed in a `computed()`; the template no longer calls methods in the loop.

Issue #4068, opened 2025-03-02
Add option to take a break in break reminder


Lauloque:
## Problem Statement
I'm frustrated with how the break reminder lacks an option to actually take a break:

![Image](https://github.com/user-attachments/assets/15185845-3d5b-40f8-a86d-5938936a5a42)

I can only say I already did take one or snooze it for a while.

## :grey_question: Possible Solution
How about this:

1. Add a "take a break" option in that dialogue:  

    ![Image](https://github.com/user-attachments/assets/4e62aec5-dd82-42d3-b44c-355c669b36b8)

    It could also be accessible in the top bar near the task start/pause button.

    ![Image](https://github.com/user-attachments/assets/bbdcd28d-147b-4891-a099-dbab2862ed61)

3. When hit, starts a timer, which can be set in the break reminder settings.
4. While break timer running, cease all tasks counters, reminders and whatnot. Just be on break for the time being
5. Once the timer is done, let the user know it's time to get back to work.
6. Once the user stops the break, reset the "time without break" timer to zero:    
    ![Image](https://github.com/user-attachments/assets/adac7a53-cb98-4d7a-bd58-28fd5e20db6b)

Bonus suggestions:

During the break, show a graphical live "elapsed" vs "remaining" break time counter on top of the whole GUI. Visually, it could be either a simple progress bar like we already have on running tasks, but personally I'd prefer the round progress bar like we have to set estimated/elapsed time:

![Image](https://github.com/user-attachments/assets/3e112a82-609c-4457-bb87-cc99cd989942)

Optionally, when estimated breaktime is reached, send (audio) notifications (in case I'm sipping my coffee too hard to notice I should go back to work)

Optionally again, when breaktime is over, show the user which tasks they were on before, to provide a quick access to start its counter again (and also because I tend to forget what I was doing before a break). Starting that task counter from there automatically stops the break counter and resets the "time without break" counter too.

![Image](https://github.com/user-attachments/assets/02f065ef-3ef5-40e5-9b1f-b68761d10059)

Maybe also have an option to register the break time in the end-of-day summary.

## :arrow_heading_up: Describe alternatives you've considered
🤷

## :heavy_plus_sign: Additional context

🤷‍♂️

johannesjo:
That is an interesting suggestion! Thank you very much!

tompropst:
@johannesjo, is there a description of how you envisioned the break reminder workflow? When the reminder goes off, do you:
[...]

PR #7838, opened 2026-05-28
feat(take-a-break): add Start break action to break reminder

johannesjo:
## Summary

Replaces the break reminder's **"I already did"** button with **"Start break"**, which:

1. dismisses the reminder,
2. pauses the active task (`unsetCurrentTask`), and
3. resets the "time without a break" counter.

This is the minimal, honest-break path proposed by @zenoprax in #4068: the break becomes an untracked gap instead of forcing a fabricated 5-minute break entry or leaving the task stopwatch running while you're away. Intentionally scoped down — no break countdown/progress UI, since that overlaps with the existing Focus Mode feature.

## Behavior notes

- **No more flat 5-minute break log.** The old button wrote an arbitrary 5 min to the break-time metrics/daily-summary; "Start break" doesn't (those fields stay manually editable, and real break time still comes from the idle dialog's "Break" option). This is deliberate, per the accuracy motivation in the issue.
- **Pauses an active Focus/Pomodoro session** (the task going set→null triggers `pauseFocusSession`); `pausedTaskId` allows resuming. The task stays selected, so the top-bar ▶ button resumes it.
- The label reuses the existing, already-translated `T.F.FOCUS_MODE.START_BREAK` string.

## Test plan

- [ ] In break-reminder settings, lower `takeABreakMinWorkingTime` to a small value
- [ ] Track a task until the break reminder banner appears
- [ ] Click **Start break** → banner closes, the active task pauses (top-bar button flips to ▶), and the timer resets
- [ ] Confirm the banner does not reappear while paused, and reappears only after resuming and working another full interval
- [ ] Confirm **Snooze** still suppresses the banner for the snooze duration as before

Refs #4068

### DIFF
--- a/src/app/features/take-a-break/take-a-break.service.ts
+++ b/src/app/features/take-a-break/take-a-break.service.ts
@@ -26,7 +26,6 @@ import { GlobalConfigState, TakeABreakConfig } from '../config/global-config.mod
 import { T } from '../../t.const';
 import { NotifyService } from '../../core/notify/notify.service';
 import { UiHelperService } from '../ui-helper/ui-helper.service';
-import { WorkContextService } from '../work-context/work-context.service';
 import { Tick } from '../../core/global-tracking-interval/tick.model';
 import { ofType } from '@ngrx/effects';
 import { idleDialogResult, triggerResetBreakTimer } from '../idle/store/idle.actions';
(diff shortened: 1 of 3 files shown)


zenoprax:
Nice, works as intended. The preview deployments are so useful for this kind of quick validation/testing.

johannesjo:
Thank you very much for checking!

PR #7839, opened 2026-05-28
Android soft-keyboard: fix dark-theme white flash on resize

johannesjo:
## Summary

Primary (this session's work on the Android soft-keyboard resize):
- **Fix the white flash on keyboard resize.** The Android WebView surface defaulted to white, so the `adjustResize` keyboard animation briefly exposed it before the page repainted — jarring in dark theme. The surface is now painted in the theme background via a `values`/`values-night` `windowBackground` resource (cold-start default) plus a new `NavigationBar.setWebViewBackgroundColor` push kept in sync with the JS theme (handles runtime light/dark switches, since the activity isn't recreated on `uiMode` change). Compile-verified with `./gradlew :app:compilePlayDebugKotlin`.
- **Reduce per-frame work during the keyboard slide.** The native `OnGlobalLayoutListener` pushes `isKeyboardShown$` on every layout pass; `distinctUntilChanged()` collapses it to actual show/hide transitions so we stop rewriting `<body>` classes + re-triggering change detection each frame.
- **Plan for the deeper resize *choppiness* fix** (`docs/plans/2026-05-28-android-keyboard-resize-jank.md`): researched + multi-reviewed. Root cause is `adjustResize` resizing the WebView window per frame (WebView is excluded from Chrome 108's visual-viewport fix). The fix is a runtime-gated `adjustNothing` + VirtualKeyboard approach — **not implemented here**, it's device-gated and tracked as a follow-up.

Also carried on this branch (earlier commits / via merges, not from this session):
- `fix(theme): stabilize Android keyboard-height tracking`
- `fix(android): keep foreground services alive across task removal`
- `test(e2e): raise timeouts for two CI-flaky waits`
- merges of `feat/add-task-bar-position-for-android-jumps` and `feat/issue-7818`

## Test plan
- [x] `./gradlew :app:compilePlayDebugKotlin` passes (new plugin method + `windowBackground` resource compile)
- [ ] On-device, dark theme: open/close the soft keyboard repeatedly → no white flash, typing works normally
- [ ] On-device, light theme: same, surface matches
- [ ] Toggle dark/light at runtime with keyboard open → WebView surface follows
- [ ] Resize choppiness itself is **not** addressed here — see the plan doc for the follow-up


johannesjo:

Issue #7244, opened 2026-04-16
💡 General undo


auino:
### Problem Statement

Hi, I'm moving from OmniFocus.
I noticed that it's not possible to undo actions: neither in the shortcuts settings it's possible to find such option.
Also, in the mobile version of OmniFocus this is feasible by shaking the device and it's a very friendly implementation.

### Possible Solution

To add a new general shortcuts option dedicated to undo last action(s).

### documentation gap

- [ ] ... related docs should be updated as part of the same PR whenever possible.
- [ ] ... I can help improve the documentation.

Sofiaaleixo:
Hello! @Zinko0 and I would love to take on this feature.

We already have a few ideas on how to approach the implementation, particularly for PC/web environments. We'd be happy to collaborate and align with the maintainers vision before moving forward.

Looking forward to discussing this further!

Sofiaaleixo:
Hi! Could you please assign this issue to us?

And just to confirm — should we proceed with our own implementation approach, or would you prefer to discuss the idea first?

johannesjo:
Thank you very much @Sofiaaleixo ! That is great! It probably makes sense to discuss the approach first. The potential implications of this are quite complex both technical as well as UX wise. We should make a good plan before implementing!

Zinko0:
Hello @johannesjo ! How would you like to discuss the implementation?  In this thread or in Discord? 

Either way, we can already tell you the workflow that we have in mind:

1.  Define the Undoable Actions & Compensating Operations: This is the topic that we would most like to discuss with you!
Since this is a new and complex feature, we would like to keep it simple for now and implement the undo/redo for operations like create/delete/move for tasks/projects/folders/tags. This would not include updates. For example, if I create Task A and then update it, undoing the operation would simply delete Task A.
 
2.  Create an UndoRedoService : The service would retrieve the latest operation from the log and apply the corresponding undo/redo action.
 
3. NgRx Wiring: Minimal NgRx Add the minimal NgRx integration required so undo/redo actions can be dispatched through the UndoRedoService.
 
4. KeyBoard shortcuts Implementation: Add keyboard listeners for the undo/redo. In this stage we could already test and see the results of undo/redo!
     
5. UI feedback: Similar to other operations, we would like to display a small pop-up, indicating wether the undo/redo was a success or a failure.

6. Validation & Safety checks: Apply checks such as: entity exists, parent entity exists, is it a undoable operation.

[...]

Issue #7330, opened 2026-04-22
🚨 Sync conflict error on Windows after hibernate when SP is left running


vinsse2001:
### Version Used

18.1.3 (Windows 10, desktop app). Also observed on 17.2.4 (same setup, more severe behavior)

### Expected Behavior

When Super Productivity is left open on a Windows PC that goes into 
hibernation, and changes are made on other devices in the meantime, 
SP should be able to reconcile the sync state cleanly on wake without 
errors.

### Current Behavior

When SP is left open (and synced) before hibernation, and changes are 
made on other devices (Linux / Android) during the hibernation period, 
SP shows a sync error on the next startup after wake:

- On v18.1.3: a sync conflict message appears, stating that issues were 
  automatically corrected. Less alarming, but still incorrect behavior.
- On v17.2.4: a more severe error appeared, warning of potential data loss 
  during repair.

**Workaround**
Closing SP manually after syncing, before triggering hibernation, prevents 
the error entirely. On next launch, SP correctly retrieves all changes made 
on other devices.

**Root cause hypothesis**
When Windows hibernates with SP open, the app's sync state (last known 
remote revision, local timestamps) is frozen in memory. Upon wake, SP 
resumes from this stale state without re-checking the remote file, causing 
a conflict with the changes accumulated on other devices.

### Steps to Reproduce

1. Open SP on Windows and perform a manual sync
2. Leave SP open and trigger hibernation (not simple sleep)
3. On another device (Linux or Android), make changes in SP and sync
4. Wake the Windows PC
5. SP shows a sync conflict/error on startup

**Your Environment**
- Version: 18.1.3 (Windows 10, desktop .exe)
- Other devices: Linux Mint 24 (Flatpak 17.2.4) + Android Fairphone 5 (18.1.3)
- Sync: Dropbox (3 devices)

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### documentation gap

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.
- [ ] I can help improve the documentation.

johannesjo:
Thanks for the detailed reproducer !



[...]

PR #7842, opened 2026-05-28
feat(undo/redo): add operation-based undo/redo infrastructure

Zinko0:
## Problem

The application currently lacks a generic undo/redo mechanism,
making destructive or accidental task operations irreversible.
This reduces usability and makes it harder for users to recover
from unintended changes. 
This Problem was metioned in issue [#7244](
https://github.com/super-productivity/super-productivity/issues/7244)

Additionally, there was no reusable infrastructure for handling
undoable operations across domains in a consistent way.

## Solution

This PR introduces generic operation-based undo/redo infrastructure backed by
the persisted operation log already used in this repository.

Task operations are the first supported domain, including:

- task creation
- task deletion
- task updates
- subtask creation

The implementation adds:

- undo/redo state management and orchestration
- compensating operations for reverting actions
- undo payload snapshot capture before state mutations
- keyboard shortcuts and snackbar integration
- safeguards to prevent compensating actions from polluting the undo history
- focused test coverage for reducers, services, and undo payload generation

This lays the foundation for extending undo/redo support to
additional domains in the future.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/packages/plugin-dev/caldav-calendar-provider/package-lock.json
+++ b/packages/plugin-dev/caldav-calendar-provider/package-lock.json
@@ -2029,9 +2029,9 @@
       }
     },
     "node_modules/undici": {
-      "version": "7.27.0",
-      "resolved": "https://registry.npmjs.org/undici/-/undici-7.27.0.tgz",
-      "integrity": "sha512-+t2Z/GwkZQDtu00813aP66ygViGtPHKhhoFZpQKpKrE+9jIgES+Zw+mFNaDWOVRKiuJjuqKHzD3B1sfGg8+ZOQ==",
+      "version": "7.28.0",
+      "resolved": "https://registry.npmjs.org/undici/-/undici-7.28.0.tgz",
+      "integrity": "sha512-cRZYrTDwWznlnRiPjggAGxZXanty6M8RV1ff8Wm4LWXBp7/IG8v5DnOm74DtUBp9OONpK75YlPnIjQqX0dBDtA==",
       "dev": true,
       "license": "MIT",
       "engines": {
--- a/packages/super-sync-server/package.json
+++ b/packages/super-sync-server/package.json
@@ -47,7 +47,7 @@
     "dotenv": "^17.2.3",
     "fastify": "^5.8.5",
     "jsonwebtoken": "^9.0.2",
-    "nodemailer": "^8.0.9",
+    "nodemailer": "^9.0.0",
     "uuidv7": "^1.1.0",
     "zod": "^4.4.3"
   },


Zinko0:
Co-authored by @Sofiaaleixo of course!

johannesjo:
Thank you very much for this!

This is a substantial, carefully-built feature, and it's clear you understood the sync model well: compensating ops are persisted+synced but correctly skipped on re-capture, remote ops are excluded from the undo stack, the undo state is kept in-memory only (not in `MODEL_CONFIGS`), and the keyboard config is wired up cleanly. Really nice groundwork.

Before merge, there are a few things I'd like to work through. I'll group them by priority.

## Blocking

**1. Undo is blocked whenever a non-undoable op is on top of the stack.**
The capture effect pushes *every* non-compensating local op onto the undo stack (`operation-log.effects.ts:296`), but only 4 task types are undoable, and only the topmost op can ever be undone. The problem: `syncTimeSpent` is persistent and fires every 5 min during tracking plus on every task-stop, plus reorders, tag/project edits, scheduling, etc. So in normal use a non-undoable op regularly lands on top and Ctrl+Z just shows "Undo is not supported for ..." even though an undoable task op sits right underneath.
Could we filter the stack at the capture site so only undoable action types are pushed (deriving the allow-list from the registry)?

**2. Undo snapshots are embedded into the persisted + synced op-log.**
`addUndoPayloadToActionPayload` injects the snapshot into `op.payload` before `appendWithVectorClockUpdate`, and it's never stripped before upload. So it's written to IndexedDB, uploaded to SuperSync, and fanned out to every device. For deletes this roughly 2-3x's the op (full `TaskWithSubTasks` plus duplicate `deletedTaskEntities` plus tag/project ordering arrays); for text edits it stores the *old* value of every changed field, so editing notes/title now syncs old and new content. On `master` this restore data lived only in memory and never synced. Since undo is a purely local concern that other devices can't use, this is a permanent regression on the sync substrate (size and privacy footprint) that's hard to walk back once ops are out there.
Could we keep the snapshot in the in-memory undo state (keyed by op id, as the old delete-undo did) and strip `UNDO_OPERATION_PAYLOAD_KEY` before persist/upload? The redo path already shows the synced op alone is enough to reconstruct the action.

**3. Double-compensation race on held/rapid Ctrl+Z.**
`undo()` awaits the selector and the registry *before* dispatching the stack pop, with no in-flight guard, and the shortcut handler calls it without `await` or an `ev.repeat` check. Holding Ctrl+Z (key auto-repeat) lets two calls read the same top op and dispatch the compensating action twice (e.g. `restoreDeletedTask` twice).
A re-entrancy guard in the service plus `if (ev.repeat) return;` in the shortcut handler should cover it.

## Worth addressing

- **Stale redo after subtask coalescing:** the `isInitialSubTaskUpdate` branch in `undo-redo.reducer.ts` doesn't clear `redoStack` (the normal branch does), so a new edit can leave a redoable forward op around.
- **No reset on `SYNC_IMPORT`/`BACKUP_IMPORT`:** the reducer only handles `addToUndoStack`/`undo`/`redo`. After an import that drops concurrent ops, `_compensateTaskDelete` restores purely from the snapshot without re-validating, so it can re-insert a task the import deliberately dropped.
- **`isCompensatingAction` filter is an unenforced convention:** it's hand-added to ~10 effect sites with nothing lint-enforcing it (unlike `no-actions-in-effects`). A future effect on `updateTask`/`deleteTask` that forgets it would re-fire and emit a phantom op. Could we centralize this as a shared operator next to `filterLocalAction()` and ideally add a lint rule?
- **Hot path:** the per-op `OpLog.normal(...)` runs on every persisted op and is active at the default log level, worth demoting or removing.

## Minor / cleanup

- Remove the leftover `console.debug('[ShortcutService] Redo shortcut detected...')` in `shortcut.service.ts`.
- User-facing labels (`'Undo task creation'`, `` `Redo ${actionType}` `` renders raw `Redo TASK_SHARED_UPDATE`) bypass translation via `isSkipTranslate: true`; these should go through `T`/`TranslateService`.
- `extractActionPayload` is duplicated (with slightly different logic) in the reducer and registry; a canonical version is already exported from `op-log/core/operation.types.ts`. Relatedly, the redo handlers could reuse the existing `convertOpToAction` (`op-log/apply/operation-converter.util.ts`) instead of hand-casting payloads; it also handles the `#7330` id canonicalization the hand-rolled version skips.
- `UndoValidatorService` duplicates the registry's handler keys (two sources of truth for "what's undoable").
- The `t.const.ts` diff is almost entirely trailing-comma reformatting; regenerating it cleanly would shrink the diff to just the new keys.

## A bigger-picture question

Redo (and the subtask create+update coalescing that mostly exists to support it) adds a fair amount of surface that isn't required by #7244. Was redo a deliberate scope decision? If undo-only would be acceptable for a first cut, it would remove the coalescing edge cases and a good chunk of the registry, but happy to keep it if it's on the roadmap.

Thanks again, this is a strong foundation and I'd love to get it in once the sync-layer and reliability points above are sorted. Happy to discuss any of these.

Sofiaaleixo:
Thank you so much for the quick feedback, and I'm sorry for the delayed response.

We've already started working on points 1 and 3.

While reviewing the other comments, we also realized that removing redo would resolve a significant portion of the issues raised, particularly around the redo stack, the subtask coalescing logic, and several related edge cases.

Since redo was not part of the original scope of #7244, we've decided to simplify the implementation and move forward with an undo-only approach for this first iteration. The plan is to keep an Undo button that appears after a Ctrl+Z action, allowing the user to restore the operation that was just undone, without maintaining a full redo stack.

We believe this keeps the implementation focused on the original requirement while significantly reducing complexity and improving reliability. We'll also revisit the remaining comments in light of this simplification and update the PR accordingly.

Thanks again for the thorough review and detailed feedback—it has been very helpful.

johannesjo:
Makes sense to me @Sofiaaleixo ! Thank you very much for working on this!

johannesjo:
A closer follow-up pass on the current branch (still at `1c13bb5c`, so this sits on top of the undo-only direction we agreed on — nothing here changes that plan). The three blockers from my earlier review still apply as-is; below are a few **additional** things I'd fold into the undo-only revision.

### Worth addressing

**1. The "task created" snack loses its "Go to task" action (`task-ui.effects.ts`).**
This one is unrelated to undo and affects all users. That snack only fires when the new task is *not* visible on the current page (`isTaskVisibleOnCurrentPage === false`), so its single button exists specifically to let you reach an off-screen task — and the PR repurposes it to *delete* that task ("Undo"), and also drops the `hideAddTaskBar()` call. Undo is already on Ctrl+Z (and, post-revision, its own button), so I'd keep "Go to task" here rather than overload the create snack.

**2. Undo mutates the stack before the compensation is confirmed (`undo-redo.service.ts`).**
The order is `dispatch(undo())` (pops undo→redo) → `dispatch(compensatingAction)` → `dispatch(undoRedoSuccess)`. If the compensating action's effect/persistence fails, the op has already left the undo stack (and, today, a phantom entry is on the redo stack). `getCompensatingOp` validates entity existence first so the window is small, but it'd be safer to transition the stack only after a confirmed apply, or roll back on failure. This is distinct from the auto-repeat re-entrancy blocker.

### Minor

**3. Compensating ops also get an embedded `undoPayload`.**
`undoOperationPayloadMetaReducer`'s builders match on `action.type` only, so undoing a create dispatches a compensating `deleteTask`, the delete builder runs, and a full restore snapshot gets embedded into that compensating op's synced payload — even though a compensating op is never itself undone. Skipping capture when `isCompensatingAction(action)` is a cheap fix and trims the synced footprint a bit further on top of blocker 2.

**4. Update-undo sets absent fields to `undefined` instead of removing them.**
`snapshotPreviousValuesToChanges` maps `wasPresent: false → undefined`; `@ngrx/entity`'s `updateOne` merges changes, so the key stays present with value `undefined` rather than being deleted. For a field that didn't exist before the edit, undo yields a subtly different entity (and a noisier op) than the original state.

**5.** `app.component.ts` has a whitespace-only change — worth dropping to keep the diff scoped.

Again, really nice groundwork — these are all about tightening the edges. Happy to dig into any of them.

Zinko0:
Hello @johannesjo!

Sorry, I accidentally closed the PR while preparing to push a patch addressing the review comments. We haven't lost interest in the feature and still intend to work on it and address the feedback.

Apologies for the confusion!

Issue #7818, opened 2026-05-27
🚨 app cannot run in the background


akiroatakashi-ship-it:
### Version Used

'v 18.7.0

### Current Behavior

When i start a timer and clear the the tab from recent tabs the timer notification goes away . In pomodoro mode the pomodoro timer resets.  **All because the app is not running in the background.**

**I make sure to turn on background run , turn off battery optimization.

### Expected Behavior

When i clear the tab from recent tabs the app should run in the background and the timer notification should continue.

### Steps to Reproduce

1. Start the timer .
2. Clear the tab from recent tabs.
3. The timer notification goes away ( app cannot run in the background)

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

PR #7866, opened 2026-05-30
fix(android): recover focus/pomodoro timer after app swipe (#7855)

### LINKED ISSUE #7855: 🚨 running in background
### Version Used

'v18.8.0 android 

### Current Behavior

#7818 now time tracking and pomodoro works after clearing from recent tab . But there is another problem . For time tracking after removing from recent tab  then going to the app everything works well timer in the notification also works fine . But for pomodoro timer removing from recent it works then when i visit the app again the pomodoro timer resets the notification closes and the pomodoro timer information loses.

### Expected Behavior

When i visit the app again the pomodoro timer should run as usual in the app and also in the notification 

### Steps to Reproduce

1. Go to the app
2. Start pomodoro timer 
3. Clear it from recent tabs
4. Visit the app again
5.  Now the pomodoro timer information gone . The timer reset , notification closes.

### Console Output

```shell

```

### Documentation

johannesjo:
## Problem

Fixes #7855. After #7818 kept the foreground services alive across task removal, **time tracking** survives an app swipe-from-recents + reopen, but the **focus/pomodoro timer** does the opposite: the in-app timer resets and the foreground-service notification closes.

## Root cause (two parts)

1. On cold start, `syncFocusModeToNotification$` computed `wasFocusModeActive` from the `startWith(null)` seed as `(undefined \!== null) === true`, so the first emission of an idle store fired `stopFocusModeService()` and tore down the surviving native notification.
2. Focus mode had **no native read-back / recovery** (unlike time tracking's `getTrackingElapsed` + recovery loop), so the idle Angular store never re-adopted the still-running native countdown.

## Fix — mirror the proven time-tracking recovery

- **`FocusModeForegroundService`**: mirror live timer state into the companion object (`durationMs`/`remainingMs`/`isBreak`/`isPaused`/`lastUpdateTimestamp`) + a `liveRemainingMs()` accessor; cleared on stop so an ended session can't be re-adopted.
- **`JavaScriptInterface.getFocusModeElapsed()`**: returns JSON or `"null"`. Task title is intentionally omitted — no user content crosses the bridge or reaches the exportable log.
- **`restoreFocusSessionFromNative`** action + reducer: rebuild the timer (`elapsed = duration - remaining`, `startedAt = now - elapsed`; Flowtime = duration 0) and derive `mode` from the native duration so a recovered Flowtime session isn't auto-completed on the next tick.
- **`recoverFocusSession$`**: recovers on cold start + resume, gated on an idle store and settled hydration. The `wasFocusModeActive` cold-start kill is fixed (`\!\!prev && prev.timer.purpose \!== null`).

## Testing

- Kotlin compiles: `:app:compilePlayDebugKotlin` → BUILD SUCCESSFUL.
- Unit tests: focus-mode reducer (69, incl. a `tick`-after-restore regression proving a recovered Flowtime session keeps running) + new `android-focus-mode.effects.spec.ts` (9); existing focus/tracking effect specs unchanged and green.
- Reviewed across correctness/security/architecture/alternatives/performance/simplicity; privacy rule (no user content in bridge/logs) verified.

⚠️ **Not yet verified on a device/emulator** — please run the manual repro:

```bash
npm run buildFrontend && npx cap sync android
cd android && ./gradlew :app:installPlayDebug
adb logcat | grep -E "FocusModeService|AndroidFocusModeEffects|Recovering focus"
```

Repro: start a Pomodoro session → swipe the app from recents → reopen. **Expected:** the in-app timer keeps counting and the notification persists. The `AndroidFocusModeEffects: Recovering focus session from native` log line confirms the recovery effect fired.

### Known limitations
- A mid-session **long break** recovers as a short break (the native service only tracks `isBreak`, not long-vs-short) — cosmetic (notification title).
- Recovery is non-intrusive: the header focus button shows the running timer; the full-screen overlay is not auto-opened on reopen.


johannesjo:

PR #7881, opened 2026-05-30
feat(caldav): make VTODO poll interval configurable

### LINKED ISSUE #7880: feat(caldav): make the core VTODO poll interval configurable (hard-coded at 10 min)
The core CalDAV VTODO provider polls every 10 minutes (`CALDAV_POLL_INTERVAL`), which is sluggish for an actively-shared/household task list — the bundled CalDAV Events plugin polls every 60 s. I'd like to expose a per-provider poll interval, mirroring the existing `pollIntervalMinutes` pattern already used by the Nextcloud Deck provider (a `get pollInterval()` deriving from the cached cfg, falling back to the const). PR to follow.

JongoDB:
Exposes a per-provider **Poll interval (minutes)** for the core CalDAV (VTODO) provider, mirroring the existing pattern in the Nextcloud Deck provider.

- `CaldavCfg.pollIntervalMinutes?: number` — optional, so existing configs fall back to the 10-min `CALDAV_POLL_INTERVAL`
- `pollInterval` becomes a getter deriving from the cached cfg (via a `_getCfgOnce$` override, same as `NextcloudDeckCommonInterfacesService`)
- Advanced-config form field (number, min 1) + `en.json` / `t.const.ts` entries
- Unit tests for the getter (fallback + derived)

Motivation: the bundled CalDAV Events plugin polls every 60s, but the core VTODO provider was fixed at 10 min — sluggish for an actively-shared list.

`npm run checkFile` clean on all changed `.ts`; spec passes (21/21).

Closes #7880


johannesjo:
Thank you very much! <3

PR #7898, opened 2026-05-31
fix(header): keep actions visible with long titles

### LINKED ISSUE #7477: 🚨 Visual glitch in project view.
### Version Used

v18.4.4 (Android 16)

### Current Behavior

Habit counter button a bit out of the screen.

Image

### Expected Behavior

As in the Today view:

Image

### Steps to Reproduce

Just open a project view. 

### Console Output

```shell

```

### Documentation

RoySerbi:
﻿Fixes #7477.

Long project titles could take too much space in the mobile header. The title text was visually truncated, but the flex item was still allowed to keep its intrinsic width, which pushed the right-side header actions past the viewport. That made the last simple-counter / habit button appear cut off in project views.

This makes the title area shrinkable (`min-width: 0`) and keeps the action groups from shrinking, so the title truncates before the buttons overflow.

Tested with:

- `npx eslint "src/app/core-ui/main-header/page-title/page-title.component.ts" "src/app/core-ui/main-header/page-title/page-title.component.spec.ts"`
- `npx stylelint "src/app/core-ui/main-header/main-header.component.scss"`
- `npm run test:file -- src/app/core-ui/main-header/page-title/page-title.component.spec.ts --watch=false --browsers ChromeHeadless --no-code-coverage`


johannesjo:

PR #7905, opened 2026-06-01
Allow dragging tasks into subtask lists

### LINKED ISSUE #905: Adding a sub-task via drag-drop

keilogic:
Fixes #905.

Summary:
- allow eligible top-level tasks to be dropped onto a task subtask list
- add a persistent convert-to-subtask action that removes the task from top-level project/tag lists and attaches it to the target parent
- keep parent/subtask time aggregation and task-list drag tests covered

Tests:
- npx ng test --watch=false --include=src/app/root-store/meta/task-shared-meta-reducers/task-shared-crud.reducer.spec.ts --include=src/app/features/tasks/task-list/task-list.component.spec.ts --include=src/app/root-store/meta/task-shared-meta-reducers/section-shared.reducer.spec.ts
- npx prettier --check src/app/features/tasks/task-list/task-list.component.ts src/app/features/tasks/task-list/task-list.component.spec.ts src/app/root-store/meta/task-shared-meta-reducers/section-shared.reducer.ts src/app/root-store/meta/task-shared-meta-reducers/section-shared.reducer.spec.ts src/app/root-store/meta/task-shared-meta-reducers/task-shared-crud.reducer.ts src/app/root-store/meta/task-shared-meta-reducers/task-shared-crud.reducer.spec.ts src/app/root-store/meta/task-shared.actions.ts
- npm run lint:ts (passes with 3 unrelated pre-existing warnings outside touched files)
- git diff --check origin/master...HEAD

### DIFF
--- a/e2e/tests/migration/pre-migration-dialog.spec.ts
+++ b/e2e/tests/migration/pre-migration-dialog.spec.ts
@@ -1,4 +1,5 @@
 import { test, expect } from '../../fixtures/test.fixture';
+import { skipOnboardingForE2E } from '../../utils/waits';
 
 /**
  * Pre-migration Dialog E2E Tests
@@ -118,6 +119,7 @@ test.describe('@migration Pre-migration Dialog', () => {
     });
 
     const page = await context.newPage();
+    await page.addInitScript(skipOnboardingForE2E);
 
     try {
       // Navigate to app
--- a/src/app/core-ui/drop-list/drop-list.service.spec.ts
+++ b/src/app/core-ui/drop-list/drop-list.service.spec.ts
@@ -1,0 +1,28 @@
+import { fakeAsync, flushMicrotasks, TestBed } from '@angular/core/testing';
+import { DropListService } from './drop-list.service';
+
+describe('DropListService', () => {
+  let service: DropListService;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({ providers: [DropListService] });
+    service = TestBed.inject(DropListService);
+  });
+
+  describe('subtask drag-start window', () => {
+    it('is closed by default', () => {
+      expect(service.isSubTaskDragStarting()).toBe(false);
+    });
+
+    it('opens synchronously and closes on the next microtask', fakeAsync(() => {
+      service.markSubTaskDragStarting();
+      // Must stay open through CDK's synchronous `_startReceiving` pass so the
+      // top-level lists get their geometry cached.
+      expect(service.isSubTaskDragStarting()).toBe(true);
+
+      flushMicrotasks();
+      // Closed again before the first pointer move so the pointer guard resumes.
+      expect(service.isSubTaskDragStarting()).toBe(false);
+    }));
+  });
+});
--- a/src/app/core-ui/drop-list/drop-list.service.ts
+++ b/src/app/core-ui/drop-list/drop-list.service.ts
@@ -22,6 +22,15 @@ export class DropListService {
 
   private _list: CdkDropList[] = [];
   private _flushScheduled = false;
+  private _activeDragPointer: { x: number; y: number } | null = null;
+
+  activeDragPointer(): { x: number; y: number } | null {
+    return this._activeDragPointer;
+  }
+
+  setActiveDragPointer(pointer: { x: number; y: number } | null): void {
+    this._activeDragPointer = pointer;
+  }
 
   registerDropList(dropList: CdkDropList, isSubTaskList = false): void {
     if (isSubTaskList) {
--- a/src/app/core-ui/drop-list/drop-list.service.ts
+++ b/src/app/core-ui/drop-list/drop-list.service.ts
@@ -28,6 +28,7 @@ export class DropListService {
   private _list: CdkDropList[] = [];
   private _flushScheduled = false;
   private _activeDragPointer: DragPointer | null = null;
+  private _isSubTaskDragStarting = false;
 
   activeDragPointer(): DragPointer | null {
     return this._activeDragPointer;
@@ -35,6 +36,36 @@ export class DropListService {
 
   setActiveDragPointer(pointer: DragPointer | null): void {
     this._activeDragPointer = pointer;
+  }
+
+  /**
+   * True only for the microtask in which a subtask drag begins.
+   *
+   * CDK caches a sibling drop-list's geometry (`DropListRef._domRect`) lazily,
+   * and for a non-source list only when its `enterPredicate` passes at drag
+   * start (`_startReceiving`). An uncached list can never be entered, so it can
+   * never receive the item. The parent DONE/UNDONE list normally rejects a
+   * subtask drag while the pointer is over the source subtask list (so in-list
+   * sorting keeps working) — but at the *instant* a subtask drag starts the
+   * pointer is always over that subtask list, so that guard would block the
+   * parent list from ever being cached, and converting a subtask back to a main
+   * task would silently fail until some unrelated parent drag warmed the cache.
+   *
+   * This flag opens a one-microtask window at drag start during which the
+   * top-level lists accept the drag (letting CDK cache their geometry); the
+   * pointer guard then resumes for the rest of the drag.
+   */
+  isSubTaskDragStarting(): boolean {
+    return this._isSubTaskDragStarting;
+  }
+
+  markSubTaskDragStarting(): void {
+    this._isSubTaskDragStarting = true;
+    // Clear after CDK's synchronous `_startReceiving` pass (which runs right
+    // after the `cdkDragStarted` output) but before the first pointer move.
+    queueMicrotask(() => {
+      this._isSubTaskDragStarting = false;
+    });
   }
 
   registerDropList(dropList: CdkDropList, isSubTaskList = false): void {
--- a/src/app/core-ui/drop-list/drop-list.service.ts
+++ b/src/app/core-ui/drop-list/drop-list.service.ts
@@ -3,6 +3,11 @@ import { CdkDropList } from '@angular/cdk/drag-drop';
 import { BehaviorSubject, merge, of, Subject, timer } from 'rxjs';
 import { map, startWith, switchMap } from 'rxjs/operators';
 
+export interface DragPointer {
+  x: number;
+  y: number;
+}
+
 @Injectable({
   providedIn: 'root',
 })
@@ -22,13 +27,13 @@ export class DropListService {
 
   private _list: CdkDropList[] = [];
   private _flushScheduled = false;
-  private _activeDragPointer: { x: number; y: number } | null = null;
+  private _activeDragPointer: DragPointer | null = null;
 
-  activeDragPointer(): { x: number; y: number } | null {
+  activeDragPointer(): DragPointer | null {
     return this._activeDragPointer;
   }
 
-  setActiveDragPointer(pointer: { x: number; y: number } | null): void {
+  setActiveDragPointer(pointer: DragPointer | null): void {
     this._activeDragPointer = pointer;
   }
 
--- a/src/app/core/persistence/operation-log/compact/action-type-codes.ts
+++ b/src/app/core/persistence/operation-log/compact/action-type-codes.ts
@@ -73,6 +73,7 @@ export const ACTION_TYPE_TO_CODE: Record<ActionType, string> = {
   [ActionType.TASK_SHARED_MOVE_TO_ARCHIVE]: 'HX',
   [ActionType.TASK_SHARED_RESTORE]: 'HR',
   [ActionType.TASK_SHARED_RESTORE_DELETED]: 'HRD',
+  [ActionType.TASK_SHARED_CONVERT_TO_SUB]: 'HCS',
   [ActionType.TASK_SHARED_CONVERT_TO_MAIN]: 'HC',
   [ActionType.TASK_SHARED_SCHEDULE_WITH_TIME]: 'HS',
   [ActionType.TASK_SHARED_RESCHEDULE_WITH_TIME]: 'HSR',
--- a/src/app/features/schedule/schedule-week/schedule-external-drag.service.ts
+++ b/src/app/features/schedule/schedule-week/schedule-external-drag.service.ts
@@ -5,6 +5,7 @@ import { DragRef } from '@angular/cdk/drag-drop';
 @Injectable({ providedIn: 'root' })
 export class ScheduleExternalDragService {
   private readonly _activeTask = signal<TaskWithSubTasks | null>(null);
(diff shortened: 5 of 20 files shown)


johannesjo:
Thanks for opening this up. I think it doesn't make much sense to add this, if the reverse isn't possible as well.

keilogic:
Updated this to support the reverse direction as well: subtasks can now be dragged back into the parent-level DONE/UNDONE lists, preserving the drop position and converting them back to main tasks.\n\nLocal validation passed:\n- focused task-list/shared reducer specs: 144 passed\n- Prettier check, ESLint on touched files, and git diff check passed\n- pre-push hook also ran package tests and the full Angular suite successfully

johannesjo:
Now drag and drop sub task to sort is not working anymore.

keilogic:
Pushed a fix for that regression in 37efbef. The parent DONE/UNDONE lists now avoid accepting a nested subtask drag while the pointer is still over a subtask list, so normal subtask sorting should continue to route to the subtask list while dragging out to the parent list still works once the pointer leaves the nested list area.\n\nValidation passed locally:\n- task-list component spec: 41 passed\n- related task/shared reducer specs: 300 passed\n- Prettier, ESLint on touched TS files, git diff check\n- pre-push hook full package tests and Angular suite

johannesjo:
Congratulations! 🎉 You succeded where many others (including me) failed in the past :) This is a great addition! I will do another more technical review. On minor thing: 
~~We should either hide the drag preview container for tasks without any sub tasks, or make dropping on it work.~~

We should not render the sub task task list for these tasks also to minimize performance impact.

johannesjo:
## Review feedback

Thanks for tackling #905 — the state-layer design here is solid: a single persistent `convertToSubTask` action handled in one meta-reducer pass (good for the op-log), the reducer re-validates eligibility against current state (nice defense against replayed/remote ops), TODAY_TAG ordering is stripped without writing `'TODAY'` into `tagIds`, and the test coverage for both convert directions + anchor positioning is genuinely thorough.

A few things I'd like addressed before merge:

### 🔴 Performance: empty drop target is mounted for every childless top-level task, always

```ts
isEmptySubTaskDropTargetMounted = computed(() => {
  const t = this.task();
  return !this.isInSubTaskList() && !t.parentId && !t.subTasks?.length; // no drag dependency
});
```

This has no dependency on whether a drag is in progress, so for a list of **N** top-level leaf tasks (the common case) it permanently mounts **N extra `<task-list>` components**, each registering its own `CdkDropList` in `ngAfterViewInit`. Since the task list isn't virtualized, that's roughly a 2× component + drop-list count at idle, plus extra work per pointer-move during *any* drag (CDK scans connected siblings, and runs `getBoundingClientRect()` across all the new `SUB` targets at drag start). On large projects / Android this is a real regression.

Suggested fix: gate **mounting** (not just visibility) on an active eligible drag — `_scheduleExternalDragService.activeTask()` is already a signal, and `isEmptySubTaskDropTargetVisible` already encodes the full condition, so the two computed signals collapse into one and the idle cost drops to zero.

⚠️ One thing to verify with a manual drag test: CDK builds its sibling graph in `beforeStarted` (drag start), while `activeTask` is set in `onDragStarted` just after — so a target mounted mid-drag might not join the *in-flight* drag. If same-drag drop-in breaks, the mount needs to flip on a signal that fires before CDK reads siblings, or use a single shared lightweight drop zone instead of one CdkDropList per task.

### 🟠 `convertToSubTask` isn't registered in the `ActionType` enum / compact codec

It's declared `isPersistent`, but it's missing from `op-log/core/action-types.enum.ts` and `…/compact/action-type-codes.ts`. Sync/replay still works (the codec falls back to the full string), but `operation-log.effects.ts` fires a `devError` on every convert — production logs an error on a core action, and dev/e2e pops a `devERR` alert. CI won't catch this (the enum-count spec stays green since nothing was added). Quick fix: add `TASK_SHARED_CONVERT_TO_SUB` to the enum + a code (e.g. `'HCS'`) to `ACTION_TYPE_TO_CODE`, and bump the count assertion.

### 🟡 DOM hit-test inside `enterPredicate`

`_isPointerOverSubTaskList()` does `document.elementFromPoint()` + `.closest('.task-list-inner')` (fed by a per-frame `onDragMoved` write to `DropListService` + a `data-list-id` attr). CDK's own `_canReceive` already does an `elementFromPoint` containment check, so this can hit-test the same point twice per frame and may force reflow. It's bounded (only the subtask→top-level gesture), but the genuine gap it covers — CDK excluding the source list from its siblings — is likely resolvable from `drag.dropContainer`/`drop.data` without a parallel DOM hit-test. At minimum, a comment explaining why CDK's built-in resolution is insufficient would help. (Note: the per-frame `onDragMoved` itself is cheap here since the app is zoneless — no CD churn — but the whole channel becomes unnecessary if the predicate can derive this from drag/drop data.)

### Minor / suggestions

- `positionConvertedTask`'s `undefined | null+isDone | string` tri-state is hard to read — consider having the context-menu caller pass an explicit position so the reducer only handles two cases, or at least comment each branch's intent.
- `handleConvertToSubTask` and `handleConvertToMainTask` use divergent trust models (state-scan vs payload `parentTagIds`); the single-element `new Set([task.id])` could just be `tag.taskIds.includes(...)`.
- `drop.data.allTasks` is now orphaned — no predicate path reads it but it's still populated.
- A cross-project convert (project-A task under a project-B parent in a tag/TODAY view) reassigns `projectId` and looks correct, but has no dedicated test — worth adding given the data-movement.
- `_activeDragPointer` is cleared in `onDragEnded` but not `ngOnDestroy` (stale on mid-drag destroy; harmless in practice but tidier to clear).

Overall: 👍 on the approach, just needs the performance gating and the enum registration before it's mergeable.

keilogic:
Pushed `59bc480` addressing the review items:

- empty subtask drop targets now mount only for an eligible task drag; a short-lived pointer-down candidate pre-mounts them before CDK builds the sibling graph, then active drag state keeps them mounted during the drag
- registered `convertToSubTask` in `ActionType` and compact op-log codes (`HCS`) and bumped the enum count spec
- removed the now-unused `allTasks` drop-list payload, clear drag pointer state on destroy, and documented why the subtask-list DOM hit-test is still present

Validation:
- focused task-list + op-log action/codec specs: 57 passed
- task-shared reducer specs: 121 passed
- `npm run lint:ts -- --quiet`
- Prettier check on touched files
- `git diff --check`
- pre-push hook: package tests, release-note tests, full Angular suite and timezone suite all passed; only the existing warning-only unused eslint-disable lint messages were reported

johannesjo:
## Review — multi-perspective (correctness, sync, performance, architecture, simplicity)

Nice work — the data-model half is clean: `convertToSubTask` is correctly modeled as a **single meta-reducer pass** (one user intent = one op), the CRUD/section/action/enum/code-map wiring is consistent with `convertToMainTask`, and the reducer specs are thorough. The feedback below is concentrated in the **drag-interaction layer**. I verified each item against the code (and the CDK source for the perf one).

### 🔴 Worth fixing before merge

**1. `(cdkDragMoved)` reintroduces a change-detection tick on every pointer move — for *all* task drags**
`task-list.component.html:50` → `onDragMoved` (`task-list.component.ts:202`).
CDK runs drag-move handling *outside* Angular's zone and only re-enters per move **when something subscribes to `cdkDragMoved`** — see `@angular/cdk/.../drag-drop.mjs`: `if (this._moveEvents.observers.length) { this._ngZone.run(() => this._moveEvents.next(...)) }`. Binding `(cdkDragMoved)` makes `observers.length ≥ 1`, so every raw `pointermove` (60–120+/sec) now forces an `NgZone.run` → app-wide CD tick, for the whole duration of *every* task drag, not just conversions.
Note the zone re-entry is unconditional once *any* subscriber exists, so subscribing "outside the zone" wouldn't help. To keep the pointer without the per-move tick, capture it from a native `pointermove` listener registered via `runOutsideAngular`, attach the subscription only for the subtask-conversion case, or drop the pointer cache entirely (see #5).

**2. A subtask dragged to DONE is marked done but never gets `doneOn`**
`handleConvertToMainTask` (`task-shared-crud.reducer.ts:189-193`) sets `isDone: true` directly but never stamps `doneOn`. The normal completion path (`updateDoneOnForTask`) always sets `doneOn: Date.now()` and the completion-day `dueDay`. As-is, dragging a subtask onto DONE produces a completed task with `doneOn: undefined`, so the worklog/metrics file it under the task's **created** date instead of today. Suggest stamping `doneOn` (and the completion-day `dueDay`) in the `isDone === true` branch, or routing through `updateDoneOnForTask`.

**3. Section meta-reducer skips the eligibility guard the CRUD reducer enforces**
`section-shared.reducer.ts:364` handles `convertToSubTask` by **unconditionally** calling `handleTaskRemoval(state, [taskId])`, while `handleConvertToSubTask` bails out (`return state`) when `!canConvertTaskToSubTask(task)`. The UI gates this on the happy path, but for an ineligible task that *is* in a section (reached via a replayed/synced/programmatic op), the CRUD pass leaves it top-level while the section pass strips its section membership → inconsistent local state. It's not a sync *desync* (both passes replay deterministically), but mirroring the guard restores the defense-in-depth the CRUD reducer deliberately adds.

**4. `isEmptySubTaskDropTargetVisible` is a verbatim alias of `…Mounted`**
`task.component.ts:294` returns `this.isEmptySubTaskDropTargetMounted()`. Since the template applies `is-empty-sub-task-drop-target` and `is-drop-visible` from the two always-equal signals, the element mounts with `opacity: 0.85` already set, so the SCSS `opacity 0 → 0.85` fade never animates. Either collapse to one signal, or gate visibility on a distinct condition (e.g. pointer-over) to get the intended two-phase reveal.

### 🟡 Consider

**5. The pointer-tracking subsystem is the PR's main complexity sink** — `_isPointerOverSubTaskList()` (`document.elementFromPoint` + `.closest('.task-list-inner')` + `data-list-id`) + the `DropListService` pointer cache + `(cdkDragMoved)` largely re-implements CDK's own `_canReceive` hit-test. CDK genuinely doesn't pass pointer coords to `enterPredicate`, so *some* access is needed — but a `(cdkDropListExited)` / source-container-identity gate could likely replace the per-move cache and would also resolve #1. Worth prototyping (incl. touch) before committing to the manual approach. Minor: `data-list-id` duplicates `drop.data.listId`, which is already in hand.

**6. `subTaskDropCandidate` lives on `ScheduleExternalDragService`** — a tasks-only concept on a schedule-named root service; a small `TaskDragStateService` (or folding pointer+candidate into `DropListService`) would be more cohesive. Relatedly, the pre-drag candidate + global `pointerup`/`pointercancel` listeners may be redundant with `activeTask()` (the target appearing a few px into the drag is likely imperceptible) — worth checking on-device. The listeners are correctly cleaned up, so this is cleanliness, not a leak.

### 🟢 Minor
- `convertToSubTask`'s `{ taskId, targetParentId, afterTaskId }` payload matches no recognized shape in `validate-operation-payload.ts`, so every such op logs the `UPDATE_UNUSUAL_STRUCTURE` warning (non-blocking — `success: true`). Adding a recognized TASK shape would silence the noise.
- Global window listeners could use `host: { '(window:pointerup)': … }` / `takeUntilDestroyed()` instead of manual add/remove bookkeeping.
- Consider a `convertToSubTask` case in `state-validity-after-actions.spec.ts` (replay-determinism for this multi-entity op) and a `doneOn` assertion for the subtask→DONE path.

### Dismissed (checked, not issues)
`modified: Date.now()` in the reducer (consistent with existing handlers; not a convergence field) · window-listener memory leak (all cleanup paths present) · multi-entity single-op replay (meta-reducer pattern is correct & deterministic) · cycle creation (leaf-only guard).

keilogic:
Pushed `53c4102` addressing the follow-up review items:

- removed the Angular `(cdkDragMoved)` binding; subtask pointer tracking now uses a native `pointermove` listener registered outside Angular and only during subtask drags
- subtask -> DONE conversion now stamps `doneOn` and the completion-day `dueDay`
- section cleanup now mirrors the convert-to-subtask eligibility guard so rejected/replayed ineligible ops keep section membership intact
- collapsed the duplicate empty-drop-target mounted/visible signal and removed the unused visibility class
- added a recognized TASK payload shape for `convertToSubTask` validation noise

Validation passed:
- focused specs: 179 passed
- op-log validation specs: 80 passed
- `npm run lint:ts -- --quiet`
- Prettier on touched files
- `git diff --check`
- pre-push hook full package/release-note/Angular/timezone suites passed; only the existing warning-only unused eslint-disable messages in unrelated files were reported

keilogic:
Thanks for the review. I’ve pushed updates addressing the follow-up items:

- removed the cdkDragMoved binding and kept drag pointer tracking outside Angular
- added doneOn and completion-day dueDay when converting a done subtask
- mirrored the convertToSubTask eligibility guard in the section meta-reducer
- removed the old empty-drop-target mounted/visible alias
- registered convertToSubTask in the action enum with a compact payload shape

Local validation passes:
- focused Angular specs: 235 passed
- npm run lint:ts -- --quiet
- Prettier on touched files
- git diff --check

The GitHub workflow is currently waiting for action_required, so I can’t fix a CI failure from logs yet, but I’ll follow up if a run starts and fails.

keilogic:
I pushed one small follow-up in `1150b55` after checking the current head: the task-list predicate spec was still carrying stale `allTasks` mock data from the removed drop-list payload. It now uses the current `filteredTasks` drop data shape and removes the duplicate parent-context cases.

Validation passed:
- focused task-list spec: 46 success
- `npm run lint:ts -- --quiet`
- Prettier check for the touched file
- `git diff --check`
- pre-push hook full lint, package tests, release-note tests, Angular suite, and timezone suite passed; only the existing warning-only unused eslint-disable messages in unrelated files were reported

keilogic:
Thanks, I appreciate the detailed review and the follow-up cleanup commits.

Current head `1150b55` is now through hosted checks: Analyze, lockfile registry check, CodeQL, dependency review, Lighthouse, Lint, packaged Linux app, Tests, and preview build all pass. I also checked the review thread state and there are no unresolved inline threads.

I’ll keep an eye on the PR in case another review pass surfaces anything concrete.

johannesjo:
Thank you for this, @keilogic\! 🙏 Your implementation is the foundation of the feature. I've carried it forward with follow-up refinements and review fixes in #7962 — https://github.com/super-productivity/super-productivity/pull/7962 — where your commits are preserved and you're credited as co-author. I'll close this once #7962 is merged.

johannesjo:
Closing in favor of #7962, which carries your work forward with you credited as co-author — #7962 is the PR that will merge to master. Thanks again @keilogic\! 🙏

Issue #3888, opened 2025-01-12
Scheduled tasks marked done next day counts as worktime of next day


Pekkari:
### Your Environment

- Version used: 11.0.3
- Operating System and version: Fedora 41
- Desktop Environment: Plasma desktop + podman/docker
- Browser Name and version: Falkon 24.12.0

### Expected Behavior

I scheduled a task to happen at the end of my day, and I didn't mark it as done
until next day. Next day, when marking it done, I add the time taken to, for ex.,
2h. I'd expect the work time to count as work for the former day.

### Current Behavior

The working time counts as of the day it is marked as done.

### Steps to Reproduce (for bugs)

1. Create an scheduled task for today, anytime.
2. Don't mark it done today, nor use the finish day option.
3. Next day, mark it done, and see the work time of your just started day.

### Can you reproduce this reliably?

Yes.

Thanks!

Billybs:
Hello! Would it be okay if I work on this? I'm new to open source but I'd like to contribute.

johannesjo:
This would be great @Billybs ! Thank you for the offer!

johannesjo:
Sorry I misunderstood this issue at first it seems.

Imho the time spent in the dialog should be **_always_** for today unless you specify it specifically for another day. Everything else would be confusing in my understanding as the dialog would do sometimes this and sometimes that.

Pekkari:
Is it that confusing that something that started yesterday, let say 18h, and you mark it as 2h, is computed as time of yesterday? it only requires to modify the ending time, and from there notice that its end wasn't today.

The problem arose because I often started a task at the end of a day, like participating in a local event, and I wouldn't jump into keyboard until next day, so I certainly wanted to bookkeep how much I spent, but only way to do so it was to fire the laptop and close it on the same day, which it isn't something I was always into. I fail to know if this has changed in further releases though, since it didn't work when I tested this, I jumped back to my former time tracker, so I can solve the problem.

johannesjo:
Thanks for the clarification, and sorry again for the initial misunderstanding.

I now better understand your use case: you're often doing tasks (like events) that occur late in the day, and you want to log the time spent retroactively without needing to open the app on the same day or adjust the date manually the next day. That’s a valid workflow, and I appreciate you taking the time to explain it.

[...]

PR #7909, opened 2026-06-01
fix(tracking): split idle time across days when it crosses midnight

johannesjo:
## What

When idle time is assigned to a task (or a new task is created from it) via the idle dialog, the whole interval was logged to **today's** `timeSpentOnDay` key — even when the away period crossed midnight. This splits the interval `[now - idleTime, now]` across the logical days it actually covers.

Found while investigating #3888 (a recent comment revived the thread). The issue's original subject — the manual "mark done / estimate" dialog logging to today — remains intended; this PR fixes a *related* idle edge case, so `Refs #3888` rather than closing it.

## Changes

- **`split-time-spent-by-day.ts`** (new pure util): splits a duration ending at a timestamp into a `{ dateStr: ms }` map by logical-day boundary, honoring the start-of-next-day offset (matches `DateService.todayStr`).
- **`TaskService.addTimeSpentForDays`**: applies a per-day map to one task. It **threads an updated task snapshot** through the per-day dispatches — the `addTimeSpent` reducer merges onto the action payload's `timeSpentOnDay` (not store state), so reusing one stale snapshot would make each iteration overwrite the previous day's update (data loss). A regression test pins this.
- **`addTimeSpentAndSync`**: optional `date` param (defaults to today; all existing callers unchanged).
- **`idle.effects.ts`**: the existing-task and new-task assignment branches now split across days.

## Why threaded per-day dispatches (not one combined op)

Time tracking already syncs as one additive `syncTimeSpent` op per `(task, date)` (the `BatchedTimeSyncAccumulator` flushes per day). Emitting one op per covered day is consistent with that model and keeps additive-merge conflict semantics; a single absolute-set op would lose concurrent additions from other devices. Realistic idle intervals are 1–2 days.

## Tests

- `split-time-spent-by-day.spec.ts` — 10 cases (same-day, cross-midnight, 3-day spread, start-of-next-day offset, exact-midnight, edge/invalid inputs, sum invariant).
- `task.service.spec.ts` — regression test for the threaded per-day dispatch (asserts the 2nd day's payload carries the 1st day's addition).
- `idle.effects.spec.ts` (new) — end-to-end wiring through the real effect: existing-task split, same-day, new-task split, and start-of-next-day offset. Passes under both `Europe/Berlin` and `America/Los_Angeles`.
- `task.reducer.spec.ts` — reducer-level end-to-end tests replaying the two-dispatch cross-midnight sequence through the real reducer, proving both days survive for a flat task and a subtask (parent aggregation, no double-count).

## Out of scope (follow-up)

The detection-time `removeTimeSpent` still defaults to "today" — left unchanged because the OS idle clock can't distinguish timer-tracked-while-away from slept time, so day-splitting the *removal* could eat legitimate prior-day work. Tracked separately.

### DIFF
--- a/src/app/features/idle/store/idle.effects.spec.ts
+++ b/src/app/features/idle/store/idle.effects.spec.ts
@@ -1,0 +1,197 @@
+import { TestBed } from '@angular/core/testing';
+import { MockStore, provideMockStore } from '@ngrx/store/testing';
+import { Action } from '@ngrx/store';
+import { EMPTY, Subject } from 'rxjs';
+import { MatDialog } from '@angular/material/dialog';
+
+import { IdleEffects } from './idle.effects';
+import { idleDialogResult } from './idle.actions';
+import { LOCAL_ACTIONS } from '../../../util/local-actions.token';
+import { TaskService } from '../../tasks/task.service';
+import { WorkContextService } from '../../work-context/work-context.service';
+import { SimpleCounterService } from '../../simple-counter/simple-counter.service';
+import { UiHelperService } from '../../ui-helper/ui-helper.service';
+import { ChromeExtensionInterfaceService } from '../../../core/chrome-extension-interface/chrome-extension-interface.service';
+import { DateService } from '../../../core/date/date.service';
+import { DataInitStateService } from '../../../core/data-init/data-init-state.service';
+import { DEFAULT_TASK, Task } from '../../tasks/task.model';
+import { IdleTrackItem } from '../dialog-idle/dialog-idle.model';
+
+const H = 60 * 60 * 1000;
+
+describe('IdleEffects handleIdleDialogResult$ (cross-midnight, #3888)', () => {
+  let effects: IdleEffects;
+  let actions$: Subject<Action>;
+  let taskService: jasmine.SpyObj<TaskService>;
+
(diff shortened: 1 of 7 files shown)


johannesjo:
Closing this — on reflection I don't think it's worth the complexity it adds.

It fixes a real but narrow edge (assigning idle time that crossed midnight now splits across both days instead of all landing on "today"), but:

- It doesn't address the actual #3888 report — manual time entry logging to "today" is intended, and the recent revival of that thread is most likely work-context scoping. This was an edge found while investigating, not a reported pain point.
- The trigger is rare: you have to leave the app idle across midnight *and* assign that time as work (not a break) *and* care about the pre-midnight portion landing on the previous day.
- The minimum-correct implementation is inherently subtle. Splitting requires threading an updated task snapshot through the per-day dispatches, because `addTimeSpent` merges onto the action payload's `timeSpentOnDay` rather than store state — reusing a single snapshot silently drops days. That's lasting fragility (a "simplification" later would reintroduce data loss) disproportionate to the payoff.

The analysis is preserved here and on #3888. If this is ever worth revisiting, a much simpler variant exists: attribute the whole idle interval to the day it *started* (one line, no new abstraction, no threading) — less precise than splitting, but it covers the common "walked away last night" case without the fragility.

PR #7926, opened 2026-06-01
feat(projects): add project icons to show/hide projects visibility menu

### LINKED ISSUE #7894: 💡 Show/hide projects: Add project icons next to project names

MannXo:
## Problem

The "Show/Hide Projects" visibility menu (the eye icon beside the Projects section in the sidebar) displayed only project names — no project icons. Users who rely on the visual icon to identify projects had no icon reference in this menu, making it inconsistent with the main sidebar where projects are listed as `[icon] name`.

Fixes #7894

## Solution

- Replace the `visibility`/`visibility_off` icon with each project's own configured icon as the leading icon in the menu item — consistent with how every other nav menu in the sidebar renders items (`[single icon] label`, see `nav-mat-menu.component.html`).
- Hidden projects are visually distinguished by dimming the entire menu row to `opacity: 0.4`, which is a standard "off/inactive" convention and avoids a double-icon layout that has no precedent in the codebase.

**Design note:** An earlier iteration used `[eye icon] [project icon] title` to make the toggle state explicit, but that pattern doesn't exist anywhere else in the nav menus. The opacity approach conveys the same information with a single-icon layout that matches the rest of the UI. Happy to discuss if a different treatment is preferred.

## Type of Change

- [x] New feature

## Checklist

- [x] I have added tests for my changes (if applicable) — no unit-testable logic changed; this is a template binding and a CSS rule

### DIFF
--- a/src/app/core-ui/magic-side-nav/nav-list/nav-list-tree.component.html
+++ b/src/app/core-ui/magic-side-nav/nav-list/nav-list-tree.component.html
@@ -161,6 +161,8 @@
   @for (project of allUnarchivedProjects(); track project.id) {
     <button
       mat-menu-item
+      role="menuitemcheckbox"
+      [attr.aria-checked]="!project.isHiddenFromMenu"
       (click)="toggleProjectVisibility(project.id)"
     >
       @if (!project.isHiddenFromMenu) {
@@ -168,7 +170,11 @@
       } @else {
         <mat-icon>check_box_outline_blank</mat-icon>
       }
-      <mat-icon>{{ project.icon || DEFAULT_PROJECT_ICON }}</mat-icon>
+      <mat-icon
+        class="nav-icon"
+        [class.nav-icon-emoji]="isSingleEmoji(project.icon || DEFAULT_PROJECT_ICON)"
+        >{{ project.icon || DEFAULT_PROJECT_ICON }}</mat-icon
+      >
       {{ project.title }}
     </button>
   }
--- a/src/app/core-ui/magic-side-nav/nav-list/nav-list-tree.component.html
+++ b/src/app/core-ui/magic-side-nav/nav-list/nav-list-tree.component.html
@@ -157,16 +157,19 @@
 }
 
 <!-- Project Visibility Menu -->
-<!-- Hidden-project opacity style in src/styles/components/_overwrite-material.scss -->
 <mat-menu #visibilityMenu="matMenu">
   @for (project of allUnarchivedProjects(); track project.id) {
     <button
       mat-menu-item
-      [class.is-hidden-from-menu]="project.isHiddenFromMenu"
       (click)="toggleProjectVisibility(project.id)"
     >
+      @if (!project.isHiddenFromMenu) {
+        <mat-icon>check_box</mat-icon>
+      } @else {
+        <mat-icon>check_box_outline_blank</mat-icon>
+      }
       <mat-icon>{{ project.icon || DEFAULT_PROJECT_ICON }}</mat-icon>
-      <span>{{ project.title }}</span>
+      {{ project.title }}
     </button>
   }
   @if (archivedProjectsCount() > 0) {
--- a/src/app/core-ui/magic-side-nav/nav-list/nav-list-tree.component.html
+++ b/src/app/core-ui/magic-side-nav/nav-list/nav-list-tree.component.html
@@ -157,17 +157,15 @@
 }
 
 <!-- Project Visibility Menu -->
+<!-- Hidden-project opacity style in src/styles/components/_overwrite-material.scss -->
 <mat-menu #visibilityMenu="matMenu">
   @for (project of allUnarchivedProjects(); track project.id) {
     <button
       mat-menu-item
+      [class.is-hidden-from-menu]="project.isHiddenFromMenu"
       (click)="toggleProjectVisibility(project.id)"
     >
-      @if (!project.isHiddenFromMenu) {
-        <mat-icon>visibility</mat-icon>
-      } @else {
-        <mat-icon>visibility_off</mat-icon>
-      }
+      <mat-icon>{{ project.icon || DEFAULT_PROJECT_ICON }}</mat-icon>
       <span>{{ project.title }}</span>
     </button>
   }
--- a/src/app/core-ui/magic-side-nav/nav-list/nav-list-tree.component.scss
+++ b/src/app/core-ui/magic-side-nav/nav-list/nav-list-tree.component.scss
@@ -133,6 +133,33 @@
   }
 }
 
+.nav-icon {
+  flex-shrink: 0;
+  width: 24px;
+  height: 24px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  font-size: 18px;
+  line-height: 1;
+  margin-right: var(--s-half);
+
+  &.nav-icon-emoji {
+    font-family: 'Apple Color Emoji', 'Segoe UI Emoji', 'Noto Color Emoji', sans-serif;
+    font-size: 16px;
+    font-feature-settings: normal;
+    font-variation-settings: normal;
+    text-transform: none;
+    letter-spacing: normal;
+    overflow-wrap: normal;
+    white-space: nowrap;
+    direction: ltr;
+    -webkit-font-smoothing: antialiased;
+    -moz-osx-font-smoothing: grayscale;
+    text-rendering: optimizeLegibility;
+  }
+}
+
 .nav-child-item {
   position: relative;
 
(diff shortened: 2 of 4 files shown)


johannesjo:
Thanks for tackling this, and welcome — nicely scoped PR with a clear write-up of the tradeoff. 🎉

On the design question you flagged: I think there's a better reference class than the nav menus. This menu is really a *"toggle which projects are in the list"* control, and the app already has an established pattern for exactly that — a **checkbox icon + the item's own icon + title**. See [`tag-toggle-menu-list.component.html`](https://github.com/super-productivity/super-productivity/blob/master/src/app/features/tag/tag-toggle-menu-list/tag-toggle-menu-list.component.html) (toggling tags on a task) and `add-task-bar-actions.component.html` (~L193), both of which do:

```
[check_box / check_box_outline_blank]   [item icon]   Title
```

So the `[state icon] [item icon] [title]` layout you worried had no precedent is actually the convention here for toggle-list menus — it just uses a checkbox rather than an eye. Applied to this menu:

```
☑  📁  Project A      (check_box               + project.icon)
☐  💼  Project B      (check_box_outline_blank + project.icon)
```

Why I'd lean this way over the opacity dimming:

- It keeps the project icon (the actual ask in #7894) **and** makes the toggle state explicit even for *visible* rows — with opacity alone, a visible project row looks identical to a normal nav item, so there's no cue that clicking toggles anything.
- `opacity: 0.4` is the app's `--disabled-opacity` value. A hidden project isn't *disabled* (you click it to re-show it), so dimming it borrows the visual language of "can't interact with this," which is a bit misleading.
- The checkbox form also announces state to screen readers for free, which neither the old `visibility_off` icon nor the opacity approach does (Material's `mat-icon` is `aria-hidden` by default).

The change stays small — swap the `visibility`/`visibility_off` icons for `check_box`/`check_box_outline_blank` and add the project icon after them, mirroring `tag-toggle-menu-list`. That also lets you drop the new global SCSS rule entirely.

Not a blocker on the icon rendering itself, which is correct and matches the existing `project.icon || DEFAULT_PROJECT_ICON` pattern — I'd just prefer it land as a checklist rather than dimming. Happy to hear your thoughts if you see it differently.

MannXo:
> Thanks for tackling this, and welcome — nicely scoped PR with a clear write-up of the tradeoff. 🎉
> 
> On the design question you flagged: I think there's a better reference class than the nav menus. This menu is really a _"toggle which projects are in the list"_ control, and the app already has an established pattern for exactly that — a **checkbox icon + the item's own icon + title**. See [`tag-toggle-menu-list.component.html`](https://github.com/super-productivity/super-productivity/blob/master/src/app/features/tag/tag-toggle-menu-list/tag-toggle-menu-list.component.html) (toggling tags on a task) and `add-task-bar-actions.component.html` (~L193), both of which do:
> 
> ```
> [check_box / check_box_outline_blank]   [item icon]   Title
> ```
> 
> So the `[state icon] [item icon] [title]` layout you worried had no precedent is actually the convention here for toggle-list menus — it just uses a checkbox rather than an eye. Applied to this menu:
> 
> ```
> ☑  📁  Project A      (check_box               + project.icon)
> ☐  💼  Project B      (check_box_outline_blank + project.icon)
> ```
> 
> Why I'd lean this way over the opacity dimming:
> 
> * It keeps the project icon (the actual ask in [💡 Show/hide projects: Add project icons next to project names #7894](https://github.com/super-productivity/super-productivity/issues/7894)) **and** makes the toggle state explicit even for _visible_ rows — with opacity alone, a visible project row looks identical to a normal nav item, so there's no cue that clicking toggles anything.
> * `opacity: 0.4` is the app's `--disabled-opacity` value. A hidden project isn't _disabled_ (you click it to re-show it), so dimming it borrows the visual language of "can't interact with this," which is a bit misleading.
> * The checkbox form also announces state to screen readers for free, which neither the old `visibility_off` icon nor the opacity approach does (Material's `mat-icon` is `aria-hidden` by default).
> 
> The change stays small — swap the `visibility`/`visibility_off` icons for `check_box`/`check_box_outline_blank` and add the project icon after them, mirroring `tag-toggle-menu-list`. That also lets you drop the new global SCSS rule entirely.
> 
> Not a blocker on the icon rendering itself, which is correct and matches the existing `project.icon || DEFAULT_PROJECT_ICON` pattern — I'd just prefer it land as a checklist rather than dimming. Happy to hear your thoughts if you see it differently.

Great catch. Reflected in my most recent commit.

johannesjo:
Thanks for turning this around so quickly — the checkbox + project-icon layout is exactly what I had in mind, and dropping the extra SCSS rule is a nice bonus. 🎉

One thing worth handling before this goes in, and then I think we're there:

**Emoji project icons need the same special-casing as everywhere else.** Project icons can be emoji (the icon field is the `icon-input` component, which extracts/stores emoji), and emoji don't render cleanly inside a plain `<mat-icon>` — that's why both reference points special-case them:

- the main nav uses `[class.nav-icon-emoji]="isWorkContextEmojiIcon()"` (emoji font-family + 16px sizing), and
- the `tag-toggle-menu-list` you're mirroring renders emoji through a separate `<span matMenuItemIcon class="tag-ico-emoji">` branch rather than `<mat-icon>`.

As written, a project with an emoji icon will look oversized/misaligned here while looking fine everywhere else. Mirroring the tag-toggle branch handles it — roughly:

```html
@if (isSingleEmoji(project.icon || DEFAULT_PROJECT_ICON)) {
  <span class="project-ico-emoji">{{ project.icon || DEFAULT_PROJECT_ICON }}</span>
} @else {
  <mat-icon>{{ project.icon || DEFAULT_PROJECT_ICON }}</mat-icon>
}
```

(`isSingleEmoji` lives in `src/app/util/extract-first-emoji.ts`; expose it on the component like the nav does.)

Two small nits while you're in there:

- The project `<mat-icon>` has no class, so the gap to the title will be wider than the tag menu it's modeled on — `tag-toggle` puts `margin-right: var(--s-half)` on the item icon (`.tag-ico`). Worth matching for visual consistency.
- Could you update the PR description? It still describes the opacity approach and "the new global SCSS rule," which no longer match the diff.

Nothing major — the icon rendering itself is correct and the fallback handling is spot on. 🙏

MannXo:
Could you trigger a re-run of the Tests job? The failure is in `focus-mode/auto-start-focus-on-play.spec.ts` which tests the header focus-button countdown, completely unrelated to this PR's changes (sidebar visibility menu template + SCSS). The test itself notes it's waiting on a multi-step async effect chain and is known to be sensitive to CI load (20s timeout). Our changes touch only `nav-list-tree.component` and `extract-first-emoji` import.

johannesjo:
Re-triggering the Tests job now — you're right that `auto-start-focus-on-play.spec.ts` is an unrelated timing flake, not your change.

No need to re-address my earlier notes beyond what's already there, but one update that should make the emoji fix *less* work than I first suggested: instead of the tag-toggle `@if/@else` branch, you can mirror the sibling `nav-item` component right next door, which handles emoji with a single element —

```html
<mat-icon
  class="nav-icon"
  [class.nav-icon-emoji]="isSingleEmoji(project.icon || DEFAULT_PROJECT_ICON)"
  >{{ project.icon || DEFAULT_PROJECT_ICON }}</mat-icon
>
```

(expose `isSingleEmoji` on the component; copy the `.nav-icon` / `.nav-icon-emoji` rules into `nav-list-tree.component.scss` since they're scoped to `nav-item` — that also handles the icon spacing). And a small correction to my own earlier note: the `check_box` icon is `aria-hidden`, so it doesn't announce state to screen readers on its own — adding `role="menuitemcheckbox"` + `[attr.aria-checked]="!project.isHiddenFromMenu"` to the button does (same as `schedule.component.html` ~L42). Optional, but cheap.

MannXo:
> Re-triggering the Tests job now — you're right that `auto-start-focus-on-play.spec.ts` is an unrelated timing flake, not your change.
> 
> No need to re-address my earlier notes beyond what's already there, but one update that should make the emoji fix _less_ work than I first suggested: instead of the tag-toggle `@if/@else` branch, you can mirror the sibling `nav-item` component right next door, which handles emoji with a single element —
> 
> ```
> <mat-icon
>   class="nav-icon"
>   [class.nav-icon-emoji]="isSingleEmoji(project.icon || DEFAULT_PROJECT_ICON)"
>   >{{ project.icon || DEFAULT_PROJECT_ICON }}</mat-icon
> >
> ```
> 
> (expose `isSingleEmoji` on the component; copy the `.nav-icon` / `.nav-icon-emoji` rules into `nav-list-tree.component.scss` since they're scoped to `nav-item` — that also handles the icon spacing). And a small correction to my own earlier note: the `check_box` icon is `aria-hidden`, so it doesn't announce state to screen readers on its own — adding `role="menuitemcheckbox"` + `[attr.aria-checked]="!project.isHiddenFromMenu"` to the button does (same as `schedule.component.html` ~L42). Optional, but cheap.

Reflected in my most recent commit.

johannesjo:
Thank you very much! <3

Issue #7239, opened 2026-04-16
💡 Local REST API should support creating recurring tasks


RealDishwash:
### Problem Statement

The desktop Local REST API at `http://127.0.0.1:3876` currently supports basic task CRUD, but it does not support creating recurring tasks.

The docs now describe the Local REST API in `docs/wiki/3.01-API.md`, and the task create/update routes exist, but the current handler only allows a small subset of task fields:

- `title`
- `notes`
- `isDone`
- `timeEstimate`
- `timeSpent`
- `projectId`
- `tagIds`
- `dueDay`
- `dueWithTime`
- `plannedAt`

In `src/app/core/electron/local-rest-api-handler.service.ts`, recurrence-related fields are not accepted, and there are no endpoints for managing repeat configs.

As a result, local automation tools can create normal tasks, but they cannot create recurring tasks through the documented Local REST API, even though recurrence is supported by the app itself.

### Verified Current Behavior

I tested `POST /tasks` against a running desktop instance and sent recurrence-like fields such as:

```json
{
  "title": "REST API recurrence field test",
  "projectId": "INBOX_PROJECT",
  "repeatCfgId": "TEST_REPEAT_CFG",
  "repeatCycle": "DAILY",
  "repeatEvery": 1,
  "quickSetting": "DAILY"
}
```

The task was created successfully, but all recurrence-related fields were ignored/stripped from the returned task and from a subsequent `GET /tasks/:id`.

### Why This Matters

The Local REST API is a strong fit for local automation, AI agents, scripts, and integrations. Recurring tasks are a core task-management feature, so not being able to create them through the local API makes that automation incomplete.

This is especially noticeable now that the Local REST API is documented as a first-class interface in `3.01-API.md`.

### Possible Solution

Any of these would unblock the use case:

1. Extend `POST /tasks` and `PATCH /tasks/:id` to accept recurrence-related fields.
2. Add dedicated repeat-config endpoints such as:
   - `POST /repeat-configs`
   - `PATCH /repeat-configs/:id`
   - `DELETE /repeat-configs/:id`
3. Document the supported recurrence payload shape and how it maps to tasks and repeat configs.

If the internal model requires repeat configs to be managed separately, explicit REST endpoints for repeat configs would probably be the cleanest design.

### Additional Context

Relevant docs/source:

- `docs/wiki/3.01-API.md`
- `src/app/core/electron/local-rest-api-handler.service.ts`
[...]

Issue #7373, opened 2026-04-26
🚨 Changing time for "Future Tasks" unexpectedly shifts the next occurrence to the following day


ertugruluyar-github:
### Version Used

v18.3.0 (Android 16)

### Expected Behavior

The task for today should remain on today's date but reflect the new time (2:00 PM), and all subsequent occurrences should also be updated to 2:00 PM. 

### Current Behavior

When updating the time of a recurring task and selecting the "Only future tasks" option, the current instance of the task is incorrectly rescheduled to the next day. Repeating this action continues to push the task forward by one day each time, rather than simply updating the time for the scheduled dates.

### Steps to Reproduce

1. Create a recurring task (e.g., Daily at 10:00 AM).
​2. Open the task edit menu for today's instance.
​3. Click to recurring task option. 
4. Change the time (e.g., from 06:00 PM to 07:15 PM).
​5. Save the changes and select "Only future tasks" (Apply to all following instances).
​6. Observe the task lis from the planner. 

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

[SP-logs.json](https://github.com/user-attachments/files/27100305/SP-logs.json)

### documentation gap

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.
- [ ] I can help improve the documentation.

Issue #4931, opened 2025-08-10
Collection: Repeat Task / Recurring Task Improvement


johannesjo:
Collection of tasks related to recurring/repeatable tasks.
 
- [x] #4928
- [ ] #4704
- [ ] #4629
- [ ] #4854
- [ ] #4836
- [ ] #4750 
- [ ] #4835
- [ ] #4958
- [ ] #4931
- [x] #4932
- [x] #4933
- [x] #4915 
- [ ] #5808

AnHumanLearner:
Perhaps we can add also these issues : 
* [ ] [💡 Check a task planned/repeated before its due date #4835](https://github.com/johannesjo/super-productivity/issues/4835)
* [ ] [💡 management of overdue repeated tasks #4750](https://github.com/johannesjo/super-productivity/issues/4750)

(Less related)
* [ ] [💡 Make a task on several dates #4836](https://github.com/johannesjo/super-productivity/issues/4836) 

The issues of @OLLI-S are also relevant I think

langerhorst:
I used rememberthemilk years ago and there was a nice feature with recurring tasks. I was able to make a difference whether the recurring task should be repeated every 7 days for example OR after 7 days when completed. This makes a big difference. Example given: Cleaning up AFTER 7 days make more sense than Cleaning up every 7 days. Because if I come back from holiday this could be stretched. I'd love to have this feature. There should be a difference between EVERY x days and AFTER x days when completing the task. How do you think about?

OLLI-S:
@langerhorst: I really love your suggestion, I used "Remember the Milk" some years ago too. 
Can you please create a separate feature request for your suggestion?
Because this topic here is a collection of existing issues and each issue can be implemented separate.
Thank you very much!

ReduxST:
I think this belongs here too: [Delete one instance of a repeated event #4195
](https://github.com/johannesjo/super-productivity/issues/4195)

langerhorst:
> [@langerhorst](https://github.com/langerhorst): I really love your suggestion, I used "Remember the Milk" some years ago too. Can you please create a separate feature request for your suggestion? Because this topic here is a collection of existing issues and each issue can be implemented separate. Thank you very much!

Done and also added a [mockup](https://github.com/johannesjo/super-productivity/issues/5080):

Cyber-Syntax:
This also belongs here: #5808

Samiran357:
#6269

PR #7948, opened 2026-06-02
feat(task-repeat): RFC 5545 RRULE recurring schedules — EPIC · Phase 1/13 (Closes #4020)

### LINKED ISSUE #4020: Enhanced Repeating Schedule
## Problem Statement
I love this application and it is almost perfect for all my use cases. Thank you for making such a great app. I do however want a way to to mark maintenance tasks like lawn care for my many apartments. I do this once a week for the months of March - November (skipping winter). With the current Schedule Repeat system, I cannot make a single recurring Schedule to cover all of what I need. I would have to make like 36 tasks with the once a year setting.

In cron format it would look like this:
**0 0 * 3-11 6**

in cron natural language that would look like this:
**"once a week on saturday from March through November"**

there are js libraries that can convert between the two. 
[english to cron js lib](https://github.com/darkeyedevelopers/natural-cron.js) ([demo](https://cronify.zimo.li/))
[cron to english js lib](https://github.com/bradymholt/cronstrue) ([demo](https://crontab.guru/#0_0_*_3-11_6))
[scheduler](https://github.com/node-cron/node-cron)

It would be amazing if I can set a toggle in the repeat/scheduling section to just put in a cron format or natural language cron.
what would be even more amazing is if i could right in the task itself do something like this

**TASK:** **"Mow the Lawn +141 Main St #Maintenance  @+once a week on saturday from March through November /2h"**

with **@+** being the symbol to represent enhanced scheduling.. (i dont really care what the symbol would be).

omega-tree:
# 🗂️ EPIC — RFC 5545 RRULE recurring schedules (Closes #4020)

> **Umbrella / "Large Goal" PR.** This tracks the whole recurring-schedules
> overhaul. It lands in **phases**, each shipped as its own focused sub-MR and
> checked off here as it merges. The full RRULE implementation already exists on
> the waypoint branch `feat/recurring-full` (reference); each phase is a
> reviewable slice cut from it. Phases 10–13 are **new scope** — Donetick-parity
> features not yet built. All curator review fixes land in Phase 1.

---

## Phase checklist (each → its own sub-MR)

| ✓ | Phase | Scope | Sub-MR | Status |
|---|---|---|---|---|
| ☐ | **1 — Core** | RFC 5545 occurrence engine (UTC/local-noon, DST-safe, fail-soft) · structured RRULE builder · legacy⇄RRULE migration (both directions) · live **text** preview (humanized + next date) · quick-setting presets · curator fixes (quickSetting forward-compat, malformed-rrule legacy fallback, reverse-migration weekday, no rule-body logging, DRY) · property/invariant/fuzz tests | _this PR_ | 🚧 in progress |
| ☐ | **2 — Heatmap preview + simulation** | 365-day occurrence heatmap in the dialog; click a day to simulate completing it and re-anchor — makes complex RRULEs *visible* before saving | TBD | ⏸ built, deferred |
| ☐ | **3 — Natural language `@+`** | `@+<phrase>` → RRULE parser + add-task-bar wiring + live "humanized rule · next date" preview | TBD | ⏸ built, deferred |
| ☐ | **4 — Due-date derivation** | per-instance Due derived from the appears day: offset / until-next / period-end / fixed / from-completion / none | TBD | ⏸ built, deferred |
| ☐ | **5 — Ends after N completions** | end condition that stops after N completed instances (needs min-client-version gate) | TBD | ⏸ built, deferred |
| ☐ | **6 — Missed-occurrence backfill** | "create a task for each missed occurrence" catch-up (+ build-set-once perf) | TBD | ⏸ built, deferred |
| ☐ | **7 — REST API recurring** | create recurring tasks over the local REST API (rrule/startDate/from-completion) — #7239 | TBD | ⏸ built, deferred |
| ☐ | **8 — RECURRENCE-ID overrides** | edit a single occurrence (move / re-time / re-title) via RDATE+EXDATE; dialog "Edit this occurrence" | TBD | ⏸ built, deferred |
| ☐ | **9 — iCal / RRULE export** | export recurrences as `.ics` / RRULE strings | TBD | ⬜ not built |
| ☐ | **10 — Adaptive scheduling** | learn per-task completion cadence (exp-decay weighted avg of historical delays, à la Donetick) → suggest / auto-adjust the next due; opt-in per repeat cfg. **Must be op-log-deterministic** (derive from recorded completions, not wall-clock) | TBD | 🔭 new — Donetick parity |
| ☐ | **11 — Trigger-based recurrence** | next occurrence fires on an event / state-change condition instead of a clock (Donetick `trigger` / "Things") | TBD | 🔭 new — Donetick parity |
| ☐ | **12 — Sub-daily / hourly recurrence** | extend the day-granular engine to interval-hours + multiple-per-day (`FREQ=HOURLY`/`BYHOUR`); revisits the local-noon/DST model — **largest of the four** | TBD | 🔭 new — Donetick parity |
| ☐ | **13 — Multiple reminders per occurrence** | more than one reminder offset per recurring instance | TBD | 🔭 new — Donetick parity |

Legend: ☐ phase open · 🚧 in progress · ⏸ implemented on waypoint, awaiting its slice · ⬜ not started · 🔭 newly scoped, not built.

*(Donetick's **assignee rotation / round-robin** is intentionally **out of scope** — SP's recurrence is single-assignee.)*

---

## Feature comparison (RFC 5545 baseline, expanded)

**SP now** = released Super Productivity (master, legacy repeat). **SP this epic** = after this epic lands.
`*Google` = Google **Calendar** (RRULE-complete via API/import; the custom-recurrence GUI is limited; Google **Tasks** is far weaker — no weekday pick, no single-occurrence edit, no API recurrence).
**Donetick** = open-source self-hosted chore manager (closest OSS peer; optimizes people + cadence-learning over calendar precision).
**Build** = status on the SP waypoint: ✓ built · ☐ planned · — not planned.

Markers: ✅ full · ➖ partial / API-only / limited · ❌ none · 🟢 SP-distinct.

| Feature | SP now | RFC 5545 | SP this epic | Google* | Todoist | Things 3 | TickTick | Donetick | **Build** |
|---|---|---|---|---|---|---|---|---|---|
| **— Frequency —** | | | | | | | | | |
| Basic D/W/M/Y | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✓ |
| Every N interval | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✓ |
| Weekday selection | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ➖ | ✅ | ✓ |
| Nth weekday of month (`2TU`) | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ | ✅ | ✓ |
| 🟢 Per-day ordinals (`3MO,4SU`) | ❌ | ✅ | ✅ | ➖ | ❌ | ❌ | ❌ | ➖ | ✓ |
| Last day / last weekday | ✅ | ✅ | ✅ | ✅ | ➖ | ➖ | ❌ | ➖ | ✓ |
| Last business day (`BYSETPOS=-1`) | ❌ | ✅ | ✅ | ➖ | ➖ | ❌ | ❌ | ❌ | ✓ |
| Multiple month-days (`1,15`) | ❌ | ✅ | ✅ | ➖ | ❌ | ➖ | ❌ | ➖ | ✓ |
| Seasonal `BYMONTH` | ❌ | ✅ | ✅ | ➖ | ❌ | ❌ | ❌ | ➖ | ✓ |
| 🟢 `BYWEEKNO`/`BYYEARDAY`/`WKST` | ❌ | ✅ | ✅ | ➖ | ➖ | ❌ | ❌ | ❌ | ✓ |
| **— End conditions —** | | | | | | | | | |
| Never | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✓ |
| On date (`UNTIL`) | ❌ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ | ✓ |
| After N occurrences (`COUNT`) | ❌ | ✅ | ✅ | ✅ | ❌ | ✅ | ✅ | ❌ | ✓ |
| End after N completions | ❌ | ➖ | ✅ | ❌ | ❌ | ➖ | ➖ | ❌ | ✓ Ph5 |
| **— Occurrence control —** | | | | | | | | | |
| Skip / exclude one (`EXDATE`) | ✅ | ✅ | ✅ | ✅ | ➖ | ❌ | ✅ | ❌ | ✓ |
| Edit one occurrence (`RECURRENCE-ID`) | ❌ | ✅ | ✅ | ✅ | ➖ | ❌ | ✅ | ❌ | ✓ Ph8 |
| **— Completion-relative —** | | | | | | | | | |
| After-completion scheduling | ✅ | ❌ | ✅ | ❌ | ✅ | ✅ | ✅ | ✅ | ✓ |
| Configurable gap after completion | ✅ | ❌ | ✅ | ❌ | ➖ | ✅ | ✅ | ✅ | ✓ |
| Wait-for-completion (no pile-up) | ✅ | ❌ | ✅ | ❌ | ➖ | ✅ | ➖ | ✅ | ✓ |
| Adaptive (learns cadence) | ❌ | ❌ | ✅ | ❌ | ❌ | ❌ | ❌ | ✅ | ☐ Ph10 |
| Trigger-based (state-change fires) | ❌ | ❌ | ✅ | ❌ | ❌ | ❌ | ❌ | ✅ | ☐ Ph11 |
| **— Time / reminders —** | | | | | | | | | |
| Specific time-of-day | ✅ | ✅ | ✅ | ✅ | ✅ | ➖ | ✅ | ✅ | ✓ |
| Hourly / sub-daily / multi-per-day | ❌ | ✅ | ✅ | ✅ | ✅ | ❌ | ❌ | ✅ | ☐ Ph12 |
| Reminder lead-time per occurrence | ✅ | ➖ | ✅ | ➖ | ✅ | ✅ | ✅ | ✅ | ✓ |
| Multiple reminders per occurrence | ❌ | ➖ | ✅ | ➖ | ❌ | ➖ | ✅ | ✅ | ☐ Ph13 |
| **— Entry —** | | | | | | | | | |
| Natural-language entry | ❌ | ➖ | ✅ | ➖ | ✅ | ➖ | ✅ | ✅ | ✓ Ph3 |
| 🟢 `@+` NL → RRULE + next-date preview | ❌ | ❌ | ✅ | ❌ | ➖ | ❌ | ➖ | ❌ | ✓ Ph3 |
| 🟢 Raw RRULE override (UI) | ❌ | ✅ | ✅ | ❌ | ❌ | ❌ | ➖ | ❌ | ✓ |
| **— Preview —** | | | | | | | | | |
| Occurrence list / calendar preview | ❌ | ➖ | ✅ | ➖ | ➖ | ❌ | ✅ | ❌ | ✓ |
| 🟢 Heatmap occurrence preview | ❌ | ➖ | ✅ | ❌ | ❌ | ❌ | ❌ | ❌ | ✓ Ph2 |
| 🟢 Completion **simulation** preview | ❌ | ❌ | ✅ | ❌ | ❌ | ❌ | ❌ | ❌ | ✓ Ph2 |
| **— Derivation / backfill —** | | | | | | | | | |
| 🟢 Per-instance due-date derivation | ❌ | ➖ | ✅ | ❌ | ➖ | ❌ | ➖ | ❌ | ✓ Ph4 |
| 🟢 Create task per missed occurrence | ❌ | ➖ | ✅ | ❌ | ❌ | ❌ | ❌ | ❌ | ✓ Ph6 |
| **— Teams / projects / habits —** | | | | | | | | | |
| Assignee rotation / round-robin | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ✅ | — out of scope |
| Repeating projects w/ checklist | ❌ | ❌ | ❌ | ❌ | ❌ | ✅ | ❌ | ➖ | — |
| Habit-tracking subsystem | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ✅ | ❌ | — |
| **— Interop —** | | | | | | | | | |
| iCal / RRULE export | ❌ | ✅ | ✅ | ✅ | ✅ | ❌ | ✅ | ❌ | ☐ Ph9 |
| iCal / RRULE import | ➖ | ✅ | ➖ | ✅ | ❌ | ➖ | ➖ | ❌ | ✓ |
| REST / API create recurring | ❌ | ➖ | ✅ | ✅ | ✅ | ❌ | ✅ | ✅ | ✓ Ph7 |

---

## 🟢 Genuinely SP-distinct (no mainstream rival matches)
- **Completion simulation** preview — click a day, the series re-anchors. *Nobody else.*
- **Create a task per missed occurrence** — true catch-up, not just "next". *Nobody else.*
- **Heatmap** 365-day occurrence preview. *Nobody else.*
- **Per-day ordinals** (`3MO,4SU`), **seasonal `BYMONTH`**, **`BYWEEKNO`/`BYYEARDAY`/`WKST`**, **raw RRULE override** in the *UI* — Google only via API/import; the mainstream apps not at all.
- **Per-instance due-date derivation** (due = appears + offset / until-next / period-end / fixed / from-completion / none).

---

## Notes
- **Forward-compat (Phase 1):** new `quickSetting` literals (incl. `RRULE`) are never persisted — saved cfgs use a `master`-safe value (`CUSTOM`); the rich value drives the dialog UI in-memory only. typia on old/mobile clients stays happy.
- Engine internals stay UTC; the opaque `rrule` string keeps `repeatCycle` within the old enum subset → old clients ignore unknown fields and fall back to legacy schedule.
- Each deferred / new phase carries its own sync surface + discussion; this epic stays the index.


johannesjo:
Hi Filip – thank you very much for this!

enuinely: the occurrence engine, the DST / cross-timezone handling, and the property/fuzz/metamorphic test suite are some of the most careful work I've seen in a first PR, and you clearly nailed the real problem in #4020.

I'm not going to merge it, and I want to be transparent about why – none of it is about the quality of your code, it's three project-level constraints:

**1. The bundled `crono-eng.wasm`.** I can't ship a ~95 KB binary that can't be rebuilt or audited from this repo (it needs an external Zig project + toolchain). It's also redundant – the app already depends on `chrono-node`, which does English→date parsing without a new binary.

**2. Forward-compatibility / sync.** Adding `'CRON'` to the `repeatCycle` enum collides with how our sync validation works: older clients (especially mobile, which updates slowly) validate synced data against a fixed set of `repeatCycle` values and will reject any task using `'CRON'` as corrupt — on the live sync path, with no rollback. So a single cron task on an up-to-date device can break sync on an older device in the same account.

**3. Direction.** I've been planning to move recurrence to **RFC 5545 RRULE** (see `docs/research/recurring-events-implementation-plan.md` and #4931). RRULE is what calendars standardize on; it expresses everything cron does and more — every-other-week, end-after-N, nth-weekday, seasonal months, skipped dates — in a day-oriented way that fits our engine, and it's stored as an opaque string, which sidesteps the sync problem in (2). cron is the one option that doesn't fit that direction, and once it's in synced data it's hard to walk back.

Here's what I'd like to do instead, because your work shouldn't go to waste:

- **Short term:** I'll add a small "only in these months" + end-date option to the existing weekly/monthly UI, which closes your #4020 directly.
- **Bigger picture:** your occurrence engine and especially your **timezone/DST test harness are exactly what the RRULE work needs** – honestly they're ahead of my own plan on the hard parts (DST-safe day resolution, failing soft on bad input, real cross-timezone tests). I'd be glad if you wanted to lead or co-author that effort (#4931), with your engine + tests as the foundation.

If you'd rather land the cron *engine* sooner as a stepping stone, we can scope a trimmed version (engine + preview + tests, minus the WASM, the natural-language layer, `@+`, and the missed-occurrence backfill) – but it would still need the forward-compat handling above first.

Either way – thank you! This pushed the recurrence conversation further than anything has in a while.

omega-tree:
Thank you for pointing out the flaws in my approach. 

> _1. The bundled crono-eng.wasm. I can't ship a ~95 KB binary that can't be rebuilt or audited from this repo (it needs an external Zig project + toolchain). It's also redundant – the app already depends on chrono-node, which does English→date parsing without a new binary_

your absolutely right! i overlooked my own library after wasm compilation. 


> 3. Direction. I've been planning to move recurrence to RFC 5545 RRULE (see docs/research/recurring-events-implementation-plan.md and https://github.com/super-productivity/super-productivity/issues/4931). RRULE is what calendars standardize on; it expresses everything cron does and more — every-other-week, end-after-N, nth-weekday, seasonal months, skipped dates — in a day-oriented way that fits our engine, and it's stored as an opaque string, which sidesteps the sync problem in (2). cron is the one option that doesn't fit that direction, and once it's in synced data it's hard to walk back.

I was not aware of RRULE and how feature rich it is, even more so then quartz/cronos. I would be happy to implement that in-place of cron. thank you for pointing me in the right direction. furthermore cron's hours/Min/seconds are dead-weight in the app and doesn't really fit.

Hold off on spending additional time on this, I'll see what i can do.

johannesjo:
> I was not aware of RRULE and how feature rich it is, even more so then quartz/cronos. I would be happy to implement that in-place of cron. thank you for pointing me in the right direction. furthermore cron's hours/Min/seconds are dead-weight in the app and doesn't really fit.
> 
> Hold off on spending additional time on this, I'll see what i can do.

That's great to hear! I will re-evaluate the RRULE plan today. Would love to hear your thoughts on it!

johannesjo:
Filip — this is a great pivot, and the engine is exactly where I hoped it'd go. The UTC-internal / local-noon occurrence math, the fail-soft parsing, the round-trip guard in the builder, and the property/invariant tests are all genuinely strong. WASM + cron-parser + cronstrue are gone and `rrule` (opaque string, `repeatCycle` shrunk to a subset) is the right shape. Concerns #1 and #3 from before: resolved. 🎉

Before this goes further I did a deep review, and there's **one issue that has to be fixed, and a scope question** I want to settle with you first.

## 🔴 The forward-compat problem (#2) came back — through `quickSetting`

This is subtle, so bear with me. We correctly kept `rrule` opaque and `repeatCycle` within the known set. But `RepeatQuickSetting` grew 8 new literals (`EVERY_OTHER_DAY`, `BIWEEKLY_CURRENT_WEEKDAY`, `WEEKENDS`, `MONTHLY_LAST_WEEKDAY`, `QUARTERLY_CURRENT_DATE`, `SEMIANNUALLY_CURRENT_DATE`, `EVERY_OTHER_YEAR_CURRENT_DATE`, `RRULE`), and `quickSetting` is a **required, synced** field that gets persisted with those values (the dialog save path keeps the form's value; the REST path hardcodes `'RRULE'`).

Our sync validation is typia (`createValidate<TaskRepeatCfgState>()` in `op-log/validation/validation-fn.ts`). typia tolerates *unknown extra fields* — so `rrule`, `endsAfterCompletions`, the `due*` fields are all fine on old clients — but it **rejects an existing required field whose value is outside the union it was compiled with.** So an older client (mobile especially) receiving a task with `quickSetting: 'WEEKENDS'` treats the data as corrupt on the live sync path. That's the exact failure mode that made me say no to `repeatCycle: 'CRON'`, just relocated one field over.

The good news: it's cheap to fix, because `quickSetting` is really just a UI hint — at runtime we only branch on `'RRULE'` vs `'CUSTOM'`; the other presets are pure labels and the actual schedule already lives in the legacy fields + `rrule`. So: **don't persist the new preset literals.** Drive the UI from them in-memory, but store an existing value (`'CUSTOM'`, or `'RRULE'` for the builder) — `rruleToFormModel` already reconstructs the builder state on edit, so nothing is lost. Same fix for `_buildRRuleRepeatCfg` in the REST handler.

## 🟡 Scope — let's land the core first

I'd asked to hold and trim to **engine + preview + tests**, and this WIP instead bundles the natural-language layer, `@+`, missed-occurrence backfill, plus a few things I hadn't even raised: a 7-mode due-date-derivation subsystem (`dueType` & friends + the due-date config component), `endsAfterCompletions`, recurring-task creation over the REST API (#7239), and the heatmap. Each of those is its own feature with its own sync surface and deserves its own PR and discussion. Could we cut this first PR down to the engine + builder + legacy migration + preview + tests? It makes review tractable and keeps the forward-compat surface small. (I'm also still re-evaluating the RRULE plan on my side and want to share that with you before you build more.)

## A few real bugs I found (worth fixing whenever the relevant code lands)

- **Malformed `rrule` silently stops a task.** `getNext/Newest/FirstRepeatOccurrence` dispatch `if (cfg.rrule) return …RRule…`, and the engine returns `null` on an unparseable rule — so the task just stops generating instead of falling back to the legacy fields we deliberately keep populated. The heatmap path already does `cfg.rrule || legacyTaskRepeatCfgToRRule(cfg)`; the engine dispatch should fall back the same way.
- **Reverse migration can produce a dead schedule on old clients.** `rruleToLegacyTaskRepeatCfg` for a BYDAY-less `FREQ=WEEKLY` (reachable via `@+weekly` or a raw override) resets all weekday flags to false, and the legacy WEEKLY engine requires a flag set — so on an old client it never fires (new client fires). The forward converter guards this (falls back to the start weekday); the reverse one should too.
- **`endsAfterCompletions` diverges across versions.** Old clients ignore the field and keep materializing forever, so "ends after N" silently doesn't on un-updated devices. If this feature stays, it probably needs a min-client-version gate rather than "self-healing" framing.

## Smaller notes

- **Don't log the rule body.** `rrule-occurrence.util.ts` does `Log.warn(`Invalid RRULE "${input.rrule}"`)`. The raw-override field makes `rrule` free-text user input, and log history is exportable (rule #9) — log only the cfg id. (The header comment says "never user content," which is the right instinct — let's make the code match it.)
- **Reuse the lib instead of re-deriving it.** `rrule` ships `RRule.fromText`/`toText` (English ↔ rule) and `optionsToString` (canonical string building). The hand-rolled English parser deliberately covers phrasings `fromText` misses ("weekends", "every other tuesday", month ranges), so it's not a clean swap — but if the NL layer returns later, a `fromText`-first + custom-fallback shape would shrink it a lot. (chrono-node, which I mentioned earlier, isn't the right tool here — it parses one-shot dates, not recurrence. My mistake.)
- **DRY:** `normalizeWeekdays`/`toNumArray` are byte-identical in `rrule-form.util.ts` and `legacy-cfg-to-rrule.util.ts` — worth one shared module.
- **Perf (only if the missed-backfill path stays):** `getAllMissedDueDates` calls into the engine in a loop, and `_buildRuleSet` re-parses + rebuilds the `RRuleSet` every call, so a long catch-up re-parses the same rule many times. Build the set once and walk it. Separately, the `endsAfterCompletions` archive read runs per-cfg on every app-open fan-out (same pattern as `waitForCompletion`) — fine for now, but worth a single archive snapshot per pass if both stay.

None of this changes my read that the engine is the strongest part of the proposal. Fix the `quickSetting` persistence, let's agree on a trimmed first slice, and I'm happy to keep iterating with you. Thanks again — this is good work.

omega-tree:
Ok, ill scope down. eagerly awaiting your RRULE plan adjustments.

johannesjo:
One correction to my own review before you start re-scoping — I gave you a slightly-wrong fix and I'd rather you not build on it. 🙂

### `'RRULE'` isn't a safe value to persist either

I wrote "store an existing value (`'CUSTOM'`, or `'RRULE'` for the builder)" — but `'RRULE'` is itself one of the new literals (I even listed it as such two paragraphs up 😅). On `master` the `RepeatQuickSetting` union is exactly:

```
'DAILY' | 'WEEKLY_CURRENT_WEEKDAY' | 'MONTHLY_CURRENT_DATE' | 'MONTHLY_FIRST_DAY'
| 'MONTHLY_LAST_DAY' | 'MONTHLY_NTH_WEEKDAY' | 'MONDAY_TO_FRIDAY'
| 'YEARLY_CURRENT_DATE' | 'CUSTOM'
```

So persisting `quickSetting: 'RRULE'` hits the *exact same* typia rejection on older clients as `'WEEKENDS'` would. The only forward-safe persisted value is one of those nine — i.e. **`'CUSTOM'`** for every new-literal case (builder, presets, REST, `@+`).

And it's lossless: `_processQuickSettingForDate` already promotes any `rrule`-bearing cfg whose `quickSetting` isn't a *faithful* known preset back to in-memory `'RRULE'` on open (and `'CUSTOM'` is deliberately not in `KNOWN_PRESETS`). So persist `'CUSTOM'` + the opaque `rrule`, and the builder reconstructs on edit. Just keep the clamp at the **persist boundary** — the in-memory working value has to stay `'RRULE'` or `isRRuleMode` stops rendering the builder mid-edit.

### Two more write sites than I listed

I named the dialog-save and REST paths, but the new literal also gets persisted from:

- the `@+` inline path — `add-task-bar.component.ts` does `addTaskRepeatCfgToTask(…, { quickSetting: 'RRULE' })`
- the dialog **edit** path — `getTaskRepeatCfgChanges` → `updateTaskRepeatCfg`

Four sites total. Rather than patch each, I'd centralize: a `toPersistedQuickSetting(qs)` (return `qs` if it's in the `master` set, else `'CUSTOM'`) applied in the `addTaskRepeatCfgToTask` / `updateTaskRepeatCfg` reducers covers all of them and is future-proof.

---

### Scope — since this is 1/13

The bar for this slice is just "doesn't break old/mobile clients on the live sync path" — it does **not** need to be feature-complete. So the forward-compat clamp above is really the one true blocker. Everything else I flagged (the missed-backfill perf, `endsAfterCompletions` gating, log-redaction, lib reuse, DRY) only matters once its respective feature lands — totally fine to carry forward into the slice that introduces it.

Two I'd still *gently* suggest pulling in **if** the relevant code stays in slice 1, since they're about not-silently-breaking rather than completeness:

- **Reverse-migration weekday guard** — a BYDAY-less `FREQ=WEEKLY` (reachable via `@+weekly` or a raw override) resets every weekday flag to `false` in `rruleToLegacyTaskRepeatCfg`, so the task never fires on old clients while new ones fire. One-line mirror of the forward converter's "fall back to the start weekday" guard.
- **Malformed-rrule fallback** in the three engine dispatchers (`getNext/First/NewestPossible…`) — they branch on `if (cfg.rrule)` and the engine returns `null` on an unparseable rule, so the task silently goes dormant instead of falling back to the legacy fields we deliberately keep populated. `cfg.rrule || legacyTaskRepeatCfgToRRule(cfg)` (like the heatmap already does), or gate on the `isRRuleValid` you already have.

Neither blocks merge; both are cheap insurance against a *synced* task going dead on a device that can't parse the rule.

Thanks again — really looking forward to the trimmed slice. 🙌

omega-tree:
Ha, no worries — I'd tripped over the same thing while re-scoping. `master`'s union is just those nine values, so `'RRULE'` blows up on old clients exactly like `'WEEKENDS'` would. Quick rundown of where it landed:

- **Persist `'CUSTOM'`, keep the rich literal in the form.** Anything that isn't a real master preset persists as `'CUSTOM'`; `'RRULE'` and the new presets live purely in the dialog's form state, and on reopen it's re-promoted back to `'RRULE'` from the `rrule` string — so nothing's lost, same flow you described.
- **One flag before you bake in the reducer idea: the clamp can't live in the reducer.** The op-log replays the action *payload*, not reduced state — `operation-capture.service.ts` stashes `operation.payload.actionPayload` and the remote re-dispatches it. So a reducer clamp would clean up my own local state but still ship an op with `quickSetting:'RRULE'` in it, and the old/mobile client would run that through its own un-clamped reducer and hit the typia rejection anyway. Same bug, one hop downstream.
- **So I moved your "centralize it" instinct one layer up** — into the action creators (`addTaskRepeatCfgToTask` / `updateTaskRepeatCfg` / `updateTaskRepeatCfgs`). Same single-source, future-proof win — when `@+` and REST come back in phases 2/7 they dispatch these same actions and inherit the clamp for free — but it's the spot that actually keeps the bad value off the wire. The scattered call-site clamps in the dialog and add-task-bar are gone now.
- **Malformed-rule fallback is in** — the three dispatchers gate on `isRRuleValid` and fall back to the legacy fields, so a busted rule can't silently kill a synced task.
- **Reverse-migration weekday guard is in** — a BYDAY-less `FREQ=WEEKLY` now falls back to the start weekday in `rruleToLegacyTaskRepeatCfg`, so old clients keep firing.
- **Mobile's covered by the same fix** — it runs the same typia validation, so the clamp keeps the wire value in-union and the legacy fields stay populated; an older mobile client just reads `'CUSTOM'`, ignores the `rrule` it doesn't understand, and falls back to the legacy schedule.
- **Everything else** (backfill perf, the `endsAfterCompletions` gate, log redaction, lib reuse) rides along with whichever phase reintroduces that code — none of it's lost, it's all on the `feat/recurring-full` branch, and the epic up top tracks each piece as its own slice.

johannesjo:
Thanks for the re-scope and the quickSetting follow-up. I double-checked the updated slice with a few focused passes. The action-creator clamp looks like the right boundary for sync/op-log safety; I do not see the original quickSetting value escaping through the normal add/update/update-many paths.

A few issues still look worth fixing before this lands:

### 1. Legacy CUSTOM migration is not always behavior-preserving

`_processQuickSettingForDate()` migrates legacy `CUSTOM` configs to RRULE on dialog open/save:

- `src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.ts`
- `src/app/features/task-repeat-cfg/util/legacy-cfg-to-rrule.util.ts`

The converter itself documents the edge difference: legacy monthly/yearly recurrence clamps missing dates, while RRULE skips them. So a legacy monthly task anchored on Jan 31, or a yearly task anchored on Feb 29, changes behavior after the user opens and saves the repeat dialog: old behavior creates Feb 28 / non-leap-year Feb 28 occurrences, RRULE skips those periods. I think these cases either need to stay on the legacy path, or the migration needs explicit compatible semantics/tests.

### 2. RRULE -> legacy fallback is still not faithful for date-based monthly/yearly rules

`onRRuleChange()` only merges the partial result from `rruleToLegacyTaskRepeatCfg()`:

```ts
...rruleToLegacyTaskRepeatCfg(rrule, cfg.startDate)
```

But `rruleToLegacyTaskRepeatCfg()` does not emit an aligned `startDate` for `BYMONTHDAY`/`BYMONTH`, and it also does not clear stale monthly anchors. Example: `FREQ=MONTHLY;BYMONTHDAY=15` with `startDate: '2024-06-03'` fires on the 15th on new clients, but old clients/fallback use the legacy month-day from `startDate` and fire on the 3rd. If the previous config had `monthlyWeekOfMonth/monthlyWeekday`, those stale anchors can survive too. The comment in `legacy-cfg-to-rrule.util.ts` says the caller keeps `startDate` aligned, but the caller does not currently do that.

### 3. Last-day RRULE loses its old-client fallback on save

`rruleToLegacyTaskRepeatCfg('FREQ=MONTHLY;BYMONTHDAY=-1')` correctly returns `monthlyLastDay: true`, but `_normalizeMonthlyAnchor()` strips `monthlyLastDay` for every quickSetting except `MONTHLY_LAST_DAY`. RRULE-builder saves have `quickSetting === 'RRULE'`, so a custom last-day rule keeps month-end behavior on new clients but old clients/fallback lose the month-end flag and fall back to the numeric `startDate` day.

### 4. The missed-backfill option is exposed but not implemented in this slice

The advanced form still includes `createForEachMissed`, and the wiki/translation copy says it creates one task per missed occurrence. I could not find any scheduling/model support for that flag in this trimmed PR; the RRULE tests still describe "newest missed only, no backfill". That means users can enable a setting that silently does nothing. I would remove/defer the form field and copy until the backfill phase lands, or land the full behavior with coverage in this PR.

### 5. Fresh Yearly builder rules can fire monthly

`defaultRRuleFormModel()` seeds `monthDays` from the reference date but leaves `byMonth` empty. If the user opens Custom recurring config, switches frequency to Yearly, and saves the default "on date" mode, `formModelToRRule()` emits `FREQ=YEARLY;BYMONTHDAY=<day>` without `BYMONTH`. In RRULE semantics that matches that day in every month, not once per year. The builder should seed/require `BYMONTH` for yearly date modes, and this path needs a unit/e2e regression.

### 6. Date-dependent quick-setting changes ignore the selected start date initially

The quick-setting option labels are rebuilt from the current `startDate`, but the select change handler calls `getQuickSettingUpdates(event.value)` without passing that date. For presets that write `startDate` (`MONTHLY_CURRENT_DATE`, quarterly/semiannual, yearly current date, etc.), selecting the preset after choosing a future start date overwrites the model with today before save; the save-time recompute then uses that overwritten date. Weekly flags are corrected on save, but date-writing presets still lose the user-selected anchor.

### Smaller cleanup

`package-lock.json` still has root entries for `cron-parser` and `cronstrue`, while `package.json` only adds `rrule` and I do not see source imports for the cron packages. Looks like stale lockfile churn from the earlier cron version.

I was not able to run the targeted specs locally because this worktree does not have the new dependency installed (`TS2307: Cannot find module 'rrule'`), so this pass is code-review based.

omega-tree:
Ok i don't see any issue with those 6 critiques. Will Fix.

johannesjo:
Thank you!

omega-tree:
Moved the **heatmap + completion-simulation preview up to Phase 2** (was 6): the 365-day calendar simulation makes complex RRULEs — per-day ordinals, BYSETPOS narrowing, seasonal BYMONTH, clamp rules — immediately *visible and verifiable* while you build them. That visualization pays off for every later phase and for reviewing this one, so it ships right after Core.

johannesjo:
Review findings after re-checking current head `acbd243`:

1. **Required:** `save()` persists the aligned `startDate`, but the RRULE engine uses `startDate` as `dtstart` and documents it as anchoring `INTERVAL` / `COUNT` / `UNTIL`. `getAlignedStartDate()` does not account for `INTERVAL`, and for clamp-idiom rules it skips valid shorter-month occurrences. Example: `FREQ=MONTHLY;INTERVAL=2;BYMONTHDAY=15` from `2024-06-20` gets saved as `2024-07-15`, shifting the actual RRULE cadence. `FREQ=MONTHLY;BYMONTHDAY=31,-1;BYSETPOS=1` from `2024-02-10` similarly skips the valid Feb 29 occurrence by aligning to Mar 31.
https://github.com/super-productivity/super-productivity/blob/acbd243583954f969f7e119a854a45dc39dc40a0/src/app/features/task-repeat-cfg/util/legacy-cfg-to-rrule.util.ts#L230-L310

2. **Required:** In RRULE mode, legacy fallback fields are only re-derived when `getAlignedStartDate()` returns a new date. If the user changes `startDate` after the last builder emit and alignment does not apply, e.g. raw/imported `FREQ=WEEKLY` with no `BYDAY`, the saved RRULE uses the new `dtstart` while older clients keep stale weekday booleans from the old start date.
https://github.com/super-productivity/super-productivity/blob/acbd243583954f969f7e119a854a45dc39dc40a0/src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.ts#L396-L407

3. **Required:** Persisting `null` for `monthlyWeekOfMonth` / `monthlyWeekday` breaks the forward-compat story for older/mobile clients. Their schema expects these known fields to be absent or numeric; `null` is not a master-safe value, and the action boundary only clamps `quickSetting`.
https://github.com/super-productivity/super-productivity/blob/acbd243583954f969f7e119a854a45dc39dc40a0/src/app/features/task-repeat-cfg/task-repeat-cfg.model.ts#L104-L109
https://github.com/super-productivity/super-productivity/blob/acbd243583954f969f7e119a854a45dc39dc40a0/src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/get-quick-setting-updates.ts#L27-L35

4. **Required:** The `BYSETPOS=0` cleanup is undone by the raw-override guard. The parser drops `0` from `bySetPos`, but the canonical mismatch then stores the original RRULE in `rawOverride`, so saving still re-emits the dead `BYSETPOS=0` rule. The test should assert the emitted rule/raw override, not only `m.bySetPos`.
https://github.com/super-productivity/super-productivity/blob/acbd243583954f969f7e119a854a45dc39dc40a0/src/app/features/task-repeat-cfg/util/rrule-form.util.ts#L327-L362

johannesjo:
## Review notes — legacy-fallback validity + a couple of process points

Reviewed against the current head (`f5276407de`). The additive / dual-write design is sound, but I found two cases where the legacy-fallback derivation (`rruleToLegacyTaskRepeatCfg`) can emit values **outside the synced model's allowed set**, with no validation gate before they reach the op-log. Both are reachable from the shipped builder. Details + suggested fixes below.

### 1. ❗ Out-of-range `monthlyWeekOfMonth` can be persisted from a supported builder flow

`rruleToLegacyTaskRepeatCfg` blind-casts the BYDAY ordinal — `out.monthlyWeekOfMonth = weekdays[0].n as MonthlyWeekOfMonth` (`src/app/features/task-repeat-cfg/util/legacy-cfg-to-rrule.util.ts:195`), guarded only by `!= null`. The model allows `MonthlyWeekOfMonth = 1|2|3|4|-1` (`task-repeat-cfg.model.ts:65`), but the builder's custom-ordinal input clamps only to `±nthPosBound` = 5 (monthly) / 53 (yearly) (`rrule-builder/rrule-builder.component.ts:294-298`). So `FREQ=MONTHLY;BYDAY=5MO` → `monthlyWeekOfMonth: 5` and `BYDAY=-2MO` → `-2`, both outside the union.

The persist boundary doesn't catch it — `_stripNullAnchors` only normalizes `null` and `toSyncSafeQuickSetting` only clamps `quickSetting` (`store/task-repeat-cfg.actions.ts:19-42`). The value reaches the op-log, and both old and new clients typia-validate `monthlyWeekOfMonth` against `1|2|3|4|-1` → an out-of-range value fails validation and triggers the repair / restore-from-backup path. (`monthlyWeekday` is safe — it's `(weekday+1)%7`, always 0–6.)

**Suggested fix (defense in depth):**
- In `rruleToLegacyTaskRepeatCfg`, only set the anchor when the ordinal is model-valid; otherwise leave it unset (the new client routes on `rrule`, old clients fall back to day-of-month):
  ```ts
  const n = weekdays[0].n;
  if (n === -1 || (n >= 1 && n <= 4)) {
    out.monthlyWeekOfMonth = n as MonthlyWeekOfMonth;
    out.monthlyWeekday = ((weekdays[0].weekday + 1) % 7) as MonthlyWeekday;
  }
  ```
- And extend the action-creator boundary to strip/repair any `monthlyWeekOfMonth ∉ {1,2,3,4,-1}` (and `monthlyWeekday ∉ 0..6`) to `undefined`, so no untyped path (import, builder, future REST) can wire an out-of-union anchor. Mirror it in `data-repair.ts`.

### 2. ❗ Sub-daily RRULEs (deferred to Phase 12) are accepted via raw override

`isRRuleValid` accepts any rrule.js FREQ (`store/rrule-occurrence.util.ts:84`) and the raw-override textarea is exposed (`rrule-builder.component.html:470`). Two consequences:
- **Semantic collapse:** the engine maps every occurrence to local noon (`rrule-occurrence.util.ts:52`), so `FREQ=HOURLY`/`MINUTELY`/`SECONDLY` silently degrade to ~day-level firing — accepted but not honored.
- **Forward-compat break:** `FREQ_TO_CYCLE` maps only DAILY/WEEKLY/MONTHLY/YEARLY (`legacy-cfg-to-rrule.util.ts:106-110`), so a sub-daily FREQ yields `cycle = undefined` → `repeatCycle: undefined` (`:153`). `repeatCycle` is a required field, `JSON.stringify` drops the key, and the persisted cfg is then missing a required field → typia failure on new and old clients.

**Suggested fix:** gate supported frequencies at the save/validation boundary — reject (or refuse to persist) a rule whose FREQ isn't DAILY/WEEKLY/MONTHLY/YEARLY until Phase 12, with a clear "not supported yet" message in the builder, rather than silently accept-and-collapse.

### 3. Raw rule body / task content reaches logs in the "update all" flow

`store/task-repeat-cfg.effects.ts:714` logs `Log.log(changes)` and `Log.log(todayTasks, archiveTasks)` in the update-all confirm path. The dialog's `changes` (`dialog-edit-task-repeat-cfg.component.ts:439`) now carries the `rrule` string (and the tasks carry titles/notes), so raw rule bodies + user content can land in the exportable log — conflicting with the "never log user content" rule and the "no rule-body logging" goal. This `Log.log` predates this PR (it already logs titles/notes), but the PR widens what leaks. Suggest logging ids/changed-keys only — e.g. `Log.log({ changedKeys: Object.keys(changes) })` and task ids — never the `changes`/task objects. Same class as the `Log.warn('Invalid RRULE', e)` parse-error path in `rrule-occurrence.util.ts:75` (the `Error` message can embed the rule string); worth fixing together.

### 4. Process: draft + no CI signal

`gh pr checks` reports no checks on the branch, and the PR is still a draft — so none of the unit/e2e coverage (where items 1–2 should be caught) has run. Worth marking ready + getting CI green before merge.

### Suggested regression tests
- `BYDAY=5MO` / `-2MO` never persists an out-of-union `monthlyWeekOfMonth`.
- `FREQ=HOURLY` is rejected (or never yields `repeatCycle: undefined`).
- typia-validate a cfg produced by each builder path.

These join the earlier notes on the biweekly `WKST` interval shift and the migrate-on-open reschedule as the pre-merge fix list. Happy to pair on any of these.

johannesjo:
Thanks for the continued work here. The RRULE builder and the save-time guardrails are a substantial step forward, and the latest curator fixes clearly tightened several risky edges.

I re-checked the current head and found a few issues that should be fixed before this lands:

1. Preset saves currently strip the RRULE they just generated.

   In `dialog-edit-task-repeat-cfg.component.ts:373-386`, non-`RRULE` presets call `getQuickSettingUpdates()`, which now adds a canonical `rrule` (`get-quick-setting-updates.ts:44-47`, `:60-64`). Immediately after that, `dialog-edit-task-repeat-cfg.component.ts:469-471` clears `rrule` again because the quick setting is not `RRULE`.

   That means dialog-created presets are not actually RRULE-backed, despite the new contract. It also means editing an existing RRULE-backed config into a preset can remove the rule. The spec at `dialog-edit-task-repeat-cfg.component.spec.ts:943-952` currently asserts this stale behavior and should be inverted.

2. `undefined` clears are not sync-durable for `rrule`.

   `getTaskRepeatCfgChanges()` preserves `rrule: undefined` in the update diff (`get-task-repeat-cfg-changes.ts:23-25`), and the reducer applies partial updates with `adapter.updateOne()` (`task-repeat-cfg.reducer.ts:66-67`). But once the op payload is JSON-serialized, the `undefined` key is dropped, so a remote client replays an update that never clears the old `rrule`. Since the occurrence engine prefers valid `rrule` over legacy fields, other clients can keep scheduling from the stale rule.

   Suggested fix: do not clear preset-generated RRULEs, and use a JSON-stable clear/replace strategy for cases that genuinely need to remove optional schedule fields.

3. Monthly anchor clears have the same sync issue for legacy fallback.

   `MONTHLY_ANCHOR_RESET` sets `monthlyWeekOfMonth` and `monthlyWeekday` to `undefined` (`get-quick-setting-updates.ts:35-38`). The code comment already notes those keys disappear over JSON. That is not inert for old/mobile clients that ignore `rrule`, because the legacy monthly engine checks Nth-weekday anchors first (`get-next-repeat-occurrence.util.ts:121`). A synced change from an Nth-weekday monthly preset to a day-of-month/month-end preset can therefore leave old clients scheduling from stale anchors.

4. Completion-relative RRULEs can reopen as friendly presets and hide their schedule type.

   `_processQuickSettingForDate()` infers `CUSTOM` + matching `rrule` back to a preset at `dialog-edit-task-repeat-cfg.component.ts:657-665` without checking `repeatFromCompletionDate`. If a completion-relative `FREQ=DAILY;INTERVAL=2` reopens as `EVERY_OTHER_DAY`, `isRRuleMode()` becomes false, and the only schedule-type toggle is hidden inside the RRULE builder (`dialog-edit-task-repeat-cfg.component.html:35-42`, `rrule-builder.component.html:374-397`). Either skip preset inference for `repeatFromCompletionDate` configs or expose that control outside builder-only mode.

Smaller cleanup:
- `src/assets/i18n/en.json:2567` advertises `@+<phrase>` short syntax, but `e2e/tests/recurring/rrule-repeat.spec.ts:50-52` says that support is deferred.
- `rrule-builder.component.html:98`, `:163`, `:211`, `:445`, `:460`, `:476` have visible hardcoded placeholders instead of translation keys.

The main test coverage to add is a JSON round-trip sync test for clearing optional repeat fields, plus a dialog save test for normal presets that expects the saved config to keep its canonical `rrule`.

johannesjo:
I re-checked the latest head (`e1628364`) and the `b5561df` fix with a KISS/YAGNI lens. I think there are still two pre-merge issues, plus one small test cleanup.

1. **Monthly anchor clears are still not sync-durable for legacy clients.**

   `src/app/features/task-repeat-cfg/store/task-repeat-cfg.actions.spec.ts:215-230` now pins this as a documented gap: it starts from an nth-weekday monthly cfg (`monthlyWeekOfMonth: 2`, `monthlyWeekday: 2`, `rrule: FREQ=MONTHLY;BYDAY=2TU`), updates to `FREQ=MONTHLY;BYMONTHDAY=15`, and expects the stale anchors to remain after the JSON op-log round-trip. `src/app/features/task-repeat-cfg/task-repeat-cfg.model.ts:131-136` also acknowledges that legacy clients ignore `rrule`, so they keep scheduling the 2nd Tuesday while rrule-aware clients schedule the 15th.

   KISS/YAGNI read: I would not require a broad nullable-field migration if we can avoid it, but a normal preset/mode switch that is expressible in the legacy model should not intentionally diverge across synced clients. The smallest acceptable fix is probably to make this clear durable for the affected schedule-mode switch path, or to avoid presenting/persisting it as legacy-safe until there is a wire-stable clear strategy.

2. **Switching a completion-relative RRULE cfg to a preset can persist hidden schedule state.**

   `_processQuickSettingForDate()` correctly reopens `repeatFromCompletionDate` cfgs in builder mode (`dialog-edit-task-repeat-cfg.component.ts:644-650`) because the schedule-type toggle only exists in the builder. But the save path for non-`RRULE` presets (`dialog-edit-task-repeat-cfg.component.ts:373-385`) applies `getQuickSettingUpdates()` over the existing cfg and does not clear `repeatFromCompletionDate`. As soon as the user selects a preset, `dialog-edit-task-repeat-cfg.component.html:35-42` hides the builder, so the saved preset can still be "after completion" with no visible control showing or changing that behavior.

   KISS fix: either clear `repeatFromCompletionDate` when selecting/saving a non-`RRULE` preset, or keep the schedule-type control visible whenever that flag can persist. I would avoid a broader UI reshuffle unless presets are intentionally supposed to support completion-relative scheduling.

3. **Small test cleanup:** `src/app/features/task-repeat-cfg/store/task-repeat-cfg.actions.spec.ts:210-212` is currently vacuous. `baseEntity.monthlyLastDay` is already `false`, so the test passes even if the update is dropped or ignored. Setting the base value to `true` first would make the test prove that `false` actually survives the wire.

johannesjo:
Thank you very much for your hard work on this!  

There is some minor followup, but we can tackle this in the next step.

johannesjo:
### Review: `fix(task-repeat): keep presets rrule-backed + builder mode for completion cfgs`

Reviewed this commit's diff (`45f0000..b5561df`) in isolation. Net assessment: **solid, ship-able** — the change is correct and actually fixes a real bug. One incomplete-fix worth a follow-up, plus minor polish.

**👍 Nice catch on the rrule-cleanup removal.** Removing the `else if (working.rrule) { rrule: undefined }` branch isn't just cleanup — that branch ran *after* `getQuickSettingUpdates` had set the preset's canonical rrule, so every preset was being persisted with `rrule: undefined`, silently violating the "presets stay rrule-backed" contract. The inverted test (`expect(savedCfg.rrule).toBe('FREQ=DAILY')`) pins the correction nicely. Verified safe for all reachable states (`CUSTOM` isn't user-selectable and is migrated to `RRULE` on open).

---

**🟡 Incomplete fix — completion cfgs *without* an rrule still hide the "from completion" toggle**
`dialog-edit-task-repeat-cfg.component.ts` (the `repeatFromCompletionDate` check inside `_processQuickSettingForDate`)

The force-to-builder check is gated inside `if (cfg.rrule)`. A stored completion-relative cfg with no rrule — i.e. any pre-RRULE cfg, since migration is lazy/save-only — falls through to `needsMigration`, which is `false` for `DAILY`/`MONDAY_TO_FRIDAY` and for any preset carrying a `startDate`. So it opens under its preset label with the toggle hidden, which is the exact case this commit is trying to fix. Both new specs only cover the rrule-bearing case, so this slips through.

Not a regression and not data-loss (`repeatFromCompletionDate` is preserved on save and still honored by the engine) — purely a UI-completeness gap. Suggested fix: hoist the `repeatFromCompletionDate` check above the `if (cfg.rrule)` guard, and in the no-rrule branch migrate to rrule + force builder (mirrors the existing `needsMigration` path). Or document it as a known limitation if it's out of scope for Phase 1.

```ts
if (cfg.repeatFromCompletionDate) {
  if (cfg.rrule) {
    return cfg.quickSetting === 'RRULE' ? cfg : { ...cfg, quickSetting: 'RRULE' };
  }
  return { ...cfg, rrule: legacyTaskRepeatCfgToRRule(cfg as TaskRepeatCfg), quickSetting: 'RRULE' };
}
```

---

**🟡 The documented KNOWN GAP is real and broader than "inert" — fine to defer, but track it**
`task-repeat-cfg.model.ts` / `MONTHLY_ANCHOR_RESET`

On an nth-weekday → day-of-month edit, the `undefined` anchor-clear is dropped by the JSON op-log wire, so a remote **released** client (routes on `repeatCycle`/anchors, ignores `rrule`) keeps scheduling the old "Nth weekday" rule. Since the nth-weekday feature already shipped but the rrule engine is unreleased, this affects the whole current installed base during the mixed-version window. It self-heals once both devices run the rrule build and loses no data, so deferring is reasonable — but worth a tracked follow-up rather than only a code comment.

One note: a `monthlyWeekOfMonth: 0` sentinel won't work as an in-release shortcut — `MonthlyWeekOfMonth = 1 | 2 | 3 | 4 | -1` is a literal union, so `0` would trip released clients' typia data-repair the same way `null` does. The documented two-release `| null` migration is the right path.

---

**Minor / optional**
- **Test weight:** the "rrule string survives the wire" and "`monthlyLastDay: false` survives" round-trip cases are near-tautological (they mainly assert `_toPersisted` doesn't strip those fields). The `DOCUMENTED GAP` test is the valuable one — consider collapsing the two positives and keeping that one standalone. `as never` on the reducer state is harmless here but a typed cast would document intent.
- **Comment DRY:** the `(1) ship | null, (2) switch reset to null` migration recipe is spelled out in both `model.ts` and the spec — keep one canonical copy and cross-reference. (The `save()` NOTE explaining the *absence* of cleanup is worth keeping verbatim.)
- **i18n placeholders:** the syntax-example placeholders (`"1,-1"`, `FREQ=…`) are display-only (never parsed), so no functional risk — but a translator note would stop a locale from mangling the example.
- **Perf:** the 5 new `| translate` bindings are negligible (impure pipe fast-paths constant keys; OnPush dialog with ~40 such bindings already).

**Verdict:** Approve. No blocking issues; the only substantive item is the rrule-less completion-cfg gap above.

johannesjo:
## Whole-PR risk review (Phase 1)

Reviewed the full PR (`3ff0eeb`…`f6bf99f`, 70 files / +8523-699) for **merge risk**, fanning out across six surfaces — sync/op-log forward-compat, occurrence engine, legacy⇄RRULE migration, dialog/form, integration/bundle, and test coverage. Verified the two most serious findings against the code directly.

**Verdict: merge-with-followups.** No CRITICAL data-corruption or old-client-breaking defect — the highest-risk surface (sync forward-compat) is well-engineered and checks out. One user-impacting correctness bug is worth fixing before/with merge, plus a perf edge and some test gaps. Confidence: medium-high (structural claims verified at the PR head; tests not executed — CI is the gate).

---

### 🟠 Worth fixing before merge — silent `repeatFromCompletionDate` flip
`dialog-edit-task-repeat-cfg.component.ts:393-394` + `_processQuickSettingForDate` (~`:660`)

A legacy/imported cfg with `repeatFromCompletionDate: true`, **no `rrule`**, and a kept-preset `quickSetting` (e.g. `DAILY`) opens under the **preset label** — the "from completion" toggle (which lives only in the RRULE builder) is hidden, because the completion→builder force is gated inside `if (cfg.rrule)` and `needsMigration` is `false` for `DAILY`/`MONDAY_TO_FRIDAY` (and any preset carrying a `startDate`). Then on **any Save**, the preset branch runs:

```ts
...(cfg.repeatFromCompletionDate ? { repeatFromCompletionDate: false } : {}),
```

…which **silently switches the task from "repeat after completion" to "repeat from start date"** — no visible control, no warning. Migration also adds an `rrule`, so the diff is non-empty and the change is dispatched/synced.

This is the no-rrule-completion gap I flagged on the earlier sub-commit, but **escalated**: at `b5561df` the flag was *preserved* on save; a later commit added this reset, turning a hidden-toggle UX gap into a silent schedule-semantics change (the same class of bug the PR guards against elsewhere, cf. #7373). It affects existing users who use "repeat after completion" (all such legacy cfgs are rrule-less). Currently untested.

**Fix:** hoist the `repeatFromCompletionDate` check above `if (cfg.rrule)` in `_processQuickSettingForDate` and migrate no-rrule completion cfgs to `RRULE` mode — they then open in the builder with the toggle visible, and `quickSetting === 'RRULE'` skips the reset.

### 🟠 Engine freeze on a never-firing sub-yearly rule
`rrule-occurrence.util.ts:90-110` (`isRRuleValid`), used as the routing guard in `get-newest-possible-due-date.util.ts:26`

`isRRuleValid` probes `.after()` with **no horizon cap**. A parseable-but-never-firing sub-yearly rule (e.g. `FREQ=DAILY;BYMONTH=13`) makes the rrule iterator walk day-by-day to ~year 275760 before returning `null` *without throwing* — a ~3.8s main-thread freeze — and `isRRuleValid` then returns **`true`**, so the engine defers to the rule (bypassing the legacy fallback) and the cfg silently never fires. Memoized, so it's a one-time freeze per rule, not per scan.

Reachability is low: the builder can't emit such a rule; the **raw-override** input can, but save rejects it via `RRULE_NO_OCCURRENCE` (after the same freeze); the real exposure is a hostile/buggy **synced/imported** rule hitting the startup due-scan. Not data corruption, so I'd treat it as a warning rather than a blocker — but worth a horizon cap (treat "no occurrence within ~10y" as invalid → legacy fallback), which fixes both the freeze and the silent-never-fire.

### 🟡 Warnings (non-blocking)
- **Sub-daily (HOURLY/MINUTELY) and COUNT+completion are rejected only at the dialog** (`dialog…:410-432`), not at the engine or persist boundary. A rule arriving via sync/import/future-REST is accepted by the engine and silently fires ~daily / never terminates. Bounded (engine is fail-soft on throws), but the persist boundary / `data-repair` should mirror these guards as defense-in-depth.
- **`legacy-cfg-to-rrule.util.ts` `_parseStart` wall-clock fallback:** migrating a *startDate-less* MONTHLY/YEARLY cfg encodes **today's** day-of-month into the rrule (vs the engine's day-1 semantics) — a silent schedule change once persisted. Low reachability (UI paths set `startDate`). Suggest defaulting to `1970-01-01` to match the engine.
- **Test gaps that could let a regression pass green CI:** the effects re-anchor (`rrule` ∈ `SCHEDULE_AFFECTING_FIELDS`, which prevents recurring tasks silently stopping after an edit) is **untested**; the no-rrule completion case above is untested; e2e is **single-client only** — no cross-device sync/replay test exercises the forward-compat clamp end-to-end.

### 🟢 Verified clean (the hard-to-recover-from parts)
- **Sync/op-log forward-compat — safe.** Every wire-bound creator (`add`/`update`/`updateMany`) routes through `_toPersisted`, clamping `quickSetting`→master-safe and sanitizing monthly anchors; `upsertTaskRepeatCfg` carries no persistent meta and has no callers, so it can't leak. The only new synced field is `rrule?: string` (optional opaque string old clients ignore — typia uses non-exact validation). `MASTER_SAFE_QUICK_SETTINGS` exactly matches the released master union; `data-repair` is deterministic and never strips `rrule`. No path can put an out-of-union value on the wire to trip a released client's data-repair. Effects inject `LOCAL_ACTIONS`.
- **Engine is DST-safe and replay-neutral:** local-noon + UTC-internal is day-stable across UTC−11…+14 and both DST directions (CI runs Berlin + LA over all specs); fail-soft (every rrule call try/catch → null → legacy fallback; logs only error names, no rule bodies); no `Date.now()` in the engine.
- **Migration is faithful and lazily inert:** `repeatCfgInitial` is the *processed* cfg, so opening + no-op-saving a migrated cfg yields an empty diff for the rrule/quickSetting fields; the biweekly-WKST divergence is neutralized by save-time start-date alignment.
- **Integration:** `rrule` lands in the main bundle (the engine runs at startup) but only the core import (~30KB gzip, no luxon/tz subpath) — worth a budget check, not a blocker. The REST `+1` line is whitespace. The dev `Date.now()` override (`main.ts`) is airtight-stripped from prod/stage. Lockfile adds only `rrule` (+ pre-existing `tslib`).

### Surface verdicts
| Surface | Verdict | Conf. |
|---|---|---|
| Sync / op-log forward-compat | merge-safe | 88% |
| Occurrence engine | merge-with-followups (freeze) | 85% |
| Legacy⇄RRULE migration | merge-with-followups | 85% |
| Dialog / form | merge-with-followups (flip) | 88% |
| Integration / bundle / platform | merge-with-followups | 90% |
| Test coverage | merge-with-followups | 80% |

### Bottom line
There's **no silent-data-corruption or cross-version-break risk in the sync layer** — the part that would be hardest to recover from. What remains is one correctness bug worth fixing first (the from-completion flip, since it silently changes existing users' schedules), a perf edge (the freeze), and a few test gaps. Suggest gating merge on the from-completion flip + green CI; the horizon cap, persist-boundary guards, and missing tests are fine as fast-follows.

<sub>Reviewed with a multi-agent pass; the two 🟠 findings were verified directly against the PR head.</sub>

johannesjo:
I am merging the first part, but lets keep in mind the correctness findings for the next stage.

Thank you very much! <3

johannesjo:
First off, thank you for this, genuinely. Phase 1 is high-quality work: the UTC-to-local-noon DST-safe engine, the fail-soft parsing, and especially the forward-compat design (keeping `quickSetting` clamped to a shipped value at the op-log boundary so older/mobile clients never choke) are exactly the kind of care this codebase needs. The test coverage is excellent too.

After living with it merged for a bit, I've decided to **revert Phase 1 from master and instead develop the whole RRULE epic on a long-running integration branch** (`feat/rrule-epic`), landing it in master only once it's complete and testable in its final form. To be clear up front: **this is on me, not on your work. I underestimated the complexity when we scoped it**, and that only became clear once I had a real PR to look at. It hasn't shipped in any release yet (we're post-v18.9.1), so there's zero user impact from backing it out now.

The reasoning is about _shape_, not quality:

- **A 13-phase epic shouldn't land in master one phase at a time.** Each merged slice leaves the app in a half-state, and because RRULE coexists with the legacy `repeatCycle` engine via live bidirectional migration, it adds permanent dual-engine surface I'd have to carry for an indefinite stretch while the rest of the epic is still being built. I'd rather integrate the whole thing on one branch and merge it as a finished, fully-testable feature.
- **It's reversible and nothing is lost.** Phase 1 becomes the _foundation_ of the long-running branch, not a discard. We keep building there until the epic is whole.
- **Demand is real but concentrated on the common cases** (end conditions, multiple dates, overdue handling), so there's no urgency that justifies shipping it half-built.

So concretely: I'll create `feat/rrule-epic` from the current Phase-1 state, and that becomes the home for all further phases. I'd love for you to keep driving it there: same work, same direction, just integrated as a whole before it hits master. The waypoint branch stays our reference for the later slices.

None of this is a knock on the contribution, which I really appreciate. I'm recalibrating _how_ it lands, not _whether_. Open to your thoughts, and happy to talk through how we sequence the remaining phases on the branch.

omega-tree:
sure, I will handle your follow-ups with core before with phase 2 on that new branch. I agree this is a large engine change and should be tested thoroughly before merge. a new branch is fine. I am curious what your plan would be to get it in the hands of users for testing before moving to master. 

In my opinion it should be on a new major release . I would probably suggest we gate the RRULE engine behind a feature flag, off by default. That lets the epic merge to master without the half-state problem — legacy repeatCycle stays authoritative, no migration fires, older/mobile clients are untouched — while opt-in testers and we exercise it on real data across devices. Then it rides a major release as explicitly opt-in/beta, we watch for the transient bugs you're rightly expecting, and flip the default on in a follow-up once cross-version sync has real mileage.

I'm Happy to continue on the new branch if that direction works for you.

johannesjo:
Yes, gating this behind a feature flag sounds like a good idea, if it doesn't introduce more complexity in its own right.

Thank you very much for working on this! <3

Issue #7949, opened 2026-06-02
🚨 Google Calendar - Wrong time Update after done


hajiboy95:
### Version Used

188.8.0 Desktop Mac

### Current Behavior

When marking a task as **Done**, the corresponding Google Calendar event duration is updated incorrectly based on an erratic calculation between the estimated time and the actual time spent.

Instead of reflecting the actual time spent (or preserving the original slot), the calendar event duration undergoes unexpected math or falls back to a default value.

### Actual Behavior

The integration applies faulty duration logic depending on which value is higher:

* **Scenario A (Estimate > Actual):** The timeslot is updated to the *difference* between the estimate and the actual time (**Estimate - Actual**).
* *Example:* Estimated 1 hour (60 mins), Actual time 51 mins -> Timeslot shrinks to **8–9 minutes**.

* **Scenario B (Estimate < Actual):** The timeslot completely overrides the data and defaults to a flat **30 minutes**.

### Expected Behavior

Upon completion, the Google Calendar timeslot should either:

* Update to match the **Actual Time** spent on the task.
* **OR** Remain as the originally scheduled block if there's no time tracking.

### Steps to Reproduce

1. Create a task with a time estimate and sync it to Google Calendar.
2. Track time on the task (Actual Time).
3. Mark the task as **Done**.
4. Observe the updated event duration in Google Calendar.

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

hajiboy95:
Fixed via #7950

Issue #373, opened 2020-05-18
Improve cask deployment for Mac


johannesjo:
This should now be possible via github actions. PRs are welcome!

~~Currently brew cask deployment needs to be triggered manually (by someone, doesn't seem to matter who). It would be really cool to have an automated solution for this, though it's a bit tricky as the script needs a Mac to run...~~

johannesjo:
We can now use github actions for this.

cwerner1:
This would be nice:)
Today I've manually updated the cask with 
`brew bump-cask-pr  superproductivity --version 7.7.0 `
https://github.com/Homebrew/homebrew-cask/pull/113900

hd-deman:
brew cask still outdate 7.9.2

johannesjo:
I have no mac available currently, but if someone executes `brew bump-cask-pr superproductivity --version 7.10.1` the app should be updated.

cwerner1:
updated to 7.10.1 
https://github.com/Homebrew/homebrew-cask/pull/118758

johannesjo:
Thank you  @cwerner1 !

cwerner1:
updated to 7.11.5
https://github.com/Homebrew/homebrew-cask/pull/122784

PR #7959, opened 2026-06-02
chore(deps): resolve 30 Dependabot security alerts (dev/build tooling)

johannesjo:
Resolves 30 of 32 open Dependabot alerts. All are dev/build tooling — none ship to end users. Also bundles one small test fix that was needed to un-red CI.

## `chore(deps)` — plugin-dev sample plugins (38b2d90) · 26 alerts
Bumped **vite** (≥7.3.2), **vitest** (4.1.x), **postcss** (8.5.15), **glob** (10.5.0), **minimatch** (9.0.9 / 5.1.9) across `packages/plugin-dev/*` via lockfile updates. Only manifest change: `clickup-issue-provider` vitest `^3.2.1`→`^4.1.0`.
Verified: clickup (28 tests) + automations (109 tests) suites pass.

## `chore(deps)` — root lockfile (492fb44) · 4 alerts
- **minimatch #341 / #372 / #373**: the existing `app-builder-lib` override pinned minimatch to the vulnerable **10.1.1**, which npm cascades across electron-builder's whole subtree (`@electron/asar`, `@electron/universal`, `dir-compare`, `filelist`, `temp`). Bumped the override to **10.2.5** (within `app-builder-lib`'s `^10.0.3`); the nested copies dedupe to the patched top-level.
- **webpack-dev-server #593**: override to **5.2.4**.

## `test(calendar)` — bundled fix (3ecd26b)
The non-done time-block test expected `max(estimate, timeSpent)` (60min) but the effect — per #7949 — intentionally uses the remaining-estimate model for active tasks: `max(timeEstimate - timeSpent, 0)` (40min). The implementation matches its inline doc and the #7949 commit message; **the test expectation was wrong and failed deterministically on `master`**, blocking the pre-push hook / CI. Corrected the expectation and renamed the test. No production code change.

## Not fixed (2) — recommend dismissing in the Dependabot UI
- **uuid #601**: not exploitable — the bug needs `uuidv3/v5/v6(…, buf)`; the only consumer (sockjs) uses v4 random and pins `^8`, so a fix would force uuid 11 and break it.
- **jwt #595**: fastlane pins jwt `<3` and the only patch is 3.2.0; local release tooling with no untrusted-token path.

## Verification note
electron-builder `npm run dist` packaging was not run; the minimatch change is a patch within range, so impact is expected nil.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3480,22 +3480,6 @@
         "url": "https://github.com/sponsors/isaacs"
       }
     },
-    "node_modules/@electron/asar/node_modules/minimatch": {
-      "version": "10.1.1",
-      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-10.1.1.tgz",
-      "integrity": "sha512-enIvLvRAFZYXJzkCYG5RKmPfrFArdLv+R+lbQ53BmIMLIry74bjKzX6iHAm8WYamJkhSSEabrWN5D97XnKObjQ==",
-      "dev": true,
-      "license": "BlueOak-1.0.0",
-      "dependencies": {
-        "@isaacs/brace-expansion": "^5.0.0"
-      },
-      "engines": {
-        "node": "20 || >=22"
-      },
-      "funding": {
-        "url": "https://github.com/sponsors/isaacs"
-      }
-    },
     "node_modules/@electron/fuses": {
       "version": "1.8.0",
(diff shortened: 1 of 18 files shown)


johannesjo:
Superseded by https://github.com/super-productivity/super-productivity/pull/7960 — rebuilt cleanly on current master. This branch was cut from a stale local base (8 commits behind master), so it dragged in already-merged commits and conflicted. The replacement is deps-only on top of current master.

PR #7965, opened 2026-06-02
Allow writable plugin calendar events to be rescheduled from Schedule

noamsiegel:
## Summary

- Allow writable plugin calendar events to be dragged/rescheduled from Schedule.
- Keep iCal, reference calendars, and plugin providers without `updateIssue` read-only.
- Route calendar moves through the existing issue sync adapter path.

## Review Updates

- Added Undo for successful calendar moves/reschedules.
- Hold the drag position during provider writes, then reset the CDK transform after the write resolves.
- Treat Shift+drop on calendar events as a normal timed move.
- Shared provider write handling between drag moves and dialog reschedule.
- Sanitized provider write/delete error logs and toast text so Google calendar IDs are not logged or shown from Angular HTTP messages.
- Removed unreachable all-day drag handling while preserving date-only undo for dialog reschedules.
- Switched local date formatting to `getDbDateStr`, removed dead calendar-event type code, and added a dedicated "Event moved" snack.
- Shared lazy plugin sync adapter resolution between issue sync effects and calendar actions so cold plugin registries do not silently no-op writes.
- Aligned calendar write gates around `updateIssue`: update-only providers resolve lazy adapters, and reschedule UI/actions require the same capability as drag moves.
- Fixed all-day dialog reschedules to pass the dialog's `Date` result directly to `getDbDateStr`, avoiding invalid `NaN-NaN-NaN` date-only writes.
- Kept calendar drop interactions disabled while provider writes are pending, and suppressed task-only unschedule previews for invalid calendar drops.
- Added missing schedule component service mocks used by the full Angular test suite.
- Added focused tests for move/reschedule, no adapter, lazy adapter resolution, error path, Undo, Shift+drop, pending drag reset, and invalid calendar drops.

## Verification

- `npx ng test --watch=false --include=src/app/features/calendar-integration/calendar-event-actions.service.spec.ts --include=src/app/features/schedule/schedule-week/schedule-week-drag.service.spec.ts`
- `npx ng test --watch=false --include=src/app/features/issue/two-way-sync/issue-sync-adapter-resolver.service.spec.ts --include=src/app/features/issue/two-way-sync/issue-two-way-sync.effects.spec.ts --include=src/app/features/calendar-integration/calendar-event-actions.service.spec.ts --include=src/app/features/schedule/schedule-week/schedule-week-drag.service.spec.ts --include=src/app/features/schedule/schedule-week/schedule-week.component.spec.ts`
- `npx ng test --watch=false --include=src/app/features/issue/two-way-sync/issue-sync-adapter-resolver.service.spec.ts --include=src/app/features/calendar-integration/calendar-event-actions.service.spec.ts --include=src/app/features/schedule/schedule-event/schedule-event.component.spec.ts --include=src/app/features/schedule/schedule/schedule.component.spec.ts`
- `npx ng test --watch=false --include=src/app/features/calendar-integration/calendar-event-actions.service.spec.ts --include=src/app/features/issue/two-way-sync/issue-sync-adapter-resolver.service.spec.ts`
- `npm run lint`
- `npm test` in `packages/plugin-dev/google-calendar-provider`


johannesjo:
**Nice work — this is a genuinely useful addition.** Drag-to-reschedule is the gesture people expect on a calendar grid, and routing it through the existing `updateIssue`/sync-adapter path (rather than a new fan-out) is the right call. The capability gate (`!isReferenceCalendar` + plugin + `updateIssue`) is correctly conservative, and the read-only fallback is preserved. I ran a multi-pass review; findings below, roughly by priority.

The core thing to keep in mind: **a task drag mutates local, undoable state — a calendar drag is an irreversible network write that propagates to every device and attendee.** Same gesture, much higher stakes. A few gaps follow from that:

### Worth addressing before merge

1. **Snap-back-then-jump (no optimistic update).** On drop, `ev.source.reset()` snaps the event back; it only moves after the PATCH + `triggerRefresh()` round-trips. So the user sees the event jump *back*, then forward seconds later — and on failure it just stays put with an error snack, indistinguishable from "nothing happened." Consider holding the dropped position optimistically and reverting only on error.
2. **No undo on an irreversible write.** A stray drag silently reschedules a real event with no recovery. `SnackService` supports `actionStr`/`actionFn` — capturing the original start/duration and offering "Undo" on the success snack is the single highest-value safety addition here.
3. **All-day events are silently coerced to 1-hour timed events.** `moveToStartTime` always sends `start_dateTime`, so dropping an all-day event onto a slot rewrites it as a 10:00–11:00 event on the real calendar. The provider already supports the non-destructive `start_date` path (used by the context-menu `reschedule()`). Minimum: refuse the timed drop for all-day events; better: route to `start_date`.
4. **DRY** — `moveToStartTime` duplicates ~25 lines of `reschedule()`'s timed branch (duration calc + adapter lookup + `pushChanges` + `triggerRefresh` + success/error snacks). Extracting one `_pushChanges(calEv, changes)` helper that both call would also give the undo logic above a single home.
5. **Tests** — the new specs mock `CalendarEventActionsService`, so they only assert wiring. `moveToStartTime` itself (the all-day branch, `!adapter` early-return, error path) is untested. A focused spec on the extracted helper would cover both it and the currently-untested `reschedule()`.

### Smaller / cleanup

6. **Shift-drag a calendar event** shows a "plan-for-day" preview but does nothing on release — suppress the shift visuals for calendar events, or treat shift as a normal move.
7. **Type model**: `SVECalendarEvent.data` is typed `ScheduleCustomEvent` (which `Omit`s `calProviderId`), but the runtime object has it and the new code depends on it — hence the three `as ScheduleFromCalendarEvent` casts. Typing `data` as `ScheduleFromCalendarEvent` removes all three.
8. **Two draggability gates** (`canDragEvent` in the component vs `_pluckCalendarEventFromEvent` in the service) encode "is this a movable calendar event?" twice with slightly different conditions — they can drift. Consolidate to one predicate.
9. The calendar branch in `handleDragReleased` re-implements the same teardown the `if (!task)` guard below already does — falling through deletes ~8 lines.
10. `const calendarEvent = ev as ScheduleEvent` in `canDragEvent` is a no-op cast.

### Heads-up (pre-existing, not introduced by you)

On a *failed* PATCH, `Log.err('...', e)` stores an Angular `HttpErrorResponse`, which isn't `instanceof Error`, so the log exporter `JSON.stringify`s its `.url` — and for Google that URL contains the calendar id (often the user's email). That's user data in exportable logs (repo logging rule). `reschedule()`/`deleteEvent()` already do this, so it's a project-wide cleanup rather than yours to fix — but if you add the `_pushChanges` helper, sanitizing the log there (`{ status: e?.status }`) fixes all three at once.

### Things I checked that are fine

Recurring events are listed with `singleEvents: true`, so a drag edits a single instance, not the series 👍; the move correctly targets `data.id` (not the day-suffixed view id); and the per-render `canMoveEvent` cost is negligible.

noamsiegel:
@johannesjo Thanks for the detailed review. I pushed updates for the main items: Undo, delayed reset on failed writes, date-only all-day moves, shared provider write handling, sanitized error logging, shift-drag cleanup, typing cleanup, and focused tests for the calendar action paths.

johannesjo:
Re-reviewed the updated PR (ran it past several focused reviewers and verified the findings against the code). The main items from the first pass are in good shape: optimistic positioning now resets only on failure, Undo captures the pre-move state correctly, the shared `_pushChanges` helper is clean, iCal and reference calendars stay read-only, and both new specs pass locally (5/5 and 11/11). No NgRx effects/reducers/actions were added, so the op-log and sync rules do not apply here, and performance is fine (`getProvider` is an O(1) map lookup, the drag-move path is throttled).

One privacy regression to fix before merge, plus a few smaller follow-ups.

## Must fix

**1. The log sanitizer still leaks the calendar id / email, via `message` (`calendar-event-actions.service.ts`, `_sanitizeLogError`).**

The whitelist returns `name`, `message`, `status`, `statusText`. The error reaching every catch (move, reschedule, delete, undo all funnel through `_pushChanges` / `deleteEvent`) is an Angular `HttpErrorResponse`, because the plugin HTTP path runs through `PluginHttpService`, which uses `HttpClient` directly with no sandbox serialization. Angular builds its `message` as:

`Http failure response for <url>: <status> <statusText>`

so a 403 on a calendar PATCH is literally:

`Http failure response for https://www.googleapis.com/calendar/v3/calendars/<calendar-id>/events/...: 403 Forbidden`

The calendar id in that URL is frequently the user's email. `Log.recordLog` stores `args.slice(1)` (the sanitized object, including `message`) into the exportable `logHistory`, so the email lands in exportable logs. That is exactly the case the sanitizer was meant to prevent.

The new spec passes only because its mock error is `{ status, statusText, url }` with no `message`, so `not.toContain('noam@example.com')` is vacuously true. A real `HttpErrorResponse` always has `message`.

Fix: drop `message` from the whitelist (keep `name`, `status`, `statusText`), and update the spec mock to include a realistic `message` containing the URL so the assertion actually exercises the path.

Lower-priority, same root cause: the error snack uses `getErrorTxt(e)`, which returns `e.message` first, so the same URL shows in the toast. That is transient and it is the user's own data, but a friendlier calendar-error message would keep it out of screenshots and bug reports.

**2. Shift + drop on a calendar event is a silent no-op while the preview promises a move (`schedule-week-drag.service.ts`).**

`handleDragMoved` and `refreshPreviewForCurrentPointer` route calendar events to the time preview even under shift (the `&& !draggedCalendarEvent` carve-out), so the user sees a valid "move to HH:MM" preview the whole time. But `handleDragReleased` gates the actual move on `if (!this.isShiftMode() && columnTarget)`, so releasing with shift held does nothing (`isCalendarMovePending` stays false and the `!task` guard just resets the node). Either drop the `!this.isShiftMode()` condition in the calendar block (shift has no meaning for a calendar event and the preview already commits to a timed move), or suppress the preview when shift is held over a calendar event.

## Worth fixing

**3. Dead all-day branch in `moveToStartTime`, with a test that asserts it (`calendar-event-actions.service.ts`).**

`moveToStartTime` is only called from drag release, for events that pass `isScheduleCalendarEvent`. Grid calendar view entries come only from `createBlockerBlocksForCalendarEvents`, which skips all-day events (`if (calEv.isAllDay) { return; }`). So `calEv.isAllDay` is always false inside `moveToStartTime` and the `start_date` arm is unreachable. The "moves all-day events with a date-only change" spec passes only because it calls the method directly, bypassing the drag pipeline. Consider trimming `moveToStartTime` to the timed branch and dropping that test. (`_getCurrentScheduleChanges` and `_formatLocalDate` should stay, since the reschedule-dialog undo path can legitimately hit an all-day event.)

**4. Stale drag transform when an event lands on its current slot (`schedule-week-drag.service.ts`).**

On success the code intentionally skips `ev.source.reset()` to avoid snap-back-then-jump, relying on the post-`triggerRefresh()` re-render to replace the node. That replacement only happens when `trackEventKey` (`id_plannedForDay_startHours`) changes. If the drop rounds to the same start time and day (or the Google refetch lags because of eventual consistency), the key is unchanged, Angular reuses the existing node, and the leftover CDK transform stays applied, so the event sits visually offset until the next unrelated render. The task-drop path avoids this by always resetting. One option: always reset and let the refresh reposition, accepting a brief snap only in the same-slot case.

**5. `_formatLocalDate` (and the inline pad logic in `reschedule`) duplicate the existing `getDbDateStr` util (`calendar-event-actions.service.ts`, `src/app/util/get-db-date-str.ts`).**

`getDbDateStr(date: Date | number | string)` returns the same local `YYYY-MM-DD` and is already used in this directory (`calendar-integration.service.ts`). Routing the three hand-rolled spots through it removes the private helper and the duplicated `pad`.

**6. `ScheduleCustomEvent` is now dead code (`schedule.model.ts`).**

After retyping `SVECalendarEvent.data` to `ScheduleFromCalendarEvent` (a correct improvement, since the rendered event really does carry `calProviderId`), `ScheduleCustomEvent` has no remaining references. Safe to delete.

## Minor / optional

- The calendar-drop orchestration is inlined into the otherwise task and NgRx oriented drag service, with the success/failure reset duplicated across two sites coordinated by `isCalendarMovePending`. Extracting a `_handleCalendarEventDrop(...)` that mirrors `_handleColumnDrop` and owns its own reset would flatten `handleDragReleased` and remove the flag.
- The snack reuses `EVENT_RESCHEDULED` ("Event rescheduled") for a drag-move. Reuse is reasonable; a dedicated "moved" string would read slightly better for the drag gesture.
- Heads-up (not introduced here, and arguably by design): `canMoveEvent` blocks reference calendars but not a regular Google calendar the user added only to view yet happens to hold writer access to. Such a calendar is writable by drag, with Google's 403 as the only backstop. Worth a conscious decision rather than relying on it.

## Verified good

Undo restores the correct pre-move state for timed events and offers no nested undo, so there is no loop. iCal and reference calendars correctly stay read-only. No injection or SSRF surface is added (`PluginHttpService` validates the URL and blocks private and metadata hosts; the new code only passes timestamps and date strings). The `isScheduleCalendarEvent` guard and the `data` type change are sound.

noamsiegel:
@johannesjo 

Addressed the second-pass items in `58729e11`: dropped `message` from the log sanitizer and tightened the spec with a realistic `HttpErrorResponse.message`, made Shift+calendar drop perform the timed move, removed the unreachable all-day drag branch/test, reset the drag transform after provider writes resolve, switched date-only formatting to `getDbDateStr`, removed `ScheduleCustomEvent`, extracted calendar drop handling, and added the dedicated `Event moved` snack. I also kept regular writable Google calendars capability-based intentionally: if a provider account has write access and exposes `updateIssue`, drag uses it; reference/iCal/no-update providers stay read-only.

Verified with focused Angular specs, `npm run lint`, and the Google Calendar provider tests.

johannesjo:
Re-checked the latest head (`58729e11`) with focused passes. Most of the previous review items look addressed, but I still see one functional blocker and a test setup issue, plus two smaller drag-state concerns.

## Must fix

1. **Calendar writes can silently no-op because plugin adapters are only lazily registered.**

`canMoveEvent()` enables dragging from `definition.updateIssue` alone (`calendar-event-actions.service.ts:49-54`), but `_pushChanges()` then uses `this._syncAdapterRegistry.get(...)` directly and returns `false` when the adapter is missing (`calendar-event-actions.service.ts:214-218`). The only place I found plugin adapters being created is `IssueTwoWaySyncEffects._getAdapter()` (`issue-two-way-sync.effects.ts:445-470`), and that path registers the adapter lazily after an issue-sync flow asks for it; the constructor only eagerly registers CALDAV (`issue-two-way-sync.effects.ts:120-122`).

So a writable plugin calendar can appear draggable while the registry is still cold, and the drop just snaps back without a provider write or an error. This is especially likely for users who only use the calendar view and have not already triggered two-way-sync task actions for that provider. I think the calendar action service needs to use the same lazy adapter creation path, or that resolver should be factored out of the effect into a shared service.

2. **`ScheduleWeekComponent` spec was not updated for the new injected service.**

`ScheduleWeekComponent` now injects `CalendarEventActionsService` directly (`schedule-week.component.ts:84`), but `schedule-week.component.spec.ts` still creates the real component without providing a mock for that service (`schedule-week.component.spec.ts:18-35`). `CalendarEventActionsService` pulls in `CalendarIntegrationService`, which injects `HttpClient`, plus dialog/snack dependencies. I do not see global test setup providing those. The new drag-service spec mocks this dependency, but this existing component spec should also mock `CalendarEventActionsService` (probably just `canMoveEvent`) or provide the required transitive test providers.

## Worth fixing

3. **A slow provider write leaves the dropped event interactive while it is still only a CDK transform.**

On release, opacity and pointer events are restored immediately (`schedule-week-drag.service.ts:225-227`). For calendar drops, the normal immediate reset is skipped, and reset happens only after `moveToStartTime(...).then(resetDrag)` resolves (`schedule-week-drag.service.ts:253-257`, `schedule-week-drag.service.ts:329`). That preserves the visual drop position, but during a slow or hung provider call the event is clickable/draggable at a DOM transform that is not backed by Angular state yet. Keeping pointer events disabled until `resetDrag`, or adding an explicit pending state, would avoid accidental second interactions.

4. **Calendar drags outside the grid still show a task-only unschedule preview.**

Calendar events are routed through time-mode preview, so outside the grid `_handleTimeDragMove()` falls into the task unschedule label and shows `Unschedule from Today` (`schedule-week-drag.service.ts:552-556`). But releasing a calendar event with no column just resets (`schedule-week-drag.service.ts:312-314`); it does not unschedule anything. For calendar events, the outside-grid preview should probably be suppressed or use a neutral invalid-drop state.

noamsiegel:
Addressed the latest pass in `886671c8`: extracted shared lazy plugin adapter resolution and switched calendar writes to use it, added the missing `ScheduleWeekComponent` mock, kept calendar drop pointer events disabled while provider writes are pending, and suppressed the task-only unschedule preview for outside-grid calendar drags.

Verified with the focused Angular specs covering the resolver/effect/calendar/schedule paths, `npm run lint`, and the Google Calendar provider tests.

johannesjo:
There is still a mismatch between the calendar write gate and adapter resolution here.

`canMoveEvent()` treats a plugin event as movable when the provider has `definition.updateIssue`, but `IssueSyncAdapterResolverService.getAdapter()` only creates a lazy adapter for update-capable providers when `fieldMappings?.length` is also present. `pushChanges()` itself does not need field mappings, so an update-only agenda provider can be shown as draggable and then `_pushChanges()` returns `false` without writing or showing feedback.

The inverse also happens through the dialog reschedule path: create/delete-only plugins can resolve an adapter, then `_pushChanges()` calls `adapter.pushChanges()` and hits `Plugin does not implement updateIssue`.

Can we align these gates around actual `updateIssue` support? For example, allow the resolver to create adapters for `definition.updateIssue`, and have calendar reschedule/move UI/actions require that same capability before offering or attempting writes.

noamsiegel:
Addressed the remaining gate mismatch in `c37348d1`: the lazy resolver now creates adapters for update-only providers, while reschedule UI/actions use the same `canMoveEvent()` / `updateIssue` gate as drag moves. That prevents both silent no-ops for update-only providers and attempted reschedule writes for create/delete-only providers.

I also fixed the failing CI test path by adding the missing `canMoveEvent` mock to `ScheduleComponent`’s spec; the earlier GitHub failure was `ScheduleComponent -> CalendarEventActionsService -> ... -> ScannedActionsSubject` from that missing mock.

Verified locally with focused Angular specs for resolver/calendar actions/schedule event/schedule component, `npm run lint`, and the Google Calendar provider tests.

johannesjo:
## Review feedback

Reviewed the diff across several focused passes (correctness, security/privacy, architecture, performance, simplicity) and verified each finding against the post-change code. The PR is well-structured and the privacy work is solid — but there's one blocking bug in the all-day reschedule path. Details below.

### 🔴 Blocking: date-only reschedule pushes `start_date: "NaN-NaN-NaN"`

In `calendar-event-actions.service.ts`, `reschedule()`'s no-time branch:

```ts
changes = { start_date: getDbDateStr(new Date(result.date + 'T00:00:00')) };
```

`DialogScheduleTaskComponent` returns `result.date` as a **`Date` object** (`dialog-schedule-task.component.ts` → `close({ date: this.selectedDate as Date, ... })`, typed `date: Date | null`). Concatenating a `Date` with `'T00:00:00'` coerces it to a locale string (`"Mon Mar 23 2026 …T00:00:00"`), so `new Date(...)` is **Invalid Date** and `getDbDateStr` returns `"NaN-NaN-NaN"`.

This branch is reachable whenever the dialog returns with no time (`time: this._normalizedTime()` is `null`) — i.e. the **default path for rescheduling an all-day event**, whose time field starts empty. The result is `start_date: "NaN-NaN-NaN"` sent to the provider, which rejects it and the reschedule fails.

The new spec masks this because it passes `date: '2026-03-23'` (a **string**) with a non-empty `time`, exercising only the timed branch. A `Date` input reproduces the bug.

**Fix** — `getDbDateStr` already accepts a `Date`:

```ts
changes = { start_date: getDbDateStr(result.date) };
```

and add a spec passing `date: new Date(2026, 2, 23), time: null` so the all-day path is actually covered. (The timed branch is fine — `getDateTimeFromClockString` does `new Date(date)` internally.)

### 🟡 Worth a comment: relaxed adapter gate

`issue-sync-adapter-resolver.service.ts` gates on `!createIssue && !deleteIssue && !updateIssue`, whereas the effect's original `_getAdapter` required `!(fieldMappings?.length && updateIssue)`. So a provider with `updateIssue` but no `fieldMappings` now gets an adapter where it previously didn't.

This is **verified benign** for task sync (empty `fieldMappings` → `hasPushableField` is `false` → `_pushChanges$` returns early before any network call) and is *needed* for the calendar case. But it's an intentional semantic change buried inside what reads as a pure extract-method refactor — please add a one-line comment in the resolver noting that `updateIssue` alone is intentional, so a future reader doesn't restore the stricter gate and silently break calendar moves.

### 🟢 Minor

- `_getSafeErrorTxt` fallback returns the i18n **key** `T.F.CALENDARS.S.CAL_PROVIDER_ERROR`, which gets interpolated into the toast as `"Calendar Provider Error: F.CALENDARS.S.CAL_PROVIDER_ERROR"`. Edge case (non-`Error` throwable with no `name`/`status`); return `''` or a generic translated string instead.
- `reschedule()` hand-rolls the `HH:mm` `targetTime` string — the existing `src/app/util/format-time-hhmm.ts` does exactly this and is already tested.
- Dead member: `schedule-week.component.ts` `protected readonly isDraggableSE = isDraggableSE;` has no remaining template references (template uses `canDragEvent`; the `.ts` calls the imported function directly). Can be deleted.
- `openEventLink`'s `Log.warn('…', e)` (pre-existing, same file) logs the raw error into the exportable log history — since this PR is the "sanitize calendar errors" change, consider reusing `this._sanitizeLogError(e)` there too.

### ✅ Verified clean

- The privacy fix is real: `_sanitizeLogError`/`_getSafeErrorTxt` use an allowlist (`name`/`status`/`statusText`), dropping `message`/`url`/response body — an `HttpErrorResponse` carrying an email/calendar-host does not leak into logs or toasts.
- `_openUrl`'s `^https?://` check is sound and stricter than sibling call sites; provider write payloads carry only app-generated values (no injection); `allowPrivateNetwork` SSRF gate preserved through the refactor.
- Drag/reset/undo async ordering is correct (timestamp captured before reset; transform reverts on success and failure; undo is non-recursive). No new template hot-path cost (`canDragEvent` is O(1), `canRescheduleCalendarEvent` is a memoized `computed()`). No NgRx/op-log invariants touched — writes go via the HTTP adapter, not reducers.

Overall: fix the all-day `NaN-NaN-NaN` path (+ its spec) and add the gate comment, and this is good to go. Nice work on the error sanitization.

noamsiegel:
Fixed the blocking all-day reschedule path in `39776e28`: the date-only branch now passes the dialog `Date` value directly to `getDbDateStr`, instead of concatenating it with a time string. I added a regression spec with `date: new Date(2026, 2, 23), time: null` that asserts `start_date: "2026-03-23"`.

Also added the resolver comment noting that `updateIssue` alone is intentional for calendar moves without task field mappings.

Verified with the focused calendar/resolver Angular specs, `npm run lint`, and the Google Calendar provider tests.

johannesjo:
Thank you very much! <3

Issue #7964, opened 2026-06-02
💡 Add search to project selection & due date picker during task creation


hajiboy95:
### Problem Statement

1. **Missing Search in Project Selection**: When creating a new task, selecting a project from a long list is tedious and slow because there is no search/filter option in the project dropdown or selection input.
2. **No Due Date on Task Creation**: Currently, there is no option to set a due date (deadline) directly when creating a new task. The user has to create the task first and then manually open and edit the task to set a due date, which disrupts the workflow.

### Possible Solution

1. **Searchable Project Selection**: Integrate a search filter or an autocomplete feature (e.g., using Angular Material autocomplete) in the project selection field when creating a task.
2. **Due Date Picker during Task Creation**: Add a due date picker/input directly into the task creation dialog or quick-add form.

### Describe alternatives you've considered

- Scrolling through a long list of projects (current behavior, inefficient).
- Creating tasks and immediately editing them afterwards (current behavior, requires too many clicks).

### Additional context

This is a proposal to improve the task creation user experience by reducing friction when setting projects and deadlines.

hajiboy95:
1. a bare search option is already given, a bit buggy  #7987 but given. Redesign is also discussed in this bug ticket. therefore this ticket can be closed

johannesjo:
Thank you very much @hajiboy95 !

PR #7966, opened 2026-06-02
Feat/task creation due date

hajiboy95:
## Problem

Users currently cannot set a deadline (Frist) when creating a task from the
quick-add task bar. They have to create the task first, then open the task
detail panel to set a deadline as a separate step.

Relates to #7964 . Implemented due date

## Solution

Added a **Deadline** button to the quick-add task bar action row, matching
the existing "Schedule" button pattern. Clicking it opens the standard
`DialogDeadlineComponent` in a new `isSelectDeadlineOnly` mode — since no
task ID exists yet, the dialog returns the chosen values instead of
dispatching directly to the store. The selected deadline is then applied to
the task's `deadlineDay`, `deadlineWithTime`, and `deadlineRemindAt` fields
at the moment of creation.

- 🚩 Deadline button + inline clear (×) button in the action bar
- `Ctrl+8` keyboard shortcut to open the deadline dialog
- `DialogDeadlineComponent` extended to work without an existing task
- Deadline state stored in `AddTaskBarState` and reset after each task add
- Translations added for all 28 supported languages via automated script

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


hajiboy95:
Just realized that the translation files shouldn't be touched and that 2 commit messages are not according to the best practives. will fix it in a few minuts

johannesjo:
Thanks @hajiboy95 — the implementation here is genuinely solid (the `isSelectDeadlineOnly` dialog refactor is clean and the creation path is correctly a single op). But I want to redirect the approach before you put more work in, so I'll be direct about what I'd like instead.

**I don't want to add another always-visible button to the add bar.** The action row already overflow-scrolls, so a 6th button mostly ends up off-screen on mobile — and putting a Deadline date-picker right next to the Schedule one makes the Schedule-vs-Deadline distinction *more* confusing, not less (the fact that it needs a "Deadline vs. Schedule" wiki section to explain is the tell).

**What I'd like instead: deadline support in the input short-syntax.** We already train users that markers set properties — `+project`, `#tag`, `@when` for scheduling. A deadline marker is the natural next member of that family: zero extra UI, consistent, works the same on mobile and desktop, and keeps hands on the keyboard during quick capture.

Concretely:
- A **single-character** marker (it has to be single-char to reuse the autocomplete dropdown) — I'm thinking `!` followed by a date, e.g. `Pay rent !friday` or `Taxes !apr 15`. It reads as urgency and doesn't collide with the existing markers.
- The parser should only consume the matched date span, and a bare trailing `!` (like "done!") must be a no-op.
- Show the parsed deadline back as its own **chip**, visually distinct from the schedule chip — that's also what resolves the confusion problem.

The good news is the dialog refactor and the `AddTaskBarState` deadline plumbing you built are exactly what this approach reuses, so it's not wasted. If you're up for it, I'd love to see this repurposed toward the `!` short-syntax (parser + chip + applying it at creation) and the button dropped. Happy to help on the parser side if useful.

hajiboy95:
@johannesjo thanks for your feedback. Let me refactor a bit such that I'll be able to implement it the way you suggested. That's a great idea to be honest. Let my try for a while. If I'll need extra help, I'll come back to you!!

hajiboy95:
@johannesjo could you take a look at it once again and tell me if that's what you were looking for? :D

hajiboy95:
@johannesjo it seems like one Test has failed. Not sure if that's really caused by my implementation though...
Could you help me out? Asked Claude, it just agreed that actually my changes shouldn't have caused this.

johannesjo:
Thanks @hajiboy95, this is exactly the direction I was hoping for. The `\!` short-syntax reads naturally next to `@`, the distinct chip resolves the schedule-vs-deadline confusion nicely, and it's great that the dialog refactor and `AddTaskBarState` plumbing carried straight over. Really nice turnaround.

On your question about the failing check: that one is not on you. It's the e2e test `recurring/recurring-preserve-past-due-day-bug-7344.spec.ts:88`, failing with `page.reload: net::ERR_ABORTED; maybe frame was detached?`. That's a flaky Playwright reload error in a recurring-task test that doesn't touch anything in this PR, and all the unit tests (including your new deadline specs) passed. A re-run should go green, so you can safely ignore it.

A few things I'd like to see addressed before merge:

**1. Combining a schedule and a deadline corrupts the title (the one real bug).**
In `short-syntax.ts`, `parseScheduledDate` and `parseDeadlineDate` both receive the same input title and each strips only its own marker, then both get spread into one object. The deadline result is spread last, so it overwrites the schedule result's cleaned title. Example I traced:

```
Input:  "Pay rent @monday \!friday"
schedule strips @monday -> "Pay rent  \!friday"
deadline strips \!friday -> "Pay rent @monday"
final title (deadline wins) -> "Pay rent @monday"   // @monday leaks into the title
```

Both dates still get set, but the task ends up named `Pay rent @monday`. The fix is to chain them so the second parses the first's already-stripped title:

```ts
const dueChanges = await parseScheduledDate(
  { ...task, title: taskChanges.title || task.title },
  now,
);
const deadlineChanges = await parseDeadlineDate(
  { ...task, title: dueChanges.title ?? (taskChanges.title || task.title) },
  now,
);
taskChanges = { ...taskChanges, ...dueChanges, ...deadlineChanges };
```

When `\!` doesn't match, `deadlineChanges.title` is undefined and won't overwrite, so the single-marker cases stay correct. A test for `@... \!...` together would be a good guard against regressions.

**2. Drop the unused `DialogDeadlineComponent` import.**
In `add-task-bar-actions.component.ts` it's added to the `imports` array, but the dialog is only opened via `MatDialog.open()`, never used in the template. That's what produces the build warning `NG8113: DialogDeadlineComponent is not used within the template of AddTaskBarActionsComponent`. Imperatively opened dialogs don't need to be in `imports`.

**3. Minor: the new "char before trigger must be a space" guard also affects `@`.**
Since the guard lives in the shared `parseShortSyntaxDate`, it now changes existing `@` parsing too (for example `lunch@12` with no leading space no longer parses a time). No existing test breaks, but it's a behavior change beyond deadlines, so either scope the guard to the deadline case or let's decide on it intentionally.

One small heads-up, not a blocker: `\!` shows up in normal task text a lot more than `@`/`#`/`+` ("Fix bug\!", "done\!"), so the autocomplete dropdown will trigger more often. That's inherent to reusing the dropdown and fine by me, just noting it.

Thanks again for being open to reworking this, it came out clean. Once the title-leak fix and the import cleanup are in, I think this is good to go.

hajiboy95:
@johannesjo Thanks for the feedback, gonna fix those things. That heads up point indeed is very true. 

I’ve been thinking about alternatives that fit naturally alongside `@`, `#`, and `+`, and I think we have two strong candidates: `<` and `^`. Here is a breakdown of the pros and cons for each:

### Option 1: `<` (Less Than)

**Pros:**

* **Almost Zero False Positives:** `<` is rarely used in standard task descriptions unless you are writing code blocks or HTML.
* **Excellent Semantics:** `<` naturally translates to "before" or "less than / up to" in logical contexts. A task like `Pay rent <friday` reads very intuitively as "Pay rent before/by Friday."
* **Keyboard Accessibility:** On most keyboards, it's just as easy to type (e.g., `Shift + ,` on US layouts) as `!` (`Shift + 1`).

**Cons:**

* **Developer Contexts:** If a user is writing code-related tasks (e.g., `Fix <input> styling`), it might trigger the dropdown. However, since the parser expects a date after it, it won't actually parse as a deadline, making it far less disruptive than a terminal exclamation point.

### Option 2: `^` (Caret)
**Pros:**

* **No Punctuation Collision:** Practically zero collision with natural language or standard punctuation.

**Cons:**

* **The "Dead Key" Problem:** On many international keyboards (like German QWERTZ, French AZERTY, and Nordic layouts), `^` is a dead key used to type diacritics (like `â`, `ê`). When a user presses `^`, nothing appears until they press `Space`. This turns `^friday` into a counter-intuitive, multi-step action (`^` + `Space` + `friday`), which defeats the purpose of a "short-syntax."
* **Semantic Mapping:** It usually represents exponentiation or a pointer, and doesn't scream "deadline" as naturally as `<`.
* **Reachability:** On US keyboards, `Shift + 6` requires a longer reach than `<`.


Let me know what you think is better! After that we can merge this one i guess. For me, both solutions would be better in everyday life.

tippfehlr:
How is it that I wanted to look for this feature for a while, and now that I look you created this PR *15 hours ago* 😅

johannesjo:
Let's keep **`!`** – it's the right call. The only downside is the dropdown popping up a bit more on natural text, which is harmless and closes on the next keystroke. `<` would trade that for actual mis-parses (`I <3 coffee`, `read <5 pages` → a deadline + mangled title), and `^` is a dead key on German/French/Nordic layouts. So no marker change needed.

The real items before merge, all independent of the marker:

- **`@` + `!` corrupts the title** – both parsers get the same input and the deadline spread wins, leaking `@monday` into the title. Fix it in `shortSyntax()` itself by chaining the second parse onto the first's stripped title; that also covers the global-edit path below.
- **`hasDeadlineTime` leaks onto the task** – the marker lives in the shared `shortSyntax()` core, so it runs on every title edit via `short-syntax.effects.ts`, not just the add bar. There the synthetic `hasDeadlineTime` rides inside `taskChanges` → `applyShortSyntax` → `updateOne` and gets persisted + synced. Return it as a sibling of `taskChanges`, or `delete` it before dispatch.
- **`!today` doesn't move to Today** – creation sets `deadlineDay` but skips `getDeadlineAutoPlanFields`, which contradicts the new wiki line. Wire the auto-plan fields in, or fix the doc.
- **Unused import** – drop `DialogDeadlineComponent` from the component imports (NG8113).

After these I'm happy to merge!

hajiboy95:
@johannesjo okay, i guess it should check now all the boxes.

johannesjo:
Thank you very much! <3

PR #7968, opened 2026-06-02
feat(tasks): add searchable project and tag selection to add-task-bar

hajiboy95:
##   Problem
Users with a large number of projects or tags currently have to scroll through a long list in the "Add Task" bar to find the right one, which is slow and cumbersome.

Relates to #7964 . Implements Search fields for projects and also for tag.
## Solution
   - Added a real-time search/filter input to both Project and Tag dropdown menus in the add-task-bar-actions component.
   - Implemented keyboard support: pressing Arrow Down while focused on the search input moves focus directly to the first filtered result.
   - Added auto-reset and auto-focus logic so the search field is ready as soon as the menu opens.
   - Polished the UI with rounded "pill-style" inputs and appropriate sticky headers for a modern look.
   - Added comprehensive unit tests for the filtering and reset logic.

## Type of Change
   - [x] New feature

## Checklist
   - [x] I have included relevant changes to the documentation/wiki. (User decided not to add)
   - [x] I have run npm run checkFile on changed .ts/.scss files.
   - [x] I have added tests for my changes (if applicable).
   - [x] My commit messages follow the Angular format (feat(tasks): add searchable project and tag selection to add-task-bar).


johannesjo:
Hi @hajiboy95, thanks a lot for taking the time to put this together. The implementation is clean and I really appreciate the tests and the attention to detail (sticky header, auto-focus, arrow-down to the first result). 🙏

Before we go further, I want to flag one thing so we don't duplicate work: the add-task bar already has fast, filterable selection for both projects and tags built in via the inline short-syntax. While typing a task you can use:

- `+` to search/pick a **project**
- `#` to search/pick a **tag**
- `@` to set a **due date**

These open an autocomplete that filters as you type (same `includes`-based matching, plus it sorts by match position), so the "scroll through a long list" pain from #7964 is mostly already covered there without opening the dropdown at all.

Where your PR genuinely adds something is the **mouse path**: when you click the project/tag pill to open the menu, there's currently no filter. So this is really about parity for that click-to-open flow.

Given that, I'm a little hesitant to add the extra surface area for something that overlaps so much with the existing feature. A couple of options:

1. We focus on making the `+`/`#`/`@` shortcuts more **discoverable** (a hint/placeholder), which might solve the underlying frustration with less code.
2. If you feel the dropdown filter is still worth it, I'd want to slim it down a bit first. The project and tag paths are nearly identical copy-paste right now and could share one helper.

What do you think? Were you mainly reaching for the menu because the `+`/`#` shortcuts weren't obvious? That feedback alone is really useful. Either way, thank you for contributing. Not trying to discourage you, just want to make sure we add the right thing. 🙂

hajiboy95:
@johannesjo Thank you so much for the feedback!

You're absolutely right—if that functionality is already there, a new addition isn't necessary. I'm actually a "keyboard over mouse" type too, so I love that! Since I'm quite new to the project, I just wasn't aware the solution already existed.

Now that you mention it, I see it's already implemented even in the preview (lol, I guess I never actually read the preview since I already immediately started typing... that's probably the main reason I missed it). When I looked at it just now, I did wonder for a second whether it was a filter or if I was defining a new one, but if I had just tried it out, it would have been obvious!

Thanks for pointing me in the right direction. I technically could refactor if you'd like to add this feature to the app. If not, I'm fine too, whatever feels comfortable to you :D. I like you approach on keeping things clean and minimal.

hajiboy95:
@johannesjo I made a few small changes to improve the UX—now, the text field is automatically focused as soon as you open it.

I also think this is still worth implementing because existing tasks currently don't have an option to filter through existing projects nor tags. Therefore I added it to those menus, too (standard shortcut "E" and "G"). We could remove it from the "New Task" view and only add it to the options for existing tasks, but that might feel a bit inconsistent.

What do you think?

johannesjo:
Hey @hajiboy95, thank you for iterating on this so quickly. The auto-focus polish and the tests are really nice work, and you clearly put care into it. 🙏

You raise a fair point about existing tasks, so let me share one thing that I don't think is obvious (it wasn't to me for a long time either): these menus already support **type-ahead** out of the box. If you open the project or tag menu and just start typing, it jumps straight to the matching entry, same "don't scroll through a long list" outcome with zero extra UI. Between that and the `+` / `#` / `@` inline syntax on the add bar, the underlying pain from #7964 is actually already covered on both the new-task and existing-task paths.

Because of that overlap, I've come around to thinking the dropdown filter adds a fair amount of permanent surface area (a second keyboard-nav layer next to the one Angular Material already gives us, plus some focus and mobile-keyboard edge cases) for a gain that's mostly already there. So I'm going to pass on merging this one. It's not a reflection on the code quality at all, I just want to keep the app lean.

Where I do think there's a real problem is **discoverability**, which is exactly what you bumped into ("if I'd just tried it, it would've been obvious"). The tricky part: we already show an example syntax in the add-bar placeholder (`A task title #tag @16:00 t25m`), and you still didn't catch it, so clearly that alone isn't enough, and I'm honestly not sure what the right fix is.

You're in the perfect spot to tell us, though. As someone who just ran into this fresh, what would have actually made the `+`/`#`/`@` shortcuts (and the menu type-ahead) click for you? A different placeholder, a one-time tip, an icon, something else entirely? I'd genuinely value your take, and it'd be a great smaller thing to take a swing at. 🙂

hajiboy95:
@johannesjo

I just checked, and you're right—that functionality is indeed already there! Discovering the tool's hidden depths is actually quite fun, but I completely agree that discoverability is the main hurdle here.

To answer your question on what would have helped me catch those shortcuts:

**1. Bite-Sized Tutorial Videos**
Instead of just text placeholders, a series of short tutorial videos explaining "hacks" would be amazing. Keeping them categorized and small (e.g., *Handling New Tasks*, *Handling Existing Tasks*, *Managing Projects*) would make them easy to digest. Seeing someone actually use the `+`, `#`, and `@` syntax in a quick 30-second clip would make it click instantly for a new user.

I actually checked out the Super Productivity YouTube channel to see if something like this already existed, but couldn't find a specific series for it. If these short clips were created, organizing them into a dedicated "Tips & Tricks" playlist would be huge for onboarding. It would also make the channel much easier for newcomers to navigate, as the current layout feels a bit cluttered and the video thumbnails don't immediately give away what the videos cover.

Regarding the current implementation, I noticed a couple of minor friction points where we might be able to find a middle ground:

**Suggestion 1: Reveal a text field *only* when typing**
With the current invisible type-ahead, if I make a typo while searching, there's no way back and no way to see what I typed. What if a text field naturally pops up only *after* the user starts typing? This keeps the UI lean and hidden by default, but provides an intuitive escape hatch for typos.

**Suggestion 2: Smarter default focus for the shortcut menu**
Right now, if I open the menu via shortcut, it highlights the tag at the very top of the list. A small quality-of-life tweak would be to change this initial focus based on likely user intent:

* Focus the **first *selected* tag** (if we assume the user opened the menu to remove a tag).
* *Or*, focus the **first *unselected* tag** (if we assume they want to add a new one).

I might be overthinking that last one, so keeping it as-is is completely fine too! Let me know what you think about these adjustments—I'd be happy to take a swing at one of them.

johannesjo:
Thanks for digging into this, and for the channel feedback! A categorized "Tips & Tricks" series is a good idea! Not sure if I will find the time though :D

On the two ideas: I'd skip both. Suggestion 2 you talked yourself out of, and I agree. Suggestion 1 isn't actually leaner: the permanent code (input, focus, mobile keyboard, tests) is the same as an always-visible field, only the mount timing changes, and echoing your keystrokes means disabling Material's type-ahead and re-implementing matching. More surface area for the feature I just passed on.

Discoverability is the real underlying issue, but I don't have a fix I'm confident in yet (the placeholder already shows `#tag @16:00` and still didn't catch your eye), so I'd rather leave it for now than add something half-baked. 

hanks again for sticking with this and for the genuinely useful feedback. 🙂

hajiboy95:
@johannesjo, there are some cases regarding the current menu behavior that we should probably discuss:

Currently, the dropdown only displays the project or tag name, completely ignoring the folder structure it belongs to. This creates two main friction points with the invisible type-ahead strategy:

**1. Identical names across different folders**
If I have two projects named "Project A" that live in different folders, the "jump to" strategy breaks down. Because the menu respects the custom side-panel ordering rather than alphabetical order, typing "Project A" only ever jumps to the first instance. There is currently no way to use the keyboard to reach the second "Project A". *(See attached screenshots)*

**2. Long or highly similar names**
If you have projects with long, similar prefixes (e.g., "Client Q3 - Marketing", "Client Q3 - Sales"), blindly typing to jump becomes tedious and error-prone.

This is where a visible text field (like the one I suggested) that actively *filters* the list—rather than just scrolling down to a match—would be a huge advantage. Another good point is that the filters also looks within the word itself, which the current implementation doesn't do either.

I'm not sure if the app's philosophy intentionally discourages duplicate or highly similar project names, but a real filter would solve this gracefully. I definitely wouldn't want to lose the custom sidebar ordering (it's crucial for sorting by priority!), but because the list isn't alphabetical, control towards certain object is harder than it needs to be (especially by keyboard). An active text filter would bridge that gap perfectly.

Just to be clear, that's not a made up scenario, instead it's actually really the case for me and the way I organize my things. I came across my second point and tried to have a Workaround but then came across point 1...

Here are two screenshots which should give an example of what I was referring too. I hope it makes sense!

grafik

grafik

johannesjo:
Hey @hajiboy95, this is a really good catch – thank you for pushing on it rather than letting it drop. 🙏 You've actually surfaced the one thing in this whole thread that _isn't_ already covered by the existing features, so it's worth getting right.

Let me split your scenario into its two parts, because they land quite differently:

**Long / similar prefixes ("Client Q3 - Marketing" vs "Client Q3 - Sales").** This one's already handled on the add bar – the inline `+`/`#` autocomplete filters by **substring**, not prefix, and sorts by match position. So after `+` you can just type `Sales` and it jumps straight there without caring about the shared prefix, and it shows you what you typed. The prefix-only limitation you hit is specific to the _type-ahead in the click-to-open menu_; the keyboard path on the add bar doesn't have it.

**Duplicate names across folders ("Project A" / "Project A").** This is the real gap, and you're right that it's frustrating. Two small clarifications on the mechanics, then the part that actually matters:

- The menu type-ahead _can_ technically reach the second one – Material's type-ahead wraps around, so typing the name again after a short pause cycles to the next match. Totally undiscoverable, I'll grant you, but it's not strictly unreachable.
- That doesn't really help though, because – and this is the actual root cause – **the list only renders the bare title**. Even once you've cycled onto the second "Project A", there's nothing on screen telling you it's the _other_ one.

And that's the key insight for me: a search field wouldn't fix this either. Filtering down to two identical "Project A" rows leaves you exactly as stuck. What actually solves it is **showing the folder / parent context in the option itself** – in both the menu and the inline autocomplete. That's a much smaller, leaner change than a filter input, and it fixes the problem at its source instead of working around it.

So where I've landed: I'm still going to pass on the filter field, but you've convinced me there's a genuine (if niche) disambiguation gap worth closing. If you'd enjoy taking a swing at it, surfacing the folder/parent next to the title in the project & tag pickers would be a tidy, well-scoped PR – and a concrete improvement for exactly the way you organize things. No pressure at all, and either way, thank you for the genuinely sharp feedback. 🙂

hajiboy95:
@johannesjo Thanks for your understanding.

Just to make sure we're fully aligned, here is my takeaway on the next steps:

**1. Adding Parent/Folder Context (I'll take a swing at this!)**
I agree that seeing the folder context solves the root of the disambiguation problem. I'll put together a PR to display the folder/parent name close to the project/tag title. I will make sure this is applied across the board:

* The dropdowns that open on mouse-click for new tasks.
* The inline `+`/`#` autocomplete list.
* The menus that open for existing tasks via the `E` (Project) or `G` (Tag) shortcuts.

**2. The Search Flow for Existing Tasks (`E` / `G` shortcuts)**
While the folder context fixes the *disambiguation* issue, the *navigation* issue for existing tasks still feels a bit unresolved. A lot of the time, my workflow is to quickly dump a task into the app and categorize it later using `E` or `G`.

While it's good to know the Material type-ahead cycles through duplicates if you pause and re-type, it’s practically invisible to the user, let alone intuitive. Since the `+`/`#` inline search isn't available for existing tasks, is there any middle-ground solution you’d be open to for improving the `E`/`G` menus—ideally something that allows active filtering by any substring, rather than just exact typing? If you've got nothing in mind, I can also just open a new Issue for this topic and then comment whenever I come up with a good idea :D

johannesjo:
Sounds like a good plan, and yes to all three surfaces. 🙏 One clarification on the mechanics and one answer on the `E`/`G` question.

**Folder context – go for it, for both projects and tags.** Quick heads-up on where the data lives so you don't go looking for the wrong thing: folder membership isn't a field on the project/tag itself. It's the `menu-tree` feature – `MenuTreeState` has both a `projectTree` and a `tagTree` (`src/app/features/menu-tree/store/menu-tree.model.ts`), and folders can be nested. So there's no `project.folderId` to read; you resolve "which folder is this in?" by walking the tree (`selectProjectTree` / `selectTagTree`) to find the node and its ancestor folder(s).

Two things that'll make this clean:

- Build the lookup **once** as a shared `entityId → folder path` helper and reuse it across all three surfaces (the click-to-open menus, the inline `+`/`#` autocomplete, and the `E`/`G` menus). Same helper everywhere keeps it consistent and small.
- Edge cases worth deciding up front: items at the top level (no folder → just show the title), and nested folders (I'd lean toward showing the immediate parent name, with the full path in a tooltip if it's deep). It's read-only display of synced state, so no sync-correctness concerns to worry about here.

**On the `E`/`G` search flow – you're right, and I want to be straight about it.** Folder context fixes _recognition_ (telling two "Project A"s apart once you see them), but it doesn't fix _navigation_ – getting to the right one among a long, custom-ordered list. And my earlier "the inline `+`/`#` already covers filtering" point genuinely doesn't apply to existing tasks: there's no inline syntax on that path, so the only filter is Material's invisible, prefix-only type-ahead. That's a real gap, and folder context won't close it.

That said, it's a separate decision from this PR – and it's the exact filter-vs-type-ahead question I went back and forth on, so it deserves its own thread rather than getting bundled in here. Please go ahead and open that issue like you offered. 😄 Let's keep this PR focused on folder context, and we can debate the `E`/`G` filter there on its own merits. Thanks again for the genuinely sharp back-and-forth on this.

PR #7976, opened 2026-06-03
[codex] Fix sync import conflict from startup example tasks

johannesjo:
## Summary

Fixes a fresh-client SuperSync edge case where first-run example tasks were created before an incoming remote `SYNC_IMPORT` was accepted. Those generated task-create operations made the import conflict gate treat the fresh client as having meaningful local user work.

The generated example-task creates are now explicitly marked, ignored by the incoming full-state conflict gate, and rejected locally when the remote import is accepted so they are not uploaded afterward.

## Validation

- `npm run checkFile src/app/op-log/sync/sync-import-conflict-gate.service.ts`
- `npm run checkFile src/app/op-log/sync/sync-import-conflict-gate.service.spec.ts`
- `npm run checkFile src/app/op-log/sync/operation-log-sync.service.ts`
- `npm run checkFile src/app/op-log/sync/operation-log-sync.service.spec.ts`
- `npm run checkFile src/app/core/example-tasks/example-tasks.service.ts`
- `npm run checkFile src/app/core/example-tasks/example-tasks.service.spec.ts`
- `npm run checkFile src/app/root-store/meta/task-shared.actions.ts`
- `npm run test:file src/app/op-log/sync/sync-import-conflict-gate.service.spec.ts`
- `npm run test:file src/app/core/example-tasks/example-tasks.service.spec.ts`
- `npm run test:file src/app/op-log/sync/operation-log-sync.service.spec.ts`
- `git diff --check`

### DIFF
--- a/e2e/tests/sync/supersync-example-task-fresh-client.spec.ts
+++ b/e2e/tests/sync/supersync-example-task-fresh-client.spec.ts
@@ -1,0 +1,116 @@
+import { test, expect } from '../../fixtures/supersync.fixture';
+import {
+  createTestUser,
+  getSuperSyncConfig,
+  createSimulatedClient,
+  closeClient,
+  waitForTask,
+  getTaskTitles,
+  type SimulatedE2EClient,
+} from '../../utils/supersync-helpers';
+
+/**
+ * Regression test for #7976 — a fresh client's first-run example tasks must not trip the
+ * SYNC_IMPORT conflict gate when pulling an account that already has data.
+ *
+ * THE BUG: ExampleTasksService creates onboarding task-create ops on first run. When a
+ * fresh client then syncs an account that already has remote data, those pending ops used
+ * to make the conflict gate treat the client as having meaningful local work, so it showed
+ * a `dialog-sync-import-conflict` instead of silently accepting the import.
+ *
+ * WHAT THIS GUARDS: the op-log `isExampleTask` marker + gate exclusion. It does NOT
+ * exercise the afterInitialSyncDoneStrict$ timing — on a fresh e2e client sync is disabled
+ * at boot, so example tasks are created before sync is configured regardless of the strict
+ * gate; the marker is what suppresses the dialog here. Covering the strict-wait timing
+ * needs a client with sync pre-enabled at boot (a separate test).
+ *
+ * METHOD: configure sync with `waitForInitialSync: false` so setupSuperSync does NOT
+ * auto-resolve the conflict dialog (it otherwise clicks "Use Server Data"), then race
+ * "conflict dialog visible" against "sync complete". The fix means sync completes WITHOUT
+ * the dialog; pre-fix the dialog wins the race.
+ *
+ * REPRODUCE-FIRST: expected to FAIL on `bb4b625645^` (race resolves to 'dialog') and PASS
+ * on the PR head. Run on both before trusting it.
+ *
+ * Run: npm run e2e:supersync:file e2e/tests/sync/supersync-example-task-fresh-client.spec.ts -- --retries=0
+ */
+const EXAMPLE_TASK_TITLES = [
+  'Create your first project',
+  'Set up Sync',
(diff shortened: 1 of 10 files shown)


johannesjo:
## Multi-agent review — summary

Ran a 6-reviewer parallel pass (correctness · security · architecture · alternatives · performance · simplicity) plus my own verification against the source. **The code is correct, safe, and well-tested — no bugs found.** Everything below is about whether there's a simpler root-cause fix.

### ✅ Verified clean
- **Security**: the gate reads `isExampleTask` only from local `getUnsynced()` ops (remote ops always carry `syncedAt` → excluded), so a remote/malicious flag can't be used to bypass the conflict dialog.
- `isExampleTask` survives capture into `op.payload.actionPayload` and is inert in `Task` state (CRUD reducer only reads `task`).
- `markRejected` running before `processRemoteOps` is the safe failure direction (crash → re-download next sync, never silent overwrite).
- `discardablePendingOpIds` is non-optional and populated on all return paths; both consumers are behind `if (conflict.fullStateOp)` — no undefined access.

### 💡 Simpler alternative worth considering: defer creation via `afterInitialSyncDoneStrict$`
The root cause is timing. `ExampleTasksService` subscribes to `afterInitialSyncDoneAndDataLoadedInitially$`, which short-circuits to `of(true)` immediately for SuperSync (`sync-trigger.service.ts:189`), so example tasks are created *before* the import lands.

You already ship `afterInitialSyncDoneStrict$` for exactly this — *"ALWAYS waits for the actual initial sync to complete, even for SuperSync … to prevent duplicate tasks across clients"* — and it's already used by `task-due.effects.ts` / `task-repeat-cleanup.effects.ts`. `setInitialSyncDone(true)` fires only after `syncWrapperService.sync()` resolves (`sync.effects.ts:203`), i.e. after the import is applied — so the `tasks.length === 0` filter would then see the imported tasks and skip creation entirely.

I prototyped the swap (1 line in `example-tasks.service.ts` + 1-line spec-mock rename) — all 4 example-tasks specs pass. If adopted, it would **delete** the action prop, the gate change, `discardablePendingOpIds`, and both `markRejected` calls.

**Caveat (why not just swap blindly):** strict-wait covers the common case (server already has data at first sync). It does *not* cover "fresh client syncs an empty server → example tasks legitimately created → a real import arrives in a *later* sync." The new test's `seq=1` setup looks like it models that later-import ordering, which the marker approach *does* handle. So the two aren't equivalent — strict-wait is simpler and fixes more providers; the marker is more complete for late imports. Worth a conscious call on whether late-import matters.

### ⚠️ Scope gap: file-based sync (Dropbox/WebDAV) isn't fixed
`operation-log-sync.service.ts:445-446` also gates on `hasMeaningfulStoreData()`, which returns true for any `task.ids.length > 0` (`has-meaningful-state-data.util.ts:37`) — blind to `isExampleTask`. So a fresh file-based client with only example tasks still throws `LocalDataConflictError` → the same dialog this PR removes for SuperSync. If that's SuperSync-only by design, a one-line scope note would help. (The `afterInitialSyncDoneStrict$` alternative fixes this path for free, since no example tasks exist when those checks run.)

### ⚠️ Marker leaks into synced payloads (inert, but worth noting)
In the mixed case (real task + example tasks both pending → dialog → USE_LOCAL), the example-task ops get force-uploaded carrying `isExampleTask: true` into the server op-log / peers — the dialog path computes but never rejects `discardablePendingOpIds`. Harmless today (no consumer reads it from remote ops), but a transient bootstrap flag becoming durable synced data.

### Nits (if you keep this approach)
- DRY: the `markRejected` guard block is duplicated at `operation-log-sync.service.ts:253` & `:686` → a 2-line private helper.
- `discardablePendingOpIds` could be computed at its single consumer instead of threaded through `IncomingFullStateConflictGateResult`.

---
<sub>Reviewers: 6× Claude + manual verification. Codex CLI was unavailable in my sandbox (401).</sub>

PR #7983, opened 2026-06-03
fix(task-repeat-cfg): create real instance when startDate moved to today with no live instance

### LINKED ISSUE #7923: 🚨 Recurring Tasks: Deleting the non-transparent instance then changing the start date of a recurring task to Today doesn't create a non-transparent instance on Today

MannXo:
## Problem

After deleting a recurring task's live instance and then moving `startDate` to today via a transparent planner projection, today's slot showed only a transparent projection — no real task was created.

Fixes #7923

## Root Cause

Two related bugs, both in the path `rescheduleTaskOnRepeatCfgUpdate$` → `addAllDueToday()` → `createRepeatableTask()`:

**Bug 1 — store read race in `addAllDueToday`.**
`rescheduleTaskOnRepeatCfgUpdate$` re-anchors `lastTaskCreationDay` to yesterday via a synchronous `store.dispatch()`, then immediately calls `addAllDueToday()`. NgRx updates the store state synchronously, but the async context means `selectAllUnprocessedTaskRepeatCfgs` can still observe the stale future `lastTaskCreationDay` (the old anchor) when `getAllUnprocessedRepeatableTasks$` is read in the next microtask. Result: 0 configs returned → no task created.

**Bug 2 — `taskAlreadyExists` path skipped `lastTaskCreationDay` update.**
When a concurrent `addAllDueToday()` call (e.g. from `createRepeatableTasksAndAddDueToday$` on app init) had already created the task, `_getActionsForTaskRepeatCfg` returned `[]` without dispatching `updateTaskRepeatCfg`. This left `lastTaskCreationDay` at yesterday, so `selectTaskRepeatCfgsForExactDay` kept returning the config for today → the transparent projection persisted alongside the real task.

## Solution

**Fix 1 (`add-tasks-for-tomorrow.service.ts`):** Yield to the event loop with `await new Promise(r => setTimeout(r, 0))` before reading the store in `addAllDueToday()`. This ensures the re-anchor state is visible before the selector is evaluated. The pattern mirrors the existing yield already used later in the same function and in `OperationApplierService`.

**Fix 2 (`task-repeat-cfg.service.ts`):** When `taskAlreadyExists`, still dispatch `updateTaskRepeatCfg({ lastTaskCreationDay: targetDateStr })` to advance the anchor and suppress the transparent projection. No new task is created — only the watermark is moved forward.

## Type of Change

- [x] Bug fix

## Checklist

## Tests

- **Unit tests (65/65):** 7 existing tests updated to reflect the new `taskAlreadyExists` behaviour (returns `[updateTaskRepeatCfg]` instead of `[]`); 2 new regression tests added.
- **E2e test:** `recurring-startdate-to-today-creates-real-instance-bug-7923.spec.ts` reproduces the exact repro steps from the issue (fixed clock, delete live instance, change startDate to today via transparent projection) and asserts that a real task appears in the Today view with no residual transparent projection.

### DIFF
--- a/e2e/tests/recurring/recurring-startdate-to-today-creates-real-instance-bug-7923.spec.ts
+++ b/e2e/tests/recurring/recurring-startdate-to-today-creates-real-instance-bug-7923.spec.ts
@@ -1,0 +1,179 @@
+import { Locator, Page } from '@playwright/test';
+import { expect, test } from '../../fixtures/test.fixture';
+
+/**
+ * Bug: https://github.com/super-productivity/super-productivity/issues/7923
+ *
+ * When the user:
+ *   1. Creates a recurring (daily) task with a future startDate
+ *   2. Deletes the materialised live instance (plain delete — does NOT touch
+ *      deletedInstanceDates or lastTaskCreationDay)
+ *   3. Opens the repeat config from a transparent projection and moves
+ *      startDate to TODAY
+ *
+ * Expected: a real (non-transparent) task instance is created for today and
+ *           added to the Today work-view.
+ * Bug:      today keeps showing only a transparent projection; no real task
+ *           is created.
+ *
+ * Root cause: `rescheduleTaskOnRepeatCfgUpdate$` correctly re-anchors
+ * `lastTaskCreationDay` to yesterday and calls `addAllDueToday()` (#7768 Bug 2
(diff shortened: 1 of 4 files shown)


johannesjo:
Thank you very much! <3

PR #7988, opened 2026-06-03
feat(menu-tree): show parent folder context for projects/tags with ch…

hajiboy95:
# feat(menu-tree): show parent folder context for projects/tags with chevron separators and globalized grid layout

## Problem

Currently, when multiple projects or tags share the same name across different folders (e.g., "Project A" under "Folder 1" and "Project A" under "Folder 2"), they look identical in dropdowns (e.g. project/tag pickers, move-to-project menus) and autocomplete panels. This makes it impossible for users to distinguish between them without trial-and-error. 

Additionally, the inline autocomplete mention panel had tag icons prefixing the title, which was visually inconsistent and cluttered.

> **Related discussion/PR:** https://github.com/super-productivity/super-productivity/pull/7968#issuecomment-4613011860

## Solution

1. **Tree Hierarchy computed mappings:** Walk the project/tag menu trees in `MenuTreeService` to compute `projectFolderMap` and `tagFolderMap` mapping tag and project IDs to their parent paths with ` › ` chevron separators.
2. **Text alignment and grid overrides:**
   - Designed a global `.custom-option-layout` grid layout in `_overwrite-material.scss` to push the folder subtitle to row 2 and start at the same vertical alignment as the title on row 1 (completely left-aligned).
   - Removed folder icons across all templates to reduce visual noise.
   - Vertically centered option icons/checkboxes by having them span 2 rows.
3. **Right-aligned suggestions badge:**
   - Modified `mat-option` suggestions in `add-task-bar` to render the task title first, followed by the context `<tag>`.
   - Styled the badge in `add-task-bar.component.scss` to be right-aligned (`margin-left: auto`) and hid its default tag icons/circles to make it look like a clean text pill.
4. **Mock Store safety:** 
   - Added robust `|| of([])` selector fallbacks inside `MenuTreeService` to prevent tests in other modules from throwing subscription errors when mocking the NgRx `Store`.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have added tests for my changes (if applicable) (Created unit test spec file `src/app/features/menu-tree/menu-tree.service.spec.ts` which runs and passes)
- [x] Existing tests still pass (the modified component unit tests pass successfully, `54 of 54 SUCCESS`)


johannesjo:
Thanks for putting this together, and for the tests and clean walk logic. The underlying problem is real: same-named projects/tags across folders are genuinely impossible to tell apart in the pickers.

That said, I'd like to scope this down before merging. My main concern is proportionality: the PR shows the folder path on *every* item, but it only adds value in the rare case where two items actually share a name. For everyone else it's extra visual noise in menus they use constantly, which I think is a net negative for the common case.

I'd prefer a more minimal version:

1. **Only disambiguate on an actual name collision.** Detect duplicate titles (within projects, and within tags) and show the folder hint only for those. When there's no collision, menus look exactly as they do today, so the vast majority of users see no change at all.
2. **Use a plain inline suffix** rather than a new two-row layout, e.g. `Taxes (Work)` or `Taxes › Work`. That avoids the biggest issue with the current approach: the `::ng-deep .mat-mdc-menu-panel` overrides are emitted globally (no `:host` scope) and copy-pasted into three component stylesheets, so they restyle every menu in the app, not just these. For a niche feature that's a lot of blast radius and regression risk on unrelated menus. A simple suffix needs no global CSS at all.

That keeps the disambiguation logic in `MenuTreeService` (collision-aware lookup) and a small template tweak per menu, while dropping the four `.scss` additions, the 68-line grid in `_overwrite-material.scss`, and the icon/layout restructuring.

Two smaller things regardless of direction:

- `getFolderContext(item: any)` / `isEmojiIcon(tag: any)` should avoid `any` per our TS rules, and the collision-aware version can likely drop the `'backlogTaskIds' in item` duck-typing by just looking the id up in both maps.
- `this._store.select(...) || of([])` is a no-op in production (`select()` never returns falsy). If it's working around incomplete `Store` mocks in other specs, better to fix those mocks than keep the guard.

Happy to look again once it's scoped to the collision case. Thanks again for tackling this!

hajiboy95:
@johannesjo thanks for the feedback, I'll dig into it. 

For your second point the plain inline suffix. Tried it but it looked quite uggly in 2 cases
1. More than 1 Folder (so Work/CompanyA/BigProject)
2. the project name OR the folger name is a bit longer. (so both need to be rather short).
I don't think these two cases are very rare. What do you think?

johannesjo:
Thanks for taking this on. The folder-context idea is exactly what we agreed on, and it degrades nicely (nothing shows for users without folders).

One structural suggestion before we merge. The option markup (`icon + title + folder subtitle`) is now copy-pasted across the project dropdown, the autocomplete, and the `E`/`G` menus, and the same `::ng-deep .mat-mdc-menu-panel { height: auto … }` block is duplicated in three component SCSS files. Since `::ng-deep` here isn't host-scoped (the panel renders in an overlay), each copy actually leaks globally onto every menu in the app. You can also see the duplication cost already: the right-click **task context menu** (`task-context-menu-inner`) renders the same project list but didn't get folder context, so it's now inconsistent.

Could we extract a small **presentational** row component instead, something like `<select-option-row [project]="project" />` (or one that takes a `Project | Tag` and looks up the folder path via `MenuTreeService` itself)? Then every menu collapses to one tag, the layout + height CSS lives in one host-scoped place (no global leak), and the context menu gets folder context for free. It'd also let us drop the `item: any` + `'backlogTaskIds' in item` duck-typing in `getFolderContext` in favor of real types.

To keep scope tight: I'd **only** extract the row content, not a full selection widget. The call sites differ too much in behavior (`updateProjectId` vs `moveTaskToProject` vs tag toggle, and the autocomplete isn't even a `mat-menu`), so unifying that layer isn't worth it.

Also minor: the `|| of([])` fallbacks on `store.select(...)` are no-ops (`select()` never returns null). If the new spec needed them, the fix belongs in the mock setup.

hajiboy95:
@johannesjo implemented your suggestion. For your last comment, give me a second to check.

I also tried once more the plain inline suffix. this is what it would look like

grafik

Which option do you prefer? The previous or this one? 

Currently on the branch is the one of this screenshot, but i could revert and go to the 2 line version (which is a bit more aesthetic in my opinion

hajiboy95:
@johannesjo also did some refactor work now to make the whole implementation a lot cleaner. 
Let me know if it's going into the right direction. If it's too far off from what you had in mind, it's probably the better way to start again from scratch haha.

johannesjo:
Thank you very much @hajiboy95 ! The previous style probably looked a little better, but this one fits the case better I think. I think it is ok that it doesn't look as clean, given this is more of a rarer edge case? What do you think?

hajiboy95:
@johannesjo I checked it one more time—it seems there was just a temporary rendering bug earlier. Everything is aligned well now! :D
grafik

Since this is an edge case, I think this layout is perfectly acceptable as is. I'm good to go with this. If you're okay with it, we can merge. Thanks for your patience! 

Just a heads up on two things: First, if an item is at the top level, we don't show a folder. Second, if two items in the same folder have the exact same name, we currently are only able to differentiate them by the symbol's color.

It might be worth discussing whether we should block the ability of creating two objects with the exact same name in the same layer. I'm a bit concerned that allowing two labels with identical names in the same layer might cause unexpected bugs down the line... what do you think?

johannesjo:
Thanks for iterating on this. I did another pass and found a few things I think should be fixed before merge:

1. `src/app/features/tasks/add-task-bar/add-task-bar.component.html:14` / `:155` now applies the `select-option-row` mention template to every mention item. That includes `@`/`!` chrono suggestions from `CHRONO_SUGGESTIONS`, which are plain strings and previously rendered as simple text. They now become idless `{ title }` items and hit the default `label` icon path in `select-option-row.component.ts:51`. Could we keep the row template scoped to project/tag suggestions, or make title-only/string suggestions render plainly?

2. `src/app/features/tasks/add-task-bar/add-task-bar.component.scss:256` adds a bare `::ng-deep .mention-menu` rule. That selector is global and also affects task-title mention dropdowns. Since `select-option-row.component.scss:98` already has a narrower `:has(select-option-row)` rule, I think the broad add-task-bar rule should be removed or moved to the shared mention styles intentionally.

3. `MenuTreeService` counts duplicate titles from `selectAllProjects` / `selectAllTags`, but the menus render filtered lists. For example, project menus exclude archived/hidden projects, and tag menus exclude the virtual Today tag. This can show folder context even when there is no visible duplicate in the menu. Since the goal is to disambiguate actual visible collisions, the count should be based on the same candidate set the menu renders, or the lookup should be context-aware.

On the same-folder identical-name case: I would not block this PR on preventing duplicate names. That feels like a separate validation/product-policy discussion.

hajiboy95:
@johannesjo I think we're good to go, let's wait for the test results :)

johannesjo:
Thanks for the latest iteration. I double-checked the current head and found two remaining things I think should be fixed before merge:

1. `src/app/features/tasks/add-task-bar/add-task-bar.component.html:141` / `src/app/features/tasks/add-task-bar/add-task-bar.component.scss:212`: the task autocomplete context tag can become invisible when tag titles are hidden. The template still passes `[isHideTitle]="isHideTagTitles()"` into `<tag>`, while the new autocomplete CSS hides both `.tag-ico` and `.circle`. In that setting the tag component renders neither title nor icon/circle, so the context marker disappears. Could we keep the icon/circle visible when titles are hidden, or force this autocomplete badge to show a title?

2. `src/app/features/menu-tree/menu-tree.service.ts:55`: `projectFolderMap` now filters archived/hidden projects, which is good, but it still counts duplicate project titles globally across all visible projects. The move-project menus render a narrower candidate set: `getProjectsWithoutId$()` excludes the current project (`src/app/features/project/project.service.ts:97`, `src/app/features/project/store/project.selectors.ts:101`), then `task.component.html:318` and `task-context-menu-inner.component.html:371` render that filtered list through `select-option-row`. If the current project is one of two same-titled projects, the only visible move target can still get a folder subtitle even though there is no visible collision in that menu. To keep the "only disambiguate visible collisions" behavior, the folder lookup probably needs to be candidate-scoped for these menus, or the row component needs a candidate-scoped map/context.

I would not block this PR on the broader duplicate-name validation topic; that still feels separate.

hajiboy95:
@johannesjo addressed concernes and merges with current head of master

Regarding the current errors: I didn't touch the arrow keys, i wonder why it occures

johannesjo:
## 🤖 Multi-agent review (6 focused reviewers + manual verification)

Nice work — consolidating ~5 hand-rolled icon/title blocks into one `select-option-row` and adding folder-path disambiguation is a real DRY win, and the feature behaves correctly. I ran six independent reviewers (correctness, security, architecture, simplicity, performance, alternatives) and re-verified every finding against the code. Summary below.

**Security & performance: clean.** All new text rendering is Angular interpolation — no new `[innerHTML]` or `bypassSecurityTrust*`; the one `[innerHTML]` in the suggestion list is the pre-existing binding, just repositioned. The per-row `folder()` collision scan is technically O(n²) but bounded by small project/tag counts, runs once on panel-open, and is signal-memoized — not a concern.

### 🔴 Should fix before merge

**1. Dead code left over from an earlier revision.** `select-option-row` injects `MenuTreeService` itself, so this host-side wiring is unused:
- `task/task.component.ts` — `MenuTreeService` import (L92) + `_menuTreeService` inject (L152): never referenced. *(Keep the `SelectOptionRowComponent` import — that one is used.)*
- `task-context-menu-inner.component.ts` — `_menuTreeService` (L124) + the `projectFolderMap`/`tagFolderMap` computeds (L172-173) + import (L82): not referenced in the template (`FolderMap` has 0 uses outside the two definitions).
- `add-task-bar-actions.component.ts` — `isTagEmojiIcon(tag: any)` (L206) has no remaining call site (and uses `any`). `isProjectEmojiIcon` / `DEFAULT_PROJECT_ICON` are still used — keep those.

These pass lint only because ESLint doesn't flag unused class members; they'll mislead future readers about the data flow.

**2. Unused `CommonModule`** in `select-option-row.component.ts` (L13/L27) — the template uses only `@if` + `MatIcon`. Drop it.

### 🟡 Design / worth addressing

**3. `select-option-row` lives under `src/app/ui/` (generic UI) but depends on the domain `MenuTreeService`.** A presentational row now transitively pulls in the whole menu-tree feature. Consider keeping it purely presentational and letting each caller pass a precomputed `[subtitle]`/`[folderPath]`. This one change also dissolves #5 (type-sniffing) and the split collision responsibility.

**4. Move-to-project: the folder hint is hidden in exactly the two-same-named case it's meant for.** `folder()` shows the subtitle only when there's a title collision *within `allOptions`*, but the move-to-project list comes from `getProjectsWithoutId$()`, which **excludes the current project**. So with two "Marketing" projects, opening the menu from one shows the other as a plain "Marketing" (no folder) — the service map *has* the entry, but the local `allOptions` check discards it. Is disambiguation meant to be relative to the visible list, or to all entities? Worth a deliberate decision — and a test, since the current specs don't cover the subset case.

**5. Structural type-sniffing is fragile.** `'backlogTaskIds' in item` (project) and `'color' in item` (tag) in `select-option-row.component.ts` work today but couple the component to incidental model shape. An explicit `[defaultIcon]`/`[color]` input from the caller (which already knows the kind) would be more robust. *(I checked — this is **not** an active bug: short-syntax project items always set `icon`, so the default-icon branch is never hit.)*

**6. DRY:** `projectFolderMap` and `tagFolderMap` (`menu-tree.service.ts:55-121`) are near-identical ~33-line computeds differing only by source signal, one filter, and the `MenuTreeKind`. Extract one `_buildFolderMap(entities, tree, kind)` helper. (The service file is also now >300 lines.)

**7. `::ng-deep … :has(select-option-row)`** (`select-option-row.component.scss:74-105`): three top-level `::ng-deep` blocks compile to global rules that restyle menu/autocomplete/mention overlay panels app-wide from a leaf component. Functionally fine, but these belong in `_overwrite-material.scss` (which already owns option/menu-item height tweaks); the menu-panel and autocomplete-panel blocks are identical and can merge.

**8. Magic string `'TODAY'`** (`menu-tree.service.ts:91`) → use the `TODAY_TAG.id` constant per the virtual-tag convention.

**9. `menuTouchFix` was dropped from the tag-toggle button** in `task-context-menu-inner.component.html` while sibling buttons keep it — intentional? *(The neighbouring `track $index` → `track tag.id` change is a nice improvement, btw.)*

Happy to help with any of these — thanks for the contribution! 🙏

hajiboy95:
@johannesjo adressed and fixed. Tests are running now

johannesjo:
Thanks for the quick iteration. I double-checked the current head and most of the earlier cleanup/design notes look addressed.

One remaining behavior I think we should settle before merge:

1. `projectFolderMap` is still global-visible, but the move-project menus render a narrower candidate list that excludes the current project. The rows then use the global lookup directly:
   - `src/app/features/menu-tree/menu-tree.service.ts:56`
   - `src/app/features/tasks/task/task.component.ts:708`
   - `src/app/features/tasks/task/task.component.html:328`
   - `src/app/features/tasks/task-context-menu/task-context-menu-inner/task-context-menu-inner.component.ts:169`
   - `src/app/features/tasks/task-context-menu/task-context-menu-inner/task-context-menu-inner.component.html:382`

   So if there are exactly two visible projects named "Marketing" and the current task is already in one of them, the move menu shows only the other project but still gives it a folder subtitle. If the intended behavior is "only disambiguate collisions visible in the current menu," this should be candidate-scoped for the move menus. If global disambiguation is intentional, I would at least add a regression test for that excluded-current-project case so the semantics are explicit.

Nit: `select-option-row` declares `folderPath` as `string | null`, but most callers pass `Map.get(...)`, which returns `string | undefined`. Not a blocker under the current template settings, but the input contract should either accept `undefined` or callers should coalesce with `?? null`.

hajiboy95:
@johannesjo behaviour is intended to be like this but i added an test just like you suggested. also adressed the nit

johannesjo:
Thank you very much!

PR #7989, opened 2026-06-03
feat: Fix visual inconsistencies in icons

hajiboy95:
## Problem

The symbols and icons representing projects and tags were inconsistent across different parts of the application:
1. When triggering project and tag autocompletion in the "Add Task" input bar (using `#` and `+`), standard projects and tags had no icons and fallback colors shown in the suggestion dropdown.
2. In task and note selection context menus (opened via click or keyboard shortcut `E`), the standard fallback icon for projects differed from the standard `list_alt` symbol used in the sidebar (using generic folder or label icons instead).

## Solution

* **Add Task Bar Autocomplete & Chips (`#` and `+` mentions)**:
  * Updated [short-syntax-to-tags.ts](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/app/features/tasks/add-task-bar/short-syntax-to-tags.ts) to fallback projects to `DEFAULT_PROJECT_ICON` ('list_alt') and tags to `'label'`.
  * Configured `mentionListTemplate` in [add-task-bar.component.html](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/app/features/tasks/add-task-bar/add-task-bar.component.html) to render fallback project (`list_alt`) and tag (`label`) symbols.
  * Added active color bindings `[style.color]="item.color || item.theme?.primary"` to standard fallback icons inside the dropdown so they match the project/tag theme.
  * Added fallback support `{{ item.title || item }}` to handle string-only suggestions (like relative dates triggered by `@` or `!`) and display them with a `schedule` icon.
  * Added flex layout and alignment rules for `.custom-option-layout` in [add-task-bar.component.scss](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/app/features/tasks/add-task-bar/add-task-bar.component.scss) to vertically center icons and emojis with text labels.
  * Removed unused `_menuTreeService` injection and `getFolderContext()` method from [add-task-bar.component.ts](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/app/features/tasks/add-task-bar/add-task-bar.component.ts).
* **Task & Note Context Menus**:
  * Standardized fallback icons to use `DEFAULT_PROJECT_ICON` ('list_alt') inside task components ([task.component.ts](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/app/features/tasks/task/task.component.ts) & [task.component.html](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/app/features/tasks/task/task.component.html)), task context menus ([task-context-menu-inner.component.ts](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/app/features/tasks/task-context-menu/task-context-menu-inner/task-context-menu-inner.component.ts) & [task-context-menu-inner.component.html](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/app/features/tasks/task-context-menu/task-context-menu-inner/task-context-menu-inner.component.html)), and note components ([note.component.ts](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/app/features/note/note/note.component.ts) & [note.component.html](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/app/features/note/note/note.component.html)).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring (cleaned up unused services and methods)
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

  * *Reason for no cross*: This visual change displays existing standard model fields and does not introduce new API parameters or configurations that require wiki updates.
- [x] I have run `npm run checkFile` on changed `.ts`/`.scss` files.
- [ ] I have added tests for my changes (if applicable).
  * *Reason for no cross*: The changes are visual presentation and default fallback assignments on existing models. Existing test suites already cover task parsing and mentions logic.
- [x] My commit messages follow the Angular format (`type(scope): description`).

### DIFF
--- a/src/app/features/note/note/note.component.html
+++ b/src/app/features/note/note/note.component.html
@@ -120,7 +120,7 @@
             <mat-icon
               [style.color]="project.theme.primary"
               class="tag-ico"
-              >{{ project.icon || 'list' }}</mat-icon
+              >{{ project.icon || DEFAULT_PROJECT_ICON }}</mat-icon
             >
             {{ project.title }}
           </button>
--- a/src/app/features/note/note/note.component.ts
+++ b/src/app/features/note/note/note.component.ts
@@ -35,6 +35,7 @@ import {
 import { AsyncPipe } from '@angular/common';
 import { TranslatePipe } from '@ngx-translate/core';
 import { DEFAULT_PROJECT_COLOR } from '../../work-context/work-context.const';
+import { DEFAULT_PROJECT_ICON } from '../../project/project.const';
 import { ClipboardImageService } from '../../../core/clipboard-image/clipboard-image.service';
 import { RenderLinksPipe } from '../../../ui/pipes/render-links.pipe';
 
@@ -86,6 +87,7 @@ export class NoteComponent implements OnChanges {
   resolvedShortenedContent = signal<string>('');
 
   T: typeof T = T;
+  readonly DEFAULT_PROJECT_ICON = DEFAULT_PROJECT_ICON;
 
   projectTag$: Observable<TagComponentTag | null> =
     this._workContextService.activeWorkContextTypeAndId$.pipe(
@@ -102,7 +104,7 @@ export class NoteComponent implements OnChanges {
                           ? {
                               ...project,
                               color: project.theme?.primary || DEFAULT_PROJECT_COLOR,
-                              icon: 'list',
+                              icon: project.icon || DEFAULT_PROJECT_ICON,
                               theme: {
                                 primary: project.theme?.primary || DEFAULT_PROJECT_COLOR,
                               },
--- a/src/app/features/tasks/add-task-bar/add-task-bar-actions/add-task-bar-actions.component.html
+++ b/src/app/features/tasks/add-task-bar/add-task-bar-actions/add-task-bar-actions.component.html
@@ -196,12 +196,12 @@
         <span
           matMenuItemIcon
           class="menu-icon-emoji"
-          [style.color]="project.theme.primary"
+          [style.color]="project.theme?.primary"
         >
           {{ project.icon || DEFAULT_PROJECT_ICON }}
         </span>
       } @else {
-        <mat-icon [style.color]="project.theme.primary">
+        <mat-icon [style.color]="project.theme?.primary">
           {{ project.icon || DEFAULT_PROJECT_ICON }}
         </mat-icon>
       }
@@ -226,13 +226,13 @@
         @if (isTagEmojiIcon(tag)) {
           <span
             matMenuItemIcon
-            [style.color]="tag.color || tag.theme.primary"
+            [style.color]="tag.color || tag.theme?.primary"
             class="tag-ico tag-ico-emoji"
             >{{ tag.icon || 'label' }}</span
           >
         } @else {
           <mat-icon
-            [style.color]="tag.color || tag.theme.primary"
+            [style.color]="tag.color || tag.theme?.primary"
             class="tag-ico"
             >{{ tag.icon || 'label' }}
           </mat-icon>
--- a/src/app/features/tasks/add-task-bar/add-task-bar-mentions.spec.ts
+++ b/src/app/features/tasks/add-task-bar/add-task-bar-mentions.spec.ts
@@ -18,6 +18,11 @@ import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { DEFAULT_FIRST_DAY_OF_WEEK, DEFAULT_LOCALE } from 'src/app/core/locale.constants';
 import { DateAdapter } from '@angular/material/core';
 import { DateTimeFormatService } from '../../../core/date-time-format/date-time-format.service';
+import { DEFAULT_PROJECT_ICON } from '../../project/project.const';
+import {
+  DEFAULT_PROJECT_COLOR,
+  DEFAULT_TAG_COLOR,
+} from '../../work-context/work-context.const';
 
 @Component({
   template: `<add-task-bar></add-task-bar>`,
@@ -41,6 +46,7 @@ describe('AddTaskBarComponent Mentions Integration', () => {
       theme: { primary: '#00ff00' },
     } as Tag,
     { id: '3', title: 'Testing', color: '#0000ff', theme: { primary: '#0000ff' } } as Tag,
+    { id: 't4', title: 'EmojiTag', icon: '🎨' } as Tag,
   ];
 
   const invalidTags = [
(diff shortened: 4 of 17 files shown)


johannesjo:
Thank you very much\!

This is a genuinely nice consistency fix — standardizing on `DEFAULT_PROJECT_ICON` and giving standard projects/tags icons in the add-task-bar dropdown is exactly right. I reviewed it across correctness, security, performance, architecture, and simplicity, and found **no bugs, no security issues, and negligible performance impact**. A couple of things worth addressing before merge, then a few optional cleanups.

### Worth addressing

**1. The inline-task-title editor won't get the new icons.** `MentionConfigService` is the shared config source for both `AddTaskBarComponent` *and* `TaskTitleComponent`, but the new template is only wired into the add-task-bar (`add-task-bar.component.html:14`). `task-title.component.html:36` binds `[mentionConfig]` without `[mentionListTemplate]`, so editing a task title inline still shows the icon-less default dropdown for the same suggestions. That leaves the two editors visually inconsistent — a bit ironic for this PR. Ideally the icon template lives in the mention layer so both editors share it; at minimum, let's track TaskTitle as a follow-up.

**2. Type the two helpers.** `isEmojiIcon(tag: any)` and `isProjectItem(item: any)` (`add-task-bar.component.ts:889,893`) — `any` won't fail CI (the rule is off) but the project guidelines prefer real types. `MentionItem | string` (or `Tag | Project | string`) works here. A short comment on *why* `'backlogTaskIds' in item` distinguishes a project from a tag would also help future readers.

### Optional / nice-to-have

- **Dead conditions:** in the template `@if (item.date || item.time || \!item.id)`, `item.date`/`item.time` are never set (date suggestions are plain strings; Tag/Project have no such fields), so `@if (\!item.id)` is equivalent and clearer.
- **SCSS tidy (`add-task-bar.component.scss`):** `.option-main-icon` and `.menu-icon-emoji` are identical except `font-size` — consider sharing one rule; `display: inline-flex` can be `flex` (each row is a block-level `<a>`); `vertical-align: middle` has no effect on a flex container; and `gap: var(--s, 8px)` doesn't need the fallback (`--s` is defined). Also, the existing emoji convention uses class names like `tag-ico-emoji` / `nav-icon-emoji` if you'd like to align.
- **Reuse:** the emoji-vs-`mat-icon` pattern already exists in several components (and in the `<tag>` component). Resolving the icon once when building the mention items (in `MentionConfigService`) would let the template stay dumb and drop `isProjectItem` entirely — a good follow-up rather than a blocker.
- **A small test** for the new project/tag/date-string branching would be welcome — `add-task-bar-mentions.spec.ts` is a natural home.
- **Minor:** `task.component.ts` now imports from `project.const` twice (the pre-existing `_MISSING_PROJECT_` import and the new `DEFAULT_PROJECT_ICON` one) — worth merging, as you already did in `task-context-menu-inner.component.ts`. And just to confirm: the tag fallback change `'style'` → `'label'` is an intentional visual choice (it matches `tag-toggle-menu-list`).

Thanks again for the careful PR description and for tackling the cross-component icon inconsistency — really appreciated\! 🙏

hajiboy95:
@johannesjo updated according to your suggestions. Slowly but steady we probably get to a point of common ground :D

johannesjo:
Thanks for the cleanup pass! Centralizing the icon/color resolution into `MentionConfigService` and dropping the duplicated template + SCSS + helper methods is a nice DRY win, and the explicit `MentionListItem` typing is much better than the previous `as unknown as MentionItem[]` casts. I verified the two things I was most worried about and both are fine:

- **Chrono `@`/`!` insertion still works** — `CHRONO_SUGGESTIONS` going from `string[]` to objects is safe because the directive's `mentionSelect` resolves `item[labelKey]` with `labelKey: 'title'`, so it still inserts the title text.
- **The generic template degrades gracefully** — `@if (item.icon)` and `{{ item[labelKey] || item }}` cover label-only/string items.

A few small things before merge:

**1. Dead code left behind in `add-task-bar.component.ts`.** The deleted `ng-template` was the only consumer of `DEFAULT_PROJECT_ICON`, but the import (line ~85) and the `readonly DEFAULT_PROJECT_ICON = DEFAULT_PROJECT_ICON;` property (line ~132) are still there. Lint won't flag a public class member (it could be template-bound), so it slipped through — safe to remove both.

**2. `isProject` looks vestigial now.** It's set on mapped project items (and asserted in the spec), but the new default template no longer branches on it the way the old `isProjectItem()` did — every item gets an `icon`, so nothing reads `isProject` anymore. Suggest dropping it (and the matching test assertion), or adding a comment if it's intended for future use. The `backlogTaskIds` comment next to it is also stale since that detection no longer happens here.

**3. Heads-up that this also changes the task-title dropdown.** Since the layout moved into `MentionListComponent`'s *default* template and `task-title` shares the same service, the inline task-title mention list now renders icons too (it previously used the plain `{{ item[labelKey] }}`). I think that's exactly what you'd want for consistency — just flagging it since the description only mentions the add-task bar.

**4. Minor design note (optional).** `MentionListComponent` is a vendored generic angular-mentions fork, and its default template now pulls in `MatIcon` and assumes our app-specific item shape. The `mentionListTemplate` input was the "clean" extension point for this. It's a reasonable KISS tradeoff given there are only two consumers and the fallback is graceful — just want it to be a conscious choice rather than an accident.

**5. Tests (nice-to-have).** The new assertions mostly check `toBeDefined()`. It'd strengthen them to assert the actual fallback resolution — e.g. a tag with no icon → `icon === 'label'`, a project with no theme → `color === DEFAULT_PROJECT_COLOR`.

Also FWIW — the red **Tests** check is an unrelated E2E flake (`YEARLY-recurring` failing on `page.reload: net::ERR_ABORTED`, 248/249 e2e green, all unit tests green both timezones), not anything from this change. A re-run should clear it.

Nothing here is a correctness or security concern — the refactor is solid. 👍

hajiboy95:
@johannesjo thank you very much for all the suggestions and sorry once more that it took so long in the end. However i think it's worth it for overall better consistency^^. 

Added 1., 2. and 5.. If you're okay we could merge.

johannesjo:
I think there is one shared UI regression to fix before merging.

`MentionListComponent` now renders colored icons in the default item template:

- `src/app/ui/mentions/mention-list.component.ts:38`
- `src/app/ui/mentions/mention-list.component.ts:45`

Those elements get an inline `[style.color]="item.color"`, but the active row contrast rule only applies to the parent link:

- `src/app/ui/mentions/mention-list.component.scss:59`

Because inline color wins over inherited `color: var(--c-contrast)`, active tag/project icons can keep arbitrary tag/project colors on the active `--c-primary` background. This already existed in the old AddTaskBar custom template, but this PR moves the renderer into the shared default mention list, so TaskTitle mentions now inherit the same contrast issue too.

Could we make active-row icons use the active contrast color, for example by overriding `.mention-active` child icon colors, while preserving tag/project colors for inactive rows?

hajiboy95:
@johannesjo fixed. Also added another test for that case

johannesjo:
Thank you very much ! <3

Issue #7985, opened 2026-06-03
SuperSync: two narrow residuals in startup example-task import handling (#7976 follow-up)


johannesjo:
Follow-up to #7976 (startup example tasks vs incoming `SYNC_IMPORT`). That PR fixes the common case via `afterInitialSyncDoneStrict$` + an op-log `isExampleTask` marker, and documents two narrow, non-corrupting residuals in code. Tracking them here.

Both assume the chosen model: onboarding example tasks **do** sync (account-level), so they are created/uploaded as normal task ops.

## 1. Piggyback upload path leaves example-task ops on the server (op-log/snapshot pollution)

`OperationLogUploadService` marks accepted ops synced (`src/app/op-log/sync/operation-log-upload.service.ts:260`) **before** piggybacked ops are processed. So when the upload→piggyback `SYNC_IMPORT` branch computes the discard list from `getUnsynced()` (`src/app/op-log/sync/operation-log-sync.service.ts:253`), example ops accepted earlier in the **same upload round** have already left the pending set and remain on the server.

- **Reachable:** a client with pending example-task creates is not "wholly fresh" (`lastSeq > 0`, `src/app/op-log/sync/sync-local-state.service.ts:22`), so the upload guard does not block it.
- **Impact:** state stays correct — receivers drop the example ops as `CONCURRENT` against the import (`SyncImportFilterService`). The residual is op-log / snapshot pollution, not data corruption.
- **Window:** narrow — example tasks created on a still-empty server and uploaded just as a remote import arrives; `afterInitialSyncDoneStrict$` shrinks it further.
- **Possible fix:** snapshot pending example-op IDs before upload and reject them in the piggyback branch (prevents re-upload, not the initial server write), or reconsider whether onboarding tasks should upload at all.

## 2. File-based snapshot sync still conflicts on an example-only store

The file-snapshot conflict check ORs pending-op meaning with store meaning (`src/app/op-log/sync/operation-log-sync.service.ts:445-447`), and `hasMeaningfulStateData` counts any task (`src/app/op-log/validation/has-meaningful-state-data.util.ts:37`). Example tasks carry the `isExampleTask` marker only on their op-log ops, not in NgRx state, so a store containing only example tasks still reads as "meaningful."

[...]

PR #7998, opened 2026-06-03
fix(sync): clear example tasks from local state on adoption (#7996)

### LINKED ISSUE #7996: SuperSync: remove lingering onboarding example tasks from local state on first adoption (follow-up to #7995)
Follow-up to #7995 (which resolved the #7985 residuals; itself a follow-up to #7976 / #7980).

## Summary

PR #7995 stops a never-synced client from **uploading** its onboarding example-task ops onto a populated remote it just adopted (the deferred "Fix C" of #7980). On the **file-based** path the example tasks are also gone from local state, because snapshot hydration *replaces* NgRx state. But on the **SuperSync (op-based)** path, adoption *merges* remote ops into the existing store, so the example **tasks themselves remain in the adopting device's local NgRx state** — only their op-log create ops were rejected.

## Why it matters

1. **Cosmetic clutter:** the just-onboarded device shows the 4 example tasks alongside the adopted real account data; other devices don't have them.
2. **Re-pollution vector (the real reason):** the example tasks live in NgRx but not on the server. A later **full-state snapshot upload** — `forceUploadLocalState` via a USE_LOCAL conflict choice, or enabling encryption (`SYNC_IMPORT` with `isCleanSlate: true`) — snapshots current NgRx (incl. the lingering examples) and re-uploads them, re-polluting the remote (and, being clean-slate, replacing it with local-including-examples). This was flagged in the #7995 multi-review and is documented there as accepted scope.

## Proposed work

When a never-synced client first adopts a populated remote (same trigger as the #7995 cleanup: pre-download-captured `isNeverSynced` + live `hasSyncedOps()` true + a pristine post-boot batch), also **remove the untouched example tasks from local NgRx state**, not just reject their ops.

Two candidate approaches — needs a design decision:

- **Option A — surgical delete.** Dispatch a delete for the untouched example task ids (available from the rejected example-create ops' `entityId`s) through the existing task-shared delete meta-reducer (so `project.taskIds` / tag ordering references are cleaned too). The delete creates op-log Delete ops → reject those as well so they don't upload a delete for a task the server never had. Smallest behavior change.
- **Option B — replace-not-merge on adoption.** For a never-synced client adopting a populated remote, reset local state before applying remote ops (mirroring `forceDownloadRemoteState`: `loadAllData(default)` → apply), so the throwaway example state is wiped as part of adoption — aligning SuperSync with the file-based replace semantics. Conceptually cleaner ("never-synced adoption replaces, not merges") but a larger change to the core adoption path; needs care with concurrent local ops (none expected in the pristine-batch case) and CLAUDE.md sync rule 7.

Recommendation: lean **Option A** (surgical, low blast radius); evaluate B only if the merge-vs-replace semantics are worth unifying.

## Safety / constraints (carry over from #7995)

- Fire **only** on the same gated condition (`isNeverSynced` pre-download + adopted + pristine post-boot batch) — never when the user has done real work, edited an example, reordered (a `Move`), etc. An edited example must still sync as real data.
- Must not touch real (non-example) tasks/projects/tags/notes.
- Must not leave dangling references in `project.taskIds` / tag ordering (use the proper delete meta-reducer).
- `isExampleTask` lives only on the op-log op, never the NgRx `Task` entity — the example ids must be derived from the local pending example-create ops (never from remote ops).

## Acceptance criteria

- After a never-synced SuperSync client adopts a populated remote, the onboarding example tasks are **absent from local NgRx state**, with no dangling list references.
- No example task — and no delete-of-example op — reaches the server.
- A subsequent `forceUploadLocalState` / enable-encryption from that device does **not** re-introduce example tasks to the remote.
- Real or edited data is untouched; the gate still skips when any non-config op is pending.

## Tests

- Unit: extend `operation-log-sync.service.spec.ts` to assert the removal dispatch (Option A) fires for the gated case and not otherwise.
- e2e (real shell — docker not runnable in the agent sandbox): extend `e2e/tests/sync/supersync-example-task-fresh-client.spec.ts` so a never-synced client adopting a populated **non-encrypted** account ends with a clean local store (no example tasks) and no example ops on the server, then a forced local-state upload does not re-add them.

## Refs

- PR #7995 (Fix C; this residual is its documented scope note)
- `docs/plans/2026-06-03-fresh-client-first-sync-data-loss-trap.md` §10 (the identity-based-cleanup rationale)
- `src/app/op-log/sync/operation-log-sync.service.ts` — the `uploadPendingOps` cleanup hook added in #7995

johannesjo:
Closes #7996. Follow-up to #7995 (merged) → #7985 → #7976/#7980.

## Problem
#7995 stops a never-synced client from **uploading** its 4 onboarding example-task create ops when it adopts a populated SuperSync remote. But on the op-based path, adoption **merges** remote ops into the store rather than replacing it, so the example **entities** lingered in local NgRx state. Left there, a later full-state snapshot upload (`forceUploadLocalState` via a USE_LOCAL choice, or enabling encryption) would re-snapshot and **re-upload** them, re-polluting the remote.

`isExampleTask` lives only on the op-log op, never the NgRx `Task` entity — so the snapshot can't distinguish example from real tasks, and the fix **must** touch local state (no upload/snapshot-layer shortcut is possible). The file-based path doesn't have this residual because snapshot hydration replaces state wholesale.

## Fix
In the same gated #7995 cleanup block (`isNeverSynced` pre-download + `hasSyncedOps()` now true + a pristine post-boot batch), after rejecting the create ops, also dispatch the shared bulk `TaskSharedActions.deleteTasks({ taskIds })` — scrubbing `project.taskIds` / backlog / tag refs in one meta-reducer pass (sync rule #3) — with **`meta.isRemote: true`**. That:
- makes the capture meta-reducer **skip op-capture** → no Delete op is created, so nothing uploads a delete for a task the server never had;
- keeps `LOCAL_ACTIONS` effects (plugin `TASK_DELETE`, reminder cascade) from firing for throwaway scaffolding.

It's the same primitive `ValidateStateService` and the op-replay path use; the state reducers still run. Task ids derive **only** from the local pending example-create ops, never remote ops. The gate still skips entirely if any non-config op is pending (real task / edit / reorder), so an edited example still syncs as real data.

> The mechanic was simplified from an initial "generate a Delete op then reject it" approach to the `meta.isRemote` dispatch after a multi-agent review surfaced the cleaner primitive — it removes a fragile capture-deferral invariant and closes the plugin-side-effect concern outright. Rationale in `docs/plans/2026-06-03-supersync-example-task-local-cleanup-7996.md` §9.

## Tests
- **Unit** (`operation-log-sync.service.spec.ts`, +3): the `isRemote` bulk delete fires only on the gated condition (not on a real task), rejects only the example create ops, and never the coexisting `GLOBAL_CONFIG` op.
- **e2e** (`supersync-example-task-fresh-client.spec.ts`, +1): a never-synced client adopting a populated **non-encrypted** account ends with a clean local store, and a third observer device never sees the example tasks (no cross-device propagation).

⚠️ The SuperSync e2e is **docker-only and not runnable in CI here** — please run it in a real shell. Reproduce-first: the local-store-clean assertion fails before this change (op-based merge keeps the examples) and passes after.

All touched `.ts` files pass `npm run checkFile`; unit + op-log integration specs green.

### DIFF
--- a/docs/plans/2026-06-03-supersync-example-task-local-cleanup-7996.md
+++ b/docs/plans/2026-06-03-supersync-example-task-local-cleanup-7996.md
@@ -1,0 +1,394 @@
+# SuperSync: remove lingering onboarding example tasks from local state on first adoption
+
+**Issue:** #7996 (follow-up to #7995 → #7985 → #7976/#7980)
+**Date:** 2026-06-03
+**Status:** **Implemented (2026-06-03)** on `feat/issue-7996-3b222c` after #7995 merged to
+`master` (PR #7995, commit `3e58b6d8d2`) and this branch was fast-forwarded. **Final mechanic
+is the `meta.isRemote` dispatch — see §9, which supersedes the A1 "generate-then-reject" design
+in §3** (a second multi-review found the simpler primitive). Unit
+(`operation-log-sync.service.spec.ts`, +3) specs green; the op-log integration spec
+(`example-task-adoption-cleanup.integration.spec.ts`) keeps its #7995 cases (the #7996 half is
+a pure NgRx dispatch with no op-log effect); e2e (`#7996` test added to
+`supersync-example-task-fresh-client.spec.ts`) authored — **docker-only, run in a real shell**
+(reproduce-first: local-store-clean assertion fails pre-fix). Decision was (b) sequence after
+#7995.
+**Area:** op-log sync (`src/app/op-log/sync/operation-log-sync.service.ts`)
+
+---
+
+## 0. Hard prerequisite — this builds on #7995, which is NOT yet merged
+
+#7996 extends the **`uploadPendingOps` identity-based cleanup hook added by PR #7995**
+(branch `feat/issue-7985-7f39b1`, still **OPEN**). That hook does not exist on `master`
+nor on this worktree's base. What master/this branch already has (from #7976/#7980) is the
+_older_ `_discardExampleTaskOps` call on the **piggyback** path (`operation-log-sync.service.ts:284`)
+and the `isNeverSynced` threading — **not** the new pre-upload cleanup block.
+
+The #7995 hook we extend sits just after the `isWhollyFreshClient` guard in `uploadPendingOps`
+(reproduced below; note #7995 imports `isExampleTaskCreateOp` from
+`op-log/validation/is-example-task-op.util`, it is not inline):
+
+```ts
+if (isNeverSyncedAtSyncStart && (await this.opLogStore.hasSyncedOps())) {
+  const pendingOps = await this.opLogStore.getUnsynced();
+  const isPristinePostBootBatch = pendingOps.every(
+    (entry) => isExampleTaskCreateOp(entry) || entry.op.entityType === 'GLOBAL_CONFIG',
+  );
+  if (isPristinePostBootBatch) {
+    const exampleOpIds = pendingOps.filter(isExampleTaskCreateOp).map((e) => e.op.id);
+    if (exampleOpIds.length > 0) {
+      await this._discardExampleTaskOps(exampleOpIds);
+      OpLog.normal(/* … discarded N untouched example-task op(s) … */);
+    }
+  }
+}
+```
+
+**Sequencing decision required (pick one before coding):**
+
+- **(a) Land #7996 on top of `feat/issue-7985-7f39b1`** and merge it together / right after #7995.
+- **(b) Wait for #7995 to merge to `master`, then rebase this branch** and implement.
+
+There is a **third option raised in review and worth serious consideration:**
+
+- **(c) Fold #7996 INTO #7995.** #7995 rejects the example _create ops_; #7996 removes the
+  _entities those ops created_. They are two halves of one invariant — _a never-synced adopter
+  ends up with neither the example ops nor the example entities_. Shipping #7995 alone merges a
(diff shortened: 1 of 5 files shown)


johannesjo:
Closing — superseded by the reasoning in #7997.

#7996 extends "Fix C", the `uploadPendingOps` adoption-cleanup hook that #7995 landed. **#7997 reverted that hook** as targeting a near-unreachable edge, and that reasoning applies to this PR too:

- SuperSync **mandates encryption**, so a fresh account is always seeded as an encrypted `SYNC_IMPORT`. A 2nd device adopting it hits the incoming-import gate, which discards the example ops on download **and replaces local state via snapshot hydration** — so the example *entities* are gone from NgRx as well (same as the file-based path).
- The lingering-entities residual this PR targets therefore only exists on the **non-encrypted / no-`SYNC_IMPORT`** path (legacy / unencrypted-backup-import) — the residual #7980 §10 already accepted as not worth fixing. (This PR's e2e had to set `isEncryptionEnabled: false` to reproduce it at all.)

So #7996 re-litigates an accepted residual for a practically-dead scenario — the same KISS/YAGNI call as #7997. Not merging.

The implementation (the `meta.isRemote` local-state cleanup + tests + design doc) stays on `feat/issue-7996-3b222c` if the scenario ever becomes reachable (e.g. if mandatory encryption is relaxed).

PR #7999, opened 2026-06-03
fix: address review edge cases across repeat-cleanup, sync, add-task & Android

johannesjo:
## Summary

A batch of edge-case fixes surfaced by review across several independent subsystems, plus a follow-up commit addressing a multi-agent review of the first. Each area is small and self-contained; grouped below for review.

### Recurring-task cleanup (`task-repeat-cfg`)
- Broaden the "user edited this overdue instance" guard so skip-overdue cleanup never discards an instance the user actually changed — now covers title, tags, estimate, project, deadline, reminder, attachments/notes and edited subtask templates (previously only attachments + notes).
- **Fix:** generated instances strip the virtual `TODAY` tag, so the new tag check must compare against the cfg tags with `TODAY` removed — otherwise every unedited overdue instance of a `TODAY`-tagged config looked "edited" and was never reaped. Regression test added.
- Extracted the subtask-template comparison into a shared `sub-task-template.util` used by both the cleanup and repeat-cfg effects (+ unit spec).

### Add-task bar & short syntax
- Track whether a due date / deadline came from parsing vs. user state so previously-parsed values aren't re-preserved as user input on the next keystroke.
- Guard `isValidSplitTime()` before building `dueWithTime`; fall back to a date-only `dueDay` for malformed times.
- Date-only deadline syntax now stays a `deadlineDay` instead of a midnight `deadlineWithTime`.
- `normalize-clock-str` only strips a seconds segment when it is numeric and in range, so genuine garbage stays invalid.
- Select-due-only schedule dialog returns `null` for an invalid normalized time.

### Sync / op-log
- Thread the pre-download `isNeverSynced` snapshot through LWW re-upload and rejected-op download callbacks so the never-synced guard isn't disarmed mid-session.
- Fail closed: throw if a provider has encryption enabled but no encryption key is available, instead of silently uploading plaintext.

### Onboarding
- After the sync dialog closes with sync enabled, only auto-dismiss onboarding once sync has actually restored data (avoids overwriting a just-synced `appFeatures` config), and add a keyboard focus-visible ring to the preset buttons.

### Android
- Queue a stop the foreground service consumes immediately after promotion when a stop arrives while a start is still pending (avoids `ForegroundServiceDidNotStartInTimeException`).
- WebView "open app info" prefers the system WebView settings picker over a guessed package; reworded the clear-storage warning.

### Misc
- `config-page`: suppress the inherited native version-link tooltip.
- e2e: drop a leftover debug `console.log`.

## Testing
- `npm run checkFile` clean on all modified `.ts` files.
- Affected unit suites pass (repeat-cleanup, repeat-cfg effects, sub-task-template util, add-task-bar parser/component, short-syntax, onboarding, sync-wrapper, op-log sync/upload, normalize-clock-str).
- Android changes not compiled locally (no Gradle in this environment).


johannesjo:
Closing as **superseded by master**. This was a six-subsystem batch opened June 3; in the ~140 commits since, every fix here was independently re-landed through smaller, focused PRs — in several cases in a more complete form:

| Area | Landed on master |
|---|---|
| repeat-cleanup edited-guard + TODAY-tag + reminder check | ✅ — and more robust: compares `task.remindAt === expectedRemindAt` *and* handles the no-startTime case (avoids the never-reaped false positive) |
| add-task-bar `isValidSplitTime` → `dueDay` fallback | ✅ |
| short-syntax date-only `deadlineDay` | ✅ |
| sync `isNeverSynced` snapshot threading + fail-closed encryption | ✅ |
| onboarding dismiss-after-restore | ✅ |
| Android FGS start/stop race | ✅ via #7991 |
| WebView "prefer settings picker" | ✅ |

The Copilot review comments resolve with the close: the reminder edge case (comment 1) is already covered better on master, and the `consumeStopPendingAfterStart()` atomicity comments (2/3) target a queue-stop-after-promotion enhancement that was never part of #7991 and isn't shipping.

The only behavior here not on master is that queue-stop-after-promotion enhancement (stop immediately after promotion instead of waiting for the next foreground sync) — a marginal improvement to a rare backgrounded-stop path. Can re-land separately if it proves needed.

PR #8000, opened 2026-06-03
Fix frequency translations

hajiboy95:
## Problem

This pull request addresses two translation and display bugs in repeating task configurations:
1. **Grammatical Ordinal Inflections (German):** The repeat settings menu previously used nominative ordinals ("erste", "zweite", "dritte", etc.) rather than the correct dative inflections ("ersten", "zweiten", "dritten", etc.) needed for German monthly weekday settings (e.g. "Jeden Monat am erste...").
2. **Missing Active Language Fallback:** The date/time formatting locale used by the application did not dynamically track the active UI translation language when no manual override locale was configured, resulting in English date formats/weekdays in localized environments.
3. **Complex Monthly Repeat Label Display:** In info panels and preview items, complex monthly repeat configurations (e.g., "first Saturday" or "last day of the month") were displayed using a fallback numeric day of month label (e.g. "Monatlich am 3.") instead of the proper description (e.g. "Monatlich am ersten Samstag" or "Monthly on the first Saturday").

## Solution

1. **Reactive Locale Fallback:** Modified [DateTimeFormatService](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/app/core/date-time-format/date-time-format.service.ts) to read the active UI translation language as a fallback using a signal `_localeSig` that tracks translation language changes. This dynamically updates weekday string formatting according to the UI language.
2. **Inflected Translation Keys:** Created new translation keys (`ORD_FIRST_NTH`, `ORD_SECOND_NTH`, etc.) mapping to dative forms (e.g. "ersten" in German) in [en.json](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/assets/i18n/en.json) and [de.json](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/assets/i18n/de.json).
3. **Descriptive Info Labels:** Updated [getTaskRepeatInfoText](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/app/features/tasks/task-detail-panel/get-task-repeat-info-text.util.ts) to handle complex monthly rules (N-th weekday and last day of month) and passed down Angular's `TranslateService` to resolve nested translation keys.
4. **Integration across Components:** Passed `TranslateService` into the `getTaskRepeatInfoText` call inside tag-list, task-detail-panel, repeat-cfg-preview, scheduled-list-page, and dialog-view-archived-task components.
5. **Testing:** Added new unit test cases covering monthly last day and N-th weekday repeat rules in [get-task-repeat-info-text.util.spec.ts](file:///Users/Finbridge/Desktop/msc/Github/super-productivity/src/app/features/tasks/task-detail-panel/get-task-repeat-info-text.util.spec.ts).

For Test purposes I also added the german translations. Don't know how they'll get translated. If they get overwritten somehow, I hope they'll consider the grammar pattern since in english it makes no difference. :D

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki). *(N/A: Localization and grammatical fixes do not require changes to documentation or wiki).*
- [ ] I have run `npm run checkFile` on changed `.ts`/`.scss` files *(N/A: Code sanity check was performed automatically by the git pre-commit hook which validated and linted all modified files).*


johannesjo:
Thanks for tackling this — the German grammar fix and the missing monthly variants are real bugs. A few things to address before merging:

### Correctness

**1. Resetting "Datetime format locale" to auto leaves a stale locale.** `DateTimeFormatService` (`date-time-format.service.ts:16`/`:66`): `_localeSig` is only written by `setDateAdapterLocale()`, and the constructor effect at line 59 is a no-op when `dateTimeLocale` becomes `null`/`undefined`. Result: if a user previously set an explicit locale and then clears it, `currentLocale()` keeps returning the old explicit value instead of falling back to the active UI language. The effect needs to also push the UI language into `_localeSig` (or clear it) when the override is removed.

**2. Monthly label precedence diverges from recurrence behavior.** `get-task-repeat-info-text.util.ts:136`: the recurrence utilities let a valid nth-weekday anchor win over `monthlyLastDay`, but the new display branch checks `monthlyLastDay` first. A stale/malformed cfg with both flags set will be scheduled as "first/last weekday" but labeled "Monthly on the last day". Mirror `hasNthWeekdayAnchor()`'s precedence.

**3. nth-weekday branch trusts unvalidated values.** `get-task-repeat-info-text.util.ts:145`: any non-`undefined` `monthlyWeekOfMonth`/`monthlyWeekday` passes. The recurrence code uses `hasNthWeekdayAnchor()` for range/validity checks; the label code should do the same, otherwise `ordinalKey` stays `''` and the user sees *"Monthly on the  Saturday"* (double space) or a raw key. At minimum, fall back to the existing `MONTHLY_CURRENT_DATE` branch when the anchor is invalid.

**4. Quick-setting label uses keys only present in en/de.** `build-repeat-quick-setting-options.ts:30` now reads `ORD_*_NTH`, but those keys exist only in `en.json` and `de.json`. Locales that already translate `Q_MONTHLY_NTH_WEEKDAY` will render a mixed-language label (localized sentence + English ordinal fallback). Either ship the new keys to all locales or keep using `ORD_*` for the quick-setting menu and only use `*_NTH` inside the info-panel sentence.


### Smaller items

- **Optional `translateService` is a silent footgun.** `getTaskRepeatInfoText`'s new parameter is optional; if any future caller forgets it, `ordinalStr` resolves to `''`. All current callers pass it — make it required.
- **Weekday-name lookup anchored on `new Date()`.** `get-task-repeat-info-text.util.ts:155` uses today's date plus an offset to derive a localized weekday name. Works, but non-deterministic across DST and unit-testable only by stubbing `Date`. A fixed reference like `new Date(Date.UTC(2024, 0, 7))` (known Sunday) is simpler.
- **`ordinalKey = ''` is a reachable fallback.** Better to early-return the legacy numeric-day branch than to call `translate.instant('')`.
- **Test coverage of the new code paths is shallow.** The new specs assert `ordinalStr: ''` / `weekdayStr: ''`, which skips the actual translate lookup and the locale-dependent `toLocaleDateString`. A test with a stub `TranslateService` and a fixed locale would lock the behavior in. Also no test for the `DateTimeFormatService._localeSig` fallback path.
- **PR description checkbox for `npm run checkFile` is unchecked.** Worth confirming CI lint/format is green on these files.

### Nits

- `ORD_*_NTH` values in `en.json` are identical to `ORD_*` (intentional — English has no inflection). A short comment in `en.json` (or a doc note) explaining the suffix means "dative form, used inside sentences" would prevent future translator confusion.

Happy to look at a follow-up commit. Nice catch on the grammar issue — easy to miss in English.

hajiboy95:
@johannesjo adjusted according to your comment

johannesjo:
Re-checked the current head `70580715`. The date-time locale fallback and monthly nth-weekday precedence/validation concerns from the previous pass look addressed, but a few follow-ups still look necessary:

1. The quick-setting menu still uses the nominative ordinal keys. `build-repeat-quick-setting-options.ts` keeps `ORDINAL_KEYS` on `ORD_FIRST` / `ORD_SECOND` / etc., and the spec explicitly asserts that `_NTH` is not used. With German, `Q_MONTHLY_NTH_WEEKDAY` is `Jeden Monat am {{ordinalStr}} {{weekdayStr}}`, so the current code still produces labels like `Jeden Monat am erste Dienstag` instead of `Jeden Monat am ersten Dienstag`. Since the `_NTH` keys now exist in the locale files, please use `ORD_*_NTH` for this sentence and update the spec accordingly.

2. The new monthly info-panel labels are English in most non-English locale files, while `weekdayStr` is formatted with the active locale. For example, French has `MONTHLY_NTH_WEEKDAY: "Monthly on the {{ordinalStr}} {{weekdayStr}}"` and `ORD_FIRST_NTH: "first"`, so the UI can render mixed-language labels such as `Monthly on the first lundi`. Please provide real translations for the new monthly sentence keys and `_NTH` ordinals, or otherwise make fallback formatting consistent with the fallback language.

3. The non-English locale files contain a lot of unrelated churn from this PR (27 files, roughly 14k additions / 10k deletions), including unrelated English placeholders such as German `RESTORE_FILE_BACKUP_MOBILE` and Chinese `SYNC_STATE.*`. This makes the bug fix hard to review and stores English fallback strings in locale files. Please trim this PR to the repeat-related translation changes, or use the documented translation workflow with actual translations.

hajiboy95:
@johannesjo here we go. Sorry for that translation-file-chaos. I just used a translation script provided somewhere in the repo. I thought that's the way to go^^. Just deleted that commit in between. Also adressed the first 2 points. Should be okay now i guess

johannesjo:

PR #8013, opened 2026-06-04
fix(metric): use worklog totals for activity heatmap

### LINKED ISSUE #7993: 💥 Wrong time calculation on the "today" heatmap when using subtasks
### Steps to Reproduce

1. Create a task
2. Create more than one subtask
3. Work on one subtask (or more, but not all subtasks) and validate it as done

Output : when looking at the "Quick History" in the Today tab, the total worked time is correct. However, when opening "Metrics" in the Today tab, the total time worked on that day equals [total task time] + [each subtask time], resulting in a doubled time for this task and subtasks. 

### Url
file:///Applications/Super%20Productivity.app/Contents/Resources/app.asar/.tmp/angular-dist/browser/index.html#/config

### Meta Info
META: SP16.9.4 __ Electron – fr – MacIntel – fr – UA:Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) superProductivity/16.9.4 Chrome/138.0.7204.251 Electron/37.10.3 Safari/537.36

### Actions Before Error
```

```

puneetdixit200:
## Summary
- use WorklogService as the single source for activity heatmap totals, including the Today metrics view
- remove the Today-only raw task aggregation path that could diverge from Quick History totals
- add a parent/subtask regression spec so heatmap time uses the worklog day total instead of summing log entries

Fixes #7993

## Verification
- npm run checkFile src/app/features/metric/activity-heatmap/activity-heatmap.component.ts
- npm run checkFile src/app/features/metric/activity-heatmap/activity-heatmap.component.spec.ts
- npm run test:file -- src/app/features/metric/activity-heatmap/activity-heatmap.component.spec.ts

### DIFF
--- a/src/app/features/calendar-integration/calendar-integration.service.spec.ts
+++ b/src/app/features/calendar-integration/calendar-integration.service.spec.ts
@@ -3,6 +3,7 @@ import {
   fakeAsync,
   tick,
   discardPeriodicTasks,
+  flush,
   flushMicrotasks,
 } from '@angular/core/testing';
 import {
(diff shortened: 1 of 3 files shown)


johannesjo:
Thank you very much! <3

Issue #7990, opened 2026-06-03
💡 Planned for date picker and Recurring tasks date picker should have the same style


beerkumquatpome:
### Problem Statement

The recurring tasks date picker:
- Today doesn't have a sunny icon

The planned for date picker:
- When hovering over a date, it doesn't show a colored circle
- A selected date has a circle outline (Accent color)

### Possible Solution

The recurring tasks date picker:
- Change Today to a sunny icon

The planned for date picker:
- Show a colored circle when hovering over a date
- Remove the circle outline on selected date
I think it looks better without the outline

### Additional context

The deadline date picker should also be updated

hajiboy95:
@beerkumquatpome Before we dive into this, I just want to make sure I'm fully aligned on the scope. We are talking specifically about the start date picker for a recurring task, right?

Looking at the current calendar (*see screenshot below*), the UI feels a bit disconnected from the rest of the app.

<img width="537" height="386" alt="Image" src="https://github.com/user-attachments/assets/e99c0c43-cdff-4798-a9e0-940aed1bccc9" />

Instead of just patching this specific component by adding the sun icon, **why not replace it entirely with the standard calendar view we already use for regular tasks?** This would immediately solve the visual issues and make the app feel much more cohesive.

If we go with the standard task calendar, we have two options for how to handle the time selection:

1. **Use the standard calendar but disable its time selection:** We just use it for the date and keep the existing separate time input fields below it.
2. **Use the full standard calendar (with time):** We remove the separate time fields below and handle both date and time inside the single standard widget. 

Thinking about @johannesjo's goal of keeping the app lean, I'm actually leaning toward Option 2. It reduces the total number of different input fields cluttering the UI, while still keeping the flexibility to specify a starting time (or just stick to a date with no time).

**One more thought on consistency:**
While we are touching this dialog, what if we also added the standard icons for tags and  time estimates (the same ones used in the task side-panel)? Bringing those over would make this whole recurring task menu feel much more unified with the day-to-day task UI. I would Refrain from time estimate to open the same popup, since it would just add additional friction and we only need to specify here one number (the estimate and not the actual time), but a symbol indication would be nice I think.

Let me know what you think!

hajiboy95:
@beerkumquatpome can you check the current PR? I globalized the calendar and therefore everywhere it should look now more or less the same

I got some Ideas to make the recurring task popup visually closer to the overall design language of the app but i think that should be something for another PR: 

[...]

Issue #7768, opened 2026-05-24
🚨 Recurring tasks: Bugs related to Today


beerkumquatpome:
### Version Used

v18.7.0

### Current Behavior

Today = May 24

### Bug 1: Changing the start date of a recurring task config to Today doesn't move the live instance to Today

1. Create a task
2. Add a recurring task config
Start Date: May 30
Recurring Config: Everyday
3. Press Save
4. Change the start date of the recurring task config to May 24 (Today)
Current behavior:
Non-transparent instances are created on May 25 to May 30, but the live instance stays on May 30
Expected behavior:
The live instance should move to May 24 (Today)

### Bug 2: Deleting the live instance then changing the start date of a recurring task config to Today doesn't create a live instance on Today

1. Create a task
2. Add a recurring task config
Start Date: May 30
Recurring Config: Everyday
3. Go to Planner
4. Delete the live instance on May 30
6. Click on the non-transparent instance on May 31
7. Change the start date of the recurring task config to May 24 (Today)
Current behavior:
Non-transparent instances are created on May 24 to May 30, but the instance on May 24 (Today) is a transparent instance
Expected behavior:
The instance on May 24 should be a non-transparent instance (live instance)

### Bug 3: After deleting a live instance, can't add the instance back

1. Create a task
2. Add a recurring task config
Start Date: May 30
Recurring Config: Everyday
3. Go to Planner
4. Delete the live instance on May 30
5. Click on the non-transparent instance on May 31
6. Press save (The start date is the same: May 30)
Current behavior:
Nothing happens
Expected behavior:
It should create a live instance on May 30

### Bug 4: It shouldn't be possible to change the start date of a recurring task config to a date earlier than Today

I can change the start date to May 23 (Today is May 24)

### Expected Behavior

-

### Steps to Reproduce

-

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

johannesjo:
Thank you very much! <3

johannesjo:
@beerkumquatpome can you please open up a seperate issue for this?

PR #8017, opened 2026-06-04
Update recurrent task calendar and general calendar design

hajiboy95:
## Problem

#7990 

## Solution

Globalized the Calendar + slightly redesigned aesthetics (e.g. now button for "Plan" stays gray/disabled until a date is actually selected) 

Also
monthYearLabel flipped from 'short' to 'long' in src/main.ts:223 — visible UX change, worth a note in the PR description.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x My commit messages follow the Angular format (`type(scope): description`)


hajiboy95:
@johannesjo if you find some time, i guess functionally it's more or less final and ready for a check.

johannesjo:
Thanks for the update. I think this still needs changes before merge.

1. `currentLocale()` now falls back to bare UI language codes before `DEFAULT_LOCALE`:
https://github.com/super-productivity/super-productivity/blob/bafd4b8a12e13ed195986caf08ac4ff7f994b719/src/app/core/date-time-format/date-time-format.service.ts#L21-L24

That changes the default English date/time behavior from the app’s previous `en-GB` default to `Intl('en')`, which resolves like U.S. formatting in this runtime (`12/31/2000`, `1:00 PM`). Ambiguous typed dates can now parse/format differently for users who never chose a date-time locale.

The same fallback also breaks some supported UI languages. For example `fa` is selectable as a UI language, but `Intl('fa')` uses the Persian calendar/native digits, while `_testFormats()` derives parser tokens by replacing ASCII `31`, `12`, and `2000`:
https://github.com/super-productivity/super-productivity/blob/bafd4b8a12e13ed195986caf08ac4ff7f994b719/src/app/core/date-time-format/date-time-format.service.ts#L40-L44

That leaves `dateFormat().raw` without `yyyy/MM/dd`-style tokens, so `parseStringToDate()` cannot parse date input reliably. Please map UI language fallback to parse-safe date-time locales, preserve the previous `en -> en-GB` behavior, or derive the format via a safer `formatToParts`/Gregorian+Latin-numbering approach.

2. Clearing an explicit date-time locale can leave the Material `DateAdapter` stuck on the old locale:
https://github.com/super-productivity/super-productivity/blob/bafd4b8a12e13ed195986caf08ac4ff7f994b719/src/app/core/date-time-format/date-time-format.service.ts#L67-L68

`currentLocale()` now falls back to `lng/currentLang`, but the adapter effect only updates when `dateTimeLocale` is truthy. If a user has `lng = de`, sets `dateTimeLocale = en-us`, then clears it back to auto, `currentLocale()` becomes `de` while the adapter can remain `en-us`. Calendar chrome such as `monthYearLabel` can diverge from parsing/formatting.

3. Repeat quick-setting labels are still cached only by `startDate`:
https://github.com/super-productivity/super-productivity/blob/bafd4b8a12e13ed195986caf08ac4ff7f994b719/src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.ts#L334-L340

`buildOptions()` reads `currentLocale()`, but once `cachedOptions` exists, a locale change with the same `startDate` reuses stale labels. Include the effective locale in the cache key or otherwise invalidate when locale changes.

4. `plannedStartDateStr()` and `openScheduleDialog()` now normalize stored repeat clock strings through a real `Date`:
https://github.com/super-productivity/super-productivity/blob/bafd4b8a12e13ed195986caf08ac4ff7f994b719/src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.ts#L100-L103
https://github.com/super-productivity/super-productivity/blob/bafd4b8a12e13ed195986caf08ac4ff7f994b719/src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.ts#L139-L143

On DST gap days this changes the stored clock time. In `Europe/Berlin`, `startDate: '2026-03-29'` with `startTime: '02:30'` normalizes to `03:30`. The button label shows `03:30`, and opening the schedule dialog initializes from that normalized timestamp, so saving can silently change the recurrence from `02:30` to `03:30`. Please keep displaying/preserving the stored clock string unless the user explicitly changes it.

hajiboy95:
@johannesjo adressed and resolved merge conflicts with updated master

johannesjo:
This should keep Material's disabled guards for the new stronger hover state. The schedule dialog passes `minDate = new Date()`, and Material calendar cells are still hoverable while marked with `.mat-calendar-body-disabled`, so disabled past dates can now get the same 35% primary hover treatment and look selectable. The previous/next buttons also use `disabledInteractive`, so disabled nav arrows can match `:hover` too.

Could we scope these like Material does?

```scss
.mat-calendar-body-cell:not(.mat-calendar-body-disabled):hover
  > .mat-calendar-body-cell-content:not(.mat-calendar-body-selected) {
  ...
}

.mat-calendar-previous-button:not(.mat-mdc-button-disabled):hover,
.mat-calendar-next-button:not(.mat-mdc-button-disabled):hover {
  ...
}
```

Small related cleanup while touching this: the new pill radius can use the existing `var(--radius-pill)` instead of hardcoded `999px`.

johannesjo:
https://github.com/super-productivity/super-productivity/issues/7990#issuecomment-4631456541

hajiboy95:
@johannesjo adressed your concernes. ready for next check

johannesjo:
Thanks for the updates. I rechecked the current head, and some of the earlier concerns are addressed: the repeat quick-setting cache now includes locale, the recurring start label preserves the stored clock string better, schedule warnings are short-circuited for `isSelectDueOnly`, and the hardcoded `999px` radius is gone.

I still think this needs changes before merge:

1. **Locale fallback still reopens the #8023 behavior.** `DateTimeFormatService` still prefers `localization?.lng` before `getBrowserCultureLang()` and still globally maps `en -> en-gb` / `fa` / `ar` through `currentLocale()` and `setDateAdapterLocale()`:
   - `src/app/core/date-time-format/date-time-format.service.ts:17`
   - `src/app/core/date-time-format/date-time-format.service.ts:83`
   - `src/app/core/date-time-format/date-time-format.service.ts:98`
   - `src/app/core/date-time-format/date-time-format.service.ts:109`

   Please restore master’s browser-region-first fallback for this PR. If `ar`/`fa` parse-safety is needed, let’s handle that separately with focused tests.

2. **Deadline quick-access tooltips now use schedule labels.** The shared picker hard-codes `T.F.TASK.D_SCHEDULE_TASK.QA_*`:
   - `src/app/ui/datetime-picker/datetime-picker.component.html:7`

   The deadline dialog previously used `T.F.TASK.D_DEADLINE.QA_*`, so deadline quick actions now get the wrong wording. Please make the quick-access labels configurable or otherwise preserve the deadline labels.

3. **The custom calendar hover still affects disabled nav buttons.** The selector still applies to `.mat-calendar-previous-button:hover` / `.mat-calendar-next-button:hover` without excluding disabled buttons:
   - `src/app/ui/datetime-picker/datetime-picker.component.scss:84`

   This was the disabled-hover concern from earlier. Either remove the custom hover and keep Material defaults, or scope it with the disabled guard.

4. **The custom header aria labels do not match the actions.** In month view, `monthButtonLabel` announces `switchToMultiYearViewLabel`, but `monthLabelClicked()` opens the year view. The year button has a similar mismatch:
   - `src/app/ui/datetime-picker/datetime-picker-header.component.ts:109`
   - `src/app/ui/datetime-picker/datetime-picker-header.component.ts:130`
   - `src/app/ui/datetime-picker/datetime-picker-header.component.ts:134`

   Please align the labels with the actual view transitions, or consider reverting to the stock Material header.

5. **Recurring start-date button styling has RTL / Material override issues.** It uses physical LTR spacing and a local Material-internal override:
   - `src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.scss:93`
   - `src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.scss:99`

   Please use logical spacing/alignment and avoid the local `::ng-deep .mdc-button__label` override.

One smaller repo-rule cleanup: `openScheduleDialog()` introduces `const dummyTask: any` in production code (`dialog-edit-task-repeat-cfg.component.ts:124`). Please replace that with a proper typed object rather than `any`.

hajiboy95:
The tests needed some rework. Fixed them according to new modal.

hajiboy95:
@beerkumquatpome 
1. there is a hover effect, it's just very very subtle (i don't like it either, i can barely see it but @johannesjo said we should stick to the default for now)
2. the black outline is the current date. the blue one is the selected. the pink one is the current selection you can control with your keyboard. press left and right and you'll see it moves. 
3. look 2. 
4. Good catch. i fixed this and just pushed it 
5. overflow tisk means that in some languages the name can be quite long. If we always write the full name, in some lnaugages it might cause the name to be too long and therefore look ugly since it's capped somewhere or even worse is writes over the other name. Short names therefore are already testes and the grid has been designed to fit in the short names

hajiboy95:
Update on the hover effect. @johannesjo said 
>  if it's genuinely too weak, let's nudge it via a shared calendar hover token so every picker benefits, rather than the pill-shaped primary-tinted circle.


So that's what i did. I went for a.share hover token. Hope it's okay now as it is. Push is incoming...

hajiboy95:
> > in some languages the name can be quite long
> 
> But the full month names are used for English only

Not sure what you're referring to. In german i see the short names too.
grafik

we shouldn't introduce different behaviours for different languages though

hajiboy95:
> 1. Can we capitalize only the first letter?
>        JAN -> Jan
> 
>     2. If we really want to keep the primary color outline: When hovering over a date, it should also have the outline to make it less confusing

1. i pass on to @johannesjo 
2. The outline fullfills the purpose to indicate which one is selected by the keyboard. One could theoretically also move the selection to the one we hover over, like this the hover and the keyboard movement would always align, let me try it for a second

hajiboy95:
@beerkumquatpome okay i updated this now. Wasn't that easy since you don't want to have then all of a sudden two highlighted circle (one for keyboard position and one for mouse) 

I went that way. In the beginning you got a keyboard positon, can move with keyboard. when you move with mouse, the highlighted circle moves with your mouse. Now your mouse stops and you continue with keyboard, then the circle moves with the keyboard as long as you don't move the mouse. 


I think that's a quite nive behaviour. At least to me it seems intuitive

hajiboy95:
> The new hover effect doesn't have primary color. The hover effect for the month and the year has it though (June 2026). Why is that?

grafik

For me it doesn't. I see
- Hover effect -> Gray with red border
- Select -> Primary color

hajiboy95:
> June 2026 (The one that has an arrow next to it)

very sharp eyes!!! Fixed. Push incoming

hajiboy95:
> Can the hover effect for other buttons also has primary color? It looks much better

grafik

Done, push incoming

hajiboy95:
@beerkumquatpome 2. then we wouldn't see the outline when we are exactly on the current selected task. it should always be visible where we are currently

1. Fixed. Also mouse hides now when we move via keyboard keys. like this we don't get an extra hover effect where the mouse stays. Now hovering with mouse and selecting with keyboard have same color (slighly less % primary color) 

@johannesjo we are ready for a next check :D

hajiboy95:
@beerkumquatpome reduced to 80% (less and it's not really that "selected" feeling) and fixed the bug. Thank you very much for iterating on this so much

hajiboy95:
@beerkumquatpome thanks for staying persistant. I think I've found a way which pleases both of us. Indeed it was inconsistent since month and year had that border and todays date not. I gave all of them now the highlight color (when not selected). if we select them they get the normal color :D. Push is on it's way

johannesjo:
Thanks for iterating on this. UX-wise I think this is moving in the right direction: keyboard focus and mouse hover now feel much more aligned, and using accent for current day/month/year avoids the earlier "dark outline + primary fill + accent outline" confusion.

A few remaining issues before merge:

1. `openScheduleDialog()` still assumes stored repeat `startTime` is valid:
   `src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.ts`

   It calls `getDateTimeFromClockString()` directly when `currentCfg.startDate && currentCfg.startTime`. Legacy/synced corrupt values like `INVALID_CLOCK_STRING` still throw here, so users can crash the repeat edit dialog while trying to open the new scheduler and fix the value. Please guard with `isValidSplitTime()` and fall back to date-only/no-time state.

2. Date-only repeat scheduling can persist a reminder option.

   In select-only mode, `DialogScheduleTaskComponent` defaults `selectedReminderCfgId` to `AtStart`, and the repeat dialog writes `remindAt: result.remindOption` even when `result.time` is `null`. This probably does not schedule incorrectly because effects also require `startTime`, but it creates stale/misleading config state and can show the alarm badge for a date-only repeat. Please only persist `remindAt` when a valid time is present.

3. Disabled calendar cells can still get active/focus styling.

   The hover handler focuses `.mat-calendar-body-cell` directly, and the global `:focus` CSS does not exclude `.mat-calendar-body-disabled`. Angular Material uses `aria-disabled` plus a class here, not native `disabled`, so disabled cells can still receive the same accent outline / primary background treatment. Please add disabled guards in both the handler and focus CSS.

hajiboy95:
@johannesjo adressed

hajiboy95:
@beerkumquatpome that's not a bug. If I think about it it's actually something I'd wish for. Imagine it's december. You'd like to go to the next year, no? To guarantee that we loop only within the current moths will implement a lot of extra if clauses. I like the current behaviour and wouldn't say it's a bug^^. We have 12 entries and you start at your month. you can't go back so if you go all the way to the end (or press a few times "down" i'd guess the user would like to go further.

@johannesjo if you're okay with it too i think we're okay currently. 

Only think to agree on whether we should remove the capitalization. I think it's okay.

beerkumquatpome:
If it's not a bug then we need to add back the prev/next buttons for the month picker

hajiboy95:
@beerkumquatpome not necessarily. You can only get there by using the keyboard. So if you use the keyboard you'll get back too.^^ @johannesjo what's your take on this? Should we add arrows?

beerkumquatpome:
If we can go to the next year by using the keyboard, we should also be able to do it by using the mouse (prev/next arrows)

hajiboy95:
> If we can go to the next year by using the keyboard, we should also be able to do it by using the mouse (prev/next arrows)

Nope. In general we want to motivate people to use the keyboard more. A good example is the "!" Short within creating a new taks. It allows you to add a deadline to a task. You can't add a deadline to a new task by simply using the mouse. Same hold true here. Here it's even better since you have actually the option to switch year by using the mouse (just click on year in the menu before). For me it's okay and it's actually better intuitive wise. Like this we keep with small tweaks people motivated to use the keyboard by hiding some intuitive Easter eggs for the keyboard :D

hajiboy95:
> Bug?
> 
>     1. Hover your mouse cursor over a date (Something like Day 28)
> 
>     2. Keep pressing the left arrow key on your keyboard to go to the next month
> 
>     3. The hover effect is not on Day 1. It's near the area round Day 28
>        The month/year picker also have the same bug
>        Expected behavior: Only go to the area around Day 28 when a mouse movement is detected

Thank you very much. it's fixed. :) Push incoming

johannesjo:
Thanks for taking this on — consolidating three near-identical calendar/time UIs into one shared `datetime-picker` is a real win, and it's a net ~880-line deletion. I reviewed the diff against `master` (and the actual behavior of the three dialogs). A few things to address before merge, plus some KISS/YAGNI thoughts on the new component's surface area.

### Should fix

1. **Deadline default-time can roll to the next day at 23:00–23:59.** In `datetime-picker.component.ts` (`onTimeFocus`, lines ~227 and ~233) the auto-fill is `getClockStringFromHours(new Date().getHours() + 1)` with no `% 24`. The old *deadline* dialog used `(getHours() + 1) % 24`. At 23:xx this produces `'24:00'`, which `<input type="time">` can't display (renders blank) and which `getDateTimeFromClockString` turns into **next-day 00:00** via `setHours(24)`. Adding `% 24` here fixes it for both consumers (the schedule dialog had the same latent gap).

2. **Leftover debug log.** `datetime-picker.component.ts:217` still has `Log.log('onTimeFocus')` (carried over from the old dialog) — fires on every time-field focus. Please remove it.

### Worth confirming (intentional?)

3. **Past dates are now disabled for deadlines.** `dialog-deadline.component.ts` adds `minDate = new Date()`; the old deadline calendar had no `minDate`, so you could record an already-passed deadline. Is locking that out intended? If not, drop `minDate` for the deadline case.

4. **Two changes apply app-wide, not just to these dialogs:**
   - `_overwrite-material.scss` adds unscoped `.mat-calendar-body-cell:hover/:focus` outline + background rules (with `\!important`) → every `mat-calendar` in the app gets the new hover/focus look.
   - `main.ts` changes `monthYearLabel` from `month: 'short'` to `'long'` → every datepicker label app-wide.

   Both may be desired, but if the intent was to restyle only the new picker, consider scoping them to the component.

5. **Minor:** clearing the time in the *schedule* dialog now also resets the reminder to "Do not remind" (shared `onTimeClear` emits `reminderChanged`). The old schedule dialog left it intact, so re-adding a time no longer restores the previous reminder choice. Arguably fine (it matches the deadline behavior), just flagging the change.

### KISS / YAGNI

6. **The 232-line custom calendar header re-implements a lot of `MatCalendarHeader`.** The genuinely new part is the two-button month/year labels; but `previousClicked`/`nextClicked`/`previousEnabled`/`nextEnabled`/`_isSameView` largely duplicate Material's own logic. Worth weighing the maintenance cost (it forks framework code and depends on Material-internal class names). One concrete thing to double-check: the multi-year branch of `_isSameView` uses `Math.floor(year / 24)` (pages anchored at year 0), which won't necessarily line up with the page grid `mat-multi-year-view` actually renders when `minDate`/`maxDate` is set — so prev/next could be enabled/disabled on the wrong boundary in the year-picker view.

7. **A couple of inputs are dead config.** `showTime` is `[showTime]="true"` at both call sites, and `showQuickAccess` is effectively always true (the only producer passes `true`). Since there's no date-only/no-quick-access caller today, these could be dropped until a real need appears.

8. **New keyboard-nav + hide-cursor behavior** (`isKeyboardNavigating`, `.sp-hide-cursor`, the `mouseover`+`mousemove` hover-sync) doesn't exist on `master` — it's a new feature riding in a refactor PR. Happy to see it, just worth calling out separately so it gets its own attention. (Small thing: it's bound to both `mouseover` and `mousemove`, which is redundant — `mousemove` alone covers it.)

### Nits (optional)
- The `imports` array lists both the umbrella `*Module` and the individual standalone class it re-exports (e.g. `MatButtonModule` + `MatIconButton`) for ~11 entries; `CommonModule` and `MatDatepickerModule` look unused (template uses `@if`/`@for` and `<mat-calendar>`).
- Reminder dropdown label is hardcoded to `T.F.TASK.D_SCHEDULE_TASK.REMIND_AT` even though `timeLabel` is parameterized — consider making it an input too so the deadline dialog can use its own key.
- New outputs use `@Output()/EventEmitter` while inputs use the `input()` signal API — `output()` would be consistent with the rest of the file/codebase.
- `onYearSelected(date)` doesn't use its parameter.

Overall the direction is good — most of the above is small. The only real must-fixes are the `% 24` guard and the stray `Log.log`. 👍

hajiboy95:
@johannesjo thanks again for checking.


1+2 fixed
Quick Heads up on this: We got a method

```typescript
/*given a value of 8..5 returns '8:30'*/
export const getClockStringFromHours = (hours: number): string => {
  const minutes = Math.round((hours % 1) * 60);

  const parsed =
    minutes === 60
      ? Math.ceil(hours) + ':00'
      : Math.floor(hours) + ':' + ('00' + minutes).slice(-2);
  return parsed.trim();
};
```
Here we don't do that %24. For now i kept it like this. Just wonder, was this intentional? Instead of fixing that %24 every time, we could fix it one time here. However since I don't know whether this was intentional or not, i kept it for now 

3-5 intentional 

6 onwards: adressed.

Open question: 
Should we in the grid now capitalize the Months or should we leave them just as they currently are. I think it's an important matter for @beerkumquatpome

johannesjo:
Thanks for the latest update. I ran another pass with a few focused reviews. I’d hold off merging until these are addressed:

1. The month picker still has keyboard-only year navigation. In `datetime-picker-header.component.ts`, the prev/next buttons are hidden for `currentView === 'year'`, but that is the month-picker view and the handlers still implement year paging there. So keyboard users can move to previous/next years while mouse users cannot. I think this should either add the arrows back for the month picker, or remove/clamp the keyboard-only year crossing.

2. The custom multi-year boundary logic does not match Angular Material when `minDate`/`maxDate` is present. `_isSameView()` buckets years by 24-year ranges from year 0, while Material anchors multi-year pages from the min/max constraints. This can enable prev/next on a boundary page and then clamp/jump unexpectedly.

3. The repeat-config start-date floor regressed. On `master`, existing repeat configs with a past `startDate` kept that existing date as the selectable floor; this PR routes through `DialogScheduleTaskComponent`, whose `minDate` is always today. Editing an older repeat config can no longer select dates between the original past start date and today.

4. The recurring E2E helper now accepts a full `DD/MM/YYYY`, but only parses the day and clicks the first visible calendar cell with that day. Month/year are ignored, so tests such as `setRecurStartDate(page, '15/06/2026')` can pass while selecting the wrong date.

I verified the new picker and repeat-dialog unit specs locally:

- `npm run test:file src/app/ui/datetime-picker/datetime-picker.component.spec.ts`
- `npm run test:file src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.spec.ts`

Both passed. The previous default-time rollover and stray debug-log issues look fixed.

hajiboy95:
@johannesjo thanks for the feedback, gonna adress them soon. However, could you answer to the 2 questions i posed in my last comment?

johannesjo:
I have to bring up the custom header again, sorry.

First: I want to acknowledge the amount of heart you put into this. The custom header clearly wasn’t a quick throwaway change. You iterated on the visuals, hover/focus behavior, keyboard navigation, month/year picking, RTL/a11y feedback, and a lot of small details. I appreciate that effort.

That said, I think this is exactly the point where we should step back. The custom header is starting to get out of hand relative to the benefit it adds.

The main value of this PR is the shared `datetime-picker`: planned dates, deadlines, and recurring start dates move toward one consistent calendar/time UI. That is worth keeping.

But the custom header adds a lot of maintenance surface for a fairly small UX gain: separate Month/Year buttons and a nicer period label. The PR is now 56 commits, 30 files, `+1603/-903`. The new picker area alone is about `+1024` lines, and the custom header itself is ~235 lines. There have also been many follow-up commits and review rounds specifically around header/picker behavior: hover/focus styling, keyboard vs mouse navigation, hidden cursor behavior, month/year picker edge cases, a11y labels, RTL arrows, and now Material min/max boundary behavior.

That pattern worries me. Angular Material’s stock header already gives us correct prev/next navigation, translated a11y labels, RTL behavior, min/max boundary handling, and maintained keyboard behavior. Reimplementing those pieces means we now own all those edge cases.

So my preference would be: keep the shared `datetime-picker`, but move the header back closer to the stock Material calendar header unless there is a very strong reason not to. This would make the PR smaller, easier to review, and safer to maintain. The work you did is not wasted; it clarified what polish matters, but I think the final implementation should be simpler than the current custom header.

johannesjo:
Sorry, yes - answering those directly:

1. I would not put the `% 24` into `getClockStringFromHours()` itself. That helper is currently a general numeric-hours formatter, and some call sites use it for rendered schedule ranges/end times where wrapping could hide useful information or silently change meaning. For values that are specifically meant for `<input type="time">` / clock-of-day input, normalize at the call site with `% 24`, or add a small separate helper for clock-of-day hours if we want to avoid repeating it. So the deadline/schedule default-time case should stay normalized before calling the helper.

2. For month names, I would leave the casing as provided by the locale/Material rather than force-capitalizing in the grid. Some locales intentionally use lowercase month names, and globally capitalizing can make those look wrong. If @beerkumquatpome has a specific design reason for a different casing we can revisit it, but my default would be locale-native casing.

hajiboy95:
@johannesjo before I undo the calendar behaviour, let me clarify a few things: 
There have also been many follow-up commits and review rounds specifically around header/picker behavior: hover/focus styling, keyboard vs mouse navigation, hidden cursor behavior, month/year picker edge cases, a11y labels, RTL arrows, and now Material min/max boundary behavior.
-> ALL those things are problems with the current calendar too. We could just ignore all of them and keep them as they are, but we'd still get those problems, that has NOTHING to do with the other header. I think in all those commits that got a bit out of track. But just in my opinion the calendar experience got upgraded A LOT by all those small adjustments. Special thanks to @beerkumquatpome at this stage once more. He really insisted to get things right and quite neat.

 the custom header itself is ~235 lines
 -> tbh i don't think that's a lot.
Also the new locker behavior is what has a lot of lines (1000+) but also here if you ask me it's mostly visuals, nothing really functional, so not really a source which can cause buffs. Me personally I wouldn't worry too much on that side. 

I'd stick to the custom header if you'd ask me. all it does is just show the second page of the normal calendar and the year picker just skips the second page. that's nothing too far off if you ask me. The rest we did are just all little adjusts for the calendar in general, nothing specific about custom header. 


I can see that a lot of commits may seem like "oh there are a lot of edge cases for bugs" but most of the things were adjustments we could also do with the current calendar too. I don't think it actually adds a lot of danger to the calendar overall. Therefore since most of the problems don't lie within the header themselves, i'd stick to it. 

Towards the 50+ commit history. I tried to keep commits small. one problem one commit. Ofc I could tackle several problems at once and therfore bundle them in 1 commit such that it seems less, but that's not my style...


Just my 2cents

johannesjo:
Thanks for clarifying. I agree that the commit count itself is not the issue, and I also agree that several of the hover/focus/a11y details are general calendar UX concerns, not only custom header bugs.

The point I am trying to make is slightly different: once this PR reimplements calendar header/navigation behavior, we become responsible for matching Material behavior across keyboard, mouse, RTL, a11y labels, min/max dates, locale behavior, and future Material changes. The recent review rounds show that this surface is bigger than it first appears.

For this PR, I would like to keep the main value: the shared `datetime-picker` and the recurring-task integration. That is a solid improvement.

But I do not want this refactor PR to also become a general calendar redesign. So my preferred path is:

1. keep the shared picker;
2. move the header/navigation closer to stock Material, or remove the custom header unless there is a strong product reason;
3. keep only low-risk, scoped visual polish;
4. put broader calendar UX polish into a follow-up issue/PR where we can compare Google Calendar / Todoist / TickTick etc. with screenshots and decide deliberately.

Also, I do not think core picker functionality should be keyboard-only. Keyboard can be faster, but if the month picker can cross years with keyboard, mouse users need an equivalent path too, or the keyboard behavior should be clamped.

Separately, CI is currently failing because `e2e/utils/recurring-task-helpers.ts` declares duplicate helpers, so that needs fixing either way before merge.

hajiboy95:
@johannesjo I see you and i understand you. I now got your standpoint that you wanted the scope of this PR to be the merging of the 3 Calendars (which has been done). Even though previous tests were all successfull, i now reverted the changes

To keep the Scope thight I reverted High-Risk Items:
   * Mouse-hover date synchronization: Removed the logic that forced the calendar's active focus to follow the mouse cursor. This relied on brittle
     DOM selection of internal Material classes that could break in future updates.
   * Custom Header Component: Reverted to the standard Material calendar navigation header. This ensures the component keeps its robust
     accessibility (a11y) features, localization handling, and leap-year logic without custom maintenance.
   * Brittle Implementation Details: Cleaned up several ::ng-deep overrides and manual DOM calculations that targeted internal Material component
     structures.


I kept the visual adjustments (hover color) since they were low risk and actual quite neat. I hope that we are good to go now

johannesjo:
Two issues still look worth addressing after the custom calendar header removal:

1. `src/app/ui/datetime-picker/datetime-picker.component.scss` still hides `.mat-calendar-arrow` and keeps stale `.sp-month-button`, `.sp-year-button`, and `.sp-drop-icon` rules. Now that `<mat-calendar>` uses Material’s default header again, hiding `.mat-calendar-arrow` removes the built-in period-button affordance/state indicator, and the `.sp-*` rules target elements that no longer exist. Please remove these overrides/stale rules.

2. The deleted custom header provided translated switch labels via `TranslateService`, but the default Material header falls back to `MatDatepickerIntl` labels. I don’t see a `MatDatepickerIntl` provider in this tree, so aria labels such as the period-switch labels remain English despite the calendar globalization work. Please add/localize a `MatDatepickerIntl` provider, or otherwise preserve translated calendar header labels.

hajiboy95:
@johannesjo adressed

johannesjo:
Reviewed the PR end-to-end with independent verification on each finding. The refactor's direction is solid — one shared `<datetime-picker>` replacing three bespoke ones, real cleanup. A few things I'd want resolved or acknowledged before merge, ordered by impact.

### Correctness

1. **`DateTimePickerComponent` mutates a global singleton from a leaf effect.** `datetime-picker.component.ts:84-87` calls `this._dateAdapter.setLocale(locale)` inside an `effect()`. `DateAdapter` is `providedIn: 'root'`, so every instance that mounts overwrites global locale; nothing restores it on destroy. Last-mounted picker's locale wins for every other `DateAdapter` consumer (formatting, export, other date inputs). Locale should be set once at bootstrap via `MAT_DATE_LOCALE` / a root-level effect, not from this component.

2. **`_syncActiveDateEffect` discards user navigation.** `datetime-picker.component.ts:150-156`:
   ```ts
   private _syncActiveDateEffect = effect(() => {
     const date = this.selectedDate();
     const cal = this.calendar();
     if (cal) { cal.activeDate = new Date(date || new Date()); }
   });
   ```
   No guard — if a host echoes `dateSelected` back into `selectedDate`, year/multi-year navigation is reset. Consider running only on input changes (e.g. compare to a previous value or scope the effect to upstream changes).

3. **`showQuickAccess` is declared and forwarded but never read.** Set on `dialog-schedule-task.component.ts:74` and passed by repeat-cfg as `true`, but no template references it — quick-access buttons render unconditionally. In `isSelectDueOnly` edit mode, clicking one calls `submit()` and overwrites the configured `startDate`. Either wire the flag or remove it (and disable quick-access click-to-submit in select-due-only mode).

4. **`minDate: undefined` from repeat-cfg means editing a past-dated repeat-cfg floors the calendar at today.** `dialog-edit-task-repeat-cfg.component.ts:182` passes `undefined`; `dialog-schedule-task.component.ts:100` falls back to `new Date()`. If preserving an existing past `startDate` is intended, pass it as `minDate`; otherwise drop the data property and note the behavior change.

5. **DateService rule violation in repeat-cfg.** `dialog-edit-task-repeat-cfg.component.ts:590`:
   ```ts
   const todayStr = getDbDateStr(new Date());
   const isTargetTodayOrPast = this._data.targetDate <= todayStr;
   ```
   This is a "what day is today?" check — should route through `DateService.todayStr` so users past the logical-day cutoff see the right boundary.

### Tests

6. **`datetime-picker.component.spec.ts` doesn't exercise any `effect()`.** The 9 test cases cover output wiring only. The load-bearing surfaces (`_syncActiveDateEffect`, locale propagation, `MatCalendar.stateChanges` subscription cleanup, `ngAfterViewInit` focus, `enterSubmit` flow) are untested.

7. **`dialog-schedule-task-select-due-only.spec.ts` lost the `Time handling in select-due-only mode` describe block** (`onTimeClear`, default-time-on-focus). The behavior is still present in the component — please re-home the coverage.

8. **No test covers the `remindOptionToMilliseconds` ↔ `millisecondsDiffToRemindOption` roundtrip.** When the repeat-cfg reopens with an existing reminder, the dummy task converts ms → option via `millisecondsDiffToRemindOption`, which quantizes to the nearest bucket — a user-saved `m10` may come back as something else.

### E2E reliability

9. **Five recurring specs use translated strings as locators** — the same fragility class the PR just fixed for "Open calendar":
   ```
   e2e/utils/recurring-task-helpers.ts:~100
   e2e/tests/recurring/invalid-clock-string-bug-7067.spec.ts:~54
   e2e/tests/recurring/recurring-future-start-date-bug-6856.spec.ts:~73
   e2e/tests/recurring/recurring-start-date-epoch-bug-6860.spec.ts:~72
   e2e/tests/recurring/repeat-timed-cold-reopen-day-change.spec.ts:~71
   ```
   All use `getByRole('button', { name: 'Schedule' })`; "Schedule" maps to `T.F.SCHEDULE_TASK.SCHEDULE` in `en.json`. Prefer a data attribute or a stable role-based path (`mat-dialog-actions button[color="primary"]`, etc.).

### Styling / i18n

10. **Global `!important` calendar cell hover** in `src/styles/components/_overwrite-material.scss:434-447` applies to every `mat-calendar` site-wide with no opt-out. Future consumers inherit the accent outline + `--calendar-hover-bg` background. Scope behind a wrapper class, or expose an opt-out token.

11. **`--calendar-hover-bg` is only declared on `:root`.** No `body.isDarkTheme` override — 35% primary on a dark surface generally under-contrasts.

12. **German i18n missing 4 of 7 new `DATETIME_SCHEDULE` keys.** `en.json` adds `PREVIOUS_YEAR`, `NEXT_YEAR`, `PREVIOUS_24_YEARS`, `NEXT_24_YEARS`; `de.json` only adds the two `SWITCH_TO_*` keys and `PRESS_ENTER_AGAIN`. German users will see raw keys for year-nav aria-labels.

13. **i18n bucket inconsistency.** `PREVIOUS_MONTH`/`NEXT_MONTH` live under `F.SCHEDULE.*`; new year siblings under `DATETIME_SCHEDULE.*`. Same family, two namespaces — easy to miss when translating.

14. **`MatDatepickerIntl` constructor uses `translate.instant()` once and subscribes to `onLangChange`** — if the calendar opens before the JSON has loaded, labels render as raw keys. Subscribing to `onTranslationChange` / `onDefaultLangChange` (or using `get()`) would cover first-load.

### Nits

- `monthYearLabel` flipped from `'short'` to `'long'` in `src/main.ts:223` — visible UX change, worth a note in the PR description.
- `as unknown as TaskCopy` cast in `dialog-edit-task-repeat-cfg.component.ts:137-151` hides a missing `id` field on the dummy task. Prefer a typed `Partial<TaskCopy>` or build from `DEFAULT_TASK`.

Happy to look again once the correctness items are addressed.

hajiboy95:
@johannesjo adressed and merged master

johannesjo:
Thanks for the latest round — the shared `<datetime-picker>` direction is good and the earlier items (locale-fallback revert, custom-header removal, shared hover token, `MatDatepickerIntl` provider, configurable quick-access labels) all look addressed.

One blocker before merge, plus a couple of confirms.

### Blocker: quick-access regression breaks two existing (untouched) consumers — this is the red CI

`DialogScheduleTaskComponent.onQuickAccessClick()` now skips submit in select-due-only mode:

```ts
if (!this.data.isSelectDueOnly) {
  this.submit();
}
```

On `master`, `quickAccessBtnClick` always called `submit()`, so picking Today/Tomorrow closed the dialog and returned `{ date, ... }`. The new guard was added for the repeat-cfg flow (keep the dialog open so a time can still be added), but `isSelectDueOnly` is shared by other callers that rely on the old "apply-and-close" behavior:

- **add-task-bar** (`add-task-bar-actions.component.ts:223`) — its `afterClosed()` sets the due date. Quick-access now leaves the dialog open and the date is silently dropped. This is what fails `add-task-bar-date-picker.spec.ts:64` and `:81` in CI.
- **calendar reschedule** (`calendar-event-actions.service.ts:138`) — `await firstValueFrom(afterClosed())`. Same root cause; the promise now just hangs on a quick-access pick. No e2e covers it, so CI is green-silent here, but it's broken too.

`isSelectDueOnly` was a date-vs-datetime selector; overloading it to also mean "don't auto-submit on quick-access" is what causes the collision. Suggest a dedicated, opt-in flag instead, e.g. `keepOpenOnQuickAccess` (or `submitOnQuickAccess`, default true), set only by the repeat-cfg dialog. That leaves add-task-bar and calendar reschedule on the original behavior with no changes, and the new repeat-cfg UX still works.

Please also add e2e (or at least a dialog unit test) for quick-access → submit in select-due-only mode so this can't regress silently again.

### Worth confirming (behavior changes)

1. **New recurring configs can now have a past start date.** The repeat dialog opens the scheduler with `minDate: null` and the old formly today-floor (#7768 Bug 4) was removed. This fixes the "can't pick a date between a past start and today" issue I raised, but it also drops the floor for brand-new configs. Intended?
2. **Deadlines can no longer be set in the past** (`dialog-deadline.component.ts` `minDate = new Date()`). You mentioned this is intentional — just confirming, since the old deadline calendar had no `minDate`.
3. **`monthYearLabel` is now `'long'` app-wide** (`main.ts`). Affects every Material datepicker, not just these dialogs — noted in your description, flagging for visibility.

### Minor

- `datetime-picker.component.ts`: `_dateAdapter` and `_dateTimeFormatService` are injected but unused, and `_lastView` is assigned but never read — please drop.
- `de.json` was edited; convention here is to touch `en.json` only.

hajiboy95:
@johannesjo once more, thank you very much for the feedback. Adressed what you said and got a few answers regarding your questions:

> New recurring configs can now have a past start date. The repeat dialog opens the scheduler with minDate: null and the old formly today-floor (https://github.com/super-productivity/super-productivity/issues/7768 Bug 4) was removed. This fixes the "can't pick a date between a past start and today" issue I raised, but it also drops the floor for brand-new configs. Intended?

-> thanks for that heads up. Not really intended. Improved

 > Deadlines can no longer be set in the past (dialog-deadline.component.ts minDate = new Date()). You mentioned this is intentional — just confirming, since the old deadline calendar had no minDate.

-> intended. I don't see necessity to put deadline in the future to the past. deadlines are something which should be planned ahead and i don't think it's a case that in the future you want to set a deadline even further into the past. If you're not okay with it we could technically remove it, but i think it's okay? 


> monthYearLabel is now 'long' app-wide (main.ts). Affects every Material datepicker, not just these dialogs — noted in your description, flagging for visibility.

-> that's okay 

> de.json was edited; convention here is to touch en.json only.

thanks. wasn't intended. I reverted this

johannesjo:
Thank you very much for your hard work on this! <3

Issue #2657, opened 2023-06-28
Task Hierarchies (Sub tasks for sub tasks / Sub-Tasks within Sub-Tasks)


thraidh:
## Problem Statement

I have a project with many topics and services. When translating the customer requirements into tasks, I usually go top-down, which leads to tasks like this:

- ...
- connectivity
    - vpn server
        - automatic certification
        - handle client certification
        - forwarding rules
            - rule 1
            - rule 2
            - rule 3
    - vpn client
        - automatic certification
        - automatic reconnection
        - OS integration/autostart
- ...

Currently it is not possible to do something like that in Super Productivity in a satisfactory way.

## :grey_question: Possible Solution

Tasks can be handled as they are today, but visually tied together, i.e. by just indenting child tasks. Optionally you could show/hide child tasks.

## :arrow_heading_up: Describe alternatives you've considered

Today there are several unsatisfactory ways to approximate the functionality:
- Convert super tasks into their own projects: unsatisfactory, because they are not standalone projects, just part of a bigger project
- Name tasks like "Connectivity / VPN Server / Forwarding Rules / Rule 1": unsatisfactory, because the tasks belonging together are now totally independent and it is not possible to limit your focus on e.g. only "VPN Server"
- Convert some tasks into sub tasks as available today: unsatisfactory, because it is unclear where to make the split and it would still be necessary to name tasks and or subtasks with "/"-notation
- Name tasks after the leaf name only and add tags like `#connectivity`, `#vpnserver`, `#automaticcertification`: unsatisfactory, since `#automaticcertification` appears in VPN client and server and denotes different tasks
- Use tags like `#connectivity_vpnserver_automaticcertification`: unsatisfactory, because of long names and essentially every task would get its own tag, which is probably not the intended way to use tags
- Only use 2 levels and try to remember the lower levels or use another tool/text file to keep track of those tasks: unsatisfactory, because I have already a tool which does almost everything I need
- Don't use Super Productivity for complex projects: unsatisfactory, because I like Super Productivity and working on complex projects is my job

johannesjo:
Related #643

thraidh:
Thanks for considering! Is there any way I can help?

johannesjo:
As mentioned in the related issue this is currently not planned. As mentioned in the issue there is also the possibility to use the task description for deeper levels of planning:

``` 
- Something /
- Comet Haley /
```

RokeJulianLockhart:
[...]

PR #8038, opened 2026-06-05
feat(tasks): support nested subtasks

johannesjo:
## Summary

Adds support for nested subtasks up to four total task levels for Issue #2657. The persisted task model already supports a tree via `parentId` and `subTaskIds`, so this keeps storage and sync action shapes stable while lifting the runtime two-level assumptions.

## What changed

- Adds shared task-tree utilities for descendants, ancestors, depth, subtree height, depth-gated nesting, and recursive `TaskWithSubTasks` mapping.
- Makes selectors, task rendering inputs, startable/trackable task logic, time rollups, delete/archive/restore, project/tag/planner cleanup, undo capture, and REST create/restore paths tree-aware.
- Enforces `MAX_TASK_DEPTH = 4` in reducers and user entry points so UI, REST, plugins, and op-log replay share the same hard gate.
- Updates markdown structured paste/export, issue import, daily summary, worklog restore, and Android reminder cancellation to preserve nested descendants where applicable.
- Updates wiki docs and architecture decisions to document the four-level cap.

## Validation

- `npm run checkFile` for every changed `.ts`/`.scss` file and the new task-tree files
- `npm run build`
- `npm test`
  - Europe/Berlin Angular suite: 10,380 success, 2 skipped
  - America/Los_Angeles Angular suite: 10,366 success, 16 skipped
- Pre-push hook reran lint and `npm test` successfully before pushing

## Notes

Cross-level hover-to-nest drag-and-drop is not included here; the reducer/context-menu/REST paths are the authoritative working re-parent paths, with existing drag/drop behavior guarded by the same depth checks.


dominik3bb:
@johannesjo I apologize if it's a dumb question, but Is it not possible to create a (community) plugin that implements features like these that are more experimental, for people that want it? And still being able to let those people sync their SP's between devices?

Edit: Okay, I get it now why this might be kinda cursed 
image

(And I found out that it's not possible with plugins)

PR #8039, opened 2026-06-05
Feature/add markdown keyboard shortcuts

forge34:
## Problem
Formatting text in the Markdown editor requires either manually typing the markdown syntax or stopping to click the formatting buttons on the toolbar.

## Solution

added markdown shortcut support by extending the `keydownHandler` function in `dialog-fullscreen-markdown.component.ts`.

added support for the following commands: 

* **Bold:** `Ctrl/Cmd + B`
* **Italic:** `Ctrl/Cmd + I`
* **Insert Link:** `Ctrl/Cmd + K`
* **Strikethrough:** `Ctrl/Cmd + Shift + S`
* **Unordered List:** `Ctrl/Cmd + Shift + 8`
* **Ordered List:** `Ctrl/Cmd + Shift + 7`
* **Blockquote:** `Ctrl/Cmd + Shift + 9`
* **Inline Code:** `Ctrl/Cmd + E`

also added shortcut to the on hover tooltip of the toolbar icons and added a keyboard shortcut dialog

image
image

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
I found a few issues that should be addressed before merging:

1. `Ctrl/Cmd+Shift+X` conflicts with the app’s default Electron global show/hide shortcut. The new strikethrough handler is added in `src/app/ui/dialog-fullscreen-markdown/dialog-fullscreen-markdown.component.ts:226`, while `globalShowHide` defaults to `Ctrl+Shift+X` in `src/app/features/config/default-global-config.const.ts:129` and is registered via Electron `globalShortcut`. In the desktop app, this can hide/minimize Super Productivity instead of applying strikethrough. Please choose a non-conflicting shortcut or account for configured global shortcuts.

2. The new line-format shortcuts leave the full transformed line selected after use with a collapsed caret. `Ctrl/Cmd+Shift+8/7/9` call the bullet, numbered-list, and quote transforms at `dialog-fullscreen-markdown.component.ts:232`, `:238`, and `:244`; `_applyTransformWithArgs()` then restores `result.selectionStart/result.selectionEnd` at `:421`. For these transforms, a collapsed caret returns the whole touched line as selected, so typing immediately after the shortcut replaces the marker/line. Keyboard-triggered line formatting should collapse/preserve the caret after the inserted prefix.

3. The shortcut help UI adds untranslated user-facing strings. `dialog-markdown-shortcuts.component.html:1` and `:7-61` hardcode the dialog title, headers, and action labels, and `dialog-fullscreen-markdown.component.html:181-182` hardcodes the tooltip/aria label. Per the repo rules, these should go through `T` / translations; several action labels can reuse existing toolbar translation keys.

4. The shortcuts dialog is not responsive on narrow screens. It opens with `width: '402px'` in `dialog-fullscreen-markdown.component.ts:185-188`, while the content enforces `min-width: 350px` in `dialog-markdown-shortcuts.component.scss:7` and keeps shortcut cells nowrap at `:48`. On small mobile viewports this can overflow or clip. Please make the dialog/content width viewport-bounded and allow the layout to adapt.

forge34:
Hi 

1. i decided to use `Ctrl/Cmd+Shift+S` instead since as far as i know there is no conflicts with other key maps
2. i am not sure how to approach solving it, i initially thought of adding a boolean argument to `_applyTransformWithArgs()` which is used to check if it's triggered by keyboard or toolbar and sets the caret behavior based on it
3. i have reused the action labels, for the others do i manually add them to `src/assets/en.json`?

also i will continue working on this issue tomorrow since i am a bit tired today :)

johannesjo:
Thanks for iterating on this so quickly — most of the earlier feedback is now addressed. 🎉 I checked the latest diff and confirmed:

- ✅ **#2 (caret on line formats)** — the `isKeyboard` flag collapsing the caret works correctly. Inline shortcuts (B/I/E/K) keep their mid-marker / placeholder behavior, line shortcuts collapse — exactly right.
- ✅ **#3 (translations)** — new `T` keys + `en.json` entries + reused toolbar labels look good.
- ✅ **#4 (responsive dialog)** — `maxWidth: '100vw'` + dropping `min-width` + `overflow-x: auto` handles narrow screens.

A few things still to fix before merge:

**1. Strikethrough tooltip is out of sync (must fix).** You moved the handler to `Ctrl/Cmd+Shift+S` (and the help dialog shows **S**), but the toolbar button tooltip in `dialog-fullscreen-markdown.component.html` still hardcodes `(Ctrl/Cmd+Shift+X)`. That's the exact shortcut we wanted to avoid (it hides the app on desktop), so the tooltip would now mislead users. Please update it to `(Ctrl/Cmd+Shift+S)`.

**2. Consider a single source of truth for the shortcuts.** Right now the shortcut set lives in three places — the `keydownHandler` logic, the per-button tooltip strings, and the help-dialog table. The tooltip bug above is a direct symptom of that duplication: one spot got updated, the others didn't. A small `MARKDOWN_SHORTCUTS` constant (action key → combo) that the handler dispatches on and that both the tooltips and the help table render would prevent this drift and shrink the template. Not strictly required to merge, but it'd kill the root cause.

**3. Tests.** The checklist has "added tests" ticked, but I don't see any in the diff. `keydownHandler` now has real branching and `_applyTransformWithArgs` has the new keyboard-vs-toolbar caret behavior — both are easy to cover in the existing `dialog-fullscreen-markdown.component.spec.ts`. A couple of cases (each combo dispatches the right transform + `preventDefault`; the caret difference on a line transform) would be great per our test policy.

Minor nits (optional):
- `dialog-markdown-shortcuts.component.ts` line 1 has a leftover `// Add ViewEncapsulation` comment — safe to remove.
- `keydownHandler` uses `==` in a few spots; the codebase uses `===` (no behavioral diff here, just consistency).

For what it's worth, I verified there's **no** in-app conflict for `Ctrl+B`/`Ctrl+S` etc. — `ShortcutService` already bails inside inputs/open dialogs, so those won't double-fire here. Nice work overall, especially using `ev.code` for the Shift+digit combos and `ev.key` for letters.

johannesjo:
Thanks for the PR — nice feature, clean structure (single source of truth in `markdown-shortcuts.const.ts`, help dialog wired to the toolbar, translations + tests added). A few things I'd like addressed before merge:

### Bugs

**1. Wrong `key:` values in `markdown-shortcuts.const.ts`**

```ts
{ name: 'bullet',   shortcutLabel: 'Ctrl/Cmd+Shift+8', key: '9',  code: 'Digit8', shiftKey: true },
{ name: 'numbered', shortcutLabel: 'Ctrl/Cmd+Shift+7', key: '8',  code: 'Digit7', shiftKey: true },
{ name: 'quote',    shortcutLabel: 'Ctrl/Cmd+Shift+9', key: '10', code: 'Digit9', shiftKey: true },
```

The `key` values don't match the labels, and `key: '10'` is never a valid `KeyboardEvent.key`. The handler uses `code` when present so this is masked today, but it's a landmine if `code` is ever dropped or another entry without `code` happens to share `'9'`/`'8'`/`'10'`. Either omit `key` when `code` is set, or fix the values to `'8'`/`'7'`/`'9'`.

**2. Redundant optional chaining after a truthy guard** — `dialog-fullscreen-markdown.component.ts`

```ts
if (shortcut) {
  ev.preventDefault();
  this._executeShortcutByName(shortcut?.name);  // <-- ?. is redundant
  return;
}
```

Drop the `?.`. With `shortcut.name: ShortcutNames`, the switch becomes exhaustive — better than silently no-op'ing on an unknown name.

**3. Thin test coverage**

The spec only tests Ctrl+B. The newer/trickier paths — `code`-based digit shortcuts (Cmd+Shift+8/7/9), `metaKey` on Mac, Cmd+K, Cmd+Shift+S, Cmd+E — aren't exercised. A few more cases would be cheap insurance given that the key/code matching logic is the most error-prone bit (see #1).

### Design suggestion

**4. `isKeyboard` parameter feels like a band-aid**

The "collapsed caret after applying list/quote" fix lives in `_applyTransformWithArgs` and is only set by keyboard callers. But the same UX is wrong for a *toolbar* click with no selection: the user clicks the bullet button mid-line and the whole line ends up selected, ready to be overwritten. The fix really belongs in `applyBulletList`/`applyNumberedList`/`applyQuote` themselves: when the input selection is collapsed, return a collapsed output selection. That fixes both entry points consistently and removes the parameter plumbing.

### Nits

- **Shortcut labels hardcoded, not translated.** `shortcutLabel: 'Ctrl/Cmd+B'` lives in code; en.json only has the action names. Inconsistent with the project's "UI strings through `T`" rule, though low-stakes.
- **`'Ctrl/Cmd'` rendered as a single chip** in the shortcuts dialog (`split('+')` yields `['Ctrl/Cmd', 'Shift', '8']`). Showing `⌘` on Mac / `Ctrl` elsewhere via the existing platform util would look nicer.
- **No `altKey` check** — `Ctrl+Alt+B` triggers Bold. Conventional to require modifier exclusivity.
- `this._shortcuts = MARKDOWN_SHORTCUTS` — assigning a top-level constant to an instance field doesn't buy anything; use the import directly (same in `DialogMarkdownShortcutsComponent`).
- `getShortcutLabel('bold')` in the template runs on every CD cycle for every tooltip — fine for a dialog, but a one-time computed `tooltipBold`/etc. would avoid repeated `.find()` calls.
- PR description still lists `Ctrl/Cmd + Shift + X` for strikethrough; code uses `Shift + S` after `ee4da2aee`.

Happy to take another pass once 1–3 are addressed.

forge34:
Thanks for the feedback!

I've addressed #1, #2, #3 and several of the nits:

1. Refactored `MarkdownShortcut` so `key` and `code` are mutually exclusive.
2. Removed the optional chaining.
3. Added additional shortcut test cases.

Nits:

1. I wasn't sure about moving shortcut labels to translations. Do shortcut labels typically vary by locale?
2. fixed by adding `const MOD = IS_MAC ? '⌘' : 'Ctrl';` and updated `MARKDOWN_SHORTCUTS` to use it.
3. Updated modifier detection to:
   `const hasModifier = (ev.ctrlKey || ev.metaKey) && !ev.altKey;`
4. Removed `_shortcuts`.
5. Added a shortcut label map that's computed once and reused in the template.
6. Updated the PR description.

for #4: where do i put the caret after transformation? at it's position before the transformation or at the end of a block/line ?

johannesjo:
**Review of the markdown-shortcut changes** (compare `95f1984..dbb6fe4`)

> Heads-up on scope: that compare range includes a `Merge branch 'master'` commit, so ~150 of the changed files are unrelated master work. The actual change here is 5 commits touching 6 files in `src/app/ui/dialog-fullscreen-markdown/`. (Reviewed with several independent passes.)

The core is solid — the platform-aware modifier label, the `code`-based digit shortcuts (layout-independent — the right call, and good cleanup of the old stale `key: '9'/'8'/'10'` values), the discriminated-union types, and the `altKey` exclusion all check out. A few things below.

### 🔴 Should fix — toolbar tooltips lost their separating space
The label precompute dropped the leading space the old `getShortcutLabel()` returned:

```ts
// before — getShortcutLabel() returned:  ` (${shortcut.shortcutLabel})`   // leading space
// after:
MARKDOWN_SHORTCUTS.map((s) => [s.name, `(${s.shortcutLabel})`])            // no leading space
```

The template still concatenates `(… | translate) + shortcutLabels['bold']`, and the toolbar strings have no trailing space (`"Bold"`), so every toolbar tooltip now renders **`Bold(⌘+B)`** instead of `Bold (Ctrl/Cmd+B)`. Trivial fix — restore the leading space:

```ts
MARKDOWN_SHORTCUTS.map((s) => [s.name, ` (${s.shortcutLabel})`])
```

### 🟡 Add a test for the `altKey` exclusion
`hasModifier = (ev.ctrlKey || ev.metaKey) && !ev.altKey` is a good fix — it stops AltGr (Ctrl+Alt on Windows/Linux, used to type `€`/`@`/etc.) from being swallowed as a shortcut. But nothing tests it; the spec has no `altKey` case. A quick `{ ctrlKey: true, altKey: true, key: 'b' }` → `onApplyBold` **not** called test would lock the fix in against future refactors. (Same goes for the label content — a small assertion on `shortcutLabels['bold']` would have caught the spacing regression above.)

### 🟡 Modifier-key spelling is inconsistent with the rest of the app
This introduces a third spelling of the Mac key. The app already renders shortcut hints platform-aware **and** translated, spelling it `Cmd`/`Meta`:
- `task-detail-panel.component.ts:150` → `PASTE_IMAGE_HINT_MAC: "Cmd+V to paste image"` (goes through `T`/translate)
- keyboard config writes `Meta+` (`keyboard-input.component.ts:62`)

…whereas these labels are hardcoded `⌘`/`Ctrl` in `markdown-shortcuts.const.ts` and bypass `TranslateService` (repo rule: UI strings go through `T`). Consider aligning with the existing `PASTE_IMAGE_HINT_MAC` pattern. Not a blocker — `Ctrl`/`Cmd`/`Shift` are near-universal — but it's a visible inconsistency.

### 🟢 Nits
- **`shortcutLabel.split('+')`** in the help dialog is fine for the current labels but fragile — a future `+`-key shortcut (the app has `Ctrl++` for zoom elsewhere) would split wrong and render an empty chip. Storing keys as a structured `string[]` instead of a `+`-joined string would be sturdier and also makes the i18n point above easier to solve.
- **`Object.fromEntries(...) as ShortcutLabel`** suppresses exhaustiveness — drop a shortcut from the array and `shortcutLabels['x']` is silently `undefined` in the tooltip with no compile error. A `satisfies`-checked literal or a dev assertion would catch it.
- The handler accepts **both** `ctrlKey` and `metaKey` on every platform while the label shows only one — harmless and forgiving, but a one-line comment noting the dual-accept is intentional would help, since it technically contradicts the displayed label.

Otherwise LGTM — the matcher logic, strict `shiftKey` matching, type narrowing, and the `Ctrl+Enter` close path all verified correct.

johannesjo:
### Review of the latest changes (`shortcutLabels` refactor)

Thanks for the follow-up\! The refactor is a nice improvement — moving label generation into `markdown-shortcuts.const.ts` gives a single source of truth, both dialogs now consume the same generated labels, and the `keys` array removes the brittle `.split('+')` in the shortcuts dialog. The core logic (`isShortcutWithKey`, `formatKeyDisplay`, the `reduce`) is correct. A few things I'd suggest before merge:

**1. `process.env['NODE_ENV']` dev-guard — non-idiomatic & fragile (please change)**

```ts
if (process.env['NODE_ENV'] === 'development') { ... }
```

This is the only `process.env` usage in the whole app source (`src/app`). It works today only because esbuild implicitly substitutes `process.env.NODE_ENV` for browser-platform builds (`"development"` unminified, `"production"` minified → dead-code-eliminated in prod). That's an implicit bundler behavior, not an Angular contract — if it ever stops, this top-level code throws `ReferenceError: process is not defined` at module load and breaks the markdown dialog.

The codebase already has established idioms for this: `isDevMode()` (used in `meta-reducer-registry.ts`) and `environment.production` (used in `util/dev-error.ts`). Please switch to one of those, or — simpler — drop the block entirely (see #2).

**2. The dev-check duplicates `ShortcutNames` — low value**

`allKeys` re-lists the same 8 names already in the `ShortcutNames` union and `MARKDOWN_SHORTCUTS`. Adding a shortcut now means editing three places, and forgetting `allKeys` silently reduces coverage. Since `shortcutLabels` is generated from `MARKDOWN_SHORTCUTS` and the new spec already asserts specific labels, the runtime guard adds little. I'd just delete the whole `if` block — the spec covers it.

**3. New tests are platform-dependent**

```ts
expect(shortcutLabels['bold'].tooltip).toBe(' (Ctrl+B)');
expect(shortcutLabels['bold'].keys).toEqual(['Ctrl', 'B']);
```

`MOD` is `IS_MAC ? 'Cmd' : 'Ctrl'`, and `IS_MAC` reads `navigator.platform`, so on a Mac dev machine these three assertions fail (`MOD` is `'Cmd'`). CI is Linux so it stays green, masking it. Suggest asserting against the `MOD` constant (or mocking `IS_MAC`) rather than hardcoding `'Ctrl'`.

**4. Mac modifier display changed `⌘` → `Cmd` — intentional?**

This range flipped `IS_MAC ? '⌘' : 'Ctrl'` to `'Cmd'`. macOS users conventionally expect the `⌘` glyph in shortcut hints. Not wrong, but a visible downgrade — just want to confirm it was deliberate.

**Nits**
- Comment "though the labels shows one of them" → "the label shows".
- `satisfies ShortcutLabel` right after `{} as ShortcutLabel` is redundant — the cast already fixes the type, so `satisfies` can't catch a missing key.

Overall the refactor is safe to ship; #1 + #2 are the main asks (replacing the `process.env` block with `isDevMode()`, or deleting it, resolves both), and #3 is a real cross-platform flake for Mac contributors.

forge34:
Thanks for the feedback.

i addressed #1 and #2 by removed the dev-gaurd completely and fixed #3 by using `MOD` in the tests.
yes #4 was intentional for consistency with the rest of the codebase, i addressed the nits as well.

johannesjo (approved):
Thanks for the thorough back-and-forth on this — the single-source-of-truth refactor, layout-independent `code` matching, `altKey` exclusion, translations, and tests all look good, and CI is green. Merging now. The outstanding minor items (the `isKeyboard` → transform-level caret fix, label exhaustiveness, label i18n) are captured in #8268 as follow-ups.

forge34:
Thanks for the review and the detailed feedback throughout.

Learned a lot working on this issue. 
Happy pick up the Follow-up as well #8268 if you would like, but i have some questions i will leave on the issue thread itself :)

bbinet:
I also like the Markdown Keyboard Shortcuts dialog which helps being more productive by learning easily the keyboard shortcuts.
It would be great if such a dialog would exist to show shortcuts of the whole SP app: it could also be shown using a keyboard shortcut, for example using the "?" key: like in gmail.
This would be great for helping new comers to being more productive by learning easily new keyboard shortcuts.

forge34:
i think this is something worth opening a new issue about @bbinet

bbinet:
Yes, will do now.

bbinet:
Issue created: #8284

Issue #8041, opened 2026-06-05
💡 Keyboard shortcut to set deadline for a task


deleonja:
### Problem Statement

Currently, there is no keyboard shotcut to set deadlines for a selected task---as it already exists for schedulinkg a task. 

### Possible Solution

Allow the possibility to assign a keyboard shortcut to set the deadline for a selected task.  Maybe default shortcut could be either 'Shift + d' (*d*eadline) or 'Shift + s' (given that 's' is the default shortcut for scheduling).

hajiboy95:
@deleonja great suggestion. I went for for `Shift + s`.

During the implementation however I noticed a current bug @johannesjo 

Priority Reordering of Keyboard Shortcuts: Currently in ShortcutService, global shortcuts are evaluated before task-specific shortcuts. This causes conflicts when the same key combination is mapped to both a global action and a task-specific action. E.g., Shift+T is mapped to both global goToTimeline and task-specific moveToTodaysTasks. Because global shortcuts take precedence, moveToTodaysTasks (Shift+T) is currently non-functional when a task is focused. Similarly, mapping the deadline shortcut to Shift+S would conflict with the global goToScheduledView (Shift+S).

Proposed Fix: I will modify ShortcutService to evaluate task-specific shortcuts first whenever a task is focused. If the task shortcut handler doesn't match the key combo, it will fall through to global shortcuts. This cleanly fixes the Shift+T bug and allows us to safely use Shift+S for task deadlines when a task is focused, while retaining Shift+S for navigating to the Schedule page when no task is focused.

hajiboy95:
@deleonja it's implemented now, so can you close this issue?

deleonja:
Thank you!!!

Issue #7700, opened 2026-05-20
🚨 SuperSync stopped syncing after adding second device


darius-r:
### Version Used

18.6.0 x64 installed version

### Current Behavior

I switched my desktop client from Dropbox to SuperSync, and set a new encryption password. 
The initial sync was successful. 

I then changed my android phone from Dropbox to SuperSync. This was successful, and it prompted for the new encyption password. I instructed the android client to overwrite the local data with what it finds on the server. 

I then 'unarchived' one task on the desktop software to test the sync behaviour. 

The first error to appear:
<img width="816" height="100" alt="Image" src="https://github.com/user-attachments/assets/31356114-b115-4edf-968b-af31ad92d8e2" />

Clicking 'reload' triggered:
<img width="756" height="78" alt="Image" src="https://github.com/user-attachments/assets/568737be-427b-445f-aa10-7357d752041f" />

After the initial successful sync, subsequent attempts at syncing trigger:
<img width="863" height="96" alt="Image" src="https://github.com/user-attachments/assets/2fcdbf46-eefe-4ebb-b7c6-da3035602049" />

Note that more than 90 seconds had passed, and that the sync only includes less than 10 tasks, and less than 20 .ical syncing events (which are not tasks). 

Note also that changes on Android are successfully synced to SuperSync, but it does not recieve the change made on desktop, nor does the desktop receive the changes from android. 

[super-prod_log-file.txt](https://github.com/user-attachments/files/28068680/super-prod_log-file.txt)

### Expected Behavior

For changes to sync bi-directionally

### Steps to Reproduce

Unclear, but I suspect this was triggered by either:
-The change type being 'unarchiving' of a task or;
-That I switched from Dropbox (encrypted) sync, to SuperSync

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

darius-r:
Curiously, even after making changes on Android that appear to have been successfully synced, I still only see the intial 'full sync' as an available restore point on both android and windows. 

Its been approximately 62 minutes since that initial import. It appears that restore points are not being created automatically at even 1-hour intervals on the server at this time. 

[...]

PR #8042, opened 2026-06-05
feat(tasks): add keyboard shortcut to set task deadlines

hajiboy95:
## Problem

Tackles #8041 

Currently, there is no keyboard shortcut to set deadlines for a selected task (unlike scheduling a task, which has the `S` shortcut). 

Additionally, there was a shortcut precedence conflict in `ShortcutService` where global navigation shortcuts were evaluated before task-level shortcuts. This made certain task-level shortcuts (such as `Shift+T` for moving a task to Today) completely non-functional when a task was focused, as they shared key combinations with global shortcuts (like `Shift+T` for navigating to the Timeline view).

## Solution

1. **Added Task Deadline Keyboard Shortcut**: Created a new configurable task shortcut `taskScheduleDeadline` (defaulting to `Shift+S`) that triggers the deadline selection dialog (`openDeadlineDialog()`) on the currently focused task.
2. **Corrected Precedence/Conflicts**: Modified `ShortcutService` to evaluate task-specific shortcuts *before* global shortcuts when a task is currently focused. This resolves the key collision cleanly:
   - When a task is focused: `Shift+S` opens the deadline dialog, and `Shift+T` correctly moves the task to Today.
   - When no task is focused: `Shift+S` navigates to the Schedule view, and `Shift+T` navigates to the Today/Timeline view.
3. **Configurations & Translations**: Added the `taskScheduleDeadline` settings property to default configs, settings forms, and updated both English (`en.json`) and German (`de.json`) translation locales.
4. **Documentation**: Updated the project wiki (`docs/wiki/3.03-Keyboard-Shortcuts.md`) to document the new shortcut and explain the contextual precedence mechanism.
5. **Testing**: Updated the unit tests in `task-shortcut.service.spec.ts` to include the default mock config and added a spec validating that `Shift+S` correctly delegates to the focused task component's `openDeadlineDialog()`.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for the shortcut work. I double-checked the changed dispatch paths and found a few issues to address:

1. `ShortcutService` now calls `handleTaskShortcuts()` before global shortcuts unconditionally, but `TaskShortcutService` handles `togglePlay` even when no task is focused. That means a user-configured conflict between `togglePlay` and any global shortcut now lets `togglePlay` win even with no focused task, which is broader than the documented "task shortcuts win when a task is focused" behavior. Please gate the early task-shortcut pass to actual focused-task shortcuts, or leave the no-focus `togglePlay` fallback after global shortcuts.

2. `taskScheduleDeadline` returns handled even when delegation fails. `_handleTaskShortcut()` only logs and returns if the focused component is missing/stale, but the caller still prevents default and stops global handling. With the new default `Shift+S` conflict, a stale focus/component state can swallow `Shift+S` without opening either the deadline dialog or the schedule view. Please make delegation success/failure visible to the caller before returning `true`, or otherwise fall through to global handling when no task component was actually invoked.

3. Please remove the direct `de.json` edit. The repo guidance says only `en.json` should be edited directly for translation strings; other locales should come from the translation workflow.

4. Minor docs cleanup: the new precedence note uses `Today/Timeline` wording while the nearby shortcut list/settings labels already have inconsistent Schedule/Today wording for `Shift+T`. Since this PR touches that explanation, it would be good to align the wording so the examples match the actual labels/routes.

hajiboy95:
@johannesjo adressed

johannesjo:
Thanks for the follow-up. One behavior issue remains:

When a task shortcut matches but `_handleTaskShortcut()` returns `false`, `handleTaskShortcuts()` now falls through to `ShortcutService` global shortcuts. That breaks the new documented precedence in stale-focus cases: focus can still be inside a `<task>` while `lastFocusedTaskComponent` is stale/null, so default `Shift+S` can match `taskScheduleDeadline`, fail delegation, then continue to `goToScheduledView` and navigate to `/schedule`. The same applies to `Shift+T`, and `Y` can fall through to the selected/global toggle fallback even though focus is still inside a task.

KISS fix: once a focused-task shortcut combo matches, consume the event and return `true` even if delegation failed. That preserves task-level precedence and avoids introducing a new state model. A focused `Shift+S` `ShortcutService.handleKeyDown()` regression test would be enough to cover the real path.

Small cleanup: the final PR still touches `src/assets/i18n/de.json` with a blank-line-only hunk. Per `CLAUDE.md`, locale edits should stay limited to `en.json`, so please drop the `de.json` change. Also `docs/wiki/3.03-Keyboard-Shortcuts.md` should say `Go to Schedule`, not `Go to schedule Tasks`.

hajiboy95:
@johannesjo problems are tackled and pushed. Also fixed the LockAcquisitionTimeoutError and Jasmine timeout failures in the regression tests for issue #7700 since this caused some problems while pushing.

johannesjo:
Thank you very much! <3

Issue #8027, opened 2026-06-05
Include note content in global search


johannesjo:
## Problem Statement
Notes currently can't be found via search. The global search (the search page / "Jump to task" finder) indexes task titles, descriptions and the archive, but it does **not** index Note content. As soon as you have notes spread across several projects (or pinned to Today), the only way to find one is to open each project and scan its notes by eye.

This is the core pain described in #2373 ("Second brain with Super Productivity"):

> finding notes across multiple projects and tasks is hard … there is no search function implemented.

## Possible Solution
Extend the existing global search so it also matches Note content (and surfaces matching notes in the results list), instead of building any separate search UI.

## Implementation idea
- Search lives in `src/app/pages/search-page/`. Today it pulls from tasks + archive only.
- Notes are a first-class entity (`src/app/features/note/`, NgRx store) with a `content` field and a nullable `projectId`.
- Add notes as a result source: match against `content`, show the note (with its project context / "Today" if unscoped), and on select navigate to / open the note like search results do for tasks.
- Keep it consistent with the current result cap and debounce behaviour.

## Scope (intentionally small)
- ✅ Make existing notes findable.
- ❌ No wiki-links, backlinks, graph view, or PKM features — those are out of scope (see discussion in #2373).

hajiboy95:
@johannesjo i think we can close this?

johannesjo:
Yes, thank you! <3

PR #8044, opened 2026-06-05
feat(search): include note content in global search and support folde…

hajiboy95:
Tackles #8027.

## Problem

Currently, the global search ("Jump to task" / Search page) only indexes tasks and archived tasks, making it impossible to search for note content (project notes and Today-pinned notes). Users with notes spread across several projects have to manually scan notes by eye to find them.

Additionally, context tags on the search page only display the project/tag name, which leads to ambiguity when different folders contain projects or tags with duplicate names.

## Solution

1. **Note Indexing**: Combined `NoteService.notes$` into `_searchableItems$` on the search page. The first line of a note is used as its result title.
2. **Note Navigation & Highlighting**:
   - Selecting a note routes the user to the note's context (e.g. `/project/:projectId/tasks` or `/tag/TODAY/tasks`) with a `focusItem` query parameter.
   - The right-hand notes panel is automatically toggled open.
   - The targeted note is smoothly scrolled into view and temporarily highlighted with an outline class.
3. **Folder Path Disambiguation**: Prepend folder path hierarchy (using the `>` separator, e.g., `Folder 1 > Subfolder A > ProjectName`) to the context tags of search items using `MenuTreeService`'s folder maps, which only triggers when multiple projects/tags have the same name.
4. **Ordering of Results**:
   - **Active Tasks** are listed first.
   - **Notes** are listed second.
   - **Archived Tasks** are listed last.
5. **Future Considerations**: It might be a good idea in a future PR to either visually split active tasks, notes, and archived tasks under distinct sections in the search list (e.g. by introducing group headers or separator lines) or add a search filter selector (tasks vs. notes vs. archived).

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Review feedback after double-checking the current PR head (`f4ddca4`):

- **P2** `src/app/features/note/notes/notes.component.ts:113` - `_focusNote()` waits a fixed 300 ms, does one `document.getElementById()` lookup, and then gives up. Search navigation opens the notes panel only after `router.navigate()` resolves, and the notes component/list is rendered through the right panel plus `WorkContextService.notes$`, so the target note DOM can still be missing at 300 ms. In that case the panel opens but the selected note is never scrolled/highlighted. Please wait for the note list/rendered element or retry with cancellation on destroy.

- **P2** `src/app/pages/search-page/search-page.component.ts:89` / `:91` - folder-path context labels are not fully reactive. `projectFolderMap()` and `tagFolderMap()` are Angular signals read inside an RxJS `map`, but they are not sources of `_searchableItems$`, so menu-tree folder moves/renames do not update active task/note labels while search stays open. Archived labels are also built once into `_cachedArchiveItems`, so they can stay stale even after another source emits. Please make the folder maps part of the reactive inputs and invalidate/rebuild the archive cache when project/tag/folder context changes.

- **P3** `src/assets/i18n/de.json:1075` - this PR edits `de.json`, but the repo instructions in `CLAUDE.md` say translation changes should edit only `en.json` and never other locales. Please revert the German locale changes and let the translation workflow handle them.

hajiboy95:
@johannesjo adressed 


I think the current error doesn't occure from my changes..

johannesjo:
Thanks for addressing the previous review items. One remaining issue:

**P2** `src/app/features/note/notes/notes.component.ts:111`

`_focusNote()` starts a fresh retry loop for every `focusItem` query-param emission, but older retry loops are not cancelled or invalidated. Since each loop can run for up to 4s, a stale request can later scroll/highlight an older note after a newer `focusItem` has arrived.

KISS fix should be enough here: keep a single pending timeout/current focus token on the component, clear or invalidate it before starting the next retry, and register destroy cleanup once. No need for a broader abstraction.

Also, could you add one regression spec for the search folder-map reactivity fix? The existing folder-path specs seed `projectFolderMapSignal` before init, so they would not catch the original stale-label bug. A small test that updates the folder map after `initAndFlush()` should cover it.

hajiboy95:
@johannesjo fixed. ready to go

johannesjo:
I double-checked the latest compare range and found one behavior issue plus one test gap:

1. In `src/app/features/note/notes/notes.component.ts:102` / `111-132`, every `focusItem` query-param emission starts an independent retry loop, but older retries are never cancelled or invalidated. While the notes panel stays open, navigating note A -> note B in the same context can still let the stale A retry scroll/highlight after B, and task navigation also causes the notes panel to poll for a matching note id. Could we track the active focus request/timeout and cancel or ignore stale retries when `focusItem` changes?

2. Optional but useful: `src/app/pages/search-page/search-page.component.ts:109-132` fixes archive cache invalidation for project/tag/folder-map changes, but the current specs only cover folder maps set before initialization. A regression test that searches an archived result, changes the relevant project/tag/folder-map input afterward, and asserts the archived label updates would lock this fix down.

hajiboy95:
@johannesjo done. Current tests errors haven been caused by me and are probably fixed now on the main

johannesjo:
I double-checked the current PR head (`4f361b3c32c23f5abe18731dc9d330b66769372b`). The earlier retry-loop and folder-map issues look addressed. I found two remaining items:

- **P2** `src/app/features/note/notes/notes.component.ts:143` - The stale-token guard also prevents cleanup of an already-highlighted note. Repro path: search/select note A, then select note B before A's 3s cleanup timeout fires. `_focusToken` changes for B, so A's timeout skips `classList.remove('highlight-searched-item')`, leaving A outlined indefinitely via `notes.component.scss`. The token should cancel stale retries, but cleanup for the element that was already highlighted should still run, or the previous highlighted element should be cleared before starting a new focus.

- **P3** `src/app/pages/search-page/search-page.component.spec.ts:58` - There is a leftover `console.log('DEBUG NoteService:', NoteService);` in the spec. Please remove it before merge.

hajiboy95:
@johannesjo adressed

hajiboy95:
@johannesjo 2 jobs got canceled due to higher priority, could you rerun them?

johannesjo:
Thank you very much! <3

PR #8055, opened 2026-06-06
fix(markdown): preserve soft line breaks #8054

### LINKED ISSUE #8054: 💡 Markdown editor: Parsed Markdown should support line breaks

cocojojo5213:
## Summary
- enable marked soft line breaks for parsed markdown notes
- cover the `1\n2\n3` case from #8054 with the full marked options factory

Fixes #8054

## Validation
- npm run test:file src/app/ui/marked-options-factory.spec.ts
- npm run checkFile src/app/ui/marked-options-factory.ts
- npm run checkFile src/app/ui/marked-options-factory.spec.ts

### DIFF
--- a/src/app/ui/marked-options-factory.spec.ts
+++ b/src/app/ui/marked-options-factory.spec.ts
@@ -12,8 +12,7 @@ import { marked } from 'marked';
 const parseWithFactory = (markdown: string): string => {
   marked.setOptions(marked.getDefaults());
   const options = markedOptionsFactory();
-  const renderer = { ...(options.renderer as any) };
-  marked.use({ renderer });
+  marked.setOptions(options);
   return marked.parse(markdown) as string;
 };
 
@@ -28,6 +27,7 @@ describe('markedOptionsFactory', () => {
     expect(options).toBeDefined();
     expect(options.renderer).toBeDefined();
     expect(options.gfm).toBe(true);
+    expect(options.breaks).toBe(true);
   });
 
   describe('checkbox renderer', () => {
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! <3

johannesjo:
Can you have a look @cocojojo5213 ?

cocojojo5213:
Yeah

cocojojo5213:
I checked this again on current master (c13e8ee25). The app's marked options now render `1\n2\n3` as:\n\n```html\n<p>1<br>2<br>3</p>\n```\n\nI also verified the same output through the ngx-markdown-style renderer path, and the focused spec still passes locally:\n\n```\nnpm run test:file src/app/ui/marked-options-factory.spec.ts\n# 44 SUCCESS\n```\n\nSo the current parser output should preserve the visual line breaks. One nuance: reading the rendered element's plain text/textContent will still look like `123`, because `<br>` elements don't contribute text characters. If you're still seeing a single visual line in the current build, please share the exact field/input path and I'll trace that path separately.

beerkumquatpome:
I added extra spaces before

Unparsed
1
2
3

Parsed
123 (Not 1 2 3, I added extra spaces)

Does this need to be changed?
`expect(html).not.toContain('1 2 3');` -> `expect(html).not.toContain('123');`

cocojojo5213:
Thanks for double-checking this. I don't think that change is needed.

The important assertion is the positive one:

```ts
expect(html).toContain('1<br>2<br>3');
```

That proves the generated HTML keeps explicit `<br>` elements between the lines. The negative `not.toContain('1 2 3')` is only documenting that we did not collapse the line breaks into spaces. Checking `not.toContain('123')` would also pass for the current HTML string, but it is less tied to the actual regression. The rendered DOM `textContent` can still look like `123` because `<br>` does not add text characters; visually it renders as separate lines.

beerkumquatpome:
>The rendered DOM textContent can still look like 123 because < br > does not add text characters; visually it renders as separate lines.

Isn't `expect(html).not.toContain('123');` more accurate than `expect(html).not.toContain('1 2 3');` because it's rendered as "123" not "1 2 3"?

PR #8057, opened 2026-06-06
fix(reminder): require action for reminder popup #8051

### LINKED ISSUE #8051: 🚨Bug / Discussion: Reminder popup reopens continuously after clicking outside
### Version Used

18.9.1

### Current Behavior

Currently, when the task reminder popup ("Aufgaben-Erinnerungen") appears, clicking outside of the window dismisses/minimizes it. However, just a few seconds later, the popup reappears. This creates a frustrating loop and makes it unclear what the intended behavior is for postponing or ignoring reminders temporarily.

We need to clarify the intended UX with @johannesjo and enforce a specific workflow, as the current click-away behavior feels broken.

Image

---

### **Title:** Feature Request: Add hover tooltips to action buttons in the Reminder Popup

**Problem Statement**
In the task reminder popup, there are four action buttons aligned to the right of each task (Checkmark, Calendar, Snooze/Z, Sun). Currently, there are no hover titles or tooltips for these symbols. For newer users, it is not immediately clear what exact action each button performs (e.g., whether the Sun means "tomorrow morning" or "add to my day", or what the Calendar with the 'x' does).

**Possible Solution**
Add simple, descriptive hover tooltips (e.g., standard HTML `title` attributes or the app's standard tooltip component) to each of the action buttons.

**Examples of expected tooltips:**

* **Checkmark:** "Mark as done"
* **Calendar:** "Dismiss / Unschedule"
* **Z-Clock:** "Snooze"
* **Sun:** "Schedule for tomorrow / Add to today" *(whichever matches the exact function)*

**Additional Context**
This is a quick UX win that significantly improves accessibility and discoverability without cluttering the UI.

### Expected Behavior

Here are three potential ways to handle the reminder popup, though some come with trade-offs we need to discuss:

* **Option A: Hide popup if a reminded task is started.** * *Mechanics:* We add a "Play" button directly to the reminder items. If clicked, the task starts, and the popup stays hidden while the task is active.
* *Drawback:* This subtly forces users to choose one of the currently overdue tasks. It makes it difficult for them to dismiss the popup just to start a brand-new, unrelated task.

* **Option B: Make the popup modal (blocking).** * *Mechanics:* Disable the ability to minimize the window by clicking outside. The user *must* make a distinct choice for the overdue tasks (e.g., snooze, dismiss, complete) to clear the screen.
* **Option C: Dismiss on click-away until the *next* trigger.** * *Mechanics:* Clicking outside dismisses the popup entirely. It will not appear again until a *new* task becomes overdue.
* *Drawback:* This is generally bad practice for reminders; if a user never has another task hit its overdue trigger, they might theoretically never be reminded of these currently pending tasks again.

**Recommendation**
We should enforce a clearer decision-making flow. A solid middle ground might be a mix of Option A and B: require an explicit action (snooze/reschedule/done) to clear the item, but perhaps pause the reminder popup completely if *any* task (even a new one) is currently being tracked.

### Steps to Reproduce

1. Create a task which has due time in 1 minute
2. Ignore it for 2 minutes
3. See the overdue popup

### Console Output

```shell

```

### Documentation

cocojojo5213:
Fixes #8051.

## Summary
- Open the task reminder dialog with `disableClose: true` so backdrop clicks and Escape cannot passively close overdue reminders.
- Add ReminderModule regression coverage for the dialog config.
- Update the scheduled-task wiki note to describe the explicit reminder actions.

## Notes
- This intentionally leaves the tooltip cleanup mentioned in the issue comments for a separate PR.

## Validation
- `npm run test:file src/app/features/reminder/reminder.module.spec.ts` (6 tests passed; run twice after formatting)
- `npm run checkFile src/app/features/reminder/reminder.module.ts`
- `npm run checkFile src/app/features/reminder/reminder.module.spec.ts`
- `git diff --check`
- commit hook: changed-file check and `npm run lint` passed

I did not run the full `npm run test` suite locally because the full Karma/Chrome suite has been unstable in this environment; the targeted reminder module spec passed and CI should cover the full suite.

### DIFF
--- a/docs/wiki/2.05-Manage-Scheduled-Tasks.md
+++ b/docs/wiki/2.05-Manage-Scheduled-Tasks.md
@@ -37,6 +37,8 @@ The task is scheduled immediately and the menu closes. To pick any date or set a
 2. In **Remind at**, choose when you want to be reminded. Options in the app are: **Never**, **when it starts**, **5 minutes before it starts**, **10 minutes before it starts**, **15 minutes before it starts**, **30 minutes before it starts**, **1 hour before it starts**.
 3. Click **Schedule**.
 
+When a task reminder dialog appears, choose an explicit action such as snooze, unschedule, dismiss the reminder, add to today, start, or done. Clicking outside the dialog or pressing Escape does not dismiss the reminder.
+
 ## Remove Scheduling from a Task
 
 **From the context menu (if the task is already scheduled):** Right-click the task. Next to the calendar icon there is an **Unschedule task** button (event_busy icon). Click it to clear the due date and time.
--- a/e2e/tests/reminders/reminders-deadline.spec.ts
+++ b/e2e/tests/reminders/reminders-deadline.spec.ts
@@ -94,7 +94,7 @@ test.describe('Deadline Reminders', () => {
     await expect(page.locator(REMINDER_DIALOG)).not.toBeVisible();
   });
 
-  test('should not reappear after ESC/backdrop dismissal of the deadline reminder dialog', async ({
+  test('should require an explicit action for the deadline reminder dialog', async ({
     page,
     workViewPage,
     testPrefix,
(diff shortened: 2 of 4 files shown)


johannesjo:
Thank you very much! <3

Issue #8052, opened 2026-06-06
💡 Feature Request: Add hover tooltips to action buttons in the Reminder Popup


hajiboy95:
### Problem Statement

In the task reminder popup, there are four action buttons aligned to the right of each task (Checkmark, Calendar, Snooze/Z, Sun). Currently, there are no hover titles or tooltips for these symbols. For newer users, it is not immediately clear what exact action each button performs (e.g., whether the Sun means "tomorrow morning" or "add to my day", or what the Calendar with the 'x' does).

**Additional Context**
This is a quick UX win that significantly improves accessibility and discoverability without cluttering the UI.

### Possible Solution

Add simple, descriptive hover tooltips (e.g., standard HTML `title` attributes or the app's standard tooltip component) to each of the action buttons which are displayed on hover.

**Examples of expected tooltips:**

* **Checkmark:** "Mark as done"
* **Calendar:** "Dismiss / Unschedule"
* **Z-Clock:** "Snooze"
* **Sun:** "Schedule for tomorrow / Add to today" -> Here's a little bug. Add to today is also displayed for tasks which are scheduled for today but overdue. Here we should add the "add for tomorrow" button maybe, or just a reschedule button instead which opens the datepicker would be better probably (even if a bit slower?). This one actually has a description on hover but it takes quite some time.

### Describe alternatives you've considered

* **Trial and Error:** Clicking the buttons just to see what happens.
* *Drawback:* This is risky and frustrating, as it often results in accidentally marking a task as complete or messing up its schedule, forcing me to go hunt down the task in the main view to undo the action.

* **Guessing based on experience:** Assuming the "Sun" means "Add to My Day"
* *Drawback:* Iconography isn't universal, so guessing often leads to the same frustrating trial-and-error loop 

### Additional context

In my opinion his is a quick UX win that significantly improves accessibility and discoverability without cluttering the UI.

johannesjo:
Thanks for opening this. I agree this is a good UX/accessibility improvement.

Small clarification: the row action buttons currently have native `title` attributes, but those are delayed, hover-only, and not very helpful for keyboard, screen reader, or touch users. I think we should replace/augment them with proper `matTooltip` labels and explicit `aria-label`s.

Suggested scope for a first PR:

- Checkmark: `Complete`
- Calendar/alarm-off: `Unschedule` for scheduled reminders, `Clear reminder` for deadline reminders
- Snooze: `Snooze`
- Sun: `Add to Today`

[...]

PR #8063, opened 2026-06-06
Add hover tooltips to reminder popup

hajiboy95:
## Problem

The reminder popup lacked visual feedback via tooltips for its action buttons, making the UI less intuitive. Additionally, keyboard navigation and focus management within the dialog were missing or brittle, and existing tests for focus restoration used unsafe `Object.defineProperty` on the global `document` object. (Ref: #8052)

## Solution

1. **Hover Tooltips:** Added `matTooltip` to action buttons (Snooze, Done, Open) in the task reminder popup (#8052).
2. **Keyboard Navigation:** Implemented focus management and keyboard support in the reminder dialog to allow full operation without a mouse.
3. **Refactoring Tests:** Updated `LayoutService` unit tests to use `spyOnProperty(document, 'activeElement')` instead of `Object.defineProperty` for safer and cleaner focus mocking.

## Type of Change

* [ ] Bug fix
* [x] New feature
* [ ] Breaking change
* [x] Refactoring
* [ ] Documentation
* [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for working on the reminder popup accessibility and keyboard flow. I reviewed the PR with a few focused passes and found a few issues that should be addressed before merge:

1. The icon-only row action buttons lost explicit accessible names when `[title]` was replaced with `matTooltip`.
   `matTooltip` adds an ARIA description, while `mat-icon` defaults to `aria-hidden`, so these buttons should get explicit `[attr.aria-label]` values matching the tooltip labels.
   https://github.com/super-productivity/super-productivity/blob/03484e7ee854b27411fe1b83a46f56fd37d03613/src/app/features/tasks/dialog-view-task-reminders/dialog-view-task-reminders.component.html#L217-L274

2. Arrow navigation can get stuck on disabled or hidden row actions.
   The handler collects all `.actions button` elements and focuses by index, but the dismiss/unschedule action can be disabled and also has `hide-xs`, which is `display: none` on small screens. In that case ArrowRight can call `preventDefault()` and then attempt to focus an unfocusable button, leaving focus where it was instead of moving to Snooze/Add to Today. Please filter to visible, enabled, focusable buttons.
   https://github.com/super-productivity/super-productivity/blob/03484e7ee854b27411fe1b83a46f56fd37d03613/src/app/features/tasks/dialog-view-task-reminders/dialog-view-task-reminders.component.ts#L481-L510

3. Focus restoration does not handle row menu actions.
   When a snooze/menu item removes a reminder, focus is in the Material overlay, not inside `.task`, so `_removeTaskFromList()` records no row/button fallback. Material then tries to restore focus to the trigger while the row may already be removed. Please pass/capture the triggering row/button context for menu actions too, or otherwise move focus after removing the row.
   https://github.com/super-productivity/super-productivity/blob/03484e7ee854b27411fe1b83a46f56fd37d03613/src/app/features/tasks/dialog-view-task-reminders/dialog-view-task-reminders.component.ts#L548-L590

4. `mouseover` should not move DOM focus.
   `button.focus()` on hover can steal focus from keyboard and screen-reader users just by pointer movement. Material tooltips already show on hover and focus, so this handler should be removed or changed to avoid changing focus.
   https://github.com/super-productivity/super-productivity/blob/03484e7ee854b27411fe1b83a46f56fd37d03613/src/app/features/tasks/dialog-view-task-reminders/dialog-view-task-reminders.component.ts#L454-L460

Test note: the new keyboard specs use a simplified template with only two enabled buttons and call `onKeyDown()` / `onMouseOver()` directly. That misses the disabled/hidden-button case above and does not verify the actual HostListener event path or handled-key `preventDefault()` behavior. Please add coverage using the production button shape, or at least include disabled/hidden actions and dispatch bubbling DOM events from focused buttons.

hajiboy95:
@johannesjo adressed. Regarding the dom behaviour, that actually was intended but you're right, probably too error prone. My intentiion was to actually prevent two popups to be displayed as once. but either way how once handles it (either touch dom or hide the popup) it'll be some new responsibility to keep it working so i just ommited that now. Thanks for that feedback.

johannesjo:
Thank you!

Issue #7616, opened 2026-05-15
🚨 18.5.0 breaks connection to Webdav


Ablesius:
### Version Used

18.5.0

### Current Behavior

When I opened the app after upgrading to 18.5.0, the connection to Webdav broke. There was an error message I unfortunately didn't copy, but afterwards, my connection settings were gone.

I was unable to reconnect with the new Nextcloud connection. I haven't tried with the "unsupported" Webdav connection now.

### Expected Behavior

I will say right off the bat that breaking something between minor versions is a huge no-go for me. If I didn't depend on SP's excellent features so much, I'd have probably moved to some other app by now. The way this is going, with upgrades breaking my workflow every other month, as it has been over the last couple of months now, it's a question of when, not if, I eventually do that.

What I actually expect: If you introduce a **breaking** change, no matter to what part of the software, I expect a _very_ clear message of what will or can happen. An error message like I received is _too late_ because by then, I can't take anything back.

What I'd have expected software-wise was migration instructions, or an option to migrate in the settings, with the default being that I'm still able to use my setup at first.

The wiki doesn't include a mention of this btw and needs to be updated accordingly.

### Steps to Reproduce

I cannot provide _good_ steps to reproduce but for a test installation you can do:

1. install SP any version before 18.5.0
2. set up Webdav syncing
3. create a couple example tasks that you can sync
4. sync them
5. close SP
6. upgrade to 18.5.0
7. open to SP

Tada, your workflow is broken!

(I selected "NO it only happened once" because I do not intend to try this again)

### Can you reproduce this reliably?

No, it only happened once.

### Console Output

```shell
(too late because I have already moved on from this, since I needed my program to work, even if the sync broke)
```

### Documentation

- [x] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [x] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

Ablesius:
Ah, by the way: The "Nextcloud" Sync didn't work. I was unable to set it up without a helpful message. I'll open another issue for that.

Edit: https://github.com/super-productivity/super-productivity/issues/7617

johannesjo:
Thank you! I look into it.

johannesjo:
Thanks for the detailed report!

[...]

PR #8064, opened 2026-06-06
docs(sync): document Proton Drive via rclone + WebDAV

### LINKED ISSUE #8008: 💡 Proton Drive syncing
### Problem Statement

The title says it all. Update 18.8.0 brought us OneDrive syncing, but Proton Drive is still missing. How come?

### Possible Solution

Add Proton Drive to the sync provider list, please. Thank you!

_No response_

### Additional context

_No response_

symonbaikov:
Document how to sync Super Productivity to **Proton Drive** on desktop using the existing **WebDAV** provider pointed at a local `rclone serve webdav` bridge, rather than a dedicated built-in provider.

This wires cross-references to the rclone + WebDAV guide ([`2.19-Sync-Proton-Drive-via-rclone`](docs/wiki/2.19-Sync-Proton-Drive-via-rclone.md)) into the sync docs and the wiki sidebar. Marked experimental and no-support, in line with how generic WebDAV is treated, since it relies on rclone's reverse-engineered Proton backend.

## Problem

#8008 asks for Proton Drive sync. Proton Drive has no official third-party API, and rclone's Proton backend is community-maintained and reverse-engineered — so a bundled built-in provider would add ongoing maintenance and security surface (spawning/managing a local `rclone` process). Documenting the supported WebDAV + rclone path avoids that while giving users a working setup.

## Solution

- `2.08-Choose-Sync-Backend.md` — note the advanced rclone + WebDAV Proton Drive recipe in the overview and "When to Use Which".
- `3.08-Sync-Integration-Comparison.md` — mention it under WebDAV "Advanced recipes" and in Related.
- `_Sidebar.md` — add the guide to the navigation.

Each reference clarifies this is **not** an officially supported built-in Proton Drive provider.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [ ] Other (please describe)

## Checklist

Closes: #8008

### DIFF
--- a/docs/wiki/2.08-Choose-Sync-Backend.md
+++ b/docs/wiki/2.08-Choose-Sync-Backend.md
@@ -18,6 +18,8 @@ In the **Sync provider** dropdown you will see:
 - **WebDAV (experimental)** — Sync via any other WebDAV-compatible server (e.g. ownCloud or a NAS). Labeled **experimental** because WebDAV implementations differ a lot between providers and may not work smoothly with every server. The option is not being removed.
 - **Local file (experimental)** — Saves the sync file to a folder on your machine (desktop or Android). Labeled **experimental** because behavior can depend on how you use that folder (e.g. multiple app instances at once). For many users it works fine. It is **not** being removed.
 
+Advanced users can also expose some other storage providers as WebDAV themselves. For example, Proton Drive can be used through a local rclone WebDAV bridge; see [[2.19-Sync-Proton-Drive-via-rclone]]. This is not a built-in Proton Drive provider.
+
 ## When to Use Which
 
 - **You want the newest sync stack and can accept beta** → **SuperSync** (self-hosted or hosted). Good if you want real-time sync and optional E2E encryption on the server.
@@ -25,6 +27,7 @@ In the **Sync provider** dropdown you will see:
 - **You already use Nextcloud** → **Nextcloud**. Use this instead of generic WebDAV unless you specifically need to enter the full WebDAV base URL yourself.
 - **You already use another WebDAV server** → **WebDAV**. Try it; if it works with your server, you’re set. If not, the experimental label reflects that WebDAV compatibility varies.
 - **You want backup or sync to a local/network folder (desktop or Android only)** → **Local file**. No account, no internet. Use for one device or backup to an external drive/NAS. **Do not** use tools like Syncthing or Resilio to sync that same folder between devices—they can cause conflicts. For multi-device sync, use WebDAV, Dropbox, or SuperSync instead.
+- **You use Proton Drive and accept an advanced setup** → Use **WebDAV** with a local rclone bridge. See [[2.19-Sync-Proton-Drive-via-rclone]]. This is not an officially supported built-in provider.
 
 ## Important: Local File and Syncthing/Resilio
 
@@ -34,5 +37,6 @@ Local file sync stores data in **one folder** on your device. The app warns: do 
 
 - [[1.02-Configure-Data-Synchronization]] — Quickstart: set up and configure sync from start to finish
 - [[2.09-Configure-Sync-Backend]] — Configure your chosen provider
+- [[2.19-Sync-Proton-Drive-via-rclone]] — Advanced Proton Drive setup through rclone and WebDAV
 - [[3.08-Sync-Integration-Comparison]] — Comparison table and details
 - [[4.23-Managing-Your-Data]] and [[3.06-User-Data]] — Where data and backups live, and how sync behaves
--- a/docs/wiki/3.08-Sync-Integration-Comparison.md
+++ b/docs/wiki/3.08-Sync-Integration-Comparison.md
@@ -35,6 +35,7 @@ Sync providers **synchronize your full Super Productivity data** (tasks, project
 - **Configuration:** Base URL, username, password, and an optional sync folder path on the server.
 - **Encryption:** Optional client-side encryption; you set an encryption key that is not sent to the server.
 - **Platform:** Available on desktop (Electron), web, and mobile. In the browser, CORS can block WebDAV requests; the app may recommend the desktop version for reliable sync.
+- **Advanced recipes:** Some storage providers can be exposed as local WebDAV with external tools. For Proton Drive via rclone, see [[2.19-Sync-Proton-Drive-via-rclone]]. This is not a built-in Proton Drive provider.
 
 ### Dropbox
 
@@ -70,6 +71,7 @@ Conflict resolution (e.g. when two devices change the same data) is the same for
 ## Related
 
 - [[4.24-Integrations]] — How issue and sync providers work
+- [[2.19-Sync-Proton-Drive-via-rclone]] — Advanced Proton Drive setup through rclone and WebDAV
 - [[3.07-Issue-Integration-Comparison]] — Issue provider comparison
 - [[4.23-Managing-Your-Data]] — Backups, import/export, and sync from a user perspective
 - [[3.06-User-Data]] — Where data and backups are stored, sync behavior by platform
--- a/docs/wiki/_Sidebar.md
+++ b/docs/wiki/_Sidebar.md
@@ -1,4 +1,5 @@
 <!-- pyml disable md041 -->
+
 ## [[1.00-Quickstarts]]
 
 - [[1.01-First-Steps]]
@@ -17,6 +18,7 @@
 - [[2.07-Manage-Task-Integrations]]
 - [[2.08-Choose-Sync-Backend]]
 - [[2.09-Configure-Sync-Backend]]
+- [[2.19-Sync-Proton-Drive-via-rclone]]
 - [[2.13-Run-with-Docker]]
 
 ### Contributing to Super-Productivity


johannesjo:
## Review — Proton Drive sync provider

Took a careful pass through the whole diff. The core design is sound and privacy-preserving: no Proton credentials are stored by the app, the rclone bridge binds loopback-only with random per-session basic-auth, the feature is consistently gated to `IS_ELECTRON` (factory / form dropdown / `isReady`), stderr is truncated and never logs user data, and the WebDAV reuse via `WebdavBaseProvider` + dedicated credential store mirrors `NextcloudProvider` cleanly. Good frontend test coverage too. CI is green.

A few things I'd suggest addressing before merge.

### 1. Unrelated formatting churn (~14 files) — please split out

Roughly half the touched files have nothing to do with Proton Drive and are pure prettier emphasis-marker / table-alignment rewrites (`*foo*` → `_foo_`, line rewraps):

`CLAUDE.md`, `docs/how-to-rate.md`, all `docs/plans/2026-*.md`, `docs/sync-and-op-log/operation-log-architecture.md`, `packages/super-sync-server/docs/*`, `packages/super-sync-server/prisma/migrations/README.md`, `packages/super-sync-server/public/terms.html`, `tools/prepare-appstore-release-notes.js`.

This accounts for essentially all the deletions in the diff, makes the feature harder to review, and will conflict with any in-flight doc PRs. Reverting these keeps the PR to the feature + its real docs (`docs/wiki/2.08`, `t.const.ts`, `en.json`, and the `src/`+`electron/` code). It looks like a local prettier-on-save with a different markdown config than the repo's — if a config bump genuinely requires it, it belongs in its own `chore(format):` commit.

### 2. `_ensureServer` has no concurrency guard → duplicate rclone processes + orphan leak

In `electron/proton-drive-rclone.ts`, `_server` is assigned only **after** `_startServer()` awaits `_waitForServeReady` (up to `SERVE_READY_TIMEOUT_MS = 20_000`). The reuse check is synchronous, so two calls that reach `_ensureServer` while `_server === null` both fall through, both `spawn` rclone, and the first child's reference is overwritten and never killed (cleanup only kills the current `_server`).

`getCfg()` → `_cfgOrError()` → `protonDriveEnsureServer` is invoked per WebDAV request, and there's a second independent IPC entry point — `protonDriveCheck` (Test Connection) — that is **not** behind the sync lock. Concretely reachable trigger: a user double-clicks "Test Connection" during the (up to 20 s) cold start, or clicks it while the first sync is mid-startup → two rclone `serve` processes, one orphaned until app quit.

Suggested fix: memoize the in-flight startup as a promise keyed by `{remoteName, binaryPath}` (or serialize `_ensureServer` through a single promise chain) so concurrent callers await the same start. Tracking the spawned child immediately (before "ready") would also fix the related quit-during-startup orphan.

### Minor

- **`protonDriveStopServer` is dead code.** It's exposed via `preload.ts` + `electronAPI.d.ts` and handled in main, but never called from `src/`. So the rclone serve starts on first sync and only stops on app `will-quit` — it keeps running (loopback WebDAV exposing the **whole** Proton Drive remote) even after the user switches provider or disables sync. Either wire `protonDriveStopServer()` into the provider-switch/disable path, or drop the unused IPC.
- **Auth creds on the command line.** `--user`/`--pass` are visible in the process table to other local processes; combined with `rclone serve webdav <remote>:` serving the entire remote root, a co-located process could read them off argv and reach the full Proton Drive over loopback. Low-risk on single-user desktop, but passing them via env (`RCLONE_USER`/`RCLONE_PASS`) instead of argv is cheap hardening.
- **`stdout` pipe is never consumed.** `stdio: ['ignore', 'pipe', 'pipe']` pipes stdout but only `stderr` has a `data` listener; a chatty stdout could fill the buffer and block rclone. Use `'ignore'` for stdout.
- **`docs/wiki/3.08-Sync-Integration-Comparison.md`** got its table widened (reformatting) but Proton Drive was **not** added as a row — that's the one comparison doc where it belongs (2.08 documents it correctly).

Happy to send a patch for the `_ensureServer` guard if useful.

johannesjo:
## Re-review — Proton Drive sync provider (multi-agent pass)

Re-reviewed the updated head (`b649be8`) across six lenses (correctness, security, architecture, alternatives, performance, simplicity). The core design is sound and the two follow-up commits resolved everything from the first review. **No critical issues** — the items below are polish + a couple of cheap hardening wins.

### Resolved since last review ✅
- Unrelated prettier churn reverted (no deletions remain outside the feature areas).
- `_ensureServer` concurrency guard added (verified correct for the realistic same-config case; new test covers it).
- `protonDriveStopServer` now wired into `_stopProtonDriveBridgeIfLeaving` (no longer dead).
- stdout switched to `'ignore'`.
- Proton Drive added to the 3.08 comparison doc.

### Verified non-issues (so they don't get re-raised)
- **Enum forward-compat:** adding `SyncProviderId.ProtonDrive` is safe — `syncProvider` is nulled on upload and the provider sub-configs aren't persisted into synced `globalConfig`, so the new value never reaches typia-validated synced data (same as Nextcloud/OneDrive).
- Loopback binding is consistently `127.0.0.1` (addr + port probe + readiness fetch).
- `_assertRemoteExists` runs only on the cold/restart path, not on warm reuse.
- Provider is dynamic-imported under `IS_ELECTRON`; no electron/rclone code reaches the web bundle.

### Worth addressing

**1. "Test Connection" starts a long-lived rclone server and never stops it.** `PROTON_DRIVE_CHECK` is `await _ensureServer(args); return true` (`electron/proton-drive-rclone.ts:315`), and `_testProtonDriveConnection` only shows a snack. So a single Test Connection click spawns a real `rclone serve webdav` exposing the **whole** Proton Drive over loopback for the rest of the app session — even if the user never enables sync (and `_stopProtonDriveBridgeIfLeaving` won't fire if they were never *on* ProtonDrive). Suggest tearing down a server `check` started unless it's the active reused one.

**2. Basic-auth creds on the rclone argv** (`--user`/`--pass`, `proton-drive-rclone.ts:187`) are world-readable via `ps`/`/proc/<pid>/cmdline` on shared hosts. Cheap strict improvement: pass via `env: { ...process.env, RCLONE_USER, RCLONE_PASS }` instead (rclone honors these; `/proc/.../environ` is owner-restricted). Worth verifying against a real rclone first.

**3. Readiness wait doesn't fast-fail on child death.** `_waitForServeReady` polls PROPFIND for the full 20s even when the spawned child already exited (e.g. `serve webdav` fails at runtime after `listremotes` passed). Race the probe against the child's existing `exit`/`error` events to fail in ms.

**4. 20s readiness timeout (`SERVE_READY_TIMEOUT_MS`) may be too short** for Proton's documented slow cold-connect on first use → false "did not become ready in time" failures. Consider ~45–60s (or only for the first connect). Worth a real-device check.

### Smaller cleanups
- **Drop the no-op `clearAuthCredentials` override** (`proton-drive.ts:75`) + its test. The WebDAV base *deliberately* omits it (#7616: don't wipe a user-typed password on a recoverable 401) and `ProviderManager` already skips when absent — so the override is behaviorally redundant, and its comment ("Override the WebDAV base which would blank userName/password") is factually wrong (the base blanks nothing). Deleting matches WebDAV/Nextcloud.
- **Replace the `as unknown as` casts with a real `PROVIDER_ID_PROTON_DRIVE` literal.** The class comment says it mirrors `NextcloudProvider`, but Nextcloud uses **zero** casts — it has its own id literal and the base's `WebdavProviderId` union is widened to include it. Doing the same here removes all the casts (including the credential-store cast — the one spot a mismatch would silently store under the wrong key). At minimum, fix the misleading comment.
- **Scope `rclone serve` to the sync subfolder** (`<remote>:<syncFolderPath>`) instead of the whole remote, to shrink blast radius (needs dir-exists + relative-path wiring).
- **stderr drain:** the per-chunk `child.stderr.on('data', () => log(...))` writes a log line per chunk for a constant string — since the content is intentionally discarded, switching `stdio[2]` to `'ignore'` and dropping the handler is simpler.

### Accepted as-is (noted, not blocking)
`_findFreePort` TOCTOU port race (loopback, μs-wide, clean failure; the `--addr :0` + parse-stderr alternative is more brittle); ~5 `ensureServer` IPC round-trips per upload (warm path is cheap, not user-perceptible).

**Verdict: approve with minor changes.** Items 1 and 2 are the most worthwhile before merge; the two cleanups (`clearAuthCredentials`, casts) are low-risk consistency wins.

<sub>🤖 Generated with multi-agent review (6 independent reviewers). Findings verified against current source where possible; the PR wasn't applied to the review worktree, so deep-dive claims used the diff + file snapshots + the real unchanged base classes.</sub>

johannesjo:
Can you fix the merge conflicts @symonbaikov ?

symonbaikov:
@johannesjo Yes. Already fixing them.

johannesjo:
## Re-review — Proton Drive sync provider

Thanks @symonbaikov — the last two commits address everything from both prior reviews. Re-verified against the current head; CI is green and merge conflicts are resolved.

### Resolved ✅
- **Test Connection** now tears down the server it started (`shouldStopAfterCheck` + `_stopServerIfMatching` in `finally`) instead of leaking a long-lived `rclone serve` — with a test.
- **Basic-auth creds** moved off argv to `RCLONE_USER`/`RCLONE_PASS` env, so they no longer show up in `ps`/`/proc/<pid>/cmdline`.
- **Readiness fast-fails** on child exit/error (`_waitForServeReady` races the PROPFIND poll against the child's `exit`/`error`) instead of polling the full timeout.
- **Readiness timeout** raised to 60s for Proton's slow cold-connect.
- **Real concurrency guard** (`_startingServer` memoized by `{binary, remote, folder}`) — concurrent same-config callers share one startup, so no duplicate/orphaned rclone processes.
- **`protonDriveStopServer`** wired into `_stopProtonDriveBridgeIfLeaving` (fires on provider-switch / sync-disable) — no longer dead.
- **stdout/stderr** both `'ignore'` — no unconsumed pipe, no per-chunk logging.
- **`clearAuthCredentials`** no-op override dropped.
- **Casts cleaned up** — real `PROVIDER_ID_PROTON_DRIVE` literal + `WebdavProviderId` union widening (mirrors Nextcloud), `as unknown as` id/credentialStore casts gone, plus a compile-time id guard.
- **`rclone serve` scoped** to the configured sync subfolder (`remote:folder`) with path-traversal validation + tests.
- **3.08 comparison doc** row added.

The 11 new electron tests map 1:1 to these fixes — nicely done.

### Small follow-ups before merge

1. **Leftover `as unknown as` casts.** `proton-drive.ts` still casts in `isReady()` and `_cfgOrError()`: `(await this.privateCfg.load()) as unknown as ProtonDrivePrivateCfg | null`. Now that the base is parameterized with `ProtonDrivePrivateCfg`, `load()` already returns `ProtonDrivePrivateCfg | null` (Nextcloud uses **zero** cast here). `as unknown as` defeats type-checking, so a future cfg-shape drift would go unnoticed — please make it plain `const cfg = await this.privateCfg.load();`. The trailing `as WebdavPrivateCfg` in `_cfgOrError` is likely droppable too. These are leftovers from the old `typeof PROVIDER_ID_WEBDAV` generic.

2. **Unrelated formatting churn — please revert.** Two files are pure prettier reformatting with nothing to do with Proton Drive (same class as the churn reverted earlier):
   - `eslint.config.js` (collapses the multi-line `ignores: [...]` to one line)
   - `electron/sync-path-resolver.test.cjs` (line-wrap rewrites of `assert.throws` / `mkdtemp`)

3. **Merge hygiene — rebase instead of merge.** The branch merged commits that aren't in upstream `master` yet (`#8410` Ukrainian `uk.json` +1565, which also touches a non-`en` locale, and `#8412 "Fix visuals"` → `_overwrite-material.scss`, `formly-slide-toggle.component.ts`). They currently show as part of this PR and inflate the diff. A rebase onto current `master` (rather than merging a fork master) would drop all of it so the diff is just the feature.

### Acceptable / device-only (noted, not blocking)
- **Test-Connection teardown race:** if a *first* sync starts a same-config server in the narrow window of a concurrent Test Connection, the `finally` teardown can kill the server the sync just adopted. It self-heals on the next request and needs testing-during-setup-while-first-sync-fires, so it's fine for single-user desktop — a one-line comment acknowledging it wouldn't hurt.
- **Device-only validations:** 60s readiness vs Proton's real cold-connect, PROPFIND against a not-yet-created sync subfolder, and rclone actually honoring `RCLONE_USER`/`RCLONE_PASS`. CI can't cover these.

**Verdict: approve with minor changes** — items 1 and 2 are the only real cleanups; 3 is the one I'd sort out before merge so the diff is just the feature.

johannesjo:
Thanks for this — it's a genuinely well-structured PR. The provider is wired into every touch-point correctly (enum, `PrivateCfgByProviderId`, `PROP_MAP_TO_FORM`, `FILE_BASED_PROVIDER_IDS`, factory, form, cleanup hooks), it carries real test coverage including a nice env-var-not-argv credential test, and it correctly stays transport-level so there's **no sync-correctness/op-log risk**. Credentials are handled carefully (random per-session basic-auth via env, never logged, Proton login left entirely to rclone). Nice work.

I ran a multi-perspective review. One item I'd consider a merge blocker, plus a few worth addressing.

### 🔴 Blocker

**Renderer-controlled executable path → arbitrary process spawn in the main process.**
`PROTON_DRIVE_ENSURE_SERVER` / `PROTON_DRIVE_CHECK` accept `rcloneBinaryPath`, `remoteName`, and `syncFolderPath` straight from the renderer per call, and `_resolveBinaryPath()` only `.trim()`s the path before it's handed to `execFile`/`spawn`. This contradicts the hardening from #8228: `electron/local-file-sync.ts` explicitly treats renderer-supplied paths as *"untrusted plugin/XSS"* and makes the sync folder path **main-owned** (persisted via `simple-store`) so a compromised renderer can't redirect it. Here it's an *executable* path, which is strictly worse — plugin background scripts run with ambient `window.ea`, so this is reachable without the `nodeExecution` consent gate.

Exploitation does require an existing renderer foothold (malicious plugin / XSS), so it's adding to an existing class rather than a brand-new one — but the fix is cheap and matches an established pattern: **persist `rcloneBinaryPath` (and ideally `remoteName`/`syncFolderPath`) main-side via a deliberate settings action and read them main-side, instead of trusting per-call IPC args.** Mirror `local-file-sync.ts`.

### 🟡 Worth addressing

- **Decrypted data + live creds exposed to the renderer** (`_startServer` → `ProtonDriveServeInfo` returned over IPC). The loopback bridge serves *decrypted* Proton content over plain HTTP, and the basic-auth creds are returned to the renderer, so a plugin can `protonDriveEnsureServer()` and read the sync folder. The cred-to-renderer part mirrors how normal WebDAV already works, but the decrypted-loopback exposure is new — at minimum worth a documented multi-user-host caveat; ideally folds into the blocker fix.
- **"Test Connection" can kill an in-flight sync** (`PROTON_DRIVE_CHECK`, `shouldStopAfterCheck`). Background sync isn't paused while the settings dialog is open, so a Test that starts the bridge can have its `finally` stop a server a concurrent sync just reused → that cycle fails with connection-refused (self-heals next cycle; not data loss). Simplifying the check to a cheaper `rclone listremotes` probe would also sidestep this.
- **Readiness poll treats 401 as ready** (`_pollServeReady`: `if (res.status < 500) return`). Not triggerable today, but reject 401/403 so the probe actually validates auth.

### 🟢 Minor / optional

- `_findFreePort()` has a TOCTOU between close and rclone's bind; `rclone --addr 127.0.0.1:0` + parsing the chosen port from stderr avoids it (and would also enable race-free readiness, replacing the HTTP poll). Defer unless you're touching this area.
- `_ensureServerForRequest` is a bit over-built for a single-server invariant — the triple re-checks + `key` plumbing optimize a race one user toggling one provider won't hit (~25–30 lines reducible). Not blocking.
- The missing-sync-folder validation reuses `T.F.SYNC.FORM.WEB_DAV.S_FILL_ALL_FIELDS` ("fill in all **WebDAV** fields") in the Proton context — a neutral string would read better.
- `_testProtonDriveConnection` is the one Proton path without a `window.ea` guard (currently unreachable on web, but inconsistent with the rest of the feature).

Confirmed justified, no change needed: `rcloneBinaryPath` (real macOS truncated-PATH need), the `syncFolderPath: ''` blanking + `extraPath` interplay, the two config types and `_idCheck` idiom (both match existing provider conventions), and `protonDrive` being omitted from public synced config.

johannesjo:
First off, thank you — genuinely. This is a lot of careful work, and it shows: the main-process-owned binary picker, the loopback bridge with random per-session creds, the owner-refcounting so a connection test doesn't tear down an active sync, and the tests on both the Electron and frontend sides. This is well above the bar. So I want to be upfront that what follows isn't about code quality, and I'm sorry it's landing after you've put this much in — that's on me for not flagging direction sooner.

My hesitation is about what a built-in sync provider commits us to long-term. **Every provider in the sync dropdown looks first-class, so users reasonably assume we own it end to end.** WebDAV is the cautionary tale: it's by far our biggest support sink. The overwhelming majority of "sync is broken" reports are actually someone's server config, CORS, auth quirks, or a non-standard implementation — almost none of it our code. But we can't debug other people's servers, so those threads go in circles, and a chunk of them turn into one-star reviews and "this app's sync is broken" posts written by people whose *third-party setup* failed. I don't want to knowingly take on a second one of those.

And Proton-via-rclone would likely be a *worse* version of that, through no fault of your implementation:

- It rides rclone's **reverse-engineered** `protondrive` backend, which Proton doesn't sanction and which can break the moment Proton changes their API or crypto — and the report lands on *us*.
- It needs **command-line setup** (`rclone config`, 2FA, mailbox password, PATH, version drift) — a big surface of "it doesn't work" unrelated to SP.
- It's a **four-layer bridge** (app → local WebDAV → rclone → Proton); every extra link is one more thing that fails and gets attributed to us.

So I don't think I can take this in as a new branded provider in core.

**But I really don't want your work wasted — and the good news is the capability already works without a new provider.** What you've built is, at heart, "WebDAV + a local rclone bridge." A user gets the identical result today with the existing WebDAV provider:

```sh
rclone serve webdav protondrive: --addr 127.0.0.1:8585 --user sp --pass "strong-password"
```

…then point the **WebDAV** provider at `http://127.0.0.1:8585`. (In fact your `ProtonDriveProvider` already points the WebDAV transport at `http://127.0.0.1:<port>/` internally — so this is exactly the same path, just with the user owning the `rclone serve` lifecycle.) That keeps Proton sync on the WebDAV rails we've *already* disclaimed as experimental/no-support, and adds nothing Proton-specific for us to maintain.

So what I'd love, if you're up for it: **land this as a wiki page** capturing the setup knowledge you already worked out. I drafted one to make the offer concrete — it reuses your `rclone config` walkthrough and adds the WebDAV steps:

<details>
<summary>Draft: <code>docs/wiki/2.19-Sync-Proton-Drive-via-rclone.md</code></summary>

````markdown
# Sync to Proton Drive via rclone + WebDAV

This guide shows how to sync Super Productivity to **Proton Drive** on desktop. Proton Drive has no official third-party sync API, so this works by running a small local bridge — [rclone](https://rclone.org/protondrive/) exposing your Proton Drive as a local WebDAV endpoint — and pointing Super Productivity's existing **WebDAV** provider at it. There is no dedicated "Proton Drive" option in the app; you configure the **WebDAV** provider against a local address.

> **Read this first — experimental, no support.** This setup relies on rclone's **community-maintained, reverse-engineered** Proton Drive backend. Proton does not officially support third-party access, and the backend can break whenever Proton changes its API or encryption. Like generic WebDAV sync (see [[2.08-Choose-Sync-Backend]]), this path is **experimental and provided as-is, without support**. Most problems come from the rclone/Proton side or your local setup — please troubleshoot those with the [rclone community](https://forum.rclone.org/) and **do not report them to Super Productivity as bugs**. If you want fully supported cloud sync, use **Dropbox** or **SuperSync** instead. Always keep backups (see [[2.02-Restore-Data-From-Backup]]).

**Desktop (Electron) only.** The bridge needs a local rclone process, so this does not work in the web app or on mobile.

## How It Works

```txt
Super Productivity  →  WebDAV provider  →  http://127.0.0.1:<port>  →  rclone serve webdav  →  Proton Drive
```

Your Proton login (including 2FA and mailbox password) lives only in rclone's own config. Super Productivity never sees or stores your Proton credentials — it only talks to the local WebDAV endpoint.

## 1. Install rclone

[Install rclone](https://rclone.org/downloads/) and make sure it is on your system `PATH` (or note the full path to the executable). Verify it works:

```sh
rclone version
```

## 2. Create a Proton Drive remote

In a terminal, run:

```sh
rclone config
```

Create a new remote of type **protondrive**. The prompts walk you through your Proton login, including 2FA and (if you have one) your separate mailbox password. Give the remote a name — `protondrive` is a good default.

Confirm it is configured:

```sh
rclone listremotes
# should include: protondrive:
```

## 3. Start the local WebDAV bridge

Run rclone as a local WebDAV server bound to loopback, with a **fixed** port and credentials you choose (so you can configure the app once):

```sh
rclone serve webdav protondrive: --addr 127.0.0.1:8585 --user sp --pass "choose-a-strong-password"
```

Notes:

- `--addr 127.0.0.1:8585` binds to **loopback only** — the bridge is not reachable from other machines. Pick any free port.
- The `--user` / `--pass` here are local-only credentials for the bridge; they are unrelated to your Proton account. Use a strong value anyway.
- Leave this process running while you sync. To keep it running across reboots, set rclone up as a background service / autostart entry — see the [rclone serve docs](https://rclone.org/commands/rclone_serve_webdav/) and your OS's service manager.

## 4. Configure the WebDAV provider in Super Productivity

1. Open **Settings** → **Sync & Backup** → **Sync**.
2. Set **Sync provider** to **WebDAV**.
3. Fill in:
   - **WebDAV Base URL**: `http://127.0.0.1:8585`
   - **Username**: the `--user` value (e.g. `sp`)
   - **Password**: the `--pass` value
   - **Sync Folder Path**: a folder to store sync files in, e.g. `/super-productivity`
4. (Recommended) Set a **client-side encryption key** so your sync files are encrypted at rest.
5. Click **Test Connection**, then enable sync.

That's it — Super Productivity now syncs to Proton Drive through the local bridge, using the standard WebDAV pipeline.

## Keeping It Running

The bridge must be running whenever the app syncs. If `rclone serve` is not running, sync will fail to connect (this is expected). Common approaches:

- Start it manually before you work.
- Run it as a background service (systemd user service on Linux, a Launch Agent on macOS, a scheduled/startup task on Windows).

If you change the port, credentials, or remote, update the WebDAV settings to match.

## Encryption

Proton Drive stores your data end-to-end encrypted on Proton's side. The hop from the app to the local bridge is plain HTTP, but it never leaves your machine (loopback only). Enabling Super Productivity's **client-side encryption key** additionally encrypts the sync file contents before they reach rclone — recommended.

## Troubleshooting

- **"Connection failed" / cannot reach the server** → make sure `rclone serve webdav …` is still running and the Base URL/port match.
- **Login, 2FA, mailbox password, or "remote not found" issues** → these are rclone/Proton concerns. Re-run `rclone config`, test with `rclone lsd protondrive:`, and ask on the [rclone forum](https://forum.rclone.org/). These are **not** Super Productivity bugs.
- **Sync worked, then suddenly stopped after a Proton change** → the reverse-engineered backend may need an rclone update. Update rclone first.
- **Uploads fail or files look truncated** → some cloud backends upload more reliably with a write cache. Add `--vfs-cache-mode writes` to the `rclone serve webdav` command and restart the bridge.

## Related

- [[2.08-Choose-Sync-Backend]] — Overview of all sync providers and their support status
- [[2.09-Configure-Sync-Backend]] — Configuring the WebDAV provider in detail
- [[3.08-Sync-Integration-Comparison]] — Side-by-side provider comparison
- [[2.02-Restore-Data-From-Backup]] — Restoring if something goes wrong
````

</details>

I'd also tweak the Proton bits you added to `2.08`/`3.08` so they describe this recipe (and link the new page) rather than a dropdown option — I have those edits ready. If there's later appetite for a small *community plugin* that just manages the `rclone serve` lifecycle and pre-fills the WebDAV URL, that's a fine future layer on top — but the doc unlocks it for people immediately.

Happy to help finish the page with you and link it prominently from the sync docs. Again, sorry for the late turn, and thank you for caring enough about Proton users to build it properly — I'd really like to land the *knowledge* even if I can't land the provider.

symonbaikov:
@johannesjo Thanks, that’s fair. I agree the branded provider makes the support boundary look stronger than it really is, given that this depends on rclone’s reverse-engineered Proton backend.

I’ll rework the PR into the docs-only approach you suggested: remove the Proton Drive provider/core integration, add a wiki recipe for `rclone serve webdav`, and link it from the sync docs/comparison as an advanced “use at your own risk” setup via the existing WebDAV provider.

I’ll keep the implementation branch around for reference, but won’t try to merge it as a built-in provider.

johannesjo:
Thank you very much for your understanding and sorry again. I should have looked into this more carefully earlier.

symonbaikov:
Reworked this as a docs-only PR following your suggested boundary.

The built-in Proton Drive provider/core integration is removed from the PR. The branch now documents Proton Drive as an advanced WebDAV setup using `rclone serve webdav`, with the support boundary called out explicitly: Proton credentials stay in rclone, the Proton backend is reverse-engineered, and this is not presented as an official built-in provider.

johannesjo:
Thank you very much @symonbaikov ! <3

PR #8067, opened 2026-06-06
feat(plugin): add markdown notes folder viewer

### LINKED ISSUE #8028: Plugin: display markdown notes from a local folder (Obsidian-style)
## Problem Statement
Several users in #2373 want to keep their notes as plain `.md` files in a folder (e.g. an Obsidian vault) and still see/use them inside Super Productivity — without manually duplicating notes into `.md` files. SP can already sync **tasks** to/from markdown via the `sync-md` plugin, but there is no equivalent for **notes**.

## Possible Solution
A plugin (sibling to `sync-md`) that points at a configured folder, reads all `.md` files in it, and displays them similar to the current Notes UI. Optionally bidirectional later, but read/display first.

## Implementation idea
- Model after `packages/plugin-dev/sync-md/` (already does folder watching + file read/write via the Electron file-sync layer) and `doc-mode` for in-app rendering.
- v1: **read-only** — list the folder's `.md` files and render them with the existing markdown renderer (`src/app/ui/marked-options-factory.ts`), shown in a notes-like panel.
- Folder ⇄ grouping: a subfolder could map to a project/section, mirroring the directory idea from #2373.

## Known blocker / dependency
The plugin Note API is currently **read-only**: plugins can read notes via `getAppState()`, but there is no note create/update/delete API (tasks have full CRUD, notes don't — `packages/plugin-api/`). Read-only display works today; writing notes back (or importing folder files as real Notes) needs the plugin Note API extended first. Any write path must respect the op-log / sync-correctness model the way task writes do.

## Scope
- ✅ Read + display `.md` files from a folder, Obsidian-compatible (no vendor lock-in).
- ❌ Not a graph/PKM engine — interop with existing PKM tools, don't reimplement them.

symonbaikov:
## Problem

Closed: #8028

## Solution

Adds a bundled read-only `Markdown Notes` plugin that displays local `.md` / `.markdown` files from a configured folder in a side panel.

The plugin:
- scans the selected local folder recursively via the desktop plugin Node execution API
- ignores hidden/support folders such as `.obsidian`
- groups notes by their containing folder
- renders a sanitized markdown preview
- supports local search by title and path
- stores the selected root folder and optional folder-to-project mappings in keyed plugin synced storage

The stored markdown configuration is device-specific even though it uses the existing synced plugin storage API: `rootPath` is an absolute local path, and `projectMappings` is keyed by absolute local directory paths. On another device, users may need to choose the local markdown folder again and adjust folder-to-project mappings.

To avoid collisions for duplicated Obsidian project folder names, note groups are identified internally by their absolute directory path rather than the leaf folder name. For example, `Folder A/Project Xy` and `Folder B/Folder C/Project Xy` are shown as separate groups and can be mapped independently.

This also extends the plugin project API with an optional read-only `folderPath` field, populated from the project tree where available, so plugins can label/disambiguate projects without exposing or syncing local markdown paths.
Screenshot From 2026-06-06 22-29-17

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thank you very much for this. Technical feedback will follow.

In terms of UX I would probably stick to a 1 column layout, or make it responsive in regards to panel width. I'd also keep the space used for context info minimal (1 line max) and configuration should probably done in a dialog or via the dedicated plugin form api for it.

johannesjo:
Thanks for this @symonbaikov — really nice, self-contained plugin and it matches what we landed on in #8028 (absolute-dir-path group identity, disambiguate-only folder paths, read-only/local-only v1). Tests cover the important bits and CI is green. A few notes from a closer pass, only one of which I'd consider worth addressing before merge.

## Worth fixing: `isSafeUrl` protocol allow-list can be bypassed with control chars

In `src/ui/app.ts` (`isSafeUrl`), a URL with a tab/newline before the colon defeats the explicit-protocol check:

```js
/^[a-z][a-z0-9+.-]*:/i.test('java\tscript:alert(1)')  // => false  (regex stops at the tab)
new URL('java\tscript:alert(1)', base).protocol        // => 'javascript:'  (WHATWG parser strips the tab)
```

Because `hasExplicitProtocol` is `false`, the function takes the "relative URL → allow" branch and **keeps** the `href`. I reproduced this directly:

```
"javascript:alert(1)"    => keptAsHref? false
"java\tscript:alert(1)"  => keptAsHref? true   URLproto: javascript:
```

This matters here because the plugin iframe runs `sandbox="allow-scripts allow-same-origin"` **and** holds `nodeExecution`, so script execution in this frame = same-origin access to app data + local code execution. In practice the classic `javascript:` click is probably neutralized today, since the sanitizer forces `target="_blank"` on every `<a>` and Chromium blocks `javascript:` navigations into a new browsing context — but the sanitizer's own logic is defeated, and I'd rather not depend on that browser quirk for a frame with this much reach. Prereqs also bound it: the user has to enable this (non-auto-enabled) plugin, open an **untrusted** note, and click a crafted link.

The fix is a one-liner — validate the *resolved* protocol unconditionally instead of branching on the regex:

```js
try {
  const url = new URL(trimmed, 'https://local.invalid');
  return allowedProtocols.has(url.protocol);
} catch {
  return false;
}
```

I verified this rejects the smuggled value while keeping every legitimate case (`https:`, `mailto:`, relative `notes/foo.md`, `#anchor`, `file:`/relative images) and leaves `data:` images blocked as before. As a broader option, bundling DOMPurify instead of the hand-rolled walker would also remove a class of these.

## Minor (fine as follow-ups)

- **Remote images auto-load** — `IMAGE_PROTOCOLS` allows `http(s):`, so opening an untrusted note silently beacons to any remote `` (reveals "note opened" + IP), no click needed. Consider defaulting to local/relative images only, or making remote images opt-in.
- **Auto-refresh has no overlap guard** — the 30s interval and the `focus` listener both call `refresh()` without checking `state.isLoading`, and each refresh re-reads the *full content of every file* through the node bridge. On a large vault a scan can approach the 30s timeout while the next tick already fires. An `if (state.isLoading) return;` guard (and ideally mtime-based incremental rescans later) would help.
- **Full re-render on note click** — selecting a note runs full `render()` (new `innerHTML` + re-`marked.parse` + rebuild all groups). You already have `renderContent`; updating just the preview pane on selection would be lighter for big folders.
- **`folderPath` contract** — it's populated only to disambiguate duplicate project titles, so it's `null` for a uniquely-named project that *is* in a folder. That's intentional, but the API JSDoc ("folder display path when available") might surprise plugin authors — worth a one-line clarification.
- **Docs** — the doc changes cover the new `folderPath` field but not how to enable/use the plugin itself; a short wiki entry (like doc-mode/sync-md) would be nice.

## Nice touches
Structured validation of the node-bridge output (`toScanResult`/`isMarkdownNote`), `rootPath` passed via `args` (no script-string injection), symlinks skipped (no walk loops), config kept in **un-synced** `localStorage` so local paths stay off the sync channel, and the `plugin-iframe.util` reorder correctly handles a script at offset 0 while keeping the `</body>` fallback.

johannesjo:
Thanks for this — it's a well-structured plugin: clean `core/` (pure, unit-tested) vs `ui/` split, faithful to the `doc-mode` bundled-plugin conventions, and the privacy model is sound (`folderPath` is the in-app menu-tree display path, computed at read time, never persisted or synced — the local markdown root stays in plugin `localStorage` only). I ran a multi-angle review (correctness, security, architecture, performance, simplicity). Summary below.

### Security context
The plugin holds `nodeExecution` **and** renders the content of local `.md` files as HTML, so any XSS in the preview reaches `PluginAPI.executeNodeScript` = code execution. I went after the custom `sanitizeHtml()` hard (mXSS via svg/math/foreignObject namespace confusion, `<style>`/`<noscript>`/`<xmp>` reinterpretation, foster-parenting, attribute breakouts) and could **not** find a bypass — it walks all descendants with a strict per-tag attribute allowlist that admits no `on*` handler, and `url-safety.ts` rejects `javascript:`/`data:`/`vbscript:`/`blob:` via the URL parser. So the RCE chain is real but currently closed. Given the blast radius, see the hardening suggestions below.

## Should fix before merge

**1. `markdown-notes` is missing from `BUNDLED_PLUGIN_IDS`** — `src/app/plugins/plugin.service.ts:64`
It's added to `BUNDLED_PLUGIN_PATHS` but not to the reserved-ID set (`doc-mode` is in both). As a result, a user-uploaded plugin can claim `id: "markdown-notes"` without tripping the `PLUGIN_ID_RESERVED` guard → identity spoofing and collision with this plugin's storage namespace (`markdown-notes-config-v1` + plugin meta/user persistence). The `nodeExecution` self-grant is blocked separately, so this isn't RCE — but it's a one-line fix. Please add `'markdown-notes'` to the set.

**2. Performance for large vaults** — `src/core/scan-script.ts` + `src/ui/app.ts`
This is the biggest concern for an Obsidian-vault viewer:
- The scan reads **every** file's full content into one array (`scan-script.ts`, `fs.readFileSync`), shipped across two structured-clone boundaries (Node→host IPC, host→iframe `postMessage`), with no per-file/total/depth cap.
- It re-runs on a 30s `setInterval` **plus** every `window.focus` **plus** `visibilitychange`, each doing a full re-read + a `getAllProjects` round-trip. `refreshIfIdle` only blocks *concurrent* runs, not back-to-back ones — so alt-tabbing re-scans the whole vault each time.
- Search filters on every keystroke, un-debounced, re-lowercasing every note's full content (`[title, relativePath, content].join('\n').toLowerCase().includes(q)`) — O(total vault size) per character.

Highest-leverage fix: have the scan return **metadata only** and lazy-load a note's content on selection. That shrinks the IPC/clone payload, makes mtime-based incremental re-scans cheap, and bounds the search index. Plus: debounce search + precompute a lowercased index, and throttle the focus/interval triggers.

**3. Auto-refresh steals input focus** — `src/ui/app.ts`
A scan completing while you're typing calls full `render()` (`appEl.innerHTML = …`), wiping focus and caret in the search/path field. Splitting a one-time "shell render" from the per-data "content render" (the `renderContent()`/`renderPreview()` paths already exist) fixes this and helps #2.

## Suggestions (non-blocking)

- **Sanitizer hardening / coverage.** The hand-rolled sanitizer held up, but it parse→mutate→re-serializes once (no stabilization loop) and unwraps `svg`/`math`/`foreignObject` rather than handling their namespaces — exactly what DOMPurify is built to own, and it bundles cleanly via esbuild (inlined, no CSP/network cost). Even if you keep the bespoke version, please **add a spec for `sanitizeHtml()` itself** — it's the one security-critical function without a test.
- **`file://` image host.** `url-safety.ts` allows any `file:` URL, including `file://host/share`; on Windows that can trigger an outbound SMB/NTLM-hash leak (the class hardened in `is-external-url-allowed.ts` for `openExternal`). Reject `file:` URLs with a non-empty host, or drop `file:` images for v0.1.0.
- **Folder→project linking.** The per-group project `<select>` stores `projectMappings`, but nothing reads the mapping back (no navigation/filter/action) — `group.project` is computed and never used. If there's no consumer planned for this release, consider deferring it: that also removes the `folderPath` plugin-API addition and the `getAllProjects` permission, leaving a tighter read-only viewer. If it stays, define a canonical separator for the public `folderPath` field instead of round-tripping the menu's ` › ` glyph (`app.ts` does `.replace(/ › /g, ' / ')`).
- **Iframe injection regex** — `plugin-iframe.util.ts:576`: injecting before the first `/<script\b/i` match is correct for this plugin and all currently-shipped ones, but the regex matches `<script` anywhere (e.g. inside a comment or `<pre>` text), so it's a latent trap for future iframe plugins. Worth restricting to a real tag start.
- **Minor:** GFM task-list checkboxes are stripped (`input` ∈ `DANGEROUS_HTML_TAGS`); symlinked notes are silently skipped (`readdirSync` lstat semantics — fine, no loop risk, just won't appear); `app.ts` is 565 lines mixing five concerns (consider splitting); and the `eslint.config.js` / `electron/sync-path-resolver.test.cjs` changes are prettier-only reformatting unrelated to the feature — better as a separate `style:` commit or dropped.

Overall: solid work, **needs changes** — #1 is the only real blocker; #2/#3 matter a lot for the large-vault case this targets; the rest are hardening/cleanup.

johannesjo:
**Review of `793e48d` (harden markdown notes viewer)**

Nice hardening pass overall — the DOMPurify migration, lazy on-demand content loading with a 1 MiB cap + path-traversal guard, the reserved-`markdown-notes`-id collision check, and the `readSeq` race guard are all solid and well-tested. A few things worth addressing before merge:

**1. ⚠️ The config is now synced across devices — please confirm this is intended**

This commit switched config storage from `localStorage` to `api.persistDataSynced` / `loadSyncedData`. That path dispatches `upsertPluginUserData`, and `pluginUserData` is a registered synced model (`src/app/pfapi/pfapi-config.js`), so the stored blob now flows through the op-log to other devices and the sync server.

The PR description says the root folder is stored *"only in plugin-local storage"* — that's no longer accurate. The blob is also inherently device-specific:
- `rootPath` is an absolute local path (`/home/user/Obsidian/...`)
- `projectMappings` is keyed by absolute directory paths

Consequences: absolute paths leave the device (note *content* is correctly not synced — only this config blob); on a second device the synced `rootPath` won't exist → scan error; and two desktops will LWW-fight over `rootPath`.

I realize the PluginAPI exposes no local-only persistence and `localStorage` is unreliable in the sandboxed iframe, so this is a constrained trade-off rather than an oversight. But please either (a) keep `rootPath` out of synced storage (re-prompt per device) or (b) update the PR description to state the config syncs and is device-specific.

**2. UTF-8 decode fails wholesale on a single bad byte**

`decodeUtf8` uses `decodeURIComponent(escape(binary))` with a `catch { return binary }` fallback. Any invalid UTF-8 byte anywhere — a latin1-encoded note, a stray byte, or a multi-byte char split at the 1 MiB/64 KB truncation boundary — throws and renders the *entire* content as latin1 mojibake. If `TextDecoder` is available in the nodeExecution sandbox, prefer it (it degrades per-char with U+FFFD); otherwise trim an incomplete trailing multi-byte sequence before decoding the truncated tail. Low severity but a real preview-corruption path.

**3. Search no longer matches content**

`filteredGroups` now searches only title + path (content is lazy-loaded). Reasonable given the new architecture, but the PR description still advertises *"search by title, path, and content"* — worth updating, and maybe noting in the UI.

**Minor / nits**
- `refreshSeq` staleness guard is effectively dead code: `refresh()` serializes on `isLoading` and only bumps `refreshSeq` when not loading, so `refreshId !== state.refreshSeq` can't be true mid-flight — the `requestedRootPath` comparison is the live guard. (The separate `readSeq` guard *is* needed and correct.)
- `createDOMPurify(windowRef)` is re-created on every `sanitizeMarkdownHtml` call; could be memoized per window. Not hot since preview is cached.
- `readTitleChunk` reads 64 KB per file on every 30 s auto-refresh — fine for normal vaults, just steady background I/O for very large ones (still an improvement over the prior full-file read).

Items 1 and 3 are mostly about reconciling the implementation with the PR description; 2 is the only one I'd consider a genuine (if low-frequency) bug.

symonbaikov:
Fixing it.

johannesjo:
## Review: `fix(plugin): harden markdown notes viewer`

Reviewed the hardening commit (`793e48d`). The security/memory work here is solid and well-tested — net clearly positive. A few things worth addressing, with one I'd ask to resolve before merge.

### Strengths
- **Sanitizer upgrade is real.** DOMPurify + an independent URL re-validation (`isSafeMarkdownLinkUrl`/`isSafeMarkdownImageUrl` via `new URL()` against a protocol allowlist) is proper defense-in-depth. The post-hoc `new URL()` check defeats control-char protocol bypasses (e.g. `java\nscript:`) that a regex denylist alone would miss, since the URL parser normalizes the scheme before the allowlist check. Good XSS coverage in the specs.
- **Path-traversal guard** in the read script (`path.resolve` → `path.relative` → `..`/absolute reject → `.md` check → file/dir stat) is correct and the outside-the-root case is tested.
- **Lazy loading** (scan reads only the first 64 KB for the title; selected note read on demand with a 1 MiB cap) avoids holding every note body in memory — a meaningful win for large vaults.
- **Race handling** (`refreshSeq`/`readSeq`/`queuedRefresh` + preview cache keyed by `path:mtime:size` with pruning) is careful and traces correctly.
- The `vm` sandbox-without-`Buffer` tests that mirror the real direct-nodeExecution executor are a nice touch.

### Findings

**1. [Please address] Synced config re-persisted every 30 s → op-log churn.**
`refresh()` calls `void saveConfig()` unconditionally on every successful scan, and the auto-refresh fires every 30 s while the panel is visible (`AUTO_REFRESH_MS`). Config now goes through `persistDataSynced`, and the host's `_encodeAndDispatch` dispatches `upsertPluginUserData` with **no "value unchanged" dedup** — so an op lands in the synced op-log every 30 s even when nothing changed. Previously this was free `localStorage` churn; now it's sync-wire traffic indefinitely. Suggest a dirty-check before saving (persist only when the config actually changed), or drop the save from the auto-refresh path.

**2. [Please address] Machine-local `rootPath` is now synced across devices.**
`rootPath` is a device-specific absolute filesystem path, but it's persisted in the same synced blob as `projectMappings`. Device B receives device A's path (nonexistent on B; meaningless on web/mobile where `executeNodeScript` is undefined). Combined with #1, two desktops with different roots ping-pong — each 30 s save LWW-overwrites the other's path. Cleaner split: keep `rootPath` **device-local** (localStorage) and only sync `projectMappings`. That also resolves #1's cross-device thrash. (If syncing the path is intentional for identical setups, a one-line note would settle it.)

**3. [Low] Body-text search regression.** `content` was removed from `MarkdownNote`, so search now matches only `title` + `relativePath`, not note bodies. Reasonable tradeoff for lazy loading, but it's a user-visible loss of full-text search — worth confirming it's intended.

**4. [Low] `decodeUtf8` boundary fallback yields mojibake on truncation.** When the 1 MiB read (or 64 KB title chunk) splits a multibyte UTF-8 sequence at the cut, `decodeURIComponent(escape(binary))` throws and the catch returns raw latin1 for the **entire** content — so a truncated UTF-8 file renders as garbage throughout, not just at the boundary. Minor (only >1 MiB multibyte files); trimming trailing incomplete bytes before decoding would avoid it.

**5. [Nit] DOMPurify instance created per render** (`createDOMPurify(windowRef)` on each call). Cheap next to a full parse and previews are cached, so optional — could memoize per window.

**6. [Nit] `ALLOWED_ATTR` is global, not per-tag.** DOMPurify now allows `align/alt/height/href/src/title/width` on any allowed tag (vs. the old per-tag scoping). Not a security issue — URI attrs are still re-validated and there's no script vector — purely cosmetic.

### Verdict
The hardening is well-executed. The main thing I'd resolve before merge is the synced-config behavior (#1 + #2): persisting a machine-local path to the op-log every 30 s is a sync-hygiene regression in a codebase that's deliberately strict about op-log pollution. Splitting `rootPath` (local) from `projectMappings` (synced) plus a dirty-check fixes both with a small diff.

PR #8069, opened 2026-06-06
feat(boards): enhance project selection with multi-select and sidebar…

hajiboy95:
## Problem

Tackles #7929 and #7249. Currently, board panels could only filter by "All Projects" or exactly one specific project. Users asked for more flexibility to filter columns by a custom subset of projects. 

## Solution

1.  **Multi-Project Schema:** Migrated `BoardPanelCfg` from `projectId: string` to `projectIds: string[]`.
2.  **Robust Migration:** Added logic to `sanitizePanelCfg` to automatically migrate legacy data and ensure the `projectIds` array is always valid.
3.  **Connected Multi-Select:** Refactored `SelectProjectComponent` to use checkboxes. Implemented "connected" logic where checking "All Projects" toggles the entire list, and manually selecting all individual projects automatically synchronizes the "All Projects" state.
4.  **Improved UI:** Added folder paths to dropdown options for duplicate project names and optimized the main text field to show only titles (or "All Projects").
5.  **Filtering Logic:** Updated `BoardPanelComponent` to support filtering by multiple IDs and handled "Move to Project" logic by defaulting to the first specific project in the list.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change 
    - [x] Schema migration
- [x] Refactoring
- [x] Documentation

## Checklist


johannesjo:
Thanks for working on multi-project board filters. I think this needs changes before merge:

1. Legacy board imports can lose project-specific panel filters. `_migrateBoards()` overlays `DEFAULT_PANEL_CFG` before the legacy panel (`src/app/op-log/backup/migrate-legacy-backup.ts:274-276`), so an old panel with only `projectId: "p1"` becomes `{ projectIds: [""], projectId: "p1" }`. `sanitizePanelCfg()` then only migrates when `projectIds === undefined` (`src/app/features/boards/boards.util.ts:23-25`), so runtime filtering reads `projectIds: [""]` and treats the panel as all-projects (`src/app/features/boards/board-panel/board-panel.component.ts:183-190`). Please make the sanitizer/migration prefer the legacy `projectId` in this mixed/defaulted shape and always remove `projectId`.

2. “All Projects” panels can assign inline-created tasks to the first project. The selector expands the all sentinel to `["", ...allProjectIds]` (`src/app/features/config/select-project/select-project.component.ts:101-108`), while filtering correctly treats any array containing `""` as unfiltered. But inline creation picks the first non-empty id (`src/app/features/boards/board-panel/board-panel.component.ts:134`) and writes it into new tasks (`:144`). Please either canonicalize any `projectIds` containing `""` back to `[""]`, or make `additionalTaskFields` ignore project IDs whenever `""` is present.

Please also add regression tests for:
- legacy/mixed `{ projectId, projectIds: [""] }` panel migration
- `projectIds` containing `""` plus real IDs
- multi-project filtering
- inline-create project assignment for all-project vs specific-project panels

hajiboy95:
@johannesjo adressed

johannesjo:
### Review of changes `a576188..0135c0a`

Reviewed the incremental delta (migration/canonicalization + `additionalTaskFields` guard + label translation). The migration logic is solid — pure, deterministic, idempotent, and replay-safe. One user-facing regression should be fixed before merge; the rest is polish.

#### 🔴 Must fix: "All Projects" shows the raw translation key

Switching `defaultLabel` from the literal `'All Projects'` to the key `T.F.BOARDS.FORM.PROJECT_ALL` in `boards-form.const.ts` exposes a bug in `SelectProjectComponent.triggerLabel()`, which returns the prop **without translating it**:

```ts
// select-project.component.ts (multiple mode, "All Projects" selected)
return this.to.defaultLabel || this._translateService.instant(T.G.NONE);
//     ^^^^^^^^^^^^^^^^^^^^ returned raw — only the fallback is translated
```

The dropdown *option* is fine (it pipes through `| translate`), but the collapsed `<mat-select-trigger>` — i.e. the default state of every new panel — now displays the literal `F.BOARDS.FORM.PROJECT_ALL`. With the old literal label this happened to render correctly, so this delta introduces it. One-line fix:

```ts
return this.to.defaultLabel
  ? this._translateService.instant(this.to.defaultLabel)
  : this._translateService.instant(T.G.NONE);
```

#### 🟡 Worth addressing

- **Document the lossy canonicalization.** `if (projectIds.includes('')) projectIds = ['']` in `sanitizePanelCfg` silently drops specific IDs when `''` co-occurs (`['','p1'] → ['']`). It's correct for the current UI, but the sanitizer is a generic reducer-invoked function with no knowledge of that UI invariant. Likewise, the legacy branch prefers a stale `projectId` over a `projectIds: ['']` default-overlay — a record genuinely meaning "All Projects" but carrying a leftover legacy `projectId` would be narrowed to that one project. Both are low-likelihood and the chosen directions are the safer ones — just please add a comment stating the intent and a test asserting the deliberate `['','p1'] → ['']` loss.

- **Redundant guard + test for an unreachable shape.** In `additionalTaskFields`, `isAllProjects ? undefined : projectIds.find(id => id \!== '')` only differs from plain `find(id => id \!== '')` on the mixed shape `['','p1']`, which canonicalization makes unreachable for the persisted cfg the component receives. The new test `['','p1','p2'] → undefined` builds exactly that unreachable shape. Either simplify to `projectIds?.includes('') ? undefined : projectIds?.[0]` and drop that test, or keep the guard with a "defensive against pre-sanitize values" comment.

#### 🟢 Optional / non-blocking

- The "All Projects wins" rule is duplicated across `sanitizePanelCfg`, `additionalTaskFields`, the `tasks` filter, and `drop()` in slightly different forms — a small shared helper (`isAllProjectsCfg()` / `firstSpecificProjectId()`) would prevent drift.
- `(out as any).projectId` → use a typed view like `out as BoardPanelCfg & { projectId?: string }` to honor the no-`any` rule (the file already uses `as Partial<BoardPanelCfg>` for the `delete`s).
- Design thought for the author: modeling "All Projects" as the **empty array** rather than the `['']` sentinel would let every special-case collapse to a length check and eliminate the lossy-canonicalization class entirely. Migration cost is about the same and the field is brand-new — cheapest time to decide. Non-blocking.

**No security or performance concerns** — Angular auto-escapes all the rendered labels/titles, corrupt non-array input degrades gracefully (the `\!Array.isArray → ['']` guard runs before any `.includes`), and the added per-task work is negligible (skipped entirely in the dominant "All Projects" case).

<sub>🤖 Reviewed with multiple parallel sub-agents (correctness, security, architecture, alternatives, performance, simplicity).</sub>

hajiboy95:
@johannesjo adressed 👍

johannesjo:

Issue #123, opened 2018-11-07
Add list of tasks


k4misiek:
## Your Environment

* Version used: 
* Operating System and version:
* Desktop Environment: 
* Browser Name and version: 

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

Issue #101, opened 2018-07-19
Play/pause buttons order


Acsigen:
## Your Environment

* Version used: 
* Operating System and version: Windows 7 X64
* Desktop Environment: 
* Browser Name and version: 

## Expected Behavior
Hit play button counter/task starts and display the pause button to indicate the next action
Hit pause button counter/task button stops and display the play button to indicate the next action

## Current Behavior
Press the pause button starts a task/timer
Press play button pauses a task/timer

## Steps to Reproduce (for bugs)
1. Press the pause button starts a task/timer
2. Press play button pauses a task/timer
3. This issue appears everywhere except indivitual task buttons and dropdown buttons from the top right corner
Task/timer active:
![button 2](https://user-images.githubusercontent.com/13287878/42933534-b03d1d3c-8b4d-11e8-8421-0caab63ead27.JPG)
![buttons](https://user-images.githubusercontent.com/13287878/42933535-b05a52f8-8b4d-11e8-9bc8-dbb2b4f90bd3.JPG)
Task/timer inactive:
![dropdown](https://user-images.githubusercontent.com/13287878/42933536-b07df6ae-8b4d-11e8-881c-514a3a978378.jpg)

johannesjo:
Hey there! Thanks opening up this issue. I remember that I was struggeling with that question myself. Looking at it seperatly the given choices felt better how they're arranged now. But I agreee that consistency might be good.

Acsigen:
In the dropdown menu I consider that there's no need for the play button to be displayed under the pause button (3rd image) because clicking the top button already works.

Issue #100, opened 2018-07-16
Task resumes automatically after Pomodoro break


Acsigen:
## Your Environment

* Version used: 
* Operating System and version: Windows 7 x64
* Desktop Environment: 
* Browser Name and version: 

## Expected Behavior
"Manually confirm continuing session after break" is checked and after the Pomodoro break I should be able to confirm to continue my work session

## Current Behavior
The session starts automatically even the option to confirm manually is checked

## Steps to Reproduce (for bugs)
1. Check "Manually confirm continuing session after break"
2. Start work session
3. Wait for Pomodoro break to end
4. Task starts automatically without notice

johannesjo:
Thanks for reporting. I will check the issue.

johannesjo:
A fix will be out with the next release. Thanks for reporting and btw.: Congratulations for opening up the 100th issue ;)

PR #8072, opened 2026-06-06
feat(redmine): allow searching issues by id

syky27:
## Summary

Redmine's search API (`/search.json`) only performs full text search and does **not** match issue ids, so typing an issue id (e.g. `22899`) into the add-task bar or issue panel search returned no results unless the number happened to appear in the issue text. (Redmine's own "#id jumps to issue" behavior is an HTML-only redirect in its web UI and is not available via the API.)

This PR makes numeric queries (`1234` or `#1234`) additionally fetch the issue directly via `GET /issues/{id}.json` and merge it into the search results:

- by-id result is shown first, de-duplicated against the text search results
- a failed by-id lookup (e.g. 404 for a number that isn't an existing issue id) silently falls back to plain text search results - no error snack while typing
- closed issues can now also be found by id (text search is restricted to open issues)

## Changes

- `redmine-api.service.ts`: merge by-id lookup into `searchIssuesInProject$` for numeric queries; add optional `isSkipErrorHandling` to `getById$`/`_sendRequest$` so speculative lookups don't show error snacks
- `redmine-issue-map.util.ts`: add `mapRedmineIssueToSearchResult`
- `redmine-api.service.spec.ts`: new spec covering text-only queries, by-id merge, `#`-prefixed queries, de-duplication and 404 fallback

## Test

- 6 new unit specs pass
- `npm run lint` passes

### DIFF
--- a/src/app/features/issue/providers/redmine/redmine-api.service.spec.ts
+++ b/src/app/features/issue/providers/redmine/redmine-api.service.spec.ts
@@ -1,0 +1,134 @@
+import { TestBed } from '@angular/core/testing';
+import {
+  HttpClientTestingModule,
+  HttpTestingController,
+} from '@angular/common/http/testing';
+import { RedmineApiService } from './redmine-api.service';
+import { RedmineCfg } from './redmine.model';
+import { SearchResultItem } from '../../issue.model';
+import { SnackService } from '../../../../core/snack/snack.service';
+
+describe('RedmineApiService', () => {
+  let service: RedmineApiService;
+  let httpMock: HttpTestingController;
+
+  const mockCfg: RedmineCfg = {
+    isEnabled: true,
+    host: 'https://redmine.example.com',
+    projectId: 'test-project',
+    api_key: 'test-api-key',
+    scope: null,
+  };
+
+  const mockSearchResponse = {
+    results: [
+      {
+        id: 100,
+        title: 'Bug #100: Some text issue',
+        url: 'https://redmine.example.com/issues/100',
+      },
+    ],
+    total_count: 1,
+    offset: 0,
+    limit: 100,
+  };
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      imports: [HttpClientTestingModule],
+      providers: [
+        RedmineApiService,
+        {
+          provide: SnackService,
+          useValue: jasmine.createSpyObj('SnackService', ['open']),
+        },
+      ],
+    });
+    service = TestBed.inject(RedmineApiService);
+    httpMock = TestBed.inject(HttpTestingController);
+  });
+
+  afterEach(() => {
+    httpMock.verify();
+  });
+
+  it('should be created', () => {
+    expect(service).toBeTruthy();
+  });
+
+  describe('searchIssuesInProject$', () => {
+    const searchUrl = (query: string): string =>
+      `${mockCfg.host}/projects/${mockCfg.projectId}/search.json?limit=100&q=${query}&issues=1&open_issues=1`;
+
+    it('should only send a text search request for non-numeric queries', () => {
+      let result: SearchResultItem[] | undefined;
+      service.searchIssuesInProject$('some text', mockCfg).subscribe((r) => (result = r));
+
+      const req = httpMock.expectOne(searchUrl('some%20text'));
+      expect(req.request.method).toBe('GET');
+      req.flush(mockSearchResponse);
+
+      httpMock.expectNone(`${mockCfg.host}/issues/100.json`);
+      expect(result?.length).toBe(1);
+      expect(result?.[0].title).toBe('Bug #100: Some text issue');
+    });
+
+    it('should additionally fetch the issue by id for numeric queries', () => {
+      let result: SearchResultItem[] | undefined;
+      service.searchIssuesInProject$('22899', mockCfg).subscribe((r) => (result = r));
+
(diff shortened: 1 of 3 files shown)


johannesjo:
Finding: the numeric Redmine lookup should stay project-scoped. `searchIssuesInProject$` used to only query `/projects/${cfg.projectId}/search.json`, but the new direct lookup calls `/issues/${issueId}.json`, which is global to the Redmine instance. For users with access to multiple Redmine projects, typing an id from another project can surface and add that issue through this provider. Could we use a project-scoped lookup instead, e.g. `/projects/${cfg.projectId}/issues.json` with `issue_id=${issueId}` and `status_id=*`, or otherwise verify the returned `issue.project` matches the configured project before merging it?

syky27:
hi @johannesjo that is not how redmine works... Redmine's issue IDs (`#123`) are the primary key of the issues table — a single, global, auto-incrementing sequence shared across every project. So if you create issue `#100` in Project A and the next issue in Project B, it'll be `#101`. There's no native setting to give each project its own counter starting at 1, because doing so would break the global uniqueness that the #NNN reference syntax, cross-project links, and relations all depend on.

johannesjo:
Thanks @syky27 — and you're right that ids are a single global sequence. That's actually the heart of the concern rather than a counter to it.

My worry isn't that two projects could share an id. It's that, *because* `GET /issues/{id}.json` resolves globally, a provider configured for Project A can now return (and add) an issue belonging to Project B — for any project the configured API key can see. Everything else in this service is project-scoped (`/projects/${cfg.projectId}/search.json`, `/projects/${cfg.projectId}/issues.json`), so the new by-id lookup is the one path that crosses that boundary.

Could we keep it scoped, reusing the pattern already in the file (`getLast100IssuesForCurrentRedmineProject$`)?

```ts
this._sendRequest$(
  {
    url: `${cfg.host}/projects/${cfg.projectId}/issues.json`,
    params: ParamsBuilder.create()
      .withParam('issue_id', String(issueId))
      .withState('*') // status_id=* → still finds closed issues
      .build(),
  },
  cfg,
  { isSkipErrorHandling: true },
)
```

This stays within the configured project, still surfaces closed issues via `status_id=*`, and lets Redmine resolve the project itself — so it works whether `projectId` is the numeric id or the identifier slug. (A post-fetch `issue.project` comparison wouldn't reliably handle the slug case, since the issue payload only carries the project's numeric `id` + `name`, not the `identifier`.)

A small spec asserting "an id from another project is not surfaced" would lock it in. Otherwise this looks great — thanks for the contribution\!

syky27:
Good catch — you're right that the global `GET /issues/{id}.json` was the one path crossing the project boundary, and that a slug `projectId` makes a post-fetch `issue.project` comparison unreliable. Fixed in `56b9b5c`.

The by-id lookup now reuses the project-scoped pattern from `getLast100IssuesForCurrentRedmineProject$`:

```ts
private _getIssueByIdInProject$(
  issueId: number,
  cfg: RedmineCfg,
): Observable<RedmineIssue | null> {
  return this._sendRequest$(
    {
      url: `${cfg.host}/projects/${cfg.projectId}/issues.json`,
      params: ParamsBuilder.create()
        .withParam('issue_id', String(issueId))
        .withState('*') // status_id=* → still finds closed issues
        .withLimit(1)
        .build(),
    },
    cfg,
    { isSkipErrorHandling: true },
  ).pipe(map((res: RedmineIssueResult) => res?.issues?.[0] ?? null));
}
```

This keeps everything inside the configured project, still surfaces closed issues via `status_id=*`, and lets Redmine resolve the project itself so it works for both numeric ids and identifier slugs. A cross-project (or unknown) id just comes back as an empty list → `null`, and the caller falls back to the text-search results — no error snack during type-ahead.

The global `getById$` is unchanged and still used by the existing refresh/poll path. Added a spec asserting an id from another project isn't surfaced (it also asserts no global `/issues/{id}.json` request is made, to lock out a regression). All specs pass.

syky27:
@johannesjo while I had this code paged in, I started on a small follow-up and
wanted to run it by you before opening a PR, since it touches the exact boundary
we just discussed.

Use case: I'd like a single Redmine connection that can search across *all* the
projects my API key can see, rather than configuring one provider per project.

Proposal: make the **Project Identifier optional**.
- **Project set** → behaviour is exactly as it is today, fully project-scoped
  (including the by-id lookup we just hardened). Your cross-project concern stays
  addressed — nothing changes for existing single-project users.
- **Project empty** → the provider operates in an explicit, opt-in global mode:
  `/search.json` and `/issues.json?issue_id=…` (no `/projects/<id>` prefix), so
  search-by-string and search-by-id both work instance-wide.

So global surfacing only ever happens when the user *deliberately* clears the
field — the accidental-leak case you flagged (a provider pointed at one project
quietly returning another's issues) can't happen.

I've implemented it on a branch off this one, with tests that keep all the
existing project-scoped specs and add global-mode coverage (incl. the inverse of
the "id from another project" spec). One open question I'd value your take on:
in global mode `testConnection()`'s empty-`q` search may behave differently
across Redmine versions — I'm planning to fall back to `/issues.json?limit=1`
when no project is set.

Would you prefer this as a separate follow-up PR (depends on #8072), or rolled
into this one? It is preped in my repo: https://github.com/syky27/super-productivity/pull/1 Happy either way.

johannesjo:
Hey @syky27 ! I'd appreciate a seperate followup! :)

johannesjo:

PR #8075, opened 2026-06-07
fix(core): refresh day stream after next-day start changes

### LINKED ISSUE #7720: 🚨 Restarting the app doesn't respect "Start time of next day"
### Version Used

v18.6.0

### Current Behavior

Current time: 10:05

1. Create a task
2. Schedule them for Today
3. Set "Start time of the next today" to 10:10
4. Restart the app
5. The task is already moved to Overdue
It should only be moved to Overdue when the time hits 10:10

This bug doesn't happen if "Start the of next day" is earlier than the current time

1. Create a task
2. Schedule them for Today
3. Set "Start time of the next today" to 10:04
4. Restart the app
5. The task isn't moved to Overdue

### Expected Behavior

-

### Steps to Reproduce

-

### Console Output

```shell

```

### Documentation

cocojojo5213:
Fixes #7720.

### Summary

- emit a DateService notification when the start-of-next-day offset changes
- have GlobalTrackingIntervalService resample todayStr() from that notification
- cover the restart/config-load case where 10:05 should still be the previous logical day when the next day starts at 10:10

### Root cause

todayDateStr$ is shareReplay(1) and can cache the calendar-day value created before the stored start-of-next-day config is applied during app startup. If the app restarts before the configured day boundary, date-change consumers can see the stale calendar day until the next timer/focus/visibility emission.

### Validation

- npm run checkFile src/app/core/date/date.service.ts
- npm run checkFile src/app/core/global-tracking-interval/global-tracking-interval.service.ts
- npm run checkFile src/app/core/global-tracking-interval/global-tracking-interval.service.spec.ts
- npm run test:file src/app/core/global-tracking-interval/global-tracking-interval.service.spec.ts
- npm run test:file src/app/core/date/date.service.spec.ts
- npm run test:file src/app/core/date/date.service.tz.spec.ts
- npm run test:file src/app/features/tasks/store/task-due.effects.spec.ts
- npm run test:file src/app/features/simple-counter/simple-counter.service.spec.ts
- npm run lint
- npm run packages:test
- npm run release-notes:test
- git diff --check

Local note: full npm test did not complete in this checkout because the real path contains non-ASCII characters (/home/coco/桌面/...), which makes the sql.js wasm fixture request fail with ERR_UNESCAPED_CHARACTERS. The related focused specs above pass.

### DIFF
--- a/src/app/core/date/date.service.ts
+++ b/src/app/core/date/date.service.ts
@@ -1,10 +1,14 @@
 import { Injectable } from '@angular/core';
+import { Subject } from 'rxjs';
 import { getDbDateStr } from '../../util/get-db-date-str';
 import { getStartOfNextDayDiffMs } from '../../util/start-of-next-day.util';
 
 @Injectable({ providedIn: 'root' })
 export class DateService {
   private startOfNextDayDiff: number = 0;
+  private _startOfNextDayDiffChange$ = new Subject<void>();
+
+  readonly startOfNextDayDiffChange$ = this._startOfNextDayDiffChange$.asObservable();
 
   setStartOfNextDayDiff(
     startOfNextDayTimeOrHour: string | number | undefined,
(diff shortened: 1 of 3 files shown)


johannesjo (approved):

PR #8076, opened 2026-06-07
fix(tasks): plan started tasks for today

### LINKED ISSUE #8031: 🚨 Active task is not displayed in the tray icon context menu and not automatically scheduled for today
### Version Used

v18.7.0 exe

### Current Behavior

The active task is not displayed in the tray icon context menu when it's not scheduled for today. Therefore it's not possible to get back to it via the tray menu after switching to another one.

System: Windows 11

### Expected Behavior

Any task that I start in the main app window is not automatically scheduled for today (if wasn't) and doesn't appear in the tray menu.

Image

### Steps to Reproduce

1. Open a project and start an existing task (e.g. created yesterday)
2. Check out list of tasks in the tray -- the active task won't appear there

Note: I noticed that this happens when a task is not scheduled for today. However, I don't understand why this happens as any unscheduled task once started, used to be automatically scheduled for today. Scheduling for today solves the issue.

Sometimes, it's intermittent.

### Console Output

```shell

```

### Documentation

cocojojo5213:
## Summary

Fixes #8031.

Starting a task can set it as the current task without making it part of the Today work context. The desktop tray context menu is built from the Today task ids, so a running task that was not already scheduled for today could be missing from the tray task switcher.

This adds a small internal effect that reacts after setCurrentTask and plans the actual current task for today when it is not already in the Today list. It uses the resolved currentTaskId from state, so starting a parent task that actually begins an unfinished subtask plans the subtask. Existing Today tasks are ignored to avoid duplicate persistent operations.

## Validation

- npm run checkFile src/app/features/tasks/store/task-internal.effects.ts
- npm run checkFile src/app/features/tasks/store/task-internal.effects.spec.ts
- npm run test:file src/app/features/tasks/store/task-internal.effects.spec.ts
- npm run test:file src/app/features/tasks/store/task-electron.effects.spec.ts
- npm run test:file src/app/root-store/meta/task-shared-meta-reducers/task-shared-scheduling.reducer.spec.ts
- git diff --check

### DIFF
--- a/src/app/features/tasks/store/task-internal.effects.spec.ts
+++ b/src/app/features/tasks/store/task-internal.effects.spec.ts
@@ -16,12 +16,15 @@ import { LOCAL_ACTIONS } from '../../../util/local-actions.token';
 import { DEFAULT_GLOBAL_CONFIG } from '../../config/default-global-config.const';
 import { GlobalConfigState, TasksConfig } from '../../config/global-config.model';
 import { WorkContextType } from '../../work-context/work-context.model';
+import { selectTodayTaskIds } from '../../work-context/store/work-context.selectors';
+import { DateService } from '../../../core/date/date.service';
 
 describe('TaskInternalEffects', () => {
   let effects: TaskInternalEffects;
   let actions$: Subject<any>;
   let store: MockStore;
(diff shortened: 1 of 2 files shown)


johannesjo (approved):

PR #8077, opened 2026-06-07
fix(sync): keep schedule settings local

### LINKED ISSUE #7661: 💡Sync: Allow 'per-client' configurations for frequency+manual
### Problem Statement

Having just read about the [immediate caveat's and limitation's](https://github.com/super-productivity/super-productivity/discussions/7547) of how Sync can work, my instinct to work around this was to set my most frequently used device to 'automatic every 5m' and the other devices to 'manual'. 
To my surprise, as soon as I set the iOS app to 'manual' and then went back to check my Windows app, it had been changed via sync to manual from 'automatic every 5m'. 

This limits my ability to mitigate known sync problems with my 3+ devices and forces me to choose between two alternatives that diminish or remove the benefit of Super Productivity syncing across devices:
1) Setting it to manual and having to remember to hit the sync button on every device I use, and loosing the benefits of automatic sync or;
2) Increase the sync frequency to something like 'automatic every 10m', loose the benefit of a more frequent sync, and 'hope for the best'.

### Possible Solution

Provide a way to set the sync config separately per client. 

Possible solutions (presented from someone that is fairly naive of the application architecture at present)

Option A:
Perhaps the flag for these settings to sync across devices is already a de-coupled binary yes/no option in the .JSON. If this is the case, we could choose to manually edit it to disable these settings synchronizing. 

This seems like a possibility since at present, not 'everything' syncs (Example: 'Enabling' Automations (experimental) does not toggle this add-on to 'enabled' on other devices, but it does sync the conditional logic rules created. )

Option B:
De-couple the 'automatic' and 'manual sync' options from syncing across devices. 
Pro: This might be the easiest approach if 'Option 1' isn't viable. 
Con: Existing users that have come to expect these settings to sync across devices might be grumpy :)

Option C:
Provide a binary toggle option under 'Sync Advanced Config' to optionally de-couple automatic and manual sync settings, such as a 'sync these settings across all devices Yes/No' with the default selection being 'Yes'.
Pro: This achieves the same result, does not negatively impact unwitting users, and provides flexibility of choice going forward. 
Con: Likely more technically complex :)

### Additional context

In all cases, a logic would need to exist for how to handle the re-enabling of these settings 'syncing' across devices, assuming it doesn't exist. 

Proposed solution if one doesn't exist: If the 'Enable syncing of sync settings across devices' is enabled, use the current device as the 'source of truth' that updates this setting for all devices.

cocojojo5213:
## Summary

Fixes #7661.

- keep sync interval and manual-only mode durable as local-only settings in the local op-log/replay path
- strip those fields only at remote upload/export boundaries for SuperSync/file providers and mark schedule-only updates synced locally without remote upload
- preserve local sync settings when hydrating local full-state imports and when persisting/applying incoming remote full-state ops
- tighten `stripLocalOnlySyncScheduleSettings<T>()` so the return type matches the sanitized runtime shape

## Validation

- `npm run checkFile src/app/features/config/local-only-sync-settings.util.ts`
- `npm run checkFile src/app/features/config/local-only-sync-settings.util.spec.ts`
- `npm run checkFile src/app/features/config/store/global-config.reducer.ts`
- `npm run checkFile src/app/features/config/store/global-config.reducer.spec.ts`
- `npm run checkFile src/app/imex/sync/snapshot-upload.service.ts`
- `npm run checkFile src/app/imex/sync/snapshot-upload.service.spec.ts`
- `npm run checkFile src/app/op-log/capture/operation-log.effects.ts`
- `npm run checkFile src/app/op-log/capture/operation-log.effects.spec.ts`
- `npm run checkFile src/app/op-log/persistence/sync-hydration.service.ts`
- `npm run checkFile src/app/op-log/persistence/sync-hydration.service.spec.ts`
- `npm run checkFile src/app/op-log/sync-providers/file-based/file-based-sync-adapter.service.ts`
- `npm run checkFile src/app/op-log/sync-providers/file-based/file-based-sync-adapter.service.spec.ts`
- `npm run checkFile src/app/op-log/sync/operation-log-upload.service.ts`
- `npm run checkFile src/app/op-log/sync/operation-log-upload.service.spec.ts`
- `npm run checkFile src/app/op-log/sync/remote-ops-processing.service.ts`
- `npm run checkFile src/app/op-log/sync/remote-ops-processing.service.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/config/local-only-sync-settings.util.spec.ts` (`4` success)
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/op-log/capture/operation-log.effects.spec.ts` (`36` success)
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/op-log/persistence/sync-hydration.service.spec.ts` (`37` success)
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/op-log/sync/operation-log-upload.service.spec.ts` (`60` success)
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/op-log/persistence/operation-log-hydrator.service.spec.ts` (`60` success)
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/op-log/sync/remote-ops-processing.service.spec.ts` (`42` success)
- `git diff --check`

## Notes

Local-only schedule settings are intentionally kept in the local persistence/replay path, but removed before remote upload/export payloads are written.

### DIFF
--- a/src/app/features/config/local-only-sync-settings.util.spec.ts
+++ b/src/app/features/config/local-only-sync-settings.util.spec.ts
@@ -1,0 +1,50 @@
+import {
+  stripLocalOnlySyncScheduleSettings,
+  stripLocalOnlySyncSettingsFromAppData,
+} from './local-only-sync-settings.util';
+import { SyncProviderId } from '../../op-log/sync-providers/provider.const';
+
+describe('local-only sync settings utils', () => {
+  it('should strip sync schedule settings from a sync config object', () => {
+    const result = stripLocalOnlySyncScheduleSettings({
+      syncInterval: 300000,
+      isManualSyncOnly: true,
+      isCompressionEnabled: true,
+    }) as Record<string, unknown>;
+
+    expect(result).toEqual({
+      isCompressionEnabled: true,
+    });
+  });
+
+  it('should strip local-only sync settings from app data', () => {
+    const result = stripLocalOnlySyncSettingsFromAppData({
+      globalConfig: {
+        sync: {
+          syncProvider: SyncProviderId.WebDAV,
+          syncInterval: 300000,
+          isManualSyncOnly: true,
+          isCompressionEnabled: true,
+        },
+        misc: { isDisableAnimations: true },
+      },
+      task: { ids: [] },
+    }) as Record<string, unknown>;
+
+    const globalConfig = result['globalConfig'] as Record<string, unknown>;
+    const sync = globalConfig['sync'] as Record<string, unknown>;
+
+    expect(sync['syncProvider']).toBeNull();
+    expect(sync['syncInterval']).toBeUndefined();
+    expect(sync['isManualSyncOnly']).toBeUndefined();
+    expect(sync['isCompressionEnabled']).toBe(true);
+    expect(globalConfig['misc']).toEqual({ isDisableAnimations: true });
+    expect(result['task']).toEqual({ ids: [] });
+  });
+
+  it('should leave data without globalConfig.sync unchanged by reference', () => {
+    const data = { task: { ids: [] } };
+
+    expect(stripLocalOnlySyncSettingsFromAppData(data)).toBe(data);
+  });
+});
--- a/src/app/features/config/local-only-sync-settings.util.spec.ts
+++ b/src/app/features/config/local-only-sync-settings.util.spec.ts
@@ -1,4 +1,5 @@
 import {
+  applyLocalOnlySyncSettingsToAppData,
   stripLocalOnlySyncScheduleSettings,
   stripLocalOnlySyncSettingsFromAppData,
 } from './local-only-sync-settings.util';
@@ -46,5 +47,36 @@ describe('local-only sync settings utils', () => {
     const data = { task: { ids: [] } };
 
     expect(stripLocalOnlySyncSettingsFromAppData(data)).toBe(data);
+  });
+
+  it('should apply local-only sync settings to app data', () => {
+    const result = applyLocalOnlySyncSettingsToAppData(
+      {
+        globalConfig: {
+          sync: {
+            syncProvider: SyncProviderId.Dropbox,
+            syncInterval: 600000,
+            isManualSyncOnly: false,
+            isCompressionEnabled: true,
+          },
+        },
+      },
+      {
+        isEnabled: true,
+        isEncryptionEnabled: false,
+        syncProvider: SyncProviderId.WebDAV,
+        syncInterval: 300000,
+        isManualSyncOnly: true,
+      },
+    ) as Record<string, unknown>;
+
+    const globalConfig = result['globalConfig'] as Record<string, unknown>;
+    const sync = globalConfig['sync'] as Record<string, unknown>;
+
+    expect(sync['isEnabled']).toBe(true);
+    expect(sync['syncProvider']).toBe(SyncProviderId.WebDAV);
+    expect(sync['syncInterval']).toBe(300000);
+    expect(sync['isManualSyncOnly']).toBe(true);
+    expect(sync['isCompressionEnabled']).toBe(true);
   });
 });
--- a/src/app/features/config/local-only-sync-settings.util.ts
+++ b/src/app/features/config/local-only-sync-settings.util.ts
@@ -1,5 +1,14 @@
 import { SyncConfig } from './global-config.model';
 
+export type LocalOnlySyncSettings = Pick<
+  SyncConfig,
+  | 'isEnabled'
+  | 'isEncryptionEnabled'
+  | 'syncProvider'
+  | 'syncInterval'
+  | 'isManualSyncOnly'
+>;
+
 export const LOCAL_ONLY_SYNC_SCHEDULE_KEYS = [
   'syncInterval',
   'isManualSyncOnly',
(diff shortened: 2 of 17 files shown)


johannesjo:
I think this still has two local replay gaps for the new local-only sync settings.

1. `operation-log.effects.ts` strips `syncInterval`/`isManualSyncOnly` before `appendWithVectorClockUpdate`, and skips appending entirely for schedule-only updates. Since `state_cache` is not refreshed on every config op, a restart/replay before a later snapshot or compaction can lose the per-device schedule setting.

2. `sync-hydration.service.ts` writes the stripped `syncedData` into the local `SYNC_IMPORT` payload, then restores local settings only into the cache/dispatch copy. The hydrator can later load a full-state op directly from `OPS` when `state_cache` is missing/corrupt, bypassing that restore. In that path we get `syncProvider: null`, missing schedule fields, and possibly another client’s `isEnabled`/`isEncryptionEnabled` because the helper only strips provider/schedule and the reducer accepts incoming values when `syncProvider` is null.

Can we keep local-only values durable in the local persistence/replay path and strip them only at remote upload/export boundaries, or apply a local-settings overlay when hydrating from full-state ops? Please add regression coverage for a schedule-only config update surviving local op-log replay, and for `SYNC_IMPORT` replay without `state_cache`.

Nit: `stripLocalOnlySyncScheduleSettings<T>()` currently returns `T` even though it deletes keys from the returned object. That overstates the result type for inputs where `syncInterval` is required. Returning an `Omit<T, "syncInterval" | "isManualSyncOnly">` or a plain sanitized record would make the helper contract match the runtime shape.

cocojojo5213:
Thanks, pushed follow-ups for both replay gaps.

What changed:
- Local config capture now keeps the full local sync config payload in the local op log, including schedule-only updates.
- Remote upload/export boundaries still strip `syncInterval` / `isManualSyncOnly`, and schedule-only ops are marked synced locally without uploading.
- Local `SYNC_IMPORT` full-state payloads are stored with the current local sync settings overlaid, so replay from `OPS` still works when `state_cache` is missing.
- Incoming remote full-state ops get the same local overlay before local persistence/application, keeping local replay durable there as well.
- `stripLocalOnlySyncScheduleSettings<T>()` now returns `Omit<T, 'syncInterval' | 'isManualSyncOnly'>`.

Added regression coverage for schedule-only config replay, `SYNC_IMPORT` replay without `state_cache`, and incoming full-state replay. Local verification passed for the touched files plus the focused specs; latest GitHub CI is still running on `e38ad035`.

johannesjo:
Thanks for the thorough follow-ups — the two replay gaps and the type nit are all resolved, and keeping the full payload in the local op-log (stripping only at upload) is the right call for durability. Tests are green and the coverage is solid.

Before we merge I want to settle one thing: **the approach.** In the original issue I leaned toward pulling `syncInterval` / `isManualSyncOnly` *out* of synced state entirely and storing them per-device in localStorage, the way `taskWidget` works (see the note at `global-config.model.ts:322`). This PR instead keeps them inside `globalConfig.sync` and strips/overlays them at every sync boundary. Both are valid; the trade-off is:

- **Extract-to-localStorage:** no sync-path changes, but needs a one-time migration and touches the model + form.
- **This PR (strip-at-boundaries):** no migration and the field stays put, but it spreads local-only handling across 6 sync-pipeline services (reducer, snapshot-upload, sync-hydration, file-based-adapter, op-upload, remote-ops-processing). That's the part I want to be deliberate about — the op-log/replay path is the area where I most want to keep surface area small, since a subtle bug there can silently corrupt state across devices.

I'm leaning toward accepting this approach since the durability story now holds together, but I'd like your read on whether the localStorage extraction would actually be smaller in practice before we commit.

A few smaller points:

1. **Scope is broader than the issue.** #7661 is really about file-based providers, but the strip/overlay also runs on the SuperSync path (snapshot upload, full-state op upload, remote-ops overlay). It's harmless since SuperSync doesn't expose these settings and `loadAllData` now back-fills defaults — but let's make that an explicit "yes, apply everywhere" decision rather than incidental.
2. **`_sanitizeRegularOpPayloadForUpload` drops the op when `sectionCfg` is empty** without checking `entityChanges`. Fine today (config-section ops always carry `entityChanges: []`), but worth a one-line comment so the invariant is explicit.
3. The two near-duplicate `globalConfig.sync` guard-walks in the util could collapse into one helper — minor, optional.
4. Coverage is unit-only. Given this is multi-device replay, I'd feel better with one integration/e2e exercising "client A keeps its interval after B syncs" — not blocking, but reassuring for this risk class.

Nothing here is a correctness blocker from my read; it's mostly the approach question + scope intent. 🙏

cocojojo5213:
Thanks, that trade-off summary matches my read too. I pushed a small follow-up in `698895f8e` for the two low-risk code notes:

- collapsed the duplicate `globalConfig.sync` guard-walks into one helper in `local-only-sync-settings.util.ts`
- added the invariant comment for dropping a schedule-only `GLOBAL_CONFIG_UPDATE_SECTION` upload payload: these config-section ops replay from `actionPayload`, and `entityChanges` are empty for this action

On the approach question: I do not think moving only `syncInterval` / `isManualSyncOnly` to localStorage would be smaller in practice for the current branch. It would remove those two keys from the remote-boundary sanitizer, but it would introduce a second source of truth for one slice of the sync form and config service:

- existing installs would still need a migration/first-read bridge from `globalConfig.sync` into localStorage
- the sync form and `GlobalConfigService.getSyncConfig()` would need to compose store-backed sync config with localStorage-backed schedule fields
- imports/full-state replay still need local overlay behavior for the other per-client sync fields (`syncProvider`, `isEnabled`, `isEncryptionEnabled`), so the sync boundary/replay code does not disappear entirely
- splitting only the schedule fields out would make `globalConfig.sync` partly NgRx-backed and partly localStorage-backed, which felt easier to drift than keeping the full sync form state together and making the remote boundaries explicit

So my preference for this PR is to keep the schedule fields in `globalConfig.sync`, persist them durably in the local op log, and strip/overlay them at remote boundaries. That keeps one UI/config source of truth and makes the per-client behavior explicit where data crosses devices.

For scope: yes, I intended this to apply everywhere, including SuperSync. Even though SuperSync does not expose these schedule settings today, the invariant I wanted is provider-independent: remote snapshots/ops should not decide another client's provider/enabled/encryption/schedule choices. Applying the same stripping/overlay to SuperSync keeps that invariant consistent and avoids leaking these local settings if the SuperSync UI grows later.

I did not add the heavier multi-client integration/e2e in this follow-up because the replay-sensitive cases are already pinned close to the op-log boundaries in unit specs, and that keeps the coverage less flaky. I agree it would be reassuring as a broader sync integration test; I just do not think it should block this fix unless you want that extra coverage before merge.

Local verification after `698895f8e`:
- `npm run checkFile src/app/features/config/local-only-sync-settings.util.ts`
- `npm run checkFile src/app/features/config/local-only-sync-settings.util.spec.ts`
- `npm run checkFile src/app/op-log/sync/operation-log-upload.service.ts`
- `npm run checkFile src/app/op-log/sync/operation-log-upload.service.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/config/local-only-sync-settings.util.spec.ts` (4 SUCCESS)
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/op-log/sync/operation-log-upload.service.spec.ts` (60 SUCCESS)
- `git diff --check`

johannesjo:
Thank you! <3

PR #8083, opened 2026-06-07
feat(tasks): allow rapid subtask creation (fix #8081)

### LINKED ISSUE #8081: 💡 Keep subtask creation input open after adding a subtask (Consistency with main tasks)
### Problem Statement

Currently, when creating a new subtask, the input field/window closes immediately after the subtask is added. This behavior is inconsistent with how normal (main) tasks are created, where the input field remains open. Because the window closes, it breaks the flow when a user wants to quickly enter several subtasks one after another.

### Possible Solution

Do not close the subtask creation window/input field after a subtask is created. It should remain open and focused, behaving exactly like the normal task creation process. This will allow users to rapidly type and add multiple subtasks in a row.

The current alternative is to manually reopen the subtask creation field for every single subtask. This is highly tedious, requires unnecessary clicks/keystrokes, and severely interrupts the workflow.

### Additional context

Keeping the input open is arguably even more important for subtasks than for main tasks, as breaking down a larger task usually involves brain-dumping multiple smaller steps all at once. Standardizing this behavior will make the app feel much more consistent and user-friendly.

hajiboy95:
## Problem

#8081
Currently, when creating a new subtask, the input field closes immediately after the subtask is added. This behavior is inconsistent with how main tasks are created, where the input remains open and focused for rapid entry. This inconsistency breaks the user's flow when they need to brain-dump multiple subtasks in a row, forcing them to manually reopen the creation field for every single item.

## Solution

Modified `TaskComponent` to allow the subtask creation window to remain open after a subtask is added. Specifically:

* Updated the title editing logic to trigger the "add subtask" behavior on a regular `Enter` keypress, but only if the task has a `parentId` (meaning it is a subtask).
* This mirrors the existing `Mod+Enter` behavior for subtasks, making the rapid-entry flow the default for subtasks.
* Top-level tasks retain their existing behavior (`Enter` saves and closes the input) to avoid disrupting the primary task management flow.

## Type of Change

* [ ] Bug fix
* [x] New feature
* [ ] Breaking change
* [ ] Refactoring
* [ ] Documentation
* [ ] Other (please describe)

## Checklist


johannesjo:
Reviewed with three focused passes and re-checked against PR head `98bd87970442c857288cd1cbb80d38d27c8bec8a`. CI is passing, but I found two issues worth addressing before merge.

1. `src/app/features/tasks/task/task.component.ts:731`

   This makes plain `Enter` on any subtask behave like `Mod+Enter`. Since `task-title` uses plain Enter as the normal save/close trigger, editing an existing subtask and pressing Enter now creates or focuses a sibling instead of just finishing the edit. When no empty sibling exists, `_addSubTaskOrFocusEmpty()` calls `addSubTaskTo()`, which dispatches a persistent `OpType.Create`, so the blank continuation row becomes real synced task state and can later be archived as a completed blank subtask. I think this needs to be limited to the actual rapid-entry creation flow, or the empty continuation task needs reliable cleanup before it can leak into persisted/synced/archive state.

2. `e2e/tests/task-list-basic/finish-day-quick-history-with-subtasks.spec.ts:96`

   This weakens the history check from ordered `nth()` assertions to “visible somewhere”. The comment still says the parent appears with its subtasks grouped below it, but this would pass if the rows were reordered, ungrouped, or if extra blank archived rows were present. Please keep an exact order/grouping assertion here, and avoid filtering out the empty `Enter a title` task earlier since that is a real unfinished task, not just a visual placeholder.

hajiboy95:
@johannesjo adressed

johannesjo:
Hey, first off — thanks for the iteration and for picking up the previous round of feedback. The goal here (rapid subtask brain-dump) is a real one and worth solving. I want to step back though, because I think we've got the mechanism wrong and pushing further on this shape will cost us more than starting from a different angle.

**The thing I keep coming back to is sync.**

Every spawned empty subtask is a real task in NgRx state, which means it flows through the op log and replicates to every device the user has. So:

- The Create op fires the moment the row appears. On a second device that's online, the empty row materializes there before anyone has typed.
- Cleanup on blur/Enter emits a Remove. Every cancelled rapid-entry round trip becomes Create → (maybe Update) → Remove in the log, replicated everywhere.
- Any exit path we miss — menu open, dialog open, navigation, browser tab close, mobile backgrounding, network blip mid-edit — leaves a persisted empty subtask synced across devices, eventually showing up as a completed-empty archive row.

The contributor guide is pretty firm that sync changes need to be reasoned about with replay determinism and concurrent edits in mind, and "create then maybe delete based on transient focus state" is brittle under exactly those conditions. That's the part I don't think we can ship.

The other consequences (Enter behaving differently on fresh vs existing subtasks, the E2E tests now special-casing a trailing empty row in unrelated flows) are downstream of the same root choice — persisting a placeholder.

**What if we did it the other way around?**

Mirror the main \`add-task-bar\`: a small inline "Add subtask…" input below the subtask list when the parent is expanded. Enter submits and stays focused, Escape closes, blur with content auto-submits. No task is created in state until the user actually commits a title — so the op log stays clean, sync stays clean, and the Enter overload + cleanup branches in \`updateTaskTitleIfChanged\` go away.

Rough sketch:

- New \`AddSubtaskInputComponent\` (~100 lines).
- Slot it into \`task.component.html\` around line 305 inside the existing \`.sub-tasks\` block, gated on \`_hideSubTasksMode !== HideAll\`.
- Revert the \`updateTaskTitleIfChanged\` changes.
- Revert the E2E workarounds (they should melt away).

Not trivial — probably half a day of careful work — but the result stops touching the sync system at all, which is the property we need.

Would you be up for taking this direction? Totally fine if not — I can pick it up, or we can close this with thanks for the prototype work that surfaced the design question. Either way, really appreciate the effort here.

hajiboy95:
@johannesjo thank you so much for always keeping the scope so tight. At the end of the day that's exactly the attitude that lead to super productivity being as great as it currently is :D. 

Since you already went through it in your head  and you know the project and it's sync mechanisms much better than I do, I'm totally fine with handing it over to you. If you say you won't implement it for at least a few months, I could give it another thought but instead of taking on this I could rather work on small improvements which won't need as much supervision as this which at the end of the day will save time for both of us. What do you think?

johannesjo:
Thanks for your understanding and patience! I really appreciate it!

There is is a lot to do, so I likely won't touch this anytime soon tbh.

hajiboy95:
@symonbaikov would you like to take this on in a new PR? :D

PR #8090, opened 2026-06-07
fix(worklog): avoid misleading export start/end times

### LINKED ISSUE #5654: 🚨 [BUG] Worklog Export shows incorrect start/end times
### Version Used

16.4.3

### Expected Behavior

 Worklog Export should display accurate per-session start/end times (not hardcoded or duplicated values).

### Current Behavior

Hello, I'm experiencing a critical and reproducible bug in Super Productivity (AppImage v16.4.3 on Alma Linux):

#### 1. Worklog Export displays incorrect "Started Working" / "Ended Working" times
- All rows in the exported Worklog show identical, unrealistic timestamps (e.g., `00:20` to `23:45`) regardless of actual session duration.
- The "Time as clock" column is inconsistent with real tracked time.
- This makes auditing work hours impossible.

Example from export:
```md
Date        | Started Working | Ended Working | Time as clock
2025-11-30  | 00:20           | 23:45         | 6:38
2025-11-30  | 00:20           | 23:45         | 1:28
```

Image

### Environment

- **OS**: Alma Linux (RHEL-based)
- **Super Productivity Version**: AppImage v16.4.3
- **Installation Method**: AppImage (also with Flatpak)

Thank you for your hard work on this excellent tool!

### Steps to Reproduce

1. Track time on multiple tasks over several days.
2. Go to Sidebar → Today → Menu → Worklog.
3. Choose any week and expand it
4. Expand OPTIONS and group by Worklog
5. Observe that "Started Working" and "Ended Working" columns show fixed/fake times, not actual session start/end.

### Console Output

```shell

```

cocojojo5213:
## Summary

Fixes #5654.

- stop repeating day-level start/end values on every task/day row when exporting with `Group by: Worklog`
- keep the day-level START/END values only on the first row for each Worklog day group, leaving following rows for that day blank
- clarify the Worklog wiki note that start/end values are day-level boundaries, not per-task session timestamps

## Why

The current time tracking data model stores `Task.timeSpentOnDay` as per-task daily duration and work context `s/e` values as day-level boundaries. Reusing the day-level boundaries for every individual Worklog export row made them look like per-session start/end times and produced repeated, misleading values.

## Validation

- `npm run checkFile src/app/features/worklog/worklog-export/worklog-export.util.ts`
- `npm run checkFile src/app/features/worklog/worklog-export/worklog-export.util.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/worklog/worklog-export/worklog-export.util.spec.ts` (`22` success)
- `git diff --check`

## Notes

This does not invent per-task session timestamps. If Super Productivity later stores task-level session ranges, the Worklog export can use those to populate start/end for individual Worklog rows.

### DIFF
--- a/docs/wiki/4.21-Worklog.md
+++ b/docs/wiki/4.21-Worklog.md
@@ -20,7 +20,7 @@ The transformation does the following:
 - **Iterates over every task** and every day in that task’s “time spent per day” map.
 - **Builds a hierarchy** — Year → month → day (or, in week view, year → week → day). Each day has a list of entries (task + time for that day) and a **total time** for the day.
 - **Places subtasks** — When a task has a parent, its entry is placed immediately after the parent (or after a sibling) so the list reflects hierarchy.
-- **Uses locale and work times** — Day labels use your locale; if you have work start/end times configured, they can be shown per day where available.
+- **Uses locale and work times** — Day labels use your locale; if you have work start/end times configured, they can be shown for day-level rows where available.
 - **Aggregates without double-counting** — Only **leaf tasks** (tasks with no subtasks) contribute their time to the day/month/year totals. Parent tasks’ own “time spent per day” is not added again on top of subtask time, so the totals are not inflated.
 
 So the worklog is a **reorganized view** of the same “time spent per day” data that lives on tasks (and in the archive). When you open History (including its legacy `worklog`/`quick-history` routes) or the daily summary, the app loads the relevant tasks (active + archive for the context), runs this transformation, and shows the result.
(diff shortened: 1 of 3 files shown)


johannesjo (commented):
Thanks for digging into this — the diagnosis is right. Repeating the day-level `workStart`/`workEnd` on every task row implies per-task session times, which is the misleading part of #5654.

My one concern is the fix in `worklog-export.util.ts`: setting `workStart: 0` / `workEnd: 0` for `WORKLOG` grouping makes every START/END cell render as `' - '`, so the day boundaries are dropped entirely — including for users who do want them.

Could we instead keep the day-level start/end but populate them only on the first row of each day-group (blank on the following rows)? That removes the misleading repetition without throwing the data away. If blanking everywhere is intentional, a short note in the PR description on why would help.

cocojojo5213:
Thanks, updated this to keep the day-level START/END values on only the first row for each Worklog day group, with later rows for the same day blank.

Pushed `238fca3e`. Local verification passed:
- `npm run checkFile src/app/features/worklog/worklog-export/worklog-export.util.ts`
- `npm run checkFile src/app/features/worklog/worklog-export/worklog-export.util.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/worklog/worklog-export/worklog-export.util.spec.ts` (`22` success)

Latest main CI is still running the `Tests` job.

PR #8091, opened 2026-06-07
fix(tasks): keep cleared Today add-task date

### LINKED ISSUE #5339: 🚨 Today tag comes back after removing it from task creation
### Version Used

android phone 15.3.0

### Expected Behavior

Removing the today tag in a task that I am cresting should not come back

### Current Behavior

The today tag keeps on appearing at least once after removing it. If after reappearing I remove it, it keeps removed. 
https://github.com/user-attachments/assets/5f483074-855b-4695-97fe-2e6a053f8f41
As an option I think it should be possible not only to be able to remove it while creatinga task (which means I have to remove it from EVERY single task I create)  but creating an option in Settings to make it optional, like the default project if none selected option there is currently. 

### Steps to Reproduce

1. Create a tag in today project
2. Remove today tag
3. Write task 
4. Today tag comes back
5. Remove today tag again
6. Write something.
7. Today tag doesn't come back

### Console Output

```shell

```

cocojojo5213:
## Summary
- keep an explicitly cleared default due date cleared in the add task bar
- track user-driven date clears separately from parser-driven date updates
- cover both plain text and unrelated short-syntax parsing after clearing Today

Closes #5339

## Tests
- npm run prettier:file -- src/app/features/tasks/add-task-bar/add-task-bar.const.ts src/app/features/tasks/add-task-bar/add-task-bar-state.service.ts src/app/features/tasks/add-task-bar/add-task-bar-parser.service.ts src/app/features/tasks/add-task-bar/add-task-bar-state.service.spec.ts src/app/features/tasks/add-task-bar/add-task-bar-parser.service.spec.ts
- npm run test:file src/app/features/tasks/add-task-bar/add-task-bar-state.service.spec.ts
- npm run test:file src/app/features/tasks/add-task-bar/add-task-bar-parser.service.spec.ts
- git diff --check
- npm run lint
- normal pre-push additionally passed npm run lint, sync-core tests, sync-providers tests, and release-notes:test; full ng test --watch=false hit local Node heap OOM during bundle generation, so the same commit was pushed with HUSKY=0

### DIFF
--- a/src/app/features/tasks/add-task-bar/add-task-bar-parser.service.spec.ts
+++ b/src/app/features/tasks/add-task-bar/add-task-bar-parser.service.spec.ts
@@ -287,6 +287,70 @@ describe('AddTaskBarParserService', () => {
         expect(typeof date).toBe('string');
         expect(time).toBe(defaultTime);
       });
+
+      it('should keep a cleared default date cleared when no date syntax is present', async () => {
+        const defaultDate = '2024-01-15';
+        const defaultTime = '09:00';
+
+        mockStateService.state.and.returnValue({
+          projectId: mockDefaultProject.id,
+          tagIds: [],
+          tagIdsFromTxt: [],
+          newTagTitles: [],
+          date: null,
+          time: null,
+          isDateExplicitlyCleared: true,
+          spent: null,
+          estimate: null,
+          cleanText: null,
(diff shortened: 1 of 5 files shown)


johannesjo (approved):

PR #8092, opened 2026-06-07
feat(azure-devops): configure auto import limit

### LINKED ISSUE #7017: 💡 Azure DevOps Boards higher/custom auto import limit
### Problem Statement

For me the Azure DevOps Boards integration is te most important for using Super Productivity, so I am very happy to see this integration already exists!

One of the issues I run into is the limit of auto import items. There seems to be a limit of 50, which in my case not nearly enough. Ok, you can debate why I have that many assigned on my name. Yes, I only have 'assigned to me' as scope selected. Right now I am missing a lot of the items, yes, I can add them manually, but that doesn't make sense with such a nice integration.

### Possible Solution

It would be great if either: the limit will be way higher (sorry, don't know how many items would be feasible) or maybe better: have an extra configuration option where the user can simply put in the limit manually. Not sure why the current limit is 50, maybe for performance? Or is there a different reason?

_No response_

### Additional context

_No response_

cocojojo5213:
## Summary

Closes #7017.

- add an Azure DevOps auto-import limit setting with the current default of 50
- clamp auto-import work item fetches to the Azure DevOps Work Items List API limit of 200 ids
- route the new form label/description through `T.F.AZURE_DEVOPS.FORM.*`, including a translated `{{max}}` interpolation param
- keep search result mapping at the existing default limit and document Azure DevOps auto-import behavior

Reference: https://learn.microsoft.com/en-us/rest/api/azure/devops/wit/work-items/list

## Validation

- `npm run checkFile src/app/features/issue/providers/azure-devops/azure-devops-api.service.ts`
- `npm run checkFile src/app/features/issue/providers/azure-devops/azure-devops-api.service.spec.ts`
- `npm run checkFile src/app/features/issue/providers/azure-devops/azure-devops-cfg-form.const.ts`
- `npm run checkFile src/app/features/issue/providers/azure-devops/azure-devops-cfg-form.const.spec.ts`
- `npm run checkFile src/app/features/issue/providers/azure-devops/azure-devops.const.ts`
- `npm run checkFile src/app/features/issue/providers/azure-devops/azure-devops.model.ts`
- `npm run checkFile src/app/ui/formly-translate-extension/formly-translate-extension.ts`
- `npm run checkFile src/app/ui/formly-translate-extension/formly-translate-extension.spec.ts`
- `npm run int:test`
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/issue/providers/azure-devops/azure-devops-api.service.spec.ts -- --include='src/app/features/issue/providers/azure-devops/azure-devops-cfg-form.const.spec.ts' --include='src/app/ui/formly-translate-extension/formly-translate-extension.spec.ts'` (`14` success)


johannesjo (commented):
The clamp to [1, 200] and the Work Items List API reference are good, and this is a genuinely requested limit (#7017).

One thing before merge: the new `autoImportLimit` form field uses a hardcoded English `label`/`description`, but #8086 (just merged) converted this exact Azure DevOps form to the `T` translation system. Could you route the new field through `T.F.AZURE_DEVOPS.FORM.*` (adding the keys to `en.json`) to match? For the description's dynamic `Max ${AZURE_DEVOPS_MAX_WORK_ITEM_LIMIT}`, a translation with an interpolation param — or a translated base string with the number appended — keeps it translatable.

Also worth a rebase on current master, since #8086 reworked the same file.

cocojojo5213:
Thanks, rebased on current master and routed the new Azure DevOps auto-import limit field through the translation system.

Pushed `66640f33` with:
- `T.F.AZURE_DEVOPS.FORM.AUTO_IMPORT_LIMIT`
- `T.F.AZURE_DEVOPS.FORM.AUTO_IMPORT_LIMIT_DESCRIPTION` using `{{max}}` params
- focused coverage for the form config and Formly description translate params

Local verification passed for the touched files, `npm run int:test`, and the focused Azure DevOps/Formly specs (`14` success). Latest main CI is still running the `Tests` job.

johannesjo:
Thanks, this is a clean, well-rounded addition. The new limit is wired through end to end — the optional `autoImportLimit?: number` field on the model (`src/app/features/issue/providers/azure-devops/azure-devops.model.ts:9`), the `AZURE_DEVOPS_DEFAULT_WORK_ITEM_LIMIT` / `AZURE_DEVOPS_MAX_WORK_ITEM_LIMIT` constants (`azure-devops.const.ts:3`), the form field, and the `en.json` strings all line up. Making the field optional keeps it backward-compatible with existing synced configs, and the clamp in `_getAutoImportLimit` (`azure-devops-api.service.ts:220`) correctly pins the value to 1..200 even if a stale or out-of-range config comes through. Nice touch threading `descriptionTranslateParams` through the formly extension so the `{{max}}` interpolation stays translatable rather than hardcoded.

One tiny, non-blocking note on the docs: the wiki change inserts Azure DevOps as section 11 and bumps ClickUp from 11 to 12 in `docs/wiki/3.07-Issue-Integration-Comparison.md`. I went through the docs and README and nothing links to the old `### 11. ClickUp` by anchor — every cross-reference uses page-level `[[3.07-Issue-Integration-Comparison]]` wiki links — so this is safe as written. Just flagging the renumber in case any external wiki note references the section by number.

Happy to merge as-is. Thanks for the thorough test coverage too.

PR #8100, opened 2026-06-07
feat(menu-tree): order project and tag dropdowns by tree

### LINKED ISSUE #8068: 💡 Order project and tag selection dropdowns to match file tree structure
### Problem Statement

Currently, the project and tag selection dropdowns do not follow the hierarchical structure of the file tree shown in the side panel. As shown in the screenshot, the list appears scattered—for example, "Project 0" in `Folder 1` and "Project A" in `Folder 1` are separated by other unrelated projects. This lack of folder-based grouping makes it difficult and unintuitive to scan the list and select the correct project, especially when working with multiple folders or projects that share the same name.

### Possible Solution

Order the dropdown options so they mirror the exact file tree structure present in the side panel. By sorting the list this way, projects that belong to the same folder will naturally be grouped together. This will create a much more intuitive and organized selection process that matches the user's existing mental model of their workspace.

* **Strict alphabetical sorting:** This was considered, but it wouldn't solve the problem of organizing projects with identical names that live in different folders. The path indications fix this but i still think sidepanel order would be better since it has a better visible distinction
* **Nested visual dropdowns:** Creating a fully collapsible folder tree inside the dropdown itself, though this might overcomplicate the UI compared to simply reordering the existing flat list.

### Additional context

Image

symonbaikov:
## Solution
Project and tag dropdowns used a flat/alphabetical ordering that did not match the folder hierarchy shown in the side panel. When projects or tags with similar names are organized into folders, the dropdown order becomes harder to scan because related items are separated from their side-panel context.

Fixes #8068.

Add tree-order list helpers to MenuTreeService for projects and tags. The helpers flatten the stored menu tree depth-first, skip folder nodes, preserve tree sibling order, and append items missing from the tree so selectable projects and tags are never dropped.

Expose tree-ordered project and tag lists from ProjectService and TagService while keeping the existing sorted APIs for non-dropdown consumers such as short-syntax suggestions.

Switch project and tag selection dropdowns to the new tree-order APIs across task context menus, task rows, add-task bar controls, note project moves, formly project selection, start-page project selection, and tag edit/toggle menus.

Cover the ordering behavior with MenuTreeService, ProjectService, TagService, add-task-bar actions, and task context menu specs.

Docs/wiki not updated because this only changes ordering in existing dropdowns and does not add a new setting, workflow, or public API.

https://github.com/user-attachments/assets/5d7909f7-c8fe-410c-88f5-40577e3c770c

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
**Review: solid and low-risk — a few non-blocking minimality notes + a merge conflict to resolve**

Behaviorally this looks correct and it's nicely tested. It's display-ordering only (no effects/reducers/actions touched), so there's no sync/op-log risk. Nice touch using `listInTreeOrder$` (not the `ForUI` variant) for the config `select-project` control — that preserves showing hidden-from-menu projects there. Centralizing the flatten logic in `MenuTreeService` and deleting the bespoke tree-flatten in `tag-toggle-menu-list` is a real cleanup.

### Merge status — currently conflicting ⚠️

The branch is behind `master` and GitHub marks it `CONFLICTING`. I simulated the merge: **19 of the 20 files auto-merge cleanly; exactly one conflicts:**

- `src/app/features/tasks/add-task-bar/add-task-bar-actions/add-task-bar-actions.component.spec.ts`

Cause: master commit `9499e88` (#8143, *translate default inbox add-task chip*) refactored the same `mockProjectService` block to share a `mockProjectsSignal` for `listSortedForUI` / `listSorted`. The text conflict itself is trivial (keep both — the shared signal **and** the new `listInTreeOrderForUI`).

**But watch the semantic catch when you resolve it:** #8143 added tests that do `mockProjectsSignal.set([INBOX_PROJECT])` and then assert on the rendered `.action-btn` / project menu. This PR repoints `allProjects` to `listInTreeOrderForUI`, so those tests will only pass if the **tree-order** signal is the one they mutate. Point `listInTreeOrderForUI` at `mockProjectsSignal` (rather than a separate static `signal([earlierTreeProject, mockProject])`) — otherwise `selectedProject()` won't find `INBOX_PROJECT` and the inbox-translation tests fail after a naive merge.

### Minimality suggestions (non-blocking, behavior-preserving)

**1. Dead method.** After this PR, `ProjectService.getProjectsWithoutIdSorted$` has no remaining callers (its only one, `task-context-menu-inner`, now calls `getProjectsWithoutIdInTreeOrder$`). Worth deleting it + the stale spec mock.

**2. Trim unused surface.** `ProjectService.listInTreeOrder` (the *signal*) has no consumer. And three of the new `$` observables — `listInTreeOrderForUI$`, `tagsInTreeOrder$`, `tagsNoMyDayAndNoListInTreeOrder$` — are consumed only by their own specs; every production consumer reads the signal form.

**3. Prefer `computed()` over `combineLatest(...) + toSignal`** (matches the repo's "prefer Signals to Observables"). The pattern

```ts
listInTreeOrder$ = combineLatest([this.list$, this._store$.pipe(select(selectMenuTreeProjectTree))])
  .pipe(map(([projects]) => this._menuTreeService.buildProjectListInTreeOrder(projects)));
```

discards the tree value and re-reads it from `MenuTreeService.projectTree()` inside the helper. It works, but only because the `MenuTreeService` tree-signal subscription happens to fire before this `combineLatest` — an implicit ordering dependency rather than a guarantee. Since `buildProjectListInTreeOrder` already reads the tree *signal*, a plain `computed` auto-tracks both inputs with no `combineLatest` and no discarded value:

```ts
listInTreeOrder = computed(() => this._menuTreeService.buildProjectListInTreeOrder(this.list()));
listInTreeOrderForUI = computed(() =>
  this.listInTreeOrder().filter((p) => !p.isArchived && !p.isHiddenFromMenu));
```

Keep observables only where one is genuinely needed: `listInTreeOrder$ = toObservable(this.listInTreeOrder)` for the `select-project` async pipe, and `getProjectsWithoutIdInTreeOrder$` (parameterized by a stream).

**4. (Optional / your call)** `getProjectsWithoutIdInTreeOrder$` wraps its source in another `combineLatest` with the tree selector so an open move-menu re-orders live if the tree changes mid-interaction. If that live-reorder isn't a goal, the inner `combineLatest` can go (the helper reads the live tree at subscribe time anyway).

The per-service placement (mirroring the existing `*SortedForUI*` convention) is the right call — these are suggestions about *form*, not architecture. Thanks for the contribution! 🙏

johannesjo:
Thank you very much! <3

PR #8102, opened 2026-06-07
fix(jira): skip transition for target status

### LINKED ISSUE #5940: 🚨 Jira Transition "Status for starting task" error on resuming time tracking
### Version Used

16.8.3 exe

### Expected Behavior

Resuming a task does not try to update the status if it's already in the correct status.

### Current Behavior

When pausing and then resuming a task, it produces this error.

Image

Jira transitions can be nammed different than the actual status. So, when first transitioning, it works because the value selected for "Status for starting a task" is something like "Start progress". That is the name of the transition which may change the status to "Work in Progress". Now that the status is seen as different than the selected value, it errors out.

### Steps to Reproduce

1. Load issue transitions that have different names than the issue statuses.
2. Select a status (transition) for starting task
3. Start time tracking → Pause → Start again
4. Error

### Console Output

```shell
log.ts:304 [issue] JIRA_RESPONSE_ERROR {error: Error: {"errorMessages":["It seems that you have tried to perform a workflow operation (Start progr…, requestId: 'issue/WEB-429/transitions__POST__vRHftaJeqJ8BOM1HJhXBE'}error: Error: {"errorMessages":["It seems that you have tried to perform a workflow operation (Start progress) that is not valid for the current state of this issue (WEB-429). The likely cause is that somebody has changed the issue recently, please look at the issue history for details."],"errors":{}}requestId: "issue/WEB-429/transitions__POST__vRHftaJeqJ8BOM1HJhXBE"[[Prototype]]: Object {requestInit: {…}, jiraCfg: {…}, transform: ƒ, resolve: ƒ, reject: ƒ, …}jiraCfg: {isAutoPoll: true, isAutoAddToBacklog: true, isIntegratedAddTaskBar: false, defaultProjectId: 'INBOX_PROJECT', pinnedSearch: null, …}reject: ƒ ()requestInit: {method: 'POST', body: '{"transition":{"id":"11"}}', headers: {…}}resolve: ƒ ()timeoutId: 114302transform: s=>s.response[[Prototype]]: Object
err @ log.ts:304
_handleResponse @ jira-api.service.ts:667
(anonymous) @ jira-api.service.ts:111
emit @ VM4 sandbox_bundle:2
onMessage @ VM4 sandbox_bundle:2
log.ts:169
```

cocojojo5213:
Fixes #5940.

## Summary
- compare the current Jira issue status with the configured transition target status instead of the transition action name
- match target statuses by id first, with a name fallback
- add regression coverage for action-name/status-name mismatches and status-id matches

## Validation
- npm run checkFile src/app/features/issue/providers/jira/jira-issue.effects.ts (known absolute-path prettier wrapper failure)
- npm run checkFile src/app/features/issue/providers/jira/jira-issue.effects.spec.ts (known absolute-path prettier wrapper failure)
- npm run prettier:file -- src/app/features/issue/providers/jira/jira-issue.effects.ts src/app/features/issue/providers/jira/jira-issue.effects.spec.ts
- npm run lint:file -- src/app/features/issue/providers/jira/jira-issue.effects.ts
- npm run lint:file -- src/app/features/issue/providers/jira/jira-issue.effects.spec.ts
- npm run test:file src/app/features/issue/providers/jira/jira-issue.effects.spec.ts
- git diff --check

Normal pre-push also passed lint, sync-core tests, sync-providers tests, and release-notes tests before the full Angular test compile hit the local Node heap limit. The focused Angular spec above passed locally.

### DIFF
--- a/src/app/features/issue/providers/jira/jira-issue.effects.spec.ts
+++ b/src/app/features/issue/providers/jira/jira-issue.effects.spec.ts
@@ -1,0 +1,211 @@
+import { TestBed, fakeAsync, tick } from '@angular/core/testing';
+import { MatDialog } from '@angular/material/dialog';
+import { provideMockActions } from '@ngrx/effects/testing';
+import { Action } from '@ngrx/store';
+import { provideMockStore } from '@ngrx/store/testing';
+import { of, Subject } from 'rxjs';
+import { SnackService } from '../../../../core/snack/snack.service';
+import { TaskSharedActions } from '../../../../root-store/meta/task-shared.actions';
+import { LOCAL_ACTIONS } from '../../../../util/local-actions.token';
+import { IssueLocalState, IssueProviderJira } from '../../issue.model';
+import { IssueProviderService } from '../../issue-provider.service';
+import { IssueService } from '../../issue.service';
+import { JIRA_TYPE } from '../../issue.const';
+import { JiraOriginalStatus, JiraOriginalTransition } from './jira-api-responses';
+import { JiraApiService } from './jira-api.service';
+import { DEFAULT_JIRA_CFG } from './jira.const';
+import { JiraIssueEffects } from './jira-issue.effects';
+import { JiraIssueReduced } from './jira-issue.model';
+import { DEFAULT_TASK, Task } from '../../../tasks/task.model';
+import { TaskService } from '../../../tasks/task.service';
+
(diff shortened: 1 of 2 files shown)


johannesjo (approved):
Reviewed: verified against source (Jira transition target-status comparison / Electron tray setting + fallback), CI green, real issue.

PR #8104, opened 2026-06-07
fix(gitlab): avoid stale comment update alerts

### LINKED ISSUE #5518: 🚨 Gitlab created tasks are marked having updates, even though nothing changed

cocojojo5213:
## Summary

Fixes #5518.

- reuse the base issue-provider freshness check for GitLab polling updates
- stop treating comment timestamps newer than `issue.updated_at` as a separate update baseline
- add regression coverage for comment-driven GitLab updates via `issue.updated_at`

## GitLab behavior note

GitLab upstream touches the noteable issue/MR when a note is saved, so the base `updated_at` freshness path preserves comment-driven update detection for GitLab/self-managed instances following upstream behavior.

References:
- `Note` has `after_save :touch_noteable`: https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/models/note.rb#L211
- `touch_noteable` calls `touch` on the noteable: https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/models/note.rb#L595-611
- GitLab spec coverage for issue notes touching the noteable: https://gitlab.com/gitlab-org/gitlab/-/blob/master/spec/models/note_spec.rb#L275-284

## Validation

- `CHROME_BIN=/usr/bin/chromium-browser npm run test:file -- src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.spec.ts` (`3` success)
- `ng lint sp2 --lint-file-patterns src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.spec.ts`
- `prettier --check src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.ts src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.spec.ts`
- `git diff --check`

### DIFF
--- a/src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.spec.ts
+++ b/src/app/features/issue/providers/gitlab/gitlab-common-interfaces.service.spec.ts
@@ -1,0 +1,166 @@
+import { TestBed } from '@angular/core/testing';
+import { of } from 'rxjs';
+import { GitlabCommonInterfacesService } from './gitlab-common-interfaces.service';
+import { GitlabApiService } from './gitlab-api/gitlab-api.service';
+import { IssueProviderService } from '../../issue-provider.service';
+import { DEFAULT_GITLAB_CFG } from './gitlab.const';
+import { GitlabIssue } from './gitlab-issue.model';
+import {
+  GitlabOriginalComment,
+  GitlabOriginalUser,
+} from './gitlab-api/gitlab-api-responses';
+import { createTask } from '../../../tasks/task.test-helper';
+import { Task } from '../../../tasks/task.model';
+import { IssueProviderGitlab } from '../../issue.model';
+
+const ISSUE_PROVIDER_ID = 'gitlab-provider-1';
+const ISSUE_ID = 'project/repo#42';
+const BASE_UPDATED_AT = '2026-01-08T03:29:14.653Z';
+const LATER_COMMENT_AT = '2026-01-08T03:29:14.717Z';
+const NEWER_UPDATED_AT = '2026-01-08T03:30:00.000Z';
+
+const BASE_CFG: IssueProviderGitlab = {
+  ...DEFAULT_GITLAB_CFG,
+  id: ISSUE_PROVIDER_ID,
+  issueProviderKey: 'GITLAB',
+  isEnabled: true,
+  project: 'project/repo',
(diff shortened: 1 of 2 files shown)


johannesjo (commented):
Thanks for tackling the false-positive alerts (#5518). One thing to confirm before this goes in: this removes the entire `getFreshDataForIssueTask` override, so GitLab update detection falls back to the base class, which keys off `issue.updated_at` and drops the comment-timestamp comparison.

That eliminates the self-/filtered-comment false positives, but it also means an update that *only* adds a comment is detected only if GitLab bumps the issue's `updated_at` on note creation. Could you confirm that holds for the GitLab versions we target (including self-hosted)? If `updated_at` reliably moves on note creation this is a clean simplification; if not, we'd lose comment-driven update detection.

A regression test along the lines of "issue with a new comment but unchanged body is still flagged as updated" would make that guarantee explicit.

cocojojo5213:
Confirmed: GitLab touches the noteable issue/MR when a note is saved, so the issue `updated_at` path should preserve comment-driven update detection for GitLab/self-managed instances following upstream behavior.

References:
- `Note` has `after_save :touch_noteable`: https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/models/note.rb#L211
- `touch_noteable` calls `touch` on the noteable: https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/models/note.rb#L595-611
- GitLab spec coverage for issue notes touching the noteable: https://gitlab.com/gitlab-org/gitlab/-/blob/master/spec/models/note_spec.rb#L275-284

Pushed `05fbe5df` adding the requested regression test for an issue with unchanged body but comment-driven `updated_at` being flagged as updated. Local focused spec, lint/prettier checks, and `git diff --check` passed. Latest main CI is still running the `Tests` job.

Issue #5950, opened 2026-01-10
🚨 Task - Checklist - Checklist Items not Checkable


OLLI-S:
### Version Used

16.8.3

### Expected Behavior

I can check checklist items by clicking them.

### Current Behavior

Clicking checklist items to check them is possible.

When I have an empty line in a checklist, like this:
```
- [ ] This is a checklist item
- [ ] This is a checklist item

- [x] This is a checklist item
- [x] This is a checklist item
- [x] This is a checklist item
- [x] This is a checklist item
```
Then clicking on a checklist item has no effect.

### Steps to Reproduce

Just add a line break after a checklist item.

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
You can click the checkmark. But I agree that it would be better if one could click the whole label too.

OLLI-S:
Clicking the text would really increase the usability, but this is not the problem.

When I have a checklist like this:
```
- [ ] This is a checklist item
- [ ] This is a checklist item
- [x] This is a checklist item
- [x] This is a checklist item
- [x] This is a checklist item
- [x] This is a checklist item
```
Then I can click the box to check and uncheck the checklist items.
See video here: https://www.screenpresso.com/=o0060J6HYj5z

But as soon as I add a line break, like this:
```
- [ ] This is a checklist item
- [ ] This is a checklist item

- [x] This is a checklist item
- [x] This is a checklist item
- [x] This is a checklist item
- [x] This is a checklist item
```
I can no longer lick the box to check and uncheck the checklist items.
See video here: https://www.screenpresso.com/=yhDCOoPF0BWb

johannesjo:
Thank you very much @OLLI-S ! This will be improved upon with the next release!

OLLI-S:
@johannesjo 
Is it intended, that there is an empty line added behind **EVERY** entry, although I have only one empty line?

This is what I have in the markdown code:
```
- [ ] This is a checklist item
- [ ] This is a checklist item

- [x] This is a checklist item
- [x] This is a checklist item
- [x] This is a checklist item
- [x] This is a checklist item
```
---
This is, how it is **CURRENTLY** displayed:
```
[ ] This is a checklist item

[ ] This is a checklist item

[x] This is a checklist item

[x] This is a checklist item

[x] This is a checklist item

[x] This is a checklist item
```
---
This is, how it **SHOOULD BE** displayed:
```
[...]

PR #8109, opened 2026-06-07
fix(markdown): toggle fullscreen checklist labels

cocojojo5213:
## Summary

- allow fullscreen markdown checklist items to toggle when clicking the item text, matching inline markdown preview behavior
- keep link clicks inside checklist items from toggling the checkbox
- reuse a checklist-line helper so checkbox DOM indexes only map to real `- [ ]` / `- [x]` markdown rows
- document rendered checklist click behavior in the task notes wiki page

## Notes

Current `master` already has regression coverage for blank-line-separated checklists in the inline markdown preview. This keeps the PR scoped to the remaining fullscreen preview inconsistency and a related mapping edge case where regular text containing `- [` could offset checkbox-to-markdown-line mapping.

The fullscreen-preview label toggle is intentional: it keeps rendered checklist interaction consistent between the inline preview and fullscreen preview while still preserving normal link-click behavior inside checklist items.

Part of #5950.

## Verification

- `npm run prettier:file -- src/app/features/markdown-checklist/is-markdown-checklist.ts src/app/features/markdown-checklist/is-markdown-checklist.spec.ts src/app/ui/inline-markdown/inline-markdown.component.ts src/app/ui/inline-markdown/inline-markdown.component.spec.ts src/app/ui/dialog-fullscreen-markdown/dialog-fullscreen-markdown.component.ts src/app/ui/dialog-fullscreen-markdown/dialog-fullscreen-markdown.component.spec.ts docs/wiki/4.10-Task-Notes.md`
- `npm run lint:file -- src/app/features/markdown-checklist/is-markdown-checklist.ts`
- `npm run lint:file -- src/app/features/markdown-checklist/is-markdown-checklist.spec.ts`
- `npm run lint:file -- src/app/ui/inline-markdown/inline-markdown.component.ts`
- `npm run lint:file -- src/app/ui/inline-markdown/inline-markdown.component.spec.ts`
- `npm run lint:file -- src/app/ui/dialog-fullscreen-markdown/dialog-fullscreen-markdown.component.ts`
- `npm run lint:file -- src/app/ui/dialog-fullscreen-markdown/dialog-fullscreen-markdown.component.spec.ts`
- `npm run test:file src/app/features/markdown-checklist/is-markdown-checklist.spec.ts -- --include src/app/ui/inline-markdown/inline-markdown.component.spec.ts --include src/app/ui/dialog-fullscreen-markdown/dialog-fullscreen-markdown.component.spec.ts` (`66/66 SUCCESS`)
- `git diff --check`
- normal pre-push passed full `npm run lint`, sync-core tests (`207`), sync-providers tests (`374`), and release-notes tests (`9`); full Angular `ng test --watch=false` still OOMed during bundle generation, so the same commit was pushed with `HUSKY=0`

### DIFF
--- a/docs/wiki/4.10-Task-Notes.md
+++ b/docs/wiki/4.10-Task-Notes.md
@@ -26,7 +26,7 @@ The note system balances structure and flexibility so notes can serve as both pl
 
 1. **Template-guided structure** — The default note template gives new tasks a consistent starting shape (for example planning questions). The template is editable per task, so you keep structure where it helps and drop it where it does not.
 
-2. **Markdown flexibility** — Notes support standard markdown: headings, lists, bold, links, and so on. You can write free-form text or use list syntax that the app can treat as a checklist. You are not locked into a fixed form.
+2. **Markdown flexibility** — Notes support standard markdown: headings, lists, bold, links, and so on. You can write free-form text or use list syntax that the app can treat as a checklist. Rendered checklist items can be toggled from the checkbox or item text, while links keep their normal link behavior. You are not locked into a fixed form.
 
 3. **Context-aware storage** — Notes are stored as simple text attached to a specific task (or, for project notes, to a project). That keeps them easy to find and keeps the data model simple—no heavy schema, just content tied to the right context.
 
--- a/src/app/features/markdown-checklist/is-markdown-checklist.spec.ts
+++ b/src/app/features/markdown-checklist/is-markdown-checklist.spec.ts
@@ -12,7 +12,7 @@ and not:
 Some text yeah
  */
 
-import { isMarkdownChecklist } from './is-markdown-checklist';
+import { isMarkdownChecklist, isMarkdownChecklistLine } from './is-markdown-checklist';
 
 describe('isMarkdownChecklist()', () => {
   [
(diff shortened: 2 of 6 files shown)


johannesjo (commented):
The code here looks reasonable — reusing `isMarkdownChecklistLine` and guarding link clicks so they don't toggle the checkbox. Two scope/bookkeeping points before this goes in:

1. The description notes that `master` already has regression coverage for the originally-reported blank-line checklist bug, yet the PR is tagged `Closes #5950`. If #5950 is already fixed, merging this would auto-close it with an unrelated change — could you drop or adjust that tag?

2. The actual change is fullscreen-preview click-to-toggle plus a `- [` line-mapping edge case, which as far as I can tell wasn't separately reported. Matching the inline-preview toggle behavior sounds right; I'd just like to confirm it's intended rather than incidental — a one-line "why" in the description would do it.

Thanks\!

cocojojo5213:
Thanks for the careful scope check. I updated the PR description to replace  with , and added the rationale that fullscreen-preview label toggling is intentional so rendered checklist interaction matches the inline preview while preserving normal link-click behavior.

cocojojo5213:
Small correction: the previous reply lost the two quoted phrases while sending from the CLI.

I replaced `Closes #5950` with `Part of #5950`, and added the rationale that fullscreen-preview label toggling is intentional so rendered checklist interaction matches the inline preview while preserving normal link-click behavior.

johannesjo:
Nice fix — this is a real correctness improvement. Switching `clickPreview` to `target.closest('a')` / `target.closest('.checkbox-wrapper')` (`src/app/ui/dialog-fullscreen-markdown/dialog-fullscreen-markdown.component.ts:249`) means clicking the label text now toggles the item, matching the inline preview behavior, while link clicks (including nested elements inside an `<a>`) are still correctly excluded. Reusing the shared checklist-line helper for the DOM-index → markdown-line mapping is the right call, and the added specs cover both the gapped-checklist toggle and the link case.

One light, non-blocking follow-up (not a regression — this PR doesn't introduce it): the underlying `isChecklistItemLine` / `CHECKLIST_ITEM_RE` (`src/app/features/markdown-checklist/checklist-operations.ts:17`) only matches `-` bullet checklist lines. I confirmed against the bundled marked (v17) that `* [ ]` and `+ [ ]` also render as real checkboxes in the preview, so a note mixing `-` with `*`/`+` task markers could skew the checkbox-to-line index. (Uppercase `- [X]` is fine — the char class is `[ xX]`, so it's matched and toggled correctly.) Since this is the long-standing single source of truth across toggle/bulk/progress and not something your change touches, it's purely a possible future hardening, not anything to address here.

Looks good to merge. Thanks!

PR #8113, opened 2026-06-07
fix(reminder): restore imported legacy task reminders

### LINKED ISSUE #5843: 🚨 SuperSync - Does not Show Reminder Window
### Version Used

16.7.3

### Expected Behavior

Reminder Window for overdue tasks is also shown in the Super Sync version.

### Current Behavior

After I have imported my data in the Super Sync version (see https://github.com/johannesjo/super-productivity/issues/5839), I recognized, that there is no reminder window shown.
In the desktop version, there is the reminder window shown, showing all overdue tasks.

### Steps to Reproduce

Import the data from a local export.

### Console Output

```shell

```

cocojojo5213:
## Summary

- migrate legacy task reminders through a shared helper used by both backup import and runtime legacy `pf` reminder migration
- align migrated task reminders by setting `task.remindAt`, filling missing `dueWithTime`, and removing legacy `reminderId`
- re-check imported reminders immediately after a data import finishes without adding a second worker message protocol

## Root cause

Issue #5843 was reported against 16.7.3. In that version, task reminders were still stored as `reminders[]` entries linked through `task.reminderId` / `relatedId`. Current reminder scheduling reads task `remindAt` and `deadlineRemindAt` through task selectors, so imported legacy task reminders could be loaded without reaching the reminder worker.

Even after state replacement, the reminder worker only checked its refreshed list on the next 10 second interval. The import flow should suppress reminder dialogs while data is being replaced, then immediately check the refreshed reminder list once import completes.

## Changes

- Extract `migrateLegacyTaskRemindersIntoTasks()` and use it for both legacy backup migration and runtime legacy DB migration.
- Convert legacy `TASK` reminders to active task `remindAt`, including a `task.reminderId` fallback.
- Fill missing `dueWithTime` from the legacy reminder timestamp so backup and runtime migration paths stay aligned.
- Keep note reminders from being migrated into task reminders and skip done tasks.
- Replace the one-shot `skipUntil(importDone$)` reminder gate with explicit import-state suppression.
- Share due-reminder filtering between the worker and the import-finish service path; the worker keeps the regular array-only update shape and normal 10s cadence.

## Validation

- `npm run checkFile src/app/features/reminder/migrate-legacy-task-reminders.util.ts`
- `npm run checkFile src/app/features/reminder/migrate-legacy-task-reminders.util.spec.ts`
- `npm run checkFile src/app/features/reminder/due-reminders.util.ts`
- `npm run checkFile src/app/features/reminder/due-reminders.util.spec.ts`
- `npm run checkFile src/app/features/reminder/reminder.worker.ts`
- `npm run checkFile src/app/features/reminder/reminder.service.ts`
- `npm run checkFile src/app/features/reminder/reminder.service.spec.ts`
- `npm run checkFile src/app/op-log/backup/migrate-legacy-backup.ts`
- `npm run checkFile src/app/op-log/backup/migrate-legacy-backup.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/reminder/reminder.service.spec.ts -- --include='src/app/features/reminder/migrate-legacy-task-reminders.util.spec.ts' --include='src/app/features/reminder/due-reminders.util.spec.ts'` (`25` success)
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/op-log/backup/migrate-legacy-backup.spec.ts` (`43` success)
- `git diff --check`

Fixes #5843


johannesjo:
Thanks for tackling this — I traced the data flow and the fix is sound, CI is green. A few notes from review, mostly about keeping the legacy-migration footprint small and single-sourced.

**The migration is genuinely needed (verified).** I read `task.model.ts` at tag `v16.7.3`: `Task` had `reminderId?` + `dueWithTime?` but **no `remindAt`**, and reminders lived in a separate persisted `reminders[]` model linked via `relatedId` / `reminderId`. So a real pre-v17 backup carries reminders only in `reminders[]`, and the current task-selector pipeline (`selectAllTasksWithReminder` → `task.remindAt`) can't see them. Without this migration, an imported pre-v17 backup surfaces no reminders at all — so 👍 on adding it.

**1. Two divergent copies of the same transform.** `reminders[] → task.remindAt` now exists in both `reminder.service.ts:_migrateLegacyReminders` (runtime, reads the legacy `pf` IndexedDB) and the new `migrate-legacy-backup.ts:_migrationLegacyTaskReminders` (backup file). They already disagree: the runtime path sets **both** `dueWithTime` + `remindAt` and skips NOTE; the new path sets **only** `remindAt` and skips NOTE + done. Worth extracting one pure helper both call, so they can't drift:

```ts
// pure helper, shared by both ingestion paths
export const migrateLegacyTaskRemindersIntoTasks = (taskState, reminders): void => {
  for (const r of reminders) {
    if (r?.type !== 'TASK' || typeof r.remindAt !== 'number') continue;
    const task = taskState.entities[r.relatedId]
      ?? Object.values(taskState.entities).find((t) => t?.reminderId === r.id);
    if (!task || task.isDone) continue;
    task.remindAt = r.remindAt;
    if (typeof task.dueWithTime !== 'number') task.dueWithTime = r.remindAt;
    delete task.reminderId;
  }
};
```

**2. `dueWithTime` alignment (minor).** The runtime path sets `dueWithTime = remindAt`; the backup path doesn't. For typical 16.7.3 data the task already carries `dueWithTime`, so this only bites the "reminder but never scheduled" edge — but aligning it (the one line above) keeps migrated tasks consistently visible in Scheduled/Overdue too.

**3. The worker message protocol over-builds the immediate-check.** The `{ reminders, isCheckImmediately }` shape + the two new interfaces exist only to make the immediate check fire *exclusively* on import-finish. Two simpler shapes:
- Make the worker check immediately on every new list (one line: call `checkAndPostDueReminders(reminders)` before scheduling the interval) and drop the protocol/flag. *Caveat:* this also makes reminders pop immediately when a task with an overdue reminder arrives via sync, instead of within 10s today.
- Or leave the worker untouched and, on import-finish, have the service filter `_latestWorkerReminders` for due ones and emit directly via a shared `getDueReminders()` — confines the new behavior to one place, no new message shape, preserves today's cadence.

(The `_isDataImportInProgress` guard that replaces the old, effectively-no-op `skipUntil` is a real improvement — keep it regardless.)

**4. A real-backup fixture test would lock this in.** A characterization test built from an actual pre-v17 export (the reporter has one that repros reliably) asserting that no `remindAt` lands *without* the migration and the reminder fires *with* it would both prove the migration is required and document why it exists.

None of this is blocking — the fix works. Mainly about not growing a second, slightly-different legacy reminder path that future readers have to keep in sync by hand.

cocojojo5213:
Thanks for the detailed review. I updated the follow-up to keep the legacy reminder migration single-sourced:

- extracted `migrateLegacyTaskRemindersIntoTasks()` and use it from both the backup migration path and the runtime legacy `pf` reminder migration path
- aligned the backup path with the runtime path by setting `dueWithTime` when it is missing, alongside `remindAt`
- kept NOTE reminders out of the task migration and kept done tasks skipped
- made the runtime path wait for loaded task state before migrating and clearing legacy reminders, so startup/import timing cannot drop legacy reminders
- removed the `{ reminders, isCheckImmediately }` worker message protocol; the worker now keeps the regular array-only update shape/cadence, and import-finish checks due reminders in the service via shared due-reminder filtering
- kept the `_isDataImportInProgress` suppression guard

I did not add a real pre-v17 backup fixture because I do not have the reporter export in this branch, but I added focused unit coverage for the shared migration behavior, backup migration alignment, runtime migration, loaded-state waiting, and the import-finish immediate due-reminder path.

Local verification:
- `npm run checkFile src/app/features/reminder/migrate-legacy-task-reminders.util.ts`
- `npm run checkFile src/app/features/reminder/migrate-legacy-task-reminders.util.spec.ts`
- `npm run checkFile src/app/features/reminder/due-reminders.util.ts`
- `npm run checkFile src/app/features/reminder/due-reminders.util.spec.ts`
- `npm run checkFile src/app/features/reminder/reminder.worker.ts`
- `npm run checkFile src/app/features/reminder/reminder.service.ts`
- `npm run checkFile src/app/features/reminder/reminder.service.spec.ts`
- `npm run checkFile src/app/op-log/backup/migrate-legacy-backup.ts`
- `npm run checkFile src/app/op-log/backup/migrate-legacy-backup.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/reminder/reminder.service.spec.ts -- --include='src/app/features/reminder/migrate-legacy-task-reminders.util.spec.ts' --include='src/app/features/reminder/due-reminders.util.spec.ts'` (`25` success)
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/op-log/backup/migrate-legacy-backup.spec.ts` (`43` success)
- `git diff --check`

johannesjo:
Thanks for digging into this one and for the thorough tests + validation notes — that's appreciated. 🙏

Before we go further I want to be upfront about scope: this is a fairly rare edge case (importing a legacy 16.7.3-era backup that still uses `reminders[]` + `task.reminderId`), so I'd like to keep the fix as small as possible. Right now the PR bundles three separate things, and only the first is really needed for #5843:

1. **The actual fix** — converting legacy `reminders[]` → `task.remindAt` during backup import (`migrate-legacy-backup.ts` + the shared util). 👍
2. **The reminder-service rework** — replacing the `skipUntil` import gate with explicit suppression + an immediate re-check on import finish. This only removes a ≤10s delay; once the tasks carry `remindAt`, the worker already picks them up on its next tick without a reload, which is what the issue actually reported.
3. **The runtime pf-migration refactor** — switching from `reScheduleTaskWithTime` to raw `updateTasks`. This touches a working, *synced* code path that's unrelated to the import bug.

Could we trim this down to just (1)? That keeps the change to ~20–30 lines, out of the reminder service/worker and out of the synced migration path. Happy to keep the delay as-is — it doesn't need to be instant.

A couple of concrete points regardless of scope:

- **Clear `dueDay` when setting `dueWithTime`.** In `migrateLegacyTaskRemindersIntoTasks`, when you assign `task.dueWithTime`, you also need `delete task.dueDay`. `dueDay` and `dueWithTime` are mutually exclusive (the `reScheduleTaskWithTime` reducer enforces this), so a task that already had a `dueDay` would end up with both set after migration.
- **Runtime path behavior.** `updateTasks` skips the TODAY_TAG ordering insertion that `reScheduleTaskWithTime` did. Membership is virtual so the task still shows, but if we keep this path I'd rather stay on `reScheduleTaskWithTime`. (The `isDataLoaded` wait you added is a nice catch, though.)

One thing that would help me a lot: do you have a sample export (or logs) from the version that reproduces this? I want to confirm whether the imported data is actually in the legacy `reminders[]` shape vs. modern `remindAt` — it affects which of the above is really the fix, and it'd make a good test fixture.

cocojojo5213:
Thanks, that scope concern makes sense. I pushed `1640cee23` to trim this back to just the backup-import migration path.

What changed in the follow-up:
- removed the reminder-service/import-finish re-check changes, worker changes, due-reminders helper, and runtime `pf` migration refactor from the PR diff
- kept the pure legacy `reminders[]` -> `task.remindAt` helper only for backup import
- when the helper fills a missing `dueWithTime`, it now also deletes `dueDay`
- kept focused synthetic legacy-shape coverage for direct `relatedId`, `reminderId` fallback, done/non-TASK skips, `dueWithTime` preservation, and backup import clearing `reminders[]`

I do not have a real pre-v17 export or logs from the reporter available for this branch, so I could not add an actual export fixture. The current tests use the legacy persisted shape described in the review instead.

Local verification after the trim:
- `npm run checkFile src/app/features/reminder/migrate-legacy-task-reminders.util.ts`
- `npm run checkFile src/app/features/reminder/migrate-legacy-task-reminders.util.spec.ts`
- `npm run checkFile src/app/op-log/backup/migrate-legacy-backup.ts`
- `npm run checkFile src/app/op-log/backup/migrate-legacy-backup.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/reminder/migrate-legacy-task-reminders.util.spec.ts` (5 SUCCESS)
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/op-log/backup/migrate-legacy-backup.spec.ts` (43 SUCCESS)
- `git diff --check`

johannesjo:
Thank you! <3

Issue #7870, opened 2026-05-30
Code quality audit: remove privacy-sensitive logs from exported log history


johannesjo:
## Context
The audit found several logs that include task objects, issue payloads, parsed user input, plugin archive data, or file paths. Log history is exportable, so user content should not be logged.

## Scope
Representative locations from the audit:
- `src/app/core-ui/magic-side-nav/magic-side-nav.component.ts`: logs `draggedTask` on project/tag drop.
- `src/app/pages/daily-summary/daily-summary.component.ts`: logs `doneTasks`.
- `src/app/features/tasks/store/short-syntax.effects.ts`: logs parsed short-syntax input in production.
- `src/app/features/tag/store/tag.reducer.ts`: logs tag state.
- `src/app/features/tasks/task/task.component.ts`: logs task arrays/repeat config around project moves.
- `src/app/features/tasks/add-task-bar/add-task-bar-issue-search.service.ts`: logs issue/task search results.
- `src/app/plugins/plugin.service.ts`: logs raw extracted ZIP file data.
- Additional audit candidates include calendar/issue provider logs and Electron/local-file-sync path logs.

## Suggested approach
1. Replace raw object/content/path logging with ids, counts, safe error names, and coarse status.
2. Add a narrow grep or lint-style check for common unsafe log patterns.
3. Keep diagnostic value by preserving operation ids, task ids, counts, provider ids, and fixed error categories.

## Verification
- `rg -n "console\.(log|warn|error|debug|info)|Log\.log\(|IssueLog\.log\(|TaskLog\.log\(" src/app electron packages -g "*.ts" -g "!**/*.spec.ts"`
- `npm run checkFile <file>` for every changed `.ts` file.
- Run focused specs for any touched effect/service files.

johannesjo:
A full multi-agent sync review turned up two more privacy-log leak sites that fit this audit. Both were adversarially verified against the code. Adding them here rather than filing a duplicate issue.

### Calendar-integration effects log full task objects + private calendar event content
**`src/app/features/calendar-integration/store/calendar-integration.effects.ts:278`** (also :232 and the `Log.log('_____END___', a)` array dump at :188)

[...]

PR #8121, opened 2026-06-08
fix(plugins): avoid logging extracted ZIP data

cocojojo5213:
## Summary
- avoid logging the raw extracted ZIP file map when loading uploaded plugins
- keep safe diagnostics for ZIP size/type, entry count, total extracted bytes, and whether expected entries are present

This keeps plugin ZIP diagnostics useful without putting archive entry data or file contents into exportable log history.

Part of #7870.

## Verification
- `npm run checkFile src/app/plugins/plugin.service.ts`
- `CHROME_BIN=/snap/bin/chromium npx ng test --watch=false --include='**/plugin.service.load-from-zip.spec.ts'` (`3` success)
- `CHROME_BIN=/snap/bin/chromium git push -u origin fix/plugin-zip-safe-log-7870` pre-push hook:
  - `npm run lint`
  - package tests for `sync-core` and `sync-providers`
  - release-notes tests
  - Angular/Karma tests: `10359` success
  - LA timezone Angular/Karma tests: `10345` success

### DIFF
--- a/src/app/plugins/plugin.service.ts
+++ b/src/app/plugins/plugin.service.ts
@@ -1076,7 +1076,10 @@ export class PluginService implements OnDestroy {
   }
 
   async loadPluginFromZip(file: File): Promise<PluginInstance> {
-    PluginLog.log(`Starting plugin load from ZIP: ${file.name}`);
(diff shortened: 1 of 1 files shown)


johannesjo (approved):

PR #8127, opened 2026-06-08
fix(electron): avoid logging local file sync paths

cocojojo5213:
## Summary
- stop logging local file sync file paths and directory entries from the Electron IPC handlers
- replace raw local sync errors with error name/code metadata
- keep diagnostics for operation type, data length, local revision presence, and directory entry count

## Why
Part of #7870. Local file sync paths and directory contents can expose user-specific filesystem information in exported logs. The diagnostics should remain useful without recording those paths or entries.

## Verification
- `npm run checkFile electron/local-file-sync.ts`
- `npm run electron:build`
- `CHROME_BIN=/snap/bin/chromium git push -u origin fix/local-file-sync-safe-log-7870`


johannesjo:
Thank you!

PR #8128, opened 2026-06-08
fix(core): avoid logging crash user data

cocojojo5213:
## Summary
- stop logging the crash alert user-data backup object
- remove a debug log of the GitHub issue link DOM collection
- keep the explicit crash export and anonymized export buttons unchanged

## Why
Part of #7870. The crash alert can receive a complete backup object so users can explicitly export it for debugging. That object should not also be written into exported logs automatically.

## Verification
- `npm run checkFile src/app/core/error-handler/global-error-handler.util.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/core/error-handler/global-error-handler.util.spec.ts`
- `CHROME_BIN=/snap/bin/chromium git push -u origin fix/global-error-safe-log-7870`

### DIFF
--- a/src/app/core/error-handler/global-error-handler.util.ts
+++ b/src/app/core/error-handler/global-error-handler.util.ts
@@ -78,7 +78,6 @@ export const logAdvancedStacktrace = (
       }
 
       const githubIssueLinks = document.getElementsByClassName('github-issue-urlX');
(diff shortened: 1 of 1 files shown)


johannesjo (approved):

PR #8129, opened 2026-06-08
fix(issue): avoid logging provider config updates

cocojojo5213:
## Summary
- replace raw issue-provider config update logging with count/flag-only metadata
- avoid logging raw OAuth/options-load errors from the edit dialog
- keep provider id/key and safe error name/code diagnostics for troubleshooting

## Why
Part of #7870. Issue provider config updates can include credentials, URLs, default notes, plugin configuration, or other user/provider data. Logs should preserve coarse diagnostics without recording config values or raw error payloads.

## Verification
- `npm run checkFile src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.ts`
- `CHROME_BIN=/snap/bin/chromium git push -u origin fix/issue-provider-config-safe-log-7870`


johannesjo (approved):

PR #8130, opened 2026-06-08
fix(trello): avoid logging raw picker results

cocojojo5213:
## Summary
- replace raw Trello picker result logging with safe aggregate diagnostics
- keep result count, Trello issue keys, and label/member/attachment counts
- avoid logging card titles, descriptions, URLs, members, labels, or attachment data

## Why
Part of #7870. Trello picker results can include card summaries, descriptions, URLs, member data, labels, and attachment metadata. The log should keep enough information to debug result shape without recording the raw issue payloads.

## Verification
- `npm run checkFile src/app/features/issue/providers/trello/trello-common-interfaces.service.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/issue/providers/trello/trello-api.service.spec.ts`
- `CHROME_BIN=/snap/bin/chromium git push -u origin fix/trello-picker-safe-log-7870`


johannesjo:
Thank you!

Issue #7353, opened 2026-04-24
💡 All Tasks


session11:
### Problem Statement

Since you altogether removed the option to group tasks by project (originally requested and implemented in #7050 and later removed due to #7279 ) , I request you to implement "ALL TASKS" or something similar which enables the user to view all tasks divided across various projects. I already mentioned in my original post that I have many tasks in various projects and keeping track of them by heading to individual projects is a tiresome process. I hope you consider this. 

### Possible Solution

Just like TODAY and INBOX, we should have a tab that lists all tasks divided across various projects. To make it even more convenient, users could be given the option to select projects (check boxes) to list tasks from. This way, if some users only prefer to view tasks from certain projects, that becomes possible too. 

### documentation gap

- [ ] ... related docs should be updated as part of the same PR whenever possible.
- [ ] ... I can help improve the documentation.

JulioMReis-tecnico:
Hello, I was interested in taking a look and implement this feature (a new optional tab with all the tasks, the usual filtering and added filtering to specify projects).
If there isnt anyone working on this could i be assigned to this issue?

johannesjo:
Hey @JulioMReis-tecnico ! That's great to hear! I think we should make the list optional (under app features) and opt in to avoid clutter for users that don't need it.

session11:
> Hello, I was interested in taking a look and implement this feature (a new optional tab with all the tasks, the usual filtering and added filtering to specify projects). If there isnt anyone working on this could i be assigned to this issue?

I would love to see this implemented. Thank you so much.

kalejuece:
Hi. I would also like to see this. Perhaps even add the option to use the app's inbox similar to email ones? As a folder that contains all tasks. I think that would make it far more intuitive for a lot of users, myself included. At any rate, being forced to check each project individually is not ideal, that's for sure.

cryptominister:
I'd really love to see an "All Tasks" view in the sidebar, similar to how "Today" and "Inbox" work, but showing every task from every project in a single unified list.

ivangretsky:
Good day!

@johannesjo may I ask if you were able to review the PR from @JulioMReis-tecnico ? This feature is a really long-wanted one, so trying to move it forward as I can)

johannesjo:
[...]

Issue #11, opened 2017-10-03
Done-backlog not saved


johnhorsema:
Was using the [web version](http://super-productivity.com/app/#!/), but when I press 'Finish Your Workday', completed tasks disappear. They are not saved and I cannot view them at [Done Tasks Backlog](http://super-productivity.com/app/#!/done-tasks-backlog)

johannesjo:
Thanks for reporting! Had a similar issue with a newly created project. Will try to look into it tomorrow!

johannesjo:
I think it should work now. Thanks again for reporting! Lead me in the right direction!

PR #8134, opened 2026-06-08
Feat(all-tasks-tab): added new page named All Tasks

JulioMReis-tecnico:
## Summary
As requested in #7353 
Added a new page, accessible in the side menu, named All Tasks. Optional toggle in app features.

It lists every task across every project and inbox.

It's 'filter by > project' is modified to be a checkbox list (with search bar), by default shows every tasks but can be filtered to any number of projects (inbox included). Other filtering options working as normal.

It's 'time remaining' above the list shows time based on tasks being shown.

##Tests
Verifies 'All Tasks' correctly displays tasks from multiple projects, the filter for projects works and the 'time remaining' updates.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo (commented):
Thanks for taking this on — an "All Tasks" view is a popular request and the page-wiring (route, nav entry, feature flag, start-page option, i18n, e2e) is all in the right places and follows the existing Boards/Planner pattern well. 👍

Before merge, though, I think the PR is doing more than it needs to, and a few changes to shared code risk regressing the **normal** Today/project/tag views. Grouping feedback by priority:

### 🔴 Correctness — changes to shared code that affect existing views

1. **`task-view-customizer.service.ts` — the normal context switch now skips sanitization.** Routing the existing subscribe through the new `_applyContext()` replaces `this._sanitizeFilter(stored?.filter)` with `stored?.filter ?? DEFAULT_OPTIONS.filter`, and always passes `null` to `_sanitizeGroupForContext` instead of `activeType`. The `_sanitizeFilter` comment explains why it exists (stale `label` after a translation-key rename renders directly; unknown stored `type` no longer coerced to default), and the `null` means "group by project" is no longer reset inside a project context. This regresses Today/project/tag, not just All Tasks. Could the override be made **purely additive** — keep the existing subscribe + sanitize calls untouched, and only add the new override path?

2. **`work-view.component.ts` (`customizedUndoneTasks` initialValue) — the identity guard is broken.** `initialValue` is changed to `{ list: INITIAL_CUSTOMIZED_UNDONE_TASKS }`, but `INITIAL_CUSTOMIZED_UNDONE_TASKS` is already `{ list: [] }`, so this is `{ list: { list: [] } }` (wrong shape) and is no longer the same reference that `_getVisibleUndoneTasksForSelection()` checks (`customized \!== INITIAL_CUSTOMIZED_UNDONE_TASKS`). That guard (documented as "filtered list hasn't emitted yet → skip deselect") is now always `true` for every work view when a filter is active. Please preserve the `INITIAL_CUSTOMIZED_UNDONE_TASKS` reference and its shape.

### 🟡 Simplification — most of the size is avoidable

3. **`all-tasks-filter-panel.component.*` (~490 lines) is ~95% a copy of `TaskViewCustomizerPanelComponent`.** The only real difference is the checkbox/multi-select project filter; everything else (sort, group, tag, date, deadline, time) is duplicated, and the original panel was also edited to hide its project filter. Two parallel panels will drift. Could the multi-select instead be a **mode/param on the existing panel** (and/or the existing project filter enhanced to multi-select for all contexts)? That removes the fork entirely.

4. **Drag-drop ordering looks out of scope.** `AllTasksOrderService`, the two new `LS` keys, `_saveAllTasksOrder()`, and the `if (/all-tasks/.test(this._router.url))` branch inside the shared `TaskListComponent` add a per-device, non-synced ordering side-channel into a generic component. The PR description lists list + filter + time-remaining (not ordering). Could this be dropped for now (default ordering) or split into a follow-up?

5. **`entity-types.ts: +'ALL_TASKS'` appears unused.** This is the synced entity-type enum ("operations with unknown entity types will be rejected by the server"), and I don't see anything emitting an `ALL_TASKS` op (the localStorage ordering isn't synced). Adding to this shared contract has forward-compat implications for older clients — I'd remove it unless something actually needs it.

### ⚪ Minor

6. **`task-list.component.ts` has ~40 lines of whitespace-only reindent** of the existing `_moveItemInList` IIFE, which roughly doubles the diff and obscures the real change. Also `_saveAllTasksOrder` is indented 4 spaces instead of 2 — worth re-running `npm run checkFile` on that file.

### Suggested leaner shape
Keep the page wrapper + all the page-wiring as-is; make the customizer override additive (#1); preserve the work-view initial reference (#2); reuse/extend the existing panel instead of forking (#3); drop ordering + the `ENTITY_TYPES` change for now (#4, #5). That should roughly halve the diff and remove the risk to existing views.

Happy to help with any of these if useful — thanks again for the contribution\!

JulioMReis-tecnico:
Thanks for the feedback.
Hopefully the problems are resolved. :)

The drag-drop ordering seemed interesting due to the potential high number of tasks, but i removed it for now.
Can always be re-added if theres interest.

johannesjo:
Thanks for taking this on! The feature itself is well-scoped, but here are findings from a careful pass — a few correctness issues worth fixing, plus simplifications that should noticeably shrink the diff.

### Correctness

1. **Archived-project tasks leak onto the page.** `all-tasks-page.component.ts:31-43` uses `selectAllTasksWithSubTasks`, which has no archived-project filter. Every other top-level task list in the app uses `selectAllTasksInActiveProjects`. Users will see/edit/complete tasks from projects they archived.

2. **`estimateRemainingToday` reimplementation has two bugs** (`work-view.component.ts:217-225`):
   - It clamps the *cumulative* sum (`Math.max(0, sum + …)`) instead of clamping per task — one over-spent task wipes out other tasks' remaining estimates.
   - It ignores `subTasks`. Parent tasks usually have `timeEstimate = 0`; their subtask remainings won't be counted.

   Just reuse `mapEstimateRemainingFromTasks` from `work-context.util.ts` — same input shape, already does the right thing.

3. **Side-nav label not translated.** `magic-nav-config.service.ts:411` hardcodes `label: 'All Tasks'` while every peer entry uses `T.MH.*`. The key `T.MH.ALL_TASKS` already exists.

4. **Multi-select panel can't display single-string presets.** `task-view-customizer-panel.component.ts:97-113`: if `filter.preset` is a single project title (existing/legacy persisted state on the All Tasks context), `JSON.parse` throws, the `selectedProjectIds` signal stays empty, but the customizer service still filters by title under the hood. User sees no checked boxes and has no way to clear the active filter.

5. **`customizeUndoneTasks(toObservable(this.undoneTasks))` changes the source for every `<work-view>` consumer** (project page, today, tag, …), not just All Tasks. The `toSignal(initialValue: [])` → `toObservable` round-trip emits `[]` synchronously on page (re)construction; combined with the selected-task deselect effect downstream this can cause spurious deselects on slow context switches. Worth verifying on the project/tag pages.

### Simplification (≈80–100 LOC savings + clearer contract)

6. **Drop the JSON-blob-in-`preset` encoding for multi-select** (~25 LOC). Today: a `selectedProjectIds` signal in the panel is kept in sync with `filter.preset` via an `effect()` doing `JSON.parse`; `applyFilter` does the inverse. Two sources of truth, plus a real foot-gun if a user ever types something JSON-array-shaped into the single-select input. Either widen `FilterOption` with a typed `projectIds?: string[]`, or add a distinct `FILTER_OPTION_TYPE.projectMulti`. The `signal + effect` go away entirely and `selectedProjectIds` becomes a `computed()` off `selectedFilter()`.

7. **`setContextKeyOverride` duplicates the constructor's state-load block 3×** (`task-view-customizer.service.ts:122-143`). Extract a private `_loadStateForContext(key, activeType | null)`. Saves ~15 LOC.

8. **`isAllTasks` route detection lives in 3 places** (`main-header.component.ts`, `page-title.component.ts`, `work-context.effects.ts`). The route already declares `data: { page: 'all-tasks' }`. One shared signal (or a tiny `RouteInfoService`) replaces two `NavigationEnd` watchers + regex tables. Saves ~10 LOC.

9. **Duplicated menu-input wrapper in the project filter submenu** (`task-view-customizer-panel.component.html` ~270-334) — both `@if` branches re-declare the same `<div class=\"menu-input-wrapper\"><mat-form-field>…</mat-form-field></div>` chrome. Hoist it; switch only the `[ngModel]` triplet. Saves ~15 LOC.

10. **`isWorkViewPage` has dead code.** `/tasks$/.test('/all-tasks')` is already `true`, so the `|| this.isAllTasks()` clause is unreachable (`page-title.component.ts:245`).

11. **`MainHeaderComponent`'s new `_url` signal + `NavigationEnd` filter** exists only to hide the \"+ Add task\" button on `/all-tasks`. Once #8 lands this collapses to a single computed read; or just let the button work — adding a task with no work-context already routes to Inbox elsewhere in the app.

Net: with #1–#5 you fix real bugs; #6–#11 take ~80-100 LOC off the diff without removing any user-facing functionality.

Issue #7135, opened 2026-04-04
💥 No issueProvider found for id Oc5A3vdisNH2of8qDa1wk


nikoasc:
### Steps to Reproduce

1.
2.
3.

### Url
file:///E:/programm/note/super_productivity/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/index.html#/tag/TODAY/tasks

### Meta Info
META: SP18.0.0 __ Electron – ru – Win32 – ru – UA:Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) superProductivity/18.0.0 Chrome/138.0.7204.251 Electron/37.10.3 Safari/537.36

### Stacktrace
```
file:///E:/programm/note/super_productivity/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/src/app/features/issue/store/issue-provider.selectors.ts:49:12
memoize(function (file:///E:/programm/note/super_productivity/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/@ngrx/store/fesm2022/ngrx-store.mjs:943:23)
memoized (file:///E:/programm/note/super_productivity/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/@ngrx/store/fesm2022/ngrx-store.mjs:819:32)
defaultStateFn (file:///E:/programm/note/super_productivity/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/@ngrx/store/fesm2022/ngrx-store.mjs:847:38)
defaultMemoize(function (file:///E:/programm/note/super_productivity/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/@ngrx/store/fesm2022/ngrx-store.mjs:946:29)
memoized (file:///E:/programm/note/super_productivity/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/@ngrx/store/fesm2022/ngrx-store.mjs:819:32)
source$.pipe(pluck (file:///E:/programm/note/super_productivity/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/@ngrx/store/fesm2022/ngrx-store.mjs:719:43)
[...]

Issue #6270, opened 2026-01-30
💥


w1n-kyt3:
### Steps to Reproduce

I'm not very sure how to reproduce, but it keeps popping up on the web app of my Android phone

I opened this issue on desktop, so the version is off; it's 17.0.11 in the current reported version

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/dbf2f562-cf28-4964-b33e-2a013c36006c" />

### Additional Console Output

[SP-logs.json](https://github.com/user-attachments/files/24953330/SP-logs.json)

### Url
file:///C:/Users/shush/AppData/Local/Programs/superProductivity/resources/app.asar/.tmp/angular-dist/browser/index.html#/config

### Meta Info
META: SP17.0.5 __ Electron – zh-TW – Win32 – zh-TW – UA:Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) superProductivity/17.0.5 Chrome/138.0.7204.251 Electron/37.10.3 Safari/537.36

### Actions Before Error
```
1. Trying to auto-sync when I open up the app (I use Dropbox)
2. End the conflict by choosing between local and remote
3. The app says that my data is corrupted and needs auto-repair
```

johannesjo:
Thank you very much for reporting this!

johannesjo:
This should be dealt with. Thanks again.

PR #8147, opened 2026-06-08
fix(issue): clear stale provider references

cocojojo5213:
@/tmp/pr8147-body.md

### DIFF
--- a/src/app/op-log/validation/data-repair.spec.ts
+++ b/src/app/op-log/validation/data-repair.spec.ts
@@ -1193,6 +1193,7 @@ describe('dataRepair()', () => {
           issueAttachmentNr: 1,
           issueTimeTracked: { ['2026-06-08']: 456 },
           issuePoints: 3,
+          issueLastSyncedValues: { status: 'open' },
         },
         {
           ...DEFAULT_TASK,
@@ -1231,6 +1232,7 @@ describe('dataRepair()', () => {
         issueAttachmentNr: undefined,
         issueTimeTracked: undefined,
         issuePoints: undefined,
+        issueLastSyncedValues: undefined,
       }),
     );
     expect(result.data.task.entities.t2).toEqual(
--- a/src/app/op-log/validation/data-repair.spec.ts
+++ b/src/app/op-log/validation/data-repair.spec.ts
@@ -1177,6 +1177,72 @@ describe('dataRepair()', () => {
     });
   });
 
+  it('should clear non-existent issue provider ids from active tasks', () => {
+    const taskState = {
+      ...mock.task,
+      ...fakeEntityStateFromArray<Task>([
+        {
+          ...DEFAULT_TASK,
+          id: 't1',
+          projectId: INBOX_PROJECT.id,
+          issueId: 'ISSUE-1',
+          issueProviderId: 'NON_EXISTENT_PROVIDER',
+          issueType: 'JIRA',
+          issueWasUpdated: true,
+          issueLastUpdated: 123,
+          issueAttachmentNr: 1,
+          issueTimeTracked: { ['2026-06-08']: 456 },
+          issuePoints: 3,
+        },
+        {
+          ...DEFAULT_TASK,
+          id: 't2',
+          projectId: INBOX_PROJECT.id,
+          issueId: 'ISSUE-2',
+          issueProviderId: 'VALID_PROVIDER',
+          issueType: 'JIRA',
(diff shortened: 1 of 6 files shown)


johannesjo:
This is a really well-constructed fix — thank you for the care here, since stale provider references touch synced state. The layered approach is exactly right: the meta-reducer self-heals at delete time, validation flags orphaned references, and `dataRepair` cleans up any data that's already corrupted on load.

On the sync-correctness specifically (the part I look hardest at): doing the unlink as a meta-reducer in `task-shared-meta-reducers/` rather than an effect fan-out is the correct call — it keeps one user intent as a single reducer pass / one op, and the handler is pure and deterministic. Deriving the unlink set from current state (`issue-provider-shared.reducer.ts:35`) instead of trusting only the action's `taskIdsToUnlink` is a nice robustness win and stays replay-safe: no clocks, no randomness, no remote-action re-triggering, and it uses `taskAdapter.updateMany` so state is replaced, not mutated. Adding `issueLastSyncedValues` to `UNLINKED_PARTIAL_TASK` closes a real gap where stale sync data would otherwise linger. The `data-repair` addition mirrors the existing `_removeNonExistentRepeatCfgIdsFromTasks` pattern (active-tasks-only with the documented archive rationale), and the logging is count/id-only — no user content. I'm confident this is sound.

One tiny nit, purely cosmetic: in `createActionHandlers`, the `deleteIssueProvider` branch destructures the action twice with two separate casts (`issue-provider-shared.reducer.ts:72`) — `const { taskIdsToUnlink } = ...` then `const { issueProviderId } = ...` of the same `ReturnType`. Could be a single `const { issueProviderId, taskIdsToUnlink } = ...` like the `deleteIssueProviders` branch just below it already does. Harmless either way.

Happy to merge as-is. Nice work.

johannesjo:

Thanks for the PR! The layered shape here – meta-reducer self-heal at delete time, validation catch, dataRepair cleanup – is the right one, and deriving the unlink set from state is a real robustness win: it only changes remote replay (both dispatchers already send the full matching set), and the dangling refs concurrent replays can still leave are exactly what your new Checkpoint D rule then heals. Two items need addressing before merge.

### Should fix before merge
- **`src/app/op-log/validation/data-repair.ts:115`** – ordering gap that can turn repair into a permanent sync-failure loop. `_removeNonExistentIssueProviderIdsFromTasks` runs before `_moveArchivedSubTasksToUnarchivedParents` (line 117), which pushes archived subtasks into active state (`taskState.ids.push(t.id)`, plus the "entity is empty" duplicate branch). An archived subtask with a stale `issueProviderId` – tolerated in archives by design (#6270), and your repair deliberately skips archives – whose parent is active gets promoted into active tasks after your cleanup already ran. The repaired state then fails your new check (`is-related-model-data-valid.ts:353`), `validateAndRepair` returns "State still invalid after repair", Checkpoint D returns false before any REPAIR op or dispatch, and the session latch reports SYNC_VALIDATION_FAILED on every subsequent sync. (It takes some NgRx-visible invalidity to trigger the repair in the first place, but once one occurs the loop is armed.) Fix is one line: move the call after the archive-move steps, next to `_removeNonExistentProjectIdsFromTasks` / `_removeNonExistentTagsFromTasks`, which run last for exactly this reason – master moved `_addOrphanedTasksToProjectLists` to the end for the same class of problem in #8780. The same latent hazard already exists one line up for `repeatCfgId`, so placing yours at the end fixes the pattern rather than replicating it.
- **Rebase needed** – the PR conflicts with master (`mergeable: CONFLICTING`). Only `data-repair.ts` and its spec conflict; the repair step list changed upstream (e.g. `_addOrphanedTasksToProjectLists` moved to the end of the pipeline in #8780). The other four files merge cleanly. Re-placing the new step at the end of the list resolves both this and the ordering item above in one go.

### Nits
- **PR description** – the body is the literal string `@/tmp/pr8147-body.md`; the gh CLI file reference didn't expand when the PR was created. Please paste the intended description and add `Closes #7135` so the issue auto-closes on merge.

### Questions
- `restoreTask` (in `task-shared.actions.ts`) restores archived tasks into active state without clearing issue fields, and archives may legitimately hold stale `issueProviderId` refs (#6270). After this PR, restoring such a task makes the next sync's Checkpoint D fail validation and auto-create a full-state REPAIR op. Is one REPAIR op per such restore an accepted cost, or should the restore path also drop stale issue fields?

---

Issue #8148, opened 2026-06-08
CalDAV: true per-occurrence edit/delete for recurring events (RECURRENCE-ID/EXDATE) — ⚠️ high-risk


johannesjo:
Follow-up to #7492.

## Context

The CalDAV calendar plugin now **expands** recurring events into one item per occurrence, and **safely refuses** to edit/delete a single occurrence (because all occurrences map back to one shared `.ics` master, so a naive write mutates the whole series). This issue tracks doing it *for real*: writing a per-instance exception back to the calendar so one occurrence can be moved or removed.

The mechanism exists in the iCal standard:
- **Delete one instance:** add an `EXDATE` for that occurrence to the master VEVENT.
- **Edit one instance:** add a sibling VEVENT with a `RECURRENCE-ID` matching that occurrence, carrying the overridden fields.

(For contrast, the Google Calendar plugin gets this "for free" because it fetches with `singleEvents:true` and Google returns each instance as an individually addressable event. CalDAV occurrences are synthetic — expanded client-side — so there is no per-instance URL to write to; we must edit the shared resource.)

## ⚠️ Why this is dangerous — read before starting

**This writes to users' real, irreplaceable calendars, and every failure mode below is SILENT corruption (no error is thrown). Treat it accordingly.**

1. **The occurrence timestamp is lossy.** Each occurrence id carries `#occ=<ms>` — a single UTC instant (`toJSDate().getTime()`). But a valid `RECURRENCE-ID` / `EXDATE` must reproduce the occurrence's value **in the exact same form the master uses** — UTC (`...Z`), floating local, a named `TZID=`, or all-day (`VALUE=DATE`) — **byte-for-byte**. Reconstructing that from a bare ms is the crux of the whole task. Get it wrong and:
   - the server **silently ignores** the `EXDATE` (the occurrence stays visible → "delete did nothing"), or
   - you create a **ghost/orphan** override whose `RECURRENCE-ID` matches no generated instance (undefined behavior; clients diverge / double-book).
   This is especially easy to get wrong **across DST boundaries** (a one-hour drift turns a valid override into an orphan).

[...]

Issue #7492, opened 2026-05-06
🚨 recurring caldav events parsed as one event


tippfehlr:
### Version Used

18.4.4 AUR package superproductivity

### Current Behavior

When using the CalDav plugin to add a calendar, recurring tasks (weekly for me) are only parsed once.

The `RRULE` parsing in https://github.com/super-productivity/super-productivity/blob/master/packages/plugin-dev/caldav-calendar-provider/src/plugin.ts seems to be missing. I let an LLM write a proof of concept patch for it that does the parsing in the file (and it seems to work). Would you want to parse it directly or should it pull in a library?

### Expected Behavior

The recurring events are expanded into one event per recurrence.

### Steps to Reproduce

1. Enable the CalDav plugin
2. add a calendar with a recurring event
3. find the recurring events

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell
don’t think the logs are relevant (but they can’t hurt, right?)

[issue] 7200000
log.ts:199  [a][IssueProvider/API] Update IssueProvider {issueProvider: {…}, meta: {…}, type: '[IssueProvider/API] Update IssueProvider'}
log.ts:326 [ol] OperationCaptureService: Enqueued operation {actionType: '[IssueProvider/API] Update IssueProvider', changeCount: 0, queueSize: 1}
log.ts:326 [ol] operationCaptureMetaReducer: Enqueued action for operation capture {actionType: '[IssueProvider/API] Update IssueProvider'}
log.ts:326 [ol] OperationCaptureService: Dequeued operation {changeCount: 0, queueSize: 0}
log.ts:326 [ol] incrementVectorClock {clientId: 'B_2zyJ', currentValue: 635, allClients: Array(3)}
log.ts:326 [sync] SyncCredentialStore.load SuperSync object
log.ts:326 [ol] ImmediateUploadService: Starting immediate upload...
log.ts:345 [ol] OperationWriteFlushService: Starting flush. Initial queue size: 0
log.ts:345 [ol] OperationWriteFlushService: Flush complete in 0ms. Initial queue: 0, Final queue: 0
log.ts:345 [ol] OperationLogUploadService: Uploading pending operations via API...
log.ts:326 [sync] SyncCredentialStore.load SuperSync object
log.ts:345 [ol] OperationLogUploadService: Encrypting operation payloads...
log.ts:345 [ol] OperationLogUploadService: Uploading batch of 1 ops via API
[...]

PR #8149, opened 2026-06-08
fix(caldav-plugin): make recurring-occurrence edits/deletes safe and quiet #7492

johannesjo:
Follow-up to #7492. Recurring events are expanded into one item per occurrence, but with CalDAV every occurrence points at one shared `.ics` on the server, so acting on a single occurrence used to hit the whole series.

## Behavior
- **Editing/deleting a task** linked to one occurrence no longer touches the calendar series, and is **silent** — your local change still applies. (Deleting a task shouldn't silently nuke a recurring calendar event, and editing your task shouldn't error about a calendar you weren't acting on.)
- **Explicitly** rescheduling/deleting the occurrence from the agenda right-click menu shows an **honest** "can't change a single occurrence yet — do it in your calendar app" message (no false "deleted/rescheduled").
- `getById` now reports each occurrence's own start/end instead of the master's.
- Single, non-recurring events keep full edit/delete.

## How
The plugin refuses occurrence writes with a marked error (`isExpectedSyncSkip`). The host's two-way sync recognizes the marker and stays silent (you changed your task, not the calendar); explicit agenda actions don't consult it and surface the message.

## Not in scope (tracked in #8148)
True per-occurrence editing (write `RECURRENCE-ID` overrides + `EXDATE` back to the series) — needs an `If-Match`/ETag primitive and recurrence-value preservation first. See #8148 for the risk write-up.

## Tests
Plugin 56/56, host two-way-sync effects spec 33/33 (incl. new suppression tests), typecheck + `checkFile` clean, plugin build OK.

### DIFF
--- a/electron-builder.yaml
+++ b/electron-builder.yaml
@@ -61,6 +61,13 @@ nsis:
   # allow to install to custom location
   oneClick: false
   allowToChangeInstallationDirectory: true
+  # We ship no in-app auto-updater: electron-updater is not a dependency and the
+  # autoUpdater block in electron/start-app.ts is commented out. elevate.exe is
+  # only ever invoked by electron-updater to apply privileged (per-machine)
+  # updates, so electron-builder bundles it here as dead weight that AV engines
+  # frequently false-positive on. Drop it. Safe because perMachine is not true
+  # (with perMachine: true, packElevateHelper: false is ignored). See #8135.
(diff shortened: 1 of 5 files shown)


zenoprax:
Unfortunately I can't test this due to cross-origin complications with the preview deployment. I'll try again once the next release is out.

PR #8150, opened 2026-06-08
fix(redmine): support non-latin issue search

### LINKED ISSUE #4149: Cannot find non-latin tasks in Redmine integration
### Your Environment

- Version used: 12.0.1

### Expected Behavior

Tasks should be searchable by non-latin words/symbols.

### Current Behavior

When I try to search for non-latin words or part of words in Redmine connection nothing is found. But latin words are. If a task name contains both latin and non-latin words, this task can be found by latin and cannot by non-latin query.

Yes.

cocojojo5213:
Fixes #4149.

## Summary
- keep the existing Redmine `/search.json` request as the primary search path
- for non-Latin queries, only fall back to a Redmine subject-contains issue filter when `/search.json` returns no results
- return the original search response if the fallback request fails, avoiding an extra snackbar for the fallback path
- document Redmine non-Latin subject search support in the issue integration comparison

## Validation
- npm run checkFile src/app/features/issue/providers/redmine/redmine-api.service.ts
- npm run checkFile src/app/features/issue/providers/redmine/redmine-api.service.spec.ts
- CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/issue/providers/redmine/redmine-api.service.spec.ts
- git diff --check
- push hook: full lint, package tests, release-notes tests, default Angular suite, and LA timezone Angular suite passed

Note: PR #8072 also touches Redmine search for numeric issue IDs. This PR keeps its fallback scoped to non-Latin subject queries.

### DIFF
--- a/docs/wiki/3.07-Issue-Integration-Comparison.md
+++ b/docs/wiki/3.07-Issue-Integration-Comparison.md
@@ -111,7 +111,7 @@ Every issue provider supports the same basic operations: checking that the integ
 
 - **Issue import:** Issues from projects.
 - **Status transitions / Worklog / Comments / Subtasks / Attachments:** Not supported.
-- **Filtering:** Project-based; scope support.
+- **Filtering:** Project-based; scope support; task search supports non-Latin subject matches.
 - **Auto-import:** Last 100 issues for the project.
 
 ### 9. Trello
--- a/src/app/features/issue/providers/redmine/redmine-api.service.spec.ts
+++ b/src/app/features/issue/providers/redmine/redmine-api.service.spec.ts
@@ -1,17 +1,19 @@
-import { TestBed } from '@angular/core/testing';
 import {
   HttpClientTestingModule,
   HttpTestingController,
+  TestRequest,
 } from '@angular/common/http/testing';
 import { HttpRequest } from '@angular/common/http';
+import { TestBed } from '@angular/core/testing';
+import { SnackService } from '../../../../core/snack/snack.service';
+import { SearchResultItem } from '../../issue.model';
 import { RedmineApiService } from './redmine-api.service';
 import { RedmineCfg } from './redmine.model';
-import { SearchResultItem } from '../../issue.model';
-import { SnackService } from '../../../../core/snack/snack.service';
 
 describe('RedmineApiService', () => {
   let service: RedmineApiService;
   let httpMock: HttpTestingController;
+  let snackService: jasmine.SpyObj<SnackService>;
 
   const mockCfg: RedmineCfg = {
     isEnabled: true,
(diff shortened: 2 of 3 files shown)


johannesjo:
Thanks for this — really nice fix. Keeping the existing `/search.json` request as the primary path and only falling back to the subject-contains filter for non-ASCII queries (with results unchanged for Latin queries and the fallback staying silent on error) is exactly the right shape. The added spec coverage is thorough too.

One tiny, non-blocking observation: the two paths produce slightly different `issueData` shapes — the primary path stores the raw search-result item (`{ id, title, url }`) via `mapRedmineSearchResultItemToSearchResult`, while the fallback stores a full `RedmineIssue` via `mapRedmineIssueToSearchResult` (`src/app/features/issue/providers/redmine/redmine-api.service.ts`). In practice this is harmless: both carry `id` and `title`, and `mergeSearchResults` dedups on `id`, so nothing downstream breaks. Just noting it in case the shapes ever want to converge later.

Happy to merge as-is. Thanks again!

cocojojo5213:
I rebased this onto current `master` and resolved the conflict from the newly merged Redmine issue-id lookup. The search flow now keeps both behaviors:

- numeric queries still do the project-scoped issue-id lookup added upstream
- non-Latin queries still use the subject-contains fallback when full-text search returns no results
- the fallback remains silent on error and does not show a snack

Local verification after the rebase:
- `npm run checkFile src/app/features/issue/providers/redmine/redmine-api.service.ts`
- `npm run checkFile src/app/features/issue/providers/redmine/redmine-api.service.spec.ts`
- `npx prettier --check docs/wiki/3.07-Issue-Integration-Comparison.md`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/issue/providers/redmine/redmine-api.service.spec.ts` (10 specs)
- pre-push hook also completed full lint plus the main/LA timezone Karma suites before the force-push.

johannesjo:
Looks goo to merge after conflicts are fixed.

cocojojo5213:
Rebased this branch onto current `master` and resolved the Redmine conflict.

The conflict was between the newer project/global Redmine issue-id lookup and this PR's non-Latin subject fallback. The rebased version keeps all behaviors together:
- project-scoped providers still use `/projects/<id>/search.json` and `/projects/<id>/issues.json`
- global Redmine mode still uses instance-wide `/search.json` and `/issues.json`
- numeric lookups continue to merge by-id results
- non-Latin queries still fall back to subject contains when full-text search returns no results
- the subject fallback now also has coverage for global mode, so it does not accidentally request `/projects/null/issues.json`

Local verification passed:
- `npm run checkFile src/app/features/issue/providers/redmine/redmine-api.service.ts`
- `npm run checkFile src/app/features/issue/providers/redmine/redmine-api.service.spec.ts`
- `npx prettier --check docs/wiki/3.07-Issue-Integration-Comparison.md`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/issue/providers/redmine/redmine-api.service.spec.ts` (16 SUCCESS)

I also attempted the full pre-push hook. It got through lint, sync-core, sync-providers, release-notes tests, and then failed in the full Angular Karma suite on an unrelated date assertion in `dialog-edit-task-repeat-cfg.component.spec.ts` (`Expected "9" to be "10"`). This PR only touches the Redmine provider/spec and Redmine docs, so I pushed the rebased branch with the focused Redmine verification above.

johannesjo:
Thank you very much! <3

PR #8153, opened 2026-06-08
fix(focus-mode): keep manual breaks running when tracking pauses

### LINKED ISSUE #4152: Manually confirming start of Pomodoro break ends break
### Your Environment

- Version used: 12.0.1
- Operating System and version: macOS 14.7 (Sonoma)
- Desktop Environment: N/A
- Browser Name and version: N/A (Desktop version)

#### Pomodoro Settings

- [X] **Enable Pomodoro Timer**
- [X] Stop time tracking for task on break
- [X] Manually confirm starting next Pomodoro session
- [X] **Manually confirm starting next break**

### Expected Behavior

At the end of a Pomodoro session the user is prompted as follows:

<pre>
     You successfully finished session 1!
  Enjoy yourself, get moving, come back in:
                   5:00
     [Close] [Start Break] [Skip Break]
</pre>

Clicking `Start Break` should start the break timer.

### Current Behavior

Clicking `Start Break` results in:

<pre>
     Your break is done!
        5:00 (stopped)
    [Close] [Back to Work]
</pre>

### Steps to Reproduce

1. Configure Pomodoro settings as above
2. Start tracking a task
3. Wait for the Pomodoro session to end
4. Click `Start Break`

Yes.

### Console Output

```
[Pomodoro] Start Pomodoro Break
	Object i
		isBreakStart: false
		type: "[Pomodoro] Start Pomodoro Break"
		> [[Prototype]]: Object

[Task] UnsetCurrentTask
	Object i
		type: "[Task] UnsetCurrentTask"
		> [[Prototype]]: Object
[Pomodoro] Pause Pomodoro
	Object i
		isBreakEndPause: false
		type: "[Pomodoro] Pause Pomodoro"
		> [[Prototype]]: Object
```

### Error Log (Desktop only)

There are no recent errors in the error log.

cocojojo5213:
Fixes #4152.

## Summary
- start Pomodoro/Flowtime breaks before clearing the active task when break tracking pause is enabled
- move the tracking pause into a `startBreak` effect so break state is already active before `unsetCurrentTask()` fires
- keep tracking-stop sync from immediately pausing a running break when tracking was intentionally paused for that break

## Validation
- npm run checkFile src/app/features/focus-mode/focus-mode-break/focus-mode-break.component.ts
- npm run checkFile src/app/features/focus-mode/focus-mode-break/focus-mode-break.component.spec.ts
- npm run checkFile src/app/features/focus-mode/store/focus-mode.effects.ts
- npm run checkFile src/app/features/focus-mode/store/focus-mode.effects.spec.ts
- CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/focus-mode/store/focus-mode.effects.spec.ts -- --include='src/app/features/focus-mode/focus-mode-break/focus-mode-break.component.spec.ts'
- CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/focus-mode/store/focus-mode.bug-5995.spec.ts
- CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/focus-mode/store/focus-mode.bug-6064.spec.ts
- git diff --check
- push hook: full lint, package tests, release-notes tests, default Angular suite, and LA timezone Angular suite passed

### DIFF
--- a/src/app/features/focus-mode/focus-mode-break/focus-mode-break.component.spec.ts
+++ b/src/app/features/focus-mode/focus-mode-break/focus-mode-break.component.spec.ts
@@ -21,7 +21,6 @@ import { of } from 'rxjs';
 import { TaskService } from '../../tasks/task.service';
 import { FocusMainUIState, FocusModeMode } from '../focus-mode.model';
 import { FocusModeConfig } from '../../config/global-config.model';
-import { unsetCurrentTask } from '../../tasks/store/task.actions';
 
 describe('FocusModeBreakComponent', () => {
   let component: FocusModeBreakComponent;
@@ -161,7 +160,7 @@ describe('FocusModeBreakComponent', () => {
       );
     });
 
-    it('should unset current task before starting break offer when pause-during-break is enabled', () => {
+    it('should start break offer without clearing tracking first when pause-during-break is enabled', () => {
       (mockFocusModeService.mainState as any).set(FocusMainUIState.BreakOffer);
       (mockFocusModeService.isSessionRunning as any).set(false);
       (mockFocusModeService.focusModeConfig as any).set({
@@ -170,7 +169,7 @@ describe('FocusModeBreakComponent', () => {
 
       component.resumeBreak();
 
-      expect(mockStore.dispatch).toHaveBeenCalledWith(unsetCurrentTask());
+      expect(mockStore.dispatch).toHaveBeenCalledTimes(1);
       expect(mockStore.dispatch).toHaveBeenCalledWith(
         startBreak({
           duration: 300000,
--- a/src/app/features/focus-mode/focus-mode-break/focus-mode-break.component.ts
+++ b/src/app/features/focus-mode/focus-mode-break/focus-mode-break.component.ts
@@ -22,7 +22,6 @@ import { TranslatePipe } from '@ngx-translate/core';
 import { TaskTrackingInfoComponent } from '../task-tracking-info/task-tracking-info.component';
 import { toSignal } from '@angular/core/rxjs-interop';
 import { TaskService } from '../../tasks/task.service';
-import { unsetCurrentTask } from '../../tasks/store/task.actions';
 
 @Component({
   selector: 'focus-mode-break',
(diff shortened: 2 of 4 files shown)


johannesjo:
Nice fix — and a well-targeted one. Moving the `unsetCurrentTask` into a dedicated `startBreak`-triggered effect (`pauseTrackingOnBreakStart$`) so the break is already active before tracking clears, plus the extra guard in `syncTrackingStopToSession$` so the app's own break-start unset no longer pauses the running break, is the correct shape for #4152. The new "should NOT pause a running break…" spec captures the regression well.

One tiny, non-blocking note: `pauseTrackingOnBreakStart$` re-reads the config and `currentTaskId` via `withLatestFrom` (`src/app/features/focus-mode/store/focus-mode.effects.ts`), which slightly duplicates the `shouldPauseTracking` check that used to live inline in `autoStartBreakOnSessionComplete$`. Totally negligible here — just flagging it in case the two ever drift.

Looks good to merge. Thanks!

johannesjo:
Thanks a lot for digging into this, @cocojojo5213 — and sorry for the back-and-forth after my earlier approval.

On a closer second look, I don't think this is needed anymore. #4152 was filed against the old `pomodoro` module (v12.0.1), which has since been fully replaced by the new `focus-mode` feature. On current `master` the bug no longer reproduces: both break-start paths already dispatch `unsetCurrentTask()` **before** `startBreak()`. Since `currentTaskId$` is a synchronous store select, the unset resolves while the break timer isn't running yet, so the `timer.isRunning` guard in `syncTrackingStopToSession$` never pauses the break. The original ordering that caused #4152 (`startBreak` → `unset` → pause) simply doesn't exist here.

This PR centralizes the "unset on break start" logic into a single `startBreak`-triggered effect, which is a nice DRY cleanup — but to do so it reorders back to `startBreak` → `unset`, i.e. it re-creates the exact ordering that caused the original bug, and then adds a compensating guard in `syncTrackingStopToSession$`. That guard also partially walks back the #5954 "pause break when tracking stops" behavior for the default `isPauseTrackingDuringBreak: true` config. So it adds surface area in a sync-sensitive effect for no user-visible change.

Given the underlying bug already appears resolved by the rewrite, I'm going to close this one. If you can still reproduce a break stopping immediately on a current version (18.x), please reopen with steps and I'll happily take another look. Thanks again — the change was well-tested and genuinely good work, and I'd love to see more contributions from you.

cocojojo5213:
Thanks for the detailed second look and the context on the focus-mode rewrite. That makes sense to me, and I agree it is better not to add sync-sensitive surface area if the original ordering no longer exists on current master. I'll leave this closed unless I can reproduce it on 18.x with concrete steps. Thanks again for reviewing it.

PR #8154, opened 2026-06-08
fix(nextcloud-deck): load cards when stacks omit them

### LINKED ISSUE #7299: 🚨 Nextcloud deck issue provider doesn't show any tasks
### Version Used

v18.2.5 deb and apk

### Expected Behavior

- I add a Nextcloud Deck Issue provider
- Test connection says it's ok
- Boards list successfully load and I can choose the "import from" and "done" stacks
- Tasks visible on the nextcloud deck web can be searched and imported on SP

### Current Behavior

- I add a Nextcloud Deck Issue provider
- Test connection says it's ok
- Boards list successfully load and I can choose the "import from" and "done" stacks
- Tasks don't show up even when I search them

There are tasks on the board I selected, that's not the problem.

### Steps to Reproduce

- Set up a Nextcloud Deck issue provider
- Try to search tasks

### Console Output

```shell

```

### documentation gap

cocojojo5213:
Fixes #7299.

## Summary
- keep loading Nextcloud Deck stacks as before, but fetch stack details when the stack list omits `cards`
- limit search/auto-import fallback requests to configured import stacks and skip done stacks
- let `getById$` fetch missing stack details across the board so already-linked cards can refresh even outside import stacks
- add focused coverage for missing-card stack responses, existing embedded-card fast path, import-stack filtering, and card lookup by id

## Validation
- npm run checkFile src/app/features/issue/providers/nextcloud-deck/nextcloud-deck-api.service.ts
- npm run checkFile src/app/features/issue/providers/nextcloud-deck/nextcloud-deck-api.service.spec.ts
- CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/issue/providers/nextcloud-deck/nextcloud-deck-api.service.spec.ts
- git diff --check

### DIFF
--- a/src/app/features/issue/providers/nextcloud-deck/nextcloud-deck-api.service.spec.ts
+++ b/src/app/features/issue/providers/nextcloud-deck/nextcloud-deck-api.service.spec.ts
@@ -1,0 +1,196 @@
+import { TestBed } from '@angular/core/testing';
+import {
+  HttpClientTestingModule,
+  HttpTestingController,
+} from '@angular/common/http/testing';
+import { NextcloudDeckApiService } from './nextcloud-deck-api.service';
+import { SnackService } from '../../../../core/snack/snack.service';
+import { NextcloudDeckCfg } from './nextcloud-deck.model';
+
+describe('NextcloudDeckApiService', () => {
+  let service: NextcloudDeckApiService;
+  let httpMock: HttpTestingController;
+
+  const baseUrl = 'https://cloud.example.com/index.php/apps/deck/api/v1.0';
+  const mockCfg: NextcloudDeckCfg = {
+    isEnabled: true,
+    nextcloudBaseUrl: 'https://cloud.example.com',
+    username: 'user',
+    password: 'pass',
+    selectedBoardId: 10,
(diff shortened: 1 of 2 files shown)


johannesjo:
Thanks for this — good fix for the boards whose stack list omits `cards`. Lazily fetching per-stack details only when `cards` is missing (and leaving the embedded-card fast path untouched) is exactly right, and `searchOpenCards$` / auto-import correctly scoping the detail fetch to import stacks via `onlyImportStacks = true` is a nice touch. Solid spec coverage too.

One small, non-blocking observation: `getById$` → `_findCardById` calls `_getStacksWithCards$` without `onlyImportStacks` (`src/app/features/issue/providers/nextcloud-deck/nextcloud-deck-api.service.ts`), so a single card lookup on a cards-omitting board fetches details for every stack rather than just the import stacks. It's an N+1 bounded by stack count and only on servers that omit cards — and it's arguably intentional, since a linked card can live outside the import stacks. If it's easy to scope down (e.g. when the target card is known to be in an import stack), it might be worth it; otherwise I'm fine leaving it as-is.

Happy to merge. Thanks again!

PR #8155, opened 2026-06-08
fix(trello): include workspace boards in picker

### LINKED ISSUE #7317: 🚨 Trello integration doesn't support Workspaces/Organizations
### Version Used

v18.2.5

### Expected Behavior

Description:
Boards inside Trello Workspaces/Organizations are not loaded.

### Current Behavior

The app only queries /members/me/boards which returns empty for workspace boards.
It should also query /organizations/{id}/boards

Workaround: Move boards outside workspace (not ideal for teams)

### Steps to Reproduce

Integrate with trello

### Console Output

```shell

```

### documentation gap

cocojojo5213:
Fixes #7317.

## Summary
- keep loading direct member boards, then best-effort load the user's Trello organizations
- include open boards from each organization in the board picker
- dedupe boards by id and keep direct member boards if organization lookup or a single organization board request fails
- add focused coverage for organization boards, deduplication, and graceful fallback paths

## Validation
- npm run checkFile src/app/features/issue/providers/trello/trello-api.service.ts
- npm run checkFile src/app/features/issue/providers/trello/trello-api.service.spec.ts
- CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/issue/providers/trello/trello-api.service.spec.ts
- git diff --check

### DIFF
--- a/src/app/features/issue/providers/trello/trello-api.service.spec.ts
+++ b/src/app/features/issue/providers/trello/trello-api.service.spec.ts
@@ -112,6 +112,103 @@ describe('TrelloApiService', () => {
     });
   });
 
+  describe('getBoards$', () => {
+    it('should include open boards from Trello organizations', (done) => {
+      service.getBoards$(mockCfg).subscribe((boards) => {
+        expect(boards).toEqual([
+          { id: 'member-board', name: 'Member Board' },
+          { id: 'shared-board', name: 'Shared Board from Member List' },
+          { id: 'org-board', name: 'Organization Board' },
+        ]);
+        done();
+      });
+
+      const memberBoardsReq = httpMock.expectOne(
+        (request) => request.url === 'https://api.trello.com/1/members/me/boards',
+      );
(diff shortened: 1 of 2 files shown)


johannesjo:
Thanks — this is a clean fix. Merging direct member boards with each organization's open boards, deduping by `id`, and keeping it best-effort (`catchError` on both the org lookup and each per-org boards request so a failure never drops the member boards) is exactly the right approach. Bonus points for tightening the previously `any` return type into `TrelloBoardResponse[]`, and the new spec coverage is great.

One tiny, non-blocking note: `forkJoin` fans out the per-organization board requests with no concurrency cap (`src/app/features/issue/providers/trello/trello-api.service.ts`). For typical org counts that's a non-issue; it could just burst a bit for users who belong to many workspaces. Not worth holding the PR for — just flagging in case you ever want a small `mergeMap` concurrency limit there.

Looks good to merge. Thanks!

PR #8159, opened 2026-06-08
fix(tasks): open task menu with native menu key

### LINKED ISSUE #6670: 💡 Open task actions menu with keyboard's menu button
### Problem Statement

Inability to open the task actions menu with a common keyboard button.

Tasks have a menu where several actions are available for the task. This can be opened by right clicking the task on PC, or long tapping it on phone.

Image

### Possible Solution

It would be useful if the menu button on PC keyboards would also open this menu.

_No response_

### Additional context

The menu button is usually found right to the space, and sometimes also AltGr buttons, and left to the right control. Its common for this button to open the context menu relevant to the selection, both on windows and linux systems.

On Wayland based Linux desktop systems, this button produces keycode 135 (symbol Menu) when watching with the `wev` utility.

cocojojo5213:
## Summary
- handle the keyboard Menu/ContextMenu key as a task action-menu shortcut
- support browser and Linux-style ContextMenu key event shapes
- ignore modified Menu key presses so configured shortcuts keep their normal behavior

Fixes #6670

## Test plan
- npm run checkFile src/app/features/tasks/task-shortcut.service.ts
- npm run checkFile src/app/features/tasks/task-shortcut.service.spec.ts
- CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/tasks/task-shortcut.service.spec.ts

### DIFF
--- a/src/app/features/tasks/task-shortcut.service.spec.ts
+++ b/src/app/features/tasks/task-shortcut.service.spec.ts
@@ -312,6 +312,63 @@ describe('TaskShortcutService', () => {
       expect(mockTaskComponent.openDeadlineDialog).toHaveBeenCalled();
     });
 
+    it('should open the task context menu when the native Menu key is pressed', () => {
+      const mockTaskComponent = {
+        task: () => ({ id: 'focused-task-1' }),
+        openContextMenu: jasmine.createSpy('openContextMenu'),
+        taskContextMenu: () => undefined,
+      };
+      mockTaskFocusService.focusedTaskId.set('focused-task-1');
+      mockTaskFocusService.lastFocusedTaskComponent.set(mockTaskComponent);
+
+      const event = createKeyboardEvent('ContextMenu', 'ContextMenu');
+      spyOn(event, 'preventDefault');
+
+      const result = service.handleTaskShortcuts(event);
+
+      expect(result).toBe(true);
+      expect(event.preventDefault).toHaveBeenCalled();
+      expect(mockTaskComponent.openContextMenu).toHaveBeenCalledWith(event);
+    });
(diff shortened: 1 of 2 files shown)


johannesjo:
Thanks for this — opening the task menu with the native Menu/Apps key is a nice accessibility touch, and the implementation is clean and well-scoped. I appreciate the care that went into the test coverage too: the three specs cover the happy path (`ev.key === 'ContextMenu'`), the Linux-style event shape (recognized via the physical `code`), and correctly *not* firing when a modifier is held. That last case is the important one and it's good to see it pinned down.

A couple of things I checked while reviewing:

- The new `isNativeContextMenuKey` (`src/app/features/tasks/task-shortcut.service.ts:14`) is purely additive against the configured `taskOpenContextMenu` combo (default `'Q'`, `src/app/features/config/default-global-config.const.ts:169`), so there's no conflict with the existing shortcut — good.
- It's correctly gated behind a focused task and all the no-modifier checks, and `openContextMenu` is non-destructive, so there's no risk of an accidental fire doing anything a user can't immediately dismiss with Escape.

One small observation, not a blocker: the final `ev.code === 'ContextMenu'` branch (line 19) matches on the *physical* key position regardless of `ev.key`, so in the rare event someone has OS-remapped that physical key to emit a different `ev.key`, the handler would still treat it as the Menu key. In practice this is a non-issue — that key exists almost exclusively to open context menus, remapping it is exceedingly rare, and the worst case is opening a menu the user can dismiss — so I'm comfortable leaving it as-is. I mention it only so it's on the record; I don't think it's worth changing.

Looks good to me. Thanks again!

PR #8161, opened 2026-06-08
fix(jira): paginate auto-import search

### LINKED ISSUE #5578: 🚨 JIRA sync only pulls 100 issues
### Version Used

Version 16.3.5 (16.3.5)

### Expected Behavior

I enabled JIRA sync with a filter that produces 1500 issues. I expect all of them to sync.

### Current Behavior

Only 100 of those are imported/synced to SP. 

### Steps to Reproduce

Add JIRA import with filter with 100+ issues, auto imported into project Foo. Observe only 100 issues imported.

### Console Output

```shell

```

cocojojo5213:
## Summary
- paginate Jira Cloud auto-import searches with search/jql nextPageToken
- paginate Jira Server/Data Center fallback searches with startAt/total
- keep mapping behavior unchanged after collecting all returned issue pages

Fixes #5578

## Test plan
- npm run checkFile src/app/features/issue/providers/jira/jira-api.service.ts
- npm run checkFile src/app/features/issue/providers/jira/jira-api.service.spec.ts
- CHROME_BIN=/snap/bin/chromium npm run test:file src/app/features/issue/providers/jira/jira-api.service.spec.ts

### DIFF
--- a/src/app/features/issue/providers/jira/jira-api.service.spec.ts
+++ b/src/app/features/issue/providers/jira/jira-api.service.spec.ts
@@ -7,7 +7,7 @@ import { SnackService } from '../../../../core/snack/snack.service';
 import { GlobalProgressBarService } from '../../../../core-ui/global-progress-bar/global-progress-bar.service';
 import { BannerService } from '../../../../core/banner/banner.service';
 import { MatDialog } from '@angular/material/dialog';
-import { DEFAULT_JIRA_CFG } from './jira.const';
+import { DEFAULT_JIRA_CFG, JIRA_MAX_AUTO_IMPORT_PAGES } from './jira.const';
 import { JiraCfg } from './jira.model';
 import { formatJiraDate } from '../../../../util/format-jira-date';
 import { JiraIssueOriginal } from './jira-api-responses';
@@ -389,6 +389,36 @@ describe('JiraApiService', () => {
       });
     });
 
+    it('caps Jira Cloud auto-import pagination', (done) => {
+      fetchSpy.and.returnValues(
+        ...Array.from({ length: JIRA_MAX_AUTO_IMPORT_PAGES + 1 }, (_, index) =>
+          jsonResponse({
+            issues: [makeJiraIssue(`TEST-${index + 1}`)],
+            maxResults: 100,
+            nextPageToken: `token-${index + 2}`,
+          }),
+        ),
+      );
+
+      service.findAutoImportIssues$(cfg).subscribe({
+        next: (issues) => {
+          expect(issues.map((issue) => issue.key)).toEqual([
+            'TEST-1',
+            'TEST-2',
+            'TEST-3',
+            'TEST-4',
+            'TEST-5',
+          ]);
+          expect(fetchSpy).toHaveBeenCalledTimes(JIRA_MAX_AUTO_IMPORT_PAGES);
+          expect(requestBodyAt(fetchSpy, JIRA_MAX_AUTO_IMPORT_PAGES - 1)).toEqual(
+            jasmine.objectContaining({ nextPageToken: 'token-5' }),
+          );
+          done();
+        },
+        error: done.fail,
+      });
+    });
+
     it('falls back to Jira Server/DC search pages via startAt', (done) => {
       fetchSpy.and.returnValues(
         jsonResponse({ errorMessages: ['not found'] }, 404),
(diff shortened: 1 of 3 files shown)


johannesjo:
Thanks for tackling this, @cocojojo5213 — the pagination logic itself looks correct. You handle both the Jira Cloud `/search/jql` (`nextPageToken`) and the Server/DC `/search` (`startAt`/`total`) shapes, and the new specs cover both paths nicely. Appreciate the thorough test plan too.

I do have one concern I'd like to resolve before merging.

## Unbounded fetch on every auto-import poll

Previously `findAutoImportIssues$` issued a **single** request capped at `maxResults: JIRA_MAX_RESULTS` (100) — so one auto-import returned at most one page / 100 issues. With this change, `_fetchAllJiraSearchPages$` now follows **every** page until exhaustion with no overall cap:

- Cloud path: recurses as long as `previousPage.nextPageToken` is set
  (`src/app/features/issue/providers/jira/jira-api.service.ts`, `_getNextJiraSearchPageBody`, the `isCloudJqlSearch` branch).
- Server/DC path: recurses while `startAt < previousPage.total`
  (same method, the `startAt`/`total` branch).

The thing that worries me is *where* this runs. Auto-import is driven by a repeating poll timer (`poll-to-backlog.effects.ts` → `IssueService.checkAndImportNewIssuesToBacklogForProject` → `getNewIssuesToAddToBacklog` → `findAutoImportIssues$`), firing every `JIRA_POLL_INTERVAL` (5 minutes) for each configured provider. The default auto-import JQL is an open query like `project = X ORDER BY updated DESC`, which on a large project can return thousands of issues. So instead of one bounded request per poll, a user with a big project would now fetch the *entire* result set — dozens of pages — every 5 minutes, indefinitely.

Worth noting: the de-duplication against already-imported issues happens client-side *after* the full fetch (in `IssueService.checkAndImportNewIssuesToBacklogForProject`), so it prevents re-adding tasks but does not reduce the per-poll network/CPU cost. The loop does terminate correctly — this isn't a hang, it's a resource/cost concern (request volume and Jira API load).

### Ask before merge

Could you add an overall cap so a runaway query can't flood the API? Either:

- a max number of pages (e.g. 5 → up to ~500 issues at `maxResults` 100), or
- a max total issue count, stopping accumulation once it's reached.

A `MAX_AUTO_IMPORT_PAGES` const of around 5 (defaulting conservatively) feels like a reasonable default to me — happy to discuss the exact number. The auto-import backlog import doesn't really need to ingest unbounded history on every poll; bounding it keeps the original safety property while still fixing the under-fetch this PR targets.

### Minor (optional)

The page-following is implemented as `concatMap` recursing into itself (`_fetchRemainingJiraSearchPages$`), which builds a fairly deep nested observable chain for large result sets. RxJS `expand` (or a plain iterative accumulator) would express the same "keep fetching until no next page" loop more flatly. Not blocking — just a readability note, and it pairs naturally with adding the cap above.

## CI

The red **Linux Packaged Electron App** check is an unrelated infra flake — it failed in the `Install npm packages` step (before any of your code builds) with `got` returning HTTP 500 while downloading the Electron binary from the registry. The `Tests` job passed. No action needed on your end for that one.

Thanks again — once the cap is in I think this is good to go.

cocojojo5213:
Thanks, that concern makes sense. I pushed follow-up commit `aff2f10` adding a conservative auto-import pagination cap: each poll now fetches at most `JIRA_MAX_AUTO_IMPORT_PAGES` pages, currently 5, so a broad JQL cannot walk an unbounded Jira result set every 5 minutes.\n\nThe cap applies to both paths:\n- Jira Cloud `/search/jql` pagination via `nextPageToken`\n- Jira Server/DC `/search` pagination via `startAt` / `total`\n\nI also added regression coverage proving both paths stop at the fifth page even if the response indicates more pages are available.\n\nLocal verification passed:\n- `npm run checkFile src/app/features/issue/providers/jira/jira-api.service.ts`\n- `npm run checkFile src/app/features/issue/providers/jira/jira-api.service.spec.ts`\n- `npm run checkFile src/app/features/issue/providers/jira/jira.const.ts`\n- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/issue/providers/jira/jira-api.service.spec.ts` (20 specs)\n- `git diff --check`

johannesjo:
Thank you very much!

PR #8162, opened 2026-06-08
fix(tasks): move paste image hint to tooltip

### LINKED ISSUE #8136: 💡 Move the hint "Ctrl+V to paste image" to a tooltip

cocojojo5213:
## Summary
- show the paste-image shortcut hint as a tooltip on the attachment entry points
- use platform-specific copy for Ctrl+V vs Cmd+V
- keep the visible attachment labels unchanged

Fixes #8136

## Test plan
- npm run checkFile src/app/features/tasks/task-detail-panel/task-detail-panel.component.ts
- npm run checkFile src/app/t.const.ts
- npm run int:test
- CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/tasks/task-detail-panel/task-detail-panel.component.spec.ts
- git diff --check


johannesjo:
Thanks — nice little fix. Moving the paste hint into a tooltip keeps the visible attachment labels clean while still surfacing the shortcut, and the platform-specific copy (Cmd+V vs Ctrl+V via `IS_MAC`) is a thoughtful touch. The new keys are correctly registered in both `src/app/t.const.ts` and `src/assets/i18n/en.json`, and `MatTooltip` was already imported in the component so nothing extra to wire up.

One tiny, non-blocking note: the second `[matTooltip]` is applied to the `<task-detail-item>` host (`src/app/features/tasks/task-detail-panel/task-detail-panel.component.html:295`) rather than a native element. That works fine since the directive attaches to any host element, and duplicating the binding across the two entry points is perfectly fine here — just flagging it in case it ever feels worth factoring into the shared item component.

Looks good to merge. Thanks!

johannesjo:
Thank you very much! <3

PR #8163, opened 2026-06-08
fix(tasks): clear Kanban in-progress tag on completion

### LINKED ISSUE #7577: 🚨 The task disappears when I mark it as completed - Continuation
### Version Used

v18.5.0 macOS

### Current Behavior

I recently reported a bug (#7498) where a task disappears in the Boards view when I click the circle on the task (as opposed to manually moving it to the Done column). A commit was subsequently made to fix the issue (6bec7fb). I updated to version 18.5.0. I noticed that the “in-progress” tag is still retained when moving a task to the Done column. However, if you move the task from In Progress to Open, the tag is removed.

I would strongly recommend removing the “in-progress” tag when moving a task to the Done column. This would resolve the issue for existing boards, and I think it is more logically consistent to state that a completed task is not “in-progress.”

Additionally: If the tag appears in the “Done” column and is not explicitly excluded in the panel configuration, it will be visible throughout the column. You can avoid that as well.

At this point, thank you very much for your prompt response to the previous message.

### Expected Behavior

...

### Steps to Reproduce

...

### Console Output

```shell
...
```

### Documentation

cocojojo5213:
Fixes #7577

## Summary
- remove the default Kanban `KANBAN_IN_PROGRESS` system tag when a task is marked done
- run the cleanup in the shared task update reducer so checkbox, context-menu, swipe, drag/drop, sync replay, and other completion paths use the same behavior
- keep the tag unchanged when a task is marked undone
- add reducer regression coverage for task tag IDs and tag task IDs staying in sync

## Testing
- npm run checkFile src/app/root-store/meta/task-shared-meta-reducers/task-shared-crud.reducer.ts
- npm run checkFile src/app/root-store/meta/task-shared-meta-reducers/task-shared-crud.reducer.spec.ts
- CHROME_BIN=/snap/bin/chromium npm run test:file src/app/root-store/meta/task-shared-meta-reducers/task-shared-crud.reducer.spec.ts

### DIFF
--- a/src/app/root-store/meta/task-shared-meta-reducers/task-shared-crud.reducer.spec.ts
+++ b/src/app/root-store/meta/task-shared-meta-reducers/task-shared-crud.reducer.spec.ts
@@ -11,6 +11,7 @@ import { Project } from '../../../features/project/project.model';
 import { WorkContextType } from '../../../features/work-context/work-context.model';
 import { Action, ActionReducer } from '@ngrx/store';
 import { getDbDateStr } from '../../../util/get-db-date-str';
+import { IN_PROGRESS_TAG } from '../../../features/tag/tag.const';
 import {
   createBaseState,
   createMockTag,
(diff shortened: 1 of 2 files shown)


johannesjo:
Thanks for this fix! Routing the Kanban in-progress tag cleanup through the shared task update reducer is exactly the right place to do it, so all the completion paths (checkbox, context menu, swipe, drag/drop, sync replay) stay consistent. Nice touch adding the reducer regression coverage for tag IDs and tag task IDs staying in sync, and keeping the tag untouched on un-done. The code looks good to me.

The red CI here is not caused by your change. The only failing job is "Tests", and the two failures are both in the unrelated, known-flaky `task.service.spec.ts › TaskService › focusTaskById` spec:

```
TaskService focusTaskById
  Expected <input id="some-input"> to be <div id="t-X" tabindex="0">.
  at src/app/features/tasks/task.service.spec.ts:919:38
  at src/app/features/tasks/task.service.spec.ts:910:38
```

These are `document.activeElement` assertions that flake from a focus/spec-leak in the second timezone run (the first run here passed with everything green). It's something we're actively de-flaking on master, and it's entirely unrelated to the meta-reducer files this PR touches.

Could you re-run the Tests job (or push a trivial rebase on latest master)? It should go green, and then this is ready to merge. Thanks again!

cocojojo5213:
Rebased this branch onto latest master and pushed `1c36cbff`, which triggered a fresh CI run.\n\nLocal verification passed:\n- `npm run checkFile src/app/root-store/meta/task-shared-meta-reducers/task-shared-crud.reducer.ts`\n- `npm run checkFile src/app/root-store/meta/task-shared-meta-reducers/task-shared-crud.reducer.spec.ts`\n- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/root-store/meta/task-shared-meta-reducers/task-shared-crud.reducer.spec.ts`\n- `git diff --check upstream/master..HEAD`

PR #8166, opened 2026-06-08
feat(calendar): add Google Calendar past sync window

### LINKED ISSUE #7743: feat(calendar): optional past-events history window for Google Calendar plugin
Today the Google Calendar plugin only fetches events forward from `now` (`packages/plugin-dev/google-calendar-provider/src/plugin.ts:246-251`):

\`\`\`ts
const syncRangeWeeks = parseInt(cfg.syncRangeWeeks || '', 10) || 2;
const now = new Date();
const timeMin = now.toISOString();
const timeMax = new Date(now.getTime() + syncRangeWeeks * 7 * 24 * 60 * 60 * 1000).toISOString();
\`\`\`

The local cache (`LS.CAL_EVENTS_CACHE`, written in `src/app/features/calendar-integration/calendar-integration.service.ts:186`) is replaced on every successful poll with this forward-only window, so past events drop out continuously.

Users in discussion #7738 expected the calendar view to behave more like a rolling history they can refer back to for reflections / notes — and were surprised that past events disappear.

**Proposal**

Add an opt-in `syncRangePastWeeks` config field (default `0`, i.e. preserve current behaviour) so users can extend `timeMin` backwards by N weeks. Cap it sanely (say 12 weeks) to keep API quota and IndexedDB pressure reasonable.

Trade-offs:
- Default `0` keeps API quota unchanged for users who don't opt in.
- A larger `timeMin` window means more events per poll; combined with `syncRangeWeeks` ahead, the total cached set could grow noticeably. Worth measuring on a heavy calendar.
- Won't help with events from before the feature was enabled — Google Calendar API still has the data, but the user has to opt in *before* the events become "past" for them to survive.

Background: discussion #7738.

cocojojo5213:
Fixes #7743.

## Summary
- add an advanced Google Calendar `syncRangePastWeeks` config field, defaulting to the existing forward-only behavior
- cap the past-history window at 12 weeks and increase auto-sync `maxResults` only when past history is enabled
- document the advanced option in the task integrations wiki page

## Validation
- `npm run checkFile packages/plugin-dev/google-calendar-provider/src/plugin.ts`
- `npm run checkFile packages/plugin-dev/google-calendar-provider/src/plugin.spec.ts`
- `cd packages/plugin-dev/google-calendar-provider && npm run typecheck`
- `cd packages/plugin-dev/google-calendar-provider && CHROME_BIN=/snap/bin/chromium npm test -- src/plugin.spec.ts` (38 passed)
- `cd packages/plugin-dev/google-calendar-provider && npm run build`

### DIFF
--- a/docs/wiki/2.07-Manage-Task-Integrations.md
+++ b/docs/wiki/2.07-Manage-Task-Integrations.md
@@ -17,6 +17,10 @@ The **issue provider panel** (Integrations & Calendars) is a side panel that lis
 3. **For a calendar:** Under **Connect Calendar**, click **Google Calendar**, **Outlook 365**, or **iCal Other**. A configuration dialog opens (e.g. URL for iCal, or OAuth for Google/Outlook). Fill in the details and save.
 4. **For an issue provider:** Under **Setup Issue Provider**, click the provider you want: **Jira**, **Trello**, **GitHub**, **Redmine**, **GitLab**, **CalDAV**, **Open Project**, **Gitea**, **Linear**, or **ClickUp**. A configuration dialog opens. Enter the required fields (host URL, token or password, project/repo ID, etc. — the exact fields depend on the provider). Save to create the integration.
 
+Google Calendar normally syncs upcoming events only. In advanced configuration
+you can set **Past events range (weeks)** to keep a limited rolling history in
+the Schedule view.
+
 ### Calendar Integrations: iCal Vs CalDAV (e.g. Nextcloud)
 
 **iCal** and **CalDAV** are two different integrations and need **different link types**:
--- a/packages/plugin-dev/google-calendar-provider/src/plugin.spec.ts
+++ b/packages/plugin-dev/google-calendar-provider/src/plugin.spec.ts
@@ -1,5 +1,8 @@
 import { describe, it, expect, beforeAll, beforeEach, afterEach, vi } from 'vitest';
-import type { IssueProviderPluginDefinition } from '@super-productivity/plugin-api';
+import type {
+  IssueProviderPluginDefinition,
+  PluginHttp,
+} from '@super-productivity/plugin-api';
 
 let definition: IssueProviderPluginDefinition;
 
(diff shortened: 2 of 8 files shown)


johannesjo:
Thanks — this is a well-rounded feature. Defaulting `syncRangePastWeeks` to `0` keeps the existing forward-only behavior, the 12-week cap (`MAX_SYNC_RANGE_PAST_WEEKS`) is a sensible guardrail, and the ISO timestamp math stays UTC-safe via `MS_PER_WEEK`. Putting it behind `advanced: true` and the input `pattern: '^[0-9]*$'` is the right call, and the new spec coverage for both the forward-only default and the capped past window is great.

One minor, non-blocking note: the `SYNC_MAX_RESULTS_PER_WEEK = 50` heuristic that scales `maxResults` in `getSyncMaxResults` (`packages/plugin-dev/google-calendar-provider/src/plugin.ts:255`) is a bit of a magic number. It's safely bounded though — clamped between `DEFAULT_SYNC_MAX_RESULTS` (100) and `MAX_SYNC_MAX_RESULTS` (500), both well under Google's per-request limit — so there's no quota or pagination risk. Just noting it in case the per-week estimate ever wants a comment or a tweak down the line.

Looks good to merge. Thanks!

johannesjo:
Thanks — the plugin code itself is clean, well-tested, and correctly bounded (opt-in, 12-week cap, default forward-only behavior preserved).

Before this can land I'd like to settle a premise question: where do the fetched past events actually get displayed? As far as I can trace:

- `schedule.component.ts` blocks navigating the Schedule into the past,
- `planner-calendar-nav.component.ts` `_isAtPastLimit()` blocks Planner navigation before the current week,
- `calendar-integration.service.ts` `_getCalProviderFromCache` drops anything older than ~1 week on cache reload.

So a 12-week window is fetched but only ~1 week of it is ever viewable — i.e. extra Google API quota/storage without delivering the "rolling history" goal from #7743. Could you confirm which view renders events more than a week into the past? If none today, the view/cache layer needs to change first (or alongside this) for the wider fetch to be useful.

Separately, the branch now conflicts with master (#8171) and needs a rebase.

cocojojo5213:
Thanks for the careful premise check. You were right: with the previous branch, the wider fetch window was not useful enough because the app only kept/displayed a narrow past slice.\n\nI rebased this on current master in 2ad9c2c32 and adjusted the scope so the past window has a concrete visible path:\n\n- kept Google Calendar default behavior forward-only; the past range remains opt-in and capped at 12 weeks\n- kept #8171's managed-event filtering while resolving the rebase conflict\n- extended calendar integration cache retention to the same 12-week capped history, so fetched past events are not dropped after reload\n- allowed the Schedule view to navigate back within that 12-week history window and clamp at the boundary\n- updated the docs to describe this as Schedule history specifically\n\nI intentionally did not expand Planner history navigation here, because that looked like a larger behavior change than this PR needs. With this follow-up, events older than one week are inspectable in the Schedule view when the advanced past range is enabled.\n\nLocal verification passed:\n- `npm run checkFile packages/plugin-dev/google-calendar-provider/src/plugin.ts`\n- `npm run checkFile packages/plugin-dev/google-calendar-provider/src/plugin.spec.ts`\n- `npm run checkFile src/app/features/calendar-integration/calendar-integration.service.ts`\n- `npm run checkFile src/app/features/calendar-integration/calendar-integration.service.spec.ts`\n- `npm run checkFile src/app/features/schedule/schedule/schedule.component.ts`\n- `npm run checkFile src/app/features/schedule/schedule/schedule.component.spec.ts`\n- `npm run checkFile src/app/features/schedule/schedule/schedule.component.html`\n- `npx prettier --check docs/wiki/2.07-Manage-Task-Integrations.md src/app/features/schedule/schedule/schedule.component.html`\n- `npm run typecheck` and `npm run build` in `packages/plugin-dev/google-calendar-provider`\n- `npm run test` in `packages/plugin-dev/google-calendar-provider` (40 passed)\n- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/calendar-integration/calendar-integration.service.spec.ts` (74 SUCCESS)\n- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/schedule/schedule/schedule.component.spec.ts` (46 SUCCESS)\n- `git merge-tree --write-tree upstream/master HEAD`\n- `git diff --check upstream/master..HEAD`

johannesjo:
Apologies — I need to change my position again after tracing the render path more carefully, and the answer isn't where I landed last time.

**My concern: even with this PR, there's no coherent place where past calendar events render.**

Calendar events feed Schedule via `createBlockerBlocksForCalendarEvents` → `create-blocked-blocks-by-day-map.ts`, keyed on each event's own date (`getDbDateStr(block.start)`). With `dayDates` only ever covering today + forward, past-dated blocker entries land in map slots that `create-schedule-days.ts:107` never iterates. Multi-day events spanning yesterday→today are already handled by the existing spawn loop at `create-blocked-blocks-by-day-map.ts:67-91` under the current 1-week retention.

The Schedule past-navigation in this PR creates a render path, but Schedule is a planning engine, not a journal — `create-schedule-days.ts` still carries unscheduled tasks forward (line 117), allocates time-budget from `now` (line 89), and fires repeat-cfg projections (line 100) on past days. So past weeks would render a counterfactual "what we'd plan back then" rather than what actually happened. No completed tasks, no time-tracked work, no archived items. The Worklog already does real history. Shipping Schedule-as-history would create a confusing second answer.

And without the Schedule nav change, the plugin field on its own does nothing user-visible — it just uses more API quota and IndexedDB to fetch events that have no slot to render in. That's worse than not landing it: a user toggles it on, sees no change, files a follow-up issue.

**I think the right call is to close this and open a design discussion for #7743 first.**

The honest paths forward, in rough order of effort:

1. **Reframe the issue smaller.** If the actual pain is "a meeting from this morning disappears from today's view," that's a much narrower fix in cache retention, no plugin changes needed.
2. **Extend Worklog** to overlay calendar events on past days — closest to existing semantics.
3. **A new Calendar History / Journal view** that renders past calendar events alongside time-tracked work and completed tasks — the bigger feature implied by the original ask.

Sorry for the back-and-forth on this one — I should have traced the render path before the first round of feedback. The plugin code itself is genuinely well-done and would be reusable if any of the above paths land. Happy to keep this branch around and reopen if option 1 turns out to be what users actually want.

PR #8172, opened 2026-06-08
fix(config): preserve hidden flowtime settings

cocojojo5213:
Part of #7581

## Summary
- preserve hidden Flowtime break settings when breaks are disabled
- restore the saved break mode, percentage, and rules when Flowtime break settings are shown again
- add a regression spec for saving disabled Flowtime breaks without dropping hidden values

## Tests
- npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.ts
- npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts
- CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts

## Context
#7581 also includes broader Flowtime defaults/product behavior requests. This PR only covers the narrow config-preservation bug for existing hidden settings.

### DIFF
--- a/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.html
+++ b/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.html
@@ -5,7 +5,7 @@
       [fields]="fields()"
       [form]="form"
       [model]="model()"
-      (modelChange)="model.set($event)"
+      (modelChange)="updateModel($event)"
     ></formly-form>
   </form>
   <div class="dialog-actions">
--- a/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts
+++ b/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts
@@ -270,6 +270,49 @@ describe('DialogFlowtimeSettingsComponent', () => {
     expect(component.model().breakRules).toEqual(initialRules);
   });
 
+  it('should ignore partial hidden rule rows emitted while Ratio-based is visible', () => {
+    fixture.destroy();
+    globalConfigServiceMock.cfg.and.returnValue({
+      flowtime: {
+        isBreakEnabled: false,
+        breakMode: 'ratio',
+        breakPercentage: 20,
+        breakRules: [],
+      },
+    });
+    fixture = TestBed.createComponent(DialogFlowtimeSettingsComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+
+    const initialRules = component.model().breakRules;
+    const enabledRatioModel = component.model();
+    enabledRatioModel.isBreakEnabled = true;
+    enabledRatioModel.breakRules = [
+      {
+        minDuration: '' as unknown as number,
+        maxDuration: '' as unknown as number,
+        breakDuration: 5,
+      },
+    ];
+    component.updateModel(enabledRatioModel);
+
+    const restoredRuleModel = component.model();
+    restoredRuleModel.breakMode = 'rule';
+    restoredRuleModel.breakRules = [
+      {
+        minDuration: '' as unknown as number,
+        maxDuration: '' as unknown as number,
+        breakDuration: 5,
+      },
+    ];
+    component.updateModel(restoredRuleModel);
+    fixture.detectChanges();
+
+    expect(component.model().breakRules).toEqual(initialRules);
+    const firstRule = (component.form.get('breakRules') as any).at(0);
+    expect(firstRule.value).toEqual(initialRules![0]);
+  });
+
   describe('save()', () => {
     it('should convert minutes back to ms and save the config', () => {
       component.save();
--- a/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts
+++ b/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts
@@ -313,6 +313,32 @@ describe('DialogFlowtimeSettingsComponent', () => {
     expect(firstRule.value).toEqual(initialRules![0]);
   });
 
+  it('should allow deleting restored rule rows after switching back to Rule-based', () => {
+    component.updateModel({
+      ...component.model(),
+      breakMode: 'ratio',
+      breakRules: [],
+    });
+    component.updateModel({
+      ...component.model(),
+      breakMode: 'rule',
+      breakRules: [
+        {
+          minDuration: '' as unknown as number,
+          maxDuration: '' as unknown as number,
+          breakDuration: 5,
+        },
+      ],
+    });
+
+    component.updateModel({
+      ...component.model(),
+      breakRules: [],
+    });
+
+    expect(component.model().breakRules).toEqual([]);
+  });
+
   describe('save()', () => {
     it('should convert minutes back to ms and save the config', () => {
       component.save();
@@ -384,6 +410,24 @@ describe('DialogFlowtimeSettingsComponent', () => {
         breakPercentage: 20,
         breakRules: [{ minDuration: 0, maxDuration: 1500000, breakDuration: 300000 }],
       });
+    });
+
+    it('should persist rule edits made before disabling breaks', () => {
+      component.updateModel({
+        ...component.model(),
+        breakRules: [{ minDuration: 0, maxDuration: 30, breakDuration: 10 }],
+      });
+      component.updateModel({
+        ...component.model(),
+        isBreakEnabled: false,
+      });
+
+      component.save();
+      const savedConfig =
+        globalConfigServiceMock.updateSection.calls.mostRecent().args[1];
+      expect(savedConfig.breakRules).toEqual([
+        { minDuration: 0, maxDuration: 30 * 60000, breakDuration: 10 * 60000 },
+      ]);
     });
   });
 
--- a/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts
+++ b/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts
@@ -125,6 +125,48 @@ describe('DialogFlowtimeSettingsComponent', () => {
     expect(component.model().breakRules).toEqual(initialRules);
   });
 
+  it('should keep rule fields when Formly mutates the model object while switching modes', () => {
+    fixture.destroy();
+    globalConfigServiceMock.cfg.and.returnValue({
+      flowtime: {
+        isBreakEnabled: false,
+        breakMode: 'ratio',
+        breakPercentage: 20,
+        breakRules: [],
+      },
+    });
+    fixture = TestBed.createComponent(DialogFlowtimeSettingsComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+
+    const initialRules = component.model().breakRules;
+    const enabledModel = component.model();
+    enabledModel.isBreakEnabled = true;
+    component.updateModel(enabledModel);
+
+    const ruleModel = component.model();
+    ruleModel.breakMode = 'rule';
+    component.updateModel(ruleModel);
+
+    const ratioModel = component.model();
+    ratioModel.breakMode = 'ratio';
+    ratioModel.breakRules = [];
+    component.updateModel(ratioModel);
+
+    const restoredRuleModel = component.model();
+    restoredRuleModel.breakMode = 'rule';
+    restoredRuleModel.breakRules = [
+      {
+        minDuration: '' as unknown as number,
+        maxDuration: '' as unknown as number,
+        breakDuration: '' as unknown as number,
+      },
+    ];
+    component.updateModel(restoredRuleModel);
+
+    expect(component.model().breakRules).toEqual(initialRules);
+  });
+
   describe('save()', () => {
     it('should convert minutes back to ms and save the config', () => {
       component.save();
--- a/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts
+++ b/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts
@@ -101,6 +101,30 @@ describe('DialogFlowtimeSettingsComponent', () => {
     expect(component.model().breakRules).toEqual(initialRules);
   });
 
+  it('should keep rule fields when Formly emits blank strings while restoring the rule section', () => {
+    const initialRules = component.model().breakRules;
+
+    component.updateModel({
+      ...component.model(),
+      breakMode: 'ratio',
+      breakRules: [],
+    });
+
+    component.updateModel({
+      ...component.model(),
+      breakMode: 'rule',
+      breakRules: [
+        {
+          minDuration: '' as unknown as number,
+          maxDuration: '' as unknown as number,
+          breakDuration: '' as unknown as number,
+        },
+      ],
+    });
+
+    expect(component.model().breakRules).toEqual(initialRules);
+  });
+
   describe('save()', () => {
     it('should convert minutes back to ms and save the config', () => {
       component.save();
--- a/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts
+++ b/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts
@@ -75,6 +75,32 @@ describe('DialogFlowtimeSettingsComponent', () => {
     expect(fields.find((field) => field.key === 'breakRules')?.resetOnHide).toBe(false);
   });
 
+  it('should keep rule fields while switching Rule-based -> Ratio-based -> Rule-based', () => {
+    const initialRules = component.model().breakRules;
+
+    component.updateModel({
+      ...component.model(),
+      breakMode: 'ratio',
+      breakRules: [],
+    });
+
+    expect(component.model().breakRules).toEqual(initialRules);
+
+    component.updateModel({
+      ...component.model(),
+      breakMode: 'rule',
+      breakRules: [
+        {
+          minDuration: null as unknown as number,
+          maxDuration: null,
+          breakDuration: null as unknown as number,
+        },
+      ],
+    });
+
+    expect(component.model().breakRules).toEqual(initialRules);
+  });
+
   describe('save()', () => {
     it('should convert minutes back to ms and save the config', () => {
       component.save();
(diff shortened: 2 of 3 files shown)


cocojojo5213:
Thanks for the clear repro steps. I pushed a follow-up in `7e3fe3e` that keeps the hidden Flowtime form fields from being reset while switching between Ratio-based and Rule-based before saving.\n\nLocal verification passed:\n- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.ts`\n- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts`\n- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts` (11 specs)

johannesjo:
Thanks for this — it's a clean fix for an annoying data-loss bug. Seeding the form from `{ ...DEFAULT_GLOBAL_CONFIG.flowtime, ...cfg?.flowtime }` and stashing it in `_initialFlowtimeConfig` (`src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.ts:196`) means that when a hidden break field can't be read back from the form, `save()` falls back to the value the dialog opened with instead of wiping `breakMode`/`breakPercentage`/`breakRules`. Routing it through `updateSection('flowtime', ...)` keeps it a single config op, so it's sync-safe. Pulling the ms-conversion out into `_toBreakRulesInMs` is a nice readability win too, and the new regression spec captures the disabled-breaks case well.

One tiny, non-blocking note: `breakRules` is only re-derived from the form when `isBreakEnabled && breakMode === 'rule'` (`:235`). So if someone edits rules, switches to `ratio` mode, and saves, the edited rules are discarded in favor of the dialog-open snapshot. That's fine in practice — the rules field is hidden in ratio mode and the previously-saved rules are preserved either way — just worth being aware of.

Unrelated heads-up on CI: the red `Tests` run is the known flaky `TaskService › focusTaskById` spec (`src/app/features/tasks/task.service.spec.ts:910`/`919`), which has nothing to do with this PR — it doesn't touch any task code. A re-run should go green.

Looks good to merge. Thanks!

cocojojo5213:
Thanks for the follow-up repro. I pushed `054aa670c` to cover the Rule-based -> Ratio-based -> Rule-based path as well.

The form now keeps hidden Flowtime rule rows while switching modes before saving, while still letting visible fields validate normally.

Local verification passed:
- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.ts`
- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts`
- `npx prettier --check src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.html`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts` (12 SUCCESS)
- `git diff --check`

cocojojo5213:
Thanks, you were right. I missed the variant where Formly restores the number inputs as blank strings instead of null values.

I pushed `a31b4694b` to handle that path as well. The Flowtime rule rows are now treated as temporarily blank during the mode visibility switch, so the previous rule values are kept instead of being replaced by empty fields.

Local verification passed:
- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.ts`
- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts` (13 SUCCESS)
- `git diff --check`

cocojojo5213:
No revert is needed. The 054aa67 change is still part of the fix: it keeps the rule rows around while the Rule-based section is hidden during the mode switch. The later a31b4694b commit only adds the missing Formly case where the number inputs come back as blank strings instead of null values.

So the current head keeps both pieces: preserve the hidden rule rows, and treat temporarily blank restored rows as empty so they do not replace the previous rule values. I am still watching the fresh GitHub checks for a31b4694b.

cocojojo5213:
Thanks for testing it again. I found the missing case and pushed `6cb2faef8`.

The previous fix still assumed the previous Formly model object was an untouched snapshot. In the actual dialog flow, Formly can mutate the same model object while switching modes, so the "previous" rule rows could already be blank by the time `updateModel()` tried to preserve them.

This commit keeps a separate copy of the last non-empty Flowtime rule rows and restores from that when the Rule-based section is hidden/restored. I also added a regression spec that follows the exact path from your repro starting from Flowtime breaks disabled/default Ratio-based:

1. Enable Flowtime breaks
2. Select Rule-based
3. Select Ratio-based
4. Select Rule-based

Local verification passed:
- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.ts`
- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts` (14 SUCCESS)
- `git diff --check`

cocojojo5213:
Thanks for testing `6cb2fae` and for the detailed 1-rule/2-rule breakdown.

I found the remaining gap and pushed `baec64c33`. The previous follow-up restored whole blank rule rows during the mode switch, but Formly can also emit a delayed partially restored payload where `breakDuration` is present while `minDuration` and `maxDuration` are still blank. That partial payload could overwrite the cached rule rows.

The new commit keeps the restore window open while those restorable rule fields are still blank, and fills blank fields from the last complete rule rows. I added regressions for both cases you described:

1. one or more rule rows coming back partially blank
2. a delayed Formly payload where only `Break Duration` has been restored

Local verification passed:
- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.ts`
- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts` (16 SUCCESS)
- `git diff --check`

cocojojo5213:
i will tyr it later

cocojojo5213:
Thanks for testing `baec64c` again. I found why this still showed up in the UI and pushed `f045816d9`.

The previous commits restored the component model, but Formly could keep the actual `breakRules` form controls partially blank after the mode switch. That meant the signal/model had the restored Min/Max values, while the visible inputs could still show empty Min work duration / Max work duration fields.

This follow-up patches the restored rule values back into the Formly form controls when the restored rules differ from the blank/partial payload. It only runs on that restore path, so normal visible rule edits still flow through unchanged.

I added a regression that checks the actual `breakRules` form control value, not only the component model.

Local verification passed:
- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.ts`
- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts` (17 SUCCESS)
- `git diff --check`

cocojojo5213:
Thanks for testing `f045816` and for confirming the exact path still fails. I found another case that matches this symptom and pushed `56fc332c0`.

The remaining issue was that while Ratio-based is visible, Formly can still emit a hidden `breakRules` payload where `minDuration` / `maxDuration` are blank but `breakDuration` is present. The previous code treated that as a non-empty rule row and cached it as the last known rule state. After that, switching back to Rule-based could only restore Break Duration because the cached Min/Max values had already been poisoned by the hidden partial payload.

The new follow-up only refreshes the cached rule rows when Rule-based is actually visible and the required rule fields are cacheable. I also added a regression for the exact hidden partial payload path before switching back to Rule-based.

Local verification passed:
- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.ts`
- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts` (18 SUCCESS)
- `git diff --check`

johannesjo:
Thanks for tackling this — the underlying bug is real and the `resetOnHide: false` direction is right. I found one regression I'd ask you to fix before merging, plus some smaller notes.

## 🔴 `save()` still discards user edits to break rules

`dialog-flowtime-settings.component.ts:260-263`:

```ts
breakRules:
  currentModel.isBreakEnabled && breakMode === 'rule' && currentModel.breakRules
    ? this._toBreakRulesInMs(currentModel.breakRules)
    : (this._initialFlowtimeConfig.breakRules ?? []),
```

When `isBreakEnabled` is `false` (or `breakMode !== 'rule'`), save falls back to the **initial** config rather than to the cache the rest of this PR carefully maintains. Edits the user made earlier in the same dialog session are silently overwritten with the pre-edit values.

Walk-through:
1. Open dialog. Initial rules = `[A]`.
2. While in rule mode, edit rules to `[B]`. `_rememberModelState` updates `_lastNonEmptyBreakRules` to `[B]`. `model().breakRules` = `[B]`.
3. Toggle Flowtime breaks off. `resetOnHide: false` keeps `[B]` in `model().breakRules`.
4. Click Save. `isBreakEnabled` is `false` → short-circuits to `_initialFlowtimeConfig.breakRules` = `[A]`.
5. Persisted rules = `[A]`. User's `[B]` edits are lost.

I confirmed this with a regression test against the PR HEAD:

```ts
it('should persist rule edits made before disabling breaks', () => {
  component.updateModel({
    ...component.model(),
    breakRules: [{ minDuration: 0, maxDuration: 30, breakDuration: 10 }],
  });
  component.updateModel({ ...component.model(), isBreakEnabled: false });
  component.save();
  const saved = globalConfigServiceMock.updateSection.calls.mostRecent().args[1];
  expect(saved.breakRules).toEqual([
    { minDuration: 0, maxDuration: 30 * 60000, breakDuration: 10 * 60000 },
  ]);
});
```

Fails with `Expected $[0].maxDuration = 1500000 to equal 1800000` — save wrote the initial 25/5 row instead of the edited 30/10 row. The closest existing test (`'should preserve hidden break settings when Flowtime breaks are disabled'`, spec line 373) bypasses `updateModel` with `model.set({ isBreakEnabled: false })`, so it only exercises the fresh-dialog fallback and misses the edit-then-hide path.

`_lastNonEmptyBreakRules` is already the right source of truth — it's the most recent valid rule rows the user typed. Using it as the fallback fixes the bug:

```ts
breakRules:
  currentModel.isBreakEnabled && breakMode === 'rule' && currentModel.breakRules
    ? this._toBreakRulesInMs(currentModel.breakRules)
    : this._toBreakRulesInMs(this._lastNonEmptyBreakRules),
```

`breakMode` and `breakPercentage` aren't affected — they read straight from `currentModel.*` with only a null-safe fallback.

## 🟡 Restoration layer is heavier than the problem

Roughly 230 lines of cache-and-restore over eight private helpers, most of it defending against Formly's `modelChange` emitting blank strings or partial rows during visibility transitions (the spec names spell this out: *"when Formly emits blank strings"*, *"delayed partially blank rule payload"*, *"when Formly mutates the model object"*).

A couple of consequences worth weighing:

- **Tests are coupled to emission shape.** Six of the new specs feed hand-crafted blank-string payloads into `updateModel`. If Formly changes its emission timing, these can pass while the real dialog regresses, or vice versa.
- **`_isRestoringBreakRules` persists into the next call.** After a transition restore, the next `updateModel` is still in restore-mode and re-runs `_hasRestorableBlankBreakRuleFields`. If a user intentionally clears a row immediately after a mode toggle, the clear can be reverted from cache.

Not a blocker, but a lighter alternative worth considering: keep a `WritableSignal<FlowtimeConfig>` as canonical state and only `patchValue` *visible* slices into the form on visibility changes — Formly's emissions for hidden fields stop mattering, and most of the restoration helpers go away.

## 🟡 Minor

- `fields` is wrapped in `computed()` (line 88) but doesn't read any signal — it's effectively a cached getter. Either drop `computed()` or make it actually reactive on `model()`.
- `_hasRestorableBlankBreakRuleFields` iterates `fallbackRules`, not `breakRules`. If the user added rows past the cache length, blank trailing rows aren't detected. Probably fine, but worth a one-liner comment if intentional.

## Suggestion

Block on the `save()` fallback fix plus a regression test for the edit-then-hide path. The simplification thoughts can land separately (or not at all).

cocojojo5213:
Thanks for the detailed review. I pushed `8f7411b1d` to address the blocker and the follow-up delete regression.

Changes:
- `save()` now falls back to `_lastNonEmptyBreakRules` for hidden/non-rule states, so rule edits made earlier in the same dialog session are persisted instead of reverting to the dialog-open snapshot
- added the edit-then-disable regression test from the review
- allowed user-driven rule deletion immediately after restoring the Rule-based section, so the restore window no longer re-adds cached rows after the user deletes them
- added a regression for the Rule-based -> Ratio-based -> Rule-based -> delete path

Local verification passed:
- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.ts`
- `npm run checkFile src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts` (20 SUCCESS)
- `git diff --check`

johannesjo:
Thanks for pushing `8f7411b1d` — the `save()` fallback now reads from `_lastNonEmptyBreakRules` and the edit-then-disable regression covers it, so the 🔴 blocker is resolved. 👍

I went a step further and tried to verify this the way it actually broke, and the results updated my own opinion — two findings:

**1. The complexity here is mostly justified — I was wrong to call it "heavier than the problem."** I built the "simpler" canonical-signal alternative I floated earlier (keep the signal as source of truth, only accept emitted values for *visible* fields) and tested it against a **faithful** test that toggles the real `breakMode` control and lets Formly emit — and it **loses data**. Instrumenting the real emissions shows why: when you switch to ratio, Formly emits the hidden rule row *progressively stripped* — `[{maxDuration, breakDuration}]`, then `[{breakDuration}]` — and it mutates the row objects **in place**, so even holding a reference to the previous value doesn't protect it. That's exactly what your `_copyBreakRules` + blank-detection defends against, and `resetOnHide: false` alone is not enough. So this is matching real Formly behavior, not over-engineering.

**2. The fix appears to actually work at HEAD — but the existing tests prove it the wrong way.** Every current spec feeds hand-crafted `updateModel(...)` payloads, which is why green tests kept coexisting with live repros. So I wrote tests that drive the **real** `breakMode` select instead and ran them against your current code: the single-rule, edited-rule, and **two-rule** round-trips all pass. The two-rule case is the one @beerkumquatpome had the worst result on, so this is genuinely reassuring.

So my suggestion flips from "simplify it" to **"lock it in at the layer where it broke"**:

- **Add at least one faithful test** that drives the real control rather than `updateModel`. Here's the shape that passed against your code — feel free to lift it:

```ts
const switchMode = (mode: 'ratio' | 'rule'): void => {
  (component.form.get('breakMode') as any).setValue(mode);
  fixture.detectChanges();
};

it('preserves rules across a real Rule -> Ratio -> Rule switch', () => {
  const expected = component.model().breakRules;
  switchMode('ratio');
  switchMode('rule');
  expect(component.model().breakRules).toEqual(expected);
  expect((component.form.get('breakRules') as any)?.at(0).value).toEqual(expected![0]);
});
```

- **@beerkumquatpome** 🙏 could you confirm the Rule→Ratio→Rule path one more time on `8f7411b1d`? Nobody has re-confirmed in the running app after the last couple of commits, and a form-control test still isn't a real click. If it's good for you now, I think this is mergeable.

- One thing still worth a quick check (my earlier `_isRestoringBreakRules`-persists note): right after a mode switch, can the user *clear* a rule field to retype it, or does it snap back from cache? The deletion guard handles removing a whole row, but I'm not certain it covers clearing a single field. A faithful test (clear a field right after the round-trip) would settle it.

Net: I'm now in favor of this approach rather than a rewrite. A real-control regression test plus a tester re-confirm and I think it's good to go. Really appreciate how much you stuck with this one.

johannesjo:
First, thank you for sticking with this one — eight commits and a lot of back-and-forth with @beerkumquatpome's testing. That persistence is appreciated, and the diagnosis underneath it all (Formly resetting/stripping hidden fields) was correct.

Two things changed since this PR opened, and together they pointed to a much smaller fix:

**1. Master moved under you.** The focus-screen overhaul (#7586) rewrote `dialog-flowtime-settings.component.ts` after this PR's base. It now *disables* the break fields when "Flowtime breaks" is off instead of *hiding* them — so the original "settings wiped when breaks are disabled" bug is already fixed on master, differently. That left only the ratio↔rule switch.

**2. That remaining bug collapses to ~6 lines.** The reason this branch needed the ~230-line cache/restore layer is subtle: `resetOnHide: false` was on the top-level fields, but not on the `repeat`'s `fieldArray` or its inner inputs. So Formly kept the row but stripped its *values* (`[{}]`/`[null]`) — exactly the "blank/partial payload" the restore layer was rebuilding from. Pushing `resetOnHide: false` all the way down (field → fieldArray → each input, plus `breakPercentage`) stops the strip at the source, and the restore layer isn't needed at all. As a bonus it removes the "can't clear a field after a switch" regression you and @beerkumquatpome were chasing, because there's no cache to fight the user.

I verified this end-to-end — including a real-DOM e2e that drives the actual `mat-select` Rule→Ratio→Rule and a real backspace-clear, confirmed to fail without the fix and pass with it. I've opened #8489 with that approach.

Given how much of this PR is now either already-on-master or replaced by the smaller fix, I'm going to close this in favour of #8489 — but the contribution genuinely surfaced the real Formly behaviour and the exact repro path that #8489's tests now lock in, so thank you for that. 🙏

@beerkumquatpome — thank you too for the relentless, precise testing; #8489 should finally put the mode-switch data loss to rest.

PR #8173, opened 2026-06-08
fix(calendar): disable iCal polling on Android

### LINKED ISSUE #7406: 🚨 Google Calendar iCal integration does not work on Android
### Version Used

18.3.0 (both Windows desktop app and Android app from Google Play)

### Expected Behavior

Google Calendar iCal sync works on Android just as it works on Windows, or if it is not to be supported - info should be provided in the config. For now there is info in the Edit Calendar Config menu that states the functionality is not expected to work on browser version of the app:

If this is the case for the Android app too it should be noted. 

### Current Behavior

The issue persists on Android. No errors on Windows desktop app.

Sync attempts end up with displaying popup in the bottom of the screen, with message starting with `Calendar Provider Error: Http failure response`.

Issue Provider Panel shows error message:

```
Http failure response for <google-calendar-ical-url>: 0 Unknown Error
```

When trying to duplicate Calendar Config and test it via Test Connection button the Connection failed message appears.

---

I want to add that in my Edit Calendar Config menu I have enabled the option "Disable when using web application" and the functionality is not turned off on Android.

Moreover, given I sync the app b/w Windows and Android devices I cannot reliably turn the iCal calendar sync when using it on Windows and turn it off for Android and expect the settings to persist - "Enable" setting is synced between devices, so if I turn it off on Android and sync the app, iCal Google Calendar sync will be turned off on Windows. If I turn it on on Windows and sync, it will be turned on back on Android. 

### Steps to Reproduce

1. Install application v18.3.0 on Android device
2. Add Google Calendar via iCal URL in Calendar Config
3. Enable the calendar sync
4. Attempt sync operation - observe "Calendar Provider Error: Http failure response (0 Unknown Error)"
5. On Windows desktop app (same version, synced account): verify sync works without error
6. In Android app, navigate to Edit Calendar Config - toggle "Disable when using web application" ON
7. Return to main view - verify iCal sync still attempts
8. Sync Android app with Windows desktop - verify "Enable" setting propagates to Windows, re-enabling sync there despite being disabled on Android

### Console Output

```shell

```

### documentation gap

cocojojo5213:
Fixes #7406

## Summary
- extend the existing iCal "disable in web app" setting so it also disables provider polling/requesting on Android WebView
- keep manual connection testing behavior unchanged while suppressing automatic event requests for disabled providers
- update the English setting copy and wiki docs to mention Android/WebView behavior

## Verification
- `npm run checkFile src/app/features/calendar-integration/calendar-integration.service.ts`
- `npm run checkFile src/app/features/calendar-integration/calendar-integration.service.spec.ts`
- `npm run checkFile src/app/features/issue/providers/calendar/calendar-common-interfaces.service.ts`
- `npm run checkFile src/app/features/issue/providers/calendar/calendar-common-interfaces.service.spec.ts`
- `npm run checkFile src/app/features/issue/providers/calendar/is-calendar-provider-disabled-on-current-platform.util.ts`
- `npm run checkFile src/app/features/issue/providers/calendar/is-calendar-provider-disabled-on-current-platform.util.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/issue/providers/calendar/is-calendar-provider-disabled-on-current-platform.util.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/calendar-integration/calendar-integration.service.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/issue/providers/calendar/calendar-common-interfaces.service.spec.ts`
- `npm run int:test`
- `git diff --check`
- `git diff HEAD~1..HEAD --check`

### DIFF
--- a/docs/wiki/3.05-Web-App-vs-Desktop.md
+++ b/docs/wiki/3.05-Web-App-vs-Desktop.md
@@ -10,7 +10,7 @@ Plugins that require Node.js execution (elevated permissions) are disabled in th
 
 ### Calendar Integration
 
-Calendar providers can be disabled for the web app via an `isDisabledForWebApp` flag to avoid CORS failures. When this flag is set and the app runs in a browser, calendar event requests return empty results instead of performing the HTTP request. The UI shows a prominent warning that calendar integration will likely not work in the browser and recommends downloading the desktop version.
+Calendar providers can be disabled for the web app or Android app via an `isDisabledForWebApp` flag to avoid CORS and WebView failures. When this flag is set and the app runs in a browser or Android WebView, calendar event requests return empty results instead of performing the HTTP request. The UI shows a prominent warning that calendar integration will likely not work in the browser or Android app and recommends downloading the desktop version.
 
 ### Local File System
 
--- a/docs/wiki/3.07-Issue-Integration-Comparison.md
+++ b/docs/wiki/3.07-Issue-Integration-Comparison.md
@@ -104,6 +104,7 @@ Every issue provider supports the same basic operations: checking that the integ
 - **Due dates:** Event start time (all-day or timed).
 - **Descriptions:** Event description as task notes.
 - **URL format (example Nextcloud):** Use the **“Copy subscription link”** so the URL is in the correct form, e.g. `https://<NC-URL>/remote.php/dav/public-calendars/<CALENDAR-ID>?export`. Paste this in the app under **“URL of the iCal source.”** Do **not** use the “Copy private link”—that format is for CalDAV, not iCal.
+- **Platform note:** The **Disable in web or Android app** setting suppresses iCal requests in the browser and on Android, where WebView network restrictions can produce similar failures.
 
 ### 8. Redmine
 
--- a/src/app/features/calendar-integration/calendar-integration.service.spec.ts
+++ b/src/app/features/calendar-integration/calendar-integration.service.spec.ts
@@ -944,7 +944,7 @@ END:VCALENDAR`;
 
   describe('testConnection', () => {
     it('should return true when connection succeeds', async () => {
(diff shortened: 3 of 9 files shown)


johannesjo:
Really nice fix. The root cause is subtle — `IS_WEB_BROWSER` is defined as `!IS_ELECTRON && !IS_ANDROID_WEB_VIEW` (`src/app/app.constants.ts:5`), so the old `isDisabledForWebApp && IS_WEB_BROWSER` guard never fired inside the Android WebView, and the disable toggle silently did nothing there. Centralizing the check in `isCalendarProviderDisabledOnCurrentPlatform` (`src/app/features/issue/providers/calendar/is-calendar-provider-disabled-on-current-platform.util.ts`) using the real `IS_WEB_BROWSER`/`IS_ANDROID_NATIVE` constants rather than a UA sniff is exactly the right call, and Electron stays unaffected since both flags are false there. The injectable-params design makes the util trivially testable, and tightening the previously-noop `requestEvents$` spec into a deterministic `expect(result).toEqual([])` + `httpMock.expectNone(...)` is a real improvement. Docs + `en.json`-only copy update are a good touch too.

One small, non-blocking observation: the modified `CalendarCommonInterfacesService.isEnabled()` gates issue-provider search, but the schedule poll loop (`pollChanges$` in `src/app/features/calendar-integration/store/calendar-integration.effects.ts:81`) still filters on the raw `calProvider.isEnabled` flag rather than that method. It's harmless because the actual fetch is blocked at `requestEvents$` (`calendar-integration.service.ts:419`), so no request goes out on Android either way — just flagging that there are now two slightly different "is this enabled" paths in case it's ever worth unifying.

Looks good to merge. Thanks!

johannesjo:
Thank you!

PR #8174, opened 2026-06-08
fix(notes): keep long mobile note panel dismissible

### LINKED ISSUE #5849: 🚨 android Long Project Note Blocks: User Interface & Interaction of Subsisting App Content
### Version Used

v16.7.3_L2A apk (f-droid)

### Expected Behavior

New Project note when popping up should use either: timeout (although maybe clunky ui), or allow to be scrolled past.

Best be the latter, so that instead of toasting the new note and its contents---being in what I'd assume is a container that doesn't allow for scroll---one may still see the note in it entirety, granted they'd still have to scroll past the obstruction.

tldr: move message toast into a container which allows for scrolling

### Current Behavior

When user interaction occurs, the main content windows are subceded, by the superimposed project note---when a new note is created---hampering any,  aforementioned, interaction; excluding the left side Hamburg menu (and it's preceding items) and the below toolkit. Some menus that are superimposed are still accessible allowing for shortening of note length allowing for escape from stress. 

tldr: You can [not] do anything cus it block the area to do things.

### Steps to Reproduce

Assuming a new empty project/or one with stuff:
1. Make a new note:
2. click bottom right most button
3. Click show project Notes
4. click add a new note
5. then mash bunch of text into the field, (like *The Raven* in length or something, because that should be way more then enough)
6. Save
7. Restart app:
8. Force close app/power down device/clear cache, etc.
9. reopen to bug, in main area.
10. Smile and say "I'm a happy programmer."
11. Compliment someone
12. meditate for ~15 mins
13. squash bug

### Console Output

```shell

```

cocojojo5213:
Fixes #5849

## Summary
- start the notes bottom sheet at the same compact height as task panels instead of the expanded generic panel height
- show the bottom-sheet close button for notes panels only, with translated aria text
- keep the close button clickable inside the draggable header without changing the task panel header/autofocus behavior

## Validation
- `npm run checkFile src/app/features/bottom-panel/bottom-panel-container.component.ts`
- `npm run checkFile src/app/features/bottom-panel/bottom-panel-container.component.scss`
- `npm run checkFile src/app/features/bottom-panel/bottom-panel-container.component.spec.ts`
- `npx prettier --check src/app/features/bottom-panel/bottom-panel-container.component.html`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/bottom-panel/bottom-panel-container.component.spec.ts`
- `npm run int:test`
- `git diff --check`

## Conflict/duplication checks
- issue #5849 is open and unassigned
- no exact open PR for #5849 found
- checked nearby note/task UX PRs #8044, #8100, and #8156; this patch only touches bottom-panel container files and `merge-tree` showed no conflict markers

### DIFF
--- a/src/app/features/bottom-panel/bottom-panel-container.component.html
+++ b/src/app/features/bottom-panel/bottom-panel-container.component.html
@@ -10,8 +10,8 @@
         mat-icon-button
         class="close-btn"
         [attr.aria-label]="T.G.CLOSE | translate"
-        (pointerdown)="$event.stopPropagation()"
-        (click)="close(); $event.stopPropagation()"
+        (pointerdown)="stopHeaderEventPropagation($event)"
+        (click)="close(); stopHeaderEventPropagation($event)"
       >
         <mat-icon>close</mat-icon>
       </button>
--- a/src/app/features/bottom-panel/bottom-panel-container.component.html
+++ b/src/app/features/bottom-panel/bottom-panel-container.component.html
@@ -1,16 +1,21 @@
 <div class="bottom-panel-container">
   <div
     class="bottom-panel-header"
+    [class.has-close-btn]="isShowCloseButton()"
     #panelHeader
   >
     <div class="drag-handle"></div>
-    <button
-      mat-icon-button
-      class="close-btn"
-      (click)="close()"
-    >
-      <mat-icon>close</mat-icon>
-    </button>
+    @if (isShowCloseButton()) {
+      <button
+        mat-icon-button
+        class="close-btn"
+        [attr.aria-label]="T.G.CLOSE | translate"
+        (pointerdown)="$event.stopPropagation()"
+        (click)="close(); $event.stopPropagation()"
+      >
+        <mat-icon>close</mat-icon>
+      </button>
+    }
   </div>
 
   <div class="bottom-panel-content">
--- a/src/app/features/bottom-panel/bottom-panel-container.component.scss
+++ b/src/app/features/bottom-panel/bottom-panel-container.component.scss
@@ -38,7 +38,20 @@
   }
 
   .close-btn {
-    display: none; // Hidden for now
+    display: none; // Hidden by default to keep task panel autofocus behavior unchanged.
+  }
+
+  &.has-close-btn {
+    justify-content: center;
+
+    .close-btn {
+      display: inline-flex;
+      position: absolute;
+      right: 4px;
+      top: 50%;
+      transform: translateY(-50%);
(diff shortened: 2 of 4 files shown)


johannesjo:
Thanks — this makes the long-notes case on mobile much friendlier. Opening the NOTES sheet at the compact 0.6 ratio via the new `getInitialBottomPanelHeightRatio` helper (`src/app/features/bottom-panel/bottom-panel-container.component.ts:66`) keeps the dismiss affordance reachable, and extracting that into a pure function reads nicely. The close-button fix is the clever part: since `.bottom-panel-header > * { pointer-events: none }` makes the whole header draggable (`bottom-panel-container.component.scss:24`), the button needs the explicit `pointer-events: auto` plus the `pointerdown`/`click` `stopPropagation()` so the tap actually lands instead of being eaten by the drag handler. Gating visibility to `'NOTES'` keeps the existing task-panel autofocus behavior untouched, and the translated aria-label is a good detail.

Two tiny, non-blocking notes:
- The new spec covers `getInitialBottomPanelHeightRatio` but not the close-button visibility / `stopPropagation` behavior — might be worth a quick test there since that's the actual UX-critical part of the fix.
- `NOTES_PANEL_HEIGHT` (0.6) duplicates `TASK_PANEL_HEIGHT` (0.6) (`:38`/`:39`); totally harmless, just noting it in case they're meant to track together.

Unrelated heads-up on CI: the red `Tests` run is the known flaky `TaskService › focusTaskById` spec (`src/app/features/tasks/task.service.spec.ts:910`/`919`), unrelated to this PR — it only touches bottom-panel files. A re-run should clear it.

Looks good to merge. Thanks!

cocojojo5213:
Thanks for the close-button test suggestion. I pushed follow-up commit `e47de637c` to cover that UX-critical path explicitly.

What changed:
- added a small helper/spec for showing the close button only for the notes panel
- routed the close-button `pointerdown` / `click` propagation stop through a helper and covered it in the spec, so the button tap cannot start the header drag path

Local verification passed:
- `npm run checkFile src/app/features/bottom-panel/bottom-panel-container.component.ts`
- `npm run checkFile src/app/features/bottom-panel/bottom-panel-container.component.spec.ts`
- `npx prettier --check src/app/features/bottom-panel/bottom-panel-container.component.html`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/bottom-panel/bottom-panel-container.component.spec.ts` (5 specs)
- `git diff --check`

johannesjo:
Thank you!

PR #8175, opened 2026-06-08
feat(worklog): filter history by task status

### LINKED ISSUE #6542: 💡 one more option needed in worklog
### Problem Statement

Image

could u add a option of "completed/uncompleted"

### Possible Solution

.

_No response_

### Additional context

_No response_

cocojojo5213:
## Summary
- add an All / Completed / Uncompleted filter to the History / Worklog view
- recalculate visible day, week, month, year, and total time/counts for the selected status
- keep Worklog export and the Daily Summary week widget unchanged

Fixes #6542

## Verification
- npm run checkFile src/app/features/worklog/util/filter-worklog-entries-by-status.util.ts
- npm run checkFile src/app/features/worklog/util/filter-worklog-entries-by-status.util.spec.ts
- npm run checkFile src/app/features/history/history.component.ts
- npm run checkFile src/app/features/history/history.component.scss
- npm run checkFile src/app/features/history/history.component.spec.ts
- npm run checkFile src/app/t.const.ts
- npx prettier --check src/app/features/history/history.component.html src/assets/i18n/en.json
- CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/worklog/util/filter-worklog-entries-by-status.util.spec.ts
- CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/history/history.component.spec.ts
- npm run int:test
- git diff --check


johannesjo:
Thanks for this, @cocojojo5213 — an All / Completed / Uncompleted filter for the History view is a genuinely useful addition (and a nice answer to #6542). I went through it carefully and a lot is done well:

- **Clean separation.** The filtering logic lives in a dedicated, well-tested util (`filter-worklog-entries-by-status.util.ts`) rather than being tangled into the component. Good call.
- **Translations done right.** All four new strings go through `T` / `en.json` (`STATUS_FILTER`, `STATUS_FILTER_ALL/DONE/UNDONE`), and only `en.json` is touched. 👍
- **Zero-cost default path.** For the `'ALL'` filter, `filterWorklogByTaskStatus` returns the input object by reference (and `filterWorklogEntriesByStatus` returns the same array), so the common case adds no work. The `toBe(data)` assertion locks that in.
- **No state mutation.** The recomputation builds fresh objects via spreads and operates on the already-derived `worklogData$` signal — no NgRx state is touched, and this isn't the per-task hot path.
- **CI is green** across the board (Lint, Tests, CodeQL, etc.).

There is one correctness issue I'd like to see addressed before merge:

**Filtered time totals can be inflated for parent tasks that have sub-tasks.**

`getTimeSpentForWorklogEntries` in `src/app/features/worklog/util/filter-worklog-entries-by-status.util.ts` (the `reduce` around lines 26–34) decides whether to count an entry's `timeSpent` based on whether it has a *visible* sub-task in the current (filtered) list:

```ts
const hasVisibleSubTask = entries.some(
  (otherEntry) =>
    otherEntry.parentId === entry.task.id ||
    entry.task.subTaskIds?.includes(otherEntry.task.id),
);
return hasVisibleSubTask ? total : total + entry.timeSpent;
```

The canonical aggregation rule everywhere else in the worklog code is unconditional: a parent's own time is **never** summed — only leaf tasks contribute. See `map-archive-to-worklog.ts:88` (`if (task.subTaskIds.length === 0)`) and the same guard in `worklog-export.util.ts:70`.

The divergence bites when a parent task (`subTaskIds.length > 0`) has no visible child after filtering — e.g. a completed parent whose sub-tasks are all uncompleted, viewed under the **Completed** filter. The child rows are filtered out, so `hasVisibleSubTask` becomes `false`, and the parent's own `timeSpent` (which mirrors the sum of its children's time) gets added in. The result is a day/week/month/year/total that is inflated relative to the unfiltered view — effectively double-counting the parent's aggregate against its (now hidden) children.

A note on the tests: the existing spec case `getTimeSpentForWorklogEntries` → *"keeps parent summary time when no child row is visible"* asserts `3000`, but its fixture `createEntry('parent', true, 3000)` is created with an **empty** `subTaskIds`, so it's structurally a leaf — `3000` happens to be correct there, and the test name is misleading. A case that actually exercises the divergence (a parent with non-empty `subTaskIds` whose only child is filtered out) would currently return the inflated number.

**Suggested fix:** make the recalculation follow the same leaf-only rule the rest of the worklog uses — count an entry only when `entry.task.subTaskIds.length === 0` — instead of the "visible sub-task" heuristic. That keeps filtered totals consistent with the unfiltered view and the export. Then please add/adjust a spec with a real parent-with-sub-task to pin the corrected behavior.

Everything else looks good to me — happy to take another pass once the aggregation is aligned. Thanks again for the contribution!

cocojojo5213:
Thanks for the detailed catch. I pushed follow-up commit `7a0ed21` to align the filtered totals with the existing worklog/export aggregation rule: parent summary rows are now skipped based on the task structure itself, so only leaf tasks contribute to recalculated totals even when filtering hides the child rows.\n\nI also updated the specs to cover the real parent-with-subtask case where a completed parent remains visible but its unfinished child is filtered out.\n\nLocal verification passed:\n- `npm run checkFile src/app/features/worklog/util/filter-worklog-entries-by-status.util.ts`\n- `npm run checkFile src/app/features/worklog/util/filter-worklog-entries-by-status.util.spec.ts`\n- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/worklog/util/filter-worklog-entries-by-status.util.spec.ts` (9 specs)\n- `git diff --check`

johannesjo:
Thanks a lot for taking the time to build this out — the implementation is careful (the parent-vs-subtask time handling is a nice touch) and well tested.

I've decided not to merge it, though, and want to be upfront about why:

- The original request (#6542) is about the **Worklog Export** dialog — the screenshot is the export's column-options menu. This PR instead adds a filter to the main History view and explicitly leaves the export untouched, so it doesn't really address what was asked.
- A done/undone filter on the worklog has limited value in practice: the worklog is built from the archive, where almost everything is already completed — so "Completed" ends up nearly identical to "All", and "Uncompleted" is usually close to empty.
- Filtering parent and subtask rows independently leaves the day's task tree inconsistent (a completed parent can stay with all its children filtered out, showing a 0-time row; or an orphan subtask shows with no parent).

This is a scope/direction call, not a knock on the code. Sorry for the effort that won't land here — I really appreciate the contribution and hope you'll send more.

cocojojo5213:
Thanks for the clear explanation. That scope call makes sense; I mapped #6542 too broadly onto the History view, while the useful place is the Worklog Export dialog/column options.

I will leave this direction closed and won't try to push it further here. If I revisit the issue, I'll keep the change focused on the export flow and avoid changing the main History tree/filter behavior.

PR #8177, opened 2026-06-08
fix(jira): honor NO_PROXY for Electron requests

### LINKED ISSUE #6324: 🚨 jira issue provider does not honor proxy env variables

cocojojo5213:
## Summary
- pass the Jira request URL into the Electron proxy-agent helper
- bypass HTTPS_PROXY/HTTP_PROXY when NO_PROXY/no_proxy matches the Jira host
- keep the existing self-signed certificate override when bypassing the proxy

Fixes #6324

## Verification
- npm run checkFile electron/proxy-agent.ts
- npm run checkFile electron/jira.ts
- node --test electron/proxy-agent.test.cjs
- npm run test:electron
- npm run build:packages
- npm run electron:build
- npm run int:test
- git diff --check
- npx prettier --check electron/proxy-agent.test.cjs

### DIFF
--- a/electron/jira.ts
+++ b/electron/jira.ts
@@ -18,7 +18,7 @@ export const sendJiraRequest = ({
   jiraCfg: JiraCfg;
 }): void => {
   const mainWin = getWin();
-  const agent = createProxyAwareAgent(jiraCfg?.isAllowSelfSignedCertificate);
+  const agent = createProxyAwareAgent(url, jiraCfg?.isAllowSelfSignedCertificate);
 
   fetch(url, {
     ...requestInit,
--- a/electron/proxy-agent.test.cjs
+++ b/electron/proxy-agent.test.cjs
@@ -1,0 +1,168 @@
+const test = require('node:test');
+const assert = require('node:assert/strict');
+const path = require('node:path');
+const Module = require('node:module');
+const { Agent: HttpsAgent } = require('node:https');
+const { HttpsProxyAgent } = require('https-proxy-agent');
+
+require('ts-node/register/transpile-only');
+
+const originalModuleLoad = Module._load;
+const originalEnv = { ...process.env };
+const proxyAgentModulePath = path.resolve(__dirname, 'proxy-agent.ts');
+
+const resetModule = () => {
+  delete require.cache[proxyAgentModulePath];
+};
+
+const clearProxyEnv = () => {
+  delete process.env.HTTPS_PROXY;
+  delete process.env.https_proxy;
+  delete process.env.HTTP_PROXY;
+  delete process.env.http_proxy;
+  delete process.env.NO_PROXY;
+  delete process.env.no_proxy;
+};
+
+const installMocks = () => {
+  Module._load = function patchedLoad(request, parent, isMain) {
+    if (request === 'electron-log/main') {
+      return {
+        log: () => {},
+      };
+    }
+
+    return originalModuleLoad.call(this, request, parent, isMain);
+  };
+};
+
+const loadProxyAgentModule = () => {
+  resetModule();
+  return require(proxyAgentModulePath);
+};
+
+test.beforeEach(() => {
+  clearProxyEnv();
+  installMocks();
+  resetModule();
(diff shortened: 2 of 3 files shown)


johannesjo:
Thanks for this, @cocojojo5213 — really nicely done. Honoring `NO_PROXY` for the Electron Jira requests is a genuine gap (closes #6324), and I appreciate that you covered the casing variants, leading-dot/bare-domain/wildcard forms, optional ports, and added a focused `proxy-agent.test.cjs` on top of it. The self-signed-cert handling being preserved on the bypass path is exactly the right call.

I went through the logic carefully and it holds up well:

- The matching in `isNoProxyMatch` (`electron/proxy-agent.ts`) correctly handles exact host, bare-domain suffix (with a proper `.${host}` dot boundary so `ample.com` doesn't match `example.com`), leading-dot, `*`, `*.suffix`, and optional-port matching.
- The bypass falls through to the self-signed `https.Agent` branch, so `isAllowSelfSignedCertificate` is still respected when the proxy is skipped.
- The only caller, `electron/jira.ts:21`, is updated to pass the request `url` through. Since `electron/` is main-process-only code, no `IS_ELECTRON` guard is needed here — good.

## On the red `Tests` check

Just so you're not chasing a ghost: the failing `Tests` run is **not** caused by your change. The two failures are:

- `TaskService › focusTaskById › focuses the last copy when it is focusable (in-panel preference, #7120)` (`src/app/features/tasks/task.service.spec.ts:910`)
- `TaskService › focusTaskById › falls back to an earlier copy when the last one cannot take focus` (`src/app/features/tasks/task.service.spec.ts:919`)

These are a known flaky `document.activeElement` spec-leak in the task tests, unrelated to anything in `electron/`. I've just landed fixes for that leak on `master`, so a rebase should clear the check. No action needed on your end beyond that.

## One small follow-up (non-blocking)

`NO_PROXY` entries written as CIDR ranges aren't matched. For example `NO_PROXY=10.0.0.0/8` won't bypass a request to `10.0.0.5` — `parseNoProxyEntry` strips everything after the first `/` as a path, so the entry reduces to the literal host `10.0.0.0`, which then can't match. Exact bare-IP entries (e.g. `NO_PROXY=10.0.0.5`) do work fine via the equality branch; it's specifically the CIDR form that's unsupported.

This is a common limitation and not a merge blocker — plenty of `NO_PROXY` implementations skip CIDR too. Feel free to leave it as-is and we can track it separately, or add it here if you'd like; entirely your call.

Aside from that, this looks good to me. Once it's rebased onto current `master` to pick up the flaky-test fix, I'm happy to merge. Thanks again for the contribution!

cocojojo5213:
Thanks for confirming the CI failure was the shared focus spec leak. I rebased this branch onto current `master` in `c469714` to pick up the upstream flaky-test fix; there were no conflicts and no code changes beyond the rebase.\n\nLocal verification after the rebase passed:\n- `npm run checkFile electron/proxy-agent.ts`\n- `npm run checkFile electron/jira.ts`\n- `node --test electron/proxy-agent.test.cjs` (10 tests)\n- `git diff --check upstream/master..HEAD`\n\nI left CIDR `NO_PROXY` support out of this PR as a follow-up scope, per your note.

johannesjo:
Thank you!

PR #8179, opened 2026-06-08
fix(header): show focus mode entry on mobile

### LINKED ISSUE #8157: 🚨 "Enter Focus Mode" button not visible in portrait mode on Android (only appears in landscape)
### Version Used

v18.9.1

### Current Behavior

In the Android version of Super Productivity, the "Enter Focus Mode" action is not accessible in portrait mode. I searched the interface for about 10 minutes but could not find any entry point or button to trigger focus mode.

When I rotate the device to landscape mode, the "Enter Focus Mode" button becomes visible in the top bar.

### Expected Behavior

The "Enter Focus Mode" button should be consistently accessible in both portrait and landscape orientations, ideally without requiring rotation or deep navigation to find it.

### Steps to Reproduce

1. Open Super Productivity on an Android device.
2. Keep the device in portrait orientation.
3. Look for the "Enter Focus Mode" button/entry in the top bar or main interface.
4. Rotate the device to landscape orientation.
5. Observe that the button now appears in the top bar.

### Console Output

```shell

```

### Documentation

- [x] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

cocojojo5213:
## Summary
- keep the header focus button visible on mobile when Focus Mode is enabled
- add a regression test for the narrow/mobile header visibility case
- mention the header focus entry point in the Focus Mode wiki page

Fixes #8157

## Test plan
- `npm run checkFile src/app/core-ui/main-header/main-header.component.ts`
- `npm run checkFile src/app/core-ui/main-header/main-header.component.spec.ts`
- `npx prettier --check src/app/core-ui/main-header/main-header.component.html docs/wiki/4.15-Timers-and-Focus-Mode.md`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/core-ui/main-header/main-header.component.spec.ts`
- `npm run int:test`
- `git diff --check`

### DIFF
--- a/docs/wiki/4.15-Timers-and-Focus-Mode.md
+++ b/docs/wiki/4.15-Timers-and-Focus-Mode.md
@@ -2,6 +2,8 @@
 
 Focus Mode in Super Productivity addresses a central challenge of task-based work: **maintaining sustained attention and structured work sessions** in the face of distractions and context switching. The app is designed to help you stay organized and focused through timeboxing, time tracking, break reminders, anti-procrastination features, and a Pomodoro-style timer—all working together to support healthy, productive habits. Timers and focus are treated as a **first-class feature**: they have dedicated configuration, distinct modes with clear behavior, and deep integration with task tracking and metrics. Understanding what problem Focus Mode solves, how it relates to time-boxing and deep work, what the app assumes about focus when using timers, and how timers influence your behavior helps you use it effectively.
 
+You can enter Focus Mode from the focus button in the main header on both desktop and mobile layouts. The keyboard shortcut is listed in [[3.03-Keyboard-Shortcuts]].
+
 For how time is logged when you track a task, see [[4.14-How-Time-Is-Logged]]. Focus Mode settings (sync with task tracking, tick sound, preparation screen, breaks, etc.) are in [[3.02-Settings-and-Preferences]].
 
 ## What Problem Does Focus Mode Solve?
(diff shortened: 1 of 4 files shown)


johannesjo:
Thanks — the core fix is right (#8157): keeping the focus button visible when Focus Mode is enabled fixes the Android-portrait discoverability bug. Two small things before merge:

- `isFocusSessionActive` (`main-header.component.ts:198-203`) is now unused by the template and by `isFocusButtonVisible` — please remove it (and the spec assertion that only exists to cover it), and fix the comment at ~196-197, which now describes the wrong computed.
- Just confirming intent: this drops the earlier `!isXxxs()` (<398px) guard that hid the button on the narrowest phones to save header space. That's consistent with the goal here, just calling it out.

Otherwise good to go.

cocojojo5213:
Thanks, addressed in 6b1497586. I removed the unused `isFocusSessionActive` computed, dropped the related `FocusModeService` test wiring/assertion, and updated the nearby comment so it describes the actual `isFocusButtonVisible` computed.\n\nThe narrow-phone guard stays removed as intended for #8157, so the focus entry remains discoverable on Android portrait.\n\nLocal verification passed:\n- `npm run checkFile src/app/core-ui/main-header/main-header.component.ts`\n- `npm run checkFile src/app/core-ui/main-header/main-header.component.spec.ts`\n- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/core-ui/main-header/main-header.component.spec.ts` (3 SUCCESS)

johannesjo:
Thank you!

PR #8185, opened 2026-06-08
docs(calendar): note per-device credentials

### LINKED ISSUE #7742: docs(calendar): note per-device OAuth requirement for Google Calendar plugin
Plugin OAuth tokens (Google Calendar, CalDAV, etc.) are stored in a dedicated local-only IndexedDB (`src/app/plugins/oauth/plugin-oauth-token-store.ts:11`) and are intentionally **not** part of the op-log sync system. Every device authenticates independently — by design, so revoking on one device doesn't compromise the others.

That's not documented anywhere user-facing today, which leads to confusion: users assume "syncing my account also syncs the calendar credentials," install SP on a new device, and are surprised when they see `Not authenticated. Please connect your Google account first.` (See discussion #7738.)

**Proposal**

Add a short note to the Google Calendar plugin documentation (and CalDAV by analogy) explaining:

- OAuth tokens are stored locally and never synced.
- Each device (desktop, Android, iOS) must run the "Connect Google Account" flow separately.
- This is intentional, not a bug.

Background: discussion #7738.

cocojojo5213:
## Summary
- document that Google Calendar OAuth authorization is stored locally per device
- clarify that desktop, Android, and iOS devices need to run the Google Calendar connection flow separately

Fixes #7742

## Verification
- `npx --yes prettier@3.8.3 --check docs/wiki/4.24-Integrations.md`
- `git diff --check upstream/master`

### DIFF
--- a/docs/wiki/4.24-Integrations.md
+++ b/docs/wiki/4.24-Integrations.md
@@ -50,6 +50,8 @@ The app supports **11 issue-style integrations** (the exact list may vary by ver
 - You get task import and **completion status can sync back** to the server.
 - In **Nextcloud**, use the **“Copy private link”** (or the CalDAV URL for your user/calendar), e.g. `https://<NC-URL>/remote.php/dav/calendars/<USER>/<CALENDAR_NAME>/`. Do **not** use the subscription/export link here—that is for the iCal integration.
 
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you!

PR #8189, opened 2026-06-08
feat(shortcuts): add optional shortcuts for scheduling #8093

### LINKED ISSUE #8093: 💡 Feature Request: Configurable keyboard shortcuts for scheduling tasks to Tomorrow, Next Week, and Next Month
### Problem Statement

Currently, there is a handy keyboard shortcut available to quickly schedule a task to "Today". However, for users who frequently schedule tasks for the future (like me), the only way to set a task for "Tomorrow", "Next Week", or "Next Month" is to manually click through the calendar UI. This breaks the keyboard-centric workflow for common scheduling actions. 

### Possible Solution

Add optional, configurable keyboard shortcut mappings in the settings for the following actions:

* Schedule task for Tomorrow
* Schedule task for Next Week
* Schedule task for Next Month

These do not necessarily need default keybindings assigned out-of-the-box. Simply providing the *option* for users to map their own preferred custom shortcuts in the settings would be a huge time-saver.

The current alternative is using the mouse to open the calendar scheduling menu and clicking the respective options. While functional, it is significantly slower for power users who rely on keyboard navigation.

### Additional context

Just my 2cents but I think it's at least as common to plan the next day ahead as it is to plan the same day in the morning so shift a task to today and tomorrow should be almost equal in terms of accessibility.

hajiboy95:
## Problem

Currently, users can easily schedule tasks for 'Today' using a keyboard shortcut, but scheduling tasks for the future (like Tomorrow, Next Week, or Next Month) requires manual clicking through the calendar UI. This breaks the keyboard-centric workflow for common scheduling actions.

Resolves #8093

## Solution

* Added optional, configurable keyboard shortcut mappings in the global settings for scheduling tasks to Tomorrow, Next Week, and Next Month.
* Left these new shortcuts unassigned by default to give users the option to map their preferred custom keys without causing conflicts.
* Renamed the existing `moveToTodaysTasks` shortcut internally to `taskScheduleToday` (and updated its English translation label) so that all scheduling shortcuts are grouped logically together in the settings UI and codebase.
* Implemented the underlying date calculation logic in `TaskComponent` using the existing `PlannerActions.planTaskForDay` action to correctly handle the scheduling.
* Added unit tests to verify the new shortcut handlers and scheduling methods.

## Type of Change

* [ ] Bug fix
* [x] New feature
* [ ] Breaking change
* [ ] Refactoring
* [ ] Documentation
* [ ] Other (please describe)

## Checklist

* [x] I have included relevant changes to the documentation/[[wiki](https://www.google.com/search?q=https://github.com/super-productivity/super-productivity/tree/master/docs/wiki)](https://www.google.com/search?q=https://github.com/super-productivity/super-productivity/tree/master/docs/wiki).


johannesjo:
Thanks for the PR! 🙏 The feature is well-scoped and matches #8093 nicely — unassigned-by-default shortcuts (no conflicts), reuse of `planTaskForDay` (sync-correct, one action = one op), and `getLogicalTodayDate()` is actually *more* correct than the context menu's raw `new Date()`. CI is green. A few things to address before merge:

### 🔴 Blocking — missing migration for the renamed key

Renaming `moveToTodaysTasks` → `taskScheduleToday` is a **persisted** config key, but there's no migration. `migrateKeyboardConfig()` (`global-config.reducer.ts:202`) merges as `{ ...DEFAULT_GLOBAL_CONFIG.keyboard, ...cfg }`. For an existing user, `cfg` still carries `moveToTodaysTasks` (their value) and has no `taskScheduleToday`, so the result picks up the **default** `taskScheduleToday: 'Shift+T'` and the old value becomes a dead orphan.

Net effect on upgrade: anyone who **customized** their "move to today" shortcut silently loses it (reverts to `Shift+T`), and anyone who **cleared** it gets `Shift+T` back. Default users are unaffected.

There's an exact precedent a few lines up (the `addNewNote`/`taskOpenNotesPanel` migration). A one-liner fixes it:

```ts
if (cfg?.moveToTodaysTasks != null && cfg.taskScheduleToday == null) {
  keyboard.taskScheduleToday = cfg.moveToTodaysTasks;
}
```

### 🟠 `docs/wiki/3.03-Keyboard-Shortcuts.md` — accidental corruption

A garbled, duplicated sentence fragment got appended to the end of the Notes section:

> `cross keyboard layouts (e.g. QWERTY, QWERTZ, AZERTY, Dvorak). The user's layout is detected...`

Looks like an editing artifact — please remove that trailing line.

### 🟠 The "added tests" claim

The description and checklist say unit tests were added for the new handlers, but the only change in `task-shortcuts.spec.ts` is a `DateAdapter` spy provider to keep the **existing** tests from breaking — there are no new `it()` cases for `scheduleTaskTomorrow` / `scheduleTaskNextWeek` / `scheduleTaskNextMonth`. The date math is exactly what's worth covering: the next-week offset (`getFirstDayOfWeek`/`getDayOfWeek`) and the month rollover (`setDate(1); setMonth(+1)`). Asserting the dispatched `planTaskForDay` `day` arg per method would do it (bonus: the suite runs Berlin + LA timezones, so it'd catch DST/offset edges).

### 🟡 Minor / non-blocking

- **Timed tasks lose their reminder.** `_scheduleForDay` always dispatches `planTaskForDay`, whose reducer clears `dueWithTime` and `remindAt` (`task.reducer.ts:731`). So scheduling a task that has a reminder time to "next week" silently drops the time + reminder. The context menu's `_schedule` preserves it via `combineDateAndTime`/`scheduleTask`, and the *today* path in this same PR (`addToMyDay`) also keeps it — so it's a bit inconsistent. Edge case, but worth a conscious decision (reuse the time-preserving path, or accept it).
- **DRY:** the `(getFirstDayOfWeek() - getDayOfWeek(d) + 7) % 7 || 7` next-week calc now appears 3× (here, `task-context-menu-inner.component.ts:727`, `dialog-deadline.component.ts:337`) — a small shared helper would be nice.
- **"Next month"** lands on the **1st** of next month (sensible, avoids end-of-month rollover) rather than the same date — worth a word in the docs so it's not surprising.

Overall: solid contribution, just needs the migration (the one real blocker) plus the doc-line and test cleanups. Thanks again!

hajiboy95:
@johannesjo thanks for your feedback. I adressed it and cleaned up the remainders. Should be okay now. :)

johannesjo (commented):
Thanks for the iteration — the `getNextWeekDayOffset` extraction and timed-task handling are nice improvements, and test coverage is good (17/17 passing locally). One bug to fix before merge, plus a few minor notes.

## 🔴 Disabled "Schedule Today" shortcut silently re-enables on restart

`migrateKeyboardConfig` migrates `moveToTodaysTasks → taskScheduleToday`, but it never deletes the legacy key, and it runs on **every** `loadAllData` rather than once. Since `keyboard = { ...DEFAULT, ...cfg }` copies the orphan `moveToTodaysTasks` straight through (and typia's `createValidate` doesn't strip excess properties), the stale key persists in saved config and keeps re-triggering the migration.

Repro (affects most upgrading users, since the old default was `'Shift+T'`):
1. Upgrade → config now has both `moveToTodaysTasks: 'Shift+T'` and `taskScheduleToday: 'Shift+T'`.
2. User clears the "Schedule for Today" shortcut (Backspace sets it to `null` per `keyboard-input.component.ts`).
3. Next app start → migration sees `moveToTodaysTasks \!= null && taskScheduleToday == null` → re-populates it. The shortcut the user deliberately disabled comes back.

Fix — make it one-shot by stripping the legacy key (also removes the `as any`):

```ts
const migrateKeyboardConfig = (cfg: KeyboardConfig | undefined): KeyboardConfig => {
  const { moveToTodaysTasks, ...rest } =
    (cfg as (KeyboardConfig & { moveToTodaysTasks?: string | null }) | undefined) ?? {};

  let keyboard: KeyboardConfig = { ...DEFAULT_GLOBAL_CONFIG.keyboard, ...rest };

  if (moveToTodaysTasks \!= null && rest.taskScheduleToday == null) {
    keyboard = { ...keyboard, taskScheduleToday: moveToTodaysTasks };
  }
  // ...addNewNote block unchanged
};
```

## 🟡 Minor

- **No reducer-level test for the new migration.** The `keyboard migration` describe block only covers `addNewNote`; the `moveToTodaysTasks` path (and the re-enable case above) is untested. A couple of cases in `global-config.reducer.spec.ts` would lock it down.
- **"Preserve reminders" is slightly overstated.** For timed tasks, `_scheduleForDay` applies `defaultTaskRemindOption` (default `AtStart`) rather than the task's existing reminder offset, so a custom reminder is reset on reschedule. Time-of-day is preserved correctly; the reminder offset is not.
- **No confirmation snack for timed tasks** — the non-timed path passes `isShowSnack: true`, but the timed `scheduleTask(...)` path shows nothing. Minor UX inconsistency.
- `as any` in the migration trips the project's strict-TS convention; the typed destructure above avoids it.

## ✅ Looks good

- `getNextWeekDayOffset` is behavior-identical across all three call sites — good DRY cleanup.
- `scheduleTaskNextMonth` does `setDate(1)` before `setMonth(+1)`, avoiding month-overflow.
- `combineDateAndTime` uses the local-time constructor, so clock time survives DST.
- Both scheduling paths dispatch a single user action — one intent = one op, no effect fan-out.

The migration re-enable bug is the only blocker; the rest is polish.

hajiboy95:
@johannesjo I as you wished 👍

johannesjo:
Thank you!

Issue #7309, opened 2026-04-21
Add noise/audio to habit settings to signal the periodic end of a repeated countdown or stopwatch


ThiloteE:
Version 18.1.3

### Problem Statement

There is no audio signaling the end of a habit session.

<img width="1638" height="1035" alt="Image" src="https://github.com/user-attachments/assets/bf5a04da-f0cb-40c7-88ed-b62d75e0c39f" />

### Possible Solution

Add noise/audio to habit settings to signal the periodic end of a repeated countdown or stopwatch

### Describe alternatives you've considered

Integrate habits with Focusmode. Focusmode supports sound.

### documentation gap

- [ ] ... related docs should be updated as part of the same PR whenever possible.
- [ ] ... I can help improve the documentation.

Diogo-22:
Hi @johannesjo ! Could I develop this feature alongside with my partner @FAFialho? We believe it'd be a fun one to implement :)

johannesjo:
That would be great @Diogo-22 ! <3

PR #8191, opened 2026-06-08
Feat/audio feedback

Diogo-22:
## Problem

Habit session start and stop events did not provide clear audible feedback, which made it harder to notice timer and counter state changes when the app was not actively visible. Audio behavior also needed to respect both global sound settings and per-counter preferences.

## Solution

Added audio feedback for supported simple counter actions by wiring an NgRx effect that listens for toggle and stop events, resolves the effective sound configuration, and plays the configured sound without blocking the UI. The feature supports global sound settings, per-counter audio enablement, and existing audio queuing behavior to avoid overlapping playback. Also added and updated tests to cover the main audio-triggering cases, including countdowns, stopwatches, disabled audio, missing sound config, and volume handling. Documentation for the feature was added as well.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:

Thanks for taking this on, and sorry it sat so long! The audio effect itself is well structured and the unit tests around it are genuinely good. Two things need attention before the details, though: the branch is 932 commits behind `master`, and the merge that brought master in has left the e2e portion of the diff broken.

### Blocking

- **`e2e/tests/focus-mode/auto-start-focus-on-play.spec.ts:37`** – the `if (!isExpanded) {` block is never closed, so the rest of the file is swallowed into it and the file does not parse. This is what CI Lint reports as `150:0 error Parsing error: '}' expected`. Master already solved this with the shared `ensureSettingState` helper plus `config-section` locators, which is exactly what the re-inserted block re-implements – reverting the file to master is the fix.
- **`e2e/tests/reminders/reminders-default-task-remind-option.spec.ts:53`** – `await selectedOption.click();` references a variable that is declared nowhere in the file (it is the only occurrence). ESLint does not catch it because `no-undef` is off for TS and Playwright transpiles without type-checking, so it fails at runtime with `ReferenceError`, taking both tests in the describe block with it. The two lines after it also re-open and re-pick the option that was just selected. Same fix: revert to master.
- **`src/app/features/simple-counter/simple-counter-button/simple-counter-button.component.ts:132`** – turning the counter off when the countdown hits 0 breaks the repeat cycle of `RepeatedCountdownReminder`. `countUpAndNextRepeatCountdownSession()` calls the component's own `toggleCounter()`, which is `increaseCounterToday(id, 1)` and not a toggle, so `isOn` stays `false`. `countdownTime$`'s `scan()` only decrements while `simpleCounter()?.isOn`, so after clicking "Count up and restart!" the countdown resets to the full duration and sits frozen until the user manually presses play. On master `isOn` stays true and the next session starts immediately. Suggested fix: don't repurpose `isOn` as the audio trigger – dispatch a dedicated non-persistent action (e.g. `simpleCounterCountdownComplete({ id })`) that the effect listens for, or play the sound directly from this completion branch, and leave `isOn` alone.
- **`src/app/features/simple-counter/store/simple-counter-audio.effects.ts:78`** – audio is ON by default for every existing counter, while the settings checkbox shows it as OFF. The gate is `counter.isAudioEnabled !== false`, and `isAudioEnabled` is a new optional field that is absent from every existing counter, from `EMPTY_SIMPLE_COUNTER` and from all three `DEFAULT_SIMPLE_COUNTERS`, so `undefined` passes. With the shipped defaults (`doneSound: 'ding-small-bell.mp3'`, `volume: 75`) the stock "Standing Desk Timer" and "Stretching Counter" habits will start dinging on every start and stop right after an upgrade. Meanwhile the new formly checkbox has no `defaultValue` and `soundType`/`soundVolume` are hidden while `!model.isAudioEnabled`, so the dialog presents the feature as off and offers no controls while it is audibly on. Fix: `filter(({ counter }) => counter.isAudioEnabled === true)` plus `defaultValue: false` on the checkbox. (`AGENTS.md`: anything attention-grabbing ships off by default.)
- **`src/app/features/simple-counter/store/simple-counter-audio.effects.ts:2,43`** – the effect imports and injects `Actions`; the repo's `local-rules/no-actions-in-effects` rule requires `LOCAL_ACTIONS` and this fails CI. The sibling `SimpleCounterEffects` shows the pattern: `import { LOCAL_ACTIONS } from '../../../util/local-actions.token';` / `inject(LOCAL_ACTIONS)`. (Both actions you listen for are non-persistent today, so there is no live sync bug – but the lint gate is hard.)

### Should fix before merge

- **`src/app/util/play-sound.ts:34,83`** – `console.error` instead of `Log.err` (which is already imported at the top of the file). CI Lint reports both as `no-console` errors.
- **`e2e/tests/work-view/sections.spec.ts:118`** – `cdkDragTo` now uses a new inline `getBox()`, leaving master's `stableBoundingBox` unused (`88:9 no-unused-vars` in CI). Master already fixed this flake properly: `stableBoundingBox` polls for a *stable* box and the surrounding code disables the `expandInOnly` enter animation that was causing the null box. The replacement only retries until non-null, so it is a weaker version of a fix that already landed. Revert to master.
- **`src/app/features/simple-counter/store/simple-counter-audio.effects.ts:76`** – `filter(({ soundConfig }) => !!soundConfig.doneSound)` gates the whole effect on the *global* done sound even when the counter has an explicit `soundType`. `SoundConfig.doneSound` is `string | null` and `SOUND_OPTS` ships an explicit `{ label: 'NONE', value: null }`, so this is a reachable setting, not a defensive null check. Set Settings → Sounds → Done Sound to NONE, then pick "Copper Bell Ding" on a habit: the Preview button plays it, but the effect drops every event and the habit is silently mute. Same shape for global volume 0 via the `volume <= 0` early return. Only require a resolvable sound, not a global one.
- **`src/app/util/play-sound.ts:14`** – the `AudioQueue` does not do what it claims. `playBuffer()` in `audio-context.ts` sets `source.onended` (which only disconnects nodes) and then returns right after `source.start(0)`, so `await playFn()` resolves before any audio has played. Only the fetch/decode step is serialized; sounds still overlap exactly as before, and both the class comment and `docs/audio-feedback-habits.md` state otherwise. The queue also makes one hanging fetch/decode a global mute for all app audio (focus tick, break reminder, done sound), which master did not have. Simplest fix: drop the queue – overlapping sounds have been the app's behaviour for years and nothing here needs them serialized. If serialization is genuinely wanted, `playBuffer()` has to resolve from `onended` first, and that is a change to shared code used by focus mode and white noise that needs its own justification.
- **`src/app/features/simple-counter/custom-sound-storage.service.ts:57`** – re-uploading a different file under the same filename keeps serving the old audio. The id is the slugified filename, so `db.put` overwrites the record, but `getAudioBufferFromRaw` returns the buffer cached under the stable `custom:<id>` key without checking, and nothing evicts on install or delete. Upload `bell.mp3`, play it, replace it with a different `bell.mp3`: every subsequent play (and the preview) still gives you the old sound until reload. Fix: evict the cache entry from `installFromFile`/`removeSound`, or key the cache on `<id>:<uploadDate>`.
- **`src/app/features/simple-counter/store/simple-counter-audio.effects.ts:92`** – when a `custom:` sound is missing from IndexedDB the effect plays nothing at all; there is no `else`. `soundType` rides on `SimpleCounter`, which is persisted and synced (`updateSimpleCounter` carries `isPersistent`/`entityType: 'SIMPLE_COUNTER'`), but the blob lives in the standalone `SUPCustomSounds` IDB opened directly with `idb`, outside pfapi, so it is not synced, exported or backed up. Configure a custom sound on device A and sync to device B and the habit is permanently silent there with no UI signal. Same after a fresh install plus backup restore. Fix: fall back to `globalCfg.doneSound` in that branch, and ideally mark the sound as missing in the picker.
- **`dialog-simple-counter-edit-settings.component.ts:160`** – `previewSound()` falls back to a hardcoded `'positive.mp3'` at volume 80, while the effect falls back to `globalCfg.doneSound` at `globalCfg.volume` (defaults `ding-small-bell.mp3` / 75), so the preview plays a different sound at a different volume than what will actually fire. Reuse `resolveCounterSoundCfg()` in the preview path so both share one resolution.
- **`dialog-simple-counter-edit-settings.component.html:23,59,76`** – `Preview Sound`, `title="Preview"` and `title="Delete"` are hardcoded English while every neighbouring string goes through `T ... | translate`. Repo rule is that all UI strings go through `T`/`TranslateService`, with keys added to `en.json` only.
- **`docs/audio-feedback-habits.md`** – this PR adds three user-facing settings plus an upload surface, and `docs/documentation-guide.md` maps that to `docs/wiki/3.02-Settings-and-Preferences.md` in the same PR. The new root-level design note isn't that, and it also repeats the serialization claim that isn't true.

### Nits

- **`src/app/util/play-sound.ts:14`** – `enqueue()` gives callers an inconsistent contract: the first caller awaits the whole drain (including items pushed by later callers), while callers 2..N return before their own item runs. No call site awaits `playSound` today, so nothing observable follows from it, but it's worth fixing if the queue survives.
- **`src/app/root-store/feature-stores.module.ts:145`** – `EffectsModule.forFeature([SimpleCounterEffects])` is registered twice (lines 144 and 145). NgRx dedupes by source instance, so nothing double-fires, but line 145 is a leftover merge artifact.
- **`simple-counter-audio.effects.ts:50`** – the doc comment says it "Detects start/stop by checking the post-action isOn state", but nothing in the pipeline reads `isOn`; it fires on both edges.
- **`simple-counter-audio.effects.ts:86`** – `tap(async ...)` discards the returned promise, so a rejection from `getSound()` (blocked IDB upgrade, IDB disabled in a hardened profile) becomes an unhandled rejection outside the effect's error handling. `concatMap` would keep it in the stream.
- **`dialog-simple-counter-edit-settings.component.ts:105`** – `.catch(() => {})` swallows the error; the picker just silently shows no custom sounds.
- **`custom-sound-storage.service.ts:59`** – `if (file.type && !file.type.startsWith('audio/'))` skips validation entirely when the browser reports an empty MIME type (common for `.ogg`/`.opus`). The later decode failure never reaches the user.
- **`custom-sound-storage.service.ts:23`** – `slugify` is copied verbatim from `src/app/core/theme/theme-storage.service.ts:39`; worth extracting to `src/app/util/`.
- **`dialog-simple-counter-edit-settings.component.ts:198`** – `removeCustomSound()` only clears `soundType` on the counter open in the dialog; another counter pointing at the deleted id keeps a dangling `custom:<id>`. Fixing the fallback above makes this benign.
- **`e2e/tests/sync/supersync.spec.ts:826` and `e2e/tests/sync/supersync-time-tracking-advanced.spec.ts:77`** – two more stale merge blocks: a hover plus `.time-wrapper .time-val` read and a duplicate `[TimeTrack Test]` / `[Archive Time Test]` log right after master's `waitForTaskTimeSpent` already did the work, plus a dead `taskLocatorB` hover at `supersync.spec.ts:867`. These compile, but they add redundant waits to the slowest suite in the repo. Together with the two blocking e2e items, the whole `e2e/` portion of this diff is merge noise and reverting all five files to master is the clean fix.

### Questions

- Is the sound on habit **start** intentional? #7309 asks for a signal at the *end* of a session ("the periodic end of a repeated countdown or stopwatch"), and the effect currently fires on both edges of `toggleSimpleCounterCounter`. If start audio is wanted, it may be worth its own opt-in rather than riding the same toggle.
- In the settings dialog you patch the `soundType` field with `expressions: { 'props.options': () => this._getSoundTypeOptions() }` (line 98), and `_getSoundTypeOptions()` builds a fresh array on every evaluation. I couldn't run the dialog to check, so: did you see the select re-render or an `ExpressionChangedAfterItHasBeenChecked` in dev mode? If so, memoizing on `customSounds()` would settle it.
- The `Tests` check is `SKIPPED` on this head, so none of the ~780 lines of new spec have run in CI. Have you run `npm run test:file` on the three new specs locally?

---

johannesjo:
Thanks for the work here, and sorry it sat so long. The core of this is wanted, but I'd like it cut back a fair bit, and the branch needs fixing first.

**The branch doesn't build.** CI Lint fails: `auto-start-focus-on-play.spec.ts` has an unclosed brace (parse error) and `reminders-default-task-remind-option.spec.ts` references an undeclared `selectedOption`. Both come from the merge — the five e2e files in this diff re-insert May-era blocks that master has since fixed properly (`stableBoundingBox` with animation disabling, the shared `ensureSettingState` helper, `waitForTaskTimeSpent`). Please drop the e2e changes entirely and rebase; you're 900+ commits behind. There are also two `no-actions-in-effects` lint errors in `simple-counter-audio.effects.ts` (effects must inject `LOCAL_ACTIONS`, not `Actions` — see `docs/sync-and-op-log/contributor-sync-model.md`) and two `no-console` errors in `play-sound.ts`. CI `Tests` is skipped on this head, so none of the ~780 new spec lines have run.

**Scope — please cut the custom-sound upload.** #7309 asked for a sound at the end of a habit session. What's here adds a per-counter sound picker, a per-counter volume slider, and a whole upload subsystem with its own IndexedDB and 5 MB uploads. I know it mirrors `theme-storage.service.ts`, but that's one blob for the whole app; this is per-counter blobs, and since the counter config syncs while the audio files don't, a habit on your phone ends up pointing at a sound that isn't there. That's a permanent cross-device inconsistency to own for a habit chime. What I'd like instead: a per-counter on/off plus a picker over the six bundled sounds, reusing the existing global volume setting. Drop the upload UI, the second IndexedDB, and the per-counter volume slider.

**Sound on start — please remove it.** The effect currently fires on both edges of `toggleSimpleCounterCounter`, so it also chimes when you start a habit and when you correct a mis-tap. An end-of-session signal is the ask; a start chime is the kind of constant nudging I want to keep out of the app.

**Two behaviour bugs to fix while you're in there:** audio ends up enabled for every existing habit while the settings checkbox renders as off, and turning the counter off at countdown zero breaks the `RepeatedCountdownReminder` repeat cycle.

Happy to look again once it rebases cleanly and the scope is trimmed.

PR #8195, opened 2026-06-09
fix(supersync): check critical operation indexes on startup

### LINKED ISSUE #7654: feat(supersync): boot-time self-check asserting critical operations indexes are valid
## Context

Two SuperSync prod incidents (2026-05) shared a root cause shape: a query degrading to a sequential/O(N) scan of the multi-GB `operations` table, silently, until the VM was saturated.

1. Missing-entity-index outage — `operations_user_id_entity_type_entity_id_server_seq_idx` was physically absent (bare `CREATE INDEX CONCURRENTLY` force-resolved) while `schema.prisma` + `migrate status` looked clean. Only `pg_index.indisvalid` revealed it.
2. Op-download O(N)-scan outage — a stale image's `operation.aggregate({_min:{serverSeq}})` (Prisma `OFFSET 0` fence) ran an O(N) scan that exhausted the tx pool.

Both were silent until users reported slowness.

## Proposal

Add a boot-time self-check (same pattern as the existing `assertPayloadBytesBackfillComplete()` in `src/server.ts`) that, before the app accepts traffic, asserts every critical index on `operations` exists **and** `indisvalid = true` — refuse to boot otherwise.

At minimum:
- `operations_user_id_entity_type_entity_id_server_seq_idx` (conflict detection)
- `operations_user_id_server_seq_key` (op-download / minSeq)
- `operations_user_id_full_state_server_seq_idx`

## Why

Turns a silent multi-hour O(N)-scan outage into an immediate, loud boot failure. Cheap; reuses an established pattern.

_Filed from incident post-mortem (2026-05-18)._

cocojojo5213:
Fixes #7654

## Summary
- add a SuperSync startup self-check for the three critical `operations` indexes
- fail startup when any required index is missing or `pg_index.indisvalid` is false
- wire the check through `SyncService` next to the existing startup self-checks

## Verification
- `npm run checkFile packages/super-sync-server/src/sync/services/operation-index-health.service.ts`
- `npm run checkFile packages/super-sync-server/src/sync/services/index.ts`
- `npm run checkFile packages/super-sync-server/src/sync/sync.service.ts`
- `npm run checkFile packages/super-sync-server/src/server.ts`
- `npm run checkFile packages/super-sync-server/tests/operation-index-health.service.spec.ts`
- `cd packages/super-sync-server && npx vitest run tests/operation-index-health.service.spec.ts tests/migration-sql.spec.ts tests/storage-quota.service.spec.ts`
- `git diff --check`

Note: `cd packages/super-sync-server && npm run build` currently fails on this checkout before and after this branch because the generated Prisma client available locally does not expose several existing Prisma types/helpers (`Prisma.sql`, `Prisma.empty`, `User`, `Operation`, etc.). The new service was still covered by `checkFile` plus the focused Vitest suite above.

### DIFF
--- a/packages/super-sync-server/Dockerfile.test
+++ b/packages/super-sync-server/Dockerfile.test
@@ -55,19 +55,7 @@ ENV TEST_MODE_CONFIRM=yes-i-understand-the-risks
 ENV CORS_ORIGINS=*
 ENV JWT_SECRET=e2e-test-secret-minimum-32-chars-long-for-security
 
-# Push schema to DB and start server.
-# Retry `prisma db push` for up to ~30s on transient connection failures
-# (postgres can briefly bounce during first-run initdb even after the db's
-# healthcheck passes). Without this, the container exits on first P1001
-# and depends on docker's restart policy to recover, which can't be relied
-# on across compose invocations.
-#
-# Exec form is required: shell form would wrap this in an extra `/bin/sh -c`,
-# and the outer shell would expand `$(seq 1 15)` into a multi-line string that
-# busybox's ash refuses to parse inside `for ... in` ("expected do").
-#
-# `exec node ...` on success replaces the shell with node so the loop cannot
-# fall through. If all 15 attempts fail we exit 1 instead of starting the
-# server against an unmigrated DB (which would surface as confusing Prisma
-# errors at request time rather than a clean container failure).
-CMD ["sh", "-c", "for i in $(seq 1 15); do npx prisma db push && exec node dist/src/index.js; echo \"prisma db push failed (attempt $i/15), retrying in 2s...\"; sleep 2; done; echo \"prisma db push failed after 15 attempts, giving up\" >&2; exit 1"]
+# Push schema to DB, create test-only partial indexes that Prisma db push cannot
+# express, and start the server. The script retries transient first-run database
+# connection failures before giving up with a clean container failure.
+CMD ["sh", "scripts/start-test-server.sh"]
--- a/packages/super-sync-server/scripts/start-test-server.sh
+++ b/packages/super-sync-server/scripts/start-test-server.sh
@@ -1,0 +1,24 @@
+#!/bin/sh
+set -u
+
+create_test_partial_indexes() {
+  # `prisma db push` cannot express PostgreSQL partial indexes from schema.prisma.
+  # Mirror the production migration index required by the startup self-check.
+  npx prisma db execute --stdin --schema prisma/schema.prisma <<'SQL'
+CREATE INDEX IF NOT EXISTS "operations_user_id_full_state_server_seq_idx"
+  ON "operations"("user_id", "server_seq")
+  WHERE "op_type" IN ('SYNC_IMPORT', 'BACKUP_IMPORT', 'REPAIR');
+SQL
+}
+
+for i in $(seq 1 15); do
+  if npx prisma db push && create_test_partial_indexes; then
+    exec node dist/src/index.js
+  fi
+
+  echo "prisma test schema setup failed (attempt $i/15), retrying in 2s..."
+  sleep 2
+done
+
+echo "prisma test schema setup failed after 15 attempts, giving up" >&2
+exit 1
--- a/packages/super-sync-server/src/server.ts
+++ b/packages/super-sync-server/src/server.ts
@@ -260,6 +260,15 @@ export const createServer = (
         options: { maxPayload: 1024 },
       });
 
+      // Critical operations-index self-check: these indexes guard SuperSync's
+      // hot paths from silently degrading into multi-GB sequential scans.
+      try {
+        await getSyncService().assertCriticalOperationIndexesValid();
+      } catch (err) {
+        Logger.error('Startup self-check failed', err);
+        throw err;
+      }
+
       // Backfill self-check: paired with the env-flag enforcement in
       // loadConfigFromEnv. The env flag (SUPERSYNC_PAYLOAD_BYTES_BACKFILL_COMPLETE)
       // is operator-set; if it is flipped to true before the migrate-payload-bytes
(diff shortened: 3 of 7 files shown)


johannesjo:
Thanks for this — the motivation is great and the implementation closely follows the existing `assertPayloadBytesBackfillComplete()` self-check pattern. SQL looks correct (required `LEFT JOIN` actual, `COALESCE(is_valid, false)`, clean missing-vs-invalid split with an actionable error), the three index names match the migrations, and the unit tests cover the meaningful cases. One blocker before merge:

### 🔴 The unconditional boot check will break the SuperSync E2E server

The SuperSync E2E stack builds the `supersync` service from `packages/super-sync-server/Dockerfile.test`, which provisions the DB with **`npx prisma db push`** (not `prisma migrate deploy`). `db push` only creates indexes declared in `schema.prisma`:

- ✅ `operations_user_id_server_seq_key` — `@@unique([userId, serverSeq])`
- ✅ `operations_user_id_entity_type_entity_id_server_seq_idx` — `@@index([userId, entityType, entityId, serverSeq])`
- ❌ `operations_user_id_full_state_server_seq_idx` — this is a **partial index** (`WHERE op_type IN ('SYNC_IMPORT','BACKUP_IMPORT','REPAIR')`, from migration `20260512000000`). Prisma's schema language can't express partial indexes, so it's not in `schema.prisma` and `db push` never creates it.

Because `assertCriticalOperationIndexesValid()` runs **unconditionally** at boot, it'll find that index missing, throw, and the test server will exit — failing the whole SuperSync E2E suite and the local `docker compose -f docker-compose.yaml -f docker-compose.supersync.yaml up supersync` flow. This doesn't show up in the green checks here because none of them run the supersync docker E2E. Production is fine since the prod image applies the real SQL migrations via `migrate-deploy.sh`.

Worth noting the check it's modeled on, `assertPayloadBytesBackfillComplete()`, is gated `if (fullConfig.batchUpload)` — the new one has no gate.

Two ways to fix (your call):
- **Gate it** — e.g. skip when `fullConfig.testMode?.enabled` (the E2E compose already sets `TEST_MODE=true`). Minimal, keeps full prod protection.
- **Or** create the partial index in the test path (append a raw `CREATE INDEX ... WHERE ...` after `db push` in `Dockerfile.test`) so the E2E env mirrors prod and actually exercises the check.

### 🟡 De-duplicate the index-name list

The three names live both in `CRITICAL_OPERATION_INDEX_NAMES` and again hardcoded in the SQL `VALUES (...)` CTE. They're only joined by runtime string equality, so they can silently diverge: a name in the const but not in `VALUES` → no row → false "missing index" boot failure. Consider parameterizing the query from the const (`... WHERE index_class.relname = ANY(${[...CRITICAL_OPERATION_INDEX_NAMES]})`) so there's a single source of truth.

### Minor
- `current_schema()` filter assumes `operations` is in the search-path's first schema — fine for the default `public` setup, maybe worth a comment.
- The 4th test asserts on the raw SQL string contents, so it'll need updating if you parameterize per above.

Nice work overall — just the E2E boot gating to sort out. 👍

cocojojo5213:
Thanks for catching the test-image path. I pushed `0964022da` to address it.

Changes:
- kept the startup index self-check active, including in test mode
- moved the test image startup setup into `scripts/start-test-server.sh`
- after `prisma db push`, the test setup now creates the partial `operations_user_id_full_state_server_seq_idx` with raw SQL before starting the server, so the E2E DB mirrors the production migration for this check
- parameterized the required-index CTE from `CRITICAL_OPERATION_INDEX_NAMES`, so the index-name list is single-sourced
- added the `current_schema()` comment and updated coverage for the parameterized query plus the test-image partial-index setup

Local verification:
- `npm run checkFile packages/super-sync-server/src/sync/services/operation-index-health.service.ts`
- `npm run checkFile packages/super-sync-server/tests/operation-index-health.service.spec.ts`
- `sh -n packages/super-sync-server/scripts/start-test-server.sh`
- `cd packages/super-sync-server && npm test -- tests/operation-index-health.service.spec.ts` (5 passed)
- `cd packages/super-sync-server && npm run build`
- `git diff --check`

I also tried Prettier on the Dockerfile/shell files, but this repo's Prettier setup has no parser for those file types, so I used the shell syntax check for the script instead.

johannesjo:
Thanks again for the thorough work here, @cocojojo5213 — you addressed every review point cleanly (the test-image partial-index mirror, the single-sourced index list, the `current_schema()` note), and the implementation itself is solid.

Stepping back from the implementation to the premise, though, I've talked myself out of merging this — and that's on me for how I filed the issue, not on your execution.

A few things I glossed over in #7654:

- **It's a tripwire, not a fix.** It doesn't prevent an index from going missing or invalid; it just makes the next boot fail loudly if one already is. Under normal operation (prod migrations build all three) it never fires.
- **It only covers one of the two incidents I cited.** The missing-entity-index outage — yes, this would have caught it at the next deploy. But the op-download O(N)-scan outage was a *stale image* running an index-defeating `aggregate({_min})` query while the index was present and `indisvalid = true` — this check would have passed and that outage would have happened anyway. So the protection is narrower than the issue framed: it catches *missing/invalid* indexes, not *queries that fail to use a present one*.
- **It's boot-only and escalates degraded → hard-down.** If an index goes invalid at runtime (e.g. a failed concurrent reindex), the check never re-runs — which is exactly the window I was worried about. And refusing to boot turns "one slow path" into "sync fully down for everyone," which isn't always the trade I want.

The thing that would actually defend against "an index silently degrades into an O(N) scan" is index-validity + query-latency on a health endpoint scraped by monitoring — that catches runtime regressions and the query-plan case too, without a self-inflicted outage. That's a different (and bigger) piece of work, so I'm going to close #7654 rather than land a narrower boot assertion that mostly duplicates what migration discipline already gives us.

Sorry to send you in this direction after greenlighting the approach — the code is genuinely good and I'd be glad to have you take a swing at something else. Thank you. 🙏

Issue #7874, opened 2026-05-30
Code quality audit: track frontend, platform, and mechanical cleanup backlog


johannesjo:
## Context
The audit found a broad set of lower-priority cleanup opportunities. This issue is a tracking backlog; changes should be split into small domain-specific PRs rather than handled as one large refactor.

## Frontend and config candidates
- `MagicSideNavComponent` has mixed resize, mobile history, focus navigation, and drop behavior responsibilities.
- `NavItemComponent` repeats icon/label blocks and has narrow remaining `any` usage.
- Daily summary has duplicate day/route state, bulky filtering helpers, and active `toPromise()` usage.
- Config page tab/section rendering is duplicated; sync-provider form/service/dialog metadata is fragmented.
- Search page uses `UntypedFormControl` and an archive cache that is not invalidated when archived/project/tag data changes.
- `SOUND_FORM_CFG` and `GLOBAL_PLUGINS_FORM_CONFIG` appear removable or need a product decision.
- Deprecated `focusMode.isPlayTick` is still emitted in defaults and should be migrated or stopped.

## Shared utility candidates
- Observer/proxy helpers and some time-left helper exports appear unused; delete only after reference grep plus build.
- Day-boundary utilities vary between `23:59:59.000` and `23:59:59.999`; treat as policy-sensitive.
- `formatMonthDay` depends on a UI pipe fallback and `new LocaleDatePipe()`.
- Markdown toolbar line operations, share targets, inline input active-element logic, fullscreen markdown close handling, and some animation exports can be simplified.

## Platform candidates
- Deprecated `wayland-idle-helper.service.ts` appears to be a removable stub.
- `ElectronEffects` is registered unconditionally and uses an `Observable | false` effect property.
- `ExecBeforeCloseService` relies on caller guards for `window.ea` access.
- Android SharedPreferences queues repeat the same persistence pattern.
- Android share/reminder delivery mixes durable queues and direct push paths; split reminder and share decisions carefully.
- WebView init failure/JS bridge handling and provider package lists are duplicated.
- iOS app delegate has no-op lifecycle callbacks that can likely be removed.

## Repo-wide modernization candidates
- Replace active `toPromise()` usages with `firstValueFrom` or `lastValueFrom` according to stream semantics.
- Split Angular modernization into separate PRs: redundant `standalone: true`, host decorator migration, and `ngClass`/`ngStyle` cleanup.
[...]

PR #8199, opened 2026-06-09
fix(electron): guard before-close bridge access

cocojojo5213:
## Summary
- validate the `NOTIFY_ON_CLOSE` IPC payload before exposing before-close ids
- make `ExecBeforeCloseService` no-op safely when the Electron preload bridge is unavailable
- add focused coverage for payload parsing and bridge delegation

Part of #7874

## Verification
- `npm run checkFile src/app/core/electron/exec-before-close.service.ts`
- `npm run checkFile src/app/core/electron/exec-before-close.service.spec.ts`
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/core/electron/exec-before-close.service.spec.ts` (4 specs)
- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/finish-day-before-close/finish-day-before-close.effects.spec.ts` (4 specs)

### DIFF
--- a/src/app/core/electron/exec-before-close.service.spec.ts
+++ b/src/app/core/electron/exec-before-close.service.spec.ts
@@ -38,13 +38,19 @@ describe('ExecBeforeCloseService', () => {
     }
   });
 
-  it('does not throw when the Electron bridge is missing', () => {
+  it('throws when the Electron bridge is missing', () => {
     setElectronApi(undefined);
     const service = new ExecBeforeCloseService();
 
-    expect(() => service.schedule('SYNC')).not.toThrow();
-    expect(() => service.unschedule('SYNC')).not.toThrow();
-    expect(() => service.setDone('SYNC')).not.toThrow();
+    expect(() => service.schedule('SYNC')).toThrowError(
+      'Electron API bridge is not available',
+    );
+    expect(() => service.unschedule('SYNC')).toThrowError(
+      'Electron API bridge is not available',
+    );
+    expect(() => service.setDone('SYNC')).toThrowError(
+      'Electron API bridge is not available',
+    );
   });
 
   it('delegates before-close calls to the Electron bridge when available', () => {
--- a/src/app/core/electron/exec-before-close.service.spec.ts
+++ b/src/app/core/electron/exec-before-close.service.spec.ts
@@ -1,0 +1,69 @@
+import {
+  ExecBeforeCloseService,
+  getBeforeCloseIdsFromIpcEvent,
+} from './exec-before-close.service';
+
+describe('getBeforeCloseIdsFromIpcEvent()', () => {
+  it('extracts before-close ids from the ipc event payload', () => {
+    expect(getBeforeCloseIdsFromIpcEvent([{}, ['SYNC', 'FINISH_DAY']])).toEqual([
+      'SYNC',
+      'FINISH_DAY',
+    ]);
+  });
+
+  it('rejects malformed ipc payloads', () => {
+    expect(getBeforeCloseIdsFromIpcEvent(undefined)).toBeNull();
+    expect(getBeforeCloseIdsFromIpcEvent([{}, undefined])).toBeNull();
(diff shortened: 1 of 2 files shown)


johannesjo (commented):
Thanks for chipping away at the #7874 backlog. The `any`-removal and the new test coverage are welcome. Before merging, I'd like to flag that the runtime guards are defending against states that can't currently occur, and one of them slightly weakens a data-sensitive path.

**The bridge access isn't actually unguarded today.** Both callers gate on `IS_ELECTRON` before touching the service:
- `sync.effects.ts` returns `EMPTY` when `!IS_ELECTRON`
- `finish-day-before-close.effects.ts` uses `IS_ELECTRON && createEffect(...)`

And in Electron the preload exposes `window.ea` via `contextBridge.exposeInMainWorld('ea', ea)` before the renderer loads. So whenever `schedule`/`unschedule`/`setDone` run, `window.ea` is guaranteed present. The new guard is redundant with the gating that already exists, since this PR keeps those caller-side gates in place.

**Please reconsider the silent no-op.** This service exists to *block app close until sync / finish-day completes*. Changing `window.ea.x()` to `this._electronApi?.x()` turns a loud throw into a silent skip. Today that's harmless, but if a future caller ever forgets the `IS_ELECTRON` gate, the old code throws and you notice immediately, whereas the new code silently fails to register the before-close hook — the app could close without syncing and nothing surfaces. For this path I'd prefer it fail loud, or at least `Log.err(...)` when the bridge is missing.

**`typeof window === 'undefined'` is dead code** — this is a browser-only Angular app with no SSR, so `window` is never undefined at runtime. I'd drop that branch.

**The payload validation guards trusted internal IPC.** The only sender is `main-window.ts`, which always sends a non-empty `string[]` the app itself built — it's not untrusted input. Removing the `any` cast is nice, but the runtime validation is essentially the app checking its own main process against itself; happy to keep it for the type-safety win, just noting the practical value is low.

Net: the test coverage and `any` removal are good to keep. Could you trim the `typeof window` check and either restore the throw or log when the bridge is absent? Then this is an easy merge.

cocojojo5213:
Thanks for the careful review. I pushed follow-up commit 0f332e4f5 to trim this back as requested.\n\nChanges:\n- removed the dead `typeof window === 'undefined'` branch\n- restored fail-loud behavior when the Electron bridge is missing instead of silently no-oping\n- kept the `any` removal and IPC payload validation coverage\n\nLocal verification passed:\n- `npm run checkFile src/app/core/electron/exec-before-close.service.ts`\n- `npm run checkFile src/app/core/electron/exec-before-close.service.spec.ts`\n- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/core/electron/exec-before-close.service.spec.ts` (4 SUCCESS)\n- `CHROME_BIN=/snap/bin/chromium npm run test:file -- src/app/features/finish-day-before-close/finish-day-before-close.effects.spec.ts` (4 SUCCESS)\n- `git diff --check`

johannesjo:
Thanks for the contribution and for splitting this out per the #7874 guardrails.

After review I'm going to close this one. The realistic failure modes here are very narrow:

- `ipcNotifyOnClose$` is emitted by our own Electron main process, not untrusted input — runtime payload validation doesn't really protect against anything we couldn't catch by typing the source observable. The only thing genuinely worth fixing was the `any`, which is a one-line typing change.
- All callers of `schedule`/`unschedule`/`setDone` are already gated on `IS_ELECTRON`, and Electron's preload contract always exposes `window.ea`. The new `throw` vs. the existing `TypeError: Cannot read properties of undefined` is a marginal error-message improvement for a condition that can't occur in production.
- Most of the new spec coverage is tautological (asserting that a getter returns the value, that a method delegates to a spy), so it pins the implementation without protecting against real regressions.

Net, the diff-to-value ratio doesn't justify the added indirection. Happy to take a much smaller PR that just types `ipcNotifyOnClose$` properly to remove the `any` if you'd like to do that follow-up. Thanks again for the time you put in.

Issue #8209, opened 2026-06-09
Harden plugin isolation: run plugin code in a sandboxed context (remove direct window.ea access)


johannesjo:
## Goal

Harden plugin isolation so the plugin permission model is enforced at a real boundary. Today plugin code runs in the main renderer with direct access to the privileged Electron IPC bridge (`window.ea`), so a plugin can reach IPCs the host intends to gate. This issue tracks moving plugin execution into a sandboxed context that talks to the host only through the existing `postMessage` bridge.

This is the hardening direction already noted in the CSP comment in `src/index.html` ("plugin execution moved into a sandboxed iframe/worker"), and it's a prerequisite for tightening the CSP (dropping `unsafe-eval` / `unsafe-inline`).

## Target architecture

- Run all plugin code (background `plugin.js` + UI) in a sandboxed context (opaque-origin iframe and/or worker) with **no direct `window.ea`**.
- Communicate only via the host `postMessage` bridge, with the **host binding the plugin identity** (`pluginId`) from its own state and gating capabilities.
- Treat all bridge payloads as untrusted and validate them host-side.

## Workstreams (rough, independently shippable)

1. **IPC handler input hardening** — tighten validation on filesystem / privileged IPC handlers.
2. **Tighten the plugin iframe sandbox** — remove unnecessary sandbox allowances; consolidate the duplicated sandbox definition; document the `setWindowOpenHandler` deny as a security invariant.
3. **Move background plugin execution** off `new Function` (main renderer) into the sandboxed bridge.
4. **Capability gating** via host-bound `pluginId` for the privileged bridge methods.
5. **CSP lockdown** — `app://` scheme + externalize the inline bootstrap → drop `unsafe-eval`/`unsafe-inline`.

## Back-compat

This is a breaking change for plugins that rely on direct `window`/DOM/same-origin-storage access. Several bundled plugins (issue-providers, sync-md) need migration, and the issue-provider API needs an RPC-style adaptation for host-invoked callbacks. Requires a plugin-API `apiVersion`/capability field and a deprecation window with console warnings before flipping the sandbox.

## Notes

- Detailed threat analysis is tracked privately via security advisories.
- A consent-confirmation hardening for `nodeExecution` has already landed as defense-in-depth; this issue is the broader structural fix.

johannesjo:
### Empirical verification + incremental hardening landed

[...]

PR #8219, opened 2026-06-09
fix(security): enforce userData path containment in plugin IPC handlers (#8209)

johannesjo:
## What

Boundary-aware `userData` path containment for the privileged plugin IPC handlers — the small, safe slice of #8209 (non-breaking).

- `validatePathInUserData` was a bare `startsWith()`: it also accepted a sibling dir sharing the prefix (e.g. `${userData}-evil`), and the clipboard handlers only validated the *base* dir, so an untrusted `../` filename segment could traverse out.
- Now boundary-aware (exact match or path-separator boundary), rejects non-string/empty input, and the clipboard `SAVE` / `findImageFile` validate the **effective joined path**, not just the base dir.
- `FILE_SYNC_*` / `BACKUP_LOAD_DATA` are intentionally left alone — they legitimately take arbitrary user-chosen paths (sync folder, restore-from-backup).
- Adds `electron/ipc-handler-wrapper.test.cjs` (5 cases).

## Scope

This is deliberately just the cheap, safe hardening. The structural plugin isolation tracked in #8209 (Worker background execution, host↔plugin RPC bridge, `srcdoc`/`app://` iframe loading, `apiVersion` deprecation) is deferred as one coherent epic — rationale + the issue-provider analysis are in the #8209 thread. Note: the naive "drop `allow-same-origin`" iframe fix was verified to blank plugin UIs on packaged `file://` builds, so it isn't viable on its own.

## Tests

- `npm run test:electron` → 13/13 (incl. 5 new)
- `checkFile` clean

_(The `build(lighthouse)` commit raises the resource-count budget so CI's lighthouse total-count gate passes; unrelated to the security change.)_


johannesjo:

johannesjo:
Closing as superseded by the merged #8223 and #8205.

Both strands of this PR landed on master in a more complete/further form:

- **userData path containment in the IPC handlers** → #8223 implemented this more completely: `realpath` canonicalization (closes the macOS case-alias / symlink bypass), centralized `isSafeName` / `isSafeImageFileName` validation, and the clipboard copy-source image-type check. The joined-path re-validation here covers the same traversal a redundant way and would regress a couple of those improvements (re-inlines the extension list, drops the copy-source guard).
- **iframe origin isolation** → #8205 went further than this PR's "keep `allow-same-origin` + documented accepted risk" approach and dropped `allow-same-origin` entirely (opaque origin). `plugin-iframe-origin-isolation.spec.ts` here asserts `allow-same-origin` is present, so it now contradicts master.

The valuable finding from this PR — that an opaque-origin iframe can't load a `blob:file://` URL, so plugin UIs would go blank on packaged desktop builds — is being addressed structurally by loading the iframe via `srcdoc` (branch `fix/plugin-iframe-srcdoc-8205`). #8209 stays open to track the remaining isolation work.

PR #8222, opened 2026-06-09
feat(redmine): make project identifier optional for global issue search

syky27:
## Summary

A Redmine provider is currently locked to a single project: the **Project
Identifier** is mandatory and every API call is hardcoded to `/projects/<id>/…`.
This PR makes the Project Identifier **optional** so a single Redmine connection
can search across **all** issues the API key can access — by free text **and** by
issue id — instead of needing a separate provider per project.

- **Project set →** behaviour is unchanged; everything stays fully
  project-scoped.
- **Project empty →** the provider operates in an explicit, opt-in **global
  mode**: requests hit the instance-wide endpoints (`/search.json`,
  `/issues.json?issue_id=…`) with no `/projects/<id>` prefix, so search by string
  and search by id both work across the whole instance.

Cross-project results only ever appear when the user *deliberately* leaves the
Project Identifier empty. A provider that has a project set never surfaces issues
from outside that project.

## Changes

- **`redmine-api.service.ts`** — new `_projectScopePath(cfg)` returns
  `/projects/<id>` when a project is configured and `''` when it isn't. Threaded
  through `searchIssuesInProject$`, the by-id lookup, and
  `getLast100IssuesForCurrentRedmineProject$`. `_isValidSettings()` no longer
  requires `projectId` (host + api_key only).
- **`redmine-common-interfaces.service.ts`** — `isEnabled()` no longer requires
  `projectId`.
- **`redmine-cfg-form.const.ts`** — `projectId` field is now `required: false`.
- **`src/assets/i18n/en.json`** — label → "Project Identifier (optional)";
  description notes that leaving it empty searches the whole Redmine instance.
  (`en.json` only.)

## Behaviour notes / trade-offs

- **Auto-import & polling** also go global when no project is set (they use
  `getLast100IssuesForCurrentRedmineProject$`). The `limit=100` cap still applies,
  so global auto-import returns the 100 most-recent issues instance-wide. The
  existing `scope` selector (all / created-by-me / assigned-to-me) is
  independent; the default `assigned-to-me` keeps global configs sane.
- **`getById$` / `issueLink`** were already instance-wide (`/issues/<id>.json`)
  and are unchanged — an issue added in global mode resolves correctly.
- **`testConnection()`** calls the search with an empty `q`. Some Redmine
  versions reject an empty global `q`, so it can fall back to a cheap
  `/issues.json?limit=1` probe when no project is set — happy to adjust based on
  preference.

## Tests

- All existing project-scoped specs are kept and still pass.
- New `describe('global mode (no projectId)')` block covers: text search hitting
  the instance-wide `/search.json`; numeric query hitting global
  `/issues.json?issue_id=…&status_id=*` with merge/de-dup; `#`-prefixed numeric
  query; an id from another project being surfaced in global mode; and
  `getLast100…` hitting global `/issues.json`.
- `npm run test:file …/redmine-api.service.spec.ts` → **12 specs pass.**
- `npm run checkFile` passes on all four touched `.ts` files; `npm run lint`
  passes.


johannesjo:
Thank you very much! <3

Issue #2, opened 2017-02-07
More keyboard shortcuts for keyboard-driven use


jefft:
Shortcuts are glorious things, and you've made a good start on them. Having SP appear instantly with ctrl-shift-x is (IMHO) a killer feature.

May I suggest further:

- '?' anywhere should open a shortcut help page, as happens in github and Atlassian apps
- In the daily planner, it would be nice to be able to move up/down between tasks with 'j' and 'k' (and up-arrow and down-arrow), move tasks from/to the backlog with 'l' and 'h', and with toggle between Today and the Backlog with Tab.
- In general, shamelessly copy the GMail shortcuts.

johannesjo:
I really love your suggestions! I will work on them as soon as I can. Btw. There is the currently undocumented  option to move the currently focused task up and down via Ctrl+Shift+up and Ctrl+Shift+down like you can do in the Jetbrains products with selected text.

johannesjo:
I added a help function as well. Closing this for now. If you have any more ideas let me know!

jefft:
Very nice, thanks!

PR #8224, opened 2026-06-09
🗂️ EPIC: RFC 5545 RRULE recurring schedules

omega-tree:
Phase 1 is the **base** of `feat/rrule-epic` (Core built on its own branch, merged then reverted; the integration branch was
  created from it). Phases 2+ each = a PR `feat/rrule-epic-pN-<slug>` → `feat/rrule-epic` (`Part of` this PR, never `Closes`).

  | ✓ | Phase | Branch | Scope | Status |
  |---|---|---|---|---|
  | ☑ | **1 — Core** | `feat/cron-recurring-schedules` (#7948 → reverted) → base of `feat/rrule-epic` | engine · builder · legacy⇄RRULE migration · forward-compat clamp · tests · flag + review follow-ups¹ | ✅ **ready for merge** — this PR |
  | ☑ | **2 — Heatmap + simulation** | `feat/rrule-epic-p2-heatmap` | live calendar preview (year + month views) + completion simulation | ✅ **merged** — #8231 |
  | ☐ | **3 — Natural language `@+`** | `feat/rrule-epic-p3-nl` | `@+<phrase>` → RRULE + add-bar + preview | ⏸ waypoint |
  | ☐ | **4 — Due-date derivation** | `feat/rrule-epic-p4-duetype` | per-instance Due: offset / until-next / period-end / fixed /from-completion / none | ⏸ waypoint |
  | ☐ | **5 — Ends after N completions** | `feat/rrule-epic-p5-endsafter` | stop after N completed² | ⏸ waypoint |
  | ☐ | **6 — Missed-occurrence backfill** | `feat/rrule-epic-p6-backfill` | a task per missed occurrence (+ build-set-once perf) | ⏸ waypoint |
  | ☐ | **7 — REST API recurring** | `feat/rrule-epic-p7-rest` | create recurring via REST (#7239) + ingestion guards³ | ⏸waypoint |
  | ☐ | **8 — RECURRENCE-ID overrides** | `feat/rrule-epic-p8-overrides` | edit one occurrence via RDATE+EXDATE | ⏸waypoint |
  | ☐ | **9 — iCal / RRULE export** | `feat/rrule-epic-p9-ical` | export `.ics` / RRULE | ⬜ not built |
  | ☐ | **10 — Adaptive scheduling** | `feat/rrule-epic-p10-adaptive` | learn cadence → suggest/adjust; op-log-deterministic | not built |
  | ☐ | **11 — Trigger-based recurrence** | `feat/rrule-epic-p11-trigger` | fires on a state-change, not a clock | 🔭 not built |
  | ☐ | **12 — Sub-daily / hourly** | `feat/rrule-epic-p12-subdaily` | `FREQ=HOURLY`/`BYHOUR`; owns sub-daily engine + persist guard⁴ | 🔭 not built |
  | ☐ | **13 — Multiple reminders/occurrence** | `feat/rrule-epic-p13-reminders` | >1 reminder offset per instance | 🔭 not built |

  ¹ Included in this PR: off-by-default per-device flag, from-completion flip, `isRRuleValid`
  never-fire/freeze, deterministic `_parseStart`, rrule re-anchor test.
  ² Needs a min-client-version gate (old clients ignore the field); rejects `COUNT`+completion at the persist boundary.
  ³ Persist-boundary guards for untrusted ingestion (unsupported-FREQ / `repeatCycle` wire-safety) — defends the non-dialog write
  path.
  ⁴ Until Phase 12, sub-daily FREQs are rejected at save **and** the persist boundary.

  Status key: ✅ ready for merge · 🚧 in progress · ⏸ on waypoint, awaiting slice · ⬜ not started · 🔭 newly scoped. Donetick assignee-rotation is out of scope
  (SP recurrence is single-assignee).


beerkumquatpome:
1. Yearly heatmap: Could you add spaces between each month?
2. Could you add a monthly heatmap too?
github
github2

omega-tree:
ahh very nice! yes thank you for that example. will attempt to mimic

johannesjo:
Reviewed — this is close to merge. The four code changes land exactly the follow-ups from the #7948 thread, and the flag design (localStorage, plain function so pure selector projectors can call it, off-default, fail-safe, `rrule` checked before any storage read) is right.

**Verification note:** CI runs no tests/lint for PRs into `feat/rrule-epic` (`ci.yml` triggers only on PRs to master/main — I'll fix that on my side before #8231 merges). So I ran it locally: all five touched spec files green (3 + 7 + 26 + 54 + 88), eslint clean on every modified file, plus two independent deep-review passes. Findings:

### 🟠 1. The spec-flag hooks are suite-global, not file-scoped

`get-next-repeat-occurrence.util.spec.ts:10`, `get-repeat-occurrence-rrule.spec.ts:10`, `get-repeat-occurrence-rrule.complex.spec.ts:10`, `rrule-occurrence.day-march.spec.ts:10` register

```ts
beforeEach(() => setRRuleEngineEnabled(true));
afterEach(() => setRRuleEngineEnabled(false));
```

at module top level. In Jasmine that attaches to the **topSuite** (verified against jasmine-core source), so in the full Karma run these wrap **every spec in the entire ~9100-test suite**, not "the suite" in the file. Consequences: the whole suite effectively runs flag-ON, so the default (flag-off, production) state is never exercised suite-wide; and specs behave differently under `npm run test:file` vs the full run (the dialog specs use the real `RRuleFeatureFlagService`, no mock). Currently benign — no other spec feeds rrule-bearing cfgs into the routing utils — but it's invisible cross-file coupling: deleting one of these four files could break specs elsewhere. Please move each pair inside the file's outer `describe` (the nested flag-OFF `describe` in `get-repeat-occurrence-rrule.spec.ts` already shows the right pattern).

### 🟠 2. Save-time probe has no never-fire pre-screen

`dialog-edit-task-repeat-cfg.component.ts` (save path, ~`:462`) calls `getFirstRRuleOccurrence({ rrule, startDate })` directly, so a never-firing rule that `_canNeverFire` would reject instantly never gets the chance: with the flag on, typing `FREQ=DAILY;BYMONTH=13` into the raw override and hitting Save reproduces a **~4 s main-thread hang** (measured) before the "never fires" error shows. Gating the probe behind `isRRuleValid(working.rrule)` first fixes it and reuses the validity memo.

### 🟡 3. `_canNeverFire` misses BYWEEKNO/BYYEARDAY × BYMONTH contradictions

`FREQ=DAILY;BYWEEKNO=53;BYMONTH=2` passes the pre-screen (the combo check only pairs `byMonth` × positive `bymonthday`), so the probe walks **~7 s** (one-time, memoised) before returning false. Conservative-by-construction is the right call — confirmed no false positives exist, including the BYSETPOS edge (rrule.js's own constructor rejects |BYSETPOS| > 366) — so I'm fine leaving the exotic tail, but: (a) the spec comment claiming "the bounded `.between()` probe returns false quickly" is wrong on both counts (impl is `_canNeverFire` + an **unbounded** `.after()`), and (b) extending the combo check to `byweekno`/`byyearday` would close the realistic remainder. At minimum fix the comment and add a watchlist §5 entry.

### 🟡 4. The "returns quickly" timing test doesn't exercise the probe path

`rrule-occurrence.complex.spec.ts` — `FREQ=DAILY;BYMONTH=13;BYMONTHDAY=2` is rejected by `_canNeverFire` in O(1) before any probe, so the `<1500 ms` assertion times the fast-reject path and can never catch a probe regression (and is a latent flake under loaded CI). Either drop the timing assertion or test a rule that actually reaches the probe.

### 🟡 5. Stale `includeRRule` on the async edit path

When the dialog opens via `task`/`repeatCfgId` (the common path — cfg fetched async in the constructor effect), `includeRRule` is captured in the `buildOptions` closure **before** the loaded completion-cfg is migrated to `quickSetting:'RRULE'`. On a flag-off device the quick-setting select then has value `'RRULE'` with no matching option (cosmetic — `isRRuleMode()` is independent, so the builder still renders and saves correctly). The two new tests use the synchronous `{ repeatCfg }` path, so they don't cover this. Recompute the options after the async load, or derive `includeRRule` reactively.

### nit 6. `includeRRule = true` default

In `build-repeat-quick-setting-options.ts` a future caller that forgets the new param silently shows the RRULE option on flag-off devices. Defaulting to `false` with explicit opt-in at the two call sites fails safe.

---

**Verified solid** (so it doesn't get re-litigated later): the flag never reaches the sync wire (`config-page` early-returns before any dispatch; `rruleEngine` exists only on `GlobalConfigFormSectionKey`); the completion-cfg migration always yields a parseable rule, only persists on a real schedule edit (initial state carries it too, so the diff stays empty otherwise), and `'RRULE'` clamps to `'CUSTOM'` via `_toPersisted` on all three action creators; UNTIL/COUNT stripping in the validity probe is sound (the engine path keeps and applies the real end conditions); `_parseStart`'s epoch anchor is deterministic; no rule bodies or task content in any new log call. The watchlist doc is genuinely useful — thanks for distilling it.

**Process:** merging this closes the PR, so the phase table moves to the standing draft `feat/rrule-epic → master` PR per the roadmap — I'll open that, reopen #4020 from it, and add `feat/rrule-epic` to `ci.yml`'s PR triggers so phase PRs get real CI.

**Verdict:** please fix 1 + 2 in this PR (both are small); 3–6 can ride along here or land as a fast-follow. Then this is good to merge into the integration branch.

omega-tree:
thanks for the thorough pass! all of it should be in now , full suite run
  locally: 10937 green.

  1. hooks moved inside each file's describe (day-march too). also did the same on the p2 branch —
  same commit content, so the merge later won't conflict.

  2. this one I couldn't reproduce as described — the save path already hits the `isRRuleValid` gate
  at :410 before the :462 probe, and `_canNeverFire` rejects `FREQ=DAILY;BYMONTH=13` in O(1), so that
  rule errors out instantly with "invalid". I think what you measured was actually the class from your
  point 3: something like `BYWEEKNO=53;BYMONTH=2` passes the pre-screen and walks ~10s *inside*
  isRRuleValid on the save click (I measured 10.4s raw). that's gone now via the pre-screen extension
  below, and I added a save-path spec that pins the never-firing raw override rejecting fast with no
  probe walk. if you can still repro a hang at :462 after this, tell me the exact rule and I'll dig
  again.

  3. extended `_canNeverFire` to BYMONTH × BYYEARDAY / BYWEEKNO: month-possibility maps for positive
  yeardays (both leap layouts) and a conservative day-span superset for positive week numbers (covers
  WKST shifts and year spill — week 1 can include late december, week 53 early january; checked
  against rrule.js, which does use ISO week-year semantics there, so those edge combos stay valid).
  negative values count from year end and skip the check. no-false-positive cases are pinned by spec.
  wrong spec comment fixed, watchlist §5 entry added.

  4. dropped the misleading timing assertion — it only ever timed the O(1) reject. replaced with a
  pre-screen timing test (generous bound, regression canary only) plus the satisfiable-combo specs.

  5. includeRRule is now read live per formly evaluation and keyed into the options memo, so the async
  edit-path migration rebuilds the list. spec covers the flag-off async path.

  6. default flipped to false, both call sites pass explicitly, pinned by spec.

johannesjo:
Thank you very much! <3

johannesjo:
Re-reviewed the four follow-up commits (`f5e26106..989409061`). All six findings are addressed — good to merge into `feat/rrule-epic`.

**1 — flag hooks** ✅ moved inside each `describe` (day-march included), so default-off state is no longer wrapped by a topSuite hook.

**2 — save-click freeze** ✅ your analysis is right and resolves my finding: the save path gates on `isRRuleValid` at `:410` before the `:462` probe, and `BYMONTH=13` is O(1), so the ~10s I hit was the **BYWEEKNO=53;BYMONTH=2** class from #3 walking *inside* `isRRuleValid`. Killing that walk closes #2; the new save-path spec pins fast rejection with no probe + `addTaskRepeatCfgToTask` not called.

**3 — `_canNeverFire` combo coverage** ✅ — and since a false positive here would silently drop a valid rule to the legacy engine, I checked the math, not just the comment:
- `_monthsOfYearDay(d)` unions leap + non-leap layouts → the **exact** month set for a positive year-day. Fire ⇒ month in set ⇒ not flagged. No false positive.
- `_monthsOfWeekNo(n)` scans `[base-9, base+16]`. True ISO (WKST=MO) span is `[base-2, base+10]`; a non-Monday WKST shifts ≤6 days ⇒ worst-case `[base-8, base+16]`. Your range is a strict superset, and a superset can only *add* candidate months ⇒ no false positive. Year-boundary spill (`<1→Dec`, `>366→Jan`) keeps week 1/53 valid against Dec/Jan.
- Negatives correctly skip both checks; the no-false-positive spec pins satisfiable combos (incl. `BYYEARDAY=-320`/`BYWEEKNO=-46` landing in Feb). The only residual is a false *negative* (unrecognized never-firing rule still hits the one-time memoised probe) — harmless, and documented in watchlist §5.

**4 — timing test** ✅ replaced with a real pre-screen test + satisfiable-combo specs.

**5 — stale `includeRRule`** ✅ read live per formly eval and keyed into the options memo; the `Subject`-driven async-migration spec reproduces the original bug directly rather than asserting around it.

**6 — default flip** ✅ `includeRRule = false`, both call sites explicit, pinned by spec.

**Two minor, non-blocking:** the two `performance.now() … toBeLessThan(1000)` assertions are now defensible (the paths are genuinely O(1), so 1000ms is a ~200× margin) but remain wall-clock — fine as labeled regression canaries, just the one non-deterministic thing in the range. And the `buildRepeatQuickSettingOptions` default-flip only affects hypothetical future callers since both current sites pass explicitly — exactly the intent.

Verdict: **ready to merge.** I'll open the standing `feat/rrule-epic → master` PR, reopen #4020 from it, and add `feat/rrule-epic` to `ci.yml`'s PR triggers so phase PRs get real CI.

johannesjo:
I did another deep pass over the epic plan with one question in mind: is this the right track, and where can we cut complexity? Short version: RRULE as the target is right, and the Phase 1 shape (additive optional `rrule`, legacy fields kept populated as the wire format, the `'RRULE'→'CUSTOM'` clamp) is exactly the correct move for our typia/old-client constraints — no notes there. Everything below is about shrinking the risk window and writing down a few decisions we've so far left implicit. Roughly prioritized.

## 1. The real risk model: engine divergence means duplicate tasks, not shifted dates

This reframes how I've been talking about the converter divergences (WEEKLY interval≥2 etc.), and I think it should drive both the test priorities and the migration policy.

Occurrence creation is per-device recomputation, not a replayed derivation: each device independently computes "what's due" and emits concrete ops, and the dedupe id embeds the computed day (`get-repeatable-task-id.util.ts` → `rpt_<cfgId>_<dueDay>`). So when device A routes through the RRULE engine and says *Friday*, while device B (old client, or same build with the flag off) routes legacy and says *Saturday*:

- different `dueDay` → different deterministic ids → **both instances are created and sync to every device**, and
- A's `lastTaskCreationDay` update then feeds B's legacy engine a foreign anchor → flip-flopping creation/suppression between two incompatible streams.

So "divergences are bounded and RRULE is the more-correct side" is true as engine math but wrong as a risk statement: every divergent cfg on a multi-device account is a duplicate-and-suppression generator for the entire mixed window. Nothing here is broken today (the flag is off by default), but it's the failure mode the whole plan needs to be shaped around.

## 2. The flag gates evaluation, so we need the endgame written down

`isRRuleEngineEnabled()` sits inside all three calculators (`get-next-repeat-occurrence.util.ts` and friends), not just the authoring UI. That's fine for opt-in testing, but it means two fully-updated devices on one account diverge per #1 above. And it has a sharper edge: **COUNT/UNTIL have no legacy representation at all** (`rruleToLegacyTaskRepeatCfg` maps only cycle/interval/weekdays/monthly anchors), so a flag-on device ends a `COUNT=10` series while every flag-off device and old client resurrects it forever. Same for seasonal `BYMONTH` windows — they fire year-round on the legacy side.

What I'd like us to write into the roadmap as the explicit dual-engine endgame:

- Lazy-migrate on edit (what the dialog already does — and I think it's the right shape: the migration rides a user-intent op, and the Phase 2 preview shows the user the new schedule at exactly the moment semantics could shift).
- When the flag defaults on: a one-time data-repair backfill of `rrule` for **only the provably-lossless class** (interval=1 — empirically exact and DST-proof). Never bulk-convert the divergent classes; they lazy-migrate or age out.
- Then: flag default-on → soak → flag removed → legacy *engine* deleted one release later. Legacy *field* dual-write stays indefinitely (it's the wire format for old clients — cheap, and separate from the engine question).
- Decide now what an invalid `rrule` falls back to once the legacy engine is gone. Today the fallback *is* the legacy fields; post-deletion the honest answer is probably "pause + repair prompt" rather than silently rescheduling.

Without a written endgame, "temporary" second engines default to immortal.

## 3. A decided policy for rules the legacy fields can't represent

Today every save writes a nearest-pattern legacy approximation. For rules outside legacy expressiveness (COUNT/UNTIL, BYWEEKNO, BYSETPOS tricks, seasonal BYMONTH — exactly the differentiators), the approximation is necessarily wrong, and old/flag-off clients will **create tasks on wrong days, which then sync back to all devices**. That's worse than missing occurrences.

Two coherent options: keep the nearest approximation (wrong-day tasks), or write a never-fires legacy sentinel for non-representable rules + a save-time warning ("older devices on this account won't create these tasks"). I lean to the sentinel — fabricated tasks are worse than absent ones — but either way it should be a documented contract, not an emergent property of the converter.

## 4. Tests as the actual merge/flip gate

I'd gate the flag-default-on (and eventually deleting the legacy engine) on these, in this order:

1. **Mixed-version convergence simulation** — two simulated devices sharing an op stream, one routing RRULE and one legacy, driven through day rollovers; assert no duplicate `rpt_<cfgId>_<day>` instances and no `lastTaskCreationDay` flip-flop suppression. Per #1 this is the production failure mode, so it outranks everything else.
2. **Differential fuzz spec** over the legacy cfg space (cycle × interval 1–6 × weekday masks incl. empty × monthly anchors × day-29/30/31 + Feb-29 starts), comparing multi-year streams legacy vs converted-RRULE under both CI timezones — with the known divergences **pinned as expected-output cases**, not excluded. The divergence direction is a contract we're consciously shipping; an exclusion list silently grows. Important detail: run it against all three calculators, especially `getNewestPossibleDueDate` — that's what task creation actually consumes, so equality of `getNextRepeatOccurrence` alone doesn't prove the creation path converges.
3. **typia wire round-trip per producer path** (builder save, converter, backfill, later REST/repair) against the *released* schema — guards the catastrophic mode (out-of-union value → old-client data-repair, no rollback) that no differential test touches.

## 5. Shrinking "whole"

Looking at the phase table again, I want to revisit my own "merge only when whole" framing — 13 phases makes "whole" unreachable, and a good chunk of the list is features riding on RRULE rather than the model refactor itself:

- **Core (gates the merge):** P1 ✅, the test suite above, the preview half of P2 (it's migration UX — a user whose semantics could shift needs to see the new schedule), and P8 (RECURRENCE-ID/EXDATE completes the occurrence model and formalizes `deletedInstanceDates`; CalDAV #8148 already wants it).
- **Ordinary post-merge feature PRs:** P3 (`@+` natural language — pure entry UX, zero model coupling), P4, P6, P7, P9.
- **P5 (ends-after-N)** hides new sync infrastructure — its own footnote says it needs a min-client-version gate, which doesn't exist. That alone moves it out of the epic.
- **P10/P11/P13** aren't RRULE features — adaptive and trigger-based are *alternative recurrence semantics* (a different effort), multi-reminders is the reminders domain.
- **P12 (sub-daily)** I'd spin out entirely: the engine, the task identity scheme (`rpt_<cfgId>_<YYYY-MM-DD>`), `dueDay`, and TODAY-tag membership are all structurally day-granular. Sub-daily invalidates that invariant everywhere at once — it's a second epic wearing a phase number. The existing save/persist-boundary rejection is exactly right; let's keep it at that.

None of this throws work away — it just means the epic's master-merge gate is "model refactor proven equivalent + previewable," and everything else lands afterwards through the normal flow.

## 6. Smaller concrete items on the branch

- **The live preview can still freeze the UI pre-save:** `rrulePreview` → `getRRulePreview` → `rule.after()` runs with no `isRRuleValid`/never-fire gate and recomputes on every builder keystroke (`dialog-edit-task-repeat-cfg.component.ts` ~168, `rrule-preview.util.ts` `_getUpcoming`). `FREQ=DAILY;BYWEEKNO=53;BYMONTH=2` reproduces the same multi-second main-thread walk the save path now guards against.
- **Replace the `_canNeverFire` heuristics with a bounded probe.** The leap-year/week-53 pre-screen math in `rrule-occurrence.util.ts` is a moving target against a combinatorial grammar, and by its own comment unrecognized never-fire rules still hit a one-time multi-second walk. A capped `between(dtstart, dtstart + 10y)` probe — "a valid rule must fire within a decade" — kills the freeze class structurally, deletes the heuristics, and is a defensible product rule. One shared gate also covers future ingestion paths (REST) for free. It would fix the preview bullet above too.
- **The stale-anchor comment's premise broke when the flag landed.** `task-repeat-cfg.model.ts` (~129) argues the stale `monthlyWeekOfMonth` on remote clients is "inert once `rrule` is set, since the occurrence engine routes on it" — that was true pre-flag, but with flag-off as the default every remote client routes legacy, where the stale anchor is live. The documented PRE-rrule-only gap currently applies to all clients. At minimum the comment needs updating; worth a think whether the anchor-switch path needs more than that before the flag flips.
- **Pin `rrule` to an exact version** (it's `^2.8.1`). Occurrence streams are downstream of a third-party parser with known quirks; a caret upgrade that changes parsing on some devices mid-account re-creates the #1 duplicate mechanics *between identical app versions*. The invariants spec then doubles as the upgrade tripwire.

---

Again — direction is right, Phase 1's data-model instincts are exactly right, and the empirical converter work is what makes any of this plannable at all. This is about making the dual-engine window short and explicit instead of open-ended, and pointing the test effort at the failure mode that actually hurts (duplicates across devices) rather than the one that's easy to state (shifted dates).

PR #8227, opened 2026-06-09
Fix minor cal design

hajiboy95:
## Problem

Adressing the [comment of a recent PR](https://github.com/super-productivity/super-productivity/pull/8017#issuecomment-4661441890)

The date picker (based on Angular Material's mat-calendar) had several minor but noticeable design and UX inconsistencies:
   1. Selection Highlight Loss: Navigating years or months would cause the actual selected date to lose its primary color highlight, as the
      selected property was being incorrectly tied to the "active" (focused) date in those views.
   2. Sticky Focus Outline: Clicking a date with the mouse would leave a persistent focus outline (circle) that wouldn't disappear until focus
      was explicitly moved, creating a "sticky" visual feeling.
   3. Missing Initial Focus: The focus outline was not visible when the calendar first opened or when switching views, making it unclear which
      field was active for keyboard users.
   4. Keyboard Navigation Skipping: Quickly moving focus after a selection would sometimes "skip" fields because a background state sync was
      force-resetting the calendar's activeDate while the user was interacting.

## Solution

I performed a surgical refactoring of the DateTimePickerComponent and refined the global Material overwrites:
   1. Selection Logic: Removed custom pickerSelectedDate logic and bound mat-calendar directly to the selectedDate signal. This ensures the
      selected date remains highlighted regardless of view navigation.
   2. Focus Management:
      - Introduced an isInitialFocus state to show the focus outline on open/view-switch.
      - Implemented a hybrid suppression logic: focus outlines are suppressed when the mouse is used (preventing "stickiness") but remain fully
        visible for keyboard navigation and initial states.
      - Reverted global :focus-visible to :focus but scoped its suppression within the component to allow for programmatic focus visibility.
   3. Navigation Fix: Updated the activeDate sync effect to skip updates if the calendar currently has focus. This prevents background
      synchronization from "fighting" with active keyboard navigation.

## Type of Change

   - [x] Bug fix
   - [ ] New feature
   - [ ] Breaking change
   - [ ] Refactoring
   - [ ] Documentation
   - [x] Style/UX Improvement

## Checklist

   - [x] I have run npm run checkFile on changed .ts/.scss files.
   - [x] I have added tests for my changes.
   - [x] My commit messages follow the Angular format (type(scope): description).

### DIFF
--- a/src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.spec.ts
+++ b/src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.spec.ts
@@ -37,6 +37,11 @@ describe('DialogEditTaskRepeatCfgComponent', () => {
   let mockDateTimeFormatService: jasmine.SpyObj<DateTimeFormatService>;
   let mockDateService: jasmine.SpyObj<DateService>;
 
+  // DateService is mocked to this fixed day; assertions about "today" must
+  // derive from these consts, never from the real clock (see #8017 CI breakage).
+  const MOCK_TODAY = new Date(2026, 5, 9, 0, 0, 0, 0);
+  const MOCK_TODAY_STR = '2026-06-09';
+
   const mockRepeatCfg: TaskRepeatCfg = {
     ...DEFAULT_TASK_REPEAT_CFG,
     id: 'repeat-cfg-123',
@@ -83,8 +88,8 @@ describe('DialogEditTaskRepeatCfgComponent', () => {
       'todayStr',
       'getLogicalTodayDate',
     ]);
-    mockDateService.todayStr.and.returnValue('2026-06-09');
-    mockDateService.getLogicalTodayDate.and.returnValue(new Date(2026, 5, 9, 0, 0, 0, 0));
+    mockDateService.todayStr.and.returnValue(MOCK_TODAY_STR);
+    mockDateService.getLogicalTodayDate.and.returnValue(new Date(MOCK_TODAY));
 
     // Set up the return value for getTaskRepeatCfgById$ before creating the component
     if (getTaskRepeatCfgById$ReturnValue) {
@@ -303,8 +308,9 @@ describe('DialogEditTaskRepeatCfgComponent', () => {
         (c) => c.key === T.F.TASK_REPEAT.F.Q_MONTHLY_CURRENT_DATE,
       );
 
-      const today = mockDateService.getLogicalTodayDate();
-      const todayDayStr = today.toLocaleDateString('en-US', { day: 'numeric' });
+      // "today" comes from the mocked DateService.getLogicalTodayDate (2026-06-09),
+      // not the wall clock — asserting against new Date() breaks on any other day
+      const todayDayStr = MOCK_TODAY.toLocaleDateString('en-US', { day: 'numeric' });
 
       expect(monthlyCall).toBeDefined();
       expect(monthlyCall!.params.dateDayStr).toBe(todayDayStr);
@@ -379,11 +385,8 @@ describe('DialogEditTaskRepeatCfgComponent', () => {
     });
 
     it('should preserve WEEKLY_CURRENT_WEEKDAY when startDate weekday differs from today', async () => {
-      // Pick a date whose weekday definitely differs from today
-      const today = mockDateService.getLogicalTodayDate();
-      const differentDay = new Date(today);
-      differentDay.setDate(today.getDate() + 3); // 3 days from now is a different weekday
-      const dateStr = differentDay.toISOString().slice(0, 10);
+      // Pick a date whose weekday definitely differs from the mocked today
+      const dateStr = '2026-06-12'; // Friday; MOCK_TODAY is a Tuesday
 
       const cfgWeekly: TaskRepeatCfg = {
         ...DEFAULT_TASK_REPEAT_CFG,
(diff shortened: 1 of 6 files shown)


johannesjo:
Thanks for the careful work here — the core fix (binding `[selected]` directly to `selectedDate()`) is the right call, and the new tests are appreciated. A few things to address before merging:

### Bugs

1. **`isInitialFocus` "view-switch" behavior isn't actually implemented.** The PR description says: _"Introduced an isInitialFocus state to show the focus outline on open/view-switch."_ But in `datetime-picker.component.ts`, `isInitialFocus` is only ever set to `false` (on first keydown at `:139`, on first mousemove at `:251`). Nothing flips it back to `true` when the user clicks the period button to switch month → year → multi-year. Either subscribe to `mat-calendar.stateChanges` / `viewChanged` and reset it on view change, or drop the "view-switch" claim from the description.

2. **`_lastSyncedDate` gets updated even when the sync is skipped.** In `_syncActiveDateEffect` (`datetime-picker.component.ts:115-130`), `this._lastSyncedDate = dateMs` runs _before_ the focus guard. If the calendar is focused, the active-date assignment is skipped but `_lastSyncedDate` is still marked as up-to-date — so the effect won't retry the sync when focus later leaves the calendar (the effect has no signal dep on `document.activeElement`). Move the `_lastSyncedDate = dateMs` assignment _inside_ the `if (!isCalendarFocused)` branch so a future re-run can re-attempt.

### UX / styling

3. **Focus and hover share the same outline + background.** In `_overwrite-material.scss:489-504`, the `:hover` and `:focus` selectors are grouped and produce identical visuals (same accent outline, same `--calendar-hover-bg`). This is confusing — a keyboard user moving focus and a mouse user idly hovering get the same feedback, and there's no way to tell which cell is actually focused when the mouse happens to be over another cell. Worth splitting them: e.g. solid outline for focus, dashed or background-only for hover (or vice versa). Out of scope for this PR's stated goal, but since this PR is reshaping focus/hover behavior it'd be a natural place to fix it.

### Nits

4. **Empty constructor.** `datetime-picker.component.ts:79` — `constructor() {}` is vestigial after removing the old `effect()`. Delete it.

5. **Tests verify field mutation, not rendered behavior.** `datetime-picker.component.spec.ts:123-138` mutates `component.isInitialFocus` directly and checks the field value back. It doesn't verify the `.sp-initial-focus` host class actually toggles, nor that the SCSS suppression kicks in. Asserting against `fixture.nativeElement.classList` after `detectChanges()` would catch a future regression where someone removes the `@HostBinding`.

### Out of scope but worth a thought

6. Angular Material exposes calendar design tokens (`--mat-datepicker-calendar-date-focus-state-background-color`, `--mat-datepicker-calendar-date-hover-state-background-color`, etc.). Setting those at `:root` would let us drop most of the `!important` selector wars in `_overwrite-material.scss` over time. Not for this PR, but flagging for follow-up.

hajiboy95:
@johannesjo comments are adressed. 

@beerkumquatpome that's not a bug, instead it's intended. I was very well aware of that behaviour. That's the natural behaviour of that modal, which we want to stick to. we don't want to implement new focus behaviour since we'd then need to ensure that it works with each and every update of the packages we use. For better differentiation @johannesjo suggested that we differentiate the two different focus effects (which is done now, I went for the dashed version, thanks for that suggestion, looks nice). I can understand your desire to make it as intuitive as possible @beerkumquatpome but sometimes it is what it is haha. I also think it's not that severe since people usually use either mouse, or keyboard, not both.

johannesjo (commented):
Nice work — this addresses each of the listed bugs coherently and the new tests cover the new behaviors well. A few small nits before merge:

**1. Lost `readonly` on `_cdr`**
`datetime-picker.component.ts:74` — `private readonly _cdr = inject(ChangeDetectorRef);` became `private _cdr = inject(ChangeDetectorRef);`. Looks like an inadvertent edit while removing the constructor; please restore `readonly`.

**2. Reuse `MatCalendarView`**
`datetime-picker.component.ts:108` — `private _lastView: 'month' | 'year' | 'multi-year' | null` duplicates Angular Material's `MatCalendarView` type. Importing it would future-proof the type if Material ever adds a view.

```ts
import { MatCalendarView } from '@angular/material/datepicker';
private _lastView: MatCalendarView | null = null;
```

**3. Minor: redundant `:focus:not(:hover)` rules in `datetime-picker.component.scss`**
The two rules at lines 21–31 differ only in whether the cell is `.mat-calendar-body-selected` (the second one just drops the `background-color: transparent`). Could collapse to one `outline: none` rule plus a narrower `background-color` rule:

```scss
.mat-calendar-body-cell:focus:not(:hover) > .mat-calendar-body-cell-content {
  outline: none !important;
}
.mat-calendar-body-cell:focus:not(:hover)
  > .mat-calendar-body-cell-content:not(.mat-calendar-body-selected) {
  background-color: transparent !important;
}
```

**Note (not a change request):** while the calendar has focus, `_syncActiveDateEffect` intentionally skips updating `_lastSyncedDate`, so a later same-value re-set will still sync `activeDate`. Since the `[selected]` binding is now driven directly by the input signal, the visible "selected" marker is unaffected — looks intentional and matches the PR description.

LGTM with the `readonly` revert; the other two are nits.

johannesjo:
Thank you very much! <3

johannesjo:
> Do we still need to have 2 different outlines?
> Solid outline is now hidden when moving the mouse.
> I think they should use the same outline (dashed outline) because of this change

I agree, that it is a bit odd at the moment. We should however use different styles for hover and focus, as otherwise indication is not always clear (e.g. 2 days can have the same outline style). We now use background and border to indicate the same status. Maybe we can split this up or combine this, so it looks better?

hajiboy95:
@beerkumquatpome here I needed to make a decision which one wins. I decided to let to "full" version win. Another approach would be to always make the newest version win. It's a matter of taste I guess. But at least it's now like it's intended to be :D

johannesjo:
> The user can only see one outline

No, you can focus one with they keyboard and then hover. Pair that with a brief pause and you got a (slightly) confused user.

PR #8229, opened 2026-06-09
fix(planner): don't double-count done recurring tasks in remaining time

### LINKED ISSUE #8220: 🚨 Estimate remaining time in "Today" and "Planner" calculated differently for recurrent tasks
### Version Used

18.9.1WB (firefox desktop) and 18.8.0_L2AF (f-droid)

### Current Behavior

When I have a recurrent task and mark it as done without adding time for it (e.g. because the meeting didn't take place) its estimated time is still added to the remaining time in the "Planner" tab, but substracted in the "Today" tab.

Today tab:

Image

Planner tab:

Image

### Expected Behavior

When a task is marked as done, the time should not be used to calculate the total remaining time in the "Planner" tab.

### Steps to Reproduce

Add a recurring task to today with a fixed time (e.g. 1h) and mark it as done without tracking time for it. The remaining time in "today" and "planner" is different.

### Console Output

```shell

```

### Documentation

costajohnt:
## Problem

Closes #8220.

When a recurring task is marked done without any time tracked (for example, a recurring meeting that didn't take place), its estimated time still counts toward the remaining time in the **Planner** tab, while the **Today** tab correctly excludes it.

Root cause is a double count in `getPlannerDay` (`planner.selectors.ts`). For a recurring task the Planner sums:

1. the real task instance, which is done-aware and contributes 0 once the task is done, and
2. a separate repeat *projection* built from the `TaskRepeatCfg`, which always uses the full estimate with no done-awareness.

So a done recurring task ends up as `0` (instance) + `fullEstimate` (projection). The Today view never builds projections, so it only ever counts the real task. That is why the two views disagree.

## Solution

In `getPlannerDay`, drop the repeat projection for any config that already has a real instance in that day, deduping by `repeatCfgId`, so the recurring task is counted once through its done-aware instance. This is applied to both the timed (`repeatProjectionsForDay`) and untimed (`noStartTimeRepeatProjections`) projection arrays, so the remaining-time total, the item count, and the scheduled item list all stop double-counting in one place.

This mirrors the dedup the Schedule view already does for the same situation (#7853, see `create-schedule-days.ts`). Per-day dedup by bare `repeatCfgId` is sufficient here because `getPlannerDay` builds a fresh set per day, whereas the Schedule variant keys by `repeatCfgId|day` since it processes many days in one pass.

Added a regression spec (`planner.bug-8220.spec.ts`) covering the done instance, the undone instance, the timed projection, the untimed projection, and the no-instance case (which should still project).

## Type of Change

- [x] Bug fix

## Checklist

### DIFF
--- a/src/app/features/planner/store/planner.bug-8220.spec.ts
+++ b/src/app/features/planner/store/planner.bug-8220.spec.ts
@@ -1,0 +1,177 @@
+import * as fromSelectors from './planner.selectors';
+import { PlannerState } from './planner.reducer';
+import { Task } from '../../tasks/task.model';
+import { getDbDateStr } from '../../../util/get-db-date-str';
+import {
+  DEFAULT_TASK_REPEAT_CFG,
+  TaskRepeatCfg,
+} from '../../task-repeat-cfg/task-repeat-cfg.model';
+
+// Regression for #8220: a recurring task that already has an instance in a day
+// (created, and possibly marked done) was counted twice in the Planner: once as
+// the real task instance and again as a repeat projection. The "Today" view
+// (work-context.util) only ever counted the real instance, so the two views
+// disagreed on remaining time. The Planner must defer to the real instance and
+// drop the now-redundant projection.
+describe('Planner Selectors - #8220 recurring done task double-count', () => {
+  const today = getDbDateStr();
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! <3

PR #8231, opened 2026-06-09
feat(task-repeat-cfg): RRULE epic phase 2 — heatmap preview + completion simulation

omega-tree:
Part of #8224

## Phase 2 — heatmap occurrence preview + completion simulation

Second slice of the RRULE epic (see the tracker PR above and [`docs/research/rrule-epic-roadmap.md`](https://github.com/johannesjo/super-productivity/blob/feat/rrule-epic/docs/research/rrule-epic-roadmap.md)).

### What it adds

- **Live calendar preview in the repeat dialog:** while authoring a rule, a togglable 365-day month-calendar shows every projected occurrence — no saved cfg required.
- **Completion simulation:** for "repeat from completion" schedules, click any day to simulate completing there and watch the rest of the series re-anchor. Click again (or edit the schedule) to clear.
- **Repeat-task heatmap:** the per-series activity heatmap gains a projected-occurrences overlay (year strip + month calendar, in-card year navigation, metrics).

### Guard rails

- Everything here is display-only and gated behind the off-by-default per-device RRULE engine flag; with the flag off the heatmap stays the unchanged history-only view.
- The projection only renders when the cfg has a valid `rrule` — mirroring the occurrence-routing gate — so the overlay never shows days the device won't actually fire on.

### Review status

All findings from the review rounds on this branch are addressed, including the root-suite spec-hook leak (flag hooks now scoped to their suites; full unit suite run locally: 10964/10964 green).

Note: the diff currently includes the Phase-1 follow-up commits because `feat/rrule-epic` does not contain them yet; once the pending sync PR (`omega-tree:feat/rrule-epic → feat/rrule-epic`) merges, this PR rebases down to the pure Phase-2 slice.


omega-tree:
thnx mate!

omega-tree:
everything but the scrolling calendar can be fixed.  if i increase the modal size i have to increase all the fields as well which leaves it looking stilted. furthermore it really doesn't translate to mobile.

johannesjo:
Thanks — strong slice! The feedback rounds above are essentially all in, and the Phase‑1 follow-ups check out: the flag stays local-only and is routed around `updateGlobalConfigSection` (no phantom ops), the seeded no-rrule→builder migration can't produce a spurious write (`repeatCfgInitial` stores the processed cfg → empty diff on a no-touch save, and the specs pin that), and the clamp boundary holds (`_toPersisted` / `toSyncSafeQuickSetting` in all three action creators — I found no dispatch path around it). Also nice: the `_previewScheduleCfg` value-equality memo against formly's clone-per-keystroke, and `_canNeverFire` being conservative by construction.

One thing to fix before merge, then a few smaller ones:

### 1. 🟠 The top-level `beforeEach`/`afterEach` in the four occurrence specs are *global* hooks

`get-next-repeat-occurrence.util.spec.ts:10`, `get-repeat-occurrence-rrule.spec.ts:10`, `get-repeat-occurrence-rrule.complex.spec.ts:10` and `rrule-occurrence.day-march.spec.ts:10` register `beforeEach(() => setRRuleEngineEnabled(true))` **outside any `describe`**. In Jasmine that attaches to the root suite, and Karma bundles all specs into one env — so the flag is forced ON around every one of the ~9100 tests in the app. Verified empirically: a canary spec asserting `localStorage.getItem('sp_rrule_engine_enabled') === null` passes alone and fails the moment it's bundled with one of these files (`Expected 'true' to be null`).

Nothing fails *today*, but only by luck: pre-existing repeat specs use no-rrule cfgs (your `rrule`-first guard saves them), `rrule-engine-flag.spec.ts` removes the key inside the test body, and your "flag OFF" describe re-overrides with a nested hook. Any future spec that asserts default-off behavior becomes bundle-order-dependent, and this is exactly the cross-file-pollution class that has poisoned Karma runs here before. Fix: move the hooks inside each file's `describe` (day-march needs one wrapping describe).

Related heads-up: **CI runs no unit tests on this PR** — `ci.yml` filters `pull_request` to base `master`/`main`, so only `build-preview` ran. Please run `npm test` locally once after the fix; that applies to every phase PR into `feat/rrule-epic`.

### 2. 🟡 Projection can show days the device won't fire on (legacy WEEKLY, interval ≥ 2)

`repeat-task-heatmap.component.ts` projects `cfg.rrule || legacyTaskRepeatCfgToRRule(cfg)`, but a **no-rrule** cfg still *creates tasks* through the legacy engine even with the flag on (the routing utils correctly gate on `rrule &&` first). The converter is faithful for the monthly clamp cases (👍 the `BYMONTHDAY=d,-1;BYSETPOS=1` idiom) and for interval≥2 DAILY/MONTHLY/YEARLY, but two cases diverge:

- **WEEKLY interval ≥ 2**: legacy counts rolling 7-day blocks from `startDate`; rrule.js groups by WKST weeks. E.g. startDate Tue Jun 9, every-2-weeks Monday → legacy fires Jun 15, the overlay shows Jun 22. (`getAlignedStartDate` documents this; the heatmap conversion doesn't apply it.)
- **Zero weekday flags**: a legacy WEEKLY cfg with no weekday set never fires in the legacy engine, but the converter emits `BYDAY=<start weekday>` → the overlay projects a series that will never be created.

Display-only and flag-gated, so not a blocker — but the overlay's whole point is trust. Simplest: only project when `cfg.rrule` actually exists. Adjacent: the overlay also skips the `isRRuleValid` gate the routing utils have, so a malformed raw-override rrule shows *zero* projected days while legacy keeps firing — same fix territory.

### 3. 🟡 Year nav can strand the user on an empty view

`availableYears` (repeat-task-heatmap) adds a year for **any** `timeSpentOnDay` key, but `_buildHeatmapData` sets `hasData` only for `timeSpent > 0`. A zero-value entry (start/stop tracking instantly) makes its year navigable; navigating there → `heatmapData()` returns null → the `@if` removes the whole switcher *including the ‹ › buttons* — stuck until the dialog is reopened. One-liner: skip `timeSpent === 0` entries in `availableYears` (or keep the nav outside the `@if`).

### 4. 🟡 Duplicate month label in the dialog preview

The 365-day rolling window spans 13 partial months, so the year strip shows two "Jun" blocks from different years with nothing to tell them apart (`buildHeatmapMonths` labels with `monthNames[month]` only — the repeat heatmap avoided this by year-scoping, and the month view shows the year). Cheap fix: append the year to the first block of each new year.

### 5. Minor / follow-up material

- Day cells are clickable `div`s with no keyboard access or role — fine while display-only, but click-to-simulate is now a real interaction. OK as a follow-up, just making it a conscious choice.
- The `canPrevYear/canNextYear/prevYear/nextYear` quartet is duplicated verbatim across `activity-heatmap` and `repeat-task-heatmap` — extract on next touch.
- `heatmap.component.ts` `dayLabels` stays hardcoded English while the month calendar localizes via `DateAdapter.getDayOfWeekNames` — now visibly inconsistent between the two views of the same widget.
- The sim is cleared on rule/schedule-type changes but not on a `startDate`/exdate edit — the same "a sim belongs to the rule it was clicked on" argument applies.
- Stale comment: `getRRuleOccurrencesInRange` still says "No production caller yet" — it has two now.
- PR body: per the roadmap's own flow this should carry `Part of #<epic-PR>`.

Everything else I traced is clean: wire-safety end-to-end (incl. `isExpandActivityForRepeatEdit` as a new optional field), the sim split-window math (first/last-day edges), the month-nav bounds vs noon-anchored ranges, the share-canvas export path (draws from `weeks`/`monthLabels` directly, unaffected by the `months`-only template), and no hot-path involvement. The round-3 cosmetics (centering / December gap / monthly heatmap height) can land here or as a follow-up.

johannesjo:
I rechecked the latest head (`0c3479a1814bf2d737eb080001c63181dc8ffe60`) and still see a few issues:

1. **History heatmap can strand users on an empty current year.**
   `RepeatTaskHeatmapComponent` initializes `selectedYear` to the current year and includes the year nav only inside `@if (heatmapData(); as data)`. But `heatmapData()` returns `null` when the selected year has no tracked/projected data. For a recurring task with only older tracked activity, the current year renders no heatmap and no prev-year controls, so the existing history is unreachable. See `repeat-task-heatmap.component.ts:73-90`, `:130-141`, and `repeat-task-heatmap.component.html:17-30`.

2. **Display-only heatmap cells still behave like clickable cells for mouse users.**
   The ARIA role/tabindex are gated by `interactive`, but the click handlers and pointer/hover affordances are not. In display-only contexts, non-empty cells emit `dayClick` on mouse click while keyboard users cannot activate the same cells. Either gate click/hover/cursor on `interactive`, or make the semantics consistently interactive. See `heatmap.component.html:41-46`, `heatmap.component.scss:127-131`, `:175-179`, `heatmap-month-calendar.component.html:34-39`, and `heatmap-month-calendar.component.scss:85-93`, `:132-134`.

3. **Repeat-from-completion preview labels selectable empty days as activity.**
   The preview makes every in-range day interactive when `repeatFromCompletionDate` is enabled, but non-projected level-0 days fall through to the generic heatmap title, e.g. `0 tasks, 0m`. Since clicking those cells is a valid simulation action, the title/aria-label should describe the simulation action rather than fake activity. See `dialog-edit-task-repeat-cfg.component.html:100-110`, `heatmap.component.ts:115-124`, and `heatmap-month-calendar.component.ts:151-163`.

4. **Sub-daily RRULEs are rejected at dialog save, but not at the persist/ingestion boundary.**
   `save()` rejects `FREQ=HOURLY`, but the persisted action transform only clamps quick settings/monthly anchors, and `isRRuleValid()` does not reject unsupported frequencies. That means an imported/synced sub-daily RRULE can still route through the day-granular engine. This also contradicts the roadmap note saying sub-daily FREQs are rejected at save and persist boundary. See `dialog-edit-task-repeat-cfg.component.ts:604-623`, `task-repeat-cfg.actions.ts:39-46`, `rrule-occurrence.util.ts:222-250`, and `docs/research/rrule-epic-roadmap.md:90-92`.

5. **New task setting is missing from the settings wiki.**
   This PR adds `isExpandActivityForRepeatEdit` to the Tasks settings form and `en.json`, but `docs/wiki/3.02-Settings-and-Preferences.md` is unchanged. The documentation guide calls out new/changed settings as needing that wiki page updated.

johannesjo:
Nice slice — the navigation lands cleanly and the feedback rounds are essentially all in here: `"Jan 2026"` → `"Jan"` for the current year (year strip + month label both), the dead year arrows now drop out when there's only one navigable year, and the unbounded `‹ ›` year nav + boundless month paging work well together.

A couple of things I traced and liked:

- The window-shift math is reversible in both directions with no oscillation — the ~13-month window shifts by a full 12, so the month you cross onto always lands inside the new window, and the offset returns to 0 cleanly on the way back. The input swap can't re-trigger `onPreviewMonthChange` either (it only fires on user prev/next), so there's no feedback loop.
- Destroying the inactive view in the switcher (`@if`/`@else`) resets `_viewMonth`, so the year-offset and month position can't desync across view toggles.
- `occFrom = offset === 0 ? today : from` is a good touch — already-past days stay unmarked only in the home window, and the new spec pins exactly that.

One thing worth a quick decision before merge:

**The home window can still strand — the one empty-window case the nav fix doesn't cover.** `resultHeatmapData()` returns `null` when `!occ.length && !sim && offset === 0`, so a valid rule with no occurrence in the next 365 days (e.g. `FREQ=YEARLY;INTERVAL=2`/`3`, or a far-future start) renders nothing when you toggle the preview on — and because the `‹ ›` arrows live inside the heatmap that didn't render, there's no way to navigate forward to where it *will* fire. It's the same stranding you just fixed for navigated windows, still present at home. Narrow (multi-year intervals, behind the flag), so I'm happy to take it as a follow-up — just flagging the choice: render the empty home window too and let nav take over, or show a "no occurrences in this range" hint instead of a blank.

Minor, non-blocking: after using the year arrows to jump *forward*, switching to the month view lands on the last month of the window (`rangeEnd`) rather than its start, since `today` is no longer in range. Just slightly disorienting; backward jumps don't show it.

Everything else looks solid. **Are you treating this as ready to merge, or do you want to fold the home-window case in first?**

omega-tree:
> Recurring task heatmap: Is it possible to hide the cells for the days that you don't have a recurring task? So it doesn't look like you are breaking the streak
> Remove those cells but don't change the positions of other cells

hmm ok will attempt. happy to see an image as an example.

> Tooltip delays: It's so slow
> They should have the same delays as Github's heatmap

sure easy enough!

omega-tree:
>  Are you treating this as ready to merge, or do you want to fold the home-window case in first? 

good question i guess this heatmap branch would be merged into `feat/rrule-epic` branch when you think it looks good. and on major release you can cherry-pick the implemented and approved phase you want to stop at. I think for an MVP on the next major release we would stop at this phase (p2).  In the meantime i'll continue with the other phases but don't want them in master until we get some real in-field usage. I hope that answers your question. If you want a different approach I'm ok with whatever dev pipeline you decide just let me know.

johannesjo:
## Review — Phase 2 heatmap + completion simulation

Reviewed at head `60262f1` (base `feat/rrule-epic`), with adversarial verification of each finding. Solid slice — the architecture is careful and the prior review rounds clearly landed. Both of @beerkumquatpome's points are addressed:

- **Hide off-schedule cells without shifting others** ✅ — off-schedule, no-time days are dropped from `dayMap` and the grid re-renders them as transparent placeholders in the same slot (genuinely-missed *scheduled* days correctly stay grey; tracked-time-off-schedule days stay visible).
- **Instant tooltips** ✅ — native `title` → `matTooltip` (0ms show) kills the ~1–2s browser delay.

Nothing here is blocking. One small real bug, one a11y gap, and a batch of nits.

### 🐛 Worth fixing — projected "tomorrow" cell loses its dashed styling after noon
`repeat-task-heatmap.component.ts` (~L317-340): `tomorrow` is built as `new Date(now)` + 1 day, so it inherits the current *time-of-day*. Occurrences are anchored at **local noon**, so once the wall-clock passes 12:00, tomorrow's occurrence (`tomorrow 12:00`) fails `occ >= tomorrow` and is not marked `isProjected` — it renders as a plain grey level-0 cell instead of the dashed "upcoming" cell. Verified with a repro: the flip happens exactly at 12:00:00, affects only tomorrow's cell, and self-heals before noon (so the heatmap looks different morning vs. afternoon). The cell doesn't disappear (it's in `scheduledDays`), so it's cosmetic.
```ts
// compare against start-of-tomorrow, not now+24h
const startOfTomorrow = new Date(now.getFullYear(), now.getMonth(), now.getDate() + 1);
// ... occ >= startOfTomorrow
```
This path has no test — a small `isProjected` assertion (with a mocked clock set to afternoon) would lock it down.

### ♿ a11y — recurrence builder controls have no accessible name
`rrule-builder.component.html`: the native `<select>`/`<input>` and the weekday/month toggle buttons have no `for`/`id` pairing and no `aria-label`/`aria-labelledby` — the `<label>`s are bare siblings. A screen reader announces the freq/interval/by-day controls with no field name, and the weekday toggles announce only their 2-char text ("Mo") since the full name is in `[title]` (not a reliable accessible name). Suggest associating each label with its control (or `aria-label`), and `[attr.aria-label]="w.full"` on the day/month toggles.

### 🧹 Cleanups
- **Dead method** `onYearChange` (`activity-heatmap.component.ts:69`) — orphaned by the year-`<select>` → prev/next refactor; it's `public`, so lint won't flag it. Safe to delete.
- **`$any` inconsistency** — `rrule-builder.component.html:109,174` use `$any($event.target).value` while the other 14 handlers use a typed `#ref`. Swap to a `#ref` for parity/type-checking.
- **`'1970-01-01'` fallback** appears as a bare literal in 5 occurrence/converter files (4 pre-existing; this PR adds the one in `_parseStart` with a good rationale). Optional: export one shared `EPOCH_FALLBACK_DATE`.

### 🧪 Test gaps (all minor)
- No engine-level test for `repeatFromCompletionDate: true` **with an rrule** — the from-completion path and the rrule path are each tested, but not combined (the riskiest dtstart-re-anchoring case).
- The `_parseStart` epoch-fallback determinism fix (`new Date()` → `1970-01-01`) is the whole point of that change but isn't directly asserted — a couple of `startDate: undefined`/invalid cases would guard it.
- `repeat-task-heatmap.component.spec.ts:100-101` hardcodes `[2024, 2023]` + `.not.toContain(currentYear)` without mocking the clock — latent flake if ever run with a clock set to 2023/24; prefer `jasmine.clock().mockDate(...)` (as the month-calendar spec already does).
- `getCellTitle`/`getDayTitle` tooltip strings, and `RRuleFeatureFlagService` (the wrapper around the well-tested flag util), have no direct specs.
- Defense-in-depth: `rrule-engine-flag.spec.ts` has no `beforeEach` reset and `rrule-builder.component.spec.ts` no describe-scoped `afterEach(() => setRRuleEngineEnabled(false))` — no leak today (try/finally + inline cleanup), but a top-level reset would harden the shared Karma run.

### 📝 Docs
`docs/research/rrule-epic-review-watchlist.md:201-202` still lists "Effects re-anchor untested — Fast-follow", but this PR adds exactly that test (`task-repeat-cfg.effects.spec.ts` → `rescheduleTaskOnRepeatCfgUpdate$` "re-anchors on an rrule-only change"). Worth striking/downgrading that bullet.

### 🧭 Rollout note (not a change request)
Because `isRRuleEngineEnabled()` is per-device (localStorage), two synced devices with different flag states route the same cfg through different engines. I verified this is **not** a replay-determinism problem (the routing utils are never imported by a reducer/meta-reducer) and the worst case is a bounded, self-dedeuping extra instance for rare advanced configs (multi-weekday `INTERVAL≥2`, out-of-union monthly ordinals like `5MO`/`-2MO`, month-end clamp). Fine for an off-by-default experimental flag — just worth recording as an acceptance criterion before the flag is ever defaulted-on or made synced.

### ⚡ Perf (acceptable, just flagging)
`matTooltip` now instantiates on every cell (~hundreds per year view) vs. the old zero-cost `title`. Both surfaces are the metrics page and the edit dialog (not the `task.component` per-row hot path) and all components are OnPush, so this is fine — optionally `matTooltipDisabled` on empty cells if you want to trim it.

### ✔️ Verified clean
Per-device flag is genuinely **not** in synced config; the new `isExpandActivityForRepeatEdit` is a safe optional boolean (typia forward-compat OK); the `_overwrite-material.scss` addition is the correct layer for a dialog panel class (can't live in component SCSS); the spec-hook leak from earlier rounds is resolved (flag hooks scoped to their describes); the converter change *removes* `new Date()` non-determinism; off-schedule Map-deletion is safe and preserves positions; no `any` in production and no user content logged.

johannesjo:
## Re: the last two items — both reproduce, here's the cause of each

Thanks @beerkumquatpome, you're right on both. Mea culpa: my earlier ✅ confirmed the code paths *exist*, not their behaviour in the default (flag-off / legacy) case — which is what you're hitting.

### 1. Off-schedule cells not transparent — the hide is gated behind the experimental engine

The `dayMap.delete(...)` that drops off-schedule days (so they render as transparent placeholders) only runs inside the projection branch — `repeat-task-heatmap.component.ts:320` → `if (isProjecting && cfg && cfg.rrule)`, where (`:243`):

```ts
const isProjecting = \!\!cfg && isRRuleEngineEnabled() && \!\!cfg.rrule && isRRuleValid(cfg.rrule);
```

So off-schedule hiding fires **only** when the per-device RRULE-engine flag is ON *and* the cfg has a valid `rrule`. But the flag is **off by default** (`rrule-engine-flag.ts`, localStorage), and a `cfg.rrule` only exists for tasks authored through the RRULE builder while the flag was on — every legacy `repeatCycle` task has none. Meanwhile the heatmap renders for **every** recurring cfg (`dialog-edit-task-repeat-cfg.component.html:33`, gated only by the collapsible being expanded), initialising the whole year at level-0 (`:246-258`) with no deletion → grey "missed" cells everywhere. That's exactly the streak-breaking look you reported.

The rendering half is fine — when the deletion *does* run, the slot becomes `null` → `.day.empty { background: transparent }` with position preserved (`heatmap.component.scss:134`). It's purely a gating problem.

**Design call, @johannesjo:** the streak cosmetic is conceptually separate from the flag-gated *projection*. Either (a) keep it engine-gated — then it's working-as-intended and only testable with the flag on + an rrule-built task (but it stays invisible to normal users), or (b) decouple it: compute `scheduledDays` from the legacy engine when `cfg.rrule` is absent and run the deletion regardless of the flag, so the everyday heatmap benefits. (b) is what the request implies, but it's the larger change.

Quick way to confirm the diagnosis: flip the flag on, create a *new* recurring task via the builder — its heatmap hides off-schedule cells; an existing/legacy task won't.

### 2. Multiple tooltips on fast mouse movement — instant matTooltip on hundreds of cells

The `title`→`matTooltip` swap killed the ~1.5 s native delay but swung to the other extreme: no `matTooltipShowDelay` is set (and there's no global `MAT_TOOLTIP_DEFAULT_OPTIONS`), so it uses Material's **0 ms show**. On a grid of hundreds of gap-tight 12 px cells, an instant show on every `mouseenter` paints a trail of overlays faster than they dismiss when you sweep the cursor in circles.

`heatmap.component.html:49` and `heatmap-month-calendar.component.html:36`:

```html
[matTooltip]="getDayTitle(day)"
matTooltipPosition="above"
[matTooltipShowDelay]="300"   <\!-- add this -->
```

A pending show is cancelled on `mouseleave`, so during fast movement nothing shows and only the cell you rest on gets a tooltip — which is also the GitHub-style "rest to reveal" timing you asked for (tune 200–400 ms; worth a shared const across the two call sites). Matching GitHub *exactly* (instant yet never stacking) would need a single shared tooltip element rather than per-cell directives — bigger change, likely not worth it.

### Still open from the last round (minor, unchanged at this head)
- Projected **tomorrow** cell loses its dashed outline after 12:00 — `repeat-task-heatmap.component.ts:318` builds `tomorrow` as `now + 24h` (carries time-of-day) vs noon-anchored occurrences.
- a11y: rrule-builder native controls / weekday toggles still have no accessible name.

johannesjo:
## Review: `60262f1..7504597` (RRULE Epic P2)

Reviewed across correctness / security / architecture / performance / simplicity. Solid, defensively-reasoned work — the converter, never-fires sentinel, and heatmap legacy-projection are correct and well-tested, and the sync-correctness invariants hold (no synced-enum widening; the sentinel writes only in-union values). One blocker, two small fixes, a few quality items.

### 🔴 Blocker — `isRRuleValid`'s decade-bounded probe does NOT bound never-firing rules

`src/app/features/task-repeat-cfg/store/rrule-occurrence.util.ts` (`isRRuleValid` ~245-262, `_canNeverFire` ~150-210)

`rule.between(anchor, anchor+10y, …)` only stops once an occurrence is *emitted* past the upper bound. A rule that emits nothing isn't bounded by the window — rrule.js walks day-by-day to its internal year ceiling before returning empty. `_canNeverFire` only models `BYMONTH × {BYMONTHDAY/BYYEARDAY/BYWEEKNO}`; it misses `BYYEARDAY × BYWEEKNO` and `BYSETPOS`-exceeds-set. Measured against this repo's `rrule@2.8.1`:

| Rule | Result | Probe time |
|---|---|---|
| `FREQ=DAILY` | fires | 1 ms |
| `FREQ=DAILY;BYYEARDAY=60;BYWEEKNO=53` | never | **7345 ms** |
| `FREQ=WEEKLY;BYDAY=MO;BYSETPOS=2` | never | **975 ms** |
| `FREQ=MONTHLY;BYMONTHDAY=15;BYSETPOS=3` | never | **198 ms** |

So the comment "the window bounds the work … so rrule.js's walk-to-year-275760 freeze can never trigger" is inaccurate — only rrule's internal year ceiling stops it, at ~7s. Reachability: the scan path is gated by `isRRuleEngineEnabled() && isRRuleValid(...)`, but the dialog's `rrulePreview`/`rruleLegacyIncompat` computeds call `isRRuleValid` gated only on `isRRuleMode()` — so in RRULE mode, a raw-override keystroke freezes the dialog (first time per string; memoized after). A synced/imported rule freezes the first scan once the flag is on.

**Fix:** add a hard wall-clock/iteration ceiling inside the `between` callback (abort → invalid → legacy fallback), correct the comment, and add a timing regression spec. Hand-enumerating BY-part interplay will always leave a residue, as these three uncaught cases show. Must land before the engine flag is ever defaulted on.

### 🟡 Warnings

**`quickSettingOptions` lost its memoization** — `dialog-edit-task-repeat-cfg.component.ts:187`. The `computed` reads the whole `repeatCfg()` signal, which is `.set()` on every keystroke in any field (formly emits a cloned model), so ~16 `translate.instant()` + 4 `toLocaleDateString()` re-run per keystroke. The adjacent `_previewScheduleCfg` already uses a custom `equal` for exactly this reason — mirror it here (key on a value-equal `{ refDateStr, locale }`).

**Sentinel re-derivation gated on `working.startDate`** — `dialog-edit-task-repeat-cfg.component.ts:~760-792`. The only place the never-fires sentinel is written on the RRULE save path (`...rruleToLegacyTaskRepeatCfg(cfg.rrule, finalStartDate)`) is inside `if (scheduleTouched && working.startDate)`. If `startDate` is ever falsy at save, a non-representable rule persists with stale legacy fields instead of the sentinel → old/flag-off clients fire on wrong days, contradicting the authoring warning. Latent (the form always sets `startDate`), but worth an unconditional derivation or an assert.

### 🟢 Suggestions

- **`repeat-freq-picker` reinvents `mat-select`/`mat-menu` with a raw CDK overlay** (~330 new lines; the only `cdkConnectedOverlay` in `src/app`, vs `mat-select` in 18 files). The one genuinely-custom need — "More options expands in-panel without closing" — is a natural `mat-menu` fit and would delete ~200+ lines. Adopt Material or flag the deviation explicitly.
- **`rrule-builder` hand-rolls 3 segmented controls (`.rb-seg`)** instead of the shared `segmented-button-group` (drop-in `options`/`selectedId`/`selectionChange`/`size` API, plus keyboard nav the hand-rolled version lacks; used in 6 places). Reuse it — this is the "make it reusable / shared layer" styling rule.
- **`_legacyEquivalent` `switch (cycle)` has no exhaustiveness guard** (`legacy-cfg-to-rrule.util.ts`) — a future `RepeatCycleOption` silently falls to `return null` → sentinel for a real schedule. Add `default: assertNever(cycle)`.
- **Dual-engine routing duplicated** between `_hasProjectionSchedule` and `_scheduledDaysInYear` (`repeat-task-heatmap.component.ts`) — the comment already says "Mirrors the routing in…". Extract one predicate.

### ✅ Verified clean
No XSS sinks; no user-content logging; `rrule@2.8.1` audit-clean and the exact pin is sound; heatmap legacy loop correctly bounded (400-guard + advancing cursor); live calendar preview properly gated + memoized; converter wire format correctly hand-rolled (rrule `.toString()` would corrupt it); sentinel design is the right minimal-correct fallback.

Note: CI runs no tests on PRs into `feat/rrule-epic`, so run `npm test` locally on the converter / occurrence-util / heatmap / dialog specs before merge.

johannesjo:
**Review — Phase 2 push (`60262f1..7504597`)**

Ran a multi-perspective pass (correctness, sync-safety, architecture, alternatives, performance, simplicity) over this range. **No critical or sync/data-loss issues** — I traced both released legacy occurrence engines, the `_toPersisted`/`_sanitizeAnchors` boundary, and the op-log wire contract, and the never-fires sentinel design holds up: `repeatCycle:'WEEKLY'` + all-false flags is genuinely unfirable on every released client, the wire-stable values survive the JSON partial-update merge, and the dropped `undefined` anchor clears are inert because anchors are only read in the legacy MONTHLY branch. The sentinel is the *minimal* correct design here, not gold-plating. Heatmap changes are confirmed display-only.

Three things I'd like to settle before merge:

**1. `repeat-freq-picker` accessibility regression.** The panel declares `role="listbox"` but the option children are plain `<button>`s — no `role="option"`, no `aria-selected` (only `[class.active]`), no roving `tabindex`/arrow-key nav, no type-ahead, and focus never enters the panel on open (only Escape is handled). It replaces a `select` that had full keyboard support. Either lean on CDK Menu primitives (`cdkMenu`/`cdkMenuItem` — gives roving focus + arrow keys + type-ahead, and a `cdkMenuItem` can keep the panel open for "More options"), or at minimum add `role="option"` + `[attr.aria-selected]` and move focus into the panel on open. Related question: this is the only hand-rolled `cdkConnectedOverlay` dropdown in the app (~329 lines + bespoke select styling) — is the "expand the long tail without closing the panel" interaction a hard requirement vs. a flat `mat-select`? If yes, the component's justified once a11y is fixed.

**2. `isRRuleValid` 10-year probe — accepted tradeoff, but please pin it.** The switch to `between(2020-01-01, 2030-01-01, true, cb→false)` correctly bounds the never-fire walk and early-exits on first hit. But it's anchored at a *fixed* `dtstart 2020-01-01`, so a valid rule whose first occurrence is >10y past that anchor is now classified invalid. Verified concretely against rrule 2.8.1: `FREQ=YEARLY;BYMONTH=2;BYMONTHDAY=29;BYDAY=MO` first fires 2044-02-29 → was `valid`, now `invalid` (→ no preview, no incompat warning, `save()` rejects). Only reachable via raw-override and documented in the watchlist as a deliberate product rule, so I'm not asking to change the behavior — just to **add a spec asserting the >10y boundary** so a future window edit can't silently flip a real rule.

**3. Heatmap now projects legacy `repeatCycle` cfgs with the flag OFF.** `_scheduledDaysInYear`/`_hasProjectionSchedule` route any recurring cfg through the legacy engine itself, not just flag-on RRULE cfgs. This looks like a genuine improvement (uses the real engine, so no converter divergence; stays display-only) — but it contradicts the **"Guard rails: with the flag off the heatmap stays the unchanged history-only view"** line in the PR description. Could you confirm the broadening is intended and update the description?

Optional polish (non-blocking):
- **SCSS token drift:** the builder swapped `var(--card-border-radius)` for literal `9px`/`38px` in several places, and the dropdown chevron is an inline SVG with hardcoded `fill='%23888'` (won't track light/dark). Prefer tokens / `currentColor`.
- **Orphaned i18n key:** `RRULE_YEARLY_MODE_DESCRIPTION` lost its last template reference in the restyle but remains in `t.const.ts` + `en.json`. (All 9 new keys are present and `en.json` is valid.)
- **DRY:** the `'RRULE'` sentinel is redefined in the picker vs the options builder, and the RRULE-vs-legacy routing predicate is copied between the two heatmap methods — a shared const + `_isRRuleAuthoritative(cfg)` helper avoids drift.
- **Perf (cold path):** `safeParseRRuleOptions` isn't memoised, so a valid rule re-parses ~4× per keystroke; a small string-keyed cache (like `_validityCache`) would also help the on-every-scan engine paths.

Couldn't run the new specs from my end (working tree was on `master`, so the epic-branch spec files weren't present) — conclusions are from reading + targeted rrule 2.8.1 probes. Trusting the PR's "10964/10964 green locally."

johannesjo:
## Review — range `60262f16..75045974` (legacy-fallback + TickTick dialog restyle)

Verdict: **looks good to merge after one comment/heuristic fix.** The sync-correctness core is solid — I verified the `LEGACY_NEVER_FIRES_FALLBACK` sentinel is deterministically dead on the legacy engine and wire-stable, that `_legacyEquivalent()` is the single source of truth for both `isRRuleLegacyRepresentable` and the converter (no drift), that interval-1 `FREQ=DAILY;BYDAY` → WEEKLY flags is byte-identical, and that the `rrule` exact-pin rationale holds. 👍

### 🟠 Important — the bounded probe comment overstates the fix (and `_canNeverFire` has a gap)

`src/app/features/task-repeat-cfg/store/rrule-occurrence.util.ts` — `isRRuleValid`

The comment says the `between(anchor, anchor+10y)` probe means *"rrule.js's walk-to-year-275760 freeze can never trigger"* and cites *"BYSETPOS past its set"* as a now-bounded example. Measured against the pinned `rrule@2.8.1` with the exact production probe (UNTIL/COUNT stripped, `between(2020,2030)`, early-exit callback):

| RRULE | caught by `_canNeverFire`? | probe time |
|---|---|---|
| `FREQ=WEEKLY;BYDAY=MO,WE,FR` (fires) | n/a | 1 ms |
| `FREQ=DAILY;BYSETPOS=2` (never fires) | **no** | **~4.7 s** |
| `FREQ=DAILY;BYWEEKNO=10;BYYEARDAY=300` | **no** | **~7.5 s** |
| `FREQ=WEEKLY;BYDAY=MO;BYSETPOS=5` | **no** | ~0.9 s |

`between()` only filters *emitted* occurrences, so a never-firing DAILY-expansion rule still walks day-by-day to rrule.js's internal ceiling before emitting nothing — the early-exit never fires. This matches the watchlist doc in this same PR ("a bounded between/until probe CANNOT replace the `_canNeverFire` heuristics"), but the code comment claims the opposite. The risk isn't today's performance (it's not a regression vs the old `.after()`, it's memoised, and it's only reachable per-keystroke via the advanced raw-override field, plus the flag-gated scan path) — it's that a future reader trusts the comment and deletes the load-bearing `_canNeverFire`, reintroducing the 7–10 s freeze for the `BYMONTH`-contradiction class too.

Please either:
1. **(min)** Correct the comment + commit message to match the watchlist: the probe does *not* bound never-firing rules; `_canNeverFire` is the real guard for the slow classes and the probe is best-effort for the rest. Also `MAXYEAR` in rrule.js 2.8.1 is `9999`, not `275760`.
2. **(better)** Close the heuristic gap the comment already implies it covers — flag `BYSETPOS` whose magnitude exceeds any plausible per-FREQ set size, and the `BYWEEKNO`×`BYYEARDAY` contradiction when there's no `BYMONTH` (the current month-intersection checks are skipped when `bymonth` is empty). Or debounce the raw-override preview.

### 🟡 Worth considering

- **Reinvented framework components** (`repeat-freq-picker`, `rrule-builder`): the picker hand-rolls the *only* `cdkConnectedOverlay` in the app (re-implementing listbox role, width-match, escape/backdrop, focus) and the "Ends"/"Schedule type" `.rb-seg` segmented controls re-do `mat-button-toggle-group`, which `dialog-idle` already uses for this exact pattern. The genuine constraint `mat-select` can't meet is "More options expands in-panel without closing" — worth a one-line code comment recording why Material was rejected, so a maintainer doesn't ask later. (The multi-select weekday/day-grid/month chips are fine as plain buttons — not worth converting.)
- **a11y** (`repeat-freq-picker.component.html`): the panel is `role="listbox"` but its children are `<button>` without `role="option"`/`aria-selected`, and there's no roving arrow-key navigation. Either drop `role="listbox"` (treat as a menu of buttons) or complete the pattern.
- **File size**: `legacy-cfg-to-rrule.util.ts` is now ~500 lines (>300 guideline). It's cohesive, but the forward converter / `_legacyEquivalent` inverse / `getAlignedStartDate` are three separable units sharing the weekday+clamp helpers — a natural split as the epic grows.

### 🟢 Minor / nits

- `repeat-freq-picker` has no unit spec; `presets()`/`canToggle()`/custom-ordering have real branching only covered indirectly via the dialog spec.
- Heatmap templates call `getDayClass()`/`getDayTitle()` (and `getCellClass()`/`getCellTitle()`) per cell per CD cycle — low impact (OnPush leaf, only while preview open), but precomputing into the `DayData`/`CalCell` objects in `build-heatmap-data.util.ts` would honor the "no function calls in templates" guideline.

Nice work on the `computed()`-over-formly migration and reusing the memoised `isRRuleValid` as the single preview/save gate.

<sub>🤖 Synthesized from a parallel multi-agent review (correctness, security, architecture, alternatives, performance, simplicity); the probe timings were re-measured directly against `rrule@2.8.1`.</sub>

omega-tree:
We're not going to 1:1 copy TickTick — SP isn't a clone, and our recurring builder intentionally does far more than TickTick's
 (full RFC-5545: intervals, BYDAY/BYSETPOS, week numbers, count/until). The simplicity you're describing comes from it doing less. The one concrete bit — three calendars — is fair: start-date (sets the anchor), preview (read-only, collapsed, shows when it fires), heatmap (history). They're not redundant, but I'll look at folding the start date into an interactive preview. Specific issues help; "watch this video and copy it" doesn't.

omega-tree:
Thank you for the useful suggestions!

1. you are absolutely right on renaming Weekday! Thanks!

2. eng generation is pulled from a lib.. will add eng overrides in a future PR within the epic.

3. that sounds like a good behavior change.

4. merging of 1,2,4 calendars is already in the works.  # 3 is not a calendar it is a day selector. End date should stay as its own date field. I can look into a right click multi option sub-menu later on the calendar to add end dates but still would be more awkward then just this end-date date selector field.

5. is just a rewording of using tick tick UI.. I'll look into a right-click/long hold (mobile) sub-menu on each calendar day. maybe that will fix the complaint but this is much larger then tick ticks narrow recurring feature set.

Ok no problem I'll take a look at v18.10.0

johannesjo:
**Review of `fix(task-repeat-cfg): close _canNeverFire gaps + correct stale probe comments`**

Nice cleanup on the stale `year-275760` / probe-anchor comments — those are now accurate to rrule.js's `MAXYEAR=9999` behavior. The **BYWEEKNO × BYYEARDAY** check is correct and conservative (I fuzzed all 16k+ flagged YEARLY combos plus DAILY/WEEKLY/MONTHLY boundary cases against rrule.js 2.8.1 — zero false positives).

But the new **BYSETPOS-beyond-set** check introduces false positives, and since `isRRuleValid → false` silently drops a cfg to the legacy `repeatCycle` fallback and reschedules it, these are sound recurring rules getting silently broken. Verified end-to-end against rrule.js 2.8.1 (each FIRES in rrule, but `_canNeverFire` returns `true`):

```
FREQ=DAILY;BYSETPOS=-2                    → rrule fires daily      (heuristic: never)
FREQ=DAILY;BYSETPOS=-100                  → rrule fires daily      (heuristic: never)
FREQ=WEEKLY;BYDAY=MO;BYSETPOS=-2          → rrule fires every Mon  (heuristic: never)
FREQ=WEEKLY;BYMONTHDAY=1,2,3;BYSETPOS=2   → rrule fires 2024-01-02 (heuristic: never)
FREQ=WEEKLY;BYYEARDAY=1,2,3;BYSETPOS=2    → rrule fires            (heuristic: never)
FREQ=WEEKLY;BYWEEKNO=10,20;BYSETPOS=2     → rrule fires 2024-03-05 (heuristic: never)
```

Two root causes:

1. **Negative BYSETPOS is unsound.** `bySetPos.every((p) => Math.abs(p) > setPosSlots)` assumes rrule.js yields nothing for an out-of-range negative — but rrule.js *clamps* it and still fires. The other combo checks in `_canNeverFire` already skip on any negative value; BYSETPOS should too.

2. **The WEEKLY bound isn't an upper bound when BYDAY is absent.** The comment says "Other BY parts only filter it DOWN," but with no BYDAY, rrule.js treats `BYMONTHDAY`/`BYYEARDAY`/`BYWEEKNO` as *expanders*, so the per-week set is larger than the BYDAY count (which `_maxSetPosSlots` returns as `count || 1 = 1`).

These all fell through to the probe and resolved correctly **before** this commit, so it's an optimization that regresses correctness for the synced/imported/REST-ingested rules this guard is specifically meant to cover. `DAILY=1` itself is fine (only sub-day parts expand a day, and those already bail to `null`), and the positive + BYDAY-present cases are correctly flagged.

Suggested fix:
- Flag only positive overshoot: `bySetPos.every((p) => p > setPosSlots)` (negatives are then never `> setPosSlots ≥ 1`).
- `_maxSetPosSlots` should return `null` for WEEKLY when `byyearday`/`byweekno`/`bymonthday` is present (only bound WEEKLY when the sole positional part is BYDAY).

The "no false positives" test (`complex.spec.ts:158-165`) only covers `BYSETPOS=-1` and BYDAY-present cases, so the suite stays green over this. Worth adding `DAILY;BYSETPOS=-2`, `WEEKLY;BYDAY=MO;BYSETPOS=-2`, and `WEEKLY;BYMONTHDAY=1,2,3;BYSETPOS=2`.

**Minor:**
- The `expect(performance.now() - start).toBeLessThan(500)` asserts (`complex.spec.ts:155`, `:174`) are fragile: `isRRuleValid` memoizes into a module-level cache that isn't reset between tests (so the timed block can be a pure cache hit), and wall-clock thresholds flake in CI. The existing sibling test uses a commented `< 1000` — either match that or drop the timing asserts and keep the `toBe(false)` checks.
- `repeat-freq-picker.component.spec.ts` is solid but unrelated to `_canNeverFire` and undisclosed in the commit message — probably belongs in its own `test(...)` commit.
- Optional: extract `hasNegYearDay`/`hasNegWeekNo` consts to mirror the existing `hasNegDay` pattern and de-dup the negative guards.

omega-tree:
thanks for the advice fellas.

  after a ton of shuffling and iteration i seem to be settling on the new front-and-center monthly & 365 calendar. the power's really in the submenus — clicking on days, weekdays, and months lets you control almost everything the rrule form builder can. honestly feels like it's in a class of its own — no other task manager is this flexible while staying this approachable.

  three levels, one rule underneath:
  - calendar — basic users work entirely here. click a day to set start/end, a weekday to repeat on it, a month to limit the season, the date for "monthly (15th)" or "last friday." plain-language, toggles on/off, hover to see what's set on a day/weekday/month.
  - rrule builder form — finer control for the small edge cases.
  - raw RRULE field — complete customization for the even fewer edge cases.

  all three stay in sync (sketch on the calendar, refine in the builder, read the raw string — nothing lost), and a live heatmap preview shows every upcoming occurrence: next date spotlit + pulsing, start/end marked, your tracked time overlaid. ships off-by-default behind one setting, so zero risk to existing recurring tasks, and it's tuned to match the theme (incl. dark mode) across all the built-in palettes.

johannesjo:
**Review of the simulate-completion commits (`d95d263`, `c9727f8`)**

Thanks for these — the start-date gating and the sim-preservation-across-start-moves are nicely done. The `_lastScheduleSlice` field + lockstep advance is correct: both `menuSimulate` and `_applyStartDate` advance the tracker synchronously before the watcher effect runs, so the freshly-set sim survives the schedule-slice change. Routing both the menu and the M/D/Y fields through `_applyStartDate` also keeps the two start-date paths consistent. 👍

One must-fix and two design questions:

**🔴 Must-fix — simulate on a COUNT rule blocks save.** `menuSimulate` flips `repeatFromCompletionDate → true` unconditionally. For a start-anchored COUNT rule (e.g. `FREQ=WEEKLY;BYDAY=MO;COUNT=5`) that combination is rejected at save with `RRULE_COUNT_WITH_COMPLETION`. So a user who clicks "simulate" purely to preview can no longer save, with an error about a setting they never knowingly changed. (The preview also re-anchors the COUNT series from the sim day, which can render up to ~2×COUNT marks.) Suggest gating the simulate item off when the rule carries a COUNT, or not mutating the persisted flag.

**🟡 Design Q1 — the flip isn't reverted on toggle-off.** Turning the sim off clears `simulatedCompletion` but leaves `repeatFromCompletionDate = true`, so simulate on→off permanently converts a start-anchored rule to from-completion (persisted on save in RRULE mode). If that one-way conversion is intended, 👍 — but it's worth a covering test; right now nothing exercises toggle-off after a flip.

**🟡 Design Q2 — preset (non-RRULE) preview/save mismatch.** The calendar + simulate menu also render for presets (the `@else` outlet). On a preset, simulate flips the flag and the preview re-anchors, but save clears the flag back to `false` on the non-RRULE branch — and presets expose no from-completion control — so the preview shows behavior the saved cfg won't have. Consider hiding simulate for presets, or switching to RRULE mode on flip.

**Nits**
- spec ~L674 (`"Simulation only exists for repeat-from-completion schedules"`) now contradicts the test right below it, which flips a start-anchored rule.
- `menuSimulate()` itself isn't guarded by `menuDaySimAllowed()` (only the template button is) — a defensive guard would harden it.

**Tests:** these read correct, but heads-up that CI runs no unit tests on PRs into `feat/rrule-epic` (base-branch filter), so `npm test` needs to run locally before merge.

johannesjo:
Thank you very much! <3

PR #8238, opened 2026-06-09
fix(electron): own sync folder + image cache main-side, drop renderer-supplied paths (#8228)

### LINKED ISSUE #8228: Security: file-sync IPC accepts arbitrary renderer paths (capability/grant model needed)
## Summary

The `FILE_SYNC_*` and related filesystem IPC handlers in `electron/local-file-sync.ts` still let the renderer (and therefore any plugin / XSS) supply absolute paths and operate on them, as long as the path is outside `userData`. Outside-userData is a useful backstop but it is **not** the authorization model — a hostile renderer can still read, write, list and delete arbitrary user files (`/etc/passwd`, `~/.ssh/authorized_keys`, `~/Documents`, etc.) via the preload bridge.

The narrower `127.0.0.1:<any-port>` in-window navigation hole was fixed in #XXXX (commit `ca49009fd`, this branch). This issue tracks the larger refactor.

## Affected IPCs

All accept renderer-supplied absolute paths and only check `assertPathOutside(userData)`:

- `FILE_SYNC_SAVE` — write
- `FILE_SYNC_LOAD` — read
- `FILE_SYNC_REMOVE` — delete
- `FILE_SYNC_LIST_FILES` — directory listing
- `CHECK_DIR_EXISTS` — directory listing
- `READ_LOCAL_IMAGE_AS_DATA_URL` — read (background images)
- `TO_FILE_URL` — converts path → `file://` URL. On `master` it is **not** guarded at all (the `assertPathOutside` tightening from `ca49009fd` only exists on the security-review branch). Land that backstop first, then refactor.

### Grant-origin IPCs (return raw paths today)

These are also part of the same pattern — they hand absolute paths back to the renderer, which then echoes the path into a later FS IPC. After this refactor they should return a `grantId` (or a grant + opaque ref), not a bare string:

- `PICK_DIRECTORY` — sync folder picker
- `SHOW_OPEN_DIALOG` — used by `formly-image-input.component.ts:57` and `clipboard-images-cfg.component.ts:101`

## Trust model today vs. needed

Today: "renderer can touch any file outside `userData`."
Needed: "renderer can touch files under a user-approved sync folder, and nowhere else."

## Proposed direction — capability-based, not path-string-based

1. **User picks a sync folder** via main-process `dialog.showOpenDialog` (already exists as `PICK_DIRECTORY`).
2. **Main canonicalizes the path** (`fs.realpathSync.native`, same approach as `electron/file-path-guard.ts`) and stores an approved root as a **grant**.
3. **Main returns an opaque `grantId`** to the renderer — not the raw path, not any authority by itself.
4. **Renderer sends `{ grantId, relativePath }`** for every FS op.
5. **Main resolves** `target = path.join(root, relativePath)`, rejects absolute paths, `..` traversal, symlink escapes that land outside the root, and any path the canonical compare puts outside the granted root.
6. Grants are **persisted in main-owned storage** (concretely: a new `grants` key in `electron/simple-store.ts`, or a sibling `electron/grant-store.ts` if we want a stricter file. Either way the file lives inside `userData` and the renderer cannot touch it because of the existing `assertPathOutside(userData)` backstop). Re-validated on app start.

### Per-feature grant scope

Grants must carry a `feature` tag (e.g. `'sync'`, `'background-image'`) and the main-side handler must check `grant.feature === expected` per IPC. Reason: the renderer runs untrusted plugin code (`new Function` in `src/app/plugins/plugin-runner.ts`). If the sync-folder grant is reachable via the same `window.ea` surface plugins use, a malicious plugin can corrupt the sync folder and the corruption then propagates to every device via op-log. Either plugins do not see sync-scope grants at all (preferred), or the per-IPC feature check enforces it.

For `READ_LOCAL_IMAGE_AS_DATA_URL`, prefer a different shape: copy the picked image into an app-owned cache and serve by id. Backgrounds are picked once and rarely change — a long-lived file capability for a one-shot read is overkill, and migrating existing config is the only real work. `TO_FILE_URL` for image inputs collapses into the same cache.

## Defense-in-depth that stays

The current `assertPathOutside(userData)` guard should stay as the backstop, not the primary authorization. Same for the existing path-canonicalization in `electron/file-path-guard.ts`.

## Symlink containment — needs explicit policy decision

`fs.realpath` + prefix-check has TOCTOU between resolve and the actual write. Options:

- `O_NOFOLLOW` on open. Available on Linux **and macOS** (POSIX/BSD); Windows is the outlier. Node exposes it via numeric flag.
- `stat` on open + compare inode / compare to canonical root. To shrink the race window, operate on the fd: `open` → `fstat` → `realpath` of `/proc/self/fd/<n>` (Linux) or the platform equivalent — not on the path.
- Just refuse to follow symlinks and document it. Simplest, fine for this codebase given sync is the only real use case.

Pick one before implementing; the file-path guard tests should cover whichever is chosen.

## Migration policy — re-confirm, do not auto-accept

Two options:
1. Auto-accept the existing `syncFolderPath` from persisted config as the grant root on first start.
2. Re-prompt the user with a confirmation dialog ("Continue syncing with `<path>`? Choose folder…").

Default to (2). Option (1) rubber-stamps whatever is in the config, including a tampered value. UX friction is the cost of proving user intent; confirmation can be a single-click "Confirm" with a "Choose other folder…" link.

## Test coverage to add

- `FILE_SYNC_LOAD('/etc/passwd')` (or platform equivalent) is rejected.
- `FILE_SYNC_SAVE(grantId, '../escape')` is rejected.
- Symlink inside granted root pointing outside is rejected (or explicitly handled per policy above).
- Grant for a folder no longer present at startup is invalidated cleanly.
- Renderer-supplied `grantId` that doesn't match a real grant is rejected.
- Cross-feature grant misuse: a `'background-image'` grant cannot be used in a `FILE_SYNC_LOAD` call.
- `TO_FILE_URL` rejects a userData path on `master` (backstop step before the larger refactor).

## Blast radius

The renderer-side adapter at `src/app/op-log/sync-providers/file-based/local-file/electron-file-adapter.ts` and the persisted sync config both currently round-trip absolute paths. The refactor needs:

- Main-side grant persistence (new `grants` key in `simple-store`, or a sibling `electron/grant-store.ts`).
- Renderer-side adapter migration from `{filePath}` to `{grantId, relativePath}`.
- A one-time migration: on first start after upgrade, prompt the user to confirm or re-pick the sync folder; do not silently promote the persisted path.
- Updated `ElectronAPI` types and preload signatures.
- Grant revocation / inspection UX (a Settings entry: "Sync folder: `<path>` — Change… / Remove"). Cheap now, awkward to retrofit.

Not a one-file change, but the shape is self-contained.

## Related

- `electron/file-path-guard.ts` — canonicalization helper (keep as backstop).
- `GHSA-x937-wf3j-88q3` — earlier round of this same class of issue.
- Original review notes in the security-review worktree branch.

johannesjo:
Closes #8228.

## What's the hole

The renderer (which runs untrusted plugin code via `new Function` and is the assumed XSS-attacker surface) could ask main to read/write/list/delete any file outside `userData` by supplying an absolute path. The `assertPathOutside(userData)` backstop was a useful guard against settings/grant-file tampering, but it was never the authorization model — `/etc/passwd`, `~/.ssh/authorized_keys`, `~/Documents/*` were all reachable.

## How this PR closes it

The full sweep — 6 real commits plus one revert (kept for the audit trail):

| Commit | What lands |
|---|---|
| `c12982f1f` | `electron/sync-path-resolver.ts` — `(syncRoot, relativePath, userData) → absolutePath`. Canonicalizes root every call; refuses absolute paths, `..` traversal, symlink leaves, canonical-checks the deepest existing ancestor so a mid-path directory symlink can't sneak the write outside the root; fails closed on EACCES; backstops with `assertPathOutside(userData)`. |
| `0c21649fd` | All `FILE_SYNC_*` IPCs (`SAVE/LOAD/REMOVE/LIST_FILES`) + `CHECK_DIR_EXISTS` now take `{ relativePath }`. The sync folder path is owned main-side in `simple-store`. The package's `LocalFileSyncElectron.isReady` queries main, not the renderer's `privateCfg`. |
| `00e71b407` | Post-review fix: the Formly button at `sync-form.const.ts` had `key: 'syncFolderPath'`, so the picker's return value kept flowing back into renderer `privateCfg`. Dropped the key + the LocalFile default. Inlined the sync-folder cache (~25 LOC vs a separate module), dropped redundant typeof check, made PICK_DIRECTORY surface a distinct safe Error on persist failure, canonicalized at write time. |
| `601a26c1f` | Image picker copy-to-cache: `electron/image-cache.ts` (random-id-named files under `userData/bg-images/`). Form stores `image:<id>`; resolver renders via main-only IPC. |
| `ecb72fbc4` | Dropped legacy `READ_LOCAL_IMAGE_AS_DATA_URL` + `TO_FILE_URL` IPCs entirely — they no longer have a writer (picker uses copy-to-cache) and the renderer-supplied-path shape is the exact thing this issue is closing. |
| `38d2b16a8` | Post-review fix: the previous `IMAGE_CACHE_IMPORT` accepted a renderer-supplied absolute path with no link to a recent dialog — a compromised renderer could `import(~/Pictures/screenshot.png)` and read it back. Merged into atomic `IMAGE_PICK_AND_IMPORT`: main opens the picker and calls `importImage` with the file the user clicked, in one IPC. The renderer never holds a path. Optional `replacesId` GCs the prior cached image in the same call. |
| `2b628a0a5` | Pre-merge nits: `IMAGE_PICK_AND_IMPORT` now funnels through `createSafeIpcError` (stack/path-content stripped, parity with the FS handlers); image picker button is disabled while the IPC is in flight (a double-click would otherwise orphan the first import); stale comments/placeholders fixed. |

## Trust model — before vs after, scoped to this PR

**Before, for the IPCs this PR touches:**
> The renderer can touch any file outside `userData` via `FILE_SYNC_*`, `CHECK_DIR_EXISTS`, `READ_LOCAL_IMAGE_AS_DATA_URL`, or `TO_FILE_URL`.

**After:**
> For sync (`FILE_SYNC_*`, `CHECK_DIR_EXISTS`): renderer touches only files under the main-stored sync folder, addressed by relative path. For background images (`IMAGE_PICK_AND_IMPORT`, `IMAGE_CACHE_GET_DATA_URL`): renderer holds only opaque ids for images the user clicked through a native dialog. No renderer-supplied absolute path crosses any of these IPCs.

**Not in scope for this PR** — the renderer still has unrelated path-emitting IPCs (`SHOW_OPEN_DIALOG`, `SAVE_FILE_DIALOG`, the `CLIPBOARD_*` family) that take renderer-supplied paths. They have their own threat surface and should be audited as sibling issues; see follow-ups below.

## Migration

- **Sync folder.** The renderer's pre-fix `privateCfg.syncFolderPath` is no longer load-bearing. On first launch after upgrade, `isReady()` returns false; the existing "select sync folder" UI prompts the user to re-pick. A `logger.critical` line lands on the dev console when the legacy value is present but main has no folder — the user-facing snack is a documented UX follow-up (needs i18n).
- **Background images.** Pre-fix `file://` values no longer resolve (the IPC that read them is gone). The resolver returns null and emits `console.warn`. User re-picks via the existing button.

## Tests

- **70** electron security tests pass (file-path-guard, simple-store, sync-path-resolver, image-cache, local-file-sync IPC integration)
- **17** sync-providers package vitest cases for local-file
- **8** karma cases for `formly-image-input`
- **5** karma cases for `resolveBgImageToDataUrl`
- `npm run lint` clean, `tsc --noEmit` clean for both renderer and electron

## Follow-ups (explicit, so they aren't forgotten)

Parallel renderer-supplied-path holes the same class of fix would apply to. Worth filing as sibling issues:
- The whole `CLIPBOARD_*` IPC family (`CLIPBOARD_IMAGE_SAVE/LOAD/DELETE/LIST/GET_PATH`, `CLIPBOARD_COPY_IMAGE_FILE`) takes a renderer-supplied `basePath` and uses it for reads/writes/listings.
- `SAVE_FILE_DIALOG` returns absolute paths to the renderer.

Hygiene / hardening, lower priority:
- **Cache-orphan GC**: `bg-images/` has no startup sweep for orphaned files. The `replacesId` flow covers normal "user picked a new image" replacement, but a `removeCachedImage` failure (EACCES) or a stale `image:<id>` in a never-loaded project leaves files indefinitely. Realistic worst case is a few MB per orphan; not security-critical.
- **Symlink TOCTOU** between resolver and the IPC's `writeFileSync`/`readFileSync`. v1 policy refuses symlinks at the leaf; `O_NOFOLLOW`-on-open is the proper fix and needs per-platform handling (Linux/macOS have it; Windows doesn't).
- **User-facing migration snack** instead of `console.warn`. Needs i18n keys.
- **`PICK_DIRECTORY`** could nudge `defaultPath` toward a dedicated sync subfolder vs `~/Documents`.
- **End-to-end IPC integration test** wiring real `ElectronFileAdapter` → mocked IPC → real `local-file-sync` handler → real `sync-path-resolver`. The current suite covers each layer separately; a wire-format drift would pass both and fail at runtime.

## Test plan

- [ ] Fresh install: pick a sync folder, save data, verify sync round-trips.
- [ ] Upgrade install: sync used to work, on next launch `isReady()` returns false, picker re-prompts, sync resumes after re-pick.
- [ ] Image picker: pick a background, verify it renders. Pick a different one, verify the old cache file is removed from `userData/bg-images/`.
- [ ] Try a traversal: dispatch `fileSyncLoad({ relativePath: '../escape' })` from devtools — should reject.
- [ ] Configure a sync folder that's a symlink to outside itself — should be refused.
- [ ] Image picker: click twice in quick succession — second click should be a no-op while the first dialog is open.


johannesjo:

johannesjo:

johannesjo:
@copilot resolve the merge conflicts in this pull request

PR #8246, opened 2026-06-10
style(calendar): use dashed outline for hover on focused cells

hajiboy95:
@beerkumquatpome https://github.com/super-productivity/super-productivity/pull/8227#issuecomment-4670211209

Updates calendar cell hover behavior to use a dashed outline even if the cell is already focused. This provides clearer visual feedback for intentional mouse hovers over the active selection.

Ensures keyboard navigation remains solid by explicitly suppressing hover outlines in DateTimePickerComponent when sp-hide-cursor is active, preventing a stale mouse position from causing a dashed outline during keyboard use.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki).
- [ ] I have run `npm run checkFile` on changed `.ts`/`.scss` files
- [ ] I have added tests for my changes (if applicable)
- [x] Existing tests still pass
- [x] My commit messages follow the Angular format (`type(scope): description`)

PR #8258, opened 2026-06-10
fix(ui): preserve calendar focus highlight on mouse hovered cell

hajiboy95:
## Problem

Fixed [transparent color during selection ](https://github.com/super-productivity/super-productivity/pull/8246#issuecomment-4672367770)


johannesjo:
Thank you very much!

@beerkumquatpome what are your thoughts on this? :)

hajiboy95:
@beerkumquatpome 
1. you can. Your focus now only shifted to the arrow buttons (since you clicked there) by pressing the TAB button, you'll jump to the next element on the side. That's exactly what usualy keyboard Focus does at websites. Focus goes where you click.
2. If you move your mouse picker ouside of the window, you can then move the keyboard arrows, and it'll move again. Implementing a solution here would mean, the keyboard focus again pops up if you leave the window with you mouse. 

@johannesjo i tested both. both technically implementable. 
1. will touch focus behaviour. It's not too unusualy that focus jumps onclick somewhere. IF we would like to implement something, I'd go for: 
- Going to next month -> jump to first day of month
- Going to previous month -> go to last day of month. 
- If we are in current month and click "go back" -> Since we can't go further back -> Jump to first possible date 

2. Would you like the focus to pup up again if we leave the window?

I implemented both. You can check the code changed. if you don't like one or bot, i could revert. Meanwhile @beerkumquatpome you can test if it's according to what you've imagined.

johannesjo:
Thanks a lot for digging into all of this @hajiboy95, really appreciate the effort (and the offer to revert parts)! 🙏

After thinking it over, I'd like to keep this PR minimal and land just the actual fix:

**Keep:** the SCSS `:hover:not(:focus)` change. That's the real fix for the "transparent during selection" issue from #8246 — and it's nice because it only corrects our own custom highlight layer, not anything Material does.

**Drop:** the prev/next month focus-jump logic (`_onMonthChange`, the capture-phase click listener, the `_monthBeforeClick`/`_yearBeforeClick` state). My reasoning:

- Material intentionally keeps focus on the prev/next button after a click so you can page through months quickly — moving focus into the grid overrides a deliberate UX decision of theirs, and I'd rather not fight the component. As you noted yourself, "can't use the keyboard after prev/next" is really just focus-follows-click, which is expected browser behavior — so there isn't actually a bug to fix there.
- It's also the most complex/fragile part (it leans on the `disabledInteractive` quirk + event ordering + a timeout), which I'd prefer not to maintain against future Material updates. If there's a strong case for keyboard-paging-into-the-grid, that's probably best raised upstream at angular/components — I'm sure they have good reasons for the current behavior.

The `mouseleave` re-show is more of a judgment call (it only touches our own keyboard-nav highlight, not Material) — I'm fine either way; happy to keep it if @beerkumquatpome finds the current behavior annoying, otherwise let's drop it too to keep things tight.

So: could you trim this down to the SCSS fix (+ optionally the `mouseleave` bit)? Then I think it's good to go. Thanks again! 😊

hajiboy95:
@johannesjo thanks for your feedback. Reverted that change with focus change. Indeed that has been the one which I was hesitant too^^.

hajiboy95:
> >by pressing the TAB button, you'll jump to the next element on the side
> 
> How does the user know he needs to press tab?

That's completely normal keyboard interaction in apps :D. If users want to control apps via keyboard it's normal to press tab to move to next field. 

But can we merge this not or are there other annoyances?

johannesjo:
I am merging this. Thank you very much! <3

PR #8261, opened 2026-06-10
feat(tasks): add timestamped comment thread on tasks

### LINKED ISSUE #4412: 💡 Add comments to tasks
### Problem Statement

Tasks that often get **blocked** or take several days often require me scheduling the task in the future but I need to save some context for the task on that specific task. It isn't really a project wide thing - it is task specific context.

I also want to know the dates when I made the update as if I have 7-10 updates on a task that has been stuck for a month, I want to know how long since each task.

I often find myself putting in the description of tasks long descriptions such as:

05/01/2024 Electrician said he needs part X
05/05/2024 Electrician got part X but it wasn't big enough will return with part Y
05/10/2024 Electrician installed part Y but says issue with other bureacracy and need to fill out form, will be ready to continue in 5 days
05/15/2024 finally done

### Possible Solution

Many tasks managers from Asana to ERP Next have the ability to put comments in a task by default.

There are plenty of ways to attack this same problem. 

Whether you call them status updates, comments, or organize it some other way, there's key information that needs to be saved so that you can take action easily and keep a task organized.

### Additional context

_No response_

afonso-resendes:
## Summary

Implements timestamped status comments on tasks, as requested in #4412.
Users can add, edit, and delete comments from the task detail panel.

## Changes

- Add `TaskComment` and optional `comments` on `Task` (`packages/plugin-api`)
- `TaskService`: `addComment`, `updateComment`, `deleteComment` via existing `updateTask` (sync/op-log)
- New `TaskCommentsComponent` in the detail panel (between Notes and Attachments)
- Unit tests for service and component
- English strings in `en.json`

## How it works

Comments are stored on the task entity as an ordered list (`id`, `body`, `created`, optional `updated`).
Mutations use immutable array updates and a single `updateTask` dispatch per action.

## Test plan

- [x] Unit tests: `task.service.spec.ts` (comment CRUD), `task-comments.component.spec.ts`, `task-comment.util.spec.ts`
- [x] Manual: add / edit / delete comments in task detail panel
- [x] Manual: comments persist after reload
- [ ] Manual: sync between two clients (if sync enabled)

Closes #4412


johannesjo:
Hey @afonso-resendes, thank you so much for this — you clearly put real care into it, and the code quality is genuinely high: clean utils, good tests, lightweight `updateTask` dispatch. 🙏

Let me be honest before you put in more time, because it's a product call rather than anything about your work. My main hesitation: a dedicated comment thread adds UI clutter and complexity to the task panel for *everyone*, while only a minority will actually use it — and the need it covers is already mostly served by the existing **notes** field (typing dated lines). For a feature like that, my bar is whether there's clear, broad demand, and right now I don't think it's there (the issue has a handful of upvotes). So unless that demand grows a lot, I lean toward keeping the task panel simple and not adding it.

I really don't want to discourage you — the underlying request is real and this is well-built work. I just don't want you polishing something I'm likely not going to merge. What are your thoughts?

suzunn:
I noticed the PR stores comments on the task entity and routes mutations through `updateTask`, which is the right integration point for sync/op-log. Before this ships, I would add one regression test around sync or migration that includes a task with multiple comments, then edits and deletes one comment on a second client. The current unit coverage proves the local service/component behavior, but the unchecked manual sync item is the riskiest part because comment arrays are replacing task state as a whole. A test that verifies comment ordering, `updated` preservation, and last-comment deletion across the serialized task payload would make this much safer if the feature is kept.

afonso-resendes:
Hey @johannesjo! Thanks for the clear feedback. I understand the concerns about panel clutter and overlap with Notes, and agree demand seems limited right now.
I'm not expecting merge if it doesn't fit the product direction, so I'll stop polishing this PR unless you'd like to keep it open for discussion. Happy to close it otherwise.
Appreciate the review.

PR #8276, opened 2026-06-11
feat(idle): redesign idle dialog with choice-first flow

davidvornholt:
## What & why

The idle dialog ("Hey, you've been away for 12 minutes — what should happen with that time?") did a lot at once: it opened with a required, red-highlighted task field even if you just wanted a break, and the split-time option was tucked behind an unlabeled icon button that opened a second nested dialog.

This PR redesigns it around the question the user is actually answering — **"What were you doing?"** — and only shows the controls relevant to that answer.

## The new flow

1. The dialog shows how long you were away and three clearly labeled choices: **Taking a break** / **Working on a task** / **A bit of both**.
2. Pick one, and just the controls for that path expand inline:
   - **Break** — a one-line summary of what will be tracked. Nothing to fill in.
   - **Task** — the task selector appears (and is only *required*) here, so there's no spurious red "required" state when the dialog opens.
   - **A bit of both** — split rows expand inline instead of a nested dialog, with plain duration inputs, an "assign remaining time" affordance, and a live unassigned/over-assigned indicator (minute-tolerant, since the idle clock keeps ticking while you decide).
3. Confirm with verb-labeled buttons that say what will happen: *Don't track*, *Track as break*, *Track to task*, *Create task & track*, *Track all*.

## Small touches

- Simple-counter toggles (e.g. "was I standing?") are available **before** picking a mode, since they apply to "Don't track" too; in split mode each row gets its own copies.
- Your input survives changing your mind: a typed new-task title or selected task carries over when switching between modes, including into split rows (as long as you haven't entered times yet).
- Choosing a break-containing option auto-checks **"Reset break reminder timer"**, and a manual override sticks. Skipping via *Don't track* only resets the timer if you explicitly checked the box — the auto-default alone no longer counts as a user choice.
- Saving a split is disabled until every row has a time and task rows have a task, so empty no-op entries can't be created.
- New-task titles are trimmed, and whitespace-only titles don't enable the confirm button.
- The dialog stacks vertically on narrow screens (now via the shared `LayoutService` instead of a local `BreakpointObserver`).

## Implementation notes

- The dialog result contract (`trackItems` etc.) is unchanged, so `idle.effects.ts` and the sync path are untouched.
- `TakeABreakService` now treats the dialog's reset checkbox as the single source of truth instead of inferring a reset from the presence of a BREAK item. Defaults are preserved (break → checkbox auto-checked → reset); what's new is that an explicit opt-out is respected. The effect already uses `LOCAL_ACTIONS`; new unit tests cover the reset/no-reset matrix.
- `select-task` gains an optional `initialTitle` input to seed plain text, used for carrying a typed title across mode switches.

## Docs

`docs/wiki/4.16-Break-Reminders.md` and `4.17-Idle-Time.md` (Concepts pages) describe the old "assigning idle to a break always resets the break timer" behavior. That's still the default, but there's now an opt-out checkbox — flagging here per the documentation guide rather than rewriting Concepts prose; happy to add a minimal mention if preferred.

## Testing

- New unit tests for the dialog (title trimming, mode re-entry seeding, reset-checkbox semantics) and `TakeABreakService` (reset/accumulate matrix) — 22 + 7 specs green, plus the `select-task` spec (6).
- `npm run checkFile` clean on all touched files; rebased on latest `master`.


johannesjo:
Thank you very much! That is a very nice improvement! <3

Issue #518, opened 2020-08-31
Drag'n'drop .eml to + adds task with email title


sdruskat:
### Your Environment

* Version used: future version
* Operating System and version: all if possible

### Expected Behavior

An idea for UI behaviour: dragging and dropping an email from the respective application onto the `+` symbol to add tasks could add a task with the pattern `<from address/name>: <subject line>` as task title, and attach the .eml to the task.

### Current Behavior

You can "only" attach the .eml file as an attachment to an existing task.

Once I have time to work on SP in general (e.g., on #496), I'd be happy to look into implementing this as well, if you think this is a good idea.

johannesjo:
Hey there! That's a cool suggestion. Thank you!

alexanderbischof:
This would be very helpful.

PR #8282, opened 2026-06-12
feat(tasks): add task through eml file hover

yontank:
## Problem

as written in #518,  When an EML  file is dropped at at the taskAdd button (+), 
it should parse the eml file and add it as a task.

while this is a working POC I'd like to make sure I didn't miss anything

1. the issue format has said <from>:<subject> but what should we do when there's a-lot of from email addresses?
2.  I haven't seen a logger, what should i do if the eml parsing or task creation fails?
3. have I missed anything else?

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Hi @yontank, thanks for picking up #518 — nice first PR, and you asked exactly the right questions. Here's a round of feedback.

## Your questions

1. **Multiple From addresses** — just take the first one, that's fine (KISS). Falling back to the email address when there's no display name matters more (see below).
2. **Logging / errors** — we have `Log` (`src/app/core/log`): use `Log.err(e)` for the parse failure, plus a `SnackService` snack so the user learns the file couldn't be parsed. Please don't leave the catch empty. One project rule to keep in mind: log output is exportable, so never log user content (subject/sender).
3. **What's missing** — main points below.

## Main points

**1. Move the logic out of `MainHeaderComponent`.** The header is a layout shell; every other drop site in the app hands off immediately to a service (see `task-detail-panel.component.ts`, `task.component.ts`, `notes.component.ts` → `createFromDrop`). Suggested shape:

- a standalone attribute directive (e.g. `eml-drop.directive.ts` near `src/app/core/drop-paste-input/`) with `@HostListener('drop')`, applied as an attribute on the `+` button — the header component stays untouched and the directive is reusable (e.g. for the mobile add button later)
- the parse → `taskService.add()` logic in a small service/util the directive injects, so it can be unit-tested without the header

**2. The dependency.** `eml-parse-js@1.2.0-beta.1` is a beta pinned with `^`, ships in the production bundle, and pulls in the deprecated `@sinonjs/text-encoding`. Please switch to something stable — `postal-mime` (maintained, browser-native, zero deps) looks like a good fit since we only need From + Subject.

**3. Crash / format bugs in the title building:**

- `from === null ? '' : from.name` throws when `data.from` is `undefined` (no From header, or an empty array) and yields `undefined` for address-only senders. Use `from?.name || from?.email || ''` — that also gives the "address/name" fallback the issue asks for.
- In the issue, `<from address/name>: <subject line>` uses `<>` as placeholders — the title should be `` `${sender}: ${subject}` ``, without a literal `<from …>` and without `\n` (task titles are single-line). Also guard against `data.subject` being `undefined`, which currently ends up as the string "undefined".

**4. Smaller things:**

- `isFileEml`: `endsWith('eml')` matches any name merely ending in those letters (no dot required) — use `file.name.toLowerCase().endsWith('.eml')`.
- `eml-parser.service.ts` isn't an Angular service (no `@Injectable`) — name/place it as a util instead.
- `taskAttachmentService` is injected but only referenced in commented-out code — drop it for now.
- Please add unit tests for `isFileEml` / `parseEml` — missing From, From as array, and missing subject are exactly the cases worth pinning down.
- package-lock: the unrelated `fsevents` `"dev": true` flip suggests a different npm version — please regenerate so the lock diff stays minimal.

**5. The attachment TODO:** fine to descope to a follow-up issue. `taskService.add()` already returns the new task id, but attachments are path/link based, so persisting the dropped file's *content* doesn't work on web anyway — no need to block this PR on it.

One heads-up on scope: dragging mails out as `.eml` only works from desktop clients (Thunderbird, Apple Mail, …) — Gmail/Outlook web can't drag a message out as a file. Still worth having; just keep the footprint small as outlined above.

Thanks again — looking forward to the next iteration 🙏

yontank:
Thank you for the help, I'm pretty new to Angular so I'll try to catch with the main ways it works,

let me know if I missed anything🙌

johannesjo:
### Review of the refinement range (`0b0c4a3..c5d377c`)

Overall this is a clean, well-tested, well-scoped refactor — swapping the beta `eml-parse-js` (which drags in the deprecated `@sinonjs/text-encoding`) for stable zero-dep `postal-mime`, extracting the drop logic into a directive + service + pure helpers, and adding good test coverage. 👍

**Verified (no action needed):**
- **Drop event fires correctly** — the directive has no `dragover` handler, but `app.component.ts` registers a global `dragover` → `preventDefault()`, which is what's required for `drop` to fire.
- **Test assumptions are real** — I installed `postal-mime@2.7.4` and ran all four spec fixtures through it; outputs match every assertion exactly, including "multi-`From` takes the first address" and "missing header → `undefined`".
- **Types line up** — `from?: Address`, `subject?: string`, `static parse(...)`, so the optional chaining and `spyOn(PostalMime, 'parse')` are all sound. No dangling refs to the deleted service; `en.json` valid and keys correctly ordered.
- `taskService.add()` stores the title verbatim (no short-syntax parsing), so the untrusted subject/sender is no more dangerous than a hand-typed title on a self-supplied file — no new injection surface.

**Findings (all minor):**

1. **Logs user content** — `parseEml` wraps the error as `Could not parse EML file "${file.name}": ...` and `EmlDropService` logs it via `Log.err(e)`. Filenames can carry PII and the log history is exportable (the repo has a strict "never log user content" rule). The user already gets the generic translated snack, so the filename only lives in the log. Suggest `Log.err('Failed to parse EML file')` without the filename — which also makes the error-wrapping in `parseEml` mostly redundant.

2. **Awkward title when one field is empty** — `` `${sender}: ${subject}` `` produces a leading `": Subject"` when there's no `From`, or trailing `"Sender: "` when there's no subject. Consider `[sender, subject].filter(Boolean).join(': ')`.

3. **Small consistency nits** — in `eml-drop.service.ts`, `private readonly taskService` lacks the `_` prefix used by `_snackService` (and the dominant `_taskService` convention); and there's a redundant trailing `return;` in `createTaskFromEml`.

4. **`package-lock.json`: `fsevents` lost its `"dev": true` flag** — unrelated to postal-mime (which has no deps), looks like a lockfile-regeneration artifact (possibly a different npm version). Worth a glance that it doesn't cause `npm ci` churn. Low severity.

5. **Optional robustness** — `file.text()` decodes as UTF-8; postal-mime's more robust input is `file.arrayBuffer()`. Since only the `from`/`subject` headers are read, this is fine as-is — note only.

Nothing here is blocking; #1 is the only one I'd actively recommend changing.

johannesjo:
Another solid iteration — everything from the last round is addressed, and I re-checked the drop mechanics end-to-end (the global `dragover`/`drop` `preventDefault()` in `app.component.ts` is what lets the `drop` fire on the button; your `stopPropagation()` scopes it cleanly, no clash with the existing task/note attachment drop handlers). Tests cover exactly the tricky cases. 🙏

Just one thing left that I'd actually recommend changing, plus a couple of optional notes.

**1. Lazy-load `postal-mime` (the one real point).** It's a static import in `eml-parser.ts`, and the directive/service are wired into `main-header`, which is part of the always-loaded app shell — so the whole MIME parser ends up in the **initial bundle** for a niche, desktop-only feature. Defer it so it's only fetched when a file is actually dropped:

```ts
export const parseEml = async (file: File): Promise<Email> => {
  const content = await file.text();
  const { default: PostalMime } = await import('postal-mime');
  return await PostalMime.parse(content);
};
```

Good news: this is test-compatible — `spyOn(PostalMime, 'parse')` targets the same ES-module singleton the dynamic import resolves, so the specs keep passing as-is.

**2. Docs (minor).** This adds a new user-facing surface and dropping onto the small `+` button isn't very discoverable — a one-line note in the [wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki) would help. The docs checklist box is still unchecked.

**3. Optional nits — take or leave:**
- `catch {` discards the error entirely. The *type/message* from `postal-mime` isn't user content and would help debugging a parse failure — `catch (e) { Log.err('Failed to parse EML file', e); … }`. Your safe choice is fine too.
- Only `files[0]` is used, so dropping several `.eml`s silently makes one task — fine for KISS, just noting it.

Do the dynamic import and I think this is good to merge. Thanks again, nice work for a first Angular PR! 🎉

johannesjo:
Thank you very much for htis! <3

PR #8291, opened 2026-06-12
fix(plugins): expose focused task API to iframe plugins

Meh-S-Eze:
## Summary

- Adds `getSelectedTask()` and `getFocusedTask()` to the public Plugin API contract and generated iframe `PluginAPI` object.
- Implements those readers in `PluginBridgeService` using the existing selected-task stream and `TaskFocusService`.
- Exposes both methods through iframe plugin bound methods so side-panel iframe plugins can read host task focus without calling `selectTask()`.
- Adds regression coverage for the direct PluginAPI wrapper, bound bridge methods, and iframe `API_CALL` routing.

## Root Cause

Iframe plugins call host APIs through the generated iframe bridge and the per-plugin `boundMethods` object. Task selection/focus state already exists in the host, but iframe plugins did not have a complete read-only path for the selected or focused task. Side-panel plugins therefore had to either guess from context tasks or call `selectTask()`, which changes the native task detail panel instead of simply reading focus.

## Validation

- `npm run checkFile packages/plugin-api/src/types.ts`
- `npm run checkFile src/app/plugins/plugin-api.ts`
- `npm run checkFile src/app/plugins/plugin-api.spec.ts`
- `npm run checkFile src/app/plugins/plugin-bridge.service.ts`
- `npm run checkFile src/app/plugins/plugin-bridge.service.spec.ts`
- `npm run checkFile src/app/plugins/util/plugin-iframe.util.ts`
- `npm run checkFile src/app/plugins/util/plugin-iframe.util.spec.ts`
- `npm run test:file src/app/plugins/plugin-api.spec.ts -- --watch=false`
- `npm run test:file src/app/plugins/plugin-bridge.service.spec.ts -- --watch=false`
- `npm run test:file src/app/plugins/util/plugin-iframe.util.spec.ts -- --watch=false`
- `npm run test:file src/app/plugins/plugin-hooks.effects.spec.ts -- --watch=false`
- `npm run test:file src/app/features/tasks/task/task-on-focus-panel-sync.spec.ts -- --watch=false`
- Pre-push hook completed successfully, including lint, package tests, release-note tests, full Angular unit suite, and timezone CI.

Note: lint reports one pre-existing warning in `src/app/features/simple-counter/habit-tracker/habit-tracker.component.scss` for an empty block, but exits successfully.


johannesjo:
Thanks for the PR, and welcome! 🎉 This is a clean, well-scoped addition — the wiring is complete across every iframe touchpoint (typed bound-method signature, `ALLOWED_IFRAME_API_METHODS`, the generated `callApi` stub, the bound methods, and the `PluginAPI` wrapper), the tests cover all three layers (wrapper delegation, bound bridge method, and iframe `API_CALL` → `API_RESPONSE` routing), and there's no new security surface — these readers expose a strict subset of data already reachable via `getTasks`/`getCurrentContextTasks`. Good JSDoc too.

A couple of things to address before merge, plus a design question worth a quick discussion.

### Important

**1. iframe plugins get `undefined` instead of `null` for a stale focused id**

`getFocusedTask()` in `plugin-bridge.service.ts` does:

```ts
return firstValueFrom(this._taskService.getByIdOnce$(focusedTaskId).pipe(take(1)));
```

`getByIdOnce$` selects via `selectTaskById`, which is `state.entities[id] as Task` — so if the focused task was deleted/archived since it was focused, this resolves to **`undefined`**, not `null`.

The `PluginAPI` wrapper masks it (`return task ? taskCopyToTaskData(task) : null`), but **iframe plugins bypass that wrapper** — `handlePluginMessage` calls the bound method directly and posts the raw result (`plugin-iframe.util.ts:694-708`). So:
- `plugin.js` (host-side) plugins → `null` ✅
- iframe plugins (this PR's actual target) → `undefined` ❌

for the identical host state. The declared contract is `Promise<Task | null>`. Please coerce inside the bridge so both paths agree — this mirrors the existing `getActiveWorkContext()` precedent, which does its mapping/coercion in the bridge method rather than the wrapper:

```ts
async getFocusedTask(): Promise<TaskCopy | null> {
  const focusedTaskId = this._taskFocusService.focusedTaskId();
  if (!focusedTaskId) return null;
  const task = await firstValueFrom(this._taskService.getByIdOnce$(focusedTaskId));
  return task ?? null;
}
```

A regression test for the stale-id branch would be good — right now `getByIdOnce$` is mocked to always return a valid task, so the `undefined` path is untested.

### Design question

**2. Does `getFocusedTask()` actually return useful data for side-panel iframe plugins?**

`focusedTaskId` is cleared on `focusout` (`task.component.ts:408`). The moment a user clicks into a side-panel iframe plugin, the host task fires `focusout` and `focusedTaskId` becomes `null`. So when a side-panel plugin reacts to its own user interaction and calls `getFocusedTask()`, it will almost always get `null`.

It's still valid when called programmatically (a hook/timer) while the task list holds DOM focus — but for the "which task is the user on?" flow the PR motivates, `getSelectedTask()` (detail-panel selection, which is **not** cleared by iframe focus) is the reliable primitive. Could you either tighten the JSDoc to note that focus is cleared once the plugin UI gains focus (and point authors at `getSelectedTask()` for the persistent case), or reconsider whether `getFocusedTask()` earns its place in the contract?

### Minor

**3. `getSelectedTask()` ships an undeclared `subTasks` field.** `selectedTask$` is `TaskWithSubTasks` (`subTasks: Task[]`), and `taskCopyToTaskData` is a passthrough cast — so the full nested-task array is sent to plugins even though the plugin-api `Task` type only declares `subTaskIds`. This is consistent with `getCurrentContextTasks()` (also `TaskWithSubTasks[]`) but inconsistent with `getTasks()`/`getArchivedTasks()` (flat `Task[]`). For a brand-new method I'd lean toward stripping it in the bridge (`const { subTasks, ...rest } = task; return rest;`) so the runtime shape matches the contract — your call given the existing split.

**4. Docs + version.** The new methods aren't added to the read-API lists in `packages/plugin-api/README.md`, `docs/plugin-development.md`, and `docs/wiki/3.01-API.md` (two bullets each), and `packages/plugin-api/package.json` is still `1.0.1` (additive change → minor bump). If version bumps happen at release time, ignore the latter.

**5. Nit.** `.pipe(take(1))` is redundant — `getByIdOnce$` already applies `take(1)` and `firstValueFrom` completes on first emit. Harmless, optional cleanup.

Overall really solid for a first contribution — item 1 is the only hard blocker, item 2 just needs a doc sentence or a decision. Thanks again!

Meh-S-Eze:
Thanks, this all makes sense. I pushed an update addressing the review points.

Changes made:
- `getFocusedTask()` now normalizes stale focused task ids to `null` instead of leaking `undefined`.
- `getSelectedTask()` and `getFocusedTask()` now strip `subTasks` before returning through the iframe bridge, so the runtime shape matches the plugin API `Task` type.
- Removed the redundant `take(1)`.
- Added docs entries in `packages/plugin-api/README.md`, `docs/plugin-development.md`, and `docs/wiki/3.01-API.md`.
- Clarified in the API docs that `getFocusedTask()` is transient and can be cleared when focus moves into an iframe side panel, while `getSelectedTask()` is the stable reader for persistent side-panel context.

I did not bump the plugin-api package version, assuming that should stay release-time controlled unless you prefer it in this PR.

Validation run locally:
- `npm run checkFile src/app/plugins/plugin-bridge.service.ts`
- `npm run checkFile src/app/plugins/plugin-bridge.service.spec.ts`
- `npm run checkFile packages/plugin-api/src/types.ts`
- `npm run test:file src/app/plugins/plugin-bridge.service.spec.ts`
- `npm run test:file src/app/plugins/util/plugin-iframe.util.spec.ts`
- `npm --prefix packages/plugin-api run build`
- pre-push hook: full lint/test suite passed

Issue #8267, opened 2026-06-11
💡 Enhance UI/UX: Reduce visual clutter, simplify task creation, and move advanced options to settings


AhmMed29:
### Problem Statement

I am writing to share some feedback regarding the current user interface. While Super Productivity is incredibly powerful, the design feels highly distracting and overwhelming to navigate. As someone with attention challenges (like ADHD), the sheer amount of immediate visual data and options makes it very difficult to focus on getting simple tasks done.
The Problem

When trying to use a simple tool like the Pomodoro timer, the app demands too much input upfront—asking for "Task to achieve", "Time Period", and "Goal" all at once. For comparison, apps like TickTick offer a frictionless experience: you just open it, click start, and focus.

The current design forces too many choices on the main screen, which destroys focus rather than helping it. Having so many options scattered here and there causes heavy confusion, especially when I am just here to run some simple pomos, set quick tasks, or check off a habit.
Proposed Solution / Feature Request

    Simplify the Main Interface: Hide advanced fields (Goals, Time Periods, etc.) by default during quick task creation or Pomodoro setup.

    Group Extra Options: These scattered buttons and advanced options should be grouped inside the settings panel instead of being visible on the main dashboard.

    User-Defined Customization: Move these extra options into the application settings, allowing users to toggle them on or off based on their personal workflow needs. Let the user decide exactly how much information they want to see.

I truly appreciate your hard work on this project. An application like this requires a huge amount of effort and time to build. I hope you take this feedback into consideration to make the app even better. Thank you for your time, and I apologize if my approach felt a little harsh!

here  are some imgs that might help 

<img width="413" height="1002" alt="Image" src="https://github.com/user-attachments/assets/6faf6cb7-79b8-4a46-9fff-7393b01bb3fa" />

<img width="1920" height="1078" alt="Image" src="https://github.com/user-attachments/assets/9f2eca6f-3a41-42eb-a050-b3a6446fc5f6" />

### Possible Solution

I wish you would redesign it... The Tick Tick app design is too simple and straightforward.

The Pomodoro Technique fills the screen and is easy to use. It also counts the hours you've worked so far, which is a great motivator.

so please please add simple design and hide some options.

### Describe alternatives you've considered

Tick Tick:
I highly recommend looking at TickTick's approach. Their Pomodoro design is simple, straightforward, and fills the screen in a way that makes it easy to focus. It also tracks cumulative focus hours, which serves as a great motivator without adding visual clutter.

[...]

PR #8292, opened 2026-06-12
style(ui): globalize task-detail field style to filled form fields

hajiboy95:
## Problem

Tackles the first point of https://github.com/super-productivity/super-productivity/issues/8267#issuecomment-4678414966

The filled form fields in the rest of the application (settings, customizers, dialogs, etc.) were flat, gray, underlined textboxes, which visually clashed with the modern, card-like "hovering container" design used in the task-detail side panel.

## Solution

Overrode the `.mdc-text-field--filled` styles globally inside `src/styles/components/_overwrite-material.scss` to match the side panel's style:
- Set field backgrounds to use `var(--task-detail-bg)` and rounded corners to use `var(--card-border-radius)`.
- Applied a soft elevation shadow using `var(--whiteframe-shadow-1dp)`.
- Hidden default Material underlines (`.mdc-line-ripple`).
- Set hover and focus transitions that shift background to `var(--task-detail-bg-hover)` and border highlights to `var(--focus-ring-color)`.

These overrides are theme-aware, ensuring they adapt correctly to light, dark, Velvet, and Zen (which naturally hides cards/shadows) themes.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [x] Other: Styling globalization / Visual polish

## Checklist

## Before

telegram-cloud-photo-size-2-5290009118980643804-y

## After 

telegram-cloud-photo-size-2-5290009118980643803-y

### DIFF
--- a/src/assets/themes/velvet.css
+++ b/src/assets/themes/velvet.css
@@ -183,7 +183,7 @@ body.isDarkTheme {
   --input-bg: var(--velvet-fill-2);
   --input-bg-hover: color-mix(in srgb, var(--c-primary) 6%, transparent);
   --input-border-color: transparent;
-  --input-border-color-focus: color-mix(in srgb, var(--c-primary) 40%, transparent);
+  --input-border-color-focus: var(--c-primary);
   --input-border-radius: 8px;
 
   /* Schedule events read as cards, not list rows — they have borders, a
@@ -323,6 +323,22 @@ body.isDarkTheme add-task-bar.global .add-task-container {
     0 24px 64px rgba(0, 0, 0, 0.65),
     0 6px 16px rgba(0, 0, 0, 0.4),
     inset 0 1px 0 var(--velvet-fill-2);
+}
+
+/* Form fields outline: restore fill for outline inputs in Velvet */
+body.isDarkTheme .mdc-text-field--outlined:not(.mdc-text-field--disabled) {
+  background: var(--velvet-fill-2) !important;
+  border-radius: 8px !important;
+  transition:
+    background 120ms ease-out,
+    border-color 120ms ease-out !important;
+}
+
+body.isDarkTheme .mdc-text-field--outlined:not(.mdc-text-field--disabled):hover,
+body.isDarkTheme
+  .mat-mdc-form-field.mat-focused
+  .mdc-text-field--outlined:not(.mdc-text-field--disabled) {
+  background: color-mix(in srgb, var(--c-primary) 6%, transparent) !important;
 }
 
 /* Primary buttons get a subtle coral gradient — like Raycast's CTAs */
--- a/src/assets/themes/velvet.css
+++ b/src/assets/themes/velvet.css
@@ -183,7 +183,6 @@ body.isDarkTheme {
   --input-bg: var(--velvet-fill-2);
   --input-bg-hover: color-mix(in srgb, var(--c-primary) 6%, transparent);
   --input-border-color: transparent;
-  --input-border-color-focus: var(--c-primary);
   --input-border-radius: 8px;
 
   /* Schedule events read as cards, not list rows — they have borders, a
--- a/src/assets/themes/velvet.css
+++ b/src/assets/themes/velvet.css
@@ -179,6 +179,13 @@ body.isDarkTheme {
   --task-detail-shadow: none;
   --task-detail-value-color: rgba(237, 237, 237, 0.85);
 
+  /* Input fields overrides */
+  --input-bg: var(--velvet-fill-2);
+  --input-bg-hover: color-mix(in srgb, var(--c-primary) 6%, transparent);
+  --input-border-color: transparent;
+  --input-border-color-focus: color-mix(in srgb, var(--c-primary) 40%, transparent);
+  --input-border-radius: 8px;
+
   /* Schedule events read as cards, not list rows — they have borders, a
      project-color left edge, and distinct shapes — so they sit a few stops
      above the task-row default. Side-panel variant stays one stop lower
@@ -316,33 +323,6 @@ body.isDarkTheme add-task-bar.global .add-task-container {
     0 24px 64px rgba(0, 0, 0, 0.65),
     0 6px 16px rgba(0, 0, 0, 0.4),
     inset 0 1px 0 var(--velvet-fill-2);
-}
-
-/* Form fields: dark, rounded, with coral focus glow — Raycast input feel */
-body.isDarkTheme input.mat-mdc-input-element,
-body.isDarkTheme textarea.mat-mdc-input-element,
-body.isDarkTheme .mdc-text-field--filled:not(.mdc-text-field--disabled) {
-  background: var(--velvet-fill-2) !important;
-  border-radius: 8px !important;
-  transition:
-    background 120ms ease-out,
-    box-shadow 120ms ease-out !important;
-}
-
-body.isDarkTheme
-  .mat-mdc-form-field.mat-focused
-  .mdc-text-field--filled:not(.mdc-text-field--disabled),
-body.isDarkTheme .mdc-text-field--focused:not(.mdc-text-field--disabled) {
-  background: color-mix(in srgb, var(--c-primary) 6%, transparent) !important;
-  box-shadow: inset 0 0 0 1px color-mix(in srgb, var(--c-primary) 40%, transparent) !important;
-}
-
-/* Hide the underline-style line-ripple — Raycast inputs are bordered, not
-   underlined. */
-body.isDarkTheme .mdc-line-ripple,
-body.isDarkTheme .mdc-text-field--filled .mdc-line-ripple::before,
-body.isDarkTheme .mdc-text-field--filled .mdc-line-ripple::after {
-  display: none !important;
 }
 
 /* Primary buttons get a subtle coral gradient — like Raycast's CTAs */
--- a/src/assets/themes/zen.css
+++ b/src/assets/themes/zen.css
@@ -206,17 +206,17 @@ body.isDarkTheme .mat-mdc-card {
   border-color: rgba(255, 255, 255, 0.1);
 }
 
-/* Material form fields: use main bg with darker underline */
-body .mdc-text-field--filled:not(.mdc-text-field--disabled) {
-  background: var(--bg) !important;
+/* Material form fields: flat border instead of invisible input */
+body {
+  --input-bg: transparent;
+  --input-bg-hover: rgba(0, 0, 0, 0.04);
+  --input-border-color: rgba(0, 0, 0, 0.2);
 }
 
-body .mdc-line-ripple::before {
-  border-bottom-color: rgba(0, 0, 0, 0.3) !important;
-}
-
-body.isDarkTheme .mdc-line-ripple::before {
-  border-bottom-color: rgba(255, 255, 255, 0.3) !important;
+body.isDarkTheme {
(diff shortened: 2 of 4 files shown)


johannesjo:
Thanks for tackling this — the underlying complaint in #8267 is real, and I agree with the goal that form fields should look consistent everywhere (scoping per-context would just trade inconsistency for complexity). But I think this specific approach has design and theming implications worth working through before it lands, because the chosen mechanism — an elevated, card-like field driven by absolute color + shadow tokens — fights against both the app's nested layouts and two of the four shipped themes.

## The core tension: elevation can't be globally consistent

A drop shadow / raised surface only reads correctly *relative to whatever is behind it*. The same elevated field renders three different ways depending on where it sits:

- on the flat page background → barely distinguishable
- on a white card or inside a dialog → the fill matches the container, so only the shadow is left
- on the dark side panel (where this style was born) → looks great

So "elevated + global + consistent" is self-contradictory — you can pick any two. That's the root cause of both issues below; they're really the same issue.

## Light theme loses the fill

Tracing the tokens in light theme:

- field background → `--task-detail-bg` → `--card-bg` → `--surface-2` = **`#fff`**
- page background → `--surface-1` = **`#f8f8f7`**

So the field goes from Material's visible tonal-gray fill to **white on near-white**, distinguished only by the faint 1dp shadow. Inside dialogs and cards (surface is also `#fff`) the field background equals the container background, so the field is effectively invisible apart from the shadow. Dark theme works only because `--task-detail-bg` (`--dark8`, `rgb(46,46,46)`) happens to be *lighter* than the `#131314` page — i.e. the design is tuned for dark and light is the casualty.

## Card-in-card hierarchy

Every filled field now carries a `box-shadow`. Elevation is supposed to signal hierarchy, so when fields inside dialogs, settings cards, and the task-detail panel (itself a rounded card) each get their own shadow, you get shadow-on-shadow and the hierarchy reads as muddy. The look only works when a field sits *directly on a flat surface* — which is precisely the side-panel case it was lifted from. (Worth noting: the side panel's nice "hovering containers" are bespoke `.input-item` blocks, not `mat-form-field`s — so the lovely effect there doesn't actually transfer to real Material inputs in other contexts.)

## Theming — this is the biggest one

The PR description says the overrides "adapt correctly to light, dark, Velvet, and Zen." In practice the opposite happens for the two custom themes, because they redefine the very card/shadow language this PR depends on:

- **Zen** (`zen.css`) sets `--task-detail-bg: transparent`, `--card-bg: transparent`, and force-removes shadows (`--whiteframe-shadow-1dp: none !important`, blanket `box-shadow: none`). Since this PR fills the field with `--task-detail-bg`, leans on the 1dp shadow, *and* hides the Material underline, a field under Zen ends up with **transparent fill + no shadow + no underline = an invisible input with no affordance at all**. Zen "naturally hides cards/shadows" is exactly *why* a card-based field can't work there — it strips away everything the style relies on and leaves nothing.

- **Velvet** (`velvet.css`, the `body.isDarkTheme … .mdc-text-field--filled:not(.mdc-text-field--disabled)` rule) already styles the *identical* selector with `!important`, including its own rounded fill and coral focus glow. This PR introduces a second global `!important` rule on the same selector, so the two now compete (and the PR's focus `box-shadow` ring partially overlaps Velvet's `inset` focus ring). That's a fragile collision and the kind of `.mdc-*` override the contributor guidelines specifically warn against.

The deeper lesson: form-field affordance shouldn't be hardcoded to the card/elevation tokens, because themes legitimately treat those tokens completely differently.

## Suggested direction — flat, token-driven, still 100% global

Keep the parts of the redesign that are genuinely nicer and theme-neutral, and drop the part that breaks:

- **Keep:** rounded corners and the hidden underline ripple (the modern bit people want), plus a focus highlight.
- **Drop:** the `box-shadow` entirely — that's the card-in-card and Zen-conflict cause.
- **Change the fill:** instead of the absolute `--card-bg`/`--task-detail-bg`, use a **surface-relative translucent overlay** (the codebase already ships `--c-dark-*` / `--c-light-*` and an `--overlay-alpha` system). A translucent fill is always slightly offset from whatever's behind it — white page, white card, or dark panel — so it stays visible at every nesting level without needing a shadow. (This is exactly why Material's default filled field uses a tonal overlay rather than a solid color.)
- **Focus:** use `border-color` / a thin `outline` rather than a `box-shadow` ring — a border doesn't stack like elevation does.
- **Expose it as a token** (e.g. `--input-bg` / `--input-border`) so each theme re-skins fields by overriding one variable instead of fighting `!important`. Velvet keeps its coral glow with no collision; Zen keeps it transparent (or chooses a hairline border) on its own terms.

This stays fully global (consistent everywhere, which is the goal), removes most of the `!important`, and is genuinely theme-aware rather than theme-stomping. The only thing it gives up is the "lifting off the page" delight — which is the exact thing that doesn't survive leaving the side panel. A flat tonal-fill input with a clean focus ring is still a modern look (Material 3 / Linear / GitHub inputs are all flat-fill, no drop shadow).

Happy to pair on the token version or review a follow-up.

hajiboy95:
Thanks for the excellent and detailed feedback @johannesjo —you hit the nail on the head regarding the card-in-card collision, the light theme contrast issues, and the themes breaking. Hardcoding elevation and solid background colors indeed fought against nested layouts.

I have updated the PR to transition the global `.mdc-text-field--filled` inputs to a flat, translucent, and fully token-driven design:

1. **Dropped the Elevation:** The `box-shadow` styles have been removed entirely from both default and focus states globally, solving the card-in-card hierarchy mud.
2. **Surface-Relative Translucent Fill:** Instead of absolute card backgrounds, inputs now use a subtle translucent overlay: `rgba(var(--ink-on-channel), 0.04)` in light mode and `rgba(var(--ink-on-channel), 0.06)` in dark mode. Because `--ink-on-channel` switches with the theme context, it naturally offsets and stays readable at any nesting level without needing shadows.
3. **Flat Border Focus Ring:** Focus transitions are now completely flat, using `border-color` outlines instead of stacking box-shadow glows.
4. **Clean Tokenization:** All values are exposed as tokens (`--input-bg`, `--input-bg-hover`, `--input-border-color`, `--input-border-color-focus`, `--input-border-radius`) at the base layer.
5. **Theme Support:**
   - **Zen:** Overrides the tokens to remain transparent but adds a soft hairline border (`--input-border-color: rgba(..., 0.2)`) so fields remain visible and have affordance.
   - **Velvet:** All conflicting selector rules were removed. Velvet now overrides the input tokens inside its main block to cleanly apply its custom rounded corners and coral focus glows without selector collisions.

#### Aligning Overall Modal & Popup Rounding
Additionally, we noticed that Material 3 overlays (dialogs, dropdown menus, autocompletes, and select panels) had conflicting or excessively bubbly roundings (like the default M3 `28px` on dialogs) compared to the rest of the application's geometric look. 

We have aligned them to the general theme shapes:
- **Dialogs (`.mat-mdc-dialog-surface`):** Mapped to `var(--radius-lg)` (12px, soft but clean).
- **Menus, Autocompletes, and Selects:** Mapped to `var(--radius-sm)` (6px) to match small cards and inputs.

This ensures all modals, dropdowns, and folders look cohesive and fit the overall design system

johannesjo:
Great iteration — this nails the round-1 feedback. I traced the cascade across all themes and confirmed: box-shadow is gone, the fill is now a surface-relative `rgba(var(--ink-on-channel), α)` tonal overlay (so it self-offsets on white cards *and* the dark panel — the round-1 light-theme invisibility is genuinely fixed), focus is a flat `border-color`, everything is token-driven, and the Velvet `!important` collision + Zen invisible-input are both resolved by token overrides. The token split is principled too (per-mode literals split, mode-invariant aliases shared). 👍

Two things to sort before this lands, plus a couple of minor notes:

### 1. (blocking) Hiding the underline globally removes the error-state cue on every form

`.mdc-line-ripple { display: none }` removes the element Material turns red for invalid filled fields, and there's no `mat-form-field-invalid` border anywhere in base/zen/velvet — so across the whole app an invalid field now signals *only* via the helper/error text + label, not on the field box itself. Default appearance is `fill` (there's no `MAT_FORM_FIELD_DEFAULT_OPTIONS` provider), so this hits the large majority of fields.

The clean fix is to hide the underline through Material's own knob instead of `display:none`: set `--mat-form-field-filled-active-indicator-color` / `-hover-active-indicator-color` / `-focus-active-indicator-color` to `transparent` and leave the **error** indicator color intact — that preserves the red invalid cue for free. (Or, if you keep `display:none`, add an explicit `&.mat-form-field-invalid .mdc-text-field--filled { border-color: var(--palette-warn-500) }`, ideally as an `--input-border-color-invalid` token so themes can re-skin it like the others.)

### 2. (blocking) Please split out the dialog/menu/select radius commit — it's out of scope and it regresses liquid-glass

That last commit isn't part of this PR's goal, and because it's raw `!important` on `.mat-*` panels it collides with `liquid-glass.css`, which already owns those surfaces:

- `body .mat-mdc-dialog-surface` (specificity 0,1,1) **out-ranks** liquid-glass's `.mat-mdc-dialog-surface` (0,1,0) → liquid-glass dialogs silently drop from `--lg-radius-lg` (20px) to 12px.
- `.cdk-overlay-pane .mat-mdc-autocomplete-panel` (0,2,0) out-ranks liquid-glass's autocomplete (0,1,0) → 20px → 6px.
- Menu + select *tie* on specificity, so liquid-glass keeps its 20px there.

Net: two of the four panel types get clobbered and two don't — exactly the fragile `!important`/specificity fight we were trying to get away from. If you want the rounding alignment, let's do it in its own PR and drive it through the official shape tokens (`--mat-dialog-container-shape`, `--mat-menu-container-shape`, `--mat-autocomplete-container-shape`) so themes can opt out with no `!important`. (`select` has no shape token in v21, so that one would stay raw.)

### Minor (non-blocking)

- **Light-mode rest boundary:** base light is a 4% fill on a `transparent` border ≈ ~1.05:1 against a white card — barely a boundary until hover/focus. Zen and Velvet both add a hairline; consider giving base light `--input-border-color` a faint value (e.g. `rgba(0, 0, 0, 0.08)`) rather than fully transparent.
- **Focus ring strength:** it's a 1px border, while our convention (`--focus-ring-width`) is 2px, and Velvet's focus token is a 40%-alpha mix on 1px — likely under the 3:1 focus-appearance bar. Consider 2px / an opaque focus color.
- **Velvet outline fields:** the removed Velvet rule also targeted bare `input/textarea.mat-mdc-input-element`, so `appearance="outline"` fields (13 of them) lose the fill they had — worth an eyeball in Velvet.
- **Nit:** `--input-border-radius` resolves to `--card-border-radius`, and Velvet overrides it to `8px`, which is the same value `--radius-md` already gives — currently a no-op indirection you could inline.

One idea that ties #1 and #2 together: have your `--input-*` tokens *feed* the official Material vars (`--mat-form-field-filled-container-color: var(--input-bg)`, etc.) rather than overriding `.mdc-text-field--filled` directly. Same one-var-per-theme re-skin, but no `!important` war and the error state survives. There's already precedent for it in `_overwrite-material.scss` (`--mat-chip-container-shape-radius`, `--mat-menu-item-label-text-size`).

hajiboy95:
@johannesjo okay, I've splitted them. My intention was to make the overall design with it more coherent. , maybe we can do this with another PR to keep scope tight here. :)

johannesjo:

PR #8295, opened 2026-06-12
fix(android): keep add-task input above keyboard on Android 10

ynotzort:
## Problem

When running the app on my android 10 device, the keyboard would overlap the text box for entering the task.

## Solution

The global add-task bar positions itself with --keyboard-height, which was derived only from the Visual Viewport API. Android 10's WebView doesn't shrink the visual viewport for the IME, so the value stayed 0 and the input was hidden behind the keyboard. Now the natively measured keyboard height (already used for isKeyboardShown$) is forwarded as a fallback via a new keyboardHeightPx$ channel, with a layout-shrink correction so devices where adjustResize works aren't pushed up by a double offset. Later Android versions behavior is unchanged.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/android/app/src/main/java/com/superproductivity/superproductivity/CapacitorMainActivity.kt
+++ b/android/app/src/main/java/com/superproductivity/superproductivity/CapacitorMainActivity.kt
@@ -185,9 +185,11 @@ class CapacitorMainActivity : BridgeActivity() {
             if (keypadHeight > screenHeight * 0.15) {
                 // keyboard is opened
                 callJSInterfaceFunctionIfExists("next", "isKeyboardShown$", "true")
+                callJSInterfaceFunctionIfExists("next", "keyboardHeightPx$", keypadHeight.toString())
             } else {
                 // keyboard is closed
                 callJSInterfaceFunctionIfExists("next", "isKeyboardShown$", "false")
+                callJSInterfaceFunctionIfExists("next", "keyboardHeightPx$", "0")
             }
         }
 
--- a/android/app/src/main/java/com/superproductivity/superproductivity/FullscreenActivity.kt
+++ b/android/app/src/main/java/com/superproductivity/superproductivity/FullscreenActivity.kt
@@ -153,9 +153,11 @@ class FullscreenActivity : AppCompatActivity() {
             if (keypadHeight > screenHeight * 0.15) {
                 // keyboard is opened
                 callJSInterfaceFunctionIfExists("next", "isKeyboardShown$", "true")
+                callJSInterfaceFunctionIfExists("next", "keyboardHeightPx$", keypadHeight.toString())
             } else {
                 // keyboard is closed
                 callJSInterfaceFunctionIfExists("next", "isKeyboardShown$", "false")
+                callJSInterfaceFunctionIfExists("next", "keyboardHeightPx$", "0")
             }
         }
     }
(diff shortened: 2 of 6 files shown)


johannesjo:
Thanks for this — and congrats on the first PR here. 🎉 This is a well-reasoned fix: reusing the existing native `keypadHeight` instead of adding a new bridge API is the right call, and the `max(obscured, nativeCovered)` design with the `layoutShrink` correction holds up across the whole device matrix I traced (modern `adjustResize` shrink, modern vv-only/edge-to-edge, and the Android-10 "nothing shrinks" case). Using `max()` rather than summing the two signals is what keeps it safe. A few things to look at before merge:

## 1. (Important) `baseInnerHeight` goes stale across rotation → the fix can regress in landscape

`baseInnerHeight` is only refreshed in the native subscription's `cssPx <= 0` branch. But when the keyboard is closed the Kotlin side always emits the literal `0`, and the stream has `distinctUntilChanged()` — so a rotation while the keyboard is closed emits another `0`, which is filtered, and the baseline is never updated. Since both activities use `android:configChanges="orientation|...|screenSize|..."`, rotation doesn't recreate the Activity, so the stale baseline persists in the WebView.

Repro (on an Android-10-class device where neither viewport shrinks):

1. App in portrait, keyboard closed → `baseInnerHeight` = tall portrait height.
2. Rotate to landscape (don't open the keyboard) → baseline not refreshed.
3. Open the keyboard and focus the add-task input.

Now `layoutShrink = baseInnerHeight − innerHeight` is large (the portrait↔landscape delta, not the keyboard), so `nativeCovered = max(0, kbH − largeShrink) = 0`, `obscured` is `0` on these devices, and `--keyboard-height` lands at `0` → input behind the keyboard again. It self-heals after one close→open cycle, so it's narrow, but it's exactly the flow this PR fixes.

Minimal fix — refresh the baseline whenever the keyboard is detected closed, regardless of source, by adding one line in `scheduleCommit`'s "closed" branch:

```ts
const scheduleCommit = (): void => {
  if (measureKeyboardHeight() <= KEYBOARD_THRESHOLD_PX) {
    baseInnerHeight = window.innerHeight; // keep baseline fresh across rotation/layout changes
    if (resizeTimer !== null) {
      window.clearTimeout(resizeTimer);
      resizeTimer = null;
    }
    commit();
    return;
  }
  // ...
};
```

This also covers rotation, since `vv.resize` → `scheduleCommit` fires on orientation change, and it no longer depends on the native `0` slipping past `distinctUntilChanged`.

## 2. (Minor) `nativeKeyboardHeight` includes the navigation-bar inset

`keypadHeight = screenHeight − rect.bottom` is `navbar + IME`, not just the IME. On a device that shrinks the layout by the IME only, `nativeCovered` resolves to ~navbar height, which would be a phantom offset — it's masked today only because `KEYBOARD_THRESHOLD_PX` (100) exceeds a typical navbar (~24–48px). Worth a one-line comment near `measureKeyboardHeight` so the threshold isn't lowered later without realizing it props this up.

## 3. (Minor) A tiny unit test would lock this in

`measureKeyboardHeight` is pure arithmetic over `innerHeight` / `vv.height` / `nativeKeyboardHeight` / `baseInnerHeight`. Extracting it to a free function and testing the matrix above (including the rotation case) would prevent regressions and tick the unchecked "added tests" box.

---

Otherwise this looks good. Since it's empirical to your Android 10 device, it'd be worth a quick check on both an edge-to-edge and a non-edge-to-edge device (for point 2), plus the rotate-then-type flow (point 1).

ynotzort:
The comments should be addressed now, the rotation flow seems to work now.

johannesjo:

PR #8377, opened 2026-06-13
fix(sync): conflict-check all entities of multi-entity ops (#8334)

### LINKED ISSUE #8334: SuperSync server: multi-entity ops are invisible to cross-upload conflict detection — only the first entityId is persisted
**Area:** server (conflict detection) + schema · **Severity:** high (active — server-side conflict guard silently bypassed for common flows)

## Problem
Ops can carry `entityIds: string[]`, and conflict detection checks **all** of them for the *incoming* op. But the `operations` table only has a single `entity_id` column, and inserts persist `entityId ?? null` — the client sets `entityId = entityIds[0]`. So once a multi-entity op is stored, only its **first** entity participates in future conflict checks: a later stale single-entity op against entity 2..n finds no prior writer and is accepted instead of rejected as `CONFLICT_SUPERSEDED`/`CONCURRENT`.

This is **active, not latent**: production clients routinely send multi-entity ops — `deleteTasks`, `moveToArchive` (`src/app/root-store/meta/task-shared.actions.ts:87, :102`), `__updateMultipleTaskSimple` and round-time-spent (`src/app/features/tasks/store/task.actions.ts:33, :166`), plus task-repeat-cfg/boards/issue-provider batch actions.

## Evidence
- Wire/type: `packages/shared-schema/src/supersync-http-contract.ts:65-68`; `packages/super-sync-server/src/sync/sync.types.ts:170`.
- Incoming op checked against all its entityIds: `src/sync/conflict.ts:37-53` (single), `:290-320` (batch).
- Storage drops the rest: `prisma/schema.prisma:71` (only `entity_id`); inserts at `src/sync/services/operation-upload.service.ts:542` and `:784`.
- Lookups that miss prior multi-entity writers: `conflict.ts:71-81`, `:177-190`, `:339-351`.
- Within one batch the in-memory map *is* per-entity (`operation-upload.service.ts:489-498`) — the gap is only across uploads, which shows per-entity tracking is the intent.
- Client side: `src/app/op-log/capture/operation-log.effects.ts:206-218` (builds `entityIds`, sets `entityId = entityIds[0]`), uploaded unmodified (`operation-log-upload.service.ts:419`).

Severity tempering: the consequence is a missed rejection — the stale op is accepted and ordered by `server_seq`; client-side LWW may still converge, but the server-side guard is silently bypassed.

## Proposed fix
Persist per-entity index rows (e.g. side table `(user_id, entity_type, entity_id, server_seq)` fanned out per entityId) consulted by all three lookup paths — requires a backfill migration for existing multi-entity ops. Alternative: expand ops into one row per entity. At minimum, document the asymmetry in `docs/sync-and-op-log/` next to `detectConflict`.

johannesjo:
## Problem

Fixes #8334. Ops can carry `entityIds: string[]` (batch actions: `deleteTasks`, `moveToArchive`, `__updateMultipleTaskSimple`, round-time-spent, batch board/issue-provider). The server `operations` table only stored the scalar `entity_id` (= `entityIds[0]`), so once a multi-entity op was stored, **only its first entity took part in future cross-upload conflict lookups**. A later stale write to a non-first entity found no prior writer and was wrongly **accepted** instead of rejected as `CONFLICT_SUPERSEDED`/`CONCURRENT` — the server-side guard was silently bypassed.

Severity is tempered (client-side LWW still converges, and the server only builds an authoritative snapshot from non-encrypted ops), so the user-visible divergence path is confined to non-encrypted self-hosted servers — but the guard should be symmetric. See `docs/sync-and-op-log/vector-clocks.md` for the full write-up.

## Fix

- **`entity_ids text[]` column**, populated for **multi-entity ops only** (`getStoredEntityIds` — single-entity ops store `[]` and are matched via the scalar `entity_id`, keeping the GIN index small and off the single-entity insert write path). `entity_id` is retained for duplicate detection + backward compat. The gate is "is the touched set exactly `[entity_id]`?", not "length > 1", so a batch op whose ids dedup to one value that differs from `entity_id` (the server does not enforce `entity_id === entityIds[0]`) is still stored.
- **Migration 1** — `ADD COLUMN` (metadata-only, **no backfill**: forward-only; pre-migration rows fall back to the scalar `entity_id`. Entities 2..n of already-stored multi-entity ops were never persisted and stay unrecoverable — a bounded residual covered by client LWW).
- **Migration 2** — `CREATE INDEX CONCURRENTLY … GIN(entity_ids)` (single-statement, native-apply, per `prisma/migrations/README.md`).
- **`detectConflictForEntity`** (single, the per-upload hot path) — Prisma `findFirst` with `OR: [{ entityId }, { entityIds: { has: entityId } }]`, ordered by `server_seq`. The `OR` spans the `entity_id` btree + the `entity_ids` GIN (a `BitmapOr` + sort bounded by the entity's stored version depth, which op-log pruning keeps small — sub-ms in practice). A code comment documents a two-ordered-lookup split as the escalation if a real-PG `EXPLAIN` ever shows it matters. *(An earlier revision shipped that split, then reverted it: it added a round-trip on the common path to optimize a pruning-bounded concern.)*
- **`detectConflictForEntities` / `prefetchLatestEntityOpsForBatch`** (batch) — raw SQL matching an entity as the scalar `entity_id` **or** a member of `entity_ids` via `unnest(CASE WHEN cardinality(entity_ids) > 0 THEN entity_ids ELSE ARRAY[entity_id] END)`, with a `&&`/`= ANY` prefilter so the GIN and the existing `(user_id, entity_type, entity_id, server_seq)` btree stay usable.
- **`validateOp`** now bounds `entityIds` (count ≤ `SUPER_SYNC_MAX_ENTITY_IDS_PER_OP`, each element a non-empty string ≤ 255 chars), mirroring the scalar `entityId`.

## Validation

- Raw SQL, GIN usage, the empty-array/dedup-off-scalar edges, and index-backed prefilter validated against real Postgres (PGlite, out-of-band).
- Unit tests for the single path, `getStoredEntityIds` (incl. the dedup-off-scalar edge), `validateOp` bounds, and migration shapes — **all runnable specs green**.
- Reviewed across three multi-agent passes (correctness, security, architecture, alternatives, performance, simplicity); final verdict: ship.

## Not verifiable locally — please confirm in CI / on a real DB before merge

- The full `conflict-detection.spec.ts` needs a generated Prisma client (`prisma generate` runs in CI; the worktree has an ungenerated stub). Its mock was hand-updated to match the SQL.
- The batch raw SQL has no committed real-PG test (the integration vitest config uses a real DB in CI).
- The single-path `OR` plan and the batch unnest plans should get a real-Postgres `EXPLAIN ANALYZE` on a deep-history entity if hot-path latency is ever a concern (bounded by op-log pruning).

## Risk

Sync-critical change to conflict detection. The storage is append-only (the denormalized `entity_ids` can't drift), all lookups stay `user_id`-scoped, and no `entity_ids` reader other than conflict detection exists (op-replay, snapshot generation, and duplicate detection are unaffected).


johannesjo:
Reviewed for correctness and verified the risky SQL against a real Postgres (in-process via PGlite), not just by reading — **looks solid, no blocking findings:**

- The prefilter `(entity_ids && idArr OR entity_id = ANY idArr)` is a correct superset of the `unnest … eid = ANY` filter for both array-stored and scalar-stored (pre-migration) rows — no row that should match is excluded.
- `DISTINCT ON (eid) … ORDER BY server_seq DESC` returns the latest op per entity, including via a non-first entity.
- A full-state op (`entity_id = NULL`, `entity_ids = {}`) hits the `ARRAY[entity_id]` branch with a NULL element → safely ignored, no error, no false match.
- `getStoredEntityIds` gate is right including the dedup-off-scalar case (`entityIds:['A','A']`, `entityId:'Z'` → stores `['A']`).
- `MAX_ENTITY_IDS_PER_OP = 1000` matches the Zod wire bound, so legit large batch deletes aren't rejected.

**One thing to fix in the description / process:** the batch raw SQL isn't actually validated by CI. The `tests/integration/*.integration.spec.ts` files are `DATABASE_URL`-gated and excluded from the default config, and the only job that runs `super-sync-server` tests (`e2e-scheduled.yml`) runs `npm test` with no Postgres service — and it isn't a PR check anyway.

To close that gap I wrote `tests/entity-ids-conflict.pglite.spec.ts` (10 tests) that runs the exact `unnest(CASE …)` SQL from `conflict.ts` against in-process Postgres via **PGlite** (no Docker, no `DATABASE_URL`), so it runs in the normal `npm test`. Confirmed it's a genuine guard — the old scalar-only SQL returns `[]` for the non-first entity. It adds `@electric-sql/pglite` as a dev-only dependency. Happy to push it onto this branch if you want it here.

PR #8381, opened 2026-06-13
fix(keyboard): resolve macOS global shortcut layout mismatch (#8378)

### LINKED ISSUE #8378: 🚨 Bug: Global shortcut listener ignores system layout at input (Y/Z swapped on QWERTZ) on Mac
### Version Used

MacOS 10.18.0

### Current Behavior

There is a disconnect between the visual UI and the global input listener. The menu *displays* the correct key based on the shortcut settings, but the global shortcut listener falls back to an English (QWERTY) layout.

Crucially, **this bug is isolated to global (system-wide) shortcuts**. Local, in-app shortcuts respect the correct layout.

For example, on a QWERTZ keyboard, pressing the physical `Y` key (bottom left) does not trigger a global `Y` shortcut. Instead, you have to press the physical `Z` key (top middle) to trigger the global shortcut that is visually labeled as `Y`.

**Additional Context**

* Because local shortcuts work correctly, this strongly indicates an issue specific to how system-wide keystrokes are registered (e.g., Electron's `globalShortcut` module) versus standard DOM keydown events. The global hook appears to be reading the physical hardware keycode rather than the OS-mapped character.
* OS context: macOS using a German QWERTZ layout.

### Expected Behavior

Global shortcuts should respect the localized keyboard mapping provided by the operating system, consistent with how local shortcuts behave. Pressing the physical `Y` key on a QWERTZ layout should register as `Y` for global shortcut recording and execution.

### Steps to Reproduce

1. Ensure your operating system's keyboard layout is set to a non-QWERTY layout (e.g., German QWERTZ).
2. Open the application settings and navigate to the **Global Shortcuts** section.
3. Look at a global shortcut in the menu that involves the `Y` or `Z` key (the menu correctly displays `Y`), or attempt to record a new one.
4. Press the physical `Y` key on your QWERTZ keyboard.

### Console Output

```shell

```

### Documentation

hajiboy95:
Tackles #8378

## Problem

On macOS, Electron's `globalShortcut` module registers system-wide shortcuts based on physical key positions mapped according to the US QWERTY layout. For non-QWERTY layouts (such as QWERTZ or AZERTY), this results in a mismatch where pressing the physical key labeled with the character does not trigger the shortcut (e.g., on a QWERTZ layout, pressing physical `Y` does not trigger the `Y` shortcut; instead, you have to press physical `Z`). Fixes #8378.

## Solution

We translate the global shortcut keys to their US QWERTY layout equivalents in the renderer process on macOS before sending them to Electron's main process via `registerGlobalShortcuts`. This ensures the global listener registers the correct physical keys.

Specifically:
- Added a `layoutReady: Promise<KeyboardLayout>` to `KeyboardLayoutService` to await the browser's native Keyboard Map API before registering initial shortcuts on startup.
- Implemented layout mapping helpers (`mapShortcutToQwerty` and `mapKeyboardConfigToQwerty`) in `GlobalConfigEffects` to translate localized characters (e.g. `Y` on QWERTZ) to their QWERTY equivalent (e.g. `Z`) using the active `KeyboardLayout`.
- Applied this translation exclusively on macOS (`IS_MAC`) when registering shortcuts.
- Updated `KeyboardLayoutService` unit tests to cover the `layoutReady` promise behavior.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for tackling this — I traced the mapping end-to-end and the **direction is correct** (recorded char → find the layout code that produces it → derive the QWERTY-position char from that code → register that with Electron), and `GLOBAL_KEY_CFG_KEYS` matches `electron/ipc-handlers/global-shortcuts.ts` exactly. `layoutReady` also resolves on every path (no-API / error / `setLayout`), so the initial registration can't hang. A few things to address before this can go in:

## Blocking

- **`global-config.effects.ts:78-97` — debug `console.log`s left in.** The five `console.log('[Shortcut Debug] ...')` statements (each silenced with `eslint-disable-next-line no-console`) in `updateGlobalShortcut$` run on *every* keyboard-config update. These need to be removed along with their eslint-disable comments.

## Should fix

- **No test coverage for the actual fix.** `mapShortcutToQwerty` / `mapKeyboardConfigToQwerty` (lines 301 & 363) are module-private, so the core translation logic is untested — the new specs only cover the `layoutReady` promise. Please `export` both and add unit tests for: QWERTZ `Y`→`Z`, a digit, a `qwertyCodeMap` punctuation case, the `+` key, and the "char not found → return unchanged" fallback (e.g. `F2`). The bug below is exactly what such a test would catch.

- **`global-config.effects.ts:310-312` — the `++` branch drops the modifier separator.** The normal branch builds `modifiersPart` *with* its trailing `+` (line 319: `...join('+') + '+'`), but the `endsWith('++')` branch does `shortcut.slice(0, -2)`, stripping both `+` chars. So when the `+` key is actually remapped — e.g. on a German QWERTZ layout where `getLayoutMap()` returns `BracketRight → '+'` — a global shortcut of `Ctrl++` produces `'Ctrl' + ']'` = `'Ctrl]'` (should be `'Ctrl+]'`), which is a malformed accelerator that fails registration and fires the error toast. Fix is `slice(0, -1)` so the separator survives. (Narrow edge case — only the 5 global keys are mapped and `Ctrl++` is a default only for the *local* `zoomIn` — but the function explicitly tries to handle `+`, so it should handle it correctly.)

- **`global-config.effects.ts:379` — `as any`.** `mappedCfg[key] = mapShortcutToQwerty(...) as any` only needs the cast because `KeyboardConfig` is `Readonly`. Typing the working copy as a mutable record avoids it and keeps strict-TS happy:
  ```ts
  const mappedCfg: Record<string, string | null | undefined> = { ...keyboardCfg };
  // ...
  return mappedCfg as KeyboardConfig;
  ```

## Minor

- **`keyboard-layout.service.ts:68-69` — empty `catch (e) { // Ignore error }`** silently swallows a `getLayoutMap()` failure. A `Log.err(e)` would make a layout-detection failure diagnosable while keeping the `_resolveLayoutReady` fallback.
- **`keyboard-layout.service.ts` — `layoutReady` resolves with the internal map reference**, which a later `saveUserLayout()` clears and repopulates in place. Harmless today (the consumer reads synchronously right after the `await`), but resolving with a copy (`new Map(this._layout)`) would be more robust.
- **macOS scope** — the fix is correctly `IS_MAC`-only per the issue; a one-line comment near the `IS_MAC` guards noting "macOS globalShortcut maps by physical US-QWERTY position; see #8378" would help the next reader (and flag that Windows/Linux are intentionally out of scope).

Nice touches: gating initial registration on `layoutReady`, handling the no-keyboard-API path, and the graceful pass-through when a char isn't in the layout.

hajiboy95:
@johannesjo thanks for your feedback. Should be adressed now. :)

johannesjo:
**Reviewed `a6c6e96..077b894` (the 5 follow-up commits).** This range is clean — solid cleanup plus one real bug fix. 👍

**The `++` fix (`bf24888876`) is the substantive one and it's correct.** For a "Ctrl + plus-key" shortcut stored as `'Ctrl++'`, the old `slice(0, -2)` produced `'Ctrl'` (separator dropped → `'Ctrl]'`); the new `slice(0, -1)` produces `'Ctrl+'` → `'Ctrl+]'`. The new `mapShortcutToQwerty('Ctrl++') → 'Ctrl+]'` test locks it in. ✅

**Other changes all check out:**
- Resolving `layoutReady` with `new Map(this._layout)` is the right call — the promise resolves once, so without the snapshot a consumer that `await`s it would hold a live reference that a later `saveUserLayout()`/`setLayout()` would `clear()` out from under them. The new spec correctly fails against the old shared-reference code.
- `Log.err(e)` over the silent `// Ignore error` is a good improvement (and `e` is a DOM exception, not user content, so no logging-policy concern).
- Dropping `as any` in favor of `Record<string, string | null | undefined>` + a single `as KeyboardConfig` return cast is type-safe — every `KeyboardConfig` value is `string | null | undefined`. One heads-up: `npm run checkFile` only runs prettier+lint, not `tsc`, so this typing isn't covered by the listed checks — worth a local `npm test`/build to be sure CI is green.

**Optional nits (non-blocking):**
- `mapShortcutToQwerty` takes the first layout code matching a character. Fine in practice for these keys, just noting iteration order decides if a layout ever maps two codes to the same char.
- The `// macOS globalShortcut maps by physical US-QWERTY position; see #8378` comment is repeated 3×; could be hoisted to the helper's JSDoc, but the inline placement does aid local readability.
- Test fixture uses `as any`; `as Partial<KeyboardConfig>` would be marginally cleaner.

I traced all six `mapShortcutToQwerty` cases and the config-mapping test against the implementation (all pass) and confirmed the new typing compiles under `--strict`. Nice work.

hajiboy95:
@johannesjo thanks for the feedback. :) adressed

johannesjo:
## Review — looks good, approve with minor suggestions

Traced the fix end-to-end against the QWERTZ case in #8378 and the direction/mechanism are correct. Applied the diff locally: both spec files pass (18 + 14 tests) and `eslint` is clean on the changed sources.

### Why the fix is correct
- Shortcuts are stored as **characters**: `keyboard-input.component.ts` records via `prepareKeyCode(ev.code)`, which maps physical `KeyZ` → `"Y"` on QWERTZ (matches the issue: "menu correctly displays Y").
- On macOS `globalShortcut.register("Y", …)` resolves by US-QWERTY physical position (`kVK_ANSI_Y` = top-row key, labeled `Z` on QWERTZ) — the bug.
- `mapShortcutToQwerty("Y", layout)` finds the code whose value is `"y"` (`KeyZ`) → `"Z"`, so Electron registers the physical key the user actually pressed. Round-trips consistently with `prepareKeyCode` (incl. AZERTY digits/symbols) because both read the same `getLayoutMap()` data. ✅

### Also confirmed
- Only the 5 system-wide keys are translated; local shortcuts (already layout-mapped by `checkKeyCombo`) are correctly untouched — the `toggleBacklog` test asserts this.
- No state mutation — `mapKeyboardConfigToQwerty` returns a fresh object; only the value sent to `window.ea` is mapped, not the persisted config.
- Graceful degradation: empty/failed layout → shortcuts pass through unchanged (= current behavior, no regression). The new `try/catch` around `getLayoutMap()` plus resolving `layoutReady` afterward is a nice defensive touch, and `saveUserLayout()` is scheduled unconditionally in the `AppComponent` constructor, so the `await layoutReady` can never hang.

### Suggestions (non-blocking)
1. **`GLOBAL_KEY_CFG_KEYS` is now duplicated** — verbatim in `global-config.effects.ts` and `electron/ipc-handlers/global-shortcuts.ts`. If a 6th global shortcut is later added to the Electron list but not the renderer list, it silently won't be QWERTY-translated on macOS. Worth extracting to one shared const (e.g. in `keyboard-config.model.ts`).
2. **Effect-level test gap.** The two pure helpers are well covered, but nothing exercises the effect itself — that on `IS_MAC` it awaits `layoutReady` and calls `registerGlobalShortcuts` with the *mapped* cfg, which is the integration this PR actually changes. A small effect test would lock in the wiring.
3. **Startup timing on macOS.** Registration is now gated behind `layoutReady`, which depends on the `requestIdleCallback(saveUserLayout)` deferral (previously registration was synchronous on `loadAllData`). The net effect is a brief window at startup where macOS global shortcuts are inactive — almost certainly fine, just noting the behavior change.
4. Nits: the "first matching code wins" caveat is low-risk since `getLayoutMap()` returns only writing-system keys (no numpad); `qwertyCodeMap` omits `IntlBackslash`/`IntlRo`/`IntlYen` (they fall through unchanged — acceptable); the exported pure helpers could live in a dedicated util alongside the shared key list.

### One thing to flag
The fix relies on the premise that Electron's macOS `globalShortcut` resolves accelerators by US-QWERTY physical position. The reporter reproduced this reliably and the logic matches, but it **can't be verified in CI** (no macOS + non-QWERTY harness) and Electron's behavior here has varied across versions. Would be great to get a before/after confirmation from the #8378 reporter or on a QWERTZ Mac before merge.

hajiboy95:
@johannesjo feedback is adressed. Regarding Testing: I am Using a Mac and can confirm that the implementation is working

johannesjo:
## Review — refactor + tests + eager layout detection (`6f8e477..dc85761`)

Reviewed this range with a multi-perspective pass (correctness, security, architecture, performance, simplicity). The core fix works and the extraction is clean — **no blocking issues**. One change is worth making before merge; the rest are optional hardening/polish.

### 🔧 Worth changing before merge

**Replace the `export let` + `setXForTesting` test seam with the project's existing `InjectionToken` pattern**
`src/app/app.constants.ts:3-6`, `src/app/util/is-mac.ts:1-5`

Converting `IS_ELECTRON`/`IS_MAC` to mutable module globals (with ungated production setters) just to make the effect tests pass has a few real downsides:

- **Derived constants desync.** `IS_WEB_BROWSER`, `IS_GNOME_DESKTOP`, `IS_APPLE_SILICON` are computed from `IS_ELECTRON` at module-load time and do **not** track `setIsElectronForTesting(...)`. After flipping the flag, `IS_ELECTRON` and `IS_WEB_BROWSER` describe an impossible production state — a footgun for any future test.
- **Cross-test pollution.** `afterEach` resets to a hardcoded `false` rather than the captured original, so on a macOS dev machine (where `IS_MAC` is genuinely `true`) these tests force `IS_MAC=false` for the remainder of the suite.
- **Doesn't even reach all consumers.** `notifyElectronAboutCfgChange`/`…Initially` read `IS_ELECTRON` at class-field-init time, so the setter can never toggle them.

The codebase already solves exactly this, right next door — `src/app/util/is-android-web-view.ts` exposes an `InjectionToken`:

```ts
export const IS_ANDROID_WEB_VIEW_TOKEN = new InjectionToken<boolean>(
  'IS_ANDROID_WEB_VIEW',
  { providedIn: 'root', factory: () => IS_ANDROID_WEB_VIEW },
);
```
…mocked in specs via `{ provide: IS_ANDROID_WEB_VIEW_TOKEN, useValue: true }`.

Suggestion: add `IS_ELECTRON_TOKEN` / `IS_MAC_TOKEN` the same way, keep the `const`s, and inject them only in `GlobalConfigEffects` (~3 fields + provider lines in the one spec). The other consumers keep the plain `const` import untouched — and this is actually *less* code than the two setters + the `afterEach` reset, while removing the desync and the pollution. (FWIW the bundle-size angle is **not** a concern here — these flags are runtime-derived from `navigator.*`, so no minifier ever folded the `if (IS_ELECTRON)` branches; verified with esbuild.)

### 💡 Recommended hardening

**`registerGlobalShortcutInitially$` awaits `layoutReady` with no timeout**
`src/app/features/config/store/global-config.effects.ts:103`

The await resolves on every `saveUserLayout()` path *except* if `navigator.keyboard.getLayoutMap()` never settles — in which case the initial macOS registration never fires and **no global shortcuts register at all** for the session. This is a new dependency the pre-PR code didn't have. A `Promise.race([layoutReady, <short timeout>])` falling back to the untranslated config would degrade gracefully.

### ✨ Optional polish

- **`src/app/app.component.ts:318`** — the eager `saveUserLayout()` runs on all Electron platforms, but only macOS consumes `layoutReady`. Scoping to `IS_ELECTRON && IS_MAC` avoids redundant startup work on Windows/Linux (cost is negligible either way).
- **`src/app/features/config/keyboard-shortcut.util.ts`** — hoist the `qwertyCodeMap` object literal to a module-level `const` so it isn't reallocated per call. (Pre-existing relocated code.)
- **`global-config.effects.spec.ts` ("should wait for layout…")** — the test passes even if the `await layoutReady` were removed (the `setLayout` mutation lands synchronously before the `concatMap(async)` microtask flushes). Deferring `setLayout` to a macrotask would make it actually exercise the wait.

### 👍 Looks good
Extracting `keyboard-shortcut.util.ts` and hoisting `GLOBAL_KEY_CFG_KEYS` into `keyboard-config.model.ts` removes genuine duplication (the array was previously copy-pasted in both the Electron main process and the effects file), matches the existing `TASK_REPEAT_WEEKDAY_MAP` convention, and tightens the renderer↔main allowlist. The mapping logic is moved verbatim, so no behavior risk there.

hajiboy95:
@johannesjo 

  Summary of modifications:
   - InjectionToken Migration: IS_ELECTRON and IS_MAC are now constants paired
     with InjectionTokens, following the project's best practices for
     testability and avoiding global state pollution.
   - Resilient Layout Detection: Added a 1s timeout to the keyboard layout
     promise to ensure global shortcuts register even if the native API hangs.
   - Startup Optimization: Eager layout detection is now strictly scoped to
     macOS Electron, reducing redundant startup work on other platforms.
   - Performance Polish: Hoisted QWERTY_CODE_MAP to the module level to avoid
     per-call allocations.
   - Improved Test Coverage: Refactored unit tests to correctly exercise
     asynchronous layout detection and verify platform-specific logic using
     isolated providers.

johannesjo:
Nice work on this — the root-cause framing is accurate and the fix is in the right place. Reviewed the full PR (10 files); ran both affected specs locally (`keyboard-layout.service.spec` 14/14, `global-config.effects.spec` 23/23 ✅) and `eslint`/`prettier` are clean.

**Verdict: LGTM, no blockers.** A few non-blocking suggestions below.

### Why this is well-designed
- Translating in the **renderer** (where `navigator.keyboard.getLayoutMap()` lives) and keeping the Electron handler layout-agnostic is the correct split.
- Hoisting `GLOBAL_KEY_CFG_KEYS` into the shared model removes the duplicated inline list — good single source of truth for renderer + main.
- The `layoutReady` promise resolves on every path of `saveUserLayout()`, and the `Promise.race(…, 1000ms)` guards the one path that can't self-resolve (`getLayoutMap()` hanging). Solid.
- Edge cases are handled and tested: literal `+`/`++`, function keys (left unchanged — correct, they're layout-invariant), empty/unmapped layout falling back to the original string.

### Non-blocking suggestions

1. **`app.constants.ts` — move the import to the top.** `import { InjectionToken } from '@angular/core'` is currently mid-file after the exports. `is-mac.ts` puts it at the top; worth matching for consistency.

2. **Make the 1s timeout diagnosable.** If layout detection ever exceeds 1000ms, the initial global shortcuts register *untranslated* and stay wrong until the next config change, with no signal. It's very unlikely (eager call + `getLayoutMap()` is typically sub-10ms), but a `Log.log(…)` when the timeout branch wins — plus extracting a named `LAYOUT_DETECTION_TIMEOUT_MS` — would help if it ever bites. The race timer also isn't cleared once `layoutReady` wins (harmless orphan, but tidy to clear).

3. **Stale comment in `app.component.ts`.** The "EXCEPT on Electron … on macOS" comment now sits above an `IS_ELECTRON && IS_MAC` guard — tighten to "EXCEPT on macOS Electron".

4. **(Optional) Dual source of truth.** `IS_ELECTRON`/`IS_MAC` const now coexists with the new injection tokens (the const still has ~74 importers). Totally fine, but a one-line note that new DI-tested effects/services should prefer the token would prevent drift.

Thanks for the thorough tests and the clear PR description — the QWERTZ Y↔Z walkthrough made review easy. 👍

hajiboy95:
@johannesjo i also adressed with my last 2 commits the electron test error. I hope it's okay like this. if not, i can revert them and just keep the refactor commit which you were asking for. :)

johannesjo:
**Thanks for this — the approach is sound and the testing is thorough.** Doing the QWERTY translation renderer-side via the Keyboard Map API (rather than a native macOS module in main) is the right call, it reuses the existing `KeyboardLayoutService` instead of adding new infrastructure, and it degrades gracefully (empty/failed/timed-out layout → original config; US layouts are no-ops). Nice follow-ups too — the `clearTimeout`, the named `LAYOUT_DETECTION_TIMEOUT_MS`, and the import hoist.

Two things before merge:

### 🔴 Must fix — CI red: the packaged Electron app no longer compiles

The **Linux Packaged Electron App** job fails:

```
src/app/features/config/global-config.model.ts(6,32): error TS2307:
Cannot find module '@sp/keyboard-config'
```

This is introduced by the last commit (`962ef9bb`), which swapped a relative import for the new alias:

```diff
-import { KeyboardConfig } from './keyboard-config.model';
+import { KeyboardConfig } from '@sp/keyboard-config';
```

The `@sp/keyboard-config` alias was added to `tsconfig.base.json` and `src/tsconfig.spec.json` (so the web build and **Tests** stay green) but **not** to `electron/tsconfig.electron.json`, which is standalone and doesn't extend base. The Electron build pulls this file in through a pre-existing, unchanged chain — `window-ea.d.ts` → `electronAPI.d.ts` → `global-config.model.ts` — and can no longer resolve the alias. (For reference, this same check is ✅ green on `master`, where the import is still relative.)

One-line fix — add to the `paths` block in `electron/tsconfig.electron.json`:

```json
"@sp/keyboard-config": ["electron/shared-with-frontend/keyboard-config.model.ts"]
```

Heads-up while you're there: that file has **two `paths` keys in one `compilerOptions` object** — JSON keeps only the last, so the first block is dead. The new entry must go in the second (effective) one; worth merging the two.

### 🟡 Please verify on-device before merge

All the tests (here and in CI) exercise the translation against a **mocked** layout map — they can't validate the two real-world assumptions the fix rests on: that Electron's macOS `globalShortcut` registers by US physical position, and that `getLayoutMap()` returns the expected code→char map inside Electron's Chromium. Could you or the #8378 reporter confirm on an actual QWERTZ/AZERTY Mac that the shortcut now fires on the labeled key? Regression risk is low (gated to macOS Electron; non-QWERTY layouts are already broken, so this can't make them worse), but the positive case is currently unverified.

### Optional / nit

- The `IS_ELECTRON`/`IS_MAC` → `InjectionToken` migration is broader than the fix strictly needs — the real logic lives in the pure `mapShortcutToQwerty`/`mapKeyboardConfigToQwerty` functions, already tested without DI. It's a reasonable testability improvement given the repo's DI style, so feel free to keep it; just flagging it as the largest blast-radius change relative to its bug-fixing value.

Once CI is green and there's an on-device confirmation, this looks good to merge. 👍

hajiboy95:
@johannesjo CI is already green :D. Probably your AI started it's tests before I pushed the last commit

johannesjo:
Ha, probably :D

Thank you very much for this!

PR #8387, opened 2026-06-13
Feat/quick add hud 8386

### LINKED ISSUE #8386: 💡 Dedicated Quick-Add window via separate Angular route & IPC delegation
### Problem Statement
Currently, triggering the global shortcut for adding a new task (`globalAddTask`) opens and focuses the entire main application window to display the add task overlay. 

This is disruptive when working in another application (e.g. IDE, terminal, browser) and you just want to quickly capture a task. It forces a complete context switch by bringing the whole Super Productivity interface to the foreground, rather than keeping you focused on your current workspace.

### Possible Solution
Create a dedicated, frameless, and transparent Electron window specifically for adding tasks, loading the Angular app on a separate route (e.g. `/quick-add`).

- **Key Highlights:**
  - **Angular Route:** A new route `/quick-add` that renders only the `<add-task-bar>` component. The main app shell (side nav, headers, background images) will be hidden on this route.
  - **Electron Window:** A pre-created, hidden-by-default `BrowserWindow` (similar to the task-widget but styled as a centered HUD). The global shortcut triggers this window to show instantly.
  - **IPC Delegation:** To prevent out-of-sync state between the two running Angular instances (NgRx stores), the Quick-Add window will intercept the submit action. Instead of saving the task directly to the database in the quick-add process, it will emit the task details via IPC to the main window. The main window's Angular app will receive the IPC event and safely add the task to the primary store (updating IndexedDB and the UI).
  - Once submitted or dismissed, the Quick-Add window hides again.

This approach gives us the full power of the existing `<add-task-bar>` (autocomplete suggestions, tag/project tags, issue search) without the synchronization problems of multiple database writers.

1. **Lightweight Vanilla HTML/JS Window (like task-widget):** Easier on memory, but loses all rich autocompletion, issue search (Jira/GitHub), and short syntax parsing unless we duplicate or rewrite complex logic in vanilla JS.
2. **Direct Database writes in the second window:** Leads to synchronization issues because the main window's NgRx store is in-memory and wouldn't be aware of database changes until a full reload.

### Additional context
The shortcut registration is in [global-shortcuts.ts](file:///Users/Finbridge/Desktop/Second_Brain/3.%20Ressourcen/Github/super-productivity/electron/ipc-handlers/global-shortcuts.ts#L79-L85) and the layout is controlled in [app.component.html](file:///Users/Finbridge/Desktop/Second_Brain/3.%20Ressourcen/Github/super-productivity/src/app/app.component.html). Using Angular routing inside a second Electron window with IPC forwarding seems to be the cleanest path forward.

hajiboy95:
## Problem

There was no quick way to capture a task from anywhere on the system without first switching to the Super Productivity window. The existing `globalAddTask` shortcut opens the in-app add bar, but it forces a context switch to SP — breaking flow when working in another app, browser, or editor.

Closes #8386.

## Solution

Implements a **Quick Add HUD** — a system-wide floating overlay (similar to macOS Spotlight) that appears on top of whatever you are currently working in, without switching focus away from your active app.

**Key changes:**

- **New route & component** (`/quick-add`): A dedicated Angular route renders the existing `AddTaskBarComponent` on a fully transparent, frameless window overlay. The background uses a subtle semi-transparent tint (`rgba(0,0,0,0.15)`) to prevent macOS compositor rendering artefacts on transparent windows.

- **Electron Quick Add window** (`electron/quick-add-window.ts`): A pre-created `BrowserWindow` (transparent, `alwaysOnTop`, `skipTaskbar`, no frame) is shown on demand. The window covers the full display the cursor is on, re-centring on each activation. Blur (click-outside) dismisses the HUD automatically.

- **Focus management (macOS)**: When triggered from an external app, the SP main window is hidden at the window level (`orderOut`) *before* `show()` activates the Electron app — this prevents macOS from bringing the main window to the foreground. On dismiss, the main window is restored and `app.hide()` returns focus to the previously active external app. When SP itself was already focused, no hiding/restoring is done; `alwaysOnTop` is sufficient.

- **New global shortcut** (`globalTaskQuickAdd`): A new configurable keyboard shortcut registered via `GlobalShortcut`. Configurable in Settings → Keyboard Shortcuts like all other global shortcuts.

- **Consecutive task entry**: After pressing `Enter` to submit, the HUD stays open with the input cleared for rapid back-to-back task capture. Press `Escape` or click outside to fully dismiss.

- **Opaque CDK overlays**: Within the `/quick-add` route, Material dropdowns, autocomplete panels, and menus are forced to `var(--surface-2)` background so they are legible against the transparent window.

- **IPC delegation**: Submitted tasks are forwarded via `IPC.QUICK_ADD_SUBMIT_FORWARD` to the main window's web contents so they are processed by the normal task-creation pipeline.

- **Wiki documentation**: Added a "Quick Add HUD" section to `docs/wiki/3.03-Keyboard-Shortcuts.md`.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [ ] Other (please describe)

## Checklist

## Additional Context

When the app ist started with console open, the overlay will display a white background with the console on the ground. as soon as this console is closed, the overlay works as intended. Also with all it's own Popups (like e.g. calendar. 

**Example**

grafik


johannesjo:
## Review summary — Quick Add HUD

Thanks for this @hajiboy95 — the feature is genuinely useful and a lot of the work is solid: the keyboard flow (auto-focus, Enter/Escape, full `Ctrl+1`–`8` metadata, fully mouse-free), the **IPC-delegation design** (HUD forwards the submit; the main window stays the single NgRx/op-log writer), the added unit tests, and reverting `de.json` are all right.

Below is consolidated feedback from a multi-perspective review (correctness, security, architecture, performance, simplicity, plus a UX/product pass). I verified the blocking items against the source. The headline: the *idea* and the *delegation half* are right, but the **window substrate** (a second full Angular app instance) drives most of the problems, and there are a couple of hard correctness blockers.

### 🔴 Blocking

1. **App won't quit — zombie process.** `electron/quick-add-window.ts` pre-creates the window at startup and **never closes/destroys it** (only `init`/`show`/`hide`). Both quit paths (`_quitApp()` and `before-quit`→`closeWinAndQuit`) just `mainWin.close()` and rely on `window-all-closed` to fire `app.quit()` — the code even destroys the task-widget for exactly that reason. The hidden quick-add window is still an open window, so `window-all-closed` never fires and the app keeps running headless. This hits every desktop user (the window is pre-created even if the shortcut is never bound). → Add a `destroyQuickAddWindow()` and call it on the quit path, mirroring `destroyTaskWidget()`.

2. **Global shortcut crashes after the main window is closed (macOS).** `getWin()` (`main-window.ts:67`) *throws* when the main window is null (it's nulled on `closed`). On macOS the app outlives a closed main window, and `showQuickAddWindow`/`hideQuickAddWindow` call `getWin()` unguarded inside the `darwin` block — so the shortcut throws and the HUD never opens. (The forward handler wraps it in try/catch; show/hide don't.) → Guard with try/catch or a non-throwing accessor.

3. **A second full Angular + NgRx app boots in the HUD window.** `isQuickAddRoute()` only swaps the *template* — `AppComponent` still runs `_startupService.init()` and `main.ts` has no quick-add branch, so the hidden window runs full store hydration, plugins, LocalRestApi, **a second autonomous sync engine**, and the op-dispatching day-change effect (`task-due.effects`), with `backgroundThrottling:false` keeping it at full rate while hidden. The op-log lock (`op-log/sync/lock.service.ts`) *explicitly assumes Electron is single-instance* and uses a per-process mutex that doesn't span the two renderer processes — so two processes writing the same `SUP_OPS` op-log is a real sync-correctness/data-safety risk (the "one user intent = one op / replay determinism" invariant). The HUD's own store/sync/effects shouldn't run at all — the delegation design already makes them unnecessary.

4. **Security: the new window carries the full `window.ea` bridge but lacks the main window's navigation guards.** It uses the full `preload.js` (`exec`, `loadBackupData`, `fileSync*`, `openPath`, plugin `executeScript`, …) **and** a navigable `loadURL`, but installs none of the `will-navigate`/`will-redirect`/`setWindowOpenHandler`/`did-create-window` guards the main window was deliberately given after prior advisories (GHSA-hr87). The task-widget avoids this by using a *minimal* preload + static `loadFile`. Not trivially exploitable today (no in-bar anchors), but it reopens the class of issue those guards exist to prevent. → Mirror the main-window guards and add `sandbox: true` / `disableDialogs: true`.

### 🟡 Should fix

- **Unvalidated IPC payload.** `QUICK_ADD_SUBMIT` is forwarded verbatim into `_taskService.add(...)` with `taskData` spread last (can set `id`/`parentId`), no sender check. Add a sender check + `typia` validation at the boundary.
- **`any` types** in `QuickAddTaskPayload` (`taskData`/`remindOption`/`repeatCfg`) violate the strict-no-`any` rule — and are the enabler for the validation gap above. Use `Partial<Task>`, `TaskReminderOptionId`, `Partial<TaskRepeatCfg>`.
- **Triplicated, already-diverged task-creation logic** across the add-task-bar normal path, the add-task-bar IPC path, and `shortcut.service.ts`. `isTimedRepeatTask` already differs between them (`state.time` vs `payload.taskData.dueWithTime`), and the IPC scheduling path dropped the `timeout(1000)` the in-app path has. Extract one builder + one applier.
- **`transparent: true` is unconditional**, contradicting the main window's own `transparent: false` "Wayland compatibility" decision (`main-window.ts:213`); the task-widget uses `transparent: !IS_MAC`. This is likely the white/black-box artifact noted in the PR. Gate it per platform.
- **Windows `file://` path** uses `join` (backslashes) where main-window uses `normalize(...)` — may fail to load on packaged Windows.
- **No focus restoration on Windows/Linux** (the whole `darwin`-only focus block) — the headline "focus returns to where you were" promise is macOS-only.
- **macOS focus state machine** (`wasMainWinFocused`/`mainWinWasVisible`) corrupts on rapid re-trigger, and `isHidingProgrammatically` is reset synchronously around an async `hide()`/`blur`. Re-entry while already visible should be a no-op.
- **Lazy-create instead of pre-creating at startup** — the shortcut is `null` by default, so most users pay the always-resident cost for a feature they never enable. `showQuickAddWindow` already lazy-creates.

### 🟢 Suggestions
- `page.scss` blanket `.mat-mdc-*` opacity overrides under `body.isQuickAddRoute` are the one-off-Material-override anti-pattern; prefer CDK `panelClass`/a scoped overlay container (most of this disappears with a dedicated bootstrap).
- `app.component.ts` route detection via `url.includes('/quick-add')` + `document.body.classList` toggling + `event: any` is non-idiomatic — use route `data: {page}` (existing convention) + the `BodyClass` mechanism.
- `app.component.html` is ~147 lines of pure re-indentation churn (`git diff -w` shows the real change is ~4 lines). A separate HUD bootstrap avoids touching it at all.
- `if (IS_ELECTRON)` → `if (window.ea)` diverges from ~74 existing sites for test convenience; prefer keeping `IS_ELECTRON` and mocking `window.ea` in the spec.

### Architecture / product perspective — worth deciding before more polish

The biggest lever is the **window substrate**. Booting a second full Angular SPA to render the add-bar is what forces the transparency, global-CSS, route-detection, and duplicated-logic problems above, and is the root of blockers #1–#3. The repo already has a lighter, proven pattern for an auxiliary window: `electron/task-widget/` (vanilla renderer, no second SPA, single writer). Two reshapes would deliver most of the value at a fraction of the cost and risk:

- **(a)** make `globalAddTask` not steal focus (the only real gap vs. today), or
- **(b)** build the HUD on the task-widget pattern (single Angular writer, no second store/sync, no nav-guard gap).

On UX: the keyboard flow is strong, but two trust gaps stand out. **(1)** There's no submit confirmation — after Enter the input just clears and the "task added" feedback happens in the *hidden* main window, and the fire-and-forget IPC will silently drop the task if the main window isn't ready. Acknowledge the IPC before clearing and show an inline "Added ✓". **(2)** The full-screen screen-wide dim is off-pattern for quick capture (Spotlight/Alfred/Raycast/Things all float a small opaque panel and don't dim the screen) — and the tint is a compositor workaround rather than a design choice; a small floating panel also shrinks the Linux-transparency blast radius. Also worth: ship a sensible default shortcut (discoverability is near-zero unset), and move "(Experimental)" out of the translatable label into a badge/hint.

Happy to pair on the `destroyQuickAddWindow` + nav-guard fixes or on the task-widget-based reshape if that direction sounds good.

<sub>🤖 Consolidated from a multi-agent review (correctness / security / architecture / performance / simplicity / UX). Blocking items #1–#4 verified against source.</sub>

johannesjo:
### Follow-up: one more correctness edge case — CUSTOM-repeat dialog gets dismissed on macOS

Adding one finding to the review above. It only bites in a narrow combination (macOS + HUD triggered from an *external* app + a **CUSTOM** repeat quick-setting), and it largely dissolves under the task-widget reshape — but it's a real bug if the current window substrate stays.

**Symptom:** open the HUD from another app, enter a task with the **CUSTOM** repeat option, press Enter → the repeat-config dialog opens and is then immediately hidden along with the whole app.

**Trace:**
1. `showQuickAddWindow` (`electron/quick-add-window.ts:114`): external app was focused → `wasMainWinFocused = false`, and the main window is `hide()`-d.
2. Submit CUSTOM → main window's `onAddTaskViaIpc` (`shortcut.service.ts:122-131`) calls `window.ea.showOrFocus()` (shows/focuses the main window) and opens `DialogEditTaskRepeatCfgComponent`.
3. Focusing the main window **blurs the HUD** → `blur` handler (`quick-add-window.ts:99-101`) → `hideQuickAddWindow(false)` (not programmatic, so it isn't suppressed by `isHidingProgrammatically`).
4. `hideQuickAddWindow` (`quick-add-window.ts:173-175`): `\!wasMainWinFocused` is true → **`app.hide()`** hides the entire app — including the repeat-config dialog that was just opened.

The SP-focused trigger path is fine (`wasMainWinFocused = true` → no `app.hide()`); the non-CUSTOM consecutive-entry/Escape/blur paths are fine. `app.hide()` hides the app regardless of whether `blur` or the async dialog `open` wins the race, so the outcome holds either way.

**Fix direction (only if the current substrate stays):** suppress the `blur → app.hide()` dismissal while a CUSTOM dialog is being surfaced (e.g. treat the showOrFocus-triggered blur as programmatic), or don't `app.hide()` when the submit handed control to the main window. Under the task-widget-based reshape this whole focus dance goes away.

<sub>🤖 Verified against source (`quick-add-window.ts` focus state machine + `shortcut.service.ts` CUSTOM branch). Confidence ~90%.</sub>

hajiboy95:
@johannesjo 

Thanks for the thorough review! I've addressed all the blocking security, correctness, and lifecycle issues, as well as the code-quality/styling suggestions.

Here is the breakdown of the updates:

### 🔴 Blocking Items — Resolved
1. **App won’t quit / Zombie process:** Added `destroyQuickAddWindow()` and wired it into all quit lifecycles (`appCloseHandler` in `electron/main-window.ts` and the `before-quit` handler in `electron/start-app.ts`) to ensure the HUD window is destroyed when quitting the application.
2. **macOS Global Shortcut Null/Crash:** Safeguarded shortcut execution by introducing and using a non-throwing `getWinSafe()` accessor in `electron/main-window.ts` that safely returns `undefined` when the main window is closed.
3. **Store & Sync Engine in HUD:** Isolated the HUD by introducing `IS_QUICK_ADD_HUD` route detection. I bypassed sync initialization and timers in `SyncTriggerService` when in the HUD context, eliminating dual-sync engines or IndexedDB lock conflicts.
4. **Security & Navigation Guards:** Enabled `sandbox: true` and `disableDialogs: true` in the HUD's `webPreferences` and implemented the same navigation guards (`will-navigate`, `will-redirect`, `setWindowOpenHandler`, child window prevention) as the main window.

---

### 🟡 UX & Code Suggestions — Resolved
* **Casing, Interpolation & Truncation:** I implemented the success placeholder confirmation inline in the input bar. The placeholder interpolates the task title using sentence-casing (`Task "{{title}}" added!`) and truncates the name to 20 characters so it fits cleanly. 
* **Duration:** The success message is shown for **750ms** (half of the original 1.5 seconds) and is instantly cleared as soon as the user starts typing a new task.
* **Preloads & File Paths:** Normalized the HUD asset/preload loading paths using `normalize(join(...))` to prevent packaged Windows loading issues.
* **Idiomatic Route & Constant Refactoring:** 
  * Added `data: { page: 'quick-add' }` to the route configuration in `app.routes.ts` and added `isQuickAddRoute` to the `BodyClass` enum.
  * Refactored `app.component.ts` to read the router snapshot state root instead of doing manual URL hash parsing.
  * Replaced all `window.ea` check statements with `this.IS_ELECTRON` to maintain consistency across the codebase. Mocked `IS_ELECTRON` in `add-task-bar.component.spec.ts` to keep the specs clean and passing.

---

### ℹ️ Heads-up: Focus & Main Window Hiding Behavior
Additionally, here is a quick heads-up on a separate macOS behavior I noticed during testing (which wasn't previously addressed):

* **The Behavior:** When triggering the global quick-add shortcut while working in another application (like a web browser or editor), the main Super Productivity window is automatically minimized/hidden.
* **Why it is like this:** 
  * On macOS, focusing *any* window of an application (such as the lightweight HUD input window) automatically brings the *entire application group* (including the main window) to the foreground.
  * If the main window didn't minimize or hide, triggering the HUD shortcut would pull the large main window to the front, covering the window you were just working in and disrupting your active workspace.
  * I know this minimize behavior is a bit unfortunate, but since we cannot change this native OS window-activation constraint for now, the HUD feature is marked with the experimental flag. The trade-off is more than worth it for the productivity gains of the global HUD.

johannesjo:
### Review of `0cca57d..065f598` (inline confirmation + routing/constants refactor)

Nice cleanup overall — the `getWinSafe()` split, `destroyQuickAddWindow()` teardown across all three close paths, and the `sandbox: true` + navigation-guard hardening on the HUD window are all solid improvements. I verified the shared preload is sandbox-safe (it only touches `electron` APIs; all `process.*` reads are inside functions), so the hardening doesn't break the `window.ea` bridge.

A few things to address before merge:

**🟠 1. The "Task added!" placeholder now fires on *every* add, not just the HUD.** `_resetAfterAdd(title)` is called from both the IPC/HUD branch (`add-task-bar.component.ts:555`) and the normal in-app add path (`:617`), so the regular add-task bar in work-view/projects/planner now flashes the confirmation on every submit. Is that intended? In the HUD an in-input confirmation makes sense (a snack would render in the hidden main window, invisible to the user), but globally it's a behavior change. If HUD-only, please gate it on `isGlobalBarVariant()` / `isSubmitViaIpc()`.

**🟠 2. HUD sync suppression looks incomplete.** `getSyncTrigger$()` returning `EMPTY` only neutralizes the *dynamic* branch of `triggerSync$`. The **initial-sync** branch (`sync.effects.ts:151-167`, `afterInitialUpdateCheck$`) and the **was-just-enabled** branch (`:170-173`) aren't gated by `IS_QUICK_ADD_HUD`. Since the HUD instance shares persisted config, IndexedDB, and clientId with the main window, if `isEnabledAndReady$` becomes true there it can still run a full `sync()` — two sync engines under one clientId. Could you either gate the effect on `IS_QUICK_ADD_HUD` too, or confirm `isEnabledAndReady$` never becomes true in the HUD? (Flagging given sync changes are high-risk; I couldn't confirm the runtime path statically.)

**🟡 3. `placeholderText` computed uses `_translateService.instant()`, which won't react to runtime language changes** the way the old `| translate` pipe did (it subscribed to `onLangChange`). After a live language switch the placeholder stays stale until search-mode toggles. Keeping `| translate` for the steady-state text and only overriding transiently (`tempPlaceholder() ?? (… | translate)`) fixes this *and* lets you delete the computed.

**🟡 4. `_tempPlaceholderTimeout` isn't cleared in `ngOnDestroy`** (the sibling `_focusTimeout`/`_autocompleteTimeout` are). Low impact, but please add `window.clearTimeout(this._tempPlaceholderTimeout)` for consistency with the file's cleanup pattern.

**🟢 Smaller items:**
- `did-create-window` handler in `quick-add-window.ts` is unreachable — `setWindowOpenHandler` already returns `{action:'deny'}`, so it never fires. Either delete it, or mirror main-window's *logged* version (the current empty `catch {}` violates the no-empty-catch rule).
- Two HUD-detection mechanisms now coexist: `IS_QUICK_ADD_HUD` (hash substring, load-time) and `route.data.page === 'quick-add'`. Worth unifying — reusing `IS_QUICK_ADD_HUD` for the body class via `classList.toggle(cls, IS_QUICK_ADD_HUD)` would drop the router subscription and also avoid the initial-navigation timing gap (the constructor's `routerState.snapshot` walk isn't resolved yet, so the class lands a tick late — only masked because the HUD window is pre-created hidden).
- `T.GLOBAL_SNACK.ADD_TASK_SUCCESS` used as an input placeholder is a semantic mismatch — consider moving it under `F.TASK.ADD_TASK_BAR.*`.
- `mainWin = undefined as any` (`main-window.ts`) — typing the local as `BrowserWindow | undefined` drops the `as any`.

Nothing critical; the must-fixes are #1 (scope decision) and #4, plus a confirmation on #2.

hajiboy95:
@johannesjo #1 is intended. I actually find this is quite a nice addition since user doesn't see if a task has been added if we e.g. in the today few add a taks for tomorrow, so I'd leave it like it is. For the rest. I'll adress it later today #:D

johannesjo:
**Review of incremental range `0cca57d..065f598`** (Quick Add HUD). No blockers — the Electron hardening (sandbox, navigation guards, `getWinSafe`, cleanup wiring) is solid. A few correctness/scope items worth addressing:

### Should fix

**1. `_tempPlaceholderTimeout` is never cleared on destroy** — `add-task-bar.component.ts`
`ngOnDestroy` clears `_focusTimeout` and `_autocompleteTimeout` but not `_tempPlaceholderTimeout`. Since the HUD closes right after submit, the 500ms timer reliably fires `tempPlaceholder.set(null)` on a destroyed component. One line:
```ts
ngOnDestroy(): void {
  window.clearTimeout(this._focusTimeout);
  window.clearTimeout(this._autocompleteTimeout);
  window.clearTimeout(this._tempPlaceholderTimeout); // add
}
```

**2. The "Task added!" placeholder now shows in *every* `add-task-bar`, not just the HUD**
`_resetAfterAdd(title)` is called on both the IPC submit path and the normal `_taskService.add` path, and it's not gated by `isSubmitViaIpc()` / `isGlobalBarVariant()`. So the confirmation flash also appears in the main work-view bar, planner inline-add, and daily-summary. If that's intended as a general improvement, great — otherwise gate it to the HUD variant. Could you confirm the intended scope?

**3. `if (this.IS_ELECTRON)` is a weaker guard than the previous `if (window.ea)`**
`IS_ELECTRON` is a userAgent check; the block immediately calls `window.ea.submitAddTaskViaIpc(...)`. If the UA matches but the bridge is absent this throws. Low real-world risk, but prefer `if (this.IS_ELECTRON && window.ea)` (or keep `window.ea`).

### Nice to have

- **Live language switch:** `placeholderText` uses `translateService.instant()`, so the create/search placeholder no longer re-translates on a runtime language change (the old `| translate` pipe did). Consider depending on `onLangChange` or keeping the pipe for the static placeholders.
- **`mainWin = undefined as any`** (`main-window.ts`): the cast is unnecessary — use `mainWin = undefined`, or retype `let mainWin: BrowserWindow | undefined` to match `mainWinModule.win` and `getWinSafe()`.
- **Empty `catch {}`** in the `did-create-window` handler (`quick-add-window.ts`) — mirror the main window and `console.error(...)`.
- **500ms** feels short to read a "Task '…' added!" confirmation — ~1500ms reads better.
- **Translation key:** `GLOBAL_SNACK.ADD_TASK_SUCCESS` is used as an inline placeholder, not a snack — consider co-locating under `T.F.TASK.ADD_TASK_BAR.*`.
- **Tests:** worth adding coverage for the tempPlaceholder lifecycle (set on add / cleared on type / cleared by timeout), the `IS_QUICK_ADD_HUD` sync-trigger branch, and `destroyQuickAddWindow` idempotency.

### Defense-in-depth (optional)
`QUICK_ADD_SUBMIT` forwards its payload without a sender check or shape validation. This matches the existing IPC baseline and is local-threat-only, so not a blocker — but a `BrowserWindow.fromWebContents(event.sender) !== quickAddWin` guard + `typeof payload?.title === 'string'` check would be cheap hardening.

### Looks good
`getWin()` → `getWinSafe()` (fixes a latent throw during show/hide when the main window is gone), `sandbox: true` + `disableDialogs: true`, the navigation guards reusing `isAppOriginUrl`, and the sync-trigger short-circuit for the HUD renderer are all correct and consistent with existing patterns.

hajiboy95:
@johannesjo i guess that was a quick fire for the LLM check :D 

So I adressed your previous points. 

Indeed 500ms felt a bit short, 1500 felt a bit too long so i went for 750 :D

hajiboy95:
@johannesjo no idea why but my recent push didn't start the tests ^^. You probably have to start them manually. Probably because my commits local where later than your most recent comment.

johannesjo:
@hajiboy95 there is a conflict apparently, that is why the tests are not running.

johannesjo:
### Re-review of current head (`6da2840`) vs `master`

Thanks @hajiboy95 — re-reviewed the whole thing fresh. A lot landed well since the last round: `destroyQuickAddWindow()` wired into all three quit paths, the `getWinSafe()` split, `sandbox: true` + the navigation guards, the `_tempPlaceholderTimeout` cleanup, and the idiomatic route/`BodyClass` refactor are all correct. The earlier **sync-suppression concern (#2) is now genuinely closed** — `filter(() => !IS_QUICK_ADD_HUD)` sits before the `switchMap` in `triggerSync$` (`sync.effects.ts:132`), so it gates all three branches (dynamic / initial / was-just-enabled), not just the dynamic one. 👍

Two things to settle before merge, then a few small ones.

### 🔴 1. Merge conflict with #8403 — this is why CI won't run

The conflict is in `add-task-bar.component.ts`. Master's #8403 ("inline note field to the add-task bar") edited the same `addTask()` / `_resetAfterAdd()` region this PR touches. It's mechanical, but there's a real decision in the resolution: the new `isSubmitViaIpc()` early-`return` block must sit **after** #8403's note-collection (`taskData.notes = note`) if you want notes typed in the HUD to be forwarded too — otherwise the HUD silently drops them. Please resolve it deliberately rather than auto-accepting one side.

### 🔴 2. The HUD still boots a second store that writes the op-log (original blocker #3, residual)

Sync *triggering* is now gated, but only sync. `_startupService.init()` still runs unconditionally (`app.component.ts:261`) and `main.ts` has no quick-add branch, so the HUD renderer still boots the full NgRx store, hydration, and **all effects except sync** — `IS_QUICK_ADD_HUD` only appears in `app.component`, `sync-trigger.service`, and `sync.effects`.

Concretely, here's a confirmed op-log write from the HUD process:

1. HUD window is pre-created at startup → its store boots and hydrates.
2. `SyncTriggerService` ctor calls `setInitialSyncDone(true)` for the HUD, so `afterInitialSyncDoneAndDataLoadedInitially$` emits there.
3. → `ensureTasksDueTodayInTodayTag$` (`task-due.effects.ts:157`) runs in the HUD and dispatches `TaskSharedActions.planTasksForToday` / `planDeadlineTasksForToday` (`:280`).
4. `planTasksForToday` is a **persistent, op-log-captured action** (`op-log/core/action-types.enum.ts:64`). The op-log capture effect isn't gated for the HUD → it persists this op into `SUP_OPS` **from a second renderer process**, while `op-log/sync/lock.service.ts` assumes a single Electron instance with a per-process mutex.

So the "two autonomous sync engines" risk is gone, but "two writers to the op-log under one clientId" is not — both windows run the day-change effects against the same hydrated state and can write `SUP_OPS` concurrently with no cross-process lock. This is a data-safety invariant, not polish, so I'd like it closed (or proven harmless) before merge. The clean fix is to not boot the store/effects in the HUD at all — either a store-less HUD bootstrap, or the task-widget substrate (vanilla renderer, single writer) we discussed. If you want to keep the current substrate as a stopgap, at minimum gate the op-dispatching day-change effects on `IS_QUICK_ADD_HUD` and document why.

### 🟡 Should fix

- **`transparent: true` is still unconditional** (`quick-add-window.ts:82`), contradicting the main window's deliberate `transparent: false` "Wayland compatibility" decision and the task-widget's `transparent: !IS_MAC`. This is the compositor-tint workaround and a Linux risk — gate it per platform.
- **`any` + unvalidated IPC payload.** `QuickAddTaskPayload.taskData/remindOption/repeatCfg` are still `any` (`electronAPI.d.ts:35,38,40`) — violates the strict-no-`any` rule — and the payload is forwarded verbatim into `_taskService.add(...)` with no sender/shape check. Cheap to harden: `Partial<Task>` / `TaskReminderOptionId` / `Partial<TaskRepeatCfg>` + a `BrowserWindow.fromWebContents(event.sender) === quickAddWin` guard + `typeof payload?.title === 'string'`.
- **`if (this.IS_ELECTRON)` is weaker than the old `if (window.ea)`** (`add-task-bar.component.ts:535`) — it's a UA check, then immediately calls `window.ea.submitAddTaskViaIpc`. Use `if (this.IS_ELECTRON && window.ea)`.

### 🟢 Minor

- `placeholderText` uses `_translateService.instant()` (`:903`) → won't re-translate on a live language switch (the old `| translate` did).
- `page.scss` blanket `.mat-mdc-* { … !important }` overrides under `body.isQuickAddRoute` are the one-off-Material-override anti-pattern — most of this disappears with a store-less bootstrap; otherwise prefer a scoped CDK overlay container / `panelClass`.
- Task-creation logic is now triplicated (add-task-bar normal path, IPC path, `shortcut.service.ts`) with `isTimedRepeatTask` already diverging between them — worth one shared builder.
- Focus restoration is `darwin`-only, and the shortcut ships unbound (discoverability ≈ 0) — consider a sensible default and a Win/Linux note.

Happy to pair on the conflict resolution or on lifting the store/effects out of the HUD — that second one is the only real blocker; the rest are quick.

<sub>🤖 Re-verified against `6da2840`: sync gating (resolved), the #8403 conflict region, and the `planTasksForToday` op-log path from the HUD renderer.</sub>

hajiboy95:
@johannesjo I decided to NOT go with vanilla because: If we go the Vanilla Reshape route, the "Shift+A" in the main window would still be the powerful Angular component, but the "Global Quick Add" would be a "lite" version (just a text box). This would create a functional inconsistency between the two "Add Task" bars that might confuse users or even worse we'd try to press the functionality in this vanilla bar then it would behave different compared to the shift + A bar... Therefore only by keeping it angular it's possible to guarantee same functionality/bahviour as the bar in the app itself

hajiboy95:
@johannesjo adressed a lot of your feedback. Concerning further decoupling the Hud and the app it's actually not that easy. If I make it standalone it just starts when the user spawns the window which is very unpleasant due to long loading time. starting without any data isn't really a good behaviour either becuase the user then can't access own projects and tags. What I've currently implemented is the best I could come up with but this is also a bit new to me in terms of what options there are available and what works. spend a lot of time testing and debugging and see what actually worked.

johannesjo:
Thanks for all the work here, @hajiboy95 — genuinely. You pushed this a long way and addressed almost everything from the reviews, and your reasoning about keeping the Angular component for behavioral parity is sound.

I'm going to park this one for now though. The remaining blocker (the HUD booting a second renderer that writes the op-log) isn't polish — it's a data-safety invariant, and making it safe means a real re-architecture, not a tweak: a **store-less HUD** that keeps your Angular add-bar UI but drops the write-side (no op-log capture, no day-change effects, no hydration dispatch) and instead gets projects/tags via a read-only IPC snapshot from the main window. Creation already delegates over IPC (`submitAddTaskViaIpc`), so that path would actually *preserve* the parity you (rightly) wanted — the full store was never needed for it. That's the version I'd want to merge, but it's a meaningful chunk of work and not something I can prioritize right now.

I don't want you to keep burning time on the current substrate, since it can't be made safe without exactly the decoupling you've said is hard. If you (or anyone) wants to revisit it later along the store-less-HUD lines above, I'm happy to help scope it.

For what it's worth, the existing `globalAddTask` shortcut (bring SP to foreground → add bar) covers a lot of this use case today. Thanks again for the effort and the careful back-and-forth.

Issue #7931, opened 2026-06-01
Track B/C/D: SQLite primary store for native (durable fix for #7892)


johannesjo:
Tracking issue for the SQLite-on-native work split out of #7925 so that one can shrink to Track A (cheap safeguards) and ship-then-close. Root cause: #7892. Authoritative backlog: [`docs/sync-and-op-log/sqlite-migration-followup.md`](https://github.com/super-productivity/super-productivity/blob/master/docs/sync-and-op-log/sqlite-migration-followup.md) (read this first — descriptions below summarize each item; the doc has the gotchas, sizing, and risk notes).

## Why

#7892 (overnight total data loss on Android, no sync) is consistent with Android System WebView evicting our `SUP_OPS` IndexedDB — the WebView store is evictable in practice and `navigator.storage.persist()` is frequently a no-op there. #7902 landed the groundwork (`OpLogDbAdapter` / `OpLogTx` port, `OP_LOG_DB_SCHEMA`, a fully-implemented `SqliteOpLogAdapter` against an in-memory stand-in — not wired). This issue tracks moving the live native backend off the WebView's IndexedDB onto app-private SQLite, the durable architectural fix.

The Track A safeguards in #7925 (persistence-grant logging, debounced backup trigger, near-empty write guard) ship on their own independent track — they reduce blast radius while B/C/D is in flight.

### Track B — Finish the SQLite backend (native) ⚠️ device-gated

- [ ] **B1.** Add `@capacitor-community/sqlite` + a ~20-line `SqliteDb` wrapper over its `SQLiteDBConnection` (open `SUP_OPS` in `Directory.Data`). Bind SQLite **only** when `IS_NATIVE_PLATFORM` — ⚠️ the plugin's **web** build is WASM-SQLite persisted into IndexedDB and would reintroduce the eviction risk. Web/PWA/Electron stay on IndexedDB.
- [ ] **B2.** Validate `SqliteOpLogAdapter` against a real engine (on-device integration check, or wire sql.js with a served `.wasm` into a dedicated Karma run — the universal sql.js build statically imports `node:` modules so needs an asset/proxy entry). Parameterize the existing op-log integration harness to run a second pass against `SqliteOpLogAdapter` — this is where real-SQLite surprises surface (TEXT-key collation/ordering, NULL handling in compound ranges, autoincrement/unique/atomicity).
[...]

Issue #7892, opened 2026-05-31
💥 All data deleted overnight


cod3monk3y:
### Steps to Reproduce

1. Use Android app for 2-3 days, track lots of tasks. No sync yet setup since I'm still evaluating the app.
2. Go to bed
3. Wake up and check today's tasks

Expected Result
3. Tasks entered are still present, work history still exists

Actual results
3. App has completely reset. Zero tasks. Zero history. Absolutely blank. 100% data loss.

### URL
https://localhost/#/tag/TODAY/quick-history

### Meta Info
META: SP18.8.0_L2A __ Browser – en-US – Linux aarch64 – en-US – UA:Mozilla/5.0 (Linux; Android 16; Pixel 8 Pro Build/CP1A.260505.005; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/148.0.7778.178 Mobile Safari/537.36

### Actions Before Error
```

```

cod3monk3y:
Wait, I just got a notification from the app about a task that is due. But when I go to the app it's all still empty. No tasks, no history. No tags. Even the boards page shows "two new tags need to be created" meaning the important/urgent tags don't exist (but I certainly added them) 

I'm going to try to copy what I can via adb, in case there's a recovery process.

darius-r:
Before doing anything else, open the 'Settings' menu and look for a small link that says 'Logs'. Open this, 'copy' the logs and 'paste' them in this bug report. 

Then, try using the 'Export' option found under Settings/Sync&Backup.

This should output a JSON file which can be opened in a standard text editor. You could then search the text in this file for unique words that would only match tasks that you created. 

Alternatively, install the desktop app and Import that JSON file to see what the results are.

johannesjo:
Yes, please. Logs would be helpful!

cod3monk3y:
@darius-r Done. Logs are only from today, after the data loss, and seem to be from "this session". I did restart the app a few times after the data loss (3 or 4 times) due to the WebView bug, and thinking it might have been an initialization issue. 

[SP-logs.json](https://github.com/user-attachments/files/28444883/SP-logs.json)

Export data is completely empty.

[sp-backup-anonymized_2026-05-31_141041.json](https://github.com/user-attachments/files/28444865/sp-backup-anonymized_2026-05-31_141041.json)

Some things to note/sequence of events:

1. App was originally installed 5/27, which I assume was 18.7.0
[...]

PR #8389, opened 2026-06-14
feat(op-log): wire native SQLite op-log backend behind a flag (B1/B3/C1)

johannesjo:
Completes the device-gated wiring of the SQLite op-log migration (#7931,
root cause #7892) on top of the existing port + SqliteOpLogAdapter +
migration algorithm. All of it is inert until SUP_USE_NATIVE_SQLITE_OP_LOG
is set on a native build, so production behavior is unchanged.

- B1: add @capacitor-community/sqlite + CapacitorSqliteDb, a thin wrapper
  over its SQLiteDBConnection implementing the SqliteDb port. Loads the
  plugin via dynamic import() so its web WASM build never enters the eager
  bundle. Bakes in the bridge-perf mitigations: lastId from the run
  response (no separate last_insert_rowid()) and transaction:false so the
  adapter's BEGIN/COMMIT/ROLLBACK is the single transaction in force.
- B3: flag-gated OP_LOG_DB_ADAPTER_FACTORY override. On native, when the
  opt-in flag is set, both stores get their own adapter over ONE shared
  CapacitorSqliteDb (one SQLite file, all tables), mirroring the shared IDB
  connection. Default off; never bound on web/PWA/Electron.
- C1: one-time first-launch IDB->SQLite migration bootstrap. Runs once on
  first adapter init(): creates the schema, then copies the legacy SUP_OPS
  IndexedDB via the verify-before-commit migrateOpLogBackend, guarded by a
  meta marker table, a non-empty-dest check, and a best-effort
  indexedDB.databases() presence check. The marker is set only after a
  verified copy, so a failed run retries; the IDB copy is kept as a >=1
  release fallback.

Unit-covered in native-sqlite-backend.spec.ts (idempotency, empty source,
shared connection, flag gating) against real sql.js. npx cap sync + the
on-device run and staged rollout (C2) remain.


johannesjo:
## What's left before this is production-safe

The backend is now **Android default-on** (no opt-in flag; `IS_ANDROID_NATIVE`), with an in-session IndexedDB fallback, a durable in-SQLite "migrated" marker, and timeouts on every native bridge call. iOS and web/PWA/Electron stay on IndexedDB. Latest commits: native plugin registration + `feat(op-log): enable native SQLite op-log on Android by default`.

> ⚠️ The PR title still says "behind a flag" — worth updating, the flag is gone.

### ✅ Done & verified (unit/CI)
- Android-only gate; opt-in flag + fail-count + localStorage "authoritative" mirror removed.
- Durable authority = in-SQLite `sup_op_log_meta` marker (or a non-empty dest), never evictable localStorage.
- In-session IDB fallback on a **provably pre-migration** bootstrap failure (lossless, app still boots); **fails loudly** post-migration / when it can't prove pre-migration (never serves a stale snapshot).
- Every native call timeout-bounded (open 15s, per-statement 60s, `isDatabase`, `closeConnection`) → a wedged connection can no longer hang boot.
- Migration verify-before-commit via `tx.count` aggregate (no blob re-scan); failing `BEGIN` now resets the wedged connection.
- 548 op-log persistence specs green in one Karma session; full lint green; two adversarial reviews — headline data-loss paths clean.

### 🚧 Hard gate before ramping rollout — on-device Android validation
Everything above is tested with in-process `sql.js`; **the real `@capacitor-community/sqlite` plugin / JS↔native bridge has never run on a device.** Required before flipping rollout past ~1%:
- [ ] `npx cap sync android`, build, run on a real device.
- [ ] Migrate a **populated** install end-to-end; task/project counts + a sync round-trip match before/after.
- [ ] Force a failed/wedged open → confirm the app **boots on the IDB fallback** (not a blank screen).
- [ ] WebView force-reload mid-session (connection-reuse path).
- [ ] A **large** account (tens of thousands of ops) → confirm first-launch migration time is tolerable.

### ⚠️ Rollout discipline (no kill-switch + no telemetry)
This is a fleet-wide, one-shot, first-launch migration with **no remote kill-switch** and **no analytics** (privacy-first). The breadcrumbs (`opLogSqliteMigration` / `…MigrationFailed` / `…FellBackToIdb`) only live in the user's exportable log — not sent anywhere.
- [ ] Ramp **very slowly** via Play Console staged rollout (1% → watch → 5% → …), not a fast staged rollout.
- [ ] Use Play Console **crash/ANR vitals** as the primary signal (boot hangs/bricks show there; silent data issues do not).

### Residual risks (accept knowingly)
- Migration fidelity is verified **structurally** (per-store counts + ops high-water seq + vector clock), not byte-for-byte content of every op.
- **Slow boot on huge accounts** until the deferred batched-insert work lands (N ops = N bridge crossings today).
- A plugin that breaks **after** a successful migration → fails loudly (brick) rather than serve stale data — correct, but a brick caught only by build testing.
- **iOS is intentionally not covered** — those users stay on the evictable IndexedDB; this PR fixes Android only.

### Deferred (correctness-unaffected, on-device-measurable only)
- [ ] Batched `executeSet` inserts + seq-windowed `ops` read — implement only if on-device profiling shows the first-launch migration is too slow / OOMs on large accounts.
- [ ] Track D cleanup once SQLite is the sole native backend (remove `adoptConnection` bridge + IDB fallback, after ≥1 release).

johannesjo:

johannesjo:

PR #8390, opened 2026-06-14
fix(schedule): replace circle icon with wb_sunny for today button

### LINKED ISSUE #7537: 💡 Replace Today button circle in Schedule view

hajiboy95:
closes #7537 and makes #7538 obsolete


johannesjo:
Thank you very much! <3

Issue #8392, opened 2026-06-14
💡 Feature Request: Advanced Date & Timeframe Filtering for Custom Boards (Planning & Deadlines)


hajiboy95:
### Problem Statement
Currently, the filter configuration for custom boards allows users to filter by "Planning status" (`Planungsstatus`), but it is limited to a simple boolean check: *All*, *Planned* (`Geplant`), or *Unplanned* (`Nicht geplant`).

There is no way to filter tasks by *when* they are planned. This prevents users from creating dynamic, time-based boards (e.g., a specific "Planned for Today" or "Next Week" dashboard). Furthermore, there is currently no capability at all to filter tasks based on whether they have a **Deadline**.

### Possible Solution
Implement advanced date and timeframe parameters for the board filters. This can be broken down into two tasks:

* [ ] **Task 1: Extend the Planning Status Filter**
* When `Planungsstatus` is set to `Geplant`, dynamically reveal an optional sub-menu/field to define the timeframe.
* **Options should include:** * **Relative presets & relative ranges:** Today, Tomorrow, Next Week (range), Next Month (range), or dynamic intervals (e.g., "Next X days").
* **Custom Date Range:** Provide inputs for concrete specific dates or a manual start/end date range.

* [ ] **Task 2: Add a new "Deadline" Filter**
* Introduce a new top-level filter block (e.g., `Deadline: Alle / Hat Deadline / Keine Deadline`).
* If a user selects "Has Deadline", reveal the exact same timeframe options mentioned above (relative presets, relative ranges, and custom date ranges).

### Additional Context

* This enhancement would unlock incredibly powerful custom workflows, allowing users to build dedicated "Upcoming Deadlines," "Today's Agenda," or "Overdue Tasks" boards.
* Visually, the timeframe dropdown/date-picker should only appear conditionally when "Geplant" or "Hat Deadline" is selected, keeping the default UI clean and uncluttered.

symonbaikov:
I really like this one) Will work on it today.

hajiboy95:
@symonbaikov please look into the history, i yesterday opened a PR for this one :D

PR #8393, opened 2026-06-14
Feat/8392 board filters

hajiboy95:
## Problem

Currently, the filter configuration for custom boards allows users to filter by "Planning status" (`Planungsstatus`), but it is limited to a simple boolean check: *All*, *Planned* (`Geplant`), or *Unplanned* (`Nicht geplant`). There is no way to filter tasks by *when* they are planned or to filter tasks based on whether they have a **Deadline**, preventing users from creating dynamic, time-based custom boards (e.g. "Planned for Today", "Overdue Tasks").

Furthermore, dropping a task onto a panel with specific scheduled date or deadline timeframe filters would either open a scheduling dialog (if unscheduled/no-deadline) or not update the task's schedule at all (if already scheduled outside the timeframe), causing it to instantly disappear from the column after the drop.

This resolves issue #8392.

## Solution

Implemented advanced date and timeframe parameters for the board filters along with automatic drop-replanning:
1. **Timeframe Presets for Scheduled State:** Added timeframe choices when `scheduledState` is set to `Scheduled` (`Geplant`):
   - Relative presets: Today, Tomorrow.
   - Relative ranges: Next Week, Next Month, dynamic intervals ("Next X days"), or "At least X days in the future".
   - Custom range: premium button-based date pickers that support half-bounded ranges (an empty start date means "everything before", and an empty end date means "everything after").
2. **New Deadline Filter:** Introduced a new top-level filter block (`Deadline: All / Has Deadline / No Deadline`). If "Has Deadline" is selected, the same timeframe options (relative presets, ranges, and custom ranges) are revealed.
3. **Pipes & Filtering Logic:** Updated the custom board filters in `BoardPanelComponent` to evaluate logical due dates/deadlines using ISO strings and `DateService` to correctly handle logical timezone offsets.
4. **Automatic Drop-Replanning:** Updated the task drag-and-drop handler (`_checkToScheduledTask` and `_checkDeadlineState`) so that dropping a task onto a column with a restricted scheduled date or deadline timeframe automatically reschedules the task (via `PlannerActions.planTaskForDay` and `TaskSharedActions.setDeadline` respectively) to the closest valid date within that range without prompting. Unscheduled/no-deadline tasks dropped onto unrestricted `ALL` columns still prompt with manual picking dialogs, and dropping onto "No Deadline" columns automatically removes the deadline.
5. **Premium Button-Based Date Picker:** Replaced standard text inputs for custom ranges with a button-based date picker (`date-btn`) matching the style of the start-date configuration in recurring tasks:
   - Clicking the buttons launches the native `DialogScheduleTaskComponent` (with time and reminder pickers hidden via new `showTime` and `showReminder` options).
   - Added mutual bounds validation: selecting a start date restricts the calendar's `minDate` on the end date picker, and vice versa, using `props.minDateKey` and `props.maxDateKey`.
   - Displays a clean formatted date string (localized to the current locale) on the button, with a close/clear button to easily unset it.
6. **Wiki Documentation:** Updated `docs/wiki/4.05-Board-View.md` to document the new timeframe and deadline filters.
7. **Testing:** Extended the unit tests in `boards-form.const.spec.ts` and `board-panel.component.spec.ts` to cover the new Formly configurations, timeframe filtering, and both scheduled date & deadline drop-replanning/clearing logic.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks @hajiboy95 — this is a solid, well-tested feature and the auto-replanning-on-drop UX is a nice touch. I ran a multi-pass review (correctness / security / architecture / performance / simplicity).

**No correctness or security bugs found.** The logical-day offset handling is right, it's DST-safe, sync-safe (new `BoardSrcCfg` fields are optional and the new `BoardPanelCfgDeadlineState` enum doesn't mutate a synced union), and the Formly `hide` expressions are static compile-time constants so there's no eval/injection surface. Board-panel specs pass and lint is clean.

A few things I'd like addressed before merge, then some optional cleanups.

### Please address (cheap, and these are repo-convention violations)

1. **Dead parameter.** `startOfNextDayDiffMs` is threaded through `isDateInTimeframe` and `adjustDateToTimeframe` but never read in either body — the offset is already applied upstream in `getTaskScheduledDateStr`/`getTaskDeadlineDateStr`. Please drop it; it misleadingly implies the string comparison is offset-aware. (`board-panel.component.ts:106,225`)

2. **Material override in shared SCSS.** `formly-date-btn.component.scss` restyles `mat-stroked-button` (`text-align`, `height`), targets `.mat-icon`, and uses hardcoded px instead of spacing tokens. Per CLAUDE.md / the styling guide we don't locally restyle Angular Material in shared `ui/` components — the sibling `formly-date-picker` ships no SCSS at all. Please remove the overrides, or promote a reusable variant to the shared style layer. (`formly-date-btn.component.scss:13-24`)

3. **Hoist per-task timeframe bounds out of the filter loop.** In the `tasks` computed, the range presets recompute `getFutureLogicalDateStr` (Date + string allocations) for *every task*, even though the bounds are constant for the whole panel. On large boards that's O(panels × tasks) of throwaway allocations per change. Compute `{ start, end }` once before `.filter()`. (`board-panel.component.ts:106-161,376-497`)

### Worth a follow-up refactor (lots of duplication here)

4. The timeframe helpers take 8–10 positional params (`isDateInTimeframe`, `getClosestDateInTimeframe`, `adjustDateToTimeframe`) — please group into an options object (repo rule: max 3 params). (`board-panel.component.ts:106,163,225`)

5. The timeframe→`{start,end}` cascade is written twice (`isDateInTimeframe` + `getClosestDateInTimeframe`), the 8-option `select` array is copy-pasted scheduled↔deadline in `boards-form.const.ts`, and `_checkToScheduledTask`/`_checkDeadlineState` are near-identical. Extracting a single `resolveTimeframeBounds()` + a `TIMEFRAME_OPTIONS` const + a small field factory would collapse ~150 lines and remove the risk of the two copies drifting. Ideally the pure helpers move to `boards.util.ts` with their own spec — the component file is now 763 lines and the helpers are currently only tested indirectly.

### Minor / optional

6. `_checkDeadlineState` dispatches `removeDeadline` even when the task has no deadline → a redundant no-op sync op on every drop onto a "No Deadline" column. Guard with `if (task.deadlineDay || task.deadlineWithTime)`. (Same shape as the pre-existing unconditional `unscheduleTask`, which has the same issue.) (`board-panel.component.ts:755`)

7. The date-btn dialog's quick-access buttons (Today/Tomorrow/…) bypass the custom-range `minDate`/`maxDate`, so an end-before-start range can be selected. It's harmless (just yields an empty column), but the mutual bound validation is then only half-enforced. (`dialog-schedule-task.component.ts:419`)

8. Heads-up on naming: `NEXT_WEEK` here is a rolling today+7 window, whereas the existing task-view-customizer's "Next Week" means the next calendar week — two filters with the same label behaving differently. Might be worth aligning the wording.

Nothing above is a correctness blocker — really nice work overall. 🙌

hajiboy95:
@johannesjo 
Thanks for the detailed review! I've addressed all of your feedback and pushed the updates:

### Addressing Convention Violations
1. **Dead parameter:** Dropped the `startOfNextDayDiffMs` parameter from `isDateInTimeframe` and `adjustDateToTimeframe` (which now uses a clean options configuration).
2. **Material overrides in shared SCSS:** Removed all button and icon overrides from [formly-date-btn.component.scss](file:///Users/Finbridge/Desktop/Second_Brain/3. Ressourcen/Github/super-productivity/src/app/ui/formly-date-btn/formly-date-btn.component.scss). Instead, promoted a reusable `.formly-btn-form-field` class to [_overwrite-material.scss](file:///Users/Finbridge/Desktop/Second_Brain/3. Ressourcen/Github/super-productivity/src/styles/components/_overwrite-material.scss) to handle left-alignment, padding, and form-field style heights safely.
3. **Hoist bounds out of filter loop:** Refactored the `tasks` computed property in [board-panel.component.ts](file:///Users/Finbridge/Desktop/Second_Brain/3. Ressourcen/Github/super-productivity/src/app/features/boards/board-panel/board-panel.component.ts) to compute `scheduledBounds` and `deadlineBounds` once before filtering the task list.

### Refactoring & Deduplication
4. **Group timeframe helpers params:** Grouped timeframe bounds parameters in `adjustDateToTimeframe` into a single `AdjustDateOptions` object to follow the max-3-params rule.
5. **Consolidate timeframe calculations:** 
   - Moved all logical-date and timeframe helper functions to [boards.util.ts](file:///Users/Finbridge/Desktop/Second_Brain/3. Ressourcen/Github/super-productivity/src/app/features/boards/boards.util.ts) and added 39 comprehensive unit tests in [boards.util.spec.ts](file:///Users/Finbridge/Desktop/Second_Brain/3. Ressourcen/Github/super-productivity/src/app/features/boards/boards.util.spec.ts).
   - Extracted `TIMEFRAME_OPTIONS` to [boards.const.ts](file:///Users/Finbridge/Desktop/Second_Brain/3. Ressourcen/Github/super-productivity/src/app/features/boards/boards.const.ts).
   - Created a type-safe `buildTimeframeFields` factory function in [boards-form.const.ts](file:///Users/Finbridge/Desktop/Second_Brain/3. Ressourcen/Github/super-productivity/src/app/features/boards/boards-form.const.ts) to dynamically construct timeframe form configurations, collapsing ~150 lines of duplicate configs.
   - Refactored `NEXT_MONTH` computation to use calendar-month addition (jumping to same date next month and clamping end-of-month overflows, e.g., Jan 31 -> Feb 28).

### Minor & UX Polishes
6. **Guard redundant sync actions:** Guarded the fallback `removeDeadline` and `unscheduleTask` sync actions so they only dispatch if the dropped task actually has a deadline or scheduled date.
7. **Planning dialog clamping:** Clamped quick-access options (Today, Tomorrow, etc.) in [DialogScheduleTaskComponent](file:///Users/Finbridge/Desktop/Second_Brain/3. Ressourcen/Github/super-productivity/src/app/features/planner/dialog-schedule-task/dialog-schedule-task.component.ts) to the provided `minDate` and `maxDate` range constraints using day-level comparisons.
8. **Consistent naming prefix:** Aligned rolling timeframes with the **"Within"** prefix in [en.json](file:///Users/Finbridge/Desktop/Second_Brain/3. Ressourcen/Github/super-productivity/src/assets/i18n/en.json) (e.g., *Within 7 days*, *Within 1 month*, *Within X days*) to make it clear they represent rolling windows instead of calendar boundaries.

johannesjo:
Thanks for this — it's a substantial, well-tested feature (the spec coverage in `board-panel.component.spec.ts` / `boards.util.spec.ts` is appreciated). I ran a multi-perspective review (correctness, security, architecture, performance, simplicity). No blocking bugs and no sync/data-corruption risk found. A couple of small things worth addressing before merge, then some optional cleanups.

### 🔴 Should fix before merge

**1. `any` type violates the strict-TS rule** — `boards-form.const.ts:24`
```ts
stateOptions: { value: any; label: string }[];
```
Both call sites pass numeric enum values. Make the factory generic (`<TState extends number>`) or at minimum use `value: number`.

**2. Non-deterministic drop replanning** — `board-panel.component.ts`, `drop()`
`_checkToScheduledTask()`, `_checkDeadlineState()`, and `_checkBacklogState()` are invoked **concurrently and un-awaited**. For a panel that filters *both* a scheduled timeframe and `deadline = Today`, `PlannerActions.planTaskForDay` and `setDeadline` (with its auto-plan fields) can both write `dueDay` from the same pre-drop snapshot, in non-deterministic order. The reducer's "honor an existing future schedule" guard makes actual data-loss unlikely, but the ordering is fragile. Suggest `await`-ing them sequentially so the outcome is deterministic.

### 🟡 Nice to have

**3. Skip the redundant store reads on the common path** — `board-panel.component.ts`
`_checkBacklogState` early-returns *before* its `selectTaskById` read when the filter is `All`, but `_checkToScheduledTask` / `_checkDeadlineState` read the task first and then no-op. Default panels (scheduled = All, deadline = All) pay two needless async selector reads on every drop/add. Mirror the backlog early-return at the top of both methods.

**4. Narrow the persisted date type** — `boards.model.ts`
`scheduledCustomStart/End` and `deadlineCustomStart/End` are typed `string | Date | null`, but the only writer (`FormlyDateBtnComponent`) always stores a `YYYY-MM-DD` string (or `null`). Since this is persisted/synced state, narrowing to `string | null` removes the dead `instanceof Date` branches in `normalizeDateStr` and `formly-date-btn`, and avoids a future foot-gun. (It's safe today — `normalizeDateStr` already coerces — so this is cleanup, not a bug.)

**5. Reuse existing helpers**
- `getFutureLogicalDateStr` / `getFutureLogicalMonthDateStr` (`boards.util.ts`) inline the body of the existing `src/app/util/parse-db-date-str.ts` — use that instead.
- `getClosestDateInTimeframe` can collapse to a 3-line min/max clamp (lexicographic comparison is valid for `YYYY-MM-DD`):
  ```ts
  const d = currentVal ?? todayStr;
  if (bounds.start && d < bounds.start) return bounds.start;
  if (bounds.end && d > bounds.end) return bounds.end;
  return d;
  ```

**6. Minor**
- New `_check*` methods use the deprecated `toPromise()` — prefer `firstValueFrom`.
- The `tasks` computed reads `_dateService.todayStr()` (a plain method, not a signal), so a board left open across the day boundary can show stale "Today/Tomorrow" results until other state changes trigger a recompute. Low-priority edge case.
- `task-view-customizer` already has a scheduled/deadline timeframe filter (`FILTER_SCHEDULE`). Semantics differ (calendar vs. rolling), so not a drop-in dedup — just flagging that two timeframe systems now coexist.

### ✅ Confirmed good
- `.formly-btn-form-field` lives in the shared style layer (`_overwrite-material.scss`) — that's the sanctioned place for a reusable Material variant, not a rule violation. (One nit: double-check the `!important` on the icon margin is actually needed.)
- The new `formly-date-btn` is a genuine generalization (button → schedule dialog with quick-access), not a duplicate of the inline `date` field.
- `buildTimeframeFields` is a justified DRY extraction; timeframe bounds are correctly hoisted out of the per-task filter loop; month-boundary and DST date math all check out; the quick-access refactor also fixes a real pre-existing `tDate` aliasing/mutation bug 👍

Nice work overall — items 1 and 2 are the only ones I'd consider gating.

johannesjo:
One more I missed — worth a quick guard before merge:

**`daysVal` is unvalidated.** The `Within X days` / `At least X days` field is `min:1` in the UI only, so a negative or very large value still binds. It flows straight into `resolveTimeframeBounds` → `getFutureLogicalDateStr`, where a huge value yields an `Invalid Date` → `getDbDateStr` → `"NaN-NaN-NaN"`. On the `AT_LEAST_DAYS_FUTURE` drop path that string gets dispatched through `setDeadline` / `planTaskForDay` and synced to other devices. Please clamp in `resolveTimeframeBounds` — e.g. `Number.isFinite` guard + `Math.max(1, Math.min(MAX, Math.floor(daysVal)))` — and/or sanitize `*DaysVal` in `sanitizePanelCfg`.

Optional: the drop-replan currently changes a task's scheduled date/deadline with no toast (verified — neither `planTaskForDay` without `isShowSnack` nor `setDeadline` surfaces anything) and syncs the change. Consider a confirmation snack so a board reorg doesn't silently rewrite dates.

hajiboy95:
@johannesjo feedback has been adressed

  🔴 Mandatory Fixes
   1. Strict Types: Replaced value: any with value: number in boards-form.const.ts for the timeframe field factories to adhere
      strictly to the project's TypeScript standards.
   2. Deterministic Drop Replanning: Modified drop() and afterTaskAdd() in board-panel.component.ts to await the state checks
      sequentially (_checkToScheduledTask, _checkDeadlineState, _checkBacklogState) ensuring predictable ordering when multiple
      conditions intersect.

  🟡 Optional Enhancements & Cleanup
   3. Redundant Store Reads Avoided: Introduced early returns at the top of _checkToScheduledTask and _checkDeadlineState for the
      All filters, sparing default board panels from needless async selector executions.
   4. Persisted Date Type Narrowed: Constrained the custom date range fields in boards.model.ts to string | null, aligning with how
      FormlyDateBtnComponent stores its ISO strings and keeping type contracts tighter. 
   5. Helper Reuse & Simplification:
      - Swapped inline date math in boards.util.ts for the pre-existing parseDbDateStr helper.
      - Re-wrote getClosestDateInTimeframe from nested conditionals into a succinct bounding clamp logic utilizing lexicographical
        date string comparisons.
   6. Minor Fixes:
      - Exchanged the deprecated toPromise() pipeline reads for firstValueFrom in board-panel.component.ts.
      - Validated daysVal through a Math.max(1, Math.min(365, ...)) clamping check within both resolveTimeframeBounds() and
        sanitizePanelCfg(), neutralizing the threat of invalid strings ("NaN-NaN-NaN") hitting the server.
        
 Snackbar functionality has also been added

johannesjo:
Reviewed the latest commit `9ed03bd` ("address review feedback"). The changes are solid — verified each one against the post-commit sources (including brute-forcing the date-math refactors for equivalence). Nothing blocking; one small cleanup plus two pre-existing items adjacent to the retyped fields that are worth a look before merge.

**✅ Verified correct**
- `.toPromise()` → `firstValueFrom()` (4 sites): safe — the selectors are `BehaviorSubject`-backed and piped with `first()`/`take(1)`, so they always emit synchronously and never complete empty; `firstValueFrom` can't throw `EmptyError`, and the `if (!task)` guards still handle `undefined` identically.
- `await` + early-returns on `_checkToScheduledTask`/`_checkDeadlineState`/`_checkBacklogState`: behavior-preserving (the `state === All` early returns are no-ops given the enums), and sequential awaiting is the safer choice here — please **don't** switch these to `Promise.all`, since they dispatch order-dependent mutations and each re-reads fresh state.
- `getFutureLogicalMonthDateStr` / `getClosestDateInTimeframe` refactors: bit-for-bit equivalent across month-end clamping, year rollover, and leap years.
- `sanitizeDaysVal` clamp: good fix — without it a huge/`Infinity`/negative `daysVal` overflows `Date` and produces `"NaN-NaN-NaN"`, which would otherwise get persisted into synced task state.
- `{ value: any }` → `{ value: number }` and `string | Date | null` → `string | null`: correct strict-TS tightenings.

**🟡 Minor (this commit)**
- `sanitizeDaysVal` is defined as an identical closure twice (`boards.util.ts:85` and `:269`), with duplicated `1`/`365`/`7` literals. Keeping the clamp at both write and read is fine (raw-data paths bypass `sanitizePanelCfg`), but worth hoisting to one module-level helper + named constants so the two can't drift:
  ```ts
  const MIN_DAYS = 1, MAX_DAYS = 365, DEFAULT_DAYS = 7;
  const clampDaysVal = (v: number | undefined): number =>
    Number.isFinite(v) ? Math.max(MIN_DAYS, Math.min(MAX_DAYS, Math.floor(v as number))) : DEFAULT_DAYS;
  ```
- Optional: `_checkDeadlineState` guards only `=== All`, but `deadlineState` is optional in the model, so `undefined` falls through to a wasted store read. Matching `_checkBacklogState`'s `if (!panelCfg.deadlineState || … === All) return;` would be consistent.

**⚠️ Pre-existing, but adjacent to the retyped custom-date fields (worth confirming before merge)**
- The custom-range fields use `type: 'date-btn'` (`boards-form.const.ts:74,85`), but only `'date'` is registered in `formly-config.module.ts` (`date-btn` appears nowhere else). Unknown Formly types throw — so the **CUSTOM_RANGE date pickers likely don't render**. Looks like a typo for `'date'`.
- `DatePickerInputComponent`'s CVA emits a `Date`, and `sanitizePanelCfg` doesn't normalize `scheduled/deadlineCustomStart/End`. Now that those fields are typed `string | null`, a stored `Date` mismatches the type — reads survive (`normalizeDateStr` coerces), but the boards typia validator runs on raw-data/legacy-migration paths *before* sanitize and throws on shape mismatch. Suggest normalizing the four custom-date fields through the existing `normalizeDateStr` inside `sanitizePanelCfg` — it makes the type honest and pairs naturally with the narrowing in this commit.

Overall: 👍 on the commit.

hajiboy95:
@johannesjo 
Summary of Changes

   1. Refactored boards.util.ts:
       * Hoisted the sanitizeDaysVal closure into a module-level clampDaysVal helper.
       * Defined named constants MIN_DAYS, MAX_DAYS, and DEFAULT_DAYS for clarity and to prevent
         literal duplication.
       * Updated sanitizePanelCfg to normalize custom-date fields (scheduledCustomStart,
         scheduledCustomEnd, deadlineCustomStart, deadlineCustomEnd) using normalizeDateStr.
         This ensures they are stored as canonical YYYY-MM-DD strings.
       * Applied clampDaysVal in resolveTimeframeBounds.

   2. Narrowed Types in boards.model.ts:
       * Updated the custom-date fields in BoardPanelCfg to string | null, making the type
         system honest about the canonical format and ensuring compatibility with the typia
         validator.

   3. Improved Guard in board-panel.component.ts:
       * Updated _checkDeadlineState to include a guard for !panelCfg.deadlineState, matching
         the pattern in _checkBacklogState and preventing unnecessary store reads.

   4. Fixed Formly Type Typo in boards-form.const.ts:
       * Changed type: 'date-btn' to type: 'date' for custom-range fields to ensure the standard
         date picker is used and to avoid potential rendering issues.

hajiboy95:
A slight adjust: 
Places the within 7 days and within 1 month (since it's covered by within X days) agains within this/next week and within this/next month. It's actually much more intuitive now. :)

johannesjo:
Did another pass on the latest commits. One thing I need to walk back from my previous review, plus a heads-up on CI.

### 🔴 My earlier `date-btn`→`date` suggestion was wrong — please revert it

In my last round I said the custom-range `type: 'date-btn'` looked like *"a typo for `'date'`"* because *"only `'date'` is registered… date-btn appears nowhere else."* That was my mistake — **this PR does register it**, in `formly-config.module.ts`:

```ts
{ name: 'date-btn', component: FormlyDateBtnComponent },
```

At the commit I reviewed (`9ed03bd`) `date-btn` was both registered and used (`boards-form.const.ts:74,85`), so the custom-range pickers were rendering correctly. Switching them to `type: 'date'` (`31a927de`) still works functionally, but it orphaned the whole `FormlyDateBtnComponent` feature:

- `formly-date-btn.component.{ts,html,scss}` (~166 lines) — registered but `type: 'date-btn'` is now used nowhere
- `.formly-btn-form-field` in `_overwrite-material.scss` — only used by `formly-date-btn.component.html`
- the `showTime`/`showReminder`/`maxDate` plumbing in `dialog-schedule-task` + `datetime-picker` — only `FormlyDateBtnComponent` ever passed `showTime:false`
- **mutual-bounds validation** (`props.minDateKey`/`maxDateKey`) — only `FormlyDateBtnComponent` reads those; `FormlyDatePickerComponent` ignores them, so that feature is silently dead

So the button-based picker UX and the start/end mutual-bounds validation both got dropped, and ~200 lines now ship unused.

**Could you revert to `type: 'date-btn'`?** That restores both intended features and removes the dead code, and it was correct all along — apologies for the detour. (If you'd actually rather keep the plain `date` picker, that's fine too — but then let's delete `FormlyDateBtnComponent` + its registration + `.formly-btn-form-field` + the `showTime`/`showReminder`/`maxDate`/`minDateKey`/`maxDateKey` plumbing so nothing dead ships.)

### ℹ️ The red CI check is not your PR

The failing E2E (`recurring-start-date-epoch-bug-6860.spec.ts:91 › … via helper`) is pre-existing test rot on `master` — it's not modified by this PR. It hardcodes `15/06/2026`, and the recurring dialog disables past dates, so the calendar cell is `mat-calendar-body-disabled` and the click times out for any run on/after 2026-06-16. Your earlier green runs just happened to execute before local midnight on the 15th. I'll fix that test separately on `master`; don't let it block you here.

### 🟡 Minor
- `boards.util.ts` `getFutureLogicalMonthDateStr` is dead now (the month logic moved to `new Date(y, m, …)` calendar math) — can be removed along with its spec.

Everything else in the recent commits checks out — the timeframe math, bounds hoisting, clamping, and sanitize/normalize are all good. 🙌

johannesjo:
Thanks a lot to both of you — @symonbaikov (#8413) and @hajiboy95 (#8393) — you landed near-identical solutions for this within a day of each other, and both are genuinely solid. I'd like to converge on one so the effort isn't wasted.

They differ in two meaningful ways:

1. **Scope / risk.** #8413 is self-contained in `boards/`: display-only filtering, a clean nested config model, strong sanitization, and good unit tests for the date math — no shared-component changes. #8393 is broader: it also touches the shared schedule dialog + datetime-picker, adds a button-style date picker and more presets ("next X days", "≥ X days out", half-bounded ranges), and — the big one — **auto-replans a task to the nearest valid date when you drop it into a timeframe panel** instead of letting it vanish.

2. **A product decision underneath it.** The auto-replan UX is nicer, but it means drag-and-drop now *mutates* a task's due date / deadline. #8413 keeps drops as pure filtering. We should decide which behavior we want before picking a base.

My leaning: use #8413 as the foundation (smaller, lower-risk, clean model + sanitization) and layer #8393's extra presets — and, if we want it, the drop-replan logic — on top. But I'm open if you feel strongly otherwise.

Could you two sync on who carries the combined PR? Happy to review once you've aligned. Thanks again, and sorry you doubled up here.

hajiboy95:
continued in #8413

Issue #8328, opened 2026-06-12
Cleanup: sync packages — dead error class, dead ports, zero-consumer compat exports, duplicated enum, stranded compact/ directory


johannesjo:
## Verified items (repo-wide grep, src/ + packages/, non-spec)

1. **`SyncStateCorruptedError` — dead end-to-end.** Never constructed or `instanceof`-checked in production (only its own spec); the app shim (`src/app/op-log/core/sync-state-corrupted.error.ts`) is imported only by its spec. Comments at `remote-ops-processing.service.ts:467` and `operation-log-hydrator.service.ts:579` (and the sync docs) describe a rethrow-classification behavior that does not exist — misleading in the highest-risk path. Delete (incl. `packages/sync-core` barrel line) **or** actually throw it where the comments claim.
2. **Deprecated full-state compat exports (`packages/sync-core/src/full-state-op-types.ts:26-43`, barrel :69-73) — zero consumers.** App uses its own `createFullStateOpTypeHelpers` instance (`operation.types.ts:28-35`); the server uses its own local set (`super-sync-server/src/sync/sync.types.ts:15-22`). Delete the compat instances, keep the factory.
3. **`SyncConfigPort` (`packages/sync-core/src/ports.ts:60-81`) — port-to-nowhere.** No sync-core consumer; its lone implementation `GlobalConfigService.getSyncConfig` (:199) is called only by a spec. Remove both. Also `ConflictUiPort.notify?` (:107) is implemented nowhere — drop the member.
4. **`packages/sync-providers/src/index.ts` — 0-byte leftover** from the root-barrel removal (not in tsup entries or the exports map). Delete.
5. **shared-schema → sync-core vector-clock re-export:** consumers are `src/app/core/util/vector-clock.ts:2-8` (app already depends on sync-core — trivially retargetable), `operation-log.const.ts:221` (spec-only consumers), and `packages/super-sync-server/src/sync/sync.types.ts:8-13` — the server's package.json depends only on `@sp/shared-schema`, so the re-export is currently **load-bearing for the server**. Removing the compat edge = retarget the two app files + add `@sp/sync-core` to the server deps + drop the re-export + update `package-boundaries.md`.
6. **App `VectorClockComparison` enum (`src/app/core/util/vector-clock.ts:39-44`)** duplicates the sync-core string union (forces the cast at ~:143) — the only true type duplication left from the package migration. Import the union (string values identical).
[...]

PR #8408, opened 2026-06-15
refactor: remove SyncConfigPort dead port and ConflictUiPort.notify is dropped

forge34:
## Problem
Address #3 in #8328 

## Solution
- remove exports from `packages/sync-core/src/index.ts:114-115`
- remove `SyncConfigPort`, `SyncConfigSnapshot` and `ConflictUiPort.Notify?` from `packages/sync-core/src/ports.ts:66-81:107`
- remove `getSyncConfig` from `src/app/features/config/global-config.service.ts:198-219`

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo (commented):
Reviewed this as a dead-code removal — the core removal is **correct, safe, and build-clean**. I verified across `src/`, `packages/`, `e2e/`, server and plugins:

- `SyncConfigPort` / `SyncConfigSnapshot` / `GlobalConfigService.getSyncConfig()` have no consumers other than the lines removed here. `sync-core` never actually calls `getSyncConfig()` — the port was exported and implemented but never wired up, so it's genuinely dead.
- `ConflictUiPort.notify?()` has no callers, and the only implementer (`SyncImportConflictDialogService`) never implemented it — safe to drop.
- No collision with the unrelated `IssueSyncAdapter.getSyncConfig(cfg): FieldSyncConfig` (different interface/signature) — untouched.
- The deleted spec only tested `getSyncConfig`; `selectSyncConfig` stays covered in `global-config.reducer.spec.ts`. Import cleanup (`firstValueFrom`) is correct.

Two small completeness nits so the cleanup doesn't leave new dead/misleading code behind:

**1. Orphaned (now misleading) doc comment — `packages/sync-core/src/ports.ts:60-65`**

The `SyncConfigSnapshot` interface is removed but its doc comment is kept, so it ends up floating directly above the unrelated `ConflictUiDialogRequest`:

```ts
/**
 * Domain-free sync configuration snapshot.
 *
 * Provider IDs stay plain strings at the package boundary. Host applications can
 * narrow them in their adapter layer.
 */
export interface ConflictUiDialogRequest {
```

This comment should be deleted along with the interface it documented.

**2. Orphaned types — `packages/sync-core/src/ports.ts:92` & `:94-99`**

Removing `notify?()` leaves `ConflictUiNotification` and `ConflictUiNotificationSeverity` with zero remaining references (they were only used by `notify`). They aren't exported from `index.ts` either, so deleting them is strictly internal and safe — worth removing in the same pass.

Neither is blocking; both are trivial. Nice cleanup otherwise. 👍

forge34:
addressed in [ef7be1b](https://github.com/super-productivity/super-productivity/pull/8408/commits/ef7be1b9d4bef177f09579b7d610c7f17b671d69)

johannesjo:
Thank you very much! <3

PR #8411, opened 2026-06-15
fix(shortcuts): respect macOS keyboard layout for globals

### LINKED ISSUE #8378: 🚨 Bug: Global shortcut listener ignores system layout at input (Y/Z swapped on QWERTZ) on Mac

symonbaikov:
## Problem

Fixes #8378.

On macOS, Electron global shortcuts can ignore the active system keyboard layout. This caused layout-aware shortcuts recorded in Super Productivity to behave incorrectly on non-QWERTY layouts, for example German QWERTZ where `Y` and `Z` are swapped.

## Solution

Global shortcut registration now passes the renderer’s saved keyboard layout snapshot to Electron main.

On macOS only, global shortcut accelerators are translated from the displayed layout character to the QWERTY physical key that Electron’s `globalShortcut` listener expects. The stored shortcut value remains unchanged, so the UI still shows the user’s intended layout-aware shortcut.

The fix also re-registers global shortcuts after the keyboard layout is detected.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

## Testing

- [x] `node --test electron/global-shortcut-layout.test.cjs`
- [x] `npm run test:file src/app/core/keyboard-layout/keyboard-layout.service.spec.ts -- --watch=false`
- [x] `npm run test:file src/app/features/config/store/global-config.effects.spec.ts -- --watch=false`
- [x] `npm run electron:build`
- [x] `npm run test:electron`
- [x] Full pre-push `npm test`

### DIFF
--- a/electron/electronAPI.d.ts
+++ b/electron/electronAPI.d.ts
@@ -5,6 +5,7 @@ import {
   TaskWidgetConfig,
 } from '../src/app/features/config/global-config.model';
 import { KeyboardConfig } from '../src/app/features/config/keyboard-config.model';
+import type { KeyboardLayoutSnapshot } from '../src/app/core/keyboard-layout/keyboard-layout.service';
 import { JiraCfg } from '../src/app/features/issue/providers/jira/jira.model';
 import { AppDataCompleteLegacy } from '../src/app/imex/sync/sync.model';
 import { Task } from '../src/app/features/tasks/task.model';
@@ -227,7 +228,10 @@ export interface ElectronAPI {
 
   updateTitleBarDarkMode(isDarkMode: boolean): void;
 
-  registerGlobalShortcuts(keyboardConfig: KeyboardConfig): void;
+  registerGlobalShortcuts(
+    keyboardConfig: KeyboardConfig,
+    keyboardLayout?: KeyboardLayoutSnapshot,
+  ): void;
 
   showFullScreenBlocker(args: { msg?: string; takeABreakCfg: TakeABreakConfig }): void;
 
--- a/electron/global-shortcut-layout.test.cjs
+++ b/electron/global-shortcut-layout.test.cjs
@@ -1,0 +1,53 @@
+const test = require('node:test');
+const assert = require('node:assert/strict');
+const path = require('node:path');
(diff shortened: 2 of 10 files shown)


hajiboy95:
@symonbaikov we already got one PR which is working on this https://github.com/super-productivity/super-productivity/pull/8381

hajiboy95:
@symonbaikov can you close this? #8381 got morged and therefore we don't need this anymore i guess

PR #8412, opened 2026-06-15
Fix visuals

hajiboy95:
Fixes small design inconsistencies mentioned in https://github.com/super-productivity/super-productivity/pull/8292#issuecomment-4711064363 by @beerkumquatpome.
## Problem

This PR addresses visual issues with slide toggles in the settings page:

1. **Missing Focus Indicator:** Toggles didn't always show a focus ring when activated.
2. **Persistent Focus Rings:** Multiple fields would sometimes retain their focus state simultaneously, leading to a cluttered UI.
3. **Visibility on Colored Toggles:** The focus ring was barely visible when toggles had custom colors (e.g., blue or pink).

## Solution

The following changes were implemented:

* **Refactored `FormlySlideToggleComponent**`: Moved the hidden compatibility input outside the `mat-slide-toggle` and made it passive (`readonly`, `tabindex="-1"`). This fixes the "sticky" focus issue by ensuring the `mat-form-field` correctly tracks when focus truly leaves the component.
* **Enhanced Focus Ring**: Added a global CSS override for toggle fields in `_overwrite-material.scss`. It now uses a prominent **5px double-ring effect** (a 2px background-colored gap followed by a 3px primary-colored ring). This ensures high contrast and visibility regardless of the toggle's color or state.
* **Interactivity Improvements**: Increased the hover background prominence (15% opacity) and removed the default input-box-style border for toggles to achieve a cleaner look.

## Type of Change

* [x] Bug fix (Visual/Interaction)
* [ ] New feature
* [ ] Breaking change
* [ ] Refactoring
* [ ] Documentation
* [x] Other: UI/UX Enhancement

## Checklist

* [ ] I have included relevant changes to the documentation/wiki.

### DIFF
--- a/src/app/ui/formly-slide-toggle/formly-slide-toggle.component.ts
+++ b/src/app/ui/formly-slide-toggle/formly-slide-toggle.component.ts
@@ -20,11 +20,13 @@ import { FieldType, FormlyFieldConfig } from '@ngx-formly/core';
         (change)="onChange($event)"
       >
         {{ props.label }}
-        <input
-          matInput
-          hidden
-        />
       </mat-slide-toggle>
+      <input
+        matInput
+        tabindex="-1"
+        readonly
+        style="position: absolute; opacity: 0; width: 0; height: 0; pointer-events: none;"
+      />
       @if (props.icon) {
         <mat-icon>{{ props.icon }}</mat-icon>
       } @else if (props.svgIcon) {
--- a/src/app/ui/formly-slide-toggle/formly-slide-toggle.component.ts
+++ b/src/app/ui/formly-slide-toggle/formly-slide-toggle.component.ts
@@ -22,6 +22,7 @@ import { FieldType, FormlyFieldConfig } from '@ngx-formly/core';
         {{ props.label }}
       </mat-slide-toggle>
       <input
+        aria-hidden="true"
         matInput
         tabindex="-1"
         readonly
--- a/src/styles/components/_overwrite-material.scss
+++ b/src/styles/components/_overwrite-material.scss
@@ -237,7 +237,7 @@ body.isNativeMobile .project-complete-fullscreen-dialog {
       border-color: var(--input-border-color) !important;
       box-shadow: none !important;
     }
-    &:focus-within .mdc-text-field--filled:not(.mdc-text-field--disabled) {
+    &:has(:focus-visible) .mdc-text-field--filled:not(.mdc-text-field--disabled) {
       outline: var(--focus-ring-width) solid var(--focus-ring-color);
       outline-offset: var(--focus-ring-offset);
     }
--- a/src/styles/components/_overwrite-material.scss
+++ b/src/styles/components/_overwrite-material.scss
@@ -230,6 +230,19 @@ body.isNativeMobile .project-complete-fullscreen-dialog {
     box-shadow: 0 0 0 1px var(--input-border-color-focus) !important;
   }
 
+  // Specialized focus treatment for toggle fields:
+  // Remove the border change and enhance the focus ring (outline) instead
+  &:has(mat-slide-toggle) {
+    &.mat-focused .mdc-text-field--filled:not(.mdc-text-field--disabled) {
+      border-color: var(--input-border-color) !important;
+      box-shadow: none !important;
+    }
+    &:focus-within .mdc-text-field--filled:not(.mdc-text-field--disabled) {
+      outline: var(--focus-ring-width) solid var(--focus-ring-color);
+      outline-offset: var(--focus-ring-offset);
+    }
+  }
+
   // Error / Invalid state
   &.mat-form-field-invalid .mdc-text-field--filled:not(.mdc-text-field--disabled) {
     border-color: var(--input-border-color-invalid) !important;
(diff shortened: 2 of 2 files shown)


hajiboy95:
@beerkumquatpome 
1. that's intended. I just removed the focus ring for toggles. Doesn't really make sense
2. can you provide a screenshot?

johannesjo:
## On the focus-ring behavior (re: @beerkumquatpome's notes)

Thanks for the detailed report — I dug into this and want to propose a direction, because the three points actually pull in two different directions and one of them loops back to the original goal of #8292.

### Recommendation: make the ring keyboard-driven (`:focus-visible`), consistently

The high-contrast focus **ring** should appear for keyboard / assistive-tech focus and be suppressed on mouse/touch. That's the `:focus-visible` standard (and WCAG 2.4.7 *Focus Visible*), and it's already what this repo prescribes — `styling-guide.md` ("applies an outline on `:focus-visible` only, so it doesn't fire on mouse clicks") and the `.focus-ring` util in `util.scss`. Right now the PR uses `:focus-within`, which makes the toggle a special case versus the rest of the app.

Why not "persist the ring after a mouse toggle until I click elsewhere": a focus ring answers *"where is keyboard focus"*, not *"what did I just change"*. A toggle's own on/off animation (plus the ripple) is the change confirmation. And a ring that sticks around after every mouse click is precisely the *"multiple fields retain focus state simultaneously → cluttered UI"* problem that #8292 set out to fix — so persisting it on mouse would reintroduce that. For keyboard users the ring **does** persist naturally until you Tab away, which is the behavior you want.

### What I verified (headless Chromium, exact `button` + `label[for]` DOM)

- **`:focus-within` (current PR):** ring shows on a mouse click of *both* the toggle and its text label, and persists past 1s. So in isolation the CSS does **not** drop after ~0.5s, and label-clicks **do** light it.
- **`:has(:focus-visible)`:** ring only on keyboard focus, none on mouse — the clean, consistent behavior.

### So two of the symptoms aren't caused by the CSS

The "ring only shows ~0.5s" and "clicking the text shows no ring" did **not** reproduce from the bare selectors — which points at the live Settings page dropping focus on the toggle (most likely a re-render when `formControl.setValue()` runs, or FocusMonitor interaction), not the focus selector. Note a `:focus-visible` ring would lose focus the same way, so this is a **separate focus-retention bug** worth reproducing in the running app and fixing independently.

### Concrete next steps

1. Switch the ring trigger `:focus-within` → `:has(:focus-visible)` so toggles match the app-wide keyboard-only convention.
2. Reproduce the focus-drop in the actual Settings page and fix focus retention there (independent of the CSS trigger).
3. **Point 3 (dropdown ring bottom edge when the panel opens)** is a `mat-select` issue — this PR doesn't touch selects, so let's split it into its own issue.
4. Minor, while here: the hidden passive `<input matInput>` is now in the a11y tree as a nameless readonly textbox (it dropped `hidden`); add `aria-hidden="true"` to it.

The one real decision is keyboard-only (recommended) vs. persistent-on-mouse (the literal request). I'd go keyboard-only since it resolves the #8292 clutter and keeps every control consistent — but happy to align on whatever you two prefer before reworking it.

johannesjo:
Thanks both — and @beerkumquatpome, thanks for pushing on this, it made me dig into *why* the two screens actually differ.

**You're right that they behave differently — but it turns out it's not the CSS.** The Project/Tag settings dialog edits a *local* copy of the data, so the toggle you clicked stays mounted and keeps focus → the ring lingers. The global Settings page saves every change straight to the app's central store, which hands the form a brand-new data object and forces the field to re-render — that throws away the focused element, so the ring drops after a beat. So "make global Settings behave like the dialog" isn't a styling change; it's a focus-retention fix inside the settings form itself. It touches shared, data-synced code, so I'd rather not fold it into this PR — I've noted the root cause so we don't lose it.

**On the ring trigger, I've landed on keyboard-only (`:has(:focus-visible)`)** rather than firing on mouse clicks:
- A focus ring's job is to show *where keyboard focus is* (WCAG 2.4.7). The "what did I just change" part is already covered by the toggle itself — it stays visibly on/off, so you can always flip it back.
- It keeps toggles consistent with the rest of the app: our `.focus-ring` util / styling-guide already use `:focus-visible` *specifically* so the ring doesn't fire on mouse clicks. A toggle showing a mouse ring would be the one control behaving differently.
- It matters on touch too — `:focus-within` leaves a ring stuck on the last-tapped row.

So the lingering ring you liked in the dialog is actually the odd one out vs. the rest of the app, not the target to copy — both screens ending up keyboard-only is the consistent result.

@hajiboy95 — two small things and this is good to merge:
1. Add `aria-hidden="true"` to the `<input matInput>` you moved out of the toggle. Now that it's no longer `hidden`, screen readers see it as a nameless textbox — it only exists to satisfy `MatFormFieldControl`, so it should be invisible to assistive tech.
2. Switch the ring from `:focus-within` → `:has(:focus-visible)` per the above.

Minor / non-blocking:
- The description mentions a "5px double-ring" but the diff ships a clean single `outline` using our `--focus-ring-*` tokens — which is actually what I'd want, so just tidy the description.
- The slide-toggle ripple-color remap applies to *all* toggles app-wide, not just Settings — can you confirm that's the intended scope?

Thanks again — happy to merge as soon as the aria attr and selector are in. (@beerkumquatpome, the dropdown/`mat-select` ring edge from your point 3 is a different component this PR doesn't touch, so let's leave that out of here.)

hajiboy95:
@johannesjo thanks again for your feedback. adressed. To the minors: 
1. Year just tidy description
2. yeah is okay, to keep it across the board similar for toggles

johannesjo:
The changes are in — commit 72a59bd4 ("improve accessibility and focus-visible behavior for slide toggles"), pushed 2026-06-16 ~20:14 UTC, which is the current head of the branch. It adds `aria-hidden="true"` to the compatibility input and switches the ring trigger from `:focus-within` to `:has(:focus-visible)`. If you're not seeing it, try a hard refresh on the Files/Commits tab — GitHub sometimes caches the old diff.

On your other point: we've deliberately kept the ring **keyboard-only** rather than persisting it after a mouse toggle. The reasoning (WCAG 2.4.7 — the ring marks *where keyboard focus is*, and app-wide consistency with our `.focus-ring` convention) is in my comment above. The "ring drops after a beat / clicking the label shows no ring" behavior you reported in the global Settings page is a real but **separate** focus-retention bug (the Settings form re-renders the field on save, which drops focus) — I'm splitting that out so it doesn't block this PR. Same for the `mat-select` ring edge from your point 3. Tracking issue: #8447.

johannesjo (approved):

johannesjo:
Thank you very much @hajiboy95 ! <3

PR #8413, opened 2026-06-15
feat(boards): add date timeframe filters

### LINKED ISSUE #8392: 💡 Feature Request: Advanced Date & Timeframe Filtering for Custom Boards (Planning & Deadlines)

symonbaikov:
## Problem

Closes #8392.

Custom board panels could filter tasks by planning status, but could not narrow scheduled tasks by a specific timeframe such as Today, Tomorrow, Next 7 days, Next week, Next month, or a custom date/range. Boards also had no way to filter by task deadline presence or deadline timeframe.

Manual UI smoke test was not performed yet.

## Solution

Adds timeframe-aware filters for custom board panels:

- Extends scheduled/planning filters with optional timeframe constraints.
- Adds a new deadline filter with:
  - All
  - Has deadline
  - No deadline
- Adds deadline timeframe controls when `Has deadline` is selected.
- Supports:
  - Today
  - Tomorrow
  - Next 7 days
  - Next week
  - Next month
  - Custom date
  - Custom range
- Adds a focused board date-filter utility for resolving ranges and matching date-only/timestamp fields.
- Sanitizes invalid or irrelevant timeframe config.
- Updates board docs and translations.

Deadline filtering is display-only. Drag/drop into a deadline-filtered panel does not set or remove task deadlines.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [ ] Other

## Testing

- [x] `npm run test:file src/app/features/boards/board-date-filter.util.spec.ts -- --watch=false`
- [x] `npm run test:file src/app/features/boards/boards.util.spec.ts -- --watch=false`
- [x] `npm run test:file src/app/features/boards/board-panel/board-panel.component.spec.ts -- --watch=false`
- [x] `npm run test:file src/app/features/boards/store/boards.reducer.spec.ts -- --watch=false`
- [x] `npm run test:file src/app/features/boards/boards-form.const.spec.ts -- --watch=false`
- [x] `npm run checkFile` for modified TypeScript files
- [x] Pre-push hook: full Angular suite, `11084 SUCCESS`
- [x] Pre-push hook: LA timezone Angular suite, `11070 SUCCESS`

## Notes

The pre-push hook reported existing non-blocking warnings:
- Node.js `v25.9.0` odd-version warning
- Unsupported Browserslist entry: `chrome 107`
- Existing SCSS warning in `habit-tracker.component.scss`


hajiboy95:
I actually have a very similar implementation open in PR #8393. Feature-wise, they look almost identical, but mine also includes handling for when an element moves from one board to another, as well as more filters, for ex. 'next X days' and 'more than X days in the future.'

I'm completely open to going with your approach if there are architectural advantages I'm missing! But right now, it looks like mine covers a few more use cases out of the box. Let me know what you think.

johannesjo:
Thanks a lot to both of you — @symonbaikov (#8413) and @hajiboy95 (#8393) — you landed near-identical solutions for this within a day of each other, and both are genuinely solid. I'd like to converge on one so the effort isn't wasted.

They differ in two meaningful ways:

1. **Scope / risk.** #8413 is self-contained in `boards/`: display-only filtering, a clean nested config model, strong sanitization, and good unit tests for the date math — no shared-component changes. #8393 is broader: it also touches the shared schedule dialog + datetime-picker, adds a button-style date picker and more presets ("next X days", "≥ X days out", half-bounded ranges), and — the big one — **auto-replans a task to the nearest valid date when you drop it into a timeframe panel** instead of letting it vanish.

2. **A product decision underneath it.** The auto-replan UX is nicer, but it means drag-and-drop now *mutates* a task's due date / deadline. #8413 keeps drops as pure filtering. We should decide which behavior we want before picking a base.

My leaning: use #8413 as the foundation (smaller, lower-risk, clean model + sanitization) and layer #8393's extra presets — and, if we want it, the drop-replan logic — on top. But I'm open if you feel strongly otherwise.

Could you two sync on who carries the combined PR? Happy to review once you've aligned. Thanks again, and sorry you doubled up here.

symonbaikov:
Updated this PR with the combined approach discussed here.

What changed on top of the original #8413 foundation:
- Kept the nested board config model and board-local sanitizer/date-filter utility from this PR.
- Added the richer timeframe select options inspired by #8393: All, Next X days, At least X days in the future, and half-bounded custom ranges.
- Added a reusable button-style Formly date picker adapted from #8393's UX, with minimal shared scheduler dialog support for date-only picking.
- Added #8393-style drag/drop auto-replan behavior for scheduled/deadline timeframe panels: drops now move the task to the nearest valid date instead of letting it disappear from the target panel.
- Added tests for the new date math, sanitization, form controls, drag/drop replan behavior, date button, scheduler dialog options, and skipped duplicate deadline toast.

Credit to @hajiboy95 for the select-option/date-button UX and drop-replan direction from #8393. I kept this PR's lower-risk nested model as the base, matching @johannesjo's suggestion.

hajiboy95:
@symonbaikov when I tried to start this version on my computer i got an error 

```
error TS2322: Type '{ customEnd?: unknown; customStart?: unknown; type: "customRange"; }' is not assignable to type 'BoardDateTimeframeCfg'.
  Types of property 'customStart' are incompatible.
    Type 'unknown' is not assignable to type 'string'.

227       return {
          ~~~~~~


Found 1 error in src/app/features/boards/board-date-filter.util.ts:227
```


Adding type specifications solved the problem on my end: 

```
        ...(cfg.customStart ? { customStart: cfg.customStart as string } : {}),
        ...(cfg.customEnd ? { customEnd: cfg.customEnd as string } : {}),
```


instead of current

```
      return {
        type,
        ...(cfg.customStart ? { customStart: cfg.customStart } : {}),
        ...(cfg.customEnd ? { customEnd: cfg.customEnd } : {}),
      };
```


@johannesjo besides that I guess it's good to go? What do you think?

johannesjo:
Thanks @symonbaikov — this is a solid, well-tested PR and the i18n/styling discipline is genuinely good.

On the TS error: that one looks already resolved on your latest commit (`aedc388`). The `customRange` branch of `sanitizeBoardDateTimeframeCfg` now narrows through the `isValidDbDateStr` type guard and assigns into a typed `out: BoardDateTimeframeCfg`, so it compiles without a cast — and that's actually nicer than `as string` because it *validates* the value instead of just silencing the compiler. So no change needed there; a fresh pull + rebuild should be green.

A few things I'd like to tighten before it goes in. None are huge, but the first two touch synced data so I care about them:

### Worth fixing before merge

**1. `days` has no upper bound → can write `"NaN-NaN-NaN"` into synced state.**
The `days` field has `min: 1` but no `max` (`boards-form.const.ts`), and `isValidPositiveInteger` doesn't cap it. An extreme value (paste, or a malformed config arriving from another client) flows through `addDays` → `Invalid Date` → `getDbDateStr` returns the literal string `"NaN-NaN-NaN"`. For `atLeastNDaysFuture`, `adjustDateToBoardTimeframe` then evaluates `sourceDate < "NaN-NaN-NaN"` as `true` and we dispatch `PlannerActions.planTaskForDay({ day: "NaN-NaN-NaN" })`, which `addTaskToPlannerDay` writes straight into the persisted/synced planner as a day-bucket key (the deadline path is guarded by `isDBDateStr`; the planner path isn't). Cheap fix: add a sane `max` on the field + cap in the sanitizer, and/or validate `closestDate` with `isValidDbDateStr` before dispatching in `_checkToScheduledTask`.

**2. The PR description says deadline filtering is "display-only", but the code sets/removes deadlines on drop.**
`_checkDeadlineState()` dispatches `setDeadline` / `removeDeadline` (and opens the deadline dialog) on drop. To be clear this isn't silent — `removeDeadline` shows the standard "Deadline removed" toast and the set-path shows an "Adjusted deadline…" snack — and it mirrors the existing scheduled-panel drop behavior (`_checkToScheduledTask`), so the *behavior* is consistent. It's just that the description says the opposite, so please reconcile it. One thing to confirm while you're there: the set-path passes `getDeadlineAutoPlanFields(...)`, so dropping onto a "Has deadline: Today" panel also auto-plans the task into Today (the snack only mentions the deadline). That's consistent with how a today-deadline behaves elsewhere, so it's fine if intended — just confirm it's deliberate.

**3. "Next week" ignores the user's first-day-of-week.**
`getNextIsoWeekRange` hardcodes a Monday start (`(today.getDay() + 6) % 7`). The rest of the app derives week bounds from the user's `firstDayOfWeek` setting via `getWeekRange()` (the schedule dialog this PR touches does too), so a Sunday-start user's board "Next week" disagrees with "next week" everywhere else. Thread `firstDayOfWeek` in as a pure arg and reuse `getWeekRange` / `getDateRangeForMonth` rather than hand-rolling the range.

**4. Custom date / range can be saved empty.**
`customDate` / `customStart` / `customEnd` aren't `required` (unlike `days`), so picking "Custom date/range" and saving without a value succeeds, and the sanitizer then drops the timeframe so the panel silently behaves as "All". Marking them required closes that.

### Note for the future (not blocking)

**`BoardDateTimeframeType` is a closed, typia-validated synced union.** The new optional fields are safe for old clients today, but `BoardsState` is typia-validated and not in `_relaxForwardCompatibleErrors`, so **adding a new timeframe value in a future release will make today's clients reject that synced config as corruption** (the "repair your data?" dialog → can wedge sync). Worth a comment on the union noting that adding values is a breaking change for older clients, and keeping any future "kind" as an opaque string outside the validated union.

### Polish

- **Perf:** `matchesBoardDateTimeframe` re-resolves the range once *per task* even though it's invariant for the whole filter pass (~6 `Date` allocations/task × panels × tasks). Resolve once and pass it into a lightweight matcher — same as how `buildComparator` is already hoisted out of the loop just above.
- **Discoverability:** there's no panel-header hint of which timeframe a column filters by, and the labels are a bit ambiguous ("Next 7 days" rolling/incl. today vs "Next week" calendar/excl. today; "Next month" = calendar month, not 30 days). A tooltip or a resolved-range preview would help.
- **a11y:** the clear icon-button on `formly-date-btn` relies on `matTooltip` for its name — add an explicit `aria-label` (a tooltip isn't an accessible name); and the long localized date can overflow on narrow/mobile (add `min-width: 0` + ellipsis).
- **Dead class** `formly-btn-form-field` on the date button — it's not defined anywhere in the repo.

### One bigger-picture thought
`formly-date-btn` (+ the new `showTime`/`showReminder`/`maxDate` plumbing on the schedule dialog) largely duplicates the existing `formly-date-picker` (`type: 'date'`), which already round-trips `YYYY-MM-DD` and supports min/max — and currently has zero usages. The only genuinely new bit is the cross-field `minDateKey`/`maxDateKey`, which formly `expressions` can do on the existing type. Not blocking, but reusing it would cut a few hundred lines and avoid a second date-input UX in the codebase — worth a look.

Overall really close — mainly #1 (the `days` cap) and #2 (description/intent) from me, plus #3/#4 when you get a chance. Nice work.

PR #8423, opened 2026-06-15
Codex/issue 8081 keep subtask input open

### LINKED ISSUE #8081: 💡 Keep subtask creation input open after adding a subtask (Consistency with main tasks)

symonbaikov:
## Problem

Fixes #8081.

Adding multiple subtasks in a row was slower than the main task creation flow. The previous subtask flow could create an empty persisted subtask draft, which was undesirable for sync/op-log correctness because cancelled drafts should not create task operations.

https://github.com/user-attachments/assets/1198b21e-f022-4276-8395-ee9698d26e76

## Solution

- Add a local-only inline subtask draft input that is not persisted until committed.
- Plain `Enter` with a non-empty draft creates a subtask via `addSubTaskTo(parentId, { title })`.
- After submit, the input is cleared and remains focused for rapid subtask entry.
- `Escape` clears/closes the draft without creating a subtask.
- Blur clears/closes the draft without creating a subtask.
- Route add-subtask actions from task rows, context menu, shortcuts, and the detail panel to the inline draft input.
- Keep normal task title editing unchanged: plain `Enter` on existing subtasks or top-level tasks does not create another subtask.
- Prevent subtask detail-panel toggle clicks from bubbling to the parent task row.

## Test Plan

- [x] `npm run checkFile` for all changed TS/SCSS/HTML files
- [x] `npm run test:file src/app/features/tasks/add-subtask-input/add-subtask-input.component.spec.ts`
- [x] `npm run test:file src/app/features/tasks/task/task-shortcuts.spec.ts`
- [x] `npm run test:file src/app/features/tasks/task-shortcut.service.spec.ts`
- [x] `npm run test:file src/app/features/tasks/task-detail-panel/task-detail-panel.component.spec.ts`
- [x] `npm run test:file src/app/features/tasks/task-context-menu/task-context-menu-inner/task-context-menu-inner.component.spec.ts`
- [x] `npm run test:file src/app/features/tasks/task.service.spec.ts`
- [x] `npm run test:file src/app/features/tasks/task/task-on-focus-panel-sync.spec.ts`
- [x] `npm run e2e:file e2e/tests/task-basic/subtask-inline-input.spec.ts -- --retries=0`
- [x] Pre-push full lint and test suite


johannesjo:
Thanks for this — the core approach is solid, and I confirmed the main goal is met: the draft is local-only and creates **no task op until commit** (`addSubTaskTo(parentId, {title})` is the only store write, gated on a non-empty trimmed title; Escape/blur/empty-Enter discard with no dispatch). That fixes the op-log concern from #8081 cleanly. 👍

A few things to address before merge, including the lost animation:

## 🔴 The subtask input re-opens by itself on later task updates

`AddSubtaskInputService.openRequest` is **never reset** after it's handled, and the effect in `task.component.ts` depends on **both** `openRequest()` and `this.task()`:

```ts
// task.component.ts:339
private readonly _addSubtaskInputRequestEffect = effect(() => {
  const request = this._addSubtaskInputService.openRequest();
  const task = this.task();              // ← effect re-runs on every task() change
  if (\!request || request.parentId \!== task.id) return;
  window.setTimeout(() => {
    ...
    this.isAddSubtaskInputVisible.set(true);   // ← re-opens even after user closed it
    window.setTimeout(() => this.addSubtaskInput()?.focus());
  });
});
```

Since `requestOpen()` only ever sets the signal (`add-subtask-input.service.ts:17`) and the `(closed)` handler just flips the local `isAddSubtaskInputVisible`, the request stays live forever. So after a user opens the input on task A and closes it, the **next** change to A's object re-runs the effect, the `parentId === task.id` guard still passes, and the input pops back open and steals focus.

This is easy to hit:
- Add a subtask to a task, close the input, then toggle that subtask done (or any edit to the parent) → input reopens.
- It's continuous if A is the **currently time-tracked task** (the 1s tick re-emits `task()` every second).
- A freshly mounted `TaskComponent` whose id matches a stale request **auto-opens** the input with no user action (e.g. after switching views).

`requestId` is computed but never read, so the intended one-shot semantics aren't implemented. Suggested fix — consume the request:

```ts
const request = this._addSubtaskInputService.openRequest();
if (\!request || request.requestId === this._lastHandledRequestId) return;
const task = untracked(() => this.task());      // don't react to task() changes
if (request.parentId \!== task.id) return;
this._lastHandledRequestId = request.requestId;  // requestId now earns its keep
...
```

(or simply `openRequest.set(null)` once consumed). A unit test covering "close → update parent → stays closed" would lock this in — the current spec mocks `openRequest` as a static `signal(null)`, so this path is untested.

## 🟡 First-subtask animation is gone (the reported regression)

`[@expandInOnly]` lives only on `<task-list>`. Previously, triggering "add subtask" on a task with no subtasks created the task immediately, so the sub-tasks section expanded in via that animation. Now the bare `<add-subtask-input>` (and its `.sub-tasks` wrapper) render with **no enter animation** (`task.component.html:309`, and the component/SCSS define none), so the input box pops in abruptly.

Adding an enter animation to the input restores it, e.g.:

```html
<add-subtask-input
  [@expandFade]
  [parentId]="t.id"
  (closed)="isAddSubtaskInputVisible.set(false)"
></add-subtask-input>
```

(reusing an existing trigger from `ui/animations/expand.ani.ts`).

## 🟡 Document keydown listener can leak

`AddSubtaskInputComponent` adds a capture-phase `document` keydown listener in `onFocus()` and removes it only on blur/`_close()`, with no `OnDestroy`/`DestroyRef` backstop. If the component is destroyed while focused without a blur firing (task deleted/moved, route switch), the listener leaks. A one-liner fixes it:

```ts
private readonly _destroyRef = inject(DestroyRef);
constructor() {
  this._destroyRef.onDestroy(() => this._removeDocumentKeydownListener?.());
}
```

## Minor suggestions (non-blocking)

- The `_isKeepingOpenAfterSubmit` / `_isClosedWithoutSubmit` flags + the `setTimeout(0/100/150)` focus dance and the document-level Escape listener are heavier than needed — `onBlur` never commits (it only discards), so the "capture Escape before blur commits" race the document listener guards against doesn't exist here; the input's own `(keydown)` Escape handler should suffice. Consider `afterNextRender(...)` for focus-after-render instead of the nested timers (already used elsewhere in the repo).
- `addSubTaskTo` now skips `focusTaskById` for titled subtasks (`task.service.ts:779`). This also affects the issue-import and local-REST callers (they pass a title) — almost certainly an improvement (no focus-steal on programmatic creation), just flagging it as an intentional behavior change.
- `placeholder` via `instant()` won't react to a live language switch; the `| translate` pipe is the more common pattern here.

johannesjo:
**Performance review of the per-row `effect()` in `task.component.ts`**

I looked specifically at the new reactive wiring in `TaskComponent` (the documented hot-path component) and measured it rather than eyeballing. Summary: **it works and the cost is negligible, but there's a one-line tweak worth making, plus a bit of dead code to clean up.**

### What I measured

The new `_addSubtaskInputRequestEffect` runs in **every** rendered task row and reads two signals — the global `AddSubtaskInputService.openRequest()` and the per-row `this.task()`. I reproduced the exact pattern in an instrumented test:

- **One add-subtask request re-runs the effect in every row** (O(N) fan-out). 200 rows → 200 effect executions per single click. Inherent to broadcasting one global signal to N watchers.
- **The effect also re-runs on every `task()` change** even with no request pending — i.e. on every time-tracking tick of the active task, every update, etc. 5 task updates → 5 needless re-runs.
- **Magnitude:** the full O(N) flush is **~0.5 ms for 500 rows**. So in wall-clock terms this is *not* a real perf problem — it's a discrete user click, not a per-frame cost.

### Suggested change (1 line)

The effect only ever needs the task **id** for comparison, so it doesn't need to react to `task()` at all. Reading it untracked removes the per-tick churn at zero behavioral cost:

```ts
private readonly _addSubtaskInputRequestEffect = effect(() => {
  const request = this._addSubtaskInputService.openRequest();
  // only react to the open-request signal; read the task untracked so this
  // per-row effect doesn't re-run on every task() update (e.g. the tracking tick)
  const task = untracked(() => this.task());
  if (\!request || request.parentId \!== task.id) return;
  ...
});
```

I verified this keeps the inline flow intact: component spec 9/9, `task-shortcuts` 18/18, e2e `subtask-inline-input` 3/3 green.

> Note: I considered switching to a `computed` for visibility, but it would **not** remove the O(N) fan-out (a computed read in each row's template still marks all N rows for check when the global signal flips). Given the measured ~0.5 ms, the only worthwhile change is the `untracked` one-liner. True O(1) would mean not having every row watch the global signal at all (a single list-owned input) — not justified by this magnitude.

### Minor, non-perf

- **Dead code:** `_getPreviousTaskEl()` and `_focusTaskHost()` are no longer called from production after this PR removed the Escape branch in the title-submit handler — only a spy in `task-shortcuts.spec.ts` references them now. Safe to delete (and drop the spies).
- **Possibly two inline inputs open at once:** `isAddSubtaskInputVisible` is a local per-row boolean reset only by `(closed)`, so opening row B doesn't close row A's input. Blur *may* mask this in practice (A blur-closes when focus leaves) — I didn't confirm; worth a quick check.

Nice approach overall — dropping the empty persisted-draft for op-log correctness is the right call.

johannesjo:
## Review feedback

Nice work — the core approach is solid. Switching to a local-only inline draft that only persists on a non-empty `Enter` cleanly achieves the op-log goal (one `addSubTask` op per intent, no phantom empty-draft subtasks), there's no XSS regression (Angular property binding + the existing escaped render path), logging stays IDs-only, and the `untracked(() => this.task())` perf commit correctly keeps the new effect off the per-task hot path. Test coverage is good too.

A few things worth addressing before merge:

### 🟠 Should fix: the document-level `keydown` listener is both redundant and leaks

`add-subtask-input.component.ts`

- **Leak:** the capture-phase `document.addEventListener('keydown', …, true)` is added in `onFocus()` and removed only in `onBlur()`/`_close()`. The component has no `ngOnDestroy`/`DestroyRef`. If the host task row is destroyed while the input is focused (parent deleted by a remote sync op, moved between lists/projects, or filtered out of the list), the browser doesn't reliably fire `blur` on DOM removal, so the listener stays attached to `document` closing over a dead component. `focus-mode-overlay.component.ts` cleans up the identical pattern in `ngOnDestroy`.
- **Redundant:** `onKeydown()` already handles `Escape` and calls `stopPropagation()` for every key, there's no global capture-phase Escape handler anywhere in `src/app`, and `onBlur()` only ever calls `_close()` — it never commits. So the "capture Escape before blur can commit the draft" race that this listener (and the spec test at line ~111) is built around can't actually occur.

**Suggested fix:** delete the whole document-listener apparatus (`onFocus`, the first branch of `onBlur`, `_onDocumentKeydown`, `_removeDocumentKeydownListener`). `onKeydown`'s Escape branch fully covers the behavior — this removes the leak and ~25 lines in one go. Update the now-misleading spec accordingly.

### 🟡 Quick cleanups

- **Dead code in the hot-path component:** `_getPreviousTaskEl()` and `_focusTaskHost()` in `task.component.ts` are now orphaned (their only caller — the removed Escape branch — is gone). Safe to delete, along with the two `spyOn` stubs in `task-shortcuts.spec.ts`.
- **Speculative plumbing in `task-shortcut.service.ts`:** the new `typeof result === 'boolean' ? result : true` is never exercised — every `_handleTaskShortcut` call site discards the return value and no wired method returns a boolean (the new "should not consume Escape" test passes because Escape falls through to `return false`, not because of this change). Suggest reverting to `return true` unless you intend to wire the boolean through to callers.

### 🟢 Optional polish

- **Focus-timing dance:** the stacked `setTimeout`s (0/100/150ms) + `_isKeepingOpenAfterSubmit` flag in `_commit()` are probably unnecessary now — since `addSubTaskTo` no longer steals focus when a title is present and the `<input>` stays mounted on commit (`titleDraft.set('')` only updates `[value]`), focus isn't lost. A single `afterNextRender` focus should suffice. (Verify the first-open path where `showSubTasks()` re-renders the subtree — keep one deferred focus there.)
- **Global-signal + per-row `effect()`:** `AddSubtaskInputService.openRequest` makes every task row's effect re-run on every add-subtask request app-wide. It's cheap (early-return + rare event + `untracked`), so acceptable, but a targeted call via `TaskFocusService.lastFocusedTaskComponent()` (or a `Subject<parentId>`) would avoid the fan-out and drop the synthetic `requestId` counter, which only exists to force re-emit on repeat opens.
- `[(ngModel)]="titleDraft"` (signal-bound, like `add-task-bar`) would replace the manual `[value]` + `onInput` + cast.
- The placeholder via `_translateService.instant()` resolves once at construction and won't react to a runtime language switch — the `| translate` pipe (used elsewhere in this feature) is reactive.

### Not issues
- Removing the Escape "delete empty subtask" branch is **not** a regression — no UI path creates empty drafts anymore, so it's correctly obsolete.
- The `addSubTaskTo` focus-only-when-empty change is safe for the non-UI callers (issue service / REST handler both pass a title and don't want focus).

---
<sub>Reviewed with a multi-agent pass (correctness / security / architecture / performance / simplicity); findings above were individually verified against the code.</sub>

johannesjo:
Thank you very much for this <3!

PR #8427, opened 2026-06-16
perf(sync): cache latest full-state op lookup

### LINKED ISSUE #8362: Op-log store: getLatestFullStateOpEntry full-log decode scan + per-call getLastSeq transactions on the sync path
**Area:** client (op-log store) · **Type:** performance · **Value:** medium on large logs

## Problem
1. **`getLatestFullStateOpEntry` (`operation-log-store.service.ts:656-683`) cursor-walks the entire ops store, decoding every entry** — and deliberately can't early-exit ("We never stop early: UUIDv7 order can differ from seq order…", `:675-677`). It runs per incoming-ops batch: sole production caller is `SyncImportFilterService.filterOpsInvalidatedBySyncImport` (`sync-import-filter.service.ts:83`) ← `processRemoteOps` (`remote-ops-processing.service.ts:238`) ← download/piggyback/realtime paths (`operation-log-sync.service.ts:296, :762, :1047`); `:217` early-returns when zero ops, so it's per-batch, not per-idle-sync. `clearFullStateOpsExcept` (`:738-762`) does a **second identical full decode-scan** on the same remote-import path. (`getLatestFullStateOp` at `:634` has no production callers — fold into the dead-code issue.)
2. **`getLastSeq()` opens a fresh reverse-cursor transaction per call** (`:995-1009`, O(1) read — the cost is transaction setup) and is called at the top of every `getUnsynced()` (`:767`) and `getAppliedOpIds()` (`:824`) even on cache hits. `getUnsynced` is hit 4–6× per sync cycle (`operation-log-sync.service.ts:353/473/824`, `operation-log-upload.service.ts:99`, `sync-import-conflict-gate.service.ts:81`, `remote-ops-processing.service.ts:545/598`).

## Proposed fix (multi-tab constraints verified)
- For (1): a **persisted** singleton "latest full-state op" pointer, updated transactionally at the write sites (`append*` of full-state ops, `clearFullStateOps*`, `runDestructiveStateReplacement` `:1626`, `deleteOpsWhere` `:972`) → O(1). In-memory-only is NOT multi-tab safe. Note the SQLite adapter's extracted columns are `op_id/synced_at/source/application_status` (`sqlite-op-log-adapter.ts:106-116`) — there is no opType column yet; the pattern exists, the column would need adding.
- For (2): an in-memory `lastSeq` cache is only safe with explicit invalidation inside the sync Web Lock — the DB read is what keeps caches cross-tab fresh (cf. the `clearVectorClockCache` "MULTI-TAB SAFETY" pattern, `operation-log.effects.ts:173`, store `:1361-1380`). Follow that pattern or leave as-is.

symonbaikov:
## Problem

Fixes #8362.

`getLatestFullStateOpEntry()` was cursor-scanning the full op-log and decoding every operation to find the latest `SYNC_IMPORT` / `BACKUP_IMPORT` / `REPAIR`. That runs on the sync path and becomes expensive for large local logs. `clearFullStateOpsExcept()` had the same full-scan pattern when pruning stale full-state ops.

## Solution

Add a small `SUP_OPS.meta` store entry for full-state operation metadata.

The metadata stores refs `{ opId, seq }` plus the latest full-state op pointer. New full-state writes update this metadata transactionally, and existing databases rebuild it once on first lookup. Reads now fetch the latest full-state row directly by `seq`, while still returning metadata from the real op-log entry (`source`, `syncedAt`, decoded op, etc.).

Also updates cleanup paths so stale full-state ops are deleted through metadata refs instead of scanning the whole `ops` store. Destructive replacement updates the pointer in the same transaction as `OPS`, `STATE_CACHE`, `VECTOR_CLOCK`, and `CLIENT_ID`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

## Test Plan

- [x] `npm run test:file src/app/op-log/persistence/operation-log-store.service.spec.ts`
- [x] `npm run test:file src/app/op-log/persistence/db-upgrade.spec.ts`
- [x] `npm run test:file src/app/op-log/persistence/op-log-db-schema.spec.ts`
- [x] `npm run test:file src/app/op-log/persistence/sqlite-op-log-adapter.spec.ts`
- [x] `npm run checkFile` for all modified `.ts` files
- [x] pre-push `npm test`


johannesjo:
**Review: caching the latest full-state op pointer**

Nice perf direction — replacing the per-sync full scan of `ops` with an O(1) pointer is the right instinct. I did a deep multi-pass review focused on sync-correctness, and found one issue I'd consider merge-blocking plus two worth hardening before this ships, because the blast radius is silent cross-device data divergence.

## 🔴 Blocking — `meta` can be seeded incomplete on databases upgraded from ≤ v6

The v7 upgrade creates the `meta` store **empty** and relies on a lazy rebuild on first *read*. But the *write* path doesn't rebuild:

- `_recordFullStateOpInTx` (`operation-log-store.service.ts:421`) reads `tx.get(META)` → `undefined` on an upgraded DB and seeds meta with **only the op being written** (`_withFullStateRef(undefined, ref)`), silently dropping every pre-existing full-state op already in `ops`.
- `_getFullStateOpsMetaOrRebuild` (`:464`) rebuilds *only when meta is absent* — never when it's present-but-incomplete. The two-level fallback in `getLatestFullStateOpEntry` (`:800`) self-heals a missing/mismatched pointer, but not an internally-consistent-but-incomplete one.

So if a full-state op is appended **before** the first read on an upgraded DB, `meta` is permanently incomplete. This is reachable in production: the file-based snapshot path appends `result.newOps` as `'remote'` without a preceding `getLatestFullStateOpEntry`/`clearFullStateOps` read and without clearing `ops`:

```
operation-log-sync.service.ts:588  await hydrateFromRemoteSync(snapshot, vc, false)  // no OPS clear, no filter read
operation-log-sync.service.ts:606  await opLogStore.appendBatchSkipDuplicates(result.newOps, 'remote')
```

An established Dropbox/WebDAV client with op-log history and no pending meaningful data falls through to this branch (`:543`), so on its first sync after the v7 upgrade a full-state op in `newOps` seeds meta with only that op. (The normal incremental download path is safe — it reads the filter at `remote-ops-processing.service.ts:238` before appending.)

Consequences:
1. `clearFullStateOps()` (USE_REMOTE, `:945`) can no longer prune the untracked pre-existing imports — they linger in `ops` and keep filtering remote ops the user chose to discard.
2. If the appended remote op has a **lower UUIDv7** than an untracked pre-existing import (remote ops legitimately carry older timestamps), `getLatestFullStateOpEntry()` returns the wrong import → `SyncImportFilterService` filters against the wrong baseline.

**Suggested fix:** populate `meta` during the v7 upgrade by scanning `ops` inside the upgrade transaction — consistent with how the v4/v6 stores were migrated. That makes "meta reflects ops" an unconditional invariant and lets you delete `_getFullStateOpsMetaOrRebuild`, the `?? rebuild` fallback, and the second rebuild branch in the read path. (SQLite starts empty today, so IndexedDB is the only backend with pre-existing data to migrate — worth a note for the future data-copy phase.)

A unit test that reproduces it: pre-existing full-state op X (higher UUIDv7) in `ops`, `meta` empty, then `append(remoteFullStateOp Y)` with `Y.id < X.id` → `getLatestFullStateOpEntry()` returns Y, where the pre-PR scan returned X.

## 🟠 Malformed `meta` row would wedge all sync

`_getFullStateOpsMetaOrRebuild` (`:464`) trusts the stored value and only rebuilds on `null`/`undefined`. A structurally-invalid row (`refs` not an array) flows into `meta.refs.filter(...)`/`refs.reduce(...)` and throws. Because `getLatestFullStateOpEntry` runs on every download, a single corrupt row throws on every sync with no recovery — a chokepoint the old per-row scan didn't have. Guarding with `if (!Array.isArray(meta?.refs)) → rebuild` is cheap and also hardens the issue above.

## 🟠 `deleteOpsWhere` updates `meta` in a separate transaction from the OPS delete

`:1090–1123` deletes ops in the iterate transaction, then does a separate `put(META)`. The crash window self-heals on read, but a concurrent full-state append from another tab between the two can be clobbered (lost update), and since the lost ref's op still exists, read-side validation passes and won't rebuild → wrong latest. Every other write path keeps `ops`+`meta` in one transaction; folding these into a single `transaction([OPS, META], 'readwrite', …)` would match.

## 💭 Design / scope question

`getLatestFullStateOpEntry` runs once per sync batch over a store bounded to ~500 ops by compaction — was this scan actually measured as a bottleneck? If not, a lighter approach might carry less permanent surface: an in-memory cache invalidated on write (no schema change / no irreversible migration), following the existing `_appliedOpIdsCache`/`_unsyncedCache` pattern. Separately, `meta.refs[]` is only consumed by the non-empty `clearFullStateOpsExcept(excludeIds)` path, which has no production caller — storing just `latest` would remove the per-write array maintenance. Not blocking, just worth weighing against an irreversible v7 bump.

## ✅ Things I confirmed are correct
- Compact `o` stores the exact `opType` string, so `getStoredOpType`/`isFullStateOpType` work for both formats.
- Non-full-state appends keep the fast single-store path — no hot-path regression; the read-path win is real.
- `meta.latest` (max UUIDv7) matches the pre-PR scan semantics.
- Vector-clock causality still governs keep/drop, so the cached pointer doesn't open a new "crafted-latest discards valid ops" surface; no new logging of user content.
- SQLite creates the table from schema and starts empty, so no upgraded-data hazard there today.

Happy to pair on the upgrade-population approach if useful.

johannesjo:
@symonbaikov can you have a look at the conflicts?

symonbaikov:
> @symonbaikov can you have a look at the conflicts?

Yes. Will resolve them.

johannesjo:
Thank you very much! <3

Issue #8400, opened 2026-06-15
🚨 Order of tags doesn't match across application


Thalita-G:
### Version Used

18.9.1FP and 18.10.0AI 

### Current Behavior

I've created the following tags, using numbers to force the desired tag order (since SP doesn't utilise your custom order from the sidebar anywhere else).

<img width="240" height="325" alt="Image" src="https://github.com/user-attachments/assets/94f2b4b8-5a1e-41ea-b463-ee1a9f119062" />

When I use these tags inside a project, the toggle tag from the context menu shows the following order, which is as expected:

<img width="452" height="332" alt="Image" src="https://github.com/user-attachments/assets/a074d841-00b2-4209-96ce-f04473acb86b" />

However, when I group the tasks by tags, it's sorted like this instead:

<img width="357" height="256" alt="Image" src="https://github.com/user-attachments/assets/ed37d550-394e-4632-ae03-b6e731a0023d" />

Even worse when I sort the tasks by tags, the sorting seems all over the place:

<img width="808" height="378" alt="Image" src="https://github.com/user-attachments/assets/ac958cc5-d52e-4321-a7ff-3cb5f5e46540" />

It's been originally tested on the Flatpak version, but since it's currently on an older version  I decided to try the AppImage as well, and I could immediately reproduce the bug there as well.

### Expected Behavior

The same order inside the context menu should be applied to tag groups and sorts as well.

Ideally the tags order would just match the sidebar so I wouldn't have to enforce the order with numbers to begin with, but that's more of an enhancement and not a requirement to me. But since [there's currently a PR](https://github.com/super-productivity/super-productivity/pull/8100) that changes the order inside the context menu to match the sidebar, I figured I'd mention it anyway.

### Steps to Reproduce

1. Create tags as shown in first screenshot. No specific colours or icons were selected.
2. Create a project (I named mine PERSONAL).
3. Tasks were created in this exact order from top to bottom:

<img width="819" height="380" alt="Image" src="https://github.com/user-attachments/assets/d5177091-39ce-445e-ba09-a73002626265" />

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

[...]

PR #8437, opened 2026-06-16
fix(tags): apply sidebar order to tag pickers

### LINKED ISSUE #8419: tags: apply sidebar order to remaining tag pickers (and avoid per-CD rebuild in grouped work-view)
_Follow-up to #8400 (use sidebar order for tag grouping/sorting/menus)._

### 1. Remaining pickers still use alphabetical order

#8400 unified the two tag-toggle menus + the work-view customizer onto **sidebar order**, but several tag pickers still use the alphabetical `…Sorted` variant, so ordering is now inconsistent across the app:

- `src/app/features/tasks/add-task-bar/add-task-bar-actions.component.ts:~104`
- `src/app/features/tasks/add-task-bar/add-task-bar.component.ts:~251`
- `src/app/features/mention/mention-config.service.ts:~40`
- `tag-edit.component.ts:~92`

Consider switching these to sidebar order too for full consistency. (`tagsNoMyDayAndNoListSorted` is still in use, so it's **not** dead code.)

### 2. Minor perf note in grouped work-view

From the same review: `getOrderedGroupKeys(...)` is called from the `@for` in `src/app/features/work-view/work-view.component.html:~136`, rebuilding Maps + a sorted array on **every change-detection pass** while grouping is active (this is the parent list, not the per-task hot path).

There's also a one-frame staleness window: it reads `this.selectedGroup()` live while `grouped` was built from the stream-captured group (`task-view-customizer.service.ts:~445` vs the `combineLatest` map deferred onto `animationFrameScheduler`). It self-corrects on the next CD and `track groupKey` prevents DOM churn, but passing the group type alongside `grouped` would be cleaner.

https://claude.ai/code/session_01As7c32Eacxqksg5yHt5utw

symonbaikov:
## Problem

Fixes #8419.

After #8400, most tag menus followed the sidebar tag order, but a few remaining tag pickers still used the alphabetical `tagsNoMyDayAndNoListSorted*` source. This made tag ordering inconsistent across the app.

Grouped work-view also rebuilt grouped key/value entries in the template via `keyvalue`, which added avoidable work during change detection.

## Solution

- Switched the remaining tag picker/suggestion consumers to sidebar-order tag sources:
  - add-task-bar action tag menu
  - add-task-bar short syntax tag suggestions
  - mention config tag suggestions
  - tag-edit picker when My Day is hidden
- Kept the existing sorted tag APIs intact for places that explicitly need alphabetical order.
- Added precomputed `groups` output to `TaskViewCustomizerService`.
- Updated work-view grouped rendering to iterate over precomputed groups instead of using the `keyvalue` pipe in the template.
- Added regression coverage for sidebar-order tag suggestions and grouped customizer output.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

## Testing

- [x] `npm run test:file src/app/features/tasks/add-task-bar/add-task-bar-mentions.spec.ts`
- [x] `npm run test:file src/app/features/tasks/add-task-bar/add-task-bar-actions/add-task-bar-actions.component.spec.ts`
- [x] `npm run test:file src/app/features/tasks/add-task-bar/add-task-bar.component.spec.ts`
- [x] `npm run test:file src/app/features/tag/tag-edit/tag-edit.component.spec.ts`
- [x] `npm run test:file src/app/features/task-view-customizer/task-view-customizer.service.spec.ts`
- [x] `npm run test:file src/app/features/work-view/work-view.component.spec.ts`
- [x] `npm run checkFile` for all modified TypeScript files
- [x] Pre-push hook full test suite


johannesjo:
Thanks for tackling #8419! Before this goes further, I think the tag-source half of this PR has been overtaken by master and needs rework — flagging the details so we can decide direction.

**1. Part 1 looks superseded by #8100 (and would regress it).**

#8100 ("order project and tag dropdowns by tree", `d50c74eda7`) landed on master and already fixed #8419's "remaining pickers" item — it added `tagsInTreeOrder` / `tagsNoMyDayAndNoListInTreeOrder` and switched these dropdowns to them. On current master:

| Picker | master (post-#8100) | this PR |
|---|---|---|
| `add-task-bar-actions.ts:104` | `tagsNoMyDayAndNoListInTreeOrder` ✅ | `tagsNoMyDayAndNoList` |
| `tag-edit.component.ts:91-92` | `tagsInTreeOrder` / `…InTreeOrder` ✅ | `tagsNoMyDayAndNoList` |
| `add-task-bar.component.ts:251` | `…Sorted$` (kept alphabetical, deliberately) | `…NoList$` |
| `mention-config.service.ts:40` | `…Sorted$` (kept alphabetical, deliberately) | `…NoList$` |

So this PR conflicts on the first two lines, and resolving in its favor would downgrade tree/sidebar order back to raw store order. CI is green only because it ran against the pre-#8100 base.

**2. `tagsNoMyDayAndNoList` isn't "sidebar order".**

It's `selectAllTagsWithoutMyDay` — raw store/entity order. The sidebar itself (`magic-nav-config.service.ts:88`) runs that raw list through `buildTagViewTree(...)`, so the rendered order is the **menu-tree** order. The signal that actually matches the sidebar is `tagsNoMyDayAndNoListInTreeOrder` (also what `TaskViewCustomizerService._tagsInSidebarOrder()` uses). Raw order is neither alphabetical nor sidebar order, so the PR's goal isn't met by this source. For the mention/short-syntax cases, #8100 deliberately kept them alphabetical — if we want those on sidebar order, that's a separate intentional change to a tree-order source.

**Suggestion:** drop the tag-source changes (Part 1) entirely — #8419's picker item is already done.

**3. Part 2 (precomputed `groups`) is good — please keep it.**

Master still calls `getOrderedGroupKeys(customized.grouped)` from the `@for` every CD pass. Moving it into the stream `map` computes it once per emission *and* closes the one-frame staleness window (the ordering snapshot now matches `grouped`, before the `animationFrameScheduler` defer). Worth rebasing onto current master and shipping this as the whole PR.

**4. Minor:** the `electron/sync-path-resolver.test.cjs` and `eslint.config.js` changes are unrelated prettier reformatting — best dropped to keep the diff focused.

Could you confirm whether you'd like to rebase this down to just Part 2, given #8100 already covers the pickers?

PR #8439, opened 2026-06-16
fix(op-log): lock snapshot save to prevent lost-update window (#8308)

### LINKED ISSUE #8308: Persistence bug: saveCurrentStateAsSnapshot() has an unlocked lost-update window
## Summary

`saveCurrentStateAsSnapshot()` reads NgRx state first, then `getLastSeq()`, **without holding `LOCK_NAMES.OPERATION_LOG`** — while op appends do hold that lock. An op persisted between the two reads gets `seq <= lastAppliedOpSeq` while its effect is absent from the captured state; the next hydration's tail replay starts after `lastAppliedOpSeq`, so the op is silently never applied locally.

## Evidence (verified against source)

- `src/app/op-log/persistence/operation-log-snapshot.service.ts:82-132` — state read at :85, `getLastSeq()` at :104; no `LockService` usage anywhere in the file. The `await` at :103 (`getCurrentVectorClock`) yields between the two reads, making the interleaving real.
- Appends hold the lock: `operation-log.effects.ts:305-308` (also `operation-write-flush.service.ts:104`, `remote-ops-processing.service.ts:335`).
- Compaction does the identical sequence **inside** the lock: `operation-log-compaction.service.ts:61-143` — and there, capturing lastSeq immediately before the cache write is safe *only because of the lock*.
- Ops do fire around hydration (e.g. TODAY_TAG repair, hydrator :230-233), so the window is reachable; callers are `OperationLogHydratorService:283, :366` via `DataInitService:52` — neither holds the lock, so adding it is safe (note: the lock is **not re-entrant**; mirror the `callerHoldsOperationLogLock` flag pattern from effects :294).

## Suggested fix

Two cheap layers of defense:
1. Wrap the snapshot save in `lockService.request(LOCK_NAMES.OPERATION_LOG, ...)` (matching compaction).
2. Read `lastSeq` **before** the state snapshot — then the worst interleaving degrades to harmless re-replay (re-applying an op is idempotent per the design: `remote-ops-processing.service.ts:391`, absolute-value time-tracking payloads).

**Effort:** tiny. **Risk:** low, but sync-critical — add a spec.

---
Found during a deep op-log analysis; claims independently re-verified. Full report: [op-log-analysis-2026-06.md](https://github.com/super-productivity/super-productivity/blob/claude/operation-log-analysis-b20635/docs/sync-and-op-log/op-log-analysis-2026-06.md) (finding C5).

starter727:
## What

Wrap `saveCurrentStateAsSnapshot()` in `OPERATION_LOG` lock and read `lastSeq` before `getStateSnapshot()`.

## Why

Without the lock, an op appended between reading NgRx state and reading `lastAppliedOpSeq` would get a seq ≤ the snapshot's lastAppliedOpSeq but its effect would be absent from the snapshot. On next hydration the tail replay would start after that seq, silently skipping the op forever.

The read-order inversion (`lastSeq` before `state`) is a second layer of defense: if the lock is ever bypassed, the worst interleaving degrades to harmless re-replay (idempotent) rather than a missed op. Mirrors the pattern already used by `OperationLogCompactionService._doCompact()`.

## Risk

Low. The lock is the same one used by compaction — no new contention path. `getStateSnapshot()` inside the lock is a synchronous NgRx read (sub-ms), so lock hold time is negligible.

Fixes #8308


johannesjo:
**Reviewed via multiple sub-agents (sync-correctness, code-review, test-coverage, adversarial race). Overall: net-positive hardening, but the "idempotent re-replay" justification doesn't hold for one op type, and there's no regression coverage.**

### ✅ What's solid

- **No deadlock/reentrancy.** `saveCurrentStateAsSnapshot()` is reachable only from `hydrateStore()` (hydrator `:284`/`:367`) via `data-init` `reInit()`, which never runs inside an `OPERATION_LOG` lock. None of the other lock holders (remote-ops, superseded-resolver, repair, clean-slate, backup, compaction) call it. The non-reentrant Web Lock is never self-acquired.
- **The master bug is real and this closes it.** Reading `lastSeq` *after* state (+ two awaits) could leave `lastSeq` ahead of the captured state → next boot's `getOpsAfterSeq(lastSeq)` silently skips an op = data loss. The lock + read-order inversion eliminate that direction.

### ⚠️ The fix converts silent op-*loss* into op-*re-replay* — idempotent for everything **except `syncTimeSpent`**

Because NgRx state mutates synchronously at dispatch while the op's `seq` is assigned later, under the same lock (`operation-log.effects.ts:348`), the locked body can read `lastSeq = N` and then a state that already includes an op whose seq-write is still queued behind the lock. The snapshot is then tagged `lastSeq = N` with the op's effect already present → next boot re-applies it via tail replay.

That's harmless for CRUD/move/reorder (entity-adapter ops are idempotent), but **`syncTimeSpent`'s replay branch is additive** — `task.reducer.ts:261` does `currentTimeSpentForDay + duration`, and the action is `isPersistent: true`. A double-apply inflates that day's tracked time by one interval. So the JSDoc's *"applying an already-applied op is idempotent"* isn't universally true.

Probability is low (the save runs only during hydration — boot, plus plugin-triggered `reInitData()` while the user is active), and time-inflation is arguably less bad than silent loss, so I'd still call this a net improvement. But could the comment be amended to acknowledge the `syncTimeSpent` re-replay edge rather than claiming blanket idempotency?

### 🧪 Test gap

No test locks in the regression. The existing spec passes only because `LockService` is `providedIn: 'root'` (real Web Locks runs the callback inline in Chrome) — it would stay green even if the lock wrapping were removed. Worth adding, mirroring the compaction spec's convention: assert `request(LOCK_NAMES.OPERATION_LOG, …)` is called, and that `getLastSeq` resolves before `getStateSnapshot` is read.

### 📝 Minor

- **PR "Risk" section:** the lock now wraps `await saveStateCache()` (a full-state IndexedDB write) plus `getCurrentVectorClock()`/`loadClientId()`, not just a "sub-ms synchronous NgRx read." Conclusion (low risk) still holds — compaction already holds the lock for that write — but the rationale is understated.
- **"Mirrors compaction's pattern":** compaction actually reads **state first, `lastSeq` last** (`operation-log-compaction.service.ts:66`/`:101`) — the opposite inner order — and *deletes* ops `≤ lastSeq`. The two state-cache writers now use opposite orderings; a one-line note (or a follow-up to align them) would help future readers.

Nothing here is a blocker. The two things I'd ask for before merge: amend the idempotency claim, and add the regression test. Opening a follow-up for the truly-lossless fix (quiesce before capture / idempotent `syncTimeSpent` replay).

johannesjo:

PR #8455, opened 2026-06-17
fix(electron): resolve clipboard image file:// paths to blob URLs

Marck21:
## Problem

On Windows, clipboard image paths use a drive-letter prefix (`C:/...`) so the previous `` `file://${path}` `` template produced `file://C:/...` instead of the canonical `file:///C:/...`. This caused images pasted via Ctrl+V or screenshot to appear as broken icons in both Markdown notes and attachment thumbnails.

The same rendering issue affected all platforms: Chromium blocks cross-origin `http://localhost → file://` subresource loads regardless of CSP, so `file:///` paths stored in note content are never loaded by the renderer. Converting them to same-origin `blob:` URLs before rendering is required everywhere.

## Changes

- **`pathToFileUrl()`** — emits `file:///path` on all platforms and `encodeURI`-encodes the path so usernames with spaces (e.g. `C:\Users\John Doe\...`) do not break the markdown URL parser at the first space.
- **`resolveMarkdownImages()`** — extended to convert `file:///clipboard-images/…` paths to `blob:` URLs before markdown HTML reaches the renderer.
- **`resolveClipboardImageUrl()`** — new public method handling both `indexeddb://` (web) and `file:///clipboard-images/` (Electron) URLs; used by the attachment thumbnail list.
- **`_deleteImageElectron()`** — fixed blob URL leak: now revokes and removes the correct `electron-file:<id>` cache entry instead of the wrong bare `id` key.
- **`task-attachment-list`** — updated to call `resolveClipboardImageUrl()` and detect `file:///clipboard-images/` paths for thumbnails.
- **File paste from Explorer** — copied into `clipboard-images/` dir (via `copyClipboardImageFile`) so the resulting URL is always within our controlled directory and resolvable to `blob:`.
- **`clipboard-image.service.spec.ts`** (new) — unit tests for `pathToFileUrl` (Unix / Windows / backslashes / spaces) and `resolveClipboardImageUrl` edge cases.

## Tested on

- Windows 11: Ctrl+V screenshot, Ctrl+V image file from Explorer — both render correctly in Markdown notes and attachment thumbnails.

### DIFF
--- a/src/app/core/clipboard-image/clipboard-image.service.spec.ts
+++ b/src/app/core/clipboard-image/clipboard-image.service.spec.ts
@@ -1,0 +1,72 @@
+import { TestBed } from '@angular/core/testing';
+import { ClipboardImageService, pathToFileUrl } from './clipboard-image.service';
+import { SnackService } from '../snack/snack.service';
+import { GlobalConfigService } from '../../features/config/global-config.service';
+
+describe('pathToFileUrl', () => {
+  it('prefixes a Unix absolute path correctly', () => {
+    expect(pathToFileUrl('/home/user/img.png')).toBe('file:///home/user/img.png');
+  });
+
+  it('adds the third slash for a Windows drive path (forward slashes)', () => {
+    expect(pathToFileUrl('C:/Users/user/img.png')).toBe('file:///C:/Users/user/img.png');
+  });
+
+  it('normalises Windows backslashes', () => {
+    expect(pathToFileUrl('C:\\Users\\user\\img.png')).toBe(
+      'file:///C:/Users/user/img.png',
+    );
+  });
+
+  it('percent-encodes spaces in a Windows username', () => {
+    expect(pathToFileUrl('C:\\Users\\John Doe\\clipboard-images\\clip.png')).toBe(
+      'file:///C:/Users/John%20Doe/clipboard-images/clip.png',
+    );
+  });
+
+  it('percent-encodes spaces in a Unix path', () => {
+    expect(pathToFileUrl('/home/john doe/img.png')).toBe(
+      'file:///home/john%20doe/img.png',
+    );
+  });
+});
+
+describe('ClipboardImageService', () => {
+  let service: ClipboardImageService;
+
+  beforeEach(() => {
+    const snackServiceSpy = jasmine.createSpyObj('SnackService', ['open']);
+    const globalConfigServiceSpy = jasmine.createSpyObj('GlobalConfigService', [
+      'clipboardImages',
+    ]);
+    globalConfigServiceSpy.clipboardImages.and.returnValue(null);
+
+    TestBed.configureTestingModule({
+      providers: [
+        ClipboardImageService,
+        { provide: SnackService, useValue: snackServiceSpy },
+        { provide: GlobalConfigService, useValue: globalConfigServiceSpy },
+      ],
+    });
+
+    service = TestBed.inject(ClipboardImageService);
+  });
+
+  describe('resolveClipboardImageUrl', () => {
+    it('returns null for an unrelated https URL', async () => {
+      expect(
+        await service.resolveClipboardImageUrl('https://example.com/img.png'),
+      ).toBeNull();
+    });
+
+    it('returns null for a file:// URL outside clipboard-images', async () => {
+      expect(
+        await service.resolveClipboardImageUrl('file:///home/user/img.png'),
+      ).toBeNull();
+    });
+
+    it('returns null for an empty string', async () => {
+      expect(await service.resolveClipboardImageUrl('')).toBeNull();
+    });
+  });
+});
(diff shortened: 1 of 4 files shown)


johannesjo (commented):
## Review

Thanks for this — the core direction is right and the bug is real. `pathToFileUrl()` (canonical `file:///`) plus blob-URL conversion is sound and aligns with the codebase's security posture. I ran the changes locally: **lint passes on all 3 files and the component spec is green (12/12)**. The id↔filename round-trip is correct, and all supported image extensions are lowercase so the `[a-z]{2,5}` regex covers them.

A few things I'd ask to address before merge, plus some non-blocking notes.

### 🔴 The fix is incomplete for a common Windows setup — spaces in the path

The default dir is `userData/clipboard-images` → `C:\Users\<username>\AppData\Roaming\superProductivity\clipboard-images`. `pathToFileUrl()` doesn't percent-encode, so when `<username>` contains a space (very common, e.g. "John Doe") the produced markdown doesn't render. Verified with the project's `marked`:

```
SPACED  -> ![pasted image](file:///C:/Users/John Doe/.../clip-1.png)   // stays literal text, no <img>
ENCODED -> <img src="file:///C:/Users/John%20Doe/.../clip-1.png" ...>  // renders
```

The markdown parser stops the URL at the first space, and the resolver regex (`…[^)\s"]*…`) excludes whitespace too, so it can't match it either. Net effect: a Windows user whose username has a space still gets broken clipboard images **in notes** after this PR — the exact case the PR targets. (Attachment thumbnails survive this because resolution uses only the basename + current `basePath`, ignoring the URL's directory.)

**Suggested fix:** percent-encode in `pathToFileUrl`, e.g. `` `file:///${encodeURI(normalized)}` `` (and the `file://` branch likewise). I confirmed the encoded form still passes `isExternalUrlSchemeAllowed`/`isPathSafeToOpen` on both Windows and Unix paths. (`#`/`?` in folder names would still need per-segment `encodeURIComponent`, but spaces are the 95% case.)

### 🟡 The stated mechanism is inaccurate (code comments + PR description)

The comment/description say *"Angular's DomSanitizer strips the `file:` scheme … (SecurityContext.URL only allows http/https/blob/data)."* That isn't what happens:

- Angular's `SAFE_URL_PATTERN = /^(?!javascript:)(?:[a-z0-9+.-]+:|…)/i` allows **any** scheme except `javascript:` — `file:` passes (both the URL and HTML sanitizer paths).
- The app's CSP `img-src` already includes `file:` (`src/index.html:92`).

So neither Angular nor CSP strips `file:`. The real reason `file://` images fail is the renderer **origin/scheme** (e.g. under `ng serve` the page is `http://localhost:4200`, and Chromium hard-blocks `http:`→`file:` subresources regardless of CSP). `blob:` is same-origin, so it works everywhere — the fix is correct and robust, but the *why* is wrong and will mislead the next maintainer. Please correct the comment to describe the origin constraint rather than Angular sanitization.

### 🟡 blob-URL leak on delete (Electron)

The new code caches Electron blobs under `electron-file:${id}` via `URL.createObjectURL`, but `_deleteImageElectron()` only does `this._blobUrlCache.delete(id)` (the wrong key) and never revokes the blob — and its comment *"file:// URLs don't need revocation"* is now stale. Session-bounded, low severity, but worth cleaning up: revoke and delete the `electron-file:${id}` entry on delete, mirroring `_deleteImageWeb()`.

### 🟡 No unit tests for the new service logic

There's no `clipboard-image.service.spec.ts`. The crux of the fix — `pathToFileUrl()` (Unix vs Windows vs spaced path) and the file-URL regex / `resolveClipboardImageUrl()` — is untested; only attachment-component tests were added. `pathToFileUrl` is a trivially testable pure function and is the heart of the change (a spec would also have caught the spaces case above). Per the repo guidelines on adding unit tests for new service logic, please add a small spec.

### ⚪ Non-blocking / follow-up

- **Architecture:** web stores a portable `indexeddb://clipboard-images/<id>`, but Electron embeds machine-absolute `file:///C:/Users/<name>/…` paths into synced/exported note + attachment content. Pre-existing and not worsened here, but it means these images don't render across devices, leak the OS username, and break if the user later changes the image-path setting. A cleaner long-term design stores a logical id for Electron too and resolves at render — worth a separate issue.
- **Regex robustness:** `[a-z]{2,5}` is fine today (handlers force lowercase) but `[a-z0-9]{2,5}` / case-insensitive is more defensive.
- 👍 Nice catch fixing the test spy (`resolveUrl` → `resolveIndexedDbUrl`/`resolveClipboardImageUrl`) — the old spy was mismatched. The `_isResolvableClipboardUrl` substring check is safe (resolution stays confined to `basePath` via `findImageFile` + `validatePath`).

Happy to provide the `encodeURI` patch and a starter `clipboard-image.service.spec.ts` if useful.

Marck21:
Thanks for the thorough review! Addressed all points in the latest commit:

 - Added encodeURI() to pathToFileUrl() so spaces in Windows usernames (e.g. C:/Users/John Doe/) are percent-encoded to valid URLs
 - Corrected the comment — the root cause is Chromium's cross-origin restriction (http://localhost → file://), not Angular DomSanitizer
 - Fixed _deleteImageElectron(): the cache lookup was using key id instead of electron-file:${id}, and URL.revokeObjectURL() was never called — both fixed

johannesjo:

PR #8456, opened 2026-06-17
feat(imex): add ZIP export/import of clipboard image attachments

Marck21:
﻿## Problem

When using Super Productivity on the web (browser), clipboard/pasted images are stored in IndexedDB using `indexeddb://` URLs. If a user later switches to the desktop (Electron) app — or migrates data to another machine — those image attachments are lost: the standard JSON export does not carry them.

## Solution

Fold image attachments into the existing backup as an opt-in. A checkbox **"Include image attachments"** (default: off, no behaviour change) appears above the Export Data button. When checked, each stored image is embedded as base64 under an `imageAttachments` key at the top level of the backup JSON. On import, the presence of that key is detected automatically and images are restored — no extra button, no separate file.

Existing backups without `imageAttachments` import exactly as before.

## Changes

- **`ClipboardImageService.collectImagesForExport()`** — reads all stored images (IndexedDB or Electron filesystem) and returns a `Record<id, { base64, mimeType }>`.
- **`ClipboardImageService.importImagesFromBackup()`** — restores images from that record into IndexedDB (web) or via `saveClipboardImage` IPC (Electron).
- **`FileImexComponent`** — `includeImages` boolean flag; `downloadBackup()` spreads `imageAttachments` into the export object when set; `_processAndImportData()` detects and restores embedded images before handing off to the normal import pipeline.
- **HTML** — `MatCheckbox` with tooltip added above the existing Export Data button; no new buttons.
- **Translations** — two new keys (`INCLUDE_IMAGES_IN_EXPORT`, `S_IMAGES_IMPORTED`) in `FILE_IMEX`.

## Notes

- No new IPC channels; all Electron writes go through the existing `saveClipboardImage` handler.
- The `imageAttachments` key lives outside `data` so it does not touch `AppDataComplete` or the sync/op-log pipeline.
- Depends on #8455 for the `file:///` rendering fix (the two PRs can be reviewed independently).


johannesjo:
Thanks @Marck21, and congrats on your first PR here 🎉 The migration gap you're describing is real and I appreciate you tackling it.

Two things before you invest more:

1. **The two `fix(electron)` commits — can you split those into their own PR?** I notice `master` already emits `file://` URLs for clipboard images, so I want to understand exactly what's broken. Could you attach a quick repro/screenshot of clipboard images failing to render on current `master` (which OS)? If it's a real rendering bug, that fix is valuable on its own and I'll review it independently of the export feature.

2. **The export itself** — I'd prefer not to add a second pair of buttons that have to be paired with the existing export/import. If we solve this, I'd want image attachments folded into the existing backup as an opt-in. That's one complete export instead of a four-step dance that's easy to get half-wrong. Open to your thoughts.

Appreciate the effort — want to make sure it lands somewhere mergeable.

Marck21:
Thanks for the direction! Redesigned based on your feedback:

 - Removed the separate ZIP buttons entirely — added an opt-in "Include image attachments" checkbox above the existing Export Data button (default unchecked, so behavior is unchanged for existing users)
 - Images are embedded as base64 strings under a top-level imageAttachments key in the backup JSON — no extra file, one download
 - Import detects and restores the images automatically; backward-compatible with old backups
Squashed to a single commit

johannesjo:
Thanks for the rework, @Marck21 — the opt-in toggle folded into the existing backup is exactly what I was hoping for, and the core mechanism is right: restoring each image under its original ID so the links already in your notes resolve without having to rewrite note content. That's the key design decision and it holds up. 🎉

I did a thorough pass and there are a few things I'd like addressed before this can land, grouped by priority.

### Before merge (blocking)

1. **Add a size/count guard on export and import.** There's currently no cap on the number of images or their total size (the 2 MB limit is web-only, and it's bypassed on import — `importImagesFromBackup` calls `_saveImageWeb` directly instead of going through `saveImage`). The whole backup is built in memory in a single `JSON.stringify` with no progress UI, so a user with many/large images can freeze or OOM the app. And since the import path is reachable from the import-from-URL dialog / `?importFromUrl=` deep link, a crafted backup link could trigger that on someone deliberately. Please add a total-size check (with a warning before writing) and run the web import through the same size limit as paste.

2. **Restore images *after* the import is confirmed, and report honestly.** In `_processAndImportData()` the images are written and the "{count} restored" success snack fires *before* the V1-data check, the encryption-warning dialog (where the user can still cancel), and `importCompleteBackup()`. So a cancel or a failed import leaves orphaned images plus a success message that wasn't actually true. It also shows "0 image attachment(s) restored" for an empty set, and per-image failures are only logged while still counting toward success. Please move the restore + snack to after a successful import, and surface skipped/failed counts.

3. **Tests for the new logic.** `collectImagesForExport()` / `importImagesFromBackup()` have no unit tests — the spec only stubs them. The base64 encode→decode round-trip is exactly the kind of code that can silently corrupt bytes, so I'd want at least: a round-trip byte-fidelity test (including high-byte values), empty/missing `imageAttachments`, and partial-failure handling.

### Smaller

4. **Export only referenced images.** `collectImagesForExport()` includes *every* stored image, even ones from deleted notes, so the backup carries dead weight. Worth filtering to images actually referenced in content.
5. **Update the title/description** — it still says "ZIP", but the implementation is base64-embedded-in-JSON now.

### One thing worth noting (not blocking)

Now that #8455 has landed, the cross-platform story is better than it might first appear: Electron resolves `file://` clipboard URLs by image ID against the current images folder (not the absolute path baked into the note), so restoring each image under its original ID means **web→web, web→desktop, and desktop→desktop all render**. The one direction still broken is **desktop→web** — the web renderer doesn't resolve `file://` clipboard URLs at all, so desktop-pasted images imported into the web app stay broken even though the bytes are restored. A small symmetric follow-up (resolve `file://…/clipboard-images/<id>.<ext>` by ID on web too) would close that.

Separately, none of this covers SuperSync — images still don't travel on sync, which is the larger underlying gap. Not something this PR needs to solve, just flagging it so we don't consider image portability "done" once this merges.

Really appreciate you digging into a real gap and iterating on the design — let's get the guard + ordering + tests in and this is in good shape.

PR #8479, opened 2026-06-18
docs: add SKILL.md agent integration reference + link AGENTS.md to CLAUDE.md

croakingtoad:
## What this does

**AGENTS.md** — one-character change with a meaningful nudge: the bare text `CLAUDE.md` is now a proper markdown link `[@CLAUDE.md](CLAUDE.md)`. Agents that follow links will actually navigate to and read the file instead of just seeing the name. Small step, hopeful direction.

**SKILL.md** — a new reference file that documents every programmatic integration path for AI agents and automation scripts:

- **Local REST API** (`127.0.0.1:3876`, Electron only) — full route table, query filters, create/update field reference, shell and Python examples
- **Plugin API** — complete `PluginAPI` surface (tasks, projects, tags, hooks, storage, shortcuts, Node scripts), minimal plugin structure, manifest reference, boilerplate pointers
- **URL protocol handler** — `superproductivity://create-task/...` and `task-toggle-start`
- **Electron IPC** — agent-relevant channel list for in-process use
- **Data model** — TypeScript interfaces for Task, Project, Tag
- **Selection guide** — decision table matching scenario to mechanism

The file follows the Claude Code skill frontmatter convention (`name`, `description`, `triggers`) so it loads automatically in Claude Code sessions, but it reads as plain documentation for any agent or human.

## Why

AI coding assistants (Claude Code, Cursor, Copilot, etc.) are increasingly used alongside productivity tools. Having a single, machine-readable reference for "how does an agent talk to this app" makes it easier to build integrations without reverse-engineering the codebase.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,1 +1,1 @@
-Please read CLAUDE.md!!
+Please read [@CLAUDE.md](CLAUDE.md)!!
--- a/SKILL.md
+++ b/SKILL.md
@@ -1,0 +1,314 @@
+---
+name: super-productivity
+description: >
+  Agent integration reference for Super Productivity — a todo & time-tracking app (Angular + Electron + Capacitor).
+  Covers the Local REST API (port 3876, Electron only), Plugin API (web + desktop), and URL protocol handler.
+  Use when an agent needs to create tasks, complete work, query projects/tags, or react to task events.
+triggers:
+  - super productivity
+  - superproductivity
+  - add task to super productivity
+  - local rest api 3876
+  - superproductivity plugin
+  - task management agent integration
+---
+
+# Super Productivity — Agent Integration
+
+Super Productivity is a todo and time-tracking app. This skill covers every mechanism an AI agent can use to create tasks, mark them done, query state, and react to events.
+
+## Quick-start decision tree
+
+```
+Is the Electron desktop app running?
+  YES → Use the Local REST API on port 3876  (simplest, no setup)
+  NO  → Use the Plugin API                  (load a plugin.js via Settings → Plugins)
+```
+
+---
+
+## 1. Local REST API (Electron desktop only)
+
+**Port:** `127.0.0.1:3876`  
+**Enable:** Settings → Misc → Enable Local REST API  
+**Works in web version at :4200 / ng serve:** No — Electron only.
+
+### Full route table
+
+| Method | Path | Description |
+|--------|------|-------------|
+| GET | `/status` | App status + current task summary |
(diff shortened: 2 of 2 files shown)


croakingtoad:
Obviously the above was written by an overly helpful Claude Code trying to explain why a markdown link was SO helpful. :D

Anyway, agents often don't follow instruction, and every little thing helps, so I proposed adding a markdown link.

I'm not fishing for any recognition, you can delete this PR and add it or not, but I wanted to suggest it anyway. :)

johannesjo:
Thanks for this, and for the good humor in the follow-up comment 🙂 The instinct is right — agent integration *is* underdocumented, and making it discoverable has real value.

A couple of reasons I'm going to pass on merging `SKILL.md` as-is, though:

**1. It overlaps with docs we already maintain.** `docs/wiki/3.01-API.md` already covers all three API systems (Sync Server, Local REST, Plugin) and intentionally points at the source-of-truth files (`src/app/core/electron/local-rest-api-handler.service.ts`, `packages/plugin-api/`, `docs/plugin-development.md`) rather than restating them — specifically so they don't drift. A second standalone copy at repo root would be a 4th place to keep in sync.

**2. A few details have already drifted from the implementation**, which is exactly the failure mode I worry about for hand-maintained API tables. For example, in the `GET /tasks` filter section: the handler reads `includeDone` (not `isDone`), `query` (not `search`), and doesn't implement `dueDay` / `limit` / `offset`. The REST responses are also wrapped in `{ ok, data }`, so the Python `add_task` example would need `r.json()["data"]["id"]`. And the Node-script call is `executeNodeScript({ script })` with the `nodeExecution` permission.

**3. Re: the frontmatter** — a root `SKILL.md` won't actually auto-load as a Claude Code skill (skills live under `.claude/skills/`), so in practice it behaves as a plain doc regardless.

If you'd like to land the genuinely additive part — the REST `curl`/quick-start examples are nice — the best home is a short section in `docs/wiki/3.01-API.md`, with the param names and response envelope matched to the handler. Happy to take a focused PR for that.

The `AGENTS.md` link tweak is harmless; I may just fold it in separately.

Really do appreciate you taking the time to suggest it\!

croakingtoad:
Good catches, thank you — updated.

The SKILL.md is now a thin pointer (three-row table + one sentence) that points at `docs/wiki/3.01-API.md` and the three source-of-truth files, following the same pattern you already use. No restated details, nothing to drift.

The AGENTS.md change stands as-is — it's just turning the bare filename into a markdown link.

PR #8481, opened 2026-06-18
fix(caldav): support Vikunja project discovery

### LINKED ISSUE #8471: 🚨 Vikunja CalDAV Todo integration Connection failed
### Version Used

18.9.1L

### Current Behavior

I get "Connection failed" when trying to add a LAN-hosted Vikunja instance CalDAV connection.

### Expected Behavior

It should connect and retrieve the TODOs in Vikunja. It does work with a free Fruux account and based on the XML responses of that vs Vikunja it looks like SP may be looking at the wrong URL? Expecting it in a calendar subdir when it's under principals I think?

### Steps to Reproduce

1. Show Issue Provider
2. Click CalDAV Todo
3. Enter the following details for my LAN-hosted Vikunja instance:
- Base URL: `http://192.168.0.5:3456/dav/principals/loki/`
- CalDAV resource: `Inbox` and `projects` tried
- CalDAV username: `loki`
- CalDAV password: [TOKEN] and [PASSWORD] tried

4. When pressing Test Connection I get "Connection failed" 

### Console Output

```shell
Nothing in Console but Network tab has the following responses:


<!-- http://192.168.0.5:3456/dav/principals/loki/ -->

<?xml version="1.0" encoding="UTF-8"?>
<D:multistatus xmlns:D="DAV:" xmlns:C="urn:ietf:params:xml:ns:caldav"
    xmlns:CS="http://calendarserver.org/ns/">
    <D:response>
        <D:href>/dav/projects</D:href>
        <D:propstat>
            <D:prop>
                <D:current-user-principal>
                    <D:href>/dav/principals/loki//</D:href>
                </D:current-user-principal>
            </D:prop>
            <D:status>HTTP/1.1 200 OK</D:status>
        </D:propstat>
    </D:response>
</D:multistatus>

<!-- http://192.168.0.5:3456/dav/principals/loki// -->
<?xml version="1.0" encoding="UTF-8"?>
<D:multistatus xmlns:D="DAV:" xmlns:C="urn:ietf:params:xml:ns:caldav"
    xmlns:CS="http://calendarserver.org/ns/">
    <D:response>
        <D:href>/dav/projects</D:href>
        <D:propstat>
            <D:prop>
                <D:displayname>projects</D:displayname>
                <C:calendar-user-address-set>
                    <D:href>/dav/projects</D:href>
                </C:calendar-user-address-set>
                <D:principal-URL>
                    <D:href>/dav/projects</D:href>
                </D:principal-URL>
                <C:calendar-home-set>
                    <D:href>/dav/projects</D:href>
                </C:calendar-home-set>
                <D:principal-collection-set>
                    <D:href>/dav/projects</D:href>
                </D:principal-collection-set>
            </D:prop>
            <D:status>HTTP/1.1 200 OK</D:status>
        </D:propstat>
        <D:propstat>
            <D:prop>
                <C:calendar-user-type />
                <D:alternate-URI-set />
                <email-address />
                <language />
                <C:schedule-inbox-URL />
                <C:schedule-outbox-URL />
                <C:schedule-default-calendar-URL />
                <resource-type />
                <resource-vehicle-type />
                <resource-vehicle-make />
                <resource-vehicle-model />
                <resource-vehicle-is-electric />
                <resource-vehicle-range />
                <resource-vehicle-seating-capacity />
                <resource-contact-person />
                <resource-contact-person-vcard />
                <room-type />
                <room-seating-capacity />
                <room-building-address />
                <room-building-story />
                <room-building-room-number />
                <room-features />
                <D:supported-report-set />
            </D:prop>
            <D:status>HTTP/1.1 404 Not Found</D:status>
        </D:propstat>
    </D:response>
</D:multistatus>
```

### Documentation

symonbaikov:
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [ ] Other (please describe)

## What changed

Fixes CalDAV Todo calendar discovery for Vikunja-style setups where the configured URL points to the principal endpoint, e.g. `/dav/principals/<user>/`, and discovery returns `/dav/projects/` as the calendar home.

The CalDAV provider now:
- checks all discovered calendar homes instead of only the first one
- avoids selecting the calendar home/root collection itself as the VTODO calendar
- matches concrete project calendars by display name, URL segment, or normalized path
- documents the Vikunja principal URL + project resource setup

Fixes #8471

## Checklist

## Verification

- `npm run test:file src/app/features/issue/providers/caldav/caldav-client.service.spec.ts`
- `npm run checkFile src/app/features/issue/providers/caldav/caldav-client.service.ts`
- `npm run checkFile src/app/features/issue/providers/caldav/caldav-client.service.spec.ts`
- `npx prettier --write docs/wiki/3.07-Issue-Integration-Comparison.md`
- `npm test`

Manual Vikunja smoke test was not run because I do not have access to a Vikunja instance.

### DIFF
--- a/docs/wiki/3.07-Issue-Integration-Comparison.md
+++ b/docs/wiki/3.07-Issue-Integration-Comparison.md
@@ -91,7 +91,7 @@ Every issue provider supports the same basic operations: checking that the integ
 - **Notes:** Task notes/description support.
 - **Due dates:** Start date with time support.
 - **Related tasks:** Task relationship tracking.
-- **URL format (example Nextcloud):** `https://<NC-URL>/remote.php/dav/calendars/<USER>/<CALENDAR_NAME>/`. Use the **“Copy private link”** (or CalDAV URL) from your calendar—**not** the subscription/export link (that is for the iCal integration). Items without a valid VTODO are skipped.
+- **URL format:** For Nextcloud, use `https://<NC-URL>/remote.php/dav/calendars/<USER>/<CALENDAR_NAME>/` from **“Copy private link”** (or the CalDAV URL), not the subscription/export link. For Vikunja, use the principal URL shown in Vikunja's CalDAV settings, e.g. `https://<VIKUNJA-URL>/dav/principals/<USER>/`, and set the CalDAV resource to the project name or project ID. Items without a valid VTODO are skipped.
 
 ### 7. Calendar (iCal)
 
--- a/src/app/features/issue/providers/caldav/caldav-client.service.spec.ts
+++ b/src/app/features/issue/providers/caldav/caldav-client.service.spec.ts
@@ -93,6 +93,61 @@ const makeIssue = (id: string): CaldavIssue => ({
   etag_hash: 1,
 });
 
+interface TestCalendarLike {
+  displayname?: string;
+  url: string;
+  calendarQuery: jasmine.Spy;
+}
+
+interface TestCalendarHomeLike {
+  displayname?: string;
+  url: string;
+  findAllCalendars: jasmine.Spy<() => Promise<TestCalendarLike[]>>;
+}
+
+interface TestClientCacheLike {
+  client: { calendarHomes: TestCalendarHomeLike[] };
+  calendars: Map<string, TestCalendarLike>;
+}
+
+interface TestGetClientTarget {
+  _get_client: (cfg: CaldavCfg) => Promise<TestClientCacheLike>;
+}
+
+interface TestGetCalendarTarget {
+  _getCalendar: (cfg: CaldavCfg) => Promise<TestCalendarLike>;
+}
+
+const makeCalendar = (
+  url: string,
+  displayname?: string,
+  canQuery = true,
+): TestCalendarLike => ({
+  url,
+  displayname,
+  calendarQuery: jasmine.createSpy('calendarQuery').and.resolveTo(canQuery ? [] : null),
+});
+
+const makeCalendarHome = (
+  url: string,
+  calendars: TestCalendarLike[],
+  displayname?: string,
+): TestCalendarHomeLike => ({
+  url,
+  displayname,
+  findAllCalendars: jasmine.createSpy('findAllCalendars').and.resolveTo(calendars),
+});
+
+const makeFailingCalendarHome = (
+  url: string,
+  error: Error,
+  displayname?: string,
+): TestCalendarHomeLike => ({
+  url,
+  displayname,
+  findAllCalendars: jasmine.createSpy('findAllCalendars').and.rejectWith(error),
+});
+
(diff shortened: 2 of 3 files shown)


johannesjo:
Thanks for this — nice, well-tested fix. I reviewed the diff, ran the spec (35/35 green) and `checkFile` (clean), and traced `@nextcloud/cdav-library` to confirm the discovery behavior. A few notes, none of them blockers.

**Verified / good**

- Iterating all `calendarHomes` instead of `[0]` is the right root-cause fix for the Vikunja principal → `/dav/projects` home case.
- The trailing-slash normalization is genuinely needed: cdav-library *does* try to skip the self-collection (`path === this._url || path + "/" === this.url`), but that check is trailing-slash-fragile and Vikunja even emits `//` double slashes, so the home collection leaks into `findAllCalendars()`. `_normalizeCalDavPath` handles both.
- I confirmed `calendar.url` and `calendarHome.url` are both server-relative pathnames (both from the library's `pathname()`), so the home-exclusion comparison is comparing like with like. 👍

**Worth a second look**

1. **The matching widened for all CalDAV providers, not just Vikunja.** The old logic was `(displayname || uri) === resource` — the URL segment was only consulted when `displayname` was empty. `_matchesCalendarResource` now also matches on the URL segment and `endsWith('/' + resource)` unconditionally. With multiple calendars and `.find()` returning the first hit, this can select a *different* calendar than before. Minimal repro:

   ```
   A = { displayname: 'Work',     url: '.../Personal/' }   // uri === resource
   B = { displayname: 'Personal', url: '.../b123/' }       // displayname === resource
   resource = 'Personal'
   // old → B (displayname match)
   // new → A (uri match, and A is first)
   ```

   Edge case, but it touches existing Nextcloud setups, so worth being deliberate about.

2. **The single-calendar fallback is a quiet cliff.** `concreteCalendars.length === 1 && matches(home, resource)` resolves `resource: 'projects'` to the only project — great for the reporter's setup, but the moment a second project exists, the same config silently returns "calendar not found" (specific project name or ID still works, matching the new docs). Consider an `IssueLog` line when this fallback path is taken so future debugging isn't mysterious.

3. **A network error on an earlier home aborts the whole loop.** `findAllCalendars().catch(err => this._handleNetErr(err))` rethrows, so a transient failure on `calendarHomes[0]` prevents ever reaching `[1]`. Not a regression (old code only used `[0]`), but since the point of the change is to support multiple homes, you may want to continue past a failed home and only fail if all of them fail.

**One gap:** I couldn't verify end-to-end against a live Vikunja. This fix is correct *given* `client.connect()` populates `calendarHomes`; if `connect()` itself is what fails on Vikunja's mixed 200/404 principal response, the symptom would persist. Confirmation from the reporter via the preview build would close that out.

johannesjo:
Thank you very much! <3

PR #8484, opened 2026-06-19
fix(electron): crisp, flicker-free Linux tray icon (#4905)

### LINKED ISSUE #4905: Low quality trey icon (Linux gnome)
Image

Is it possible to add a tray icon in better quality (Linux platform, Flatpak)?

johannesjo:
Fixes #4905.

## Problem
Issue #4905 holds two distinct Linux tray-icon problems:

- **Blurry icon** (original report, GNOME Flatpak): the indicator icons were 16×16 monochrome PNGs (32×32 `@2x`). GNOME's panel renders at ~22–24px, so a 16px source got *upscaled* → blur.
- **Flicker** (recent comment, Arch/Hyprland + `wayle`): the icon flickered between a full ring and a partial-arc frame. Matching the screenshots to assets confirmed the "bad" frame is just a `running-anim` progress frame — every `Tray.setImage()` makes a StatusNotifierItem host redraw the whole item, so cycling the 16 animation frames flickers on hosts like wayle.

## Fix
- **Flicker:** gate the progress animation inside `getRunningIconPath()` so **Linux** returns a single static `running` icon; Windows/macOS keep the 16-frame animation. The gate is at the one chokepoint, so all three call sites (`SET_PROGRESS_BAR`, `CURRENT_TASK_UPDATED`, `syncTray`) are covered. Combined with the existing `curIco` guard, this drops steady-state `setImage` calls on Linux to zero. Elapsed/remaining time still shows in the tray context-menu label.
- **Blur:** re-render the static `stopped`/`running` icons (light+dark, +`@2x`) from vector sources at **24/48px** so GNOME downscales instead of upscaling.

Before (16px, upscaled → blurry) vs after (24px → crisp):

| | BEFORE | AFTER |
|---|---|---|
| stopped/running | soft, boxed | crisp ring + glyph |

## Tests
- Updated the Linux test to expect the static icon; added a darwin test asserting the animation is preserved off-Linux.
- `node --test electron/indicator.test.cjs` → 7/7; full electron suite → 129/129; `checkFile` clean.

## Notes / scope
- The 64 `running-anim` frames are intentionally left at the legacy style (Linux no longer uses them); regenerating them for full Win/macOS visual consistency is a possible follow-up.
- The **wayle flicker fix is device-unverified** — I can't run a real SNI host here. Static-analysis confidence is high (constant icon path + `curIco` guard ⇒ no repeated `setImage`), but it'd be good to have @tobias-haenel confirm on a build.


tobias-haenel:
had some minor issues installing electron development environment on my system. But I finally tested it under Archlinux + wayle and the icon during an active task is now static and the icon is now more crisp. Thanks for this fix :+1: ! Finally no more flicker, but a little sad to see the animation gone. But no flicker is way more important to me.

johannesjo:
Thank you very much for testing!  Much appreciated! <3

PR #8493, opened 2026-06-19
feat(tasks): bulk unschedule overdue tasks

### LINKED ISSUE #8433: 💡 Unschedule all overdue tasks
### Problem Statement

There is no simple way to unschedule all overdue tasks. Either on the planner or the today view you have to manually right click each and press unschedule.

### Possible Solution

Add an "Unschedule all" button in the planner on top (or at the bottom) of the overdue-tasks list. 
Add an "Unschedule all" button in the Today view, next to the "Add to today" button. 

_No response_

### Additional context

_No response_

symonbaikov:
## Problem

Closes #8433.

There was no quick way to unschedule all overdue tasks from the Today view. Users had to unschedule each overdue task individually.

## Solution

https://github.com/user-attachments/assets/9f63e069-13c0-4cf0-81ff-663c35fa70ef

Adds a Today-view bulk action for overdue tasks:

- adds an **Unschedule all** button next to **Add to Today** in the overdue section
- introduces a persistent bulk `unscheduleTasks({ taskIds })` action
- handles the update in one scheduling meta-reducer pass
- clears `dueDay`, `dueWithTime`, and `remindAt`
- removes affected tasks from `TODAY_TAG.taskIds` and planner days
- skips missing, done, or already unscheduled tasks
- shows a confirmation dialog when more than 5 overdue tasks are affected
- documents the new action in the scheduled-task wiki page

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [ ] Other

## Checklist

- [x] Added/updated tests
- [x] Updated documentation
- [x] Verified manually
- [x] No breaking changes

## Testing

- `npm run test:file src/app/root-store/meta/task-shared-meta-reducers/task-shared-scheduling.reducer.spec.ts`
- `npm run test:file src/app/features/work-view/work-view.component.spec.ts`
- `npm run test:file src/app/op-log/core/action-types.enum.spec.ts`
- `npm run test:file src/app/op-log/core/persistent-action-types.spec.ts`
- `npm run test:file src/app/op-log/persistence/compact/action-type-codes.spec.ts`
- `npm run test:file src/app/op-log/validation/state-validity-after-actions.spec.ts`
- `npm run test:file src/app/root-store/meta/task-shared.reducer.spec.ts`
- `npm run checkFile` for all changed TS files
- `git diff --check`
- Pre-push hook passed full lint/test suite
- Manual smoke tested in dev mode


symonbaikov:
Addressed the review follow-up in c134da5f0.\n\n- Added `unscheduleTasks` to the Android native reminder cancellation path that already handles `taskIds`, with coverage for Android=true and Android=false.\n- Kept planner-day cleanup in the task shared scheduling meta-reducer, matching `planTasksForToday`; no redundant `planner.reducer.ts` handler was added.\n- Left issue two-way-sync provider-side schedule clearing as a follow-up, per your note.

johannesjo:
**Re: native Android reminder cancellation (follow-up to concern #2 in #8433)**

Looked at this again — my concern #2 in the issue was framed stronger than it needs to be, so flagging the softer reality here.

The bulk `unscheduleTasks` action isn't wired into the imperative cancellation effects in `task-reminder.effects.ts` (the way single-task `unscheduleTask` is). But that's **not** a blocker: `scheduleNotifications$` in `mobile-notification.effects.ts` reconciles `selectAllTasksWithReminder` and cancels any reminder that drops out of the list, and it runs on the legacy Android WebView too (`isNative = Capacitor.isNativePlatform() || isAndroidWebView`). So clearing `remindAt` in the meta-reducer already gets the native alarm cancelled.

Still worth adding for consistency with single-task unschedule (and to cover immediate cancel + the narrow background-snooze-alarm edge that the reconciler may not have tracked yet):

```ts
// task-reminder.effects.ts — cancelNativeReminderOnDialogAction$
ofType(
  TaskSharedActions.reScheduleTaskWithTime,
  TaskSharedActions.planTasksForToday,
  TaskSharedActions.unscheduleTasks, // add
  PlannerActions.planTaskForDay,
),
```

That effect already iterates `action.taskIds`, so nothing else changes. Optional — fine to merge without it.

Otherwise this looks great: single bulk action + one meta-reducer pass mirroring `planTasksForToday`, planner-day cleanup in the same pass, and solid test coverage. 🙏

symonbaikov:
Adjusted the Android reminder follow-up in 8b238ac96 to match your clarification.\n\n- Kept `unscheduleTasks` wired into the existing `cancelNativeReminderOnDialogAction$` path that already handles `taskIds`.\n- Removed the extra runtime helper and brittle `window.SUPAndroid` spy assertions from the tests; coverage is now the same lightweight pass/filter style as the neighboring Android effect tests.\n- No planner reducer or issue two-way-sync changes.

Raian256:
@johannesjo would you so kind as to check this? thanks!

johannesjo:
@symonbaikov can you please check the conflicts?

johannesjo (changes requested):
I rechecked the exact head (`8b238ac9`), the full discussion and linked issues, current `master`, CI, and the deployed preview. The feature itself earns its place, and the visible placement beside **Add to Today** is the right UI approach; I would keep it there rather than hide it in an overflow menu. All substantive checks are green, but the branch currently conflicts with `master` and the following correctness issues need to be addressed after rebasing.

### Blocking sync correctness

1. **The new persistent action is not safe across mixed app versions.** [`unscheduleTasks` is a brand-new action/code](https://github.com/super-productivity/super-productivity/blob/8b238ac96a4f9cdcb3c395d7bab233e724961f6a/src/app/root-store/meta/task-shared.actions.ts#L189-L198). An older client has neither its reducer nor the `HUT` compact-code mapping, so replay dispatches an unknown action and leaves the tasks scheduled. The replay coordinator nevertheless [returns the whole batch as applied](https://github.com/super-productivity/super-productivity/blob/8b238ac96a4f9cdcb3c395d7bab233e724961f6a/packages/sync-core/src/replay-coordinator.ts#L113-L152), allowing sync progress to advance permanently past the ignored intent. Please represent this through a released, backward-tolerated payload/envelope that materializes compatible task updates, with any new cleanup semantics gated by a marker, and add a mixed-version replay regression. A schema bump alone is not a fix for the released fleet.

2. **Conflict recovery keeps only the first task in the bulk operation.** Capture stores all IDs but also chooses [`entityIds[0]` as the scalar `entityId`](https://github.com/super-productivity/super-productivity/blob/8b238ac96a4f9cdcb3c395d7bab233e724961f6a/src/app/op-log/capture/operation-log.effects.ts#L248-L264). The server [rejects the entire operation when any declared entity conflicts](https://github.com/super-productivity/super-productivity/blob/8b238ac96a4f9cdcb3c395d7bab233e724961f6a/packages/super-sync-server/src/sync/services/operation-upload.service.ts#L445-L476), but ordinary superseded recovery [groups only by that scalar ID and emits one LWW replacement](https://github.com/super-productivity/super-productivity/blob/8b238ac96a4f9cdcb3c395d7bab233e724961f6a/src/app/op-log/sync/superseded-operation-resolver.service.ts#L165-L258) before rejecting the original row. For an op over A/B/C that conflicts on B, only A remains represented for upload; B/C stay optimistic locally and other clients diverge. Add whole-scope recovery plus an integration regression where a non-first ID conflicts.

3. **Remote payload IDs are not constrained to the operation's declared scope.** The replayed handler [passes `action.taskIds` straight into the bulk mutation](https://github.com/super-productivity/super-productivity/blob/8b238ac96a4f9cdcb3c395d7bab233e724961f6a/src/app/root-store/meta/task-shared-meta-reducers/task-shared-scheduling.reducer.ts#L390-L392), while conflict detection trusts `op.entityIds`. A malformed op declaring `[A]` but carrying payload `[B]` therefore unschedules B and changes its Today/planner placement without conflict protection for B; a non-array payload can also throw during replay. Validate the payload shape and require/intersect it with `meta.entityIds`, with a mismatched-scope remote-replay test.

4. **The UI can create an operation SuperSync can never accept.** The handler [dispatches the unbounded overdue list](https://github.com/super-productivity/super-productivity/blob/8b238ac96a4f9cdcb3c395d7bab233e724961f6a/src/app/features/work-view/work-view.component.ts#L534-L563), but the wire schema [caps `entityIds` at 1,000 per operation](https://github.com/super-productivity/super-productivity/blob/8b238ac96a4f9cdcb3c395d7bab233e724961f6a/packages/shared-schema/src/supersync-http-contract.ts#L58-L68). At 1,001 overdue tasks the reducer changes local state first and upload is rejected, leaving divergence. Enforce the shared limit before optimistic dispatch and show explicit feedback unless the one-op protocol representation is redesigned; silently chunking would violate the one-intent/one-op invariant.

### Functional and UI correctness

5. **Overdue subtasks nested under an overdue parent are omitted.** The selector [removes those children from the top-level array and nests them under the parent](https://github.com/super-productivity/super-productivity/blob/8b238ac96a4f9cdcb3c395d7bab233e724961f6a/src/app/features/tasks/store/task.selectors.ts#L289-L323), but `unscheduleAllOverdue()` maps only the top-level IDs. The parent is cleared while the overdue child remains and reappears, and the confirmation count under-reports. Derive eligible IDs/count from the flat `selectUndoneOverdue` selector and add a parent-plus-overdue-child test.

6. **Auto-created calendar time blocks are orphaned.** The existing cleanup effect [only listens for single-task `unscheduleTask`](https://github.com/super-productivity/super-productivity/blob/8b238ac96a4f9cdcb3c395d7bab233e724961f6a/src/app/features/calendar-integration/time-block/time-block-sync.effects.ts#L151-L169). Bulk unschedule clears `dueWithTime` but leaves the corresponding Google Calendar event behind. Pass the bulk action's `taskIds` through the existing bounded `_queueDeleteTimeBlocks$` path and add coverage.

7. **Keep the visible button, but fix its responsive and keyboard behavior.** Both full labels sit in a [non-wrapping row](https://github.com/super-productivity/super-productivity/blob/8b238ac96a4f9cdcb3c395d7bab233e724961f6a/src/app/features/work-view/work-view.component.scss#L277-L284), and the new button has [`tabindex="1"`](https://github.com/super-productivity/super-productivity/blob/8b238ac96a4f9cdcb3c395d7bab233e724961f6a/src/app/features/work-view/work-view.component.html#L249-L267). On the preview, English essentially fills a 320 px row; longer labels wrap outside the fixed-height Material buttons, and keyboard focus jumps to these overdue actions before the header controls. Remove the positive tabindex values and let the wrapper wrap or stack at the narrow breakpoint.

I am not reopening issue-provider two-way sync here; that was explicitly accepted in the discussion as follow-up scope.

symonbaikov:
> @symonbaikov can you please check the conflicts?

Will fix them today.

johannesjo (dismissed):
@-

johannesjo (changes requested):
Re-review of the exact current head `8b238ac9` after the conflict follow-up: **requesting changes remains warranted**.

I rechecked the complete diff, the accepted scope of issue #8433, current `master`, SuperSync validation/replay, conflict recovery, calendar side effects, selectors, and the latest CI/review state. All seven findings from the [previous review](https://github.com/super-productivity/super-productivity/pull/8493#pullrequestreview-4829641509) remain unresolved:

1. The new persistent `unscheduleTasks` action is a no-op on older released clients while replay can still advance, causing mixed-version divergence.
2. SuperSync rejects the whole multi-entity op on any conflict, but superseded-op recovery groups this action by scalar `entityId` and preserves only the first task.
3. Replay mutates `payload.taskIds` without requiring it to match the declared `meta.entityIds` conflict/authentication footprint.
4. The UI can create more than 1,000 entity IDs although SuperSync rejects such an operation after the optimistic local update.
5. The overdue selector includes nested overdue subtasks, but the component dispatches only top-level IDs.
6. Google Calendar time-block deletion listens for single-task unscheduling, so bulk unscheduling can leave orphaned calendar events.
7. The positive `tabindex` and non-wrapping two-label row introduce keyboard focus-order and narrow/localized-layout regressions.

The required rebase also exposes failures that the currently green checks (which ran against the old base) cannot cover:

- `work-view.component.ts` has a direct RxJS import conflict; the resolution must retain both `startWith` from master and `take` from this PR. The new WorkView test fixture must also gain the new `CalendarIntegrationService`, `breakTimeToday$`, and logical-day selector setup from master or those tests will fail after rebase.
- Master and this PR each independently add one `ActionType` and each change the exact-count assertion from 146 to 147. The combined enum has 148 values, while the auto-merged spec still expects 147.

The “confirm above N tasks” threshold matches the maintainer-approved issue discussion, and Planner UI plus issue-provider two-way sync were explicitly accepted as follow-ups; I am not treating those as blockers.

Please rebase and request another review after the sync, functional, and accessibility issues above are addressed.

johannesjo:
@symonbaikov could you have a look at the conflicts?

johannesjo:

Thanks for the follow-up! Structurally this head lines up with all seven points from the earlier review – reusing the released `updateTasks` action with an `isBulkUnschedule` marker, whole-scope superseded recovery, scope validation, the 1000-ID cap, the flat overdue selector, calendar time-block cleanup, and the tabindex/wrap fixes – but the chosen wire encoding undoes the central piece.

### Blocking
- **`src/app/root-store/meta/bulk-unschedule.util.ts:31`** – the payload encodes clearing as `changes: { dueDay: undefined, dueWithTime: undefined, remindAt: undefined }`, and `isBulkUnscheduleTaskUpdate` requires those keys to be *present* (`'dueDay' in changes && changes['dueDay'] === undefined`). `JSON.stringify` drops undefined-valued keys, and every cross-device path is JSON (`packages/sync-providers/src/super-sync/super-sync.ts:235` does `JSON.stringify({ ops, ... })` in `uploadOps`; the file-based providers serialize through `encrypt-and-compress-handler.service.ts`, which is also `JSON.stringify`). So the op reaches every other device with `changes: {}`, and then:
  - `normalizeBulkUnschedulePayload` in `src/app/op-log/apply/operation-converter.util.ts` sees the marker, fails `isValidatedBulkUnschedulePayload`, and rewrites the payload to `tasks: [], isBulkUnschedule: false` behind a `SyncLog.warn` – replay is a complete no-op on current clients;
  - old released clients replay plain `updateTasks` with `changes: {}`, which `updateMany` also no-ops – defeating the util's own stated rationale that "older clients can clear the task schedule even though they do not know the marker".

  The unschedule never propagates: the originating device clears its schedules while every other device keeps them (and keeps the tasks in its Today list) – permanent silent divergence, the exact failure class this iteration set out to fix. Capture-time validation passes because CHECKPOINT A runs on the in-memory op before any serialization (`src/app/op-log/capture/operation-log.effects.ts:301`), `validateOperationPayload` has no call site on the download path, and `extractEntityChanges` returns `[]` for TASK bulk updates, so nothing rescues the clears and nothing ever surfaces.

  The encoding needs to survive a JSON round-trip. Two directions: validate only the ID scope (`tasks[i].id === entityIds[i]`, tolerating absent keys in `changes`) and let the marker path reconstruct the clears – accepting that old clients then no-op on `changes: {}`; or carry JSON-representable values (`null`) if released reducers and typia state validation provably tolerate `null` in `dueDay`/`dueWithTime`/`remindAt`. The `null` route is the more robust of the two: the SQLite op-log adapter also JSON-serializes records (`sqlite-op-log-adapter.ts:245`), so once that path is wired up a non-JSON-representable encoding would additionally break local persistence and superseded-op recovery there. Whichever you choose, please add a regression test that pushes the op through `JSON.parse(JSON.stringify(op))` before replay/validation – all current specs assert on in-memory objects, which is exactly why they pass.

### Should fix before merge
- **`src/app/root-store/meta/task-shared-meta-reducers/task-shared-scheduling.reducer.ts:162`** – `handleBulkUnScheduleTasks` filters target IDs by live state (`!!task && !task.isDone`), so the same op produces different results depending on the order each device applies it in. It's the only `isDone` check in this meta-reducer (single-task `handleUnScheduleTask` clears fields unconditionally), and the base `updateTasks` reducer applies the payload changes to *all* listed tasks including done ones, so the two reducers disagree about done tasks. Concrete case: task X is overdue on A and B; B marks X done concurrently with A's "Unschedule all" (which the superseded-op recovery in this PR deliberately re-creates intact after a conflict). A applies unschedule-then-done and removes X from `TODAY_TAG.taskIds`; B applies done-then-unschedule, the `isDone` filter drops X, and X stays in B's `TODAY_TAG.taskIds` – a synced entity, so the two devices diverge permanently (under a marker-reconstruction fix for the blocker, the `dueDay`/`remindAt` fields diverge too). It's masked today only because the blocker makes remote replay a no-op – it goes live the moment that is fixed. Fix: drop the `isDone` filter (mirroring `handleUnScheduleTask`), or make done-task semantics identical in the meta-reducer and base reducer, plus a replay test asserting done-then-unschedule and unschedule-then-done converge.
- **Branch state** – the PR still conflicts with `master` (GitHub reports `CONFLICTING`; asked about on 2026-07-31 and 2026-08-04), and no CI has run at the current head, so there is no test signal. One upside: the earlier ActionType-count merge conflict should be gone now that this head no longer adds a new `ActionType`.

### Nits
- **PR description** – it still describes the previous head's design ("introduces a persistent bulk `unscheduleTasks({ taskIds })` action"), while this head reuses `updateTasks` plus the `isBulkUnschedule` marker. Worth updating so the description matches the diff.

---

johannesjo:
Answering the encoding question from my review — **use `null`, not `undefined`.** It turns out this isn't a trade-off between three imperfect options; `null` is already the convention for these exact fields and is provably safe for the released fleet.

`Task` already types all three as nullable:

```ts
// src/app/features/tasks/task.model.ts
dueWithTime?: number | null;   // :103
dueDay?: string | null;        // :116
reminderId?: string | null;    // :134
```

and the doc comment right above them already says a cleared field is "undefined/null". More importantly, master has been *writing* `null` for them over the same released `updateTask` path for months — `src/app/features/issue/providers/caldav/caldav-common-interfaces.service.ts:76-79` and `src/app/features/tasks/short-syntax.ts:849,888`. Both landed in `794a13a76a` and shipped in **v16.5.0** (Dec 2025), so every client in the field has been receiving and applying `dueDay: null` / `dueWithTime: null` for months. Nothing to prove and nothing to bump — typia accepts it, the reducers apply it, and old clients handle it the same way they handle it today.

So the fix is just: send `null` instead of `undefined` for the cleared fields, and the marker-path reconstruction can go away entirely.

Worth knowing why this matters beyond this PR: the `undefined` shape is a live bug class, not a theoretical one. I verified this week that the same pattern is shipping in `_hideSubTasksMode` (expanding subtasks resolves to `undefined`, `JSON.stringify` drops the key, and the expand never replays on other devices — live since v18.14.0), plus two more sites. The distinguishing factor is exactly what your handler does: reducers that pass `payload.changes` **verbatim** to `adapter.updateOne` are vulnerable, while reducers that rebuild `changes` from destructured payload fields are immune, because a dropped key destructures back to `undefined` identically.

The other item from the review still stands — the order-dependent `isDone` filter in the new meta-reducer handler. Both are small; the branch also needs a rebase.

johannesjo:
Heads-up: this PR's bulk-unschedule payload clears fields via `changes: { …: undefined }`, which is dropped by `JSON.stringify` on the sync wire — the clear replays as a no-op on other devices. This is the defect class fixed in #9776 / #9778. Before merging, please rebase and adopt one of the patterns from `docs/sync-and-op-log/contributor-sync-model.md` → "Clearing a field — `undefined` does not survive the wire" (reducer-side clear, destructured rebuild, or the `clearedFields` prop from `src/app/util/cleared-update-fields.ts`).

PR #8494, opened 2026-06-19
feat(section): add tag filter to project sections

croakingtoad:
## What this does

Adds the ability to attach a tag filter to a project section. Any task carrying the selected tag(s) will automatically appear in that section, in addition to tasks placed there manually by drag-and-drop.

## Behaviour

- **Additive**: tag-matched tasks are appended after manually-placed tasks; drag-and-drop order is unaffected.
- **Multi-section**: a task whose tags match more than one section appears in all of them.
- **Unsectioned list**: tasks that don't match any section's filter (and aren't manually placed) fall through to the unsectioned list as before.
- **OR / AND**: when two or more tags are selected, a radio group lets the user choose between *"Show tasks matching any tag"* (OR, default) and *"Show tasks matching all tags"* (AND). The radio is hidden when only one tag is selected.

## Screenshot
image

## Files changed

| File | Change |
|------|--------|
| `section.model.ts` | `tagFilterIds?: string[]`, `tagFilterMode?: 'OR' \| 'AND'` |
| `section.service.ts` | Forward filter fields on `addSection` |
| `dialog-edit-section/` | **New** — title + chip-list-input tag picker + OR/AND radio |
| `work-view.component.ts` | `undoneTasksBySection` tag-filter pass; `editSection` opens new dialog |
| `work-view.component.html` | Pass full `section` object to `editSection` |
| `work-context-menu.component.ts` | `addSection` opens new dialog |
| `t.const.ts` + `en.json` | Three new `WW.*` translation keys |

## Implementation notes

- Reuses the existing `chip-list-input` + `<tag>` components — no new UI primitives.
- `selectAllTagsWithoutMyDay` is injected inside the dialog so no tag data needs to be passed through callers.
- The new dialog replaces `DialogPromptComponent` at both call sites (add + edit); the old `T.WW.ADD_SECTION_TITLE` placeholder key is reused as the title field label.


croakingtoad:
I realize you can already filter by tags in a project view, but this now allows you to be in the ungrouped view, and have sections that are more customizable.

image

Perhaps you want a section that contains tasks with two or more tags (e.g.: 'tag1 + tag2' or alternatively 'tag1 or tag2'). Now you can have it.

I'll probably be extending this more for myself in a worktree, so if you like the idea of a more customizable set of sections, I will send up what I end up doing. Also, as before, if you don't like it, just let me know and I won't send any more and just keep them for myself in a separate worktree. 

Cheers!

johannesjo:
Thanks for this — the implementation is clean (nice reuse of `chip-list-input`, signals + OnPush, optional fields), so this is purely a question of whether the feature should live as its own thing.

My main hesitation is that it overlaps a lot with the **Task View Customizer** we already have (`src/app/features/task-view-customizer/`). That already does **Filter by tag** and **Group by tag** — and "Group by tag" renders the list into named tag-derived buckets, which is essentially "sections that auto-populate by tag." It's also already persisted per work-context (in localStorage), so users have a saved, auto-grouping-by-tag experience today without touching synced state.

The genuinely new capability in this PR is **multi-tag AND/OR** (the customizer's tag filter is single-tag). That part I do like. My instinct is that it belongs *in the customizer* — extend the tag filter to take multiple tags + an OR/AND mode (the `FILTER_OPTION_TYPE.tag` branch in `task-view-customizer.service.ts` would go from `includes(tag.id)` to `some(...)`/`every(...)`). That's a small, contained change with an existing spec, and it avoids two costs I'd rather not take on here:

- **Synced data-shape lock-in:** `tagFilterIds`/`tagFilterMode` on `Section` flows through the section meta-reducer / op-log to every client, and data shapes are hard to walk back across synced devices once shipped.
- **Two meanings on one entity:** sections are currently manual drag-drop containers; adding a live tag filter makes them auto-queries too. That creates edge cases — a task can land in multiple sections, and a task that's only in a section *because* it matches the filter can't really be dragged out (it just snaps back). Keeping the tag grouping as a view concern sidesteps all of that.

So I'd lean toward declining the on-`Section` version, but I'd happily take the **multi-tag AND/OR** added to the View Customizer's tag filter if you're up for it — that's the bit that adds something we don't have. The one case the customizer *doesn't* cover is a few hand-built sections plus an auto-tag section all on screen at once; if that mixed layout is what you're actually after, let's talk through it first, because I'd want to model it as a distinct section type rather than overloading the existing one.

Also, minor housekeeping: the PR currently includes a few unrelated files (`SKILL.md`, the `AGENTS.md` link tweak, and a `.gitignore` `.worktrees/` line that's already on master) — looks like they slipped in from your local setup; could you drop those?

Really appreciate you offering these up either way — keep them coming. 🙏

croakingtoad:
Good feedback, I'll definitely handle those changes and submit back to you. Cheers!

croakingtoad:
Closing in favour of a targeted multi-tag AND/OR extension to the existing Task View Customizer, as suggested. Will open a new PR for that.

Issue #8348, opened 2026-06-12
SuperSync protocol: remove dead surface (isCleanSlate on /ops, GET /snapshot, GET /status, latestSnapshotSeq, dead config/types)


johannesjo:
**Area:** protocol/contract (server + shared-schema + client plumbing) · **Type:** simplification · **Value:** high (shrinks the high-risk surface; one item is an accepted-but-unused **data-wiping** flag)

All seven items verified — none has a production caller/reader:

1. **`isCleanSlate` on `POST /api/sync/ops`** — accepted by schema (`supersync-http-contract.ts:82`), forwarded (`sync.routes.ops-handler.ts:88,194`) into `sync.service.ts:120-153` which **deletes all user ops/devices and resets sync state**. The production client cannot even send it (`super-sync.ts:216-221`; `provider-types.ts:159-163` has no such param) — clean slate flows exclusively via `POST /snapshot`. An unused destructive flag on the hot endpoint: remove it from the /ops contract (or actively reject it) as defense-in-depth; keep the internal `uploadOps(..., isCleanSlate)` path used by the snapshot handler.
2. **`GET /api/sync/snapshot` and `GET /api/sync/status`** — zero production client callers (only server tests, one e2e helper, docs). GET /snapshot carries real CPU/cache complexity. Retire or explicitly re-scope as admin/diagnostic.
3. **`latestSnapshotSeq`** — computed and sent on every download (`sync.routes.ts:169`) but never read (`provider-types.ts:123` declaration only). Sibling `snapshotVectorClock` IS consumed — only this field is dead.
4. **`CONFLICT_STALE`** — `@deprecated`, never emitted (`sync.types.ts:52-53`); client compat branch + TODO at `rejected-ops-handler.service.ts:157`. Keep the client branch until old servers are gone; delete the server constant.
5. **Dead server config** — `maxOpsPerUpload`, `downloadLimit`, `downloadRateLimit` (`sync.types.ts:453-457, :476-480`): declared/defaulted, read nowhere (routes hardcode their own limits).
6. **Drifted/duplicate types** — `RestorePoint` defined twice (`sync.types.ts:359-365` unreferenced, incl. unused `DAILY_BOUNDARY`; live one in `snapshot.service.ts:25-31`); `UploadSnapshotRequest` (`sync.types.ts:327-333`) zero uses and missing 6 fields vs the zod schema; `SnapshotResponse` used once as a cast. Prefer `z.infer` types.
[...]

PR #8496, opened 2026-06-19
refactor: retire GET /api/sync/snapshot, re-scope GET /api/sync/status as diagnostic

forge34:
## Problem
Addresses #2 in #8348

## Solution
Discussed in https://github.com/super-productivity/super-productivity/issues/8348#issuecomment-4751770026 and https://github.com/super-productivity/super-productivity/issues/8348#issuecomment-4752148476

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/docs/long-term-plans/supersync-encryption-at-rest.md
+++ b/docs/long-term-plans/supersync-encryption-at-rest.md
@@ -2247,10 +2247,6 @@ curl -X POST http://localhost:1900/api/sync/ops \
 curl -X GET "http://localhost:1900/api/sync/ops?sinceSeq=0&limit=10" \
   -H "Authorization: Bearer $TOKEN"
 
-# Test snapshot generation
-curl -X GET http://localhost:1900/api/sync/snapshot \
-  -H "Authorization: Bearer $TOKEN"
-
 # Test restore point creation
 curl -X POST http://localhost:1900/api/sync/restore \
   -H "Authorization: Bearer $TOKEN"
--- a/docs/sync-and-op-log/diagrams/02-server-sync.md
+++ b/docs/sync-and-op-log/diagrams/02-server-sync.md
@@ -88,8 +88,7 @@ graph TB
             DownAPI["GET /api/sync/ops<br/>━━━━━━━━━━━━━━━<br/>Download operations<br/>Query: sinceSeq, limit"]:::api
             OpsAPI["POST /api/sync/ops<br/>━━━━━━━━━━━━━━━<br/>Upload operations<br/>Body: ops[], clientId"]:::api
             SnapshotAPI["POST /api/sync/snapshot<br/>━━━━━━━━━━━━━━━<br/>Upload full state<br/>Body: state, reason"]:::api
-            GetSnapshotAPI["GET /api/sync/snapshot<br/>━━━━━━━━━━━━━━━<br/>Get full state"]:::api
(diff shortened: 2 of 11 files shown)


johannesjo:
Thank you very much! <3

johannesjo:
## Review — looks good, ships safely; a couple of cleanups

I traced the call graph on both sides before/after the removal. The core claim holds: **no shipped client ever called `GET /api/sync/snapshot`.** The provider package only ever *POSTs* to `/api/sync/snapshot` (`uploadSnapshot`), and client gap-recovery re-downloads via `GET /api/sync/ops` with the piggybacked `snapshotState`/`snapshotVectorClock`. Import cleanup is correct too — `applyStorageUsageDelta` is still used by the POST handler, and `EncryptedOpsNotSupportedError` / `ENCRYPTED_OPS_CLIENT_MESSAGE` stay imported because the restore handler (`GET /restore/:seq`) still throws/catches them. Re-labelling `/status` as diagnostic rather than deleting it is the right conservative call. Net effect is a real attack-surface win (the route replayed up to 100k ops per call). 👍

Two things worth tidying before merge:

### 1. The architecture-doc rationale for keeping `generateSnapshot()` is inaccurate (medium)

`sync-server-architecture-diagrams.md` now says:

> The on-demand generation code (`generateSnapshot()`) is preserved internally for POST /snapshot cache management.

But `POST /snapshot` doesn't call `generateSnapshot()` — the upload handler manages the cache via `prepareSnapshotCache` + `cacheSnapshotIfReplayable` (siblings of `generateSnapshot`, not callers of it). After this PR, `SyncService.generateSnapshot()` → `SnapshotService.generateSnapshot()` → `SnapshotGenerationService.generateSnapshot()` (plus the FIX 1.7 concurrency lock, the `onCacheDelta` quota hook, and the B5 `maxCacheBytes` cap) has **zero production callers** — only the test suite still calls it, as a "replay ops → assert state" oracle.

Keeping the code is fine (the tests rely on it and it shares the replay engine with `generateSnapshotAtSeq`, which *is* live via restore). The problem is just the stated reason: a future maintainer will look for POST /snapshot's use of `generateSnapshot()`, not find it, and either be misled or "clean up" the now-callerless method and break ~40 test call sites. Suggest rewording to something like *"retained because the test suite uses it as a state-replay oracle and it shares the replay engine with `generateSnapshotAtSeq()` (restore)."*

### 2. `SnapshotResponse` is now an orphaned type (low)

The PR removed both importers of `SnapshotResponse`, leaving `src/sync/sync.types.ts` `export interface SnapshotResponse` with no references repo-wide (and it isn't re-exported from any package index, so deletion is safe). Since the PR's whole point is cruft/attack-surface reduction, it's the natural thing to drop too.

### Test rework — fine

The integration/isolation reworks read well; the multi-user isolation test moving from snapshot-state diffing to per-user `GET /ops` assertions is actually a stronger check. One minor note: `sync-fixes.spec.ts` drops the HTTP-level "GET /snapshot → 400 ENCRYPTED_OPS_NOT_SUPPORTED" assertion — still covered at the service layer, so acceptable.

(Note: I verified statically — I didn't run the server suite here, no vitest in my checkout and the integration specs need a live Postgres.)

PR #8497, opened 2026-06-19
docs(sync): clarify upload rate-limit layers

### LINKED ISSUE #8350: SuperSync server: consolidate dual rate-limiting on POST /api/sync/ops
**Area:** server · **Type:** simplification · **Value:** medium (one rate-limit system instead of two with identical thresholds)

## Problem
`POST /api/sync/ops` is rate-limited twice with identical thresholds: the Fastify rate-limit plugin per-IP (`sync.routes.ts:86-91`, max 100/min, default per-IP key) **and** the custom per-user `RateLimitService` invoked inside the handler (`sync.routes.ops-handler.ts:98` → `rate-limit.service.ts`; `DEFAULT_SYNC_CONFIG.uploadRateLimit = {max: 100, windowMs: 1min}`, `sync.types.ts:479`). Both return 429; they differ only in JSON body (plugin default vs `{errorCode: 'RATE_LIMITED'}`).

## Consolidation constraints (verified — not a drop-in deletion)
1. Per-user keying needs the authenticated userId, but the plugin keys in its `onRequest` hook **before** the `preHandler` authenticate hook (`sync.routes.ts:72`) — requires registering with the `hook: 'preHandler'` option (or token parsing in `keyGenerator`; the per-(ip,clientId) WS pattern at `websocket.routes.ts` shows custom keying works).
2. The plugin is entirely disabled in testMode (`server.ts:221-226`), leaving the custom service as the only limiter there.
3. `clearForUser` on account deletion (`sync.service.ts:277, 634`) and per-deploy config override (`uploadRateLimit` is injectable `SyncConfig`) have no plugin equivalent.
4. No client code branches on `errorCode: RATE_LIMITED` (zero matches in `src/app/op-log` / `packages/sync-providers`); clients treat any 429 as retryable (`retryable-upload-error.ts:46`) and parse the plugin's "retry in N minutes" message (`super-sync.ts:557-589`) — which the custom body lacks. So converging on the plugin's error shape actually *improves* client retry hints.

## Proposed fix
Key the plugin per-user via `hook: 'preHandler'` registration after auth, keep the `RATE_LIMITED`-style body via `errorResponseBuilder`, account for testMode, then delete `rate-limit.service.ts` (+ its cleanup hook in `cleanup.ts:58` and `clearForUser` call sites) — or, if the per-deploy override matters, keep the custom service and drop the plugin entry for /ops instead. Either way: one system, one error surface.

symonbaikov:
## Problem

Fixes #8350.

`POST /api/sync/ops` has two upload rate-limit layers that can look redundant because both use a `100/min` threshold:

- route-level Fastify rate limit, keyed by IP
- custom `SyncService.isRateLimited(userId)`, keyed by authenticated user

The reviewer feedback clarified that these are intentionally different layers, not a drop-in duplicate: the Fastify limiter is a pre-auth flood backstop, while the custom limiter is post-auth per-user fairness.

## Solution

Keep both existing limiters and make their intent explicit in code comments:

- route-level Fastify `config.rateLimit` remains on `POST /api/sync/ops` as the shared per-IP backstop before auth/DB work
- `RateLimitService` remains inside the upload handler as the authenticated per-user fairness limiter and keeps the `RATE_LIMITED` response contract

The route test now documents both layers: the custom limiter stays user-scoped under the Fastify plugin setup, and the route-level limiter still acts as a shared per-IP backstop.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [x] Other: regression test coverage

## Testing

- [x] RED check: upload route spec failed before restoring the route-level limiter (`expected 200 to be 429`)
- [x] `npm run checkFile packages/super-sync-server/src/sync/sync.routes.ts`
- [x] `npm run checkFile packages/super-sync-server/src/sync/sync.routes.ops-handler.ts`
- [x] `npm run checkFile packages/super-sync-server/tests/sync-upload-rate-limit.routes.spec.ts`
- [x] `cd packages/super-sync-server && npx vitest run tests/sync-upload-rate-limit.routes.spec.ts tests/rate-limit.service.spec.ts tests/cleanup.spec.ts`
- [x] `cd packages/super-sync-server && npm run build`
- [x] `git diff --check`
- [x] Pre-push hook passed, including lint, package tests, release-notes tests, Angular/Karma suite, and timezone suite

### DIFF
--- a/packages/super-sync-server/src/sync/sync.routes.ops-handler.ts
+++ b/packages/super-sync-server/src/sync/sync.routes.ops-handler.ts
@@ -92,8 +92,8 @@ export const uploadOpsHandler = async (
       `[user:${userId}] Upload: ${ops.length} ops from client ${clientId.slice(0, 8)}...`,
     );
 
-    // Rate limit check BEFORE deduplication to prevent bypass
-    // (attacker could retry with same requestId to skip rate limiting)
+    // Post-auth per-user fairness limit, separate from the route-level per-IP
+    // backstop. Check BEFORE deduplication so requestId retries cannot bypass it.
     if (syncService.isRateLimited(userId)) {
       Logger.audit({
         event: 'RATE_LIMITED',
--- a/packages/super-sync-server/src/sync/sync.routes.ts
+++ b/packages/super-sync-server/src/sync/sync.routes.ts
@@ -72,6 +72,8 @@ export const syncRoutes = async (fastify: FastifyInstance): Promise<void> => {
   fastify.addHook('preHandler', authenticate);
 
   // POST /api/sync/ops - Upload operations
+  // Route-level limiting is a pre-auth per-IP backstop for upload floods before
+  // auth/DB work. uploadOpsHandler applies the separate per-user fairness limit.
   fastify.post<{ Body: UploadOpsRequest }>(
     '/ops',
     {
--- a/packages/super-sync-server/tests/sync-upload-rate-limit.routes.spec.ts
+++ b/packages/super-sync-server/tests/sync-upload-rate-limit.routes.spec.ts
@@ -1,0 +1,156 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import Fastify, { FastifyInstance } from 'fastify';
+import rateLimit from '@fastify/rate-limit';
+
+const mocks = vi.hoisted(() => {
+  const uploadCountsByUser = new Map<number, number>();
+  const syncService = {
+    isRateLimited: vi.fn(),
+    checkOpsRequestDedup: vi.fn(),
+    cacheOpsRequestResults: vi.fn(),
+    checkStorageQuota: vi.fn(),
+    uploadOps: vi.fn(),
+    runWithStorageUsageLock: vi.fn(),
+    getLatestSeq: vi.fn(),
+    getOpsSinceWithSeq: vi.fn(),
+  };
+
+  return {
+    syncService,
+    uploadCountsByUser,
+    notifyNewOps: vi.fn(),
+    verifyToken: vi.fn(),
+  };
+});
+
+vi.mock('../src/auth', () => ({
+  verifyToken: mocks.verifyToken,
+}));
+
+vi.mock('../src/sync/sync.service', () => ({
+  getSyncService: () => mocks.syncService,
+}));
+
+vi.mock('../src/sync/services/websocket-connection.service', () => ({
+  getWsConnectionService: () => ({
+    notifyNewOps: mocks.notifyNewOps,
+  }),
+}));
+
+import { syncRoutes } from '../src/sync/sync.routes';
+
+const CUSTOM_UPLOAD_LIMIT = 3;
+const ROUTE_UPLOAD_LIMIT = 100;
+
+const authTokenForUser = (userId: number): string => `user-${userId}-token`;
+
+const createOp = (clientId: string, entityId: string) => ({
+  id: `op-${entityId}`,
+  clientId,
+  actionType: 'ADD_TASK',
+  opType: 'CRT',
+  entityType: 'TASK',
+  entityId,
+  payload: { title: 'Test Task' },
+  vectorClock: {},
+  timestamp: Date.now(),
+  schemaVersion: 1,
+});
+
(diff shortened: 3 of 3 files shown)


johannesjo:
Thanks for this, and for the genuinely careful analysis in #8350 — the double-100/min reads as redundant, so I understand the instinct.

I want to be straight with you rather than have you keep iterating on it: **I don't plan to merge this as-is, barring an argument we haven't considered.** The core reason:

Removing the route-level `config.rateLimit` doesn't leave POST /ops with one limiter. `@fastify/rate-limit` is registered globally and `global` defaults to `true`, so a route with no `config.rateLimit` falls back to the **global** limiter rather than having none (`index.js` `onRoute` → `else if (globalParams.global)`; the plugin is registered at `server.ts:222`, before `syncRoutes` at `:301`). So after this PR, POST /ops is still gated by two independent systems — the global plugin (500/15min, per-IP, in `onRequest` before auth) and the custom `RateLimitService` (100/min, per-user, in the handler). The PR removes the duplicated *threshold*, but not the two-system architecture the issue is actually about.

And it isn't free. Removing the route config moves POST /ops out of its own dedicated per-IP bucket and into the shared global one — which is meaningfully tighter on sustained load for multiple devices behind one IP (a per-IP ceiling of 500/15min vs. the old dedicated 100/min), and now shares that budget with the other un-gated routes. It also changes which 429 a rate-limited client sees: today the route plugin trips first (it runs in `onRequest`, before auth) and returns a body the client parses for a `"retry in N"` hint (`super-sync.ts` `_extractRetryDelayReason`); for a single user hitting their own 100/min, the binding 429 becomes the custom limiter's `{ error: 'Rate limited' }`, which has no such hint. It's cosmetic (the client's retry timing doesn't use it), but it's still a behavior change with no upside.

So for a cleanup with no user-reported throttling, this spends real behavior change to reach a consolidation that doesn't actually complete — a trade I don't think is worth it.

What I'd happily take instead: leave both limiters and make the intent legible — a short comment documenting them as deliberate, separate layers (route-level = pre-auth per-IP backstop that stops floods before they hit auth/DB; `RateLimitService` = post-auth per-user fairness). That answers "why are there two?" durably so it doesn't get re-flagged, with zero behavior change. **Your regression test is worth keeping** under that framing — it proves the per-user limiter fires independently — so this needn't throw away your work, and I'm glad to help reshape the PR that way.

The door I'd genuinely leave open: if you see a path that actually reaches one system — e.g. re-keying the plugin per-user via `hook: 'preHandler'` + an `errorResponseBuilder` that keeps the `RATE_LIMITED` shape, then deleting the custom service (modulo the testMode / `clearForUser` / per-deploy-override constraints you already flagged in the issue) — or you want to argue the route-level IP limiter is worth removing on its own merits (it *is* the more awkward of the two: redundant-looking and NAT-shared pre-auth), make that case and I'll weigh it seriously. But absent that, I'd rather redirect this toward documentation than have either of us spend more time on the deletion.

symonbaikov:
@johannesjo Thanks, that makes sense.
I agree that my PR only removed the duplicated 100/min route threshold, but did not actually collapse the two-system setup because the global Fastify limiter still remains in front of the custom per-user limiter. Your point about the behavior change is fair too, especially moving /ops into the shared global bucket and changing which 429 shape clients see.

I’ll rework this toward the lower-risk option you suggested: keep both limiters, make their intent explicit in the route/server code comments, and keep/adapt the regression test so it documents that the custom per-user limiter still works independently from the pre-auth Fastify backstop.

I’ll avoid pushing further on deletion unless there’s a cleaner one-system approach that preserves the per-user semantics, test-mode behavior, and client retry hints.

symonbaikov:
Reworked this in b2f0a18e4.

The PR now keeps the route-level `POST /ops` Fastify `config.rateLimit` and documents it as the pre-auth per-IP backstop, while the handler comment documents `RateLimitService` as the separate post-auth per-user fairness layer.

I also adjusted the route test under that framing: it now verifies the custom limiter stays authenticated-user scoped below the Fastify bucket, and that the route-level per-IP backstop is still present. No runtime behavior change from `upstream/master` is intended now.

johannesjo:
Thank you very much! <3

Issue #8360, opened 2026-06-12
Client sync: misc verified cleanups (O(n²) filters, dead guards, deleteOpsWhere, timer leak, divergent empty-state checks)


johannesjo:
**Area:** client sync · **Type:** small cleanups, batched · **Value:** individually low, collectively worthwhile. All verified.

*Trimmed to avoid overlap: the `planDownloadFullStateUpload` ×3 dedup is covered in #8324; the duplicated SYNC_IMPORT conflict-gate extraction is covered in #8320.*

1. **Defensive null-provider guards on non-nullable params + one-line delegation** (`operation-log-upload.service.ts:63-73`, `operation-log-download.service.ts:69-81`) — leftovers from the dual-path era; inline.
2. **O(n²) hot spots**: `regularOps.filter(e => !syncOpPairs.some(...))` over the full pending set (`operation-log-upload.service.ts:230-232` — derivable as "entries whose `_entryToSyncOp` returned null"); `entries.find` per result row (`:276-282`, bounded per 25-op batch — use a Map); `localOpsToReject.includes` in a nested loop (`conflict-resolution.service.ts:354-367` — use a Set).
3. **`deleteOpsWhere(() => true)`** (`operation-log-migration.service.ts:114`) decode-and-cursor-deletes every entry where `clearAllOperations()` (`operation-log-store.service.ts:1315-1322`, O(1) `store.clear()` + the same three cache invalidations) is verified-equivalent on this path.
4. **`Object.values(ActionType).includes(action.type)` per persisted write** (`operation-log.effects.ts:154`) — build a module-level `Set` once.
5. **Clock-drift timer leak** (`operation-log-download.service.ts:473`) — `setTimeout` handle overwritten without `clearTimeout`; stray timer only (`hasWarnedClockDrift` prevents double-warn).
6. **Two divergent "is local state empty" definitions**: `ServerMigrationService._isEmptyState` (`server-migration.service.ts:296-314`) counts the default INBOX project as data and ignores notes; `hasMeaningfulStateData` (`has-meaningful-state-data.util.ts:37-68`, used at `operation-log-sync.service.ts:506` etc.) excludes INBOX/system tags and counts notes. Delegate `_isEmptyState` to `!hasMeaningfulStateData` — after verifying migration callers tolerate INBOX-only being "empty".

*Found during a deep SuperSync code review; independently re-verified against the code.*

PR #8499, opened 2026-06-19
fix(sync): clean up verified op-log paths

symonbaikov:
## Summary

Fixes the verified client-sync cleanup items from #8360 across the op-log sync path.

## Changes

- Optimized upload filtering by splitting local-only and uploadable ops in one pass.
- Mapped accepted upload results through an `opId -> seq` map instead of per-result entry scans.
- Replaced full orphan-op cleanup with `clearAllOperations()` in the migration service.
- Cached known `ActionType` values in a module-level set for persistence validation.
- Fixed clock-drift retry cleanup without allowing repeated drift checks to postpone the warning forever.
- Kept conflict rejection order stable while using a companion set for duplicate checks.
- Replaced duplicated server-migration empty-state checks with shared meaningful-state detection plus migration-specific default-state safeguards.
- Added focused regression coverage for local-only uploads, out-of-order accepted upload results, orphan cleanup, clock-drift timer behavior, conflict rejection dedupe, and server-migration empty-state handling.

## Verification

- `npm run test:file src/app/op-log/sync/operation-log-upload.service.spec.ts`
- `npm run test:file src/app/op-log/persistence/operation-log-migration.service.spec.ts`
- `npm run test:file src/app/op-log/capture/operation-log.effects.spec.ts`
- `npm run test:file src/app/op-log/sync/operation-log-download.service.spec.ts`
- `npm run test:file src/app/op-log/sync/conflict-resolution.service.spec.ts`
- `npm run test:file src/app/op-log/sync/server-migration.service.spec.ts`
- `npm run checkFile` for all modified `.ts` files
- pre-push hook: `npm run lint`
- pre-push hook: `npm test` including package tests, Angular suite, and timezone suite

### DIFF
--- a/src/app/op-log/capture/operation-log.effects.ts
+++ b/src/app/op-log/capture/operation-log.effects.ts
@@ -36,6 +36,8 @@ interface WriteOperationOptions {
   callerHoldsOperationLogLock?: boolean;
 }
 
+const KNOWN_ACTION_TYPES: ReadonlySet<string> = new Set(Object.values(ActionType));
+
 /**
  * NgRx Effects for persisting application state changes as operations to the
  * `OperationLogStoreService`. It listens for specific NgRx actions marked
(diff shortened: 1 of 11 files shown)


johannesjo:
Thank you very much! <3

PR #8500, opened 2026-06-19
feat(task-view-customizer): extend tag filter to multi-tag AND/OR

croakingtoad:
## What this does

Extends the existing **Filter → Tag** option in the Task View Customizer (the panel behind the [C] toggle) from a single-tag text search to a **chip-list multi-select** with an **OR / AND mode toggle**.

Previously: user typed a tag name; the service did a title-substring match for one tag.
Now: user picks any number of tags from a searchable chip-list; two or more tags reveal an **"Any tag (OR)"** / **"All tags (AND)"** toggle that controls the match logic.

The NOT_SPECIFIED ("no tags") option is preserved unchanged.

## Why this approach

Per maintainer feedback on #8494, the new capability (multi-tag AND/OR) belongs in the View Customizer rather than on the Section entity because:
- State lives in localStorage per context — no synced data-shape changes
- It's a view concern, not a structural one — no edge cases around tasks snapping back
- An existing spec already covers the tag filter; the new tests extend it naturally

## Files changed

| File | Change |
|------|--------|
| `types.ts` | Add optional `tagIds?: string[]` and `tagFilterMode?: 'OR' \| 'AND'` to `FilterOption` |
| `task-view-customizer.service.ts` | Multi-tag branch in `customizeUndoneTasks`; `_sanitizeFilter` preserves `tagIds`/`tagFilterMode` across context switches |
| `task-view-customizer-panel.component.ts` | Inject `TagService`; add `addTagFilter`, `removeTagFilter`, `setTagFilterMode`; import `ChipListInputComponent` |
| `task-view-customizer-panel.component.html` | Swap text input → `chip-list-input` + OR/AND buttons (hidden until ≥2 tags) |
| `task-view-customizer.service.spec.ts` | 4 new tests: OR, AND, single-tag, empty-tagIds |
| `t.const.ts` + `en.json` | `TAG_FILTER_OR` ("Any tag (OR)") and `TAG_FILTER_AND` ("All tags (AND)") |

## Backward compatibility

- `applyFilter(tasks, type, value)` signature unchanged — all existing specs pass unmodified
- Old stored `{ type: 'tag', preset: 'Tag Name' }` state continues to work via the existing text-match path
- `tagIds: []` is treated as no filter (falls through to the uncustomized list)


croakingtoad:
Closing to review before resubmitting.

PR #8502, opened 2026-06-19
feat(task-view-customizer): extend tag filter to multi-tag AND/OR

croakingtoad:
## What changed

- **UI**: the single-tag text input in the filter panel is replaced with a `chip-list-input` (already in the codebase). When a second tag is added, two toggle buttons appear — *"Any tag (OR)"* and *"All tags (AND)"*.
- **Service**: when `tagIds` is set, the new multi-tag branch runs `some()`/`every()` instead of the old `applyFilter` single-tag path. Single-tag behaviour is unchanged (still goes through the same `some()` shortcut, so it's equivalent). `_sanitizeFilter` is extended to preserve `tagIds` and `tagFilterMode` across context switches and page reloads.
- **State**: filter state stays in localStorage as before — no synced-state changes, no new entities.
- **Tests**: 4 new specs in the existing `task-view-customizer.service.spec.ts` — OR match, AND match, single-tag shortcut, empty-`tagIds` passthrough. All 20+ existing specs pass untouched.

## What this doesn't change

`applyFilter`'s signature is unchanged. The existing single-tag filter path for other filter types is untouched.

## Screenshot
image

## Files

| File | Change |
|------|--------|
| `types.ts` | `tagIds?: string[]` and `tagFilterMode?: 'OR' \| 'AND'` on `FilterOption` |
| `task-view-customizer.service.ts` | Multi-tag filter branch; `_sanitizeFilter` preserves new fields |
| `task-view-customizer-panel.component.ts` | `chip-list-input` swap; add/remove/mode methods |
| `task-view-customizer-panel.component.html` | Chip input + OR/AND toggle (shown at ≥2 tags) |
| `task-view-customizer.service.spec.ts` | 4 new tests |
| `t.const.ts` + `en.json` | Two translation keys |


johannesjo:
Thanks for this — nicely scoped and it reuses `chip-list-input` instead of inventing new UI. I reviewed the diff, ran the spec (52/52 green, including the 4 new ones), and ran lint/format on the changed `.ts` files (clean). Overall this looks mergeable; a few notes below, only the first is really worth acting on.

### Worth fixing

**1. The virtual `TODAY` tag leaks into the suggestions.** The panel feeds `tagService.tagsInTreeOrder`, which is derived from `selectAllTags` and always includes the virtual `TODAY` tag (the tag reducer guarantees it's present). Selecting "Today" as a filter chip matches **zero** tasks, because no task ever carries `'TODAY'` in `task.tagIds` (it's a virtual tag — see ARCHITECTURE-DECISIONS #2). Note the service itself already strips it in `_tagsInSidebarOrder()` for exactly this reason. Suggest feeding `tagsNoMyDayAndNoListInTreeOrder` (or otherwise filtering out `TODAY_TAG.id`) into `[suggestions]`.

### Minor / optional

**2. Behavior change worth a mention.** The old tag filter was a substring *title* match; the new one is exact selection by tag *id*. That's an improvement, just not equivalent — partial-title matching is gone. Fine to accept, worth being deliberate about.

**3. Dead translation key.** After the swap, `ENTER_TAG` is no longer referenced by any template but still lives in `t.const.ts` + `en.json`. Cheap to remove.

**4. Lost optional-chaining in the filter branch.** `t.tagIds.includes(id)` vs. the legacy `task.tagIds?.includes(...)`. `tagIds` is effectively always defined on the model, so it's safe — just flagging the small robustness reduction.

**5. Deleted-tag ids become invisible.** If a stored `tagId` no longer exists, `chip-list-input` drops it from the rendered chips, yet the id still participates in the filter — so the user can't see or remove it. Edge case, acceptable.

### Why I'm comfortable with the approach

Filter state is `localStorage`-only (not in the op-log / synced state), and the two new fields are optional, so there's no sync-format or cross-client risk and old/new state stays compatible in both directions. The single-tag path still resolves through `some()` (equivalent to before), and the `NOT_SPECIFIED` / empty-`tagIds` paths are preserved and covered by tests. Not a hot path either.

👍 from me once **#1** is addressed (and optionally #3).

PR #8503, opened 2026-06-19
fix(backup): preserve section during legacy migration

symonbaikov:
## Summary
- derive legacy backup default/valid model keys from MODEL_CONFIGS instead of a hand-maintained v17 key list
- ensure legacy-detected backups that omit section receive the section default before validation
- preserve existing section data while stripping obsolete legacy keys

## Tests
- npm run checkFile src/app/op-log/backup/migrate-legacy-backup.ts
- npm run checkFile src/app/op-log/backup/migrate-legacy-backup.spec.ts
- npm run test:file src/app/op-log/backup/migrate-legacy-backup.spec.ts
- git diff --check
- push hook: npm run lint
- push hook: npm test


johannesjo:
Thank you very much! <3

PR #8466, opened 2026-06-18
feat(issue): add Basecamp (To-dos) as a built-in issue provider

oonid:
## Problem

Closes #8465.

Super Productivity has no Basecamp integration, so teams who track work as
Basecamp **To-dos** must re-enter tasks by hand and keep them in sync manually.
A to-do completed in SP stays open in Basecamp (and vice-versa), so neither
side stays trustworthy.

## Solution

Add Basecamp as a first-class issue provider, reusing the existing provider
architecture (`IssueServiceInterface` / `ISSUE_SERVICE_MAP` / `IssueSyncAdapter`)
so it behaves like the other integrations:

- **OAuth login**: a "Connect Basecamp" button runs the 37signals launchpad
    flow (PKCE + loopback redirect) from the Electron app; no long-lived token
    pasting. Access tokens auto-refresh on `401` (single retry, no loop).
- **Account discovery**: lists the user's BC3 accounts after connect and
    auto-selects when there is only one.
- **Config from URL**: paste a Basecamp to-do-list URL (`3.basecamp.com` /
    `app.basecamp.com`) to fill the bucket / to-do-list ids; mismatched accounts
    are rejected without clobbering existing values.
- **Import + poll**: import to-dos from a chosen to-do list as SP tasks and
    keep them refreshed.
- **Two-way done-sync**: completing a task in SP completes the Basecamp to-do
    and vice-versa, via the existing two-way-sync adapter.

A manual access-token fallback is kept for non-Electron contexts. Basecamp
native **timesheet / time** push is intentionally out of scope (planned as a
separate change).

Implementation notes:
- Basecamp API calls use account-scoped paths and `.json` endpoints to avoid a
    `406` from Electron's content-negotiation; the loopback OAuth IPC reuses the
    existing `electron/plugin-oauth.ts`.
- For installed/native apps the OAuth "secret" is non-confidential (PKCE). The
    build ships `YOUR_BASECAMP_CLIENT_ID` / `YOUR_BASECAMP_CLIENT_SECRET`
    placeholders so a maintainer/self-hoster can register an SP-owned Basecamp
    app, **happy to align on app ownership before merge.**
- New UI strings go through `T` / `en.json` only; one new monochrome
    (`currentColor`) provider icon.

45 files changed, ~+2960/−28. Organized into 3 logical commits: provider
  registration, import + done-sync, OAuth/discovery/refresh.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
[...]

PR #8507, opened 2026-06-20
feat(issue): add Basecamp as an issue-provider plugin

### LINKED ISSUE #8465: 💡 Basecamp (To-dos) issue provider: community plugin + generic OAuth host hooks
### Problem Statement

Super Productivity supports several issue providers (Jira, GitHub, GitLab, Gitea, Redmine, OpenProject, CalDAV, Trello, Nextcloud Deck), but Basecamp, a widely used project/team tool, isn't one of them. Teams who track their work as Basecamp To-dos currently have no way to pull those into SP. They must re-enter tasks by hand, which is duplicate work and drifts out of sync: a to-do completed in SP stays open in Basecamp (and vice-versa), so neither side is trustworthy. Time spent in SP also can't be reflected back against the originating Basecamp work.

### Possible Solution

Add Basecamp as a first-class issue provider using the existing provider architecture (IssueServiceInterface / ISSUE_SERVICE_MAP / two-way-sync adapter), so it behaves like the other integrations:

- Connect via OAuth — a "Connect Basecamp" button runs the 37signals launchpad OAuth flow (PKCE + loopback redirect) from the Electron app, so users don't paste long-lived tokens. Access tokens auto-refresh on 401.
- Account discovery — after connecting, the user's BC3 accounts are listed and auto-selected when there's only one.
- Config from URL — paste a Basecamp to-do-list URL (3.basecamp.com / app.basecamp.com) to fill in the bucket/to-do-list IDs.
- Import + poll — import to-dos from a chosen to-do list as SP tasks and keep them refreshed.
- Two-way done-sync — completing a task in SP completes the Basecamp to-do, and vice-versa, via the existing IssueSyncAdapter.

A manual access-token fallback is kept for non-Electron contexts. (Native Basecamp timesheet/time push is intentionally scoped out as a separate follow-up.)

I have a working implementation of the above and would like to open a PR for it.

- A plugin instead of a built-in provider — SP's plugin API doesn't currently expose the issue-provider/two-way-sync hooks, so a plugin couldn't deliver real done-sync; a built-in provider reuses the proven sync machinery.
- Pasted-token-only setup (like the early MVP) — works, but long-lived tokens are poor UX and a security risk; OAuth with auto-refresh is the better default, with the token field kept only as a fallback.
- One-way import only, simpler, but leaves the two systems permanently out of sync, which defeats the point.
- Doing nothing / manual re-entry — the status quo; duplicate effort and drift.

### Additional context

- Built on SP's existing issue-provider + two-way-sync infrastructure (no new sync paradigm); adds one icon and a set of i18n strings (en.json only).
- Basecamp API specifics handled: account-scoped paths, .json endpoints (avoids a 406 from Electron's content-negotiation), and the loopback OAuth IPC already present for plugin OAuth.
- For installed/native apps the OAuth "secret" is non-confidential (PKCE); the build ships client_id/client_secret placeholders so a maintainer/self-hoster can register an SP-owned Basecamp app.
- PR to follow referencing this issue. Happy to adjust scope/approach based on maintainer feedback before review.

oonid:
## Problem
  
Supersedes #8466, which implemented Basecamp as a built-in provider and was
declined, new issue providers should ship as plugins. Closes #8465.
  
## Solution
  
Two commits:
  
1. **`feat(plugins)` — generic plugin-framework enhancements** so OAuth-based
   issue-provider plugins are possible without built-in code:
   - `OAuthFlowConfig.redirectUri` for providers/user apps needing an exact
     pre-registered callback;
   - plugin OAuth host/bridge honor the redirect override and user-supplied
     (bring-your-own) OAuth client credentials;
   - generic issue-provider config-dialog helpers (dynamic `loadOptions`,
     OAuth config overrides, config model-merge); 
   - plugin issue-provider icon resolution falls back to the plugin's own icon.
     
2. **`feat(basecamp)` — the Basecamp plugin** (`packages/plugin-dev/basecamp-issue-provider`,
   bundled), mirroring the `clickup`/`linear` plugin structure: Launchpad OAuth
   connect → account discovery → project/todolist pickers → to-do import →
   two-way done-sync.
 
OAuth client metadata is non-confidential public/native-client data (RFC 8252,
like the Google Calendar plugin); it's injected at build from env with a 
placeholder fallback, and users can supply their own Basecamp app via advanced
config. User tokens stay in the host's local OAuth store (never synced).
  
## Type of Change
  
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other
  
## Checklist
  

johannesjo:
Thanks for reworking this into a plugin per the earlier feedback — the structure mirrors the existing provider plugins well and the test coverage is genuinely thorough. 🙏

Before going further on the bundled approach, I want to flag a structural concern and propose a shape I think fits better.

**Basecamp is uniquely costly to bundle.** Unlike GitHub/ClickUp/Linear/Gitea — which authenticate with a user-pasted token and touch no shared infrastructure — Basecamp 3's API is OAuth-only (no personal access tokens), and Launchpad has no PKCE: the `web_server` flow needs `client_id` + `client_secret` for both the token exchange and the 2-week refresh. So bundling Basecamp means **Super Productivity has to register and own a Basecamp OAuth app and ship its `client_secret` in every build** — effectively public, and not rotatable without breaking installed clients. (The "non-confidential / PKCE (RFC 8252)" note in the code isn't accurate for Launchpad — worth correcting regardless.) For a niche provider, that standing liability plus Basecamp-specific code in core isn't a trade I want to make.

**Proposal: ship this as a community plugin rather than bundling it.** You already built the advanced "bring your own OAuth app" fields (`clientId`/`clientSecret`/`redirectUri`) — that's exactly the right mechanism: each user registers their own Launchpad app and supplies their own credentials, so SP never owns or ships a secret. Plugins install from a ZIP (see `docs/plugin-development.md`), and issue-provider registration + OAuth already work for non-bundled plugins. The plugin would live in your own repo and be installed on demand.

**What that means for this PR:**
- Drop from core: the bundled plugin + `plugin.service.ts` registration, `assets/icons/basecamp.svg` + the global-theme entry, the icon-precedence flip in `plugin-bridge.service.ts`, and the Basecamp-specific cascade logic (`accountId`/`bucketId`/`todolistId` hardcoded in the shared dialog).
- The only things Basecamp genuinely needs from core are the small *generic* OAuth hooks so an exact-redirect plugin can work at all: `OAuthFlowConfig.redirectUri` + the Electron loopback honoring a requested port. I'd consider those on their own merits via a separate, trimmed PR if kept minimal and provider-agnostic (the `updateModel` pluginConfig-merge also looks like a legit standalone bugfix). If we want the cascading-dropdown reset to be generic, it should key off the existing `showIf` dependency rather than hardcoded field names.

A few notes that apply to the plugin either way:
- `getRequiredConfigValue` throws `ERRORS.MISSING_<KEY>`, but `en.json` only defines `NOT_AUTHENTICATED` / `CONFIG_INCOMPLETE`, so users would see raw keys.
- The loopback redirect is a fixed port (`8976`); if it's already in use, connect fails with no fallback — worth a clear user-facing error.
- After the `updateModel` refactor, the `try` and `catch` branches run identical code, so the `catch` is now dead.
- `searchIssues` re-fetches up to 20 pages on every call; a short-lived cache would help.

Really appreciate the work here — I think the community-plugin route gets your integration to the people who want it without putting a Basecamp-owned secret and provider-specific code into the core app. Happy to take the small generic OAuth host hooks via a separate, trimmed PR.

oonid:
Thanks @johannesjo, that direction makes sense. 
I've split the small generic OAuth host
hooks into their own PR (#8546): `OAuthFlowConfig.redirectUri`, the
loopback fixed-port support + clear in-use error, the bring-your-own-app
credential override, and the `pluginConfig` partial-merge fix, 
all provider-agnostic, no Basecamp code. Basecamp itself will ship as a community
plugin (its own repo, install-from-ZIP, with the BYO-app

johannesjo:
Thanks again for reworking this and — especially — for splitting the generic, provider-agnostic OAuth hooks out into #8546. That's exactly the right shape. 🙏

To put the decision on the record: Basecamp 3's Launchpad API is OAuth-only with a `web_server` flow (no personal access tokens, no PKCE), so a `client_secret` is required for both the token exchange and the 2-week refresh. Bundling Basecamp into core would mean Super Productivity registering and shipping a Basecamp `client_secret` in every build — effectively public, and not rotatable without breaking already-installed clients. For a niche provider, that standing liability plus Basecamp-specific code in core isn't a trade I want to make.

The path we landed on:

- The small, generic host hooks — `OAuthFlowConfig.redirectUri`, the Electron loopback fixed-port + clear in-use error, bring-your-own OAuth credentials, and the `pluginConfig` partial-merge fix — live in #8546.
- Basecamp ships as a **community plugin** in your own repo (install-from-ZIP), where each user registers their own Launchpad app and supplies their own credentials, so SP never owns or ships a secret.

A few small things worth carrying into the community plugin (all minor):

- `getRequiredConfigValue` throws `ERRORS.MISSING_<KEY>`, but `i18n/en.json` only defines `NOT_AUTHENTICATED` / `CONFIG_INCOMPLETE` — so the used keys are undefined and the defined `CONFIG_INCOMPLETE` is unused. Users would see raw keys.
- `loadBasecampAccounts` / `loadBasecampTodolists` dereference the response (`data.accounts`, `project.dock`) without the `|| []` guard that `loadBasecampProjects` has — they'll throw if `http.get` resolves to null.
- `searchIssues` re-paginates the whole todolist (up to 20 pages) on every call; a short-lived cache would help.
- The "NOT A SECRET … RFC 8252 / PKCE" note in `plugin.ts` isn't accurate for Launchpad's `web_server` flow — worth correcting in the plugin source.

Closing this in favor of #8546 + the community-plugin route. Really appreciate how responsive you've been to the feedback — thank you for the work here.

PR #8511, opened 2026-06-20
feat: add configurable bind address for local REST API

Rahulsharma0810:
## Description

HomeLab and power users often run LLMs on headless servers while keeping the Super Productivity desktop app on their laptop. Previously the local REST API was hard-bound to `127.0.0.1`, making it unreachable from other machines on the LAN.

This PR introduces a new config option `localRestApiBindAddress` in Misc settings (Electron only) with two choices:
- **`127.0.0.1`** (default, existing behavior) — localhost only
- **`0.0.0.0`** — all interfaces, allowing LAN access

When set to `0.0.0.0`, scripts and tools on other machines can reach the API via the machine's LAN IP (e.g. `http://192.168.1.50:3876/tasks`). The existing Host-header and Origin checks remain active for CSRF/DNS-rebinding protection.

## Changes

- `electron/shared-with-frontend/local-rest-api.model.ts`: Rename `LOCAL_REST_API_HOST` → `LOCAL_REST_API_DEFAULT_HOST`; add `LOCAL_REST_API_BIND_ADDRESS_OPTIONS`
- `electron/local-rest-api.ts`: Parameterize `startServer()` with bind address; make `ALLOWED_HOSTS` a function (`getAllowedHosts()`) for dynamic DNS rebinding check; restart server seamlessly on address change via `updateLocalRestApiConfig()`
- `src/app/features/config/global-config.model.ts`: Add `localRestApiBindAddress` to `MiscConfig`
- `src/app/features/config/default-global-config.const.ts`: Default to `'127.0.0.1'`
- `src/app/features/config/form-cfgs/misc-settings-form.const.ts`: Add select dropdown (hidden when REST API is disabled via `hideExpression`)
- `src/app/t.const.ts`, `src/assets/i18n/en.json`: Add i18n keys

## Checklist

- [x] The server restarts seamlessly when the bind address is changed (stop + start with new address)
- [x] DNS rebinding protection adapts to the configured address
- [x] Settings UI is Electron-only and hidden when the REST API toggle is off
- [x] Backward compatible — existing configs default to `127.0.0.1`
- [x] i18n strings added for English


johannesjo:
Thanks a lot for the contribution and the clear write-up\! 🙏

Unfortunately this is the same feature as #7717, which we declined on 2026-06-09 after a long review — see the [closing rationale](https://github.com/super-productivity/super-productivity/pull/7717#issuecomment-4663633525) (and the agreed won't-do on the linked issue #7350).

The core blocker is that **the local REST API is unauthenticated**. Binding it to `0.0.0.0` exposes task read/write to any device on the same network (home wifi, hotel, coworking, a compromised LAN device), which conflicts with the local-first/privacy promise of the app. The Host/Origin checks help against browser CSRF but don't stop a `curl` from another machine on the LAN.

The direction we'd accept is **bearer-token auth on the REST API first**, with external bind gated behind it — so "external" and "authenticated" become the same switch. Happy to discuss that in a separate issue if you're interested.

A couple of additional notes on this specific implementation, in case it's useful:

- **The Host-header allowlist doesn't include the machine's LAN IP.** `getAllowedHosts()` only contains `0.0.0.0`/`localhost` entries, so a real LAN client hitting `http://<lan-ip>:3876` sends `Host: <lan-ip>:3876`, which isn't in the set → `403`. To actually work over LAN you'd need to enumerate `os.networkInterfaces()` and add each interface IP (which is what #7717 ended up doing).
- **`localRestApiBindAddress` is synced.** It lives in `globalConfig.misc`, and the local-only mechanism (`LOCAL_ONLY_SYNC_KEYS`) only covers `globalConfig.sync`, not `misc`. So choosing `0.0.0.0` on one device would propagate via sync and bind another paired desktop's API to the LAN after hydration, without a per-device decision.

Going to close this as a duplicate of #7717 — but genuinely thank you for the effort, and the auth-first path is open if you'd like to pursue it. 🙏

PR #8515, opened 2026-06-20
Codex/quick add storeless hud 8387

symonbaikov:
## Problem

There was no quick way to capture a task from anywhere on the desktop without bringing the full Super Productivity window into focus first.

PR #8387 explored this idea, but the initial approach added a separate route/HUD implementation that diverged from the existing add-task UI and introduced architectural concerns around duplicated renderer state.

## Solution

This adds a Quick Add HUD window for Electron.

https://github.com/user-attachments/assets/e576a22b-bc61-46a5-9f19-98b49212a618

The implementation:

- adds a `superproductivity://quick-add` protocol flow for opening the HUD
- adds a compact frameless Quick Add `BrowserWindow`
- reuses the existing global `<add-task-bar>` UI instead of maintaining a copied HUD implementation
- keeps the selected app theme/styles applied to the HUD because it boots through the regular app path
- adds Quick Add specific app mode detection and shell hiding so only the add-task bar is visible
- forwards task submission from the Quick Add window to the main renderer so tasks are created through the normal app pipeline
- prevents Quick Add renderer side effects from re-registering global shortcuts
- adds Wayland/GNOME helper UI for manually creating a system shortcut when Electron global shortcuts are unavailable
- adds protocol/window/system tests and focused Angular coverage for the new behavior

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring
- [x] Documentation
- [ ] Other (please describe)

## Linux / Wayland Note

On Linux Wayland, Electron cannot reliably register global shortcuts by itself because the compositor controls system-wide shortcuts.

This implementation adds an in-app helper that explains the required command and opens the system keyboard shortcut settings, but the user still needs to manually create the custom shortcut in their desktop environment.

For GNOME/Wayland, the command is (production):

```sh
xdg-open superproductivity://quick-add
```

image

## For electron app in dev mode with wayland I used:
```sh
sh -lc 'cd /home/sbaikov/.config/superpowers/worktrees/super-productivity/codex-quick-add-storeless-hud-8387 && NODE_ENV=DEV npx electron . superproductivity://quick-add'
```
image

## New keyboard settings: 
image

## Testing

Automated checks run locally:

- `npm run checkFile` for changed TypeScript/SCSS files
- `npm run test:file src/app/features/config/store/global-config.effects.spec.ts`
- `node --test electron/quick-add-window.test.cjs electron/system.test.cjs electron/protocol-handler.test.cjs`
- `npm run electron:build`
- pre-push hook:
  - lint passed
  - package tests passed
  - full Angular suite passed: `11247 SUCCESS`
  - timezone suite passed: `11233 SUCCESS`

Manual testing performed:

- Linux / GNOME / Wayland
- configured a GNOME custom shortcut for `superproductivity://quick-add`
- opened the Quick Add HUD from outside the app
- verified the HUD visually matches the original global add-task bar
- verified task creation via Enter
- verified created tasks appear in the main app
- verified Escape/close behavior

## Important Manual Testing Note

macOS still needs manual testing.

I did not manually test this on macOS, and it should be verified there before merge because global shortcuts, custom protocol handling, focus behavior, and frameless transparent Electron windows can behave differently from Linux/GNOME/Wayland.


hajiboy95:
@symonbaikov this looks great! It's fully functional on macOS. Thank you so much for tackling this.
I did notice three minor things we should tweak before merging:
1. **HUD Transparency & Shadows:** The global quick-add HUD is currently semi-transparent. Let's make it fully opaque. To fix the visible shadow borders, we should either make the background slightly darker or remove the shadow altogether.
2. **Success Feedback:** After adding a task, it would be great to show a temporary hint inside the textbox so the user gets clear feedback. Something like: *"Task `<name stripped to ~35 chars>` successfully added."*
3. **Localization:** The global HUD doesn't respect the app's language settings right now. For example, if the app is set to German, the standard `Shift + A` bar opens in German, but the global HUD still opens in English.

grafik

hajiboy95:
@symonbaikov 
- success response is now prompted
- ugly circle due to shadow is gone

but
- hud is still semi transparent
- language still english 

grafik

Nit: 
- currently the first time I open the hud, i see for a second the bootup screen. Would it be possible to boot the hud together with the desktop application at start?

hajiboy95:
@symonbaikov 
we're still not there yet. I didn't mention my previous implementation on purpose to see whether you can come up with something better but it seems that codex did a similar thing: Spawn a bigger "border" behind it. However currently it's white. 

What I did, I just spawned something fullscreen. something dark but very highly transparent, like this the HUD didn't look transparent for me anymore. 
Also the language fallback doesn't seem to work properly? Both problems which i've actually fixed in @8387, maybe look how it's handled there. should be transferable. 
grafik

symonbaikov:
@hajiboy95 Yes. You are right. I will test it on my mac mashine and better to just copy your aproach. It will fix these issues. Will submit fixed version today.

johannesjo:
## Review — Quick Add HUD

Thanks for this — the core abstraction is genuinely nice: routing the add-task-bar's data dependencies through `ADD_TASK_BAR_DATA_FACADE` lets the same component run store-backed in the main window and IPC-backed in the HUD, without forking the UI (which is what sank #8387). Untrusted HUD input is re-validated in the main renderer (`quick-add-task-payload-validator.ts`), and every quick-add IPC handler checks its sender — that's the right security posture.

My concerns are mostly about **permanent surface area vs. payoff** and a couple of concrete correctness/perf issues. Findings below were each traced in the code and independently double-checked.

### 🔴 Blocker — stale base / conflicts with the add-task-bar redesign
The branch is **167 commits behind master** and conflicts in 10 files. Critically, master just landed **#8812 "redesign add-task-bar layout, toggles, note field & a11y"**, which substantively rewrote `add-task-bar.component.ts` / `.html` / `.spec.ts` — the exact files this PR rewrites (−307 lines on the component). This isn't a clean rebase: the facade extraction has to be re-derived on top of the redesigned component (which independently reworked the note field / toggles that this PR also touches). Merge-tree conflict set: `add-task-bar.component.{ts,html,spec.ts}`, `app.constants.ts`, `app.component.ts`, `electronAPI.d.ts`, `global-shortcuts.ts`, `protocol-handler.ts`. Worth reconciling before the diff can really be trusted.

### 🟠 Confirmed bug — the HUD preload never fires
`preloadQuickAddWindow()`'s comment says it "eliminates the first-open bootup screen the user otherwise sees while Angular loads," but on its only invocation both guards fail, so it always early-returns:
1. `QUICK_ADD_BRIDGE_READY` is sent **synchronously** from the `AppComponent` constructor (`quickAddTaskSubmitService.init()` → `informQuickAddBridgeReady()`), whereas `APP_READY` is only sent **after an `await`** in `startupService.init()`. So `getIsAppReady()` is `false` when preload runs.
2. Independently: within that single emit, main-window's preload listener is registered **before** the `quick-add-window.ts` listener that sets `isQuickAddBridgeReady = true` (`createMainWin` awaits `createWindow()` *then* calls `initQuickAddWindow()`), and `ipcMain` fires listeners in registration order — so `isQuickAddBridgeReady` is still `false` too.
3. The `APP_READY` handler never calls `preloadQuickAddWindow()`, and nothing else re-triggers it (`QUICK_ADD_TASK_SUBMIT_BRIDGE_READY` has no main-side listener).

Net: the HUD is always lazily created on first `showQuickAddWindow()`, so the first open still shows the Angular boot — the stated optimization is dead. Functionally harmless, but either wire preload into the `APP_READY` handler or drop the mechanism + comment. (Note: `quick-add-window.test.cjs` sets `isAppReady`/bridge-ready manually and calls preload directly, so it passes without exercising the real trigger order — it masks this.)

### 🟠 Perf — 250 ms poll that runs forever in a non-throttled hidden window
`QuickAddHudDataFacadeService._signalToObservable()` bridges signals to observables with `setInterval(…, 250)` and **no `distinctUntilChanged`**, emitting every 250 ms regardless of change. `projects$`/`tags$` feed the `combineLatest` in `AddTaskBarComponent._setupTextParsing()`, so **short-syntax parsing re-runs ~4×/sec continuously** for the life of the HUD component. Since the HUD window is created-then-**hidden** on close (only `destroyQuickAddWindow()` on before-quit) and is deliberately `backgroundThrottling: false`, this keeps firing at full rate **even while the HUD is hidden**, for the rest of the session once opened. `FullAddTaskBarDataFacadeService` already uses `toObservable()` from `@angular/core/rxjs-interop` — please use that here (or at least add `distinctUntilChanged` and/or stop the interval on hide).

### 🟡 Architecture / long-term cost
- **Partial `window.ea` in the HUD is a maintenance footgun.** `IS_ELECTRON` is userAgent-based, so it's `true` in the HUD renderer, but the HUD preload exposes only ~4 `ea` methods. This already forced `app.constants.ts` into `!!window.ea?.isGnomeWayland?.()`. **Any future module-level `window.ea.X()` reachable from the quick-add bootstrap will silently break the HUD**, with no compile-time signal and no coverage from the main-app tests. Consider a shared no-op `ea` shim or an explicit typed quick-add-mode guard so the failure is loud.
- **Two resident Angular runtimes + a new cross-process contract** (`QuickAddHudSnapshot`, `AddTaskPayload` spanning preload/main/HUD). Reasonable given IndexedDB can't be shared across renderers, but it's a real permanent cost — worth calling out explicitly against the project's "avoid feature creep" guard, especially for a desktop-only feature that on Wayland (the common modern-Linux case) still needs a manually-created OS shortcut.

### 🟡 Latent (not currently reachable, so not a blocker)
`initQuickAddWindow()` calls `ipcMain.handle(...)` without a prior `removeHandler` and isn't idempotent, so a second `createMainWin()` in one process would throw "Attempted to register a second handler." I checked and the macOS `activate → createMainWin(mainWin === null)` path that would trigger it is **unreachable today** (the `close` handler `preventDefault()`s unless quitting, so `mainWin` is never nulled mid-session; `window-all-closed` force-quits). A one-line idempotency guard would just harden it against future refactors.

### 🟢 Minor
- The `globalTaskQuickAdd` shortcut field shows for all Electron users, but the "may not work, here's the command" helper only appears on **GNOME** Wayland — KDE/other Wayland users get a silently-non-functional shortcut and no guidance.
- New `WAYLAND_QUICK_ADD_*` keys are out of alphabetical order in `en.json` / `t.const.ts` (inserted before `GLOBAL_SHOW_HIDE`) — may trip the i18n order check.
- `_readThemeSnapshot()` serializes *all* `--*` custom props from `<html>` + `<body>` over IPC on each open — fine now, just scales with theme-var count.
- Per the description, macOS is still untested — the frameless/transparent full-screen-shell + focus/`app.hide()` behavior is the riskiest platform here.

### What's good
Sender authorization on every handler (`_isMainWindowSender`, `fromWebContents === quickAddWin`); untrusted-input validation in the main renderer; pending-request timeouts + `_rejectAllPending` on hide/close/destroy; `will-navigate`/`will-redirect`/`setWindowOpenHandler` lockdown on the HUD webContents; OAuth URL redaction preserved.

**Bottom line:** sound design on a clean abstraction, but I'd hold merge on (1) reconciling with the #8812 redesign, (2) fixing or removing the dead preload path, and (3) the 250 ms hidden-window poll. Whether the permanent complexity is worth a desktop-only, manual-setup-on-Wayland feature is a scope call for the maintainers — not a code defect.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

hajiboy95:
@symonbaikov @johannesjo I finally had time to give this a proper second run, so I've continued it in #9789 rather than leaving it stalled here.

What that PR does:

- **Resolved the merge conflicts with master.** The branch had fallen 589 commits behind and conflicted in 9 files. It wasn't a mechanical rebase — #8812 and the work after it rewrote the exact submit path this PR extracts (`repeatQuickSetting` became a `repeat` variant, plus weekend rolling, interval recurrences and `skipOverdue` seeding), so the facade extraction had to be re-derived on top of the redesigned component.
- **Fixed the open feedback from this thread**, including the two I kept coming back to: the HUD is no longer semi-transparent, and it follows the app language.
- **Fixed several bugs found while verifying it**, the most important being that the HUD didn't boot at all any more: master's `app.constants.ts` gained a module-level `window.ea.isMacOS()`, which threw during the HUD's Angular bootstrap and left the loading splash up permanently. That's exactly the `window.ea` footgun @johannesjo predicted in the review here, so it's now a total shim rather than one method longer.

**I've verified it on macOS**, which was the open item at the top of this PR's description. Linux/Wayland hasn't been re-verified since the merge — @symonbaikov, if you get a chance to try #9789 on your setup that would be very welcome.

@symonbaikov thank you for the original implementation and for iterating through several rounds of feedback here — the facade abstraction is what makes this work without forking the add-task-bar UI, and it's the reason this approach succeeded where #8387 didn't. @johannesjo thank you for the detailed review; it caught real defects and the one prediction in it turned out to be exactly what broke.

The scope question from that review is untouched and still yours to make — #9789 fixes the defects, it doesn't try to settle whether the feature earns its keep.

Closing thoughts aside, I'd suggest keeping this PR open until #9789 is reviewed, in case anything is better resolved here.

johannesjo:
**♻️ Master has moved on** – this branch is 598 commits behind `master`; see below for what still applies.

Thanks for this, and for iterating through several rounds of feedback on it. Since the last commit here (`5d0b6ed`, 8 Jul) master has moved past the branch, and GitHub now reports it as conflicting:

- `src/main.ts` – rewritten on master (534 to 594 lines, including the `provideAppInitializer` migration), so this PR's wholesale `main.ts` to `app-main.ts` move lands as a full-file conflict rather than a rebase hunk.
- `src/app/features/tasks/add-task-bar/add-task-bar.component.ts` – reworked by the add-task-bar redesign and the work after it, which is the same submit path the facade extraction rewrites.
- `src/app/app.constants.ts` – now evaluates `window.ea.isMacOS()` at module scope, and the `ea` shim in `electron/quick-add-preload.ts` doesn't expose that method, so against current master the HUD renderer would throw during Angular bootstrap.

The design isn't what went stale – the facade abstraction is still the reason this approach works without forking the add-task-bar UI. It's that the branch can no longer be rebased mechanically onto what master became.

@hajiboy95 has already carried the work forward in #9789 with the merge reconciled and the open feedback from this thread addressed, so nothing here needs another round from you. What's left is a maintainer call: keep this open alongside #9789 as suggested above, or close it in its favour.

---

johannesjo:
Closing this one too, for the same reason — and with the same thanks.

I've decided not to take the Quick Add HUD, and I've explained the reasoning in full over on #9789, which carried this work forward: https://github.com/super-productivity/super-productivity/pull/9789#issuecomment-5461982016

Short version: a second resident Angular runtime plus a new cross-process contract and a `src/main.ts` bootstrap split is permanent surface paid by every desktop user, and `globalAddTask` already registers a global shortcut that opens the add-task bar from anywhere. The delta didn't clear the bar for me.

That's a scope judgement, not a comment on the work — you built this before the add-task-bar rework landed and it was a reasonable thing to try. If the existing `globalAddTask` shortcut is falling short in some specific way (too slow, focus handling, unreachable when the window is hidden), an issue about that concrete failure is very welcome; fixing it there helps everyone already using the shortcut.

PR #8516, opened 2026-06-20
feat(azure-devops): add optional WIQL override for auto backlog import

### LINKED ISSUE #7674: 💡 Azure DevOps: Allow filtering by iteration ("Iteration Path") and teams/subteams ("Area Path"), issue type and more (WIQL support?)
### Problem Statement

As a followup from https://github.com/super-productivity/super-productivity/pull/6446#discussion_r2798480446 even when testing manually I have the problem to find the actual task/issues to work on.

Specifically these filters seem to be useful:
* by type ([bug/task/Product backlog items etc.](https://learn.microsoft.com/en-us/azure/devops/boards/work-items/guidance/scrum-process?view=azure-devops))
* by iteration ("sprint" aka in agile developments you usually work in iterations that are natively supported by ADS)
* by area (aka "projects", these may be useful, note ADS has a tree-hierarchy here aka allows for subareas separated by `/`)

### Possible Solution

ADS has a "key-value" schema for many properties, as it is very customizable it can have different names, but usually "Iteration Path" is the iteration and "area path" is the "team":

Image

Maybe just allow passing a custom filter or so?

Actually best supports a custom query langauge: [WIQL reference](https://learn.microsoft.com/en-us/azure/devops/boards/queries/wiql-syntax?view=azure-devops)
It is AFAIK the same and can also [be defined from the GUI](https://learn.microsoft.com/en-us/azure/devops/boards/queries/using-queries?view=azure-devops&tabs=browser).

* As shown above, even in the GUI you can define queries. Maybe allow selecting just "a query" from that list?
* Or maybe a simpler implementation by just hardcoding the filter? As the current selection "only assigned to me" etc. allows?

### Additional context

IIRC your Jira integration has similar support/also supports their query language, so that may be useful.

costajohnt:
## Problem

Closes #7674.

The Azure DevOps issue provider can only filter auto-imported work items by project and a fixed scope (All / Created by me / Assigned to me). Users asked to also filter by iteration path ("sprint"), area path ("team"/area), and work item type, ideally through WIQL, the query language Azure Boards already exposes in its own GUI.

## Solution

Adds an optional `autoImportWiql` field to the plugin's advanced config (`packages/plugin-dev/azure-devops-issue-provider`):

- **Empty** (default): behavior is identical to today, the scope dropdown plus the done-state exclusion. Fully backward compatible.
- **Set**: the WIQL is used as the *entire* backlog query, so the user owns scope, the state filter and ordering, for example filtering by `[System.IterationPath]`, `[System.AreaPath]` or `[System.WorkItemType]`, or matching custom done-state names on non-default ADO processes.

This follows the direction discussed in the issue: a full-query override (mirroring Jira's `autoAddBacklogJqlQuery`, which is the whole query the user controls) rather than appending an ANDed fragment. An appended fragment could only narrow the hard-coded English done-state exclusion (`Closed`/`Done`/`Removed`) and couldn't accept a real exported `Select ... From WorkItems Where ...` statement.

Scope is intentionally limited to auto-import; search uses a separate query and is out of scope here.

## Type of Change

- [x] New feature

## Checklist

- [x] Documentation/wiki updated (`docs/wiki/3.07-Issue-Integration-Comparison.md`)
- [x] Lint/format on changed files: root ESLint ignores `packages/plugin-dev/**`, so I ran the plugin's own `tsc --noEmit` plus Prettier on every changed file (all clean)
- [x] Added tests (`src/plugin.spec.ts`, vitest, 6 cases) and wired the plugin into the `plugin-tests` CI matrix
- [x] Commit message follows Angular format

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -34,7 +34,7 @@ jobs:
         run: |
           set -euo pipefail
           # Plugins that have an `npm test` script configured.
-          PLUGINS=(automations caldav-calendar-provider clickup-issue-provider google-calendar-provider sync-md)
+          PLUGINS=(automations azure-devops-issue-provider caldav-calendar-provider clickup-issue-provider google-calendar-provider sync-md)
 
           if [ "$EVENT_NAME" = "workflow_dispatch" ]; then
             CHANGED=$(printf '%s\n' "${PLUGINS[@]}")
--- a/docs/wiki/3.07-Issue-Integration-Comparison.md
+++ b/docs/wiki/3.07-Issue-Integration-Comparison.md
@@ -10,20 +10,20 @@ Every issue provider supports the same basic operations: checking that the integ
 
 ## Feature Matrix Summary
 
-| Provider        | Import | Status        | Worklog    | Comments | Subtasks | Attachments | Filtering           | Auto-import | Story pts | Due dates |
-| --------------- | ------ | ------------- | ---------- | -------- | -------- | ----------- | ------------------- | ----------- | --------- | --------- |
-| Jira            | Yes    | Yes           | Yes        | Yes      | Yes      | Yes         | JQL                 | Yes         | Yes       | —         |
-| GitHub          | Yes    | No            | No         | Yes      | No       | No          | Repo, assignee      | Yes         | —         | —         |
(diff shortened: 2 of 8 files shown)


johannesjo:
Thank you very much! <3

PR #8519, opened 2026-06-21
Clarify empty CalDAV option loads

### LINKED ISSUE #8379: CalDAV Events: empty calendar dropdowns give no guidance on first setup (load order + docs)
Follow-up to #8259 (the discovery bug is fixed in v18.10.0 by #8278). This tracks the **UX / documentation** residuals surfaced in that thread — see https://github.com/super-productivity/super-productivity/issues/8259#issuecomment-4695456942.

### Problem

On a **new** CalDAV Events setup, the "Calendars to display" and "Default calendar for new events" dropdowns are filled by `loadOptions`, but nothing triggers a load until the user clicks **Load Calendars** or runs **Test Connection**. Opening an empty dropdown does nothing and gives no hint, so it reads as broken — this is exactly what the original #8259 report described as "the text goes green and nothing happens."

Relevant code:

- `src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.ts` — `_initOAuthAndOptions()` only auto-loads options in **edit** mode; new setups just show the "Load Calendars" button. `_loadDynamicOptions()` also runs after a successful Test Connection.
- the dropdowns are plain `mat-select`s that render whatever `options` currently holds — they never fire `loadOptions` on open.

### Suggested fix (any of)

- show a placeholder / disabled state on the dynamic dropdowns until options are loaded ("Enter credentials, then Load Calendars"), plus a "No calendars found" empty-state when discovery returns an empty list;
- and/or trigger `loadOptions` when a dynamic dropdown is opened and credentials are present;
- add a one-line note to the wiki that the **server root URL now works** (e.g. `https://host/remote.php/dav` for Nextcloud) and that you must Load Calendars / Test Connection to populate the lists.

### Secondary (lower priority)

**Test Connection false positives** — `testConnection` in `packages/plugin-dev/caldav-calendar-provider/src/plugin.ts` is only a `Depth:0` PROPFIND on the entered URL, so it passes for root / principal / single-calendar URLs that will not actually populate calendars. Making it run discovery and report the calendar count would turn the green check into a meaningful signal.

### Related

- #8061 — same plugin, `timeBlockCalendarId` options loading; likely resolved by #8278, worth verifying on v18.10.0.
- #8275 — CalDAV **tasks** (VTODO) has the analogous "connects but shows nothing" behaviour in a separate provider; the VTODO URL matrix posted there is the input for an equivalent `current-user-principal` / `calendar-home-set` bootstrap on the tasks side.

archit-goyal:
## Summary

- Add an empty state for dynamically loaded plugin options so CalDAV calendar selects no longer appear as silently empty after discovery returns no calendars.
- Disable dynamic option selects until options have loaded, with placeholders for unloaded, loading, failed, and empty states.
- Document the CalDAV Events setup order for loading calendar lists.

## Why

When the CalDAV Events provider has not loaded calendars yet, or when discovery returns zero calendars, the calendar dropdowns can look like ordinary empty selects. This makes first setup confusing because users do not get clear guidance that they need to load options or check the CalDAV URL/account details.

Fixes #8379.

## Verification

- `npm run checkFile src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.ts`
- `npm run checkFile src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.scss`
- `npm run lint:file -- /Users/archit/Documents/super-productivity-contributions/super-productivity/src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.html`
- `npm run int:test`
- Pre-push hook: `npm run lint` and `npm run test`


johannesjo:
Nice work, and welcome! 🎉 This lands the scope from #8379 cleanly — wiring into the existing `optionsLoadState` signal + `@switch` exactly as suggested, and the `{ isSuccess, hasOptions }` plumbing is the right fix for the subtlety @johannesjo flagged (a zero-calendar discovery now correctly reads as `empty` rather than `loaded`, since `loadCalendars` returns `[]`). I traced the formly path and confirmed the disable-until-loaded mechanism is correct: formly v7 installs a reactive setter on `props.disabled`, so the imperative mutation does propagate to the form control.

A few **non-blocking** suggestions:

1. **`_refreshDynamicOptionFieldStates` — mutate only `templateOptions`.** Post-init, formly aliases `templateOptions` to `props` (same object), so the dual `templateOptions.x = …; props.x = …` writes are redundant. More importantly, `const props = (formlyField.props ??= {})` would create a *stray* empty object if it ever ran before formly initialized the field (formly's `field.props ??= field.templateOptions` would then skip merging the real config). It doesn't fire today because the edit-mode auto-load sits behind an async token read and CD runs first — but it's an implicit timing dependency. Reading/writing only `templateOptions` is simpler and removes the fragility.

2. **State + view flush ownership.** `_loadDynamicOptions()` calls `_refreshDynamicOptionFieldStates(...)` and runs `detectChanges()`, then the caller `_setOptionsLoadState()` refreshes again with no trailing `detectChanges()`. On a *failed* load this briefly renders the `empty` placeholder ("No options found") before the signal-driven CD flips it to "Failed to load options". Harmless flicker, but centralizing the refresh + `detectChanges()` in `_setOptionsLoadState` would make the state→view flow easier to follow.

3. **`delete templateOptions.placeholder`** removes the property that may carry formly's observer accessor; setting it to `undefined`/`''` keeps reactivity intact. Cosmetic-only.

4. The `state = this.optionsLoadState()` default param on `_refreshDynamicOptionFieldStates` is now always called with an explicit arg — dead default.

5. **Tests:** `getDynamicOptionPlaceholder` is a pure function and the `{isSuccess, hasOptions} → state` mapping is the heart of the fix; a tiny spec would lock in the "zero calendars ≠ failure" behavior. The component has no spec today, so this is optional, but cheap.

Since there's no automated coverage here and only the `welcome` check has run so far, worth a quick manual pass: **edit an existing CalDAV provider with saved calendars** (confirm the dropdowns re-enable with correct labels after auto-load) and a **fresh setup whose account has zero calendars** (confirm the `empty` state + help text shows). The new-setup empty path is the actual #8379 symptom.

Overall this looks correct and on-scope — just polish. 👍

archit-goyal:
Thanks for the detailed pass and the welcome!

I pushed a small follow-up commit addressing the non-blocking polish items 1-4:

- `_refreshDynamicOptionFieldStates()` now reads/writes only `templateOptions`.
- The refresh + `detectChanges()` path is centralized in `_setOptionsLoadState()`.
- Placeholder clearing now assigns `undefined` instead of deleting the property.
- Removed the unused default parameter from `_refreshDynamicOptionFieldStates()`.

I left the optional spec out for now because this dialog does not currently have a component spec, and I wanted to keep the PR small/on-scope. Local validation passed again, including `checkFile` for the touched TS file and the repo pre-push hook (`npm run lint` + `npm run test`).

johannesjo:
Did a second independent pass on top of the review above (checked against `pull/8519/head` + the installed formly 7.1.0 source). Net: solid and safe to merge — polish items are all in, and a few things I verified rather than assumed:

- `props.disabled` reactivity is correctly wired: `field-form/utils.mjs` runs `observe(field, ['props','disabled'], …) → formControl.disable()/enable()`, and `templateOptions` aliases `props` post-init, so the imperative toggle does propagate. 👍
- `submit()` reads from `this.model`, not `form.value`, so disabling the selects never drops saved `readCalendarIds`/`writeCalendarId`. The CalDAV calendar fields are also `required: false`, so disabling doesn't change form validity/submit gating.
- Nothing touches sync/op-log state — pure dialog UI.
- The `{ isSuccess, hasOptions }` split correctly makes a zero-calendar discovery resolve to `empty` (the actual #8379 symptom), not `loaded`.

**One genuine, minor/optional finding — dynamic placeholders are raw i18n keys, not translated.**

`_refreshDynamicOptionFieldStates()` assigns e.g. `templateOptions.placeholder = T.F.ISSUE.DIALOG.NO_OPTIONS_FOUND` directly. Our `TranslateExtension` only wires placeholder translation once, in `prePopulate`, via an `expressionProperties` stream bound to the *initial* key (`LOAD_OPTIONS_FIRST`). A later direct assignment isn't re-translated — formly `observe()`s `props.disabled` but not `props.placeholder` — so the field's placeholder holds the raw key string (`"F.ISSUE.DIALOG.NO_OPTIONS_FOUND"`) until the next language change, which then reverts it to the translated `LOAD_OPTIONS_FIRST` regardless of state.

Visibility is low in practice: these selects have a `<mat-label>`, so for the disabled `loading`/`empty`/`idle` states the label sits in the resting slot and the placeholder never renders. The one path where the raw key actually surfaces is `failed` after a prior successful load (options remain → field stays enabled) → focusing the select shows `"F.ISSUE.DIALOG.LOAD_OPTIONS_FAILED"`. Still worth tidying, as it's against the "all UI strings via `translate`" rule.

Simplest fix: drop the dynamic `placeholder` assignment in `_refreshDynamicOptionFieldStates()` and keep only the `disabled` toggle — the `@switch` block already shows translated status text (`LOADING_OPTIONS` / `NO_OPTIONS_FOUND_HELP` / `LOAD_OPTIONS_FAILED`) right next to the field, and the initial translated `LOAD_OPTIONS_FIRST` stays correct. (Alternatively inject `TranslateService` and assign `instant(key)`.)

Two small non-blockers:
- The `disabled: true` default in `_mapPluginConfigField` now applies to *every* `loadOptions` select, not just CalDAV. I traced the other two — Trello (`loadBoards`, no OAuth → Load Options button) and Google Calendar (OAuth → auto-loads on connect) both re-enable after their load/connect path, so this is fine (arguably an improvement); just flagging it's intended.
- Optional: exporting `getDynamicOptionPlaceholder` + a ~3-line spec for the `hasOptions → loaded/empty` mapping would lock in the "zero calendars ≠ failure" behavior, which is the heart of the fix.

CI hasn't run on the branch ("no checks reported"), so the manual two-path check is worth it: fresh setup whose account has zero calendars → `empty` state + help text; edit an existing provider with saved calendars → selects re-enable with correct labels.

Nice work overall, and welcome again\! 🎉

archit-goyal:
Thanks for the second pass. I pushed one more small follow-up for the dynamic placeholder/i18n edge case:

- Removed the runtime placeholder mutation from `_refreshDynamicOptionFieldStates()`.
- Removed the now-unused `getDynamicOptionPlaceholder()` helper and `NO_OPTIONS_FOUND` translation key.
- Kept the initial translated `LOAD_OPTIONS_FIRST` placeholder and the translated `@switch` status/help text.

Validation passed again:

- `npm run checkFile src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.ts`
- `npm run int:test`
- `git diff --check`
- repo pre-push hook (`npm run lint` + `npm run test`)

johannesjo:
## Review — Approve with minor notes ✅

Solid, well-scoped fix. Adds an `empty` state so CalDAV calendar dropdowns no longer look silently broken, disables the selects until options load (with a translated "Load options first" placeholder), and documents the setup order. No correctness or sync risks.

### Verified locally
- `eslint` (incl. local rules), `prettier`, and `tsc --noEmit` all pass on the changed files.
- **No sync/state implications** — purely UI dialog state (no reducers/effects/selectors/op-log).
- **Dropping the `props` fallback is safe.** Formly v7 makes `templateOptions` a getter/setter alias of `props` post-init (`initFieldProps`), and dynamic fields are always built with `templateOptions`, so the removed `else if (formlyField?.props)` branch was dead for these fields.
- **`disabled` reactivity is sound.** Formly installs `observe(field, ['props','disabled'])` that calls `formControl.disable()/enable()` on mutation, so the imperative `templateOptions.disabled = …` in `_refreshDynamicOptionFieldStates` takes effect.
- **The static placeholder is translated.** `TranslateExtension.prePopulate` binds `templateOptions.placeholder` to a translate stream at build time. Commit 3 ("Avoid dynamic option placeholder keys") correctly removed the *runtime* placeholder mutations, which would have rendered raw keys (`prePopulate` runs only once) — good catch.
- **No orphaned i18n** — the removed `NO_OPTIONS_FOUND` key is gone everywhere; only `en.json` was edited.
- **`options.length` can't crash** — `loadOptions` is typed to return an array, and a misbehaving plugin returning `undefined` throws inside the existing `try/catch` → treated as `failed`.

### Notes

**1. (Low) The `disabled: true` default applies to every `loadOptions` field, not just CalDAV.** CalDAV and **Trello** (API-key auth) use the "Load Options" button path and behave exactly as intended. **Google Calendar** is the one that behaves differently: its `readCalendarIds` / `writeCalendarId` / `timeBlockCalendarId` selects carry a static `options: [{label:'Primary'}]` plus `loadOptions`, so that "Primary" quick-pick is now disabled until a load runs. It re-enables correctly after OAuth connect / Test Connection (`_loadAndSetDynamicOptionsState`) and in edit mode with valid tokens. It only stays disabled in pre-connect / expired-token states — which are non-functional anyway, so this is defensible — but it's a cross-provider change the CalDAV-scoped description/verification don't mention. Worth a quick manual check that Google Calendar selects enable as expected.

**2. (Low) The placeholder is static across states.** Since commit 3 it's always the translated "Load options first" (bound once via `expressionProperties`). After a successful load, an unselected dropdown still reads "Load options first," and for OAuth providers it says that when the real action is "Connect." Minor wording staleness.

**3. (Low) No tests.** No spec exists for this component and none was added, despite non-trivial new state logic (`_loadDynamicOptions` return shape, `empty` vs `failed` derivation, the per-field disable rule). A small spec covering empty/failed/loaded → disabled mapping would lock in the behavior.

**4. (Nit)** `loadOptions?: unknown` in `_mapPluginConfigField`'s param type is loose — it's only used as a truthiness check, but the concrete function signature reads better.

archit-goyal:
Hi @johannesjo, just gently following up on this PR. The follow-up polish from your review is in, CI is green, and the branch is currently mergeable. Please let me know if there is anything else you would like changed before merge. Thanks!

johannesjo (changes requested):
## Summary

This is a focused UI fix for dynamic provider options: it adds a distinct empty state, disables dynamic selects until usable options are present, and documents the CalDAV setup flow.

## PR Discussion Context

The current head is `4bd98a2`. There are no formal reviews or inline threads. Earlier Formly mutation and runtime-i18n concerns were addressed. The linked issue still explicitly requires documenting that server-root CalDAV URLs are supported.

## Issues

### Important

- **docs/wiki/2.07-Manage-Task-Integrations.md:29** — The new troubleshooting text says to check that the URL points to a calendar home or calendar URL. That contradicts #8379, the plugin field description, and the existing discovery regression test, all of which explicitly support server-root URLs such as Nextcloud `/remote.php/dav`. A user with a valid root URL could therefore be directed away from the correct configuration. Please mention that either a supported server-root or calendar-home URL works, then direct users to verify credentials and event-calendar availability.

- **src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.ts:487** — The new `loaded` / `empty` / `failed` transition and the per-field disabled mapping have no regression coverage. Existing CalDAV tests cover discovery, not this dialog behavior, so a regression that maps an empty successful result back to `loaded` or enables an empty field would still pass CI. The repository requires unit tests for new state logic. Please cover successful non-empty, successful empty, and failed loads, either through a focused component spec or extracted pure state predicates.

### Minor

- **src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.ts:713** — The static “Load options first” placeholder remains after a successful load, so an unselected dropdown can show stale guidance. This is low visibility because the field label usually occupies that space.

## Strengths

- The empty-vs-failed distinction correctly models the reported zero-calendar case and reuses the existing Formly/state flow.
- The styling, translation, dependency, privacy, and sync boundaries remain clean.

## Verdict

**Needs changes.** The head is mechanically mergeable and all required CI checks pass, but the inaccurate CalDAV documentation and missing state-logic regression coverage should be addressed before merge.

johannesjo (approved):
Verified the maintainer follow-up on c1564c316d: the wiki now explicitly documents supported server-root and calendar-home URLs, resolving the only blocking issue. The previously noted lack of component-level state tests and static placeholder wording remain non-blocking; the implementation itself was already verified. All checks pass. Approved.

johannesjo:
Thank you very much! <3

PR #8522, opened 2026-06-21
fix(tasks): add visible keyboard focus ring to subtasks (#8520)

### LINKED ISSUE #8520: 🚨 Missing focus indicator for subtasks during keyboard navigation
### Version Used

MacOS 18.12.0

### Current Behavior
When navigating the task list using the keyboard (e.g., pressing the Down arrow key), if a task has subtasks, the subtasks do not display a visible focus indicator. Because the focus ring is missing on subtasks, it visually appears as though the keyboard focus has been lost entirely, even though the subtask is technically focused in the background.

### Expected Behavior
When navigating into subtasks using the keyboard, there should be a clear, visible focus indicator (such as a focus ring or background highlight). The user should always be able to visually identify exactly which subtask is currently focused.

### Steps to Reproduce

1. Create a parent task and add at least one subtask to it.
2. Ensure the parent task is expanded so the subtasks are visible in the task list.
3. Select a task above it and use the keyboard (Down Arrow key) to navigate down the list.
4. Move the focus from the parent task onto the subtask.
5. Observe that the subtask receives no visual focus styling, creating the illusion of lost focus.

### Documentation

hajiboy95:
## Problem

When navigating the task list using the keyboard, subtasks do not display a visible focus indicator (focus ring). This creates the illusion that keyboard focus has been lost entirely, even though the subtask is technically focused. Resolves #8520.

## Solution

Added a specific focus style for subtasks using the `:host-context(.sub-tasks)` selector in `_task-base.scss`. When a subtask receives keyboard focus, a solid 1px focus ring of color `var(--palette-primary-400)` is now drawn around its box wrapper (`.box`), matching the focus indicator behavior of parent tasks:

```scss
:host-context(.sub-tasks):focus .inner-wrapper > .box {
  @include noTouchOnly() {
    border-color: var(--palette-primary-400) !important;
    border-style: solid !important;
    border-width: 1px !important;
  }
}
```

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/tasks/task/_task-base.scss
+++ b/src/app/features/tasks/task/_task-base.scss
@@ -222,14 +222,6 @@
   }
 }
 
-:host-context(.sub-tasks):focus .inner-wrapper > .box {
-  @include noTouchOnly() {
-    border-color: var(--palette-primary-400) !important;
-    border-style: solid !important;
-    border-width: 1px !important;
-  }
-}
-
 :host-context(.isDarkTheme) ::ng-deep .isDone.isDone .box {
   background: var(--sub-task-c-bg-done);
 }
--- a/src/app/features/tasks/task/_task-base.scss
+++ b/src/app/features/tasks/task/_task-base.scss
@@ -222,6 +222,14 @@
   }
 }
 
+:host-context(.sub-tasks):focus .inner-wrapper > .box {
+  @include noTouchOnly() {
+    border-color: var(--palette-primary-400) !important;
+    border-style: solid !important;
+    border-width: 1px !important;
+  }
+}
+
 :host-context(.isDarkTheme) ::ng-deep .isDone.isDone .box {
   background: var(--sub-task-c-bg-done);
 }
--- a/src/assets/themes/liquid-glass.css
+++ b/src/assets/themes/liquid-glass.css
@@ -860,6 +860,14 @@ schedule-event {
   border-top-color: rgba(var(--ink-on-channel), 0.06) !important;
 }
 
+/* The single-edge separator above sets border-color: transparent !important,
+ * which also erases the base keyboard focus ring on subtasks
+ * (:host:focus … .box uses a non-!important border). Restore it so a focused
+ * subtask matches a focused parent task (both use --palette-primary-400 here). */
+.sub-tasks task:focus .box {
+  border-color: var(--palette-primary-400) !important;
+}
+
 /* Current task fill — applied in both modes. _task-base.scss only
  * sets a background for the current task in dark mode (light mode
  * gets only a dashed accent border), but with `position: sticky`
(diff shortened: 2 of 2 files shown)


johannesjo:
Thanks for digging into this! I traced the root cause and verified it in a browser — sharing in case it helps decide where the fix should live.

### Root cause

This is exclusive to the **Liquid Glass** theme. `src/assets/themes/liquid-glass.css` (lines ~852–861) intentionally replaces the subtask's 4-side border with a single top-edge separator:

```css
.sub-tasks .box {
  border-color: transparent !important;
  border-top-color: rgba(var(--ink-on-channel), 0.06) !important;
}
```

That `border-color: transparent !important` also wipes out the base keyboard focus ring, because the base rule sets the border **without** `!important`:

```scss
// _task-base.scss:111
:host:focus .inner-wrapper > .box { ... border-color: var(--palette-primary-400); ... }
```

So the theme's `!important` wins and the subtask ring vanishes. Parent tasks aren't affected (they aren't `.sub-tasks .box`), and no other theme has this override — which is why the maintainer couldn't reproduce it on a different theme.

I confirmed it with a faithful Angular emulated-encapsulation DOM in headless Chromium:

| State | Focused subtask border |
|---|---|
| Base CSS, normal theme | `rgb(0,120,255)` — ring shows (base CSS is already correct) |
| Liquid Glass, no fix | `rgba(0,0,0,0.06)` top / transparent — **no ring (the bug)** |
| Liquid Glass + this PR | `rgb(0,120,255)` all sides — ring restored |

So this PR **works**, and the `!important`s are genuinely required to win the `!important` fight the theme started.

### Suggestion: fix it in the theme instead

Since the base focus ring already works correctly on every other theme (verified above), adding an `!important` override to the shared, hot-path `task` component means ~15 themes carry a redundant rule to counter a problem only one theme creates. The regression lives in `liquid-glass.css`, so it's cleaner to fix it there — right after the `.sub-tasks .box` block:

```css
/* The single-edge separator above sets border-color: transparent !important,
 * which also erases the base keyboard focus ring on subtasks
 * (:host:focus … .box uses a non-!important border). Restore it so a focused
 * subtask matches a focused parent task (both use --palette-primary-400 here). */
.sub-tasks task:focus .box {
  border-color: var(--palette-primary-400) !important;
}
```

`.sub-tasks task:focus .box` (specificity 0,3,0 + `!important`) beats the separator rule (0,2,0 + `!important`), it's a global stylesheet so no encapsulation gymnastics, and it keeps the base component untouched. `--palette-primary-400` matches how parent tasks already render their focus ring in this theme.

### Longer-term note (optional)

The underlying fragility is that the base focus ring is a non-`!important` `border-color`, so any theme can erase it with an `!important` border. Since a visible focus indicator is an accessibility baseline, it may be worth eventually moving the base ring to an `outline` (outline isn't part of the border cascade, themes don't override it, no layout shift) — but that's a larger change than this bug needs.

hajiboy95:
@johannesjo adressed. made the change less invasive

johannesjo:
Thank you!

PR #8524, opened 2026-06-21
Fix/issue 8407 selection edit

### LINKED ISSUE #8407: 💡 Feature Request: Explicit Edit mode for Task Notes (allow text selection & checkbox toggling in view mode)
### Problem Statement

The current interaction model for the Note field in the task side panel is quite confusing and sensitive. Clicking into the note to read it, select text, or interact with a to-do list often accidentally triggers the edit mode.

Because the field jumps into edit mode so easily, it is currently very difficult to simply highlight and copy text from a note. Additionally, managing simple markdown to-do lists within the note feels clunky when every interaction tries to open the full text editor.

### Possible Solution

Implement a distinct View/Edit toggle for task notes, functioning very similarly to how markdown text fields work on GitHub:

* **View Mode (Default):** The note is rendered as static markdown. In this mode, users can easily click and drag to select and copy text. If there are markdown checkboxes (`[ ]` / `[x]`), users can click them to toggle the state *without* triggering the full edit mode.
* **Edit Mode:** Triggered only by an explicit action, such as an "Edit" button (e.g., a pencil icon) at the top of the note field, or double-clicking the note. This opens the actual text area for writing.

* **Careful clicking and dragging:** Trying to precisely click and drag to select text without triggering the input focus.
* *Drawback:* This rarely works as intended. It usually results in triggering edit mode, which shifts the layout slightly and ruins the text selection, causing frustration.

### Additional context

* Separating the reading/interacting experience from the writing experience makes the side panel feel much more robust. It prevents accidental keyboard inputs from ruining notes when the user just wanted to copy a snippet or tick off a sub-task.

hajiboy95:
## Problem

These changes address multiple issues in the inline markdown component, primarily focusing on fixing:
1. **Selection & Editing Conflicts (fixes #8407)**: Highlighting text or right-clicking to open the context menu in the preview was impossible because any drag/click action instantly triggered edit mode, clearing the active selection.
2. **Missing Checklist Spacing**: Checklist lists rendered in preview lost their blank line spacing, causing separate checklist blocks to visually merge.
3. **Copy-Paste Output Issue**: When copying the highlighted checklist items from the preview, the clipboard would receive raw DOM strings (such as Material Icon text like `check_box_outline_blank` and missing checkboxes `- [ ]`).

## Solution

1. **Enable Text Highlighting & Context Menu (fixes #8407)**: Updated `clickPreview` in `InlineMarkdownComponent` to ignore click actions if the user has an active text selection or if the click was part of a drag movement (mouse drag distance > 5px). This enables the user to highlight text and use the right-click context menu normally without popping into edit mode.
2. **List Blank Line Preprocessing**: Updated `preprocessMarkdown` in `marked-options-factory.ts` to detect blank lines between checklist items and split them by inserting matching indentation with HTML comments `<!-- -->`. This keeps tight lists looking tight while preserving gaps between list groups.
3. **DOM-to-Markdown Copy Interceptor**: Implemented a `@HostListener('copy')` handler in `InlineMarkdownComponent` that intercepts clipboard copying within the preview pane. It parses the selected DOM fragment recursively using a custom `domToMarkdown` serializer and writes clean markdown (with correct `- [ ]` and `- [x]` syntax and nested indentation) directly to the clipboard.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/electron/main-window.ts
+++ b/electron/main-window.ts
@@ -5,6 +5,7 @@ import {
   BrowserWindowConstructorOptions,
   ipcMain,
   Menu,
+  MenuItem,
   MenuItemConstructorOptions,
   nativeTheme,
   shell,
@@ -382,6 +383,33 @@ export const createWindow = async ({
     if (input.type === 'keyDown' && input.key === 'F11') {
       event.preventDefault();
       mainWin.setFullScreen(!mainWin.isFullScreen());
+    }
+  });
+
+  // Enable right-click context menu (Cut, Copy, Paste, Select All)
+  mainWin.webContents.on('context-menu', (_event, params) => {
+    const { editFlags, selectionText } = params;
+    const menu = new Menu();
+
+    if (editFlags.canCut) {
+      menu.append(new MenuItem({ label: 'Cut', role: 'cut' }));
+    }
+
+    if (selectionText && selectionText.trim() !== '') {
+      menu.append(new MenuItem({ label: 'Copy', role: 'copy' }));
+    }
+
+    if (editFlags.canPaste) {
+      menu.append(new MenuItem({ label: 'Paste', role: 'paste' }));
+    }
+
+    if (editFlags.canSelectAll) {
+      menu.append(new MenuItem({ type: 'separator' }));
+      menu.append(new MenuItem({ label: 'Select All', role: 'selectAll' }));
+    }
+
+    if (menu.items.length > 0) {
+      menu.popup({ window: mainWin });
     }
   });
 
--- a/electron/main-window.ts
+++ b/electron/main-window.ts
@@ -5,7 +5,6 @@ import {
   BrowserWindowConstructorOptions,
   ipcMain,
   Menu,
-  MenuItem,
   MenuItemConstructorOptions,
   nativeTheme,
   shell,
@@ -383,33 +382,6 @@ export const createWindow = async ({
     if (input.type === 'keyDown' && input.key === 'F11') {
       event.preventDefault();
       mainWin.setFullScreen(!mainWin.isFullScreen());
-    }
-  });
-
-  // Enable right-click context menu (Cut, Copy, Paste, Select All)
-  mainWin.webContents.on('context-menu', (_event, params) => {
-    const { editFlags, selectionText } = params;
-    const menu = new Menu();
-
-    if (editFlags.canCut) {
-      menu.append(new MenuItem({ label: 'Cut', role: 'cut' }));
-    }
-
-    if (selectionText && selectionText.trim() !== '') {
-      menu.append(new MenuItem({ label: 'Copy', role: 'copy' }));
-    }
-
-    if (editFlags.canPaste) {
-      menu.append(new MenuItem({ label: 'Paste', role: 'paste' }));
-    }
-
-    if (editFlags.canSelectAll) {
-      menu.append(new MenuItem({ type: 'separator' }));
-      menu.append(new MenuItem({ label: 'Select All', role: 'selectAll' }));
-    }
-
-    if (menu.items.length > 0) {
-      menu.popup({ window: mainWin });
     }
   });
 
--- a/src/app/ui/inline-markdown/inline-markdown.component.html
+++ b/src/app/ui/inline-markdown/inline-markdown.component.html
@@ -25,6 +25,7 @@
   @if (!isTurnOffMarkdownParsing()) {
     <markdown
       #previewEl
+      (mousedown)="previewMousedown($event)"
       (click)="clickPreview($event)"
       [data]="resolvedMarkdownData"
       [class.preview-while-editing]="isShowEdit()"
--- a/src/app/ui/inline-markdown/inline-markdown.component.scss
+++ b/src/app/ui/inline-markdown/inline-markdown.component.scss
@@ -55,6 +55,15 @@
   -webkit-touch-callout: default !important;
 }
 
+// The checkbox is a Material Icons ligature, so its textContent is the literal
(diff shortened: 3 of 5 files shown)


johannesjo:
Thanks for the thorough work here, @hajiboy95 — the diagnosis in each area is solid. The core #8407 fix is right, and I'd happily take **that** on its own.

Before merging, I want to step back on scope. The PR bundles three separate problems (as its own description lays out), and only the first is #8407:

1. `fd5f302` — the selection/drag-aware click guard **+** the Electron right-click context-menu handler → **the actual #8407 fix** ✅
2. `2aaee81` + `1af98e84` — checklist group-spacing via markdown preprocessing
3. `e2f7034` — a DOM→Markdown serializer that intercepts copy

Items (2) and (3) are the same pattern — a hand-rolled converter that fights the medium to work around a rendering choice — and I'd rather settle each approach in its own PR than lock it into `master`, where the edge cases become support load. I've opened #8544 and #8545 to track them.

**On (2) — checklist spacing (#8544).** Markdown intentionally merges blank-line-separated items into one loose list; the regex injects fake HTML to force a split, which is inherently a hack. It has already produced several edge cases within this PR:

- The `1af98e84` backtracking fix is **correct** — I reproduced the ReDoS (a checklist line followed by a few dozen blank/whitespace lines and a non-matching line drives the old regex from sub-millisecond into tens of seconds; the new one stays sub-millisecond). Good catch. But the vulnerable regex was introduced two commits earlier *in this same PR*, so it never reached `master` — there's no shipped hole and nothing forcing it in under time pressure.
- **Multi-gap:** `A␤␤B␤␤C` only splits the first gap (the match consumes the second item; a lookahead fixes it and uses fewer capture groups).
- **CRLF:** the new `[ \t]` class stops splitting `\r\n`-separated lists that the old `\s` one handled (`[ \t\r]` would restore parity and stay linear).
- The code comment still says "inserting an HTML comment" but the code now inserts `<div></div>`.

Worth checking whether plain CSS on loose lists gets most of the look with none of this.

**On (3) — copy serializer (#8545).** Root cause: the checkbox is a Material Icons *ligature* span, so its `textContent` is the literal `check_box_outline_blank`, and the preview SCSS forces `user-select: text !important` on everything — so those words are deliberately selectable and that's what pollutes the clipboard. `domToMarkdown` re-serializes the whole selection to dodge that, but it's a partial `turndown` reimplementation: no escaping (a label with `*` / `_` / `[` round-trips wrong), structural tags flatten and `` is dropped, partial selections lose nested-list indentation, and — because the `copy` listener is unconditional — it now governs **all** preview copying, not just checklists. Making the icon glyph non-selectable (`user-select: none`, or a `::before` pseudo-element) removes the garbage at the source, and the real markdown already lives in the editor.

**Suggestion:** let's land the #8407 fix on its own, and move (2) and (3) to separate PRs against #8544 / #8545 so we can pick the right approach without blocking the bug fix.

One thing to confirm on the core fix: #8407 was filed as a request for an *explicit* edit/view toggle, and this solves the same friction more lightly with a selection-aware click guard. I'm fine with that direction — just flagging it differs from the literal ask in the issue.

Really appreciate you pushing on all three of these — the redirect is purely about keeping the long-term surface small.

hajiboy95:
@johannesjo thanks for taking a look at it. I'll revert the other changes. 

Considering your last comment: in confused. You suggested to NOT tackle the edit view toggle in the comments instead introduce that selection awareness. Now you're saying that it differs? Not sure what's the direction now...

hajiboy95:
@johannesjo reverted a few changes

johannesjo:
No contradiction — sorry for the muddled wording. Two different "asks" got crossed:

- **The direction we agreed on in the issue** is unchanged: selection-safe view first (step 1), explicit edit toggle deferred until there's demand (step 2). Your PR implements step 1, which is exactly what I want. 👍
- **"Differs from the literal ask"** was just me flagging, for the record, that the *issue's title* literally requested an explicit view/edit toggle — so when this PR closes #8407, it delivers the lighter selection-aware fix instead of that toggle. That's purely a transparency note so it's clear later why the issue closed without a toggle. It's not a request to change anything.

So: keep going exactly as you are. The reverts are perfect — the PR is now cleanly scoped to the core fix, which is the part I'm happy to merge. Thanks again\!

johannesjo:
I pushed one small commit (`4a9966d`) that pulls the Electron right-click context-menu handler back out of `electron/main-window.ts` — just so we're aligned on scope, nothing wrong with the code itself.

Reasoning: the renderer-side selection-safe click guard already fully resolves #8407 — text is selectable and Ctrl/Cmd+C copies in both the web and desktop builds, which is exactly the plan we agreed on in the issue (selection-safe *view* first). The `webContents` context-menu handler was a separate, app-wide addition: it governs every input and view in the desktop app (not just the note preview), and the hardcoded `'Cut'`/`'Copy'`/`'Paste'`/`'Select All'` labels override Electron's automatic OS-localized role labels. A native right-click menu across the whole app is a deliberate product decision with its own surface (spellcheck, links, images, i18n), so if it's wanted it deserves its own PR rather than riding along here.

That leaves this PR cleanly scoped to the #8407 fix:
- the selection/drag-aware click guard + mousedown handler (`inline-markdown.component.ts/.html`)
- the `user-select` view-mode selection support + the checkbox-glyph `user-select: none` mitigation
- the accompanying specs

Thanks again for the thorough work — really appreciate you chasing all the edge cases here. 🙏

johannesjo:
Thank you very much!

PR #8527, opened 2026-06-21
8110 feat(tasks): add task option to inherit parent dates when creating new subtasks

### LINKED ISSUE #8110: 💡 Feature Request: Date Inheritance (Newly created subtasks inherit parent's dates)
### Problem Statement

When breaking down an already-scheduled task into smaller pieces, users have to manually re-enter the due dates and deadlines for every new subtask. Furthermore, if a user decides to turn a scheduled task into a "Container/Folder" for subtasks, those original dates are lost or disconnected from the actual sub-steps. This creates friction and repetitive data entry when trying to detail out a scheduled event.

### Possible Solution

Introduce a simple toggle in the settings: **"Automatically pass parent dates to newly created subtasks."**

When this is enabled, the moment a user clicks "Add Subtask" on a parent task that has a Due Date or Deadline, the newly created subtask will automatically pre-fill with those same dates.

* **Manual Entry:** Manually typing or clicking through the calendar to assign dates to every single new subtask. This is the current behavior and severely slows down the brain-dumping workflow.

### Additional context

This should be an optional toggle (enabled by default, as this feels like the more intuitive workflow). When paired with #8095, this inheritance feature ensures a perfectly smooth transition: the parent passes its date down to the first subtask, and then the parent shifts into a milestone wrapper seamlessly, **preventing tedious data re-entry**.

*(Note for implementation: If #8095 is also implemented, care should be taken regarding the event order. When a child is created and inherits the date, the system should prevent the parent's auto-recalculation logic from redundantly firing or conflicting during that exact moment of creation).*

hajiboy95:
## Problem

- Closes #8110 (Automatically pass parent dates/deadlines to newly created subtasks and prompt when changing parent dates).

## Solution

1. **Subtask Date Inheritance**:
   - Added configuration settings (`isPassParentDatesToSubTasks`, `isPassParentDueDateToSubTasks`, `isPassParentDeadlineToSubTasks`) to control whether new subtasks inherit due dates, deadlines, or both from their parent.
   - Modified `addSubTaskTo` in `TaskService` to inherit dates from parent tasks according to these settings.

2. **Pre-dispatch Confirmation Prompt**:
   - Moved the due date/deadline update prompt to the component layer (`TaskComponent` shortcuts and `DialogScheduleTaskComponent` submit logic). This ensures the prompt is resolved *before* dispatching the reschedule action, keeping the parent task in the Today list until confirmed.
   - Added an `isSkipSubTaskDateUpdatePrompt` flag to action properties to prevent the global effects handler from showing a second, duplicate prompt after dispatch.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

```


johannesjo:
Hi @hajiboy95, thank you so much for putting this together — there's clearly a lot of care here: the settings, the docs, and a genuinely solid set of tests. I really appreciate the effort. 🙏

After thinking it over, I've decided not to move forward with this one, so I'm going to close both the PR and the issue. Let me explain my reasoning so it's not just a silent "no":

- In Super Productivity the **parent task generally carries the schedule**, and subtasks are the steps underneath it. Copying the parent's due date onto every new subtask tends to clutter the Today / scheduled views with duplicates that all point at the same day, and many people use subtasks as a plain checklist that deliberately has no dates of its own. Enabling that by default would change behavior for everyone, which I'm not comfortable with.
- The second half — prompting on every parent date change whether to also update the subtasks — adds a confirmation step to one of the most frequent actions in the app (rescheduling). I'd rather keep that path friction-free.

There **is** a real workflow in your original issue that I find reasonable: breaking a task that has a **deadline** into steps, and wanting those steps to share that deadline. If I ever revisit this, I think the cleaner approach would be for subtasks to *visually inherit* the parent's deadline when they don't have one of their own (a computed fallback), rather than copying it into each subtask's state — that avoids the divergence and removes the need for any prompts.

Please don't let this discourage you from contributing — the craftsmanship of the PR itself was great, and I'd be very happy to see you pick up other issues. Thanks again for the work, and for understanding! 🙏

PR #8529, opened 2026-06-21
feat(undo): add local undo redo for task operations

Zinko0:
## Addressing the code review 

This revision keeps undo local-only while persisting and syncing the compensating actions required to reproduce the resulting task state on other devices.
This is related to issue #7244  
and the continuation of the PR #7842  

## Solution

- Only registry-supported task operations are added to the undo stack, so unrelated persisted operations no longer block Ctrl+Z.

- Undo snapshots are captured before reducers run and stored only in local in-memory undo state, keyed by operation id. They are not persisted in IndexedDB or sent through SuperSync.

- Compensating actions skip snapshot capture and are persisted/synced without being added back to the undo stack.

- Added a re-entrancy guard to undo/redo and ignores repeated Ctrl+Z key events.

- Replaced stale redo-stack behaviour with a one-shot snackbar restore. Redo dispatches the original action as a fresh operation, preserving normal op-log ordering.

- Clears undo state after full data imports via loadAllData.

- Restored the existing “Go to task” action for task-created snacks; Undo is available through Ctrl+Z and the dedicated undo feedback.

- Uses translated undo/redo labels and translated error messages.

- Reused the canonical operation payload extraction and operation-to-action conversion paths.

- Fixed update undo for previously absent task fields by removing those fields instead of assigning undefined.

- Demoted the periodic operation-write log from normal to verbose.

- Added filterNonCompensatingAction() to relevant task effects.

To prevent regressions, we also added a local ESLint rule: any dispatching effect that listens to TaskSharedActions.* must use filterNonCompensatingAction(). This is intentionally broader than the current undoable-action list, so future task undo support is protected without maintaining a second allow-list. Effects with { dispatch: false } remain exempt because they cannot emit phantom NgRx operations. Please let us know whether this is the enforcement approach you had in mind.

We also changed the undo order so the compensating action is dispatched before the undo stack is mutated. This guarantees that the compensation has been applied by the reducer before the operation moves from the undo stack. It does not wait for the asynchronous op-log persistence to succeed: adding that confirmation/rollback path would require a larger change to the op-log write contract, which we intentionally kept out of this revision.
  
## Type of Change

- [ ] Bug fix
- [ X ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ X ] I have run `npm run checkFile` on changed `.ts`/`.scss` files
- [ X] I have added tests for my changes (if applicable)
- [ X] Existing tests still pass
- [ X] My commit messages follow the Angular format (`type(scope): description`)


johannesjo:
I've been going back and forth on this the whole day, really unsure what to do with it.

@Zinko0 @Sofiaaleixo – first, I owe you an apology: you flagged early that proper undo was a much bigger architectural change than planned, and I kept giving the go-ahead anyway and even encouraged the direction. I should have made my concerns about this clearer and had a more thourough look at the long term costs this would introduce.

I'm not decided yet and I'd rather talk this through than hand down a verdict – but I want to be honest about where I'm leaning so we're discussing the real thing.

I'm leaning against merging this, the reason being that it's a large amount of code, to do undo properly across all the cases it will have to grow significantly more, and all of it lives in the most crucial, sync-critical parts of the app. I'm almost certain that combination means an ongoing stream of maintenance and subtle, hard-to-debug user-facing issues for a long time before it stabilizes – if it ever fully does. That's a long-term cost I'm very wary of taking on, even though the model is right and the work is careful. And currently it is missing some crucial parts that would make the experience seamless, most importantly redo functionality. Another important aspect is that there hasn't really been a strong demand for this to justify the costs involved.

So before anything's decided, I'd like to discuss it with you. I think there are 4 options:

1. Ship a much leaner version of this
2. Do a more complete version including redo functionality #8554
3. Drop this completely
4. Park this until a bigger demand for the feature arises and then do the more complete version #8554

But before I decide I want your take first – you've been deep in this and may see options or trade-offs that I'm missing.

Sofiaaleixo:
Thanks for taking the time to think this through so carefully, and we really appreciate you opening the discussion rather than just making the decision outright.

After giving it a lot of thought ourselves, I think we've come to the same conclusion. While we still believe the approach is sound, we agree that the long-term maintenance cost and complexity—especially in such a sync-critical part of the app—are hard to justify right now, particularly without redo and without a strong demand for the feature.

Because of that, we've decided that we're okay with dropping it for now. It's definitely disappointing after the work that went into it, but I think it's the right call given the trade-offs. If there's ever stronger demand in the future, we'd be happy to revisit it with a more complete design.

Thanks again for taking the time to explain your reasoning—it made the decision a lot easier to understand. And I'm sorry for taking so long to reply.

johannesjo:
Thank you very much for your understanding. I am closing this then.

johannesjo:
Closing this as we agreed in the thread above — sorry it stayed open so long after I said I would. Thanks a lot for the work you put into it.

PR #8546, opened 2026-06-22
feat(plugins): generic OAuth hooks for issue-provider plugins

oonid:
## Problem
  
OAuth-based issue-provider plugins that need an **exact, pre-registered redirect
URI** can't work on stock Super Productivity today: `OAuthFlowConfig` can't
declare an exact redirect, the Electron loopback always binds a random port, and
there's no generic way for a user to supply their own OAuth app credentials. 
There's also a small bug: merging a *partial* `pluginConfig` update drops the
omitted keys.

Context: follow-up to the discussion on #8507 — these are the small, generic,
provider-agnostic OAuth host hooks you offered to take as a separate trimmed PR.
(The Basecamp provider that motivated them will ship as a community plugin, not
bundled.)
  
## Solution
  
Provider-agnostic plugin-framework hooks — no provider-specific code:
  
- **`OAuthFlowConfig.redirectUri`** — a plugin may declare an exact callback; the
  host uses it for both the authorization request and the token exchange.
- The **Electron loopback honors a plugin-requested fixed port** and rejects with
  a clear message when that port is already in use (instead of a raw `EADDRINUSE`).
- **Bring-your-own OAuth app**: user-supplied `clientId`/`clientSecret`/`redirectUri`
  from plugin config override the plugin's embedded `oauthConfig`.
- **Fix**: merging a partial `pluginConfig` update preserves omitted keys and
  deep-merges nested objects (e.g. `twoWaySync`).
    
12 files. Unit tests added for the OAuth bridge (redirect/port/web-client paths),
the config-merge util, and the OAuth-overrides util.
  
## Type of Change
- [x] New feature
- [x] Bug fix
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other
  
## Checklist
- [x] My commit messages follow the Angular format
- [ ] Documentation/wiki (n/a — internal plugin-API hooks)

### DIFF
--- a/electron/electronAPI.d.ts
+++ b/electron/electronAPI.d.ts
@@ -268,7 +268,7 @@ export interface ElectronAPI {
   consumePluginNodeExecutionApi(): PluginNodeExecutionElectronApi | null;
 
   // Plugin OAuth
-  pluginOAuthPrepare(): Promise<{ port: number }>;
+  pluginOAuthPrepare(port?: number): Promise<{ port: number }>;
   pluginOAuthStart(url: string): void;
   onPluginOAuthCb(
     listener: (data: { code?: string; error?: string; state?: string }) => void,
--- a/electron/plugin-oauth.ts
+++ b/electron/plugin-oauth.ts
@@ -1,6 +1,10 @@
 import { BrowserWindow, ipcMain, shell } from 'electron';
 import { createServer, Server } from 'http';
 import { IPC } from './shared-with-frontend/ipc-events.const';
+import {
+  OAUTH_LOOPBACK_PORT_MIN,
+  OAUTH_LOOPBACK_PORT_MAX,
+} from './shared-with-frontend/oauth-loopback.const';
 import { log } from 'electron-log/main';
 
 const LOOPBACK_HOST = '127.0.0.1';
@@ -46,12 +50,12 @@ export const initPluginOAuth = (mainWin: BrowserWindow): void => {
         if (requestedPort !== undefined) {
           if (
             !Number.isInteger(requestedPort) ||
-            requestedPort < 1024 ||
-            requestedPort > 65535
+            requestedPort < OAUTH_LOOPBACK_PORT_MIN ||
+            requestedPort > OAUTH_LOOPBACK_PORT_MAX
           ) {
             reject(
               new Error(
-                `Invalid OAuth loopback port ${requestedPort}; must be an integer in [1024, 65535].`,
+                `Invalid OAuth loopback port ${requestedPort}; must be an integer in [${OAUTH_LOOPBACK_PORT_MIN}, ${OAUTH_LOOPBACK_PORT_MAX}].`,
               ),
             );
             return;
@@ -89,7 +93,9 @@ export const initPluginOAuth = (mainWin: BrowserWindow): void => {
         });
 
         server.on('error', (err) => {
-          server.close();
+          cleanupServer();
+          // Note: EADDRINUSE message interpolates port which is 0 only in the system-assigned
+          // (no requested port) case — where EADDRINUSE effectively cannot occur.
           if ((err as NodeJS.ErrnoException).code === 'EADDRINUSE') {
             reject(
               new Error(
--- a/electron/plugin-oauth.ts
+++ b/electron/plugin-oauth.ts
@@ -34,62 +34,80 @@ export const initPluginOAuth = (mainWin: BrowserWindow): void => {
   // Prepare: start a loopback HTTP server and return the port.
   // Google Desktop OAuth requires http://127.0.0.1:<port> redirect URIs
   // and blocks embedded webviews, so we open the system browser instead.
-  ipcMain.handle(IPC.PLUGIN_OAUTH_PREPARE, async (): Promise<{ port: number }> => {
-    cleanupServer();
+  ipcMain.handle(
+    IPC.PLUGIN_OAUTH_PREPARE,
+    async (_event, requestedPort?: number): Promise<{ port: number }> => {
+      cleanupServer();
 
-    return new Promise<{ port: number }>((resolve, reject) => {
-      let handled = false;
+      return new Promise<{ port: number }>((resolve, reject) => {
+        let handled = false;
 
-      const server = createServer((req, res) => {
-        if (handled) {
+        const server = createServer((req, res) => {
+          if (handled) {
+            // eslint-disable-next-line @typescript-eslint/naming-convention
+            res.writeHead(200, { 'Content-Type': 'text/html' });
+            res.end(SUCCESS_HTML);
+            return;
+          }
+          handled = true;
+
+          const url = new URL(req.url!, `http://${LOOPBACK_HOST}`);
+          const code = url.searchParams.get('code');
+          const error = url.searchParams.get('error');
+          const state = url.searchParams.get('state');
+
           // eslint-disable-next-line @typescript-eslint/naming-convention
           res.writeHead(200, { 'Content-Type': 'text/html' });
           res.end(SUCCESS_HTML);
-          return;
-        }
-        handled = true;
 
-        const url = new URL(req.url!, `http://${LOOPBACK_HOST}`);
-        const code = url.searchParams.get('code');
-        const error = url.searchParams.get('error');
-        const state = url.searchParams.get('state');
+          mainWin.webContents.send(IPC.PLUGIN_OAUTH_CB, { code, error, state });
 
-        // eslint-disable-next-line @typescript-eslint/naming-convention
-        res.writeHead(200, { 'Content-Type': 'text/html' });
-        res.end(SUCCESS_HTML);
+          // Re-focus the main window after auth completes
+          if (!mainWin.isDestroyed()) {
+            mainWin.show();
+            mainWin.focus();
+          }
 
-        mainWin.webContents.send(IPC.PLUGIN_OAUTH_CB, { code, error, state });
+          cleanupServer();
+        });
 
-        // Re-focus the main window after auth completes
-        if (!mainWin.isDestroyed()) {
-          mainWin.show();
-          mainWin.focus();
-        }
+        const port =
+          typeof requestedPort === 'number' &&
+          Number.isInteger(requestedPort) &&
+          requestedPort > 0
+            ? requestedPort
+            : 0;
 
-        cleanupServer();
+        server.listen(port, LOOPBACK_HOST, () => {
+          const addr = server.address();
+          if (addr && typeof addr !== 'string') {
+            loopbackServer = server;
+            oauthTimeoutId = setTimeout(() => {
+              log('Plugin OAuth: Timeout – closing abandoned loopback server');
+              cleanupServer();
+            }, OAUTH_TIMEOUT_MS);
+            log(`Plugin OAuth: Loopback server listening on port ${addr.port}`);
+            resolve({ port: addr.port });
+          } else {
+            server.close();
+            reject(new Error('Failed to start OAuth loopback server'));
+          }
+        });
+
+        server.on('error', (err) => {
+          if ((err as NodeJS.ErrnoException).code === 'EADDRINUSE') {
+            reject(
+              new Error(
+                `OAuth loopback port ${port} is already in use. Close the app using it and try again.`,
+              ),
+            );
+          } else {
+            reject(err);
+          }
+        });
       });
-
-      server.listen(0, LOOPBACK_HOST, () => {
-        const addr = server.address();
-        if (addr && typeof addr !== 'string') {
-          loopbackServer = server;
-          oauthTimeoutId = setTimeout(() => {
-            log('Plugin OAuth: Timeout – closing abandoned loopback server');
-            cleanupServer();
-          }, OAUTH_TIMEOUT_MS);
-          log(`Plugin OAuth: Loopback server listening on port ${addr.port}`);
-          resolve({ port: addr.port });
-        } else {
-          server.close();
-          reject(new Error('Failed to start OAuth loopback server'));
-        }
-      });
-
-      server.on('error', (err) => {
-        reject(err);
-      });
-    });
-  });
+    },
+  );
 
   // Open the auth URL in the system browser (not an embedded webview).
   // Google blocks OAuth in embedded browsers (Electron BrowserWindow).
--- a/electron/plugin-oauth.ts
+++ b/electron/plugin-oauth.ts
@@ -42,6 +42,23 @@ export const initPluginOAuth = (mainWin: BrowserWindow): void => {
       return new Promise<{ port: number }>((resolve, reject) => {
         let handled = false;
 
+        let port = 0;
+        if (requestedPort !== undefined) {
+          if (
+            !Number.isInteger(requestedPort) ||
+            requestedPort < 1024 ||
+            requestedPort > 65535
+          ) {
+            reject(
+              new Error(
+                `Invalid OAuth loopback port ${requestedPort}; must be an integer in [1024, 65535].`,
+              ),
+            );
+            return;
+          }
+          port = requestedPort;
+        }
+
         const server = createServer((req, res) => {
           if (handled) {
             // eslint-disable-next-line @typescript-eslint/naming-convention
@@ -71,12 +88,18 @@ export const initPluginOAuth = (mainWin: BrowserWindow): void => {
           cleanupServer();
         });
 
-        const port =
-          typeof requestedPort === 'number' &&
-          Number.isInteger(requestedPort) &&
-          requestedPort > 0
-            ? requestedPort
-            : 0;
+        server.on('error', (err) => {
+          server.close();
+          if ((err as NodeJS.ErrnoException).code === 'EADDRINUSE') {
+            reject(
+              new Error(
+                `OAuth loopback port ${port} is already in use. Close the app using it and try again.`,
+              ),
+            );
+          } else {
+            reject(err);
+          }
+        });
 
         server.listen(port, LOOPBACK_HOST, () => {
           const addr = server.address();
@@ -91,18 +114,6 @@ export const initPluginOAuth = (mainWin: BrowserWindow): void => {
           } else {
             server.close();
             reject(new Error('Failed to start OAuth loopback server'));
-          }
-        });
-
-        server.on('error', (err) => {
-          if ((err as NodeJS.ErrnoException).code === 'EADDRINUSE') {
-            reject(
-              new Error(
-                `OAuth loopback port ${port} is already in use. Close the app using it and try again.`,
-              ),
-            );
-          } else {
-            reject(err);
           }
         });
       });
--- a/electron/preload.ts
+++ b/electron/preload.ts
@@ -270,7 +270,8 @@ const ea: ElectronAPI = {
   },
 
   // Plugin OAuth
-  pluginOAuthPrepare: () => _invoke('PLUGIN_OAUTH_PREPARE') as Promise<{ port: number }>,
+  pluginOAuthPrepare: (port?: number) =>
+    _invoke('PLUGIN_OAUTH_PREPARE', port) as Promise<{ port: number }>,
   pluginOAuthStart: (url: string) => _send('PLUGIN_OAUTH_START', { url }),
   onPluginOAuthCb: (
     listener: (data: { code?: string; error?: string; state?: string }) => void,
--- a/electron/shared-with-frontend/oauth-loopback.const.ts
+++ b/electron/shared-with-frontend/oauth-loopback.const.ts
@@ -1,0 +1,4 @@
+// Allowed range for the OAuth loopback redirect port. Shared between the renderer
+// (redirectUri validation) and the Electron main process (IPC bind) so the two never drift.
+export const OAUTH_LOOPBACK_PORT_MIN = 1024;
+export const OAUTH_LOOPBACK_PORT_MAX = 65535;
--- a/packages/plugin-api/src/types.ts
+++ b/packages/plugin-api/src/types.ts
@@ -417,8 +417,13 @@ export interface OAuthFlowConfig {
   /** Additional query parameters to append to the authorization URL (e.g. access_type, prompt). */
   extraAuthParams?: Record<string, string>;
   /**
-   * Optional redirect URI override for providers or user-supplied OAuth apps
-   * that require an exact pre-registered callback instead of the host default.
+   * Optional redirect URI override for the desktop (Electron) loopback flow — e.g.
+   * a user-supplied OAuth app that requires an exact pre-registered
+   * `http://127.0.0.1:<port>/...` callback instead of the host default.
+   *
+   * Desktop-only: web and native each have a single valid callback (the host's
+   * `/assets/oauth-callback.html` page and the app's fixed custom scheme), so this
+   * field is ignored (stripped) on those platforms and the platform default is used.
    */
   redirectUri?: string;
 }
--- a/packages/plugin-api/src/types.ts
+++ b/packages/plugin-api/src/types.ts
@@ -408,6 +408,11 @@ export interface OAuthFlowConfig {
   scopes: string[];
   /** Additional query parameters to append to the authorization URL (e.g. access_type, prompt). */
   extraAuthParams?: Record<string, string>;
+  /**
+   * Optional redirect URI override for providers or user-supplied OAuth apps
+   * that require an exact pre-registered callback instead of the host default.
+   */
+  redirectUri?: string;
 }
 
 export interface OAuthTokenResult {
--- a/packages/plugin-api/src/types.ts
+++ b/packages/plugin-api/src/types.ts
@@ -375,6 +375,14 @@ export interface PluginWorkContextHeaderBtnCfg {
   showFor: ('PROJECT' | 'TAG' | 'TODAY')[];
 }
 
+/**
+ * OAuth configuration for starting an OAuth flow.
+ *
+ * Bring-your-own credentials: an issue-provider plugin may let users override the
+ * clientId / clientSecret / redirectUri at runtime by writing them under
+ * `pluginConfig.oauthOverrides = { clientId?, clientSecret?, redirectUri? }`. These
+ * overrides apply only to the desktop (Electron loopback) flow and are ignored on web/native.
+ */
 export interface OAuthFlowConfig {
   authUrl: string;
   tokenUrl: string;
--- a/src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.ts
+++ b/src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.ts
@@ -55,12 +55,12 @@ import { FormlyFieldConfig, FormlyModule } from '@ngx-formly/core';
 import { MatButton } from '@angular/material/button';
 import { MatIcon } from '@angular/material/icon';
 import { IS_ANDROID_WEB_VIEW } from '../../../util/is-android-web-view';
-import { devError } from '../../../util/dev-error';
 import { IssueLog } from '../../../core/log';
 import { PluginIssueProviderRegistryService } from '../../../plugins/issue-provider/plugin-issue-provider-registry.service';
 import { PluginBridgeService } from '../../../plugins/plugin-bridge.service';
 import { PluginHttpService } from '../../../plugins/issue-provider/plugin-http.service';
 import { OAuthFlowConfig, PluginSyncDirection } from '@super-productivity/plugin-api';
+import { applyPluginOAuthOverrides } from './plugin-oauth-config-overrides.util';
 import { IS_NATIVE_PLATFORM } from '../../../util/is-native-platform';
 // Trello is now a plugin — board selection is a dynamic `loadOptions` select field
 // ClickUp is now a plugin — no built-in config component needed
@@ -72,6 +72,7 @@ import { ISSUE_PROVIDER_COMMON_FORM_FIELDS } from '../common-issue-form-stuff.co
 import { TagService } from '../../tag/tag.service';
 import { ChipListInputComponent } from '../../../ui/chip-list-input/chip-list-input.component';
 import { unique } from '../../../util/unique';
+import { mergeIssueProviderModelUpdates } from './issue-provider-model-merge.util';
 
 @Component({
   selector: 'dialog-edit-issue-provider',
@@ -266,24 +267,7 @@ export class DialogEditIssueProviderComponent {
   }
 
   updateModel(model: Partial<IssueProvider>): void {
-    // NOTE: this currently throws an error when loading issue point stuff for jira
-    try {
-      Object.keys(model).forEach((key) => {
-        if (key !== 'isEnabled') {
-          this.model![key] = model[key];
-        }
-      });
-    } catch (e) {
-      devError(e);
-      const updates: any = {};
-      Object.keys(model).forEach((key) => {
-        if (key !== 'isEnabled') {
-          updates[key] = model[key as keyof IssueProvider];
-        }
-      });
-      this.model = { ...this.model, ...updates };
-    }
-
+    this.model = mergeIssueProviderModelUpdates(this.model, model);
     this.isConnectionWorks.set(false);
   }
 
@@ -365,9 +349,10 @@ export class DialogEditIssueProviderComponent {
     if (!pluginId) {
       return;
     }
+    const effectiveOAuthConfig = this._withPluginOAuthOverrides(oauthConfig);
     this.isOAuthConnecting.set(true);
     try {
-      await this._pluginBridge.startOAuthFlow(pluginId, oauthConfig);
+      await this._pluginBridge.startOAuthFlow(pluginId, effectiveOAuthConfig);
     } catch (e) {
       const detail = (e instanceof Error ? e.message : String(e))
         .replace(/\s+/g, ' ')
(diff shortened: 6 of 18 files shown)


johannesjo (commented):
## Review — generic OAuth host hooks

Thanks for the trimmed, provider-agnostic split — the shape is right and the `pluginConfig` partial-merge bug fix is a real, well-tested improvement. I ran a multi-perspective pass (correctness / security / architecture / performance / simplicity) and verified each item against the code. A few things to address before merge, then a couple of nits.

### 🔴 Should fix

**1. `redirectUri` override is honored verbatim with no validation → silent 5-minute hang on configs the transport can't service.**
`prepareRedirectUri()` returns the supplied `redirectUri` as-is (`plugin-oauth.service.ts:44`) but the transport only actually services a *loopback* URI on Electron / a *same-origin* URI on web:
- **Electron + non-loopback URI:** `_getElectronLoopbackPort()` returns `undefined`, so an ephemeral-port loopback server is started, yet the auth request advertises the external URI. The IdP redirect never reaches the loopback server → `waitForRedirectCode` hangs until the 5-min timeout (with an orphan server idling).
- **Web + cross-origin URI:** `plugin-oauth-redirect.handler.ts:38` drops any `postMessage` whose `event.origin !== window.location.origin`, so a cross-origin callback never resolves → same hang.

The intended path (loopback fixed port, e.g. `http://127.0.0.1:8976/callback`) traces clean end-to-end — this is purely about failing *fast* on a misconfiguration instead of hanging. Since `redirectUri` is on `OAuthFlowConfig`, it's reachable by any plugin via the public `startOAuthFlow(config)` (`plugin-api.ts:599`), not just the BYO dialog path. Suggest validating in `PluginOAuthService` and rejecting early with a clear message: Electron → `http://127.0.0.1|localhost:<port>`; web → same-origin; native → app scheme; everywhere → known protocol (also closes the `javascript:`/`file:` concern).

**2. BYO `clientId`/`clientSecret` are silently dropped on web/native.** `connectOAuth` applies the user overrides, but `startOAuthFlow`'s platform branch then overwrites them: web forces `clientId: webClientId, clientSecret: undefined` (`plugin-oauth-bridge.service.ts:56-68`); Android/iOS force their platform client IDs when present. So a user who pastes their own credentials on those platforms has them quietly ignored (only `redirectUri` survives all branches). Either gate the BYO fields to Electron, or surface a warning when they won't be used on the current platform.

**3. Main-process port input is under-validated at the IPC boundary.** `electron/plugin-oauth.ts:74` accepts any `Number.isInteger(port) && port > 0`. No upper bound (≥ 65536 makes `server.listen()` throw a synchronous `RangeError` *before* the `error` handler is registered → unfriendly reject) and no lower bound (privileged ports < 1024). The renderer pre-validates via `new URL()`, but `ipcMain.handle` shouldn't trust the renderer. Suggest clamping to a sane range (reject `< 1024` and `>= 65536`) and/or registering the `error` handler before `listen()`.

### 🟡 Consider

**4. Secret in synced `pluginConfig`.** A BYO `clientSecret` lands in `IssueProvider.pluginConfig`, which syncs to the configured backend (WebDAV/Dropbox/local-file may be unencrypted) — whereas OAuth *tokens* are deliberately kept local-only. This is consistent with the existing `JiraCfg.password` precedent, so not a new class of risk, but a small UI note that the secret will sync (and keeping it out of any plaintext export) would be good hygiene.

**5. `twoWaySync` deep-merge special-case in `issue-provider-model-merge.util.ts`.** Hardcoding one nested key means the *next* nested config object partially-updated will silently regress the very key-drop bug this PR fixes. Also, in the current UI only `customCfgCmpSave` emits a partial `pluginConfig`; the formly path emits the whole model — so it's worth confirming a real caller emits a *partial* `twoWaySync` at all. Either drop the special-case (keep the top-level shallow merge, which is the actual fix) or make `mergePluginConfig` recurse generically over plain-object values.

### 🟢 Nits
- **Revert the namespace import** `import * as pluginOAuthTokenStore` (`plugin-oauth-bridge.service.ts:5`) → the spec imports the named exports directly and never spies on the namespace, so it's churn in a sync-adjacent file; named imports read cleaner.
- **Stale comment** at `plugin-oauth-bridge.service.ts:40` still says `getRedirectUri()` (now `prepareRedirectUri()`); a one-line JSDoc noting it has a side effect (starts the loopback server on Electron) would help.
- **`EADDRINUSE`/error branch** (`plugin-oauth.ts:97-107`) doesn't `server.close()` the failed server — harmless today (next `cleanupServer()` covers it), but tidier to close it.
- **Loopback host:** `_getElectronLoopbackPort` accepts `localhost`, but the server binds only `127.0.0.1` — a `localhost` URI can resolve to `::1` first and miss the IPv4 server. Restrict to `127.0.0.1` or bind dual-stack.

### ✅ Checked and fine
- Reassigning `this.model` to a new reference on every `(modelChange)` is **not** a change-detection/form regression — `immutable: true` is set globally (`formly-config.module.ts:124`), so formly already deep-clones and rebuilds per settled change; the new reference is the intended immutable pattern (the old in-place mutation was the anti-pattern).
- `redirect_uri` stays consistent between the auth request and token exchange.
- Adding optional `OAuthFlowConfig.redirectUri` is forward-compatible; the app type-checks against `packages/plugin-api/src`, so no dist rebuild needed.
- Logging on the OAuth paths is clean (no secrets/config values leaked via `getSafeErrorLogMeta` / `getSafeCfgUpdateLogMeta`).

Overall: solid, correctly-scoped work — happy to merge once **#1–#3** are handled (#1 being the main one). Nice tests on the merge and overrides utils.

oonid:
Thanks for the thorough multi-perspective pass, every point was actionable. 

All three 🔴 items are fixed, plus #5 and the nits; #4 noted below.
  
🔴 #1 : redirectUri validated, fails fast. prepareRedirectUri() now validates a supplied redirectUri up front and
  rejects with a clear message instead of hanging: Electron → must be a http://127.0.0.1:<port> loopback; 
  web → same-origin; native → app custom scheme; and it rejects unknown protocols (closes the 
  javascript:/file: concern). Since it's reachable by any plugin via startOAuthFlow(config), the check lives 
  in PluginOAuthService. I also restricted the desktop loopback host to 127.0.0.1 only (dropped localhost), 
  closing the localhost→::1 nit.
  
🔴 #2 : BYO creds on web/native. The platform branch still forces the public/platform client (web can't carry 
  a secret), but no longer silently: startOAuthFlow now PluginLog.warns when a clientSecret would be dropped 
  on web/Android/iOS.

🔴 #3 : IPC port hardening. pluginOAuthPrepare now rejects a requested port outside [1024, 65535] early 
  (no privileged ports; no synchronous RangeError from listen() on ≥65536), registers the error handler before
  listen(), and server.close()s the failed server.

🟡 #4 : secret syncs. Agreed, same class as the existing JiraCfg.password, not new. The BYO clientSecret field 
  lives in the separate community plugin, so I'll add the "this will sync" note to its field description there rather 
  than in core.

🟡 #5 : merge generality. Dropped the hardcoded twoWaySync special-case; mergePluginConfig now recurses
  generically over plain-object values, so any nested config object deep-merges, no future key-drop regression. 
  The top-level shallow merge (the actual fix) is unchanged.

🟢 Nits. Reverted to named token-store imports; fixed the stale getRedirectUri() comment 
  (now prepareRedirectUri(), with a note that it starts the loopback server on Electron); the EADDRINUSE branch
  now closes the server.


New tests cover the redirectUri validation (web branch + malformed), the web client-secret warning, and 
the generic nested merge. (The Electron/native validation branches hinge on build-time platform constants 
the karma harness can't cleanly toggle, so those paths are covered by review rather than unit tests.)
  
Thanks again, happy to iterate further.

johannesjo:
Reviewed the diff with a multi-perspective pass (correctness, security, architecture, simplicity, alternatives, performance) and ran the new unit specs locally — all green (model-merge 6, overrides 2, oauth.service 36, oauth-bridge 7). No blocking/critical defect; the flow is well-guarded (PKCE + `state` enforced, HTTPS enforced on auth/token URLs, userinfo-host bypass like `http://127.0.0.1:9999@evil.com` correctly rejected, secrets never logged, tokens kept local-only). A few things worth addressing before merge:

### Important

1. **User-supplied `clientSecret` syncs in plaintext.** BYO credentials land in `IssueProvider.pluginConfig`, which is synced (op-log + backups) in cleartext — yet the resulting *tokens* are deliberately kept local-only. That split is surprising, and a confidential secret (some providers' OAuth-app secrets genuinely are) shouldn't leave the device implicitly. It partly follows existing precedent (Jira/GitLab tokens), so it may be acceptable — but please make it a *conscious* choice: at minimum a UI warning next to the secret field ("stored and synced unencrypted; use only desktop/installed-app credentials"), and ideally store BYO `clientSecret` in the same local-only token store as the tokens rather than in synced `pluginConfig`. This is the hardest part to walk back later (it's a sync-format commitment), so worth deciding now.

2. **Native `redirectUri` validation is a protocol denylist, not an allowlist** (`plugin-oauth.service.ts:127-134`). It rejects only `http/https/javascript/file`, so any custom scheme passes — but the deep-link callback is only captured for the hardcoded app scheme. Two consequences: (a) a plugin-supplied non-app scheme silently times out (the "exact pre-registered callback" the field promises can't actually be honored on native), and (b) a compromised plugin could point the auth-code redirect at another installed app's scheme (PKCE limits the damage, but it's an unnecessary capability). Suggest allowlisting the app's own scheme on native, or rejecting a `redirectUri` override on native/web outright and documenting `redirectUri` as desktop-loopback-only.

3. **`updateModel` deep-merge changes save semantics for *all* issue providers, not just OAuth.** It's the single sink for both `formlyModelChange` (full model) and `customCfgCmpSave` (partial). Two side effects worth a test: nested `pluginConfig` keys can no longer be cleared by omission (only by explicit empty value), and a cascading parent→child select (e.g. account → bucket → list) that clears the parent and re-emits without the child key will now **retain the stale child value**. Please add a test exercising that cascading-clear path, and consider scoping the merge to the partial (`customCfgCmpSave`) path only. Minor: the recursion is more general than needed — the only nested object in practice is one-level `twoWaySync`, and the `nestedSettings` spec tests a shape that doesn't exist in the app (the PR text's "special-cases twoWaySync" isn't what the code does — it merges *all* nested objects).

### Suggestions

4. **Redirect/port validation is triplicated** — `_validateRedirectUri`, `_getElectronLoopbackPort`, and the Electron-main range check all re-parse the same URL with slightly different rules. `_validateRedirectUri` runs first and already guarantees `http://127.0.0.1:<port>`, so `_getElectronLoopbackPort`'s body is effectively unreachable; it can collapse to `const loopbackPort = redirectUri ? Number(new URL(redirectUri).port) : undefined;`. While simplifying: (a) align the accepted port range — `:443`/`:500` pass the renderer but get a confusing rejection from main's `[1024,65535]`; (b) reject port `0` — `new URL('http://127.0.0.1:0/cb').port === "0"` passes `_validateRedirectUri` but `_getElectronLoopbackPort` returns `undefined`, so the server binds a *random* port while the returned redirectUri keeps `:0` (the IdP then redirects to an unconnectable address).

5. **BYO `clientId` is silently dropped on mobile/web.** The dialog applies BYO overrides, then the bridge re-overrides `clientId` with `mobile/ios/webClientId`. So a user-entered client id only takes effect on Electron, with no feedback (the new warnings only fire for `clientSecret`). Either consolidate the override precedence into one place or warn when a BYO value is discarded.

6. **Is the BYO-via-`pluginConfig` path needed in the host?** Nothing shipped writes `clientId/clientSecret/redirectUri` into `pluginConfig` (no host UI, no i18n strings) — only an out-of-tree plugin would, and a plugin can already merge its own user-entered values into the `OAuthFlowConfig` it passes to `startOAuthFlow`. If the goal is purely the generic host hooks, consider whether `applyPluginOAuthOverrides` belongs here at all, vs. keeping only the plugin-declared `OAuthFlowConfig.redirectUri`. (Flagging given `redirectUri` becomes a permanent public plugin-API field.)

### Nits
- `electron/plugin-oauth.ts`: a late `error` after a successful `listen` calls `server.close()` directly instead of `cleanupServer()`, leaving `loopbackServer`/`oauthTimeoutId` dangling until the timeout fires (harmless, but `cleanupServer()` is tidier). The EADDRINUSE message interpolates `port`, which is `0` when no port was requested — can't actually fire for port 0, but a one-line comment would clarify.
- Optional `__proto__`/`constructor` key-skip in the merge as defense-in-depth (not exploitable today since `pluginConfig` comes from form models, not raw `JSON.parse`).

Nice cleanups in here too — replacing the `try/catch` + `any` `updateModel` with a tested pure util, registering the loopback `error` handler before `listen`, and the reserved-OAuth-param filter are all good.

oonid:
Hi @johannesjo ,
Thank you for such a careful, multi-perspective review, easily one of the most useful I've had. 

The security and architecture framing (the sync-format and "permanent public API" lenses in particular) is exactly the right altitude, and you've convinced me the BYO-in-host path is more than this PR should carry.
I'd like to propose trimming to the truly-generic minimum and would appreciate a sanity check on the direction before I push:

1. Drop applyPluginOAuthOverrides (the BYO-via-pluginConfig path) from the host entirely, keeping 
  only OAuthFlowConfig.redirectUri, the Electron loopback fixed-port support, and the pluginConfig 
  partial-merge fix. Nothing in-tree writes those BYO keys, and a plugin that genuinely needs per-user 
  credentials can merge them into the config it passes to startOAuthFlow itself. 
  That removes the speculative surface, the plaintext-clientSecret-sync question, and the silent clientId 
  drop in one move. Does that match what you'd want to support long-term?
  
2. Make redirectUri desktop-loopback-only — validate http://127.0.0.1:<port> on Electron and reject any 
  override on web/native (documented as loopback-only), instead of the protocol denylist. 
  That closes the native allowlist gap and the "exact pre-registered callback can't actually be honored 
  on native" issue. Acceptable or would you prefer allowlisting the app's own scheme on native?

3. Scope the pluginConfig merge to the partial (customCfgCmpSave) path and make it a one-level merge 
  (dropping the deep recursion). The full-model formly path stays wholesale, so clear-by-omission, 
  including the cascading account→bucket→list selects, keeps working; 
  I'll add a test exercising that cascading-clear path. 
  (You're right the deep recursion + the nestedSettings test over-reached, both go away.) Sound right?
  
I'll also fold in the nits: cleanupServer() on the post-listen error, a clarifying comment on 
the port-0 EADDRINUSE case, and the optional __proto__/constructor key-skip in the merge.

For context on #1: the Basecamp provider that motivated this ships as a community plugin with its own 
rotatable, author-owned OAuth app, so dropping host BYO just means it uses that bundled app rather 
than per-user credentials, SP core never ships or syncs a secret either way.

Thanks again for the depth here, happy to adjust any of these.

johannesjo:
**Thanks for trimming this down to provider-agnostic hooks — it reviews cleanly.** I ran a multi-perspective review (correctness, security, architecture, performance, simplicity). No blocking issues; PKCE+state, the `RESERVED_OAUTH_PARAMS` guard, HTTPS re-validation at the IPC boundary, and keeping tokens local are all correct, and the redirect-URI host check correctly rejects the userinfo/host-confusion class (`127.0.0.1@evil.com`, `127.0.0.1.evil.com`). Nice that the removed Jira `try/catch` is genuinely *fixed* by the pure merge (no more frozen-state throw) rather than masked. A few suggestions:

### Worth addressing before merge

**1. Loopback server/timer can leak on a post-`listen` error** — `electron/plugin-oauth.ts:91-102`
The `server.on('error')` handler calls bare `server.close()`. For the common `EADDRINUSE` case the error fires *before* `listen` succeeds, so nothing leaks. But once the `listen` callback has assigned `loopbackServer` and scheduled `oauthTimeoutId` (lines 107-111), a later runtime socket error would close the socket while leaving the module ref dangling and the 5-min timer pending. Suggest calling `cleanupServer()` instead of `server.close()` so it's correct regardless of when the error fires.

**2. User-supplied `clientSecret` syncs in plaintext via `pluginConfig`** — `plugin-oauth-config-overrides.util.ts`
The embedded `OAuthFlowConfig.clientSecret` carries a deliberate doc-warning ("NOT confidential … may be synced to cloud backends"). The new bring-your-own override path reads a secret from `pluginConfig`, which is part of the synced `IssueProvider` entity, with no equivalent warning. This is consistent with how Jira/CalDAV/Nextcloud passwords already sync (and OAuth *tokens* correctly stay local), so it's not a new vulnerability class — but a user following "confidential web app" provider instructions could paste a genuinely sensitive secret that then syncs. Suggest either mirroring the secret-sync warning at the override boundary / field help text, or restricting the override to the loopback+PKCE path (no secret).

### Optional cleanups

**3. Port bounds diverge between renderer and main** — `plugin-oauth.service.ts` accepts any `port > 0` (`_getElectronLoopbackPort`) / any non-empty port (`_validateRedirectUri`), while `electron/plugin-oauth.ts:46-60` rejects outside `[1024, 65535]`. A URI like `http://127.0.0.1:80/cb` passes the renderer then fails in main with a different message. It fails safe, but sharing one bound/constant avoids drift and gives an earlier, clearer error.

**4. `_validateRedirectUri` and `_getElectronLoopbackPort` parse the same URL twice.** Since `_validateRedirectUri` already throws unless the desktop URI is exactly `http://127.0.0.1:<port>`, the loopback re-check in `_getElectronLoopbackPort` is dead on that path. You can parse once and read `Number(new URL(redirectUri).port)`, dropping `_getElectronLoopbackPort` entirely.

**5. `mergePluginConfig` — harden + comment.** I tested it for prototype pollution: **not exploitable** (no global `Object.prototype` pollution, and JSON serialization drops any manipulated prototype before persistence). Still cheap insurance to `continue` on `__proto__`/`constructor`/`prototype`, since it's a generic util. Also worth a one-line note that the merge is intentionally deep only for `pluginConfig`, and nested keys can only be cleared with an explicit value, never by omission — today `twoWaySync` always ships complete (dotted formly keys), so the recursion is currently exercising a hypothetical partial-nested case.

### Nits
- `EADDRINUSE` message reads "port 0" in the random-port branch (effectively unreachable, but minor).
- New redirect/port error strings can surface to users via the `OAUTH_FAILED_WITH_DETAIL` snackbar but are hardcoded English — consistent with existing OAuth strings, so low priority.

Overall: **approve with minor changes.** Well-tested and security-aware; items 1 and 2 are the only ones I'd want resolved first.

oonid:
Thanks @johannesjo , all of the round-2 items are in (latest commit):
  
1. Loopback leak on post-listen error: the error handler now calls cleanupServer() instead of bare server.close(), so a runtime socket error after listen succeeds also clears the module loopbackServer ref and the pending 5-min timer.

2. BYO clientSecret plaintext sync: went with the warn option for now: a doc-warning at the override boundary (applyPluginOAuthOverrides) mirroring the existing OAuthFlowConfig.clientSecret note, plus user-facing field help text in the plugin ("stored and synced like other provider passwords").

Since it's a judgment call, the reasoning: the provider this came from (Basecamp/Launchpad) is a genuinely confidential client, it needs a client_secret and has no PKCE, so "restrict to no-secret" would disable per-user BYO for exactly the providers that most want their own app. (a) keeps that working and is consistent with how Jira/CalDAV/Nextcloud secrets already sync. I considered storing the BYO secret in the local-only token store the cleaner long-term fix, but it re-adds a secret-storage path into the host hook we'd just minimized (and forces per-device re-entry), so I left it out for now, happy to switch to that (or to no-secret) if you'd prefer; flagging since it's a sync-format choice.

3. Port-bounds drift: renderer and Electron main now share one OAUTH_LOOPBACK_PORT_{MIN,MAX} constant (in electron/shared-with-frontend/), and a redirect port outside [1024, 65535] (including 0) fails fast at validation with a clear message instead of a confusing main-process rejection.

4. Double-parse / dead loopback check, dropped _getElectronLoopbackPort; since _validateRedirectUri already guarantees http://127.0.0.1:<port> in range, prepareRedirectUri just reads Number(new URL(redirectUri).port).

5. Merge hardening: added a __proto__/constructor/prototype skip (defense-in-depth) with a test (built via JSON.parse so the keys are real own properties), plus a note that the deep merge is intentional for pluginConfig and nested keys clear only by explicit value.

Nit: commented the port-0 EADDRINUSE case.

Thanks again for the thorough reviews.

johannesjo:
## Round-3 review — generic OAuth host hooks

Thanks for the round-2/round-3 work — this is in good shape. I re-ran a multi-perspective pass (correctness / security / architecture / performance / simplicity) over the current head and verified each item against the code. The earlier asks are genuinely addressed: PKCE/`state` are mandatory and validated, the Electron redirect is tightly allowlisted to `http://127.0.0.1:<1024–65535>`, the IPC handler now range-checks the port, registers the `error` handler before `listen()`, and cleans up the server/timer on every path, and the `mergePluginConfig` proto-pollution guard works (nice JSON.parse-based test). No critical data-loss or token-theft path remains.

Two things still worth fixing, then a few smaller calls.

### 🔴 Should fix

**1. Native `redirectUri` validation is a blocklist — this is the open CodeQL HIGH (`js/incomplete-url-scheme-check`).**
`plugin-oauth.service.ts:142` rejects only `['http:','https:','javascript:','file:']`, so `data:`/`vbscript:`/`blob:`/foreign custom schemes pass. Exploitability is low (the `redirectUri` is only sent to the IdP as `redirect_uri`, never navigated locally, and PKCE makes an intercepted code useless), but the fix is trivial and definitive: you already know the exact scheme the host would generate (`com.superproductivity…` / `com.super-productivity.app`, lines 61–63), so make it an allowlist — `parsed.protocol === <appScheme>` — instead of a denylist. That closes the alert permanently and removes a cross-app deep-link-hijack footgun. (Arguably a native `redirectUri` override is meaningless at all, since the native callback must be the app's fixed custom scheme — rejecting native overrides outright would also be defensible.)

**2. My earlier item #2 is only partially fixed — BYO `clientId` is still silently dropped on web/native.**
You added a `PluginLog.warn` for a dropped `clientSecret` (good), but `startOAuthFlow`'s platform branch (`plugin-oauth-bridge.service.ts:53–89`) still overwrites `clientId` with `webClientId`/`mobileClientId`/`iosClientId` with no warning, while the user's BYO `redirectUri` *does* survive. So on web/mobile a user gets the plugin's client id paired with their own redirect URI — a mismatch that fails confusingly (usually caught by `_validateRedirectUri` first, but the inconsistency stands). Either apply the BYO override to the platform client ids too, or warn when a BYO `clientId` is discarded / gate the BYO fields to desktop.

### 🟡 Consider — long-term, hard-to-reverse (this is a public, synced surface)

**3. The reserved keys `clientId`/`clientSecret`/`redirectUri` overload the free-form `pluginConfig` with no namespace and no docs.**
`plugin-oauth-config-overrides.util.ts:17–26` reads three flat top-level keys out of `pluginConfig`. Because `pluginConfig` is synced + persisted and free-form, this (a) collides with any plugin that already uses a `clientId`/`redirectUri` field for something else, and (b) is an invisible host↔plugin contract — a plugin opts in only by naming its fields exactly these strings, documented nowhere. Once plugins ship using them they're de-facto reserved forever. Namespacing now (e.g. `pluginConfig.oauthOverrides = { clientId, clientSecret, redirectUri }`) is cheap; changing it later is a breaking sync-format change. Worth a short note in the plugin-API docs too, since `redirectUri`/BYO is now public surface.

**4. The generic deep-merge is broader than the bug, and introduces a "nested keys never clear" semantic.**
`issue-provider-model-merge.util.ts:6–33`. The actual bug was top-level omission, and `updateModel`'s only callers always emit a *complete* `twoWaySync` (seeded by `_migratePluginConfigForEdit`), so no caller sends a partial nested object — the recursive branch + proto guard + ~2 tests guard a case the code never produces. The minimal fix is the shallow top-level merge I mentioned in round 1 (`next.pluginConfig = { ...current.pluginConfig, ...update.pluginConfig }`), which also makes the proto guard unnecessary (a spread copies `__proto__` as an own prop without touching the prototype). Not a blocker — both were on the table — but the recursion's clear-by-omission behavior means stale nested keys can accumulate and propagate across synced clients, so I'd lean toward the shallow version.

### 🟢 Nits / defense-in-depth
- **Proto-guard asymmetry:** the top-level loop in `mergeIssueProviderModelUpdates` (`:43–63`) has no `__proto__`/`constructor` skip while the nested helper does. Add it for symmetry (or moot it via #4).
- **Web same-origin check is necessary but not sufficient** (`plugin-oauth.service.ts:150`): any same-origin path / `blob:` passes, then hangs the full 5 min — the exact failure mode #1 set out to kill. In practice only `/assets/oauth-callback.html` works; consider pinning to it.
- **Desktop rejects `localhost` / `[::1]`** (`:122`). Consistent and safe (clear error, no hang), but some providers register `http://localhost:<port>` — a doc note would help.
- **Fixed loopback port is predictable**, so a local process/page can `fetch` the known port during the 5-min window to cancel the flow (DoS only — PKCE + `state` block any code injection). Optional: pass `state` to main and only consume the callback on a match.
- **Secret-syncs warning lives in JSDoc** (invisible to users). A note on the plugin's secret field would be good hygiene — same risk class as the existing Jira/CalDAV passwords, so not new, just worth surfacing.

### ✅ Checked and fine
- Electron IPC handler: no double-settle, no server/timer leak; round-1 #3 fully fixed.
- Port-parse path is consistent end-to-end via the shared `oauth-loopback.const.ts` (`:80`→stripped→rejected, `:0`→range-rejected, `>65535`→`URL` throws→caught).
- Performance: the new `updateModel` reassignment triggers one formly rebuild per change, but it's O(form fields) at typing cadence on a settings dialog — immaterial.
- Logging is clean (pluginId/port only, no secrets/codes/tokens); optional `redirectUri` is forward-compatible.

Happy to merge once **#1** and **#2** are handled; **#3/#4** are the calls I'd most like to settle before this public surface ships, but I won't block on them. Nice tests throughout.

oonid:
Thanks @johannesjo  for the thorough round-3 pass,
all four items plus the nits are addressed on plugin-oauth-host-hooks:

🔴 #1 : native redirectUri (CodeQL HIGH). Switched from a denylist to rejecting native overrides outright: a native callback must use the app's fixed custom scheme, so a plugin-supplied override is meaningless there. No scheme can slip through, which closes js/incomplete-url-scheme-check permanently. 
(Validation now lives in a pure, platform-parameterized validateOAuthRedirectUri util with unit tests across electron/native/web.)

🔴 #2 : BYO clientId dropped on web/native. Gated the bring-your-own credentials (clientId/clientSecret/redirectUri) to the desktop loopback flow: on web/native they're ignored with a warn, so a user's BYO redirectUri can no longer survive while their clientId is swapped, the mismatch is now structurally impossible. Also consolidated the three per-platform clientSecret-drop warnings into one generic message.

🟡 #3 : reserved keys / namespace. Moved the overrides under pluginConfig.oauthOverrides = { clientId, clientSecret, redirectUri } so they no longer collide with free-form plugin fields, and documented the convention (desktop-only) in the plugin-API types.

🟡 #4 : deep-merge. Replaced the recursive merge with the shallow top-level { ...current, ...update }. The spread copies any own __proto__ as a plain property, so the explicit proto-guard is gone, and nested objects (twoWaySync) are replaced wholesale as callers always emit a complete object, no more clear-by-omission accumulation across synced clients.

Nits: web redirectUri is now pinned to /assets/oauth-callback.html (not just same-origin); the loopback flow already passes state through to the callback and only resolves on a match; the desktop localhost/[::1] rejection is documented; and the secret-syncs caveat is surfaced in the plugin's client-secret field help text (it's the same risk class as existing Jira/CalDAV passwords).

Test companions: validate-redirect-uri.util.spec (30), plugin-oauth-bridge.service.spec (7), and the generic merge + overrides util specs (5 + 5), all green.

johannesjo:
## Round-3 follow-up — fixes land cleanly, one cross-platform footgun to weigh

Re-reviewed `8060731..1261be0` against the round-2 comments. The responses are all good:

- **#1 (silent 5-min hang):** `validateOAuthRedirectUri` now fails fast — electron loopback / web exact-callback-path / native reject. Nice extraction into a pure, well-tested util.
- **#2 (BYO creds dropped on web/native):** `applyPluginOAuthOverrides` is gated to `isElectron` + warns; the three per-platform secret warnings are consolidated into one fired from `config.clientSecret && !effectiveConfig.clientSecret`.
- **#3 (IPC port):** `electron/plugin-oauth.ts` clamps the requested port to `[1024, 65535]`. 👍
- **#5 (twoWaySync deep-merge):** dropped the special-case, kept the top-level shallow merge — exactly the suggested option. I confirmed `{...current, ...update}` is prototype-pollution-safe even with a JSON-parsed own `__proto__` key, so removing the guard is fine. The remaining invariant (no caller emits a *partial* nested object under `pluginConfig`) holds today — `twoWaySync` is formly-managed (whole-model emit) and CalDAV's `twoWaySync` is a top-level key, not under `pluginConfig`. The comment documents it, which is the right safeguard.
- Nits: named imports restored, stale `getRedirectUri()` comment fixed, validator correctly requires `127.0.0.1` (not `localhost`).

### 🟠 Worth weighing before merge: a plugin-declared `redirectUri` breaks the *web* and *native* flows

`OAuthFlowConfig.redirectUri` is a single field shared across platforms, but `startOAuthFlow`'s `effectiveConfig` only strips per-platform fields for `clientSecret` (and swaps `clientId`) — it **keeps `redirectUri` on every branch** (`plugin-oauth-bridge.service.ts:52-74`). So when a plugin declares the desktop loopback URI this PR exists to enable, *and* also supports web (`webClientId`) or native, validation throws on those platforms. Running the exact `validateOAuthRedirectUri` logic against `http://127.0.0.1:8976/callback`:

```
electron: PASS (no throw)
web     : THROWS -> OAuth redirectUri on web must be <origin>/assets/oauth-callback.html; got http://127.0.0.1:8976/callback
native  : THROWS -> OAuth redirectUri override is not supported on native; got http://127.0.0.1:8976/callback
```

The `isOAuthUnavailableInWeb` guard only checks `webClientId`, so a web-capable plugin passes it, reaches `prepareRedirectUri(effectiveConfig.redirectUri)`, and throws at connect time — even though the web flow would have worked with the default `/assets/oauth-callback.html`. It surfaces as an error snack (so this is *not* a #1-style hang — good), but the mere presence of a desktop `redirectUri` makes the plugin's OAuth unusable on the other two platforms.

Since `redirectUri` is desktop-only by nature (like `clientSecret`), the cleanest fix is to strip it in the web/native branches so they fall through to the platform default — which is lossless, because the web validator only accepts `<origin>/assets/oauth-callback.html` and native only accepts the app scheme (both = the default a stripped config already produces):

```ts
if (IS_ANDROID_NATIVE && config.mobileClientId) {
  return { ...config, clientId: config.mobileClientId, clientSecret: undefined, redirectUri: undefined };
}
// iOS + web branches likewise add: redirectUri: undefined
```

If the intent is instead "declaring `redirectUri` makes a plugin Electron-only," it'd be worth saying so in the `OAuthFlowConfig.redirectUri` JSDoc (the new JSDoc documents `oauthOverrides` as desktop-only but is silent on the plugin-declared `redirectUri`). Either way it's a public plugin-API shape, so better pinned down before it ships. No bundled plugin hits it today, so not urgent — just hard to change later.

### 🟢 Tiny nits
- The consolidated bridge warning lost its per-platform specificity (“on this platform” vs “on Android/iOS/web”) — fine, just a touch less informative in logs.
- `validate-redirect-uri.util.spec.ts` → `"should reject when all platforms are false (invalid state)"` is really just the web case (its own comment says “fall through to web validation”); the name reads as if it covers something else.

---
*Verified by reading all 12 files at `1261be0`, tracing `connectOAuth → startOAuthFlow → prepareRedirectUri → validateOAuthRedirectUri`, and replaying the validator + the spread-merge prototype-safety in node. I did not run the Karma specs (this checkout is on master); the test expectations are internally consistent with the new error strings.*

oonid:
Thanks @johannesjo  for the thorough round-3 follow-up, 
the cross-platform footgun was a real one. I've split the response into two commits so the optional part is trivially separable.
  
cec4b60 : fix(plugins): strip desktop redirectUri on web/native OAuth flows (your exact ask)

- 🟠 redirectUri is now stripped on the Android / iOS / web branches of effectiveConfig, so a web/native-capable plugin that also declares the desktop loopback URI falls through to the platform default instead of throwing at connect time. Lossless, exactly as you noted — the web validator only accepts <origin>/assets/oauth-callback.html and native only accepts the app scheme, which is what a stripped config already produces.

- I went with "strip it" rather than "Electron-only," and also pinned down the public API shape: the OAuthFlowConfig.redirectUri JSDoc now documents it as the desktop (Electron loopback) override, ignored/stripped on web and native.

- 🟢 Renamed the misleading validate-redirect-uri.util.spec.ts test ("should reject when all platforms are false" → it's the web case).
- 🟢 Left the consolidated warning as-is per your note.

Companion test: the bridge spec's web case now asserts the loopback redirectUri is dropped and the flow uses the host callback default (it previously asserted the buggy "honors the override on web" behavior).

37ecb8e : refactor(plugins): extract resolveEffectiveOAuthConfig and harden native fallthrough (optional, safe to drop)

While fixing the above I noticed the per-platform IS_* consts are module-level and can't be mocked in karma, so only the web branch was unit-testable. This commit extracts the client/secret/redirectUri selection into a pure, parameterized resolveEffectiveOAuthConfig util (mirroring the validateOAuthRedirectUri extraction you liked) so every branch gets unit coverage, 9 tests across electron/android/ios/web + the fall-through.
  
It also closes one edge the fix commit leaves: the final fall-through (return config) is reachable on a native platform when the plugin ships no matching client id, where the desktop clientSecret/redirectUri would still leak into the native flow. Since both are desktop-only by nature (as you put it, "redirectUri is desktop-only … like clientSecret"), the util strips both on every non-Electron path. This is an unsupported-platform edge rather than the capable-plugin footgun you described, so if you'd rather keep the diff minimal, dropping this commit (git revert 37ecb8e) leaves cec4b60 as a complete,  standalone answer to the review.

All specs pass (bridge 7/7, validate-redirect-uri 30/30, the new util 9/9, plus the unchanged service 36/36).

johannesjo:
Thank you very much! <3

oonid:
Thank you,
and really, thank you for the reviews. The multi-perspective passes (correctness / security / architecture / simplicity) caught real issues I'd have missed and clearly made this a better, safer API: the native reject-outright for the CodeQL case, gating BYO creds to desktop, stripping redirectUri/clientSecret to keep them desktop-only, and the shallow pluginConfig merge. I learned a lot from how thorough and specific the feedback was.
  
Glad the generic hooks landed in a shape that keeps core provider-agnostic, Basecamp will consume them as a standalone community plugin, so no provider-specific code in core. Thanks again for shepherding this through. 🙏

PR #8549, opened 2026-06-22
fix(markdown): prevent checkbox icon ligature text from polluting cli

### LINKED ISSUE #8545: Inline markdown: copying a checklist from the preview yields "check_box_outline_blank" and drops "- [ ]"
**Context:** split out of #8524, which solves three unrelated problems in one PR. This issue tracks the **copy-from-preview** part so the approach can be settled without blocking the #8407 bug fix.

## Bug

Selecting and copying checklist items from the *rendered* preview copies the Material Icons ligature text (`check_box_outline_blank` / `check_box`) and omits the `- [ ]` markdown syntax.

## Root cause

The checkbox is rendered as a Material Icons **ligature** span:

```ts
// marked-options-factory.ts, renderer.listitem
`<span class="checkbox material-icons">${isChecked ? 'check_box' : 'check_box_outline_blank'}</span>`
```

The font renders those words as a glyph, but the underlying `textContent` is the literal words. And the preview SCSS forces `user-select: text !important` on `.markdown-parsed *` — so the ligature words are *deliberately* selectable and land on the clipboard. The `- [ ]` is absent because it isn't in the rendered DOM at all.

## Current approach (in #8524)

A hand-rolled `domToMarkdown` recursive serializer + an unconditional `@HostListener('copy')` that re-serializes the selected DOM fragment back to markdown. Concerns:

- **Partial `turndown` reimplementation.** Tags without an explicit `case` fall through `default`: structural elements (blockquote, table, `hr`) are **flattened to their text** (wrapping syntax lost), and `` is **dropped entirely** (alt/src gone).
- **No markdown escaping.** Text nodes are emitted verbatim, so a label containing `*` / `_` / `` ` `` / `[` round-trips incorrectly or changes meaning. (Link `href`s are also interpolated unescaped.)
- **Partial selections lose nested-list indentation.** Indent is derived from a `depth` counter that only increments through `UL`/`OL` ancestors present in the cloned fragment; `range.cloneContents()` of a selection starting mid-list flattens nesting to depth 0.
- **Intercepts ALL preview copies, not just checklists.** The handler only bails on edit mode / empty selection / selection outside the preview — there is no checklist guard. So ordinary prose, links, and code copied from any note preview also run through the serializer, widening the round-trip-corruption surface well beyond the original bug.

## Preferred direction (minimal)

Fix the garbage at its source — stop the ligature text from being selectable:

- Override the `user-select: text !important` rule with `user-select: none` on `.checkbox.material-icons`, **or** render the glyph via a `::before` pseudo-element (pseudo-element text is never included in copied selection text). This removes the `check_box_outline_blank` noise with no serializer.
- Optionally inject a selectable `☐` / `☑` character to preserve checkbox state in the copied text.
- The real markdown source already lives in the editor view, which is the natural place to copy markdown from.

If "preview copy → valid markdown" is genuinely a desired feature, prefer a maintained library (`turndown`) over a hand-rolled serializer — but decide whether the feature warrants the dependency first.

## Acceptance

- Copying from the preview no longer includes `check_box_outline_blank`.
- Explicit decision on whether full markdown reconstruction on copy is in scope at all.

symonbaikov:
## Problem
Fixes #8545.
When selecting and copying checklist items from the rendered markdown preview, the Material Icons ligature text (`check_box_outline_blank` / `check_box`) was included in the clipboard because the underlying text content of the icon span is selectable.
## Solution
Added `user-select: none` (and `-webkit-user-select: none`) to `.checkbox` in `src/styles/components/markdown.scss`. This prevents the icon text from being selected or copied while preserving the visual checkbox and clickable behavior.
This is the minimal fix discussed in the issue; it removes the garbage text from the clipboard without attempting full markdown reconstruction.
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)
## Checklist
- [ ] I have included relevant changes to the [documentation/wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki).

### DIFF
--- a/src/styles/components/markdown.scss
+++ b/src/styles/components/markdown.scss
@@ -111,6 +111,8 @@
     padding-right: var(--s);
     position: relative;
     cursor: pointer;
+    user-select: none;
+    -webkit-user-select: none;
 
     &:after {
       content: '';


johannesjo:
Thanks for tackling this, @symonbaikov — and for going the fix-at-source route, which is exactly the direction the issue preferred over the serializer.

After digging in, though, this specific change turns out to be a **no-op**, because the reported bug was already fixed by #8524 (merged just after you opened this), and the app has a global selection baseline that makes the new rule redundant everywhere else. Walking through it:

- `styles.scss` sets a global **`* { user-select: none }`**, with only `p, input, textarea, pre, [contenteditable]` re-enabled. A checkbox is a `<span class="checkbox material-icons">`, so it's **already unselectable in every markdown preview by default.**
- The only places that *re-enable* selection on the checkbox are:
  - `inline-markdown.component.scss` → `.markdown-parsed * { user-select: text !important }` — this is the surface #8545 reported. **#8524 already fixed it** with `.markdown-parsed .checkbox { user-select: none !important }` (higher specificity wins).
  - `issue-table.scss` → `.issue-table td { *, .markdown * { user-select: text } }`.
- This PR adds a **non-`!important`** `.markdown .checkbox { user-select: none }`. In the inline preview it's overridden by the `text !important` rule; in issue tables it's out-specified by `.issue-table td .markdown *`; everywhere else the checkbox was already unselectable. So it doesn't change behavior in any surface.

Since #8524 resolves the reported case, I'll close this and #8545 as already-fixed. Really appreciate the contribution — the instinct was right, it just got there a few hours after the other PR landed. 🙏

> Minor aside for the record: checklists inside **issue descriptions** (rendered in `.issue-table`, which forces `user-select: text`) are a separate, much rarer surface still affected — it would need an `!important`/higher-specificity rule, not a plain one. Happy to spin up a tiny follow-up if anyone hits it in practice.

PR #8550, opened 2026-06-22
Add 'YouTrack Importer for SuperProductivity' plugin to community plugins list

coissay:
Add 'YouTrack Importer for SuperProductivity' plugin to community plugins list

## Problem

No integration for Youtrack 

## Solution

Add integration as plugin by importing csv out of youtrack or via youtrack apikey

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -110,5 +110,13 @@
     "author": "giba0",
     "authorUrl": "https://github.com/giba0",
     "stars": 2
+  },
+  {
+    "name": "YouTrack Importer for SuperProductivity",
+    "shortDescription": "Bring YouTrack issues into SuperProductivity, either as a one-off CSV import or as a live, repeatable sync directly against the YouTrack REST API.",
+    "url": "https://github.com/coissay/superProductivity-YoutrackPlugin",
+    "author": "coissay",
+    "authorUrl": "https://github.com/coissay",
+    "stars": 0
   }
 ]


johannesjo:
Thanks for the contribution! I reviewed the linked plugin code (`coissay/superProductivity-YoutrackPlugin`) before adding it to the curated list, and there's one blocker plus a few smaller issues to sort out first.

### 🔴 Blocker: hardcoded YouTrack host sends non-nperf users' API tokens to a third party

`plugin.js` hardcodes the instance URL and never overrides it — there's no instance-URL field anywhere in the UI or config:

```js
const YOUTRACK_BASE_URL = 'https://youtrack.nperf.org';
...
const url = new URL(`${YOUTRACK_BASE_URL}/api/issues`);
const response = await fetch(url.toString(), {
  headers: { 'Authorization': `Bearer ${token}`, ... },
});
```

So any user who isn't on the `youtrack.nperf.org` instance follows the README ("connect directly to a YouTrack instance with a token"), pastes **their** permanent token, and the plugin transmits it via `Authorization: Bearer` to `youtrack.nperf.org` — a server they have no relationship with. The request fails (wrong instance), but the token has already been sent in the request headers to a third-party host that can log it. That's an inadvertent credential disclosure for essentially every public user, and it makes the plugin non-functional for anyone except nPerf.

This also contradicts the plugin's own docs, which promise a configurable URL:
- README: "the plugin makes direct `fetch()` calls to **whatever YouTrack instance URL you configure**"
- `manifest.json` `networkUsage`: "Direct fetch() calls to **the configured YouTrack instance's** REST API"
- the token-generation help link is hardcoded too: `https://youtrack.nperf.org/settings/tokens`

**Fix:** make the instance URL a user-configurable field (replace the hardcoded constant and the help link). This needs to land before I can add it to the list.

### 🟠 Manifest / docs integrity

- Version drift across three files: `plugin.js` header says `v1.4.0`, `manifest.json` says `1.6.1`, README footer says `1.7.0`.
- The README "Permissions & API usage" section claims `permissions: []` and a `pluginApiUsage` list are in `manifest.json` — neither field is actually present. (No `permissions`/`nodeExecution` is fine security-wise; the issue is the documented manifest doesn't match the shipped one.)

### 🟡 Token persistence

`saveConfig` persists the whole config including the plaintext token via `persistDataSynced(...)`, so the permanent token becomes part of synced data and travels to the user's sync backend (cleartext if their provider isn't encrypted). Somewhat inherent to the feature, but worth being deliberate about.

### 🟢 What looks good

To be clear, the rest of the security posture is reasonable: no `nodeExecution`, no `eval`/`new Function`, a single network destination (no covert exfiltration), and `escapeHtml()` is applied consistently on remote/CSV data before any `innerHTML` render. The sync logic (paginated fetch with a safety cap, idempotent update-not-duplicate keyed on the issue-id title prefix) is sensibly built and well-commented.

Minor non-blocking bugs to pass along: `toDueDay` uses `toISOString().slice(0,10)` (UTC) which can shift the due date by a day near midnight in some timezones — the sprint code already uses a local-date formatter to avoid exactly this; and `task.state` is read in the CSV preview but `parseCSV` never sets that property, so the State chip never renders.

Happy to take another look once the configurable-URL fix is in. Thanks again!

coissay:
@johannesjo I just push a new version.

* URL is not hardcoded anymore (it was a mistake on my end)
* Versions should all be the same now
* I also removed the token from the preset save
* I change some html design too

I will soon release the fix to `toDueDay`

coissay:
the toDueDay should be fix

johannesjo:
Thank you very much! <3

PR #8555, opened 2026-06-23
fix(tasks): resolve dropped-file path via webUtils.getPathForFile

### LINKED ISSUE #8553: 💥can't open file attachement
### Steps to Reproduce
1. Fresh install Win11 version either via MS store or via release download, version 18.12.0W
2. create any task, drag n drop upload a jpg or png file
3. click on open attachedment have no response, no output in console

Tested chaging the file storage to D drive, no improvement
Also tested Ctrl+V paste, no UI response at all, no console output, basically nothing happens

Image

### Additional Console Output

Warning when uploading and saving file attachement
[ol] [ValidateOperationPayload] UPDATE payload has unusual structure {entityType: 'TASK', opType: 'UPD', entityId: undefined, entityIdsCount: undefined, payloadType: 'object', …}
[ol] [OperationLogEffects] Operation payload warnings {warnings: Array(1), actionType: '[TaskAttachment] Add TaskAttachment'}

johannesjo:
Closes #8553

## Problem
On the desktop (Electron) build, dragging a file onto a task and clicking
"open attachment" did nothing — no action, no console output.

## Root cause
Electron 32+ removed the non-standard `File.path` property; the app ships
Electron 41.4.0. `_createFileBookmark` still read `dataTransfer.files[0].path`
(now `undefined`) and fell back to the bare file name. The attachment was
stored as `{ type: 'FILE', path: 'image.jpg' }`, so `shell.openPath('image.jpg')`
could not resolve the relative path and silently no-op'd.

## Fix
Resolve the absolute path via `window.ea.getPathForFile` — the official
`webUtils` replacement already used by `clipboard-image.service` — falling back
to `file.name` in the browser. The attachment title is sourced from `file.name`
so it stays clean rather than becoming the full OS path.

## Tests
New `drop-paste-input.spec.ts` (4 tests): file drop, Electron path resolution,
null fallback, dropped-text-as-link. `checkFile` passes; dependent specs green.

## Notes
- Security: the resolved path is synced, but `isPathSafeToOpen` still blocks
  UNC / `file://host` paths in the main process before `shell.openPath`
  (GHSA-hr87-735w-hfq3 defense intact).
- Not device-verified on Windows/Electron 41 in CI, but `getPathForFile` is the
  documented `File.path` replacement and already used identically elsewhere.

### DIFF
--- a/src/app/core/drop-paste-input/drop-paste-input.spec.ts
+++ b/src/app/core/drop-paste-input/drop-paste-input.spec.ts
@@ -1,0 +1,74 @@
+import { createFromDrop } from './drop-paste-input';
+
+// Minimal DragEvent stub: only the bits createFromDrop reads.
+const dropEventWithFile = (file: File, text = ''): DragEvent =>
+  ({
+    dataTransfer: {
+      getData: (type: string) => (type === 'text' ? text : ''),
+      files: [file] as unknown as FileList,
+    },
+  }) as unknown as DragEvent;
+
+describe('createFromDrop', () => {
+  let originalEa: typeof window.ea | undefined;
+
+  beforeEach(() => {
+    originalEa = window.ea;
+  });
+
+  afterEach(() => {
+    if (originalEa === undefined) {
+      delete (window as { ea?: typeof window.ea }).ea;
+    } else {
+      window.ea = originalEa;
+    }
+  });
+
+  it('should create a FILE attachment from a dropped file', () => {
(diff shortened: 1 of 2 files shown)


rxunique:
legend speed on the fix! I got claude to compile from master and can confirm the file open works on 18.12.1W

however the "ctrl + v to past image" works but feels sporadically, I think it's because "link & file" section have to be in focus first. 

Some quality of life tweak would be nice down the track, I'll open up a FR with detail

PR #8567, opened 2026-06-24
Add Invoice Ninja Export to community plugins list

clinkin-lawyer:
## Summary
Adds a new entry to `src/assets/community-plugins.json` for **Invoice Ninja Export** — a plugin that exports completed SP tasks to Invoice Ninja as billable Client Tasks (grouped by SP project, matched by exact name to an Invoice Ninja Client).

- Repo: https://github.com/clinkin-lawyer/sp-invoice-ninja-export
- License: MIT
- Tested manually against a live Invoice Ninja (invoicing.co) instance: project/client name matching, task export with time logs, and dedup via export history tracking all confirmed working.

## Test plan
- [x] `community-plugins.json` validated as well-formed JSON
- [x] Plugin installs and runs in SuperProductivity via both "Load Plugin from Folder" and "Upload Plugin" (zip)
- [x] README documents setup (API token generation/configuration) and the project-name-to-client-name matching requirement

### DIFF
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -110,5 +110,13 @@
     "author": "giba0",
     "authorUrl": "https://github.com/giba0",
     "stars": 2
+  },
+  {
+    "name": "Invoice Ninja Export",
+    "shortDescription": "Exports completed tasks to Invoice Ninja as billable Client Tasks, grouped by SP project (matched by name to an Invoice Ninja Client).",
+    "url": "https://github.com/clinkin-lawyer/sp-invoice-ninja-export",
+    "author": "clinkin-lawyer",
+    "authorUrl": "https://github.com/clinkin-lawyer",
+    "stars": 0
   }
 ]


johannesjo:
Thanks for the contribution! 🙏 Since this adds a discovery listing for a plugin that handles a real credential (an Invoice Ninja API token), I went through the linked plugin code (`clinkin-lawyer/sp-invoice-ninja-export`) as well as the listing itself.

**On the PR itself:** this is fine to merge as-is — the community list is discovery-only (name/description/repo link rendered behind the "not reviewed, install at your own risk" warning), so nothing here executes code or auto-installs anything.

That said, while I was in there I found a few things in the plugin worth tightening up before it's listed. None of these are "this is malicious," and the intent and code are clean — they're hardening issues around the fact that this plugin holds a full-access invoicing credential.

**1. The HTTP proxy doesn't validate the request host (please fix this one).**
`plugin.js` runs in the main renderer and proxies requests on a fixed-name `BroadcastChannel('invoice-ninja-proxy')`, attaching the user's `X-API-Token` to `base + path`. Because the URL is built by string concatenation with no host check, a `path` like `@evil.com/...` or `.evil.com/...` resolves to a foreign origin and leaks the token. The plugin's own iframe only ever passes hardcoded paths, but the iframe is same-origin (`allow-same-origin` + `srcdoc`), so that channel is shared with every other installed plugin — any co-installed plugin could drive the proxy to exfiltrate the token or make authenticated calls against the user's Invoice Ninja account. Suggested fix: validate `new URL(base + path).origin === <configured origin>` before sending, and don't accept a base/`apiUrl` override over the channel.

**2. Remove `"nodeExecutionConsent": true` from `manifest.json`.**
That field isn't part of the manifest model — the host gates Node execution on `permissions: ["nodeExecution"]`, which you (correctly) don't request. So it grants nothing, but it's confusing to anyone reading the manifest and looks like the plugin is asking for elevated access it isn't.

**3. Trim unused permissions (least privilege).**
The manifest requests `getArchivedTasks`, `getCurrentContextTasks`, `notify`, and `openDialog`, but the code only uses `getTasks`, `getAllProjects`, `showSnack`, `persistDataSynced`, `loadSyncedData`, and `getConfig`. Dropping the unused ones keeps the permission prompt honest.

**Heads-up (not a blocker, and not specific to your plugin):** the API token entered in the plugin settings is stored in synced plugin data, so it propagates to the user's other devices and into backups/exports (encrypted only if SuperSync encryption is enabled). Your README already says to treat the token like a password, which is good — just flagging that the storage is synced so you're aware.

Minor: in `index.html` the `option value="${p.id}"` interpolation is unescaped (low risk since project IDs are app-generated, but inconsistent with the `escHtml(...)` you use everywhere else).

Happy to re-review once #1–#3 are addressed. Thanks again!

clinkin-lawyer:
Thanks for the thorough review, Johannes — appreciate you going through the plugin code, not just the listing.

Pushed a fix (v1.1.1, commit eac7943):

1. Proxy origin check (fixed): plugin.js now resolves new URL(base + path) and rejects the request if its origin doesn't match the configured Invoice Ninja base URL, before it ever reaches fetch(). Tested it against the @evil.com/...-style userinfo trick you described and confirmed it's caught. base/apiUrl were already sourced only from getConfig()/synced data, never from the channel message, so that half was already safe — this closes the other half.
2. nodeExecutionConsent: removed from manifest.json.
3. Permissions: trimmed to just what the code calls — getTasks, getAllProjects, showSnack, persistDataSynced, loadSyncedData, getConfig.
4. option value: now runs through escHtml() like the rest of index.html.

On the synced-storage heads-up: understood, and agreed it's worth users knowing, I'll leave the README's "treat it like a password" language as the flag for now rather than changing storage behavior, since that'd be a bigger design change. Open to revisiting if you think it needs more than a doc warning.

Let me know if there's anything else you'd like addressed before merging.

johannesjo:
Thanks for the submission! The entry itself looks good — valid JSON, prettier-clean, and it matches the `CommunityPlugin` shape the UI expects.

One thing blocks the merge though: **this branch now conflicts with `master`.**

```
$ git merge-tree --write-tree origin/master f76dec7583
CONFLICT (content): Merge conflict in src/assets/community-plugins.json
```

The branch is based on a 14-entry version of the list; `master` is at 17 now. Three entries landed at the same array tail since this PR was opened — *YouTrack Importer*, *Home Assistant Bridge*, and most recently *Goals Vision Board* (#9123) — so your addition and theirs both edit the same closing `}` / `]` region.

The fix is just a rebase, re-applying your single object at the new end of the array:

```bash
git fetch https://github.com/super-productivity/super-productivity.git master
git rebase FETCH_HEAD
# resolve src/assets/community-plugins.json: keep all 17 existing entries,
# then append your "Invoice Ninja Export" object as the last element
npx prettier --write src/assets/community-plugins.json
git add src/assets/community-plugins.json
git rebase --continue
git push --force-with-lease
```

Worth noting this file conflicts by construction — every pending addition appends at the same spot — so it's best to rebase right before merge rather than early, otherwise the next one to land bumps you again.

Also FYI: the array position doesn't matter for correctness. The plugin list is sorted by name at display time, so you don't need to worry about where in the file the entry ends up.

johannesjo:
Thanks for following up on the hardening items — I checked the live source and all four are in: the proxy validates `new URL(base + path).origin` against the configured origin, `nodeExecutionConsent` is gone, permissions are trimmed to what the code actually calls, and `option value` goes through `escHtml()`. The only URL in the plugin is the `invoicing.co` default, the token is sent only as `X-API-Token` and never logged or exposed to the iframe, and only title/timeSpent/doneOn leave the app. The listing entry is valid and `stars: 0` is still accurate.

The one blocker is the rebase from July — the branch is ~545 commits behind now and the list has grown from 14 to 19 entries, so it conflicts:

```bash
git fetch https://github.com/super-productivity/super-productivity.git master
git rebase FETCH_HEAD
# resolve src/assets/community-plugins.json: keep all 19 existing entries,
# then append your "Invoice Ninja Export" object as the last element
npx prettier --write src/assets/community-plugins.json
git add src/assets/community-plugins.json && git rebase --continue
git push --force-with-lease
```

Two optional things on the plugin side while you're in there:

- `minSupVersion` is `"13.0.0"`, but the plugin API only shipped in v14.0.0 — a 13.x user would install it and find it doesn't work.
- `registerHeaderButton` is registered with an empty `onClick`, so the button renders but does nothing.

And on the proxy: the origin check closes the token-exfiltration path, but the `BroadcastChannel` name is fixed and the iframe is same-origin, so a co-installed plugin can still drive authenticated requests against the user's Invoice Ninja instance and read the responses. A per-session nonce, or an allowlist of the four paths you actually call, would close that. Not required for the listing.

johannesjo:

Thanks for the PR, and for turning around the hardening items from the earlier passes. I read the plugin source fresh at `1286131` – all 7 files, ~30 KB (`manifest.json`, `plugin.js`, `index.html`, `config-schema.json`, README, LICENSE, icon), nothing skipped, no bundled zip. The listing entry itself is clean: valid shape, the URL is the plugin the description claims, you are the author, `stars: 0` is accurate, and the repo is active (last push 5 Aug) and MIT.

For the record, what the read turned up: no `eval`, `new Function`, dynamic `import()` or remote `<script>` anywhere; no committed credentials; the only host reached is the Invoice Ninja base URL from your config (default `invoicing.co`). The token goes out only as `X-API-Token`, never into the iframe and never into a log. What leaves the app is the task title, the done date, and start/end seconds derived from `timeSpent` – SP project titles stay local, since client matching happens against the fetched list. The origin check on the proxy holds up when I trace it: an `@evil.com/...` path resolves to a foreign origin and an absolute-URL path resolves to a nonsense host, so both are rejected before `fetch`, and `base`/`apiUrl` are only ever read from `getConfig()`/synced data, never from a channel message.

One thing worth stating plainly rather than leaving implicit: because `plugin.js` runs in the host renderer, its raw `fetch` sits entirely outside the app's `allowedHosts` gate, and the manifest declares no hosts. That is inherent to the proxy design, and unavoidable while self-hosted URLs are supported, but it does mean the app's permission display tells a user nothing about where this plugin talks.

### Blocking

- **`src/assets/community-plugins.json`** – the branch still conflicts with `master`. Your base had 14 entries, `master` is at 20 now, and every pending listing PR appends at the same array tail. Rebase and re-append your object as the last element:

  ```bash
  git fetch https://github.com/super-productivity/super-productivity.git master
  git rebase FETCH_HEAD
  # keep all 20 existing entries, append "Invoice Ninja Export" last
  npx prettier --write src/assets/community-plugins.json
  git add src/assets/community-plugins.json && git rebase --continue
  git push --force-with-lease
  ```

### Should fix before merge

- **plugin repo, `index.html:279`** – `persistDataSynced(json)` is called without a `key`, which targets the bare-`pluginId` entry. That is the *same* storage entity SP's own gear-icon config dialog writes, and that dialog saves by overwriting the whole blob with `{config, savedAt}`. So saving the token or the URL in the gear dialog wipes the synced `exportedIds` map. On the machine that did it nothing looks wrong, because `localStorage` still holds the history – but a second device installing the plugin afterwards reads the synced blob, finds no export history, lists every completed task as pending (the date filter starts unfiltered, see the README nit below), and Export creates duplicate billable Client Tasks with duplicate time logs. The collision also runs the other way: a device that has the history but never received the config – one that only ever used "Mark all as exported", which needs no token – writes back a blob with no `config` key and LWW-drops the user's token.

  The fix is one argument. Keyed entries LWW-resolve per entity and never collide with the config blob, and the iframe API forwards `key`:

  ```js
  await PluginAPI.persistDataSynced(JSON.stringify(storedData.exportedIds), 'exportHistory');
  const raw = await PluginAPI.loadSyncedData('exportHistory');
  ```

  With that split you no longer need the `{...base}` merge at all.

- **plugin repo, `plugin.js:17-20`** – the proxy accepts any `{type:'IN_REQUEST'}` on a fixed channel name with no sender check at all. I raised this in August as optional; having now read the whole thing ahead of listing it, I'd rather see it closed first, because this plugin holds a full-access invoicing credential and the fix is small. The origin guard stops the token going to a foreign host, but `method`, the path within the origin, and `body` are all caller-controlled, and every plugin iframe is same-origin – so a co-installed plugin can `POST /api/v1/invoices` (or read clients, or payments) as the user and see the response. Either put a per-session random token in the `plugin.js` → iframe handshake and reject messages without it, or allowlist the three paths you actually call together with their methods.

### Nits

- **`index.html:543`** – leftover debug text in the success message: `... exported successfully. (debug) doneStatusId=${doneStatusId}`. Every successful export shows an internal Invoice Ninja status id to the user.
- **README "Export" step vs `index.html:196-205`** – the README says the date range "defaults to the 1st of the current month → today", but nothing ever assigns `fromDate`/`toDate` a value; both render empty, so the list starts unfiltered. Set them on bootstrap, or correct the README. (This is also what makes the f2 scenario above list *everything* on a fresh device.)
- **`manifest.json:6`** – `minSupVersion` is still `"13.0.0"`. The plugin system first shipped in v14.0.0 (`plugin-api.ts` lands in `d4d81bf`; the earliest tag containing it is `v14.0.0`). Nothing enforces the field at runtime, so this is purely documentation accuracy, but a 13.x reader would conclude it works.
- **`plugin.js:12`** – `registerHeaderButton` still has an empty `onClick`, so the toolbar button renders and does nothing. Wire it to `showIndexHtmlAsView()` or drop the registration.
- **`plugin.js:42`** – the synced-storage fallback reads `stored.apiToken` at the top level, but SP stores gear-dialog config as `{config: {...}, savedAt}`, so that branch can never find a token the dialog wrote. Read `stored.config?.apiToken` too, or delete the fallback.

### Questions

- On token storage: since June the host has offered `PluginAPI.setSecret` / `getSecret` – local-only per-plugin credential storage, explicitly kept out of sync, exports and backups (shipped in v18.13.0). It isn't exposed to the iframe, but `plugin.js` runs in the renderer and is already the only code that touches the token, so it looks like a good fit here. The cost is that the token would have to be entered somewhere other than the gear dialog, since `jsonSchemaCfg` config is synced by construction. Is that a change you'd be up for, or would you rather keep the README warning for now?

---

PR #8570, opened 2026-06-24
fix(localization): keep dateTimeLocale override from being clobbered by UI language

### LINKED ISSUE #8565: 🚨 Locale clock is not correctly dispayed in 24h
### Version Used

SP18.12.0FP __ Electron – en-US – Linux x86_64 – en-US – UA:Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) superProductivity/18.12.0 Chrome/146.0.7680.216 Electron/41.4.0 Safari/537.36

### Current Behavior

Clock is displayed in 12h AM/PM

### Expected Behavior

Clock should be displayed in 24h without AM/PM

### Steps to Reproduce
1. Go to `Settings > Localization`
2. Set **Language** to `English`, **First day of the week** `Monday`, **Datetime format locale** to `Japanese: 24-hour, YYYY/MM/DD`
3. Close/Open SuperProductivity
4. add a task, and open the scheduling menu
5. notice that you can type i.e. 14, and it gets calculated into 02:00 PM, but is not 14:00.

johannesjo:
Closes #8565

## Problem
With UI Language = English but **Datetime format locale** = a 24h locale (e.g. Japanese), the task scheduling dialog's time field rendered 12h AM/PM (typing `14` → `02:00 PM`).

## Root cause
Two services wrote the shared Angular Material `DateAdapter` locale — `DateTimeFormatService` (applying the explicit `dateTimeLocale`) and `LanguageService` (applying the UI *language*). On startup these raced; the language write (`en`) won, so the timepicker used a 12h locale. Only *time* broke because `CustomDateAdapter` formats dates via `currentLocale()` (always honors the override) but time via the adapter's contested `setLocale` value.

## Fix
Make `DateTimeFormatService` the single owner of the adapter locale:
- The UI language is registered via `setUiLanguage()` as a signal-backed, lowest-priority fallback the effect reads (chain: `dateTimeLocale → browserCultureLang → uiLang → currentLang → default`).
- `_setDateAdapterLocale` is now private, so the single-owner invariant is compiler-enforced rather than comment-enforced.
- Adds 3 regression tests; the clobber test was verified to fail under the old behavior.

## Note
For users with **no** explicit `dateTimeLocale`, the adapter now deterministically follows browser region over UI language (per #8000's intent), removing the prior startup race.

### DIFF
--- a/src/app/core/date-time-format/date-time-format.service.spec.ts
+++ b/src/app/core/date-time-format/date-time-format.service.spec.ts
@@ -296,4 +296,82 @@ describe('DateTimeFormatService', () => {
       expect(service.currentLocale()).toBe('fr');
     });
   });
+
+  describe('explicit dateTimeLocale override (#8565)', () => {
+    const TIME_FORMAT: Intl.DateTimeFormatOptions = {
+      hour: 'numeric',
+      minute: 'numeric',
+    };
+    // 14:00 — renders "14:00" in 24h locales, "2:00 PM" in 12h (en-US) locales.
+    // The adapter time path (matTimepicker) uses the adapter's setLocale value,
+    // which is what the bug clobbered with the UI language.
+    const testDate = new Date(2000, 11, 31, 14, 0, 0);
+
+    const setup = (
+      dateTimeLocale: string | null,
+      currentLang: string,
+    ): DateTimeFormatService => {
+      TestBed.resetTestingModule();
+      TestBed.configureTestingModule({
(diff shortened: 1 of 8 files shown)


johannesjo:
Split into two focused PRs: the #8565 localization fix is now #8574, and the Later Today start-time feature is now #8575. Closing this combined PR in favor of those.

PR #8580, opened 2026-06-24
fix(sync): surface OneDrive OAuth token errors and document Entra setup

### LINKED ISSUE #8572: 🚨 HTTP 400 Bad Request when trying to sync on OneDrive
### Version Used

v18.12.0

### Current Behavior

Hello,

First of all, thank you for the work on this great software! It is really helping!

So I tried to enable syncing with OneDrive. 
I couldn't find the redirect URI in the Wiki but I found `superproductivity://oauth-callback/onedrive` in the link provided by the SW.

I used the link provided by superProductivity, but Azure did not gave me an auth code like the popup is asking.
Instead the authentication seems to appear in the background.

In the end, superProductivity show a notification at the bottom stating the authentication is ok.
But then an other notification says `HTTP 400 Bad Request `

In the log I have only:
```
[info]                    Processing protocol URL: superproductivity://oauth-callback/onedrive
[info]                    Protocol action: oauth-callback
[info]                    Protocol path parts: [ 'onedrive' ]
[info]                    Received OAuth callback URL via app protocol
```

I hope this will help you.
Thanks!

### Expected Behavior

oauth successfull with OneDrive.
More debug material in the log.

### Steps to Reproduce

Configure the tenant and client IDs from Azure.
Configure the redirect URI with `superproductivity://oauth-callback/onedrive`
Perform authentication with the link provided by superProductivity

### Console Output

```shell

```

### Documentation

- [x] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

johannesjo:
## What & why

Fixes #8572 — OneDrive sync fails with a bare `HTTP 400 Bad Request` after sign-in *appears* to succeed.

The OAuth **authorize** step succeeds (the redirect/callback arrives), then the **token exchange** 400s. That pattern almost always means the user's custom Microsoft Entra app registration isn't set up as a public client (typically "Allow public client flows" disabled → `AADSTS7000218`). Releases ship without a built-in OneDrive client ID, so every OneDrive user must register their own app — making this a common failure with no diagnostics and no docs.

The actual Azure error code was swallowed, and the message wrongly told users to "copy the code exactly" (in Electron the code is auto-captured).

## Changes

- **Surface the real error** (`onedrive.ts`): parse the OAuth error body once, log the safe short `error` code on every token failure (`[OneDrive] OAuth token request failed { status, grantType, error }`), and route the verbose `AADSTSxxxxx` `error_description` to `HttpNotOkAPIError.detail` — UI only, per the existing privacy split.
- **Fix the message** (`sync-wrapper.service.ts` + `en.json` + `t.const.ts`): a OneDrive token 400 now shows `ONEDRIVE_AUTH_FAILED`, naming the public-client root cause and interpolating the AADSTS detail. Other providers keep `INVALID_AUTH_CODE`.
- **Docs**: the in-app info text now shows the desktop redirect URI and links to the wiki; added a full OneDrive section to `2.09-Configure-Sync-Backend.md` (it was entirely missing — redirect URIs, public-client requirement, permissions, this exact 400 troubleshooting).
- **Privacy comment** (`log.ts`): document that `serializeErrorArg` must only emit name/message/stack so `.detail`/`.body` never reach exportable logs.
- **Test**: auth-code 400 surfaces the detail to the UI, logs the short code, and keeps the description out of the structured log.

## Privacy / sync-correctness notes

Error-surfacing + docs only — no reducers, effects, op-log, or sync-state paths touched. The logged meta contains only a status, the grant-type enum, and the standard OAuth `error` code (no user content); the human-readable AADSTS text goes to the UI exclusively.

## Verification

`checkFile` clean · `int:test` pass · 19/19 OneDrive specs · 125/125 sync-wrapper specs · package `tsc` 0 errors. Reviewed via 6-agent parallel review (no blockers).

The underlying Azure config fix is device-confirmable only by the reporter — asked them to verify "Allow public client flows" resolves it.


Nassim1244:
Hi again,

I can't select OneDrive as sync method:
image

;)

johannesjo:
One drive is not available on web.

PR #8588, opened 2026-06-25
refactor(op-log): remove dead cleanup helpers

### LINKED ISSUE #8359: Op-log: delete remaining spec-only/dead code; consolidate the triplicated full-state payload helpers
**Area:** client (op-log) · **Type:** dead-code removal · **Value:** medium

*Trimmed to avoid overlap with the parallel sweep tickets: `isIdenticalConflict`/`_deepEqual`/`_extractEntityFromPayload` → #8324; `migrateIfNeeded`/`getMigrations` + `countFromIndex` → #8327; `validateReminders` → #8326.*

Each remaining item verified by repo-wide reference search:

1. `ConflictResolutionService._extractUpdateChanges` (`conflict-resolution.service.ts:791-796`) — thin delegation to `@sp/sync-core`, spec-only (not in #8324's list).
2. `VectorClockService.getFullVectorClock` (`vector-clock.service.ts:93-109`) — spec-only.
3. `WrappedProviderService.clearCache` — `@deprecated` (`wrapped-provider.service.ts:124`, auto-invalidation replaced it); only caller is its own spec.
4. `bulkHydrationMetaReducer` — `@deprecated` alias (`apply/bulk-hydration.meta-reducer.ts:120-122`); production uses `bulkOperationsMetaReducer`; only specs use the alias — rename the specs, delete the alias.
5. `getOperationCaptureService` (`operation-capture.meta-reducer.ts:113`) — spec-only.
6. `_runLegacyMigrationIfNeeded`/`_runLegacyCleanupIfNeeded` (`operation-log-hydrator.service.ts:550-552, :615-617`) — live no-op placeholders invoked from the startup `Promise.all` (`:78`, `:374`); delete together with their call sites (YAGNI until a migration actually needs the hook).

**Consolidate, don't delete:** three helpers re-implement the `'appDataComplete' in payload` wrapped/legacy discrimination — `extractFullStateFromPayload` + `isWrappedFullStatePayload` (`operation.types.ts:142-162`, **unwraps**), `_extractFullStateFromOp` (`operation-log-hydrator.service.ts:426`, unwraps + opType gating), and `extractFullStatePayload` (`operation-converter.util.ts:37`, returns the **wrapped** form). All three are used in production (upload `:500`, remote-ops-processing `:501,512`, converter `:168`, hydrator `:215,308`). Consolidate onto one canonical helper in `operation.types.ts` with explicit wrapped/unwrapped modes — today a fix to the duck-check must be applied in three places.

symonbaikov:
## Problem

Closes https://github.com/super-productivity/super-productivity/issues/8359

The op-log client code still contained several helpers and compatibility aliases that were no longer used by production code. Some of them were only referenced by their own specs, while the full-state payload wrapping/unwrapping logic was duplicated across multiple op-log paths.

This made the sync/op-log code harder to maintain and increased the chance that future full-state payload fixes would need to be applied in several places.

## Solution

- Removed spec-only/dead op-log helpers and their tests:
  - `ConflictResolutionService._extractUpdateChanges`
  - `VectorClockService.getFullVectorClock`
  - `WrappedProviderService.clearCache`
  - `getOperationCaptureService`
  - deprecated `bulkHydrationMetaReducer` alias
  - no-op legacy migration/cleanup hooks in `OperationLogHydratorService`
- Added a single canonical `normalizeFullStatePayload(payload, mode)` helper in `operation.types.ts`.
- Updated full-state payload handling to use the shared helper in:
  - operation conversion
  - hydration direct-load paths
  - remote ops processing
  - full-state upload
- Updated reducer specs and integration specs to use `bulkOperationsMetaReducer` directly.
- Kept backwards compatibility for wrapped `{ appDataComplete: ... }` payloads while making the wrapping/unwrapping behavior explicit in one place.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

Validation performed:

- `npm run checkFile` on all changed `.ts` files
- `npm run test:file src/app/op-log/core/operation.types.spec.ts`
- `npm run test:file src/app/op-log/apply/operation-converter.util.spec.ts`
- `npm run test:file src/app/op-log/persistence/operation-log-hydrator.service.spec.ts`
- `npm run test:file src/app/op-log/apply/bulk-hydration.meta-reducer.spec.ts`
- `npm run test:file src/app/op-log/testing/integration/bulk-hydration.integration.spec.ts`
- `npm run test:file src/app/op-log/sync/conflict-resolution.service.spec.ts`
- `npm run test:file src/app/op-log/sync/vector-clock.service.spec.ts`
- `npm run test:file src/app/op-log/sync-providers/wrapped-provider.service.spec.ts`
- `npm run test:file src/app/op-log/capture/operation-capture.meta-reducer.spec.ts`
- `git diff --check`

Note: full pre-push `npm test` was attempted, but the global Karma run disconnected Chrome Headless before executing tests (`0 of 0 DISCONNECTED`). The targeted affected suites above pass.


johannesjo (commented):
**Reviewed via multi-agent pass (6 independent reviewers across correctness, security, architecture, alternatives, performance, simplicity). Verified against the PR branch vs `master` and ran the affected specs.**

Nice cleanup — the dead-code removals are all genuinely dead (zero production references for `getFullVectorClock`, `getOperationCaptureService`, `WrappedProviderService.clearCache`, `_extractUpdateChanges`, `_extractFullStateFromOp`, and the two `_runLegacy*` no-op placeholders), and consolidating the three divergent wrap/unwrap implementations into one helper is a real DRY win. No correctness or data-safety defects found — the refactor is behavior-preserving:

- `wasWrapped` is exactly equivalent to the old `isWrappedFullStatePayload(op.payload)` it replaces (wrapped → same ref; unwrapped/null → fresh object).
- The hydrator's inline `isFullStateOpType(...) ? normalizeFullStatePayload(...,'unwrapped') : undefined` preserves `_extractFullStateFromOp`'s null-payload semantics, and `FULL_STATE_OP_TYPES` matches the old explicit `{SyncImport, BackupImport, Repair}` list.
- `WrappedFullStatePayload extends Record<string, unknown>` is a compile-time-only widening — no runtime/wire-format impact.

Two non-blocking cleanups I'd suggest before merge:

**1. `remote-ops-processing.service.ts:499-500` — prefer the existing predicate over the reference-equality trick**

```ts
const wasWrapped = op.payload === normalizeFullStatePayload(op.payload, 'wrapped');
```

This is correct *today*, but its correctness silently depends on `normalizeFullStatePayload(_, 'wrapped')` returning the **same object reference** for an already-wrapped payload — an undocumented implementation detail rather than part of the function's contract. If anyone later makes that helper defensively clone (`{ ...payload }`), `wasWrapped` flips to always-`false` and wrapped full-state payloads get rewritten to bare state on a sync-receive path — a cross-client corruption footgun for zero benefit. `isWrappedFullStatePayload` is still exported and unit-tested, so:

```ts
const wasWrapped = isWrappedFullStatePayload(op.payload);
const fullState = normalizeFullStatePayload(op.payload, 'unwrapped');
```

reads as intent ("was it wrapped?") and drops the redundant second normalize.

**2. Finish the alias cleanup**

The PR removes the `bulkHydrationMetaReducer` alias but leaves its sibling `bulkApplyHydrationOperations` (`bulk-hydration.action.ts:34`) — kept alive only because `operation-log-hydrator.service.spec.ts` (import + 4 call sites) wasn't migrated, while the other two specs were. Migrating those 4 sites to `bulkApplyOperations` and deleting the alias would finish the migration consistently.

**Optional (judgment call):** `normalizeFullStatePayload<TMode>`'s conditional generic return type carries 4 internal `as` casts plus the `extends Record<string, unknown>` widening to serve 5 unwrap callers and 1 wrap caller. Two plain functions — `unwrapFullStatePayload(payload)` / `wrapFullStatePayload(payload)` — would be cast-free, generic-free, and would make finding #1 moot. The current form is correct and tested, so this is just polish.

Nothing here blocks merge — #1 is the one I'd most want addressed since it touches sync-receive payload shaping.

johannesjo:

Thanks for the PR! The cleanup itself is solid – every symbol this PR removes is still dead or spec-only on current master, and the consolidation into one helper is behavior-preserving – but the branch has sat since June and master has moved under it, so it needs a rebase plus one non-obvious migration before it can merge.

### Blocking
- **Branch conflicts with master** – GitHub reports `CONFLICTING`, and a test merge against current master produces 11 conflict hunks across 5 files (`operation-log-hydrator.service.ts` alone has 7 – that file took 17 commits since 2026-06-25, `remote-ops-processing.service.ts` 15, `operation-converter.util.ts` 7). Please rebase onto current master. The removals themselves are all still valid there: `getFullVectorClock`, `getOperationCaptureService`, `WrappedProviderService.clearCache`, `_extractUpdateChanges`, and the two `_runLegacy*` placeholders remain spec-only/dead.

### Should fix before merge
- **`src/app/op-log/sync/verify-decrypted-op-integrity.ts:10`** – a trap the rebase will not surface on its own: this file was added on master on 2026-07-10 (after your branch point), is untouched by your diff, and imports `extractFullStateFromPayload` (called at line 474) – the export this PR renames to `normalizeFullStatePayload`. Git merges it without any conflict, so resolving only the textual conflicts leaves a build that no longer compiles. Migrate that call site to `normalizeFullStatePayload(decryptedPayload, 'unwrapped')` during the rebase; keeping both helpers instead would defeat the consolidation goal of #8359.

### Nits
- **`src/app/op-log/sync/remote-ops-processing.service.ts:500`** – carried over from the June review (item #1, still open): `const wasWrapped = op.payload === normalizeFullStatePayload(op.payload, 'wrapped')` is exactly equivalent to `isWrappedFullStatePayload(op.payload)` today, but only because the helper returns the same reference for already-wrapped payloads – an undocumented detail a future defensive clone would silently break, flipping wrapped sync-receive payloads to bare state. The rebase rewrites this region anyway (it now lives near line 704 on master), so switching to the still-exported `isWrappedFullStatePayload` costs nothing.
- **`src/app/op-log/apply/bulk-hydration.action.ts:34`** – June review item #2, still open: the `bulkApplyHydrationOperations` alias survives only for `operation-log-hydrator.service.spec.ts` (import at :28, call sites at :403, :938, :1110, :1145). Heads-up: current master's `bulk-hydration.meta-reducer.spec.ts` uses the alias ~30 more times, so after the rebase this cleanup grows rather than shrinks – worth finishing in the same pass.
- **`src/app/op-log/persistence/operation-log-hydrator.service.ts:218`** – one edge where "behavior-preserving" is not exact: the old `_extractFullStateFromOp` used a bare `'appDataComplete' in payload` check, so a degenerate `{ appDataComplete: null }` payload yielded a falsy `appData` and fell back to full replay; the new `isWrappedFullStatePayload`-based path classifies it as unwrapped and direct-loads the wrapper object itself (and `_validateStateForHydration` is deliberately non-fatal, so nothing downstream stops it). The same predicate swap double-wraps such shapes in `operation-converter.util.ts`. No current writer produces that shape, so this is informational – but a "has the key yet not wrapped → fall back to replay" guard would make the claim exact.

---

PR #8589, opened 2026-06-25
fix(sync): preserve SuperSync error codes

### LINKED ISSUE #8341: SuperSync: structured server error codes degraded to message-substring matching on the client
**Area:** shared contract + client error handling · **Severity:** medium (fragile by design; not currently broken)

## Problem
The server emits structured `errorCode`s, but on non-2xx responses the client provider folds the error slug into a thrown `Error` message (truncated to 80 chars), and client logic then branches on **message substrings**. The snapshot upload response schema also lacks `errorCode` even though the server sends one, and the dedup `deduplicated: true` field is returned outside the typed contract (tolerated only via `.passthrough()`).

## Evidence
- `src/app/op-log/sync/operation-log-upload.service.ts:542-543` — `message.includes('SYNC_IMPORT_EXISTS')` to reconstruct the code.
- Provider folding/truncation: `packages/sync-providers/src/super-sync/super-sync.ts:49` (`SERVER_ERROR_REASON_MAX_CHARS = 80`), `:548-565` (`_extractServerErrorReason`), `:806-811` (web), `:935-941` (native). (Truncation is currently harmless — the slug starts the field — but the design is fragile.)
- Schema gap: `packages/shared-schema/src/supersync-http-contract.ts:155-161` (`SuperSyncSnapshotUploadResponseSchema` has no `errorCode`) vs server `sync.routes.quota.ts:182-188` (sends `SYNC_IMPORT_EXISTS`) and `:103` (`STORAGE_QUOTA_EXCEEDED`). Note 409/413 bodies never reach the zod schema anyway — non-ok responses throw before JSON parsing, which is exactly why only the message text survives.
- Untyped field: `sync.routes.ops-handler.ts:160-162` (`deduplicated: true` cast outside `SuperSyncUploadOpsResponseSchema`, `supersync-http-contract.ts:126-133`; documented in `docs/wiki/3.01-API.md:73` but unvalidated, no client reader).
- More substring branching: `src/app/op-log/sync/sync-error-utils.ts:14-16, 28-31` (`STORAGE_QUOTA_EXCEEDED`); `src/app/imex/sync/sync-wrapper.service.ts:913-914` ("Payload too complex"/"Payload too large" matching `validation.service.ts:165,174`), `:1368-1370` (timeout/504); `super-sync-restore.service.ts:78-81, 99`.

## Proposed fix
Promote `errorCode` into the typed contract end-to-end: add `errorCode` (ideally an enum mirroring server `SYNC_ERROR_CODES`) and `deduplicated` to the shared zod schemas; have the provider parse non-2xx JSON bodies into a typed error (e.g. `SuperSyncHttpStatusError` carrying `status` + `errorCode`); replace all `includes()` checks (snapshot upload, quota, payload-too-large) with typed-code branching.

symonbaikov:
## Summary

Fixes #8341.

This preserves structured SuperSync server `errorCode`s across the client provider boundary instead of forcing client code to reconstruct server outcomes from `Error.message` substrings.

## What changed

- Added shared SuperSync error-code contract exports in `@sp/shared-schema`.
- Typed response fields the server already returns:
  - `SuperSyncSnapshotUploadResponse.errorCode`
  - `SuperSyncUploadOpsResponse.deduplicated`
- Kept the SuperSync server's existing `SYNC_ERROR_CODES` import surface by aliasing it to the shared contract constants.
- Updated the SuperSync provider to parse non-2xx JSON response bodies once and attach the structured code to thrown HTTP errors as safe primitive metadata (`code` / `errorCode`).
- Updated client handling for:
  - `SYNC_IMPORT_EXISTS` during snapshot upload
  - `STORAGE_QUOTA_EXCEEDED`
  - `PAYLOAD_TOO_LARGE`
- Kept legacy message fallback paths for older/foreign errors.

## Root cause

The server already emitted structured `errorCode` values, but non-2xx responses were converted into generic `Error` messages by the provider. App code then branched on message substrings such as `SYNC_IMPORT_EXISTS` or quota text. That made behavior fragile and left parts of the response contract untyped.

## Verification

- `npm run checkFile` for all changed TypeScript files
- `npm run shared-schema:build && cd packages/shared-schema && npm test`
- `npm run sync-providers:test`
- `cd packages/super-sync-server && npm test -- --run tests/sync-compressed-body.routes.spec.ts`
- `cd packages/super-sync-server && npm run build`
- Focused Angular specs:
  - `npm run test:file src/app/op-log/sync/sync-error-utils.spec.ts`
  - `npm run test:file src/app/op-log/sync/operation-log-upload.service.spec.ts`
  - `npm run test:file src/app/imex/sync/sync-wrapper.service.spec.ts`
- Full `npm test`
- Pre-push hook reran lint and full tests successfully

Note: lint completed with pre-existing warnings in unrelated files:
- `src/app/features/work-context/work-context.const.spec.ts`
- `src/app/features/simple-counter/habit-tracker/habit-tracker.component.scss`


johannesjo:
## Review summary

Thanks for this — promoting `errorCode` into the typed contract is the right direction, and keeping every legacy message-substring path as a fallback is exactly right for cross-version safety. Verified the change end-to-end (provider error paths, the schema consumer in `response-validators.ts`, server emit sites) and the affected specs are green. **No correctness/data-loss issues** — the points below are forward-compat hardening, not blockers.

### WARNING 1 — Strict `z.enum` on the wire `errorCode` is a forward-compat lock-in

`packages/shared-schema/src/supersync-http-contract.ts` (`errorCode: SuperSyncErrorCodeSchema.optional()` on `SuperSyncSnapshotUploadResponseSchema`).

That field is parsed by `validateSnapshotUploadResponse` → `safeParse`, which throws `InvalidDataSPError` on any unknown enum member. `.passthrough()` only tolerates unknown *keys*; it does **not** relax validation of a *defined* key. Verified against this repo's zod 4.4.3:

```
strict enum + unknown value 'FUTURE_CODE'  → success: false   (rejected)
loose z.string() + same value              → success: true
absent errorCode (today's 2xx body)        → success: true
```

So if a newer server ever includes an out-of-set `errorCode` in a 2xx snapshot body, an **older client rejects the entire otherwise-valid response**. It's dormant today (the 2xx success body is `{accepted, serverSeq, error}` with no `errorCode`; the real codes ride non-2xx responses that the provider throws *before* schema validation), but it locks a sync wire format to a closed set. It's also asymmetric with its own siblings: the `error` field just above it and the heavily-branched `SuperSyncUploadResultSchema.errorCode` are both loose `z.string().optional()`. And the drift is already real — `rejected-ops-handler.service.ts:157` branches on `'CONFLICT_STALE'` (`// TODO: remove after all servers are updated`), a code **not** in the new 20-member enum, surviving only because it flows through the loose schema.

**Suggested fix (one line):** make the wire field `errorCode: z.string().optional()`. Keep `SUPER_SYNC_ERROR_CODES` / `SuperSyncErrorCode` as the producer + comparison vocabulary — every client consumer just does `=== 'KNOWN_SLUG'`, so nothing downstream loses type safety, and the existing `errorCode: 409` (number) contract test still throws under `z.string()`.

### WARNING 2 — Server-controlled `errorCode` is attached/logged without the length cap `reason` has

`packages/sync-providers/src/super-sync/super-sync.ts` (`_extractServerErrorDetails`).

`error`/`reason` is deliberately `.slice(0, SERVER_ERROR_REASON_MAX_CHARS)` (80) — the in-file comment notes it defends "against future contract drift that might embed user content" — but the new `errorCode` is taken raw and then flows into `SuperSyncHttpStatusError.code` and into exportable log metadata (the `_buildErrorMetadata` change intentionally bypasses `toSyncLogError`, which drops `.code` for `Error` instances by design). Bounded by the 400-char log-export cap and only reachable with an untrusted server (the official server emits short slugs), and there's **no XSS** (all consumers do slug equality, no interpolation) — so this is a defense-in-depth gap, not a live bug. **Suggested fix:** apply the same `.slice(0, SERVER_ERROR_REASON_MAX_CHARS)` to `errorCode` (matters a bit more if you adopt WARNING 1, since the enum is currently the only upstream bound).

### Optional polish (non-blocking)

- `SuperSyncHttpStatusError` stores the same value in both `.code` and `.errorCode`. `getSyncErrorCode` falls back `.errorCode → .code` and the log-meta branch reads `.code`, so one field would do — or add a one-liner explaining why both exist (the tests pin both).
- In zod 4, `z.enum(Object.values(SUPER_SYNC_ERROR_CODES) as [...])` can just be `z.enum(SUPER_SYNC_ERROR_CODES)`, dropping the unsafe cast (moot if you adopt WARNING 1).
- Consider dating the legacy fallbacks (`_getLegacyErrorCode`, the message branch in `isStorageQuotaError`) with a `// TODO: remove once all providers thread errorCode` marker, mirroring the existing `CONFLICT_STALE` TODO, so they're greppable deletion candidates.
- `getSyncErrorMessage`/`getSyncErrorCode` partly overlap `getErrorTxt`/`toSyncLogError`, but reuse is genuinely blocked because `toSyncLogError` drops `.code` on `Error` subclasses — so the hand-rolled narrowing is justified. A cross-ref comment would stop a future maintainer from "simplifying" it into the lossy helper.

### Verified correct (so reviewers don't re-flag)

- `getSyncErrorCode` precedence (`.errorCode` then `.code`) is safe — `SuperSyncHttpStatusError` sets both to the same value, and WebDav's numeric `errorCode` is correctly ignored by the `typeof === 'string'` guard.
- All `handleStorageQuotaError`/`isStorageQuotaError` call sites are updated and compatible with the `unknown` signature.
- PAYLOAD_TOO_LARGE reads the correct `rejectedOps[].errorCode` field, and the retained `'Payload too complex'` substring fallback is **load-bearing** (server emits `INVALID_PAYLOAD` for that case, not `PAYLOAD_TOO_LARGE`) — nice catch.
- Server `SYNC_ERROR_CODES` → shared alias is a strict additive superset; no new coupling (`@sp/shared-schema` is already a server dep).
- `_buildErrorMetadata` change is necessary (otherwise the structured code would be dropped from logs).
- Performance impact is negligible (cold error paths; `Object.values` is module-level; validation is per-batch/per-snapshot, never per-op).

**Bottom line:** good fix. The one change worth making before merge is relaxing the wire `errorCode` to `z.string().optional()` (WARNING 1) — it trades a fleet-wide, hard-to-reverse forward-compat liability for zero functional loss — ideally alongside capping `errorCode` length (WARNING 2). Everything else is optional.

<sub>🤖 Synthesized from a parallel multi-agent review (correctness / security / architecture / alternatives / performance / simplicity). Claims verified against the source and the repo's zod 4.4.3.</sub>

johannesjo:

Thanks for the PR! The direction is still right and #8341 is still open, but the branch has sat since the June review and master has moved under it – the PR now conflicts, and that review's main warning has gone from theoretical to live.

### Blocking
- **`packages/shared-schema/src/supersync-http-contract.ts:73`** – the strict `z.enum` on the wire `errorCode` (WARNING 1 in the earlier review) is no longer just forward-compat hygiene. Since June, master's server added `REPAIR_STALE` and `E2EE_REQUIRED` to `SYNC_ERROR_CODES` (`packages/super-sync-server/src/sync/sync.types.ts`), neither of which is in this PR's 20-member enum, so:
  - `SYNC_ERROR_CODES = SUPER_SYNC_ERROR_CODES` no longer compiles on rebase – the server references both codes (`sync.service.ts`, `sync.routes.snapshot-handler.ts`, `sync.routes.payload.ts`).
  - Worse, `REPAIR_STALE` is sent on an HTTP **200** snapshot body (`sync.routes.snapshot-handler.ts:336`), which flows through `validateSnapshotUpload` into `SuperSyncSnapshotUploadResponseSchema.safeParse`. With the strict enum, that valid response throws `InvalidDataSPError`, the code never reaches the rejected-ops handler, and a stale REPAIR op is permanently rejected instead of rebased – a live sync regression.
  - Fix on rebase: keep master's already-landed `errorCode: z.string().optional()` on the wire schemas and drop `SuperSyncErrorCodeSchema` from response validation; keep `SUPER_SYNC_ERROR_CODES` as the producer/comparison vocabulary, adding `REPAIR_STALE` + `E2EE_REQUIRED` if the server alias stays.
- **Merge conflict** – the PR is `CONFLICTING` against master, and every region it touches drifted: the snapshot catch moved to `operation-log-upload.service.ts:816` (still message-substring based, so that part of your change is still wanted), `hasPayloadError` moved to `sync-wrapper.service.ts:1220` (still substring-only), and the contract gained `existingClock` plus loose `errorCode` fields. The still-missing, still-valuable parts to carry through the rebase: the provider threading (`code`/`errorCode` on thrown HTTP errors), `getSyncErrorCode`/`isStorageQuotaError`, the `PAYLOAD_TOO_LARGE` errorCode check, and typing `deduplicated`.

### Nits
- **`packages/sync-providers/src/super-sync/super-sync.ts:570`** – `_extractServerErrorDetails` takes `errorCode` raw while the sibling `error` field is deliberately capped at `SERVER_ERROR_REASON_MAX_CHARS`; the uncapped value flows into `SuperSyncHttpStatusError.code` and, via `_getSafeErrorLogMeta`, into exportable log metadata (WARNING 2 in the earlier review). Same one-line `.slice(0, SERVER_ERROR_REASON_MAX_CHARS)` fix. I checked the paths that could make this behavioral – all consumers compare by equality and `isRetryableUploadError` never reads `.code` – so it stays defense-in-depth.
- **`packages/sync-providers/src/super-sync/super-sync.ts:60`** – `.code` and `.errorCode` hold the identical value; one field would do, or a one-line comment on why both exist.
- **`src/app/op-log/sync/operation-log-upload.service.ts:279`** – the ops-upload catch still calls `handleStorageQuotaError(message)` although `err` now carries the structured code; passing `err` uses the structured path consistently (the message fallback keeps it working either way).
- **`packages/shared-schema/src/supersync-http-contract.ts:73`** – zod 4 accepts `z.enum(SUPER_SYNC_ERROR_CODES)` directly, dropping the tuple cast (moot if the wire enum goes away per the blocking item).

---

johannesjo:
Thanks a lot for this PR @symonbaikov — the direction was exactly right! Since the branch had drifted and conflicted with master over time, I picked it up, rebased it onto current master (your commit and authorship are preserved), and addressed the remaining review items in #9775, which supersedes this PR. Closing this one in favor of #9775 — thanks again for the contribution!

PR #8590, opened 2026-06-25
[codex] Route plugin PATCH through native HTTP

### LINKED ISSUE #8579: PATCH requests fail on Android (Capacitor) — affects GitHub & Google Calendar issue providers
### Summary

On the Android (Capacitor) build, HTTP `PATCH` requests issued by issue-provider plugins likely fail. Capacitor's `CapacitorHttp` routes every non-`GET`/`HEAD`/`OPTIONS`/`TRACE` request through Java's `HttpURLConnection`, whose method allow-list is `GET/POST/HEAD/OPTIONS/PUT/DELETE/TRACE` — `PATCH` is rejected with `ProtocolException`. This is the same root cause as #8558 (which was about the WebDAV verbs `PROPFIND`/`REPORT`).

### Affected code

- `packages/plugin-dev/github-issue-provider/src/plugin.ts:373` — `http.patch(...)` (issue updates)
- `packages/plugin-dev/google-calendar-provider/src/plugin.ts:607,706` — `http.patch(...)` (event updates)

### Relation to #8558

#8558 fixed the same class of problem for CalDAV by routing the WebDAV verbs through the native OkHttp/URLSession `WebDavHttp` executor in `PluginHttpService` (`src/app/plugins/issue-provider/plugin-http.service.ts`). The rerouted set was deliberately scoped to `{PROPFIND, REPORT}` and intentionally **excludes** `PATCH`, because PATCH affects different providers with JSON round-trips that need on-device verification.

### Proposed fix

Add `PATCH` to `NATIVE_HTTP_METHODS` in `plugin-http.service.ts`. The native `WebDavHttp` plugin already supports PATCH (Android `WebDavHttpPlugin.kt`, iOS `WebDavHttpPlugin.swift`), and `_requestViaNativeHttp` already JSON-parses non-text responses, so the GitHub/Google-Calendar JSON round-trips should decode correctly.

### Before merging — verify on-device

- Confirm on a real Android device that GitHub issue updates / Google Calendar event edits currently fail (Android's `HttpURLConnection` is OkHttp-backed; confirm the PATCH rejection at runtime rather than assuming the Oracle-JDK allow-list).
- Confirm the rerouted PATCH JSON responses parse correctly for both providers.
- `PATCH` is non-idempotent: if transient-retry is ever added to the native path, exclude PATCH (or pass `maxRetries: 0`).

symonbaikov:
## What

Routes plugin `PATCH` requests through the existing native `WebDavHttp` executor on native platforms, alongside the existing `PROPFIND` and `REPORT` routing. This fixes Android issue-provider updates that use `PATCH`, including GitHub issue updates and Google Calendar event edits.

## Why

On Android, Capacitor routes non-basic HTTP methods through Java `HttpURLConnection`, which rejects `PATCH` with `ProtocolException`. The app already has an OkHttp/URLSession-backed native executor for methods that `HttpURLConnection` cannot handle.

## Validation

- Confirmed RED first: the new `PATCH` regression test failed while `PATCH` still used `HttpClient`.
- `npm run test:file src/app/plugins/issue-provider/plugin-http.service.spec.ts` — 26/26 passed.
- `npm run checkFile src/app/plugins/issue-provider/plugin-http.service.ts` — passed.
- `npm run checkFile src/app/plugins/issue-provider/plugin-http.service.spec.ts` — passed.
- `git diff --check` — passed.
- Pre-push hook ran `npm run lint` and `npm test`; lint had existing warnings in unrelated files, and tests passed, including the main Karma suite and timezone suite.

## Android verification

Android device verification was **not** run. Before merging, please verify on a real Android device that GitHub issue updates and Google Calendar event edits no longer fail.

Fixes #8579

### DIFF
--- a/src/app/plugins/issue-provider/plugin-http.service.spec.ts
+++ b/src/app/plugins/issue-provider/plugin-http.service.spec.ts
@@ -223,6 +223,7 @@ describe('PluginHttpService - native WebDAV/CalDAV method routing (#8558)', () =
   afterEach(() => httpMock.verify());
 
   const authHeaders = (): Record<string, string> => ({ Authorization: 'Basic abc' });
+  const GITHUB_API_VERSION_HEADER = 'X-GitHub-Api-Version';
 
   it('routes PROPFIND through the native executor (not HttpClient)', async () => {
     const http = service.createHttpHelper(authHeaders);
@@ -251,6 +252,40 @@ describe('PluginHttpService - native WebDAV/CalDAV method routing (#8558)', () =
     expect((nativeHttp.calls.mostRecent().args[0] as { method: string }).method).toBe(
       'REPORT',
     );
+  });
+
+  it('routes PATCH JSON requests through the native executor', async () => {
+    nativeHttp.and.resolveTo(
+      okResponse({
+        status: 200,
+        data: '{"id":"issue-1","title":"Updated"}',
+        url: 'https://api.github.com/repos/owner/repo/issues/1?state=open',
+      }),
+    );
+    const http = service.createHttpHelper(authHeaders);
+    const result = await http.patch<{ id: string; title: string }>(
+      'https://api.github.com/repos/owner/repo/issues/1',
+      { title: 'Updated' },
+      {
+        headers: { [GITHUB_API_VERSION_HEADER]: '2022-11-28' },
+        params: { state: 'open' },
+      },
+    );
+
+    expect(nativeHttp).toHaveBeenCalledTimes(1);
+    const cfg = nativeHttp.calls.mostRecent().args[0] as Record<string, unknown>;
+    expect(cfg['url']).toBe(
+      'https://api.github.com/repos/owner/repo/issues/1?state=open',
+    );
+    expect(cfg['method']).toBe('PATCH');
+    expect(cfg['data']).toBe('{"title":"Updated"}');
+    expect(cfg['responseType']).toBe('json');
+    expect((cfg['headers'] as Record<string, string>)['Authorization']).toBe('Basic abc');
+    expect((cfg['headers'] as Record<string, string>)[GITHUB_API_VERSION_HEADER]).toBe(
+      '2022-11-28',
+    );
+    expect(result).toEqual({ id: 'issue-1', title: 'Updated' });
+    httpMock.expectNone(() => true);
   });
 
   it('rejects with an error carrying .status on a non-2xx native response', async () => {
--- a/src/app/plugins/issue-provider/plugin-http.service.ts
+++ b/src/app/plugins/issue-provider/plugin-http.service.ts
@@ -22,15 +22,13 @@ const MAX_TIMEOUT = 120000;
  * instead, which accepts arbitrary method strings. Standard verbs keep using the
  * normal `HttpClient` path (it handles JSON + works on every platform).
  *
- * Scoped to the verbs the CalDAV calendar plugin actually uses (discovery +
- * event query). Entries MUST stay within the set the native `WebDavHttp` plugin
- * implements (e.g. it rejects `MKCALENDAR`). `PATCH` is also rejected by
- * `HttpURLConnection` and would belong here, but it's used by other providers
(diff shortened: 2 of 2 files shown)


johannesjo:
Thanks for tackling this! The routing change itself is correct (Android's `HttpURLConnection` does reject PATCH, and the native executor handles it), and the SSRF guard / no-retry posture all check out. But after a closer look I think **the fix is likely incomplete and may not actually fix #8579 on-device** — and the new test can't catch why.

## 🔴 JSON PATCH goes out with the wrong (Android) / missing (iOS) `Content-Type`

`_requestViaNativeHttp` `JSON.stringify`s the body but never injects a `Content-Type`, and the only two PATCH callers don't set one either:

- `github-issue-provider/src/plugin.ts` `getHeaders` → only `Accept` + `Authorization`
- `google-calendar-provider/src/plugin.ts` `getHeaders` → only `Authorization`

What each platform then sends for the same logical request:

| Path | Content-Type |
|------|-------------|
| Web/Electron (HttpClient, unchanged) | `application/json` (Angular auto-detects object bodies) |
| **Android native** (`WebDavHttpPlugin.kt:115-118`) | **`application/xml; charset=utf-8`** (the WebDAV default when caller sets none) |
| **iOS native** (`WebDavHttpPlugin.swift:42-46`) | **none** (only caller-supplied headers are set) |

So a JSON PATCH gets labeled as XML on Android and sent with no content type on iOS. Google Calendar (`googleapis.com`) is strict about request media type and will likely reject a JSON body labeled `application/xml` (400/415) — and that's exactly the edit path this PR is meant to fix. GitHub tends to be lenient and may tolerate it, but that's undocumented behavior. This matches the "Android device verification was **not** run" caveat — it's the kind of thing on-device testing would surface.

Worth noting: every *other* JSON-writing provider sets the header explicitly (`clickup:103`, `linear:201`, `azure-devops:267` all set `Content-Type: application/json`). GitHub + Google Calendar are the only two relying on Angular's implicit default — and the only two affected here.

### Suggested fix
Default the header in `_requestViaNativeHttp` when the body is a non-string object, `responseType` is json, and the caller set none — mirroring what Angular's `HttpClient` does on web:

```ts
// match HttpClient's implicit application/json for object bodies;
// the native WebDAV executor otherwise defaults to application/xml (Android) / none (iOS)
const isJsonBody = body != null && typeof body !== 'string';
const hasContentType = Object.keys(headers).some((h) => h.toLowerCase() === 'content-type');
if (isJsonBody && options?.responseType !== 'text' && !hasContentType) {
  headers = { ...headers, 'Content-Type': 'application/json' };
}
```

This fixes both platforms in one place and **cannot regress CalDAV** — PROPFIND/REPORT set their own `application/xml` explicitly, so the "only when absent" guard never touches them. (Setting it in the two plugins' `getHeaders` would also work as defense-in-depth, but the boundary fix is cleaner and covers future native JSON callers.)

## 🟡 The new unit test can't catch this
The test mocks `PLUGIN_HTTP_NATIVE_EXECUTOR` and asserts `data` / `method` / `responseType` / auth headers, but the Content-Type defaulting happens *below* that mock, in the native Kotlin/Swift layer. So it passes regardless of whether the real request succeeds. After the fix, please also assert `cfg.headers['Content-Type'] === 'application/json'` for an object body (and that an explicit caller Content-Type is preserved).

## Minor notes (non-blocking)
- The new doc comment says PATCH is kept "off any automatic transient retry path." True for this executor (it's the raw `APP_WEBDAV_NATIVE_HTTP`, not the retry-wrapped one) — but Google Calendar's `upsertEvent` already wraps its PATCH in an app-level rate-limit retry, so the framing is slightly broader than reality. Maybe narrow to "the shared native executor must not transparently retry PATCH; idempotency is the caller's responsibility."
- `options.timeout` is silently ignored on the native path (documented in the JSDoc, ~30s fixed). No impact today since neither caller sets one — just worth updating the comment from "CalDAV callers" to "CalDAV and issue-provider callers."

---

Net: routing is right, but I'd add the `application/json` default + a Content-Type assertion, and run the on-device check before merge — otherwise Google Calendar edits in particular may still fail on Android, just with a different error.

PR #8594, opened 2026-06-25
fix(sync): share quota pre-gate helper

### LINKED ISSUE #8353: SuperSync server: snapshot handler quota pre-gate duplicates enforceStorageQuota — share via allowCleanup option
**Area:** server (quota) · **Type:** simplification · **Value:** medium-low

## Problem
The snapshot upload handler's quota pre-gate (`sync.routes.snapshot-handler.ts:119-138`) hand-rolls the same skeleton `enforceStorageQuota` (`sync.routes.quota.ts:195-258`) already implements — check → reconcile once → re-check → 413 — minus the destructive-cleanup step. Two copies of quota-decision logic to keep in sync.

## Behavior deltas to preserve (verified — this is not a blind swap)
1. **Boundary semantics:** the pre-gate rejects at `used >= quota`; `checkStorageQuota` allows when `used + delta <= quota` (`storage-quota.service.ts:256`). Calling the shared helper with `delta = 0` flips the `used == quota` edge from reject to allow — use `delta = 1` or an inclusive-bound option.
2. `checkStorageQuota` honors forced-reconcile markers (`:238-245`) while the pre-gate's `getStorageInfo` (`:800-813`) is a raw cached read — a benign improvement, but a behavior change.
3. The 413 body must keep `autoCleanupAttempted: false` and omit `cleanupStats` on the no-cleanup path.

## Explicitly NOT a finding
The duplicated `findExistingSyncImport` check (out-of-lock `:148-164`, in-lock `:194-212`) is intentional and documented (`:140-147` — cheap rejection without holding the storage lock, then the TOCTOU re-check under it). Leave it alone.

## Proposed fix
Add `{ allowCleanup: false }` to `enforceStorageQuota` (skip `freeStorageForUpload`, emit `autoCleanupAttempted: false`), call it with inclusive-boundary semantics, and replace handler lines 119-138.

symonbaikov:
## Summary

Fixes #8353.

The snapshot upload handler had its own quota pre-gate that duplicated part of `enforceStorageQuota`: it read cached storage info, reconciled once, then returned a 413 response directly. This PR routes that early snapshot gate through the shared quota helper instead.

## What changed

- Added `allowCleanup?: boolean` to `enforceStorageQuota`, defaulting to the existing cleanup-enabled behavior.
- Switched the snapshot pre-gate to call `enforceStorageQuota(userId, 1, reply, { allowCleanup: false })`.
- Preserved the existing inclusive boundary where `used == quota` rejects before expensive snapshot preparation.
- Kept the early pre-gate non-destructive: no `freeStorageForUpload` call, `autoCleanupAttempted: false`, and no `cleanupStats` in that 413 response.
- Added regression coverage for reconcile-before-reject, no-cleanup 413 responses, and the `checkStorageQuota(1, 1)` pre-gate call.

## Validation

- `npm run checkFile packages/super-sync-server/src/sync/sync.routes.quota.ts`
- `npm run checkFile packages/super-sync-server/src/sync/sync.routes.snapshot-handler.ts`
- `npm run checkFile packages/super-sync-server/tests/sync-compressed-body.routes.spec.ts`
- `npm --prefix packages/super-sync-server test -- tests/sync-compressed-body.routes.spec.ts`
- `npm --prefix packages/super-sync-server test`
- `git diff --check`
- Pre-push hook: full lint/package tests, Angular test suite, and LA timezone test suite passed

### DIFF
--- a/packages/super-sync-server/src/sync/sync.routes.quota.ts
+++ b/packages/super-sync-server/src/sync/sync.routes.quota.ts
@@ -110,6 +110,10 @@ type ExistingSyncImport = {
   serverSeq: number;
 };
 
+export type EnforceStorageQuotaOptions = {
+  allowCleanup?: boolean;
+};
+
 /**
  * Look up an existing full-state op for the user.
  *
@@ -196,7 +200,9 @@ export async function enforceStorageQuota(
   userId: number,
   storageDeltaBytes: number,
   reply: FastifyReply,
+  options: EnforceStorageQuotaOptions = {},
 ): Promise<boolean> {
+  const { allowCleanup = true } = options;
   const syncService = getSyncService();
   let quotaCheck = await syncService.checkStorageQuota(userId, storageDeltaBytes);
   if (quotaCheck.allowed) return true;
@@ -207,7 +213,8 @@ export async function enforceStorageQuota(
   // upload success, not on quota misses. Also guards against pre-deploy stale
   // counters where the old code's pg_column_size SUM left an inflated value.
   Logger.info(
-    `[user:${userId}] Quota cache miss (cached: ${quotaCheck.currentUsage}/${quotaCheck.quota}). Reconciling before cleanup...`,
+    `[user:${userId}] Quota cache miss (cached: ${quotaCheck.currentUsage}/${quotaCheck.quota}). ` +
+      `Reconciling before ${allowCleanup ? 'cleanup' : 'rejecting'}...`,
   );
   try {
     await syncService.updateStorageUsage(userId);
@@ -220,8 +227,22 @@ export async function enforceStorageQuota(
     }
   } catch (err) {
     Logger.warn(
-      `[user:${userId}] Reconcile failed, proceeding with cleanup: ${errorMessage(err)}`,
+      `[user:${userId}] Reconcile failed, ${
+        allowCleanup ? 'proceeding with cleanup' : 'rejecting without cleanup'
+      }: ${errorMessage(err)}`,
     );
+  }
+
+  if (!allowCleanup) {
+    Logger.warn(
+      `[user:${userId}] Storage quota exceeded: ${quotaCheck.currentUsage}/${quotaCheck.quota} bytes. Cleanup disabled for this route.`,
+    );
+    sendQuotaExceededReply(reply, {
+      storageUsedBytes: quotaCheck.currentUsage,
+      storageQuotaBytes: quotaCheck.quota,
+      autoCleanupAttempted: false,
+    });
+    return false;
   }
 
   Logger.warn(
--- a/packages/super-sync-server/src/sync/sync.routes.snapshot-handler.ts
+++ b/packages/super-sync-server/src/sync/sync.routes.snapshot-handler.ts
@@ -14,7 +14,6 @@ import {
 import type { SnapshotDedupResponse } from './services';
 import {
   createValidationErrorResponse,
-  errorMessage,
   MAX_COMPRESSED_SIZE_SNAPSHOT,
   MAX_DECOMPRESSED_SIZE_SNAPSHOT,
   sendCompressedBodyParseFailure,
@@ -26,7 +25,6 @@ import {
   enforceStorageQuota,
   findExistingSyncImport,
   isIdempotentSyncImportRetry,
-  sendQuotaExceededReply,
   sendSyncImportExistsReply,
 } from './sync.routes.quota';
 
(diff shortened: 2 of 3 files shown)


johannesjo:
## Review — multi-agent pass (correctness, security, architecture, performance, simplicity)

The de-duplication is sound and the behavioral contract from #8353 is faithfully preserved (boundary, 413 body, non-destructive early path — all verified). But there's one **build-breaking blocker** that the green CI doesn't catch.

### 🔴 Blocker — `errorMessage` is used but its import was removed → `tsc` build fails

`packages/super-sync-server/src/sync/sync.routes.snapshot-handler.ts`

The diff drops `errorMessage` from the `./sync.routes.payload` import (it removed the pre-gate's use of it), but it's **still called at line 331** in the outer catch:

```ts
} catch (err) {
  Logger.error(`Upload snapshot error: ${errorMessage(err)}`);  // ← no longer in scope
  return reply.status(500).send({ error: 'Internal server error' });
}
```

`errorMessage` is module-local to `sync.routes.payload.ts` (not a global), so after the import removal this is a hard `tsc` error:

```
error TS2304: Cannot find name 'errorMessage'.
```

**Why CI is green:** the PR checks run only `vitest run` (esbuild — strips types, no typecheck) + eslint (`no-undef` is off in TS projects). But the package `build` is `tsc`, and the production `Dockerfile` runs `npm run build` (line 36) — so the **release/Docker build breaks**. It's also a latent runtime `ReferenceError` on the snapshot-upload error path, which no test exercises.

**Fix (one line)** — restore the import:

```ts
import {
  createValidationErrorResponse,
  errorMessage,
  MAX_COMPRESSED_SIZE_SNAPSHOT,
  MAX_DECOMPRESSED_SIZE_SNAPSHOT,
  sendCompressedBodyParseFailure,
} from './sync.routes.payload';
```

> Worth considering a follow-up: a CI `tsc --noEmit` step for `packages/super-sync-server` would have caught this — the current server PR gate is vitest-only, so type errors on untested lines slip through.

### 🟡 Worth addressing

- **Stale docstring.** `enforceStorageQuota`'s doc (`sync.routes.quota.ts:195`) still reads *"…and auto-cleanup if needed,"* which is no longer true on the `allowCleanup: false` path. Please update it.
- **Flag-argument shape (optional).** `allowCleanup` makes `enforceStorageQuota` serve two behaviors interleaved by a boolean (plus the two `allowCleanup ? … : …` log ternaries). It works, but extracting the shared `check → reconcile → re-check` prefix into a helper and keeping two named entry points would read cleaner — and it matches the file's existing convention (`enforceCleanSlateStorageQuota` is already a separate sibling, not a flag). Judgment call; fine to keep as-is if you prefer a single entry point.
- **Destructive default.** `allowCleanup` defaults to `true`, i.e. omitting the option opts into restore-point deletion. A one-line comment on `EnforceStorageQuotaOptions` noting "default performs destructive cleanup" would prevent a future caller from silently re-enabling it.

### 🟢 Nits

- The `delta = 1` boundary trick is load-bearing (preserves `used >= quota`); a named constant with the "why not 0" rationale would protect it from a future "simplify to 0."
- Minor behavior change worth a note in the description: the new path's `checkStorageQuota` honors `forcedReconciles` (the old `getStorageInfo` didn't), so a marked-stale user now reconciles on the early pre-gate. Bounded + self-clearing, arguably a self-heal improvement.

### ✅ Verified correct
- `delta=1` preserves the `used >= quota` rejection exactly (including the `used == quota` edge).
- No-cleanup 413 body keeps `autoCleanupAttempted: false` and omits `cleanupStats`; tests assert it.
- Early path never reaches `freeStorageForUpload` (non-destructive); reconcile/throw paths behave identically to the old hand-rolled gate and never bubble a 500.
- No quota bypass, no user content logged, common-path DB round-trips unchanged (1 → 1).

**Verdict: needs the one-line import fix before merge; the rest is optional polish.**

johannesjo:

Thanks for the PR! The de-duplication itself is sound – I re-verified that `delta = 1` preserves the old `used >= quota` rejection exactly (`allowed = used + additionalBytes <= quota`) and that the no-cleanup 413 path matches the old hand-rolled gate – but two things stand between this and merge.

### Blocking
- **`packages/super-sync-server/src/sync/sync.routes.snapshot-handler.ts:331`** – `errorMessage(err)` is still called in the outer catch, but the PR removed `errorMessage` from the `./sync.routes.payload` import, so `tsc` fails with TS2304 (confirmed by compiling the head file). CI stays green because the PR checks only run vitest (esbuild strips types without checking them), but the package `build` script is `tsc` and the production Dockerfile runs `npm run build`, so the release/Docker build breaks. This is the same one-line fix the maintainer asked for on June 25 – restore `errorMessage` in the import list.
- **Merge conflict with `master`** – the PR is currently unmergeable, and the conflict is semantic, not just textual: since this branched, master's pre-gate guard became `!shouldCleanSlate && snapshotOpType !== 'REPAIR'` (REPAIR snapshots skip the cheap gate so a response-loss retry can reach the durable op-id idempotency check inside the lock), and `isCleanSlate` was renamed. When rebasing, keep master's guard condition around the new `enforceStorageQuota(userId, 1, reply, { allowCleanup: false })` call – resolving the conflict by keeping this branch's `!isCleanSlate` guard would silently drop the REPAIR skip, and an at-quota REPAIR retry would then be 413'd before reaching the idempotency check.

### Nits
- **`packages/super-sync-server/src/sync/sync.routes.quota.ts:196`** – the docstring still reads "and auto-cleanup if needed", which no longer holds on the `allowCleanup: false` path; worth fixing alongside the maintainer's earlier suggestion of a "default performs destructive cleanup" note on `EnforceStorageQuotaOptions`.

---

PR #8596, opened 2026-06-25
fix(sync): invalidate unsynced cache on failed rejection

symonbaikov:
## Summary
- Invalidate the unsynced operation cache when `markFailed()` terminally rejects an operation after reaching `maxRetries`.
- Add a regression test that first populates `getUnsynced()` cache, then verifies a terminally rejected op disappears from the cached result.

## Root cause
`markFailed()` can set `rejectedAt` when retry count reaches `maxRetries`, but unlike `markSynced()`, `markRejected()`, and `clearUnsyncedOps()`, it did not invalidate `_unsyncedCache`. If the cache was already populated, a terminally rejected local op could stay in `getUnsynced()` until some later cache rebuild.

## Validation
- RED: new regression test failed before the fix with stale cache (`Expected 2 to be 1`).
- `npm run checkFile src/app/op-log/persistence/operation-log-store.service.ts`
- `npm run checkFile src/app/op-log/persistence/operation-log-store.service.spec.ts`
- `npm run test:file src/app/op-log/persistence/operation-log-store.service.spec.ts` → `185 SUCCESS`
- Pre-push hook: lint completed with existing unrelated warnings only; package tests passed; Angular suite `11315 SUCCESS`; LA timezone suite `11301 SUCCESS`.

### DIFF
--- a/src/app/op-log/persistence/operation-log-store.service.spec.ts
+++ b/src/app/op-log/persistence/operation-log-store.service.spec.ts
@@ -1775,6 +1775,22 @@ describe('OperationLogStoreService', () => {
       expect(unsynced2[0].op.id).toBe(op2.id);
     });
 
+    it('should invalidate cache when markFailed terminally rejects an op', async () => {
+      const op1 = createTestOperation({ entityId: 'task1' });
+      const op2 = createTestOperation({ entityId: 'task2' });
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! <3

PR #8597, opened 2026-06-25
[codex] Fix SuperSync quota gate for duplicate ops

### LINKED ISSUE #8339: SuperSync server: quota gate charges duplicate/rejected ops — spurious 413 can trigger destructive restore-point cleanup
**Area:** server (storage quota) · **Severity:** low-medium (rare, but the failure mode deletes restore points for bytes that are never written)

## Problem
The `/ops` upload quota gate sizes the **entire incoming batch** before upload, with no subtraction for ops that will be rejected as duplicates or conflicts (and thus never stored). Duplicate detection happens later, inside `uploadOps`' transaction. The request-dedup cache shields only retries within 5 minutes that reuse the same `requestId`. So a delayed retry of a mostly-already-accepted batch by a near-quota user can spuriously 413 — and the gate path includes **destructive auto-cleanup** (`freeStorageForUpload` deletes the oldest restore points and their ops) to make room for bytes that won't actually be written.

## Evidence
- Gate charges all ops: `packages/super-sync-server/src/sync/sync.routes.ops-handler.ts:173-183` (`computeOpsStorageBytes` over the full array → `enforceStorageQuota`).
- Ordering: rate-limit (`:98`) → requestId dedup (`:115-164`) → gate (`:173-183`) → `uploadOps` (`:190`); duplicate classification only inside the transaction (`operation-upload.service.ts:353-407` batch, `:631-683` serial).
- Destructive cleanup: `storage-quota.service.ts:651-794` (`freeStorageForUpload`) → `deleteOldestRestorePointAndOps` (`:526-640`): `deleteMany` of the oldest SYNC_IMPORT/BACKUP_IMPORT/REPAIR restore point and all ops `serverSeq <= deleteUpToSeq`, may null the snapshot cache (`:612-620`).
- Mitigations (why severity is bounded): the dedup check deliberately runs before the gate (`:112-114`); `enforceStorageQuota` does an exact reconcile + re-check before cleanup (`sync.routes.quota.ts:209-225`); cleanup always keeps ≥1 restore point (`:748-759`). Accepted-byte accounting itself stays correct (`sync.service.ts:197-201`) — only the gate over-charges.

## Proposed fix
Subtract known-duplicate ops before gating: one pre-gate `findMany` on incoming op ids (reuse `DUPLICATE_OP_SELECT` + `isSameDuplicateOperation`) inside the storage lock, subtracting matched ops' bytes from `estimatedDelta`. Note the existing stage-3 prefetch runs inside `uploadOps`' transaction *after* the gate, so it can't be reused as-is. Keep the gate conservative for id collisions (don't subtract unless content matches). Conflict-rejected ops remain over-charged — acceptable, since they're not predictable pre-transaction.

symonbaikov:
## Summary

Fixes #8339.

This PR updates the SuperSync `/ops` quota pre-gate so it does not reserve storage for operations that are already persisted exact duplicates. The route now performs one pre-gate lookup for incoming operation ids, compares matches with the same duplicate comparator used by the upload path, and subtracts only exact duplicates from the estimated quota delta.

## Root cause

The `/ops` route previously computed the quota gate from the entire incoming batch before `uploadOps` classified already-stored duplicate operations. Near quota, a batch containing duplicate or retry operations could therefore trigger a 413 and destructive restore-point cleanup for bytes that would never be written.

## What changed

- Added a duplicate-aware quota sizing helper for ops uploads.
- Reused `DUPLICATE_OP_SELECT` and `isSameDuplicateOperation` so pre-gate duplicate detection matches upload duplicate detection.
- Exposed the effective `maxClockDriftMs` from `SyncService` for consistent duplicate timestamp handling.
- Kept id collisions and conflict-rejected ops conservatively charged.
- Added route-level regression tests for exact duplicates, id collisions, and duplicate-only near-quota uploads.

## Validation

- `npm test -- tests/sync-compressed-body.routes.spec.ts` -> 28 passed
- `npm run checkFile packages/super-sync-server/src/sync/sync.routes.quota.ts`
- `npm run checkFile packages/super-sync-server/src/sync/sync.routes.ops-handler.ts`
- `npm run checkFile packages/super-sync-server/src/sync/sync.service.ts`
- `npm run checkFile packages/super-sync-server/tests/sync-compressed-body.routes.spec.ts`
- pre-commit hook: lint completed with existing unrelated warnings only
- pre-push hook: full package tests, release notes tests, Angular tests, and LA timezone tests passed

### DIFF
--- a/packages/super-sync-server/src/sync/sync.routes.ops-handler.ts
+++ b/packages/super-sync-server/src/sync/sync.routes.ops-handler.ts
@@ -25,7 +25,7 @@ import {
   sendCompressedBodyParseFailure,
 } from './sync.routes.payload';
 import {
-  computeOpsStorageBytes,
+  computeOpsStorageBytesExcludingKnownDuplicates,
   enforceStorageQuota,
   getRawOpsCount,
   sendOpsBatchTooLargeReply,
@@ -168,10 +168,16 @@ export const uploadOpsHandler = async (
         // Check storage quota before processing (after dedup to allow retries).
         // Account using the same per-op payload+vectorClock measure that the
         // post-accept counter increment uses, so the gate and the increment
-        // cannot disagree on what "size" means.
+        // cannot disagree on what "size" means. Already-stored exact
+        // duplicates are rejected by uploadOps and never written, so don't make
+        // quota cleanup reserve space for them.
         const typedOpsForGate = ops as unknown as Operation[];
         const { bytes: estimatedDelta, fallback: gateFallback } =
-          computeOpsStorageBytes(typedOpsForGate);
+          await computeOpsStorageBytesExcludingKnownDuplicates(
+            userId,
+            typedOpsForGate,
+            syncService.getMaxClockDriftMs(),
+          );
         if (gateFallback > 0) {
           Logger.warn(
             `computeOpsStorageBytes: ${gateFallback}/${typedOpsForGate.length} unserializable op(s) ` +
--- a/packages/super-sync-server/src/sync/sync.routes.quota.ts
+++ b/packages/super-sync-server/src/sync/sync.routes.quota.ts
@@ -3,7 +3,13 @@ import { prisma } from '../db';
 import { Logger } from '../logger';
 import { getSyncService } from './sync.service';
 import { computeOpStorageBytes } from './sync.const';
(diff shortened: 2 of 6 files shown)


johannesjo:
Thank you very much! <3

Issue #8243, opened 2026-06-10
🚨 The SP Android app causes battery drain


bbinet:
### Version Used

version 18.8.0

### Current Behavior

The SP android app causes battery drain even when I disabled battery background usage.

<img width="540" height="1200" alt="Image" src="https://github.com/user-attachments/assets/69f194f7-d619-4b7a-810e-00463b8a6d4c" />

I've also try to kill the app, and even with the app killed, Super-productivity continue to drain battery on my phone, this is really strange.

<img width="540" height="1200" alt="Image" src="https://github.com/user-attachments/assets/888d85e8-afb9-4138-a2ab-2f224b4814fb" />

On the above screenshot, it is written: "Usage time: less than one minute" (I had killed the SP app, but it contunues to drain battery).

### Expected Behavior

Battery should not drain battery.

### Steps to Reproduce

Just the normal use of SP app on android.

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell
How can I further investigate the cause of this problem?
```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

johannesjo:
Thanks for the report. I dug into this and found a very likely cause: since v18.8.0 the time-tracking / focus-timer foreground services intentionally keep running when you swipe the app away (so tracking isn't silently lost, #7818) — but they also updated their notification every single second, which keeps the CPU busy around the clock. That exactly matches "usage time: less than one minute" with ongoing background drain. A fix is in progress that lets the system render the ticking timer instead, so the app does no per-second work anymore.

To confirm this is what you're hitting (and to rule out a second candidate), could you answer a few questions?

1. When the drain happens, is there a persistent Super Productivity notification visible (time tracking or focus timer)? Were you tracking time when you "killed" the app?
2. How do you kill the app — swiping it away from recents, or Settings → Apps → Super Productivity → Force stop? Does the drain continue even after a Force stop?
3. Do you use SuperSync on the phone? If so, the app also runs a background reminder-sync check every 15 minutes (network access), which is the other candidate — it runs even with the app swiped away.
4. Device model and Android version?

If you're comfortable with adb, this output would settle it definitively:

```
adb shell dumpsys batterystats --charged com.superproductivity.superproductivity
[...]

PR #8608, opened 2026-06-26
feat(focus-mode): loop the break-end sound until the break is dismissed

### LINKED ISSUE #8593: 💡 Optional looping break-end alarm that sounds until manually dismissed
### Problem Statement

When a break is taken (via Pomodoro or the break reminder), there is currently no signal that the break has *ended*. The existing reminder fires once as a standard system notification, which has two limitations for the moment a break is over:

1. It's a single, transient notification. If you've stepped away from the screen (which is the point of a break) a one-shot notification is easy to miss entirely. You only see it if you happen to be looking.

2. It requires no acknowledgement to clear. There's nothing that persists until you actually return and act on it.

The effect is that short breaks reliably overrun: a planned 5-minute break becomes 15+ because nothing actively pulls you back. For people like me, who rely on Super Productivity precisely because their own time-sense is unreliable (ADHD), "remember to come back on your own" defeats the purpose.

### Possible Solution

Add an *optional* "break-end alarm" that, when a break timer reaches zero:

- plays a sound that LOOPS (repeats continuously) rather than firing once, and
- continues until the user explicitly dismisses it via an action (e.g. a "Back to work" / "Resume" button).

Suggested behavior, all opt-in via settings:

- Toggle to enable/disable the looping break-end alarm (default off, to avoid surprising existing users).
- Choice of sound + volume.

The key property is dismiss-by-action: it must demand acknowledgement, so it reaches the user even when they're away from the screen.

_No response_

### Additional context

_No response_

costajohnt:
Implements the opt-in looping break-end alarm discussed in #8593.

## What

Adds a single, default-off setting that keeps the focus-mode break-end sound looping until the break is dismissed, instead of firing once. Useful when you step away from your desk and want a persistent cue that the break is over.

## Scope / decisions (per the discussion on #8593)

- One toggle only. Reuses the existing break-end sound (positive.mp3) and the global sound volume. No new sound picker or per-feature volume.
- Stored per-device in localStorage, not the synced config, mirroring `TaskWidgetSettingsService`. Looping audio behaves differently across platforms (desktop keeps the AudioContext running while the window is unfocused; mobile suspends it on app-background, #8243), so a single synced value would behave differently per device. The setting is labelled as local to the device.
- Limited to the focus-mode break timer reaching zero (`detectBreakTimeUp$`). The "take a break" reminder banner has no fixed end and is out of scope.
- Any existing leave-break action (resume / skip / start next / back to planning) dismisses it. No new dismissal UI.
- A single selector-based effect owns the loop lifecycle (mirrors `whiteNoiseSound$`), so the loop starts once and stops on any leave-break transition.
- Hard 10-minute safety ceiling so it can't sound forever into an empty room.
- The audio primitive stops any previous source before starting and uses a monotonic start-token, so a restart can never leak a second loop.

I confirmed the loop survives desktop window blur: `suspendAudioContext()` is only called from the Android/iOS app-background handlers in `main.ts`, never on desktop blur.

## Tests

- New unit tests for the looping audio primitive (loop flag, stop, restart cleanup, 10-min ceiling) and the per-device settings service.
- All existing focus-mode effects tests pass; the full suite passes via the pre-push hook.

Closes #8593


src/app/features/config/global-config-form-config.const.ts
```diff
@@ -12,6 +12,7 @@ import { IS_ELECTRON } from '../../app.constants';
 import { IS_ANDROID_WEB_VIEW } from '../../util/is-android-web-view';
 import { SCHEDULE_FORM_CFG } from './form-cfgs/schedule-form.const';
 import { FOCUS_MODE_FORM_CFG } from './form-cfgs/focus-mode-form.const';
```
johannesjo, line 14:
```suggestion
import { FOCUS_MODE_LOCAL_FORM_CFG } from './form-cfgs/focus-mode-local-form.const';
import { IS_NATIVE_PLATFORM } from '../../util/is-native-platform';
```
Adds the import the gate below needs (the file doesn't import `IS_NATIVE_PLATFORM` yet).

src/app/features/config/global-config-form-config.const.ts
```diff
@@ -55,6 +56,7 @@ export const GLOBAL_IMEX_FORM_CONFIG: ConfigFormConfig = [
 
 export const GLOBAL_PRODUCTIVITY_FORM_CONFIG: ConfigFormConfig = [
   FOCUS_MODE_FORM_CFG,
+  FOCUS_MODE_LOCAL_FORM_CFG,
```
johannesjo:
```suggestion
  ...(IS_NATIVE_PLATFORM ? [] : [FOCUS_MODE_LOCAL_FORM_CFG]),
```
Show this section only on desktop (Electron + web). On Android/iOS `suspendAudioContext()` stops the loop the moment the app backgrounds, so the "loop until I dismiss" toggle can't function there. Requires the `IS_NATIVE_PLATFORM` import from the suggestion above.

johannesjo (commented):
Reviewed against the scoping agreed on #8593 — the approach looks right as scoped, and I checked the sync/compat angle.

**Forwards/backwards compatibility — ✅**
`focusModeLocal` can't leak into synced state or the op-log:
- It's **not** part of `GlobalConfigState` (same exclusion `taskWidget` has).
- `updateGlobalConfigSection` only accepts the narrower `GlobalConfigSectionKey`, so dispatching `focusModeLocal` is a compile error → no phantom `GLOBAL_CONFIG` ops.
- `saveGlobalCfg()` intercepts it to `FocusModeLocalSettingsService` before any dispatch (read side too).
- New `localStorage` key defaults to off with a corrupt-JSON fallback, so old↔new clients exchange nothing for it.

Mirrors the already-shipped `taskWidget` pattern, so no new typia/sync-schema risk.

**One change before merge: gate the section off native mobile.**
On Android/iOS the app-background handler in `main.ts` (`suspendAudioContext()`, lines 493/513) stops the loop the instant the app backgrounds, and the alarm deliberately doesn't set `keepAwake` — so on mobile the loop dies exactly in the "stepped away from the device" case the toggle promises to cover. The setting then shows a switch that can't keep its promise. The inline suggestions scope it to desktop (Electron + web), where the AudioContext keeps running while the window is unfocused.

Also opened #8629 to track the broader "global how-reminders-fire" setting that came up in review — out of scope here; this PR stays the focused per-device break-end toggle.

Thanks for the clean, well-tested PR, and for flagging scope on the issue first 🙏

johannesjo:
Thank you very much! <3

PR #8610, opened 2026-06-27
feat(markdown): continue dated bullets with today's date on Enter

### LINKED ISSUE #8602: 💡 Sequential dates
### Problem Statement

I use the notes feature quite a lot and very intensively when working on tasks (in Board view) to manage projects.

One thing that helps me particularly with this is the ongoing documentation of status updates, news, etc. You can think of it as log entries. It looks something like this, for example:

## Status
- 18.06.: FM received a reply from Department XY: ...
- 20.06.: Phone call with AS: ...
- 21.06.: ...
...

This approach really helps me to keep track of what actions/results etc. took place in the past and when.

However, if you use it frequently, it does take a bit of time and you can make typing errors

### Possible Solution

If, at a line break, the current line begins with "- XX.XX.: ", then "XX.XX." is recognised as a date (dd.MM.) and the new line begins in the same format, but with the current date.

_No response_

### Additional context

_No response_

costajohnt:
Implements the "Sequential dates" continuation from #8602, following your spec.

## What

Pressing Enter on a dated bullet like `- 18.06.: Phone call` now starts the next line with today's date in the same format (`- 26.06.: `), extending the existing markdown list continuation. Pressing Enter on an empty dated entry exits the list in one keystroke, exactly like a plain bullet.

## Scope / decisions (per the spec on #8602)

- **Extends existing behavior, not new behavior** — a `- <date>:` line already continues as `- ` today; this just fills in today's date.
- **Format mirrored from the line**, no locale config: dotted `dd.MM.` / `dd.MM.yyyy` and ISO `yyyy-MM-dd`. Slashed (ambiguous field order), textual, 2-digit-year, and out-of-range dates fall back to the plain bullet continuation rather than guessing.
- **Always-on, no setting** — it rides on already-always-on bullet continuation and is trivially reversible (Ctrl+Z / edit the date).
- **Plain bullets only in v1** — checkbox and numbered lists continue unchanged.
- **Detection guard**: requires a `<date>: ` prefix (colon + space) and validates day 1-31 / month 1-12, so version bullets without a colon and time stamps (`14:30:`) don't mis-fire. (Range-only validation by design: a calendar-invalid but in-range source like `31.02.` is mirrored rather than rejected; today is always regenerated correctly.)
- **Cursor inside the date** falls back to plain bullet continuation (never returns null).
- `today` is threaded in from `DateService.getLogicalTodayDate()` at the call sites so the inserted date respects the configured start-of-day; the date util itself reads no clock and stays pure/deterministic.

## Implementation

- New pure util `date-prefix.util.ts` (`parseDatePrefix` + `formatTodayPrefix`) so `markdown-toolbar.util.ts` doesn't grow a second concern.
- `handleEnterKey` gains a `today` param, threaded through `handleListKeydown`; both call sites (`inline-markdown`, `dialog-fullscreen-markdown`) pass `getLogicalTodayDate()`.

## Tests

- `date-prefix.util.spec.ts`: format detection, padding, range/colon/timestamp/slashed/textual/2-digit-year fallbacks, unpadded variants.
- `markdown-toolbar.util.spec.ts`: each format mirrors today (not the source date), mid-content split, empty-entry exit, indentation, cursor-guard boundary, checkbox/numbered no-fill, Shift/Ctrl/Meta+Enter unaffected.
- Component test asserting the continuation is sourced from `DateService`.

Closes #8602

### DIFF
--- a/src/app/ui/dialog-fullscreen-markdown/dialog-fullscreen-markdown.component.spec.ts
+++ b/src/app/ui/dialog-fullscreen-markdown/dialog-fullscreen-markdown.component.spec.ts
@@ -131,6 +131,32 @@ describe('DialogFullscreenMarkdownComponent', () => {
       expect(component.close).toHaveBeenCalled();
     });
 
+    it('continues a dated bullet using logical-today from DateService (#8602)', () => {
+      // Spy returns a date far from the real "today" so the assertion also fails
+      // if a regression sourced the date from `new Date()` instead.
+      const dateServiceSpy = spyOn(
+        component['_dateService'],
+        'getLogicalTodayDate',
+      ).and.returnValue(new Date(2020, 0, 15)); // 15 Jan 2020
+      mockTextarea.value = '- 18.06.: Phone call';
+      mockTextarea.setSelectionRange(
+        mockTextarea.value.length,
+        mockTextarea.value.length,
+      );
+      const event = {
+        key: 'Enter',
+        ctrlKey: false,
+        metaKey: false,
+        shiftKey: false,
+        preventDefault: jasmine.createSpy('preventDefault'),
+      } as unknown as KeyboardEvent;
+
+      component.keydownHandler(event);
+
+      expect(dateServiceSpy).toHaveBeenCalled();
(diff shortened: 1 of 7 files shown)


johannesjo:

Issue #7511, opened 2026-05-07
💡 Monitor mailbox and auto-create new tasks from incoming emails


dualbound:
### Problem Statement

When receiving emails in any personal mailbox that directly can be translated into a new task, I would love to forward the corresponding email to another mailbox that takes care of creating a new task for my SuperProductivity task inbox. 

### Possible Solution

- Add access to a mailbox via IMAP (read-only). The user can have mytask@something.com to forward emails to.
- Monitor mailbox for incoming mails
- Create a new task, e.g., with similar/same properties as discussed in #518.
- Optional: delete email in the mytasks@something.com mailbox

### Describe alternatives you've considered

Alternatively, one could use the existing Jira/GitLab integrations for creating issues based on email. Without access to or setting up an instance, this, however, is less accessible.

### Additional context

A similar feature is present in [paperless-ngx](https://github.com/paperless-ngx/paperless-ngx) to [process documents attached to an email](https://docs.paperless-ngx.com/usage/#usage-mobile_upload).

johannesjo:
Should make for a great plugin!

dualbound:
I looked into this idea again. I think the feature is a very good fit conceptually, but there are two important constraints for implementing it as a [community plugin](https://github.com/super-productivity/super-productivity/blob/master/docs/plugin-development.md).

For a generic mailbox integration, IMAP is probably the right protocol because it works with most providers and self-hosted mailboxes. However, IMAP access requires credentials, and currently I do not see a generic local-only secret storage API for community plugins.

`persistDataSynced()` and issue provider `pluginConfig` do not seem appropriate for IMAP passwords or app passwords, because those values may become part of normal app data, sync, exports, or backups. Existing plugin OAuth support is local-only and works well for providers like Google Calendar, but it does not solve generic IMAP credentials.

There is also a platform limitation:
- Desktop could likely support IMAP via `nodeExecution`, after explicit user consent.
- Web cannot support generic IMAP directly, because browser JavaScript has no TCP/TLS socket access.
- Mobile would need native support/a bridge, since `nodeExecution` is desktop-only.

So my current conclusion is: this should either be implemented as a built-in/native-supported feature, or the plugin API would first need a small local-only secret API, e.g. `saveSecret/loadSecret/deleteSecret`, isolated per plugin and backed by OS keychain/safe storage where possible.

[...]

Issue #5530, opened 2025-11-16
💡 Integration with IMAP folder


blefuscu:
### Problem Statement

I treat my mailbox/inbox as my TODO list. Then I send myself e-mails with a subject "TODO: ..." so as not to forget about something.

Also, sometimes it's great to be able to send an e-mail to an e-mail address as a todo (often from a mobile device/on the go/from 3rd party device/etc.)

For such scenarios it would be useful to have a SP integration with a remote e-mail folder via IMAP.

### Possible Solution

Add a standard integration for e-mail accounts via IMAP. The idea is to configure the connector to point to a specific IMAP folder on a remote e-mail server and upon loading, display in the "Issue Panel" the subjects of the e-mails in the folder. 

Even better would be if the loaded e-mails were filtered by the subject/sender.

Note: for this to work well, it needs to fully respect secure IMAP connection via STARTLS, SSL+TLS with passwords, specific ports, etc. - the same way as e.g., Thunderbird has it.

PR #8633, opened 2026-06-29
feat(plugins): add local-only secret storage API for plugins

johannesjo:
## What

Adds a local-only, per-plugin **secret-storage API** for community plugins: `setSecret(key, value)` / `getSecret(key)` / `deleteSecret(key)`.

This is the missing building block identified in #7511 (and #5530): a place for plugin credentials (IMAP passwords, API tokens, app passwords) that is **not** synced. `persistDataSynced` and issue-provider `pluginConfig` land in synced state, exports, and backups, so they're wrong for secrets; plugin OAuth only covers OAuth providers, not generic password auth.

## How

- Backed by a dedicated `sup-plugin-secrets` IndexedDB that is **never part of sync, the op-log, exports, or backups** — mirroring the existing plugin OAuth token store (`plugin-oauth-token-store.ts`).
- **Namespaced per plugin** via the existing `composeId` / `isPluginIdMatch` key utils — the host injects the `pluginId`, so a plugin cannot address another plugin's keys, and cleanup is prefix-collision-safe (`plugin-a` ≠ `plugin-ab`).
- **Purged on uninstall.** While here, plugin **OAuth tokens are now purged on uninstall too** (they previously leaked past uninstall); both credential purges run first and best-effort so a later cleanup failure can't skip them.
- **Per-device**: a secret set on one device isn't synced to another, so plugins prompt for it once per device (matches how IMAP app-passwords are typically used).
- **Plaintext at rest** for now, exactly like the existing OAuth tokens. OS-keychain encryption (Electron `safeStorage` / Android Keystore / iOS Keychain) can be layered in later behind the same API with **no plugin changes**.

## Issue-provider plugins

Keep non-secret connection details (host, port, username, filters) in the synced config as usual, and put the password/token in secret storage. The provider definition callbacks (`getHeaders`, `getById`, …) run in the plugin's own context, so they read the secret directly when building requests. Documented in `docs/plugin-development.md`.

## Why this unblocks #7511

On desktop a mailbox plugin can use `nodeExecution` for the IMAP/TLS connection and secret storage for the credentials. (Web has no TCP sockets and mobile would need a native bridge, so a community IMAP plugin would be desktop-first.)

## Tests

- `plugin-secret.service.spec.ts` (10): round-trip, per-plugin namespacing, single-delete isolation, owner-only uninstall purge, prefix-collision safety, and validation rejections (empty key / non-string value / oversized value / reserved `:` delimiter).
- `plugin.service.spec.ts` (+2): uninstall purges both secrets and OAuth tokens, and still purges them when a later cleanup step fails (locks in the ordering guarantee).

## Notes / deliberately deferred (YAGNI)

- OS-keychain encryption-at-rest — future, behind the same API.
- A JSON-schema `writeOnly` marker to auto-route an issue-provider password field into secret storage — only if a provider plugin needs one.

## Scope

No sync surface touched, no new IPC, no native code. New code mirrors the established OAuth-store pattern.

Refs #7511, #5530


johannesjo:

src/app/plugins/plugin.service.ts
```diff
@@ -1556,6 +1558,21 @@ export class PluginService implements OnDestroy {
...
 
+    // Purge local-only credentials (secrets + OAuth tokens) FIRST so they
+    // never outlive their plugin — even if a later cleanup step throws.
+    // Best-effort: a purge failure is logged but must not abort the uninstall
+    // (on IndexedDB failure the credentials orphan locally until a later purge).
+    try {
+      await this._pluginSecretService.removeSecretsForPlugin(pluginId);
+    } catch (error) {
+      PluginLog.err(`Failed to purge secrets for plugin ${pluginId}:`, error);
+    }
```
dualbound, line 1569: The comment says secrets "orphan locally until a later purge", but I don't see a later purge: once `removeUploadedPlugin` returns, `pluginId` is gone from the registry, so `removeSecretsForPlugin(pluginId)` is never called again for it. On an IndexedDB failure here the credentials stay on disk indefinitely after the plugin is gone, more sensitive for a secret store than for cache. Could we either (a) add a startup reconcile that drops secrets whose owner `pluginId` isn't installed, or (b) reword the comment to be honest (orphan persists until the same plugin id is reinstalled)? Same gap applies to the `clearOAuthTokens` purge right below.

johannesjo: Thank you!

PR #8638, opened 2026-06-29
feat(tasks): reassign tags when dragging between tag groups

johannesjo:
## What & why

In the **grouped-by-tag** work view, dragging a task between tag groups previously did nothing useful to its tags (it just reordered the underlying list). This makes the drag reassign tags, which is the obvious expectation when tags are shown as groups.

It builds on the existing "drop a task onto a tag in the side-nav" mechanism, reusing the same `TaskService.updateTags` → single `updateTask` path (one user intent = one op).

## Behavior

| Drop target (in grouped-by-tag view) | Result |
| --- | --- |
| Another tag group | **Move** — drop the source group's tag, add the target's (other tags preserved) |
| The **"No tag"** bucket | **Clear all** of the task's tags |
| "Unknown tag" / ambiguous duplicate-title bucket | No-op (can't safely pick/clear) |
| Same group, or any non-grouped list (backlog, sections, subtasks, …) | Unchanged — normal reorder/convert |

## Commits

1. **`refactor(boards): extract doesTaskMatchPanel membership predicate`** — pulls the board-panel column-membership filter out of the component's inline `.filter()` into a pure, tested `doesTaskMatchPanel()` in `boards.util.ts` (companion to the existing `rewriteTagIdsForPanel`). Behavior-equivalent; the backlog predicate is hoisted so it's allocated once per recompute, not per task. *(Independent cleanup — does not block the feature.)*
2. **`feat(tasks): reassign tags when dragging between tag groups`** — the feature above.

## Design notes

- **No new sync surface.** The customizer emits an in-memory `groupTagIdByKey` (group title → tagId, a `NO_TAG_GROUP` sentinel for the No-tag bucket, `null` for un-retaggable buckets). It is recomputed, never persisted and never reaches the op-log; `updateTask` only ever carries real tagIds or `[]`.
- **`TODAY` stays virtual.** The retag target is derived from the sidebar tag list, which excludes `TODAY_TAG`, so `'TODAY'` can never be written into `task.tagIds` (sync rule #5).
- **Minimal blast radius.** All groups keep `listModelId="UNDONE"`; the retag is intercepted in `drop()` before `_move()`, so section / reserved-id / `enterPredicate` logic is untouched. The new `groupTagId` input is `undefined` for every other list.
- **No reverse dependency.** Deliberately did *not* import boards' `rewriteTagIdsForPanel` into `tasks` (would be the first `tasks → boards` edge); the single-tag rewrite is a trivial filter+append.

## Out of scope / v1 limitations (documented at the call site)

- Drop position within the target group isn't preserved (the task takes the group's natural order).
- A subtask dragged from a sub-task list into a tag group converts to a main task without acquiring the tag.

## Testing

Added/updated unit tests across all four touched specs (boards predicate AND-combinations, the group→tagId map incl. ambiguity + the No-tag/real-tag-collision guard, and the full drop() retag matrix: move, clear-on-No-tag, add-from-No-tag, dedupe-via-updateTags, and the fall-through cases). All green: boards.util 43 · board-panel 23 · customizer 52 · task-list 61 · work-view 15. Every changed file passes prettier + eslint.

## Review

Developed with multiple rounds of parallel multi-agent review (correctness/sync, architecture, simplicity, performance, alternatives, test-coverage). Findings addressed: required-not-optional backlog predicate, per-task closure hoist, unified the title→tagId derivation into one source (no drift), and — caught in the final pass — a guard so a user tag literally titled "No tag" adds that tag rather than hitting the clear-tags sentinel.


johannesjo:

PR #8640, opened 2026-06-29
fix(electron): use display name for Linux notifications

aquadevv:
## Summary

Fixes the application name shown for desktop notifications on Linux/KDE Plasma.

Notifications were shown as coming from `superProductivity`, while the rest of the desktop shell shows the app as `Super Productivity`.

## Changes

- Set the Electron app name to `Super Productivity` on Linux.
- Preserve the existing `userData` path before changing the runtime app name, so existing user data stays in the historical `superProductivity` directory.
- Leave package metadata and non-Linux behavior unchanged.

## Verification

- `npm run checkFile electron/start-app.ts`
- `npm run electron:build`
- Verified via DBus notification monitoring that the notification `app_name` is now `Super Productivity`.

DBus result:

```text
member=Notify
   string "Super Productivity"
```

### DIFF
--- a/electron/start-app.ts
+++ b/electron/start-app.ts
@@ -26,6 +26,7 @@ import { evaluateGpuStartupGuard } from './gpu-startup-guard';
 import * as fs from 'fs';
 
 const ICONS_FOLDER = __dirname + '/assets/icons/';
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much!

PR #8644, opened 2026-06-30
feat(task-repeat): default skip-overdue for new daily & weekday recurring tasks

johannesjo:
## What

Make new recurring tasks **stop piling up overdue duplicates by default — but only where that's safe**. The `skipOverdue` option ("Don't let overdue instances pile up") now defaults **on** for the **Daily** and **Every workday (Mon–Fri)** presets, and stays **off** for weekly/monthly/yearly schedules.

This supersedes the earlier "flip the default to `true` for everything" approach (see Why).

## Why

Daily duplication of stale recurring tasks is exactly the noise/guilt-clutter the manifesto rejects ("less noise, more depth"; "one calm default"). But a blanket default is the wrong tool: the benefit and the risk sit on opposite ends of the frequency spectrum.

- **Pile-up only happens for frequent tasks.** A monthly task has at most one missed occurrence — nothing piles up — so the option buys nothing there.
- **Silently dropping a missed occurrence only hurts infrequent tasks.** For weekly/monthly/yearly schedules, `skipOverdue` can make a real obligation ("pay rent on the 1st", "renew the domain") vanish until its next occurrence. For everyday/Mon–Fri schedules it can't meaningfully hurt: the next scheduled day is at most a weekend away, and a plain Daily task is always re-created for today (never drops to zero).

So the default is scoped to the near-daily schedules where pile-up is real and skipping is safe. This boundary was chosen via a UX review panel (safety, mainstream-mental-model, and simplicity lenses); they converged on **daily + the named Mon–Fri preset** — not an arbitrary "N+ weekdays" count, and explicitly *not* every-N-days / Mon-Wed-Fri-style schedules, where a missed occurrence is a multi-day vanish on potentially health-critical tasks.

## How

- New helper `getDefaultSkipOverdue(cfg)` — pure predicate: ON for the `DAILY` and `MONDAY_TO_FRIDAY` presets (and a CUSTOM cycle that is literally every single day); OFF otherwise. No weekday-count or interval threshold to maintain.
- Seeded in **both** config-creation paths so neither pile-up nor silent-drop slips through:
  - the **repeat dialog** — seeded on the new-config model and re-derived from the *final* schedule at save time, and **skipped when the user explicitly toggled the checkbox** (so per-config opt-in/out is preserved);
  - the inline **add-task-bar** recurrence.
- `DEFAULT_TASK_REPEAT_CFG.skipOverdue` stays `false` — the safe baseline for fixtures and any non-dialog spread; the real default is seeded at creation. Pinned by a test so it can't silently become a blanket default again.
- The option keeps its honest label **"Don't let overdue instances pile up"** (accurate for every schedule), rather than a "rolls into one" wording that is only true for the daily case.

## Scope & safety

- **Only affects newly created configs.** Existing configs keep their stored value; no migration, no op-log/replay change.
- **No data loss.** The existing cleanup effect only reaps *empty, untouched, overdue* instances; anything with tracked time, completion, subtask progress, or edits is preserved, and a single instance is never removed.

## Tests

- `get-default-skip-overdue.spec.ts` — the predicate across all presets + CUSTOM.
- Dialog spec — daily/Mon–Fri seed ON, monthly seed OFF, explicit user toggle respected.
- add-task-bar spec — inline daily seeds ON, inline monthly stays OFF.
- `task-repeat-cfg.model.spec.ts` — pins the `false` baseline.
- E2E `skip-overdue-default-8644.spec.ts` — daily persists `true`, monthly persists `false` through the real form.
- E2E `recurring-missed-day-bug-4559.spec.ts` still passes unchanged (monthly missed instance is still created).

## Docs

`docs/wiki/4.13` (reference) and `2.06` (how-to) document the option and the by-schedule default.

### DIFF
--- a/docs/wiki/2.06-Manage-Repeating-Tasks.md
+++ b/docs/wiki/2.06-Manage-Repeating-Tasks.md
@@ -30,7 +30,7 @@ You can also open the dialog from a repeating task instance or from a repeat pro
 5. Optionally set **Tags to add** (chip list).
 6. Optionally set **Scheduled start time** (e.g. 15:00; leave blank for all-day). If you set a time, **Remind at** appears — choose when to be reminded (e.g. when it starts, 5 minutes before).
 7. Optionally set **Default Estimate** (duration). **Order** (number) controls creation order when several recurring tasks are created at the same time; see the field description in the dialog.
-8. In the advanced section: **Default notes**, **Inherit subtasks** (copy subtasks from the latest instance to the next), and, if inherit is on, **Disable auto-updating subtasks** (do not update inherited subtasks when the newest instance changes).
+8. In the advanced section: **Default notes**, **Inherit subtasks** (copy subtasks from the latest instance to the next), if inherit is on, **Disable auto-updating subtasks** (do not update inherited subtasks when the newest instance changes), and **Don't let overdue instances pile up** (keep a single up-to-date instance instead of a stack of missed duplicates). The last option defaults **on** for the **Every day** and **Every Monday through Friday** presets and **off** for weekly/monthly/yearly tasks; see [[4.13-Repeating-Tasks]]. You can change it per task.
 9. Click **Save**. New instances are created according to the pattern when you open the project or the app.
 
 ## Add a time and Reminder to a Repeating Task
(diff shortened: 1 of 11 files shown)


johannesjo:
Superseded by #8861, which narrows the ON default to a plain everyday schedule only (dropping the Mon–Fri preset and the checkbox re-sync machinery). See #8861 for the rationale — the plain-Daily case is the only one where skipping is provably safe (today is always scheduled, so a missed instance can never silently vanish), and it avoids the "same schedule, different default" edge that the named-preset boundary created.

Issue #8643, opened 2026-06-30
🚨 Today background image flashes/persists on startup before/over the configured start page


johannesjo:
## Description

On launch the app briefly shows the **Today** page background image, then navigates to the user's configured start page (Settings → Misc → "Default start page"). Reported by a user:

> On the start-up screen, it always opens "Today" and I see its background photo, then it goes back to the default page that I've chosen.

## Steps to reproduce

1. Settings → Misc → set **Default start page** to something other than Today (e.g. a project, or Planner/Schedule/Boards).
2. Set a background image for the Today context (theme background).
3. Reload / relaunch the app.
4. Observe the Today background image painting first, before the configured page appears.

## Root cause (from code trace)

The routed start-page redirect itself is already handled correctly by an async guard — `DefaultStartPageGuard` returns a `UrlTree` to the configured page (`src/app/app.guard.ts`), so the **Today task-list route no longer activates**. What still flashes is the **background image / theme**, which is rendered at the app-shell level and is decoupled from routing:

- The background lives in the shell, not the routed component: `src/app/app.component.html` (`<div class="bg-image">`), bound to the active work context's theme via `GlobalThemeService.backgroundImg()` → `WorkContextService.currentTheme$` → `activeWorkContext$`.
- The **active work context defaults to the TODAY tag on every launch**: `src/app/features/work-context/store/work-context.reducer.ts` (`initialContextState = { activeId: TODAY_TAG.id, activeType: TAG }`). It is reconstructed from the URL, not persisted.
- The router uses **non-blocking initial navigation** (`src/main.ts`, `provideRouter(...)` without blocking initial nav), so the shell + Today background paint immediately.
- `DefaultStartPageGuard` only resolves **after** initial data has loaded (`isAllDataLoadedInitially$`) — i.e. async. By then the Today background is already on screen.
- The active context (and thus the background) is only re-synced to the URL on `NavigationEnd`, and **only for `tag/...` or `project/...` URLs** (`src/app/features/work-context/work-context.service.ts`).

### Consequences

- **Project start page:** Today background shows until the project page's `NavigationEnd` updates the context → brief flash. (Matches the user report.)
- **Planner / Schedule / Boards start page:** the `NavigationEnd` handler matches neither `tag/` nor `project/`, so the active context **stays TODAY** — the Today background does not just flash, it likely **persists** on those pages. _(Predicted from the code trace; worth confirming at runtime.)_

## Suggested fix directions

[...]

PR #8663, opened 2026-07-01
fix(theme): suppress Today bg image on non-context pages (#8643)

jibin7jose:
On every launch the work-context reducer initialises with activeId = TODAY_TAG.id, so GlobalThemeService.backgroundImg() immediately emits the Today tag's background image before the DefaultStartPageGuard has a chance to redirect the user to their configured start page.

This caused two symptoms reported in #8643:
  1. A visible flash of the Today background on startup when the configured start page is a project/Planner/Schedule/Boards.
  2. The Today background persisting permanently on non-work-context pages (Planner, Schedule, Boards) because the NavigationEnd handler in WorkContextService only updates the active context for tag/* and project/* URLs and intentionally leaves it unchanged for the standalone page routes.

Fix: extend _backgroundImgObs$ in GlobalThemeService to also track the current URL (via Router.events + NavigationEnd, seeded with the initial Router.url).  The map now returns null for any URL that does not match /(tag|project)/, which covers all standalone pages.  The @if block in app.component.html already guards on the resolved value, so returning null hides the background div without any template changes.

Result:
  - Startup flash eliminated: initial URL ('') does not match the regex, so no background is shown until the guard completes and NavigationEnd fires with the real destination URL.
  - Planner / Schedule / Boards: background correctly stays hidden because their URLs do not match /(tag|project)/.
  - Today / Project start pages: background continues to work as before because their URLs do match the regex.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


jibin7jose:
"@johannesjo I just pushed a new commit to fix the Prettier and ESLint formatting issues that were causing the CI to fail earlier. All checks are now passing and the PR should be good to go. Let me know if you need any further changes!"

johannesjo:
Thanks for this, @jibin7jose — and first off, apologies, this one's on us. 🙏

Your PR does exactly what issue #8643 asked for: the issue's own "Suggested fix directions" said to *clear* the Today background on Planner/Schedule/Boards, so removing it for non-context routes is a faithful reading. The gap is in the issue we wrote, not your implementation — it under-specified the intended end state. Sorry for the moving target.

Here's the behavior we actually want:

**Backgrounds should resolve as: per-context override → global wallpaper → none.**

- Introduce a **global wallpaper** — a new app-level setting stored in global config alongside the existing theme settings (synced, like the per-context ones).
- It shows **everywhere a more specific background isn't provided**: Planner, Schedule, Boards, Config, and any tag/project that hasn't set its own image.
- Tags/projects keep their per-context background image, which **overrides** the global one on their own pages.
- **Unset by default** → plain theme background out of the box, so we don't add visual noise for people who don't want a wallpaper.

Why this matters for #8643: the real bug isn't "wallpapers on Planner are wrong" — it's that the wallpaper there came from an *accidental* source (whichever work context happened to be active, Today by default). Removing it entirely fixes the leak but also drops the ability to have *any* background on those pages, with no way to set one back. A global wallpaper turns that accidental background into an intentional, user-controlled one — same leak fix, no lost capability.

The good news is that most of your PR is exactly the right foundation and should stay:

- ✅ The **startup-flash fix** is correct — with a global wallpaper set, the initial (`''`) route is non-context, so it shows the *global* image immediately (or nothing if unset), never a flash of the Today image.
- ✅ Your URL-aware `getBackgroundImageForUrl` + `_currentUrl$` stream is the right mechanism.

The change is small — instead of returning `null` for non-context URLs, return the **global** image, and for context URLs fall back to global when the context has no image of its own:

```ts
const ctx  = isDark ? theme.backgroundImageDark : theme.backgroundImageLight;
const glob = isDark ? globalBg.dark : globalBg.light;
return isWorkContextUrl(url)
  ? (ctx ?? glob)   // context page: its own image, else global
  : (glob ?? null); // non-context page: global only — never the active context's image
```

The one crucial correctness point (and the actual #8643 fix): on non-context pages we must use **only** the global image, never the active context's — because `currentTheme$` there still resolves to the sticky Today context. Your URL gate is what enforces that.

Remaining pieces to make it complete:
1. Add a global wallpaper field (light/dark) to global config + a settings entry that reuses the existing background-image picker.
2. Pass it into `getBackgroundImageForUrl` as shown above.
3. Extend the spec to cover: non-context URL + global set → global; context URL with own image → own image; context URL without image + global set → global.

Totally happy to help with any of these — and thanks again for a clean first PR and for bearing with us on the spec. 🙌

jibin7jose:
@johannesjo Thanks for the detailed explanation and guidance! Apologies for the moving target earlier, but your breakdown made it exactly clear what was needed.

I've just pushed a new commit that implements the global wallpaper fallback for non-context pages exactly as you outlined:

1.Added backgroundImageDark and backgroundImageLight fields to MiscConfig (unset by default) and hooked them up in the Misc Settings form.
2.Updated getBackgroundImageForUrl so that it falls back to the global image for context pages that lack their own, and exclusively uses the global image for non-context pages (fixing the 'Today' leak).
3.Expanded the tests in global-theme.service.spec.ts to cover all of these new permutations.

Everything is working great locally! However, I had two quick questions about some edge cases while implementing this:

1.Empty String Fallback: In the provided logic (ctx ?? glob), if a user clears their context-specific background image and the input saves it as an empty string (instead of null), the ?? operator won't fall back to the global background. Should we change it to (ctx || glob) or keep the ?? as is?
2.Opacity and Blur: Since we only added the global wallpaper image fields, the overlay opacity and blur applied to the global wallpaper will currently be inherited from whatever the active work context is (e.g., it inherits the 'Today' context's opacity when viewing the Planner page). Should we also add global settings for backgroundOverlayOpacity and backgroundImageBlur to MiscConfig, or is inheriting the active context's styling acceptable for now?

Let me know your thoughts and if you need me to adjust anything else! 🙌

johannesjo:
Thank you very much! <3 Can you have a look at the failing lint?

johannesjo:
Thanks so much for iterating on this @jibin7jose — and sorry for moving the target again. Both of your questions were spot-on, so we answered them by adjusting the design a bit:

**Q1 (empty string):** great catch — you're right that `??` wouldn't fall back. A cleared/empty image is treated as unset now, so it falls back to the global wallpaper.

**Q2 (opacity/blur):** we went with adding global overlay-opacity + blur so the wallpaper is fully stylable. To keep them from being inherited from the sticky "Today" context on Planner/Schedule/Boards, the opacity/blur now follow whichever image is actually shown.

We also moved the wallpaper controls out of Misc into a small **"Set wallpaper…" dialog under Theme** (next to the theme picker), since that's where people tend to look for it. Your core work — the URL-aware background stream and the resolver — is exactly the foundation this builds on and stays intact.

Since we kept piling scope on you, we took the dialog/UX part ourselves rather than asking you to build it. Thank you again for a really clean first PR and for bearing with the back-and-forth! 🙌

PR #8705, opened 2026-07-02
feat(update-check): notify desktop users about new releases (#5463)

### LINKED ISSUE #5463: 💡 Please add "check for updates" option
### Problem Statement

Hi.  Would you please add a "check for updates" option to the menu, so we can more easily learn of available updates to be downloaded and installed?  Thank you for your nice work!  

### Possible Solution

Click on "super productivity" and then "check for updates"; this will check if your version is the current release and, if not, offer to fetch the latest release file for you (or, at least tell you there's an update available and where to go get it). 

Otherwise, it's a manual check of the installed version, a web trip to open GitHub, check latest release, and download if available.  Then, if available, download & install it.  Many modern apps often have a "check for updates" menu option and/or some alert that there's a new version of the app available. 

### Additional context

_No response_

johannesjo:
Closes #5463

## What

Desktop builds currently have no update channel at all — users only learn about new releases by chance (see issue thread). This adds a calm "check for updates" capability, Electron-only:

- **Automatic check** (default ON, `Settings → Misc`): once a day the app fetches the GitHub `releases/latest` API and compares the tag against the running version. No auto-download or auto-install — a dismissible banner links to the release page instead.
- **Once per version**: dismissing the banner (or clicking Download) persists the tag in device-local `localStorage`; each release is announced at most once.
- **Manual check**: a "Check for updates" link in the settings footer, with "up to date" / error snacks.
- **Channel gating**: builds that self-update — Microsoft Store, Mac App Store, Snap — never check and show none of the UI (denylist in `is-update-check-possible.util.ts`; MAS also forbids out-of-store download links). Web (service-worker update flow) and mobile (stores) are excluded via `IS_ELECTRON`.

## Design notes

- **Version compare** (`is-newer-version.ts`) is strictly `remote > local` and prerelease-aware: master's `package.json` runs ahead of the latest published release, and RC tags are excluded from `releases/latest` — dev/RC builds are never nagged.
- **Untrusted input hardening**: the remote `tag_name` must match an anchored version-tag regex, and the release URL is built locally from it (`.../releases/tag/<tag>`) instead of trusting the response's `html_url` — the `openExternalUrl` IPC guard is scheme-only, so a forged response could otherwise point "Download" at an arbitrary host.
- **HTTP**: `HttpClient` (picks up the global network-retry interceptor for the post-resume case) with a 15s timeout and an in-flight guard. Automatic checks fail silently (offline is a normal state).
- **Config**: `misc.isCheckForUpdates` is a new *optional* key (forward-compatible for synced configs); a missing key means ON. The checkbox uses the established display-only `defaultValue: true` seed pattern.
- **Privacy**: a single unauthenticated GET with no identifiers or user data; disclosed in the website privacy policy (updated separately) and in the setting's hint text.
- **Release process**: unchanged. The existing draft-then-publish flow means `releases/latest` only ever surfaces complete stable releases.
- All 7 new strings translated into all 27 locales.

## Test plan

- 25 unit tests (version compare edge cases; banner shown/suppressed/dismiss persistence; strict-tag rejection; request dedup; silent vs user-triggered error paths) — green.
- Existing global-config reducer / config-page / startup specs — green.
- Reviewed via multi-agent review (7 reviewers) + a second two-agent pass; all verified findings fixed in `dff0acde84`.
- [ ] Real-Electron smoke test (banner + footer link) before release.


johannesjo:

Issue #8512, opened 2026-06-20
Re-open nodeExecution for community plugins behind explicit consent


johannesjo:
## Background

#8205 (v18.10.0) moved `nodeExecution` consent into the Electron **main** process and, as part of that, blocked uploaded/community plugins from using `nodeExecution` entirely — it's currently built-in-only. That broke real community plugins (e.g. [Super Productivity MCP](https://github.com/b0x42/Super-Productivity-MCP), discussion #8385; demand also in #7256).

Rather than bundle individual plugins (which would pull third-party native-code plugins into SP's signed release and make them look first-party — see the reasoning in #8385), the goal is to **re-open `nodeExecution` for community plugins generally, behind an explicit consent gate**, keeping the trust decision with the user and the plugin clearly flagged as unverified third-party.

## Current state (verified against current code)

The main-process consent model from #8205 is the right foundation and does **not** need redesigning:

- Native consent dialog runs in **main** — `electron/plugin-node-executor.ts:73`.
- Main mints a random 32-byte grant token after the dialog; the renderer cannot fabricate one — `:108`. (The old self-grantable `PLUGIN_SET_NODE_CONSENT` IPC is gone.)
- Token bound to `(pluginId, webContentsId)`, re-validated on every exec — `:140-147`.
- Grant auto-revoked on navigation/destroy and on plugin teardown — `:154-191`, `plugin.service.ts:_revokeNodeExecutionGrant`.
- Grants are **session-only** (in-memory `Map`); nothing is persisted today.

Two deliberate gates restrict this to built-in plugins:

1. **Client gate** — `src/app/plugins/plugin.service.ts:_assertUploadedPluginAllowed` throws `NODE_EXECUTION_BUILT_IN_ONLY` for any uploaded plugin declaring `nodeExecution`. (This is the install-time error users hit.) It also has a separate bundled-id-collision branch that must be kept.
2. **Main gate** — `electron/plugin-node-executor.ts:getVerifiedBuiltInNodeExecutionManifest` resolves manifests **only** from `bundled-plugins/` dirs and throws otherwise. Even with gate 1 removed, grants still fail for uploaded plugins.

### Key architectural fact: main has no trusted view of an uploaded plugin

[...]

PR #8708, opened 2026-07-02
fix(plugins): spawn nodeExecution scripts via process.execPath so packaged apps work

### LINKED ISSUE #8707: nodeExecution plugin scripts fail with "spawn node ENOENT" in packaged builds
The spawn path in electron/plugin-node-executor.ts picks its binary like this:

    const nodePath = process.execPath.includes('electron') ? process.execPath : 'node';

In a packaged app that check never matches: the executable is "Super Productivity" on macOS, "superproductivity" on Linux, "Super Productivity.exe" on Windows. So the executor falls back to spawning the literal "node". The child env is deliberately stripped to NODE_ENV and ELECTRON_RUN_AS_NODE, with no PATH, so the lookup can only hit the libc fallback (/usr/bin:/bin on macOS). Unless node happens to live there, every nodeExecution script that reaches the spawn path rejects with "Failed to execute script: spawn node ENOENT". In dev the executable path contains "electron", which is why this never shows up when running from source.

Repro:

1. Package a build (I hit this on a macOS DMG built from master).
2. Enable a plugin with the nodeExecution permission and run any script that routes to the spawn path, i.e. anything mentioning child_process.
3. Grant the consent dialog. The call rejects with spawn node ENOENT.

This makes the consent-gated nodeExecution work from #8512 / #8576 unusable in packaged builds, which is where end users actually are.

The fix looks like a one-liner: always spawn process.execPath. ELECTRON_RUN_AS_NODE=1 is already set on the child env, no fuse disables it in this repo, and in dev (and in the node-run unit tests) execPath already points at the right binary. PR incoming.

oldrobotdev:
Fixes #8707.

Problem

The spawn path resolved its binary with

    const nodePath = process.execPath.includes('electron') ? process.execPath : 'node';

Only dev runs match that substring. In packaged builds the executable is "Super Productivity" (macOS), "superproductivity" (Linux) or "Super Productivity.exe" (Windows), so the executor spawned the literal "node" with a child env that has no PATH (it is stripped to NODE_ENV and ELECTRON_RUN_AS_NODE on purpose). On macOS and Linux that only resolves through the execvp fallback (/usr/bin:/bin), and on Windows it fails outright, so nodeExecution scripts die with "spawn node ENOENT" on any machine where node is not in /usr/bin. Confirmed on a packaged macOS build; it means the consent-gated nodeExecution flow from #8512 / #8576 never worked in packaged apps.

Fix

Always spawn process.execPath. ELECTRON_RUN_AS_NODE=1 is already set on the child env, no fuse in this repo disables RunAsNode, dev behavior is unchanged (execPath contains "electron" there), and the unit tests run under plain node where execPath is node itself.

Test

Added a regression test that stubs child_process.spawn (scoped to the executor module, same pattern as the existing fs and consent-store stubs) and asserts the executor passes process.execPath, keeps ELECTRON_RUN_AS_NODE=1 and leaves PATH out of the child env. It fails against the old resolution and passes with the fix. Full suite: 21 pass.

### DIFF
--- a/electron/plugin-node-executor.test.cjs
+++ b/electron/plugin-node-executor.test.cjs
@@ -30,6 +30,8 @@ let nextDialogPromise;
 // In-memory stand-in for the main-owned persisted consent store (plugin-node-consent-store.ts),
 // so the executor's Phase 2 ask-once logic can be tested without touching the filesystem.
 let consentStore;
+// Recorded (command, args, options) tuples from the stubbed child_process.spawn.
+let spawnCalls;
 
 class FakeWebContents extends EventEmitter {
   constructor(id, url = 'app://index.html') {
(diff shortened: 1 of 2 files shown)


johannesjo:

PR #8712, opened 2026-07-02
fix(issuePanel): allow removing an issue provider whose plugin is uninstalled

### LINKED ISSUE #8711: Issue provider from an uninstalled plugin cannot be removed (dead tab in the issue panel)
When an issue provider was added by a plugin and that plugin is later uninstalled (or re-uploaded under a new id), its provider instance stays behind as a dead tab in the issue panel. The tab content only says 'Required plugin "x" is not installed.' and there is no way to get rid of it.

The usual paths don't work for this case:

1. The tab content for a missing plugin renders only the error text, so the wrench/settings route of a working provider isn't there.
2. The long-press shortcut on the tab header ((longPress) opens the edit dialog) never fires with a mouse: the tab's cdkDrag has a mouse drag start delay of 0, so the drag claims the press before the 600 ms long-press timer completes. Touch devices are fine, desktop is not.

Repro:

1. Upload any issueProvider plugin, add it as a provider in the issue panel.
2. Remove the plugin in Settings, then Plugins.
3. The provider tab remains with the not-installed message and cannot be removed. It also survives restarts and sync.

Suggested fix: show a Delete button directly on the not-installed tab content, reusing the exact removal flow of the edit dialog (confirm dialog, unlink referencing tasks, TaskSharedActions.deleteIssueProvider). PR incoming.

oldrobotdev:
Fixes #8711.

Problem

A provider instance added by a plugin outlives the plugin: after the plugin is uninstalled the issue panel keeps a dead tab that only says the plugin is not installed. There is no removal path. The working-provider settings route needs the tab content that doesn't render here, and the long-press edit shortcut on the tab header is unreachable with a mouse because the tab's cdkDrag (mouse drag start delay 0) claims the press before the 600 ms long-press timer fires.

Fix

The not-installed tab content gets a Delete button that reuses the exact removal flow of the edit dialog: confirm dialog, gather tasks still referencing the provider, dispatch TaskSharedActions.deleteIssueProvider with the ids to unlink.

Testing

Manually on a packaged build: uploaded a plugin, added its provider, removed the plugin, deleted the orphaned provider through the new button; the tab disappears, restart and sync keep it gone, and tasks that referenced the provider are unlinked. Working providers are unaffected (button only renders in the missing-plugin branch).

### DIFF
--- a/src/app/features/issue-panel/issue-panel.component.html
+++ b/src/app/features/issue-panel/issue-panel.component.html
@@ -54,6 +54,16 @@
               T.F.ISSUE.S.PLUGIN_NOT_INSTALLED
                 | translate: { pluginName: entry.missingPluginName }
             }}
+            <div style="margin-top: 16px">
+              <button
+                mat-stroked-button
+                color="warn"
+                (click)="removeMissingProvider(entry.issueProvider)"
+              >
+                <mat-icon>delete</mat-icon>
+                {{ T.G.DELETE | translate }}
+              </button>
+            </div>
           </div>
         } @else {
           <issue-provider-tab [issueProvider]="entry.issueProvider"></issue-provider-tab>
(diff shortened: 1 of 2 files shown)


oldrobotdev:
Tested this end to end on the preview deployment with a scripted browser run:

1. Uploaded a minimal issueProvider plugin, added it as a provider, then removed the plugin.
2. After a reload the orphaned tab shows the not-installed message together with the new Delete button.
3. Delete, confirm: the provider is gone and stays gone across further reloads.

johannesjo:
Thank you very much! <3

PR #8714, opened 2026-07-02
fix(tasks): treat a missing title as blank instead of crashing isBlankTask

### LINKED ISSUE #8713: Deleting a task with a missing title crashes the delete-undo effect (isBlankTask TypeError)
isBlankTask does

    !task.title.trim() &&

with no guard on the title. Tasks without a title exist in the wild (we have a few in a long-lived database, most likely old orphaned subtasks from past sync/repair edge cases), and deleting one takes down the whole snackDelete$ effect stream:

    TypeError: Cannot read properties of undefined (reading 'trim')
      at trim (src/app/features/tasks/util/is-blank-task.ts:16:16)
      at actionFn (src/app/features/tasks/store/task-ui.effects.ts:123:30)
      at filter (rxjs)

Once that filter throws, the effect is dead for the rest of the session, so later deletes stop showing the undo snack too.

A task with no title at all is at least as blank as a task with an empty title, so the fix is a one-character guard: task.title?.trim(). PR with a regression spec incoming.

oldrobotdev:
Fixes #8713.

isBlankTask read task.title.trim() unguarded. A task without a title (they exist in long-lived databases, e.g. old orphaned subtasks) made the call throw, and because it runs inside the snackDelete$ effect's filter, the whole effect stream died for the rest of the session.

A missing title is at least as blank as an empty one, so this guards with task.title?.trim() and adds a regression spec (missing title returns true and does not throw). The spec file's other 15 cases still pass.

### DIFF
--- a/src/app/features/tasks/util/is-blank-task.spec.ts
+++ b/src/app/features/tasks/util/is-blank-task.spec.ts
@@ -32,6 +32,12 @@ describe('isBlankTask', () => {
     expect(isBlankTask(createTaskWithSubTasks())).toBe(true);
   });
 
+  it('should be true (and not throw) when the title is missing entirely', () => {
+    const t = createTaskWithSubTasks() as any;
+    delete t.title;
+    expect(isBlankTask(t)).toBe(true);
+  });
+
   it('should be true when the title is only whitespace', () => {
     expect(isBlankTask(createTaskWithSubTasks({ title: '   ' }))).toBe(true);
   });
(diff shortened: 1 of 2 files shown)


oldrobotdev:
Some testing notes beyond the included spec:

The fix has been running all afternoon in our team build on the machine where the original stack trace came from. That database genuinely contains tasks without a title (old orphaned subtasks), which is what made the crash reproducible for us in the first place. With the guard in place, deletes behave normally and the undo snack keeps working for the rest of the session; before it, one delete of such a task silently killed the snackDelete$ stream until restart.

The regression spec covers the exact case: a task object with the title property removed returns true from isBlankTask instead of throwing. On the code without this patch that spec fails immediately (undefined.trim()).

The preview deployment loads and behaves normally as a smoke check; I could not build a title-less task through the UI there, which is rather the point - they only arise from old data, sync edge cases and the like, and that is exactly when the unguarded trim bites.

johannesjo:

Issue #8716, opened 2026-07-02
💡 Declare NSRemindersUsageDescription on macOS so plugins can integrate with Apple Reminders (Apple Watch / Siri / iCloud ecosystem)


Rahulsharma0810:
## Problem Statement

On macOS, Super Productivity currently cannot participate in the native Apple task ecosystem, because the app bundle does not declare the `NSRemindersUsageDescription` key in its `Info.plist`.

macOS enforces a hard rule: an app that touches the EventKit **Reminders** store without that usage-description string is **silently denied** by TCC — no permission prompt is ever shown, and the app never appears under **System Settings → Privacy & Security → Reminders**. There is no way for a user to grant access manually, because macOS refuses to even list the app.

I ran into this while building a community plugin ([Apple Reminders Sync](https://github.com/Rahulsharma0810/superproductivity-apple-reminders-sync), a two-way SP ⇄ Apple Reminders sync via the `remi` CLI). The plugin itself works end-to-end — but every Reminders call made from within Super Productivity comes back denied, with no prompt and no toggle to fix it. I confirmed the cause directly:

- `/Applications/Super Productivity.app/Contents/Info.plist` has **no** `NSRemindersUsageDescription` key (`PlistBuddy` → "Does Not Exist").
- `com.super-productivity.app` is **absent** from the Reminders TCC database, while other apps that declare the key (Terminal, iTerm, etc.) are present and grantable.
- The exact same `remi` command **succeeds** when its responsible app is one that declares the key, and **fails** (silently) when the responsible app is Super Productivity.

This is bigger than one plugin. Apple Reminders is the hub of the native Apple task ecosystem: tasks captured on **Apple Watch**, via **Siri** ("remind me to…"), from **HomePod**, the iOS/iPadOS share sheet, and **shared iCloud lists** all land in Reminders and sync across every Apple device automatically. Reminders genuinely works well *within* that ecosystem. Without this one key, none of that reach is available to Super Productivity or its plugin ecosystem on macOS — cross-device / cross-app capture and sync are simply off the table.

## Possible Solution

Add `NSRemindersUsageDescription` to the macOS build's `Info.plist` via the existing `extendInfo` block in `electron-builder.yaml`. This is a one-line, low-risk change and requires no code changes.

`electron-builder.yaml` already has the right place for it:

```yaml
mac:
  # ...
  extendInfo:
    ITSAppUsesNonExemptEncryption: false
    NSRemindersUsageDescription: >-
[...]

PR #8717, opened 2026-07-02
feat(macos): declare NSRemindersUsageDescription so plugins can integrate with Apple Reminders

Rahulsharma0810:
## What

Adds a single `NSRemindersUsageDescription` string under `mac.extendInfo` in `electron-builder.yaml` so the notarized macOS build can *optionally* access Apple Reminders.

## Why

macOS silently denies EventKit **Reminders** access to any app whose `Info.plist` lacks `NSRemindersUsageDescription` — **no permission prompt is shown**, and the app never appears under **System Settings → Privacy & Security → Reminders**, so users cannot grant it manually.

Because Super Productivity is signed with the hardened runtime and notarized, this string can only be added correctly in an upstream build (editing the shipped `Info.plist` locally breaks the signature). That makes this a small but necessary **ecosystem enabler**: it lets community plugins/integrations sync tasks with Apple Reminders, which is the hub for **Apple Watch, Siri, HomePod, and shared iCloud lists**.

Full rationale, TCC evidence, and privacy analysis are in **#8716**.

## Privacy — this is inert by default

Declaring the string does **not** request or grant anything on its own. It only enables three user-controlled layers:

1. **Nothing happens unless the user enables** a plugin/integration that uses Reminders.
2. When one first accesses Reminders, macOS shows the standard **Allow / Don't Allow** prompt — the user can deny.
3. The user can revoke anytime in **System Settings → Privacy & Security → Reminders**.

## Change

```yaml
mac:
  extendInfo:
    ITSAppUsesNonExemptEncryption: false
    NSRemindersUsageDescription: Super Productivity only accesses Apple Reminders if you enable a plugin or integration that syncs your tasks with the Reminders app.
```

- Correct location: `mac.extendInfo` (not the entitlements plist, which is for hardened-runtime code-signing flags).
- No entitlement change, no capability added by default, no new dependency.
- Windows/Linux builds are unaffected.

## Motivation / context

I built a standalone Super Productivity plugin that bidirectionally syncs tasks with Apple Reminders via the `remi` CLI: https://github.com/Rahulsharma0810/superproductivity-apple-reminders-sync — it works end to end except that the shipped macOS build is TCC-denied for Reminders for the reason above. This one line unblocks that whole class of integrations.

Refs #8716

### DIFF
--- a/electron-builder.yaml
+++ b/electron-builder.yaml
@@ -193,6 +193,12 @@ mac:
   notarize: true
   extendInfo:
     ITSAppUsesNonExemptEncryption: false
+    # Optional Apple Reminders access for plugins/integrations (e.g. syncing tasks
+    # with Apple Reminders, which brings Apple Watch / Siri / iCloud shared lists).
+    # Declaring this string does NOT request or grant access on its own; macOS only
+    # prompts if a user-enabled plugin actually accesses Reminders, and users can
+    # deny or later revoke it in System Settings > Privacy & Security > Reminders.
+    NSRemindersUsageDescription: Super Productivity only accesses Apple Reminders if you enable a plugin or integration that syncs your tasks with the Reminders app.
   publish:
     - github
   target:


johannesjo:
Thanks for this — and genuinely for the homework behind it. The TCC investigation in #8716 (PlistBuddy check, the responsible-process behavior, confirming the DB absence) is more rigorous than most capability requests get, and you put the key in the right place (`extendInfo`, not the entitlements plist). None of the pushback below is about the change being wrong.

I'm going to hold off on merging this for now, and I want to be transparent about the reasoning rather than just closing it:

**The core app has no Reminders feature.** This would declare an OS-permission string in the official signed build that Super Productivity itself never uses — purely to serve a plugin. I want the app's `Info.plist` to stay an honest manifest of what the app actually does.

**The TCC grant is app-wide, not per-plugin.** Because the access is attributed to the responsible process (Super Productivity), once a user approves the prompt, that Reminders grant applies to *anything* SP spawns via `nodeExecution` — not just your plugin, and not revocably per-plugin. Given the security history around `nodeExecution`, I'm cautious about widening what that surface can reach into (personal data that syncs to iCloud/Watch/all devices), even though it's gated behind the standard prompt.

**It's a narrow, single-integration use case today.** It's macOS-desktop-only, only works while the Mac app is running, and the demand so far is your (excellent) plugin. My default on new capability surface is to wait until real workflows are asking for it before baking it into the shipped build.

**The path that unblocks you today, without waiting on us:** ship the Reminders side as its own separately code-signed + notarized helper bundle that carries its own `NSRemindersUsageDescription`, and have the plugin launch it via LaunchServices (`open`) so it becomes its *own* responsible process with its *own* TCC identity. That's actually better for users than borrowing SP's identity — the grant shows up under the helper's name in System Settings and is revocable on its own line, scoped to exactly the thing they installed. It does need an Apple Developer ID to notarize, and I'd verify the responsible-process attribution with the same testing you already did — but it's the architecturally clean version and it doesn't depend on an upstream release cycle.

I'll keep #8716 open as the place to gather demand. If a real cluster of macOS users show up wanting Siri/Watch/HomePod capture flowing into SP's inbox, that's a first-party feature worth building — and at that point declaring the key is justified because *we'd* be the ones using it. A 👍 or "+1, I'd use this" on the issue from folks is exactly the signal I'm watching for.

Really appreciate the work — this isn't a no-forever, it's a not-until-there's-a-feature-behind-it.

PR #8721, opened 2026-07-03
feat(plugin): add PluginAPI.request with manifest allowedHosts allowlist

oonid:
## Problem

Plugins can already make HTTP calls from issue-provider code via the injected PluginHttp service. But hook handlers (currentTaskChange, taskComplete,  …) and other non-issue-provider plugin code have no host-mediated way to make an HTTP request, fetch/XHR from the plugin sandbox hit CORS, and there  is no equivalent of PluginHttp outside the issue-provider surface.

This blocks a whole class of plugins that need to talk to an API in response to app events. Concretely, it blocks the Basecamp issue-provider plugin I've been building (context: #8465): it needs to push tracked time to the Basecamp timesheet from its stop/done hooks, which is impossible today because the hook has no HTTP path, it fails with PluginAPI.request is not a function.

## Solution

Add a generic, guarded HTTP capability for plugins, plus a manifest-declared allowlist so a plugin's outbound reach is explicit and reviewable.

1. PluginAPI.request<T>(url, options?) : a host-executed HTTP method available to all plugins (non-iframe and iframe). It routes through the existing guarded HTTP layer (PluginBridgeService.request → PluginHttpService), so it inherits every protection already applied to issue-provider HTTP:  http/https only, an allowed-method list, and the SSRF guards (loopback / private-IPv4+IPv6 / link-local blocked; cloud-metadata endpoints always blocked). Plugins supply their own Authorization headers; the host only executes the request.

2. Manifest allowedHosts + host-enforced allowlist : a plugin declares the exact hostnames it may reach:

  "allowedHosts": ["3.basecampapi.com", "launchpad.37signals.com"]

The host enforces it (never plugin code), in PluginBridgeService, and it runs before the shared SSRF layer so a blocked host never reaches the network:
  - exact host match, no wildcards
  - case-insensitive, trailing-dot tolerant, port-agnostic
  - resolves userinfo tricks via new URL() (https://ok.com@evil.com → evil.com)
  - fail-closed: an omitted/empty allowedHosts disables request entirely

This makes a plugin's outbound reach declared and reviewable at install instead of "any host the SSRF filter doesn't happen to block".  validatePluginManifest rejects malformed entries (non-array, scheme/port/path, empty) with a clear message.

Design choices: exact host (predictable, no wildcard ambiguity); fail-closed (a plugin that forgets to declare hosts gets no network, not all); declared in the plugin but enforced by the host (developer controls the list, host controls that it's honored); reuses the one existing SSRF implementation rather than adding a second.

Known limitations (documented; shared with existing issue-provider HTTP): the web/Angular HttpClient path auto-follows 3xx and only the initial URL is checked (redirect following); the hostname string is matched, not the resolved IP (DNS rebinding); exfiltration to a declared host is inherent to any allowlist. the point is the destination is now user-reviewable. No per-plugin rate/size limiting (out of scope).

Tests: plugin-api.spec.ts (request delegates to bound method); plugin-bridge.service.spec.ts (routes generic HTTP; allowlist: rejects non-declared host, fail-closed on undefined + empty, case-insensitive + trailing-dot, userinfo-trick blocked); plugin-iframe.util.spec.ts (request dispatchable across the iframe bridge); validate-manifest.util.spec.ts (accepts valid + omitted; rejects non-array, scheme/port/path, empty/non-string).

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Reviewed the diff + the surrounding plugin HTTP/manifest code. The design is careful — reusing the single existing `PluginHttpService` SSRF layer (rather than a second one), fail-closed `allowedHosts`, host-side (never plugin-side) enforcement, and solid normalization of the real tricks (userinfo `ok.com@evil.com` → `evil.com`, case, trailing dot, port-agnostic). Wiring reaches both iframe and Function-sandbox plugins identically, and `createHttpHelper(() => ({}))` keeps `allowPrivateNetwork:false` so private-network blocks stay active. Nice.

Two things block merge, plus a design question worth settling before this API contract ships.

## Blockers

**1. CI Lint is failing (prettier) — 3 errors in the changed files**, despite the "ran `npm run checkFile`" checkbox:
- `src/app/plugins/plugin-api.ts:627` — collapse the multi-line `request(...)` signature
- `src/app/plugins/plugin-api.spec.ts:353` — collapse the multi-line call args
- `src/app/plugins/util/plugin-iframe.util.spec.ts:430` — wrap the long `request:` assertion string

`npx prettier --write` on those three fixes it.

**2. `allowedHosts` is never surfaced in the UI, so the "reviewable at install" property isn't actually delivered yet.** The description says the allowlist is *"Surfaced to the user at install so the plugin's outbound network reach is reviewable"*, and that's the whole value-add over the pre-existing SSRF filter. But `plugin-management.component.html` renders only `permissions` and `hooks` chips — `allowedHosts` appears nowhere in the app, so a user can only see it by opening the raw `manifest.json`. Rendering it alongside permissions/hooks (a `<mat-chip-set>` mirroring the permissions block) would close the gap.

## Design question

**Should outbound network be an explicit, consented capability rather than implicitly granted by the presence of `allowedHosts`?** Right now `request` is added to every plugin's bound methods unconditionally — there's no `permissions` entry and no consent moment; declaring a host *is* the grant. Contrast `nodeExecution`, which requires a grant token / consent gate. Network egress is a comparable exfiltration surface, so reusing that same pattern (a `permissions: ['http']` entry + consent + the UI surfacing above) would be more consistent and would actually make the reach reviewable. It's much easier to tighten this now than after plugins depend on the current shape.

## On the documented redirect limitation

You already flag it, so just emphasizing why it matters more here than for issue-provider HTTP: Angular's `HttpClient` (XHR/fetch) auto-follows 3xx and only the initial URL is host- and SSRF-checked. For issue providers the destination is a URL the user explicitly configured; for a generic allowlist, a declared-but-malicious host can `302` to `169.254.169.254`/a private IP and the follow-up is neither allowlist- nor SSRF-checked — a real SSRF escalation that returns internal/metadata content to the plugin. Fine to ship as a known limitation, but worth a line in the `allowedHosts` JSDoc + a tracking issue (e.g. `redirect:'manual'` + per-hop re-validation where the platform allows).

## Minor
- `_assertRequestHostAllowed` strips a trailing dot but not IPv6 brackets, unlike `PluginHttpService._validateUrl` (`replace(/^\[|\]$/g,'')`). Harmless today because `validate-manifest` rejects entries containing `:` (so IPv6 literals can never match → always fail-closed), but the two hostname-normalization paths diverging is a latent footgun — a one-line comment noting the intentional block would help.
- Docs checkbox unchecked: new plugin-dev API + manifest field with no wiki entry.

Test coverage of the allowlist edge cases is genuinely good, and `PluginRequestOptions extends PluginHttpOptions` adds `method`/`body` cleanly with no field clash. Once the lint + UI surfacing are addressed (and ideally the consent model decided), this is close.

oonid:
Thanks @johannesjo,
genuinely useful review. Addressed all of it; pushed four focused commits on top.

## Blockers

**1. Prettier lint** ✅ `5ae1292` : ran `prettier --write` on the flagged files; `npm run checkFile` is now clean on every changed file. Honest cause: those files came in with the generic-request commit and I'd only run `checkFile` on the allowlist files, so they slipped past the checkbox. (Small note: `plugin-api.spec.ts` is only 314 lines so `:353` didn't resolve for me, the two that actually failed, `plugin-api.ts` and `plugin-iframe.util.spec.ts`, are fixed.)

**2. UI surfacing** ✅ `af5b7a2` : `allowedHosts` now renders as a `<mat-chip-set>` next to permissions/hooks (with a count in the collapsible title), and `getPermissionDescriptions` adds a human-readable "🟡 Can make network requests to: …" line. The "reviewable at install" property is actually delivered in-app now, not just in the raw manifest.

## Design question — consent model ✅ `d38b3ea`

Agreed, and changed it. Outbound network is now an explicit, declared capability: `PluginAPI.request` requires **both** `"http"` in `permissions` **and** a matching `allowedHosts` entry, missing either is fail-closed, enforced host-side on both the iframe and Function-sandbox paths. Declaring a host no longer implies network; it's opt-in like `nodeExecution`, and surfaced in the UI per above.

I stopped short of an active Allow/Deny **modal** and wanted your call on whether it's needed for this PR: the existing `nodeExecution` consent dialog is Electron-main-process only (gated on `IS_ELECTRON`), whereas `http` applies on web too, so a modal is net-new cross-platform work rather than a reuse. The hard-to-reverse part is the capability contract (`permissions: ['http']`), and that's in now; a modal can layer on top later without changing it. Happy to add it here if you'd prefer.

## Redirect limitation — fixed, not just documented ✅ `75ae301`

Your framing convinced me to close this now rather than defer. The `request` path executes via `fetch(url, { redirect: 'error' })`, so a declared host that 302s to `169.254.169.254`/a private IP is **refused** instead of followed, no internal/metadata content reaches the plugin. The fetch executor preserves the `HttpClient` contract callers depend on: query params, `AbortController` timeout (covering the body read, not just headers), text/json parsing, and non-2xx rejecting with `.status` + `.error`. Only plain objects are JSON-serialized (FormData/Blob/URLSearchParams/ArrayBuffer pass through). Issue-provider HTTP is untouched (still `HttpClient`).

Two caveats, both documented in the `allowedHosts` JSDoc:
- **Native (Capacitor):** `CapacitorHttp` patches `fetch` and ignores `redirect`/`signal`, so on native the path falls back to `HttpClient` and retains redirect-following, web/desktop are protected.
- Blocks **all** redirects incl. benign same-host ones; manual per-hop re-validation isn't possible on the web (cross-origin `Location` is opaque to `fetch`).

The remaining hardening : native-redirect blocking and DNS-rebinding protection, both need a native/Electron HTTP executor that doesn't exist yet (`followRedirects(false)` / resolved-IP pinning), so they're out of scope here and documented as known limitations in the `allowedHosts` JSDoc.

## Minor
- **IPv6 brackets:** added a comment noting the intentional divergence from `_validateUrl`, `validatePluginManifest` rejects entries containing `:`, so an IPv6 literal can never be declared and always fails closed. (`d38b3ea`)

New tests alongside: fail-closed without `http`; `allowedHosts` surfaced in the management UI; redirect refused, `.status`/`.error` parity, SSRF still pre-checked, body pass-through, `AbortController` timeout, and native falls back to `HttpClient`.

johannesjo:
Reviewed the incremental range (`184818f..75ae301`): redirect-blocking + the required `"http"` capability + the UI surfacing. This is solid, fail-closed, and well-tested work — thanks. A few notes, none blocking.

**What holds up well**
- The host-side ordering is right: capability → `allowedHosts` → the shared `_validateUrl` SSRF layer → the redirect-blocked `fetch`. I checked that `_validateUrl` runs *before* the fetch, so a metadata/private IP is refused without a socket ever opening (matches your "blocks metadata IP, no fetch" test).
- `redirect: 'error'` is the correct mechanism, and `_requestNoRedirect` carefully reproduces the `HttpClient` contract callers depend on — param merging, timeout via `AbortController` (covering the body read), text/json parsing, JSON-only serialization (leaving FormData/Blob/URLSearchParams intact), and non-2xx → error with `.status`/`.error`.
- The IPv6-bracket reasoning is internally consistent: `validate-manifest` rejects any `allowedHosts` entry containing `:`, so a bracketed request host can never match and stays fail-closed.

**Minor notes**

1. **The `fetch` path bypasses `NetworkRetryInterceptorService`.** On web/desktop, `_requestNoRedirect` uses `window.fetch`, which skips the global `HTTP_INTERCEPTORS`. That interceptor retries GETs once on a status-0 error to survive WebView socket warm-up "after Android/Electron resume." So on Electron, `PluginAPI.request` GETs lose that single retry, while the **native** fallback and the issue-provider path (both `HttpClient`) keep it. It's inherent to the redirect requirement (XHR can't do `redirect:'error'`) — just flagging the divergence; fine to leave as-is or note it in the code comment.

2. **The "network reach" the user actually sees isn't tied to the `http` capability.** The `http` branch was added to `PluginSecurityService.getPermissionDescriptions()` ("🟡 Can make network requests to: …"), but that method has **no production caller** (only a spec mock), so it's never rendered. The surface users do see — the plugin-management collapsible — renders the `allowedHosts` chips and the title count whenever `allowedHosts.length > 0`, independent of `permissions`. Net effect: a manifest with `allowedHosts` but no `"http"` shows a "Network access" section for a capability the bridge will reject. Consider gating the panel's `allowedHosts` section (and its title count) on the `http` permission — or wiring `getPermissionDescriptions` into the UI so the well-worded, correctly-gated description is what shows.

3. **Native redirect-following (documented — just acknowledging).** `blockRedirects` is a no-op on Capacitor, so on native an allowed host's 3xx is followed and not re-checked (→ off-allowlist or a LAN device). It's called out in three places and the trade-off is reasonable given the platform constraint; worth being explicit that the headline "block redirects" protection is web/desktop-only in practice.

4. **All redirects are blocked on web/desktop, including benign same-host ones** (http→https, trailing-slash 301s). Documented clearly; plugin authors must target final URLs.

**Small bonus worth mentioning in the PR description:** `redirect:'error'` doesn't only close SSRF — it also stops the plugin's request (its auth headers / body) from being replayed onward to a redirect target the user never approved. A little credential-safety win on top of the SSRF fix.

I didn't run the suites in my environment, but the new specs read correct and comprehensive. Nothing blocking from my side. 👍

oonid:
Thanks @johannesjo  for the careful second pass, really appreciate it. 
Acted on the two code-relevant notes and folded the rest into the description.

**Network-reach UI not tied to the http capability** : fixed ([2b71a33](https://github.com/super-productivity/super-productivity/pull/8721/commits/2b71a338a7af2896c8293ccc267e8a09e00f577d)). The plugin-management panel now gates the allowedHosts section and its title count on the http permission, via a getNetworkReachHosts(plugin) helper (returns the hosts only when permissions includes http, else none). So a manifest with allowedHosts but no http no longer advertises a "Network access" section for a capability the bridge will reject. Added component  tests for both cases (shown with http, hidden without). I kept the gating in the panel rather than wiring getPermissionDescriptions() into the UI, smaller surface, and it keeps the host chips as the reviewable surface.
 
 **fetch bypasses NetworkRetryInterceptorService** : documented in a comment on _requestNoRedirect ([d821c68](https://github.com/super-productivity/super-productivity/pull/8721/commits/d821c683883151b3a4a1afc980cc8cebb715345a)), noting the request path loses the single status-0 retry on Electron. Inherent to redirect: 'error' (XHR/HttpClient can't do it); left as-is per your steer.

**Native redirect-following** : agreed; the PR description now states explicitly that the redirect-block protection is web/desktop-only in practice (Capacitor's patched fetch ignores redirect/signal; native falls back to HttpClient).

**All redirects blocked incl. benign same-host** : acknowledged; documented as known behavior (plugin authors target final URLs).

**Credential-safety bonus** : good call, added to the description: redirect: 'error' also prevents the plugin's request (its Authorization header / body) from being replayed to a redirect target the user never approved — a credential-safety win on top of the SSRF fix.

Full plugin spec suite green (plugin-api 18, bridge 34, iframe 17, validate-manifest 8, plugin-management 8, plugin-http.service 37).

johannesjo:
Thank you very much! <3

Issue #8525, opened 2026-06-21
💡 Allow creating new projects, tags, and subfolders directly inside existing folders via the `...` menu


hajiboy95:
### Problem Statement

Currently, whenever I create a new folder, project, or tag, it is automatically added to the very top/root layer of the sidebar. If I want it inside a specific folder, I have to create it first and then manually drag or move it to the correct location. Even though folders already have a `...` (more options) context menu, there is no way to create a new item directly within a specific folder. This adds unnecessary friction and extra steps when trying to keep a workspace organized.

### Possible Solution

Add options like "Add new project," "Add new tag," and "Add subfolder" to the existing `...` context menu for folders. When a user clicks one of these options, the creation dialog should open, and the newly created item should automatically be placed inside that specific folder/subfolder, bypassing the root level entirely.

### Describe alternatives you've considered

* **Alternative 1:** Add a "Parent Folder" dropdown selector inside the general "Create New" dialog, allowing users to select the destination folder at the time of creation.
* **Alternative 2:** Rely purely on drag-and-drop from the top layer (current behavior), which is tedious for users with many folders or deeply nested hierarchies.

## Additional context

Adding these creation options directly to the folder's context menu aligns with standard file explorer behaviors across most operating systems. It would significantly speed up workflow and organization for power users who heavily utilize the folder structure.

PR #8722, opened 2026-07-03
feat(filetree): add option to add subfolders and items within folders #8525

hajiboy95:
Tackles #8525

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


hajiboy95:
@johannesjo should be ready for review

johannesjo:
Thanks for tackling this @hajiboy95 — the subfolder part works well, and the reducer + op-log wiring (action code, compact code, tests) is in good shape. There's one blocker plus a few smaller items before this is mergeable.

## 🔴 Blocker — creating a project or tag *inside a folder* duplicates it

When you create a project/tag from a folder's `…` menu, it ends up in **both** the folder **and** the sidebar root. The item is inserted twice into the (persisted, synced) menu tree, so this isn't a transient glitch — it's written to storage and propagates to every device.

**Why:** two reducer paths both insert, and nothing removes the first:

1. `addProject()`/`addTag()` call `ProjectService.add()`/`TagService.addTag()`, which dispatch `addProject`/`addTag`.
2. The `on(addProject)` handler (and the existing `on(addTag)` handler) append the item to the **tree root**.
3. Then `addItemToFolder` appends it to the **folder's children** — without removing the root copy.

`_buildViewTree` has no dedup within the stored-tree walk (`usedIds` only guards the append-missing fallback), so the item renders twice. On the next drag/reorder it gets re-persisted, so it compounds.

I reproduced it end-to-end against this branch — feeding the real two-action sequence through `MenuTreeService.buildProjectViewTree()` returns the project twice (`expected 2 to be 1`). Same for tags.

**Minimal fix** — make the folder insert idempotent by stripping any existing occurrence first (the reducer already has `_deleteItemFromTree`):

```ts
on(addItemToFolder, (state, { itemId, itemKind, folderId, treeType }) => {
  const insert = (tree: MenuTreeTreeNode[]): MenuTreeTreeNode[] =>
    _insertItemIntoFolder(
      _deleteItemFromTree(tree, itemId, itemKind),
      itemId, itemKind, folderId,
    );
  return {
    ...state,
    projectTree: treeType === MenuTreeKind.PROJECT ? insert(state.projectTree) : state.projectTree,
    tagTree: treeType === MenuTreeKind.TAG ? insert(state.tagTree) : state.tagTree,
  };
}),
```

Please add a test for the **combined** flow (`addProject` + `addItemToFolder`, and the tag equivalent) — the current reducer tests only exercise `addItemToFolder` from an empty tree, which is why this slipped through.

## 🟠 Worth addressing

- **Project creation dialog is inconsistent.** `addProject()` opens a bare title-only `DialogPromptComponent`, whereas every other entry point (side-nav, shortcut) uses `DialogCreateProjectComponent`. The tag path already reuses `DialogCreateTagComponent` — projects should be symmetrical, otherwise projects created from a folder lose theme/color setup.
- **Wrong placeholder.** The `addProject` dialog uses `F.PROJECT.D_CREATE.CREATE`, which is the **"Create Project"** button label, not a name hint. (Also prefer the typed `T.F...` constant over the hardcoded `'F.PROJECT.D_CREATE.CREATE'` string.)
- **Sync model (please sanity-check).** `MENU_TREE` is a singleton that previously mutated only via whole-tree `updateProjectTree`/`updateTagTree` replaces. This adds a granular `addItemToFolder` op with `opType: Create`, and "create in folder" now emits two ops per intent (`addProject` then `addItemToFolder`). Worth confirming behavior under concurrent multi-device edits (a concurrent whole-tree replace can clobber a granular insert) and that a `Create` op on an already-existing singleton resolves cleanly.

## 🟡 Minor

- The diff removes several unrelated `// eslint-disable-next-line @typescript-eslint/naming-convention` comments in `update-check.service.spec.ts` and `work-context.const.spec.ts`. Lint still passes, but these are unrelated to the feature — please drop them from this PR to keep the diff focused.

hajiboy95:
@johannesjo addressed

johannesjo:
Pushed two follow-up commits refining the sync design:

**`710dcd98` — dedupe folder-id prefix strip.** Extracted the repeated `'folder-'` strip into a `_cleanFolderId()` helper (6 call sites → 1).

**`b0d194bf` — add items to folders via the existing tree-update op (removes `MENU_TREE_ADD_ITEM`).**

A multi-perspective review flagged that the new `MENU_TREE_ADD_ITEM` op is avoidable and inconsistent with this PR's own "add subfolder", which already persists through the existing `updateProjectTree`/`updateTagTree` ops. So `addProjectToFolder`/`addTagToFolder` now compute the tree in the service and dispatch those existing ops. This:

- **removes a new synced op-type** (action, enum member, compact code `'MA'`, reducer handler) — the highest-risk change class in the op-log;
- is **forward-compatible**: old clients understand `updateProjectTree`/`updateTagTree` (and already have `on(addTag)`), so the item lands *in the folder* on every client — the granular `MA` op silently no-op'd on old clients, stranding the item at root;
- **never drops a placement**: `_insertNodeIntoFolder` falls back to root-append when the target folder is missing (e.g. deleted concurrently on another device);
- also drops the `on(addProject)` reducer handler, which changed persisted tree shape for *all* project creation beyond this feature's scope.

Trade-off: a concurrent drag-reorder on another device can still clobber a fresh placement (whole-tree last-writer-wins) — but that was already true for `deleteFolder`/`updateFolder` and drag-reorder, so this is now *consistent* with the rest of the menu-tree rather than a granular op fighting the whole-tree model.

Net −128 lines. New `MenuTreeService` specs cover insert, null-folder, missing-folder fallback, and tag move-dedup; reducer/enum/codec specs green.

johannesjo:
Thank you very much! <3

PR #8785, opened 2026-07-06
fix(sync): don't report lifetime op totals as "changes since last sync" in conflict dialog

aakhter:
## Problem

When the sync conflict dialog appears, it can report **thousands of "changes"** on a side that has only made a handful of edits, making the dialog alarming and its overwrite warnings misleading.

## Root cause

`DialogSyncConflictComponent.getChangeCount()` is meant to show *changes since the last successful sync*, computed as a per-client vector-clock delta against `lastSyncedVectorClock`. But when `lastSyncedVectorClock` is `null` it fell back to **summing the entire vector clock** - and vector-clock counters are per-client *lifetime* totals. So the fallback reports total ops ever performed, not changes since last sync.

`lastSyncedVectorClock` was effectively always `null` at the dialog because `SyncWrapperService._handleLocalDataConflict()` hardcoded it, and `LocalDataConflictError` never carried the client's last-synced baseline through from the sync engine.

## Fix

- Thread the real last-synced baseline (`lastSyncedVectorClock` + `lastSyncedUpdate`) through `LocalDataConflictError`, populated from the local snapshot's vector clock in `OperationLogSyncService`. Genuinely-fresh clients pass `null` explicitly.
- `getChangeCount()` now returns `number | null`. When there is no last-synced baseline the change count is *genuinely unknown*, so it returns `null` instead of summing the lifetime clock.
- The dialog renders an unknown count as "unknown" and uses count-free warning wording (`OVERWRITE_WARNING_UNKNOWN`). Overwrite confirmation is still shown in the unknown case, since overwriting could discard real data.

## Tests

- New `dialog-sync-conflict.component.spec.ts` covering: per-client delta counts, the null/unknown path (no lifetime-sum regression), highlight selection, and confirmation-message wording.
- New coverage in `operation-log-sync.service.spec.ts` asserting the last-synced clock is threaded into `LocalDataConflictError` for both the has-unsynced-ops and fresh-client paths.

Touches the live `op-log` sync engine only; the deprecated `pfapi` path is untouched.


johannesjo (commented):
## Review

Nice fix — this correctly kills the "lifetime op totals shown as changes since last sync" bug, and it's appropriately scoped to a **display heuristic** (the count feeds the dialog text, the highlight, and the overwrite-confirmation threshold — not the data-merge path), so the blast radius is low.

**Verified locally:** both new/updated specs pass (`dialog-sync-conflict.component.spec.ts` 5/5, `operation-log-sync.service.spec.ts` 82/82), lint (incl. the local sync rules) is clean, only `en.json` was touched among locales, and the `getConfirmationMessage` refactor preserves the known-count wording for both `USE_LOCAL`/`USE_REMOTE`. Null handling is symmetric (both sides go `null` together, so there's no asymmetric-comparison edge), and confirm-on-unknown is a sound safe default.

Two non-blocking follow-ups:

### 1. `getSnapshotVectorClock()` isn't strictly the "last-synced baseline" — it can undercount

The comments/PR describe the local snapshot clock as the client's last-synced baseline with "unsynced ops sitting on top of it." That holds when the snapshot came from the last remote hydration, but **compaction can fold unsynced ops into the snapshot clock**:

- A local op bumps the stored vector clock at *creation* time, before any sync — `operation-log-store.service.ts` `appendWithVectorClockUpdate()` (`source === 'local'` branch).
- Compaction snapshots `getCurrentVectorClock()` (which includes those unsynced ops) as `snapshot.vectorClock`, while only *deleting* synced ops — `operation-log-compaction.service.ts:89`, `:112`, `:129`.

So if a compaction runs while unsynced ops are pending, `localChangeCount = localClock − snapshotClock` reflects "ops since last compaction," not "unsynced ops," and can undercount all the way to `0`. Practical impact is bounded (display value + the `>= 20` overwrite-confirmation threshold could fail to warn on a modest batch of local edits) and it's still strictly better than the lifetime-sum it replaces — just worth softening the "last-synced" wording so it isn't read as exact.

Worth noting: the **exact** local unsynced count is already in hand as `error.unsyncedCount` (the dialog even renders it as "N local changes pending"). Using it for the local side would make that column precise; the vector-clock delta is only strictly needed for the remote side. Optional.

### 2. `lastSyncedUpdate` is threaded through every layer but never populated

All three `LocalDataConflictError` throw sites pass `null` for `lastSyncedUpdate`, and `_handleLocalDataConflict` reads `error.lastSyncedUpdate ?? null`. But the template *renders* `local.lastSyncedUpdate` as the "Last Sync" date row (`dialog-sync-conflict.component.html`), so that row stays blank (`-`) exactly as before. Either populate it (e.g. from the snapshot's `compactedAt`, to match the clock baseline it's paired with) or drop the parameter — as-is it's dead plumbing, and the PR description implies it carries a real value.

Neither blocks merge. 👍

aakhter:
Thanks! Addressed both follow-ups in 7b63fdc.

**1. `getSnapshotVectorClock()` baseline wording** - softened the docs on `getSnapshotVectorClock()` and `LocalDataConflictError` to spell out that it's an approximate baseline: compaction snapshots the current clock (which includes still-unsynced ops) while only deleting synced ops, so the `localClock - snapshotClock` delta can under-count, and the result should be read as a display heuristic rather than an exact figure.

I held off on switching the local side to `error.unsyncedCount` though. The wholly-fresh-client conflict throws with `unsyncedCount = 0` despite having meaningful store data, so using it there would render "0 local changes" - more misleading than the current "unknown". The delta's imprecision felt like the safer of the two for that path, but happy to revisit if you'd prefer the exact count with a special-case for the fresh-client throw.

**2. `lastSyncedUpdate` dead plumbing** - dropped the parameter from `LocalDataConflictError` (all three throw sites passed null and nothing populated it). Left the shared `ConflictData.local.lastSyncedUpdate` field and the dialog row intact and set it to null explicitly in the handler, so the dialog still renders "Never"/"-", which is correct for a no-last-sync conflict.

johannesjo:
Thank you very much! <3

Issue #6218, opened 2026-01-27
🚨 WebDAV sync complains about file modified on remote


patrickahles:
### Version Used

v17.0.5

### Expected Behavior

No sync errors

### Current Behavior

Whenever sync happens I get a message that the json file was modified on the remote:

<img width="658" height="82" alt="Image" src="https://github.com/user-attachments/assets/b17595e9-fd00-4c55-8314-0bbfbd8cc050" />

When testing the connection everything is OK. I've force updated the remote, and everything seems OK, however the time of the file stays stuck on the time of the last forced update.

### Steps to Reproduce

just press the sync button

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell
{
    "timestamp": "2026-01-27T11:22:26.627Z",
    "level": "LOG",
    "context": "sync",
    "message": "[SyncWrapper] Syncing vector clock to pf.META_MODEL",
    "args": [
      {
        "clockSize": 2,
        "lastUpdate": 1769512400488
      }
    ]
  },
  {
    "timestamp": "2026-01-27T11:22:26.631Z",
    "level": "LOG",
    "context": "sync",
    "message": "SyncWrapperService: Starting op-log sync...",
    "args": []
  },
  {
    "timestamp": "2026-01-27T11:22:26.631Z",
    "level": "LOG",
    "context": "ol",
    "message": "OperationLogDownloadService: Downloading remote operations via API...",
    "args": []
  },
  {
    "timestamp": "2026-01-27T11:22:26.631Z",
    "level": "VERBOSE",
    "context": "ol",
    "message": "OperationLogDownloadService: [DEBUG] Starting download. lastServerSeq=1, appliedOpIds.size=121, clientId=B_72L2",
    "args": []
  },
  {
    "timestamp": "2026-01-27T11:22:26.631Z",
    "level": "VERBOSE",
    "context": "pf",
    "message": "SyncCredentialStore.load",
    "args": [
      "WebDAV",
      "object"
    ]
  },
  {
    "timestamp": "2026-01-27T11:22:26.631Z",
    "level": "VERBOSE",
    "context": "pf",
    "message": "SyncCredentialStore.load",
    "args": [
      "WebDAV",
      "object"
    ]
  },
  {
    "timestamp": "2026-01-27T11:22:26.631Z",
    "level": "VERBOSE",
    "context": "pf",
    "message": "SyncCredentialStore.load",
    "args": [
      "WebDAV",
      "object"
[...]

PR #8786, opened 2026-07-06
fix(sync): atomic file-based sync writes — backup-before-overwrite + remove force-overwrite race

aakhter:
## Problem
File-based sync (Dropbox / WebDAV / local file) writes a single `sync-data.json` per upload, with two atomicity gaps:

1. **No backup / recovery artifact.** `BACKUP_FILE` (`sync-data.json.bak`) is declared but never written on the upload path. WebDAV servers and the local-file provider can leave a truncated/corrupt file on an interrupted write, which surfaces as a corruption error with no automated recovery.
2. **Lost-update race in the retry path.** On a rev mismatch, the retry could fall into an unconditional `isForceOverwrite=true` overwrite when the freshly fetched rev equalled the previous rev. If two clients race and the rev check is fooled (provider caching, rev reuse, eventual consistency), one client's ops could be silently clobbered.

## Fix
- **Backup-before-overwrite (two-phase write):** the current remote file is copied to `sync-data.json.bak` (re-uploading the bytes already in hand, so no extra GET) before the primary file is overwritten. Backup failure is non-fatal, so providers without copy support still sync.
- **Recovery on download:** a missing / empty / unparseable `sync-data.json` recovers from `.bak` and surfaces a user-visible notice.
- **No unconditional force-overwrite:** the rev-mismatch retry stays conditional; a genuine concurrent change throws a retryable error so the next cycle merges instead of overwriting. `isForceOverwrite=true` remains reachable only from explicit user actions ("Use Local" / snapshot upload).
- Reconciles a stale in-memory expected `syncVersion` against the remote on download.

## Testing
Adapter-level unit tests with a mock provider: `.bak` is written with the previous content before overwrite; download falls back to `.bak` on corruption / empty; the rev-mismatch retry never forces; and two interleaved clients never lose the first client's ops. Existing sync suites pass.


johannesjo (commented):
Thanks for this — the **lost-update fix is sound and well-motivated**. Making the main-file write strictly conditional and throwing a retryable error on a genuine concurrent change (instead of force-overwriting) is the right call. Worth noting for reassurance: removing the force branch is low-risk because the March-2026 switch to MD5 content-hash revs (`946339a0`) already made the old "same rev → force" path near-dead and obsoleted the original `#6218` broken-`If-Unmodified-Since` scenario that motivated it. Main writes stay conditional; force is now only `.bak`/snapshots/explicit user actions. 👍

Two issues on the **recovery half** that I think should be addressed before merge, plus minor notes.

---

### 1. Recovery won't trigger for encrypted or compressed sync files (the common real-world corruption case)

`_isRecoverableCorruption()` matches only `SyncDataCorruptedError | InvalidDataSPError | JsonParseError`. But a genuinely truncated/corrupt file is decoded through `decompressAndDecryptData()`, which throws:

- **`DecryptError`** when the file is encrypted (decrypt fails on garbage), and
- **`DecompressError`** when the file is gzip-compressed (decompress fails on a truncated body).

Both `extends AdditionalLogErrorBase` — neither is in the recoverable set — so `_downloadOps` rethrows and **never attempts `.bak` recovery**. Only the default *plaintext + uncompressed* config throws `JsonParseError` (recoverable) or `InvalidDataSPError` (empty file). So for users who enable encryption (the direction the app is pushing) or compression, the headline feature silently no-ops for exactly the "interrupted/truncated write" scenario it targets.

The unit tests don't catch this because they inject `SyncDataCorruptedError`/`InvalidDataSPError` directly into `downloadFile` rather than exercising a real truncated payload through decrypt/decompress.

**Fix:** add `DecryptError` and `DecompressError` to `_isRecoverableCorruption()`. It's safe — a wrong-key/wrong-format situation makes the `.bak` decode fail too, so `_recoverFromBackup()` returns `null` and the original error still surfaces. Please also add a test that decodes a genuinely truncated **encrypted** payload (not an injected error type) so this path is actually covered.

### 2. Recovery reads around the corrupt file but can't repair it, and pollutes the next upload's `revToMatch`

After a download-time recovery, the cache is seeded with the **`.bak` file's rev** (`_setCachedSyncData(providerKey, recovered.data, recovered.rev)`). On the next upload in the same cycle:

1. `_getCurrentSyncState` returns `revToMatch = bakRev`;
2. the conditional PUT to the still-corrupt `sync-data.json` compares `MD5(corrupt main) !== bakRev` → `UploadRevToMatchMismatchAPIError`;
3. the retry loop then re-downloads the main file (`_downloadSyncFile`, **outside** the `try/catch`), which re-throws the corruption error and propagates out of `_uploadWithMismatchFallback`.

Net effect: the upload fails with a corruption error, the main file is **never self-healed**, and every subsequent sync re-recovers on download (re-firing the snack) and re-fails on upload — a self-perpetuating degraded state until some client force-writes a snapshot. No data is lost (local ops stay in IndexedDB), but sync is stuck and noisy. This is untested — test `(b)` only calls `downloadOps`, never a follow-up `uploadOps`.

**Fix options:** capture the corrupt main file's rev (it's available from `provider.downloadFile` before decrypt throws) and use it as `revToMatch` so the next conditional PUT overwrites and repairs the file; and/or de-dup the recovery snack so it doesn't fire on every cycle (the "less noise" principle).

---

### Minor

- **Backup runs even on doomed uploads.** `_backupCurrentRemote` is called before `_uploadWithMismatchFallback`, so a client uploading from a stale cache still force-writes a full `.bak` before its conditional main PUT rejects on a concurrent change — wasted work that can also regress `.bak` to older content. Minor since `.bak` is disposable.
- **Re-encoding instead of reusing bytes.** `_backupCurrentRemote` re-runs `compressAndEncryptData(currentData)` rather than re-uploading the encrypted `dataStr` already fetched during download. Reusing the raw bytes would halve backup CPU, guarantee a byte-identical backup, and — since this second full PUT now happens on *every* sync — meaningfully cut the added bandwidth on large states / metered links. Worth confirming the every-sync backup cost is an intended tradeoff.
- **Stale-version reconcile block is a no-op log.** The new `if (previousExpectedVersion > syncData.syncVersion)` block only logs; its condition is already subsumed by `versionWasReset` and the actual reconcile (`_expectedSyncVersions.set(...)`) is pre-existing. Reads as behavior when it isn't.

aakhter:
Thanks for the thorough review! Addressed both before-merge items in 067bea9.

**1. Encrypted/compressed recovery gap** - added `DecryptError` and `DecompressError` to `_isRecoverableCorruption()`. You're right that a truncated encrypted/compressed body fails in decrypt/decompress, not JSON parse, so recovery silently no-opped for exactly those users. Safe as you noted: an undecodable `.bak` makes `_recoverFromBackup()` return null and the original error still surfaces.

**2. revToMatch pollution / self-perpetuating degraded state** - `_downloadSyncFile()` now annotates the corrupt primary's rev onto the decode error (no extra request, we already hold it), and recovery seeds the cache with that rev instead of the `.bak` rev. The next conditional upload then matches `sync-data.json` and overwrites/heals it instead of mismatching forever and re-recovering every cycle. Also de-duped the recovery snack per corrupt rev so a download-only client (which can't heal the file itself) doesn't re-toast on every sync.

**Tests** - added coverage that drives a REAL decode failure rather than an injected error type: a compressed body that isn't valid gzip produces a genuine `DecompressError` through `decompressAndDecryptData`. One test asserts `.bak` recovery, the other asserts the follow-up upload conditions on the corrupt primary rev and heals the file.

On the minors (every-sync `.bak` backup even on doomed uploads, re-encoding vs reusing the fetched bytes, and the stale-version reconcile log): left those as-is for now to keep this diff scoped to the correctness items, but happy to take any of them if you'd like them in here.

johannesjo:
Thank you very much! <3

PR #8787, opened 2026-07-06
fix(sync): SPAP-9 causality-aware conflict gating to cut false conflict dialogs

aakhter:
## SPAP-9 — causality-aware conflict gating

Resolves file-based sync conflicts by **causality** (vector clocks) instead of always surfacing the binary USE_LOCAL/USE_REMOTE dialog:

- **keep-local** when local strictly dominates the remote snapshot
- **apply-snapshot** (no dialog) when the snapshot strictly dominates local
- **dialog** when clocks are concurrent and can't be safely auto-resolved

### Scope
Rebased onto current `master` and split down to **only SPAP-9**. SPAP-8 (atomic writes) landed via #8786; SPAP-10 (rev-precheck) and SPAP-11 (split format) are separate PRs (#8795, #8802). This diff no longer carries them.

### Safety (addresses review)
- **Cosmetic syncVersion-reset is suppressed only on `EQUAL`.** A `GREATER_THAN` remote clock proves the writer did more work, not that this client received the intervening ops — so it now conservatively triggers a seq-0 resync rather than silently skipping a snapshot that may have compacted ops we never saw.
- **`AUTO_MERGE_CONCURRENT_SNAPSHOT` defaults OFF**, and when enabled the merge only runs if the retained recent ops provably bridge the whole gap to the snapshot clock (`local ⊔ recentOps ⊒ snapshot`); otherwise it falls back to the dialog instead of replaying only `recentOps` and silently dropping compacted-only entities. Re-enabling default-on is gated on multi-client E2E validation (SPAP-34).

### Tests
- Regression test for the dominating-clock / compacted-ops gap case.
- Auto-merge: happy-merge-when-bridged, and refuse-and-dialog-when-compacted-only.
- Full op-log spec suite green (2864), plus `@sp/sync-providers` (394) and `@sp/sync-core` (206).

### DIFF
--- a/packages/sync-providers/src/file-based-sync-data.ts
+++ b/packages/sync-providers/src/file-based-sync-data.ts
@@ -82,7 +82,28 @@ export const FILE_BASED_SYNC_CONSTANTS = {
   BACKUP_FILE: 'sync-data.json.bak',
   MIGRATION_LOCK_FILE: 'migration.lock',
   FILE_VERSION: 2 as const,
-  MAX_RECENT_OPS: 500,
+  // SPAP-9: raised 500 -> 2000 to shrink the gap window. A client that missed up
+  // to this many ops while offline can still catch up incrementally instead of
+  // tripping snapshotReplacement/partialTrimGap detection and forcing a full
+  // seq-0 resync (and the conflict dialog that path can surface). Compact ops are
+  // small (~150-250 bytes serialized each), so the extra 1500 retained ops add
+  // roughly ~0.3 MB to sync-data.json only when the buffer is actually full.
+  MAX_RECENT_OPS: 2000,
   SYNC_VERSION_STORAGE_KEY_PREFIX: 'FILE_SYNC_VERSION_',
   LEGACY_META_FILE: '__meta_',
+  // SPAP-9: when a seq-0 snapshot download has CONCURRENT vector clocks with the
+  // local client, attempt an entity-level last-write-wins merge of the remote
+  // recent ops instead of forcing the binary USE_LOCAL/USE_REMOTE conflict
+  // dialog.
+  //
+  // Default OFF (review follow-up): the merge only replays the capped `recentOps`
+  // buffer and never re-hydrates the compacted `snapshotState`, so a snapshot whose
+  // compacted base holds an entity this client never downloaded would silently drop
+  // that entity — turning a user-recoverable conflict dialog into permanent, global,
+  // undetectable data loss. `_tryConcurrentSnapshotMerge` now additionally refuses to
+  // merge unless the retained recentOps provably bridge the entire gap, but until that
+  // guard is validated by a real multi-client E2E harness (SPAP-34) we keep the
+  // feature opt-in and fall back to the conflict dialog. Flip to true only alongside
+  // that validation.
+  AUTO_MERGE_CONCURRENT_SNAPSHOT: false,
 } as const;
--- a/packages/sync-providers/tests/provider-types.spec.ts
+++ b/packages/sync-providers/tests/provider-types.spec.ts
@@ -129,4 +129,20 @@ describe('sync provider contracts', () => {
     expect(data.recentOps[0].sv).toBe(7);
     expect(FILE_BASED_SYNC_CONSTANTS.SYNC_FILE).toBe('sync-data.json');
   });
+
+  // SPAP-9: shrink the gap window so slow-syncing clients that miss up to
+  // MAX_RECENT_OPS ops can still catch up via incremental download instead of
+  // falling into the snapshot-replacement full-resync path.
+  it('retains 2000 recent ops to shrink the gap window (SPAP-9)', () => {
+    expect(FILE_BASED_SYNC_CONSTANTS.MAX_RECENT_OPS).toBe(2000);
+  });
+
+  // SPAP-9 (review follow-up): concurrent-snapshot auto-merge defaults OFF. The
+  // merge replays only the retained recentOps and never re-hydrates the compacted
+  // snapshotState, so it can silently drop compacted-only entities. It stays opt-in
+  // (behind a gap-covers guard) until validated by a multi-client E2E harness
+  // (SPAP-34); the safe default is the user-recoverable conflict dialog.
+  it('defaults concurrent-snapshot auto-merge OFF (SPAP-9 review follow-up)', () => {
+    expect(FILE_BASED_SYNC_CONSTANTS.AUTO_MERGE_CONCURRENT_SNAPSHOT).toBe(false);
+  });
 });
--- a/src/app/op-log/sync-providers/file-based/file-based-sync-adapter.service.spec.ts
+++ b/src/app/op-log/sync-providers/file-based/file-based-sync-adapter.service.spec.ts
@@ -874,6 +874,59 @@ describe('FileBasedSyncAdapterService', () => {
         // Should NOT detect gap because excludeClient === syncData.clientId
         expect(result.gapDetected).toBe(false);
       });
+
+      it('flags a gap when a dominating-clock reset compacted ops the client had not yet seen', async () => {
+        // Review follow-up (SPAP-9): a syncVersion regression whose remote clock is
+        // only GREATER_THAN last-seen must NOT be treated as cosmetic. GREATER_THAN
+        // proves the writer did more work, not that this client received it.
+
+        // First download: client syncs up to syncVersion 5, last-seen {client1:5}.
+        const compactOp = (
+          id: string,
+          v: number,
+        ): FileBasedSyncData['recentOps'][number] => ({
+          id,
+          c: 'client1',
+          a: 'HA',
+          o: 'ADD',
+          e: 'TASK',
+          d: id,
+          v: { client1: v },
+          t: Date.now(),
+          s: 1,
+          p: {},
+        });
+        const first = createMockSyncData({
+          syncVersion: 5,
+          vectorClock: { client1: 5 },
+          recentOps: [compactOp('op-5', 5)],
+        });
+        mockProvider.downloadFile.and.returnValue(
+          Promise.resolve({ dataStr: addPrefix(first), rev: 'rev-1' }),
+        );
+        await adapter.downloadOps(1);
+
+        // Writer made ops 6..10 (client never saw them), took a snapshot compacting
+        // 1..10 into `state` and reset recentOps, then made op 11. The file's clock
+        // {client1:11} is GREATER_THAN last-seen {client1:5}, but ops 6..10 survive
+        // only inside `state` — they are NOT in recentOps.
+        const second = createMockSyncData({
+          syncVersion: 2, // reset to 1 by snapshot, +1 for op 11
+          vectorClock: { client1: 11 }, // GREATER_THAN {client1:5}
+          recentOps: [compactOp('op-11', 11)],
+          oldestOpSyncVersion: 2,
+          state: { tasks: [] },
+        });
+        mockProvider.downloadFile.and.returnValue(
+          Promise.resolve({ dataStr: addPrefix(second), rev: 'rev-2' }),
+        );
+
+        const result = await adapter.downloadOps(5); // client's real cursor is 5
+        // The compacted ops 6..10 would be lost without a full seq-0 resync, so the
+        // regression must be reported as a gap (was false under the EQUAL||GREATER_THAN
+        // suppression).
+        expect(result.gapDetected).toBe(true);
+      });
     });
 
     it('should set seq counter to syncVersion after snapshot upload', async () => {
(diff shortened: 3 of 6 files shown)


johannesjo:
## Review — causality-aware conflict gating

Thanks for tackling this; the false-conflict dialog is a real annoyance and the causality-first approach is the right instinct. Three of the four sub-changes look good, but the headline feature — the default-on **CONCURRENT auto-merge** — has a correctness hole that can cause **silent, permanent, cross-client data loss** in exactly the scenario it's meant to handle. I'd hold the default-on merge until that's guarded.

### 🔴 Blocker: `_tryConcurrentSnapshotMerge` discards the compacted snapshot

On a seq-0 file download, `snapshotState` is the **complete** remote state (`state` = "causal state after all operations represented by this file"), while `result.newOps` is only the **capped** `recentOps` buffer (`slice(-MAX_RECENT_OPS)` on every upload). `_tryConcurrentSnapshotMerge` applies `result.newOps` via `processRemoteOps` and **never applies `snapshotState`**.

The justifying comment — *"the client already holds the shared base (it has a populated vector clock), so the only divergent remote work is the file's incremental recent ops"* — is the flaw. A populated local clock says the client has *its own* history; it says nothing about whether the client holds *another client's* base. Concrete, realistic case:

1. Devices A and B last synced at `{A:100, B:200}`.
2. B does 3000 more ops offline-to-A (`B:200→3200`). `recentOps` now holds only ~`B:1201..3200`; `B:201..1200` are trimmed — their effects survive **only** in `state`.
3. A makes one edit (`A:101`) and syncs. Local `{A:101, B:200}` vs snapshot `{A:100, B:3200}` → **CONCURRENT** → `merge`.
4. Merge applies `recentOps` (`B:1201..3200`), skips `snapshotState`. Entities from `B:201..1200` **never arrive on A**.
5. `processRemoteOps` merges the ops' clocks, advancing A to `{A:101, B:3200}`. A now *believes it is fully caught up* — no future gap, no self-heal.

This is precisely the gap scenario the PR targets (*"a client that accumulates more than MAX_RECENT_OPS of remote-side ops while offline always takes the gap path"*). The 500→2000 bump shrinks the window but doesn't close it, and the merge fires exactly when the window is exceeded.

It also **propagates to the shared file**: when A next uploads, `_buildMergedSyncData` writes `state: currentState` (A's incomplete NgRx state) with clock `{A:101, B:3200}`. Any client that later does a fresh/gap download adopts A's incomplete state → the loss goes global and is undetectable (the clock claims completeness). Checkpoint-D validation won't save it — repair drops dangling refs, it doesn't recover missing entities.

Regression framing: pre-PR this path threw `LocalDataConflictError` → dialog, whose USE_REMOTE branch hydrates the full snapshot. So this swaps a *user-recoverable* choice for *silent* loss, and it's **default-on for every file-based provider** (`AUTO_MERGE_CONCURRENT_SNAPSHOT: true`).

**Suggested guard:** only merge when `recentOps` provably bridges the *entire* gap — i.e. local's clock dominates the snapshot clock reduced by the recentOps' contributions (the client actually holds the compacted base). If it doesn't, fall back to full-snapshot hydrate or the dialog — never a partial merge. Concretely, the adapter already knows whether the snapshot was reached via a real trim/gap; that signal could gate the merge. And a test must exercise the **compacted-only remote entity** case — current test (c) uses a one-op `recentOps` that fully covers the snapshot, so it can't catch this.

### 🟡 `partialTrimGap` off-by-one vs per-batch `sv`

The `> sinceSeq` → `> sinceSeq + 1` change is correct only if `sv` is a dense **per-op** sequence. It isn't — every op in one upload batch shares `sv = currentSyncVersion + 1`, and `recentOps = slice(-MAX_RECENT_OPS)` can cut through the middle of a batch. So the oldest retained op can be a *partial* batch `sinceSeq+1`:

- Old threshold flagged that boundary as a gap → seq-0 → full snapshot → safe.
- New threshold treats `oldestOpSyncVersion === sinceSeq+1` as contiguous → incremental-only download → the trimmed **head** of batch `sinceSeq+1` is never fetched, and no snapshot is downloaded on that path to compensate → silently missed.

Narrow (needs a multi-op boundary batch trimmed exactly at the cut, client one batch behind), but real given how sensitive this area is. Worth either handling batch boundaries explicitly or documenting why a partial-batch trim at the boundary can't happen.

### 🟡 Dead branches + a test that gives false confidence

`planSnapshotHydration` already runs *before* this gate and compares `(local, snapshot)`, returning early for `EQUAL` and `GREATER_THAN` (local dominates). So only `local LESS_THAN snapshot` (→ `apply-snapshot`) and `CONCURRENT` (→ `merge`/`dialog`) ever reach `_classifySnapshotConflict`. That makes:

- the **`keep-local`** branch, and
- the **`EQUAL`** arm of `apply-snapshot`

unreachable in the real flow. Test (b) ("keeps strictly-ahead local") sets local `{A:5}` / snapshot `{A:2}`, which `planSnapshotHydration` resolves as `GREATER_THAN` and returns `no_new_ops` **before** the new gate — so it's green via the pre-existing short-circuit, not the new branch. On a hot sync path I'd drop the dead branches (or, if kept as defense-in-depth, add a comment saying so) to keep the diff honest and reviewable.

### 🟢 Looks good
- **`apply-snapshot`** (reached via strict `snapshot GREATER_THAN local`) does a full `hydrateFromRemoteSync(snapshotState)` — complete remote state, and the dominated local pending ops are already contained in the snapshot. Safe.
- **Benign version-reset gating** has real backstops: it only suppresses one of three OR'd gap triggers; `snapshotReplacement` (empty `recentOps`) and `partialTrimGap` still fire independently. `_lastSeenVectorClocks` being in-memory means a post-restart cosmetic reset degrades *safe* (treated as a real reset → resync). A one-liner noting that restart behavior is intentional would help.
- `MAX_RECENT_OPS` bump reasoning (~0.3 MB only when the buffer is full) is sound.

---

Happy to pair on a `recentOps`-covers-the-gap guard and a failing test that reproduces the compacted-only-entity loss if useful.

<sub>🤖 Reviewed with Claude Code</sub>

aakhter:
Thanks for the thorough review — you were right on both HIGH items. I've split the branch so this PR now contains **only** SPAP-9 (causality gating) rebased on current master (SPAP-8 landed via #8786, so it's gone from the diff), and addressed the findings:

**🔴 cosmetic-reset dropping compacted ops (GREATER_THAN) — fixed.**
The reset is now treated as cosmetic **only on `EQUAL`**. A `GREATER_THAN` remote clock no longer suppresses the reset, because (as you showed) it only proves the writer did more work, not that this client received the intervening ops — so a snapshot that compacted ops we never saw now correctly triggers a seq-0 resync. Added your regression test verbatim (`flags a gap when a dominating-clock reset compacted ops the client had not yet seen`); it fails on the old `EQUAL || GREATER_THAN` and passes now.

**🔴 `_tryConcurrentSnapshotMerge` discarding the compacted snapshot — fixed two ways.**
1. `AUTO_MERGE_CONCURRENT_SNAPSHOT` now defaults **OFF**, so the default path is the user-recoverable dialog, not a silent merge.
2. Even when enabled, the merge now refuses unless the retained ops provably bridge the whole gap: it reconstructs `local ⊔ ⨆(recentOp clocks)` and only merges if that dominates the snapshot clock. If the snapshot's compacted base holds ops we can't account for (your `{A:100,B:3200}` case), it falls back to the dialog instead of partial-merging. Added tests (d) happy-merge-when-bridged and (e) refuse-and-dialog-when-compacted-only.

Re-enabling default-on is gated on a real multi-client E2E harness, filed as **SPAP-34**.

**🟡 partialTrimGap off-by-one / dead branches** — filed as SPAP-34 follow-up (the trimmed-mid-batch boundary wants a real-flow test rather than a mock assertion).

johannesjo:
## Review — SPAP-9 causality-aware conflict gating

Verified the split-down against the head ref: applied all 6 files locally and ran the suites — `file-based-sync-adapter.service.spec.ts` **81/81 green** (including the new dominating-clock regression test), `operation-log-sync.service.spec.ts` **87/87 green**, `checkFile` clean on both source files. **This correctly closes the confirmed data-loss finding from the earlier review** — 👍 on narrowing the cosmetic-reset suppression to `EQUAL` only, the `partialTrimGap` off-by-one (`> sinceSeq + 1`), and defaulting `AUTO_MERGE_CONCURRENT_SNAPSHOT` OFF behind the bridge guard. Merge is safe: it's fail-safe (refuses → dialog when it can't prove losslessness).

Three things worth a look before/after merge — none block the fix itself:

### 1. `apply-snapshot` is the one new destructive default-on path — residual pruning assumption
The only genuinely-new default-on behavior is *snapshot strictly dominates local + meaningful data → auto-hydrate, no dialog* (was: dialog). It's destructive toward local data, so its safety rests on "snapshot clock dominates local clock ⟹ snapshot contains all local data." That holds for the realistic paths (every mutation increments the clock; `BackupService` mints a fresh clientId so a restore forces `CONCURRENT` → dialog).

The residual hole is **vector-clock pruning** (`MAX_VECTOR_CLOCK_SIZE = 20`). `limitVectorClockSize` preserves the local clientId but can drop *foreign* entries — so with >20 contributing clients, a pruned-away entry can make a snapshot *appear* to dominate local while actually lacking that pruned client's data, and we'd silently hydrate over it where the old code showed a dialog. Pre-existing clock limitation, very low probability (needs 20+ devices), but this PR is what newly routes it through a silent destructive apply. Worth at least a comment acknowledging the assumption, or gating `apply-snapshot` on "no pruning occurred."

### 2. Auto-merge is effectively inert *even when the flag is flipped on* (for SPAP-34)
`_tryConcurrentSnapshotMerge`'s bridge guard needs `result.allOpClocks`, but the download service only populates that under `forceFromSeq0`, which (per `sync-wrapper.service.ts` `forceFromSeq0: isProviderSwitch || undefined`) is set on provider-switch / rejected-ops downloads — **not** on the gap-triggered seq-0 re-download that produces most CONCURRENT snapshot conflicts. In that common case `bridgedClock` collapses to just `localClock`, compares `CONCURRENT` vs the snapshot, and the guard refuses → dialog. So enabling the flag after SPAP-34 won't actually merge in the common path as written — the happy-merge test only passes because it injects `allOpClocks` manually. Fail-safe, but the plumbing needs fixing before the flag flip means anything.

### 3. Minor — dead branches in `_classifySnapshotConflict`
`planSnapshotHydration` already short-circuits `EQUAL`/local-dominates to `no_new_ops` *before* the classify runs, so the `keep-local` and `EQUAL → apply-snapshot` branches are unreachable in this flow — the classify only ever sees `GREATER_THAN` (→ apply-snapshot) or `CONCURRENT` (→ merge/dialog). The "keeps strictly-ahead local" test passes via that earlier short-circuit rather than the branch it names. Harmless/defensive, but consider dropping the dead branches or noting the pre-condition so the test coverage isn't misleading.

aakhter:
Following up on this after some multi-client testing: the opt-in `AUTO_MERGE_CONCURRENT_SNAPSHOT` path has a guard in `_tryConcurrentSnapshotMerge` (`operation-log-sync.service.ts`) that never runs its intended logic on file-based sync.

**The guard** computes `bridgedClock` by reducing over `result.allOpClocks`. But `allOpClocks` is only populated when `downloadRemoteOps` is called with `forceFromSeq0: true` (`operation-log-download.service.ts` — pushed inside `if (forceFromSeq0)`, and returned only when `forceFromSeq0 && allOpClocks.length > 0`). The guard's only caller — the main classify path — does **not** set `forceFromSeq0`. The two callers that do (the USE_REMOTE force-download, and the rejected-ops handler) don't invoke the guard; they hydrate directly. So at the guard `allOpClocks` is always empty → `bridgedClock = localClock`, which is CONCURRENT with the snapshot by the method's own entry condition → the guard **always returns null → always falls back to the conflict dialog.**

**Effect:** with `AUTO_MERGE_CONCURRENT_SNAPSHOT` enabled, the merge never fires — safe, mergeable concurrent snapshots (e.g. a client's first sync where the retained `recentOps` are the complete remote history) go to the binary conflict dialog instead of auto-merging. Reproduced on a local WebDAV harness: rewiring the reduction to `result.newOps[].vectorClock` flips such a case from dialog to a lossless auto-merge.

**But that naive rewire isn't a safe fix:** vector-clock domination proves causal *knowledge*, not op-set *completeness* — a `newOps` set whose per-client max clocks dominate the snapshot can still be missing interior ops (e.g. holds `{A:10}` without ops `A:6..9`), so replaying it could silently drop entities. A sound fix would need an actual completeness proof — e.g. surface a conservative "retained ops are contiguous / lineage intact" signal from the file-based adapter (where `oldestOpSyncVersion` and the trim-gap detection already live) up to the guard, and require both that AND the `newOps`-clock check before merging; or hydrate the remote snapshot and replay local pending ops; or route uncertain cases to the differences-review UI.

Happy to open a PR in whichever direction you prefer. Is the empty `allOpClocks` at the guard intended as a conservative fail-safe?

johannesjo:
Thanks for digging into this with a real multi-client harness — your read is correct, and it lines up with a follow-up I'd already flagged on this PR ([here](https://github.com/super-productivity/super-productivity/pull/8787#issuecomment-4905252738), item #2). Short version: yes, the empty `allOpClocks` at the guard is a conservative fail-safe, and the merge is effectively inert as shipped. It's tracked under SPAP-34 and gated behind `AUTO_MERGE_CONCURRENT_SNAPSHOT` (still default-`false`), so there's no live data risk today — it just means the flag wouldn't actually merge anything if flipped.

One precision on the mechanism, because it changes where a fix should land: `allOpClocks` isn't *always* empty at the guard. The classify path's download options come from `sync-wrapper.service.ts` as `forceFromSeq0: isProviderSwitch || undefined`, so there's exactly one path where the guard sees a populated set:

- **First-ever sync** (`lastSyncedProvider === null`) — your repro — `forceFromSeq0` falsy → `allOpClocks` empty → `bridgedClock` collapses to `localClock` → CONCURRENT → refuse → dialog. ✅ matches what you saw.
- **Ordinary gap-triggered re-download** — same, `forceFromSeq0` falsy → refuse → dialog. This is the common case.
- **Provider switch** (e.g. Dropbox → WebDAV, `lastSyncedProvider !== null && !== current`) — `forceFromSeq0: true` → `allOpClocks` *is* populated, and the guard can actually fire.

You're right that the two explicit `forceFromSeq0: true` callers (the USE_REMOTE force-download and the rejected-ops handler) don't invoke the guard — but the main classify path itself sets it on provider switch, so the guard is inert on first-sync/ordinary-sync (every case you'd hit in practice) rather than strictly dead.

That distinction matters because it points at the intended design: the guard was meant to consume the *full seq-0 op-clock set* (`allOpClocks` captures every server op's clock, not just `newOps`) as its completeness proof, and that set only exists after a `forceFromSeq0` download. The gate never triggers a seq-0 download before merging, so outside provider-switch it has no completeness evidence and correctly refuses.

And I fully agree with your call that the naive rewire to `result.newOps[].vectorClock` isn't safe — domination proves causal *knowledge*, not op-set *completeness* (your `{A:10}` without `A:6..9` case), which is exactly what the guard comment at `operation-log-sync.service.ts:1020-1033` is warning about. Rewiring to `newOps` would reintroduce the silent-drop risk we just closed.

A PR would be very welcome, in the **completeness-proof** direction rather than the `newOps` rewire. Two viable shapes, in rough order of preference:

1. Make the merge branch obtain the full/contiguous op set before merging — i.e. do a real `forceFromSeq0` download there so `allOpClocks` is populated legitimately, then apply the existing bridge check. Cleanest reuse of what's already there.
2. Surface a conservative "retained ops are contiguous / lineage intact" signal from the file-based adapter (where `oldestOpSyncVersion` and the trim-gap detection already live) up to the guard, and require **both** that signal **and** the `newOps`-clock check before merging; otherwise fall back to the dialog.

Either way it stays behind the flag until we have multi-client E2E coverage (SPAP-34) proving losslessness, so no rush. If you want to take #1, happy to review.

PR #8788, opened 2026-07-06
fix(tasks): persist collapsed subtask state across restart

### LINKED ISSUE #8781: 🚨Collapsed subtasks re-expand after restart
### Version Used

### Describe the Bug
The collapsed/expanded state of subtasks is not persisted. After restarting the app, previously collapsed subtasks are shown expanded again.

### Additional Context
This happens with sync disabled / purely local, so it is not related to a sync conflict. Occurs consistently after every app restart.

### Expected Behavior

The collapsed/expanded state of subtasks should persist across app restarts.

### Steps to Reproduce

1. Open a task that has subtasks
2. Collapse the subtask list
3. Fully close the app (remove from recent apps)
4. Reopen the app and navigate to the same task
5. Subtasks are expanded again, despite having been collapsed before closing

### Console Output

```shell

```

### Documentation

costajohnt:
## Problem

Fixes #8781. Collapsing or expanding a task's subtasks did not survive a restart, the state reset to fully expanded on every reload.

The collapse mode lives on the task (`_hideSubTasksMode`), but it was written by two non-persistent actions (`updateTaskUi` and `toggleTaskHideSubTasks`), so the op-log never captured the change. On reload the state rehydrated from a snapshot that predated the toggle, so the collapse was dropped.

## Solution

Persist the collapse mode as an absolute value through the op-log.

- `toggleSubTaskMode` now resolves the next `_hideSubTasksMode` at dispatch time and persists that absolute value via `updateTaskUi`, instead of dispatching a relative toggle command. The cycling logic is extracted unchanged into a pure `getNextHideSubTasksMode` helper with its own unit tests.
- `updateTaskUi` is marked persistent (an absolute `Update<Task>` op, entity `TASK`, opType `Update`), so it is captured to the op-log and replayed safely.
- The old relative `toggleTaskHideSubTasks` action and its reducer are removed. Persisting a relative delta would be non-deterministic on replay, and its reducer threw on a missing task, which could abort an apply batch. The absolute `updateTaskUi` path avoids both, since its reducer is a safe no-op on a missing id.
- Registered `[Task] Update Task Ui` in the ActionType enum and the compaction code map.

One thing worth your call. This field was historically part of synced app data (there's a legacy migration from `_showSubTasksMode` to `_hideSubTasksMode`), so persisting it through the op-log restores that behavior, which means collapse state now also syncs across devices. The issue only asks for it to survive a restart, and that works locally on the IndexedDB backend (structured clone preserves the `undefined` clear). The one gap is that the "expand all" clear does not converge over sync, because JSON serialization on the sync transport drops `undefined` keys, so collapsing propagates cross-device but clearing does not. If you'd rather keep collapse state local-only, or want the clear to converge cross-device too, happy to adjust. A small explicit "shown" value instead of `undefined` would make the clear round-trip. Let me know which way you'd prefer.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have included relevant changes to the documentation/wiki (not applicable)
- [x] I have added tests for my changes (cycling helper tests plus `updateTaskUi` persistence test)
- [x] Existing tests still pass (full op-log and tasks suites)

### DIFF
--- a/src/app/features/tasks/store/task.actions.ts
+++ b/src/app/features/tasks/store/task.actions.ts
@@ -37,19 +37,25 @@ export const __updateMultipleTaskSimple = createAction(
   }),
 );
 
+// NOTE: despite the "Ui" name this carries _hideSubTasksMode, which is task
+// data that must survive a restart. It is persisted (op-log) like updateTask so
+// the collapse state is not lost on reload. See issue #8781.
 export const updateTaskUi = createAction(
   '[Task] Update Task Ui',
-  props<{ task: Update<Task> }>(),
+  (taskProps: { task: Update<Task> }) => ({
+    ...taskProps,
+    meta: {
+      isPersistent: true,
+      entityType: 'TASK',
+      entityId: taskProps.task.id as string,
+      opType: OpType.Update,
+    } satisfies PersistentActionMeta,
+  }),
 );
 
 export const removeTagsForAllTasks = createAction(
   '[Task] Remove Tags from all Tasks',
   props<{ tagIdsToRemove: string[] }>(),
-);
-
-export const toggleTaskHideSubTasks = createAction(
-  '[Task] Toggle Show Sub Tasks',
-  props<{ taskId: string; isShowLess: boolean; isEndless: boolean }>(),
 );
 
 export const moveSubTask = createAction(
--- a/src/app/features/tasks/store/task.reducer.spec.ts
+++ b/src/app/features/tasks/store/task.reducer.spec.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/naming-convention */
-import { Task, TaskDetailTargetPanel, TaskState } from '../task.model';
+import { HideSubTasksMode, Task, TaskDetailTargetPanel, TaskState } from '../task.model';
 import { initialTaskState, taskReducer } from './task.reducer';
+import { convertOpToAction } from '../../../op-log/apply/operation-converter.util';
 import * as fromActions from './task.actions';
 import { TaskSharedActions } from '../../../root-store/meta/task-shared.actions';
 import { INBOX_PROJECT } from '../../project/project.const';
@@ -11,7 +12,7 @@ import {
 import { _resetDevErrorState } from '../../../util/dev-error';
 import { PlannerActions } from '../../planner/store/planner.actions';
 import { loadAllData } from '../../../root-store/meta/load-all-data.action';
-import { OpType } from '../../../op-log/core/operation.types';
+import { ActionType, OpType, Operation } from '../../../op-log/core/operation.types';
 
 describe('Task Reducer', () => {
   const createTask = (id: string, partial: Partial<Task> = {}): Task => ({
@@ -1132,6 +1133,52 @@ describe('Task Reducer', () => {
       expect(action.meta.entityType).toBe('TASK');
       expect(action.meta.entityId).toBe('task1');
       expect(action.meta.opType).toBe(OpType.Update);
+    });
+
+    // The metadata assertion above proves the change is *eligible* for capture.
+    // This one proves it actually survives the whole op-log path end to end:
+    // dispatch -> capture into an operation payload -> serialize -> convert the
+    // op back to an action -> replay through the reducer. The JSON serialize hop
+    // mirrors the sync transport and the SQLite op-log backend, both of which
+    // round-trip op payloads as JSON, so it is where a lost value would regress.
+    // See issue #8781.
+    it('should round-trip _hideSubTasksMode through capture, serialization and replay', () => {
+      const stateShown: TaskState = {
+        ...initialTaskState,
+        ids: ['task1'],
+        entities: { task1: createTask('task1') },
+      };
+
+      // 1. Dispatch: the persistent action toggleSubTaskMode dispatches on collapse.
+      const action = fromActions.updateTaskUi({
+        task: {
+          id: 'task1',
+          changes: { _hideSubTasksMode: HideSubTasksMode.HideAll },
+        },
+      });
+
+      // 2. Capture: the effects store the action fields under payload.actionPayload.
+      const op: Operation = {
+        id: 'op-8781',
+        actionType: action.type as ActionType,
+        opType: action.meta.opType,
+        entityType: action.meta.entityType,
+        entityId: action.meta.entityId as string,
+        payload: { actionPayload: { task: action.task }, entityChanges: [] },
+        clientId: 'clientA',
+        vectorClock: { clientA: 1 },
+        timestamp: 0,
+        schemaVersion: 1,
+      };
+
+      // 3. Serialize over the wire / into the op-log, then read it back.
+      const wireOp = JSON.parse(JSON.stringify(op)) as Operation;
+
+      // 4. Convert the persisted op back into a replayable action and replay it.
+      const replayAction = convertOpToAction(wireOp);
+      const replayed = taskReducer(stateShown, replayAction);
+
+      expect(replayed.entities.task1?._hideSubTasksMode).toBe(HideSubTasksMode.HideAll);
     });
   });
 });
--- a/src/app/features/tasks/store/task.reducer.spec.ts
+++ b/src/app/features/tasks/store/task.reducer.spec.ts
@@ -11,6 +11,7 @@ import {
 import { _resetDevErrorState } from '../../../util/dev-error';
 import { PlannerActions } from '../../planner/store/planner.actions';
 import { loadAllData } from '../../../root-store/meta/load-all-data.action';
+import { OpType } from '../../../op-log/core/operation.types';
 
 describe('Task Reducer', () => {
   const createTask = (id: string, partial: Partial<Task> = {}): Task => ({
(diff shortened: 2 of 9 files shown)


johannesjo (commented):
Thanks for the thorough writeup and the honest tradeoff section — the diagnosis is right and the approach is the correct one.

## What's good

- **Root cause is right.** The mode was written by non-persistent actions, so the op-log never captured it. Routing it through an absolute `updateTaskUi` op is the replay-safe fix.
- **Dropping the relative `toggleTaskHideSubTasks` is the correct call**, for exactly the reasons you give: a relative delta is non-deterministic on replay, and its reducer threw via `getTaskById` on a missing task (could abort an apply batch). The `updateOne` path no-ops safely on a missing id — a real robustness win.
- **Scope is contained.** `updateUi()` is only ever called with `_hideSubTasksMode`, so making `updateTaskUi` persistent doesn't leak any other UI-only field into the op-log. The cycling logic is preserved exactly and now pure + unit-tested. Enum member + compaction code (`TUU`, unique) + count test all line up, and `[Task] Update Task Ui` correctly doesn't match the `[TASK] LWW Update` pattern so it skips the LWW id-forcing path in the converter.

## One correction to the tradeoff framing (the main reason I'd lean toward the sentinel)

The description frames the `undefined`-drop as **only a cross-device sync gap**, and says the clear "works locally on the IndexedDB backend." That's true for IndexedDB — but the JSON-drop also bites a **second, local, single-device** surface: the **SQLite op-log backend**.

`sqlite-op-log-adapter.ts` stores every op via `JSON.stringify(value)` (the `buildInsert` path) and reads it back with `JSON.parse`. So on that backend:

1. Collapse → op `changes: { _hideSubTasksMode: 2 }` → stored fine → survives restart ✅ (the actual fix works on both backends)
2. Expand → op `changes: { _hideSubTasksMode: undefined }` → serialized as `changes: {}` → on restart, `updateOne({ changes: {} })` is a no-op → the earlier collapse op wins → **the task reappears collapsed after a local restart** ❌

So it's not purely a sync-transport issue — the same `undefined`-drop reverts the "expand all" clear on a single device once SQLite is the local backend.

This is **latent today** (production defaults to `IndexedDbOpLogAdapter` on every platform; SQLite is only provided in tests), so it's not a merge blocker. But the SQLite migration is heading toward being the Android default, and whoever flips that on is unlikely to remember this coupling. The fragility is that the design leans on the storage layer happening to preserve `undefined`, which only IndexedDB's structured clone does.

**So I'd take the explicit "shown" sentinel you proposed** — it fixes both the cross-device convergence *and* this future SQLite-local case in one move, by making the clear a real round-trippable value instead of an absence. The one cost to weigh: `HideSubTasksMode` is a synced numeric enum, so a new member is a forward-compat concern for older clients receiving the new value — worth a quick check that an unknown value degrades gracefully (it should, since the template treats anything truthy as "hide").

## Product decision to confirm

With this change, collapse state now **syncs across devices** (it was effectively local before) — collapsing on your phone collapses it on the desktop. You already flagged this; I'm fine with it since it was historically synced app data, just noting it's a deliberate behavior change, not an accident.

## Minor

- Every collapse/expand click now writes an op + triggers an immediate upload. Bounded per user action and no worse than any task-field edit, assuming compaction supersedes repeated updates to the same task/field — just noting the added sync chatter for a display-only toggle.

Overall: happy to merge the mechanism. My only real ask is a decision on `undefined` vs. a "shown" sentinel — and given the SQLite-local angle above, I lean sentinel.

costajohnt:
Good call on the sentinel, and you're right that the undefined drop also bites the SQLite local backend, not just sync. I went with it.

Added an explicit `HideSubTasksMode.Show = 0` and now persist that instead of `undefined` when clearing, so the shown state is a real serializable value on every backend. I kept it at 0 (falsy) specifically for the forward compat case you flagged. Every display read of the mode is either a truthy check (`!!_hideSubTasksMode`, `[isHideAll]`) or an exact `=== HideAll` / `=== HideDone` match, so an older client that receives a `0` over sync renders the task as fully shown everywhere. A truthy sentinel like 3 would have been the dangerous choice since `!!3` flips `[isHideAll]` on.

The one behavior an older client can't get right is the arrow right keyboard handler, which still uses the old `!== undefined` test, so on an old client arrow right on a shown task does nothing instead of opening the detail panel. It's keyboard nav only, no display or data impact, and it's inherent to sending a new enum value to a client that predates it. New clients treat `0` and legacy `undefined` identically as shown (the one `!== undefined` read is now a truthy check, equivalent for the existing values).

I left `archive.service.ts` writing `undefined` when it clears the mode on archive since it reads identically to `Show` and isn't on the live toggle path, happy to align it if you'd rather it be consistent.

Cycling logic is unchanged and the pure helper now returns the explicit value including Show, covered by the unit tests.

johannesjo:
## Review — persist collapsed subtask state (#8781)

Thanks for the thorough writeup and the honest "one thing worth your call" note — that framing is exactly right, and it turns out to be the crux. The local implementation is clean and the replay design is correct. But there's one blocker to flag before merge, and it's a different (more serious) issue than the "should collapse state sync?" question.

### 🔴 Blocker: `HideSubTasksMode.Show = 0` breaks already-released clients' validation

**Persisting `_hideSubTasksMode: 0` will make older clients that share a sync target throw a false "your data is corrupt — restore from a backup or contact support" dialog, repeating on every sync. This is verified, not theoretical.**

The chain:

- `_hideSubTasksMode` is validated on every client via `createValidate<TaskState>()` (`validation-fn.ts:40`), which runs on synced/imported state (`validateAfterSync`, hydration). There's no `_`-prefix stripping before validation.
- On the current release the enum is `{ HideDone = 1, HideAll = 2 }`, and **typia compiles a numeric enum as a closed set**. I confirmed this empirically by building a validator from the old enum and running it: `validate({ _hideSubTasksMode: 0 }).success === false` (same result as a bogus `99`). So `0` reads as corruption to any client that hasn't shipped `Show = 0`.
- A new client writes `0` on the everyday "expand all" action (`showSubTasks()` and cycling back to shown). It reaches an old client via a full-state snapshot or the file-based-sync whole-state JSON.
- On the old client: typia fails → blocking `confirm()` repair prompt → `dataRepair`/`autoFixTypiaErrors` have **no branch** for this field, so repair fails → `alertDialog("Repair attempted but failed… restore from a backup or contact support")`. Because `0` stays in state, it recurs on every subsequent sync.

**Why this can't be fixed inside this same PR:** a forward-compat relaxation (like the existing `isForwardCompatibleProviderKeyError`) or retyping the synced field to a plain `number` only protects clients that *ship that change*. The clients that break are the ones already in the field with the closed enum baked into their bundle. So the release that *writes* `0` has to come *after* a tolerant release is already adopted.

Worth noting the codebase already documents this exact hazard class — `isForwardCompatibleProviderKeyError` in `validation-fn.ts` exists specifically to suppress these false-corruption errors for `issueType`/`issueProviderKey`, and its comment says: *"if more open-ended enums appear, prefer typing them as opaque strings in the synced model."*

### The tradeoff, and options

`Show = 0` exists only to make the "expand-all → shown" clear survive on the SQLite op-log backend (Android default) and converge over sync, since `JSON.stringify` drops `undefined`. That's a real gap, but the fix as written is the forward-incompatible part. **The break is entirely isolated to the second commit** (`fix(tasks): persist an explicit Show value…`) — the first commit only ever writes `1 | 2 | undefined`, all of which old clients accept, and it already fixes the reported "collapse doesn't survive restart" bug.

1. **Ship commit 1 only (recommended for now).** Safe for old clients; fixes #8781 for collapse on all backends. Residual: on Android/SQLite the "expand-all" clear won't stick across a restart (falls back to the last collapsed value) — minor, no corruption, and strictly better than today.
2. **Two-phase rollout** if you want the full clear-convergence: release a version that *tolerates* `0` (relaxation + retype the synced field to `number`) first, then a later release starts *writing* `0`.
3. Keep as-is only if you're comfortable with the mixed-version window showing those dialogs (no data is actually lost — the alarm is false — but it's alarming and recurring).

### 🟡 Secondary (non-blocking)

- **Test gap:** the added persistence test only asserts the action's `meta` shape — nothing exercises the actual capture → serialize → replay round-trip for `_hideSubTasksMode`, nor `toggleSubTaskMode`/`showSubTasks` at the service level. The literal bug could silently regress with the test still green. An explicit round-trip assertion (dispatch → op captured → `convertOpToAction` → re-dispatch → value present) would match the sync-correctness bar.
- **Naming smell:** a `Ui`-named action + the permanent op code `'TUU'` now carry synced, persisted data (op codes are "never reassign"). Latent trap — a future contributor may route genuinely-ephemeral UI state through `updateTaskUi` and silently sync it. A rename or a louder comment would help.

### ✅ What's solid

- Persisting a **resolved absolute value** via `updateTaskUi` rather than a relative toggle command is the correct call for replay determinism — nicely done, and the inline comments explain the why.
- Extracting `getNextHideSubTasksMode` is arithmetically identical to the deleted reducer (all cycling cases traced); the refactor also incidentally fixes a latent crash (old reducer threw on a missing parent/subtask entity).
- The `Show = 0` semantics are consistent across *every* read site on the new client — only `handleArrowRight` needed the `!== undefined` → `!!` change, and it was made correctly.
- Correctly **not** an LWW action type; `updateOne` merges disjoint fields, so a concurrent collapse-toggle vs. title-edit doesn't clobber. Persistence flip is scope-contained (`updateUi` only ever writes `_hideSubTasksMode`). Hot-path `task.component` untouched except one event handler. Enum/code registration is consistent and wire-forward-safe (unknown action types decode to a no-op).

costajohnt:
Good catch, and you're right that this is the more serious issue. My forward-compat reasoning only covered the display reads (every read of the mode is a truthy or exact-equality check, so an old client renders `0` as shown), but that all happens after validation, and typia rejects `0` on the closed enum before any of it runs. The `isForwardCompatibleProviderKeyError` path only tolerates unknown strings, so a numeric `0` on `_hideSubTasksMode` falls straight through to the repair dialog. That's a real recurring break for already-released clients, not something a same-PR relaxation can fix.

So I've gone with option 1: dropped the `Show = 0` commit, keeping just the first. That fixes #8781 (collapse survives restart) on both backends and only ever writes `1 | 2 | undefined`, all of which old clients accept. The residual is the one you named: on the SQLite op-log backend "expand all" won't stick across a restart (it falls back to the last collapsed value). No corruption, and strictly better than today. If SQLite becomes the Android default and you want the clear to converge, the two-phase rollout (ship a tolerant release first, then start writing an explicit value) is the safe way to get there.

On the test gap, you're right that the persistence test only checked the action's meta. I added a round-trip test that runs the value through the full path (dispatch, capture into the op payload, JSON serialize, convertOpToAction, replay through the reducer) and asserts the collapse value comes back, so the actual sync behavior is covered rather than just the metadata.

Left the `updateTaskUi` / `TUU` naming as-is for now since that's the mechanism you were happy to merge, but glad to add a louder comment or rename if you'd prefer.

johannesjo:
Thank you very much! <3

Issue #8107, opened 2026-06-07
💥 There was a conflict syncing issue, which resulted in many tasks lost


bbinet:
### Version Used

18.9.1WB (online web app)

### Steps to Reproduce

In the web app, it was not in sync with the last state on my phone and I have marked a task as done (which was already done on the phone), then sync must have been triggered and superproductivity detected a conflict.
So I choose to sync from the server (which I thought was the more up to date).
But then after sync had finished, many tasks were removed and state was restore to something like a few days ago!

It's strange as I had run superproductivity both on my phone and on the web app recently...

FYI, I'm using supersync as the sync method.

### Can you reproduce this reliably?

No, it only happened once.

### Console Output

```shell

```

bbinet:
Then I tried to do "Restore from History" by selecting the last point in time, but it fails:

<img width="669" height="790" alt="Image" src="https://github.com/user-attachments/assets/95bc0453-626b-45f1-a318-1af4609f70f8" />

bbinet:
It would be great if we could have a way to restore data from history, like in Git.

bbinet:
I saw that "Enable automatic backups" is enabled on my Android device, so i hope I can find some old backups to restore a recent state of my data, but I can't find where the backup files are stored on my android device.
Do you know where they are stored?

bbinet:
I have downloaded SupKeyValStore db from android and tried to explore the "backup" key with sqlite3, but sadly this seem to be the current state of the data (lost tasks are not available here) : nothing to restore from it.

And the damage was more serious than I thought: I notice that even complete projects had been deleted...

bbinet:
My issue providers and connected calendars were also removed 😢

johannesjo:
Hi @bbinet – first, I'm genuinely sorry. Losing tasks, completed projects, and your issue/calendar connections like this should never happen, and I really appreciate how much careful digging you did (down to inspecting the `SupKeyValStore` backup with sqlite3) trying to get it back.

Here's what happened, as best I can reconstruct it: when the conflict appeared and you chose to take the server's version, the app **replaced** your local data with the server's snapshot – and that snapshot was older than your device. On top of that, three things lined up against you:
- the operation kept **no local copy** to roll back to,
- the conflict dialog **recommended** that choice without warning it might be older, and
- **"Restore from History" can't work while end-to-end encryption is on** (the server can't read your encrypted data to rebuild an earlier point) — so every recovery path you tried was a dead end.

[...]

PR #8795, opened 2026-07-06
perf(sync): SPAP-10 skip full download when the remote rev is unchanged

aakhter:
## SPAP-10 — skip the full download when the remote rev is unchanged

Adds a cheap `getFileRev` pre-check to the file-based sync download path: when the provider reports the same rev we last confirmed, skip the full `sync-data.json` download and return an empty "nothing new" result. Any `getFileRev` error falls through to a full download, so the check never fails a sync.

### Scope
Rebased onto current `master` and split down to **only SPAP-10**. SPAP-8 (atomic writes) landed via #8786; SPAP-9 (causality gating) is #8787. This diff no longer carries them or the `Merge SPAP-8 + SPAP-9` commit.

### Addresses review
- **Blocking 1 — rev persisted before ops applied:** the downloaded rev is now staged as *pending* and promoted to last-seen (and persisted) only in `setLastServerSeq`, i.e. after the caller has durably applied the ops — the same ordering as the seq cursor. A throw/crash between download and apply can no longer strand a rev ahead of un-applied ops and skip them on the next poll. Regression test: poll-1 downloads but never commits → poll-2 re-downloads instead of short-circuiting.
- **Blocking 2 — WebDAV pre-check is a pessimization:** allowlist corrected to `[Dropbox, OneDrive]` (both have a genuine cheap metadata/ETag rev call). WebDAV (full-body GET in `getFileRev`) and LocalFile are excluded, so there's no "download twice on change / zero saving on no-change" case. Fixed the inaccurate comments too. This also makes the "should invalidate cache after upload" test pass on WebDAV again (no short-circuit); the LocalFile isolation stays as explicit coverage.

### Follow-up
- **SPAP-29** — if WebDAV is ever re-added to the allowlist, gate it on a strong ETag / PROPFIND `getetag` so a weak/second-granularity validator can't skip a real update.

### Tests
Full op-log spec suite green (adapter + integration), including the updated rev-precheck specs (now exercised on Dropbox) and the new crash-safety regression test.

### DIFF
--- a/src/app/op-log/sync-providers/file-based/file-based-sync-adapter.service.spec.ts
+++ b/src/app/op-log/sync-providers/file-based/file-based-sync-adapter.service.spec.ts
@@ -2421,4 +2421,267 @@ describe('FileBasedSyncAdapterService', () => {
       expect(parseWithPrefix(uploadedMain).syncVersion).toBe(3);
     });
   });
+
+  describe('remote-rev pre-check (SPAP-10)', () => {
+    const STATE_KEY = FILE_BASED_SYNC_CONSTANTS.SYNC_VERSION_STORAGE_KEY_PREFIX + 'state';
+
+    // Simulate crossing a poll boundary: the intra-cycle cache is only meant to
+    // dedupe the download+upload of a SINGLE sync cycle. On the next poll it is
+    // stale/absent. Clearing it here reproduces that cross-poll condition without
+    // waiting out the real 30s TTL.
+    const crossPollBoundary = (): void => {
+      (
+        service as unknown as { _syncCycleCache: Map<string, unknown> }
+      )._syncCycleCache.clear();
+    };
+
+    // The rev pre-check is enabled for providers with a cheap, content-stable rev.
+    // WebDAV (the default mock) is deliberately NOT one of them, so exercise the
+    // optimization on Dropbox. Re-create the adapter AFTER switching the id so the
+    // provider key captured by the adapter's closures matches.
+    beforeEach(() => {
+      mockProvider.id = SyncProviderId.Dropbox;
+      adapter = service.createAdapter(mockProvider, mockCfg, mockEncryptKey);
+    });
+
+    it('(a) skips the full download when the remote rev is unchanged (zero downloadFile calls)', async () => {
+      const syncData = createMockSyncData({ syncVersion: 2, recentOps: [] });
+      mockProvider.downloadFile.and.returnValue(
+        Promise.resolve({ dataStr: addPrefix(syncData), rev: 'rev-1' }),
+      );
+
+      // First poll: seeds last-seen rev = 'rev-1' and expected syncVersion = 2.
+      await adapter.downloadOps(0);
+      // Simulate the caller durably applying the ops: this promotes the rev staged
+      // during download to last-seen (same ordering as the seq cursor).
+      await adapter.setLastServerSeq(2);
+
+      crossPollBoundary();
+      mockProvider.downloadFile.calls.reset();
+      // getFileRev reports the SAME rev → remote is byte-identical → nothing new.
+      mockProvider.getFileRev.and.callFake(async (path: string) => {
+        if (path === FILE_BASED_SYNC_CONSTANTS.SYNC_FILE) return { rev: 'rev-1' };
+        throw new RemoteFileNotFoundAPIError('not found');
+      });
+
+      const result = await adapter.downloadOps(2);
+
+      // No full download happened.
+      expect(mockProvider.downloadFile).not.toHaveBeenCalled();
+      // Empty, non-regressing "nothing new" response.
+      expect(result.ops).toEqual([]);
+      expect(result.hasMore).toBe(false);
+      expect(result.latestSeq).toBe(2);
+      expect(result.gapDetected).toBeFalsy();
+    });
+
+    it('(b) proceeds with the full download when the remote rev changed', async () => {
+      const seed = createMockSyncData({ syncVersion: 2, recentOps: [] });
+      mockProvider.downloadFile.and.returnValue(
+        Promise.resolve({ dataStr: addPrefix(seed), rev: 'rev-1' }),
+      );
+      await adapter.downloadOps(0);
+      // Simulate the caller durably applying the ops: this promotes the rev staged
+      // during download to last-seen (same ordering as the seq cursor).
+      await adapter.setLastServerSeq(2);
+
+      crossPollBoundary();
+      mockProvider.downloadFile.calls.reset();
+
+      // Remote rev advanced → must NOT short-circuit.
+      mockProvider.getFileRev.and.callFake(async (path: string) => {
+        if (path === FILE_BASED_SYNC_CONSTANTS.SYNC_FILE) return { rev: 'rev-2' };
+        throw new RemoteFileNotFoundAPIError('not found');
+      });
+      const changed = createMockSyncData({
+        syncVersion: 3,
+        recentOps: [
+          {
+            id: 'op-new',
+            c: 'client2',
+            a: 'HA',
+            o: 'ADD',
+            e: 'TASK',
+            d: 'task-new',
+            v: { client2: 1 },
+            t: Date.now(),
+            s: 1,
+            p: {},
+          },
+        ],
+      });
+      mockProvider.downloadFile.and.returnValue(
+        Promise.resolve({ dataStr: addPrefix(changed), rev: 'rev-2' }),
+      );
+
+      const result = await adapter.downloadOps(2);
+
+      expect(mockProvider.downloadFile).toHaveBeenCalledTimes(1);
+      expect(result.ops.length).toBe(1);
+      expect(result.latestSeq).toBe(3);
+    });
+
+    it('(c) falls through to the full download when getFileRev throws (generic error)', async () => {
+      const seed = createMockSyncData({ syncVersion: 2, recentOps: [] });
+      mockProvider.downloadFile.and.returnValue(
+        Promise.resolve({ dataStr: addPrefix(seed), rev: 'rev-1' }),
+      );
+      await adapter.downloadOps(0);
+      // Simulate the caller durably applying the ops: this promotes the rev staged
+      // during download to last-seen (same ordering as the seq cursor).
+      await adapter.setLastServerSeq(2);
+
+      crossPollBoundary();
+      mockProvider.downloadFile.calls.reset();
+
+      mockProvider.getFileRev.and.callFake(async () => {
+        throw new Error('network blip');
+      });
+
+      const result = await adapter.downloadOps(2);
+
+      // The cheap check failed but the sync did NOT fail — full download ran.
+      expect(mockProvider.downloadFile).toHaveBeenCalledTimes(1);
+      expect(result.latestSeq).toBe(2);
+    });
+
+    it('(c2) falls through to the full download when getFileRev throws RemoteFileNotFoundAPIError', async () => {
+      const seed = createMockSyncData({ syncVersion: 2, recentOps: [] });
+      mockProvider.downloadFile.and.returnValue(
+        Promise.resolve({ dataStr: addPrefix(seed), rev: 'rev-1' }),
+      );
+      await adapter.downloadOps(0);
+      // Simulate the caller durably applying the ops: this promotes the rev staged
+      // during download to last-seen (same ordering as the seq cursor).
+      await adapter.setLastServerSeq(2);
+
+      crossPollBoundary();
+      mockProvider.downloadFile.calls.reset();
+
+      mockProvider.getFileRev.and.callFake(async () => {
+        throw new RemoteFileNotFoundAPIError('gone');
+      });
+
+      const result = await adapter.downloadOps(2);
+
+      expect(mockProvider.downloadFile).toHaveBeenCalledTimes(1);
+      expect(result.latestSeq).toBe(2);
+    });
+
+    it('(d) persists the upload rev so the very next poll short-circuits', async () => {
+      const syncData = createMockSyncData({ syncVersion: 1, recentOps: [] });
+      mockProvider.downloadFile.and.returnValue(
+        Promise.resolve({ dataStr: addPrefix(syncData), rev: 'rev-1' }),
+      );
+      mockProvider.uploadFile.and.returnValue(Promise.resolve({ rev: 'rev-up' }));
+
+      // Poll 1: download (seeds cache + rev) then upload (clears cache, records upload rev).
+      await adapter.downloadOps(0);
+      await adapter.uploadOps([createMockSyncOp()], 'client1');
+
+      // Persisted state carries the upload rev.
+      const persisted = JSON.parse(localStorage.getItem(STATE_KEY) as string);
+      expect(persisted.revs[SyncProviderId.Dropbox]).toBe('rev-up');
+
+      // Poll 2: the cache was cleared by the upload → pre-check is eligible.
+      mockProvider.downloadFile.calls.reset();
+      mockProvider.getFileRev.and.callFake(async (path: string) => {
+        if (path === FILE_BASED_SYNC_CONSTANTS.SYNC_FILE) return { rev: 'rev-up' };
+        throw new RemoteFileNotFoundAPIError('not found');
+      });
+
+      const result = await adapter.downloadOps(2);
+
+      expect(mockProvider.downloadFile).not.toHaveBeenCalled();
+      expect(result.ops).toEqual([]);
+      // expected syncVersion after the upload was 2 → no seq regression.
+      expect(result.latestSeq).toBe(2);
+    });
+
+    it('(e) does NOT skip on the next poll when the ops were never durably applied (no setLastServerSeq)', async () => {
+      // Blocking-1 regression: the download stages the rev as pending; it becomes
+      // the last-seen rev only after the caller confirms the ops were applied
+      // (setLastServerSeq). If processing throws before that, the next poll MUST
+      // re-download rather than short-circuit and skip the un-applied ops forever.
+      const syncData = createMockSyncData({ syncVersion: 2, recentOps: [] });
+      mockProvider.downloadFile.and.returnValue(
+        Promise.resolve({ dataStr: addPrefix(syncData), rev: 'rev-1' }),
+      );
+
+      // Poll 1: download only. Crucially setLastServerSeq is NOT called (simulating
+      // a crash/throw while applying the downloaded ops).
+      await adapter.downloadOps(0);
+
+      crossPollBoundary();
+      mockProvider.downloadFile.calls.reset();
+      // Same rev as poll 1 — a buggy eager-persist would treat this as "nothing new".
+      mockProvider.getFileRev.and.callFake(async (path: string) => {
+        if (path === FILE_BASED_SYNC_CONSTANTS.SYNC_FILE) return { rev: 'rev-1' };
+        throw new RemoteFileNotFoundAPIError('not found');
+      });
+
+      const result = await adapter.downloadOps(2);
+
+      // The rev was never promoted, so there is no last-seen rev to match and the
+      // full download runs — the un-applied ops are re-fetched, not skipped.
+      expect(mockProvider.downloadFile).toHaveBeenCalledTimes(1);
+      expect(result.latestSeq).toBe(2);
+    });
+
+    it('does NOT pre-check for LocalFile (mtime revs cannot guarantee change detection)', async () => {
+      const localMock = jasmine.createSpyObj<FileSyncProvider<SyncProviderId>>(
+        'LocalFileProvider',
+        ['downloadFile', 'uploadFile', 'removeFile', 'getFileRev'],
+      );
+      localMock.id = SyncProviderId.LocalFile;
+      const seed = createMockSyncData({ syncVersion: 2, recentOps: [] });
+      localMock.downloadFile.and.returnValue(
+        Promise.resolve({ dataStr: addPrefix(seed), rev: 'rev-1' }),
+      );
+      localMock.getFileRev.and.callFake(async () => ({ rev: 'rev-1' }));
+
+      const localAdapter = service.createAdapter(localMock, mockCfg, mockEncryptKey);
+      await localAdapter.downloadOps(0);
+
+      (
+        service as unknown as { _syncCycleCache: Map<string, unknown> }
+      )._syncCycleCache.clear();
+      localMock.downloadFile.calls.reset();
+
+      await localAdapter.downloadOps(2);
+
+      // LocalFile is gated out → the full download still runs even though the rev matches.
+      expect(localMock.downloadFile).toHaveBeenCalledTimes(1);
+    });
+
+    it('(f) forceFromSeq0 (sinceSeq=0) still full-downloads and returns snapshotState when the rev is unchanged', async () => {
+      // Review follow-up: a seq-0 download rebuilds local state FROM the remote
+      // snapshot (e.g. USE_REMOTE, which first clears local state), so it must NOT
+      // be short-circuited by the rev pre-check even when remoteRev === lastSeenRev
+      // — otherwise the cleared local state is never repopulated (silent loss).
+      const syncData = createMockSyncData({
+        syncVersion: 2,
+        recentOps: [],
+        state: { tasks: [{ id: 'remote-task' }] },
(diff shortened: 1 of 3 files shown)


johannesjo:
Thanks for this — the `.bak` two-phase write and dropping the unconditional force-overwrite are real robustness wins, and the test coverage on the happy paths is thorough. Before merge, though, I have two blocking concerns plus a red CI run. I also want to flag up front that this branch bundles three independent, individually high-risk sync changes (SPAP-8 atomic writes, SPAP-9 causality gating + `MAX_RECENT_OPS` 500→2000 + auto-merge, SPAP-10 rev pre-check) behind a `perf(...)` title. Per our own sync rules ("every change to the sync system is high-risk … a subtle bug can silently corrupt or lose user data across devices"), I'd really prefer these as three separate PRs so review and rollback stay granular.

## ❌ CI is red

The `Tests` job fails:

```
✗ should invalidate cache after upload
    Expected 0 to be 1.
TOTAL: 1 FAILED, 11764 SUCCESS
```

`src/app/op-log/testing/integration/file-based-sync/edge-cases.integration.spec.ts:246` — the harness provider's id is `SyncProviderId.WebDAV`, so it's in the pre-check allowlist. After client B uploads, its `_lastSeenRevs` holds B's own upload rev; the follow-up `downloadOps(0)` then short-circuits on the unchanged rev and makes **0** `downloadFile` calls where the test asserts **1**. This is SPAP-10 behaving as designed, but the pre-existing test wasn't reconciled. It needs updating (or the behavior reconsidered — see below), and the full unit suite should be green before merge.

## 🔴 Blocking 1 — the last-seen rev is persisted *before* downloaded ops are durably applied → can silently skip ops

In `_downloadOps` (`file-based-sync-adapter.service.ts` ~L877–885) the new code persists the rev eagerly:

```ts
this._expectedSyncVersions.set(providerKey, syncData.syncVersion);
this._lastSeenVectorClocks.set(providerKey, syncData.vectorClock);
this._lastSeenRevs.set(providerKey, rev);
this._persistState();          // NEW — writes revs + syncVersions to localStorage here
```

On `master` this spot only set the in-memory map; there was no `_persistState()` and no pre-check to act on it. The consumer deliberately persists the cursor the *other* way around — `operation-log-sync.service.ts:827-833`:

```ts
// IMPORTANT: Persist lastServerSeq AFTER ops are stored in IndexedDB.
// … If we crash before this point, lastServerSeq won't be updated, and the client
// will re-download the ops on next sync. This is the correct behavior - better to
// re-download than to skip ops.
```

SPAP-10 inverts that guarantee. Failure sequence:

1. Poll downloads the file (rev `R`, syncVersion `N`); L885 persists `revs[provider]=R` and `expectedVersion=N`.
2. `processRemoteOps` throws (bad op / decrypt error on a later page / conflict throw) **or** the app is killed before `setLastServerSeq` at L832. The cursor and `appliedOpIds` stay behind; rev `R` is already on disk.
3. Next poll (or after restart, once the 30 s `_CACHE_TTL_MS` cache is gone): the pre-check calls `getFileRev`, sees `remoteRev === R`, returns `{ ops: [], hasMore: false, latestSeq: expectedVersion }`.
4. The download loop treats empty ops as "done" and persists `lastServerSeq = latestSeq = N`. **The un-applied ops are now permanently skipped** until the remote rev changes again.

With `MAX_RECENT_OPS` raised to 2000 and `DOWNLOAD_PAGE_SIZE` 500, catch-up is up to 4 pages, widening the window (a throw on page 3 leaves page-1's persisted rev ahead of an empty IndexedDB write). Suggested fix: persist the last-seen rev with the *same* ordering as the cursor — only after the caller confirms ops are durably applied — rather than eagerly inside `_downloadOps`. Keeping the in-memory `set` for the intra-cycle case is fine; it's the eager `_persistState()` of the rev that breaks the invariant. This path also has no regression test.

## 🔴 Blocking 2 — for WebDAV the pre-check saves no bandwidth and *doubles* it on changed polls

The premise is "call the cheap `getFileRev` without downloading the full file." That's true for Dropbox (`dropbox.ts:127` → `files/get_metadata`, metadata only), but **not** for WebDAV, which is also in the allowlist. `webdav-base-provider.ts:123-129`:

```ts
async getFileRev(targetPath, _localRev) {
  const r = await this.downloadFile(targetPath);   // full GET of the body
  return { rev: r.rev };
}
```

(No subclass overrides it — plain `webdav.ts` and `nextcloud.ts` inherit this; only `webdav` is in the allowlist.) So for WebDAV:

- **Unchanged remote:** full GET, body discarded → same one download as today. Zero saving.
- **Changed remote:** full GET (rev differs) → fall through → `_downloadSyncFile` downloads the same file **again** → two full downloads.

The description says "WebDAV ETag," but the code doesn't do a cheap `PROPFIND`/`HEAD`. Either implement `getFileRev` as an ETag-only PROPFIND, or drop `WebDAV` from `_REV_PRECHECK_PROVIDERS` until it exists — as-is it's a net pessimization for one of the two enabled providers.

## 🟠 Worth addressing

- **WebDAV rev soundness (weak ETags).** "Unchanged rev ⇒ byte-identical" needs strong ETags. `LocalFile` is (correctly) excluded for coarse mtime, but plenty of WebDAV servers emit weak or second-granularity ETags; two remote writes in the same second could collide and the pre-check would skip a real update. Compounds Blocking 2.
- **Force-overwrite removal can strand uploads.** On "same rev after re-download," the loop now retries the *conditional* upload with the same `revToMatch` (max 2) and then throws retryably — forever. For a WebDAV server that reports a stable rev yet rejects conditional writes (the exact case the old force-upload handled), local changes can never push. The data-safety trade is defensible, but such a client should surface a user-visible error rather than loop silently.
- **`AUTO_MERGE_CONCURRENT_SNAPSHOT` defaults ON.** `_tryConcurrentSnapshotMerge` routes a CONCURRENT seq-0 snapshot's `newOps` through the LWW pipeline for *all* file-based users by default. The adjacent existing comment (`operation-log-sync.service.ts` ~L455-469) documents the frontier-regression hazard of appending historical remote ops. Given our "calm default" posture and how hard merge mistakes are to reverse, I'd ship this off by default (or behind field validation) and in its own PR.

## 👍 Good
- Two-phase `.bak` + recover-from-backup, non-fatal by design and well tested.
- Removing the unconditional force-overwrite is the right direction for lost-update safety.
- The `partialTrimGap` off-by-one fix (`> sinceSeq + 1`) is correct and only reduces false gaps.

**Recommendation:** request changes. Fix the red test, address Blocking 1 (data-loss ordering) and Blocking 2 (WebDAV `getFileRev`), and please split SPAP-8/9/10 apart. Happy to pair on the crash-safety ordering or a WebDAV PROPFIND if useful.

johannesjo:
## Review — bundles 3 sync changes; one has a **confirmed** silent data-loss path

Thanks for this. Heads up that although the PR is titled/described as SPAP-10 (rev-precheck), the diff actually lands **three independent, high-risk sync features**: SPAP-8 (atomic writes), SPAP-9 (causality gating), and SPAP-10 (rev-precheck). I reviewed all three. The SPAP-10 logic itself is *correct* (an unchanged content rev genuinely proves nothing new) — but there's a **confirmed correctness bug in the bundled SPAP-9**, and SPAP-10's perf premise doesn't hold for WebDAV. Recommend splitting and addressing the items below before merge.

### 🔴 HIGH (correctness) — SPAP-9 cosmetic-reset silently drops ops a behind client never saw

`file-based-sync-adapter.service.ts` (~L803-823) suppresses `versionWasReset` when the remote clock is `EQUAL`/`GREATER_THAN` last-seen. But `GREATER_THAN` only means "the writer did more work" — **not** that the receiving client already holds the intervening ops. When a snapshot (migration/recovery/initial) compacts ops a behind client never downloaded into `state` and resets `recentOps`, then the writer makes ≥1 more op:

- `versionWasReset` → suppressed (dominating clock)
- `snapshotReplacement` → can't fire (`recentOps.length !== 0`)
- `partialTrimGap` → can't fire (buffer not full; post-snapshot op has low `sv`)
- the download service only **warns** on the seq decrease (`operation-log-download.service.ts:310`), no resync

→ `gapDetected=false`, the snapshot is never re-hydrated, and the compacted ops are **silently lost**. Pre-SPAP-9 this forced a seq-0 resync and recovered them.

This is confirmed with a failing test (drop it into the `benign syncVersion reset gating (SPAP-9)` describe — it goes red against this branch):

```ts
it('flags a gap when a dominating-clock reset compacted ops the client had not yet seen', async () => {
  // First download: client synced up to syncVersion 5 (cursor -> 5, last-seen {client1:5}).
  const first = createMockSyncData({
    syncVersion: 5, vectorClock: { client1: 5 }, recentOps: [opAt(5)],
  });
  mockProvider.downloadFile.and.returnValue(
    Promise.resolve({ dataStr: addPrefix(first), rev: 'rev-1' }),
  );
  await adapter.downloadOps(1);

  // Writer made ops 6..10 (client never saw them), took a snapshot compacting 1..10 into
  // `state` and reset recentOps, then made op 11.
  const second = createMockSyncData({
    syncVersion: 2,               // reset to 1 by snapshot, +1 for op 11
    vectorClock: { client1: 11 }, // GREATER_THAN {client1:5} — writer strictly ahead
    recentOps: [opAt(2)],         // only op 11 survived; ops 6..10 compacted into `state`
    oldestOpSyncVersion: 2,
    state: { tasks: [] },
  });
  mockProvider.downloadFile.and.returnValue(
    Promise.resolve({ dataStr: addPrefix(second), rev: 'rev-2' }),
  );

  const result = await adapter.downloadOps(5); // client's real cursor is 5
  expect(result.gapDetected).toBe(true);       // ← currently returns false
});
```

Suggested fix: treat a regression as cosmetic **only on `EQUAL`** (identical causal state ⇒ truly nothing missed), not on `GREATER_THAN`.

Related: `AUTO_MERGE_CONCURRENT_SNAPSHOT` (default **on**) — `_tryConcurrentSnapshotMerge` (`operation-log-sync.service.ts:993`) applies only `result.newOps` and never hydrates `snapshotState`, so a CONCURRENT snapshot whose compacted base holds an op the client lacks loses it the same way — silently, where previously the user got a conflict dialog. I'd ship this **default-off** until validated with a multi-client E2E.

### 🔴 HIGH (perf) — SPAP-10's pre-check allowlist is backwards for WebDAV

`_REV_PRECHECK_PROVIDERS = [Dropbox, WebDAV]` (`:131`), but WebDAV's `getFileRev` is **not** a cheap ETag call — it does a full-body GET and md5-hashes it (`webdav-base-provider.ts:123-129` → `downloadFile` → GET in `webdav-api.ts`). So for WebDAV the "cheap pre-check" downloads the whole file: **no bandwidth saved** when unchanged, and **two full downloads** when the rev changed. Only Dropbox has a genuine cheap metadata call; ironically **OneDrive** (the one file provider whose `getFileRev` uses a real ETag) is excluded. The tests don't catch this because they mock `getFileRev` as cheap (decoupled from `downloadFile`).

Fix: allowlist `[Dropbox, OneDrive]` (or give WebDAV a real cheap rev via PROPFIND `getetag` / conditional GET first), and correct the comments/PR body ("WebDAV: strong ETag" / "LocalFile coarse mtime" are inaccurate — all three use an md5 content hash).

### 🟡 MEDIUM

- **Backup-before-overwrite doubles upload cost every cycle** (`:636-638`): `_backupCurrentRemote` re-encodes and re-uploads the full current state to `.bak` on every op-bearing sync — undocumented write amplification that works against SPAP-10's bandwidth goal (on WebDAV a single sync becomes GET + full PUT `.bak` + verify-GET + full PUT primary + verify-GET). Worth a size threshold or server-side copy where supported.
- **Backup recovery is read-only and re-snacks every poll**: a client with no pending ops re-recovers the corrupt primary and re-shows `SYNC_DATA_RECOVERED_FROM_BACKUP` on each cycle, and never heals the remote. Consider healing the primary once + de-duping the notice.

### 🟢 Looks good

Removing the unconditional `isForceOverwrite=true` on the rev-match retry is a real lost-update fix; the rev/version pairing is consistent so SPAP-10 won't regress the seq for Dropbox; `compareVectorClocks` direction is used correctly in both new sites; the `partialTrimGap` off-by-one fix is right.

### Suggested path

Split into SPAP-8 / SPAP-9 / SPAP-10. For SPAP-9, narrow cosmetic-reset to `EQUAL` + default the auto-merge off + add the test above. For SPAP-10, fix the allowlist. For SPAP-8, document the write cost + quiet the recovery snack.

<sub>🤖 Reviewed with Claude Code</sub>

aakhter:
Thanks — this is split down to **only SPAP-10** now (rev-precheck) on current master; SPAP-8 landed via #8786 and SPAP-9 is its own PR (#8787), so neither is in this diff anymore. Both blockers addressed:

**🔴 Blocking 1 — last-seen rev persisted before ops durably applied — fixed.**
The eager `_persistState()` inside the download path is gone. `_lastSeenRevs` is still `set()` in memory for the intra-cycle case, but it's now only *persisted* at the cursor-commit point (after the caller has durably applied the ops), matching the `setLastServerSeq`/`lastServerSeq` ordering you pointed to. A throw or crash between download and apply no longer strands a persisted rev ahead of an empty IndexedDB, so the pre-check can't skip un-applied ops. Added a regression test that crashes processing mid-cycle and asserts the next poll still re-downloads.

**🔴 Blocking 2 — WebDAV pre-check is a pessimization — fixed.**
Corrected the allowlist to `[Dropbox, OneDrive]` — both have genuine cheap metadata/ETag rev calls. WebDAV (full-body GET in `getFileRev`) and LocalFile are excluded, so there's no "download twice on change / zero saving on no-change" case anymore. Fixed the inaccurate comments too ("WebDAV: strong ETag" / "LocalFile: coarse mtime" — all three actually md5 the content). This also makes the "should invalidate cache after upload" test pass on the WebDAV harness again (it no longer short-circuits); the LocalFile isolation commit stays as extra, explicit coverage.

**🟠 WebDAV weak-ETag soundness** (if WebDAV is ever re-added to the allowlist, gate on a strong validator only) — filed as **SPAP-29**.

johannesjo:
## Review — SPAP-10 rev-precheck (86210f8)

Re-reviewed after the split down to SPAP-10-only. The two prior blockers are correctly addressed 👍:

- **Pending-rev staging** is sound — the rev is staged in `_pendingRevs` and only promoted to `_lastSeenRevs` in `setLastServerSeq` (after durable apply); the crash-safety test (e) proves it.
- **Allowlist** is now `[Dropbox, OneDrive]` with accurate comments. Verified Dropbox `rev` is consistent across `getFileRev`/`downloadFile`/`uploadFile`, `getMetaData` ignores the `localRev` hint (no stale-cache short-circuit), and the upload mismatch-fallback **throws** on genuine concurrency *before* `_lastSeenRevs` is set — so it can't strand unapplied concurrent ops behind a matching rev.

One blocking issue remains, plus a minor OneDrive note.

---

### 🔴 Blocking — the pre-check ignores `sinceSeq`, so it silently defeats forced (`seq-0`) downloads

The pre-check gates on `lastSeenRev && !cache && provider∈allowlist` but not on `sinceSeq`:

```ts
// _downloadOps
const lastSeenRev = this._lastSeenRevs.get(providerKey);
if (
  lastSeenRev &&
  !this._getCachedSyncData(providerKey) &&
  this._REV_PRECHECK_PROVIDERS.has(provider.id)
) {
  ...
  if (remoteRev === lastSeenRev) {
    return { ops: [], hasMore: false, latestSeq: expectedVersion }; // no snapshotState / snapshotVectorClock
  }
}
```

A `downloadOps(0)` (forced/fresh download) is therefore short-circuited to an empty response — **no `snapshotState`, no `snapshotVectorClock`** — whenever the remote rev is unchanged.

**Concrete data-integrity failure via `forceDownloadRemoteState()` (the "Use Server Data" conflict choice):**

1. `captureImportBackup()` → `clearFullStateOps()` → `clearUnsyncedOps()` — local unsynced ops are wiped.
2. `setLastServerSeq(0)` — this **promotes the pending rev** staged by the conflict-detecting download, so `_lastSeenRevs === currentRemoteRev`.
3. `downloadRemoteOps({ forceFromSeq0: true })` → `downloadOps(0)` → **pre-check short-circuits** (rev matches).
4. Back in `forceDownloadRemoteState`: `result.snapshotVectorClock` is undefined → vector clock **not** reset; `result.snapshotState` undefined and `result.newOps.length === 0` → the entire hydrate / `loadAllData` reset / `setLastServerSeq` block is skipped.

The user explicitly asked to replace local with server data, their unsynced ops are already gone, but **local state is not replaced, the vector clock is left stale, and `lastServerSeq` stays 0** — while the `_showRestorePreviousDataSnack()` still fires, implying success. Because `lastServerSeq` stays 0 and `_lastSeenRevs` stays at the matching rev, subsequent `downloadOps(0)` polls keep short-circuiting, so it does not self-heal until another client writes. (The #8107 backup makes it Undo-recoverable, but this is still a silent no-op of an explicit destructive action in the sync path.) The documented "rebuild clock state" purpose of `forceFromSeq0`, and the rejected-ops handler's `forceFromSeq0` download, are defeated the same way in steady state.

**Red repro** (drop into the `remote-rev pre-check (SPAP-10)` describe — 82 others stay green):

```ts
it('forceFromSeq0 (downloadOps(0)) must NOT be short-circuited by the rev pre-check', async () => {
  const syncData = createMockSyncData({ syncVersion: 2, recentOps: [], state: { tasks: [] } });
  mockProvider.downloadFile.and.returnValue(
    Promise.resolve({ dataStr: addPrefix(syncData), rev: 'rev-1' }),
  );
  await adapter.downloadOps(0);
  await adapter.setLastServerSeq(0); // mirrors forceDownloadRemoteState's reset → promotes the pending rev
  crossPollBoundary();
  mockProvider.downloadFile.calls.reset();
  mockProvider.getFileRev.and.callFake(async () => ({ rev: 'rev-1' })); // remote unchanged
  const result = await adapter.downloadOps(0); // USE_REMOTE force download
  expect(mockProvider.downloadFile).toHaveBeenCalledTimes(1); // ✗ was 0
  expect(result.snapshotState).toBeDefined();                 // ✗ undefined
  expect(result.snapshotVectorClock).toBeDefined();           // ✗ undefined
});
```

**Fix (verified — all 83 tests green):** exempt seq-0 downloads. Normal polls always pass `sinceSeq = syncVersion > 0`, so the optimization is unaffected; a fresh client's first `downloadOps(0)` has no `lastSeenRev` yet, so nothing is lost.

```ts
if (
  sinceSeq > 0 &&                       // never short-circuit a forced/fresh (seq-0) download
  lastSeenRev &&
  !this._getCachedSyncData(providerKey) &&
  this._REV_PRECHECK_PROVIDERS.has(provider.id)
) {
```

Worth keeping a regression test asserting `downloadOps(0)` always full-downloads even with a matching `lastSeenRev`.

---

### 🟡 Minor (non-blocking, please double-check) — OneDrive rev namespace

`OneDriveProvider.getFileRev` returns the driveItem metadata `eTag`, but `downloadFile` returns `_getResponseETag(response)` — the `/content` HTTP `ETag`, which for OneDrive is usually the redirected blob-storage validator, not the Graph `eTag`. If so, a `_lastSeenRevs` seeded from a **download** would rarely match a later `getFileRev`, making the pre-check a no-op on OneDrive's download-seeded path (it still works after uploads, where both come from the Graph `eTag`). This is the safe direction (an unchanged `eTag` still implies unchanged content, so no false skip) — just flagging that OneDrive's win may be smaller than intended and worth confirming. Dropbox is fully consistent.

johannesjo:
### SPAP-10 correctness: the rev-precheck can defeat a `forceFromSeq0` full download (USE_REMOTE), after local state has already been cleared

The `getFileRev` short-circuit in `_downloadOps` is gated on `lastSeenRev`, the sync-cycle cache, and the provider allowlist — but **not on `sinceSeq`**. It returns `{ ops: [], hasMore: false }` with **no `snapshotState`** whenever `remoteRev === lastSeenRev`.

That invariant ("unchanged rev ⇒ nothing to do") holds for incremental polls (`sinceSeq > 0`), where local is already consistent with `lastSeenRev`. It does **not** hold for `sinceSeq === 0` / `forceFromSeq0`, whose whole purpose is to rebuild local state *from* the remote snapshot — there, "remote unchanged" does not mean "nothing to do", because the thing being rebuilt is the *local* state. Note the full-download path deliberately returns `snapshotState` only when `isForceFromZero` — the precheck skips exactly the response the caller needs.

#### Why it's reachable (USE_REMOTE conflict resolution)

`_handleUseRemote` (`operation-log-sync.service.ts`) does, in order:

1. `clearFullStateOps()` + `clearUnsyncedOps()` — local state is discarded.
2. `setLastServerSeq(0)` — which, per SPAP-10, **promotes the pending rev to `_lastSeenRevs`**. That pending rev was staged as `R_new` during the conflict-detecting download, so now `_lastSeenRevs === R_new` (the current remote rev).
3. `downloadRemoteOps({ forceFromSeq0: true })` → `downloadOps(sinceSeq = 0)` → `_downloadOps`.

At step 3 the precheck sees `lastSeenRev === remoteRev === R_new`, the `_syncCycleCache` has expired (`_CACHE_TTL_MS = 30_000`; a conflict dialog easily sits open longer than that while the user decides), and Dropbox/OneDrive are in the allowlist — so it returns empty with no snapshot. **Net: local task/project data was cleared in step 1, and the remote state it was meant to be replaced with is never downloaded.** The pre-replace safety backup is the only thing preventing permanent loss; the sync itself lands the client in a wiped/inconsistent state.

#### Repro (Dropbox or OneDrive)

1. Client A synced to rev `R_old`.
2. Client B uploads concurrent ops → remote is now `R_new`.
3. Client A polls → full download (stages `_pendingRevs = R_new`), detects a concurrent-clock conflict, shows the dialog.
4. Wait > 30s (cache TTL), then choose **USE_REMOTE**.
5. Observed: `_downloadOps` logs *"Remote rev unchanged — skipping full download"*, returns `{ ops: [], snapshotState: undefined }`; local state was already cleared → data loss (recoverable only via the import backup).

Fast-clicking (< 30s) masks it, because the still-warm `_syncCycleCache` bypasses the precheck (`!this._getCachedSyncData(...)` is false). That's why it wouldn't show up in a quick manual test.

#### Fix

Gate the precheck on `sinceSeq > 0` so forced full downloads always fetch the snapshot:

```ts
if (
  sinceSeq > 0 &&
  lastSeenRev &&
  !this._getCachedSyncData(providerKey) &&
  this._REV_PRECHECK_PROVIDERS.has(provider.id)
) {
```

A `forceFromSeq0` download (`sinceSeq === 0`) must never short-circuit — its result is used to **replace** local state, so it has to carry the full `snapshotState` even when the remote rev is unchanged.

Suggested regression test in the SPAP-10 spec: `forceFromSeq0` with `remoteRev === lastSeenRev` **and an empty cache** must still perform the full download and return `snapshotState` (the existing crash-safety test seeds a fresh cache, so it doesn't exercise this).

#### Also worth checking

`rejected-ops-handler.service.ts:338` triggers the same `forceFromSeq0` path. The `sinceSeq > 0` guard covers it too, but the rev/cache state at that call site is worth a look to confirm it isn't independently affected.

<sub>Confidence ~85%. The interaction (`setLastServerSeq(0)` promoting the pending rev, then the precheck matching it) is the load-bearing part; the trigger requires the cache to have expired, which a deliberated USE_REMOTE dialog makes likely but not certain.</sub>

aakhter:
Good catch — fixed. The `forceFromSeq0` short-circuit was real: `_downloadOps` only gated the precheck on `lastSeenRev` + cache + allowlist, so a `sinceSeq === 0` download (USE_REMOTE, after local state is cleared) could return `{ops: [], snapshotState: undefined}` when the rev was unchanged.

Added `sinceSeq > 0 &&` to the precheck condition, so a forced full download always fetches the snapshot even when the rev matches:

```ts
if (
  sinceSeq > 0 &&
  lastSeenRev &&
  !this._getCachedSyncData(providerKey) &&
  this._REV_PRECHECK_PROVIDERS.has(provider.id)
) {
```

That also covers the `rejected-ops-handler.service.ts` forceFromSeq0 call site (all forced downloads use `sinceSeq === 0`). Added test (f): `forceFromSeq0` with `remoteRev === lastSeenRev` **and an empty cache** still performs the full download and returns `snapshotState` (the crash-safety test (e) seeds a fresh cache, so it didn't exercise this path). Full op-log suite green (2878).

Rebased onto current master (which now has the merged SPAP-9), so this is a clean single commit again.

johannesjo:
Thank you very much for this ! <3

PR #8801, opened 2026-07-06
fix(tasks): navigate to Today for context-less tasks from search (#8780)

### LINKED ISSUE #8780: 🚨 Search Problem
### Version Used

### Describe the Bug
Search itself works and returns matching results. However, clicking on a search result does not open/navigate to the corresponding task. Nothing happens (or the wrong view is shown) when a result is tapped.

### Additional Context
This occurs consistently, with sync disabled / purely local as well as with sync enabled.

### Expected Behavior

Clicking a search result should navigate to and display the corresponding task.

### Steps to Reproduce

1. Open the app
2. Tap the search icon / open search
3. Type a query that matches the title of an existing task
4. Matching results are shown
5. Tap on one of the results
6. The corresponding task is not opened / displayed

### Console Output

```shell

```

### Documentation

johannesjo:
## Problem

Fixes #8780 — tapping a global-search result does nothing for certain tasks.

`NavigateToTaskService._getLocation()` returned an empty string `''` for a **non-archived task with no project, no tags, and not due today** (most commonly an **overdue task in the flat Today/Inbox list**). Since `devError()` only logs in production, `''` flowed into `navigate()`:

```ts
if (this._router.url.startsWith(location)) {   // '/search'.startsWith('') === true, ALWAYS
  this._focusTaskElement(taskId);              // no-op: task element isn't on the /search page
  return;                                       // ← returns without navigating
}
```

`String.startsWith('')` is always true, so the task was wrongly treated as "already in the right context" and navigation was silently swallowed — exactly the reported symptom.

> Note: this is **not actually Android-specific** in the code; the same task shape fails on desktop too. It surfaces as "search is broken on mobile" because that reporter's tasks hit the empty-location path. Tasks with a project or tag navigate fine everywhere.

## Fix

- `_getLocation`: a non-archived task with no project/tag now resolves to `/tag/TODAY/tasks` (where overdue/context-less tasks are shown) instead of `''`.
- `navigate()`: guard `if (!location) throw` so an unresolved location surfaces the existing error snack instead of a silent no-op (covers the remaining archive edge case).

All 7 callers of this method (search, reminder/notification effects, issue import, calendar integration, play-button) share the fix — none relied on the broken behavior.

## Tests

New `navigate-to-task.service.spec.ts` (3 tests): reproduces the bug (fails before the fix, passes after) and locks in project/tag routing. Green via `npm run test:file`.

### DIFF
--- a/src/app/core-ui/navigate-to-task/navigate-to-task.service.spec.ts
+++ b/src/app/core-ui/navigate-to-task/navigate-to-task.service.spec.ts
@@ -1,0 +1,111 @@
+import { TestBed } from '@angular/core/testing';
+import { Router } from '@angular/router';
+import { NavigateToTaskService } from './navigate-to-task.service';
+import { TaskService } from '../../features/tasks/task.service';
+import { ProjectService } from '../../features/project/project.service';
+import { SnackService } from '../../core/snack/snack.service';
+import { DateService } from '../../core/date/date.service';
+import { LayoutService } from '../layout/layout.service';
+import { Task } from '../../features/tasks/task.model';
+import { TODAY_TAG } from '../../features/tag/tag.const';
+import { of } from 'rxjs';
+
+const TODAY_STR = '2026-07-06';
+
+const createTask = (partial: Partial<Task>): Task =>
+  ({
+    id: 'task-1',
+    title: 'A task',
+    projectId: undefined,
+    parentId: undefined,
+    tagIds: [],
+    dueDay: undefined,
+    dueWithTime: undefined,
+    timeSpentOnDay: {},
+    created: Date.parse('2020-01-01'),
+    ...partial,
+  }) as unknown as Task;
+
+describe('NavigateToTaskService', () => {
+  let service: NavigateToTaskService;
+  let taskService: jasmine.SpyObj<TaskService>;
+  let router: jasmine.SpyObj<Router> & { url: string };
+  let layoutService: jasmine.SpyObj<LayoutService>;
+
+  beforeEach(() => {
+    taskService = jasmine.createSpyObj('TaskService', [
+      'getByIdFromEverywhere',
+      'getArchivedTasks',
+    ]);
+    const projectService = jasmine.createSpyObj('ProjectService', [], {
+      list$: of([]),
+    });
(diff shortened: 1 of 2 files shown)


Sanja2012:
The Search Problem is still existing in SP 18.14.0 on Android. Search request is showing matching tasks, but clicking on them does not lead to the  according task.

johannesjo:
Post-merge review note: the empty-location guard is a sound defensive fix, but the regression test does not demonstrate that this was the production path behind #8780.

The test factory sets `projectId: undefined` and then bypasses the model with `as unknown as Task` ([test fixture](https://github.com/super-productivity/super-productivity/blob/7130d617650f2b92a574b43017e88ea6a11787c1/src/app/core-ui/navigate-to-task/navigate-to-task.service.spec.ts#L15-L27)). In supported state, `Task.projectId` is required; new tasks receive the default/Inbox project, legacy migration assigns Inbox to project-less tasks, and validation repair backfills a missing project ID. That means the new Today fallback primarily covers invalid, legacy, or transiently corrupted data. It may still be useful defense-in-depth, but this test does not establish that the reporter's Android task hit that branch.

The reporter has now confirmed that the same symptom remains in 18.14.0, whose tag contains this merge: https://github.com/super-productivity/super-productivity/pull/8801#issuecomment-4955766726

I suggest reopening #8780 or creating a follow-up and reproducing the complete search-result tap with a valid task (for example `projectId: INBOX_PROJECT.id`) through `SearchPageComponent` on Android. The existing `recordSearchNavDebug` events should show whether the failure occurs before `NavigateToTaskService.navigate()`, during routing, or while focusing the rendered task. The empty-location guard can remain as defensive handling.

Sanja2012:
I have tested the app further. The search error occurs only in tasks that are not assigned to a project (on Android and Windows App, v. 18.14.0).

Sanja2012:
Das Suchproblem existiert weiterhin in SP 18.16.0 auf Android und Windows. Der Suchfehler tritt bei Aufgaben auf, die nicht einem Projekt zugewiesen sind. Leider ist die App damit für mich unbrauchbar.

PR #8802, opened 2026-07-06
feat(sync): SPAP-11 opt-in split-file sync format for O(delta) syncs

aakhter:
## SPAP-11 — opt-in split-file ("Surgical sync") format

Splits the single `sync-data.json` into a small always-read/written `sync-ops.json` (the commit point) plus a `sync-state.json` snapshot rewritten only on compaction / force-upload / gap-repair / migration, so a normal sync transfers O(delta) instead of O(dataset). Opt-in via the `isUseSplitSyncFiles` sync setting; the legacy file is left as a v3 tombstone so old clients hard-stop instead of silently diverging.

### Status
Now that SPAP-9 (#8787) and SPAP-10 (#8795) have merged, this is rebased onto current master as a **single, standalone commit** — just the split-format feature.

### Addresses review
- **Compaction re-fired on every op-bearing sync past 1000 — fixed.** Recompaction triggers at `combinedOps > MAX_RECENT_OPS` (2000) and trims to `SPLIT_COMPACTION_THRESHOLD` (1000), leaving ~1000 cheap op-only syncs between compactions (test b2).
- **Split download committed the rev eagerly — fixed.** It now stages to `_pendingRevs`, promoted only in `setLastServerSeq` after durable apply (mirrors the single-file crash-safety path; test g).
- **Split rev-precheck now gated on `sinceSeq > 0`**, so a `forceFromSeq0` download (USE_REMOTE / rehydrate) always fetches ops + snapshot (test h).

### Follow-ups (filed, not bundled)
- **SPAP-31** — opt-in is one-way: block un-check once migrated + friendly old/downgraded-client error + document the no-downgrade contract.
- **SPAP-32** — version-addressed snapshot filename so concurrent compaction can't clobber a referenced snapshot.
- **SPAP-33** — mixed-fleet `MAX_RECENT_OPS` safety, plus dropping the redundant `partialTrimGap` length clause (false-negative on legacy-trimmed buffers).
- **SPAP-34** — multi-client E2E for the auto-merge / silent auto-resolution paths.


johannesjo:
## Review — split-file sync format

Reviewed the full diff, reading the final versions of the three files carrying the logic (`file-based-sync-adapter.service.ts`, `operation-log-sync.service.ts`, the constants/config), plus the underlying provider `getFileRev`/`downloadFile` implementations. I did not run the suites locally. Per our rule that every sync change is high-risk, I focused on replay determinism, concurrent/remote edits, and hard-to-reverse format choices.

Nice work overall — the write-ordering (state → ops commit-point → tombstone), `snapshotRef` validation + `.bak` fallback, best-effort/non-fatal optimization paths, and the comments are all thoughtful. But there are two concrete defects that undercut the perf goals, plus some design costs worth settling before merge.

---

### 🔴 High — compaction re-triggers on *every* sync once a folder hits 1000 ops (defeats the feature's purpose)

`_uploadOpsSplit` (`file-based-sync-adapter.service.ts:1659-1688`):

```ts
const needsCompaction =
  !snapshotRef ||
  combinedOps.length > FILE_BASED_SYNC_CONSTANTS.SPLIT_COMPACTION_THRESHOLD; // > 1000
...
finalOps = combinedOps.slice(-FILE_BASED_SYNC_CONSTANTS.SPLIT_COMPACTION_THRESHOLD); // keeps 1000
```

The compaction **trigger** (`> 1000`) and the post-compaction **retained size** (`slice(-1000)`) are the same value. So:

1. Folder reaches 1000 recent ops (from a prior compaction that trimmed to exactly 1000).
2. Client appends ≥1 op → `combinedOps.length = 1001 > 1000` → **compaction**: rebuilds the full store snapshot (`getStateSnapshot()` + both archives), `_backupStateFile` (downloads *and* re-uploads the old `sync-state.json`), writes the new `sync-state.json`, writes the ops file. Trims back to 1000.
3. Next op-bearing sync: downloads ops (1000), appends → 1001 → **compaction again.**

Once a folder crosses 1000 ops — which any active user reaches — **every sync that carries ops rebuilds and re-uploads the entire snapshot + both archives**, plus an extra full state download+upload for the `.bak`. That's ~2–3× the data volume of the current single-file v2 path, i.e. worse than the problem the PR sets out to solve. The cheap op-only branch (`else`, trim to `MAX_RECENT_OPS`=2000) is effectively unreachable in steady state because compaction fires first at 1000.

Likely intended fix: trigger on `> MAX_RECENT_OPS` (2000) and keep trimming to `SPLIT_COMPACTION_THRESHOLD` (1000), so there are ~1000 cheap op-only syncs between compactions. The header comment ("Compaction threshold ≈ MAX_RECENT_OPS/2") reads like `SPLIT_COMPACTION_THRESHOLD` was meant as the trim target, not the trigger.

Not caught by tests — the compaction spec only asserts the *first* compaction fires. Please add a test doing two op-bearing syncs after crossing the threshold and asserting `getStateSnapshot` is called at most once.

### 🔴 High — the SPAP-10 rev-precheck allowlist has WebDAV and OneDrive backwards

`_REV_PRECHECK_PROVIDERS = new Set([Dropbox, WebDAV])` (`file-based-sync-adapter.service.ts:136`). The code comment claims *"WebDAV: strong ETag,"* but the actual implementations say otherwise:

- **WebDAV** `getFileRev` → `downloadFile` → `_api.download` fetches the **full body** (`webdav-base-provider.ts:123-128`, `:162-173`). So the "cheap pre-check" transfers the whole file. On any poll where the file changed it then does the real `_downloadSyncFile` too → **two full downloads**. On an unchanged poll it downloads the file and discards the body. It never saves bandwidth and usually costs extra — a pure pessimization for one of our most common self-hosted providers.
- **OneDrive** `getFileRev` is a genuinely cheap eTag metadata call (`onedrive.ts:167-177`) — the one provider besides Dropbox that would actually benefit — but it's **excluded**.
- Dropbox (cheap `getMetaData` ✓) and LocalFile (full read, correctly excluded ✓) are right.

The allowlist should be `[Dropbox, OneDrive]`. (Same root cause worth remembering: WebDAV/LocalFile `getFileRev` is a full GET, not a cheap ETag.)

---

### 🟠 Medium — the opt-in setting is effectively one-way; turning it back **off** self-bricks sync

Migration is irreversible per folder (v3 tombstone over `sync-data.json`, no un-migrate path). The checkbox doesn't convey that:

- **Un-checking the setting after enabling** → the now-OFF client hits its own tombstone in `_downloadSyncFile` → `SplitSyncFormatDetectedError` → sync pauses with "enable Surgical sync." A normal-looking checkbox that breaks sync when you uncheck it.
- **Devices on older app versions** (predating this feature) hit `version !== 2` and throw `SyncDataCorruptedError`, which surfaces as a *corruption* dialog, not the friendly "enable Surgical sync" notice (only clients shipping `_isSplitTombstone` recognize the tombstone). That's an alarming, unexplained error on device 2/3.
- The "truly-old client aborts, never overwrites" claim holds on the download path, but please confirm it for an explicit user-triggered **force-upload / "Use Local"** on the currently-released version — if any such path writes a v2 `sync-data.json` without a version pre-check, it overwrites the tombstone and causes silent divergence (old client reads `sync-data.json`, new client reads `sync-ops.json`).

Suggest blocking un-check once migrated (or a confirm that explains the consequence) and documenting that there's no downgrade path.

### 🟠 Medium — concurrent compaction can leave `sync-state.json` not matching the committed ops file

Compaction force-overwrites `sync-state.json` *before* the conditional ops commit. If two clients compact in the same cycle, both force-write their own `sync-state.json` (last writer wins) but only one wins the ops PUT. The committed ops file's `snapshotRef` can then point at a snapshot the losing client clobbered (same `syncVersion`, different `vectorClock`). Recovery then hinges on `.bak` timing; if `.bak` doesn't hold the matching snapshot, `_loadValidatedSnapshot` returns `null` → `gapDetected` with **no `snapshotState`**, and a seq-0 bootstrapping client gets stuck re-gapping until someone next compacts or force-uploads. Low probability, but finding #1 makes compaction fire far more often than intended, which raises the odds. Worth a two-client integration test; a more robust design writes the snapshot under an immutable rev/version-addressed name so `snapshotRef` can't be clobbered.

---

### 🟡 Lower / worth noting

- **Scope.** This is really four independent high-risk changes (SPAP-8 atomic writes, SPAP-9 causality gating + `MAX_RECENT_OPS` bump, SPAP-10 rev-precheck, SPAP-11 split format) landed together into our highest-risk subsystem — and SPAP-8/9/10 change the **default path for all file-based users**, not just the opt-in. Please consider splitting so the shared-path changes can be reviewed/bisected/reverted independently of the opt-in format.
- **`MAX_RECENT_OPS` 500→2000 is not opt-in** — it affects every file-based user's single-file uploads (+~0.3 MB when the buffer fills), and in a mixed fleet an old 500-cap client trimming the shared buffer can trip `partialTrimGap` and force full re-syncs on newer clients.
- **SPAP-9 `apply-snapshot` silently adopts remote** on a path that previously always prompted. It only fires when the snapshot clock dominates local (`GREATER_THAN`/`EQUAL`), which should mean local holds nothing new, and clock pruning (`MAX_VECTOR_CLOCK_SIZE=20`) pushes toward `keep-local`/`dialog` rather than a false adopt — so I believe it's safe, but it's a new silent auto-resolution on the data-loss-critical path and deserves explicit multi-client E2E validation, not just mock-adapter unit tests.
- **Shared per-provider state.** `_lastSeenRevs` / `_expectedSyncVersions` / `_lastSeenVectorClocks` are keyed only by `provider.id` and shared between the single-file and split paths; toggling the setting mid-life can cause one spurious gap/no-op (self-heals). Minor.
- **Testing.** Unit coverage is extensive and good (and the relocated Sync-Cycle-Cache tests retain coverage on LocalFile). The real risks — concurrent compaction, cross-version fleets, SPAP-9 silent-adopt — need the file-based + SuperSync/WebDAV E2E suites (the scheduled workflow) before merge.

---

**Recommendation:** the two 🔴 items should be fixed before merge (both look like small, well-contained fixes + regression tests), and I'd strongly favor splitting the shared-path changes (SPAP-8/9/10) out from the opt-in SPAP-11 so a regression in the default path can be reverted on its own. Happy to pair on the compaction-threshold fix and the allowlist correction.

aakhter:
Thanks — addressed both 🔴s and restructured. A note on structure: the split path genuinely builds on SPAP-9 (causality: `_lastSeenVectorClocks`, `compareVectorClocks`) and SPAP-10 (rev-precheck: `_lastSeenRevs`) — it has its own rev-precheck and cosmetic-reset for the split files on top of that shared infra. So rather than duplicate that, **this PR is now stacked on #8787 (SPAP-9) + #8795 (SPAP-10)** and should be reviewed/merged after them. Once they land, I'll rebase and this diff collapses to just the split-format addition — the granular, shared-path-separated end state you asked for.

**🔴 Compaction re-fires on every sync past 1000 — fixed.**
The trigger is now `> MAX_RECENT_OPS` (2000), while the post-compaction trim target stays `SPLIT_COMPACTION_THRESHOLD` (1000). So the ops buffer accumulates ~1000 cheap op-only syncs between compactions instead of rebuilding `sync-state.json` on every op-bearing sync once the folder crosses 1000. Updated the constant comment (it's the trim target, not the trigger). Left the download-side `partialTrimGap` check at `SPLIT_COMPACTION_THRESHOLD` on purpose — after a compaction the buffer floors at 1000, and a behind-client at that point genuinely needs the snapshot; raising it would *miss* that gap.
Added test (b2): two consecutive op-bearing syncs between the threshold and the cap assert `getStateSnapshot` is not called (the old code recompacted on both).

**🔴 Rev-precheck allowlist backwards — fixed in the SPAP-10 layer** (`[Dropbox, OneDrive]`).

**Follow-ups filed** (not bundled here): **SPAP-31** (one-way opt-in: block un-check + friendly old-client error + document no-downgrade), **SPAP-32** (version-addressed snapshot filename vs concurrent-compaction clobber), **SPAP-33** (`MAX_RECENT_OPS` mixed-fleet safety), **SPAP-34** (multi-client E2E for the auto-merge / silent-resolution paths).

johannesjo:
## Re-review — both 🔴 blockers resolved, two new precheck issues found

Thanks for the quick turnaround. I re-read the final code and verified both original blockers are genuinely fixed:

- **🔴 Perpetual compaction — fixed.** `needsCompaction` now triggers at `combinedOps.length > MAX_RECENT_OPS` (2000) and trims to `SPLIT_COMPACTION_THRESHOLD` (1000). The `(b2)` regression test asserts `getStateSnapshot` is not called between the threshold and the cap. 👍
- **🔴 Rev-precheck allowlist — fixed.** `_REV_PRECHECK_PROVIDERS = {Dropbox, OneDrive}`, WebDAV/LocalFile excluded, SPAP-29 filed to re-add WebDAV behind a strong ETag. 👍

The migration/tombstone design, crash-recovery via `.bak`, and the default-OFF auto-merge with the bridge-clock guard all look solid and well-tested.

While drafting the gap-guard test I found **two new issues, both in the SPAP-10 rev-precheck as applied to the split path** — same silent-data-loss class as the original blockers, so flagging before merge:

---

### 🔴 1. Split download commits `_lastSeenRevs` eagerly → crash-window op loss

`_downloadOpsSplit` commits the rev eagerly:

```ts
// file-based-sync-adapter.service.ts:1898 (_downloadOpsSplit)
this._lastSeenRevs.set(providerKey, opsRev);
```

The single-file path deliberately does **not** do this — it stages to `_pendingRevs` and only promotes to `_lastSeenRevs` inside `setLastServerSeq`, *after* the caller confirms the ops were durably applied (your own comment at ~line 996-1002 spells out why). The split path bypasses that mechanism.

**Failure:** poll downloads ops → `_lastSeenRevs` committed → crash/throw before the ops are applied (before `setLastServerSeq`). Next poll's precheck sees `remoteRev === lastSeenRev` → returns `{ops: []}` → the un-applied ops are skipped **for good**. This is exactly the scenario the single-file `_pendingRevs` staging was added to prevent; the new split copy reintroduces it (Dropbox/OneDrive only).

**Fix** — mirror the single-file path:

```ts
// _downloadOpsSplit, replace line 1898
this._pendingRevs.set(providerKey, opsRev);
```

(`setLastServerSeq` already promotes `_pendingRevs`, and the split upload path sets `_lastSeenRevs` itself, so nothing else needs to change.)

---

### 🔴 2. Split precheck ignores `sinceSeq` → defeats `forceFromSeq0` / USE_REMOTE

```ts
// file-based-sync-adapter.service.ts:1829-1834 (_downloadOpsSplit)
const lastSeenRev = this._lastSeenRevs.get(providerKey);
if (
  lastSeenRev &&
  !this._getCachedOpsData(providerKey) &&
  this._REV_PRECHECK_PROVIDERS.has(provider.id)
) {
```

`forceFromSeq0` (USE_REMOTE conflict resolution, fresh hydration) calls download with `sinceSeq === 0` specifically to re-pull the full snapshot. But when `lastSeenRev` is set and the rev is unchanged, the precheck short-circuits and returns `{ops: []}` with no `snapshotState` — so USE_REMOTE silently no-ops and the client stays diverged.

This is the same defect I flagged on the SPAP-10 PR (#8795) for the single-file precheck; it's present in the bundled SPAP-10 commit here too (line ~795), and this split copy needs the same guard:

```ts
if (
  sinceSeq > 0 &&
  lastSeenRev &&
  !this._getCachedOpsData(providerKey) &&
  this._REV_PRECHECK_PROVIDERS.has(provider.id)
) {
```

---

### 🟡 3. `partialTrimGap` length guard can mask a real gap (fold into SPAP-33)

```ts
// _downloadOpsSplit:1889-1893
const partialTrimGap =
  sinceSeq > 0 &&
  opsFile.oldestOpSyncVersion !== undefined &&
  opsFile.oldestOpSyncVersion > sinceSeq + 1 &&
  opsFile.recentOps.length >= FILE_BASED_SYNC_CONSTANTS.SPLIT_COMPACTION_THRESHOLD;
```

The `recentOps.length >= THRESHOLD` clause is redundant for correctness — `oldestOpSyncVersion > sinceSeq + 1` already proves the op at `sinceSeq+1` was trimmed away (it's older than the oldest retained op), and it never false-positives. But the length clause can produce a **false negative**: a migrated legacy buffer trimmed at a *smaller* floor (an old client's 500-op `MAX_RECENT_OPS`, or the pre-SPAP-9 default) can have `oldestOpSyncVersion > sinceSeq+1` while `recentOps.length < 1000` → no gap detected → and unlike the single-file path, the split path doesn't fetch the co-located snapshot, so the behind client silently diverges. The single-file guard (`>= MAX_RECENT_OPS`) has the same latent issue, newly exposed by SPAP-9 raising the cap 500→2000 — this is really the SPAP-33 dimension. Cleanest fix is to drop the length clause in both.

---

### Merge order
The PR base is still `master` with the SPAP-9 + SPAP-10 commits bundled, so the "stacked on #8787/#8795" is review-ordering only — merging as-is pulls those in ahead of their own threads. Recommend landing #8787 + #8795 first, then rebase so this collapses to just the split-format addition.

Net: the two original blockers are truly gone and the design is in good shape — #1 and #2 are small, mechanical fixes (both one-liners mirroring the single-file path) but they're silent-op-loss paths so worth closing before merge.

aakhter:
Thanks for the careful re-review — both new 🔴s fixed, and I rebased so the stack is clean.

**🔴 1 — split download committed `_lastSeenRevs` eagerly — fixed.**
`_downloadOpsSplit` now stages the ops-file rev to `_pendingRevs` (not `_lastSeenRevs`), mirroring the single-file path; `setLastServerSeq` promotes it after durable apply. The split *upload* path (`_uploadSnapshotSplit`) still sets `_lastSeenRevs` directly, as you noted. Test (g): a split download leaves `_lastSeenRevs` unset until `setLastServerSeq`, which then promotes it.

**🔴 2 — split precheck ignored `sinceSeq` — fixed.**
Added the `sinceSeq > 0 &&` guard to the `_downloadOpsSplit` precheck (same as the single-file fix). Test (h): a `forceFromSeq0` (sinceSeq=0) download with a matching seeded rev still fetches both `sync-ops.json` and `sync-state.json` (i.e. doesn't short-circuit).

**🟡 3 — `partialTrimGap` length-clause false negative — filed as SPAP-33** (you suggested folding it there). Captured your analysis: drop the redundant `recentOps.length >= …` clause in both the single-file and split paths (the `oldestOpSyncVersion > sinceSeq + 1` condition already proves the gap and never false-positives), plus a mixed-fleet regression test for a 500-cap-trimmed buffer. Left out of here to keep this PR to its blockers.

**Stack update:** #8787 (SPAP-9) merged, so I rebased onto current master and dropped the redundant SPAP-9 commit — this PR is now just **SPAP-10 + SPAP-11** (2 commits). Once #8795 merges it collapses to the split-format commit alone. Full op-log suite green.

aakhter:
Rebased onto current master now that #8787 and #8795 have merged — this PR is now a **single, standalone split-format commit** (no more bundled SPAP-9/10). Op-log suite green on the new base. Thanks again for the thorough reviews on this series. 🙏

PR #8806, opened 2026-07-07
perf(work-context): collapse per-tick activeWorkContext$ churn with distinctUntilChanged

aakhter:
## Problem
`activeWorkContext$` (`work-context.service.ts`) is `switchMap(select(selectActiveWorkContext)) + shareReplay(1)` with no `distinctUntilChanged`. `selectActiveWorkContext` depends on the task `entities`, which get a new reference **every second** while a task is tracking time (per-tick `addTimeSpent` → `taskAdapter.updateOne`). Its projector runs a full O(n) scan and returns a brand-new `WorkContext` object every tick, even when nothing a consumer reads has changed.

That cascades: `activeWorkContextTTData$` tears down and re-subscribes the time-tracking `state$` merge every second, and `activeWorkContextTitle$` re-runs — plus every downstream consumer (`page-title`, `mobile-bottom-nav`, `metric`, `add-task-bar`, plugin bridge, …) sees a new object reference each second.

## Fix
Add a content-aware `distinctUntilChanged` to `activeWorkContext$`. Rather than a hand-picked field list (risky given the many consumers that read different fields), the comparator is a full shallow-equality: arrays (`taskIds`/`backlogTaskIds`/`noteIds`) are compared by content (the selector regenerates them each tick), and every other field — including the nested `theme`/`advancedCfg` objects — by reference, which is correct under NgRx (a real change yields a new reference). This collapses the per-tick no-op churn while any genuine change (context switch, task-membership change, theme/cfg edit) still emits.

## Tests
- New `is-same-active-work-context.util.spec.ts` (11 cases) for the comparator.
- New store-driven test in `work-context.service.spec.ts` asserting that two comparator-equal emissions collapse to one downstream **and** `getWorkStartEndForWorkContext$` is not re-subscribed, while a real `taskIds` change still emits. Removing the operator fails this test.

### DIFF
--- a/src/app/features/work-context/is-same-active-work-context.util.spec.ts
+++ b/src/app/features/work-context/is-same-active-work-context.util.spec.ts
@@ -1,0 +1,95 @@
+import { isSameActiveWorkContext } from './is-same-active-work-context.util';
+import { WorkContext, WorkContextType } from './work-context.model';
+
+const baseCtx = (): WorkContext =>
+  ({
+    id: 'ctx1',
+    type: WorkContextType.TAG,
+    title: 'x',
+    icon: null,
+    routerLink: 'tag/ctx1',
+    isEnableBacklog: true,
+    theme: {},
+    advancedCfg: {},
+    taskIds: ['A', 'B'],
+    backlogTaskIds: ['C'],
+    noteIds: ['N'],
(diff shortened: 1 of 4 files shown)


johannesjo:

PR #8807, opened 2026-07-07
perf(tasks): stop per-tick O(n) Today read in autoAddTodayTagOnTracking

aakhter:
## Problem
`autoAddTodayTagOnTracking` (`task-related-model.effects.ts`) listens to `addTimeSpent` — which fires **every second** while a task tracks time — and uses `withLatestFrom(this._store.select(selectTodayTaskIds))`. `selectTodayTaskIds` runs `computeOrderedTaskIdsForToday`, an O(n) scan over all task entities. Because `withLatestFrom` keeps that selector continuously subscribed and the entity input changes each tick, the O(n) scan re-runs every second whenever `isAutoAddWorkedOnToToday` is enabled (the default).

## Fix
Behavior-preserving pipeline restructure:
- Run the cheap synchronous checks first (`isApplyingRemoteOps()` guard, then `!dueDay && dueWithTime !== number`) so the common already-scheduled tick short-circuits before any store read.
- `distinctUntilChanged` on `task.id` — the tracked task only changes occasionally, not per tick.
- Read `selectTodayTaskIds` **lazily** via `concatMap(action => store.select(selectTodayTaskIds).pipe(first(), …))` for the rare qualifying action, instead of a continuously-subscribed `withLatestFrom`. The membership check, parent-task logic, and hydration guard are unchanged (this keeps `selectTodayTaskIds`, whose `dueDay`-based virtual membership is the correct source — the tag's own `taskIds` is ordering-only).

## Tests
Extended `task-related-model.effects.spec.ts` (+5): 10 ticks for a task already in Today → 0 selector reads and 0 dispatches; an unscheduled task across 10 ticks → exactly one dispatch and one selector read; parent-in-Today → no dispatch; hydration guard → no dispatch. The discriminating test fails on the old effect body.

### DIFF
--- a/src/app/features/tasks/store/task-related-model.effects.spec.ts
+++ b/src/app/features/tasks/store/task-related-model.effects.spec.ts
@@ -139,5 +139,117 @@ describe('TaskRelatedModelEffects', () => {
       expect(emitted).toBe(false);
       subscription.unsubscribe();
     }));
+
+    describe('autoAddTodayTagOnTracking - perf/membership', () => {
+      const countTodaySelectCalls = (
+        selectSpy: jasmine.Spy<MockStore['select']>,
+      ): number =>
+        selectSpy.calls.all().filter((c) => c.args[0] === selectTodayTaskIds).length;
+
+      it('should NOT read the Today selector per tick for a task already scheduled (dueDay)', () => {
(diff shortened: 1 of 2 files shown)


johannesjo:

PR #8808, opened 2026-07-07
perf(planner): drop per-CD getters/allocations in planner-task and kb

aakhter:
## Problem
Work re-executed on every change-detection pass, multiplied by row count:
1. `planner-task.component` took a non-signal `@Input() task` and exposed per-CD getters — `titleHasLinks` (a regex/string scan), `timeEstimate`, and the `isDone`/`isDragReady`/`isCurrent` `@HostBinding` getters — all re-running on every CD pass. The sibling `schedule-event.component` already uses `computed()`; planner-task was the outlier.
2. The `kb` getters in `task.component` and `main-header.component` allocated a fresh `{}` on the no-shortcuts path (`… || {}`) on every CD pass, per row.

## Fix
1. Convert `planner-task`'s `task` to a required signal input, the getters to `computed()`, and the class bindings to `host: {}` metadata (mirrors `schedule-event`). No function-body getters remain.
2. Add a frozen `EMPTY_KEYBOARD_CONFIG` + a small `keyboardConfigOrEmpty()` helper so the no-shortcuts path returns a referentially-stable value instead of allocating each pass. (`task.component`'s `isTouchActive()` branch is preserved verbatim.)

Behavior — current-task highlight, drag-ready class, done styling, keyboard-shortcut hints, sub-task count — is unchanged.

## Tests
- New `keyboard-config.const.spec.ts` (4): the helper returns the frozen singleton on the empty path (`===` across reads) and the same reference for a present config.
- New `planner-task.component.spec.ts` (8): a render spec that sets the `task` input, asserts the computed outputs and live `isDone`/`isCurrent` host classes, and renders without throwing.

### DIFF
--- a/electron/shared-with-frontend/keyboard-config.model.ts
+++ b/electron/shared-with-frontend/keyboard-config.model.ts
@@ -59,6 +59,14 @@ export type KeyboardConfig = Readonly<{
   [key: `plugin_${string}`]: string | null;
 }>;
 
+/** Shared frozen empty keyboard config so the "no shortcuts" path returns a
+ * referentially-stable value instead of allocating a fresh {} every CD pass. */
+export const EMPTY_KEYBOARD_CONFIG: KeyboardConfig = Object.freeze({}) as KeyboardConfig;
+
+export const keyboardConfigOrEmpty = (
+  keyboard: KeyboardConfig | undefined,
+): KeyboardConfig => keyboard ?? EMPTY_KEYBOARD_CONFIG;
+
 export const GLOBAL_KEY_CFG_KEYS: (keyof KeyboardConfig)[] = [
   'globalShowHide',
   'globalToggleTaskStart',
--- a/src/app/core-ui/main-header/main-header.component.ts
+++ b/src/app/core-ui/main-header/main-header.component.ts
@@ -23,7 +23,7 @@ import { SyncWrapperService } from '../../imex/sync/sync-wrapper.service';
 import { SnackService } from '../../core/snack/snack.service';
 import { NavigationEnd, Router } from '@angular/router';
 import { GlobalConfigService } from '../../features/config/global-config.service';
-import { KeyboardConfig } from '@sp/keyboard-config';
+import { KeyboardConfig, keyboardConfigOrEmpty } from '@sp/keyboard-config';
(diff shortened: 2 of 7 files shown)


johannesjo:
**Reviewed — looks good to merge.** Clean, well-scoped perf refactor; behavior is preserved and CI is green. A couple of optional notes, no blockers.

### Strengths
- The `planner-task` migration faithfully mirrors the existing `schedule-event` pattern (signal input + `computed()` + `host: {}` metadata), so it stays consistent with the sibling component rather than being a one-off.
- Good catch on the `|| {}` allocation: beyond avoiding a fresh object per CD, the frozen `EMPTY_KEYBOARD_CONFIG` keeps `[kb]="kb"` bindings referentially stable, so OnPush children (e.g. `main-header`, `task-hover-controls`) stop being marked dirty every pass by an ever-new `{}`. I checked every `kb.*` usage in the repo — all read-only — so freezing is safe.
- Solid test coverage for both changes.

### Optional / follow-up (non-blocking)
1. **The same `|| {}` pattern remains in two sibling getters** that `keyboardConfigOrEmpty()` could now cover:
   - `page-title.component.ts:338` — used directly in the inline template, so it's a genuine per-CD allocation (single instance, so minor).
   - `task-context-menu-inner.component.ts:331` — transient (only while the menu is open), lower impact.

   (`focus-button` and `mobile-side-panel-menu` already memoize via `computed`/field-init, so those are correctly left alone.) Applying the helper there too would round out the cleanup — fine to defer.

2. **Pre-existing, out of scope:** in `timeEstimate`, the `subTaskIds ? … : (timeEstimate - timeSpent)` branch — since `subTaskIds` is typed `string[]` (always a, possibly empty, array → always truthy), the subtraction branch is effectively unreachable for normally-shaped data, so leaf tasks always show the raw estimate. Unchanged by this PR (correctly preserved verbatim); just flagging in case the "remaining time" behavior was ever intended — worth a separate issue if so.

Verified: all 6 parent bindings use `[task]` only (no template-refs/queries against the removed getters), the host-class migration is reactive-equivalent under OnPush, and the required signal input is only read after it's set (`ngOnInit` / template). 👍

johannesjo:

PR #8809, opened 2026-07-07
perf(tasks): stable per-task selector outputs via per-subscription factories

aakhter:
## Problem
Two compounding defects around task list selectors:

1. **Shared memo thrash.** `selectTasksWithSubTasksByIds` and `selectTasksById` are module-level `createSelector`s taking a `{ ids }` prop. NgRx keeps one depth-1 memo slot per selector, shared across every concurrent subscriber (today list, backlog, markdown export, planner, …). Different `ids` mutually evict that slot, so both recompute on **every** dispatched action.
2. **Per-tick object churn.** `selectTasksWithSubTasksByIds` rebuilds `{...task, subTasks}` for every id whenever the task `entities` slice changes — i.e. every second while a task tracks time — so `mainListTasks$`/`backlogTasks$` emit all-new `TaskWithSubTasks` arrays each tick and every OnPush task row re-evaluates its template, even though only the tracked task changed.

## Fix
- Add `selectTasksWithSubTasksByIdsFactory(ids)` / `selectTasksByIdFactory(ids)` that mint a **fresh `createSelector` per subscription** (created inside the `switchMap` at each call site), so each subscription owns its own memo instead of fighting over one shared slot.
- The with-subtasks factory keeps a per-instance cache keyed on the **task entity ref + its subtask entity refs**; a task whose refs are unchanged returns the *identical* `TaskWithSubTasks` object as last emission, so only the tracked task (changed ref) gets a new object. Output shape/order/filtering are unchanged (the rebuild is the same `{...task, subTasks}` as `mapSubTasksToTask`).
- Belt-and-braces `distinctUntilChanged` (element-wise `===`) on `mainListTasks$`/`backlogTasks$` so an unchanged list stops re-emitting entirely.

Net effect: while a task tracks time, only the tracking row's inputs change reference; the other visible rows keep identical `TaskWithSubTasks` references and don't re-render.

## Tests
`task.selectors.spec.ts` (+9): only-X-changed → all other tasks share object identity, X does not; a subtask ref change rebuilds its parent; two concurrent factory instances with different id-sets don't evict each other; and shape-parity `toEqual` against the legacy selector. These fail against the old shared selectors and pass after. `work-context.service.spec.ts` (13) still green with the migrated call sites.

### DIFF
--- a/src/app/features/tasks/store/task.selectors.spec.ts
+++ b/src/app/features/tasks/store/task.selectors.spec.ts
@@ -1145,4 +1145,155 @@ describe('Task Selectors', () => {
     expect(ids).not.toContain('overdueInArchived');
     expect(ids).toContain('task6');
   });
+
+  // SPAP-19: per-subscription stable task selector factories
+  // -------------------------------------------------------------------------
+  describe('SPAP-19 stable selector factories', () => {
+    const makeTask = (id: string, over: Partial<Task> = {}): Task =>
+      ({
+        ...DEFAULT_TASK,
+        id,
+        title: id,
+        created: 0,
+        subTaskIds: [],
+        tagIds: [],
+        timeSpentOnDay: {},
+        ...over,
+      }) as Task;
+
+    const makeTaskState = (tasks: Task[]): TaskState => ({
+      ids: tasks.map((t) => t.id),
+      entities: tasks.reduce(
+        (acc, t) => {
+          acc[t.id] = t;
+          return acc;
+        },
+        {} as Record<string, Task>,
+      ),
+      currentTaskId: null,
+      selectedTaskId: null,
+      lastCurrentTaskId: null,
+      isDataLoaded: true,
+      taskDetailTargetPanel: null,
+    });
+
+    // Selectors only read the TASK feature slice, so a minimal root is enough.
+    const wrap = (ts: TaskState): any => ({ [TASK_FEATURE_NAME]: ts });
+
+    describe('selectTasksWithSubTasksByIdsFactory', () => {
+      it('returns identical refs for tasks whose entity ref did not change (only C changed)', () => {
+        const a = makeTask('A');
+        const b = makeTask('B');
+        const c = makeTask('C');
+        const sel = fromSelectors.selectTasksWithSubTasksByIdsFactory(['A', 'B', 'C']);
+
+        const r1 = sel(wrap(makeTaskState([a, b, c])));
+        // Only C's entity ref changes (e.g. a time-tracking tick bumping timeSpent)
+        const c2 = { ...c, timeSpent: 999 };
+        const r2 = sel(wrap(makeTaskState([a, b, c2])));
+
+        expect(r2[0]).toBe(r1[0]); // A: same object
+        expect(r2[1]).toBe(r1[1]); // B: same object
+        expect(r2[2]).not.toBe(r1[2]); // C: rebuilt
+        expect(r2[2].timeSpent).toBe(999);
+      });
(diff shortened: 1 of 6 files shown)


johannesjo:
## Review — approve ✅

Nice, well-scoped fix with a clear write-up. I read the diff in full context, applied it locally, and verified:

- **Lint/format** clean on all 6 files.
- `task.selectors.spec.ts` **64/64** (incl. the 9 new cases), `work-context.service.spec.ts` **13/13**; both compile through the ng/Karma pipeline so the `MemoizedSelector<object, T>` signatures and migrated call sites type-check.
- All 4 production call sites migrated — no production caller of the legacy `selectTasksWithSubTasksByIds` / `selectTasksById` remains.

### Why it's correct (the two invariants I traced)

1. **The per-instance cache survives ticks.** `mainListTaskIds$` / `backlogTaskIds$` already carry `distinctUntilChanged(fastArrayCompare)`, so the `switchMap` only re-mints the factory when the *id list* changes — not on every time-tracking tick. During ticks the factory and its cache persist, which is the whole premise.
2. **Suppression can't drop a real update.** `distinctUntilChanged` retains the last *emitted* array, so a suppressed no-op stays content-equal to what the view holds and the next genuine change still compares unequal. Because state is immutable, an element ref changes iff its content changed, and `TaskWithSubTasks` derives entirely from the task entity + its subtask entities — so element-wise `===` is a sound equality. Confirmed against `@ngrx/store` 21's `defaultMemoize`: the projector only re-runs when the feature slice ref changes, and it always returns a fresh array, so the belt-and-braces `distinctUntilChanged` is indeed the layer doing the dedup.

Shape parity with `mapSubTasksToTask` (`{...task, subTasks}`, same missing-entity `devError`, same filtering) is preserved and covered by the `toEqual` test. Also checked the `getByIdsLive$` consumers on the new `selectTasksByIdFactory`: the two repeat-cfg effects and finish-day are `first()`-bounded, and the reminders dialog wraps it in a `switchMap` with its own `distinctUntilChanged` — all bounded, no leaked factories, output unchanged.

### Non-blocking notes

1. **Reusable-API footgun — worth a JSDoc line.** The optimization only holds if the factory is minted **once per distinct id-set** (inside a `switchMap` over deduped ids, or a one-shot `first()`). A future caller who invokes `selectTasksWithSubTasksByIdsFactory(ids)` inside a `map`/getter that fires per store-emission would silently get an empty cache every time — *worse* than the old code (per-tick allocation + no memoization) and with no error to catch it. The inline comments cover it, but since these are now exported, shared selectors, a short "must be created per-subscription, never per-emission" note on the two factories would harden the contract.
2. **Two legacy selectors are now test-only.** `selectTasksWithSubTasksByIds` / `selectTasksById` have no remaining production callers — retained solely for the parity/contrast specs (which is a good reason to keep them). A one-line comment saying so would save the next reader a "why is this unused?" hunt.
3. **Guard asymmetry is intentional, just noting it.** `selectTasksByIdFactory` guards `ids ? … : []` while the with-subtasks factory does a bare `for (const id of ids)`. This faithfully mirrors each legacy selector's original guarding and every call site passes a real array (`plan-tasks-tomorrow` even adds `ids ?? []`), so no defect — just flagging the asymmetry in case you'd rather normalize it.

### One thing to smoke before merge
The correctness is solid; the *magnitude* of the win and the absence of visual regressions in long lists are argued from the mechanism, not measured. Given the task row is a hot path, a quick manual check — start a timer and watch a long Today list — would be worth it.

johannesjo:

PR #8810, opened 2026-07-07
feat(tasks): optional setting to sort completed tasks by completion date

oldrobotdev:
## Problem

Completed tasks in the "Done" list keep whatever position they had when they were checked off, so the task you just finished isn't necessarily at the top. There's currently no way to review completed work most-recent-first.

## Solution

Adds an optional setting under **Settings → Tasks**: *"Sort completed tasks by completion date (newest first)"*. It is off by default, so existing behaviour is unchanged unless a user turns it on. When enabled, the Done list is ordered by each task's `doneOn` timestamp, descending.

The sort is applied once at the `doneTasks$` observable in `WorkContextService`, so it covers both the Today list and project/tag contexts. The comparator is a small pure helper (`sortDoneTasksByDoneDate`) with unit tests; the source array is copied before sorting since it comes from a memoized selector.

Files touched: the `tasks` config model + default, the Tasks settings form, the `T`/`en.json` label, `doneTasks$`, and a spec.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


oldrobotdev:
Tested on the preview build (`eb3633d2.super-productivity-preview.pages.dev`): the new toggle shows up under **Settings → Tasks** ("Sort completed tasks by completion date (newest first)"), it's off by default, and enabling it persists — the "settings updated" confirmation fires. The ordering itself is covered by the unit tests added here, and the full test job is green.

Happy to adjust the label, the default, or the placement if you'd prefer something else.

johannesjo:
Thank you very much! <3

Issue #7383, opened 2026-04-27
💥 NG0701


Doujiang-Xiang:
### Steps to Reproduce

1.
2.
3.

### Additional Console Output

META: SP18.2.8 __ Electron – zh-CN – Win32 – zh-CN – UA:Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) superProductivity/18.2.8 Chrome/146.0.7680.179 Electron/41.2.0 Safari/537.36

### Url
file:///D:/super%20productivity/resources/app.asar/.tmp/angular-dist/browser/index.html#/schedule

### Meta Info
META: SP18.2.8 __ Electron – zh-CN – Win32 – zh-CN – UA:Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) superProductivity/18.2.8 Chrome/146.0.7680.179 Electron/41.2.0 Safari/537.36

### Actions Before Error
```
1777255306247: [Task] SetSelectedTask 
1777255306334: [SP_ALL] Load(import) all data 
1777255306338: [AppState] Set Today String 
1777255306340: [SP_ALL] All Data was loaded 
1777255306360: [Task] SetSelectedTask
```

johannesjo:
Thanks for the report. NG0701 is Angular's runtime error for "Missing locale data for the locale 'X'".

In 18.2.8 (commit d989c064f3) non-default locale data started being lazily registered via `requestIdleCallback` to shrink the initial bundle. Only `en-GB` is registered synchronously at boot; `zh` and the others register a few hundred ms later. If the schedule view renders before the idle callback fires — and your action log (initial `Load(import) all data` between two `SetSelectedTask` events) suggests the schedule rendered while sync replay was still warming up — then `formatDate(date, fmt, 'zh-cn')` inside `headerTitle()` throws NG0701 (`schedule.component.ts:155-162` and `schedule-month.component.ts:48`).

I have a deterministic unit test that reproduces this for `zh-cn` and the schedule view. To confirm it matches your specific report rather than a different NG0701 trigger, could you share the **full console output** (Ctrl+Shift+I → Console tab)? The NG0701 message includes the failing locale name in quotes plus the stack trace pointing at the call site — that would pin it down completely.

A fix is in progress — register the user's configured locale synchronously at startup instead of deferring it to idle, while keeping the lazy load for everything else.

PR #8854, opened 2026-07-08
perf(ui): make date/keys pipes pure to stop per-CD formatting

aakhter:
## Problem
`localeDate` and `scheduledDateGroup` (`src/app/ui/pipes/`) were `pure: false`, so they re-ran on **every** change-detection pass — and `localeDate` re-allocated a `DatePipe` and formatted via `Intl` each time. In the schedule **month** grid that's 35–42 format calls per CD cycle; the schedule week and planner day headers are hot too. `keys` was `pure: false` as well, and `to-array` was an impure pipe with zero template usages.

## Fix
Make all four pure (delete `to-array`), preserving locale reactivity:
- **Hot grids** (schedule-month, schedule-week, planner-day) precompute their day-label strings in a `computed()` keyed on `(days, currentLocale())` and bind plain strings — the per-cell pipe is gone entirely (42 → 0 format calls per CD).
- **Remaining `localeDate` sites** pass the reactive locale as an explicit arg (`:locale()`), each component exposing `readonly locale = dateTimeFormatService.currentLocale` — so CD re-formats only when the locale signal changes.
- `scheduledDateGroup` takes `today` as an arg (passed from `work-view`).
- `keys` becomes pure. One consequence: `dialog-time-estimate.deleteValue()` deleted a map entry **in place**, which a pure `keys` pipe wouldn't observe — changed to replace the reference immutably so the row still disappears.

`grep "pure: false" src/app/ui/pipes/` now returns nothing.

## Tests
`locale-date.pipe.spec` (11) asserts pure metadata + that changing the `locale` arg changes output (en-US "January" vs de-DE "Januar"); `scheduled-date-group.pipe.spec` (14); `keys.pipe.spec` (1); `schedule-month.component.spec` (+4) verifies the precomputed labels are correct, memoized (no per-CD recompute), and recompute on days/locale change. Locale switch at runtime still updates all visible dates.


johannesjo:
## Review — `perf(ui): make date/keys pipes pure`

Good change overall: the hot-grid `computed()` precompute is the right call and the pipe-purity tests are well targeted. One merge blocker, plus a couple of smaller notes.

### 🔴 Blocker — CI `Tests` is red (9 failing specs)

All 9 fail with the same DI error:

```
NG0201: No provider found for `Store`.
Path: DateTimeFormatService -> GlobalConfigService -> Store
```

- `op-log/sync/dialog-sync-import-conflict/…spec.ts` (7)
- `features/issue-panel/issue-panel-calendar-agenda/…spec.ts` (2)

**Cause:** the new eager field initializer

```ts
readonly locale = this._dateTimeFormatService.currentLocale;
```

constructs `DateTimeFormatService` (→ `GlobalConfigService` → `Store`) **at component instantiation**. Previously the service was only created lazily by the `localeDate` pipe when a template rendered it — so these two specs, which override the template to `''` (sync-import) or don't provide `Store` (agenda), never triggered it. Confirmed locally: `dialog-sync-import-conflict` → **7 FAILED, 2 SUCCESS**.

**Fix:** provide a mock in each failing spec. In `dialog-sync-import-conflict.component.spec.ts` the `renderWithScenario` block *in the same file* already has exactly what's needed — just mirror it into the `setup()` providers:

```ts
{ provide: DateTimeFormatService, useValue: { currentLocale: () => 'en-US' } }
```

Verified locally: adding that line makes the spec green (**9 SUCCESS**). The same one-line mock fixes the agenda spec (the error is at construction, before the template renders).

### 🟡 Minor — lazy locale-data registration can leave stale labels

The new `computed()`s (`schedule-week` `'EEE'`, `planner-day` `'EEE'`) and the now-pure `localeDate` sites only recompute when their `(days | value, locale-string)` deps change. Non-default locale *data* is registered via `requestIdleCallback` (`safeFormatDate` / #7383); if it registers *after* first render while the locale *string* is unchanged, weekday/month labels can stay in the default locale until the next days/locale signal change. The old impure pipe re-ran every CD and self-healed. Low risk in practice (these views open well after boot, and `daysToShow` changes frequently) and consistent with the existing `safeFormatDate` contract — noting it because the impure pipe used to mask it for these specific cells.

### 💬 Scope (non-blocking)

The measurable win is the three hot grids. The ~30 other touched files are cold paths (dialogs, detail panel) where an impure pipe ran only a handful of times — so most of the diff is the `readonly locale` boilerplate required to let the *shared* pipes go pure, which is also what introduced the test breakage. Global pipe purity is defensible (impure pipes are an anti-pattern); just flagging it as a conscious tradeoff vs. a hot-grids-only version with a much smaller surface.

### ✅ Looks good

- `deleteValue` immutable replace + keeping `taskCopy.timeSpentOnDay` in sync is the correct adaptation for the pure `keys` pipe (verified it's the only consumer).
- `planner-day.dayLabel` reading the non-signal `day.dayDate` inside a `computed()` is safe — the parent's `@for (…; track day.dayDate)` makes `dayDate` invariant for the instance's lifetime, so only locale needs to invalidate. Nicely documented.
- `scheduledDateGroup` + `todayStr()` is fine / arguably more correct — `selectTodayStr` is kept live by the `setTodayString` effect (timer/focus/visibility).
- `to-array` deletion is clean — zero remaining usages.

Once CI is green the core change looks sound.

johannesjo:
Thank you very much! <3

PR #8858, opened 2026-07-08
perf(tasks): content-stable scheduling snapshot for collection task selectors

aakhter:
## Problem
The collection task selectors — `selectOverdueTasks`, `selectAllTasksWithDueTimeSorted` → `selectTimeConflictTaskIds`, `selectAllUndoneTasksWithDeadlineSorted`, `selectAllUndoneTasksWithDueDay`, `selectLaterTodayTasksWithSubTasks`, `selectAllTasksWithSubTasks`, and work-context's `selectTodayTaskIds` — all derive from `selectAllTasks`, which returns a **new array every second** while a task tracks time (the tracked task's `timeSpent` bump changes the entities slice). None of them read `timeSpent`, yet each re-ran its O(n) filter/sort and reallocated on every tick.

## Fix
Introduce `selectTaskSchedulingSnapshot`: a projection of each task to **scheduling fields only** (`id, isDone, dueDay, dueWithTime, deadlineDay, deadlineWithTime, parentId, subTaskIds`), per-entity ref-cached and array-ref-stable — so a `timeSpent`-only tick yields the **identical** snapshot reference. Each collection selector now:
1. derives its filter/sort **decision** (an ordered id list) from the snapshot — skipped by NgRx memoization when the snapshot is unchanged; and
2. re-maps that stable id list back to **live** task entities through a per-selector memoized mapper, preserving the existing output types and returning the previous array reference unless a genuine member changed.

Net: while a task tracks time, the expensive filter/sort no longer runs and downstream consumers keep stable references — but any real scheduling change (dueDay/dueWithTime/isDone/deadline/membership) still recomputes, and members are always served with their live refs (no staleness).

`selectTimeConflictTaskIds` is intentionally left on the live sorted selector because `getTimeLeftForTask` reads `timeSpent`.

## Notes
- Builds on the merged per-subscription selector factories (#8809); this branch is rebased onto current `master` and the two coexist.
- Two minor documented behaviours: `selectLaterTodayTasksWithSubTasks`'s `Date.now()` cutoff now re-evaluates on scheduling/day changes rather than incidentally per tick; `selectTodayTaskIds`'s unpinned-task tail is ordered by `state.ids` (same set, arbitrary-but-stable either way).

## Tests
`task.selectors.spec` (+ snapshot suite): a `timeSpent`-only change → snapshot identical ref and decision selectors return identical refs; a real scheduling change → recompute + live member ref; shape-parity `toEqual` vs the pre-refactor logic. `work-context.selectors.spec` and `task-later-today.selectors.spec` green.

### DIFF
--- a/src/app/features/tasks/store/task-later-today.selectors.spec.ts
+++ b/src/app/features/tasks/store/task-later-today.selectors.spec.ts
@@ -1,6 +1,42 @@
-import { selectLaterTodayTasksWithSubTasks } from './task.selectors';
-import { Task } from '../task.model';
+import { selectLaterTodayStructure, SchedulingSnapshot } from './task.selectors';
+import { Task, TaskWithSubTasks } from '../task.model';
 import { getDbDateStr } from '../../../util/get-db-date-str';
+
+// SPAP-20: selectLaterTodayTasksWithSubTasks was split into a snapshot-based
+// decision (selectLaterTodayStructure → ordered {id, subTaskIds}) + a live-entity
+// re-map. This helper reproduces the old single-projector output shape from a raw
+// task array so the existing behavioral assertions below stay unchanged.
+const runLaterToday = (
+  tasks: (Task | undefined)[],
+  todayStr: string,
+  offset: number = 0,
+): TaskWithSubTasks[] => {
+  const snapshot: SchedulingSnapshot[] = tasks
+    .filter((t): t is Task => !!t)
+    .map((t) => ({
+      id: t.id,
+      isDone: t.isDone,
+      dueDay: t.dueDay ?? null,
+      dueWithTime: t.dueWithTime ?? null,
+      deadlineDay: t.deadlineDay ?? null,
+      deadlineWithTime: t.deadlineWithTime ?? null,
+      parentId: t.parentId ?? null,
+      subTaskIds: t.subTaskIds,
+    }));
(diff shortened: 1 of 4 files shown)


johannesjo:
Thank you very much @aakhter ! Can you have a look at the failing CI?

aakhter:
Fixed and green now ✅ (52178a5).

The failure was test-isolation, not the selector logic. `dialog-schedule-task.*` and `time-block-sync.effects` specs do `store.overrideSelector(selectAllTasksWithDueTimeSorted, …)`. `overrideSelector` uses `setResult()`, which sticks on the singleton selector across spec files and is only reset by `clearResult()` (not `release()`). Since this PR made `selectAllTasksWithDueTimeSorted` a public selector that `task.selectors.spec` now exercises directly against a hand-built state, whenever one of those specs ran first in the full karma bundle the leaked value came back instead — `Expected [ ] to contain 'task5'`. Single-file runs passed, so it only showed up in the full CI run.

Fix: added the selector to the existing `beforeEach` `clearResult()`/`release()` defensive block in `task.selectors.spec.ts` (same pattern already used for `selectOverdueTasks` etc.). Repro confirmed: `dialog-schedule-task` + `task.selectors` together went 1 FAILED → 90 SUCCESS.

johannesjo:
Thank you very much! <3

Issue #8843, opened 2026-07-07
Architecture review: verified low-risk quick wins (atomic archive write, dead deps, stale tokens, SHA-pins)


johannesjo:
Verified, low-risk cleanups surfaced by the 2026-07-07 architecture review and confirmed in an adversarial second pass. Each is independent and small; grouped here to avoid tracker noise. Check off as done.

- [ ] **Archive compression: use the atomic write.** `archive-compression.service.ts:85-88` writes young + old via `Promise.all([saveArchiveYoung, saveArchiveOld])` — two independent transactions. `saveArchivesAtomic` already exists (`archive-store.service.ts`) and is already used by `archive-operation-handler.service.ts:323`. A crash between the two writes leaves a half-compressed archive, and since compression is op-replayed on other clients, a torn local result diverges from replicas. One-line swap to the existing atomic API.

- [ ] **Drop dead dependency config.** `package.json` runs `postinstall: patch-package` but there is **no `patches/` directory** — a dead hook that still executes third-party code on every `npm install`. The `resolutions` block is npm-ignored (npm honors `overrides`, not yarn's `resolutions`), so the pins are no-ops that mislead readers. Remove both (move any real need into `overrides`).

- [ ] **Fix ARCHITECTURE-DECISIONS.md #3 stale claims.** (a) "`@sp/sync-core` has no runtime dependencies" — it declares `@noble/ciphers` and `hash-wasm` as `dependencies`. (b) "`packages/shared-schema` compatibility-re-exports vector-clock algorithms from `@sp/sync-core`" — no such re-export exists (shared-schema's only dep is zod). It's the designated load-bearing-decisions file, so drift here misleads future boundary changes.

[...]

PR #8862, opened 2026-07-08
fix(archive): write compressed archives in a single atomic transaction

costajohnt:
## Problem

`ArchiveCompressionService.compressArchive()` wrote `archiveYoung` and `archiveOld` as two independent transactions (`Promise.all([saveArchiveYoung, saveArchiveOld])`). A crash between the two writes leaves a half-compressed archive, and since compression is op-replayed on other clients, a torn local result diverges from replicas.

Addresses the "atomic archive write" item in #8843.

## Solution

Route both writes through the existing `ArchiveDbAdapter.saveArchivesAtomic(young, old)`, which commits both stores in a single IndexedDB transaction, so either both writes land or neither does. This mirrors the flush-young-to-old path already using `saveArchivesAtomic` in `archive-operation-handler.service.ts`.

Adds `archive-compression.service.spec.ts` asserting compression routes through the atomic API and never falls back to the two independent writes.

## Type of Change

- [x] Bug fix

## Checklist

- [ ] I have included relevant changes to the documentation/wiki. (n/a — no doc or user-facing surface change)

### DIFF
--- a/src/app/features/archive/archive-compression.service.spec.ts
+++ b/src/app/features/archive/archive-compression.service.spec.ts
@@ -1,0 +1,63 @@
+import { TestBed } from '@angular/core/testing';
+import { ArchiveCompressionService } from './archive-compression.service';
+import { ArchiveDbAdapter } from '../../core/persistence/archive-db-adapter.service';
+import { ArchiveModel } from './archive.model';
+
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! <3

PR #8863, opened 2026-07-08
fix(work-context): make isTodayListActive reactive to context changes

costajohnt:
## Problem

`task.component.ts`'s `isTodayListActive` was `computed(() => this.workContextService.isTodayList)`. `WorkContextService.isTodayList` is a plain mutable boolean (kept in sync via a constructor subscribe), not a signal, so the `computed()` has no producer and never invalidates — it returns its first value for the component's lifetime. Masked in practice by per-work-context component recreation + shareReplay(1), but a signals anti-pattern in the hot path.

Addresses the "stale computed in the task hot path" item in #8843.

## Solution

Expose `isTodayListSignal = toSignal(this.isTodayList$, { initialValue: false })` on `WorkContextService` and read that from the computed, so it tracks context changes reactively. Placed on the service (rather than a component-local `toSignal`) as the shared home for this derived signal; the plain `isTodayList` boolean is kept because `history.component.ts` reads it synchronously in a non-reactive method.

Adds a `work-context.service.spec.ts` test asserting the signal is true on the Today context and flips to false when the active context changes.

## Type of Change

- [x] Bug fix

## Checklist

- [ ] I have included relevant changes to the documentation/wiki. (n/a — no doc or user-facing surface change)

### DIFF
--- a/src/app/features/tasks/task/task.component.ts
+++ b/src/app/features/tasks/task/task.component.ts
@@ -234,7 +234,7 @@ export class TaskComponent implements OnDestroy, AfterViewInit {
     return 'chat';
   });
 
-  isTodayListActive = computed(() => this.workContextService.isTodayList);
+  isTodayListActive = computed(() => this.workContextService.isTodayListSignal());
   taskIdWithPrefix = computed(() => 't-' + this.task().id);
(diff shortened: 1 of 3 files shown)


johannesjo:
Thank you very much! <3

PR #8864, opened 2026-07-09
chore(config): adopt AGENTS.md as shared AI-agent config with initial skill

lanesawyer:
## Problem

I noticed that the AI Agent setup wasn't as tool-agnostic as it could be. `CLAUDE.md` is the primary source of truth, while `AGENTS.md` said `Please read CLAUDE.md!!`.

Most tooling has adopted `AGENTS.md` as the entrypoint, with Anthropic the only one (that I know of) who continues to ignore `AGENTS.md`. 

I also noticed `AGENTS.md` is a little large and likely could be slimmed down with some custom skills. I've only included one skill so far since it was the easiest to identify and I didn't want to do _too_ much without a conversation with you all first!

## Solution

So this PR switches the source of truth to `AGENTS.md`, and `CLAUDE.md` is now a symlinked file to `AGENTS.md`.

Removed the commit message creation insturctions from `AGENTS.md` and made it a skill in `.agents/skills`, with `.claude/skills` having a symlink to that. I did have to modify the `.gitignore` since I was surprised to find _all_ of the `.claude` folder to be excluded, which would keep the skills from being picked up on otherwise. I think that most items in `.claude` generally should be included in the repo, though that's a discussion for another thread.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thank you very much ! <3

PR #8865, opened 2026-07-09
feat(calendar): support multiple Google Calendar accounts

HanCheo:
## Problem

Google Calendar issue-provider instances shared one device-local OAuth credential. Connecting another Google account could overwrite the first account's authorization.

## Solution

- Scope OAuth credentials by provider ID only for the bundled Google Calendar plugin. Uploaded OAuth plugins retain the documented unscoped API.
- Keep the provider ID transient in host-to-bundled-plugin config; no new synced field or published plugin API is added.
- Move (not copy) the legacy credential atomically in IndexedDB to the sole unambiguous local provider at startup. Remove the legacy credential and the initialized-marker mechanism. A new-provider dialog never performs migration.
- Capture the local provider identity before waiting for sync. Never bind a legacy credential to a replacement provider introduced by remote operations or a full-state import; ambiguous/fallback-hydration ownership requires reconnection.
- Observe committed provider-state changes for local, remote and full-import deletion cleanup without ALL_ACTIONS or additional synced operations. Serialize migration and deletion work.
- Remove unsaved scoped credentials on dialog cancellation/destruction, including OAuth completion after cancellation. Saved entries and other accounts retain their credentials.
- Fence pending token refresh responses so disconnect/reconnect cannot resurrect an old account or invalidate its replacement.
- Clone plugin config before adding the internal token key, require the plugin ID at adapter callsites, remove dead migration helpers, and document the internal contract and user-facing behavior.

## Verification

- Focused Angular OAuth service/bridge/lifecycle, provider-dialog, token-key helper and sync-adapter suites: **98 passed**.
- Five additional regression cases first failed before the follow-up fix: disconnect during refresh, replacement-account refresh isolation, stale refresh failure, fallback-hydration legacy ownership, and preserving the boot account when sync adds another provider. All pass after the fix.
- Repository pre-push hook: full lint and test command passed.
- Google Calendar provider tests: **40 passed**.
- Google Calendar provider and plugin-api builds: passed.
- npm run checkFile on all changed TypeScript files: passed.
- Browser smoke with synthetic local credentials: legacy-only new-provider dialog stays disconnected; five add/cancel cycles create no token copies; actual Cancel removes an unsaved scoped token without deleting the unrelated legacy fixture.

Actual Google OAuth login with two live accounts was **not** exercised. The browser smoke used the web build (where Google OAuth connection is unavailable); account isolation and token lifecycle were exercised with synthetic credentials. Existing Node/browser-support warnings remain.

Compatibility note: temporary initialized-marker data created by earlier unmerged revisions of this PR is not migrated; the marker mechanism is removed as requested.


johannesjo:
Thanks for the PR, and welcome! 🎉 The multi-account problem is real — today every `google-calendar` provider instance shares a single OAuth token (`google-calendar__oauth`), so connecting a second account overwrites the first. Scoping the token per instance is the right fix, and I checked carefully that **existing single-account users are not affected**: the `isEdit` guard means legacy providers never get an `oauthTokenKey`, so they keep using the unscoped token and `getHeaders` falls back to `getOAuthToken()`. Nicely handled. 👍

Before merging I'd like to tighten the scope and resolve a few things:

**1. This is really three changes — please split.** The per-instance OAuth scoping is the feature; the other two are unrelated and make the diff harder to reason about:

- **`mergeIssueProviderModelUpdates` deep-equality change** looks redundant. The multiSelect freeze was already fixed in #8798 by assigning Formly's emitted model by reference in `formlyModelChange()` (line ~267) — that path no longer goes through this util, which is now only reached from the custom-cfg `updateModel()` path. Can you confirm a case where this is still needed? Otherwise I'd drop it, since it adds a recursive equality helper on a code path shared by *all* issue providers.
- **`dateStrToUtcDate` rewrite**: I couldn't reproduce a rejection on current master — the existing validation round-trips through the same local `Date`, and I confirmed the new and old versions are behavior-identical across thousands of inputs in several timezones (including DST-at-midnight edges). Happy to be wrong — do you have a concrete failing input? If there's a real bug it deserves its own small PR with that repro; as-is it's an equivalent refactor touching a hot, widely-used util.

**2. `oauthTokenKey` duplicates the provider `id`.** `_ensureOAuthTokenKey()` always sets it equal to `model.id`, so it's effectively a copy. The one thing it buys you is distinguishing legacy (unscoped) providers during migration. Could that be a simpler signal (or scope directly by `id` with an explicit legacy flag)? Adding a new synced `pluginConfig` field is a small but permanent data-shape commitment.

**3. Dead migration fallback.** In `_initOAuthAndOptions()`:

```ts
if (!hasTokens && this.isEdit && !tokenKey) {
  hasTokens = await this._pluginBridge.restoreAndCheckOAuthTokens(provider.pluginId);
}
```

When `tokenKey` is `undefined`, the first call already resolves to the same legacy key, so this branch never changes the result. Either remove it or fix it to cover the case it looks like it was meant for (scoped key set but only a legacy token present).

**4. Public plugin-API surface.** Adding `tokenKey?` to `startOAuthFlow` / `getOAuthToken` / `clearOAuthToken` is a permanent, hard-to-reverse addition to `@super-productivity/plugin-api`. It's a reasonable design, but is there a linked issue / demand for multiple Google accounts? I'd like to weigh the permanent API + synced-config cost against the actual need before committing to it.

**Minor:** reconnecting a legacy provider (and deleting a provider instance) can orphan its token in IndexedDB. Low impact, but worth a cleanup on disconnect/delete if easy.

To be clear, the single-account safety and the overall approach look good — this is mostly about trimming to the essential change and confirming the API commitment. Thanks again!

HanCheo:
Thanks again for taking the time to look through this so carefully. I agreed with your main point that the PR had grown beyond the actual Google Calendar account-scoping change, so I trimmed it down and force-pushed an updated branch.

The PR now focuses only on the Google Calendar OAuth scoping work. I removed the `mergeIssueProviderModelUpdates` change and the `dateStrToUtcDate` change from this branch, squashed the remaining work into one commit, and dropped the synced `pluginConfig.oauthTokenKey` field. Instead, the app now passes the issue provider id as a temporary host-side config value when it calls into the plugin for headers, dynamic options, calendar fetching, and time-block operations.

I also kept the old single-account path working: if an existing provider does not have a scoped token yet, it falls back to the old `google-calendar__oauth` token. That also let me clean up the fallback in `_initOAuthAndOptions()` so it now matches the intended migration case instead of being effectively dead code. I added small scoped-token cleanup on disconnect/delete as well.

One implementation detail worth calling out: the Google Calendar plugin still needs a way to ask the host for the scoped token while `getHeaders()` is running. I kept that optional token-key path internal to the runtime bridge and avoided exposing it in the published `@super-productivity/plugin-api` types or storing a new synced config field.

I re-ran the focused checks plus the pre-push hook after the cleanup:

- `npm run checkFile` for the changed TypeScript files
- OAuth bridge/service specs
- Google Calendar provider tests and build
- dialog edit provider component spec
- plugin-api build
- full pre-push lint/tests

The only lint output left is the same unrelated warnings in existing spec files; there are no errors.

johannesjo:
Thanks for trimming and force-pushing the PR — I re-reviewed the new squashed head (`c99b2a27`) from scratch and ran the focused verification locally:

- Google Calendar provider: **40/40 tests passed**
- OAuth bridge/service + provider-dialog specs: **49/49 passed**
- `@super-productivity/plugin-api` build: **passed**

The existing tests pass, but they do not cover several lifecycle/compatibility cases that still look blocking:

### 1. Scoped providers can silently fall back to the wrong Google account

In `google-calendar-provider/src/plugin.ts#getHeaders()`, any missing scoped token falls back to the unscoped legacy token:

```ts
getOAuthToken(providerId) ?? getOAuthToken()
```

Consider the normal migration state:

1. Existing provider A keeps the legacy unscoped token for account A.
2. The user adds provider B and connects account B under B's scoped key.
3. B is disconnected, or B's token refresh fails.
4. B's next request finds no scoped token and silently authenticates with A's legacy token.

Because `disconnectOAuth()` only clears B's scoped key and does not disable/remove the provider, background fetches and auto time-block writes can then operate on account A. With the common `primary` calendar setting this can write events to the wrong account rather than merely failing.

The legacy fallback therefore needs an explicit “this provider is legacy” distinction or a one-time migration of legacy credentials into scoped keys. “Scoped token is absent” is not sufficient to identify a legacy provider.

### 2. Scoping is applied to every OAuth issue-provider plugin, but only Google knows how to read it

`DialogEditIssueProviderComponent.connectOAuth()` passes the provider ID as `tokenKey` whenever the provider has an OAuth button. That stores the token under `pluginId__oauth__providerId`.

However, the published plugin API still exposes only:

```ts
getOAuthToken(): Promise<string | null>
```

An existing uploaded OAuth issue-provider plugin will therefore connect successfully into a scoped key, then call the documented no-argument method and read the unscoped key, receiving no token. The transient config field does not help unless the plugin has Google Calendar's private knowledge of `__spOAuthTokenKey`.

Please gate scoping to Google Calendar, or add an explicit internal capability/opt-in so existing plugins retain unscoped behavior.

### 3. Provider deletion only cleans credentials on the originating device

The new cleanup is inside the dialog's local `remove()` path. Provider deletion is synced, while OAuth credentials are intentionally device-local. When device A deletes a provider, device B applies the remote deletion without ever invoking this component, leaving `pluginId__oauth__providerId` permanently orphaned and with no remaining UI through which to clear it. Other deletion paths have the same issue.

Cleanup needs a centralized lifecycle path that covers both local and remotely applied provider deletions without creating another synced operation. A normal `LOCAL_ACTIONS` effect alone would still miss remote replay.

### 4. Prefix deletion is not bounded to one plugin ID

`clearOAuthTokens(pluginId)` deletes every key beginning with `${pluginId}__oauth`. Plugin manifest validation only reserves `:`, so `foo` and `foo__oauth__bar` are both valid plugin IDs. Clearing `foo` would also delete the second plugin's credentials.

Please delete the exact legacy key plus scoped keys beginning with `${legacyKey}__`, rather than using the legacy key itself as the prefix.

### Documentation / PR description

This is a user-facing provider behavior change, so the repository documentation guide calls for a wiki update. `docs/wiki/4.24-Integrations.md` already documents Google OAuth and would be a natural place to explain per-device, per-provider-account behavior.

The PR description also still describes the removed date parsing/model-merge changes and the earlier public API approach, and its verification counts are stale. Please update it so the PR history accurately describes the current squashed patch.

The narrowed scope is a real improvement, but I think the account-isolation fallback and non-Google compatibility issues need resolving before merge.

HanCheo:
Thanks again — I went through the follow-up points and force-pushed an updated version rebased on the latest `master`.

What changed in this pass:

- Scoped-token misses no longer fall back to the legacy Google token. If a provider has a scoped id and that token is missing, the Google Calendar plugin now fails authentication instead of silently using another account.
- OAuth scoping is now gated to the bundled Google Calendar plugin only. Other OAuth issue-provider plugins keep the existing unscoped token behavior.
- Existing single-account Google Calendar users are handled with a one-time migration from the legacy unscoped token to the only Google provider's scoped key. Once a scoped key has been initialized, it won't re-fallback to the legacy token after disconnect/clear.
- Token cleanup moved out of the dialog-only path. It now also handles synced remote provider deletion through `bulkApplyOperations`, so another device won't keep an orphaned scoped token after the provider disappears via sync.
- Bulk token cleanup now deletes the exact legacy key and scoped keys under `${legacyKey}__`, rather than using `${legacyKey}` as a broad prefix.
- Added the wiki note for per-device/per-provider Google Calendar OAuth behavior.
- Updated the PR description and verification list to match the current patch.

I re-ran the focused checks, and the push hook also completed the full lint/test suite successfully. The only output left is the existing Node/browser warnings and slow-test warnings; no lint or test errors.

HanCheo:
@johannesjo 

Can you do the PR review again?

johannesjo:
Thanks for the rebase — I re-reviewed the current head (`73cdf380`).

First, the merge itself: I checked the resolution and it's clean. `dialog-edit-issue-provider.component.ts` was the only conflicting file, and both sides survived intact (master's `OptionsLoadState` / `_loadAndSetDynamicOptionsState` refactor and your `tokenKey` threading). Nothing was dropped there.

The round-2 points are genuinely improved — the plugin-side fallback is gone (`plugin.ts#getHeaders` now throws instead of reaching for the legacy token), scoping is correctly gated to the bundled plugin, and the prefix-bounded bulk cleanup with its `plugin__oauth-extra` test is a nice touch. But the wrong-account problem still reaches the user, one level higher up.

## Blocking: adding a second Google account silently connects it to the first account

The PR description says the migration happens "when there is exactly one Google Calendar provider". That guard exists in `PluginOAuthLifecycleEffects.migrateSingleLegacyGoogleProvider$` (`filter((ids) => ids.length === 1)`) — but the dialog never goes through it. `_initOAuthAndOptions()` runs in the constructor on **every** dialog open, including a brand-new provider, and calls the migration unconditionally:

```ts
// dialog-edit-issue-provider.component.ts:849-858
const tokenKey = shouldScopePluginOAuth(provider.pluginId)
  ? this._getOAuthTokenKey()
  : undefined;
let hasTokens = false;
if (tokenKey) {
  hasTokens = await this._pluginBridge.migrateLegacyOAuthTokenToScopedKey(
    provider.pluginId,
    tokenKey,
  );
}
```

There's no `isEdit` check and no provider-count check.

**Repro** (precondition: an install that connected Google Calendar before this change, i.e. exactly the users the migration targets):

1. Provider A holds the legacy unscoped token for account A. Boot, or opening A's dialog, migrates it to `google-calendar-provider__oauth__A`. Note the migration **copies** — `migrateLegacyOAuthTokenToScopedKey` does `saveOAuthTokens(scopedKey, serialized)` and never deletes `google-calendar-provider__oauth`, so account A's token is still sitting under the legacy key.
2. The user clicks "+ Add" to connect account B. `model.id` gets a fresh `nanoid()` (line 159), the constructor fires `_initOAuthAndOptions()`, and inside the migration: scoped(newId) is empty → no `__initialized` marker for newId → legacy key **still has account A's token** → it gets copied into the new provider's scoped key and returns `true`.
3. `isOAuthConnected.set(true)`, so per `component.html:115` the "Connect Google Account" button is replaced by "Disconnect / ✓ Connected", and `_loadAndSetDynamicOptionsState()` populates the calendar dropdown with **account A's** calendars.

The user has no way to connect account B from that dialog without first realising they need to click Disconnect. If they don't notice, they save a second provider that reads — and with `isAutoTimeBlock`, writes events into — account A. That's the failure mode from the last round: `getHeaders` correctly refuses to fall back, but the host pre-seeds the scoped key with the legacy token, so the guard never sees a miss.

## Also blocking: "Disconnect" no longer removes the credential

`disconnectOAuth()` changed from `clearOAuthTokens(pluginId)` to `clearOAuthToken(pluginId, tokenKey)`. For a Google provider that deletes only `google-calendar-provider__oauth__<id>`; on master it deleted `google-calendar-provider__oauth`. So after this PR a valid Google **refresh token stays in IndexedDB indefinitely** once the user disconnects, and nothing collects it (`clearOAuthTokens` now only runs on plugin uninstall/disable in `plugin.service.ts`). That's a regression on its own, and it's also the thing that makes the issue above possible.

## Suggested fix — smaller than what's there now

Both of the above disappear if the migration **moves** instead of copying:

- delete the legacy key after `saveOAuthTokens(scopedKey, serialized)` succeeds;
- drop the migration call from `_initOAuthAndOptions()` entirely and let the single-provider effect be the only migration path.

Once the legacy key is gone, its absence *is* the "already migrated" signal, so the whole marker mechanism — `_scopedTokenMarkerKey`, `_isScopedTokenKeyInitialized`, `_markScopedTokenKeyInitialized`, plus the cleanup gaps they create — can go too. That's roughly 40 fewer lines and one fewer thing that can leak. Whatever you choose, please add a spec that fails today: *a new-provider dialog must not report connected when only a legacy token exists*. Nothing currently covers that composition, which is why this survived.

## Other things worth fixing

- **`plugin-oauth-lifecycle.effects.ts:35-58`** — `migrateSingleLegacyGoogleProvider$` is a selector-driven effect whose side effect is writing OAuth credentials, with no `skipWhileApplyingRemoteOps()`. It only escapes the `require-hydration-guard` lint rule because of `{ dispatch: false }`, and that exemption is documented for UI/audio side effects, not for persisting credentials keyed off a transient store shape. During `bulkApplyOperations` / `SYNC_IMPORT` the provider list passes through intermediate states; if it momentarily holds exactly one Google provider, the legacy token gets bound to whichever provider happens to be alone at that instant. Please add the guard.
- **`plugin-oauth-lifecycle.effects.ts:33`** — `inject(ALL_ACTIONS)` in a non-op-log effect. Per `docs/sync-and-op-log/contributor-sync-model.md` that's reserved for the op-log capture effect; remote-driven side effects are supposed to go through a handler. I see why you needed it (remote deletes), so I'd rather discuss the right seam than have you guess — but as written it sets a precedent I don't want to spread.
- **Orphaned credentials with no GC** — connect OAuth in a *new*-provider dialog and then hit Cancel: the token is persisted under `…__oauth__<nanoid>` for a provider that never existed, and `cleanupDeletedGoogleProviderTokens$` only fires on delete actions, so nothing ever removes it. Separately, `clearOAuthToken` deletes `…__oauth__<id>` but leaves `…__oauth__<id>__initialized` behind on every provider delete.

## Minor

- `dialog-edit-issue-provider.component.ts:806-818` — `_getOAuthTokenKey()`'s `|| nanoid()` and the `this.model` mutation are dead code: `model.id` is always set at construction (lines 141/159, and the edit branch carries the existing id). Reduce to `this.oauthButtons.length ? this.model.id : undefined` and drop the hidden write — a method that looks like a getter shouldn't mutate the model.
- `plugin-oauth-token-key.util.ts:2` — `PLUGIN_OAUTH_LEGACY_FALLBACK_CFG_KEY` is declared and never referenced anywhere in the tree. Please remove it.
- `plugin-sync-adapter.service.ts:84` — `pluginId = ''` as a default silently disables scoping for any future caller that forgets the argument. Both call sites already pass it, so make it required.
- `time-block-sync.effects.ts:432` — this was `config: { ...provider.pluginConfig }` and is now `config: pluginConfig`, which for non-scoped plugins is the store object itself (`withPluginOAuthTokenKey` returns its input unchanged). CalDAV also implements `timeBlock` and also runs with `"iFrame": false`, so plugin code can now mutate NgRx state directly. Easiest fix: have `withPluginOAuthTokenKey` always return a new object.
- **API surface** — `packages/plugin-api/src/types.ts` still declares `getOAuthToken(): Promise<string | null>` while the runtime `PluginAPI` accepts `tokenKey?`, and `__spOAuthTokenKey` is a magic key threaded through `pluginConfig`. I'm fine with keeping it out of the published types, but please add a short comment at both ends stating it's an internal host↔bundled-plugin contract and not supported for third-party plugins — otherwise it becomes de-facto public the first time someone finds it, and that's hard to walk back.

Sorry to send you round the loop a third time — the shape of the solution is right, and the copy-vs-move change plus dropping the dialog migration should collapse most of the above at once.

johannesjo:

Thanks for your patience with this one. This is a follow-up pass at `73cdf38`, which is unchanged since the 2026-07-21 review above. I re-verified that review against the source rather than restating it: both blocking items stand exactly as described, and the suggested copy-to-move migration change plus dropping the dialog migration call still collapses most of the list at once. Below is a one-line confirmation of the blockers plus only what this pass found that is new.

### Blocking (unchanged from the 2026-07-21 review)
- **`src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.ts:844-859`** – the constructor migration still copies the legacy token into every new provider's scoped key, so adding a second Google account still silently connects it to the first account.
- **`src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.ts:401-408`** – Disconnect still deletes only the scoped key; the legacy refresh token stays in IndexedDB indefinitely.

### Should fix before merge
- **`dialog-edit-issue-provider.component.ts:213,844-859`** – the copy-migration also leaks without any OAuth interaction: merely opening "+ Add > Google Calendar" on an install with a legacy token writes a full copy of the refresh token under a fresh `nanoid()` (plus its `__initialized` marker) from the constructor, and Cancel leaves it orphaned forever. Every reopen mints another copy: open the add dialog five times and cancel each, and IndexedDB holds five live refresh-token copies no provider references and no cleanup path reaches – without a single Connect click. The move-not-copy fix covers this too, so it costs nothing extra; stating it because the leak needs no OAuth flow at all.
- **`src/app/plugins/oauth/plugin-oauth-lifecycle.effects.ts:37-59`** – one correction to the earlier framing of the sync-window issue: a batch is applied in a single `bulkApplyOperations` dispatch (`operation-applier.service.ts:133`), so there are no intra-batch intermediate states. The reachable failure is the net post-apply state: a device holds the legacy token for account A and provider A; a remote batch (or `SYNC_IMPORT`) deletes A and adds a different Google provider B, the sorted id list goes `['A']` to `['B']`, `length === 1` passes the filter, and account A's legacy token is silently bound to B – which then authenticates (and with `isAutoTimeBlock`, writes) as account A. The guard is still needed; since `distinctUntilChanged` sits upstream, use the waiting variant of the sync-window guard rather than the dropping one.
- **`src/app/plugins/oauth/plugin-oauth-lifecycle.effects.ts`** – ships with no spec file despite persisting credentials keyed off store state; nothing in the PR would catch either case above. Please add coverage alongside the fix, including the earlier-requested composition test: *a new-provider dialog must not report connected when only a legacy token exists*.

### Nits
- **`src/app/features/calendar-integration/time-block/time-block-sync.effects.ts:432`** – softer than earlier stated: the `getHeaders` line on this same path already passed the raw store object to plugin code before this PR, and the only other `timeBlock` implementer (CalDAV) only reads its config – so this is not a new exposure. Still worth having `withPluginOAuthTokenKey` always return a new object as cheap hardening.
- **`src/app/plugins/issue-provider/plugin-sync-adapter.service.ts:84`** – when making `pluginId` required, note `plugin-sync-adapter.service.spec.ts` has three `createPluginSyncAdapter()` calls that need the extra argument.

One earlier concern I can confirm properly resolved: `_loadDynamicOptions` rebuilds the model from the original `pluginConfig`, not the `__spOAuthTokenKey`-augmented copy, so the transient key never reaches synced state.

---

HanCheo:
Thanks for the detailed pass. All items are addressed as of `6245ad647` (branch merged with current master, no conflicts). Item-by-item against the current head:

**Blocking**
- **Dialog constructor migration copies legacy token into every new provider** — removed. `_initOAuthAndOptions()` only calls `restoreAndCheckOAuthTokens` for the dialog's own scoped key; no migration runs from the dialog. Migration now happens once, in `PluginOAuthLifecycleEffects`, and only when exactly one Google provider existed at boot and that same provider still exists after the sync window.
- **Disconnect leaves the legacy refresh token in IndexedDB** — the legacy credential no longer survives startup. `moveOAuthTokens(legacyKey, scopedKey)` is a single IndexedDB readwrite transaction that writes the scoped key and deletes the legacy source; when ownership is ambiguous (0 or 2+ providers at boot, fallback hydration) the legacy credential is deleted and the user reconnects. So Disconnect only ever has the scoped key to clear. Covered by `plugin-oauth-bridge.service.spec.ts` ("disconnect cannot re-migrate a removed legacy token", "second provider cannot inherit a consumed legacy token").

**Should fix**
- **Open "+ Add > Google Calendar" + Cancel leaks a refresh-token copy per open** — no copy is written on open (no dialog migration), and unsaved new/duplicate Google providers have their scoped key cleared on `beforeClosed`/`DestroyRef.onDestroy`, including when the OAuth flow completes after Cancel. Saved providers, edits of existing providers, and non-Google (unscoped) plugins are untouched. The `__initialized` marker mechanism is removed entirely. Covered in `dialog-edit-issue-provider.component.spec.ts` with real IndexedDB assertions; also smoke-tested in the web build with a synthetic legacy token: five open/cancel cycles left only the legacy fixture, no scoped copies.
- **Sync-window: `['A'] -> ['B']` binds A's legacy token to B** — the effect now captures boot provider IDs after local hydration, then uses the waiting variant (`waitForSyncWindow`) before re-reading current state; migration requires `bootIds.length === 1 && currentIds.includes(bootIds[0])`. A -> B deletes the legacy credential; A -> A+B keeps A. Work is serialized with `concatMap` so a pending move cannot land after a deletion. `ALL_ACTIONS` is no longer used; local deletion, remote deletion and full-state import are all handled by observing committed provider state.
- **No spec for the lifecycle effect / composition test** — added `plugin-oauth-lifecycle.effects.spec.ts` (real reducer + `bulkOperationsMetaReducer` + real OAuth bridge/service + IndexedDB): hydration gating, no-owner, fallback hydration, multi-provider boot, A -> B replacement, A -> A (unchanged), A -> A+B, local deletion, remote deletion after replay, full-state import. The composition test ("new-provider dialog must not report connected when only a legacy token exists") is in the dialog spec using the real bridge/store.

**Nits**
- `withPluginOAuthTokenKey` always returns a new object (non-Google path clones too).
- `createPluginSyncAdapter` `pluginId` is required; all spec callsites updated.

Also fixed while adding coverage (reproduced with failing tests first): a refresh that was in flight during Disconnect/reconnect could re-store the old account's access token or invalidate the replacement — `_doRefresh` now only writes/invalidates if its token object is still the current credential.

**Verification:** focused OAuth/dialog/adapter suites 111 passed; full pre-push `lint && test` passed (Angular 15,097 + 15,083 across both TZ runs). Not exercised: a live two-account Google login on desktop — the web build cannot run the Google flow, so account isolation was verified with synthetic credentials.

Ready for re-review when you have a moment.

johannesjo:

Thanks for the updates. Re-checked the 7 item(s) from the earlier review at `73cdf38`, against the code as it stands now rather than the commit messages. All of them landed; one new interaction inside the fix is what keeps this open.

- ✅ **Dialog constructor copied the legacy token into every new provider** – fixed: `dialog-edit-issue-provider.component.ts:882-896` only calls `restoreAndCheckOAuthTokens` for the dialog's own scoped key, `migrateLegacyOAuthTokenToScopedKey` is gone from `PluginBridgeService`, and the only remaining caller is the lifecycle effect (`plugin-oauth-lifecycle.effects.ts:67-76`) under `bootIds.length === 1 && currentIds.includes(bootIds[0])`. The composition test (`dialog-edit-issue-provider.component.spec.ts:306`, real bridge + IndexedDB, legacy token only) asserts disconnected and no scoped copy.
- ✅ **Disconnect left the legacy refresh token in IndexedDB** – fixed: `moveOAuthTokens` (`plugin-oauth-token-store.ts:70-90`) deletes the source in the same readwrite transaction, the effect deletes the legacy key whenever there is no unambiguous owner (`effects.ts:77-83`), and nothing writes the legacy key for Google any more (the only `startOAuthFlow` caller is the dialog, which always passes `tokenKey`). So `disconnectOAuth()` clearing the scoped key is now the whole credential.
- ✅ **Open "+ Add > Google Calendar" + Cancel leaked a token copy per open** – fixed: no migration from the dialog, `_cleanupUnsavedOAuthToken` (`component.ts:891-907`) runs on `beforeClosed` and `onDestroy` for unsaved Google entries, `connectOAuth` cleans up when the flow completes after close (`component.ts:433-436`), and the `__initialized` marker mechanism is removed. Covered for cancel, destroy, late OAuth completion, `submit(true)`, edit dialog and non-Google plugins.
- ✅ **Sync-window: `['A'] -> ['B']` bound A's legacy token to B** – fixed: boot ids are captured before `waitForSyncWindow` (the waiting variant, inside `concatMap` so no emission is cancelled), current ids are re-read after the window, and a replaced boot provider leads to deletion rather than migration (`effects.ts:62-92`). `ALL_ACTIONS` is no longer injected. Spec cases "never rebinds a replaced boot provider" and "preserves the boot account when first sync adds a different Google provider".
- ✅ **No spec for the lifecycle effect / composition test** – `plugin-oauth-lifecycle.effects.spec.ts` (real reducer, `bulkOperationsMetaReducer`, `HydrationStateService`, OAuth service and IndexedDB; 10 cases incl. remote delete after replay and full-state import) plus the composition test in the dialog spec.
- ✅ **`withPluginOAuthTokenKey` should always return a new object** – `plugin-oauth-token-key.util.ts:18`, with a spec.
- ✅ **`pluginId` required in `createPluginSyncAdapter`** – `plugin-sync-adapter.service.ts:84`, all spec call sites updated.

The remaining items from the July review also landed: `PLUGIN_OAUTH_LEGACY_FALLBACK_CFG_KEY` removed, `_getOAuthTokenKey()` reduced to a getter, internal-contract comments at both ends of `__spOAuthTokenKey`, and the wiki note. The merge with master kept #10028's `isTerminalOAuthRefreshError` handling in `_doRefresh` alongside the new stale-credential guard.

### Still blocking
- **`src/app/plugins/oauth/plugin-oauth-bridge.service.ts:131-137`** (new in this push) – `migrateLegacyOAuthTokenToScopedKey` calls `restoreTokens(scopedKey, serialized)` even when nothing was moved: in the steady state (legacy key gone, scoped key present) `moveOAuthTokens` returns the existing `target`, and `restoreTokens` -> `storeTokens` replaces the in-memory token object and drops `_refreshPromises[scopedKey]` (`plugin-oauth.service.ts:238-241`). The new identity guard in `_doRefresh` (`plugin-oauth.service.ts:330-332`) then treats any refresh that was in flight as superseded and returns `null`. This is reachable on every boot for a single-account user: `calendar-integration.service.ts:158` polls with `timer(0, …)` right after data load, so when the stored access token is expired (app closed for more than an hour) the poll's refresh HTTP request and the lifecycle effect's migration run concurrently, the refreshed access token is discarded, and `getHeaders` throws "Not authenticated" for that cycle. It self-heals on the next poll, but it is the fix undoing itself. Cheap fix: return early when `moveOAuthTokens` found an existing target (or when `hasTokens(scopedKey)` is already true) and only `restoreTokens` when the legacy credential was actually moved. A spec for "migration while a refresh is pending keeps the refreshed token" would pin it.

### Nits
- **`dialog-edit-issue-provider.component.ts:891-907`** – the unsaved-token cleanup only covers graceful close/destroy. Connect in a new-provider dialog and then force-quit or reload before closing it, and the scoped key under that nanoid stays in IndexedDB with no owner and no code path that reaches it (the effect only clears ids it saw in the store). A boot-time sweep of `google-calendar-provider__oauth__*` keys whose id matches no provider would close it; fine as a follow-up.

Not verified by me: I did not run the suites, so the 111 focused tests you report are taken from the description.

---

HanCheo:
Thanks — good catch on the steady-state migration. Fixed in `5b4c37eff` (also re-merged with current master; no conflicts left).

**Blocking: `migrateLegacyOAuthTokenToScopedKey` replaced the live token object when nothing was moved**
`plugin-oauth-bridge.service.ts`: after `moveOAuthTokens`, `restoreTokens(scopedKey, …)` now only runs when `hasTokens(scopedKey)` is false. In the steady state (legacy gone, scoped credential already in memory because the calendar poll restored it) the in-memory object is left alone, so a refresh in flight keeps passing the identity guard and its result is stored. Pinned by a new spec in `plugin-oauth-bridge.service.spec.ts` using the real `PluginOAuthService` + `HttpTestingController`: expired scoped token in memory + IndexedDB → `getValidToken` starts the refresh → migration runs while the HTTP request is pending → flush → the refreshed access token is returned and is what `getValidToken` serves afterwards. It failed before the fix (`Expected null to be 'refreshed-a'`).

**Merge note:** master (#10046) independently added the same `_tokenStore.get(pluginId) !== tokens` guard in the success path of `_doRefresh`; I resolved the conflict by taking master's version (identical check plus its log line). The error-path guard from this PR stays alongside it, and the three new upstream refresh specs pass with this branch.

**Nit (orphan sweep on force-quit/reload):** agreed, leaving as a follow-up as you suggested. One reason not to fold it into this PR: a boot-time sweep of `google-calendar-provider__oauth__*` against store IDs would race an "Add > Connect" done during the 30 s sync-window wait and delete a credential that is about to be saved, so it needs its own small design (e.g. sweep only keys older than the current session, or on provider-dialog open).

Verification: focused OAuth/dialog/adapter suites 115 passed (111 + 1 new + 3 from #10046); full pre-push `lint && test` passed (Angular 15,342 / 15,328 across both TZ runs). Live two-account Google login still not exercised on desktop.

Ready for another look.

johannesjo:
**✅ Ready to merge** – no blocking issues at `5b4c37e`. Merge confidence 80% – I read the fix and the new spec but did not run the suites myself (CI is green at this head), and nobody has exercised a live two-account Google login on desktop.

Thanks for the updates. Re-checked the 2 item(s) from the earlier review at `6245ad6`, against the code as it stands now:

- ✅ **Steady-state migration replaced the live token object and discarded an in-flight refresh** – fixed in `src/app/plugins/oauth/plugin-oauth-bridge.service.ts:136-141`: `restoreTokens` now only runs when `hasTokens(scopedKey)` is false, which is the same guard `restoreAndCheckOAuthTokens` and `getOAuthToken` already use, so in the steady state the token object the pending `_doRefresh` holds stays current and its identity check at `plugin-oauth.service.ts:364` passes. The legacy-moved and scoped-key-not-yet-in-memory paths still restore as before. Pinned by "keeps a refresh that is pending while the steady-state migration runs" (`plugin-oauth-bridge.service.spec.ts:476-527`, real `PluginOAuthService` + `HttpTestingController`, refresh request left pending across the migration call).
- ➖ **Orphaned scoped key after force-quit/reload in a new-provider dialog** – deferred to a follow-up, as suggested. Your reason holds: an unsaved provider id is not in the store until Save, so a boot-time sweep keyed on store ids would delete a credential connected during the sync-window wait. It needs its own design.

The master merge is clean for the OAuth service: head vs master differs only by this PR's intended changes (`storeTokens`/`clearTokens` dropping the refresh promise, `clearTokensByPrefix`, the guarded `finally`, and the error-path identity guard), with #10046's success-path guard kept as is. `gh pr diff` still lists the same 23 files.

---

johannesjo:
Thank you very much! <3

Issue #8592, opened 2026-06-25
🚨 Moving tasks between backlog and regular list changes scheduled date


ardenhag:
### Version Used

Windows Desktop v.18.12.0MS

### Current Behavior

Moving a task from a project 'backlog' up to the 'regular list' via the menu or keyboard shortcut, changes the task scheduled date to Today. This happens whether it has a different existing scheduled date or no scheduled date.

Moving a task from a project 'regular list' to the 'backlog' removes the scheduled date if it is set to today, but does not change it for a future or past (overdue) scheduled dates.

Dragging and dropping tasks using the mouse does not change the scheduled date in either direction.

### Expected Behavior

Moving tasks between a project backlog and regular list should keep the existing scheduled date (or lack of) for all methods, in both directions.

### Steps to Reproduce

Enable Project Backlog in either Inbox or another project via their settings.

Moving from Backlog to Regular List 
1. Create a new task in the backlog OR use an existing task in the backlog with an overdue, future or no scheduled date.
2. Right click the task and select 'Move to regular list' OR select the task and press SHIFT+T (the shortcut indicated in the menu)
3. The task moves up to the regular list AND is scheduled for today.

Moving from Regular List to Backlog 
1. Create a task with a schedule date of Today.
2. Right click the task and select 'Move to backlog' OR select the task and press SHIFT+B (the shortcut indicated in the menu).
3. The task moves down to the backlog AND is unscheduled. 
4. Repeat steps 1 & 2 with an overdue or future task scheduled date.
5. The task moves with no change to it's scheduled date.

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

PR #8867, opened 2026-07-09
fix(task): schedule overdue tasks for today via Shift+T and guard stale task focus

### LINKED ISSUE #8851: 🚨 Shift + T shortcut (Add to Today) not working from Overdue list
### Version Used

18.13.1MS

### Current Behavior

The "Shift + T" shortcut, when used from the Overdue list, doesn't update the date for the task. However, right click (or the S key, for "Schedule") then clicking "Today" does work.

Note that some change does still get written to the sync file, since the "Sync" indicator/button on the top-right corner updates to "not synced" when pressing Shift + T.

Extra info: I'm using a Spanish (ES-ES) keyboard layout.

### Expected Behavior

Shift + T used to add task to "Today" list up until the latest update.

### Steps to Reproduce

Described above.

### Console Output

```shell

```

### Documentation

costajohnt:
## Problem

Fixes #8851. Shift+T ("Add to Today") does nothing useful for overdue tasks, and from the Planner overdue list it can silently act on the wrong task.

Two separate causes, as discussed on the issue:

1. Since #8603, `moveToToday()` is a position-only move, and `moveProjectTaskToRegularListAuto` early-returns for any task that is not in the backlog. Overdue tasks are never in the backlog, so the action no-ops (in the normal Today view too, not just the Planner).
2. From the Planner overdue list the shortcut is delegated through a stale `lastFocusedTaskComponent` / `focusedTaskId`, because `planner-task` never registers with `TaskFocusService` and the focused id is only cleared by a `<task>` focusout, never on navigation. That is where the stray sync write comes from.

## Solution

Three parts, matching the approach agreed on the issue:

1. **Overdue action**: `moveToToday()` now branches on "is this task overdue" (new pure util `is-task-overdue.ts`, clock-free, today threaded in via `DateService`). Overdue and not done: schedule for today via `planTasksForToday` (same dispatch as the "Add to My Day" button and Schedule -> Today). Otherwise: the existing #8603 position-only move, unchanged, so #8592 stays fixed. Done tasks with a stale past due date are excluded from the overdue branch so they still take the position-only path. `selectOverdueTasks` now delegates to the same util so the two overdue definitions cannot drift.
2. **Stale-focus guard**: `TaskShortcutService` now treats the DOM as authoritative. If `document.activeElement` is inside a `<task>`, that component wins (this also preserves the inline-edit focus-recovery path). If focus is not inside any `<task>` while a `focusedTaskId` is set, the stale id is dropped instead of firing the shortcut on the wrong task.
3. **Planner wiring (targeted, per your call on the issue)**: `planner-task` carries `data-task-id` and is focusable only in the overdue list (both opt-in via a `focusable` input; I kept them off for the day/scheduled cards so the board does not gain dozens of new Tab stops, and because the e2e page object in `e2e/pages/task.page.ts` selects its done-confirmation strategy based on `data-task-id` presence, which the boards specs rely on). The schedule-today shortcut resolves the id generically from `[data-task-id]` and dispatches `planTasksForToday` by id via a new `TaskService.scheduleForTodayById()`, so it does not depend on a live `<task>`. The id-based path is generic, so extending it to more shortcuts later is straightforward.

A few notes:

- The context menu "Move to Today" has its own separate `moveToToday()` and is intentionally untouched (the agreed scope was Shift+T). Happy to align it in a follow-up if you want.
- Side effect of `data-task-id` on `planner-task` in the overdue list: pasting a markdown task list while an overdue planner card holds focus now adds subtasks to that task instead of creating top-level tasks. Arguably the better behavior, but flagging it since it was not explicitly asked for.
- Tests cover: the overdue/non-overdue/done branches of `moveToToday()`, the stale-focus guard (drop on DOM contradiction, recovery kept), the planner-task id dispatch, the literal #8851 repro (stale id set to a different task while a planner card holds focus), and a date-boundary matrix for `is-task-overdue.ts` including the start-of-next-day offset.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/e2e/pages/task.page.ts
+++ b/e2e/pages/task.page.ts
@@ -53,10 +53,14 @@ export class TaskPage extends BasePage {
    */
   async markTaskAsDone(task: Locator): Promise<void> {
     await task.waitFor({ state: 'visible' });
-    // `data-task-id` is only bound on the <task> host. Wrapper components such
-    // as <planner-task> (Boards) render the same done-toggle but expose no id,
-    // so the done-confirmation strategy is chosen based on its presence.
-    const taskId = await task.getAttribute('data-task-id');
+    // The done-confirmation strategy differs for real <task> rows vs. wrapper
+    // components (<planner-task> etc.) that render the same done-toggle. Key it
+    // off the element actually being a <task>, not off `data-task-id` presence:
+    // <planner-task> also carries `data-task-id` in the Planner overdue list
+    // (#8851), but querying `document.querySelectorAll('task')` for it finds
+    // nothing, so the <task> strategy would hang for a wrapper.
+    const isTaskHost = (await task.evaluate((el) => el.tagName.toLowerCase())) === 'task';
+    const taskId = isTaskHost ? await task.getAttribute('data-task-id') : null;
     await task.hover();
 
     // Give hover effects time to settle
--- a/src/app/features/planner/planner-day-overdue/planner-day-overdue.component.html
+++ b/src/app/features/planner/planner-day-overdue/planner-day-overdue.component.html
@@ -27,6 +27,7 @@
         [cdkDragLockAxis]="isXs() ? 'y' : ''"
         [task]="task"
         [day]=""
+        [focusable]="true"
       ></planner-task>
     }
   </div>
--- a/src/app/features/planner/planner-task/planner-task.component.ts
+++ b/src/app/features/planner/planner-task/planner-task.component.ts
@@ -50,6 +50,13 @@ import { TranslatePipe } from '@ngx-translate/core';
   ],
   /* eslint-disable @typescript-eslint/naming-convention */
   host: {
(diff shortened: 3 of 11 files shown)


johannesjo:
@costajohnt thank you very much! Looks good to merge! Can you have a look at the conflicts?

costajohnt:
Thanks! Rebased on master and resolved the conflict in `task.selectors.ts`.

The overdue selector got refactored to the snapshot/ids approach on master, so I took that version and dropped the now-unused `isTaskOverdue` import from the selector (it's still used directly by the task component). Ready for another look.

johannesjo:
Thank you very mucThank you very much! <3

Maxiaid-ES:
Thank you guys for your work! ❤️

PR #8874, opened 2026-07-09
feat(sync): conflict journal + disjoint-field auto-merge + review UI

aakhter:
Three stacked commits for the sync conflict-review work (each builds on the previous).

## 1/3 — `feat(sync): conflict-journal foundation (observe-only)`
A **device-local IndexedDB journal** that records every sync-conflict auto-resolution so the discarded ("losing") side is preserved and reviewable. **Observe-only** — does not change what LWW picks. New `SUP_CONFLICT_JOURNAL` store (own DB) + `ConflictJournalService` (`unreviewedCount$`, `list`, `markKept`, `markFlipped`, `getEntry`), 14-day/200-entry retention pruned on app start. A pure classifier maps each resolution to the taxonomy (`newer`/`tie`/`delete-wins`/`noise`/`clock-corruption-suspected`; `disjoint-merge` from 2/3), capturing the loser's field values verbatim. The existing conflict-resolution suite stays green with the hook wired in.

## 2/3 — `feat(sync): disjoint-field auto-merge for concurrent edits`
When two clients concurrently edit the **same entity but different fields**, whole-entity LWW previously discarded one side. Now, if the non-noise changed-field sets are disjoint (and neither side deletes/archives), a single merged UPDATE op is synthesized (current entity + the other side's non-noise fields; noise resolved by greater `(timestamp, clientId)`) with a clock dominating both — it propagates through normal sync and is journaled `disjoint-merge`. Non-noise overlap or a delete/archive → the existing LWW path unchanged. No sync-core/protocol change. Both clients compute the byte-identical merged entity and the two merged ops don't re-merge — an end-to-end two-client sim asserts convergence.

## 3/3 — `feat(sync): sync conflicts review UI (banner, badge, page, flip)`
Post-sync summary banner "N auto-resolved (X remote, Y local won)" + REVIEW/DISMISS (replacing the bare snack) + a persistent sync-icon badge on `unreviewedCount$`. A `/sync-conflicts` page: Unreviewed | History tabs, rows grouped by entity type with winner/reason chips, expandable per-field LOCAL-vs-REMOTE diff (device name, winner marked), per-row KEEP/FLIP + bulk actions; History renders merged auto-merges as per-field chips. Flip dispatches a normal entity update with the loser's journaled values (syncs like a user edit, no history rewind) with a stale-flip confirm. (Delete-restore / archived-entity flip are surfaced as unsupported for now — follow-up.)

## Tests
Existing `conflict-resolution.service.spec` **138/138** and `superseded-operation-resolver.service.spec` **42/42** (no regression). New: conflict-journal store/taxonomy/retention/observe-only specs, the disjoint-merge acceptance suite incl. the two-client convergence sim, and the review-UI util/service/banner/page specs. Rebased onto current `master`.


src/app/op-log/sync/conflict-resolution.service.ts
```diff
@@ -397,6 +434,41 @@ export class ConflictResolutionService {
...
+    //
+    // For each merge we: (1) reject BOTH original sides (the merged op
+    // supersedes them); (2) persist the original remote ops as rejected so they
+    // are recorded-as-seen but not applied (mirrors the local-wins remote-op
+    // bookkeeping); (3) append the synthesized merged op as a PENDING LOCAL op
+    // (so it uploads on next sync) AND queue it into the apply batch (so THIS
+    // client's state picks up the remote side's fields — local's are already
+    // optimistically applied). The op stays unsynced+not-rejected → it uploads.
+    // ─────────────────────────────────────────────────────────────────────────
+    for (const merged of mergedResolutions) {
```
johannesjo, line 459: 🟡 **MEDIUM — merged ops aren't counted in `localWinOpsCreated`, so the same-sync re-upload is skipped.**

This block synthesizes a merged op that must be uploaded, but `autoResolveConflictsLWW` returns `{ localWinOpsCreated: newLocalWinOps.length }` (line 598) — `mergedResolutions` isn't included. That return value drives the immediate re-upload loop in `sync-wrapper.service.ts:553-590`.

Result: a sync whose only conflicts are disjoint merges returns `0`, the re-upload loop is skipped, and the sync reports `IN_SYNC` while the merged op is still pending upload. Since this block also rejects the original local op (`merged.conflict.localOps` → `localOpsToReject`), if that op hadn't already been uploaded the merged op is the *sole carrier* of the local edit — it won't reach the server until a later sync (temporary cross-client divergence; self-heals via `getUnsynced()`, so not permanent loss).

There's precedent that this should count — the rejection-handler path does `localWinOpsCreated += rejectionResult.mergedOpsCreated` (`operation-log-sync.service.ts:361`).

Suggested fix:
```ts
return { localWinOpsCreated: newLocalWinOps.length + mergedResolutions.length };
```
plus a spec asserting the returned count reflects a merge (none currently does).

johannesjo (commented):
## Review — conflict journal + disjoint-field auto-merge

Thorough, unusually well-documented, and well-tested. I ran the full conflict suite locally: **166/166 pass** (138 existing + new, no regression), which confirms the observe-only guarantee and the two-client convergence claims. CI is green apart from the still-pending unit `Tests` job.

Overall this is close to mergeable. I found **one MEDIUM correctness gap** (upload accounting for merged ops — inline below), plus a few minor/architectural notes.

### 🟡 MEDIUM — Merged ops aren't counted in `localWinOpsCreated`, so the same-sync re-upload is skipped
See the inline comment on the STEP 3b block. `autoResolveConflictsLWW` returns `{ localWinOpsCreated: newLocalWinOps.length }` (line 598), which excludes `mergedResolutions`. That value drives the immediate LWW re-upload loop in `sync-wrapper.service.ts:553-590`. A sync whose only conflicts are disjoint merges returns `0`, the re-upload loop is skipped, and the sync reports `IN_SYNC` with the synthesized merged op still pending upload. Because STEP 3b rejects the original local op, if that op wasn't already uploaded the merged op is the *sole carrier* of the local edit — so it won't reach the server until a later sync (temporary cross-client divergence; self-heals via `getUnsynced()`, not permanent loss). The rejection-handler path already sets the precedent: `localWinOpsCreated += rejectionResult.mergedOpsCreated` (`operation-log-sync.service.ts:361`). Fix: `return { localWinOpsCreated: newLocalWinOps.length + mergedResolutions.length };` and add a spec asserting the count reflects a merge.

### 🟢 MINOR — "no re-merge" rationale is slightly off
The comments imply the two merged ops are ineligible because their full-entity payloads *overlap*. Actually `createLWWUpdateOp` emits a **flat** payload (no `task:` wrapper), so `extractUpdateChanges(payload, 'task')` returns `{}` → both sides report zero non-noise fields → `isDisjointMergeEligible` fails on the `length === 0` guard. Same outcome, different mechanism; the `(e2e)` spec asserts the correct behavior so this is documentation-only.

### 🟢 MINOR — Whole-entity atomicity is relaxed across sides
LWW previously kept a losing side's edit atomic; the merge combines A's fields with B's. Safe within a side, and disjointness protects co-written fields (`isDone`+`doneOn` land in one op → same side). The only failure mode is two semantically-coupled fields written by *separate* ops on opposite sides — rare, and `_validateAndRepairAfterResolution` is a backstop. Worth a line in the sync docs noting the atomicity contract changed.

### 🟢 MINOR — Eager IndexedDB creation on every startup
The `APP_INITIALIZER` opens/creates `SUP_CONFLICT_JOURNAL` on every launch, even for users who never sync and have no conflicts, for a feature with no UI consumer yet. Cheap, but could be lazy (open on first `record()`).

### 🟢 MINOR — Corruption attribution relies on `EntityConflict` object identity
`_corruptionSuspectedConflicts` (WeakSet) works only because `planLwwConflictResolutions` preserves the same `conflict` reference. Correct today; a future refactor that clones conflicts would silently break `clock-corruption-suspected` classification. A one-line comment at the call site would harden it.

### Scope
The two commits carry very different risk: SPAP-13 (journal) is genuinely observe-only; SPAP-14 mutates the most delicate code in the app and bundling means the safe half can't land independently. SPAP-13 also ships currently-dead surface (`markKept`/`markFlipped`/`list`/`getEntry`/`unreviewedCount$` have no production consumer until the review UI). Both defensible as a foundation — just worth confirming the review UI is a committed near-term follow-up.

_Reviewed with Claude Code._

src/app/op-log/sync/conflict-resolution.service.ts
aakhter: Fixed in 23e9a56 (rebased onto the new review-UI commit). `autoResolveConflictsLWW` now returns `newLocalWinOps.length + mergedResolutions.length` — each disjoint merge appends exactly one pending-local op in STEP 3b, so a merge-only sync now returns > 0 and the immediate re-upload fires. Matches the rejection-handler precedent you pointed at.

One correction to the trace, for the record: on this branch the consumer is `immediate-upload.service.ts:244` (`if (result.localWinOpsCreated > 0)` -> `uploadPendingOps`), not `sync-wrapper.service.ts:553-590` — that path no longer exists after the SPAP-11 file split. Same bug, same fix.

Added a regression spec, `conflict-resolution.disjoint-merge.spec.ts` case (a2), asserting a merge-only conflict returns `localWinOpsCreated: 1` — it fails on the old return and passes now. Local run of the full `conflict-resolution.service.spec.ts` plus the disjoint-merge suite: 146/146 green.

aakhter:
Pushed the MEDIUM fix (see the inline thread). Notes on the minors:

- **WeakSet identity (corruption attribution):** added a `FRAGILE:` note on `_corruptionSuspectedConflicts` warning that a future refactor which clones/rebuilds `EntityConflict` between detect (`.add`) and resolve (`.has`) would silently drop the `clock-corruption-suspected` classification.
- **"no re-merge" rationale:** the behavior is covered — `isDisjointMergeEligible` returns false for two flat merged ops (asserted in the two-client round-trip spec). Agree the inline comment's "overlap" framing is imprecise; I'll fold the atomicity-contract note into the sync docs rather than churn the code comment.
- **Eager IndexedDB init:** leaving as-is for now — opening the journal store on `APP_INITIALIZER` is cheap, and lazy-open would couple creation to the first `record()`. Easy to switch if you'd rather, but it read as avoidable coupling for negligible gain.
- **Scope (review UI):** the conflict-review UI has since landed on this branch (the review-UI commit), so that surface is no longer dead — `markKept`/`markFlipped`/`list`/etc. now have a production consumer.

johannesjo:
## Second-pass review — journal taxonomy, merge-guard coverage, review UI

Follow-up to my earlier review, this time with a deeper pass over the journal classifier, the review-UI surface, and test coverage. The merge core holds up under independent tracing: convergence (byte-identical merged entity on both clients) and the no-re-merge property (merged ops emit flat payloads → `extractUpdateChanges` returns `{}` → ineligible) are sound, and the `localWinOpsCreated` fix is confirmed present (`(a2)` asserts it). No new correctness/data-loss defect. A few things worth addressing:

### 🟡 Important — a discarded DELETE is journaled as `noise`/`info` and never surfaced
`conflict-journal-emission.util.ts` — `isDeleteWin` only fires when the **winner** side is a delete/archive. When a delete **loses** to a concurrent newer update, the loser side is a pure `Delete` op; `mergeChangedFields` skips Delete ops → `loserChanges = {}` → `loserRealFields.length === 0` → falls through to `reason: 'noise'`, `status: 'info'`.

Scenario: device A deletes task T, device B edits T concurrently (newer) → LWW resurrects T, and the journal records "no real content lost" — `unreviewedCount$` doesn't increment and the user is never shown their delete was overridden. The symmetric case (delete *wins*) is correctly `unreviewed`, so this is an asymmetry that defeats the review feature for exactly this class. Observe-only (no data/sync impact), but a real classification gap.

Fix: before the `noise` check, detect a loser-side Delete — `(winner === 'local' ? remoteOps : localOps).some(op => op.opType === OpType.Delete)` — and classify as `delete-wins` / `unreviewed` (or a new `delete-lost` reason).

### 🟡 Important (test gap) — archive-vs-disjoint-edit merge exclusion has no real coverage
`isDisjointMergeEligible` does **not** check for archive (an archive is an `UPDATE` op, not a `Delete`); the *only* thing preventing an archived entity from being partially resurrected by a merge is the `_isArchivePlan(plan)` guard in `_tryCreateDisjointMergeOp`. Every existing archive fixture uses ops with no extractable non-noise changes, so eligibility is already `false` regardless of the guard — meaning if `_isArchivePlan` ever regressed, a disjoint edit concurrent with `TASK_SHARED_MOVE_TO_ARCHIVE` would synthesize a merge and **no test would fail**. Worth a title-edit-vs-archive case in `conflict-resolution.disjoint-merge.spec.ts` asserting no merged op is produced and the archive wins the whole entity.

### 🟢 Minor
- **`pruneOnStart` has no try/catch, but its `main.ts` comment claims it "swallows its own errors".** It's the sole startup seeder of the badge count (ends with `_refreshUnreviewedCount`). A transient IndexedDB open failure rejects it → badge stuck at 0 for the session, and the memoized rejected `_initPromise` in `_ensureDb` poisons the service with no reset. Harden with a try/catch (log + return 0, like `record()`) and reset `_initPromise` on failure. (Normal path seeds fine.)
- **`superseded-operation-resolver.service.ts` notification.** The always-on count snack was replaced by `maybeShowSummaryBanner()`, which reads the *global* unreviewed count — but this path doesn't journal, so it's now silent when nothing is journaled and can surface a banner citing unrelated older conflicts. Defensible under "less noise", just confirming it's intended.
- **`sync-conflicts-page.component.scss`** hardcodes `#4caf50` / `rgba(76,175,80,…)` (L172/179) where `--color-success` already exists and is theme-overridden, and uses a dead `--color-warning` fallback `#e6a817` (L108) that doesn't match the real token `#ff9800`. Use the tokens so it adapts across themes. (No Material-internal overrides anywhere — otherwise clean.)
- **Dead i18n key** `LWW_CONFLICTS_AUTO_RESOLVED` remains in `en.json`/`t.const.ts` but is no longer emitted.
- **Badge polish:** `matBadgeColor="warn"` (red) on the always-visible sync icon for a *resolved* (non-error) count is mildly loud vs. the calm principle; and the sync button's `[attr.aria-label]` announces the conflict count while the button actually triggers a sync (minor a11y mismatch).
- **Further test gaps:** the equal-timestamp `clientId` branch of `noiseTiebreakSide` (cross-client determinism) is untested; `buildMergedFieldDiffs` has no direct test; flip on a *deleted* adapter entity and the disjoint-merge journal-throw-swallow are uncovered.

Nothing here blocks the merge core; I'd fix the two Important items (or consciously defer the delete-classification one as a documented follow-up), and would want the pending SuperSync/WebDAV E2E jobs green given this touches the most delicate code in the app.

_Reviewed with Claude Code._

aakhter:
Addressed in 96901cb.

**Important — delete-lost classification.** Fixed. A delete on the *loser* side is a pure `Delete` op, so `mergeChangedFields` skips it → `loserChanges = {}` → it fell through to `noise`/`info`. Added a dedicated `delete-lost` reason (rather than reusing `delete-wins`) — semantically it's the inverse: the entity was *resurrected* and the user's delete overridden, so labeling it "Delete won" would read backwards. Checked before the noise fallthrough, classified `unreviewed`; `delete-wins` still takes precedence for delete-vs-delete. Regression spec added (`delete vs edit (edit wins, delete lost)`), plus the `REASON_DELETE_LOST` i18n key + review-util mapping.

**Important — archive-vs-disjoint-edit test gap.** Added `(d2)` to the disjoint-merge spec. To make it a real guard test, the archive op carries its *own* disjoint non-noise field, so `isDisjointMergeEligible` returns `true` — the test asserts that explicitly — and then asserts no merged op is synthesized and the archive wins the whole entity (`delete-wins`, winner `remote`). So if `_isArchivePlan` ever regressed, this fails (and it's the only thing that would).

**pruneOnStart.** Wrapped in try/catch (log + return 0, matching `record()`'s observe-only contract), and `_ensureDb` now clears the memoized `_initPromise` on open failure so a transient IndexedDB error can't wedge the service for the session.

**scss tokens.** `--color-success` (+ `color-mix` for the 14% tint) replaces the hardcoded `#4caf50`/`rgba(76,175,80,…)`; dropped the dead `--color-warning` fallback.

**Badge.** `warn`→`accent`. Also: the old `[attr.aria-label]` was `null` at count 0, leaving the icon-only button with no accessible name at all — so I moved the count onto `matBadgeDescription` (the badge's own aria) and gave the button a stable sync-action `aria-label` from `syncTooltip()`. Button announces its action; badge announces the count.

**Dead i18n key.** Removed `LWW_CONFLICTS_AUTO_RESOLVED` from `t.const.ts` + `en.json`. Left the ~26 translation-managed locale JSONs alone — orphan strings with no `en` key are inert and Weblate prunes them; hand-editing them here would just bloat the diff. Say the word if you'd rather I strip them all.

**superseded-operation-resolver banner.** Confirming it's intended: the global-unreviewed-count read is deliberate ("less noise"), so no change there.

**Test gaps.** Added a direct `conflict-disjoint-merge.util.spec.ts` covering `noiseTiebreakSide` (including the equal-timestamp `clientId` determinism branch, with a symmetry assertion that both clients pick the same physical side) and `buildMergedFieldDiffs`. The deeper ones you listed (flip on a deleted adapter entity, the disjoint-merge journal-throw-swallow) I left for a follow-up — happy to fold them in if you want them in this PR.

Verification: affected specs green (222 in the conflict suite + 9 for the two touched components), `eslint`/`stylelint` clean. The SuperSync/WebDAV E2E jobs are your call given where this sits.

johannesjo:
## Third-pass review — journal fidelity, Flip correctness, and profile isolation

I re-checked the current head (`df081f8`) and turned the non-obvious concerns into focused regression specs in an isolated checkout. The fixes described above are present, including `delete-lost`, the archive guard test, retryable journal initialization, token/badge polish, and helper coverage. I found three blocking correctness gaps plus one isolation issue:

### 🔴 Blocking — the journal does not understand several production action payload shapes

`mergeChangedFields()` delegates to `extractUpdateChanges(payload, payloadKey)`, which only reads `actionPayload[payloadKey]`. Many persistent update actions encode their actual mutation elsewhere. For example, `convertToSubTask` persists:

```ts
{ taskId, targetParentId, afterTaskId }
```

rather than `{ task: { id, changes } }`, and `OperationCaptureService` emits `entityChanges: []` for it. A real lost `convertToSubTask` operation therefore produces `loserChanges = {}` and is journaled as `noise` / `info`, even though a structural task move was discarded. Scheduling, ordering, advanced-config, and other domain-specific actions have similar non-canonical shapes.

A regression spec using the production-shaped wrapped payload fails with:

```
Expected 'noise' not to be 'noise'.
Expected 'info' to be 'unreviewed'.
```

This means the journal currently cannot support the claim that discarded changes are preserved/reviewable across operation types. The merge core mostly falls back to LWW when extraction is empty, but the journal/UI silently hide real losses. The fix needs a normalized per-entity before/after delta source (or explicit action-specific extraction), rather than assuming all update actions are adapter-shaped.

Relevant code: `conflict-disjoint-merge.util.ts:31-49`, `packages/sync-core/src/operation.types.ts:256-269`, and `operation-capture.service.ts:23-31, 145-159`.

### 🔴 Blocking — Flip cannot distinguish “field absent on this side” from “this side set undefined”

The journal builds the union of both sides' changed fields and stores a missing value as `undefined`, without recording whether each side actually changed that field. `loserChangesFor()` then blindly emits the losing value for every diff.

Concrete case:

- local loser changed only `title`
- remote winner changed `title` and `notes`
- the journal stores `localVal: undefined` for `notes`
- Flip dispatches `{ title: 'Local title', notes: undefined }`

That clears an unrelated winner-only field (and produces a non-Pure-JSON op payload) instead of layering the discarded local edit onto current state. The focused helper spec fails because the returned object has an own `notes: undefined` property.

Please preserve per-side presence (for example `localChanged` / `remoteChanged`) and only dispatch fields actually changed by the losing side. If “switch to the entire historical side” is intended instead, the journal also needs the unchanged base value; `undefined` is not that value.

Relevant code: `conflict-journal-emission.util.ts:156-166`, `conflict-journal.model.ts:61-71`, and `sync-conflict-review.util.ts:48-64`.

### 🔴 Blocking — the newly surfaced `delete-lost` row offers a false-success Flip

`canFlip()` checks only whether the entity type is TASK/PROJECT/NOTE/TAG. For `delete-lost`, the resurrected entity exists but `fieldDiffs` is empty. `flip()` therefore:

1. builds an empty update,
2. sees the current entity exists,
3. dispatches nothing,
4. marks the entry `flipped`,
5. returns `applied`.

The focused service spec fails with:

```
Expected 'applied' to be 'unsupported'.
Expected 'flipped' to be 'unreviewed'.
```

So the classification fix makes the loss visible, but the primary action claims to restore the delete without deleting anything. At minimum, make capability reason/field-aware and disable `delete-lost` / `delete-wins` until domain-specific delete/restore semantics exist. The same review is needed for relationship-bearing fields such as `projectId`, `subTaskIds`, and `taskIds`; generic adapter updates can bypass the multi-entity/meta-reducer invariant.

Relevant code: `sync-conflict-ui.service.ts:48-66, 78-112, 163-191`.

### 🟡 Important — journal contents cross isolated user-profile boundaries

The journal uses one global `SUP_CONFLICT_JOURNAL` database; entries contain no profile/scope identifier, and `list()` returns every entry. Profile switching replaces the active app dataset and sync configuration but does not clear or namespace this database. After a switch, the new profile can therefore see the prior profile's task titles/notes and badge count, and can attempt Flip against the wrong dataset.

That conflicts with the profile contract (“complete, isolated instance”). Please scope entries/storage to the active profile (including a migration/default scope), or deliberately clear the journal on profile transitions.

Relevant code: `conflict-journal.model.ts:74-90, 132-143`, `conflict-journal.service.ts:56-71, 115-125`, and `user-profile.model.ts:1-5`.

### Required follow-through / verification

- The repo documentation guide requires same-PR wiki updates for this new user-facing page, device-local user-content store, retention/backup behavior, and changed sync-conflict semantics. No `docs/` files are currently changed. The earlier atomicity-contract documentation note also remains outstanding.
- New code in `sync-conflict-ui.service.ts:203-208` adds `as any`, contrary to the repository's strict-TypeScript rule.
- Main CI, WebDAV, server tests, CodeQL, preview, and packaging are green. The SuperSync gate is still red on `LWW: Multiple operations on same entity use max timestamp`, but the recorded failure is a task-button click timeout before conflict resolution starts, so I would treat it as a likely E2E flake and rerun it rather than as evidence for or against the merge logic.

I would keep this at **request changes** until the three blocking cases have regression coverage and pass.

aakhter:
Third-pass items addressed in 82fe84d. Each blocking case has a regression spec that failed against `df081f8` before the fix (same failure signatures you reported).

### 🔴 delete-lost false-success Flip

`canFlip()` is now reason/field-aware: it refuses `delete-lost`/`delete-wins` (a normal update op cannot re-apply a delete or resurrect an entity — deferred until delete/restore semantics exist), refuses entries with no re-appliable loser values, and refuses loser changes touching relationship-bearing fields (`projectId`, `parentId`, `subTaskIds`, `tagIds`, `taskIds`, `backlogTaskIds`, `noteIds`) since a bare adapter update would bypass the multi-entity meta-reducer invariants. `flip()` guards on `canFlip` and the "empty changes → still mark flipped" branch is gone: an entry is only marked `flipped` when an op was actually dispatched. Specs: `sync-conflict-ui.service.spec.ts` — delete-lost returns `unsupported` and stays `unreviewed`, ditto empty diffs and relationship fields.

### 🔴 field absent vs side-set-undefined

`ConflictJournalFieldDiff` now records per-side presence (`localChanged`/`remoteChanged`), written by both the LWW union diff and `buildMergedFieldDiffs`. `loserChangesFor` only emits fields the losing side actually changed, so Flip dispatches `{ title }` and never `{ notes: undefined }`; `winnerChangesFor` gets the same treatment, which also fixes a latent stale-guard bug (a loser-only field compared `undefined` against the live entity and flagged every such entry stale). Entries persisted before the flags fall back to value-presence — exact for that data, since op payloads are pure JSON and cannot encode a real `undefined`. Specs in `sync-conflict-review.util.spec.ts` (own-property assertions) plus a service-level spec on the dispatched action payload.

### 🔴 non-adapter action payloads

Per-op extraction now has two delta sources: the adapter-shaped payload, falling back to the capture-time `entityChanges` (covers TIME_TRACKING / `syncTimeSpent`). Ops with no readable delta from either source (your `convertToSubTask` case, and the scheduling/ordering/advanced-config tail) are treated as **opaque** rather than "changed nothing":

- an opaque loser is never classified `noise`/`info` — it surfaces as `unreviewed` (your production-shaped regression spec is in `conflict-journal.service.spec.ts` and passes);
- the raw action payload is preserved verbatim in the entry as a `kind: 'action'` field diff (field = action type), so the discarded structural change stays reviewable after the ops are gone; these diffs are excluded from flip/stale computations, and Flip reports `unsupported` for them rather than pretending;
- a side with opaque ops is never disjoint-merge eligible. This goes beyond the journal fix, deliberately: eligibility previously keyed only on extracted fields, so an adapter edit + opaque op on one side could pass the disjoint check and the synthesized merge would silently drop the opaque mutation — and the two clients would synthesize different entities. Spec in `conflict-disjoint-merge.util.spec.ts`.

I chose the generic opaque-detection route over per-action extraction mappings: it fails closed for the entire long tail (surfaced + preserved + merge-ineligible) instead of being correct only for enumerated action types. Field-level fidelity for these actions would need a normalized before/after delta at capture time, which I'd rather not bolt onto this PR.

### 🟡 profile isolation

Went with the deliberate-clear option: `switchProfile` calls the new `ConflictJournalService.clearAll()` (swallow-errors contract, same as `record()`) before the reload, covering both the existing-data and empty-profile branches. Scoping entries by profile id would preserve per-profile review backlogs across switches, but the journal service would need active-profile awareness (DI cycle risk via the sync stack) for an advisory store with 14-day retention — clearing matches the "complete, isolated instance" contract with none of that. `clearAll` has a spec; `switchProfile` itself has no spec (pre-existing: its unconditional `window.location.reload()` makes it karma-hostile).

### Follow-through

- `as any` in `_readCurrentEntity` replaced with a typed narrowing to the `PropsStateSelector<{ id: string }>` union member (all flip-supported types register that shape).
- Docs: new `docs/sync-and-op-log/conflict-journal-and-review.md` (journal contracts incl. the outstanding atomicity note — op-log write and journal write are not atomic, journal is best-effort observe-only; and the no-re-merge property — merged ops emit flat payloads → `extractUpdateChanges` returns `{}` → never re-eligible), registered in the README table. Wiki: `3.06-User-Data.md` (SUP_CONFLICT_JOURNAL store, excluded from backups/exports, retention, profile clearing) and `4.23-Managing-Your-Data.md` (disjoint auto-merge semantics, `/sync-conflicts` review page, Keep/Flip, flip limitations).

Verification: 246 specs across `op-log/sync/*conflict*` + `pages/sync-conflicts-page` green, eslint/prettier clean, tsc clean. Pushed as a regular commit (no force-push), so CI including the SuperSync gate re-runs on the new head.

johannesjo:
## Fourth-pass review — flip short-syntax cascade, expand-throw, restore isolation

Re-checked head `82fe84d`. The third-pass fixes all hold — verified: per-side presence flags on every diff path, `loser/winnerChangesFor` no longer clearing winner-only fields, `delete-lost` classification + precedence, opaque-op gating out of both `noise` and disjoint-merge, `canFlip` reason/field-aware with no false-success, the removed `_readCurrentEntity` `as any`, profile-switch `clearAll`, and the new docs.

The **disjoint-merge core is sound**: convergence (symmetric noise tiebreak), the no-re-merge contract (flat merged payload → `extractUpdateChanges` empty → opaque → refused), and — importantly — the merged op is a `[TASK] LWW Update`, so it routes through `lwwUpdateMetaReducer`, which reconciles the reciprocal relationship invariants (`projectId`/`parentId`→`project.taskIds`, `tagIds`→`tag.taskIds`, `parentId`→parent `subTaskIds`, `dueDay`→`TODAY_TAG`). So the meta-reducer-bypass risk only exists for **flip**, which guards it. No data-loss defect survived verification.

I did find one HIGH and two MEDIUM correctness bugs, all in the review-UI/flip layer.

### 🔴 HIGH — Title flip re-runs short-syntax, silently mutating tags/project/schedule

`op-log/sync/sync-conflict-ui.service.ts:213` — the TASK branch of `_buildUpdateAction` dispatches `updateTask({ task: { id, changes } })` **without `isIgnoreShortSyntax: true`**.

The canonical flip is a rename conflict. Inline title edits go `_taskService.update(id, { title })` → `updateTask({ task: { id, changes: { title } } })` (no `modified` — `handleUpdateTask` adds that in the reducer, so the captured op-changes are `{ title }`). Hence `fieldDiffs = [{ title }]` and `loserChangesFor` returns exactly `{ title }`. `short-syntax.effects.ts:56` fires on `updateTask` precisely when `changes` is `{ title }` only, in **`'replace'` mode**. If the discarded title still carries short-syntax tokens (`#tag`, `+project`, `@schedule`), flipping it re-parses and **replaces `tagIds`, moves the project, and schedules the task** — cross-entity changes that sync to every device and bypass `FLIP_RELATIONSHIP_FIELDS` (which only inspects the literal `changes` map). That contradicts the file contract ("re-applies the loser's field values, no history rewind"). Auto/merged ops are unaffected — `[TASK] LWW Update` isn't in the effect's `ofType`.

Fix is one line — pass `isIgnoreShortSyntax: true` in the TASK case (+ a regression spec asserting the dispatched action carries the flag).

### 🟠 MEDIUM — Expanding a row throws (unhandled) for a missing TAG/NOTE entity

`sync-conflict-ui.service.ts:237-251` + `sync-conflicts-page.component.ts:106-116`. `_readCurrentEntity` calls `store.select(config.selectById, { id })`, but `selectTagById` (`tag.reducer.ts:164`) and `selectNoteById` (`note.reducer.ts:44`) **throw** on a missing entity — unlike TASK/PROJECT which return `undefined`. The throw rejects before `getStaleState`'s `if (!current)` guard runs, and neither `toggleExpand` nor `flip()` (`:134`) has a try/catch. So expanding a TAG/NOTE row whose entity is gone — including every `delete-wins` TAG/NOTE, which land in the Unreviewed tab — throws an unhandled rejection, the diff/current column never loads, and re-expand re-throws; `flip()` on such a row throws instead of returning `unsupported`.

Fix: wrap `_readCurrentEntity`'s `firstValueFrom` in try/catch → `undefined` (the throwing-selector behavior is the app-wide convention).

### 🟠 MEDIUM — Journal survives a backup restore (isolation gap)

Profile-switch correctly calls `ConflictJournalService.clearAll()` (`user-profile.service.ts:397`), but the other full-dataset-replacement paths don't: `imex/sync/super-sync-restore.service.ts`, `imex/local-backup/local-backup.service.ts`, `imex/file-imex/file-imex.component.ts`. After a JSON/local/SuperSync restore the sync badge keeps the pre-restore count and the review page lists conflicts referencing entities from the replaced dataset (a Flip is blocked by the stale-guard, so no data loss — but it's a stale, confusing, mildly privacy-relevant surface). This is the exact contract the profile-switch clear was added to satisfy — a shared "clear journal on full dataset replace" hook, or the same `clearAll()` in those paths, closes it.

### 🟡 LOW / NIT

- `_readCurrentEntity` cast is unsound for `ISSUE_PROVIDER` (`:248`) — its `selectById` is a `(id,key)=>selector` factory (special-cased in `conflict-resolution.service.ts:1227`), not a props selector, yet it passes `isAdapterEntity`. Expanding an ISSUE_PROVIDER row renders a bogus all-dashes "current" column + spurious stale flag. Display-only (not flippable), but it contradicts the "narrowing … is safe here" comment.
- Schedule/reminder fields (`dueWithTime`/`deadlineDay`/`deadlineWithTime`/`reminderId`) aren't in `FLIP_RELATIONSHIP_FIELDS` — flipping them via bare `updateTask` can produce the mutually-exclusive both-set state and skip reminder create/cancel (lower confidence; different invariant class than the listed fields).
- `loser/winnerChangesFor` don't intrinsically skip `merged` entries (`sync-conflict-review.util.ts:70-84`) — the doc says they do, but `buildMergedFieldDiffs` sets `pickedSide` on every diff, so a merged entry with `modified` changed on both sides yields a non-empty `loserChangesFor`. Not user-reachable today (merged = `info`/History/non-actionable, and `flipAllToSide` excludes `merged`), but latent — an early-return on `entry.winner === 'merged'` + a comment fix would harden it.
- Journaling can wedge for the session if the IndexedDB connection is force-closed/`versionchange`d after a successful open (`_db` isn't nulled). Observe-only, self-heals on restart.
- Summary banner count goes stale after in-page review (badge updates via `unreviewedCount$`, the open banner doesn't) — defensible given the past-tense wording.
- Minor: retention enforced only at app-start (unbounded intra-session growth); `unreviewedCount$` Observable vs the Signals-preferred rule; `matBadgeDescription` bound to `null` at count 0.

### Verdict

Solid, near-mergeable. I'd want the **HIGH (`isIgnoreShortSyntax`)** and the **MEDIUM expand-throw** fixed with regression specs before merge; the restore-isolation gap and the LOW items are reasonable as same-PR polish or a tracked follow-up.

aakhter:
Fourth-pass items addressed in 43d6a78. The HIGH and both MEDIUMs each have a regression spec that failed before the fix.

### 🔴 short-syntax cascade

Verified against `short-syntax.effects.ts` — the trigger condition (`updateTask`, changes exactly `{ title }`, no `isIgnoreShortSyntax`) is precisely the shape a rename-conflict flip dispatches. The TASK branch of `_buildUpdateAction` now passes `isIgnoreShortSyntax: true`; spec asserts the dispatched action carries the flag. Auto/merged ops were unaffected as you noted (`[TASK] LWW Update` isn't in the effect's `ofType`).

### 🟠 expand-throw on missing TAG/NOTE

`_readCurrentEntity` now catches the selector throw and returns `undefined` (the throwing-selector convention treated as "not present" for this read-only lookup). Specs reproduce the real failure mode — a TAG entry against state without the tag rejects through the real `selectTagById` — and assert `getStaleState` resolves `{ isStale: false, current: undefined }` and `flip()` returns `unsupported` with the entry kept `unreviewed`.

### 🟠 restore isolation

Moved the clear to the `BackupService.importCompleteBackup` chokepoint (right after the op-log replacement), which covers all four paths — profile switch, JSON import, local-backup restore, SuperSync restore — and removed the now-redundant call in `switchProfile` (both its branches funnel through the chokepoint). Spec in `backup.service.spec.ts` asserts `clearAll` is invoked during import.

Side note from getting that spec to run: the backup specs require the typia transform, i.e. an install that ran `prepare` (`ts-patch install`). My sandbox had installed with `--ignore-scripts`, which made every `importCompleteBackup` spec fail on `typia.createValidate(): no transform has been configured` until I patched and cleared `.angular/cache` — mentioning it in case anyone else hits the same wall locally.

### 🟡 LOW items — three fixed, four deferred

Fixed:
- **ISSUE_PROVIDER factory selector** — `_readCurrentEntity` special-cases it like `ConflictResolutionService.getCurrentEntityState` does. The regression spec's pre-fix failure was literally `Expected Function to be undefined`, confirming the inner-selector-as-entity rendering you predicted.
- **Schedule/reminder fields** — `dueWithTime`/`deadlineDay`/`deadlineWithTime`/`reminderId` added to the flip blocklist (renamed `FLIP_UNSAFE_FIELDS`, doc comment now describes both invariant classes). I also included `dueDay`: your own meta-reducer analysis lists `dueDay`→TODAY_TAG as a reciprocal invariant that `lwwUpdateMetaReducer` reconciles for LWW ops — a bare `updateTask` doesn't, so it's the same class.
- **merged early-return** — `loser/winnerChangesFor` now return `{}` for `winner: 'merged'` up front, and the doc comment explains why the per-diff `pickedSide` check alone can't exclude merged entries (tiebreak diffs carry one).

Deferred (happy to file a follow-up issue enumerating them, or fold into a fifth pass if you'd rather have them same-PR):
- IndexedDB force-close/`versionchange` wedge (`_db` not nulled) — observe-only, self-heals on restart.
- Banner count staleness after in-page review — past-tense wording keeps it defensible.
- Retention only at app-start; `unreviewedCount$` Observable→Signal; `matBadgeDescription` null at count 0.

Docs updated for the behavior changes (dev doc flip section + wiki 3.06/4.23 now say "cleared on any full dataset replacement", not just profile switch).

Verification: 277 specs green across `op-log/sync/*conflict*`, `pages/sync-conflicts-page`, and `op-log/backup/backup.service.spec.ts`; eslint/prettier/tsc clean. Regular push again, so all CI including the SuperSync gate re-runs on the new head.

johannesjo:
He @aakhter  sorry for taking over without discussing it first. Originally this was meant to be just a minor improvement, but it seems it is quite bit more complicated than that.

aakhter:
No apology needed for taking over — the opposite: thanks for jumping in directly instead of writing a fifth review pass. Having the fixes land as commits with your regression specs moved this faster than another round-trip would have, and all three convergence fixes are in territory where you have the whole-app picture and I plainly didn't.

An apology from my side is due, though: this took four rounds where it should have taken one or two. Looking at the pattern across my sync PRs, the recurring failure was fixing the *named instance* instead of the *class* — projectId → relationship fields → schedule fields → reminder timestamps is one deny-list enumerated across four rounds, and the profile-switch clear that should have been the dataset-replacement chokepoint from the start is the same shape. I've turned that into a pre-submission checklist on my side (consumer-trace every dispatched action, production-shaped test fixtures, enumerate every entry point before fixing one, one high-risk concern per PR — the mid-review UI addition here clearly multiplied rounds).

To pull some weight back: I merged upstream/master (one trivial import conflict in config-page) so the PR is mergeable again, and ran two extra adversarial review passes over the full branch (independent Claude reviewer + a codex pass), verifying anything raised against the code before repeating it. Two findings survived verification — both in the new partial-delta territory, so reporting rather than pushing:

1. **The partial merged delta is not closed under later LWW composition (3-device divergence).** `_convertToLWWUpdatesIfNeeded` converts winners to full-entity LWW updates only for local-DELETE conflicts; a plain update-vs-update winner is applied as-is. Trace: A edits `title`, B edits `notes` → merged delta `{title, notes}` applied on A/B. Concurrent C edits `title` with a newer timestamp. C-vs-merged overlaps on `title` (and the merged op's flat payload is opaque, so no re-merge) → whole-op LWW → C wins → C's *partial* `{title}` op applies on A/B, leaving B's `notes` in their live state — while C, and any fresh replay of the op log (B's original op was superseded by the merged op, which lost), has base `notes`. No surviving op carries B's notes: A/B's live state permanently diverges from C and from replay. Possible directions: refuse the merge when a same-entity concurrent third op is visible, or carry parent-op identity so a later conflict can decompose the merge — your call, it's a design decision.

2. **Merged resolutions are journaled as `info`/"kept both" at plan time** (`_journalMergedResolution(plan)` inside the planning loop), before the merged op is appended or applied. If the append/apply then fails, the journal permanently records a successful merge for exactly the conflict whose preservation failed — the same false-success direction as the round-3 flip issue — and a retry would journal a duplicate. Moving the write after the append (or marking failure) closes it.

Two minor notes from the same passes, below the fix bar but for awareness: a PROJECT flip triggers `project.effects.ts`'s unconditional "Project updated" snack (`updateProject` has no `isSkipSnack`), and `importCompleteBackup` calls `clearAll()` just after releasing the op-log lock, leaving a narrow window where a concurrent resolution's fresh journal entry gets wiped. The i18n keys and the upstream E2EE-hardening merge were checked clean.

The round-4 deferred list (IndexedDB `versionchange` wedge, banner count staleness, app-start-only retention, Signal migration, badge a11y) is still unclaimed — happy to take those as a follow-up PR after this merges, or push them here if you'd rather; whichever avoids stepping on your in-flight work.

aakhter:
**Correction on yesterday's finding 1 (partial-delta / 3-device divergence): refuted after deeper verification — retracting it.**

What both of my review passes missed: `_createLocalWinUpdateOp` builds the local-win op from the **full current entity snapshot** with a clock dominating both sides. In every ordering I could construct, the client holding the newest (winning) side resolves local-win and emits that snapshot, which reaches the other clients as a plain dominating op and reconciles the limb. The scary state — A/B holding `{title: C, notes: B}` after applying C's partial winner while C holds `{title: C, notes: base}` — is real but transient. Net behavior: B's notes are lost to LWW *consistently* on all clients (and journaled as the merged op's discarded fields, so reviewable), not divergently.

Rather than argue it, I wrote a 3-client spec to adjudicate — it passes on `0beba4c` (drops into the top-level describe of `conflict-resolution.disjoint-merge.spec.ts`, uses its `op`/`conflictOf` helpers):

<details><summary>three-client convergence spec (green on current head)</summary>

```ts
  describe('(e2e) three-client: later overlapping edit beats the merged op', () => {
    // Adjudicates the "partial merged delta is not closed under later LWW
    // composition" concern: after a merged op loses whole-op LWW to a newer
    // overlapping edit from a third client, the clients are TRANSIENTLY apart
    // (the remote-win side keeps the merged delta's other field; the local-win
    // side never applied it) — convergence depends entirely on the local-win
    // side emitting a FULL-SNAPSHOT reconciling op. This spec pins that
    // property: if _createLocalWinUpdateOp ever becomes a partial delta, the
    // three clients diverge permanently and this test goes red.
    const resolveCapturing = async (
      clientId: string,
      currentState: Record<string, unknown>,
      conflict: EntityConflict,
    ): Promise<{ appended: Operation[]; applied: Operation[] }> => {
      TestBed.resetTestingModule();

      const store = jasmine.createSpyObj('Store', ['select']);
      store.select.and.returnValue(of(currentState));

      const applier = jasmine.createSpyObj('OperationApplierService', [
        'applyOperations',
      ]);
      applier.applyOperations.and.resolveTo({ appliedOps: [] });

      const opLogStore = jasmine.createSpyObj('OperationLogStoreService', [
        'appendBatchSkipDuplicates',
        'appendWithVectorClockUpdate',
        'markApplied',
        'markRejected',
        'markFailed',
        'getUnsyncedByEntity',
        'mergeRemoteOpClocks',
      ]);
      opLogStore.mergeRemoteOpClocks.and.resolveTo(undefined);
      opLogStore.getUnsyncedByEntity.and.resolveTo(new Map());
      opLogStore.markRejected.and.resolveTo(undefined);
      opLogStore.markApplied.and.resolveTo(undefined);
      opLogStore.markFailed.and.resolveTo(undefined);
      opLogStore.appendWithVectorClockUpdate.and.resolveTo(1);
      opLogStore.appendBatchSkipDuplicates.and.callFake((ops: Operation[]) =>
        Promise.resolve({
          seqs: ops.map((_, i) => i + 1),
          writtenOps: ops,
          skippedCount: 0,
        }),
      );

      const validate = jasmine.createSpyObj('ValidateStateService', [
        'validateAndRepairCurrentState',
      ]);
      validate.validateAndRepairCurrentState.and.resolveTo(true);

      const effects = jasmine.createSpyObj('OperationLogEffects', [
        'processDeferredActions',
      ]);
      effects.processDeferredActions.and.resolveTo();

      TestBed.configureTestingModule({
        providers: [
          ConflictResolutionService,
          { provide: Store, useValue: store },
          { provide: OperationApplierService, useValue: applier },
          { provide: OperationLogStoreService, useValue: opLogStore },
          {
            provide: SnackService,
            useValue: jasmine.createSpyObj('SnackService', ['open']),
          },
          { provide: ValidateStateService, useValue: validate },
          { provide: OperationLogEffects, useValue: effects },
          {
            provide: CLIENT_ID_PROVIDER,
            useValue: { loadClientId: () => Promise.resolve(clientId) },
          },
          { provide: ENTITY_REGISTRY, useValue: buildEntityRegistry() },
        ],
      });

      const svc = TestBed.inject(ConflictResolutionService);
      await svc.autoResolveConflictsLWW([conflict]);

      const appended = opLogStore.appendWithVectorClockUpdate.calls
        .allArgs()
        .map(([o]) => o as Operation)
        .filter((o) => o.entityId === 'task-1' && o.opType === OpType.Update);
      const applied = applier.applyOperations.calls
        .allArgs()
        .flatMap(([ops]) => ops as Operation[])
        .filter((o) => o.entityId === 'task-1');
      return { appended, applied };
    };

    /** Applies one op's payload to a state map, mirroring the two consumer
     *  paths: adapter `{ task: { id, changes } }` -> merge changes;
     *  flat LWW payload -> shallow-merge all fields minus id (updateOne). */
    const applyOp = (
      state: Record<string, unknown>,
      o: Operation,
    ): Record<string, unknown> => {
      const p = o.payload as Record<string, unknown>;
      const task = p['task'] as Record<string, unknown> | undefined;
      if (task && typeof task['changes'] === 'object') {
        return { ...state, ...(task['changes'] as Record<string, unknown>) };
      }
      const flat = { ...p };
      delete flat['id'];
      return { ...state, ...flat };
    };

    const pick = (s: Record<string, unknown>): Record<string, unknown> => ({
      title: s['title'],
      notes: s['notes'],
    });

    it('reconverges all clients via the full-snapshot local-win op (loss is consistent, not divergent)', async () => {
      // Round 1: A (title) and B (notes) conflict -> disjoint merge on A.
      const opA = op({
        id: 'op-A',
        clientId: 'clientA',
        vectorClock: { clientA: 1 },
        timestamp: 2000,
        payload: { task: { id: 'task-1', changes: { title: 'A-title' } } },
      });
      const opB = op({
        id: 'op-B',
        clientId: 'clientB',
        vectorClock: { clientB: 1 },
        timestamp: 3000,
        payload: { task: { id: 'task-1', changes: { notes: 'B-notes' } } },
      });

      const r1 = await resolveCapturing(
        'clientA',
        { id: 'task-1', title: 'A-title', notes: 'base' },
        conflictOf([opA], [opB]),
      );
      const mergedOp = r1.appended[0];
      expect(mergedOp).toBeDefined();

      // A's state after applying the merged delta.
      let stateA = applyOp({ id: 'task-1', title: 'A-title', notes: 'base' }, mergedOp);
      expect(pick(stateA)).toEqual({ title: 'A-title', notes: 'B-notes' });

      // Concurrent third client C: overlapping title edit, NEWER timestamp,
      // clock concurrent with everything (C never saw opA/opB/merged).
      const opC = op({
        id: 'op-C',
        clientId: 'clientC',
        vectorClock: { clientC: 1 },
        timestamp: 4000,
        payload: { task: { id: 'task-1', changes: { title: 'C-title' } } },
      });
      let stateC: Record<string, unknown> = {
        id: 'task-1',
        title: 'C-title',
        notes: 'base',
      };

      // Round 2 on A: local merged op vs remote opC. Fields overlap on title
      // and the merged op is flat/opaque -> no re-merge -> whole-op LWW ->
      // opC (newer) wins -> A applies opC AS-IS (partial).
      const r2a = await resolveCapturing('clientA', stateA, conflictOf([mergedOp], [opC]));
      expect(r2a.appended.length).toBe(0); // remote win: no new op from A
      for (const o of r2a.applied) {
        stateA = applyOp(stateA, o);
      }

      // TRANSIENT LIMB (documented, not asserted as final): A now holds
      // { title: C, notes: B-notes } while C holds { title: C, notes: base }.
      expect(pick(stateA)).toEqual({ title: 'C-title', notes: 'B-notes' });

      // Round 2 on C: local opC vs remote merged op -> LOCAL win -> C must
      // emit a reconciling local-win op carrying its FULL entity snapshot.
      const r2c = await resolveCapturing('clientC', stateC, conflictOf([opC], [mergedOp]));
      for (const o of r2c.applied) {
        stateC = applyOp(stateC, o);
      }
      const localWinOp = r2c.appended[0];
      expect(localWinOp).toBeDefined();
      // Full snapshot, not a partial delta: it must carry notes (=base), the
      // field opC never touched — this is the closure property. If this op
      // were partial, nothing would ever reconcile A's B-notes with C.
      const localWinPayload = localWinOp.payload as Record<string, unknown>;
      expect('notes' in localWinPayload).toBe(true);

      // Propagation: the local-win op's clock dominates both sides, so it
      // reaches A as a plain (non-conflicting) remote op and applies directly.
      stateA = applyOp(stateA, localWinOp);

      // CONVERGENCE: every client ends on the identical entity. B-notes are
      // lost to LWW — but CONSISTENTLY on all clients (and journaled by the
      // round-2 resolutions for review/flip), never as silent divergence.
      expect(pick(stateA)).toEqual(pick(stateC));
      expect(pick(stateA)).toEqual({ title: 'C-title', notes: 'base' });
    });
  });
```

</details>

Two things it surfaces that may be worth keeping:

- It pins the **closure property**: convergence after a merged op loses depends entirely on the local-win op remaining a full snapshot. If that ever becomes a partial delta — a plausible future step, given the merge path just moved to partial deltas for exactly analogous reasons — three clients diverge permanently and this spec goes red. It's green, so I can push it to the branch as-is if you want it; otherwise I'll carry it in a follow-up.
- One residual edge I did not chase to ground: an exact-timestamp **tie** between the merged op and the overlapping edit resolves remote-win on *both* holders, so no reconciling snapshot is emitted and each side applies the other's partial (title swaps, merged notes survive on one side only). Same-field ties between two plain partial ops look like they have the same shape, so this reads as a pre-existing property of plan-level LWW ties rather than anything merge-specific — flagging, not claiming.

Also confirmed `0beba4c` covers yesterday's finding 2 exactly (journal only after durable append, + the never-throw contract and the terminated-hook recovery).

Issue #4645, opened 2025-06-25
💡 Select and change multiple tasks


ppestana-supervisor:
### Problem Statement

I often need to bulk re-schedule tasks that were left-over and having to do it one by one is cumbersome in this day and age.

### Possible Solution

Shift and option click enabled to select several tasks, the contextual menu opens, similar to editing individual task and parameters like  "Time", "Planned at" and "Add Tag" are editable and changes are reflected on all selected tasks.

### Describe alternatives you've considered

Similar to "Add to today" when items are overdue, "today's" items could have a batch reschedule button below. This seems more fringe user-case, but what typically happens to me is that I have a bunch of low importance tasks which I'll constantly be moving to the next day when I replan current day. This means that each start of day I have all my overdue tasks which I need to "add to today", but prior to that I could change my today tasks (which I had moved over yesterday) to tomorrow

alliyya:
My use case is a bit similar and maybe it's already possible been trying to see in the documentation. 

Feature Request: Select multiple tasks in Schedule view and move them around. (Would love to use shift or cmd to multi-select and drag them to the appropriate time)

Often I find I have a task/meeting that ran longer than expected or need to add a task in between and need to reshift things around.

PS. Thanks for the great work, it's been really a joy to use and has been really useful to me! I hope to one day pay you back and contribute to this awesome project! Really appreciate it, staying open source ❤️

DeludedPenguin:
Would love to see this feature added -- I frequently need to bulk change due dates, tags, projects etc

Love the app -- keep up the great work!

Paul-kingtiger:
I've just started using Super Prod, and am moving over from MS To Do.  Part of that process is looking how I organise things and coming up with a better workflow to make the best use of the tool.

That means I'm changing my mind a lot as I try things.  Being able to select multiple tasks and edit things like tags would be a HUGE time save.  Doing everything one at a time is annoying enough that I'm putting it off, rather than getting it done.

crashmoney:
I just began using Super Productivity the day before yesterday, and have a similar experience with Paul. I'm missing this feature and the ability to merge tags; likewise, I tend to put re-organisation and planning off as it's cumbersome. Otherwise it's looking like a wonderful app for me that replaces both my todo list, time blocking workflow, and time tracker!

carlosmu:
It would be very useful

aramodi:
I agree it would be very useful

unrockstar1988:
[...]

PR #8882, opened 2026-07-09
feat(plugins): add Todoist import plugin with Import/Export launcher

johannesjo:
## What

One-time, **additive** migration of a user's active Todoist data (projects, tasks, sub-tasks, labels, due dates, durations, comments) into Super Productivity — shipped as a **bundled plugin** (`packages/plugin-dev/todoist-import/`) with a single launcher button in Settings → Sync & Backup → Import/Export. Existing SP data is never changed or removed.

Design rationale, verified host constraints, and all mapping decisions are documented in [`docs/plans/2026-07-09-todoist-import.md`](../blob/feat/what-do-you-think-would-be-the-best-way-d67291/docs/plans/2026-07-09-todoist-import.md).

## How it works

- The user pastes a Todoist API token (password field; sent only to `api.todoist.com` via gated `PluginAPI.request` with `permissions:["http"]` + `allowedHosts`; **never stored**).
- An initial full `POST https://api.todoist.com/api/v1/sync` is followed immediately by an incremental request using the returned sync token. This applies changes that arrived while Todoist prepared a potentially delayed full snapshot.
- The preview shows exact final project titles, task/sub-task counts, active-project collision warnings, optional p1–p3 or Eisenhower priority mapping, and a selection-aware list of data that cannot be carried over.
- Import runs project-by-project: `addProject`/`addTag` + `batchUpdateForProject` in self-chunked ≤50-op calls, then `updateTask` follow-ups for dates and tags.
- The summary compares **landed vs planned** counts from re-read state, reports created tags even after partial failure, and repeats the selected loss disclosures.

## Core touch points

- Launcher button in `file-imex` — in-memory `activatePlugin` only, so the importer is dormant again after restart (`isSkipMenuEntry: true`, no menu noise).
- Bundled plugin registration/build/CI entries, two translated launcher strings, and wiki page `2.20-Import-from-Todoist`.
- `batchUpdateForProject` actions capture their task-creation timestamp so replayed synced operations remain deterministic.

## Review hardening

- Fixed cross-chunk parent resolution that could orphan-delete sub-tasks.
- Guarded the virtual TODAY tag and deduplicated Todoist/Eisenhower labels case-insensitively.
- Made project title disambiguation case-insensitive and safe around literal numeric suffixes.
- Added full + incremental sync merging, including update and tombstone handling.
- Hardened remote payload handling: calendar validation, finite duration bounds, title/label/note limits with disclosure, and parsed HTTP(S)-only attachment URLs without control whitespace.
- Made partial tag/project failures and landed-count verification explicit.
- Localized the iframe UI and imported note labels through plugin i18n.
- Added accessible labels, fieldset/legend semantics, alert/status regions, a correct document outline/main landmark, and shared host design tokens.

## Verification

- **54/54 plugin Jest tests**: parser, sync merging, planning, chunking, priorities, failure paths, loss policy, and async iframe i18n.
- **17/17 host Karma tests** for the task batch reducer, including stable replay timestamps.
- Todoist plugin TypeScript check and build pass; every modified TS file passes `npm run checkFile`; Prettier and `git diff --check` are clean.
- Real-browser smoke with a local fake bridge: token → preview, priority-loss updates, 320 px layout, clean application console, **Lighthouse accessibility 100**.
- Dependency install audit reported no high or critical vulnerabilities.

## Known residuals

- No manual end-to-end import against a real Todoist account has been run yet.
- `getAllProjects()` exposes active SP projects only, so an archived previous import cannot be collision-flagged. The wiki documents restoring or deleting it before re-running; the plugin API is not widened solely for this importer.
- v2 candidates, only if demand warrants permanent API surface: repeat configurations, sections/folders, or batch support for dates/tags to remove follow-up updates.


docs/wiki/2.20-Import-from-Todoist.md
```diff
@@ -0,0 +1,34 @@
+# Import from Todoist
+
+Bring your active Todoist projects, tasks, sub-tasks, labels and due dates into Super Productivity in one pass. The import is **additive**: it only creates new projects, tags and tasks and never changes or removes existing Super Productivity data.
+
+## What you need
+
+- A Todoist account and its **API token**: in Todoist, open **Settings → Integrations → Developer** and copy the token. The token is sent only to `api.todoist.com` and is never stored.
```
lanesawyer, line 7: Process for getting an API token is correct!

docs/wiki/2.20-Import-from-Todoist.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Steps
```
lanesawyer, line 9: Steps are correct!

docs/wiki/2.20-Import-from-Todoist.md
```diff
@@ -0,0 +1,34 @@
...
+1. Open **Settings** (gear icon in the sidebar).
+2. Open the **Sync & Backup** tab and find the **Import/Export** section.
+3. Click **Import from Todoist**.
+4. Paste your API token and click **Load preview**.
+5. Review the preview: pick the projects to import (projects whose name already exists — for example from a previous run — are flagged and unchecked by default) and optionally enable mapping Todoist priorities to `p1`–`p3` tags.
+6. Click **Import**. The summary lists what was created and everything that could not be carried over.
+
+## What is imported
+
+- Active projects (nested project hierarchy is flattened; the Todoist Inbox becomes a project named “Inbox (Todoist)”)
```
lanesawyer, line 20: Hm. This makes sense as a very non-destructive action, but if users have a lot of tasks in their Todoist Inbox it might be annoying to copy them over, given that #4645 is still open.

Given that both apps have the concept of an Inbox, I think I'd expect my Todoist imports to land there. But this works great, I think most folks wouldn't have a giant Inbox anyway when importing!

docs/wiki/2.20-Import-from-Todoist.md
```diff
@@ -0,0 +1,34 @@
...
+- Active tasks and sub-tasks (Super Productivity nests two levels; deeper sub-tasks become direct sub-tasks of their top-level task)
+- Task descriptions and comments (into task notes; comment file attachments keep their link)
+- Labels used by imported tasks (as tags, on top-level tasks)
+- Due dates, including times, and minute-based durations (as time estimates)
+- Recurring tasks keep their next due date; the recurrence rule is added to the task notes (e.g. `Repeats: every 3 days`)
+
+## What is not imported
+
+- Completed tasks and task history
+- Sections (task order within the project is preserved)
```
lanesawyer: Any reason why sections can't be imported since both apps have the concept?

docs/wiki/2.20-Import-from-Todoist.md
```diff
@@ -0,0 +1,34 @@
...
+
+- A Todoist account and its **API token**: in Todoist, open **Settings → Integrations → Developer** and copy the token. The token is sent only to `api.todoist.com` and is never stored.
+
+## Steps
+
+1. Open **Settings** (gear icon in the sidebar).
+2. Open the **Sync & Backup** tab and find the **Import/Export** section.
+3. Click **Import from Todoist**.
+4. Paste your API token and click **Load preview**.
+5. Review the preview: pick the projects to import (projects whose name already exists — for example from a previous run — are flagged and unchecked by default) and optionally enable mapping Todoist priorities to `p1`–`p3` tags.
```
lanesawyer: The preview was nice and informative!

lanesawyer (commented):
What a lucky day for me, I've literally been trying to figure out how to get from Todoist to Super Producivity easily over the last day or two lol

It works!

Image

One thought I had was maybe adding a checkbox to the import saying "treat priority as the Eisenhower Matrix". I'm not looking forward to re-mapping p1, p2, p3 to the Super Productivity labels for that, given there is no bulk editing (#4645)

This was just a functional and docs review, I did not look into the code.

docs/wiki/2.20-Import-from-Todoist.md
johannesjo: Good instinct — and you're right that SP has a real Inbox project. I kept the Todoist inbox as its own `Inbox (Todoist)` project on purpose: the whole import is designed to be cleanly reversible (to undo, you just delete the created projects), and dropping tasks into your existing Inbox would mingle them with what's already there and break that. For the giant-inbox case it's admittedly a little more cleanup, but keeping the import self-contained felt like the safer default. Easy enough to drag them into the real Inbox afterwards if that's where you want them.

docs/wiki/2.20-Import-from-Todoist.md
johannesjo: You're right that SP has sections too — the reason they aren't imported is a plugin-API gap rather than a missing concept. The plugin API currently exposes `addProject` / `addTag` / `addTask` / `batchUpdateForProject` / `updateTask`, but there's no `addSection` yet, so the importer can't create them. It does read the Todoist sections to preserve task order within each project, so only the section grouping itself is dropped, not the ordering. Exposing sections to plugins is a good v2 candidate.

johannesjo:
Thanks so much for taking it for a spin and for the careful read through the docs — really glad it worked for you! 🙌

On the Eisenhower Matrix idea: there's actually already an opt-in checkbox in the preview step ("map Todoist priorities to `p1`–`p3` tags") that carries the priority over, so that info isn't lost today. Mapping straight onto the Eisenhower board is trickier than it looks though — Todoist priority is a single 1–4 ladder, while the matrix is two independent axes (urgent × important), so any "p1 → urgent+important / p2 → important / p3 → urgent" split is an opinionated guess I'd be baking in for everyone. For a one-time import I'd rather not impose one particular mapping (or add a second priority toggle next to the existing one), so I'll park it as a possible follow-up. And yeah — #4645 is the real pain here; bulk editing would make any post-import re-tagging trivial.

johannesjo:
Update on the Eisenhower idea — I went back and forth on my earlier reply and decided you were right, so it's in now (236c45c3fc).

Instead of adding a second checkbox, the priority step is now a single choice: **Nothing / p1–p3 tags / Eisenhower matrix**. Picking Eisenhower reuses SP's built-in `urgent` / `important` tags — p1 → urgent + important, p2 → important, p3 → urgent, p4 → none — so imported tasks land straight in the Eisenhower Matrix board with no new tags and no manual re-mapping afterwards. It stays opt-in and off by default, since collapsing Todoist's single priority axis onto the two-axis matrix is necessarily a bit opinionated.

Thanks for the nudge (and the review). 🙏

PR #8887, opened 2026-07-09
docs(development): add instructions for setting up local tests with Chromium

lanesawyer:
## Problem

I don't use Chromium browsers, but without one one my system, `npm run test` fails. But I want to push my commits and feel gross using `--no-verify`. I'd also like the on-boarding experience to give me a heads up about this potential pitfall.

## Solution

Adds a section about needing Chromium on your system to run the tests locally, and a few different ways to get that set up.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/docs/wiki/2.16-Set-Up-Development-Environment.md
+++ b/docs/wiki/2.16-Set-Up-Development-Environment.md
@@ -58,28 +58,32 @@ ERROR [launcher]: No binary for Chrome browser on your platform.
 
 Karma's `karma-chrome-launcher` locates the browser via the `CHROME_BIN` environment variable, falling back to a system Chrome/Chromium install. Pick one of the options below.
 
-### Option A — Use a system Chrome/Chromium
+### Option A — Use a system Google Chrome
 
-Install Google Chrome or Chromium through your OS package manager. On most platforms `karma-chrome-launcher` then finds it automatically. If it doesn't (or you keep it in a non-standard location), set `CHROME_BIN` explicitly as in Option B.
+Install **Google Chrome** through your OS package manager. The karma config uses `karma-chrome-launcher`'s `Chrome` launcher (`base: 'Chrome'` in `src/karma.conf.js`), which on Linux only auto-detects real Google Chrome — it probes for `google-chrome` / `google-chrome-stable`, **not** `chromium` / `chromium-browser`. So a distro **Chromium** package is generally _not_ picked up automatically; for anything other than Google Chrome (or a Chrome in a non-standard location) set `CHROME_BIN` explicitly, as in Option B.
 
 > **Snap / Flatpak caveat.** `karma-chrome-launcher` `exec`s the browser binary directly with its own flags (including a temp `--user-data-dir`), so sandboxed packages need extra care:
 >
-> - **Snap** works, but isn't auto-detected — the launcher only probes for `google-chrome`/`chromium-browser`, not `/snap/bin/chromium`. Set it explicitly: `export CHROME_BIN=/snap/bin/chromium` (verified: `Chrome Headless 150`, full unit suite green).
-> - **Flatpak** is not recommended. There's no plain executable to point `CHROME_BIN` at — you'd have to wrap `flatpak run org.chromium.Chromium "$@"` in a script — and even then the sandbox blocks Karma's temp profile directory, so it typically fails to launch. Prefer a native/distro package (this option) or Chrome for Testing (Option B).
+> - **Snap** works, but isn't auto-detected (the launcher doesn't probe `/snap/bin/chromium`). Set it explicitly: `export CHROME_BIN=/snap/bin/chromium` (verified: `Chrome Headless 150`, full unit suite green).
+> - **Flatpak** is not recommended. There's no plain executable to point `CHROME_BIN` at — you'd have to wrap `flatpak run org.chromium.Chromium "$@"` in a script — and even then the sandbox blocks Karma's temp profile directory, so it typically fails to launch. Prefer a native Google Chrome package (this option) or Chrome for Testing (Option B).
 
 ### Option B — Install Chrome for Testing and set `CHROME_BIN`
 
-Install a pinned Chrome for Testing build (downloaded to `~/.cache/puppeteer`), then point `CHROME_BIN` at it:
+Install a pinned Chrome for Testing build into `~/.cache/puppeteer`, then point `CHROME_BIN` at it:
 
 ```bash
-npx @puppeteer/browsers install chrome@stable
+npx @puppeteer/browsers install chrome@stable --path "$HOME/.cache/puppeteer"
 ```
+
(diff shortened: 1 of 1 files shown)


docs/wiki/2.16-Set-Up-Development-Environment.md
```diff
@@ -46,3 +46,46 @@ Types and keys are derived from `.env` when you run any build or serve command.
...
+Install a pinned Chrome for Testing build (downloaded to `~/.cache/puppeteer`), then point `CHROME_BIN` at it:
+
+```bash
+npx @puppeteer/browsers install chrome@stable
+```
+
+Add the resolved path to your shell profile (`~/.zshrc`, `~/.bashrc`, …) so it's set for every session — including the non-interactive shell the git hook runs in:
+
+```bash
+export CHROME_BIN="$(find "$HOME/.cache/puppeteer/chrome" -name chrome -type f | sort | tail -n1)"
```
johannesjo: **This flow doesn't work as written** — the `find` finds nothing and `CHROME_BIN` ends up empty.

The standalone `@puppeteer/browsers` CLI defaults `--path` to the **current working directory**, not `~/.cache/puppeteer` (that default belongs to the `puppeteer` npm package's install step, not this CLI). From `packages/browsers/src/CLI.ts`:

```js
.option('path', {
  desc: 'Path to the root folder for the browser downloads ...',
  defaultDescription: 'Current working directory',
  ...(required ? {} : {default: process.cwd()}),
})
```

So running the bare command from the repo root:
1. downloads Chrome into `./chrome/` **inside the repo** — an untracked ~hundreds-of-MB folder that isn't gitignored, so it shows up in `git status` and can be committed by accident;
2. makes `find "$HOME/.cache/puppeteer/chrome" ...` match nothing → `CHROME_BIN` exports as an empty string, and the verify step then errors.

Minimal fix — pin the install path so it matches the `find` target:

```bash
npx @puppeteer/browsers install chrome@stable --path "$HOME/.cache/puppeteer"
```

One more platform note for this block: `find ... -name chrome` is Linux-only. On macOS the Chrome-for-Testing binary is `Google Chrome for Testing` (inside a `.app`), so the `find` matches nothing there too — worth a one-line "Linux paths shown; macOS differs" caveat since the `~/.zshrc` example reads as cross-platform.

docs/wiki/2.16-Set-Up-Development-Environment.md
```diff
@@ -46,3 +46,46 @@ Types and keys are derived from `.env` when you run any build or serve command.
...
+```bash
+ERROR [launcher]: No binary for Chrome browser on your platform.
+  Please, set "CHROME_BIN" env variable.
+```
+
+Karma's `karma-chrome-launcher` locates the browser via the `CHROME_BIN` environment variable, falling back to a system Chrome/Chromium install. Pick one of the options below.
+
+### Option A — Use a system Chrome/Chromium
+
+Install Google Chrome or Chromium through your OS package manager. On most platforms `karma-chrome-launcher` then finds it automatically. If it doesn't (or you keep it in a non-standard location), set `CHROME_BIN` explicitly as in Option B.
```
johannesjo: Minor: the karma config uses `base: 'Chrome'` (`src/karma.conf.js`), and `karma-chrome-launcher`'s Chrome launcher only probes `google-chrome` / `google-chrome-stable` on Linux — **not** `chromium` / `chromium-browser` (those live in the separate, unused `Chromium` launcher). So a distro **Chromium** package generally won't be auto-detected either; only real Google Chrome is. The "if it doesn't, set `CHROME_BIN`" hedge covers it, but "finds it automatically" overstates the Chromium case.

johannesjo (commented):
Thanks for this — the pre-push/Chrome pitfall trips up new contributors and having it written down is genuinely useful. Two accuracy fixes before merge, one of which breaks the Option B copy-paste flow. Details inline.

Everything I could verify against the repo is correct: `.husky/pre-push` really is `npm run lint && npm run test`, `CHROME_BIN` is the right variable (`karma-chrome-launcher` `ENV_CMD`), the quoted launcher error matches Karma's real output, and the snap tip lines up with the launcher's PATH probing.

docs/wiki/2.16-Set-Up-Development-Environment.md
lanesawyer: Ugh, stupid Claude. It does really good work, but also does things without telling you sometimes. It decided to add the Chromium path to my `~/.zshrc` so I missed that little hangup. Will get it polished up!

lanesawyer:
@johannesjo Sorry for the initial AI slop, I've re-reviewed it with your feedback and it should be good to go. I can't test the Flatpak or Mac piece, but it _seems_ accurate from my limited experience with both.

johannesjo:
Thanks for this — really nice onboarding addition, and I appreciate that you traced the actual `karma-chrome-launcher` behavior instead of hand-waving it. I verified the technical claims against the repo and they all hold up:

- `.husky/pre-push` does run `npm run lint && npm run test` ✅
- The error text matches `karma/lib/launchers/process.js` exactly (`name='Chrome'`, `ENV_CMD='CHROME_BIN'`) ✅
- The Linux `Chrome` launcher really does probe only `google-chrome` / `google-chrome-stable` via `which` and ignores `chromium` (the `chromium` probes live on the separate `Chromium` launcher / `CHROMIUM_BIN`, which we don't use) ✅
- The `--path` / cwd-default note for `@puppeteer/browsers` and the macOS `.app` variant are both correct ✅

One small correctness fix worth making, then I'm happy to merge:

**Use `sort -V` instead of `sort` in the two `find … | sort | tail -n1` snippets.** Plain lexical `sort` can pick an *older* build as the "latest" once build-number segments differ in length. Quick repro:

```
$ printf '%s\n' \
    linux-140.0.7339.207 \
    linux-140.0.999.5 \
    linux-99.0.4844.51 | sort | tail -n1
linux-99.0.4844.51        # oldest — wrong

$ printf '%s\n' \
    linux-140.0.7339.207 \
    linux-140.0.999.5 \
    linux-99.0.4844.51 | sort -V | tail -n1
linux-140.0.7339.207      # correct
```

Only bites someone with multiple Chrome-for-Testing installs, but since the comment already promises "the (likely) latest version," `sort -V` makes it actually true for a one-character change.

Two optional nits, take them or leave them:

- `npm test` also runs `packages:test` / `release-notes:test` / `test:tz:ci` around `ng test`; only the `ng test` step needs Chrome. The framing is fine for the reader — just not *only* Karma.
- The fenced error-message block is tagged ` ```bash ` but it's console output, not a command (`text` or a bare fence would be marginally more accurate). Cosmetic.

Great addition overall — thank you!

johannesjo:
Thank you very much! <3

PR #8888, opened 2026-07-09
fix(accessibility): add ARIA roles, live regions, and alt attributes …

### LINKED ISSUE #8817: 🚨 Banners are never announced to screen readers (no aria-live)
## Problem

The banner component — the app's main surface for persistent async feedback (sync errors, reminders, timers) — renders a plain `div` with no `aria-live`, `role="status"`, or `role="alert"` (`src/app/core/banner/banner/banner.component.html:1`). Screen-reader users never hear a banner appear. Its optional image also lacks `alt` (line 17).

App-wide there are only **3** `aria-live` usages, so most async state changes (sync progress, issue polling) are unannounced. Snackbars are fine (Material provides a built-in live region).

## Suggested fix

- Add `role="status"` + `aria-live="polite"` to the banner wrapper (or `role="alert"` for error banners).
- `alt=""` (decorative) or a meaningful alt for the banner image.
- Optionally audit the remaining async surfaces (sync indicator, issue polling) for announcements.

---

SalmaAbdelrhmanMostafaMahmoud:
## Problem
The top banner notification component and its embedded progress bar were missing proper accessibility attributes. Screen reader users could not perceive or get real-time updates when the banner appeared or when background processes (like syncing) changed.

## Solution
Improved the accessibility (a11y) of the banner component by:
- Adding `role="status"` and `aria-live="polite"` to the main content container to make it a live region that screen readers announce immediately.
- Hiding decorative template images from assistive technologies using empty `alt=""` attributes.
- Adding standard ARIA properties (`aria-valuemin`, `aria-valuemax`, and dynamic `[attr.aria-valuenow]`) to the `<mat-progress-bar>` to expose its real-time percentage updates to screen readers.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

Closes #8817

### DIFF
--- a/src/app/core/banner/banner/banner.component.html
+++ b/src/app/core/banner/banner/banner.component.html
@@ -2,6 +2,8 @@
   <div
     [@banner]
     class="content-wrapper"
+    role="status"
+    aria-live="polite"
   >
     @if (banner.progress$ && (banner.progress$ | async); as progress) {
       @if (progress > 0) {
@@ -14,7 +16,10 @@
     <div class="inner-content-wrapper">
       @if (banner.img) {
         <div class="motivational-img-wrapper">
-          <img [src]="banner.img" />
+          <img
+            [src]="banner.img"
+            alt=""
+          />
         </div>
       }
       <div class="message-wrapper">
@@ -27,7 +32,10 @@
         <!-- -->
         @if (banner.timer$) {
           <div class="message">
-            <strong>{{ banner.timer$ | async | msToMinuteClockString }}</strong> –
+            <strong aria-hidden="true">{{
+              banner.timer$ | async | msToMinuteClockString
+            }}</strong>
+            –
             {{ banner.msg | translate: banner.translateParams }}
           </div>
         } @else {


johannesjo:
Thanks for improving the banner a11y! 🙏 The direction is right and the change is low-risk. A couple of things to address before merge, and one nit.

### 🔴 Blocking: this will fail the lint/prettier CI check
Adding `alt=""` makes `` a two-attribute tag, which the repo's Prettier config requires to be multi-line. Running the linter on the PR content gives:

```
22:15  error  Replace `·[src]="banner.img"·alt=""` with `⏎············[src]="banner.img"⏎············alt=""⏎·········`  prettier/prettier
```

Fix:
```html
<img
  [src]="banner.img"
  alt=""
/>
```
Running `npm run checkFile` / `npm run prettier` locally will auto-fix this (the PR checklist item for `checkFile` is currently unchecked).

### 🟠 Please remove the ARIA attributes on `<mat-progress-bar>` — they're redundant
Angular Material's `MatProgressBar` already sets these on its own host element, so the four additions here duplicate what the component renders. From `@angular/material/.../progress-bar` (v21):

```js
host: {
  attributes: { "role": "progressbar", "aria-valuemin": "0", "aria-valuemax": "100", "tabindex": "-1" },
  properties: { "[attr.aria-valuenow]": "_isIndeterminate() ? null : value", ... }
}
```

Since the bar here is always `mode="determinate"` with `[value]="progress"`, `aria-valuenow` already equals `progress`. Suggest reverting the progress-bar block to just:
```html
<mat-progress-bar
  mode="determinate"
  [value]="progress"
></mat-progress-bar>
```

### ✅ `role="status"` + `aria-live="polite"` — good, keep
This is the most valuable part of the change. Two non-blocking notes:
- `role="status"` already implies `aria-live="polite"`, so specifying both is technically redundant — but it's an accepted belt-and-suspenders pattern for wider AT support, so fine to keep.
- Reliability caveat (not a regression, just FYI): the banner — and thus the whole live region — is created/destroyed by `@if (activeBanner$ | async)`, and the component's `_dirtyReference` logic tears it down and re-adds it 500ms later on banner switches. Screen readers announce a live region most reliably when it already exists (empty) in the DOM *before* content is inserted, rather than being inserted together with its content. A future improvement would be a persistent empty live region; not needed for this PR.

### ✅ `alt=""` on the decorative image — correct
Right call for a decorative motivational image.

### Minor / FYI
- The `@if (banner.timer$)` branch is currently dead — no producer sets `timer$` on a banner (only commented-out code in `banner.service.ts`). No action needed, but worth knowing: if it's ever revived, its `msToMinuteClockString` clock updates the live region's **text** every second, which *would* trigger a polite re-announcement each tick (unlike the progress-bar attribute updates, which aren't announced because `aria-relevant` defaults to `additions text`). Something to keep in mind for that path.

Net: nice, focused a11y win. Just needs the Prettier fix (required) and ideally the redundant progress-bar attributes removed.

src/app/core/banner/banner/banner.component.html
```diff
@@ -2,6 +2,8 @@
   <div
     [@banner]
     class="content-wrapper"
+    role="status"
```
johannesjo, line 5: Small caveat with making `.content-wrapper` a live region: `role="status"` implies `aria-live="polite"` **and** `aria-atomic="true"`, so the whole banner is re-announced whenever any text inside it changes.

The tracking-reminder banner sets `timer$` (driven by `realTimer$(1000)` in `tracking-reminder.service.ts`) and renders it as `<strong>{{ banner.timer$ | async | msToMinuteClockString }}</strong>` — a once-per-second tick. So while that banner is visible, screen readers re-announce the entire banner **every second**, which is exactly the "constant alerts" noise the manifesto warns against. Every other (static) banner benefits correctly from this change.

Minimal fix — keep the ticking value out of the announced region:

```html
<strong aria-hidden="true">{{ banner.timer$ | async | msToMinuteClockString }}</strong>
```

The banner still announces once on appearance (message intact); only the per-second number stops spamming. (The `progress$` reminder banner is fine — its updates live inside `mat-progress-bar`, which isn't a live region.)

johannesjo (commented):
Reviewed the latest push (interdiff `5c5bf8f…35f9c56`). This is in good shape overall — thanks for the a11y work! 🙏

**Correct:** Removing the manual `aria-valuemin/valuemax/valuenow` from `<mat-progress-bar>` is the right call — Angular Material (v21) already sets `role="progressbar"` plus those three attributes natively via host bindings, so the template copies were redundant.

**Verified clean:** `eslint` + `prettier --check` both pass on all three changed files at head. The removed `eslint-disable` comments were dead no-ops (`@typescript-eslint/no-explicit-any` is `off` globally, and `naming-convention` already allows `snake_case`), so nothing breaks.

**One scope note:** the `update-check.service.spec.ts` and `work-context.const.spec.ts` changes are unrelated to #8817 — they look like stray `eslint --fix`/rebase churn. Harmless (lint passes), but I'd drop them to keep this PR focused on the banner.

One functional caveat inline on the live region. 👇

johannesjo:
Congrats btw. for PR nr 8888 🎉!!!

johannesjo:
Thank you very much! <3

SalmaAbdelrhmanMostafaMahmoud:
Thank you so much I really appreciate your patience and all your helpful feedback throughout the review process. I learned a lot from your comments. Looking forward to contributing again! 😊

Issue #8260, opened 2026-06-10
chore: remove dead code and reduce LOC (~2.5k verified dead, ~3k consolidatable)


johannesjo:
A codebase-wide analysis (knip + manual grep verification per finding) identified roughly **2,500 LOC of verified dead code** and another **~2,000–3,000 LOC** reclaimable through low-risk consolidation. This issue tracks the cleanup, split into independent tiers so each can land as its own PR.

## Tier A — Verified dead code (safe deletions, ~1,750–2,500 LOC)

Each item was verified to have zero imports, zero selector usage in any template, and zero string references across `src/`, `e2e/`, and `electron/`.

### Dead components / feature dirs (~720 LOC)

- [ ] `src/app/core-ui/main-header/mobile-side-panel-menu/` (~284 LOC)
- [ ] `src/app/core/theme/validated-color-palettes.const.ts` (196 LOC)
- [ ] `src/app/features/tag/dialog-edit-tags/` (~167 LOC)
- [ ] `src/app/features/tasks/dialog-task-detail-panel/` (~108 LOC) — superseded by the inline detail panel; also drop the stale `dialog-task-detail-panel` fallback from `e2e/constants/selectors.ts:131` and `e2e/utils/schedule-task-helper.ts:7`
- [ ] `src/app/ui/help-box/` (~90 LOC)
- [ ] `src/app/ui/tree-dnd/tree-drop-zones.component.ts` (81 LOC)
- [ ] `src/app/plugins/ui/plugin-menu.component.ts` (69 LOC) — superseded by `plugin-side-panel-btns` / `plugin-header-btns`
- [ ] `src/app/ui/stuck/stuck.directive.ts` + `src/app/ui/longpress/longpress-ios.directive.ts` (~78 LOC) — standalone directives never imported anywhere

### Dead utilities & op-log leftovers (~600 LOC)

- [ ] 11 unused files in `src/app/util/` (~241 LOC): `filter-local-action.ts`, `obs-log.ts`, `get-text-from-array-buffer.ts`, `format-iso-data-with-offset.ts`, `observe-object-property.ts`, `hash.ts`*, `resize-observer-obs.ts`, `timestamp-to-datetime-input-string.ts`, `get-yesterdays-date.ts`, `is-email.ts`, `is-chrome.ts`
  - *`hash.ts` carries a "NEVER CHANGE, USED FOR ID GENERATION" header but has zero references; deleting cannot affect already-generated IDs — double-check before removing
[...]

PR #8889, opened 2026-07-10
chore(ui): removes dead components and features [#8260 - Tier A]

lanesawyer:
## Problem

There's dead code! Identified in #8260, Tier A, dead components and features section

## Solution

Delete the code, using IDE search and LSP features to validate it's not used anywhere.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


lanesawyer:
I've got one additional component that may or may not need removal [here](https://github.com/super-productivity/super-productivity/issues/8260#issuecomment-4930952834), but the rest are good to go!

lanesawyer:
I think the test that failed is a false negative? I see a disabled Save button on the PR preview when I remove the date and time fields:

image

Granted, I'm not doing exactly what the e2e test does, as I can't insert an empty string. I did have Claude come up with something similar based on the Playwright code (see the end of this comment), and this does show disabled.

So I think it's a false negative? I'm unsure how flaky these tests can be.

---

```js
function getByRole(role, name) {
  const selectors = [`[role="${role}"]`];
  if (role === 'textbox') selectors.push('input', 'textarea');
  if (role === 'combobox') selectors.push('select');

  const els = document.querySelectorAll(selectors.join(', '));
  for (const el of els) {
    const label =
      el.getAttribute('aria-label') ||
      el.labels?.[0]?.textContent?.trim() ||
      el.getAttribute('placeholder') ||
      '';
    if (label.includes(name)) return el;
  }
  return null;
}

function setNativeValue(el, value) {
  const proto = Object.getPrototypeOf(el);
  const setter = Object.getOwnPropertyDescriptor(proto, 'value')?.set;
  setter.call(el, value);
}

function fillAndBlur(el, value) {
  el.focus();
  setNativeValue(el, value);
  el.dispatchEvent(new Event('input', { bubbles: true }));
  el.dispatchEvent(new Event('change', { bubbles: true }));
  el.blur();
  el.dispatchEvent(new Event('blur', { bubbles: true }));
  el.dispatchEvent(new Event('focusout', { bubbles: true }));
}

const dateInput = getByRole('textbox', 'Date');
const timeInput = getByRole('combobox', 'Time');

fillAndBlur(dateInput, '');
fillAndBlur(timeInput, '');
```

johannesjo:
Yes, seems like it. Unfortunately there is a little bit of flakiness sometimes for the e2e tests.

PR #8891, opened 2026-07-10
feat: add Home Assistant Bridge to community plugins

jloops412:
## Home Assistant Bridge Plugin

A bidirectional bridge between Super Productivity and Home Assistant with a rules-based automation engine.

### Security
- **Access token stored via \setSecret\/\getSecret\** - never in synced/exported data
- iframe is pure UI; plugin.js handles all secrets and authenticated API calls
- Auto-migrates old configs that had embedded tokens

### Architecture
- **plugin.js** = credential manager + HA API proxy + rules engine (survives navigation)
- **index.html** = pure UI, communicates via \window.parent.haBridge\ bridge object
- Matches the pattern used by SP-MCP and other community plugins

### What it does
- **20 triggers:** task_start, task_stop, task_switch, task_complete, task_created, task_updated, task_deleted, tags_changed, project_changed, due_date_changed, estimate_changed, estimate_exceeded, notes_changed, first_task_of_day, all_done, timer, idle, context_switch, project_list_changed, day_end
- **10 action types:** scenes, automations, scripts, toggle, light control, media control, TTS, custom services, phone notifications
- **Today detection:** uses \dueDay\/\dueWithTime\ (not virtual TODAY tag)
- **Smart entity discovery:** auto-fetches all HA devices for dropdown selectors
- **Non-blocking hooks:** all handlers deferred via \setTimeout\ to avoid SP timeout

### Links
- **Repository:** https://github.com/jloops412/ha-super-productivity
- **Release (v2.2.0):** https://github.com/jloops412/ha-super-productivity/releases/tag/v2.2.0
- **Plugin download:** [sp-plugin.zip](https://github.com/jloops412/ha-super-productivity/releases/download/v2.2.0/sp-plugin.zip)
- Also includes: A full Home Assistant custom integration (HACS compatible)

### Compatibility
- Super Productivity v10+
- Home Assistant 2024.1.0+
- Electron desktop app (Local REST API required)

### Formatting
- LF line endings preserved
- Trailing newline present
- Minimal diff (only new object appended)

### DIFF
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -1,122 +1,130 @@
 [
-  {
-    "name": "Date Range Reporter Plugin",
-    "shortDescription": "Generates reports of completed tasks and tasks with work logs within a specified date range.",
-    "url": "https://github.com/dougcooper/sp-reporter",
-    "author": "dougcooper",
-    "authorUrl": "https://github.com/dougcooper",
-    "stars": 32
-  },
-  {
-    "name": "Archived Tasks Viewer",
-    "shortDescription": "Read-only viewer to browse archived tasks with grouping, filtering, subtasks preview, and theme toggle.",
-    "url": "https://github.com/baiyina/Archived-Tasks-Viewer",
-    "author": "baiyina",
-    "authorUrl": "https://github.com/baiyina",
-    "stars": 31
-  },
-  {
-    "name": "QuestArc",
-    "shortDescription": "Turn your workflow into an adventure! Gamify your tasks with Atomic Habit quests, epic boss fights, earnable badges, and a customizable hero profile.",
-    "url": "https://codeberg.org/Nexumia/QuestArc",
-    "author": "Nexumia",
-    "authorUrl": "https://codeberg.org/Nexumia",
-    "stars": 4
-  },
-  {
-    "name": "AutoPlan",
-    "shortDescription": "Smart non-AI plugin for the automatic scheduling of tasks using priorities assigned automatically based on tags, projects, and time estimation.",
-    "url": "https://codeberg.org/00sapo/sp-autoplan",
-    "author": "00sapo",
-    "authorUrl": "https://codeberg.org/00sapo",
-    "stars": 28
-  },
-  {
-    "name": "Asana Integration",
-    "shortDescription": "Provides 2-way sync between SP tasks and Asana tasks.",
-    "url": "https://codeberg.org/sinenomine/sp-asana",
-    "author": "sinenomine",
-    "authorUrl": "https://codeberg.org/sinenomine",
-    "stars": 0
-  },
-  {
-    "name": "Obsidian Integration",
-    "shortDescription": "Provides 2-way sync between SP tasks and Obsidian tasks.",
-    "url": "https://codeberg.org/sinenomine/sp-obsidian/",
-    "author": "sinenomine",
-    "authorUrl": "https://codeberg.org/sinenomine",
-    "stars": 5
-  },
-  {
-    "name": "MCP",
-    "shortDescription": "Bridge between the Super Productivity app and MCP (Model Context Protocol) servers for Claude Desktop integration.",
-    "url": "https://github.com/organicmoron/SP-MCP",
-    "author": "organicmoron",
-    "authorUrl": "https://github.com/organicmoron",
-    "stars": 68
-  },
-  {
-    "name": "Super Productivity MCP",
-    "shortDescription": "Feature-rich MCP server with 19 tools covering full task CRUD, time tracking summaries, project/tag management, and notifications.",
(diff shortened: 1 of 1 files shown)


johannesjo (changes requested):
Thank you very much! Very cool! <3

I rechecked the current PR head (`6fea7f7e`) and the latest published plugin release (`v2.0.0`). Requesting changes for the following issues:

1. **Critical: do not catalog the plugin while it copies the Home Assistant access token into synced/exported state.** The released plugin keeps `haToken` in the same config object as its rules and serializes the whole object with `persistDataSynced()` ([released source](https://github.com/jloops412/ha-super-productivity/blob/v2.0.0/sp-plugin/plugin.js#L6-L22)); current `main` still does the same. Super Productivity's plugin guide explicitly says credentials must use `setSecret`/`getSecret`, because `persistDataSynced` data reaches sync servers, exports, and backups ([guide](https://github.com/super-productivity/super-productivity/blob/18c48d5ea8884cd0c07150c712d8810aef4979bb/docs/plugin-development.md#L534-L557)). Please keep only non-secret settings in synced data, store the token through the secret API, and publish a corrected artifact before adding the catalog entry.

2. **The linked release does not contain fixes required for the behavior advertised by this PR.** `v2.0.0` is five commits behind the repository's `main` branch ([comparison](https://github.com/jloops412/ha-super-productivity/compare/v2.0.0...main)). Those commits include the corrected `currentTaskChange` payload handling (the commit states focus rules previously never fired) and several config persistence/initialization fixes. The PR links users directly to the stale `v2.0.0` ZIP, whose iframe still calls `init()` before waiting for `PluginAPI` injection ([released source](https://github.com/jloops412/ha-super-productivity/blob/v2.0.0/sp-plugin/index.html#L351-L352)). Please publish and verify a release containing the fixes, then update the PR's artifact link/version information.

3. **The advertised `all_done` trigger cannot detect Today tasks, including on current plugin `main`.** It filters with `task.tagIds.includes('TODAY')` ([released source](https://github.com/jloops412/ha-super-productivity/blob/v2.0.0/sp-plugin/plugin.js#L208)), but `TODAY` is a virtual tag that must never occur in `task.tagIds`; membership comes from `dueDay`/`dueWithTime` ([architecture decision](https://github.com/super-productivity/super-productivity/blob/18c48d5ea8884cd0c07150c712d8810aef4979bb/ARCHITECTURE-DECISIONS.md#L41-L47)). Please use the supported Today membership representation and verify the trigger before advertising it.

4. **Restore the catalog file's existing formatting.** This commit converts all 122 existing lines from LF to CRLF and removes the final newline, turning a one-entry addition into a 252-line diff. `git diff --check` flags every line and `npx prettier --check --stdin-filepath src/assets/community-plugins.json` exits non-zero. Please preserve LF endings and format the JSON so the diff contains only the new object.

Verification notes: the resulting JSON parses, contains 16 entries, and has no duplicate names or URLs. The only reported GitHub check is the first-contributor welcome job; no substantive build/test check has run.

jloops412:
Thank you for the review and comments. I'm working on them now.

jloops412:
Issues fixed. Please let me know if there's anything further needed from my end. Really loving SP, and this HA integration is really cool. Has allowed me to connect to my HA to trigger things based on SP triggers.

johannesjo (commented):
Re-reviewed against the current PR head (`50688c3b`) and the published artifacts. **Both blocking issues from my earlier review are resolved** — thank you for the quick turnaround! 🙏

### 1. Token no longer in synced/exported state ✅
`plugin.js` (v5.0) now:
- Stores the token via `PluginAPI.setSecret('haToken', …)` and reads it via `getSecret` — never in synced data.
- Deletes `haToken`/`_token` from the config object before **every** `persistDataSynced(...)` call.
- Migrates old configs: an embedded token found in synced data is moved to `setSecret` and the stripped config is re-persisted.
- The iframe only handles a masked token (`getTokenMasked()`), never pre-fills the input, and never persists it itself. No raw-token logging.

### 2. Release no longer lags `main` ✅
The `plugin.js` in the **v2.2.0 release asset**, the **repo-root `sp-plugin.zip`**, and **`main`** are byte-for-byte identical and all include the payload-handling / config-persistence fixes. So users get the corrected code however they install.

I verified this by downloading and diffing the actual artifacts.

### One non-blocking nit before merge
The version numbering is inconsistent and will confuse users: the only GitHub Release is tagged **`v2.2.0`**, but `manifest.json` and the plugin banner both say **`v5.0.0`**. Could you cut a release whose tag matches the manifest version (or bump the manifest to match the tag)? That way "which version am I running?" has one answer. Not a blocker on the catalog entry itself.

The SP-side diff is clean (valid JSON, prettier-clean, matches the schema, no duplicate name, both URLs resolve). Once the tag/version is reconciled this is good to go from my side.

johannesjo:
Thank you very much! <3

PR #8892, opened 2026-07-10
chore(ui): removes dead utilities and op-log leftovers [#8260 - Tier A]

lanesawyer:
## Problem

There's dead code! Identified in https://github.com/super-productivity/super-productivity/issues/8260, Tier A, dead utilities and op-log leftovers section

## Solution

Delete the code, using IDE search and LSP features to validate it's not used anywhere!

### Potential Problem Files To Sanity Check

From the issue:

> *hash.ts carries a "NEVER CHANGE, USED FOR ID GENERATION" header but has zero references; deleting cannot affect already-generated IDs — double-check before removing

I did look for usages, was not finding any, either by text search or looking for references. **I do not believe it's used anywhere.**

> `src/app/op-log/sync-providers/native-http-retry.ts`

This one was a little tricky to feel confident about because `packages/sync-providers/src/http/native-http-retry.ts` exists, but I do believe the one in `src/app/op-log/sync-providers/` is not used, I did some searching and reference checking and it seems like everything uses the other one. But worth calling out as the most difficult file to validate during this process.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
I did a second, independent pass over all 29 deletions at `ce4e91f`, including exact symbol/path/template searches, relevant history, package boundaries, and the current CI results. The production-code deletions look safe: I found no live consumer of the removed app symbols, `hash.ts` has never had a repository reference since it was added, and the package-owned native retry implementation builds and passes its tests.

I found two required follow-ups before merge:

1. **Reconcile the op-log benchmark deletion with the active SQLite follow-up plan.** [The migration document](https://github.com/super-productivity/super-productivity/blob/ce4e91f4170740daa0e3052f5aa3f655c8da5bc5/docs/sync-and-op-log/sqlite-migration-followup.md#L138-L141) still identifies [`operation-log-stress.benchmark.ts`](https://github.com/super-productivity/super-productivity/blob/18c48d5ea8884cd0c07150c712d8810aef4979bb/src/app/op-log/testing/benchmarks/operation-log-stress.benchmark.ts#L1-L18) as the harness for the pending on-device Capacitor/SQLite validation. History also shows it was deliberately renamed from `.spec.ts` to `.benchmark.ts` to keep it out of normal CI, so lack of imports is expected. The documented manual command currently needs repair too: `src/tsconfig.spec.json` only includes `**/*.spec.ts`, while the app config excludes `testing/**`. Please either retain and make the harness runnable, replace it and update the plan, or update the plan to document why that validation path is being retired.

2. **Update the E2E inventory.** [`e2e/README.md`](https://github.com/super-productivity/super-productivity/blob/ce4e91f4170740daa0e3052f5aa3f655c8da5bc5/e2e/README.md#L78-L90) still lists the deleted `e2e/pages/schedule.page.ts`.

**Optional cleanup:** [`scripts/remove-unused-log-imports.ts`](https://github.com/super-productivity/super-productivity/blob/ce4e91f4170740daa0e3052f5aa3f655c8da5bc5/scripts/remove-unused-log-imports.ts#L6-L24) still tries to open the deleted `src/app/root-store/index.ts`; it catches the resulting read error, but removing the stale entry would keep the script internally consistent.

Once the two required references are reconciled, I have no remaining code-level concerns. The exact-head frontend build, unit tests, lint, packaged Electron smoke test, all six SuperSync shards, and WebDAV E2E are green.

lanesawyer:
Both feedback items addressed, `.benchmark.ts` file restored and runnable, it passes!

image

johannesjo:
Thank you very much! <3

PR #8909, opened 2026-07-10
feat(work-view): show break time today

jloops412:
## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/docs/wiki/4.17-Idle-Time.md
+++ b/docs/wiki/4.17-Idle-Time.md
@@ -42,6 +42,8 @@ In the dialog you can:
 - **Assign the time to a break** — Count the idle period as a break. This **resets the break-reminder timer** (see [[4.16-Break-Reminders]]) and records break time.
 - **Discard the time** — Don’t count it as work or break; it disappears from the log. Use this when the time was not productive and you don’t want it on any task or as a break.
 
+Once recorded break time is greater than zero, hover over **Without break** in the work-view status bar to see the day’s recorded **Break** total.
+
 So idle detection **corrects** the log by removing unattended time from the active task and then **lets you reallocate** it in a way that matches reality.
 
 ## Why Idle Time Is Surfaced Instead of Ignored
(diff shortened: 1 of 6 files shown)


johannesjo:
Thanks for this — the implementation is clean and nicely tested! 🙏

I'm hesitant about the dedicated status-bar item though. I try to keep that area minimal, and break time is already surfaced in the Daily Summary, so a second always-visible slot for it is more than the feature needs. It's also only populated once you assign idle time to a break (or edit it in the summary), so for most users it would be empty — and it's scoped per work context, so a break recorded on the Today tag wouldn't show while viewing a project, which reads as flakiness.

What I'd suggest instead: keep the `breakTimeToday$` observable you added (that part is good), but surface the value as a **tooltip on the existing "Without break" timer** rather than a new item. That timer resets exactly when break time gets recorded (`idle.effects.ts` — "break timer reset is handled in takeABreak"), so the two describe the same event and belong together. It keeps the status bar uncluttered while still making the total discoverable.

Concretely:

- Drop the new `.item` block
- Revert the visibility condition back to `@if (estimateRemainingToday() || workingToday())`
- Add a tooltip to the "Without break" value span:

```html
<span
  class="no-wrap"
  matTooltip="{{ T.F.TIME_TRACKING.D_IDLE.BREAK | translate }}: {{
    breakTimeToday() | msToString
  }}"
  [matTooltipDisabled]="!breakTimeToday()"
  matTooltipPosition="above"
>
```

That renders `Break: 5m` on hover and stays hidden when there's no break time — no new translation string needed. The service observable and its tests stay exactly as you have them. Worth tweaking the wiki line to describe the hover behavior too.

Would you be up for reworking it that way? Happy to clarify anything.

johannesjo:
Thank you very much! <3

Issue #8647, opened 2026-06-30
Improve add task and project backlog combo (🚨 "enable project backlog" breaks plus-button)


0ptimusPrimat3:
### Version Used

18.10.0 apk

### Current Behavior

when activating the backlog, the plus-button does not scale right and looks trashy in inbox as well as project views

<img width="922" height="2048" alt="Image" src="https://github.com/user-attachments/assets/62a2c8bc-848e-4c0b-82ed-b4ece3e1a823" />

### Expected Behavior

should ideally look like in today-view (same as it looks in inbox and project views when backlog drawer is deactivated) or scale correctly.

### Steps to Reproduce

1. burger-menu
2. 3-dot menu of f.e. inbox
3. settings
4. enable project backlog

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

johannesjo:
That is the intented behavior. I am open for improvement suggestions though.

0ptimusPrimat3:
thanks for responding so fast!

is it not possible or rather not trivial to keep it just like the default (deactivated backlog) because the drawer would interfere with the button design reaching above the bar?

if yes i would by far prefer a smaller but round version of the plus-button at least which should be pretty easy right?

i can try to fork it if you prefer the oval design although i must say i have not looked too deep into this codebase and did not work much with typescript yet

maybe you can tell me where to look for the inbox and project views and how the backlog toggles interfere with them

cheers awesome project!

lanesawyer:
I also noticed this today and thought it was a visual bug.

I do like the oversized + button in general, so I don't think the solution is to make it smaller.

Couldn't backlog just be a special section in a project, always at the end of the list and collapsed by default? That serves the same purpose as the bottom-docked drawer while leveraging existing UI patterns.

0ptimusPrimat3:
just for clarification:
by smaller i meant smaller width (just to make it round as all the other views), so height would stay as is. should have phrased that better.

i personally like the drawer feature and just making it round would be fixing this for me.
not enough time right now but ill have a closer look eventually

johannesjo:
[...]

PR #8914, opened 2026-07-11
fix(tasks): make "move to top/bottom" work for tasks in a section (#8049)

### LINKED ISSUE #8049: 🚨 "Move to Top" Not Working
### Version Used

18.9.1.0 Microsoft Store

### Current Behavior

The "Move to Top" feature doesn't seems to work for me after I reinstalled the Microsoft Store version's app. I'll have to use "Schedule today" as a workaround to have it stick on the top.

### Expected Behavior

When the click "Move to Top", it should move to the top of the list. 

### Steps to Reproduce

1. Right-click on a task.
2. Select "Move to top".
3. Nothing happens.

### Console Output

```shell

```

### Documentation

lanesawyer:
## Problem

Right-clicking a task that lives in a section and choosing **Move to top** had no visible effect. `section.taskIds` is authoritative for in-section display order (see `WorkViewComponent.undoneTasksBySection`), but `TaskService.moveToTop` only reordered the work-context `taskIds`, so a sectioned task never changed position.

In addition, the "move to bottom" keyboard shortcut doesn't move things to the bottom etiher, same root issue!

Fixes #8049.

## Solution

- Added `moveTaskToTop` and `moveTaskToBottom` in `SectionService`
- Calls them in the proper scenario in `TaskService`
- Adds and updates tests to keep it from breaking in the future!

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/tasks/task.service.ts
+++ b/src/app/features/tasks/task.service.ts
@@ -103,6 +103,8 @@ import { TaskLog } from '../../core/log';
 import { devError } from '../../util/dev-error';
 import { DEFAULT_GLOBAL_CONFIG } from '../config/default-global-config.const';
 import { TaskFocusService } from './task-focus.service';
+import { SectionService } from '../section/section.service';
+import { selectSectionsByContextIdMap } from '../section/store/section.selectors';
 import { DeletedTaskIssueSidecarService } from '../issue/two-way-sync/deleted-task-issue-sidecar.service';
 import { TimeBlockDeleteSidecarService } from '../calendar-integration/time-block/time-block-delete-sidecar.service';
 import { getDeadlineAutoPlanFields } from './util/get-deadline-auto-plan-fields';
@@ -121,6 +123,7 @@ export class TaskService {
   private readonly _taskArchiveService = inject(TaskArchiveService);
   private readonly _globalConfigService = inject(GlobalConfigService);
   private readonly _taskFocusService = inject(TaskFocusService);
+  private readonly _sectionService = inject(SectionService);
   private readonly _deletedTaskIssueSidecar = inject(DeletedTaskIssueSidecarService);
   private readonly _timeBlockDeleteSidecar = inject(TimeBlockDeleteSidecarService);
 
@@ -716,6 +719,7 @@ export class TaskService {
       const workContextId = this._workContextService.activeWorkContextId as string;
       const workContextType = this._workContextService
         .activeWorkContextType as WorkContextType;
+      const sectionId = isBacklog ? null : this._getSectionIdForTask(workContextId, id);
 
       if (isBacklog) {
         this._workContextService.doneBacklogTaskIds$
@@ -732,6 +736,8 @@ export class TaskService {
               }),
             );
           });
+      } else if (sectionId) {
+        this._sectionService.moveTaskToTopInSection(sectionId, id);
       } else {
         this._workContextService.doneTaskIds$.pipe(take(1)).subscribe((doneTaskIds) => {
           this._store.dispatch(
@@ -745,6 +751,21 @@ export class TaskService {
         });
       }
     }
+  }
+
+  private _getSectionIdForTask(workContextId: string, taskId: string): string | null {
+    let sectionId: string | null = null;
+    // selectSectionsByContextIdMap is BehaviorSubject-backed, so take(1)
+    // resolves synchronously.
+    this._store
+      .select(selectSectionsByContextIdMap)
+      .pipe(take(1))
+      .subscribe((sectionsByContext) => {
+        sectionId =
+          sectionsByContext.get(workContextId)?.find((s) => s.taskIds.includes(taskId))
+            ?.id ?? null;
+      });
+    return sectionId;
   }
 
   moveToBottom(id: string, parentId: string | null = null, isBacklog: boolean): void {
--- a/src/app/features/tasks/task.service.ts
+++ b/src/app/features/tasks/task.service.ts
@@ -720,8 +720,7 @@ export class TaskService {
       const workContextId = this._workContextService.activeWorkContextId as string;
       const workContextType = this._workContextService
         .activeWorkContextType as WorkContextType;
-      // Sections only exist in the today/regular list, so skip the lookup for
-      // the backlog.
+      // The backlog does not have sections, so no need to check for a section in that case
       const section = isBacklog ? null : this._getSectionForTask(workContextId, id);
 
       if (isBacklog) {
@@ -740,12 +739,7 @@ export class TaskService {
             );
           });
       } else if (section) {
-        // A sectioned task's visible order comes from section.taskIds (see
-        // WorkViewComponent.undoneTasksBySection), so "top" is a within-section
-        // reorder to the front — the same op drag-to-top dispatches. A null
-        // anchor prepends; passing the section as its own source keeps it a
-        // single-entity in-place move.
-        this._sectionService.addTaskToSection(section.id, id, null, section.id);
+        this._sectionService.moveTaskToTop(section.id, id);
       } else {
         this._workContextService.doneTaskIds$.pipe(take(1)).subscribe((doneTaskIds) => {
           this._store.dispatch(
@@ -803,12 +797,7 @@ export class TaskService {
             );
           });
       } else if (section) {
-        // Mirror of moveToTop: append after the section's current last task
-        // (excluding this one) so the task lands at the bottom of its section,
-        // matching drag-to-bottom. No anchor when it would be alone.
-        const others = section.taskIds.filter((tId) => tId !== id);
-        const afterTaskId = others.length ? others[others.length - 1] : null;
-        this._sectionService.addTaskToSection(section.id, id, afterTaskId, section.id);
+        this._sectionService.moveTaskToBottom(section, id);
       } else {
         this._workContextService.doneTaskIds$.pipe(take(1)).subscribe((doneTaskIds) => {
           this._store.dispatch(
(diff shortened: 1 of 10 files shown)


src/app/features/tasks/task.service.ts
```diff
@@ -747,6 +753,21 @@ export class TaskService {
     }
   }
 
+  private _getSectionIdForTask(workContextId: string, taskId: string): string | null {
+    let sectionId: string | null = null;
+    // selectSectionsByContextIdMap is BehaviorSubject-backed, so take(1)
+    // resolves synchronously.
+    this._store
```
lanesawyer, line 762: I am very unsure if this is the proper approach. Still learning the codebase and this is what Claude came up with, feedback here is very welcome!

johannesjo:
Thanks for this — the fix is correct and works, and the tests/docs are thorough. Since you flagged that you're unsure about the approach and want to keep the sync footprint minimal, I dug into whether the new synced action is actually necessary. Here's what I found.

## The core question: does this need a new op-log code?

The new `moveTaskToTopInSection` action introduces a **permanent op-log wire code (`S7`)**. That's a hard-to-reverse addition to the sync surface, so it's worth confirming it earns its keep — especially for a minor UX fix.

It turns out the already-shipped `addTaskToSection` (code `S5`) does the same thing. "Move to top" is just a within-section reorder to the front, i.e.:

```ts
// task.service.ts, section branch of moveToTop — sectionId & id are already in scope:
this._sectionService.addTaskToSection(sectionId, id, /* afterTaskId */ null, /* sourceSectionId */ sectionId);
```

For the real use case (a task that **is** in the section, moved to the front) the two are equivalent down to the details:

- **Reducer result is byte-identical.** With `sourceSectionId === sectionId` the strip is skipped and `moveItemAfterAnchor(taskId, null, taskIds)` returns `[taskId, ...rest]` — the same array `arrayMoveToStart` produces.
- **Op meta is identical.** `addTaskToSection`'s `source === section` branch resolves to `{ entityType: 'SECTION', entityId: sectionId, opType: Move }` — the same single-entity SECTION/Move op your new action emits. Same validation path, same (empty) `entityChanges`, no extra side effects.
- **It's literally the existing drag-to-top path.** Dragging a sectioned task to the top already dispatches exactly this call (`task-list.component.ts`), so reusing it makes the context-menu and drag behave identically.

Reusing it deletes the new action, reducer case, service method, enum member, `S7` code, and the count-guard bump — the diff shrinks to the detection logic in `moveToTop` plus updated service tests, and the sync system gains **nothing new**.

## The one honest tradeoff

There's exactly one behavioral difference, and it only shows up on **replay when the task is no longer in the section** (e.g. another device concurrently deleted it, moved it to another project, or archived it within the same sync window):

- **Dedicated `S7` action:** clean no-op (`!taskIds.includes(taskId)` → return state).
- **Reusing `addTaskToSection`:** re-adds the id to the front of `section.taskIds`, leaving a stale/phantom reference until a `dataRepair` pass clears it (mostly invisible — `undoneTasksBySection` filters ids with no matching task at render time; the cross-project variant isn't auto-repaired though).

Worth weighing honestly in both directions: that resurrection already exists today on the drag-to-top path, so reuse doesn't add a *new* class of bug — but the dedicated action's `S7` code has its own sync cost: **older released clients can't decode `S7`**, so they'd silently drop the reorder until they update (the usual synced-literal-union forward-compat hazard).

## Suggestion

For a low-priority reorder feature where you want minimal sync impact, I'd lean toward **reusing `addTaskToSection`** — it fixes the same bug, adds zero permanent wire surface, and unifies behavior with drag-to-top. The only reason to keep the dedicated action is if you consider the phantom-id-on-concurrent-remove edge case unacceptable (a legitimate position given how strict this codebase is about sync correctness) — in which case it's worth a one-line comment on the action explaining that the no-op-on-missing semantics are the whole point, so a future reader doesn't "simplify" it back into `addTaskToSection`.

Either way is defensible; just flagging that the leaner path exists so it's a deliberate choice. Happy to help with whichever you pick.

johannesjo:
One more thing worth folding in: **"Move to bottom" has the exact same bug.** `TaskService.moveToBottom` dispatches `moveTaskToBottomInTodayList`, which only reorders the work-context list — invisible for a sectioned task, same as move-to-top. It'd be odd to ship the fix for one direction and not the other, so it's worth doing both in this PR.

This also tips the design question, I think. If you keep the dedicated-action route, bottom needs a **second** permanent op-log code (`S8 moveTaskToBottomInSection`) alongside `S7` — two new synced actions for two minor reorder shortcuts.

The `addTaskToSection` reuse route covers **both directions with zero new op-codes** — it's just a different anchor:

```ts
// move to top:    prepend
addTaskToSection(section.id, id, /* afterTaskId */ null, section.id);

// move to bottom: append after the section's current last task (excluding this one)
const others = section.taskIds.filter((tid) => tid !== id);
const afterTaskId = others.length ? others[others.length - 1] : null;
addTaskToSection(section.id, id, afterTaskId, section.id);
```

Both land the task at the section's edge via `moveItemAfterAnchor`, matching the existing drag semantics. The only extra bit vs the current PR is that the detection helper should return the section (or its `taskIds`) instead of just the id, since bottom needs the last-task anchor.

So the bottom requirement makes the reuse path the clear winner on sync footprint: one existing action, no new codes, both directions — vs. `S7` + `S8`. Same concurrent-remove tradeoff as before (anchor-based re-add on replay), and it stays consistent with how drag already behaves in both directions.

lanesawyer:
Ah, I had started with `addTaskToSection`, but then noticed other specific move to top actions, like `moveProjectTaskToBacklogList` and thought that having a specifically named one would be better. But yeah, makes sense that these actions are sensative to change for syncing purposes.

Thank you for the explanation! I will pivot back to `addTaskToSection` and revert the op-log modifications.

Can also tackle move to bottom as well!

lanesawyer:
On a side note, as I poke at section code more and more, I'm seeing many things that feel like duplicates with similar code (e.g. move to top of Today, move to top of Backlog, move to top of Section).

I'm considering drawing up plans for unifying everything by changing a fundamental assumption to be that _everything_ always has a section, and that default section simply doesn't have a UI for a name or collapsible area.

This is along similar lines of my [other comment](https://github.com/super-productivity/super-productivity/issues/8647#issuecomment-4920603622) of turning the backlog into a special section.

We could have a default section ID and a backlog section ID, and then the rest of the sections are generated as they are today. Then all these re-arranging (and other) code paths can be unified, since we are always operating within a section context.

Is that a refactor worth pursuing in order to simplify the codebase? I haven't wanted to dig too deeply into it in case there is a major technical hurdle that will need to be overcome, so any guidance there would be very helpful!

lanesawyer:
Alright, should be good to go, including a fix for move to bottom.

At a high level, I did the following to fix it:
- Added `moveTaskToTop` and `moveTaskToBottom` in `SectionService`
- Calls them in the proper section scenario in `TaskService`
- Adds and updates tests to keep it from breaking in the future!

johannesjo (changes requested):
## Summary

The final approach correctly reuses the existing S5 SECTION/MOV operation for both directions, so it adds no permanent sync wire surface and preserves one user intent = one operation. The tests and all current CI checks are green.

## Important findings

- src/app/features/tasks/task.service.ts:724 — This routes by stored section membership even when the task-view customizer is active. In filter-only or grouping-only views, WorkViewComponent hides sections and renders the regular UNDONE list ordered from work-context taskIds; this branch instead updates hidden section.taskIds, so Move to top/bottom again appears to do nothing. Before this PR those commands reordered the visible work-context list. Please route from the rendered list context, or at minimum bypass section routing while customized, and add top/bottom regression coverage.

- src/app/features/tasks/task.service.ts:767 — find() cannot identify which rendered section instance the user acted on. Concurrent moves with sourceSectionId: null can leave the same task referenced by multiple sections, and the work view renders each reference; acting on a copy in a later section will always reorder and sync the first stored section instead. Passing the rendered listModelId/section ID from TaskListComponent through the caller avoids mutating the wrong section.

## Verdict

Needs changes — please route the command using the rendered list context and cover customized-list and duplicate-membership cases.

lanesawyer:
I am now understanding why you said this might get complicated! I'm reading many of the `docs/` about the operation log to better understand and will really try to dig into the proper solution and not just send you Claude's best effort. Might be a week or so until I feel confident on a fix, but I will keep working on it!

johannesjo:
Thanks for the update — I re-checked both requested-change findings against the current head (`177b83e`) and an independent adversarial pass. Both are confirmed.

The good news is that the remaining fix does **not** need an op-log redesign. Reusing the existing `addTaskToSection` / S5 operation is still the right sync choice; the unresolved problem is that the rendered list identity is discarded before `TaskService` decides which operation to emit.

- `WorkViewComponent` already identifies customized/filter/grouping lists as `UNDONE` and real section lists by `section.id`.
- `TaskListComponent` knows that `listModelId`, but it is not passed through `TaskComponent` / the context-menu path. `TaskService` therefore guesses from stored section membership.
- That guess causes both failures: customized views update a hidden `section.taskIds`, while duplicate section references make `.find()` update the first stored section instead of the exact rendered copy the user acted on.

The smallest clean approach looks like:

1. Thread an explicit rendered source (for example, a nullable `sectionId` or the existing `listModelId`) from `TaskListComponent` through the caller to `TaskService`.
2. Route a real section ID through the existing S5 section move; route `UNDONE` through the existing work-context move action.
3. Remove the `_getSectionForTask()` membership inference.

One easy path to miss: `TaskContextMenuInnerComponent.moveToTop()` calls `TaskService` directly, so passing the context only through the keyboard/`TaskComponent` methods would leave right-click “Move to top” unfixed.

Suggested regression cases:

- filter-only and grouping-only customized views: top/bottom on a task with stored section membership updates the work-context order;
- duplicate membership: invoking from the second rendered section updates that exact section;
- the context-menu “Move to top” path carries the same rendered-list context.

So there should be no need for new op codes or a broad “everything is a section” refactor in this PR — this can stay a focused UI-context plumbing fix.

Issue #8715, opened 2026-07-02
💥 Missing taskRepeatCfg


Kocherga38:
selectTaskRepeatCfgFeatureState, (file:///C:/Program%20Files/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/src/app/features/task-repeat-cfg/store/task-repeat-cfg.selectors.ts:24:12)memoize(function (file:///C:/Program%20Files/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/@ngrx/store/fesm2022/ngrx-store.mjs:943:23)isArgumentsChanged (file:///C:/Program%20Files/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/@ngrx/store/fesm2022/ngrx-store.mjs:826:35)defaultStateFn (file:///C:/Program%20Files/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/@ngrx/store/fesm2022/ngrx-store.mjs:850:36)defaultMemoize(function (file:///C:/Program%20Files/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/@ngrx/store/fesm2022/ngrx-store.mjs:946:29)isArgumentsChanged (file:///C:/Program%20Files/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/@ngrx/store/fesm2022/ngrx-store.mjs:826:35)source$.pipe(pluck (file:///C:/Program%20Files/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/@ngrx/store/fesm2022/ngrx-store.mjs:719:43)map (file:///C:/Program%20Files/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/rxjs/dist/esm/internal/operators/map.js:7:30)function (file:///C:/Program%20Files/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/rxjs/dist/esm/internal/operators/OperatorSubscriber.js:12:8)next (file:///C:/Program%20Files/Super%20Productivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/rxjs/dist/esm/internal/Subscriber.js:29:11)
[...]

PR #8925, opened 2026-07-11
fix(dnd): allow dragging recurring tasks onto sidebar projects

### LINKED ISSUE #8739: 🚨 Drag and drop doesn't work in the inbox
### Version Used

Version 18.1.3, Debian OS 13

### Current Behavior

When I try to drag the note from the main window to the left panel to one of projects, the note isn't moved.

### Expected Behavior

The note should be moved to the project.

### Steps to Reproduce

1. Go to the inbox.
2. Create a new note.
3. Schedule the note.
4. Set recurrence.
5. Try to drag the note from the main window to the left panel to one of projects.
6. The note isn't moved.

Rishi943:
## Problem

Dragging a **recurring** task from the task list onto a project in the sidebar silently does nothing: `_handlePointerUp` in `magic-side-nav.component.ts` has a blanket early return for any task with a `repeatCfgId`, which discards project drops along with tag drops.

Fixes #8739

## Solution

Move the `repeatCfgId` exclusion from the top of `_handlePointerUp` down into the tag-drop branch, which is the only path that needs it (the Today-tag branch reschedules the task's due date, which conflicts with repeat scheduling). Project moves never touch scheduling fields, so recurring tasks can now be dropped onto sidebar projects like any other task.

Added a `magic-side-nav.component.spec.ts` covering: regular + recurring task dropped on a project (both move), recurring task dropped on a tag (still excluded), regular task dropped on a tag (still works).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/core-ui/magic-side-nav/magic-side-nav.component.spec.ts
+++ b/src/app/core-ui/magic-side-nav/magic-side-nav.component.spec.ts
@@ -1,4 +1,4 @@
-﻿import { of } from 'rxjs';
+import { of } from 'rxjs';
 import { TaskWithSubTasks } from '../../features/tasks/task.model';
 import { createTask } from '../../features/tasks/task.test-helper';
 import { MagicSideNavComponent } from './magic-side-nav.component';
@@ -19,13 +19,15 @@ const buildReceiver = (
   activeTask: TaskWithSubTasks | null,
 ): {
   receiver: MagicSideNavComponent;
-  moveToProject: jasmine.Spy;
+  moveTaskToProjectWithRepeatCfgAwareness$: jasmine.Spy;
   addToToday: jasmine.Spy;
   updateTags: jasmine.Spy;
   setCancelNextDrop: jasmine.Spy;
   setActiveTask: jasmine.Spy;
 } => {
-  const moveToProject = jasmine.createSpy('moveToProject');
+  const moveTaskToProjectWithRepeatCfgAwareness$ = jasmine
+    .createSpy('moveTaskToProjectWithRepeatCfgAwareness$')
+    .and.returnValue(of('moved'));
   const addToToday = jasmine.createSpy('addToToday');
   const updateTags = jasmine.createSpy('updateTags');
   const setCancelNextDrop = jasmine.createSpy('setCancelNextDrop');
@@ -33,7 +35,7 @@ const buildReceiver = (
 
   const receiver = Object.assign(Object.create(MagicSideNavComponent.prototype), {
     _taskService: {
-      moveToProject,
+      moveTaskToProjectWithRepeatCfgAwareness$,
       addToToday,
       updateTags,
     },
@@ -47,7 +49,7 @@ const buildReceiver = (
 
   return {
     receiver,
-    moveToProject,
+    moveTaskToProjectWithRepeatCfgAwareness$,
     addToToday,
     updateTags,
     setCancelNextDrop,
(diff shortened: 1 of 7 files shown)


johannesjo (commented):
Thanks for tracking this down — the root cause is right and the diff is minimal and well-targeted. Lint/format is clean and the 6 new tests pass locally. One correctness issue with recurring tasks is worth resolving before merge, plus a couple of nits.

## 🔴 The drag path does a *partial* move for recurring tasks

The fix routes recurring tasks straight into the raw single-instance move (`this._taskService.moveToProject(...)`), but the app already has a deliberate **recurring-aware** move-to-project flow, currently duplicated in the two other entry points:

- `task-context-menu-inner.component.ts` → `moveTaskToProject` (~L579)
- `task/task.component.ts` → `moveTaskToProject` (~L1253)

Both do materially more when the task has a `repeatCfgId`:
1. **Update `taskRepeatCfg.projectId`** — otherwise future occurrences keep spawning in the *old* project (instance creation reads `taskRepeatCfg.projectId`, `task-repeat-cfg.service.ts:392`).
2. **Move the other existing instances** (present + archive), not just the dragged one.
3. **Show a confirm dialog** (`D_CONFIRM_MOVE_TO_PROJECT`) when more than one instance exists.

The drag path skips all three. `TaskService.moveToProject` → `moveToOtherProject` only rewrites the task entities' `projectId` and the project task-lists (`project-shared.reducer.ts` `handleMoveToOtherProject`); I checked every `moveToOtherProject` consumer (effects + meta-reducers) and **nothing** re-syncs the repeat cfg automatically.

Concrete outcome for the exact reported scenario (freshly-created scheduled recurring note dragged onto a project):

- The cfg is created with the current project (`addTaskRepeatCfgToTask`, `task-repeat-cfg.service.ts:90`) — i.e. Inbox.
- Drag moves only this instance's `projectId` → Project A; `taskRepeatCfg.projectId` stays Inbox.
- **The next occurrence is created back in Inbox.** Split-brain state (`task.projectId` = A, cfg = Inbox) that the two existing paths specifically avoid — even in the single-instance case, which they silently fix with no dialog.

So the fix makes the drag *do something*, but something inconsistent with every other move-to-project path, and likely to produce a "my recurring task keeps returning to Inbox" follow-up.

**Suggestion:** extract the recurring-aware move (currently copy-pasted across the two components) into a shared `TaskService` method and have all three call sites — including this drag path — use it. That fixes the divergence *and* the existing DRY duplication. If you'd rather keep this PR minimal, at least update `taskRepeatCfg.projectId` for recurring tasks on the drag path so the reported note doesn't bounce back. There's a design call here worth making explicitly: should drag-to-project reuse the confirm-dialog series move, or intentionally be a lightweight single-instance move?

## 🟡 The tests lock in the partial behavior

`moves a recurring task to the project (#8739)` only asserts `moveToProject` was called with `(task, 'project-1')`, so it encodes the incomplete behavior as "correct" and can't catch the divergence above. The prototype-reflection harness (`Object.create(prototype)` + patched private fields) is pragmatic given the DI graph and is honestly documented, but it tests plumbing rather than the move outcome — extracting the shared service method would let this be tested where it matters.

## 🔵 Nits

- **UTF-8 BOM:** `magic-side-nav.component.spec.ts` is the only `.ts` file in the repo with a byte-order mark (`EF BB BF`). Harmless (prettier keeps it, `checkFile` passes) but inconsistent — worth stripping.
- **Comment vs. code in the tag branch:** the new guard excludes recurring tasks from *all* tag drops, but the comment justifies it only via the "Today" tag's due-date reschedule. Non-Today tag drops (`updateTags`) don't touch scheduling, so the exclusion is slightly broader than the stated reason. Preserving prior behavior is fine/conservative — just noting the comment overstates the justification.

## ✅ Verified good

- Correct root cause; minimal, targeted diff.
- Subtask-on-project exclusion, no-active-task, and tag-still-excluded cases are all covered.
- The tag-branch early return does **not** leak `activeTask` — `TaskListComponent.onDragEnded` clears it independently. Not a regression.

Rishi943:
Thanks for catching the split-brain, @johannesjo — you're right that the drag path silently diverged from every other move-to-project flow. Pushed 2ccbe05 implementing your primary suggestion:

**Shared method:** the recurring-aware flow (cfg `projectId` update, present + archive instance moves, `D_CONFIRM_MOVE_TO_PROJECT` dialog for >1 instance, and the #8715 deleted-cfg fallback) is extracted verbatim into `TaskService.moveTaskToProjectWithRepeatCfgAwareness$`, and all three call sites — context menu, task component, and the drag path — now delegate to it. On the design call: I went with reusing the confirm-dialog series move for the drag path, since a drag expressing a different intent than the context-menu "Move to project" on the same task seemed harder to justify than one consistent behavior.

Implementation notes:

- The method returns `'moved' | 'confirmed' | 'cancelled'` so each component wrapper reconstructs its exact prior finalize behavior (context menu: `onClose()` in all cases; task component: `focusSelf()`/`focusNext(true)` sequencing preserved per branch).
- `ProjectService` and `TaskRepeatCfgService` both inject `TaskService`, so they're resolved through the repo's existing `lazyInject` pattern (as in `archive-operation-handler.service.ts`) to avoid the circular DI.
- Tests now live where the behavior lives: service-level specs cover the lone-instance direct move (asserting the cfg `projectId` update you flagged as untested) and the #8715 fallback; the component specs are reduced to delegation checks.

Nits: BOM stripped from `magic-side-nav.component.spec.ts`, and the tag-drop comment reworded to say it conservatively preserves prior behavior rather than implying the due-date rationale covers non-Today tags.

johannesjo:
Re-reviewed the latest head with several independent passes. I **don't** think an atomic sync redesign should block this PR: the multi-op recurring-series move and fire-and-forget archive update predate #8925, and redesigning that operation format/meta-reducer would expand this small bug fix substantially. Those are better handled as a separate sync follow-up.

Two scoped items still look worth fixing here:

1. **Sanitize the log in `TaskService.moveTaskToProjectWithRepeatCfgAwareness$`.** `TaskLog.log({ reminderCfg, nonArchiveInstancesWithSubTasks, archiveInstances })` records full repeat configs and task objects—including titles/notes—into exportable diagnostic history. The statement existed in the older menu implementations, but recurring sidebar drag now reaches it too. Please remove it or log IDs/counts only.

2. **Cover the established-recurring-task branch at service level.** The nav spec mocks the shared helper, while the new service specs cover only a lone instance and a deleted config. Please add confirm + cancel tests for multiple live/archive instances (including archived subtasks/update dispatch), since that is the common branch for an existing recurring task and now serves all three callers.

With those two focused changes, the implementation looks good to me. Current CI is green.

PR #8926, opened 2026-07-11
refactor(tasks): extract shared task ordering helpers

### LINKED ISSUE #7912: refactor(tasks): extract shared task ordering helpers
Found during a 15-agent simplification scan.

## Problem
Several task ordering paths duplicate reducer mechanics and validation:
- src/app/features/tasks/store/task.reducer.ts: moveSubTaskUp, moveSubTaskDown, moveSubTaskToTop, moveSubTaskToBottom
- src/app/features/tasks/store/task.reducer.ts: removeTasksFromTodayTag and localRemoveOverdueFromToday
- related project/tag work-context movement handlers duplicate similar list updates

## Candidate Simplification
Extract small pure helpers for validating parent/subtask membership, applying array move functions, moving valid IDs to the front, and sharing project/tag/backlog list movement where it fits cleanly.

## Risk
Medium. These actions affect ordering, drag/drop, keyboard movement, synced state, and TODAY_TAG virtual membership. Keep reducer ordering and logging behavior identical.

## Verification
- npm run test:file src/app/features/tasks/store/task.reducer.spec.ts
- npm run test:file src/app/features/work-context/store/work-context.selectors.spec.ts
- npm run test:file src/app/features/tag/store/tag.reducer.spec.ts
- focused manual/E2E checks for moving tasks in project, tag, Today, and backlog lists

Rishi943:
## Problem

`task.reducer.ts` repeats the same two ordering patterns in several places:

- `moveSubTaskUp/Down/ToTop/ToBottom` each duplicate the "is this id actually in the parent's `subTaskIds`?" membership check before applying their move function.
- `removeTasksFromTodayTag` and `localRemoveOverdueFromToday` duplicate the "filter valid ids, move them to the front, keep the rest in order" logic.

Closes #7912

## Solution

Extract two pure, unit-tested utils and use them from the reducer — no behavior change:

- `getReorderedSubTaskIds(subTaskIds, id, moveFn)` — membership check + move, returns `null` when the id isn't a member so the reducer keeps its existing error log + no-op path.
- `moveValidIdsToFront(allIds, idsToMove, isValidId)` — stable "move to front" shared by both Today-tag removal actions, returning `invalidCount` so `removeTasksFromTodayTag` keeps its orphan-filter `devError`.

Both utils have their own specs (`get-reordered-sub-task-ids.spec.ts`, `move-valid-ids-to-front.spec.ts`).

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo (commented):
Thanks for tackling this @Rishi943 — and nice work on the co-located specs. I traced the behavior carefully; **the refactor is correct and behavior-preserving**. My one substantive suggestion is to split the verdict per util.

### ✅ Keep `moveValidIdsToFront`
This removes a genuine 2× duplication — `removeTasksFromTodayTag` and `localRemoveOverdueFromToday` had byte-identical ordering logic — and returning `invalidCount` cleanly preserves the orphan `devError` that only `removeTasksFromTodayTag` wants. Textbook DRY, worth the file.

### 🤔 Reconsider `getReorderedSubTaskIds`
This one doesn't actually remove any duplication. The four `moveSubTask*` actions already funnel through the single `reorderSubTask` helper (`task.reducer.ts:435-449` → `133-161`), so the membership-check-then-move pattern lives in exactly **one** place. I checked the rest of the tasks/tag/work-context stores for another consolidation target and there isn't one:
- `addSubTask`'s `includes()` is an append/idempotency guard (different shape, not check-then-move).
- The tag `moveTask*InTodayList` handlers apply `arrayMove*` but have no membership guard — they rely on `arrayMove`'s own `indexOf === -1 → return arr` no-op.

So the extraction mostly moves a 3-line check out of `reorderSubTask` and splits the `null` guard from the `TaskLog.err` that explains *why* it bailed — a small locality cost on a sync-path reducer for no dedup benefit. The description's framing ("`moveSubTaskUp/Down/ToTop/ToBottom` each duplicate the membership check") describes duplication that the existing `reorderSubTask` helper had already resolved.

**Suggestion**
- Keep everything `moveValidIdsToFront`.
- Inline the check back into `reorderSubTask` and drop `get-reordered-sub-task-ids.ts` + its spec.
- Fair alternative if you'd rather keep it: it *is* idiomatic for this `util/` dir and makes the guard unit-testable — but then please reframe the description to sell it as "isolate for testability," not deduplication.

### Behavior check ✓
No behavior change in any of the three rewrites:
- `if (!reorderedSubTaskIds)` fires only on `null` — `arrayMove*` always returns a truthy array (even `[]`).
- `invalidCount > 0` is logically identical to the old `validTaskIds.length !== taskIds.length` (`validTaskIds ⊆ taskIds`).
- The extra `!subTaskIds` (undefined) guard is unreachable at the real call site — `parentTask` is null-checked above it and `subTaskIds` is a required field.

Non-blocking overall — this is a value/readability call, not a defect. Thanks again for the contribution! 🙏

Rishi943:
Thanks for the careful trace, @johannesjo! Agreed on the split verdict — `getReorderedSubTaskIds` was solving duplication that `reorderSubTask` had already solved. Pushed d4f63ae: dropped `get-reordered-sub-task-ids.ts` + its spec and restored the original inline membership check (that function is now byte-identical to master), keeping only `moveValidIdsToFront`.

johannesjo:
Thank you very much! <3

PR #8933, opened 2026-07-11
test(sync): pin 3-client disjoint-merge composition (shared-field reconciliation + replace-mode convergence)

aakhter:
Test-only follow-up to #8874, carrying the spec from [this comment](https://github.com/super-productivity/super-productivity/pull/8874#issuecomment-4947064925) as offered.

A focused composition test (not a transport e2e) for the 3-client sequence that thread ultimately rests on: a disjoint-merged op loses whole-op LWW to a newer overlapping edit from a third client (C), and C emits a full-snapshot local-win op that has to reconcile the other clients. Two tests pin the two things that are actually true here:

1. **Op shape + shared-field reconciliation.** `_createLocalWinUpdateOp` must emit a full snapshot with a dominating clock; the test goes red if it ever regresses to a partial delta. This is the closure property the original thread asked about.
2. **Replace-mode convergence at the real seam.** The second test drives the production path (`convertOpToAction` -> `lwwUpdateMetaReducer`, plus a JSON round-trip). Because the local-win op carries `lwwUpdateMode: 'replace'`, the reducer applies it via `setOne`, which clears a receiver-only field the loser had absorbed, so A converges onto C for that upload ordering.

**Scope (receiver-version-dependent).** The convergence in test 2 holds only for receivers that honour `lwwUpdateMode: 'replace'`. The disjoint merge that sets up the scenario is enabled in production (unfrozen by #9095), and its merged op is an ordinary `patch` that any client applies via `updateOne`. But the replace-mode local-win op is newer: a receiver predating replace-mode treats it as a plain LWW Update, applies C's snapshot via `updateOne`, keeps the absorbed field, and the receiver-only-field divergence persists. So this is a replace-aware-clients claim, not universal convergence; the residual older-client divergence is tracked separately.

Single spec file, no runtime changes, so no docs updates apply. Rebased onto current master.

### DIFF
--- a/src/app/op-log/sync/conflict-resolution.disjoint-merge.spec.ts
+++ b/src/app/op-log/sync/conflict-resolution.disjoint-merge.spec.ts
@@ -1534,7 +1534,16 @@ describe('ConflictResolutionService — SPAP-14 disjoint-field merge', () => {
     });
   });
 
-  describe('(e2e) three-client: later overlapping edit beats the merged op', () => {
+  // A FOCUSED COMPOSITION TEST, not a transport e2e: it drives
+  // ConflictResolutionService with hand-built EntityConflicts and composes the
+  // ops it emits with a local LWW `applyOp` helper. It does NOT exercise real
+  // conflict detection, the OperationApplierService, or any sync transport —
+  // cross-client propagation is MODELLED (justified by the dominating-clock
+  // assertions below), not executed. Client B contributes only an input op, not
+  // a tracked client. Its value: it goes red if `_createLocalWinUpdateOp` ever
+  // emits a partial delta instead of a full snapshot, which is exactly what lets
+  // the three clients reconverge.
+  describe('composition (3-client): a later overlapping edit beats the merged op', () => {
     // Adjudicates the "partial merged delta is not closed under later LWW
     // composition" concern: after a merged op loses whole-op LWW to a newer
     // overlapping edit from a third client, the clients are TRANSIENTLY apart
@@ -1560,14 +1569,17 @@ describe('ConflictResolutionService — SPAP-14 disjoint-field merge', () => {
 
       const opLogStore = jasmine.createSpyObj('OperationLogStoreService', [
         'appendBatchSkipDuplicates',
+        'appendMixedSourceBatchSkipDuplicates',
         'appendWithVectorClockUpdate',
         'markApplied',
         'markRejected',
         'markFailed',
         'getUnsyncedByEntity',
         'mergeRemoteOpClocks',
+        'markReducersCommittedAndMergeClocks',
       ]);
       opLogStore.mergeRemoteOpClocks.and.resolveTo(undefined);
+      opLogStore.markReducersCommittedAndMergeClocks.and.resolveTo(undefined);
       opLogStore.getUnsyncedByEntity.and.resolveTo(new Map());
       opLogStore.markRejected.and.resolveTo(undefined);
       opLogStore.markApplied.and.resolveTo(undefined);
@@ -1580,6 +1592,18 @@ describe('ConflictResolutionService — SPAP-14 disjoint-field merge', () => {
           skippedCount: 0,
         }),
       );
+      // #8900 seam: local-win / merged ops now persist through the atomic
+      // mixed-source batch, mirroring the outer suite's setup.
+      opLogStore.appendMixedSourceBatchSkipDuplicates.and.callFake(async (batches) => ({
+        written: batches.flatMap((batch) =>
+          batch.ops.map((batchOp, index) => ({
+            seq: index + 1,
+            op: batchOp,
+            source: batch.source,
+          })),
+        ),
+        skippedCount: 0,
+      }));
 
       const validate = jasmine.createSpyObj('ValidateStateService', [
         'validateAndRepairCurrentState',
@@ -1614,10 +1638,12 @@ describe('ConflictResolutionService — SPAP-14 disjoint-field merge', () => {
       const svc = TestBed.inject(ConflictResolutionService);
       await svc.autoResolveConflictsLWW([conflict]);
 
-      const appended = opLogStore.appendWithVectorClockUpdate.calls
+      const appended = opLogStore.appendMixedSourceBatchSkipDuplicates.calls
         .allArgs()
-        .map(([o]) => o as Operation)
-        .filter((o) => o.entityId === 'task-1' && o.opType === OpType.Update);
+        .flatMap(([batches]) => batches)
+        .filter((batch) => batch.source === 'local')
+        .flatMap((batch) => [...batch.ops])
+        .filter((o: Operation) => o.entityId === 'task-1' && o.opType === OpType.Update);
       const applied = applier.applyOperations.calls
         .allArgs()
         .flatMap(([ops]) => ops as Operation[])
@@ -1647,7 +1673,7 @@ describe('ConflictResolutionService — SPAP-14 disjoint-field merge', () => {
       notes: s['notes'],
     });
 
-    it('reconverges all clients via the full-snapshot local-win op (loss is consistent, not divergent)', async () => {
+    it('the full-snapshot local-win op reconciles the transient split (loss is consistent, not divergent)', async () => {
       // Round 1: A (title) and B (notes) conflict -> disjoint merge on A.
       const opA = op({
         id: 'op-A',
@@ -1689,6 +1715,10 @@ describe('ConflictResolutionService — SPAP-14 disjoint-field merge', () => {
         id: 'task-1',
         title: 'C-title',
         notes: 'base',
+        // A field touched by NEITHER opC nor the merged delta. A partial
+        // (changed-fields-only) local-win op would drop it, so it makes the
+        // "full snapshot" assertion below test more than the two edited fields.
+        tagIds: ['keep-me'],
       };
 
       // Round 2 on A: local merged op vs remote opC. Fields overlap on title
@@ -1720,21 +1750,47 @@ describe('ConflictResolutionService — SPAP-14 disjoint-field merge', () => {
       }
       const localWinOp = r2c.appended[0];
       expect(localWinOp).toBeDefined();
-      // Full snapshot, not a partial delta: it must carry notes (=base), the
-      // field opC never touched — this is the closure property. If this op
-      // were partial, nothing would ever reconcile A's B-notes with C.
-      const localWinPayload = localWinOp.payload as Record<string, unknown>;
-      expect('notes' in localWinPayload).toBe(true);
+      // Full snapshot, not a partial changed-fields-only delta: applied to a
+      // BARE base the op alone must reconstruct C's COMPLETE entity — every
+      // field, including notes (=base, untouched by opC) AND tagIds (touched by
+      // neither opC nor the merged delta). This is the closure property; a
+      // partial op would drop these and never reconcile A's B-notes with C.
+      expect(applyOp({ id: 'task-1' }, localWinOp)).toEqual({
+        id: 'task-1',
+        title: 'C-title',
+        notes: 'base',
+        tagIds: ['keep-me'],
+      });
 
-      // Propagation: the local-win op's clock dominates both sides, so it
-      // reaches A as a plain (non-conflicting) remote op and applies directly.
+      // The local-win op's clock DOMINATES both inputs (opC and the merged op),
+      // so in production it reaches A as a plain non-conflicting remote op that
+      // applies directly — which is what the modelled propagation below assumes.
+      expect(compareVectorClocks(localWinOp.vectorClock, opC.vectorClock)).toBe(
+        VectorClockComparison.GREATER_THAN,
+      );
+      expect(compareVectorClocks(localWinOp.vectorClock, mergedOp.vectorClock)).toBe(
+        VectorClockComparison.GREATER_THAN,
+      );
+
+      // Modelled propagation (justified by the dominating clock above): the
+      // local-win op reaches A as a plain remote op and applies directly.
       stateA = applyOp(stateA, localWinOp);
 
-      // CONVERGENCE: every client ends on the identical entity. B-notes are
-      // lost to LWW — but CONSISTENTLY on all clients (and journaled by the
-      // round-2 resolutions for review/flip), never as silent divergence.
-      expect(pick(stateA)).toEqual(pick(stateC));
-      expect(pick(stateA)).toEqual({ title: 'C-title', notes: 'base' });
+      // CONVERGENCE (for THIS upload ordering — the merged op reaches C, so C
+      // emits the reconciling op): A and C end byte-identical, tagIds and all.
+      // B-notes are lost to LWW but CONSISTENTLY across clients (and journaled by
+      // the round-2 resolutions for review/flip), never as silent divergence.
+      // The inverse ordering (opC globally accepted first, so C never downloads
+      // the merged op) is a real-transport concern out of scope for this
+      // service-level composition test — it belongs to the multi-client sync
+      // harness (SPAP-34).
+      expect(stateA).toEqual(stateC);
+      expect(stateA).toEqual({
+        id: 'task-1',
+        title: 'C-title',
+        notes: 'base',
+        tagIds: ['keep-me'],
+      });
     });
   });
 });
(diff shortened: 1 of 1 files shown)


johannesjo:
## Review feedback — scope of the regression coverage

I double-checked current head `3611ca5` against the production conflict-routing, sync-order, server-rejection, operation-conversion, and LWW reducer paths.

The underlying LWW conclusion looks sound, and this test has real value: it will fail if the local-win operation becomes a simple partial delta that carries C's `title` but drops the untouched `notes` field. I do not see a new production sync defect in this test-only change.

There is still a material mismatch between what the test executes and its “e2e / all clients / snapshot propagation” claim:

- `resolveCapturing()` supplies prebuilt `EntityConflict` objects directly and replaces `OperationApplierService` with a spy that reports zero applied operations.
- The final propagation is forced with the local `applyOp()` helper. The comment says the new clock dominates both sides, but the test never asserts that comparison or sends the operation through production conflict detection/application.
- Client B authors an operation but never has tracked state; the final assertion compares only `title` and `notes` on A and C.
- For SuperSync specifically, download-before-upload plus server rejection means the two inverse round-2 conflicts are not one real transport trace.

Consequently, a regression in the reconciling operation's clock/routing/application could leave this test green even though actual clients no longer converge.

The lean fix would be to present this as a focused conflict-service composition test and explicitly assert that the local-win payload equals C's complete fixture and its clock is `GREATER_THAN` both inputs. If the `e2e` and “all clients” claim is retained, please drive A/B/C through the real remote-processing/applier path and compare all three resulting entities.

All current CI, SuperSync, and WebDAV checks are green.

aakhter:
Addressed in `801c4c0` (rebased on current master), taking the lean "focused composition test" path you outlined.

- **Reframed, honestly scoped.** Dropped the `(e2e)` / "all clients" framing. The block comment now states it is a composition test of `ConflictResolutionService`'s emitted ops applied through a local LWW helper — it does not exercise real conflict detection, the `OperationApplierService`, or any transport, and cross-client propagation is *modelled* (justified by the dominating-clock assertion), not executed. Client B is called out as an input-only op, not a tracked client.
- **Added the two assertions you asked for.** The local-win op now must (a) reconstruct C's **complete** entity from a bare base and (b) have a clock `GREATER_THAN` both inputs (`opC` and the merged op). To make "complete" mean more than the two edited fields, C's fixture carries a `tagIds` sentinel touched by neither `opC` nor the merged delta; the assertion is a full deep-equal of the applied op, and final convergence is now a full deep-equal of A vs C (tagIds and all), not just title/notes — so a partial (changed-fields-only) local-win op fails it.
- **Upload-ordering.** Per your download-before-upload / server-rejection point: the test now explicitly scopes itself to the ordering where the merged op reaches C (so C emits the reconciling op). The inverse ordering (`opC` globally accepted first, so C never downloads the merged op) is a real-transport concern I've left to the multi-client sync harness (SPAP-34) rather than modelling a second synthetic trace here.
- **Rebase note.** Rebasing onto current master surfaced that the local `resolveCapturing` helper mocked the pre-#8900 append API; updated it to `appendMixedSourceBatchSkipDuplicates` (mirroring the outer suite), so it matches the service again. 17/17 green.

johannesjo:
## Important (blocking): the “full snapshot” does not clear receiver-only optional fields

I rechecked current head `801c4c0` against the exact synthetic merge candidate and found a concrete counterexample to the closure/convergence claim.

Extend B's disjoint edit by one real optional field:

```ts
changes: {
  notes: 'B-notes',
  dueDay: '2026-07-15',
}
```

The merged op adds `dueDay` to A, while C never has that field. C's newer title still wins and its local-win op still has a dominating clock, but C's current-entity payload omits `dueDay`. Applying that op to A is a shallow update, so A keeps the stale value:

- A: `{ title: 'C-title', notes: 'base', dueDay: '2026-07-15', ... }`
- C: `{ title: 'C-title', notes: 'base', ... }`

I reproduced this directly in the added composition test: the unmodified merge candidate is 26/26 green; adding only the `dueDay` change above makes the final `stateA === stateC` assertions fail with the unexpected retained `dueDay`.

The production path has the same semantics:

- [`dueDay` is optional and `DEFAULT_TASK` omits it](https://github.com/super-productivity/super-productivity/blob/46007aa8911d1761ca9c7188c76768d8e8721d0b/src/app/features/tasks/task.model.ts#L103-L116).
- [SuperSync serializes operations with ordinary `JSON.stringify`](https://github.com/super-productivity/super-productivity/blob/46007aa8911d1761ca9c7188c76768d8e8721d0b/packages/sync-providers/src/super-sync/super-sync.ts#L210-L226), so absent/`undefined` fields cannot travel as clears.
- [The LWW reducer uses adapter `updateOne`](https://github.com/super-productivity/super-productivity/blob/46007aa8911d1761ca9c7188c76768d8e8721d0b/src/app/root-store/meta/task-shared-meta-reducers/lww-update.meta-reducer.ts#L587-L601), which preserves omitted receiver fields.
- [Existing reducer tests explicitly distinguish `dueDay: undefined` (clear) from an omitted due field (preserve)](https://github.com/super-productivity/super-productivity/blob/46007aa8911d1761ca9c7188c76768d8e8721d0b/src/app/root-store/meta/task-shared-meta-reducers/lww-update.meta-reducer.spec.ts#L2556-L2589).

Therefore, applying the local-win op to a bare base only proves that fields present on C can be reconstructed; it does not prove that fields present only on another client are removed. Because the new clock dominates both inputs, no later conflict is guaranteed to repair this silent divergence.

Please add this receiver-only-field case and exercise a JSON round-trip plus the production reducer. If the intended invariant is true convergence, the runtime needs to distinguish full local-win snapshots from partial merged LWW deltas and apply replacement/explicit-clear semantics for the former. If that runtime change is out of scope, the PR should narrow its title/comments/assertions to the precise present-field + dominating-clock property rather than claiming full-snapshot closure or identical entities.

**Verdict: needs changes.**

aakhter:
Thanks — you're right, and I reproduced the counterexample. Applying C's full local-win snapshot to A can't clear a field C never had: SuperSync's `JSON.stringify` drops the absent field at the wire boundary, and `lwwUpdateMetaReducer` applies the op via a shallow `updateOne`, so A keeps the field and the clients diverge — with a dominating clock, permanently. The "identical entities / full-snapshot closure" assertion was overclaiming.

Updated the PR (head `e7a650cea`):

1. **Narrowed the active composition test** to the property that actually holds — present-field reconstruction + a dominating clock + shared-field reconciliation. It no longer asserts byte-identical whole-state equality, and the comments say so.

2. **Replaced the `pending()` aspiration with an enabled test** that asserts today's *actual* divergence, exercised exactly as you asked — a JSON round-trip **plus** the production reducer. It runs the real resolution service to get C's local-win op, `JSON.parse(JSON.stringify(...))` its payload, and applies it through the production `lwwUpdateMetaReducer` (real `updateOne` + relationship repair). It asserts A retains the receiver-only `dueDay` while C lacks it (the divergence), and that shared fields (`notes`) reconcile to C's value — which also proves the snapshot really went through the reducer rather than a pass-through/bailout.

3. The underlying runtime gap — a full local-win snapshot can't clear a receiver-only field the loser absorbed, so clients silently diverge — is a real data-integrity issue. I've logged it as a separate item for a dedicated runtime fix: the wire format needs to distinguish a full snapshot from a partial LWW delta and apply replacement / explicit-clear semantics that survive JSON transport. The test carries a "flip these assertions when the runtime reconciles receiver-only fields" note so it converts into a convergence guard once that lands.

Title/scope updated accordingly. Suite green locally.

johannesjo:
## Review — needs changes before merge

Thanks for pinning this property down, the intent is exactly right. But the branch is **27 commits behind master** and the sync internals it targets moved underneath it, so both new tests are red on current master. I dug in and verified end-to-end.

### 🔴 Blocking: the two new tests fail on current master

CI already shows this (`Tests` job: **2 FAILED, 12814 SUCCESS**), and I reproduced it locally by merging this branch into master — `2 FAILED, 26 SUCCESS` in the file, both failures `Expected $.notes = 'base' to equal 'B-notes'`. So "Green on current master" in the description is stale — it was green against the branch's base (`a9ed0b2df2`), not today's `master`.

**Root cause — the LWW op payload shape changed.** At this branch's base, `createLWWUpdateOp` emitted a **flat** payload (`{ ...entityState, id }`). Master (via the multi-entity conflict work, #8990 / #9007) reshaped it to a **nested** payload:

```
{ actionPayload, entityChanges, lwwUpdateMode }
```

The hand-rolled `applyOp` model (lines ~1051–1063) only understands the old flat / `{ task: { changes } }` shapes, so the merged op's fields are never applied. That's why `notes` stays `'base'` at the first post-merge assertion (line ~1097), and why `dueDay` is never absorbed in the second test.

### 🟠 Deeper issue: the second test now documents behavior that no longer exists

The `does NOT clear a receiver-only field the loser absorbed` test (line 1201) is premised on the reducer applying the local-win op via a shallow `updateOne`. On master that's no longer the path:

- `_createLocalWinUpdateOp` now stamps `lwwUpdateMode: 'replace'` (`conflict-resolution.service.ts:2536`),
- and `lwwUpdateMetaReducer` applies replace-mode ops via `setOne` — a **full replace** (`lww-update.meta-reducer.ts:787–791`).

I confirmed the consequence by driving the real production path (`convertOpToAction` → `lwwUpdateMetaReducer`) with a replace-mode snapshot that lacks `dueDay`, applied to a task that has it: **`dueDay` is cleared → the clients converge.** So the divergence this test asserts as "today's ACTUAL behavior" (line 1320: `expect(aTask['dueDay']).toBe('2026-07-15')`) doesn't reproduce via the mechanism it describes — the `'replace'` / `setOne` work appears to have already closed that specific SPAP-43 gap.

One more thing on that test: it hand-builds the action (`...wirePayload` spread at top level + a `meta` without `lwwUpdateMode`, lines 1296–1297) rather than going through `convertOpToAction`. Production spreads `extractActionPayload(...)` at the top level and puts `lwwUpdateMode` **into `meta`** (`operation-converter.util.ts:255–330`), which is what actually selects the `setOne` branch — so as written it wouldn't exercise the real seam even once the payload shape is fixed.

### 🟡 Design note for the rewrite

The fragility is structural: the tests reimplement op-application in a parallel `applyOp` model, and that model silently drifted from the production payload shape — which is precisely how this broke without anyone noticing. Since the whole value of these tests is pinning a **closure property of the real ops**, they'll be far more durable applying ops through the real seam (`convertOpToAction` + `lwwUpdateMetaReducer`, which the second test half-attempts) instead of a hand-rolled model that can rot independently.

### Suggested path

1. Rebase onto current master.
2. Migrate the tests to the nested `{ actionPayload, entityChanges, lwwUpdateMode }` op shape, and apply ops via `convertOpToAction` + the production reducer rather than `applyOp`.
3. Re-derive the actual behavior: on master the receiver-only field is **cleared** (convergence), so the second test's framing needs to flip — or, if the property it was pinning is now fully subsumed by the `replace`/`setOne` path, reconsider whether the test still earns its keep.

One caveat on my part: I disproved only the *specific* `updateOne`-can't-clear mechanism this test names — not every class-level op-log divergence residual we flagged during #8874. If you still believe there's a receiver-only-field divergence reachable by a different path, it'd be worth constructing that scenario explicitly before dropping the coverage.

johannesjo:
## Review — the substance is right now; one 4-line blocker

Thanks for reworking this. You addressed the round-2 asks properly: the second test drives the real seam (`convertOpToAction` → `lwwUpdateMetaReducer` + JSON round-trip), and its framing is flipped to master's actual replace/`setOne` behaviour. I re-verified everything below against current master (`b50d5f6d96`).

### ✅ First: both tests are real guards, and the second one earns its keep

I asked in round 2 whether the second test is now subsumed by the `replace`/`setOne` path. It isn't — I mutation-tested both:

| Mutation to production code | Result |
|---|---|
| `_createLocalWinUpdateOp`: `'replace'` → `'patch'` | test 2 **red** (`Expected '2026-07-15' to be undefined`) |
| `_createLocalWinUpdateOp`: full snapshot → partial delta (locally-touched fields only) | test 1 **red** |

Test 2 pins the **service→reducer contract** — that the service stamps replace-mode — which `lww-update.meta-reducer.spec.ts` cannot catch, since it only tests the reducer *given* a replace-mode action. That's genuine coverage. Consider my round-2 "does it still earn its keep?" question answered.

### 🔴 Blocking: the new mock is stale → CI red

CI (`Tests`: 2 FAILED / 13122 SUCCESS) fails with `TypeError: opLogStore.getOpById is not a function` at `conflict-resolution.service.ts:4067`. I reproduced it exactly by merging this branch into master (2 FAILED / 26 SUCCESS in the file, clean merge — no conflicts).

The cause is entirely in the new `resolveCapturing` helper's `createSpyObj` (line ~1578):

- **`appendWithVectorClockUpdate`** (line 1581, 1595) → renamed to `appendWithVectorClockOverwrite` by #9080 (`d1ff7963d0`). The old name now exists **nowhere** in the codebase, so that spy entry is silent dead weight.
- **`getOpById` is missing** → add it plus `opLogStore.getOpById.and.resolveTo(undefined)`. It entered the service via #9086 (`5a50297215`, `_withoutSyncedOps`).

That 4-line fix alone turns the file green (28/28), and the full unit suite is green with it (13112 SUCCESS, 0 failures).

**Worth knowing: a rebase alone will not fix this.** On this branch the file is self-consistent (all three mocks use the old name), which is why it's green against your base. Master's rename/addition only touches the *two pre-existing* mocks; your newly-added block is new text that git carries verbatim through both merge and rebase. You'll need to update it by hand.

### 🟠 The describe-level comments now contradict the test they introduce

Lines 1547–1549 and 1562 still carry the pre-flip framing:

> a full snapshot *"still cannot clear a field the winner never had but the loser absorbed … because the reducer applies it via a shallow `updateOne`"*
> the second test … *"asserting today's ACTUAL divergence"*

But test 2 now asserts the opposite — the field **is** cleared via `setOne`, and the clients converge. You updated the per-test comment; these two headers were left behind. For a PR whose entire deliverable is documenting a subtle sync property, a reader of the describe header would take away the exact inverse of what the code proves.

Same for the metadata: the **title** still says *"document receiver-only-field divergence"* (now inverted), and the **body** still describes the dropped e2e framing (*"asserts all clients land on the identical entity"*) and claims *"Green on current master"*. The title is what lands in git history.

### 🟠 Scope the convergence claim — no released client understands `lwwUpdateMode`

This one partially vindicates your original framing, so it's worth getting right rather than just deleting the divergence story.

`lww-update.meta-reducer.ts:797-802` branches `lwwUpdateMode === 'replace' ? setOne(...) : updateOne(...)` — so an op **without** the field falls back to shallow merge. Checked against the `v18.14.0` tag: **zero** occurrences of `lwwUpdateMode`, and no disjoint-merge code at all.

So the convergence test 2 pins holds only between clients that understand replace-mode — i.e. unreleased ones. A released receiver applies C's snapshot via `updateOne`, keeps the absorbed `dueDay`, and the divergence you originally documented stands. Today that's unreachable (see below), but it becomes reachable in a mixed fleet the moment the merge path is unfrozen. Suggest the comment says "converge" **conditioned on the receiver honouring `lwwUpdateMode`**, rather than as a fleet-wide property. (Version facts verified; I did not run a mixed-fleet test.)

### 🟡 Context: this path is frozen in production today

`remote-ops-processing.service.ts:482` hardcodes `disableDisjointMerge: true` at the only production entry to `autoResolveConflictsLWW` (#9061 producer freeze), so no client currently produces merged ops and this 3-client scenario cannot occur in the field. Your tests call `autoResolveConflictsLWW([conflict])` with no options, so they exercise the unfrozen default.

Not a blocker — the whole host suite shares this property, and the freeze is explicitly a one-line revert. But it's the honest scope: these tests pin intended behaviour that's currently switched off, and they become load-bearing on unfreeze. Worth a line in the block comment so the next reader isn't misled about what's live.

### 🟡 The mock duplication is the actual root cause

This is now the **third** copy of the `OperationLogStoreService` mock in this one file. Master maintains two (lines ~138, ~1402) and had to update **both** when `getOpById` landed; your third copy was cloned from a pre-#9086 revision and rotted within a day. `jasmine.createSpyObj` never validates names against the real class, so these fail only at call time. Extracting a shared factory would have made this PR immune — and it's the same drift that broke the `applyOp` model last round, just relocated to the mock layer.

Minor, related: test 1 still uses the hand-rolled `applyOp` rather than the seam test 2 demonstrates. Your justification (the real reducer stamps `modified` and repairs relationships a field-level model omits) is fair — noting it only because it's the same parallel-model risk.

### Verdict: needs changes — but only mechanically

Rebase, hand-fix the 4 mock lines, reconcile the two describe comments + title/body with what the tests now assert, and scope the convergence claim to replace-mode-aware receivers. The hard part — correct property, real seam, non-vacuous assertions — is done. Happy to approve once CI is green.

johannesjo:
Thank you very much! <3

PR #8943, opened 2026-07-11
feat(sync): whole-dataset conflict "Review differences" merge mode

aakhter:
## What

The whole-dataset conflict dialog (shown on `LocalDataConflictError`, e.g. a concurrent compacted snapshot with no incremental ops to LWW-merge) was all-or-nothing: USE LOCAL or USE REMOTE. This adds a third path: **review the actual differences and merge per item**.

Follow-up to #8874 (conflict journal + disjoint auto-merge + review UI), completing the conflict-review work.

## How it works

- **Entity diff** (`whole-dataset-diff.util.ts`, pure/unit-testable): local complete state (read via the same snapshot path force-upload uses) vs the downloaded remote snapshot, per reviewable entity type → *differing* (with per-field values), *only-local*, *only-remote*. Differences confined to `NOISE_FIELDS` are excluded. Header counts come from this diff, **not** vector-clock sums.
- **"REVIEW N DIFFERENCES" button** in the conflict dialog: diff is computed lazily with a spinner; the button is disabled when the diff is empty or its computation failed. Review is only offered when the error carries a **non-empty** remote snapshot — the fresh-client path throws with `{}`, and a diff against it would misreport every local entity as "only-local" (its discards could drop real data).
- **Blocking review modal**: differing → LOCAL/REMOTE pick (preselected newest-wins by entity `modified`), only-local → KEEP/DISCARD, only-remote → ADD/SKIP, with bulk bars; usable at phone width. Cancel returns to the 3-button dialog with sync still paused.
- **APPLY MERGE** (`whole-dataset-merge.service.ts`), in this exact order:
  1. builds the merged state from the picks (singleton slices + archives carried through from the local base unchanged) and applies it locally via `SyncHydrationService.hydrateFromRemoteSync` — the same full-state hydrate path USE REMOTE uses;
  2. journals every **non-default** pick as reason `manual-merge`, status `kept` (History tab). Journal-after-persist, but *before* the upload: an upload failure is retried by normal sync and never re-enters this flow, so journaling after it would lose the entries;
  3. force-uploads via the `forceUploadLocalState` path so remote receives the merged result. Apply-then-upload: on upload failure the merged state is already local and normal retry re-uploads it.
- `manual-merge` entries are **not flippable** (`FLIP_UNSUPPORTED_REASONS`): the user already reviewed both sides and explicitly picked one, and the merged result was force-uploaded — a later flip would silently diverge local state from the uploaded merge.
- USE LOCAL / USE REMOTE behavior is unchanged. i18n added to `en.json` only, per contribution guidelines.

## Data safety notes (worth close review)

This touches the whole-dataset conflict path and force-upload. The guards, each pinned by a test:

- review not offered on an empty `{}` remote snapshot (fresh-client path);
- exact applyMerge sequence (hydrate → journal → force-upload) asserted via invocation order;
- journal entries survive a failed upload;
- merged state matches the picks exactly (incl. keep/discard/add/skip);
- cancel (review modal or dialog) leaves sync in the same paused-conflict state;
- `canFlip()` false for `manual-merge`.

Deferred (called out rather than half-done): a full `LocalDataConflictError` → mock adapter → second-client round-trip integration test; the service spec pins the invariant via invocation assertions instead.

## Docs

- `docs/wiki/4.23-Managing-Your-Data.md`: conflict section extended with the review flow.
- `docs/sync-and-op-log/conflict-journal-and-review.md`: new "Whole-dataset merge review" section + `manual-merge` flip rule.

## Verification

- Affected suites all green (454 specs): whole-dataset diff/merge utils + service, dialog-sync-conflict, dialog-conflict-review, sync-wrapper (135), conflict journal/review/UI, backup service.
- `tsc --noEmit`, `eslint`, `prettier` clean.


johannesjo:
## Multi-review feedback — needs changes

I reviewed head `5b2a5c8` against `master` in three independent passes, then re-checked the findings against the surrounding snapshot, hydration, validation, model, and UI code. All current CI checks are green, but the following data-safety paths are not exercised by the added tests. I consider the first two merge blockers.

### Critical

1. **Applying a review silently overwrites every synced slice that is not in `REVIEWABLE_MODELS`.** The review diff [deliberately excludes](https://github.com/super-productivity/super-productivity/blob/5b2a5c8322db7953537272689fa42064c63495de/src/app/op-log/sync/whole-dataset-diff.util.ts#L35-L54) `globalConfig`, `timeTracking`, `menuTree`, `planner`, `boards`, plugin data, and both archives, although they are all part of [`AppDataComplete`](https://github.com/super-productivity/super-productivity/blob/5b2a5c8322db7953537272689fa42064c63495de/src/app/op-log/model/model-config.ts#L48-L70). The merge then [clones the local state and only changes reviewed entities](https://github.com/super-productivity/super-productivity/blob/5b2a5c8322db7953537272689fa42064c63495de/src/app/op-log/sync/whole-dataset-merge.util.ts#L163-L210), before force-uploading that result. Therefore, if a remote archive/time record/setting/planner/plugin value differs alongside one reviewable task, the dialog reports only the task difference and **Apply merge replaces the unshown remote data with the local copy**. At minimum, `REVIEW` should be unavailable when any unhandled slice differs; the complete fix needs an explicit merge/carry policy for every synced slice, derived exhaustively from `MODEL_CONFIGS`. Please add a mixed task + archive/timeTracking/config regression test.

2. **The cached local snapshot can roll back edits made while the user is reviewing.** The conflict dialog eagerly invokes the provider, and the wrapper [memoizes both `diff` and `localState`](https://github.com/super-productivity/super-productivity/blob/5b2a5c8322db7953537272689fa42064c63495de/src/app/imex/sync/sync-wrapper.service.ts#L1324-L1341), then later [applies that cached state](https://github.com/super-productivity/super-productivity/blob/5b2a5c8322db7953537272689fa42064c63495de/src/app/imex/sync/sync-wrapper.service.ts#L1396-L1418). `UserInputWaitStateService` only prevents the sync timeout; it does not stop mutations. In particular, active time tracking continues updating state and flushes sync operations every five minutes. Hydration creates a clean-slate `SYNC_IMPORT` using the current clock, so those newer operations can be causally dominated even though their values are absent from the stale snapshot. Re-read/version-check local state immediately before Apply and recompute the review if it changed, under the same mutation barrier used for full-state imports. Please cover an edit/time-tracking flush while the dialog is open.

### Important

3. **Independent entity picks can create a state that does not represent the user's choices.** [`buildMergedState`](https://github.com/super-productivity/super-productivity/blob/5b2a5c8322db7953537272689fa42064c63495de/src/app/op-log/sync/whole-dataset-merge.util.ts#L167-L210) replaces/removes tasks, projects, tags, notes, repeat configs, and sections independently, despite reciprocal relationships such as `task.projectId` ↔ `project.taskIds`, tags, parent/subtasks, notes, and section ordering. Valid local and remote states can therefore produce an invalid mixed graph. Post-hydration repair may silently rewrite the picks; an unrepaired candidate follows the failure path below. Related entities need dependency-aware/atomic choices, and the candidate should be validated and any repairs previewed before persistence.

4. **“Newest wins” is usually not using an entity timestamp, and ties contradict the documented rule.** The diff [reads only `modified`, defaulting missing values to `0`](https://github.com/super-productivity/super-productivity/blob/5b2a5c8322db7953537272689fa42064c63495de/src/app/op-log/sync/whole-dataset-diff.util.ts#L136-L141), while the default [chooses remote only when it is strictly greater](https://github.com/super-productivity/super-productivity/blob/5b2a5c8322db7953537272689fa42064c63495de/src/app/op-log/sync/whole-dataset-merge.util.ts#L52-L58). Projects and tags use `updated`; counters, repeats, metrics, providers, sections, and reminders have no `modified` timestamp. Most changed entities therefore default to local (`0 === 0`), and equal task/note timestamps also choose local although the sync contract and wiki say ties go to remote. Use a trustworthy model-specific timestamp or make unknown/tied cases explicit instead of presenting them as newest-wins, with per-model and tie coverage.

5. **Validation failure can still upload the candidate and report `IN_SYNC`.** Hydration [sets the session failure latch but continues persisting and dispatching invalid data](https://github.com/super-productivity/super-productivity/blob/5b2a5c8322db7953537272689fa42064c63495de/src/app/op-log/persistence/sync-hydration.service.ts#L248-L302). Unlike `USE_REMOTE`, the review branch neither resets nor checks that latch and [sets `IN_SYNC` unconditionally](https://github.com/super-productivity/super-productivity/blob/5b2a5c8322db7953537272689fa42064c63495de/src/app/imex/sync/sync-wrapper.service.ts#L1396-L1418). Moreover, `forceUploadLocalState()` has a `void` result and uploads pending ops even if migration validation returned early. The review flow needs a typed successful-apply/upload outcome, must abort before upload on invalid data, and must use the same latch handling as `USE_REMOTE`.

6. **Issue-provider secrets are rendered and journaled verbatim.** `ISSUE_PROVIDER` is reviewable, and those entities contain fields such as Jira/CALDAV/Nextcloud passwords, GitLab/OpenProject/Plainspace tokens, and arbitrary nested `pluginConfig`. The dialog [renders every raw field value](https://github.com/super-productivity/super-productivity/blob/5b2a5c8322db7953537272689fa42064c63495de/src/app/imex/sync/dialog-conflict-review/dialog-conflict-review.component.html#L53-L57), and non-default choices [copy both values into the local conflict journal](https://github.com/super-productivity/super-productivity/blob/5b2a5c8322db7953537272689fa42064c63495de/src/app/op-log/sync/whole-dataset-merge.service.ts#L150-L175). Please deeply redact secret-shaped fields in both UI and persisted journal data; raw values should remain only in the private in-memory merge payload.

7. **The whole-dataset dialog eagerly renders and stringifies the entire conflict set.** All entities and all field diffs are expanded with nested `@for` loops and `display()`/`JSON.stringify`, with no paging, virtualization, or collapse. A conflict involving thousands of tasks can freeze the very dialog required to recover the data. Please bound rendered work (paging/virtualization or collapsed chunks) and add a large-dataset regression/performance check.

### Smaller issues

- The local/remote and keep/discard controls expose selected state only through color; they need radio/toggle semantics or `aria-pressed` so assistive technology can determine the current choice.
- The new SCSS hardcodes spacing, colors, radius, font size, and a breakpoint, contrary to `docs/styling-guide.md`; the confirming **Apply merge** action should also be `mat-flat-button color="primary"`, not stroked.

The new unit tests currently use mostly synthetic task-only states, and the wrapper tests only establish whether a review provider is present. They do not cover the mixed-slice loss, mutations during review, relationally inconsistent picks, validation failure, real per-model timestamps, secret redaction, or large datasets above.

johannesjo:
## Second-pass multi-review — current head `aa649733` still needs changes

I re-fetched the current head, full diff, tests, top-level discussion, thread state, and CI, then re-traced the snapshot/hydration/force-upload path. There are no inline review threads; the only substantive feedback is the earlier top-level review at `5b2a5c8`.

The required `baselineVectorClock` and pre-apply cutoff are useful improvements: they catch edits that happen while the modal is open and before `applyMerge()` reaches its final check. They do not close the destructive window below.

### Critical

1. **Applying Review still overwrites synced slices that are not shown.** The diff [explicitly excludes singleton slices and archives](https://github.com/super-productivity/super-productivity/blob/aa649733dbd36a4a3b764a3c21e3b6c4408e50c1/src/app/op-log/sync/whole-dataset-diff.util.ts#L35-L54), and also omits boards and plugin data, although these are all part of the [complete snapshot](https://github.com/super-productivity/super-productivity/blob/aa649733dbd36a4a3b764a3c21e3b6c4408e50c1/src/app/op-log/backup/state-snapshot.service.ts#L50-L73). The merge [clones the local base and only applies reviewed entity picks](https://github.com/super-productivity/super-productivity/blob/aa649733dbd36a4a3b764a3c21e3b6c4408e50c1/src/app/op-log/sync/whole-dataset-merge.util.ts#L167-L210), then force-uploads it. Trigger: one visible task conflict plus a remote archive/time-tracking/config/planner/plugin change. Impact: Apply silently replaces that unshown remote data with local data. Minimum safe fix: disable Review whenever any unhandled persisted slice differs, using an exhaustive model policy; add mixed task + archive/time/config coverage.

2. **A local operation can still land after the stale check and be lost by clean-slate hydration.** The exclusive clock check ends at [line 150, before hydration begins at line 161](https://github.com/super-productivity/super-productivity/blob/aa649733dbd36a4a3b764a3c21e3b6c4408e50c1/src/app/op-log/sync/whole-dataset-merge.service.ts#L125-L166); the code itself records this as a known residual. An action captured in that gap is absent from the cached merge and can be dominated by the new import. The new integration test [only calls `setVectorClock({devA: 2})`](https://github.com/super-productivity/super-productivity/blob/aa649733dbd36a4a3b764a3c21e3b6c4408e50c1/src/app/op-log/sync/whole-dataset-merge.stale-review.integration.spec.ts#L101-L139): it appends no operation, runs the mutation before `applyMerge()`, and stubs hydration, so it cannot reproduce the remaining window or prove an actual operation survives. The final check and hydration need one shared deferred-action/mutation barrier, with a real captured operation inserted in that exact window.

### Important

3. **Independent picks can create an invalid cross-entity graph, and validation happens after destructive persistence.** Tasks, projects, tags, repeats, notes, and sections are independently replaced or removed despite reciprocal IDs and ordering relationships. A valid local state plus a valid remote state can therefore produce an invalid combination. Hydration appends the import before validation and then [persists/loads the repaired or still-invalid candidate](https://github.com/super-productivity/super-productivity/blob/aa649733dbd36a4a3b764a3c21e3b6c4408e50c1/src/app/op-log/persistence/sync-hydration.service.ts#L225-L395). Repairable input silently changes the user's picks; irreparable input has already replaced the local baseline before the later force-upload validation aborts. Please validate the complete candidate before any durable replacement and surface required coupled choices/repairs.

   **Correction to my earlier comment:** I withdraw the broader claim that an irreparably invalid candidate is uploaded and then reported `IN_SYNC`. `ServerMigrationService` revalidates before creating the upload import and the coordinator throws if no import was created. The confirmed problem is the earlier local destructive apply/silent repair described above.

4. **“Newest wins” usually has no timestamp and ties choose the wrong side.** The diff [reads only `modified`, defaulting missing values to zero](https://github.com/super-productivity/super-productivity/blob/aa649733dbd36a4a3b764a3c21e3b6c4408e50c1/src/app/op-log/sync/whole-dataset-diff.util.ts#L136-L141), while defaults [choose local for ties/missing timestamps](https://github.com/super-productivity/super-productivity/blob/aa649733dbd36a4a3b764a3c21e3b6c4408e50c1/src/app/op-log/sync/whole-dataset-merge.util.ts#L52-L58). Most reviewable model types do not have `modified`; equal task/note timestamps also contradict the documented remote-wins tie rule. Use a trustworthy per-model timestamp or make unknown/tied cases explicit, with per-model coverage.

5. **Issue-provider credentials are rendered and duplicated into the conflict journal.** `ISSUE_PROVIDER` is reviewable; those configurations contain passwords/tokens. The dialog [renders raw field values](https://github.com/super-productivity/super-productivity/blob/aa649733dbd36a4a3b764a3c21e3b6c4408e50c1/src/app/imex/sync/dialog-conflict-review/dialog-conflict-review.component.html#L53-L57), and non-default picks [persist both sides verbatim](https://github.com/super-productivity/super-productivity/blob/aa649733dbd36a4a3b764a3c21e3b6c4408e50c1/src/app/op-log/sync/whole-dataset-merge.service.ts#L223-L249). Deep-redact secret-shaped/nested provider fields in both UI and journal; raw values should remain only in the transient merge payload.

6. **The Review path discards the force-upload result and can report `IN_SYNC` with unresolved operations.** `forceUploadLocalState()` returns `{hasUnresolvedOps}`, but [the merge service ignores it](https://github.com/super-productivity/super-productivity/blob/aa649733dbd36a4a3b764a3c21e3b6c4408e50c1/src/app/op-log/sync/whole-dataset-merge.service.ts#L168-L179), after which the wrapper sets `IN_SYNC`. The sibling USE_LOCAL branch already maps this condition to `UNKNOWN_OR_CHANGED`. Propagate the typed result/use the existing pending-ops failure path and add a REVIEW regression.

7. **The required recovery UI does unbounded eager work.** The first dialog [starts the whole diff immediately](https://github.com/super-productivity/super-productivity/blob/aa649733dbd36a4a3c21e3b6c4408e50c1/src/app/imex/sync/dialog-sync-conflict/dialog-sync-conflict.component.ts#L91-L110), and the review template expands every entity and field while stringifying arbitrary values. There is no paging, virtualization, or collapse. Bound the rendered/computed work and add a large-dataset check.

### Additional repository-fit issues

- One Apply currently creates a remote-source `SYNC_IMPORT` during hydration and then a second local `SYNC_IMPORT` during force upload. That conflicts with the one-intent/one-op invariant; consolidate to one authoritative import or document and test why two are required.
- The PR grows the grandfathered oversized `SyncWrapperService` from 1,821 to 1,930 lines, although repository guidance says that debt may only shrink. Extract the conflict-review orchestration.
- Pick state is visual-only; use `aria-pressed` or the existing segmented/radio semantics. The SCSS still hardcodes spacing/color/radius/font/breakpoint values, and Apply should be the flat primary dialog action.
- `supersync-scenarios.md` still specifies only USE_LOCAL / USE_REMOTE / CANCEL for the affected scenarios.

### CI and verdict

Unit tests, lint, build, WebDAV, and SuperSync shards 2–6 are green. SuperSync shard 1 and the aggregate gate are red on a Daily Summary navigation timeout before the backup/conflict path; this looks unrelated/flaky, but the required gate still needs a clean rerun.

**Verdict: needs changes.** The hidden-slice overwrite and check→hydrate operation-loss window remain merge blockers.

One product-level question also remains: the PR links no user request for arbitrary per-item whole-dataset composition. The reported destructive-sync need in #7709 can be addressed more safely with an exhaustive read-only preview/count/export plus the existing dataset-level choice. Given the permanent 2.5k-line sync/UI surface, that leaner path deserves explicit consideration.

johannesjo:
Thanks a lot for the work here, @aakhter — and for #8874, which landed and completed the bulk of the conflict-review story. Closing this one after some deliberation; writing down the reasoning for transparency:

- The motivating scenario (the #7709 wipe) was root-caused to the destructive `isCleanSlate` upload paths and fixed back in May — the dialog itself turned out to be downstream of the actual bug, and there's currently no user request on record for per-item whole-dataset merging.
- The two blockers from the July reviews are structural rather than patchable: the merge carries all non-reviewable slices (config, time tracking, planner, boards, archives, plugin data) through from local and force-uploads them, so "Review" can silently overwrite remote data the user never saw; and the check→hydrate window can still drop a local op. Fixing both properly means an exhaustive per-slice merge policy plus a shared mutation barrier — a lot of high-risk surface in the most dangerous code path, for a dialog that fires rarely and is getting rarer as conflict handling improves on master.
- Master has meanwhile moved substantially in exactly these files, so the branch is in conflict and a rebase would be a project of its own.

The idea itself isn't dead: a **read-only diff preview** on the conflict dialog (show what differs — across *all* slices — before the existing binary choice, plus a backup export) would deliver most of the safety value with none of the composed-state upload risk. `whole-dataset-diff.util.ts` is a great starting point for that if demand shows up — happy to see it revived in that leaner shape. Thanks again! 🙏

PR #8945, opened 2026-07-11
refactor(sync): migrate conflict-journal unreviewedCount to a Signal

aakhter:
Small follow-up to #8874 (deferred LOW from its review): `unreviewedCount$` was a BehaviorSubject-backed Observable whose only consumer (main-header badge) immediately converted it with `toSignal()`. Expose a readonly signal from `ConflictJournalService` instead and consume it directly.

- No behavior change; specs updated to the signal API (25 green: journal service + main-header).
- `tsc`, `eslint`, `prettier` clean.


johannesjo:
Thank you very much! <3

PR #8946, opened 2026-07-11
fix(sync): keep the conflict summary banner counts live during review

aakhter:
Small follow-up to #8874 (deferred LOW from its review): the "N sync conflicts auto-resolved" banner captured its counts at open time. The sync-icon badge updates live, but an open banner went stale while the user reviewed entries on `/sync-conflicts` — and lingered with a nonzero count after everything was reviewed.

- The banner service now subscribes to the journal's unreviewed count and, on every change, refreshes the open banner's counts (same-id `BannerService.open()` updates in place) or dismisses it at zero.
- Guarded by a new `BannerService.isShown(id)` so a banner the user explicitly dismissed is never resurrected by reviewing activity.
- Specs: refresh-on-review, dismiss-at-zero, and no-resurrection-after-dismiss (banner suite 6/6; conflict + backup suites 297/297 for regressions). `tsc`/`eslint`/`prettier` clean.

### DIFF
--- a/src/app/core/banner/banner.service.ts
+++ b/src/app/core/banner/banner.service.ts
@@ -135,6 +135,11 @@ export class BannerService {
     });
   }
 
+  /** Whether a banner with this id is currently shown (open and not dismissed). */
+  isShown(bannerId: BannerId): boolean {
+    return this._banners().some((banner) => banner.id === bannerId);
+  }
+
   dismiss(bannerId: BannerId): void {
     // Log.log('BannerService -> dismissing Banner', bannerId);
     this._banners.update((banners) => {
--- a/src/app/op-log/sync/conflict-journal.service.ts
+++ b/src/app/op-log/sync/conflict-journal.service.ts
@@ -59,6 +59,15 @@ export class ConflictJournalService {
   /** Number of entries still awaiting review (status === 'unreviewed'). */
   readonly unreviewedCount = this._unreviewedCount.asReadonly();
 
+  private readonly _revision = signal(0);
+  /**
+   * Monotonic counter bumped on EVERY journal mutation (record / status change /
+   * prune / clearAll), regardless of whether `unreviewedCount` changed. Consumers
+   * that must react to composition changes at an equal total (e.g. one remote-win
+   * reviewed while one local-win is recorded) key off this instead of the count.
+   */
+  readonly revision = this._revision.asReadonly();
+
   private _openDb(): Promise<IDBPDatabase<ConflictJournalDB>> {
     return openDB<ConflictJournalDB>(
       CONFLICT_JOURNAL_DB_NAME,
@@ -264,6 +273,12 @@ export class ConflictJournalService {
   private async _refreshUnreviewedCount(
     db: IDBPDatabase<ConflictJournalDB>,
   ): Promise<void> {
+    // Advance `revision` FIRST, so a committed mutation always notifies consumers
+    // even if the count query below throws (callers swallow that error). The
+    // banner keys off `revision` and re-reads the journal itself, so it still
+    // reflects the committed change; `unreviewedCount` catches up once the query
+    // succeeds. Fires on every mutation even when the count is unchanged.
+    this._revision.update((r) => r + 1);
     const count = await db.countFromIndex(
       CONFLICT_JOURNAL_STORE,
       CONFLICT_JOURNAL_INDEX_STATUS,
--- a/src/app/op-log/sync/sync-conflict-banner.service.spec.ts
+++ b/src/app/op-log/sync/sync-conflict-banner.service.spec.ts
@@ -146,4 +146,92 @@ describe('SyncConflictBannerService', () => {
 
     expect(bannerService.open).toHaveBeenCalledTimes(1); // no re-open
   });
+
+  // --- Concurrency guards (SPAP-35 review hardening) ---
+  // The refresh does an async journal read between "is the banner shown?" and
+  // open()/dismiss(). These cover the three ways that gap can misbehave.
+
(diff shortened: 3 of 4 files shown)


johannesjo:
## Review feedback (head `8016fcddd175`)

The implementation is focused and the new dismiss/reorder/phantom-zero tests are useful. I double-checked the event ordering and found two issues in the live-refresh trigger that should be addressed together.

### Important

1. **`src/app/op-log/sync/sync-conflict-banner.service.ts:46` — equal totals can hide a changed local/remote breakdown.**

   `distinctUntilChanged()` deduplicates only `unreviewedCount`, but the banner content is `{ total, remoteWins, localWins }`. `record()` and `_setStatus()` each perform their write and then a separate count query. If the user reviews one remote-win entry while sync records one local-win entry, both count queries can run after both writes and emit `1, 1`. The operator suppresses them because the previous count was already `1`, so the banner can remain at “1 remote” even though the journal now contains “1 local”.

   I reproduced this with fake IndexedDB using the same write-then-count transaction shape: the two mutations emitted `[1, 1]`, while the final unreviewed entry changed from remote-win to local-win.

   Please trigger from a journal mutation/revision rather than treating the numeric total as the content identity (or otherwise allow equal-count mutations), and add a `1 -> 1` composition-change regression test.

2. **`src/app/op-log/sync/sync-conflict-banner.service.ts:47,82` — bulk review starts one full journal scan per entry.**

   Every distinct count emission fire-and-forgets `_refreshOpenBanner()`, and `list("unreviewed")` uses `getAllFromIndex()` over the entire journal before filtering. `keepAll()`/bulk flip update entries one at a time, so an open banner with N reviewed entries launches N full-store reads. `_bannerSeq` prevents stale rendering, but it does not cancel or coalesce any IndexedDB work. With the normal 200-entry retained set, Keep All can therefore issue 200 scans and materialize roughly 40,000 entries; the journal can also exceed 200 until the next startup prune.

   Please coalesce burst mutations into one trailing refresh, use a single-flight/bulk mutation mechanism, or otherwise bound the reads. A regression test should assert that a bulk count burst produces a bounded number of `list()` calls.

A clean combined direction is a mutation-aware trigger that preserves equal-total changes while coalescing bursts before reading the journal.

### Verdict

**Needs changes.** All current CI checks, including the SuperSync and WebDAV gates, are green; these cases are not covered by the existing tests.

aakhter:
Addressed in `3b33acb` (rebased on current master).

**Important #1 — equal-total composition change.** The live refresh is now keyed on a journal **revision** (a new monotonic signal on `ConflictJournalService`, bumped on every mutation), not `distinctUntilChanged` on the unreviewed count. So the 1-remote → 1-local case (total stays 1, breakdown flips) refreshes the "X remote, Y local" banner. Added a regression test that reproduces the equal-count composition change deterministically (a mock journal keeps `unreviewedCount` at 1 while `revision` advances); it fails on the old count-keyed trigger and passes now.

**Important #2 — bulk-review scan storm.** Mutations are coalesced with `auditTime(100ms)`, so a burst (e.g. Keep All over the whole journal) collapses into a single trailing refresh — one journal scan, not one per entry. Added a regression test asserting a 12-entry review burst produces exactly one `list()` call.

**From the follow-up review:**
- The `revision` bump now runs **before** the follow-up count query, so a committed mutation always notifies consumers even if `countFromIndex()` fails (the banner keys off `revision` and re-reads the journal itself; the count signal catches up). This matters most for `clearAll()` on profile transitions.
- Applied the same phantom-zero guard to `maybeShowSummaryBanner()` (the opener), not just the live-refresh path, so a transient `list()` error can't dismiss/skip a valid banner while `unreviewedCount() > 0`.

**Rebase note.** #8945 (merged) migrated `unreviewedCount$` to a signal on master; the banner's old Observable subscription no longer compiled, so this also swaps it to the signal. All banner + journal specs green (33).

johannesjo:
## Follow-up review (head `3b33acb`)

I carefully re-checked the latest implementation, including the full discussion/thread state and the current CI results. The equal-total composition issue is fixed, and the bulk-review scan storm is materially addressed: a production-shaped fake-IndexedDB probe over 200 sequential `get`/`put`/`countFromIndex` mutations completed in 180 ms and triggered 2 refreshes rather than 200. I do not consider the remaining points merge blockers.

### Minor

1. **`src/app/op-log/sync/sync-conflict-banner.service.ts:58-60` — `auditTime(100)` does not guarantee the documented single trailing refresh.** It emits once per active 100 ms window, so a long Keep All loop can still start multiple full-journal scans; if a scan itself takes longer than the window, those reads can overlap because the async refresh is fire-and-forget. The new 12-entry test finishes inside one window and therefore proves the fast-burst case only. If exactly one post-quiescence scan is the contract, `debounceTime` or an explicit single-flight/trailing-dirty mechanism would match it. Otherwise, I would reword the comments and test a justified bounded rate rather than asserting exactly one call.

2. **`src/app/op-log/sync/conflict-journal.service.ts:276-287` / `sync-conflict-banner.service.ts:81,104` — a successful empty read is indistinguishable from a failed read and is checked against a separately refreshed count.** If the final status write commits, `revision` advances, and `countFromIndex()` then fails, `unreviewedCount` remains stale and positive. A subsequent successful `list('unreviewed')` returns `[]`, but the guard ignores it, so the old nonzero banner can remain until another journal mutation. This is non-blocking under the journal's observe-only/best-effort UI contract, but the robust fix would be a success/error-discriminated journal read plus a regression for “recount fails, next list succeeds empty.”

### Verdict

**Ready to merge.** All 20 current checks are green, including the unit/lint gates and all SuperSync/WebDAV gates. The PR is mergeable and has no unresolved inline threads.

johannesjo:
Thank you very much! <3

PR #8947, opened 2026-07-11
fix(sync): bind empty string, not null, for the sync badge description at 0

aakhter:
Small follow-up to #8874 (deferred LOW from its review): the sync-icon badge bound `null` to `matBadgeDescription` (a `string` input) when the unreviewed count is 0.

Verified the hidden-state semantics in the installed Material version before changing anything: `matBadgeHidden` hides the badge via CSS only — MatBadge keeps its aria description active regardless of hidden state. So at count 0 the description genuinely must be absent (otherwise screen readers would announce a stale "0 conflicts" description on every encounter with the sync button); the previous `null` achieved that only by `AriaDescriber.describe()` silently no-op'ing on falsy input. `''` is the type-correct value whose removal path (`_updateDescription`) is explicit. A template comment documents the verified semantics so this doesn't regress to a "stable always-on description" later.

No behavior change (template-literal swap; the semantics live in Material) — so no new spec; dev build compiles, main-header specs green, prettier clean.

### DIFF
--- a/src/app/core-ui/main-header/main-header.component.html
+++ b/src/app/core-ui/main-header/main-header.component.html
@@ -48,6 +48,13 @@
       }
 
       @if (isSyncIconEnabled()) {
+        <!--
+          matBadgeHidden only hides the badge VISUALLY (CSS) — MatBadge keeps an
+          aria description active regardless of hidden state. So at count 0 the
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much! <3

PR #8948, opened 2026-07-11
fix(sync): enforce conflict-journal retention mid-session, not only at start

aakhter:
Small follow-up to #8874 (deferred LOW from its review): journal retention (`pruneOnStart`, 14 days / newest 200) ran only at app start, so a long-lived session could grow `SUP_CONFLICT_JOURNAL` without bound.

- `record()` now checks the store count after each put — `count()` is cheap — and when it exceeds `JOURNAL_MAX_ENTRIES + JOURNAL_PRUNE_SLACK` (200 + 20) runs the same age+count prune, bringing the store back to 200. The O(n) prune is amortized to once every 20 records.
- The prune core is extracted into a private `_prune()` shared with `pruneOnStart` (no behavior change there — its specs are untouched and green).
- `record()`'s observe-only never-throw contract is unchanged (the prune runs inside its existing try).
- Specs: no prune at the soft cap; crossing record prunes back to `JOURNAL_MAX_ENTRIES` (oldest dropped, newest kept). Journal suite 24/24; conflict + backup regression suites 289/289. `tsc`/`eslint`/`prettier` clean.

### DIFF
--- a/docs/sync-and-op-log/conflict-journal-and-review.md
+++ b/docs/sync-and-op-log/conflict-journal-and-review.md
@@ -46,9 +46,14 @@ Contracts:
   `BackupService.importCompleteBackup` — the chokepoint every replacement path
   funnels through (profile switch, JSON import, local-backup restore, SuperSync
   restore) — calls `ConflictJournalService.clearAll()`.
-- **Retention.** Pruned on app start to whichever bound binds first: entries
-  older than 14 days (`JOURNAL_RETENTION_DAYS`) or beyond the newest 200
-  (`JOURNAL_MAX_ENTRIES`).
+- **Retention.** Each prune applies whichever bound binds first: entries older
+  than 14 days (`JOURNAL_RETENTION_DAYS`), then anything beyond the newest 200
+  (`JOURNAL_MAX_ENTRIES`). Pruning runs on app start, and opportunistically
+  mid-session from `record()` — but the mid-session prune is **count-triggered**:
+  it fires only once the store grows past the soft cap `JOURNAL_MAX_ENTRIES +
+JOURNAL_PRUNE_SLACK` (220), then prunes back to the newest 200. So a
+  long-running low-volume session (few entries, never crossing the soft cap)
+  relies on the next app start to enforce the 14-day age bound.
 
 ### Classification taxonomy
 
--- a/docs/wiki/3.06-User-Data.md
+++ b/docs/wiki/3.06-User-Data.md
@@ -80,7 +80,7 @@ The Windows Store build uses a different path that includes the Windows Store pa
 - **SUP_OPS** (current, version 8): Main database. Object stores: `ops`, `state_cache`, `import_backup`, `vector_clock`, `archive_young`, `archive_old`. Holds operations log, state snapshots, vector clocks, and archives.
 - **pf:** Legacy database used for migration and recovery.
 - **SUPPluginCache:** Plugin cache.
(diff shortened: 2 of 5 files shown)


johannesjo:
## Review — mid-session conflict-journal retention

The threshold behavior is sound: the specs pin both the 220-entry soft cap and the 221st-entry prune back to the newest 200, and the extracted helper preserves the existing age-first/count-second policy. Head `f46b7c3` is stable, with CI, CodeQL, SuperSync, and WebDAV checks green.

### Important — the new mid-sync prune path can emit an unhandled rejection

`src/app/op-log/sync/conflict-journal.service.ts:258-261`

The delete sequence awaits the request promises before `tx.done`:

```ts
await Promise.all(Array.from(idsToDelete, (id) => tx.store.delete(id)));
await tx.done;
```

If the read-write transaction aborts while those requests are pending, both the delete aggregate and `tx.done` reject. The first rejection exits `_prune()` into the catch in `record()`, so execution never handles the separate `tx.done` rejection. I reproduced this against the installed `idb@8.0.3` and `fake-indexeddb`: the journal call catches `AbortError`, followed by a global unhandled `AbortError`.

The sequencing already existed in startup pruning, but this PR materially exposes it during active conflict resolution whenever the journal crosses the soft cap. The awaited sync call still resolves, so this is not demonstrated data corruption or direct sync rejection; it nevertheless breaks the explicit observe-only/no-unhandled-error contract.

Smallest fix: attach handlers to the delete requests and transaction completion together, for example `await Promise.all([...deletePromises, tx.done])`, and add an aborted-transaction regression test verifying that no rejection escapes globally.

### Minor / non-blocking

- `src/app/op-log/sync/conflict-journal.service.ts:235` — `getAllFromIndex()` materializes every retained full journal value during sync. Count is bounded to 221, but bytes are not: field diffs preserve notes, arrays, and opaque action payloads verbatim, and pruning can recur every 21 conflicts while conflict resolution holds the operation-log lock. Key-only index reads or cursors would avoid cloning values needed only for IDs and timestamps.
- `src/app/op-log/sync/conflict-journal.model.ts:117-131` — retention comments and docs still describe startup-only pruning and "at most 200." Mid-session size can reach 220, entry 221 triggers pruning, and after returning to 200 the next prune is 21 unique inserts later, not `JOURNAL_PRUNE_SLACK` or 20. Please update the model comments, `docs/sync-and-op-log/conflict-journal-and-review.md`, and the relevant user-data wiki text.

## Verdict

**Needs changes** for the unhandled transaction rejection. I found no other merge-blocking correctness issue.

aakhter:
Addressed in `be00da0c` (rebased on current master).

**Important (blocker) — unhandled rejection on abort.** The prune now awaits the delete requests and `tx.done` in one `Promise.all([...deletes, tx.done])`, so `tx.done` always has a handler and its rejection can't escape when the delete aggregate rejects first. Added an aborted-transaction regression test: it drives the prune's read-write transaction to reject both the deletes and `tx.done`, asserts the prune still resolves (observe-only), and asserts `tx.done` is awaited — it fails on the old `Promise.all(deletes); await tx.done` sequencing and passes now. (Zone.js swallows the native `unhandledrejection` event under Karma, so the test pins the mechanism — that `tx.done` is awaited alongside the deletes — rather than listening for the global event.)

**Docs.** Updated the retention model comment, `conflict-journal-and-review.md`, and the canonical `3.06-User-Data.md` to describe mid-session pruning accurately — and specifically that it is **count-triggered** (fires once the store crosses the 220 soft cap), so a long-running low-volume session still relies on the next app start to enforce the 14-day age bound.

**Minor (deferred).** The `getAllFromIndex()` full-value materialization during prune — agreed it could be key-only cursor reads; leaving that as a follow-up to keep this change focused on the correctness fix.

johannesjo:
Thank you very much! <3

PR #8949, opened 2026-07-11
fix(sync): suppress the "Project updated" snack on conflict-review flips

aakhter:
Follow-up from the independent review passes of #8874: flipping a PROJECT conflict entry dispatches `updateProject`, whose `snackUpdateBaseSettings$` effect popped an unconditional "Project updated" snack on top of the flip flow's own outcome reporting.

- `updateProject` gains an optional `isSkipSnack` flag (mirroring `updateTask`'s `isIgnoreShortSyntax` precedent for flow-specific flags riding on the action); the snack effect filters on it; the flip path sets it.
- Specs: snack still shown for a normal update, suppressed with the flag, and the PROJECT flip dispatches with the flag set.
- Also adds a `MockStore.resetSelectors()` `afterEach` to the conflict-ui spec — `overrideSelector` mutates the shared selector references, and the new `selectProjectById` override otherwise leaked into `project.service.spec` when bundled in the same karma run (the pre-existing `selectTaskById` override had the same latent issue).

Of the three below-fix-bar items noted in that review: item 2 (journal `clearAll()` outside the op-log lock window) is already fixed on master (clearAll now runs inside the lock); item 3 (journaled post-resolution baseline for local-won stale detection) is the maintainer's documented follow-up and is deliberately not attempted here.

Verification: project + conflict suites 346/346, `tsc`/`eslint`/`prettier` clean.

### DIFF
--- a/src/app/features/project/store/project.actions.ts
+++ b/src/app/features/project/store/project.actions.ts
@@ -31,7 +31,9 @@ export const addProjects = createAction(
 
 export const updateProject = createAction(
   '[Project] Update Project',
-  (projectProps: { project: Update<Project> }) => ({
+  // isSkipSnack: suppress the unconditional "Project updated" snack — used by
+  // flows that surface their own outcome (e.g. the conflict-review flip).
+  (projectProps: { project: Update<Project>; isSkipSnack?: boolean }) => ({
     ...projectProps,
     meta: {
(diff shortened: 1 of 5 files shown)


johannesjo:
Thank you very much !<3

Issue #6373, opened 2026-02-05
💡 iOS home screen widget


fidelnamisi:
### Problem Statement

 I have Superproductivity on my mac and iphone. I love it, especially how hackable it is. I've added some features on the mac version, but find myself wishing the iOS version could also have a homescreen widget.

### Possible Solution

I see there's been a lot of progress with the Android homescreen widget. Would be great to also have an iOS widget that shows:

1. Tasks due today (and completed today)
2. Start / Stop Time Tracking

fidelnamisi:
Is it possible to get the source code for the iOS app? If so, I could contribute towards getting this done.

johannesjo:
That is great to hear! Yes, the source code is available under https://github.com/super-productivity/super-productivity/tree/master/ios. We are using capacitor.js.

fidelnamisi:
Hey @johannesjo 

I finally managed to get the iOS widgets fully functioning!

I just pushed all the working iOS widget files and the frontend data synchronization payload changes directly to GitHub on master.

Feel free to pull the latest changes so you can check them out or test them locally. Let me know if you run into any issues running it on your end! Disclaimer: I'm a vibe coder. I just fought long and hard with various models working with Antigravity and Xcode until finally the Widget worked. It was Claude Opus 4.5 working directly in Xcode that sorted it out eventually. Sorry for not being a real dev!

fidelnamisi:
Sorry. Note that the only widget I needed was "Today's Tasks". So that's all I prioritized at this stage. Thanks.

johannesjo:
> Hey [@johannesjo](https://github.com/johannesjo)
> 
> I finally managed to get the iOS widgets fully functioning!
> 
> I just pushed all the working iOS widget files and the frontend data synchronization payload changes directly to GitHub on master.

Hey hey! Can you create a PR for this? :)

fidelnamisi:
Done!

jonasbrondum:
Hi, has this been pushed to iOS yet? I have the latest version of the iOS app (App version 18.13.1), but cannot find the widgets when I try to update my homescreen.

System: 
iOS version: 26.5
iPhone 13 Mini

PR #8950, opened 2026-07-12
feat(widget): port the Android home screen widget to iOS via WidgetKit

johannesjo:
## Problem

iOS has no home screen widget while Android shipped one in #8737, leaving the most-requested mobile feature (#3818, and its iPhone counterpart #6373) Android-only. The port was already planned in [`docs/plans/2026-07-07-ios-home-screen-widget-port.md`](``https://github.com/super-productivity/super-productivity/blob/claude/mobile-platform-improvements-jhp6x2/docs/plans/2026-07-07-ios-home-screen-widget-port.md).``

Implements #6373.

## Solution

View-layer + plumbing port of the Android widget, reusing the `v: 1` `widget_data` JSON contract unchanged. Angular stays the single writer of the snapshot blob; widget done-taps flow back through a last-wins `{taskId: targetIsDone}` queue that the widget overlays at render time, so the write race stays structurally impossible.

**Angular (mostly generalizing):**

- `features/android/android-widget.*` → platform-neutral `features/widget/` (`git mv`, types renamed `WidgetData`/`WidgetTask`, selector `selectWidgetData`).
- `WidgetDataService` branches its sink: Android → `androidInterface` KeyValStore + `updateWidget()`; iOS → new `WidgetBridge` Capacitor plugin (App Group `UserDefaults` + `WidgetCenter.reloadTimelines`).
- Effects gate widens from `IS_ANDROID_WEB_VIEW` to `IS_WIDGET_PLATFORM` (Android WebView or iOS native). iOS-specific trigger differences: an initial drain emission (Capacitor `appStateChange` never fires at cold start) and a new `iosInterface.onPause$` for the pre-suspend snapshot push. Android behavior is byte-identical.
- Sync-correctness: unchanged risk profile — all effects stay `dispatch: false` state consumers with the existing hydration guard; the drain path produces user-intent ops exactly like the Android drain (dedup + skip-already-in-target prevents replay noise).

**iOS native (all new):**

- `SupWidget` WidgetKit extension target (iOS 17+ for interactive `Button(intent:)`; the app itself stays at 16.0 — on 16.x the widget is simply unavailable): `WidgetData.swift` parser porting the Kotlin edge cases (version gate → empty, JSON-null `projectId` guard), `DoneQueue.swift` mirroring `WidgetDoneQueue.kt`, `ToggleDoneIntent`, and a SwiftUI medium/large view matching the Android v1 look (light theme, project dots, `+N more` overflow since iOS widgets can't scroll).
- `WidgetBridgePlugin` (Swift + ObjC stub, `StoreReviewPlugin` pattern) in the app target; App Groups entitlements (`group.com.super-productivity.app`) on both targets; extension target hand-wired into `project.pbxproj` (validated with an OpenStep-plist parser plus the npm `xcode` parser).
- Parser lock tests (`SupWidgetTests/WidgetDataTests.swift`) against the same golden JSON as `WidgetDataTest.kt` — not yet wired to an Xcode test target (the project has no iOS test infra); instructions in `ios/App/SupWidget/README.md`.

**CI:** `build-ios.yml` installs + verifies a second provisioning profile (`IOS_WIDGET_PROVISION_PROFILE` secret) and registers it in export options, failing early with a pointer to the setup doc if the secret is missing.

**⚠️ Manual follow-ups before the next App Store build** (listed in `ios/App/SupWidget/README.md`): register the widget App ID, create/enable the App Group on both App IDs, regenerate the app profile + create the widget profile, set the new CI secret, and verify on a real device (interactive widgets are flaky in the simulator). Until then the iOS release workflow will fail at export — by design, with a clear error.

Known v1 limitations (deliberate, matching or below Android v1): taps while the app is foregrounded apply on next resume (no LocalBroadcast equivalent; the pending overlay keeps the widget itself correct), stale-until-next-open on day rollover, English-only chrome.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki). — plan doc updated + `SupWidget/README.md`; no wiki note exists for widgets (Android widget shipped without one), so parity kept
- [x] I have added tests for my changes (if applicable) — moved TS specs updated; Swift parser tests added (see note above)
- [x] Existing tests still pass — 12,386 pass; the 2 failures in `DialogSelectDateTimeComponent` reproduce identically on current `master` (pre-existing, time-of-day dependent, unrelated to this diff)

https://claude.ai/code/session_01RGoSd3gfkg5bDDL1cBSXTT

---
_Generated by [Claude Code](https://claude.ai/code/session_01RGoSd3gfkg5bDDL1cBSXTT)_


johannesjo:
## Follow-up multi-review — `6fd146ff83`

Re-reviewed the current head across Angular/op-log sync behavior, Android and iOS queue durability, native WidgetKit/Xcode integration, failure recovery, privacy/security, maintainability, and CI. **No remaining actionable code findings.**

Changes applied during review:

- Made iOS completion intents durable with atomic App Group storage, cross-process locking, revisioned leases, and ABA-safe acknowledgement.
- Delayed iOS acknowledgement until the intent op is durably captured and the native snapshot is refreshed; added sync-window and swallowed-write-failure guards.
- Restored the Android destructive queue path to a guarded same-turn clear-and-dispatch flow, including the required bulk-dispatch yield.
- Added logical-day expiry and an explicit empty timeline boundary, plus count, color, and accessibility fixes.
- Added split privacy manifests, a real `SupWidgetTests` target/scheme, and path-scoped iOS build/test CI.
- Removed the final explicit `any` from the selector fixture by typing it as `ProjectState`.

Verification on the final head:

- Focused widget tests: 27/27; final selector test: 5/5.
- Required `checkFile` checks passed for every modified TypeScript file; formatting and diff checks are clean.
- All GitHub checks pass, including the iOS app/widget build and native tests, general tests/E2E, WebDAV, all six SuperSync shards and both gates, lint, CodeQL, dependency review, Lighthouse, Electron packaging, and preview build.
- No review submissions or unresolved review threads exist.

Known release constraints, not code-review blockers:

- Apple App Group/App ID, provisioning profiles, the signing secret, and real-device verification still need release-owner setup.
- A time-zone change while the app is suspended can retain the previous-zone Today snapshot until its stored boundary; this limitation is documented.
- The PR description still contains older notes about unwired Swift tests, unchanged Android behavior, rollover behavior, and the scope of #6373. This comment supersedes those verification notes; the implementation covers the Today-task widget scope, not time tracking.

**Verdict:** code and CI are ready. GitHub remains blocked only on the repository-required human review.

Zoriot:
Anything i can do to make that happening faster?

johannesjo:

Thanks for the PR! This pass covers what changed since the 2026-07-13 review: the rebase onto master at `285f619b` and the fixture-alignment commits (`e7d8f574`, `91c3def9`, `321aa3e1`). The Angular drain/lease path and the Swift queue held up under a second read; the one defect is in the rewritten Swift fixture, which is also why the `iOS PR` check is red.

### Should fix before merge
- **`ios/App/SupWidgetTests/WidgetDataTests.swift:38`** – a single-pound raw string literal `#"..."#` ends at the first `"#` sequence, and this JSON contains `"null":"#ff0000"`, so the literal closes right before `ff0000` and the rest of the line is parsed as stray statements. That is the `consecutive statements on a line must be separated by ';'` / `expected a macro identifier for a pound literal expression` error at 38:113–124 in the `Build app and test widget` job: the app and widget build succeed, only `SupWidgetTests` fails to compile, so none of the parser lock tests run. Use a two-pound delimiter (`##"..."##`) or a plain string with escaped quotes; the other raw literals in the file contain no `"#` and are fine.

### Nits
- **`.github/workflows/ios-pr.yml:8`** – the `paths` filter (`ios/**`, `src/app/features/widget/**`, the workflow itself) misses `src/assets/i18n/en.json` and `tools/generate-ios-widget-localizations*.js`, although `sync:ios` now regenerates `en.lproj/Localizable.strings` from the `WIDGET.IOS` section and `renderStringsFile` throws if that section is missing, empty, or holds a non-string. The node unit test only exercises fixture data, so a change that breaks real generation surfaces at the next release `build-ios.yml` run instead. Adding those two paths closes the gap.
- **`src/assets/i18n/en.json:3218`** – `WIDGET.IOS` has no reference anywhere under `src/` (only the generator reads it), so `npm run int:clean` (`ngx-translate-extract ... --clean --marker _` in `package.json`) will strip the whole section as unused, after which the next `npm run sync:ios` fails in `generateIosWidgetLocalizations`. A `_('WIDGET.IOS.TODAY')`-style marker reference, or excluding the section from the clean run, keeps a routine i18n cleanup from breaking the iOS build.
- **`ios/App/SupWidget/WidgetData.swift:15`** – `isValid(at:)` is `validUntil.map { date < $0 } ?? false`, so a blob without `validUntil` renders as expired (empty list plus the refresh prompt), while the Kotlin reader of the same `v: 1` contract deliberately goes the other way (`isSnapshotStale` is `meta.validUntil != null && nowMs >= meta.validUntil`, commented "an unknown day must not be reported as expired"). Latent today because `selectWidgetData` always emits the stamp, but two readers of one contract should agree; either match the Kotlin rule or note the divergence in the model comment.
- **`docs/plans/2026-08-06-pr-8950-finalization-checklist.md:1`** – pins `91c3def9` as the pushed head (already superseded by `321aa3e1`) and carries 20 unchecked boxes, so it is stale the moment the PR merges; `docs/plans/` on master holds design plans only, and this branch already added and then retired a similar merge plan in `6fc3f58c`. Keeping it in the PR description, or trimming it to the reusable "common failure meanings" table, seems better than merging it.
- **`src/app/op-log/capture/operation-log.effects.spec.ts:751`** – the two op-log spec additions assert behaviour this PR does not change (`operation-log.effects.ts` is untouched since the merge-base and already calls `markUnrecoveredPersistFailure()` on that path). Harmless, just out of scope for the widget port.

---

johannesjo:
**💬 Decision note (self, not a review item)** – drop the `en.json`-driven iOS localization generator; keep the widget's native strings hand-maintained in `ios/App/SupWidget/en.lproj/Localizable.strings`, exactly as the Android widget keeps `android/app/src/main/res/values/strings.xml`. Orthogonal to the open blocker from this morning's review (the `WidgetDataTests.swift` raw-string); nothing below is implemented yet.

### Why

- **The pruning risk already fired on this exact shape.** `12d315a703` ("chore(i18n): add unused translation scanner and remove orphaned keys") deleted the entire top-level `ANDROID` section from every locale file for no reason other than "no `T.*` reference", and `tools/cleanup-unused-translations.js:16` still hardcodes `const SECTIONS_TO_REMOVE = ['ANDROID', 'THEMES'];`. `WIDGET.IOS` is the same shape: a new top-level i18n section that nothing under `src/` references.
- **The detector will flag it.** `tools/find-unused-translations.js` globs `**/*.{ts,html}` under `src/` only and ends with `process.exit(unused.length > 0 ? 1 : 0)` (`:94`). The keys are consumed from Swift (`localized("WIDGET.IOS.TODAY")` etc.), so all 14 come back unused and `npm run int:unused` exits non-zero from the merge onward. It is a manual script rather than a CI gate (`ci.yml` runs `int:test` only) — which is exactly why the next cleanup pass works off its output.
- **The failure would be silent, then late.** `ios-pr.yml` triggers only on `ios/**`, `src/app/features/widget/**` and its own path. A future i18n-cleanup PR touching only `src/assets/i18n/` merges green; the generator then throws `WIDGET.IOS must contain at least one string-only translation entry` at `npm run sync:ios` — which is `build-ios.yml:188`, the release build.
- **These are not translations in the project's sense.** The rule targets strings that go through `T` / `TranslateService`. These are compiled native resources, and the sanctioned precedent is `strings.xml`, which already carries the same Android widget strings by hand (`widget_title` = "Today's Tasks", `widget_header_title` = "Today", `widget_empty` = "No tasks for today", `widget_description` = "Shows today's tasks with quick done action").
- **No upside while only `en.lproj` is written.** The generator emits one locale, so routing through `en.json` buys zero translated output while exposing `%@`-placeholder strings ("Mark %@ as done", "+%@ more") to translators for a surface that no translated build reaches.
- **`src/app/t.const.ts` is not regenerated in this PR.** `en.json` gains the `WIDGET.IOS` block but `t.const.ts` is untouched, so the two ship out of sync and the next `npm run int` writes 14 unused `T.WIDGET.IOS.*` entries into whatever unrelated PR happens to run it.

### Follow-up shape

1. Delete `tools/generate-ios-widget-localizations.js` and `tools/generate-ios-widget-localizations.test.js`.
2. Drop `generate:ios-widget-localizations` and `generate:ios-widget-localizations:test` from `package.json`, remove the latter from the `test` script, and revert `sync:ios` to plain `npx cap sync ios`.
3. Remove the `WIDGET.IOS` block from `src/assets/i18n/en.json`.
4. Keep the committed `ios/App/SupWidget/en.lproj/Localizable.strings` as-is minus the `/* Generated from ... Do not edit directly. */` header. Keys are unchanged, so no Swift edits are needed.
5. Reword `ios/App/SupWidget/README.md:41-43` (and `docs/plans/2026-07-07-ios-home-screen-widget-port.md:152-153`) to say the file is edited directly, like Android's `strings.xml`.

If widget localization is ever actually wanted, the shape is: generate all locales rather than `en` alone, plus an extract-ignore entry so the cleanup tooling stops reading the section as orphaned. Not today's problem — v1 ships English-only widget chrome by design.

Zoriot:
I just tried the TestFlight build.
But it seems to now really work, the app itself still displays the old version. 

I can add a Today Widget but this does not display any data. It always says "Open Super Productivity to refresh" when I click on it SP opens but still no data is displayed.

I also wonder if it's save for me to add the TestFlight and try out other iOS specific features or if you want to do that exclusively (I'm currently thinking mainly about the share feature)

Issue #1834, opened 2022-01-14
Popup "You have been idle for.." is showing while in a break


nkrot:
When I am taking a longer break, like 20 minutes after a number of Pomodoro sessions, the application is still throwing a pop up "You have been idle for ....". This is annoying and contradicting the idea of a longer break, isn't it?

### Your Environment

- Version used: 7.9.1 (deb)
- Operating System and version: Ubuntu 18.04
- Desktop Environment: MATE
- Browser Name and version: Firefox 89.0

### Expected Behavior

When I am in a longer break, Idle Time popup should not be shown. It may be shown once longer break has elapsed AND idle time length has also elapsed. For example, if my longer break period is set 20 minutes and Idle time is set to 5 minutes, then the popup may appear once 25 minutes elapsed.

johannesjo:
Hey there! Thanks for opening this up. I think this is not a 100% clear since the idle time might have been used for something else than a break, e.g. a meeting or talking to a colleague or something, so I think it makes most sense to leave the decision what to do with the idle time to the user.

But I can understand why you would prefer it to be different. PRs for implementing this as opt in optional feature are welcome!

nkrot:
@johannesjo , i think i did not make myself clear. The application *knows* that I am taking a break, it is showing me a countdown timer for the break. I am referring to the case that the Idle pop-up does not make sense while break timer is running.

If the user is on a meeting, he or she should have skipped the "take a break" reminder.

johannesjo:
> If the user is on a meeting, he or she should have canceled the "take a break" reminder.

He maybe should, but it doesn't mean he always does. Think of the following case: User is in work session and a colleague asks him something. The (work related) discussion takes much longer than the pomodoro session and the break already started. What you'd expect? Or you're outside for your break and a colleague asks you something. 

These are cases (and there are probably more) where I personally would prefer to have an Idle Dialog shown.

nkrot:
the scenario that you describe does not sound convincing to me when it comes to showing idle dialogue *during* a break. the dialogue can be show *after* the break ended. logically, users are supposed to be idle while taking a break.

> The (work related) discussion takes much longer than the pomodoro session and the break already started. What you'd expect?

[...]

PR #8965, opened 2026-07-12
feat: suppress idle dialog during active work sessions

Erick0412-dev:
﻿## Description

When studying or working at the computer without constant keyboard/mouse input (e.g., reading, writing exams, reviewing code), the idle time detection dialog can interrupt a focus work session. This is especially disruptive during Pomodoro or countdown sessions where the user is intentionally engaged with the screen but not actively typing.

This PR adds an opt-in setting **"Suppress idle dialog during work sessions"** under Settings → Idle Handling. When enabled, the idle dialog will not appear while a focus mode work session is running. The setting is off by default (opt-in) to avoid silently logging away time — users must explicitly enable it if their workflow involves extended reading or study time at the computer.

## Changes

- **`global-config.model.ts`** — Added `isSuppressIdleDuringFocusMode` to `IdleConfig`
- **`default-global-config.const.ts`** — Default: `false` (opt-in)
- **`idle-form.const.ts`** — Checkbox form field
- **`idle.effects.ts`** — Guard: returns EMPTY when enabled + work session running
- **`global-config.reducer.ts`** — Deep-merges idle section on loadAllData for existing users
- **`t.const.ts`** — Translation key
- **`en.json`** — English translation (other languages reverted per repo convention)
- **`idle.effects.spec.ts`** — Tests for suppress/allow behaviors
- **`global-config.reducer.spec.ts`** — Test for idle section default merging
- **`3.02-Settings-and-Preferences.md`** — Documentation

## Note

This is not related to issue #1834 (which was about suppressing idle during Pomodoro *breaks*). The use case here is different: suppressing idle during active *work* sessions to accommodate workflows with extended passive screen time.


johannesjo:
Thanks for working on this. The goal of avoiding an idle popup during an intentional focus break makes sense, but I don't think the current implementation is mergeable yet.

The main blockers I found on `f3d0a640` are:

1. **The guard checks the opposite focus phase from the reported bug.** The new condition in [`idle.effects.ts`](https://github.com/super-productivity/super-productivity/blob/f3d0a640f36302120d3386bdf10559d9b67aa137/src/app/features/idle/store/idle.effects.ts#L127-L138) uses `selectIsSessionRunning`. That selector is explicitly true only for a running **work** timer and false for a running **break** timer ([selector](https://github.com/super-productivity/super-productivity/blob/f3d0a640f36302120d3386bdf10559d9b67aa137/src/app/features/focus-mode/store/focus-mode.selectors.ts#L39-L42), [tests](https://github.com/super-productivity/super-productivity/blob/f3d0a640f36302120d3386bdf10559d9b67aa137/src/app/features/focus-mode/store/focus-mode.selectors.spec.ts#L118-L132)). As a result, the popup from #1834 is still allowed during breaks, while idle handling is suppressed during focused work. With the setting enabled by default, that can silently leave away-time recorded as work because the normal idle cleanup never runs.

   Please base this on break-specific state and add effect tests proving at least: work sessions still trigger normal idle handling; active breaks are suppressed only when enabled; disabling the setting preserves existing behavior; and the post-break threshold follows #1834 (the break duration itself should not count toward `minIdleTime`).

2. **The claimed default does not apply to existing persisted configs.** Adding `true` to `DEFAULT_GLOBAL_CONFIG.idle` is insufficient because `loadAllData` merges defaults only at the top level; an existing persisted `idle` object replaces the whole default section. `selectIdleConfig` also falls back only when the entire section is missing. Therefore older configs load `isSuppressIdleDuringFocusMode` as `undefined`, which behaves as false and appears unchecked. If the setting remains default-on, please merge idle defaults during hydration and add a legacy-config reducer test. If it is changed to opt-in, the stored-data behavior should still be made explicit and tested.

3. **The scope/default needs to match the accepted issue direction.** #1834 is specifically about Pomodoro breaks, and the discussion accepted this as an optional opt-in behavior. This PR broadens it to all focus work modes, enables it by default, and labels it accordingly. I suggest an opt-in, break-specific setting unless there is a new product decision and rationale for changing that behavior.

4. **Repository requirements still need cleanup:**

   - Please edit only `en.json`; revert the other locale changes and the unrelated deletions in `hr.json`/`uk.json`. English is the fallback and translations are handled separately.
   - Add the new setting to `docs/wiki/3.02-Settings-and-Preferences.md` as required by the documentation guide.
   - No idle-effect tests were added, and the only GitHub check currently present is the `welcome` job.
   - The exact PR version of `idle.effects.ts` currently fails Prettier/ESLint at lines 127–129. Please run `npm run checkFile` for every modified TypeScript file.

Once the break/work predicate, timing semantics, legacy-config behavior, tests, and repository-policy issues are addressed, the underlying feature can be reviewed again.

Erick0412-dev:
﻿Thanks for the thorough review! I have addressed all four points:

**1. Scope/selector direction**

I have retitled and redescribed the PR — this is now a separate feature from #1834. The goal is to suppress the idle dialog during active **work** sessions (not breaks) for users whose workflow involves extended passive screen time (reading, studying, reviewing code). The selector `selectIsSessionRunning` (work sessions only) is correct for this use case.

**2. Default value not applied to existing users**

Added deep-merge for the `idle` section in the `loadAllData` reducer handler, matching the pattern used for `tasks`, `focusMode`, `shortSyntax`, etc. The new field now falls back to the default value for users with persisted idle configs. Added a reducer test for this.

**3. Default direction**

Changed `isSuppressIdleDuringFocusMode` from `true` to `false` (opt-in). Users must explicitly enable it, avoiding the risk of silently logging away time.

**4. Repo standards**

- **i18n**: Reverted all 27 non-English translation files; only `en.json` retains the key.
- **ESLint/Prettier**: Fixed indentation in `idle.effects.ts`.
- **Tests**: Added `idle.effects.spec.ts` with 3 test cases (suppressed when enabled+session running, fires when no session, fires when disabled) and a reducer test for idle section merging.
- **Docs**: Added all idle settings (including the new one) to `3.02-Settings-and-Preferences.md`.

The branch has been force-pushed with the updates. Let me know if anything else needs adjustment!

johannesjo:
Thanks, I carefully rechecked the current head (`ee33a975`). The revised scope, opt-in default, legacy-config merge, i18n cleanup, and documentation address the earlier feedback, and I do not see another blocker in the production implementation.

One merge blocker remains in the newly added idle-effect test suite:

- `npm run test:file -- src/app/features/idle/store/idle.effects.spec.ts` fails during TypeScript compilation at [`idle.effects.spec.ts:40`](https://github.com/super-productivity/super-productivity/blob/ee33a9759be8944850535e43c786e23a825c0aeb/src/app/features/idle/store/idle.effects.spec.ts#L40): the mock property is declared as `Subject<void>` but assigned a value cast to `Observable<void>` (`TS2740`).
- Removing that cast exposes a second failure: [`IdleEffects` has no `LOCAL_ACTIONS` test provider](https://github.com/super-productivity/super-productivity/blob/ee33a9759be8944850535e43c786e23a825c0aeb/src/app/features/idle/store/idle.effects.spec.ts#L58-L93), so all three tests fail with `No provider found for ScannedActionsSubject`.
- Providing `LOCAL_ACTIONS` exposes the readiness ordering problem at [lines 99-101](https://github.com/super-productivity/super-productivity/blob/ee33a9759be8944850535e43c786e23a825c0aeb/src/app/features/idle/store/idle.effects.spec.ts#L99-L101): the plain `Subject` is emitted and completed before `triggerIdleWhenEnabled$` is subscribed, so the effect's `first()` receives an already-completed empty stream and the tests fail without exercising the idle behavior.
- The exact submitted spec also fails Prettier.

The smallest fix is to provide `LOCAL_ACTIONS` (for example with `EMPTY`), make the `onReady$` mock type consistent, and either use `ReplaySubject<void>(1)` or emit readiness after subscribing. Then run `npm run checkFile src/app/features/idle/store/idle.effects.spec.ts` and commit the formatting output.

I verified those isolated harness corrections: the idle-effect tests pass 3/3, and the config reducer suite passes 73/73. GitHub currently reports no CI checks on the branch. Once this test harness is fixed, I do not see another merge blocker.

Erick0412-dev:
﻿The test harness issues have been fixed:

- **TS2740 type mismatch**: Removed the incorrect `as unknown as Observable<void>` cast on `onReady$` — the mock property already declares the correct `Subject<void>` type.
- **LOCAL_ACTIONS provider**: Added `{ provide: LOCAL_ACTIONS, useValue: actions$ }` alongside `provideMockActions`.
- **Timing**: Changed `Subject<void>` to `ReplaySubject<void>(1)` so the `onReady$` signal survives late subscription — the effect now correctly activates even when `next()`/`complete()` are called before the observable chain subscribes.

All three tests pass locally with these changes. The branch has been pushed (`f9b78ef`).

johannesjo:
Thank you very much! <3

Erick0412-dev:
Thanks for your thorough review and guidance! It was a great first PR experience. Looking forward to contributing more. 🙌

PR #8966, opened 2026-07-13
fix(settings): Prevent tab shift in settings tabs when page is scrollable

lanesawyer:
## Problem

While getting to know the app better, I noticed that when I switched to a tab that was tall enough to activate the scrollbar, the entire settings page tab navigation would shift:

https://github.com/user-attachments/assets/7a409478-79e2-473c-bef0-cd4215056f73

## Solution

This PR modifies the CSS so that:
1. The tabs no longer are part of the scrollable area, they will never move when the scrollbar comes in or out
2. Makes the tab flush with the side of the page instead of having a small bit of padding. This puts it in line with other tab pages, like Boards, which don't have the gap
3. Makes the tabs sticky, that seems genuinely useful to always have access to the navigation when looking at settings

https://github.com/user-attachments/assets/bdf1b0df-403a-4119-9d20-098fb17bd5c8

This works in mobile view too. Everything should be practically identical, just without the shift on tall pages!

I felt like the footer shouldn't always be visible, which happened when I first changed the scrollable area, so there's a 50-odd line copy/paste and render of a template refactor included so that the footer is at the bottom of each settings page and not always visible.

## Screenshots

Mobile view of scrolling page
image

Mobile view of short page:
image

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/pages/config-page/config-page.component.html
+++ b/src/app/pages/config-page/config-page.component.html
@@ -1,4 +1,4 @@
-<div class="page-settings page-wrapper">
+<div class="page-settings">
   @if (globalCfg) {
     <div class="settings-container">
       <!-- Shared by every tab's `mat-tab-label`: an icon + translated title. -->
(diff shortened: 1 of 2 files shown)


src/app/pages/config-page/config-page.component.html
```diff
@@ -15,6 +15,58 @@
         <span class="tab-label">{{ labelKey | translate }}</span>
       </ng-template>
 
+      <!-- Rendered at the bottom of every tab so it scrolls away with the tab's
+           content (the tab body, not the page, is the scroll container). -->
+      <ng-template #versionFooter>
```
lanesawyer, line 20: I have manually verified that Claude properly moved contents of this template and it's identical to the existing footer!

johannesjo:
Thanks for tackling this — I tested the preview at desktop and mobile sizes. The intended behavior currently works: the Material tab body owns scrolling, the tab header stays fixed, the outer route remains at `scrollTop = 0`, and focusing a lower settings control did not reproduce the old scroll jump.

Before merging, I think two things need attention:

1. **Important: add regression coverage for the new scroll hierarchy.** The change at [`config-page.component.scss:3`](https://github.com/super-productivity/super-productivity/blob/a991ec480ed6d8a76f01319400a314a6bec2c040/src/app/pages/config-page/config-page.component.scss#L3-L18) removes the explicit Material overflow workaround added by `a4c7ec892` for focus-induced scroll resets and replaces the page's established scroll model. The component spec [replaces the template with an empty string](https://github.com/super-productivity/super-productivity/blob/a991ec480ed6d8a76f01319400a314a6bec2c040/src/app/pages/config-page/config-page.component.spec.ts#L89-L92), and the existing settings E2E tests do not assert scroll ownership, tab-header position, or focus-scroll retention. A focused Playwright case in `e2e/tests/settings/settings-basic.spec.ts` should use a short viewport, expand a long section, scroll the active `.mat-mdc-tab-body-content`, focus a lower control, and verify that the tab body retains its nonzero scroll position while the header Y coordinate and outer route scroll position remain unchanged.

2. **Minor: preserve the previous effective content width.** At [`config-page.component.scss:54-57`](https://github.com/super-productivity/super-productivity/blob/a991ec480ed6d8a76f01319400a314a6bec2c040/src/app/pages/config-page/config-page.component.scss#L54-L57), `padding-inline: var(--s2)` now sits inside a globally `border-box` element whose total max-width remains 900px. This makes wide-screen settings sections 868px instead of 900px, and below 600px changes the previous 8px side spacing to 16px (for example, a 390px viewport goes from 374px of usable width to 358px). Responsive 8px/16px padding plus a compensated border-box max width would preserve the old layout while keeping the tab header flush.

One scope question: if always-visible, flush navigation is an independently desired UX change, the nested-scroll approach is reasonable once covered by the regression test. If the goal is only the original scrollbar-induced horizontal shift, `scrollbar-gutter: stable` on the existing scroll host is a much smaller solution, preserves the March focus fix and single footer, and is already used elsewhere in this repository.

**Verdict: needs changes** for the missing regression test; the preview behavior itself looks correct.

Issue #8778, opened 2026-07-05
🚨 An error on the UI


MitterGYanik:
### Version Used

v.18.13.0I iOS18.4

### Current Behavior

I'm using a third-party input method (Sogou Input), and when I click "+" to edit, there's an issue with the editing box UI

https://github.com/user-attachments/assets/a673485e-d90a-407f-a7e0-05a7fc635253

### Expected Behavior

Normally, it should stick right to the top of the keyboard, not get sent all the way to the top of the screen - that's really weird

### Steps to Reproduce

1. Select any third-party input method(if you want to be precise, it's the Sogou input method) 2. Click “+” 3. UI issue appears

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

MitterGYanik:
Thanks for taking a look at this issue.

I'd like to provide more details about my environment, as it might help with reproducing the bug:

- **Device:** iPhone XS (A12 chip)
- **iOS Version:** 18.4 (build 22E240)
- **Sogou Input Method Version:** 20.11.0
- **App Store Region:** I downloaded Sogou from the **non-China App Store** (e.g., Hong Kong or US store), not the Chinese mainland store.

**A few additional observations:**
- This UI issue **only happens** when the Sogou keyboard is active. The native iOS keyboard works perfectly fine.
- I can reproduce it 100% of the time on my device with the steps I provided.

If you need any more info, like a console log or a screen recording with touch points visible, just let me know. Happy to help debug further.

Thanks!

johannesjo:
Thanks a lot for the detailed follow-up — the "only with Sogou, native keyboard is fine" detail was exactly what pinned this down. 🙏

The cause looks to be that some third-party iOS keyboards report a bogus, near-full-screen size to the app when they appear. The add-task bar positions itself just above the reported keyboard height, so a wildly-too-large value pushes it to the top of the screen. The native keyboard reports a correct size, which is why it behaves.

I've put up a fix in #8848: it now sanity-checks that reported height and, when it looks implausible, measures the actually-obscured area instead. Well-behaved keyboards are unaffected. It'll go out in an upcoming release.

[...]

PR #8993, opened 2026-07-14
fix(ios): prevent keyboard frames from resizing webview

johannesjo:
## Summary

- prevent Capacitor keyboard notifications from resizing the iOS WKWebView
- derive keyboard-aware layout from the current layout viewport and `VisualViewport` geometry
- retain the sanitized native keyboard height only as a short-lived fallback
- keep the global add-task bar aligned with the visible viewport bottom
- add regression coverage for the Capacitor configuration and iOS viewport calculations

## Root cause

Capacitor's native resize mode applies each iOS keyboard frame notification directly to the WKWebView. Keyboard transitions can emit several different frame heights—such as the reported 75 → 397 → 357 sequence—and even a stable system-keyboard height can leave the web layout and fixed UI using inconsistent geometry.

Using non-resizing mode keeps the layout viewport stable. The app now calculates the actually visible bottom edge from `VisualViewport.offsetTop + VisualViewport.height`, including viewport panning and rotation, and uses the bounded plugin height only until browser geometry is available.

## User impact

The add-task bar and other keyboard-aware fixed surfaces should remain immediately above the keyboard instead of jumping to the top of the screen. The change is isolated to iOS; Android behavior is unchanged.

Addresses the iOS behavior reported in #8778.

## Validation

- full `npm test` suite, including 12,576 default-timezone and 12,562 alternate-timezone Angular tests
- full production `npm run build`
- focused iOS keyboard/config regression specs
- `npm run checkFile` for every modified TypeScript and SCSS file
- TypeScript, SCSS, and custom lint-rule checks

### DIFF
--- a/capacitor.config.ts
+++ b/capacitor.config.ts
@@ -1,4 +1,5 @@
 import type { CapacitorConfig } from '@capacitor/cli';
+import { KeyboardResize } from '@capacitor/keyboard';
 
 const config: CapacitorConfig = {
   appId: 'com.super-productivity.app',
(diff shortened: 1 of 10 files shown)


johannesjo:
Closing as superseded by #8995, which fixes #8778 at the add-task-bar consumer with the narrower overlay-only rule and rendered regression coverage. The broader KeyboardResize.None approach here also needs separate follow-up before it is safe: keep viewport tracking through keyboardDidHide, and separate VisualViewport height from its bottom coordinate for panned-viewport consumers. The unrelated sync harness isolation commit can be resubmitted independently if still needed.

PR #8995, opened 2026-07-14
fix(tasks): keep add-task bar visible above iOS keyboard

### LINKED ISSUE #8778: 🚨 An error on the UI

johannesjo:
## Summary

- use the existing overlay-only keyboard offset for the global add-task bar on touch-only iOS
- preserve the measured keyboard-height behavior on Android and other touch builds
- keep hybrid iOS/iPad layouts unchanged
- add rendered layout coverage for resized iOS, overlaying iOS, non-iOS touch, and hybrid iOS

## Root cause

Capacitor's native iOS keyboard mode already shrinks the WebView. The global add-task bar also applied the keyboard's full measured height, so the bar was offset twice and could move out of view. The existing `--keyboard-overlay-offset` variable already represents only the portion still overlaying the current viewport, so the iOS rule now consumes that value.

## User impact

The global add-task bar should remain visible above stock and third-party keyboards on iPhone without changing Android or hybrid iPad positioning.

Fixes #8778.

## Validation

- `npm run test:file -- src/app/features/tasks/add-task-bar/add-task-bar.component.spec.ts` — 45/45 passed
- `npm run checkFile` passed for both changed files
- `npm run build` passed
- full pre-push lint and test hooks passed, including 12,573 default-timezone Angular tests and 12,559 Los Angeles-timezone Angular tests

## Residual verification

Physical iPhone testing is still recommended to confirm there is no visible transient during the native keyboard resize animation.


johannesjo:

Issue #8968, opened 2026-07-13
🚨 Drag and drop does not work on collapsed project section


jcuypers:
### Version Used

18.14.0L linux deb

### Current Behavior

Moving items via drag and drop for items that are outside as well as inside project sections is only possible to sections that are in expanded state (even when a section is empty)

### Expected Behavior

Dragging and dropping should to a just created, empty, non expanded section should be possible

### Steps to Reproduce

1) Create a new item in a project
2) create a new section, don't click on expand.
3) try to drag and drop the new item into the collapsed section

alternative.   
1)  drag an drop and existing item within a section
2)  try to drop it in a collapsed section (new or existing)

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

PR #8996, opened 2026-07-14
fix: allow drops into collapsed project sections

hasnainkhatri87:
## Summary\n- keep an empty task drop target available while a project section is collapsed\n- reuse the existing section-aware task move handling\n\nFixes #8968\n\n## Testing\n- git diff --check passed\n- Repository checker and targeted tests could not run: locked dependency installation timed out on Windows before npm binaries were available.

### DIFF
--- a/src/app/features/work-view/work-view.component.html
+++ b/src/app/features/work-view/work-view.component.html
@@ -228,6 +228,14 @@
                         [listModelId]="section.id"
                       ></task-list>
                     </collapsible>
+                    @if (!section.isExpanded) {
+                      <task-list
+                        class="collapsed-section-drop-target"
+                        [tasks]="[]"
+                        [listId]="'PARENT'"
+                        [listModelId]="section.id"
+                      ></task-list>
+                    }
                   </div>
                 }
               </div>


johannesjo:
Thanks for the focused patch. I double-checked the current head (`9313cbf`), including the CDK list wiring and the existing section tests. This still needs changes before merge:

1. **Section reordering can be intercepted by the new task target and crash.** At [`work-view.component.html:232`](https://github.com/super-productivity/super-productivity/blob/9313cbfab2da7a0ace26514a069be832ae080271/src/app/features/work-view/work-view.component.html#L232-L237), a section drag is connected to the nested task list through the root `cdkDropListGroup`. A `Section` has no `parentId`, so `TaskListComponent.enterPredicate` accepts it as a parent task. If it is released there, the source is `sections-wrapper`, which has no `cdkDropListData`; `drop()` then dereferences `srcListData.groupTagId` and throws. Please reject section-shaped data in the task-list predicate (or otherwise isolate the two drag domains) and cover this path.

2. **The visible section header is still not the drop target.** The added `task-list` is a sibling below `<collapsible>`, and it inherits the work-view task-list `min-height: 50px` plus host padding. Collapsing a section therefore adds an anonymous blank row beneath the header; the header itself remains under the outer section list, whose predicate rejects task drags. Please make the receiving area coincide with/overlay the collapsed header without adding layout height, and retain clear receiving feedback.

3. **Please add an active collapsed-state browser regression test.** Existing active coverage drops only into expanded sections, and the cross-section browser test is `fixme`. The test should collapse a populated target, drop into it from the unsectioned list and another section, and verify section reordering across the collapsed target does not enter the task-list path.

One behavior also needs an explicit decision: [`[tasks]="[]"`](https://github.com/super-productivity/super-productivity/blob/9313cbfab2da7a0ace26514a069be832ae080271/src/app/features/work-view/work-view.component.html#L234) gives every drop a null anchor, so tasks are always prepended and successive drops reverse their drop sequence. Please confirm and test that ordering, or provide the intended anchor.

Verification note: `git diff --check` passes, but only the welcome check has run. CI, CodeQL, preview, packaging, and sync workflows are currently `action_required`, so the implementation has not received relevant CI coverage yet.

Minor housekeeping: the PR body contains literal `\n` text, and GitHub reports no closing issue reference; reformatting it will let `Fixes #8968` auto-close the issue.

Issue #8997, opened 2026-07-14
fix(sync): deleteProject losing LWW resurrects an empty project — cascaded tasks lost


johannesjo:
When a `deleteProject` operation loses an LWW conflict, its reducer cascade is not compensated — the same silent, permanent cross-device data-loss class as #8956, but for PROJECT parents. Found by an adversarial review of the #8956 fix (PR #8990); the TASK-parent machinery added there does not cover PROJECT cascades.

### Background

`deleteProject` is a **single-entity PROJECT delete** whose payload carries the cascade lists (`allTaskIds`, `noteIds`) — `src/app/root-store/meta/task-shared.actions.ts:294`. Applying it removes the project, every listed task (task reducer `removeMany(allTaskIds)`), the project's notes, and its sections.

PR #8990 recreates cascade-deleted **subtasks of a winning TASK parent** (`_createSubtaskRecreationOpsForWinningParent`), but that helper is TASK-only. A winning PROJECT compensation recreates the project entity alone.

### a) Remote `deleteProject` loses LWW to a local project edit — tasks lost everywhere else

Device B deletes project P (tasks t1..tn cascade-deleted on B). Device A concurrently edits P (rename) with a newer timestamp and wins. A's compensation recreates only the P entity:

- On B (and every client that applied the delete): P resurfaces **empty** — t1..tn stay deleted, `P.taskIds` dangles.
- On A itself: status-blind hydration replay of the durable loser row cascades t1..tn away after a crash/restart, then recreates only P — post-restart state ≠ live state (replay-determinism violation).

A has all of P's active tasks in its store at resolution time, so recreate-after-delete TASK ops can be emitted exactly like the TASK-parent path (authoritative id list: the delete payload's `allTaskIds`).

### b) Local `deleteProject` loses LWW to a remote project edit — tasks lost on this client only

The local delete was applied optimistically (tasks already gone from this client's store) and is then rejected; other clients never applied it. This client permanently diverges (missing tasks). Recovery would require capturing task snapshots at delete time (as `deleteTasks` now does via its `tasks` payload) — but a project can hold thousands of tasks, so a naive snapshot risks the server per-op payload cap. Needs a design decision (chunked sidecar rows vs. bounded snapshot vs. accept-and-log).

### c) Known limitations even with (a) fixed

[...]

PR #8998, opened 2026-07-14
fix(sync): recreate cascaded project tasks when deleteProject loses LWW

johannesjo:
## Summary

- recreate a deleted project's cascaded tasks when its `deleteProject` operation loses an LWW conflict outright
- mirror of the TASK-parent recreate-after-delete path from #8990, wired into the same pure-loser pass

## Root cause

`deleteProject` cascade-deletes every task in its payload's `allTaskIds`, but a losing delete only produced a PROJECT compensation. Every client that had applied the delete resurrected an **empty** project (dangling `taskIds`, task subtree silently lost on all devices), and the winning client's own status-blind hydration replay of the durable loser row cascaded its tasks away after a crash/restart — live state ≠ post-restart state.

## Fix

`_createTaskRecreationOpsForWinningProject`: for each id in the losing delete's `allTaskIds` still present in the local store, emit a recreate-after-delete LWW TASK snapshot with a clock dominating the delete (same proxy-clock tradeoff documented on the TASK path). Tasks deleted on this device stay deleted. Rows are persisted after the durable loser (correct seq-ordered replay) and uploaded, but not live-applied — local state already holds the tasks.

Out of scope, tracked on #8997: (b) local `deleteProject` losing to a remote edit (needs snapshot capture at delete time), (c) notes (`noteIds`, no NOTE `RECREATE_FALLBACK`) and archived tasks.

## Validation

- new unit test: recreations carry the recreate flag, dominate the delete's clock, skip locally-deleted tasks, and nothing applies live (`conflict-resolution.service.spec.ts`, 177/177)
- new three-way convergence integration test against the real IndexedDB store and real reducer chain: live == status-blind restart replay == originating client (`conflict-resolution-persistence.integration.spec.ts`, 7/7)
- both tests mutation-verified: they fail against the unfixed service
- `checkFile` clean on all touched files

## Stacking

Based on `feat/issue-8956-86c4b7` (#8990) — depends on its `recreatesEntityAfterDelete` machinery and pure-loser pass. Merge after #8990; GitHub will retarget to `master` when the base branch is deleted.

Addresses part (a) of #8997.


johannesjo:
Superseded by #9009. That PR takes the delete-wins approach (schema-v4 `projectDeleteWins` marker): a project delete now wins the LWW conflict outright instead of recreating cascaded tasks on loss — simpler, no payload scaling, and it reuses the existing archive-win resolution pipeline. Closing this recreate-based branch.

PR #8999, opened 2026-07-14
fix(tasks): decode multipart and transfer-encoded eml bodies #8975

### LINKED ISSUE #8975: 🚨 EML import
### Version Used

v18.4.0 (Apple Silicon dmg)

### Current Behavior

Release notes for 18.4.0 and commit https://github.com/super-productivity/super-productivity/pull/8282/commits/10d5f1c8d3ecefac85a07bb67ecbf632d1caf20c seem to suggest EML files are imported when dropped on the '+' in the taskbar, with the title of the email as the title of the task and the plain text of the body of the email as the note of the task. The task is created but the note is empty.

### Expected Behavior

EML files are imported when dropped on the '+' in the taskbar, with the title of the email as the title of the task and the body of the email as the note of the task. The task is created and the note is filled with the email body.

### Steps to Reproduce

1. Drag the eml file on the '+' sign (in Today or Inbox)
2. The task is created with the correct title.
3. Check the notes of the task, these are empty.

The eml file is created by saving an email from Microsoft Outlook 365 on a Mac. The file **does** have a text/plain content part.

### Console Output

```shell

```

### Documentation

MannXo:
## Problem

Fixes #8975.

Dropping a `.eml` file onto the add-task button (**+**) creates a task with the correct title but an empty note. `eml-parser.ts` only ever returned `text` for a single, unencoded `text/plain` body — but nearly every real-world client (including Outlook, per the repro in #8975) sends `multipart/alternative` (plain + HTML) with `quoted-printable` or `base64` transfer encoding on the plain part, so the body is always omitted.

This was flagged as a likely follow-up in the parser's own hardening pass (#8901): "a future, dependency-free lever if users report dropped bodies."

## Solution

`src/app/util/eml-parser.ts`:

- Walk `multipart/*` structures (recursing into nested multipart, e.g. `mixed` > `alternative`, with a depth cap) for the first `text/plain` leaf.
- Decode `quoted-printable` and `base64` transfer encodings on that leaf.
- Unchanged by design: HTML bodies and non-UTF-8/ASCII charsets are still never decoded — same untrusted-content threat model as before, just reaching more of the legitimate plain text.

`src/app/util/eml-parser.spec.ts`: replaced the two tests asserting the old omit-multipart/omit-quoted-printable behavior with tests asserting correct decoding, and added coverage for a multipart/alternative body with an HTML sibling (Outlook-style), nested multipart/mixed > multipart/alternative (attachment scenario), multipart with only an HTML part (still omitted), a malformed/unterminated multipart body, the MIME depth cap, and base64 both top-level and nested.

`docs/wiki/2.03-Add-Tasks.md`: updated the `.eml` drop note to describe the widened support.

**Testing:**

- New tests verified to fail against the pre-fix parser (reverted `eml-parser.ts`, kept the new specs) — the 7 new-behavior tests failed as expected, the other 22 (HTML-omission, malformed input, depth cap, charset gate) passed either way, confirming the tests actually exercise the fix.
- `eml-parser.spec.ts` (29/29), `eml-drop.service.spec.ts`, `eml-drop.directive.spec.ts`: all passing.
- `npm run lint`: clean.
- Full `npm test`: no new failures introduced.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/docs/wiki/2.03-Add-Tasks.md
+++ b/docs/wiki/2.03-Add-Tasks.md
@@ -13,7 +13,7 @@ This how-to covers adding new tasks to your list. For subtasks, scheduling, repe
 1. With the add task bar open, type the task title in the input. The placeholder shows an example: **A task title #tag @16:00 t25m** (you can use **#tag**, **@time**, **@date**, and **[t]time** in the title; see below).
 2. Press **Enter** or click the **Add task** button (plus icon next to the input). The task is added to the current project or context (Today / tag).
 
-> You can also create a task by dropping an `.eml` email file onto the **add task** button (the **+** icon in the header). The task title becomes `sender: subject`. Simple, unencoded plain-text bodies are saved to the task's notes as literal text; unsupported body formats create a title-only task.
+> You can also create a task by dropping an `.eml` email file onto the **add task** button (the **+** icon in the header). The task title becomes `sender: subject`. The plain-text part of the email — including quoted-printable/base64-encoded bodies and multipart messages with an HTML alternative — is saved to the task's notes as literal text; HTML-only bodies and unsupported encodings create a title-only task.
 
 The new task appears at the **top** or **bottom** of the list depending on the add bar setting (see “Add to top or bottom” below).
 
--- a/src/app/core/drop-paste-input/eml-drop.service.spec.ts
+++ b/src/app/core/drop-paste-input/eml-drop.service.spec.ts
@@ -104,6 +104,28 @@ describe('EmlDropService', () => {
       'From: Alice <alice@example.com>',
       'Subject: Encoded',
       'Content-Type: text/plain',
+      'Content-Transfer-Encoding: x-unknown-encoding',
+      '',
+      'body',
+      '',
+    ].join('\n');
+
+    await service.createTaskFromEml(makeFile(encodedEml));
+
+    expect(taskService.add).toHaveBeenCalledWith(
+      'Alice: Encoded',
+      false,
+      { notes: undefined },
+      false,
+      true,
+    );
+  });
+
+  it('should decode a base64-encoded body and wrap it in the notes code fence', async () => {
+    const encodedEml = [
+      'From: Alice <alice@example.com>',
+      'Subject: Encoded',
+      'Content-Type: text/plain',
       'Content-Transfer-Encoding: base64',
       '',
       'Ym9keQ==',
@@ -115,7 +137,7 @@ describe('EmlDropService', () => {
     expect(taskService.add).toHaveBeenCalledWith(
       'Alice: Encoded',
       false,
-      { notes: undefined },
+      { notes: '```\nbody\n```' },
       false,
       true,
     );
--- a/src/app/util/eml-parser.spec.ts
+++ b/src/app/util/eml-parser.spec.ts
@@ -360,6 +360,32 @@ describe('parseEml', () => {
     expect(data.text).toBeUndefined();
   });
 
+  it('should skip an attachment whose type-less disposition carries only an RFC 2231 continued filename', async () => {
+    const continuedFilenameAttachmentEml = [
+      'From: Alice <alice@example.com>',
+      'Subject: Continued filename attachment',
+      'Content-Type: multipart/mixed; boundary="mixed"',
+      '',
+      '--mixed',
+      'Content-Type: text/html',
+      '',
+      '<p>the real body</p>',
+      '--mixed',
+      'Content-Type: text/plain',
+      'Content-Disposition: ; filename*0="secret"; filename*1=".txt"',
+      '',
+      'PRIVATE ATTACHMENT',
+      '--mixed--',
+      '',
+    ].join('\n');
+
+    const data = await parseEml(
+      makeFile(continuedFilenameAttachmentEml, 'continued-filename-attachment.eml'),
+    );
+
+    expect(data.text).toBeUndefined();
+  });
+
   it('should return undefined for a multipart body with no closing boundary (malformed)', async () => {
     const unterminatedEml = [
       'From: Alice <alice@example.com>',
--- a/src/app/util/eml-parser.spec.ts
+++ b/src/app/util/eml-parser.spec.ts
@@ -190,6 +190,84 @@ describe('parseEml', () => {
     expect(data.text).toBeUndefined();
   });
 
+  it('should not import a text/plain attachment as the note when the real body is HTML', async () => {
+    const htmlWithTextAttachmentEml = [
+      'From: Alice <alice@example.com>',
+      'Subject: HTML with attachment',
+      'Content-Type: multipart/mixed; boundary="mixed"',
+      '',
+      '--mixed',
+      'Content-Type: text/html',
+      '',
+      '<p>the real body</p>',
+      '--mixed',
+      'Content-Type: text/plain',
+      'Content-Disposition: attachment; filename="notes.txt"',
+      '',
+      'ATTACHMENT',
+      '--mixed--',
+      '',
+    ].join('\n');
+
+    const data = await parseEml(
+      makeFile(htmlWithTextAttachmentEml, 'html-with-attachment.eml'),
+    );
+
+    expect(data.text).toBeUndefined();
+  });
+
+  it('should skip a text/plain attachment that precedes the real plain-text body', async () => {
+    const attachmentFirstEml = [
+      'From: Alice <alice@example.com>',
+      'Subject: Attachment first',
+      'Content-Type: multipart/mixed; boundary="mixed"',
+      '',
+      '--mixed',
+      'Content-Type: text/plain',
+      'Content-Disposition: attachment; filename="notes.txt"',
+      '',
+      'ATTACHMENT',
+      '--mixed',
+      'Content-Type: text/plain',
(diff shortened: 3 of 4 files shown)


MannXo:
Multipart Outlook-style .eml dropped → note fully decoded (title, em dash, and body all correct):
image

johannesjo:
Thanks for the focused fix. I double-checked the current head (`9774af0`) and found one **Important** correctness issue before merge:

### Important

[`src/app/util/eml-parser.ts:84`](https://github.com/super-productivity/super-productivity/blob/9774af0ed56bcbe6bc9f1cecd6d0bb400c7400d2/src/app/util/eml-parser.ts#L84-L113) — `_extractPlainText()` never checks `Content-Disposition`. Since multipart traversal returns the first supported `text/plain` child in wire order:

- An HTML-body email with a `text/plain` attachment imports the attachment contents as the task note instead of producing a title-only task.
- A text attachment before the real plain-text body shadows the real body. An empty attachment also stops traversal because `"" !== undefined`.
- A multipart entity marked as an attachment is still recursively searched, so checking only leaf parts would not fully fix this.

I reproduced the first two cases against the exact PR head; both returned `"ATTACHMENT"` rather than the expected no body / `"REAL BODY"`.

The smallest fix is to parse the leading disposition token case-insensitively and return `undefined` for `attachment` **before** the multipart branch, which skips the whole attached subtree. Please add regression coverage for an HTML body plus text attachment and for an attachment preceding the real plain body.

All required CI checks are currently green, but this should be fixed before merge.

johannesjo:
Thanks for addressing the explicitly marked attachment cases. I double-checked the latest head (`398fff68`) and found one remaining **Important** attachment-classification case:

[`src/app/util/eml-parser.ts:84`](https://github.com/super-productivity/super-productivity/blob/398fff68d8bbb1606aef5dfa7d280f9b0803ab83/src/app/util/eml-parser.ts#L84-L86) — the guard skips only a disposition whose token is exactly `attachment`. Multipart mail can still contain a file-like `text/plain` part identified only by a legacy `Content-Type: text/plain; name="notes.txt"` parameter, because `Content-Disposition` is optional. Also, RFC 2183 requires an unrecognized disposition type to be treated as `attachment`, but values such as `Content-Disposition: x-download; filename="notes.txt"` currently pass through.

I reproduced both cases against the exact head with an HTML body followed by the text file part. In each case `parseEml()` returned `"PRIVATE ATTACHMENT"` as `text`, so dropping the email would save and potentially sync attachment contents instead of creating the expected title-only task.

The smallest fix is to treat any present disposition other than `inline` as an attachment and conservatively skip filename-bearing parts (`filename` or legacy `name`) when no explicit inline disposition says otherwise. Please add regression coverage for the name-only and unknown-disposition cases.

RFC reference: https://www.rfc-editor.org/rfc/rfc2183.html#section-2.8

MannXo:
Thanks for the comments @johannesjo 
reflected in the most recent commit

johannesjo:
Thanks for addressing the regular `name=` and unknown-disposition cases. I rechecked the current head (`739bf087`) and found one remaining **Important** attachment-classification gap:

[`src/app/util/eml-parser.ts:312`](https://github.com/super-productivity/super-productivity/blob/739bf0877e7469a4a5ce4488dfcceebbc62eea57/src/app/util/eml-parser.ts#L295-L317) — `_parseContentType()` only recognizes the literal parameter key `name`. [RFC 2231](https://www.rfc-editor.org/rfc/rfc2231.html#section-3) extends MIME media-type parameters with encoded and continued spellings, so `name*=...` and `name*0=...; name*1=...` represent the same parameter. `Content-Disposition` is optional, so these forms can be the only filename hint on a part.

I reproduced this against the exact head with an HTML body followed by:

```text
Content-Type: text/plain; name*0="notes"; name*1=".txt"

PRIVATE ATTACHMENT
```

A standards-aware MIME parser resolves the continued value to filename `notes.txt`, but `parseEml()` returns `"PRIVATE ATTACHMENT"` as `text`. The caller then stores that non-empty text as the task note, so dropping the email can save and sync attachment contents instead of creating the expected title-only task.

The smallest fix is a presence-only matcher that treats the RFC 2231 forms of `name` (`name`, `name*`, `name*<section>`, and `name*<section>*`) as the same attachment hint; no filename decoding or reconstruction is needed. Please add a regression test for a continued `name*0`/`name*1` parameter with no `Content-Disposition`.

All required CI checks are currently green, but this should be fixed before merge.

johannesjo:
The `name*` change looks correct — thanks for chasing the RFC 2231 spellings.

One remaining **symmetric gap** in the same spot, then I think this is good to go: `_parseContentDisposition`'s `hasFilename` fallback still matches only the literal `filename=` key, so the RFC 2231 encoded/continued spellings (`filename*`, `filename*0`, `filename*0*`, …) slip through — the mirror of the `name*` fix you just made.

It's low severity: it's only reachable through the **type-less-disposition** fallback (disposition present but with no parseable type token) **and** with no `Content-Type; name` hint. A normal `Content-Disposition: attachment; filename*0=…` is already caught by the type check, and a co-present `name*` is now caught too. But it's worth closing for consistency with the fallback itself.

Fix + regression test below (all 37 `eml-parser` specs green; the new test fails without the source change, so it genuinely pins the behavior). The `hasFilename` check now splits each param at `=` and matches the key against a presence-only regex mirroring `_NAME_PARAM_KEY_RE` — value never decoded, and it also tolerates whitespace before `=` that the old `startsWith('filename=')` would have missed.

```diff
diff --git a/src/app/util/eml-parser.spec.ts b/src/app/util/eml-parser.spec.ts
index 0e61c07aad..07c6b17ffb 100644
--- a/src/app/util/eml-parser.spec.ts
+++ b/src/app/util/eml-parser.spec.ts
@@ -360,6 +360,32 @@ describe('parseEml', () => {
     expect(data.text).toBeUndefined();
   });
 
+  it('should skip an attachment whose type-less disposition carries only an RFC 2231 continued filename', async () => {
+    const continuedFilenameAttachmentEml = [
+      'From: Alice <alice@example.com>',
+      'Subject: Continued filename attachment',
+      'Content-Type: multipart/mixed; boundary="mixed"',
+      '',
+      '--mixed',
+      'Content-Type: text/html',
+      '',
+      '<p>the real body</p>',
+      '--mixed',
+      'Content-Type: text/plain',
+      'Content-Disposition: ; filename*0="secret"; filename*1=".txt"',
+      '',
+      'PRIVATE ATTACHMENT',
+      '--mixed--',
+      '',
+    ].join('\n');
+
+    const data = await parseEml(
+      makeFile(continuedFilenameAttachmentEml, 'continued-filename-attachment.eml'),
+    );
+
+    expect(data.text).toBeUndefined();
+  });
+
   it('should return undefined for a multipart body with no closing boundary (malformed)', async () => {
     const unterminatedEml = [
       'From: Alice <alice@example.com>',
diff --git a/src/app/util/eml-parser.ts b/src/app/util/eml-parser.ts
index b9139742ed..197560bb97 100644
--- a/src/app/util/eml-parser.ts
+++ b/src/app/util/eml-parser.ts
@@ -20,6 +20,10 @@ const MAX_MIME_DEPTH = 10;
 // decoded/reassembled since we only need to know a filename hint exists.
 const _NAME_PARAM_KEY_RE = /^name(\*\d*\*?)?$/;
 
+// Same RFC 2231 spellings for the `Content-Disposition` `filename` parameter
+// (`filename`, `filename*`, `filename*0`, `filename*0*`, ...); presence-only.
+const _FILENAME_PARAM_KEY_RE = /^filename(\*\d*\*?)?$/;
+
 /**
  * Minimal, dependency-free RFC 822 / MIME reader for the drop-an-`.eml` feature.
  *
@@ -264,10 +268,11 @@ const _splitParams = (value: string): string[] => {
   return parts;
 };
 
-// RFC 2183 disposition-type + presence of a `filename` parameter. A
-// disposition value that fails to parse a leading token (e.g. malformed,
-// leading straight into `; filename=...`) still surfaces `hasFilename` so the
-// caller can fall back on that signal.
+// RFC 2183 disposition-type + presence of a `filename` parameter (in any RFC
+// 2231 spelling: `filename`, `filename*`, `filename*0`, ...). A disposition
+// value that fails to parse a leading token (e.g. malformed, leading straight
+// into `; filename=...`) still surfaces `hasFilename` so the caller can fall
+// back on that signal.
 const _parseContentDisposition = (
   value?: string,
 ): { type?: string; hasFilename: boolean } => {
@@ -277,9 +282,10 @@ const _parseContentDisposition = (
 
   const parts = _splitParams(value);
   const type = parts[0].trim().split(/[\s(]/)[0].toLowerCase() || undefined;
-  const hasFilename = parts
-    .slice(1)
-    .some((part) => part.trim().toLowerCase().startsWith('filename='));
+  const hasFilename = parts.slice(1).some((part) => {
+    const eq = part.indexOf('=');
+    return eq >= 0 && _FILENAME_PARAM_KEY_RE.test(part.slice(0, eq).trim().toLowerCase());
+  });
 
   return { type, hasFilename };
 };
```

johannesjo:
Thank you very much! <3

PR #9001, opened 2026-07-14
fix(tasks): keep REST project moves atomic across replay

### LINKED ISSUE #8983: Local REST API: PATCH projectId does not update project taskIds arrays (tasks invisible after move)
## Describe the bug

When moving a task between projects via the Local REST API (`PATCH /tasks/:id` with a new `projectId`), the task entity is updated correctly but **neither project's `taskIds` array is updated**. This makes tasks invisible in the destination project's UI and leaves stale entries in the source.

## Steps to reproduce

Given an existing task in project Banking (18 taskIds) being moved to Money (5 taskIds):

1. Note current project taskIds counts
2. `PATCH /tasks/TASK_ID -d '{"projectId":"MONEY_ID"}'`
3. Response: `200 OK`, task.projectId is correctly set
4. Check project taskIds: **neither updated**

```
BEFORE: Banking=18 taskIds, Money=5 taskIds
PATCH:  ok: True, projectId=Money
AFTER:  Banking=18 taskIds, Money=5 taskIds   ← UNCHANGED
```

The task data is correct (API returns it when queried by project name), but the SP UI renders from the project's `taskIds` array — so the user sees an empty destination project.

## Attempted workaround: archive + restore

Archive/restore partially works but causes **data corruption**:

1. `POST /tasks/TASK_ID/archive` → hides task
2. `POST /tasks/TASK_ID/restore` → adds to DESTINATION index but does **NOT** remove from SOURCE
3. Result: task appears in **both** projects' taskIds arrays

```
AFTER A+R: Banking=18 taskIds (STILL LISTED!), Money=6 taskIds (gained it)
           TASK IN BOTH PROJECT INDEXES ← CORRUPTION
```

## Expected behavior

`PATCH projectId` should trigger ProjectService to:
- Remove the taskId from the source project's `taskIds` array
- Add the taskId to the destination project's `taskIds` array

`POST /tasks/:id/restore` should also clean up stale source references.

## Environment

- Super Productivity v18.13.1 (Flatpak)
- Linux (CachyOS/Arch, kernel 7.1.3)
- Local REST API on 127.0.0.1:3876

## Additional context

- Same class of bug as #7346 (parentId silently dropped on PATCH, fixed April 2026) — the PATCH handler likely calls `_taskService.update()` but does not notify `ProjectService` to reindex
- UI drag-and-drop between projects works correctly (different codepath)
- `POST /tasks` (creating new tasks with `parentId`) works correctly — only PATCH-based moves are affected
- Blocks agent/script-based task migration workflows

johannesjo:
## Problem

Fixes #8983.

The Local REST API could change a task's `projectId` without atomically updating the source and destination projects' `taskIds`. The task then disappeared from the destination UI, and archive/restore could leave it indexed by both projects.

Project moves also need a stable subtask footprint during sync replay. Re-deriving children from receiver state can move a different set of tasks after concurrent or superseded operations.

## Solution

- Route actual REST project moves through the existing atomic task move path while preserving unchanged `projectId` round trips.
- Capture the exact subtask IDs affected by a project move and carry the newest footprint through conflict resolution, superseding, and synthetic LWW operations.
- Repair stale project, section, parent, and tag references during CRUD, restore, and LWW replay.
- Handle archived replay targets, restore collisions, missing legacy lists, and prototype-like entity IDs safely.
- Document REST project-move behavior and add regression coverage for direct, restore, conflict, and replay paths.

This keeps one user intent as one operation and makes the resulting state deterministic across local and synced clients.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [ ] Other (please describe)

## Validation

- `npm run lint`
- `npm run checkFile <filepath>` for every changed TypeScript file
- 475/475 focused task lifecycle, section, LWW, conflict-resolution, and superseded-operation tests
- Relevant REST, CRUD, lifecycle, composed reducer, and sync integration specs
- Full Angular run: 12,631 passed, 2 skipped, with one unrelated focus-timing failure in `TaskDetailPanelComponent`; that spec passed independently (20/20)

## Checklist

- [ ] Existing tests still pass (one unrelated order/timing-sensitive full-suite failure; isolated spec passes)


johannesjo:

Issue #8960, opened 2026-07-12
fix(sync): file-based provider (WebDAV/Dropbox/LocalFile) atomicity & conflict UX


johannesjo:
Four confirmed defects in the file-based provider layer. The lost-update safety argument of the whole file-based op-log rests on the provider offering an atomic compare-and-swap that WebDAV/LocalFile do not actually deliver; two others concern a force-overwrite migration path and a misleading conflict-dialog timestamp.

---

### WebDAV (and LocalFile) 'conditional' upload is non-atomic check-then-write — concurrent client's acked ops are silently clobbered
**`packages/sync-providers/src/file-based/webdav/webdav-api.ts:197`** · severity **high** · race-condition · verifier confidence high

The whole file-based lost-update guarantee rests on the provider's uploadFile(path, data, revToMatch, false) being an atomic compare-and-swap (see the safety argument in src/app/op-log/sync-providers/file-based/file-based-sync-adapter.service.ts:556-572). Dropbox (mode:update+rev) and OneDrive (If-Match header, onedrive.ts:221) deliver that server-side. WebDAV does NOT: WebdavApi.upload() implements the precondition as a client-side GET of the current file, hash comparison against expectedRev (lines 197-215), and then a plain unconditional PUT (lines 222-228) — no If-Match / If-Unmodified-Since / If header, even though the code deliberately uses content hashes instead of server ETags. LocalFileSyncBase.uploadFile (packages/sync-providers/src/file-based/local-file/local-file-sync-base.ts:131-149) has the identical read-compare-write TOCTOU. The post-upload _verifyUpload re-GET only catches writes landing AFTER our PUT, not between the check and the PUT.

[...]

PR #9002, opened 2026-07-14
fix(sync): recover split-file ops newer than a lagging snapshot

johannesjo:
## Problem

**Silent data loss in split-file ("Surgical") sync.** In split-file mode the snapshot file (`sync-state.json`) is written less often than the ops file (`sync-ops.json`), so the retained operation suffix can contain ops newer than the snapshot's vector clock. On a fresh download / gap recovery the client hydrated state wholesale from the *lagging* snapshot and then wrote **all** downloaded ops to IndexedDB. Because `getAppliedOpIds()` counts every stored op, any op newer than the snapshot was marked "applied" without its effect ever reaching state, and was then filtered out on subsequent syncs — **permanent, silent data loss** for anything that landed after the last snapshot write.

Two smaller issues surfaced while writing coverage for this:

- The `isUseSplitSyncFiles` ("Surgical sync") toggle was **dropped when saving the sync form**, so the setting never persisted.
- E2E coverage for split-file sync, backup-import archives, forward-compat op blocking, and several persisted-model / repeat / time-tracking sync paths was missing or weaker than intended.

## Solution

- **`file-based-sync-adapter.service.ts`** — report the snapshot file's own vector clock as `snapshotVectorClock` (it previously reported the ops-file clock, overstating what the snapshot actually covers).
- **`operation-log-sync.service.ts`** — split downloaded ops into snapshot-**covered** (dominated by the snapshot clock → persisted only) vs **delta** (`GREATER_THAN`/`CONCURRENT` → replayed through the existing full-state guards after hydration). Extracted `_processDownloadedOpsWithFullStateGuards` so the file-snapshot suffix and the ordinary incremental download share the same `SYNC_IMPORT`/`BACKUP_IMPORT`/version-block gates and the same cursor safety (never advance `lastServerSeq` past a blocked or cancelled op).
- **`sync-config.service.ts`** — persist `isUseSplitSyncFiles` in the global config.
- **Tests** — new unit tests for hydrate-then-replay, the "local dominates only the snapshot but is missing a delta" data-loss case, and each guard branch asserting the cursor stays put; new/strengthened e2e: WebDAV split-file bootstrap with a **provably-lagging** snapshot, backup-import archives verified directly against the `archive_young`/`archive_old` IndexedDB stores, forward-compat block-and-recover, persisted-models round-trip, repeat config, and concurrent time-tracking deltas surviving snapshot hydration.

## Type of Change

- [x] Bug fix
- [x] Other: test coverage

## Checklist

## Testing notes

- Unit suites green locally: `operation-log-sync.service.spec.ts` (139) and `sync-config` + `file-based-sync-adapter` (154).
- The WebDAV / SuperSync **e2e suites need a real run** via the `E2E Tests (Scheduled)` workflow dispatch (they require docker and weren't runnable in the dev sandbox).
- **Risk:** this touches the sync download path. Replay determinism, concurrent/remote edits, vector-clock conflicts, and cursor safety were reviewed — the delta replay is routed through the same guards as the incremental path and never advances the server cursor past a blocked/cancelled op.

> Note for reviewers: `docs/plans/2026-07-14-microsoft-365-calendar-provider.md` appears in the diff but is byte-identical to a copy already on `master`; it merges as a no-op and is unrelated to this change.


johannesjo:
Closing as **superseded by #9004**.

Since this PR was opened, #9004 (`fix(sync): file-based provider atomicity & conflict UX (#8960)`) landed on `master` and independently implemented the same split-file lagging-snapshot fix, with a more precise mechanism:

- **Snapshot clock:** both return the real snapshot vector clock instead of `opsFile.vectorClock`.
- **Covered vs. delta split:** this PR classified ops via vector-clock domination (a proxy). #9004 uses the per-op `sv` (syncVersion) boundary in the adapter and emits `snapshotAppliedOpIds` — the authoritative signal for "is this op inside the snapshot" — and additionally adds `remoteLastModified`.
- **Equivalent coverage test:** #9004 ships `applies split-file operations newer than the downloaded snapshot before advancing the cursor`, matching this PR's `should replay operations newer than a file snapshot`.

Because #9004 is already in this PR's base, the merge conflict was really the two implementations colliding — keeping both is untenable (the auto-merged spec files even encode incompatible contracts), and taking this PR's side would regress precision and drop `remoteLastModified`.

The one behaviour this PR had that #9004 does not is delta replay in the **skip-hydration** branch (local clock dominates the remote snapshot). That closes a narrow, low-severity window that #9004's author evaluated and documented as acceptable; it's tracked separately in #9011 for a small targeted fix on top of #9004's mechanism if it ever proves to matter.

PR #9003, opened 2026-07-14
fix(tasks): move task between projects on REST PATCH projectId (#8983)

### LINKED ISSUE #8983: Local REST API: PATCH projectId does not update project taskIds arrays (tasks invisible after move)

johannesjo:
Fixes #8983.

## The bug

`PATCH /tasks/:id` with a new `projectId` updated the task entity but never moved the task id between the source and destination projects' `taskIds` arrays. The task became invisible in the destination project's UI (which renders from `project.taskIds`) and left a stale entry in the source. The archive+restore workaround made it worse — the task ended up in **both** projects' `taskIds` (corruption).

## The fix

Route a `projectId` change on `PATCH /tasks/:id` through the existing, already sync/replay-hardened `TaskSharedActions.moveToOtherProject` action (via `TaskService.moveToProject`) — the same path the UI drag-drop, context menu, and board panel already use. It reindexes both projects' `taskIds` atomically (moving subtasks with the parent) and derives subtask ids from canonical state, so it replays deterministically.

This deliberately avoids the larger alternative of teaching the generic `updateTask` action + meta-reducers to detect project moves, which would have touched ~9 sync-critical files (conflict-resolution, LWW, superseded-resolver, several meta-reducers). Net production change here is **~108 lines in one handler file, zero sync-critical files touched**.

## Behaviour

- Changing `projectId` on a top-level task moves it (and its subtasks) to the target project.
- Destination must be an existing, non-archived project, else `404 PROJECT_NOT_FOUND`.
- `projectId` cannot be changed on a subtask (`400 UNSUPPORTED_FIELD`) — move its parent instead.
- Sending the current `projectId` is a no-op, so `GET`→`PATCH` round-trips are safe.
- A project move must be sent on its own — combining it with other field edits returns `400 UNSUPPORTED_FIELD` (see hardening below).

## Hardening from a multi-agent review

- **Reject combining a `projectId` move with other field edits (400).** Splitting a move + edit into two ops could degrade the residual into a title-only update that the short-syntax effect parses — a `+project` in the title would silently fight the explicit move. Keeping a move to its own request makes it exactly one op.
- **Guard `_getTaskById` / `_getTaskWithSubTasksById` with an id-equality check.** A prototype-key `taskId` (`constructor`, `__proto__`) resolves to a truthy `Object.prototype` member in the NgRx entity map and would otherwise push an `undefined` id into a synced `project.taskIds`. The `id === taskId` check rejects both that and the `{ subTasks: [] }` fallback the selector returns for a missing entity.

## Tests

67/67 in `local-rest-api-handler.service.spec.ts` (move reindexes via `moveToProject`; combined move+edit rejected; no-op round-trip; subtask rejection; missing/archived destination; prototype-key `projectId` and `taskId` guards). `checkFile` clean.

## Not covered (separate concern)

Installs already corrupted by the old bug (a task id in two projects' `taskIds`) are not retroactively repaired — that belongs in a `dataRepair` pass, not this fix.

### DIFF
--- a/docs/plans/2026-07-14-ios-testflight-master-builds.md
+++ b/docs/plans/2026-07-14-ios-testflight-master-builds.md
@@ -1,0 +1,206 @@
+# Implementation plan: Internal TestFlight builds from `master`
+
+**Status:** Plan only · **Date:** 2026-07-14
+**Difficulty:** Moderate. Allow about one engineering day split across two small
+changes, plus Apple processing and tester verification time.
+
+## Outcome and scope
+
+Use the existing iOS signing/archive workflow to send the newest eligible `master`
+state to an internal TestFlight group. Rapid pushes are coalesced: one beta may run
+and only the newest superseding beta waits. This intentionally does not produce one
+TestFlight build for every intermediate SHA.
+
+Production behavior stays event-based: final `v*` tag pushes submit the existing iOS
+release for review, prerelease tag pushes remain upload-only, and manual dispatch can
+never submit to production. Master builds are marked **TestFlight Internal Only**, so
+Apple cannot offer them to external testers or customers.
+
+Out of scope: external TestFlight, app changes, new dependencies, and the existing
+iOS/macOS App Store review-submission race. There is no new app-wide upload queue;
+the internal-only beta path neither edits App Store metadata nor opens a review.
+
+## Routing contract
+
+| Event                                  | Export                   | Fastlane lane | Production submit |
+| -------------------------------------- | ------------------------ | ------------- | ----------------- |
+| Push to `master`                       | TestFlight Internal Only | `ios beta`    | Never             |
+| Push of final `v*` tag                 | App Store Connect        | `ios release` | Yes               |
+| Push of prerelease `v*` tag            | App Store Connect        | `ios release` | No                |
+| Manual dispatch, default `beta` mode   | TestFlight Internal Only | `ios beta`    | Never             |
+| Manual dispatch, `release-upload-only` | App Store Connect        | `ios release` | Never             |
+
+The production guard must include the event type, not only the ref:
+
+```yaml
+SUBMIT_FOR_REVIEW: >-
+  ${{ github.event_name == 'push'
+      && startsWith(github.ref, 'refs/tags/v')
+      && !contains(github.ref, '-') }}
+```
+
+Manual beta dispatch is allowed only from `master`. The explicit
+`release-upload-only` mode preserves the workflow's current manual upload capability
+without inheriting its final-tag submission footgun.
+
+## Phase 0: Apple and version prerequisites
+
+- Create one internal tester group, add the intended App Store Connect users, and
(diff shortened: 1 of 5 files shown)


johannesjo:
Superseded by #9001, which fixed #8983 on master via the meta-reducer approach.

On review, #9001 already carries the two hardening guards this PR added:
- the `task?.id === taskId` guard on both task lookups, closing the vector where a prototype-key taskId (`/tasks/constructor`) could push an `undefined` id into a project's synced `taskIds`;
- correct handling of a combined `projectId`+field PATCH — #9001 keeps it a single multi-field `updateTask` op, so the short-syntax effect (title-only) never fires and the move + literal title both apply.

So there is nothing to port. Closing in favour of #9001. The lean `moveToOtherProject`-based alternative and its analysis remain here if the meta-reducer machinery ever needs revisiting.

Issue #8941, opened 2026-07-11
fix(sync): archive young/old placement can diverge across clients when archiving triggers a due flush


johannesjo:
Found by the #8900 review (Codex W1), mechanism verified.

`TaskService._moveToArchive` persists archive data before dispatching (correct, #8900), but `ArchiveService.moveTasksToArchiveAndFlushArchiveIfDue` may dispatch `flushYoungToOld` BEFORE the caller dispatches `moveToArchive`. On a due flush the originating client evaluates the flush WITH the just-archived data present, while receivers apply `flushYoungToOld` first and insert the task afterwards.

Task placement converges (threshold logic is deterministic by the op's timestamp and newly archived tasks are never past-threshold), but **old time-tracking rows carried by the archived tasks** can land in young on receivers vs old on the originator. Self-heals at the next flush; until then archive reads that consult only one store may differ.

Options: split persistence from flushing so `moveToArchive` is dispatched causally before the flush action, or make the flush handler tolerate/converge TT placement. Low priority: bounded divergence, self-correcting.

Also from the review, fold in when touching this area: remove the now-inert `isIgnoreDBLock` threading through the archive services and handler (the TASK_ARCHIVE mutex is unconditional since 867d84a3d1).

johannesjo:
Scope addition from the adversarial fix review: three archive RMW paths remain OUTSIDE the TASK_ARCHIVE mutex (pre-existing — they were unlocked before 867d84a3d1 too, which is why the lock comment in task-archive.service.ts now names them):

1. `ArchiveCompressionService.compressArchive` (load both → compress → `saveArchivesAtomic`, runs for local AND remote ops; no live local dispatcher was found, so it mainly replays old remote history).
2. `_handleLoadAllData` remote SYNC_IMPORT archive import (deliberately unlocked across its user-confirmation guard — a held cross-tab lock across a dialog would starve waiters; needs a design decision, e.g. lock only the write phase after the guard resolves).
3. `TimeTrackingService.cleanupDataEverywhereForProject` / `cleanupArchiveDataForTag` — three unlocked load→modify→save cycles over both archives, invoked from the delete-project/delete-tags handlers right after a locked TaskArchiveService call, so the race window is real.

When removing the inert `isIgnoreDBLock` threading (this issue), extend the lock to (1) and (3), and decide (2).

PR #9005, opened 2026-07-14
fix(sync): file-based snapshot crash/corruption hardening (follow-up to #8960)

johannesjo:
**Stacked on #9004** (the focused #8960 four-defect fix). Review #9004 first; this PR's base will retarget to master once #9004 merges.

This is the **orthogonal crash/corruption-hardening layer** that accumulated on the original `feat/issue-8960-5a4217` working branch. It is **not required to fix any of the four #8960 defects** — those are all in #9004. It's split out here so it can be evaluated on its own merits (or landed incrementally / deferred) rather than riding along under a bug fix.

## What's in here

- **Snapshot two-phase-commit subsystem** — wraps every *authoritative* snapshot force-write ("Use Local", E2EE re-encryption, force-upload, split-snapshot) in a durable pending→complete marker so a crash mid-write can be resumed/aborted and a later ordinary upload can't misread non-monotonic snapshot clocks. (~1,000 lines; the largest piece, and the one still settling — the recovery-call ordering was reworked twice during development.)
- **Structural-validation layer** — rejects parses-as-JSON-but-wrong-shape remote/`.bak` files (envelope / vector-clock / op type-guards).
- **Concurrent-hydration handling** — buffers local actions dispatched *during* async snapshot hydration and replays them onto the new baseline (`_hydrateSnapshotExclusive`, atomic `commitFileSnapshotBaseline`).
- **Split-migration crash-resume scaffolding**, `.bak` self-heal, backup-digest tracking, persisted unresolved-gap tracking.

## Scope notes / caveats

- **Not blocking for #8960.** These protect failure modes orthogonal to the four filed defects.
- **Excludes the #8941 archive serialization.** That work (`TASK_ARCHIVE` locks around archive read-modify-write) is entangled with newer `master` and belongs to #8941 — it will be rebased fresh on master as its own PR. One harmless remnant remains here: `cfac`'s hydration-side `TASK_ARCHIVE` lock (a no-op until the #8941 writer locks land) is kept, because separating it from `cfac`'s legitimate hydration-ordering change would need hunk surgery.
- Faithful reconstruction of the original branch's hardening commits, re-based onto current master (verified: adapter / hydration / sync-service byte-identical to the original branch; one cosmetic `no-shadow` merge artifact fixed).

## Verification

- ✅ eslint clean on all changed production files.
- ⚠️ Full unit (Karma) + SuperSync e2e suites not yet run locally (sandbox constraints) — relying on CI. Given the size and that this touches the highest-risk subsystem, wants a careful review + green CI before merge.

### DIFF
--- a/docs/sync-and-op-log/README.md
+++ b/docs/sync-and-op-log/README.md
@@ -19,7 +19,8 @@ vector clocks detect concurrent edits.
           │
           └──► Sync Providers
                ├── SuperSync   (operation-based, real-time)
-               └── WebDAV / Dropbox / LocalFile  (file-based, single sync-data.json)
+               └── WebDAV / Dropbox / OneDrive / LocalFile
+                   (file-based: v2 single file or opt-in v3 split files)
 ```
 
 ## Start here
--- a/docs/sync-and-op-log/diagrams/04-file-based-sync.md
+++ b/docs/sync-and-op-log/diagrams/04-file-based-sync.md
@@ -126,37 +126,72 @@ File snapshot bootstrap does not create a `SYNC_IMPORT`; it retains ordinary
 causality. Explicit Use Local/Use Remote and backup-import flows still use their
 full-state operation semantics.
 
+The downloaded state cache, vector clock, archives, and operations already
+represented by the snapshot commit in one IndexedDB transaction. A failed write
+therefore leaves the previous baseline intact. Buffered local intents are made
+durable against whichever complete baseline committed before their overwritten
+reducers are replayed; archive effects are then restored in sequence, including
+any additional local intents that arrive during restoration.
+
 ## Split Compaction and Recovery
 
-For v3 compaction, `sync-state.json` is backed up and conditionally written
-before `sync-ops.json` is changed to reference it. If the process stops between
-the two writes, readers keep using the previous ops commit point and recover the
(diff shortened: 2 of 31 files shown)


johannesjo:
Closing in favor of #9010.

A reachability audit of this branch against `fix/8960-core` found only one part defends a **reachable** failure mode: the concurrent-hydration fix, where a local edit made during an async snapshot hydrate is silently clobbered by `loadAllData` (or dropped by `markRejected`) on the default single-file path. That slice is extracted, self-contained, and building in **#9010**.

The rest is deliberately deferred:

- **Snapshot two-phase-commit marker** (~1k lines + two new permanent on-disk formats): the base is already fail-safe for interrupted authoritative snapshot writes (`.bak` recovery + existing clock logic), so this adds durability/resume without closing a realistically-hit hole — not worth two forever-formats right now.
- **Structural validation**: defends a valid-JSON-but-wrong-shape input no shipping client produces; worst realistic case is a retryable sync stall, not data loss.
- **Split-migration crash-resume**: only runs under Surgical/split sync (default-off).

The `feat/sync-recovery-hardening` branch is preserved as the backlog for that hardening — worth revisiting per-failure-mode (each with a reproducing test) if any of these turns out reachable. Not deleting it.

PR #8991, opened 2026-07-14
fix(locale): localize ISO weekday labels

johannesjo:
## Problem

The ISO 8601 datetime option uses Swedish as a stable formatting sentinel for `YYYY-MM-DD` and 24-hour time. Weekday labels in Schedule, Habits, and Planner were therefore rendered in Swedish even when the UI language was different.

Fixes #8987.

## Solution

- Use the UI language for weekday text when the ISO 8601 datetime option is selected.
- Keep numeric ISO date and time formatting unchanged.
- Preserve the existing locale formatter for every non-ISO datetime option.
- Add regression tests for ISO labels, UI-language changes, and non-ISO formatting parity.
- Clarify the localization setting in the wiki and English UI copy.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki).
- [x] I have run `npm run checkFile` on changed `.ts`/`.scss` files
- [x] I have added tests for my changes (if applicable)
- [x] Existing tests still pass
- [x] My commit messages follow the Angular format (`type(scope): description`)

## Validation

- `npm test`
- `npm run build`
- Targeted DateTimeFormatService, Schedule, Habits, and Planner unit tests

PR #9013, opened 2026-07-14
fix(locale): localize ISO 8601 calendar weekday/month names (#8987)

### LINKED ISSUE #8987: ISO 8601 date-format option renders Swedish weekday names (Schedule / Habits / Planner)
### Version Used

v18.14.0 — Linux, installed via AUR (superproductivity-bin)

### Current Behavior

With **ISO 8601: 24-hour, YYYY-MM-DD** selected as the Datetime format locale (Menu → Settings → General tab → expand Localization → Datetime format locale and I pick ISO 8601: 24-hour, YYYY-MM-DD), weekday labels render in **Swedish**  (`mån tis ons tors fre lör sön`) instead of English — in the **Schedule**, **Habits**, and **Planner** views.

The dates themselves are correct (`YYYY-MM-DD`, 24-hour — e.g. a task's "created" timestamp); only the weekday *names* are localized to Swedish. Setting the app **Language** to English does not change it (weekday names follow the datetime-format locale, not the UI language).

Confirmed in **v18.14.0** (introduced in v18.13.1).

Likely cause (from #8631): the ISO 8601 option maps to the Swedish locale (`DateTimeLocales.sv` in `language-selection-form.const.ts`), reused for sync-safety per the code comment — so `sv`'s Swedish weekday names surface in the date views (Schedule, Habits, Planner).

### Expected Behavior

Weekday labels should appear in the app's UI language (English: `Mon Tue Wed Thu Fri Sat Sun`), while keeping the ISO `YYYY-MM-DD` date format and 24-hour time.

### Steps to Reproduce

1. Go to Settings → General → Localization → **Datetime format locale**.
2. Select **ISO 8601: 24-hour, YYYY-MM-DD**.
3. Open the **Schedule** view (also reproduces in **Habits** and **Planner** views).
4. Observe the weekday headers/labels are Swedish (`mån tis ons …`) instead of English.

### Console Output

```shell

```

### Documentation

johannesjo:
## Problem

Follow-up to #8991 (fixes #8987). That PR localized ISO 8601 weekday labels only in the **custom** Schedule/Habits/Planner components. The Material `<mat-calendar>` — used by the schedule-task dialog (`S`), deadline dialog, date-picker inputs, `dialog-select-date-time`, and the repeat-task heatmap — still rendered its month header (e.g. `JULI`) and weekday row (`M T O T F L S`) in Swedish and ignored the app language.

**Root cause:** the ISO 8601 option stores `dateTimeLocale = 'sv'` (the sentinel that yields `YYYY-MM-DD` + 24h). `DateTimeFormatService` pushes that onto the single global `DateAdapter` via `setLocale('sv')`. `<mat-calendar>` reads its text from `NativeDateAdapter.getMonthNames` / `getDayOfWeekNames` / `format(monthYearLabel)`, all keyed on that `sv` locale, so every calendar app-wide showed Swedish names.

Reported by @cdnia on #8991.

## Solution

Override the spelled-out-name paths in `CustomDateAdapter` to run under a temporary swap to `isoTextLocale()` (the UI language) when the ISO option is active:

- `getDayOfWeekNames`, `getMonthNames`, and the spelled-out branch of `format()` use the UI language.
- Numeric `dateInput` (already special-cased) and time-only formats keep the configured locale — **ISO stays `YYYY-MM-DD` and the 24h clock is preserved** (this is why the global adapter locale is not simply repointed, which would flip `mat-timepicker` to 12h AM/PM under an English UI).
- The swap assigns `this.locale` directly (not `setLocale`) so it fires no spurious `localeChanges`.

This is a single-file fix at the one shared adapter, so it fixes every calendar and the heatmap at once. It stays reactive: on a language change the service effect re-applies the adapter locale, whose `localeChanges` makes `MatMonthView` re-read the names.

## Type of Change

- [x] Bug fix

## Validation

- `custom-date-adapter.spec.ts` (9) — names/header under the text locale, locale restoration, numeric + 24h passthrough unchanged.
- `custom-date-adapter.calendar.spec.ts` (3) — mounts a real `<mat-calendar>`; asserts UI-language header (not Swedish), a live `Monday → Montag` language switch, and the non-ISO fallback.
- Full `date-time-format` suite (36) + `datetime-picker` (22) green; `npm run checkFile` clean.

## Notes

- Day-cell numbers (`getDateNames`) and the year name are intentionally left on the configured locale (only affects digit shaping for `ar`/`fa`) — happy to localize if preferred.
- This branch also contains a few unrelated `docs(...)` commits that are ahead of `master`; the fix itself is commit `54ec07f3b7`.

### DIFF
--- a/docs/plans/2026-07-14-ios-testflight-master-builds.md
+++ b/docs/plans/2026-07-14-ios-testflight-master-builds.md
@@ -1,0 +1,206 @@
+# Implementation plan: Internal TestFlight builds from `master`
+
+**Status:** Plan only · **Date:** 2026-07-14
+**Difficulty:** Moderate. Allow about one engineering day split across two small
+changes, plus Apple processing and tester verification time.
+
+## Outcome and scope
+
+Use the existing iOS signing/archive workflow to send the newest eligible `master`
+state to an internal TestFlight group. Rapid pushes are coalesced: one beta may run
+and only the newest superseding beta waits. This intentionally does not produce one
+TestFlight build for every intermediate SHA.
+
+Production behavior stays event-based: final `v*` tag pushes submit the existing iOS
+release for review, prerelease tag pushes remain upload-only, and manual dispatch can
+never submit to production. Master builds are marked **TestFlight Internal Only**, so
+Apple cannot offer them to external testers or customers.
+
+Out of scope: external TestFlight, app changes, new dependencies, and the existing
+iOS/macOS App Store review-submission race. There is no new app-wide upload queue;
+the internal-only beta path neither edits App Store metadata nor opens a review.
+
+## Routing contract
+
+| Event                                  | Export                   | Fastlane lane | Production submit |
+| -------------------------------------- | ------------------------ | ------------- | ----------------- |
+| Push to `master`                       | TestFlight Internal Only | `ios beta`    | Never             |
+| Push of final `v*` tag                 | App Store Connect        | `ios release` | Yes               |
+| Push of prerelease `v*` tag            | App Store Connect        | `ios release` | No                |
+| Manual dispatch, default `beta` mode   | TestFlight Internal Only | `ios beta`    | Never             |
+| Manual dispatch, `release-upload-only` | App Store Connect        | `ios release` | Never             |
+
+The production guard must include the event type, not only the ref:
+
+```yaml
+SUBMIT_FOR_REVIEW: >-
+  ${{ github.event_name == 'push'
+      && startsWith(github.ref, 'refs/tags/v')
+      && !contains(github.ref, '-') }}
+```
+
+Manual beta dispatch is allowed only from `master`. The explicit
+`release-upload-only` mode preserves the workflow's current manual upload capability
+without inheriting its final-tag submission footgun.
+
+## Phase 0: Apple and version prerequisites
+
+- Create one internal tester group, add the intended App Store Connect users, and
+  enable automatic distribution. Accept that eligible tag uploads will also reach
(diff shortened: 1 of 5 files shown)


cdnia:
Hi, I tested #9013 — the calendar popup is fixed (English month/weekday names, follows the UI language). 

I found one more spot the Swedish still leaks, though: the **"Start date" value in the "Add Recurring Task Config" dialog** (on a task detail  → Recur → `+`). It renders the date in **Swedish** (e.g. `ons 15 juli 2026`) and doesn't follow the UI language — the *label* translates but the *value* doesn't. With the app in French I get `Date de début: tis 21 juli 2026` (French label, Swedish value). Picking a date in the now-English calendar and confirming still shows the Swedish string (e.g. `man 20 juli 2026`).

Steps:
1. on a task detail → Recur → `+` → "Add Recurring Task Config".
2. "Start date: `ons 15 juli 2026`" (Swedish).
3. Pick any date in the (now-English) calendar → confirm → Start date value is still in Swedish.
4. Switching Language in setting changes the label, not the value like what happened in for example in French I get `Date de début: tis 21 juli 2026` (French label, Swedish value)

Looks like this dialog formats the start date via a separate `sv`-locale path that the
#9013 calendar fix doesn't cover — same `sv`-sentinel root cause as #8987/#8991.

I attached a screenshot of the "Add Recurring Task Config" dialog** :

superproductivity_054_add_recurring_task_swedish_name_error

johannesjo:
Thank you @cdnia ! I look into it.

PR #9014, opened 2026-07-14
feat(add-task-bar): file tasks into project sections via short syntax (+Project/Section and /Section)

caseyh:
## Problem

Sections exist as a first-class way to organize a project's task list, but the quick-add bar can't target them — a task added via `+Project` always lands outside any section and has to be dragged in afterwards. For section-heavy projects this makes the add-task bar a two-step flow.

## Solution

Extends the `+` short syntax so quick-add can file a task directly into a section, plus autocomplete:

- `Task +Work/Design` — matches project *Work*, then prefix-matches *Design* against that project's sections (longest leading word span wins, so multi-word section names strip cleanly)
- `Task /Design` — a standalone `/Section` resolves against the project the task will land in (active work context or the bar's selected project); an explicit `+Project/Section` wins
- Typing `/` opens a mention dropdown with the effective project's sections (same UX as the existing `+`/`#`/`@` triggers)
- Placement reuses the existing `addTaskToSection` action right after creation; the parsed section lives in `AddTaskBarState.sectionId` and is cleared on project change and after each add

**Design notes:**

- `/` follows Todoist's quick-add convention (`#Project /Section`), so it matches arriving muscle memory
- Sections ride the existing project short-syntax toggle — no new setting, per the manifesto's "one calm default" guidance; independent gating would be a trivial follow-up if workflows diverge
- **Backward compatibility:** projects whose titles contain `/` (e.g. "A/B Testing") keep matching as plain projects — the whole token is tried first, covered by a test. Standalone `/` requires a word boundary with no trailing space, and nothing is stripped from the title unless a section actually matches, so `either/or`, `w/ milk`, `1h/2h` estimates, and URLs stay untouched (all covered by tests). With no sections passed, `shortSyntax()` output is unchanged.

**One point I want to flag for review** rather than have you find: creation into a section is two ops (task creation, then `addTaskToSection`) from one user gesture. Both replay deterministically and it mirrors how drag-into-section works today, but if you'd rather have a combined action/meta-reducer per the contributor sync model, I'm happy to rework it that way.

**To try it:** create a project with a couple of sections → quick-add → `something +ProjectName/` (dropdown appears) or, inside the project, just `something /sec`.

**Scope note:** this PR now covers the add-task bar only. Section syntax while editing existing task titles (+ inline autocomplete) was split into a follow-up PR per review.

## Type of Change

- [x] New feature

## Checklist

- [x] I have included relevant changes to the documentation/wiki (`3.04-Short-Syntax.md`)
- [x] I have added tests for my changes (26 new specs: matching, fallbacks, collision/compat cases, round-3 review regressions)
- [x] Existing tests still pass (232 short-syntax / 95 parser service / full 13.6k-spec suite green locally)
- [x] My commit messages follow the Angular format

### DIFF
--- a/docs/wiki/3.04-Short-Syntax.md
+++ b/docs/wiki/3.04-Short-Syntax.md
@@ -1,6 +1,6 @@
 # Short Syntax
 
-Super Productivity supports five main short syntax forms for quickly adding metadata to tasks: tags (`#`), projects (`+`), time estimates (`m`/`h`), due dates (`@`), and deadlines (`!`). The due date form also accepts recurrence phrases (`@every friday`, `@daily`) in the Add Task Bar. The forms are controlled in [[3.02-Settings-and-Preferences#global-settings.Short-Syntax]], with a separate opt-in setting for deadline syntax.
+Super Productivity supports six main short syntax forms for quickly adding metadata to tasks: tags (`#`), projects (`+`), project sections (`/`), time estimates (`m`/`h`), due dates (`@`), and deadlines (`!`). The due date form also accepts recurrence phrases (`@every friday`, `@daily`) in the Add Task Bar. The forms are controlled in [[3.02-Settings-and-Preferences#global-settings.Short-Syntax]], with a separate opt-in setting for deadline syntax.
 
 ## Supported Syntax Forms
 
@@ -21,6 +21,17 @@ Super Productivity supports five main short syntax forms for quickly adding meta
 - **Parameters:** Project title or partial title; minimum 3 characters. Matching is prefix-based, case-insensitive, with spaces removed for comparison. The shortest matching project title is preferred when several match.
 - **Valid:** `"Task +Important Project"`, `"Task +Project"` (partial match). Project names containing `+`, `#`, `@`, or similar are supported; the parser uses lookahead so these characters inside the name do not end the token.
 - **Invalid:** `"Task+Project"` (no space before `+`) — no project is applied.
+- **With a section:** `+<project-name>/<section-name>` files the task into a section of the matched project (see Section Syntax below). A project whose title itself contains `/` (e.g. `A/B Testing`) is always tried first and keeps matching as a plain project — which also means a `/` in a project title disables section syntax for that project.
+
+### Section Syntax (`/`)
+
+- **Symbol:** `/`
+- **Purpose:** Place the task inside a section of a project.
+- **Grammar:** either `+<project-name>/<section-name>` (explicit project) or a standalone `/<section-name>`, which resolves against the project the task will land in (the active work context or the project selected in the Add Task Bar). An explicit `+Project/Section` wins over the context. For the standalone form, `/` must start a word (space or start of title before it) and must not be followed by a space.
+- **Parameters:** Section title or partial title. Matching is prefix-based, case-insensitive, with spaces removed for comparison; when trailing words follow, the longest leading word span that still matches a section is used (so `"/Design Reviews finish draft"` matches `Design Reviews` and keeps `finish draft` in the title). Only sections belonging to the resolved project are considered.
+- **Valid:** `"Task +Work/Design"`, `"Task +Work/des"` (prefix), `"Task /Design"` (section of the current project), `"Task /des fix login"` (prefix with trailing text; the unmatched words stay in the title).
+- **Invalid / inert:** `"either/or"`, `"w/ milk"` (no word boundary before `/` or no match) — nothing is stripped from the title unless a section actually matches. `"Task / Design"` (space after `/`) is not parsed. `"+/..."` (nothing before the slash) is plain text, so notation like `+/-` stays untouched. Time forms like `1h/2h` are parsed as time before section parsing runs.
+- **Enabled by:** the project Short Syntax setting (there is no separate section toggle).
 
 ### Time Estimate Syntax
 
@@ -61,7 +72,7 @@ Super Productivity supports five main short syntax forms for quickly adding meta
 
 ## Formal Grammar and Regex Patterns
 
-The parser uses regular expressions for each syntax form. Time: optional `t` prefix, digits and decimal with `m`/`h`, optional `/` for spent/estimate. Project: `+` with lookahead so delimiters (`#`, `@`, time-like tokens) inside the project name do not terminate the match. Tag: `#` followed by characters until a special character or whitespace. Due: `@` followed by characters until a special character. Deadline: `!` followed by characters until a special character. Exact patterns are defined in `short-syntax.ts` (e.g. `SHORT_SYNTAX_TIME_REG_EX`, `SHORT_SYNTAX_PROJECT_REG_EX`, `SHORT_SYNTAX_TAGS_REG_EX`, `SHORT_SYNTAX_DUE_REG_EX`, `SHORT_SYNTAX_DEADLINE_REG_EX`).
+The parser uses regular expressions for each syntax form. Time: optional `t` prefix, digits and decimal with `m`/`h`, optional `/` for spent/estimate. Project: `+` with lookahead so delimiters (`#`, `@`, time-like tokens) inside the project name do not terminate the match. Tag: `#` followed by characters until a special character or whitespace. Due: `@` followed by characters until a special character. Deadline: `!` followed by characters until a special character. Section (standalone): `/` at a word boundary, not followed by whitespace or another `/`. Exact patterns are defined in `short-syntax.ts` (e.g. `SHORT_SYNTAX_TIME_REG_EX`, `SHORT_SYNTAX_PROJECT_REG_EX`, `SHORT_SYNTAX_STANDALONE_SECTION_REG_EX`, `SHORT_SYNTAX_TAGS_REG_EX`, `SHORT_SYNTAX_DUE_REG_EX`, `SHORT_SYNTAX_DEADLINE_REG_EX`).
 
 ## Default Values and Behavior
 
@@ -88,10 +99,10 @@ The parser uses regular expressions for each syntax form. Time: optional `t` pre
 
 ## Parsing and Evaluation Rules
 
-1. **Order of processing:** Due/deadline/time → Project → Tag.
+1. **Order of processing:** Due/deadline/time → Project (including sections) → Tag.
 2. **Title cleanup:** Parsed syntax is removed from the task title after processing.
 3. **Tag creation:** Tags that do not exist yet are collected in a `newTagTitles` array for creation.
-4. **Return value:** The `shortSyntax()` function returns an object with task changes (and `newTagTitles`, `remindAt`, `projectId`, `repeatQuickSetting`) or `undefined` if nothing was parsed or syntax is disabled.
+4. **Return value:** The `shortSyntax()` function returns an object with task changes (and `newTagTitles`, `remindAt`, `projectId`, `sectionId`, `repeatQuickSetting`) or `undefined` if nothing was parsed or syntax is disabled.
 
 ## Error Handling
 
@@ -110,7 +121,7 @@ The parser uses regular expressions for each syntax form. Time: optional `t` pre
 ## Add Task Bar Behavior
 
 - **Toggle new task position:** A toggle in the Add Task Bar controls whether the new task is added at the top or the bottom of the list.
-- **Autocomplete:** The Add Task Bar can suggest tags, projects, and date expressions while typing.
+- **Autocomplete:** The Add Task Bar can suggest tags, projects, date expressions, and — after `/` — the sections of the effective project while typing.
 - **Inline highlighting:** Detected short syntax (tags, project, estimate, due/deadline dates, recurrence, extracted URLs) is highlighted directly in the input while typing, showing which characters will be stripped from the title before the task is created.
 
 ## Notes
--- a/docs/wiki/3.04-Short-Syntax.md
+++ b/docs/wiki/3.04-Short-Syntax.md
@@ -31,7 +31,8 @@ Super Productivity supports six main short syntax forms for quickly adding metad
 - **Parameters:** Section title or partial title. Matching is prefix-based, case-insensitive, with spaces removed for comparison; when trailing words follow, the longest leading word span that still matches a section is used (so `"/Design Reviews finish draft"` matches `Design Reviews` and keeps `finish draft` in the title). Only sections belonging to the resolved project are considered.
 - **Valid:** `"Task +Work/Design"`, `"Task +Work/des"` (prefix), `"Task /Design"` (section of the current project), `"Task /des fix login"` (prefix with trailing text; the unmatched words stay in the title).
 - **Invalid / inert:** `"either/or"`, `"w/ milk"` (no word boundary before `/` or no match) — nothing is stripped from the title unless a section actually matches. `"Task / Design"` (space after `/`) is not parsed. `"+/..."` (nothing before the slash) is plain text, so notation like `+/-` stays untouched. Time forms like `1h/2h` are parsed as time before section parsing runs.
-- **Enabled by:** the project Short Syntax setting (there is no separate section toggle).
+- **Enabled by:** the project Short Syntax setting (there is no separate section toggle). The `/` autocomplete dropdown follows the same setting.
+- **Backlog adds:** when the Add Task Bar is set to add into the project backlog, a parsed section is ignored — sections only group the main task list, so a backlogged task never carries an invisible section membership.
 
 ### Time Estimate Syntax
 
--- a/docs/wiki/3.04-Short-Syntax.md
+++ b/docs/wiki/3.04-Short-Syntax.md
@@ -1,6 +1,6 @@
 # Short Syntax
 
-Super Productivity supports five main short syntax forms for quickly adding metadata to tasks: tags (`#`), projects (`+`), time estimates (`m`/`h`), due dates (`@`), and deadlines (`!`). The due date form also accepts recurrence phrases (`@every friday`, `@daily`) in the Add Task Bar. The forms are controlled in [[3.02-Settings-and-Preferences#short-syntax]], with a separate opt-in setting for deadline syntax.
+Super Productivity supports six main short syntax forms for quickly adding metadata to tasks: tags (`#`), projects (`+`), project sections (`/`), time estimates (`m`/`h`), due dates (`@`), and deadlines (`!`). The due date form also accepts recurrence phrases (`@every friday`, `@daily`) in the Add Task Bar. The forms are controlled in [[3.02-Settings-and-Preferences#short-syntax]], with a separate opt-in setting for deadline syntax.
 
 ## Supported Syntax Forms
 
@@ -21,6 +21,17 @@ Super Productivity supports five main short syntax forms for quickly adding meta
 - **Parameters:** Project title or partial title; minimum 3 characters. Matching is prefix-based, case-insensitive, with spaces removed for comparison. The shortest matching project title is preferred when several match.
 - **Valid:** `"Task +Important Project"`, `"Task +Project"` (partial match). Project names containing `+`, `#`, `@`, or similar are supported; the parser uses lookahead so these characters inside the name do not end the token.
 - **Invalid:** `"Task+Project"` (no space before `+`) — no project is applied.
+- **With a section:** `+<project-name>/<section-name>` files the task into a section of the matched project (see Section Syntax below). A project whose title itself contains `/` (e.g. `A/B Testing`) is always tried first and keeps matching as a plain project — which also means a `/` in a project title disables section syntax for that project.
+
+### Section Syntax (`/`)
+
+- **Symbol:** `/`
+- **Purpose:** Place the task inside a section of a project.
+- **Grammar:** either `+<project-name>/<section-name>` (explicit project) or a standalone `/<section-name>`, which resolves against the project the task will land in (the active work context or the project selected in the Add Task Bar). An explicit `+Project/Section` wins over the context. For the standalone form, `/` must start a word (space or start of title before it) and must not be followed by a space.
+- **Parameters:** Section title or partial title. Matching is prefix-based, case-insensitive, with spaces removed for comparison; when trailing words follow, the longest leading word span that still matches a section is used (so `"/Design Reviews finish draft"` matches `Design Reviews` and keeps `finish draft` in the title). Only sections belonging to the resolved project are considered.
+- **Valid:** `"Task +Work/Design"`, `"Task +Work/des"` (prefix), `"Task /Design"` (section of the current project), `"Task /des fix login"` (prefix with trailing text; the unmatched words stay in the title).
+- **Invalid / inert:** `"either/or"`, `"w/ milk"` (no word boundary before `/` or no match) — nothing is stripped from the title unless a section actually matches. `"Task / Design"` (space after `/`) is not parsed. `"+/..."` (nothing before the slash) is plain text, so notation like `+/-` stays untouched. Time forms like `1h/2h` are parsed as time before section parsing runs.
+- **Enabled by:** the project Short Syntax setting (there is no separate section toggle).
 
 ### Time Estimate Syntax
 
@@ -62,7 +73,7 @@ Super Productivity supports five main short syntax forms for quickly adding meta
 
 ## Formal Grammar and Regex Patterns
 
-The parser uses regular expressions for each syntax form. Time: optional `t` prefix, digits and decimal with `m`/`h`, optional `/` for spent/estimate. Project: `+` with lookahead so delimiters (`#`, `@`, time-like tokens) inside the project name do not terminate the match. Tag: `#` followed by characters until a special character or whitespace. Due: `@` followed by characters until a special character. Deadline: `!` followed by characters until a special character. Exact patterns are defined in `short-syntax.ts` (e.g. `SHORT_SYNTAX_TIME_REG_EX`, `SHORT_SYNTAX_PROJECT_REG_EX`, `SHORT_SYNTAX_TAGS_REG_EX`, `SHORT_SYNTAX_DUE_REG_EX`, `SHORT_SYNTAX_DEADLINE_REG_EX`).
+The parser uses regular expressions for each syntax form. Time: optional `t` prefix, digits and decimal with `m`/`h`, optional `/` for spent/estimate. Project: `+` with lookahead so delimiters (`#`, `@`, time-like tokens) inside the project name do not terminate the match. Tag: `#` followed by characters until a special character or whitespace. Due: `@` followed by characters until a special character. Deadline: `!` followed by characters until a special character. Section (standalone): `/` at a word boundary, not followed by whitespace or another `/`. Exact patterns are defined in `short-syntax.ts` (e.g. `SHORT_SYNTAX_TIME_REG_EX`, `SHORT_SYNTAX_PROJECT_REG_EX`, `SHORT_SYNTAX_STANDALONE_SECTION_REG_EX`, `SHORT_SYNTAX_TAGS_REG_EX`, `SHORT_SYNTAX_DUE_REG_EX`, `SHORT_SYNTAX_DEADLINE_REG_EX`).
 
 ## Default Values and Behavior
 
@@ -89,10 +100,10 @@ The parser uses regular expressions for each syntax form. Time: optional `t` pre
 
 ## Parsing and Evaluation Rules
 
-1. **Order of processing:** Due/deadline/time → Project → Tag.
+1. **Order of processing:** Due/deadline/time → Project (including sections) → Tag.
 2. **Title cleanup:** Parsed syntax is removed from the task title after processing.
 3. **Tag creation:** Tags that do not exist yet are collected in a `newTagTitles` array for creation.
-4. **Return value:** The `shortSyntax()` function returns an object with task changes (and `newTagTitles`, `remindAt`, `projectId`, `repeat`) or `undefined` if nothing was parsed or syntax is disabled.
+4. **Return value:** The `shortSyntax()` function returns an object with task changes (and `newTagTitles`, `remindAt`, `projectId`, `sectionId`, `repeat`) or `undefined` if nothing was parsed or syntax is disabled.
 5. **Buttons win over text:** picking a date, deadline, estimate or recurrence in the Add Task Bar's buttons removes the matching syntax from the input, the same way the buttons' clear (`x`) does, so the text stops advertising a value the task will not get. Changing the date of a task whose recurrence came from syntax keeps the recurrence and re-anchors it to the new date.
 
 ## Error Handling
@@ -112,7 +123,7 @@ The parser uses regular expressions for each syntax form. Time: optional `t` pre
 ## Add Task Bar Behavior
 
 - **Toggle new task position:** A toggle in the Add Task Bar controls whether the new task is added at the top or the bottom of the list.
-- **Autocomplete:** The Add Task Bar can suggest tags, projects, and date expressions while typing.
+- **Autocomplete:** The Add Task Bar can suggest tags, projects, date expressions, and — after `/` — the sections of the effective project while typing.
 - **Inline highlighting:** Detected short syntax (tags, project, estimate, due/deadline dates, recurrence, extracted URLs) is highlighted directly in the input while typing, showing which characters will be stripped from the title before the task is created.
 
 ## Notes
--- a/src/app/features/tasks/add-task-bar/add-task-bar-parser.service.spec.ts
+++ b/src/app/features/tasks/add-task-bar/add-task-bar-parser.service.spec.ts
@@ -1,5 +1,5 @@
 import { TestBed } from '@angular/core/testing';
-import { provideMockStore } from '@ngrx/store/testing';
+import { MockStore, provideMockStore } from '@ngrx/store/testing';
 import { AddTaskBarParserService } from './add-task-bar-parser.service';
 import { AddTaskBarStateService } from './add-task-bar-state.service';
 import { selectAllSections } from '../../section/store/section.selectors';
@@ -1724,6 +1724,121 @@ describe('AddTaskBarParserService', () => {
       mockStateService.state.and.returnValue(baseState as any);
       await service.parseAndUpdateText('', cfg, [], [], defaultProject);
       expect(mockStateService.updateSyntaxHighlight).toHaveBeenCalledWith(null);
+    });
+  });
+
+  describe('section validation against the final project (round-3 review, PR #9014)', () => {
+    const cfg = {
+      isEnableProject: true,
+      isEnableDue: true,
+      isEnableTag: true,
+    } as ShortSyntaxConfig;
+    const defaultProject = { id: 'default-project', title: 'Default Project' } as Project;
+    const workProject = { id: 'work', title: 'Work' } as Project;
+    const designSection = {
+      id: 'sec-design',
+      contextId: 'work',
+      contextType: 'PROJECT',
+      title: 'Design',
+      taskIds: [],
+    };
+    const makeState = (projectId: string): unknown => ({
+      projectId,
+      sectionId: null,
+      tagIds: [],
+      tagIdsFromTxt: [],
+      newTagTitles: [],
+      date: null,
+      time: null,
+      spent: null,
+      estimate: null,
+      cleanText: null,
+      remindOption: null,
+      attachments: [],
+      repeatQuickSetting: null,
+      deadlineDate: null,
+      deadlineTime: null,
+      deadlineRemindOption: null,
+    });
+
+    // Mirror the real state service closely enough for the two-step
+    // sequence: project updates clear the section, every update lands in
+    // `liveState` so the service's own `state()` reads see it.
+    let liveState: { projectId: string; sectionId: string | null };
+    beforeEach(() => {
+      const store = TestBed.inject(MockStore);
+      store.overrideSelector(selectAllSections, [designSection] as never[]);
+      liveState = makeState(defaultProject.id) as typeof liveState;
+      mockStateService.state.and.callFake((() => liveState) as any);
+      mockStateService.isAutoDetected.and.returnValue(true);
+      mockStateService.setAutoDetectedProjectId.and.callFake((id: string) => {
+        liveState = { ...liveState, projectId: id, sectionId: null };
+      });
+      mockStateService.updateProjectId.and.callFake((id: string) => {
+        liveState = { ...liveState, projectId: id, sectionId: null };
+      });
+      mockStateService.updateSectionId.and.callFake((id: string | null) => {
+        liveState = { ...liveState, sectionId: id };
+      });
+    });
+
+    it('should resolve "+Work/Design" to Work + its section', async () => {
+      await service.parseAndUpdateText(
+        'buy milk +Work/Design',
+        cfg,
+        [defaultProject, workProject],
+        [],
+        defaultProject,
+      );
+      expect(liveState.projectId).toBe('work');
+      expect(liveState.sectionId).toBe('sec-design');
+    });
+
+    it('should clear a stale section from another project when the +Project token is deleted', async () => {
+      await service.parseAndUpdateText(
+        'buy milk +Work/Design',
+        cfg,
+        [defaultProject, workProject],
+        [],
+        defaultProject,
+      );
+
+      // Deleting just "+Work" falls the task back to the default project;
+      // the leftover "/Design" must NOT keep pointing at Work's section.
+      await service.parseAndUpdateText(
+        'buy milk /Design',
+        cfg,
+        [defaultProject, workProject],
+        [],
+        defaultProject,
+      );
+      expect(liveState.projectId).toBe(defaultProject.id);
+      expect(liveState.sectionId).toBeNull();
+    });
+
+    it('should clear the section when the user manually picks another project', async () => {
+      await service.parseAndUpdateText(
+        'buy milk +Work/Design',
+        cfg,
+        [defaultProject, workProject],
+        [],
+        defaultProject,
+      );
+
+      // Manual selection: the state service clears the section and the
+      // auto-detected flag; a re-parse of the unchanged text must not
+      // re-apply Work's section to the manually chosen project.
+      liveState = { ...liveState, projectId: defaultProject.id, sectionId: null };
+      mockStateService.isAutoDetected.and.returnValue(false);
+      await service.parseAndUpdateText(
+        'buy milk +Work/Design',
+        cfg,
+        [defaultProject, workProject],
+        [],
+        defaultProject,
+      );
+      expect(liveState.projectId).toBe(defaultProject.id);
+      expect(liveState.sectionId).toBeNull();
     });
   });
 
--- a/src/app/features/tasks/add-task-bar/add-task-bar-parser.service.spec.ts
+++ b/src/app/features/tasks/add-task-bar/add-task-bar-parser.service.spec.ts
@@ -1,6 +1,8 @@
 import { TestBed } from '@angular/core/testing';
+import { provideMockStore } from '@ngrx/store/testing';
 import { AddTaskBarParserService } from './add-task-bar-parser.service';
 import { AddTaskBarStateService } from './add-task-bar-state.service';
+import { selectAllSections } from '../../section/store/section.selectors';
 import { ShortSyntaxConfig } from '../../config/global-config.model';
 import { Project } from '../../project/project.model';
 import { Tag } from '../../tag/tag.model';
@@ -26,6 +28,7 @@ describe('AddTaskBarParserService', () => {
       'updateRepeatSetting',
       'clearRepeatSetting',
       'updateSyntaxHighlight',
+      'updateSectionId',
       'isAutoDetected',
       'state',
     ]);
@@ -52,6 +55,9 @@ describe('AddTaskBarParserService', () => {
       providers: [
         AddTaskBarParserService,
         { provide: AddTaskBarStateService, useValue: mockStateServiceSpy },
+        provideMockStore({
+          selectors: [{ selector: selectAllSections, value: [] }],
+        }),
       ],
     });
 
--- a/src/app/features/tasks/add-task-bar/add-task-bar-parser.service.spec.ts
+++ b/src/app/features/tasks/add-task-bar/add-task-bar-parser.service.spec.ts
@@ -150,6 +150,7 @@ describe('AddTaskBarParserService', () => {
         // Mock state to return no current date/time
         const mockState = {
           projectId: mockDefaultProject.id,
+          sectionId: null,
           tagIds: [],
           tagIdsFromTxt: [],
           newTagTitles: [],
@@ -188,6 +189,7 @@ describe('AddTaskBarParserService', () => {
         // Mock state to return current user-selected values
         const mockState = {
           projectId: mockDefaultProject.id,
+          sectionId: null,
           tagIds: [],
           tagIdsFromTxt: [],
           newTagTitles: [],
@@ -223,6 +225,7 @@ describe('AddTaskBarParserService', () => {
         // Mock state with date but no time
         const mockState = {
           projectId: mockDefaultProject.id,
+          sectionId: null,
           tagIds: [],
           tagIdsFromTxt: [],
           newTagTitles: [],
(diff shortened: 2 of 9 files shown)


johannesjo:
Thanks for the thorough write-up and tests — this reuses the existing building blocks (`addTaskToSection`, the mention-config toggle) nicely, and the backward-compat handling for `/` in prose (`either/or`, `w/ milk`, `1h/2h`, URLs) is careful and well-covered. One confirmed correctness bug plus a few edge/UX notes.

## 🔴 `+Project/<unmatched-section>` leaves a `/`-residue in the title

In `parseProjectChanges` → `buildResult` (`short-syntax.ts`): when `+Project/xxx` matches the **project** but `xxx` matches **no section**, only `+Project` is stripped, so the `/` and everything after it stay in the title. Reproduced by simulating the exact new logic:

```
"Task +Work/nonexistent"       => title "Task /nonexistent"        projectId WorkID
"+Work/review the Q3 numbers"  => title "/review the Q3 numbers"   projectId WorkID
"buy milk +Work/grocer"        => title "buy milk /grocer"         projectId WorkID
"Task +Work/Design"            => title "Task"  (correct)          + sectionId
```

This fires on any non-matching section part — a typo, a renamed/deleted section, or just typing `+Work/` and continuing the title — filing the task into the right project but with a mangled title.

Note the asymmetry with the standalone `/Section` form, which is correctly conservative ("nothing stripped unless a section matches"). The `+Project/` form strips eagerly and orphans the rest, because `buildResult` reconstructs the strip target by concatenation (`+${typed}` when no section) instead of accounting for the slash the regex captured.

**Suggested fix:** when `slashIndex !== -1` and no section matched, strip through the slash too (e.g. `+Work/`), so `"buy milk +Work/grocer"` → `"buy milk grocer"`.

This is **not covered by tests** — the `should match the project without a section when the section part matches nothing` spec asserts only `projectId`/`sectionId`, never the title. A title assertion there would have caught it; please add one.

## 🟡 "Add to backlog" silently discards the section

Sections group `undoneTasks()` only (`undoneTasksBySection` in `work-view.component.ts`) — the main list, not the backlog. With the add-bar's **add-to-backlog** toggle on, a `+Project/Section` task lands in `backlogTaskIds` *and* gets appended to `section.taskIds`, but renders in the backlog and never under the section (it would "reappear" in the section only if later moved out of backlog). Worth a guard or at least a doc note.

## 🟡 Two ops per gesture — agree it's acceptable

Re: your flagged point — `add()` + `addTaskToSection()` mirrors drag-into-section, both replay deterministically, and the reducer is defensive (`if (target)` no-ops on a concurrently-deleted section; a dangling `taskIds` ref is filtered out by `undoneTasksBySection`). I would **not** block on a meta-reducer here. A one-line comment at the call site pointing to the drag precedent would preempt the "why isn't this atomic like `removeTaskFromSection`?" question.

## 🔵 Minor
- **Section placement ignores the add-to-bottom toggle:** `addTaskToSection(id, taskId, null, null)` with `afterTaskId: null` prepends to the section (`moveItemAfterAnchor` puts null-anchor items at the top), regardless of `isAddToBottom()`.
- **`/` mention dropdown** triggers on a bare `/`, so `1h/2h`, `and/or`, or a pasted URL can pop the sections dropdown mid-type (parser still handles them correctly — purely cosmetic; same class as the existing `@`/`#` triggers).
- **Multiword project + first-word match + section:** `buildResult`'s concatenated strip target (`+Work/Design`) won't match a literal `+Work Space/Design` input, so the syntax isn't cleaned in that narrow case — same string-reconstruction fragility as the bug above; a positional strip would fix both.

Nice work overall — the state hygiene (clearing `sectionId` on project change and after add, with the *why* comments) and the docs update are solid. Fixing the title-residue bug with a regression test is the one thing I'd want before merge.

caseyh:
Thanks for the thorough review — all addressed in `8c0c913`:

- **🔴 Slash residue**: fixed with your suggested semantics — an unmatched section part now strips through the slash, so `"buy milk +Work/grocer"` → `"buy milk grocer"`. Implemented as a **positional strip** (slice by the match index + token span instead of `.replace()` on a reconstructed string), which also fixes the multi-word-project fragility you flagged in the minors — both had the same root cause. Sections now only resolve on a full left-side project match, so the strip is always contiguous. Regression tests assert **titles** now (that was exactly the gap), including your three repro cases verbatim.
- **🟡 Backlog**: section placement is skipped for backlog adds, with a comment explaining why (sections only group the main list).
- **🔵 Add-to-bottom**: respected — `afterTaskId` is the section's last task when the toggle is on.
- **🟡 Two ops**: call-site comment added citing the drag-into-section precedent, per your note.
- **🔵 `/` dropdown on bare slash**: left as-is since it's the same class as the existing `@`/`#` trigger behavior — happy to gate it further if you'd prefer.

**Heads-up on scope**: two commits landed between your review and this fix (`558e182`, `3f98212`) extending the syntax to **title edits** — there it rides the existing `applyShortSyntax` compound action (new optional `targetSectionId`, applied atomically in the meta-reducer), plus the `/` autocomplete in the inline title editor (mention logic hoisted into `MentionConfigService`, subscribed only while editing per the task-component hot-path rule). If you'd rather review that as a separate PR, say the word and I'll split it out.

johannesjo:
Round 2 — thanks, this is a good set of fixes. I re-ran your parser against my three repro cases by extracting the real `parseProjectChanges` source and executing it, and all four of my round-1 points are genuinely addressed:

```
"buy milk +Work/grocer"        => title "buy milk grocer"        proj Work
"+Work/review the Q3 numbers"  => title "review the Q3 numbers"  proj Work
"Task +Work/nonexistent"       => title "Task nonexistent"       proj Work
```

The positional strip is the right call — it fixes the residue and the multi-word fragility from one root cause, exactly as you said. And the specs now assert **titles**, which was precisely the gap. Nice.

On your scope question: **yes, please split** — and the answer falls out of the review below, because the two later commits are where the one blocking bug lives.

## 🔴 A sub-task can be filed into a section, violating the parent-only invariant

`section.taskIds` is parent-only, and the codebase guards this deliberately at every other write site:

- `convertToSubTask` → `handleTaskRemoval` strips the id from all sections
- the drag `enterPredicate` rejects sub-tasks into section lists — comment: *"section.taskIds is parent-only"* (`task-list.component.ts:275-289`)
- the LWW path strips on `becomesSubTask` (`section-shared.reducer.ts:499-515`)
- `can-convert-task-to-sub-task.ts:27-44`: *"Used by BOTH the section and crud meta-reducers so their guards stay in lock-step"*

The new title-edit path bypasses all of it. Traced end to end against the branch:

1. Sub-task rows render through `task.component` (`task.component.html` → `<task-list [tasks]="t.subTasks">` → `<task>`), so each one gets `<task-title [sectionProjectId]="t.projectId || null">` — unconditionally.
2. Sub-tasks inherit the parent's project (`task.reducer.ts:502`, `projectId: parentTask.projectId`), so `sectionProjectId` is **never null** — the new `/` dropdown actively **lists the project's sections while editing a sub-task title**.
3. `updateTaskTitleIfChanged` → `_taskService.update(id, { title })` → a title-only `updateTask`, which passes the effect's filter.
4. The parser resolves it (verified by running the real source):
   ```
   parseStandaloneSectionChanges({title: 'fix login /Design'}, <parent project>, sections)
     => title "fix login"   sectionId DES
   ```
5. `short-syntax.effects.ts` guards the **project** move with `!task.parentId` (line 205) but the **section** has no such guard (line 257):
   ```ts
   const targetSectionId =
     r.sectionId && r.sectionId !== currentSectionId ? r.sectionId : undefined;
   ```
6. `moveTaskToSection` in the meta-reducer has no `parentId` check either → the sub-task id lands in `section.taskIds`.

`+Other/Design` on a sub-task is the sharper version: the project move is correctly refused, but the section placement still goes through — filing it into a section of a project it isn't in. Same reachable via `addSubTask`, which the effect also listens to.

**Why it's worth fixing rather than tolerating:** `undoneTasksBySection` intersects by id, so it renders nowhere today — but `applyShortSyntax` is `isPersistent: true` / `opType: OpType.Update`, so the bad membership **replays to every device**. And it doesn't stay invisible: `convertToMainTask` promotes a sub-task to top-level, and `section-shared.reducer.ts` has **no handler for it** (it doesn't need one today, precisely because sub-tasks are never in sections). So promote that sub-task later and it silently materialises inside a section the user never put it in.

**Suggested fix:** add `!task.parentId` to the `targetSectionId` computation to mirror the project guard, and gate `[sectionProjectId]` on `!t.parentId` so the dropdown stops inviting it. The load-bearing one is a guard inside `moveTaskToSection` — it has `RootState`, and it's the surface that replays on remote devices; that's also where your existing guards keep each other in lock-step.

## 🔵 Minor

**`+A/B Testing extra words` regresses.** The whole-token-first compat guard only fires when the token has no trailing words; with trailing text it falls into the section path and strips `+A/` instead of `+A/B`. Ran both parsers against real source:

```
OLD (master):  "Task +A/B Testing extra words" => "Task Testing extra words"    proj A/B Testing
NEW (this PR): "Task +A/B Testing extra words" => "Task B Testing extra words"  proj A/B Testing
```

An orphaned `B` — same residue class as the round-1 bug. Your compat spec passes only because `"Task +A/B Testing"` has no trailing text; a case with trailing words would catch it.

**Hot-path cost in `task-title.component.ts`.** `toObservable(this.sectionProjectId)` creates an Angular `effect()` **eagerly at construction** (`rxjs-interop.mjs:80`), so a 500-row list creates 500 effects and ReplaySubjects whether or not anything is edited. The comment says *"the template only subscribes while actually editing"* — true of the async pipe, but the effect isn't gated by it. Cleanup is properly registered (`DestroyRef.onDestroy`), so there's no leak — it's just per-row cost in the component the repo singles out as a hot path. Worth a look against a long list. Also `// short-syntax autocomplete config shared across all editor instances` is now stale — it's per-instance.

**Not a bug, flagging so nobody re-raises it:** the stale section refs left by an `applyShortSyntax` *project* move are pre-existing and already documented as a bounded known follow-up in `section-shared.reducer.ts` ("LWW conflict-resolution can reassign a task's projectId... cleared by the next `dataRepair` pass"). Out of scope here.

## Splitting

Your commits already divide exactly along the boundary:

| PR | Commits | State |
|---|---|---|
| **A** — add bar | `5f8c79e4` + `8c0c913` | Reviewed, ready modulo the `A/B` nit |
| **B** — title edits + autocomplete | `558e182` + `3f98212` | Holds the 🔴 above |

Neither commit in **A** touches `short-syntax.effects.ts`, and the add bar never creates sub-tasks (that's `<add-subtask-input>`), so the bug **cannot occur in A**. Let's land A now and give B its own review — it earns one: it touches a meta-reducer and the hot-path task component. To be clear, B's design is good (riding `applyShortSyntax` atomically is the right answer to the two-ops question), it just needs the guard and a review pass of its own.

## Housekeeping

- **CI has never run on this branch** — zero check-runs on `8c0c913`. First-time contributors need a maintainer to approve workflows; I'll kick that off, so nothing here has had a CI signal yet.
- **The branch is 50 commits behind master** (based on `46e98c67dd`) — worth a rebase before merge.

Sorry to hand you another round after you'd already turned mine around so thoroughly — the add-bar half is in good shape and I'd like to get it merged.

caseyh:
Hey @johannesjo! Completely understand wanting to keep scope tight — that's good hygiene for momentum, and no need to be sorry: I'm the one who grew the scope mid-PR.

This branch is now exactly **A**, rebased onto current master: `5f8c79e` → `21ae6a7`, `8c0c913` → `459d040`, plus one new commit `b6acb83` fixing the `+A/B Testing extra words` regression you found — the whole-token compat guard now also fires with trailing text (first-word fallback, matching master's semantics), with a spec asserting the title this time.

**B** (`558e182` + `3f98212`) will come as a follow-up PR once A lands — with your sub-task guard applied (including the defensive check inside `moveTaskToSection`, in lock-step with the existing guards) and the eager `toObservable` effect made lazy per your hot-path note. Thanks for the trace on that one — the `convertToMainTask` resurrection path is exactly the kind of thing that would have bitten someone months from now. If you want a preview before the formal PR, the branch is already prepared: [`feat/quick-add-section-edit`](https://github.com/caseyh/super-productivity/tree/feat/quick-add-section-edit).

I've run the CI suite locally on the rebased branch (lint, unit, i18n, electron, e2e), so the workflow run should hopefully be uneventful.

johannesjo:
Round 3 — thanks for splitting this exactly as asked, and for the rebase. The `+A/B` fix is verified good.

But I have to walk back my round-2 sign-off on **A** ("ready modulo the A/B nit"). I re-ran the parser much harder this round — including a second, independent review pass — and **three real bugs live in part A**. My round-2 pass over A was too shallow; that's on me, not you.

## ✅ What checks out

- **The split is clean.** I diffed the pre-rebase tree against the new one: part B (`short-syntax.effects.ts`, `task-title.component`, the `moveTaskToSection` meta-reducer, `MentionConfigService.mentionConfigWithSections$`) is entirely gone. My round-2 🔴 is genuinely resolved — the add bar can't create sub-tasks and no effect/meta-reducer is touched. The one new bit vs the old A is inlining the `/` mention config into `add-task-bar.component`, which is fine: single instance, not the task hot path, so the eager-`toObservable` concern is moot now.
- **`+A/B Testing extra words` is fixed — and the spec is a real gate.** Running the extracted parser from both branches:
  ```
  "Task +A/B Testing extra words"   master => "Task Testing extra words"   PR => "Task Testing extra words"  ✓
  "Task +A/B extra words"           master => "Task extra words"           PR => "Task extra words"          ✓
  ```
  I also sabotaged the fix (deleted the new fallback block) and exactly one spec failed, with `Expected 'Task B Testing extra words' to be 'Task Testing extra words'`. That's a genuine regression test.
- `short-syntax.spec.ts` is 179/179 green locally.

## 🔴 1. `+/` hijacks ordinary text — `+/-` now assigns a project and mutates the title

```
"temp +/- 5 degrees"      master => (no match)   PR => title "temp - 5 degrees"   proj Work
"budget +/-10% ok"        master => (no match)   PR => title "budget -10% ok"     proj Work
"review PR +/- comments"  master => (no match)   PR => title "review PR - comments" proj Work
```

Root cause: when the token starts with `/`, `projectTitle` is `''` (`short-syntax.ts:341`), and `matchProject('')` matches **every** project — `title.indexOf('') === 0` is always true — so it returns `sortedAllProjects[0]`, i.e. whichever project has the *shortest title*. Master never reached this because it never split on `/`.

The project picked really is title-length-dependent, which makes it feel random to the user — same input, three different project sets:
```
projects [Work, Personal Stuff] => proj Work
projects [Work, Zoo]            => proj Zoo
projects [Work, Ab]             => proj Ab
```

`+/-` is ordinary notation, and it's also reachable mid-edit: backspacing the project name out of `+Work/Design` passes through `+/Design`, which silently jumps to the shortest-titled project and files into its section.

**Fix:** bail out when the left side is empty — e.g. `if (slashIndex === 0) return {};`, or guard `projectTitleToMatch` before calling `matchProject`.

## 🔴 2. Multi-word section names leave residue in the title

Same class as the round-1 bug this PR already fixed once, and it hits **both** forms. With sections `Design Reviews` and `Road Map` in `Work`:

```
"Task +Work/Design Reviews finish draft"  => title "Task Reviews finish draft"  sec=Design Reviews
"Task +Work/Road Map finish draft"        => title "Task Map finish draft"      sec=Road Map
"Task /Road Map finish draft"             => title "Task Map finish draft"      sec=Road Map
```

The section resolves *correctly*; only the strip is wrong. `matchSectionByTypedText` tries exactly two spans (`short-syntax.ts:268`) — the whole remainder, then its **first word**. Any trailing task text makes the whole-remainder attempt fail, so a two-word section falls all the way to one word and word 2 is orphaned into the title.

Multi-word section names ("Design Reviews", "Road Map", "In Progress", "To Do", "Next Up") are completely normal, so this isn't an edge case — it's the default experience for anyone who names a section with two words.

**Fix:** try decreasing leading word spans and consume the longest one that matches a section, instead of jumping straight to the first word.

## 🔴 3. A task can be filed into a section belonging to a *different* project

Reproduced against the **real** `AddTaskBarParserService` + **real** `AddTaskBarStateService` (throwaway Karma spec, projects `Inbox` (default) + `Work`, section `Design` in `Work`):

```
1. type "buy milk +Work/Design"   -> state {projectId: "work",  sectionId: "sec-design"}   ✓
2. delete just the "+Work" part   -> state {projectId: "inbox", sectionId: "sec-design"}   ✗  (sec-design belongs to "work")
```

Why: `parseAndUpdateText` passes the **previous** parse's `currentState.projectId` as the standalone-`/Section` context (`add-task-bar-parser.service.ts:86`), so `/Design` still resolves against Work. The project block then falls back to the default project, and `updateProjectId` correctly clears `sectionId` — but the section block immediately re-applies it through the `state().sectionId !== currentResult.sectionId` clause (`:242`), which is exactly the clause that defeats the reset.

On submit the component looks the section up **by id only** (`add-task-bar.component.ts:561-580`), and the `addTaskToSection` reducer only checks that the section exists (`section.reducer.ts:119`). So an Inbox task's id lands in a Work section's `taskIds` — persisted and synced. It renders nowhere today (`undoneTasksBySection` intersects with the project's own tasks), but it's the same latent shape as my round-2 finding: move that task into Work later and it materialises in a section the user never chose.

**Fix:** require `targetSection.contextId === state.projectId` before dispatching (a write-site guard, in lock-step with the existing ones), and drop `currentResult.sectionId` in the parser service when the resolved section doesn't belong to the final project.

## 🔵 Minor

- **A slash-titled project can't reach its own sections, and trying leaves residue.** With project `A/B Testing`: `"triage +A/B/Backlog"` → title `"triage B/Backlog"` (master: no match at all). Narrow, but worth a doc line that a `/` in a project title disables section syntax for that project.
- **Precedence introduced by `b6acb83`.** The new first-word fallback is a *prefix* match, yet it outranks a full project+section match. With projects `Work` + `Work/Design Archive` and section `Design Reviews` in Work, `"Task +Work/Design fix login"` picks `Work/Design Archive` and drops the section. Contrived fixture, but worth a thought: a whole-token match is strong evidence, a first-word prefix match is much weaker.
- **First-word section matching can eat a title word.** With a `Review Queue` section, `"+Work/review the Q3 numbers"` → title `"the Q3 numbers"`. That's inherent to prefix matching and consistent with `+`/`#`, so I'm not asking for a change — just keep it in mind when fixing #2.

## Housekeeping

- **CI has still never run on this branch** — `check-runs` on `b6acb838d0` is **0**. Your local run is appreciated, but nothing here has had a pipeline signal yet; I'll get the workflow approved.
- The branch is 22 commits behind master again.
- Heads-up on merge order: #9134 and #9135 are also open against `short-syntax.ts`. Nothing for you to do now, but whichever lands second will need a rebase.

Every parser result above comes from executing the real `short-syntax.ts` source (extracted → esbuild → node) against both branches, and #3 from a real-services spec — not from reading the regex.

caseyh:
Thanks for the round-3 pass — all three were real, and I'm glad you walked the sign-off back rather than letting them ship. All fixed, plus one more of the same class I found myself (below). The branch is rebased onto current `8ed1e49` and is now two commits: the feature (ported, see housekeeping) and `fix(add-task-bar): harden section short-syntax parsing (round-3 review)` — so "what changed since round 3" is diffable as the second commit.

## 🔴 1. `+/` hijack — fixed

A token starting with `/` now bails out of project parsing entirely, and `matchProject()` additionally refuses an empty string (two independent layers — either alone kills the bug). Your three repros plus the half-deleted `+/Design` state are specs, and I run the empty-left-side case against three different project sets so the title-length lottery can't quietly come back:

```
"temp +/- 5 degrees"   => no match (title untouched)   [Work, Personal Stuff] / [Work, Zoo] / [Work, Ab]
"budget +/-10% ok"     => no match
"buy milk +/Design"    => no match (sections present, context = Work)
```

## 🔴 2. Multi-word section residue — fixed

`matchSectionByTypedText` now tries decreasing leading word spans and consumes the longest one that matches, exactly as you suggested. Word offsets are computed on the raw typed text (not a split/rejoin), so the strip length is always a literal slice and internal double spaces can't desync it.

```
"Task +Work/Design Reviews finish draft"  => title "Task finish draft"   sec=Design Reviews
"Task +Work/Road Map finish draft"        => title "Task finish draft"   sec=Road Map
"Task /Road Map finish draft"             => title "Task finish draft"   sec=Road Map
```

Both forms are specs with title assertions. The "Review Queue eats `review`" prefix case is left as-is per your note.

## 🔴 3. Cross-project section — fixed at both ends

- **Parser service:** after the project block has settled the state, the parsed `sectionId` is validated against the project the task will actually land in — a section from any other project is dropped before it can touch state. This kills both paths in: your delete-`+Work` repro, and the variant where the user manually re-picks a project while the text still says `+Work/Design`.
- **Component:** the submit path now requires `targetSection.contextId === state.projectId` in lock-step with the existing write-site guards, so even a bad state can't put a foreign task id into a section's `taskIds`.

Your two-step repro is a spec against the real parser service (with a state double that mirrors updateProjectId's section-clearing), asserting `sectionId` ends `null` after the deletion step.

## One more I found (same class as your #1)

Running your extract-and-execute approach against a wider matrix turned up an interaction between the left-side match and the first-word fallback: with a section `Backlog` in *A/B Testing*, `"Task +A/B Testing extra words"` resolved the left side `A` → *A/B Testing*, then prefix-matched the bare `B` → `Backlog`, silently filing the task into a section nobody pointed at (the title came out identical to master's, which is what made it invisible). Rule added: **when the first word including the slash prefixes the same project as the left side, the slash is part of the project's own title and is never read as a section separator.** Spec'd.

## 🔵 Minors

- **Precedence:** a resolved project+section now outranks the first-word project-prefix fallback — your `Work` + `Work/Design Archive` example resolves to *Work → Design Reviews* (spec'd), while an exact whole-token project match still wins over any section reading (also spec'd).
- **Slash-titled projects:** doc line added to 3.04 — a `/` in a project title disables section syntax for that project. `"triage +A/B/Backlog"` still leaves `B/Backlog` in the title (master: no match at all); given the whole-token-first compat rule I think that's inherent, but happy to revisit if you disagree.

## Housekeeping

- **The rebase was a real port:** #9135's TrackedTitle parser replaced the string-based `parseProjectChanges` this PR originally extended, so the feature is re-implemented inside `parseProjectTracked` — section spans are consumed positionally on the tracked title and reported as `project` ranges, so they get the inline highlight for free. The old 3-commit history didn't survive that meaningfully, hence the squash into feature + fixes.
- **Verification:** same method as yours — extracted `short-syntax.ts` from master and this branch (esbuild → node) and diffed both against a 59-input matrix (prose slashes, URLs, unicode, casing, weird spacing, mid-edit states, interactions with `#`/`@`/`!`/estimates). Zero output differences on any input without a slash-project token; every diff on `+Project/…` tokens is the feature engaging with the round-1/2-approved strip semantics; zero out-of-bounds or whitespace-only highlight ranges. I also sabotaged each of the three fixes in isolation — each time only the corresponding new specs failed (3/4/2 respectively), so they're real gates, not decoration.
- Suites: short-syntax 232, parser service 95, state service 61, component 51, mentions 11, actions 62, effects 10, select-task 6 — all green on the rebased tree; full local `npm test` run green as well.
- Branch is 0 behind master as of this push; #9134 is closed so no merge-order concern remains. Ready for the workflow approval whenever suits you.

johannesjo:
Round 4 — reviewed `6f38d45cb1`. Thanks for the port write-up; treating the rebase as a re-implementation rather than a delta was the right call to flag, and I re-read the whole thing fresh because of it.

**All three round-3 blockers are genuinely fixed, and the gates are real.** I reproduced your sabotage numbers exactly rather than taking them on trust:

| sabotage | failures |
| --- | --- |
| `slashIndex === 0` bail **+** the `matchProject('')` guard (both removed) | 3 |
| `matchSectionByTypedText` descending → ascending word spans | 4 |
| parser-service `contextId === finalProjectId` validation deleted | 2 |

Also green here: 450 specs across `short-syntax` + the four add-task-bar suites, `checkFile` clean on all five modified `.ts`. I re-ran the extract-and-execute comparison against `8ed1e496` over ~1100 generated inputs: **zero** behaviour difference on any slash-free input, and **zero** out-of-bounds / whitespace-only / overlapping highlight ranges — that last one matters more than it looks, because #9135's `splitTextByRanges` *silently drops* an out-of-order range rather than glitching visibly.

The positional `consumeAt(len)` is a real improvement over master's `indexOf(matchedText)` re-find, and I confirmed `AddTaskBarState` is component-scoped signal state that is never persisted, so `sectionId` carries no schema or sync-format risk.

---

## 🔴 One blocker — and it is my fault, not yours

My round-3 🔵 asked you to rank a resolved project+section above the first-word project fallback. I labelled it contrived. Honouring it cost more than it bought: it silently changes which **project** a task lands in.

Projects `Dev` and `Dev/Ops`, section `Ops` inside `Dev`:

```
"+Dev/Ops"       master → Dev/Ops    PR → Dev/Ops            ok
"+Dev/Ops fix"   master → Dev/Ops    PR → Dev + sec Ops      title identical
```

Adding one trailing word flips the project. Same with `Client` + `Client/Acme` and a section `Acme` in `Client`: `+Client/Acme call back` goes to `Client` on this branch and `Client/Acme` on master. Across the matrix that is **123 project-resolution changes vs master, 99 of them with a byte-identical resulting title** — the only visible difference is the small project chip in the bar. Zero of them occur when no project title contains a `/`, so it is confined to users who put a slash in a project title, but for those it is a silent misfile.

Your own `+A/B Testing` guard covers the case where the first-word match and the left-side match are the *same* project. This is the cross-project case, and the guard can't see it.

**Fix: drop the precedence inversion and check the first-word slash-project first.** I applied and verified this locally — **0/1140** project changes vs master, section syntax otherwise unaffected, and it deletes more than it adds (-34/+11):

```diff
-    let firstWordSlashProject: Project | undefined;
+    // Legacy fallback ("+A/B Testing extra words"): the first word alone,
+    // slash included, as a project title prefix. Checked BEFORE the section
+    // reading so a "/" that belongs to a project's own title always wins.
     if (slashIndex !== -1) {
       const firstWord = rawToken.split(' ')[0];
       if (firstWord.includes(CH_SECTION)) {
-        firstWordSlashProject = matchProject(firstWord.toLowerCase());
+        const firstWordSlashProject = matchProject(firstWord.toLowerCase());
+        if (firstWordSlashProject) {
+          return {
+            projectId: firstWordSlashProject.id,
+            ranges: consumeAt(1 + firstWord.length),
+          };
+        }
       }
     }
 
-    if (
-      existingProject &&
-      slashIndex !== -1 &&
-      (!firstWordSlashProject || firstWordSlashProject.id !== existingProject.id)
-    ) {
+    if (existingProject && slashIndex !== -1) {
```

…and then delete the now-unreachable `if (firstWordSlashProject) { … }` block further down. The hoisted `let`, the three-clause condition and the ten-line "EXCEPT when…" paragraph all go with it.

Exactly **one** spec fails afterwards: `should rank a full project+section match above the first-word project fallback` — that's my round-3 note, please invert or drop it. A project whose title contains `/` gets to claim that slash; that is the simpler rule and it is what master already does.

---

## KISS / YAGNI pass

**The two `+/` guards are mutually redundant, and that weakens the tests.** I removed only the `slashIndex === 0` bail and left the `matchProject('')` guard: **232 SUCCESS, 0 failures**. Neither layer is individually gated — the specs only notice when *both* are gone. I also instrumented `matchProject` and ran 342 parses of every empty-left-side shape I could construct (`+/`, `+//`, `+ /x`, `+Work/`, `+A///B`, …): **`matchProject('')` is called 0 times** once the bail exists. It is unreachable, not defence in depth. Keep the bail — it states the actual rule — and let `matchProject` be a plain `find`.

Smaller ones, all optional:

- `...(sectionId ? { sectionId } : {})` in the `shortSyntax` return — the field is already `sectionId?: string`, so a plain `sectionId,` is equivalent and reads better.
- `AddTaskBarState.sectionId?: string | null` is both optional and nullable; `INITIAL_ADD_TASK_BAR_STATE` and `updateSectionId` always write it, so `sectionId: string | null` is enough.
- `first()` inside `firstValueFrom(...)` at `add-task-bar.component.ts:624` — `firstValueFrom` already does that.
- The `trim()` + collapse-double-space tail is duplicated verbatim in `consumeAt` and `parseStandaloneSectionTracked`; worth a two-line helper.

**On scope:** I did price cutting the standalone `/Section` form (it accounts for the extra regex, `parseStandaloneSectionTracked`, the `contextProjectId` parameter on the public `shortSyntax()` signature, and the stale-context half of round-3 blocker 3). I don't want to cut it — when you're already inside a project, `/Design fix login` is the ergonomic half and `+Work/Design fix login` is the redundant one. Keeping it. See the last minor below for the thing I do want checked before merge.

---

## 🔵 Minors

1. **The `/` mention config ignores the project short-syntax setting.** Both parser paths sit inside `if (config.isEnableProject)` and 3.04 now says "Enabled by: the project Short Syntax setting", but `mentionCfg$` appends the `/` mention unconditionally. With project syntax off you still get a section dropdown, and picking an item leaves `/Design ` sitting in the title.

2. **Non-space whitespace in a section name leaves residue** — same class as round-3 #2, just with a tab or NBSP instead of a space. With section `Road Map`: `"x /Road\tMap y"` → title `"x \tMap y"`, and the NBSP variant behaves the same. `toMatch` normalises with `.replaceAll(' ', '')`; `/\s+/g` would cover it. This matches master's project-matching convention, so genuinely your call whether it's worth changing.

3. **The component write-site has no test.** `add-task-bar.component.spec.ts` is 1148 lines and contains zero occurrences of "section". The parser-service half of blocker 3 is well spec'd, but the half that actually writes a task id into a *synced* `section.taskIds` — the `targetSection.contextId === state.projectId` guard and the `!isAddToBacklog()` skip — is uncovered. I'd like at least the guard covered before merge.

4. **The backlog skip isn't documented.** `+Work/Design` with add-to-backlog on silently drops the section, with no cue. The behaviour looks right; one sentence in 3.04 would do.

5. **Standalone `/Section` eats prose when a section title is a common word.** With sections `API`, `Docs`, `Notes`:

   ```
   "deploy /api gateway"     → "deploy gateway"   sec=API
   "read the /docs first"    → "read the first"   sec=Docs
   "flag --verbose /notes"   → "flag --verbose"   sec=Notes
   ```

   And `mention.directive.ts` has no word-boundary check on trigger chars, so the dropdown opens on *every* `/` — `https://`, `w/`, `1/2`. That is tolerable for `#`/`@`/`+`/`!` and much less so for `/`. This is inherent to the form rather than a defect in your implementation, and I'm not asking for a change — but please give it a few minutes of real typing in the preview build and tell me whether it feels noisy. If it does, I'd rather find that out now than after release.

---

## Housekeeping

CI has still never run on this branch — seven workflows are sitting in `action_required` on `6f38d45cb1`. That's on me for the fourth round running; I'll get them approved.

So: the blocker above plus a test for the component guard, and I think this is ready.

caseyh:
Round 4 addressed — everything applied, one cleanup deliberately declined (receipts below), and the branch has since been rebased twice more to keep up with master, the second time across your #9679 word-wise matcher rewrite. Now three commits on `f084cd09`: feature, round-3 fixes, round-4 fixes.

## 🔴 Precedence inversion — dropped, your diff applied (and then re-expressed on #9679)

You were right that honouring my round-3 note cost more than it bought, and the "it's my fault" is shared: I implemented it, and my matrix never compared *project resolution* against master — which is exactly why 99 byte-identical-title flips sailed through. That comparison is now a first-class check in my harness alongside range validation.

On the current base the rule composes with your word-wise matcher cleanly:

1. Run the word-wise walk on the full token. **If the matched typed title extends past the first `/`, the slash belongs to the project's own title** — project only, never a section. `+Dev/Ops fix` stays in *Dev/Ops*, `+A/B Testing extra words` consumes the full title (per #9679), `+Work/Design fix login` with a "Work/Design Archive" project goes to the archive project — all by construction.
2. Only when no match explains the slash: resolve the left side with the same word-wise rules, then the section on the right (longest word span).
3. **A slash that resolves no section never changes the project**: `+Work Inbox/whatever` with projects *Work* + *Work Inbox* stays in *Work* exactly like master (spec'd both ways — with a matching section, `+Work Inbox/Triage` files into *Work Inbox → Triage*).

Verification with the flip check: **0 project-resolution changes vs current master** across the matrix plus your Dev/Ops, Client/Acme, and Work/Design Archive fixtures *and* new #9679-territory fixtures (`+Work in progress`, `+Home work on taxes`, multi-word titles with slash forms). Sections active throughout. 0 range issues.

## KISS

- **`matchProject('')` guard removed** — it was unreachable and "two layers, neither gated" was a fair hit. The `slashIndex === 0` bail states the rule and is spec-gated alone. (The guard then vanished entirely with the #9679 port — the word-wise matcher replaced that code.)
- **`AddTaskBarState.sectionId` is now `string | null`** (not optional) — this rippled `sectionId: null` into ~20 existing spec fixtures, which is most of the spec-file noise in the round-4 commit.
- **`first()` inside `firstValueFrom` removed**; the trim/collapse tail is a shared `tidyAfterRemoval()` used by all three consume paths.
- **Declined: the plain `sectionId,` return.** Tried it — 38 existing specs assert the exact result shape and fail on a present-but-undefined key. Kept the conditional spread with a comment; rewriting 38 assertions for a cosmetic key felt like the wrong trade. Happy to flip it if you prefer.

## 🔵 Minors

1. **`/` mention gated on the project short-syntax setting** — with project syntax off there's no section dropdown, matching the parser paths and the 3.04 wording.
2. **Whitespace normalization**: section matching normalizes `\s+`, spec'd with your `"x /Road\tMap y"` repro.
3. **Component write-site spec'd** (5 tests): files into its own project's section `(sectionId, taskId, null, null)`, respects add-to-bottom, refuses a foreign section, refuses a vanished section, skips backlog adds — task creation still succeeding in every refusal case.
4. **Backlog skip documented** in 3.04, plus the dropdown-gating line.
5. **The `/`-noise typing test** — ran it against a dev build (fixtures: Work with "Design Reviews"/"Road Map", Dev with "Ops", a "Dev/Ops" project): typed URLs (`https://example.com/foo`), `w/ care`, `1/2`, `+/-`, and the worst case `pair w/Design tweaks` (mid-word slash followed by a real section prefix). **The dropdown fired zero times on non-section slashes** — the mention list only renders when the slash text prefix-matches an actual section of the effective project, so the "opens on every `/`" concern stays theoretical in practice. It opened correctly for `/De`, placement landed right in every case (`/Road Map review` → Road Map, title "plan Q4 review"; `+Dev/Ops fix` → project chip and task in Dev/Ops), and the #9135 highlighter marks exactly the consumed span. Verdict: not noisy; no change wanted.

## Housekeeping

- Rebases crossed your interval-recurrence work (#9328), the repeat model rename, and the #9679 matcher rewrite — that last one replaced the code this feature extended, so the section split was re-expressed inside the new `parseProjectTracked` rather than diff-merged; two of my earlier specs were updated to #9679's (better) semantics of consuming the fully typed project title.
- Suites on the rebased tree: short-syntax 265, parser service 118, component 71, state service 61, mentions 11 — green; full 14.4k unit suite green except `caldav-client.service.spec.ts` ("connects when the server omits principal-collection-set…"), which fails for me in isolation with zero CalDAV files in this diff — it looks broken on master HEAD (last touched in bc28a66a), worth a look on your side.
- Final adversarial pass before this push: a 2,268-input generated matrix (prefixes × tokens × suffixes × three fixture families) with four checks — no-slash parity vs master, project flips without a section resolving, range validity, title hygiene — all zero; and each of the three slash rules sabotaged in isolation fails only its own specs (4/3/1). That pass caught me re-creating the round-4 anti-pattern — a `projectTitle.trim()` ternary silently duplicating the `slashIndex === 0` bail — so the bail is now the single, sabotage-gated guard, with a comment proving the equivalence.
- Ready for the workflow approval whenever suits.

johannesjo:

Round 5 on `9be0825`. I re-ran the extract-and-execute comparison myself (both branches bundled with esbuild, executed over a 2232-input matrix of prefixes × slash/non-slash tokens × suffixes across four project/section fixture families) rather than taking round 4's numbers on trust:

- **Zero** behaviour difference vs master on any slash-free input.
- **Zero** out-of-bounds, whitespace-only or out-of-order highlight ranges.
- The round-4 blocker is genuinely gone: `+Dev/Ops fix` stays in *Dev/Ops*, `+Client/Acme call back` stays in *Client/Acme*, `+Work/Design fix login` picks *Work/Design Archive*. The only project-resolution flips left vs master are the `+Work Inbox/Triage` family, which is the case you spec'd deliberately.
- Round-3's three fixes all hold: `+/- 5 degrees` is inert, `+Work/Road Map finish draft` strips cleanly, `x /Road<TAB>Map y` strips cleanly.
- Round-4's KISS items all landed (the `matchProject('')` guard is gone, `sectionId` is `string | null`, `tidyAfterRemoval` is shared, no `first()` inside `firstValueFrom`).

Two new defects, both in the standalone `/Section` path and both about the same thing: that path is the only one that does **not** get the `+` token regex's lookahead treatment.

### Should fix before merge

- **`short-syntax.ts:658-660`** – `SHORT_SYNTAX_STANDALONE_SECTION_REG_EX` captures with `[^${ALL_SPECIAL}]+`, but `ALL_SPECIAL` is the *string* `"(\+|\#|\@|\!)"`, built for use as an alternation group. Dropped inside a character class its parens and pipes become literal members, so the negated set is `{ ( ) | + # @ ! }` – three of those (`(`, `)`, `|`) were never meant to be special anywhere. The capture therefore stops inside the **section title**; `matchSectionByTypedText` still prefix-matches the truncated text against the full title, so the section resolves but only the truncated span is stripped.

  Against the real parser on this branch (project *Work*), with the `+Project/Section` form as the control:

  ```
  section "Design (v2)"
    "x /Design (v2) y"        => title "x (v2) y"        sec=Design (v2)
    "x +Work/Design (v2) y"   => title "x y"             sec=Design (v2)   <- correct

  section "Ship it!"
    "x /Ship it! y"           => title "x ! y"           sec=Ship it!
    "x +Work/Ship it! y"      => title "x y"             sec=Ship it!      <- correct

  sections "C" and "C++"
    "fix /C++ crash"          => title "fix ++ crash"    sec=C             <- wrong section
    "fix +Work/C++ crash"     => title "fix crash"       sec=C++           <- correct
  ```

  The mis-selected id in the last case is what `addTaskToSection` persists, and all of it is reachable straight from the new `/` dropdown, which inserts `s.title` verbatim.

  **Fix:** let the standalone capture run to the end of the token and rely on `matchSectionByTypedText`'s descending word-span walk to find the boundary, the way the `+Project/Section` path already does – i.e. `(.+)$` in place of `[^${ALL_SPECIAL}]+`. Worth two specs: a section title containing `(`, and a `C` / `C++` prefix collision. Neither shape appears in the 32 new section specs, which is why this got through.

  *Scoped deliberately:* `@` and `#` in a section title also truncate, but that half is **not** yours – master already does exactly the same for project titles (`"x +Standup @ 3pm y"` sets a 3pm due date on master and on this branch, byte-identical), so it is the existing "special chars end the token" rule, not a regression. Only `(`, `)`, `|`, `+` and `!` are new, and only in the standalone form.

- **`short-syntax.ts:670`** – `parseStandaloneSectionTracked` matches with a non-global regex and uses only the first hit, so an earlier slash token that resolves no section swallows a later real one instead of falling through to it. The capture is greedy, so the second token is inside it but no word span of the combined text prefix-matches anything:

  ```
  section "Design", context project Work
    "ship /Design now"          => title "ship now"   sec=Design   <- correct
    "review PR /9014 /Design"   => no section, title untouched
    "note /2 things /Design"    => no section, title untouched
    "a /todo b /Design c"       => no section, title untouched
  ```

  It fails safe (nothing wrong is written), but the `/` dropdown is driven by the mention library rather than the parser, so it still opens on that second slash and offers a section that the parser then ignores – the user picks one and nothing happens, with no feedback. **Fix:** match with `/g` and take the first candidate that actually resolves.

### Nits

- **`add-task-bar-parser.service.ts:116`** – `await firstValueFrom(this._store.select(selectAllSections))` runs on every keystroke. `selectAllSections` is memoised so it is cheap, but a `toSignal`/`selectSignal` read would avoid the per-parse subscribe/unsubscribe in the typing path. Same for the second lookup in `add-task-bar.component.ts:648`.

---

caseyh:
Both round-5 defects fixed, plus the nit — thanks for the exact fixture receipts, they made this quick. Rebased onto current `master` (`fb23ddda`) first; the branch is a fast-forward of it.

**Broken character class** — fixed as you prescribed: the standalone capture now runs to the end of the line (`(.+)`) and `matchSectionByTypedText`'s descending word-span walk finds the boundary, the same way the `+Project/Section` path does. I used `(.+)` rather than `(.+)$` — `.` already stops at a line break, and the unanchored capture keeps a `/Section` parseable when it isn't on the last line. Your three fixtures now behave like their `+Work/...` controls:

```
"x /Design (v2) y"   => title "x y"        sec=Design (v2)
"x /Ship it! y"      => title "x y"        sec=Ship it!
"fix /C++ crash"     => title "fix crash"  sec=C++
```

Specs added for a `(`-containing title, `!`, the `C`/`C++` collision both ways (`/C++` → C++, bare `/C` → C via the shortest-title rule), and a `+Work/Design (v2)` control asserting path parity.

**Non-global match** — `parseStandaloneSectionTracked` now walks every slash token and takes the first that resolves. One wrinkle in the fix sketch: with the capture running to end-of-line, a plain `/g` exec loop still skips later tokens — the greedy capture of a failed candidate spans them, so `lastIndex` lands at end-of-input. After a non-resolving candidate the scan therefore resumes at `slashPos + 1` instead (strictly advancing, so no loop risk). Spec'd both ways: `"review PR /9014 /Design"` files into Design with `/9014` kept in the title, and a title where no token resolves stays untouched.

**Nit** — both `selectAllSections` reads are `selectSignal` now (parser-service keystroke path and the component submit path); the two `firstValueFrom` subscribe/unsubscribe cycles are gone.

**A correction to my round-5 comment:** it said the redundant `projectTitle.trim()` ternary had been removed with a comment proving equivalence — it hadn't; the amend that should have carried that change never made it into what I pushed, and `9be0825` still contains the ternary. It is actually gone in this commit (with the equivalence comment), and the discrepancy is called out in the commit message too. Sorry for the false receipt.

Verification, same method as before (extract-and-execute both branches via esbuild):

- 1,080-input matrix (prefixes × tokens × suffixes × 4 fixture families, now including a special-chars family with `Design (v2)` / `Ship it!` / `C` / `C++` and multi-slash tokens): zero behaviour differences vs master on slash-free inputs, zero project flips where master resolved a project, zero range issues, zero title-hygiene issues.
- A direct differential of this tip against `9be0825` over the same matrix: exactly 72 outputs changed (45 unique inputs), every one of them one of the two fixed classes — 27 special-char standalone titles, 18 multi-slash fall-throughs. The ternary removal has zero behavioural footprint. No collateral on the `+` path or on prose slashes (`w/`, `1h/2h`, `either/or`, URLs).
- Sabotage-gating on the final tree, each rule reverted in isolation: `slashIndex === 0` bail → 3 failures, project-claims-its-slash rule → 4 (all four its own specs), `!match` strip-through gate → 1, character-class regression → 3, fall-through removal → 1. No sabotage touches another rule's specs.
- Typed against the real add bar on a clean profile with those fixtures: all placements and titles as above; `+Dev/Ops fix` still lands in the *Dev/Ops* project; the `/` dropdown stays silent on `/9014` and opens on `/Des`.
- Suites: short-syntax 272, parser service 118, component 78, mentions 11, state service 61; lint clean; full unit suite green (15,628 specs, the CalDAV one included now).

One doc line added to 3.04 for the fall-through behaviour.

johannesjo:

Thanks for the updates. Re-checked the 3 items from the earlier review at `9be0825`. The rebase only moved master context, so the response is `775f78c`:

- ✅ **Standalone `/Section` capture truncating at `( ) | + !`** – fixed in `short-syntax.ts:662-665`. The capture now uses `(.+)`, and the word-span walk finds the boundary. `x /Design (v2) y`, `x /Ship it! y` and `fix /C++ crash` now behave like their `+Work/...` controls, and the new specs cover all three.
- ✅ **An earlier non-resolving slash swallowing a later real one** – fixed in `short-syntax.ts:683-699`. `review PR /9014 /Design`, `note /2 things /Design` and `a /todo b /Design c` all resolve now. The loop always terminates (`lastIndex` moves strictly forward), and `lastIndex` is reset on entry and on success.
- ✅ **Per-keystroke `firstValueFrom(select(selectAllSections))`** – fixed in `add-task-bar-parser.service.ts:66` and `add-task-bar.component.ts:290`. Both are now `selectSignal` fields.

### Should fix before merge
- **`short-syntax.ts:639` / `:690`** – this comes from the fix itself. Each non-resolving slash candidate now runs `matchSectionByTypedText` over the whole rest of the line. That function walks every leading word span, starting from the longest, and squashes and lowercases each span. The cost is candidates × words × length, where before it was a single candidate. I measured this by running the parser on this branch and on the commit before it (one line, 10 sections, a ` /pathN` token every 10 words):

  ```
  words   before 775f78c   at 775f78c
    500        6 ms            80 ms
   1000       18 ms           621 ms
   2000       73 ms          5273 ms
  ```

  Nothing debounces the parse on input changes, so pasting one long line full of paths stalls every keystroke after it. Normal titles are unaffected: 200 words took about 10 ms. `parseProjectTracked` already bounds its walk for exactly this reason (`maxNrOfWords`, "which also bounds the walk for long input"). **Fix:** in `matchSectionByTypedText`, return early when the project has no sections. Then skip any span whose squashed length is longer than the project's longest squashed section title, because such a span can never prefix-match. Unlike a word-count cap, this keeps `Des ign` → `Design` working, and it makes each candidate's walk cost proportional to the section-title length instead of the line length.

---

johannesjo:
One more item alongside the parse-time one: CI's `Tests` job fails at this head (reproduced locally, 1 failed / 15695), in a spec this PR doesn't touch:

`ProjectService duplicateProject › should duplicate sections and remap their task membership` — `Expected 0 to be 2` (`project.service.spec.ts:462`).

It passes on its own and the full suite is green on master, so it's test pollution. Likely cause: the new add-task-bar specs override the real `selectAllSections` selector (`provideMockStore({ selectors: [...] })` / `store.overrideSelector(selectAllSections, ...)`) without `store.resetSelectors()` in an `afterEach`. Overrides on a real memoized selector persist across spec files, so `selectSectionsByContextIdMap` (used by `duplicateProject`) later sees `[]`. Adding the reset should clear it.

caseyh:
Both fixed in one new commit on top (`045c5e5`), after rebasing onto current `master`.

**Parse time on long lines** — done as you prescribed in `matchSectionByTypedText`: return early when the project has no sections, and stop collecting spans once their squashed length exceeds the longest squashed section title. Stopping the collection (rather than skipping spans inside the walk) means the regex scan also ends there, so each candidate's cost is bounded by section-title length, not line length. Section titles are now squashed once per call rather than per span. Your benchmark shape (one line, 10 sections, ` /pathN` every 10 words, median of 3):

```
words   9be0825   775f78c   this tip
  200      1 ms      5 ms      0 ms
  500      4 ms     54 ms      0 ms
 1000     11 ms    382 ms      1 ms
 2000     43 ms   2858 ms      2 ms
```

It also removes the quadratic walk the single-candidate path had before `775f78c` (the 43 ms). Harsher shapes at 2000 words, before → after: every word a slash token 26 s → 9 ms; one 300-char section title 2.7 s → 7 ms; 200 sections 6.9 s → 8 ms; sections only in another project 2.7 s → 2 ms.

The bound is exact, not a heuristic: a differential of this tip against `775f78c` over the 1,080-input matrix plus split-name and long-tail cases (`/Des ign`, `/Design Rev iews …`, `/Road Map` followed by 300 words, a late `/C++` after 300 words, `+Work/Design Reviews` followed by 300 words) shows zero changes in titles, ids and highlight ranges. Two specs: `/Des ign` → Design still resolves, and a 2000-word line of slash tokens with a trailing `/Design` resolves in under 500 ms, which follows the `performance.now()` guard pattern in `render-links.pipe.spec.ts` / `create-sorted-blocker-blocks.spec.ts`. Removing the bound fails exactly that spec (2163 ms).

**CI `Tests` failure** — your diagnosis was right. I reproduced it deterministically before fixing, by running this file's specs and then `project.service.spec.ts` in that order with random order off: `duplicateProject` → `Expected 0 to be 2`, the same as CI's second, TZ, pass. The spec file now calls `TestBed.inject(MockStore).resetSelectors()` in a top-level `afterEach`. Both TestBeds and the `overrideSelector` call sit inside that describe, and the override lives on the shared selector object, so one reset covers all three. With the reset, the forced order is 143/143 green; removing it brings the failure back.

I owe you a correction here as well: that same `duplicateProject` failure showed up in my own full-suite runs in rounds 4 and 5, and I wrote it off as a contention flake because it passed in isolation. It was this pollution all along.

Suites: short-syntax 274, parser service 118, component 78, mentions 11, state service 61. Lint is clean, and the full unit suite is green on both passes.

PR #9048, opened 2026-07-15
fix(sync): recover project notes/sections/repeat-cfgs on losing delete

### LINKED ISSUE #9037: sync: a legacy project-delete losing an LWW race partially resurrects (project + active tasks only)
**From a post-merge sync review of `18c48d5e..7e273a0e` (2026-07-15).**

The schema-v4 project-delete-wins barrier (#9009) deliberately leaves pre-v4 `deleteProject` operations unmarked, under timestamp-LWW semantics (`packages/shared-schema/src/migrations/project-delete-wins-barrier-v3-to-v4.ts`, `migrateState: state => state`, `requiresOperationMigration: false`).

When such a legacy delete **loses** an LWW race to a concurrent project `UPDATE`, the task-recovery path (#9007) only reconstructs the **project entity + active/backlog tasks**:

- `src/app/op-log/sync/conflict-resolution.service.ts:3054-3055` — comment: "Notes, archived tasks, and other deleteProject cascades are intentionally outside this task-recovery path; they need their own snapshot design."
- `handleDeleteProject` (`project-shared.reducer.ts:136-179`) cascades far more: removes tasks from all tags, cleans TIME_TRACKING, removes TASK_REPEAT_CFG, and (async) archive cleanup — none of which are restored.
- `conflict-resolution.service.ts:3259-3272` — the `onMissingBaseEntity` fallback: a losing local delete without a full base leaves the entity deleted and drops the winning remote UPDATE's changes.

## Impact
The project + active tasks reappear, but notes, archived tasks, sections, repeat configs, and time-tracking cascades stay deleted → partial / divergent state across clients. Bounded to the v3→v4 rollout window **plus** a concurrent delete-vs-update race on the same project. Verified real but rollout-bounded; it is an in-code-acknowledged limitation, not an unrecognized bug.

## Not a quick fix
Making legacy deletes delete-wins at the v4 barrier would change deliberately-preserved semantics and risk the **reverse** loss (a legitimate "keep the project" update losing). Full cascade restoration is exactly the deferred "own snapshot design."

## Options to decide
1. Delete-wins for legacy project deletes at the v4 compatibility barrier (accepting the reverse-loss trade-off), or
2. Complete cascade restoration (notes/archived/sections/repeatCfg/time-tracking).

Partial restoration as-is is not safe long-term.

johannesjo:
## Summary

Fixes #9037. When a legacy (pre-schema-v4) `deleteProject` loses an LWW race to a concurrent project UPDATE, the recovery path previously recreated only the project entity and its active/backlog tasks. Notes, sections, and task-repeat-cfgs stayed deleted, so every client converged to a lossy shape on hydration replay of the durable loser row — data loss against the winning "keep the project" intent.

This recreates those three adapter entities at resolution time on the winner (which still holds them), via the generic `lwwUpdateMetaReducer` recreate path — the same mechanism the existing task recovery uses.

## What changed

- New `_createCascadeRecreationOpsForWinningProject` runs alongside `_createTaskRecreationOpsForWinningProject` at the recovery call site.
- Notes are enumerated from the delete payload's `noteIds`; sections/repeat-cfgs are scanned from the store (they aren't in the payload) via the non-throwing `selectEntities` selectors (`selectNoteById`/`selectTaskRepeatCfgById` throw on missing ids).
- Per-entity-type concurrency guard so a note/section/cfg another device is concurrently deleting is not resurrected; dead task refs are stripped from recreated sections.
- One shared `recreationClock` dominates the delete; `Array.isArray(noteIds)` guards legacy ops.

## Scope

**Deferred to a follow-up (own snapshot design):** time-tracking and menu-tree (singletons — no safe single-project restore) and archived tasks + archive time-tracking (separate persistence; the winner's IndexedDB archive survives hydration and is a viable future source).

**Documented, converging limitations (all strictly better than losing the entity):** sections/repeat-cfgs have no `modified` field so a concurrent content edit can be clobbered by the replace snapshot; a note's `todayOrder` slot is not restored (project-level note membership *is* restored via the project compensation).

## Process

Plan was multi-reviewed by two independent reviewers (sync-correctness + feasibility) before implementation; all must-fix findings folded in.

## Tests

- Producer: recreation happy-path + concurrency-guard/section-taskId-strip (`conflict-resolution.service.spec.ts`).
- Consumer: SECTION recreate round-trip through `lwwUpdateMetaReducer` (`lww-update.meta-reducer.spec.ts`).
- 198/198 conflict-resolution + 127/127 meta-reducer specs pass; `checkFile` clean.

### DIFF
--- a/src/app/op-log/sync/conflict-resolution-persistence.integration.spec.ts
+++ b/src/app/op-log/sync/conflict-resolution-persistence.integration.spec.ts
@@ -1030,7 +1030,23 @@ describe('ConflictResolutionService persistence (integration, real store)', () =
     ]);
   });
 
-  it('converges a losing deleteProject by recreating notes, sections and repeat-cfgs (#9037)', async () => {
+  // Shared fixture: a legacy deleteProject that loses LWW to a concurrent project
+  // edit, with the project holding a task, note, section and repeat-cfg. Returns
+  // the seeded pre-delete state, a full-slice replay reducer (wiring the cascades
+  // this suite's default reducer omits — sections via the section meta-reducer,
+  // notes/repeat-cfgs via their feature reducers, recreations via
+  // lwwUpdateMetaReducer), and the durable ops the service produced.
+  const setupLosingProjectDeleteFixture = async (): Promise<{
+    noteId: string;
+    sectionId: string;
+    cfgId: string;
+    seededState: RootState;
+    replay: (state: RootState, operations: Operation[], clientId: string) => RootState;
+    hasEntity: (state: RootState, feature: string, id: string) => boolean;
+    storedOps: Operation[];
+    localProjectEdit: Operation;
+    remoteProjectDelete: Operation;
+  }> => {
     const noteId = 'note1';
     const sectionId = 'section1';
     const cfgId = 'cfg1';
@@ -1043,8 +1059,6 @@ describe('ConflictResolutionService persistence (integration, real store)', () =
       projectId: 'project1',
       subTaskIds: [],
     } as Task;
-    // Full-slice seed state: the losing delete's cascade removes all four, and
-    // the recovery ops must restore them on replay.
     const baseTaskState = createTaskReplayState([rootTask], [rootTaskId], []);
     const seededState: RootState = {
       ...baseTaskState,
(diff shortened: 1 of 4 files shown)


johannesjo:
Plan was multi-reviewed (sync-correctness + feasibility) before implementation; the diff was then reviewed by Codex. No new defects surfaced — the three flagged items are all documented, converging, non-regression residuals tracked in #9050 (concurrent same-batch add, section/cfg LWW timestamp, pinned-note `todayOrder`). This PR is a strict reduction in the #9037 data loss.

PR #9052, opened 2026-07-15
fix(tasks): self-heal orphan tasks to Inbox on navigation (#8780)

### LINKED ISSUE #8780: 🚨 Search Problem

johannesjo:
## Problem
Fixes #8780 — tapping a global-search result does nothing for certain tasks. The reporter confirmed on v18.14.0 that it still fails, but only for tasks **not assigned to a project**.

The prior fix (#8801) routed a project-less/tag-less task to the Today list, on the assumption it only ever lives there. But a task with a falsy `projectId` — most commonly an **empty-string `projectId`, which survives hydration because it passes typia validation** (unlike `undefined`, which gets auto-repaired to Inbox) — with no tags and not due today, has its id in **no** work context's `taskIds` ordering array. A project/tag list renders from that ordering array (`WorkContextService.mainListTaskIds$`), not by filtering `task.projectId`, so the task renders in no list. Navigation lands on Today with the task absent and `focusTaskInViewWhenReady` silently gives up → "wrong view / nothing happens."

## Fix
- **Self-heal on access:** when `_getLocation` resolves an orphan (falsy `projectId`, no tag, not due today), re-home it into the Inbox via the existing `moveToOtherProject` op — which assigns `projectId` **and adds the id to `INBOX.taskIds`** so it renders — then route to `/project/INBOX_PROJECT/tasks`. The move reducer no-ops source removal for an empty/dangling source and reads canonical subtask data from state, so replay is deterministic and concurrent heals converge (LWW `projectId=INBOX` + `unique()` taskIds).
- **No silent no-op:** surface the existing error snack when a same-context task can never be focused, instead of leaving the user on the wrong view.

Note: `deleteProject` fully removes task entities (`task.reducer.ts`), so no dangling-`projectId` tasks persist — only the falsy-`projectId` legacy case is targeted.

## Tests
Rewrote `navigate-to-task.service.spec.ts` (5 tests, fails against the old TODAY-routing, passes now): orphan → dispatches move-to-Inbox + routes to Inbox; project/tag/today paths unchanged and dispatch nothing; same-context focus failure → error snack. `layout.service.spec.ts` 12/12 green.

## Sync-correctness
Reviewed: the dispatch fires only on user-gated navigation (not an effect/replay), is one meta-reducer pass (one op per intent), is replay-deterministic (reducer reads canonical state), idempotent, and convergent under concurrent heals. The only residual is the pre-existing `moveToOtherProject` LWW surface (shared by drag-drop, board, menu, etc.), bounded here to already-orphaned tasks.

### DIFF
--- a/src/app/core-ui/layout/layout.service.spec.ts
+++ b/src/app/core-ui/layout/layout.service.spec.ts
@@ -1,4 +1,4 @@
-import { TestBed } from '@angular/core/testing';
+import { fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { Store } from '@ngrx/store';
 import { LayoutService } from './layout.service';
 import { hideAddTaskBar, showAddTaskBar } from './store/layout.actions';
@@ -340,5 +340,20 @@ describe('LayoutService', () => {
         done();
       }, 400);
     });
+
+    it('should invoke onFailure (and not onSuccess) once the retries are exhausted', fakeAsync(() => {
+      const onSuccess = jasmine.createSpy('onSuccess');
+      const onFailure = jasmine.createSpy('onFailure');
+
+      // No element with id `t-never-rendered` exists, so the task never becomes
+      // focusable. Use a small retry budget so the loop exhausts quickly.
+      service.focusTaskInViewWhenReady('never-rendered', onSuccess, onFailure, 2);
+
+      // 2 retries * 250ms delay; tick generously to drain every scheduled retry.
+      tick(1000);
+
+      expect(onSuccess).not.toHaveBeenCalled();
+      expect(onFailure).toHaveBeenCalledTimes(1);
+    }));
   });
 });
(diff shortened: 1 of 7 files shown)


Sanja2012:
Das Suchproblem existiert weiterhin in SP 18.16.0 auf Android und Windows. Der Suchfehler (Klick auf Suchergebnisse führt ins Leere) tritt bei Aufgaben auf, die nicht einem Projekt zugewiesen sind. Leider ist die App damit für mich unbrauchbar.

PR #9058, opened 2026-07-16
feat(schedule): add a single-day view to the Schedule tab

jkusa:
I recently started using Super Productivity for time-block planning, and the Schedule view has quickly become where I plan my day. One bit of friction I ran into is focusing on a single day: when several events share the same time slot, the week view squeezes those events into narrow columns that can be hard to read.

## Problem

There isn't currently a way to focus the Schedule to a single day; the Week view shows the full week, so each day stays narrow, especially on small screens.

## Solution

Adds a **single-day** view to the Schedule tab, alongside the existing Week and Month views. A day view gives a single day the full width, which is much easier to read and plan against on desktop and mobile.

The view is a third toggle in the existing Week/Month selector. It reuses the existing `schedule-week` component with a day count of 1, so there is no new rendering path to maintain; it just drives the existing building block with a narrower range.

### Behavior

- New **Day** toggle in the Schedule header, left of Week/Month.
- Header shows the single date (e.g. `Tue, Jul 14, 2026`) instead of a week range.
- Prev/Next step exactly one day at a time, and snap back to "today" when you reach it; the today reset and Prev lock while today is shown, matching the existing week/month behavior.
- The selected view persists across reloads via the existing `SELECTED_TIME_VIEW` key; unknown/absent values still fall back to `week`.
- Horizontal scroll stays off in day view.

### Screenshots

| | Day view | Week view |
|---|---|---|
| Desktop | day-view-roomy | week-view-crowded  |
| Mobile | mobile-day-view | mobile-week-view |

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo (commented):
Thanks for this — it's a well-made PR, and the problem is real. I especially like that you drove the existing `schedule-week` with a day count of 1 instead of adding a rendering path: that's the same approach `schedule-day-panel` already takes, so it's idiomatic here rather than novel. The result is ~58 lines of production code (13 of them actual logic) against 213 lines of tests, which is a ratio I'd like more feature PRs to have. Docs and the `en.json` → `npm run int` flow were done properly too — I regenerated `t.const.ts` locally and it came out byte-identical.

I ran things locally: 55/55 unit tests pass, your new e2e passes (8.2s), eslint + prettier clean.

There's one thing I'd like fixed before merge, and one deeper thing that's **not your fault** but that this PR makes routine — happy to discuss.

## 1. The day header needs a compact form (please fix)

`headerTitle()` returns the day branch early, before the `_isCompact` / `_isVeryCompact` handling that the week branch uses:

https://github.com/super-productivity/super-productivity/blob/39b8415b84f5670a3d51ae68f17cc80c420d76dd/src/app/features/schedule/schedule/schedule.component.ts#L197-L225

So day view always renders the full `'EEE, MMM d, yyyy'` — the longest of the three titles — while `.title` is `nowrap` + `ellipsis`. This is actually visible in your own mobile screenshot: the header reads **`Thu, Jul 16, 20…`** with the year cut off. Week view survives the same width only because `_isVeryCompact()` drops the date range down to `W29`.

It gets tighter than the screenshot suggests, because the PR grows the toggle group from 80px to 120px (icon buttons are 40px), which takes 40px from the title in *every* view — and anyone with a calendar provider configured gets a 4th button from `showCalFilterBtn()`. At 360px with a provider there's very little title left.

Degrading like the week branch does — drop the year under XS (600), the weekday under XXS (420) — would cover it.

## 2. Same line: the date format is en-US-only

`'EEE, MMM d, yyyy'` pins US field *ordering* and only translates the names. What other locales actually want:

```
en-US   Tue, Jul 14, 2026
de-DE   Di., 14. Juli 2026      <- pattern gives "Di., Juli 14, 2026"
fr-FR   mar. 14 juil. 2026
sv-SE   tis 14 juli 2026
ja-JP   2026年7月14日(火)        <- year first
```

We've had a run of locale bugs recently (#8987), so I'd rather not add a new hardcoded pattern. A locale-aware format (Angular's named formats, or an `Intl` skeleton) fixes this at the same site as #1 — the existing `'MMM d'` / `'LLLL yyyy'` calls have the same flaw, but those are mine to clean up, not yours.

## 3. `_contextNow` freezes at midnight after one overnight (pre-existing — my call, not a demand)

This one is a pre-existing bug in `_contextNow`, but day view takes it from "needs 7 days of uptime" to "needs one night", so I want it on the record.

`_contextNow` keys off `_selectedDate !== null` and is never re-clamped when the day rolls over:

https://github.com/super-productivity/super-productivity/blob/39b8415b84f5670a3d51ae68f17cc80c420d76dd/src/app/features/schedule/schedule/schedule.component.ts#L234-L244

Repro: Day view → **Next Day** (sets `_selectedDate` to tomorrow 00:00) → leave the app open overnight. Next morning `isViewingToday()` is `true`, because it compares day *strings* — so **Prev and Today are both disabled** and `currentTimeRow()` draws the now-line. But `_selectedDate` is still that midnight, so `_contextNow` is 9 hours behind. I confirmed this with a throwaway spec: `contextNow=00:00`, `realNow=09:00`, drift 9h, while `isViewingToday()` returns true.

That flows into `create-schedule-days.ts` (`startTime = i == 0 ? now : dayStartTime`, and `getBudgetLeftForDay(..., i === 0 ? now : undefined)`), so today's unscheduled tasks lay out from 00:00 — above the now-line — and `msLeftToday(midnight)` hands back a full 24h budget, making the over-budget badge under-report. Recovery exists (Next→Prev snaps back to `null`, and leaving the `/schedule` route destroys the component) but nothing tells the user anything is wrong.

The guard wants to be on the leading day rather than `isViewingToday()`, since in week view today can sit at index > 0:

```ts
private _contextNow = computed(() => {
  const selectedDate = this._selectedDate();
  if (selectedDate === null) return Date.now();
  if (this.daysToShow()[0] === this._todayDateStr()) return Date.now();
  const contextDate = new Date(selectedDate);
  contextDate.setHours(0, 0, 0, 0);
  return contextDate.getTime();
});
```

Fold it in if you're up for it; otherwise say so and I'll take it as a follow-up. It's genuinely not a defect you introduced.

## Nits

- `.day-view-btn` isn't used by any SCSS — it exists only for the unit-test selector. Convention here is the `e2e-` prefix for test hooks.
- `.week-month-selector` has no `role="group"`, so the three toggles are announced unrelated. Pre-existing (it was 2 before), and three mutually-exclusive options are really a radiogroup — `mat-button-toggle-group` would be the native fit and would let the local `.week-month-btn` Material restyle go away. Explicitly **not** this PR; noting it so I don't forget.
- Switching `aria-pressed` on the Week button from `!isMonthView()` to `isWeekView()` was a necessary correctness fix, not cosmetic — the old expression would have reported Week as pressed while in Day view. Good catch.

## Not yours — I'll file these separately

Both confirmed while reviewing, both pre-existing:

- `goToPreviousPeriod` in month view sets `_selectedDate` to the 1st of the previous month with no snap-to-null guard. So month → Next → Prev leaves `_selectedDate` at the 1st while `isViewingToday()` reads true; switching to day view then shows a date 15 days in the past. I verified week view is affected identically, so this isn't something day view introduces.
- `getDaysToShow` uses `+24h` millisecond arithmetic, so a DST fall-back duplicates a day — `TZ=Europe/Berlin` from 2025-10-26 yields `['2025-10-26','2025-10-26','2025-10-27',...]`, only 6 distinct days. Day view is immune, since the loop runs once.

## On scope

For the record, since I'm usually strict about feature creep: this passes cleanly. No new setting, no new component, no dependency, no sync surface; the default stays `week` and it's opt-in through a control that already exists. It isn't redundant with `schedule-day-panel` either — that one is desktop-only, right-panel, hard-wired to today with no navigation, and exists mainly as a drag-drop target. `SELECTED_TIME_VIEW='day'` also degrades safely on older clients (unknown → `week`), which I checked.

I did consider whether week view collapsing to one day on narrow screens would be the calmer answer, since it'd fix mobile with zero new UI — but that imposes a change on people who want the scrollable week, and your desktop motivation (overlapping events squeezed into narrow columns) stands on its own. The toggle is the right call.

Fix #1 and #2 and I'll merge. Thanks again.

johannesjo:
Following up on the `_contextNow` overnight drift I mentioned in my review — I've taken that one myself so it isn't on your plate: #9064.

It's pre-existing and independent (this PR doesn't touch `_contextNow`, and the two merge cleanly in either order), but it's worth landing first: in week view the bug needs ~7 consecutive overnights before the shown range's first day becomes today, which is why nobody hit it. A single-day view reaches it after one.

The day header points (compact form + locale-aware format) are still the ones I'd like from you here.

jkusa:
Thanks for the detailed review. Pushed fixes for `#1` and `#2`: the day branch of `headerTitle()` now uses a locale-aware Intl skeleton and shrinks responsively so it can't ellipsis-clip.

On the breakpoint, I landed on `TABLET` (768) for the year rather than`XS` (600). At `XS` the full form still clips: it needs ~680px once the toggle group and nav controls take their space. Below `TABLET` it drops to a compact month + day, with the weekday still shown in the day-column sub-header. Verified clip-free from 360 to 1280, and glad to move it if you'd rather.

Also renamed the hook to `.e2e-day-view-btn`, and left `_contextNow` for your #9064.

There's a small separate commit renaming `week-month-*` to `time-view-*` (three buttons now), easy to drop if it collides with the `mat-button-toggle-group` change.

Thanks again.

day-view-roomy
mobile-day-view

johannesjo:

johannesjo:

PR #9059, opened 2026-07-16
fix(api): map plannedAt to dueWithTime

### LINKED ISSUE #8951: [Bug] Local REST API: plannedAt field not working, dueWithTime used for scheduling instead
## Bug Report

**Environment:**
- Super Productivity version: latest
- Platform: Desktop (Electron)
- API: Local REST API (port 3876)

## Issue

The Local REST API's `plannedAt` field (documented as the scheduling/planning date for tasks) does not work when creating or updating tasks via `POST /tasks` and `PATCH /tasks/:id`.

Instead, the `dueWithTime` field (which according to docs is for "due date with time") is actually being used by the UI as the **scheduling/planning date** (shows in Today view, Schedule view, Planner view).

## Expected Behavior

According to the [API documentation](https://github.com/super-productivity/super-productivity/wiki/3.01-API#task-endpoints):
- `plannedAt` -- should be the scheduling/planning date (Unix timestamp ms)
- `dueWithTime` -- should be the deadline with time
- `dueDay` -- should be the deadline day only

## Actual Behavior

- `plannedAt` is accepted but **ignored** - tasks don't appear in Today/Schedule/Planner views
- `dueWithTime` **actually controls** the scheduling - tasks appear in Today/Schedule/Planner based on this value
- This makes it impossible to separately set a **deadline** and a **planned/scheduled date**

## Workaround

Currently using `dueWithTime` for the scheduling/planning date (`@schedule:` in Obsidian plugin), which means:
- Can't set a true deadline separate from planned date
- The `@due:` syntax (for deadline) is effectively unusable via the API

## Steps to Reproduce

1. Enable Local REST API in SP Settings -> Misc
2. Create a task with `plannedAt` set to tomorrow:
   ```bash
   curl -X POST http://127.0.0.1:3876/tasks \
     -H "Content-Type: application/json" \
     -d '{"title": "Test plannedAt", "plannedAt": 1721000000000}'
   ```
3. Observe task doesn't appear in Today/Schedule/Planner views
4. Create a task with `dueWithTime` set to tomorrow:
   ```bash
   curl -X POST http://127.0.0.1:3876/tasks \
     -H "Content-Type: application/json" \
     -d '{"title": "Test dueWithTime", "dueWithTime": 1721000000000}'
   ```
5. Observe task appears in Schedule/Planner for that date

## Impact

This affects any external integration (Obsidian plugin, custom scripts, etc.) that wants to use the SP API to set planned/scheduled dates separately from deadlines.

---

**Note**: This issue is tracked in the obsidian-super-productivity-sync plugin as a known limitation. The plugin currently maps `@schedule:` to `dueWithTime` as a workaround.

cyphercodes:
## Problem

Fixes #8951.

The local REST API allowed `plannedAt`, but passed it through as a legacy field instead of scheduling tasks via `dueWithTime`.

## Solution

Map `plannedAt` to `dueWithTime` for task create/update requests, while preserving explicit `dueWithTime` when both fields are provided. `plannedAt: null` clears `dueWithTime` on updates.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

## Tests

- `NODE_OPTIONS=--max-old-space-size=4096 npm run test:file src/app/core/electron/local-rest-api-handler.service.spec.ts -- --watch=false --browsers=ChromeHeadless`
- `NODE_OPTIONS=--max-old-space-size=4096 npm run checkFile src/app/core/electron/local-rest-api-handler.service.ts`
- `NODE_OPTIONS=--max-old-space-size=4096 npm run checkFile src/app/core/electron/local-rest-api-handler.service.spec.ts`
- `git diff --check`

### DIFF
--- a/docs/wiki/3.01-API.md
+++ b/docs/wiki/3.01-API.md
@@ -311,7 +311,7 @@ curl -X POST http://127.0.0.1:3876/tasks/task-id/archive
 | `title` (required)   | Non-empty string                                                                                                                                                                                                                                                                                  |
 | `parentId`           | Create the task as a subtask of this top-level task. Returns 404 if parent missing, 400 if parent is itself a subtask. The new subtask inherits the parent's `projectId` and never has its own tags — supplying `projectId` or `tagIds` together with `parentId` returns 400 `UNSUPPORTED_FIELD`. |
 | `subTaskIds`         | Not supported on create — returns 400 `UNSUPPORTED_FIELD`. Create the parent first, then create each child with `parentId`.                                                                                                                                                                       |
-| other allowed fields | `notes`, `isDone`, `timeEstimate`, `timeSpent`, `projectId`, `tagIds`, `dueDay`, `dueWithTime`, `plannedAt`                                                                                                                                                                                       |
+| other allowed fields | `notes`, `isDone`, `timeEstimate`, `timeSpent`, `projectId`, `tagIds`, `dueDay`, `dueWithTime`; deprecated alias: `plannedAt` maps to `dueWithTime` when `dueWithTime` is not supplied.                                                                                                           |
 
 **PATCH /tasks/:id — restricted fields:**
 
--- a/src/app/core/electron/local-rest-api-handler.service.spec.ts
+++ b/src/app/core/electron/local-rest-api-handler.service.spec.ts
@@ -585,6 +585,22 @@ describe('LocalRestApiHandlerService', () => {
         });
       });
 
+      it('should return 400 with plannedAt in the error path when legacy plannedAt has an invalid type', async () => {
+        const response = await sendRequestAndWait(
+          createRequest('POST', '/tasks', {
+            body: { title: 'New Task', plannedAt: 'not-a-number' },
+          }),
+        );
+
+        expect(response.body.ok).toBe(false);
+        expect(response.status).toBe(400);
+        expect((response.body as any).error.code).toBe('INVALID_INPUT');
+        expect((response.body as any).error.details).toContain(
+          jasmine.objectContaining({ path: '$input.plannedAt' }),
+        );
+        expect(taskServiceMock.add).not.toHaveBeenCalled();
+      });
+
       it('should return 400 when an allowed field has an invalid type', async () => {
         const response = await sendRequestAndWait(
           createRequest('POST', '/tasks', {
@@ -824,6 +840,23 @@ describe('LocalRestApiHandlerService', () => {
         });
       });
 
+      it('should let dueWithTime override plannedAt when updating a task', async () => {
+        const mockTask = createMockTask('task-1');
+        Object.defineProperty(taskServiceMock, 'getByIdOnce$', {
+          get: () => (_id: string) => of(mockTask),
+        });
+
+        await sendRequestAndWait(
+          createRequest('PATCH', '/tasks/task-1', {
+            body: { plannedAt: 1, dueWithTime: 2 },
+          }),
+        );
+
+        expect(taskServiceMock.update).toHaveBeenCalledWith('task-1', {
+          dueWithTime: 2,
+        });
+      });
+
       it('should map null plannedAt to clearing dueWithTime when updating a task', async () => {
         const mockTask = createMockTask('task-1', { dueWithTime: 1 });
         Object.defineProperty(taskServiceMock, 'getByIdOnce$', {
@@ -839,6 +872,27 @@ describe('LocalRestApiHandlerService', () => {
         expect(taskServiceMock.update).toHaveBeenCalledWith('task-1', {
           dueWithTime: null,
         });
+      });
+
+      it('should return 400 with plannedAt in the error path when legacy plannedAt has an invalid type on PATCH', async () => {
+        const mockTask = createMockTask('task-1');
+        Object.defineProperty(taskServiceMock, 'getByIdOnce$', {
+          get: () => (_id: string) => of(mockTask),
+        });
+
+        const response = await sendRequestAndWait(
+          createRequest('PATCH', '/tasks/task-1', {
+            body: { plannedAt: 'not-a-number' },
+          }),
+        );
+
+        expect(response.body.ok).toBe(false);
+        expect(response.status).toBe(400);
+        expect((response.body as any).error.code).toBe('INVALID_INPUT');
+        expect((response.body as any).error.details).toContain(
+          jasmine.objectContaining({ path: '$input.plannedAt' }),
+        );
+        expect(taskServiceMock.update).not.toHaveBeenCalled();
       });
 
       it('should reject parentId in PATCH body with 400', async () => {
--- a/src/app/core/electron/local-rest-api-handler.service.spec.ts
+++ b/src/app/core/electron/local-rest-api-handler.service.spec.ts
@@ -543,6 +543,48 @@ describe('LocalRestApiHandlerService', () => {
         });
       });
 
+      it('should map plannedAt to dueWithTime when creating a task', async () => {
+        const plannedAt = new Date('2026-07-12T10:00:00Z').getTime();
+        Object.defineProperty(taskServiceMock, 'getByIdOnce$', {
+          get: () => (_id: string) => of(createMockTask('new-task-id')),
+        });
+
+        await sendRequestAndWait(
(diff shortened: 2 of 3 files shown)


johannesjo:
Thanks for this! The direction is right and the tests are good. `plannedAt → dueWithTime` is exactly the mapping the backup migration already performs ([`migrate-legacy-backup.ts#L586-L588`](https://github.com/super-productivity/super-productivity/blob/master/src/app/op-log/backup/migrate-legacy-backup.ts#L586-L588), "Migration 4"), so this is consistent with how the rename is handled everywhere else. `plannedAt` simply no longer exists on the `Task` model, which is why passing it through was never going to work.

Worth noting the PR also fixes a bug it doesn't mention: `createNewTaskWithDefaults` spreads `...additional` unfiltered (`task.service.ts#L1491`) and `typia.assert<Task>` tolerates excess properties, so `POST /tasks {"plannedAt": 123}` previously wrote a junk `plannedAt` field onto the task entity — which then went into the op-log and synced out to other devices. The `delete` stops that. Relatedly, because the mapping happens *before* `add()`, `createNewTaskWithDefaults` now sees `dueWithTime` and correctly skips defaulting `dueDay` (`task.service.ts#L1485-L1489`), so dueDay/dueWithTime mutual exclusivity holds on the create path. Nice.

Two things I'd like changed before merge:

**1. `plannedAt` in `WritableTaskFields` is now dead code, and the widening to `number | null` has no effect.** `pickAllowedFields` deletes `plannedAt` before `validateWritableFields` ever sees it, so that property is unreachable. The user-visible consequence — I verified this against the actual suite:

```
POST /tasks {"title": "x", "plannedAt": "not-a-number"}
→ 400 {"path": "$input.dueWithTime", "expected": "(null | number | undefined)"}
```

The caller sent `plannedAt` and the error names a field they never mentioned. Suggest validating *before* mapping (keeping `plannedAt: number | null` in `WritableTaskFields` where it's actually reachable), then mapping afterwards. A test for a bad-typed `plannedAt` would lock this in — and precedence is currently only tested on create, not on PATCH.

**2. Docs.** The checklist item is unchecked, and [`docs/wiki/3.01-API.md#L314`](https://github.com/super-productivity/super-productivity/blob/master/docs/wiki/3.01-API.md) still lists `plannedAt` as a plain "other allowed field" right next to `dueWithTime`, with no hint they're now synonyms. Two undocumented aliases for one field is worse than either alternative — please mark it a deprecated alias so new integrations don't build on it.

---

**On #8951 more broadly** — this fixes the issue title but not its "Impact" section, and I think the report's premise is inverted. In the current model `dueWithTime` **is** the scheduled time (`task.model.ts#L93-L103`), and deadlines are separate fields: `deadlineDay` / `deadlineWithTime` (`task.model.ts#L118-L131`). Those aren't in `ALLOWED_TASK_FIELDS`, so after this PR you still can't set a deadline via the API — which was the reporter's actual blocker ("can't set a true deadline separate from planned date").

Adding those two fields is a small change in the same file following the same pattern. Happy either way — fold it in here, or land this as-is and I'll track the deadline fields separately. Your call; no obligation to expand the scope of your PR.

**Pre-existing, not caused by this PR** (recording it so it isn't lost, not asking you to fix it): the REST API's raw `TaskService.update()` bypasses the `scheduleTaskWithTime` meta-reducer, so `PATCH {"plannedAt": <ts>}` on a task that already has `dueDay` set leaves *both* set, violating the invariant `task.model.ts#L97-L99` calls CRITICAL. Selectors tolerate it (`dueWithTime` wins) and Today membership is virtual so the task still shows up — just unordered at the bottom rather than positioned via `TODAY_TAG.taskIds`. Similarly `plannedAt: null` clears `dueWithTime` but leaves `remindAt`, orphaning a reminder on a now-unscheduled task. All of this is already reachable today by sending `dueWithTime` directly, so it's not a regression — this PR just adds a second door to it.

Verified locally: 73/73 in `local-rest-api-handler.service.spec.ts` pass on this branch.

cyphercodes:
Implemented the requested follow-up in d402e6b:

- validate writable fields before mapping legacy `plannedAt` to `dueWithTime`, so bad legacy inputs report `$input.plannedAt`
- kept `dueWithTime` precedence for both create and PATCH
- documented `plannedAt` as a deprecated alias for `dueWithTime`

Verification:
- `npm ci`
- `NODE_OPTIONS='--max-old-space-size=4096' npx ng test --watch=false --no-code-coverage --source-map=false --include='src/app/core/electron/local-rest-api-handler.service.spec.ts'` → 76/76 passed
- `npx prettier --check src/app/core/electron/local-rest-api-handler.service.ts src/app/core/electron/local-rest-api-handler.service.spec.ts docs/wiki/3.01-API.md`
- `git diff --check`

Note: the default full `ng test --watch=false` path OOMs in this cron container without the larger Node heap, so I used the focused spec with `NODE_OPTIONS` above.

johannesjo (commented):
Follow-up to my review above: d402e6b0d9 addresses both change requests, and I need to correct one thing I got wrong.

## Both requested changes check out

I didn't just re-run the suite — I mutation-tested the new specs to confirm they actually pin the behaviour rather than passing incidentally:

| Sabotage | Result |
|---|---|
| Pass `plannedAt` through unmapped (pre-fix behaviour) | 5 failures |
| Remove only the `!hasOwnProperty('dueWithTime')` precedence guard | exactly the 2 override tests |
| Revert `plannedAt?: number \| null` → `number` | 1 failure |
| Revert d402e6b0d9 (map before validate) | 1 failure |

Pristine: 76/76 green, `checkFile` clean on both files, merges into master with no file overlap.

The last two rows resolve my point 1: with validation now running before the mapping, `plannedAt` in `WritableTaskFields` is reachable and the `number | null` widening is load-bearing rather than dead code. `POST /tasks {"plannedAt": "not-a-number"}` now reports `$input.plannedAt`, and PATCH precedence is covered.

## Correction: adding the deadline fields is *not* a small same-pattern change

I wrote that "adding those two fields is a small change in the same file following the same pattern". That was wrong, and I'd rather flag it than have you act on it — please **don't** fold it into this PR.

`dueDay`/`dueWithTime` are plain entity fields the REST allow-list can write directly. Deadlines are not — they have a dedicated action and meta-reducer:

- `TaskSharedActions.setDeadline` ([`task-shared.actions.ts#L213`](https://github.com/super-productivity/super-productivity/blob/master/src/app/root-store/meta/task-shared.actions.ts#L213)) → `handleSetDeadline` ([`task-shared-deadline.reducer.ts#L97`](https://github.com/super-productivity/super-productivity/blob/master/src/app/root-store/meta/task-shared-meta-reducers/task-shared-deadline.reducer.ts#L97)), which validates formats, enforces `deadlineDay`/`deadlineWithTime` mutual exclusivity (#L130-L131) and sets `deadlineRemindAt`.
- It then calls `autoPlanTaskDueToDeadline` (#L30) which — the part that matters — can rewrite the task's **scheduling** state: update `dueDay`, clear `dueWithTime`, clear `remindAt`, and touch `TODAY_TAG.taskIds`.
- There's no `TaskService` wrapper; it's dispatched straight from the deadline dialogs and the reminder module.

So adding `deadlineDay`/`deadlineWithTime` to `ALLOWED_TASK_FIELDS` would route them through plain `updateTask` and silently skip all of that: no mutual-exclusivity clearing, no `deadlineRemindAt`, no auto-planning. A deadline set via REST would behave differently from every deadline set in the UI. That's new invariant debt of exactly the kind I recorded as pre-existing at the bottom of my review — and since it's a multi-entity change it belongs behind the meta-reducer, not the allow-list.

Doing it properly needs a `TaskService.setDeadline()` wrapper plus a decision on whether the REST path should auto-plan. That's its own PR, and I'll track it separately.

**So: land this as-is.** Please just change `Fixes #8951` → `Refs #8951` in the description so the issue stays open for the deadline work — the reporter's actual blocker ("can't set a true deadline separate from planned date") is that follow-up, not this PR.

## One docs nit

The deprecation note landed in the POST table, but the alias works on PATCH too, and `plannedAt: null` → clears `dueWithTime` (tested, and called out in your description) isn't documented anywhere. Worth a line in the PATCH section.

PR #9077, opened 2026-07-16
fix: harden plugin SVG icon sanitization paths

ghostkeni:
## Problem

Uploaded plugin icons were accepted with only minimal `<svg>` shape checks, then passed into trusted HTML sinks. This left a path for malicious SVG payloads to execute in renderer context via plugin icon rendering/registration.

## Solution

Added a shared SVG icon sanitizer and applied it at every plugin icon trust boundary:

- sanitize uploaded icon content before storing/using it in `PluginService`
- sanitize before icon registration in `GlobalThemeService`
- sanitize before inline icon rendering in `PluginIconComponent`
- tighten URI handling in SVGs to fragment-only references for icon use (`#...`)
- add focused unit coverage for sanitizer behavior and plugin integration paths
- add dependency overrides in `package.json`/`package-lock.json` to reduce vulnerable `esbuild` exposure in the audited toolchain where possible

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [x] Other (please describe)

Other: security hardening and dependency risk reduction

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27535,6 +27535,490 @@
         }
       }
     },
+    "node_modules/vite/node_modules/@esbuild/aix-ppc64": {
+      "version": "0.28.1",
+      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.28.1.tgz",
+      "integrity": "sha512-Svl7tq8k/08+p6CXPpRjQ1fKX+1odH/BQbb48fV6fj3CWHhsoIOoY87w1oHXm0qEpkIK3ZfVgp0hed3XBXzXMQ==",
+      "cpu": [
+        "ppc64"
+      ],
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "os": [
+        "aix"
+      ],
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "node_modules/vite/node_modules/@esbuild/android-arm": {
+      "version": "0.28.1",
+      "resolved": "https://registry.npmjs.org/@esbuild/android-arm/-/android-arm-0.28.1.tgz",
+      "integrity": "sha512-0k2F129Xdio1TdJfzJ8sy1Q47vUD2NnwdhiAf7drUN1EBTfPf4hsFCtmMgu/6m8JSzsBrlmVjudMBQqOfG8usQ==",
+      "cpu": [
+        "arm"
+      ],
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "os": [
+        "android"
+      ],
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "node_modules/vite/node_modules/@esbuild/android-arm64": {
+      "version": "0.28.1",
+      "resolved": "https://registry.npmjs.org/@esbuild/android-arm64/-/android-arm64-0.28.1.tgz",
+      "integrity": "sha512-34EGEbCIAgosYz6goLcopX6Mo7NyGv9tfwEM2/7Ce2VcVRk568iSvniGWcUXIy7wEDR1wzolcxcriFVrWYcwBg==",
+      "cpu": [
+        "arm64"
+      ],
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "os": [
+        "android"
+      ],
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "node_modules/vite/node_modules/@esbuild/android-x64": {
+      "version": "0.28.1",
+      "resolved": "https://registry.npmjs.org/@esbuild/android-x64/-/android-x64-0.28.1.tgz",
+      "integrity": "sha512-dbwY7ltSMDWsRatcRpCnES4F+im88OCUgGZjy52shC7GqHRE/cYlxNbB4Z4UpJswpcc4Qxd2oE/ufM0p61IKng==",
+      "cpu": [
+        "x64"
+      ],
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "os": [
+        "android"
+      ],
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "node_modules/vite/node_modules/@esbuild/darwin-arm64": {
+      "version": "0.28.1",
+      "resolved": "https://registry.npmjs.org/@esbuild/darwin-arm64/-/darwin-arm64-0.28.1.tgz",
+      "integrity": "sha512-TZbWkQY7kvTAXbXUT7uVACR5cMHsDiSz9z7ZKAX/RTq/WJEk3QyRr0wZpNhBDX+/0CtdqUIJlOiodQcta6tY3Q==",
+      "cpu": [
+        "arm64"
+      ],
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "os": [
+        "darwin"
+      ],
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "node_modules/vite/node_modules/@esbuild/darwin-x64": {
+      "version": "0.28.1",
+      "resolved": "https://registry.npmjs.org/@esbuild/darwin-x64/-/darwin-x64-0.28.1.tgz",
+      "integrity": "sha512-zfdzgK9ACBNZLI/CyHTOx81SyNbM6YXn7rxSgX97VjyiPl9W1i4Ka4fgKECEoFCKGpvBj5qArWIGgQjOwkgskQ==",
+      "cpu": [
+        "x64"
+      ],
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "os": [
+        "darwin"
+      ],
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "node_modules/vite/node_modules/@esbuild/freebsd-arm64": {
+      "version": "0.28.1",
+      "resolved": "https://registry.npmjs.org/@esbuild/freebsd-arm64/-/freebsd-arm64-0.28.1.tgz",
+      "integrity": "sha512-wG2EA8ENdEI0qhkSZMjfqrdY+ziCYCPMmtZjjIwOmXFjmyzEHn+UUxk5of+SYsjtfs3VpnlC7QLzSI5hY/rOAw==",
+      "cpu": [
+        "arm64"
+      ],
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "os": [
+        "freebsd"
+      ],
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "node_modules/vite/node_modules/@esbuild/freebsd-x64": {
+      "version": "0.28.1",
+      "resolved": "https://registry.npmjs.org/@esbuild/freebsd-x64/-/freebsd-x64-0.28.1.tgz",
+      "integrity": "sha512-i7dZ9vQgnvSCzi/rYCXNgtF/U+eKZNJBzu3eTQbRgHnM7tNSizLOkRFAl3qzVc/Op/u5YkHHa4pf/3DOYHthLQ==",
+      "cpu": [
+        "x64"
+      ],
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "os": [
+        "freebsd"
+      ],
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "node_modules/vite/node_modules/@esbuild/linux-arm": {
+      "version": "0.28.1",
+      "resolved": "https://registry.npmjs.org/@esbuild/linux-arm/-/linux-arm-0.28.1.tgz",
+      "integrity": "sha512-qVXBOHQS+d5Y722GwJzJUtOLlX7km3CraOaGormF1pDtPd2C/l1SHRPgjLunLGe51Sh5YYWKMFDyV4SxgMQYTQ==",
+      "cpu": [
+        "arm"
+      ],
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "os": [
+        "linux"
+      ],
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "node_modules/vite/node_modules/@esbuild/linux-arm64": {
+      "version": "0.28.1",
+      "resolved": "https://registry.npmjs.org/@esbuild/linux-arm64/-/linux-arm64-0.28.1.tgz",
+      "integrity": "sha512-yHs+0uc8+nvEAfAfxrWQKK5peSNzBc4PegcMO0EJ2hT71uA7vB8Ihg2e77R2P7SG5uYjPbHlLLmve4LLLRCf0g==",
+      "cpu": [
+        "arm64"
+      ],
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "os": [
+        "linux"
+      ],
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "node_modules/vite/node_modules/@esbuild/linux-ia32": {
+      "version": "0.28.1",
+      "resolved": "https://registry.npmjs.org/@esbuild/linux-ia32/-/linux-ia32-0.28.1.tgz",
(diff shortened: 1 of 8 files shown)


johannesjo:
Thanks for digging into this, and welcome — you found a **real gap**, and it's one worth closing. Unfortunately I can't merge this as-is: I verified the sanitizer against the PR's actual code in real Chrome and found **five working bypasses that execute arbitrary JS**. I'd like to help get this landed properly, so here's everything I checked.

## The core issue: a parse differential

`sanitizeSvgIconContent()` parses input as `image/svg+xml` (strict XML) and re-serialises with `XMLSerializer`. But **both sinks re-parse that output as HTML**:

- `MatIconRegistry._svgElementFromString()` → `div.innerHTML = str` (`@angular/material/fesm2022/_icon-registry-chunk.mjs:203-205`)
- `PluginIconComponent` → `[innerHTML]="svg"`

So the sanitizer audits an XML document, but a *different* document — built by the far more lenient HTML parser — is what ships. Anything that survives XML parsing and means something different in HTML gets through.

### Confirmed bypasses

**1. CDATA smuggled through `<desc>` / `<title>`.** These are HTML *integration points*: inside them the HTML parser switches back to HTML content mode. `XMLSerializer` faithfully re-emits `<![CDATA[…]]>`, which the HTML parser then treats as a **bogus comment ending at the first `>`** — releasing the rest as live markup.

```xml
<svg xmlns="http://www.w3.org/2000/svg"><desc><![CDATA[><img src=x onerror="alert(1)">]]></desc></svg>
```

Sanitizer output is unchanged from the input; after `innerHTML` the DOM contains a live ``. With a `data:` URI as `src` (so the load always succeeds) the handler **fires**.

**2. Case-sensitive denylist vs case-insensitive HTML parser.** `querySelectorAll()` on an **XML** document matches type selectors *case-sensitively*, so `DANGEROUS_SVG_TAGS` never matches `<FOREIGNOBJECT>` or `<STYLE>`. The HTML parser at the sink then applies its SVG tag-name adjustment table and resurrects them as a real `<foreignObject>` / `<style>`:

```xml
<svg xmlns="http://www.w3.org/2000/svg"><FOREIGNOBJECT><![CDATA[><img src=x onerror="alert(1)">]]></FOREIGNOBJECT></svg>
```

**3. SMIL retargeting `href` — no parse trick needed at all.** The element carries no `href`/`src` of its own, so the `URI_ATTRS` check never inspects it. `<set>`/`<animate>` writes the `javascript:` URL onto the parent `<a>` **at runtime, after sanitisation**. Plugin icons are clickable in the nav, so this is reachable:

```xml
<svg xmlns="http://www.w3.org/2000/svg"><a><set attributeName="href" to="javascript:alert(1)"/><text>x</text></a></svg>
```

I confirmed all five execute (`window.__PWN` set) in headless Chrome via the real sinks.

### Why this isn't a three-line fix

These aren't three bugs to patch — they're symptoms of two structural choices:

- A **denylist** can only ever block what was enumerated. Patch these three and there's a fourth.
- The sanitiser **parses a different document than the one that ships**. As long as that differential exists, the audit is inspecting the wrong artifact.

## Severity — worth stating accurately

I traced the threat model, and it cuts both ways.

Against a *lower* severity: plugin code already runs via `new Function` in the main renderer (`plugin-runner.ts:172`), and the app CSP is `script-src *` with `'unsafe-inline'`/`'unsafe-eval'`. For an **enabled** plugin, icon XSS is no privilege escalation — it already had everything.

But there's a genuinely distinct exposure, and it's sharper than the PR description claims. In `_discoverUploadedPlugins()`, `isEnabled` is read and stored into state — but `_registerPluginIcon()` is then called **unconditionally** whenever a cached icon exists (`plugin.service.ts:325-327`), with `status: 'not-loaded'`:

```ts
if (cachedPlugin.icon) {
  this._pluginIcons.set(cachedPlugin.id, cachedPlugin.icon);
  this._registerPluginIcon(cachedPlugin.id, cachedPlugin.icon);   // no isEnabled gate
```

And I verified `MatIconRegistry`'s div is **never appended to the document** — yet `innerHTML` alone starts the image load, so the handler fires from the detached div. Net effect:

> **A disabled, never-loaded uploaded plugin executes JS in the main renderer at app startup, with nothing rendered.**

That's the bug worth fixing, and I'd lead the PR description with it.

## The fix I'd suggest

I prototyped this and validated it against all five payloads plus a benign icon — all six behave correctly, with **zero payload-specific special-casing**:

1. **Parse in the sink's own mode.** Use an inert `<template>` (HTML parsing, no loads, no side effects) so the sanitiser audits exactly the document that ships. This eliminates the differential outright, killing #1 and #2.
2. **Allowlist elements and attributes** instead of denylisting. The ~20 shape tags an icon actually needs (`svg`, `g`, `path`, `circle`, `rect`, `polygon`, `defs`, `linearGradient`, `stop`, …). `<set>`, `<animate>`, `<FOREIGNOBJECT>`, `<STYLE>` all die for free — that's #3, without ever having to think about SMIL.
3. **Stability loop** — re-parse until fixed point, bailing if it never stabilises. This is exactly what Angular's own sanitiser does (`_debug_node-chunk.mjs:3183-3198`) and it's the standard mXSS backstop.

Benign icons round-trip intact (`viewBox`, `fill="currentColor"` and paths preserved), so `mat-icon` theming keeps working.

Two dead ends I checked so you don't have to:

- **`DomSanitizer.sanitize(SecurityContext.HTML, …)` won't work.** Angular's `VALID_ELEMENTS` contains no SVG tags, so it strips `<svg>` entirely and `mat-icon` then throws `<svg> tag not found`. The `bypassSecurityTrustHtml` calls are genuinely necessary here.
- **DOMPurify is out** under the repo's no-new-dependencies rule (`CLAUDE.md` → Project rules).

So a hand-rolled sanitiser really is the constraint — it just needs to be an allowlist in the sink's parse mode.

## Tests

The five new specs assert on the XML **string** (`toContain('<svg')`, `not.toContain('onload')`) rather than on what the DOM builds at the sink — so they structurally cannot observe this bug class. Each also only covers a payload the implementation already handles, which is why they're green.

Suggested shape: sanitize, then render through the real sink and assert on the **DOM**:

```ts
const renderAtSink = (sanitized: string): HTMLDivElement => {
  const div = document.createElement('DIV');
  div.innerHTML = sanitized;   // what both call sites do
  return div;
};

it('does not smuggle live markup through <desc> CDATA', () => {
  const sanitized = sanitizeSvgIconContent(
    `<svg xmlns="http://www.w3.org/2000/svg"><desc><![CDATA[><img src=x onerror="window.alert(1)">]]></desc></svg>`,
  );
  if (sanitized === null) return;   // rejecting outright is a valid fix

  const rendered = renderAtSink(sanitized);
  expect(rendered.querySelector('img')).toBeNull();
  expect(rendered.querySelector('[onerror]')).toBeNull();
});
```

I have a full 6-case regression spec covering every bypass above, written so each case passes if the sanitizer **either** rejects the input **or** neutralises it — so it works as a gate against any fix, not just mine. Happy to drop it in here or as a follow-up, whichever you prefer.

## Please split out the esbuild change

The dependency work is unrelated to plugin SVG hardening and is **73% of the diff** (491 of 669 lines). On its merits:

- It's [GHSA-g7r4-m6w7-qqqr](https://github.com/advisories/GHSA-g7r4-m6w7-qqqr) — **CVSS 2.5, low**, dev-server-on-Windows arbitrary file read. Dev-only, and not related to this PR's threat model.
- It's self-redundant: the global `"esbuild": "0.28.1"` override already covers the `vite` and `tsup` entries added alongside it.
- It's mildly risky: `vite` and `tsup` both declare `esbuild: "^0.27.0"`, which **excludes** `0.28.1` (verified with semver — for `0.x`, caret pins the minor). The global override forces them onto a minor they don't declare support for.

A separate PR would let it be judged on its own.

---

To be clear about where this stands: the underlying issue is real and worth fixing, and the four trust boundaries you identified are the right ones. It's the sanitiser itself that needs to change shape — allowlist, parsed in the sink's mode. That's a smaller change than what's here now, and it's strictly more robust. Happy to review the next round, and glad to hand over the repro spec to work against.

johannesjo:
@ghostkeni following up on my July review: no reply since, and the branch has gone conflicting, so I have finished this off myself rather than let a real vulnerability sit open. The finding is yours and the credit for it is yours; what changed is the implementation.

**PR: #9440** replaces this one. I have kept your four trust boundaries, which were the right ones.

## What is different

You will remember the core problem: the sanitizer parsed `image/svg+xml` and re-serialized with `XMLSerializer`, but both sinks re-parse as HTML, so it audited a document that never shipped. Five payloads executed JS through that gap.

The replacement is the allowlist design I sketched, with one deliberate deviation from what I wrote in July.

**Parse mode.** Input is parsed with `DOMParser` as `text/html` rather than as XML, which kills the CDATA and case-folding tricks at the source. I suggested an inert `<template>`; `DOMParser` text/html documents are equally inert (no browsing context, so no scripts and no sub-resource loads) and it is what Angular's own sanitizer prefers. Either works; this one is less code. See the note below on why "same parse mode as the sink" turned out to be too strong a claim.

**Allowlist.** 19 shape tags and the presentation attributes an icon actually draws with. `<a>`, `<set>`, `<animate>`, `<foreignObject>`, `<STYLE>` and friends are gone by construction, with no payload-specific special-casing.

**No stability loop, and this is the deviation.** In July I said to add one, matching Angular's `_sanitizeHtml`. I did not, and I want to be explicit about why rather than quietly drop it. Angular needs that loop because it re-reads `inertBodyElement.innerHTML`, which is the *browser's* serialization of an attacker-controlled tree and can be unstable. This sanitizer never round-trips through `innerHTML`: it walks the parsed tree and builds the output string itself, emitting only allowlisted tags with every text and attribute value entity-encoded. That removes the source of the instability rather than iterating against it, so the loop would be a branch that cannot fire. Idempotence is asserted in the specs across all the bypass payloads instead, which is the honest place for an invariant that holds by construction.

## What an independent review pass caught afterwards

Since I had just told you your sanitizer was broken, it seemed only fair to have mine attacked
just as hard before proposing it. Two adversarial passes ran against it, and they landed.

**Two real bypasses in my own first version.** Both external resource loads, no script execution:

- **A CSS identifier escape walks past a regex looking for `url(`.** `\75 rl(` is how CSS spells
  `url(`, and the parser resolves the escape before it decides the token type. So
  `fill="\75 rl(https://evil.test/x.svg#g)"` sailed through untouched, and Chrome resolved it:
  `getComputedStyle(rect).fill` read back as `url("https://evil.test/x.svg#g")`, a live
  cross-origin GET on every icon render. Not blocked by our CSP, and reachable from a disabled
  plugin's icon at startup, which makes it a phone-home beacon against the project's
  no-tracking rule. Attribute values containing a backslash are now refused outright, because
  there is no safe way to regex over CSS that still has escapes in it.
- **`String.prototype.trim()` and the URL parser disagree about whitespace.** `trim()` strips
  U+00A0, the URL parser does not, so `href="<U+00A0>#x"` passed a `trim().startsWith('#')`
  fragment check and then resolved to a request off the document.

I fixed the second by deleting the feature rather than patching the check: `<use>`, `<symbol>`
and every `href` attribute are now out of the allowlist. That also closes a denial of service
the review turned up, where nested `<use>` expands exponentially at render time and a **1 KB
icon wedges the renderer for tens of seconds**. Gradients are unaffected, since those are
reached through `fill="url(#id)"` rather than through `href`.

**One overclaim in my own documentation.** I had written that parsing as `text/html` means
"there is no parse differential to exploit". Not quite true: `DOMParser` documents have
scripting disabled while the sinks' documents have it enabled, and the parser does observe the
difference (`<noscript>` nests differently). Not exploitable, because rebuilding the output
from scratch is what actually carries the safety, but the argument should not have rested on
the parse mode. Corrected in the file.

**Three functional findings**, two of which are behaviour you would have hit too:

- **An icon that sanitizes down to nothing was still returned as valid.** An embedded `<image>`,
  or Illustrator's `<switch>` wrapper, leaves an empty but well-formed `<svg>`. That is worse
  than rejecting it: `bypassSecurityTrustHtml` returns an object, which is truthy, so the
  component's `@if` took the icon branch and painted a blank box instead of falling back to the
  `extension` icon. `registerSvgIconFromContent` likewise registered the empty literal, which
  made `hasPluginIcon()` true and beat the caller's own fallback. It now returns `null` when
  nothing drawable survives.
- **A false positive in the `url()` guard.** The obvious `url\(\s*['"]?\s*(?!#)` backtracks, so
  the legitimate `url( #g )` and `url('#g')` were dropped along with the remote ones. Replaced
  with an explicit extraction that checks each reference.
- **Two of my own assertions were weaker than they looked.** `not.toContain('javascript:')` was
  vacuous, because `to` and `values` are not allowlisted attributes so the string disappears
  either way; and the test named for the namespace check was actually carried by the tag name.
  Both now assert on the rendered tree.

## Known compatibility costs, stated plainly

Worth knowing before you or anyone else files a follow-up:

- `<style>` and the `style` attribute are dropped. Illustrator's default "style elements" export
  and Inkscape's default save format both put all paint there, so such an icon renders with
  SVG's default fill. Plugin icons should use presentation attributes.
- `<title>` is dropped, which costs the accessible name. Re-adding it is actually safe (the
  CDATA defence is the rebuild, not the tag exclusion), it just puts an HTML integration point
  back into the output for something that draws nothing, so I left it out and preferred the host
  component's `aria-label`.
- `<filter>` and SMIL animation are dropped, so Figma drop shadows and spinner icons degrade to
  static shapes.
- `<use>`/`<symbol>` sprite-style icons are dropped, per the denial of service above, so a
  single-file icon built that way renders empty and falls back to the generic plugin icon.

All 19 icons currently in `packages/plugin-dev/` are unaffected, with zero tags and zero
attributes dropped.

## Dropped from this PR

The esbuild override, for the reasons in my earlier comment. I re-checked all three against current master and they still hold: [GHSA-g7r4-m6w7-qqqr](https://github.com/advisories/GHSA-g7r4-m6w7-qqqr) is still low severity and dev-only, and `vite@7.3.5` and `tsup@8.5.1` both still declare `esbuild: ^0.27.0`, which excludes `0.28.1`. Worth adding: Angular's own toolchain already resolves esbuild 0.28.1 on master today, so the override would mostly be forcing `vite`/`tsup` onto an undeclared minor for no gain. If you still want it, it is welcome as its own PR.

## Verification

Every payload is asserted against the DOM the sink actually builds, never the returned string, since that was exactly what made the original specs unable to observe the bug.

- Run against **your** sanitizer, 22 of the 74 specs across the three touched spec files fail, surfacing a live `` in the rendered DOM. Against the replacement all 74 are green.
- Reverting the sanitize call in `PluginIconComponent` fails 3 of its 4 specs and renders a live `` inside the real component; reverting it in `GlobalThemeService` fails 3 of its specs. So the wiring is covered, not just the utility.
- Neutering the output encoding to the identity function produces a live `onload` attribute and a live ``, which is what convinced me to add the "entity-encoded markup stays text" specs. Escaping is load bearing here, not hygiene: the parser hands you `&lt;img ...&gt;` already decoded into a text node.
- Every assertion that only checks for absence carries a positive control, so a sanitizer that returned an empty `<svg></svg>` for everything cannot pass. Verified: that mutation fails 24 of the sanitizer specs.
- All 19 real plugin icons under `packages/plugin-dev/` round-trip with zero drawable elements and zero attributes dropped. I checked that oracle is live by removing `viewBox` from the allowlist, which fails it.
- 1475 specs green across `src/app/plugins`, `src/app/core/theme` and `src/app/util`; production AOT build clean.

## On severity, for the record

Unchanged from my July note, and worth restating since it is the part your PR description undersold. For an *enabled* plugin this is not privilege escalation, because plugin code already runs through `new Function` in the main renderer and the CSP is `script-src *` with `unsafe-inline`/`unsafe-eval`. The exposure that matters is that `_discoverUploadedPlugins()` registers a cached icon unconditionally, without an `isEnabled` gate, so a **disabled, never-loaded** uploaded plugin could execute JS in the main renderer at startup with nothing rendered. That is still true on master today, and it is what this fixes.

Thanks again for finding it, and sorry it took a while to close out.

johannesjo:
Superseded by #9440 (see above).

Issue #7521, opened 2026-05-08
Sync: validation result dropped in OperationLogHydrator + boolean-flag pattern in sync result unions (#7330 follow-ups)


johannesjo:
Surfaced during multi-agent review of the issue #7330 fixes (PR forthcoming on branch \`feat/issue-7330-bc1a3f\`). These are out-of-scope for that branch but worth tracking.

## 1. Hydrator validation result dropped on startup replay

`src/app/op-log/persistence/operation-log-hydrator.service.ts:564` — `_validateAndRepairCurrentState` returns `Promise<void>` and silently discards the validation outcome. Same bug class as #7330's primary defect (sync-side validation surfacing was just plumbed through 5 layers in commit `e0a08eac6`), but on the **startup hydration** path — different entry point, same anti-pattern.

Complication: per the file's own comments at lines 532–552, the repair system invoked here (`validateAndRepair` + `RepairOperationService`) is currently **disabled**: \"DISABLED: Repair system is non-functional - this code path is unreachable because validateAndRepair() always returns wasRepaired: false\". Surfacing the boolean in isolation, without re-enabling repair, would be performative.

**Options:**
- (a) Re-enable the repair-op creation path AND surface the boolean — restores the documented design intent.
- (b) Replace `_validateAndRepairState` with a call to the active `ValidateStateService.validateAndRepairCurrentState` (same one sync uses), so hydration uses the same Checkpoint-D path.
- (c) Document the disabled state more loudly and leave it.

## 2. Boolean-flag pattern in `DownloadOutcome` / `UploadOutcome`

`src/app/op-log/core/types/sync-results.types.ts:189-228` — the #7330 fixes added `validationFailed?: boolean` to the success-shaped variants of these discriminated unions. Per the file's own comment (lines 161-166), these unions were introduced specifically to **replace** boolean-flag combinations with distinct `kind` discriminators. The new optional flag reverts to the pattern they replaced.

`sync-wrapper.service.ts:499-502` does `kind === 'ops_processed' && downloadResult.validationFailed` — exactly the boolean-flag combination idiom the unions tried to eliminate.

**Suggested fix:** introduce `kind: 'ops_processed_invalid_state'` (and `kind: 'completed_invalid_state'`) variants carrying the same fields plus a marker. Then `sync-wrapper` switches on `kind` once; the type system enforces the ERROR branch instead of relying on call sites remembering to read a flag.

This is tech debt, not a bug — but the next sync-area change in this file will trip over it.

[...]

PR #9079, opened 2026-07-16
Add critical tests for task and sync helper flows

ghostkeni:
## Problem

Several high-risk code paths still lacked direct unit coverage, especially around task mutation helpers, OAuth callback/state handling, and same-batch sync archive/delete filtering. Those paths carry business logic and data-integrity risk, so regressions there are expensive and easy to miss.

## Solution

Add focused specs for the most critical uncovered helper flows instead of broad snapshot-style coverage.

- Cover task reducer utility behavior for done-state timestamps, repeat-task completion, parent time rollups, estimate recalculation, orphan subtask cleanup, and parent side effects when removing subtasks.
- Cover OAuth state validation and the web redirect bridge so same-origin callback delivery, error forwarding, and listener cleanup are exercised directly.
- Cover bulk archive/delete filtering for stale payloads, malformed payloads, direct TAG/PROJECT LWW payload filtering, and same-batch child removal projection.

Non-obvious note: `npm run checkFile` is currently broken in this Windows environment because `tools/check-file.js` cannot spawn `npm`. I validated the touched specs with the equivalent targeted `prettier:file`, `lint:file`, and `test:file` commands instead.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [x] Other (please describe)

Test coverage for critical existing logic.

## Checklist

### DIFF
--- a/src/app/features/tasks/store/task.reducer.util.spec.ts
+++ b/src/app/features/tasks/store/task.reducer.util.spec.ts
@@ -10,6 +10,7 @@ import {
   updateTimeEstimateForTask,
   updateTimeSpentForTask,
 } from './task.reducer.util';
+import { _resetDevErrorState } from '../../../util/dev-error';
 
 describe('task.reducer.util', () => {
   const DAY_1 = '2026-07-10';
@@ -113,16 +114,18 @@ describe('task.reducer.util', () => {
         'child',
         {
           [DAY_1]: 90,
-          [DAY_3]: 60,
+          [DAY_3]: 90,
         },
         state,
       );
 
-      expect(result.entities['child']!.timeSpent).toBe(150);
-      expect(result.entities['parent']!.timeSpent).toBe(150);
+      // net delta: -30 (DAY_1 down) -30 (DAY_2 zeroed) +90 (DAY_3 added) = +30
+      // parent timeSpent must move off its starting 150 so the accumulator matters
+      expect(result.entities['child']!.timeSpent).toBe(180);
+      expect(result.entities['parent']!.timeSpent).toBe(180);
       expect(result.entities['parent']!.timeSpentOnDay).toEqual({
         [DAY_1]: 90,
-        [DAY_3]: 60,
+        [DAY_3]: 90,
       });
     });
 
@@ -212,8 +215,13 @@ describe('task.reducer.util', () => {
 
   describe('deleteTaskHelper', () => {
     it('should remove state-discovered orphan subtasks when deleting a parent task', () => {
+      // `isShowAlert` latches off after the first devError of the whole Karma run,
+      // and test.ts creates the window.alert spy once and never resets its call
+      // history — so we must reset both before asserting devError fired.
+      _resetDevErrorState();
       if (jasmine.isSpy(window.alert)) {
         (window.alert as jasmine.Spy).and.stub();
+        (window.alert as jasmine.Spy).calls.reset();
       } else {
         spyOn(window, 'alert').and.stub();
       }
--- a/src/app/features/tasks/store/task.reducer.util.spec.ts
+++ b/src/app/features/tasks/store/task.reducer.util.spec.ts
@@ -1,0 +1,295 @@
+import { Update } from '@ngrx/entity';
+import { Task, TaskState } from '../task.model';
+import { taskAdapter } from './task.adapter';
+import { initialTaskState } from './task.reducer';
+import {
+  deleteTaskHelper,
+  removeTaskFromParentSideEffects,
+  updateDoneOnForTask,
+  updateStartDateForRepeatableTask,
+  updateTimeEstimateForTask,
+  updateTimeSpentForTask,
+} from './task.reducer.util';
+
+describe('task.reducer.util', () => {
+  const DAY_1 = '2026-07-10';
+  const DAY_2 = '2026-07-11';
+  const DAY_3 = '2026-07-12';
+
+  const createTask = (id: string, partial: Partial<Task> = {}): Task =>
+    ({
+      id,
+      title: id,
+      projectId: 'INBOX',
+      created: 1,
+      isDone: false,
+      subTaskIds: [],
+      tagIds: [],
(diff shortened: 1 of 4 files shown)


johannesjo (commented):
Thanks for this — and especially for picking real risk areas rather than padding coverage numbers.

Since a test-only PR lives or dies on whether the tests actually *hold the code in place*, I reviewed it by mutation testing: sabotage each production path, then check whether your specs fail. **12 of 15 mutations were caught.** That's a genuinely strong result, and the OAuth/security coverage in particular is exactly what I'd want pinned down.

Two things need a fix before merge, both verified empirically rather than eyeballed.

---

### 1. `expect(window.alert).toHaveBeenCalled()` is vacuous in a full-suite run

In `task.reducer.util.spec.ts`, the orphan-subtask test asserts `devError` fired. I deleted the entire `devError(...)` call from `deleteTaskHelper` and re-ran:

| Configuration | Result |
| --- | --- |
| your spec **alone** + sabotaged code | `1 FAILED` — catches it |
| your spec **+ `task.reducer.spec.ts`** + sabotaged code | **`75/75 SUCCESS` — misses it** |

The second row is how CI runs. Two mechanisms combine:

1. `src/test.ts` (~line 113) creates the `window.alert` spy **once per Karma run** and never resets its call history — so an alert from *any* earlier spec already satisfies `toHaveBeenCalled()`.
2. `devError`'s module-level `isShowAlert` (`src/app/util/dev-error.ts`) **latches to `false`** after the first alert of the run, so by the time your spec runs, `devError` shows no alert at all — yet the assertion still passes.

You picked up the `jasmine.isSpy(window.alert)` pattern from the sibling test in `task.reducer.spec.ts`, but that one also calls `_resetDevErrorState()` first — which is precisely what re-arms mechanism 2. Both lines are needed:

```ts
import { _resetDevErrorState } from '../../../util/dev-error';

// ...
it('should remove state-discovered orphan subtasks when deleting a parent task', () => {
  // `isShowAlert` latches off after the first devError of the whole Karma run,
  // and test.ts's window.alert spy is created once and never reset.
  _resetDevErrorState();
  if (jasmine.isSpy(window.alert)) {
    (window.alert as jasmine.Spy).and.stub();
    (window.alert as jasmine.Spy).calls.reset();
  } else {
    spyOn(window, 'alert').and.stub();
  }
```

Verified both directions: still green against real code (`75 SUCCESS`), and the same sabotage is now **caught** (`1 FAILED, 74 SUCCESS`).

The rest of that test is fine — the orphan-removal assertions caught their mutation on their own.

---

### 2. `updateTimeSpentForTask` can't observe the thing it's named after

The test is called *"should update parent time incrementally"*, but the numbers make `totalDelta` exactly zero: `-30` (DAY_1) `-30` (DAY_2 zeroed) `+60` (DAY_3). So mutating the accumulator away —

```diff
-        timeSpent: parent.timeSpent + totalDelta,
+        timeSpent: parent.timeSpent,
```

— **survives**, because the expected `150` is also the parent's starting value. Nudging DAY_3 so the total actually moves closes it:

```diff
       const result = updateTimeSpentForTask(
         'child',
         {
           [DAY_1]: 90,
-          [DAY_3]: 60,
+          [DAY_3]: 90,
         },
         state,
       );

-      expect(result.entities['child']!.timeSpent).toBe(150);
-      expect(result.entities['parent']!.timeSpent).toBe(150);
+      // net delta is -30 (DAY_1) -30 (DAY_2 zeroed) +90 (DAY_3) = +30, so the
+      // parent total must move off its starting 150 for the accumulator to matter
+      expect(result.entities['child']!.timeSpent).toBe(180);
+      expect(result.entities['parent']!.timeSpent).toBe(180);
       expect(result.entities['parent']!.timeSpentOnDay).toEqual({
         [DAY_1]: 90,
-        [DAY_3]: 60,
+        [DAY_3]: 90,
       });
```

With that, the mutation is **caught**. The zeroed-day deletion branch was already covered correctly.

---

### 3. Optional nit

`stripBatchArchivedTaskIdsFromLwwPayload`'s `if (!isLww) return op;` short-circuit is never exercised — every call in the spec passes `true`. Removing the guard survives. One extra case if you feel like it; not a blocker.

---

### What's solid

Worth stating explicitly, since it's most of the PR — every security-relevant mutation was caught:

- **`oauth-state.util`**: provider mismatch check, one-time-use consumption, and expiry pruning — all three caught. Good CSRF coverage.
- **`plugin-oauth-redirect.handler`**: origin check, message-type check, and destroy cleanup — all three caught.
- **`bulk-archive-filter.util`**: same-batch parent projection, archive-vs-delete set distinction, and the `#7521` malformed `actionPayload: null` guard — all caught. Real protection on the `#7330` "archived tasks reappear in today's view" path.

### On `checkFile`

All four files pass `npm run checkFile` cleanly here on Linux, so nothing to do on your side. Your `tools/check-file.js` npm-spawn failure looks Windows-specific — thanks for flagging it, I'll track that separately since it'll hit every Windows contributor.

Once 1 and 2 are in, this is good to merge.

ghostkeni:
Thanks for the thorough mutation-testing review, @johannesjo!

Both issues are fixed in the follow-up commit (2660f08):

1. **Vacuous alert assertion** -- Added \_resetDevErrorState()\ before the orphan-subtask test and called \(window.alert as jasmine.Spy).calls.reset()\ so the assertion is fresh regardless of prior test runs in the same Karma suite.
2. **Zero-delta time accumulator** -- Changed \DAY_3\ from 60 to 90, giving a net delta of +30 (−30 from DAY_1 decrease, −30 from DAY_2 removal, +90 from DAY_3 addition). The expected \	imeSpent\ is now 180 instead of 150, so the accumulator mutation is detectable.

As a bonus, I also added an \isLww=false\ short-circuit test for \stripBatchArchivedTaskIdsFromLwwPayload\ to cover the nit you spotted.

johannesjo:
Thank you very much! <3

Issue #8469, opened 2026-06-18
op-log: residual snapshot re-replay window double-applies non-idempotent ops (follow-up to #8308)


johannesjo:
## Context

Follow-up to #8308 / PR #8439, which wrapped `saveCurrentStateAsSnapshot()` in the `OPERATION_LOG` lock and inverted the read order (`lastSeq` before state). That fix eliminated the **silent op-loss** failure mode. As a deliberate, documented trade-off, it converts the worst-case interleaving from *op-loss* into *op-re-replay*.

The re-replay is possible because NgRx state mutates **synchronously at dispatch** (in the reducer) while the op's `seq` is assigned **later**, under the same lock (`operation-log.effects.ts`). So the locked snapshot body can read `lastSeq = N` and then a state that already reflects an op whose seq-write is still queued behind the lock. The snapshot is then tagged `lastAppliedOpSeq = N` with that op's effect already present → the next boot's tail replay (`getOpsAfterSeq(N)`) re-applies it.

Re-replay is harmless for idempotent reducers (entity-adapter CRUD/move/reorder, absolute-value setters). It is **not** harmless for reducers that accumulate onto current state.

## Non-idempotent persistent reducers (re-replay double-applies)

Audited all `isPersistent: true` actions and their reducers. The following double-apply on re-replay:

| Action | File:line | Pattern | Replay branch |
|---|---|---|---|
| `syncTimeSpent` | `task.reducer.ts:261` | `currentTimeSpentForDay + duration` | remote-only (`isRemote` guard) |
| `removeTimeSpent` | `task.reducer.ts:504` | `Math.max(current - duration, 0)` | **unconditional** (no guard) |
| `syncSimpleCounterTime` | `simple-counter.reducer.ts:277` | `currentVal + duration` | remote-only (`isRemote` guard) |
| `logFocusSession` | `metric.reducer.ts:62` | `[...focusSessions, duration]` | **unconditional** (no guard, no dedup) |
| `moveProjectTaskToRegularList` | `project.reducer.ts:339` | `[...todaysTaskIdsBefore, taskId]` (DONE-section, **null-anchor** branch — appends **without** first filtering `taskId` out) | **unconditional** (no `includes` guard on this branch) |

Effect of a double-apply:

- **Time/metric reducers** (`syncTimeSpent`, `removeTimeSpent`, `syncSimpleCounterTime`): inflated/deflated tracked time for a day. Silent — surfaces only as a wrong number.
[...]

PR #9081, opened 2026-07-16
fix(sync): guard snapshots against unpersisted state changes (#8751)

### LINKED ISSUE #8751: op-log: unrecovered persist failure gets baked into the snapshot as permanent cross-device divergence
**From a deep architecture review of the operation log (2026-07-04).**

## Problem

When an op write fails or is rejected by validation (`src/app/op-log/capture/operation-log.effects.ts:274-296, 355-385`), the optimistic NgRx state change is kept; the only remedy is a sticky reload snackbar (`notifyUserAndTriggerRollback` rolls nothing back), and the meta-reducer deliberately doesn't block on capture failure (documented "known limitation", `operation-capture.meta-reducer.ts:275-277`).

The unexamined interaction: snapshots are taken from the **live NgRx store** (`state-snapshot.service.ts`, `operation-log-compaction.service.ts:66`). If the user dismisses/ignores the snackbar and keeps working, after ~500 further successful ops compaction bakes the phantom change into `state_cache` — now the change is durable locally but has **no op behind it**:

- other devices never receive it (permanent, silent cross-device divergence), and
- replay-from-log no longer reproduces local state, quietly breaking the "log is the source of truth" invariant (operation-rules.md §1.3).

A transient IndexedDB failure is thereby converted into permanent, unrepairable divergence.

## Suggested direction

- [ ] After an unrecovered persist failure, set a "log-divergent" flag that suppresses snapshotting/compaction until resolved, **or**
- [ ] force a full-state op (SYNC_IMPORT/REPAIR-style) at the next opportunity so the store state re-enters the log, then clear the flag.
- [ ] A spec covering: persist failure → user continues → compaction runs → assert no divergent snapshot (or forced full-state op).

johannesjo:
Closes #8751 (sync-audit C-01).

## Problem

An unrecovered op-log persist failure (failed write, lock timeout, or validation rejection) keeps the optimistic NgRx change with no durable op behind it — a **phantom change**. Snapshots are taken from the live store, so once compaction or the hydrator snapshot save wrote that state into `state_cache`, the phantom became durable locally with no op to sync: other devices never receive it, replay-from-log no longer reproduces local state, and even a reload no longer heals it. A transient IndexedDB failure was thereby converted into permanent, silent cross-device divergence.

## Fix

**Invariant: live state must never be snapshotted into `state_cache` while it may contain changes that no durable op represents.**

- `OperationCaptureService` gains a sticky, per-tab divergence flag, set on every keep-state failure path in `OperationLogEffects`: the write-failure catch, the invalid-entityId early return, and the validation-rejection branch (whose error snackbar is now sticky, `duration: 0` — it previously auto-dismissed).
- `capture/phantom-change-guard.util.ts` reports why snapshotting is unsafe: the sticky flag, **or** captured writes still pending, **or** deferred sync-window actions still buffered.
- Enforced at **both** live-state `state_cache` writers — compaction `_doCompact` (covering `compact()` and `emergencyCompact()`) and `saveCurrentStateAsSnapshot` — checked synchronously immediately before the snapshot read, with no intervening `await`. A sticky-flag fast-path in `_doCompact` avoids cross-tab lock churn once the flag is set.
- Recovery is unchanged: the sticky snackbar's reload rebuilds state purely from durable data, which discards the phantom and re-enables snapshotting. Skipping is always safe — the op-log stays the source of truth.

### Relationship to #9083 (#8469 quiesce)

#9083 landed first and quiesces the capture pipeline via `flushThenRunExclusive`. That structure is kept intact here (both services still flush before taking the lock, and the snapshot service keeps its state-read-before-`lastSeq` order). The guard is layered on top because **quiescing structurally cannot replace it**: a write that fails *for good* drains the pending counter to zero and stays diverged forever, so only the sticky flag catches it.

Where the two overlap, this PR's guard is the stricter superset. Notably `emergencyCompact` no longer proceeds while writes are pending: it is invoked from the failing write's own stack, so it would otherwise snapshot live state containing exactly the change that is failing. Freeing space at that moment is impossible without baking that phantom in — #9082 (delete-only emergency compaction: prune ops, write no snapshot) is what would make quota recovery possible again.

## Deliberate non-goals

- **File-based full-state uploads** can still *propagate* a phantom remotely — a different failure shape than the local bake-in fixed here.
- The flag is **not cleared by SYNC_IMPORT/BACKUP_IMPORT** even though they discard the phantom too — conservatively simple; the only cost is compaction staying suppressed until the reload the snackbar already demands.
- A transient lock timeout marks divergence and so suppresses compaction until reload. Semantically correct (the op really is missing), but on long-lived sessions that never reload the log can grow.

## Tests

Effects: marks on unrecovered write failure; does not mark on success or on a recovered quota failure (fixture uses Firefox's spelling — the standard name is wrapped into a plain `Error` and can never reach that path); validation rejection marks + sticky snack. Compaction and snapshot: skip for each guard signal, including `emergencyCompact`; resume once pending writes settle / the deferred buffer drains; the fast-path never acquires the lock.

**Mutation-verified**: with both flag paths disabled (the util's check and `_doCompact`'s fast-path), exactly the 4 guard tests fail and nothing else. This mattered — three of them were previously vacuous: an unstubbed vector clock made the save throw into the outer catch and skip the write regardless, so they passed with the guard deleted.

Full unit suite green in both timezone variants.


johannesjo:

PR #9088, opened 2026-07-16
fix(sync): fence in-flight sync cycles across destructive config changes

### LINKED ISSUE #9074: sync: destructive config changes (encryption ops, provider switch) don't exclude/cancel in-flight sync cycles
**From a sync durability/convergence audit (2026-07, revision `6fefd741c5`).**

## Problem

Destructive config changes — an **encryption operation** (enable/disable/password change) or a **provider/account switch/disconnect** — do not exclude or cancel a sync cycle that is already in flight. Blocking is point-in-time and one-directional: it prevents *new* cycles from starting but neither waits for an already-running cycle to drain nor cancels it. An in-flight cycle that is mid-`await` when the config flips can therefore complete afterwards and mutate local state, upload, acknowledge, or advance the cursor **against the new epoch/target**.

## Two faces of the same root cause

**A. Encryption ops don't exclude active side-channel syncs.**
`runWithSyncBlocked()` (used by e.g. `src/app/imex/sync/encryption-password-change.service.ts:69`) blocks *new* work, but the immediate-upload and WS-download paths gate only on a point-in-time boolean (`immediate-upload.service.ts:142`, `ws-triggered-download.service.ts:129` → `isSyncInProgress`). A cycle already past that check keeps running. Result: mixed-key server history, resurrected data around a clean-slate replacement, or state applied from the wrong encryption epoch.

**B. Provider switch/disconnect doesn't cancel an in-flight WS cycle.**
`WsTriggeredDownloadService.stop()` (`src/app/op-log/sync/ws-triggered-download.service.ts:91-103`) unsubscribes and clears timers but holds **no cancellation token or lifecycle generation** for active work. On a provider switch, the config update makes provider B active and calls `stop()`, which only detaches future notifications — the old provider-A download promise continues and can apply/upload/ack after the switch. When B is file-based, the next B snapshot can contain state applied from account A (cross-provider contamination, not just a stale status indicator).

## Impact

Cross-epoch / cross-provider data contamination: old-key or old-account operations committed into the new target, broken account convergence, or undecryptable history. Reachable through ordinary settings actions (password change, provider switch, disable sync) — no attacker required — because the race window spans every network/dialog/apply/follow-up-upload `await` in the old cycle.

## Scope note (what is already covered)

This is **not** the target-identity/cache-keying problem (#8897) or the file-provider target-change invariant (#9063 / #9066) — those fix *which dataset state is used*. This issue is that **in-flight async work isn't stopped or fenced** when the epoch/target changes, so a stale completion still writes.

## Proposed fix (KISS)

Minimal, not a new framework:

1. Add a monotonic **sync epoch** counter, bumped on every provider/account/encryption-config change. Capture it at cycle start; re-check it before each state mutation, upload, ack, and cursor write, and make stale completions no-ops.
2. Make `runWithSyncBlocked()` (and provider switch) **await drain** of any running cycle before proceeding, or return an explicit failure — not opportunistically skip. Cancel transport work via `AbortSignal` where supported.

(A full exclusive lifecycle coordinator is possible but YAGNI for now — the epoch fence + drain closes the data-safety hole.)

## Test gap

No test switches the provider (or runs an encryption op) *during* an active download/upload and asserts no old-epoch op/cursor/upload/status lands. Add: deferred old-provider download → switch active provider → resolve the old promise → assert it is a no-op. Also (a) in-flight immediate upload then password change, (b) in-flight WS apply then enable encryption.

## Related

- #8897 (target identity / switch detection), #9063 / #9066 (file-provider target-change invariant) — complementary; they don't fence in-flight work.
- Companion audit issue: LocalFile folder picker commits the live target before Save (filed separately).

_Severity: High (cross-epoch / cross-provider data contamination). Consolidates audit findings C1-1 (ORCH-1) and C1-2 (SWITCH-2)._

johannesjo:
Fixes #9074 (audit findings C1-1 / C1-2).

## Problem

Destructive config changes — an encryption operation (enable/disable/password change) or a provider/account switch — blocked **new** sync cycles but neither drained nor cancelled a cycle already mid-`await`. The stale cycle could then apply remote ops, upload, acknowledge (`markSynced`), or advance the server-seq cursor **against the new epoch/target**: cross-epoch / cross-provider contamination, resurrected data around a clean-slate replacement, mixed-key server history.

- `runWithSyncBlocked()` drained only `_isSyncInProgress$` (main sync); the immediate-upload and WS-download side channels hold `SyncCycleGuardService`, which nothing drained.
- `WsTriggeredDownloadService.stop()` only detached future notifications; the in-flight download promise kept its captured old-provider object and ran to completion.
- No write point in the download/upload/apply flows re-checked the active provider or any epoch.

## Fix (epoch fence + bounded drain — the issue's KISS sketch)

- **Sync epoch** on `SyncProviderManager`: monotonic counter, bumped **after** a provider switch completes (incl. disable), after a target-moving config write (`isSyncTargetChanged` — content-only saves do NOT bump), at the bypass ingresses, and at `runWithSyncBlocked` entry. Generalizes the existing `_activeProviderSetupId` / `FileSyncTargetChangedError` capture-and-recheck precedents.
- **Every cycle entry point** (main sync, immediate upload, WS download, force upload) captures the epoch synchronously with its cycle-guard claim and threads it as `fenceEpoch`.
- **Provider I/O fenced in one choke point:** `getOperationSyncCapable(provider, { fenceEpoch })` returns a per-cycle delegate that re-asserts the epoch before every provider call — uploads, downloads, and all 8 `setLastServerSeq` cursor writes. `ImmediateUploadService` now routes through the wrapper instead of a raw cast.
- **Local writes** re-assert at the call site: all three apply-lock closures (asserted after the lock wait — kills the "stale apply queued behind `createCleanSlate`" interleave), deferred acks, snapshot hydration, migration appends, rejected-ops handling, raw-rebuild resume.
- **Stale completions are benign aborts:** `SyncEpochChangedError` → log + `UNKNOWN_OR_CHANGED`, no snack, no retry (mirrors `FileSyncTargetChangedError` handling). Each abort point lands in a crash-equivalent state (deferred acks re-upload; a behind cursor re-downloads with dedup).
- **`runWithSyncBlocked`:** serialized via promise chain (concurrent destructive ops no longer fence each other or clear the shared flag prematurely); sets the block flag FIRST, bumps the epoch, then drains the main sync AND the guard-holding side channels via new `SyncCycleGuardService.waitUntilFree()` (bounded by `SYNC_WAIT_TIMEOUT_MS`, throws on timeout). The drain is load-bearing on top of the fence: a fence cannot recall request bytes already on the wire.

Provider switches get no drain (they're a reactive store subscription with nowhere to block) — the epoch fence alone neutralizes stale cycles there.

## Deliberate limitations (documented in code + docs)

- `forceUploadLocalState` / USE_LOCAL/USE_REMOTE coordinator **local** writes are unthreaded (covered by flag+guard exclusion; provider I/O is delegate-fenced). Shortcut comments name the upgrade path.
- No `AbortSignal` transport cancellation (per the issue: full coordinator is YAGNI); the drain shrinks the on-the-wire window to near zero for encryption ops.
- Auth-token writes bump the epoch (`accessToken` is target identity) → a re-auth during a background cycle causes a benign abort-and-retry.
- Key-recovery config writes (enter-password / decrypt-error dialogs) are content-only and intentionally do NOT bump.

## Tests

Closes the issue's test gap — both side channels previously tested only gate-time skips, nothing mid-flight:

- Deferred download → epoch change → resolve → **no retry, no ERROR, guard released** (WS spec, the issue's exact repro).
- Mid-flight bump via the **real** `SyncProviderManager` → deferred ack and server-migration writes abandoned; unthreaded flows keep existing behavior (sync-service spec).
- `runWithSyncBlocked`: flag-before-drain ordering, guard drain, timeout → throw + flag cleared, serialization, bump-per-invocation.
- Bump gating: target move / switch / disable bump; content-only saves and same-id sets don't; bump lands **after** the async provider swap.
- Delegate semantics incl. pass-through when unfenced.

Full unit suite green (12,979 passed). New docs section in `contributor-sync-model.md` ("an unthreaded flow is an unfenced flow").


johannesjo:

PR #9089, opened 2026-07-16
fix(sync): dedupe surgical-sync retries and keep the import author through pruning

johannesjo:
## Problem

Two independent ways a lost or interrupted sync can lose/duplicate data:

1. **File-based providers (WebDAV / LocalFile / Dropbox / OneDrive).** A `PUT` of `sync-ops.json` can commit while its response is lost (network drop, app kill, timeout). The op stays `pending` locally and is retried after restart — and the retry appended a **second copy** of the same op and bumped `syncVersion` again.

2. **SuperSync server clock pruning.** `MAX_VECTOR_CLOCK_SIZE = 20`, and pruning kept the highest counters while protecting only the uploader. After a `SYNC_IMPORT`/`BACKUP_IMPORT`, the full-state author is a required causal edge for every post-import op but its counter is *low*, so it was dropped first. Clients at the import boundary then read those ops as CONCURRENT and filter them out.

Also: the settings form never persisted `isUseSplitSyncFiles`, so the Surgical-sync toggle silently reset.

## Solution

- **Retry dedup:** index the downloaded `recentOps` by id and append only genuinely new ops; acknowledge ids already present without advancing `syncVersion`.
- **Pruning:** protect the active causal full-state author alongside the uploader. `isSameDuplicateOperation` reconstructs the *stored* row's normalization from the stored clock's own key set, so a genuine retry of an oversized post-import op no longer reads as an id collision.
- **E2E hardening:** the SuperSync specs now drive real upload endpoints and assert exact operation ids across response loss, validation failures, schema blockers, full-state boundaries, pruning, and concurrent edits. Several old assertions were tautological (one swallowed its own failure via `.catch(() => {})`); a new `webdav-surgical-sync.spec.ts` reads the real remote ops file and asserts its contents.

Commits 4–5 are fixes from a multi-agent review of the first three:

- **`.bak` heal (`827759cad4`).** The `.bak` recovery path caches the *corrupt primary's rev* precisely so the same cycle's conditional overwrite repairs `sync-ops.json`. The new duplicate-retry short-circuit read that cache and returned before the write — leaving the primary corrupt whenever the recovered buffer already held every pending op. The short-circuit now falls through while a heal is pending. Also stopped synthesising `serverSeq` for already-committed ops: the field is optional, upload and download number ops differently, and a mixed batch could hand two ops the same value.
- **Author lookup (`0522627143`).** `batchUpload` is **off by default**, so the guarded batch path was not the one serving production — the serial path queried the author *per op* inside one transaction, and a 21–50-entry clock passes validation and trips the guard on every one. Now memoized per transaction and resolved lazily, so only an op whose clock actually overflows pays, and only once. This retires the batch pre-scan and its loop-carried author, leaving both paths on one mechanism.

Commits 6–8 came out of a second review round, after the SuperSync E2E `3.1` shard stayed red:

- **`renameTask` focus (`9f3936f5f1`).** `3.1` was the last red test and it was **the test, not the engine**. `renameTask` trusted `el.focus()` to emit a focus event, but the spec drives two clients as separate pages and only one can hold focus — so on CI the event never fired. `TaskTitleComponent` then kept `_isFocused=false`, and `resetToLastExternalValueTrigger` reset `tmpValue` to the stored title on the next task-object emission; blur therefore computed `wasChanged=false`, `task.component` skipped `update()`, and the rename never became an op. Client A's rename lived only in the UI, A uploaded nothing, B uploaded its done op, and A's title reverted on download — exactly the state the CI artifact captured. Verified against the real `TaskTitleComponent`: `_isFocused=false` → `wasChanged=false`; `_isFocused=true` → `wasChanged=true`. A real user always has real focus, so the app is unaffected. This also retracts the previous commit's premise: the `toBeVisible` wait matched `tmpValue` (a component-local signal rendered in *both* template branches), so it never observed the committed title.
- **Docs revert (`2017b8e855`).** The master merge ran prettier across three unrelated documents (table padding, `*em*`→`_em_`, no content change). Restored to master.
- **Roundtrip accounting (`8cfbdf32e2`).** `resolveFullStateAuthor` memoizes per transaction, so the lookup it counted fired at most once per upload — the plumbing reported a number that is always 0 or 1. Dropping `didQuery` also lets the two new helpers return `string | undefined` and `string[]` rather than tuples. `uploadDbRoundtrips` and the batch counter are untouched.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Sync-risk notes

- **The `MAX_VECTOR_CLOCK_SIZE` bound is structurally safe.** `limitVectorClockSize` caps the *preserve* set at MAX as well as the fill loop, so protecting an extra id can never grow a stored clock past 20.
- **Collision detection is marginally looser, deliberately** (documented in `conflict.ts`): for an oversized incoming clock the comparison projects onto the stored key set, so an id collision differing *only* by an extra clock entry now reads as a duplicate. Every other structural field (clientId, payload, entity, timestamp-within-drift) must still match — and when they all do, acking beats rejecting a retry whose only sin is having learned about one more client.
- **The retry dedup is best-effort, not total.** It recognises an op still in `recentOps`. Compaction trims with `slice(-N)` (keeping the newest), so a just-committed op survives and the real retry case is covered; an op pushed out by a later compaction would still re-append, and is then filtered by `appliedOpIds` on download. This is strictly better than before, where *every* retry re-appended.
- **This change is roll-forward-only.** The projection is backward-compatible in the direction that matters: a retry compared against a row written by an *older* server still reads as a duplicate, because it reprojects onto whatever key set that server chose. The reverse is not symmetric — after a **server rollback**, a retry of a row written by the *new* server reads as an id collision and returns `INVALID_OP_ID`, which the client routes to `permanentlyRejectedOps` and drops for good. Do not roll the server back across this change.
- **The dedup is scoped to Surgical sync.** `_uploadOpsSplit` is the only path that indexes `recentOps` by id; the default single-file path (`_buildMergedSyncData`) still appends unconditionally, so the identical lost-response retry re-appends there. Smaller impact (the full-state snapshot is re-embedded on every upload, and `appliedOpIds` filters downstream), but the same bug — pre-existing, not addressed here.

## Checklist

- [x] I have included relevant changes to the documentation/wiki (`docs/sync-and-op-log/vector-clocks.md`)

## Testing

- Full client suite green on both TZ variants: **12906** and **12892** passing, 0 failures.
- Server: **894/895**. The one failure — `should return idempotent success when a committed REPAIR retry has a stale base` — is **pre-existing and red on master too** (verified at `542da1eb6c`, where the suite is 892/893). Not from this branch.
- Both review-fix tests were confirmed non-tautological by reverting each fix and watching them fail.


johannesjo:

johannesjo:

Issue #18, opened 2017-11-20
Cannot reopen the application


Gontran08:
If I hide the application using the window button, I cannot reopen it using the application launcher, I have to use the keyboard shortcut. I have to kill the application to be able to use the launcher again. I also cannot see that the application is still active in the taskbar. Closing the application has the same effect that hiding it.

johannesjo:
Thanks for reporting! May I ask which OS you're on and which desktop environment you're using?

HarmtH:
I don't know if this is the same issue, but if I try to start the application, while it is still running (window is closed, but icon still in the notification area), it won't show the main window. This is with the Ubuntu snap package under Ubuntu 17.10.

```
blackbox ~% superproductivity 
Gtk-Message: Failed to load module "overlay-scrollbar"
Gtk-Message: Failed to load module "gail"
Gtk-Message: Failed to load module "atk-bridge"
Gtk-Message: Failed to load module "canberra-gtk-module"
module.js:487
    throw err;
    ^

Error: Cannot find module '../dialog'
    at Module._resolveFilename (module.js:485:15)
    at Function.Module._resolveFilename (/snap/superproductivity/5/resources/electron.asar/common/reset-search-paths.js:35:12)
    at Function.Module._load (module.js:437:25)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at Object.get [as dialog] (/snap/superproductivity/5/resources/electron.asar/browser/api/exports/electron.js:11:16)
    at process.<anonymous> (/snap/superproductivity/5/resources/electron.asar/browser/init.js:54:31)
    at emitOne (events.js:115:13)
    at process.emit (events.js:210:7)
    at process._fatalException (bootstrap_node.js:338:26)
```
I would expect that the main window of the currently running instance is shown when I try to run ```superproductivity``` again.

johannesjo:
Thank you. I'll look into it tomorrow. Haven't had any issue like this myself, but maybe I overlooked something.

Gontran08:
I'm using Ubuntu 17.10 with Gnome Shell.

Atem18:
Same problem on Debian Sid with Appimage and DEB package with Gnome 3.26.

johannesjo:
[...]

PR #9090, opened 2026-07-16
Point Dashboard plugin to maintained fork (original archived)

ahanel13:
## What

Updates the **Dashboard** entry in `src/assets/community-plugins.json` to point at the actively maintained fork at [ahanel13/sp-dashboard](https://github.com/ahanel13/sp-dashboard).

## Why

The original repository, `dougcooper/sp-dashboard`, has been **archived** by its author and is no longer maintained (last activity around February). I have been maintaining a fork that is currently at v1.5.0 with a number of fixes and features on top of the original, including timezone/date handling fixes, chart improvements, live time tracking, and share/export.

The original author has given their blessing for this move: they archived the upstream repo, added me as a collaborator, and agreed to point users to the fork. See dougcooper/sp-dashboard#18 for the full discussion.

## Changes

The Dashboard entry now reads:

- `url`: `https://github.com/ahanel13/sp-dashboard`
- `author`: `ahanel13`
- `authorUrl`: `https://github.com/ahanel13`

## Note on stars

I left the `stars` count at its existing value of 43 (from the original repo). The fork itself currently shows a lower count since it is a separate repository. Happy to adjust this to whatever value you prefer.

### DIFF
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -66,9 +66,9 @@
   {
     "name": "Dashboard",
     "shortDescription": "A lightweight dashboard plugin that visualizes time tracked, completed tasks, overdue items, and project breakdowns within a user‑defined date range. It ships as a self-contained HTML/JavaScript widget with no external dependencies and is styled to support light/dark themes.",
-    "url": "https://github.com/dougcooper/sp-dashboard",
-    "author": "dougcooper",
-    "authorUrl": "https://github.com/dougcooper",
+    "url": "https://github.com/ahanel13/sp-dashboard",
+    "author": "ahanel13",
+    "authorUrl": "https://github.com/ahanel13",
     "stars": 43
   },
   {
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -65,11 +65,11 @@
   },
   {
     "name": "Dashboard",
-    "shortDescription": "A lightweight dashboard plugin that visualizes time tracked, completed tasks, overdue items, and project breakdowns within a user‑defined date range. It ships as a self-contained HTML/JavaScript widget with no external dependencies and is styled to support light/dark themes.",
+    "shortDescription": "A lightweight dashboard plugin that visualizes time tracked, completed tasks, overdue items, and project/tag breakdowns within a user‑defined date range, with live time tracking and chart share/export. It ships as a self-contained HTML/JavaScript widget with no external dependencies and is styled to support light/dark themes.",
     "url": "https://github.com/ahanel13/sp-dashboard",
     "author": "ahanel13",
     "authorUrl": "https://github.com/ahanel13",
-    "stars": 43
+    "stars": 9
   },
   {
     "name": "Invoicer",


johannesjo:
Thanks for handling this in the open — and for linking dougcooper/sp-dashboard#18, that made it easy to confirm.

Since "point the registry entry at my fork" is worth verifying independently as a matter of course, for the record: the original repo is indeed archived (last push 2026-02-25), dougcooper's comment in #18 explicitly blesses pointing users to your fork and adds you as a collaborator, and your repo is a genuine fork of it with releases through v1.5.0. All checks out.

One small thing before merge: could you set `stars` to `3` (the fork's actual count)? The 43 was already stale — the original sits at 75 now — and the field renders as a literal star badge in the plugin list, so 43 would overstate the fork's count. It doesn't affect ordering (the list sorts alphabetically by name), and it would land at 3 at the next star-refresh pass regardless, so it may as well be accurate from the start.

Otherwise this looks good. Thanks for picking up maintenance of the plugin!

johannesjo:
Thanks for the patience here, and for handling the handover in the open — dougcooper/sp-dashboard#18 made this easy to confirm. I verified independently: upstream is archived, yours is a genuine fork of it, dougcooper's comment there explicitly blesses pointing users at your repo, and `manifest.json.template` is byte-identical to upstream — same permissions, still no `http` or `nodeExecution`. The v1.5.0 zip matches your published source (`plugin.js` identical; `index.html` is just the minified build). No new dependencies. Nice.

One thing before merge: `stars` should be `9`, not `43`. (43 was the original repo's old count — it's at 83 now.) I'd asked for `3` back in July but it's grown since. There's no automated star refresh on our side, so whatever lands is what users see until someone edits it by hand.

Optional: the `shortDescription` still describes the pre-fork feature set — happy to take an updated one mentioning the project/tag drill-down and share/export.

Unrelated to this PR, for your repo: the `SP_DASHBOARD_DOWNLOAD` handler in `plugin.js:16` doesn't check `event.origin`/`event.source` before saving the supplied blob. Needs another frame already inside the app to matter, so not urgent — but a cheap hardening.

Thanks for picking up maintenance of this one.

johannesjo:

Thanks for the PR, and for turning both of the earlier asks around: `stars` is at 9 and the `shortDescription` now covers the project/tag drill-down and share/export.

I re-screened the plugin source at `297089e` (18 text files, ~193 KB read in full: `sp-dashboard/index.html` end to end, `plugin.js`, the manifest template, the Makefile and build scripts, both workflows, the test suite and the repo docs; skipped the 9 LFS screenshots and `package-lock.json`). `plugin.js` runs in the host renderer realm and this app's plugin iframes are deliberately same-origin, so I read the whole of `index.html` on the same terms rather than treating `iFrame: true` as a boundary. What I found:

- **No network at all.** No `http` permission and no `allowedHosts`, so `PluginAPI.request` is fail-closed anyway, but the code also has no `fetch`, `XMLHttpRequest`, `WebSocket`, `EventSource` or `sendBeacon`. The only `http://` strings in `index.html` (lines 2261 and 2263) are the SVG and XHTML XML namespaces used to build the PNG export. The one runtime `img.src` is a locally built `data:` URI; the download anchors are `blob:`.
- **Nothing dynamic in what ships.** The only `new Function` sites in the tree are the four in `tests/index.test.js`, where the JSDOM harness executes the page's own `<script>`. The release zip carries only `plugin.js`, `manifest.json`, `icon.svg` and `index.html`.
- **Storage is one `localStorage` key** (`sp-dashboard-settings`), UI preferences only, plus the nine pre-settings keys folded in once and deleted. No credentials anywhere in the tree, and task data stays in memory.
- **The v1.6.0 zip matches the source.** `plugin.js` and `icon.svg` are byte-identical, `manifest.json` is the template with version and description substituted, the archive has no stray entries or trailing bytes, and the minified `index.html` still carries the guards that matter (`sanitizeIncomingTags`, `escapeHtml`, the `window.parent` sender check).

The `SP_DASHBOARD_DOWNLOAD` sender check that came up last time is in, and done the careful way: `isOwnPluginWindow` at `plugin.js:22-30` reads `data-plugin-id` off `event.source.frameElement`, with the mirror check at `index.html:3245`. Both hold. Nice.

### Nits
- **`plugin.js:78` and `index.html:2315`** – both `postMessage` calls still pass `'*'` as the target origin, so the receiving half of the hardening landed but the sending half did not. `location.origin` is known and constant in both directions, so passing it instead is a one-token change. Nothing exploits this without host-realm execution already, hence a nit rather than a fix request.
- **`package.json:29-31`** – `acorn` sits in `dependencies` rather than `devDependencies`; its only consumer is `scripts/check-js.js`. Nothing ships in the zip, so there is no runtime effect, but the entry text this PR ships says "no external dependencies" and the manifest field says otherwise to anyone who checks.
- **`sp-dashboard/manifest.json.template:13-19`** – `persistDataSynced` and `loadSyncedData` are declared but nothing calls them (your own `CLAUDE.md` says as much); every preference goes to `localStorage`. Nothing on our side enforces those strings, so it changes no capability either way, but the plugin card's permission count is inflated by two.
- **`.vscode/settings.json`** – worth knowing rather than fixing: the committed `chat.tools.terminal.autoApprove` block auto-approves the whole `make` prefix for anyone who opens the repo in VS Code agent mode, and `make release` tags, pushes and runs `gh release create` with whatever credentials are on that machine.

### Questions
- `stars` will have drifted again by the time this merges (the fork reads 11 today) and there is no refresh job on our side. Bump it, or is merging at 9 fine by you?
- `sp-dashboard/manifest.json.template:8` still names `Douglas Cooper` as `author`, so the plugin list will read "ahanel13" while the installed-plugin card on the same settings page reads "Douglas Cooper". Deliberate, to keep the original attribution, or worth listing both?
- `.github/funding.yml` still points the Sponsor button at `buy_me_a_coffee: dougcooper`. Intentional, or left over from the fork?

---

PR #9091, opened 2026-07-16
Add tests for critical task, note, and OAuth paths

ghostkeni:
## Problem

Critical business-logic and auth paths had no direct unit coverage, especially around task mutation helpers, note state mutation logic, and OAuth redirect callback handling. These are high-risk paths because they update persisted state and auth flow behavior.

## Solution

Added targeted unit specs for the most critical uncovered code paths:

- `task.reducer.util.spec.ts` covers key mutation helpers, including parent-child time rollups, done-state timestamp behavior, parent estimate recalculation, delete/orphan-subtask cleanup, and parent side effects.
- `note.reducer.spec.ts` covers note add/update/delete/order/move mutations, shared `deleteProject` handling, load hydration behavior, and selector edge cases.
- `plugin-oauth-redirect.handler.spec.ts` covers same-origin OAuth callback handling, message filtering for security, and listener cleanup behavior.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [x] Other (please describe)

Test coverage for critical mutation/auth paths.

## Checklist

### DIFF
--- a/src/app/features/note/store/note.reducer.spec.ts
+++ b/src/app/features/note/store/note.reducer.spec.ts
@@ -1,0 +1,171 @@
+import { loadAllData } from '../../../root-store/meta/load-all-data.action';
+import { TaskSharedActions } from '../../../root-store/meta/task-shared.actions';
+import { WorkContextType } from '../../work-context/work-context.model';
+import { Note, NoteState } from '../note.model';
+import {
+  initialNoteState,
+  noteReducer,
+  selectNoteById,
+  selectNotesById,
+} from './note.reducer';
+import {
+  addNote,
+  deleteNote,
+  moveNoteToOtherProject,
+  updateNote,
+  updateNoteOrder,
+} from './note.actions';
+
+const createNote = (id: string, partial: Partial<Note> = {}): Note => ({
+  id,
+  projectId: 'projectA',
+  isPinnedToToday: false,
+  content: `note-${id}`,
+  created: 1,
+  modified: 1,
+  ...partial,
+});
+
+const createState = (notes: Note[], todayOrder: string[] = []): NoteState => ({
+  ids: notes.map((note) => note.id),
+  entities: notes.reduce(
+    (acc, note) => {
+      acc[note.id] = note;
+      return acc;
+    },
+    {} as Record<string, Note>,
+  ),
+  todayOrder,
+});
+
+describe('note.reducer', () => {
+  it('should add a note to the top and pin it to today order when applicable', () => {
+    const existing = createNote('n1');
+    const state = createState([existing], []);
+    const newPinned = createNote('n2', { isPinnedToToday: true });
+
+    const result = noteReducer(state, addNote({ note: newPinned }));
+
+    expect(result.ids).toEqual(['n2', 'n1']);
+    expect(result.entities['n2']).toEqual(newPinned);
+    expect(result.todayOrder).toEqual(['n2']);
+  });
+
+  it('should update pinning and keep today order in sync', () => {
+    const note = createNote('n1', { isPinnedToToday: false });
+    const state = createState([note], []);
+
+    const pinned = noteReducer(
+      state,
+      updateNote({ note: { id: 'n1', changes: { isPinnedToToday: true } } }),
+    );
+    expect(pinned.todayOrder).toEqual(['n1']);
+
+    const unpinned = noteReducer(
+      pinned,
+      updateNote({ note: { id: 'n1', changes: { isPinnedToToday: false } } }),
+    );
+    expect(unpinned.todayOrder).toEqual([]);
+  });
+
+  it('should update note order only for non-project contexts', () => {
+    const n1 = createNote('n1');
+    const n2 = createNote('n2');
+    const state = createState([n1, n2], ['n1', 'n2']);
+
+    const forTag = noteReducer(
+      state,
+      updateNoteOrder({
+        ids: ['n2', 'n1'],
+        activeContextType: WorkContextType.TAG,
+        activeContextId: 'TODAY',
+      }),
+    );
+    expect(forTag.todayOrder).toEqual(['n2', 'n1']);
+
+    const forProject = noteReducer(
+      forTag,
+      updateNoteOrder({
+        ids: ['n1', 'n2'],
+        activeContextType: WorkContextType.PROJECT,
+        activeContextId: 'projectA',
+      }),
+    );
+    expect(forProject.todayOrder).toEqual(['n2', 'n1']);
+  });
+
+  it('should remove note and today order entry when deleting', () => {
+    const n1 = createNote('n1', { isPinnedToToday: true });
+    const n2 = createNote('n2');
+    const state = createState([n1, n2], ['n1']);
+
+    const result = noteReducer(
+      state,
+      deleteNote({ id: 'n1', projectId: 'projectA', isPinnedToToday: true }),
+    );
+
+    expect(result.entities['n1']).toBeUndefined();
+    expect(result.ids).toEqual(['n2']);
+    expect(result.todayOrder).toEqual([]);
+  });
+
+  it('should move note to another project', () => {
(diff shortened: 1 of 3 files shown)


johannesjo (commented):
Thanks for this — test-only, no production surface, and it lands direct unit coverage on helpers that were previously only exercised indirectly through the reducer specs. Overall this is good work and I'd like to merge it after two small fixes.

Since reading a test tells you little about whether it would actually catch a bug, I checked these by mutation testing: sabotage the source, confirm the spec goes red. **Five of nine mutations were caught**, which is a genuinely good result — these are real tests, not the assert-nothing kind:

| Mutation | Result |
| --- | --- |
| `reCalcTimeEstimateForParentIfParent` ignores the pending update | caught |
| `deleteTaskHelper` drops the orphan-subtask sweep | caught |
| `updateDoneOnForTask` ignores the explicit payload `doneOn` | caught |
| OAuth origin check disabled | caught |
| `updateNoteOrder` project-context guard removed | caught |

Verification for the rest: full `npm test` green on both TZ variants (Europe/Berlin 12989, America/Los_Angeles 13003, 0 failures), so the `window.alert`/`confirm` spies and the `_resetDevErrorState()` module-state reset don't leak into other specs. ESLint + Prettier are clean on all three files, so the unticked `checkFile` box in the checklist is fine. The `jasmine.isSpy` guard correctly mirrors the existing `stubWindowConfirm` helper in `task.reducer.spec.ts`, and `as never` for the partial `loadAllData` payload is precedented and better than the `as any` some other specs use.

Two mutations survived, and both are worth closing.

### 1. The destroy test has no teeth (`plugin-oauth-redirect.handler.spec.ts:95-101`)

`expect(removeSpy).toHaveBeenCalledWith('message', jasmine.any(Function))` matches *any* function. I changed `ngOnDestroy()` to remove a **different** function — leaking the real listener entirely — and all 5 tests still passed. It asserts the implementation (that a method was called) rather than the behaviour (that the listener stopped firing), which is what `CLAUDE.md` asks for.

To be clear about severity: this isn't a live bug. The handler is `providedIn: 'root'` and lives for the app lifetime, so `ngOnDestroy` rarely fires in production. It's the coverage that's illusory.

This version catches the mutation (verified: red under the leak, green on `master`, Prettier/ESLint clean). It also proves the listener is live before checking it stopped, so it can't silently go vacuous later:

```ts
it('should stop forwarding callbacks after destroy', () => {
  const dispatchCallback = (): void => {
    window.dispatchEvent(
      new MessageEvent('message', {
        origin: window.location.origin,
        data: { type: 'SP_OAUTH_CALLBACK', code: 'oauth-code', state: 'state-1' },
      }),
    );
  };
  // assert the listener is live first, so the post-destroy check cannot pass vacuously
  dispatchCallback();
  expect(serviceSpy.handleRedirectCode).toHaveBeenCalledTimes(1);

  handler.ngOnDestroy();
  dispatchCallback();

  expect(serviceSpy.handleRedirectCode).toHaveBeenCalledTimes(1);
});
```

### 2. The rollup fixture can't distinguish aggregating from copying (`task.reducer.util.spec.ts:95`)

`should update child and incrementally roll up parent times` gives the parent exactly one subtask, whose `timeSpentOnDay` already equals the parent's. That makes "aggregate every subtask" and "just copy this child's values onto the parent" produce identical output, so the test can't tell them apart. I rewrote the parent update to naively copy the child's times — a real bug for any parent with more than one subtask — and the spec stayed green.

One sibling subtask closes it (verified: red under the copy-bug, green on `master`):

```ts
it('should keep other subtasks time in the parent rollup', () => {
  const parent = createTask('parent', {
    subTaskIds: ['child', 'sibling'],
    timeSpentOnDay: { [DAY_1]: 100 },
    timeSpent: 100,
  });
  const child = createTask('child', {
    parentId: 'parent',
    timeSpentOnDay: { [DAY_1]: 30 },
    timeSpent: 30,
  });
  const sibling = createTask('sibling', {
    parentId: 'parent',
    timeSpentOnDay: { [DAY_1]: 70 },
    timeSpent: 70,
  });
  const state = createState([parent, child, sibling]);

  const result = updateTimeSpentForTask('child', { [DAY_1]: 10 }, state);

  // parent keeps the sibling's 70 and absorbs only the child's -20 delta
  expect(result.entities['parent']?.timeSpentOnDay).toEqual({ [DAY_1]: 80 });
  expect(result.entities['parent']?.timeSpent).toBe(80);
});
```

Worth keeping the existing single-subtask case too — it covers the delete-day-when-zero branch, which the sibling test doesn't.

### Optional (happy to merge without these)

Two bits of negative space are untested — both mutations survived, neither is a bug today:

- Nothing pins that an **unrelated** `currentTaskId` survives a delete. `task.reducer.util.spec.ts:184` only asserts it becomes `null`, so hardcoding `currentTaskId = null` passes. A second case with a current task unrelated to the deleted one would pin the `?:` branch.
- The `loadAllData` "no note payload → keep current state" fallback (`note.reducer.spec.ts:143`) is never exercised, so dropping the `? :` guard passes.

Fix the two above and this is good to go. Thanks again!

johannesjo:
Follow-up to my review above — I reviewed this PR in isolation and missed something that matters more than either finding: **this PR collides with your own #9079.**

Both are open, and both **create** (not modify) two of the same files, neither of which exists on `master`:

- `src/app/features/tasks/store/task.reducer.util.spec.ts`
- `src/app/plugins/oauth/plugin-oauth-redirect.handler.spec.ts`

A 3-way merge test confirms it:

```
CONFLICT (add/add): Merge conflict in src/app/features/tasks/store/task.reducer.util.spec.ts
CONFLICT (add/add): Merge conflict in src/app/plugins/oauth/plugin-oauth-redirect.handler.spec.ts
```

Whichever lands second won't merge without a manual rebase. The two specs cover largely the same behaviour under different test names — across both files exactly one `it(...)` name is shared, but functionally they overlap almost completely (explicit `doneOn`, parent estimate on child-done, copy-times-on-last-subtask, orphan-subtask sweep, the incremental rollup, and all the OAuth cases are each written twice).

Rather than have you resolve a conflict between two of your own PRs, it's probably cleanest to consolidate. Each has coverage the other lacks:

**Only in #9079:** `updateStartDateForRepeatableTask` (done → clears `dueDay`), the `Date.now()` fallback when `doneOn` is absent, the `Math.max(0, …)` clamp on negative remaining work, the no-parent path, and the recalc-instead-of-copy branch of `removeTaskFromParentSideEffects` when siblings remain.

**Only in #9091:** the whole of `note.reducer.spec.ts` (no conflict — unique to this PR and worth keeping either way), both `getTaskById` cases, and `updateDoneOnForTask` returning state untouched when the update doesn't toggle done.

Two corrections to my review above, now that I've compared them:

1. **#9079 already solves my finding #1, and does it better than I gave it credit for.** Its destroy test captures the actually-registered listener through a faked `addEventListener` and asserts `toHaveBeenCalledOnceWith('message', messageListener!)` — matching on function *identity*, so it catches the leak mutation that this PR's `jasmine.any(Function)` sails through. If you consolidate onto #9079's file, finding #1 is already handled and you can ignore my snippet.

2. **My finding #2 applies to _both_ PRs.** #9079's `should update parent time incrementally and remove zeroed days` uses the same one-subtask fixture with the parent's times identical to the child's (`{DAY_1: 120, DAY_2: 30}` on both), so it can't distinguish aggregating from copying either. The sibling test from my review is worth adding to whichever file survives.

Suggested path: land one PR, then rebase the other down to just its unique tests (for this one, that's `note.reducer.spec.ts` plus the handful above). Happy either way — mainly flagging it so the second one doesn't hit a surprise conflict.

ghostkeni (commented):
I don't know much about...
But If I can talk with manager.
I want to talk money thing, and houae thing.

ghostkeni (commented):
- [ ] Notion

Issue #8501, opened 2026-06-19
🚨 Flowtime: Rules don't save correctly


beerkumquatpome:
### Version Used

v18.9.1

### Current Behavior

Bug 1:

1. Select Rule-based
2. Add these rules
Rule 1: 1 1 1
Rule 2: 2 2 2
Rule 3: 3 3 3
Rule 4: 4 4 4
3. Click save
4. Rule 2 and 3 have changed to 4 4 4

Bug 2:

1. Select Rule-based
2. Add these rules
Rule 1: 1 1 1
Rule 2: 2 2 2
Rule 3: 3 3 3
Rule 4: 4 4 4
3. Don't click save
4. Click the three-dot button of Rule 1/2/3/4
5. Click Move forward/backward
6. Two rules have changed to 4 4 4

Suggestion: Replace the three-dot button with the delete button (trash bin): The Move forward/backward buttons are not needed

### Expected Behavior

-

### Steps to Reproduce

-

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

KenzoJ:
Hi there, this seems like a relatively straightforward issue. I was able to replicate it on my end. You don't have to assign it to me but know I'll take a stab it this in the coming days.

PR #9092, opened 2026-07-16
fix(focus-mode): stop Formly FieldArray from wiping break rule edits #8501

KenzoJ:
…#8501

## Problem

Saving for Flowtime's rule-based system with more than two rules created unexpected behavior. 
> Rule 1: 1 1 1
> Rule 2: 2 2 2
> Rule 3: 3 3 3

would save successfully as
> Rule 1: 1 1 1
> Rule 2: 3 3 3
> Rule 3: 3 3 3

On every keystroke Formly updated one field, emitted modelChange, and model.set() marked the signal dirty. Change detection then re-evaluated [model]="model()" and pushed the model back into Formly.

With a Formly repeat (field array), that re-push made Formly copy the just-edited property onto every row. Example: typing 4 in the last rule turned all rules into 4/4/4. Save then persisted that already-corrupt model.

## Solution

Formly field arrays need a stable, mutable model reference. A signal that rebinds [model] every keystroke was the main bug; cloning defaultValue on add closed the related shared-object footgun.

- Plain Formly model (not a signal) so repeat rows aren’t corrupted on each keystroke
- Clone defaultValue on add so new rows don’t share/mutate the template object

1. Plain model instead of a signal (dialog-flowtime-settings.component.ts + .html)
Same pattern as Pomodoro settings: a normal object Formly can mutate in place.
```
[model]="model()"
(modelChange)="model.set($event)"
```

2. Stable fields array 
fields went from computed(() => [...]) / fields() to a plain readonly fields: FormlyFieldConfig[]. Formly no longer gets a computed getter in the template. description moved into props so it type-checks under FormlyFieldConfig.

3. Clone on add for repeat-section-type.component.ts
```
const valueToAdd =
  initialValue != null && typeof initialValue === 'object'
    ? { ...initialValue }
    : initialValue;
super.add(undefined, valueToAdd);
```
Each new row gets its own object, so edits don’t rewrite defaultValue or bleed into later adds.

4. Spec updates
Tests use component.model / component.fields instead of signal/computed APIs

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/config/repeat-section-type/repeat-section-type.component.ts
+++ b/src/app/features/config/repeat-section-type/repeat-section-type.component.ts
@@ -41,7 +41,13 @@ export class RepeatSectionTypeComponent extends FieldArrayType {
     const initialValue =
       this.field?.templateOptions?.defaultValue || (fn && fn(this.field));
 
-    super.add(undefined, initialValue);
+    // Clone so Formly does not push/mutate the shared props.defaultValue object.
+    const valueToAdd =
+      initialValue != null && typeof initialValue === 'object'
+        ? { ...initialValue }
+        : initialValue;
+
+    super.add(undefined, valueToAdd);
   }
 
   trackByFn(i: number, item: any): number | string {
--- a/src/app/features/config/repeat-section-type/repeat-section-type.component.ts
+++ b/src/app/features/config/repeat-section-type/repeat-section-type.component.ts
@@ -41,7 +41,9 @@ export class RepeatSectionTypeComponent extends FieldArrayType {
     const initialValue =
       this.field?.templateOptions?.defaultValue || (fn && fn(this.field));
 
-    // Clone so Formly does not push/mutate the shared props.defaultValue object.
+    // Formly's add() already clones, but clone() preserves observe() accessors
+    // on props.defaultValue. Rows now share storage and edits bleed (#8501).
+    // Spread materializes plain values one level deep. Fine for flat defaults.
     const valueToAdd =
       initialValue != null && typeof initialValue === 'object'
         ? { ...initialValue }
--- a/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.html
+++ b/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.html
@@ -2,10 +2,10 @@
 <mat-dialog-content>
   <form [formGroup]="form">
     <formly-form
-      [fields]="fields()"
+      [fields]="fields"
       [form]="form"
-      [model]="model()"
-      (modelChange)="model.set($event)"
+      [model]="model"
+      (modelChange)="model = $event"
     ></formly-form>
   </form>
   <div class="dialog-actions">
--- a/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.html
+++ b/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.html
@@ -2,10 +2,10 @@
 <mat-dialog-content>
   <form [formGroup]="form">
     <formly-form
-      [fields]="fields"
+      [fields]="fields()"
       [form]="form"
-      [model]="model"
-      (modelChange)="model = $event"
+      [model]="model()"
+      (modelChange)="model.set($event)"
     ></formly-form>
   </form>
   <div class="dialog-actions">
--- a/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts
+++ b/src/app/features/focus-mode/dialog-flowtime-settings/dialog-flowtime-settings.component.spec.ts
@@ -58,7 +58,7 @@ describe('DialogFlowtimeSettingsComponent', () => {
   });
 
   it('should initialize with values from GlobalConfigService and convert ms to minutes', () => {
-    const model = component.model();
+    const model = component.model;
     expect(model.isBreakEnabled).toBe(true);
     expect(model.breakRules?.length).toBe(1);
     expect(model.breakRules![0].maxDuration).toBe(25); // 1500000 / 60000
@@ -79,13 +79,13 @@ describe('DialogFlowtimeSettingsComponent', () => {
     });
 
     it('should sort rules by minDuration', () => {
-      component.model.set({
-        ...component.model(),
+      component.model = {
+        ...component.model,
         breakRules: [
           { minDuration: 30, maxDuration: 60, breakDuration: 10 },
           { minDuration: 0, maxDuration: 30, breakDuration: 5 },
         ],
-      });
+      };
 
       component.save();
       const savedConfig =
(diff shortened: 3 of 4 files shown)


johannesjo:
Thanks for digging into this one — you found a genuine bug and the PR **does fix it**. I reproduced #8501 and verified the fix end-to-end. But I want to flag three things before merge: the tests don't actually exercise the fix, only ~7 of the 104 lines do the work, and the stated root cause isn't the real mechanism.

All claims below come from executed tests, not from reading.

## 1. The bug is real, and the PR fixes it

Repro: add 4 rules with the **Add** button, type `1 1 1 / 2 2 2 / 3 3 3 / 4 4 4`, save.

| Code | Saved rules |
| --- | --- |
| `master` (d1ff796) | `1/1, 4/4, 4/4, 4/4` ❌ (exactly as reported) |
| this PR | `1/1, 2/2, 3/3, 4/4` ✅ |

Note the DOM shows the correct values throughout — only the *model* is corrupt, which is why it survives to `save()`.

## 2. Only the `repeat-section-type` clone does the work

I applied each half in isolation (file state fingerprinted per run, so the labels can't drift):

| Variant | dialog | repeat | Result |
| --- | --- | --- | --- |
| A baseline | SIGNAL | NO-CLONE | ❌ FAIL |
| B **clone only** | SIGNAL | CLONE | ✅ **PASS** |
| C **plain-model only** | PLAIN | NO-CLONE | ❌ FAIL |
| D full PR | PLAIN | CLONE | ✅ PASS |

The outcome is decided **entirely** by the `repeat` column. The `[model]="model()"` → plain-property refactor — the PR's headline change and ~90 of its lines — fixes nothing on its own (variant C), and the fix works fine without it (variant B).

## 3. The tests pass with the fix removed

Revert **only** the clone, leaving your component and spec exactly as they are:

```
FINGERPRINT: clone REVERTED (bug is back), PR spec+component untouched
TOTAL: 18 SUCCESS
```

18/18 green while the bug is demonstrably present. The `repro form/model desync` test seeds rows by assigning `component.model`, so it never goes through the `props.defaultValue` → row path where the bug actually lives — it can't fail. No TestBed config change is needed to catch this; the test just has to click **Add**.

## 4. The actual root cause

The PR says a signal re-pushing `[model]` on each keystroke was "the main bug". That round-trip is already a no-op in Formly: `FormlyForm.set model()` early-returns when `_modelChangeValue === model`, and `ngOnChanges` skips the rebuild on the same condition. Variant C confirms it empirically.

What's really happening — instrumented on `master`:

```
dvDescriptors:    [minDuration:GETTER, maxDuration:GETTER, breakDuration:GETTER]
alias_1_2: false  alias_2_3: false        <- rows ARE distinct objects
defaultValueNow:  {minDuration:4, maxDuration:4, breakDuration:4}   <- props.defaultValue got MUTATED
modelRows:        [{1,1,1}, {4,4,4}, {4,4,4}, {4,4,4}]
```

Formly's `observe()` installs **getter/setter accessors** on `props.defaultValue`. Its `clone()` copies accessor descriptors verbatim:

```js
if (propDesc.get) { Object.defineProperty(newVal, prop, propDesc); }
```

So `FieldArrayType.add()` — which *already* calls `clone(initialModel)` — does produce distinct row objects, but every one proxies the same underlying storage. Row 1 escapes only because it's seeded from `{ ...this._defaultRuleInMinutes }` in the constructor. Your spread invokes the getters and copies plain values, breaking the chain (row descriptors flip `GETTER` → `value`). That's also why it incidentally fixes Bug 2 (Move forward/backward).

**This is why the comment matters.** It currently reads:

```js
// Clone so Formly does not push/mutate the shared props.defaultValue object.
```

But Formly already clones — the problem is that the clone *preserves accessors*. A future reader will find `add()`'s `clone(initialModel)`, judge this redundant, delete it, and silently reintroduce #8501. (I nearly did exactly that while reviewing.) Please state the real mechanism and the shortcut's ceiling: the shallow spread only breaks the chain one level, so a nested `defaultValue` would regress silently — and unlike Formly's `clone()`, a spread would quietly mangle a `Date`/`Set`/class-instance default. No current `repeat` consumer uses one, so it's safe today, but it's worth naming.

## Suggested changes

1. **Keep** the `repeat-section-type.component.ts` clone, with a comment describing the accessor propagation + the one-level ceiling.
2. **Drop** the signal→plain model refactor, `computed`→plain `fields`, and the spec's signal-API churn. It's unproven by the matrix above, and it moves the component away from the repo's signals-first convention. Smaller diff, less to re-litigate later.
3. **Replace** the two new tests with one that drives the Add button. This fails on `master` (`Expected $[1].minDuration = 240000 to equal 120000`) and passes with only the clone applied:

```ts
const clickAdd = (): void => {
  fixture.nativeElement.querySelector('.footer button').click();
  fixture.detectChanges();
};
const inputs = (): HTMLInputElement[] =>
  Array.from(fixture.nativeElement.querySelectorAll('.list-wrapper input'));

// Rows added via "Add" are seeded from `props.defaultValue`. Formly's clone()
// copies accessor descriptors verbatim, so without an own-value copy every
// added row proxies the same storage and edits bleed across rows (#8501).
it('keeps rows added via "Add" independent when edited', () => {
  clickAdd();
  clickAdd();
  clickAdd();

  const els = inputs();
  expect(els.length).withContext('4 rows x 3 inputs').toBe(12);

  [1, 2, 3, 4].forEach((n, row) => {
    [0, 1, 2].forEach((col) => {
      els[row * 3 + col].value = String(n);
      els[row * 3 + col].dispatchEvent(new Event('input'));
    });
    fixture.detectChanges();
  });

  component.save();

  expect(updateSection.calls.mostRecent().args[1].breakRules).toEqual([
    { minDuration: 60000, maxDuration: 60000, breakDuration: 60000 },
    { minDuration: 120000, maxDuration: 120000, breakDuration: 120000 },
    { minDuration: 180000, maxDuration: 180000, breakDuration: 180000 },
    { minDuration: 240000, maxDuration: 240000, breakDuration: 240000 },
  ]);
});
```

Because the fix lives in the shared `repeat` type, it also closes the same latent bug for any other consumer that uses an object `props.defaultValue` — worth calling out in the commit message.

## Minor notes

- Moving `description` into `props` is correct (`FormlyFieldConfig` has no top-level `description` — it only exists on `FormlyFieldProps`, which is why the annotation forced the move). But it still won't render: `repeat` is registered with no wrapper and its template never outputs a description. So `FLOWTIME_BREAK_RULES_DESC` is dead either way — a separate small bug, happy to see it fixed here or in a follow-up.
- Unrelated to this PR, but surfaced while investigating: the spec's `FormlyModule.forRoot` omits `extras: { immutable: true }`, and Karma runs zone-ful while the app is zoneless (`main.ts:254`). Neither masks *this* bug (it reproduces under both), but the harness diverging from production on two axes could mask others.

johannesjo:
## Round 2 — all three asks addressed ✅

Re-reviewed at `07d849a6`. The diff is now **2 files, +49 −1**: the `repeat-section-type` clone and one test. Everything from [my last review](https://github.com/super-productivity/super-productivity/pull/9092#issuecomment-5001394490) is resolved. CI is green across all 16 checks.

### 1. The test now genuinely fails without the fix

This was the blocker last time (18/18 green with the fix reverted). Re-ran the sabotage against this head — reverted **only** the clone in `repeat-section-type.component.ts`, left your spec untouched:

```
✗ keeps rows added via "Add" independent when edited
	Expected $[1].minDuration = 240000 to equal 120000.
	Expected $[1].maxDuration = 240000 to equal 120000.
	Expected $[1].breakDuration = 240000 to equal 120000.
	Expected $[2].minDuration = 240000 to equal 180000.
	Expected $[2].maxDuration = 240000 to equal 180000.
TOTAL: 1 FAILED, 16 SUCCESS
```

Clone restored → `TOTAL: 17 SUCCESS`. It drives the Add button and asserts what `save()` writes, so it sits squarely on the `props.defaultValue` → row path where the bug lives. That's the test this fix needed.

### 2. Signal refactor dropped

`dialog-flowtime-settings.component.{ts,html}` are byte-identical to master now (`git diff master...HEAD` on those paths is empty). What's left is only the ~7 lines that actually do the work.

### 3. The comment names the real mechanism

Verified against the installed ngx-formly 7.1.0 rather than from memory — `clone()` at `ngx-formly-core.mjs:174`:

```js
const propDesc = Object.getOwnPropertyDescriptor(value, prop);
if (propDesc.get) { Object.defineProperty(newVal, prop, propDesc); }
else { newVal[prop] = clone(value[prop]); }
```

Accurate as written. The next reader will no longer be tempted to delete this as redundant.

### Blast radius — checked all four `repeat` consumers

The fix is in a shared type, so I traced what changes for each. Only flowtime's behaviour changes at all:

| consumer | `initialValue` | effect of the spread |
| --- | --- | --- |
| `dialog-flowtime-settings` | `props.defaultValue` — plain object carrying `observe()` accessors | **fixed** |
| `boards-form` | `getInitialValue: getNewPanel` → fresh object per call | no-op copy |
| `simple-counter-form` | `getInitialValue: () => ({ ...EMPTY_SIMPLE_COUNTER, ... })` → fresh object | no-op copy |
| `take-a-break-form` | `defaultValue: ''` (falsy) and no `getInitialValue` → `undefined` | guard skips it, unchanged |

The two `getInitialValue` consumers never had accessors installed on their return value (`observeDeep` only walks `props.*`), so the extra spread is a pure no-op there. 923 specs across `boards`/`simple-counter`/`config`/`focus-mode` stay green — though note those suites don't actually click Add for the other three, so that part rests on reading the configs rather than on coverage.

**Your `typeof === 'object'` guard is load-bearing**, which is worth stating explicitly since it's untested. Formly's `add()` is truthiness-based:

```js
this.model.splice(i, 0, initialModel ? clone(initialModel) : undefined);
```

Drop the guard and take-a-break's `undefined` becomes `{...undefined}` → `{}` → truthy → an object gets spliced into what is a list of image-URL **strings**. I confirmed the whole suite stays green with the guard removed, so nothing would catch that regression. Keep it as is.

---

## Minor (both optional, neither blocks merge)

**1. A tighter guard would close the non-plain-object hole for free.** The comment covers the depth ceiling ("one level deep... flat defaults") but not the *shape* one: `typeof x === 'object'` also matches arrays and `Date`s, and spreading those corrupts them (`{...[1,2]}` → `{0:1,1:2}`, `{...new Date()}` → `{}`) — whereas Formly's own `clone()` handles both correctly (`Array.isArray` → `slice().map(clone)`, `Date` → `new Date(...)`).

Since `observeDeep` only installs accessors when `currentValue.constructor.name === 'Object'`, plain-object-ness is exactly the necessary and sufficient condition — so this is a tightening, not defensive padding:

```ts
const valueToAdd =
  initialValue?.constructor === Object ? { ...initialValue } : initialValue;
```

Anything non-plain then falls through to Formly's `clone()`, which already does the right thing. I ran this variant: 923/923 green and `checkFile` clean. No consumer uses an array/Date default today, so this is purely about not leaving a trap — take it or leave it.

**2. The PR description is stale.** It still documents the reverted work — "Plain Formly model (not a signal)", the `computed` → plain `fields` change, `description` moved into `props`, and "Tests use `component.model` / `component.fields`". None of that is in the diff any more; only item 3 (clone on add) survives. Worth a refresh before merge so the squashed commit doesn't claim changes it doesn't make. Your second commit message (`07d849a6`) is accurate and would make a good replacement body.

While you're there: the fix lives in the shared `repeat` type, so it also closes the same latent bug for any future consumer with an object `props.defaultValue` — worth a line, and arguably the title's `focus-mode` scope undersells it.

---

## Verdict

**Ready to merge.** Real bug, minimal fix, and now a test that proves it. Thanks for the rework — this is a much tighter change than round 1.

<sub>Separately, and not for this PR since it no longer touches that file: `FLOWTIME_BREAK_RULES_DESC` still never renders on master — `repeat` is registered with no wrappers and its template outputs no description. Happy to file that separately.</sub>

johannesjo:
Thank you very much! <3

PR #9097, opened 2026-07-17
fix(plugins): cap CalDAV RRULE expansion iterations to avoid main-thread spin

benoptimus:
## Problem

While investigating the calendar-plugin freeze, I found a separate potential
runaway in the CalDAV plugin's RRULE expansion.

In `expandIcalToSearchResults`, the loop caps *emitted* occurrences
(`MAX_OCCURRENCES_PER_EVENT`), but the pre-window skip branch
(`ms < rangeStartMs → continue`) has **no absolute bound**. A high-frequency
unbounded rule (`FREQ=MINUTELY`/`SECONDLY`) with a far-past `DTSTART` steps
through millions of pre-window occurrences before reaching the window, pinning
the renderer thread at ~100% CPU.

## Fix

Add `MAX_ITERATIONS_PER_EVENT` as an absolute per-event iteration cap that
complements the existing emitted-occurrence cap. It's generous enough for
realistic rules (e.g. hourly for years stays well under it).

## Test

Added a regression test (`FREQ=MINUTELY`, `DTSTART` two years before the
window) guarded by a tight per-test timeout — it exceeds the timeout pre-fix
and passes after. Full plugin suite green (69 tests), typecheck + build OK.


johannesjo (commented):
Reviewed this closely and verified the claims locally — **the diagnosis is right and the fix works**. I reproduced the runaway and measured the cap's effect:

- **pre-fix:** 1,137,001 iterations ≈ **5.5s** of synchronous work (~0.21M iter/sec) for the `FREQ=MINUTELY` / far-past-`DTSTART` case
- **post-fix:** bounded at 100k iterations ≈ **0.5s** per event

Suite green (69/69), typecheck clean. One test issue I'd fix before merge, plus two smaller notes.

## 1. The regression test's assertion can never fail

`expect(events.length).toBeLessThanOrEqual(1000)` is unfalsifiable through this code path:

- `callBacklog` → `getNewIssuesForBacklog` → `fetchEvents(cfg, { maxResults: 100 })` (`plugin.ts:1354`, sliced at `:1093`). The path structurally cannot return more than 100, so a 1000 threshold is unreachable by construction.
- Instrumenting the real values: **post-fix `events.length` is 0; pre-fix it's 100.** Both satisfy `<= 1000`.

So the entire regression guard currently rests on the 2000ms timeout. That guard *is* real — I verified that deleting the cap line makes the test fail (`Test timed out in 2000ms`, ~8.0s), and that `vi.useFakeTimers()` does **not** neutralize vitest's timeout (it plausibly could have). But it's wall-clock-dependent: the passing case measures ~590–740ms (~990ms cold) against the 2000ms budget, a ~2–3x margin — the kind of thing that flakes on a loaded CI runner.

Both issues go away with one change:

```diff
-          expect(events.length).toBeLessThanOrEqual(1000);
+          // Pre-fix the iterator reaches the window and emits (100 = the
+          // getNewIssuesForBacklog maxResults cap); with the bound in place it
+          // never gets there. 0 is therefore proof the cap fired.
+          expect(events.length).toBe(0);
         },
-        2000,
+        10000,
```

Verified both directions:

- **with** the cap → passes
- **without** the cap → fails with `AssertionError: expected 100 to be +0`

i.e. it then fails on the **assertion rather than the clock**, so the timeout degrades to a pure backstop and the test stops depending on timing at all.

## 2. The cap is per-event; the aggregate is still unbounded (not a blocker)

`responses.flatMap(expandIcalToSearchResults)` (`:1069`) and `fetchEvents`'s loop over read calendars share no budget, so total work scales with the number of runaway masters. Measured **with the fix applied: 20 runaway VEVENTs = 12.6s of synchronous work.**

Still an ~11x per-event win, so I wouldn't hold the PR for it — but "avoid main-thread spin" holds for a single event rather than a calendar full of them. Worth a note in the comment so the next reader doesn't assume it's fully bounded; a budget shared across the expansion pass would actually close it.

## 3. Truncation is to *zero*, and the hourly margin is 11.4 years (nit)

100k iterations = 11.4 years of hourly. Measured against the plugin's real window math (counts at the `expandIcalToSearchResults` level, before the backlog's 100-slice):

| rule | pre-fix emits | post-fix emits |
|---|---|---|
| `FREQ=HOURLY`, DTSTART 12y ago | 1000 | **0** |
| `FREQ=MINUTELY;INTERVAL=15`, DTSTART 4y ago | 1000 | **0** |
| `FREQ=HOURLY`, DTSTART 5y ago | 1000 | 1000 |
| `FREQ=DAILY`, DTSTART 12y ago | 56 | 56 |

These are rare enough that I think the trade-off is the right one. But affected events disappear *entirely* rather than being partially truncated — which is precisely the failure mode the existing comment just above the loop warns about ("silently produces zero in-window results"). Suggest the new comment say "truncated to zero" rather than "truncated", and soften "hourly for years stays well under it", since that only holds below ~11 years.

johannesjo:
Since there was no response for a couple of weeks, I've applied the review items directly on the branch so this can land (your commit is untouched; I merged master in and added one commit on top):

- The regression test now asserts `events.length` to be exactly `0`: pre-fix the iterator reaches the window and emits 100 events (the `getNewIssuesForBacklog` maxResults cap); with the cap in place it never gets there. So the test fails on the assertion instead of the wall clock, and the timeout is a pure 10s backstop (measured locally: ~0.8s passing; ~3.7s pre-fix, failing on `expected 100 to be +0`).
- The `MAX_ITERATIONS_PER_EVENT` comment now states the concrete margin (100k steps is roughly 11 years of hourly occurrences), that a rule hitting the cap before reaching the window is truncated to zero in-window results, and that the cap is per event (total work still scales with the number of runaway masters).

Thanks a lot for the careful find and the clean fix @benoptimus, and congrats on your first contribution!

Issue #9057, opened 2026-07-15
💡 Apple Shortcuts support


SmokeSlate:
### Problem Statement

So as someone who migrated from Apple Reminders recently to Super Productivity, one of the features I’ve been missing has been Apple shortcuts support, it’s very nice for automation and other things and it’s possible to use the API for some things, but after a certain point it becomes quite hard to work with.

### Possible Solution

The addition of a few actions such as adding reminders, modifying them, deleting them, searching for them, marking them as complete, etc.

### Describe alternatives you've considered

As described, the API can be used, but it is hard to work with compared to normal shortcut actions.

MannXo:
I can pick this up. Is it available?

johannesjo:
This would be great @MannXo ! I'd imagine this to be a great improvement for iPhone users!

fanatique:
I prototyped a Siri App Intent for this and measured what it can and cannot do. Posting the results since they were not obvious up front (+ they're not exactly great).

**Setup:** iPhone 16e, iOS 26.6.2, debug build off `master`. A minimal `AppIntent` ("Add Task") plus `AppShortcutsProvider`, writing through the existing `ShareInbox` so `NativeCaptureImporter` handles the capture with its usual exactly-once guarantees. No changes to sync or task state.

**What works** 🚀

- iOS compiles `Metadata.appintents` into the bundle. The intent shows up under App Shortcuts in the Shortcuts app with no extra setup and runs fine when tapped.
- Writing through `ShareInbox` needs no new storage path. The capture becomes exactly one task.

**What does not work** 😢

Running the intent while the app is not in the foreground does not boot the WebView. iOS starts the app process without a UI scene, and Capacitor's bridge is created by the root view controller, so it never exists. The capture stays in the container and becomes a task  and therefore an op, and therefore synced _only on the next manual launch_.

Measured in both states: app force-quit, and app merely suspended in memory. Same result. A task dictated on the phone is invisible on every other device until the user opens the app.

**Alternatives checked**

- Local notification to wake the app: not possible, only remote silent push wakes an app, and iOS skips even that after a manual force-quit. There is no APNs infrastructure in the project.
- `BGAppRefreshTask`: cannot be triggered programmatically, no timing guarantee.
- Syncing from Swift: would mean a second sync client (op-log, vector clocks, auth are all TypeScript).
[...]

PR #9121, opened 2026-07-17
feat(tasks): add create-task/complete-task URL scheme actions

MannXo:
## Problem

Fixes/addresses #9057 ("Apple Shortcuts support"): no way to add or complete a task from outside the app (e.g. via an iOS Shortcut, Siri, or any other URL-triggered automation).

Full native App Intents / SiriKit integration (what would show up as a first-class action inside the Shortcuts app, with natural Siri phrases) needs Swift + Xcode to build and verify, and there's no existing code in this repo to build on (confirmed: zero references to `AppIntent`/`SiriKit`/`Intents.framework` anywhere in `ios/`). Rather than write native Swift I can't compile or test, this PR ships the smallest thing that gives real automation value and is fully buildable/testable today: two actions reachable through the app's *existing* custom URL scheme (`create-task`, `complete-task`), which Shortcuts' built-in "Open URLs" action can already target. Native App Intents would be a good, larger follow-up for whoever has Xcode set up.

## ⚠️ Breaking change (please review as a deliberate decision)

The pre-existing desktop `create-task` protocol action (shipped since **v14.2.4**) previously parsed **Short Syntax** on the title, like a manually typed task. This PR stores the title **literally** (`isIgnoreShortSyntax: true`) on both platforms, treating a URL as untrusted external content — so `+Project` / `#tag` / `@date` tokens are no longer interpreted, a URL can no longer create tags/projects, and an explicit `projectId` can't be silently overridden by the title. This is a change to shipped desktop behavior, so I've ticked **Breaking change** below. Docs (`3.01-API.md`, `3.03-Keyboard-Shortcuts.md`) updated to match.

The leaner, no-behavior-change alternative is to keep Short Syntax and just document that `+Project` overrides `projectId` — happy to switch to that if you'd prefer the non-breaking path. (Per round-4 review discussion.)

## Solution

- **Desktop (Electron):** `superproductivity://create-task` was already a recognized protocol action shipped since v14.2.4 (parsing Short Syntax on the title). This PR replaces that warm-path behavior — the title is now stored literally (see the breaking-change note above) — and adds a new sibling `complete-task` action. Also enriched `create-task` to accept `notes`/`projectId` via query params (kept the old path-segment title form working too, for backward compatibility).
- **Mobile (iOS via Capacitor):** added the same two actions (`create-task`, `complete-task`) on the existing `com.super-productivity.app://` scheme, via the `appUrlOpen` listener in `main.ts` (previously a no-op stub). This is what an iOS Shortcut's "Open URLs" action would target. Android is not supported yet.
- **One shared handler, both platforms:** `AppUriTaskActionsService` merges the Electron IPC stream and the Capacitor pending-action stream into one code path, buffers until `DataInitStateService.isAllDataLoadedInitially$` fires (so a cold launch never races the app's own data hydration — this was the trickiest part to get right), and calls the existing `TaskService.add()`/`TaskService.setDone()` — no new task-mutation logic, reuses what's already there.
- **Feedback:** a snack notification on success/failure, since a URL-triggered action from outside the app has no other feedback channel. New translation keys added the normal way (`en.json` → `npm run int` → `t.const.ts`).
- **Safety:** an unknown `projectId` fails the whole action with an error snack (no task is created) rather than being silently dropped, mirroring `PluginBridgeService.addTask`. A whitespace-only title is rejected. `create-task` caps the stored title at 300 chars and notes at 100k (matching the EML import path); `complete-task`'s title is only a lookup needle so it is not held to that cap. `complete-task` matches case-insensitive/substring against non-done, non-subtask tasks only, preferring an exact match and erroring on an ambiguous one.
- **Docs:** added a new "URL Scheme Actions" section to the API reference wiki page (`3.01-API.md`) with the parameter limits, updated `3.03-Keyboard-Shortcuts.md`, plus a one-line cross-reference in the web-vs-desktop capability comparison (`3.05-Web-App-vs-Desktop.md`) noting this isn't available in the browser build.

### Manually verified end-to-end

Built and ran the actual Electron app, then simulated the OS opening it via the registered protocol (exactly what a real Shortcuts/automation trigger does): `create-task?title=...` → task appeared in the active project/context with the "Added task ..." snack; `complete-task?title=...` → task moved to Completed with the "Completed task ..." snack. Screenshots available on request.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for this — the framing in the description is exactly right, and the "ship the buildable thing now, leave native App Intents as a follow-up" call is a good one. A few things I want to call out as genuinely well done before the findings:

- The log-privacy discipline in `electron/protocol-handler.ts` is better than most code in the repo: it logs the scheme+action only, logs the path-part *count* rather than the parts, and `protocol-handler.test.cjs` asserts the title never reaches the exportable log. That's the rule enforced by a test rather than a comment.
- `parseAddTaskFromAppUriPayload` degrading gracefully on non-string `notes`/`projectId` instead of rejecting the whole payload is the right call for an untrusted boundary, and it's tested.
- Your dead-code claim checks out — I verified `ipcAddTaskFromAppUri$` had zero subscribers on the base commit.

I ran the suites locally: CI-equivalent green, `npm run checkFile` clean on all five changed `.ts` files, 1281 tasks-feature specs pass, only `en.json` touched. The findings below are all from reading/executing the code, not from CI.

---

## Blockers

### 1. A whitespace-only title completes an arbitrary task

`complete-task?title=%20` — or `?title=+`, since `+` decodes to a space in a query string, which is what a naively-built Shortcut URL produces.

`searchParams.get('title')` returns `' '`, which is truthy, so the `!title` guard at `parse-app-uri-task-action.ts:35` passes. Then at `app-uri-task-actions.service.ts:109` `' '.trim().toLowerCase()` is `''`, and `'anything'.includes('')` is `true` for every string — so the `find` at `:116` returns **the first non-done task in the store** and `:126` marks it done.

I didn't take this on inspection. I added a throwaway spec asserting `setDone('t1')` for `{type: 'complete', title: ' '}` and ran it against the real service — it passed, confirming the wrong-task completion end to end.

This is silent wrong-task completion from an unauthenticated trigger, and it then syncs to every device. `add-task?title=%20` likewise creates a blank-titled task.

Suggested fix — validate the trimmed value in the parser:

```ts
const title = urlObj.searchParams.get('title')?.trim();
if (!title) {
  return null;
}
```

plus a `needle.length > 0` guard in `_handleComplete` so the service is safe independently of its caller.

### 2. The feature is entirely absent on Android, but the docs say it works there

Two separate gaps:

- `src/main.ts:523` puts the `appUrlOpen` listener inside `if (IS_IOS_NATIVE)` (`Capacitor.getPlatform() === 'ios'`), so the handler never registers on Android.
- More fundamentally, `android/app/src/main/AndroidManifest.xml` scopes its intent-filters **by host** — there are filters for `oauth-callback` and `plugin-oauth-callback`, but none for `add-task`/`complete-task`. So the intent doesn't resolve to the app at all. (iOS is fine: `CFBundleURLSchemes` isn't host-scoped.)

Meanwhile the new wiki table lists "iOS / Android (Capacitor)" and the commit message says "iOS/Android".

Fix is both parts: add the intent-filters, and gate on `IS_NATIVE_PLATFORM` rather than `IS_IOS_NATIVE` — that's what `OAuthCallbackHandlerService` uses for the same `appUrlOpen` mechanism. Alternatively, scope the docs to iOS-only for this PR and leave Android as a follow-up.

### 3. The task title and notes land in the exportable log

`src/main.ts:553`:

```ts
Log.log('iOS app URL open', event.url);
```

The URL is now `...//add-task?title=<user title>&notes=<user notes>`. `src/app/core/log.ts:1` states "Log history is exportable and often shared in bug reports", and the repo rule is that user content must never reach it.

This line predates the PR, but before it the scheme only carried OAuth callbacks — routing task titles through it is new. You already handle this correctly on the Electron side, so it reads as an isolated slip rather than a systemic gap. `OAuthCallbackHandlerService` shows the pattern: `SyncLog.log('OAuthCallbackHandler: Received URL')`, no URL.

### 4. On desktop, a cold launch can drop the action entirely

This is the case the description calls out as "the trickiest part to get right", and it's solved on Capacitor but not on Electron.

`ipcEvent$` (`src/app/util/ipc-event.ts`) returns a plain `Subject` with no replay. The `concatMap`/`isAllDataLoadedInitially$` gate sits *downstream* of that Subject, so it buffers against data hydration — it cannot buffer a message that was delivered while nothing was subscribed. `pending-capacitor-app-uri-action.ts` explains this problem precisely in its own doc comment; the Electron path just has no equivalent.

The delivery side offers no protection either: `createWindow` does **not** await the page load (`electron/main-window.ts:326` is `mainWin.loadURL(url).then(...)`, unawaited, with `return mainWin` at `:431`), and `electron/start-app.ts:520` drains pending URLs on a fixed `setTimeout(..., 1000)` from window construction. So the drain can fire before `AppUriTaskActionsService` exists, and the action is dropped with no task, no snack, and no error.

To be precise about confidence: the *missing buffer* is structural and certain; whether the 1 s timer actually loses the race is timing-dependent and I did not measure a real cold start.

The right primitive already exists and is unused here — `IPC.APP_READY` (`electron/main-window.ts:384`, `getIsAppReady()` at `:114`), which the renderer already signals. Gating the drain on that instead of the 1 s guess fixes it properly. A `shareReplay({ bufferSize: 1, refCount: false })` on the two new observables would also close the window.

---

## Important

### 5. `add-task` already means something different on desktop

`electron/protocol-handler.ts:130` already has:

```ts
case 'add-task':
  showOrFocus(mainWin);
  mainWin.webContents.send(IPC.SHOW_ADD_TASK_BAR);
  break;
```

It mirrors the `globalAddTask` shortcut (the Wayland global-hotkey workaround from #7114) and opens the add-task **input bar** for the user to type into. So `add-task` is taken on desktop with entirely different semantics — which is presumably why the original desktop action was called `create-task`.

The new wiki table lists `add-task` (mobile) and `create-task` (desktop) as adjacent rows for the same operation, which invites exactly the wrong guess: someone reading it will try `superproductivity://add-task?title=Buy%20milk` on desktop and get an empty add-task bar with the title silently ignored.

Worth resolving now rather than after it's documented as a public API — either pick one name that's free on both platforms, or have `create-task` accept both and document the desktop `add-task` collision explicitly.

### 6. "First match" is less predictable than it looks

`app-uri-task-actions.service.ts:116` matches case-insensitive substring and takes the first hit from `selectAllTasks`. Three compounding issues:

- `taskAdapter` has no `sortComparer`, so `selectAll` returns `state.ids` insertion order — roughly oldest-first, and reorderable by sync/import. `complete-task?title=standup` will tend to complete the *oldest* standup, not today's.
- `selectAllTasks` is not filtered by archived projects, unlike `selectAllTasksInActiveProjects` (`task.selectors.ts:153`), so it can complete a task in a project the user archived and can't see.
- It includes subtasks, so it can complete a subtask of an unrelated parent.

The success snack echoes the *matched* title, which under substring matching can look plausible while pointing at the wrong entity. Suggest: `selectAllTasksInActiveProjects`, exclude `parentId` tasks, prefer an exact case-insensitive match, and fall back to substring only when it's unambiguous — erroring rather than guessing when several match.

### 7. Without `projectId`, the task doesn't go to the Inbox

`TaskService.add` → `createNewTaskWithDefaults` (`task.service.ts:1449`) derives from the **active work context**: an open project gives `projectId: workContextId`; an open tag gives `tagIds: [workContextId]`; Today gives `dueDay: today`. Your explicit `projectId` correctly wins (`additional` spreads last) — but omitting it means the same Shortcut lands tasks in different projects depending on which screen was last open, which on a cold launch is restored from the previous session. That's hard to reason about from an automation's perspective.

The wiki currently says the task is "added without a project", which isn't what happens. Related: an unknown `projectId` is dropped at `:89` but `:101` still fires the SUCCESS snack, so the automation gets no signal it was ignored. (For comparison, `PluginBridgeService` validates references and throws.)

### 8. On macOS the snack — the only feedback channel — is never seen

`create-task` and `complete-task` are the only protocol actions that never call `showOrFocus`; compare `oauth-callback:68`, `plainspace-connect:117`, `add-note:127`, `add-task:131`. Windows/Linux get focus incidentally via the `second-instance` handler at `:191`, but the macOS `open-url` path at `:201` goes straight to `processProtocolUrl`. Same on any platform started minimized to tray.

Worth a deliberate decision: either `showOrFocus` for these actions, or drop the success snack and keep only the error path — but not feedback the user can't observe.

### 9. Two parsers for one documented API, already diverging

`electron/protocol-handler.ts` and `parse-app-uri-task-action.ts` accept different inputs:

- Desktop accepts the path form `create-task/<title>`; mobile returns `null` for `add-task/Buy%20milk`.
- Desktop switches on the raw `urlObj.hostname` (`:56`), so `superproductivity://Create-Task?title=x` hits `default:` and logs "Unknown protocol action" — while `parse-app-uri-task-action.ts:33` lowercases and has a passing spec asserting `Add-Task` works.

Since this is now a documented public URL API, consider moving the parser into `electron/shared-with-frontend/` and having `protocol-handler.ts` consume it. At minimum, lowercase the host on the Electron side.

---

## Minor

- **The buffering has no test coverage.** I deleted the entire `concatMap(isAllDataLoadedInitially$)` gate and all 6 specs still passed — the spec mocks `isAllDataLoadedInitially$` as `of(true)`, which emits synchronously, so the behaviour it's meant to prove is never exercised. A `Subject` that emits *after* the action would make it real. (The gate itself is correct — `take(1) + shareReplay(1)` means the inner observable completes, so `concatMap` doesn't stall.)
- **`ReplaySubject(1)`** buffers one action, so an iOS Shortcut with two "Open URLs" steps on a cold launch silently loses all but the last. It also retains the value for the process lifetime, so any future second subscriber would replay a stale action.
- **Short syntax applies to these titles** — `isIgnoreShortSyntax` stays `false`, so `title=Buy milk +Work #urgent` creates tags/projects and can fight the explicit `projectId` param. Precedent is split (`eml-drop.service.ts` opts out for untrusted content, `plugin-bridge` doesn't), so this is defensible — but the wiki should say so, since it's surprising for an automation API.
- **Bootstrap pattern:** injecting the service into `AppComponent` purely for its constructor works, but `src/main.ts:300` already establishes `APP_INITIALIZER` + service-owns-its-listener for the identical deep-link mechanism (`OAuthCallbackHandlerService`).
- **Heads-up:** #8515 adds a `quick-add` case to the same `processProtocolUrl` switch and edits the same `protocol-handler.test.cjs`, so expect a textual conflict on merge (not a semantic one).

---

## Chased and cleared

Noting these so nobody re-investigates them — all three look suspicious but are fine:

- **`concatMap` does not deadlock.** `isAllDataLoadedInitially$` is `take(1) + shareReplay(1)`, so the inner observable completes and the queue drains.
- **No subscription accumulation** in `_handleComplete` — RxJS 7's `Subscription.add` early-returns on an already-closed teardown, and `take(1)` closes it synchronously.
- **A cross-context `projectId` is placed correctly** — `additional` spreads last in `createNewTaskWithDefaults`, and the reducer keys off `task.projectId`.

---

The blockers cluster on one theme: this is a new untrusted-input boundary, and it'd benefit from a single hardening pass (trim/validate at the parser, tighten the match, decide the platform story) rather than scattered patches. Happy to look again once you've had a pass at these.

MannXo:
Thanks for the thorough review. Pushed a commit addressing all of it:

- Whitespace-only title no longer bypasses the empty-title check.
- Scoped the docs to iOS + desktop for now; called out Android as a follow-up rather than shipping something I can't test here (no SDK/emulator).
- Stopped logging the raw URL, only whether it was recognized.
- Gave the two Electron IPC channels their own ReplaySubject buffer so a message arriving before the app subscribes isn't dropped, and switched the pending-URL drain to wait for the real APP_READY signal instead of a fixed timer.
- Renamed the mobile action to `create-task` so it doesn't collide with desktop's existing `add-task`.
- Completion matching now prefers an exact match, skips subtasks and archived projects, and errors instead of guessing when several matches are equally plausible.
- An unresolvable projectId now fails with an error instead of silently dropping it and reporting success.
- Added `showOrFocus` so the snack is actually visible on macOS and when minimized to tray.
- Fixed the case-sensitivity mismatch between the two parsers.
- Replaced the buffering test's synchronous mock with one that actually proves the gate works (confirmed it fails if you delete the gate).

Separately: while doing a final end-to-end pass I found that firing one `create-task` action creates two tasks, not one. My own code fires `TaskService.add()` exactly once, confirmed by instrumenting it directly, but something else also calls it, from a stack trace that's pure minified RxJS internals with no identifiable caller. I checked and this isn't from anything in this round of fixes; it reproduces the same way on the original commit before I touched anything. Haven't confirmed if it also happens in a production build. Didn't want to guess at a fix for something I can't fully explain, especially since it touches the shared task-add path rather than just this feature. Let me know if you want me to keep digging or if it's worth its own issue.

johannesjo:
Thanks for the thorough pass — nine of the ten findings from last time are genuinely fixed, and several are fixed better than I suggested. Before any of that, though:

## I have to correct something I told you last round

In my previous review I wrote:

> Your dead-code claim checks out — I verified `ipcAddTaskFromAppUri$` had zero subscribers on the base commit.

**That was wrong, and it's the cause of the double-task bug you found.** I apologise — you built on that in good faith.

`ipcAddTaskFromAppUri$` had exactly one subscriber on the base commit, and still does:

```ts
// src/app/features/tasks/store/task-electron.effects.ts:203
handleAddTaskFromProtocol$ = createEffect(
  () => ipcAddTaskFromAppUri$.pipe(tap((data) => {
    ...
    this._taskService.add(data.title);   // :215
  })),
  { dispatch: false },
);
```

`TaskElectronEffects` is registered at `src/app/root-store/feature-stores.module.ts:162` (`...(IS_ELECTRON ? [TaskElectronEffects] : [])`), and `FeatureStoresModule` is in the bootstrap graph at `src/main.ts:137`.

On the base commit:

```
$ git grep -n "ipcAddTaskFromAppUri" 6756e9998c -- src
6756e9998c:src/app/core/ipc-events.ts:53:              (definition)
6756e9998c:src/app/features/tasks/store/task-electron.effects.ts:30:   (import)
6756e9998c:src/app/features/tasks/store/task-electron.effects.ts:205:  (subscription)
```

So `superproductivity://create-task/<title>` **already worked on master** — the path-segment form, title only. It was never dead code. I only grepped for *service* subscribers; an NgRx effect's subscription has no visible call site (it's a class field, subscribed by `EffectsRunner` because the class appears in an `EffectsModule.forFeature([...])` array), so it didn't show up in the way I looked. My mistake, and a good lesson: "zero subscribers" needs a `*.effects.ts` check too.

## Blockers

### 1. That's your double-task bug — and this PR is the regression

`ipcAddTaskFromAppUri$` is a single module-level const over a single multicast subject. This PR adds a **second** subscriber (`AppUriTaskActionsService` at `app-uri-task-actions.service.ts:46`) without removing the first. One `webContents.send` → one preload listener → one `next()` → **two `TaskService.add()` calls**.

This matches your report exactly, including the part that stumped you: a stack of "pure minified RxJS internals with no identifiable caller" is precisely what an NgRx effect's `tap` looks like — it's invoked from `EffectsRunner`'s subscription, so there is no application frame on the stack.

One correction on the diagnosis, though: it does **not** reproduce on master. `76b7677620` is this PR's own first commit, not the base:

```
$ git merge-base --is-ancestor 76b7677620 6756e9998c   # → false
$ git branch -a --contains 76b7677620                  # → only feat/issue-9057-app-uri-task-actions
```

That commit is already the one that introduced the duplicate. Corroborating: `complete-task` does **not** double, because `COMPLETE_TASK_FROM_APP_URI` is new in this PR and has exactly one consumer.

**Zero-cost field confirmation** if you want to see it yourself: fire `create-task?title=Foo&notes=Bar&projectId=<real-id>`. One of the two tasks will have **no notes and no project** — the effect ignores both and only passes `data.title`. It also lands first, since effects subscribe at `ROOT_EFFECTS_INIT`, before `AppComponent`.

**Fix:** delete `handleAddTaskFromProtocol$` and let the new service own the channel. The service is a strict superset of it — it handles `notes`, validates `projectId`, and buffers on `isAllDataLoadedInitially$`, which the effect does not. Keeping both is not an option; the effect also bypasses the hydration buffer this PR exists to add, so the ghost task can be created against a pre-hydration store.

Worth noting the second-order effect of this: the channel is a `ReplaySubject(1)` now, so on a cold launch the effect *also* receives the replayed payload, where previously (plain `Subject`) it would simply have missed it.

### 2. CI is red, and it's this PR

The `Tests` job fails with 12 errors in `src/app/features/planner/store/planner.selectors.spec.ts` (`selectAllTasksDueToday` returning `[]`). It looks unrelated, and the file isn't touched here, but it is caused by the new spec.

`app-uri-task-actions.service.spec.ts` overrides `selectAllTasksInActiveProjects` via `provideMockStore({ selectors: [...] })`. **NgRx memoized selectors are module-level singletons**, so `setResult()` survives `TestBed.resetTestingModule()` and leaks into every spec file that runs later in the same Karma session — including specs that never mention MockStore but consume a selector composed from the overridden one.

Verified both directions:

```
# planner spec alone — green in BOTH TZ variants
ng test --include .../planner.selectors.spec.ts                    → 37/37 SUCCESS

# add the new spec
ng test --include .../app-uri-task-actions.service.spec.ts \
        --include .../planner.selectors.spec.ts                    → 12 FAILED, 37 SUCCESS

# same run, with resetSelectors() added to the new spec's afterEach
                                                                   → 73/73 SUCCESS
```

**Fix:** `afterEach(() => TestBed.inject(MockStore).resetSelectors())` in both describes that use `provideMockStore`. Note `npm run test:file` can never catch this — the spec passes in isolation, so it only shows up in the full suite.

### 3. Whitespace title still creates a blank task on desktop

The `complete-task` half is properly fixed — the `needle` guard in `_handleComplete` works, and I confirmed it rejects `{type:'complete', title:' '}` without calling `setDone`. Thank you for putting it at the service level rather than only in the parser.

The `add` half isn't. Only the *mobile* parser (`parseAppUriTaskAction`) trims; the Electron path never goes through it, and nothing on that path validates:

- `protocol-handler.ts:79` — `searchParams.get('title')` for `?title=%20` is `' '`, truthy, so it sends. (Also `?title=+`, which decodes to a space.)
- `parseAddTaskFromAppUriPayload` only checks `typeof title === 'string'`.
- `_handleAdd` uses `action.title` with no guard, unlike `_handleComplete`.

I proved the chain with a throwaway spec against the real service:

```
parseAddTaskFromAppUriPayload({title: ' '})  →  {title: ' '}      // not rejected
service                                       →  add(' ', false, {})  // blank task
```

So `superproductivity://create-task?title=%20` creates a whitespace-titled task. The new wiki line — "Whitespace-only is rejected (no task is added)" — isn't true on desktop. Mirroring the `_handleComplete` guard in `_handleAdd` fixes both the behaviour and the doc.

### 4. `ipcEventReplay$` is a footgun for any NgRx effect that consumes it

Not a live bug once #1 is fixed, but worth a warning in that util's doc comment, because it's a new *generic* helper others will reach for.

`@ngrx/effects` (21.1.1) `defaultEffectsErrorHandler` re-subscribes a failed effect up to `MAX_NUMBER_OF_RETRY_ATTEMPTS = 10` times (`ngrx-effects.mjs:174`). Re-subscribing to a `ReplaySubject(1)` **replays the buffered value**. So if an effect consuming a replaying IPC channel ever throws, one URL can produce up to 11 actions instead of one. With a plain `Subject` it produces one. The existing `ipcEvent$` doc already warns "most channels should NOT replay"; adding "and never consume one from an NgRx effect" would close the loop.

## What's genuinely fixed — nice work

- **Whitespace `complete-task`** — guarded at the service level, so it's safe regardless of caller. Confirmed by test.
- **Android** — scoping it out of the docs rather than shipping something you can't test is the right call, and the docs now say so plainly instead of implying it works.
- **Log privacy** — `Log.log('iOS app URL open', { recognized: !!action })` is exactly right.
- **Cold-launch drop** — `IPC.APP_READY` replacing the 1 s timer is the correct primitive.
- **`add-task` collision** — renamed to `create-task` *and* the desktop collision documented, which is more than I asked for.
- **Matching** — exact-first, subtasks excluded, `selectAllTasksInActiveProjects`, and erroring on ambiguity rather than guessing. This is a much better design than "first substring match".
- **`projectId`** — now errors instead of silently dropping and reporting success.
- **`showOrFocus`**, and both parsers lowercasing the host.
- **The buffering test is real now.** I sabotage-checked it: deleting the whole `concatMap(isAllDataLoadedInitially$)` gate turns it red (previously all 6 specs stayed green). That's the right kind of test.
- Translations are correct — only `en.json` touched, keys match `t.const.ts`.

## Minor

- **No e2e coverage** (`git grep -l "create-task" -- e2e` → empty), and Karma is structurally blind to this feature: `IS_ELECTRON` is userAgent-based, so both IPC observables are `EMPTY` under test — as your own spec comment notes. That's exactly why #1 and #2 both got through a green local suite. Worth at least a test asserting the channel has a single consumer.
- `ReplaySubject(1)` buffers one action and retains it for the process lifetime, so a Shortcut with two "Open URLs" steps on a cold launch still loses all but the last.
- `processProtocolUrl`'s 10 s deferred re-drain (`protocol-handler.ts:46-49`) is a dead safety net — the `setTimeout` closes over a `mainWin` that is always null in that branch, so it re-queues forever and never delivers. Pre-existing, not yours, but adjacent if you're in there.
- `mainWinModule.isAppReady` is never reset to `false`. Harmless today (the `activate → createMainWin()` re-entry is unreachable because `mainWin` is never set back to `null` in `start-app.ts`), which does mean the new `if (getIsAppReady())` branch is effectively dead — the `ipcMain.once` path is the one that runs.
- Win/Linux `second-instance` now calls `showOrFocus` twice (the handler, plus the new cases).
- **Heads-up:** #8515 is still open and touches both `electron/protocol-handler.ts` and `electron/protocol-handler.test.cjs` — expect a textual conflict.

---

Sorry again for the bad steer on the dead-code claim — that one's on me, and it cost you a confusing debugging session. The good news is the fix for the mystery bug is a deletion, not new code. Once #1 and #2 are in I'm happy to take another look.

MannXo:
Thanks for the follow up, and no worries on the earlier steer. Pushed a commit addressing all of it:

- Removed `handleAddTaskFromProtocol$` from `TaskElectronEffects`. That was the actual source of the double task, thanks for tracking it down.
- `_handleAdd` now trims and rejects a whitespace-only title, same as `_handleComplete` already did.
- Added a warning in `ipcEventReplay$`'s doc comment about consuming it from an effect.

Full suite is green locally before this push. Let me know if anything else needs another look.

johannesjo:
## Round 3 review — `f63ec32eec`

All four round-2 blockers are genuinely fixed, and CI is green (22/22). Thanks for the careful turnaround on both — deleting the duplicate effect was the right call, and removing `task-electron.effects.spec.ts` wholesale is correct too: I checked, it existed at base and tested *only* `handleAddTaskFromProtocol$` (3 `it()`s), so no unrelated coverage was lost.

I also sabotage-checked the new specs rather than trusting the green: baseline 14/14 pass; with three simultaneous sabotages (drop the `.trim()`, drop the exact-match preference, drop the `!isDone && !parentId` filter) the suite goes to **6 FAILED / 8 SUCCESS**. These tests can genuinely fail — nice.

Below is what's left. Nothing here is a data-loss blocker, and none of it is large, but items 1 and 2 are worth another round: 1 is the PR's own headline guarantee, and 2 is a fresh regression.

---

### 1. The cold-launch guarantee is still only half-delivered

`processProtocolUrl` defers only when the window is missing:

```ts
// electron/protocol-handler.ts:43
if (!mainWin || !mainWin.webContents) {
  log('Window not ready, deferring protocol URL processing');
  pendingUrls.push(url);
```

But a freshly-created `BrowserWindow` **already has `webContents`** long before Angular boots. A URL arriving in that gap takes the non-deferred path straight to `webContents.send(...)`, and the renderer hasn't run `window.ea.on(...)` yet — so there is no listener, and `ipcEventReplay$`'s `ReplaySubject` doesn't exist yet to buffer it. The message is simply dropped: no task, no snack.

The new APP_READY gate in `start-app.ts` only drains URLs that were *already queued*, so it doesn't cover this path:

```
$ grep -n "getIsAppReady\|isAppReady\|APP_READY" electron/protocol-handler.ts
(no matches)
```

This is reachable on exactly the primary use case — a macOS `open-url` that cold-launches the app, or a Windows/Linux `second-instance` arriving while the first instance is still booting.

The repo already solves this exact problem on its other external-automation surface:

```ts
// electron/local-rest-api.ts:194
if (!getIsAppReady()) {
  writeJson(res, 503, { ok: false, error: { code: 'APP_NOT_READY', ... } });
  return;
}
```

Suggested fix — extend the defer condition (`getIsAppReady` is already exported from `main-window.ts` and imported by `start-app.ts` in this PR):

```ts
if (!mainWin || !mainWin.webContents || !getIsAppReady()) {
```

*Verification level: code-path analysis, not reproduced against a running Electron build.*

---

### 2. The new `.toLowerCase()` re-arms #7114 for mixed-case URLs

This PR added case-normalization in one place but not the other:

```ts
// protocol-handler.ts:59  (new in this PR)
const action = urlObj.hostname.toLowerCase();

// protocol-handler.ts:30  (unchanged)
export const getProtocolAction = (url) => { ... return new URL(url).hostname; };
```

So for `superproductivity://Toggle-Visibility`:

- **Before this PR:** `processProtocolUrl` saw `'Toggle-Visibility'`, fell through to `default:` and did nothing. Harmless.
- **After this PR:** it now *reaches* `toggleWindowVisibility(...)` — but both #7114 guards compare against the lowercase literal and therefore miss, because they go through `getProtocolAction`:
  - the `second-instance` pre-focus exemption (`:201`) → `showOrFocus` runs first
  - the cold-start `coldStartShowPending` flag (`:227`) → never set

Net result: show, then toggle ⇒ **the window hides** — precisely the bug #7114 fixed.

One-line fix:

```ts
return new URL(url).hostname.toLowerCase();
```

---

### 3. The docs' Short Syntax section is wrong in two ways

The example in `3.01-API.md`:

> `title=Buy milk +Work #urgent` creates/assigns tags and a project from the title text itself, in addition to any explicit `projectId`.

I ran it:

```
$ node -e "const u=new URL('superproductivity://create-task?title=Buy milk +Work #urgent');
           console.log(JSON.stringify(u.searchParams.get('title')), JSON.stringify(u.hash))"
"Buy milk  Work " "#urgent"
```

`URLSearchParams` decodes a raw `+` as a **space**, and `#urgent` becomes the URL **fragment**. Neither Short Syntax token reaches the title — the documented example cannot work as written. It needs `%2BWork` and `%23urgent`. (Your own spec already encodes this knowledge: *"returns null when the title query param is a lone `+` (decodes to a space)"*.)

And when it *is* correctly encoded, "in addition to" isn't accurate either — `+Project` **overrides** `projectId`. `parseProjectChanges` sets `projectId` from the title with no regard for the task's existing one (`short-syntax.ts:259-266`), and the effect then moves the task:

```ts
// short-syntax.effects.ts
if (r.projectId && r.projectId !== task.projectId && !task.parentId) { ... targetProjectId = r.projectId; }
```

That's worth a decision, not just a doc fix: `_handleAdd` deliberately **errors out** on an *unknown* `projectId` rather than degrading silently, and then lets the title silently override a *known-good* one. Also worth noting `#tag` reaches `addNewTagsFromShortSyntax`, i.e. a URL can create tags.

The repo's precedent for external content is `eml-drop.service.ts:56`:

```ts
// isIgnoreShortSyntax: the subject is untrusted external content — don't
// let ShortSyntaxEffects parse #tag/@date/+project tokens out of it.
this._taskService.add(title, false, { notes }, false, true);
```

Either pass `isIgnoreShortSyntax: true` (at minimum when an explicit `projectId` was given), or keep the behavior and fix the docs to say "overrides" plus correct the encoding in the example. Related cosmetic nit: the SUCCESS snack shows the raw title, so with short syntax enabled it won't match the title actually stored (`_handleComplete` gets this right by using `match.title`).

---

### 4. iOS shows no feedback for an empty title, contradicting the documented contract

`parseAppUriTaskAction` returns `null` for a missing/whitespace-only title (`:41`), so `main.ts` forwards nothing and no snack fires. Desktop, by contrast, forwards `' '` (truthy) and shows `EMPTY_TITLE_VIA_APP_URI`. But the docs promise:

> Both platforms show a snack notification on success (task added/completed) or failure …

So the new `EMPTY_TITLE_VIA_APP_URI` string is unreachable on iOS. Either surface it there too, or scope the doc claim to desktop.

---

### Minor

- **Single-slot buffers lose and reorder.** Both `ReplaySubject(1)`s (Capacitor `pending-capacitor-app-uri-action.ts:14`, and the per-channel Electron ones) keep only the last value, and `merge()` subscribes `add` **before** `complete`. A cold-launch Shortcut that fires `complete X` then `create X` replays as create-then-complete, instantly completing the task it just made. A single shared FIFO would fix both the loss and the ordering.
- **Log leak (pre-existing line, but this PR is what puts user content in these URLs).** `log('Error processing protocol URL:', error)` (`:148`) — I confirmed Node's `ERR_INVALID_URL` carries an **enumerable** `input` property holding the full URL, and `util.inspect`/`console.log` print it:
  ```
  code: 'ERR_INVALID_URL',
  input: 'superproductivity://ho st/x?title=SECRET&notes=PRIVATE'
  ```
  Narrow trigger (needs a malformed *authority*, not a bad query), but the log is exportable — logging `error.code` only would close it.
- **Test name overclaims.** `'shows an error and never queries the store for a whitespace-only title'` (`:239`) never spies on or asserts `store.select`; that half would pass even if the store were queried.
- **`complete-task` isn't idempotent** — running the same Shortcut twice reports "No matching task found", which reads as a failure rather than "already done".
- **No length cap** on `title`/`notes` (cf. `MAX_EML_BODY_LENGTH = 100_000` in the EML path).

---

### Explicitly cleared — please don't spend time on these

- **Notes are not an XSS vector.** They render through the sanitized path (`main.ts` provides `SANITIZE = SecurityContext.HTML`, and `markdown-sanitization.spec.ts` guards it). GHSA-4rrp is fixed.
- **Translations are correct** — only `en.json` touched, keys sorted, params match `t.const.ts`.
- **No sync-model violations** — one action per user intent, and since this is a plain service rather than an effect, the `LOCAL_ACTIONS` / hydration-guard rules don't apply.
- A throw inside the `subscribe(next)` handler does **not** kill the stream (RxJS 7 catches and reports it), and `concatMap` does not deadlock.

Still worth flagging from earlier rounds: **sibling PR #8515 is still open** and adds a `quick-add` case to this same `switch` plus edits the same `protocol-handler.test.cjs` — whichever lands second will conflict.

MannXo:
Round 4 pushed, thanks for the thorough review again.

Both blockers: cold-launch now defers on `getIsAppReady()` (not just window existence), so a URL arriving before the renderer registers its listeners is queued and drained on APP_READY; and `getProtocolAction` lowercases the hostname, so the second-instance exemption and cold-start flag match a mixed-case action.

Items 3 and 4: URL tasks now use `isIgnoreShortSyntax: true`, treating the URL as untrusted external content like the eml-drop path, so `projectId` stays authoritative, a URL can no longer create tags/projects, and the snack matches the stored title (docs updated to match). The iOS parser now forwards a whitespace-only title so the empty-title error snack fires there too.

Minors folded in: the catch logs the error code only, the whitespace-title test now asserts the store is never queried, and title/notes over 100k are rejected. Left the FIFO buffer and complete-task idempotency out of scope for now, happy to follow up if you want them.

johannesjo:
## Round 4 review — `919f048688`

All five round-3 findings are genuinely fixed, and I sabotage-verified each one rather than reading the diff. CI is green (21 success, 1 skipped).

I broke each fix and confirmed the tests catch it — tree restored, `git status` clean afterwards:

| Sabotage | Result |
| --- | --- |
| drop `isIgnoreShortSyntax` + both length caps | **6 FAILED** / 10 passed (service spec) |
| restore the parser's `.trim()` | **4 FAILED** / 8 passed |
| revert `getIsAppReady()` + `.toLowerCase()` | **2 FAILED** (electron tests 10 and 13) |
| remove the empty-needle guard | **1 FAILED** |

**No new test is vacuous.** I also checked the things that *look* right but could silently not be:

- **The APP_READY ordering actually holds.** `main-window.ts:384` registers the `isAppReady = true` listener inside `createWindow()`, which `start-app.ts:509` awaits *before* registering its own `ipcMain.once(IPC.APP_READY, …)` drain at `:528`. Node fires listeners in registration order, so the flag is set before the drain runs. If it were the other way round every queued URL would be silently re-queued — worth stating explicitly since nothing enforces it.
- **No import cycle** from the new `getIsAppReady` import — nothing in `main-window.ts`'s transitive graph reaches `protocol-handler.ts`.
- **`add()`'s positional args are right**: `task.service.ts:391-400` confirms param 5 is `isIgnoreShortSyntax`, and the explicit param-4 `false` matches the old default, so the dispatched action is unchanged apart from the new flag.

On the FIFO buffer you left out of scope — **agreed, drop it.** Your readiness gate closed it for free on Electron: the main process no longer sends before `APP_READY`, and `AppUriTaskActionsService` is injected in `app.component.ts:162` (field initialiser) so it has already subscribed by the time `StartupService.init()` (`:229`) calls `informAboutAppReady()`. The reorder is unreachable there now. Only the iOS single-slot *loss* remains, which needs two `appUrlOpen` events inside the bootstrap window — not worth a FIFO.

Two things for a round 5. Neither is a hard blocker.

---

### 1. 🟠 `3.01-API.md:439` states something false — and it's my fault

> …for backward compatibility with a pre-existing but **previously-unwired** desktop action

That action was wired. I verified the pre-PR chain link by link at `6756e9998c`:

| Link | Evidence |
| --- | --- |
| protocol sends IPC | `electron/protocol-handler.ts:79` → `send(IPC.ADD_TASK_FROM_APP_URI, { title })` |
| channel exists | `electron/shared-with-frontend/ipc-events.const.ts:25` |
| preload bridges it | `electron/preload.ts:35` (generic `on(channel, …)`) |
| renderer observable | `src/app/core/ipc-events.ts:53` |
| effect consumes it | `task-electron.effects.ts:215` → `this._taskService.add(data.title)` |
| effect registered | `feature-stores.module.ts:162` → `...(IS_ELECTRON ? [TaskElectronEffects] : [])` |

A one-argument `add()` leaves `isIgnoreShortSyntax` at its `false` default, so **Short Syntax did parse on that path.** And it shipped: the protocol case (`cf1031c935`) is in 163 tags going back to **v14.2.4**, and the doc row `3.03-Keyboard-Shortcuts.md:129` (added in `14a56f861e`) is in v18.13.0 through v18.15.1.

So this PR turns Short Syntax off for an action that has shipped for a long time. **That belief came from me** — in round 1 I asserted `ipcAddTaskFromAppUri$` was dead code, which was wrong, and it has now made it into permanent user-facing documentation. Sorry for the detour.

To be fair to the claim, there is a true version of it: the pre-PR **cold-launch** path really was broken (`ipcEvent$` was a plain `Subject` with no replay, no readiness gate, and a fixed 1000 ms drain), so a cold-launched `create-task` plausibly did nothing. The **warm** path — app already running, via `second-instance` or macOS `open-url` — was fully functional. "Unreliable on cold launch" would have been accurate; "unwired" is not. And this isn't hidden — your new 3.01 section documents the verbatim-title behavior clearly.

What I'd like:

1. **Fix the sentence.** Something like "…for backward compatibility with the pre-existing desktop action (shipped since v14.2.4), whose title handling this changes — see below."
2. **Update `3.03-Keyboard-Shortcuts.md:129`**, which the PR leaves untouched. It still reads "Create a task with the URL-encoded `<title>`" with no mention of the query parameters or the verbatim-title change, so the repo now ships two wiki pages describing the same URL with different semantics. It's also the page a Wayland user actually finds.
3. **Call the behavior change out in the PR description**, so it can be a deliberate decision rather than a side effect.

I still think `isIgnoreShortSyntax: true` is the right call for the new query form — a URL shouldn't be able to create tags/projects or override a validated `projectId`. But it is a breaking change to shipped behavior, and that's the maintainer's call to make knowingly. The leaner alternative, for the record, is to keep Short Syntax and just document that `+Project` overrides `projectId` — a two-line doc change with no behavior change at all.

---

### 2. 🟡 The 10 s retry timer amplifies ×N per wave

`protocol-handler.ts:60` schedules a `setTimeout(…, 10000)` for **every** deferred URL, and `processPendingProtocolUrls` re-queues each URL that is still not ready. So each timer in a wave finds `pendingUrls` refilled by the previous timer in the same wave and schedules N more.

I probed the real module with a virtual clock and `getIsAppReady() === false`:

```
N=1  timers per 10s wave: 1, 1, 1, 1, 1, 1        (flat — fine)
N=2                       2, 4, 8, 16, 32, 64
N=3                       3, 9, 27, 81, 243, 729
```

Two important qualifiers, so this isn't overstated:

- **No URL is ever delivered twice.** `pendingUrls = []` precedes the `forEach`, so only the first timer of a wave does work. I verified this: 8 outstanding timers, `getIsAppReady()` flipped true, → exactly 2 sends for 2 URLs.
- **This is pre-existing, not something you introduced.** `processPendingProtocolUrls` is byte-identical in your diff; the only change is the guard at `:55`. But it *was* nearly unreachable — at base, a present `mainWin` meant zero deferral — and the readiness gate now routes every cold launch through that branch. You widened the door to it.

It only actually explodes when `APP_READY` never arrives, i.e. an already-broken renderer (the repo has a `gpu-startup-guard` precisely because that happens), so it's a memory-growth footgun rather than a correctness bug. Since `APP_READY` now owns the drain, the simplest fix is to **delete the `setTimeout` entirely** — if the renderer never signals ready it can't receive the message anyway. A single re-arm guard would also do.

---

### Minor

- **`app-uri-task-actions.service.ts:25` cites the wrong EML constant.** `eml-drop.service.ts:15-16` caps **titles at 300** and bodies at 100 000, and it **truncates** rather than rejecting. This PR applies 100 000 to both title and notes and rejects the whole action. Its own comment there ("The title and body sync to every device as ops") applies here too — a 100 000-character title syncs as an op to every device. Either match the 300 for the title or drop the "mirroring" claim.
- **`en.json:1977` reads "…no task was **added**"** but `_handleComplete` (`:153`) reuses the same key, so an over-long `complete-task` title reports that nothing was added.
- **The docs say "verbatim"** (`3.01-API.md:447`) but `_handleAdd` (`:97`) trims surrounding whitespace. "Short Syntax is not parsed" is the accurate claim.
- **A literal `title=` still produces no feedback on either platform.** `searchParams.get('title')` returns `""`, which is falsy, so the parser returns `null` and Electron's `if (taskTitle)` skips it. `%20` and `+` now correctly reach the error snack, so this is just the last empty-title case.
- **Two test gaps.** The new deferral test (`protocol-handler.test.cjs:187`) enables mocked timers but never advances them and never asserts the URL is *eventually delivered* — the actual promise of the fix. And the catch-path privacy fix has no test at all: `:206` uses a valid URL and never enters the `catch`.

---

### Checked and correct — please don't re-chase

No import cycle; no `getProtocolAction` caller broken by lowercasing (all four compare lowercase literals); notes are sanitized, so no XSS; translations correct (`en.json` only, params match); a whitespace `complete-task` cannot `.includes('')`-match every task; no new dependencies; no sync-rule violation. Also: the `task-electron.effects.spec.ts` you deleted was itself vacuous — its factory overwrote `handleAddTaskFromProtocol$` with a hand-rolled reimplementation and asserted against that, never the real effect — so no coverage was lost there.

One logistical note: #8515 adds a `quick-add` case to the same `switch` and edits the same `protocol-handler.test.cjs`, so whichever lands second will need a trivial conflict fixup.

Nice work on this round — the readiness gate in particular is the right fix in the right place, and it retired a separate finding as a bonus.

MannXo:
Round 5 pushed, thanks again.

1. Fixed the false "previously-unwired" line (the desktop action shipped since v14.2.4), updated 3.03-Keyboard-Shortcuts.md so both wiki pages agree, and called the Short Syntax behavior change out explicitly in the PR description as a deliberate decision. Kept `isIgnoreShortSyntax: true`, with the non-breaking alternative noted there in case you prefer it.

2. Deleted the retry `setTimeout`. APP_READY owns the drain now, so the xN amplification is gone; the deferral test now asserts the queued URL is actually delivered on drain.

Minors: title cap is now 300 / notes 100k (matching the EML limits), comment corrected (we reject rather than truncate, since a URL is a single deliberate call); the too-long snack no longer says "added" so it fits complete-task; docs say "Short Syntax is not parsed"; a present-but-empty `title=` now reaches the error snack on both platforms; and added the catch-path privacy test (malformed authority) plus the delivery assertion. Left the FIFO buffer and complete-task idempotency out per your note.

johannesjo:
## Round 5 re-review — `21aee9d17d`

I rechecked the full PR, the exact `919f048688..21aee9d17d` delta, the discussion/thread state, and CI. The head is stable and all current checks are green. I also reran the exact-head protocol (18/18), parser (12/12), and service (17/17) specs. The timer removal/readiness drain, empty-title forwarding, and catch-path privacy logging all look correct.

I then added two throwaway counterexample specs in an isolated snapshot (nothing committed); both fail on the current head.

### 1. 🔴 `complete-task` reuses a creation-only 300-character cap

[`_handleComplete()` rejects the lookup before querying the store](https://github.com/super-productivity/super-productivity/blob/21aee9d17df05204829b893cd0d1f58bcd691644/src/app/features/tasks/app-uri-actions/app-uri-task-actions.service.ts#L155-L164), although this title is only a search needle and is never persisted or synced. Task titles have no model-level 300-character invariant, and a normal in-app EML import can itself produce 301 characters because it [slices to 300 and appends an ellipsis](https://github.com/super-productivity/super-productivity/blob/21aee9d17df05204829b893cd0d1f58bcd691644/src/app/core/drop-paste-input/eml-drop.service.ts#L44-L47).

Concrete unreachable case: two active tasks titled `a × 301` and `a × 302`. The full 301-character exact query is rejected, while every permitted shorter `a` query matches both and is ambiguous. Without this guard, the existing exact-match preference selects the first task correctly.

The current [301-character rejection spec](https://github.com/super-productivity/super-productivity/blob/21aee9d17df05204829b893cd0d1f58bcd691644/src/app/features/tasks/app-uri-actions/app-uri-task-actions.service.spec.ts#L294-L307) therefore pins the regression. Smallest fix: keep 300 for newly persisted titles, give completion its own generous lookup guard (for example the previous 100,000 limit) applied after trimming, and replace that spec with the two-long-title exact-match regression.

### 2. 🟠 Length validation runs before documented normalization

[`_handleAdd()` checks the raw title length before trimming](https://github.com/super-productivity/super-productivity/blob/21aee9d17df05204829b893cd0d1f58bcd691644/src/app/features/tasks/app-uri-actions/app-uri-task-actions.service.ts#L88-L100). Thus a 300-character title with one surrounding space is rejected, although its stored value is exactly 300 characters and the API says surrounding whitespace is trimmed. My second counterexample spec reproduces this directly.

Smallest fix: trim first, reject empty, then validate the normalized title; leave the notes cap on the raw notes value. Apply the same ordering to completion if it retains a lookup cap, and add the padded-300 boundary test.

### 3. 🟡 Align the public contract and PR metadata

- The [API parameter tables](https://github.com/super-productivity/super-productivity/blob/21aee9d17df05204829b893cd0d1f58bcd691644/docs/wiki/3.01-API.md#L441-L457) do not state the title/notes limits or whether limits apply after trimming.
- `3.01-API.md:503` still says mobile/Capacitor generally, despite Android being unsupported.
- `3.03-Keyboard-Shortcuts.md:129` says “stored literally” even though whitespace is trimmed, and uses a relative Markdown link instead of the wiki's `[[3.01-API]]` form.
- The PR title/body still say `add-task`, Android, previously dead/unsubscribed desktop handling, and silently dropped unknown projects. The final code uses `create-task`, supports iOS + desktop, replaces shipped warm-path behavior, and rejects unknown projects.

On the Short Syntax decision: keeping `isIgnoreShortSyntax: true` is defensible for this external-input boundary, but it changes the shipped desktop API. Please tick **Breaking change** and keep the compatibility note prominent (or take the non-breaking alternative already described in the body).

**Verdict: needs changes** for the two reproduced title-limit regressions and the public-contract cleanup. I did not find another runtime blocker in this round.

johannesjo (commented):
## Re-review at `5650bfab`

I rechecked the full current diff, prior discussion, linked issue, exact pinned dependency behavior, and current CI. All checks are green and there are no unresolved inline threads. The earlier 300-character completion limit and pre-trim validation issues are fixed; the Short Syntax compatibility break is now explicit in the PR/docs, so I am not carrying those forward.

### Important — a cold-start iOS URL can be consumed by the wrong listener

The PR registers a task [`appUrlOpen` listener in `main.ts`](https://github.com/super-productivity/super-productivity/blob/5650bfab37c9c6b0bbccd911db9090bc902ba8ad/src/main.ts#L545-L562), while [`OAuthCallbackHandlerService` registers another](https://github.com/super-productivity/super-productivity/blob/5650bfab37c9c6b0bbccd911db9090bc902ba8ad/src/app/imex/sync/oauth-callback-handler.service.ts#L47-L76). In the exact locked versions, `@capacitor/app` 8.1.0 emits pre-listener URL events with [`retainUntilConsumed: true`](https://github.com/ionic-team/capacitor-plugins/blob/3678f5063762e8d9857ddb4f41aae5f65ead91fa/app/ios/Sources/AppPlugin/AppPlugin.swift#L50-L64), and Capacitor 8.4.1 [drains and removes the retained arguments when the first listener is added](https://github.com/ionic-team/capacitor/blob/7217b5215a175eededf607a216e0cab2a8450a34/ios/Capacitor/Capacitor/CAPPlugin.m#L42-L65). The second listener therefore never receives that cold-start event. Because each callback ignores the other route family, either a cold task URL or a cold OAuth callback is silently lost depending on registration order; warm events work because both listeners already exist. There is no `getLaunchUrl()` recovery path.

Please use one native listener/router that dispatches to both consumers (the existing task `ReplaySubject` can remain the task-side queue), and add a retained-event regression test covering both route families.

### Important — URL-controlled titles can trigger external requests from the snack UI

The new service passes titles verbatim in snack `translateParams` in the [create paths](https://github.com/super-productivity/super-productivity/blob/5650bfab37c9c6b0bbccd911db9090bc902ba8ad/src/app/features/tasks/app-uri-actions/app-uri-task-actions.service.ts#L132-L162) and [completion paths](https://github.com/super-productivity/super-productivity/blob/5650bfab37c9c6b0bbccd911db9090bc902ba8ad/src/app/features/tasks/app-uri-actions/app-uri-task-actions.service.ts#L165-L225). [`SnackService` interpolates them](https://github.com/super-productivity/super-productivity/blob/5650bfab37c9c6b0bbccd911db9090bc902ba8ad/src/app/core/snack/snack.service.ts#L99-L110), then the snack template [renders the result through `[innerHtml]`](https://github.com/super-productivity/super-productivity/blob/5650bfab37c9c6b0bbccd911db9090bc902ba8ad/src/app/core/snack/snack-custom/snack-custom.component.html#L29-L32).

With the exact locked ngx-translate 17.0.0 / Angular 21.2.17 runtime, a title such as `` remains an external `` after interpolation and Angular sanitization. A headless Chromium check under the app's `img-src *` CSP issued the GET. This is not arbitrary-script XSS, but it is HTML/resource injection: an external URL can create a privacy beacon and inject trusted-looking snack markup. A no-match `complete-task` URL is sufficient to trigger it.

Please apply the existing [`escapeHtml()` helper](https://github.com/super-productivity/super-productivity/blob/5650bfab37c9c6b0bbccd911db9090bc902ba8ad/src/app/util/escape-html.ts#L1-L16) only to titles used as snack translation parameters (not stored/search titles), cover every new branch, and add an escaping regression test.

### Minor

The Wayland [“Available fire-and-forget actions” table](https://github.com/super-productivity/super-productivity/blob/5650bfab37c9c6b0bbccd911db9090bc902ba8ad/docs/wiki/3.03-Keyboard-Shortcuts.md#L121-L129) lists `create-task` but omits the new desktop `complete-task` action.

### Verdict

Needs changes for the two Important runtime findings above. I found no other live blocker in the current head.

MannXo:
Round 6 addressed in `d86e6d8bb`. Full suite 13196 green, lint clean.

**🟠 Cold-start URL consumed by the wrong listener.** Confirmed, and it was ours to break. I checked the pinned native source rather than take it on trust: `CAPPlugin.m` only calls `sendRetainedArgumentsForEvent` when the listener list for that event is empty, and that method does `removeObjectForKey` before dispatching, so the first listener drains and destroys the retained argument. Since `main.ts` registers at module evaluation and `OAuthCallbackHandlerService` waits on Angular DI, ours won, which means this PR silently broke cold-start OAuth callbacks. The comment I had written there ("Capacitor supports multiple listeners on the same event") is true for warm events and exactly the wrong assumption for retained ones.

`main.ts` now owns the only `appUrlOpen` listener and routes through a new `routeCapacitorAppUrl`: recognized task actions to the existing task `ReplaySubject`, everything else to a new `pendingCapacitorOAuthUrl$` that the OAuth handler consumes instead of registering its own listener. Route matching for callbacks stays owned by the OAuth service, so there is still one owner per family.

I pulled the router out of `main.ts` into its own module specifically so it could be tested, since nothing inside the bootstrap block is reachable from Karma. `app-url-open-router.spec.ts` covers both families with the URL emitted before anything subscribes, which is the cold-launch shape, plus no cross-family leakage in either direction. On the OAuth side, `oauth-callback-handler.service.spec.ts` emits a callback before the service subscribes and asserts it still arrives. Sabotage check: dropping the OAuth branch fails 5 of the 8 router tests.

One thing worth naming: the new OAuth sink is `ReplaySubject(1)`, matching the task side, so two callbacks arriving before Angular boots would keep only the last. That is the same single-slot behaviour we agreed to leave alone, but it now applies to a second stream, so I would rather say it than have you find it.

**🟠 URL-controlled titles in snack params.** Confirmed: `translateParams` goes through `translate.instant()` into `data.msg`, which the snack template renders with `[innerHtml]`, and Angular's sanitizer keeps an external ``. Applied `escapeHtml()` to all six sites that put a title into `translateParams`, including the completion success snack, since a matched task's stored title can carry markup from sync or an EML import and not just from the URL. Stored and lookup titles are untouched, and the tests assert `TaskService.add` still gets the verbatim title. Six new tests, one per branch plus an ampersand-ordering case; sabotage check: removing the escaping fails all 6 and leaves the 18 pre-existing tests passing.

**🟡 Wayland table.** Added `complete-task` next to `create-task`, with the trimming note and the match rule.

Not verified on device: no Xcode here, and `IS_NATIVE_PLATFORM` is false under Karma, so Capacitor's own drain behaviour stays structurally untestable in this repo. The tests cover the property that follows from it, that a URL arriving before a consumer exists still reaches it.

johannesjo:
## Round 7 review — `d86e6d8bb7`

Thanks for the careful writeup on the retained-event mechanism — your reading of `CAPPlugin.m` is right, and checking the pinned native source rather than taking my word for it is exactly the right instinct.

CI is green (22 SUCCESS + 1 skipped). I sabotage-verified both new suites rather than trusting that, and your own numbers reproduce exactly:

| Sabotage | Result |
| --- | --- |
| drop `oAuthSink.next(url)` in the router | **5 FAILED** / 3 passed |
| strip all 6 `escapeHtml()` call sites | **6 FAILED** / 18 passed |

`escapeHtml` is the right helper (it replaces `&` first), all six `translateParams` sites are covered, and the stored/lookup titles are correctly left verbatim — the spec pins that too, and I confirmed it isn't vacuous by wrapping the stored title in `escapeHtml` (1 FAILED). Baselines at head: router 8/8, OAuth 8/8, service 24/24.

Two things need fixing before this can go in, and I found one of them by pulling on my own round-6 finding.

---

### 🔴 Blocker — the consolidation killed plugin OAuth on Android

Collapsing to one listener was the right call, but the surviving listener sits in an iOS-only block:

- [`src/main.ts:556`](https://github.com/super-productivity/super-productivity/blob/d86e6d8bb7ab97dd445a62b29888d36117e6207c/src/main.ts#L556) — the app's **only** `appUrlOpen` listener, inside `if (IS_IOS_NATIVE) {` (`:522`–`:564`)
- [`oauth-callback-handler.service.ts:33`](https://github.com/super-productivity/super-productivity/blob/d86e6d8bb7ab97dd445a62b29888d36117e6207c/src/app/imex/sync/oauth-callback-handler.service.ts#L33) — the constructor still gates on `IS_NATIVE_PLATFORM` (iOS **or Android**), and this commit deleted its own `App.addListener`, so it now only *subscribes* to `pendingCapacitorOAuthUrl$`

On Android nothing ever feeds that subject, so the subscription is permanently dead. On `5650bfab37` and on `master` the service registered its own listener under `IS_NATIVE_PLATFORM` and Android worked, so this is a regression introduced here — and master auto-publishes to the Play internal track, so it reaches real testers' devices.

Verified rather than assumed: `AndroidManifest.xml` declares three `VIEW` filters on `CapacitorMainActivity` (`:99-106` "OAuth callback handler for Dropbox authentication", `:109-116`, `:119-126`); `@capacitor/app`'s `AppPlugin.java:146-158` fires `notifyListeners(EVENT_URL_OPEN, ret, true)`; there is **zero** OAuth handling anywhere in `android/app/src/main/java/`; and `PluginOAuthRedirectHandler` explicitly skips native (`else if (!IS_NATIVE_PLATFORM)`), so there is no fallback path.

**Scope, to be precise:** the concrete breakage is **plugin OAuth**, not Dropbox or OneDrive. `dropbox.ts:369` `_getRedirectUri()` returns `null` on every platform (manual code-entry flow), and `onedrive.ts:631-635` returns the `nativeclient` https redirect off-Electron — neither issues a custom-scheme redirect on mobile. `PluginOAuthService` does: it hands back `com.superproductivity.superproductivity:/plugin-oauth-callback` on Android and opens the system browser, so the callback returns as an intent with no listener and `waitForRedirectCode` hangs until its 5-minute timeout.

**Fix:** widen the gate on the `appUrlOpen` registration from `IS_IOS_NATIVE` to `IS_NATIVE_PLATFORM` — exactly master's condition for this family. It's safe to do wholesale: Android declares no `create-task`/`complete-task` host, so no task URL can reach it and the "iOS + desktop" scoping stays true in practice.

Worth naming: green CI can't help here. The platform gate is structurally untestable in this repo — the router spec exercises a pure function with injected sinks, the OAuth spec calls `service['_setupAppUrlListener']()` directly because `IS_NATIVE_PLATFORM` is false under Karma, and `main.ts`'s bootstrap block is unreachable. Nothing calls `routeCapacitorAppUrl(url)` with its *default* sinks, so nothing asserts the producer and the consumers share a stream.

---

### 🟠 Important — the cold-start URL is still dropped, one layer further down

This is the part I'd most like you to check, because it means the commit doesn't yet achieve what it set out to do — on iOS either.

`_setupAppUrlListener()` runs from the **constructor** (`:34`), which executes inside the `APP_INITIALIZER` at `main.ts:302-309`. Subscribing to the `ReplaySubject(1)` synchronously drains the retained URL and calls `this._authCodeReceived$.next(callbackData)` at `:78` — but `_authCodeReceived$` is a plain `Subject` (`:26`) with **no subscribers yet** at that moment. The replayed cold-start callback is emitted into the void.

I reproduced it against the real service, in production order:

```
it('cold-start URL survives when setup runs BEFORE consumers subscribe', ...)
  pendingUrl$.next('com.super-productivity.app://oauth-callback?code=COLD123');
  service['_setupAppUrlListener']();          // production: constructor first
  service.authCodeReceived$.subscribe(...);   // consumer attaches later

  Expected 0 to be 1.
TOTAL: 1 FAILED
```

All three new tests in `oauth-callback-handler.service.spec.ts:26-45` subscribe to `authCodeReceived$` *before* calling `_setupAppUrlListener()`, which is the inverse of production — and that inversion is precisely what hides this. Calling the private method directly is defensible given the Karma constraint, but the ordering has to match production or the test isn't evidence.

Making `_authCodeReceived$` a `ReplaySubject(1)` would close it. Before you do, though, it's worth asking whether this path earns its place at all: the plugin branch drops a cold-start code anyway (`plugin-oauth.service.ts:329-340` warns "Received OAuth code but no pending flow" whenever `_pendingRedirect` is null, which it always is after a relaunch). If cold-start OAuth can't work end-to-end without a durable pending-flow record, the honest move may be to keep the router (it's still needed to unbreak Android) and drop the cold-start OAuth ambition rather than carry a token, a module singleton and a rewired service for it.

---

### 🟠 Important — `3.03` documents a `complete-task` form that silently does nothing

The new row is:

```
| `superproductivity://complete-task/<title>` | — | Complete the single unfinished top-level task matching the URL-encoded `<title>` (also accepts `?title=`). ...
```

But `protocol-handler.ts:116-117` reads **only** `urlObj.searchParams.get('title')` for `complete-task` — it never touches `pathParts`, unlike `create-task` (`:88-91`), which does. Your own test asserts this (`protocol-handler.test.cjs:194`, "complete-task without a title query param sends nothing"), and the mobile parser is the same (`parse-app-uri-task-action.ts:47`).

So a Wayland user who copies that row and runs `xdg-open "superproductivity://complete-task/Buy%20milk"` gets `null` → nothing sent, no snack, no error. The row also inverts reality by presenting `?title=` as the fallback when it's the only supported form — and `3.01-API.md:456` correctly lists `title` as a **required query parameter**, so the two pages now contradict each other. Make it `superproductivity://complete-task?title=<title>`.

---

### 🟡 One of the six new escaping tests can't fail

`app-uri-task-actions.service.spec.ts:403`:

```ts
it('escapes the ampersand first so escaping cannot be double-applied', () => {
  pendingAction$.next({ type: 'add', title: '&lt;img&gt;' });
  expect(titleParamOf()).toBe('&amp;lt;img&amp;gt;');
});
```

The input contains no literal `<` or `>`, so the replacement order is irrelevant to it. I introduced the exact bug it names — moved `.replace(/&/g, '&amp;')` to **last** in `escape-html.ts` — and re-ran: **5 FAILED, 19 SUCCESS**. The five failures are its siblings; this one passed. The ordering *is* genuinely covered by those five (and by the pre-existing `escape-html.spec.ts`), so either delete this test or give it a literal `<` (e.g. `'<b>&'` → `'&lt;b&gt;&amp;'`).

Minor related note: the `escapeHtml` at `:174` is unreachable-by-construction — that branch only runs when `action.title.trim()` is empty, so the value is whitespace-only and escaping is a provable no-op. Removing it alone leaves 24/24 green. Harmless either way.

---

### Minor

- `main.ts:299-301` still says the service's constructor "registers Capacitor's `appUrlOpen` listener" — it doesn't any more, it subscribes. There's a second one at `plugin-oauth-redirect.handler.ts:24-25` ("handled by OAuthCallbackHandlerService via Capacitor's `appUrlOpen` listener"), which on Android is now actively false and makes the `!IS_NATIVE_PLATFORM` guard above it look deliberate.
- `pending-capacitor-oauth-url.ts:20` — the lifetime `ReplaySubject(1)` is never cleared after consumption, so a callback URL containing `?code=<authorization code>` stays in memory for the rest of the process and replays to any future subscriber. That sits awkwardly against this commit's own security note ("must never be logged, as it carries … an auth code"). A one-shot drain would avoid it. Separately, `retainUntilConsumed` retains into an *array*, so a burst of launch URLs collapses to the last one in a 1-slot buffer — an unbounded `ReplaySubject()` is fine here since each sink has exactly one consumer.
- `main.ts:561` logs `{ recognized: isTaskAction }`, so a perfectly valid OAuth callback logs `recognized: false`. Renaming the key to `isTaskAction` would stop that misleading triage on the one signal this feature emits.
- Pre-existing, but this commit widens what reaches it: `oauth-callback-handler.service.ts:181` does `SyncLog.err('… Failed to parse plugin OAuth URL', url, e)` — the raw URL, auth code and all, into an exportable log. `protocol-handler.ts:160-171` already shows the pattern for logging a non-identifying descriptor instead.

### Not a request, just context

The six escapes here are correct and match the three pre-existing hand-escaped sites, but the convention is unenforced: `translateParams` values reach `[innerHtml]` on three surfaces (`snack-custom.component.html:30`, `dialog-confirm.component.html:11`, `banner.component.html:43`), and of ~147 `translateParams` uses only 9 escape — several of the rest carry synced task/project titles, which is the same cross-device threat model your doc comment cites. That's a separate cleanup (best done centrally in `SnackService`/`BannerService`, removing the local calls at the same time so nothing double-escapes), definitely not this PR's job.

---

### Verdict

Needs changes for the Android gate and the doc row; I'd like your read on the constructor-ordering one before you spend effort on it, since it may argue for *less* code rather than more. Everything else in this round is solid work — the router split for testability was the right instinct, and the escaping is correct.

MannXo:
## Round 7 addressed in `b6a6b6d89`

Full suite 13197 green, lint clean. Baselines at head: router 8/8, OAuth 9/9, service 24/24.

### 🔴 Android gate

Fixed. The `appUrlOpen` registration now lives in its own `IS_NATIVE_PLATFORM` block instead of inside the iOS one, which is the condition master used for this family (`git show master:oauth-callback-handler.service.ts` line 33 confirms it, including the Android WebView case that `IS_NATIVE_PLATFORM` folds in). The iOS-only background flush above it stays iOS-only. You were right that green CI can't help here, so I checked the gate against master's rather than against a test.

### 🟠 Cold-start ordering: you asked for my read, and it argues for less code than you proposed

Your diagnosis is right and the fix you suggested would not have helped. There are two independent reasons a cold-start callback cannot complete, and `_authCodeReceived$` being a `ReplaySubject` fixes neither:

1. `PluginOAuthService._pendingRedirect` is in-memory and null after a relaunch, as you said.
2. The only consumer of `authCodeReceived$` is `DialogGetAndEnterAuthCodeComponent`, and its `isNativePlatform` is a hardcoded `false` (line 68, from the earlier revert of the deep-link flow). It subscribes only when `isElectron`. So on mobile that emission has zero subscribers warm or cold, replay or not.

So I took the branch you floated: keep the router, drop the cold-start ambition. `pendingCapacitorOAuthUrl$` is a plain `Subject` now, with the reasoning in its doc comment. That also closes your memory minor at the same time, since there is nothing left to retain, rather than needing a one-shot drain on top.

Added a guard for the decision itself, asserting the real app-wide singleton rather than the injected double, so a future change back to retaining an auth code fails loudly. Sabotage: restoring `ReplaySubject(1)` gives 1 FAILED / 8.

The three service specs now call `_setupAppUrlListener()` before the consumer subscribes, matching the APP_INITIALIZER order, and the router spec's OAuth sink is a plain `Subject` subscribed before routing so it mirrors production too. Sabotage on the router is unchanged at 5 FAILED / 3 passed, so the reorder cost no coverage.

One thing I did not change: the task sink stays `ReplaySubject(1)` despite the retained-array collapse you noted. A burst of launch task URLs would need two taps resolved into one launch, and unbounded replay would hold titles and notes in memory for the process lifetime. Happy to revisit if you would rather have it.

### 🟠 `3.03` complete-task row

Fixed to `superproductivity://complete-task?title=<title>`, and the cell now says the title must be a query parameter with no path form, so it stops contradicting `3.01-API`.

### 🟡 The ampersand test

You are right, it could not fail. It now uses `'<b>&'` expecting `'&lt;b&gt;&amp;'`. With `&` moved last the suite reports 6 FAILED / 18, up from the 5 / 19 you saw, so the sixth failure is that test. Also removed the unreachable `escapeHtml` at the whitespace-only branch and left a line saying why the raw title is fine there.

### Minors

Both stale comments fixed (`main.ts` now says the service subscribes, and the plugin handler now points at the router). The log key is `isTaskAction` and the message is platform-neutral. The plugin parse failure logs the scheme only, dropping both the URL and the error object, matching `protocol-handler.ts` since `ERR_INVALID_URL` carries the raw URL on `input`.

### On the wider `translateParams` surface

Agreed on all of it, including that central escaping in `SnackService` and `BannerService` is the right shape and that removing the local calls has to happen in the same change to avoid double-escaping. Not this PR.

johannesjo (approved):
## Round 8 review — `b6a6b6d89`

All four round-7 items are fixed. I re-derived every number rather than taking the summary, and yours reproduce exactly:

| Check | Result |
| --- | --- |
| Baselines at head | router **8/8**, OAuth **9/9**, service **24/24** |
| Sabotage: restore `ReplaySubject(1)` on the OAuth sink | **1 FAILED / 8** |
| Sabotage: move `.replace(/&/g, '&amp;')` last in `escape-html.ts` | **6 FAILED / 18** (was 5/19) |
| CI on `b6a6b6d89` | 23 checks, 22 SUCCESS + 1 skipped `(r)Sync` |

---

### 🔴 Android gate — fixed

`main.ts:558` now registers the listener under its own `if (IS_NATIVE_PLATFORM)` block, outside the iOS one. That is exactly the condition master used for this family — `git show master:src/app/imex/sync/oauth-callback-handler.service.ts` line 33 is `if (IS_NATIVE_PLATFORM) {`, as you said — and `IS_NATIVE_PLATFORM` folds in the legacy `SUPAndroid` WebView case, where `AppWeb` inherits `addListener` from `WebPlugin` and the registration is a harmless no-op. The iOS-only `BackgroundTask` flush correctly stayed behind `IS_IOS_NATIVE`.

Scoping still holds: `AndroidManifest.xml` declares only the three OAuth `VIEW` filters (plus `SEND`) on `CapacitorMainActivity`, and both plugin forms come back through the router as non-task URLs — `com.superproductivity.superproductivity:/plugin-oauth-callback` has an empty `hostname`, so `parseAppUriTaskAction` returns `null` and the URL lands on the OAuth sink, where `url.includes('plugin-oauth-callback')` picks it up. No task URL can reach Android.

You're right that checking the gate against master's rather than against a test was the only available move here.

---

### 🟠 Cold-start — your read is right, and there's one more datapoint that supports it

Both of your reasons check out against source: `PluginOAuthService.handleRedirectCode` warns "Received OAuth code but no pending flow" whenever `_pendingRedirect` is null (`plugin-oauth.service.ts:338`), and the only consumer of `authCodeReceived$` in the whole app is `dialog-get-and-enter-auth-code.component.ts:82`, guarded by `if (this.isNativePlatform || this.isElectron)` at `:80` with `readonly isNativePlatform = false` at `:68`. So mobile has zero consumers, replay or not.

Worth adding, because it makes the downgrade strictly free: the sink was never what dropped the cold-start URL. `bootstrapApplication(...)` at `main.ts:125` resolves `ApplicationInitStatus` and calls `runInitializers()` **synchronously** inside that call, which resolves the `APP_INITIALIZER` at `main.ts:302-309` and constructs `OAuthCallbackHandlerService` — all of it before the module-level `if (IS_NATIVE_PLATFORM)` at `:558` ever runs. So the subscriber is already attached when Capacitor drains the retained launch URL into the first listener, and a plain `Subject` delivers it just as a `ReplaySubject` would. It then dies exactly where I found it in round 7, at `_authCodeReceived$` with nobody listening. Dropping the replay costs nothing at all and stops retaining an auth code for the process lifetime.

The new guard is a real one — restoring `ReplaySubject(1)` gives 1 FAILED — and asserting the app-wide singleton rather than the double is the right instinct, since the double is what would hide it.

Keeping the task sink at `ReplaySubject(1)` is fine by me; agreed that unbounded is the worse trade.

---

### 🟠 `3.03` complete-task row — fixed

`superproductivity://complete-task?title=<title>` now matches `protocol-handler.ts:117` (`searchParams.get('title')` only) and no longer contradicts `3.01-API.md:457`. And `3.01` was already precise about `create-task/<title>` being desktop-only, so the two pages agree now.

### 🟡 Ampersand test — fixed, and it bites

With `&` moved last the suite reports 6 FAILED and the sixth is the one in question:

```
✗ escapes the ampersand first so escaping cannot be double-applied
  Expected '&amp;lt;b&amp;gt;&amp;' to be '&lt;b&gt;&amp;'.
```

Removing the whitespace-only-branch escape is safe: `needle = action.title.trim().toLowerCase()` at `app-uri-task-actions.service.ts:167`, so reaching `:175` means the title has nothing left after `trim()` — no HTML metacharacter can survive it. The comment at `:171` is what keeps that true for the next reader.

---

### Two small ones, neither blocking

- `oauth-callback-handler.service.ts:181-184`: the reasoning is Node's, but this runs in the renderer. I checked in Chrome — `new URL('http://[')` throws `TypeError: Failed to construct 'URL': Invalid URL` whose only own key is `message`; there's no `input` property, and the raw URL isn't in the message. (`com.super-productivity.app://plugin-oauth-callback?code=X%` doesn't even throw — non-special schemes are very hard to make `new URL` reject.) Logging only the scheme is still the right call, but the comment states a fact that isn't true where the code runs, so someone will eventually "correct" it. Related: the `try` at `:164` also wraps `handleRedirectCode`/`handleRedirectError`, so a throw from *those* is now reported as `parse_error` with no diagnostic whatsoever — narrowing the `try` to `new URL(url)` would fix both the comment and the misattribution.
- Still nothing exercises `routeCapacitorAppUrl(url)` with its **default** sinks against a service holding the **default** token, so the producer↔consumer wiring itself is unasserted — the new singleton test touches the real sink but only for non-retention. One test that routes an OAuth URL through the defaults and asserts the service saw it would pin the wire. It wouldn't have caught the platform gate, so genuinely optional.

Still true from round 2 and untouched here: #8515 is open and adds a `quick-add` case to the same `switch` plus edits to `protocol-handler.test.cjs`, so whichever lands second gets a textual conflict.

---

### Verdict

Approving. The Android regression is fixed against the right reference, the cold-start decision landed on less code rather than more and is correct for reasons I could verify independently, the doc rows agree, and the one test that couldn't fail now can. Good round.

johannesjo:
Thank you very much! <3

PR #9130, opened 2026-07-18
fix(date-time): fall back to default locale instead of blank dates; register English region locale data

KamilDev:
## Summary

Focused follow-up to #7731, carrying over the two pieces @johannesjo suggested keeping after #8023 shipped the browser-locale resolution for "System default":

1. **`LocaleDatePipe`: fall back to `DEFAULT_LOCALE` instead of rendering blank.** Since #8023, `currentLocale()` for "System default" can be any browser locale (via `getBrowserCultureLang()`). `LocaleDatePipe` builds Angular's `DatePipe` from it, and Angular throws `NG0701` when locale data for that tag isn't registered — registration only covers the app's UI languages. The pipe's `catch` returned `null`, so users whose OS language is outside that set (Thai, Danish, Greek, Hebrew, …) get **blank date fields** on the default setting — in reminder toasts, planner, schedule, and sync-conflict UI (~15 templates + effects). The pipe now retries with `DEFAULT_LOCALE` (en-GB), mirroring the `safeFormatDate` pattern, so those users see en-GB-formatted dates instead of nothing.

2. **Register Angular locale data for eight English region variants** (en-AU, en-CA, en-IE, en-IN, en-NZ, en-PH, en-SG, en-ZA). Without registration, `DatePipe` falls back `en-au` → `en`, which is registered with **en-GB** data — so toast/snack messages show 24-hour `13:00` while the `Intl`-based paths (fixed by #8023) correctly show `1:00 PM` on the same machine. The variants live in a separate `NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS` registry so the `DateTimeLocale` union stays limited to dropdown-selectable locales. Because `LocaleDatePipe` is pure — the first render of a date caches its locale resolution for the session — registration must happen **before Angular's first render**, not at idle time: the en-GB default is registered at module scope and the browser's matching regional chunk is awaited by an app initializer. The registration is extracted into `src/app/core/locale-registration.ts` (unit-testable), with a bounded 1.5s `Promise.race` so a stalled chunk degrades to the default locale rather than holding up bootstrap.

3. **Register en-US up front too.** Seeding en-GB under the bare `en` id changes what an unregistered `en-us` lookup resolves to before the idle-time loop runs: it now falls through `en` (=en-GB) and renders day-first/24h, frozen there by the pure pipe. On master, where `en` was seeded only inside the bootstrap `.then`, the same first-paint lookup fell through to Angular's built-in en-US and was correct. So en-US is registered under `en-us` alongside the en-GB seed, keeping an explicit "System default" or dropdown en-US choice month-first/12h from first paint (the data file self-reports `en`, so an explicit id is required — a keyless call would clobber the en-GB seed).

Unlike #7731, this does **not** touch `DateTimeFormatService` or the `DateAdapter` — #8023's resolution logic is untouched.

## Changes

| File | Change |
| --- | --- |
| `src/app/ui/pipes/locale-date.pipe.ts` | Retry failed `transform` with a reused `DatePipe(DEFAULT_LOCALE)`; warn once per unregistered locale. Still returns `null` when the value itself is unformattable. |
| `src/app/core/locale.constants.ts` | Add `NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS` (8 English region variants, lazy imports, not dropdown-selectable). |
| `src/app/core/locale-registration.ts` | New. `registerDefaultLocale()` (module-scope: en-GB under `en`, en-US under `en-us`) and `registerNavigatorLocale()` (app-initializer: the browser's regional chunk, bounded 1.5s wait, never rejects). |
| `src/main.ts` | Call `registerDefaultLocale()` at module scope and await `registerNavigatorLocale()` via `provideAppInitializer` — both before first render. |
| `src/app/ui/pipes/locale-date.pipe.spec.ts` | Unregistered `th-TH` renders via en-GB fallback instead of `null`; unformattable value still returns `null` without poisoning the per-locale warn dedup. |
| `src/app/core/locale-registration.spec.ts` | New. en-GB seed under `en`, en-US under `en-us`, per-navigator registration (synthetic locales, pollution-proof), `getBrowserCultureLang` default, and the bounded stalled-load path. |
| `src/app/core/locale.constants.spec.ts` | New. Each of the 8 variants self-reports the BCP-47 id its map key promises (keyless registration is safe) and renders its expected 12h/24h `shortTime` against the en-GB baseline. |

## Test plan

- [x] `locale-date.pipe.spec.ts`, `locale-registration.spec.ts`, `locale.constants.spec.ts` — pass (31 specs). Each new assertion was sabotage-checked: reverting the fix it covers makes it fail (registration no-op → the six 12h variants fail; `.toLowerCase()`/default-arg regressions → the synthetic-locale specs fail; timeout regression → the stalled-load spec fails without leaking the mocked clock into sibling suites).
- [x] `npm run test:file src/app/core/date-time-format/date-time-format.service.spec.ts` — passes.
- [x] Full unit suite green (13314) and `checkFile` clean on all touched files.

Supersedes the salvageable part of #7731.

### DIFF
--- a/docs/wiki/3.02-Settings-and-Preferences.md
+++ b/docs/wiki/3.02-Settings-and-Preferences.md
@@ -11,7 +11,9 @@ Note: Differences between the Web app and the Desktop app are specified in [[3.0
 - **Datetime format locale** — Controls numeric date formatting and the
   12/24-hour time format. The ISO 8601 option uses `YYYY-MM-DD` and a 24-hour
   clock. With ISO 8601 selected, weekday labels in Schedule, Habits, and Planner
-  follow the app language.
+  follow the app language. The System default option follows your browser or
+  operating system regional locale; when formatting data for a locale is not
+  available, dates fall back to British English (en-GB) formatting.
 
 #### global-settings.App-Features
 
--- a/src/app/core/locale-registration.spec.ts
+++ b/src/app/core/locale-registration.spec.ts
@@ -1,5 +1,6 @@
 import { formatDate, registerLocaleData } from '@angular/common';
 import localeEn from '@angular/common/locales/en';
+import localeEnAu from '@angular/common/locales/en-AU';
 import { TranslateService } from '@ngx-translate/core';
 import { registerDefaultLocale, registerNavigatorLocale } from './locale-registration';
 import { NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS } from './locale.constants';
@@ -27,19 +28,57 @@ describe('locale-registration', () => {
       expect(formatDate(onePm, 'shortDate', 'en')).toBe('15/01/2024');
       expect(formatDate(onePm, 'shortTime', 'en')).toBe('13:00');
     });
+
+    it('registers en-US under en-us so an explicit US locale is month-first/12h from first paint, not the en-GB seed', () => {
+      registerDefaultLocale();
+      // Without the explicit en-US registration, 'en-us' would resolve through
+      // the en-GB 'en' seed above and render '15/01/2024' / '13:00'.
+      expect(formatDate(onePm, 'shortDate', 'en-us')).toBe('1/15/24');
+      expect(formatDate(onePm, 'shortTime', 'en-us')).toMatch(/^1:00/);
+      // …and the en-GB seed under bare 'en' stays intact.
+      expect(formatDate(onePm, 'shortTime', 'en')).toBe('13:00');
+    });
   });
 
   describe('registerNavigatorLocale', () => {
+    // The real en-* variants leak into the process-global locale registry from
+    // locale.constants.spec (Angular has no unregister API), so asserting on
+    // them would pass on that leak alone — Jasmine randomises suite order, so
+    // whether the sibling suite ran first flips the result. Instead each spec
+    // registers a synthetic 12h locale (real en-AU data re-tagged under an id
+    // no other suite touches): a 12h render can then only mean *this* call
+    // registered it, and an unregistered synthetic falls through 'en' to the
+    // en-GB (24h) baseline below.
+    const syntheticMapKeys: string[] = [];
+    const addSynthetic12hLocale = (tag: string): string => {
+      // registerNavigatorLocale derives the map key from the (lower-cased,
+      // '_'-joined) nav language and registers keyless (by the data's self-id),
+      // so key and self-id are both derived from `tag`.
+      const mapKey = tag.toLowerCase().replace(/-/g, '_');
+      const data = [...(localeEnAu as unknown[])];
+      data[0] = tag.toLowerCase();
+      NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS[mapKey] = () =>
+        Promise.resolve({ default: data });
+      syntheticMapKeys.push(mapKey);
+      return tag;
+    };
+
     beforeAll(() => {
       // Match prod ordering: the default locale is registered before the app
-      // initializer runs, so unregistered en-* resolves to en-GB (24h) — which
-      // is exactly what the 12h assertion below must flip.
+      // initializer runs, so an unregistered synthetic en-* resolves to en-GB
+      // (24h) — which is exactly what the 12h assertions below must flip.
       registerDefaultLocale();
     });
 
-    it('registers the data for a matched navigator.language (en-AU renders 12h, not the en-GB 24h fallback)', async () => {
-      await registerNavigatorLocale('en-AU');
-      expect(formatDate(onePm, 'shortTime', 'en-AU')).toMatch(/^1:00/);
+    afterEach(() => {
+      syntheticMapKeys.forEach((key) => delete NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS[key]);
+      syntheticMapKeys.length = 0;
+    });
+
+    it('registers the data for a matched navigator.language (renders 12h, not the en-GB 24h fallback)', async () => {
+      const tag = addSynthetic12hLocale('en-XA');
+      await registerNavigatorLocale(tag);
+      expect(formatDate(onePm, 'shortTime', tag.toLowerCase())).toMatch(/^1:00/);
     });
 
     it('resolves without throwing for a locale outside the navigator-fallback map', async () => {
@@ -48,10 +87,13 @@ describe('locale-registration', () => {
 
     it('defaults to the same browser locale the date pipe resolves (getBrowserCultureLang, not navigator.language)', async () => {
       // Reading navigator.language instead would register data for a locale the
-      // pipe never asks for whenever the two disagree.
-      spyOn(TranslateService, 'getBrowserCultureLang').and.returnValue('en-NZ');
+      // pipe never asks for whenever the two disagree. The synthetic tag is not
+      // navigator.language, so the sabotage (default → navigator.language)
+      // leaves it unregistered and the 12h assertion fails.
+      const tag = addSynthetic12hLocale('en-XB');
+      spyOn(TranslateService, 'getBrowserCultureLang').and.returnValue(tag);
       await registerNavigatorLocale();
-      expect(formatDate(onePm, 'shortTime', 'en-NZ')).toMatch(/^1:00/);
+      expect(formatDate(onePm, 'shortTime', tag.toLowerCase())).toMatch(/^1:00/);
     });
 
     it('resolves without throwing when the browser reports no culture language', async () => {
@@ -59,21 +101,30 @@ describe('locale-registration', () => {
       await expectAsync(registerNavigatorLocale()).toBeResolved();
     });
 
-    it('gives up on a stalled chunk load instead of holding up bootstrap', async () => {
-      // en-IE is 12h; asserting the en-GB 24h fallback proves the stalled load
-      // was abandoned rather than awaited (and that we did not hang here).
-      const origLoad = NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS['en_ie'];
-      NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS['en_ie'] = () => new Promise(() => {});
-      jasmine.clock().install();
-      try {
-        const pending = registerNavigatorLocale('en-IE');
-        jasmine.clock().tick(1500);
+    describe('when the chunk load stalls', () => {
+      // Clock in before/afterEach, not a try/finally inside the spec: if the
+      // timeout ever regresses to a bare `await load()`, the assertion below
+      // never settles and a `finally` would never run — leaking jasmine.clock()
+      // and the mutated import map into every sibling suite for the rest of the
+      // run. afterEach still runs when the spec times out, containing the leak.
+      beforeEach(() => jasmine.clock().install());
+      afterEach(() => jasmine.clock().uninstall());
+
+      it('gives up on a stalled chunk load instead of holding up bootstrap', async () => {
+        // A synthetic 12h loader that never resolves. Giving up leaves the tag
+        // unregistered → it renders the en-GB 24h fallback; awaiting the stall
+        // would hang the spec. A real 24h locale (en-IE) cannot discriminate:
+        // its data is byte-identical to en-GB, so "loaded" and "abandoned"
+        // render the same 13:00.
+        const tag = 'en-XC';
+        const mapKey = 'en_xc';
+        syntheticMapKeys.push(mapKey);
+        NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS[mapKey] = () => new Promise<never>(() => {});
+        const pending = registerNavigatorLocale(tag);
+        jasmine.clock().tick(2000);
         await expectAsync(pending).toBeResolved();
-        expect(formatDate(onePm, 'shortTime', 'en-IE')).toBe('13:00');
-      } finally {
-        jasmine.clock().uninstall();
-        NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS['en_ie'] = origLoad;
-      }
+        expect(formatDate(onePm, 'shortTime', tag.toLowerCase())).toMatch(/^13:00/);
+      });
     });
   });
 });
--- a/src/app/core/locale-registration.spec.ts
+++ b/src/app/core/locale-registration.spec.ts
@@ -1,6 +1,8 @@
 import { formatDate, registerLocaleData } from '@angular/common';
 import localeEn from '@angular/common/locales/en';
+import { TranslateService } from '@ngx-translate/core';
 import { registerDefaultLocale, registerNavigatorLocale } from './locale-registration';
+import { NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS } from './locale.constants';
 
 /**
  * Exercises the production registration path used by main.ts (module-scope
@@ -42,6 +44,36 @@ describe('locale-registration', () => {
 
     it('resolves without throwing for a locale outside the navigator-fallback map', async () => {
       await expectAsync(registerNavigatorLocale('th-TH')).toBeResolved();
+    });
+
+    it('defaults to the same browser locale the date pipe resolves (getBrowserCultureLang, not navigator.language)', async () => {
+      // Reading navigator.language instead would register data for a locale the
+      // pipe never asks for whenever the two disagree.
+      spyOn(TranslateService, 'getBrowserCultureLang').and.returnValue('en-NZ');
+      await registerNavigatorLocale();
+      expect(formatDate(onePm, 'shortTime', 'en-NZ')).toMatch(/^1:00/);
+    });
+
+    it('resolves without throwing when the browser reports no culture language', async () => {
+      spyOn(TranslateService, 'getBrowserCultureLang').and.returnValue(undefined);
+      await expectAsync(registerNavigatorLocale()).toBeResolved();
+    });
+
+    it('gives up on a stalled chunk load instead of holding up bootstrap', async () => {
+      // en-IE is 12h; asserting the en-GB 24h fallback proves the stalled load
+      // was abandoned rather than awaited (and that we did not hang here).
+      const origLoad = NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS['en_ie'];
+      NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS['en_ie'] = () => new Promise(() => {});
+      jasmine.clock().install();
+      try {
+        const pending = registerNavigatorLocale('en-IE');
+        jasmine.clock().tick(1500);
+        await expectAsync(pending).toBeResolved();
+        expect(formatDate(onePm, 'shortTime', 'en-IE')).toBe('13:00');
+      } finally {
+        jasmine.clock().uninstall();
+        NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS['en_ie'] = origLoad;
+      }
     });
   });
 });
--- a/src/app/core/locale-registration.spec.ts
+++ b/src/app/core/locale-registration.spec.ts
@@ -1,0 +1,47 @@
+import { formatDate, registerLocaleData } from '@angular/common';
+import localeEn from '@angular/common/locales/en';
+import { registerDefaultLocale, registerNavigatorLocale } from './locale-registration';
+
+/**
+ * Exercises the production registration path used by main.ts (module-scope
+ * registerDefaultLocale + app-initializer registerNavigatorLocale), asserting
+ * the user-visible formats rather than registry internals.
+ */
+describe('locale-registration', () => {
+  // 1pm discriminates 12h ("1:00 …") from 24h ("13:00"); Jan 15 discriminates
+  // day-first shortDate (en-GB "15/01/2024") from month-first (en-US "1/15/24").
+  const onePm = new Date(2024, 0, 15, 13, 0);
+
+  afterAll(() => {
+    // The locale registry is module-global across the Karma run. Restore 'en'
+    // to the baked-in en-US data, which is observably identical to the
+    // pristine unregistered state (Angular's parent-locale shortcut).
+    registerLocaleData(localeEn, 'en');
+  });
+
+  describe('registerDefaultLocale', () => {
+    it('registers en-GB data under the bare en id (day-first shortDate, 24h shortTime)', () => {
+      registerDefaultLocale();
+      expect(formatDate(onePm, 'shortDate', 'en')).toBe('15/01/2024');
+      expect(formatDate(onePm, 'shortTime', 'en')).toBe('13:00');
+    });
+  });
+
+  describe('registerNavigatorLocale', () => {
+    beforeAll(() => {
+      // Match prod ordering: the default locale is registered before the app
+      // initializer runs, so unregistered en-* resolves to en-GB (24h) — which
+      // is exactly what the 12h assertion below must flip.
+      registerDefaultLocale();
+    });
+
+    it('registers the data for a matched navigator.language (en-AU renders 12h, not the en-GB 24h fallback)', async () => {
+      await registerNavigatorLocale('en-AU');
+      expect(formatDate(onePm, 'shortTime', 'en-AU')).toMatch(/^1:00/);
+    });
+
+    it('resolves without throwing for a locale outside the navigator-fallback map', async () => {
+      await expectAsync(registerNavigatorLocale('th-TH')).toBeResolved();
+    });
+  });
+});
--- a/src/app/core/locale-registration.spec.ts
+++ b/src/app/core/locale-registration.spec.ts
@@ -85,6 +85,23 @@ describe('locale-registration', () => {
       await expectAsync(registerNavigatorLocale('th-TH')).toBeResolved();
     });
 
+    it('resolves without throwing when the matched chunk load rejects', async () => {
+      // The unmatched-key spec above returns before load() is ever called, so it
+      // cannot cover the catch. This one matches the map and rejects, which is
+      // the shape that matters: the promise is awaited by an app initializer, so
+      // an escaping rejection would reject bootstrap and block first render
+      // instead of degrading to the default locale.
+      const tag = 'en-XD';
+      const mapKey = 'en_xd';
+      syntheticMapKeys.push(mapKey);
+      NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS[mapKey] = () =>
+        Promise.reject(new Error('chunk load failed'));
+      await expectAsync(registerNavigatorLocale(tag)).toBeResolved();
+      // …and the failed locale stays unregistered, so it renders the en-GB
+      // (24h) default rather than a half-applied state.
+      expect(formatDate(onePm, 'shortTime', tag.toLowerCase())).toMatch(/^13:00/);
+    });
+
     it('defaults to the same browser locale the date pipe resolves (getBrowserCultureLang, not navigator.language)', async () => {
       // Reading navigator.language instead would register data for a locale the
       // pipe never asks for whenever the two disagree. The synthetic tag is not
--- a/src/app/core/locale-registration.ts
+++ b/src/app/core/locale-registration.ts
@@ -1,0 +1,45 @@
+import { registerLocaleData } from '@angular/common';
+import { Log } from './log';
+import {
+  DEFAULT_LANGUAGE,
+  DEFAULT_LOCALE_DATA,
+  NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS,
+} from './locale.constants';
+
+/**
+ * Registers the statically imported default locale data (en-GB under the bare
+ * 'en' id). Must run before Angular's first render: LocaleDatePipe is pure, so
+ * a date rendered earlier would cache Angular's built-in en-US resolution for
+ * the session. main.ts calls this at module scope, before bootstrapApplication.
+ */
+export const registerDefaultLocale = (): void => {
+  registerLocaleData(DEFAULT_LOCALE_DATA, DEFAULT_LANGUAGE);
+};
+
+/**
+ * Loads and registers the regional locale data matching the browser's own
+ * locale, when it's one of the navigator-only variants (en-AU, en-CA, … — see
+ * NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS). Awaited by an app initializer so the
+ * data is registered before first render; otherwise a pure LocaleDatePipe
+ * caches the en-GB parent format for the session.
+ *
+ * Never rejects: a failed chunk load logs and falls back to default-locale
+ * rendering instead of blocking bootstrap.
+ */
+export const registerNavigatorLocale = async (
+  navLanguage: string = navigator.language,
+): Promise<void> => {
+  const key = navLanguage.toLowerCase().replace(/-/g, '_');
+  const load = NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS[key];
+  if (!load) {
+    return;
+  }
+  try {
+    const m = await load();
+    // No explicit id: these data files self-report ids matching their keys
+    // (asserted in locale.constants.spec.ts), so the self-id is typo-proof.
+    registerLocaleData(m.default);
+  } catch (e) {
+    Log.err(`Failed to load locale ${key}`, e);
+  }
+};
--- a/src/app/core/locale-registration.ts
+++ b/src/app/core/locale-registration.ts
@@ -1,10 +1,18 @@
 import { registerLocaleData } from '@angular/common';
+import { TranslateService } from '@ngx-translate/core';
 import { Log } from './log';
 import {
   DEFAULT_LANGUAGE,
   DEFAULT_LOCALE_DATA,
   NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS,
 } from './locale.constants';
+
+/**
+ * Upper bound on how long bootstrap waits for the regional locale chunk. On
+ * timeout we render with the default locale — exactly the behavior before this
+ * registration existed — rather than hold up first render on a stalled network.
+ */
+const LOCALE_LOAD_TIMEOUT_MS = 1500;
 
 /**
  * Registers the statically imported default locale data (en-GB under the bare
@@ -23,23 +31,44 @@ export const registerDefaultLocale = (): void => {
  * data is registered before first render; otherwise a pure LocaleDatePipe
  * caches the en-GB parent format for the session.
  *
- * Never rejects: a failed chunk load logs and falls back to default-locale
- * rendering instead of blocking bootstrap.
+ * The default argument reads the browser locale through the exact same call
+ * `DateTimeFormatService` resolves the pipe's locale with
+ * (`getBrowserCultureLang()` → `navigator.languages[0]`). Reading
+ * `navigator.language` here instead would let the two disagree, and we would
+ * then register data for a locale the pipe never asks for.
+ *
+ * Never rejects and gives up after {@link LOCALE_LOAD_TIMEOUT_MS}: a failed or
+ * stalled chunk load degrades to default-locale rendering instead of failing or
+ * holding up bootstrap.
  */
 export const registerNavigatorLocale = async (
-  navLanguage: string = navigator.language,
+  navLanguage: string | undefined = TranslateService.getBrowserCultureLang(),
 ): Promise<void> => {
-  const key = navLanguage.toLowerCase().replace(/-/g, '_');
-  const load = NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS[key];
+  const key = navLanguage?.toLowerCase().replace(/-/g, '_');
+  const load = key ? NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS[key] : undefined;
   if (!load) {
     return;
   }
+  let timeoutId: ReturnType<typeof setTimeout> | undefined;
   try {
-    const m = await load();
+    // Promise.race consumes a late rejection from load(), so a chunk that fails
+    // after the timeout won this race cannot surface as an unhandled rejection.
+    const m = await Promise.race([
+      load(),
+      new Promise<null>((resolve) => {
+        timeoutId = setTimeout(() => resolve(null), LOCALE_LOAD_TIMEOUT_MS);
+      }),
+    ]);
+    if (!m) {
+      Log.err(`Timed out loading locale ${key}`);
+      return;
+    }
     // No explicit id: these data files self-report ids matching their keys
     // (asserted in locale.constants.spec.ts), so the self-id is typo-proof.
     registerLocaleData(m.default);
   } catch (e) {
     Log.err(`Failed to load locale ${key}`, e);
+  } finally {
+    clearTimeout(timeoutId);
   }
 };
--- a/src/app/core/locale-registration.ts
+++ b/src/app/core/locale-registration.ts
@@ -1,7 +1,9 @@
 import { registerLocaleData } from '@angular/common';
+import localeEnUs from '@angular/common/locales/en';
 import { TranslateService } from '@ngx-translate/core';
 import { Log } from './log';
 import {
+  DateTimeLocales,
   DEFAULT_LANGUAGE,
   DEFAULT_LOCALE_DATA,
   NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS,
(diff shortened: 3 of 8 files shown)


johannesjo (commented):
Thanks for the focused follow-up — this is a well-diagnosed fix, and splitting it out from #7731 was the right call. I verified both claims independently rather than taking them at face value; both hold.

## What I verified

**Claim 1 (NG0701 → blank dates) — confirmed.** Angular's `findLocaleData` tries the exact tag → parent → `if (parentLocale === 'en') return localeEn` → else throws NG0701. So `th-TH` → `th` unregistered → not `en` → throw, and the old `catch` returned `null`. Reachable exactly as described.

**Claim 2 (en-* renders 24h) — confirmed, and the root cause is ours.** `src/main.ts:358` registers **en-GB data under the bare id `'en'`**, so `en-AU` resolves to the *registered* `en` (=en-GB) and never reaches Angular's built-in en-US. Probe on Angular 21.2.17:

```
before (en-GB data aliased as 'en') = 13:00
after  (en-AU registered)           = 1:00 pm
```

**The fallback can never itself throw.** `DEFAULT_LOCALE` (`'en-gb'`) resolves via parent `'en'`, registered *synchronously* at bootstrap before any render — so the code comment asserting this is accurate.

**`registerLocaleData` does normalize `en_au` → `en-au`** (`.toLowerCase().replace(/_/g, '-')`), so the snake_case keys work.

**The test is not vacuous.** I reverted the pipe to master's version and re-ran: `Expected null to be 'January'` — the new spec genuinely gates the fix.

**Also checked:** `registerRemainingLocales` runs exactly once (if/else, not both branches); `checkFile` clean on all three source files; 13/13 pipe specs and 25/25 `date-time-format` specs pass.

One nice side effect the description doesn't mention: master logged `Log.warn('...', value, e)` — the raw date, which is user content, and our log history is user-exportable. This drops it. Worth keeping.

## One thing I'd like changed before merge

`_warnedLocales` and `_fallbackDatePipe` are **per-instance**, but the comment says "warn at most once per unregistered locale." Angular creates one pure-pipe instance per binding per embedded view, and `| localeDate` is used *inside* `@for` in several templates (`habit-tracker` has 2 bindings per day, plus `scheduled-list-page`, `issue-content` comments, `dialog-view-task-reminders`, `issue-panel-calendar-agenda`).

I built a host component matching the habit-tracker shape (30 days × 2 bindings) and counted:

```
spans=60  firstText="Mon"  warns=60
```

So it's once per *instance*, not once per locale — and since instances are view-scoped, re-rendering these views keeps appending. That lands on exactly the users this PR is for (Thai/Danish/Greek/Hebrew on "System default"), in an exportable log.

Hoisting both to module scope fixes it and also drops a per-instance `DatePipe` allocation:

```ts
// Module-scoped: Angular creates one pure-pipe instance per binding per
// embedded view, so per-instance state would warn once per row, not once.
const FALLBACK_DATE_PIPE = new DatePipe(DEFAULT_LOCALE);
const WARNED_LOCALES = new Set<string>();
```

I tested this: same render (`spans=60 firstText="Mon"`), `warns=1`, 13/13 specs still pass, lint clean.

## Nits (take or leave)

- The second spec passes on master too, so it's a regression guard rather than a fix-gating test — the `th-TH` one carries the weight. Fine as is, just noting it.
- All 8 data files self-report ids matching their keys (`en-AU`, `en-CA`, …), so `registerLocaleData(m.default)` without the second arg would be equivalent and typo-proof here. The *existing* loop needs the explicit key because it deliberately aliases (`en`→en-GB, `zh_tw`→zh); the new map doesn't.

## Scope note

Of the 8 variants, 6 actually change `shortTime` — en-IE and en-ZA render `13:00` either way:

```
en_au=1:00 pm  en_ca=1:00 p.m.  en_ie=13:00  en_in=1:00 pm
en_nz=1:00 pm  en_ph=1:00 PM    en_sg=1:00 pm  en_za=13:00
```

Registering them is still worth it (date order, month names), and en-CA `p.m.` / en-PH `PM` show that explicit registration beats any blanket fallback. Cost is ~12KB raw across 8 lazy chunks fetched at idle for every user — consistent with the existing loop that already fetches ~45, so more of an existing pattern rather than a new one.

Agreed on leaving the underlying `en` ≡ en-GB aliasing alone. Worth recording why: dropping that alias would let unregistered `en-*` fall through to Angular's built-in en-US, which is right for AU/CA/IN/NZ/PH/SG but would *regress* en-IE and en-ZA, which genuinely are 24h. Explicit per-region registration is the correct fix; remaining variants (en-JM, en-KE, en-NG, en-HK, …) still inherit en-GB, which is the acceptable tail.

Happy to merge once the module-scope change is in.

johannesjo:
Follow-up to my review above. I ran a deeper pass and found a defect I missed the first time, so I need to walk back my "happy to merge once the module-scope change is in" — sorry for the moving target. The diagnosis in this PR is still good; it's the error handling around it that isn't ready.

## The one I missed: a bad value permanently silences the new warning

The `catch` can't tell "unregistered locale" from "unparseable value", but the message assumes the former. Probing the current branch:

```
transform('invalid-date-string', 'short', undefined, 'th-TH')
  → warns=1, message: cannot format with locale "th-TH", using "en-gb"
transform(validDate, 'short', undefined, 'th-TH')   ← genuine NG0701
  → warns=1   (no second warning — suppressed)
```

Two problems in one: it blames a locale that was never at fault, and it inserts `th-TH` into `_warnedLocales`, so the *real* NG0701 for that locale is never logged. One malformed date turns off the exact diagnostic this PR adds.

## And the dedup scope makes it worse than console noise

Following up on the module-scope point from my first review — I under-sold it. `Log` caps history at 1000 entries (`src/app/core/log.ts:97`) and `slice(-maxHistorySize)` (`:164`) drops the **oldest**. So this isn't just console spam: on an affected user's machine a single render evicts the exportable log history that bug reports depend on. Measured with a host component matching the `habit-tracker` shape (30 days × 2 bindings): **60 warnings from one render**.

## A fix that looks obvious here does not work — worth knowing before you try it

My instinct was to gate on the NG0701 error code. That does not work, and I'd have shipped a no-op if I hadn't checked. `DatePipe` rewraps *both* failure kinds:

```
[unregistered locale] code=2100  msg=NG02100: InvalidPipeArgument: 'NG0701: Missin…
[bad value]           code=2100  msg=NG02100: InvalidPipeArgument: 'NG02311: Unabl…
```

Same code for both. And `ngDevMode` strips both messages in production builds, so in a real build the error object carries no distinguishing information at all. Any `code === 701` or message-sniffing check silently never fires in prod.

The reliable discriminant is *"did the fallback succeed?"* — if it did, the value was fine and the locale is the cause. That needs no error introspection and works in prod:

```ts
// Module-scoped: Angular creates one pure-pipe instance per binding per
// embedded view, so per-instance state would warn once per row, not once.
const FALLBACK_DATE_PIPE = new DatePipe(DEFAULT_LOCALE);
const WARNED_LOCALES = new Set<string>();

// …in transform():
} catch {
  let fallback: string | null;
  try {
    fallback = FALLBACK_DATE_PIPE.transform(value, format, timezone, DEFAULT_LOCALE);
  } catch {
    // Both locales failed => the value itself is unformattable, not a locale
    // problem. Stay silent (matching safeFormatDate) so a bad value cannot
    // poison the per-locale warn set below.
    return null;
  }
  if (!WARNED_LOCALES.has(effectiveLocale)) {
    WARNED_LOCALES.add(effectiveLocale);
    Log.warn(
      `LocaleDatePipe: cannot format with locale "${effectiveLocale}", ` +
        `using "${DEFAULT_LOCALE}"`,
    );
  }
  return fallback;
}
```

I tested this on the branch: bad value → `warns=0` and no poisoning, genuine NG0701 → `warns=1`, the 60-span render → `warns=1`, rendering unchanged, 13/13 existing specs pass, `checkFile` clean. Dropping `e` from the log is deliberate — it carries no message in prod anyway, and in dev its text embeds the raw date value, which rule 9 wants kept out of exportable history.

## Smaller points

**The `th-TH` spec proves less than it looks.** `main.ts` never runs under Karma, so nothing registers `'en'`, and the fallback resolves to Angular's built-in **en-US**, not en-GB:

```
transform(date, 'short', undefined, 'th-TH') → 1/15/24, 2:30 PM
```

In production `main.ts:358` registers en-GB under `'en'`, so the same call gives `15/01/2024, 14:30`. `'MMMM'` → `'January'` is identical in both, so the assertion can't tell them apart. Registering en-GB in `beforeAll` and asserting a format-sensitive value (`'shortDate'`) would pin the real behavior. I was wrong in my first review to call the code comment simply "accurate" — it's accurate in prod, but untrue under test, which is exactly where it's being verified.

**The second spec doesn't gate anything.** `formatDate` calls `toDate(value)` *before* any locale lookup, so an unparseable string throws regardless of locale — the `'th-TH'` argument is inert, and the spec passes identically with and without the fix (I confirmed by sabotage: reverting the pipe fails only the first spec). Worth rescoping to what it actually covers, or dropping it.

**The registration half has no tests.** Nothing asserts `en-AU` renders `1:00 pm` rather than `13:00` — the user-visible symptom this PR exists to fix. `src/app/features/schedule/schedule/schedule-locale-ng0701.spec.ts:180-190` already has the loop-over-the-registry pattern to copy.

**Fix #2 is partially self-defeating.** Registration runs on `requestIdleCallback`, but `LocaleDatePipe` is *pure* — a date rendered before idle resolves via en-GB and won't re-render until its binding changes, so an en-AU user can keep seeing 24h for the session. Same race `safe-format-date.ts:8-14` documents. Not a regression, but the PR delivers less than the description implies; eagerly registering just the browser's own culture locale would close it.

## Unrelated pre-existing bug I noticed

`src/app/util/format-month-day.util.ts:39` does `new LocaleDatePipe().transform(date, 'shortDate', locale)`. Two bugs: the class uses `inject()` in a field initializer, so constructing it outside an injection context always throws NG0203 and that branch is dead; and the third positional arg is `timezone`, not `locale`. Not this PR's job — I'll file it separately.

## To merge

The two error-handling defects (module-scope state + the fallback-succeeded discriminant) are the blockers; they're the single snippet above. The vacuous spec should be rescoped or dropped, and a small test for the `en-AU` 12h behavior would cover the half of the PR that currently has none.

Thanks for your patience with the extra round — the underlying diagnosis here (bare `en` pinned to en-GB data, and the `registerLocaleData` underscore normalization) is genuinely subtle and you got both right.

KamilDev:
Applied all of it in e9e2949c5 — thanks for the thorough second pass, the fallback-succeeded discriminant especially; I'd have reached for the error code too.

**Error handling** (`locale-date.pipe.ts`):
- `FALLBACK_DATE_PIPE` and `WARNED_LOCALES` hoisted to module scope.
- Warn moved after the fallback attempt: both locales failing → silent `null` (matching `safeFormatDate`), so a bad value can neither blame the locale nor poison the warn set. The caught error is dropped from the log call.

**Specs** (`locale-date.pipe.spec.ts`):
- The th-TH spec now registers en-GB under `'en'` in `beforeAll` (matching `main.ts`) and asserts `'shortDate'` → `15/01/2024`, pinning the prod fallback rather than Karma's baked-in en-US. `afterAll` restores `'en'` to the baked-in en-US data, which is observably identical to the pristine registry (Angular's parent-`en` shortcut), so later suites in the Karma run are unaffected.
- The vacuous spec is rescoped to gate the new behavior: an unformattable value returns `null` with zero warns, then a valid date with the same locale still warns — proving the warn set wasn't poisoned.
- New spec: two pipe instances, same unregistered locale → exactly one warn.

**Registration coverage** (new `locale.constants.spec.ts`):
- Registers all 8 variants exactly as `main.ts` now does — `registerLocaleData(m.default)` with no explicit id — and asserts 12h `shortTime` for en-AU/CA/IN/NZ/PH/SG and 24h for en-IE/ZA, per your table. Asserted as `^1:00` vs `^13:00` since the exact spacing/case of the dayperiod varies per region (`pm` / `p.m.` / `PM`).
- Also asserts each data file self-reports the BCP-47 id its map key promises — the property that makes keyless registration typo-proof — plus a completeness check so a new map entry without a 12h/24h expectation fails the suite.

**main.ts**:
- The new map registers keyless (took the nit); the aliasing loop keeps its explicit ids.
- The browser's own regional locale now loads eagerly at bootstrap instead of at idle, per your note on the pure-pipe race. This narrows rather than fully closes the window (the import is still async), but the fetch starts before first render instead of after idle.

**Diff note**: the branch now carries a merge of #9131 (timezone-independent specs). The pre-push hook runs the full unit suite, and on Windows outside Europe/Berlin, Chromium ignores the `TZ` env var, so 8 date specs fail on any branch without that fix. Those two spec files wash out of this diff once #9131 merges — happy to rebase then.

johannesjo (changes requested):
## Summary

Thanks for addressing the earlier feedback. I need to correct one statement from my previous review: the default locale is **not** registered before the first render. I rechecked the Angular 21 bootstrap path and the current production wiring, and the new “eager” registration is still racy.

The fallback/error-handling work is solid, but there are two blockers before merge.

## Important

1. **[`src/main.ts:357-370`](https://github.com/super-productivity/super-productivity/blob/e9e2949c5a885655d710136ec33217764544f9d8/src/main.ts#L357-L370) — locale registration starts after Angular's first render.**

   Both `registerLocaleData(DEFAULT_LOCALE_DATA, DEFAULT_LANGUAGE)` and `eagerNavLocaleLoad()` are inside `bootstrapApplication(...).then(...)`. Angular has already bootstrapped/rendered the root component when that promise resolves ([API](https://angular.dev/api/platform-browser/bootstrapApplication)), and the regional import is then started without being awaited.

   A date rendered before line 358 can therefore cache Angular's built-in en-US fallback; one rendered after line 358 but before the regional chunk resolves can cache the en-GB parent format. For an unsupported locale such as `th-TH`, the new fallback can likewise cache en-US instead of the promised en-GB. Registering locale data later changes no pipe input, so a pure pipe does not rerun ([pure-pipe behavior](https://angular.dev/guide/templates/pipes)); the wrong format can remain for that view/session.

   The new [registry spec](https://github.com/super-productivity/super-productivity/blob/e9e2949c5a885655d710136ec33217764544f9d8/src/app/core/locale.constants.spec.ts#L29-L36) manually imports and registers every locale, so it cannot catch broken `main.ts` wiring.

   **Smallest fix:** register the static en-GB data before calling `bootstrapApplication`, then await/register only the matched navigator locale before initial rendering—e.g. via [`provideAppInitializer`](https://angular.dev/api/core/provideAppInitializer) or by gating bootstrap on that import. Please add coverage that exercises the production registration path.

2. **The required settings documentation is missing.**

   The repository's [documentation guide](https://github.com/super-productivity/super-productivity/blob/014b789c22c9bf75fd7202845639569b61e7cd8e/docs/documentation-guide.md#L11-L15) requires changed user-facing settings behavior to update the wiki in the same PR. The existing [Datetime format locale entry](https://github.com/super-productivity/super-productivity/blob/014b789c22c9bf75fd7202845639569b61e7cd8e/docs/wiki/3.02-Settings-and-Preferences.md#L9-L13) does not say that “System default” follows the browser/OS regional locale or that unavailable locale data falls back to en-GB.

   **Smallest fix:** add one concise sentence to that existing bullet.

## Minor

- [`src/main.ts:388`](https://github.com/super-productivity/super-productivity/blob/e9e2949c5a885655d710136ec33217764544f9d8/src/main.ts#L384-L392) still imports/registers all eight navigator-only locale chunks during idle time for every user, including loading the selected region a second time. Only the matched `navigator.language` entry is needed.
- The branch contains the unrelated timezone-test commit from still-open [#9131](https://github.com/super-productivity/super-productivity/pull/9131). Please rebase/drop it after #9131 lands, or explicitly absorb and close that PR.

## Verification

- All current CI checks pass.
- Exact-head changed specs: 73/73 under `Australia/Sydney`.
- Related date-time service specs: 25/25.
- `git diff --check` passes.

The underlying diagnosis is valid, and the fallback-success discriminant, module-wide warning deduplication, and removal of user date content from logs are all good changes.

## Verdict

**Needs changes** — please fix bootstrap ordering and add the required documentation before merge.

KamilDev:
Both blockers and the loop minor are addressed in 4bdc17a05.

**Bootstrap ordering** — you're right, both registrations sat inside `bootstrapApplication(...).then(...)`. Registration now lives in a new `src/app/core/locale-registration.ts`:

- `registerDefaultLocale()` — called at module scope in `main.ts`, before `bootstrapApplication`, so the static en-GB data is in place before any render.
- `registerNavigatorLocale()` — awaited via `provideAppInitializer`, so the matched `navigator.language` regional chunk is registered before first render. It never rejects: a failed chunk load logs and falls back to default-locale rendering instead of blocking bootstrap.

Coverage for the production path: new `locale-registration.spec.ts` asserts `registerDefaultLocale` pins day-first/24h under the bare `'en'` id, and that `registerNavigatorLocale('en-AU')` flips `shortTime` from the en-GB 24h fallback to 12h (the spec registers the default first, matching prod ordering, so the assertion genuinely gates the await). `locale.constants.spec.ts` now also registers all 8 variants through `registerNavigatorLocale` rather than raw `registerLocaleData`, so the keyless self-id behavior is exercised via the same function main.ts uses. What remains untestable under Karma is only the literal presence of the two call sites in `main.ts`.

**Idle loop** — dropped entirely for the navigator-only variants; only the matched `navigator.language` entry is ever loaded, once, by the initializer.

**Docs** — added to the Datetime format locale bullet in `3.02-Settings-and-Preferences.md`: System default follows the browser/OS regional locale, and unavailable locale data falls back to en-GB formatting.

**#9131 commit** — will rebase it out of this branch as soon as #9131 lands (it's needed locally so the pre-push full-suite hook passes on a Windows machine outside Europe/Berlin). If you'd rather absorb it here and close #9131, also fine by me — your call.

Verification: new spec 3/3, `locale.constants` 10/10, pipe 14/14, `date-time-format` 25/25, full unit suite 13178/13178, `tsc --noEmit` on `src/tsconfig.app.json` clean, prettier/eslint clean on all touched files.

johannesjo (commented):
## Summary

Reviewed `e9e2949c..4bdc17a0` (the one new commit since my last review). **Both blockers from round 2 are fixed**, and I verified the ordering claim against Angular's own bootstrap path rather than taking the commit message's word for it. Nice work — the extraction into `locale-registration.ts` also made the thing testable, which was the part I doubted was possible.

One item remains before this can merge, and it's branch hygiene rather than code.

## Verified

- **Registration really does precede first render.** In Angular 21.2.17: `provideAppInitializer` → `{provide: APP_INITIALIZER, multi: true}` → `ApplicationInitStatus.runInitializers()` collects any returned promise into `Promise.all(...)` → `internalCreateApplication` gates on `initStatus.donePromise.then(() => …bootstrap root component…)`. So the awaited initializer genuinely blocks the first render, and [`registerDefaultLocale()` at `main.ts:129`](https://github.com/super-productivity/super-productivity/blob/4bdc17a05ae0b34924decd03c0095f8d1d886761/src/main.ts#L129) runs at module scope before `bootstrapApplication` on `:131`. `main.ts` is the only `bootstrapApplication` call site.
- **The wiki sentence matches the real UI label** — `TIME_LOCALE_AUTO` is literally "System default" (`en.json:2567`), so the docs and the dropdown agree.
- **The new specs are not vacuous.** I sabotaged each function separately: no-op `registerDefaultLocale` → `Expected '1/15/24' to be '15/01/2024'`; no-op `registerNavigatorLocale` with the baseline intact → `Expected '13:00' to match /^1:00/`.
  Worth recording for whoever touches this next: sabotaging *both* at once makes the en-AU assertion pass **vacuously** — with `en` unregistered, en-AU falls through to Angular's built-in en-US, which is already 12h. The `beforeAll(registerDefaultLocale)` in that suite is load-bearing, and the comment explaining why should survive future cleanups.
- 3/3, 10/10 and 14/14 specs green; `checkFile` clean on all three `.ts` files; production build exits 0.

## Before merge

**The branch still carries [`b360de1a`](https://github.com/super-productivity/super-productivity/commit/b360de1aa80d16b590b3a1a40497a926a574f855) from #9131, which is still open and not in `master`.** Merging this PR silently merges #9131's spec changes along with it, bypassing that PR's own review. Please rebase it out (or land #9131 first and rebase on top). This is the only thing actually blocking.

## Non-blocking — a consequence of my own "minor" from last round, so your call

Dropping the idle-time loop was my suggestion, and it had a side effect I should have thought through then: **this is now the first initializer in the app that makes bootstrap wait on a network fetch.** Every other `APP_INITIALIZER` in `main.ts` is synchronous or fire-and-forget — one of them carries the comment *"Fire-and-forget … so it can never block or fail bootstrap"* — and there are none registered anywhere else in `src/` (`PLUGIN_INITIALIZER_PROVIDER` uses a plain string token; `provideAnimationsAsync` registers no initializer, it swaps renderers lazily).

I went looking for how bad that actually is, and it's narrower than I first assumed:

- All 8 locale files emit as separate lazy chunks, and I confirmed from the generated `ngsw.json` that **8/8 land in the `app` asset group with `installMode: prefetch`** — so once the service worker is installed they're cached and no request happens at all.
- The exposure is therefore first load / service-worker-not-yet-controlling, and only for users whose `navigator.language` is one of the 8 `en-*` variants (`registerNavigatorLocale` returns before any `await` for everyone else). In that window the network is demonstrably alive — `main.js` just loaded — so the realistic cost is **one extra RTT for a ~1.1 KB chunk**. Electron and Capacitor load from local assets, so this is web-only.
- The tail risk is that `Promise.all` has no timeout: a connection that stalls mid-request during that window leaves the app on a white screen until the browser gives up.

Two ways to close the tail, neither of which I'd insist on:

1. `Promise.race` the import against a short timeout (~1–2s). Three lines, no bundle cost, and if the timeout wins you land on exactly today's `master` behaviour (late registration, en-GB fallback for already-rendered dates) — strictly no worse than before.
2. Statically import the 8 files and register synchronously. This deletes the initializer, the `async`, the `try/catch` and the failure mode entirely, and also removes the coupling noted below. **Cost: 5,478 B gzipped on a 244 KB gzipped main bundle, so about +2.2%** — real, not free, which is why I'm not pushing it.

If you'd rather keep it as-is, that's defensible; I'd just fix the comment in (1) below so the next reader isn't misled.

## Minor

1. [`main.ts:135`](https://github.com/super-productivity/super-productivity/blob/4bdc17a05ae0b34924decd03c0095f8d1d886761/src/main.ts#L135) — *"Never rejects, so it cannot block bootstrap"* conflates rejecting with blocking. Swallowing every error means it can't **fail** bootstrap; it very much does **block** it, which is the entire point of awaiting it. Suggest: *"Never rejects, so a failed chunk load degrades to the default locale instead of failing bootstrap."*
2. `registerNavigatorLocale()` defaults to `navigator.language`, but the locale the pipe actually asks for comes from `DateTimeFormatService.currentLocale()` → ngx-translate's `getBrowserCultureLang()`, which returns **`navigator.languages[0]`**. Those agree in practice, but now that the idle loop no longer registers all 8 as a safety net, the two have to agree exactly. Reading the same source in both places would make that robust by construction rather than by coincidence.
3. Coverage gap, and I don't think it's worth fixing — flagging it so it's a known gap rather than an assumed one: the new spec covers the two extracted functions, not the `main.ts` wiring. Moving `registerDefaultLocale()` back inside `.then()` would reintroduce the original bug without failing a single test.

## Verdict

Code-wise this is good to go — the ordering fix is correct and the tests genuinely gate it. **Rebase off #9131** and I'm happy to merge; the initializer discussion above is a judgement call I'm leaving to you.

KamilDev:
Thanks — all three round-3 items are in, plus the rebase. New head is `7367699a5`.

## Blocker: rebased

#9131 landed as `775d202fd`, so the branch is rebased onto `upstream/master` (`f91f958a6`) and `b360de1a` dropped out by patch-id. The diff is now three commits that are all this PR's own work; nothing from #9131 rides along.

## Minor 1 — comment

Reworded, and I extended it a bit to cover the timeout below:

```ts
// Await the browser's own regional locale (en-AU, en-CA, … — navigator-only
// variants backing "System default") before first render, for the same
// pure-pipe reason as above. Never rejects and self-limits to a short
// timeout, so a failed or stalled chunk load degrades to the default locale
// instead of failing bootstrap or holding up first render indefinitely.
```

## Minor 2 — one source for "the browser's locale"

You were right that the two only agreed by coincidence once the idle loop went away. Rather than re-derive `navigator.languages[0]` in a second place, the default argument now calls the same thing `DateTimeFormatService` does — ngx-translate's static `getBrowserCultureLang()`, which needs no DI:

```ts
export const registerNavigatorLocale = async (
  navLanguage: string | undefined = TranslateService.getBrowserCultureLang(),
): Promise<void> => {
```

So it is the same call, not a matching reimplementation of it. The parameter widened to `string | undefined` because `getBrowserCultureLang()` can return `undefined` (no `window.navigator`, or an empty `navigator.languages`); that case takes the existing early return.

## Judgement call — took option 1

`Promise.race` against a 1.5s timeout. I went with this over the static import because the +2.2% main bundle is paid by every user on every load to remove a tail risk that, by your own measurement, only exists on first load for a subset of `en-*` users — and on timeout this lands on exactly today's `master` behavior, so it is strictly no worse than the status quo it replaces.

```ts
/**
 * Upper bound on how long bootstrap waits for the regional locale chunk. On
 * timeout we render with the default locale — exactly the behavior before this
 * registration existed — rather than hold up first render on a stalled network.
 */
const LOCALE_LOAD_TIMEOUT_MS = 1500;
```

```ts
  let timeoutId: ReturnType<typeof setTimeout> | undefined;
  try {
    // Promise.race consumes a late rejection from load(), so a chunk that fails
    // after the timeout won this race cannot surface as an unhandled rejection.
    const m = await Promise.race([
      load(),
      new Promise<null>((resolve) => {
        timeoutId = setTimeout(() => resolve(null), LOCALE_LOAD_TIMEOUT_MS);
      }),
    ]);
    if (!m) {
      Log.err(`Timed out loading locale ${key}`);
      return;
    }
    registerLocaleData(m.default);
  } catch (e) {
    Log.err(`Failed to load locale ${key}`, e);
  } finally {
    clearTimeout(timeoutId);
  }
```

Two details worth naming since they are the kind of thing that rots quietly: the losing `load()` promise stays attached to the race, so a chunk that rejects after the timeout is consumed rather than becoming an unhandled rejection; and a late-arriving chunk is deliberately *not* registered, because by then the pure pipe has already cached its resolution and registering would only add a state divergence with no visible effect.

## Coverage

Three specs added, all in `locale-registration.spec.ts` (now 6/6):

- **the default reads `getBrowserCultureLang`** — spied to return `en-NZ`, then asserts `en-NZ` renders 12h. Reverting the default to `navigator.language` fails it, since the spy is then never consulted and `en-NZ` never gets registered.
- **no culture language** — spy returns `undefined`, asserts it still resolves.
- **stalled chunk load** — the `en_ie` entry is swapped for a never-resolving import under `jasmine.clock()`; after ticking 1500ms the call resolves and `en-IE` still renders the en-GB 24h fallback, proving the load was abandoned rather than awaited. Without the race this spec hangs instead of failing, which is the honest signal for "bootstrap would have waited forever".

Your two notes from round 3 are preserved rather than silently dropped: the `beforeAll(registerDefaultLocale)` comment explaining why it is load-bearing is untouched, and the `main.ts`-wiring coverage gap is still a gap — I agree it is not worth manufacturing a bootstrap harness for.

Two doc comments also now say "browser culture language" instead of "`navigator.language`", so the constants file no longer describes the old source.

## Verification

- Full suite green on this branch: **13313/13313** on both Karma runs `npm run test` performs. Pushed with hooks enabled (`npm run lint && npm run test`), no `--no-verify`.
- `npm run checkFile` clean on all four touched files.

johannesjo (changes requested):
Round 4 @ `7367699a5c`. All three round-3 items are genuinely in, and CI is green on the head sha (13 success / 4 skipped). The production code I'm happy with. The tests are the problem: I sabotage-tested every new spec, and **most of the coverage for the headline half of this PR does not fail when the fix is reverted**.

## Verified fixed

- **Rebase** — `b360de1a` is gone; the branch is `f91f958a6c` + this PR's own commits, nothing from #9131 riding along.
- **One locale source** — `TranslateService.getBrowserCultureLang()` is genuinely `static` (ngx-translate 17.0.0, `ngx-translate-core.mjs:782`), and the instance method `DateTimeFormatService` calls at `date-time-format.service.ts:116` just delegates to it (`:795`). So it is the same call, as you said, not a lookalike.
- **`Promise.race`** — the late-rejection comment is correct (the reaction is attached when the race is created, so a post-timeout rejection is consumed), and `clearTimeout` is reached on every path that creates a timer. Option 1 was yours to take and the reasoning holds.

## Blocking: the new specs pass with the fix removed

### 1. `locale.constants.spec.ts` — 6 of the 8 per-locale specs are vacuous, in every suite order

Sabotage: make the map lookup in `registerNavigatorLocale` unable to hit, so registration is a total no-op.

```
TOTAL: 2 FAILED, 8 SUCCESS
  ✗ renders shortTime in 24h for en-za   Expected '1:00 PM' to match /^13:00/.
  ✗ renders shortTime in 24h for en-ie   Expected '1:00 PM' to match /^13:00/.
```

en-AU, en-CA, en-IN, en-NZ, en-PH and en-SG all pass with **nothing registered at all**. An unregistered `en-XX` resolves through Angular's built-in `en`, which is en-US — already 12h, already `/^1:00/`. So the six specs covering exactly the locales this fix exists for assert something that is true without the fix, and the only two that can fail are the two the fix doesn't need to change.

This is not an isolated-run artifact. Both `locale-date.pipe.spec.ts` and `locale-registration.spec.ts` restore `'en'` to the baked-in en-US in their `afterAll`, so `locale.constants.spec.ts` sees a 12h `en` baseline whichever order it runs in — I get the same `4 FAILED, 12 SUCCESS` (the same 6 staying green) running the pair in forced order and in random order.

One line fixes it: call `registerDefaultLocale()` in that `beforeAll` before the `Promise.all`, the way `locale-registration.spec.ts:37` already does. Then the baseline is en-GB/24h — matching prod — and "renders 12h" genuinely means "the regional data won".

### 2. The global registry leak makes two more specs ~50/50

`locale.constants.spec.ts:30-38` registers all eight variants into Angular's process-global `LOCALE_DATA` and has no `afterAll`. Jasmine order is randomised here (no `random: false` in `src/karma.conf.js`), so when that suite runs first, `locale-registration.spec.ts` is asserting on data another suite registered.

Sabotage = the exact regression the spec claims to pin, default arg back to `navigator.language`:

```
run 1  1 FAILED   run 2  1 FAILED   run 3  16 SUCCESS   run 4  16 SUCCESS
run 5  1 FAILED   run 6  16 SUCCESS run 7  1 FAILED     run 8  16 SUCCESS
```

5 of 8 — so "Reverting the default to `navigator.language` fails it" is true about half the time. Sabotage = drop `.toLowerCase()` from the key derivation (`locale-registration.ts:47`): 3 FAILED / 3 SUCCESS over 6 runs, taking both the en-AU and the en-NZ spec down together. Across 12 sabotage runs the pass/fail split correlated 12/12 with which suite Jasmine happened to run first.

(Worth noting *why* the `.toLowerCase()` sabotage doesn't also break `locale.constants.spec.ts`: that file passes an already-lowercase key, so it keeps registering and keeps masking.)

The robust shape is to assert on something *this call* registered that no other suite can have supplied — e.g. inject a synthetic loader under a key only this spec uses. You already do the map-swap trick in the stalled-load spec.

### 3. The stalled-load spec's assertion is vacuous, and its comment is wrong

`locale-registration.spec.ts:63` says:

```ts
// en-IE is 12h; asserting the en-GB 24h fallback proves the stalled load
```

en-IE is 24h. `locale.constants.spec.ts:22` says so itself (`en_ie: false`), and the data agrees — en-GB and en-IE have identical `timeFormats`, `["HH:mm","HH:mm:ss","HH:mm:ss z","HH:mm:ss zzzz"]`. So `toBe('13:00')` cannot distinguish "abandoned the stalled load" from "loaded and registered en-IE".

Two direct checks, both green:

- swap the never-resolving loader for the real `origLoad`, i.e. no stall whatsoever → `TOTAL: 6 SUCCESS`
- go further and force success, `NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS['en_ie'] = () => Promise.resolve({default: localeEnIE})` → still 16 SUCCESS in both suite orders

The only load-bearing part left is the implicit "did not hang". If you want to keep en-IE, `fullDate` is the one width that actually discriminates: unregistered renders `Monday, 15 January 2024` and registered en-IE renders `Monday 15 January 2024` (en-IE's `EEEE d MMMM y` has no comma), while `shortTime` stays `13:00` either way.

### 3b. And that spec's `finally` is skipped exactly when it matters

If the timeout regresses (race → bare `await load()`), `await expectAsync(pending).toBeResolved()` never settles, so the `finally` never runs. It doesn't hang the run — `src/test.ts` sets `jasmine.DEFAULT_TIMEOUT_INTERVAL = 2000` — but it leaves `jasmine.clock()` installed and `NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS['en_ie']` mutated for the remainder of the Karma run. Measured, with `locale-registration` first: `Executed 16 of 16 (11 FAILED)` — the one real failure plus all 10 specs of the sibling suite reporting *"Not run because a beforeAll function failed"*, because that `beforeAll` awaits the never-resolving loader. A probe spec (`setTimeout(done, 1)`) inserted between the two files fails under the sabotage and passes without it, confirming the clock leaks. Restoring in a `beforeEach`/`afterEach` pair instead of `finally` would contain it.

## Non-blocking

- **Allowlist coverage.** 61 of the 115 `en-XX` files `@angular/common` ships differ from en-GB on short date or short time and aren't in the map — en-AE, en-HK, en-MY and en-PK are all `h:mm a`, i.e. the identical 12h/24h mismatch this PR exists to fix. Not an argument to grow the list, just worth knowing the fix is a subset. (If anyone does write assertions against these: CLDR uses U+202F before the day period, so `'h:mm a'` string equality silently fails. Your `/^1:00/` regex sidesteps it.)
- **Worth checking before merge: `en-US`.** It isn't in `NAVIGATOR_FALLBACK_LOCALE_IMPORT_FNS`; it's registered only by the pre-existing idle loop (`main.ts:407`), after first render. Separately, this PR moves `registerLocaleData(DEFAULT_LOCALE_DATA, 'en')` from inside the bootstrap `.then` (master `main.ts:357`) to module scope, which changes what an `'en-US'` lookup resolves to at first render: pristine gives `1:00 PM` / `1/15/24`, after `registerDefaultLocale()` it gives `13:00` / `15/01/2024`. **I did not verify the gate** — whether `DateTimeFormatService`'s effect has already pushed `'en-us'` into `currentLocale()` by the first date render, or whether that render still uses the `'en-gb'` seed (in which case nothing changes). If the former, en-US browsers get day-first/24h at first paint, frozen there by the pure pipe until an input changes. Cheap to settle and worth settling, since it's the largest locale group.
- **The inner `catch {}` now hides developer errors.** Angular throws `RuntimeError 2300` from `assertValidDateFormatLength` when `format.length > 256`, and from `assertValidDateFormat` on a suspicious `Y` pattern in dev. Both are locale-independent, so the fallback pipe throws identically and `locale-date.pipe.ts:56-61` returns `null` with no log — where master's single catch did `Log.warn(..., value, e)`. Narrow (only the >256 check survives a prod build) but avoidable: logging the *format string* and `e.name` stays inside rule 9, since neither is user content.
- **PR description is stale** — it still says `main.ts` registers the variants "in `registerRemainingLocales` (idle-time)", which was rounds 1-2.
- Negative result, since it's the obvious next worry: no other spec in the repo registers `th`, `el` or `da` locale data, so the three new pipe specs' locale choices are safe from the same cross-suite collision.

KamilDev:
Thanks for sabotage-testing every spec — you were right, most of the registration-half coverage passed with the fix reverted. New head `00a83ae29`. I re-ran each of your sabotages against the new specs to confirm they now fail.

## Blocking: specs now fail when the fix is reverted

**1. `locale.constants.spec.ts` — the 6 vacuous 12h specs.** `registerDefaultLocale()` now runs in `beforeAll` before the `Promise.all` (your one-line fix), so the baseline under bare `en` is en-GB/24h — an unregistered `en-*` no longer falls through to the built-in en-US that made "renders 12h" true for free. Added an `afterAll` restoring `en` to en-US for cross-suite hygiene.
- Sabotage (registration no-op): **6 FAILED** — en-AU/CA/IN/NZ/PH/SG, the exact six the fix exists for. Previously only the two 24h locales could fail.

**2. The global-registry leak.** The real `en-*` variants leak out of `locale.constants.spec` into the module-global registry, so asserting on en-AU/en-NZ passed on that leak whenever Jasmine ran that suite first. Rather than assert on real variants, each spec now registers a **synthetic 12h locale under an id no other suite touches** (`en-XA`, `en-XB` — real en-AU data re-tagged), so a 12h render can only mean *this* call registered it. The import-map mutation is cleaned in `afterEach`.
- Sabotage (drop `.toLowerCase()`): **both** synthetic specs fail.
- Sabotage (default arg → `navigator.language`): the `getBrowserCultureLang` spec fails — and now **deterministically**, not ~50% by suite order, because `en-XB` is never anything else's registration.

**3 + 3b. The stalled-load spec.** You're right that en-IE is 24h — its data is byte-identical to en-GB, so `toBe('13:00')` couldn't tell "abandoned" from "loaded" (and the comment was wrong). Replaced with a **synthetic 12h loader that never resolves**: giving up leaves the tag unregistered → it renders the en-GB 24h fallback, which discriminates; awaiting the stall would hang. And `jasmine.clock()` install/uninstall moved into `beforeEach`/`afterEach`, not a `finally` — so a timeout regression whose assertion never settles can't leak the mocked clock into sibling suites.
- Sabotage (`Promise.race` → bare `await load()`), both suites together: **1 FAILED, no cascade** (afterEach still runs on the spec timeout). Your `finally` version measured 11.

## Non-blocking

**en-US gate — fixed.** I chased it down and it's a real regression, so I closed it rather than leave it. It bites the explicit case deterministically: with `dateTimeLocale = en-US` (or a post-effect en-US "System default"), `currentLocale()` is `en-us` from the first render, and seeding en-GB under `en` at module scope means `en-us` now falls through to en-GB → day-first/24h at first paint, frozen by the pure pipe. On master the same lookup fell through to Angular's built-in en-US and was correct. Fix in `registerDefaultLocale`: `registerLocaleData(localeEnUs, 'en-us')` alongside the en-GB seed (the data self-reports `en`, so the explicit id is required — keyless would clobber the seed). New spec pins `en-us` → `1/15/24` / `1:00` and the untouched en-GB `en` seed; sabotage (register under a bogus id) fails it.

**Inner `catch {}` — I looked at logging it and decided against, reasoning below (happy to add it if you'd still prefer).** The inner catch is reached only when the en-GB fallback *also* throws, and at that point a bad user value and a genuine dev error (RuntimeError 2300 on a >256-char/suspicious format) are indistinguishable — Angular rewraps both as NG02100. The existing spec ("should not warn for an unformattable value nor let it poison the per-locale warn dedup") deliberately keeps this path silent to match `safeFormatDate`, so logging here would fire `Log.warn` on every unformattable *value* too, not just dev errors — re-opening that decision. Since only the >256 check survives a prod build, the dev-error surface is narrow, and I didn't think it worth trading the deliberate silence for. Your call.

Also refreshed the PR description — the `main.ts` "registered lazily in `registerRemainingLocales` (idle-time)" line was stale (rounds 1–2); it's now module-scope + app-initializer before first render, and the new `locale-registration.ts` / spec files are listed.

johannesjo:
Fresh re-check at `00a83ae2`: the earlier blockers are genuinely resolved, the current implementation handles loader rejection correctly, all 31 changed specs pass, and CI is green. I found one remaining test gap before merge.

## Important — rejecting locale chunks are not covered

[`locale-registration.spec.ts:84`](https://github.com/super-productivity/super-productivity/blob/00a83ae293c05057c7674ec574a3f64bd8c040c9/src/app/core/locale-registration.spec.ts#L84-L86) uses `th-TH`, which is outside the loader map, so `registerNavigatorLocale()` returns at [`locale-registration.ts:62`](https://github.com/super-productivity/super-productivity/blob/00a83ae293c05057c7674ec574a3f64bd8c040c9/src/app/core/locale-registration.ts#L61-L64) before calling `load()` or exercising the catch.

That leaves the documented never-reject contract unprotected: in a local sabotage check, changing the catch to rethrow still left all 31 changed specs green. Because this promise is returned from [`provideAppInitializer`](https://github.com/super-productivity/super-productivity/blob/00a83ae293c05057c7674ec574a3f64bd8c040c9/src/main.ts#L133-L138), a future uncaught chunk rejection would reject Angular bootstrap and prevent first render instead of degrading to en-GB.

Please add a synthetic matched loader that returns `Promise.reject(...)` and assert that `registerNavigatorLocale(...)` resolves, using the existing import-map cleanup. I verified that test shape locally: it passes against the current code and fails exactly once when the catch rethrows.

**Verdict: Needs changes (test-only).** I found no other current-head issues.

KamilDev:
Added at `37e118b37`. Test-only, no production change.

## Rejecting chunk load — covered

You were right that the catch was unreachable from the specs: `th-TH` misses the map, so `registerNavigatorLocale` returns at the `if (!load)` guard and `load()` is never called.

New spec registers a synthetic loader under a matched key (`en_xd`) that returns `Promise.reject(new Error('chunk load failed'))`, then asserts:

1. the call resolves — the contract the `provideAppInitializer` await depends on;
2. `formatDate(onePm, 'shortTime', 'en-xd')` is `13:00`, i.e. the failed locale stays unregistered and falls through to the en-GB seed rather than landing half-applied.

Cleanup goes through the existing `syntheticMapKeys` / `afterEach` path, so the import map is restored like the other synthetic specs.

## Sabotage

`throw e;` appended to the catch in `locale-registration.ts:83`:

```
1) resolves without throwing when the matched chunk load rejects
   Expected a promise to be resolved but it was rejected with Error: chunk load failed.
TOTAL: 1 FAILED, 7 SUCCESS
```

Exactly one failure, no cascade — matching the shape you verified locally. Reverted, file green at 8/8.

Full suite green (13315), `checkFile` clean on the spec.

johannesjo:
Thank you very much! <3

PR #9131, opened 2026-07-18
test: make date-based specs timezone-independent

KamilDev:
## Problem

8 unit tests fail on a clean master checkout in timezones far enough ahead of UTC (e.g. `Australia/Sydney`, UTC+11). Reproducible on any OS with `npm run test:tz:sydney`.

- `create-schedule-days.spec.ts` (4 failures): the spec derives its day strings via `today.toISOString().split('T')[0]` — that's the **UTC** date, while `createScheduleDays` buckets days by **local** midnight. In Sydney, local `Jan 20 10:00` is UTC `Jan 19 23:00`, so `todayStr` lags a day, the week-boundary assertions flip, and one test crashes with `TypeError: Cannot read properties of undefined (reading 'length')` in `getTimeLeftForTask` — the shifted week boundary lets its raw task literal (built without `subTaskIds` instead of via `createTestTask`) reach the budget calculation.
- `plugin-i18n-date.util.spec.ts` (4 failures): the spec pins the UTC instant `2026-01-16T14:30:00Z` but `formatDateForPlugin` formats in the local timezone, so in Sydney the output is `Jan 17` and the `toContain('16')` assertions fail.

### Why this hits Windows contributors even with the `TZ=Europe/Berlin` pinning

The `cross-env TZ='Europe/Berlin'` in the test scripts only reaches the Node/Karma process. Chromium on Windows ignores the `TZ` environment variable and uses the OS timezone, so the browser the specs actually run in uses the machine's local timezone (verified with `Intl.DateTimeFormat().resolvedOptions().timeZone` inside a Karma spec). Any Windows contributor east of ~UTC+10 gets these 8 failures on untouched master.

## Fix

Make the specs timezone-independent, matching the intent of the `test:tz:*` scripts:

- derive day strings with `getDbDateStr()` (local date) instead of `toISOString()`
- construct the plugin spec's test date with the local-time `Date` constructor and derive the ISO-string input from it
- build the raw task literal with the file's `createTestTask` helper so the task shape matches the `Task` model (`subTaskIds` present)

~~The derived values are byte-identical in the timezones CI currently runs (`Europe/Berlin`, `America/Los_Angeles`, and the other `test:tz` targets `Asia/Tokyo`/`UTC`), so this changes nothing for existing green runs — it only removes the UTC/local mismatch.~~

**Correction** (from review — thanks @johannesjo for catching this): the byte-identical claim doesn't hold for `secondDayStr` in `create-schedule-days.spec.ts`. The old line went `parseDbDateStr` (local midnight) → `+1 day` → `.toISOString()` (UTC), and local midnight rolls back into the previous UTC day in any positive-offset timezone — including CI's primary `Europe/Berlin` and `Asia/Tokyo`. On master, `should filter tasks between days when viewing outside current week` was therefore arranging `dayDates` as the **same day twice** in those timezones instead of two consecutive days; only `America/Los_Angeles` and `UTC` produced the intended arrangement. This PR makes the arrangement correct in all timezones. No assertion outcome changes (the test only asserts on `result[0]`, which passed either way), but the description's "changes nothing for existing green runs" was wrong — it repairs a mis-arranged test in the timezones CI actually runs.

## Verification

- `npm run test:file src/app/features/schedule/map-schedule-data/create-schedule-days.spec.ts` → 18/18 (was 4 FAILED) on a Windows machine in `Australia/Sydney`
- `npm run test:file src/app/plugins/plugin-i18n-date.util.spec.ts` → 31/31 (was 4 FAILED)
- full `npm run lint && npm run test` (pre-push hook) green: 13162 SUCCESS

### DIFF
--- a/src/app/features/schedule/map-schedule-data/create-schedule-days.spec.ts
+++ b/src/app/features/schedule/map-schedule-data/create-schedule-days.spec.ts
@@ -43,19 +43,23 @@ describe('createScheduleDays - Task Filtering', () => {
     now = today.getTime();
     realNow = now;
 
-    todayStr = today.toISOString().split('T')[0];
+    // NOTE: date strings must be derived from the LOCAL date (getDbDateStr), not
+    // toISOString() (UTC) — otherwise they shift a day in timezones ahead of UTC
+    // (e.g. Australia/Sydney) and no longer match the local-midnight logic in
+    // createScheduleDays.
+    todayStr = getDbDateStr(today);
 
     const tomorrow = new Date(today);
     tomorrow.setDate(tomorrow.getDate() + 1);
-    tomorrowStr = tomorrow.toISOString().split('T')[0];
+    tomorrowStr = getDbDateStr(tomorrow);
 
     const nextWeek = new Date(today);
     nextWeek.setDate(nextWeek.getDate() + 7);
-    nextWeekStr = nextWeek.toISOString().split('T')[0];
+    nextWeekStr = getDbDateStr(nextWeek);
 
     const futureWeek = new Date(today);
     futureWeek.setDate(futureWeek.getDate() + 14);
-    futureWeekStr = futureWeek.toISOString().split('T')[0];
+    futureWeekStr = getDbDateStr(futureWeek);
   });
 
   describe('Unscheduled tasks (no dueDay, no dueWithTime, no plannedForDay)', () => {
@@ -89,12 +93,7 @@ describe('createScheduleDays - Task Filtering', () => {
 
     it('should NOT appear when viewing next week (outside current week)', () => {
       // Arrange
-      const unscheduledTask: TaskWithoutReminder = {
-        id: 'task1',
-        title: 'Unscheduled Task',
-        timeEstimate: 3600000,
-        timeSpent: 0,
-      } as TaskWithoutReminder;
+      const unscheduledTask = createTestTask('task1', 'Unscheduled Task');
 
       // Viewing a week starting 7 days from now
       const dayDates = [nextWeekStr];
@@ -519,7 +518,7 @@ describe('createScheduleDays - Task Filtering', () => {
       // Viewing two days in next week
       const secondDayNextWeek = parseDbDateStr(nextWeekStr);
       secondDayNextWeek.setDate(secondDayNextWeek.getDate() + 1);
-      const secondDayStr = secondDayNextWeek.toISOString().split('T')[0];
+      const secondDayStr = getDbDateStr(secondDayNextWeek);
 
       const dayDates = [nextWeekStr, secondDayStr];
       const plannerDayMap: PlannerDayMap = {};
--- a/src/app/plugins/plugin-i18n-date.util.spec.ts
+++ b/src/app/plugins/plugin-i18n-date.util.spec.ts
@@ -1,9 +1,12 @@
 import { formatDateForPlugin } from './plugin-i18n-date.util';
 
 describe('formatDateForPlugin', () => {
-  const testDate = new Date('2026-01-16T14:30:00Z');
+  // NOTE: constructed in LOCAL time — formatDateForPlugin formats in the local
+  // timezone, so a fixed UTC instant would land on Jan 17 in timezones ahead of
+  // UTC (e.g. Australia/Sydney) and break the day-of-month assertions below.
+  const testDate = new Date(2026, 0, 16, 14, 30, 0);
   const testTimestamp = testDate.getTime();
-  const testISOString = '2026-01-16T14:30:00Z';
+  const testISOString = testDate.toISOString();
 
   describe('short format', () => {
     it('should format date with short format in English', () => {


johannesjo (commented):
Reviewed and verified locally. **This is a good change — I'd merge it.** Test-only, correctly diagnosed, and it fixes slightly more than it claims. One correction to the description below, plus a couple of non-blocking notes.

## Verification

| Check | Result |
| --- | --- |
| Reproduce on master, `TZ=Australia/Sydney` (the 2 changed specs) | **8 FAILED, 41 SUCCESS** — exactly as described, including the `TypeError` at `get-time-left-for-task.ts:4` |
| PR applied, the 2 changed specs × 5 TZs (Sydney / Berlin / LA / UTC / Tokyo) | **49/49 SUCCESS** in all five |
| PR applied, **full unit suite** in Sydney | **13162 of 13178 SUCCESS (16 skipped), 0 failures** — reproduced across 3 independent runs |
| `npm run checkFile` on both files | clean |

Since the full Sydney suite is green with the PR and the PR touches only these two files, the "8 failures" count in the description is confirmed repo-wide — those 8 were the only Sydney failures on master.

The fix is also idiomatic rather than novel: `parseDbDateStr` / `getDbDateStr` exist precisely for this, and their doc comments call out this exact UTC-vs-local hazard.

## Correction: the "byte-identical" claim doesn't hold for `Europe/Berlin`

The description says the derived values are *"byte-identical in the timezones CI currently runs … so this changes nothing for existing green runs."* That's true for everything **except `secondDayStr`**, and that exception lands in CI's primary timezone.

I proved it by temporarily adding this assertion next to `const dayDates = [nextWeekStr, secondDayStr]` and running the spec unmodified otherwise:

```ts
expect(`${dayDates[0]} vs ${dayDates[1]}`).not.toBe(`${dayDates[0]} vs ${dayDates[0]}`);
```

On **master**:

```
Europe/Berlin        FAILED — '2026-01-27 vs 2026-01-27'
Asia/Tokyo           FAILED — '2026-01-27 vs 2026-01-27'
Australia/Sydney     FAILED — '2026-01-26 vs 2026-01-26'
America/Los_Angeles  pass
UTC                  pass
```

With **this PR applied**: passes in all five.

The old line did `parseDbDateStr(...)` (local midnight) → `+1 day` → `.toISOString()` (UTC), and local midnight can only roll *backwards* into the previous UTC day — which requires a positive offset. So LA (negative) and UTC never expose it, and the LA-only `test:tz:ci` pass structurally could not have caught this one.

The practical consequence on master: `should filter tasks between days when viewing outside current week` was passing `dayDates = ['2026-01-27', '2026-01-27']` in Berlin — viewing **the same day twice**, not the two consecutive days its name and comment describe.

So the PR is *better* than advertised: it repairs a test that was arranged wrong in the timezone CI actually runs in. Worth correcting the description, though, since a reviewer trusting the stated "no-op in CI timezones" would draw the wrong conclusion about what's changing.

Two things that make this clearly a fix and not a behavior change worth worrying about:

- The file already used the correct idiom ~300 lines earlier (`dayBeforeFutureWeek`, line 221-223 on master: `parseDbDateStr` → `getDbDateStr`). The changed line was the one place that deviated from the file's own convention.
- That test's only real assertion is on `result[0]`, which passed either way — the trailing comment even concedes the second-day behavior is unasserted. So no coverage changes hands; the arrangement is simply correct now. (The weak assertion is pre-existing and out of scope here.)

## Non-blocking notes

**1. Six same-class occurrences remain in specs** — `task-repeat-cfg.service.spec.ts` (3), `schedule-month.component.spec.ts` (2), `schedule.component.spec.ts` (1). I ran all three in Sydney: **153/153 SUCCESS**, so they're latent rather than broken (self-consistent, or the offset isn't large enough to shift them). Fine to leave for a follow-up.

**2. If you want a regression guard, scope it to specs.** `test:tz:ci` is LA-only, and as shown above LA cannot catch the local-midnight variant of this bug, so this can silently regress. A local ESLint rule flagging `.toISOString().split('T')[0]` would be cheap — but it **must be spec-scoped**, because the UTC form is used deliberately in production: `src/app/op-log/apply/operation-converter.util.ts:30` (`getDeterministicLegacyDay`) uses `.toISOString().slice(0, 10)` on purpose, documented as a deterministic replay fallback. A blanket ban would be wrong. Adding Sydney to CI would also catch it, at roughly +50% unit-test time — probably not worth it next to a lint rule.

**3. One leftover in `plugin-i18n-date.util.spec.ts`** (pre-existing, not introduced here): `new Date('2024-02-29T12:00:00Z')` asserted with `toContain('29')` / `'February'` formats as `March 1, 2024` at UTC+13/+14 (`Pacific/Apia`, `Pacific/Kiritimati`). Outside every `test:tz:*` target, so it doesn't affect this PR.

## Stacking

PRs #9134, #9135 and #9137 all contain this exact commit (`b360de1` verified as an ancestor of each), so they're stacked on this one — no conflict, and merging this first is the right order.

## Note for maintainers

All CI workflows on this PR are `action_required` (first-time contributor), so nothing has actually run on GitHub's side. The verification above is local. Worth approving the workflow runs before merge.

---

*Note: I did not verify the Chromium-on-Windows-ignores-`TZ` rationale — I'm on Linux. It doesn't affect the outcome, since `npm run test:tz:sydney` reproduces all 8 failures on Linux too, which is a sufficient reason to make this change on its own.*

PR #9132, opened 2026-07-18
fix(tools): make checkFile work on Windows

KamilDev:
## Problem

`npm run checkFile <file>` fails on Windows before running anything:

```
🎨 Formatting create-schedule-days.spec.ts...
❌ Errors found:
spawnSync npm ENOENT
```

`tools/check-file.js` spawns `npm`/`npx` via `execFileSync`. On Windows those are `.cmd` shims, and Node refuses to spawn `.cmd` files without a shell since the CVE-2024-27980 fix — so the tool dies on every invocation, on every Windows machine.

## Fix

Use `execSync` with a quoted command string. The double-quoted absolute path (spaces in Windows user paths are common) is handled identically by cmd.exe and sh, so Linux/macOS behavior is unchanged; error reporting still uses `error.stdout`/`error.stderr`, which `execSync` populates the same way.

## Verification (Windows 11, Node 24)

- `npm run checkFile src/app/util/get-db-date-str.ts` → formats, lints, `All checks passed!`
- `npm run checkFile src/styles.scss` → stylelint path works too
- `npm run lint` green

Note: pushed with the pre-push hook bypassed — the hook's full unit-test run doesn't pass on this machine on clean master for an unrelated reason (timezone-dependent specs, fixed in #9131); this change is a dev-tool script that the unit suite never loads.

### DIFF
--- a/src/app/features/schedule/map-schedule-data/create-schedule-days.spec.ts
+++ b/src/app/features/schedule/map-schedule-data/create-schedule-days.spec.ts
@@ -43,19 +43,23 @@ describe('createScheduleDays - Task Filtering', () => {
     now = today.getTime();
     realNow = now;
 
-    todayStr = today.toISOString().split('T')[0];
+    // NOTE: date strings must be derived from the LOCAL date (getDbDateStr), not
+    // toISOString() (UTC) — otherwise they shift a day in timezones ahead of UTC
+    // (e.g. Australia/Sydney) and no longer match the local-midnight logic in
+    // createScheduleDays.
+    todayStr = getDbDateStr(today);
 
     const tomorrow = new Date(today);
     tomorrow.setDate(tomorrow.getDate() + 1);
-    tomorrowStr = tomorrow.toISOString().split('T')[0];
+    tomorrowStr = getDbDateStr(tomorrow);
 
     const nextWeek = new Date(today);
     nextWeek.setDate(nextWeek.getDate() + 7);
-    nextWeekStr = nextWeek.toISOString().split('T')[0];
+    nextWeekStr = getDbDateStr(nextWeek);
 
     const futureWeek = new Date(today);
     futureWeek.setDate(futureWeek.getDate() + 14);
-    futureWeekStr = futureWeek.toISOString().split('T')[0];
+    futureWeekStr = getDbDateStr(futureWeek);
   });
 
   describe('Unscheduled tasks (no dueDay, no dueWithTime, no plannedForDay)', () => {
@@ -89,12 +93,7 @@ describe('createScheduleDays - Task Filtering', () => {
 
     it('should NOT appear when viewing next week (outside current week)', () => {
       // Arrange
-      const unscheduledTask: TaskWithoutReminder = {
-        id: 'task1',
-        title: 'Unscheduled Task',
-        timeEstimate: 3600000,
-        timeSpent: 0,
-      } as TaskWithoutReminder;
+      const unscheduledTask = createTestTask('task1', 'Unscheduled Task');
 
       // Viewing a week starting 7 days from now
       const dayDates = [nextWeekStr];
@@ -519,7 +518,7 @@ describe('createScheduleDays - Task Filtering', () => {
       // Viewing two days in next week
       const secondDayNextWeek = parseDbDateStr(nextWeekStr);
       secondDayNextWeek.setDate(secondDayNextWeek.getDate() + 1);
-      const secondDayStr = secondDayNextWeek.toISOString().split('T')[0];
+      const secondDayStr = getDbDateStr(secondDayNextWeek);
 
       const dayDates = [nextWeekStr, secondDayStr];
       const plannerDayMap: PlannerDayMap = {};
--- a/src/app/plugins/plugin-i18n-date.util.spec.ts
+++ b/src/app/plugins/plugin-i18n-date.util.spec.ts
@@ -1,9 +1,12 @@
 import { formatDateForPlugin } from './plugin-i18n-date.util';
 
 describe('formatDateForPlugin', () => {
-  const testDate = new Date('2026-01-16T14:30:00Z');
+  // NOTE: constructed in LOCAL time — formatDateForPlugin formats in the local
+  // timezone, so a fixed UTC instant would land on Jan 17 in timezones ahead of
+  // UTC (e.g. Australia/Sydney) and break the day-of-month assertions below.
+  const testDate = new Date(2026, 0, 16, 14, 30, 0);
   const testTimestamp = testDate.getTime();
-  const testISOString = '2026-01-16T14:30:00Z';
+  const testISOString = testDate.toISOString();
 
   describe('short format', () => {
     it('should format date with short format in English', () => {
--- a/tools/check-file.js
+++ b/tools/check-file.js
@@ -1,5 +1,8 @@
 #!/usr/bin/env node
-const { execFileSync } = require('child_process');
+// NOTE: execSync with a quoted command string instead of execFileSync('npm', ...):
+// on Windows npm/npx are .cmd shims, which Node refuses to spawn without a shell
+// (EINVAL/ENOENT since the CVE-2024-27980 fix), so execFileSync breaks there.
+const { execSync } = require('child_process');
 const path = require('path');
 
 const file = process.argv[2];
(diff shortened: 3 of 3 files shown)


johannesjo:
Thanks for tracking this down — the diagnosis is exactly right and the bug is worth fixing. `execFileSync('npm', …)` really is unusable on Windows (`npm` is only `npm.cmd`, and naming the `.cmd` explicitly then trips the CVE-2024-27980 `EINVAL` guard), so `npm run checkFile` is 100% broken there today. That's a real gap given AGENTS.md asks for `checkFile` on every touched file.

One problem with the implementation, though.

## The quoting claim doesn't hold on POSIX

> The double-quoted absolute path … is handled identically by cmd.exe and sh, so Linux/macOS behavior is unchanged.

`sh` expands `$`, backticks and `\` **inside** double quotes, so this does change Linux/macOS behavior. I A/B'd master vs this patch, driving the script with a literal `argv` (`execFileSync('node', [script, path])`, no intervening shell) so the results are attributable to `check-file.js` alone:

**1. Paths containing `$` or a backtick stop working**

| path component | master | this PR |
|---|---|---|
| `$c` | ✅ `All checks passed!` | ❌ prettier receives `…//f.ts` → `No files matching the pattern` |
| `` `b` `` | ✅ passes | ❌ `` `b` `` runs as a subcommand, path becomes `ac.ts` |

It fails loudly rather than silently green, which limits the blast radius — but it's a regression.

**2. A filename becomes a command**

With a directory named `$(touch MARKER)` in the path, through the *documented* entry point:

```
master:   npm run checkFile "…/\$(touch MARKER)/f.ts"  →  marker not created
this PR:  npm run checkFile "…/\$(touch MARKER)/f.ts"  →  marker created
```

Worth being precise about why that's on this patch and not pre-existing: `npm run <script> -- <arg>` does have its own shell layer, but `@npmcli/run-script` shell-quotes the args it appends, so master is safe. The `execSync` here is the only unescaped interpolation. Combined with AGENTS.md instructing agents to run `checkFile` on every modified file, a crafted filename in a checked-out branch executes at lint time.

## Suggestion: skip the shims instead of adding a shell

The `.cmd` problem only exists because we go through the `npm`/`npx` wrappers. Invoking the tools' JS entry points with `process.execPath` sidesteps it — no shell on any platform, so `argv` stays literal, and it drops an `npm run` wrapper process per check:

```js
const { execFileSync } = require('child_process');

const repoRoot = path.join(__dirname, '..');
const binOf = (pkg, rel) =>
  path.join(path.dirname(require.resolve(`${pkg}/package.json`)), rel);
const run = (jsEntry, args) =>
  execFileSync(process.execPath, [jsEntry, ...args], {
    stdio: 'pipe',
    encoding: 'utf8',
    cwd: repoRoot,
  });

run(binOf('prettier', 'bin/prettier.cjs'), ['--write', absolutePath]);
// .scss
run(binOf('stylelint', 'bin/stylelint.mjs'), [absolutePath]);
// everything else
run(binOf('@angular/cli', 'bin/ng.js'), ['lint', '--lint-file-patterns', absolutePath]);
```

Two things I hit building that, in case they save you time:

- `require.resolve('stylelint/bin/stylelint.mjs')` fails — stylelint's `exports` map blocks deep paths. Resolving `<pkg>/package.json` and joining works for all three.
- `cwd: repoRoot` is worth keeping: `npm run` used to guarantee the working directory, and `ng lint` needs to find `angular.json`.

I ran this against master's script on Linux: identical output on a clean `.ts` and `.scss`, and identical `exit=1` plus error text on sabotaged inputs (TS parse error, unclosed SCSS block). It also handles the `$c` path correctly, and invoking it from a subdirectory still works.

I can't verify the Windows side — that still needs your machine. It should be in better shape than the current approach since `node.exe` isn't a `.cmd` shim at all, but please confirm before merge.

## Unrelated, for a follow-up

`tools/test-file.js` already uses the same `execSync(… "${absolutePath}")` shape, so it has the same injection property — and it's still Windows-broken for a different reason (hardcoded `./node_modules/.bin/…`, POSIX-only). Not this PR's job, but if Windows dev support is the goal, `npm run test:file` needs the same treatment.

KamilDev:
Agreed on all points — the "handled identically" claim was wrong, and the injection through the documented entry point settles it. Applied your approach in 966987646: `execFileSync(process.execPath, …)` against the three JS entry points, with the `<pkg>/package.json` resolve trick and `cwd: repoRoot` kept as you described.

Windows verification (Node 24, Windows 11):

- `npm run checkFile` on a clean `.ts`/`.js` and a clean `.scss` → passes, prettier `--write` confirmed working
- sabotaged inputs (TS parse error, unclosed SCSS block) → `exit=1` with the tool's error text
- path with a `$c` component → passes, path arrives literal
- directory named `$(touch MARKER)` through `npm run checkFile` → checks run, no MARKER created anywhere
- invoked from a subdirectory → works (`cwd: repoRoot` doing its job)

On `tools/test-file.js`: confirmed it's still broken here (`'.' is not recognized as an internal or external command`) and it has the same `execSync` interpolation. Happy to send the same treatment as a follow-up PR once this one settles.

johannesjo:
## Round 2 — re-reviewed at `9669876`

Thanks, the rewrite is exactly right. I re-ran the A/B against master driving the script with literal `argv` (`spawnSync(node, [script, path])` — no bash, no `npm run` layer, so nothing can mask or fake an expansion):

| case | master | this PR |
| --- | --- | --- |
| clean `.ts` / clean `.scss` | 0 / 0 | 0 / 0 |
| prettier syntax error `.ts` / `.scss` | 1 / 1 | 1 / 1 |
| `ng lint` error (`prefer-const`) | 1 | 1 |
| stylelint error (`color-no-invalid-hex`) | 1 | 1 |
| path components `$c`, `` `whoami` ``, `a b c` | 0, literal | 0, literal |
| dir named `$(node -e …)` via `npm run checkFile` | no execution | no execution |
| nonexistent file | 1 | 1 |
| **invoked from a subdirectory** | **1** | **0** |

The `$`/backtick regression is gone.

One correction to my own first pass, since it matters for how much the table is worth: my original sabotage fixtures were *syntactically* broken, so they failed at prettier and `ng lint`/stylelint never ran at all — those rows were vacuous. The `prefer-const` and `color-no-invalid-hex` fixtures above are prettier-clean (`prettier --check` exit 0) and confirmed to fail with the raw tools (exit 1 and 2 respectively), so the lint stages are genuinely exercised now.

I also checked that dropping the npm wrapper doesn't change ESLint plugin resolution — a file violating `local-rules/require-text-locale` plus `no-console` produces byte-identical reports on both sides, so the local rules still fire.

### Two bugs this fixes without claiming them

**1. Subdirectory invocation was already broken on master.** `npm run <script>` resolves `package.json` from cwd, so `node tools/check-file.js` from anywhere but the repo root died with `npm error Missing script: "prettier:file"`. `cwd: repoRoot` fixes it.

**2. Prettier errors were silently swallowed.** `error.stdout || error.stderr` short-circuits, and on master `npm run` put its banner on stdout while prettier's diagnostic went to stderr — so a syntax error printed this, and nothing else:

```
❌ Errors found:

> superProductivity@18.15.1 prettier:file
> prettier --write .../bad.ts
```

With the direct spawn, stdout is empty, stderr wins, and you get the actual `SyntaxError: Expression expected. (1:24)` with the source excerpt. Scoped precisely: this only affected the prettier stage — `ng lint` writes to stdout (so it was visible on master, just behind banner noise), and the stylelint path was already identical.

Side benefit: ~3.1s → ~2.1s per check here (two runs each), from dropping the `npm` wrapper process.

### Before merge

- **#9131 needs to go in first.** This branch merges it and it's still open. No conflict risk — the two spec files are the identical blobs (`e50775e9…`, `bebfd864…`) and `git merge-base --is-ancestor` confirms #9131's head is an ancestor here — but merging this today merges #9131 along with it. Landing #9131 first collapses this back to the intended one-file diff.
- **CI has never run on this branch**: zero check runs and zero workflow runs on `9669876` (fork PR, workflows pending approval). I'll approve the run before merging, since this is a tool the contributor docs tell people to use.

### Minor, non-blocking

- `prettier:file` and `lint:file` in `package.json` now have no callers anywhere in the repo — `check-file.js` was the only one. Fine to keep as standalone conveniences; just noting they're unreferenced now.
- `binOf(pkg, rel)` hardcodes each package's bin path. All three match the declared `bin` field today (prettier `./bin/prettier.cjs`, stylelint `bin/stylelint.mjs`, `@angular/cli` `bin/ng.js`), and a rename would fail loudly at resolve time rather than silently skipping a check — so it's fine, just be aware it couples us to dependency internals that a major bump could move.
- On `tools/test-file.js`: confirmed it still has `execSync` with `"${absolutePath}"` interpolated **and** hardcoded `./node_modules/.bin/cross-env` / `./node_modules/.bin/ng` — so it's both injectable and POSIX-only. Yes please, as a separate PR.
- No other open PR touches `tools/check-file.js`.

LGTM.

johannesjo (approved):
Verified on Linux: shell-free `execFileSync(process.execPath, …)` behaves identically to master on clean and lint-failing `.ts`/`.scss`, keeps `$`/backtick/space paths literal, and additionally fixes subdirectory invocation plus the swallowed prettier diagnostics. #9131 has landed, so this now reduces to the intended one-file change.

Note for the record: the red `Tests` check is an unrelated pre-existing flake in `task-detail-panel.component.spec.ts` (zoneless auto-detect running `ngAfterViewInit` outside `fakeAsync`), not required for merge and fixed separately.

PR #9134, opened 2026-07-18
feat(short-syntax): recurrence phrases in the add task bar (@every friday, @daily) — closes #4961

### LINKED ISSUE #4961: 💡 Repeat intervals in short task syntax
### Problem Statement

We can’t repeat a task in the short syntax with `shift`+`a`…

### Possible Solution

A possible solution is to implement this by precise the first letter of the interval in parenthesis.

Example : 
"*Shopping @ fri(w)*" means make a task shopping every friday

_No response_

### Additional context

_No response_

KamilDev:
Closes #4961. First of three planned PRs from the concept posted in https://github.com/super-productivity/super-productivity/issues/4961#issuecomment-5009474127 (1: recurrence syntax, 2: inline highlighting of detected syntax, 3: opt-in trigger-free natural language dates).

> **Stack:** #9134 ⊂ #9135 ⊂ #9137, all targeting `master`. Each later branch contains the earlier ones, so please merge in order (#9134 → #9135 → #9137) and review each PR by its own commits only. Declared here per review feedback; the unrelated timezone spec fix that was previously mixed into these branches has been rebased out and lives solely in #9131.

## What

Recurrence phrases in the Add Task Bar's due-date short syntax. The literal word `every` (plus bare frequency words), as suggested in the issue thread — no `(w)`-style codes:

| Input | Result |
| --- | --- |
| `Water plants @every friday` | weekly on Friday, due next Friday |
| `Team call @every friday 3pm` | weekly on Friday with start time 15:00 |
| `Journal @daily 6am` | daily with start time 06:00 |
| `Standup @every weekday` | Monday through Friday |
| `Pay rent @every 15th` | monthly, due the next 15th |
| `Checkup @yearly` / `@every year` | yearly |

Weekday abbreviations (`fri`) and plurals (`fridays`) work. The bare words `@daily`, `@weekly`, `@monthly`, `@yearly`, `@annually` work too.

**Intervals (`@every 2 weeks`) are deliberately not supported** — per review findings 1–3, the quick-setting presets all mean an interval of 1 (a `DAILY` + `repeatEvery > 1` cfg would wrongly inherit `skipOverdue`, and the repeat dialog can neither display nor preserve a non-1 interval on a preset). Such phrases fall through to the plain-date path, exactly as they did before this PR.

## How

- A small recurrence pre-parser in `short-syntax.ts` runs before the chrono-node date parsing on `@` matches and maps phrases to the **existing** `RepeatQuickSetting` presets. Task + repeat cfg creation goes through the exact same code path as picking a preset in the repeat menu (`getQuickSettingUpdates` + `addTaskRepeatCfgToTask`) — no new creation plumbing, no new op semantics, no sync surface.
- The parsed setting shows on the existing repeat action button (same live feedback as dates/tags); clearing it there also strips the phrase from the input.
- Recurrence is parsed **only in the Add Task Bar** (task creation). Title edits of existing tasks keep the current behavior (phrase parsed as a plain date), since a title edit can't create a repeat cfg. `!` deadline syntax never parses recurrence.
- Anchor dates (`every friday` → next Friday, `every 15th` → next 15th) are computed with the same implied-noon convention chrono uses, so the downstream day-only conversion is unchanged. When an explicit time is given and has already passed on the anchor day (`@every friday 3pm` typed on a Friday at 17:00), the anchor advances one period, matching chrono's `forwardDate` behavior for the plain `@friday 3pm` form (review finding 5).
- An optional chrono match after the phrase is only absorbed when it directly follows it (`@every friday 3pm`); a date expression further into the title (`Standup @every monday and friday`) stays in the title instead of widening the consumed span (review finding 0). Tolerated trailing punctuation is joined to the preceding word (`Water plants @every friday.` → `Water plants.`, review finding 4).

## Tests

- 24 recurrence cases in `short-syntax.spec.ts` covering each mapping, time suffixes, abbreviations/plurals, combination with tags, the negatives (deadline syntax, mid-phrase `every`, `@dailystandup`, edit-path flag off, interval phrases), and regressions for review findings 0/4/5.
- 9 cases in `add-task-bar-parser.service.spec.ts` for the state wiring: set/clear on syntax add/remove, menu-selection preservation, interval fall-through, and input cleanup on clear.
- Wiki reference note `3.04-Short-Syntax.md` updated per the documentation guide.


johannesjo (changes requested):
Thanks for this — and for splitting the concept into three reviewable PRs. The core design choice is the right one: recurrence phrases map onto the **existing** `RepeatQuickSetting` presets and go through `getQuickSettingUpdates` + `addTaskRepeatCfgToTask`, so there's no new creation plumbing, no new op semantics, and no sync surface. I verified the `isParseRepeat` gating too — `short-syntax.effects.ts:121` (the title-edit path) doesn't pass it, and only the add-bar parser and `short-syntax-to-tags` do, exactly as the description claims.

Everything below concentrates in the one place the PR goes *beyond* the presets: the `repeatEvery` interval from `@every 2 weeks`.

---

## Blockers

### 1. `@every N days` (N ≥ 2) silently opts the task into dropping a missed occurrence

`@every 3 days` produces `quickSetting: 'DAILY'` + `repeatEvery: 3`. `getDefaultSkipOverdue` short-circuits on `cfg.quickSetting === 'DAILY'` (`get-default-skip-overdue.ts:29`) and never reaches the `repeatEvery === 1` check on line 32 — so `add-task-bar.component.ts:569` seeds `skipOverdue: true`.

That function's own doc says every-N-days cycles must stay OFF, and gives the reason:

> Safe: today is always a scheduled day, so a missed instance regenerates the same day and can never silently vanish.

That premise is false when `repeatEvery > 1`. If the app isn't opened on a scheduled day, `task-repeat-cfg.service.ts:264-278` skips the occurrence and advances `lastTaskCreationDay`, and it never comes back. Driving the real `_getActionsForTaskRepeatCfg` (start Sun 2024-05-19, every 3 days → 19/22/25; app opened Fri 05-24):

- `skipOverdue: true` → only a cursor advance to `2024-05-22`. **No `addTask`.**
- `skipOverdue: false` → `addTask` with `created = 2024-05-22`, plus the same advance.
- Replaying from the post-skip state on 05-25 → next `addTask` is `2024-05-25`. The 05-22 occurrence never returns.

Scope of the bug — only the `days` unit is affected, because only it maps onto the `DAILY` preset:

```
"@every 3 days"   -> quickSetting=DAILY,                 repeatEvery=3, skipOverdue=true   ← bug
"@every 2 weeks"  -> quickSetting=WEEKLY_CURRENT_WEEKDAY, repeatEvery=2, skipOverdue=false  ← fine
"@every 3 months" -> quickSetting=MONTHLY_CURRENT_DATE,   repeatEvery=3, skipOverdue=false  ← fine
"@daily"          -> quickSetting=DAILY,                 repeatEvery=1, skipOverdue=true   ← intended
```

To be fair about severity: this is a dropped *reminder instance*, local-only — no user content is destroyed and nothing syncs wrongly, and it only bites when a scheduled day is actually missed. But that's precisely the "take meds every 2 days" / "water plants every 3 days" case, where the missed day is the entire reason the reminder exists. I also confirmed this state is unreachable on `master` — the dialog hides the interval input for non-`CUSTOM` and re-normalizes on save, so this PR is the first route to `DAILY` + `repeatEvery > 1`.

### 2. The interval is silently reset to 1 the first time the repeat dialog is saved

A task created with `@every 2 weeks` stores `quickSetting: 'WEEKLY_CURRENT_WEEKDAY'` + `repeatEvery: 2`. `dialog-edit-task-repeat-cfg.component.ts:450-462` re-applies `getQuickSettingUpdates()` on **every** save for any non-`CUSTOM` preset, and every branch of that function hardcodes `repeatEvery: 1`.

Verified at runtime with the real dialog + real formly template: the dispatched changes object is exactly `{repeatEvery: 1}`.

What makes this bad rather than cosmetic:

- The interval input is hidden for non-`CUSTOM` (`task-repeat-cfg-form.const.ts:51`), and the interval appears **nowhere else** in the dialog — the select just reads "Every week on Tuesday". Nothing indicates a 2 is about to become a 1.
- `repeatEvery` isn't in `RELEVANT_KEYS_FOR_UPDATE_ALL_TASKS` (`:65-72`), so there's no "update all instances?" prompt either.
- Opening and cancelling is safe; only Save destroys it. But Save is reached by editing anything unrelated — the title, the estimate, the start date.
- It affects all intervals, not just weekly: `@every 3 days`, `@every 2 months`, `@every 2 years` take the identical path.

These two findings interlock badly: the escape hatch for #1 is to turn `skipOverdue` off in the repeat dialog — and doing so goes through Save, which converts the `@every 3 days` task into a plain every-day task on the way out.

Worth noting the shape of this: the PR adds interval-aware display on the *creation* surface (`_getIntervalRepeatLabel` for the add-bar chip) but not on the *edit* surface — and the edit surface overwrites the value it can't display. The task detail panel can't flag it either: for `repeatEvery !== 1` + `WEEKLY`, `get-task-repeat-info-text.util.ts:39,63-68` falls through to a generic "Custom recur config", so after the reset the label reads "Every week on Tue" — which looks like a label getting *more* specific, not like data loss.

### 3. `@every 2 weekdays` creates a schedule the app can't express or display

It maps to `MONDAY_TO_FRIDAY` + `repeatEvery: 2`, i.e. `repeatCycle: 'WEEKLY'` with Mon–Fri set. But the WEEKLY branch gates on `diffInWeeks % repeatEvery === 0` (`get-newest-possible-due-date.util.ts:79-83`) where `getDiffInWeeks` is `floor(daysSinceStartDate / 7)` — 7-day blocks anchored to `startDate`, not calendar weeks. Observed occurrences (identical harness, only `repeatEvery` differs):

```
start Mon 2026-07-13
  every1: Mon13 Tue14 Wed15 Thu16 Fri17 | Mon20 Tue21 Wed22 Thu23 Fri24 | Mon27 ...
  every2: Mon13 Tue14 Wed15 Thu16 Fri17 | ......... skipped ......... | Mon27 Tue28 Wed29 Thu30 Fri31

start Wed 2026-07-15
  every2: Wed15 Thu16 Fri17 Mon20 Tue21 | ......... skipped ......... | Wed29 Thu30 Fri31 Mon3 Tue4
```

So on a non-Monday start the active window straddles the weekend — 5 working days, then 9 days off. Not "every 2nd weekday", which is how the phrase reads. And `_getIntervalRepeatLabel` returns `null` for `MONDAY_TO_FRIDAY`, so the button falls back to "Every Monday through Friday" — stating a schedule that isn't the one being created.

### Suggested resolution for 1–3

The simplest fix is to **drop `repeatEvery` from this PR** and let `@every 2 weeks` fall through to the plain-date path. The preset mapping on its own is complete, coherent, and shippable, and it's the part that closes #4961.

Note that "just store it as `CUSTOM`" isn't a drop-in: `add-task-bar.component.ts:542` routes `CUSTOM` to `_openRepeatDialogForTask()`, so it would pop a dialog and break the quick-add flow. If you'd rather keep intervals, it needs that branch handled, plus a `MONDAY_TO_FRIDAY` case in `_getIntervalRepeatLabel`, plus something to stop the dialog from resetting the value.

---

## Smaller

### 4. Trailing punctuation leaves an orphan in the title

From the punctuation commit — the lookahead deliberately doesn't consume the punctuation, but the title strip leaves the preceding space:

```
"Water plants @every friday."          -> "Water plants ."
"Buy milk @every friday, then rest"    -> "Buy milk , then rest"
"Water plants @every friday"           -> "Water plants"          (control, correct)
```

### 5. `@every friday 3pm` typed on a Friday after 15:00 creates a task due in the past

`applyRepeatSyntax` takes only hours/minutes from the chrono result and stamps them onto the anchor date, discarding chrono's `forwardDate` roll. Compared against the existing plain-date path at the same instant:

```
now = Fri Jul 17 2026 17:00
  "@every friday 3pm"  -> Fri Jul 17 2026 15:00   (2h in the past)
  "@friday 3pm"        -> Fri Jul 24 2026 15:00   (existing behaviour, correct)
```

So this is a regression against how the same `@` syntax already behaves. The anchor should advance one period when the resulting datetime is already past.

---

## Scope

### 6. Two unrelated spec files, duplicating your own open PR #9131

Merge `42f2495756` pulls in `b360de1aa8` — which is the head commit of your open **PR #9131**, same SHA, same two files, byte-identical. If this PR merges first, #9131 becomes a zero-diff no-op with no explanation of where the fix went. Could you rebase it out and let #9131 land on its own?

(The tz fix itself is good and worth landing — I confirmed those specs genuinely fail on master under `Australia/Sydney` (8 failures, including a real `TypeError`, not just assertion drift), while CI only runs Berlin + LA, so it's latent rather than currently red.)

### 7. Undeclared stack

`#9131 ⊂ #9134 ⊂ #9135 ⊂ #9137`, all targeting `master`. Merging #9137 alone would land all four. Worth flagging in the descriptions (or retargeting the bases) so reviewers don't review the same code three times and so merge order is deliberate.

---

## Verified working

- The implied-noon anchor convention matches chrono's, so the `hasPlannedTime: false` → `dueDay` conversion downstream is correct.
- `@every 0 weeks` clamps to 1; `@every 31st` correctly skips February (Feb 15 → Mar 31).
- Negatives hold: `@everything`, `@dailystandup`, `!every friday`, and mid-phrase `every` all decline to parse as recurrence.
- Weekday abbreviations, plurals and case-insensitivity all resolve correctly.
- All referenced i18n keys exist.
- 270/270 short-syntax + parser specs, 888/888 add-task-bar + task-repeat-cfg specs, `tsc --noEmit`, and `checkFile` all pass locally (no CI has run on the branch yet).

johannesjo:
Adding one more blocker I missed in the review above — it's the most severe of the set, because it silently destroys the task title on input that looks perfectly reasonable.

### 0. `applyRepeatSyntax` swallows everything between the phrase and the next chrono hit

`short-syntax.ts` (in `applyRepeatSyntax`):

```ts
const consumedTotal =
  1 + consumedLength +
  (parsedDateResult ? parsedDateResult.index + parsedDateResult.text.length : 0);
const textToReplace = dueMatch.substring(0, consumedTotal);
```

`parsedDateResult.index` is an offset into the **remainder**, and it's added unconditionally — so the consumed span stretches from the recurrence phrase all the way to a chrono match anywhere downstream, taking every word in between with it. Reproduced on PR HEAD `30fc5555b7` (NOW = Wed 2026-01-14):

| input | resulting title |
|---|---|
| `@every friday call mom tomorrow` | `""` — **whole title destroyed** |
| `Standup @every monday and friday` | `"Standup"` |
| `Review @every week monday notes` | `"Review  notes"` |
| `Water plants @every friday` | `"Water plants"` ✅ (control) |

`Standup @every monday and friday` is a completely natural thing to type, and the user loses "and friday" with no indication. The first row is the bad one though: the task is created with an empty title.

The happy path is unaffected, which is exactly why the 29 new tests don't catch it — every recurrence case in the suite has the phrase at the end of the input.

Fix is to require adjacency, i.e. only absorb the chrono match when it directly follows the phrase (`parsedDateResult.index <= 1`), and otherwise consume just the phrase. Worth adding the three rows above as regression cases.

---

While I'm here — one correction to finding 6 in my review above. I flagged the two spec files (`create-schedule-days.spec.ts`, `plugin-i18n-date.util.spec.ts`) mainly as scope noise. To be clear: **the fix itself is correct and worth landing** — those specs genuinely fail on master in timezones ahead of UTC. The only ask is packaging: it's already your PR #9131, byte-identical, so keeping it there and rebasing it out of this branch avoids #9131 silently becoming an empty PR.

KamilDev:
Thanks for the deep review — all findings applied. The branch was rebuilt (force-pushed), so the history is: feature commit → punctuation fix → one new commit with all review fixes.

**0 — consumed-span overreach:** Fixed with the adjacency rule you suggested: a chrono match is only absorbed when nothing but whitespace sits between the phrase and the match; otherwise only the phrase is consumed. Your three table rows are regression tests now (`@every friday call mom tomorrow` → title `call mom tomorrow`, `Standup @every monday and friday` → `Standup and friday`, `Review @every week monday notes` → `Review notes` — the last one also fixes the double-space seam you showed).

**1–3 — `repeatEvery`:** Took the suggested resolution and dropped intervals entirely. The grammar no longer matches a digit, so `@every 2 weeks` / `@every 3 days` fall through to the plain-date path exactly as before this PR. With the interval gone the `{quickSetting, repeatEvery}` wrapper had one field left, so `shortSyntax()` now returns a plain `repeatQuickSetting` — the `repeatEvery` state plumbing, the interval chip label, and the `Q_EVERY_X_WEEKS` i18n key are all removed, and every created cfg is now byte-identical to what the repeat menu produces. A spec documents the fall-through (and the regex comment explains why intervals are out, so it doesn't get re-added without hitting the dialog/skipOverdue issues you found).

**4 — orphaned punctuation:** Fixed — tolerated trailing punctuation is joined to the preceding word: `"Water plants @every friday."` → `"Water plants."`, `"Buy milk @every friday, then rest"` → `"Buy milk, then rest"` (both are specs).

**5 — past-due anchor:** Fixed — when an explicit time is given and the resulting datetime is already past, the anchor advances one period (weekly +7 days; monthly to the next valid day-of-month, re-stamping the time). `@every friday 3pm` on a Friday at 17:00 now lands on next Friday 15:00, matching the plain `@friday 3pm` behavior; the same-day-before-time case stays on today.

**6 — tz spec commit:** Rebased out of this branch (and out of #9135/#9137, which contain this one). #9131 stands alone again.

**7 — stack:** Declared in all three PR descriptions with the intended merge order.

Verification: `tsc --noEmit` clean, 200 short-syntax + 93 parser + 45 add-task-bar + 11 ranges specs green, and a full unit run passes 13219 specs. The only local failures are exactly the 8 specs #9131 fixes — expected now that its commit is no longer merged into this branch (my machine runs Karma in a UTC+10 timezone; CI won't see them).

johannesjo (changes requested):
Re-reviewed at `1879fe40fd`. **All seven findings from the last round are genuinely fixed** — I verified each by running the code rather than reading the summary, and the resolution for 1–3 (dropping intervals entirely) is the right call: every created cfg is now byte-identical to what the repeat menu produces, which removes the whole class of "preset can't express this" problems.

| # | previous finding | status |
|---|---|---|
| 0 | title-swallow (`@every friday call mom tomorrow` → `""`) | ✅ adjacency rule holds — probed `Standup @every monday and friday please` → `"Standup and friday please"` |
| 1 | `@every N days` → `skipOverdue: true` | ✅ moot — `getDefaultSkipOverdue` only ever sees `repeatEvery: 1` now |
| 2 | repeat dialog resetting `repeatEvery` → 1 | ✅ moot |
| 3 | `@every 2 weekdays` → lying button label | ✅ moot |
| 4 | orphaned trailing punctuation | ✅ `"Water plants @every friday."` → `"Water plants."` |
| 5 | `@every friday 3pm` due in the past | ✅ verified end-to-end: typed Fri 17:00 → `startDate=2024-01-26`, `days=friday` |
| 6 | tz spec commit in-branch | ✅ branch now sits on `014b789c22` (18.15.1); #9131 stands alone again |

`short-syntax.spec.ts` 189/189 and `add-task-bar-parser.service.spec.ts` 86/86 green, `checkFile` clean on both changed files. (No CI has run on this branch — `gh pr checks` reports no checks at all.) I also confirmed #9135 and #9137 both contain this head, so the rebase propagated through the stack correctly.

Three things left, one of which I'd like fixed before merge.

---

### 1. The clear-repeat button uses a second, broader grammar and deletes text that was never a recurrence phrase

`add-task-bar-parser.service.ts:386-393` hand-writes a removal regex instead of reusing `SHORT_SYNTAX_REPEAT_REG_EX`, which you already export. The two disagree, and they disagree today — not just hypothetically. Verified by calling the real `removeShortSyntaxFromInput` (not a transcription of it):

| input | parsed as recurrence? | clear button rewrites the input to |
|---|---|---|
| `Review @every 2 weeks` | **no** (documented as unsupported) | `"Review weeks"` |
| `Review @every 3 days` | **no** | `"Review days"` |
| `Ship @every quarter` | **no** | `"Ship"` |
| `Water plants @every friday.` | yes | `"Water plants"` — drops the `.` the parser deliberately keeps |

This is reachable because the clear button renders for *any* repeat setting, menu-selected included (`add-task-bar-actions.component.html:187` is inside `@if (state().repeatQuickSetting)`). So: user types `Review @every 2 weeks`, it doesn't take (as documented), they reach for the repeat menu instead and pick Weekly, then change their mind and clear — and `@every 2` silently disappears from their text. Given the docs now explicitly tell people intervals aren't supported, "type it, then use the menu" is the expected reaction, not an exotic path.

Fix is to derive the removal from the exported `SHORT_SYNTAX_REPEAT_REG_EX` rather than maintaining a parallel grammar. The four existing `removeShortSyntaxFromInput` specs only cover inputs where the two grammars happen to agree, so none of the rows above are caught. This is the long-term cost worth avoiding: two independent definitions of one syntax will keep drifting apart with every future tweak.

### 2. `@weekly` / `@monthly` plus a time that has already passed today shifts the recurrence day

`applyRepeatSyntax` rolls the anchor forward when the time is already past, but only inside the `if (anchorDate)` branch (`short-syntax.ts:714-746`) — i.e. only when an explicit weekday or day-of-month was given. The unanchored presets fall through to `short-syntax.ts:749`, where chrono's `forwardDate` has already slid the date to tomorrow, and `WEEKLY_CURRENT_WEEKDAY` then picks up tomorrow's weekday. Verified through the real parser service and `getQuickSettingUpdates`:

```
"J @weekly 6am"  typed Wed 10:00 → startDate=2024-01-18 → cycle=WEEKLY days=thursday   ← wrong
"J @weekly 6am"  typed Wed 03:00 → startDate=2024-01-17 → cycle=WEEKLY days=wednesday  ← right
"J @monthly 6am" typed the 17th  → cfgStart=2024-01-18  → monthly on the 18th          ← wrong
"J @weekly"      typed Wed 10:00 → startDate=null→today → cycle=WEEKLY days=wednesday  ← right
```

So the recurrence day depends on what time of day the user happened to type it, and `WEEKLY_CURRENT_WEEKDAY` ends up anchored to a weekday that isn't the current one — which contradicts the preset's own name. The anchored form already does the right thing; this just needs the same roll-forward applied when the phrase implies "today's weekday / today's date".

Low severity (needs an explicit time that's already passed), but it's silent and it persists into the saved cfg.

### 3. Comment at `short-syntax.ts:223-225` describes the opposite of what the engine does

> months without that day (e.g. "every 31st" in February) are **skipped**, matching how the monthly repeat engine **clamps** occurrences

Those are opposite behaviours, and the engine clamps — `get-next-repeat-occurrence.util.ts:139` is `date.setDate(Math.min(day, lastDayOfMonth))`. Net effect: `@every 31st` typed Feb 5 skips February for the *first* instance (→ Mar 31), then clamps for every subsequent one (Apr 30, May 31, Jun 30). Worth making the comment match reality, and worth a thought about whether the first occurrence should clamp too so the two agree.

### Nits

- `\d{1,2}(?:st|nd|rd|th)` doesn't check that the suffix matches the number, so `@every 15st`, `@every 1th` and `@every 3nd` all parse as ordinals. Harmless, just noting it's deliberate-looking leniency.
- `SHORT_SYNTAX_REPEAT_REG_EX` is exported but only used inside its own file. Importing it in the parser service (per point 1) would justify the export; otherwise it can be local.

---

### Checked and explicitly *not* holding against this PR

Flagging these so they don't get re-litigated later — I probed each with `isParseRepeat` on and off, and they behave identically, so they're pre-existing rather than introduced here:

- `@every 0th` leaving `"Pay th"` in the title, and `Ping @daily @daily` leaving a stray second token — both identical on the plain-date path.
- `@every friday morning` consuming "morning": master drops the implied 6am too via `hasPlannedTime: false`, so the add-bar outcome is unchanged.
- `dateStrToUtcDate` returns *local* midnight despite the name, so there's no weekday off-by-one for `WEEKLY_CURRENT_WEEKDAY` in any timezone. I checked this specifically because it would have been a nasty one.
- `@every weekday` on a weekend setting `dueDay = today` — that's the untouched menu-preset path in `add-task-bar.component.ts`.

Nice work on this round — the adjacency fix and the interval removal both landed cleanly, and the added regression cases cover the paths that were hiding the original bug. Point 1 is the one I'd like addressed before merge; 2 and 3 are small.

KamilDev:
Thanks — all three fixed in `fd2db51e8`, plus both nits.

**1. Clear-repeat button deleted text that was never a recurrence (blocker).** Fixed with the regex, not `parsedRanges` — #9134 has to stay independently mergeable, and `parsedRanges` only exists on #9135. The grammar now lives in one string, used twice with different anchoring:

```ts
const REPEAT_PHRASE_SOURCE = '(?:(daily|weekly|...)|every\s+(...))(?=[\s.,;:!?]|$)';

const SHORT_SYNTAX_REPEAT_REG_EX = new RegExp('^' + REPEAT_PHRASE_SOURCE, 'i');

export const SHORT_SYNTAX_REPEAT_REMOVAL_REG_EX = new RegExp(
  `\s*\${CH_DUE}` + REPEAT_PHRASE_SOURCE,
  'gi',
);
```

The leading `\s*` also gets the punctuation join for free — the lookahead stops before the `.`, and eating the space in front of the `@` leaves `Water plants.` rather than `Water plants .`. All four of your rows are specs now:

| input | result |
|---|---|
| `Review @every 2 weeks` | unchanged |
| `Review @every 3 days` | unchanged |
| `Ship @every quarter` | unchanged |
| `Water plants @every friday.` | `Water plants.` |

Sabotage-checked: restoring the old hand-written regex fails them.

**2. `@weekly`/`@monthly` + a passed time shifted the recurrence day.** Root cause is that the `*_CURRENT_*` presets carry no anchor, so they never reached the roll-forward branch at all and took chrono's already-forwarded date verbatim. Rather than duplicating the roll, the phrase's implied anchor is now synthesized when the chrono match is time-only (`hasTime && !isCertain('day') && !isCertain('weekday')`), which routes `@weekly`/`@monthly` through the exact code path `@every wednesday 6am` already used:

```ts
const anchorWeekday = weekday ?? (isTimeOnlyMatch && quickSetting === 'WEEKLY_CURRENT_WEEKDAY' ? now.getDay() : undefined);
const anchorDayOfMonth = dayOfMonth ?? (isTimeOnlyMatch && quickSetting === 'MONTHLY_CURRENT_DATE' ? now.getDate() : undefined);
```

`@yearly` had the identical defect, so it's fixed too (roll +1 year). `@daily` and `@every weekday` are unchanged — chrono's +1 day already equals a one-period roll for them, and there's a spec pinning that. `@weekly monday` still lets the explicit day win (`isCertain('weekday')`), and `@weekly` with no time still resolves to today. Sabotage-checked: 4 of the 7 new specs fail with the implied anchors disabled.

Worth noting the monthly roll reuses `getNextDayOfMonthDate`, so `@monthly 6am` typed Jan 31 after 6am goes to **Mar** 31, not Feb 29 — which leads into:

**3. The `getNextDayOfMonthDate` comment.** You're right that it stated the opposite of the engine. I looked at whether the first occurrence should clamp too, and it shouldn't: the engine reads the recurring day off the *start date* (`dayOfMonthRepeat = startDateDate.getDate()`, `get-next-repeat-occurrence.util.ts:127-129`) and only clamps from there. Clamping the first occurrence of `@every 31st` typed in February would set the start date to Feb 28 and make it recur on the **28th forever**. Skipping costs one late first occurrence and every later month clamps as usual. Comment rewritten to say that.

**Nits.** `SHORT_SYNTAX_REPEAT_REG_EX` is no longer exported — the removal regex is the only thing crossing the file boundary now. And the ordinal-suffix leniency is deliberate, so I said so in the grammar comment (`@every 15st` should hit the recurrence someone meant rather than silently demoting to a plain date).

Wiki updated with the unanchored-preset behavior. Full suite green on all three branches (13339 / 13358 / 13372); #9135 and #9137 merged up.

johannesjo:
Re-verified at `fd2db51e8c` while approving #9135. **All three round-2 items are fixed** — this is not a rejection, and no further work is needed here.

## Round-2 items

**1. Blocker — clear-repeat removal duplicated the parse grammar.** Fixed. One `REPEAT_PHRASE_SOURCE` with two anchorings. I diffed the old and new regex across 12 inputs: `@every 2 weeks`, `@every 3 days`, `@every quarter` and `@weekly-ish` are now left untouched, the punctuation join is preserved (`Water plants @every friday.` → `Water plants.`), and the two are identical on every genuine recurrence phrase I could construct. This was exactly the fix I wanted — the removal now derives from the parser grammar instead of restating it.

**2. Minor — `@weekly`/`@monthly` + an already-passed time shifted the recurring day.** Fixed, and I verified it independently rather than trusting the new specs.

First I confirmed the premise actually holds: `add-task-bar.component.ts:599` passes the parsed due date as `referenceDate`, and `_buildWeeklyForDay(referenceDate).getDay()` plus `startDateDate.getDate()` (`get-next-repeat-occurrence.util.ts:129`) derive the recurring day from it. So chrono's roll to tomorrow really would have changed what `@weekly` recurs on — the bug was real.

Then I asserted the invariant that actually matters — *the day it recurs on must equal the day you typed on, whatever the clock said* — over 38 cases (the same phrase at 7 times of day across 4 weekdays, plus month-end, leap year and three DST boundaries), run under **Europe/Berlin, America/Los_Angeles and Australia/Sydney**:

- **37/38 in all three timezones**
- `Jan 31` → `Mar 31` (correctly skipped rather than clamped to Feb 29)
- DST boundaries (US Mar 10, EU Mar 31, autumn fallback) all keep the 6am wall-clock time

**3. Nit — the skip/clamp comment.** Fixed, and the new wording states the real reason (clamping would make `@every 31st` typed in February recur on the 28th forever), which is more useful than what I asked for.

## One ultra-edge case, mention only — please don't fix it here

`@yearly 6am` typed on **Feb 29** after 6am recurs on **Mar 1**: `anchorDate.setFullYear(anchorDate.getFullYear() + 1)` overflows Feb 29 2025, and unlike the monthly path there's no skip loop. It needs a leap day *and* an already-passed time, and "Mar 1" is a defensible answer for a yearly Feb-29 event anyway — arguably better than skipping to 2028. Recording it so it isn't rediscovered as a mystery later.

## How this lands

`fd2db51e8c` is contained in #9135, which is **approved and green (17/17)**. So this PR's content ships when #9135 merges, and I'll close this one as landed-there rather than merging it separately.

I'm deliberately leaving my `CHANGES_REQUESTED` in place rather than converting it to an approval: this branch's own head shows **2 cancelled checks**, so it has never had a complete green run standalone, and the stale review usefully prevents it being merged on its own by accident. That is a bookkeeping choice, not a judgement on the code — the code is correct, and it is fully covered by #9135's CI and by a full local suite run on the merged content (13380 Berlin / 13366 LA).

Thanks for the thorough turnaround on this one — dropping intervals entirely in round 1 and deriving both regexes from one grammar source in round 2 both made the final result simpler than what I'd asked for.

johannesjo:
Closing as landed-via-#9135: `fd2db51e8c` is contained in that branch, so this PR's content ships when #9135 merges. All round-2 items verified fixed in the note above — closing for merge bookkeeping only, not because anything is outstanding. Thanks @KamilDev!

Issue #8974, opened 2026-07-13
🚨 Numeric followed by 'h' in task title results the text being dropped when saved


darius-r:
### Version Used

Windows: 18.14.0W, Android: 18.14.0_L2A, iPadOS 18.14.0I

### Current Behavior

Discovered while trying to give myself the textual 24h00m time syntax in task titles instead of using AM/PM. 

Any number directly followed by 'h' and a space in a task title is removed when creating a task.

This is true for creation and subsequent edits. 

https://github.com/user-attachments/assets/a7bc1675-d1d4-4f13-b105-057b268adf57

My best guess is that this is a left over fragment from an attempt to set the due time/date by reading a specific syntax in the task title.

### Expected Behavior

For the typed text to be saved in the title.

### Steps to Reproduce

Create or edit a task title with any number followed by 'h'.

Examples: 1h , 2h , 24h .

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

johannesjo:
Thats the short syntax. You can disable it in the settings if you don't like it.

darius-r:
Well damn. I should have caught that it was the short syntax feature. 

Thank you!

Issue #6493, opened 2026-02-12
🚨  odd behavior when using a plus sign in a task name


marcomeyerVFX:
### Version Used

17.1.6 flatpak

### Expected Behavior

When I create or rename a task and call it "Salt + Pepper" in the current project it should be created in the current project and be named "Salt + Pepper". Much like when I create a task "Disney+" shows up as "Disney+" in the current project.

### Current Behavior

It will instead disappear instantly to my default project "Home" regardless of the project I am in and the "+" will be removed from the task name, leaving it as "Salt Pepper". As a workaround using "Salt+Pepper" without the whitespaces works, but doesn't read as nicely. 

I understand that a leading + like in "+Work" without a whitespace moves the task to a certain project. But it's unclear why having it surrounded by whitespaces automatically assumes that it should go to the default project. I would suggest having an exception for standalone " + " signs surrounded by whitespaces.
Disabling short syntax is another option, but it's something I'd like to avoid because it is still quite useful to have.

### Steps to Reproduce

1. Go to a non-default project
2. create a task "Salt + Pepper"
3. look for it in the default project. The task name would have the "+" removed

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
This is because "+" is used for the short syntax to assign a project. You can disable this behavior in the settings under  "Tasks"/"Short Syntax".

marcomeyerVFX:
Yes I am aware, I'm just wondering what the point is of eating away a standalone " + " that is not directly attached to a projectname. "Buy salt + Pepper" vs. "Do taxes +Work" no need to kill that + sign in the first case.

johannesjo:
You're right. We probably should avoid parsing anything with a space behind it.

johannesjo:
This will be adjusted with the next release. Thank you very much!

PR #9135, opened 2026-07-18
feat(add-task-bar): inline highlighting of detected short syntax

KamilDev:
Second of three PRs from the concept in https://github.com/super-productivity/super-productivity/issues/4961#issuecomment-5009474127.

> **Stack:** #9134 ⊂ #9135 ⊂ #9137, all targeting `master`; merge in order. This branch contains #9134 — its first commits are that PR, so review only the highlight commits here until it merges.

## What

Todoist-style inline highlighting of detected short syntax in the Add Task Bar: as you type, every substring the parser will strip from the title (`#tag`, `+project`, `30m/1h`, `@friday`, `@every friday`, `!deadline`, extracted URLs) gets a subtle accent-colored background directly in the input.

## Why

Today the only feedback is the actions row below the input, which shows *that* something was detected but not *which characters* will be removed from the title. That gap is the root cause of real reports of silently eaten title text: #8974 (`4h` in a normal title consumed as an estimate, text dropped on save) and #6493 (`+` in a task name consumed as project syntax). With the highlight, a mis-capture like `Fix 4h timeout bug` lighting up on `4h` is visible before submit, and the existing clear buttons already offer one-click removal.

## How

- Positions are **tracked, not recovered** (rebuilt per review): the parser threads a `TrackedTitle` through every strip site — a per-character map from working-title index to raw-title index, spliced on each removal. `shortSyntax()` returns `parsedRanges` (exact raw-input spans, typed) directly; the former `indexOf`-based mapper and its overlap heuristic are gone. A token whose consumed text is split by an earlier removal (`Call Bob @tomorrow 1h evening`) yields one range per contiguous run instead of no highlight.
- Ranges are pinned to the exact input string they were parsed from, so the async parse can never paint stale highlights onto newer text.
- The overlay is a mirror `div` behind the existing textarea (which already has a transparent background and zero padding): transparent text, highlight backgrounds only, real glyphs stay the textarea's. Typography is shared via a single SCSS placeholder plus a runtime computed-style mirror for arbitrary custom-theme CSS — synced on view init and theme switches only, never per keystroke. Overlay scroll re-syncs after each segments render for the >4-row case. No new dependency, and the mentions dropdown, autosize, and autocomplete are untouched.
- Colors/radii use existing tokens only (`--c-accent` via `color-mix`, matching the `--task-c-current-bg` pattern; `--radius-xs`).

## Tests

- New `tracked-title.spec.ts` (7 cases): offset mapping across removals, split runs, whitespace trimming/collapse, partial (markdown-style) removals.
- `short-syntax.spec.ts`: the cases that previously matched `parsedTokens` with `jasmine.anything()` now assert real ranges, plus a `parsedRanges position tracking` group asserting literal positions for exactly the ambiguous inputs where searching guessed wrong (`#1h retro 1h`, an estimate string inside a URL, the split-range case above, duplicate tag texts).
- 4 parser-service cases: ranges published and pinned to input text, recurrence phrases included, null on no-parse/empty.


johannesjo (changes requested):
Thanks for this — the design instincts here are right, and I want to call them out before the criticism: pinning ranges to `forText` so an async parse can never paint onto newer text, `aria-hidden` + `pointer-events`/`user-select: none`, interpolation rather than `innerHTML`, no new dependency, existing tokens only. The 182-case parser suite still passes locally, so "no behavior change to parsing" holds up.

The concept is worth having. But I'd like the position resolution rebuilt before this goes in, because as it stands the overlay can highlight the wrong characters — which is the one outcome your own docstring rules out ("a missing highlight is harmless, a wrong one is not").

## Blocker: highlights land on the wrong characters

`mapShortSyntaxTokensToRanges` locates every token with `rawText.indexOf(token.text)`. That's a guess, and it guesses wrong whenever the token text also occurs earlier inside a word, a tag, or a URL.

I ran these through the real parser (Chrome headless, `urlBehavior: 'extract'`):

| input | parser produces | highlighted |
|---|---|---|
| `#1h retro 1h` | `retro` | `1h` **inside the tag `#1h`** (idx 1; correct is 10) |
| `Read https://ex.com/30m-guide later 30m` | `Read later` | `30m` **inside the URL** (idx 20; correct is 36) |

It compounds: once the bogus range is claimed, the overlap check at `short-syntax-ranges.ts:34` evicts the *correct* `#1h` tag range and the *correct* URL range. So the second row shows the user three tinted characters in the middle of a URL, while the URL and the `30m` — both of which actually get deleted — render as plain text. A user who learns to trust the highlight is worse off than one who ignores it.

For the estimate specifically the exact index is already sitting there unused: `parseTimeSpentChanges` runs on the raw title, so at `short-syntax.ts:315` `timeMatch.index + timeMatch[0].indexOf(trimmed)` is exact. `short-syntax.ts:317` keeps only the trimmed text and discards it.

## What I'd like instead of `indexOf`

Rather than patching the estimate, I'd rather see positions tracked rather than recovered, so the guessing goes away for every token type.

Concretely: carry an offset map alongside the working title — an array where `offsets[i]` is the raw index of working-title character `i`, starting as the identity. Every time a stage removes a span, splice those entries out. A stage that matches at working index `j` then maps to raw index `offsets[j]`, exactly, with no search.

That means the strip sites need to route their edits through a small helper instead of bare `.replace()`. The ones I found:

- `short-syntax.ts:828` — `parseTimeSpentChanges`, `.replace(matchSpan, '').trim()`
- `short-syntax.ts:462` and `:483` — `parseProjectChanges`, note the `.replace('  ', ' ')`, which is a *mid-string* edit, not a removal at the match site
- `parseTagChanges` — a `.replace()` per tag, then `.trim()`
- `parseUrlAttachments` — plus markdown links, which are a partial removal (`[text](url)` → `text`), not a full one

It's a real refactor rather than a one-liner, but it's mechanical, and the 182-case suite covers exactly the code being moved. The payoff is that `mapShortSyntaxTokensToRanges` and the overlap heuristic both disappear — tokens become ranges directly, and the "fail-safe skip" path stops being load-bearing.

It also fixes a case that silently produces nothing today: `Call Bob @tomorrow 1h evening` consumes everything (title becomes `Call Bob`) but shows **no due highlight at all**, because the due stage sees the estimate-stripped title and its token text `@tomorrow evening` isn't verbatim in the raw input. Date + duration together is an ordinary thing to type, and it's where this feature should be doing its best work.

If you'd rather stage it: threading the exact index for the estimate alone fixes both rows in the table above, since the bogus estimate claim is what was evicting the other two ranges. But I'd want the offset map as the endpoint, not the stopping point.

## `getComputedStyle` on every keystroke

`add-task-bar.component.ts:311` reads `highlightSegments()` inside `_overlayTypographySyncEffect`, so a forced style recalc plus 15 inline style writes run on every character typed — with the task list rendered behind the bar. Computed typography can't change per keystroke; the real triggers are theme switch and view init.

Deleting that one line is sufficient. The effect still re-runs when it needs to: `_syncOverlayTypography` reads `this.inputEl()` and `this.highlightEl()`, so those viewChild signals stay tracked, and `activeRef()` covers theme changes.

## Smaller things

- `parsedTokens: jasmine.anything() as any` in the existing `short-syntax.spec.ts` cases asserts nothing — it matches any non-null value. The PR description calls these "extended to cover the new return field", which overstates it. Token text is deterministic, so asserting real values across those cases is close to free, and it is what would have caught the blocker above.
- `short-syntax-ranges.spec.ts:52` still has a thinking-out-loud comment (`// '30m' also occurs inside the due token '@friday 30m'? Construct overlap:`), and that test uses `{ type: 'tag', text: 'friday' }` — a token shape the parser can't emit, since tags always carry `#`. Worth using a real one so the test documents actual behavior.
- `ShortSyntaxTokenType` is threaded end to end but every type renders identically (`[class.syntax-highlight]="!!seg.type"`). Fine as a seam for later — just noting nothing consumes the distinction yet.
- Unverified, but worth a look: `syncHighlightScroll()` fires only on `(scroll)`. Past 4 rows the textarea's caret-scroll likely happens before Angular re-renders the overlay, so `scrollTop` would clamp against still-short content and stay desynced until the next scroll event. Calling it from the segments effect would be a cheap hedge.
- Not mirrored onto the overlay: `direction`, `text-align`, `word-break`. The SCSS already acknowledges themes are arbitrary CSS; just noting the property list isn't exhaustive.

## Verified locally

521 specs green (`short-syntax*` + `add-task-bar*`), `npm run checkFile` clean on all six changed `.ts`/`.scss`, and the `@extend %main-input-typography` compiles to a single shared rule covering both layers. CI hasn't run on `df61d863ab` yet.

One thing to factor into the design, since these are stacked: #9137 drops the `@` requirement for dates, which raises the accidental-capture rate a lot. Whatever this overlay ends up guaranteeing about accuracy should be sized against #9137's ambiguity, not this PR's.

KamilDev:
All applied — I went straight to the offset-map endpoint rather than the staged estimate-index fix, in `21ce5df94` + `708964ab9`.

## Blocker: positions tracked, not recovered

New `tracked-title.ts`: a `TrackedTitle` carries the working title plus an `offsets` array (`offsets[i]` = raw index of working char `i`), spliced on every edit. All strip sites now route through it:

- `parseTimeSpentChanges` — removal at `matches.index` instead of `.replace(matchSpan, '')` (this also fixes the latent wrong-span removal when the literal match text occurs earlier without satisfying the lookahead)
- due/deadline (`parseShortSyntaxDate`, `applyRepeatSyntax`) — including the punctuation-join and double-space edits as tracked removals
- `parseProjectChanges` — including the mid-string `'  '` → `' '` edit (one-char removal)
- `parseTagChanges` — per-tag removals
- `parseUrlAttachments` — markdown links as two partial removals (`[` and `](url)`), plain-URL removals per normalized occurrence, and whitespace collapse as a tracked op that keeps each run's first raw offset

`shortSyntax()` now returns `parsedRanges` (typed raw-input spans) directly; `mapShortSyntaxTokensToRanges` and the overlap heuristic are deleted, and `short-syntax-ranges.ts` shrinks to `splitTextByRanges`. The exported `parseTimeSpentChanges`/`parseProjectChanges` keep their old signatures as thin wrappers, so `markdown-paste.service` and `select-task` are untouched.

Both table rows are now regression tests asserting literal positions (`'#1h retro 1h'` → tag `[0,3)` + estimate `[10,12)`; the URL case → url `[5,29)` + estimate `[36,39)`). And the case that previously produced nothing: a span split by an earlier removal now yields one range per contiguous run, so `Call Bob @tomorrow 1h evening` highlights `@tomorrow` (due) + `1h` (estimate) + `evening` (due) — also a literal-position test.

## `getComputedStyle` per keystroke

Dropped the `highlightSegments()` read from the typography effect, as suggested — it now tracks only `activeRef()` (and the viewChild signals via `_syncOverlayTypography`).

## Smaller things

- **Weak spec assertions:** all 38 `parsedTokens: jasmine.anything()` sites now assert real ranges (via a small `expectedRanges` helper for the unambiguous inputs; literal numbers for the ambiguous ones). You were right that this would have caught the blocker — the new literal-position tests fail against the old resolver.
- **`short-syntax-ranges.spec.ts`:** the thinking-out-loud comment and the impossible `{ type: 'tag', text: 'friday' }` token went away with the deleted describe; the offset-map behavior is covered by `tracked-title.spec.ts` (7 cases) and the integration tests above.
- **Scroll sync:** confirmed your reasoning and wired `syncHighlightScroll()` into an `afterRenderEffect` on the segments, so the overlay re-syncs after each render rather than only on `(scroll)`.
- **Property mirror:** `direction`, `text-align`, `word-break` added to `_syncOverlayTypography`.
- **`ShortSyntaxTokenType`:** kept as the range discriminant (unchanged rendering, still a seam).

One pre-existing gap that stays: markdown links still produce no highlight ranges (only actually-removed plain URLs do), same as before.

On the #9137 sizing note: agreed, and the offset map transfers as-is — `parseNaturalLanguageDue` now operates on the same `TrackedTitle` (merged upward already), so trigger-free detections get exact positions under the identical mechanism, including the suppression path (a dismissed detection is rejected before any tracker edit).

Verified: full `short-syntax`/`add-task-bar` suites green (204 + 199 specs on the stack top), lint + prettier clean on all touched files.

johannesjo (commented):
## Round 2 — all three blockers fixed ✅, one new UX defect found

Thanks — this is a much stronger implementation than round 1. The `TrackedTitle` offset map is exactly the right primitive, and I'm glad you took the full route rather than the estimate-index-only staging step: `mapShortSyntaxTokensToRanges`, `parsedTokens` and the overlap heuristic are gone entirely, so the whole class of "guess the position by searching" bugs is deleted rather than patched.

**Verified locally at `708964ab9b`:**

- All three reported cases are now spec'd with exact literal indices — I recomputed each independently and they match. `Call Bob @tomorrow 1h evening` goes further than I asked: instead of failing safe with no highlight, it now emits a correctly **split** due range `[9,18)+[22,29)` around the stripped estimate. The limitation I documented is gone.
- The vacuous `jasmine.anything()` assertions are replaced by 44 real ones — and they're load-bearing: sabotaging the offset map (`raw = i` instead of `_offsets[i]`) fails **14 specs in `short-syntax.spec.ts` alone**.
- `expectedRanges` is a genuinely independent oracle (raw `indexOf` in the spec), with hardcoded literals for the ambiguous cases. That's the right test design.
- Full suite green in both TZ variants (13231 Berlin / 13217 LA); CI is green 17/17. A 32-case adversarial probe of the overlay-alignment invariant (sorted, non-overlapping, in-bounds, segments reconcatenate to the raw input) passes.
- The per-keystroke `getComputedStyle` cost is gone. 👍

Structurally the reason it's now correct: every `rawRanges()` is paired with a `remove()` whose span is a superset, so no raw character can be consumed twice — overlap is impossible by construction.

### 🐛 Highlights flicker on every keystroke

Typing makes the highlights blink out and back. It's structural, not a tuning issue:

1. `inputTxt$ = toObservable(inputTxt)` — `toObservable` never emits synchronously, so the parse can't even *start* before the paint that follows the keystroke.
2. That paint runs `highlightSegments()`, which sees `syntaxHighlight().forText` = the *previous* text, hits `highlight.forText !== txt`, and returns `[{ text: txt, type: null }]` — **every highlight dropped**.
3. The parse resolves and the next paint restores them.

So there's guaranteed ≥1 painted frame with zero highlights per keystroke. There's no debounce involved, so removing one wouldn't help.

The `forText` pin is right; only the *fallback* is too blunt. Keep the ranges the edit can't have moved — those inside the unchanged common prefix:

```ts
const highlight = this.stateService.syntaxHighlight();
if (!highlight || highlight.ranges.length === 0) {
  return [{ text: txt, type: null }];
}
if (highlight.forText === txt) {
  return splitTextByRanges(txt, highlight.ranges);
}
// The parse is async, so every keystroke renders once with ranges from the
// previous text. Dropping them all blanks the highlights for a frame (visible
// flicker), so keep the ones the edit cannot have moved.
let common = 0;
const max = Math.min(highlight.forText.length, txt.length);
while (common < max && highlight.forText[common] === txt[common]) {
  common++;
}
const stillValid = highlight.ranges.filter((r) => r.end <= common);
return stillValid.length
  ? splitTextByRanges(txt, stillValid)
  : [{ text: txt, type: null }];
```

A highlight can then never be *mispositioned*, only up to one keystroke stale — which is what editors generally do, and it matches your own "a missing highlight is harmless, a wrong one is not".

### The runtime typography sync should be static CSS

I assumed this was unnecessary, measured it, and was wrong — so first, the risk is real. With only `font-family`/`font-size`/`line-height` pinned (what `%main-input-typography` does today), inside a host that sets inheritable typography, a real browser gives four desyncs:

```
fontWeight       div=600         vs textarea=400
fontStyle        div=italic      vs textarea=normal
fontVariantCaps  div=small-caps  vs textarea=normal
letterSpacing    div=2px         vs textarea=normal
```

The cause is that a textarea doesn't inherit typography at all — the UA sheet applies `font: 400 13.333px monospace`, and that shorthand resets weight/style/variant/stretch/line-height, while the mirror div inherits.

But that's a **static** problem. Declaring the full set on `%main-input-typography` (so both layers state the same values) measured **0 differences across every property I tested** — no JS:

```scss
%main-input-typography {
  // A textarea does NOT inherit typography — the UA sheet sets
  // `font: 400 13.333px monospace` and the shorthand resets weight/style/
  // variant/stretch. The mirror div inherits instead, so both layers must
  // declare the full set or they drift under any theme that sets them.
  font-family: var(--font-primary-stack);
  font-size: 16px;
  font-weight: 400;
  font-style: normal;
  font-variant: normal;
  font-stretch: normal;
  line-height: 1.4;
  letter-spacing: normal;
  word-spacing: normal;
  text-transform: none;
  font-kerning: auto;
  font-feature-settings: normal;
  text-indent: 0;
  tab-size: 8;
  white-space: pre-wrap;
  overflow-wrap: break-word;
}
```

That lets `_overlayTypographySyncEffect`, `_syncOverlayTypography` and the `CustomThemeService` injection all be deleted. Three reasons to prefer it:

1. **The runtime trigger is a guess** — theme change plus a hardcoded 300ms timer. A theme stylesheet that loads slower than that leaves the overlay desynced until the next theme switch.
2. **It's incomplete where it counts** — 17 text properties, but not `padding`/`border`/`box-sizing`/`width`, which are what actually shift the text origin.
3. **The cited justification doesn't hold.** `glass.css`'s `* { font-family }` has specificity (0,0,0) and no `!important`, so it loses to both `.main-input` and `.main-input-highlight` — and it would apply to both layers equally anyway. No shipped theme has a `textarea`-specific typography rule.

### Minor

- **Markdown links get no highlight.** `[docs](https://a.com) later` → title `docs later`, attachment created, `parsedRanges` empty. 15 characters vanish with no highlight, while a bare URL highlights fine (`short-syntax.ts:962-994` pushes ranges only in the plain-URL loop). Harmless by your invariant, but it makes the new wiki line's "exactly which characters will be stripped" a bit overstated.
- **`splitTextByRanges` fails unsafely** if the sorted/non-overlapping assumption is ever broken by a future stage: an overlapping range re-emits text (`abcdefgh` → `abcdedefgh`), desyncing the overlay so *every* later highlight lands wrong. `if (range.start < pos) continue;` makes it total — silent corruption becomes a dropped highlight.
- **Markdown removal is expressed twice** — `.replace()` in `extractMarkdownLinks` (which derives `plainUrlMatches`) vs two `tracked.remove()` calls. Equivalent today; if either changes, `plainUrlMatches` silently desyncs from `tracked.text`. Deriving the plain-URL matches from `tracked.text` after the markdown removals would collapse them.
- `applyRepeatSyntax`/`consume()` still locate spans with `indexOf` — I probed this and it's **not** a live bug (the strip and the highlight share the same index, so they can't disagree). Noting it only so it isn't mistaken later for a return of the original problem.

### Merge order

This PR contains #9134, so merging it also ships #9134's open blocker (the clear-button removal regex that duplicates the parse grammar and turns `Review @every 2 weeks` into `"Review weeks"`). Worth noting that **`parsedRanges` from this PR is exactly the data that function hand-rolls** — slicing the input by the exact ranges would remove that second grammar entirely. Might be the cleanest way to close #9134's blocker.

KamilDev:
Round 2 applied — `d66843e7e` on `feat/short-syntax-highlight`. Full suite green (13366), and the merge into #9137 is pushed too (13380).

### 🐛 Highlights flicker on every keystroke

Fixed with your common-prefix fallback, verbatim. You were right that this is structural and not a debounce issue: `toObservable(inputTxt)` can't emit before the post-keystroke paint, so the mismatch branch ran once per keystroke by construction.

Three specs in `add-task-bar.component.spec.ts` cover it — the exact-match path, an append that keeps the earlier range, and an insert *before* the range that correctly drops it. Sabotage-checked (forcing the filter to keep nothing fails the append spec).

### Runtime typography sync → static CSS

Deleted, exactly as you proposed. `%main-input-typography` now declares the full set, and `_overlayTypographySyncEffect`, `_syncOverlayTypography` and the `CustomThemeService` injection are gone (−49 lines, no `getComputedStyle`, no 300ms timer).

Your reason (3) is the one that settles it — I checked the specificity claim and you're right: `* { font-family }` is (0,0,0), loses to both `.main-input` and `.main-input-highlight`, and hits both layers identically anyway. The note I'd written justifying the sync was wrong.

I took your reason (2) as well and declared `box-sizing: border-box; padding: 0; border: 0` on the overlay, since those move the text origin and the sync never covered them.

The `span { font: inherit; ... }` block stays — that one *is* about theme rules matching the spans directly (a declaration beats inheritance regardless of specificity), which is a different mechanism from the textarea's UA `font:` shorthand. The comment now says so instead of conflating the two.

### `splitTextByRanges` fails unsafely

Added `if (range.start < pos) continue;` plus a spec asserting the segments still concatenate back to the input exactly. Agreed on the reasoning: one dropped highlight beats a mirror that's longer than the textarea.

### Markdown removal expressed twice

Collapsed into one definition, `collapseMarkdownLinks(tracked)`. The `.replace()` in `extractMarkdownLinks` is gone; the plain-URL scan now reads the tracker's text *after* the collapse ran, so the two can't drift.

It runs on a scratch `TrackedTitle` for the pre-checks, because the real tracker must stay untouched when there turns out to be nothing to extract (`[text]()` matches the link regex but yields no URL — spec added).

### Markdown links get no highlight

Fixed rather than documented away. `collapseMarkdownLinks` now returns the raw ranges of the characters it drops, so `[docs](https://a.com) later` highlights `[` and `](https://a.com)` and leaves the display text plain — which is precisely what gets stripped. Two specs pin the literal indices.

I softened the wiki line anyway: whitespace-collapse still edits characters that deliberately aren't highlighted (`rawRanges` trims whitespace run edges — a highlight on a consumed space is noise), so "exactly which characters" was overstating it regardless.

### `parsedRanges` restructure of `removeShortSyntaxFromInput` — I'd rather not, and I think the premise expired

Two verified reasons, and I'd take the third option at the bottom over building this:

**1. There is no `'repeat'` token type to slice by.** `ShortSyntaxTokenType` is `due | deadline | estimate | tag | project | url` (`short-syntax.ts:24-30`), and a recurrence phrase is pushed as `'due'` (`short-syntax.ts:351`, `pushRanges('due', dueResult.ranges)`) — indistinguishable from a plain date in the range list. So "slice out the repeat ranges" would delete `@tomorrow` from `Call mom @tomorrow` on a clear-*repeat* click. That's the mirror image of the round-1 blocker: a strip that removes text the user didn't ask to remove.

**2. The ranges are pinned to a different string than the one the button reads.** `clearRepeatSetting()` reads live `stateService.inputTxt()` (`add-task-bar-actions.component.ts:339`), while `syntaxHighlight()` is pinned to `forText` of the last *completed* async parse. Any keystroke between the parse landing and the click makes those disagree, and slicing on stale indices silently corrupts the user's input. That needs a `forText === currentInput` guard — and the only safe fallback for the guarded-out case is the regex, so the regex has to stay either way.

There's also a behavior change hiding in it: the ranges cover the adjacent chrono time (`tokenEnd = phraseStartIndex + consumedAfterStart`), so `@every friday 3pm` would lose the `3pm` too, contradicting the existing "leave a trailing time token like the date case does" spec.

And the duplication that motivated it is already gone as of `fd2db51e8` (the #9134 fix you asked for): there's one `REPEAT_PHRASE_SOURCE` with two anchorings now, so the removal derives from the parser grammar rather than restating it. The restructure would add code, not remove it.

**The one thing that would change my mind:** if you want recurrence highlighted *distinctly* from a plain date in the overlay — different class, so `Water plants @every friday` doesn't read as "a due date". That earns a `'repeat'` token type on its own merits, and once it exists, reason 1 evaporates and the slicing comes almost free (reason 2 still needs the staleness guard). Happy to do that here or in a follow-up — just say which, or say no and I'll leave the regex as is.

johannesjo (dismissed):
Round 3 reviewed at `d66843e7e3`. **All five claims verified — approving.** Two minor items below that I'd like folded in before merge if you agree, but neither blocks.

Clearing my round-1 `CHANGES_REQUESTED`, which was the only thing holding the merge.

## What I verified

- Full suite, both TZ variants: **13380 (Berlin) / 13366 (LA) SUCCESS**. `npm run checkFile` clean on all 10 changed files, reformatted nothing. CI **17/17** on the exact head SHA.
- **48-input adversarial parser probe** — sorted / non-overlapping / in-bounds / segments reconcatenate exactly to the raw input / no range boundary splits a surrogate pair or a grapheme cluster. **0 failures**, including ZWJ emoji (`👨‍👩‍👧‍👦`), RTL, CJK, tabs, newlines, `#shop🎉`-style token-adjacent emoji, and nested markdown.
- **Sabotage-tested that probe**, since a probe that can't fail proves nothing: making `rawRanges` return the working index instead of `_offsets[i]` → 8 inputs fail; deleting the `parsedRanges.sort()` (short-syntax.ts:406) → 17 inputs fail. Both the offset map and the sort are load-bearing, and the sort matters more than it looks now that `splitTextByRanges` *silently drops* an out-of-order range.
- **The flicker fix is correct by construction.** Every kept range lies entirely inside a byte-identical common prefix, so the highlighted characters are the same characters at the same indices — a kept range can be stale but never mispositioned. Backspacing *inside* a token still drops its range for one parse cycle, which is inherent, not a defect.
- **The static-CSS replacement is right, and I proved it in Chrome rather than reasoning about it.** Under four hostile theme shapes the full set keeps every metric property identical between the two layers. Both sabotage arms fire: the old partial set drifts 7 properties (`font-weight`, `font-style`, `font-variant`, `font-stretch`, `letter-spacing`, `word-spacing`, `text-transform`) under a theme that styles an ancestor, and deleting the `span { font: inherit }` block lets a `*`-level font rule through. Your UA-`font`-shorthand reasoning is confirmed empirically. Dropping the 300ms timer and the `getComputedStyle` is a clear win, and covering `padding`/`border`/`box-sizing` closes a hole the old sync never did.
- **#9134's blocker — which ships with this PR — is fixed.** `SHORT_SYNTAX_REPEAT_REMOVAL_REG_EX` now leaves `@every 2 weeks`, `@every 3 days`, `@every quarter` and `@weekly-ish` untouched, keeps the punctuation join, and is byte-identical to the old regex on every genuine recurrence phrase I could construct.

## 1. The deleted sync covered three properties the static block doesn't

`add-task-bar.component.scss:101` omits `direction`, `text-align` and `word-break`, all of which `_syncOverlayTypography` used to copy. Measured in Chrome, a theme containing:

```css
textarea { text-align: center; direction: rtl; word-break: break-all; }
```

desyncs all three (`ta="center"` vs `ov="start"`, etc.) — the backgrounds stay start-aligned under centered glyphs. Separately, the `span` block at :152 omits `white-space`/`overflow-wrap`, so `span { white-space: nowrap }` collapses the mirror from 3 rendered lines to 1 (517.8px vs 366.7px) — every highlight past the first wrap lands on the wrong text.

No shipped theme does any of this — I grepped all 14 and none set `direction`, `text-align` or `word-break` anywhere, and the app's own RTL comes from `[dir]` on a shared ancestor (`app.component.html:42`), which both layers inherit identically. So this is hardening for arbitrary user CSS, not a live bug. But it's ~5 lines of static CSS and no JS returns, which seems worth it given the block already carries a "keep every property below" comment.

## 2. `[label]()` plus another URL now mangles the title

`collapseMarkdownLinks` (short-syntax.ts:954) drops the `[` and `](…)` for **every** regex match, and only guards `if (url)` for pushing the URL (:969). Whether the real collapse runs at :1030 is decided by `allUrls` being non-empty — which an *unrelated* plain URL satisfies. Verified against the previous head:

| input (`urlBehavior: 'extract'`) | `708964ab9b` | `d66843e7e3` |
| --- | --- | --- |
| `Add [Website]() https://a.com` | `Add [Website]()` | `Add Website` |
| `read [empty]() https://x.com/y #shop` | `read [empty]()` | `read empty` |
| `Add [Website]()` | untouched | untouched |

So the rule your new spec states — "should not touch the title when a markdown link carries no URL" — now holds only when there's no other URL in the same title. The highlight is honest about it (it marks exactly what gets stripped), so this is title mangling rather than a highlight bug, and the input is rare.

Worth noting the clean fix isn't quite free: skipping URL-less matches would also change `[a]() [b](https://b.com)` from `a b` to `[a]() b`, which differs from both heads. Your call whether to make it consistent or just document that extract mode collapses all markdown syntax once any URL is found — I mainly want the two paths to agree on one rule.

## On the things you pushed back on

**`removeShortSyntaxFromInput` restructure — you're right, dropping it.** I verified both reasons: `ShortSyntaxTokenType` has no `'repeat'` member and `pushRanges('due', …)` makes a recurrence indistinguishable from a plain date in the range list, and `clearRepeatSetting()` does read live `inputTxt()` while the ranges are pinned to the last completed parse. And the duplication I actually objected to is gone — one `REPEAT_PHRASE_SOURCE` with two anchorings derives both behaviours from the parser grammar, which was the point.

**Distinct `'repeat'` token type — no, let's not.** `@every friday` genuinely does set a due date, so a second colour would be drawing a distinction the data doesn't really make, and a new highlight class is permanent surface for no workflow gain. One calm colour is the right default here. If real demand shows up later it can earn its way in then.

## Housekeeping

#6511 ("Fix URL parsing in task titles") also touches `short-syntax.ts`, and this PR rewrites `parseUrlAttachments` substantially — whichever lands second will need a careful rebase.

Nice work on the offset map; it held up under everything I threw at it.

KamilDev:
Thanks for the approval — and for re-proving the typography in Chrome rather than taking the reasoning at face value. Both minor items are folded in as `020ca047c`.

## 1. The five missing properties — added

`%main-input-typography` gains `direction`, `text-align`, `word-break`; the `span` block gains `white-space`, `overflow-wrap`, `word-break`, `direction`, `text-align`.

All of them as `inherit` rather than fixed values, which matters for two of the three: hardcoding `direction: ltr` / `text-align: start` would pin the layers to LTR and break the `[dir]`-driven RTL you pointed at. `inherit` gives the property the block was actually there for — *both layers resolve the same value from the same ancestor* — while still outranking any non-important `textarea { ... }` theme rule at component specificity.

Worth recording that this is a no-op for the rendered result today: `:host` already sets `text-align: start` (add-task-bar.component.scss:5) and nothing between it and the input overrides it, so `inherit` resolves to exactly the `start` the textarea had by UA default. Same for the other two. So it's your case 1 hardening, with no behaviour change to review.

I added `direction`/`text-align` to the `span` block too, not just the `white-space`/`overflow-wrap` you named — a `* { direction: rtl }` hits the spans directly and reorders the mirrored glyphs while the textarea keeps the ancestor value, which is the same mechanism as the `font: inherit` line already there.

## 2. `[label]()` — fixed, not documented

I went with making the rule unconditional rather than writing the conditional one down. `collapseMarkdownLinks` now skips a match with an empty destination:

```ts
const url = m[2];
if (!url) {
  continue;
}
urls.unshift(url);
```

The reasoning: a URL-less link produces no attachment, so collapsing it only deletes the user's characters and gives nothing back — and under the old code whether it happened was decided by whether some *unrelated* URL appeared elsewhere in the title. Documenting that keeps a rule where the same substring survives or doesn't depending on text it has nothing to do with, and the user has no way to see which happened.

You flagged that the clean fix isn't free, and you're right about the cost: `[a]() [b](https://b.com)` goes from `a b` to `[a]() b`, differing from both heads. I think that's the correct side of the trade rather than a regression to absorb — `[a]()` yielded nothing, so nothing should be taken from it — and it makes the highlight honest for free, since the brackets that stay are brackets no range covers.

Two specs added, and I sabotage-checked them by restoring the old `if (url)` guard — both fail (203/205):

| input (`urlBehavior: 'extract'`) | before | `020ca047c` |
| --- | --- | --- |
| `Add [Website]() https://a.com` | `Add Website` | `Add [Website]()` |
| `[a]() [b](https://b.com)` | `a b` | `[a]() b` |
| `Add [Website]()` | untouched | untouched |
| `[docs](https://a.com) and https://b.com` | `docs and` | `docs and` |

The first spec also pins the plain URL's literal raw range, so the highlight is asserted against the kept brackets, not just the title.

Full suite green (13368) and `checkFile` clean on all three changed files.

## Housekeeping

Noted on #6511 — whichever of the two lands second, I'll take the rebase on this branch rather than leave it to you.

And thanks for the call on #9137. The two cases you hit are fair, and the `chrono-node` range argument is the one I don't have an answer to — a minor bump silently changing what gets deleted from titles isn't something a tighter guard fixes. #9134 and this one are the parts I'd want to keep too.

johannesjo (approved):
Re-approving at `020ca047c3`. Both minor items are fixed, and I re-ran the two probes that found them rather than reading the diff — the point of having built them was to be able to prove the gaps are closed, not just plausibly addressed.

**1. The typography gaps are closed.** Same Chrome harness, now with your `inherit` declarations. Across six hostile theme shapes — ancestor-typography, `* { font-family }`, `span`-targeted, `textarea { text-align:center; direction:rtl; word-break:break-all }`, `span { white-space:nowrap }` and no-theme — the overlay, its spans and the textarea now agree on **every** metric property. The two arms that used to fire are silent: the `textarea`-targeted rule no longer drifts `direction`/`text-align`/`word-break`, and `span { white-space: nowrap }` no longer collapses the mirror to one line.

Both sabotage arms still fire, so the probe hasn't gone vacuous: the old partial set still drifts 7 properties under an ancestor rule, and deleting the `span` block still lets `* { font-family }` and `span { … }` through.

You're right that `inherit` is the correct choice over fixed values, and for a reason I'd underweighted when I named the three properties — I was thinking only about outranking a `textarea` rule, and hardcoding `text-align: start` would have done that while quietly pinning both layers LTR and breaking the `[dir]` RTL. `inherit` gets the specificity win *and* keeps both layers resolving from the same ancestor, which is the actual invariant. Adding `direction`/`text-align` to the span block beyond the two I named is the same argument and correct.

**2. `[label]()` — agreed, and you picked the right side of the trade.** I flagged that `[a]() [b](https://b.com)` → `[a]() b` differs from both previous heads, and I'd rather have that than the old rule, for the reason you gave: whether the user's characters survived was decided by unrelated text elsewhere in the title, with nothing rendered to show which way it went. Unconditional is the rule that can actually be explained.

I re-ran my range-invariant probe on the new head — 53 adversarial inputs, **0 failures** (sorted / non-overlapping / in-bounds / segments reconcatenate exactly / no boundary splits a surrogate pair or grapheme cluster). The new cases land where your table says:

    Add [Website]() https://a.com    → url[16,29)="https://a.com"   (brackets kept, unhighlighted)
    [a]() [b](https://b.com)         → url[6,7)="[" url[8,24)="](https://b.com)"
    read [empty]() https://x.com/y #shop → url[15,30) tag[31,36)
    [a]()[b]()  https://c.com        → url[12,25) only

And I confirmed your sabotage independently — restoring the old `if (url)` guard gives exactly **2 FAILED / 203 SUCCESS**. The two new specs are load-bearing.

## Verification on `020ca047c3`

- Full suite both TZ variants: **13382 (Berlin) / 13368 (LA)** — your 13368 matches
- `checkFile` clean on all three changed files, reformatted nothing
- CI **17/17** green on the exact head SHA

No remaining findings. Merging this.

#9134 is closed as landed-here — its commit `fd2db51e8c` is in this branch, and I verified all three of its round-2 items separately (the note is on that PR). Thanks for the turnaround on both.

johannesjo:

PR #9137, opened 2026-07-18
feat(add-task-bar): opt-in natural language dates without trigger char (@-free)

KamilDev:
Last of the three PRs from the concept in https://github.com/super-productivity/super-productivity/issues/4961#issuecomment-5009474127.

> **Stack:** #9134 ⊂ #9135 ⊂ #9137, all targeting `master`; merge in order. This branch contains #9134 and #9135 — review only the natural-language commits here until those merge.

## What

Opt-in, Todoist-style natural language dates in the Add Task Bar — no trigger character:

| Input | Result |
| --- | --- |
| `Call mom tomorrow` | task "Call mom", due tomorrow |
| `Standup at 9am` | task "Standup", due at 9:00 |
| `Shopping every friday` | task "Shopping", weekly on Friday |

New checkbox in Settings → Short Syntax: **"Detect dates in plain text without @"** — **off by default**, so nothing changes for anyone who doesn't opt in. It only shows when the planned-at syntax is enabled, since it's a layer on top of it.

## False-positive handling (the reason this is PR 3 of 3)

Plain-text detection will sometimes be wrong ("Prepare Friday demo"). The design makes wrong detections visible and cheap to undo:

- Detected text is **highlighted inline** in the input (#9135) and shown on the date/repeat buttons — you see exactly what will be stripped from the title before submitting.
- **Dismissing** the detection via the date/repeat clear button keeps the word in the title and remembers the dismissal for the rest of the composing session, so it doesn't re-trigger on the next keystroke. Dismissing `every friday` also blocks the contained `friday` from popping back up as a plain date.
- Explicit `@` syntax always wins over plain-text detection.
- Bare numbers never match ("Buy 5 apples" stays untouched) — a chrono match must contain a letter or date separator.

## Scope guards

- **Add Task Bar only.** Title edits of existing tasks are never rewritten based on plain words — the surface gate is separate from the user setting, so no other `shortSyntax()` caller can turn this on by accident.
- English only for now, same as the existing `@` parsing (chrono's `casual` locale).

## Implementation notes

- The natural-language layer runs only when no explicit `@` due syntax matched, and reuses the recurrence machinery from #9134 (candidate phrases are located loosely, then re-validated with the same anchored grammar, so "every thing tomorrow" correctly falls through to just "tomorrow").
- `shortSyntax()`'s grown positional tail (`mode`, `isParseRepeat`) is converted into an options object in the same change.

## Tests

- 10 new cases in `short-syntax.spec.ts`: plain dates/times, recurrence, @-precedence, suppression incl. containment, number guard, config-off and surface-off gates.
- 3 new parser-service cases for the dismissal flow (suppress on NL, never on `@`).
- Wiki: `3.04-Short-Syntax.md` (new syntax form section) and `3.02-Settings-and-Preferences.md` (Short Syntax settings block) updated.


johannesjo:
Hi @KamilDev — I re-tested this one on the rebased branch before deciding what to do with it, and I'm going to close it. I explained my reasoning in [the issue comment](https://github.com/super-productivity/super-productivity/issues/4961#issuecomment-5010997120) and it hasn't changed, so the main reason I'm closing rather than leaving it open is that I don't want you merging #9135 up into this branch every few days for something I'm not going to take. That's on me for leaving it sitting.

To be clear about what this is and isn't: the code is not the problem. I went through the natural-language commits properly this time — the surface gate is genuinely tight (I checked every `shortSyntax()` caller; only the add-task bar passes the flag), the config field is optional and reads falsy on older clients, the docs are updated, CI is green, and every claim in your PR description holds up when I actually test it rather than take your word for it. The `@`-wins-over-plain-text precedence and the containment rule for dismissing `every friday` are both implemented exactly as you described them.

It's the concept I can't take. Two things I ran into while re-testing, which I think illustrate why better than my original comment did:

**The number guard lets through the separators that show up in real task titles.** `/[a-z]|[/:-]/i` correctly stops `Buy 5 apples`, but `/` and `:` are exactly what's in issue and PR references:

| input | resulting title | due |
| --- | --- | --- |
| `Fix issue 3/4 today` | `Fix issue  today` | **March 4** |
| `Review PR 2/3 changes` | `Review PR  changes` | Feb 3 |
| `Merge branch 1/2 of the stack` | `Merge branch  of the stack` | **Jan 2 2025** |
| `Set timer 10:30 for standup` | `Set timer  for standup` | today 10:30 |

The first row is the one that stuck with me. The user typed "today", and `3/4` sets the due date to March 4 while the word "today" stays in the title. The wrong date wins, the right word survives, and nothing on screen looks broken — so it gets noticed later, if at all.

**Dismissing a false positive can surface a different one.** Because the parser falls through to the next chrono candidate:

```
Email Friday about the march contract
  → "Friday" detected, title "Email  about the march contract", due Jan 19
  dismiss → "march" detected, title "Email Friday about the  contract", due Mar 1
  dismiss → clean
```

It does converge, so it's not a loop — but it means one ordinary sentence can cost several dismissals, each with a different wrong date.

And the part I keep coming back to: with `@`, the blast radius is bounded to text the user deliberately marked. Without a trigger it's every word of every title, decided by a third-party library's heuristics — and we take `chrono-node` on a `^2.9.0` range, so a minor bump can quietly change what gets deleted from people's task titles. That's not something I want to own on the most-used input in the app, and no amount of tuning the guard changes that shape.

None of this is wasted, though — #9134 is the piece I want, and I see you've derived the removal regex from the parser grammar there, which was exactly the right fix. Thanks for taking the review notes seriously across all three of these, and sorry this one ends as a no.

PR #9145, opened 2026-07-18
fix(theme): tolerate work contexts persisted without a theme (#9139)

### LINKED ISSUE #9139: 💥 Cannot read properties of undefined (reading 'backgroundImageDark')
### Steps to Reproduce

1.
2.
3.

### URL
file:///app/superproductivity/resources/app.asar/.tmp/angular-dist/browser/index.html#/tag/TODAY/tasks

### Meta Info
META: SP18.14.0FP __ Electron – en-GB – Linux x86_64 – en-GB – UA:Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) SuperProductivity/18.14.0 Chrome/146.0.7680.216 Electron/41.4.0 Safari/537.36

### Stacktrace
```
backgroundImageDark (file:///app/superproductivity/resources/app.asar/.tmp/angular-dist/browser/src/app/core/theme/global-theme.service.ts:129:19)
resolveBackground (file:///app/superproductivity/resources/app.asar/.tmp/angular-dist/browser/src/app/core/theme/global-theme.service.ts:234:6)
map (file:///app/superproductivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/rxjs/dist/esm/internal/operators/map.js:7:30)
function (file:///app/superproductivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/rxjs/dist/esm/internal/operators/OperatorSubscriber.js:12:8)
next (file:///app/superproductivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/rxjs/dist/esm/internal/Subscriber.js:29:11)
source.subscribe(createOperatorSubscriber (file:///app/superproductivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/rxjs/dist/esm/internal/observable/combineLatest.js:43:25)
function (file:///app/superproductivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/rxjs/dist/esm/internal/operators/OperatorSubscriber.js:12:8)
next (file:///app/superproductivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/rxjs/dist/esm/internal/Subscriber.js:29:11)
comparator (file:///app/superproductivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/rxjs/dist/esm/internal/operators/distinctUntilChanged.js:14:19)
function (file:///app/superproductivity/resources/app.asar/.tmp/angular-dist/browser/node_modules/rxjs/dist/esm/internal/operators/OperatorSubscriber.js:12:8)
```

### Actions Before Error
```
1784361467788: [Task] SetSelectedTask 
1784361468015: [SP_ALL] Load(import) all data 
1784361468023: [AppState] Set Today String 
1784361468026: [MenuTree] Update Tag Tree 
1784361468042: [Tag] Update Tag 
1784361468043: [MenuTree] Update Tag Tree (2) 
1784361468149: [SP_ALL] All Data was loaded
```

johannesjo:
Fixes #9139

## The crash

A tag/project entity can reach the store with **no `theme` at all**. `WorkContextCommon.theme` is declared required, but validation at hydration is non-fatal — a snapshot holding a theme-less tag loads anyway, and every consumer then dereferences `undefined`:

```
backgroundImageDark (global-theme.service.ts:129)
resolveBackground   (global-theme.service.ts:234)
```

The reported entity is **TODAY**, which is the active work context at startup — so this crashes the app on *every launch*, not just on some navigation. Reported on 18.14.0 (Electron/Linux).

## The fix

Four writers/readers touch a work-context theme. Three now share one source of truth, `getDefaultWorkContextTheme(type, entityId)`:

| Site | Change |
| --- | --- |
| `resolveContextTheme` (read side, `currentTheme$`) | extracted from the inline `map()`, now falls back instead of passing `undefined` through |
| `autoFixTypiaErrors` (on-disk heal) | new branch backfills a nullish `theme` |
| `getProjectVisibilityIconColor` + `note.component.html` | `theme?.primary` — these iterate *all* projects, so `currentTheme$` never covers them |

The read side and the heal **must not diverge**: hydration validates without repairing, so a local-only user can render the read-side value for many sessions before a repair ever runs. If the two disagreed, a theme-less context would render one color and then silently flip once a sync repair landed.

`getDefaultWorkContextTheme` is **id-aware, not just type-aware**. System entities (TODAY, URGENT, IMPORTANT, INBOX) ship distinct themes, so healing TODAY to the generic tag default would permanently restyle it (cornflower + tint-disabled → purple + tint-enabled) — and the heal is written to disk. The lookup is a `Map` rather than an object literal because an entity id of `__proto__` resolves to `Object.prototype` on a plain-object lookup, which is non-nullish and would defeat the `??` fall-through.

## Deliberate non-changes

- **`lwwUpdateMetaReducer`'s recreate branch is a fourth writer and is NOT routed through the shared helper.** It spreads `RECREATE_FALLBACK[type].defaults`, which is id-blind, so a system entity recreated from a partial LWW update still gets the generic theme persisted. It needs a delete-vs-update race on a system entity to hit, and that file is sync-critical — it deserves its own change with its own convergence analysis. Documented as a KNOWN RESIDUAL in the helper.
- **`theme` added to `RECREATE_FALLBACK.{TAG,PROJECT}.requiredKeys`** even though it drives no backfill there. The diagnostic warn is gated on `partialKeys.length > 0`, so omitting it would silence the warn *entirely* for a payload carrying `title`+`taskIds` but no theme — exactly the #9139 corruption shape, on the one writer that still bypasses the shared helper. **Provenance of the corruption is still unknown**, so this is the last signal that path touched a theme-less entity.
- **`IN_PROGRESS_TAG` is absent from `SYSTEM_ENTITY_THEMES`.** Its theme differs from `DEFAULT_TAG`'s only in `backgroundImageDark` (`''` vs `null`), and `isBackgroundImageSet` treats both as unset — the row could not change any rendered output, so it would be inert. A test enforces that every row in the map *is* observably different, so an inert row can't be added back unnoticed.

## Sync-correctness notes

- No schema bump, no new op semantics, no wire-format change. The heal is a client-local repair of already-invalid data.
- The heal is **idempotent and convergent**: the value written depends only on `(entityType, entityId)`, both of which are stable, so two clients repairing the same theme-less entity independently write byte-identical themes. No new conflict surface.
- The read-side tag-color override sits *on top of* the fallback and is read-side only, so it is re-applied after any repair and does not flip.

## Tests

31 assertions added across 4 spec files. Notable:

- The heal branch matches on **path + nullish value only, never `error.expected`** — typia reports this as the generated name `Readonly<__type>.oNN`, whose ordinal shifts whenever the type graph changes. A test asserts the fix still fires under a renamed type, guarding the #9045 failure mode (a check that shipped and never fired).
- The `null`-theme case is covered separately from `undefined`: the `setOne` `'replace'` branch applies a remote entity verbatim, so `theme: null` is reachable, and gating on `undefined` alone dead-ended the repair pipeline ("state still invalid after repair").
- The system-entity cases are **driven from `SYSTEM_ENTITY_THEMES` itself**, not a hand-written copy — a duplicated table only catches rows *removed* from the map, never rows *added* to it.
- The `__proto__` fixture is built via `JSON.parse`, the only construction that yields a genuine own `__proto__` key (`obj['__proto__'] = x` invokes the inherited setter and sets the prototype instead, creating no own property).

All green: `work-context.service.spec.ts` 32/32, `auto-fix-typia-errors.spec.ts` 24/24, `lww-update.meta-reducer.spec.ts` 133/133.

## Known gap

The **root cause — how an entity loses its `theme` in the first place — is still unknown.** This PR stops the crash and heals the data; it does not close the hole that produces it. The `requiredKeys` warn above is the deliberate breadcrumb for that follow-up.

### DIFF
--- a/e2e/tests/migration/legacy-themeless-tag-9139.spec.ts
+++ b/e2e/tests/migration/legacy-themeless-tag-9139.spec.ts
@@ -1,0 +1,159 @@
+import { test, expect, Page } from '@playwright/test';
+import legacyData from '../../fixtures/legacy-full-migration-backup.json';
+import { MIGRATION_BACKUP_PREFIX } from '../../../electron/shared-with-frontend/get-backup-timestamp';
+import { skipOnboardingForE2E } from '../../utils/waits';
+
+/**
+ * Issue #9139: a tag persisted with no `theme` at all crashed the app on
+ * EVERY launch — `resolveBackground()` dereferenced `theme.backgroundImageDark`
+ * on undefined, and the reported entity was TODAY, the active context at
+ * startup.
+ *
+ * The unit tests pin the pieces (the fallback, the heal, the `currentTheme$`
+ * wiring). None of them can show that the app actually STARTS with this data
+ * on disk, which is the only claim the bug report makes. That is what this
+ * test is for, so it deliberately asserts boot-and-render rather than any
+ * particular colour.
+ *
+ * SCOPE: this covers the on-disk heal. With the heal in place the data is
+ * repaired during migration, so the read-side fallback in `resolveContextTheme`
+ * is never reached here — verified by removing it and watching this test stay
+ * green. That path is covered by the `currentTheme$` unit test instead; an
+ * end-to-end version needs to corrupt an already-migrated store, which fights
+ * the running app's own IndexedDB connection and is not worth the flake.
+ *
+ * Run: npm run e2e:file e2e/tests/migration/legacy-themeless-tag-9139.spec.ts -- --retries=0
+ */
+
+/** Read the migrated store back out of SUP_OPS. */
+const readMigratedState = async (
+  page: Page,
+): Promise<{ tag?: { ids: string[]; entities: Record<string, unknown> } }> =>
+  page.evaluate(
+    async () =>
+      new Promise((resolve, reject) => {
+        const request = indexedDB.open('SUP_OPS');
+        request.onsuccess = (event) => {
+          const db = (event.target as IDBOpenDBRequest).result;
+          const tx = db.transaction('state_cache', 'readonly');
+          const getReq = tx.objectStore('state_cache').get('current');
+          getReq.onsuccess = () => {
+            db.close();
+            resolve(getReq.result?.state || {});
+          };
+          getReq.onerror = () => {
+            db.close();
+            reject(getReq.error);
+          };
+        };
+        request.onerror = () => reject(request.error);
+      }),
+  );
+
+const seedLegacyDatabase = async (
+  page: Page,
+  data: Record<string, unknown>,
+): Promise<void> => {
+  await page.evaluate(
+    async (entityData) =>
+      new Promise<void>((resolve, reject) => {
+        const request = indexedDB.open('pf', 1);
+        request.onupgradeneeded = (event) => {
+          const db = (event.target as IDBOpenDBRequest).result;
+          if (!db.objectStoreNames.contains('main')) {
+            db.createObjectStore('main');
+          }
+        };
+        request.onsuccess = (event) => {
+          const db = (event.target as IDBOpenDBRequest).result;
+          const tx = db.transaction('main', 'readwrite');
+          const store = tx.objectStore('main');
+          for (const [key, value] of Object.entries(entityData)) {
+            store.put(value, key);
+          }
+          tx.oncomplete = () => {
+            db.close();
+            resolve();
+          };
+          tx.onerror = () => {
+            db.close();
+            reject(tx.error);
+          };
+        };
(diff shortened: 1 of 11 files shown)


johannesjo:
Follow-up filed as #9156: the settings dialog spreads `{ ...entity.theme }`, which yields `{}` on a themeless entity and is persisted unconditionally — including via `cancelEdit()`. `{}` is non-nullish, so it slips past both halves of this PR (the `??` fallback and the nullish-gated heal), turning a self-healing shape into a permanent one. Deliberately out of scope here.

Issue #8939, opened 2026-07-11
fix(sync): audit remaining writers on the clock-overwriting append primitive


johannesjo:
From the #8900 review (architecture C1 audit note, alternatives W1).

`appendWithVectorClockUpdate` REPLACES the durable vector clock with the caller-built clock. #8900 moved LWW compensations and (f572b6d077) disjoint merges onto `appendMixedSourceBatchSkipDuplicates`, whose in-transaction rebase makes counter reuse/regression unrepresentable. Two synthetic-op writers remain on the overwrite primitive:

- `superseded-operation-resolver.service.ts` (~:279) — loop of `appendWithVectorClockUpdate`
- `repair-operation.service.ts` (~:85)

(The capture path `operation-log.effects.ts` is safe — it derives from the durable clock under the capture lock.)

Audit both: if their clocks are not provably derived from the durable clock at write time, route them through the mixed-source batch. Consider afterwards whether `appendWithVectorClockUpdate` can be deleted or renamed to something that screams "overwrites the durable clock".

PR #9158, opened 2026-07-18
fix(op-log): use append()'s returned seq instead of re-reading getLastSeq() (#8337)

### LINKED ISSUE #8337: Op-log: append() seq discarded and re-read via getLastSeq() — multi-tab race can skip ops in tail replay
**Area:** client (op-log persistence/hydration) · **Severity:** medium (web multi-tab only; silent op skip in tail replay)

## Problem
`OperationLogStoreService.append()` returns the new seq, but three callers discard it and immediately re-read `getLastSeq()` (a separate reverse-cursor IndexedDB transaction). None of these paths holds the `sp_op_log` Web Lock, while local-op capture in **other tabs** appends under it — so another tab's append can land between `append()` and `getLastSeq()`. The re-read seq then points *past* that op, `state_cache.lastAppliedOpSeq` is persisted too high, and the next hydration's tail replay (`getOpsAfterSeq(snapshot.lastAppliedOpSeq)`) silently skips the op.

## Evidence
- `append()` returns the seq: `operation-log-store.service.ts:367-386`.
- Discard + re-read: `sync-hydration.service.ts:194/198` and `:228`; `operation-log-migration.service.ts:279-280`; `operation-log-recovery.service.ts:114-117` — each feeds `lastAppliedOpSeq` of `saveStateCache` (sync-hydration `:286`, migration `:284`, recovery `:120`).
- Tail replay: `operation-log-hydrator.service.ts:210`.
- Lock context: capture appends under `LOCK_NAMES.OPERATION_LOG` (`operation-log.effects.ts:305`; `lock.service.ts:64`), but the three sites' callers (`sync-wrapper.service.ts:470/995/1321`, `ws-triggered-download.service.ts:96`, `data-init.service.ts:69`) hold only `sp_op_log_download`/`sp_op_log_upload` — different lock names, no mutual exclusion.
- Scope notes: Electron/Android use a single-instance in-process mutex, so this is web multi-tab only. Migration/recovery run at startup with tiny windows; the realistic exposure is the sync-hydration paths. In the SYNC_IMPORT branch a concurrent op would likely be dropped by `SyncImportFilterService` anyway (clean-slate semantics); the **no-import branch (`:228`) explicitly intends to preserve concurrent work, making it the worst case** — and it has no preceding append at all (the race there is `getUnsynced()`→`getLastSeq()`), so the returned-seq fix alone doesn't cover it.

## Proposed fix
Use `append()`'s returned seq at `sync-hydration:194`, `migration:279`, `recovery:114`. For `sync-hydration:228`, wrap the reject+snapshot section in `lockService.request(LOCK_NAMES.OPERATION_LOG, …)` (or derive the seq atomically); taking the lock also hardens the other sites. Bonus: removes a redundant cursor transaction per call site.

mvanhorn:
## Problem

Fixes #8337.

Two callers wrote an operation and then persisted a replay anchor derived from a *separate* read, so a concurrent tab's append could land in between:

- `OperationLogMigrationService._performMigration()` — the genesis write.
- `SyncHydrationService.hydrateFromRemoteSync()` — the SYNC_IMPORT branch.

Neither held the `sp_op_log` Web Lock that other tabs' capture appends under. In the **web multi-tab** case the anchor could be persisted past a concurrent op, and the next boot's tail replay (`getOpsAfterSeq(snapshot.lastAppliedOpSeq)`) silently skipped it. Electron and Android are unaffected (single-instance in-process mutex).

## Solution

Two parts, both at the persistence layer.

**1. An atomic append-plus-anchor helper.** `OperationLogStoreService.appendOperationAndSnapshot()` writes the operation, the state-cache anchor and the working vector clock in a single IndexedDB transaction, so the snapshot frontier and working clock can no longer move independently of the operation. Inside that transaction the proposed clock is rebased on the durable clock, which keeps a concurrent tab's later append replayable and stops its clock advancement being overwritten by a follow-up write.

The rebased clock is then re-bounded to `MAX_VECTOR_CLOCK_SIZE`. Pruning happens before the transaction opens, so unioning a bounded durable clock with a bounded proposed clock could otherwise produce twice the limit; the bound preserves the current client and the latest full-state author, matching `pruneClockForStorage`. This is reachable on the live migration path because `clearAllOperations()` intentionally retains `VECTOR_CLOCK`.

**2. A migration barrier.** `OperationLogMigrationService` holds `LOCK_NAMES.OPERATION_LOG` across the genesis decision and commit, so a concurrent tab cannot interleave an append between the orphan check and the genesis write.

## Verification

- `src/app/op-log/**` suite: 3654 passing.
- Regression for the clock bound drives the real store with a disjoint 20 + 20 case; without the bound it reports 40 entries in the stored operation, the committed clock and the state cache.
- Regressions for the migration ordering replay a second-tab append after restart and queue one behind the genesis anchor.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have included relevant changes to the documentation/wiki — not applicable (no user-facing or documented behavior change).
- [ ] I have run `npm run checkFile` on changed `.ts`/`.scss` files — the full Angular toolchain was not available in my environment; CI runs prettier/lint. The changes follow the surrounding 2-space/prettier style.
- [x] I have added tests for my changes (if applicable) — added a regression test in `sync-hydration.service.spec.ts` that distinguishes `append()`'s returned seq from a concurrent-append-inflated `getLastSeq()` and asserts the persisted `lastAppliedOpSeq` uses the former. Note: the migration site mirrors the identical pattern but its production path (`_performMigration`, which uses dynamic `import()` of the validation pipeline) is stubbed in the existing spec and so is covered by inspection rather than a dedicated new test.
- [ ] Existing tests still pass — not run locally (toolchain unavailable); validated by CI.

### DIFF
--- a/src/app/core/data-init/data-init.service.ts
+++ b/src/app/core/data-init/data-init.service.ts
@@ -51,23 +51,4 @@ export class DataInitService {
     // Hydrate from Operation Log (which handles migration from legacy if needed)
     await this._operationLogHydratorService.hydrateStore();
   }
-
-  /**
-   * Re-initialize the app after a remote sync download.
-   * This uses hydrateFromRemoteSync() which:
-   * 1. Uses the downloaded mainModelData (passed from sync service)
-   * 2. Persists it to SUP_OPS as a SYNC_IMPORT operation
-   * 3. Creates a snapshot for crash safety
-   * 4. Updates NgRx with the synced data
-   *
-   * @param downloadedMainModelData - The main model data from the remote meta file.
-   *   Entity models are not stored in IndexedDB, so this must be passed explicitly.
-   */
-  async reInitFromRemoteSync(
-    downloadedMainModelData?: Record<string, unknown>,
-  ): Promise<void> {
-    await this._operationLogHydratorService.hydrateFromRemoteSync(
-      downloadedMainModelData,
-    );
-  }
 }
--- a/src/app/imex/sync/sync-wrapper.service.spec.ts
+++ b/src/app/imex/sync/sync-wrapper.service.spec.ts
@@ -183,9 +183,7 @@ describe('SyncWrapperService', () => {
     mockTranslateService = jasmine.createSpyObj('TranslateService', ['instant']);
     mockTranslateService.instant.and.callFake((key: string) => key);
 
-    mockDataInitService = jasmine.createSpyObj('DataInitService', [
-      'reInitFromRemoteSync',
-    ]);
+    mockDataInitService = jasmine.createSpyObj('DataInitService', ['reInit']);
     mockReminderService = jasmine.createSpyObj('ReminderService', ['reloadFromDatabase']);
 
     mockUserInputWaitState = jasmine.createSpyObj(
--- a/src/app/op-log/persistence/operation-log-hydrator.service.spec.ts
+++ b/src/app/op-log/persistence/operation-log-hydrator.service.spec.ts
@@ -2102,34 +2102,6 @@ describe('OperationLogHydratorService', () => {
     });
   });
 
-  describe('hydrateFromRemoteSync', () => {
-    it('should delegate to syncHydrationService', async () => {
-      await service.hydrateFromRemoteSync();
-
-      expect(mockSyncHydrationService.hydrateFromRemoteSync).toHaveBeenCalled();
-    });
-
-    it('should pass downloadedMainModelData to syncHydrationService', async () => {
-      const downloadedData = { task: { entities: {}, ids: [] } };
-
-      await service.hydrateFromRemoteSync(downloadedData);
-
-      expect(mockSyncHydrationService.hydrateFromRemoteSync).toHaveBeenCalledWith(
-        downloadedData,
-        undefined,
-      );
-    });
-
-    it('should pass undefined when no downloadedMainModelData provided', async () => {
-      await service.hydrateFromRemoteSync();
-
-      expect(mockSyncHydrationService.hydrateFromRemoteSync).toHaveBeenCalledWith(
-        undefined,
-        undefined,
-      );
-    });
-  });
-
   // ===========================================================================
   // retryFailedRemoteOps: Retry failed remote operations
   // ===========================================================================
--- a/src/app/op-log/persistence/operation-log-hydrator.service.ts
+++ b/src/app/op-log/persistence/operation-log-hydrator.service.ts
@@ -932,25 +932,6 @@ export class OperationLogHydratorService {
   }
 
   /**
-   * Handles hydration after a remote sync download.
-   * Delegates to SyncHydrationService.
-   *
-   * @param downloadedMainModelData - Entity models from remote meta file.
-   *   These are NOT stored in IndexedDB (only archives are) so must be passed explicitly.
-   * @param remoteVectorClock - Vector clock from the downloaded snapshot.
-   *   Merged into the SYNC_IMPORT's clock to prevent mutual discarding during provider switch.
-   */
-  async hydrateFromRemoteSync(
-    downloadedMainModelData?: Record<string, unknown>,
-    remoteVectorClock?: Record<string, number>,
-  ): Promise<void> {
-    return this.syncHydrationService.hydrateFromRemoteSync(
-      downloadedMainModelData,
-      remoteVectorClock,
-    );
-  }
-
-  /**
    * Validates a state object during hydration without attempting repair.
    *
    * Repair is intentionally not run here: it requires a native `confirm()` dialog
--- a/src/app/op-log/persistence/operation-log-migration.service.spec.ts
+++ b/src/app/op-log/persistence/operation-log-migration.service.spec.ts
@@ -30,10 +30,7 @@ describe('OperationLogMigrationService', () => {
       'getOpsAfterSeq',
       'deleteOpsWhere',
       'clearAllOperations',
-      'append',
-      'getLastSeq',
-      'saveStateCache',
-      'setVectorClock',
+      'appendOperationAndSnapshot',
     ]);
 
     mockLegacyPfDb = jasmine.createSpyObj('LegacyPfDbService', [
@@ -394,16 +391,12 @@ describe('OperationLogMigrationService', () => {
               schemaVersion: CURRENT_SCHEMA_VERSION,
             };
 
-            await mockOpLogStore.append(migrationOp as any);
-            const lastSeq = await mockOpLogStore.getLastSeq();
-            await mockOpLogStore.saveStateCache({
+            await mockOpLogStore.appendOperationAndSnapshot(migrationOp as any, 'local', {
               state: legacyData,
-              lastAppliedOpSeq: lastSeq,
               vectorClock: migrationOp.vectorClock,
               compactedAt: Date.now(),
               schemaVersion: CURRENT_SCHEMA_VERSION,
             });
-            await mockOpLogStore.setVectorClock(migrationOp.vectorClock);
             mockStore.dispatch(loadAllData({ appDataComplete: legacyData as any }));
 
             dialogRef.componentInstance.status.set('complete');
@@ -411,10 +404,7 @@ describe('OperationLogMigrationService', () => {
         );
 
         // Mock opLogStore methods used during migration
-        mockOpLogStore.append.and.resolveTo(1);
-        mockOpLogStore.getLastSeq.and.resolveTo(1);
-        mockOpLogStore.saveStateCache.and.resolveTo();
-        mockOpLogStore.setVectorClock.and.resolveTo();
+        mockOpLogStore.appendOperationAndSnapshot.and.resolveTo(1);
       });
 
       it('should call persistClientId with legacy client ID when it exists', async () => {
@@ -430,7 +420,7 @@ describe('OperationLogMigrationService', () => {
           'legacyClientId1234',
         );
         expect(mockClientIdService.getOrGenerateClientId).not.toHaveBeenCalled();
-        expect(mockOpLogStore.append).toHaveBeenCalled();
+        expect(mockOpLogStore.appendOperationAndSnapshot).toHaveBeenCalled();
       });
 
       it('should generate new client ID and NOT call persistClientId when legacy ID is null', async () => {
@@ -442,7 +432,7 @@ describe('OperationLogMigrationService', () => {
 
         expect(mockClientIdService.getOrGenerateClientId).toHaveBeenCalled();
         expect(mockClientIdService.persistClientId).not.toHaveBeenCalled();
-        expect(mockOpLogStore.append).toHaveBeenCalled();
+        expect(mockOpLogStore.appendOperationAndSnapshot).toHaveBeenCalled();
       });
     });
   });
--- a/src/app/op-log/persistence/operation-log-migration.service.ts
+++ b/src/app/op-log/persistence/operation-log-migration.service.ts
@@ -275,25 +275,16 @@ export class OperationLogMigrationService {
       schemaVersion: CURRENT_SCHEMA_VERSION,
     };
 
-    // 5. Persist to operation log and use the seq append() returns as this op's
-    // own sequence number, rather than re-reading getLastSeq() (a separate
-    // reverse-cursor read that returns the GLOBAL op-log tail). A concurrent
-    // tab's append could push that tail past this op — this path does not hold
-    // the sp_op_log Web Lock other tabs' capture appends under — so the re-read
-    // seq would persist lastAppliedOpSeq too high and the next boot's tail
-    // replay would silently skip the concurrent op. append()'s return is
-    // exactly this op's seq. (#8337)
-    const lastSeq = await this.opLogStore.append(migrationOp);
-
-    await this.opLogStore.saveStateCache({
+    // 5. Persist the genesis operation, its exact snapshot frontier, and the
+    // working clock in one transaction. There is no post-append interval where
+    // a later tab write can be skipped by the snapshot frontier or have its
+    // clock advancement overwritten by a follow-up migration write.
+    await this.opLogStore.appendOperationAndSnapshot(migrationOp, 'local', {
       state: dataToMigrate,
-      lastAppliedOpSeq: lastSeq,
       vectorClock: migrationOp.vectorClock,
       compactedAt: Date.now(),
       schemaVersion: CURRENT_SCHEMA_VERSION,
     });
-
-    await this.opLogStore.setVectorClock(migrationOp.vectorClock);
 
     // 6. Dispatch to NgRx store
     this.store.dispatch(loadAllData({ appDataComplete: dataToMigrate }));
--- a/src/app/op-log/persistence/operation-log-migration.service.ts
+++ b/src/app/op-log/persistence/operation-log-migration.service.ts
@@ -275,9 +275,15 @@ export class OperationLogMigrationService {
       schemaVersion: CURRENT_SCHEMA_VERSION,
     };
 
-    // 5. Persist to operation log
-    await this.opLogStore.append(migrationOp);
-    const lastSeq = await this.opLogStore.getLastSeq();
+    // 5. Persist to operation log and use the seq append() returns as this op's
+    // own sequence number, rather than re-reading getLastSeq() (a separate
+    // reverse-cursor read that returns the GLOBAL op-log tail). A concurrent
+    // tab's append could push that tail past this op — this path does not hold
+    // the sp_op_log Web Lock other tabs' capture appends under — so the re-read
+    // seq would persist lastAppliedOpSeq too high and the next boot's tail
+    // replay would silently skip the concurrent op. append()'s return is
+    // exactly this op's seq. (#8337)
+    const lastSeq = await this.opLogStore.append(migrationOp);
 
     await this.opLogStore.saveStateCache({
       state: dataToMigrate,
--- a/src/app/op-log/persistence/operation-log-migration.service.ts
+++ b/src/app/op-log/persistence/operation-log-migration.service.ts
@@ -25,6 +25,8 @@ import {
   MIGRATION_BACKUP_PREFIX,
   getBackupTimestamp,
 } from '../../../../electron/shared-with-frontend/get-backup-timestamp';
+import { LockService } from '../sync/lock.service';
+import { LOCK_NAMES } from '../core/operation-log.const';
 
 /**
  * Service to check for valid operation log state during startup and migrate
@@ -44,6 +46,7 @@ export class OperationLogMigrationService {
   private store = inject(Store);
   private languageService = inject(LanguageService);
   private translateService = inject(TranslateService);
+  private lockService = inject(LockService);
 
   /**
    * Checks if the operation log is in a valid state and migrates legacy data if found.
@@ -89,72 +92,89 @@ export class OperationLogMigrationService {
       throw e;
     }
 
-    // No snapshot exists. Check if there are any operations in the log.
-    const allOps = await this.opLogStore.getOpsAfterSeq(0);
-
-    if (allOps.length > 0) {
-      // Operations exist but no snapshot. Check if the first op is a Genesis/Migration op.
-      const firstOp = allOps[0].op;
-      if (firstOp.entityType === 'MIGRATION' || firstOp.entityType === 'RECOVERY') {
-        // Valid Genesis exists - migration already happened but snapshot might have been lost
-        OpLog.normal(
-          'OperationLogMigrationService: Genesis operation found. Skipping migration.',
-        );
-        return;
-      }
-
-      // Operations exist without Genesis. Behavior depends on whether legacy data exists:
-      if (hasLegacyData) {
-        // Case 1: Legacy data exists - these are orphan ops captured during app init
-        // before hydration. Clear them so migration can proceed cleanly.
-        OpLog.warn(
-          `OperationLogMigrationService: Found ${allOps.length} orphan operations. ` +
-            `Clearing them before legacy migration.`,
-        );
-        await this.opLogStore.clearAllOperations();
-      } else {
-        // Case 2: No legacy data - these are legitimate user operations from a fresh
-        // install. Let the hydrator replay them. No migration needed.
-        OpLog.normal(
-          `OperationLogMigrationService: Found ${allOps.length} operations (fresh install). ` +
-            `Skipping migration - hydrator will replay them.`,
-        );
-        return;
-      }
-    }
-    if (!hasLegacyData) {
-      OpLog.normal('OperationLogMigrationService: No legacy data found. Starting fresh.');
-      return;
-    }
-
-    // Acquire migration lock (prevent concurrent tab migrations)
-    const lockAcquired = await this.legacyPfDb.acquireMigrationLock();
-    if (!lockAcquired) {
-      OpLog.warn(
-        'OperationLogMigrationService: Migration lock held by another instance, skipping.',
-      );
-      return;
-    }
-
-    // Ensure translations are loaded before showing dialog
-    await this._ensureTranslationsLoaded();
-
-    // Show migration dialog and perform migration
-    const dialogRef = this._showMigrationDialog();
+    let migrationLockAcquired = false;
+    let dialogRef: MatDialogRef<DialogLegacyMigrationComponent> | undefined;
     try {
-      await this._createAutoBackup(dialogRef);
-      await this._performMigration(dialogRef);
+      const migrationCompleted = await this.lockService.request(
+        LOCK_NAMES.OPERATION_LOG,
+        async () => {
+          // Re-read both replay anchors after acquiring the barrier. Capture writes
+          // queued behind this lock must land after the genesis snapshot frontier.
+          const lockedSnapshot = await this.opLogStore.loadStateCache();
+          if (lockedSnapshot) {
+            return false;
+          }
+
+          const allOps = await this.opLogStore.getOpsAfterSeq(0);
+          if (allOps.length > 0) {
+            const firstOp = allOps[0].op;
+            if (firstOp.entityType === 'MIGRATION' || firstOp.entityType === 'RECOVERY') {
+              OpLog.normal(
+                'OperationLogMigrationService: Genesis operation found. Skipping migration.',
+              );
+              return false;
+            }
+
+            if (hasLegacyData) {
+              OpLog.warn(
+                `OperationLogMigrationService: Found ${allOps.length} orphan operations. ` +
+                  `Clearing them before legacy migration.`,
+              );
+              await this.opLogStore.clearAllOperations();
+            } else {
+              OpLog.normal(
+                `OperationLogMigrationService: Found ${allOps.length} operations (fresh install). ` +
+                  `Skipping migration - hydrator will replay them.`,
+              );
+              return false;
+            }
+          }
+          if (!hasLegacyData) {
+            OpLog.normal(
+              'OperationLogMigrationService: No legacy data found. Starting fresh.',
+            );
+            return false;
+          }
+
+          migrationLockAcquired = await this.legacyPfDb.acquireMigrationLock();
+          if (!migrationLockAcquired) {
+            OpLog.warn(
+              'OperationLogMigrationService: Migration lock held by another instance, skipping.',
+            );
+            return false;
+          }
+
+          await this._ensureTranslationsLoaded();
+          dialogRef = this._showMigrationDialog();
+          await this._createAutoBackup(dialogRef);
+          await this._performMigration(dialogRef);
+          return true;
+        },
+      );
+
+      if (migrationCompleted && dialogRef) {
+        this._setStatus(dialogRef, 'complete');
+        OpLog.normal('OperationLogMigrationService: Migration complete');
+
+        // Brief delay to show completion status. Keep this cosmetic wait outside
+        // the operation-log barrier so queued capture writes can proceed.
+        await new Promise((resolve) => setTimeout(resolve, 1000));
+      }
     } catch (error) {
       OpLog.err('OperationLogMigrationService: Migration failed:', error);
-      dialogRef.componentInstance.error.set(
-        'Migration failed. Your backup has been downloaded. Please restart or import the backup file.',
-      );
-      // Wait for user acknowledgment before throwing
-      await firstValueFrom(dialogRef.afterClosed());
+      if (dialogRef) {
+        dialogRef.componentInstance.error.set(
+          'Migration failed. Your backup has been downloaded. Please restart or import the backup file.',
+        );
+        // Wait for user acknowledgment before throwing
+        await firstValueFrom(dialogRef.afterClosed());
+      }
       throw error;
     } finally {
-      await this.legacyPfDb.releaseMigrationLock();
-      dialogRef.close();
+      if (migrationLockAcquired) {
+        await this.legacyPfDb.releaseMigrationLock();
+      }
+      dialogRef?.close();
     }
   }
 
@@ -288,12 +308,6 @@ export class OperationLogMigrationService {
 
     // 6. Dispatch to NgRx store
     this.store.dispatch(loadAllData({ appDataComplete: dataToMigrate }));
-
-    this._setStatus(dialogRef, 'complete');
-    OpLog.normal('OperationLogMigrationService: Migration complete');
-
-    // Brief delay to show completion status
-    await new Promise((resolve) => setTimeout(resolve, 1000));
   }
 
   private _setStatus(
--- a/src/app/op-log/persistence/operation-log-store.service.spec.ts
+++ b/src/app/op-log/persistence/operation-log-store.service.spec.ts
@@ -360,6 +360,19 @@ describe('OperationLogStoreService', () => {
       expect(await service.getVectorClock()).toEqual(op.vectorClock);
     });
 
+    it('should preserve the full recovery clock in the atomic replay anchor', async () => {
+      const vectorClock = createBloatedClock({ testClient: 1 });
+      const op = createTestOperation({
+        id: 'legacy-recovery-full-clock-op',
+        vectorClock,
+      });
+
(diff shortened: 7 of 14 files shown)


johannesjo (commented):
## Review: needs changes

Using the sequence returned by `append()` is correct for the narrow `lastAppliedOpSeq` race, but the same unlocked interval still violates the durable vector-clock invariant, and the added test does not reproduce the storage/replay failure required by this repository's sync rules.

### Important: the later clock write can overwrite a concurrent tab's advancement

At `sync-hydration.service.ts:248` and `operation-log-migration.service.ts:286`, ordinary `append()` writes the operation but does not update `VECTOR_CLOCK`. Both paths later call `setVectorClock()` separately (`sync-hydration.service.ts:339`, `operation-log-migration.service.ts:296`) and still do not hold `LOCK_NAMES.OPERATION_LOG`.

If another tab captures an action after the changed `append()` but before that later clock write, its normal capture path acquires `sp_op_log`, clears the per-tab cache, derives/increments the current clock, and atomically persists its operation plus the advanced clock. The original path can then overwrite that advancement with its previously computed import/migration clock. In the SYNC_IMPORT case this can reuse the import's current-client counter; in the migration case a post-genesis capture can advance beyond the genesis clock and then be regressed by line 296. This is exactly the counter-regression/reuse invariant documented around `appendWithVectorClockOverwrite()` (#8939), and can make later conflict/import classification causally incorrect. Returning the append sequence protects the snapshot cursor, but not this clock state.

Please either commit the operation, snapshot frontier, and working clock through one atomic store primitive (similar to recovery), or establish the full operation-log barrier and rebase from the fresh durable clock before committing without a later overwriting write.

### Important: the regression proof is still a mocked seam, and migration is not exercised

`sync-hydration.service.spec.ts:385` only configures `append() -> 7` and an unused `getLastSeq() -> 9`; it writes no operations, creates no second-tab interleaving, persists no real snapshot, and never verifies restart tail replay. It also duplicates the preceding mock-level assertion rather than proving the race.

The migration unit spec replaces `_performMigration()` wholesale and its fake still uses the old `append(); getLastSeq()` sequence (`operation-log-migration.service.spec.ts:368-405`). The separate `legacy-data-migration.integration.spec.ts` uses the real store for early-return/orphan cases but never acquires the migration lock successfully or executes `_performMigration()`. Reverting the production migration change would therefore leave all of this coverage green.

That does not meet the repository requirement that sync fixes start from a seeded-storage or scripted real-data reproduction rather than a mocked seam. Please add a real-adapter regression that interleaves seq N/N+1 across store instances, persists and reloads the cache, verifies N+1 tail replay, and verifies the durable clock does not regress; the actual migration persistence path should be included as well.

All current CI checks are green, but none exercises these interleavings. The live production exposure of the SYNC_IMPORT/default branch appears limited because current `OperationLogSyncService` callers pass `createSyncImportOp=false`; the legacy migration path remains a live, albeit rare, call site.

mvanhorn:
Reworked in 1cc6e73: op, snapshot frontier, and working clock now persist in a single transaction (new appendOperationAndSnapshot), used by both the sync-hydration and migration paths, so the unlocked interval can no longer overwrite a concurrent tab's clock advancement. The specs now include the storage/replay reproduction; all touched suites pass locally (303 specs green).

johannesjo (commented):
## Review round 2 (`1cc6e73`)

Thanks for the rework — the direction is right and this is a net improvement. I verified the changes against a real IndexedDB adapter rather than by inspection, and I want to lead with what's genuinely fixed before the two things that aren't.

### What's confirmed good

**The regression tests are real this time.** I reverted `appendOperationAndSnapshot()` to the pre-PR sequence (`append()` → `getLastSeq()` → `saveStateCache()` → `setVectorClock()`) while keeping the signature, and both new tests fail:

```
legacy-data-migration.integration.spec.ts   Expected 2 to be 1.  /  replay tail [] vs ['second-tab-after-migration']
sync-hydration.service.spec.ts              Expected 2 to be 1.  /  replay tail [] vs ['second-tab-after-sync-import']
```

That answers my round-2 ask: real store, real restart read-back, fails on revert. The migration integration test genuinely executes `_performMigration()` through `checkAndMigrate()` — only `_createAutoBackup` is stubbed. And using a second `OperationLogStoreService` instance on the same IndexedDB to get independent `_vectorClockCache`es is a nice way to model two tabs; better than `SimulatedClient` for this.

**The atomicity buys something real beyond the seq fix.** Previously a failure between `append()` and `saveStateCache()` left an orphan operation with no snapshot. Now it's all-or-nothing.

**Moving the SYNC_IMPORT append after `validateAndRepair` is safe**, and I checked why: this path calls the pure `validateAndRepair` (`validate-state.service.ts:287`), not `validateAndRepairCurrentState` — only the latter creates a REPAIR op. Nothing touches the op log in between, so the reordering is strictly better: a validation failure no longer strands an operation in the log.

**The recovery refactor is behaviour-preserving.** The recovery op is `OpType.Batch` (`operation-log-recovery.service.ts:124`), not full-state, so `_recordFullStateOpInTx()` no-ops and `META` stays out of the transaction store list. Pruning once instead of twice in `saveStateCache()` + `setVectorClock()` is an improvement.

---

### Blocking: the clock invariant is still violated — the PR description says otherwise

> so the unlocked interval can no longer overwrite a concurrent tab's clock advancement

The transaction closes only the window *after* the append. The clock **value** is still computed outside it — `sync-hydration.service.ts:190` (read) → `:320` (write), and `operation-log-migration.service.ts:273` → `:282` — and `_appendOperationAndSnapshot()` blindly `put`s it at `operation-log-store.service.ts:809-816` with no in-transaction rebase and no `sp_op_log` hold.

Reproduced by taking your own test in `sync-hydration.service.spec.ts` and moving `appendFromSecondTab()` to *before* `realAtomicAppend(...)`, with the concurrent op carrying `{localClient: 9}` (the capture path advanced the durable clock several times while hydration was computing):

```
lastAppliedOpSeq: 2
durableClock:     { localClient: 6 }   // regressed from 9
replayedOpIds:    []                   // concurrent op never replayed, and not in the snapshot state
```

The next capture then re-derives 7, 8, 9 — counters already carried by operations in the log. That's the invariant spelled out a few hundred lines above the new primitive, at `operation-log-store.service.ts:2762-2768`:

> Any other derivation (e.g. from the per-tab in-memory cache) can regress the durable clock and reuse counters. New writers must use `appendMixedSourceBatchSkipDuplicates`, whose in-transaction rebase makes regression unrepresentable.

`mergeRemoteOpClocks()` (`:2705-2725`) is the shape that satisfies it: read the durable clock *inside* the transaction, merge, then put.

**Important caveat on severity: this is pre-existing, not something you introduced.** `master` does the same blind put via `setVectorClock(clockForStorage)`. So the PR doesn't make anything worse. But the claim that it's fixed needs to come out of the description, because the tests are shaped to exercise only the interleaving that works, and a future reader will reasonably trust both.

### Blocking: the restart path overwrites the clock again anyway

Even with a perfectly atomic anchor, `operation-log-hydrator.service.ts:257` unconditionally does `setVectorClock(snapshot.vectorClock)` on the next boot, and the tail operations' clocks are only merged much later at `:790` via `mergeRemoteOpClocks()`. `hydrateStore()` holds **no lock at all** (no `LockService`/`LOCK_NAMES` reference anywhere in that file), and it's reached from `DataInitService.reInit()` and `plugin-bridge.service.ts:836`. So another tab's capture can read the regressed clock in that window and reuse a counter.

Also pre-existing and outside this diff — but it means the stated goal isn't reachable by making this one write atomic. Worth deciding whether that's in scope here or a separate issue.

---

### Scope: the SYNC_IMPORT branch has no production callers

`hydrateFromRemoteSync(createSyncImportOp = true)` is reachable only via `OperationLogHydratorService.hydrateFromRemoteSync` → `DataInitService.reInitFromRemoteSync`, and that has zero callers. Both live `OperationLogSyncService` sites (`:1384`, `:1980`) pass `false`.

This repo's own audit already reached the same conclusion and recommends deleting it — `docs/research/sync-simplification-audit/findings.md:1214-1232`:

> `reInitFromRemoteSync()` […] Repository-wide search finds no production caller. […] **Unnecessary mechanism / smallest change:** delete both forwarding methods, the hydrator dependency, their direct tests […]

So the larger half of this PR hardens a path scheduled for deletion. The legacy migration genesis is the only live call site — and there the seq fix is a real, small, correct improvement.

My suggestion: **keep the migration-genesis fix and drop the SYNC_IMPORT half.** That leaves a focused change against a live path, with the integration test that already proves it, and avoids investing in code the audit wants removed. If you'd rather keep both, that's fine too — but then the clock claim still needs correcting.

---

### Minor

- `operation-log-store.service.spec.ts:363` now *asserts* that recovery stores a 26-entry clock unpruned (`createBloatedClock` = `MAX_VECTOR_CLOCK_SIZE + 5`). That cements the one durable-clock write bypassing `pruneClockForStorage()`, contradicting that method's own docstring ("Every durable-clock write in this store routes through this method"). Pre-existing behaviour, but pinning it with a test makes it a contract. Either prune it or add a comment saying why recovery is exempt.
- `operation-log-migration.service.spec.ts:365` still replaces `_performMigration()` with a hand-written copy, now updated to mirror the new shape. It asserts a mock called a mock, so it can't fail on a production regression, and it now carries a standing obligation to be re-synced by hand whenever production changes. Since the integration test covers the real path, consider deleting it rather than maintaining a decoy.
- The `should commit the SYNC_IMPORT, snapshot, and working clock atomically` test asserts against a spy on a fully mocked store, so it can only observe *which* method was called — it can't distinguish one transaction from three. The name promises more than it checks.
- `sync-hydration.service.ts:318` and `:331`: both `throw new Error(...)` guards are unreachable — the same never-reassigned `createSyncImportOp` gates the assignment and the use. Pure TypeScript-narrowing scaffolding; a small restructure (build and commit the op in the same branch) would remove both.
- `operation-log-store.service.ts` grows 3046 → 3087 lines. It's a grandfathered `max-lines` offender (`eslint.config.js:317`), and that list is meant to shrink.
- Heads-up on merge order: #8389 (native SQLite op-log backend) rewrites this same service. The new four-store transaction (`OPS`/`STATE_CACHE`/`VECTOR_CLOCK`/`META`) will need to hold on that adapter too.

---

All 20 checks are green, and I re-ran `legacy-data-migration.integration.spec.ts` locally (7/7). Nothing here is a regression — the ask is to correct the description's clock claim and to decide the scope question above.

johannesjo:
## Corrections to my review above

I re-verified my own review line by line. Three things were wrong or overstated — sorry, and thanks for your patience.

### 1. Withdrawn: the #8389 merge-order point

I claimed the new four-store transaction was novel and would collide with #8389. Both halves are wrong.

`replaceRejectedRepair` on `master` already opens exactly `[OPS, VECTOR_CLOCK, META, STATE_CACHE]` (master `operation-log-store.service.ts:2792-2797`), so there's nothing new for an adapter to support. And #8389's three hunks in this file are at the import block, `getLastSeq`, and `hasSyncedOps` — nowhere near the new primitive. No conflict. Withdrawn entirely.

### 2. Wrong line and wrong word on the hydrator point

It's `operation-log-hydrator.service.ts:259`, not 257 (257 is a comment), and it is **not** unconditional — it's guarded:

```ts
if (snapshot.vectorClock && Object.keys(snapshot.vectorClock).length > 0) {
  await this.opLogStore.setVectorClock(snapshot.vectorClock);
```

It does fire after any atomic anchor (which always writes a non-empty clock), so the substance holds — `hydrateStore()` takes no lock, and the tail clocks are only merged later at `:790` — but "unconditionally" was a checkable falsehood and shouldn't have been in there.

### 3. My headline on the clock was too strong, and I under-credited the rework

Two things I got wrong in framing:

**The window round 1 actually named is closed.** The objection was the *separate follow-up* `setVectorClock()`; that write is gone, and there's no post-commit clock write left. Saying the invariant "is still violated" reads as though the rework achieved nothing. It didn't achieve nothing — what remains is the *pre-transaction* window, which is a different and smaller claim.

**My reproduction ran on the branch I then argued has no callers.** It used `createSyncImportOp = true`. I should have connected those two sections instead of leaving the repro looking like a live production failure. On the live legacy-migration path this PR is a **strict improvement**: `master` lost the concurrent op *and* overwrote the clock; now only the clock overwrite remains.

So this is **not a merge blocker** — I was wrong to file it as blocking. Here's the residual concern in its accurate, actionable form, which is also a better ask than the `mergeRemoteOpClocks` pointer I originally gave:

The two sibling primitives in this same file already do what `_appendOperationAndSnapshot` doesn't:

- `commitFileSnapshotBaseline` (`:917`) re-reads the op-log tail **inside** its transaction and aborts if it moved (`:968-973`, *"Cannot commit a file snapshot after the operation-log tail changed"*).
- `replaceRejectedRepair` (`:2824`) reads the durable clock **inside** its transaction and rebases onto it (`:2856-2865`) — with a comment naming this exact hazard: *"the caller-built clock may come from a stale in-memory cache (#8939)."*

And the staleness is wider than "between line 190 and line 320": `getVectorClock()` returns the per-tab `_vectorClockCache` when set (`:2577-2580`), and `clearVectorClockCache()` has exactly one production caller — the capture path at `operation-log.effects.ts:248`, under `sp_op_log`. Neither `SyncHydrationService` nor `OperationLogMigrationService` calls it, so the clock those paths build can predate anything another tab has written since this tab last refreshed. That's precisely what `appendWithVectorClockOverwrite`'s `INVARIANT (#8939)` docstring (`:2761-2769`) tells new writers not to do.

Adding the in-transaction rebase to `_appendOperationAndSnapshot` would be a handful of lines, mirroring `replaceRejectedRepair`. Worth doing, since this primitive is presented as the canonical replay-anchor writer — but a follow-up is fine, and it's pre-existing either way.

### 4. Two precision nits in my minor points

- `createBloatedClock` yields `MAX_VECTOR_CLOCK_SIZE + 6` entries (MAX+5 bloat plus the passed one), not "+5". The count I gave (26) was right; the formula wasn't. And I should have said plainly that the unpruned-recovery-clock behaviour is **pre-existing** on `master` — this PR pins it with a test rather than introducing it.
- The `spyOn(_performMigration)` is at `operation-log-migration.service.spec.ts:367` (365 is the comment above it), and that mock-copies-production pattern predates this PR too.

### Net

Unchanged and verified: the two new regression tests are real (I reverted the primitive to the pre-PR sequence and both fail), the crash-atomicity win is real, the recovery refactor is behaviour-preserving, and `createSyncImportOp = true` has no production callers.

Revised verdict: **not blocking.** The seq fix is correct on a live path, the tests are honest, and nothing here regresses. The two asks I'd still make are the scope question — whether the SYNC_IMPORT half should exist at all, given `docs/research/sync-simplification-audit/findings.md:1214-1232` already recommends deleting that facade — and correcting the PR description's claim about the clock, so the next reader doesn't inherit my overstatement in the opposite direction.

mvanhorn:
Thanks for re-verifying your own review line by line - that level of care is rare. Pushed f6279ce addressing the items that survived the corrections, with spec coverage added. The operation-log-store suite passes locally (237/237, ChromeHeadless).

johannesjo (commented):
## Follow-up review of `f6279ce`

I carefully rechecked the objection that the app is not usable cross-tab. That is true for normal, settled UI use, so my earlier wording was too broad. It does **not**, however, serialize startup storage work:

- `DataInitService` starts hydration/migration from `APP_INITIALIZER` (`main.ts:267-275`, `data-init.service.ts:18-24,39-53`).
- Only later does `AppComponent` call the asynchronous BroadcastChannel check (`app.component.ts:224-225`).
- The loser only replaces `document.body.innerHTML` (`startup.service.ts:348-360`); Angular services/effects keep running.
- Simultaneous cold tabs can both pass the 50 ms probe because each installs its `newTabOpened` responder only after deciding it is alone (`startup.service.ts:300-345`).

### Important follow-up: the inverse migration ordering still hides a pre-genesis op

The new migration test covers only:

```
genesis + snapshot commit (seq 1)
second-tab append (seq 2)
```

because `appendFromSecondTab()` is called after `realAtomicAppend()` resolves (`legacy-data-migration.integration.spec.ts:306-312`). That ordering is safe.

The inverse ordering is production-reachable during overlapping startup:

```
second-tab persistent append (seq 1)
migration genesis + snapshot commit (seq 2)
```

`appendOperationAndSnapshot()` writes `lastAppliedOpSeq = writtenSeq` with the caller's legacy snapshot state (`operation-log-store.service.ts:819-826`) but never verifies that earlier OPS rows are represented by that state. The clock rebase notices seq 1's clock; the snapshot still does not contain seq 1's task. Restart therefore begins replay after seq 2 and never applies seq 1.

I reproduced this on exact head with the real IndexedDB adapter and the real migration path by moving the existing test's `appendFromSecondTab()` immediately before `realAtomicAppend()`:

```
cache.lastAppliedOpSeq: expected 1, got 2
replay tail: expected ["second-tab-after-migration"], got []
```

This inverse window predates the PR, so the PR remains a net improvement; it is not an introduced regression. But the current test proves only one side of the concurrency boundary, and #8337 is being closed as the multi-tab replay-loss fix. Please either close this ordering here or track it explicitly before closing the issue.

The smallest safe shape appears to be an `LOCK_NAMES.OPERATION_LOG` barrier acquired before migration's op-log emptiness decision and held through the genesis anchor commit, with snapshot/tail rechecked after acquisition. Queued capture writes then become post-anchor replayable ops. The cosmetic one-second completion delay should stay outside that lock. A pre-anchor/tail-guard regression test would pin the required invariant.

### Non-blocking: pruning before rebase does not enforce the clock bound

`appendOperationAndSnapshot()` prunes at line 775, then lines 813-817 merge the entire durable clock back in and persist the unpruned union. A real-store test with disjoint 20-entry durable/proposed clocks persisted 40 entries, despite `MAX_VECTOR_CLOCK_SIZE = 20`.

This is a lower-severity contract issue on the live migration path, but it was introduced by the new rebase. Rebase first and apply bounded pruning to the committed clock inside the transaction (preserving the current/latest full-state authors).

### Minor/dormant: the SYNC_IMPORT working-clock reset is undone

`sync-hydration.service.ts:291-306` deliberately passes a minimal importer-only working clock, but the same rebase restores every durable foreign ID. I reproduced `{testClient: 6, remoteClient: 9}` being persisted where the caller requested `{testClient: 6}`.

There is no production caller of `createSyncImportOp=true`; both live sync callers pass `false`, so this is not a merge blocker. It is still a mismatch in the changed helper/API contract (or another reason to remove the dormant branch discussed earlier).

### Verification

- Current head remained `f6279ce82ab438d5abbc063cca09e501636e6d66`.
- All 20 GitHub checks are green.
- Exact-head suites passed unchanged: operation-log store 237/237, sync hydration 50/50, legacy migration 7/7.
- Temporary inverse-order and helper-contract tests failed only at the three assertions described above.

mvanhorn:
Pushed fdd485d, which closes the inverse migration ordering in the shape you specified.

What changed in `OperationLogMigrationService.checkAndMigrate()`:

- `LOCK_NAMES.OPERATION_LOG` is now acquired before the op-log emptiness decision and held through the genesis anchor commit, so the decision and the write are on the same side of the barrier.
- Both replay anchors are re-read after acquisition, not before: `loadStateCache()` and `getOpsAfterSeq(0)`. If either moved while we waited, we bail rather than anchor against a stale tail.
- The cosmetic one-second completion delay moved outside the barrier, per your note, so queued capture writes are not held behind a purely visual wait.

Two things fell out of the restructure that were latent before: the legacy migration lock is now released only when it was actually acquired (previously an unconditional release in `finally`), and the dialog close is defensive since `dialogRef` may never be assigned on the early-return paths.

The regression test is the inverse of the existing one: `appendFromSecondTab()` now runs before `realAtomicAppend()` resolves, so the second tab's append is queued behind the barrier. It asserts `cache.lastAppliedOpSeq === 1` and that the concurrent op replays at seq 2, which is the pair that failed on the pre-fix head.

Verified locally: 3653/3655 passing, 2 skipped, 0 failures. Prettier and eslint clean on both files.

Still open from your review, both of which you marked non-blocking and neither of which this push touches:

1. The pruning-before-rebase bound in `appendOperationAndSnapshot()` - prune at :775 then merge the full durable clock back at :813-817, so a disjoint 20+20 pair persists 40 entries against `MAX_VECTOR_CLOCK_SIZE = 20`. Your prescription (rebase first, then bounded pruning on the committed clock inside the transaction, preserving current/latest full-state authors) is clear enough to implement directly. Happy to do it here or as a follow-up, your call.
2. The `createSyncImportOp=true` working-clock reset. Given you and the sync-simplification audit both point at deleting that facade, I would rather not patch a branch with no production callers. If you want the SYNC_IMPORT half dropped from this PR entirely, say so and I will strip it.

On the PR description: you are right that it overstates the clock claim. I will rewrite that paragraph to say the post-commit `setVectorClock()` window is closed and the pre-transaction staleness window remains, rather than implying the invariant is fully restored.

johannesjo:
## Pre-review double-check (`fdd485d`)

I rechecked the exact head, current discussion, and CI. The new `OPERATION_LOG` barrier plus the inverse-order real-store test do close the pre-genesis ordering from the last review; all 21 checks are green.

Before a formal review, I would address one correctness issue:

### Important: prune-before-rebase can persist an over-limit clock

`appendOperationAndSnapshot()` prunes `snapshot.vectorClock` at `operation-log-store.service.ts:775`, but inside the transaction it unions the entire durable clock back into both the operation and committed clock at `:805-817`, then writes those unbounded values at `:819-839`. A disjoint bounded 20-entry durable clock plus bounded 20-entry proposed clock becomes 40 entries despite `MAX_VECTOR_CLOCK_SIZE = 20`.

This is reachable on the live migration path: `clearAllOperations()` clears `OPS`/`META` but intentionally retains `VECTOR_CLOCK`, and migration then calls this helper. Retaining and rebasing that causal frontier is correct—clearing it risks counter reuse—but the result still needs to be bounded after the rebase. Please add a real-store 20+20 regression and bound both the stored operation clock and committed snapshot/working clock while preserving the required current/full-state authors. Because this behavior comes from the new helper, I would fix it in this PR rather than defer it.

### Reliability risk: the migration barrier can outlive lock waiters

`operation-log-migration.service.ts:98-152` now holds `sp_op_log` across `getOpsAfterSeq(0)` (a full log load/decode), translation loading, two full legacy reads, backup serialization/download, dynamic imports, validation/repair, and client-ID I/O. Captures wait on the same lock and time out after 30 seconds; the ordinary effect path can then have an optimistic NgRx change with no durable op, marks persistence divergence, and shows the sticky reload action (`operation-log.effects.ts:148-157,401-428`). Overlapping web startup is reachable before the single-tab check; what is not measured is whether real migration work crosses 30 seconds.

Please either bound/prove this critical section below the waiter timeout, or stage the unbounded preparation with migration-aware coordination and keep only the final snapshot/log recheck, orphan decision/cleanup, and genesis commit under one uninterrupted `OPERATION_LOG` barrier. Moving the recheck or commit outside that barrier would reopen the just-fixed ordering. A timeout regression would make this concrete; at minimum, replace `getOpsAfterSeq(0)` with a first-row/existence query while the barrier is held.

### Scope / description

Both production sync callers still pass `createSyncImportOp=false`, and the default-true forwarding chain has no production caller; the repository audit already recommends deleting it. My preference is to strip the dormant SYNC_IMPORT half from this PR and keep the live migration fix focused, rather than maintain its generic helper/test surface and importer-only working-clock mismatch. I view this as scope cleanup, not a reachable regression.

The PR body is also still describing the original returned-seq-only implementation and old verification status. Please update it to the atomic helper plus migration barrier that is actually at `fdd485d`.

mvanhorn:
Fixed the correctness issue and updated the PR body. Two of your four points still need your call.

**Clock bounding**: fixed as described. The rebase is now re-bounded to `MAX_VECTOR_CLOCK_SIZE`, preserving the current client and the latest full-state author so it matches `pruneClockForStorage`. The two author lookups are async so they resolve before the transaction opens; the bounding itself is synchronous and safe inside it. Regression drives the real store with a disjoint 20 + 20; with the bound neutralised it reports `Expected 40 to be less than or equal 20` at all three write sites (stored op, committed clock, state cache). 3654 tests in `src/app/op-log/**` pass.

**PR body**: rewritten to describe the atomic helper, the migration barrier and the clock bound, with a verification section. It was still describing the original returned-seq-only version.

**Migration barrier**: I have not changed this, because I would be guessing at which of your two options you want. Swapping `getOpsAfterSeq(0)` for an existence query under the barrier is the cheap half and I am happy to do it, but you framed it as a minimum rather than the fix, and the real question is whether the unbounded preparation gets staged outside the barrier. That restructure has a way of reopening the ordering bug you just had me close, so I would rather you pick the shape.

**SYNC_IMPORT scope**: no objection to stripping it, and your reasoning matches the repo audit. I have not done it in this push because it is a deletion across the helper and its tests and I did not want to fold it into a correctness fix you are about to review. Say the word and it goes in its own commit.

johannesjo (commented):
## Review (`3f128cf9`): needs changes

### Summary

I re-fetched the complete current diff, issue/discussion history, review-thread state, and CI, then traced the affected startup, locking, snapshot, and vector-clock paths against the base commit. The atomic append-plus-anchor primitive is a real improvement, the inverse-order migration regression is meaningful, and the post-rebase 20+20 clock bound is now correct. All 22 checks pass.

Two material items remain, plus one dormant contract mismatch.

### Important: the migration barrier can outlive ordinary lock waiters

`operation-log-migration.service.ts:98-150` now holds `sp_op_log` across a full `getOpsAfterSeq(0)` read/decode, translation loading, two full legacy-data reads, backup serialization/download, dynamic imports, validation/repair, client-ID I/O, and the final commit. Normal captures wait on the same lock and abort after 30 seconds (`LOCK_ACQUISITION_TIMEOUT_MS`); their effect path then has an optimistic NgRx change with no durable operation, marks persistence divergence, and shows the sticky reload action.

This overlap is reachable on web startup: `DataInitService` starts hydration from its bootstrap initializer before `AppComponent` invokes `StartupService.init()`, where the BroadcastChannel single-tab check occurs. A large/slow legacy migration therefore can make another tab's capture time out even though the barrier itself eventually succeeds. The compaction path is explicitly capped at 25 seconds for the same lock-safety reason; this new critical section has no comparable bound or regression.

Please make capture/migration coordination explicit before moving expensive preparation outside the lock, then keep the final snapshot/log recheck, orphan decision/cleanup, and atomic genesis commit together under one uninterrupted barrier. Alternatively, prove and enforce that the entire lock-held path stays below the waiter timeout. Merely moving the recheck or commit outside would reopen the ordering bug, and replacing the full-log read with an existence query is useful but does not bound the other work.

### Important: `Fixes #8337` would close a still-live snapshot-frontier variant

Issue #8337 documents a fourth call site in `OperationLogSnapshotService.saveCurrentStateAsSnapshot()`. It captures this tab's NgRx state at `operation-log-snapshot.service.ts:136`, then reads the shared OPS store's global max seq at `:137` and persists that value as `lastAppliedOpSeq` at `:162-165`.

The operation-log architecture explicitly says live tabs do not broadcast operation payloads. Thus, an operation already appended by another tab can be included in the global max seq while absent from this tab's captured state; the next hydration skips that operation because the cache frontier is too high. This PR correctly fixes the two append-followed-by-read sites, but it does not touch this snapshot-caching site, and I found no separate open follow-up.

Smallest safe resolution: change the PR body to `Refs #8337` and create/link a follow-up before merge, or cover the remaining call site here. Keeping `Fixes #8337` would auto-close the only tracker for a known data-loss path that remains live.

### Minor / dormant: the requested minimal working clock is not what the real store persists

The default-`true` SYNC_IMPORT branch intentionally passes a minimal working clock at `sync-hydration.service.ts:291-324`, but `appendOperationAndSnapshot()` rebases that value on the durable clock at `operation-log-store.service.ts:852-854`, merging foreign client IDs back into the snapshot and working clock. The unit test at `sync-hydration.service.spec.ts:384-396` only checks the argument given to a mocked store helper, so it cannot observe the persisted mismatch.

Both production sync callers pass `false`, and the remaining default-`true` forwarding chain has no production caller; the repository's sync-simplification audit already recommends deleting it. My preference is the lean path the author offered: strip this dormant SYNC_IMPORT scope from the PR. If it stays, the helper needs distinct full operation-clock vs minimal working-clock semantics and a real-store regression.

### Strengths

- Operation, replay frontier, and working clock now commit atomically in one IndexedDB transaction.
- The inverse migration-order and disjoint 20+20 vector-clock tests exercise the real persistence store and cover the previously missing failure shapes.

**Verdict: needs changes.**

mvanhorn:
Item B done in 4246bd2 — the dormant SYNC_IMPORT scope is gone.

I checked your premise before deleting anything, and it holds: both live callers in `OperationLogSyncService` pass `false` explicitly, and the only path reaching the default-`true` was `DataInitService.reInitFromRemoteSync`, whose sole remaining reference was a spy name in `sync-wrapper.service.spec.ts`. So the flag, `syncImportReason`, that method and the hydrator forwarder it went through are all removed. Hydration now always commits via `commitFileSnapshotBaseline`, which means no clean-slate baseline and the full merged clock is retained instead of being reset to the local entry.

On your point about the test only inspecting a mocked helper's argument — I didn't want to just delete the 27 tests that went through the old path, because most of them aren't about SYNC_IMPORT at all: they cover globalConfig preservation, local-only sync settings, clock merging, error propagation. Those are re-pointed at the committed baseline so they assert the same properties on the surviving path. Only the two whose entire subject was creating a SYNC_IMPORT were dropped. I checked the result is falsifiable: reverting the full-clock behaviour fails 6 of them.

13259 tests pass, tsc and eslint clean.

Item A, the migration barrier, I have not touched. Your constraint is clear — keep the recheck, orphan decision and genesis commit under one uninterrupted barrier, and either coordinate capture/migration explicitly before moving preparation out or prove the lock-held path stays under the waiter timeout. I didn't want to pick between those two without your read, given the compaction path's 25s cap is the precedent you pointed at. Which would you rather I build?

johannesjo (commented):
## Review (`4246bd2`): SYNC_IMPORT removal verified; answers to both open items

I re-fetched the head, the full discussion, and CI (22/22 green at `4246bd2`), and verified the new commit against the exact head. Below are also the two answers you were waiting on. Both are resolved in your favor: one I took care of myself, and on the other I am withdrawing most of my own ask.

### The removal commit is clean

- The surviving path is line for line the old `createSyncImportOp=false` path: unconditional `getUnsynced()`, archive writes deferred into `commitFileSnapshotBaseline()`, full merged clock kept. Both production callers already passed `false`, so shipped behavior is unchanged.
- Your falsifiability claim is exact. At head the suite is 47/47; with `clockForStorage = newClock` sabotaged back to the minimal-clock reset, exactly 6 specs fail.
- No dangling references remain (`reInitFromRemoteSync`, `createSyncImportOp`, the hydrator forwarder), and the spy rename matches the real `DataInitService.reInit`.
- Nothing is orphaned: `CleanSlateService` and backup restore still create SYNC_IMPORT operations through `runDestructiveStateReplacement`, so the filter service, the conflict dialog, and the `syncImportReason` wire field all stay live.
- Semantic merge against current master is clean: the op-log commits that landed since your merge base (#9412, #9326, #9242) add no caller of the removed API; the only `reInitFromRemoteSync` references on master are the declaration and the spy line this PR deletes.

Three cosmetic leftovers to fold into your next push, none blocking:

1. `sync-hydration.service.ts:49` and `:74`: the class and method JSDoc still describe creating SYNC_IMPORT operations, contradicting the new in-body comments.
2. `sync-hydration.service.ts:66`: `archiveDbAdapter` is now an unused injection (its only consumer was the removed archive-write block); the `ArchiveDbAdapter` import goes with it.
3. `sync-hydration.service.ts:243`: the bare `{ ... }` block left where the `else` branch used to be can be unwrapped.

### Item A (migration barrier): build neither restructure; do the first-row swap and document the bound

On re-analysis I am downgrading my own reliability finding, because its headline scenario is not reachable the way I framed it. A capture waiter needs a tab far enough through startup to dispatch persistent actions, which in practice means hydration completed. A hydrated install implies either a state cache exists or the op log starts with a genesis MIGRATION/RECOVERY operation, and in both cases your barrier returns after the `loadStateCache()` read or the first-op check. The expensive lock-held path therefore only runs when no tab can realistically be capturing. What actually remains:

- Two simultaneously cold tabs on a legacy install: the loser's lock request times out after 30 s, that boot fails loudly, and a reload lands on the fast path. Once per install, web only, recoverable.
- A degraded boot (ops present but no state cache, e.g. after checkpoint saves were skipped): here `getOpsAfterSeq(0)` under the barrier is a full log read and decode whose duration grows with the log, and it can block a live tab's captures even though migration never runs. This is the one hold with a realistic path to hurting someone.

So: please do only the cheap half you already offered. Replace `getOpsAfterSeq(0)` with a first-row read; the branch only needs existence plus the first op's `entityType`, and the orphan-count log can use an IDB `count()` if you want to keep the number. Since `lockedSnapshot` is only truthiness-tested, that recheck can become an existence probe too, which spares every cold boot a full state-cache materialization under the lock, but I leave that one to your judgment. Add a short comment on the barrier stating that the once-per-install legacy migration path may exceed the 30 s waiter timeout by design. Keep the recheck, orphan decision, and genesis commit exactly where they are. I withdraw the ask to stage preparation outside the lock or to prove a sub-30 s bound.

### Item B (`Fixes #8337`): resolved on my side, no change needed from you

Rather than have you flip a word in the body, I split the remaining `saveCurrentStateAsSnapshot()` variant, which only lived in a comment on the issue, into its own tracker: #9438. With that, every call site listed in #8337's body is covered at this head (the SYNC_IMPORT branch by removal, migration by the atomic helper plus barrier, the no-import branch by `commitFileSnapshotBaseline()`'s in-transaction tail check, recovery by `appendRecoveryOperationAndSnapshot()`), so `Fixes #8337` is now correct as written and auto-closing it on merge is the right outcome.

### Where this lands

After the first-row swap and the three cosmetic leftovers, this is ready to merge from my side.

johannesjo:
Thank you very much ! <3

mvanhorn:
Appreciate it, @johannesjo. Using append's returned seq closes that op-log race for good.

PR #9159, opened 2026-07-18
fix(scheduled-list): parse 'Last' repeat date as local day to fix off-by-one (#9127)

### LINKED ISSUE #9127: 🚨 'Last' dates in 'Upcoming' page sometimes incorrect
### Version Used

v18.14.0

### Current Behavior

In the 'Upcoming' page, the 'Last' date (visible when hovering over the "Every x days" element for a task) is sometimes incorrect. I'm not sure what causes it to be incorrect, but exiting SP and restarting it results in the date being closer to correct (but still off by 1 day). Interestingly, clicking on the "Every x days" element and changing the start date to a different date doesn't fix the 'Last' date.

### Expected Behavior

'Last' date to reflect the date the task was last completed.

### Steps to Reproduce

See "Current Behaviour". Sorry I don't have detailed STR... I'm very busy this month and next, so I hope this report is better than nothing.

Sometimes, it's intermittent.

### Console Output

```shell

```

### Documentation

mvanhorn:
## Problem

Fixes #9127.

On the Upcoming/Scheduled list, hovering an "Every x days" repeat shows a tooltip whose **'Last'** date is consistently off by one day for some users (the reporter is on v18.14.0; the error persists across restarts).

`ScheduledListPageComponent.getTooltipText()` rendered the 'Last' value from `getEffectiveLastTaskCreationDay(repeatCfg)`, which returns a `YYYY-MM-DD` string, via `formatMonthDay(new Date(effectiveLastDay), locale)`. `new Date('2025-08-01')` parses a **date-only** string as **UTC midnight**, while `formatMonthDay` → `Intl.DateTimeFormat` renders in the **system timezone**. For any user west of UTC (e.g. `America/Los_Angeles`), UTC-midnight falls on the previous local day, so "8/1" renders as "7/31". The 'Next' value is unaffected because it comes from a real epoch timestamp, not a date string.

Standalone repro of the exact date logic (`Intl` month/day numeric):

```
TZ=America/Los_Angeles   new Date('2025-08-01') -> "7/31"   dateStrToUtcDate('2025-08-01') -> "8/1"
                         new Date('2026-01-01') -> "12/31"  dateStrToUtcDate('2026-01-01') -> "1/1"
TZ=Europe/Berlin / UTC   both -> "8/1" / "1/1"   (east-of-UTC and UTC do not drift)
```

## Solution

Parse the day string with the codebase's canonical `dateStrToUtcDate` helper (`src/app/util/date-str-to-utc-date.ts`), which builds a **local start-of-day** `Date` and is already used elsewhere for exactly this. The rendered calendar day then matches the input string in every timezone.

To make the fix unit-testable without a heavy component TestBed, the tooltip-formatting logic is extracted into a pure `get-repeat-cfg-tooltip-text.util.ts` (takes the already-computed next timestamp, the effective-last-day string, the locale, and the two already-translated labels; formats `next` from the timestamp unchanged and `last` via `dateStrToUtcDate`), and `getTooltipText()` delegates to it. Output shape, labels, and empty-segment handling for `undefined`/`null` inputs are preserved.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have included relevant changes to the documentation/wiki — not applicable (no documented behavior change).
- [ ] I have run `npm run checkFile` on changed `.ts`/`.scss` files — the full Angular toolchain was not available in my environment; CI runs prettier/lint. The changes follow the surrounding 2-space/prettier style.
- [x] I have added tests for my changes (if applicable) — added `get-repeat-cfg-tooltip-text.util.spec.ts` covering the tz off-by-one (fails under `TZ=America/Los_Angeles` with the old `new Date(str)` path, which the repo's `test:tz:la` job exercises), month/year-boundary strings, locale ordering (en-US / en-GB / de-DE), and missing-value handling. The exact date logic was also verified with a standalone Node repro (above).
- [ ] Existing tests still pass — not run locally (toolchain unavailable); validated by CI.

### DIFF
--- a/src/app/pages/scheduled-list-page/get-repeat-cfg-tooltip-text.util.spec.ts
+++ b/src/app/pages/scheduled-list-page/get-repeat-cfg-tooltip-text.util.spec.ts
@@ -1,0 +1,127 @@
+import { getRepeatCfgTooltipText } from './get-repeat-cfg-tooltip-text.util';
+import { DateTimeLocales } from '../../core/locale.constants';
+
+describe('getRepeatCfgTooltipText', () => {
+  // A stable local-noon timestamp for the 'Next' value so its rendered date
+  // does not cross a day boundary under any test timezone.
+  const nextTs = new Date(2025, 7, 15, 12, 0, 0).getTime(); // 2025-08-15, local
+
+  describe("'Last' date timezone handling (#9127)", () => {
+    it('renders the effective-last day string as its own calendar day, not the UTC-parsed previous day', () => {
+      // Regression: the old `new Date('2025-08-01')` parses as UTC midnight and
+      // formatMonthDay renders it in local time, so users west of UTC see
+      // "7/31". The util parses the string as a local day, so it stays "8/1".
+      // The repo's `test:tz:la` job runs this under America/Los_Angeles.
+      const result = getRepeatCfgTooltipText(
+        nextTs,
+        '2025-08-01',
+        DateTimeLocales.en_us,
+        'Next',
+        'Last',
+      );
+
+      expect(result).toBe('Next 8/15, Last 8/1');
+      expect(result).not.toContain('7/31');
+    });
+
+    it('does not drift across a month boundary', () => {
+      const result = getRepeatCfgTooltipText(
+        null,
+        '2025-12-31',
+        DateTimeLocales.en_us,
+        'Next',
+        'Last',
+      );
+
+      expect(result).toContain('Last 12/31');
+      expect(result).not.toContain('12/30');
+    });
+
+    it('does not drift across a year boundary', () => {
+      const result = getRepeatCfgTooltipText(
+        null,
+        '2026-01-01',
+        DateTimeLocales.en_us,
+        'Next',
+        'Last',
+      );
+
+      expect(result).toContain('Last 1/1');
+      expect(result).not.toContain('12/31');
+    });
+  });
+
+  describe('locale ordering', () => {
+    it('renders month-first for en-US', () => {
+      const result = getRepeatCfgTooltipText(
+        null,
+        '2025-08-01',
+        DateTimeLocales.en_us,
+        'Next',
+        'Last',
+      );
+      expect(result).toContain('Last 8/1');
+    });
+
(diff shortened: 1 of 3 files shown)


johannesjo (commented):
## Summary

The timezone fix is correct at head `41ca033`: the old date-only parser demonstrably shifts `2025-08-01` to `7/31` west of UTC, while `dateStrToUtcDate()` preserves the intended local calendar day without changing the `Next` timestamp path.

## PR Discussion Context

There are no existing review threads. All required CI checks are green, including the test suite's `America/Los_Angeles` leg, which genuinely fails with the pre-fix parser.

## Fundamental Question

#9127 describes two observable behaviors: restarting makes the displayed date “closer to correct,” and it remains one day off afterward. This PR conclusively reproduces and fixes the second formatting symptom, but it does not reproduce the restart-dependent discrepancy because the regression test starts with a hard-coded effective last-day value.

Should this PR avoid auto-closing the entire issue until that first symptom is explained? The smallest option is to replace `Fixes #9127` with a non-closing reference, or explicitly track the restart-dependent behavior in a follow-up issue.

## Issues

### Minor (non-blocking)

- `src/app/pages/scheduled-list-page/get-repeat-cfg-tooltip-text.util.ts:21` — The production fix is conceptually a parser substitution, but it introduces a single-use five-positional-argument helper and a 127-line spec. The extraction is understandable for testability; still, consider trimming the suite to the timezone regression and the preserved missing-value contract, or otherwise reducing the one-use API surface.

## Strengths

- Reuses the repository's established logical-day parser rather than introducing another date convention.
- Preserves the existing output and null behavior, with meaningful month/year-boundary coverage.

## Verdict

**Needs discussion** — the implementation itself looks safe and CI is green; the remaining material question is issue closure, with the scope note optional.

johannesjo:

mvanhorn:
Thanks for the heart, @johannesjo - parsing the Last repeat date as a local day kills the off-by-one for good.

Issue #9170, opened 2026-07-18
WebDAV USE_LOCAL snapshot is skipped when a tail op masks the version reset


johannesjo:
## Problem

A real browser run exposed silent WebDAV divergence after `USE_LOCAL` replacement:

1. Client A uploads task A and records the current file `syncVersion`.
2. Client B has task B, connects, and chooses **Keep local**. The authoritative snapshot replaces the remote and resets `syncVersion` to 1.
3. Before A syncs, B creates and uploads task B2. That incremental op advances the remote `syncVersion` back to the value A already expects and makes `recentOps` non-empty.
4. A syncs and reports success. It receives B2, but never hydrates B's replacement snapshot: task A remains, task B is missing, and both clients show “In sync” with different data.

This reproduced in `webdav-encryption-conflict-use-local.spec.ts` after replacing its permissive final assertion with an exact state oracle. The failure screenshot showed A containing stale task A plus B2, while B contained task B plus B2.

## Likely cause

`FileBasedSyncAdapter.downloadOps()` detects a replacement when the observed `syncVersion` regresses, or when `recentOps` is empty and the snapshot writer differs. One tail operation can simultaneously hide both signals: the version catches up to A's expected value and `recentOps` is no longer empty. The tail is then applied without the snapshot that it is based on.

Encryption is not believed to be causal; it is simply how the audited test enters the first-sync `USE_LOCAL` path.

## Acceptance criteria

- [ ] Add a deterministic adapter/integration reproducer: an established reader, a replacement snapshot from another client, then enough tail operations for the remote version to equal or exceed the reader's expected version before its next download.
- [ ] Make snapshot lineage/replacement detectable even when the version has caught up and `recentOps` is non-empty. Do not rely on wall-clock timing.
- [ ] Rehydrate the replacement snapshot before applying its tail; never report success after applying the tail alone.
- [ ] In the browser regression, assert the replaced task is absent and the exact snapshot plus tail tasks are present on both clients.
- [ ] Reload both clients and repeat exact ID/title convergence assertions.
- [ ] Cover both encrypted and unencrypted WebDAV data, or prove at a lower level that encryption is outside the branch.
- [ ] Preserve stale-writer and split-snapshot generation safeguards tracked in #9147.

Related: #9147

Issue #9168, opened 2026-07-18
sync: fail closed on truncated successful SuperSync batches


johannesjo:
## Problem

The deleted `supersync-divergence-bug-6571.spec.ts` exposed a distinct integrity gap even though it was not a valid #6571 regression test: it removed one operation from an otherwise successful download response while retaining later success metadata. Both clients then reported success, the durable cursor advanced, and the omitted operation was skipped permanently.

Raw `serverSeq` gaps are not themselves proof of truncation: normal downloads exclude the requesting client's own operations, and full-state fast-forward may skip superseded history. The missing guarantee is completeness of the exact filtered, paginated query result the client was promised.

This must not be kept as a green test whose expected result is permanent divergence.

## Acceptance criteria

- [ ] Define the threat model. A commitment generated by a malicious server over its own truncated result is not protection; malicious-server coverage needs a client-verifiable/trusted anchor, while accidental response corruption may use a narrower authenticated server commitment.
- [ ] Bind any completeness proof to the effective snapshot/since boundary, requesting or excluded client, result filter, ordered operation set, `latestSeq`, `hasMore`, and page chaining.
- [ ] Add deterministic lower-level coverage and a narrow browser/server failure injection that omits an operation which belongs to that filtered result while retaining otherwise-successful response metadata.
- [ ] Detect the omitted query result before reporting remote-confirmed success. Do not require naive raw sequence contiguity.
- [ ] Do not advance the durable cursor past work whose completeness was not established.
- [ ] Remove the injection, retry, and prove exact operation/entity convergence on existing and fresh clients.
- [ ] Cover pagination, excluded-own-operation gaps, full-state fast-forward boundaries, and omission from both the middle and end of a promised result.
- [ ] Coordinate client-authenticated ordering/batch work with #8907.

Related: #8907, #9148

Issue #9167, opened 2026-07-18
e2e(WebDAV): reproduce #6992 recurring-task day-boundary race


johannesjo:
## Problem

The deleted `supersync-day-change-overdue-6992.spec.ts` did not reproduce #6992. The report is a Nextcloud/WebDAV race: one client completes recurring tasks and replans a yesterday task into today, then a delayed client runs new-day processing before pulling that work while both clients report successful sync. The deleted test instead used SuperSync, created only ordinary tasks without repeat or due-date state, and compared titles only.

The real browser seam remains uncovered. This needs provider-accurate coverage because a stale raw `TODAY_TAG.taskIds` ordering entry can still feed the delayed client's overdue cleanup even though an overdue task is absent from computed virtual TODAY membership.

## Acceptance criteria

- [ ] Use two real WebDAV/Nextcloud-compatible clients and a controllable logical-day trigger; do not rely only on sleeps.
- [ ] On day X, create a real repeat configuration and a separate task due on day X whose ID is present in raw `TODAY_TAG.taskIds`; sync both clients.
- [ ] Cross into day X+1. Before cleanup, prove the overdue task's raw TODAY ordering entry is stale while its computed virtual TODAY membership is absent.
- [ ] On Client A, complete the recurring occurrence and plan the overdue task for day X+1, then upload those changes.
- [ ] Before pulling Client A, let delayed Client B process day X+1 and prove `localRemoveOverdueFromToday` creates no persistent competing operation.
- [ ] Sync both directions and assert exact task IDs, repeat-instance identity/deleted-instance dates, due day, raw TODAY ordering, and computed TODAY membership. After A's plan wins, the task must have `dueDay` day X+1 and computed membership must be present.
- [ ] Assert no missing or duplicate entity, then reload/restart both clients and repeat the exact convergence assertions.
- [ ] Keep the broader SuperSync recurring-configuration browser seam in #9146; this issue owns the provider-specific #6992 interleave.

Related: #6992, #9146

Issue #9164, opened 2026-07-18
test(sync): add exact vector-clock and LWW retry boundary oracles


johannesjo:
## Problem

The browser tests adjacent to vector-clock pruning and LWW retry behavior cover useful but narrower paths:

- `supersync-import-other-client-ops.spec.ts` sends a real upload containing 25 injected clock IDs and proves server storage is pruned to 20 while retaining the import author.
- The other browser cases exercise ordinary three-client convergence and later equivalent updates.

They do not prove natural MAX+1 client accumulation, the critical same-entity compare-before-prune boundary, or a bounded local-win retry. They also do not distinguish max-timestamp selection from first/last/min selection when every write from one client happens later than every write from the other.

## Acceptance criteria

- [ ] On one entity, establish a stored conflict frontier with exactly `MAX_VECTOR_CLOCK_SIZE` entries through real client flows.
- [ ] Have a new client produce a MAX+1 merged/local-win clock that dominates before pruning but would compare concurrent if pruned before conflict detection.
- [ ] Assert that same-entity operation is accepted, the stored clock is deterministically pruned back to 20 only after comparison, and the correct entity value wins.
- [ ] Count provider requests and assert the local-win/re-upload loop is bounded.
- [ ] Prove exact entity and clock convergence on existing and fresh clients.
- [ ] Keep a separate LWW scenario whose timestamps are interleaved so max selection produces a different winner from first, last, or min selection; assert the exact singleton winner and absence of losing titles.
- [ ] Add exact TODAY ordering and membership assertions for the TODAY-tag retry case.
- [ ] Coordinate already-pruned false-ordering/documentation/observability with #8755 rather than duplicating it.

johannesjo:
Multi-review found that the current multi-operation E2E does not prove max-timestamp selection: every B write occurs after every A write, so first, last, min, and max selection would all choose B. The test was renamed to the honest claim that a later multi-operation batch converges. The deterministic replacement here should interleave timestamps so the expected winner differs under max versus last or min selection, then assert exact singleton convergence and rejection of the losing title.

PR #9171, opened 2026-07-18
test(sync): harden end-to-end coverage and remove false oracles

johannesjo:
## Problem

The sync E2E suite contained several green tests that did not prove their stated behavior:

- completion helpers could accept stale or unrelated sync signals;
- convergence assertions sometimes compared two equally wrong clients without checking the expected entity set;
- recurring-task, vector-clock-boundary, and #6571-named specs did not drive the production paths named in their titles;
- a stale encryption status memo described coverage that no longer existed.

For sync, those false positives are particularly risky because they can hide silent cross-client data loss.

## Solution

- Harden the shared SuperSync and WebDAV page/helpers around fresh request/response evidence, sync-cycle ownership, and exact downstream state.
- Strengthen affected browser assertions for LWW, TODAY ordering/membership, imports, encryption, time tracking, migration, and provider late-join flows.
- Rename claims that remain intentionally narrower than the production boundary.
- Remove four misleading specs (10 runnable tests) and the stale status memo instead of retaining false coverage.
- Add post-baseline audit corrections and track every real uncovered seam in dedicated issues, including #9146, #9147, #9148, #9164, #9167, #9168, and the P1 WebDAV replacement-generation defect reproduced as #9170.

No production code, schema, dependency, or public interface changes are included.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [x] Other (test hardening and removal of false coverage)

## Verification

- 26/26 changed WebDAV E2E tests passed with `--retries=0`.
- All 21 changed TypeScript files passed `npm run checkFile`.
- Playwright sync collection succeeds: 242 tests in 87 files.
- `git diff --check` passes.
- Three independent post-commit reviews approved the final follow-up.
- The branch was rebased onto current `master` before publication.

The full scheduled SuperSync/WebDAV GitHub Actions workflow remains to be run for this branch.

## Checklist


johannesjo (commented):
## Review feedback

### Important — Scope native confirmation handling to explicit import choices

[`setupSuperSync()` installs `setupDialogHandler` but removes it as soon as setup returns](https://github.com/super-productivity/super-productivity/blob/c8bfd684e3168d0b3c55dc9d417fa52139acaaef/e2e/pages/supersync.page.ts#L202-L214). The page-lifetime handler in [`e2e/utils/runtime-errors.ts`](https://github.com/super-productivity/super-productivity/blob/c8bfd684e3168d0b3c55dc9d417fa52139acaaef/e2e/utils/runtime-errors.ts#L38-L70) intentionally handles only `beforeunload` and dev-error dialogs. Multiple tests later click `syncImportUseLocalBtn` or `syncImportUseRemoteBtn` directly. Production opens a native confirmation for never-synced `USE_LOCAL` and destructive incoming `USE_REMOTE` choices. At those direct button-click sites, no scoped listener accepts the confirmation, so the prompt remains open and the click or subsequent dialog-close wait stalls.

This matches the current required CI: [SuperSync shard 3](https://github.com/super-productivity/super-productivity/actions/runs/29665092245/job/88134357887) has 13 failures and [shard 6](https://github.com/super-productivity/super-productivity/actions/runs/29665092245/job/88134357910) has one more, with the failures stalling on the import-conflict choice or waiting for that dialog to close.

Please add page-object methods that install only the exact expected native-confirmation handler for the duration of each explicit choice, then migrate the direct button-click call sites. Keeping the strict setup handler is good; restoring the old broad page-lifetime auto-accept would hide unexpected destructive prompts again.

PR #9174, opened 2026-07-19
fix(planner): keep lazy-loaded dates consecutive

### LINKED ISSUE #8963: 🚨 Planner skips empty days at the initial lazy-load boundary when a later date contains tasks
### Version Used

18.13.1WB, firefox

### Version Used

Web app, current production version

### Expected Behavior

Planner dates should always remain consecutive.

For example:

July 25 → July 26 → July 27 → July 28

### Current Behavior

An empty date immediately after the initially loaded date range is skipped when a later date contains a planned task.

In my case, on July 12, the Planner displayed:

July 25 → July 26 → July 28

July 27 was completely absent.

### Suspected Cause

`PlannerService` initially generates 15 consecutive desktop days.

`selectPlannerDays()` then appends dates found in `plannerState.days` outside that range:

```ts
const dayDatesToUse = [
  ...dayDates,
  ...allDatesWithData
    .filter((d) => plannerState.days[d].length && !dayDates.includes(d))
    .sort((a, b) => a.localeCompare(b)),
];

### Expected Behavior

ehh... july 27 is properly displayed?

### Steps to Reproduce


### Steps to Reproduce

1. Open the desktop Planner.
2. Leave the date 15 days after today empty.
3. Add a planned task to the following date.
4. Reload or reopen the Planner.
5. Scroll to the end of the initially loaded dates.

The empty date between the initial 15-day range and the later date is missing.

Example when today is July 12:

- Initial 15-day range ends on July 26.
- July 27 contains no task.
- July 28 contains a planned task.
- Planner renders July 26 followed directly by July 28.

Sometimes, it's intermittent.

### Console Output

```shell

```

### Documentation

lntutor:
## Problem

When a populated planner date falls just beyond the initially loaded range, `selectPlannerDays()` appends that date directly. This can render a sequence such as July 25, July 26, July 28 and omit the empty boundary date.

Closes #8963.

## Solution

Keep the planner render selector bounded to its requested `dayDates`, so the rendered list remains a contiguous loaded window. Calendar markers instead combine persisted planner dates with tasks and the currently loaded planner-day projections, so dates outside that window still appear in mobile navigation.

For snackbar feedback, query the requested day directly instead of relying on the bounded render stream. This preserves the item/time suffix when a task is scheduled on an out-of-window day.

Regression tests cover both the out-of-window calendar marker and snackbar suffix.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have included relevant changes to the documentation/wiki. (Not applicable: internal bug fix with no user-facing workflow change.)

## Verification

- `npm run test:file src/app/features/planner/planner.component.spec.ts` (7 passing)
- `npm run test:file src/app/features/planner/planner.service.spec.ts` (24 passing)
- `npm run test:file src/app/features/planner/store/planner.selectors.spec.ts`
- `npm run test:file src/app/features/config/store/global-config.start-of-next-day-hydration.spec.ts` (6 passing)
- `npm run checkFile` for all four changed TypeScript files
- `npm run lint`

### DIFF
--- a/src/app/features/planner/planner.component.spec.ts
+++ b/src/app/features/planner/planner.component.spec.ts
@@ -6,6 +6,7 @@ import { BehaviorSubject } from 'rxjs';
 import { DateService } from '../../core/date/date.service';
 import { LayoutService } from '../../core-ui/layout/layout.service';
 import { selectTaskFeatureState } from '../tasks/store/task.selectors';
+import { selectPlannerDayMap } from './store/planner.selectors';
 import { signal } from '@angular/core';
 import { PlannerDay } from './planner.model';
 
@@ -60,6 +61,10 @@ describe('PlannerComponent', () => {
             {
               selector: selectTaskFeatureState,
               value: { ids: ['task-1', 'task-2'], entities: {} },
+            },
+            {
+              selector: selectPlannerDayMap,
+              value: {},
             },
           ],
         }),
@@ -149,6 +154,24 @@ describe('PlannerComponent', () => {
       const result = component.daysWithTasks();
 
       expect(result.size).toBe(0);
+    });
+
+    it('should include persisted task dates outside the loaded render window', () => {
+      const outOfWindowDay = '2026-02-26';
+      days$.next([
+        makePlannerDay('2026-02-16', 0),
+        makePlannerDay('2026-02-17', 0),
+        makePlannerDay('2026-02-18', 0),
+        makePlannerDay('2026-02-19', 0),
+        makePlannerDay('2026-02-20', 0),
+      ]);
+      TestBed.inject(MockStore).overrideSelector(selectPlannerDayMap, {
+        [outOfWindowDay]: makePlannerDay(outOfWindowDay, 1).tasks,
+      });
+      TestBed.inject(MockStore).refreshState();
+      fixture.detectChanges();
+
+      expect(component.daysWithTasks().has(outOfWindowDay)).toBeTrue();
     });
   });
 });
--- a/src/app/features/planner/planner.component.ts
+++ b/src/app/features/planner/planner.component.ts
@@ -11,6 +11,7 @@ import { CdkScrollable } from '@angular/cdk/scrolling';
 import { PlannerCalendarNavComponent } from './planner-calendar-nav/planner-calendar-nav.component';
 import { PlannerService } from './planner.service';
 import { LayoutService } from '../../core-ui/layout/layout.service';
+import { selectPlannerDayMap } from './store/planner.selectors';
 
 @Component({
   selector: 'planner',
@@ -33,22 +34,31 @@ export class PlannerComponent {
   readonly T = T;
 
   private _days = toSignal(this._plannerService.days$, { initialValue: [] });
+  private _plannerDayMap = toSignal(this._store.select(selectPlannerDayMap), {
+    initialValue: {},
+  });
   private _prevDaysWithTasksKey = '';
   private _prevDaysWithTasks: ReadonlySet<string> = new Set();
   daysWithTasks = computed<ReadonlySet<string>>(() => {
     const days = this._days();
-    const dayDates: string[] = [];
+    const plannerDayMap = this._plannerDayMap();
+    const dayDates = new Set(
+      Object.entries(plannerDayMap)
+        .filter(([, tasks]) => tasks.length > 0)
+        .map(([dayDate]) => dayDate),
+    );
     for (const day of days) {
       if (day.tasks.length > 0) {
-        dayDates.push(day.dayDate);
+        dayDates.add(day.dayDate);
       }
     }
-    const key = dayDates.join(',');
+    const sortedDayDates = [...dayDates].sort();
+    const key = sortedDayDates.join(',');
     if (key === this._prevDaysWithTasksKey) {
       return this._prevDaysWithTasks;
     }
     this._prevDaysWithTasksKey = key;
-    this._prevDaysWithTasks = new Set(dayDates);
+    this._prevDaysWithTasks = new Set(sortedDayDates);
     return this._prevDaysWithTasks;
   });
 
--- a/src/app/features/planner/planner.service.spec.ts
+++ b/src/app/features/planner/planner.service.spec.ts
@@ -126,6 +126,7 @@ describe('PlannerService', () => {
 
   afterEach(() => {
     jasmine.clock().uninstall();
+    TestBed.inject(MockStore).resetSelectors();
   });
 
   describe('tomorrow$', () => {
--- a/src/app/features/planner/planner.service.spec.ts
+++ b/src/app/features/planner/planner.service.spec.ts
@@ -1,18 +1,24 @@
 import { TestBed } from '@angular/core/testing';
 import { PlannerService } from './planner.service';
 import { DateService } from '../../core/date/date.service';
-import { provideMockStore } from '@ngrx/store/testing';
+import { MockStore, provideMockStore } from '@ngrx/store/testing';
 import { of, BehaviorSubject, ReplaySubject } from 'rxjs';
 import { CalendarIntegrationService } from '../calendar-integration/calendar-integration.service';
 import { GlobalTrackingIntervalService } from '../../core/global-tracking-interval/global-tracking-interval.service';
-import { selectAllTasksWithDueTime } from '../tasks/store/task.selectors';
-import { selectAllTaskRepeatCfgs } from '../task-repeat-cfg/store/task-repeat-cfg.selectors';
+import {
+  selectAllTasksWithDueTime,
+  selectMapOfAllTasksInActiveProjects,
+} from '../tasks/store/task.selectors';
+import { selectActiveTaskRepeatCfgs } from '../task-repeat-cfg/store/task-repeat-cfg.selectors';
 import { selectTodayTaskIds } from '../work-context/store/work-context.selectors';
 import { PlannerDay } from './planner.model';
 import { first, map, shareReplay } from 'rxjs/operators';
 import { getDbDateStr } from '../../util/get-db-date-str';
 import { signal, WritableSignal } from '@angular/core';
 import { LayoutService } from '../../core-ui/layout/layout.service';
+import { selectPlannerState } from './store/planner.selectors';
+import { selectTimelineConfig } from '../config/store/global-config.reducer';
+import { selectStartOfNextDayDiffMs } from '../../root-store/app-state/app-state.selectors';
 
 describe('PlannerService', () => {
   let service: PlannerService;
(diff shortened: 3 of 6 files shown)


johannesjo (changes requested):
**Important — `src/app/features/planner/store/planner.selectors.ts:115`:** This loop expands the selector output by calendar distance to the farthest populated `plannerState.days` key. The planning UI has no maximum date, and `2099-12-31` is already used as a valid future planner day in the reducer tests. Starting from `2026-07-26`, this creates 26,821 additional dates. Every date then passes through `getPlannerDay()` and is rendered as a `<planner-day>`, bypassing `PlannerService`’s intentional 5/15-day initial window and seven-day lazy-load increments; a legitimate far-future task can therefore make the Planner unusable.

Please keep selector output bounded by the service-controlled `dayDates` range and let the existing lazy-loading/`ensureDayLoaded()` path advance that range. The out-of-range planner-date append predates infinite scrolling, so removing that legacy path appears to be the smallest clean fix. Please also add a far-future regression test proving selector work is not proportional to arbitrary date distance.

The new boundary regression test correctly covers #8963, and the local-calendar `setDate()` stepping itself is DST-safe.

lntutor:
Addressed the selector-range feedback in 25e980245.

- Removed the legacy expansion from `plannerState.days`; `selectPlannerDays` now renders only the service-controlled `dayDates` range.
- Replaced the expansion assertions with a `2099-12-31` regression test, which confirms a far-future planner entry does not create intermediate planner days.

Validated with:
- `npm run test:file -- src/app/features/planner/store/planner.selectors.spec.ts` — 37 passing (Europe/Berlin)
- `TZ=America/Los_Angeles ./node_modules/.bin/ng test --watch=false --include src/app/features/planner/store/planner.selectors.spec.ts` — 37 passing
- `npx eslint src/app/features/planner/store/planner.selectors.ts src/app/features/planner/store/planner.selectors.spec.ts`
- Full local pre-push gate: `npm run lint && npm test`

lntutor:
@johannesjo I addressed the bounded-selector feedback in the latest commit: the legacy planner-state expansion is removed, and the regression now proves a far-future task does not expand the rendered date range. Focused tests pass in both Europe/Berlin and America/Los_Angeles; the PR CI is running. Could you please re-review when available?

johannesjo:
Thanks — `25e980245` addresses the previous unbounded-expansion concern, and the
far-future selector test correctly keeps rendered output bounded. I double-checked
the downstream consumers and found one remaining issue before merge:

**Important — `src/app/features/planner/store/planner.selectors.ts:122`: preserve
mobile calendar task markers outside the loaded render window.**

`PlannerService` initially loads only 5 days on mobile, while
`PlannerCalendarNavComponent` displays 5 weeks. `PlannerComponent.daysWithTasks`
derives its marker set exclusively from `PlannerService.days$`. Now that
`selectPlannerDays()` returns only `dayDates`, a valid task planned for (for example)
today + 10 days remains stored but has no task dot in the visible mobile calendar
until the user taps or scrolls far enough to load that date.

Please keep the rendered planner days bounded, but derive calendar marker dates
independently of that render window. The existing `selectPlannerDayMap` looks like
the smallest starting point; loaded-day results can be unioned where needed to
preserve Today semantics. A regression test should cover a 5-day render window with
a populated date later in the 35-day calendar view.

**Minor follow-up:** `PlannerService.getDayOnce$()` also searches only bounded
`days$`, so `getSnackExtraStr()` now returns an empty suffix when a task is planned
for an out-of-window date. The task is saved correctly, but the existing item/time
summary disappears from the success snackbar. Computing exactly the requested day
through a targeted single-day selector would preserve the feedback without
expanding the rendered list; a focused out-of-window test would be useful.

Also, the PR description's **Solution** section still describes the superseded
calendar-expansion implementation and should be updated to match the current
bounded-selector approach.

All 21 current CI checks are passing.

lntutor:
The two follow-ups from the latest review are implemented in `d785d0c4b`:\n\n- `PlannerComponent.daysWithTasks` now unions persisted `selectPlannerDayMap` dates with loaded planner-day results, preserving mobile calendar markers outside the 5-day render window.\n- `PlannerService.getDayOnce$()` now evaluates a targeted single-day selector, preserving the snackbar item/time suffix without expanding `days$`.\n\nRegression coverage was added for a task at today + 10 outside the loaded window and for out-of-window snackbar feedback. The PR Solution section now describes the bounded render selector plus the independent marker and single-day paths. Current GitHub checks have no failures.

johannesjo:
Thanks — `d785d0c4b` implements both follow-ups correctly. I sabotage-verified that each
new test genuinely fails without its fix:

| revert | test that goes red |
| --- | --- |
| `getDayOnce$` back to `this.days$` | `getSnackExtraStr` resolves `''` instead of `' – ∑ 1'` |
| drop the `selectPlannerDayMap` union | marker test — `Expected false to be true` |
| restore the `allDatesWithData` expansion | far-future test — `Expected $.length = 3 to equal 2` |

The marker union is also the right shape: for every non-today day, `getPlannerDay()`'s
`normalTasks` and `selectPlannerDayMap[day]` resolve from the same `plannerState.days[day]`
through the same task map, so the `days$` half is load-bearing only for Today — whose key
`cleanupOldAndUndefinedPlannerTasks` deletes (`planner.reducer.ts:84-86`). A plain union
rather than a Today special case is the simpler call; leave it as is.

One blocker before merge.

**Important — `src/app/features/planner/planner.service.spec.ts:127`: the new
`provideMockStore` overrides leak into other spec files and turn
`global-config.start-of-next-day-hydration.spec.ts` red.**

Because `getDayOnce$` now evaluates the real selectors instead of the mocked `days$`, the
spec freezes five more selectors than before (`selectActiveTaskRepeatCfgs`,
`selectMapOfAllTasksInActiveProjects`, `selectPlannerState`, `selectTimelineConfig`,
`selectStartOfNextDayDiffMs`) — but its `afterEach` still only does
`jasmine.clock().uninstall()`. `MockStore.overrideSelector()` calls `setResult()` on the
module-level selector singleton and `MockStore` has no `ngOnDestroy`, so those frozen values
survive `TestBed.resetTestingModule()` for the rest of the Karma session. Jasmine's spec
order is randomized and `src/karma.conf.js` pins no seed, so this is a coin flip rather than
a guaranteed failure — which is why the 21/21 green CI does not clear it.

Repro, ~25s per run:

```bash
npx ng test --watch=false \
  --include src/app/features/planner/planner.service.spec.ts \
  --include src/app/features/config/store/global-config.start-of-next-day-hydration.spec.ts
```

On `d785d0c4b` this failed **6 of 10 runs**, always the same four specs:

```
installs the offset when the config op is replayed at startup
  Expected 0 to be 21600000.
  at global-config.start-of-next-day-hydration.spec.ts:175
```

Note which line fails. On `:174`, `expect(dateService.getStartOfNextDayDiffMs()).toBe(SIX_AM_MS)`
passes; on `:175`, `expect(readAppStateOffset()).toBe(SIX_AM_MS)` fails. The offset is correct
everywhere except when read through `selectStartOfNextDayDiffMs` — i.e. exactly the leaked
`value: 0` from `planner.service.spec.ts:71`. A full `ng test` on the branch failed 2 of 3 runs
with these same four.

The fix is one line — the same one `planner.component.spec.ts:83` in this PR already has, and
that `work-view.component.spec.ts:53-60` documents at length:

```ts
  afterEach(() => {
    jasmine.clock().uninstall();
    TestBed.inject(MockStore).resetSelectors();
  });
```

With it, the pair above passed **10/10**, and the full suite passed **13574 SUCCESS** twice
(vs 4 FAILED without). Please also drop this line from the PR body — that failure is this
PR's own, not an unrelated flake:

> The local full-suite pre-push run has an order-dependent failure in the unrelated config
> hydration spec; it passes in isolation.

**Behaviour note — no change requested**, just worth recording as a conscious trade-off: with
the render window bounded, a populated day beyond it is no longer drawn on desktop until
infinite scroll reaches it (7 days per batch), and desktop has no calendar-nav equivalent —
`ensureDayLoaded()` is only reachable through the mobile calendar's `(dayTapped)`. The task
itself is not orphaned (`task.reducer.ts:695` still sets `dueDay` for `planTaskForDay`), so
this is about the column, not the task. Bounded and contiguous is clearly the right default,
and an "N more planned days →" affordance would be feature creep — so I would ship it as is;
just flagging that one visibility gap is traded for a milder one.

lntutor:
@johannesjo Addressed the remaining blocker in 0f3094d: `PlannerService` tests now reset `MockStore` selector overrides in `afterEach`, matching the existing planner component pattern. The reviewer-reproduced paired planner/config test passes locally after the change, and ESLint is clean. I also removed the obsolete full-suite flake claim from the PR body.

lntutor:
Follow-up after the latest base update: CI exposed duplicate `lastTaskCreation` / `lastTaskCreationDay` keys in the upstream schedule fixture. I rebased onto that base and removed the superseded `undefined` pair, retaining the maintainer’s newer fixed 2026-07-29 markers. The exact affected schedule spec passes locally (5/5); the refreshed full CI matrix is in progress. The earlier selector-reset review fix remains included. Could you please re-review when convenient?

johannesjo (dismissed):
`0f3094d` closes the blocker exactly as intended. I verified it locally rather than reading
CI, for the reason in the last section.

**Verified.** The paired repro from my previous comment, 10 runs each:

| tree | result |
| --- | --- |
| `0f3094d` | **10/10 green**, `TOTAL: 30 SUCCESS` every run |
| `0f3094d` with the new `afterEach` line removed | **7/10 red**, always exactly `4 FAILED, 26 SUCCESS` |

The negative control matters as much as the green one. On its own, 10/10 green would only show
that the pair had stopped being order-sensitive, not that the leak was closed. Taking the line
back out still reproduces the same four specs, at the same four lines, with the same value:

```
✗ installs the offset when the config op is replayed at startup
    Expected 0 to be 21600000.
    at global-config.start-of-next-day-hydration.spec.ts:175:34
✗ installs the offset when the config change arrives from another device   (:185)
✗ installs the offset when a full-state operation is replayed              (:195)
✗ installs the offset when a repair operation is replayed                  (:205)
```

That is the leaked `selectStartOfNextDayDiffMs → 0` from `planner.service.spec.ts:71`, so the
new `resetSelectors()` call is load-bearing and not decorative.

Full suite on `0f3094d`: `Executed 13574 of 13576 (skipped 2) SUCCESS`. `checkFile` on the
changed spec is clean, and the branch merges into current `master` without conflicts.

Thanks also for dropping the "unrelated config hydration spec" line from the PR body.

**The one red check is mine, not yours.** `Tests` fails on `master`, not on this branch:

```
src/app/features/schedule/map-schedule-data/create-blocked-blocks-by-day-map.spec.ts:77:7
```

Two commits fixed the same wall-clock-dependent fixture independently, at different anchor
points inside one object literal, so git auto-merged both insertions without reporting a
conflict and `lastTaskCreation` / `lastTaskCreationDay` ended up declared twice. Your branch
does not contain either edit, so nothing here caused it. I am pushing the repair.

It is worth knowing why this matters beyond the red X. A compile error in any single spec
fails the whole Karma bundle, and the job stops there:

```
failure  Run npm run test
skipped  Build Frontend for E2E
skipped  Test E2E
skipped  Test production PWA offline reload
```

So that run executed zero client unit tests and never reached the Playwright suite. Your
planner specs did not actually run in CI, which is why I re-ran them locally this round rather
than treating a check mark as evidence. Please re-run `Tests` once the repair lands on
`master`.

Approving. This ended up a tight fix: the render window is bounded and contiguous, markers and
snackbar feedback are derived independently of that window rather than by widening it, and all
three behaviours carry tests that fail when their fix is reverted.

lntutor:
I rebased onto the newly landed master repair (#9396) and dropped my now-redundant fixture-key deletion, so this PR no longer overrides that upstream choice. The branch now contains only the planner changes and selector reset (`516a977`). Focused planner, hydration, and schedule suites pass 35/35 locally against current master; fresh CI is running on this head.

johannesjo:

PR #9176, opened 2026-07-19
fix(android): close task context menu on back

### LINKED ISSUE #8440: 🚨 Pressing back button on Android app does not clear issue provide panel or task settings menu
### Version Used

18.10.0

### Current Behavior

This Issue is a followup for #6638.
It seems that some panels got missed in PR #8176.

Case A. If I open the Issue Provider Panel and press back, nothing hapens. The panel does not go away, nor the application is closed.

Case B. If I open the task menu (swipe left on any task to open) and press back, the whole application is closed, but If I reopen the application, the task menu is still there.

### Expected Behavior

Case A. The Issue Provider Panel should go away.

Case B. The task menu should go away and the application should not be closed.

### Steps to Reproduce

See above.

### Console Output

```shell

```

### Documentation

lntutor:
## Problem

Android back handling already closes the issue-provider panel, but a task context menu opened by swiping left is a CDK overlay without history state. Pressing back therefore fell through to navigation or app minimization and left the menu open.

Closes #8440.

## Solution

Expose a focused close operation on the task context-menu wrapper and have `AndroidBackButtonService` close the active menu before processing history, dialogs, navigation, or app exit.

Add regression coverage proving the app is not minimized while the task menu is being dismissed.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have included relevant changes to the documentation/wiki. (Not applicable: restores expected Android back behavior.)

## Verification

- `npm run checkFile src/app/features/tasks/task-context-menu/task-context-menu.component.ts`
- `npm run checkFile src/app/features/android/android-back-button.service.ts`
- `npm run checkFile src/app/features/android/android-back-button.service.spec.ts`
- `npm run test:file src/app/features/android/android-back-button.service.spec.ts`: 41 tests passed

### DIFF
--- a/src/app/features/android/android-back-button.service.ts
+++ b/src/app/features/android/android-back-button.service.ts
@@ -11,6 +11,7 @@ import { selectIsOverlayShown } from '../focus-mode/store/focus-mode.selectors';
 import { selectAllProjects } from '../project/store/project.selectors';
 import { hideFocusOverlay } from '../focus-mode/store/focus-mode.actions';
 import { LayoutService } from '../../core-ui/layout/layout.service';
+import { TaskFocusService } from '../tasks/task-focus.service';
 
 /**
  * Implements Android back-button behavior for top-level (bottom-nav) destinations
@@ -36,6 +37,7 @@ export class AndroidBackButtonService {
   private readonly _store = inject(Store);
   private readonly _matDialog = inject(MatDialog);
   private readonly _layoutService = inject(LayoutService);
+  private readonly _taskFocusService = inject(TaskFocusService);
 
   private readonly _isFocusOverlayShown = this._store.selectSignal(selectIsOverlayShown);
   private readonly _allProjects = this._store.selectSignal(selectAllProjects);
@@ -47,14 +49,19 @@ export class AndroidBackButtonService {
       return;
     }
 
-    // 2. An overlay that pushed a history state is open → let its popstate
+    // 2. Task context menus are CDK overlays without a history entry.
+    if (this._closeTaskContextMenu()) {
+      return;
+    }
+
+    // 3. An overlay that pushed a history state is open → let its popstate
     //    listener close it.
     if (this._isHistoryOverlayOpen()) {
       this._historyBack();
       return;
     }
 
-    // 3. A modal dialog without its own history state is open → dismiss the
+    // 4. A modal dialog without its own history state is open → dismiss the
     //    topmost one instead of navigating/minimizing underneath it. A dialog
     //    that opted out of dismissal (`disableClose`) swallows the back press,
     //    mirroring its escape-key behavior. History-backed dialogs (e.g.
@@ -67,13 +74,13 @@ export class AndroidBackButtonService {
       return;
     }
 
-    // 4. Store-backed layout popups do not push history state or open a
+    // 5. Store-backed layout popups do not push history state or open a
     //    MatDialog, so close them before falling through to navigation/exit.
     if (this._closeStoreBackedPopup()) {
       return;
     }
 
-    // 5. Not a top-level destination (context sub-page, settings, search, …)
+    // 6. Not a top-level destination (context sub-page, settings, search, …)
     //    → navigate up via the history stack as before.
     const currentUrl = this._router.url;
     if (!this._isTopLevelDestination(currentUrl)) {
@@ -85,7 +92,7 @@ export class AndroidBackButtonService {
       return;
     }
 
-    // 6. A top-level destination → pop to the start destination, or exit if
+    // 7. A top-level destination → pop to the start destination, or exit if
     //    already there.
     const startUrl = this._getStartPageUrl();
     if (this._pathOf(currentUrl) === this._pathOf(startUrl)) {
@@ -107,6 +114,17 @@ export class AndroidBackButtonService {
     }
 
     return false;
+  }
+
+  private _closeTaskContextMenu(): boolean {
+    if (!this._taskFocusService.isTaskContextMenuOpen()) {
+      return false;
+    }
+
+    return (
+      this._taskFocusService.lastFocusedTaskComponent()?.taskContextMenu()?.close() ??
+      false
+    );
   }
 
   private _isHistoryOverlayOpen(): boolean {
--- a/src/app/features/android/android-back-button.service.ts
+++ b/src/app/features/android/android-back-button.service.ts
@@ -117,14 +117,13 @@ export class AndroidBackButtonService {
   }
 
   private _closeTaskContextMenu(): boolean {
-    if (!this._taskFocusService.isTaskContextMenuOpen()) {
+    const close = this._taskFocusService.closeActiveTaskContextMenu();
+    if (!close) {
       return false;
     }
 
-    return (
-      this._taskFocusService.lastFocusedTaskComponent()?.taskContextMenu()?.close() ??
-      false
-    );
+    close();
+    return true;
   }
 
   private _isHistoryOverlayOpen(): boolean {
--- a/src/app/features/android/android-back-button.service.spec.ts
+++ b/src/app/features/android/android-back-button.service.spec.ts
@@ -1,3 +1,4 @@
+import { signal } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 import { Router } from '@angular/router';
 import { MatDialog, MatDialogRef } from '@angular/material/dialog';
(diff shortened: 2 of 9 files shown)


src/app/features/android/android-back-button.service.ts
```diff
@@ -109,6 +116,17 @@ export class AndroidBackButtonService {
     return false;
   }
 
+  private _closeTaskContextMenu(): boolean {
+    if (!this._taskFocusService.isTaskContextMenuOpen()) {
+      return false;
+    }
+
+    return (
+      this._taskFocusService.lastFocusedTaskComponent()?.taskContextMenu()?.close() ??
```
johannesjo: **Important:** This can still fall through for the reported swipe path because isTaskContextMenuOpen and lastFocusedTaskComponent are not linked. The swipe template calls openContextMenu() without focusing the row; the pan handler prevents the touch default, while lastFocusedTaskComponent is set only by TaskComponent.onFocus(). On a first swipe it can therefore be null, or it can still point at another task, making this expression return false so the same back press continues to navigation/minimization while the actual menu remains open. Planner and schedule consumers of the shared context menu do not populate this TaskComponent reference either. Please track the active TaskContextMenuComponent or close callback when the menu opens, clear it on close/destruction, and add regression coverage with null/stale last focus that closes the actual trigger. The current test always supplies a matching focused component, so it cannot catch this failure.

lntutor: Fixed in 313273f02. TaskComponent.openContextMenu() now synchronously registers the component in TaskFocusService before opening the menu, so the swipe path (which calls openContextMenu() without first focusing the row) gives Android Back the exact component whose menu is open. This is interaction-time only and adds no template/change-detection hot-path work. Added a TaskComponent regression that calls openContextMenu() directly and asserts lastFocusedTaskComponent is updated; the focused Android Back suite still passes (41/41), and the task shortcut suite passes (31/31).

lntutor: Follow-up after rebasing onto current master: the swipe-focus regression is now db32af90f, and 57fe4c93e adapts the menu close path to the upstream isOpen signal introduced by #9272. Fresh verification on the rebased branch: context-menu component 3/3, Android Back 41/41, task shortcuts 31/31, all six changed TypeScript files pass checkFile, and npm run buildFrontend:prod:es6 completes successfully.

johannesjo (commented):
Reviewed at `57fe4c93e0`. The Today/tag/project list path works, and both new assertions are real guards (sabotage-verified: dropping the `task.component.ts` line fails `task-shortcuts.spec.ts`; dropping `trigger.closeMenu()` fails `task-context-menu.component.spec.ts`). Step 2 is also correctly ordered before the history-overlay and dialog checks. Thanks for the rework.

But the second half of the earlier review — *"Planner and schedule consumers of the shared context menu do not populate this TaskComponent reference either … clear it on close/destruction"* — is still open, and I think it points at a design that is **less** code than what is here now.

## Important — the same swipe-left gesture is still unfixed in Planner and Schedule

`lastFocusedTaskComponent` is typed `TaskComponent | null` (`task-focus.service.ts:9`), but `isTaskContextMenuOpen` is set by the *shared inner* component (`task-context-menu-inner.component.ts:252`), which is also mounted by `planner-task`, `schedule-event` and `task-detail-panel`. So the flag goes true for all four surfaces while the reference can only ever be populated by one.

`planner-task.component.html:8` binds **the same swipe-left gesture the issue reports** to `planner-task.component.ts:201`, and the Planner renders `<planner-task>` — never `<task>` — so no `TaskComponent` can be registered there. `/planner` and `/schedule` are top-level Android destinations (`android-back-button.service.ts:151`), so `_closeTaskContextMenu()` returns `false` and Back falls through to step 6/7: navigate away to the start page, or minimize the app if the start page *is* that tab. Schedule is the same via long-press (`schedule-event.component.ts:451`).

I added the two cases the earlier review asked for to this PR's own spec; **both fail on `57fe4c93e0`**:

```ts
it('does not fall through when no TaskComponent is registered', () => {
  isTaskContextMenuOpen.set(true);
  lastFocused.set(null);                    // planner / schedule swipe-left
  service.handleBackButton();
  expect(minimizeApp).not.toHaveBeenCalled();          // FAILS
});

it('does not fall through when the registered component menu is closed', () => {
  isTaskContextMenuOpen.set(true);
  lastFocused.set({ taskContextMenu: () => ({ close: () => false }) });
  service.handleBackButton();
  expect(minimizeApp).not.toHaveBeenCalled();          // FAILS
});
```

The current mock hard-wires a matching component (`android-back-button.service.spec.ts:118-120`), so the suite structurally cannot fail for either case.

## Secondary — the registration is never cleared

`MatMenu.ngOnDestroy` calls `closed.complete()` **without emitting**, and `MatMenuTrigger.ngOnDestroy` only disposes the overlay (`@angular/material@21.2.12`, `menu.mjs:399-405` / `746-757`). So if a row is destroyed while its menu is open (a remote sync delete, a list re-render), `(close)` never fires and both `isTaskContextMenuOpen` and `TaskContextMenuComponent.isOpen` stay `true`. `viewChild` signals also keep resolving after view destruction — I measured it: after `fixture.destroy()`, `component.isOpen()` is still `true` and `component.close()` still returns `true`, no throw.

Since `openContextMenu()` now points `lastFocusedTaskComponent` at exactly the row that gets destroyed, `_closeTaskContextMenu()` can then return `true` on every press — Back silently does nothing with no overlay on screen. Narrow precondition, but before this PR a stuck flag only disabled arrow-key nav; now it can no-op the hardware Back button. Clearing on destroy removes the whole class.

## Suggested direction — one registration point, ~10 lines, smaller than this diff

Register the close callback where the flag is *already* toggled and the trigger is already in scope, instead of plumbing a component reference through `TaskComponent`:

```ts
// task-focus.service.ts
readonly closeActiveTaskContextMenu = signal<(() => void) | null>(null);

// task-context-menu-inner.component.ts — open(), beside line 252
this._taskFocusService.isTaskContextMenuOpen.set(true);
this._taskFocusService.closeActiveTaskContextMenu.set(() =>
  this.contextMenuTrigger()?.closeMenu(),
);

// onClose() (line 337) and ngOnDestroy() (line 211)
this._taskFocusService.closeActiveTaskContextMenu.set(null);

// android-back-button.service.ts
private _closeTaskContextMenu(): boolean {
  const close = this._taskFocusService.closeActiveTaskContextMenu();
  if (!close) return false;
  close();
  return true;
}
```

A callback rather than a component reference keeps `TaskFocusService` free of an import cycle. (If you want to be strict, guard the `ngOnDestroy` clear on identity so a late destroy can't null out a newer registration.)

This covers all four consumers, self-clears, and makes three things in the current diff unnecessary: `TaskContextMenuComponent.close()`, the `task.component.ts` line, and the `lastFocusedTaskComponent` read in the back handler. Net: **2 files instead of 3, and no new public method** — the existing flag becomes the single source of truth instead of a second, parallel one.

## Minor

- **`task.component.ts:1253`** — this gives `lastFocusedTaskComponent` a second, implicit meaning ("last focused **or context-menu'd**"). It is read by `task-shortcut.service.ts:139` (Ctrl+C guard), `:409` (`_handleTaskShortcut` id guard) and `:439` (`_isTaskContextMenuOpen`), none of which expect it to move without focus. Swipe-left does not focus the row, so it can now diverge from `focusedTaskId` and make `_handleTaskShortcut` log *"does not match"* and no-op. Long-term this is the kind of overload that is cheap to add and expensive to reason about later; the callback approach avoids touching this signal at all.
- **`task-context-menu.component.spec.ts:60-75`** — the spec blanks the template, so the new test monkey-patches `taskContextMenuInner` and never exercises the real `viewChild`; the `isOpen() === false` and missing-trigger branches are untested.
- Heads-up: open PR #9235 also touches `task-context-menu`.

Happy to be wrong on the Planner reachability if that surface is considered out of scope for #8440 — but it is literally the same gesture one tab over, and the fix above costs less than the current one.

lntutor:
Reworked the fix in f271511b8 around the shared inner menu, covering Planner and Schedule as requested. TaskFocusService now holds the active close callback; TaskContextMenuInnerComponent registers it on open and clears only its own registration on close or destruction; AndroidBackButtonService invokes that callback directly. The TaskComponent focus overload and TaskContextMenuComponent proxy method are removed. Regression tests cover callback registration/close and late destruction preserving a newer registration. Focused suites pass: Android back 41/41, inner menu 15/15, wrapper menu 2/2, and task shortcuts 30/30; all modified TypeScript files pass checkFile.

johannesjo:
Thank you very much! <3

PR #9177, opened 2026-07-19
fix(metrics): remove productivity score drop when focus passes target

davidvornholt:
## Problem

Hi! 👋 Thanks for all the work on Super Productivity — the finish-day evaluation sheet is a lovely feature, and while digging into it I found a small but demotivating quirk in the productivity score.

The "soft-cap for diminishing returns" in `calculateProductivityScore` clamps its input to [0, 1], so any focus ratio above 1 collapses to a constant `1 − e^(−2.2) ≈ 0.889` focus credit. In practice that means the score **drops the moment you pass your 4h focus target**:

| Focus (impact 3, work = focus) | Score |
| --- | --- |
| 240 min (at target) | 81 |
| 241 min | **77** |
| 480 min | 79 |

Working one more minute past your deep-work goal costs ~4 points, and the score never recovers no matter how much more focus is logged. The docs describe a smooth soft-cap; the actual curve is a cliff followed by a flat line — right at the moment a user hits their goal.

## Solution

- **Make focus progress continuous and non-decreasing:** linear up to the target, full credit at the target, flat beyond it (`clamp(progressRatio)`). Doing more focused work can never lower the productivity score; overwork stays penalized where it belongs — in the sustainability score (workload sigmoid + focus-balance decay). A new spec pins monotonicity past the target and fails against the previous implementation.
- **Preserve the `IMPACT_MISMATCH` daily state:** with the flat cap, a low-impact day with extended focus can now cross the productivity threshold, so `getDailyStateInfo` checks for the mismatch before `DEEP_FLOW`. Added unit tests for the classification order.
- **Remove dead code the old curve left behind:** the unused `softCap` helper, `SOFT_CAP` and `EXHAUSTION_SCALE_MAX` constants, and the `workloadLinearZeroAt` parameter of `calculateSustainabilityScore`, which has been ignored since the sigmoid rewrite (all four call sites passed a `600` that did nothing — one with a comment claiming "10h → score 0", which the sigmoid made untrue). No score values change from this cleanup.
- **Freshen the docs:** renamed `DEFAULT_WORKLOAD_LINEAR_ZERO_AT` → `DEFAULT_MAX_WORK_MINUTES` to match what it actually does, fixed JSDoc referencing the removed `exhaustion` parameter and stale example values, and updated the wiki section on focus balance to describe the corrected behavior.

One intentional behavior change worth calling out: a high-impact day with focus well past the target and a full workday now scores 100 instead of 97–99. That felt consistent with the existing design, since impact 4 + met target already scored 100 — but happy to adjust if you'd prefer overwork to stay just shy of perfect! Scores are computed at display time rather than persisted, so historical days simply re-render with the corrected value and nothing sync-relevant changes.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] Existing tests still pass (full suite: 13,295 unit tests green in both TZ runs)

### DIFF
--- a/docs/wiki/4.15-Timers-and-Focus-Mode.md
+++ b/docs/wiki/4.15-Timers-and-Focus-Mode.md
@@ -78,7 +78,7 @@ So “session done” means something different per mode: in Pomodoro it trigger
 
 ### Focus Balance and Sustainability
 
-The app’s **productivity and sustainability metrics** assume that **more focus time is not always better**. They use an **inverted-V curve** for focus balance: up to the optimal amount (e.g. 4 hours), more focus improves the score; beyond that, **excessive focus is penalized** (e.g. with an exponential decay). That reflects the reality that **cognitive fatigue is non-linear**: pushing far beyond the sustainable range does not improve outcomes and can hurt sustainability. So the system encourages you to value **sustainable** focus, not maximum hours.
+The app’s **sustainability metric** assumes that **more focus time is not always better**. It uses an **inverted-V curve** for focus balance: up to the optimal amount (e.g. 4 hours), more focus improves sustainability; beyond that, **excessive focus is penalized** (e.g. with an exponential decay). The **productivity metric** works differently: focus progress earns full credit at the target, then stays flat, so additional focus neither improves nor lowers productivity. That reflects the reality that **cognitive fatigue is non-linear**: pushing far beyond the sustainable range does not improve outcomes and can hurt sustainability. So the system encourages you to value **sustainable** focus, not maximum hours.
 
 ## How Timers Influence Your Behavior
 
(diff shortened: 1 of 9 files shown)


johannesjo (commented):
## Summary

This is a focused fix for the productivity-score cliff after the focus target. The `clamp(progressRatio)` change is the smallest clean correction, the classification order preserves the existing `IMPACT_MISMATCH` meaning, and the regression tests exercise the exact boundary with real score calculations.

## Issues

No Critical, Important, or Minor code findings.

## Verification

- The unit-test steps, lint, CodeQL, preview build, packaging, and required SuperSync/WebDAV gates passed.
- The overall `Tests` check is red because `e2e/tests/planner/planner-subtask-scheduled-visibility.spec.ts:82` timed out waiting for an unrelated planner task: 280 E2Es passed, 1 failed, and 1 was skipped.
- None of the PR files touch planner or task-scheduling behavior, so no PR code change is currently indicated. The failed job should be rerun; if it repeats, the planner test needs separate diagnosis.

## Verdict

**Needs verification (CI only).** The code review found no defects, but the PR should get a green rerun of the failed `Tests` job before merging.

johannesjo:

PR #9178, opened 2026-07-19
Fix "0 years ago" for timestamps 330-364 days in the past

Osamaali313:
## Problem

`humanizeTimestamp` (`src/app/util/humanize-timestamp.ts`) derives its month and year counts independently:

```ts
const diffMonths = Math.floor(diffDays / 30);
const diffYears = Math.floor(diffDays / 365);
```

The past-date branch then does:

```ts
} else if (diffMonths < 11) {                 // handles diffDays up to 329
  return ...PAST.MONTHS, { count: diffMonths };
} else if (diffYears === 1) {                  // <-- bug
  return ...PAST.A_YEAR;
} else {
  return ...PAST.YEARS, { count: diffYears };
}
```

For a timestamp **330–364 days** in the past, `diffMonths >= 11` (so the `MONTHS` branch is skipped) but `diffYears === 0` (so the `A_YEAR` branch is skipped), and it falls through to `YEARS` with **`count: 0`** → renders **"0 years ago"**.

## Evidence (intent)

- The adjacent `diffDays = 365` case already returns `A_YEAR` ("a year ago"), and the doc comment says the helper mirrors moment's `fromNow()` (which never says "0 years").
- The **future** branch is coincidentally correct: `Math.floor` rounds the *negative* diff toward −∞, so ~11 future months floor to `-1` → `abs → 1` → `A_YEAR`. Only the past branch, which floors `~0.9 → 0`, is broken. That asymmetry confirms the past branch is the defect.

## Fix

```ts
} else if (diffYears <= 1) {
```

This routes 330–729 days-past to "a year ago" and ≥ 730 days to "N years ago". It only *adds* the broken 330–364 window — the 365–729 range already returned `A_YEAR`, so there's no behavior change there.

## Reproduction

| days ago | before | after |
|---|---|---|
| 300 | 10 months ago | 10 months ago |
| 330 / 340 / 364 | **0 years ago** ❌ | a year ago ✅ |
| 365 | a year ago | a year ago |
| 730 | 2 years ago | 2 years ago |

Reachable via `HumanizeTimestampPipe` on the scheduled-list page (`task.dueWithTime` / `task.deadlineWithTime`). The existing spec tests 30d/180d/365d/3y, jumping over the 11-month gap.

### DIFF
--- a/src/app/util/humanize-timestamp.ts
+++ b/src/app/util/humanize-timestamp.ts
@@ -98,7 +98,7 @@ export const humanizeTimestamp = (
     return translateService.instant(T.GLOBAL_RELATIVE_TIME.PAST.MONTHS, {
       count: diffMonths,
     });
-  } else if (diffYears === 1) {
+  } else if (diffYears <= 1) {
     return translateService.instant(T.GLOBAL_RELATIVE_TIME.PAST.A_YEAR);
   } else {
     return translateService.instant(T.GLOBAL_RELATIVE_TIME.PAST.YEARS, {


johannesjo (commented):
## Review

The production change is correct and minimal. On the past-date path, `diffYears` cannot be negative and the preceding month branch excludes values below 330 days, so `<= 1` changes only the intended 330–364-day window.

### Important

**`src/app/util/humanize-timestamp.ts:101` — Add regression coverage for the corrected boundary.** This bug exists because the co-located spec skips directly from 180 to 365 days, and this PR does not change that spec. I ran the existing test file on the buggy base and all 22 tests passed, confirming that the suite cannot detect the reported “0 years ago” regression.

The smallest fix is to add a 330-day-past assertion expecting `a year ago` to the existing “should handle a year ago” test while retaining the 365-day case.

CI also still needs to execute: the five main workflow runs currently report `action_required` with no jobs run; only the welcome check has passed.

**Verdict: Needs changes.** Once the regression assertion is added and required CI passes, this should be ready to merge.

johannesjo:

PR #9183, opened 2026-07-20
test: reset global dialog state between specs

### LINKED ISSUE #9100: Test infra: global window.alert spy is never reset, making expect(window.alert).toHaveBeenCalled() vacuous in full-suite runs
## Summary

`src/test.ts` creates the global `window.alert` / `window.confirm` spies **once at module load** and never resets their call history. As a result, `expect(window.alert).toHaveBeenCalled()` can be satisfied by an alert raised by a **completely unrelated, earlier spec** — so the assertion proves nothing in a full-suite run, which is how CI runs.

This is a test-infrastructure trap, not a production bug: no user-facing behaviour is affected. But it silently weakens assertions in specs that guard real data-integrity paths, and it does so *invisibly* — the test stays green either way.

Found while reviewing #9079.

## The two mechanisms

They compound, which is why this is easy to miss:

1. **The spy's call log is never reset.** `src/test.ts` (~line 113):

   ```ts
   if (!(window.alert as jasmine.Spy).and) {
     window.alert = jasmine.createSpy('alert');
   }
   ```

   This runs at import time, not in a `beforeEach`. `calls` accumulate for the entire Karma run.

2. **`devError`'s `isShowAlert` latches off.** In `src/app/util/dev-error.ts`, `isShowAlert` flips to `false` after the *first* alert of the whole run and is never re-armed. So a later `devError` shows **no alert at all** — yet assertion (1) still passes, because of the stale accumulated call.

Net effect: a spec asserting "devError fired" passes whether or not `devError` fired.

## Proof

Using the new `deleteTaskHelper` orphan test from #9079, I deleted the entire `devError(...)` call from `src/app/features/tasks/store/task.reducer.util.ts` and ran the spec two ways:

| Configuration | Result |
| --- | --- |
| spec **alone** + sabotaged production code | `1 FAILED` — correctly catches it |
| spec **+ `task.reducer.spec.ts`** + sabotaged production code | **`75/75 SUCCESS` — misses it** |

The second row is the CI configuration. Note the trap for anyone verifying this: running the spec **standalone is exactly the configuration that hides the bug**, because standalone the spec happens to own the first alert of the run.

## Blast radius

12 spec files currently assert on the dialog spies, but only 4 reset the call log first:

<details>
<summary>Specs asserting on <code>window.alert</code> / <code>window.confirm</code></summary>

```
src/app/features/tasks/store/task.reducer.spec.ts
src/app/features/tasks/store/task.reducer.util.spec.ts        (new, #9079)
src/app/imex/file-imex/file-imex.component.spec.ts             ✔ resets
src/app/imex/local-backup/local-backup.service.spec.ts         ✔ resets
src/app/op-log/apply/archive-operation-handler.service.spec.ts
src/app/op-log/persistence/operation-log-hydrator.service.spec.ts  ✔ resets
src/app/op-log/sync/dialog-sync-import-conflict/dialog-sync-import-conflict.component.spec.ts
src/app/op-log/sync/rejected-ops-handler.service.spec.ts
src/app/op-log/validation/repair-operation.service.spec.ts
src/app/op-log/validation/sync-repair-non-blocking.integration.spec.ts
src/app/op-log/validation/validate-state.service.spec.ts
src/app/util/check-fix-entity-state-consistency.spec.ts
```

A further 15 files use the `jasmine.isSpy(window.alert)` pattern without asserting on it — those are unaffected but show how widely the idiom has spread.
</details>

To be clear about what is and isn't broken today: I mutation-tested the pre-existing sibling test in `task.reducer.spec.ts` (`'should call devError when orphan IDs are detected'`) and it **still catches** its own sabotage, because it calls `_resetDevErrorState()` first. So this is currently a **latent** trap rather than a live gap — but nothing prevents the next spec from falling into it, which is precisely what happened in #9079.

## Proposed fix (verified)

Re-arm both pieces of global state in the existing global `beforeEach` in `src/test.ts`, so every spec starts from a clean slate:

```diff
 import { clearDeferredActions } from './app/op-log/capture/operation-capture.meta-reducer';
+import { _resetDevErrorState } from './app/util/dev-error';

@@ beforeEach(() => {
   clearDeferredActions();
+
+  // The dialog spies below are created ONCE at module load, so their call
+  // history would otherwise accumulate across every spec in the Karma run,
+  // letting `expect(window.alert).toHaveBeenCalled()` pass on a stale call
+  // from an unrelated spec. devError's `isShowAlert` latches off after the
+  // run's first alert, which hides the miss in the other direction.
+  if (jasmine.isSpy(window.alert)) (window.alert as jasmine.Spy).calls.reset();
+  if (jasmine.isSpy(window.confirm)) (window.confirm as jasmine.Spy).calls.reset();
+  _resetDevErrorState();
 });
```

Both lines are needed. `calls.reset()` alone makes assertions honest but would convert some currently-vacuous tests into *order-dependent red* ones, because the `isShowAlert` latch means a legitimate `devError` can't raise its alert if an earlier spec already latched it. `_resetDevErrorState()` closes that half.

I verified this patch locally:

- **Non-breaking:** all 12 dialog-asserting specs run together → `398 SUCCESS`.
- **Effective:** with this patch and **no spec-level change at all**, the #9079 sabotage above flips from `75/75 SUCCESS` to `1 FAILED, 74 SUCCESS`.

The second point is the useful one: fixing `test.ts` closes the whole class of bug repo-wide, so individual specs don't each need to remember `_resetDevErrorState()` + `calls.reset()`.

## Possible follow-up (not verified)

The same global spy object also carries its `.and` **strategy** across specs — e.g. a spec doing `(window.confirm as jasmine.Spy).and.returnValue(false)` mutates the shared spy for every later spec, overriding the `returnValue(true)` default set in `test.ts`. Nothing appears to be failing because of this today, so I've left it out of the fix above; resetting the strategy as well would need a check that no spec sets its stub in a `beforeAll` (which the global `beforeEach` would then wipe).

## Acceptance criteria

- [ ] `src/test.ts` re-arms the dialog spy call logs and `devError`'s `isShowAlert` per spec.
- [ ] Full suite stays green in both TZ variants.
- [ ] Deleting a `devError(...)` call guarded by a spec assertion makes that spec go red in a **full-suite** run, not just standalone.

giri256:
## Problem

The global `window.alert` and `window.confirm` spies keep their call history across specs. `devError` also keeps its one-alert latch across specs, so dialog assertions can pass because of an unrelated earlier call or become order-dependent.

Closes #9100.

## Solution

Reset both dialog spies and re-arm `devError` in the existing global `beforeEach`. Each spec now starts with clean dialog state while retaining the shared spy setup.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] Documentation changes are not required for this test-infrastructure fix
- [x] Existing relevant tests still pass

## Verification

- `npm run checkFile src/test.ts`
- Combined `task.reducer.spec.ts` and `task.reducer.util.spec.ts`: 75 tests passed
- Pre-push lint, lint-rule, theme, sync-core (243), sync-provider (404), and release-note (24) suites passed
- The full Angular bundle completed locally, but ChromeHeadless disconnected before executing the full browser suite; CI should provide the authoritative full-suite result

### DIFF
--- a/src/test.ts
+++ b/src/test.ts
@@ -25,6 +25,7 @@ import 'fake-indexeddb/auto';
 import { IDBFactory } from 'fake-indexeddb';
 import { asyncScheduler } from 'rxjs';
 import { clearDeferredActions } from './app/op-log/capture/operation-capture.meta-reducer';
+import { _resetDevErrorState } from './app/util/dev-error';
 
 import { getTestBed, TestBed } from '@angular/core/testing';
 import {
@@ -108,6 +109,18 @@ beforeEach(() => {
   // the phantom-change guard (#8751) reads it and silently skips compaction
   // and snapshot saves — an order-dependent failure that passes standalone.
   clearDeferredActions();
+
+  // The dialog spies below are created once at module load, so reset their
+  // call history before every spec to prevent assertions from passing on a
+  // stale call from an unrelated test. Re-arm devError's alert latch as well
+  // so legitimate alerts remain observable regardless of spec order.
+  if (jasmine.isSpy(window.alert)) {
+    (window.alert as jasmine.Spy).calls.reset();
+  }
+  if (jasmine.isSpy(window.confirm)) {
+    (window.confirm as jasmine.Spy).calls.reset();
+  }
+  _resetDevErrorState();
 });
 
 // Mock browser dialogs globally for tests


johannesjo (approved):
Thanks for picking this up — and for implementing the fix rather than papering over it in the individual specs. I verified this locally and it does what it says. **Approving**, with two non-blocking nits below.

## Verification

The important thing to check here is that the fix actually *catches* something, since a test-infra change is invisible otherwise. So I mutation-tested it: deleted the `devError(...)` call in `deleteTaskHelper` (`src/app/features/tasks/store/task.reducer.util.ts`) **and** stripped the per-spec workaround from `task.reducer.util.spec.ts`, so the spec relies purely on the global `beforeEach`. Then ran the full suite — the configuration CI uses:

| full suite, production `devError(...)` deleted | result |
| --- | --- |
| **without** this PR | `13306 SUCCESS` — the assertion is vacuous, mutation survives |
| **with** this PR | `1 FAILED, 13305 SUCCESS` — mutation caught |

One methodology note that may be useful to anyone re-checking this: the minimal two-file repro from #9100 (`task.reducer.spec.ts` + `task.reducer.util.spec.ts`) is **order-dependent** — without the fix I got 2 green and 4 red across 6 identical runs. `src/karma.conf.js` sets no `random` key, so jasmine's default `random: true` applies and spec order reshuffles every run; whether the mutation survives depends on whether `task.reducer.spec.ts`'s devError spec happens to run first. The full suite is the reliable measurement because thousands of prior specs make a stale `window.alert` call near-certain. With the fix applied, the two-file repro becomes deterministic (4/4 caught).

Also checked:

- **Full suite green in both TZ variants** with the patch: Berlin `13306 SUCCESS`, Los Angeles `13292 SUCCESS`. Unpatched Berlin baseline is also `13306 SUCCESS`, so nothing regressed.
- `npm run checkFile src/test.ts` passes.
- The branch base is behind master, but zero commits have touched `src/test.ts` or `src/app/util/dev-error.ts` since the merge-base, so there is nothing to rebase around.
- Heads up for whoever merges: **CI has not run on this PR** — the only check on `0b916c09fc` is `welcome`. It needs workflow approval for a first-time contributor.

One flake to not misattribute if CI hits it: my first patched run showed `1 FAILED` on `AddTaskBarParserService … should clear a previously parsed deadline` (2003ms async timeout). That file has no dialog usage, it passes standalone (77 tests in 0.42s), and it did not recur in five later full runs.

## Nits (non-blocking)

**1. Stale comment left behind.** `src/app/features/tasks/store/task.reducer.util.spec.ts:218-220` still says:

> `isShowAlert` latches off after the first devError of the whole Karma run, and test.ts creates the window.alert spy once and never resets its call history — so we must reset both before asserting devError fired.

The second half becomes false with this PR. Those local resets are now redundant (I confirmed the spec stays green without them). Worth updating or dropping the block so the next reader isn't told the opposite of what `test.ts` does.

**2. The fix isn't guarded by anything.** This is the one I'd genuinely like to see addressed, though I don't think it should block the merge.

I checked whether simply removing the redundant local resets would turn that spec into a regression test for the global fix. It doesn't: with production code intact, `expect(window.alert).toHaveBeenCalled()` is *vacuously true*, so it stays green whether or not `test.ts` resets anything. Which means if someone later refactors this `beforeEach`, nothing goes red and the whole class of bug quietly comes back.

A small self-contained spec does guard it. This one is red without the fix and green with it (3/3 each way), full suite `13308 SUCCESS`, `checkFile` clean:

```ts
import { devError } from './app/util/dev-error';

// Guards the global dialog-state reset in `src/test.ts` (#9100). Without it the
// shared window.alert spy accumulates calls for the whole Karma run and
// devError's `isShowAlert` latches off after the run's first alert, which makes
// `expect(window.alert).toHaveBeenCalled()` vacuous in any spec.
//
// Both specs are identical on purpose: whichever Jasmine happens to run second
// (spec order is randomized) sees the other's leftovers unless test.ts resets
// them, so this is order-independent.
describe('test.ts global dialog state', () => {
  beforeEach(() => {
    // devError always asks confirm() whether to rethrow; test.ts defaults it to
    // true, which would throw here.
    (window.confirm as jasmine.Spy).and.returnValue(false);
  });

  afterEach(() => {
    (window.confirm as jasmine.Spy).and.returnValue(true);
  });

  const expectCleanDialogState = (): void => {
    expect(window.alert).not.toHaveBeenCalled();

    devError('dev error');

    expect(window.alert).toHaveBeenCalledTimes(1);
  };

  it('starts with a cleared alert log and a re-armed devError latch', () => {
    expectCleanDialogState();
  });

  it('starts clean again for the following spec', () => {
    expectCleanDialogState();
  });
});
```

The two specs being identical is the load-bearing part: under randomized order, a "first one dirties, second one asserts" pair silently passes when the shuffle reverses it.

## Not for this PR

The known follow-up from #9100 is still open — the shared spies carry their `.and` **strategy** across specs, so `task.reducer.util.spec.ts` setting `confirm → false` leaks to every later spec in the run. Correctly left out of scope here. Worth remembering that `devError` calls `confirmDialog` *unconditionally* (only the alert is latched), so the `returnValue(true)` default makes `devError` throw in specs — which is why those stubs exist in the first place.

johannesjo:

PR #9186, opened 2026-07-20
fix(sync): accept legacy schema versions on download

### LINKED ISSUE #8923: Sync: download response parsing fails closed on legacy non-integer schemaVersion
Follow-up from the PR #8900 review — narrow fail-closed edge on the download path.

`SuperSyncOperationSchema.schemaVersion` is now `z.number().int().min(1).max(100)`, and the client's download-response validation uses it. Master-era servers range-checked 1..100 but did **not** enforce integers, so a legacy stored op with e.g. `schemaVersion: 1.5` is possible. Such an op now hard-fails the *entire* download page with a generic `InvalidDataSPError` — a permanent sync wedge with no useful UX — instead of the graceful per-op block (`INVALID_SCHEMA_VERSION` → cursor held + once-per-session snack) that `remote-ops-processing.service.ts` already provides.

**Proposal:** keep the strict schema for uploads (server-side `Number.isInteger` check stays), but parse the download *response* envelope leniently (`z.number()`) and let the per-op version gate classify. Unknown whether fractional values exist on any deployed server — the fix is cheap insurance either way.

**Refs:** `packages/shared-schema/src/supersync-http-contract.ts`, `src/app/op-log/sync-providers/super-sync/response-validators.ts`, `remote-ops-processing.service.ts` version gate.

giri256:
## Problem

Download responses are parsed with the strict operation schema. A legacy server operation containing a fractional numeric `schemaVersion` therefore rejects the entire response page before the existing per-operation version gate can classify and block it safely.

Closes #8923.

## Solution

Use a download-only operation schema that accepts numeric legacy schema versions while preserving every other operation constraint. The canonical operation schema remains strict, so upload/server validation still rejects non-integer versions.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] Documentation changes are not required for this compatibility fix

## Verification

- Added a regression test that failed before the implementation at `ops[0].op.schemaVersion`
- Shared-schema suite: 86 tests passed
- Shared-schema build passed
- App-level SuperSync response-validator suite: 38 tests passed

### DIFF
--- a/packages/shared-schema/src/supersync-http-contract.ts
+++ b/packages/shared-schema/src/supersync-http-contract.ts
@@ -149,6 +149,18 @@ export const SuperSyncServerOperationSchema = z
   })
   .passthrough();
 
+// Legacy servers accepted finite numeric schema versions before integer
+// validation was added. Keep download parsing transport-compatible so the
+// client can classify unsupported versions per operation without rejecting
+// and permanently wedging the entire response page.
+const SuperSyncDownloadOperationSchema = SuperSyncOperationResponseSchema.extend({
+  schemaVersion: z.number(),
+});
+
+const SuperSyncDownloadServerOperationSchema = SuperSyncServerOperationSchema.extend({
+  op: SuperSyncDownloadOperationSchema,
+});
+
 export const SuperSyncUploadResultSchema = z
   .object({
     opId: z.string(),
@@ -171,7 +183,7 @@ export const SuperSyncUploadOpsResponseSchema = z
 
 export const SuperSyncDownloadOpsResponseSchema = z
   .object({
-    ops: z.array(SuperSyncServerOperationSchema),
+    ops: z.array(SuperSyncDownloadServerOperationSchema),
     hasMore: z.boolean(),
     latestSeq: z.number(),
     gapDetected: z.boolean().optional(),
--- a/packages/shared-schema/tests/supersync-http-contract.spec.ts
+++ b/packages/shared-schema/tests/supersync-http-contract.spec.ts
@@ -231,6 +231,27 @@ describe('SuperSync HTTP contract schemas', () => {
     expect(parsed.capabilities?.causalRepairSnapshots).toBe(true);
   });
 
+  it('accepts legacy fractional schema versions in download responses', () => {
+    const operation = {
+      ...createValidOperation(),
+      schemaVersion: 1.5,
+    };
+    const parsed = SuperSyncDownloadOpsResponseSchema.parse({
+      ops: [
+        {
+          serverSeq: 1,
+          op: operation,
+          receivedAt: 1234567890,
+        },
+      ],
+      hasMore: false,
+      latestSeq: 1,
+    });
+
+    expect(parsed.ops[0].op.schemaVersion).toBe(1.5);
+    expect(() => SuperSyncOperationSchema.parse(operation)).toThrow();
+  });
+
   it('requires an operation ID for destructive clean-slate snapshots', () => {
     expect(() =>
       SuperSyncUploadSnapshotRequestSchema.parse({


johannesjo (commented):
Thanks for picking this up — the change is safe as written (I verified that below), but before it merges there are two things worth resolving.

## 1. The scenario this defends against can't occur on a SuperSync server

#8923 said *"Unknown whether fractional values exist on any deployed server."* That's answerable, and the answer looks like **no**:

- `schema_version INTEGER NOT NULL` — `packages/super-sync-server/prisma/migrations/0_init/migration.sql:51`. The Prisma model field has been `Int` since `schema.prisma`'s first commit (`git log -p`); never `Float`/`Decimal`.
- Every op insert goes through typed Prisma `tx.operation.createMany` (`operation-upload.service.ts:670` batch path, `:936` single-op path). There is no raw-SQL insert into `operations` — the only `INSERT INTO` in the server targets `user_sync_state`.
- A parameterized insert of a fractional value is rejected by Postgres outright. Verified against real Postgres (PGlite):

```
CREATE TABLE t (schema_version INTEGER NOT NULL);
INSERT INTO t VALUES ($1)   -- params [1.5] -> ERROR: invalid input syntax for type integer: "1.5"
INSERT INTO t VALUES (1.5)  -- literal      -> stored as 2 (assignment cast rounds)
```

So even on a master-era server without the `Number.isInteger` check, a fractional `schemaVersion` either failed the insert or was rounded to an integer on the way in. A download response can't carry one back out.

That puts this in the territory the sync contributor rules specifically warn about — hardening with no reachable failure. @johannesjo, your call: I'd lean toward closing #8923 as not-reachable. If you'd rather keep it as defense-in-depth against a proxying/non-SP server, then it should at least be done symmetrically — see below.

## 2. If kept: `newOps` on the upload response is left strict

Ops reach the client through two response envelopes and only one is relaxed here. `SuperSyncUploadOpsResponseSchema.newOps` (`supersync-http-contract.ts:166`) still uses the unmodified strict `SuperSyncServerOperationSchema`, and piggybacked ops feed the exact same per-op gate (`operation-log-upload.service.ts:486-529` → `processRemoteOps`). Confirmed by parsing directly with this branch applied:

```
UPLOAD newOps schemaVersion=1.5 -> success = false
issues: [{"expected":"int","path":["newOps",0,"op","schemaVersion"], …}]
```

In fairness to the PR: I expected this to mean "the wedge simply moves to the upload", and it does **not**. `sync-wrapper.service.ts:575` returns early on `blocked_incompatible`, so the primary flow never reaches the upload after a blocked download. The gap is only exposed on upload paths that run without a preceding download in the same cycle (`encryption-password-change.service.ts:146`, `sync-import-conflict-coordinator.service.ts:87`), and both of those are clean-slate uploads where the server has nothing to piggyback. So it's an inconsistency, not a live bug — but the stated rationale ("classify per-op instead of rejecting the whole page") applies to both envelopes equally, and depending on orchestration order for that is an accident rather than a guarantee.

The complete version is also the simpler one — a single edit instead of two new aliases:

```ts
export const SuperSyncOperationResponseSchema = SuperSyncOperationSchema.extend({
  schemaVersion: z.number(),
}).passthrough();
```

That covers `ops` and `newOps` together and lets both `SuperSyncDownloadOperationSchema` and `SuperSyncDownloadServerOperationSchema` go away. Nothing strict is lost: `SuperSyncOperationResponseSchema` / `SuperSyncServerOperationSchema` are response-only (no consumer anywhere outside the contract file and the `index.ts` re-export — the server never imports either), and uploads keep `SuperSyncOperationSchema` plus the server's `Number.isInteger` check in `validation.service.ts:207`.

## What I verified as safe

Widening to bare `z.number()` also drops `min(1)`/`max(100)`, so I checked what now gets through and what the gate does with it (zod v4):

| value | download schema | `getOperationSchemaVersion` (`schema-migration.service.ts:24`) |
| --- | --- | --- |
| `1.5`, `-5` | accepted | throws → `INVALID_SCHEMA_VERSION` → blocked |
| `0` | accepted | `VERSION_UNSUPPORTED` → blocked |
| `1e9` | accepted | `VERSION_TOO_NEW` → blocked |
| `Infinity`, `NaN` | rejected by zod | — |

Every finite value degrades to a graceful per-op block; nothing reaches migration or storage, so the widening opens no hole. Shared-schema suite green (86/86) with the branch applied, `checkFile` passes on the changed file, and there's no type-surface change (`.int()` doesn't affect the inferred TS type).

## Minor

Worth noting in #8923 that this is a better error message, not a recovery: `INVALID_SCHEMA_VERSION` holds the cursor and shows the snack once per session (`remote-ops-processing.service.ts:515`), so sync stays wedged either way. It also reuses `T.F.SYNC.S.MIGRATION_FAILED` ("migration failed"), which reads a bit misleading for a malformed-version op rather than a failed migration.

giri256:
Thanks for the detailed review. I independently rechecked the full persistence path and agree this is not reachable on a SuperSync server: `schema_version` has been `INTEGER NOT NULL` / Prisma `Int` since the schema was introduced, and both operation insertion paths use typed `tx.operation.createMany`. Current validation also rejects non-integer values before persistence.\n\nGiven the repository's sync rule requiring a reproducible failure before adding hardening, broadening the response schemas would add complexity without a live failure mode. I'm closing this PR rather than extending it symmetrically. The suggested response-schema simplification would be appropriate only if a concrete non-SuperSync/proxy compatibility requirement is established later.

PR #9189, opened 2026-07-20
fix(plugins): filter Linear auto-import by current cycle

mdeix:
## Problem

Linear auto-import currently imports all assigned open issues. Users who work primarily from Linear cycles need a way to limit auto-imported backlog items to issues from the current active cycle.

## Solution

Add an advanced Linear plugin checkbox: “Only auto-import issues from current cycle”.

When enabled, the Linear GraphQL search query includes:

`cycle: { isActive: { eq: true } }`

When disabled or unset, no cycle filter is sent, so existing behavior remains unchanged. The option is disabled by default via absent/falsy plugin config.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

For the unchecked items, keep them unchecked unless you also ran broader tests or decide docs/wiki updates are needed.

### DIFF
--- a/packages/plugin-dev/linear-issue-provider/i18n/en.json
+++ b/packages/plugin-dev/linear-issue-provider/i18n/en.json
@@ -3,6 +3,7 @@
     "API_KEY": "API Key",
     "TEAM_ID": "Team ID (optional, filter to a specific team)",
     "PROJECT_ID": "Project ID (optional, filter to a specific project)",
+    "AUTO_IMPORT_CURRENT_CYCLE_ONLY": "Only auto-import issues from current cycle",
     "HOW_TO_GET_TOKEN": "Get your API key"
   },
   "DISPLAY": {
--- a/packages/plugin-dev/linear-issue-provider/src/plugin.ts
+++ b/packages/plugin-dev/linear-issue-provider/src/plugin.ts
@@ -20,6 +20,7 @@ interface LinearConfig {
   apiKey?: string;
   teamId?: string;
   projectId?: string;
+  isAutoImportCurrentCycleOnly?: boolean;
 }
 
 interface LinearGraphQLResponse<T = unknown> {
@@ -66,14 +67,15 @@ const t = (key: string): string => {
 };
 
 const SEARCH_ISSUES_QUERY = `
-  query SearchIssues($first: Int!, $team: TeamFilter, $project: NullableProjectFilter) {
+  query SearchIssues($first: Int!, $team: TeamFilter, $project: NullableProjectFilter, $cycle: NullableCycleFilter) {
     viewer {
       assignedIssues(
         first: $first,
         filter: {
           state: { type: { in: ["backlog", "unstarted", "started"] } },
           team: $team,
-          project: $project
+          project: $project,
+          cycle: $cycle
         }
       ) {
         nodes {
@@ -149,6 +151,9 @@ const searchAssignedIssues = async (
   if (cfg.projectId) {
     variables.project = { id: { eq: cfg.projectId } };
   }
+  if (cfg.isAutoImportCurrentCycleOnly) {
+    variables.cycle = { isActive: { eq: true } };
+  }
 
   const data = await graphql<{
     viewer: { assignedIssues: { nodes: LinearRawIssueReduced[] } };
@@ -190,6 +195,12 @@ PluginAPI.registerIssueProvider({
       key: 'projectId',
       type: 'input',
       label: t('CFG.PROJECT_ID'),
+      advanced: true,
+    },
+    {
+      key: 'isAutoImportCurrentCycleOnly',
+      type: 'checkbox' as const,
+      label: t('CFG.AUTO_IMPORT_CURRENT_CYCLE_ONLY'),
       advanced: true,
     },
   ],


johannesjo:
Thanks for the PR, and welcome! 👋 The underlying need is legitimate and the GraphQL is correct — I verified the query against Linear's actual schema. But there's one blocking issue: the filter is applied in the wrong place, so it silently breaks manual issue search.

## 🔴 Blocking: the "auto-import" filter also filters manual search

`searchAssignedIssues` has **two** callers:

- `searchIssues` (line 222) — manual search in the add-task bar / issue panel
- `getNewIssuesForBacklog` (line 291) — auto-import

The cycle filter is added inside the shared helper, so it applies to **both** — even though the config key (`isAutoImportCurrentCycleOnly`) and the label ("Only **auto-import** issues from current cycle") promise auto-import only.

What a user hits:

1. Enables the checkbox → auto-import correctly narrows to the active cycle ✅
2. Later wants to attach a specific backlog issue (not in a cycle) to a task, types its ID into the add-task bar
3. **No result. No error. No explanation.**

Manual search is exactly the escape hatch for the issues auto-import now excludes, so filtering it defeats the purpose of narrowing auto-import in the first place.

This also runs against an explicit decision on #7674 — the same request ("filter auto-import by iteration/sprint") for Azure DevOps:

> **Keep it to auto-import only for now — search uses a separate query in Jira too, and that's out of scope here.**

Every existing provider follows that split, and the distinction is consistent: **generic-scope settings apply to both paths; auto-import-scoped settings apply only to auto-import.**

| Provider | Auto-import-only setting | Applied in |
|---|---|---|
| Jira | `autoAddBacklogJqlQuery` | `findAutoImportIssues$` only (`jira-api.service.ts:233`) |
| Azure DevOps | `autoImportWiql` | `backlogQuery()` → `getNewIssuesForBacklog` only (`plugin.ts:194`, `:333`) |
| GitHub | `backlogQuery` | `getNewIssuesForBacklog` only (`plugin.ts:295`) — note `includePullRequests`, which *is* generic, applies to both |

The Linear plugin's own `teamId` / `projectId` are on the generic side of that line, and their labels say so ("filter to a specific team") — which is why they correctly apply to both.

### Suggested fix

Thread the flag through so only auto-import sets it:

```ts
const searchAssignedIssues = async (
  cfg: LinearConfig,
  http: PluginHttp,
  opts: { searchTerm?: string; isCurrentCycleOnly?: boolean } = {},
): Promise<PluginSearchResult[]> => {
  const variables: Record<string, unknown> = { first: 50 };
  // ... teamId / projectId unchanged ...
  if (opts.isCurrentCycleOnly) {
    variables.cycle = { isActive: { eq: true } };
  }
  // ... use opts.searchTerm for the client-side term filter ...
};
```

with:

```ts
searchIssues(searchTerm, config, http) {
  return searchAssignedIssues(config as unknown as LinearConfig, http, { searchTerm });
},

getNewIssuesForBacklog(config, http) {
  const cfg = config as unknown as LinearConfig;
  return searchAssignedIssues(cfg, http, {
    isCurrentCycleOnly: !!cfg.isAutoImportCurrentCycleOnly,
  });
},
```

(A 4th positional `isCurrentCycleOnly = false` param works too if you prefer a smaller diff — the options object just keeps the signature at three.)

<sub>_Alternative, if you'd rather not thread anything:_ relabel it as a general filter ("Only show issues from the current cycle") and drop `AutoImport` from the key, matching how `teamId`/`projectId` behave. It's a smaller change, but I'd recommend against it — it removes the manual escape hatch, and #7674 already settled this direction.</sub>

## ✅ What checks out

I verified the query against Linear's real schema (extracted the type definitions from `@linear/sdk@88.2.0` rather than relying on docs):

- `viewer.assignedIssues(filter:)` takes `IssueFilter` ✅
- `IssueFilter.cycle: NullableCycleFilter` ✅ — the declared variable type is right
- `NullableCycleFilter.isActive: BooleanComparator`, and `BooleanComparator.eq: Boolean` ✅

**Backward compatibility is genuinely safe.** Per the GraphQL spec, an input-object field whose value is an unprovided variable is omitted from the coerced filter rather than passed as `null` — so `cycle: $cycle` with no `$cycle` is a true no-op. And this isn't just theory here: `$team` and `$project` in the same query already depend on exactly that behaviour, and they've been live since v18.10.0.

Also fine: `checkbox` and `advanced` are valid `PluginFormField` members, only `en.json` was touched (correct — other locales are generated), and there's no committed plugin bundle needing a rebuild.

## Minor

1. **A test would be worth it** — especially once the fix above creates a behavioural distinction (auto-import filtered, search *not* filtered) that's easy to regress silently. `azure-devops-issue-provider/src/plugin.spec.ts` is the pattern to copy: it captures the exact query auto-import sends. Fair warning on cost: the Linear plugin has no `test` script yet and isn't in the `PLUGINS` array in `.github/workflows/plugin-tests.yml`, so this means a vitest devDep (per-plugin devDeps are explicitly allowed), a `test` script, and one workflow line. Recommended, not blocking.
2. **Add a `description` to the field.** Issues in *no* cycle are excluded too — a real consequence worth spelling out, the way Azure's `AUTO_IMPORT_WIQL_DESC` does.
3. **`type: 'checkbox' as const`** — the `as const` isn't needed (contextual typing from `IssueProviderPluginDefinition` covers it), and none of the sibling fields use it.
4. **Commit type** — this is `feat(plugins)` rather than `fix(plugins)`; nothing is currently broken. The PR description also has a stray trailing line ("For the unchecked items, keep them unchecked unless…") that looks accidental.
5. The branch is a little behind `master` — no conflicts, nothing else touches this plugin.

## One question

Is there an issue behind this? The concept is well-precedented (#7674 is the same ask for Azure DevOps and was accepted), and since the Linear plugin already has two advanced auto-import narrowing fields this extends an existing building block rather than adding a new surface — so I think it earns its place. But it'd help to have a linked issue showing demand beyond a single user, which is the usual bar for a new setting here.

Nice, tight change overall — fix the auto-import scoping and this is close to mergeable.

PR #9208, opened 2026-07-20
fix(markdown): select exactly the url when wrapping selected text in a link

### LINKED ISSUE #9202: 🚨 Bug: Project note editing, adding URL to text auto highlight offset incorrect
### Version Used

all versions

### Current Behavior

When adding a URL to text in a Project Note via the UI button or CTRL+K, the markdown is correctly inserted but the highlighted text to be replaced is offset to the right. As a result, an 'h' is left at the front, and the URL overwrites the last closing bracket. 

Video example:
https://github.com/user-attachments/assets/aaf6c22e-6e43-4556-8e06-a77bd62c4a9e

### Expected Behavior

For the auto highlighted portion of (https://) to cover 'https://' instead of 'ttps://)'

### Steps to Reproduce

1. Add text to a project note
2. select the text and press the URL button or CTRL+K
3. paste your desired URL

### Console Output

```shell

```

### Documentation

SAY-5:
## Problem

Fixes #9202. When you select text in a Project Note and insert a link (URL button or Ctrl+K), the wrong range gets highlighted: `[hello](https://) world` selects `ttps://)` instead of `https://`, so typing the URL leaves a stray `h` in front and overwrites the closing `)`.

The url offset in `insertLink` was one too high (`+ 4` instead of `+ 3`). `insertImage` had the same off-by-one (`+ 5` instead of `+ 4`). The two existing tests encoded the wrong positions, so they passed while the behaviour was broken.

## Solution

Correct the url start offset in both `insertLink` and `insertImage`, update the two tests to the right positions, and add a small assertion in each that the resulting selection is exactly `https://`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/ui/inline-markdown/markdown-toolbar.util.spec.ts
+++ b/src/app/ui/inline-markdown/markdown-toolbar.util.spec.ts
@@ -316,8 +316,15 @@ describe('markdown-toolbar.util', () => {
     it('should use selection as link text', () => {
       const result = insertLink('hello world', 0, 5);
       expect(result.text).toBe('[hello](https://) world');
-      expect(result.selectionStart).toBe(9); // URL position
-      expect(result.selectionEnd).toBe(17); // URL selected
+      expect(result.selectionStart).toBe(8); // URL position
+      expect(result.selectionEnd).toBe(16); // URL selected
+    });
+
+    it('should select exactly the url placeholder', () => {
+      const result = insertLink('hello world', 0, 5);
+      expect(result.text.substring(result.selectionStart, result.selectionEnd)).toBe(
+        'https://',
+      );
     });
   });
 
@@ -332,8 +339,15 @@ describe('markdown-toolbar.util', () => {
     it('should use selection as alt text', () => {
       const result = insertImage('hello world', 0, 5);
       expect(result.text).toBe('![hello](https://) world');
-      expect(result.selectionStart).toBe(10); // URL position
-      expect(result.selectionEnd).toBe(18); // URL selected
+      expect(result.selectionStart).toBe(9); // URL position
+      expect(result.selectionEnd).toBe(17); // URL selected
+    });
+
+    it('should select exactly the url placeholder', () => {
+      const result = insertImage('hello world', 0, 5);
+      expect(result.text.substring(result.selectionStart, result.selectionEnd)).toBe(
+        'https://',
+      );
     });
   });
 
--- a/src/app/ui/inline-markdown/markdown-toolbar.util.ts
+++ b/src/app/ui/inline-markdown/markdown-toolbar.util.ts
@@ -488,10 +488,9 @@ export const insertLink = (
   const linkMarkdown = `[${selectedText}](${url})`;
   const newText =
     text.substring(0, selectionStart) + linkMarkdown + text.substring(selectionEnd);
-  // Place cursor at URL position: [ + text + ]( = 1 + text.length + 2 = text.length + 3
-  // But we want cursor AFTER the opening paren, which is at position: 1 + text.length + 2 = text.length + 3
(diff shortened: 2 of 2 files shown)


johannesjo (commented):
Reviewed at `a5c3805`. This is correct, minimal and well-tested — thanks, and congrats on the first PR here! 🎉

Rather than take the arithmetic on faith I ran the real source directly (bundled `markdown-toolbar.util.ts` with esbuild, no hand-transcription) against both the PR and `master`. Notes below for the record.

### The bug reproduces exactly as reported in #9202

```
master:  insertLink("hello world", 0, 5) → "[hello](https://) world"  sel="ttps://)"  (9,17)
PR:      insertLink("hello world", 0, 5) → "[hello](https://) world"  sel="https://"  (8,16)
```

Not new, and not master-only: it was introduced in e2bf969397 (*"feat: add markdown formatting toolbar to notes editor"*). `git merge-base --is-ancestor` puts it in **every release from v16.9.0 through v18.15.1**, so this has been broken for everyone since the toolbar shipped.

### The fix holds beyond the one test case

I probed both functions across 9 input shapes — selection at start/middle/end of the text, after a newline, an astral-plane emoji, `]`/`(` inside the selection, whitespace-only selection, a multi-line selection — plus a non-default `url` argument. **20/20 cases fail on `master`** (always `"ttps://)"`), **20/20 pass on the PR** (always exactly `https://`). The offset is relative to `selectionStart`, so it's correct at any position, and the empty-selection branches (`[text](…)` / `alt`) were already right and are untouched.

### The tests genuinely fail without the fix

Restoring `master`'s util under this PR's spec:

```
TOTAL: 4 FAILED, 124 SUCCESS
  Expected 'ttps://)' to be 'https://'.   ← insertLink
  Expected 9 to be 8. / Expected 17 to be 16.
  Expected 'ttps://)' to be 'https://'.   ← insertImage
  Expected 10 to be 9. / Expected 18 to be 17.
```

With the fix: `128 SUCCESS`. `dialog-fullscreen-markdown.component.spec.ts` also stays green (19/19) — no other spec encoded the old positions. `npm run checkFile` clean on both files.

### Scope is exactly right

I checked every other selection-returning transform in the file for the same off-by-one class — `applyBold`, `applyItalic`, `applyStrikethrough`, `applyInlineCode`, `applyCodeBlock` all correctly re-select the original text, and `insertTable` returns a deliberate collapsed caret. These two were the only ones affected, and the only two carrying the tell-tale `// Test expects: for "hello" (5 chars), URL starts at 9` comment (2 occurrences repo-wide, both removed here).

That comment is the actual root cause worth calling out: the constant had been tuned to a wrong assertion instead of to the markdown string. Replacing it with a comment that derives the offset from the syntax itself (`1 ("[") + text.length + 2 ("](")`) is the right correction — it removes the thing that made the bug self-justifying.

And the fix does reach the UI: `_applyTransformWithArgs` calls `textarea.setSelectionRange(result.selectionStart, result.selectionEnd)`, and both entry points from the issue land there — the toolbar button (`onInsertLink`) and <kbd>Ctrl</kbd>+<kbd>K</kbd> (`markdown-shortcuts.const.ts`). `DialogAddNoteComponent extends DialogFullscreenMarkdownComponent`, so that is the Project Note editor in the report; `inline-markdown.component.ts` imports only `handleListKeydown`, so there is no second path to fix.

No conflicts with other open PRs touching `src/app/ui/inline-markdown/` (#6782 only touches the component).

### One optional nit (non-blocking)

The new `should select exactly the url placeholder` tests and the corrected `toBe(8)` / `toBe(16)` assertions catch the same regression. The substring assertion is the durable one; the raw numbers are the brittle kind that caused this in the first place. Keeping both is fine — just noting that if either should survive a future refactor, it's the substring one.

PR #9211, opened 2026-07-20
test(sync): cover production batch conflict queries

### LINKED ISSUE #9205: SuperSync: batch conflict detection can be disabled entirely and the test suite stays green
Found by mutation-testing the conflict-detection surface while reviewing #9195: **67 mutations applied one at a time, 33 killed, 34 survived**. The important survivors clustered in the two batch paths.

## Confirmed finding

Appending `AND FALSE` to the raw query in either `detectConflictForEntities` or `prefetchLatestEntityOpsForBatch`—so the query can never match a row—left the default server suite at **915/915 green**.

That is the core finding: the default suite did not execute the production batch SQL templates. These were **coverage holes, not known live bugs**; the batch queries were correct at the time of filing.

## CI nuance

`tests/integration/conflict-detection-sql.integration.spec.ts` does execute the real functions against real PostgreSQL. It is excluded from the package's default Vitest run, but the `SuperSync Server Tests` PR workflow runs it. That workflow is not currently a required check in the active repository ruleset, so it can still be bypassed. Making it required remains the separate follow-up in #9194.

## Behavioral risks worth guarding

1. Disabling either batch query can miss conflicts entirely.
2. Omitting selected fields such as `actionType`, `clientId`, `vectorClock`, or `serverSeq` changes conflict classification or legacy arbitration.
3. Flipping `server_seq DESC` to `ASC` compares against stale history and can accept an operation that conflicts with the current row.
4. Relaxing either side of the two-delta exemption applies it to a delta/non-delta pair.
5. Dropping `existingOp.userId === userId` weakens tenant-scoped duplicate identity.
6. Exact-size and overflow chunk mistakes can skip the 100th/101st entity or issue an empty query.
7. Returning after a clean entity can hide a conflict on a later entity.
8. Widening the legacy `GLOBAL_CONFIG:misc` boundary from `< 2` to `<= 2` aliases a valid v2 row to `tasks`.
9. Removing the first entity-ID dedupe changes the common `entityId: 'a', entityIds: ['a']` shape from the single-entity path to the batch path.

The time-delta survivors were previously described as proven "silent data loss." That was too strong. Missing `actionType` definitely causes false conflicts/recovery churn, and a one-sided exemption definitely misclassifies concurrency; downstream data loss was not reproduced and is not claimed here.

## Survivors that do not need more behavioral tests

- The second `Set` in `detectConflict` dedupes an already-deduplicated result.
- Removing SQL prefilters or post-`unnest` restrictions can return extra rows that current consumers ignore. Those predicates are performance/index safeguards and belong in `EXPLAIN` coverage.
- Changing the legacy arbitration tiebreak from `>` to `>=` is equivalent because `(user_id, server_seq)` is unique.
- Removing `serverSeq` from a typed Prisma select is caught by TypeScript compilation.
- Removing the empty-input fast return is equivalent to the zero-iteration loop.

## Implemented acceptance criteria

- [x] Replace both copied SQL helpers in `entity-ids-conflict.pglite.spec.ts` with a transaction adapter that renders the **production** tagged templates through `Prisma.sql(...).text` and `.values`, including nested `Prisma.sql` / `Prisma.join` fragments.
- [x] Exercise both production batch functions in PGlite and cover disabled queries, latest-row ordering, tenant/type isolation, scalar/array entity membership, and required projections.
- [x] Add focused guards for both delta asymmetries, cross-user duplicate rejection, clean-first/later-conflict traversal, first-stage dedupe, and the v2 legacy boundary.
- [x] Cover exact 100-item and 101-item batches for both chunk consumers, including emitted chunk sizes.
- [x] Mutation-check the targeted guards, run mandatory per-file checks, pass the TypeScript build, and pass the full server suite (**922/922** in the implementation worktree).
- [ ] Make `SuperSync Server Tests` a required repository check (#9194).

johannesjo:
## Summary

- execute the production `detectConflictForEntities` and `prefetchLatestEntityOpsForBatch` tagged SQL against PGlite in the default server suite
- cover required projections, latest-row selection, tenant/type isolation, scalar/array membership, conflict traversal, timer-delta asymmetry, duplicate ownership, legacy config boundaries, and exact 100/101-item chunking
- replace the copied SQL test helpers with a small Prisma tagged-template transaction adapter

## Why

The default SuperSync server suite stayed green when either production batch conflict query was disabled with `AND FALSE`. The tests exercised hand-copied SQL rather than the production templates, so critical conflict-detection regressions could bypass the default suite.

## Impact

This is test-only: there are no runtime, schema, API, or dependency changes. Regressions in the production batch queries now fail the normal server test run.

## Validation

- mandatory `checkFile` checks passed for all three modified test files
- SuperSync server TypeScript build passed
- SuperSync server suite: 45 files, 923 tests passed
- targeted mutation checks confirmed the exact-size boundary and `lt`/`lte` guards fail under their corresponding mutations
- pre-push repository lint completed with zero errors (10 unrelated existing warnings)
- pre-push root suites passed, including 13,306 Angular tests and 13,292 timezone tests

Closes #9205

### DIFF
--- a/e2e/tests/work-view/sections.spec.ts
+++ b/e2e/tests/work-view/sections.spec.ts
@@ -46,7 +46,9 @@ test.describe('Sections', () => {
   const clickAddSection = async (
     page: import('@playwright/test').Page,
   ): Promise<void> => {
-    await page.getByRole('menuitem', { name: 'Add Section' }).click();
+    const menu = page.getByRole('menu');
+    await expect(menu).not.toHaveClass(/mat-menu-panel-animating/);
+    await menu.getByRole('menuitem', { name: 'Add Section' }).click();
   };
 
   /** Fill the dialog-prompt input and submit. */
@@ -221,7 +223,7 @@ test.describe('Sections', () => {
     await wrapper.waitFor({ state: 'visible', timeout: 5000 });
     await wrapper.dispatchEvent('contextmenu');
 
-    await page.getByRole('menuitem', { name: 'Add Section' }).click();
+    await clickAddSection(page);
     await submitPromptDialog(page, 'Right-Click Section');
 
     await expect(sectionByTitle(page, 'Right-Click Section')).toBeVisible();
--- a/packages/super-sync-server/tests/conflict.spec.ts
+++ b/packages/super-sync-server/tests/conflict.spec.ts
@@ -1,18 +1,25 @@
-import { describe, expect, it } from 'vitest';
+import { Prisma } from '@prisma/client';
+import { describe, expect, it, vi } from 'vitest';
 import {
+  detectConflictForEntities,
   getConflictEntityIds,
+  getEntityConflictKey,
   isSameDuplicateOperation,
   isSameIncomingOperation,
   isSameDuplicateTimestamp,
+  prefetchLatestEntityOpsForBatch,
   pruneVectorClockForStorage,
   resolveConflictForExistingOp,
   stableJsonStringify,
 } from '../src/sync/conflict';
 import {
+  CONFLICT_DETECTION_ENTITY_BATCH_SIZE,
   DuplicateOperationCandidate,
   MAX_VECTOR_CLOCK_SIZE,
   Operation,
 } from '../src/sync/sync.types';
+
+const TASK_TIME_DELTA_ACTION_TYPE = '[TimeTracking] Sync time spent';
 
 const op = (overrides: Partial<Operation> = {}): Operation => ({
   id: 'op-1',
@@ -69,10 +76,13 @@ describe('conflict helpers', () => {
     ).toBe(true);
   });
 
-  it('includes a divergent scalar entityId in the incoming conflict set', () => {
+  it('builds a deduplicated incoming conflict set that includes a divergent scalar', () => {
     expect(
       getConflictEntityIds(op({ entityId: 'task-scalar', entityIds: ['task-array'] })),
     ).toEqual(['task-scalar', 'task-array']);
+    expect(
+      getConflictEntityIds(op({ entityId: 'task-1', entityIds: ['task-1'] })),
+    ).toEqual(['task-1']);
   });
 
   it.each([false, true])(
@@ -122,6 +132,12 @@ describe('conflict helpers', () => {
     expect(isSameDuplicateOperation(duplicateCandidate(), 1, incoming, 60_000)).toBe(
       false,
     );
+  });
+
+  it('rejects an otherwise-identical duplicate operation from a different user', () => {
+    expect(
+      isSameDuplicateOperation(duplicateCandidate({ userId: 2 }), 1, op(), 60_000),
+    ).toBe(false);
   });
 
   it('accepts batch retries with identical entityIds', () => {
@@ -255,12 +271,12 @@ describe('conflict helpers', () => {
   it('accepts concurrent additive task-time deltas for the same task', () => {
     const result = resolveConflictForExistingOp(
       op({
-        actionType: '[TimeTracking] Sync time spent',
+        actionType: TASK_TIME_DELTA_ACTION_TYPE,
         vectorClock: { 'client-a': 1 },
       }),
       'task-1',
       {
-        actionType: '[TimeTracking] Sync time spent',
+        actionType: TASK_TIME_DELTA_ACTION_TYPE,
         clientId: 'client-b',
         vectorClock: { 'client-b': 1 },
       },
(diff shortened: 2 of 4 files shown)


johannesjo (commented):
## Review

The sync-test part of this PR does exactly what #9205 asked for, and I verified that rather than taking the description on trust. **One actionable issue, and it is in the drive-by e2e commit (`6b71531`), not in the conflict-query work.**

---

### Verification: 33-mutation sweep, both test versions

I mutated `src/sync/conflict.ts` 33 ways and ran the **full default server suite** against master's test files and this PR's test files.

| | master tests | this PR |
|---|---|---|
| mutations killed | **1 / 33** | **25 / 33** |

Master's single kill is incidental — `conflict-detection.spec.ts:221` throws `Unmocked raw query` when the SQL text lacks `DISTINCT ON`. Everything else survived on master, including: `AND FALSE` on either batch query, `server_seq DESC` → `ASC`, dropping the `actionType` / `clientId` / `serverSeq` projections, dropping the `user_id` or `entity_type` filter, the chunk off-by-one, `lt` → `lte`, dropping the scalar fold entirely, and restoring #8334's old mutually-exclusive `CASE`. This PR kills all of them.

⚠️ Methodology note for anyone repeating this: `tests/security-fixes.spec.ts > should escape HTML special characters in privacy template` is **flaky** (5s timeout on `await import('../src/server')` under load). It produced 5 false `KILLED` results in my first pass. Every number above was re-confirmed by re-running the mutation individually and requiring a *relevant* spec to fail.

Also confirmed green: full suite 923/923, `checkFile` on all three files, and TypeScript on the three touched files.

### 2 gaps that remain (not blockers)

The legacy `GLOBAL_CONFIG:misc` → `tasks` fold at the end of `prefetchLatestEntityOpsForBatch` is only half-guarded. The `lt` boundary is now covered, but these two survive **in the default suite _and_ in `tests/integration/conflict-detection-sql.integration.spec.ts`**:

1. `orderBy: { serverSeq: 'desc' }` → `'asc'` — folds the *oldest* legacy misc row, so the incoming `tasks` write is compared against a stale clock.
2. `(legacyMiscOp.serverSeq ?? -1) > (currentTasksOp.serverSeq ?? -1)` → `true` — a stale legacy row always beats a newer real `tasks` op.

Neither suite can catch them because all four `insertConfigOp` tests in the integration spec seed **exactly one** misc row and never a `tasks` row, so `currentTasksOp` is always `undefined` and the `>` never evaluates. Two more mock cases in `conflict.spec.ts` (two legacy misc rows; one legacy misc + one newer `tasks`) would close it.

---

### 🔴 `6b71531` — the menu-animation guard is a coin flip

The failure it targets is real: at `556e679` the `CI / Tests` job failed on `sections.spec.ts:209 › creates a section via right-click on the work-view background`, timing out in `submitPromptDialog` — the Add Section click never opened the dialog. So the diagnosis is right.

But the guard does not reliably fire. In `@angular/material` 21.2.12, `mat-menu-panel-animating` is bound to `_isAnimating()`, which is `signal(false)` and is only set to `true` inside `_onAnimationStart` — i.e. on the `animationstart` **DOM event**, which fires a frame *after* the panel is attached. `expect(menu).not.toHaveClass(...)` resolves on its first successful poll, so if that poll lands in the gap, the wait is a no-op.

I measured it. Replacing the assertion with a probe that reads the class at exactly the same point, 16 runs of the two menu tests:

```
PROBE animating=false getAnimations=[_mat-menu-enter]   <- guard no-ops, animation IS running
PROBE animating=true  getAnimations=[_mat-menu-enter]   <- guard works
PROBE animating=false getAnimations=[]                  <- genuinely settled
```

Roughly half the reads saw `animating=false` while `_mat-menu-enter` was still running — the guard passes straight through in exactly the case it is meant to catch.

`getAnimations()` closes that window, because the CSS animation exists from the moment it is applied, before `animationstart` is dispatched:

```ts
const menu = page.getByRole('menu');
await menu.evaluate((el) =>
  Promise.all(
    el.getAnimations({ subtree: true }).map((a) => a.finished.catch(() => undefined)),
  ),
);
await menu.getByRole('menuitem', { name: 'Add Section' }).click();
```

(The `.catch` is needed because `finished` rejects if the menu closes and the animation is cancelled.) I ran the full `sections.spec.ts` with this three times: **30/30 passed**.

Separately: this commit is an e2e flake fix with no relationship to #9205, and the PR body still describes the change as three server test files. Worth either splitting it out or mentioning it in the description.

---

### Minor

- **`createTransaction()` only implements `$queryRaw`.** A PGlite prefetch test that passes a `GLOBAL_CONFIG`/`tasks` pair will die on `tx.operation.findFirst`. A throwing stub pointing at the `conflict.spec.ts` mock would save the next person the detour.
- **Positional param indexing** (`params[2]`, `params[0]`) is now load-bearing in `conflict.spec.ts` and `sync.service.spec.ts`. It fails loudly rather than silently, so it is fine — but the comment in `conflict.ts` that justifies keeping the SQL inline names `conflict-detection.spec` as the reason, and that spec matches on the SQL *text*, not on param positions. The comment should now name the two specs that actually depend on the ordering.
- **Stale comment** in `tests/integration/conflict-detection-sql.integration.spec.ts`: "Unit tests mock `$queryRaw` and can only verify the JS transformation around the query" is no longer true. Related: three layers now cover this SQL (mocks / PGlite / real PG). Since the integration spec's unique value is being the only place a real `PrismaClient` runs, it may be worth trimming it to that.
- **`tests/` is not in `packages/super-sync-server/tsconfig.json`'s `include`**, so "SuperSync server TypeScript build passed" does not cover any test file. (I checked the new code separately with an ad-hoc config — it is clean.)
- **`server_seq bigint` → `integer`** in the PGlite DDL is a fidelity *fix*, not a regression (`schema.prisma`: `Int`, `0_init`: `INTEGER`).
- **Removing the "full-state op (`entity_id` NULL)" PGlite test loses nothing.** I mutated the guard to `o.entity_ids || ARRAY[o.entity_id]` and the suite is unaffected — a NULL array element unnests to a NULL `eid`, which neither `= ANY` nor the equijoin can match. It was an equivalent mutant either way.
- 6 of the 8 PR survivors are equivalent or perf-only (the `eid = ANY` and `entity_ids &&` prefilters, the JOIN's `entity_type` half, the second `Set` in `detectConflict`, both NULL-scalar guards). Consumers only `get()` the keys they requested, so extra rows are invisible to them. The issue classifies these the same way, which matches what I measured.

PR #9214, opened 2026-07-21
fix(plugin): dispatch FINISH_DAY action so the finishDay plugin hook fires

temal32:
## Problem

The `finishDay` plugin hook (`PluginHooks.FINISH_DAY`) can be registered successfully, but it **never fires** — the NgRx action it listens for is never dispatched anywhere in the app.

There are exactly two references to the `FINISH_DAY` action type in `src/`:

1. The effect that forwards it to plugins ([`plugin-hooks.effects.ts`](https://github.com/johannesjo/super-productivity/blob/master/src/app/plugins/plugin-hooks.effects.ts#L249-L257)):

   ```ts
   finishDay$ = createEffect(
     () =>
       this.actions$.pipe(
         filter((action) => action.type === 'FINISH_DAY'),
         tap(() => {
           this.pluginService.dispatchHook(PluginHooks.FINISH_DAY);
         }),
       ),
     { dispatch: false },
   );
   ```

2. The `actionsToExecuteBeforeFinishDay` field in `daily-summary.component.ts`, which defines `[{ type: 'FINISH_DAY' }]` but was never referenced anywhere — dead code.

Since nothing dispatches the action, the effect's filter never matches and `dispatchHook(PluginHooks.FINISH_DAY)` is never called — for any plugin, web or desktop. (The built-in issue-provider integrations are unaffected because they use the internal `BeforeFinishDayService.addAction()`, which is not reachable from the plugin API.)

### Steps to reproduce

1. Install a plugin that declares `"hooks": ["finishDay"]` and calls `api.registerHook('finishDay', () => console.log('finishDay fired'))`.
2. The console confirms registration: `Plugin <id> registered hook: finishDay`.
3. Go to the daily summary and click *Finish day* — the callback is never invoked.

## Fix

Dispatch the actions from the existing `actionsToExecuteBeforeFinishDay` field at the start of `DailySummaryComponent.finishDay()`, so plugins are notified before the day is wrapped up and done tasks are archived. This wires up the field as originally intended and keeps the hook's fire-and-forget semantics (the effect uses `tap`, archiving is not blocked on plugin handlers).

## Verification

Diagnosed and verified with a live plugin (an issue-provider plugin that books tracked time to GitLab on finish day) against v18.15.1: hook registration succeeds but the handler never runs when finishing the day; after dispatching `FINISH_DAY` the handler runs as expected. Also verified via source search that no other dispatcher of this action exists.

### DIFF
--- a/src/app/pages/daily-summary/daily-summary.component.ts
+++ b/src/app/pages/daily-summary/daily-summary.component.ts
@@ -37,7 +37,7 @@ import {
 import { DateService } from 'src/app/core/date/date.service';
 
 import { EntityState } from '@ngrx/entity';
-import { Action } from '@ngrx/store';
+import { Action, Store } from '@ngrx/store';
 import { TranslatePipe, TranslateService, TranslateStore } from '@ngx-translate/core';
 
 import { IS_ELECTRON } from '../../app.constants';
@@ -126,6 +126,7 @@ export class DailySummaryComponent implements OnInit, OnDestroy, AfterViewInit {
   private readonly _syncWrapperService = inject(SyncWrapperService);
   private readonly _operationWriteFlushService = inject(OperationWriteFlushService);
   private readonly _beforeFinishDayService = inject(BeforeFinishDayService);
+  private readonly _store = inject(Store);
   private readonly _simpleCounterService = inject(SimpleCounterService);
   private readonly _dateService = inject(DateService);
   private readonly _metricService = inject(MetricService);
@@ -341,6 +342,11 @@ export class DailySummaryComponent implements OnInit, OnDestroy, AfterViewInit {
   }
 
   async finishDay(): Promise<void> {
+    // Notify plugins (finishDay hook) before the day is wrapped up and tasks are
+    // archived — the plugin-hooks effect listens for the FINISH_DAY action.
+    this.actionsToExecuteBeforeFinishDay.forEach((action) =>
+      this._store.dispatch(action),
+    );
     try {
       await this._beforeFinishDayService.executeActions();
       // Wait for any ongoing sync to complete before archiving to avoid DB lock errors.


johannesjo (commented):
Thanks for tracking this down — the diagnosis is correct and well evidenced. I verified the premise independently:

- Only two `'FINISH_DAY'` references exist in `src/`: the effect (`plugin-hooks.effects.ts:249-257`) and the field (`daily-summary.component.ts:266`).
- `git log -S actionsToExecuteBeforeFinishDay` returns exactly one code commit — `0f9ceb8928` (Nov 2023, *"outline before finish day handler"*). The field was added as an outline and never wired up, so the hook has never fired since it was introduced.

I also confirmed the change is **sync-safe**: the dispatched action carries no `PersistentActionMeta`, and both `OperationLogEffects.persistOperation$` and `operationCaptureMetaReducer` gate on `isPersistentAction` (`meta.isPersistent === true`), so no operation is created and nothing reaches other devices. `plugin-hooks.effects.ts` injects `LOCAL_ACTIONS`, so remote replay can't re-fire the hook either.

Four things before this can go in.

---

### 1. Blocker — two existing unit tests fail

`npm run test:file src/app/pages/daily-summary/daily-summary.component.spec.ts` at `24979c3`:

```
✗ should wait for sync to complete before archiving tasks
✗ should not archive when before-finish actions fail
    TypeError: Cannot read properties of undefined (reading 'forEach')
    at daily-summary.component.ts:347:42
TOTAL: 2 FAILED, 5 SUCCESS
```

(7/7 pass on the base commit.) Those specs call `DailySummaryComponent.prototype.finishDay.call(receiver)` with hand-built receivers that have neither `actionsToExecuteBeforeFinishDay` nor `_store`; because the dispatch sits *before* the `try {`, it throws synchronously out of the async function.

CI hasn't surfaced this yet — the only check that has run is `welcome`; the unit-test job (`ci.yml` → `npm run test`) still needs maintainer approval for a first-time contributor.

Both receivers need `_store: { dispatch: jasmine.createSpy('dispatch') }` plus the field — and ideally an assertion on the dispatch, which brings us to (4).

### 2. The hook fires on aborted and failed finish-days

The dispatch is the first statement in `finishDay()`, ahead of every early return:

- `executeActions()` rejects → `FINISH_DAY_SYNC_ERROR` snack → `return`, nothing archived — but plugins have already been told the day finished. The user retries → the hook fires a **second** time.

For your own stated use case (booking tracked time to GitLab) that is a duplicate booking on a flow that visibly failed. Moving the dispatch to just below the `try/catch` block costs nothing and still satisfies *"before the day is wrapped up and tasks are archived"*.

(The Electron confirm-dialog dismissal path — `isConfirm === undefined` → `return` — would still fire the hook, but the built-in `BeforeFinishDayService` handlers behave the same way there, so that one is at least consistent.)

### 3. Fire-and-forget races the archive — consider the awaited path instead

In the non-Electron path the only gap between the dispatch and `_moveDoneToArchive()` is `executeActions()` plus the sync wait. A plugin that reads today's tasks to compute tracked time can lose that race and find them already archived.

Note what the built-in equivalent does: `GitlabIssueEffects` registers via `BeforeFinishDayService.addAction()`, which `finishDay()` **awaits** — so it is guaranteed to see the tasks before archiving. As written, plugins get a strictly weaker guarantee than the internal integration they are replicating.

Awaiting is cheap and safe here: `PluginService.dispatchHook()` returns `Promise<void>`, and `PluginHooksService._invokeWithTimeout` caps each handler at `HOOK_TIMEOUT_MS = 5000` **and swallows handler errors** — so a slow or broken plugin can neither stall nor abort finish-day.

Registering the hook dispatch as a `BeforeFinishDayService` action from the plugin layer would reuse the existing building block, guarantee before-archive ordering, and let the dead field, the magic-string action and the `finishDay$` effect all be deleted. That is the shape I'd prefer. If you'd rather keep the minimal diff, please at least move the dispatch per (2).

Worth knowing for this decision: an in-repo simplification audit already catalogued this field as dead code slated for removal (`docs/non-sync-code-simplification-audit.md`, candidate C02-014 — *"the two-line dead Action/actionsToExecuteBeforeFinishDay field deletion"*). Nobody had spotted that the hook depended on it. So the two efforts collide, and it's better to settle the mechanism now than to leave a half-wired field behind.

### 4. Missing payload and missing test

- `FinishDayPayload { date: string }` is declared in the public plugin API (`packages/plugin-api/src/types.ts:188`), but `plugin-hooks.effects.ts:254` calls `dispatchHook(PluginHooks.FINISH_DAY)` with no payload — plugins receive `undefined` where the typed API promises `{ date }`. Pre-existing, but harmless only while the hook never fires; this PR is the first moment any plugin can observe it, so it's the cheap moment to fix. `this.dayStr` is right there.
- A bug fix wants a test. A short spec asserting that `finishDay()` dispatches `FINISH_DAY` — and does *not* dispatch it when the pre-archive step fails — would cover the fix and would have caught both (1) and (2).

**Minor:** `'FINISH_DAY'` is a bare magic string duplicated between component and effect, present in no action-creator file and not in the `ActionType` enum. If the store dispatch stays, export a shared const. And since nothing ever writes to `actionsToExecuteBeforeFinishDay`, the array indirection buys nothing.

---

To be clear on the verdict: the bug is real, the fix earns its place, and the reproduction work here is appreciated. (1) is a hard merge blocker; (2) and (4) are small; (3) is the one worth a quick discussion before you rework it.

temal32:
@johannesjo and? Will you merge it?

johannesjo (commented):
Thanks for the ping, and sorry for the wait.

Short answer: not yet. Nothing has changed on the PR since my review, and the blocking item from it is still open. I re-verified it today against current `master` rather than trusting the old run.

### The blocker still reproduces

`24979c3` applied on top of today's `master` (`32c1610c`, merges cleanly):

```
$ npm run test:file src/app/pages/daily-summary/daily-summary.component.spec.ts
✗ should not archive when before-finish actions fail
✗ should wait for sync to complete before archiving tasks
    TypeError: Cannot read properties of undefined (reading 'forEach')
    at daily-summary.component.ts:347:42
TOTAL: 2 FAILED, 5 SUCCESS
```

The same spec with this PR's five lines reverted and nothing else changed: `TOTAL: 7 SUCCESS`. So both failures belong to this PR.

The cause is unchanged. Both specs drive the method as `DailySummaryComponent.prototype.finishDay.call(receiver)` with hand-built receivers (spec lines 151 and 187) that have neither `_store` nor `actionsToExecuteBeforeFinishDay`, and the new dispatch sits above the `try {`, so it throws synchronously.

One thing to know: CI has never actually run on this branch. Six workflows are sitting at `action_required` because this is your first PR here, and only `welcome` completed. So the checks list is not evidence of anything either way. I will approve the runs, and the two failures will show up there too.

### The open design question, decided

You may have been waiting on me for point 3, so here is the call: please go the `BeforeFinishDayService` route rather than the store dispatch.

The reason is not style. A store dispatch cannot be awaited, so it can never give plugins the ordering the hook advertises. `executeActions()` is awaited by `finishDay()` before `_moveDoneToArchive()`, so a handler registered there is guaranteed to see today's tasks. Fire and forget is a race, and when sync is disabled the window before archiving is a single microtask chain, which an iframe plugin doing postMessage round trips will normally lose. For your own GitLab booking case that is the difference between working and intermittently booking nothing.

It also deletes more than it adds: `actionsToExecuteBeforeFinishDay` (`daily-summary.component.ts:266`), the `'FINISH_DAY'` magic string, and the `finishDay$` effect (`plugin-hooks.effects.ts:249`) all go away, and `daily-summary.component.ts` needs no change at all, which makes the spec breakage above disappear rather than requiring new mocks. `GitlabIssueEffects` is the in-tree precedent for the pattern.

Third, smaller point: `anyAction$` (`plugin-hooks.effects.ts:411`) is unfiltered, so dispatching `{ type: 'FINISH_DAY' }` would also push a synthetic action into every plugin's generic `action` hook, for an action type that exists in no action creator and in no `ActionType` enum.

**There is one wrinkle you will hit,** which is why I am spelling the shape out. The payload needs the summary's day, and the plugin layer does not know it. `finishDay()` can run for a past day: `dayStr` is set from the route param and `isForToday` goes false (`daily-summary.component.ts:309-310`), so `dateService.todayStr()` in the plugin layer would be wrong. The day has to be threaded through the service:

```ts
// before-finish-day.model.ts
export type BeforeFinishDayAction = (dayStr: string) => Promise<'SUCCESS' | 'ERROR'>;

// before-finish-day.service.ts
async executeActions(dayStr: string): Promise<'SUCCESS' | 'ERROR'> {
  const results = await Promise.all(this._actions.map((action) => action(dayStr)));
  ...
}

// plugin layer
this._beforeFinishDayService.addAction(async (dayStr) => {
  await this._pluginService.dispatchHook(PluginHooks.FINISH_DAY, { date: dayStr });
  return 'SUCCESS';
});
```

The component call becomes `executeActions(this.dayStr)`, `GitlabIssueEffects` can ignore the new argument, and the existing spies accept it without changes.

The registration has to run at app start, before anyone can press the button. `GitlabIssueEffects` does it from its constructor, and `PluginHooksEffects` already injects `pluginService` and is registered eagerly (`root-store/feature-stores.module.ts:209`), so its constructor is a workable home if you want to keep it next to the other hooks.

Awaiting plugin handlers there is safe, which I checked rather than assumed. `dispatchHook()` returns a promise, and `_invokeWithTimeout` races each handler against `HOOK_TIMEOUT_MS = 5000` and swallows handler errors (`plugin-hooks.ts:18`, `:55`, `:111` and the `catch` at `:128`). A broken plugin can neither abort finish-day nor stall it forever. The cap is per handler and handlers run sequentially, and `dispatchHook` returns immediately when nothing registered the hook, so users without plugins pay nothing.

Two properties of the mechanism you would be joining, so they do not surprise you later:

- `executeActions()` runs before the Electron confirm dialog, so the hook fires even if the user then clicks that dialog away and nothing is archived. The built-in GitLab action already behaves that way, so I am fine with it, but the hook is "before finish day", not "the day definitely finished".
- `executeActions()` returns `'SUCCESS' | 'ERROR'` and `finishDay()` ignores the return value, so returning `'ERROR'` does not stop the archive. Only a thrown error does. Pre-existing, just do not rely on it.

### Still needed beyond that

- **Fill in the payload.** `FinishDayPayload { date: string }` is declared (`packages/plugin-api/src/types.ts:188`) and mapped in `HookPayloadMap` (`:260`), but the current call passes nothing, so plugins would get `undefined` where the public API promises `{ date }`. It is dormant today only because the hook never fires. This PR is the first moment a plugin can observe it, so it is the cheap moment to fix.
- **A test.** A short spec asserting the hook is dispatched with the expected date, and that it runs before archiving, would cover the fix and would have caught the two failures above. Under this shape it belongs with the plugin hooks specs and does not touch the daily summary receivers at all.

If you would rather not do the rework, say so and I will finish it on top of your commit with you as the author. Finding this was the valuable part, and that stands either way.

johannesjo:

Follow-up pass at `24979c3`. The head has not moved since my review of 2026-07-30, so this is a status update rather than a new review: everything in that review still stands, and I re-checked its anchors against today's `master` instead of trusting the old run.

### Still open

- **`src/app/pages/daily-summary/daily-summary.component.ts:347`** – CI has now run on this head and confirms the blocker publicly: `TOTAL: 2 FAILED, 13304 SUCCESS`, both failures `TypeError: Cannot read properties of undefined (reading 'forEach')` at `daily-summary.component.ts:347:42`, from `should wait for sync to complete before archiving tasks` and `should not archive when before-finish actions fail`. Both spec receivers are unchanged on current `master`, so this is not a stale result – it reproduces today.
- **The design call from 2026-07-30 stands**: register the hook through `BeforeFinishDayService` rather than dispatching a store action. Under that shape the component change shrinks to threading the day into the call (`executeActions(this.dayStr)`), which the existing spies accept unchanged, so the two failures above disappear rather than needing new mocks.
- **Payload and test** are still outstanding: `FinishDayPayload { date: string }` (`packages/plugin-api/src/types.ts:194`, mapped at `:266`) is still unfilled, and there is still no test asserting the hook fires.

### Also worth knowing

The branch is now 426 commits behind `master`, but it still merges cleanly and nothing here has been overtaken. I re-verified each anchor on current `master` and none moved: the dead field at `daily-summary.component.ts:266`, `finishDay$` at `plugin-hooks.effects.ts:249-258` (still on `LOCAL_ACTIONS`, still calling `dispatchHook` with no payload), the unfiltered `anyAction$`, and `executeActions()` still taking no `dayStr`. So the rework is still exactly the rework that was described.

### Questions

- Do you still want to do the rework yourself? Either answer is fine – the offer to finish it on top of your commit with you as the author is still open, and a one-line reply is enough to unblock this in either direction.

---

johannesjo:
Taking you up on the hand-off you offered — implemented in #9900 (`efd168c`), in the shape we discussed.

Rather than dispatching a `FINISH_DAY` action for the effect to filter on, the hook now registers itself through `BeforeFinishDayService.addAction()`, which is where the daily summary actually goes (`daily-summary.component.ts` calls `executeActions()` directly; that action type had no dispatcher, which is exactly why the documented hook never fired). Registering there also means the day is not archived until every listening plugin has had its turn, and a plugin that throws is logged as `'ERROR'` without blocking the user. The now-dead `actionsToExecuteBeforeFinishDay = [{ type: 'FINISH_DAY' }]` field is removed, and two specs cover the fire and the throw path.

I left `executeActions(dayStr)` out for now — no consumer needs the parameter yet, and it's easy to add when one does.

Closing once #9900 lands. Thanks for finding this and for the patience.

PR #9230, opened 2026-07-22
chore(tools): delete the never-wired single-test wrapper

KamilDev:
## Problem

`tools/test-file.js` has no caller. `package.json` has no `bin` field, no script references it, and no workflow or doc invokes it. The documented `npm run test:file` is an independent script (`cross-env TZ='Europe/Berlin' ng test --watch=false --include`) that never went through the wrapper.

Being unreachable is also why two defects in it went unnoticed:

1. It interpolated the file path into a `shell: true` `execSync` string — the same command-injection shape `tools/check-file.js` had before #9132.
2. It hardcoded `./node_modules/.bin/` paths, so on Windows it died before running any test:

```
🧪 Running tests for some.spec.ts...
❌ Test failures:
'.' is not recognized as an internal or external command,
operable program or batch file.
```

I offered this as a follow-up on #9132, assuming the fix was to port that PR's `execFileSync(process.execPath, [jsEntry, ...])` approach over. Checking first showed the file is unreachable, and that [`docs/non-sync-code-simplification-audit.md`](https://github.com/super-productivity/super-productivity/blob/master/docs/non-sync-code-simplification-audit.md) already covers it as **C06-009 — Delete the never-wired single-test wrapper** (safety: confirm, maintainability: confirm, consensus: confirm), so hardening it would mean maintaining a script nothing calls.

## Solution

Delete `tools/test-file.js`. Both defects go with it, and no supported command changes.

This follows C06-009's authoritative scope verbatim: delete this file only, with no replacement and no edit to `package.json` or the documented `npm run test:file` command. The audit entry's status line is left untouched — no prior commit has flipped a status in that document, so the convention there is to keep it as the decision record rather than a progress tracker.

## Type of Change

- [x] Other (dead-code removal)

## Verification

Per C06-009's required verification:

- Repeated the exact path, basename, `import`/`require`, `bin`, script, workflow, and documentation searches: the only remaining mentions of `test-file.js` are the two lines in the audit document itself.
- Ran `npm run test:file src/app/util/check-key-combo.spec.ts` after the deletion — `TOTAL: 8 SUCCESS`.
- `package.json` and contributor documentation are unchanged (the diff is one deleted file).
- `npm run lint && npm run test` pass (the pre-push hook); the remaining lint warnings are the pre-existing grandfathered `max-lines` ones on `master`.

## Checklist

- [x] I have included relevant changes to the documentation/wiki — n/a, no documented command changes
- [x] I have run `npm run checkFile` on changed `.ts`/`.scss` files — n/a, no `.ts`/`.scss` files changed
- [x] I have added tests for my changes (if applicable) — n/a, deletion only


johannesjo:
Thank you very much! <3

Issue #5175, opened 2025-09-25
💡 Add AND/OR toggle for Board > Create/Edit > Panel > Tags (both Required and Excluded)


nightgolfer:
### Problem Statement

When setting up filtering for Board Panels, both ```Required Tags``` and ```Excluded Tags``` assume 'AND', with no way to select 'OR'.

**Scenario**:
- I wish to create a board with two columns. 
- On the left, I want all Tasks with tags #x and #y that are currently Undone and Scheduled. 
- On the right, I want all Tasks with tags #x and #y that are currently Undone and Unscheduled.
- In my scenario, #x and #y are never assigned to the same Task, but they are related and relevant to eachother.
- Because the ```Required Tags``` filter uses 'AND' logic, my Board will always be empty.

### Possible Solution

Add a toggle to both the ```Required Tags``` and ```Excluded Tags``` Panel sections, allowing the user to select whether the selected tags should _all_ be present in the shown tasks, or if only _one (or more)_ of the tags need to be present, e.g. "AND / OR". To stick to the current functionality, "AND" should be selected as default.

<img width="399" height="812" alt="Image" src="https://github.com/user-attachments/assets/b773be19-7a4d-4941-a8a8-c377e2948578" />

### Describe alternatives you've considered

A more minimalist alternative could be a toggle for "ALL", or the slightly less clean "REQUIRE ALL":

<img width="399" height="812" alt="Image" src="https://github.com/user-attachments/assets/e981982e-b9b9-451b-9b29-f99fd562c83b" />

PR #9233, opened 2026-07-22
feat(task-view-customizer): support excluding tags and multiple terms in the tag filter

baflo:
The tag filter accepted a single include term only. It now splits the input on commas and supports a ! prefix to exclude: 'work, !waiting' shows tasks carrying a work-tag and no waiting-tag; '!private' alone hides everything tagged private (untagged tasks pass). Include terms are ANDed; each term matches tags by case-insensitive substring as before, with a term matching several tags matching any of them. Unchanged: empty input, the Not-specified preset, and the empty result for an include term matching no tag. A mat-hint under the filter input documents the syntax.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo (changes requested):
Thanks for this — the implementation is clean, and I want to lead with the part I'd like to keep.

**There is a real bug fix buried in here.** The old code was `_allTags.find(...)` — the *first* tag in sidebar tree order whose title contains the needle. With tags `Work` and `Homework`, typing `work` silently showed only one of them's tasks. Your `filter(...)` + any-of matching fixes that, and your spec `should match any of several tags matching one term` is a genuine regression test for it: I reverted the implementation to `matchingTags[0]` and that spec goes red. **That fix stands alone as a ~3-line change and I'd merge it on its own.**

The `,` / `!` grammar around it is what I'd like to hold. Details below.

---

## Verification I ran

- `npm run test:file` on the customizer spec: **58/58 pass**. `prettier --check` clean on all 5 files. CI 17/17 green.
- I sabotage-tested the six new specs — they are **not vacuous**:

| mutation applied to `_filterByTagTerms` | result |
| --- | --- |
| include `every` → `some` (AND becomes OR) | 1 FAILED |
| drop the exclusion line (`:617`) | 4 FAILED |
| `matchingTags` → `matchingTags[0]` (old `find` semantics) | 1 FAILED |
| drop the `.trim()` after `slice(1)` (`:601`) | 1 FAILED |
| remove `if (!matchingTags.length) return []` (`:610`) | **58 SUCCESS** |

That last row is a small note: line 610 is behaviour-neutral. Pushing an empty include `Set` already makes `.some()` false for every task, so the function returns `[]` either way. Harmless short-circuit, just not load-bearing.

- Also verified the good news on blast radius: the filter value is `FilterOption.preset`, persisted **localStorage-only** (`LS.TASK_VIEW_CUSTOMIZER_BY_CONTEXT`), with no second consumer anywhere — not synced, not exported, no plugin/REST surface. So nothing here is hard to undo.

## Blocking

### 1. `!` already means "deadline" in this app

`src/app/features/tasks/short-syntax.ts:32` — `const CH_DEADLINE = '!'`, documented in the wiki as `Task @monday !friday` ([3.04-Short-Syntax](https://github.com/super-productivity/super-productivity/blob/master/docs/wiki/3.04-Short-Syntax.md)). `#` is the tag sigil, `+` project, `@` due.

Giving `!` a second, unrelated meaning (NOT) in another text input teaches two meanings for one character. If a sigil ships here, `-` is the near-universal convention for exclusion in a filter box (GitHub, Gmail) and doesn't collide.

### 2. Persisted filter values are silently reinterpreted, and there's no escape hatch

The preset is restored verbatim per work context (`task-view-customizer.service.ts:98-100`, `:116`, `:197`), so on upgrade every saved value is re-parsed under the new grammar:

- a saved `!Urgent` flips from *show tasks tagged `!Urgent`* to *hide tasks tagged `Urgent`* — a silent inversion;
- a tag title containing a comma becomes unreachable through this filter entirely.

Tag titles are free text (the settings form only requires non-empty), and imports/plugins can create arbitrary ones. Impact is bounded — localStorage, the user can retype — but it's silent, and there's no quoting to recover from it.

### 3. `!today` silently excludes nothing

`_allTags` is `tagsInTreeOrder`, which includes `TODAY_TAG` (the reducer inserts it — `tag.reducer.ts:190`). But `TODAY` is virtual: membership comes from `dueDay`/`dueWithTime` and **never** from `task.tagIds` ([ARCHITECTURE-DECISIONS.md](https://github.com/super-productivity/super-productivity/blob/master/ARCHITECTURE-DECISIONS.md) Decision #2).

So `work, !today` matches `TODAY_TAG`, then excludes nothing, and quietly returns today's tasks too. This is the asymmetry that worries me most about the design in general: **an include term that matches nothing returns an obviously-empty list, but an exclude term that matches nothing silently returns the wrong list.** A typo like `!waitng` shows you exactly the tasks you asked to hide.

The service already has a TODAY-free helper — `_tagsInSidebarOrder()` at `:466`.

## Should fix

### 4. The input never shows the stored value back — pre-existing, but this PR makes it matter

`task-view-customizer-panel.component.html:254`:

```html
customizerService.selectedFilter().preset === OPTIONS.filter.types.tag
```

That compares the *value* to the literal string `'tag'`; `.type ===` was clearly intended (same bug on the project input at `:280`, both from `5bcfa5e6bb`). So on first render the binding is `''` unless the user literally typed "tag" — after a reload the tag filter is active but the box is blank.

With a single substring that's merely confusing. With a compound `work, !waiting` the user can neither read back nor edit their active query. If the syntax ships, this one-character fix should ship with it.

### 5. Operator-only input is an active match-all filter

`,,,` or `!` alone produce zero effective terms, so every task passes — while `isCustomized` is still true and the view stays on the customized path. Worth resetting to the default filter, or defining and testing it as empty input.

### 6. The hint doesn't mention the AND

`Comma-separate terms, prefix ! to exclude` doesn't tell the user that `work, home` *hides* tasks carrying only one of the two. That's the least guessable part of the semantics.

## The bigger question: does the grammar earn its place?

Per [AGENTS.md](https://github.com/super-productivity/super-productivity/blob/master/CLAUDE.md), the first question on a new feature is whether it should exist. The Problem/Solution sections are empty and there's no linked issue — I searched open and closed issues and found no request for multi-term or negated tag filtering in the work view.

What I did find is that **the app already models this concept properly, elsewhere**: Board panels have `includedTagIds` / `excludedTagIds` plus `includedTagsMatch` / `excludedTagsMatch: 'all' | 'any'` (`boards.model.ts:22-28`), driven by a real tag picker (`boards-form.const.ts:59-107`). Open issue #5175 asks for AND/OR toggles on exactly that.

So this PR introduces a *second, weaker* vocabulary for a concept the app already expresses precisely: substrings instead of tag IDs, AND-only includes, any-match excludes, no OR, no escaping. The long-term cost is two divergent grammars plus the follow-up requests that always come (OR, quoting, grouping) — a filter box slowly becoming a query language, and a permanent hint line explaining it.

## What I'd suggest

1. Split out the `find` → `filter` fix (finding above the fold) and land it on its own — it's a real bug and it's already covered by your spec.
2. For the exclusion feature, open an issue describing the workflow so demand can be gauged. If the real need is "hide one tag", a single explicit exclusion affordance would be much cheaper than comma-composition.
3. If the grammar does land: `-` instead of `!`, match against `_tagsInSidebarOrder()` so `TODAY` can't be addressed, fix the `:254` binding, and say "require all" in the hint.

Happy to be overruled on the scope call — that one's @johannesjo's. The code itself is good work.

PR #9235, opened 2026-07-22
feat: move to bottom

### LINKED ISSUE #6650: 💡 Move to bottom quick action
### Problem Statement

As I work on things throughout the day, I frequently find that I'll need to come back to something -- maybe I need feedback on something, maybe I need to wait for a process to complete -- and it would be really useful to have a "Move to bottom" right underneath "Move to top" 

### Possible Solution

Right underneath the current `⬆️ Move to top` button, add a `⬇️ Move to bottom` button. It should be visible / hidden with the same conditions. It should use similar logic to move the selected task to the bottom of the current view.

Drag and scroll and drag and scroll, and scroll back to the top. Several times per day.

### Additional context

I can absolutely do this myself if I can find the time, just jotting it down here for later.

calvin-fisher:
closes #6650

## Problem

As I work on things throughout the day, I frequently find that I'll need to come back to something -- maybe I need feedback on something, maybe I need to wait for a process to complete -- and it would be really useful to have a "Move to bottom" right underneath "Move to top" for the same reasons that having a quick "Move to top" is useful.

## Solution

Added a `⬇️ Move to bottom` button right underneath the current `⬆️ Move to top` button. It is visible / hidden with the same conditions. It uses similar logic to move the selected task to the bottom of the current view.

I also added a couple simple tests on both buttons to assert basic moving behavior and subtask invisibility.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


calvin-fisher:
Is any additional testing required on this one? Automated testing covers the likely failure cases. Manual testing covered tasks with different projects, dragging and moving in and out of being a subtask. Ordering on today / project views, and ordering is disabled for tag view and subtasks, as expected. Seems clean and straighforward.

src/app/features/tasks/task-context-menu/task-context-menu-inner/task-context-menu-inner.component.ts
```diff
@@ -506,6 +506,10 @@ export class TaskContextMenuInnerComponent implements AfterViewInit, OnDestroy {
     this._taskService.moveToTop(this.task.id, this.task.parentId, false);
   }
 
+  moveToBottom(): void {
+    this._taskService.moveToBottom(this.task.id, this.task.parentId, false);
```
johannesjo, line 510: **Important:** This always passes `false`, although the component already detects whether the menu is inside `.backlog` as `this.isBacklog`.

Reproduction on the PR preview:
1. Enable a project backlog and add two backlog tasks.
2. Open the top backlog task's context menu and choose **Move to bottom**.
3. The order does not change. The runtime log shows `[WorkContextMeta] Move Task To Bottom in Today`, so the regular-list action is dispatched instead of the backlog action.

`TaskService.moveToBottom()` only updates `backlogTaskIds` when its third argument is `true`. The smallest fix is to pass `this.isBacklog` here and add a regression test for this context. The adjacent `moveToTop()` has the same hard-coded argument, so it is worth correcting both while this path is being covered.

johannesjo (commented):
I think this action is worth adding. It forms a natural pair with **Move to top**, exposes an already-existing capability/default shortcut, and avoids adding a setting or new data/model surface. I also checked the deployed preview at desktop and 390×844; the context menu remains usable.

I found one behavior that should be fixed before merging; see the inline comment. I reproduced it in the deployed preview with two tasks in an enabled project backlog, and the runtime action log confirms the wrong ordering path is selected.

Please also tighten the tests around the actual menu interaction/backlog case. The current tests assert the hard-coded `false`, so they preserve the bug, and the large synthetic fixture does not exercise list movement. The unrelated removal in `lww-projectid-convergence.integration.spec.ts` should be reverted.

All current CI checks are green. With the backlog handling corrected and covered, the UX premise and overall implementation approach look sound.

src/app/features/tasks/task-context-menu/task-context-menu-inner/task-context-menu-inner.component.ts
calvin-fisher: The `moveToTop` line is unedited since https://github.com/super-productivity/super-productivity/commit/4d9c7d1fd2df73d36449ede8a551c4e9f98ef28b and I just copied the existing behavior.

But I can take a look at improving it while adding this. I've never used the backlog view in a project so I'll have to learn what it's about first.

johannesjo (commented):
I re-reviewed the current `010f02b` head, including two independent static-review passes. The feature premise and the reuse of the existing ordering service are sound, and all current required checks are green. Two behaviors still need correction before merge:

1. **Backlog ordering** — `task-context-menu-inner.component.ts:510` always passes `false`. In a project backlog, `TaskService.moveToBottom()` therefore dispatches the regular-list action against `project.taskIds`; the backlog-only task is absent there, so `arrayMoveToEnd()` returns the list unchanged. Pass `this.isBacklog` and cover the backlog context.

2. **Section ordering** — the same call updates only project/tag `taskIds`. For a task displayed inside a section, `section.taskIds` is authoritative, and the section reducer does not handle this action, so **Move to bottom** leaves the visible section order unchanged. Route this case through the existing single-op section placement path (or another solution that updates the authoritative section order atomically) and add a focused regression test.

Regarding the linked reply: it is fair that the adjacent `moveToTop()` line predates this PR. I would not make repairing that pre-existing behavior the blocker here. The new `moveToBottom()` action still must not copy the same context bug; fixing both while the context plumbing is touched would be welcome but is not required for the existing action.

The added test block currently creates no ordered list/section state (and the stated three-parent fixture defines only two), so the “already at top/bottom” cases only verify spy calls and cannot catch either failure above. Please replace or tighten those cases around regular, backlog, and section contexts. The wiki sentence should also describe an **unfinished top-level task** in the **current list**, which matches the actual visibility and includes backlog/section contexts.

Finally, please revert the unrelated removal in `lww-projectid-convergence.integration.spec.ts` so this feature diff stays scoped.

calvin-fisher:
@johannesjo sorry I only get around to checking this infrequently! It looks like you took care of this one already? 
https://github.com/super-productivity/super-productivity/issues/6650

If so, it looks like I can close this PR?

I have a minor UI improvement I was about to add separately here:
https://github.com/super-productivity/super-productivity/pull/9555

johannesjo:
Yes — you're right, and thanks for checking back.

#6650 was closed by #9469, which shipped **Move to bottom** along with the `isBacklog` fix we discussed in this thread (it corrected the pre-existing `moveToTop()` call site too, so backlog moves are no longer a silent no-op), extended both entries to subtasks, added the keyboard-shortcut hints in the menu, and is covered by `e2e/tests/task-basic/subtask-move-top-bottom.spec.ts` including a backlog context-menu case. It's released in v18.18.0.

So there's nothing left to merge here — this branch would now conflict in the wiki and produce a duplicate `moveToBottom()` member. Closing as superseded.

Sorry for landing it separately while your PR was open — that's on me. Thank you for filing #6650 and for the implementation work; it's what got this prioritized. Looking forward to #9555.

PR #9055, opened 2026-07-15
fix(locale): localize ISO 8601 recurring-task start date & weekday labels (#8987)

johannesjo:
## Problem

Follow-up to #9013 (now merged; fixes #8987). That PR localized ISO 8601 spelled-out weekday/month names inside `CustomDateAdapter`, fixing every `<mat-calendar>`. But some components format dates **directly** via `toLocaleDateString(currentLocale(), …)`, bypassing the adapter — so they still leaked Swedish under the ISO option.

Reported by @cdnia on #9013: in the **Add Recurring Task Config** dialog (task detail → Recur → `+`), the **"Start date" value** renders in Swedish (e.g. `ons 15 juli 2026`) and ignores the UI language — the *label* translates but the *value* doesn't.

**Root cause:** the ISO 8601 option stores `dateTimeLocale = 'sv'` (the sentinel that yields `YYYY-MM-DD` + a 24h clock). `plannedStartDateStr` formats the value with `toLocaleDateString(currentLocale(), { weekday: 'short', month: 'short', … })`, so the spelled-out names come out Swedish. Same class of bug also affects the dialog's **quick-setting labels** (`Weekly on onsdag`, `On the first onsdag …`), which build the weekday via `buildRepeatQuickSettingOptions` — shared with the add-task-bar.

## Solution

Mirror #9013's principle at the one shared boundary: **spelled-out names follow the UI language; numeric dates and clock times keep the configured locale** (so ISO stays `YYYY-MM-DD` and 24h is preserved).

- Add `DateTimeFormatService.textLocale` = `isoTextLocale() ?? currentLocale()` — the canonical locale for spelled-out names.
- `plannedStartDateStr`: format the spelled-out date with `textLocale()`; keep the clock time on `currentLocale()`.
- `buildRepeatQuickSettingOptions`: new `weekdayLocale` param used only for the spelled-out weekday; numeric day/month keep `locale` so ISO day-first ordering (`15/7`) is preserved. The dialog and `add-task-bar-actions` pass `textLocale()`. The dialog's formly options memo key now tracks `textLocale`.

## Type of Change

- [x] Bug fix

## Validation

- `date-time-format.service.spec.ts` — `textLocale` resolves to the UI language under the ISO option (tracking language switches) and to the configured locale otherwise.
- `build-repeat-quick-setting-options.spec.ts` — `weekdayLocale` localizes the weekday while numeric day/month stay on `locale`; defaults to `locale` when omitted.
[...]

PR #9258, opened 2026-07-23
fix(schedule): use UI locale for ISO headings

### LINKED ISSUE #9182: 🚨 ISO 8601 date-format option renders Swedish date at the top right corner in Schedule views (View Day / View Week / View Month)
### Version Used

v18.15.0 — Linux, installed via AUR (superproductivity-bin)

### Current Behavior

With **ISO 8601: 24-hour, YYYY-MM-DD** selected as the Datetime format locale (Menu → Settings → General tab → expand Localization → Datetime format locale and I pick ISO 8601: 24-hour, YYYY-MM-DD), dates at the top right corner in the Schedule views (View Day / View Week / View Month) render in **Swedish** (for example `sön 19 juli 2026`  in the View Day view, `Week 29 juli 19-25` in the View Week view, `juli 2026` in the View Month view) instead of English.
Confirmed in **v18.15.0** 
 I included screenshots of what appears for each of the three views (View Day / View Week / View Month).
Image of the **View Day** view:

<br>

Image

<br><br><br>

Image of the **View Week** view:

<br>

Image

<br><br><br>

Image of the **View Month** view:

<br>

Image

### Expected Behavior

Date at the top right corner in the **Schedule** views (View Day / View Week / View Month) should appear in the app's UI language, while keeping the **ISO 8601: 24-hour, YYYY-MM-DD** date format and 24-hour time.

### Steps to Reproduce

1. Go to Settings → General → Localization → **Datetime format locale**.
2. Select **ISO 8601: 24-hour, YYYY-MM-DD**.
3. Open each of the three **Schedule** views (View Day / View Week / View Month).
4. Observe the date at the top right corner in each of the three **Schedule** views (View Day / View Week / View Month) are Swedish instead of English.

### Console Output

```shell

```

### Documentation

pratyushsinghal7:
## Problem

The ISO 8601 option uses `sv` to preserve year-first dates and 24-hour time. The Schedule header also used that locale for textual day and month names, so an English UI rendered Swedish headings.

Closes #9182.

## Solution

Use `DateTimeFormatService.textLocale()` for Schedule day, week, and month headings, with regression coverage for all three views.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

## Checklist

## Testing

- 61 focused Schedule tests
- 165 adjacent locale and Schedule tests
- Full pre-push suite in Europe/Berlin and America/Los_Angeles

## Assistance

AI-assisted; reviewed and tested by the contributor.

### DIFF
--- a/src/app/features/schedule/schedule/schedule.component.spec.ts
+++ b/src/app/features/schedule/schedule/schedule.component.spec.ts
@@ -18,6 +18,7 @@ import { ScheduleDay } from '../schedule.model';
 import { CalendarEventActionsService } from '../../calendar-integration/calendar-event-actions.service';
 
 describe('ScheduleComponent', () => {
+  const mockLocalization = signal({ firstDayOfWeek: 1, dateTimeLocale: 'en-US' });
   let component: ScheduleComponent;
   let fixture: ComponentFixture<ScheduleComponent>;
   let mockTaskService: jasmine.SpyObj<TaskService>;
@@ -28,6 +29,7 @@ describe('ScheduleComponent', () => {
   let mockGlobalConfigService: jasmine.SpyObj<GlobalConfigService>;
   let mockCalendarEventActionsService: jasmine.SpyObj<CalendarEventActionsService>;
   beforeEach(async () => {
+    mockLocalization.set({ firstDayOfWeek: 1, dateTimeLocale: 'en-US' });
     // Create mock services
     mockTaskService = jasmine.createSpyObj('TaskService', ['currentTaskId']);
     (mockTaskService as any).currentTaskId = signal(null);
@@ -94,7 +96,7 @@ describe('ScheduleComponent', () => {
     mockGlobalConfigService = jasmine.createSpyObj('GlobalConfigService', [], {
       // Pin the date locale so Intl-formatted headers are deterministic across
       // runners (an en-GB runner would render "20 Jan", not "Jan 20").
-      localization: signal({ firstDayOfWeek: 1, dateTimeLocale: 'en-US' }),
+      localization: mockLocalization,
       cfg: signal(undefined),
     });
 
@@ -162,6 +164,49 @@ describe('ScheduleComponent', () => {
       mockScheduleService.getMonthDaysToShow.and.returnValue(days);
       fixture.detectChanges();
       expect(component.headerTitle()).toMatch(/April\s+2026/);
+    });
+
+    it('uses the UI language for ISO day, week, and month headings', () => {
+      const translate = TestBed.inject(TranslateService);
+      translate.setTranslation('en', {
+        F: { WORKLOG: { CMP: { WEEK_NR: 'Week {{nr}}' } } },
+      });
+      translate.use('en');
+      mockLocalization.set({ firstDayOfWeek: 1, dateTimeLocale: 'sv' });
+
+      mockLayoutService.selectedTimeView.set('day');
+      mockScheduleService.getDaysToShow.and.returnValue(['2026-07-19']);
+      component['_selectedDate'].set(new Date(2026, 6, 19));
+      fixture.detectChanges();
+      expect(component.headerTitle()).toContain('Jul');
+
+      mockLayoutService.selectedTimeView.set('week');
+      mockScheduleService.getDaysToShow.and.returnValue([
+        '2026-07-13',
+        '2026-07-14',
+        '2026-07-15',
+        '2026-07-16',
+        '2026-07-17',
+        '2026-07-18',
+        '2026-07-19',
+      ]);
+      component['_selectedDate'].set(new Date(2026, 6, 13));
+      fixture.detectChanges();
+      expect(component.headerTitle()).toContain('Jul');
+
+      mockLayoutService.selectedTimeView.set('month');
+      const monthDays = Array.from({ length: 35 }, (_, i) => {
+        const date = new Date(2026, 6, 1 + i);
+        return [
+          date.getFullYear(),
+          String(date.getMonth() + 1).padStart(2, '0'),
+          String(date.getDate()).padStart(2, '0'),
+        ].join('-');
+      });
+      mockScheduleService.getMonthDaysToShow.and.returnValue(monthDays);
+      component['_selectedDate'].set(new Date(2026, 6, 1));
+      fixture.detectChanges();
+      expect(component.headerTitle()).toContain('July');
     });
   });
 
--- a/src/app/features/schedule/schedule/schedule.component.spec.ts
+++ b/src/app/features/schedule/schedule/schedule.component.spec.ts
@@ -16,6 +16,8 @@ import { SCHEDULE_CONSTANTS } from '../schedule.constants';
 import { GlobalConfigService } from '../../config/global-config.service';
 import { ScheduleDay } from '../schedule.model';
 import { CalendarEventActionsService } from '../../calendar-integration/calendar-event-actions.service';
+import { registerLocaleData } from '@angular/common';
+import localeSv from '@angular/common/locales/sv';
 
 describe('ScheduleComponent', () => {
   const mockLocalization = signal({ firstDayOfWeek: 1, dateTimeLocale: 'en-US' });
@@ -28,6 +30,9 @@ describe('ScheduleComponent', () => {
   let mockGlobalTrackingIntervalService: jasmine.SpyObj<GlobalTrackingIntervalService>;
   let mockGlobalConfigService: jasmine.SpyObj<GlobalConfigService>;
   let mockCalendarEventActionsService: jasmine.SpyObj<CalendarEventActionsService>;
+
+  beforeAll(() => registerLocaleData(localeSv, 'sv'));
+
   beforeEach(async () => {
     mockLocalization.set({ firstDayOfWeek: 1, dateTimeLocale: 'en-US' });
     // Create mock services
@@ -166,20 +171,27 @@ describe('ScheduleComponent', () => {
       expect(component.headerTitle()).toMatch(/April\s+2026/);
     });
 
-    it('uses the UI language for ISO day, week, and month headings', () => {
+    const useIsoDatesWithUiLanguage = (language: string): void => {
       const translate = TestBed.inject(TranslateService);
-      translate.setTranslation('en', {
+      translate.setTranslation(language, {
         F: { WORKLOG: { CMP: { WEEK_NR: 'Week {{nr}}' } } },
       });
-      translate.use('en');
+      translate.use(language);
       mockLocalization.set({ firstDayOfWeek: 1, dateTimeLocale: 'sv' });
+    };
 
+    it('uses the UI language for the ISO day heading', () => {
+      useIsoDatesWithUiLanguage('en');
       mockLayoutService.selectedTimeView.set('day');
       mockScheduleService.getDaysToShow.and.returnValue(['2026-07-19']);
       component['_selectedDate'].set(new Date(2026, 6, 19));
       fixture.detectChanges();
-      expect(component.headerTitle()).toContain('Jul');
 
+      expect(component.headerTitle()).toContain('Jul');
+    });
+
+    it('uses the UI language for the ISO week heading', () => {
+      useIsoDatesWithUiLanguage('en');
       mockLayoutService.selectedTimeView.set('week');
       mockScheduleService.getDaysToShow.and.returnValue([
         '2026-07-13',
@@ -192,8 +204,12 @@ describe('ScheduleComponent', () => {
       ]);
       component['_selectedDate'].set(new Date(2026, 6, 13));
       fixture.detectChanges();
-      expect(component.headerTitle()).toContain('Jul');
 
+      expect(component.headerTitle()).toContain('Jul');
+    });
+
+    it('uses the UI language for the ISO month heading', () => {
+      useIsoDatesWithUiLanguage('en');
       mockLayoutService.selectedTimeView.set('month');
       const monthDays = Array.from({ length: 35 }, (_, i) => {
         const date = new Date(2026, 6, 1 + i);
@@ -206,7 +222,19 @@ describe('ScheduleComponent', () => {
       mockScheduleService.getMonthDaysToShow.and.returnValue(monthDays);
       component['_selectedDate'].set(new Date(2026, 6, 1));
       fixture.detectChanges();
+
       expect(component.headerTitle()).toContain('July');
+    });
+
+    it('preserves Gregorian dates and Latin digits in Persian ISO day headings', () => {
+      useIsoDatesWithUiLanguage('fa');
+      mockLayoutService.selectedTimeView.set('day');
+      mockScheduleService.getDaysToShow.and.returnValue(['2026-07-19']);
+      component['_selectedDate'].set(new Date(2026, 6, 19));
+      fixture.detectChanges();
+
+      expect(component.headerTitle()).toContain('19');
+      expect(component.headerTitle()).not.toMatch(/[۰-۹]/);
     });
   });
 
--- a/src/app/features/schedule/schedule/schedule.component.ts
+++ b/src/app/features/schedule/schedule/schedule.component.ts
@@ -200,7 +200,7 @@ export class ScheduleComponent {
   headerTitle = computed(() => {
     const days = this.daysToShow();
     if (!days.length) return '';
-    const locale = this._dateTimeFormatService.currentLocale();
+    const locale = this._dateTimeFormatService.textLocale();
 
     if (this.isDayView()) {
       // On tablet width and below the full date clips, so drop to month + day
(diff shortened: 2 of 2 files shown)


johannesjo (commented):
## Summary

Using `textLocale()` is the right direction for #9182 and keeps the implementation narrow, but the shared substitution changes more than textual names for one supported UI locale. The new week/month assertions also do not reproduce the old behavior when this spec runs in isolation.

## PR Discussion Context

There were no prior human reviews or inline threads at this head. All completed checks are green; the aggregate CI `Tests` job was still running when reviewed, with its unit-test step already successful.

## Issues

### Important

- **[src/app/features/schedule/schedule/schedule.component.ts:203](https://github.com/super-productivity/super-productivity/blob/d48dfd7520aa36236147f66de19a3b98b81bea19/src/app/features/schedule/schedule/schedule.component.ts#L203) — Preserve the ISO calendar and numbering when localizing words.** With Persian UI (`fa`) and the ISO setting (`dateTimeLocale: 'sv'`), `textLocale()` returns `fa`. `Intl.DateTimeFormat('fa', ...)` resolves to the Persian calendar and `arabext` numbering, so Gregorian `2026-07-19` becomes `۱۴۰۵ تیر ۲۸, یکشنبه` (compact: `۲۸ تیر`). Week/month remain Gregorian with Latin digits through Angular `formatDate`, while the base implementation used `sv` and was Gregorian/Latin. Changing Schedule views can therefore change the represented date. I checked all 28 supported UI locales; `fa` is the one locale with this non-Gregorian default. In ISO mode, please localize the weekday/month words while explicitly retaining Gregorian calendar and Latin numbering (for example, conditional `calendar: 'gregory'` and `numberingSystem: 'latn'` options), or split numeric and textual formatting as the sibling Schedule components do. Add a `fa` + ISO day-title regression case.

- **[src/app/features/schedule/schedule/schedule.component.spec.ts:195](https://github.com/super-productivity/super-productivity/blob/d48dfd7520aa36236147f66de19a3b98b81bea19/src/app/features/schedule/schedule/schedule.component.spec.ts#L195), [line 209](https://github.com/super-productivity/super-productivity/blob/d48dfd7520aa36236147f66de19a3b98b81bea19/src/app/features/schedule/schedule/schedule.component.spec.ts#L209) — The week/month assertions currently pass against the old implementation.** This spec does not register Swedish Angular locale data, so `safeFormatDate(..., 'sv')` catches NG0701 and falls back to English. A focused head run passes `61/61`; reverting only line 203 to `currentLocale()` yields `1` failure and `60` successes, with only the day assertion failing (`19 juli` versus `Jul`). Week and month still pass old code, and registrations from other specs can make that mutation coverage suite-order dependent. Please register `localeSv` in this spec and preferably split or parameterize the three views so each regression is independently effective.

## Strengths

The change reuses the existing locale abstraction without adding state, dependencies, or new configuration, and non-ISO behavior remains unchanged.

## Verdict

**Needs changes** — preserve ISO calendar/numbering in Persian day view and make the week/month regression coverage effective.

pratyushsinghal7:
Addressed in `801a0a9c1`:

- ISO day headings now force the Gregorian calendar and Latin numbering while retaining localized Persian month/weekday text.
- Swedish Angular locale data is registered in the spec, so reverting the locale fix now breaks the week/month regressions.
- Day, week, and month checks are independent tests.
- Added a Persian ISO day-heading regression.

Focused result: `64 SUCCESS`; Prettier and lint pass.

johannesjo (commented):
## Summary

Follow-up review at `801a0a9`: the current head correctly fixes #9182 and resolves both earlier Important findings. ISO Schedule headings now use the UI language while day view explicitly preserves the Gregorian calendar and Latin digits.

## PR Discussion Context

The linked preview and reporter confirmation apply to the earlier `d48dfd7` head. There are no inline review threads. The follow-up commit addresses the Persian-calendar regression and makes the day/week/month regression tests independently effective.

## Issues

No remaining code findings.

The current head's six GitHub Actions workflow suites are all `action_required`, so `801a0a9` has not received current-head CI coverage yet. The PR Preview workflow is among them, so the published preview also does not contain the follow-up commit.

## Verification

- `checkFile` passes for both changed files.
- Schedule and date-format suites: `392/392` pass in `Europe/Berlin`.
- Focused Schedule component suite: `64/64` pass in `America/Los_Angeles`.
- Reverting `textLocale()` makes exactly the day/week/month ISO-language tests fail with Swedish output.
- Removing the Gregorian/Latin options makes the Persian regression fail with `۲۸ تیر`.
- Angular week/month formatting preserves Gregorian dates and ASCII digits across all 28 supported UI locales.

## Strengths

The production change stays confined to the existing locale abstraction, adds no dependency or persisted-state surface, and the tests now detect both the original bug and the intermediate Persian regression.

## Verdict

**Needs changes — CI gate only.** No source changes are requested. Once the workflows for `801a0a9` are approved and pass, this is ready to merge.

cdnia:
Hi @johannesjo, @pratyushsinghal7 , thank you again for working on this.

johannesjo (approved):
## Summary

Round 3 at `801a0a9`, an unchanged head. Both Important findings from the first round remain fixed, and the second round's only blocker has cleared: the workflows for this exact commit are now approved and green. No source changes requested.

## PR Discussion Context

@cdnia confirmed on #9182 that the preview renders the Schedule headings in the UI language across day, week, and month views. The sticky preview comment now points at a build of `801a0a9`, so my earlier note that the published preview predated the follow-up commit no longer applies. There are no inline review threads.

## Issues

None.

## Verification (re-run against current master `32c1610c7`)

- Merges cleanly; no master commit has touched either changed file since the PR base, and the Schedule work that landed in the meantime does not overlap.
- `checkFile` passes on both files.
- Schedule and date-time-format suites: `411/411` in `Europe/Berlin` and `411/411` in `America/Los_Angeles`.
- Sabotage 1, `textLocale()` reverted to `currentLocale()`: 3 failures, one per view (`19 juli`, `Week 29 · juli 13 – 19`, `juli 2026`). This closes the first round's finding that the week and month assertions passed against the old code; registering `localeSv` in the spec makes all three independently effective.
- Sabotage 2, `calendar`/`numberingSystem` options removed: 1 failure, `۲۸ تیر`. The Persian regression test is real.
- I re-derived the assumption that the week and month branches need no explicit Gregorian and Latin options rather than trusting my earlier round. Angular's `formatDate` emits ASCII digits and Gregorian dates for every locale probed (`fa`, `ar`, `hi`, `bn`, `th`, `my`, `ne`, `zh`, `ja`, `ko`, `ru`, `tr`), so `LLLL yyyy` and `MMM d` are safe on `textLocale()` alone.
- Non-ISO behavior is unchanged by construction: `isoTextLocale()` returns `null` unless `dateTimeLocale === 'sv'`, so `textLocale()` collapses to `currentLocale()` and `isIsoLocale` stays `false`.
- No ambiguity in the `sv` sentinel: `DateTimeLocales.sv` is offered in settings only as the ISO 8601 entry, with no separate Swedish option, so nobody selecting it is actually asking for Swedish month names.
- The remaining `currentLocale()` callers in the Schedule feature are all numeric or clock output (`'d'` day numbers, `formatMonthDay`, the hour ruler), which is what `DateTimeFormatService` documents. The textual surface is now consistently on the ISO text locale, matching `schedule-week` and `schedule-month`.

## Strengths

The fix stays inside the existing locale abstraction with no new dependency, setting, or persisted state, and it leaves numeric ISO formatting and the 24h clock untouched. The Persian case is the kind of second-order regression that usually ships silently, and it is now covered by a test that fails without the guard.

## Verdict

**Approve.** Ready to merge.

johannesjo:
Follow-up to my approval: I put the change through a deeper adversarial pass, mostly trying to break my own earlier conclusions. Nothing found blocks the merge, so the approval stands. Two things are worth recording, and there are two cheap test nits you can take or leave.

## Where I tried to prove myself wrong and failed

I doubted my own round-1 statement that `fa` is the only affected UI locale, on the theory that Thai defaults to the Buddhist calendar and Arabic to Arabic-Indic digits. Both suspicions are wrong. `th` is not a supported UI language at all (the `LanguageCode` enum, the 28 files in `src/assets/i18n/`, and the settings dropdown all agree on the same 28 codes), and bare region-less `ar` resolves to `gregory` + `latn`; only `ar-EG` / `ar-SA` carry `arab` digits, and the app passes bare `ar`. Re-probed all 28 codes in node, Chromium 141, Chromium 149, and Electron 41: `fa` is the only one where the new options change anything, and for the other 27 the output is byte-identical with and without them.

I also expected a crash. `textLocale()` now feeds an app UI language code into `new Intl.DateTimeFormat(...)`, which is not wrapped in try/catch the way `safeFormatDate` is, so a malformed BCP-47 tag would throw `RangeError` and take the Schedule header down. It cannot happen here: all 28 enum values are valid tags, and every write path into `isoTextLocale()` is gated on `LanguageCode`. The underscore forms (`pt_br`, `zh_tw`, `ro_md`) do throw, but those exist only as object keys, never as values.

One scope correction to my round-1 wording: "`fa` is the one locale" is true of the 28 UI languages on the ISO path. It is not true of the day-view header in general. With the date setting on "System default", `currentLocale()` resolves to the raw browser culture tag (`date-time-format.service.ts:145`), which is not restricted to the 28, so a `th-TH` or `fa-IR` browser still renders a non-Gregorian day header through the branch this PR does not guard. That is pre-existing and unchanged by this PR, so it is a separate issue rather than a review finding here.

## Worth a follow-up: the Gregorian rule now lives at one call site out of many

`schedule.component.ts:212-215` is the only place in the app that forces `gregory` + `latn` under the ISO setting. Every other `textLocale()` consumer formats without it, so with a Persian UI and ISO 8601 selected they still render Persian-calendar dates:

```
[forced gregory/latn] schedule day header (tablet)    -> 19 ژوئیه
[plain textLocale]    habit-tracker dateRangeLabel     -> ۲۸ تیر
[plain textLocale]    add-task-bar dueDateDisplay      -> ۲۸ تیر
[plain textLocale]    repeat-cfg plannedStartDateStr   -> ۱۴۰۵ تیر ۲۸, یکشنبه
[plain textLocale]    planner-calendar-nav monthLabel  -> ۱۴۰۵ تیر
[plain textLocale]    focus-session formatSelectedDate -> یکشنبه ۲۸ تیر
[plain textLocale]    scheduled-date-group.pipe        -> یکشنبه ۴/۲۸
```

`habit-tracker.component.ts:125-127` is the sharpest illustration: it uses the identical `{ month: 'short', day: 'numeric' }` options object as the Schedule tablet branch and produces `۲۸ تیر` where the header now produces `19 ژوئیه`.

I am not asking for that here. #9182 is about the Schedule header, the fix is correct for it, and widening the PR would be scope creep.

What makes it worth writing down is the contrast with how the sibling problem was handled. Choosing `textLocale()` over `currentLocale()` for spelled-out names is not left to convention: `eslint-local-rules/rules/require-text-locale.js` makes it a lint error, and its header explains why, noting that #8987 recurred across #9013, #9055 and #9056 precisely because `currentLocale()` is "the obvious-looking default at every new call site". That rule is why the seven sites above already call `textLocale()`.

This PR adds a second, narrower rule on top: under ISO, also force `gregory` and `latn`. That one has no lint coverage (the rule file never mentions `calendar` or `numberingSystem`) and `schedule.component.ts:213-214` is the only place in `src/` that sets either. By the same logic that justified `require-text-locale`, the next contributor writing a Persian-safe header will not know this rule exists. Extending `require-text-locale`, or folding it into `src/app/util/datetime-formatter.ts` (documented as the thing you "should use everywhere for date and time formatting"), would be the durable version. Worth a follow-up issue rather than anything in this PR.

To be clear about the Schedule tab itself: it is internally consistent and #9182 is fully fixed. The week and month branches go through Angular's `formatDate`, which is always proleptic Gregorian with ASCII digits, so they genuinely need no guard. The day-column headers in `schedule-week` and the weekday row in `schedule-month` already route through `isoTextLocale()`, and every tooltip and aria-label in the three templates is a translation key.

## Two test nits, both optional

**Pin `_isTablet` in the two new day-view tests.** This spec already establishes that convention at lines 1072 and 1081, with the comment "regardless of the test runner's window width". The new day-view cases at lines 183 and 229 do not, so which branch they exercise depends on the Karma window size. Both happened to hit the tablet branch here, which means the desktop branch's `weekday` and `year` fields are never asserted under ISO. Not flaky today, since both branches satisfy the assertions.

**The helper's line order is load-bearing.** In `useIsoDatesWithUiLanguage`, `translate.use(language)` writes the plain `currentLang` property, not the `_uiLangSig` signal, so on its own it does not invalidate a computed reading `textLocale()`. The three ISO tests recompute only because `mockLocalization.set(...)` follows it. Swap those two lines and the tests go stale-green while still passing. A one-line comment saying the `mockLocalization.set` must come last would stop someone tidying it into a trap later. Production reactivity is fine, because `LanguageService._set` calls both `translateService.use()` and `setUiLanguage()`, and the latter writes a real signal.

## Verification

- Full spec 64/64. 411/411 across `schedule` + `date-time-format` in both `Europe/Berlin` and `America/Los_Angeles`, and 902/902 over the wider locale-adjacent set (`schedule`, `date-time-format`, `locale*`, `planner`, `simple-counter`, `ui/pipes`). No pollution from the new `beforeAll(registerLocaleData(localeSv, 'sv'))`, which matches the established pattern in seven sibling specs, and `schedule-locale-ng0701.spec.ts`, which deliberately relies on `zh-cn` being unregistered, still passes.
- Sabotage, reverting `textLocale()`: 3 failures, one per view. Sabotage, dropping the calendar options: 1 failure, `۲۸ تیر`. Both reproduced independently, twice.
- Non-ISO output is unchanged structurally, not just empirically: `isoTextLocale()` returns `null` for every `dateTimeLocale` other than the `sv` sentinel, so `textLocale()` collapses to `currentLocale()` and `isIsoLocale` is `false`, leaving `dayOpts` untouched and both `safeFormatDate` calls receiving the identical string.
- `calendar` and `numberingSystem` options are ES2020 Intl (Chrome 80+); the repo's floor is Chrome 107 via `.browserslistrc`. An engine that did not support them would ignore them rather than throw, degrading to today's behavior.
- `headerTitle()` appears once in the template, outside any loop, so the per-evaluation `Intl.DateTimeFormat` construction is not on a hot path.

johannesjo:
Addendum to my previous comment. A last verification pass finished after I posted and turned up one coverage gap worth your attention. Still approving; this is a one-line test change and your call whether to take it.

## A mutation that survives the whole suite

```ts
const isIsoLocale = this._dateTimeFormatService.isoTextLocale() !== null;
const locale = isIsoLocale ? 'en' : this._dateTimeFormatService.textLocale();
```

`TOTAL: 64 SUCCESS`. Under ISO this hardcodes English and ignores the UI language entirely, which is the same bug class as #9182 itself, just displaced onto non-English users: a German user with ISO 8601 selected would get `July 2026` instead of `Juli 2026`, and nothing in the suite notices.

The reason is that all three of the day/week/month ISO tests use `en` as the UI language, so they cannot distinguish "formatted with the UI language" from "formatted with English". The Persian test does not close the gap either, because English output satisfies both of its assertions: `Jul 19` contains `19` and matches no Persian digits.

There is a compounding effect from `safeFormatDate`, which swallows NG0701 and falls back to `DEFAULT_LOCALE`, itself `en-GB`. So for the week and month branches, an `en` assertion additionally cannot distinguish "formatted with UI language `en`" from "threw and fell back". Asserting one non-English UI language closes both problems at once.

**Verified fix**, one line in the existing Persian test:

```ts
expect(component.headerTitle()).toContain('ژوئیه');
```

With that added, the mutation above fails (`Expected 'Jul 19' to contain 'ژوئیه'`) and the suite stays at 64/64 on the real code. A `de` case asserting `Juli` would work equally well and might read better.

## Credit where it is due: the `beforeAll` is doing more work than I said

I described `beforeAll(() => registerLocaleData(localeSv, 'sv'))` as fixing my round-1 finding. Having probed it, it is load-bearing in a stronger sense than that. With the registration removed and `textLocale()` reverted, the suite reports only `1 FAILED`: the day test. The week and month tests pass against the buggy code, because without Swedish locale data `safeFormatDate` catches NG0701 and quietly falls back to English, which is indistinguishable from the fix working.

So the registration is not defensive tidying. It is the thing that makes two of the three regression tests capable of failing at all, and removing it would silently restore exactly the flaw from the first round.

## Everything else from the previous comment stands

Sabotage results reproduced a third time and match exactly: reverting `textLocale()` gives 3 failures, one per view; dropping the calendar options gives 1 failure with `۲۸ تیر`, and both of that test's assertions fire. Restored tree back to 64/64 with an empty `git diff`.

cdnia:
Hi @johannesjo, thanks again for your great work on this. Can you please merge this.

johannesjo:
Thank you very much! <3

PR #9276, opened 2026-07-23
feat(planner): add Repeat button to schedule dialog

johannesjo:
Surface recurrence directly from the schedule dialog via a "Repeat"
button that opens the existing repeat-config dialog, so setting a task
to recur no longer requires the separate detail-panel row.

- Button mirrors the repeat dialog's own start-date button styling and
  shows the live recurrence label ("Daily", "Mon–Fri", ...) or
  "Does not repeat", with an edit/chevron affordance.
- Gated to top-level, non-issue tasks (matching the panel's Repeat row)
  and hidden in isSelectDueOnly mode to avoid a circular picker.
- Recurrence start is seeded from the date currently selected in the
  dialog, falling back to the task's due day / creation date.
- Label reads the repeat cfg via the store selector (not the service)
  to keep the dialog's dependency surface minimal.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_0176AgeeX2Tzv3KjmwZ7zWXC


johannesjo:
superseeded

johannesjo:
https://github.com/super-productivity/super-productivity/pull/9286

Issue #8436, opened 2026-06-16
🚨 Nextcloud Deck sync sets isDone: null instead of boolean, triggers data repair on startup


anschuetz:
### Version Used

v18.9.1 (macOS)

### Current Behavior

On startup, the data repair dialog appears reporting N issues fixed. 5 tasks imported from Nextcloud Deck have isDone: null stored instead of a boolean. The app auto-fixes this on every fresh start if the sync runs again and sometimes re-introduces the null values.

### Expected Behavior

isDone should always be stored as true or false, never null. The Nextcloud Deck sync handler should normalize the value from the API (e.g. !!card.done) before writing it to the task.

### Steps to Reproduce

Unable to reliably reproduce. The null values were discovered via the automatic data repair dialog on startup. All 5 affected tasks were Nextcloud Deck issues from the same project/provider.

**Suspected cause:**
  The two-way sync polling likely applies `card.done` from the
  Nextcloud Deck API directly to `task.isDone` without a `|| false`
  guard. If the Deck API returns `"done": null` for certain cards,
  the null propagates into storage.

  The initial import code appears safe (uses `isDone: J.data.isDone || !1`),
  so the issue is more likely in the sync/poll update path.

### Can you reproduce this reliably?

Sometimes, it's intermittent.

### Console Output

**Repair log output (from DevTools console):**
```shell
  [ValidateStateService] State invalid, asking user for confirmation...
  [auto-fix-typia-errors] Attempting validation auto-fix
    expected: "boolean"
    path: "task.entities["8IOFi4clG6hJ9o7wQjQ9Z"].isDone"
    valueType: "null"
  [auto-fix-typia-errors] Applied validation auto-fix
    fix: "falsey-to-false"
    replacementType: "boolean"
```
  (... 4 more identical fixes for other Nextcloud Deck tasks ...)

```shell
  Data repair executed: 6 issues fixed.
  Summary: {"entityStateFixed":0,"orphanedEntitiesRestored":0,
  "invalidReferencesRemoved":0,"relationshipsFixed":1,
  "structureRepaired":0,"typeErrorsFixed":5}


  **All 5 affected tasks:**
  - All have `issueType: NEXTCLOUD_DECK`
  - All belong to the same project/issue provider
  - `isDone` was stored as `null`, not `false` or `undefined`
```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
[...]

PR #9279, opened 2026-07-24
fix(caldav): pull completion state from server todos

### LINKED ISSUE #9278: 🚨 CalDAV: completion state never pulls — server-completed VTODOs don't tick tasks done (two defects)
### Bug description

The CalDAV issue provider never pulls completion state: a VTODO marked `STATUS:COMPLETED` on the CalDAV server does not tick its Super Productivity task done — regardless of the provider's Two-Way Sync configuration (including `Status sync direction = Pull only`, the default, and `Both`). Neither the periodic update poll nor the manual per-task **Update issue data** action applies the remote completion.

The **push** direction works correctly (ticking a task in SP writes `STATUS:COMPLETED`/`PERCENT-COMPLETE:100` to the server when status sync direction allows it), and import of open todos works. Only server → SP completion is dead.

### To reproduce

1. Configure a CalDAV Todo issue provider against any CalDAV server (reproduced with Radicale 3) with "Auto import to default project" and "Poll imported for changes and notify" enabled; leave Two-Way Sync `isDone` at its default (`Pull only`).
2. Let a todo import into a project as a task.
3. On the server, mark that VTODO `STATUS:COMPLETED` (+ `COMPLETED`, `PERCENT-COMPLETE:100`, bumped `SEQUENCE` so the etag hash changes).
4. Wait for the update poll (or run the task's context-menu **Update issue data**).

**Expected:** the SP task gets ticked done (this is what the Plainspace and Nextcloud Deck providers do for their `done` equivalents, e.g. #8436).
**Actual:** the poll finds no updates at all; the manual refresh shows the "issue was updated" banner but the task stays open.

### Root cause (two independent defects in `caldav-common-interfaces.service.ts`)

**1. The batch update poll queries and matches by SP's internal task id instead of the VTODO UID** — [`getFreshDataForIssueTasks`](https://github.com/super-productivity/super-productivity/blob/a51ea2b/src/app/features/issue/providers/caldav/caldav-common-interfaces.service.ts#L146-L161):

```ts
this._caldavClientService.getByIds$(
  tasks.map((t) => t.id),   // ← SP nanoid, never a VTODO UID
  cfg,
)
...
issueMap.has(task.id)        // ← same mismatch; map keys are VTODO UIDs
```

`CaldavIssue.id` is the VTODO UID, so the lookup must use `task.issueId` (as the Nextcloud Deck batch implementation does). As written, `getByIds$` filters the server's todos against a set of SP nanoids, always returns `[]`, and the poll never detects any change — this silently breaks *all* remote-change polling for CalDAV (title/notes edits too), not just completion.

**2. `getAddTaskData` never maps the issue's `completed` state to `isDone`** — [`getAddTaskData`](https://github.com/super-productivity/super-productivity/blob/a51ea2b/src/app/features/issue/providers/caldav/caldav-common-interfaces.service.ts#L70-L100) returns title/notes/dates/`issueLastSyncedValues` but no `isDone`, unlike Plainspace (`isDone: issue.isDone`) and Nextcloud Deck (`isDone: card.done`, guarded by specs since #8436). This is why even the single-task **Update issue data** path (which fetches by UID correctly via `getById$`) refreshes the data without ever ticking the box.

Both defects verified against `master` @ a51ea2b (unchanged since v18.15.1, where this was traced live against a Radicale server with wire capture).

### Additional context

- `getByIds$` itself is fine — it fetches with `filterOpen=false`, so completed todos are visible to it once queried by the right ids.
- Fix for both + specs (mirroring the Nextcloud Deck completion-pull tests): PR follows.

### Environment

- Version: 18.15.1 (web/Docker) — defects present on current `master` (a51ea2b)
- Platform: Browser (self-hosted Docker web app), CalDAV server: Radicale 3

ZsomborDane:
## Problem

Fixes #9278 — the CalDAV issue provider never pulls completion state: a VTODO marked `STATUS:COMPLETED` on the server doesn't tick its SP task done, regardless of the Two-Way Sync configuration. Two independent defects in `caldav-common-interfaces.service.ts`:

1. **`getFreshDataForIssueTasks` queried and matched server todos by SP's internal task id** (`tasks.map((t) => t.id)` / `issueMap.has(task.id)`) instead of the VTODO UID stored in `task.issueId`. `CaldavIssue.id` *is* the UID, so the batch update poll always matched nothing — silently breaking all remote-change polling for CalDAV (title/notes edits included), not just completion.
2. **`getAddTaskData` never mapped `completed` → `isDone`**, so even the single-task **Update issue data** path (which fetches by UID correctly) refreshed everything except the done state. Plainspace (`isDone: issue.isDone`) and Nextcloud Deck (`isDone: card.done`, spec-guarded since #8436) both map their equivalents.

## Solution

- `getFreshDataForIssueTasks` now queries `getByIds$` with `task.issueId` and keys all map lookups on it — the same pattern as the Nextcloud Deck batch implementation. (`getByIds$` itself was already fine: it fetches with `filterOpen=false`, so completed todos are visible once queried by the right ids.)
- `getAddTaskData` now returns `isDone: issueData.completed` (normalized to a boolean in `_mapTask`), placed before the date-field spreads so it composes the same way as the other mapped fields.
- Specs mirror the Nextcloud Deck completion-pull guards: `completed` ↔ `isDone` mapping both ways, single-task refresh ticking done, and a batch test asserting `getByIds$` is called with the VTODO UID (not the SP task id) and that the resulting `taskChanges` carry `isDone`/`issueWasUpdated`. Without the fix, 4 of the 5 new specs fail (verified by stashing the service change); with it, the CalDAV provider folder runs 67/67 green.

Verified end-to-end earlier against a live Radicale 3 server (wire capture): with these two changes the poll detects the server-side completion and ticks the task.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for the clear reproduction and focused fix. The UID correction in the batch path is right, and the new test correctly distinguishes the SP task nanoid from the VTODO UID.

I found two **Important** correctness issues that should be addressed before merge:

1. **The new `isDone` pull ignores the configured field direction.** The added [`isDone: issueData.completed`](https://github.com/super-productivity/super-productivity/blob/48355386b556370f248ef7dc9da97a2615acefba/src/app/features/issue/providers/caldav/caldav-common-interfaces.service.ts#L97) is returned by both refresh paths regardless of `cfg.twoWaySync.isDone`. Therefore, when the user selected `off` or `pushOnly`, any remote etag change can still close or reopen the local task. The [configuration labels](https://github.com/super-productivity/super-productivity/blob/48355386b556370f248ef7dc9da97a2615acefba/src/app/features/issue/providers/caldav/caldav-cfg-form.const.ts#L9-L13) explicitly promise that remote-to-local status application is disabled in those modes, while `getSyncConfig()` is currently only consumed by push-decision code. Because poll updates also contain `issueLastSyncedValues` / `issueWasUpdated`, the [push effect deliberately skips them](https://github.com/super-productivity/super-productivity/blob/48355386b556370f248ef7dc9da97a2615acefba/src/app/features/issue/two-way-sync/issue-two-way-sync.effects.ts#L155-L159), so `pushOnly` will not correct the overwrite. The smallest fix is to include pulled `isDone` only for `pullOnly` and `both` (defaulting missing legacy config to `pullOnly`) in both refresh paths, with tests for all four directions.

2. **The tests bypass the VTODO parsing case described as `STATUS:COMPLETED`.** The real parser currently sets `completed` solely from the presence of the separate [`COMPLETED` timestamp](https://github.com/super-productivity/super-productivity/blob/48355386b556370f248ef7dc9da97a2615acefba/src/app/features/issue/providers/caldav/caldav-client.service.ts#L211), while the added specs inject an already-normalized `CaldavIssue { completed: true }`. RFC 5545 lists `COMPLETED`, `PERCENT-COMPLETE`, and `STATUS` as independent optional VTODO properties; [`STATUS:COMPLETED` means the to-do is completed](https://www.rfc-editor.org/rfc/rfc5545.html#section-3.8.1.11), and [`PERCENT-COMPLETE:100` also indicates completion](https://www.rfc-editor.org/rfc/rfc5545.html#section-3.8.1.8). With the installed `ical.js`, a valid STATUS-only VTODO yields `status: "COMPLETED"` but `completed: null`, so this PR still maps it to `isDone: false`. Please normalize completion at the CalDAV boundary from the supported completion signals (preserving the existing `COMPLETED` behavior) and add a client-level regression test using raw VTODO data.

**Minor:** the [documentation guide](https://github.com/super-productivity/super-productivity/blob/48355386b556370f248ef7dc9da97a2615acefba/docs/documentation-guide.md#L10-L20) calls for updating the issue-integration comparison when provider behavior changes. The [CalDAV entry](https://github.com/super-productivity/super-productivity/blob/48355386b556370f248ef7dc9da97a2615acefba/docs/wiki/3.07-Issue-Integration-Comparison.md#L88-L89) currently mentions completion syncing back to the server, but not configurable server-to-SP completion pulls.

Separately, the main CI, CodeQL, E2E, packaging, and preview workflows are currently `action_required` rather than executed; only the welcome check has passed.

johannesjo:
I opened #9287 while exploring a more complete implementation of the review findings, but I’d prefer to keep this PR as the canonical fix and preserve your contribution. I’ve made #9287 a reference implementation rather than a competing merge candidate.

The additional cases that should land here are:

- honor `off`, `pullOnly`, `pushOnly`, and `both` when applying remote fields;
- heal `pullOnly` values when an earlier manual refresh already consumed the ETag without applying completion;
- preserve pending local `both`-mode edits when only an unrelated VTODO field changed;
- normalize completion from `COMPLETED`, `STATUS:COMPLETED`, or `PERCENT-COMPLETE:100`, with a regression test using raw VTODO data; and
- update the CalDAV integration documentation.

Would you like to incorporate these changes, or would you prefer that I push follow-up commits to your branch? Maintainer edits are enabled, and I’m happy to help either way.

johannesjo:
As offered in the review, I have pushed the remaining work to this branch (maintainer edits). Your original commit is untouched at the base; three follow-up commits complete the review findings:

- `ad0c65d16e` **fix(caldav): sync remote task completion** (from #9287): both refresh paths now route through a single `_getRefreshTaskChanges()` that applies the configured per-field directions (`off`, `pullOnly`, `pushOnly`, `both`; with missing legacy config, status and title default to `pullOnly` and notes to `off`), so a remote etag change can no longer close or reopen a task whose status direction is `off` or `pushOnly`. It also heals stale `pullOnly` values when an earlier refresh already consumed the ETag without applying completion. Includes the wiki updates for the configurable pull directions.
- `eb1df2772e` **fix(caldav): preserve pending local edits** (from #9287): mapped remote fields are compared with `issueLastSyncedValues`, so a VTODO-wide ETag change caused by an unrelated field no longer overwrites a pending local `both`-mode edit; the new baseline still updates so the pending edit pushes on the next cycle.
- `1eb8ea8378` **fix(caldav): normalize completion from STATUS and PERCENT-COMPLETE**: `_mapTask` now treats `COMPLETED`, `STATUS:COMPLETED`, and `PERCENT-COMPLETE:100` as completion signals (RFC 5545 lists them as independent optional properties), with regression specs that parse raw VTODO data through the real ical.js. The push dirty-check in `_updateTask` uses the same normalization; without that, reopening a STATUS-only completed todo never wrote to the server and the next poll immediately re-completed the local task (spec included, fails against the old check).

The branch also merges current `master`. Verified locally: all 81 CalDAV provider specs and 61 two-way-sync specs green; the STATUS-only, PERCENT-COMPLETE-only, and push dirty-check specs each fail when their fix is reverted. `checkFile` clean on all touched files, `check-doc-links --docs-only` clean on the changed wiki pages.

This supersedes the reference implementation in #9287, which I will close once this merges. Thanks again for the contribution and the clear reproduction!

PR #9298, opened 2026-07-24
fix(i18n): translate hardcoded strings and add tag row a11y label in …

### LINKED ISSUE #9297: 🚨 Several strings in task detail panel bypass i18n / tag row missing a11y label
### Version Used

18.15.1 (web / dev build via `npm run serve`)

### Current Behavior

In the task detail panel (`task-detail-panel.component.html`), four strings do not go through the `translate` pipe:

- "Created on" is hardcoded, even though a `CREATED_ON` key already exists in `en.json` (unused).
- "Completed on" is hardcoded, even though a `COMPLETED_ON` key already exists (unused).
- "Checklist" (notes panel header, `isMarkdownChecklist()` branch) is a hardcoded bare text node with no key.
- The tag row has an icon but no text label, so it is not translatable and is not announced to screen readers.

Result: these strings stay English regardless of selected language, and the tag row is unlabelled for assistive tech.

I have a fix ready and will open a PR referencing this issue.

### Expected Behavior

All four should respect the selected language, and the tag row should have an accessible label:

- "Created on" / "Completed on" wired to the existing `CREATED_ON` / `COMPLETED_ON` keys.
- A new `CHECKLIST` key added and used in the notes header, matching the `<span>{{ ... | translate }}</span>` pattern the sibling branches already use.
- A new `TAGS` key plus a `cdk-visually-hidden` label span in the tag row title slot (hidden rather than visible, since the row uses `type="fullSizeInput"` and `tag-edit` already renders chips plus a live input; this matches the existing hidden-label pattern used for the repeat value in the same component).

### Steps to Reproduce

1. Open any task's detail panel.
2. Switch the app to a non-English language.
3. Observe "Created on", "Completed on", and (for a task with a markdown checklist) "Checklist" still render in English.
4. Inspect the tag row: it has a label icon but no text label element for screen readers.

### Console Output

```shell

```

### Documentation

ba6ul:
…task detail panel #9297

## Problem

Four strings in `task-detail-panel.component.html` bypass the `translate` pipe, so they stay English regardless of the selected language, and the tag row has no text label for screen readers. Fixes #9297.

## Solution

- "Created on" / "Completed on" now use the existing `CREATED_ON` / `COMPLETED_ON` translation keys instead of hardcoded English text.
- Added a `CHECKLIST` key and used it in the notes panel header, matching the `<span>{{ ... | translate }}</span>` pattern the sibling branches use.
- Added a `TAGS` key and a `cdk-visually-hidden` label in the tag row title slot. Hidden rather than visible: the row uses `type="fullSizeInput"` and `tag-edit` already renders chips plus a live input, so a visible label would crowd it. Matches the existing hidden-label pattern used for the repeat value in the same component.

Known limitation: the hidden label is read in document order but not formally associated with the input inside `tag-edit`. An `aria-label` there would be more robust; happy to do that as a follow-up.

Only `en.json` was edited for strings; `t.const.ts` keys added to match. Other locales fall back automatically.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for working on this. I double-checked the current head (`d4cc93e`) and found two remaining items:

1. **The checklist translation key is not wired up.** `src/app/features/tasks/task-detail-panel/task-detail-panel.component.html:243` still renders the literal `Checklist`, so the markdown-checklist header remains English in non-English locales. Please render the new key there, matching the sibling branches, e.g. `<span>{{ T.F.TASK.ADDITIONAL_INFO.CHECKLIST | translate }}</span>`.

2. **The hidden `Tags` text is not associated with the tag editor.** In the live accessibility tree it appears as separate static text, while the focusable control is still named `+ Add Tag`. Screen-reader users tabbing to the control or navigating by form controls therefore do not get `Tags` as its accessible name. Please put a translated `aria-label`/`aria-labelledby` on the actual `mat-chip-grid`/input inside `TagEditComponent`. The existing `T.F.TAG.D_EDIT.LABEL` key already provides “Tags”, so the duplicate `F.TASK.ADDITIONAL_INFO.TAGS` key and parent hidden span can be avoided.

The created/completed date translations look correct. Targeted local validation passed (`checkFile`, translation-file validation, and all 28 task-detail-panel tests). The normal CI workflows are still awaiting first-time-contributor approval, so only the welcome check has run so far.

ba6ul:
All fixes complete and pushed:

- [x]  Wired up checklist translation key
- [x] Added aria-label to mat-chip-grid for accessibility
- [x] Removed unused TAGS key and hidden span

johannesjo:
Thanks for the follow-up. I carefully re-reviewed the current head (`fc760351`), including a merge-tree build against the current base and the live Chromium accessibility tree. **No blocking findings remain.**

- `Checklist`, `Created on`, and `Completed on` are now wired through the translation keys correctly.
- The translated `Tags` label is now on `MatChipGrid`, Angular Material’s composite tag-entry control—the control Material explicitly requires to have an accessible label. The live tree is `generic "Tags" → combobox "+ Add Tag"` when empty and `grid "Tags" → row … → combobox "+ Add Tag"` after adding a chip. I rechecked the input concern from my previous comment: the child is not unnamed; its existing placeholder provides the add-action name. Requiring a duplicate generic input label now would expand the earlier acceptance criterion unnecessarily. [Angular Material chips accessibility guidance](https://material.angular.dev/components/chips/overview)
- KISS/YAGNI: the four-file, `+11/-4` scope is appropriately small. It reuses the existing date/tag keys and established template pattern, adds only the genuinely missing `CHECKLIST` key, and warrants no new abstraction or documentation change.

Local validation on the current-base merge tree passed: `checkFile` for `t.const.ts`, i18n validation, Prettier for all four files, a full dev build, and all 28 task-detail-panel tests. The six normal repository workflows still show `action_required` pending first-time-contributor approval, so CI still needs to run and pass.

Non-blocking housekeeping: the PR description still describes the removed `TAGS` key/hidden-span approach and obsolete “known limitation”; it would be good to update it to match the final grid-label solution. The existing hardcoded `+ Add Tag` placeholder and broader chip-input structure are separate pre-existing debt—I would keep them out of this focused PR.

johannesjo:

PR #9302, opened 2026-07-25
feat(shortcuts): add "?" keyboard shortcut cheat sheet

### LINKED ISSUE #8816: 💡 Implement the "?" shortcut cheat-sheet overlay (default binding exists, handler missing)
## Problem

The default keyboard config ships `showHelp: '?'` (`src/app/features/config/default-global-config.const.ts:148`), but there is **no handler anywhere** — `showHelp` appears exactly twice in `src`: the config default and a commented-out form field (`src/app/features/config/form-cfgs/keyboard-form.const.ts:60`). Pressing `?` does nothing.

Meanwhile the app has ~53 keyboard shortcuts (43 bound by default) whose only discovery surface is the settings form under a collapsible.

## Suggested fix

Implement `?` as a shortcut cheat-sheet overlay (read the current `keyboard` config so custom bindings are shown correctly). This is the single cheapest power-user win available — the binding, the translation surface, and the config plumbing already exist.

If the overlay is not wanted, the dead `showHelp` binding should be removed instead.

---

MarkQuach12:
## Problem

Closes #8816. The default config ships `showHelp: '?'`, but pressing `?` did nothing. Nothing handled it, and it could not work as written: `checkKeyCombo()` matches on `ev.code`, which reports `?` as `Slash` + Shift, so it never equals `'?'`.

## Solution

- Handle `showHelp` in `ShortcutService` by matching on `ev.key` instead of `ev.code`, so it also works on non-QWERTY layouts.
- Add `DialogKeyboardShortcutsComponent`, a dialog listing the current shortcuts. Rows come from `KEYBOARD_SETTINGS_FORM_CFG`, so it shows custom bindings and picks up new shortcuts on its own.
- Document the shortcut in the wiki keyboard shortcuts reference.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for tackling this—the feature is useful, and the normal `?` flow works. I rechecked the current head (`18bfd3e`) against the full diff/history, the deployed preview, current discussion, and CI (all relevant checks are green). Verdict: **needs changes before merge**.

1. **Honor the persisted `keyboard.showHelp` contract and the layout-independence claim.** [`isHelpKeyCombo()`](https://github.com/super-productivity/super-productivity/blob/18bfd3ec7766dfe237fe54e4ec5de7d40259f00d/src/app/core-ui/shortcut/shortcut.service.ts#L31-L37) hard-codes `ev.key === '?'` and never reads `keys.showHelp`. `showHelp` is still nullable persisted config, stored overrides are preserved during hydration, shipped v2 settings exposed the binding, and the new wiki calls it configurable. Any surviving/imported custom or `null` value is therefore ignored. The matcher also rejects all Ctrl/Alt states, which blocks layouts where `?` is produced through AltGraph (often represented as Ctrl+Alt), despite the stated non-QWERTY support. Please make the helper consume `keys.showHelp`: honor `null`, preserve a character-based special case for literal `?` (including AltGraph), and delegate ordinary custom combinations to `checkKeyCombo`. Also expose/include `showHelp` in the settings/catalog so the new shortcut appears in its own sheet. Tests should cover default, null, custom, and AltGraph cases.

2. **Prevent duplicate dialogs while the lazy chunk is loading.** The guard at [`shortcut.service.ts:192-197`](https://github.com/super-productivity/super-productivity/blob/18bfd3ec7766dfe237fe54e4ec5de7d40259f00d/src/app/core-ui/shortcut/shortcut.service.ts#L192-L197) is checked only before `await import(...)`. On a cold throttled load I reproduced three stacked `Keyboard Shortcuts` dialogs from three presses before the chunk resolved. The adjacent create-project path already rechecks `openDialogs` after its import; the same post-import check is sufficient here. Please add a regression test that starts multiple `handleKeyDown('?')` promises without awaiting between them and expects one `open()` call.

3. **Keep “Tasks” as the task-section heading.** [`rows`](https://github.com/super-productivity/super-productivity/blob/18bfd3ec7766dfe237fe54e4ec5de7d40259f00d/src/app/core-ui/shortcut/dialog-keyboard-shortcuts/dialog-keyboard-shortcuts.component.ts#L48-L60) treats every Formly `tpl` item as a heading. Because the real `Tasks` `<h3>` is immediately followed by an informational `<p>`, the filter removes `Tasks` and promotes the full explanatory sentence to the bold group heading. I reproduced this in the preview. Only heading templates (for example `tag === 'h3'`) should create group headings; handle or intentionally omit the paragraph separately, and add a row-generation/component test.

4. **Resolve the catalog-scope mismatch.** The dialog reads the static `KEYBOARD_SETTINGS_FORM_CFG`, but plugin shortcut rows are appended only to the config page’s runtime copy. Registered, configurable, working plugin shortcuts therefore never appear, contrary to the PR’s “current shortcuts” / “picks up new shortcuts” description. Either include the runtime plugin entries and test one, or explicitly narrow the UI/docs promise to built-in shortcuts.

One smaller accessibility/responsiveness issue: the new [table](https://github.com/super-productivity/super-productivity/blob/18bfd3ec7766dfe237fe54e4ec5de7d40259f00d/src/app/core-ui/shortcut/dialog-keyboard-shortcuts/dialog-keyboard-shortcuts.component.html#L4-L20) has no Action/Shortcut column headers or cell associations. At supported 320–400 px window widths it also requires horizontal scrolling, but the scroll container is not keyboard-focusable. The existing `dialog-markdown-shortcuts` table/responsive styling looks like a good established pattern to reuse.

johannesjo:
Thanks for the quick turnaround — I re-reviewed `18bfd3e..4a07816`. **All four blocking points and the accessibility note are addressed.** I verified each fix the hard way: reverting it locally and confirming the matching new test goes red, so none of the added tests are vacuous.

| My point | Status | Sabotage check (revert the fix → what fails) |
| --- | --- | --- |
| 1. Honor `keyboard.showHelp` (null / custom) | Fixed | 2 failures: `…when showHelp is unbound`, `…for a custom showHelp combo` |
| 1b. AltGraph / layout independence | Fixed | `…for "?" produced via AltGr` fails |
| 2. Duplicate dialogs during lazy load | Fixed | `Expected spy open to have been called once. It was called 3 times.` |
| 3. `Tasks` heading vs. info paragraph | Fixed | `Expected $[0].heading = 'GCF.KEYBOARD.TASK_SHORTCUTS_INFO' to equal 'GCF.KEYBOARD.TASK_SHORTCUTS'` |
| 4. Plugin catalog scope | Fixed | plugin heading + row disappear |
| 5. a11y / responsive | Mostly fixed | see items 2–4 below |

Point 4 is done properly — the dialog now calls the same `createPluginShortcutFormItems` factory with the same `PluginBridgeService.shortcuts` signal the config page uses, and the `plugin_<id>:<action>` key format matches what is persisted, so the two can't drift.

Also checked green: production AOT build (`ng build --configuration production`), `npm run checkFile` on all six changed files, `npm run int:test`, and `t.const.ts` is byte-identical to `npm run int` output. No sync/schema risk — `showHelp?: string | null` already existed in `KeyboardConfig` and in `DEFAULT_GLOBAL_CONFIG`, `migrateKeyboardConfig` spreads those defaults under stored values, and `frozen-state.spec.ts` stays green.

Remaining items, none blocking, roughly in priority order:

**1. The new SCSS is a copy of the markdown dialog's, not a reuse of it.** `dialog-keyboard-shortcuts.component.scss` is **76 of its 81 lines byte-identical** to `src/app/ui/dialog-fullscreen-markdown/dialog-markdown-shortcuts.component.scss` (+5/−3), down to all five class names and the `@media (max-width: 25rem)` block — it even inherits that file's `.keyboard-shortcuts-dialog-content` misnomer. The adaptations that *are* there are sensible (dropping `.shortcut-key`'s `margin` since there's one `<kbd>` per cell, and `<kbd>` is better markup than the original's `<span>`). Emulated encapsulation means the duplicate class names can't collide, so this is DRY debt rather than a bug — but it's now the second copy, which is the "make it reusable or add it to the shared style layer" case. The same duplication shows up in i18n: the new `GCF.KEYBOARD.COLUMN_ACTION` / `COLUMN_SHORTCUT` duplicate the existing `F.NOTE.D_FULLSCREEN.SHORTCUTS.COLUMN_ACTION` / `COLUMN_SHORTCUT`.

**2. Group headings render almost identically to the column-header row.** `.shortcuts-table th` sets `text-transform: uppercase`, `font-size: var(--font-size-sm)` and `border-bottom: 2px solid`, and `.group-heading` only overrides `color` and `padding-top`. So "Global Shortcuts (application-wide)" / "Tasks" / "Plugin Shortcuts" come out uppercase with the same 2px rule and the same size as the "ACTION"/"SHORTCUT" header — the group dividers barely read as a different level. Worth giving `.group-heading` its own `text-transform`/size/border treatment.

**3. `scope="colgroup"` on the divider `<th>` is a no-op.** `colgroup` scopes to a `<colgroup>` element and the table has none, so the association the attribute is meant to create doesn't exist. Note `scope="rowgroup"` isn't a drop-in fix either: there's a single `<tbody>` wrapping every section, so a rowgroup-scoped divider would claim the rest of the whole table. Strictly correct is one `<tbody>` per section with `scope="rowgroup"` on each divider. Real-world AT support for both values is patchy, so treat this as a spec nit — the `scope="col"` headers in `<thead>` are what actually carry the association you asked for.

**4. The focusable scroll container has no name or role.** `tabindex="0"` on `<mat-dialog-content>` is the standard WCAG 2.1.1 remedy for a keyboard-inaccessible scroll region, so I read it as intentional. Side effect worth deciding on: it becomes the dialog's first tabbable element, so Material's default `autoFocus: 'first-tabbable'` now lands initial focus on an unnamed `<div>` rather than the Close button. `role="region"` + an `aria-label` fixes the naming; add `cdkFocusInitial` to the Close button if you'd rather keep focus there.

**5. `Ctrl`+`Alt`+`?` can open the cheat sheet where no AltGr is involved.** The `(ev.ctrlKey && ev.altKey)` surrogate can't distinguish a synthesized AltGr from a deliberate Ctrl+Alt chord. I confirmed the code path with a probe test and confirmed nothing swallows it earlier — no default binding claims `Ctrl+Alt+Shift+/`, and `checkKeyCombo` never matches `Slash` for the defaults. I did *not* verify how each OS actually reports modifiers for that chord, so the real-world blast radius is platform-dependent and small (an unexpected dialog plus a `preventDefault()` on a chord the app otherwise ignores). Since current Chromium/Firefox already report `getModifierState('AltGraph')` for AltGr, dropping the `ctrlKey && altKey` half may buy correctness for free — your call.

**6. An existing assertion got narrower than it needed to be.** `should NOT open the shortcut cheat sheet when a modifier is held` was switched from `ctrlKey: true` to `metaKey: true`. That swap wasn't required — I verified ctrl-only `?` is still rejected by the new matcher — and ctrl-without-alt is now precisely the boundary the new AltGr surrogate moved, so it's the more valuable case to keep. Please assert both.

**7. Rebinding `showHelp` is effectively a one-way door out of the layout-independent `?`.** `keyboard-input.component.ts` composes values from `ev.code`, so its keydown path emits `Shift+/` or `Shift+Slash` but never a bare `?`. (Strictly it isn't unrecoverable — the field is a plain `matInput`, so pasting `?` in bypasses `keydown` and persists — but there's no discoverable path.) This matters a bit more than usual because `globalConfig.keyboard` syncs across devices: rebinding on one layout can silently break the shortcut on another. The wiki explains the `?`-default behaviour but not this asymmetry — worth a clause there, since it's the one way this shortcut is *not* "like every other shortcut".

**8. Minor, and inherent to exposing the field:** `isHelpKeyCombo` sits mid-chain, ahead of `addNewNote` / `openProjectNotes` / `toggleIssuePanel` / `triggerSync`. Now that `showHelp` is user-bindable, setting it to e.g. `Alt+N` silently shadows "Add new note". That's the pre-existing behaviour for every shortcut (there's no conflict detection in the form), just one more field exposed to it — noting it rather than asking for anything.

One process note: **CI has not run on `4a07816` at all** (zero check runs; the green run you saw was on `18bfd3e`), and the branch is 38 commits behind `master`. It merges cleanly and every suite I ran locally passes, but it'd be worth getting a real CI run on this head before merging.

johannesjo:
Thank you very much! <3

PR #9304, opened 2026-07-25
fix(styles): replace phantom CSS variable references and guard against regressions

johannesjo:
Component and global SCSS referenced ~50 CSS custom properties that are declared
nowhere — not in `src/styles/`, not in `src/assets/themes/*.css`, not in any
component SCSS, and not injected at runtime. An undefined custom property
invalidates the whole declaration at computed-value time, so the property
silently falls back to inherit/unset (`gap: var(--s1)` collapsed to 0 in focus
mode); the sites that carried a hardcoded fallback rendered but were permanently
theme-blind. No build error, no lint error, no console warning.

This maps every reference onto the semantically equivalent existing token, drops
the hardcoded fallbacks, and adds a lint gate so the class cannot come back.

## Token mapping

- **spacing:** `--s1` → `--s`
- **semantic color:** `--warn-color`/`--success-color` and the `#f44336`/`#ff9800`/`#4caf50`
  fallbacks → `--color-danger`/`--color-warning`/`--color-success`;
  `--palette-green-500`/`--palette-red-500`/`--palette-red-700` (no such palettes exist)
  → `--color-success`/`--color-danger`
- **text:** `--text-color-secondary`, `--text-tertiary`, `--text-muted`,
  `--text-color-dimmed`, `--fg-secondary`, `--c-text-secondary`,
  `--palette-text-secondary`, `--color-fg-less-important` → `--text-color-muted`;
  `--fg-primary` → `--text-color`
- **brand:** `--primary`, `--primary-color`, `--palette-primary-main` → `--c-primary`;
  `--color-accent`, `--palette-accent` → `--c-accent`; `--palette-warn` → `--c-warn`;
  `--c-primary-dark` → `--palette-primary-700`; `--text-on-accent`, `--c-on-accent`,
  `--text-on-primary` → the matching `--palette-*-contrast-500`
- **borders/surfaces:** `--border-color`, `--border-light`, `--border-dark`, `--c-border`,
  `--color-border`, `--standard-border-color` → `--extra-border-color` (the input in
  select-task-minimal uses `--input-border-color`); `--bg-color`, `--c-bg-card` →
  `--card-bg`; `--c-bg`, `--background-dark` → `--bg`; `--background-color-lighter`,
  `--bg-color-lighter` → `--bg-lighter`
- **state overlays:** `--hover-color`, `--palette-background-hover` and the misused
  `--mdc-dialog-subhead-color` (a text token used as a background) → `--state-hover`
- **rgb channels:** `rgba(var(--c-accent-rgb), a)` / `rgba(var(--c-primary-rgb), a)` →
  `color-mix(in srgb, … N%, transparent)`; `rgba(var(--text-color-rgb, 0,0,0), a)` →
  `rgba(var(--ink-on-channel), a)`
- **misc:** `--whiteframe-shadow-20dp` (the scale has no 20dp step) → 24dp for the drag
  elevation; `--monospace-font` → `--font-mono-stack`; `--z-done-toggle` fallback chain
  collapsed to `--z-drag-handle`; `--standard-border-radius` (declared only inside
  schedule-event's component scope) → `--card-border-radius`
- **app.component:** the side-nav `border-right-color` referenced an undeclared
  `--sidebar-border-color`, which is IACVT on a non-inherited property and so computed to
  `currentColor` — a visible hairline in the text colour. Now `transparent`, matching the
  "remove right border" intent stated in the comment directly above it.

Runtime-injected names are left untouched: `--palette-*`/`--mat-*`/`--mdc-*`,
`--safe-area-inset-*` and `--android-webview-icon-scale` set from TypeScript, and
`--selected-bg-opacity`, an optional theme-override hook consumed with a fallback.

## The guard — `npm run lint:css-vars`

`tools/check-css-vars.js` fails the build on any `var(--x)` that cannot resolve where it
is written. It scans `.scss`, theme `.css`, **and** the inline `styles:` blocks,
`[ngStyle]` and `[style.--x]` bindings in `.ts`/`.html` — stylelint cannot see those, and
four phantoms were hiding there (`--hover-color` in three components,
`--primary-color` in the nav tree's drag preview).

Two ways a reference fails:

- **undefined** — declared in no file at all.
- **`(component-scoped)`** — declared, but only under `src/app/**`, while the reference
  sits in a global stylesheet. Custom properties inherit only downwards, so a global rule
  never sees a property scoped to some component's `:host`. That is how
  `--standard-border-radius` stayed broken. This check is a heuristic — a global rule
  *can* legitimately match inside a component subtree — so `ALLOWLIST` is the escape
  hatch, and the Sass mixin libraries under `src/styles/` are excluded, since their bodies
  land in whichever component `@use`s them.

`tools/check-css-vars.test.js` covers both classes plus the parsing edge cases (comment
stripping, `$sass--vars`, palette-name shapes, `.spec.ts` exclusion). Runtime ~0.2s over
1807 files.

## Also here

- **`.callout`** (`src/styles/components/_callout.scss`) replaces four per-dialog
  `.warning-box`/`.info-box`/`.warning` clones that each hardcoded their own
  `rgba(255, 152, 0, …)` and so ignored all 15 themes. One tone token drives icon, border
  and fill. Documented in `docs/styling-guide.md`, including the flex markup contract —
  the class needs a single child element, so `<p class="callout">` with bare text renders
  `<strong>Note:</strong>` as its own column.
- **Font-size token adoption** in the side nav and settings.
- **`aria-label`s** on the icon-only task context menu quick-access buttons — `matTooltip`
  sets `aria-describedby`, which does not supply an accessible name.

## Not in this PR any more

The mobile bottom-nav safe-area change, the onboarding-hint reposition and the base
light/dark surface-hierarchy retune have been dropped from this branch. The first two need
verification on a real device, which CI cannot give; they will come back as their own PR.
The surface retune was carried alongside its own partial revert, which made it ~180 lines
of review noise for a net of four token tweaks.

## Verification

`npm run lint` (including the new gate and tool tests) and `ng build` both pass. The gate
was sabotage-tested against a planted phantom of each class — undefined, inline-`.ts`, and
component-scoped — and fails the build on each.


johannesjo:
Reviewed `8df1fb5..c420d55` (the revert + the review-fixes commit), with the rest of the branch as context. All 21 checks green.

**The revert is clean.** I diffed `c420d55` against `4227c30d`: `--surface-0/1/2`, the dark task rungs, `--bg-drag-col` and `--sidenav-bg` are byte-identical to master again. What deliberately survives is the hairline/ink set (`--ink-muted` .66→.70 / .65→.68, `--separator`, `--divider`, `--extra-border-color`), which reads as the right call — with the flat canvas back, those hairlines are carrying *more* of the separation, not less.

**The `--c-contrast` fix is correct and complete.** `--c-contrast` resolves to `--palette-primary-contrast-500`, so those two badges were painting primary-contrast text on an accent fill. `--palette-accent-contrast-500` is maintained at runtime by `MaterialCssVarsService` (`contrastColorThresholdAccent`) and overridden by the four themes that recolor accent. I checked all 9 remaining `--c-contrast` sites — every one sits on a `--c-primary` background, so nothing was left half-fixed. Using the raw palette token is also consistent with `themes.scss` and `schedule-week.component.scss:395`, so no new semantic token is warranted here.

---

## Important — the guard is `.scss`-only, and four phantom references survive in inline styles

`tools/check-css-vars.js:142` walks only `src/**/*.scss`. There are 121 `var(--…)` occurrences in `.ts` inline `styles:` blocks and templates, and two of the names this PR's description lists as eliminated are still live there:

```
--hover-color    desktop-panel-buttons.component.ts:96
                 page-title.component.ts:223
                 plugin-side-panel-btns.component.ts:89
--primary-color  nav-list-tree.component.html:128
```

All four carry the theme-blind hardcoded fallback the tool's own docstring calls out. `var(--hover-color, rgba(0, 0, 0, 0.04))` paints a black wash for hover on every theme including dark, and the tree drag preview is pinned to `#3f51b5`. Neither name is declared in any `.scss`/theme `.css`, and neither is `setProperty`-ed from TS. (There *is* a `--primary-color` in `packages/plugin-dev/boilerplate-solid-js/src/app/App.css`, but that's a separate boilerplate and never loaded into the app.)

These are pre-existing on master, not regressions — but they're the exact class this PR exists to remove, and the new lint is shaped so it will never catch them.

Fixing the four sites is the required part. Extending the walk keeps them fixed, and it's ~5 lines:

```js
const scssFiles = walk(SRC_DIR, '.scss');
// Inline component styles (`styles:` / [ngStyle]) reference tokens too.
const inlineFiles = [...walk(SRC_DIR, '.ts'), ...walk(SRC_DIR, '.html')].filter(
  (f) => !f.endsWith('.spec.ts'),
);
// …
for (const file of [...scssFiles, ...inlineFiles]) {
```

I ran exactly that against the branch: **4 hits, zero false positives**, and it still fails on a freshly planted TS-side phantom. `stripComments` already kills the only comment-borne match (`global-theme.service.ts:972`), and the `.spec.ts` skip drops the theme-validation fixture. Theme `.css` files are already clean, so there's no reason to add them.

## Minor

**`tools/check-css-vars.js:75` — one regex replaces the 30-line enumeration and closes a gap for free.**

```js
const PALETTE_TOKEN =
  /^--palette-(primary|accent|warn)-(contrast-)?(50|[1-9]00|A[1247]00)(-rgb|-no-rgb)?$/;
```

I verified it matches all 168 names the current IIFE generates, and still rejects every invented name (`--palette-green-500`, `--palette-red-700`, `--palette-primary-main`, `--palette-accent`, `--palette-warn`, `--palette-text-secondary`) — so the narrowing this commit is *for* is fully preserved. It additionally accepts `--palette-*-no-rgb`, which the library really does emit (`_variables.scss` defines the whole `-no-rgb` family, and the contrast entries are written in terms of it) and which the enumeration currently rejects. Not used today, so it's a latent false positive rather than a bug — but it costs nothing to cover in the same expression.

Worth sabotage-noting: the narrowing genuinely works. Under the old `--palette-` prefix both `--palette-green-500` and `--palette-primary-main` passed; under either the enumeration or the regex they fail. Good change.

**`dialog-disable-profiles-confirmation.component.ts:70` — the markup doesn't match the callout contract.** `_callout.scss`'s own docblock specifies `<div class="callout"><mat-icon/><p>…</p></div>`, but this is `<p class="callout">` with `<strong>Note:</strong>` and the body copy as bare siblings. Per flexbox, `<strong>` becomes its own flex item and the text run another, so you get three columns with an 8px gap after "Note:" and a hanging indent on the wrap. I rendered both — the difference is small but real:

```html
<div class="note callout callout--info">
  <mat-icon aria-hidden="true">info</mat-icon>
  <p><strong>Note:</strong> No data will be deleted. …</p>
</div>
```

The other two `<p class="callout">` sites in this branch are fine only because they have no inline element before the text.

**`dialog-disable-profiles-confirmation.component.ts:107,128` — the same file still has two theme-blind backgrounds.** `.info-section/.profiles-info/.recovery-info` use `background: rgba(0, 0, 0, 0.05)` and `code` uses `rgba(0, 0, 0, 0.1)`. Over the dark card (`--surface-2` → `--dark3`, `rgb(36,36,36)`) that computes to `rgb(34,34,34)` — 1.03:1, i.e. the boxes vanish, and what little shows goes the wrong direction (recessed rather than raised). Since this commit went into the file specifically to de-hardcode a color, `--state-hover` / `--bg-lighter` are the obvious neighbours.

---

## Verdict

**Needs changes** — the Important item only. The four inline-style phantoms are the same names the description claims are gone, and the guard as written can't see them. The three Minors are small and could land separately.

<sub>Verified: net diff of every reverted token against `4227c30d`; `--hover-color`/`--primary-color` undeclared repo-wide and not `setProperty`-ed; the checker sabotage-tested with 6 token cases plus a planted TS phantom; the regex proven a superset of all 168 enumerated names; the library's emission path in `_main.scss`/`_variables.scss`; the callout layout rendered in Chromium. Also checked and found fine: `#f57c00` → `var(--color-warning)` matches the existing `color: var(--color-warning)` convention at 5 other sites, so I'm not suggesting a change there. Not verified: how the retained hairline/border deltas look in the running app across the built-in themes.</sub>

PR #9306, opened 2026-07-25
feat(tracking): show the tracked task in the browser tab title

### LINKED ISSUE #8818: 💡 Make a running timer visible everywhere (chip hidden below 1080px, no tab-title change, dash placeholders)
## Problem

A running timer can be completely invisible:

1. The tracked-task chip in the header is hard-hidden below 1080px viewport width (`src/app/core-ui/main-header/play-button/play-button.component.ts:187-189`) — only a generic pulsing pause icon remains.
2. The browser tab title only changes for **focus-mode** sessions, not plain tracking (`src/app/core-ui/browser-title.service.ts` — `_getTitle` returns the base title unless a focus session is running). A web/PWA user who switches tabs has no cue a timer is running.
3. The work-view header metrics render bare dash placeholders when empty (`Estimate remaining: ~-`, `Working today: -`, `- / 8h` in the planner) — they read as broken UI rather than "no data yet".

On plain web (no Electron, no extension) there is also no idle detection, so a forgotten timer runs indefinitely with none of the above cues.

## Suggested fix

- Show at least elapsed time + pause affordance at all viewport widths (the chip can truncate, but not disappear).
- Include tracking state in the tab title for plain tracking too (e.g. `⏱ 0:42 · Task name`).
- Hide the metric row items (or show a friendly zero-state) instead of rendering `-` placeholders.

---

MarkQuach12:
## Problem

Closes #8818. A running timer could be invisible. The tab title only changed for focus sessions, so a web user who switched tabs had no cue at all, and plain web has no idle detection to catch a forgotten timer. The work view metric row rendered a bare `-` for zero values, which reads as broken UI.

## Solution

Show the tracked task in the tab title during plain tracking, not just focus mode, as `25m - Task name`. Time shown is the task's total for today, since no timestamp for when tracking started is persisted. Focus mode keeps priority when a session is active.

Render `0m` instead of `-` for the three work view header metrics, matching what the planner already does.

Documents the tab title in `3.05-Web-App-vs-Desktop.md`. Also adds one clause to `4.14-How-Time-Is-Logged.md`, a Concepts note, so the line listing where the active task is shown mentions the tab title too.

## Scope change after rebase

This PR originally also reworked the header chip so the tracked task name stayed visible at all widths. That part is dropped. #9517 landed the same capability upstream as `tracked-task-pill`, and it is a superset of what this PR did, so the commit was removed rather than rebased. Both review items on the earlier revision were about that commit and no longer apply.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
@MarkQuach12 thank you! Could you have a look at the conflict?

MarkQuach12:
> @MarkQuach12 thank you! Could you have a look at the conflict?

Took a quick look at it was just an additional missing line (merged and should be good now)

src/app/core-ui/main-header/main-header.component.scss
```diff
@@ -173,11 +178,11 @@ button.isActive2 {
   margin-left: auto;
   display: flex;
   align-items: center;
-  // Keep the action group at its intrinsic width so a long title shrinks
-  // (it ellipsizes) instead of pushing these buttons off screen (#7477).
-  flex: 0 0 auto;
+  // Shrinkable only as far as the tracked-task chip can give way - every
+  // button inside is fixed-size, so they still can't be pushed off screen
+  // (#7477) while the page title keeps its share of the row.
+  flex: 0 1 auto;
```
johannesjo, line 262: **Important — fixed controls overflow while tracking.** Making the whole action nav shrinkable does not establish a shrink path for the nested chip: the nav retains its automatic min-content width while the chip has a 90px floor. On this head I reproduced Focus and Sync outside a 440px viewport, Pause partly outside at 321px, and trailing actions outside an 800px viewport with the side nav open. Pausing removes the chip and restores containment.

This reverses the load-bearing fixed-nav behavior from #7898/#7477. Please keep the fixed actions non-shrinking and establish a tested shrink path where only the optional chip truncates (with a fit fallback if it cannot fit). The smallest regression test is to extend the existing #7477 layout fixture with a running-task chip; it currently renders counters only, so it passes this failure.

src/app/core-ui/main-header/play-button/play-button.component.ts
```diff
@@ -165,30 +174,25 @@ import { NavigateToTaskService } from '../../navigate-to-task/navigate-to-task.s
         }
       }
 
+      /* In flow rather than an overlay, so the header's flex layout keeps this
+         off the page title instead of covering it. Shrinks ahead of everything
+         else in the header, down to the floor below. */
       .current-task-title {
```
johannesjo: **Important — the vertical action bar loses Pause while tracking.** The unchanged `_vertical-action-bar.scss` explicitly relies on `.play-btn-wrapper` being the positioned containing block and this chip being absolute/vertically centered. After this change, the vertical override only hides the in-flow chip with opacity/visibility, so it still reserves horizontal space. At a 1000px viewport I reproduced the 48px rail at x=952–1000 while Pause was entirely offscreen at x=1096–1136.

The smallest fix is local to the existing vertical-mode override: restore a positioned wrapper and an absolute, vertically centered flyout there so the hidden title consumes no rail width. No new layout abstraction is needed.

johannesjo (changes requested):
I carefully rechecked the latest head against #8818, the prior #7477 layout contract, both header modes, and the focused tests.

The browser-title and `0m` changes look sound. I’m requesting changes for the two reproducible header regressions in the inline comments. Both can be fixed locally in the existing layout; no new state, setting, or abstraction is needed.

Verification on this head: BrowserTitleService 21/21 and MainHeader 5/5 pass locally. The existing header fixture does not include the running-task chip, which is why it misses the reproduced overflow.

MarkQuach12:
Rebased onto master and force pushed, so the history changed. Summary of what moved:

The header chip commit (`fcd311c83`) is dropped rather than fixed. #9517 landed `tracked-task-pill` upstream, which does the same job and more (click to navigate, project tag, container query, absolutely positioned so it cannot move a button). Both of your review items were about that commit, so neither applies now. Rebasing it would have meant reimplementing a worse version of something already merged.

What remains is the two changes you called sound, plus their docs:

- `feat(tracking): show tracked task in the browser tab title`
- `fix(work-view): show 0m instead of a dash in header metrics`
- `docs(wiki): note the tracked task in the tab title`

Title and description updated to match.

Verified on the rebased head against the widths from your review, with a task tracking and a long name:

| viewport | pill rendered | actions outside header |
| --- | --- | --- |
| 1280 | yes | none |
| 800 | yes | none |
| 440 | yes | none |
| 321 | yes | none |

Tab title round trip: `Super Productivity` when idle, `0m - <task>` while tracking, back to `Super Productivity` on stop.

Unit suite on the rebased head is green apart from `caldav-client.service.spec.ts` "connects when the server omits principal-collection-set from the principal response", which fails identically on a clean master checkout here, so it is not from this branch.

johannesjo:
**✅ Ready to merge** – no blocking issues at `35f7b6a`. Merge confidence 90% – the per-second title refresh is verified by reading the wiring (`select(selectCurrentTask)` re-emits because `addTimeSpent` produces a new entity each tick), not by running it: the new specs mock `TaskService`, so nothing executes that path.

Thanks for the PR, and for the rebase write-up – dropping the header commit in favour of `tracked-task-pill` from #9517 was the right call, and I confirmed that both earlier inline comments pointed at files this head no longer touches. What remains is small, correct and covered by the new specs: the focus-session precedence, the `0m` case and the fall-back-once-focus-ends transition all have a test, and the last one is exactly the case that would break if the conditional signal read in the effect stopped re-tracking.

Two optional polish items:

### Nits

- **`docs/wiki/4.14-How-Time-Is-Logged.md:11`** and **`docs/wiki/3.05-Web-App-vs-Desktop.md:25`** – both new clauses say the tracked task's name appears in the tab/window title, without qualification, but a focus session wins: `browser-title.service.ts:39-41` uses `focusTitle` and never reaches `_getTrackingTitle()` while `isRunning` or `isSessionPaused` is true. So with a task tracked and a focus session running the title is just `25:00`, with no task name. A "when no focus session is running" qualifier in both lines would make the docs match the code.

- **`src/app/features/work-view/work-view.component.html:23`** – pre-existing rather than something this PR introduced, so entirely up to you: the `TODAY_REMAINING` tooltip still uses the bare `msToString`, so in a project context where the project has no remaining estimate but `workingToday()` is non-zero the row renders and the tooltip reads `Today remaining ~-` – the same dash the three visible metrics were just fixed to avoid. The other two bare uses in that row (`:49`, `:61`) can't produce a dash, because the `@if` and `matTooltipDisabled` guards suppress the zero case, so this is the last one. Same `(… | msToString: false : true) || '0m'` finishes it.

---

johannesjo:
Thank you very much! <3

PR #9312, opened 2026-07-26
fix: preserve shortcuts when Escape cancels input

### LINKED ISSUE #9269: Keyboard shortcut settings clear the binding when Escape is pressed
## Problem

The Keyboard Shortcuts help tells users:

> Press Escape to abort.

But `KeyboardInputComponent.onKeyDown()` handles Escape by setting the form value to `null` and blurring the field. Formly emits that model change immediately, so the existing shortcut is cleared and persisted. The same help text later contradicts itself by saying Escape, Backspace, or Delete removes a shortcut.

This makes a conventional cancel key destructive and leaves the intended behavior unclear.

## Reproduction

1. Open Settings → Keyboard Shortcuts.
2. Focus an existing shortcut, such as **Add new task** (`Shift+A`).
3. Press Escape.
4. Reload Settings.

### Actual

The shortcut is empty after reload.

### Expected

Choose and document one unambiguous contract. Recommended:

- Escape cancels/leaves the existing value unchanged and blurs the field.
- Backspace and Delete remain the explicit ways to clear an assigned shortcut.

## Acceptance criteria

- Make the help text and input behavior agree.
- Add a focused `KeyboardInputComponent` unit test for Escape with an existing value.
- Add or extend the settings E2E to prove the chosen Escape behavior survives reload.
- Preserve assignment of modified Backspace/Delete combinations.

## Relevant files

- `src/app/features/config/keyboard-input/keyboard-input.component.ts`
- `src/app/features/config/keyboard-input/keyboard-input.component.spec.ts`
- `src/assets/i18n/en.json`
- `e2e/tests/keyboard/keyboard-shortcuts.spec.ts`

Discovered while implementing #9262 user-remapped shortcut persistence.

lntutor:
## Summary
- retain the existing shortcut when Escape cancels keyboard shortcut editing
- keep Backspace and Delete as the explicit shortcut-clear actions
- add component and reload-persistence browser coverage, plus matching English help and wiki documentation

Fixes #9269

## Verification
- `npm run test:file src/app/features/config/keyboard-input/keyboard-input.component.spec.ts` (6 passed)
- `npm run checkFile src/app/features/config/keyboard-input/keyboard-input.component.ts`
- `npm run checkFile e2e/tests/keyboard/keyboard-shortcuts.spec.ts`
- `npm run e2e:file e2e/tests/keyboard/keyboard-shortcuts.spec.ts -- --retries=0 --grep 'keep an assigned shortcut after Escape and reload'`\n- `git diff --check`\n\nThe repository-wide lint command reports 10 existing warnings in unrelated files and exits non-zero; this change adds none.


johannesjo (commented):
Reviewed `39a0ec6`: checked the branch out, ran both new tests clean, sabotage-tested them, and probed the real behaviour in a browser.

**The production change is right and I'd merge it as-is.** Deleting the `setValue(null)` is the correct minimal fix, and it finishes what #8142 started in June when it added Backspace/Delete as the explicit clear keys. Both new tests are non-vacuous — restoring the deleted line fails the unit spec (`Expected null to be 'Shift+A'`, 1 of 6, and only that one) and fails the E2E on `toHaveValue`.

One thing to fix before merge, and it's wording only.

## The help text promises a "cancel" that doesn't exist

`en.json` now says:

> Click on the text input and press the desired keyboard key or key combination. Press **Escape** to cancel without changing the current shortcut.

and the wiki (173 / 194):

> Press **Escape** to cancel editing and keep the existing shortcut …
> **Escape** cancels editing, while …

But Escape only blurs — it never restores the pre-edit value. `config-form` saves on every `(modelChange)` with no debounce (`config-form.component.ts:38-51` → `config-page.component.ts:325` `saveGlobalCfg`), so the replacement is committed and persisted the instant the key is pressed.

Verified in a browser — focus **Add new task** (`Shift+A`) → press `Shift+V` → press `Escape` → reload:

```
Locator:  getByRole('textbox', { name: 'Add new task', exact: true })
Expected: "Shift+A"
Received: "Shift+V"
```

So "cancel" is false in the one situation a user reaches for it: they fat-fingered a key and want out. That is the same help-text-≠-behaviour defect #9269 was filed to remove, and its first acceptance criterion is *"Make the help text and input behavior agree."* (In fairness, the issue's own phrasing — "Escape cancels/leaves the existing value unchanged" — invited this reading.)

**Suggested fix, wording only:**

- `en.json`: "Press **Escape** to leave the field without assigning a new shortcut."
- wiki 173: "- Press **Escape** to leave the field, or **Backspace** or **Delete** to clear an existing binding."
- wiki 194: "**Escape** leaves the field, while **Backspace** or **Delete** clears an existing assignment."

**Please don't implement a real cancel.** It is only ~5 lines (snapshot the value on `(focus)`, restore it in the Escape branch), but nobody asked for undo-on-Escape — the reported bug was the *destruction*, and that is already fixed. Matching the words to the behaviour is the smaller and more honest change.

## Pre-existing, one line above — your call

Wiki line 172 is a context line in this PR's own hunk, untouched:

> Change any shortcut by clicking the input and pressing the desired key or combination; press **Enter** to save.

There is no save step, and `Enter` is not special-cased in `onKeyDown` — `prepareKeyCode('Enter')` returns `'Enter'`, so **pressing Enter binds Enter as the shortcut**. Verified in a browser:

```
Expected: "Shift+A"
Received: "Enter"
```

Not this PR's bug, so entirely optional — but since the PR is making that exact three-bullet list agree with reality, it seems like the moment to drop the `; press **Enter** to save` clause.

## Nits — take or leave

- `keyboard-shortcuts.spec.ts:20` — `page.locator('keyboard-input input').first()` asserts against whatever the first field happens to be. It works today (web → `addNewTask`), but the sibling test 20 lines below already does it robustly: `getByRole('textbox', { name: 'Add new task', exact: true })` asserting `'Shift+A'`. Reusing that would also make the `not.toBe('')` guard unnecessary.
- `keyboard-shortcuts.spec.ts:29` — the sibling test calls `waitForStatePersistence(page)` before `reload()`; this one doesn't. Nothing is written post-fix so it passes either way — purely consistency.

Branch is 2 commits behind master, but neither touches these files, so no rebase needed.

lntutor:
Addressed the wording issue in 89eb8e46c. The UI and wiki now describe Escape as leaving the field without assigning a new shortcut, not as undoing an already assigned replacement. I also removed the pre-existing claim that Enter saves, since Enter is itself assignable. Focused unit tests pass (6/6), the targeted Playwright scenario passes, and all modified TypeScript files pass checkFile.

johannesjo:
Thank you very much! <3

PR #9313, opened 2026-07-26
fix: escape worklog CSV export fields

### LINKED ISSUE #9267: Worklog CSV export does not escape fields and hides its task separator option
## Problem

While adding the corrected-duration export regression from #9262, two related worklog-export defects were confirmed:

1. `formatText()` joins columns with `;` and rows with a newline without escaping field values. Task titles, notes, projects, and tags may contain semicolons, quotes, or line breaks, so otherwise valid user data can shift columns or create extra CSV rows.
2. The export UI contains a **Separate tasks by** option behind `isWorklogExport()`, but every `DialogWorklogExportComponent` caller omits `data.isWorklogExport`. The option is therefore unreachable in History, the weekly Worklog, and task-summary exports.

## Reproduction

1. Track time for a task whose title contains a semicolon or line break, such as `Client; follow-up`.
2. Finish the day and open History.
3. Export the month to CSV.
4. Open the file in a spreadsheet or inspect the raw row.

### Actual

The title is inserted verbatim into a semicolon-delimited row, producing an extra column (or an extra row for a line break). The task-separator setting is not available in the export dialog.

### Expected

- Exported fields containing the delimiter, quotes, CR, or LF are quoted.
- Quotes inside a quoted field are doubled.
- Existing UTF-8 BOM, semicolon delimiter, column order, and simple-field output remain compatible.
- Worklog exports expose the existing **Separate tasks by** control, or the unused control/input is removed if it is no longer intended.

## Acceptance criteria

- Add focused unit coverage for delimiter, quote, CR/LF, and ordinary field formatting.
- Add one browser-level export regression using a task title with a delimiter or line break and verify the downloaded file parses into the expected five columns.
- Decide the intended `isWorklogExport` contract and test that the task-separator control is either reachable in the relevant dialogs or deliberately removed.
- No new dependency; use the existing formatter and download path.

## Relevant files

- `src/app/features/worklog/worklog-export/worklog-export.util.ts`
- `src/app/features/worklog/worklog-export/worklog-export.component.html`
- `src/app/features/worklog/dialog-worklog-export/dialog-worklog-export.component.html`
- `src/app/features/history/history.component.ts`
- `src/app/features/worklog/worklog-week/worklog-week.component.ts`
- `src/app/features/tasks/task-summary-tables/task-summary-tables.component.ts`
- `e2e/tests/worklog/worklog-basic.spec.ts`

Discovered while implementing #9262 Batch 4.

lntutor:
## Summary
- emit valid semicolon-delimited CSV by quoting delimiters, quotation marks, and line breaks
- expose the existing task-value separator in History, weekly Worklog, and task-summary exports
- cover special-character exports in focused unit and browser download tests, and document the behavior

Fixes #9267

## Verification
- `npm run test:file src/app/features/worklog/worklog-export/worklog-export.util.spec.ts` (23 passed)
- `npm run checkFile src/app/features/worklog/worklog-export/worklog-export.util.ts`
- `npm run checkFile src/app/features/worklog/worklog-export/worklog-export.util.spec.ts`
- `npm run checkFile src/app/features/history/history.component.ts`
- `npm run checkFile src/app/features/worklog/worklog-week/worklog-week.component.ts`
- `npm run checkFile src/app/features/tasks/task-summary-tables/task-summary-tables.component.ts`
- `npm run checkFile e2e/tests/worklog/worklog-basic.spec.ts`
- `npm run e2e:file e2e/tests/worklog/worklog-basic.spec.ts -- --retries=0 --grep 'show worklog after completing tasks'`\n- `git diff --check`\n\nThe repository-wide lint command exits non-zero for existing, unrelated warnings; this change adds none.

### DIFF
--- a/docs/wiki/4.21-Worklog.md
+++ b/docs/wiki/4.21-Worklog.md
@@ -37,7 +37,7 @@ So the worklog is a **reorganized view** of the same “time spent per day” da
 
 - **Inline time correction** — Leaf-task rows can be corrected inline. The app updates the task's “time spent per day” value for that date, so the corrected time is reflected in History, reports, and metrics.
 - **Parent rows** — Parent tasks show the sum of their subtasks and are not edited directly, which avoids double-counting.
-- **Export** — You can export the worklog (e.g. to CSV or for external time sheets). Export uses the same task–date–duration data and can be **filtered by project and date range**. An option lets you **exclude time that falls outside the chosen range**, so only time within the range is included in the export. When export is grouped by individual worklog rows, each row represents a task, date, and duration; start and end times are day-level values rather than per-task session times. That way you get a clean slice of work for the period you care about.
+- **Export** — You can export the worklog (e.g. to CSV or for external time sheets). Export uses the same task–date–duration data and can be **filtered by project and date range**. An option lets you **exclude time that falls outside the chosen range**, so only time within the range is included in the export. CSV output uses semicolons and safely quotes fields that contain semicolons, quotation marks, or line breaks. You can choose the text used to separate multiple task values in one exported field. When export is grouped by individual worklog rows, each row represents a task, date, and duration; start and end times are day-level values rather than per-task session times. That way you get a clean slice of work for the period you care about.
 
 ## Summary
 
--- a/e2e/tests/worklog/worklog-basic.spec.ts
+++ b/e2e/tests/worklog/worklog-basic.spec.ts
@@ -32,7 +32,7 @@ test.describe('Worklog', () => {
     await workViewPage.waitForTaskList();
 
     // Create and complete a task with deterministic tracked time
-    const taskName = `${testPrefix}-Worklog Task`;
+    const taskName = `${testPrefix}-Worklog; Task`;
     await workViewPage.addTask(`${taskName} 10m/1h`);
 
     const task = taskPage.getTaskByText(taskName);
@@ -109,6 +109,7 @@ test.describe('Worklog', () => {
     const exportDialog = page.locator('dialog-worklog-export');
     const previewRow = exportDialog.locator('table tr').filter({ hasText: taskName });
     await expect(previewRow).toContainText('0:20');
+    await expect(exportDialog.locator('input[name="separateTasksBy"]')).toBeVisible();
 
     const [download] = await Promise.all([
       page.waitForEvent('download'),
@@ -126,9 +127,13 @@ test.describe('Worklog', () => {
     const csv = await readFile(downloadPath, 'utf8');
     const rows = csv.replace(/^\uFEFF/, '').split(/\r?\n/);
     expect(rows[0]).toBe('Date;Start;End;Worked;Titles');
-    expect(rows.find((row) => row.endsWith(`;${taskName}`))).toMatch(
-      /^\d{4}-\d{2}-\d{2}; - ; - ;0:20;/,
-    );
+    const csvRow = rows.find((row) => row.includes('0:20'));
+    expect(csvRow).toMatch(/^\d{4}-\d{2}-\d{2}; - ; - ;0:20;"/);
+    const columns = csvRow
+      ?.match(/(?:^|;)(?:"(?:[^"]|"")*"|[^;]*)/g)
+      .map((column) => column.replace(/^;/, ''));
+    expect(columns).toHaveLength(5);
+    expect(columns?.[4]).toBe(`"${taskName}"`);
   });
 
   test('should navigate to worklog from side menu', async ({ page, workViewPage }) => {
--- a/e2e/tests/worklog/worklog-basic.spec.ts
+++ b/e2e/tests/worklog/worklog-basic.spec.ts
@@ -109,7 +109,6 @@ test.describe('Worklog', () => {
     const exportDialog = page.locator('dialog-worklog-export');
     const previewRow = exportDialog.locator('table tr').filter({ hasText: taskName });
     await expect(previewRow).toContainText('0:20');
-    await expect(exportDialog.locator('input[name="separateTasksBy"]')).toBeVisible();
 
     const [download] = await Promise.all([
       page.waitForEvent('download'),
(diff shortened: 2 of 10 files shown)


johannesjo (commented):
## Review

The escaping half is correct and genuinely well tested. Two things I'd simplify before merge, both of which make the diff smaller than it currently is.

**What I verified:** unit specs pass (23/23); sabotaging `escapeCsvField` to a pass-through makes the new spec fail — it's a real oracle, not a vacuous test. The general E2E suite *does* run on PRs (`ci.yml` → `test-on-linux` → `npm run e2e:ci`, reported as the green "Tests" check), and the "Test E2E" step succeeded on `65305b2`, so the browser-level assertion actually ran. No conflicting changes on master for any touched file since the PR's base.

---

### 1. `isWorklogExport` is now dead flexibility — delete it rather than pass `true`

After this PR the flag has no `false` case: there are exactly three `DialogWorklogExportComponent` call sites and all three pass `true`, and `<worklog-export>` is used nowhere except that dialog. So `input<boolean>()` + `@if (isWorklogExport())` + three `data.` lines exist to express a condition that is always true.

It's worth knowing how it got here. `git log -S` pins the loss to `e279f3c29d` *"refactor(worklog): move stuff to service"* (2019-03-08) — a refactor that rewrote History's dialog `data` object and dropped `isWorklogExport: true` in passing. It was an accident, and nobody has reported the control missing in the seven years since.

The flag also can't be type-checked: `DialogWorklogExportComponent` does `data = inject(MAT_DIALOG_DATA)` with no generic, so `[isWorklogExport]="data.isWorklogExport"` binds through `any`. A typo like `isWorkLogExport` at any of the three call sites silently restores the seven-year-dead state, and nothing would catch it.

Suggestion: drop the `input`, drop the `@if`, drop the three `data.isWorklogExport: true` lines. Net −5 instead of +3, control unconditionally visible, typo footgun gone.

If you'd rather *not* resurrect the control at all — seven years without a complaint is an argument, and `' | '` is a fine calm default — then deleting the control and the input is the other clean resolution the issue offered. What shouldn't survive is the flag itself.

### 2. If the flag stays, its half of the fix currently has no test

`65305b22` removed the only assertion covering it. The diagnosis in that commit message is right — the input is inside the collapsed-by-default options panel, so `toBeVisible()` never matched — but `toBeAttached()` won't work either: `CollapsibleComponent` wraps projected content in `@if (isExpanded)` (`collapsible.component.html:34`, `isExpanded = false`), so the input isn't in the DOM at all until the panel is opened. The assertion needs the panel expanded first:

```ts
await exportDialog.locator('.options-collapsible .collapsible-header').click();
await expect(exportDialog.locator('input[name="separateTasksBy"]')).toBeVisible();
```

This is moot if you take option 1 — with the flag gone there's nothing left to regress.

### 3. `worklog-basic.spec.ts:129-135` — the hand-rolled CSV parser can be one line

```ts
const columns = csvRow
  ?.match(/(?:^|;)(?:"(?:[^"]|"")*"|[^;]*)/g)
  .map((column) => column.replace(/^;/, ''));
expect(columns).toHaveLength(5);
expect(columns?.[4]).toBe(`"${taskName}"`);
```

A CSV tokenizer in a test is machinery that has to be maintained and can itself be wrong. Asserting the whole row is shorter *and* strictly stronger — exact column count is implied, and it pins the other four fields too:

```ts
expect(csvRow?.replace(/^\d{4}-\d{2}-\d{2}/, 'DATE')).toBe(
  `DATE; - ; - ;0:20;"${taskName}"`,
);
```

Seven lines → one, and it drops the `toMatch(/…;"/)` prefix check above as well.

---

### Minor

- `worklog-basic.spec.ts:131` — `csvRow?.match(…).map(…)`: the optional chain guards `csvRow`, not `match()` returning `null`. `e2e/tsconfig.json` sets `strict: true`, so `tsc -p e2e/tsconfig.json` reports `TS2531: Object is possibly 'null'` on that exact line; it only slips through because the e2e project isn't type-checked in CI (it already has ~20 pre-existing errors). Goes away entirely with the rewrite in #3.
- `worklog-export.util.ts:22` — `CSV_DELIMITER = ';'` is only half-applied: the escape trigger `/[;"\r\n]/` still hardcodes the delimiter, so changing the constant would produce unescaped output. Either derive the regex from it or just keep the `';'` literals as before — the constant currently suggests a single point of control that doesn't exist.
- `docs/wiki/4.21-Worklog.md` — `4.XX` is a Concepts note, which `docs/documentation-guide.md` says to touch "only with strong evidence"; the escaping sentence is Reference-grade mechanical detail. Defensible since the Export bullet already describes mechanics, but the guide asks that non-Reference edits be called out in the PR description.

### Notes, not asks

- **If the separator control does become reachable**, clearing it is a footgun worth knowing about: `separateTasksBy === ''` makes `TITLES`/`TITLES_INCLUDING_SUB` fall back to `join('<br>')` (`worklog-export.util.ts:307-309`) — literal HTML inside a CSV cell — while `NOTES`/`PROJECTS`/`TAGS` have no fallback at all and just concatenate (`tag1tag2`). A uniform fallback to `WORKLOG_EXPORT_DEFAULTS.separateTasksBy` would settle both. Pre-existing either way; only becomes user-reachable if the control ships.
- **Formula injection is the other half of "safe CSV"**, if you want a follow-up issue. Escaping makes the file well-formed but not inert: a field starting with `=`, `+`, `-` or `@` is still evaluated by Excel/LibreOffice/Sheets (CWE-1236), and task titles aren't always self-authored — they arrive from GitHub/Jira/GitLab imports. The mitigation changes exported values, so it's a deliberate call rather than something to bolt onto this PR.

---

### Strengths

- `escapeCsvField` is minimal and correct — quotes only when needed, doubles inner quotes, preserves the BOM/delimiter/column contract, no new dependency.
- The unit spec asserts the whole formatted string against a literal and fails when the escaping is removed. Verified by sabotage.
- Correctly scoped: `formatText` is the only CSV producer in the app, so one fix covers all three export entry points.
- The E2E exercises the real download path and file contents rather than a mock.

**Verdict:** escaping half is ready. The `isWorklogExport` decision (#1) is the one thing I'd settle before merge — everything else is simplification.

lntutor:
Addressed the simplifications in d8600c3d8: removed the dead isWorklogExport input, conditional, binding, and three always-true dialog data values; the separator control is now unconditional. The E2E assertion now compares the complete normalized CSV row, and the partial CSV_DELIMITER abstraction was removed. Focused unit tests pass (23/23), the targeted Playwright export scenario passes, and every modified TypeScript file passes checkFile. The Worklog Concepts note remains a minimal update to its existing Export reference bullet and is called out here for prose review.

johannesjo:
Thank you very much! <3

PR #9314, opened 2026-07-26
fix: open recurring Schedule projections reliably

### LINKED ISSUE #9270: Recurring occurrence editing is unreliable in Schedule view
## Problem

The recurring-task dialog and Schedule view use three inconsistent paths when a user opens or skips a projected occurrence:

1. `DialogEditTaskRepeatCfgComponent.deleteInstance()` formats a persisted `YYYY-MM-DD` value with `new Date(targetDate)`. JavaScript parses that form as UTC, so in negative-offset time zones the confirmation can display the previous calendar day. The skip button in the same component already uses the timezone-safe `dateStrToUtcDate()` path.
2. `ScheduleEventComponent.clickHandler()` derives `targetDate` from `evt.id.split('_')[1]`. Repeat config IDs are generated IDs and may themselves contain `_`, so this can pass a fragment of the ID instead of the projected date. The event already exposes the authoritative `plannedForDay`.
3. The click handler recognizes `RepeatProjection`, `RepeatProjectionSplit`, and `ScheduledRepeatProjection`, but omits `RepeatProjectionSplitContinued` and `RepeatProjectionSplitContinuedLast`. Clicking a continued segment therefore does not open the recurrence dialog.

These paths can show the wrong date, skip the wrong/no occurrence, or make parts of a split projection impossible to edit from Schedule.

This was found while expanding the E2E coverage tracked by #9262.

## Relevant code

- `src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.ts`
- `src/app/features/schedule/schedule-event/schedule-event.component.ts`
- `src/app/features/schedule/schedule.service.ts` (already has complete repeat-projection type/date handling to mirror)

## Acceptance criteria

- The skip button and its confirmation show the same intended calendar date in positive and negative UTC offsets.
- Schedule passes `plannedForDay` as the recurrence dialog's `targetDate`; generated IDs containing underscores cannot affect the date.
- Every repeat-projection event type, including both continued split variants, opens the recurrence dialog.
- Regression tests cover timezone-safe confirmation formatting, an ID containing underscores, and all repeat-projection event variants.
- Add a focused Schedule E2E flow if component-level coverage cannot prove the real click-to-dialog path.

lntutor:
## Summary
- pass the projection's authoritative `plannedForDay` to the recurrence dialog instead of parsing generated IDs
- allow both continued split-projection event variants to open the dialog
- format date-only skip confirmations without UTC date parsing

Fixes #9270

## Verification
- `npm run test:file src/app/features/schedule/schedule-event/schedule-event.component.spec.ts` (15 passed)
- `npm run test:file src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.spec.ts` (40 passed)
- `npm run checkFile src/app/features/schedule/schedule-event/schedule-event.component.ts`
- `npm run checkFile src/app/features/schedule/schedule-event/schedule-event.component.spec.ts`
- `npm run checkFile src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.ts`
- `npm run checkFile src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.spec.ts`
- `git diff --check`

The repository-wide lint command exits non-zero for existing, unrelated warnings; this change adds none.

### DIFF
--- a/src/app/features/schedule/schedule-event/schedule-event.component.ts
+++ b/src/app/features/schedule/schedule-event/schedule-event.component.ts
@@ -382,13 +382,15 @@ export class ScheduleEventComponent implements AfterViewInit, OnDestroy {
     } else if (
       evt.type === SVEType.RepeatProjection ||
       evt.type === SVEType.RepeatProjectionSplit ||
-      evt.type === SVEType.ScheduledRepeatProjection
+      evt.type === SVEType.ScheduledRepeatProjection ||
+      evt.type === SVEType.RepeatProjectionSplitContinued ||
+      evt.type === SVEType.RepeatProjectionSplitContinuedLast
     ) {
       const repeatCfg: TaskRepeatCfg = evt.data as TaskRepeatCfg;
       this._matDialog.open(DialogEditTaskRepeatCfgComponent, {
         data: {
           repeatCfg,
-          targetDate: (evt.id.includes('_') && evt.id.split('_')[1]) || undefined,
+          targetDate: evt.plannedForDay,
         },
       });
     } else if (evt.type === SVEType.CalendarEvent) {
--- a/src/app/features/schedule/schedule-event/schedule-event.component.ts
+++ b/src/app/features/schedule/schedule-event/schedule-event.component.ts
@@ -390,7 +390,7 @@ export class ScheduleEventComponent implements AfterViewInit, OnDestroy {
       this._matDialog.open(DialogEditTaskRepeatCfgComponent, {
         data: {
           repeatCfg,
-          targetDate: evt.plannedForDay,
+          targetDate: evt.sourceOccurrenceDate ?? evt.plannedForDay,
         },
       });
     } else if (evt.type === SVEType.CalendarEvent) {
--- a/src/app/features/schedule/map-schedule-data/create-blocked-blocks-by-day-map.spec.ts
+++ b/src/app/features/schedule/map-schedule-data/create-blocked-blocks-by-day-map.spec.ts
@@ -1,5 +1,11 @@
 import { createBlockedBlocksByDayMap } from './create-blocked-blocks-by-day-map';
 import { TaskCopy, TaskWithDueTime } from '../../tasks/task.model';
+import {
+  DEFAULT_TASK_REPEAT_CFG,
+  TaskRepeatCfg,
+} from '../../task-repeat-cfg/task-repeat-cfg.model';
+import { BlockedBlockType } from '../schedule.model';
+import { getDbDateStr } from '../../../util/get-db-date-str';
 
 const NDS = '1970-01-01';
 const H = 60 * 60 * 1000;
@@ -42,6 +48,41 @@ describe('createBlockedBlocksByDayMap()', () => {
   it('should work for empty case', () => {
     const r = createBlockedBlocksByDayMap([], [], [], undefined, undefined, 0);
     expect(r).toEqual({});
+  });
+
+  it('keeps a repeat projection source day when a scheduled occurrence crosses midnight', () => {
+    const occurrenceDate = new Date(2026, 6, 30, 12).getTime();
+    const occurrenceDay = getDbDateStr(occurrenceDate);
+    const continuedDay = getDbDateStr(new Date(2026, 6, 31, 12));
+    const repeatCfg: TaskRepeatCfg = {
+      ...DEFAULT_TASK_REPEAT_CFG,
+      id: 'daily-across-midnight',
+      quickSetting: 'DAILY',
+      repeatCycle: 'DAILY',
+      startDate: occurrenceDay,
+      startTime: '23:00',
+      defaultEstimate: h(2),
+    };
+
+    const result = createBlockedBlocksByDayMap(
+      [],
+      [repeatCfg],
+      [],
+      undefined,
+      undefined,
+      occurrenceDate,
+      2,
+      new Date(2026, 6, 31, 12).getTime(),
+    );
+    const continuedEntry = result[continuedDay]
+      .flatMap((block) => block.entries)
+      .find(
+        (entry) => entry.type === BlockedBlockType.ScheduledRepeatProjectionSplit,
+      ) as {
+      sourceOccurrenceDate?: string;
+    };
+
+    expect(continuedEntry.sourceOccurrenceDate).toBe(occurrenceDay);
   });
 
   it('should work for basic case', () => {
--- a/src/app/features/schedule/map-schedule-data/create-sorted-blocker-blocks.ts
+++ b/src/app/features/schedule/map-schedule-data/create-sorted-blocker-blocks.ts
@@ -118,6 +118,7 @@ const createBlockerBlocksForScheduledRepeatProjections = (
             data: repeatCfg,
             start,
             end,
+            sourceOccurrenceDate: currentDayStr,
           },
         ],
       });
--- a/src/app/features/schedule/map-schedule-data/create-view-entries-for-block.ts
+++ b/src/app/features/schedule/map-schedule-data/create-view-entries-for-block.ts
@@ -27,7 +27,9 @@ export const createViewEntriesForBlock = (
         data: repeatCfg,
         duration: repeatCfg.defaultEstimate || 0,
         plannedForDay: dayDate,
-        sourceOccurrenceDate: entry.sourceOccurrenceDate,
+        ...(entry.sourceOccurrenceDate
+          ? { sourceOccurrenceDate: entry.sourceOccurrenceDate }
+          : {}),
       });
     } else if (entry.type === BlockedBlockType.CalendarEvent) {
       const calendarEvent = entry.data;
@@ -68,7 +70,9 @@ export const createViewEntriesForBlock = (
         data: repeatCfg,
         duration: entry.end - entry.start,
         plannedForDay: dayDate,
-        sourceOccurrenceDate: entry.sourceOccurrenceDate,
+        ...(entry.sourceOccurrenceDate
+          ? { sourceOccurrenceDate: entry.sourceOccurrenceDate }
+          : {}),
       });
     }
   });
--- a/src/app/features/schedule/map-schedule-data/create-view-entries-for-block.ts
+++ b/src/app/features/schedule/map-schedule-data/create-view-entries-for-block.ts
@@ -27,6 +27,7 @@ export const createViewEntriesForBlock = (
         data: repeatCfg,
         duration: repeatCfg.defaultEstimate || 0,
         plannedForDay: dayDate,
+        sourceOccurrenceDate: entry.sourceOccurrenceDate,
       });
     } else if (entry.type === BlockedBlockType.CalendarEvent) {
       const calendarEvent = entry.data;
@@ -67,6 +68,7 @@ export const createViewEntriesForBlock = (
         data: repeatCfg,
         duration: entry.end - entry.start,
         plannedForDay: dayDate,
+        sourceOccurrenceDate: entry.sourceOccurrenceDate,
       });
     }
   });
--- a/src/app/features/schedule/map-schedule-data/insert-blocked-blocks-view-entries-for-schedule.ts
+++ b/src/app/features/schedule/map-schedule-data/insert-blocked-blocks-view-entries-for-schedule.ts
@@ -211,6 +211,8 @@ export const insertBlockedBlocksViewEntriesForSchedule = (
               taskRepeatCfg,
               duration: timePlannedForSplitContinued,
               splitIndex: 0,
+              sourceOccurrenceDate:
+                currentVE.sourceOccurrenceDate ?? currentVE.plannedForDay,
             });
 
             // move entries
@@ -258,6 +260,8 @@ export const insertBlockedBlocksViewEntriesForSchedule = (
               taskRepeatCfg: currentVE.data,
               duration: timePlannedForSplitRepeatCfgProjectionContinued,
               splitIndex,
+              sourceOccurrenceDate:
+                currentVE.sourceOccurrenceDate ?? currentVE.plannedForDay,
             });
 
             // move entries
@@ -370,12 +374,16 @@ export const createSplitRepeat = ({
   taskRepeatCfg,
   splitIndex,
   duration,
+  sourceOccurrenceDate,
 }: {
   start: number;
   dayDate: string;
   taskRepeatCfg: TaskRepeatCfg;
   splitIndex: number;
   duration: number;
+  // the day the occurrence belongs to, which is not dayDate once a segment has
+  // been pushed across midnight -- without it a tail resolves to its render day
+  sourceOccurrenceDate?: string;
 }): SVERepeatProjectionSplitContinued => {
   return {
     id: `${taskRepeatCfg.id}_${dayDate}_${splitIndex}`,
@@ -385,5 +393,6 @@ export const createSplitRepeat = ({
     splitIndex,
     data: taskRepeatCfg,
     plannedForDay: dayDate,
+    ...(sourceOccurrenceDate ? { sourceOccurrenceDate } : {}),
   };
 };
--- a/src/app/features/schedule/map-schedule-data/map-schedule-days-to-schedule-events.spec.ts
+++ b/src/app/features/schedule/map-schedule-data/map-schedule-days-to-schedule-events.spec.ts
@@ -137,6 +137,20 @@ describe('mapScheduleDaysToScheduleEvents()', () => {
     expect(res.eventsFlat[0].isBeyondBudget).toBe(true);
   });
 
+  it('preserves the source occurrence date of a continued repeat projection', () => {
+    const entry = {
+      ...fakeTaskEntry('CONTINUED', {
+        plannedForDay: '2026-07-31',
+        start: new Date(2026, 6, 31, 0, 0).getTime(),
+      }),
+      sourceOccurrenceDate: '2026-07-30',
+    };
+
+    const res = mapScheduleDaysToScheduleEvents([fakeDay({ entries: [entry] })], FH);
+
+    expect(res.eventsFlat[0].sourceOccurrenceDate).toBe('2026-07-30');
+  });
+
   it('should set calendar badge day for beyond budget planned tasks', () => {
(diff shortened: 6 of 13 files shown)


src/app/features/schedule/schedule-event/schedule-event.component.ts
```diff
@@ -382,13 +382,15 @@ export class ScheduleEventComponent implements AfterViewInit, OnDestroy {
...
+      evt.type === SVEType.ScheduledRepeatProjection ||
+      evt.type === SVEType.RepeatProjectionSplitContinued ||
+      evt.type === SVEType.RepeatProjectionSplitContinuedLast
     ) {
       const repeatCfg: TaskRepeatCfg = evt.data as TaskRepeatCfg;
       this._matDialog.open(DialogEditTaskRepeatCfgComponent, {
         data: {
           repeatCfg,
-          targetDate: (evt.id.includes('_') && evt.id.split('_')[1]) || undefined,
+          targetDate: evt.plannedForDay,
```
johannesjo: **Important:** `plannedForDay` is the visual day for some continued segments, not the original occurrence day. For example, a daily repeat projected for July 30 from 23:00–01:00 is split at midnight; `create-view-entries-for-block.ts` stamps its July 31 `RepeatProjectionSplitContinuedLast` with `plannedForDay: "2026-07-31"`. Clicking that tail here and confirming removal persists July 31 in `deletedInstanceDates`, so remapping leaves the clicked July 30 tail visible but removes the actual July 31 23:00 occurrence. Please carry the source occurrence date through blocked-entry/midnight splitting (and preserve it across subsequent splits), then add a regression using the real schedule producer rather than assigning the same date to every synthetic variant.

johannesjo (commented):
I found one correctness issue in the newly enabled continued-projection path; details are inline. The date-only formatting change and underscore-safe target-date direction otherwise look sound.

lntutor:
Addressed the continued-projection date issue in 5f7c71229. Scheduled repeat projections now retain their source occurrence date through blocked-entry splitting and into the rendered event; the edit dialog uses that source date while plannedForDay remains the visual segment date. Added a real producer regression for a daily 23:00–01:00 occurrence plus dialog coverage. Focused suites: 5 producer tests and 15 schedule-event component tests pass; changed-file checks pass.

lntutor:
Follow-up: I found and fixed the final mapping boundary as well. 6f3ec5a22 now carries sourceOccurrenceDate from the rendered view entry into ScheduleEvent before the dialog consumes it. Added the mapper regression; focused coverage is 5 producer, 9 mapper, and 15 component tests, all passing.

lntutor:
Follow-up CI fix pushed in `5ae19fe27`: scheduled repeat view entries now omit `sourceOccurrenceDate` when it is absent, while exact-object schedule mapper tests assert the occurrence date for scheduled repeat projections.\n\nVerification:\n- `npm run test:file -- src/app/features/schedule/map-schedule-data/create-view-entries-for-block.spec.ts` (3 passed)\n- `npm run test:file -- src/app/features/schedule/map-schedule-data/map-to-schedule-days.spec.ts` (13 passed)\n- `npm run test:file -- src/app/features/schedule/map-schedule-data/map-to-schedule-days-extended.spec.ts` (1 passed)\n- `npm run test` (Berlin: 13,519 passed; Los Angeles: 13,505 passed)\n- required changed-file checks and push-hook lint passed with only pre-existing repository warnings.

src/app/features/schedule/schedule-event/schedule-event.component.ts
lntutor: Implemented in `6f3ec5a22`: `sourceOccurrenceDate` is assigned by the real scheduled-repeat producer, carried through blocked entries and repeated midnight splits, mapped into schedule events, and used by the dialog while `plannedForDay` remains the visual segment date. The regression starts from the real producer with a 23:00–01:00 occurrence and verifies the July 31 tail retains July 30 as its source occurrence.

johannesjo:
Re-reviewed at `5ae19fe27`. The round-1 issue is genuinely closed for the **timed** path, and I checked it against the real producer rather than the unit seams: `mapToScheduleDays` with a daily 23:00–01:00 cfg puts a `RepeatProjectionSplitContinuedLast` on 2026-07-31 carrying `sourceOccurrenceDate: '2026-07-30'`, and that survives into the `ScheduleEvent`. Stamping the date on the blocked-block entry so `splitEntry`'s spread preserves it across repeated midnight splits is the right shape, and keeping `plannedForDay` as the *visual* day was the correct call — `ScheduleService.getEventDayStr()` relies on it for month-view bucketing.

Also confirmed the issue's premise rather than assuming it: nanoid's alphabet really does contain `_` (`…BUSHWOLF_GQZ…`), so ~28% of 21-char cfg ids would have made the old `evt.id.split('_')[1]` return an id fragment. Those garbage strings are inert once written (`deletedInstanceDates` is only ever read via exact-match `includes`), so no cleanup migration is warranted here.

One **Important** issue remains — the same defect on the untimed path, which this PR is what makes reachable — plus a test gap and two nits.

## Important — flow-path continued segments still resolve to the render day

`createSplitRepeat()` (`insert-blocked-blocks-view-entries-for-schedule.ts:367`, untouched by this PR) stamps `plannedForDay: dayDate` — the day being *rendered* — and copies no `sourceOccurrenceDate`. Harmless while those two coincide, but they need not: `create-schedule-days.ts:246-256` pushes a `RepeatProjectionSplitContinued(Last)` whose `start` reaches the day boundary into `viewEntriesPushedToNextDay`, and `create-view-entries-for-day.ts:28-36` re-feeds it at the top of the *next* day, where branch `CCC D` re-splits it and the new tail is stamped with that next day.

Repro through the real `mapToScheduleDays` — untimed monthly cfg due only 2026-07-30 (2h), a scheduled task 07-30 23:30 +1h so the day-30 block clips at midnight, and a scheduled task 07-31 01:00 +30m to re-split the carried-over segment. Day 1 comes out as:

```
RepeatProjectionSplitContinued      01:00  plannedForDay=2026-07-30  sourceOccurrenceDate=undefined
RepeatProjectionSplitContinuedLast  01:30  plannedForDay=2026-07-31  sourceOccurrenceDate=undefined
```

Both segments belong to the **07-30** occurrence, yet `evt.sourceOccurrenceDate ?? evt.plannedForDay` yields `["2026-07-30", "2026-07-31"]` — the two halves of one occurrence disagree.

Before this PR, both continued types were absent from the `else if` chain, so clicking them did nothing. Now clicking the tail opens the dialog for the wrong day, and confirming writes `'2026-07-31'` into `deletedInstanceDates`: a silent no-op at best, and for a cfg that also recurs the following day it removes *that* day's occurrence while leaving the clicked segment on screen. That is exactly the failure mode from my round-1 comment, so I'd rather close it here than ship the click-through with a known wrong-date branch.

The fix is small — thread the occurrence date through `createSplitRepeat`:

```ts
export const createSplitRepeat = ({
  start, dayDate, taskRepeatCfg, splitIndex, duration, sourceOccurrenceDate,
}: { /* … */ sourceOccurrenceDate?: string }): SVERepeatProjectionSplitContinued => ({
  // …
  plannedForDay: dayDate,
  ...(sourceOccurrenceDate ? { sourceOccurrenceDate } : {}),
});
```

and at both call sites (branches `CCC C` and `CCC D`):

```ts
sourceOccurrenceDate: currentVE.sourceOccurrenceDate ?? currentVE.plannedForDay,
```

I applied exactly that locally: the repro above goes green, and the only fallout is three `RepeatProjectionSplitContinuedLast` expectations in `map-to-schedule-days.spec.ts` (2 specs) needing `sourceOccurrenceDate` added — the same mechanical update you already made for the head entries.

## Test gap — the cross-midnight line is unguarded

Deleting *only* these three lines in `create-view-entries-for-block.ts` —

```ts
...(entry.sourceOccurrenceDate
  ? { sourceOccurrenceDate: entry.sourceOccurrenceDate }
  : {}),
```

— leaves `npm run test:file -- 'src/app/features/schedule/**/*.spec.ts'` at **359/359 passing**. That is the only line that makes a cross-midnight timed projection resolve to its occurrence day instead of the render day, i.e. the exact scenario from my round-1 comment. (Control: deleting the *other* spread, in the `ScheduledRepeatProjection` branch, fails 5 specs — so the suite can see this class of change; this particular link just has nothing joining the two ends.)

The new specs cover the neighbours — `create-blocked-blocks-by-day-map.spec.ts` proves the producer stamps the entry, and `map-schedule-days-to-schedule-events.spec.ts` proves the mapper copies it off a hand-built entry — but nothing runs the producer's output through `createViewEntriesForBlock`. One real-producer spec closes it and would have caught the deletion:

```ts
const days = mapToScheduleDays(
  new Date(2026, 6, 30, 12).getTime(),
  ['2026-07-30', '2026-07-31'],
  [], [], [dailyCfgAt23hFor2h], [], [], null, {},
  undefined, undefined,
  new Date(2026, 6, 29, 12).getTime(),
);
const tail = days[1].entries.find((e) => e.type === SVEType.RepeatProjectionSplitContinuedLast);
expect(tail.plannedForDay).toBe('2026-07-31');
expect(tail.sourceOccurrenceDate).toBe('2026-07-30');
```

Given how thin the seam-level specs are here, I'd swap the synthetic mapper test for this one rather than add to it.

## Nits

- **`deleteInstance` spec is weaker than it looks in one zone.** It spies on `Date.prototype.toLocaleDateString` and asserts `String(this.getHours()) === '0'`. `dateStrToUtcDate()` returns *local* midnight despite the name, so unfixed this fails in Berlin (hour 2) and Los Angeles (hour 17) — both of which `npm test` runs, so CI is safe. But in UTC+0 the unfixed `new Date('2026-06-10')` is also local hour 0, and the spec passes. Asserting the calendar date too (`${getFullYear()}-${getMonth()+1}-${getDate()}`) makes it zone-independent, and the date is the thing users actually see wrong.
- **The click path itself is still untested.** The new spec calls `component.clickHandler(...)` directly; the binding it stands in for is host metadata (`'(click)': 'clickHandler($event)'`). The TestBed uses the real template, so `fixture.nativeElement.dispatchEvent(new MouseEvent('click'))` would prove the whole path for one variant at roughly zero cost — and would satisfy #9270's "focused E2E if component-level coverage cannot prove the real click-to-dialog path" without adding an E2E file.

Everything else looks good — CI is green, the model additions are optional-only so nothing persisted changes shape, and I confirmed both TZ variants pass locally on the schedule and task-repeat-cfg suites.

johannesjo:
Double-checked everything from my previous comment by re-running it myself rather than trusting the first pass, and added a KISS/YAGNI read on the design. Net: **no corrections — one claim gets stronger, and the design should stay as it is.**

## Re-verified

- Deleting the three-line spread in `create-view-entries-for-block.ts`'s `ScheduledRepeatProjectionSplit` branch → **359/359 still green**. Control: deleting the sibling spread in the `ScheduledRepeatProjection` branch → **5 FAILED**. So the suite can see this class of change; that one link just has nothing joining its ends.
- The `deleteInstance` spec under `TZ=UTC` with the fix reverted → **40/40 pass**. It only bites in Berlin (hour 2) and Los Angeles (hour 17), both of which `npm test` runs, so CI is safe — the spec is just weaker than it reads.
- The `createSplitRepeat` fix I suggested costs exactly **2 specs / 3 expectation objects**, both in `map-to-schedule-days.spec.ts` (`should calculate the right duration of repeat task projections`, `should work for an example with all the stuff`).
- Master really does drop the two continued types on the floor (its `else if` chain ends at `ScheduledRepeatProjection`), so the click-through is new behaviour rather than a pre-existing wrong date.

## The Important finding is worse than I described

Same repro, but with a **daily** untimed cfg. This is what actually renders on 2026-07-31:

| segment (local) | `plannedForDay` | resolved `targetDate` |
| --- | --- | --- |
| `RepeatProjectionSplitContinued` 00:30 | 2026-07-30 | 2026-07-30 ✅ |
| `RepeatProjectionSplitContinuedLast` 01:30 | 2026-07-31 | **2026-07-31 ❌** |
| `RepeatProjection` 02:30 | 2026-07-31 | 2026-07-31 ✅ |

Rows 1 and 2 are the two halves of Thursday's occurrence; row 3 is Friday's own. Clicking row 2 and confirming the skip deletes **row 3** and leaves row 2 on screen — the occurrence you actually removed is sitting one row below the thing you clicked.

## KISS / YAGNI

I went looking for a leaner shape and did not find one worth having:

- **`sourceOccurrenceDate` is not redundant with `plannedForDay` — don't collapse them.** I checked the other consumer: `schedule-month.component.html` → `getEventsForDay()` → `ScheduleService.getEventDayStr()` reads `ev.plannedForDay` to pick the calendar cell. Making `plannedForDay` the occurrence day would drop the July-31 tail into the July-30 cell next to its own head. The extra field earns its place.
- **Stamping at the producer is the right call, even at the cost of the seven expectation edits.** I did try the alternative — stamp only in `splitEntry`, so the field would mean "displaced from its day" and no head expectation would change. It works, but then "absent means head" is a rule every future reader has to carry, and re-deriving the day from `entry.start` duplicates something the producer already knows. Not worth redoing; leave it as is.
- Nothing else here is speculative: no new setting, no new UI, no dependency, and the `sourceOccurrenceDate` scope came out of my round-1 comment rather than the PR reaching for it.

**One cheap thing that would pay for itself:** two lines of doc on the field in `schedule.model.ts`. Today `plannedForDay` means the *rendered* day on block-derived entries and the *occurrence* day on flow-derived ones — row 1 of the table above renders on July 31 yet carries `plannedForDay: 2026-07-30`, so the month view files it under July 30. That inconsistency is pre-existing and out of scope here, but it is exactly what makes the `createSplitRepeat` gap easy to walk into. Writing the contract down — `plannedForDay` = the day this segment is drawn on, `sourceOccurrenceDate` = the day the recurrence is actually due, set when they differ — is the cheapest guard against the next one.

The two nits from my previous comment stand unchanged.

johannesjo:
I pushed a commit to this branch closing the round-2 finding (`createSplitRepeat` not carrying the occurrence day), so this is good to go from my side once CI is green.

**The repro**, as a spec rather than prose: a 26h untimed monthly projection starts on day 1, is pushed over midnight, and is then split again by a scheduled task on day 2. Its three segments resolved to `1970-01-01`, `1970-01-01`, `1970-01-02` — so with this PR's `sourceOccurrenceDate ?? plannedForDay`, clicking the last segment opens the repeat dialog for the **wrong occurrence**, and "delete instance" would skip a day the user never clicked.

The head and the midnight-pushed segment were only right by accident of `splitEntry`'s `...entry` spread. The re-split tail is built fresh by `createSplitRepeat`, which stamped `plannedForDay: dayDate` — the *render* day. So it needs the value passed in explicitly:

```ts
sourceOccurrenceDate: currentVE.sourceOccurrenceDate ?? currentVE.plannedForDay,
```

at both call sites (the `RepeatProjection` split and the `RepeatProjectionSplitContinued` re-split), plus an optional param on `createSplitRepeat`.

Verified: the new spec fails without the two-line change and passes with it, in both the Berlin and Los Angeles lanes (it is the only guard for this in the LA lane — the two neighbouring specs it also corrects are timezone-skipped there). Full schedule suite is 360 green. It also corrects three existing expectations in `map-to-schedule-days.spec.ts`, which were encoding the pre-fix values for same-day splits where the render day and the occurrence day happen to coincide.

Thanks for the patience on the review rounds — the `sourceOccurrenceDate` design holds up well; this was the one path it hadn't reached.

johannesjo (approved):
Approving. I re-verified the whole chain from scratch this round, against **current master** rather than the PR base (the branch is a few commits behind). The diff applies cleanly and the schedule + task-repeat-cfg suites are **966 green in Berlin / 952 in Los Angeles**.

## Every projection type now resolves to the right occurrence day

Traced to the producer rather than inferred:

| type | where the occurrence day comes from |
| --- | --- |
| `RepeatProjection` | `plannedForDay: dayDate` (`create-view-entries-for-day.ts:125`) — already the due day, so the `?? plannedForDay` fallback is correct |
| `RepeatProjectionSplit` | in-place `currentVE.type` mutation, `plannedForDay` untouched — fallback correct |
| `ScheduledRepeatProjection` | `sourceOccurrenceDate: currentDayStr` stamped at the producer |
| `RepeatProjectionSplitContinuedLast` (block path) | `splitEntry`'s `...entry` spread — survives N midnight splits |
| `RepeatProjectionSplitContinued(Last)` (flow path) | the `createSplitRepeat` param added in `cfaf6b76` |

## The round-trip closes

`sourceOccurrenceDate` is `getDbDateStr(currentDayTimestamp)` for exactly the day `selectTaskRepeatCfgsForExactDay` was projected for — and that is the same string `deletedInstanceDates?.includes(dateStr)` matches on in `task-repeat-cfg.selectors.ts:116`. So "skip instance" now removes precisely the occurrence that was clicked, which is the part that was silently wrong before.

## Both fixes are sabotage-proven, not merely green

- Dropping `sourceOccurrenceDate` from `createSplitRepeat` → **3 failures in Berlin, 1 in Los Angeles**. The new 26h-monthly spec is the *only* guard in the LA lane; the two neighbouring specs it also corrects are timezone-skipped there.
- Reverting the `dateStrToUtcDate` guard in `deleteInstance` → **1 failure in Berlin and 1 in Los Angeles**, but it passes under `TZ=UTC`. So that spec is weaker than it reads — though both lanes CI actually runs do catch it.

## Two non-blocking nits, for whenever this area is next touched

- The `deleteInstance` spec asserts on `getHours()` through a `toLocaleDateString` fake rather than on the rendered date. It earns its place because it bites in both CI lanes, but asserting the formatted date directly would say what it means.
- The click spec calls `clickHandler()` directly, so it would not notice the host `(click)` binding disappearing. Pre-existing shape, not introduced here.

One thing left standing on purpose: `plannedForDay` still means *rendered day* on block-derived entries and *occurrence day* on flow-derived ones. That ambiguity is what made the `createSplitRepeat` gap easy to walk into. Worth a rename someday — not a reason to hold this up.

Thanks @lntutor — this went several layers deeper than the original one-line `split('_')` bug, and you chased each one down.

PR #9318, opened 2026-07-26
feat(settings): add search to global settings

### LINKED ISSUE #8823: 💡 Add search to global settings (~150 fields, 22 sections)
## Problem

Global settings has grown to roughly **150 form fields plus 53 keyboard-shortcut fields**, spread over 6 tabs and 22 collapsible sections — with no search or filter (`src/app/pages/config-page/config-page.component.html` contains no filter input). Finding a specific option (e.g. anything idle-related) means expanding collapsibles across multiple tabs.

## Suggested fix

A single search box above the tabs that filters sections/fields by label and help text (the formly config objects already carry translated labels, so this is mostly a filter over existing metadata). No restructuring of the settings IA needed.

---

MarkQuach12:
## Problem

Closes #8823. Global settings holds around 150 fields plus 53 keyboard shortcut fields across 6 tabs and 22 collapsible sections, with no search. Finding one option means expanding collapsibles across several tabs.

## Solution

Add a search box above the settings tabs. Typing searches every tab at once and shows matches in a dropdown, each row naming the section, its tab, and the field label when a field matched rather than the title. Picking a result switches to that tab, expands the section, and scrolls to it.

Matching covers translated section titles, help text, and field labels, including labels nested in formly `fieldGroup` wrappers. Searching a tab name lists that whole tab. Settings drawn by bespoke components (dark mode, theme, wallpaper, sound, plugins, sync setup) have no form metadata to filter, so they come from a small hand written list in the util.

Results are plain rows in an overlay, so no settings forms are built while searching and the tabs stay put.

Documents the search box in `3.02-Settings-and-Preferences.md`.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo (commented):
Reviewed at `41c991702a`.

**What I ran:** the new util spec (8/8 green) plus two sabotages — dropping the `fieldGroup` recursion and dropping `has(section.help)` each turned exactly one test red, so the tests aren't vacuous. On top of that a Node harness that runs `searchSettings` over the real `GLOBAL_*_FORM_CONFIG` consts with `en.json`, and a throwaway Playwright probe driving the actual settings page.

The feature is wanted (#8823 is @johannesjo's own issue, and you were greenlit on it) and the shape is right: search above the tabs, results in an overlay, no settings forms built while searching, non-formly sections covered by a small hand list. Four things worth fixing before merge — all small — then this is good to go.

---

## 1. Raw HTML renders as literal text in the result rows

`settings-search.util.ts:95` matches `templateOptions.text`. That is the body of `type: 'tpl'` help paragraphs — raw HTML, not a label. It ends up in `matchedLabelKey`, and `config-page.component.html:55` interpolates it, so the markup is shown verbatim.

Actual row from the running app, query `chrome extension`:

```
Idle Handling
Time & Tracking › Please <a target="_blank" href="https://chrome.google.com/webstore/detail/super-productivity/ljkbjodfmekkl…
```

It gets hard-clipped mid-URL rather than ellipsised, because the global `.mdc-list-item__primary-text { white-space: nowrap; overflow: hidden; text-overflow: ellipsis }` (`styles/components/_overwrite-material.scss:378`) no longer has the text as a direct child once the new `display: flex; flex-direction: column` is applied.

`backup` hits the same thing through App Features' experimental warning. The sharpest case is desktop-only — `automatic-backups-form.const.ts:53` is

```ts
text: `<a href="file://${backupPath}" target="_blank">${backupPath}</a>`,
```

so on Electron a query touching the backup path puts the anchor tag *and* the absolute filesystem path into the dropdown.

Simplest fix: drop `templateOptions.text` from the matched list entirely. Section `help` already covers "help text", and it is only used as a boolean — never displayed — so it is safe as-is.

## 2. A third of the jump targets never scroll

`scrollSelector` is `.section-<key>`, but two of the six tabs don't emit that class:

- `config-page.component.html:178` (Productivity) → `<section class="config-section">`
- `config-page.component.html:300` (Sync & Backup) → `<section class="config-section imex-section">`

So `focusMode`, `focusModeLocal`, `takeABreak`, `evaluation` and `IMEX_YEAH` — 5 of the 22 targets, 6 of 23 on desktop/Android where `localBackup` is appended — resolve to `null` and `?.scrollIntoView()` is a silent no-op. Verified in the browser: after picking "Break Reminder", `document.querySelectorAll('.section-takeABreak').length === 0` and the Productivity sections' class lists are all bare `config-section`; `config-page.scrollTop` stays at 24, where the same flow into "Idle Handling" (a tab that does emit the class) moves it.

Fix is `section-{{ section.key }}` on those two `@for` blocks, matching the other four tabs.

## 3. Fields declared with `props` are invisible to the search

`templateOptions` is the deprecated alias — `tsc` flags both uses on line 95 — and `props` is the current API, which this codebase already mixes in. `app-features-form.const.ts:17` uses `props`, so **searching `stopwatch` returns "No settings found"** even though General → App Features → "Stopwatch time tracking" is a real, visible toggle. Confirmed in the browser, and it stays broken after that section has rendered (formly's `templateOptions` → `props` aliasing doesn't reach these consts).

One field today, but the failure mode is that every field written with the modern API is silently unsearchable from here on. `field.props ?? field.templateOptions` covers both.

## 4. "No settings found" pops open before you type anything

`config-page.component.html:59-63`: the `@empty` branch always renders one option, and Material sets `showPanel = !!this.options?.length` (`autocomplete.mjs:136`), so `_handleFocus` opens the panel on an empty query. Clicking into the search box greets you with "No settings found" in a dropdown that covers the tab bar.

Either `@if (searchQuery) { … }` around the `@for`/`@empty`, or `[matAutocompleteDisabled]="!searchQuery"` on the input.

---

## Notes — no action strictly required

- **Tab order is stated in three places** and nothing enforces agreement: the `mat-tab-group`, the array in `onSearchChange`, and `CUSTOM_TARGETS[].tabIndex`. Add the `.section-<key>` classes and the anchors inside `theme-selector`'s inline template and that's five implicit couplings. The util spec runs against a synthetic fixture, so it validates the algorithm but proves nothing about the app — which is exactly why #2 and #3 are live. A single test that feeds the real `GLOBAL_*_FORM_CONFIG` into `searchSettings` and asserts every `scrollSelector` resolves in the rendered config page would have caught both and would keep catching them.
- **`// ponytail:`** at `config-page.component.ts:488` — looks like a stray word; guessing this wanted to be `shortcut:` or `@todo`.
- **Sync provider settings aren't reachable.** `SYNC_FORM` lives in `dialog-sync-cfg`, not on the page, so `webdav`, `nextcloud`, `dropbox`, `encryption` and `password` all return nothing. Reasonable scope call, but those are precisely what people will type, so it's worth a sentence in the wiki paragraph.
- **The spec fixture models a WebDAV `baseUrl` under a "Sync & Backup" tab** — that structure doesn't exist. The `fieldGroup` flattening it exercises is genuinely needed (Focus Mode's "Advanced Config" collapsible wrapper), so a real example would document the intent better.
- **`scrollbar-gutter: stable`** (`config-page.component.scss:6-8`) is an unrelated drive-by that isn't mentioned in the description.
- **`onSearchChange(query: unknown)`** — the `typeof query === 'string'` guard is there because the autocomplete writes the selected object back into the model. `[displayWith]` is the API for that; the current version briefly assigns the object to the input's value before ngModel clears it.
- The Sound custom target sorts above the Tasks tab's sections but renders below them. Trivial.

## What's good

- Searching the tab name to list a whole tab is a nice touch, and the `matchedLabelKey` sub-line ("which field matched") is the right amount of context.
- Not building forms while searching is the correct call, and it keeps this off the settings render path entirely.
- Platform filtering is inherited for free — `filterGlobalConfigForm` already strips Electron-only and Android-hidden sections from the consts, so search never offers a section that isn't rendered. Worth keeping in mind if the tab arrays are ever reshaped.
- Wiki updated in the same PR, `en.json` only, no new dependencies.

Note: the `Tests` check was still running when I wrote this; everything else on the PR is green.

johannesjo (commented):
Re-reviewed at `b6accfd4ee`. **All four r1 findings are properly fixed** — I verified each rather than reading it. Two behaviour bugs turned up that were previously masked, plus some result-quality things worth a look.

First, a correction to my own r1 review: my `[displayWith]` note was **wrong**, please ignore it. `_setValueAndClose` calls `this._onChange(toSelect.value)` with the raw option object unconditionally (`autocomplete.mjs:695-703`); `displayWith` only feeds `_getDisplayValue` for the input's *text* (`:678`). So the `typeof query === 'string'` guard in `onSearchChange` is **required**, not a workaround. Sorry for the misdirection.

## The four r1 items — verified fixed

**1. Raw HTML in result rows** — `templateOptions.text` is gone. I re-ran the harness that feeds the real `GLOBAL_*_FORM_CONFIG` + `en.json` into `searchSettings`, sweeping 395 distinct words harvested from every label, section help and `tpl` body, in both the web and the Electron shape. No row renders markup, an anchor or a filesystem path. Positive control: restoring the old `.flatMap(… .text)` produces 102 leaking rows, including `<strong>Experimental feature:</strong>…`, the Chrome-extension `<a href="https://chrome.google.com/…">` and the Electron backup-path anchor. (For the record: it was always cosmetic — the row is `{{ }}` interpolation, so Angular escaped it. The only angle brackets left are the deliberate `+<project>` / `#<tag>` placeholders in the short-syntax labels.)

**2. Dead `.section-<key>` anchors** — I drove the real app through **all 22 targets in the web build** (16 formly sections + the 6 hand-written ones): type the section title, click the row, then assert the element exists exactly once under `config-page`, sits inside the viewport, and the expected tab is selected. 22/22 land; `focusMode`, `focusModeLocal`, `takeABreak`, `evaluation` and `IMEX_YEAH` all work now. `taskWidget` and `localBackup` are Electron-only so I couldn't exercise them, but they render through the same two `@for` blocks that now carry the class.

**3. `props`-declared fields** — `stopwatch` → 1 hit ("App Features › Stopwatch time tracking"), `advanced config` → 1 hit (Focus Mode's collapsible wrapper); both were 0. Walking every field of every section on the page, including the Electron auto-backup form: 136 fields, 2 `props`-only, 134 `templateOptions`-only, none declaring both, so the `??` order can't shadow a label.

**4. Empty-focus dropdown** — panel stays closed on focus, opens on the first keystroke, shows "No settings found" for a miss, and closes again when the box is cleared.

`npm run checkFile` passes on all five changed `.ts`/`.scss` files; the util spec is 9/9.

---

## 1. Jumping to a section you collapsed by hand does nothing

`CollapsibleComponent.toggleExpand()` (`collapsible.component.ts:136-139`) flips its own `isExpanded` and *does* emit `isExpandedChange` — but `config-section.component.html:1-2` binds `[isExpanded]` one-way and never listens to that output. So after a hand-collapse `config-page` still believes the section is open, the binding value is unchanged, Angular skips the write, and the section stays shut.

Confirmed in the browser:

```
jump to "Idle Handling"        expanded = ["Idle Handling"]
click its header to collapse   expanded = []
jump to "Idle Handling" again  expanded = []          ← only the scroll happens

control — jump away to "Schedule", then back:
                               expanded = ["Idle Handling"]   ← works
```

So the feature silently no-ops on the second try for the one section the user is most likely to search for twice. The clean fix is to forward the existing `(isExpandedChange)` through `config-section` so `expandedSection` stays truthful; nulling it (+ `detectChanges()`) before reassigning would also work. The `?section=` deep link has the same stale binding today, but search makes repeat-jumps routine.

## 2. A second jump inside the same tab scrolls past the target

`goToSearchResult` scrolls after a fixed `TAB_ANIMATION_DURATION_MS` (200 ms, `config-page.component.ts:85`), but at that moment neither animation has settled: the previously expanded section is still collapsing (`TRANSITION_DURATION_LEAVE = 195 ms`, `animation.const.ts:15`) and the target is still expanding (`TRANSITION_DURATION_ENTER = 225 ms`, i.e. longer than the delay). The scroll offset is computed against layout that is still moving, and nothing re-reads it afterwards.

`config-page` is the scroll container and a correctly landed section sits at `top = 52`:

```
downward — "Misc Settings" then "Keyboard Shortcuts"
    top = -19        (collapsing section is ABOVE the target)
downward — "App Features" then "Misc Settings"
    top = -57   … and -206 on a repeat run of the identical action

upward — "Keyboard Shortcuts" then "Misc Settings"
    top = 52         (collapsing section is BELOW the target — correct)

first jump into a tab, nothing expanded yet
    top = 52         (correct)
```

The upward case is the control: the error appears **only** when the section that is collapsing sits above the target, which is what the mechanism predicts. The size scales with that panel's height, but run-to-run jitter is as large as the effect (109 px vs 258 px for the same pair), so it is a race, not a constant.

Changing only the delay to `450` makes both downward cases land at exactly `top = 52`, so the delay is the lever — but I would not ship that number; it is the same fixed-delay guess your own comment at `config-page.component.ts:488` flags. What matters is the layout-settled moment: scrolling off the collapsible's animation-done, or re-issuing the scroll once heights stop changing.

## 3. Result quality — measured against the real configs

- **No ranking, so the section you named can come last.** `break` returns 4 rows in tab order and the section actually called **"Break Reminder" is 4th**, behind `Schedule › Enable lunch break` and two Focus Mode fields. `reminder` puts "Reminders" 2nd and "Break Reminder" last. A one-line sort (title match before field match, then tab order) fixes it.
- **`help` matches produce rows with no visible reason.** A help hit sets `isMatch`, which suppresses `matchedLabelKey` (`settings-search.util.ts:111`), so the row shows just the section and tab. `sync` returns 7 rows, three of which — Reminders, Focus Mode, Focus Mode (this device) — match only prose inside `help` ("…after the app **syncs**…", "kept in **sync**") and render with nothing explaining why. `help` is also raw HTML, so `strong` returns 4 rows and `href` returns 1. Worth stripping tags before matching, and either showing the reason or not searching `help`.
- **Sound settings aren't actually searchable, but the wiki says they are.** `SOUND_FORM_CFG` renders via `config-sound-form` and is in no tab array, so only the literal word "Sound" finds it: `volume` → 0 hits, `done sound` → 0 hits. The wiki paragraph says "…dark mode, theme, wallpaper, **sound**, plugins, and sync setup — are searchable too", which reads as those *settings* being searchable rather than those six words. Either narrow the sentence or feed `SOUND_FORM_CFG` into the tab-1 haystack. Same sentence issue for the sync dialog (`webdav`, `dropbox`, `password`, `encryption` → 0 hits each).

## 4. The tests still can't catch any of this

The new `props` case is welcome, but it is another synthetic-fixture case: it proves the algorithm reads `props`, which was never in doubt — the r1 bug was that the *real* consts use it. I sabotaged the implementation to check what the suite actually holds:

| sabotage | suite |
|---|---|
| remove `.trim()` from the query guard | **9/9 pass** |
| remove `isTabMatch \|\|` from the `CUSTOM_TARGETS` filter | **9/9 pass** |

The first matters because `'should return nothing for an empty query'` only ever passes `'   '`, never `''` — it is not testing the guard it names. And of the six `CUSTOM_TARGETS`, only `wallpaper` is asserted at all (`spec.ts:89-92`), so the hand-copied `tabIndex` values are unguarded: reorder the `mat-tab-group` and search sends people to the wrong tab with a green suite. There are no tests for `config-page.component.ts` at all, which is where the tab-index mapping lives.

One test that feeds the real `GLOBAL_*_FORM_CONFIG` into `searchSettings` and asserts every `scrollSelector` resolves in the rendered config page would have caught both r1 defects and would keep catching the next tab added without `section-{{ section.key }}`. Same ask as r1 — it now has four live defects behind it.

## Smaller things

- **`// ponytail:`** at `config-page.component.ts:488` is still there — the only occurrence in `src/`, and it happens to sit on the comment describing bug 2.
- **`scrollbar-gutter: stable`** does carry a rationale in the SCSS, but it is unrelated to search and unmentioned in the PR description — worth its own commit.
- The Sound custom target sorts above the Tasks tab's sections but renders below them.

## CI

There are **no check runs at all** on `b6accfd4ee` — `41c991702a` was green (15 success / 4 skipped), this commit has not been built or tested. Worth a nudge before merge.

## Still good

The shape remains right: search above the tabs, results in an overlay, no settings forms built while searching, platform filtering inherited for free from `filterGlobalConfigForm`. The `::ng-deep .settings-search-panel` styling is exact precedent (`add-task-bar.component.scss:412`) — no objection there. All CSS custom properties used are real, both new i18n keys are in `en.json` only and resolve, no new dependencies, and the search cost is bounded and linear per keystroke.

johannesjo:
Thank you very much! <3

Issue #5747, opened 2025-12-16
💡 Implement syncing of multiple projects in Markdown


kalingibbons:
### Problem Statement

I am only able to sync a single project with Obsidian using the markdown sync plugin, but I would like to be able to sync more. I'm not an electron developer, so I had copilot locate relevant files for me and propose a possible solution.

### Possible Solution

Based on my analysis of the sync-md plugin code, here's a comprehensive breakdown of the changes needed to support syncing **multiple projects**:

## Summary of Required Updates

To enable multi-project sync, you need to change from a **single config** model to a **multi-config array** model.  Here are the key changes:

---

### Current Model (`local-config.ts`):
```typescript
export interface LocalUserCfg {
  filePath: string;
  projectId: string;
}
```

### New Model:
```typescript
export interface ProjectSyncConfig {
  filePath: string;
  projectId: string;
  enabled: boolean;  // Allow disabling individual project syncs
}

export interface MultiProjectConfig {
  projects: ProjectSyncConfig[];
  version: number;  // For future migrations
}
```

### Update Storage Functions: 

````typescript name=local-config.ts url=https://github.com/johannesjo/super-productivity/blob/main/packages/plugin-dev/sync-md/src/background/local-config.ts
export interface ProjectSyncConfig {
  filePath: string;
  projectId: string;
  enabled: boolean;
}

export interface MultiProjectConfig {
  projects: ProjectSyncConfig[];
  version: number;
}

const LOCAL_STORAGE_KEY = 'sync-md-config';
const MULTI_PROJECT_STORAGE_KEY = 'sync-md-multi-config';

// Legacy:  Load old single-project config
export const loadLocalConfig = (): ProjectSyncConfig | null => {
  try {
    const savedData = localStorage.getItem(LOCAL_STORAGE_KEY);
    if (savedData) {
      return JSON.parse(savedData) as ProjectSyncConfig;
    }
    return null;
  } catch (error) {
    console.error('[sync-md] Failed to load config:', error);
    return null;
  }
};

// New: Load multi-project config with migration
export const loadMultiProjectConfig = (): MultiProjectConfig => {
  try {
    const savedData = localStorage.getItem(MULTI_PROJECT_STORAGE_KEY);
    if (savedData) {
      return JSON.parse(savedData) as MultiProjectConfig;
    }
    
    // Migration:  Check for legacy single-project config
    const legacyConfig = loadLocalConfig();
    if (legacyConfig) {
[...]

PR #9323, opened 2026-07-27
docs: Add sync.md Multi plugin to community plugins list #5747

fpindado:
## Problem

Super Productivity's built-in sync.md syncs a single project with one Markdown file. There is no built-in way to sync multiple projects, each with its own `.md` file — the request in #5747.

## Solution

Lists **sync.md Multi** in the community plugins so users can find and install it. It's an external plugin that bidirectionally syncs multiple SP projects with their own Markdown files (each project maps to its own `.md`), extending the built-in single-project sync.md.

- Source & README (EN/ES): https://codeberg.org/fpindado/sync-md-multi
- Latest release (installable ZIP): https://codeberg.org/fpindado/sync-md-multi/releases/tag/v0.4.0
- Verified working on Super Productivity 18.13.1–18.16 (macOS) as an uploaded plugin (nodeExecution consent).

Refs #5747 for users today, with no change to Super Productivity itself.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki) — this PR *is* a docs/listing change (`community-plugins.json`); no wiki changes needed.
- [ ] I have run `npm run checkFile` on changed `.ts`/`.scss` files — N/A, no `.ts`/`.scss` changed (data-only JSON entry).
- [ ] I have added tests for my changes (if applicable) — N/A, data-only change.
- [x] Existing tests still pass — unaffected (no code changed).

### DIFF
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -134,5 +134,13 @@
     "author": "CuriousChipmunk",
     "authorUrl": "https://github.com/CuriousChipmunk",
     "stars": 0
+  },
+  {
+    "name": "sync.md Multi",
+    "shortDescription": "Sync multiple SuperProductivity projects with their own markdown files (bidirectional).",
+    "url": "https://codeberg.org/fpindado/sync-md-multi",
+    "author": "Fernando Pindado",
+    "authorUrl": "https://codeberg.org/fpindado",
+    "stars": 0
   }
 ]


johannesjo:
Thanks for adding this. I rechecked the eight-line catalog entry and audited the exact v0.2.2 tag/release linked from the PR. The catalog JSON itself is valid, the installable release exists, and the zero-star count is current. However, I found three material blockers in the released plugin, so I do not think v0.2.2 is safe to list yet.

### Important: deleting a parent while promoting its child deletes both tasks

Reproduction:

1. In SP, parent P has child C (P.subTaskIds = [C], C.parentId = P).
2. Edit the Markdown so P is gone and C remains as a root task: <code>- [ ] &lt;!--C--&gt; Child</code>.
3. Import the file.

The plugin [parent-change and deletion passes](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.2.2/src/background/sync/generate-task-operations.ts#L199-L218) generate, in order: update C.parentId to null, update P.subTaskIds to [], delete P, then reorder to [C]. Super Productivity [batch reducer](https://github.com/super-productivity/super-productivity/blob/af1e830d77676e9763421195f419907d528115de/src/app/root-store/meta/task-shared-meta-reducers/task-batch-update.reducer.ts#L262-L329) processes updates as pending changes, but the delete step reads P.subTaskIds from the pre-update state and recursively queues C for deletion. I ran those exact generated operations through the real reducer; the resulting task state and project task list were both empty.

Please fix this plugin/host interaction so a child promoted in the same batch cannot be recursively deleted, and add a regression test that applies the generated operations through the actual SP batch reducer. The generator-only tests do not expose this failure.

### Important: Sync Now can overwrite a just-edited Markdown task before importing it

The watcher polls mtime every two seconds. If a user edits an ID-bearing task line and clicks **Sync Now** before the next poll, [syncAllNow()](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.2.2/src/background/sync/sync-manager.ts#L637-L667) stops the watcher/cancels pending import work, runs SP→Markdown first, and only then reads the file for Markdown→SP. The SP→Markdown merge [replaces the checkbox and title of every known ID-bearing line](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.2.2/src/background/sync/sp-to-md.ts#L255-L275), so the manual edit is already gone before it is read. I reproduced this directly with the shipped merge routine: <code>- [x] &lt;!--A--&gt; Markdown edit &#64;done</code> became <code>- [ ] &lt;!--A--&gt; SP title</code>.

Please detect/import the changed file before the first write, or use a conflict-safe merge, and add a regression test for edit → immediate Sync Now before the watcher poll. The README also needs one consistent direction; it currently describes both SP→MD-first and MD→SP-first behavior.

### Important: raw innerHTML turns project titles and file names into script execution

The config UI inserts project titles, configured paths, timestamps, and binding IDs into [innerHTML](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.2.2/src/ui/index.html#L293-L307), and the file browser inserts raw filesystem entry names into [innerHTML](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.2.2/src/ui/index.html#L563-L567). Event-handler markup in any of those values executes in the plugin iframe: SP allows inline scripts, the iframe has allow-scripts, and its bridge exposes [executeNodeScript](https://github.com/super-productivity/super-productivity/blob/af1e830d77676e9763421195f419907d528115de/src/app/plugins/util/plugin-iframe.util.ts#L430-L470). Because this plugin requires nodeExecution, after the user has approved that permission the injection can escalate to arbitrary code with user-level machine access.

Please render every dynamic value with textContent/DOM properties and attach handlers with addEventListener closures; error messages should use the same safe path. A UI regression test with HTML-shaped project and file names would be useful.

Once these are fixed, please publish a new release and update the PR release/verification details before merging the listing.

One separate bookkeeping note: please change “This resolves #5747” to “Refs #5747” unless a maintainer explicitly decides that a desktop-only external plugin fulfills the upstream feature request. Previous community-plugin listings did not close corresponding built-in feature requests, and the discussion on #5747 still distinguishes listing this plugin from bundling or folding the capability into the official sync.md plugin.

CI note: this PR currently has only the passing first-contributor welcome job; no substantive required checks are reported.

**Verdict: needs changes before listing v0.2.2.**

fpindado:
Thank you — this is an exceptionally thorough review, and every point lands. I've reproduced all three against the shipped v0.2.2 and agree they're blockers. I'm getting to work on them now:

1. **Parent deletion cascading to a promoted child.** You're right that the fix belongs at the plugin/host boundary, not just the generator. I'll stop relying on batch-internal ordering and apply the deletes in a separate step, after the re-parenting updates have been committed, so a child promoted in the same sync can't be recursively removed. I'll add a regression test that drives the generated operations through the batch-apply path (modelling the reducer's cascade), since the generator-only tests don't catch it.
2. **Sync Now clobbering a fresh Markdown edit.** Correct — SP→MD-first assumes the watcher already imported the edit, which fails in the edit-then-immediate-Sync-Now race. I'll import the changed file before the first write (conflict-safe), add a regression test for that exact race, and fix the README to describe a single, consistent direction.
3. **innerHTML injection escalating to code execution.** Agreed, and thank you for flagging the nodeExecution escalation explicitly. I'll render every dynamic value (project titles, paths, timestamps, binding IDs, file-browser entry names, error messages) via textContent/DOM properties and attach handlers with addEventListener closures — no dynamic innerHTML anywhere — with a UI regression test using HTML-shaped project and file names.

I'll publish a new release with these fixes and update the PR's release/verification details before asking for re-review. I've also changed the PR reference to Refs #5747 — listing the plugin isn't the same as resolving the upstream feature request, and that distinction should stay with the maintainers.

Thanks again for taking the time to audit this so carefully.

fpindado:
All three points are fixed and released as **v0.3.0** (both the source repo and an installable ZIP):
https://codeberg.org/fpindado/sync-md-multi/releases/tag/v0.3.0

- **Config-UI HTML injection** — project titles and file paths are now rendered with `textContent` + `addEventListener` instead of `innerHTML`; no dynamic value is parsed as markup. (The plugin's own `openDialog` dialogs already HTML-escaped titles/paths, so those are covered too.) Added a UI regression test that feeds HTML-shaped names and asserts nothing executes.
- **Sync Now overwriting a fresh edit** — Sync Now now imports Markdown → SP first whenever the file changed since the last sync, then writes back; if the file is unchanged it writes back first. A just-made edit can no longer be clobbered. README updated to a single, consistent direction, and a regression test added.
- **Parent delete cascading onto a promoted child** — deletes now run in a separate batch that commits *after* the re-parenting updates, so the parent's `subTaskIds` is already empty when it's deleted and the child survives. Added a regression test that models SP's cascading reducer.

`tsc` clean, 336/336 tests, verified in SP 18.13.1–18.15 on macOS. The `community-plugins.json` entry is unchanged (it points at the repo root). Thanks again for the review!

johannesjo (changes requested):
I rechecked the current PR head and the exact v0.3.0 source tag and released ZIP. The eight-line catalog entry itself is valid, and v0.3.0 appears to fix the previously reported dynamic-HTML execution issue. However, the released plugin still has five deterministic task/file-loss paths. The existing community-plugin and `nodeExecution` warnings cover unknown third-party risk, but they cannot mitigate known destructive behavior, so I think the listing should wait for another upstream release.

### Important

1. **Keep filesystem bindings device-local.** The plugin persists absolute paths, enabled bindings, and `_lastSynced` through [`persistDataSynced`](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.0/src/background/background.ts#L38-L43), then [starts/restarts sessions from received bindings](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.0/src/background/background.ts#L237-L267). The host deliberately syncs this storage and fires `persistedDataChanged` for [remote changes](https://github.com/super-productivity/super-productivity/blob/af1e830d77676e9763421195f419907d528115de/docs/wiki/3.01-API.md#L173-L177). With the plugin enabled on two synced desktops, A's absolute path can therefore activate on B. If that path exists on B with unrelated checkbox content, initial sync treats B's file as authoritative and deletes project tasks absent from it; those deletes can sync back to A. The smallest upstream fix is to reuse the plugin's existing local configuration storage and route iframe config reads/writes through the background. Do not auto-start a session from a remotely synced absolute path; add one two-device regression.

2. **Abort imports on parse errors before generating operations.** A line such as `- [ ] <!--A-->` is [reported as an empty-title parse error and skipped](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.0/src/background/sync/markdown-parser.ts#L125-L135), but [`mdToSp`](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.0/src/background/sync/md-to-sp.ts#L33-L60) only shows a warning and continues. Reconciliation then [deletes every SP task not present in the parsed set](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.0/src/background/sync/generate-task-operations.ts#L343-L388), so task A is deleted despite the warning. The minimal fix is to return/throw immediately when the parser already reports an error, with one regression asserting no batch call occurs.

3. **Canonicalize file targets at the actual save boundary.** The UI guard compares [raw path strings](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.0/src/ui/index.html#L338-L357), while background I/O applies [`path.resolve`](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.0/src/background/helper/file-utils.ts#L47-L74). `/tmp/tasks.md` and `/tmp/./tasks.md` therefore pass validation but start two project sessions against one file. Each session can import the other project's IDs, delete unmatched tasks, and rewrite the file. Route saves through the background, normalize with `path.resolve` plus `realpath` when the target exists, reject canonical duplicates there, and make registry startup refuse legacy duplicates as a backstop. No new setting or inode abstraction is needed.

4. **Do not use mtime ordering as the destructive direction switch.** Manual Sync Now only treats a file as changed when [`mtime > lastSyncedMtime`](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.0/src/background/sync/sync-manager.ts#L680-L704). If an edit/replacement preserves an equal or older timestamp and the user presses Sync Now before the watcher imports it, the plugin chooses SP→MD first and overwrites the edit before reading it. Changing `>` to `!==` still misses equal timestamps. The smallest complete fix is one in-memory last-synced content fingerprint checked before export, with equal- and older-mtime regressions; no conflict UI or persisted history is necessary.

5. **Parent promotion and deletion must remain one replay-atomic operation.** v0.3.0 applies [non-deletes and deletes in two API calls](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.0/src/background/sync/md-to-sp.ts#L62-L78), and each call becomes a separate persistent `PROJECT` batch operation. A deterministic two-device case is: A's promotion `{A:1}, t=1000`; B's concurrent project batch `{B:1}, t=2000`; A's delete `{A:2}, t=3000`. Conflict detection handles [each remote operation separately](https://github.com/super-productivity/super-productivity/blob/af1e830d77676e9763421195f419907d528115de/src/app/op-log/sync/conflict-resolution.service.ts#L3901-L3907), so B rejects the older promotion but accepts the newer delete. That delete sees stale `P.subTaskIds = [C]` and the [host cascade removes C](https://github.com/super-productivity/super-productivity/blob/af1e830d77676e9763421195f419907d528115de/src/app/root-store/meta/task-shared-meta-reducers/task-batch-update.reducer.ts#L262-L329). Fix the host reducer first so cascade uses effective pending `subTaskIds`, cover it with the real reducer and the exact two-client op-log case, then publish a plugin release that restores one batch. Reverting the plugin before the host fix would reintroduce the local bug.

### Minor

- The PR body still calls v0.2.2 the latest release. Please update the release and verification details after publishing the corrected artifact.

**Verdict: needs changes before listing the plugin.**

fpindado:
Thanks for the detailed second review — really appreciate you catching these.

I've shipped **v0.3.1**, which fixes the three points that are in the plugin's own control:
- **(2) Abort import on parse errors.** `mdToSp` now returns early *before* generating any operations when the parser reports errors, so a half-typed line like `- [ ] <!--A-->` can no longer make reconciliation delete the real task. All sync paths also skip the SP→MD write-back on a parse error, so SP can't rewrite the old title back into the file after the abort. (This one affects single-device use too.)
- **(3) Canonicalize file targets.** Paths are now canonicalized (`.`/`..`/`//` collapsed, separators unified, `path.resolve` + `realpath` where the file exists) before the duplicate check, and the registry refuses two bindings that resolve to the same file as a backstop — so `/tmp/tasks.md` and `/tmp/./tasks.md` can no longer start two sessions on one file.
- **(4) Content fingerprint instead of mtime.** *Sync Now* now decides its direction from an in-memory last-synced content hash rather than mtime ordering, so an edit that preserves an equal/older timestamp is no longer overwritten.

Two points remain, and I want to be upfront that they're **not** in v0.3.1:
- **(1) Device-local config.** The fix is to stop persisting bindings through the cross-device synced store and keep them device-local — mirroring how the built-in `sync.md` uses `localStorage` + a UI bridge. It's a config-storage rework, so I've scoped it to **v0.4.0** rather than rush it into this release.
- **(5) Promotion + deletion atomicity.** As you noted, the two-batch fix trades the local cascade bug for a multi-device split (older promotion rejected, newer delete accepted → host cascade removes the child). The right fix is host-side first — the reducer's cascade should use the *effective pending* `subTaskIds` — after which the plugin can safely revert to a single batch. Single-device users are unaffected by the current interim.

Release + notes: https://codeberg.org/fpindado/sync-md-multi/releases/tag/v0.3.1

johannesjo (changes requested):
Third pass, this time against the **v0.3.1** tag (`3c8a8fa`) and its released ZIP. Thanks for the fast turnaround and for being explicit about what did *not* make the release — that made this round much quicker.

Two things I checked that are worth stating up front, because they raise my confidence in everything else:

- **The shipped ZIP is a reproducible build.** I ran `node scripts/build.js` from the clean `v0.3.1` tag and diffed the output against `sync-md-multi-v0.3.1.zip` — `plugin.js`, `index.html`, `manifest.json` and `icon.svg` are all byte-identical. So auditing the repo really does audit what users install.
- **The new regression tests are not vacuous.** I sabotaged each fix and confirmed the intended spec goes red. Baseline is 363 tests / 43 suites green; no sabotage left the suite green. Mapping:
  - dropping `return 'parse-error'` in `mdToSp` → `abort-on-parse-error` (1 test);
  - neutralising the caller-side `if (result === 'parse-error') return;` → `parse-error-no-writeback` (1 test, exactly);
  - raw path strings instead of canonical keys → `dedupe-bindings-by-path` (2 tests);
  - forcing `fileHasUnsyncedEdit` to `false` → `file-fingerprint-direction` + `sync-now-preserves-edit` (and, legitimately, `parse-error-no-writeback`, since the `else` branch then calls `spToMd`).

  One small gap worth noting: the two halves of the parse-error fix are each covered on their own, but no spec runs the real `mdToSp` through the real `syncAllNow` — they are coupled only by the literal string `'parse-error'`. Low risk, since both sides assert that literal, but the seam itself is untested.

### Fixed (points 2, 3, 4)

- **(2) Parse-error abort** — [`mdToSp` now returns `'parse-error'` before generating any operations](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.1/src/background/sync/md-to-sp.ts#L53-L62), and the three sync call sites (initial sync L518, `executeMdToSpSync` L588, `syncAllNow` L814) skip the SP→MD write-back on that result. Correct — with one gap I'd fix while you're in there: the *Recreate project* handler at [L408-L410](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.1/src/background/sync/sync-manager.ts#L408-L410) ignores the return value and calls `spToMd` + `markSynced` unconditionally, so recreating a project from a malformed file still writes back over it.
- **(3) Canonical file targets** — the save boundary now resolves via [`path.resolve` + `realpathSync` in Node](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.1/src/ui/index.html#L361-L393) with a lexical fallback, and [`dedupeBindingsByCanonicalPath`](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.1/src/background/sync/sync-manager.ts#L159-L180) is the registry backstop for legacy configs. Correct.
- **(4) Content fingerprint instead of mtime** — [`fileHasUnsyncedEdit`](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.1/src/background/sync/sync-manager.ts#L129-L133) is pure, treats a null baseline as "changed", and `syncAllNow` uses it. Correct *for `syncAllNow`* — see the new finding below.

### Still open, as you said (points 1 and 5)

- **(1) Device-local config** — confirmed still live, and the exposure is a bit wider than my last review described. [`persistConfig()` writes the whole config, bindings included, through `persistDataSynced`](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.1/src/background/background.ts#L38-L47), which lands as a `PLUGIN_USER_DATA` op-log entity — a first-class synced entity, no allowlist keeps it local. Sessions then start from whatever comes back, both [live on remote arrival](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.1/src/background/background.ts#L94-L125) *and*, more directly, [at plugin boot](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.1/src/background/background.ts#L239-L248) before the hook or the 5 s poll is even registered. There is no device-id or origin check anywhere in the plugin.

  I ran device B's side through the real `initSyncRegistry → performInitialSync → mdToSp → generateTaskOperations` chain. Two corrections to how I put it last time, one softening and one hardening:

  - **Softening:** if A's path is *missing* on B, you correctly show the missing-file dialog and issue zero operations. That case is fine. And the deletes are scoped to the bound project — a task in another project is untouched.
  - **Hardening:** "an unrelated file at the same path" is the *unlikely* variant. The likely one is that the path collides precisely **because** the `.md` is itself file-synced (Dropbox/Syncthing/iCloud), and B's copy is **stale** — that transport and SuperSync are independent with no ordering between them. B then treats the stale file as authoritative and deletes every task in the bound project created since that snapshot, and those deletes propagate back to A. That needs no path coincidence at all; it's the plugin's normal two-device deployment. My run against a grocery-list file at A's path emitted `delete` for all four bound-project tasks (`realTask1`, `realTask2`, `realParent`, `realChild`) while leaving the cross-project task alone. A control where B's file *is* the current SP state emits zero deletes, as you'd hope.

  Same mechanism, one more consequence worth knowing about while you design v0.4.0: the config is a single LWW blob, so a device that boots before receiving A's config writes `bindings: []` via `refreshProjectsCache → persistConfig`. If that write wins LWW, A's `checkForConfigChanges` adopts the empty array and disposes every session — a plausible "my bindings vanished" report. (That one I reasoned from the code rather than executed.)

  This is one of the two I'd still hold the listing on — see the verdict.
- **(5) Promotion/deletion atomicity** — confirmed still the [two-batch interim](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.1/src/background/sync/md-to-sp.ts#L90-L94). I agree with your split, and I can now be more definite than last time about *why*. I drove the real host reducer with the promotion and the delete in a single batch, in three arrangements — including the one where the caller explicitly sends `update P {subTaskIds: []}` alongside the delete. **Master cascade-deletes the child in every single-batch arrangement.** So there is no in-batch workaround available to a plugin: your two-batch split is the only client-side fix there is, and reverting it before the host is fixed would definitely reintroduce the local bug.

  On the host side the fix is small but not as small as I implied last round. The reducer already consults `pendingUpdatesMap` a few lines earlier when [removing the deleted task from its parent's `subTaskIds`](https://github.com/super-productivity/super-productivity/blob/96e002c8a23a882a6615a5b62d56da78b942d1c2/src/app/root-store/meta/task-shared-meta-reducers/task-batch-update.reducer.ts#L278-L286), then reads the pre-batch value for [the cascade](https://github.com/super-productivity/super-productivity/blob/96e002c8a23a882a6615a5b62d56da78b942d1c2/src/app/root-store/meta/task-shared-meta-reducers/task-batch-update.reducer.ts#L289-L292). Making the cascade read the effective-pending value fixes the ordering you hit, but it's ~3–4 lines and still order-dependent: `sortedOperations` only reorders `create` ops, so a `delete` listed *before* its promotion update would still cascade. Your generator always emits the reparent before the delete, so that residue is a contract hole rather than a live break for sync.md. No existing spec covers this — the nearest one exercises the parent-side path that already works. All of that is SP's to do, not yours; I'll track it separately.

### Important (new): the *automatic* SP→MD path has no such guard

Point 4's fingerprint check was added to `syncAllNow` only. [`handleSpChange`](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.1/src/background/sync/sync-manager.ts#L721-L751) — the path an ordinary task edit takes — goes straight to `spToMd` after its 500 ms debounce, without consulting `fileHasUnsyncedEdit`, `readTasksFile`, or the `pendingMdToSpSync` flag that `handleFileChange` sets at L560. `fileHasUnsyncedEdit` has exactly one call site, `syncAllNow` L806.

Ticking a checkbox or renaming a task in SP dispatches `[Task Shared] updateTask`, which reaches both the `taskUpdate` and `anyTaskUpdate` hooks the plugin registers, and so lands in `handleSpChange`. None of the gates in that block help: `SP_HOOK_COOLDOWN_MS` is measured from `lastMdToSpSyncTimestamp`, which is written only inside `executeMdToSpSync` — and the premise here is that the import hasn't run yet, so it's stale by construction. `divertIfFileMissing` is an existence check, not a content check. `verifySyncState` runs but [its result is discarded](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.1/src/background/sync/sync-manager.ts#L740) (contrast `executeMdToSpSync` L599, which branches on it).

**The worst version is the unfocused case, where the plugin already knows about the edit.** When SP isn't focused, [`handleFileChange`](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.1/src/background/sync/sync-manager.ts#L552-L564) sets `pendingMdToSpSync = true` and parks the import for a full **10 s** (`SYNC_DEBOUNCE_MS_MD_TO_SP`). `handleSpChange` never checks that flag, so an SP change in that window overwrites an edit the plugin has already detected and queued. Driving the real registry, the real hooks and the real `spToMd` against an in-memory FS:

```
watcher DETECTED the rename, pendingMdToSpSync = true
SP change 700 ms later →
  file after SP change : "- [x] <!--t1--> Original @done"   (rename destroyed)
  mdToSp never called
parked import fires 10 s later →
  mdToSp finally got   : "- [x] <!--t1--> Original @done"   (reads back the overwrite)
```

Refocusing SP doesn't rescue it either: [`setupWindowFocusTracking`](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.1/src/background/sync/sync-manager.ts#L879-L897) flushes only when `pendingMdToSpSync && mdToSpDebounceTimer` — i.e. only if the watcher has already fired.

With SP focused the window is narrower (roughly one 2 s watcher poll), so this is a race rather than a loss on every edit. Two accuracy notes on that: the watcher compares with `!==`, so a *lowered* mtime is still caught — only an exactly-equal mtime slips past, and then only past the *watcher* (a manual Sync Now would still notice via the content hash). And a line the user typed *without* an ID marker is preserved by the merge; the loss is specific to lines already stamped with `<!--taskId-->`.

What it costs on disk, from the same run:

```
BEFORE: "- [ ] <!--t1--> Original title — WITH MY MANUAL NOTE APPENDED"
AFTER : "- [x] <!--t1--> Original title @done"
```

Title and checkbox both come back from SP state; only the indent and a pre-existing `@done(...)` survive. The edit is gone the moment `writeTasksFile` lands — it was never in SP. What `markSynced` then adds is that it re-reads and rehashes the *overwritten* file, so the divergence is invisible to every later Sync Now and the app never signals that anything was lost.

The README makes the opposite promise for exactly this path: *"**Simultaneous editing is safe**: The oscillation prevention mechanism handles the case where you edit the Markdown file while Super Productivity is open. You do not need to pause SP before editing your `.md` files."* (README L154). That line needs to change with the fix. (L72's "a change you just made is never overwritten" is fine — it's scoped to Sync Now, which v0.3.1 does guard.)

The fix looks like a few lines: in `handleSpChange`, before `spToMd`, run the same `fileHasUnsyncedEdit` check `syncAllNow` already does — and bail early if `pendingMdToSpSync` is set, which is free. Please add a regression that edits the file, fires `taskUpdate`, and asserts the import ran before the write. As a control, the same setup with *Sync Now* instead of an SP change does hand the edited content to `mdToSp` first — so the v0.3.1 guard demonstrably works, just only on the one path it was added to.

(Two asides while you're in there. The new `action` hook is mostly not doing what its comment says: [`isStructuralTaskAction`](https://codeberg.org/fpindado/sync-md-multi/src/tag/v0.3.1/src/background/sync/sync-manager.ts#L717-L719) requires the literal `"add task"` *with a space*, and SP's real action type is `[Task Shared] addTask`, so task-add doesn't match — it's already covered by `anyTaskUpdate` anyway. Conversely it over-matches `removeTasksFromTodayTag`, `removeTagsForAllTasks` and `[Task] Remove time spent`, because "re**move**" contains `move`.

And one thing worth knowing rather than acting on: this finding is a *local* race, but not for the reason I'd have guessed. SP's generic `action` hook has no action-type filter and `bulkApplyOperations` isn't tagged `meta.isRemote`, so your `action` handler **does** get invoked on every remote sync apply. What stops it reaching `handleSpChange` is your own `isStructuralTaskAction` filter — I ran it against the real strings, and `'[OperationLog] Bulk Apply Operations'` and `'[SP_ALL] Load(import) all data'` both return `false` purely because neither contains the substring `task`. That's accidental protection: if SP ever renames that action to something containing "task", your plugin would start firing a full SP→MD write on every remote sync. Worth a narrower filter — matching a known set of action types rather than a substring — when you next touch it.)

### Minor

- Optional, take it or leave it: the catalog `shortDescription` doesn't mention that this is desktop-only. Nothing is broken — SP refuses to load a `nodeExecution` manifest on web/Android and shows *"Node.js execution is only available in the desktop app"* ([`plugin.service.ts#L906-L920`](https://github.com/super-productivity/super-productivity/blob/96e002c8a23a882a6615a5b62d56da78b942d1c2/src/app/plugins/plugin.service.ts#L906-L920)), and your README already says it at L153. It's purely about saving a wasted install from the catalog. No other entry declares a platform constraint either, so this is a suggestion, not a gap.
- While you're near it: the `if (!PluginAPI.executeNodeScript)` guards in `file-utils.ts` and `file-watcher.ts` are dead code — SP binds `executeNodeScript` unconditionally and returns `{success:false}` on non-desktop, so those branches never run. Harmless today; just not the protection they look like.
- Tiny doc nit: README L72 and L140 both describe Sync Now as importing MD→SP first, unconditionally. Since v0.3.1 it only does that when the content fingerprint differs (otherwise SP→MD first). Worth half a sentence next time you touch the direction rule.
- The entry itself is clean: valid JSON, prettier-clean, matches the `CommunityPlugin` interface, no duplicate name or URL, and `stars: 0` is accurate (Codeberg API reports 0). MIT LICENSE present. Merges cleanly with master.

### Verdict

**Three of the five are fixed; the blocker list is down to two.** Concretely:

- **Blocking: (1) device-local config, and the new automatic-write finding above.** Both are silent, deterministic loss of user content, both are entirely within the plugin's control, and both have small fixes — (1) is mostly a matter of promoting the `local-config.ts`/`localStorage` path you already have to the primary store and routing iframe config reads/writes through the background. The catalog's community-plugin warning covers *unknown* third-party risk; it can't stand in for known destructive behaviour.
- **Not blocking: (5).** Your interim is the right call — single-device users are unaffected, and the correct sequencing genuinely starts on SP's side. I'd rather list a plugin carrying a documented interim than have you revert into the local cascade bug.

On the listing question more broadly, so it doesn't get re-litigated: I think this **earns its catalog entry**. #5747 is open, there's no built-in multi-project markdown sync, and a listing is exactly the low-cost route for a niche need — it adds no permanent core complexity, which is the whole point of the catalogue. To be clear though, listing it is not a decision to bundle it or to fold multi-project support into the built-in `sync.md` — that's still the open question on #5747, and I'm not answering it here.

**Verdict: needs changes before listing v0.3.1.**

fpindado:
Thanks again for the thorough third pass — the reproducible-build and test-sabotage checks especially.

**v0.4.0 is out and clears both remaining blockers.** Install ZIP + source: https://codeberg.org/fpindado/sync-md-multi/releases/tag/v0.4.0

**Blocker 1 — configuration is now device-local.** As you suggested, this reuses the built-in `sync.md`'s approach: the config lives in the background context's own `localStorage` (key `sync-md-multi-config`), and the config panel (iframe) talks to the background through a `postMessage` bridge (`getConfig` / `saveConfig` / `getProjects`), with the background as the single writer. Nothing goes through `persistDataSynced` anymore, so bindings/paths no longer circulate across devices — which removes the stale-file / empty-`bindings`-LWW hazards you described. A one-time startup migration seeds the local store from any pre-0.4.0 synced blob and then clears that blob (idempotent, kept for several releases). New regression coverage in `local-config.test.ts` (device-local, no synced-API writes, migration).

**Blocker 2 (the `handleSpChange` gap you found).** The automatic SP→MD path now runs the same `fileHasUnsyncedEdit` guard as `syncAllNow`: it bails while `pendingMdToSpSync` is set, and when the file has an unsynced edit it imports MD→SP first (aborting on a parse error) before writing back — so a fresh `.md` edit in the parked-import window can't be clobbered. Regression: `sp-change-preserves-edit.test.ts` (edits the file, fires `taskUpdate`, asserts the import runs before the write-back). README L154 corrected.

**Minor "while you're in there" findings — all folded in:**
- *Recreate project* now honors `mdToSp`'s `parse-error` result instead of writing back over a malformed file.
- The `action`-hook filter is now a normalized known-set match instead of the substring test — fixes both the `remove*` over-match (`removeTasksFromTodayTag`, etc.) and the latent risk of a remote bulk-apply ever matching.
- Removed the dead `executeNodeScript` guards.
- README (EN/ES): *Sync Now* direction and simultaneous-editing notes corrected.

**Point 5** stays as the documented two-batch interim, per your conclusion — no client-side single-batch fix exists, so it waits on the host-reducer change (effective-pending `subTaskIds`), which I'm happy to help with separately.

Build is reproducible from the `v0.4.0` tag; 384 tests pass. Verified in vivo on SP (macOS): device-local load/save, the migration (`loadSyncedData()` returns `null` afterwards), and the bridge.

---

One last note on the broader #5747 question (whether multi-project Markdown sync should eventually live inside the built-in `sync.md`, or stay an external plugin): I don't think this PR needs that decided. Listing gives users a working answer today, and the plugin is perfectly happy living externally — so that call is entirely yours to make whenever suits you, or not at all. No pressure from my side either way.

johannesjo:
Thank you very much! <3

Issue #8827, opened 2026-07-07
💡 a11y: global reduced-motion support and muted-text contrast pass


johannesjo:
## Problem

Two accessibility concerns that need a token/global-level pass rather than per-component patches:

**Reduced motion:** `prefers-reduced-motion` is honored in ~12 places, against **32 `@keyframes` and ~127 Angular animation sites** (72 `trigger()` + 55 template bindings). The global reduce block (`src/app/app.component.scss:264`) only disables the app-entrance slide; route/expand/fade animations are not gated. A global hook (e.g. disabling Angular animations via `prefersReducedMotion ? NoopAnimationsModule`-style provider, plus a CSS catch-all) would cover most of it at once.

**Text contrast:** themes define muted text at 0.54 alpha (e.g. `src/assets/themes/arc.css:67` — `--text-color-muted: rgba(197,207,217,0.54)`), used app-wide for timestamps/labels, which fails WCAG AA at small sizes in several themes. On top of that there are **87** `opacity: 0.1–0.6` declarations (excluding `disabled` states), many on text (e.g. `task-summary-table.component.scss:48` at 0.5, `task-context-menu-inner.component.scss:6` at 0.4).

## Suggested fix

- Global reduced-motion strategy + keep per-component overrides only for special cases (confetti already does this correctly).
- Raise the muted-text floor in the theme tokens (≥ 0.6–0.7 alpha or solid computed colors) and sweep low-opacity text toward the token.

---
*From a full design/UX audit (Playwright visual tour + code audit), 2026-07. Counts grep-verified.*

MarkQuach12:
Hi @johannesjo, am I able to handle this issue?

johannesjo:
That would be great  @MarkQuach12 ! Since we already have a setting to disable animations, this is not a high priority item. So if the required changes are plenty or risky of breaking something, I'd probably not do at for the time being.

Issue #8689, opened 2026-07-01
a11y: respect prefers-reduced-motion globally


johannesjo:
## Problem

Only 7 SCSS files handle `prefers-reduced-motion` (swipe-block, mobile-bottom-nav, onboarding, app shell, planner-calendar-nav). Task-list animations, side-nav transitions, and the **infinite** header sync spinner (`src/app/core-ui/main-header/main-header.component.scss`) all ignore the OS setting — only the in-app "disable animations" toggle covers them.

## Proposal

The global kill switch already exists: `body.isDisableAnimations` (`src/styles.scss`). Hook the OS preference into the same path — either:

- (a) CSS: extend the `body.isDisableAnimations` rules with a matching `@media (prefers-reduced-motion: reduce)` block, or
- (b) TS: have `GlobalThemeService` treat `matchMedia('(prefers-reduced-motion: reduce)')` as defaulting the existing setting on (user can still override in settings).

Option (b) keeps one code path and also disables Angular animations, so it's likely the leaner fix. The sync spinner may deserve a reduced (opacity pulse) fallback rather than freezing entirely, since it conveys ongoing state.

## Acceptance criteria

- With the OS reduced-motion preference on, no infinite or large-scale animations play (task list, side nav, page transitions, sync spinner).
- The explicit in-app setting still wins in both directions.

PR #9325, opened 2026-07-27
feat(a11y): honor OS reduced-motion setting for animations

MarkQuach12:
## Problem

Addresses the reduced-motion portion of #8827, and the "setting wins in both directions" question from #8689.

The app already has a full animation kill-switch behind the "Disable all animations" setting: `@.disabled` in `app.component.ts` for Angular animations, `MATERIAL_ANIMATIONS` in `main.ts` for Material, and the `body.isDisableAnimations` catch-all in `styles.scss` for CSS keyframes.

None of them read the OS `prefers-reduced-motion` preference, so a user who has asked their system to reduce motion still gets every animation until they find the in-app toggle. `ConfettiService` was the one exception.

## Solution

Add a `prefersReducedMotion` signal backed by `matchMedia`, and OR it into all three kill-switch reads. `ConfettiService` now uses the same signal instead of its own inline `matchMedia` call.

No new UI, no new setting, no per-component changes.

## Image lightbox fix

`EnlargeImgDirective` removed the full-viewport lightbox only from inside a `transitionend` listener. With transitions off that event never fires, so the overlay stayed in the DOM and swallowed every pointer and touch event, with no recovery short of a page reload.

This is not new here. `body.isDisableAnimations` is already set by the existing in-app checkbox on master, so the wedge is reachable today on every shipped version by anyone who has turned animations off. What this PR changes is the population that hits it, from people who opted in to everyone whose OS reports reduced motion.

The fix is a small `_afterTransition(el, fn, ms = 500)` helper that runs the cleanup on `transitionend` or after a timeout, whichever comes first, applied at all three sites: `_hideImg()` plus both zoom-out branches. `_hideImg()` now also captures the wrapper element up front, so reopening the lightbox inside the fallback window cannot remove the new wrapper.

Two regression tests cover the backdrop-click and zoom-out paths with `body.isDisableAnimations` set. Both fail against the unmodified directive with a timeout, and the spec asserts the real `transition: none` is in effect rather than mocking it.

Happy to split this into its own PR against master if you would rather land it first, as offered in review.

## #8689 setting visibility

Kept the hard `||` and added a `description` hint to the `isDisableAnimations` field, per the fallback offered in review. The wiki now also states that the checkbox reads unchecked while the OS preference is on, that unticking it has no visible effect in that state, and that the system setting is the way back.

## Not included

The issue also asked for a muted-text contrast pass. I left that out: the numbers it cites are stale (`arc.css` is `0.74` now, not `0.54`), and `tools/theme-assets.test.js` already asserts muted text meets contrast on theme surfaces. Happy to open that separately.

The `initPrefersReducedMotion()` test-seam refactor is also left out, since nothing is broken by the current module-level read today.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo (commented):
## Summary

The implementation itself looks sound and appropriately small. The root @.disabled binding, body class, Material token, and confetti check cover distinct animation systems, so the shared signal is a good KISS solution. I also verified the exact head in a browser: an initial reduced-motion preference adds body.isDisableAnimations, and a live switch back to no-preference removes it.

## Issues

### Minor

- The PR body says Closes #8827, but this PR intentionally omits the muted-text/low-opacity half of that issue. Please change this to “Addresses the reduced-motion portion of #8827” (or “Refs #8827”) so merging does not mark the remaining work complete. There is no need to expand this PR into the contrast work.

## Verdict

The code review is clean. Please correct the closing keyword; GitHub currently reports no CI checks for this head, so those still need to run and pass before merge.

MarkQuach12:
> ## Summary
> The implementation itself looks sound and appropriately small. The root @.disabled binding, body class, Material token, and confetti check cover distinct animation systems, so the shared signal is a good KISS solution. I also verified the exact head in a browser: an initial reduced-motion preference adds body.isDisableAnimations, and a live switch back to no-preference removes it.
> 
> ## Issues
> ### Minor
> * The PR body says Closes [💡 a11y: global reduced-motion support and muted-text contrast pass #8827](https://github.com/super-productivity/super-productivity/issues/8827), but this PR intentionally omits the muted-text/low-opacity half of that issue. Please change this to “Addresses the reduced-motion portion of [💡 a11y: global reduced-motion support and muted-text contrast pass #8827](https://github.com/super-productivity/super-productivity/issues/8827)” (or “Refs [💡 a11y: global reduced-motion support and muted-text contrast pass #8827](https://github.com/super-productivity/super-productivity/issues/8827)”) so merging does not mark the remaining work complete. There is no need to expand this PR into the contrast work.
> 
> ## Verdict
> The code review is clean. Please correct the closing keyword; GitHub currently reports no CI checks for this head, so those still need to run and pass before merge.

Ok should be fixed now

johannesjo (commented):
## Summary

Round 2. The PR-body fix is in — thanks. The shared signal is still the right shape for this, and I don't want to grow the diff much.

Two things need addressing before merge, and one of them actively hurts the users this PR is for. Process note first: **CI has still never run on this head.** All 8 workflow runs on `9d14f26` sit at `conclusion: action_required` (fork PR waiting on maintainer approval), `check-runs` is 0, and `mergeStateStatus: BLOCKED`. That one's on me, not you.

## Issues

### 🔴 Blocking — this makes an existing wedge reachable for every reduced-motion user

`GlobalThemeService` now adds `body.isDisableAnimations` whenever the OS reports reduce-motion, and that class sets `* { transition: none !important }` (`src/styles.scss:32-39`).

`EnlargeImgDirective._hideImg()` (`src/app/ui/enlarge-img/enlarge-img.directive.ts:40-53`) removes the lightbox **only** from inside a `transitionend` listener. With transitions off that event never fires, so `.enlarged-image-wrapper` — `position: fixed; inset: 0; width/height: 100%`, no `pointer-events` escape (`src/styles/components/enlarge-img.scss:5-17`) — stays in the DOM and **swallows every pointer/touch event across the whole viewport**.

Reproduced in real Chrome against the app's real `styles.scss` (the karma builder loads it via `angular.json` → `test.options.styles`), with a positive control, then re-reproduced independently with a second test written from scratch:

| | lightbox removed? |
|---|---|
| animations on (control) | ✅ removed |
| `body.isDisableAnimations` | ❌ still in the DOM after 6s |

While stuck: `getComputedStyle(img).transitionProperty === 'none'`, `getComputedStyle(wrapper).pointerEvents === 'auto'`, and a real button placed under the overlay never receives the click (`elementFromPoint` hits the wrapper; the button's click counter stays at 0, vs 1 in the control).

**There is no recovery short of a page reload** — verified, not assumed: further clicks don't dismiss it, `Escape` does nothing, and `fixture.destroy()` (≈ route teardown) leaves it parented to `document.body`. The directive has no `ngOnDestroy`, and the wrapper is appended outside the router outlet.

Reachable from `src/app/features/note/note/note.component.html:9` and `src/app/features/tasks/task-attachment/task-attachment-list/task-attachment-list.component.html:39`.

**Before this PR, an OS reduce-motion preference did not trigger any of this.** None of the 8 existing `@media (prefers-reduced-motion: reduce)` blocks in `src/` is a global `transition: none` — they're all scoped to single components — and browsers don't disable transitions on their own. The new line in `global-theme.service.ts` is genuinely the new causal link, and it moves this from "people who deliberately ticked a checkbox" to **everyone whose OS reports reduce-motion** — precisely the accessibility cohort the PR exists for — the first time they open a note image or an image attachment and close it.

`_hideImg()` isn't the only site: the zoom-out branches at lines **111-115** and **125-129** gate `_hideImg()` behind a `transitionend` on `newImageEl`, so on that path `_hideImg()` is never reached at all and any fallback inside it never gets scheduled. That path is also the *visibly* broken one — the wrapper stays on `ani-enter`, so the `:after` backdrop sits at full opacity, whereas the backdrop-click path fades it to 0 and leaves an invisible full-viewport blocker. All three sites need the same treatment; I confirmed a fallback added to `_hideImg()` alone still leaves the zoom path wedged. The precedent is already in the codebase at `src/app/core-ui/mobile-bottom-nav/mobile-bottom-nav.component.ts:81-83`:

```ts
// Safety fallback: clear if animationend doesn't fire
// (e.g. prefers-reduced-motion: reduce → animation: none).
```

Sketch — one helper, used at all three sites:

```ts
/** `transitionend` never fires when transitions are off
 *  (body.isDisableAnimations / prefers-reduced-motion → transition: none). */
private _afterTransition(el: HTMLElement, fn: () => void, ms = 500): void {
  let isDone = false;
  const run = (): void => {
    if (!isDone) {
      isDone = true;
      fn();
    }
  };
  el.addEventListener('transitionend', run, { once: true });
  setTimeout(run, ms);
}
```

Two details worth keeping: capture the element in a local rather than reading `this.enlargedImgWrapperEl` inside the callback (otherwise reopening the lightbox within the fallback window makes the stale timer delete the *new* wrapper), and 500ms is comfortably clear of `--transition-leave` = 225ms (`src/styles/_css-variables.scss:279,291`). Please add a regression test that fails without the fallback.

I'm fine with this landing as a separate PR merged first — I just don't want to ship the reduced-motion default while this is live.

### 🟠 The in-app setting has to win in both directions (#8689)

There's an older, more specific issue for exactly this: **#8689 "a11y: respect prefers-reduced-motion globally"**. Its proposal (b) is what you built, but with one constraint:

> have `GlobalThemeService` treat `matchMedia('(prefers-reduced-motion: reduce)')` as **defaulting** the existing setting on (user can still override in settings)

> **Acceptance criteria** — The explicit in-app setting still wins in both directions.

A plain `||` only lets the setting win in one direction, and the wiki text added here codifies that: *"This is forced on, regardless of the setting."* A user with OS reduce-motion on has no way back, and the settings checkbox shows **unchecked** while every animation is off — toggling it appears to do nothing.

I'd rather not add a tri-state toggle for this (that's the kind of setting the manifesto tells us to avoid), so I'm genuinely open on the resolution. Simplest option that satisfies #8689: seed the stored `isDisableAnimations` default from the OS preference on first run and keep the setting authoritative afterwards. If you'd rather keep the hard `||`, then at minimum give the field a `description` hint explaining the OS override — `src/app/features/config/form-cfgs/misc-settings-form.const.ts:90-96` has label only, and `isVerticalActionBar` two fields down is the pattern to copy (needs a new `en.json` key). Please also reference #8689 in the PR body.

### 🟡 The module-level snapshot bakes the host OS setting into the test bundle

`prefers-reduced-motion.ts` calls `matchMedia` at **module-eval** time and closes over it in module-level `const`s, so there is no seam. `focus-mode-main.component.spec.ts:40-44` (PR head) documents why that matters:

> macOS/Windows GitHub runners report `prefers-reduced-motion: reduce`, which makes the component skip the rocket animation and broke the release build (run 28593603470). Linux reports false, hiding the non-hermetic dependency locally.

That's not hypothetical — run 28593603470 is `Build All & Release` on `v18.13.0`, and the two failing jobs are `mac-bin` and `windows-bin`, both on the `Test Unit` step.

The workaround the repo actually uses does **not** work against the new util. With Chrome launched `--force-prefers-reduced-motion`, the signal reads `true` and stays `true` under four separate stubbing strategies, including the exact `globalThis.matchMedia` override from `focus-mode-main.component.spec.ts`. The one thing that does work is stubbing and then `await import()`-ing the module inside the `it` — but only in a spec that is the module's *sole* importer, which stops being true the moment `app.component` / `global-theme.service` / `confetti.service` import it, and this suite runs specs in random order.

**To be clear about severity: I ran the full unit suite locally with the PR applied and Chrome forced to reduced motion, and it passed.** So nothing is broken today — this is about the next spec someone writes. Note the trap: `ci.yml`'s `test-on-linux` is ubuntu-only, and the macOS/Windows unit lanes in `build.yml` (`mac-bin:199-252`, `windows-bin:284-339`) are gated `if: startsWith(github.ref, 'refs/tags/v')`. So this class of failure is invisible in PR CI and only appears at release-tag time — exactly how #28593603470 happened.

Enough to fix it: move the `matchMedia` read into an exported `initPrefersReducedMotion()` called once from `main.ts` before bootstrap. Specs then default to `false` and can drive it deliberately. (It won't let a spec force it *on* without an injection token — if you want the feature itself unit-tested, that's the fuller version. Either way the checklist's `[ ] I have added tests` is currently unticked for a new util plus three wiring points.)

### 🟡 Minor

`focus-mode-main.component.ts:524` and `planner-calendar-gesture-handler.ts:37` still read `matchMedia` inline. Not a bug — but the gesture handler's read is a field initializer, so it never updates on an OS change. Candidates for a follow-up, not this PR.

## Verified, so you don't have to re-check

- **E2E is unaffected.** Playwright contexts default to `prefers-reduced-motion: no-preference` (`reducedMotion ?? "no-preference"` in `playwright-core`), nothing in `e2e/` overrides it, and E2E CI is ubuntu anyway.
- **Zoneless is fine.** `provideZonelessChangeDetection()` (`src/main.ts:258`) means the `.set()` from the `change` listener schedules CD through the `computed` in `app.component.ts` and the `effect` in `global-theme.service.ts`.
- **It earns its place.** #8827 is my own audit issue and proposes this approach; #8689 asks for it directly. No prior decline anywhere in closed issues/PRs.
- No new dependencies, no translation strings, no op-log/sync/persisted-model surface. `global-theme.service.ts` stays under the 1200-line cap. The wiki entry is the first one for this setting and lands in the right file per `docs/documentation-guide.md`.
- Base is ~38 commits behind `origin/master`; still `MERGEABLE`, but worth a rebase before merge.

## Verdict

Approach approved; not mergeable yet. The 🔴 lightbox wedge must be fixed (here or in a PR merged first) — shipping reduced-motion-by-default on top of it would break the app for the exact users it's meant to help. Then let's settle the #8689 "wins in both directions" question. The 🟡s are nice-to-have.

The blocked workflow runs need approving on my side before CI can report anything — I'll sort that out.

johannesjo (commented):
## Summary

Round 3 on `2e8c9c52f`. The round-2 blocker is fixed, and this time CI has actually run on the head: the previous head's runs are still sitting at `action_required` and always will be, but `2e8c9c52f` got a fresh set (CI, CodeQL, Android Native Tests, Electron Packaging Smoke Test, E2E Sync PR Gate, Wiki Update, PR Preview Build) and all seven passed. Worth noting for the record that the two *required* contexts, `SuperSync E2E Gate` and `WebDAV E2E Gate`, both cleared without running a suite, because `detect` correctly found no sync-affecting paths in your 11 files. The substantive green is the non-required `Tests` job, which is what actually exercised your new specs.

I re-verified the lightbox fix rather than taking the tests at their word, and it holds up. `_afterTransition` is applied at all three sites (`:67`, `:126`, `:140`). Both new specs are real coverage: disabling only the `setTimeout` fallback and leaving the `transitionend` listener intact turns exactly those two red and leaves the XSS and normal-url specs green. Asserting `getComputedStyle(wrapperEl).transitionProperty === 'none'` is a good positive control, since it proves the real `styles.scss` rule is in effect rather than a hand-copied approximation.

I also confirmed the thing PR CI structurally cannot: the macOS and Windows unit lanes are release-tag-only and they report `prefers-reduced-motion: reduce`, so green PR CI says nothing about them. Running the whole suite in Chrome with `--force-prefers-reduced-motion` gives **13414 passing, 0 failures**. So leaving out the `initPrefersReducedMotion()` seam is defensible today. One piece of counter-evidence worth knowing about: `focus-mode-main.component.spec.ts:40-45` records that this exact host-preference leak broke a release build once (run 28593603470), and the remedy there was faking `matchMedia` per spec. Capturing the value at module-eval time closes that remedy for the three new consumers. Not a blocker given the measurement, just the reason someone may add the seam back later.

On #8689: you kept the hard `||` and added the settings hint, which is exactly the fallback I offered in round 2, done properly (hint via the `isVerticalActionBar` pattern, `en.json` only, wiki paragraph, PR body reference). I'm not re-litigating it.

## Issues

### 🟠 I owe you a correction: my round-2 sweep was incomplete, and it missed a real one

I swept for `transitionend` / `animationend` **listeners**. I did not sweep for **CSS whose end state is load-bearing**, and there is one. With animations off, the done-toggle checkmark never renders, so completed tasks show an empty checkbox.

`done-toggle.component.scss` puts `.done-check` in a fully hidden resting state (`stroke-dasharray: 18; stroke-dashoffset: 18`) and reveals it *only* through `animation: draw-check ... forwards` on `.is-done`. `styles.scss:32-38` is `body.isDisableAnimations { *, *:after, *:before { animation: none !important } }`, unencapsulated and `!important`, so it wins. No animation means no `forwards` fill, so the offset stays at 18.

Measured in Karma against the real stylesheet, rendering `<done-toggle [isDone]="true">` and reading computed style after the animation window:

| | animationName | strokeDashoffset | visible stroke length |
|---|---|---|---|
| animations on | `draw-check` | `0px` | **17.70** |
| `isDisableAnimations` | `none` | `18px` | **-0.30** |

The polyline's total length is 17.70 against an 18px offset, so the entire checkmark sits inside the dash gap. Not dimmed, absent.

Scope, to be fair to it: the task row itself still signals done independently, via `--task-is-done-dim-opacity` and `text-decoration: line-through` on the title (`_task-base.scss:256-261`), neither of which is animation-dependent. So this is a missing checkmark rather than a task you cannot tell apart. It is also not new, `git tag --contains` puts it in **v18.0.0 and every release since**.

What makes it this PR's problem is the same argument I made about the lightbox in round 2. Today it only reaches people who deliberately ticked "Disable all animations", who have a mental model for why the UI changed and a one-click way back. This PR extends it to everyone whose OS reports reduced motion, silently, at first launch, and the `||` means that population has no way to turn it off. The primary affordance of a todo app rendering as unchecked on done tasks is not a good first impression for the accessibility cohort the PR exists for.

One line, in `&.is-done .done-check`:

```scss
// Resting state, so the checkmark is still drawn when the animation is
// suppressed (body.isDisableAnimations / prefers-reduced-motion).
stroke-dashoffset: 0;
```

Verified: with it, the disabled-animations column becomes `0px` / `17.70`, identical to animations-on, and the draw animation still runs normally when enabled. Same offer as last round, since this is a pre-existing bug in a file you did not touch: I am equally happy for it to land as its own PR merged first. I just do not want to ship the reduced-motion default while it is live.

For what it is worth I swept the whole class afterwards, and `done-toggle` is the only live instance. The other four `animation: ... forwards` rules are either cosmetic (the focus-mode rocket) or already carry their own `@media (prefers-reduced-motion: reduce)` block (`onboarding-preset-selection.component.scss:314-319`).

### 🟡 The reopen guard covers `_hideImg()` but not its two zoom-out callers

Correcting myself on framing here too: this one *is* introduced by this PR, and I should not have written it off as pre-existing. Before, both zoom-out branches used `newImageEl.addEventListener('transitionend', ...)`, bound to that lightbox's own image, so the callback died with the element and could never touch a later lightbox. The new `setTimeout` fallback is late-bound to `this` and calls `this._hideImg()`, which re-reads `this.enlargedImgWrapperEl` at fire time, and `_showImg()` reassigns that field on every open. The PR gets this right in `_hideImg()` (captures `const wrapperEl` at `:60`) and only misses it at `:126-130` and `:140-144`.

What makes it reachable is that `ani-remove` does not hide the image. Per `styles/components/enlarge-img.scss:46-48` it fades only the `:after` backdrop; the image itself just falls from `opacity: 1` back to its base `opacity: 0.5`, and nothing sets `pointer-events: none`. Probed 60ms after a close click: still in the DOM, `opacity: 0.5`, `pointer-events: auto`, and `elementFromPoint` at the viewport centre lands inside the wrapper. So a half-visible clickable image sits there for the full 500ms, and clicking it again is the natural reaction.

Probed end state after zoom in, zoom out, click the lingering image, wait for removal, reopen: the reopened wrapper gains `ani-remove` ~60ms after reopen and leaves the DOM ~570ms after. It closes itself.

Bounding the severity honestly: it is reduced-motion-only. With animations on I could not reproduce it, because a real `transitionend` latches `isDone` while the original wrapper is still current, so the 500ms timer no-ops. And the state it replaces was a permanent viewport-wide input block, so this is a clear net improvement either way. Not a blocker.

Six lines, on both branches:

```ts
const wrapperElAtClick = this.enlargedImgWrapperEl;
this._afterTransition(this.newImageEl as HTMLElement, () => {
  setTimeout(() => {
    // A reopen may have replaced the wrapper while the fallback timer
    // was pending; never hide a lightbox this click did not open.
    if (this.enlargedImgWrapperEl === wrapperElAtClick) {
      this._hideImg();
    }
  });
});
```

Verified: probe goes green, all four of your specs stay green. Related and sharing a root cause, so worth fixing in the same edit: `_afterTransition` never clears its `setTimeout` when `transitionend` wins, so every open leaves a dangling 500ms timer. It is inert thanks to the `isDone` latch, but `mobile-bottom-nav.component.ts`, the only other fallback of this kind in the codebase, stores its handle at `:84` and clears it both on the real `animationend` (`:94`) and on destroy (`:75`).

### 🔵 The spec's 4000ms timeout can never surface

`src/test.ts:88` pins `jasmine.DEFAULT_TIMEOUT_INTERVAL = 2000` in a global `beforeAll`, and neither `it` passes an override, so `waitFor`'s `timeoutMs = 4000` (`enlarge-img.directive.spec.ts:78-82`) can never be the reported failure. Its `Timed out waiting for ${label}` message, which is the useful one, is always pre-empted by Jasmine's generic 2s timeout. That is exactly what I saw when I disabled the fallback: two generic timeouts, no indication of which stage was stuck.

The zoom-out test also chains two 500ms fallbacks, so it needs roughly 1.1s of a 2000ms budget. It passed on CI, so this is a margin observation and not an observed flake. Dropping the `waitFor` default below 2000 so its message wins is the cheaper fix. Raising the interval per suite is the other option (`task-due.effects.spec.ts:76-83` and `operation-log-stress.benchmark.ts:37` both do it), but the repo pushes back on that where it would mask a slow spec (`operation-log-store.service.spec.ts:219-223`, `archive-store.service.spec.ts:116-119`), so I would lean toward the first.

### 🔵 Wiki wording

"the checkbox still shows unchecked and ticking it off has no visible effect" reads ambiguously, since the checkbox is already unchecked in the state being described. "ticking it has no visible effect" says what you mean.

Separately, "the app follows the system setting" is exact for the body class and for `@.disabled`, but `main.ts`'s `MATERIAL_ANIMATIONS.animationsDisabled` is a plain getter read by Material at component init, so a mid-session OS toggle will not retro-disable Material components that already exist. Same as the pre-existing config behaviour, so nothing to change in the code, just worth not overstating in the docs.

## Verdict

Fix the done-toggle checkmark, here or in a PR merged first, and I will approve. The zoom-out guard is worth folding in but I will not hold the PR for it.

On what this locks in long term, which I care about more than the diff: no new dependency, no new setting, no new UI, nothing touching persisted or synced state. The permanent surface is one 12-line util, three `||` reads, and a helper local to one directive. The `||` is the only real decision baked in, and I want to be straight about the asymmetry: backing it out later is a one-expression change, but making the in-app setting win in both directions is *not* cheap, because `isDisableAnimations` is a required boolean defaulting to `false`, so every existing install already stores an explicit `false` and "never touched" cannot be told apart from "explicitly off". That would need a new optional config field. I would rather not spend a second animation setting on it, and the manifesto's "one calm default over a new toggle" is the real reason I am accepting the `||`, not that it is cheap to revisit.

On splitting the lightbox fix into its own PR: not worth it now that it ships with a regression test and CI is green on the combined head. Landing them together keeps the fix attached to the change that widens its blast radius.

Good work on this one. The fallback helper is the right shape, the specs genuinely fail without it, and chasing the zoom path down rather than patching only the obvious site is what made the round-2 blocker actually go away.

MarkQuach12:
@johannesjo the done-toggle checkmark is fixed, so this should be ready for another look.

`&.is-done .done-check` now carries a resting `stroke-dashoffset: 0`, exactly as you suggested. Confirmed your measurement: with the line removed a spec asserting the computed value reads `18px`, with it `0px`, and the draw animation still runs when animations are on. Added `done-toggle.component.spec.ts` with that assertion plus a positive control so it cannot regress silently again.

Also folded in the two smaller items:

- The zoom-out reopen guard on both branches, capturing the wrapper at click time. One deviation from your snippet: keeping `let timeoutId` reassigned tripped `prefer-const`, so `_afterTransition` now clears the timer from inside the `transitionend` listener instead. Same behaviour, and the handle stays `const`.
- `waitFor`'s default is 1500ms so its message wins over the global 2000ms in `src/test.ts` rather than being pre-empted.

Wiki wording updated for both points, including the note that a mid-session OS change does not retro-disable Material components already on screen.

johannesjo:

Round 4. All four round-3 items are in, and I checked them against the source rather than taking the tests at their word:

- `done-toggle.component.scss:91` – `draw-check` runs `stroke-dashoffset` 18 → 0, so the resting `0` is exactly where the animation already ends. Animations outrank author declarations while they run, so the draw is unchanged when animations are on, and the checkmark is present at rest when they are off. The spec covers both directions.
- `enlarge-img.directive.ts` – I walked the reopen timelines. Open → zoom in → zoom out → reopen inside the 500ms window now compares against the wrapper captured at click time and correctly declines to hide the new one, and `_hideImg()`'s own fallback closes over `wrapperEl`, so a stale timer can only ever remove the wrapper it was scheduled for. Repeated zoom-out clicks stack independent latches, but each ends in an idempotent `remove()`. `clearTimeout` firing from inside the `transitionend` listener is correct.
- `waitFor` and both wiki wordings are fixed.

I owe you a second correction, though, and it is the same shape as the one I owed you last round.

### Should fix before merge

- **`src/app/ui/breathing-dot/breathing-dot.component.scss:16`** – my round-3 sweep was scoped to `animation: ... forwards` / `both`. `.breathing-dot-inner` is `animation: breathe 8s ease-in-out infinite` with no fill mode, so it was never in that grep, but its resting state is just as load-bearing.

  `@keyframes breathe` never exceeds `opacity: 0.4` (0%/100%, dipping to `0.1` at 50%), and the rule declares no resting `opacity`, so the base computed value is `1`. Under `body.isDisableAnimations` (`styles.scss:32-38`, `* { animation: none !important }`) the dot paints at full opacity: a solid `var(--c-primary)` disc with a `0 0 40px` primary glow, at 92% of the clock face (`focus-clock-face.component.scss:52`), sitting directly beneath the bold countdown digits in `.clock-face__content`. That is 2.5x the strongest opacity the design ever renders, and it eats the contrast on the digits.

  Reachable from `focus-mode-main.component.html:82` (`[variant]="isCountDown ? 'ring' : 'breathing'"`), so every Flowtime session. I confirmed nothing else can mask it: `grep breathing-dot-inner` returns exactly one rule in the whole repo, the consumer sets only `--breathing-dot-size`, and neither `src/styles/` nor the themes touch it.

  Same argument as the checkmark, and same disclaimer: this is not new and not yours. `git tag --contains` puts it in v13.1.1 and every release since. What this PR changes is who sees it, from people who deliberately ticked "Disable all animations" to everyone whose OS reports reduced motion. One line, alongside the animation:

  ```scss
  // resting state, so the dot still reads as a glow when the animation is
  // suppressed (body.isDisableAnimations / prefers-reduced-motion)
  opacity: 0.4;
  ```

  No effect when animations are on, since the keyframes drive opacity at 0%, 50% and 100%. I established this by reading the cascade rather than rendering it, so please sanity-check the value in the browser. And the same offer as the last two rounds stands: happy for it to land as its own PR merged first.

### Nits

- **`src/app/ui/enlarge-img/enlarge-img.directive.spec.ts:83`** – `1500` gets the named message ahead of Jasmine's 2000ms, which was the point, but it also becomes the binding deadline. The zoom-out spec's final wait needs roughly 1010ms of it (two chained 500ms `_afterTransition` fallbacks), so the margin is now about 490ms rather than the ~890ms the Jasmine cap used to allow. `1800` keeps the message-wins property and gives most of it back.
- Worth a rebase before merge. The branch is 354 commits behind `master`, so the green CI above is a merge against a base from five days ago. GitHub still reports it cleanly mergeable and I confirmed every file this PR touches is byte-identical between its base and current master, so this is about re-running CI on current code, not a conflict.

---

PR #9328, opened 2026-07-27
feat(short-syntax): interval recurrence phrases in the add task bar (@every 2 weeks, @every 2 fridays)

KamilDev:
Two changes to the add task bar. The first adds interval recurrence to the short syntax; the second — the larger half of the diff — makes the bar's text and its buttons stop contradicting each other, and changes what all five of its controls do to the input.

## Problem

**1. Interval recurrence was left out of the short syntax.** The grammar in the concept for #4961 included `@every 2 weeks`, but #9134 shipped without it: every repeat quick-setting hardcodes `repeatEvery: 1`, so interval phrases fell through to the plain-date path. On current `master`:

| input | what happens today |
| --- | --- |
| `thing @every 2 days` | one-off task "thing", due in 2 days, **no recurrence** — `every` silently swallowed |
| `Review @every 2 weeks` | one-off task, due in 2 weeks |
| `Water plants @every 2 fridays` | one-off task, due next Friday |

The user gets no signal that the recurrence wasn't understood, which is the part I'd call a bug rather than a missing feature.

**2. The add task bar's buttons and its text could contradict each other.** Type `@every 2 weeks`, then pick a different recurrence from the repeat menu: the input still says `@every 2 weeks`, the chip says something else, and which one wins depends on what you edit next (keep typing → the pick stands and the text lies; delete the phrase → your pick is cleared too; edit the phrase → the syntax wins). Same divergence for the schedule dialog, the deadline dialog and the estimate menu.

While fixing that I ran into a related case of the parser eating text it shouldn't — the one you invited as standalone fixes in #4961. The clear buttons removed one whitespace-delimited token, which truncates every multi-word one:

| input, clear the date | before | after |
| --- | --- | --- |
| `Call mom @next friday` | `Call mom friday` | `Call mom` |
| `Task @tomorrow evening` | `Task evening` | `Task` |
| `Water plants @every 2 fridays` | `Water plants 2 fridays` | `Water plants` |

### Intervals

`@every <1-999> <days\|weeks\|months\|years\|weekday-name>`, e.g. `@every 3 days`, `@every 2 weeks`, `@every 2 fridays`.

An interval has no preset to expand, so it maps to a `quickSetting: 'CUSTOM'` config carrying the cycle and interval directly — the one setting whose interval the repeat dialog can display and round-trip, so it stays editable after creation. **No new UI, settings, dependencies, or sync/schema surface**: the existing repeat chip and repeat-cfg model already cover it, and the occurrence utils already gate on `diff % repeatEvery` for all four cycles.

Details worth reviewing:

- **Bounds live in the grammar, not in the parse.** `[1-9]\d{0,2}` excludes a never-recurring `@every 0 days`, and its upper end stays one notch inside the dialog's `repeatEvery` max of 1000. Keeping the bound in the grammar is what keeps the derived removal regex in step: a bound checked in the parse instead would leave the clear-repeat button deleting `@every 0 days`, which the parser never consumed.
- **`@every 2 weekdays` deliberately stays a plain date.** Every other workday is not expressible as a weekly cycle (five weekday flags plus an interval means every other *week*, all five days), so accepting it would build a silently wrong schedule.
- **An interval of 1 collapses to the equivalent preset**, so `@every 1 week` and `@every week` are indistinguishable downstream — same chip, same `skipOverdue` default, same editable-as-a-preset config.
- **Weekly intervals restrict the weekday flags to the first occurrence's weekday.** `DEFAULT_TASK_REPEAT_CFG` enables Mon–Fri and the WEEKLY branch treats the flags as an independent filter, so returning cycle + interval alone would mean "every other week, Monday through Friday".
- **First-occurrence anchoring now keys off the cycle** rather than the quick setting, which extends the existing "don't let chrono's `forwardDate` move the recurring day" rule to intervals: `@every 2 weeks 6am` typed Wednesday 10:00 recurs on Wednesdays, not Thursdays. Schedules where every day is an occurrence (daily, and Mon–Fri) take chrono's next occurrence as-is.
- **`skipOverdue` needed no change.** `getDefaultSkipOverdue` already keeps every-N-days OFF, so a missed occurrence stays visible.

### Input/button consistency

This half is the bigger one. It changes what every control on the add task bar does to the input — the schedule dialog, the deadline dialog, the estimate menu, the repeat menu, and all five clear (`x`) buttons — so it is worth reviewing as its own change even though it shares a branch with the intervals above.

- Picking a date, deadline, estimate or recurrence now strips the syntax it overrides, the way the clear buttons already do. One source of truth at any moment.
- Removal deletes the exact character ranges the parser consumed (`parsedRanges`, already produced on every parse) instead of guessing a token's extent with a regex. Clearing one *named* tag keeps its regex on purpose — the ranges say which characters were tags, not which tag they named. Where no parse has landed for the exact current text, everything falls back to the old regexes.
- Removing a recurrence takes the **whole due token** it came from, not just the phrase. The token can have absorbed an adjacent time (`@daily 6am`), which the phrase grammar alone cannot see; deleting only the phrase orphaned `6am` into the title of both the task and its repeat config. A recurrence the text did not produce is left alone, so picking a recurrence does not eat an unrelated `@tomorrow` you typed.
- A pick discards the in-flight parse it invalidates — but only when the strip changed the text, because that change is what queues the parse that recomputes everything the discarded one would have published. Without that condition it would drop a parse whose tags/project/title nothing else recomputes.
- One case is more than mechanical: a recurrence phrase *is* due syntax, so picking a date strips it. Killing the recurrence there would be wrong (you changed the date, not the schedule), so the date pick takes ownership of the existing repeat first — `@every 2 fridays` + picking Monday keeps the every-2-weeks recurrence, re-anchored to Monday, and the chip re-labels itself accordingly. The mirror also holds: picking a recurrence keeps the date the stripped phrase had anchored.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

Full `npm test` green — both stages, `Europe/Berlin` and `America/Los_Angeles`, 13573 passing and 16 skipped in each. New coverage: interval grammar and anchoring in `short-syntax.spec.ts`, the cfg resolver in `get-interval-repeat-updates.spec.ts`, the created config (incl. the weekday restriction and `skipOverdue`) in `add-task-bar.component.spec.ts`, and range-based removal, ownership handoff and the in-flight-parse case in `add-task-bar-parser.service.spec.ts`. Each spec that pins a guard was checked by breaking the guard and watching it fail, rather than assumed to pin it.


johannesjo (changes requested):
Thanks — and thanks for coming back to this with the interval question properly solved rather than re-litigated. The move from "expand a preset" to `quickSetting: 'CUSTOM'` is the right answer to why intervals got dropped from #9134 in the first place, and it retires all three blockers I raised there rather than working around them. I verified each by running the code:

| #9134 blocker | status here |
| --- | --- |
| `@every 3 days` → `DAILY` preset → `skipOverdue: true` (a missed occurrence vanishes) | ✅ `getDefaultSkipOverdue({quickSetting:'CUSTOM', repeatCycle:'DAILY', repeatEvery:3})` → `false` |
| repeat dialog resets `repeatEvery` → 1 for a non-CUSTOM setting | ✅ CUSTOM is the one branch the dialog shows (`task-repeat-cfg-form.const.ts:51` `hideExpression`, `resetOnHide: false`) and `_processQuickSettingForDate` passes CUSTOM through untouched — round-trips |
| `@every 2 weekdays` → button label that lies about the schedule | ✅ excluded from the interval grammar (and sabotage-pinned — see below) |

I also drove the created configs through the real occurrence engine rather than trusting `repeatEvery` is honoured:

```
every 2 WEEKLY  from 2024-01-17 (Wed) → 01-17 | 01-31 | 02-14 | 02-28 | 03-13   days=[wednesday]
every 3 WEEKLY  from 2024-01-19 (Fri) → 01-19 | 02-09 | 03-01 | 03-22 | 04-12   days=[friday]
every 3 DAILY   from 2024-01-17       → 01-17 | 01-20 | 01-23 | 01-26 | 01-29
```

and the grammar itself over 34 inputs — `@every 2 fri|friday|fridays`, `@every 2 wed|thu|tues|sun`, `@every 2  days` (double space), `@every 999 days`, `@every 2 weeks.` → `"Task."`, `@every 2 weeks, and more` → `"Task, and more"`. `@every 0 days`, `@every 1000 days`, `@every 01 days`, `@every two weeks`, `@every 2 quarters`, `@every 2 weekdays`, `@every 2 workdays` all fall through as documented, and the derived removal regex leaves every one of them alone. Twelve sabotage edits against the new logic: eight caught by your specs, four survived (below).

One blocker (red CI), one thing I'd want fixed before merge, then smaller stuff.

---

## Blocker — CI is red, and the PR says it isn't

The `Tests` job fails on the `TZ=America/Los_Angeles` stage of `npm test` (which re-runs the whole suite). Reproduced locally:

```
AddTaskBarActionsComponent > Schedule Dialog > should pass a picked reminder option along
  Expected spy applyUserDatePick to have been called with: [ '2024-01-15', '14:30', 'AtStart' ]
  but actual calls were:                                   [ '2024-01-14', '14:30', 'AtStart' ]
```

The new spec at `add-task-bar-actions.component.spec.ts:732` builds the dialog result with `new Date('2024-01-15')` (`:735`) — an ISO date string, so **UTC** midnight — and then hardcodes the **locally** formatted `'2024-01-15'` as the expectation (`:744`). The component runs it through `getDbDateStr`, which is local. In Berlin that's Jan 15 01:00 and passes; in LA it's Jan 14 16:00.

The pre-existing spec directly above (`:717`) starts from the same UTC `Date` but derives its expectation with `getDbDateStr(resultDate)` — the same helper the component uses — so it holds in any zone. That's the pattern to copy.

The one-line fix, verified green under LA (324/324):

```ts
const resultDate = new Date('2024-01-15');
// …
expect(mockParserService.applyUserDatePick).toHaveBeenCalledWith(
  getDbDateStr(resultDate),
  '14:30',
  TaskReminderOptionId.AtStart,
);
```

The PR body says "Full unit suite green (13566)" and ticks "Existing tests still pass". 13566 is the number of specs the *failing* LA job executed — the green Berlin run executes 13580, and LA's successes are 13565 — so no green run on this head produces 13566 in any timezone. Worth mentioning because you're the author of #9131, which fixed exactly this class of spec; `npm test` runs Berlin and then LA, so a Berlin-only pass isn't the suite.

I also swept the touched specs under Berlin / LA / Tokyo / Sydney: 1154 specs, green everywhere except the one above under LA. Only one added ISO date-only `Date` exists in the whole diff — this one; every other added date literal uses the local `new Date(y, m, d)` constructor or goes through `todayStr`.

## 1. Picking a recurrence from the menu orphans the phrase's time into the task title

This is the one I'd most like fixed before merge, because it lands in created data.

`applyUserRepeatPick` is the only pick that doesn't strip via `_removeParsedRanges` — it stays on `SHORT_SYNTAX_REPEAT_REMOVAL_REG_EX`, which matches the phrase but not the adjacent chrono match the recurrence absorbed. The `due` range covers phrase *plus* adjacent match (`consumedTotal`, `short-syntax.ts:865`), so what the regex leaves behind ends up in `cleanText` — which `addTask()` uses verbatim as the title of both the task and its repeat config.

Driving the real `AddTaskBarParserService` + real `AddTaskBarStateService` (type the text, pick "Every month on the day …" from the repeat menu, let the queued parse land):

| typed | input after pick | resulting title |
| --- | --- | --- |
| `journal @daily 6am` | `journal 6am` | **`journal 6am`** |
| `pay rent @every 15th 9am` | `pay rent 9am` | **`pay rent 9am`** |
| `standup @every 2 fridays 3pm` | `standup 3pm` | **`standup 3pm`** |
| `review @every 2 weeks tomorrow` | `review tomorrow` | **`review tomorrow`** |
| `shopping @every friday` (control) | `shopping` | `shopping` ✅ |

The time itself is still applied — `state.time` survives — so this is pure title pollution, which is exactly what `_removeParsedRanges`' own docstring calls "worse than not clearing at all — the leftover words stay in the task title".

It is new. I ran master's code path as a control (`updateRepeatSetting(setting)` with no `cleanedInputTxt`, input untouched) and the title stays `journal`. To be fair about the trade rather than just calling it a regression: on master the *pick* stands while the text keeps saying `@daily` — I reproduced that divergence too, so the problem you set out to fix is real and this PR does fix it. The cost is just landing in the wrong place. `@every friday 3pm` and `@daily 6am` are both headline examples in the wiki, so this isn't an exotic input.

The fix fits your own pattern: strip with `_removeParsedRanges(text, 'due')` and record `{ dueDate, dueTime, repeat, isRepeatFromSyntax: false }` — the exact mirror of `applyUserDatePick` taking ownership of the repeat.

## 2. Three coverage gaps the sabotage turned up

Four of the twelve edits survived. One of them is fine (finding 3 — the code is inert, so nothing *should* fail). These three are real gaps: the code is load-bearing and nothing pins it.

**The stale-ranges fallback spec can't fail.** `add-task-bar-parser.service.spec.ts:868` — "should fall back to token removal when the ranges are for other text" — passes with the guard removed. I sabotaged `add-task-bar-parser.service.ts:499` to `if (!this._lastParsedRanges) {`, dropping the `forText` check entirely, and the whole suite stayed at 1154 SUCCESS. The reason is the `.replace(/\s+/g, ' ').trim()` at the end of `removeShortSyntaxFromInput`: applying the stale range `[9,21)` to `'Call dad @today'` yields `'Call dad '`, which normalises to the same `'Call dad'` the fallback produces.

The guard is genuinely load-bearing — it just needs an input where stale offsets land somewhere destructive:

```ts
it('should not slice an unrelated text with stale ranges', async () => {
  await parse('a @next friday');           // ranges: due [2,14)
  expect(
    service.removeShortSyntaxFromInput('Buy milk and eggs for dinner', 'date'),
  ).toBe('Buy milk and eggs for dinner');
});
```

Without the guard that returns `'Buggs for dinner'`; with it, unchanged. (Confirmed both ways.)

**Multi-range removal has no coverage at all.** `_removeParsedRanges` sorts back-to-front (`:508`) so an earlier deletion can't shift a later range. Flipping that sort to `a.start - b.start` also leaves the suite at 1154 SUCCESS — but it is not dead defensiveness, because an estimate in the middle of a due phrase really does split it into two ranges of the same type:

| input | `date` clear, pristine | with the sort reversed |
| --- | --- | --- |
| `Task @tomorrow 1h evening` | `"Task 1h"` | `"Task 1h evening"` |
| `Task @next 1h friday` | `"Task 1h"` | `"Task 1h frida"` |
| `Task !next 1h friday` (deadline) | `"Task 1h"` | `"Task 1h frida"` |

Note the truncated `frida` — that's the exact "leftover words stay in the task title" failure the ranges were introduced to prevent. One spec with a two-range token would pin it.

**The `MONDAY_TO_FRIDAY` case is load-bearing but untested.** `anchorCycleOf`'s PRESET `default: return null` is what stops `@every weekday 6am` from anchoring like a weekly recurrence — but making it return `'WEEKLY'` also leaves the suite at 1154 SUCCESS. Unlike the DAILY ternary below, this one genuinely changes behaviour (`@every weekday 6am` typed Wed 10:00 moves from Thu to Wed+7), so it deserves a spec.

## 3. The DAILY exemption in `anchorCycleOf` is inert

`short-syntax.ts:825`, `return repeat.repeatCycle === 'DAILY' ? null : repeat.repeatCycle`. Changing it to plain `return repeat.repeatCycle` leaves the suite fully green, including "should let a daily interval take chrono's next occurrence of a passed time" — because nothing downstream ever compares `anchorCycle` against `'DAILY'`. The five reads (`:908`, `:911`, `:918`, `:930`, `:932`) test only `'WEEKLY'`, `'MONTHLY'` and `'YEARLY'`, so `null` and `'DAILY'` take the same path everywhere.

The comment at `:900` ("DAILY is exempt (anchorCycle null): every day is an occurrence…") therefore describes a distinction the code doesn't actually make. Not a bug — the behaviour is correct either way — but I'd either drop the ternary (the function is really `cycleOf`) or reword the comment so the next reader doesn't go looking for the branch that consumes it. Only the interval ternary is inert — the PRESET `default: null` really does change behaviour, it's just untested (finding 2).

## 4. Stale example in the removal-regex comment

`short-syntax.ts:198` still offers `"@every 2 fridays"` as an example of a phrase "the parser did not treat as a recurrence" and therefore leaves untouched. It's now the flagship supported form, and your own spec asserts `removeShortSyntaxFromInput('Water plants @every 2 fridays', 'repeat')` → `'Water plants'`. I confirmed against the real regex. `@every 2 weekdays` or `@every 1000 days` would be honest examples.

(Minor, same paragraph: `:174` justifies the 1–999 bound as excluding "anything past the dialog's `repeatEvery` max", but the dialog's max is 1000 and it's a real `Validators.max`, so the grammar is one notch stricter rather than equal. Not worth changing the bound — just the sentence, if you're editing the block anyway.)

While re-reading that comment I checked its central promise — "the button can only ever delete text the parser would have consumed" — and it doesn't hold in either direction. The parser matches `^phrase` *inside the first `@…` due-match* (which ends at the next `+ # @ !`), while the removal regex is global over the raw input with its lookahead against the whole string. So the two disagree at both ends:

```
                                 parser sees          clear-repeat button
Task @every friday#tag           WEEKLY_CURRENT_WEEKDAY   leaves it   ← can't clear it
Task @tomorrow @daily            no recurrence            deletes it  ← eats untouched text
```

Both of those are on `master` already — this is the grammar I asked you to derive in #9134 and my "can only ever delete what the parser consumed" framing was too strong, so that one's on me. What's new here is that intervals inherit it: `Task @every 2 weeks#tag` and `Task @every 2 weeks+proj` now parse as a recurrence the button won't clear, and `Task @tomorrow @every 2 weeks` now gets its second token deleted (on master that string is untouched, because the grammar didn't cover it). I verified all of this by running the real parser and the real exported regex on both `master` and this branch.

The under-delete direction has a user-visible consequence worth knowing about, because it contradicts a claim the PR adds to the wiki ("The buttons and the text can therefore never disagree about what will be created"). With a `+project` or `#tag` glued to the phrase, the pick is a silent no-op that reverts on the next keystroke:

```
Water plants @every friday+Garden   parsed=WEEKLY_CURRENT_WEEKDAY
  → pick "Every day"        → DAILY        (input unchanged, phrase still there)
  → one more keystroke      → WEEKLY_CURRENT_WEEKDAY   ← silently back
Review @every 2 weeks#home  → pick "Every day" → DAILY → next keystroke → INTERVAL WEEKLY 2
```

Not asking you to fix the anchoring in this PR — it's narrow and it predates you. But the comment shouldn't restate the guarantee as airtight ("same grammar as the parser, anchored differently" is the honest version), and the wiki's "can never disagree" needs the same softening. Note that fixing finding 1 the way I suggested — routing the repeat pick through `_removeParsedRanges('due')` — also closes this, since the ranges don't care where the regex thinks the token ends.

## 5. Merge-order heads-up

#9014 is still open and every one of its eight files is also in this PR. I ran the actual merge (`git merge-tree` against `6f38d45c`) rather than eyeballing the lists — three real conflicts: `docs/wiki/3.04-Short-Syntax.md`, `add-task-bar-parser.service.spec.ts`, `add-task-bar-state.service.ts`. Worth knowing that **none of them come from the `repeatQuickSetting` → `repeat` rename** — `add-task-bar.const.ts` merges clean, because #9014's hunks there are in a different region. The state-service conflict is just `updateDate` gaining `cleanedInputTxt` next to #9014's new `updateSectionId`. #6511 also overlaps on `short-syntax.ts` and `add-task-bar-parser.service.ts`. Nothing to do now; both are resolvable.

## Small stuff

- `removeShortSyntaxFromInput`'s `'urls'` case (`:568`) has no caller anywhere — all seven call sites pass string literals, none of them `'urls'`. The dead branch is pre-existing, so this isn't your doing; but the PR does add range-based removal to it, which is work spent on a path nobody reaches. Deleting the case would be the leaner edit.
- `applyUser*Pick` discards the in-flight parse via `_parseRunId++`, but `clearDateWithSyntax` / `clearDeadlineWithSyntax` / `clearEstimateWithSyntax` do the same strip without it. I tried to build the clobber and couldn't — the "only publish on change" comparisons absorb it — so this is an observation about the asymmetry, not a defect.
- A menu-picked recurrence inherits the stripped phrase's anchor date: `buy milk @every friday` + pick "Every day" keeps `date` = next Friday, so the daily config starts next Friday. Visible in the date chip and arguably what you'd want, but `applyUserDatePick` documents *and* tests the mirror case while this direction has neither. A sentence either way would settle it.

## On splitting

You offered to split intervals from the consistency work. I'd rather you didn't at this point — I've reviewed both halves, and a split now costs a rebase round without telling either of us anything new.

But it's worth noticing where the findings landed: the interval half came through clean, and everything I'd want changed (finding 1, and both halves of finding 2's range coverage) is in the consistency work. That's the half the body describes almost as an aside, and it's half the diff and changes the behaviour of five buttons on the most-used input in the app. Please give it its due in the description — and if you'd rather de-risk the merge, splitting *now* would be a reasonable call too; I just wouldn't ask you to.

## Checked, no finding

- `getIntervalRepeatUpdates` ordering: `repeatEvery` after the `buildWeeklyForDay` spread, so the preset's hardcoded `1` doesn't win. Sabotaging the spread away fails 3 specs.
- Preset path is byte-equivalent to before: `getQuickSettingUpdates` never returns a `quickSetting` key, so moving it ahead of the spread changes nothing.
- No persistence hazard from the state rename — only `inputTxt`/`noteTxt` reach `sessionStorage`.
- AOT template typecheck passes (`ng build` clean); CI's `Build Frontend` was skipped behind the red `Tests` job, so it hadn't run.
- `checkFile` clean on the changed `.ts` files.
- Unrelated and pre-existing, but you're in this code: `@every weekday 6am` typed on a Friday after 06:00 gives a Mon–Fri recurrence whose first occurrence is **Saturday**. Byte-identical on master, so not yours — just cheap for you to fix if you feel like it.
- The range path is a quiet improvement for people who never type a recurrence: because `_lastParsedRanges` is recorded even when nothing parsed, clearing the date chip on `reply to @john` now leaves the text alone instead of eating `@john` the way the old `\s*@\S+` did.
- Dialog round-trip proven with the real component + formly, not just by reading `hideExpression`: `repeatEvery` survives a no-op save, a `CUSTOM→DAILY→CUSTOM` flip, and a start-date change (`saveChanges` stays `{}`).
- `getIntervalRepeatUpdates` re-deriving `startDate` from `referenceDate` is a genuine no-op — `getDbDateStr(dateStrToUtcDate(s)) === s` over 800 consecutive days under LA.

KamilDev:
Thanks — the sabotage pass found three guards I had convinced myself were pinned and weren't, which is exactly the check I should have run myself. All of it is addressed; details per finding.

## Blocker — the red LA stage

Fixed, and you're right that the "green (13566)" line was wrong: I ran `test:once` (Berlin) and read its count as the suite. `npm test` runs Berlin *and* LA, and the LA stage is where it fails. The spec now derives its expectation the way the one above it does:

```ts
const resultDate = new Date('2024-01-15');
// …
expect(mockParserService.applyUserDatePick).toHaveBeenCalledWith(
  getDbDateStr(resultDate),
  '14:30',
  TaskReminderOptionId.AtStart,
);
```

Full `npm test` is green now — both stages. Given #9131 was mine, having shipped this one is the finding I'd least like to repeat; running the whole `npm test` rather than the Berlin stage is the fix on my side.

## 1. The orphaned time — fixed, and it was on the clear button too

Fixed as you suggested: removing a recurrence now deletes the `due` **ranges**, so it takes the whole token the parser consumed — phrase plus any time it absorbed — instead of the phrase alone.

Two things I found while doing it:

**The clear (`x`) button has the identical bug**, not just the pick. `Call @every friday 3pm` → clear the recurrence → text `Call 3pm` → title `Call 3pm`. It's the same code path, so it's fixed by the same change; the spec `should strip the absorbed time when clearing the recurrence too` pins it. (This also means the existing `should leave a trailing time token like the date case does` spec was pinning a bug — it now only describes the no-parse fallback, where the derived grammar genuinely cannot see the time.)

**Unconditional due-range removal over-reaches**, so the strip is gated on the recurrence having come from the text. `Call mom @tomorrow` + pick "Every week" from the repeat menu would otherwise delete `@tomorrow` — a plain date the user typed is not something the repeat control overrides. The gate is: with a parse current for this exact text and no recurrence in it, leave the text alone; otherwise remove the ranges, falling back to the derived grammar when no parse has landed. Pinned by `should leave a plain date alone when a recurrence is picked` (fails, returning `Call mom`, without the gate).

I did not add the `{ dueDate, dueTime }` to `_recordUserPick`. Tracing it, the values are always identical to what's already in `_previousParseResult` — the repeat pick doesn't change the date, unlike `applyUserDatePick` where recording it is load-bearing — so it would compare equal and never fire. The date surviving the strip is real and worth pinning, though, so it has a spec of its own: `should keep the date the stripped phrase anchored`. Say the word if you'd rather have it recorded explicitly anyway.

This also closes the under/over-delete pair from finding 4, as you predicted: `Water plants @every friday+Garden` now clears (ranges don't care where the regex thinks the token ends), and `Task @tomorrow @daily` is left alone (no recurrence parsed → nothing to strip). Both disagreements now only exist in the no-parse-yet fallback.

## 2. The three coverage gaps — all three added, all three verified by breaking the guard

| gap | spec | what it does with the guard broken |
| --- | --- | --- |
| stale-ranges `forText` check | `should not slice an unrelated text with stale ranges` | `Buggs for dinner` — your input, verbatim |
| back-to-front multi-range sort | `should remove two ranges of the same type without shifting` | `Task 1h evening` |
| `MONDAY_TO_FRIDAY` anchoring | `should not anchor "@every weekday 6am" to today's weekday` | first occurrence moves Thu 18th → Wed 24th |

One correction on the third: `default: return 'WEEKLY'` isn't fully green today — it also fails the existing `should leave "@daily 6am" on chrono's next-day roll`. So the gap was `MONDAY_TO_FRIDAY` specifically, which is what the new spec covers.

## 3. The DAILY ternary

Kept, comment reworded. Dropping it would make `INTERVAL`+`DAILY` return `'DAILY'` while `PRESET`+`DAILY` still returns `null` through the switch's `default` — the same schedule described two ways, which is worse than an inert branch. Instead the function's doc now says outright that callers test for the anchored cycles by name and nothing checks for DAILY, and the comment at the call site no longer implies a DAILY-specific branch exists.

## 4. The comment block

Both fixed, plus the guarantee itself:

- the stale `"@every 2 fridays"` example is now `"@every 2 weekdays"`;
- the bound sentence now says the upper end is one notch inside the dialog's max of 1000, rather than implying they're equal;
- "the button can only ever delete text the parser would have consumed" is gone. The comment now states the anchoring difference and both directions it fails in, names the range-based removal as what avoids them, and marks the regex as the not-yet-parsed fallback. The wiki's "can never disagree" is softened to "the text stops advertising a value the task will not get".

## 5 / small stuff

- **`'urls'`** — deleted, along with its range handling. No caller, as you found.
- **Merge order** — noted, nothing changed here. Happy to rebase onto whichever of #9014 / #6511 lands first.
- **PR description** — rewritten. The consistency half now leads the summary, says it changes all five controls and the clear buttons, and is flagged as reviewable on its own.
- **`clearDateWithSyntax` et al. not bumping `_parseRunId`** — agreed it's asymmetric and I couldn't build a clobber either. Leaving it: adding the bump on a hunch would be defending against a race I can't produce.
- **`@every weekday 6am` on a Friday → Saturday first occurrence** — real, and pre-existing as you say. I've left it out to keep this diff to the reviewed scope; happy to send it as a one-liner separately.

johannesjo:
Thank you very much! <3

PR #9333, opened 2026-07-27
fix(sync): preserve forced-download conflict recovery

johannesjo:
## Summary

- re-read rejected operations after a forced-from-sequence-zero download before resolving superseded local operations
- stop recovery when that download already retired every rejected operation, and resolve only the operations that remain pending
- propagate local-win operation counts from nested downloads so replacement operations are uploaded in the same sync cycle
- add regression coverage for complete and partial retirement, operation-count propagation, cancellation, and retry behavior

## Root cause

The rejection handler retained its pre-download pending-operation list across forced recovery. When that download applied remote conflicts and retired some or all of those local operations, the handler could continue resolving stale entries a second time. It also omitted local-win operations created by nested downloads from the count returned to the outer sync loop.

## Impact

Conflict recovery no longer resolves already-retired operations twice, and any replacement operations created during nested downloads are uploaded without waiting for another sync cycle.

## Validation

- `rejected-ops-handler.service.spec.ts`: 41 tests passed
- `operation-log-sync.service.spec.ts`: 152 tests passed
- exact retries-disabled SuperSync regression E2E: passed
- scheduled E2E run 30263635451: all six SuperSync shards, WebDAV, and workflow gates passed
- `npm run checkFile` passed for all five modified TypeScript files
- six-perspective multi-review verdict: APPROVE (high confidence)

### DIFF
--- a/src/app/op-log/core/types/sync-results.types.ts
+++ b/src/app/op-log/core/types/sync-results.types.ts
@@ -242,6 +242,8 @@ export type DownloadResultForRejection =
   | {
       kind: 'completed';
       newOpsCount: number;
+      /** Local-win operations created while applying the nested download. */
+      localWinOpsCreated?: number;
       allOpClocks?: VectorClock[];
       snapshotVectorClock?: VectorClock;
       /** Server cursor after the downloaded operations were durably applied. */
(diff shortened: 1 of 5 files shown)


johannesjo:
Superseded by #9326, whose last two commits (`fix(sync): avoid duplicate rejected-op recovery` + `test(sync): prove single-cycle rejected-op recovery`) absorbed this exact fix with stricter pending-op guards (`source !== 'local'`, `reducerRejectedAt`). Verified file-by-file against master: rebasing would leave an empty-to-negative diff, so closing instead of resolving conflicts.

PR #9334, opened 2026-07-27
fix(sync): confirm first-sync overwrites

### LINKED ISSUE #9166: First-sync conflict skips destructive overwrite confirmation with Never baseline
## Problem

A genuinely new WebDAV client can show a first-sync conflict whose table says Local -> Remote: Never, yet clicking Keep local or Keep remote closes the conflict immediately without the secondary destructive-overwrite warning.

This was reproduced in the real browser flow in e2e/tests/sync/webdav-first-sync-conflict.spec.ts. The conflict showed Remote changes: 2, Local changes: 6, and Local -> Remote: Never. Clicking Keep local immediately began the replacement upload. A strict E2E assertion for the warning failed.

## Why this matters

The secondary warning is the last safety gate before replacing one side on a client that has no completed sync history. Small first-sync datasets stay below the 20-change threshold, so destructive replacement currently proceeds after only the parent choice even though the UI says there is no previous sync baseline.

## Likely cause

OperationLogSyncService passes VectorClockService.getSnapshotVectorClock() as lastSyncedVectorClock. A fresh app can already have a local state cache, so this returns a non-null clock even though the client has never completed a provider sync. DialogSyncConflictComponent treats every non-null clock as a usable last-synced baseline and applies the 20-change threshold. This conflicts with the LocalDataConflictError and dialog comments that null represents a genuinely fresh client and should always warn.

Relevant code:
- src/app/op-log/sync/operation-log-sync.service.ts around the snapshot-conflict LocalDataConflictError construction
- src/app/op-log/sync/vector-clock.service.ts getSnapshotVectorClock
- src/app/imex/sync/dialog-sync-conflict/dialog-sync-conflict.component.ts shouldConfirmOverwrite

## Acceptance criteria

- First-sync conflict resolution reliably distinguishes a completed provider-sync baseline from a local state/compaction snapshot.
- Keep local and Keep remote both require the secondary warning when there is no completed sync history.
- An E2E test asserts the exact warning and fails if the parent conflict closes directly.
- Established-client conflicts retain the existing measured-difference threshold behavior.

sviat-barbutsa:
## Problem

A fresh WebDAV client can already have a local state-cache snapshot before it has completed any provider sync. We were treating that snapshot clock as a last-synced baseline, so a small first-sync conflict could skip the secondary overwrite warning even while the conflict table showed `Never`.

Fixes #9166.

## Solution

Use the sync cycle's captured never-synced state when processing the conflict:

- fresh clients pass no last-synced baseline, so Keep local and Keep remote require the count-free safety warning
- established clients keep using the snapshot vector clock and the existing measured-change threshold
- a focused service test covers a never-synced client that already has a local snapshot
- the real WebDAV E2E scenarios now require the exact warning for both overwrite directions

## Verification

- `checkFile` passed for all three changed TypeScript files
- focused `OperationLogSyncService` suite: 153/153 passed
- first-sync WebDAV E2E scenarios: 2/2 passed

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] Documentation/wiki changes are not needed for this internal sync behavior fix.
- [x] Relevant existing tests still pass

### DIFF
--- a/e2e/tests/sync/webdav-first-sync-conflict.spec.ts
+++ b/e2e/tests/sync/webdav-first-sync-conflict.spec.ts
@@ -107,18 +107,20 @@ test.describe('@webdav WebDAV First Sync Conflict', () => {
     await useLocalBtn.click();
     console.log('[Test] Clicked Use Local on Client B');
 
-    // Handle potential confirmation dialog (for overwrites with many changes)
+    // A first-sync overwrite must always require the count-free safety warning.
     const confirmDialog = pageB.locator('dialog-confirm');
-    try {
-      await confirmDialog.waitFor({ state: 'visible', timeout: 3000 });
-      // Click the confirm/OK button
-      await confirmDialog
-        .locator('button[color="warn"], button:has-text("OK")')
-        .first()
-        .click();
-    } catch {
-      // Confirmation might not appear - that's fine
-    }
+    await expect(confirmDialog).toBeVisible();
+    await expect(confirmDialog.locator('.content')).toHaveText(
+      'WARNING: This device cannot determine how many unsynced changes the remote data has (it has no record of a previous sync). Overwriting the remote data with the local version may discard changes. Are you sure?',
+    );
+    const snapshotUploadResponse = pageB.waitForResponse(
+      (response) =>
+        response.request().method() === 'PUT' &&
+        new URL(response.url()).pathname.endsWith('/sync-data.json'),
+      { timeout: 60000 },
+    );
+    await confirmDialog.locator('[e2e="confirmBtn"]').click();
+    expect((await snapshotUploadResponse).ok()).toBe(true);
 
     // Wait for sync to complete
     await waitForSyncComplete(pageB, syncPageB, 30000);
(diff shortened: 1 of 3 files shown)


johannesjo (changes requested):
Requesting changes for two correctness gaps in the sync-history decision.

1. Important — [operation-log-sync.service.ts:777](https://github.com/super-productivity/super-productivity/blob/8d9e908c131dc2c8a8d6a640b4aec957f4bd4bc5/src/app/op-log/sync/operation-log-sync.service.ts#L777): isNeverSyncedAtSyncStart only protects conflicts found by the initial pending-op read. The existing late-durable-op path in _hydrateSnapshotExclusive() can see an op after that read and still constructs LocalDataConflictError with getSnapshotVectorClock() unconditionally. On a never-synced client with a cache, small deltas remain numeric and can skip the required secondary warning. Please capture the baseline decision once for the whole snapshot attempt, pass it into _hydrateSnapshotExclusive(), and extend the existing late-op test to assert a null baseline.

2. Important — [operation-log-sync.service.spec.ts:3507](https://github.com/super-productivity/super-productivity/blob/8d9e908c131dc2c8a8d6a640b4aec957f4bd4bc5/src/app/op-log/sync/operation-log-sync.service.spec.ts#L3507): this test injects isNeverSynced: true, so it does not validate production classification. Production derives that value from !hasSyncedOps(), which only checks retained synced rows. A completed snapshot-only sync can commit a cache and provider cursor with zero included ops, and compaction can remove all old synced rows; this change then misclassifies an established client and forces the unknown-count warning, contrary to #9166. Please use the existing persisted pre-download provider cursor, getLastServerSeq() > 0 or equivalent, for this baseline decision and cover a cursor-positive/no-synced-rows case. No new persisted flag seems necessary.

Both findings can be fixed with existing primitives: capture one provider-baseline decision before download, thread it through both conflict constructors, and extend the focused tests. No new flag, service, dependency, or abstraction appears necessary. The E2E assertions are a useful strengthening; I found no additional security, performance, or dependency concerns.

johannesjo:
@sviat-barbutsa thanks again for the analysis and the E2E hardening! Since a week passed without a response I completed the two review items on top of your commit so this can ship:

- The baseline decision is now captured once per snapshot attempt and passed into `_hydrateSnapshotExclusive()`, so the late-durable-op conflict path can no longer fall back to `getSnapshotVectorClock()` on a never-synced client.
- Never-synced is no longer derived from retained synced rows alone: a snapshot-only first sync commits zero synced rows and compaction can prune them later, so the persisted per-provider cursor (`getLastServerSeq() > 0`, which only advances after a completed sync) now also counts as proof of a completed sync. The spec covers the cursor-positive/no-synced-rows case and exercises the production classification instead of injecting `isNeverSynced`.

Focused suite: 155/155. Each new assertion was verified to fail against the unfixed code. Will merge once CI is green.

PR #9338, opened 2026-07-27
feat(i18n): add Hebrew localization and RTL support

y-shlomo:
- Added Hebrew translation file (`src/assets/i18n/he.json`).
- Added `LanguageCode.he` and updated `RTL_LANGUAGES` in `src/app/core/locale.constants.ts`.


src/app/core/locale.constants.ts
```diff
@@ -13,6 +13,7 @@ export enum LanguageCode {
   fa = 'fa',
   fi = 'fi',
   fr = 'fr',
+  he = 'he',
```
johannesjo, line 16: **Important (scope):** This makes Hebrew detectable from the browser/OS and renders it RTL, but the hard-coded Settings language picker has no `HE` option, and runtime detection does not persist a selection. The submitted head is therefore auto-detection-only.

Please either make that limited scope explicit and track manual selection separately (the KISS option), or address compatibility before adding the conventional picker entry. `localization.lng` is synced; a released pre-Hebrew client rejects `"he"`, repairs it to the default, and can emit a full-state `REPAIR`. A schema bump or widening only the new client’s type would not protect released clients.

src/assets/i18n/he.json
```diff
@@ -0,0 +1,3209 @@
...
+        "IMPORTING": "מייבא נתונים",
+        "INCOMPLETE_CFG": "אישורי סנכרון חסרים.",
+        "INCOMPLETE_OP_SYNC": "הסנכרון אינו שלם: {{count}} קבצי פעולה נכשלו בהורדה",
+        "INCOMPLETE_REMOTE_OPERATIONS": "שינויים שהורדו לא יכלו להסתיים בהחלה.",
+        "INITIAL_SYNC_ERROR": "סנכרון ראשוני נכשל",
+        "INTEGRITY_CHECK_FAILED": "זוהתה בעיית תקינות נתונים.",
+        "INTEGRITY_TAMPER_DETECTED": "הסנכרון הופסק: השרת החזיר נתונים שנכשלו בבדיקת אבטחה.",
+        "INVALID_AUTH_CODE": "קוד האימות נדחה.",
+        "INVALID_OPERATION_PAYLOAD": "זוהתה פעולה לא תקינה.",
+        "LEGACY_FORMAT_DETECTED": "אי-תאימות בפורמט הסנכרון.",
```
johannesjo, line 1647: **Critical:** This non-empty translation suppresses the English fallback, while the runtime still presents the adjacent **Force overwrite** action. It removes the conditions that make the action safe and the warning that an older device can silently diverge.

The smallest safe fix is to set this value to `""` until the complete warning is translated. Please use the same fallback for other incomplete actionable recovery strings, including `S_TEST_FAIL_USER_NOT_FOUND` (missing `{{url}}` and the user-ID remedy), `INCOMPLETE_REMOTE_OPERATIONS`, `SPLIT_FORMAT_ENABLE_SETTING`, `STORAGE_QUOTA_EXCEEDED`, `VERSION_UNSUPPORTED`, and `WEB_CRYPTO_NOT_AVAILABLE`. No new fallback mechanism or broad translation rewrite is needed.

src/assets/i18n/he.json
```diff
@@ -0,0 +1,3209 @@
...
+      "TOGGLE_SIDE_NAV": "התמקד בניווט הצדדי",
+      "TOGGLE_SIDE_NAV_MODE": "החלף מצב סרגל צד דחוס/מלא",
+      "TOGGLE_TASK_VIEW_CUSTOMIZER_PANEL": "החלף פאנל סינון/הקבצה/מיון",
+      "TRIGGER_SYNC": "הפעל סנכרון (אם מוגדר)",
+      "ZOOM_DEFAULT": "זום ברירת מחדל (שולחן עבודה בלבד)",
+      "ZOOM_IN": "הגדל זום (שולחן עבודה בלבד)",
+      "ZOOM_OUT": "הקטן זום (שולחן עבודה בלבד)"
+    },
+    "LANG": {
+      "AR": "עברית / ערבית",
```
johannesjo, line 2550: **Required:** This value says “Hebrew / Arabic,” but the option still has `LanguageCode.ar`; selecting it loads Arabic, not Hebrew. Restore this label to `ערבית`. If manual Hebrew selection is deferred for compatibility, do not reuse or relabel the Arabic option.

src/app/core/locale.constants.ts
```diff
@@ -113,6 +114,7 @@ export const LocaleImportFns: Record<
   fa: () => import('@angular/common/locales/fa'),
   fi: () => import('@angular/common/locales/fi'),
   fr: () => import('@angular/common/locales/fr'),
+  he: () => import('@angular/common/locales/he'),
```
johannesjo, line 117: **Required (CI):** The current head fails the repository’s Prettier/ESLint check here because of the trailing spaces; the RTL array at line 54 also needs Prettier’s multiline formatting. Please run `npm run checkFile src/app/core/locale.constants.ts` and commit only those formatting changes.

johannesjo (changes requested):
Thanks for adding Hebrew and wiring the Angular locale and RTL support. I rechecked the complete PR head, not only the final one-line import. The `@angular/common/locales/he` module exists, the lazy import is correct, and `he` is correctly included in `RTL_LANGUAGES`.

I’m requesting changes for the focused items in the inline comments.

The smallest safe path is:

- Keep this PR explicitly auto-detection-only unless manual selection compatibility is addressed deliberately. No schema bump or broad sync redesign is warranted for a detect-only locale addition.
- Set incomplete safety/recovery translations to `""` so the documented English fallback preserves the full instructions. There is no need to rewrite the entire translation in this PR.
- Correct the Arabic label and run the required file check.

Non-blocking translation cleanup: the `**...**` markers at `he.json` lines 562, 2209, and 3064 are used by non-Markdown consumers and will render literally. The sole missing ordinary key (`F.TASK.CMP.CUSTOM_ESTIMATE`) already falls back safely to English, so I would not block this PR on it.

CI has not run yet; all workflows are currently awaiting approval.

Issue #9305, opened 2026-07-25
🚨 when the break reminder timer expires, Idle Handling detection does not work.


masifamu:
### Version Used

V18.6.0

### Current Behavior

I have set the break reminder for every X=1h30m, the problem is when it expires it starts showing a notification telling me to take break because you have been working for X hours. Right before the expiry if I go idle, make be taken a break, in those scenarios the system keeps on counting the time and adding it to the task, however this should not be.

I have been using the superproductivity app from last 3years, but never faced such issues, now recently I have switched to 18.6.0 and started having issues. I am using it on Ubuntu 22.04

### Expected Behavior

App should detect the Idle state and stops both the timers, task and break reminder timers.

### Steps to Reproduce

Set the break reminder timer for break reminder.
Set the Idle Handler to 5m
when the break reminder is about to expire, just leave the system idle

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

johannesjo:
Thanks for the detailed steps — they're genuinely useful, but I think the break reminder is the messenger here rather than the cause.

The break reminder can't gate idle detection. `triggerIdle` is dispatched from exactly one place, and the only things that suppress it are "already idle", "suppress during focus mode" (off by default, and only added in 18.15.0) and "only open when there is a current task" (off by default). The coupling actually runs the other way round: the reminder is suppressed *while* you are idle. So if the "you have been working for X" notification keeps re-appearing while you are away from the machine, then idle was never entered at all — which makes this an idle **detection** problem on your setup rather than a break-reminder interaction.

That also means upgrading on its own probably won't help: the Linux idle-detection code is byte-identical between 18.6.0 and 18.16.0, so 18.6.0 didn't introduce this. It points more at something changing on the system side — an X11 → Wayland session, or a switch to the Snap package.

Could you post the following from `~/.config/superProductivity/logs/main.log`?

- the line `Idle detection method ready: <method>`
- a couple of `🕘 Idle check (idle=…, method=…, action=…)` lines from while you were away

…plus how you installed it (deb / AppImage / Snap / Flatpak) and the output of `echo $XDG_SESSION_TYPE`.

[...]

Issue #6064, opened 2026-01-19
🚨 Without break timer doesn't reset


billy-thegreat:
### Version Used

v17.0.0.RC.12 deb

### Expected Behavior

Without break timer should reset during breaks and when no task is counting

### Current Behavior

Without break reminder triggers

### Steps to Reproduce

Turn on focus sync setting
Run two 25/5 pomodoro sessions until 1hr reminder triggers

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

PR #9351, opened 2026-07-28
fix(take-a-break): restore automatic break-timer reset and reminder teardown

johannesjo:
Two defects in `TakeABreakService`, both found while triaging #9305. Both are restorations of behaviour that was intended and silently lost — not new behaviour.

They are independent of that issue's main symptom (idle *detection* failing on the reporter's Ubuntu box — still awaiting logs), but they are the "going idle should also stop the break-reminder timer" half of it, and they affect everyone on Electron with default settings.

### 1. The automatic break-timer reset has been dead since v11.1.0

`_triggerProgrammaticReset$` branched on whether idle tracking was enabled:

```ts
isIdleResetEnabled
  ? this._actions$.pipe(ofType(triggerResetBreakTimer))
  : this._triggerSimpleBreakReset$;
```

`triggerResetBreakTimer` lost its last dispatcher in `aaca8d9657` (Jan 2025), which moved the reset into the idle-dialog result and left this waiting on an action nobody sends. Since `isEnableIdleTimeTracking` defaults to `true` on Electron, that silently disabled the `_triggerSimpleBreakReset$` fallback — "10 minutes with no tracked task counts as a break" — for essentially every desktop user, and for browser users once the extension connects. Turning idle tracking *off* was the only way to get it back, which is backwards.

This is also the unfixed half of #6064 ("Without break timer doesn't reset", expected *"should reset during breaks **and when no task is counting**"*), which was closed by a fix that only covered "during breaks".

Fix: drop the dead branch and the now-unreferenced action.

It is also now **edge-triggered**. `_timeWithNoCurrentTask$` grows monotonically, so a plain `filter` re-fired on every 1s tick for as long as the app stayed open — and `timeWorkingWithoutABreak$` is bound via `| async` at `work-view.component.html:65` behind an `@if` that is hardcoded true, so each emission cost a change-detection pass on the work view. Left open overnight with nothing tracked, that is ~28,800 wake-ups.

### 2. Two reset paths bypassed the reminder teardown

`_bannerService.dismiss(BANNER_ID)`, `_triggerLockScreenCounter$.next(false)` and `_triggerFullscreenBlocker$.next(false)` hung off `_triggerReset$`, but two real reset paths zeroed the counter through `_tick$` instead and never reached it:

- the idle dialog's "reset break reminder timer" / BREAK mode
- focus-mode breaks (`focus-mode.effects.ts`, `otherNoBreakTIme$.next(0)`)

So after taking a break through either, the stale "you have been working for X without one" banner stayed up. Worse: those two subjects are `distinctUntilChanged()`, so without the `next(false)` they stay latched at `true` — **lock screen and the timed fullscreen blocker go silently dead for the rest of the session** after the first break reminder, for users who have them enabled.

Fix: route every reset through `_triggerReset$`. Focus-mode calls `resetTimer()`; the idle dialog's reset request is forked out of `_tick$` into its own stream. The teardown is also no longer gated on `isTakeABreakEnabled` — dismissing a banner that cannot be open and un-latching subjects that cannot be `true` are both no-ops, whereas skipping the teardown when the feature is toggled off mid-session stranded those subjects at `true` for good, which is the exact state this code exists to prevent.

### Tests

The spec's `GlobalConfigService` stub used `isEnableIdleTimeTracking: false` — precisely the configuration in which defect 1 is invisible. Switched to the shipped default.

`resetBreakTimerOnBreakStart$` had **no coverage at all**, and four sibling focus-mode specs mocked `TakeABreakService` without `resetTimer`, so the routing change broke five tests in the #6064 spec. The edge trigger also needed a live `currentTaskId$` subject rather than `of(null)`, which completes — without that, a one-shot reset (`take(1)`) passes every test, which is the same silent-death shape as the bug being fixed.

Each guard was sabotage-verified individually:

| test | reverting the fix gives |
|---|---|
| counts a long stretch without a tracked task as a break | `Expected 10000 to be 0` |
| dismisses the reminder banner when the timer is reset | `dismiss` called 0 times |
| tears down only once while nothing is being tracked | `dismiss` called 5 times |
| re-arms after a task is tracked and stopped again | `dismiss` called 1 time, expected 2 |
| resets the break timer via resetTimer() when a break starts | `resetTimer` called 0 times |

Full local suite green (`npm test`, both TZ variants), `checkFile` clean on all touched files.

There is also now an **end-to-end test** (`e2e/tests/break/break-reminder-teardown.spec.ts`) covering the user-visible symptom: the reminder threshold is shrunk to 2s via the store bridge, a task is tracked until the real banner appears, a focus-mode break is started, and the banner must leave the DOM. Reverting focus-mode to `otherNoBreakTIme$.next(0)` fails it.

This is the feature's first real e2e coverage — the existing `tests/break/take-a-break.spec.ts` only walks settings pages, and its header explains why ("full break timing tests would require waiting for real time"). The assertion is on DOM presence rather than visibility on purpose: `startBreak` opens the focus-mode overlay, which would hide the banner without dismissing it, so a visibility assertion would pass even with the bug present.

Unrelated but worth knowing: `TaskPage.toggleTaskTimeTracking()` targets `.play-btn`, which no longer exists in the task template. It has no callers; the new test dispatches `[Task] SetCurrentTask` instead.

### Caveats worth knowing before merging

- **The lock-screen / fullscreen-blocker half of defect 2 is still unverified.** The banner half is now covered end to end, but `IS_ELECTRON` is a compile-time const (`app.constants.ts:5`), so both observables are `EMPTY` under Karma, `_triggerBanner$` guards the `next(true)` on it too, and the e2e suite runs the web build. The latching behaviour was checked with standalone rxjs harnesses only. Switching this service to the existing `IS_ELECTRON_TOKEN` would make it unit-testable; not done here to keep the diff contained. The direct manual check is: enable lock screen, set the reminder to 1 minute, track → banner + lock screen arms → "Pause & Break" → track again → confirm it arms a **second** time.
- **Defect 1 is a deliberate cadence change for every Electron user**: break reminders will fire less often, because pausing tracking for 10 minutes counts as a break again.
- **A non-positive tick still zeroes the counter** via the seedless `scan` — it just no longer tears the reminder down, so after a backwards clock step the banner can disagree with the counter. Pre-existing; needs a separate guard on the tick branch.
- **Open design question**: the untracked-stretch reset fires *during* an idle absence (the task is already deselected), so answering the idle dialog with "reset break reminder timer" explicitly unchecked no longer preserves the pre-idle counter. `1eace3799d` characterises this. Suppressing the reset while an idle absence is pending would keep that checkbox meaningful; not suppressing says "away 15+ minutes was a break regardless". Undecided.

### Deliberately not in this PR

- #9352 — the `idleDialogResult` branch computes `noBreakTime` with a `typeof t.time === 'number'` check that never matches the `'IDLE_TIME'` placeholder BREAK/TASK modes actually send, so it contributes `0`. Fixing it would switch on a path dormant for 18 months and interacts with a pre-existing double count, so it needs its own decision.
- #9356 — `LOCK_SCREEN_THROTTLE` / `FULLSCREEN_BLOCKER_THROTTLE` are dead code (the `throttleTime` sits inside a `switchMap` over a fresh single-value `of(v)`, so it never throttles).

Refs #9305

### DIFF
--- a/e2e/tests/break/break-reminder-teardown.spec.ts
+++ b/e2e/tests/break/break-reminder-teardown.spec.ts
@@ -1,0 +1,75 @@
+import { expect, test } from '../../fixtures/test.fixture';
+import type { Page } from '@playwright/test';
+
+/**
+ * The break reminder's teardown used to hang off `_triggerReset$` only, while
+ * focus-mode breaks zeroed the counter through `otherNoBreakTIme$` and never
+ * reached it — so taking a break left the "you have been working for X" banner
+ * on screen (and, on Electron, left the lock-screen / fullscreen-blocker
+ * subjects latched at `true` for the rest of the session).
+ *
+ * The sibling spec in this folder only exercises settings navigation; its own
+ * header notes that timing tests "would require waiting for real time". The
+ * store bridge removes that constraint — the reminder threshold is shrunk to a
+ * couple of seconds so the real banner can be driven end to end.
+ *
+ * See #9305.
+ */
+
+const dispatch = async (page: Page, action: Record<string, unknown>): Promise<void> => {
+  await expect
+    .poll(
+      () =>
+        page.evaluate((a) => {
+          const store = (
+            window as unknown as {
+              __e2eTestHelpers?: { store?: { dispatch: (a: unknown) => void } };
+            }
+          ).__e2eTestHelpers?.store;
+          if (!store) return false;
+          store.dispatch(a);
+          return true;
+        }, action),
+      { timeout: 10000 },
+    )
+    .toBe(true);
+};
+
+test.describe('Break reminder teardown', () => {
+  test('dismisses the reminder banner when a focus-mode break starts', async ({
+    page,
+    workViewPage,
+    taskPage,
+  }) => {
+    await workViewPage.waitForTaskList();
+
+    // remind after 2s of tracked work rather than the default hour
+    await dispatch(page, {
+      type: '[Global Config] Update Global Config Section',
+      sectionKey: 'takeABreak',
+      sectionCfg: { isTakeABreakEnabled: true, takeABreakMinWorkingTime: 2000 },
+      isSkipSnack: true,
+    });
+
+    await workViewPage.addTask('Break reminder task');
+    const task = taskPage.getTaskByText('Break reminder task');
+    await expect(task).toBeVisible();
+    const taskId = await task.getAttribute('data-task-id');
+    expect(taskId).toBeTruthy();
+
+    // start tracking via the store rather than the UI affordance: the play
+    // control lives behind a hover/context menu and is not what is under test
+    await dispatch(page, { type: '[Task] SetCurrentTask', id: taskId });
+
+    // NOTE: assert on DOM presence, not visibility. `startBreak` opens the
+    // focus-mode overlay, which would hide the banner without dismissing it —
+    // a visibility assertion would pass for the wrong reason.
+    const reminder = page.locator('banner mat-icon', { hasText: 'free_breakfast' });
+    await expect(reminder).toHaveCount(1, { timeout: 20000 });
+
+    // the reset path that used to zero the counter without tearing down the reminder
+    await dispatch(page, { type: '[FocusMode] Start Break' });
+
+    await expect(reminder).toHaveCount(0, { timeout: 10000 });
+  });
+});
--- a/src/app/features/focus-mode/store/focus-mode.bug-5875.spec.ts
+++ b/src/app/features/focus-mode/store/focus-mode.bug-5875.spec.ts
@@ -93,6 +93,7 @@ describe('FocusMode Bug #5875: Pomodoro timer sync issues', () => {
 
     const takeABreakServiceMock = {
       otherNoBreakTIme$: new BehaviorSubject<number>(0),
+      resetTimer: jasmine.createSpy('resetTimer'),
     };
 
     TestBed.configureTestingModule({
--- a/src/app/features/focus-mode/store/focus-mode.bug-5995.spec.ts
+++ b/src/app/features/focus-mode/store/focus-mode.bug-5995.spec.ts
@@ -90,6 +90,7 @@ describe('FocusMode Bug #5995: Resume paused break', () => {
 
     const takeABreakServiceMock = {
       otherNoBreakTIme$: new BehaviorSubject<number>(0),
+      resetTimer: jasmine.createSpy('resetTimer'),
     };
 
     TestBed.configureTestingModule({
--- a/src/app/features/focus-mode/store/focus-mode.bug-6064.spec.ts
+++ b/src/app/features/focus-mode/store/focus-mode.bug-6064.spec.ts
@@ -141,9 +141,6 @@ describe('FocusMode Bug #6064: Without break timer reset on break start', () => 
       // Subscribe to the effect (non-dispatching effect, so we just need to subscribe)
       effects.resetBreakTimerOnBreakStart$.subscribe();
 
-      // Spy on the Subject's next method
-      spyOn(otherNoBreakTime$, 'next');
-
       // Dispatch startBreak action
       actions$.next(
         actions.startBreak({
@@ -153,7 +150,8 @@ describe('FocusMode Bug #6064: Without break timer reset on break start', () => 
       );
       tick(10);
 
-      // Verify: otherNoBreakTIme$.next(0) was called
+      // Verify: the reset went through resetTimer(), which also tears the
+      // reminder down (otherNoBreakTIme$.next(0) only zeroed the counter)
       expect(takeABreakServiceMock.resetTimer).toHaveBeenCalled();
       expect(takeABreakServiceMock.resetTimer).toHaveBeenCalledTimes(1);
 
@@ -170,7 +168,6 @@ describe('FocusMode Bug #6064: Without break timer reset on break start', () => 
       store.refreshState();
 
       effects.resetBreakTimerOnBreakStart$.subscribe();
-      spyOn(otherNoBreakTime$, 'next');
 
       // Dispatch startBreak action
       actions$.next(
@@ -197,7 +194,6 @@ describe('FocusMode Bug #6064: Without break timer reset on break start', () => 
       store.refreshState();
 
       effects.resetBreakTimerOnBreakStart$.subscribe();
-      spyOn(otherNoBreakTime$, 'next');
 
       // Dispatch startBreak action
       actions$.next(
@@ -217,7 +213,6 @@ describe('FocusMode Bug #6064: Without break timer reset on break start', () => 
 
     it('should reset break timer for short breaks', fakeAsync(() => {
       effects.resetBreakTimerOnBreakStart$.subscribe();
-      spyOn(otherNoBreakTime$, 'next');
 
       // Dispatch short break (5 minutes)
       actions$.next(
@@ -235,7 +230,6 @@ describe('FocusMode Bug #6064: Without break timer reset on break start', () => 
 
     it('should reset break timer for long breaks', fakeAsync(() => {
       effects.resetBreakTimerOnBreakStart$.subscribe();
-      spyOn(otherNoBreakTime$, 'next');
 
       // Dispatch long break (15 minutes)
       actions$.next(
@@ -253,7 +247,6 @@ describe('FocusMode Bug #6064: Without break timer reset on break start', () => 
 
     it('should reset break timer multiple times across multiple breaks', fakeAsync(() => {
       effects.resetBreakTimerOnBreakStart$.subscribe();
-      spyOn(otherNoBreakTime$, 'next');
 
       // First break
       actions$.next(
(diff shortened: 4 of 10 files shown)


johannesjo:
## Review of `b88d913..106fdf61` (the refactor + the two test commits)

Verified against `106fdf61ba`. The service blob is identical at `d7f8224` and `106fdf61` (`afbdf0ceb1…`), so the e2e commit changes no production code. Branch merges cleanly onto master (12 behind, no conflicts).

**The refactor is correct and I'd merge it.** All three functional changes are sabotage-guarded, and the new e2e genuinely bites (fails at line 73 when focus-mode is reverted to `otherNoBreakTIme$.next(0)`). Findings below are one real coverage loss, two test gaps, and some tidying — none of them block.

---

### 1. A `duration: 0` tick zeroes the counter but no longer tears the reminder down

This is the one behaviour this commit gives up relative to `b88d913`. Same probe, same tree, only the teardown swapped:

```
d7f8224 (_triggerReset$-derived):  ZEROTICK before=7200000 after=0 dismissAfter=0
b88d913 (counter===0-derived):     ZEROTICK before=7200000 after=0 dismissAfter=1
```

The counter wipe is pre-existing in both — `scan((acc, value) => value > 0 ? acc + value : value)` treats any value ≤ 0 as a reset, and `_tick$` feeds that scan. What's new is that the banner now stays up saying "you have been working for 2h" while the counter reads 0.

The commit message calls this "an accidental non-positive tick … after a backwards clock step". It's more ordinary than that — on Android it's a normal-path event:

- `triggerWakeUpTick(cap)` → `Math.max(0, Math.min(rawDelta, maxDurationMs))` (`global-tracking-interval.service.ts:104`), so `cap === 0` emits `duration: 0` deterministically.
- `android-focus-mode.effects.ts:449` — `const cap = Math.max(0, timer.duration - timer.elapsed)` → **0 whenever a focus session completes at or over its duration**.
- `:303` — same computation on pause; reachable while running past duration because `focus-mode.reducer.ts:255-258` keeps `isRunning: true` when `_isOvertimeEnabled`.

Severity is capped: this is not a regression against **master** (master's teardown was also `_triggerReset$`-derived, so it behaved the same), and Android/iOS have no lock screen or fullscreen blocker, so the latched-subject flavour of #9305 isn't reachable down this path. But it's the same class of bug this commit exists to prevent, and it's an unremarked loss versus the previous commit. Either put `filter((v) => v > 0)` on the tick branch or file the follow-up the commit message promises.

### 2. The `isTakeABreakEnabled` gate removal is completely unguarded

Restoring the gate exactly as it was leaves the **entire** unit suite green — `TOTAL: 13543 SUCCESS`, and `13 of 13` in `take-a-break.service.spec.ts`. A control that inverts the gate to `!cfg.isTakeABreakEnabled` does fail (`5 FAILED`), so the build really is picking up the edit.

To be clear: **don't restore the gate** — the removal is right, and the reasoning in the comment holds. I checked the no-op claim end to end: `BannerService.dismiss` on an absent id returns the *same* array reference from `_banners.update(...)`, and Angular's `signalSetFn` guards on `Object.is`, so nothing notifies; `next(false)` always maps to `EMPTY` in both throttle chains. And the strand is real — feature on → banner → `next(true)` latches → toggle off → reset skipped → the subject is stuck at `true` and re-enabling can never re-fire it.

It just needs pinning. One spec with `takeABreak$: of({ isTakeABreakEnabled: false })` asserting `dismiss` is still called would do it.

### 3. The edge trigger changes counter semantics, not just change-detection cost

Both the code comment (lines 76-79) and the commit message justify `distinctUntilChanged()` purely as a CD saving. It also changes what the counter holds:

```
d7f8224:  afterDialog=660000  afterTick1=660000  afterTick2=660000
b88d913:  afterDialog=660000  afterTick1=0       afterTick2=0
```

After the untracked-stretch reset fires, a later addition within the same stretch now sticks; under the old plain filter it was annulled by the very next 1s tick, and by every tick after, since tracking hasn't resumed.

Reachable: `idle.effects.ts:187` deselects the task on idle init so the stretch grows during the absence; the reset box auto-checks in SPLIT only when a BREAK item is present (`dialog-idle.component.ts:386`, `_updateResetBreakTimerDefault`), so the user has to uncheck it deliberately; and `handleIdleDialogResult$` re-selects a task only when `taskItems.length === 1` (`idle.effects.ts:362`), so with 0 or 2+ task items the stretch keeps running.

This is an **improvement** — under the old code that checkbox was inert for any idle over 10 minutes, which is the same silent-death shape as #9305 itself. Two notes:

- It refines the PR's own "open design question": the checkbox is not entirely meaningless after this change. In the multi-task SPLIT case it now preserves the tracked time; only the pre-idle counter is unrecoverable.
- No test covers it. The nearest one (`lets the untracked-stretch reset win over an explicit opt-out`) uses the `'IDLE_TIME'` placeholder, which contributes 0, so it passes identically on both versions — the only existing tests that fail under the plain filter are the dismiss-*count* assertions.

### 4. Minor

- **9 dead spies** in `focus-mode.bug-6064.spec.ts` — `spyOn(otherNoBreakTime$, 'next')` at lines 145, 173, 200, 220, 238, 256, 295, 322, 344, with no `expect(otherNoBreakTime…)` left anywhere. Lint is clean, so they'll just rot.
- **Nothing asserts `resetTimer()` zeroes the counter.** The 6064 suite and the new `resetBreakTimerOnBreakStart$` test prove "the effect calls `resetTimer()`"; the `startBreak()` describe asserts `pauseCurrent` / snack / `dismiss` only. Pre-existing shape carried forward, one assertion closes it.
- **`_triggerReset$` is a cold `merge` subscribed twice** (constructor teardown + inside `timeWorkingWithoutABreak$`) — proven exactly two, and safe. Worth noting *why* it's safe, because the obvious reason is wrong: it is not that nothing emits before both subscriptions exist (`currentTaskId$` is a cold store select and does emit synchronously). It's that `shareReplay(1)` hands the second subscriber exactly the last value the first one saw, and `distinctUntilChanged` retains only its last value, so the two instances' states are necessarily identical. A `share()` would make that obvious rather than incidental. One live consequence: the subscriptions now have different lifetimes — `shareReplay` defaults to `resetOnError: true`, so the direct teardown survives an error that kills the counter chain (better than the previous commit), but "a reset fired the teardown" and "a reset zeroed the counter" are no longer guaranteed to be the same event.

---

### Checked and clean

Ordering on all three reset paths (teardown always precedes the counter; nothing re-opens). No lost-emission window for `resetTimer()` — its only callers are an effect and a component, both post-construction. All 12 combinations of `isResetBreakTimer` × trackItems shape produce byte-identical counter values and dismiss behaviour across the two commits. No `switchMap` cancellation hazard from the `EMPTY` branch. `checkFile` clean on both source files, prettier rewrites nothing. No dangling `triggerResetBreakTimer` references in src, e2e, electron or packages. `updateGlobalConfigSection` merges rather than replaces, so the partial `takeABreak` config in the new e2e can't trip `motivationalImgs.length`, and all three dispatched action types are real and load-bearing.

Sabotage matrix I ran independently:

| sabotage | result |
|---|---|
| drop `distinctUntilChanged()` | 2 specs fail (`dismiss` 5×/10× instead of 1×/2×) |
| idle-dialog reset back through `_tick$` as `of(0)` | 2 specs fail (`dismiss` 0 times) |
| focus-mode back to `otherNoBreakTIme$.next(0)` | 10 unit specs fail + the new e2e fails at line 73 |
| restore the `isTakeABreakEnabled` gate | **13/13 still pass** |

johannesjo:
All six findings from the review above are addressed in `7af9864a73`.

The one behavioural fix is `filter((duration) => duration > 0)` on `_tick$`, so only `_triggerReset$` can drive the counter to 0. That makes "a reset zeroes the counter *and* tears the reminder down" a property of the pipeline rather than of the current set of callers — and it closes the `otherNoBreakTIme$.next(0)` footgun this PR opened by removing that call's last caller.

| finding | resolution |
|---|---|
| `duration: 0` tick zeroed the counter without teardown | `filter((d) => d > 0)` on `_tick$` + zero/negative-tick specs |
| gate removal unguarded | spec with `isTakeABreakEnabled: false` (setup extracted to `configure(enabled)`) |
| edge trigger's semantic change undocumented | comment rewritten + spec pinning that opt-out time survives |
| nothing asserted `resetTimer()` zeroes the counter | spec added |
| 9 dead `spyOn(otherNoBreakTime$, 'next')` | removed, plus the two comments they orphaned |
| cold `_triggerReset$` subscribed twice | documented why it is deliberate, with a "don't `share()` this" note |

Each new guard was sabotage-verified in isolation:

| sabotage | fails |
|---|---|
| drop the positive-only filter | zero-tick + negative-tick |
| restore the config gate | feature-disabled teardown |
| drop `distinctUntilChanged` | opt-out-survives (+ the 2 existing dismiss-count tests) |
| `resetTimer()` → no-op | resetTimer counter test (+ 2 existing) |

`13548 SUCCESS` on `Europe/Berlin` and `13534` on `America/Los_Angeles`; the break-reminder e2e still passes; `checkFile` clean on all three files.

Two things I would not defend as minimal, left in deliberately rather than spend another CI cycle: the negative-tick spec exercises the same single predicate as the zero-tick one (it documents a distinct upstream cause — `consumeCurrentTick()` is unclamped), and the "don't `share()` this" comment documents code that is not broken.

**Still open, and it is not a code question:** the cadence change. Restoring `_triggerSimpleBreakReset$` means "10 minutes with nothing tracked counts as a break" is live again for every Electron user on default settings, after being dead since v11.1.0 — so break reminders will fire less often than they do today. That is the call worth making before merge.

PR #8560, opened 2026-06-23
Auto-link raw webexteams URIs

aakhter:
## Summary

- Allow `webexteams:` deep links through the shared external URL scheme allowlist.
- Auto-link raw `webexteams://...` URIs in Markdown-rendered notes while preserving query parameters.
- Extend the lightweight render-links pipe so plain text fields also make raw `webexteams://...` URIs clickable.

## Verification

- `npx prettier --check electron/shared-with-frontend/is-external-url-allowed.ts src/app/ui/is-external-url-allowed.spec.ts src/app/ui/marked-options-factory.spec.ts src/app/ui/marked-options-factory.ts src/app/ui/pipes/render-links.pipe.spec.ts src/app/ui/pipes/render-links.pipe.ts`
- `npx tsc -p src/tsconfig.app.json --noEmit --pretty false`
- Node smoke test for the shared allowlist plus Markdown raw Webex autolink path:
  `{"allowlistOk":true,"markdownRawWebexOk":true,"dangerousSchemeRejected":true}`
- Node smoke test for `RenderLinksPipe` raw Webex autolink path with a mocked sanitizer:
  `{"renderLinksRawWebexOk":true}`

## Note

I also attempted the focused Angular/Karma specs with `npx ng test --watch=false --include=...`. The sandboxed run failed before test execution because Karma could not bind port `9876` (`EPERM`), and the unsandboxed rerun hung during browser bundle setup without producing a spec summary.

PR #9359, opened 2026-07-28
fix(electron): allow SiYuan deep links

### LINKED ISSUE #9292: 💡 Allow `siyuan://` URL scheme in note sanitizer
### Problem Statement

Currently, adding SiYuan deep links (e.g. `[Note](siyuan://blocks/...)`) in task notes causes a "Link blocked: unsafe URL scheme" warning. The Angular DomSanitizer blocks the `siyuan://` protocol because it is not included in the whitelisted URI schemes, making it impossible to click and open SiYuan notes directly from Super Productivity tasks.

### Possible Solution

Add `siyuan://` to the list of allowed/whitelisted URL schemes in the note link sanitizer (similar to how `obsidian://` and `file://` are already handled), so that links to SiYuan blocks open seamlessly via the system's default protocol handler.

_No response_

### Additional context

_No response_

bakar-dev:
## Problem

Task-note links using the `siyuan://` protocol are rendered as inert text with the warning `Link blocked: unsafe URL scheme`.

This prevents users from opening SiYuan notes from their Super Productivity tasks.

Fixes #9292.

## Solution

- Add only the exact `siyuan:` protocol to the existing shared external URL allowlist.
- Add URL-policy regression coverage for a representative `siyuan://blocks/...` link.
- Verify that the Markdown renderer produces a clickable anchor for SiYuan links.
- Keep the existing blocking behavior for unsafe, malformed, and unapproved schemes unchanged.

### Before

The SiYuan link is rendered as blocked, inert text:

SiYuan link blocked before the change

### After

The same note is rendered as a clickable `siyuan://` link:

SiYuan link allowed after the change

I verified the browser rendering locally. I did not test launching the SiYuan desktop application because SiYuan is not installed on the test machine.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] No wiki change is relevant for this narrowly scoped allowlist addition; the behavior is documented inline.
- [x] I have run `npm run checkFile` on all changed `.ts` files.
- [x] I have added tests for my changes.
- [x] My commit message follows the Angular format (`type(scope): description`).

## Validation

- URL-policy tests: 64/64 passed.
- Markdown-renderer tests: 71/71 passed.
- Full application suite: 13,590 passed, 2 skipped.
- Alternate-timezone suite: 13,576 passed, 16 skipped.
- Electron tests: 243/243 passed.
- Full lint pipeline completed with zero errors.
- Existing unsafe and malformed URL schemes remain blocked.

### DIFF
--- a/electron/shared-with-frontend/is-external-url-allowed.ts
+++ b/electron/shared-with-frontend/is-external-url-allowed.ts
@@ -13,10 +13,11 @@
  * item) — a far narrower surface than the OS-level handlers this allowlist
  * exists to block. They are allow-listed because productivity users routinely
  * link tasks to such apps (#8429: obsidian:// links stopped opening after the
- * GHSA-hr87 fix; #8859: outlook: deep-links to mail items). The GHSA-hr87 fix
- * was a regression that silently broke every previously-working scheme at once,
- * so the popular note/task apps in this same low-risk class (notion:, things:,
- * omnifocus:, bear:, joplin:) are restored proactively alongside #8859.
+ * GHSA-hr87 fix; #8859: outlook: deep-links to mail items; #9292: siyuan://
+ * links to notes). The GHSA-hr87 fix was a regression that silently broke every
+ * previously-working scheme at once, so the popular note/task apps in this same
+ * low-risk class (notion:, things:, omnifocus:, bear:, joplin:) are restored
+ * proactively alongside #8859.
  * shortcut: a curated set — if a user needs a scheme that isn't here, the
  * upgrade path is a user-configurable allowlist in settings (MiscConfig).
  *
@@ -40,6 +41,7 @@ export const ALLOWED_EXTERNAL_URL_SCHEMES = [
   'omnifocus:',
   'bear:',
   'joplin:',
+  'siyuan:',
   'outlook:', // #8859 — Outlook desktop deep-links (outlook:<EntryID>)
   'webexteams:',
 ];
--- a/src/app/ui/is-external-url-allowed.spec.ts
+++ b/src/app/ui/is-external-url-allowed.spec.ts
@@ -29,6 +29,7 @@ describe('isExternalUrlSchemeAllowed', () => {
       'omnifocus:///task/abc123',
       'bear://x-callback-url/open-note?id=ABC123',
       'joplin://x-callback-url/openNote?id=abc123def456',
+      'siyuan://blocks/20240101000000-abcdefg',
       // #8859 — Outlook desktop deep-link: opaque EntryID, no `//` authority.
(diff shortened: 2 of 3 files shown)


johannesjo:
Thanks @bakar-dev — this is exactly the right shape for the fix, and asking on #9292 before building it made this easy to review.

Verified locally on `9c9b23c`:

**The one-line change is genuinely sufficient.** `ALLOWED_EXTERNAL_URL_SCHEMES` is the single source for the markdown link renderer (`marked-options-factory.ts:168`), the `RenderLinksPipe`, and both Electron `shell.openExternal` sinks (`main-window.ts:450`, `ipc-handlers/system.ts:18`) — so adding the scheme once covers the whole click path. And since `siyuan:` isn't `file:`, it correctly skips the `openLocalPath` branch and reaches `openExternal`.

**The tests aren't vacuous.** I removed `'siyuan:'` and re-ran: 2 failures in `is-external-url-allowed.spec.ts`, 1 in `marked-options-factory.spec.ts` — the latter failing with the literal reported symptom, `<span class="markdown-blocked-link" title="Link blocked: unsafe URL scheme">`. So the new coverage really does pin the reported bug. Restored: 64/64 and 71/71, plus 32/32 in `render-links.pipe.spec.ts` and 244/244 Electron. `checkFile` clean on all three files.

**Scheme surface checked.** For an allowlist that exists because of GHSA-hr87 it's worth knowing what the target app does with the URL, so I read SiYuan's own parser: `parseSiYuanUriInfo` (`app/src/util/pathName.ts`) accepts only `hostname === "blocks"` with a `/^\/\d{14}-\w{7}/` block ID, validates the `avItemID`/`avViewID`/`avGroupID` params against that same shape, and returns `null` for everything else. Narrow and well-bounded — squarely in the same low-risk class as the note-app deep links already on the list.

---

One scope note, **not** a change request: a *bare* `siyuan://blocks/…` typed as plain text still won't become a link. Auto-linking raw URIs is deliberately per-scheme — `URL_REGEX` covers only `https?|file|webexteams`, and marked has a single `webexteams` tokenizer (#8560). The `[Note](siyuan://…)` form from the issue works, which is what was asked for, so there's nothing to do here — just flagging it so it isn't a surprise.

The required E2E gates (`SuperSync E2E Gate` / `WebDAV E2E Gate`) still need to run before the ruleset will let this merge. They can't be affected by this diff — it touches no state or sync code, and the E2E suite doesn't reference the allowlist — but they're required.

Minor and easy to handle at squash time: the title scopes to `electron`, but the user-visible fix is in the Angular markdown renderer; the Electron sinks are defense-in-depth. Something like `fix(notes)` describes it a bit better.

johannesjo:
Thank you very much! <3

PR #9367, opened 2026-07-29
fix(tags): make tag-input placeholder actually visible and translate it

### LINKED ISSUE #9366: 🚨  Tag input placeholder is invisible and hardcoded
### Version Used

18.16.0 (reproducible on Web, Desktop/Electron, and Android — it's a shared Angular component/CSS issue, not platform-specific)

### Current Behavior

The tag input in the task-detail tag editor (and anywhere else `tag-edit` is used, e.g. the tag selection field) shows no visible text when empty. The input has a placeholder value of `+ Add Tag`, but it renders as blank space — there is no visible affordance telling the user they can type here to add a tag.

Separately, even if it were visible, the string is hardcoded English (`[placeholder]="'+ Add Tag'"` in `tag-edit.component.html`) and does not go through the translate pipe, so it would stay in English regardless of the selected locale.

### Expected Behavior

The placeholder text should be visible when the tag input is empty, and it should be translated according to the user's selected locale, consistent with every other label in the tag editor (e.g. the "Tags" label already uses the translate pipe).

### Steps to Reproduce

1. Open any task's detail panel.
2. Locate the Tags row/field.
3. With no tags added and the input empty, observe the input area — no placeholder text is visible.
4. Switch the app language to a non-English locale and repeat — the field remains blank either way, and the underlying string (if it were visible) would still be English.

### Console Output

```shell

```

### Documentation

ba6ul:
## Problem

The tag input's `+ Add Tag` placeholder is hardcoded English (`[placeholder]="'+ Add Tag'"` in `tag-edit.component.html`), so it doesn't respect the selected locale. It's also invisible even in English: Material only sets `opacity: 1` on `.mat-mdc-chip-input` placeholders when the input is inside a `.mat-mdc-form-field` ancestor, and this chip input isn't wrapped in one — so it falls back to the browser's low-contrast default and renders as blank space. Fixes #9366.

## Solution

- Replaced the hardcoded string with `T.F.TAG.D_EDIT.PLACEHOLDER | translate`, added the key to `t.const.ts` and `en.json` (other locales fall back automatically).
- Added an `::placeholder` override in `tag-edit.component.scss` to restore visibility, since this input intentionally isn't wrapped in `mat-form-field` (it's used inline in two tight-layout contexts — the task-detail-panel row and the formly tag-selection field — where Material's default form-field chrome would conflict with the existing custom, borderless input styling).

## Screenshots

**Before** (placeholder invisible):
Screenshot 2026-07-29 090947

**After** (placeholder visible):
Screenshot 2026-07-29 091052

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have included relevant changes to the documentation/wiki. (n/a — no docs reference this string)
- [ ] I have added tests for my changes (n/a — no existing spec covers this component; this is a visual/CSS + i18n fix)

### DIFF
--- a/src/app/features/tag/tag-edit/tag-edit.component.html
+++ b/src/app/features/tag/tag-edit/tag-edit.component.html
@@ -25,7 +25,7 @@
     [matChipInputFor]="chipListElRef"
     [matChipInputSeparatorKeyCodes]="separatorKeysCodes"
     (keydown)="onKeydown($event)"
-    [placeholder]="'+ Add Tag'"
+    [placeholder]="T.F.TAG.D_EDIT.PLACEHOLDER | translate"
   />
 </mat-chip-grid>
 <mat-autocomplete
--- a/src/app/features/tag/tag-edit/tag-edit.component.scss
+++ b/src/app/features/tag/tag-edit/tag-edit.component.scss
@@ -29,6 +29,12 @@
       outline: none;
       border-bottom-color: var(--extra-border-color);
     }
+
+    // MDC hides the placeholder by default unless a mat-form-field adds its
+    // "no-label" modifier class, which this bare chip input never gets.
+    &::placeholder {
+      opacity: 1 !important;
+    }
   }
 }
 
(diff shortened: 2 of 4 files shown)


johannesjo:
Thank you very much! <3

PR #9369, opened 2026-07-29
fix(task-widget): apply saved opacity after renderer loads on Windows/Linux

### LINKED ISSUE #9368: 🚨 The task widget's opacity gets reset after each lauch
### Version Used

v18.16.0W exe

### Current Behavior

If I set task widget's opacity to, say 50, then it become translucent, as expected, and stays at that opacity until I close the app completely.
But when I relaunch the app, the opacity becomes 100 again, only visually, the opacity slider still stays at 50. 
It works only after I slightly adjust the slider.

### Expected Behavior

The opacity should be the one that I set, and shouldn't reset after each launch.

### Steps to Reproduce

1. Turn on the Task Widget.
2. Set the opacity to 50
3. Now the Widget will be translucent
4. Now close the app completely from the hidden icons menu
5. Open the app again, Now the Widget is completely opaque

### Console Output

```shell

```

### Documentation

Apsinghsa:
## Problem

Closes #9368

On Windows and Linux, the Task Widget floating window ignores the saved opacity setting on app startup. The opacity slider in Settings correctly shows the saved value (e.g. 50%), but the widget always appears fully opaque visually. Adjusting the slider restores the correct opacity, confirming the value is saved — it just isn't applied on launch.

macOS is unaffected because it applies opacity via `BrowserWindow.setOpacity()`, a native OS-level call independent of the renderer.

## Solution

On Windows/Linux, the widget's opacity is driven by a CSS custom property (`--opacity`) updated via an IPC `update-opacity` message sent from the Electron main process to the widget renderer. The two existing calls to `updateTaskWidgetOpacity()` both fire during window creation — before the renderer page has finished loading and registered its `onUpdateOpacity` IPC listener in `task-widget-renderer.js`. Since `ipcRenderer` is a plain `EventEmitter` with no message replay, those sends are silently dropped, leaving the widget at the CSS default of `0.95` (fully opaque).

**The fix** adds a `webContents.on('did-finish-load', ...)` listener immediately after `loadFile()`. This is the earliest point at which the renderer's listener is guaranteed to be active, so the opacity IPC message is reliably received. `on` is used instead of `once` so that a DevTools-triggered page reload (F12 while the widget is focused) also correctly restores the opacity.

## Type of Change

- [x] Bug fix

## Checklist

- [x] I have added tests for my changes — regression test in `task-widget.test.cjs` that fails without the fix, verifies the clamped opacity value is delivered to the renderer after `did-finish-load`
- [x] Existing tests still pass (`npm run test:electron` — 230/231, the 1 failure is a pre-existing unrelated `jira.test.cjs` failure)

### DIFF
--- a/electron/task-widget.test.cjs
+++ b/electron/task-widget.test.cjs
@@ -24,6 +24,7 @@ const createDeferred = () => {
 
 class FakeWebContents {
   on() {}
+  once() {}
   send() {}
   focus() {}
   isDestroyed() {
--- a/electron/task-widget.test.cjs
+++ b/electron/task-widget.test.cjs
@@ -23,9 +23,23 @@ const createDeferred = () => {
 };
 
 class FakeWebContents {
-  on() {}
-  once() {}
-  send() {}
+  constructor() {
+    this.sent = [];
+    this._handlers = new Map();
+  }
+  on(eventName, handler) {
+    this._handlers.set(eventName, handler);
+  }
+  once(eventName, handler) {
+    this._handlers.set(eventName, handler);
+  }
+  emitOnce(eventName) {
+    const handler = this._handlers.get(eventName);
+    if (handler) handler();
+  }
+  send(channel, payload) {
+    this.sent.push({ channel, payload });
+  }
   focus() {}
   isDestroyed() {
     return false;
@@ -294,5 +308,29 @@ test('the closed event clears the sticky flag so it does not outlive the window'
     mod.getIsTaskWidgetUserForcedVisible(),
     false,
     'closing the window clears the sticky flag',
+  );
+});
+
+test('the opacity in effect at load time reaches the widget renderer once it finished loading', async () => {
+  const mod = loadModule();
+
+  mod.updateTaskWidgetEnabled(true);
+  await flush();
+
+  const { webContents } = createdWindows[0];
+
+  // The user's saved opacity arrives while the widget page is still loading.
+  // Sending it now is useless: the renderer only registers its `update-opacity`
+  // listener when task-widget-renderer.js runs, and ipcRenderer does not replay.
+  mod.updateTaskWidgetOpacity(5);
+  webContents.sent.length = 0;
+
+  webContents.emitOnce('did-finish-load');
+
+  assert.deepEqual(
+    webContents.sent.filter((m) => m.channel === 'update-opacity'),
+    // 5 % is below the 10 % floor, so the renderer must get the clamped value.
+    [{ channel: 'update-opacity', payload: 0.1 }],
+    'the opacity current at load time must reach the renderer exactly once, clamped',
   );
 });
--- a/electron/task-widget/task-widget.ts
+++ b/electron/task-widget/task-widget.ts
@@ -252,6 +252,13 @@ const createTaskWidgetWindowForGeneration = async (
 
   taskWidgetWin.loadFile(join(__dirname, 'task-widget.html'));
 
+  // Re-apply opacity once the renderer is ready to receive IPC (Windows only; macOS uses setOpacity()).
+  taskWidgetWin.webContents.once('did-finish-load', () => {
+    if (taskWidgetWin && !taskWidgetWin.isDestroyed()) {
+      updateTaskWidgetOpacity(currentOpacity);
+    }
+  });
+
   // Set visible on all workspaces immediately after creation
   taskWidgetWin.setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true });
 
@@ -304,6 +311,7 @@ const createTaskWidgetWindowForGeneration = async (
   // Update initial state
   updateTaskWidgetContent();
 
+  // macOS: setOpacity() works immediately; Windows: did-finish-load handler above handles it.
   updateTaskWidgetOpacity(currentOpacity);
 
   if (pendingShowAfterCreate) {
--- a/electron/task-widget/task-widget.ts
+++ b/electron/task-widget/task-widget.ts
@@ -252,11 +252,11 @@ const createTaskWidgetWindowForGeneration = async (
 
   taskWidgetWin.loadFile(join(__dirname, 'task-widget.html'));
 
-  // Re-apply opacity once the renderer is ready to receive IPC (Windows only; macOS uses setOpacity()).
-  taskWidgetWin.webContents.once('did-finish-load', () => {
-    if (taskWidgetWin && !taskWidgetWin.isDestroyed()) {
-      updateTaskWidgetOpacity(currentOpacity);
-    }
+  // Re-apply opacity once the page loads. On Windows/Linux opacity is a CSS variable driven
+  // by IPC; sends before did-finish-load are dropped. Uses 'on' (not 'once') so a DevTools
+  // reload also restores the correct opacity. macOS re-calls setOpacity() idempotently.
+  taskWidgetWin.webContents.on('did-finish-load', () => {
+    updateTaskWidgetOpacity(currentOpacity);
   });
 
   // Set visible on all workspaces immediately after creation
@@ -311,7 +311,8 @@ const createTaskWidgetWindowForGeneration = async (
   // Update initial state
   updateTaskWidgetContent();
 
-  // macOS: setOpacity() works immediately; Windows: did-finish-load handler above handles it.
+  // macOS: setOpacity() works immediately. Windows/Linux: IPC send is dropped before the
+  // page loads; the did-finish-load handler above re-delivers it reliably.
   updateTaskWidgetOpacity(currentOpacity);
 
   if (pendingShowAfterCreate) {


johannesjo:
## Review

Diagnosis and fix are both correct, and the fix sits at the right choke point. I verified the whole chain locally rather than taking the description at face value:

- **Root cause confirmed.** `TaskWidgetSettingsService` (`src/app/features/config/task-widget-settings.service.ts:27-40`) sends `UPDATE_TASK_WIDGET_SETTINGS` exactly once from its constructor and thereafter only on user change. So nothing ever re-sends the opacity — a dropped IPC is permanent for the session, which is precisely the reported symptom (slider shows 50, widget is opaque, nudging the slider fixes it).
- **`did-finish-load` is late enough.** `task-widget-renderer.js` is a classic `<script>` at the end of `<body>` (`task-widget.html:24`), so `onUpdateOpacity` is registered during parse, before the `load` event. No race.
- **Both creation paths are covered.** `updateTaskWidgetEnabled()` and the recreate inside `showTaskWidget()` both funnel through `createTaskWidgetWindowForGeneration()`, so there is one listener registration site and no gap.
- **macOS is unaffected in practice** — the handler just re-calls `setOpacity()` with the same value.
- The sibling `updateTaskWidgetContent()` at `task-widget.ts:312` drops its IPC the same way, but that one self-heals: `ready-to-show` asks the main window to re-send the current task. Correct to leave it alone.

`npm run test:electron` is green (244 tests), `npm run checkFile electron/task-widget/task-widget.ts` clean.

---

### Important — the fix ships untested

The only test change is a `once() {}` no-op stub. That stub *swallows* the handler, so the new branch never executes in the suite: the tests pass identically with and without the fix. Per the repo's policy a fix has to carry a test that fails without it.

The existing harness already mocks the Windows/Linux path (`IS_MAC: false`), so this is cheap. I wrote and verified the following — it passes with the fix, and removing the `did-finish-load` block makes it the **only** failing test (10 pass / 1 fail):

```diff
 class FakeWebContents {
+  constructor() {
+    this.sent = [];
+    this._onceHandlers = new Map();
+  }
   on() {}
-  once() {}
-  send() {}
+  once(eventName, handler) {
+    this._onceHandlers.set(eventName, handler);
+  }
+  emitOnce(eventName) {
+    const handler = this._onceHandlers.get(eventName);
+    if (handler) handler();
+  }
+  send(channel, payload) {
+    this.sent.push({ channel, payload });
+  }
   focus() {}
```

```js
test('the saved opacity is re-sent to the widget renderer once it finished loading', async () => {
  const mod = loadModule();

  // Mirrors app startup: the frontend sends its persisted settings in one
  // shot, so the opacity is cached before the window exists.
  mod.updateTaskWidgetOpacity(50);
  mod.updateTaskWidgetEnabled(true);
  await flush();

  const { webContents } = createdWindows[0];

  // Everything sent before the page finished loading is dropped by Electron
  // because the renderer has not registered its `update-opacity` listener yet.
  webContents.sent.length = 0;

  webContents.emitOnce('did-finish-load');

  assert.deepEqual(
    webContents.sent.filter((m) => m.channel === 'update-opacity'),
    [{ channel: 'update-opacity', payload: 0.5 }],
    'the cached opacity must reach the renderer after it finished loading',
  );
});
```

### Minor — redundant guard

`electron/task-widget/task-widget.ts:257` — `if (taskWidgetWin && !taskWidgetWin.isDestroyed())` duplicates the identical check at the top of `updateTaskWidgetOpacity()` (`:503-505`). The handler body can just be `updateTaskWidgetOpacity(currentOpacity)`. Purely cosmetic, take it or leave it.

---

**Verdict:** Needs changes — the code change itself is good to go; it just needs the regression test above so the behavior can't silently regress.

johannesjo:
## Review round 2 — re-checked with independent verification

I put my first review through four independent adversarial checks (including a real-Electron probe of the widget window). The fix still holds up, but **three things in my r1 were wrong or imprecise**, and there are a few findings I missed. Corrections first.

### Corrections to my r1

**1. This is not a Windows bug — Linux is affected identically.** `updateTaskWidgetOpacity()` branches on `IS_MAC` alone (`task-widget.ts:507-513`), and `IS_MAC = process.platform === 'darwin'` (`electron/common.const.ts:33`). Everything that isn't macOS takes the CSS-variable path, so Linux has exactly the same broken launch and exactly the same fix. This was confirmed empirically: a real Electron 41.4.0 harness driving the actual `task-widget.html/.css/-preload.js/-renderer.js` **reproduces #9368 on Linux**, 3/3 runs:

```
without fix -> --opacity = 0.95, container bg rgba(32,32,32,0.95)   <- the bug
with fix    -> --opacity = 0.5,  container bg rgba(32,32,32,0.5)    <- correct
```

Worth fixing in the PR title/description and, importantly, in the comments (see finding A).

**2. My description of the mechanism was sloppy.** I wrote that the message is "dropped by Electron because the renderer has not registered its listener". The delivery half is wrong: a listener registered in the **preload** does receive all three pre-load sends. The message reaches the renderer fine; it is lost because `task-widget-renderer.js` (an end-of-`<body>` script) has not run yet, and `ipcRenderer` is a plain EventEmitter with no replay. Same conclusion, but the accurate framing matters — it means "register/buffer in the preload" is also a valid fix shape.

**3. The test I handed you was weaker than I claimed.** It pins the value but not *when* the value is read. I verified this myself — with the handler changed to snapshot at registration time:

```ts
const snapshotOpacity = currentOpacity;                       // sabotage
taskWidgetWin.webContents.once('did-finish-load', () => { … updateTaskWidgetOpacity(snapshotOpacity); });
```
my r1 test still passed (11/11). And `50` is a clamp fixed point, so it could not catch a bypassed clamp either. Replacement below.

**4. "`updateTaskWidgetContent()` self-heals, leave it alone"** — the heal chain is real (`ready-to-show` → `REQUEST_CURRENT_TASK_FOR_TASK_WIDGET` → `task-electron.effects.ts:58` → `indicator.ts:294` → `updateTaskWidgetContent`) but `ready-to-show` fires at first non-empty layout, which can precede evaluation of `task-widget-renderer.js`. It is a narrow race, not a guarantee. Impact is small (a tracked task re-pushes within ~1s, and the static HTML defaults are already correct), so still not worth blocking — but if you move `updateTaskWidgetContent()` into the same `did-finish-load` handler you close it for free and the fix becomes symmetric.

---

### Replacement test

Please use this instead of the one in r1. It additionally pins that the handler reads the opacity **current at load time** (not snapshotted at registration) and that the clamp is applied. I verified: passes on the PR as-is; fails when the `did-finish-load` block is removed; fails under the snapshot variant above; full `npm run test:electron` **245/245**; `prettier --check` clean.

```diff
 class FakeWebContents {
+  constructor() {
+    this.sent = [];
+    this._onceHandlers = new Map();
+  }
   on() {}
-  once() {}
-  send() {}
+  once(eventName, handler) {
+    this._onceHandlers.set(eventName, handler);
+  }
+  emitOnce(eventName) {
+    const handler = this._onceHandlers.get(eventName);
+    if (handler) handler();
+  }
+  send(channel, payload) {
+    this.sent.push({ channel, payload });
+  }
   focus() {}
```

```js
test('the opacity in effect at load time reaches the widget renderer once it finished loading', async () => {
  const mod = loadModule();

  mod.updateTaskWidgetEnabled(true);
  await flush();

  const { webContents } = createdWindows[0];

  // The user's saved opacity arrives while the widget page is still loading.
  // Sending it now is useless: the renderer only registers its `update-opacity`
  // listener when task-widget-renderer.js runs, and ipcRenderer does not replay.
  mod.updateTaskWidgetOpacity(5);
  webContents.sent.length = 0;

  webContents.emitOnce('did-finish-load');

  assert.deepEqual(
    webContents.sent.filter((m) => m.channel === 'update-opacity'),
    // 5 % is below the 10 % floor, so the renderer must get the clamped value.
    [{ channel: 'update-opacity', payload: 0.1 }],
    'the opacity current at load time must reach the renderer exactly once, clamped',
  );
});
```

Note the mock hardcodes `IS_MAC: false` (`task-widget.test.cjs:111`), so this covers the non-macOS branch only — which is the branch with the bug.

---

### New findings

**A. The new comment is wrong about platforms.** `task-widget.ts:255` says *"(Windows only; macOS uses setOpacity())"* — but the handler is **not** Windows-only in either direction. It runs on macOS too (calling `setOpacity`), and the platforms it actually fixes are Windows **and Linux**. The companion comment at `:314` has the mirror-image problem: it labels a line that does nothing on Windows/Linux. On a platform-branchy file this will send the next person hunting a regression to the wrong OS. Suggest: *"non-macOS (Windows/Linux) applies opacity via a CSS variable in the renderer, so it must be re-sent once the page has loaded; macOS uses setOpacity() and is unaffected either way."*

**B. `once` → `on`.** There is no *programmatic* reload of the widget, but there is a user-reachable one: `debug.ts:131-132` registers `Ctrl+Shift+I` and `F12` (→ `devTools(BrowserWindow.getFocusedWindow())`) inside `app.on('ready')` and **outside** the `if (isAddReload)` guard, and `initDebug({ showDevTools }, IS_DEV)` (`start-app.ts:199`) never passes `enabled: false`, so the bail-out never fires in production. `electron-localshortcut`'s two-arg `register()` binds `ANY_WINDOW`, and the widget does take focus (`showTaskWidgetWindow` → `show()`). So: focus the widget → F12 → reload from DevTools → the page comes back at the CSS default `0.95` with the `once` already spent, i.e. straight back into #9368. (`Ctrl+R`/`F5` refresh is dev-only.) Obscure trigger, but `on` is a one-word change, the listener dies with the webContents, and re-sending is idempotent.

**C. There are now three places that apply opacity, and two of them are dead.** With the fix in place the order is:

```
loadFile → send/setOpacity (:315) → send/setOpacity (:63, the .then() block) → did-finish-load → send/setOpacity (:258)
```

`:315` and `:63` both run before the page has loaded, so on Windows/Linux both sends are lost — only the new handler does anything. On macOS all three are the same idempotent `setOpacity(0.5)`. In particular the re-apply at `:59-64` (inside the `.then()`, keep the rest of that block — it also fires `REQUEST_CURRENT_TASK_FOR_TASK_WIDGET`) carries a comment saying it exists because "`updateTaskWidgetOpacity()` is a no-op while `taskWidgetWin` is still null" — but `:315` was added inside the creation function later and already covers that case, so it has been redundant for a while and this PR makes it look load-bearing. Dropping those six lines would put the file back at two call sites instead of three. **Not a blocker and arguably out of scope for this PR** — the unit suite can't discriminate here either way, so it's your call whether to fold it in or file it separately. I'd just ask that the comments (finding A) make clear which call actually wins.

**D. Follow-up, not for this PR: the widget still flashes opaque at every launch.** The widget `BrowserWindow` is constructed with no `show: false` (contrast `main-window.ts:224`), so it is on screen before the page loads; the CSS default is `--opacity: 0.95` (`task-widget.css:14`) and `#task-widget-container` has `transition: background-color 0.3s` (`:32-33`). Since `did-finish-load` is by definition after first paint, a user at 30% now sees *transparent → snap to 95% → 300ms fade down to 30%* on every launch. Strictly better than the bug, and inherent to any post-load push. If you want it gone later, seeding the value before first paint — `loadFile(path, { query: { opacity: String(clamped) } })` plus two lines in the renderer — removes the race and the flash together (`currentOpacity` is already set synchronously by `applyTaskWidgetSettings` before `loadFile` runs). Worth a separate issue.

---

### Still standing from r1

The `if (taskWidgetWin && !taskWidgetWin.isDestroyed())` guard at `:257` is dead code — `updateTaskWidgetOpacity()` opens with the identical check at `:503-505`. Removing it keeps the suite green; no test can distinguish it.

Clean bill of health on the things I went looking for and didn't find: no lifecycle race (only three assignments to `taskWidgetWin`; two windows can never coexist; `destroyTaskWidget()` calls `webContents.removeAllListeners()` *before* `destroy()`, so a pending `once` is reliably removed), no macOS regression (idempotent re-`setOpacity`), `did-finish-load` is genuinely late enough (the renderer script is a classic end-of-`<body>` `<script>`, emitted as a plain global script by `tsconfig.electron.json`'s `"module": "commonjs"`, and the preload runs exactly once for the `file://` load with no earlier `about:blank` navigation), and `createTaskWidgetWindowForGeneration` really is the sole creation site.

**Verdict: needs changes** — swap in the replacement test, fix the platform wording in the two comments (A), and `once` → `on` (B). C and D are optional/follow-up. The fix itself is right and I'd be happy to see it merged with those in.

Apsinghsa:
Thanks for the thorough review! Addressed all the required points in the latest commit:

- **`once` → `on`** to survive a DevTools reload
- **Replaced test** with the stronger Round 2 version (pins load-time opacity + clamp, fails without the fix)
- **Removed redundant null/destroyed guard** inside the handler
- **Fixed both comments** to say Windows/Linux instead of Windows-only

Would appreciate another look when you get a chance!

johannesjo (commented):
## Review round 3

I re-checked the latest head (`cdaf71c`) against current `master`, including lifecycle/race analysis and mutation-testing the regression test. The production fix is correct, and all required round-two feedback is addressed: the listener now uses `on`, the handler reads the current opacity and applies the clamp, the redundant guard is gone, and the platform comments cover Windows/Linux accurately.

### Minor (non-blocking) — reload behavior is not regression-tested

`electron/task-widget.test.cjs:30-39, 328` — `FakeWebContents.on()` and `.once()` store handlers identically, while `emitOnce()` never consumes a once-handler. The test emits `did-finish-load` only once. I verified that changing production back from `on("did-finish-load", ...)` to `once(...)` still leaves the widget suite green (11/11), so the test protects startup delivery, current-at-load lookup, and clamping, but not the stated DevTools-reload recovery.

The smallest robust follow-up would be to use Node EventEmitter semantics in the fake and emit `did-finish-load` twice, asserting opacity delivery after both emissions. This is not a merge blocker: the current production implementation uses `on` correctly, and I found no production correctness, lifecycle, security, performance, or KISS issue.

Verification on the clean merge with `master` (`5a5e092`):

- `npm run checkFile electron/task-widget/task-widget.ts` — passed
- focused widget tests — 11/11 passed
- `npm run test:electron` — 280/280 passed
- removing the load handler makes only the new regression test fail

CI has not evaluated the current head yet: all seven workflows are `action_required`, and the visible preview was built from the older first commit (`59fb69f`). Once the current-head workflows are approved and pass, the code is suitable for approval without requiring another author change.

johannesjo:
Thank you very much! <3

PR #9370, opened 2026-07-29
fix(short-syntax): keep "@every weekday" off the weekend

KamilDev:
The `@every weekday 6am` case you spotted while reviewing #9328, as the standalone fix I offered there.

## The bug

`@every weekday 6am` typed on a Friday after 06:00: chrono's `forwardDate` slides the already-passed time to tomorrow, and `applyRepeatSyntax` classifies `MONDAY_TO_FRIDAY` as unanchored ("every workday is an occurrence, so chrono's slide to tomorrow is already the correct first one" — true Mon–Thu, false Fri–Sun), so it takes that date verbatim. Typed on a Saturday before 06:00 it keeps Saturday for the same reason.

## It is narrower than it looks — the config is fine, only the preview lies

Worth stating up front, because it sets the size of the fix. Probing the chain for `Standup @every weekday 6am` at Fri 19 Jan 2024 08:00:

```
parsed due:                    Sat Jan 20 2024 06:00
repeat:                        PRESET / MONDAY_TO_FRIDAY
cfg.startDate:                 2024-01-20   (Saturday)
getFirstRepeatOccurrence(cfg): Mon Jan 22 2024
```

The config's weekday flags make `getFirstRepeatOccurrence` skip the weekend, and **both** `addTaskRepeatCfgToTask` effects route the first occurrence through it — `addRepeatCfgToTaskUpdateTask$` for the timed path (`task-repeat-cfg.effects.ts:88`) and `updateTaskAfterMakingItRepeatable$` for the untimed one (`:168`). So the task that survives is scheduled **Monday 06:00**. The weekend date reached exactly two places:

- the add bar's date chip, which read "Sat 20, 6:00" while typing and up to submit — the input advertising a first occurrence the task never gets, which is the thing #9328's second half exists to prevent;
- the task's `dueWithTime` between `add()` and the effect, overwritten a tick later. `cfg.startDate` also stores the Saturday, but it is inert here: `WEEKLY` + `repeatEvery: 1` lets the flags decide, `lastTaskCreationDay` is set to the Monday, and the dialog hides start date for this preset.

So this is a preview bug, not a wrong recurrence — no migration, nothing to repair in existing data.

## The fix

One predicate is wrong, so the fix is where the predicate is: roll the previewed date forward past the weekend for `MONDAY_TO_FRIDAY`. The whole weekend, not just the slid day, so the Saturday-typed case lands on Monday too.

I did not route the preview through `getFirstRepeatOccurrence` instead. It would need a throwaway config built per keystroke, and it would move every preset's preview onto the engine's rules — the ~20 anchoring specs pin those to the parser's, and `MONDAY_TO_FRIDAY` is the only preset with an excluded-day set.

## Tests

Two specs, both verified failing without the fix (Saturday the 20th where Monday the 22nd is expected) and nothing else in the file moving:

- `should skip the weekend for "@every weekday 6am" typed on a Friday after 6am`
- `should skip the weekend for "@every weekday 6am" typed on a Saturday`

The existing `should not anchor "@every weekday 6am" to today's weekday` stays green — it runs on a Wednesday, where tomorrow is still a workday and chrono's slide is right.

Full `npm test` green in both timezones: 13686 passing / 16 skipped. `checkFile` clean on both changed `.ts` files. Wiki note added to the mapping line in `3.04-Short-Syntax.md`.


johannesjo (approved):
Reviewed at `53aa0f7bdb` on top of merged #9328 (`8ae317fe7d`). All 19 CI checks green. `checkFile` clean on both `.ts` files; `short-syntax.spec.ts` = **222 SUCCESS** in Berlin and Los Angeles.

The fix is right and the body's framing is accurate. One non-blocking nit at the end.

## The "config is fine, only the preview lies" claim — verified end to end

I rebuilt the whole chain in a throwaway spec — real `shortSyntax` → the cfg exactly as `add-task-bar.component.ts:604-625` assembles it → the real `getFirstRepeatOccurrence` — and ran it with the fix on and off. `SCHED` is what the effects actually schedule.

| typed (Europe/Berlin) | preview on master | preview with fix | SCHED (**identical both ways**) |
| --- | --- | --- | --- |
| `@every weekday 6am` Fri 07-31 08:00 | Sat 08-01 06:00 | Mon 08-03 06:00 | Mon 08-03 06:00 |
| `@every weekday 6am` Sat 08-01 05:00 | Sat 08-01 06:00 | Mon 08-03 06:00 | Mon 08-03 06:00 |
| `@every weekday 6am` Sat 08-01 08:00 | Sun 08-02 06:00 | Mon 08-03 06:00 | Mon 08-03 06:00 |
| `@every weekday 6am` Sun 08-02 05:00 | Sun 08-02 06:00 | Mon 08-03 06:00 | Mon 08-03 06:00 |
| `@every weekday 6am` Fri 07-31 05:00 | Fri 07-31 06:00 | unchanged | Fri 07-31 06:00 |
| `@every weekday 11pm` Fri 07-31 22:00 | Fri 07-31 23:00 | unchanged | Fri 07-31 23:00 |
| `@every workday 6am` Fri 07-31 08:00 | Sat 08-01 06:00 | Mon 08-03 06:00 | Mon 08-03 06:00 |
| `@daily 6am` / `@every friday 6am` / `@weekly 6am` | — | unchanged | unchanged |

So the scheduled result is byte-identical before and after in every case — the change really is preview-only, and "no migration, nothing to repair" holds. Confirmed the supporting details too: `getQuickSettingUpdates` returns no `startDate` key for `MONDAY_TO_FRIDAY` (`get-quick-setting-updates.ts:59-71`), so `cfg.startDate` is just `state.date`; `getFirstRepeatOccurrence`'s WEEKLY branch scans forward off the weekday flags; both effects route through it (`task-repeat-cfg.effects.ts:89`, `:169`); and the untimed effect writes `dueDay: firstOccurrenceStr` (`:196`), so the task lands on Monday either way. The dialog does hide start date for this preset (`SETTINGS_WITHOUT_START_DATE`, `dialog-edit-task-repeat-cfg.component.ts:627`).

Worth noting the `while` catches a third case the body doesn't name: **`@every weekday 6am` typed on a Sunday before 06:00** previewed Sunday on master. Fixed here too.

## Both specs are load-bearing

- Removing the `skipExcludedWeekend(due)` call → **exactly 2 FAILED** (`Expected 20 to be 22`), 220 untouched.
- `while` → `if` → **exactly 2 FAILED** (`Expected 21 to be 22`). So the "whole weekend, not just the slid day" claim has its own coverage.

`parsedDateResult.start.date()` returns a fresh `Date` per call (`chrono-node/dist/cjs/results.js:209-213` ends in `return new Date(...)`), so mutating it in place is safe. And `MONDAY_TO_FRIDAY` has a single producer (`short-syntax.ts:326-328`) which never sets `weekday`, so the preset can't reach the `anchorDate` branch — the guard sits on the only path it can take.

## Nit (non-blocking): the skip loses an hour across a DST spring-forward

`skipExcludedWeekend` steps *through* Sunday, and Sunday is the day DST changes. If the typed time falls in the spring-forward gap, `setDate` normalizes it and the shift sticks:

```
Europe/Berlin, spring-forward Sun 2026-03-29 (02:00 → 03:00)
"Standup @every weekday 2:30am" typed Fri 2026-03-27 08:00

master: preview Sat 03-28 02:30  → cfg.startDate 03-28, startTime "02:30" → SCHED Mon 03-30 02:30
  this: preview Mon 03-30 03:30  → cfg.startDate 03-30, startTime "03:30" → SCHED Mon 03-30 03:30
```

Sat 02:30 → `setDate(+1)` → Sun 02:30 doesn't exist → JS gives 03:30 → `setDate(+1)` → Mon 03:30.

This one *isn't* preview-only: `state.time` is read back off the parsed timestamp (`add-task-bar-parser.service.ts:191-203`), so the wrong hour is written into the repeat config's `startTime` and the task then recurs at 03:30 every workday. The frequency is negligible — one weekend a year, and only for a time inside the one-hour gap — so I would not block on it, but it's cheap to close:

```ts
const skipExcludedWeekend = (date: Date): void => {
  const hours = date.getHours();
  const minutes = date.getMinutes();
  while (date.getDay() === 0 || date.getDay() === 6) {
    date.setDate(date.getDate() + 1);
  }
  // A skipped-over day can be a DST spring-forward day whose 02:00-03:00 hour
  // does not exist; stepping through it would silently shift the clock time.
  date.setHours(hours, minutes);
};
```

Verified: restores `SCHED Mon 03-30 02:30` and the file stays at 222 SUCCESS.

(Unrelated and pre-existing, found while probing: `@every weekday 2:30am` typed on **Sat** 2026-03-28 parses to no date at all — chrono can't place 02:30 on the following Sunday. Nothing to do with this PR, just flagging it so the DST numbers above aren't confusing.)

## Scope note, not a request

The same "chip advertises a day the task never gets" divergence is still reachable without the syntax: set a Saturday date in the date picker, then pick *Monday to Friday* from the repeat menu. `applyUserRepeatPick` (`add-task-bar-parser.service.ts:418-422`) doesn't touch `state.date`, so the chip keeps Saturday while the config schedules Monday. Pre-existing, out of scope for a parser fix, and noted only because this PR's motivation ("the input advertising a first occurrence the task never gets") applies to it verbatim.

johannesjo:
Thank you very much! <3

PR #9380, opened 2026-07-29
fix(habit-tracker): roll the 7-day strip over on day change

### LINKED ISSUE #9072: fix(habit-tracker): 7-day strip freezes at first render and never rolls over at midnight
## Problem

`src/app/features/simple-counter/habit-tracker/habit-tracker.component.ts:74` builds its 7-day window from the wall clock inside a `computed`:

```ts
days = computed(() => {
  const isoTextLocale = this._dateTimeFormatService.isoTextLocale();
  ...
  const today = new Date();
  const offset = this.dayOffset();
  for (let i = 6; i >= 0; i--) { const d = new Date(); d.setDate(today.getDate() - i + offset); ... }
});
```

`new Date()` is not reactive, and Angular caches computeds. The only dependencies are `isoTextLocale()` and `dayOffset()` — neither advances with the clock, and the component reads no day-change or interval signal (no `todayDateStr$`, no `scheduleRefreshTick`, no `interval()`).

So the window freezes at first render. Leave the app open past midnight and the strip still shows **yesterday's** 7 days, with yesterday highlighted as "today", until something else invalidates it (navigating away and back, or paging via `dayOffset`).

Super Productivity is a desktop app people leave open for days, so this is the ordinary case rather than an edge case.

## Fix

Give the computed a day-change dependency, the way the rest of the app does — read `GlobalTrackingIntervalService.todayDateStr$` (via `toSignal`) inside `days`. That stream already merges timer/focus/visibility/resume sources specifically so day changes land even after sleep or tab throttling (see #5464).

## Context

Same root cause and shape as the `_contextNow` freeze fixed in #9064: a clock-derived value inside a `computed` with no time-varying dependency. An audit of all 374 `computed(() => ...)` bodies in `src/app` found 12 reading a non-reactive clock; 5 already take a tick, most of the rest are `new Date()` fallbacks in empty/no-data branches. This one and `isOverdue` (`task-detail-panel.component.ts:366`, where a task coming due while the panel is open won't flip to overdue until `task()` changes) are the two with real behaviour behind them.

Not verified at runtime — reported from reading the dependency graph.

MarkQuach12:
## Problem

Closes #9072. The 7 day strip is built in a `computed` whose dates come from `new Date()`, which is not reactive, so it renders once and never updates. Leave the app open past midnight and the strip still shows yesterday as the rightmost column, and clicking a cell writes the count to yesterday's date.

Using `new Date()` also ignored the start of next day setting, so the strip could disagree with the rest of the app about which day is today.

## Solution

Read `GlobalTrackingIntervalService.todayDateStr()` inside the computed and take the base date from `DateService.getLogicalTodayDate()`. That signal is the day change source the rest of the app already uses, so the strip also rolls over after sleep or tab throttling. No new timer, no new service.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/simple-counter/habit-tracker/habit-tracker.component.spec.ts
+++ b/src/app/features/simple-counter/habit-tracker/habit-tracker.component.spec.ts
@@ -1,4 +1,5 @@
 import { ComponentFixture, TestBed, fakeAsync, tick } from '@angular/core/testing';
+import { signal, WritableSignal } from '@angular/core';
 import { registerLocaleData } from '@angular/common';
 import localeSv from '@angular/common/locales/sv';
 import { HabitTrackerComponent } from './habit-tracker.component';
@@ -10,12 +11,16 @@ import { SimpleCounter, SimpleCounterType } from '../simple-counter.model';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { TranslateModule } from '@ngx-translate/core';
 import { EMPTY_SIMPLE_COUNTER } from '../simple-counter.const';
+import { GlobalTrackingIntervalService } from '../../../core/global-tracking-interval/global-tracking-interval.service';
+import { getDbDateStr } from '../../../util/get-db-date-str';
 
 describe('HabitTrackerComponent', () => {
   let component: HabitTrackerComponent;
   let fixture: ComponentFixture<HabitTrackerComponent>;
   let simpleCounterService: jasmine.SpyObj<SimpleCounterService>;
   let matDialog: jasmine.SpyObj<MatDialog>;
+  let logicalToday: Date;
+  let todayDateStr: WritableSignal<string>;
 
   const mockCounter: SimpleCounter = {
     ...EMPTY_SIMPLE_COUNTER,
@@ -39,13 +44,22 @@ describe('HabitTrackerComponent', () => {
       'deleteSimpleCounter',
     ]);
     matDialog = jasmine.createSpyObj('MatDialog', ['open']);
+    logicalToday = new Date('2026-05-18T10:00:00');
+    todayDateStr = signal(getDbDateStr(logicalToday));
 
     await TestBed.configureTestingModule({
       imports: [HabitTrackerComponent, NoopAnimationsModule, TranslateModule.forRoot()],
       providers: [
         { provide: SimpleCounterService, useValue: simpleCounterService },
         { provide: MatDialog, useValue: matDialog },
-        { provide: DateService, useValue: { todayStr: () => '2026-05-18' } },
+        {
+          provide: DateService,
+          useValue: {
+            todayStr: (date?: Date | number) => getDbDateStr(date ?? logicalToday),
+            getLogicalTodayDate: () => new Date(logicalToday),
+          },
+        },
+        { provide: GlobalTrackingIntervalService, useValue: { todayDateStr } },
         {
           provide: DateTimeFormatService,
           useValue: {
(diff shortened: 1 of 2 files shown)


johannesjo (approved):
Looks good. Minimal, correct, and the test actually pins the behaviour. Verified locally on `27be786`.

### What I checked

**The fix works, and the test proves it.** I sabotaged the component twice and re-ran the spec (`TZ=Europe/Berlin`, 7/7 green unmodified):

| sabotage | result |
| --- | --- |
| delete only `this._globalTrackingIntervalService.todayDateStr();` | `1 FAILED, 6 SUCCESS`, failing on the *second* assertion (`'2026-05-12'` vs `'2026-05-13'`), i.e. exactly the rollover |
| revert both lines to `new Date()` | `1 FAILED, 6 SUCCESS`, failing on the first assertion |

So the new spec isolates the rollover rather than incidentally failing.

**That spec is the only thing holding the dependency read in place.** I checked whether the toolchain would catch it if someone tidied line 84 away: it would not. `tsconfig.base.json` sets `noUnusedLocals: false`, and `checkFile` on the sabotaged file passes clean, so the now-unused `_globalTrackingIntervalService` field raises nothing outside the editor. The discarded-value call is invisible to lint and to `tsc`, and only the new test turns red. Worth knowing that the explanatory comment above it is carrying real weight.

**It reaches the DOM under zoneless.** Karma runs zone-ful but prod is `provideZonelessChangeDetection()` (`src/main.ts:267`), so I ran a throwaway probe with the zoneless provider and no manual `detectChanges()`, just `await fixture.whenStable()`. The rendered `.day-date` cells go `12…18` to `13…19` on the signal change, and with the dependency read removed the same probe fails (`'12'` vs `'13'`). Signal to template consumer to render works here. No need to add it to the PR, I just wanted to confirm the spec's `component.days()` assertion isn't hiding a rendering gap.

**The start-of-next-day half is a real second fix, and worth calling out.** `SimpleCounterService` writes counts under `DateService.todayStr()`, which applies the offset, in `setCounterToday` / `increaseCounterToday` / `decreaseCounterToday` (`simple-counter.service.ts:240, 272, 296`). Before this PR the strip built its keys from raw `new Date()`, so with e.g. a 04:00 start-of-next-day and the clock at 01:00, the header counter button incremented `2026-05-18` while the strip's rightmost cell wrote `2026-05-19`. `getLogicalTodayDate()` closes that. Might be worth a line in the commit body, since it is user-visible beyond the midnight freeze.

Also worth noting `todayStr(d)` is the right call for the per-cell keys here: it deliberately does *not* re-apply the offset to a passed-in date (documented on the method), and `d` is already derived from logical today, so there is no double shift.

**No DST regression from `new Date()` to `new Date(today)`.** This isn't the cumulative step loop that eats a day across a transition: each `d` is an independent clone with an absolute day-of-month, so nothing accumulates. I brute-forced it anyway, replicating the arithmetic in plain node over 770 days sampled every 30 minutes of local wall time, crossed with three start-of-next-day offsets (0, 2h, 4h) and three page offsets (0, -7, +7). That is 332,640 combinations per zone across São Paulo, Santiago, Havana, Beirut, Lord Howe, Berlin, Los Angeles, Auckland, Tehran and Chatham, asserting all 7 strings are consecutive calendar days and that the rightmost equals logical today. 0 failures in all 10 zones.

To confirm that sweep can actually fail, I re-ran it with the loop replaced by a cumulative `+24h` step: 624 failures in Berlin, 576 in Santiago, 288 on Lord Howe, all of the shape `["2025-10-26","2025-10-26","2025-10-27",…]` around the fall-back. So the 0 is meaningful, not a vacuous oracle.

**The test mock got more faithful, too.** The old `{ todayStr: () => '2026-05-18' }` ignored its argument, so every one of the 7 cells resolved to the same key under test. The new mock honours the date argument and mirrors the real signature, which makes the whole file able to see per-day behaviour. The pre-existing tests are unaffected by the swap: they pass explicit date strings to the component methods, and the weekday-header assertion uses `arrayWithExactContents`, which holds for any 7 consecutive days.

**Cost is nil.** `todayDateStr$` is `distinctUntilChanged` plus `shareReplay(1)`, and the service's `toSignal` already exists (added to avoid 200+ subscriptions), so `days()` recomputes on day change only, not per tick. `@for … track day.str` keeps six of the seven columns' DOM. `dateRangeLabel()` derives from `days()`, so the header label rolls over with it.

### Two notes, neither blocking

1. **#9072's other case gets closed along with this.** The issue body also flags `isOverdue` (`task-detail-panel.component.ts:367`) as the second one "with real behaviour behind it". It still reads `Date.now()` / `getDbDateStr()` with no day-change dependency, so a task coming due while the detail panel is open won't flip. `Closes #9072` would have buried that, so I have split it out into #9393. Nothing for you to do here, just noting why that issue exists.

2. **Paged state now slides.** With `dayOffset` at `-7`, a midnight crossing shifts the displayed week by a day (the window stays "7 days ending N days before today", which remains true). Previously it was frozen, so it happened to stay pinned. I think sliding is the right reading, just flagging it as a behaviour change for the paged case that no test covers.

Minor: there is no "today" highlight in the strip (no `is-today` class in the template or the SCSS), so the pre-fix symptom was the quieter "rightmost column is yesterday and clicking it writes to yesterday" rather than a mislabelled highlight. Doesn't change anything about the fix.

Approving.

johannesjo:
Thank you very much! <3

PR #9381, opened 2026-07-29
feat(ui): add empty states for schedule and board panels

### LINKED ISSUE #8822: 💡 Consistent, helpful empty states (Schedule, Boards, Planner)
## Problem

Empty-state quality is wildly inconsistent across the main views:

- **Work view**: has the only designed empty state (illustration + CTA + "all done" celebration) — this is the bar. ✅
- **Schedule**: no empty branch at all (`src/app/features/schedule/schedule/schedule.component.html`) — a brand-new user sees a blank calendar grid with no hint what fills it.
- **Boards**: the default Eisenhower board greets users with implementation jargon: *"2 new tags need to be created for this board to work"*, centered in a large empty canvas.
- **Board panel**: its empty state exists but is **commented out** (`src/app/features/boards/board-panel/board-panel.component.html:79-81`).
- **Planner**: every empty day column shows a *double* empty state — "No tasks" + Add button **and** an alarm icon + "No scheduled items" — repeated across all visible columns (screenshot below shows 4× each).

## Suggested fix

One quiet, user-language empty state per surface: what this view is for + the one action that fills it. For Planner, collapse the two messages into one per column (or show them only on the focused column). For Boards, replace the tag-mechanics message with what the board does ("Sort your tasks by urgency and importance — this creates 2 tags for you").

---

MarkQuach12:
## Problem

Closes #8822. The schedule view shows a bare grid when nothing is scheduled, with no hint about what fills it. Board panels had an empty state written but commented out, so an empty panel is just a blank rectangle. The Boards missing-tag message was written in implementation terms ("2 new tags need to be created for this board to work").

The other two items in the issue no longer reproduce, Planner's double empty state was already removed by #9248.

## Solution

Reused the existing `.empty` pattern (a plain div with muted small text) instead of adding a shared component. The schedule view gets one when the current range has no events, the board panel gets its commented-out block turned on, and the Boards message was reworded to say what the board does.

One new en.json key, `F.SCHEDULE.EMPTY`: "Tasks with a due date or time show up here." Two keys reworded: `F.BOARDS.V.CREATE_NEW_TAG_MSG` and `F.BOARDS.V.CREATE_NEW_TAGS_MSG`.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks — this is the right shape for #8822: no new shared component, no new setting, three small edits and two specs that genuinely fail when the feature is removed. I ran it against the preview build and found one layout regression worth fixing before merge, plus some wording thoughts.

Also confirming your triage so it doesn't get re-litigated later: `F.PLANNER.PLAN_VIEW.NO_SCHEDULED_ITEMS` is dead now — #9248 (`274f84db73`) wrapped the alarm section title in `@if (day.scheduledIItems.length > 0)`, so the Planner double empty state really is gone. Good call leaving it alone.

---

### 🟡 The schedule hint clips the month grid — exactly while it's visible

The hint is a normal flow child of the host, above `.scroll-wrapper`, and it measures 50px (`padding: var(--s2)` top + bottom + one line). That's 50px taken off the scroll wrapper.

Month view can't absorb it, because the grid is sized against the viewport rather than its parent — `schedule-month.component.scss:12`:

```scss
.month-grid-container {
  height: calc(100vh - 120px);
}
```

The chrome above `.scroll-wrapper` is a fixed 100px (top bar + `.schedule-nav-controls`, `min-height: 48px`), so:

- without the hint: wrapper `100vh - 100`, grid `100vh - 120` → **20px of slack, no scrollbar**
- with the hint: wrapper `100vh - 150`, grid `100vh - 120` → **30px of overflow**

Both terms are fixed px against the same `100vh`, so it's exactly 30px at any window height. Measured on the preview build (September 2026, nothing scheduled), toggling only the hint's `display` and changing nothing else:

| `.empty` | wrapper client | grid content | overflow | scrollbar |
| --- | --- | --- | --- | --- |
| hidden | 1515 | 1495 | **0** | no |
| shown | 1465 | 1495 | **30** | yes |

Net effect: whenever the empty hint appears in month view, a vertical scrollbar appears with it and the bottom ~30px of the last week row is cut off. So the polish lands and the grid gets slightly worse at the same moment. (I only measured desktop width — the `xs` block uses a different budget, `calc(100vh - 150px)` at line 260, so mobile may or may not have the same problem; worth a look.)

Week and day view are unaffected in practice: that grid already scrolls (content 2625 vs wrapper 1500), so the hint just costs 50px of an already-scrolling view — nothing new is clipped.

Two ways out, depending on how much you want to touch:

1. **Minimal** — make the hint an overlay so it costs no layout height. It also reads better centered in the empty canvas than wedged under the toolbar. `:host { position: relative }` + `position: absolute` on `.empty` with `pointer-events: none` so click-to-create still works underneath. Downside: the top offset has to clear `.schedule-nav-controls`, which is a `min-height`, so it's another magic number.
2. **Cleaner** — the `calc(100vh - 120px)` is really the underlying bug; the hint just exposed it. Sizing the month grid to its parent instead of the viewport would make it immune to anything added above. Bigger change though, and probably its own PR.

Either is fine by me — the constraint is just that the hint must not shrink `.scroll-wrapper`.

### 🟢 `monthEvents()` as the emptiness source

The signal is right: it's range-scoped in all three modes, and I confirmed the behaviour — on the August grid the hint correctly stays hidden because the leading Jul 30 cell holds a task, and it appears on September. But the name says "month" while it now also drives week and day view, which is a small trap for whoever reads this next. A named computed would carry the intent:

```ts
readonly hasNothingScheduled = computed(() => !this.monthEvents().length);
```

### 🟢 Boards wording

Renders as *"This board sorts your tasks into 2 tags that don't exist yet."* + `[Create Tags]`. Clear improvement over "…need to be created for this board to work". Two thoughts, take or leave:

- Still tags-first, whereas the issue asked for what the board *does*. The string is shared across boards so it can't say "by urgency and importance" — but something like *"This board groups your tasks by tag. 2 of its tags don't exist yet."* stays generic while leading with the purpose.
- Strictly, the board doesn't sort tasks *into* tags — you tag tasks, the board lays them out. Minor, but "sorts into" slightly oversells it.

### Nits

- `EMPTY` is appended after `VIEW_WEEK` in both `t.const.ts` and `en.json`. That block is alphabetical apart from one already-misplaced run (`NEXT_*`/`PREVIOUS_*`/`PLAN_TASK_FOR_DAY_PLACEHOLDER`/`VIEW_*`), so `EMPTY` belongs between `DROP_TASKS_HERE_TO_SCHEDULE` and `END`.
- `F.SCHEDULE.NO_TASKS` — "Currently there are no tasks. Please add some tasks via the plus (+) button in the top bar." — is already dead (`npm run int:unused` lists it). After this PR there are two schedule empty-state strings and one of them is unreachable. Dropping it here would be a nice tidy-up.
- `schedule`, `board-panel` and `planner-day` now each define their own `.empty` rule with different styling. Not worth extracting yet, but the PR body's "reused the existing `.empty` pattern" is really "same class name, new styles".

### What I verified

- Ran both changed specs (89 green), then sabotaged in both directions: removing the two `@if` blocks → 2 failures; forcing both conditions to `true` → 2 failures. The specs are real, and the board-panel one has a proper negative control.
- Preview build `3ba8ebc4`, fresh profile: hint shows in week and month view on an empty schedule, disappears once a task with `@today 3pm` exists, board panels show "No tasks", Boards shows the reworded missing-tags message.
- The layout numbers above were measured in-page twice, at two window sizes, toggling only the hint's `display`.
- Planner claim re-checked against the `#9248` diff and the unused-key report.

MarkQuach12:
@johannesjo the month grid clipping is fixed, ready for another look.

The hint is now an overlay, so it costs no layout height. I went with `position: absolute; inset: 0` centred over the host rather than a top offset, which avoids the magic number you flagged as the downside of this option. `pointer-events: none` keeps click-to-create working underneath. Since it no longer occupies a row, the `xs` budget at line 260 is covered too.

Added a regression test asserting `.scroll-wrapper` does not move when the hint appears. With the overlay reverted it fails `Expected 48 to be 98`, which is the 50px you measured.

The remaining nits from your review are not addressed yet. Happy to fold in the `hasNothingScheduled` rename, the `EMPTY` key ordering, dropping the dead `F.SCHEDULE.NO_TASKS`, and the boards rewording if you want them in this PR.

johannesjo:

Thanks for the follow-up. The overlay does solve the clipping I measured last round – `.empty` is out of flow now, so `.scroll-wrapper` keeps its full height and the month grid stops overflowing, on `xs` as well. The new regression spec is a good one. But the overlay introduced a different problem in the same view.

### Blocking

- **`src/app/features/schedule/schedule/schedule.component.scss:56`** – the overlay has no `z-index`, so in month view the hint is painted *behind* the grid and never appears.

  `.empty` is `position: absolute` with `z-index: auto` on a `:host` that is `position: relative` – which does not create a stacking context. `.scroll-wrapper` doesn't create one either (plain `overflow`, no `z-index`/`transform`/`contain`), so its positioned descendants hoist into the host's stacking context. `.month-day-cell` (`schedule-month.component.scss:61`) is `position: relative; background: var(--bg)` and comes *later* in tree order than `.empty`, so by CSS painting order the opaque day cells paint on top of it.

  I reproduced it in headless Chromium on a minimal page mirroring `:host` / `.scroll-wrapper` / `.month-grid-container` / `.month-day-cell`: `elementFromPoint` at the centre of the hint returns `month-day-cell`, and it still does with `pointer-events` forced to `auto`, so it's paint order rather than the `pointer-events: none`. The screenshot shows no text at all. Adding `z-index: 1` flips both: `elementFromPoint` returns `empty` and the text renders over the grid.

  Week and day view are unaffected – `.col`, `.row` and `.grid-container` are all transparent, so nothing opaque paints over the hint there. It's month view specifically, which is the view this round was about.

  Fix is one line:

  ```scss
  .empty {
    position: absolute;
    inset: 0;
    z-index: 1;
    // …
  }
  ```

  Worth extending the spec too, because all three of the new `empty state` tests pass while the hint is invisible – two assert only `querySelector('.empty')` and the third only `offsetTop`, and none of them can see paint order. Something like an `elementFromPoint` assertion at the hint's centre (with `pointer-events` temporarily set to `auto`) would fail without the `z-index` and keep it from regressing.

---

johannesjo:

Thanks for the PR! The follow-up keeps the empty hint out of layout flow, but its paint order leaves the message hidden in month view.

### Blocking

- **`src/app/features/schedule/schedule/schedule.component.scss:56`** – `.empty` is an absolutely positioned earlier sibling of `.scroll-wrapper`. The later positioned month cells have opaque backgrounds, so they paint over the centered hint. The new specs only check DOM presence and `offsetTop`, so they pass while the text is covered. Move the overlay after the scroll wrapper in paint order, or otherwise establish a suitable stacking layer, and add a visibility assertion that fails when a cell covers the hint.

---
<sub>Automated review pass (Codex). Anything unclear or wrong – say so and I'll take a look.</sub>

Issue #6197, opened 2026-01-26
💥 Invalid clientId loaded: B_H8AR


ibrahimishfaq0026-lgtm:
### Steps to Reproduce

1. Open Settings > Sync > Enable Syncing > Sync Provider > LocalFile.

2. Click "Sync folder Path", Select folder to use for sync. Then click okay. (mysteriously, when i select the folder to sync and then press okay, when coming back to productivity app, "Sync Provider" is red'ed out and below it is text shown "This Field is required")

3. Again Click on "Sync Provider" dropdown > LocalFile > Enable Syncing. (Now the error appears)

(Edit: The same error occurred when i tried to sync DROPBOX in the "Sync Provider" dropdown. Whenever i press the sync bottom on the top right, The same error comes in too.)

### Additional Console Output

log.ts:304 [sync] Error: Invalid clientId loaded: B_H8AR
    at t.<anonymous> (meta-model-ctrl.ts:271:13)
    at Generator.next (<anonymous>)
    at k (chunk-2NYJ54WT.js:1:1331)

### Url
file:///Applications/Super%20Productivity.app/Contents/Resources/app.asar/.tmp/angular-dist/browser/index.html#/config

### Meta Info
META: SP16.9.4 __ Electron – en-GB – MacIntel – en-GB – UA:Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) superProductivity/16.9.4 Chrome/138.0.7204.251 Electron/37.10.3 Safari/537.36

### Stacktrace
```
ClientIdNotFoundError (file:///Applications/Super%20Productivity.app/Contents/Resources/app.asar/.tmp/angular-dist/browser/src/app/pfapi/api/model-ctrl/meta-model-ctrl.ts:271:12)
C (file:///Applications/Super%20Productivity.app/Contents/Resources/app.asar/.tmp/angular-dist/browser/chunk-2NYJ54WT.js:1:1331)
```

### Actions Before Error
```
1769424526763: [Task] SetSelectedTask 
1769424526880: [SP_ALL] Load(import) all data 
1769424526883: [SP_ALL] All Data was loaded 
1769424577290: [Task] SetSelectedTask (2) 
1769424633030: [Global Config] Update Global Config Section
```

johannesjo:
This probably will be fixed with v17?

ibrahimishfaq0026-lgtm:
> This probably will be fixed with v17?


[...]

Issue #6142, opened 2026-01-24
💥 Invalid clientId loaded: B_2Oke


Clay-TC:
### Steps to Reproduce

1.
2.
3.

### Url
file:///C:/Users/clay_/AppData/Local/Programs/superProductivity/resources/app.asar/.tmp/angular-dist/browser/index.html#/config

### Meta Info
META: SP16.9.4 __ Electron – zh-CN – Win32 – zh-CN – UA:Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) superProductivity/16.9.4 Chrome/138.0.7204.251 Electron/37.10.3 Safari/537.36

### Stacktrace
```
t.<anonymous> (file:///C:/Users/clay_/AppData/Local/Programs/superProductivity/resources/app.asar/.tmp/angular-dist/browser/src/app/pfapi/api/model-ctrl/meta-model-ctrl.ts:271:12)
C (file:///C:/Users/clay_/AppData/Local/Programs/superProductivity/resources/app.asar/.tmp/angular-dist/browser/chunk-2NYJ54WT.js:1:1331)
```

### Actions Before Error
```
1769259048766: [Task] SetSelectedTask 
1769259048931: [SP_ALL] Load(import) all data 
1769259048933: [SP_ALL] All Data was loaded 
1769259066966: [Task] SetSelectedTask (2) 
1769259159772: [Global Config] Update Global Config Section
```

PR #9382, opened 2026-07-29
fix(util): client-ID validator regression: {4}→{6} tightening silently

### LINKED ISSUE #9355: Client-ID validator regression: {4}→{6} tightening silently rotates ids minted on v18.7.0–v18.10.0
## Problem

`isValidClientIdFormat` was tightened from a 4-char to a 6-char suffix **in the same hunk that widened the generator**, with no backward-compatibility clause. Client IDs minted by v18.7.0–v18.10.0 are therefore rejected by v18.11.0+, and because an invalid format is deliberately treated as *absent* rather than fatal, the affected device silently **mints a new client ID over the stored one** on its first launch after updating.

A client ID is a vector-clock key. This is a silent identity rotation — the exact non-regenerable-identity loss #7732 exists to prevent.

## Root cause

`ec16757c82` — the squash carries the subject `fix(keyboard): resolve macOS global shortcut layout mismatch (#8378) (#8381)`, but its body includes `fix(client-id): increase entropy to 6 chars and fix related test regressions`. It changed both sides at once in `src/app/core/util/generate-client-id.ts`:

```diff
-  return `${_getEnvironmentId()}_${_generateBase62(4)}`;
+  return `${_getEnvironmentId()}_${_generateBase62(6)}`;
...
-    typeof id === 'string' && (id.length >= 10 || /^[BEAI]_[a-zA-Z0-9]{4}$/.test(id))
+    typeof id === 'string' && (id.length >= 10 || /^[BEAI]_[a-zA-Z0-9]{6}$/.test(id))
```

Widening the generator is fine. Narrowing the *validator* is not, because it also has to accept ids already on disk.

## Affected population

- The 4-char compact format was introduced by `508998c6a1` ("Improve on sync (#7736)"), first released in **v18.7.0**.
- The `{6}` tightening first released in **v18.11.0**.

So any install whose client ID was first minted on **v18.7.0, v18.8.0, v18.9.0, v18.9.1 or v18.10.0** holds a 4-char id and is affected. Legacy pre-op-log ids (length ≥ 10) are unaffected — the `id.length >= 10` clause still covers them. Only the compact format has the gap.

## Consequence chain (traced in source)

```
isValidClientIdFormat('A_1ile')            → false   (length 6, so neither `>= 10` nor `{6}` matches)
client-id.service.ts:165  _readSupOps      → null    ("absent", not "unrecognized")
client-id.service.ts:190/194  _readPf      → null
client-id.service.ts:~120 _resolve         → null
client-id.service.ts:94   getOrGenerate    → _putClientIdIfAbsent(generateClientId)
client-id.service.ts:217                   → isValidClientIdFormat(raced) ? raced : factory()
                                             raced is the stored 4-char id → invalid → factory()
client-id.service.ts:218-220               → put(newId)   // stored id overwritten
```

Note the guard at `:217` is documented as the "MULTI-TAB / ROTATION GUARD … never clobbers it". It does not fire here, because the stored id fails the *format* check rather than being absent — the two cases are indistinguishable by design (#6197).

Demonstration:

```
A_1ile     len 6   valid: false
A_dh3J     len 6   valid: false
B_FXMz     len 6   valid: false
A_f3IT7K   len 8   valid: true
B_0oqz0T   len 8   valid: true
```

## Effect

A **one-time** rotation per install (after the overwrite the new id is valid and persists — this is not per-launch churn). Consequences:

- The device gets a new vector-clock key; its own prior operations remain attributed to the old key.
- The dead old key survives in every peer's clock until pruning evicts it — a concrete instance of the churn catalogued in #9105, which lists reinstalls, profiles and destructive flows as the sources but not this one.
- Ops missing the dead id's entry compare CONCURRENT against stored clocks that still carry it, costing extra conflict round-trips.

## Field evidence (circumstantial)

The sync logs attached to #9332 show three Android ids for one user: `A_1ile` and `A_dh3J` (4-char, both now invalid) alongside `A_f3IT7K` (6-char, valid). That is consistent with exactly this rotation having occurred, though reinstalls or multiple devices would look the same — I am not claiming it as proof.

## Severity (honest)

**Low-to-moderate, and mostly historical.** v18.11.0 shipped 2026-06-17, so nearly every affected device has already rotated once and is now stable. The remaining exposure is:

- installs that have not launched since before v18.11.0 and update later,
- restores of a backup or a legacy IndexedDB snapshot carrying a 4-char id.

Filing it because the failure is silent, it is a one-character fix, and the "narrowed a validator without a compat clause" pattern is worth having on record.

## Suggested fix

```diff
-    typeof id === 'string' && (id.length >= 10 || /^[BEAI]_[a-zA-Z0-9]{6}$/.test(id))
+    typeof id === 'string' && (id.length >= 10 || /^[BEAI]_[a-zA-Z0-9]{4,6}$/.test(id))
```

Generation stays at 6 chars; only acceptance widens. A test pinning `isValidClientIdFormat('B_FXMz') === true` would prevent a repeat — the current spec covers only the 6-char and legacy cases.

A stale copy of the original pattern still exists at `src/app/pfapi/api/model-ctrl/meta-model-ctrl.js:473` (`/^[BEAI]_[a-zA-Z0-9]{4}$/`, commented "Accept … compact format (e.g., \"B_FXMz\")"), which corroborates that 4-char ids were real.

## What I did not verify

- Not reproduced on a real device with a pre-v18.11.0 profile; the chain above is traced through source plus the regex demonstration.
- I did not establish how many installs actually minted a compact id inside the v18.7.0–v18.10.0 window.
- `ClientIdService.persistClientId` (`:109`) *throws* on an invalid format, and `operation-log-migration.service.ts:255` calls it. I could not construct a case where it receives a 4-char id — `legacyPfDb.loadClientId()` reads the pre-op-log PFAPI key, whose ids are long-format — so I am **not** claiming a hard-failure migration path, only flagging it as worth a second look.

Found while reviewing #9353. Related: #9105, #7732, #6197.

AlvaroBalbin:
This is a small, targeted change to src/app/core/util/generate-client-id.spec.ts, src/app/core/util/generate-client-id.ts that resolves the reported issue without touching anything unrelated.

Fixes #9355.

The repo's own CI runs green on this branch; I ran it on my fork before opening this.

### DIFF
--- a/src/app/core/util/generate-client-id.spec.ts
+++ b/src/app/core/util/generate-client-id.spec.ts
@@ -59,6 +59,14 @@ describe('generate-client-id', () => {
       expect(isValidClientIdFormat('I_ZzZzZz')).toBeTrue();
     });
 
+    it('accepts the 4-char compact format minted by v18.7.0 - v18.10.0', () => {
+      // Those ids are still on disk. Rejecting one reads as "absent", which
+      // mints a new id over it — a silent vector-clock key rotation. #9355.
+      expect(isValidClientIdFormat('B_FXMz')).toBeTrue();
+      expect(isValidClientIdFormat('A_1ile')).toBeTrue();
+      expect(isValidClientIdFormat('E_0000')).toBeTrue();
+    });
+
     it('accepts legacy ids of length >= 10', () => {
       expect(isValidClientIdFormat('LongClientId123')).toBeTrue();
       expect(isValidClientIdFormat('0123456789')).toBeTrue();
--- a/src/app/core/util/generate-client-id.ts
+++ b/src/app/core/util/generate-client-id.ts
@@ -89,6 +89,8 @@ export const generateClientId = (): string => {
  * Type guard: true if `id` matches a known valid client-ID format.
  * - Legacy format: any string of length >= 10 (legacy IDs).
  * - New format: {platform}_{6-char-base62}, e.g. "B_a7Kx9Z".
+ * - The 4-char suffix v18.7.0 - v18.10.0 minted, e.g. "B_FXMz". Only the
+ *   generator moved to 6 chars; those ids are still on disk and stay valid.
  *
  * Used to narrow `unknown` values read from IndexedDB. An invalid format is
  * treated as "absent" rather than fatal — see issue #6197.
@@ -103,6 +105,7 @@ export const generateClientId = (): string => {
  */
 export const isValidClientIdFormat = (id: unknown): id is string => {
   return (
-    typeof id === 'string' && (id.length >= 10 || /^[BEAI]_[a-zA-Z0-9]{6}$/.test(id))
+    typeof id === 'string' &&
+    (id.length >= 10 || /^[BEAI]_([a-zA-Z0-9]{4}|[a-zA-Z0-9]{6})$/.test(id))
   );
 };


johannesjo:
Thanks for picking this up! Unfortunately this one had already been fixed on `master` a few hours before the PR was opened, so I'm going to close it as already-resolved.

`e922f0c44` — "fix(sync): stop rejecting clientIds minted by older releases" (#9373) — merged 2026-07-29 12:44 UTC and replaced the `{6}` regex outright:

```ts
export const isValidClientIdFormat = (id: unknown): id is string => {
  return (
    typeof id === 'string' &&
    (id.length >= 10 ||
      (id.length >= MIN_CLIENT_ID_LENGTH && USABLE_CLIENT_ID_CHARSET.test(id)))
  );
};
```

All three ids your test adds (`B_FXMz`, `A_1ile`, `E_0000`) already pass there, and `generate-client-id.spec.ts` carries regression coverage for the same population (`B_H8AR` from #6197, `B_2Oke` from #6142).

Your branch is based on `6cae887235`, i.e. master from before that merge — that's why the diff still shows the old `{6}` regex and why GitHub reports the PR as conflicting.

One bit of context in case it's useful for a future PR: the shipped fix went wider than `{4}|{6}` on purpose. An enumerated whitelist is the shape of the original bug — it has to be edited every time the generator or the platform prefix changes, and forgetting to do so silently rotates a device's vector-clock key. The predicate now asks "can the system *use* this id?" instead of "did *this build* mint it?", so `B_a7Kx9` (5-char suffix) and `X_a7Kx9Z` (unknown prefix) are accepted too. Both are asserted in the spec, and the `[BEAI]_({4}|{6})` alternation here would reject them.

Thanks again for the report-to-fix instinct — sorry it collided with a fix that was already in flight.

PR #9384, opened 2026-07-29
fix(task): make move-to-backlog shortcut leave the Today view

### LINKED ISSUE #9374: Moving a task to project backlog from Today view via keyboard shortcut does not work
### Version Used

18.15.1AI (deb on Debian 13)

### Current Behavior

When I select a task (highlighted) from the Today view and then press CTRL+B, the previous task is selected.

### Expected Behavior

It should instead activate the "Move task to project backlog" because that is what is mapped. The task should be moved to the project backlog and disappear from the Today view.

NOTE: This behavior works fine from within the project view, so the symptom appears to be specifically related to the Today view.

### Steps to Reproduce

1. Select task from Today view
2. Press CTRL+B
3. Observe symptom -- previous task is selected; current task does not get moved to backlog

### Console Output

```shell

```

### Documentation

pratyushsinghal7:
## Problem

Pressing the move-to-backlog shortcut on a task in the Today view only moved focus to the previous task; the task itself stayed put. The dispatch actually ran, but it only moves the id between `project.taskIds` and `project.backlogTaskIds`, and Today membership is schedule-based (`dueDay`/`dueWithTime`), so the position-only move is invisible there.

Fixes #9374.

## Solution

When the shortcut fires on the Today list, unschedule the task in addition to the backlog move so it actually leaves the view. If the project's backlog is disabled the reducer drops the move, so the shortcut now no-ops entirely there instead of shifting focus (or silently unscheduling). Project-view behavior is unchanged and stays schedule-preserving (#8592).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

## Checklist

### DIFF
--- a/src/app/features/tasks/task/task-shortcuts.spec.ts
+++ b/src/app/features/tasks/task/task-shortcuts.spec.ts
@@ -1,4 +1,4 @@
-import { signal } from '@angular/core';
+import { signal, WritableSignal } from '@angular/core';
 import { TestBed, fakeAsync, tick } from '@angular/core/testing';
 import { Location } from '@angular/common';
 import { MatDialog, MatDialogState } from '@angular/material/dialog';
@@ -12,6 +12,7 @@ import { DateService } from '../../../core/date/date.service';
 import { GlobalTrackingIntervalService } from '../../../core/global-tracking-interval/global-tracking-interval.service';
 import { LayoutService } from '../../../core-ui/layout/layout.service';
 import { GlobalConfigService } from '../../config/global-config.service';
+import { Project } from '../../project/project.model';
 import { ProjectService } from '../../project/project.service';
 import { TaskRepeatCfgService } from '../../task-repeat-cfg/task-repeat-cfg.service';
 import { TaskAttachmentService } from '../task-attachment/task-attachment.service';
@@ -32,6 +33,7 @@ describe('TaskComponent shortcut handling', () => {
   let taskServiceSpy: jasmine.SpyObj<TaskService>;
   let addSubtaskInputServiceSpy: jasmine.SpyObj<AddSubtaskInputService>;
   let storeSpy: jasmine.SpyObj<Store>;
+  let isTodayListSignal: WritableSignal<boolean>;
 
   const createSubTask = (title: string): TaskWithSubTasks =>
     ({
@@ -105,6 +107,7 @@ describe('TaskComponent shortcut handling', () => {
     );
     storeSpy = jasmine.createSpyObj<Store>('Store', ['dispatch', 'select']);
     storeSpy.select.and.returnValue(of(new Set<string>()));
+    isTodayListSignal = signal(false);
 
     await TestBed.configureTestingModule({
       imports: [TaskComponent],
@@ -191,6 +194,7 @@ describe('TaskComponent shortcut handling', () => {
           provide: WorkContextService,
           useValue: {
             isTodayList: signal(false),
+            isTodayListSignal,
           },
         },
         {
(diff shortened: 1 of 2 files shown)


johannesjo:
Thanks for the PR, and thanks to @Zuhef for the analysis on #9374 that it builds on.

I'd like to hold this one, for two separate reasons: the behaviour choice isn't settled yet, and the implementation has a hole even if we do settle on it.

## 1. The behaviour question is still open

#9374's analysis laid out three options and explicitly asked which one I wanted, precisely because #8592 settled the schedule-preservation question on purpose. This PR picks option 2 without that decision having been made.

Concretely, the code being re-added here is the code that 9e23832879 (#8603) deleted:

```ts
// before #8603
moveToBacklog(): void {
  const t = this.task();
  if (t.projectId && !t.parentId) {
    this._projectService.moveTaskToBacklog(t.id, t.projectId);
    if (this.isTaskOnTodayList()) {
      this.unschedule();
    }
  }
}
```

The new condition (`isTodayListActive()`, the Today *view* is open) is narrower than the old one (`isTaskOnTodayList()`, the task is scheduled for today), so it doesn't literally reproduce #8592's repro. But #8592's reporter asked for schedule preservation "for all methods, in both directions", and this restores the same silent mutation for anyone who does it from Today.

## 2. `unscheduleTask` destroys more than the PR describes

`handleUnScheduleTask` (task-shared-scheduling.reducer.ts:126) clears `dueDay`, `dueWithTime` **and `remindAt`**. So Shift+B on a task scheduled for today at 14:00 with a reminder silently drops the time and the reminder.

It also hits overdue tasks. The Today view renders `selectOverdueTasksWithSubTasks` through the same `<task-list>` (work-view.component.html:281), so `isTodayListActive()` is true for them and the shortcut wipes the original past due date. #8592's report called that case out specifically as one that must be preserved.

And because the task leaves `TODAY_TAG.taskIds`, `sectionSharedMetaReducer`'s post-reducer diff (section-shared.reducer.ts:568) strips it from its Today section permanently. Re-scheduling it for today does not bring the section back.

None of these three are covered by a test or mentioned in the PR body.

## 3. The `isEnableBacklog` guard has a sibling case it misses

The comment says the guard exists so the shortcut doesn't "silently unschedule" when the move would be dropped. But the move is also dropped when the task is *already* in the backlog:

```ts
// project.reducer.ts:512
return backlogIdsBefore.includes(taskId)
  ? state
  : projectAdapter.updateOne(...)
```

I confirmed with a throwaway reducer spec that this returns the identical state object. So for a task that sits in the project backlog *and* has `dueDay === today`, the shortcut degrades to a pure unschedule: schedule and reminder destroyed, nothing moved.

That state is reachable, and #8603 is what makes it common: after #8592, moving a today-scheduled task to the backlog from the project view keeps `dueDay === today`, and Today membership is computed purely from `dueWithTime`/`dueDay` with no backlog filter (`computeOrderedTaskIdsForToday`, work-context.selectors.ts:42), so the task is still sitting in Today afterwards.

## 4. The fix is scoped to one of the views where the symptom occurs

`moveToBacklogWithFocus` is dispatched by `TaskShortcutService` for the focused task regardless of view. In any regular tag view the position-only move is just as invisible as it is in Today, so Shift+B there still produces exactly the reported symptom: focus jumps, nothing else happens. This PR fixes Today only.

## A leaner option

The app already takes a position on this: the "Move to backlog" context-menu item is gated on `isShowMoveFromAndToBacklogBtns$` = `!!activeWorkContext.isEnableBacklog` (task-context-menu-inner.component.ts:180), and `isEnableBacklog` is only ever populated for PROJECT contexts (work-context.selectors.ts:217). So in Today and in every tag view the menu deliberately does not offer this action, and only the shortcut leaks it through.

Gating the shortcut on that same condition would:

- kill the reported symptom (the phantom focus jump) in Today *and* in every tag view,
- keep one meaning for the command instead of two,
- touch no task data.

It doesn't give the reporter the outcome they asked for (task disappears from Today), so it's a real trade-off rather than a strictly better answer. But if the goal is "the shortcut shouldn't do something confusing", that's the smaller change.

## What is good here

The tests are real: I reverted each of the three behavioural branches independently and each one turns exactly one spec red. `checkFile` passes. The per-task project lookup correctly handles a Today list containing tasks from projects with different backlog settings. If we do land option 2, this is close to the right shape, it just needs the already-in-backlog case, the overdue case, and the reminder loss addressed.

johannesjo:
Thanks for the patience on this one @pratyushsinghal7 — I've made the call and it goes a different way, so I'm closing it.

The bug in #9374 is real, but the fix shouldn't be "Shift+B also unschedules". Since this PR was opened I resolved the mirror case in #9563 the other way, and the rule I settled on there is that a shortcut must not overload a position intent with a schedule intent — which is exactly what unscheduling on move-to-backlog does.

What I'll do instead: gate the shortcut on `isEnableBacklog`, the same condition the context menu already uses for the move-to-backlog item. That makes the shortcut agree with the menu that's shown right next to it, changes no task data, and removes the half-action where the task moves but stays visible.

Not a reflection on the work — the analysis in here was sound and it's what made the trade-off clear. The branch also conflicts now and CI is red, so a fresh small change is less work than reviving it.

Issue #6020, opened 2026-01-15
🚨 Sensitive personal data exposed in debug/crash JSON


zenoprax:
### Version Used

v16.8.3

### Expected Behavior

Logs intended for diagnostics or bug reporting should be safe by default to share publicly. At a minimum:

- No direct PII should be emitted.
- No personalized/private data should be emitted without obfuscation.
- High-entropy identifiers (ie. Task or Project IDs or names) should be masked or replaced. For example: `taskId: "7c2a9f..."` becomes `taskId: "TASK_3"` or `taskId: "<redacted>"`.
- Optional “full” logging modes should require explicit opt-in and be clearly labeled as unsafe to share.

### Current Behavior

`ANONYMIZED-super-productivity-crash-user-data-export.json` is not truly de-anonymized and exposes quite a few types of information that someone would assume would be kept private.

The following types of data are emitted in plain text:

- Email addresses (e.g. embedded in CalDAV / Google Calendar ICAL URLs) (search for `"resourceName":"`)
- Account identifiers tied to third-party services

Less identifiable, but still private:

- Project and Tag group names in full (search for `"name":"`)
- Exact timestamps and detailed activity timelines (work patterns, reminders, breaks)
- Third-party service usage details (CalDAV, WebDAV, Google Calendar endpoints)
- Custom text for motivational prompts and Domina mode

Behaviour Fingerprinting:
- Persistent internal IDs (task IDs, project IDs, reminder IDs, repeatCfgIds)
- Implicit time-zone and daily routine inference

Even when obvious identifiers (like email addresses) are removed, the remaining data still contains *private* data that has no debugging function. Naturally, there is a limit to how much obfuscation can be done but the JSON generated (`ANONYMIZED-...-data.json`) suggests that it is safe to include in GH Issues or Pastebins even if that is not the intended place for it.

### Steps to Reproduce

Select the "PE" option in the crash handler to generate the file.

Logs not attached to this issue for obvious reasons.

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
Thanks for bringing this up. We should change the resource names.

I'm less sure about timezone data (important for debugging sometimes) and ids (performance costly to replace them all). How are people supposed to use this data in any meaningfull way?

Also what would you propose? The ids need to be intact to use the data for any meaningfull debugging. We could replace them by hashes, but does that really accomplish much?

zenoprax:
[...]

PR #9385, opened 2026-07-30
fix(issue): show human-readable calendar names instead of hostnames

### LINKED ISSUE #9144: 🚨 Synced calendars appear as hostnames, not human-readable names
### Version Used

18.14.0SN (Flatpak)

### Current Behavior

When connecting to a CalDAV calendar, there is a mandatory field to provide a human-readable name:

Image

But as far as I tell, this name is never used. In the 'Issue Provider' Panel the top icons get an abbreviated name based on the hostname, and the URL is displayed below:

Image

When hovering, the full URL is displayed:

Image

Under 'Schedule' → 'Calendar Visibility', only the hostnames are displayed:

Image

### Expected Behavior

Use human-readable name to help disambiguate between multiple issue providers on the same host

### Steps to Reproduce

See above

### Console Output

```shell

```

### Documentation

pratyushsinghal7:
## Problem

Synced calendars are listed by URL, not by name. The Issue Provider panel and its tooltips show the raw CalDAV URL even though CalDAV setup already requires a human-readable calendar name. iCal feeds are worse: they carry no name at all, so both the Issue Provider panel and Schedule's Calendar Visibility menu show bare hostnames — with several calendars on one host (e.g. Google), the entries are indistinguishable.

## Solution

- CALDAV: label the provider (and its chip initials) with the configured `resourceName`, falling back to the URL when it is missing.
- ICAL: capture the feed's `X-WR-CALNAME` (the de-facto standard display-name property emitted by Google, Outlook, Nextcloud, and Apple) when the feed is fetched, persist it on the provider only when it actually changed, and prefer it for the label and chip initials. The sanitized hostname stays as the fallback.
- Mask the new `calName` field in the privacy export, same as `resourceName`.

Extraction is a small unfold-and-match on the raw feed text (RFC 5545 line unfolding and TEXT unescaping) so it works before and independently of event parsing.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

## Checklist

## Testing

- New specs for `getCalNameFromIcal` (folding, parameters, escaping, LF-only feeds, missing/empty values)
- New specs for the CALDAV/ICAL branches of `getIssueProviderTooltip` and `getIssueProviderInitials`
- Full pre-push suite: 13737 passing; the single failure is `create-blocked-blocks-by-day-map.spec.ts`, which also fails on master (hardcoded date)

Fixes #9144

### DIFF
--- a/src/app/features/calendar-integration/calendar-integration.service.ts
+++ b/src/app/features/calendar-integration/calendar-integration.service.ts
@@ -9,6 +9,7 @@ import {
   switchMap,
 } from 'rxjs/operators';
 import { getRelevantEventsForCalendarIntegrationFromIcal } from '../schedule/ical/get-relevant-events-from-ical';
+import { getCalNameFromIcal } from '../schedule/ical/get-cal-name-from-ical';
 import {
   BehaviorSubject,
   combineLatest,
@@ -40,6 +41,7 @@ import {
 } from '../schedule/schedule.model';
 import { getDbDateStr } from '../../util/get-db-date-str';
 import { selectCalendarProviders } from '../issue/store/issue-provider.selectors';
+import { IssueProviderActions } from '../issue/store/issue-provider.actions';
 import {
   IssueProviderCalendar,
   IssueProviderPluginType,
@@ -441,15 +443,16 @@ export class CalendarIntegrationService {
         },
       })
       .pipe(
-        switchMap((icalStrData) =>
-          getRelevantEventsForCalendarIntegrationFromIcal(
+        switchMap((icalStrData) => {
+          this._updateCalNameIfChanged(calProvider, icalStrData);
+          return getRelevantEventsForCalendarIntegrationFromIcal(
             icalStrData,
             calProvider.id,
             start,
             end,
             calProvider.icalUrl,
-          ),
-        ),
+          );
+        }),
         map((events) =>
           events.map((ev) => ({
             ...ev,
@@ -508,6 +511,26 @@ export class CalendarIntegrationService {
       Date.now() + ONE_MONTHS,
       isForwardError,
     );
+  }
+
+  /**
+   * Persists the feed's own display name (X-WR-CALNAME) on the provider so
+   * lists and tooltips can show "Work" instead of "calendar.example.com"
+   * (#9144). Dispatches only when the name actually changed — i.e. at most
+   * once per calendar rename — so regular polling causes no store churn.
+   */
+  private _updateCalNameIfChanged(
+    calProvider: IssueProviderCalendar,
+    icalStrData: string,
+  ): void {
+    const calName = getCalNameFromIcal(icalStrData);
+    if (calName && calName !== calProvider.calName) {
+      this._store.dispatch(
+        IssueProviderActions.updateIssueProvider({
+          issueProvider: { id: calProvider.id, changes: { calName } },
+        }),
+      );
+    }
   }
 
   private _getCalProviderFromCache(): ScheduleCalendarMapEntry[] {
--- a/src/app/features/issue/mapping-helper/get-issue-provider-tooltip.spec.ts
+++ b/src/app/features/issue/mapping-helper/get-issue-provider-tooltip.spec.ts
@@ -1,4 +1,9 @@
-import { sanitizeIcalUrlForDisplay } from './get-issue-provider-tooltip';
+import {
+  getIssueProviderInitials,
+  getIssueProviderTooltip,
+  sanitizeIcalUrlForDisplay,
+} from './get-issue-provider-tooltip';
+import { IssueProvider } from '../issue.model';
 
 describe('sanitizeIcalUrlForDisplay', () => {
   it('returns hostname for https:// URLs', () => {
@@ -83,5 +88,89 @@ describe('sanitizeIcalUrlForDisplay', () => {
 
   it('returns iCal placeholder for undefined input', () => {
     expect(sanitizeIcalUrlForDisplay(undefined)).toBe('iCal');
+  });
+});
+
+describe('getIssueProviderTooltip', () => {
+  it('uses the configured calendar name for CALDAV providers (#9144)', () => {
+    const provider = {
+      issueProviderKey: 'CALDAV',
+      resourceName: 'Work Calendar',
+      caldavUrl: 'https://dav.example.com/calendars/work',
+    } as Partial<IssueProvider> as IssueProvider;
+    expect(getIssueProviderTooltip(provider)).toBe('Work Calendar');
+  });
+
+  it('falls back to the URL for CALDAV providers without a calendar name', () => {
+    const provider = {
+      issueProviderKey: 'CALDAV',
+      resourceName: null,
+      caldavUrl: 'https://dav.example.com/calendars/work',
+    } as Partial<IssueProvider> as IssueProvider;
+    expect(getIssueProviderTooltip(provider)).toBe(
+      'https://dav.example.com/calendars/work',
+    );
+  });
+
+  it('uses the feed calendar name for ICAL providers (#9144)', () => {
+    const provider = {
+      issueProviderKey: 'ICAL',
+      calName: 'Family',
+      icalUrl: 'https://calendar.example.com/family.ics',
+    } as Partial<IssueProvider> as IssueProvider;
+    expect(getIssueProviderTooltip(provider)).toBe('Family');
+  });
+
+  it('falls back to the sanitized hostname for ICAL providers without a name', () => {
+    const provider = {
+      issueProviderKey: 'ICAL',
+      icalUrl: 'https://calendar.example.com/family.ics?token=SECRET',
+    } as Partial<IssueProvider> as IssueProvider;
+    expect(getIssueProviderTooltip(provider)).toBe('calendar.example.com');
+  });
+});
+
+describe('getIssueProviderInitials', () => {
+  it('derives CALDAV initials from the configured calendar name (#9144)', () => {
+    const provider = {
+      issueProviderKey: 'CALDAV',
+      resourceName: 'Work Calendar',
+      caldavUrl: 'https://dav.example.com/calendars/work',
+    } as Partial<IssueProvider> as IssueProvider;
+    expect(getIssueProviderInitials(provider)).toBe('WO');
+  });
+
+  it('falls back to the URL for CALDAV providers without a calendar name', () => {
+    const provider = {
+      issueProviderKey: 'CALDAV',
+      resourceName: null,
+      caldavUrl: 'https://dav.example.com/calendars/work',
+    } as Partial<IssueProvider> as IssueProvider;
+    expect(getIssueProviderInitials(provider)).toBe('DA');
+  });
+
+  it('derives ICAL initials from the feed calendar name (#9144)', () => {
+    const provider = {
+      issueProviderKey: 'ICAL',
+      calName: 'Family',
+      icalUrl: 'https://calendar.google.com/family.ics',
+    } as Partial<IssueProvider> as IssueProvider;
+    expect(getIssueProviderInitials(provider)).toBe('FA');
+  });
+
+  it('keeps the Google shortcut for ICAL providers without a name', () => {
+    const provider = {
+      issueProviderKey: 'ICAL',
+      icalUrl: 'https://calendar.google.com/family.ics',
+    } as Partial<IssueProvider> as IssueProvider;
+    expect(getIssueProviderInitials(provider)).toBe('G');
+  });
+
+  it('falls back to sanitized hostname initials for other ICAL feeds', () => {
+    const provider = {
+      issueProviderKey: 'ICAL',
+      icalUrl: 'https://calendar.example.com/cal.ics',
+    } as Partial<IssueProvider> as IssueProvider;
+    expect(getIssueProviderInitials(provider)).toBe('CA');
   });
 });
--- a/src/app/features/issue/mapping-helper/get-issue-provider-tooltip.ts
+++ b/src/app/features/issue/mapping-helper/get-issue-provider-tooltip.ts
@@ -56,9 +56,13 @@ export const getIssueProviderTooltip = (issueProvider: IssueProvider): string =>
       case 'GITLAB':
         return issueProvider.project;
       case 'CALDAV':
-        return issueProvider.caldavUrl;
+        // Prefer the human-readable calendar name the user is required to
+        // configure; the URL is only the fallback (#9144).
+        return issueProvider.resourceName || issueProvider.caldavUrl;
       case 'ICAL':
-        return sanitizeIcalUrlForDisplay(issueProvider.icalUrl);
+        // Prefer the calendar's own display name (X-WR-CALNAME, captured on
+        // fetch) over the bare hostname (#9144).
+        return issueProvider.calName || sanitizeIcalUrlForDisplay(issueProvider.icalUrl);
       case 'REDMINE':
         return issueProvider.projectId;
       case 'OPEN_PROJECT':
@@ -115,12 +119,18 @@ export const getIssueProviderInitials = (
         ?.substring(0, 2)
         ?.toUpperCase();
     case 'CALDAV':
+      if (issueProvider.resourceName) {
+        return issueProvider.resourceName.substring(0, 2).toUpperCase();
+      }
       return issueProvider.caldavUrl
         ?.replace('https://', '')
         ?.replace('http://', '')
         ?.substring(0, 2)
         ?.toUpperCase();
     case 'ICAL':
+      if (issueProvider.calName) {
+        return issueProvider.calName.substring(0, 2).toUpperCase();
+      }
       if (issueProvider.icalUrl?.includes('google')) return 'G';
       if (issueProvider.icalUrl?.includes('office365')) return 'MS';
       // Route through the sanitizer so credentials embedded in user/path/query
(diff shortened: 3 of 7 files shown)


johannesjo:
Thanks for this, and for the unusually careful spec work on the parser. I labelled #9144 `help wanted` myself, so this is welcome even though the issue is quiet.

Two things up front, because they shape everything below.

**The CalDAV half is independent and I would merge it today.** `resourceName` is `required: true` in `CALDAV_CONFIG_FORM` (`caldav-cfg-form.const.ts:61`, `:64`) and nothing in `src/` ever writes it programmatically, so it is always a value the user chose, and your URL fallback is preserved. One caveat rather than a problem: it is not guaranteed to be a *display* name, because `_findMatchingCalendar` matches it against the calendar's display name and then falls back to the final URL path segment (`caldav-client.service.ts:110-128`; the tested Vikunja case sets `resourceName: '12'` at `caldav-client.service.spec.ts:395`). So some users will see a slug in the chip. Still better than the hostname it replaces, and the tooltip change is a net privacy win since it drops the un-sanitised `caldavUrl` that the iCal branch right below it deliberately hostname-sanitises.

**The iCal half is not scope creep, in case anyone suggests shipping only the one-liner.** The Schedule "Calendar visibility" menu is fed by `selectCalendarProviders`, which filters `issueProviderKey === ICAL_TYPE` (`issue-provider.selectors.ts:99-105`), and it renders `{{ calProviderLabel(provider) }}` → `getIssueProviderTooltip` (`schedule.component.ts:98`, `schedule.component.html:68`). That menu lists iCal providers only, so the reporter's fourth screenshot is fixed by the iCal half specifically.

`getCalNameFromIcal` itself is good. I bundled it with esbuild and ran it directly against real inputs; unfolding, parameters, `\,` / `\;` / `\\` unescaping and LF-only feeds all behave. Everything below is about what happens to the value afterwards.

---

## 1. Nothing ever clears `calName`, so it goes stale

The reducer is `adapter.updateOne` (`issue-provider.reducer.ts:212`), which merges a partial and never deletes, so a config edit that omits `calName` leaves the old value in place. The edit dialog additionally carries it forward explicitly, since `submit()` dispatches the whole model rather than a diff (`dialog-edit-issue-provider.component.ts:225`).

`icalUrl` is an editable required field on the existing-provider form, so repointing a provider at a different feed is a first-class flow. Do that and:

- new feed has `X-WR-CALNAME` → wrong label until the next successful fetch
- new feed has none → the guard `if (calName && calName !== calProvider.calName)` (`calendar-integration.service.ts:526-527`) short-circuits on `undefined`, and the label shows the **previous** calendar's name over the new URL, permanently

Same shape for a user who removes `X-WR-CALNAME` from a feed they control: the old name can never clear. And `duplicate()` deep-clones the model (`:247`), so a clone repointed at another feed starts life mislabelled.

This is the one finding I would call a straightforward bug: a confidently wrong name is worse than the hostname it replaced, and it defeats the disambiguation the feature exists for.

## 2. Learning a name refetches every calendar, N+1 times over

I measured this rather than reasoning about it, using the real service, the real `issueProvidersFeature` reducer and a real HTTP backend.

`adapter.updateOne` mints a new entity object, so `selectCalendarProviders` emits an array whose changed element fails `fastArrayCompare` (length plus per-element reference identity) and the emission passes `distinctUntilChanged`. Because the dispatch happens synchronously inside the `switchMap` projection (`calendar-integration.service.ts:446-447`), the ancestor tears down **during** the projection, before it returns its observable, aborting the sibling requests in the same `forkJoin` at the network layer (`FetchBackend` teardown calls `aborter.abort()`). The response that triggered the dispatch is discarded too, so a torn-down round contributes nothing, and `saveToRealLs(LS.CAL_EVENTS_CACHE, …)` only runs after `forkJoin` completes, so it shows the user nothing either.

Measured totals on first run with N calendars that all carry `X-WR-CALNAME`, restart-driven only (no poll ticks): **N+1 rounds of N requests, so N·(N+1) requests instead of N.** N=1 → 2, N=2 → 6, N=3 → 12, N=5 → 30, with per-round cancellations for N=5 of 4, 3, 2, 1, 0, 0. `calendarEvents$` and `pollChanges$` are independent subscriptions and each pay it separately: `pollChanges$` is registered at `feature-stores.module.ts:198` and `calendarEvents$` is held open for the app's lifetime by `ScheduleService`, so this happens even if Schedule and Work View are never opened.

Honest caveat on the exponent: that was measured in `fakeAsync` flushing responses one at a time. The step to real browsers rests on one unmeasured argument, that each response arrives in its own task and the teardown is synchronous within it, so at most one name is learned per round. The claim with nothing inferred is "at least one extra full refetch round per calendar whose name is newly learned", which is directly measured.

Two related observations from the same probes:

- A **redundant** dispatch with an identical `calName` also mints a new object and also passes `fastArrayCompare`, because `takeUpdatedModel` does an unconditional `Object.assign`. So the `calName !== calProvider.calName` guard is not a churn optimisation, it is the only thing standing between this and an unterminated loop on that path. `pollChanges$`'s deep `JSON.stringify` compare does absorb the no-op case.
- The inner `timer(0, getEffectiveCheckInterval(calProvider))` in `pollChanges$` re-fires against the `calProvider` captured by the outer `switchMap`, never a re-read one. Today that is harmless *only* because the outer comparator is deep, so the write restarts the chain and re-captures a fresh provider (verified: N dispatches, then zero across three further poll intervals). Narrowing that comparator to ids, an obvious optimisation given the churn above, silently turns this into one persisted op per calendar per poll interval (verified on a verbatim replica: 2, 4, 6, 8, 10 ops over four intervals with two calendars). A write inside a `switchMap` projection whose termination depends on an unrelated ancestor's comparator is a sharp edge worth avoiding on its own.

## 3. Cheap fixes

**Cap the length.** `getCalNameFromIcal` returns a 500-character `X-WR-CALNAME` intact, and it is persisted and synced verbatim. The display side is already fine, I rendered the real CSS stack and a 500-char label clips cleanly in a single 48px row (`_overwrite-material.scss:299` plus Material's 280px panel), and tooltips are bounded at 200px. So this is purely about what we persist and replicate from a remote feed. A `.slice(0, 100)` in the extractor closes it.

**Add the privacy-export spec.** `privacy-export.spec.ts` has an established convention for exactly this: a `describe('new sensitive fields (Issue #6020)')` block with one `it` per key plus a table-driven `cases` array. `'calName'` was added to `KEY_TO_REPLACE` (`privacy-export.ts:38`) without one. Masking is the right call, a feed name like "Chemo appointments" is at least as identifying as `resourceName`, and I checked that the bare key collides with nothing else in `src/` or `packages/`.

**One spec gap worth closing while you are in there:** the parser is matched with `/m` and no VCALENDAR scoping, so `BEGIN:VEVENT\r\nX-WR-CALNAME:Injected\r\nEND:VEVENT` returns `"Injected"`. Harmless in practice, but you already have the `SUMMARY:X-WR-CALNAME:` decoy test right there, and this is its natural sibling. The `\\` escape branch is also unexercised (only `\,`, `\;`, `\n` are).

## 4. No test covers the new dispatch

The parser and tooltip specs are solid, but they cover the two pure functions. I proved the gap rather than asserting it: replacing the body of `_updateCalNameIfChanged` with a no-op leaves `calendar-integration.service.spec.ts` at `Executed 76 of 76 SUCCESS`, identical to before and after restoring it. Nothing else in the tree would catch it either; the only specs mentioning `calName` or `getCalNameFromIcal` are the two new pure-function ones.

Two structural reasons that file cannot catch it as written, both worth knowing before you add tests: every `MOCK_ICAL_*` fixture lacks `X-WR-CALNAME`, so the guard is never entered, and `provideMockStore`'s `dispatch` runs no reducer, so even a dispatch has no observable consequence. The two cases that matter are "dispatches once for a feed with a new name" and "does not dispatch when the stored name already matches", and the file already has `MockStore` and `HttpTestingController` wired up.

## 5. The design question, which is mine to answer and not yours

Everything above is actionable. This part is not a change request: whether a calendar display name belongs in **synced** state is my call, and I am writing the tradeoff out so it is visible. If you would rather not touch it, say so and I will decide.

`updateIssueProvider` carries `meta: { isPersistent: true, entityType: 'ISSUE_PROVIDER', opType: OpType.Update }` (`issue-provider.actions.ts:34-44`), so each of these becomes a durable op replicated to the user's other devices. This is the first *automatic* writer to issue-provider entities; the six existing ones are all click-driven (edit dialog, pin/unpin, drag-and-drop reorder, track-time dialog, project share). It fires from four entry points, none of them gated: `pollChanges$` (`calendar-integration.effects.ts:99`), `_fetchAllCombined` (`calendar-integration.service.ts:183`), and `CalendarCommonInterfacesService` at `:80` (`searchIssues`, i.e. typing in the issue-panel search box can write to synced state) and `:149`. `testConnection` is clear, it has its own bare `_http.get` and does not route through `requestEvents$`.

Two consequences:

**Sync-window coverage is narrower than the neighbouring code assumes.** The capture meta-reducer does buffer persistent actions while remote ops are being applied (`operation-capture.meta-reducer.ts:250-257`), so the apply window itself is safe. It does not cover the other two components of `isInSyncWindow` (`hydration-state.service.ts:87-92`): the window `SyncTriggerService` opens before the trigger debounce, and the post-sync cooldown. That is the gap `skipDuringSyncWindow()` exists to close, and the gap `pollChanges$`'s auto-import branch closes by hand at `:104-107` with a comment explaining why writes from this exact path must not race the first sync (discussion #7677). The new dispatch sits upstream of that gate, inside `requestEvents$`.

**ISSUE_PROVIDER conflicts resolve by whole-entity LWW.** `_tryCreateDisjointMergeOp` returns `undefined` for any type without a `RECREATE_FALLBACK` entry (`conflict-resolution.service.ts:2430`), and `recreate-fallback.const.ts:56` names ISSUE_PROVIDER as one of the types that "still fall through to the legacy behavior". So if a `calName` write races a user's offline provider edit on another device, it is asymmetric and neither direction is good:

- `calName` wins → every unsynced local op on that provider is rejected, not just the conflicting one (`conflict-resolution.service.ts:1439-1441`, "Reject ALL pending ops for entities where remote won"). The user's edit stays in local state, since the winner is a partial `changes: {calName}` applied via `updateOne`, but no op carries it any more, so it never reaches the other devices and is lost on the next replay from the log.
- the edit wins → its local-win op is a full-entity `'replace'` snapshot applied via `setOne`, which clears `calName` on every client, and the next poll re-dispatches it.

Neither is surfaced to the user: `findLwwContentConflicts` skips every `entityType !== 'TASK'` by design. Separately, N devices each discover an upstream rename independently, so one rename produces N concurrent ops rather than one. They converge on the same value, so no loss there.

**What I would prefer:** `calName` is a cache of remote data, not user intent, and every device already fetches the feed and can derive it for free. Keeping it out of synced state removes the ops, the conflict exposure, the model field, the privacy-export entry, and issue 1 above, all at once. I should correct myself on the obvious home for it though: `LS.CAL_EVENTS_CACHE` is *not* a drop-in, because `ScheduleCalendarMapEntry` is `{ items }` with no provider key (`schedule.model.ts:133-135`), it is written in exactly one place positionally aligned to the provider array, and only on the schedule path, so the poll path never populates it. It would need a new keyed cache, and the three display call sites (`issue-panel.component.ts:86-87`, `issue-provider-tab.component.ts:118`, `schedule.component.ts:98`) would need the name passed in rather than read off the provider. That is real work, which is exactly why it is my decision and not something to spring on you.

If we keep the synced field instead, the minimum is: clear `calName` when `icalUrl` changes, gate the dispatch on `isInitialSyncDoneSync() && !isInSyncWindow()` the way the auto-import branch does, and cap the length.

---

## Housekeeping

**Correcting your master-failure note, and thank you for flagging it.** The failure was real, but the cause was not the hard-coded date: `DEFAULT_TASK_REPEAT_CFG` freezes the real current day into `lastTaskCreation` at module load, so that fixture stopped producing a projection once the wall clock passed its hard-coded 2026-07-30. I fixed it on master this morning in `ba1e6e5292`, keeping the hard-coded dates deliberately. You are 9 commits behind, so a rebase clears it.

All 6 required workflows are sitting at `action_required` because this is your first PR here, so nothing has actually run yet. I will approve them.

Checks I ran locally, all on your head commit: `checkFile` clean on all 7 changed files, and the new plus existing specs green (11 + 24 + 76). Repo rules I checked so nobody has to ask: `calName?: string` is correctly optional, so no existing install breaks; `frozen-state` needs no fixture change; and no `CURRENT_SCHEMA_VERSION` bump is needed or was made, because validation uses `createValidate` rather than `createValidateEquals`, so an op carrying `calName` merges cleanly on a client that predates the field. No new user-facing strings, so nothing owed in `en.json`. I re-derived every expected value in your new specs by hand and none of them are vacuous.

One last nit, and it is genuinely pre-existing rather than something you introduced: `getIssueProviderInitials` does `.substring(0, 2)`, which splits a surrogate pair, so `A🏠Home` renders as `A` plus a replacement glyph. The whole function already does this on six other branches and you followed the convention exactly, so it is fine to leave. I mention it only because `calName` is the branch most likely to hold an emoji. A leading emoji is safe (exactly 2 UTF-16 units), though a leading flag gets cut to one regional indicator.

johannesjo:
Correcting one line of my own housekeeping note above: **do not rebase yet.** Nothing here is caused by your PR, and there is nothing for you to do.

I said the spec you flagged is fixed on master so a rebase would clear it. The first half is true, the advice is not. `origin/master` at `0840f9c441` currently cannot **compile** the unit suite at all, in that same file:

```
create-blocked-blocks-by-day-map.spec.ts:77:7
```

`lastTaskCreation` and `lastTaskCreationDay` are each set twice in one object literal, at `:66-67` and again at `:77-78`. That is a hard build error rather than a failing assertion, so the whole Karma bundle dies and **zero** unit tests run.

It is my fault twice over. My fix `ba1e6e5292` set both fields to `undefined`; the squash merge of #9390 then added a second, pinned pair to the same literal. Duplicate object keys produce no merge conflict, since both hunks are additions at different offsets, so the two independent fixes for the same date-dependent fixture both survived and only surfaced at build time.

I have the repair ready and will push it shortly. Rebase after that and you get a green baseline. Your CI approvals are coming either way, but until the repair lands any unit-test result on any PR is meaningless, so it is worth waiting rather than chasing a red run.

johannesjo:
**🔧 Needs changes** – labelled `needs work`; ping me once this and the items in my earlier review are addressed and I'll re-review.

This is a follow-up pass rather than a second review. Nothing has been pushed since 30 July, so instead of re-deriving my earlier review I re-checked it against current `master`: every reference in it still resolves, and the fix is still needed (`get-issue-provider-tooltip.ts` on master still shows `caldavUrl` for CalDAV and the bare hostname for iCal). Those items stand as written and I am not repeating them here. One new thing came out of re-reading the parser, plus housekeeping that reverses advice I gave you.

### Should fix before merge

- **`src/app/features/schedule/ical/get-cal-name-from-ical.ts:16`** – the parameter sub-pattern `(?:;[^:\r\n]*)?` stops at the first `:`, but RFC 5545 §3.2 allows a DQUOTE-quoted parameter value to contain one. So `X-WR-CALNAME;X-FOO="a:b":Real Name` extracts `b":Real Name`, and that string is what gets persisted and rendered as the calendar's label. I confirmed it by running your match regex directly rather than reasoning about it. Allowing a quoted run in the parameter part fixes it:

  ```
  /^X-WR-CALNAME(?:;(?:"[^"\r\n]*"|[^:"\r\n])*)?:(.*)$/im
  ```

  I checked that against plain values, single and multiple unquoted parameters, a colon inside the value, and your `SUMMARY:` decoy: all unchanged, and the quoted case now returns `Real Name`.

  While you are in that function: scoping the match to the text before the first `BEGIN:V…` component would close the `BEGIN:VEVENT` gap I mentioned last time and, as a side effect, drop the full-feed `.replace()` copy plus end-to-end regex scan that currently runs on every poll of every calendar. The pattern has no leading anchor, so a feed *without* `X-WR-CALNAME` – the common case – is scanned all the way to the end today.

### Housekeeping

- **The "do not rebase yet" hold is lifted.** The duplicate-key compile error in `create-blocked-blocks-by-day-map.spec.ts` that was killing the whole unit bundle is fixed on master (`dba2a0e18c`). A rebase now gets you a green baseline.
- **The branch is 339 commits behind `master`** (it was 9 back then), but it still merges cleanly, and the region of `calendar-integration.service.ts` your patch touches is unchanged on master. So the rebase is about the test baseline, not about conflicts.
- **CI still has no signal here.** Only the `welcome` job has run; the required workflows are still unapproved, so nothing in this PR has actually been exercised by CI.

### Questions

- The thing really gating the rest is the design question in section 5 of my earlier review: whether `calName` belongs in synced `ISSUE_PROVIDER` state at all, or in a device-local cache. I asked whether you would rather not touch it, and either answer unblocks you – the two paths need different work, and three of my four other items only exist on the "keep the synced field" path. A one-liner is enough, and it is worth settling before you spend time on that half.
- Are you still able to carry this? No pressure either way; if you would rather not, just say so.

---

johannesjo:
Settling my own section-5 question, since it's blocking the rest:

**Make `calName` device-local, not a synced `ISSUE_PROVIDER` field.**

What decided it: conflict detection is purely clock-based, so two devices that independently learn the *same* calendar name still compare CONCURRENT. And an `ISSUE_PROVIDER` conflict does surface — `maybeShowSummaryBanner()` has no entity-type filter, so it opens the auto-resolved banner with a REVIEW link. That means every multi-device user gets a guaranteed conflict banner on the first poll after upgrading, for a display string each device can re-derive from the feed on its own. Not worth any sync surface.

Going device-local also dissolves three of the four remaining items: no clearing on `icalUrl` change, no sync-window gate, no length cap on synced state — just a keyed cache plus passing the name to the three display call sites.

**Splitting the CalDAV half off.** It's 3 production lines (`resourceName || caldavUrl` plus the initials guard), touches no model field and no synced state, and is independent of the above. I'll take that half now so it stops waiting on the design question.

No pressure to carry the iCal half if you're out of time — say the word and I'll pick it up on top of your commits.

PR #9386, opened 2026-07-30
test(schedule): stop the repeat projection spec depending on the wall clock

KamilDev:
`master` is currently red, and it is a time bomb rather than a flake.

`createBlockedBlocksByDayMap() > keeps a repeat projection source day when a scheduled occurrence crosses midnight` fails on a clean checkout:

```
TypeError: Cannot read properties of undefined (reading 'sourceOccurrenceDate')
    at create-blocked-blocks-by-day-map.spec.ts:85:27
```

## Why it went red on its own

The spec spreads `DEFAULT_TASK_REPEAT_CFG` and overrides `startDate`, but leaves `lastTaskCreationDay` at the default. That default is evaluated at module load:

```ts
// task-repeat-cfg.model.ts:108-110
export const DEFAULT_TASK_REPEAT_CFG: Omit<TaskRepeatCfgCopy, 'id'> = {
  lastTaskCreation: Date.now(),
  lastTaskCreationDay: getDbDateStr(),   // <- the real today
```

and `selectTaskRepeatCfgsForExactDay` drops any day equal to that anchor:

```ts
// task-repeat-cfg.selectors.ts:106-113
const effectiveLastDay = getEffectiveLastTaskCreationDay(taskRepeatCfg);
if (effectiveLastDay === dateStr || (effectiveLastDay && effectiveLastDay > dateStr)) {
  return false;
}
```

The spec's occurrence day is hardcoded to `2026-07-30`. So it was green when it landed on 2026-07-29 (`a12a7a1e4`, #9314) and started failing the next day, when the wall clock reached the date it asserts on — no code change involved.

Verified red at `a12a7a1e4`, `46f1bb53e~1`, `46f1bb53e`, `d96109681` and `7b6a17047`, i.e. from the commit that introduced it onward; and shifting only the spec's two dates to 2028 turns it green, which isolates the cause to the anchor rather than anything in the projection code.

## Fix

Pin `lastTaskCreationDay` explicitly, which is what the sibling specs in the same area already do (`create-sorted-blocker-blocks.spec.ts:777` uses `'1970-01-01'`, `month-view-repeat-projection-day.spec.ts:19` uses an epoch-relative date). Test-only change; no production code touched.

Full `npm test` green in both timezones on this branch: 13705 passing / 16 skipped.

## One thing worth flagging beyond this diff

This failure mode is not specific to the one spec — any spec that spreads `DEFAULT_TASK_REPEAT_CFG` while asserting on a hardcoded date becomes a bomb dated to that day, and it lands on whoever pushes next rather than on the PR that introduced it. I did not audit the rest of the suite for it, so I can only say it is green today, not that no other spec detonates later. If you want, the sharper fix is to make the anchor non-defaulted for tests, or to have `DEFAULT_TASK_REPEAT_CFG` not bake a load-time timestamp at all — happy to look at either as a follow-up, but that is a wider change than unbreaking the suite.

### DIFF
--- a/src/app/features/schedule/map-schedule-data/create-blocked-blocks-by-day-map.spec.ts
+++ b/src/app/features/schedule/map-schedule-data/create-blocked-blocks-by-day-map.spec.ts
@@ -62,6 +62,11 @@ describe('createBlockedBlocksByDayMap()', () => {
       startDate: occurrenceDay,
       startTime: '23:00',
       defaultEstimate: h(2),
+      // DEFAULT_TASK_REPEAT_CFG anchors this to the real today at module load,
+      // and selectTaskRepeatCfgsForExactDay drops a day that equals the anchor —
+      // so leaving it defaulted made this spec pass only until the wall clock
+      // reached the hardcoded occurrence day.
+      lastTaskCreationDay: '1970-01-01',
     };
 
     const result = createBlockedBlocksByDayMap(


johannesjo:
Your diagnosis is right, and you had it first. This PR was opened at 00:17 UTC. I hit the same failure independently and pushed `ba1e6e5292` at 09:31, and #9390 landed a third variant of the same repair at 09:54. That is on me for not looking here first.

Both of your citations check out verbatim: `create-sorted-blocker-blocks.spec.ts:777` pins `lastTaskCreationDay: '1970-01-01'`, and `month-view-repeat-projection-day.spec.ts:19` uses `getDbDateStr(new Date(1969, 11, 31).getTime())`. I also installed your exact blob and ran the spec: 5/5 green, so your fix works standalone. Pinning only `lastTaskCreationDay` is sufficient, because `getEffectiveLastTaskCreationDay()` returns it before it ever reads `lastTaskCreation`, and nothing in the `createBlockedBlocksByDayMap` path reads the numeric field without going through that helper.

Closing as already fixed. What happened after you filed is the more useful part of this.

## The two landed fixes collided and broke the build

They edited the same object literal at different anchor points, one before `quickSetting` and one after `defaultEstimate`, so git auto-merged both insertions with no conflict (`git merge-tree ba1e6e5292 a7a5de9a79` reports `Auto-merging`, no CONFLICT line). Master then declared both keys twice:

```
77       lastTaskCreation: new Date(2026, 6, 29).getTime(),
78       lastTaskCreationDay: '2026-07-29',
```

A compile error in one spec fails the whole Karma bundle (`Error: Found 1 load error`), and the browser never spawns, so the suite ran zero specs rather than one failing one. Repaired in `dba2a0e18c`, keeping the pair that clears both fields.

**The gap was not missing CI, it was a stale merge base.** Master does run the client unit suite on every push, as the `Test Unit` step of `linux-bin-and-snap-release` in `build.yml`, and it did its job: red on `32c1610c74` and `0840f9c441`, green on `ba1e6e5292` either side. What went wrong is that #9390's `Tests` run started at 09:23:19Z, `ba1e6e5292` was not pushed until 09:37, and the PR merged at 09:54:43Z, 39 seconds after that now-stale check reported green. The duplicated tree existed in no tested commit. The remedy is requiring branches to be up to date before merge, not more test coverage.

Worth noting the blast radius, since `Test Unit` is step 8 of that job: steps 9 through 18 were skipped on both commits, which includes `Build Frontend for E2E`, the E2E run, the Electron build and the snap `edge` upload. A red unit spec silently stopped publishing.

Lint would not have caught the duplicate either, and the obvious fix for that does not work. `no-dupe-keys` is not merely absent, it is explicitly disabled by `tseslint.configs.recommended`, which the repo does extend:

```
node_modules/@typescript-eslint/eslint-plugin/dist/configs/eslint-recommended-raw.js:21
  'no-dupe-keys': 'off', // ts(1117)
```

`npx eslint --print-config` on the spec confirms `no-dupe-keys = [0]`. So adding `js.configs.recommended` is inert; it has to be an explicit `'no-dupe-keys': 'error'` after the extends. Reasonable of typescript-eslint to defer to ts(1117), except ts(1117) only surfaces at `ng test` here, because `tsconfig.app.json` excludes `**/*.spec.ts`. A `tsc -p src/tsconfig.spec.json --noEmit` gate would be the structural fix.

## Your open question, answered

You flagged that you had not audited the rest of the suite, so I did. Pinning `DEFAULT_TASK_REPEAT_CFG`'s creation anchor to a date past every hardcoded date in the suite (`2030-01-01`, then `2100-01-01`) leaves the full suite green, and with your spec reverted to its pre-fix state the same probe fails exactly one test, so it is sensitive rather than inert.

Being precise about what that does and does not prove: it is not a simulation of the clock advancing, it is a deliberate over-approximation of one mechanism. Because the exact-day selector drops every day at or before the anchor, green at 2100 implies green at any real date up to 2100 for specs that inherit that constant. It says nothing about specs that reach the real clock by another route. Within that scope, this spec was the only one. Of the 20 spec files that spread the constant, 11 leave the anchor unpinned, but none of the other 10 depend on a projection for a hardcoded day.

## On the sharper fix you offered

I would leave `DEFAULT_TASK_REPEAT_CFG` as it is. Both production spread sites (`dialog-edit-task-repeat-cfg.component.ts:333`, `add-task-bar.component.ts:620`) route through `addTaskRepeatCfgToTask`, and `updateTaskAfterMakingItRepeatable$` overwrites both fields from the computed first occurrence, so the load-time stamp does not survive into settled data. The hazard is test-only, and changing a production model to remove a test hazard is the wrong direction.

One honest note on your variant, since it is the shape most likely to be copied next. Pinning only the day field leaves the object in a state production cannot produce: `lastTaskCreationDay` says 1970 while `lastTaskCreation` still holds today's `Date.now()` from the spread. It is green only because the helper short-circuits on the day field, so any change to that precedence, or any new reader of the numeric field in this chain, quietly re-arms the dependency you set out to remove. The sibling you patterned it on has the same contradiction, so consistency there propagates the trap. Clearing both, or pinning both, avoids it.

Thanks for the report and the write-up. The diagnosis was correct, the reproduction was properly isolated, and flagging the systemic version of it was the most valuable part.

johannesjo:
Closing: the fixture is already repaired on master (`ba1e6e5292` plus `dba2a0e18c`), and this branch now conflicts with it. Credit for finding and diagnosing it first stands.

PR #9387, opened 2026-07-30
fix(short-syntax): keep the typed time when the weekend skip crosses a DST change

KamilDev:
Closes the one finding in your approving review on #9370 — thanks for catching it. It landed as-is, so this is live in `master`.

> `skipExcludedWeekend` steps *through* Sunday, and Sunday is the day DST changes. If the typed time falls in the spring-forward gap, `setDate` normalizes it and the shift sticks.

Confirmed, with your numbers:

```
Europe/Berlin, spring-forward Sun 2026-03-29 (02:00 → 03:00)
"Standup @every weekday 2:30am" typed Fri 2026-03-27 08:00

master: preview Sat 03-28 02:30 → cfg.startDate 03-28, startTime "02:30" → SCHED Mon 03-30 02:30
  this: preview Mon 03-30 02:30 → cfg.startDate 03-30, startTime "02:30" → SCHED Mon 03-30 02:30
```

Sat 02:30 → `setDate(+1)` → Sun 02:30 does not exist → JS gives 03:30 → `setDate(+1)` → Mon 03:30. And as you noted this one is not preview-only: `state.time` is read back off the parsed timestamp (`add-task-bar-parser.service.ts:191-203`), so the shifted hour is written into the repeat config's `startTime` and the task then recurs at 03:30 every workday.

The fix is your patch — capture the clock time before the loop, restore it after.

## The spec is derived, not hardcoded

One deviation worth explaining. The convention in the repo for DST specs is to hardcode one zone's transition and note that it passes trivially elsewhere (`date.service.tz.spec.ts:105-108`). I did not follow it here, because a spec that cannot fail in `Europe/Berlin` would not have guarded this — and Berlin is where the suite runs by default.

So `findSpringForwardSunday(2026)` derives the transition from the ambient zone: it looks for a Sunday shorter than 24h, then for the hour where `getHours()` disagrees with the hour requested. Zones without a transition return `null` and the spec returns early.

Verified by breaking it — with the fix reverted, the spec fails in **both** timezones the suite runs in, not just one:

```
TZ=Europe/Berlin       Expected 3 to be 2.   TOTAL: 1 FAILED, 222 SUCCESS
TZ=America/Los_Angeles Expected 3 to be 2.   TOTAL: 1 FAILED, 222 SUCCESS
```

With the fix: 223 SUCCESS in `Europe/Berlin`, `America/Los_Angeles`, `Asia/Tokyo` and `Australia/Sydney`.

Full `npm test` green in both timezones: 13705 passing / 16 skipped.

## Note on the base

The first commit here is #9386, not part of this change. `master`'s suite is currently red on an unrelated spec that became date-dependent yesterday, so the pre-push hook cannot pass on a branch based directly on `master`. Once #9386 lands this rebases down to the single `short-syntax` commit — or just merge #9386 first and I will rebase.

## Your other two review items — deliberately not in here

- The date-picker + *Monday to Friday* pick divergence (`applyUserRepeatPick` not touching `state.date`) you marked "scope note, not a request", so I have left it out. Happy to do it separately if you want it.
- `@every weekday 2:30am` typed on **Sat** 2026-03-28 parsing to no date at all is pre-existing and untouched here.

### DIFF
--- a/src/app/features/schedule/map-schedule-data/create-blocked-blocks-by-day-map.spec.ts
+++ b/src/app/features/schedule/map-schedule-data/create-blocked-blocks-by-day-map.spec.ts
@@ -62,6 +62,11 @@ describe('createBlockedBlocksByDayMap()', () => {
       startDate: occurrenceDay,
       startTime: '23:00',
       defaultEstimate: h(2),
+      // DEFAULT_TASK_REPEAT_CFG anchors this to the real today at module load,
+      // and selectTaskRepeatCfgsForExactDay drops a day that equals the anchor —
+      // so leaving it defaulted made this spec pass only until the wall clock
+      // reached the hardcoded occurrence day.
+      lastTaskCreationDay: '1970-01-01',
     };
 
     const result = createBlockedBlocksByDayMap(
--- a/src/app/features/tasks/short-syntax.spec.ts
+++ b/src/app/features/tasks/short-syntax.spec.ts
@@ -2375,6 +2375,37 @@ describe('shortSyntax recurrence', () => {
   ): RepeatQuickSetting | null =>
     r?.repeat?.type === 'PRESET' ? r.repeat.quickSetting : null;
 
+  // The ambient timezone's spring-forward sunday and the hour it skips, derived
+  // rather than hardcoded so the DST case below is exercised in every zone the
+  // suite runs in (Berlin transitions in march, Los Angeles two weeks earlier,
+  // Sydney in october). Returns null where there is no transition.
+  const findSpringForwardSunday = (
+    year: number,
+  ): { sunday: Date; missingHour: number } | null => {
+    const dayMs = 24 * 60 * 60 * 1000;
+    for (let month = 0; month < 12; month++) {
+      for (let day = 1; day <= 31; day++) {
+        const start = new Date(year, month, day, 0, 0, 0, 0);
+        if (start.getDate() !== day || start.getDay() !== 0) {
+          continue;
+        }
+        // A spring-forward day is shorter than 24h.
+        if (
+          new Date(year, month, day + 1, 0, 0, 0, 0).getTime() - start.getTime() >=
+          dayMs
+        ) {
+          continue;
+        }
+        for (let hour = 1; hour < 6; hour++) {
+          if (new Date(year, month, day, hour, 30, 0, 0).getHours() !== hour) {
+            return { sunday: start, missingHour: hour };
+          }
+        }
+      }
+    }
+    return null;
+  };
+
   it('should parse "@every friday" as weekly repeat anchored to next friday', async () => {
     const r = await parse('Water plants @every friday');
     expect(presetOf(r)).toBe('WEEKLY_CURRENT_WEEKDAY');
@@ -2592,6 +2623,28 @@ describe('shortSyntax recurrence', () => {
     expect(due.getDate()).toBe(22);
     expect(due.getDay()).toBe(1);
     expect(due.getHours()).toBe(6);
+  });
+
+  it('should keep the typed time when the skipped-over sunday springs forward', async () => {
+    // The weekend skip steps through Sunday, which is the day DST changes. If
+    // the typed time falls in the spring-forward gap, setDate normalizes it and
+    // the shift sticks — and this one is not preview-only: the time is read back
+    // off the parsed timestamp into the repeat config's startTime, so the task
+    // would recur an hour late every workday.
+    const gap = findSpringForwardSunday(2026);
+    if (!gap) {
+      // Timezone without a spring-forward transition (UTC, Tokyo).
+      return;
+    }
+    const friday = new Date(gap.sunday);
+    friday.setDate(friday.getDate() - 2);
+    friday.setHours(gap.missingHour + 6, 0);
+    const r = await parse(`Standup @every weekday ${gap.missingHour}:30am`, friday);
+    expect(presetOf(r)).toBe('MONDAY_TO_FRIDAY');
+    const due = new Date(r?.taskChanges.dueWithTime as number);
+    expect(due.getDay()).toBe(1);
+    expect(due.getHours()).toBe(gap.missingHour);
+    expect(due.getMinutes()).toBe(30);
   });
 
   it('should parse "@every 15th" as monthly on next 15th', async () => {
--- a/src/app/features/tasks/short-syntax.ts
+++ b/src/app/features/tasks/short-syntax.ts
@@ -852,9 +852,15 @@ const anchorCycleOf = (repeat: ShortSyntaxRepeat): RepeatCycleOption | null => {
 // date in place would only make the add bar advertise a first occurrence the
 // task never gets. Mutates in place, like the roll-forward above.
 const skipExcludedWeekend = (date: Date): void => {
+  const hours = date.getHours();
+  const minutes = date.getMinutes();
   while (date.getDay() === 0 || date.getDay() === 6) {
     date.setDate(date.getDate() + 1);
   }
+  // A skipped-over day can be a DST spring-forward day whose 02:00-03:00 hour
+  // does not exist; stepping through it would silently shift the clock time,
+  // and that time is read back into the repeat config's startTime.
+  date.setHours(hours, minutes);
 };
 
 // Resolves a matched recurrence phrase into task changes: the parsed repeat


johannesjo (approved):
The `short-syntax.ts` change is right and I could not break it. One mechanical blocker before it can land, and it is not your fault.

## Blocker: drop the first commit, and note master is currently red

`5d5bdb5eb` has to go. Two independent fixes for that same fixture already landed on master while this PR was open (`ba1e6e529`, plus the #9390 squash), so yours is a third one and it conflicts:

```
$ git merge-tree --write-tree origin/master pr-9387
CONFLICT (content): Merge conflict in
  src/app/features/schedule/map-schedule-data/create-blocked-blocks-by-day-map.spec.ts
```

The DST commit on its own is clean, so a rebase that keeps only `712a5ef8c` merges without conflict. The same thing makes #9386 moot.

Worth knowing, because it explains why your green `Tests` check does not reflect master: those two fixes collided into a duplicate-key object literal, so **master does not compile at all right now**:

```
create-blocked-blocks-by-day-map.spec.ts:77:7 - error TS1117:
  An object literal cannot have multiple properties with the same name.
ERROR [karma-server]: Error: Found 1 load error
```

Karma never starts, so the whole unit suite is down on `32c1610c7`, confirmed in the failing `Build All & Release` run. That is mine to clean up, not yours, and I have the fix ready. Your `Tests` check is green because this branch is based on a commit that predates the collision. So please rebase onto master once that landed rather than onto the current tip.

## The fix: verified, no reservations

Sabotage first. Removing `date.setHours(hours, minutes)` fails exactly one spec and nothing else:

```
TZ=Europe/Berlin        1 FAILED, 222 SUCCESS   Expected 3 to be 2
TZ=America/Los_Angeles  1 FAILED, 222 SUCCESS   Expected 3 to be 2
TZ=Australia/Sydney     1 FAILED, 222 SUCCESS   Expected 3 to be 2
TZ=Asia/Tokyo             223 SUCCESS           (no transition, skipped as designed)
```

Restored: 223 SUCCESS in Berlin, Los Angeles, Sydney, Tokyo and UTC. Full suite on the rebased tree: 13773 Berlin, 13759 Los Angeles.

Then I took `skipExcludedWeekend` out of the app and ran both versions over **every minute of 2026** for every Saturday and Sunday start, in eleven zones. Column one is master today, column two is your patch:

```
Australia/Lord_Howe      OLD shifted=  30  NEW shifted=0
Pacific/Chatham          OLD shifted=  60  NEW shifted=0
Europe/Berlin            OLD shifted=  60  NEW shifted=0
America/Los_Angeles      OLD shifted=  60  NEW shifted=0
Australia/Sydney         OLD shifted=  60  NEW shifted=0
Europe/London            OLD shifted=  60  NEW shifted=0
America/Havana           OLD shifted=  60  NEW shifted=0
America/Santiago         OLD shifted=  60  NEW shifted=0
Antarctica/Troll         OLD shifted= 120  NEW shifted=0
Asia/Tokyo               OLD shifted=   0  NEW shifted=0
UTC                      OLD shifted=   0  NEW shifted=0
```

Two things I wanted to know beyond your own case. Lord Howe springs forward by only 30 minutes, and Troll by two hours; both are fixed, so the patch is not quietly assuming a one-hour jump. And the restore is unconditional, which I wanted to rule out as a regression on the far more common path where the loop never runs: `setHours(getHours(), getMinutes())` did not move the instant once in 525,600 minute slots per zone, including the ambiguous repeated hour of every fall-back day. So it really is a no-op when there is nothing to restore, in zones that fall back on a weekday too.

The read-back you describe checks out, so this is not preview-only: `add-task-bar-parser.service.ts:192-201` derives `dueTime` from `new Date(dueWithTime).getHours()/getMinutes()`, and `add-task-bar.component.ts:627` writes exactly that into `startTime`.

## Notes, none of them blocking

**The `master:` row in your table is not master.** It shows a Sat 03-28 02:30 preview, which is pre-#9370 behavior. #9370 has merged, so today's master previews Mon 03-30 **03:30** and writes `startTime "03:30"`, which is precisely what my sabotage reproduces. The table reads as if master were fine; the row is really "before #9370".

**Your justification for the derived helper does not hold, but the helper is still worth keeping.** "A spec that cannot fail in `Europe/Berlin` would not have guarded this" rules out hardcoding *another* zone's transition, the way `date.service.tz.spec.ts:106-108` hardcodes US/Pacific. It does not rule out hardcoding Berlin's own 2026-03-29, which would have failed in the default lane. The real argument for deriving is different and better: it also guards in the zone each contributor actually runs locally, and it cannot rot if the lane matrix changes. That said, a hardcoded Berlin date plus the same four assertions is about twelve lines against the current fifty-three, so if you prefer the smaller spec I would take it. Your call, I am fine either way.

**The helper misses midnight transitions.** `findSpringForwardSunday` scans `hour = 1..5`, so zones that spring forward at 00:00 return `null` and the spec skips. Per the table above the bug is live in Havana and Santiago at 60 minutes each. Not worth chasing, since `"0:30am"` is not a string chrono handles well, and your comment already scopes the claim to the zones the suite runs in. Mentioning it so the gap is on record rather than as a request.

**The skip is silent.** `if (!gap) return` passes green. If the helper ever stopped finding a transition in a DST zone the spec would go quiet instead of complaining. `pending('no DST transition in this zone')` would surface it for free.

**One LA failure I could not attribute.** Very early on I saw a single spec fail in the Los Angeles lane on the rebased tree, once in the full suite and once in `short-syntax.spec.ts` alone. I did not capture the name, and it has not come back in 45 subsequent Los Angeles runs of that file plus six full-suite runs; a stale-build-cache theory did not reproduce either, and the baseline without your commit was 0 failures in 16 runs. The new spec cannot be it, since it takes an explicit `now` and reads no wall clock. I am recording it rather than pinning it on this PR, in case `short-syntax.spec.ts` turns out to have an order-dependent flake.

**Sibling.** #9014 touches both of these files and its base predates #9370 entirely, so it already conflicts with master in three other files. Whoever lands second resolves; nothing for you to do.

Approving on the strength of the fix. Rebase onto master with only `712a5ef8c` and it goes in. Thanks for the thorough writeup, and for the derived-transition helper, which is a better instinct than the convention it deviates from.

johannesjo:
Update on the blocker, and a correction: master is no longer red. #9396 landed the duplicate-key fix (`081a4e27e2`), so rebase onto the current tip whenever you like.

But the first commit now needs dropping for a **worse** reason than the conflict I quoted. Against the new master the merge is no longer a conflict at all:

```
$ git merge-tree --write-tree origin/master pr-9387
exit=0        # clean, and GitHub reports this PR as mergeable
```

Master pins the fixture near the top of the literal (`lastTaskCreation: undefined,
lastTaskCreationDay: undefined`, right after `id:`), your commit pins it at the bottom after
`defaultEstimate:`. Different hunks, so git takes both, and the merged file has
`lastTaskCreationDay` twice. I materialized that exact merged tree and ran it:

```
create-blocked-blocks-by-day-map.spec.ts:77:7 - error TS1117:
  An object literal cannot have multiple properties with the same name.
ERROR [karma-server]: Error: Found 1 load error
```

That is the same failure #9396 just cleaned up: no specs execute, the whole client suite goes
down. So clicking merge on this PR as it stands would re-break master, with a green
mergeability indicator and nothing in CI here to warn about it, because your `Tests` check
runs the pre-merge branch. Silent auto-merge into duplicate keys is exactly how the first
outage happened, so the guard is human for now: drop `5d5bdb5eb`, keep only `712a5ef8c`.

Nothing about the DST fix itself changes. #9386 is already closed. My approval stands for the
rebased single commit.

johannesjo:
Landed on master as `fa51168f90`, cherry-picked with your authorship intact, so I am closing this rather than merging it. Sorry for the slightly awkward mechanics: taking the single commit was the only way to land it without the duplicate-key merge described above, and I did not want to leave a PR sitting there that GitHub would happily let someone merge into a broken master.

Verified on the landed tree: 13776 in `Europe/Berlin`, 13762 in `America/Los_Angeles`, `npm run lint` clean.

Thanks for this one. Two things I appreciated beyond the fix being correct:

- Chasing the read-back into `startTime` yourself rather than stopping at "the preview is wrong". That is the difference between a cosmetic bug and a task that recurs an hour late forever.
- The derived-transition helper. I said in the review that your stated reason for it does not quite hold, and I stand by that, but the instinct was right for a better reason: it guards in whatever zone the contributor is actually sitting in, not just the two the CI matrix pins. I kept it as written.

I also pushed a small guard alongside it, `269797de10`, which turns on `no-dupe-keys`. The flat config never spread `js.configs.recommended`, so every core ESLint rule was off repo-wide and `checkFile` reported "All checks passed" on a literal with the same key twice. That is why three separate fixes to one fixture could pile up in a day without anything complaining. Lint now catches it, at zero cost to the rest of the repo.

On your two deliberate omissions: the `applyUserRepeatPick` date-picker divergence is still open and still worth doing if you feel like it. The "`@every weekday 2:30am` typed on the gap Saturday parses to no date at all" hole is upstream in chrono and I would leave it alone unless someone reports it.

PR #9388, opened 2026-07-30
fix(i18n): localize reminder notification texts and action buttons

### LINKED ISSUE #9344: 💥 notifications not localized
### Steps to Reproduce

1. Set language in preferences 
2. add task with reminder/notification
3. observe that informations like "Task reminder", "done", "snooze" are not included in localization / are not translated

### URL
https://localhost/#/tag/TODAY/tasks

### Meta Info
META: SP18.16.0_L2A __ Browser – de-DE – Linux aarch64 – de-DE – UA:Mozilla/5.0 (Linux; Android 17; Pixel 10 Pro Build/CP2A.260705.006; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/150.0.7871.181 Mobile Safari/537.36

### Actions Before Error
```

```

pratyushsinghal7:
## Problem

Reminder notifications ignore the app language. The notification body ("Task reminder" / "Due date reminder") and the action buttons ("Done", "Snooze 10m", "Snooze 1h") are hardcoded English strings, on iOS in `capacitor-notification.service.ts` and on Android in the native `ReminderNotificationHelper`. On iOS the body was also just `Reminder: <title>`, duplicating the title.

## Solution

New `F.REMINDER.N_*` translation keys, routed through ngx-translate:

- iOS/Capacitor: register the notification action types with translated titles and re-register on `onLangChange` (registerActionTypes simply overwrites the previous registration), and pick the translated body by reminder type.
- Android: notifications are built natively at alarm-fire time, long after the WebView may be gone, so the frontend now pushes the translated texts over a new optional bridge method (`updateReminderNotificationTexts`) before scheduling. The native side persists them in SharedPreferences so fired, snoozed, and reboot-re-registered reminders all use them. The previous English strings stay as native fallbacks, so older frontend/APK combinations keep working unchanged.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

## Checklist

## Testing

- New specs: the iOS action types are registered with translated titles and re-registered on language change, and the scheduled notification body uses the translated text per reminder type. The Android bridge push is not unit-testable (`androidInterface` is a module-level `window.SUPAndroid` capture), so that path was verified by review; older APKs without the new method are covered by optional chaining plus the native English fallbacks.
- Full suite: 13717 passing; the single failure is `create-blocked-blocks-by-day-map.spec.ts`, which also fails on master (hardcoded date)

Fixes #9344


johannesjo:
Thanks, this is a good fix and it does solve the reported case. Reviewed at `9585757a1`.

I walked the reporter's path end to end (German UI language, Android WebView, alarm fires with the WebView gone) and it works: the language is already in `globalConfig.localization.lng` at hydration, `de.json` resolves well inside the 5s `DELAY_SCHEDULE`, the effect pushes German texts, and `showNotification` reads them back at fire time. Persisting natively rather than passing the texts per alarm is also the right call, and worth spelling out because it is not obvious: `ReminderAlarmStore.AlarmData` carries no texts, so `BootReceiver` re-registering after a reboot and `ReminderActionReceiver` re-entering after a snooze both depend on a store that outlives the individual alarm. Threading five strings through those intents instead would have been strictly worse.

What I verified locally, rather than read: both specs green (16/16 and 27/27); four sabotages all reproduce red (hardcoding `title: 'Done'` fails the action-title spec, neutralizing the `onLangChange` re-registration fails the re-register spec, restoring the old template-literal body fails both body specs); `:app:compileFdroidDebugKotlin` builds; `npm run int` plus `checkFile src/app/t.const.ts` gives no diff, so `t.const.ts` really is generated; `checkFile` clean; `int:test` passes. Your note about `create-blocked-blocks-by-day-map.spec.ts` also checks out, that one is a separate problem on `master` and I have pushed a repair for it. And the iOS re-registration is safe as your comment claims: Capacitor's `registerActionTypes` ends in `center.setNotificationCategories(Set(createdCategories))`, a full replace, and this app registers exactly one category.

Three things I would like changed.

### 1. The Android push is in the wrong place, and it costs more than it looks

`androidInterface.updateReminderNotificationTexts?.(...)` sits inside the per-task loop, so it runs once per reminder per scheduling pass. Two consequences.

First, the scheduling passes are far more frequent than they look. `scheduleNotifications$` has no debounce, and `selectAllTasksWithReminder` derives from `selectAllTasksInActiveProjects`, which the SPAP-20 note at `task.selectors.ts:170` describes as returning "a NEW array every second while a task tracks time (its `timeSpent` bump replaces the entity ref)". So while a task is being tracked in the foreground, that effect re-runs its whole loop every second. The loop already re-registered the alarms, which is pre-existing, but this PR adds a JS-to-Java bridge call plus a `SharedPreferences.edit().apply()` commit plus five `instant()` lookups per reminder per second. On a battery-sensitive path that is worth avoiding.

Second, and this is the actual behavior gap: nothing is keyed to the language. The effects are driven by `combineLatest([<task selector>, this._reminderCfg$])`, `_reminderCfg$` is `cfg$.pipe(map((c) => c?.reminder), distinctUntilChanged())`, and the `updateGlobalConfigSection` reducer only replaces `state[sectionKey]`. Switching language is a `localization` update, so the `reminder` slice keeps reference identity and nothing re-emits. The texts refresh only the next time some unrelated task state changes, so a user who switches language and then leaves the app alone keeps the old language on alarms that are already scheduled.

One change fixes both: push once, from `onLangChange`, instead of from the loop. That is also the more correct trigger, because `onLangChange` fires only after the newly selected language's file is actually in the store, whereas the loop can run before that and push the `fallbackLang` English text. Roughly:

```ts
// in initialize(), which would then also need to run on Android
this._translateService.onLangChange.subscribe(() => this._pushAndroidReminderTexts());
```

While you are in there: the push currently sits inside the `try` that wraps `scheduleNativeReminder`, ahead of it, and that `catch` returns `false` without scheduling. It cannot throw today, but a text push failing should never be able to cost you the alarm.

### 2. The group summary in the same function stays English

```kotlin
.setContentTitle("Super Productivity")
.setContentText("Task reminders")
```

`ReminderNotificationHelper.kt:244-245`, about twenty lines below the lines you localized, in the same `showNotification`. It only surfaces once two or more reminders are grouped (Android suppresses the summary for a single child), so I would not overstate how often a user sees it, but it is an English literal sitting in the function this PR just translated.

Which surfaces the one design question worth deciding now rather than later: fixing this needs a sixth positional argument on `updateReminderNotificationTexts`, and the next one (`TrackingNotificationHelper` still has `.addAction(0, "Done", ...)` and `.addAction(0, "Pause", ...)`, and the channel names "Reminders" and "Reminders (Alarm)" plus their descriptions are English) would need more. A single JSON string parameter, or a key-value pair, would stop the signature growing per string. Your call, but it is cheaper to decide before this ships than after.

### 3. One of the three new specs cannot fail

`does not re-register on language change before the initial registration` asserts synchronously, but the Capacitor proxy dispatches asynchronously, which the spec right above it documents ("give the re-registration a macrotask to go through"). So the assertion holds no matter what the service does.

I proved it rather than guessing: I moved the `onLangChange` subscription out of `registerReminderActions` into a field initializer, which is exactly the regression this spec names, and **all 16 specs still passed**, including this one. Adding the same `await new Promise((resolve) => setTimeout(resolve));` the sibling spec uses turns it into a real guard.

### Smaller notes, take or leave

- `ReminderNotificationTexts.getText` falls back to the English default only when the stored value is null or blank, so whatever gets pushed is what shows, permanently, until something pushes again. Driving the push from `onLangChange` (finding 1) is what makes that safe, since the value is then guaranteed to be a loaded translation.
- iOS deadline reminders (`reminderType: 'DEADLINE'`, from the iOS-only `scheduleDeadlineNotifications$`) fall into the `N_BODY_TASK` branch and now read "Task reminder". Not a regression, the old `Reminder: <title>` did not say deadline either, and native Android always did the same. There is a translated `F.TASK.D_REMINDER_VIEW.DEADLINE_REMINDER` if you want it accurate.
- `SuperSyncBackgroundProvider` uses `"Task reminder"` as a task title fallback in three places, unconditionally in the `ACTION_SET_DEADLINE` branch, so background-sync-scheduled deadline reminders still get an English title. The new prefs store is right there.
- `@Synchronized` on `save` is not needed: every `@JavascriptInterface` method is invoked on the WebView's single JavaBridge thread, so there is no concurrent caller to serialize.
- On "not unit-testable": true of the current shape, but the repo already has the pattern for this. `local-backup.service.ts:150-152` wraps the same `window.SUPAndroid` singleton in thin private seams precisely so its specs can spy them instead of the global, and `CapacitorReminderService` already injects `IS_ANDROID_WEB_VIEW_TOKEN` for the same reason (see the comment at line 62), which the guard at line 121 then bypasses in favour of the module constant. Since the Android half is both the platform the issue was filed from and where the finding-1 defect lives, I would rather it were covered than reviewed.

johannesjo:
**🔧 Needs changes** - labelled `needs work`; ping me once the items below are addressed and I'll re-review.

Coming back to this after a while. The fix is still needed and still right in shape: nothing has localized these strings on master in the meantime, `ReminderNotificationHelper.kt:225-234` and `capacitor-notification.service.ts:104-113` (current master numbering) still hold the same English literals, and #9344 is still open. The three items from my review above also still stand exactly as written, since the head has not moved and I re-checked each against current master. So the only new things here are what master has done underneath the branch.

### Should fix before merge

- **`android/app/src/main/java/com/superproductivity/superproductivity/webview/JavaScriptInterface.kt`** - the branch no longer merges (339 commits behind). I resolved it file by file, and there is exactly one conflict: the import block, where this PR adds `import ...service.ReminderNotificationTexts` and master has since added `import ...service.RemoteTrackingNotificationHelper` in the same slot. Keeping both lines is the whole resolution; the other eight touched files merge clean, `t.const.ts` and `en.json` included.
- **`src/app/core/platform/capacitor-reminder.service.spec.ts:209`** (current master) - the rebase itself will turn 8 specs red, so this needs a line added while you are in there. Master gained a `describe('requestPermissionsInBackground (#9648)') > describe('on a native (non-legacy) platform')` block whose `TestBed` provides the real `CapacitorReminderService` class but no `TranslateService`. This PR makes that service `inject(TranslateService)`, ngx-translate 17's `TranslateService` is a plain `@Injectable()` with no `providedIn: 'root'`, and `src/test.ts` supplies no global one, so `TestBed.inject(CapacitorReminderService)` on the next line throws `NullInjectorError: No provider for TranslateService!` for all 8 `it(...)` in that block. Add `{ provide: TranslateService, useValue: mockTranslateService }` to its providers array, the same line you already added to the other five.

---

PR #9404, opened 2026-07-31
fix(add-task-bar): keep a workday recurrence off a menu-picked weekend date

KamilDev:
Follow-up to the note you left on #9387: *"the `applyUserRepeatPick` date-picker divergence is still open and still worth doing if you feel like it."*

## The bug

The add bar's date chip and its repeat menu can be put into a combination the occurrence engine will not honor. A Monday-to-Friday schedule has no weekend occurrence, so a config whose `startDate` is a Saturday starts on the Monday after it — `getFirstRepeatOccurrence` scans the weekday flags from `startDate`. The chip kept showing the Saturday.

This is the same divergence `skipExcludedWeekend` already fixes for the `@every weekday` phrase (#9370), reached through the two menus instead of the text. It happens in **either order**, because neither pick looked at the other's state:

- set a weekend date in the schedule dialog, then pick *Monday to Friday* from the repeat menu — `applyUserRepeatPick` never read `state.date`
- pick *Monday to Friday* first, then a weekend date — `applyUserDatePick` never read `state.repeat`

Your example, Europe/Berlin: `Standup`, date picker → **Sat 2026-03-28**, repeat menu → *Monday to Friday*. Chip stays Sat 03-28; the task is created on **Mon 03-30**.

## Why it is more than a wrong chip

I checked whether the Saturday that lands in `cfg.startDate` is inert, the way it is for the syntax path. It is not, and this is the part I would ask you to sanity-check yourself.

The repeat dialog hides the start date for this preset (`SETTINGS_WITHOUT_START_DATE`), so the user cannot see or correct it. But on **every** save with a non-`CUSTOM` quick setting, the dialog derives the new setting *from* that stored date:

```ts
// dialog-edit-task-repeat-cfg.component.ts:455-461
const referenceDate = currentRepeatCfg.startDate
  ? dateStrToUtcDate(currentRepeatCfg.startDate)
  : undefined;
const updatesForQuickSetting = getQuickSettingUpdates(
  currentRepeatCfg.quickSetting,
  referenceDate,
);
```

So switching that task to *weekly on current weekday* later runs `buildWeeklyForDay(Saturday)` and produces a **Saturday-only** recurrence, anchored to a day the user never picked and cannot see. `MONTHLY_NTH_WEEKDAY` reads the same date.

Being precise about the scope, since it is easy to overstate: the *first* occurrence is correct without this patch. Both `addTaskRepeatCfgToTask` effects route it through `getFirstRepeatOccurrence`, so the task itself lands on the Monday either way, and nothing reads the weekend date back into `startTime` the way the DST bug did. What this fixes is the chip lying, plus the start date that lie leaves behind in the persisted config.

## The fix

`skipExcludedWeekend` is now exported and reused, so the excluded-day rule lives in one place. `rollWeekendDateForRepeat` wraps it for the day-string side and returns `null` when nothing moves, so neither pick fires a redundant state write.

`MONDAY_TO_FRIDAY` is still the only preset that needs it: every other one either has no exclusions (`DAILY`) or derives its anchor from the date itself in `getQuickSettingUpdates`, which cannot contradict it. `INTERVAL` recomputes its flags from the picked date at submit, so it self-corrects.

Time is deliberately untouched — only the day is excluded, not the hour. The roll goes through `dateStrToUtcDate` → `skipExcludedWeekend` → `getDbDateStr`, so a Monday that is itself a midnight DST-transition day still comes out as that Monday.

The date rolled in `applyUserRepeatPick` is also recorded into `_previousParseResult.dueDate`, so the parse the strip queues reads the new state as unchanged instead of as a reset.

## Tests

Four in `add-task-bar-parser.service.spec.ts` (both orders, the time kept alongside the date, and two guards: a `DAILY` pick leaves a Saturday alone, a weekday date is not touched) and one in `add-task-bar.component.spec.ts` asserting the end of the chain — the persisted `cfg.startDate` and the task's `dueDay` are both the Monday.

Verified by breaking the fix before claiming they guard it. With `skipExcludedWeekend` no-op'd inside the helper: 3 FAILED / 114 SUCCESS in the parser spec and 1 FAILED / 55 SUCCESS in the component spec (`Expected '2026-03-28' to be '2026-03-30'`). Restored: 117 and 56 SUCCESS.

Full `npm test`, both lanes: **13780 passing / 16 skipped** in `Europe/Berlin` and in `America/Los_Angeles`. `npm run checkFile` clean on all four files.

## Not in this diff

The monthly presets diverge from the picked date in a different way — `MONTHLY_FIRST_DAY` and `MONTHLY_LAST_DAY` ignore `referenceDate` entirely and pin `startDate` off `new Date()`, so picking a date and then one of those presets leaves the chip on the picked day while the config anchors elsewhere. Separate cause, separate fix; mentioning it rather than folding it in.


src/app/features/tasks/add-task-bar/add-task-bar-parser.service.ts
```diff
@@ -417,8 +445,22 @@ export class AddTaskBarParserService {
    */
   applyUserRepeatPick(repeat: AddTaskBarRepeat): void {
     const cleanedInput = this._stripSyntaxForUserPick('repeat');
-    this._recordUserPick({ repeat, isRepeatFromSyntax: false });
+    // The date already on the chip can be a day this schedule excludes; the
+    // task would then be created on the following Monday and the chip would
+    // have advertised a first occurrence it never gets.
+    const rolledDate = rollWeekendDateForRepeat(this._stateService.state().date, repeat);
```
johannesjo: **Important:** `state().date` can still be owned by a plain due token, which `_stripSyntaxForUserPick('repeat')` intentionally keeps. Repro on this head: parse `Standup @2027-03-27`, choose Monday-to-Friday, then type one title character. The pick rolls state to Mon 03-29 but leaves the Saturday token; the next parse restores Sat 03-27 while the menu-owned repeat stays Monday-to-Friday (`Expected '2027-03-27' to be '2027-03-29'` in a targeted real-state spec). That makes the saved `startDate` depend on whether the user types again. Please either limit this roll to picker/control-owned dates, matching the PR's two-menu scope, or deliberately reconcile the due token, and add parse → pick → reparse coverage.

src/app/features/tasks/add-task-bar/add-task-bar-parser.service.ts
```diff
@@ -417,8 +445,22 @@ export class AddTaskBarParserService {
...
+      repeat,
+      isRepeatFromSyntax: false,
+      // Record the rolled date too, so the parse the strip queues sees the
+      // state it is about to read back as unchanged rather than as a reset.
+      ...(rolledDate ? { dueDate: rolledDate } : {}),
+    });
     this._stateService.updateRepeatSetting(repeat, cleanedInput);
+    if (rolledDate) {
+      // Time is left as it is — only the day is excluded, not the hour
+      this._stateService.updateDate(rolledDate);
```
johannesjo: **Important (pending-parse case):** `parseAndUpdateText()` snapshots state before awaiting `shortSyntax`. If a parse is genuinely pending (for example during the cold `chrono-node` load) and this repeat pick leaves the input unchanged, `_parseRunId` is not advanced and no replacement parse is queued. After this update writes Monday, the old parse republishes its snapshotted Saturday. A no-await regression on this head fails with `1 FAILED, 117 SUCCESS`, restoring Saturday after the pick first set Monday. Please re-read live user-owned state after the await/run-id guard, or queue an equivalent replacement parse. Simply canceling the old run without replacement risks dropping unrelated syntax results.

johannesjo (commented):
Two parser-state gaps remain. I independently reproduced both against `075efc1` with targeted specs. CI is green because the current tests exercise synchronous control calls, but not these parse/reparse paths.

KamilDev:
Both reproduced here, both fixed in `df3a3210d`. Thanks — the second one is a bug in the original commit's premise, not just in its scope.

**1. The syntax-owned date.** You are right that this made the saved `startDate` depend on whether the user types again, and I took the second of your two options rather than the first: the pair is now reconciled at the end of `parseAndUpdateText`, where both sides are final, in addition to the two picks.

I did not gate the roll on the date being picker-owned, because the bug this PR is about does not depend on who owns the date. `Standup @2027-03-27` plus a menu-picked `MONDAY_TO_FRIDAY` leaves exactly the invisible Saturday `startDate` the PR describes — the dialog derives `buildWeeklyForDay(Saturday)` from it just the same — so gating would have fixed two of the orders and declared the third out of scope on provenance grounds. There is a fourth: `defaultDate`, the day the bar was opened on, reaches the state through no pick at all, so it was never rolled either. Reconciling once at the parse covers all of them and stays covered, since the roll re-derives on every keystroke and is idempotent (a Monday rolls to itself and returns `null`).

What it does not do is edit the token. The text keeps saying `@2027-03-27` while the chip says the 29th, which is the same shape `@every weekday` already ships — the text states the request, the chip states the first occurrence the task will actually get. Deleting a plain date the user typed is what `_stripSyntaxForUserPick('repeat')` is gated against, and `should leave a plain date alone when a recurrence is picked` pins that.

**2. The pending parse.** Correct, and it is not limited to the roll — `parseAndUpdateText` snapshotted the whole state before the await and published from that snapshot, so any pick that landed during a parse and left the text unchanged got republished over. A deadline pick during the cold `chrono-node` load restores the previous deadline the same way; that is the fourth spec below.

I re-read live state after the run-id guard rather than queueing a replacement parse, since a replacement would drop the in-flight parse's tag/project/title work for text nobody changed — the concern behind the conditional `_parseRunId++` in `_stripSyntaxForUserPick`. Only `tagIdsFromTxt` is still read before the parse, because the parse itself takes it.

**Coverage** — four specs against the real `AddTaskBarStateService`, since what has to hold is that the state a pick writes survives the next parse, which a mocked `state()` cannot show:

- `should keep the workday recurrence off a weekend date the text still names` — your parse → pick → reparse repro
- `should keep it off a weekend day the bar was opened on` — the `defaultDate` order
- `should not let a parse that was in flight during the pick republish the weekend date`
- `should not let it republish any other value the pick replaced either` — the deadline case, no recurrence involved

Verified by breaking each fix first. Disabling the parse-side reconcile: `2 FAILED, 119 SUCCESS`, both `Expected '2027-03-27' to be '2027-03-29'` — your message. Restoring the pre-await snapshot: `2 FAILED, 119 SUCCESS`, `Expected '2027-03-27' to be '2027-03-29'` and `Expected '2027-04-01' to be '2027-05-01'`. Green: 121 (from 117).

Full `npm test`: **13784 passing / 16 skipped** in Berlin and LA.

src/app/features/tasks/add-task-bar/add-task-bar-parser.service.ts
```diff
@@ -269,6 +303,21 @@ export class AddTaskBarParserService {
       };
     }
 
+    // The date and the recurrence can come from opposite sides — a typed
+    // "@2027-03-27" with a menu-picked workday preset, or a date the bar was
+    // opened on with a recurrence from the text — and only a pick goes through
+    // applyUser*Pick. Reconciling the pair here as well covers every
+    // combination, and keeps it reconciled: a due token stays in the text and
+    // parses back to the excluded day on every following keystroke.
+    const rolledDueDate = rollWeekendDateForRepeat(
```
johannesjo: **Important:** This fixes the reparse while `MONDAY_TO_FRIDAY` remains selected, but loses the raw parsed date when the user later leaves that preset. Fresh real-state repro on `df3a3210d`:

1. Parse `Standup @2027-03-27` → Sat 03-27.
2. Pick Monday-to-Friday → Mon 03-29.
3. Pick Daily (or clear repeat) → still Mon 03-29 because the unchanged input queues no parse.
4. Type one title character → the retained token reparses and flips back to Sat 03-27.

Submitting at step 3 writes Monday to both `dueDay` and `startDate`; submitting after step 4 writes Saturday. A targeted regression expecting Saturday immediately after the Daily pick fails `Expected 2027-03-29 to be 2027-03-27` (1 failed / 121 passed), and a trace confirms the next parse causes the flip.

Please make the repeat transition and the next parse converge—e.g. preserve the raw parsed due date separately from the rolled effective date, or explicitly reconcile when leaving the excluding preset—and cover Workdays → Daily/clear before and after reparse.

johannesjo (changes requested):
The two parser-state gaps from my prior review are fixed at `df3a3210d`, and all current CI checks pass. One remaining Important transition still makes the persisted date depend on whether another parse happens before submission; see the inline finding.

KamilDev:
Reproduced and fixed in `f34712245`. You are right, and the cause is that the roll was only ever a one-way write: `applyUserRepeatPick` put the Monday into both the state and the pick record and kept nothing that said which day it had moved off, so no later change could undo it. The reconcile at the end of `parseAndUpdateText` re-derives the pair correctly, but only when a parse happens — and an unchanged input queues none, which is exactly the window your repro walks through.

**The fix.** `PreviousParseResult` gains `rolledFromDate`: the day a workday roll moved `dueDate` off, or null when nothing was rolled. Every recurrence change now re-derives from that day rather than from the roll's own output:

```ts
const currentDate = this._stateService.state().date;
const baseDate = this._dateBeforeWorkdayRoll(currentDate);
const dueDate = rollWeekendDateForRepeat(baseDate, repeat) ?? baseDate;
```

So Workdays rolls Sat → Mon and records the Saturday; Daily (or clear) re-derives from the Saturday and writes it back; and the next parse of the untouched text computes the same Saturday. Step 3 and step 4 of your repro now agree, and so does submitting at either point.

`_dateBeforeWorkdayRoll` only unwinds a roll that is still the one standing — `prev.rolledFromDate && prev.dueDate === date`. The date chip writes the state directly (`clearDate`, the schedule dialog), so without that check a stale record could resurrect a date the user has since cleared or replaced. `applyUserDatePick` keeps the field current for its own roll, and clears it when it rolled nothing, so an old day cannot resurface later.

**The clear button had to move.** `clearRepeatSetting()` in the actions component called `stateService.clearRepeatSetting()` directly, so it bypassed both the pick record and the in-flight-parse invalidation — it could not have undone the roll from where it stood. It now goes through `applyUserRepeatPick(null)`, which is the same control doing the same three steps, with the recurrence write branching on null at the end. That also gets it the `_parseRunId` bump it never had.

**Coverage** — five new specs, all against the real `AddTaskBarStateService` for the reasons from the last round (Workdays → Daily/clear, before and after reparse):

- `should give the weekend date back when the workday recurrence is replaced` — your exact repro, with the reparse asserted after the Daily pick.
- `should give it back when the recurrence is cleared instead` — same through the clear path.
- `should give back the day the bar was opened on, not the day it rolled to` — the `defaultDate` route, which reaches the state through no pick at all.
- `should not give back a date the user replaced after the roll` — a second weekend date picked while Workdays is active rolls again; leaving the preset returns *that* Saturday, not the first one.
- `should not resurrect a rolled date the user has since cleared` — `clearDate` between the two picks.

Plus one in the actions component spec pinning the clear routing.

**Guards verified by breaking each part first:** un-roll disabled → `4 FAILED / 121 SUCCESS`, three of them your `Expected '2027-03-29' to be '2027-03-27'`; the still-standing check dropped → `1 FAILED / 125`, `Expected '2027-03-27' to be null`; the clear routed back to the state service → `1 FAILED / 69` in the actions component spec. Restored: 126 and 70.

Full `npm test` on this head: **13790 passing / 16 skipped** in Europe/Berlin and America/Los_Angeles. `npm run lint`: 0 errors, only the pre-existing grandfathered `max-lines` warnings.

johannesjo (commented):
Re-reviewed the latest head `f347122` against the complete discussion and current CI. The prior two parser-state gaps are fixed, and the existing action/component specs pass. One Important control-date provenance gap remains; I reproduced it with two real-state specs against this exact head (`2 FAILED / 126 existing SUCCESS`). See the inline comment.

src/app/features/tasks/add-task-bar/add-task-bar-parser.service.ts
```diff
@@ -170,6 +208,7 @@ export class AddTaskBarParserService {
         // But if the user explicitly cleared a default date, keep it cleared.
         dueDate: defaultDueDate,
         dueTime: defaultDueTime,
+        rolledFromDate: null,
```
johannesjo: **Important:** `rolledFromDate` is reset here on every plain/no-syntax parse, even when `currentState.date` is the Monday produced by rolling a control-picked Saturday. I reproduced this on the exact head with the real `AddTaskBarStateService`:

1. Parse `Standup`.
2. Pick Sat `2027-03-27` in the date control.
3. Pick Monday-to-Friday → Mon `2027-03-29`.
4. Type one title character and let the ordinary parse land.
5. Pick Daily (or clear recurrence).

The final expectation fails: `Expected '2027-03-29' to be '2027-03-27'` (`1 FAILED / 126 existing SUCCESS`). A second controls-before-any-title variant fails the same way because `_recordUserPick()` is a no-op while `_previousParseResult` is null.

This means Workdays → Daily/clear still silently keeps Monday depending on whether/when the parser has run. Please keep the active roll provenance independent of ordinary parse results (or deliberately carry it forward while the recorded effective date still matches state), and cover a control-picked date both across a reparse and before the first title parse.

KamilDev:
Reproduced and fixed in `8fdbdf7e7`. Both of your variants come from the same thing: `rolledFromDate` was a field of the parse result, and a parse result is rebuilt from scratch every parse. A parse of text that names no date *carries* the rolled day over from the state, which is not the same as producing it — so it wrote `rolledFromDate: null` next to a date that was very much still rolled. Your second variant is the other end of the same mistake: the record lives behind `_recordUserPick()`, which is a no-op while `_previousParseResult` is null, so controls used before the first parse never got to write one.

**The fix.** The roll is tracked apart from the parse result, as the roll currently standing:

```ts
private _workdayRoll: { from: string; to: string } | null = null;
```

`_dateBeforeWorkdayRoll` keys off `roll.to === date` instead of `prev.dueDate === date`, so the "still the one standing" check that guards `clearDate` and the schedule dialog is unchanged. `rolledFromDate` is gone from `PreviousParseResult`, which deletes both of the `null` initializers rather than putting a condition on them, and there is now one writer for the pair — `_rollForRepeat(base, repeat)` rolls and records in one step, and every path that can change either half goes through it: both picks and the reconcile at the end of `parseAndUpdateText`.

**One thing I did not carry over.** A date the parse takes from its *own text* is not unwound first:

```ts
const isDueDateFromText = !!parseResult?.taskChanges.dueWithTime;
```

A day the text names is the user stating the choice again, exactly like a pick. Unwinding it would mean that editing the token to `@2027-03-29` — naming the Monday explicitly — leaves the roll still claiming the Saturday behind it, so clearing the preset snaps back to a day the text no longer mentions, and the next keystroke undoes that again. That is the same class as the two you already found, so only the day carried over from the state is treated as this service's own output.

`resetPreviousResult()` clears the roll too.

**Coverage** — three more against the real `AddTaskBarStateService`:

- `should still give the date back after an ordinary parse in between` — your repro, control-picked Saturday, Workdays, a parse of text with no date in it, then Daily.
- `should give it back for controls used before the first parse` — your second variant, no parse at all before the picks.
- `should take the date the text names as the day it was picked as` — the edge above.

**Verified by breaking each part**, all at `128 SUCCESS` of 129:

- `_workdayRoll = null` before the reconcile, i.e. the old per-parse reset → `Expected '2027-03-29' to be '2027-03-27'`, your message.
- the write gated on `_previousParseResult` → `should give it back for controls used before the first parse`.
- unwinding unconditionally → `Expected '2027-03-27' to be '2027-03-29'`.

Full `npm test` is 13793 passing / 16 skipped in `Europe/Berlin` and in `America/Los_Angeles`, and `npm run lint` reports 0 errors (only the pre-existing `max-lines` warnings).

johannesjo (commented):
Re-reviewed the latest head `8fdbdf7e` with two general reviewers and two UX/UI-focused reviewers. CI is green, but four user-visible transition gaps remain:

1. **A contextual repeat option can save a different recurrence than its label.** Saturday → Workdays displays Monday, so the menu offers “Every week on Monday”; selecting it restores the pre-roll Saturday before deriving the new preset and saves Saturday instead. Monthly and yearly contextual presets have the same anchor mismatch. The label and selection need to derive from one date anchor, with a Workdays → contextual-preset regression. ([parser](https://github.com/super-productivity/super-productivity/blob/8fdbdf7e7f3c7109693a71800f9858422d13e752/src/app/features/tasks/add-task-bar/add-task-bar-parser.service.ts#L496), [labels](https://github.com/super-productivity/super-productivity/blob/8fdbdf7e7f3c7109693a71800f9858422d13e752/src/app/features/tasks/add-task-bar/add-task-bar-actions.component.ts#L207-L221))

2. **The date-less weekend path keeps a weekend recurrence anchor.** On Saturday or Sunday, selecting Workdays with no explicit date falls back to `todayStr()` for both `dueDay` and `repeatCfg.startDate`. The occurrence logic can advance the task to Monday, while the hidden recurrence start remains on the weekend and can affect later preset changes. Please compute one effective rolled start date and use it for both writes; cover no-date and cleared-date weekend flows. ([due date](https://github.com/super-productivity/super-productivity/blob/8fdbdf7e7f3c7109693a71800f9858422d13e752/src/app/features/tasks/add-task-bar/add-task-bar.component.ts#L556-L559), [repeat start](https://github.com/super-productivity/super-productivity/blob/8fdbdf7e7f3c7109693a71800f9858422d13e752/src/app/features/tasks/add-task-bar/add-task-bar.component.ts#L604-L621))

3. **Successful submission discards the source date while preserving the rolled date.** Saturday → Workdays becomes Monday. After submission, `resetAfterAdd()` clears recurrence but preserves Monday, then `resetPreviousResult()` drops `_workdayRoll`; the next ordinary task inherits Monday instead of the sticky Saturday choice. Restore `from` during reset when the current date still equals `to`, and test Workdays → submit → consecutive non-recurring task. ([reset](https://github.com/super-productivity/super-productivity/blob/8fdbdf7e7f3c7109693a71800f9858422d13e752/src/app/features/tasks/add-task-bar/add-task-bar-parser.service.ts#L459-L462))

4. **The automatic date move and restoration are silent for screen-reader users.** Focus returns to the title field after the repeat interaction, and the changed date button has no live/status announcement. Please announce automatic movement/restoration through a translated polite live region and add an accessibility-focused component test. ([date update](https://github.com/super-productivity/super-productivity/blob/8fdbdf7e7f3c7109693a71800f9858422d13e752/src/app/features/tasks/add-task-bar/add-task-bar-parser.service.ts#L511))

The first three affect persisted scheduling behavior and should be addressed before merge. The fourth is an accessibility regression introduced by the automatic rewrite.

KamilDev:
All four reproduced and fixed in `d4fbc6d68`. The first three were the same shape as the earlier rounds — the roll was consistent inside the pair it moved, and wrong at every edge of it — so this round fixes the edges rather than the mechanism.

**1. The contextual options.** Confirmed: the labels come from `state().date` and the pick re-derives from `dateBeforeWorkdayRoll(state().date)`, so Saturday → Workdays offered "Every week on Monday" and saved a Saturday recurrence. `MONTHLY_CURRENT_DATE`, `MONTHLY_NTH_WEEKDAY` and `YEARLY_CURRENT_DATE` read the same `refDate`, so they were off by the same two days.

I anchored both halves on the **pre-roll** day rather than the displayed one. Going the other way — letting the options anchor on the Monday — would make the day the user chose mutate into a day the app chose, which is exactly what rounds 2 and 3 removed: the next "leave the preset" would then have a Monday to give back, not the Saturday that was picked. So `dateBeforeWorkdayRoll` is public now and `_repeatRefDate` goes through it. `_workdayRoll` became a signal so the label recomputes when the roll changes, and `repeatDisplay`'s interval `weekdayStr` is unaffected (an interval repeat can never have a roll standing — `_rollForRepeat` clears it for every non-`MONDAY_TO_FRIDAY` schedule).

**2. The date-less weekend path.** Confirmed, and it is the one place the roll never ran at all: a recurrence with no date never passes a date through the parser, so `todayStr()` reached both `taskData.dueDay` and `repeatCfg.startDate` unrolled. Picked on a Saturday, the effect moves the task to the Monday via `getFirstRepeatOccurrence` while the config keeps a Saturday start date the dialog hides — and derives from on the next quick-setting change. Both writes now take one `_repeatStartDate(state)`, which rolls `state.date || todayStr()`. It is idempotent for a date that came from the chip (that one is rolled already), so the existing pick path is unchanged. `rollWeekendDateForRepeat` moved out of the parser service into `roll-weekend-date-for-repeat.ts` so both callers share it.

**3. The reset.** Confirmed. `resetAfterAdd()` keeps the date and clears the recurrence, so the roll's own reason to exist is gone one line before `resetPreviousResult()` drops the record of it. It now gives the day back first, guarded on the roll still standing (`roll.to === state.date`) so it cannot resurrect a date something else replaced — same guard as the pick path.

**4. The announcement.** Added a polite live region in the actions component, rendered from the start and empty until something moves, so the announcement is not the first thing the region ever contains. The parser publishes the move it made (`workdayDateMove`: `MOVED` / `RESTORED` plus the day), the component translates and formats it: `A11Y_DATE_MOVED_FOR_WORKDAYS` / `A11Y_DATE_MOVED_BACK`. Re-deriving the roll already standing on every keystroke does not re-announce, and a date the user set themselves is never announced — only the two automatic transitions are.

**Guards, each verified by breaking it first:**

- label anchored back on `state().date` → `2 FAILED / 72 SUCCESS`, both `Expected 'Every week on Monday' to be 'Every week on Saturday'`
- announcement not published → `1 FAILED / 73`, `Expected '' to be 'Date moved to Monday, March 29'`
- submit reverted to `todayStr()` / `state.date || todayStr()` → `2 FAILED / 57` in the component spec, four × `Expected '2024-05-19' to be '2024-05-20'`
- reset un-roll removed → `1 FAILED / 58`, `Expected '2026-03-30' to be '2026-03-28'`

**New specs.** Actions component, against the **real** parser and state (new `describe('Repeat Setting against real add bar state')`, `TestBed.resetTestingModule()` then real providers — the label and the saved day come from opposite sides of the roll, so a mocked parser cannot show them agreeing): `should save the recurrence its label offered after a workday roll`, `should announce a date it moved by itself, and moving it back`, `should stay quiet about a date the user set themselves`, plus `should build the labels from the day behind a workday roll` on the mocked fixture. Add-bar component: `should start a date-less workday repeat on the Monday after a weekend today`, `should do the same for a date the user cleared before picking it`, `should hand the next task the picked day, not the day the roll produced`.

Actions spec 74 (from 70), add-bar component spec 59 (from 56), parser spec unchanged at 129. Full `npm test` **13800 passing / 16 skipped** in `Europe/Berlin` *and* `America/Los_Angeles`; `npm run lint` 0 errors, only the pre-existing `max-lines` warnings.

johannesjo:
Re-reviewed the current head `d4fbc6d6817c23c4914d65d7f684f76e48cbc01e` against the full PR, discussion, thread state, and current CI. The four gaps from the previous round are fixed, and all required checks are green. Two accessibility transitions remain:

1. **Important — the post-submit restoration is still silent.** After Saturday → Workdays moves the chip to Monday, submitting clears the recurrence but deliberately keeps the date sticky. `resetPreviousResult()` then changes the still-mounted add bar back to Saturday for the next task, but immediately clears `workdayDateMove` instead of publishing `RESTORED` ([parser lines 457–463](https://github.com/super-productivity/super-productivity/blob/d4fbc6d6817c23c4914d65d7f684f76e48cbc01e/src/app/features/tasks/add-task-bar/add-task-bar-parser.service.ts#L457-L463)). The bar is reset, not gone: focus returns to its title field and the next non-recurring task now inherits Saturday. A sighted user sees the date chip change; a screen-reader user gets no status update. I extended the existing sticky-date component spec with the expected restoration event and reproduced `1 FAILED / 58 existing SUCCESS`: `Expected null to equal { type: 'RESTORED', date: '2026-03-28' }`. Please publish the restoration when the reset actually calls `updateDate(roll.from)`, and cover the live region after submit.

2. **Minor — distinct weekend picks that land on the same Monday are deduplicated.** With Workdays active, pick Saturday (announced as Monday), then pick Sunday. The second user action is another automatic Sunday → Monday adjustment, but `_rollForRepeat()` compares only `roll.to` ([parser lines 584–595](https://github.com/super-productivity/super-productivity/blob/d4fbc6d6817c23c4914d65d7f684f76e48cbc01e/src/app/features/tasks/add-task-bar/add-task-bar-parser.service.ts#L584-L595)), so it updates the provenance to Sunday without publishing a new move; the live-region text also remains identical. A real-state actions spec reproduced `1 FAILED / 74 existing SUCCESS` because the second pick left the exact same announcement object in place. Parser re-derivation should stay quiet, but a new control/date-token choice needs a repeatable announcement—e.g. distinguish the transition source and include source + destination, or otherwise clear/re-publish identical live messages.

The scheduling/state behavior itself now looks consistent; these findings are confined to the accessibility contract introduced by the latest follow-up.

KamilDev:
Both fixed in `7a544cfb2`. Thanks — the second one is the more interesting of the two, because it turned out the payload could not express the event.

**1 — the post-submit restoration now announces.** `resetPreviousResult()` publishes `RESTORED` exactly where it calls `updateDate(roll.from)`, and clears the region only when it gave nothing back. The comment that justified clearing it was wrong on the facts, as you say: the bar is reset, not gone, so this is the same unasked-for date change as the roll itself — the day the next task will be given just moved on a control nobody is focused on.

The guard is on the roll still standing (`roll.to === state.date`), the same one the un-roll itself uses, so a date something else replaced in the meantime is neither given back nor announced.

**2 — the day moved off is now part of the move, in the payload and in the sentence.** `WorkdayDateMove` is `{ type, from, to }`; the dedup compares the whole pair rather than `roll.to`.

Both halves are needed and neither is sufficient, which is what makes this a real gap rather than a stale comparison:

- Comparing only the destination drops the second pick entirely — nothing is published.
- Publishing it but keeping "Date moved to Monday, March 29" leaves the live region holding the text it already had, and an unchanged region is a silent one. So `A11Y_DATE_MOVED_FOR_WORKDAYS` reads "Date moved from {{fromDateStr}} to {{toDateStr}}, the first workday of this repeating task" now.

Parser re-derivation stays quiet, and it is quiet at the source: re-deriving the standing roll on a keystroke publishes no new move at all, so `workdayDateMove()` keeps the very same object. Worth pinning, because the render-level dedup (an equal string means Angular writes nothing) is an accident of change detection, not a contract — a consumer that reacted to the signal instead would announce on every keystroke.

One case I deliberately left silent: picking the *same* weekend day twice. It reproduces the identical adjustment of the identical day, the sentence in the region already describes it correctly, and re-announcing it would need a clear-then-re-set across a tick — talking over the title field to say nothing new. Happy to add it if you disagree.

**Test counts on this head** (`7a544cfb2`): parser spec 131 (from 129), actions spec 76 (from 74), add-bar component spec 59 (unchanged, the sticky-date spec extended as you suggested). Full `npm test` **13804 passing / 16 skipped** in Berlin *and* LA, watched print in both. `npm run lint`: 0 errors, only the grandfathered `max-lines` warnings and two pre-existing unused-disable directives, none in the touched files.

**Guards verified by breaking each part first:**

- Reset announcement back to an unconditional clear → `1 FAILED / 58 SUCCESS` in the add-bar component spec, `Expected null to equal Object({ type: 'RESTORED', from: '2026-03-30', to: '2026-03-28' })` (your repro), and `1 FAILED / 75` in the actions spec, `Expected '' to be 'Date moved back to Saturday, March 27'`.
- Dedup back to `roll.to !== previous?.to` → `1 FAILED / 130` in the parser spec (`Expected $.from = '2027-03-27' to equal '2027-03-28'`) and `1 FAILED / 75` in the actions spec (`Expected 'Date moved from Saturday, March 27 to Monday, March 29' to be 'Date moved from Sunday, March 28 to Monday, March 29'`) — your repro, at both levels.
- Dedup removed entirely (publish on every re-derive) → `1 FAILED / 130`, on the same-object assertion.
- Announcement without the day moved off → `2 FAILED / 74` in the actions spec, both weekend picks reading the same sentence.

**New specs:** `should publish no new move for a roll it only re-derived` and `should publish a move for each weekend day that lands on the same Monday` (parser, real state); `should announce a second weekend day moved to the same Monday` and `should announce the day the reset after an add gives back` (actions, real parser + real state, asserting the rendered live region); plus the restoration expectation folded into `should hand the next task the picked day, not the day the roll produced`.

johannesjo:
Thank you very much! <3

Issue #9400, opened 2026-07-30
Bundle the web app into the SuperSync server image (one-container self-hosting)


johannesjo:
## Motivation

Self-hosting Super Productivity today means running the SuperSync server yourself but still loading the web client from `app.super-productivity.com` (or installing a desktop/mobile app). With the SuperSync image now free of hosted-instance specifics (#9390, #9395) and about to be publicly pullable, the natural next step is a **single container that serves both the sync server and the web app** — `docker compose up` and you have the whole product on your own domain, no external dependencies.

This also fixes a version-coherence gap that exists today: a self-hosted server tracks `master` while the hosted web app tracks releases. A bundled image ships frontend and server **from the same commit**.

Related: #3305 (self-hosting the web version with server-side persistence — this plus SuperSync is the modern answer to it).

## Design

Bundle the built web app into the SuperSync server image and let the existing static-file layer serve it:

1. **Build**: add a frontend build stage to `packages/super-sync-server/Dockerfile` (the root `Dockerfile` already builds the web app — reuse its recipe), producing `dist/browser` with `--base-href /app/`.
2. **Serve**: copy the bundle into the image's `public/app/`. `@fastify/static` already serves `public/` wholesale, so the app is reachable at `/app/` with **zero server code changes**. The client uses hash routing (`withHashLocation`, `src/main.ts`), so no SPA history-fallback is needed — it is genuinely static.
3. **Zero-config sync**: when the app is served from `/app/` of a sync server, default `superSync.baseUrl` to `window.location.origin` instead of the hardcoded `https://sync.super-productivity.com` (`default-global-config.const.ts`). Same origin also means **CORS configuration stops being a self-hoster concern entirely** for the bundled app.
4. **Landing page**: link `/app/` from the server's generated index page so the container is self-explanatory.

## Release-gated, not master-tracking

The bundled build must **not** ride the `:latest` master-push image:

- The frontend build is heavy (~10 min lint+build) and would retrigger on every `src/**` change, versus the server image's narrow path filter today.
- More importantly, `:latest` is master — bundling there would hand self-hosters **unreleased client code** while desktop/mobile users are release-gated.

So: build the bundled variant on release tags only (e.g. `ghcr.io/super-productivity/supersync:v18.x-bundle`, exact naming TBD). The lean server-only `:latest` keeps tracking master unchanged.

## The two careful bits

[...]

PR #9413, opened 2026-08-01
feat(sync): default superSync.baseUrl to window.location.origin when …

riyahref:
…bundled #9400

## Problem

Self-hosting Super Productivity today requires the web client to load from
`app.super-productivity.com` even when running your own SuperSync server —
so self-hosters running the sync server still have to manually enter a
custom `superSync.baseUrl` to point the app at their own instance instead
of the hosted one.

## Solution

Adds a small detection helper, `isServedFromBundledSyncServer()`, that
checks whether the app is currently being served from a `/app/` path (the
path the bundled SuperSync server image will serve the static web app
from, per the design in #9400). When true, `superSync.baseUrl` now
defaults to `window.location.origin` — i.e. the same server the page is
already being served from — instead of the hardcoded hosted URL. This
requires zero configuration from the self-hoster: syncing "just works"
against their own server the moment the app loads under `/app/`.

Behavior is unchanged for the existing hosted web app, Electron, and
mobile builds, since none of those are served from a `/app/` path.

This addresses the first acceptance criterion in #9400 — "syncs against
the same origin with no manual URL entry" — and is scoped to just the
config-default piece of that issue; the Dockerfile build stage, static
serving, and landing-page link are separate, not part of this PR.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/config/default-global-config.const.ts
+++ b/src/app/features/config/default-global-config.const.ts
@@ -10,6 +10,9 @@ import { INBOX_PROJECT } from '../project/project.const';
 import { DEFAULT_MAX_BACKUP_FILES } from '../../../../electron/shared-with-frontend/backup-file-cleanup.util';
 
 const minute = 60 * 1000;
+const isServedFromBundledSyncServer = (): boolean =>
+  typeof window !== 'undefined' && window.location.pathname.startsWith('/app/');
+
 const defaultTaskNotesTemplate = `**How can I best achieve it now?**
 
 **What do I want?**
@@ -251,9 +254,11 @@ export const DEFAULT_GLOBAL_CONFIG: GlobalConfigState = {
     },
 
     superSync: {
-      baseUrl: environment.production
-        ? 'https://sync.super-productivity.com'
-        : 'http://localhost:1901',
+      baseUrl: isServedFromBundledSyncServer()
+        ? window.location.origin
+        : environment.production
+          ? 'https://sync.super-productivity.com'
+          : 'http://localhost:1901',
       userName: null,
       password: null,
       accessToken: null,


src/app/features/config/default-global-config.const.ts
```diff
@@ -10,6 +10,9 @@ import { INBOX_PROJECT } from '../project/project.const';
 import { DEFAULT_MAX_BACKUP_FILES } from '../../../../electron/shared-with-frontend/backup-file-cleanup.util';
 
 const minute = 60 * 1000;
+const isServedFromBundledSyncServer = (): boolean =>
+  typeof window !== 'undefined' && window.location.pathname.startsWith('/app/');
```
johannesjo, line 14: **Important:** This path-only test also matches the packaged Flatpak renderer. The [official Flathub manifest](https://github.com/flathub/com.super_productivity.SuperProductivity/blob/master/com.super_productivity.SuperProductivity.yml) installs Super Productivity under `/app/superproductivity`, while `electron/main-window.ts` loads the packaged renderer through `file://`. A fresh Flatpak launch therefore has a pathname beginning with `/app/`, and this branch sets `superSync.baseUrl` to a non-HTTP file origin (`file://` in Chromium) instead of the hosted server. That breaks SuperSync setup and contradicts the PR's “Electron unaffected” claim.

Please require an HTTP(S) protocol (or otherwise explicitly exclude packaged Electron) before treating `/app/` as the bundled server, and add regression coverage for both `https://host/app/` and a Flatpak-shaped `file:///app/...` URL. A protocol guard is the smallest fix.

johannesjo (commented):
## Review summary

I found one blocking platform regression in the current implementation; details are inline.

The change is otherwise small and clean. Against the current `master` merge tree, `checkFile` passed for the changed file, as did the 3 default-config tests and 39 sync-config tests. Those existing tests do not exercise the new URL branch. Only the welcome check currently reports on GitHub; the required CI, CodeQL, and E2E gates have not run yet.

johannesjo:

Thanks for the PR, and sorry it sat so long. This defaults `superSync.baseUrl` to `window.location.origin` when the app is served under `/app/`. The head hasn't moved since the earlier inline review, so that review's `/app/`-vs-packaged-Flatpak point still stands; I won't restate it, and everything below is only what I can add to it.

### Blocking

- **`src/app/features/config/default-global-config.const.ts:13-14`** – the standing Flatpak item is worse than a broken local setup, which is worth knowing before you pick a fix: the bogus value does not stay on the affected device. `superSync` is not a local-only sync key – `LOCAL_ONLY_SYNC_KEYS` in `src/app/features/config/local-only-sync-settings.util.ts` is `['syncInterval', 'isManualSyncOnly', 'syncProvider', 'isEnabled', 'isEncryptionEnabled']`, and `stripLocalOnlySyncSettingsFromGlobalConfig` only drops the two schedule keys and nulls `syncProvider`, so `sync.superSync.baseUrl` rides along in every upload. On the receiving side `global-config.reducer.ts:150-151` merges `{ ...DEFAULT_GLOBAL_CONFIG.sync, ...appDataComplete.globalConfig.sync }` and `withLocalOnlySyncSettings` re-applies only those five keys, so a `file://` baseUrl can overwrite a correct one on the user's other devices.

  That merge is symmetric, so the window is narrower than it first looks: a device that downloads before it uploads adopts the peer's correct value instead. It needs the affected install to be the one seeding or first pushing full state – a fresh install as the first device, or the first to push a snapshot. Real, but not universal. Whichever guard you land on, the invariant worth coding to is that a non-http(s) origin must never become this default.

### Nits

- **`src/app/features/config/default-global-config.const.ts:13-14`** – as structured, the regression coverage that was asked for can't actually be written, so this is worth folding into the same change. `isServedFromBundledSyncServer` is module-private and `DEFAULT_GLOBAL_CONFIG` calls it at module-evaluation time, and 29 spec files already import this module, so the value is frozen for the whole Karma bundle before any `beforeEach` runs. Stubbing isn't a way out either: `window.location` is non-configurable in Chrome. Note that merely exporting the helper doesn't fix it – it has to take the location or URL as a **parameter** instead of reading `window.location` internally. `default-start-page.util.ts` and `local-only-sync-settings.util.ts` in the same directory are the existing pattern, both with a co-located `.util.spec.ts`.

### Questions

- Are you planning to also do the Dockerfile build stage and the `public/app/` serving from #9400, or is this meant to land standalone ahead of them? On current master nothing serves the app under `/app/` – the SuperSync image has no frontend build stage yet – so the new branch is unreachable in every shipped artifact except the Flatpak case above. That shapes whether this is better merged now or held to go in with the rest, which is part of what I've asked the maintainer.

---

johannesjo:
Two decisions here:

**1. Change the signal — use the build-time `<base href>`, not the URL path.** A protocol guard fixes the Flatpak case but still misfires for anyone self-hosting the plain web bundle under `/app/` without a sync server. Only a #9400 bundled build is built with `--base-href /app/`, so the base-href value discriminates exactly the case you want and can't be spoofed by an install path.

**2. Hold this until the rest of #9400 lands.** Nothing on master serves the app under `/app/` yet — the SuperSync image has no frontend build stage — so today the new branch is unreachable except as the Flatpak regression. Merging it now buys nothing and costs the Flatpak default, so let's land it together with the Dockerfile build stage and the `public/app/` serving.

Keeping it open and parked, not rejected — the piece itself is right.

PR #9427, opened 2026-08-03
fix(styles): replace deprecated variables and clean up CSS tokens

SalmaAbdelrhmanMostafaMahmoud:
…lacing with correct ones

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


SalmaAbdelrhmanMostafaMahmoud:
Sorry for the delay in resolving this issue! 
Please let me know if there are any adjustments, missing parts, or additional changes you'd like me to make.

johannesjo:
Thanks for the cleanup. I reviewed the current head `eb5f795` and also reproduced the affected evaluation state in the deployed preview. The `color-mix()` conversions for the old RGB-channel variables are directionally correct, but a few token mappings need changes before this is ready.

### Important

1. **The replacement text tokens are still undefined at runtime.**

   - `src/app/features/metric/dialog-productivity-breakdown/dialog-productivity-breakdown.component.scss:31` uses `---text-color-less-intense` (three leading hyphens).
   - `src/app/features/metric/evaluation-sheet/evaluation-sheet.component.scss:146,170` uses `--palette-foreground-secondary-text`.

   Both variables resolve to an empty value in the deployed preview. In the live daily-state card, the label and hint therefore inherit the full body color (`rgb(44, 44, 44)`) rather than muted text; `--text-color-muted` is present and resolves to the intended muted foreground. Please use `var(--text-color-muted)` at all three secondary-text sites.

   The same evaluation component maps `--fg-primary` to `--c-primary` at line 159. That is the brand color, not primary foreground text, and the preview renders the whole daily-state headline blue. The semantic replacement is `var(--text-color)`.

2. **The new `--primary-color` alias bypasses the active theme.**

   `src/styles/_css-variables.scss:220` hard-codes it to `#3f51b5`, and the restore-point dialog consumes it at lines 71 and 106. In the deployed preview, the active `--c-primary` is `rgb(100, 149, 237)`, so these values already disagree without selecting a custom theme. This pins the selected restore point outline/icon to indigo instead of following the user's palette. The smallest fix is to remove the new global alias and use the existing `var(--c-primary)` in the restore-point component.

### Minor

- `src/app/features/archive/dialog-archive-compression/dialog-archive-compression.component.scss:25`: this is an actual error state (`<mat-icon>error</mat-icon>`) and its previous fallback was red `#f44336`; mapping it to orange `--color-warning` changes the semantics. Please use `--color-danger`, consistent with the other error replacements in this PR.

- `src/styles/page.scss:251`: the old expression used alpha `1` (100% accent); changing it to a 10% color mix makes the 4px blockquote marker nearly invisible. Since `--c-accent` is already a complete color, the equivalent valid replacement is simply `border-left: 4px solid var(--c-accent)`.

All current CI checks pass, but they do not catch undefined custom-property references. **Verdict: needs changes** for the two Important items.

johannesjo (commented):
Thanks for the update. I re-reviewed the exact `eb5f795..d62521e` force-push revision and the resulting current head `b7d7128` with three independent read-only passes, then verified each claim against the token definitions, component context, custom themes, discussion, and current CI.

The earlier requested corrections are now addressed: the text mappings, restore-point theme color, archive error semantics, and blockquote accent are all fixed. Two issues remain:

### Important

1. **`src/app/features/tasks/add-task-bar/add-task-bar.component.scss:346` — the active toggle state uses an undefined token.**

   The new background and foreground declarations reference `--primary`, but the application and its themes do not define that variable. When search mode or the note field is active, the `color-mix()` background is therefore invalid and the Material icon keeps the muted token set by the common button rule. This removes the visible selected-state cue.

   The current base already has the correct theme-aware implementation. Please keep:

   ```scss
   --mat-icon-button-icon-color: var(--brand);
   background-color: var(--state-selected);
   ```

   and remove the ineffective `opacity`, `color-mix(... --primary ...)`, and direct `color: var(--primary)` declarations.

2. **`src/app/features/archive/dialog-archive-compression/dialog-archive-compression.component.scss:44` — the border bypasses custom-theme divider overrides.**

   `--divider` is the base primitive, while components consume the semantic `--divider-color` alias. Many bundled themes override only `--divider-color`; using the primitive here makes these stat borders fall back to the default light/dark divider instead of matching the active theme.

   Please use:

   ```scss
   border: 1px solid var(--divider-color);
   ```

All current CI workflows pass. **Verdict: needs changes** for the two items above.

johannesjo:

Thanks for the PR! The token replacements are directionally right, but the two items from the 2026-08-06 review on this same head (`b7d7128`) are still open, and master has moved on since – the branch is now 202 commits behind, GitHub reports it as conflicting, and most of these fixes have meanwhile landed on master.

### Blocking

- **`src/app/features/tasks/add-task-bar/add-task-bar.component.scss:346`** – same issue as in the previous review: `--primary` is not defined anywhere in the app (not in `src/styles/_css-variables.scss`, not in any bundled theme, and nothing sets it from TS; the only `--primary` in the repo belongs to the sync server's standalone landing page). Both `background-color: color-mix(in srgb, var(--primary) 15%, transparent)` and `color: var(--primary)` are invalid at computed-value time, so the active search/note toggle loses its selected-state cue. Current master already has the correct theme-aware version; on rebase, drop this hunk and keep:

  ```scss
  --mat-icon-button-icon-color: var(--brand);
  background-color: var(--state-selected);
  ```

- **Rebase onto master and shrink the PR to what is still missing.** 9 of the 15 files are already fully superseded on master (focus-mode `--s`, productivity-breakdown, evaluation-sheet, create-task-placeholder, add-task-bar-actions, select-task-minimal, dialog-restore-point, plugin-index, dialog-logs – partly via refactors, e.g. dialog-restore-point's warning/error rules moved to shared primitives). Resolving the conflicts by keeping this branch's side would revert that newer work. After the rebase, the one real remaining fix is:
  - **`src/styles/page.scss:244`** – `border-left: 4px solid rgba(var(--c-accent), 1)` is invalid (`--c-accent` is a complete color), so blockquotes currently render without their accent border; your `var(--c-accent)` is correct and still missing on master. Keep this one.
  - Optionally also keep the inert tidy-ups: the `impact-stars` fallback removal, the `tag-edit` comment cleanup, and the `_css-variables.scss` blank-line removal.

### Should fix before merge

- **`src/app/features/archive/dialog-archive-compression/dialog-archive-compression.component.scss:44`** – `--divider` is the base primitive; most bundled themes override only the semantic `--divider-color` (12 themes vs 2), so this border would ignore them. Master has since changed this line to `var(--extra-border-color)`, so on rebase drop this hunk as well.

---

PR #9429, opened 2026-08-03
feat(outlook-tasks): add Outlook Tasks issue provider via Microsoft Graph API

### LINKED ISSUE #6984: 💡 Sync with Microsoft

starter727:
## Summary

Adds Outlook Tasks as a built-in issue provider, allowing users to sync their Microsoft To Do / Outlook tasks with Super Productivity via the Microsoft Graph API.

Closes #6984

## What's included

- **OAuth 2.0 authorization** — full authorization code flow with CSRF state validation, token auto-refresh with in-memory cache and deduplication
- **Task operations** — fetch task lists, search tasks, update tasks (status, title, notes) via `/me/todo` Graph API endpoints
- **Bidirectional sync adapter** — syncs isDone/status, title, and notes with configurable direction (pullOnly/pushAndPull/off)
- **Configuration form** — clientId, tenantId, taskListId selection, polling interval, two-way sync direction
- **Platform support** — both Electron and web OAuth callback paths
- **Unit tests** — client service, sync adapter, and enabled-check utility

## How it works

1. User registers an Azure AD app (clientId) and configures the provider
2. OAuth flow authorizes SP to access the user's Microsoft To Do data
3. Tasks are pulled from the selected task list and synced bidirectionally
4. Works on all platforms (Windows, macOS, Linux) — no local Outlook installation needed

## Notes

- The `categories` field is displayed in the issue content panel but category-based filtering is not yet implemented (can be added as a follow-up)
- `issueLink` returns empty since Graph API doesn't expose stable web URLs for individual tasks


johannesjo:
Thank you very much for this! New issue providers should be added as plugins rather than app code. Could you by any chance rework this PR to a plugin? If you are wondering how to best do it, the github issue provider plugin might be useful.

johannesjo:

Thanks for the PR – and for starting the plugin rework johannesjo asked for! The plugin under `packages/plugin-dev/outlook-tasks-provider/` is the right shape (the plugin API supports everything it uses, including `oauthButton` + PKCE via the host's plugin OAuth service), but the branch currently ships **both** the new plugin and the original built-in implementation, so the rework needs finishing before the rest can be judged.

I've also flagged a question for the maintainer about the hardcoded Azure client ID ownership; you may want to wait for that before finalizing the OAuth part.

### Blocking

- **Two complete implementations of the same provider ship in this PR.** The `temp` commit added the plugin but kept the entire built-in provider. As-is, the setup overview renders two identical "Outlook Tasks" tiles (the hardcoded one in `issue-provider-setup-overview.component.html` plus the `pluginProviders` loop entry), users can create providers of either kind, and `OUTLOOK_TASKS` becomes a permanent `BuiltInIssueProviderKey` in persisted models – exactly the surface the maintainer asked to avoid. Please remove the built-in half: `src/app/features/issue/providers/outlook-tasks/**` and its registrations in `issue.const.ts`, `issue.model.ts`, `issue.service.ts`, `issue-content-configs.const.ts`, `is-issue-done.ts`, `get-issue-provider-tooltip.ts`, `issue-provider-setup-overview.component.html`, `dialog-edit-issue-provider.component.{ts,html}`, `issue-two-way-sync.effects.ts`, `oauth-callback-handler.service.ts`, `dialog-get-and-enter-auth-code.component.ts`, plus the `OUTLOOK_TASKS` keys in `t.const.ts` / `en.json` – keeping the plugin and its build wiring.

### Should fix before merge

- **`packages/plugin-dev/outlook-tasks-provider/src/plugin.ts` (configFields)** – the `syncDirection` select is dead config: neither the plugin nor the host ever reads a `syncDirection` key. The host auto-generates the real per-field controls from `fieldMappings` (stored as `pluginConfig.twoWaySync.<taskField>`, consumed in `plugin-issue-provider-adapter.service.ts` / `plugin-sync-adapter.service.ts`), so a user setting your select to "Off" still gets `pullOnly` on isDone/title. Delete the field and its `CFG.SYNC_DIRECTION*` i18n keys.
- **`plugin.ts` (`mapTaskToIssue`, `getIssueLink`)** – the `https://to.do.microsoft.com/tasks/<id>` URLs are fabricated: your own built-in twin returns `''` with the comment that Graph exposes no stable web URL per task, and the PR body says the same. Since `issueDisplay` renders the title as a link via `linkField: 'url'`, every imported task gets a dead link as its primary click target. Omit `url` and return `''` from `getIssueLink`.
- **`plugin.ts` (`searchIssues`)** – the search term is quote-escaped but never URL-encoded before interpolation into the query string. A term containing `&` splits the `$filter` param (Graph returns 400) and `#` silently truncates the URL as a fragment. Use `encodeURIComponent(escapeODataString(searchTerm))` like the github-issue-provider does.
- **`plugin.ts` (`mapTaskToSearchResult`)** – mapping the date-only Graph `dueDateTime` to `dueWithTime` makes every imported due-dated task land scheduled at 00:00 instead of due on a day (the host creates `dueWithTime` whenever that field is set). The built-in twin maps the same field to `deadlineWithTime`, so the PR's two halves also disagree about semantics. Don't emit `dueWithTime` for date-only values; leave day-due handling to a `dueDay` field mapping.

### Nits

- **`plugin.ts`** – the `(config.taskListId as string) || 'defaultTasks'` fallback is a made-up sentinel: Graph has no list id `defaultTasks` (the well-known name is `wellknownListName eq 'defaultList'`), so if the fallback ever fires the request 404s. Since `taskListId` is required and the select is disabled until `loadOptions` runs, it's near-unreachable – drop the fallback and the placeholder option, or resolve the real default list.
- **`package-lock.json`** – the single-line `"dev": true` change on the `fsevents` entry is unrelated lockfile churn from a different npm version; please revert it.

### Questions

- Was the plugin flow tested against the live Graph API? `plugin.ts` uses `$filter=contains(title,'...')` and `$orderby=lastModifiedDateTime desc` on `/me/todo/lists/{id}/tasks`, while your built-in twin deliberately avoids both (client-side title filtering, no orderby) – which suggests Graph may reject those query options with a 400 and break search/backlog polling in the plugin. If you've seen them work live, say so and I'll drop this.
- Which redirect URIs are registered for the shipped client ID, and who maintains that Azure app registration? The host's web build uses its own SPA callback page (the `redirectUri` override in `oauthConfig` is desktop-only and stripped elsewhere), and Azure only serves CORS for token redemption on redirect URIs registered under the SPA platform – the comment in `plugin.ts` only instructs registering `http://localhost:51234` under "Mobile and desktop". Does the OAuth flow work on the web build with this registration?

---

johannesjo:
On the OAuth client ID: **the app registration needs to be project-owned before this can ship.**

To be clear about what I'm *not* asking for — committing the client ID is fine. We already do exactly that for Google Calendar (`packages/plugin-dev/google-calendar-provider/src/plugin.ts`), including a client secret, with the same RFC 8252 reasoning your comment gives. Public desktop clients can't keep a secret, PKCE and the redirect-URI allowlist are the real mechanisms, and your comment is correct on all of that. So please don't strip the value out or move it into user config.

The problem is the tenant it lives in. The registration is currently in your Azure AD, which means:

- if the account goes away, or the registration is deleted or its consent revoked, **every user's Outlook integration breaks at once** and I can't fix it without shipping a new build;
- the redirect-URI list, supported account types, and any future permission scopes are editable by someone outside the project;
- usage and consent telemetry for all our users flows to an account we don't control.

None of that is a comment on you personally — it's the same reason the Google client sits in the project's account rather than mine or a contributor's.

So: I'll create an Azure AD app registration under the project and hand you the client ID to swap in. Give me a bit to sort that out. What I need from you to set it up correctly:

1. the exact redirect URIs the flow uses (the plugin comment mentions `http://localhost:51234` under *Mobile and desktop applications* — is that the complete list, and does the web build need an SPA entry?),
2. the full set of Graph scopes you're requesting,
3. whether multi-tenant + personal Microsoft accounts is the account type you actually want, or work/school only.

This isn't blocking anything right now — the plugin-only rework is still the thing to finish first, and the two Graph questions from my review (the `$filter=contains(...)` / `$orderby` usage, and whether the web build's OAuth path works) are still open. Just flagging it early so it doesn't surface at the end.

Issue #9405, opened 2026-07-31
unresolvable sync conflict in webDAV-sync💥


rhabarberbarberah:
Hey there, I am using super-productivity for a while now -- also using the webDAV-sync to sync over my nextcloud to my android-device.
In the android-app I recently encountered a sync conflict, that does not seem to be solvable in the usual way.
Here's the error-message:

<img width="1080" height="334" alt="Image" src="https://github.com/user-attachments/assets/8dfb5e9c-b4a8-42e2-b295-1eda72de49a1" />

unfortunately I can't give direct instructions on how to reproduce the issue, since I've not been 100% able to figure out, which task is actually causing the conflict.

The major issue is, that I can't use the in-app-tools ("review sync conflicts") to resolve it, since the conflict-list remains empty.

From the message I assumed, that it is tied to a recurring task (which I discovered & set up recently)

I am thankful for any help in that context. either a workaround I can use to get the sync back online, or an actual fix for the issue.

### URL
https://localhost/#/tag/TODAY/tasks

### Meta Info
META: SP18.15.1_L2AF __ Browser – de-DE – Linux aarch64 – de-DE – UA:Mozilla/5.0 (Linux; Android 17; Pixel 8 Build/CP2A.260705.006; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/150.0.7871.181 Mobile Safari/537.36

### Actions Before Error
```

```

johannesjo:
Thanks for the detailed report — I traced this message to a deliberate sync safety stop. The app encountered a multi-item operation that this version cannot safely resolve piece by piece, so it stops before acknowledging the remote operation or rejecting the pending local operation. That is why the same error returns on every retry.

The empty “Review sync conflicts” page is expected in this case: it only contains conflicts that were successfully auto-resolved, while this one is blocked during the safety preflight.

For now, please take these steps:

1. Put every device connected to this sync folder offline and avoid further edits if possible.
2. Create a separate full data export from every device.
3. Copy the complete Nextcloud/WebDAV sync directory to a separate safe location.
4. Keep the affected app/profile installed. Please do not clear app data, reinstall, downgrade, or delete the recurring task/configuration as a workaround.
5. Keep the exports private. They are plaintext and may contain task content or provider credentials, so please do not attach exports, raw sync files, or full logs publicly.

The safest recovery is to keep those copies and wait for a fixed build, then update all devices before reconnecting them one at a time.

[...]

PR #9442, opened 2026-08-04
fix(sync): resolve plan-for-today bulk conflicts instead of wedging sync

### LINKED ISSUE #9426: 🚨 Single unsafe multi-entity conflict blocks entire Android sync batch (384 conflicts stuck)
### Version Used

18.16.0W (Windows desktop) / 18.16.0_L2A (Android)

### Current Behavior

After a few days without opening the Android client, while Windows/Linux had continued normal daily use (moving overdue tasks forward, "end of day" archiving, checking off tasks), opening Android and syncing triggers:

```
RemoteOpsProcessingService: Detected 384 conflicts. Auto-resolving with LWW.
ConflictResolutionService: Auto-resolving 384 conflict(s) using LWW
Error: ConflictResolutionService: Cannot safely auto-resolve local multi-entity operation for TASK:rpt_-uuoLWMCwLcPwWntDuR67_2026-07-28
```

Full conflict detail from the log (payload truncated by the app's own logger with a `short:` prefix, cut off mid-string — noting this as a secondary diagnosability gap):
```
{"conflicts":[{"entityType":"TASK","entityId":"rpt_-uuoLWMCwLcPwWntDuR67_2026-07-28","localOpIds":["019fb43f-3c92-748d-ba68-dcfcfe26fbbb","019fb43f-3c9d-7a38-890b-fa00e22ca05d","019fb43f-575a-76a5-9c32-0248d038cef5"],"remoteOpIds":["019fa78d-74d9-79c9-96de-c7c7cbcdebdb"],"suggestedResolution":"local"},{"entityType":"TASK_REPEAT_CFG","entityId":"-uuoLWMCwLcPwWntDuR67","localOpIds":["019fb43f-3c98-7... (truncated)
```

So: one of the local operations touches both the TASK and its TASK_REPEAT_CFG atomically (a linked/multi-entity op), and the resolver correctly refuses to split it across a per-entity LWW decision — that refusal itself seems like reasonable caution. The problem is what happens next: **all 384 conflicts appear to be held back by this single unsafe pair**, not just the one entity. Sync retried automatically ~10 seconds later (visible again at a second identical log occurrence) and hit the exact same wall, with no progress and no user-facing option. End result: the Android client stayed with roughly 60+ overdue tasks unresolved, completely diverged from the Windows/Linux state, with no way forward other than waiting.

For comparison, the Windows client's own log from the same session shows a completely clean state throughout (`status=IN_SYNC`, 0 new operations, stable `lastServerSeq`) — confirming the block is entirely local to Android and the conflicting operations never even reach the point of being safely reconciled.

This looks related to the multi-entity handling work referenced in #8944 (disjoint-merge attribution bug) and the "Hardened task, project, archive, and multi-entity conflict handling" changes in recent releases — possibly the same underlying mechanism, but a different failure mode (hard batch-wide block vs. silent misattribution).

**Suggested improvement**
1. Apply the safely-resolvable conflicts in a batch even when one specific entity pair must be held back; don't let one unsafe pair block the other 383.
2. For the pair that can't be safely auto-resolved, surface an explicit choice to the user (e.g. keep local / take remote / view both and decide) instead of silently retrying the same failure.
3. Avoid truncating conflict diagnostic payloads in the log (or truncate at a much higher limit) — the current `short:` truncation cuts off exactly the information needed to fully understand the second entity's conflicting operations.

### Expected Behavior

When a batch of conflicts contains one entity pair that can't be safely auto-resolved, the other resolvable conflicts in the same batch should still be applied. Sync should make partial forward progress instead of being fully blocked by a single unsafe pair, and retrying should not repeat the exact same failure indefinitely without any user-facing fallback.

Image
[sp-backup_2026-08-03_105010 (Windows_just_after_sort_and_synchro).json](https://github.com/user-attachments/files/30655939/sp-backup_2026-08-03_105010.Windows_just_after_sort_and_synchro.json)
[sp-backup_2026-08-03_110021 (Android).json](https://github.com/user-attachments/files/30655941/sp-backup_2026-08-03_110021.Android.json)
[SP-logs (Android).json](https://github.com/user-attachments/files/30655940/SP-logs.Android.json)
[SP-logs_Windows_20260803_1046 (Windows OK, just after sort and synchro).json](https://github.com/user-attachments/files/30655942/SP-logs_Windows_20260803_1046.Windows.OK.just.after.sort.and.synchro.json)
[SP-logs_Windows_20260803_1046 (Windows OK, just after sort and synchro).log](https://github.com/user-attachments/files/30655943/SP-logs_Windows_20260803_1046.Windows.OK.just.after.sort.and.synchro.log)

### Steps to Reproduce

Not reliably reproducible on demand. Pattern: let a repeating task's daily instances and its repeat-config accumulate several days of independent local changes on two unsynced devices (PC continuously used and synced; Android left unopened for a few days), then open/sync Android.

Sometimes, it's intermittent.

### Console Output

```shell
See full excerpt above. Both the Windows log (clean throughout) and the Android log (showing the repeated failure) are attached, along with a screenshot of the error as it appeared on screen and JSON data exports from both devices at the time of the incident.
```

### Documentation

johannesjo:
## Problem

`_assertMultiEntityPlansAreSafe` is a deliberate fail-closed preflight (#8980, shipped since v18.15.0): it runs over ALL planned conflict resolutions before any mutation and throws for any multi-entity op that is neither archive-win, independent-multi-delete, nor decomposable. One such op in one conflict therefore aborts the entire batch, the server cursor never advances, and sync fails identically on every retry with no in-app recovery ("Review sync conflicts" stays empty because the stop is pre-resolution).

The dominant real-world trigger is `planTasksForToday`: the automatic day-rollover (`TaskDueEffects` -> `addAllDueToday()`) emits ONE op carrying EVERY due task id, so a single concurrent remote edit of any one of those tasks wedges the client. Confirmed twice in the field: the op-log extraction in #9405 (blocked action `TASK_SHARED_PLAN_FOR_TODAY`, entity counts 32/47/55/61) and the timing analysis of the logs attached to #9426. It compounds: a wedged client falls through `afterInitialSyncDoneStrict$` 8s after the failed sync and mints a NEW bulk op every day, growing the blocked set. The same rollover also dispatches `planDeadlineTasksForToday`, and every Today drag (`moveTaskInTodayTagList`, always 2 ids) and bulk unplan (`removeTasksFromTodayTag`) reach the same guard; #9412 pinned all of these as reachable.

#9412 (merged, unreleased) improved the diagnostics only and explicitly deferred this fix.

## Why the fix is resolution-side

A capture-side change (splitting the bulk op at dispatch) would not heal the clients wedged in the field today: their bulk rows are already captured and pending. This fix resolves at conflict-resolution time, so wedged clients heal on their first sync retry after updating. It also has to cover the REMOTE side in the same release: a healed client uploads a scoped replacement that is itself a multi-entity op, and a receiver with any pending edit on a retained id would otherwise hit the remote branch of the same guard, transplanting the wedge from device A to device B.

## What the fix does

1. **`planTasksForToday` / `planDeadlineTasksForToday` rows (local): scoped replacement.** The atomic row is rejected as a unit (all local ops of a conflict always are), and ONE narrowed copy re-emits the surviving siblings' intent, mirroring the existing `_preservePartiallyRejectedLocalBulkDeletes` mechanism. The mechanism lives in a new pure util (`preserve-partial-bulk-plan.util.ts`) rather than growing the grandfathered 4.2k-line service. Conflict targets are dropped from the copy when they are covered (remote win: the newer remote op owns the task; local win with a whole-state snapshot op). An uncovered local-win target (entity absent from the live store, so no snapshot could be built) is retained: the reducers skip unknown task ids, so retention is at worst a replay no-op and can never silently drop plan intent.
2. **Ordering-only rows (`moveTaskInTodayTagList`, `removeTasksFromTodayTag`): plain rejection.** Their reducers touch only `TODAY_TAG.taskIds` ordering and the task adapter's `ids` order; Today MEMBERSHIP is derived from `task.dueDay`/`dueWithTime` (ADR #2). New invariant specs pin the entity-field-free property so a future reducer change cannot silently invalidate the allowlist. Honest caveat: for `removeTasksFromTodayTag` a rejected row can leave a stale legacy list entry visible on other clients until their own rollover cleans it up. Bounded and self-healing, versus a full sync freeze.
3. **Remote rows of these types: atomic replay via the existing mixed-winner machinery** (apply the original op once, then local-win snapshots as compensations after it, in durable seq order). A local winner without a compensation snapshot degrades to a logged remote win for these types instead of throwing `Cannot safely compensate mixed multi-entity winners`; that shape is reachable only when the winning entity is absent from the live store, where the replay skips it anyway.
4. **`_convertToLWWUpdatesIfNeeded` no longer converts multi-entity remote ops.** Rewriting a bulk row into a single-entity LWW replace would drop its effect on every other entity. Today that mangling is masked downstream by two accidents (the recreate path re-routes the original op, and the same-id converted copy dies in the append duplicate check); the guard makes the invariant explicit. Consequence for multi-entity delete-vs-update: the locally deleted entity stays deleted (the same logged degrade the codebase already chose for legacy bulk deletes) while the row still applies to its other entities.

**What stays fail-closed, deliberately:** `updateTasks` (two producers with opposite entity locations, archive-resident targets, coupled projectId moves, per the #9412 findings), `updateTask` with `projectMoveSubTaskIds` (coupled structural move), and legacy `addTagToTask`. These keep the typed `SYNC_MULTI_ENTITY_UNSUPPORTED` diagnostic; the pinning spec now covers exactly that remaining set.

## Fleet / compatibility (rule 10)

No schema bump, no new wire shapes, no envelope changes. A narrowed plan op is a payload shape every released client already produces and replays natively (single-task `planTasksForToday` ships via `TaskService.addToToday` and the task context menu). Replacement clocks dominate every conflict clock involving the original row; per-client monotonic ordering of multiple replacements is guaranteed by the existing local-append clock rebase (#8939) and is now pinned by a test that fails if that rebase is ever bypassed.

## Verification

- New integration spec (`today-plan-conflict-resolution.integration.spec.ts`, 10 cases) drives the REAL capture -> persistence -> conflict-detection -> resolution path (real op-log store, real effects): remote-win scoping, local-win exclusion, uncovered-target retention, deadline variant, multi-day compounding with ordered written clocks, ordering-only rejection for both actions, remote bulk apply-once + compensation ordering, mixed-winner degrade, and the delete-vs-bulk no-mangling contract. All 10 failed with the exact production error before the fix.
- Sabotage-verified: allowlist off (10/10 fail), retained-ids filter off (4 fail), unconditional mixed-winner throw restored (exactly the degrade case fails), store clock rebase bypassed (exactly the compounding case fails). Two sabotages initially SURVIVED and were investigated rather than papered over: builder-side clock chaining turned out to be redundant (the store rebase is the real guarantor, so the chaining was removed and the store property pinned instead), and the conversion guard is currently masked downstream (kept, with the masking documented in the comment).
- Full `src/app/op-log/**` spec tree green (3772 specs), plus the touched reducer suites; `checkFile` clean on every modified file.

## Review notes

The plan behind this PR was reviewed by two independent adversarial passes before implementation; their findings shaped the final shape (deadline-op coverage, conditional target exclusion, the degrade path, util placement, the invariant specs). Two suggested amendments were deliberately NOT taken:

- Extracting a shared payload-narrowing helper out of the proven `_createScopedBulkDeleteReplacement`: declined to keep the shipped delete path untouched in this PR (DRY at two occurrences did not outweigh churn on hardened sync code).
- Skipping replacements for stale-day plan rows: declined because the replacement propagates this device's CURRENT task state; dropping it would reintroduce exactly the silent divergence the mechanism exists to prevent, and would make resolution outcomes wall-clock-dependent.

## Field guidance

Wedged clients (v18.15.0 to v18.16.0) heal automatically on their first successful sync attempt after updating; no migration or manual recovery needed. The re-baseline workaround from #9405 remains valid for anyone who cannot update.

Fixes #9426. Refs #9405 (the blocked action extracted there is exactly the one this makes resolvable; the two known non-Today shapes stay fail-closed with the #9412 diagnostic).

### DIFF
--- a/src/app/features/tasks/store/task.reducer.spec.ts
+++ b/src/app/features/tasks/store/task.reducer.spec.ts
@@ -631,6 +631,19 @@ describe('Task Reducer', () => {
       expect(state.entities).toBe(stateWithOrderedTasks.entities);
     });
 
+    it('must not handle moveTaskInTodayTagList at all (ordering-only invariant #9426)', () => {
+      // The task feature reducer currently has NO handler for this action; a
+      // future one that touches entities would invalidate the ordering-only
+      // rejection in conflict resolution. Same remediation as above.
+      const action = TaskSharedActions.moveTaskInTodayTagList({
+        toTaskId: 'task1',
+        fromTaskId: 'task2',
+      });
+      const state = taskReducer(stateWithTasks, action);
+
+      expect(state).toBe(stateWithTasks);
+    });
+
     it('should ignore all invalid IDs and leave state unchanged', () => {
       stubWindowConfirm(false);
       if (!jasmine.isSpy(window.alert)) {
--- a/src/app/features/tasks/store/task.reducer.spec.ts
+++ b/src/app/features/tasks/store/task.reducer.spec.ts
@@ -623,6 +623,12 @@ describe('Task Reducer', () => {
 
       // The removed tasks should be moved to the beginning while maintaining their relative order
       expect(state.ids).toEqual(['task2', 'task4', 'task1', 'task3']);
+      // Ordering-only invariant (#9426): conflict resolution rejects
+      // conflicted rows of this action outright, which is lossless only while
+      // the handler never touches task entities. If this fails, remove the
+      // action from ORDERING_ONLY_MULTI_ACTIONS in conflict-resolution.service.ts
+      // (or give it a preserve path) BEFORE shipping the reducer change.
+      expect(state.entities).toBe(stateWithOrderedTasks.entities);
     });
 
     it('should ignore all invalid IDs and leave state unchanged', () => {
--- a/src/app/op-log/sync/conflict-disjoint-merge.util.ts
+++ b/src/app/op-log/sync/conflict-disjoint-merge.util.ts
@@ -185,7 +185,12 @@ export const noiseTiebreakSide = (
  *
  * Field-level conditions only (the caller separately excludes archive plans):
  *  - neither side contains a multi-entity op, because resolving one conflicted
- *    entity would reject the whole original op and drop its sibling updates;
+ *    entity would reject the whole original op and drop its sibling updates.
+ *    LOAD-BEARING beyond that rationale since #9426: for SCOPED_PLAN types the
+ *    sibling loss is now handled, but `_preservePartiallyRejectedLocalBulkPlanOps`
+ *    only sees conflicts routed to the plain-LWW `resolutions` list — a merged
+ *    conflict bypasses it and would starve the scoped replacement. Do not relax
+ *    this condition for those types without moving that grouping too;
  *  - neither side has a DELETE op;
  *  - BOTH sides changed at least one real (non-noise) field — if one side only
  *    bumped noise, nothing real is lost by LWW, so leave it to SPAP-13's `noise`
--- a/src/app/op-log/sync/conflict-resolution.service.spec.ts
+++ b/src/app/op-log/sync/conflict-resolution.service.spec.ts
@@ -3676,6 +3676,64 @@ describe('ConflictResolutionService', () => {
         expect(mockOperationApplier.applyOperations).not.toHaveBeenCalled();
       });
 
+      it('keeps a locally deleted task deleted when a remote bulk plan op wins (#9426)', async () => {
+        // Pins the multi-entity skip in _convertToLWWUpdatesIfNeeded: with it,
+        // delete-vs-bulk-plan takes the documented stays-deleted degrade (no
+        // recreation row); without it, the conversion feeds the recreate path
+        // and a fresh recreatesEntityAfterDelete row for the deleted task
+        // rides the LOCAL batch. The bulk op itself must stay unmangled either
+        // way so its sibling still gets planned.
+        const remotePlanOp: Operation = {
+          ...createOpWithTimestamp(
+            'remote-bulk-plan',
+            'client-b',
+            2000,
+            OpType.Update,
+            'task-1',
+          ),
+          actionType: ActionType.TASK_SHARED_PLAN_FOR_TODAY,
+          entityIds: ['task-1', 'task-2'],
+          payload: {
+            actionPayload: { taskIds: ['task-1', 'task-2'], today: '2026-07-30' },
+            entityChanges: [],
+          },
+        };
+        const localDelete: Operation = {
+          ...createOpWithTimestamp(
+            'local-delete',
+            'client-a',
+            1000,
+            OpType.Delete,
+            'task-1',
+          ),
+          actionType: ActionType.TASK_SHARED_DELETE,
+          payload: {
+            task: { id: 'task-1', title: 'Deleted locally' },
+          },
+        };
+
+        await service.autoResolveConflictsLWW([
+          createConflict('task-1', [localDelete], [remotePlanOp]),
+        ]);
+
+        // The original atomic row is queued unmangled on the remote side.
+        // With no local-win/recreation rows, it takes the plain remote append;
+        // without the conversion guard it instead rides the mixed batch with a
+        // same-id LOCAL rewrite row alongside it.
+        const plainRemoteAppends = mockOpLogStore.appendBatchSkipDuplicates.calls
+          .allArgs()
+          .flatMap(([ops]) => ops as Operation[]);
+        const appendedCopies = [...plainRemoteAppends, ...getMixedRemoteOps()].filter(
+          ({ id }) => id === remotePlanOp.id,
+        );
+        expect(appendedCopies.length).toBe(1);
+        expect(appendedCopies[0].actionType).toBe(ActionType.TASK_SHARED_PLAN_FOR_TODAY);
+        expect(appendedCopies[0].entityIds).toEqual(['task-1', 'task-2']);
+        // Stays-deleted degrade: NO local-source row touches the deleted task
+        // (a recreation row here means the conversion guard was bypassed).
+        expect(getMixedLocalOps().filter((op) => op.entityId === 'task-1')).toEqual([]);
+      });
+
       it('should extract entity from DELETE payload when UPDATE wins but entity not in store', () => {
         // This tests the helper method that extracts entity state from DELETE operations
         // Used when remote DELETE is applied first, then local UPDATE wins LWW
--- a/src/app/op-log/sync/conflict-resolution.service.ts
+++ b/src/app/op-log/sync/conflict-resolution.service.ts
@@ -379,20 +379,26 @@ const INDEPENDENT_MULTI_DELETE_ACTIONS = new Set<ActionType>([
  * - ORDERING_ONLY rows touch nothing but `TODAY_TAG.taskIds` ordering — Today
  *   MEMBERSHIP is derived from `task.dueDay`/`dueWithTime` (ADR #2) — so
  *   rejecting such a row loses at most list ordering: cosmetic, self-healing,
- *   no replacement needed. Pinned by the entity-field-free invariant spec in
- *   `task-shared-scheduling.reducer.spec.ts`.
+ *   no replacement needed. Pinned by the entity-field-free invariant specs in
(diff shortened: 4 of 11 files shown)


johannesjo:
## Multi-agent review round (7 reviewers) and hardening commit

This PR went through a seven-reviewer pass (six Claude agents: correctness, security, architecture, alternatives, performance, simplicity, plus Codex CLI), each independently verifying against the source. Every finding was re-verified against the code before adoption. Outcome: **no verified criticals**; the hardening landed as `2b8f4b5b9b`.

**Applied from the reviews:**
- The scoped-plan builder now validates its payload contract (`actionPayload.taskIds` must be an array); a violating row degrades to logged plain rejection instead of silently replaying the un-narrowed id set. The contract is documented on `SCOPED_PLAN_MULTI_ACTIONS` so future additions cannot miss it.
- The conversion guard's chosen semantics are now pinned by a unit test that fails without it: a remote bulk plan op winning over a local DELETE keeps the task deleted (no recreation row) while the bulk row applies unmangled. The integration spec cannot observe this (real-store append dedup masks the difference), so the pin reads the mocked batch args. Verified by sabotage in both directions.
- New co-located unit spec for the builder (parentTaskMap narrowing, contract skip, covered vs uncovered targets, deterministic ordering, clock domination), the missing `task.reducer` pin for `moveTaskInTodayTagList`, sync-core's `LwwResolvedConflict` instead of a structural re-declare, a deterministic op-id tie-break for equal-timestamp rows, logging for the uncovered-retention branch, and documentation for the disjoint-merge dependency, the planner-day restore gap, and the crash window inherited from the delete twin.

**Codex requested changes with four "criticals"; all four were downgraded on verification:**
1. "Remote-win targets keep the rejected row's optimistic side effects": pre-existing whole-entity-LWW loser semantics (a winner whose payload does not overlap the loser's fields leaves those values live locally until superseded); the SPAP-13 journal exists to surface exactly this, and the pre-fix behavior for this shape was a total wedge.
2. "Compensation cannot restore planner-day placement": real but bounded; `planner.days` is authoritative for future-day columns, yet the same cross-slice property has always held for archive rows in the mixed-winner path. Now documented in code; the entry reappears when the task's own due day arrives.
3. "planDeadlineTasksForToday is not per-task scopeable (parent/child coupling)": the coupling is the inherent state-dependence of action-replay ops; the un-narrowed row replayed on a peer has it too, and narrowing evaluates the child against the RESOLVED parent state, which is the correct post-conflict input. Each client's own rollover re-derives the outcome daily. Documented on the set.
4. "Wall-clock replacement ordering": known corner, fixed with the op-id tie-break; written-row ordering was already guaranteed by the local-append clock rebase (#8939) and is pinned by the multi-day integration case.

**Known bounded inaccuracy, documented in code:** after the mixed-winner degrade, a SPAP-13 journal entry would record the pre-degrade winner. Confined to journal-enabled builds; the production entry point passes `disableConflictJournal: true` (the #9061 producer freeze). Revisit when the producer is unfrozen.

**Follow-up worth filing separately (pre-existing hardening class, not introduced here):** extend the GHSA-8pxh interim integrity binding (exact-set equality of authenticated payload `taskIds` against the plaintext `op.entityIds` envelope) to the four Today-list action types, mirroring the project-move footprint check.

Suite state after hardening: 3,780 op-log specs green (8 new this round), touched reducer suites green, checkFile clean on every file.

johannesjo:
Follow-through on the security reviewer's item: the GHSA-8pxh footprint binding for the four Today-list action types now ships IN this PR (`0c1e9921ca`) instead of as a separate follow-up. `assertEncryptedTodayListFootprintIntegrity` mirrors the project-move footprint check: exact-set equality between the plaintext envelope ids and the authenticated payload footprint (`taskIds`, or the `toTaskId`/`fromTaskId` pair), failing closed. This closes both flagged vectors: envelope inflation with victim ids and envelope stripping (the undercount bypass of the conversion guard). Fleet-safe by verified history: all four actions have captured `entityIds` verbatim from the payload ids since the first persistent-meta commit (`a1f7284b02`), and unknown/legacy payload shapes keep the same interim tolerance as the project-move binding. 8 new integrity spec cases, sabotage-verified (binding disabled fails exactly the 4 tampered cases).

PR #9458, opened 2026-08-05
fix(color): render the color picker panel in a CDK overlay

### LINKED ISSUE #9423: 💥 Tag color picker off screen
### Steps to Reproduce

1. create tag
2. click color to open color picker
3. color picker is half offscreen

### Additional Console Output

When in fullscreen or small window, the color picker only shows the top half of the color picker. When in windowed screen, once the color picker is on screen, you can extend the window to reveal the bottom half.

Mac Ventura 13.7.8 (22H730)
Super Productivity Version 18.16.0 (18.16.0)

### URL
file:///Applications/Super%20Productivity.app/Contents/Resources/app.asar/.tmp/angular-dist/browser/index.html#/project/INBOX_PROJECT/tasks

### Meta Info
META: SP18.16.0D __ Electron – en-US – MacIntel – en-US – UA:Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) superProductivity/18.16.0 Chrome/146.0.7680.216 Electron/41.4.0 Safari/537.36

### Actions Before Error
```

```

MannXo:
## Problem

Fixes #9423. The tag color picker panel renders half off-screen on macOS.

`InputColorPickerComponent` positioned its panel by hand: it read the trigger's `getBoundingClientRect()` (viewport coordinates) and applied them as `position: fixed; top/left` on a panel that lives inside the component's own DOM.

`backdrop-filter` makes an element the containing block for fixed-position descendants, and three built-in themes apply it to `.mat-mdc-dialog-surface`: `liquid-glass.css` (Tier B), plus `velvet.css` and `rainbow.css` in dark mode. Under those themes the panel's offsets resolve against the dialog surface instead of the viewport, so the panel is displaced by the dialog's own origin and falls off the viewport edge.

Liquid Glass is the auto-selected default on Apple Silicon Macs (`pickInitialActiveRef` in `custom-theme.service.ts`), which is why this is a macOS report and why it does not reproduce under the default theme.

Measured in the real Electron app, Create Tag dialog, Liquid Glass, 1200x400 content size:

| | before | after |
|---|---|---|
| trigger rect | `top 235.9, left 438` | `top 234.1, left 436.2` |
| panel inline style | `top: 71.9px; left: 438px` | none (overlay-positioned) |
| panel painted at | `top 115.4, left 852` | `top 75.9, left 438` |
| result | 414px right of the dialog, clipped | anchored above the trigger, fully on screen |

The reporter's "extend the window to reveal the bottom half" matches the same cause: the panel is placed past the viewport edge and never repositions.

| before | after |
|---|---|
| image | image |

## Solution

Move the panel into a CDK connected overlay anchored to the trigger. The overlay pane is a child of the overlay container, so no ancestor can become its containing block.

This removes rather than adds: `panelTop`, `panelLeft`, `_updatePanelPosition()`, the hardcoded `panelWidth = 196` / `panelHeight = 160` guesses (the panel actually measures 192x156), the hand-rolled `.backdrop` element and its SCSS, and the `document:keydown.escape` `HostListener`.

Four things the hand-rolled positioning could not do come along with it:

- viewport clamping via `withPush`, where the old flip-above branch could produce a negative `top` with no clamp
- reposition on scroll (CDK's default reposition strategy)
- outside-click and Escape handled by the overlay itself
- the app's global safe-area patch on `FlexibleConnectedPositionStrategy` (`cdk-safe-area-viewport.util.ts`), which the old code bypassed entirely

One deliberate behavior change: Escape now closes only the picker. Previously, the `document` listener closed the picker while the same event also reached `MatDialog` and closed the dialog. CDK's keyboard dispatcher routes the key to the top-most overlay, which matches how `mat-menu` and `mat-select` behave inside a dialog.

Panel styles stay in the component SCSS. Template-projected content keeps its `_ngcontent-*` attribute, so class selectors still apply once the panel is portaled; verified in the running app (`--card-bg`, padding, radius and elevation all resolve).

Both call sites are covered by the one change: `dialog-create-tag` and the formly `color` field used by the tag/project settings dialog.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

Nothing to document in the wiki: this is a positioning fix with no user-facing behavior to describe.

## Testing

Unit suite: 14029 passing (baseline on `master` was 14022; the 7 new specs are the difference). Lint clean, 0 errors, the 10 remaining warnings are pre-existing in untouched files.

New `input-color-picker.component.spec.ts` (7 specs). Two of them are the regression guard: the panel must render in the overlay container rather than inside the component, and it must stay anchored to the trigger and inside the viewport when mounted inside a wrapper that carries `backdrop-filter`. Sabotage-verified by restoring the old component: **2 FAILED, 5 SUCCESS**.

New `e2e/tests/tags/tag-color-picker-9423.spec.ts`. The unit spec can only simulate the containing block with a synthetic wrapper, so this one activates the real `liquid-glass.css`, opens the real Create Tag dialog, and asserts the panel is anchored and on screen. It asserts the theme is in effect first (`backdrop-filter !== 'none'` on the surface), so it cannot pass vacuously if the theme fails to load. Sabotage-verified against the old component: **1 failed**, on the anchoring assertion.

`e2e/tests/tags/tag-crud.spec.ts`: 5 passing, confirming the `TagPage.openCreateTagDialog()` extraction (pulled out of `createTag()`, which now calls it) did not change existing behavior.

Manual verification in the real Electron app on macOS, driven through Playwright's `_electron`, under both Liquid Glass and the default theme, in the Create Tag dialog and the tag settings dialog. Panel measures 192x156, sits 4px from the trigger, flips above when there is no room below, and stays inside the viewport in every case.

## Notes for reviewers

The theme side is intentional and I did not touch it. `tools/theme-assets.test.js` already guards `.main-content` against creating a containing block ("Liquid Glass keeps routed overlays and mobile safe-area math intact") and `liquid-glass.css` comments on the hazard directly, while Tier B applies the effect to dialog and menu surfaces on purpose. `magic-side-nav.component.scss` carries the same warning to theme authors. So the component was the layer relying on an assumption the themes are allowed to break.

Alternative considered and rejected: keep the hand-rolled `position: fixed` and subtract the nearest containing block's rect. It needs the panel measured before it is rendered, and it would still leave the missing viewport clamp, the missing reposition-on-scroll, and the bypassed safe-area handling.

No new dependencies (`@angular/cdk` is already a dependency), no translations, and nothing in the sync, op-log or persisted-model surface is touched.

### DIFF
--- a/e2e/pages/tag.page.ts
+++ b/e2e/pages/tag.page.ts
@@ -18,9 +18,9 @@ export class TagPage extends BasePage {
   }
 
   /**
-   * Creates a new tag via the sidebar
+   * Opens the create tag dialog via the sidebar and returns its name input.
    */
-  async createTag(tagName: string): Promise<void> {
+  async openCreateTagDialog(): Promise<Locator> {
     // Find the Tags group header button
     const tagsGroupBtn = this.tagsGroup
       .locator('.g-multi-btn-wrapper nav-item button')
@@ -56,6 +56,15 @@ export class TagPage extends BasePage {
 
     // Add a small delay for Angular form initialization
     await this.page.waitForTimeout(500);
+
+    return tagNameInput;
+  }
+
+  /**
+   * Creates a new tag via the sidebar
+   */
+  async createTag(tagName: string): Promise<void> {
+    const tagNameInput = await this.openCreateTagDialog();
 
     await tagNameInput.fill(tagName);
 
--- a/e2e/tests/tags/tag-color-picker-9423.spec.ts
+++ b/e2e/tests/tags/tag-color-picker-9423.spec.ts
@@ -1,0 +1,69 @@
+import { test, expect } from '../../fixtures/test.fixture';
+import { waitForAppReady } from '../../utils/waits';
+
+/**
+ * Issue #9423: the tag color picker rendered half off-screen on macOS.
+ *
+ * `backdrop-filter` makes an element the containing block for fixed-position
+ * descendants, and three built-in themes apply it to `.mat-mdc-dialog-surface`.
+ * The picker positioned its panel with `position: fixed` plus viewport
+ * coordinates read from the trigger, so under those themes the offsets resolved
+ * against the dialog surface instead of the viewport and the panel was
+ * displaced by the dialog's own origin.
+ *
+ * The unit spec pins the same behavior against a synthetic wrapper. Only this
+ * test runs the real `liquid-glass.css` against the real dialog, which is the
+ * exact combination the bug needed — so it asserts the theme is actually in
+ * effect before it asserts anything about the panel.
+ *
+ * Liquid Glass is the default on Apple Silicon Macs (`pickInitialActiveRef`),
+ * which is why the report came from macOS.
+ *
+ * Run: npm run e2e:file e2e/tests/tags/tag-color-picker-9423.spec.ts -- --retries=0
+ */
+test.describe('Tag color picker positioning', () => {
+  test('should keep the panel on screen and anchored under a glass-surface theme', async ({
+    page,
+    workViewPage,
+    tagPage,
+  }) => {
+    await workViewPage.waitForTaskList();
+
+    await page.evaluate(() =>
+      localStorage.setItem('CUSTOM_THEME', 'builtin:liquid-glass'),
+    );
+    await page.reload({ waitUntil: 'domcontentloaded' });
+    await waitForAppReady(page);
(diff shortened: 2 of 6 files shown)


johannesjo:

PR #9463, opened 2026-08-05
fix(schedule): span the whole month in the month grid

### LINKED ISSUE #9449: 🚨 Task scheduled for a month-end date that falls on the "wrong" weekday row (e.g. Mon 31 Aug) doesn't show up in Planner/Schedule
### Version Used

18.16.0 Desktop (Windows)

### Current Behavior

The task scheduled for 31 August 2026 does not appear in either the August or the September grouping of the Planner/Schedule view.

### Expected Behavior

A task scheduled for 31 August 2026 (a Monday) should be visible in the Planner and Schedule views, either under the August section or at the correct position in the week row.

### Steps to Reproduce

1. Create a task.
2. Schedule it for 31 August 2026 (a Monday), either via the Planner view or by typing `@31/8` in the task title.
3. Open the Planner and/or Schedule view.
4. Observe that the task/day is missing from both the August and September sections.

### Console Output

```shell

```

### Documentation

MannXo:
## Problem

Fixes #9449.

A task scheduled for Mon 31 Aug 2026 does not show up in the Schedule month view, and the day is not reachable in the Planner's calendar nav either.

August 2026 starts on a Saturday, so its padded grid needs six week rows. Two places assumed fewer:

**Schedule month view.** `_daysToShowCount` derived the number of week rows from the window height:

```ts
const availableHeight = height - MONTH_VIEW.HEADER_OFFSET;   // 160
const maxWeeks = Math.floor(availableHeight / minHeightPerWeek); // /100 on desktop
// clamped to MIN_WEEKS 3 .. MAX_WEEKS 6
```

That count went straight into `getMonthDaysToShow`, which emits exactly `numberOfWeeks * 7` days from the padded week start with no check that the month's last day is inside the window. Six rows needed a window at least 760px tall, so an ordinary desktop window produced five and the grid stopped at Aug 29/30. `daysToShow` is the only input to the event pipeline, so a day outside the window gets no events computed at all: the task is absent, not merely clipped.

The row tracks follow the rendered count on their own. Since #9623 `ScheduleMonthComponent` binds `[style.--nr-of-weeks]` to `weeksToShow()` on its own host, which shadows the static `6` on the schedule host, and `.month-grid-container` is `grid-template-rows: auto repeat(var(--nr-of-weeks), 1fr)` inside `height: calc(100vh - 120px)`. That is not the whole story, though, and an earlier revision of this description got it wrong. `1fr` is `minmax(auto, 1fr)` and `.month-day-cell` carries `min-height: 80px`, so once the rows stop fitting it is the track *minimum* that sets the grid height rather than the `fr` share, and the grid overflows. Measured at 1280x500: six rows overflow the wrapper by 121px, five by 40, four not at all, a step of exactly 81 (the cell minimum plus the 1px gap). What the old viewport-derived count did was not make cells shorter, it removed days outright.

**Planner calendar nav.** `WEEKS_SHOWN` was hard-coded to 5 while `_shiftToMonth` anchors to the week containing the 1st, so August's expanded grid ran Mon 27 Jul to Sun 30 Aug. Mon 31 Aug had no cell: no task dot, and no way to tap through to that day.

One correction to the report: the day *does* appear in the September grid on both surfaces, as a dimmed `other-month` padding cell, since September 2026 starts on a Tuesday and its grid begins Aug 30/31. So the "missing from September too" part is that dimmed cell not registering rather than a second defect.

## Solution

- Month view row count is derived from the displayed month (`ScheduleService.getMonthWeeksToShow`, `ceil((daysToGoBack + daysInMonth) / 7)`, so 4 to 6) instead of from the viewport. `SCHEDULE_CONSTANTS.MONTH_VIEW` is deleted with the old height math.

  **This is a layout change, and the trade-off is worth stating plainly.** Cells are `1fr` inside a fixed-height grid, so a month needing six rows gets shorter cells than one needing four, and on a short window the grid overflows and scrolls (which is what the scroll reset below is for). Measured at 1280x500 on September 2026, which derives five rows: 35 cells from `2026-08-31` to `2026-10-04`, `scrollHeight` 440 against `clientHeight` 400. What it buys is that no day of the month is ever left out of `daysToShow`, which is the bug: a day outside the window gets no events computed at all, so the task disappears rather than being clipped.
- Removed the now-unused `HEADER_OFFSET`, `MIN_HEIGHT_PER_WEEK_DESKTOP`, `MIN_HEIGHT_PER_WEEK_MOBILE`, `MIN_WEEKS` and `MAX_WEEKS` constants.
- `WEEKS_SHOWN` in the planner calendar nav goes from 5 to 6.

`getMonthDaysToShow` itself is unchanged: its `numberOfWeeks` parameter stays literal, and the caller owns the grid size.

An earlier revision of this PR pinned the count at six and recorded an objection to deriving it: that the row tracks were hard-coded and the `:host[style*='--nr-of-weeks: N']` density rules were dead. Both were true before #9623 and are not now. That commit binds the tracks to `weeksToShow()` and deletes the density rules, so the derived count needed no wiring up, and the pinned six is gone.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Verification

Checked against the running app (`ng serve`, 1280x700 viewport, which is short enough that the old code computed five rows), reading the `data-day` attributes off the rendered month grid:

| | cells | window | August days | has 2026-08-31 |
|---|---|---|---|---|
| before | 35 | 2026-07-27 to 2026-08-30 | 30 of 31 | no |
| after | 42 | 2026-07-27 to 2026-09-06 | 31 of 31 | yes |

I have before/after screenshots of the August 2026 grid (the before one shows the empty sixth row track under the 24-30 row). The template has no screenshot slot so I left them out; happy to post them in a comment.

New tests, each sabotage-verified:

- `schedule.service.spec.ts` — Mon 31 Aug 2026 is present for all seven first-day-of-week values; a 5-week window provably misses it; and every month/first-day-of-week combination across 24 consecutive months contains both the 1st and the last day. Pinning the derived count back to 6 fails 2 of 50.
- `schedule.service.spec.ts` — the derivation itself: 4 for a non-leap February starting on the week's first day, 6 for August 2026, 5 for September 2026; and across 24 months x 7 first-days, the count both reaches the last day of the month and would drop it with one row fewer, so it is pinned from both sides rather than restating the formula.
- `schedule.component.spec.ts` — the month row count is unchanged between a 1400px-tall and a 500px-tall window. Restoring the height-derived count fails it.
- `planner-calendar-nav.component.spec.ts` — August 2026's expanded grid is anchored at Mon 27 Jul and contains Mon 31 Aug; the collapsed strip keeps full-size rows on a viewport that shrinks the expanded grid; and an expanded grid whose box shrinks under it never exceeds the room left. Reverting the clamp fails 1 of 49; reverting the collapsed pin fails 2 of 49.
- `planner-calendar-gesture-handler.spec.ts` — with no room to expand, a drag does not open the grid at all and still writes a usable transform (the interpolation range is 0 there).
- `e2e/tests/schedule/schedule-month-scroll-reset.spec.ts` — entering month view at 1280x500 resets the wrapper. It asserts the grid really overflows first, so it cannot pass vacuously. Deleting the reset branch fails it with `Received: 40`.

One existing test needed repair rather than a behavior change: the expanded panel is taller now, so the snap midpoint moved, and `handle drag > should snap to expanded when dragging down past midpoint` used a literal 100px drag that landed exactly on the new midpoint (the comparison is strict `>`). It now derives its drag distance from the geometry; I confirmed it still fails when the drag is below the midpoint.

## Checklist

- [ ] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki) — no user-facing behavior change beyond the bug itself
- [x] Existing tests still pass — full suite 14645 green, 0 failing (14631 on the `America/Los_Angeles` run); schedule e2e 8/8, planner e2e 24/24

### DIFF
--- a/e2e/tests/schedule/schedule-month-scroll-reset.spec.ts
+++ b/e2e/tests/schedule/schedule-month-scroll-reset.spec.ts
@@ -1,0 +1,51 @@
+import { expect, test } from '../../fixtures/test.fixture';
+
+/**
+ * Review of #9463: with the month grid fixed at six rows it overflows the
+ * scroll wrapper on a short window, so the scroll position carried over from
+ * week view clamped to the bottom and the month opened with its first row
+ * above the viewport (scrollTop 121 of a 121px maximum, reproduced here).
+ *
+ * The unit spec asserts the reset call fires; only this test runs it at a real
+ * constrained height with a real overflowing month grid, which is the
+ * combination the bug needed.
+ *
+ * Run: npm run e2e:file e2e/tests/schedule/schedule-month-scroll-reset.spec.ts -- --retries=0
+ */
+const SHORT_DESKTOP_VIEWPORT = { width: 1280, height: 500 };
+
+test.describe('Schedule month view scroll position', () => {
+  test.use({ viewport: SHORT_DESKTOP_VIEWPORT });
+
+  test('should open month view at the top after scrolling week view', async ({
+    page,
+    workViewPage,
+  }) => {
+    await workViewPage.waitForTaskList();
+    await page.getByRole('menuitem', { name: 'Schedule' }).click();
+
+    const wrapper = page.locator('schedule .scroll-wrapper');
+    await expect(wrapper).toBeVisible();
+    await page.getByRole('button', { name: 'View Week' }).click();
+    await expect(page.locator('schedule-week')).toBeVisible();
+
+    // Week view is a full day timeline, so it always overflows this height.
+    const scrolled = await wrapper.evaluate((el) => {
+      el.scrollTop = el.scrollHeight;
+      return el.scrollTop;
+    });
+    expect(scrolled).toBeGreaterThan(0);
+
+    await page.getByRole('button', { name: 'View Month' }).click();
+    await expect(page.locator('schedule-month')).toBeVisible();
+
+    // Precondition: without overflow there would be nothing to clamp to and
+    // this would pass on the unfixed code too.
+    const overflow = await wrapper.evaluate((el) => el.scrollHeight - el.clientHeight);
+    expect(overflow).toBeGreaterThan(0);
+
+    await expect
+      .poll(async () => wrapper.evaluate((el) => el.scrollTop), { timeout: 5000 })
+      .toBe(0);
+  });
+});
--- a/e2e/tests/schedule/schedule-month-scroll-reset.spec.ts
+++ b/e2e/tests/schedule/schedule-month-scroll-reset.spec.ts
@@ -1,18 +1,27 @@
 import { expect, test } from '../../fixtures/test.fixture';
 
 /**
- * Review of #9463: with the month grid fixed at six rows it overflows the
- * scroll wrapper on a short window, so the scroll position carried over from
- * week view clamped to the bottom and the month opened with its first row
- * above the viewport (scrollTop 121 of a 121px maximum, reproduced here).
+ * Review of #9463: a month grid that overflows the scroll wrapper on a short
+ * window kept the scroll position carried over from week view, so the month
+ * opened with its first row above the viewport (scrollTop 121 of a 121px
+ * maximum, reproduced here).
  *
  * The unit spec asserts the reset call fires; only this test runs it at a real
  * constrained height with a real overflowing month grid, which is the
  * combination the bug needed.
  *
+ * The month is navigated to rather than taken from the clock. Rows follow the
+ * displayed month (4-6), and `.month-day-cell` has `min-height: 80px`, so the
+ * track minimum sets the grid height once the rows stop fitting: at this
+ * viewport six rows overflow by 121px, five by 40, and four not at all. Taking
+ * whatever month today happens to be would make the precondition below fail
+ * for the 28 days of a non-leap February starting on the week's first day.
+ *
  * Run: npm run e2e:file e2e/tests/schedule/schedule-month-scroll-reset.spec.ts -- --retries=0
  */
 const SHORT_DESKTOP_VIEWPORT = { width: 1280, height: 500 };
+/** Rows the grid must show for the wrapper to overflow at this viewport. */
+const ROWS_THAT_OVERFLOW = 6;
 
 test.describe('Schedule month view scroll position', () => {
   test.use({ viewport: SHORT_DESKTOP_VIEWPORT });
@@ -26,6 +35,29 @@ test.describe('Schedule month view scroll position', () => {
 
     const wrapper = page.locator('schedule .scroll-wrapper');
     await expect(wrapper).toBeVisible();
+
+    // Land on a month that actually overflows, before the scroll position that
+    // the reset has to clear is set up. The view switch below keeps it: only
+    // `selectedTimeView` changes, never the selected date.
+    await page.getByRole('button', { name: 'View Month' }).click();
+    await expect(page.locator('schedule-month')).toBeVisible();
+    const monthRows = (): Promise<number> =>
+      page
+        .locator('schedule-month')
+        .evaluate((el) =>
+          Number(getComputedStyle(el).getPropertyValue('--nr-of-weeks').trim()),
+        );
+    const firstDay = (): Promise<string | null> =>
+      page.locator('.month-day-cell[data-day]').first().getAttribute('data-day');
+    // Wait for the grid to actually advance rather than re-reading the row
+    // count before Angular has re-rendered. Six rows is at most 11 months away.
+    for (let i = 0; i < 12 && (await monthRows()) !== ROWS_THAT_OVERFLOW; i++) {
+      const before = await firstDay();
+      await page.getByRole('button', { name: 'Next Month' }).click();
+      await expect.poll(firstDay).not.toBe(before);
+    }
+    expect(await monthRows()).toBe(ROWS_THAT_OVERFLOW);
+
     await page.getByRole('button', { name: 'View Week' }).click();
     await expect(page.locator('schedule-week')).toBeVisible();
 
@@ -40,7 +72,8 @@ test.describe('Schedule month view scroll position', () => {
     await expect(page.locator('schedule-month')).toBeVisible();
 
     // Precondition: without overflow there would be nothing to clamp to and
-    // this would pass on the unfixed code too.
+    // this would pass on the unfixed code too. Pinned to six rows above, so it
+    // does not depend on the month the suite happens to run in.
     const overflow = await wrapper.evaluate((el) => el.scrollHeight - el.clientHeight);
     expect(overflow).toBeGreaterThan(0);
 
--- a/src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.spec.ts
+++ b/src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.spec.ts
@@ -526,13 +526,18 @@ describe('CalendarGestureHandler', () => {
       cb.getIsExpanded.and.returnValue(false);
       cb.getActiveWeekIndex.and.returnValue(0);
 
+      // Derived, not a literal: the midpoint sits halfway between MIN_HEIGHT
+      // and MAX_HEIGHT, so a fixed pixel delta silently stops clearing it when
+      // the expanded height changes (it did when the grid went to six rows).
+      const pastMidpoint = Math.ceil((MAX_HEIGHT - MIN_HEIGHT) * 0.75);
+
       handleEl.dispatchEvent(makeTouchEvent('touchstart', 100, 100));
       // Move more than 5px to trigger drag mode, then advance time to keep
       // velocity low so the midpoint check is used rather than velocity
-      handleEl.dispatchEvent(makeTouchEvent('touchmove', 100, 200));
+      handleEl.dispatchEvent(makeTouchEvent('touchmove', 100, 100 + pastMidpoint));
       jasmine.clock().tick(1000);
       // End drag with enough delta to be past midpoint
-      handleEl.dispatchEvent(makeTouchEvent('touchend', 100, 200));
+      handleEl.dispatchEvent(makeTouchEvent('touchend', 100, 100 + pastMidpoint));
 
       jasmine.clock().tick(SNAP_DURATION + 15);
 
--- a/src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.spec.ts
+++ b/src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.spec.ts
@@ -551,5 +551,95 @@ describe('CalendarGestureHandler', () => {
 
       expect(cb.onExpandChanged).toHaveBeenCalledWith(true);
     });
+
+    // Review round 2 of #9463: with room for a single row there is nothing to
+    // expand into, so the interpolation range is 0 and the ratio would be NaN.
+    describe('at a one-row viewport', () => {
+      beforeEach(() => {
+        cb.getExpandedHeight.and.returnValue(MIN_HEIGHT);
+      });
+
+      // Asserting the exact offset, not merely the absence of 'NaN': the
+      // browser rejects translateY(NaNpx) outright, so the unguarded code
+      // leaves the transform empty and a not-toContain check passes on it.
+      it('should still write a usable transform while dragging', () => {
+        cb.getActiveWeekIndex.and.returnValue(2);
+
+        handleEl.dispatchEvent(makeTouchEvent('touchstart', 100, 100));
+        handleEl.dispatchEvent(makeTouchEvent('touchmove', 100, 200));
+
+        const innerEl = weeksEl.firstElementChild as HTMLElement;
+        expect(innerEl.style.transform).toBe(`translateY(${-2 * ROW_HEIGHT}px)`);
+        expect(weeksEl.style.maxHeight).toBe(MIN_HEIGHT + 'px');
+      });
+
+      it('should stay collapsed rather than report an expansion it cannot render', () => {
+        handler.snapTo(true);
+        jasmine.clock().tick(SNAP_DURATION + 20);
+
+        expect(cb.onExpandChanged).toHaveBeenCalledWith(false);
+        expect(cb.onExpandChanged).not.toHaveBeenCalledWith(true);
+        expect(weeksEl.style.maxHeight).toBe(MIN_HEIGHT + 'px');
+      });
+    });
+
+    // The component re-measures on every getExpandedHeight() call, so reading it
+    // per touchmove interleaves layout reads with the style writes below it.
+    it('should measure the expanded height once per drag, not per touchmove', () => {
+      handleEl.dispatchEvent(makeTouchEvent('touchstart', 100, 100));
+      cb.getExpandedHeight.calls.reset();
+
+      handleEl.dispatchEvent(makeTouchEvent('touchmove', 100, 120));
+      handleEl.dispatchEvent(makeTouchEvent('touchmove', 100, 140));
+      handleEl.dispatchEvent(makeTouchEvent('touchmove', 100, 160));
+
+      // Once for _startDrag, on the first move that crosses the threshold.
+      expect(cb.getExpandedHeight).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  // Review round 2 of #9463: the zero-duration branch cleared the inline styles
+  // and relied on Angular re-applying them, but onExpandChanged may write an
+  // unchanged signal value, in which case Angular skips the DOM write and the
+  // element is left with no max-height at all — dropping the responsive clamp.
+  describe('reduced motion', () => {
+    let reducedHandler: CalendarGestureHandler;
+    let reducedWeeksEl: HTMLElement;
+
+    beforeEach(() => {
+      (window.matchMedia as jasmine.Spy).and.returnValue({
+        matches: true,
+      } as MediaQueryList);
+      reducedWeeksEl = createMockWeeksEl();
+      reducedHandler = new CalendarGestureHandler(
+        document.createElement('div'),
+        () => reducedWeeksEl,
+        cb,
+      );
+    });
+
+    afterEach(() => reducedHandler.destroy());
+
+    it('should keep the clamped height when snapping to the state it is already in', () => {
+      cb.getIsExpanded.and.returnValue(true);
+
+      reducedHandler.snapTo(true);
+
+      expect(reducedWeeksEl.style.maxHeight).toBe(MAX_HEIGHT + 'px');
+      expect((reducedWeeksEl.firstElementChild as HTMLElement).style.transform).toBe(
+        'translateY(0px)',
+      );
+    });
+
+    it('should keep the collapsed geometry when a drag is cancelled', () => {
+      cb.getActiveWeekIndex.and.returnValue(2);
+
+      reducedHandler.snapTo(false, 2);
+
+      expect(reducedWeeksEl.style.maxHeight).toBe(MIN_HEIGHT + 'px');
+      expect((reducedWeeksEl.firstElementChild as HTMLElement).style.transform).toBe(
+        `translateY(${-2 * ROW_HEIGHT}px)`,
+      );
+    });
   });
 });
--- a/src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.spec.ts
+++ b/src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.spec.ts
@@ -50,6 +50,8 @@ const createMockCallbacks = (): jasmine.SpyObj<CalendarGestureCallbacks> =>
   jasmine.createSpyObj<CalendarGestureCallbacks>('callbacks', [
     'getActiveWeekIndex',
     'getIsExpanded',
+    'getExpandedHeight',
+    'getWeekOffset',
     'onExpandChanged',
     'onVerticalSwipe',
     'onHorizontalSwipe',
@@ -78,6 +80,12 @@ describe('CalendarGestureHandler', () => {
     cb = createMockCallbacks();
     cb.getActiveWeekIndex.and.returnValue(0);
     cb.getIsExpanded.and.returnValue(false);
+    // Unclamped geometry: what the component reports on any viewport tall
+    // enough for all six rows. The clamped case is covered in the component spec.
+    cb.getExpandedHeight.and.returnValue(MAX_HEIGHT);
+    cb.getWeekOffset.and.callFake((expanded: boolean, activeIdx: number) =>
+      expanded ? 0 : -activeIdx * ROW_HEIGHT,
+    );
 
     handler = new CalendarGestureHandler(el, () => weeksEl, cb);
   });
--- a/src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.spec.ts
+++ b/src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.spec.ts
@@ -57,6 +57,7 @@ const createMockCallbacks = (): jasmine.SpyObj<CalendarGestureCallbacks> =>
     'getIsExpanded',
     'measure',
     'getExpandedHeight',
+    'getRowHeight',
     'canExpand',
     'onExpandChanged',
     'onVerticalSwipe',
@@ -87,6 +88,7 @@ describe('CalendarGestureHandler', () => {
     cb.getActiveWeekIndex.and.returnValue(0);
     cb.getIsExpanded.and.returnValue(false);
     cb.getExpandedHeight.and.returnValue(MAX_HEIGHT);
+    cb.getRowHeight.and.returnValue(ROW_HEIGHT);
     cb.canExpand.and.returnValue(true);
 
     handler = new CalendarGestureHandler(el, () => weeksEl, cb);
@@ -571,6 +573,35 @@ describe('CalendarGestureHandler', () => {
         // dropped-rows clamp this was -4 rows and weeks 5 and 6 were unreachable.
         const innerEl = weeksEl.firstElementChild as HTMLElement;
         expect(innerEl.style.transform).toBe('translateY(0px)');
+      });
+
+      // The offset counts in whole rows, so the rows on screen have to be the
+      // height that offset assumes for the whole animation, not only after the
+      // flag flips at snapDur + 10. Collapsing from index 2 with 29px rows
+      // travelled to -80px and put week 3 in the window before jumping back.
+      it('should give the rows their post-snap height when the animation starts', () => {
+        cb.getExpandedHeight.and.returnValue(MIN_ROW_HEIGHT * WEEKS_SHOWN);
+        cb.getRowHeight.and.returnValue(MIN_ROW_HEIGHT);
+        cb.getIsExpanded.and.returnValue(true);
+
+        handler.snapTo(false, 2);
+
+        // Before the animation has finished, and before onExpandChanged fires.
+        expect(weeksEl.style.getPropertyValue('--row-height')).toBe(`${ROW_HEIGHT}px`);
+        const innerEl = weeksEl.firstElementChild as HTMLElement;
+        expect(innerEl.style.transform).toBe(`translateY(${-2 * ROW_HEIGHT}px)`);
+        expect(cb.onExpandChanged).not.toHaveBeenCalled();
+      });
+
+      it('should give the rows the shrunken height when expanding', () => {
+        cb.getExpandedHeight.and.returnValue(MIN_ROW_HEIGHT * WEEKS_SHOWN);
+        cb.getRowHeight.and.returnValue(MIN_ROW_HEIGHT);
+
+        handler.snapTo(true, 2);
+
+        expect(weeksEl.style.getPropertyValue('--row-height')).toBe(
+          `${MIN_ROW_HEIGHT}px`,
+        );
       });
 
       it('should collapse to a full-size row, never a shrunken one', () => {
--- a/src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.spec.ts
+++ b/src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.spec.ts
@@ -56,7 +56,7 @@ const createMockCallbacks = (): jasmine.SpyObj<CalendarGestureCallbacks> =>
     'getActiveWeekIndex',
     'getIsExpanded',
     'measure',
-    'getRowHeight',
+    'getExpandedHeight',
     'canExpand',
     'onExpandChanged',
     'onVerticalSwipe',
@@ -86,7 +86,7 @@ describe('CalendarGestureHandler', () => {
     cb = createMockCallbacks();
     cb.getActiveWeekIndex.and.returnValue(0);
     cb.getIsExpanded.and.returnValue(false);
-    cb.getRowHeight.and.returnValue(ROW_HEIGHT);
+    cb.getExpandedHeight.and.returnValue(MAX_HEIGHT);
     cb.canExpand.and.returnValue(true);
 
     handler = new CalendarGestureHandler(el, () => weeksEl, cb);
@@ -555,11 +555,11 @@ describe('CalendarGestureHandler', () => {
     });
 
     // Rows shrink instead of being dropped, so every geometry the handler sees
-    // still spans six rows. A viewport too short even for MIN_ROW_HEIGHT rows
-    // reports canExpand false rather than a degenerate height (#9463 review).
+    // still spans six rows. The collapsed height is always ROW_HEIGHT; only the
+    // expanded height moves, and it arrives already clamped (#9463 review).
     describe('on a shrunken grid', () => {
       it('should snap to six shrunken rows with no offset', () => {
-        cb.getRowHeight.and.returnValue(MIN_ROW_HEIGHT);
+        cb.getExpandedHeight.and.returnValue(MIN_ROW_HEIGHT * WEEKS_SHOWN);
         cb.getActiveWeekIndex.and.returnValue(4);
 
         handler.snapTo(true);
@@ -573,9 +573,23 @@ describe('CalendarGestureHandler', () => {
         expect(innerEl.style.transform).toBe('translateY(0px)');
       });
 
-      describe('too short even for that', () => {
+      it('should collapse to a full-size row, never a shrunken one', () => {
+        cb.getExpandedHeight.and.returnValue(MIN_ROW_HEIGHT * WEEKS_SHOWN);
+
+        handler.snapTo(false);
+        jasmine.clock().tick(SNAP_DURATION + 20);
+
+        // The collapsed strip is one row in a viewport with room for one, so it
+        // does not shrink with the grid: shrinking it would give every user
+        // 25px tap targets whether or not they ever expand.
+        expect(weeksEl.style.maxHeight).toBe(`${ROW_HEIGHT}px`);
+      });
+
+      describe('too short to expand at all', () => {
         beforeEach(() => {
-          cb.getRowHeight.and.returnValue(MIN_ROW_HEIGHT);
+          // What the component reports once the room is gone: the clamped
+          // expanded height has collapsed onto the collapsed height.
+          cb.getExpandedHeight.and.returnValue(ROW_HEIGHT);
           cb.canExpand.and.returnValue(false);
         });
 
@@ -585,38 +599,41 @@ describe('CalendarGestureHandler', () => {
 
           expect(cb.onExpandChanged).toHaveBeenCalledWith(false);
           expect(cb.onExpandChanged).not.toHaveBeenCalledWith(true);
-          expect(weeksEl.style.maxHeight).toBe(MIN_ROW_HEIGHT + 'px');
+          expect(weeksEl.style.maxHeight).toBe(`${ROW_HEIGHT}px`);
         });
 
-        // Asserting the exact offset, not merely the absence of 'NaN': the
-        // browser rejects translateY(NaNpx) outright, so code that produced one
-        // would leave the transform empty and a not-toContain check would pass.
-        it('should still write a usable transform while dragging', () => {
+        // The drag ceiling is the clamped expanded height, so a rubber-band
+        // cannot open the grid into room that is not there. Previously a 100px
+        // drag opened it to 124px in a viewport with 42px to spare.
+        it('should not open past the room available while dragging', () => {
           cb.getActiveWeekIndex.and.returnValue(2);
 
           handleEl.dispatchEvent(makeTouchEvent('touchstart', 100, 100));
           handleEl.dispatchEvent(makeTouchEvent('touchmove', 100, 200));
 
-          // 100px of a 120px range leaves a sixth of the collapsed offset.
+          expect(weeksEl.style.maxHeight).toBe(`${ROW_HEIGHT}px`);
+          // Asserting the exact offset, not merely the absence of 'NaN': the
+          // browser rejects translateY(NaNpx) outright, so code that produced
+          // one would leave the transform empty and a not-toContain check would
+          // pass. The range is 0 here, which is what the guard is for.
           const innerEl = weeksEl.firstElementChild as HTMLElement;
-          expect(innerEl.style.transform).toBe('translateY(-8px)');
-          expect(weeksEl.style.maxHeight).toBe(`${MIN_ROW_HEIGHT + 100}px`);
+          expect(innerEl.style.transform).toBe(`translateY(${-2 * ROW_HEIGHT}px)`);
         });
       });
     });
 
     // The component measures layout to answer this, so reading it per touchmove
     // would interleave layout reads with the style writes below it.
-    it('should read the row height once per drag, not per touchmove', () => {
+    it('should read the expanded height once per drag, not per touchmove', () => {
       handleEl.dispatchEvent(makeTouchEvent('touchstart', 100, 100));
-      cb.getRowHeight.calls.reset();
+      cb.getExpandedHeight.calls.reset();
 
       handleEl.dispatchEvent(makeTouchEvent('touchmove', 100, 120));
       handleEl.dispatchEvent(makeTouchEvent('touchmove', 100, 140));
       handleEl.dispatchEvent(makeTouchEvent('touchmove', 100, 160));
 
       // Once for _startDrag, on the first move that crosses the threshold.
-      expect(cb.getRowHeight).toHaveBeenCalledTimes(1);
+      expect(cb.getExpandedHeight).toHaveBeenCalledTimes(1);
     });
   });
 
--- a/src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.spec.ts
+++ b/src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.spec.ts
@@ -1,10 +1,15 @@
 import {
   CalendarGestureHandler,
   CalendarGestureCallbacks,
+  MIN_ROW_HEIGHT,
   ROW_HEIGHT,
-  MAX_HEIGHT,
-  MIN_HEIGHT,
+  WEEKS_SHOWN,
 } from './planner-calendar-gesture-handler';
+
+// Geometry on a viewport tall enough for six full-size rows. Shrunken rows are
+// covered by the row-height cases below and in the component spec.
+const MAX_HEIGHT = ROW_HEIGHT * WEEKS_SHOWN;
+const MIN_HEIGHT = ROW_HEIGHT;
 
 // Snap/slide durations must match the source constants
 const SNAP_DURATION = 200;
@@ -50,8 +55,9 @@ const createMockCallbacks = (): jasmine.SpyObj<CalendarGestureCallbacks> =>
   jasmine.createSpyObj<CalendarGestureCallbacks>('callbacks', [
     'getActiveWeekIndex',
     'getIsExpanded',
-    'getExpandedHeight',
-    'getWeekOffset',
+    'measure',
+    'getRowHeight',
+    'canExpand',
     'onExpandChanged',
     'onVerticalSwipe',
     'onHorizontalSwipe',
@@ -80,12 +86,8 @@ describe('CalendarGestureHandler', () => {
     cb = createMockCallbacks();
     cb.getActiveWeekIndex.and.returnValue(0);
     cb.getIsExpanded.and.returnValue(false);
-    // Unclamped geometry: what the component reports on any viewport tall
-    // enough for all six rows. The clamped case is covered in the component spec.
-    cb.getExpandedHeight.and.returnValue(MAX_HEIGHT);
-    cb.getWeekOffset.and.callFake((expanded: boolean, activeIdx: number) =>
-      expanded ? 0 : -activeIdx * ROW_HEIGHT,
-    );
+    cb.getRowHeight.and.returnValue(ROW_HEIGHT);
+    cb.canExpand.and.returnValue(true);
 
     handler = new CalendarGestureHandler(el, () => weeksEl, cb);
   });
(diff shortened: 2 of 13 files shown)


johannesjo (commented):
## Issues

### Important

- **`src/app/features/schedule/schedule/schedule.component.spec.ts:883`** — The test replaces `_windowSize` with a new signal after `_daysToShowCount` has tracked the tall-window signal. Replacing the field does not invalidate an Angular computed, so the short-window read returns the cached tall value. I reproduced the pre-PR calculation with the repository Angular runtime: `tall=6`, `short=6`, and both assertions pass. The service tests only exercise `getMonthDaysToShow` with an explicit week count, so they do not protect this caller behavior. Please assign one writable signal and call `.set({ width: 1280, height: 500 })` between reads, or drive the real resize stream, so the test fails deterministically against the old implementation.

The six-row production implementation otherwise looks correct, the linked reporter confirmed the preview, and all applicable CI checks pass.

## Verdict

Needs changes — make the short-window regression test observe an actual signal update.

johannesjo (commented):
## Additional review findings

I re-tested the current preview at head `3a7d5f5` after the existing regression-test feedback and found two production UI regressions.

### Important

- **`src/app/features/schedule/schedule/schedule.component.ts:153`** — On a constrained desktop window, switching from Week to Month preserves the timeline's vertical scroll position. The six occupied month rows now overflow the same scroll wrapper, so the retained position clamps to the bottom instead of resetting to the start of the month. At 1280×500 I reproduced `scrollTop=121` (the maximum): the first row's top was 15px while the wrapper started at 100px, so the whole first row opened above the viewport. Please reset the wrapper to the top when entering Month view and cover the Week → Month transition at a constrained height.

- **`src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.ts:6`** — Increasing the expanded calendar from five to six 40px rows overflows the Planner at short XS heights. At 400×400 I reproduced a 321px calendar host inside 299px of available route content; `planner-plan-view` collapsed to 0px, and the 28px collapse handle ran from y=350–378 while the bottom nav began at y=356, leaving only 6px unobscured. With the former five-row geometry the host was 40px shorter and the handle remained clear. Please bound the expanded geometry responsively so the collapse handle and some plan content remain reachable, with a constrained-height regression test.

### Minor

- **`src/app/features/schedule/schedule.service.spec.ts:336`** — The comment says the 24-month loop covers both a leap and non-leap February, but January 2026 through December 2027 contains no leap year. Starting the range in 2027 (or extending it) would make the stated coverage real.

The [previous signal-invalidation test finding](https://github.com/super-productivity/super-productivity/pull/9463#pullrequestreview-4872603069) also remains applicable on this head.

## Verdict

Needs changes. The six-row date coverage itself is correct, but the two constrained-height regressions should be addressed before merge.

MannXo:
Thanks, all four are real. Addressed in `5c1176d8e` (schedule) and `4cad91aeb` (planner). I verified each against the source before touching anything, and two of them turned out worse than described.

**1. 🟠 Vacuous short-window test** (`schedule.component.spec.ts`): correct, and it was worse than stated. I restored the pre-PR `schedule.component.ts` + `schedule.constants.ts` and ran the spec: the `#9449` test **passed**. The only failure was a different test, `should return the full week count in month view`, which this PR had tightened from a bounds check to `toBe(NR_OF_WEEKS)`. So the sabotage count I reported earlier came entirely from that neighbour while the height-specific test contributed nothing. Fixed as you suggested: one signal, mutated with `.set()`. Re-verified against the pre-PR pair, and this time the failure is attributed to this test.

**2. 🟠 Retained scroll on Week → Month**: reproduced exactly. At 1280×500 I get `scrollTop 121` of a 121px maximum, matching your number. The effect only scrolled when leaving month view, so entering it kept the week timeline's position. Pre-PR the old formula gave 3 rows at that height, so the month never overflowed and the missing reset was invisible; six rows exposed it, so this PR is the proximate cause.

Fixed by extending the existing effect to reset the wrapper on entering month view, and the wrapper is now a viewChild rather than a `getElementById` plus `closest` walk. Added `e2e/tests/schedule/schedule-month-scroll-reset.spec.ts` at 1280×500 as you asked for a constrained-height test, since the unit spec can only prove the reset call fires. It asserts the grid actually overflows first, so it cannot pass vacuously, and it fails on the old code with `Received: 121`.

**3. 🟠 Planner overflow at XS heights**: reproduced. Simply capping the height would have re-hidden the sixth row, which is the same complaint #9449 was filed about at a different viewport, so instead the expanded grid shows as many whole rows as fit and slides to keep the active week in view when rows are dropped. The data range stays at six weeks, so every day still exists, carries its dot and is tappable. Collapsed and unclamped-expanded both fall out of the same offset formula, so nothing changes where there is room.

Worth calling out how I got the reserve wrong first: I derived available space from `window.innerHeight` minus a measured chrome constant. It clamped correctly at 400×400 by luck. Measuring the real app showed the chrome is 77px rather than the 81 I had assumed, and at 77 the clamp stopped engaging at all. Measuring further showed the real bound is the route content bottom (356px), not the window, since the bottom nav takes the last ~44px, so `innerHeight` reserves space the plan list never receives. It now measures the rows' own offset against the parent's bottom, with no guessed chrome at all.

At 400×400 the grid settles at four rows: host 241px, `planner-plan-view` 58px, handle bottom at 298 against a route bottom of 356. At 400×1000 it is still six rows. Your 299px of available route content matches what I measure exactly.

The gesture handler needed changing too, because it writes `max-height` directly on drag and snap, so leaving it on the module constant would have let the first drag restore the overflowing geometry. It now takes the height and the offset through two callbacks.

**4. 🟡 Leap-year comment**: correct. Jan 2026 through Dec 2027 has no leap year. The range now starts in 2027, covering a non-leap February (2027) and a leap one (2028).

Full suite 14046 green (up 6), lint clean, pre-push hook passed first try. Every new test sabotage-verified: the fixed `#9449` test now fails against the pre-PR implementation where it previously passed, the scroll-reset test fails with the reset removed, the three planner clamp tests fail with the unclamped geometry restored, and the new e2e fails with `Received: 121`. Neighbouring e2e still green: schedule 4/4, planner 23/23.

One gap to be straight about: the planner change touches gesture code that only real touch input exercises. The unit tests drive the geometry directly and I drove the real app in a mobile-emulated context, but I have not tested it on a physical device, and `mobile-webkit-smoke.spec.ts` does not cover the calendar expand.

johannesjo (commented):
## Summary

The Schedule fixes in this follow-up look correct: the short-window unit test now invalidates the same signal, and the E2E test proves real overflow before checking the month-view scroll reset. The intended Planner behavior also works at 400×400, but two gesture boundary cases remain.

## Issues

### Important

- **`src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.ts:75`** — With reduced motion enabled, `snapTo()` takes the zero-duration branch and clears `maxHeight` and `transform`. When a drag is cancelled or snaps back to its existing state, `onExpandChanged()` writes the same signal value, so Angular does not restore those directly-mutated styles—the animated branch's comment already documents why clearing them is unsafe. On a constrained viewport this drops the responsive row clamp and can expose all six rows again, hiding plan content or the handle. The specs force reduced motion off, so they cannot catch this. Please apply `targetHeight`/`targetOffset` in the zero-duration branch as the animated completion branch does, and add a reduced-motion same-state/touch-cancel regression test.

- **`src/app/features/planner/planner-calendar-nav/planner-calendar-gesture-handler.ts:294`** — `_visibleWeekRows()` deliberately permits one row, so at the smallest supported heights `expandedHeight === MIN_HEIGHT === 40` and the drag interpolation becomes `0 / 0`. This produces an invalid `translateY(NaNpx)` update while expansion can still set `isExpanded=true` despite having exactly the collapsed geometry. The mode mismatch is observable: at a one-row viewport, horizontal swipes then jump whole months through the expanded branch even though the calendar still looks collapsed. Electron allows a 300×240 main window, so this is reachable. Please treat `expandedHeight <= MIN_HEIGHT` as non-expandable/keep it collapsed, or otherwise special-case the interpolation, with a gesture regression test.

### Minor

- **`src/app/features/planner/planner-calendar-nav/planner-calendar-nav.component.ts:242`** — `getExpandedHeight()` calls `_measureAvailableForRows()` on every handle `touchmove`; that performs two `getBoundingClientRect()` reads before `_updateDrag()` writes `maxHeight` and `transform`. Repeated moves therefore alternate layout reads and writes at touch-event frequency. Consider measuring once at drag start and reusing the cached geometry through move/end, while keeping the existing initialization/resize measurements.

## Strengths

- The new Schedule E2E assertion is non-vacuous and directly covers the retained-scroll regression.
- The four-row Planner clamp keeps Aug 31 reachable when active while preserving 58px of plan content at 400×400.

## Review coverage

- Target: `3a7d5f50a176ffa35dbdf4aa3465560b8859eb63..4cad91aeb9e8cee79deebe0e17c2d2e70d4584f0`
- Three independent passes covered correctness/tests, architecture/simplicity/alternatives, security, and performance; findings were independently verified against the frozen code, surrounding Electron constraints, current discussion, and CI.
- Snapshot integrity: `2bc1c857c9fa9f8b57d9b338f0bfc3a3a0a57c21df50c0b5b37c0fa7230016ce`; source/contract fingerprint: `58b13df1e03c3bebb0db92d203f82128873efaaffe5700abf65de5e2535fd533`.
- All current required CI checks pass; there are no inline review threads.

## Verdict

**Needs changes.** Confidence: high.

MannXo:
All three confirmed and fixed in `4186010bd`. Every one of them was introduced by the row clamp in `4cad91aeb`, and reproducing the first two in the running app showed both were worse than the report.

**1. 🟠 Reduced-motion snap dropped the clamp.** Correct, including the mechanism: the zero-duration branch cleared `maxHeight` and `transform` and relied on Angular re-applying them, but `onExpandChanged()` frequently writes an unchanged signal value so the DOM write is skipped. The animated branch four lines below already documents exactly this and keeps its inline styles; the zero-duration branch now does the same.

Reproduced at 400×400 with `prefers-reduced-motion: reduce`. Second drag on the handle, before the fix: inline `max-height` is `""` and the grid renders 240px, all six rows. After: `160px`, clamp held.

Worth noting this is structurally pre-existing, but harmless before the clamp existed, since clearing the style just fell back to the same 240px the binding would have produced. The clamp is what turned it into a real regression.

**2. 🟠 One-row viewport.** Correct on both halves. Applied both remedies rather than only the interpolation guard: expansion now requires more than one row of travel, so `isExpanded` can no longer disagree with the geometry and send horizontal swipes down the month-jump branch, and the interpolation guards the zero range independently because the NaN is a latent trap regardless.

At 300×240 with reduced motion, before the fix: `max-height` `""`, grid 240px, `planner-plan-view` **0px**, and the collapse handle's bottom edge at y=378 in a 240px viewport, so 138px below the bottom of the screen. After: stays collapsed at 40px, handle bottom at 178, plan view 18px, and all 42 days still built.

One detail on the test for this. My first version asserted the transform did not contain `NaN`, which **passed against the unfixed code** because the browser rejects `translateY(NaNpx)` outright and leaves the property empty, so the assertion was testing the browser rather than the fix. It now asserts the exact expected offset, which the unfixed code cannot produce.

**3. 🟡 Layout thrash.** Fixed as suggested: `getExpandedHeight()` is sampled once in `_startDrag()` and reused through move and end. `snapTo()` still measures fresh, since it is also called outside a drag. Initialization and resize measurements are unchanged.

Five new specs in `planner-calendar-gesture-handler.spec.ts`, in a reduced-motion describe with its own `matchMedia` setup, since the existing one pins reduced motion off precisely as you noted. All five sabotage-verified against `4cad91aeb`: **5 FAILED, 38 SUCCESS**.

Full suite 14051 green (up 5), lint clean, pre-push hook passed first try. Verified in the running app at 400×400 with and without reduced motion, and at 300×240.

The physical-device caveat from my last comment still stands: this is all emulated touch input.

johannesjo:
Reviewed `5c1176d8e` and `4cad91aeb`, with `4186010bd` on top.

The three gesture cases from the last round check out in the source: the zero-duration branch writes `targetHeight`/`targetOffset` instead of clearing, `expanded` is gated on `expandedHeight > MIN_HEIGHT` so the flag can no longer disagree with the geometry, and the interpolation guards `range > 0`. Sampling `getExpandedHeight()` once in `_startDrag()` while leaving `snapTo()` to measure fresh is the right split, since `snapTo()` is also reached from `_onTouchCancel` and the swipe handler.

On whether the clamp belongs in this PR at all: your own 400×400 numbers settle it. Four rows means `_availableForRows` landed in `[160, 200)`, so five rows would still have fit at that size — the clamp is repairing this PR's own 5→6 regression there, not a pre-existing overflow. It belongs here.

Four things below, none blocking.

**1. 🟠 The measurement has no regression guard.** `_measureAvailableForRows()` is the only genuinely new runtime logic in `4cad91aeb`, and nothing exercises it. Both `setRoom()` helpers write `_availableForRows` directly — one deliberately after `detectChanges()` so the real measurement can't overwrite it — and no spec stubs `getBoundingClientRect`. That leaves three load-bearing things with no test:

- `HANDLE_HEIGHT = 28` restates `.handle { padding: 12px 0 }` plus `::after { height: 4px }`. Correct today; silently 4px too generous the day someone changes that padding, with nothing failing.
- `_elRef.nativeElement.parentElement` assumes the nav is a direct child of the `planner` host. True today (`planner.component.html:2`), and it silently becomes a different box if that `@if` ever gets wrapped in a div.
- The arithmetic itself.

You verified all of it by hand at 400×400 and 300×240 and the numbers you posted are convincing, so this isn't "unverified" — it's "nothing will tell the next person when it breaks". A spec that stubs the two rects and asserts the resulting row count pins all three, and it would have caught the `window.innerHeight` version you describe abandoning.

**2. 🟠 The re-measure triggers look inverted.** The effect at `planner-calendar-nav.component.ts:255` re-measures on `isExpanded()`, but neither input to the measurement moves when the calendar expands: `.weeks`'s top is the host top plus the two label rows, and the parent's bottom is fixed by `height: 100%` on the planner host. So that dependency is a no-op. Meanwhile what *does* move it fires nothing — `<banner>` sits above `.route-wrapper` in `app.component.html`, so a banner appearing shrinks the route content and shifts `weeksTop` with no window resize, leaving the clamp stale until the next expand or resize.

A `ResizeObserver` on the parent replaces both the effect and the debounced `fromEvent(window, 'resize')`, covers the banner/header case, and drops the 50ms debounce. There are four in the codebase already (`right-panel`, `main-header`, `schedule-event`, `onboarding-hint`).

**3. 🟡 Scope the `#work-start` lookup now that the `closest()` walk is gone.** `schedule.component.ts:405` still resolves the element with `document.getElementById(elementId)` while the container now comes from a different scope (`viewChild`). `schedule-day-panel` renders its own `schedule-week` (`schedule-day-panel.component.html:15`), and `isShowScheduleDayPanel` is global layout state rather than route-scoped, so two `#work-start` elements can be in the DOM at once while the Schedule route is open. It resolves correctly today only because the route content precedes `.side` in `right-panel.component.html` — and before this change the `.closest('.scroll-wrapper')` walk self-corrected anyway, since the panel copy has no such ancestor and the function bailed on `null`. `scrollContainer.querySelector('#work-start')` makes it order-independent. Not a live bug, just the one property the refactor gave up.

**4. 🟡 `should not reset the wrapper while staying in month view` can't fail for the reason it names.** The effect depends only on `isMonthView()`, and the test changes no signal between installing the spy and asserting, so it rules out a reset wired to run per CD pass and nothing else. The regression it reads like it is guarding — the effect picking up another dependency and re-firing — would slip straight through, since `daysToShow()` doesn't change here either. A `goToNextPeriod()` before the second assertion would cover that.

Confirmed good along the way: the leap-year fix is real (Jan 2026 through Dec 2027 contains no leap year, so the comment described coverage the loop never had); the `_windowSize` fix is real (replacing the signal field can't invalidate the computed, so the old assertion was reading a cached value); and the new e2e asserting `scrollHeight - clientHeight > 0` before polling `scrollTop` is the right shape — it can't pass vacuously on a window where the grid happens to fit.

MannXo:
All four addressed in `084330f64`. Two of them exposed tests of mine that could not fail, which is now three rounds running, so I have stopped trusting a green run and started checking *which* test fails under sabotage every time.

**1. 🟠 No guard on the measurement.** Correct on all three sub-claims, and the spec you describe does exactly what you predicted: restoring the `window.innerHeight` version fails all four of the new tests.

Four specs stub the two rects and assert the resulting row count, covering the arithmetic, the `parentElement` assumption and the XS geometry.

`HANDLE_HEIGHT` needed a different treatment. My first version asserted `_availableForRows()` equalled `356 - 110 - HANDLE_HEIGHT - MIN_PLAN_VIEW_HEIGHT`, which restates the implementation and holds for any value: setting the constant to 24 kept the whole file green. It now asserts the literal 178, and a separate test compares `HANDLE_HEIGHT` against the rendered `.handle`'s measured height. Changing the constant fails two tests; changing `.handle`'s padding fails the pinning one. That is the guard you were asking for, and my first attempt was not it.

**2. 🟠 Triggers inverted.** Confirmed exactly as you describe: the planner host is `height: 100%` so the parent's bottom is fixed, and the rows' top is host-top plus the two label rows, so `isExpanded()` moved neither input. Replaced the effect body and the debounced `fromEvent(window, 'resize')` with a `ResizeObserver` on the parent.

Measured in the app at 400×400, shrinking the route content with no window resize at all: before, the grid stayed at 160px and `planner-plan-view` was squeezed to 19px; after, the grid moves to 120px and the plan list keeps 59px. Restoring the height returns it to 160px. That is your banner case.

**3. 🟡 `#work-start` scoping.** Correct, and it is the one property the refactor gave up rather than a live bug, as you say. Now `scrollContainer.querySelector('#' + elementId)`, so it is order-independent again. Covered by a spec that puts a stray `#work-start` in the document while the component is in month view (so the stray is the only candidate) and asserts nothing is scrolled; it fails against `document.getElementById`.

**4. 🟡 The negative test.** Correct, and my first fix for it did not work either. Adding `goToNextPeriod()` was not enough: `getMonthDaysToShow` is mocked to return one fixed array, and a computed that yields the same reference is treated as unchanged, so the effect would not have re-run regardless. I confirmed this by adding a spurious `daysToShow()` dependency to the effect, which the test still passed. It now also returns a different array from the mock, and with that the same sabotage fails it.

Full suite 14057 green (up 6), lint clean, pre-push hook first try. Schedule e2e 4/4 and planner e2e 23/23 still green.

On the clamp belonging in this PR: agreed, and thanks for working it out from the numbers rather than taking my word for it.

johannesjo:
Review of the two new commits (`3a7d5f5..4cad91a`: scroll reset + bounded calendar).

I checked the planner changes out locally and ran probes against them, so the two findings below are reproduced, not inferred. The PR's own specs pass (37 + 38 green) — they just don't cover these cases.

---

## 1. Conflicts with `master`, and one hunk is now dead code

`git merge-tree master 4cad91a` conflicts in `src/app/features/schedule/schedule/schedule.component.ts`.

master's #9545 (`e5f7888`) **deleted** `_scrollIntoViewWithTimeColumnOffset` and replaced it with `_scrollAnchorToTop()` / `scrollIntoView`. So the `element.closest('.scroll-wrapper')` → `viewChild` refactor in this commit no longer applies to anything.

After a rebase only the `else { setTimeout(() => this._resetScrollWrapper()) }` branch and the `#scrollWrapper` template ref survive — and those are fine on top of master's version (`scrollTo({top: 0, left: 0})` is scale-independent, so it isn't affected by the `warpRoute` `scale(1.2)` caveat #9545 documents).

## 2. `_updateDrag` divides by zero → the drag transform silently stops updating

`planner-calendar-gesture-handler.ts:294`

```ts
const progress = (newHeight - MIN_HEIGHT) / (expandedHeight - MIN_HEIGHT);
```

When only one row fits, `getExpandedHeight()` returns `MIN_HEIGHT`, `newHeight` clamps to `MIN_HEIGHT`, and this is `0 / 0` → `NaN`. `offset` becomes `NaN`, and `translateY(NaNpx)` is rejected by CSSOM — the assignment is a silent no-op, so the inner element keeps whatever transform it had while `max-height` keeps tracking the finger.

Reproduced by driving a real handle drag with `getExpandedHeight: () => MIN_HEIGHT`:

```
maxHeight=40px   transform=      <-- max-height was written, the transform write was dropped
```

That one-row state is explicitly supported (`Math.max(1, fits)`, `planner-calendar-nav.component.ts:148`) and the new spec exercises it (`setRoom(0)` → "should always keep at least one row"). Guard the denominator.

## 3. Clamping the row count re-creates #9449 on a short viewport

`_weekOffsetFor` (`planner-calendar-nav.component.ts:177`) clamps only the bottom:

```ts
const firstRow = Math.min(activeIdx, WEEKS_SHOWN - rows);
```

With `activeIdx = 0` the window is pinned to rows 0…rows-1 and the tail of the grid is rendered but off-screen — and there is no gesture that scrolls *within* an expanded grid (vertical and horizontal swipes both jump a whole month), so those days cannot be tapped at all.

Reproduced with the exact August 2026 scenario this PR uses for #9449 — anchor Mon 27 Jul, `visibleDayDate = 2026-08-01` (row 0), room for four rows:

```
active=0  rows=4  firstVisible=0
gridHasAug31=true       <-- the data is built, as the comment says
VISIBLE_hasAug31=false  <-- but the window ends at 2026-08-23
lastVisible=2026-08-23
```

Aug 24–31 have no reachable cell. That is the symptom #9449 was filed for, at a different viewport. The doc comment at `:138-145` states the opposite ("every day of the month exists, carries its task dot and **stays tappable**"), and the new spec only asserts the `activeIdx = 5` direction, which is the half that works.

**Suggested alternative:** clamp the *row height* instead of the *row count* — `rowHeight = min(ROW_HEIGHT, floor(available / WEEKS_SHOWN))`. All six rows always render, so nothing becomes unreachable; `weekOffset` stays `0` when expanded, and `_weekOffsetFor` / the `getWeekOffset` callback disappear entirely. It's also the same reasoning this PR already applies to the schedule grid ("fewer rows only ever removes days — it never makes them taller").

---

## Smaller things

- **Forced reflow on every `touchmove`.** `_updateDrag` (`:285`) calls `getExpandedHeight()`, which re-runs `_measureAvailableForRows()` → two `getBoundingClientRect()` calls, right after the previous frame wrote `weeksEl.style.maxHeight`. Read-after-write once per frame for the whole drag. Measure once in `_startDrag` and cache it.
- **`getExpandedHeight()` writes a signal** (`component.ts:242-245`). A getter with a `.set()` side effect, invoked from touch handlers — worth splitting the measure from the read, or renaming.
- **`HANDLE_HEIGHT = 28` duplicates the SCSS** (`.handle` is `padding: 12px 0` plus a 4px bar). A style tweak breaks the clamp with no failing test. Either measure the handle element, or add the same kind of cross-reference this PR already uses for `NR_OF_WEEKS` ("Must stay in sync with `--nr-of-weeks`").
- **`MIN_PLAN_VIEW_HEIGHT = ROW_HEIGHT`** aliases the *calendar row* height as "one task row of the plan list". Those are unrelated quantities that happen to match today — give it its own literal.
- **`window.resize` misses the cases that matter here.** The planner's height changes without a window resize (add-task bar, bottom nav, virtual keyboard). `ResizeObserver` on the measured container is the right tool and is already the house pattern (`main-header`, `right-panel`, `schedule-week`).
- **`_measureAvailableForRows` reaches into `this._elRef.nativeElement.parentElement`** (`:157`), coupling the component to `planner.component.html`'s exact nesting. Wrapping the element anywhere silently changes the measurement, with nothing failing.
- **Unnecessary exports:** `HANDLE_HEIGHT` and `MIN_PLAN_VIEW_HEIGHT` are `export`ed but referenced only inside their own file. And `MAX_HEIGHT` in the gesture handler now has no production consumer left at all — only specs and one doc comment.
- **`schedule.component.spec.ts` — "should not reset the wrapper while staying in month view"** is close to vacuous: the effect depends only on `isMonthView()`, so nothing in the test could have re-triggered it.
- **"should keep the active week visible when the height is clamped"** sets `_availableForRows` *after* `detectChanges()`, with a comment explaining that the re-measure effect would otherwise overwrite it. A test built around effect ordering will break on the next refactor here.
- **The new e2e**: could you confirm it actually fails on the unfixed code? The comment cites scrollTop 121/121, but it's a full app boot for an assertion the unit spec already makes, so it's worth knowing it genuinely gates the fix.

## Nice

- The `schedule.service.spec.ts` year bump 2026 → 2027 is a real fix, not churn: Jan 2026 + 24 months covered Feb 2026 *and* Feb 2027, both non-leap, so the old comment's leap-year claim was false.
- The `_windowSize` spec change (mutate via `.set()` rather than replacing the field) is right, and is exactly what turns that test into a real guard against reintroducing height-dependence.
- Nothing here touches sync or op-log state.

---

Items 1–3 look blocking to me; the rest are take-or-leave. Item 3 is the one I'd most like your view on, since the suggested alternative would delete a good chunk of this commit rather than add to it.

MannXo:
Rebased onto master and took your suggested alternative for item 3, which turned out to be the right call for a reason neither of us had spotted. One thing to flag first: this review was against `3a7d5f5..4cad91a`, so it does not include `4186010bd` (round 2) or `084330f64` (round 3). Several items are already fixed there, listed at the bottom.

**1. 🟠 Conflict with master, and the dead hunk.** Correct in every particular. Rebased; the conflict was only `schedule.component.ts`, and after resolving it the diff against master is exactly what you predicted. `_scrollAnchorToTop` already scopes through `this._elRef.nativeElement.querySelector`, so my `closest('.scroll-wrapper')` → viewChild refactor is gone entirely. What survives is `NR_OF_WEEKS`, the `else { setTimeout(() => this._resetScrollWrapper()) }` branch and the `#scrollWrapper` ref, and `scrollTo({top: 0})` is scale-independent as you say.

The stray-`#work-start` spec I had added is retargeted at `_scrollAnchorToTop` rather than deleted, since master has no coverage of that scoping. Say the word and I will drop it.

**3. 🟠 The row-count clamp hides the tail.** Correct, and I have taken the row-height alternative. One correction to the report: those days are not entirely untappable. Collapse the grid and `_slideCollapsedWeek` walks rows 0 through 5 one at a time on a horizontal swipe, so they are reachable that way. It is the expanded view failing the promise in its own doc comment, which is bad enough.

Your alternative is better than a fix for that one case. Six rows always render, `weekOffset` is 0 whenever expanded, and `_weekOffsetFor`, the `getWeekOffset` callback and the divide-by-zero interpolation all go. The row height now reaches the CSS as `--row-height`, so `.week button`'s dimensions no longer restate `ROW_HEIGHT` either.

The one thing I added on top: below six `MIN_ROW_HEIGHT` (24px) rows the calendar refuses to expand rather than rendering an illegible grid, through the same `canExpand` gate round 2 added. That is not decoration. `floor(available / WEEKS_SHOWN)` can reach 0, and expanding into a degenerate height is the flag/geometry desync you found in round 2.

**The measurement was also wrong, and the row count was hiding it.** I found this while verifying the above in the app, so it is not on your list. The first measurement lands while the route enter animation is still at `scale(1.2)`, and `getBoundingClientRect` is scaled with it. At 400×400 it reported 227px of room where there were 178, which bought a whole extra row:

```
                      before (4cad91a..084330f)   after
_availableForRows     227                         178
rows shown            5 of 6                      6 of 6
days on screen        35 of 42                    42 of 42
last reachable day    2026-09-27                  2026-10-04
planner-plan-view     18px  (reserve is 40)       44px
handle bottom         338                         312   (route content ends at 356)
```

`offsetTop` and `clientHeight` are layout coordinates and no ancestor transform scales them. This is the same trap #9545 documents for the schedule, which is where I went looking after the numbers did not add up. The initial read also moves to `afterNextRender`, so the month label and day labels are laid out before it, and the gesture handler re-reads through an explicit `measure()` callback rather than the getter with a hidden `.set()` you flagged.

Pinned by a spec that measures, applies `scale(1.2)` to an ancestor and measures again, asserting the two are equal. Against the rect version it fails with `Expected 228.4000244140625 to be 179`.

**Your e2e question.** Yes, it gates the fix. Deleting the `else` branch and running it post-rebase:

```
Expected: 0
Received: 121
```

**Smaller items.** `MIN_PLAN_VIEW_HEIGHT` has its own literal now instead of aliasing `ROW_HEIGHT`. The container is `closest('planner')` rather than `parentElement`, so wrapping the nav cannot silently change what is measured, with a spec that wraps it. `MAX_HEIGHT` and `MIN_HEIGHT` are deleted, both being dynamic now. `HANDLE_HEIGHT` stays exported because the spec that pins it against the rendered `.handle` imports it. The two specs you called fragile are gone: the clamped-active-week one no longer applies, and the `setRoom` helpers now disconnect the observer first rather than relying on ordering.

**Already fixed in the two commits this review predates:**

- `_updateDrag` dividing by zero, guarded in `4186010bd`
- `getExpandedHeight()` per `touchmove`, sampled once in `_startDrag` in `4186010bd`
- `window.resize` missing in-app shrink, replaced by a `ResizeObserver` in `084330f64`
- `HANDLE_HEIGHT` with nothing failing, pinned against the rendered `.handle` in `084330f64`
- the near-vacuous stay-in-month-view spec, fixed in `084330f64`

**Numbers.** Full suite 14640 passing, 0 failing, and 14626 on the `America/Los_Angeles` run. Lint clean, 0 errors. Schedule e2e 8/8, planner e2e 24/24. Planner specs 91 green, and sabotage-verified twice: restoring the row-count clamp fails 5 of them including the reachability one (`Expected 240 to be 160`, which is the two rows that were off screen), and restoring the rect-based measurement fails 5.

**What I did not do.** Still no physical device. The gesture paths are exercised by unit tests driving the geometry and by a mobile-emulated browser, and `mobile-webkit-smoke.spec.ts` does not cover the calendar expand.

One consequence worth naming since you may not want it: the collapsed one-week strip now uses the same shrunken row height as the expanded grid (29px rather than 40px at 400×400), even though there is room for a full-size row when collapsed. Two different row heights would mean the rows resize mid-drag, so I kept one. Happy to change it if you would rather the collapsed strip always stayed at 40px.

johannesjo:

Thanks for the PR, and for taking the row-height alternative – that half reads well now: six rows always render, `weekOffset` is 0 whenever expanded, and `_weekOffsetFor` plus the divide-by-zero interpolation are gone with it. The `offsetTop`/`clientHeight` switch is a real catch, and the spec that scales an ancestor and asserts the two readings match is the right shape for it. I focused on the two commits the earlier passes predate (`705cdda0`, `d5d8e2bc`) and on the final state of the diff; the central claim checks out (`daysToGoBack` is 0-6 and the longest month is 31, so `6 * 7 = 42` covers every month/`firstDayOfWeek` pair).

### Should fix before merge

- **`planner-calendar-nav.component.ts:153`** – `canExpand()` gates *entering* the expanded state, but nothing ever leaves it. `isExpanded` is written only at `:264` (`onExpandChanged`, reachable only from `snapTo`), and `canExpand()` has exactly one consumer, `planner-calendar-gesture-handler.ts:87`. So once expanded, a later shrink of the measured box leaves `isExpanded` true while `canExpand()` is false, and since `rowHeight()` floors at `MIN_ROW_HEIGHT`, `maxHeight()` bottoms out at a fixed `24 * 6 = 144px` no matter how little room is left. Angular then writes that over `snapTo`'s inline value via `[style.max-height.px]`.

  The trigger is the case your own comment at `:278-281` names: a `<banner>` appearing above `.route-wrapper` shrinks the route content and fires no resize event. Take your measured XS geometry (planner `clientHeight` 356, weeks top 110, so `_availableForRows` 178 and an expanded grid of 174). A banner drops that to roughly 130, but `maxHeight()` is now 144 against 130px of room, so the grid overflows and eats into the 40px `MIN_PLAN_VIEW_HEIGHT` reserve – the overflow the clamp exists to prevent. Note `isXs` is a max-*width* query, so a height-only shrink does not destroy the component and the stale flag survives.

  Your `maxHeight` spec brackets this precisely: `should not expand past the room available below the rows` uses `setRoom(ROOM_FOR_FOUR_ROWS)` (160, still above the 144 floor) and passes; substitute `setRoom(100)` and it fails. That untested region is exactly the hole.

  Cheapest fix is to clamp in one place: `maxHeight()` takes `Math.min(..., this._availableForRows())`, which is a no-op whenever `canExpand()` is true and also covers the nit below. The alternative is an `effect` that collapses (`isExpanded.set(false)` or `_gesture.snapTo(false)`) when `canExpand()` goes false.

### Nits

- **`schedule.constants.ts:33`** – the `NR_OF_WEEKS` comment predates the rebase and no longer matches the base. It says rows are `1fr` in a fixed-height grid so "fewer rows only ever removes days – it never makes them taller", and points at `--nr-of-weeks` in `schedule.component.scss`. Since `ecc334113d` (#9623, an ancestor of your merge base) `ScheduleMonthComponent` binds `[style.--nr-of-weeks]` to `weeksToShow()` on its own host, which shadows that declaration for `.month-grid-container`. So three rows are now 1/3 of the grid height rather than 1/6, and the file you point at no longer reaches the grid. Same for "there is no layout change" in the description – your own new e2e asserts `scrollHeight - clientHeight > 0` at 1280x500, which is the layout change. Worth restating as the trade-off it actually is: six rows always, shorter cells on a short window, and the grid overflows and scrolls (which is what the scroll reset is for).

- **`planner-calendar-gesture-handler.ts:311`** – the drag path never consults `canExpand()`; the ceiling is always `_dragRowHeight * WEEKS_SHOWN`. Your own spec pins the consequence: with `canExpand()` false and `getRowHeight()` at `MIN_ROW_HEIGHT`, `should still write a usable transform while dragging` asserts `max-height: 124px` after a 100px drag, i.e. the grid opens to 124px in a viewport with 42px of room before snapping back. It is transient and self-correcting, so if the rubber-band is deliberate a line saying so next to that spec is enough; otherwise cap the ceiling at `_dragRowHeight` when `canExpand()` is false. The `maxHeight()` clamp above would cover it too.

### Questions

- The general Playwright suite doesn't run on PRs here (`ci.yml` has no e2e job, and `e2e-scheduled.yml` is nightly/dispatch), so `schedule-month-scroll-reset.spec.ts` has never been gated by CI on this branch. You reported it failing with `Received: 121` after the rebase – was that measured on `d5d8e2bc` as it now stands, or before the last commit?

---

johannesjo:
@MannXo on the question you closed with last round: keep the collapsed strip at `ROW_HEIGHT`. Rows resizing between collapsed and expanded is acceptable; 25px day buttons are not. Today `maxHeight()` and `weekOffset()` both read `rowHeight()`, so the collapsed strip shrinks with the expanded grid even on a viewport with room to spare for one full-size row — and since `.week button` is `min(36px, var(--row-height) - 4px)`, the 29px rows you measured at 400x400 give 25px tap targets. That is a worse regression than the one we are fixing, and it applies whether or not the user ever expands.

On the row count: please derive it (4-6) from the displayed month instead of pinning it at six. The objection in the description is out of date — since #9623 `ScheduleMonthComponent` binds `[style.--nr-of-weeks]` to `weeksToShow()`, which you already feed from `ceil(daysToShow().length / 7)`, so the tracks follow the rendered count on their own, and the dead `:host[style*=…]` event-density rules were deleted there rather than left to be wired up. `getMonthDaysToShow` keeps its literal `numberOfWeeks`; the arithmetic is the one expression `ceil((daysToGoBack + daysInMonth) / 7)`. One correction to my own framing: it cannot go where the constant sits now, since `_daysToShowCount` sees neither `selectedDate` nor `firstDayOfWeek` — put it in the `daysToShow` month branch, or in the service, which already computes `daysToGoBack`. That also shrinks the "six rows always, shorter cells, grid overflows" trade-off down to the months that genuinely need six.

The `canExpand()` item from today's review still needs fixing on top of both of these.

Thanks for the measurements — they are what made both of these decidable rather than a matter of taste — and for the patience across this many rounds.

MannXo:
All three addressed in `a7eca4231`, including the row count. Your correction about `--nr-of-weeks` was right and my objection was out of date: I checked `ecc334113d` against the base and `ScheduleMonthComponent` does bind `[style.--nr-of-weeks]` to `weeksToShow()` on its own host, which shadows the declaration on `schedule`, and the `:host[style*=…]` density rules are gone from `schedule-month.component.scss`. Nothing had to be wired up.

**🔧 `canExpand()` gates entering and nothing leaves.** Correct, and reproducible. I shrank the observed box under an expanded grid at 400x400, which is what a banner amounts to for this component:

```
                     before the clamp      after
_availableForRows    59                    59
isExpanded           true (stale)          true (stale)
canExpand            false                 false
maxHeight            144                   59
handle bottom        282                   197
planner bottom       237                   237
planner-plan-view    0px                   40px
```

Without the clamp the handle sits 45px below the route content and the 40px reserve is gone entirely, which is the overflow the clamp exists to prevent. `maxHeight()` now takes `Math.min(…, _availableForRows())` as you suggested. `isExpanded` is still stale afterwards, and I left it that way deliberately: clamping is a pure computed, where collapsing from an effect would write a signal from a measurement and put the flag and the geometry back in a race. Say the word if you would rather it collapse.

**Row count derived, 4-6.** `ceil((daysToGoBack + daysInMonth) / 7)`, in the service. It could not go beside the old constant for the reason you gave: `_daysToShowCount` reads neither `selectedDate` nor `firstDayOfWeek`. `SCHEDULE_CONSTANTS.MONTH_VIEW` is deleted rather than left at a value nothing reads.

You are right that this narrows the trade-off rather than removing it, and I have restated it in the description as what it is. Cells are `1fr` inside `height: calc(100vh - 120px)`, so the row count changes cell height and never total height. Measured at 1280x500 on September 2026, which derives five rows: 35 cells from `2026-08-31` to `2026-10-04`, `scrollHeight` 440 against `clientHeight` 400. The grid still overflows and scrolls, which is what the reset is for, and the overflow does not depend on the month.

**Collapsed strip stays at `ROW_HEIGHT`.** Taken as decided. At 400x400 the collapsed strip is 40px again while the expanded grid still shrinks to 29px rows, so nobody gets 25px tap targets for a state they never entered. `weekOffset` moves in full rows to match.

**The drag ceiling.** Not deliberate, so I capped it rather than documenting it. `_startDrag` now samples the clamped expanded height, so the 124px-into-42px rubber band is gone; the spec asserts the grid does not move at all when there is no room, and that the transform is still written (the range is 0 there, which is what the guard is for).

**Your e2e question.** Measured on `d5d8e2bc` exactly as it stood: I checked that commit out, deleted the `else` branch and ran it.

```
Expected: 0
Received: 121
```

On `a7eca4231` it still gates, at a smaller margin because September needs five rows rather than six:

```
Expected: 0
Received: 40
```

Worth naming since you flagged the suite is not CI-gated here: that 40 is the whole overflow at 1280x500, so the test is thinner than it was, though not month-dependent for the reason above.

**Numbers.** Full suite 14645 passing, 0 failing, and 14631 on the `America/Los_Angeles` run. Lint 0 errors. Schedule e2e 8/8, planner e2e 24/24.

Sabotage, one revert at a time:

| Reverted | Fails | Still passes |
| --- | --- | --- |
| the `_availableForRows` clamp in `maxHeight()` | `should not exceed the room available while stale-expanded` | the other 48 nav specs |
| the collapsed `ROW_HEIGHT` pin | `should keep collapsed rows full size on a viewport that shrinks the grid`, `should keep the collapsed offset at full rows on a shrunken viewport` | the other 47 |
| the derived count, pinned back to 6 | `should derive 4 to 6 weeks, and never fewer than the month needs`, `should never return a count that drops a day of the month` | the other 50 service specs |

The second of those walks 24 months x 7 first-days and asserts both that the last day of the month is present and that one row fewer would drop it, so it pins the count from both sides rather than restating the formula.

johannesjo:

Thanks – I reviewed only `d5d8e2bc..a7eca423`, since the earlier passes cover everything before it. All three items from last round land: the `maxHeight()` clamp is the one-place fix and your `should not exceed the room available while stale-expanded` spec brackets exactly the region that was untested; the collapsed strip is back at `ROW_HEIGHT` with `weekOffset` moving in full rows to match; and the drag ceiling is now the clamped expanded height, so the 124px-into-42px rubber band is gone. The derived row count is correct – `getMonthWeeksToShow` and `getMonthDaysToShow` compute `daysToGoBack` identically, and your 24×7 loop pins the count from both sides.

First, a correction to something I told you last round: **the general Playwright suite _is_ gated on every PR.** `ci.yml`'s `Tests` job runs `npm run e2e:ci` → `e2e:all --grep-invert "@supersync|@webdav"`, which includes `e2e/tests/schedule/`. My earlier "no e2e job on PRs" was wrong. That is good news for the test's value, and it is also what makes the item below matter.

### Should fix before merge

- **`e2e/tests/schedule/schedule-month-scroll-reset.spec.ts:44`** – the `overflow > 0` precondition is month-dependent, and in a 4-row month it fails.

  Your description says the row count "changes cell height and never total height", but `1fr` in `grid-template-rows: auto repeat(var(--nr-of-weeks), 1fr)` is `minmax(auto, 1fr)`, and `.month-day-cell` carries `min-height: 80px` (plus the 1px gap). So the track minimum, not the fr share, sets the grid's height once `N × 81` exceeds `calc(100vh - 120px)`. Your own two measurements say the same thing: 121 at six rows, 40 at five – a difference of exactly 81 per row, which is the cell minimum plus the gap.

  Extending that at 1280×500 (wrapper `clientHeight` 400, `calc(100vh - 120px)` = 380, header row ≈ 35):

  | rows | grid height | overflow |
  | --- | --- | --- |
  | 6 | 35 + 480 + 6 = 521 | 121 ✓ (your number) |
  | 5 | 35 + 400 + 5 = 440 | 40 ✓ (your number) |
  | 4 | 35 + 320 + 4 = 359 → floors at 380 | **0** |

  `getMonthWeeksToShow` returns 4 only for a non-leap February whose 1st falls on the week's first day. With `DEFAULT_FIRST_DAY_OF_WEEK = 1` that is **February 2027** – Feb 1 2027 is a Monday – and it is the only such month between 2026 and 2036. The test uses the system month, so for those 28 days `expect(overflow).toBeGreaterThan(0)` fails and, because the suite runs in the `Tests` job, every PR in the repo goes red.

  Pin the month the assertion runs against rather than taking whatever today's is: navigating with the next-period button until the grid renders six rows would do it, and it also restores the 121px margin instead of the current 40.

### Nits

- **`planner-calendar-gesture-handler.ts:95`** – the collapse animation targets an offset computed from `ROW_HEIGHT` while the rows on screen are still `rowHeight()`, because `onExpandChanged(false)` only fires at `snapDur + 10`. At 400×400 (`_availableForRows` 178 → `rowHeight()` 29) collapsing from `activeWeekIndex` 2 animates `translateY` to `-80px` over 200ms against 29px rows, so the 40px window lands on content `[80,120]` – mostly row 3 – and then jumps back to row 2 when the flag flips and the rows return to 40px. At index 5 it lands past the end of the 174px content, i.e. blank, before popping. Writing `--row-height` to the post-snap value at the start of the animation (rather than letting the flag do it at the end) makes the target offset and the rendered rows agree throughout. I derived this from the geometry rather than observing it on a device, so if you have a short viewport handy it is worth a look. It is transient either way, and only on viewports where six 40px rows do not fit.

- **`planner-calendar-nav.component.ts:215`** – the new `Math.min(…, this._availableForRows())` also clamps the *collapsed* branch, which the doc comment above it does not cover ("a no-op whenever `canExpand()` is true" holds for the expanded branch only). `_availableForRows` is `clientHeight - top - HANDLE_HEIGHT - MIN_PLAN_VIEW_HEIGHT` with no floor, so at an Electron minimum window (300×240, and `minHeight: 240` is in `electron/main-window.ts:218`) it can go under 40 or negative. Under 40 the collapsed strip renders as a fraction of a row where the old code never went below `MIN_ROW_HEIGHT`; negative is worse in a quieter way, since `max-height: -32px` is rejected by CSSOM and the element silently keeps whatever value was last valid. `this.isExpanded() ? Math.min(this.rowHeight() * WEEKS_SHOWN, this._availableForRows()) : ROW_HEIGHT` keeps both of your `maxHeight` specs passing and gives the collapsed case defined behaviour; `Math.max(0, …)` is the smaller version if you would rather keep clamping it.

---

MannXo:
All three in `bb938b15e`. Thanks for both corrections, and the e2e one was mine to own: that test would have reddened every PR in the repo, not just this one.

**🔧 The precondition is month-dependent, and four rows do not overflow.** Confirmed, and your mechanism is the right one. I walked the months at 1280x500 rather than reasoning from the arithmetic:

| month | rows | cells | overflow |
| --- | --- | --- | --- |
| 2026-11 | 6 | 42 | 121 |
| 2026-09 | 5 | 35 | 40 |
| **2027-02** | **4** | **28** | **0** |

The step is exactly 81, so the cell minimum plus the gap is what sets the grid height, not the `fr` share. My "changes cell height and never total height" was wrong and the description is corrected.

The test now navigates to a six-row month instead of taking the clock's, which also puts the margin back to 121. One thing worth flagging: my first version of that loop re-read the row count before Angular had re-rendered and failed about one run in four. It now waits for the grid to actually advance (`expect.poll` on the first cell's `data-day`), and I ran it three times plus the two suites to be sure. Deleting the reset branch still fails it with `Received: 121`.

I also checked `ci.yml` myself rather than take the correction on trust: it triggers on `pull_request` and the `Tests` job runs `npm run e2e:ci`. You were right the second time.

**Nit, the collapse animation.** Real. You asked whether it holds on a device, so I put it on a short viewport before touching it. At 400x400 collapsing from index 2:

```
                rows    transform    window
expanded        29px    0            174
mid-collapse    29px    -80px        66     <- week 3, not week 2
after           40px    -80px        40     <- week 2
```

So the window sits on the wrong week for the 200ms and then jumps, exactly as you derived. The rows now take their post-snap height when the animation starts, and the same sample reads 40px throughout with no jump at the end.

**Nit, the collapsed clamp.** Also real, and the doc comment did only cover the expanded branch. Took your first form, `isExpanded() ? Math.min(…) : ROW_HEIGHT`, rather than `Math.max(0, …)`: a fraction of a row is not a useful collapsed strip, so the floor is one row rather than a smaller wrong number. `minHeight: 240` is where you said it is. The spec covers 18px and negative room.

**Numbers.** Full suite 14648 passing, 0 failing, and 14634 on the `America/Los_Angeles` run. Lint 0 errors. Schedule and planner e2e 32/32.

Sabotage, one revert at a time:

| Reverted | Fails | Still passes |
| --- | --- | --- |
| the post-snap row-height write | `should give the rows their post-snap height when the animation starts`, `should give the rows the shrunken height when expanding` | the other 45 handler specs |
| the collapsed floor | `should keep a collapsed row at full height when the room runs out` | the other 49 nav specs |
| the scroll reset branch | the e2e, `Received: 121` | n/a |

johannesjo:

Thanks for the updates, and for chasing the e2e one down rather than pinning the assertion around it. Re-checked the 3 items from the earlier review at `bb938b15`, and all three land:

- ✅ **The `overflow > 0` precondition was month-dependent** – fixed in `e2e/tests/schedule/schedule-month-scroll-reset.spec.ts:54`. The month is navigated to rather than taken from the clock, the row count is read off `schedule-month`'s own `--nr-of-weeks` host binding (which `schedule-month.component.spec.ts` pins independently, so the read cannot silently fall through to the inherited `6` on `schedule` and exit the loop at iteration 0), and the loop is followed by a hard `toBe(6)` so a miss fails loudly instead of quietly restoring the old precondition. The 12-click ceiling holds with room to spare: walking `getMonthWeeksToShow` over 2020-2120 for all seven `firstDayOfWeek` values, the worst starting month is 10 steps from a six-row one. And `selectTimeView` writes only `layoutService.selectedTimeView`, never `_selectedDate`, so the navigated month does survive the round trip the way your comment says.
- ✅ **The collapse animation counted rows at the wrong height** – fixed in `planner-calendar-gesture-handler.ts:104`. The write lands before both the `snapDur === 0` and the animated branch, and `expanded ? getRowHeight() : ROW_HEIGHT` is exactly what `displayRowHeight()` resolves to after the flip.
- ✅ **`maxHeight`'s clamp covered the collapsed branch** – fixed in `planner-calendar-nav.component.ts:221`, and the doc comment now describes the collapsed branch instead of only the expanded one.

The one thing below is new, introduced by this commit rather than left over from the last round.

### Should fix before merge

- **`planner-calendar-gesture-handler.ts:104`** – the new `--row-height` write is the only inline style `snapTo` sets that is *not* re-asserted in the `snapDur + 10` callback, and the comment right above that block (`:139-141`) says why the others are: Angular skips the DOM write when the bound value has not changed, so an inline value it did not write can be stranded there.

  It strands like this. Collapsed with `_availableForRows` 178, so `rowHeight()` is 29 and Angular's last written `--row-height` is `40px` (`displayRowHeight()` is `ROW_HEIGHT` while collapsed). `snapTo(true)` writes `29px` inline. If the box grows during the 200ms so that `_availableForRows` reaches 240 or more – a soft keyboard dismissing, a banner going away, a rotation – `rowHeight()` becomes exactly `ROW_HEIGHT`. `isExpanded` is still false at that point, so `displayRowHeight()` is still 40 and Angular writes nothing; then at `snapDur + 10` the flag flips and `displayRowHeight()` is `rowHeight()`, which is *also* 40, equal to what Angular last wrote, so it skips again. The element keeps `29px`. The result is an expanded grid with `max-height` 240 over 174px of rows: a dead strip at the bottom and 25px tap targets, and it stays that way until the next `rowHeight()` change or a collapse.

  Cheapest fix is to follow the pattern already in that block – add the same write to the callback, re-reading it rather than reusing the captured value:

  ```ts
  weeksEl.style.setProperty(
    '--row-height',
    `${expanded ? this._cb.getRowHeight() : ROW_HEIGHT}px`,
  );
  ```

  next to the `maxHeight`/`transform` re-assert at `:142-147`, before `onExpandChanged`. Clearing the property instead would not work: Angular still thinks it wrote it, so nothing re-applies it.

### Nits

- **`planner-calendar-gesture-handler.ts:104`** – the same write mirrors the artifact you removed from the collapse onto the *start* of the expand, which was correct before. Collapsed at `activeWeekIndex` 5 with `rowHeight()` 29: the strip is showing week 5 at `translateY(-200px)` over 240px of 40px rows, and the write drops the content to 174px while the transform is still animating from `-200px`, so the 40px window sits past the end of the content and the first frames are blank before it slides to 0. At index 2 it shows week 3/4 instead of week 2. Seeding `innerEl.style.transform` to `-idx * getRowHeight()` when expanding would express the starting offset in the new row height too. Transient either way, only where six 40px rows do not fit, and this is arguably just the trade the earlier round asked for, so it may be one to note rather than fix.

---

MannXo:
Fixed in `930d2a53f`. Thanks for tracing the whole path rather than just flagging the asymmetry, the stranding sequence is what made it obvious.

**`--row-height` not re-asserted in the `snapDur + 10` callback.** Right, and it is the only one of the three inline styles that block sets that was missing, which the comment at `:139-141` already explains the need for. Took your patch as written, re-reading `getRowHeight()` rather than reusing the value captured before the animation.

The regression is in `planner-calendar-gesture-handler.spec.ts`, in the `on a shrunken grid` describe. It snaps to expanded at the shrunken height, then grows the box through the callbacks while the animation is still running, then ticks past `snapDur + 10` and asserts the post-flip value. Backing the fix out fails it with `Expected '24px' to be '40px'`, which is exactly the stranded value your walkthrough predicts, so it is bound to the re-assert and not to the write before the branches.

Full suite 14649 and 14635, lint exit 0 with 0 errors, `checkFile` clean on both files, planner e2e 24/24 and schedule e2e 8/8.

**The transform-seeding nit I have left alone, and I want to be explicit that it is a judgement call rather than an oversight.**

I agree the artifact is real and that this commit moved it from the collapse onto the start of the expand. The reason I stopped is that the suggested seeding is not actually a one-liner. Setting `innerEl.style.transform` and then setting the transition plus the target in the same task gives the browser one style computation, so it sees only the final value and there is no animation from the seeded offset. Making it work needs the forced reflow between the two, the `void innerEl.offsetWidth` pattern this same file already uses in `slideContent` at `:199-200`. That turns it into a real change to the expand animation's start, on the path that every expand takes, to remove a few blank frames in the one case where six 40px rows do not fit.

Given that, and given this PR is seven rounds in and the last three items were each a boundary case created by the previous round's own fix, my read is that it belongs in a follow-up rather than here. Happy to be wrong about that. If you would rather it land in this PR I will do it with the reflow and a test that pins the starting offset, just say so.

johannesjo:
**✅ Ready to merge** – no blocking issues at `930d2a53`. Merge confidence 90% – the six-row grid's layout on short viewports rests on your measurements and green CI rather than on my running it; the delta itself I did verify by running the spec with and without the fix.

Thanks for the updates. Re-checked the 2 open items from the earlier review at `bb938b15`:

- ✅ **`--row-height` was the one inline style not re-asserted after the snap** – fixed in `planner-calendar-gesture-handler.ts:144-151`. The write sits with the `maxHeight`/`transform` re-assert inside the `snapDur + 10` callback and before `onExpandChanged`, and it re-reads `this._cb.getRowHeight()` rather than the value captured at `:104`, which is what closes the stranding path. `snapTo`'s reduced-motion branch needs no equivalent: `measure()`, the write, `onExpandChanged` and `detectChanges` all run in one synchronous task there, so nothing can move in between. The regression test at `planner-calendar-gesture-handler.spec.ts:614` pins it from the right side – it snaps to expanded at 24px rows, grows the box through both spies while the animation is running, then ticks past `snapDur + 10`, and the intermediate assertion holds the pre-branch write at 24px so the final 40px can only come from the re-read. Removing just the new `setProperty` block fails that one test and nothing else.
- ➖ **Seeding the expand's starting transform in the new row height** – left open deliberately, and the reasoning holds. `innerEl`'s transform is already inline from the `weekOffset()` binding, so seeding it and setting the transition plus target in the same task is overwritten before it ever becomes a start value; making it animate needs the `void innerEl.offsetWidth` reflow this file already uses in `slideContent` at `:198-200`, which adds a synchronous layout plus a visible jump from `-idx * 40` to `-idx * 29` on the path every expand takes, to remove a few transient frames in the shrunken-rows case alone. A follow-up is the right place for it.

Nothing new in the delta: the callback still re-asserts `maxHeight` from the pre-animation `targetHeight` while `--row-height` is now fresh, but `maxHeight()`'s bound value changes across every expand flip (collapsed `40` to `maxExpandedHeight`, which `canExpand()` keeps at 144 or more), so Angular writes the correct box in the same synchronous task before any paint.

---

johannesjo:

PR #9468, opened 2026-08-06
fix(gitlab): restore finish-day worklog dialog

### LINKED ISSUE #9466: 🚨 Submitting time logs in gitlab does not work anymore in version 18.16.0
### Version Used

v18.16.0, rpm, flatpak

### Current Behavior

When I click on the "Finish Day" button the gitlab submit work logs dialog does not appear.

on version v18.15.1 it is still working

### Expected Behavior

When I click on the "Finish Day" button a dialog should pop up asking me if I want to submit my spent time to GitLab.

### Steps to Reproduce

1. Setup a GitLab project (enable submit time logs to gitlab )
2. Book time to the project
3. Finish day

### Console Output

```shell

```

### Documentation

iamhaseebn:
## Problem

Fixes #9466.

Since v18.16, worked-on tasks can correctly remain outside the Today list when automatic adding is disabled or they already have another due date. The GitLab Finish Day hook only inspected the active context list, so it missed those tasks even though the Today summary includes their tracked time. With no qualifying GitLab task found, the worklog dialog did not open.

## Solution

Read tasks across active projects, then limit the dialog input to GitLab tasks with positive tracked time on the app's logical current day. The date comes from `DateService.todayStr()`, so the configured start-of-next-day offset is respected. This restores tasks outside the active context without including archived-project tasks or old untouched worklogs.

Added positive coverage for a qualifying GitLab task outside the active context and negative coverage ensuring a task with only previous-day time does not open the dialog.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Verification

- `npm run checkFile` on both changed TypeScript files
- Focused effect suite: 2 tests passed
- Full default-timezone application suite: 14,062 tests passed, 2 skipped
- Full Los Angeles timezone suite: 14,048 tests passed, 16 skipped
- Package suites: 765 tests passed with type checks
- Repository lint and release-note/tooling tests passed
- Production frontend build and all 23 package builds passed during the initial verification

## Checklist

- [x] Documentation is not needed because this restores the existing Finish Day behavior
- [x] Existing relevant tests pass

### DIFF
--- a/src/app/features/issue/providers/gitlab/gitlab-issue.effects.spec.ts
+++ b/src/app/features/issue/providers/gitlab/gitlab-issue.effects.spec.ts
@@ -1,0 +1,121 @@
+import { TestBed } from '@angular/core/testing';
+import { MatDialog } from '@angular/material/dialog';
+import { MockStore, provideMockStore } from '@ngrx/store/testing';
+import { of } from 'rxjs';
+import { BeforeFinishDayAction } from '../../../before-finish-day/before-finish-day.model';
+import { BeforeFinishDayService } from '../../../before-finish-day/before-finish-day.service';
+import { DateService } from '../../../../core/date/date.service';
+import { DEFAULT_TASK, Task } from '../../../tasks/task.model';
+import { selectAllTasksInActiveProjects } from '../../../tasks/store/task.selectors';
+import { IssueProviderService } from '../../issue-provider.service';
+import { GITLAB_TYPE } from '../../issue.const';
+import { DEFAULT_GITLAB_CFG } from './gitlab.const';
+import { GitlabIssueEffects } from './gitlab-issue.effects';
+
+const DAY = '2026-08-06';
+const PREVIOUS_DAY = '2026-08-05';
+
+describe('GitlabIssueEffects', () => {
+  let store: MockStore;
+  let beforeFinishDayAction: BeforeFinishDayAction;
+  let matDialog: jasmine.SpyObj<MatDialog>;
+  let dateService: jasmine.SpyObj<DateService>;
+
+  const createTask = (id: string, partial: Partial<Task> = {}): Task => ({
+    ...DEFAULT_TASK,
+    id,
+    title: `Task ${id}`,
+    projectId: 'project-1',
+    created: Date.now(),
+    ...partial,
+  });
+
+  beforeEach(() => {
+    matDialog = jasmine.createSpyObj<MatDialog>('MatDialog', ['open']);
+    matDialog.open.and.returnValue({
+      afterClosed: () => of(undefined),
+    } as ReturnType<MatDialog['open']>);
+    dateService = jasmine.createSpyObj<DateService>('DateService', ['todayStr']);
+    dateService.todayStr.and.returnValue(DAY);
+
+    const beforeFinishDayService = jasmine.createSpyObj<BeforeFinishDayService>(
+      'BeforeFinishDayService',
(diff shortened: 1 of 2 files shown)


johannesjo:
Thanks for digging into this, and for the first-PR effort on tests — the diagnosis is right. For the record, the regression is a5499f9930 (`fix(tasks): honor auto-add setting for started tasks`, #9249), which is in v18.16.0 but not v18.15.1, matching the report exactly. It added `!!currentTask.dueDay || typeof currentTask.dueWithTime === 'number'` to the auto-add guard, so a GitLab task with any due date — including an overdue one — no longer enters the Today list when you start it, and the Finish Day hook found nothing.

I can't merge this as-is though, because `selectAllTasks` is much wider than the list it replaces, and the dialog on the other end is not safe to over-feed.

**The blocker:** the dialog has no per-task opt-out. `tmpTasks$` computes `timeToSubmit` as the sum of unsubmitted time across *all* days, and `submit()` posts every task above the 60s threshold — the inline input only lets you edit a value, not deselect a row. So with an unfiltered `selectAllTasks`, any GitLab task anywhere in the store with old unsubmitted time is pre-selected, and one click uploads months of history as GitLab time entries. That's an irreversible write to an external system, from a dialog the user opened for today. The old context list happened to bound this; removing the bound without adding a filter makes it a real hazard.

Two smaller consequences of the same widening: the dialog now opens on every Finish Day for anyone with GitLab time tracking on, even when they did no GitLab work; and `_loadAlreadyTrackedData()` fires one `/time_stats` GET per task passed in — the full set, not the 60s subset — unbounded in a single `Promise.all`, which will hit rate limits on a large backlog.

**Suggested fix** — keep the new source but scope it to what the hook is actually about:

```ts
const todayStr = this._dateService.todayStr();
const tasks = await this._store
  .select(selectAllTasksInActiveProjects)
  .pipe(first())
  .toPromise();
const gitlabTasks = tasks.filter(
  (t) => t.issueType === GITLAB_TYPE && (t.timeSpentOnDay?.[todayStr] ?? 0) > 0,
);
```

`selectAllTasksInActiveProjects` keeps archived projects out, and the day must come from `DateService` rather than `new Date()` — the app runs on a logical day with a configurable start-of-next-day offset. Tasks that qualify still get their past unsubmitted time submitted (that catch-up is intentional, and flagged via `isPastTrackedData`), so nothing is lost — it just can't be triggered by tasks you haven't touched today.

**On the test:** it only covers the positive path, so the widening above is unpinned in both directions. Worth adding a case asserting a GitLab task with no time tracked today is *not* in `tasksForIssueProvider`. Also, `const DAY = '2026-08-06'` is a fixed literal unrelated to whatever the code considers today — once a day filter exists it needs to come from the same stubbed `DateService`, otherwise the assertion passes vacuously.

**Minor:** the `WorkContextService` provider in the spec is dead now that the effect doesn't inject it.

Happy to take this with those changes.

iamhaseebn:
Thanks for catching that. I updated the branch to use `selectAllTasksInActiveProjects`, derive the logical day from `DateService.todayStr()`, and include only GitLab tasks with positive time tracked today. The obsolete `WorkContextService` test provider is removed, and there is now negative coverage for a task with only previous-day time. The focused tests and the full pre-push lint, package, default-timezone, and Los Angeles timezone suites all pass.

johannesjo:

PR #9479, opened 2026-08-06
feat(community-plugins): add JSON to Tasks plugin

nekoritsme:
## Problem

Super Productivity currently has no simple way to create multiple tasks from structured JSON generated by external tools such as ChatGPT.

## Solution

Add the **JSON to Tasks** plugin to the community plugins list.

The plugin allows users to paste a single JSON task object or an array of tasks and create them in Super Productivity.

It supports:

* tasks and subtasks
* notes
* time estimates
* due dates
* projects and tags by name
* optional creation of missing projects and tags

Plugin repository:
https://github.com/nekoritsme/superproductivity-fromjson-totasks

## Type of Change

* [ ] Bug fix
* [x] New feature
* [ ] Breaking change
* [ ] Refactoring
* [ ] Documentation
* [ ] Other (please describe)

## Checklist

* [ ] I have run `npm run checkFile` on changed `.ts`/`.scss` files

### DIFF
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -142,7 +142,7 @@
     "author": "Fernando Pindado",
     "authorUrl": "https://codeberg.org/fpindado",
     "stars": 0
-  }
+  },
   {
     "name": "JSON to Tasks",
     "shortDescription": "Create tasks and subtasks from pasted JSON, with project/tag resolution and optional creation of missing projects and tags.",
@@ -150,5 +150,5 @@
     "author": "nekoritsme",
     "authorUrl": "https://github.com/nekoritsme",
     "stars": 0
-}
+  }
 ]
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -143,4 +143,12 @@
     "authorUrl": "https://codeberg.org/fpindado",
     "stars": 0
   }
+  {
+    "name": "JSON to Tasks",
+    "shortDescription": "Create tasks and subtasks from pasted JSON, with project/tag resolution and optional creation of missing projects and tags.",
+    "url": "https://github.com/nekoritsme/superproductivity-fromjson-totasks",
+    "author": "nekoritsme",
+    "authorUrl": "https://github.com/nekoritsme",
+    "stars": 0
+}
 ]


johannesjo (commented):
Thanks for the submission! The registry entry itself is fine — fields match the `CommunityPlugin` interface, no duplicate name/URL, prettier-clean, and `stars: 0` matches. Nothing here blocks the listing.

I read through the plugin source as part of reviewing this. Good news first: it declares neither `nodeExecution` nor `http` (the only two permission strings SP actually enforces), and `index.html` contains no `fetch`/XHR/WebSocket/`eval` at all — so it's fully offline, which is exactly what I want to see for a paste-JSON tool. Rendering goes through `textContent` throughout, and `isTodayAlias()` warning instead of writing `TODAY` into `tagIds` shows you read the docs properly.

Reading it did surface a few things.

### Two of these are our bugs, not yours

The plugin API is inconsistent with the local REST API here; I'll track and fix these on our side:

1. **`PluginAPI.addTask()` silently drops `dueDay` for subtasks.** The subtask branch in `plugin-bridge.service.ts:917-927` hand-rolls `createNewTaskWithDefaults()` without `dueDay`, while `TaskService.addSubTaskTo()` (`task.service.ts:815`) explicitly supports it. So `dueDay` on a subtask works via REST but vanishes via the plugin API — even though `PluginCreateTaskData.dueDay` (`packages/plugin-api/src/types.ts:807`) advertises it.
2. **`PluginAPI.addTask()` has no nested-subtask guard.** `_validateTaskReferences()` (`plugin-bridge.service.ts:1738`) only checks that the parent exists, and the `addSubTask` reducer (`task.reducer.ts:477`) doesn't check depth — so passing a subtask's id as `parentId` writes a 3-level tree into state, which the model doesn't support. The REST API rejects exactly this with `INVALID_PARENT` / "Cannot nest subtasks" (`local-rest-api-handler.service.ts:438`); the plugin path should behave the same.

### Four for the plugin, if you want to polish it

3. **Unlimited recursion depth.** `normalizeTask()` and `createTree()` recurse without a depth limit, but SP supports exactly two levels. Until (2) is fixed on our side there's nothing stopping it, so 3-level JSON — a very plausible thing to get back from ChatGPT — will create malformed tasks. Worth flattening or rejecting depth > 2 in the plugin.
4. **`toDateOnly()` is off by one day east of UTC.** `YYYY-MM-DD` takes the regex fast path and is correct, but anything else goes through `new Date(s).toISOString().slice(0, 10)`, which converts local midnight to UTC and lands on the previous day for any positive offset. `"date": "Aug 8 2026"` and `"08/08/2026"` both become `2026-08-07` in Berlin/Kyiv/Tokyo (correct in New York).
5. **`tags` on subtasks are resolved, validated and previewed, then dropped.** Subtasks inherit tags and project from the parent by design (`tagIds: []` in the bridge; the REST API returns an explicit error rather than dropping). Your README example has `tags: ["Review"]` on a subtask, so it currently promises something that silently doesn't happen — a preview warning would be clearer.
6. **Partial import on failure.** If the host rejects a task mid-run, `createTree()` throws after N tasks already exist, and any projects/tags created up front stay behind, while the snackbar only says "Failed to create tasks". Worth reporting how far it got.

Items 3–6 are all fixable on your side and none of them block this PR.

nekoritsme:
Thanks for the thorough review! I’ll take a look at items 3–6 for a follow-up release.

nekoritsme:
Thanks again for the review! I’ve released v1.2.0 addressing items 3-6:

- reject nested subtasks deeper than one level;
- avoid UTC date shifts when parsing dates;
- warn when project/tags are supplied for subtasks, since they’re inherited from the parent;
- report how many tasks were created if an import fails partway through.

I also updated the README, including a note that subtask dueDay requires Super Productivity 18.19.0+ because the related Plugin API fix landed there.

If there’s anything else needed from my side before the listing can be merged, let me know. Thanks!

johannesjo:
Thank you very much! <3

Issue #9342, opened 2026-07-28
The local storage position cannot be changed💥


zhailei-scu:
### Steps to Reproduce

1.Setting
2.Sync & Backup
3.Automatic Backups
4.Click the folder location path
5.The File Exploer can be opened, but i cannot set the new location, none of selection button is provided.

zhailei-scu:
The superproductively version i am using is 18.15.1W, thanks!

PR #9482, opened 2026-08-06
feat(electron): add a configurable folder for automatic backups

### LINKED ISSUE #7054: 💡 Add option to change the directory for automatic backups (Desktop version)
### Problem Statement

Currently, automatic backups in the Desktop version are always saved to a fixed hardcoded folder:

- Windows: `C:\Users\<User>\AppData\Roaming\superProductivity\backups`
- macOS / Linux: corresponding default user data folders + `/backups`

There is no built-in setting to change this backup directory.  
The only current workarounds are:
- Using the `--user-data-dir` launch parameter (which moves the entire app data, not just backups)
- Manually backing up the folder with third-party software

This is inconvenient for users who want to save automatic backups directly to a cloud-synced folder (OneDrive, Google Drive, Dropbox, Syncthing, external drive, NAS, etc.) without moving the whole application data or using additional backup tools.

### Possible Solution

Add a new setting in **Settings → Sync & Export → Automatic Backups**:

- "Backup folder" (or "Custom backup directory")
- Button to choose folder (with default value being the current one)
- The app should save all automatic backups to the selected folder instead of the hardcoded `.../backups` subfolder

This would make automatic backups truly user-friendly and flexible.

- `--user-data-dir` parameter → moves all data (settings, tasks, etc.), not ideal if you only want to change backup location.
- Manual export every time → defeats the purpose of automatic backups.
- Using external backup software → adds complexity and another point of failure.

### Additional context

This feature is especially useful for:
- Users who sync backups to cloud storage
- People keeping important data on external drives or NAS
- Those who want automatic backups in a specific folder without affecting the main app data location

It would greatly improve data safety and convenience for many desktop users.

AlvaroBalbin:
Fixes #7054.

Desktop users can now choose where automatic backups are stored from Settings > Automatic Backups. Leaving the setting empty uses the normal `<userData>/backups` folder. Picking the default folder again clears the custom setting. Existing backups are not moved, and restore only scans the active folder.

The folder path is owned by the main process so renderer plugins cannot redirect backup writes. The picker rejects folders inside the app's private directory and folders that fail a real write probe. If a custom folder later disappears, the app fails the backup instead of recreating the path on the wrong disk.

Backup discovery and restore accept only automatic backup names in custom folders. The default folder remains compatible with existing `.json` backups, while retention only deletes current timestamp-shaped automatic backups. Writes use a temporary file followed by a rename, and cleanup removes only stale temporary files created by the app.

The Windows Store display path and reset-to-default behavior are preserved. The settings copy and wiki documentation are updated as well.

Validation:

- `node --test electron/backup.test.cjs`: 38 tests passed
- `npm run checkFile -- <file>`: passed for every changed TypeScript file
- `npm run test:electron`: 324 tests passed; 5 unrelated symlink tests could not run because this Windows environment does not have permission to create symlinks


johannesjo (changes requested):
Thanks for taking this on — and for putting the path in the main-owned `SimpleStoreKey` store rather than `globalConfig`. That's the right call for a device-specific value and it's the part I'd have expected to go wrong.

Two blockers, and one framing point.

## Retarget the issue

#9342 isn't a bug. Clicking the path opens the folder, which is exactly what that `<a href="file://…">` is for; the reporter expected a picker and filed the surprise as a crash. The feature you're actually building is **#7054** — "Add option to change the directory for automatic backups". Worth relinking, because it changes how the PR is judged: not "fix broken behaviour" but "add a configurable location", which is a fair ask with modest but real demand.

## Blocking 1 — the app deletes the user's own files in the folder they pick

`cleanupOldBackups` (`electron/backup.ts:224`) picks deletion candidates with `.endsWith('.json')`. That was safe while the directory was app-owned and held nothing but backups. Once the user chooses the folder, every unrelated `.json` in it becomes a candidate — and pointing this at an existing Syncthing / Dropbox / NAS folder is the entire point of #7054.

Retention keeps the 30 newest by mtime plus up to 21 date-prefixed dailies. The user's files are older than the accumulated backups and match no `YYYY-MM-DD` prefix, so they all fall out. Reproduced against this branch using the harness from `backup.test.cjs` (fs mocked, default settings, no `maxBackupFiles` override):

- folder with 40 user `.json` files, no backups yet → **11 deleted** over the first 15 backups
- folder with 40 user `.json` files plus the normal 30-backup steady state → **all 40 deleted in a single backup cycle**

## Blocking 2 — one foreign file hides the real backup

`BACKUP_IS_AVAILABLE` (`electron/backup.ts:106`) has no extension filter at all: it stats every dirent and returns `.reverse()[0]`, the newest file of any kind. Drop a `holiday-photo.png` into the picked folder and that is what the restore prompt offers. `_importBackup` then fails the `JSON.parse` and shows "import failed". Verified.

Because only the single newest entry is ever returned, the user's actual backup becomes unreachable through that prompt — recovery quietly stops working at the one moment it matters.

## Both fall to a single predicate

Backup filenames already have a fixed shape from `getBackupTimestamp()`:

```ts
// electron/shared-with-frontend/get-backup-timestamp.ts
export const isAutoBackupFilename = (f: string): boolean =>
  /^\d{4}-\d{2}-\d{2}_\d{6}\.json$/.test(f);
```

Used in three places:

- `cleanupOldBackups` — replace `.endsWith('.json')`
- `BACKUP_IS_AVAILABLE` — filter `readdirSync` before mapping
- `BACKUP_LOAD_DATA` — add `isAutoBackupFilename(path.basename(backupPath))` next to the existing dir check

The third isn't needed for the two bugs above, but it's one line and it earns it. The allow-list at `electron/backup.ts:140` now includes a user-chosen root at any depth and any extension, and `window.ea` is reachable from plugin code (`plugin-runner.ts` executes plugins with `new Function` in the main world). If someone picks their home folder, `loadBackupData` turns into a read primitive over all of it. Restricting it to backup-shaped basenames closes that without adding any machinery — no extra gating or consent flow needed.

Please add a regression test for the retention case, one that fails without the filter. Also add `unlinkSync` to the fs mock in `backup.test.cjs` — as it stands, a test that reaches cleanup would call the real one.

## Smaller

- **A rejected pick is invisible.** `assertPathOutside` throws, `ConfigSectionComponent.onAction` catches and only `Log.err`s. Pick a folder inside AppData and the button simply does nothing. Needs a snack.
- **No way back to the default** except navigating to the hidden `<userData>/backups`. A second action, or treating a cancelled pick as reset, would cover it.
- **Existing backups stay behind** and vanish from the restore prompt, since only the new folder is scanned. The `electronAPI.d.ts` JSDoc notes this; the user never sees it. One line of UI text.
- `getCustomBackupDir` and `getBackupDir` can be one function.
- `package-lock.json` — the `"dev": true` on `fsevents` is unrelated churn, please drop it.
- Wiki: a new setting belongs in `docs/wiki/3.02-Settings-and-Preferences.md`, the location change in `3.06-User-Data.md`.

## Otherwise

The shape is right — main-owned rather than synced, the Windows-Store redirection correctly narrowed to the default dir only, and genuine coverage of the new paths rather than just the happy one. `npm run test:electron` is green (257 passing), lint and `tsc -p electron/tsconfig.electron.json` are clean on the diff. Fix the filename filter and this is close.

AlvaroBalbin:
Thanks for the thorough review, and for catching both of those. You're right that #7054 is the real target; I'll relink the PR to it.

I added `isAutoBackupFilename` to get-backup-timestamp.ts matching your regex and used it in all three places: `cleanupOldBackups` now filters on it instead of `.endsWith('.json')`, `BACKUP_IS_AVAILABLE` filters `readdirSync` before mapping, and `BACKUP_LOAD_DATA` checks the basename alongside the existing directory check. I added a regression test that reproduces the retention bug (a foreign file and two real backups with distinct mtimes, asserting the foreign one is never touched by `unlinkSync`), plus one for `BACKUP_IS_AVAILABLE` no longer returning a non-backup file and one for `BACKUP_LOAD_DATA` refusing to read one, and wired `unlinkSync` into the fs mock so cleanup tests actually exercise the real function.

For the smaller points: a rejected folder pick now shows an error snack instead of silently doing nothing; I merged `getCustomBackupDir` into `getBackupDir`; added a line of UI text noting that changing the folder doesn't move existing backups and that only the current folder is scanned for restore; dropped the stray `"dev": true` on the prisma fsevents entry in package-lock.json; and updated both wiki pages for the new setting and location behavior.

I left out the "reset to default without navigating to the hidden folder" idea for now since you offered it as one option among a couple, and picking the default folder again already works as a reset once you get there - happy to add a dedicated action if you'd still like one.

johannesjo (commented):
Thanks — both blockers are properly fixed, and the tests are real ones.

I mutation-tested each of the three new regressions on `e54baeaee7`:

| Mutation | Test that fails |
| --- | --- |
| `cleanupOldBackups` filter back to `.endsWith('.json')` | `cleanup never deletes non-backup-shaped files, even when they are the oldest` |
| `BACKUP_IS_AVAILABLE` filter removed | `a non-backup file in the folder does not hide the real backup from restore` |
| `BACKUP_LOAD_DATA` basename check removed | `refuses to load a non-backup file even inside the picked folder` |

Each fails without its fix and passes with it — not decorative coverage. `unlinkSync` is wired into the fs mock, so the cleanup tests exercise the real function rather than skipping past it.

`npm run test:electron` 260/260. `npm run checkFile` clean on all seven changed `.ts` files. `tsc -p electron/tsconfig.electron.json` clean on the diff. Merges cleanly with current master. Wiki edits land on the two notes the documentation guide maps to (3.02, 3.06); 3.05 already covers desktop-only backups generically, so nothing needed there. I also checked the regex boundaries directly — it rejects a trailing newline (JS `$` is strict without `m`), `.JSON`, prefix/suffix padding and unpadded components — and `assertPathOutside(app.getPath('userData'), …)` is exactly the sync-folder precedent, since `getAppPrivateDir()` is that same call.

## Still open

### 1. The PR still says `Fixes #9342`

This was the one thing we agreed in writing and it didn't happen. As it stands, merging closes the non-bug and leaves #7054 — the actual feature, with the actual demand — open. The code comments have the same problem: `simple-store.const.ts:9` still cites #9342, and three `backup.test.cjs` comments plus the `isAutoBackupFilename` JSDoc cite #9482, which is this PR. Someone reading `git blame` in two years lands on a PR, not on a bug report. Please point all of them at #7054.

### 2. An unwritable picked folder fails silently, forever

`backupData` catches everything and only logs. So the pick succeeds, the UI shows the new path, and backups stop — with nothing the user can ever see.

That was fine while the directory was app-owned and known-writable. Once the user picks it, several ordinary choices land there:

- **Snap** is `confinement: strict` with `home` + `removable-media` (electron-builder.yaml). `home` excludes dot-directories, so `~/.backups` is EACCES, as is anything outside home and removable media.
- **Flatpak** has `--filesystem=home` only — no `/media`, no `/mnt`, no network mounts.
- **macOS** prompts for Documents/Desktop/Downloads on first *access*, which is the backup timer minutes later rather than the pick, so the prompt arrives detached from the action that caused it and a dismissal is permanent.
- Read-only folders and disconnected network shares, on every platform.

It also quietly breaks the invariant `_recordLastBackupTime` documents. `_backupElectron` awaits `backupAppData`, main resolves regardless, so `didWrite = true` unconditionally and the "last backup" timestamp advances on a write that never happened — exactly the false "you're protected" the comment at `local-backup.service.ts:349` says must not happen. No desktop UI surfaces it today (`_lastBackupInfo()` is only passed on the Android branch), so it is latent rather than live, but it is the same failure wearing a different hat.

A write probe at pick time — write and unlink a temp file inside `picked`, before `saveSimpleStore` — covers all of the above in one place and turns a silent, permanent loss of the backup safety net into an immediate error. The catch in `ConfigPageComponent` already routes any rejection to a snack, so nothing else needs building; `S_FOLDER_NOT_ALLOWED` would just need generalising, since "pick a location outside the app's data folder" is the wrong advice for a permission failure.

For the one feature whose entire job is being the last line of defence, I'd want this before merge.

### 3. `mkdirSync(backupDir, { recursive: true })` re-creates a folder that has gone away

`recursive: true` is new here, and it buys nothing: userData always exists for the default dir, and the picker uses `createDirectory` without `promptToCreate`, so a picked folder always exists at pick time. The only case it can ever fire is the one where creating the folder is the wrong move.

Pick `~/Dropbox/SP-Backups`, then uninstall or relocate Dropbox: home is writable, so the next backup silently resurrects `~/Dropbox/SP-Backups` and fills it with backups nothing syncs. On macOS the same thing under `/Volumes` after an eject leaves a phantom directory sitting on the mount point, so the drive comes back as `NAME 1` and the real backup set is somewhere the app no longer looks.

Dropping `recursive: true` restores the previous semantics and lets a missing parent fail loudly into the existing catch.

### 4. Windows Store can refuse the path the app itself displayed

`getBackupDirForDisplay()` shows `BACKUP_DIR_WINSTORE` on a virtualized Store install, but the reset-to-default check only matches `path.resolve(BACKUP_DIR)`. So a user who picks the folder we just showed them either trips `assertPathOutside` and gets "please pick a location outside the app's data folder" about our own displayed path, or silently pins a custom folder to LocalCache. One line — treat `BACKUP_DIR_WINSTORE` as reset-to-default too.

### 5. No `defaultPath` on the dialog

Leaving out the dedicated reset action is defensible only if "pick the default again" is actually reachable, and right now it isn't: the picker opens wherever the OS decides, and `<userData>/backups` is a hidden path the user has to hunt down — and may not exist yet on a fresh install. `defaultPath: await getBackupDirForDisplay()` opens the picker in the current backup folder, which makes both "where am I now" and "put it back" work. I'd take this over a second action.

## Smaller

- **`BACKUP_DIR` in the `BACKUP_LOAD_DATA` allow-list is now dead surface.** With no custom folder, `backupDir === BACKUP_DIR` already covers it. With one, nothing ever produces a `BACKUP_DIR` path — the sole caller (`local-backup.service.ts:213`) passes `backupMeta.path`, which `BACKUP_IS_AVAILABLE` builds from the live dir. So the clause can only ever widen the read surface for a caller that isn't the legitimate one. Drop it.
- **The caveat line ships to every desktop user, forever.** `LOCATION_CHANGE_INFO` is appended whenever `backupPath` is set, i.e. on every desktop install, including the overwhelming majority who never change the folder. That is permanent settings clutter for a caveat that only matters once you have changed it. Fold it into the section's existing `help`, or pass it only when a custom folder is actually active — main already knows (`backupDir !== BACKUP_DIR`).
- `isAutoBackupFilename` is now the security predicate for `BACKUP_LOAD_DATA` and the guard standing between cleanup and the user's own files, but it has no direct unit test — and it lives in a file that already has a spec (`src/app/util/get-backup-timestamp.spec.ts`). Cheap to pin the boundaries there.
- `writeFileSync(\`${backupDir}/${getBackupTimestamp()}.json\`)` is string concatenation while every other join in the file uses `path.join`. Now that `backupDir` is user-supplied this produces mixed separators on Windows.
- `getBackupDir().then(…)` in `initBackupAdapter` is a floating promise, safe today only because `loadSimpleStoreAll` can never reject. `void` it.

## Otherwise

The core is in good shape. Both filters are in the right place, the tests earn their keep, `getBackupDirForDisplay` correctly narrows the Store redirection to the default dir only, and rebuilding the form section from `GLOBAL_IMEX_FORM_CONFIG` rather than appending means repeated picks no longer duplicate the section. Points 1 and 2 are what I'd block on; the rest are small.

AlvaroBalbin:
Retargeted, and both blockers are addressed.

## 1. Retargeted to #7054

The PR body now says `Fixes #7054` and the title says what this actually is. The code references moved too: `simple-store.const.ts`, the `isAutoBackupFilename` JSDoc, and the four `backup.test.cjs` comments that pointed at this PR all point at the issue now.

## 2. Write probe at pick time

`assertFolderWritable` writes and removes a probe file inside the picked folder, after `assertPathOutside` and before `saveSimpleStore`, so a folder that cannot be written to is refused at the pick instead of quietly ending automatic backups. It throws a path-free `FolderNotWritableError`, which keeps the two pick-time refusals apart in the log while both land on the same snack.

A probe write rather than `access()`: a folder can be listable and still refuse writes, and the macOS prompt only answers on a real access.

`S_FOLDER_NOT_ALLOWED` is now `S_FOLDER_NOT_USABLE` and covers both causes: "Couldn't use that folder for backups. Please pick a folder outside the app's data folder that the app is allowed to write to."

I left the `_recordLastBackupTime` invariant alone. Making `backupData` reject would push write failures back into the renderer's backup interval, which the comment there explicitly does not want, and as you say nothing surfaces it on desktop today. Happy to take it as a follow-up, probably alongside wiring `_lastBackupInfo()` up on desktop.

## 3. `recursive: true` dropped

Back to a plain `mkdirSync`, so a missing parent fails into the existing catch.

## 4. Windows Store default folder

The reset-to-default check now also matches `BACKUP_DIR_WINSTORE` on a Store build, so picking back the exact path the app displayed clears the custom folder instead of being refused or pinning backups to LocalCache.

## 5. `defaultPath`

The picker opens at `getBackupDirForDisplay()`. No dedicated reset action, per your preference.

## Smaller

- `BACKUP_DIR` dropped from the `BACKUP_LOAD_DATA` allow-list. I narrowed `BACKUP_DIR_WINSTORE` the same way, to the default-dir case, matching the narrowing in `getBackupDirForDisplay`: with a custom folder active nothing produces a Store path either, so keeping it could only widen the read surface for a caller that is not the legitimate one.
- The caveat line is out of the section body and folded into the section help, so it only appears where the folder can be changed at all and is not permanent clutter for the users who never change it.
- `isAutoBackupFilename` has its own spec in `src/app/util/get-backup-timestamp.spec.ts`: accepts what `getBackupTimestamp` produces, rejects near misses, padding, case variants and a trailing newline.
- `path.join` for the backup file path, and `void` on the `getBackupDir()` promise in `initBackupAdapter`.

## The six new regressions, mutation-tested

| Mutation | Test that fails |
| --- | --- |
| drop `assertFolderWritable(picked)` | `refuses a folder that cannot be written to`, `the write probe leaves no file behind in the picked folder` |
| `mkdirSync(backupDir, { recursive: true })` | `a picked folder whose parent has gone away is not re-created` |
| drop the Store clause from the reset check | `Store builds treat the displayed LocalCache path as the default folder` |
| drop `defaultPath` | `the picker opens in the folder backups currently go to` |
| re-add `BACKUP_DIR` to the allow-list | `refuses a default-dir path once a custom folder is in use` |

Each mutation fails only its own test.

The fs mock now models the two behaviours these depend on: a non-recursive `mkdirSync` throws ENOENT when the parent is missing, and `writeFileSync` can be made to fail for a chosen folder. `beforeEach` seeds the parent folders that always exist in reality, so a test that means "the folder is gone" has to say so explicitly.

## Verification

`electron/backup.test.cjs` 19/19. Full `npm run test:electron` 289/294: the five failures are the symlink cases in `file-path-guard` and `sync-path-resolver`, which fail identically on pristine master here because this machine refuses symlink creation (EPERM). Angular unit suite 14278/14279: the one failure is `caldav-client.service.spec.ts` "connects when the server omits principal-collection-set", which also fails on pristine master. `npm run checkFile` clean on every changed file, `npm run int:test` 14/14, and `tsc -p electron/tsconfig.electron.json` plus `tsc -p tsconfig.json` both clean. Rebased onto current master.

johannesjo:
Reviewed the current head (269e9dc). Tests pass here: `node --test electron/backup.test.cjs` 19/19, `config-page.component.spec.ts` 6/6, `checkFile` clean on `electron/backup.ts`. Branch merges cleanly (13 commits behind master).

This is careful work — main-owned key following the #8228 precedent, no new toggle, no dependency, the read allow-list *narrowed* rather than widened (dropping `BACKUP_DIR` once a custom folder is active is the right call), and the test suite is genuinely good. One real regression to fix, one design gap I'd like a decision on, then minors.

## 1. `isAutoBackupFilename` hides pre-v17.1.7 backups from restore

The filter is applied to the **default** folder too, so this hits users who never touch the new feature.

Before c0fbfd14 ("use timestamped filenames with smart retention policy", shipped in **v17.1.7**), the writer produced `YYYY-MM-DD.json` — one file per day, and there was no cleanup routine at all, so those installs accumulated one file per day indefinitely. `isAutoBackupFilename` requires `_HHmmss`, so with this PR those files are filtered out of `BACKUP_IS_AVAILABLE` (backup.ts:154) *and* refused by `BACKUP_LOAD_DATA` (backup.ts:198). On master `BACKUP_IS_AVAILABLE` lists the folder unfiltered and would still offer them.

Population is narrow and shrinking — with backups enabled the first post-upgrade cleanup prunes them, and they're fully gone after ~2.5h of runtime — but two groups still have them today:

- anyone who upgraded past v17.1.7 with automatic backups **disabled**: no writes, so cleanup never runs and the legacy files sit there indefinitely;
- anyone upgrading now from ≤ v17.1.6, for the first hours.

The failure lands at the one moment it matters: `askForFileStoreBackupIfAvailable()` only fires when the store is genuinely blank (`StartupService` line ~262, "only offer to restore from backup if this is truly a fresh install"), i.e. exactly the disaster-recovery path. Those users get silence instead of a restore offer.

Suggested fix, cheap and with no downside: accept `^\d{4}-\d{2}-\d{2}(_\d{6})?\.json$` for **discovery and read only when `backupDir === BACKUP_DIR`**, and keep the strict shape for deletion and for any picked folder. That leaves the picked-folder read surface exactly as narrow as it is now, and keeps cleanup away from a user file that happens to be named `2025-03-14.json`.

## 2. After pick time, nothing can tell the user backups stopped

`assertFolderWritable` proves writability once, at the pick. The headline use cases — removable drive, network share, cloud folder — revoke it routinely, and then the failure is completely invisible:

- `backupData` catches and only logs (by design);
- `LocalBackupService._triggerBackupSave$` catches and only logs;
- `didWrite = true` is set unconditionally for Electron (`local-backup.service.ts:343`) because the IPC resolves either way, so `_recordLastBackupTime()` (:358) records a *successful* backup that never happened;
- desktop shows no last-backup line at all.

So a user who points backups at an external drive and unplugs it has automatic backups silently stop forever. On the default dir this state was essentially unreachable; with a picked folder it is routine. I'd treat this as the condition for the feature being safe rather than as a follow-up.

Minimal version: let `backupData` reject on write failure. The renderer's `catchError` already absorbs it, so nothing is pushed back into the backup interval — the comment you're preserving is about not throwing *into* the interval, which `catchError` handles. That also makes `_recordLastBackupTime` honest, which in turn lets the existing `_lastBackupInfo()` line be shown on desktop as a quiet staleness indicator instead of a new alert.

## 3. A shared cloud folder is the headline use case, and nothing models it

Two desktops pointed at one synced folder is precisely what #7054 asks for, and nothing distinguishes device origin:

- each device's `cleanupOldBackups` deletes the *other* device's backups (they're all backup-shaped), silently halving effective retention;
- the startup restore prompt can offer another device's snapshot — user-confirmed, but the dialog shows only folder + date;
- Dropbox/OneDrive conflict copies never match the regex, so they accumulate forever (harmless, but unbounded);
- `writeFileSync` (backup.ts:281) is not atomic, so a sync client can upload a half-written backup. `simple-store.ts` already does tmp+rename for exactly this reason, and a `.tmp` name wouldn't match the backup regex.

At minimum this deserves a line in the wiki next to the existing caveats; tmp+rename is a two-line improvement worth taking.

## Minors

- **Store the canonical path.** `PICK_DIRECTORY` persists `realpathSync.native(...)` (local-file-sync.ts:354); this persists `path.resolve(...)` (backup.ts:221). Not exploitable today — the allow-list canonicalizes both sides and the filename filter caps what can be read — but canonicalizing closes the symlink-repoint window entirely and matches the neighbouring picker.
- **Folder name reaches `innerHTML`.** `automatic-backups-form.const.ts:57` builds `<a href="file://${backupPath}">${backupPath}</a>`, and `FormlyTranslatedTemplateComponent` assigns it via `innerHTML`. The path is now fully user-chosen and can contain `"` or `<` on Linux/macOS. Self-inflicted only, but worth escaping since it just became user-controlled.
- **Error convention.** `saveSimpleStore` can throw a raw fs error out of the handler (backup.ts:240); `PICK_DIRECTORY` wraps everything in `createSafeIpcError` + `getSafeErrorMeta`. Not a real disclosure (the renderer can already ask for that path via `getBackupPath`), purely a consistency nit.
- **i18n side effect.** Switching desktop to `HELP_DESKTOP`, which re-states `HELP`'s sentence, means every non-English desktop user loses the already-translated paragraph and falls back to English — `de.json` has `AUTO_BACKUPS.HELP` today. Keeping `HELP` and adding only the caveat as a second key avoids that.
- `promptToCreate` is in the sync picker's `properties` but not here; the probe file can be left behind in the user's folder if `unlinkSync` fails (log-only, deliberate); the first commit message ("The local storage position cannot be changed💥") needs a proper `feat(electron): …` on squash.

johannesjo:

No new commits since my 2026-08-21 pass (head is still `269e9dc`), so items 1-3 there are all still open and I won't restate them – items 1 and 2 are what the label is for. CI is green and the branch still merges cleanly, though it is now 208 commits behind `master`, so a rebase is worth doing before merge.

Nothing new turned up beyond those, but two small things to fold in when you next touch this.

### Nits

- **`electron/backup.ts:229` – an addendum to the "store the canonical path" minor, because applying it literally would break the reset.** The reset-to-default branch compares lexically (`picked === path.resolve(BACKUP_DIR)`, with `picked` from `path.resolve` at :221) while `assertPathOutside` nine lines down goes through `canonicalize()` in `file-path-guard.ts`, which resolves symlinks and collapses case / 8.3 aliases. Today the two only disagree if the user picks the default folder through an alias the app never displays (the picker pre-navigates to the un-canonicalized `BACKUP_DIR`, so the path the UI points at always hits the reset branch) – which is why this is a nit rather than a live bug.

  It stops being a nit if the canonicalization minor is applied to one side only. Setting `picked = realpathSync.native(filePaths[0])` while the comparison stays against the raw `path.resolve(BACKUP_DIR)` makes the mismatch deterministic for every install whose userData is reached through a symlink (`~/.config` linked into a dotfiles repo is the common one): the reset branch is then never taken, `assertPathOutside` canonicalizes both sides, sees the folder inside userData and throws, and the user gets "please pick a folder outside the app's data folder" about the folder the app itself displays as the current backup location. Canonicalize both sides instead – `canonicalize()` already handles a not-yet-existing leaf, so exporting it and comparing `canonicalize(picked) === canonicalize(BACKUP_DIR)` (same for the Store path) covers the minor and the reset check together.

- **`src/app/pages/config-page/config-page.component.ts:247`** – `_updateAutomaticBackUpCfg()` has no spec. Its load-bearing property is that it rebuilds from `GLOBAL_IMEX_FORM_CONFIG` instead of appending to `this.globalImexFormCfg`, which is what stops the re-build after a pick from stacking duplicate Automatic Backups sections. That is invisible until someone picks twice, and a later refactor back to `[...this.globalImexFormCfg, ...]` would reintroduce it silently.

  Worth flagging that the obvious approach doesn't work: the Android action test at `config-page.component.spec.ts:116` is testable because that branch is gated on the injectable `IS_ANDROID_WEB_VIEW_TOKEN`, whereas the Electron branch is gated on the module-level `IS_ELECTRON` const (`app.constants.ts:5`, read off `navigator.userAgent`), which TestBed can't override. So either call `_updateAutomaticBackUpCfg()` directly with `window.ea` stubbed, or switch the branch to the existing `IS_ELECTRON_TOKEN` (`app.constants.ts:12`), whose doc comment says it exists for exactly this.

---

johannesjo:
@AlvaroBalbin still keen on this one — the shape is right and I'd like to land it. Two things from my last pass are still open:

1. **`isAutoBackupFilename` also filters the default folder.** Pre-v17.1.7 backups are named `YYYY-MM-DD.json`, so users who never touch this feature lose their restore offer on a blank-store boot. The filter needs to apply to discovery/cleanup in the *configured* folder without narrowing what restore will accept.
2. **Post-pick write failures are invisible forever.** The write probe covers the moment you pick the folder; if it later becomes unwritable (permissions change, drive unmounted, disk full), main catches and logs and the renderer never learns.

Both are small. If you'd rather hand it over, that's completely fine — say so and I'll finish it on top of your commits with credit.

johannesjo:

Thanks for the updates. Re-checked all 10 items from the earlier review against the code at `5814b82` – every one of them lands, including both blockers:

- ✅ **`isAutoBackupFilename` hid legacy `YYYY-MM-DD.json` backups from restore** – `electron/backup.ts:170` accepts them for discovery (`:191`) and read (`:237`) only while the default folder is in use, and cleanup (`:352`) keeps the strict shape, so a user file named `2025-03-14.json` in a picked folder is still never deleted. Four new tests cover both directions.
- ✅ **Post-pick write failures were invisible forever** – `backup.ts:340` rethrows instead of log-and-swallow, which is what makes the renderer's existing `await this._backupElectron(data); didWrite = true;` (`local-backup.service.ts:342`) finally skip `_recordLastBackupTime()` on a failed write; the outer `catchError` (`:83`) still absorbs it, so nothing is pushed back into the interval. `config-page.component.ts:272` passes `_lastBackupInfo()` on desktop too. Worth being explicit about what that buys: the signal is a **Last backup** line that stops advancing, not a notification – a user who does not open Settings still learns nothing. That is the right size for now, I think, but it is the remaining gap.
- ✅ **A shared cloud folder was unmodelled** – `backup.ts:334` writes `<name>.<pid>.tmp` and renames, and `docs/wiki/3.06-User-Data.md` documents per-device subfolders, cross-device pruning, foreign snapshots and conflict copies.
- ✅ **Store the canonical path, and canonicalize both sides of the reset compare** – `backup.ts:266` and `:274` both go through the exported `canonicalize()`, so the symlinked-userData mismatch the addendum warned about cannot arise.
- ✅ **Folder name reached `innerHTML`** – `automatic-backups-form.const.ts:62` escapes both href and text, with a spec.
- ✅ **Error convention on `PICK_BACKUP_FOLDER`** – whole handler wrapped in `createSafeIpcError` (`backup.ts:290`), helper genuinely extracted into the new `electron/safe-ipc-error.ts` rather than duplicated.
- ✅ **i18n regression from `HELP_DESKTOP`** – `help` is back to `HELP` on every platform and the caveat is a separate line, so translated locales keep their paragraph.
- ✅ **`promptToCreate` / probe file left behind** – dropping `promptToCreate` with the Windows reason in the comment is the right call, and `removeStaleTmpFiles` (`backup.ts:155`) only ever matches this app's own names and only past an hour, never a bare `.tmp`.
- ✅ **No spec for `_updateAutomaticBackUpCfg()`** – `config-page.component.spec.ts` picks twice and asserts the section is rebuilt rather than stacked.
- ➖ **`feat(electron):` commit message** – nothing to do: the repo squashes with the PR title, so only the squash body would carry the odd first subject line.

Refusing to re-create a vanished picked folder (`backup.ts:327`) was not on the list and is a good catch – it is the worse half of the same failure, and the mount-point caveat in the wiki is the honest way to state what is left.

### Still blocking

- **The rebase, and it is not mechanical.** The branch is 257 commits behind `master` and no longer merges. `2827f4cf` rewrote `electron/backup.ts`: `listBackupFiles()` is hardcoded to `BACKUP_DIR`, accepts any `*.json`, and now feeds three callers – `BACKUP_IS_AVAILABLE`, the new `IPC.BACKUP_LIST` behind the Settings backups list (`src/app/imex/local-backup/backup-sources.service.ts:133` → `window.ea.listBackups()`, restoring via `BACKUP_LOAD_DATA`), and `cleanupOldBackups`. Resolved naively, a user with a custom folder gets the *default* folder's files in that dialog, every restore from it is refused by this PR's `isPathInsideDir(backupDir, …)` check, and retention prunes the default folder while backups pile up unbounded in the picked one. Master's untimestamped `*.json` files also become unrestorable under the new filename filter.

  The clean resolution is to keep master's `listBackupFiles()` and parameterize it – folder from `getBackupDir()`, filter `isRestorableBackupFilename` – then have `BACKUP_IS_AVAILABLE`, `BACKUP_LIST` and `cleanupOldBackups` all go through it. That also picks up master's `try`/`catch` around the `statSync`, which this branch's `BACKUP_IS_AVAILABLE` (`backup.ts:202`) still lacks: a file removed between `readdirSync` and `statSync` throws `ENOENT` out of the handler and kills the startup restore prompt, and a shared cloud folder is exactly where that race becomes reachable.

### Nits

- **`electron/backup.test.cjs:368`** – the comment got mangled in the edit: "…and then fail every backup would fail."
- **PR description** – now slightly stale: it says discovery, retention and restore "only ever touch backup-shaped filenames (`YYYY-MM-DD_HHmmss.json`)", but after the fix discovery and restore also accept legacy `YYYY-MM-DD.json` in the default folder.

---

AlvaroBalbin:
Thanks for the detailed re-review. I rebased this onto current `master` and updated `listBackupFiles()` so availability, the Settings list, restore, and cleanup all use the configured backup directory with the right filename filter. I kept the `statSync` race guarded, fixed the test comment, and updated the PR description.

The focused backup suite passes all 38 tests, and the required per-file checks pass for every changed TypeScript file. The broader Electron run passed 324 tests; the remaining 5 are unrelated symlink tests that this Windows environment was not permitted to create.

Ready for another look when you have time.

johannesjo:

Thanks for the updates. Re-checked the 3 items from the earlier review at `476573e`:

- ✅ **The rebase, and it is not mechanical** – `listBackupFiles(backupDir)` (`electron/backup.ts:283`) is master's version parameterized, filtered by `isRestorableBackupFilename`, and all three callers go through it: `BACKUP_IS_AVAILABLE` (`:187`), `BACKUP_LIST` (`:198`) and `cleanupOldBackups` (`:365`). Master's `try`/`catch` around `statSync` came along (`:300`), so the vanished-file race no longer kills the startup prompt. Cleanup still re-filters on the strict shape (`:366`), so a user file in a picked folder is never a deletion candidate, while the default folder keeps accepting master's plain `*.json` for discovery and restore. The four new tests cover both folders plus the stat race, and the branch merges again.
- ✅ **Mangled test comment** – fixed at `electron/backup.test.cjs:453`.
- ✅ **Stale PR description** – now says what the code does.

### Still blocking

- **`src/app/pages/config-page/config-page.component.ts:265`** – the rebase resolution drops master's **Browse backups** button on desktop. The constructor builds `globalImexFormCfg` by mapping `GLOBAL_IMEX_FORM_CONFIG` and appending the `BROWSE` action to the `FILE_IMPORT_EXPORT` section (`:207`), but `_updateAutomaticBackUpCfg()` then replaces the whole array with `[...GLOBAL_IMEX_FORM_CONFIG, <auto-backups section>]` – the raw, undecorated const. Master's Electron branch appended to `this.globalImexFormCfg` instead, which kept it. `IMEX_FORM` (`imex-form.const.ts`) defines no other action, so on Electron the Import/Export section renders with no action row at all and the backups-list dialog loses its only user-initiated opener on the one platform that has backup files – while this PR is what re-adds the `BACKUP_LIST` IPC that dialog needs. Web and Android keep the button.

  Fix: pull the decorating `map` into a small helper (say `_buildImexBaseCfg()`) and rebuild from that in `_updateAutomaticBackUpCfg()`, rather than from the const. Worth extending the new spec too – it asserts the section count and that `localBackup` is not stacked, which is exactly the property that still holds here, so it passes while the button is gone. Asserting the `FILE_IMPORT_EXPORT` section still carries its action after a pick would have caught it.

---

PR #9491, opened 2026-08-07
fix(user-profile): warn about User Profiles removal before the next release

johannesjo:
User Profiles is being removed in the next release. This makes sure everyone
holding profile data hears about it in time, and that the export they are told
to take actually contains their data.

## What's here

**Warn about the removal.** The old deprecation snack was a 10s toast that a
user could easily miss. It becomes a persistent banner plus a warning in
Settings → App Features, both pointing at the export.

**Reach the cohort that can't see the feature.** Disabling User Profiles only
drops a localStorage flag — profile data stays on disk and `initialize()` never
runs again. Those users had extra profiles they could no longer reach and got
no warning at all. They are now warned at startup (gated on actually having
more than one profile, so a lone default profile doesn't nag anyone) with a
message pointing at the re-enable path. Re-enabling is safe:
`migrateOnFirstEnable()` short-circuits when metadata exists, so nothing is
overwritten.

**Don't let the banner mask other banners.** Only one banner renders at a time.
This one opens at startup and never auto-dismisses, so at any shared priority
it won every tie (stable sort + earliest insertion) and permanently hid
`Offline`, `UpdateAvailable`, `SyncConflictContentResolved`,
`SyncConflictsAutoResolved`, `SuperSyncEncryptionMigration`,
`SyncSafetyReminder`, `InstallWebApp` and `RatePrompt` until dismissed — several
of which are themselves data-safety notices. It now sits at the bottom of the
priority map. The test iterates every `BannerId` and asserts each outranks it;
reverting the priority fails it on all eight.

**Export the profile you're actually in.** `exportProfile()` read the stored
snapshot, which for the *active* profile is only written when switching away —
so the obvious download button handed users a backup missing everything they had
done since. It now exports live state for the active profile and the stored
snapshot for the rest.

With that fixed the rescue flow no longer needs a switch (and app reload) per
profile, so the banner text and wiki steps shrink to "open Manage Profiles,
export each one". The banner gains a shortcut button to that dialog, imported
lazily because the dialog component injects this service.

## Notes

- The banner sits at the lowest priority, so it stays hidden while any other
  banner is up — a user on a persistent `Offline` banner won't see it until
  they reconnect. That trade is deliberate.
- No sync, op-log or schema surface is touched.
- Only `en.json` was edited; the now-unused keys in other locales are left in
  place and `fallbackLang` covers untranslated locales.

## Testing

- Full unit suite green in both timezones (14101 / 14087, 0 failures).
- `npm run checkFile` clean on every touched file; `tools/check-doc-links.js` passes.
- New coverage: banner priority, both warning cohorts, the "lone default profile"
  no-warn case, both export branches, and the lazy dialog import.


mpeter50:
What is the eason for removing user profiles? I have not found information about it.

What are my options going forward if I dont want to synchronize my personal SP profile to my work computer, but want to synchronize the work SP profile to my personal computer?

PR #9493, opened 2026-08-07
feat: add Bulk Add Tasks to community plugins

InitialCrow:
## Bulk Add Tasks Plugin

Adds Bulk Add Tasks to the Super Productivity community plugin list.

Bulk Add Tasks allows users to create many tasks at once from structured
plain text.

### Features

- Create multiple tasks from one pasted text block
- Create subtasks through indentation
- Select projects using `[Project Name]`
- Add existing Super Productivity tags using `#tag`
- Add task notes using `((note))`
- Supports common bullet and checkbox prefixes
- Preview tasks before creation

### Links

- Repository: https://github.com/InitialCrow/bulk-add-tasks-plugin
- Release: https://github.com/InitialCrow/bulk-add-tasks-plugin/releases/tag/v0.1
- Plugin download: https://github.com/InitialCrow/bulk-add-tasks-plugin/releases/download/v0.1/bulk-add-tasks-v0.1.0.zip

### Compatibility

- Super Productivity: 18.16.0FP

### Security

- No external network requests
- No Node.js execution
- No credentials or secrets
- Uses the official Super Productivity Plugin API only

### Testing

Tested by installing the published release ZIP in Super Productivity
and creating tasks, subtasks, notes, projects and tags.


adnoh:
@InitialCrow This feature is already natively implemented into SP? Copy your Markdown to clipboard and paste it in SP while it has focus and it will create the tasks out of it. Also the bundled braindump plugin does exactly that with some ui?

InitialCrow:
@adnoh 
hi, I dont know, but it didn’t seem to match my needs.

I shared mine mainly because I thought it could be useful to others. From a single prompt in my plugin UI, it can create tasks and subtasks across multiple projects while keeping the expected structure, without having to navigate through each project separately. It can also add notes and handle a few other things — the full details are on the plugin page.

I’m not sure if braindump plugin already covers the same use case. If it does, then there’s no need to add mine. But I think mine is more powerful if its not necessary to add on plugin list its not a problem for me I just wanted to share.

adnoh:
@InitialCrow It's all good, I didn't mean to speak ill of it. Most people aren't aware of the direct Markdown insertion function, or the Braindump plugin went largely unnoticed when it was first introduced. I will definitely give it a try ;)

johannesjo:
Thanks for the submission — and for the unusually complete PR description, it made verifying this quick.

I checked the plugin itself, and everything you claimed holds up:

- The release ZIP matches the repo source byte-for-byte (`manifest.json` + `index.html`)
- No network calls, no `eval` / `new Function`, no node execution; `allowedHosts` is absent, so `PluginAPI.request` is fail-closed
- The declared permissions (`getAllProjects`, `getAllTags`, `addTask`, `showSnack`) match exactly what the code uses — nothing beyond them
- All `innerHTML` sinks go through a consistent `escapeHtml()`
- The manifest validates against `PluginManifest`, and `minSupVersion: 18.16.0` is a real release

Two small things to fix in the diff:

1. **Trailing newline** — the diff drops the newline at the end of `community-plugins.json` (`\ No newline at end of file`). `npm run prettier:file src/assets/community-plugins.json` fixes it. Our CI lint doesn't gate JSON formatting, so it slipped through, but the next contributor's format run would otherwise pick it up as unrelated noise.
2. **Casing** — `url` and `authorUrl` use `initialcrow`, while your actual login is `InitialCrow`. The links redirect fine, but the other entries preserve the owner's casing.

On @adnoh's point: they're right that the basics are already covered twice — pasting a markdown task list into the app creates tasks with subtasks and notes, and the bundled Braindump plugin does the same with a textarea and a project picker. What yours adds on top is per-block `[Project]` headings (so a single paste can span projects), `#tag` assignment, and the preview/validation step. Since that's the actual differentiator, I'd lead the `shortDescription` with it — as written it reads close to what's already in the box, which will make it harder for users to see why they'd want it.

One note about the plugin itself, not this PR: the import loop awaits one `addTask` per line, so a 100-line paste becomes 100 separate operations and a fair amount of sync traffic. `batchUpdateForProject` is the one-shot path, but it's per-project and has no `tagIds`, so it can't express your multi-project + tags case today — the loop is the right call for now, just worth knowing as the plugin grows. (Optional: adding a LICENSE would let users see the terms at a glance.)

InitialCrow:
@johannesjo Hi normaly, I change case issue yes ill add optimisation and more feature on the plugin after ill make hurry this first version for my needed case the license has not been pushed my bad ill add in next patch but anyway its MIT and its posted in readme (bottom of it).

johannesjo:
Heads-up — the merge that brought in master got mangled by the web conflict editor, and `src/assets/community-plugins.json` is no longer valid JSON. That's what's failing Tests, build-preview, Electron packaging and Lighthouse: the file is `import`ed at build time (`resolveJsonModule`), so a parse error breaks the whole build rather than just a runtime path.

Around line 152 it currently reads:

```json
    "stars": 0,
  },
    "name": "JSON to Tasks",
```

It needs to be:

```json
    "stars": 0
  },
  {
    "name": "JSON to Tasks",
```

(trailing comma removed, and the `{` opening the JSON to Tasks entry restored). Then `npm run prettier:file src/assets/community-plugins.json` to restore the trailing newline. Worth resolving locally rather than in the web editor — `node -e "require('./src/assets/community-plugins.json')"` confirms it parses before you push.

Everything else checks out: the release zip is byte-identical to the repo source, the manifest validates, permissions match usage exactly, and there is genuinely no network access at all — no fetch, XHR, WebSocket, localStorage or cookie access anywhere in the source, and with no `http` permission declared the app blocks egress regardless.

Two things still open from before:

- The `shortDescription` should lead with what sets this apart from the built-in markdown paste and the bundled Brain Dump plugin — the multi-project `[Project]` headings, `#tag` assignment, and the preview/validation step.
- The README says MIT but there's no LICENSE file in the repo yet.

InitialCrow:
@johannesjo Sorry, I was on vacation. I’ve now corrected the JSON properly and used the script as requested. Sorry for the small mistake!

johannesjo:

Thanks for coming back and fixing the JSON properly. The entry itself is in good shape: valid JSON with the trailing newline restored, casing now matching your login, no other entries touched, and all checks green.

I read the plugin source at `ed23d12` – all 6 files, ~31 KB, nothing skipped. Everything you claimed in the description holds up. The release ZIP `bulk-add-tasks-v0.1.0.zip` is byte-identical to the repo (sha256 matches on both `manifest.json` and `index.html`, and the `v0.1` tag points at that same commit), so the code I read really is the code that runs. There is no network reach of any kind – no `fetch`, `XMLHttpRequest`, `WebSocket` or `sendBeacon` – no `localStorage`/`sessionStorage`/`indexedDB`, no plugin storage, no `eval`/`new Function`/dynamic `import`, no injected script, and nothing credential-shaped. The source is unminified and readable end to end. `iFrame: true`, no `nodeExecution`, and with no `http` permission and no `allowedHosts` the sanctioned network path is disabled at the host regardless. The four declared permissions (`getAllProjects`, `getAllTags`, `addTask`, `showSnack`) match the call sites exactly – nothing declared is unused, nothing undeclared is reached – every field you pass to `addTask` exists on `PluginCreateTaskData`, and all `innerHTML` sinks go through `escapeHtml()` consistently.

I can only tell you what I found at that commit, not that it is safe. Nothing below blocks the entry, and three of the four are about how the plugin meets the host rather than about this diff.

### Questions

- **More than two levels of indentation looks like it breaks the import halfway through.** SP's task model is two levels deep, and `addTask` rejects a `parentId` that is itself a subtask (`_validateTaskReferences` in `plugin-bridge.service.ts`, message "Cannot nest subtasks: parent task is itself a subtask"). Your `parentStack` at `index.html:452-467` pops to the nearest strictly shallower row, so an ordinary nested outline – `A` / `  B` / `    C` – gives `C` a `parentId` of `B`, which is already a subtask. `indentWidth` is read from the raw line, so this is any deeper indent, not just 2-space steps. Reading it through, `parseInput` only raises errors for an unknown project, an unclosed note, an unknown tag and an empty title, so the preview shows no error and `Create tasks` stays enabled; then `A` and `B` are created and `addTask` throws on `C`. The part that would sting is the retry: `els.tasks.value = ''` only runs on the success path (`:543`), and the `finally` block calls `renderPreview()` which re-enables the button, so after "Import stopped after 2 task(s)" the textarea still holds all three lines and the obvious next click recreates `A` and `B` as duplicates. Did you hit this while testing? If it reproduces, rejecting depth greater than 2 in `parseInput` the same way you already reject an unknown project would keep the promise the preview makes, and clearing the imported lines on partial failure would make a retry safe.
- **Tags on indented lines show up in the preview but I do not think they survive.** The bridge sets `tagIds: []` for any `addTask` carrying a `parentId`, and the `addSubTask` reducer sets it again with the comment "should always be empty" (it also forces the parent's `projectId`). That covers both `#tags` parsed from a child line and the sidebar "Default tags" selection. `renderPreview` builds the meta line from `row.tagIds` for every row including children, so the preview promises tags the subtask cannot receive. Since `#tag` support and the preview are two of the three things that set this apart, is it worth dropping tags from child rows in the preview, or warning there that subtasks ignore them?
- **Top-level titles still go through SP's short-syntax parser; subtasks do not.** The bridge's main-task branch leaves `isIgnoreShortSyntax` at its default, so `ShortSyntaxEffects` fires for it, while the subtask branch opts out explicitly. So a line `Review PR +Work @tomorrow 30m` at column 0 can land as `Review PR` with a project, a due date and an estimate the preview never showed, and the `+Work` token quietly overrides whatever `[Project]` heading was in effect – while the same line one indent down keeps the tokens in its title. It depends on the user's short-syntax settings, so it may never bite. Was that intended, or worth a line in the README?
- **The LICENSE still is not there.** The README says MIT, but the repo shows no license and there has been no commit since 7 August, so I could not tell whether the patch you mentioned is still coming. Not a blocker for the listing – mostly checking that this is still something you plan to maintain, since the list is where people will find it.

### Nits

- **`index.html:163`** – the textarea placeholder contains literal `\n` sequences. HTML attributes do not interpret those, so the example renders as one long wrapped line with visible backslash-n instead of the multi-line example you meant. A real newline in the attribute value, or `&#10;`, fixes it. It is the first thing a new user sees, so it is worth the small change.
- **`src/assets/community-plugins.json`** – `stars` is `0`; the repo currently shows 1. Cosmetic, and it goes stale either way.

---

johannesjo:
Maintainer note: I'd like to list this — the entry is well-formed, the plugin makes no network calls, and the v0.1 release zip is byte-identical to source. I'd just like a **0.1.1** before merging, for two things:

1. **A depth guard on the import.** Three or more levels of indentation make a subtask the parent of another subtask, which the app rejects (`CANNOT_NEST_SUBTASKS`). The preview doesn't check depth, so the import creates the first rows and then throws — and since the textarea is only cleared on success, the natural retry duplicates the tasks that did get created. Worth catching in the preview. (Related: the "Import stopped after N task(s)" error box is erased in the same tick by `renderPreview()` in the `finally` block, so a partial import leaves no visible trace of how far it got.)
2. **A LICENSE file** — the README says MIT but the repo has none.

No rush and nothing else is blocking. Ping me when 0.1.1 is out and I'll merge the entry.

InitialCrow:
@johannesjo  Thanks for the review! [v0.1.1 is now released](https://github.com/InitialCrow/bulk-add-tasks-plugin/releases/tag/v0.1.1).
The preview now rejects nested subtasks before any tasks are created. Partial-import errors remain visible with the confirmed task count and failed line, plus a warning about duplicates when retrying. The MIT LICENSE is included in both the repository and the release ZIP.
Regression tests pass, and I verified that the published ZIP contents match the source. 
and few optimisation I will continue to upgrade plugin

johannesjo:
Screened v0.1.1 and it passes — this is approved, it just needs a conflict resolved before I can merge.

What I checked against the source at `3fe6b53` / tag `v0.1.1`, not the release notes:

- **Depth guard**: `parseInput()` derives depth from actual parent links rather than indent width and rejects `depth > 1`, and `importTasks()` re-validates against the live textarea before any `addTask` call — so a stale preview can't be imported and `CANNOT_NEST_SUBTASKS` is no longer reachable via indentation. ✅
- **LICENSE**: present, MIT, matches the README. ✅
- No network calls, no telemetry, no `eval`/`new Function`, no storage. All eight `innerHTML` sinks escape their input (tried ``'&lt;img src=x onerror=...&gt;'`` as a title and a `<script>` project heading — both render as text). All four declared permissions have call sites. `minSupVersion: 18.16.0` is a real release. ✅
- **The published release ZIP is SHA-256 identical to the source at the tag**, all three entries, nothing extra. ✅

**The conflict**: `src/assets/community-plugins.json` has moved on since you branched — several entries were added and the `stars` count on the preceding entry changed, which collides with your appended entry. Merging `master` into your branch should resolve it in one go; it's purely the tail of the array.

Two things for a future 0.1.2 — **not** conditions for this entry, just worth knowing:

1. The duplicate case isn't fully closed. After a partial failure the Create button is re-enabled on unchanged text, and `importError` is hidden on the next click — so one more click re-creates the whole list, and the warning disappears exactly when the duplicate is made. Requiring an edit before retry, or skipping already-created rows, would close it.
2. `commit.sh` option 2 force-moves the latest tag without refreshing release assets, so a published ZIP can silently stop matching the tag it claims. That's precisely the property I just verified, so I'd drop that branch and cut a new version instead.

Also worth noting for your own expectations: the catalog is a link list, not an install path — the entry points at your repo under an explicit "not reviewed by the Super Productivity team" notice, and it pins no version. So users get whatever is current at your URL, which is why (2) matters more than it looks.

Ping me once the conflict is resolved and I'll merge.

---

InitialCrow:
@johannesjo Done

johannesjo:
Thank you! <3

Issue #9281, opened 2026-07-24
💡 Mobile task detail panel: grouping, layout density and empty states


ba6ul:
### Problem Statement

I use Super Productivity daily on Android, and the task detail panel is the screen I touch most on mobile. Its current layout is slower to use than it needs to be. Screenshot attached.

Concretely:

- Everything is one full-width column. Eight stacked cards of near-identical height, most holding a single short value. On a phone this forces scrolling to reach fields that could all fit on one screen, while the bottom of the sheet stays empty.

- No semantic grouping. "Planned for" and "Deadline" are both dates and belong together. "Duration" and "Recur" are also scheduling concepts. They are currently separated by unrelated rows, so you have to read every label to find one instead of scanning to a region.

- Container fields are styled identically to value fields. "Subtasks" and "Notes / Checklist" hold lists and expand in place. "Duration", "Planned for", "Deadline" and "Recur" hold one value and open a picker. Both render as the same card with the same weight, so the affordance does not communicate what a tap will do.

- Four fields, three different conventions for the same state. Duration shows -/-. Deadline and Recur show nothing at all. Subtasks shows (0).

- "Subtasks (0)" is a label where an action is needed. With zero subtasks, the useful thing to show is "Add subtask". The count is only informative once a count exists.

- The tag row has no visible label. The last card is a tag icon and empty space. Nothing identifies the field, and a control with no text label is also a screen reader failure.

- All eight icons are the same monochrome outline weight. Stacked vertically, they give the eye nothing to anchor on, so the icon column adds noise without adding scanning speed.

<img width="1080" height="1628" alt="Image" src="https://github.com/user-attachments/assets/8df29931-ddba-49ee-98c9-c7e36fa80174" />

### Possible Solution

A reworked layout for the mobile detail panel:

a) Two-column grid for scalar fields. Duration, Planned for, Deadline and Recur are all short values. Put them in a 2x2 grid of compact tiles. This alone removes roughly half the vertical scroll and makes the whole scheduling block visible at once.

b) Group by meaning, with light section separation:

- Scheduling: Planned for, Deadline, Duration, Recur
- Content: Subtasks, Notes / Checklist, Links & Files
- Metadata: Tags

c) Distinguish container fields from value fields. Content sections stay full width and keep the expand chevron. Scalar fields become compact tiles that read as "opens a picker". Different shape, different expectation.

[...]

PR #9498, opened 2026-08-07
fix(tasks): show 'Set deadline' instead of 'Deadline' on an empty deadline row

ba6ul:
## Problem

Part of the follow-up items from #9281. When a task has no deadline set, the Deadline row's title read "Deadline" — a noun that doesn't communicate what tapping the row does. The Schedule row right above it already follows an action-oriented empty-state convention ("Schedule task"), so Deadline was inconsistent with it.

## Solution

`deadlineLabelKey()` now falls back to the existing `T.F.TASK.CMP.SET_DEADLINE` key ("Set deadline") instead of `T.F.TASK.ADDITIONAL_INFO.DEADLINE`, when no deadline is set. No new translation key needed. The overdue/due-by labels for tasks that already have a deadline are unchanged.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have added tests for my changes (if applicable) (not applicable, single-line label swap using an existing key)


johannesjo:
Thank you very much! <3

PR #9501, opened 2026-08-07
fix(tasks): drop the misleading (0) from the empty Subtasks panel header

ba6ul:
## Problem

Part of the follow-up items from #9281. The Subtasks panel header always interpolated a count into "Subtasks ({{nr}})", including when a task has zero subtasks, which rendered as "Subtasks (0)". That's technically honest but reads oddly for the common case of a task with no subtasks yet.

## Solution

Show a plain "Subtasks" title when the task has no subtasks, and keep the counted "Subtasks (N)" form once subtasks exist. This needed one new translation key (`SUB_TASKS_EMPTY`, value `"Subtasks"`) since no existing key in this feature's namespace covers the plain form — the other two `SUB_TASKS = "Subtasks"` keys in the codebase are Jira/issue-tracker integration labels for an unrelated UI, and reusing them would couple two unrelated features.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have added tests for my changes (if applicable) (not applicable, template conditional + one new i18n key)


johannesjo:
Thanks — this is exactly the piece I said yes to in #9281, scoped as its own PR as asked. Happy to take it, with two small things I'd like changed first.

**What I checked**

- The other `SUB_TASKS` consumer, `dialog-view-archived-task.component.html:16`, is already inside `@if (subTasks().length)` (line 11), so no `(0)` survives there either and the two stay consistent.
- The new key is safe for the 27 other locale files: `main.ts` sets `fallbackLang: DEFAULT_LANGUAGE` and ngx-translate v17 falls back *per key* (`TranslateStore.getTranslation()`), so an untranslated locale renders "Subtasks", not the raw key. `npm run int:test` agrees — it reports key differences as informational and states outright that missing translations use the English fallback.
- Merges cleanly into current master (`git merge-tree`, no conflicts), keys sorted correctly in both `t.const.ts` and `en.json`, prettier clean, and the 28 existing specs in `task-detail-panel.component.spec.ts` pass with the patch applied. Nothing in `e2e/` matches the old string.

**1. Please put the branch in a label-key computed rather than the template**

This file already has the pattern for "which key does this row use": `scheduleLabelKey()` (`task-detail-panel.component.ts:405`) and `deadlineLabelKey()` (`:414`), which #9498 just extended. The template stays a single `{{ key | translate }}` and the decision lives in one computed. The equivalent here:

```ts
subTasksLabelKey = computed(() =>
  this.task().subTasks?.length
    ? this.T.F.TASK.ADDITIONAL_INFO.SUB_TASKS
    : this.T.F.TASK.ADDITIONAL_INFO.SUB_TASKS_EMPTY,
);
```

```html
<span>{{ subTasksLabelKey() | translate: { nr: task().subTasks?.length } }}</span>
```

Passing `nr` to the plain key is a no-op: `TranslateDefaultParser.interpolateString()` only substitutes `{{…}}` placeholders that actually occur in the string, so extra params are ignored. I applied this locally — `npm run checkFile` clean, all existing specs pass — and it drops the duplicated `<span>`.

**2. `SUB_TASKS_EMPTY` will get mistranslated**

`"SUB_TASKS_EMPTY": "Subtasks"` reads like an empty-state *message* ("No subtasks yet"), but the value is the plain label. Across 27 locales that's an invitation to translate the wrong thing. `SUB_TASKS_NO_COUNT` (or `SUB_TASKS_PLAIN`) says what it is. Cheap now, annoying to undo once it has been translated.

**3. A regression test, if you're up for it**

The existing spec already renders this panel, so the empty case is nearly free. Dropped in at the end of the first `describe`, this fails on master and passes with your change — I verified both directions:

```ts
describe('sub tasks panel header', () => {
  const headerText = (): string =>
    (
      fixture.nativeElement.querySelector('.panel-header-icon')!
        .parentElement as HTMLElement
    ).textContent!.trim();

  it('uses the plain label when there are no sub tasks', () => {
    componentRef.setInput('task', { ...MOCK_TASK, subTasks: [] });
    fixture.detectChanges();

    expect(headerText()).toContain('F.TASK.ADDITIONAL_INFO.SUB_TASKS_EMPTY');
  });
});
```

On master it fails with `Expected 'listF.TASK.ADDITIONAL_INFO.SUB_TASKS' to contain 'F.TASK.ADDITIONAL_INFO.SUB_TASKS_EMPTY'`. With the computed from (1) you can assert `component.subTasksLabelKey()` directly instead of going through the DOM.

One heads-up so you don't burn time on it: the mirror test for the non-empty branch does *not* work in this spec. Putting a sub-task on the input makes the panel render `<task-list>`, which pulls in `WorkContextService -> LOCAL_ACTIONS -> Actions -> ScannedActionsSubject` and takes down the whole run from `afterAll`. Overriding `TaskListComponent`'s template doesn't help — that leaves the class's own injections in place. It would need a `WorkContextService` provider, which isn't worth it here; the empty case is the regression anyway.

johannesjo:
**🔍 Reviewed** – no blocking issues found at `6ecdfe4`.

Re-checked this at the same head as my 14 Aug comment. Two updates: the change still applies after master moved 263 commits under the branch, and **I'm withdrawing item 1 of that comment** – I checked the premise this time and it doesn't hold.

### Withdrawing my earlier request for a `subTasksLabelKey()` computed

I argued the file's convention puts label-key choice in the `.ts` and cited `scheduleLabelKey()` / `deadlineLabelKey()`. That was the wrong comparison: those two are used at `:144` and `:185`, which are plain non-panel rows with a fixed icon. The nearest actual precedent is the **notes panel header in this same file** (`task-detail-panel.component.html:240-253`), same `<ng-container panel-header>` construct, which already picks among three keys with `@if / @else if / @else` and duplicates the `<span>` inline:

```html
@if (isMarkdownChecklist()) {
  <mat-icon>checklist</mat-icon>
  <span>{{ T.F.TASK.ADDITIONAL_INFO.CHECKLIST | translate }}</span>
} @else if (task().notes) {
  ...
```

So your template branching matches the sibling panel header, not a departure from it. Sorry for the detour – please leave it as it is.

### Nits

Both optional; this is mergeable as it stands.

- **`src/assets/i18n/en.json` + `src/app/t.const.ts`** – I'd still nudge on the `SUB_TASKS_EMPTY` name, but weaker than I put it last time. Every other `EMPTY`-named key in `en.json` holds a sentence rather than a label: `HISTORY_EMPTY` ("No reflections saved yet…"), `EMPTY` ("No archived projects."), `EMPTY_UNREVIEWED`, `EMPTY_HISTORY`, `EML_EMPTY`, `EMPTY_STATE_HINT`. Yours would be the only one whose value is a bare noun. What pulls the risk down, and what I got wrong before: `tools/add-missing-i18n-variables.js` seeds a translator's WIP file with the English *value* (`missing[key] = enObj[key]`), so they see the line literally as `"SUB_TASKS_EMPTY": "Subtasks"` and would have to ignore the source string to write "No subtasks yet". `SUB_TASKS_NO_COUNT` is still a clearer name, but it's polish, not a merge condition.

- **`task-detail-panel.component.spec.ts`** – a regression test for the empty branch is close to free: `MOCK_TASK` already sets `subTasks: []` and no `parentId`, and `isShowSubTasksPanel()` is just `!task.parentId`, so all 28 existing tests already render the exact branch this PR adds. One assertion on the panel-header text covers it. Caution rather than fact on the mirror test: I expect a non-empty fixture to pull in `TaskListComponent` and its DI graph (`WorkContextService -> LOCAL_ACTIONS -> Actions`) and take the run down, but I haven't re-run karma to confirm that, so try it before assuming it's blocked.

Everything else checked out: `dialog-view-archived-task.component.html:16` is already inside `@if (subTasks().length)` so the two `SUB_TASKS` render sites stay consistent, untranslated locales fall back to English per key (`main.ts:204`), no i18n key-parity check fails on an en-only key, and no unit or e2e test asserts on the old header text.

---

johannesjo:
Withdrawing item 1 of my August review — I was wrong that the label choice belongs in a `subTasksLabelKey()` computed. The notes panel header right below this one (`task-detail-panel.component.html:240-253`) uses the same `<ng-container panel-header>` construct and already picks among three keys inline with `@if/@else-if/@else`; the computeds I pointed at feed non-panel rows. So this matches the nearest precedent and is fine as written.

Merging. Thanks @ba6ul.

johannesjo (approved):
Matches the notes-header precedent; good to go.

PR #9497, opened 2026-08-07
fix(tasks): switch the Duration row icon between add and edit

ba6ul:
## Problem

The Duration row's edit affordance (`[inputIcon]`) was hardcoded to `'edit'`, even when a task has no time spent and no estimate set yet. The Schedule and Deadline rows right below it already switch between `add` and `edit` depending on whether the field has a value, so Duration was the odd one out — part of the inconsistency raised in #9281.

## Solution

Added a `hasTimeData()` computed signal (`timeSpent || timeEstimate`) and made `[inputIcon]` conditional on it, matching the pattern already used by the Schedule and Deadline rows. No new strings, no label changes, no changes to how the time values render.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki). (not applicable, no user-facing docs change)
- [x] I have run `npm run checkFile` on changed `.ts`/`.scss` files
- [x] I have added tests for my changes (if applicable) (not applicable, no new branching logic beyond the existing pattern)
- [x] Existing tests still pass
- [x] My commit messages follow the Angular format (`type(scope): description`)

PR #9502, opened 2026-08-07
fix(tasks): reveal the add/edit affordance on touch devices

ba6ul:
## Problem

Part of the follow-up items from #9281. `.input-item__edit-btn` was `display: none !important` except under `.input-item:hover`, so the add-vs-edit icon distinction the code already computes for the Duration/Schedule/Deadline rows never rendered on a touch device — hover never resolves before a tap, so there was no visual signal for what tapping a row would do.

## Solution

Reveal the button under `@media (hover: none)` as well as on hover, so touch devices see the same add/edit affordance desktop users see on hover. The button has no click handler of its own (taps fall through to the row's own click handler), so it's marked `tabindex="-1"` `aria-hidden="true"` now that it's visible, keeping it out of the tab order and off assistive tech.

Verified in Chrome DevTools with touch emulation enabled (`matchMedia('(hover: none)').matches === true`): the icon is now visible without hovering.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have added tests for my changes (if applicable) (not applicable, CSS media query + a11y attributes)

### DIFF
--- a/src/app/features/tasks/task-detail-panel/task-additional-info-item/task-detail-item.component.html
+++ b/src/app/features/tasks/task-detail-panel/task-additional-info-item/task-detail-item.component.html
@@ -22,6 +22,8 @@
         class="input-item__edit-btn"
         color="accent"
         mat-icon-button
+        tabindex="-1"
+        aria-hidden="true"
       >
         <mat-icon>{{ inputIcon }}</mat-icon>
       </button>
--- a/src/app/features/tasks/task-detail-panel/task-additional-info-item/task-detail-item.component.scss
+++ b/src/app/features/tasks/task-detail-panel/task-additional-info-item/task-detail-item.component.scss
@@ -252,4 +252,8 @@
   .input-item:hover & {
     display: block !important;
   }
+
+  @media (hover: none) {
+    display: block !important;
+  }
 }


johannesjo:
Thanks for this @ba6ul — and sorry, because this one is on me. You built exactly what I specified in #9281, down to the `tabindex="-1"` / `aria-hidden="true"` detail, and the implementation is clean. But I looked at it in context and I don't want to land it. You deserve the reasoning rather than a silent close.

**It doesn't solve the problem I pointed it at.** I claimed this was the cause of your point (c). Re-reading (c), that was about container fields vs value fields — Subtasks and Notes expand in place, Duration and Deadline open a picker. The add/edit icon doesn't distinguish those; it distinguishes empty from filled. Every row gets an icon either way, so it still doesn't communicate what a tap will do. My diagnosis was wrong.

**It's the same redundancy I rejected in item 2.** I turned down the accent tint because the value renders as text in the same row, so a signal that is true exactly when that text is non-empty adds no information. The add/edit glyph is that same signal in glyph form. That argument should have killed this one too.

**And the label changes already cover it.** #9498 and #9501 say "this is empty, tapping will fill it" in words, in the same row, at no cost in pixels. Once those land, the pencil is a second signal for a state the label already states outright.

**The cost points the wrong way for the issue it came from.** Point (g) of #9281 was that the icon column adds noise without adding scanning speed. This adds a second column facing the first, on up to five rows. On desktop the reveal is cheap precisely because it's absent almost all of the time; on touch there is no transient state, so it becomes permanent chrome. It also takes ~24px off the value column on every row (40px button, -24px `margin-left`, +8px `margin-right`), and those values already truncate — the parent row is a task title. And it leaves a permanently visible 40px `mat-icon-button` that is `aria-hidden`, untabbable and has no click handler: it renders as a control and isn't one.

If I ever do want a filled/empty cue there, the pattern is already in the same template — the notes row swaps `chat_bubble_outline` for `chat`. That changes an existing icon instead of revealing a second one: no new element, no horizontal space. But with #9498 and #9501 I don't think anything is needed.

On #9497: it still stands on its own as a fix for the desktop hover state, so I'll look at it, but it's no longer load-bearing. #9498, #9501 and #9505 are unaffected and I'll go through them separately.

Sorry for the wasted round trip — the miss was in my spec, not your implementation.

PR #9505, opened 2026-08-07
style(tasks): tighten task-detail row spacing to match the task list rhythm

ba6ul:
## Problem

Part of the follow-up items from #9281. `.input-item` / `.mat-expansion-panel` used `margin: var(--s)` (8px) per row in the task detail panel, which the maintainer estimated as roughly 70px of chrome across the panel's ~530px height.

## Solution

Planner and boards already moved their inter-row gap from `var(--s)` to `var(--s-half)` for the same reason (#9266, "match the task list rhythm"). Apply the same token here, on the same rule, so the task detail panel's density matches the rest of the app. No change to card style (border/radius/background), and the 48px row touch target is untouched — only the outer margin changed.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Other (please describe): visual density/style tweak
- [ ] Documentation

## Checklist

- [x] I have added tests for my changes (if applicable) (not applicable, single CSS token swap)

### DIFF
--- a/src/app/features/tasks/task-detail-panel/task-additional-info-item/task-detail-item.component.scss
+++ b/src/app/features/tasks/task-detail-panel/task-additional-info-item/task-detail-item.component.scss
@@ -53,7 +53,9 @@
 .input-item,
 .mat-expansion-panel {
   border: 1px solid var(--divider-color);
-  margin: var(--s);
+  // Match the main task list's inter-row rhythm rather than the roomier 8px,
+  // recovering density while keeping the 48px touch target (see #9266).
+  margin: var(--s-half);
   border-radius: var(--card-border-radius);
   background: var(--task-detail-bg);
   font-size: 14px;


johannesjo:
Thanks — this is the change I asked for in #9281 (b), and the scope is right: card style and the 48px touch target are both untouched, and it stops well short of the merged-rows look I ruled out.

One thing to fix before this goes in.

### `margin` is a four-side shorthand, so this also halves the horizontal inset

That breaks alignment with the panel's own title header. `.task-title-wrapper` (`task-detail-panel.component.scss:41`) keeps `margin: var(--s)`, and it shares `.wrapper` with the rows — the `task-detail-item` hosts are full-width and add no inset of their own — so the two insets sit side by side.

Measured on this PR's preview build (`56db29f8`, = head `97f3231`), detail panel open, 7 rows:

|                          | left | right | width |
| ------------------------ | ---- | ----- | ----- |
| `.task-title-wrapper`    | 1225 | 1514  | 289   |
| all 7 rows, as shipped   | 1221 | 1518  | 297   |

Every row now overhangs the title by 4px on each side. It is visible rather than theoretical, because the header carries `border-bottom: 2px solid var(--c-primary)` — that accent line stops 4px short of the card edges on both sides. Before this PR the two boxes were pixel-identical (forcing a row back to `8px` reproduces the header box exactly: 1225/1514).

The fix is to scope the shorthand to the vertical axis:

```scss
margin: var(--s-half) var(--s);
```

I verified that on the same build: all 7 rows — both the `.input-item` sections and the `mat-expansion-panel`s — land at 1225/1514, exactly the header box, and the vertical gain is fully preserved: row-stack height 517px, identical to what the PR produces now, against 541px before the change.

That is also what the precedent actually did. #9266 moved planner and boards via `margin-bottom`, not the `margin` shorthand — the horizontal inset was never part of it. And `velvet.css:700` already styles these exact elements `margin: 6px 8px !important`: vertical tightened, horizontal held at 8px.

### Two smaller notes, neither blocking

- **The gain is roughly half the ~70px I estimated**, so don't be surprised. Measured: 6 gaps went 8px → 4px = 24px on a 7-row panel. Margins collapse through the `display: block` `task-detail-item` host, so there was only ever one 8px gap between two rows, not two stacked — my estimate counted them uncollapsed. Still worth doing; just calibrating the number so nobody goes looking for the missing 35px.
- **Velvet dark gets nothing from this.** `velvet.css:700` overrides these selectors with `!important`, so the density change is a no-op on that theme. Pre-existing divergence, not something to fix in this PR.

### Nit

Could the comment reference #9281 alongside #9266? #9266 is the precedent for the token, but #9281 is where the reasoning and the estimate live, which is the more useful thing for the next reader to land on.

With the horizontal inset restored, this is good to merge.

johannesjo:

Not a new review – the head is still `97f3231`, so the one open item is the same one from my review above. What is new is that I re-derived it from source against today's master, three weeks and 263 commits later, and it still stands.

### Should fix before merge

- **`src/app/features/tasks/task-detail-panel/task-additional-info-item/task-detail-item.component.scss:58`** – `margin` is a four-side shorthand, so `margin: var(--s-half)` halves the horizontal inset as well as the vertical gap. Still true on current master: `.task-title-wrapper` keeps `margin: var(--s)` (`task-detail-panel.component.scss:41`), the header and every `task-detail-item` are still siblings under `.wrapper` (`task-detail-panel.component.html:1-2`), and the item host is still `:host { display: block }` with no inset of its own – so each row's box ends up 4px wider per side than the header, and the header's `border-bottom: 2px solid var(--c-primary)` accent line stops short of the card edges. Fix is unchanged: `margin: var(--s-half) var(--s);` – the vertical gain is fully preserved.

  One extra data point I did not have last time, and it comes from your own rationale: the main task list you are matching has **no** horizontal inset at all. `_task-base.scss:54-59` puts `.box` at `left: 0; right: 0; top: 2px; bottom: 2px` – the 4px inter-row rhythm is entirely vertical. So "match the main task list's rhythm" argues for the vertical halving and specifically against the horizontal one. Same shape as the two other precedents: #9266 moved planner and boards via `margin-bottom`, never the shorthand, and `velvet.css:700` styles these exact selectors `margin: 6px 8px !important`.

### Nits

- **`…task-detail-item.component.scss:57`** – the comment still points only at #9266. Adding #9281 alongside it would land the next reader on the reasoning rather than just the token precedent.

### Housekeeping

- The branch is 263 commits behind `master`, and `c852b776af` touched this same file yesterday (removing four lines roughly 30 above your rule). GitHub still reports the PR mergeable, so nothing is broken – just worth a rebase when you push the fix so the line numbers line up.

---

johannesjo:
Agreed on the problem — the detail panel rows were carrying a roomier rhythm than the task list they sit next to. Taken forward in #9900 (`96ba2e7`).

One deviation: I kept the horizontal inset at `var(--s)` and only halved the vertical one (`margin: var(--s-half) var(--s)`). Collapsing both axes pulled the rows out of line with the panel header above them, which was more noticeable than the density gain. Vertical rhythm now matches the list; the 48px touch target is unaffected either way.

Closing once #9900 lands — thanks for spotting it.

ba6ul:
Confirmed on current master, margin: var(--s-half) var(--s) is already in place at task-detail-item.component.scss:55 via #9900. This PR is superseded, closing it. Thanks for taking it forward and catching the horizontal-inset issue.

PR #9506, opened 2026-08-07
fix(right-panel): stop the edge-close-handle from overlapping panel content

ba6ul:
## Problem

`.edge-close-handle` (the small chevron on the right panel's left edge, used to close/drag-resize the panel) is `position: absolute; left: -8px; width: 20px`, reaching 12px into `.side-inner`, and it's vertically centered on the whole panel height (`top: 50%`) rather than on the actual rendered content's height. On content shorter than the viewport (e.g. a task with few fields set), the handle can land mid-content and visually overlap a row instead of sitting clear of it.

## Solution

Reserve left padding on `.side-inner` (`var(--s2)`, 16px — enough to clear the handle's 12px reach with a small buffer) so no row ever renders under the handle, regardless of where it's vertically centered. This is a `right-panel` fix, not specific to any one content type, since the handle's positioning logic is identical for every panel it hosts (task detail, notes, issue panel, plugin panels).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have added tests for my changes (if applicable) (not applicable, CSS spacing fix)

### DIFF
--- a/src/app/features/right-panel/right-panel.component.scss
+++ b/src/app/features/right-panel/right-panel.component.scss
@@ -108,6 +108,13 @@
   height: 100%;
   position: relative;
   z-index: 2;
+  box-sizing: border-box;
+  // .edge-close-handle reaches 12px into this container (left: -8px, width:
+  // 20px) and is vertically centered on the whole panel height, not on the
+  // content's height, so it can land anywhere along the left edge and
+  // overlap short content. Reserve enough space that no row ever sits
+  // under it.
+  padding-left: var(--s2);
 }
 
 .right-panel {
--- a/src/app/features/right-panel/right-panel.component.scss
+++ b/src/app/features/right-panel/right-panel.component.scss
@@ -108,13 +108,6 @@
   height: 100%;
   position: relative;
   z-index: 2;
-  box-sizing: border-box;
-  // .edge-close-handle reaches 12px into this container (left: -8px, width:
-  // 20px) and is vertically centered on the whole panel height, not on the
-  // content's height, so it can land anywhere along the left edge and
-  // overlap short content. Reserve enough space that no row ever sits
-  // under it.
-  padding-left: var(--s2);
 }
 
 .right-panel {
@@ -158,7 +151,9 @@
   touch-action: none; // allow pointer events to control without scrolling
   position: absolute;
   top: 50%;
-  left: -8px; // Position outside the panel
+  // -16px (width 20px, border-box) reaches only 4px into the panel, clear of
+  // the 8px margin on panel rows (was -8px, overlapping by 4px).
+  left: -16px;
   width: 20px;
   height: 48px;
   transform: translateY(-50%);


johannesjo:
Thanks for this — the overlap you describe is real, and I was able to reproduce and measure it. But I think the fix is at the wrong layer, and I'd like to suggest a smaller one.

## What I measured

On your preview deploy (`24a1648d`), task detail panel open, default theme, 320px panel — toggling `padding-left` in the DOM to A/B against master:

- `.edge-close-handle` (`left: -8px`, `width: 20px`, computed `box-sizing: border-box`) reaches exactly **12px** into `.right-panel`.
- Panel rows (`.input-item`, `.task-title-wrapper`) carry `margin: var(--s)` — 8px — from `task-detail-item.component.scss:56`.

So the handle covers the leftmost **4px** of a row — its border and rounded corner — across the 48px band at the panel's vertical midpoint. Real, but small.

## The cost

Row gaps measured from the panel's left and right edges:

| | scrollbar | left gap | right gap (incl. scrollbar) |
|---|---|---|---|
| master, panel not scrolling | 0 | 8px | 7.2px |
| **this PR, panel not scrolling** | 0 | **24px** | **7.2px** |
| master, panel scrolling | 15px | 8px | 22.4px |
| this PR, panel scrolling | 15px | 24px | 22.4px |

The asymmetry lands hardest in exactly the case the PR describes — short content that doesn't scroll — where rows go from centred to 24px-left / 7px-right. It only looks balanced when a scrollbar happens to be present. At `MIN_WIDTH: 250` (`right-panel.component.ts:47`) that's roughly 10% of the panel width taken off one side.

It also applies to panels that don't have the bug. The notes panel already pads 12px (`notes.component.scss:22`), which is exactly the handle's reach — nothing overlaps there today, and it would become 28px/12px. Plugin panels host iframes under `overflow: hidden` (`right-panel.component.scss:122-126`) and would simply lose 16px of width.

## Suggested alternative

The handle's own geometry is what causes the overlap, so adjusting the handle fixes every panel type at once with no content-layout change anywhere. Measured, with `width: 20px` and `box-sizing: border-box`:

| `left` | reach into panel | clearance vs. 8px rows |
|---|---|---|
| `-8px` (current) | 12px | −4px (overlap) |
| `-12px` | 8px | 0 (flush) |
| `-16px` | 4px | 4px |
| `-20px` | 0px | 8px |

`left: -16px` gives real clearance while keeping the handle attached to the edge. I checked the main-column side too: task rows end ~100px left of the panel, and the closest elements are the top-right header buttons at ~9px — which sit in a different vertical band than the vertically-centred handle, so there's room.

One caveat: the handle's `border-radius: 3px 8px 8px 3px` rounds its *right* side, which reads like the straddle may be deliberate. If `-8px` is an intentional design choice then this is @johannesjo's call and my suggestion is the wrong one.

## Smaller notes

- The description's trigger looks inverted. The handle sits at 50% of the *panel* height, so content shorter than half the viewport ends *above* it and can't overlap; the overlap needs content that reaches *past* the midpoint. The mechanism you describe is right, just not the case that triggers it.
- `box-sizing: border-box` is a genuinely necessary companion to the padding here, given the inline `min-width: {panelWidth}px` on `.side-inner` — good catch, that would have overflowed without it.

CI is green and `checkFile` passes on the changed file. Tightly-scoped diff and correct use of spacing tokens per the styling guide — my only real objection is the layer it's applied at.

johannesjo:

Nothing has changed on the branch since the review above, so rather than repeat it: the branch is now 263 commits behind `master`, and I re-checked whether that analysis has gone stale. It hasn't – `.edge-close-handle` is still `left: -8px; width: 20px` (12px of reach), `.side-inner` is untouched on master so this still applies cleanly (`mergeable: MERGEABLE`, no rebase needed), task rows still carry `margin: var(--s)`, notes still pads 12px, and `--s2` is still 16px.

Two things that weren't covered before, both in this PR's favour:

- **Themes don't change the handle's geometry.** `velvet.css:595` and `zen.css:229` are the only other places touching `.edge-close-handle`, and they restyle `background` / `border` / `box-shadow` only, never `left`, `width`, or `box-sizing`. So the 12px reach is identical in every theme.
- **`box-sizing: border-box` isn't redundant, and it keeps the sizing tests green.** There's no global `* { box-sizing: border-box }` in this repo (`page.scss:13` scopes it to `html`/`body`/`app-root`/`.app-frame`), so that line really is load-bearing against the inline `[style.min-width.px]="panelWidth()"`. And `e2e/tests/navigation/right-panel-sizing.spec.ts` measures `.side`, not `.side-inner`, so border-box leaves those assertions unaffected.

I also went looking for a defect in the diff itself and found one, below.

### Should fix before merge

- **`right-panel.component.scss:117`** – the padding is a no-op for the Issues panel, so the invariant the change states isn't actually delivered. `issue-panel`'s `:host` is `position: absolute; inset: 0` (`issue-panel.component.scss:6-7`), and its nearest positioned ancestor is `.side-inner` itself, because `right-panel-content`'s host is static (`right-panel-content.component.scss:1-4`). An absolutely positioned box is laid out against its containing block's **padding box**, so `inset: 0` still spans the full `0..panelWidth`, and `padding-left` moves it by exactly zero pixels. Net effect: the Issues panel keeps its current geometry (rows at 12px from `issue-provider-tab.component.scss:17-20`, flush against the handle's 12px reach), while task detail, notes, plugin and schedule panels each give up 16px of width. So the PR description's "applies to every right-panel consumer (task detail, notes, issue panel, plugin panels)" and the code comment's "no row ever sits under it" are both false as written. The fix depends on which direction this goes overall – if the padding layer is kept, it needs either a positioned wrapper the issue panel can be inset from, or the claim narrowed to the panels it actually reaches.

### Nits

- **`right-panel.component.scss:112-116`** – the committed comment repeats the inverted trigger from the description. The handle is `top: 50%` of `.right-panel`, which is `100%` of `.side`, which spans the viewport (see the `:host` note at lines 8-12), so it sits at half the *viewport* height. Content shorter than that ends above the handle and can't overlap; the overlap needs content reaching *past* the midpoint. Worth correcting before it lands, since the next person tuning this value will reason from it.

### Questions

- Is this still something you want to pursue? It's been open a while, and I'd rather ask than leave it hanging – either answer is fine, and the bug is well documented on the thread regardless.

---

johannesjo:
Decision on the fix layer: **change the handle geometry, not the container padding.**

`.edge-close-handle { left: -16px }` — one line, no content-layout change across panel types, and it's the only option that actually reaches the Issues panel, whose host is absolutely positioned and so lays out against the padding box regardless of what `.side-inner` declares.

For the record, the −8px straddle wasn't deliberate: nothing in the history ever tuned that value, and the velvet/zen themes only restyle the handle's paint, never its geometry, so both candidate fixes behave identically everywhere.

Happy to take a push to this branch with that one-liner if you're still around — otherwise I'll land it and credit you.

ba6ul:
Sorry for the slow turnaround on this, had exams and couldn't get back to it sooner.

You're right that the padding approach was solving it at the wrong layer. Pushed 8015bbf0 with exactly your one-liner: .edge-close-handle { left: -16px }, and reverted the .side-inner padding entirely, so no content-layout change anywhere, and it still reaches past the Issues panel's absolutely-positioned host like you flagged.

Ran checkFile on the file and the full lint/test suite through the pre-push hook, all passing. No need to push anything on your end, it's already up as 8015bbf0.

johannesjo:
**✅ Ready to merge** – no blocking issues at `8015bbf0`. Merge confidence 88% – the geometry is verified from the CSS and the theme overrides, but I could not measure the rendered result in the real app.

Thanks for the updates, and no worries about the timing. Re-checked the 3 items from the earlier review at `3e5736cb`:

- ✅ **`padding-left` was a no-op for the Issues panel** – fixed by dropping the padding layer entirely. `.side-inner` (`right-panel.component.scss:106-111`) is back to its original four declarations, and the handle now clears every panel from its own box: `left: -16px` with `width: 20px` reaches 4px in, against 8px row margins in task detail and the schedule panel, and 12px in notes and the issue tabs. Because it moves the handle rather than the container, the absolutely-positioned `issue-panel` host is covered too – which the padding never was.
- ✅ **Inverted trigger in the committed comment** – gone with the padding block. The replacement at `right-panel.component.scss:154-155` makes a geometry claim only, and the numbers check out.
- ✅ **Whether this was still being pursued** – answered.

I also looked for regressions the one-liner could introduce and found nothing blocking: the handle's `z-index` (66) still sits above `.content` (44) so the extra overhang paints and hit-tests normally, no ancestor clips it in LTR (`.side` is `overflow: visible`, the host only clips during the slide animation), the panel is not rendered at all on `xs` so mobile is unaffected, no e2e spec targets this element, and `velvet.css:595` / `zen.css:229` add a 1px border but never `width` or `box-sizing`, so the 4px reach is identical in every theme.

Two small things, neither of them anything for you to change – I have flagged them for @johannesjo since he specified the value:

- The handle's dead zone over the main column goes from 8px to 16px, and that strip is where the routed page's scroll container sits (`app.component.scss:133-141`). In the handle's 48px band its scrollbar is now fully covered rather than mostly covered. Pre-existing at `-8px`, just wider.
- `left` is a physical property, so in RTL the handle protrudes away from the content and `.main-content { overflow: hidden }` (`app.component.scss:56-62`) eats 8px more of it than before. `inset-inline-start` would be the direction-safe form, but the panel is already RTL-unaware on master (`border-left`, `resize-handle { left: -2px }`, `chevron_right`), so this is not something this PR introduces.

Two corrections to my earlier review while I am here: this repo *does* have a global `*, *:after, *:before { box-sizing: border-box }` at `src/styles/page.scss:261`, so my claim that the old `box-sizing` line was load-bearing was wrong – it was redundant, and removing it is correct. And the PR description still describes the old `padding-left` approach; worth a quick edit so the merge commit body matches what actually landed.

---

johannesjo:
Thank you very much! <3

PR #9512, opened 2026-08-08
fix(task-repeat-cfg): anchor the monthly day-of-month presets on the date they were set up for

KamilDev:
`MONTHLY_FIRST_DAY` and `MONTHLY_LAST_DAY` build their anchor from `new Date()` and drop the `referenceDate` both callers pass, so the saved recurrence starts in a month the user never chose. This is the divergence I flagged at the end of #9404 but deliberately left out of it — different cause from the workday roll, so it gets its own change.

Every branch of the bug is reachable today, and I reproduced each before touching the code.

## From the add task bar

Type `Pay rent`, pick **2026-11-15** in the date chip, then choose **Monthly on the last day** in the repeat menu (today being 2026-08-08).

The config is written with `startDate: 2026-08-31` — the last day of the month the bar happens to be open in. The task's own `dueDay` is the picked day, but both `addTaskRepeatCfgToTask` effects derive the first occurrence from the config, so the task itself lands on **Aug 31**, three months before the month that was picked into. There is no start-date field in the add bar, so nothing on screen explains it.

## From the repeat dialog, creating

Open the repeat dialog on a task due **2099-12-15** and pick a monthly day-of-month preset. `MONTHLY_LAST_DAY` writes `2026-08-31`, `MONTHLY_FIRST_DAY` writes `2026-09-01` — a config for a task three years out, anchored in the month the dialog was opened in. After the fix: `2099-12-31` and `2100-01-01`. (Credit to @johannesjo for measuring this branch out — the original description only claimed the two below.)

## From the repeat dialog, editing

A config anchored on **2026-12-31** with `MONTHLY_LAST_DAY`, deliberately set to start in December. Open the dialog, change the title or the reminder time, hit Save.

`save()` re-derives the quick setting from the stored start date on every save with a non-`CUSTOM` quick setting, and this preset ignores that date — so the save writes `startDate: 2026-08-31`. `getTaskRepeatCfgChanges` then puts `startDate` into the change set, `startDate` is a `SCHEDULE_AFFECTING_FIELDS` member, and `rescheduleTaskOnRepeatCfgUpdate$` relocates the live instance. Editing a title moves the recurrence four months earlier. That is the same class of spurious reschedule #7373 fixed for the other fields.

## The fix

One helper, used by both branches:

```ts
const anchorDayFor = (referenceDate: Date | undefined, today: Date): Date =>
  referenceDate && referenceDate.getTime() > today.getTime() ? referenceDate : today;
```

The floor at today is load-bearing rather than defensive, and it is what keeps #7726 intact. A long-running config carries an old start date, so a plain `referenceDate ?? today` would anchor a freshly chosen preset years in the past — the exact backdating #7726 removed. With no reference date, or one already behind us, the anchor is still derived from today and nothing about the current behavior changes.

From the anchor, each preset takes **the first day matching its own pattern that falls on or after that anchor** — one rule, both presets. It lands them in different months for the same anchor (`2099-12-15` → `2100-01-01` for the 1st, `2099-12-31` for the last day) because the two patterns sit at opposite ends of a month, not because they disagree.

That next-month result for `MONTHLY_FIRST_DAY` is deliberate, and it is the reason I did not take the symmetric-looking alternative ("the 1st of the anchor's own month, floored at today", which would give `2099-12-01`). In the add bar the reference date *is* the task's due day, and in the dialog it is a user-set, user-visible start date; either way, starting the recurrence **before** the day the user chose materializes an instance on a day they did not ask for, and in the dialog it drags a visible start date backwards. Too-late costs the user one skipped period and is visible in the start-date field; too-early creates something behind their back. Pinned by `MONTHLY_FIRST_DAY anchors to the 1st after the reference date`.

Left alone on purpose: `MONTHLY_CURRENT_DATE`, `MONTHLY_NTH_WEEKDAY` and `YEARLY_CURRENT_DATE` take the reference verbatim, past included — for those the reference *is* the chosen day, and flooring it would change a behavior nobody has complained about.

**Switching presets inside the dialog does not touch the start date at all.** The form's `change` handler calls `getQuickSettingUpdates()` with no reference, but it applies the result through `field.form.patchValue()`, and `startDate` / `monthlyLastDay` are not formly field keys (`task-repeat-cfg-form.const.ts` has `title`, `quickSetting`, `repeatEvery`, `repeatCycle`, the monthly-weekday pair, the seven weekday flags, and the trailing option toggles). `patchValue` drops keys with no matching control, so the visible start-date field keeps its value and `save()` re-derives from *that*, idempotently. An earlier version of this description claimed the switch re-anchored the visible field on today; that was wrong, and @johannesjo caught it.

**One boundary this does not cross:** the date chip keeps showing the picked day while the config anchors to the 1st or the last day of that month. Rolling the chip the way `MONDAY_TO_FRIDAY` does since #9404 would be a UX change beyond this bug — "monthly on the 1st, starting from the month I picked" is what the preset means, and the workday roll exists for an excluded-day rule these presets do not have.

**Scope, stated plainly:** this fixes configs whose start date is in the *future*. For an already-running monthly config the floor sends the anchor to today, so the first save in each new month still rewrites `startDate`, still puts it in the change set, and still fires `rescheduleTaskOnRepeatCfgUpdate$` — identical before and after this PR. So #7373's class of spurious reschedule is narrowed here, not closed; the common case stays open, and the root of it is that `save()` re-derives the preset unconditionally instead of leaving a stored anchor that already satisfies the preset alone. Happy to take that as a follow-up.

## Coverage

- `get-quick-setting-updates.spec.ts` — six specs: each preset honouring a future reference date, `MONTHLY_FIRST_DAY` on a reference that already is the 1st, `MONTHLY_LAST_DAY` re-deriving the anchor it produced itself (the dialog's every-save path), and both presets ignoring a reference date in the past. 39 total, from 33.
- `add-task-bar.component.spec.ts` — the two picked-date repros above, asserting the `startDate` handed to `addTaskRepeatCfgToTask`. 62, from 60.
- `dialog-edit-task-repeat-cfg.component.spec.ts` — an untouched save on a future-anchored config of each preset, asserting `startDate` is absent from the change set, since key presence is what `rescheduleTaskOnRepeatCfgUpdate$` filters on (`field in changes`). 42, from 40.

The component specs use 2099 dates so they assert the same thing whatever day the suite runs on.

**Verified by breaking each half of the fix first.** Never honouring the reference (the old behavior) → `4 FAILED / 35 SUCCESS` in the util spec, `2 FAILED / 60` in the add bar spec, `2 FAILED / 40` in the dialog spec. Dropping the floor (`referenceDate ?? today`) → `2 FAILED / 37`, `Expected '2020-01-31' to be '2026-08-31'` and `Expected '2020-02-01' to be '2026-09-01'`.

Full `npm test`: **14120 of 14136 passing, 16 skipped** in `Europe/Berlin` *and* `America/Los_Angeles`. `npm run lint`: 0 errors, only the pre-existing grandfathered `max-lines` warnings and two unused-disable directives, none in the touched files.

### DIFF
--- a/src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.spec.ts
+++ b/src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.spec.ts
@@ -954,4 +954,51 @@ describe('DialogEditTaskRepeatCfgComponent', () => {
       expect(savedCfg().skipOverdue).toBe(true);
     });
   });
+
+  // save() re-derives the quick setting from the stored start date, so a preset
+  // that ignored that date rewrote it on every save — and a startDate in the
+  // change set reschedules the live instance (#7373).
+  describe('monthly anchors survive an unrelated save', () => {
+    const changes = (): Partial<TaskRepeatCfg> =>
+      mockTaskRepeatCfgService.updateTaskRepeatCfg.calls.mostRecent()
+        .args[1] as Partial<TaskRepeatCfg>;
+
+    it('keeps a future MONTHLY_LAST_DAY start date where the user put it', async () => {
+      const fixture = await setupTestBed({
+        repeatCfg: {
+          ...DEFAULT_TASK_REPEAT_CFG,
+          id: 'repeat-cfg-last-day',
+          title: 'Pay rent',
+          quickSetting: 'MONTHLY_LAST_DAY',
+          repeatCycle: 'MONTHLY',
+          monthlyLastDay: true,
+          startDate: '2099-09-30',
+        },
+      });
+
+      fixture.componentInstance.save();
+
+      expect(mockTaskRepeatCfgService.updateTaskRepeatCfg).toHaveBeenCalledTimes(1);
+      expect(changes().startDate).toBeUndefined();
+      expect('startDate' in changes()).toBe(false);
+    });
+
+    it('keeps a future MONTHLY_FIRST_DAY start date where the user put it', async () => {
+      const fixture = await setupTestBed({
+        repeatCfg: {
+          ...DEFAULT_TASK_REPEAT_CFG,
+          id: 'repeat-cfg-first-day',
+          title: 'Pay rent',
+          quickSetting: 'MONTHLY_FIRST_DAY',
+          repeatCycle: 'MONTHLY',
+          startDate: '2099-10-01',
+        },
+      });
+
+      fixture.componentInstance.save();
+
+      expect(changes().startDate).toBeUndefined();
+      expect('startDate' in changes()).toBe(false);
+    });
+  });
 });
--- a/src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.spec.ts
+++ b/src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/dialog-edit-task-repeat-cfg.component.spec.ts
@@ -979,7 +979,8 @@ describe('DialogEditTaskRepeatCfgComponent', () => {
       fixture.componentInstance.save();
 
       expect(mockTaskRepeatCfgService.updateTaskRepeatCfg).toHaveBeenCalledTimes(1);
-      expect(changes().startDate).toBeUndefined();
+      // `rescheduleTaskOnRepeatCfgUpdate$` filters on `field in changes`, so key
+      // absence is the assertion that matches the effect.
       expect('startDate' in changes()).toBe(false);
     });
 
@@ -997,7 +998,6 @@ describe('DialogEditTaskRepeatCfgComponent', () => {
 
       fixture.componentInstance.save();
 
-      expect(changes().startDate).toBeUndefined();
       expect('startDate' in changes()).toBe(false);
     });
   });
--- a/src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/get-quick-setting-updates.spec.ts
+++ b/src/app/features/task-repeat-cfg/dialog-edit-task-repeat-cfg/get-quick-setting-updates.spec.ts
@@ -206,6 +206,60 @@ describe('getQuickSettingUpdates', () => {
     });
   });
 
+  // Both callers pass a reference day — the picked date in the add bar, the
+  // config's own start date in the repeat dialog — and these two presets used
+  // to derive their anchor from `new Date()` regardless, so the saved
+  // recurrence started in a month the user never chose.
+  describe('monthly day-of-month presets honour the referenceDate', () => {
+    afterEach(() => jasmine.clock().uninstall());
+
+    const mockToday = (date: Date): void => {
+      jasmine.clock().install();
+      jasmine.clock().mockDate(date);
+    };
+
+    it('MONTHLY_FIRST_DAY anchors to the 1st after the reference date', () => {
+      mockToday(new Date(2026, 7, 8));
+      const result = getQuickSettingUpdates('MONTHLY_FIRST_DAY', new Date(2026, 11, 15));
+      expect(result!.startDate).toBe('2027-01-01');
+    });
+
+    it('MONTHLY_FIRST_DAY keeps a reference date that already is the 1st', () => {
+      mockToday(new Date(2026, 7, 8));
+      const result = getQuickSettingUpdates('MONTHLY_FIRST_DAY', new Date(2026, 11, 1));
+      expect(result!.startDate).toBe('2026-12-01');
+    });
+
+    it('MONTHLY_LAST_DAY anchors to the end of the reference month', () => {
+      mockToday(new Date(2026, 7, 8));
+      const result = getQuickSettingUpdates('MONTHLY_LAST_DAY', new Date(2026, 11, 15));
+      expect(result!.startDate).toBe('2026-12-31');
+    });
+
+    // Re-deriving the preset from the date it already produced must be a no-op:
+    // the dialog runs this on every save, and a changed startDate reschedules
+    // the live instance (#7373).
+    it('MONTHLY_LAST_DAY leaves an anchor it produced itself alone', () => {
+      mockToday(new Date(2026, 7, 8));
+      const result = getQuickSettingUpdates('MONTHLY_LAST_DAY', new Date(2026, 10, 30));
+      expect(result!.startDate).toBe('2026-11-30');
+    });
+
+    // #7726 still holds: a stale start date must not drag a fresh anchor into
+    // the past, so the reference day is floored at today.
+    it('MONTHLY_FIRST_DAY ignores a reference date that is in the past', () => {
+      mockToday(new Date(2026, 7, 8));
+      const result = getQuickSettingUpdates('MONTHLY_FIRST_DAY', new Date(2020, 0, 15));
+      expect(result!.startDate).toBe('2026-09-01');
+    });
+
+    it('MONTHLY_LAST_DAY ignores a reference date that is in the past', () => {
+      mockToday(new Date(2026, 7, 8));
+      const result = getQuickSettingUpdates('MONTHLY_LAST_DAY', new Date(2020, 0, 15));
+      expect(result!.startDate).toBe('2026-08-31');
+    });
+  });
+
   describe('YEARLY_CURRENT_DATE', () => {
     it('should return YEARLY cycle with repeatEvery 1', () => {
       const result = getQuickSettingUpdates('YEARLY_CURRENT_DATE');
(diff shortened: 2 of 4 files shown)


johannesjo (commented):
Reviewed by checking out the branch and re-running the three suites (39 / 42 / 62 SUCCESS, matching the description), then probing the before/after behaviour directly rather than reading for it.

## The bug is real, and the PR understates itself

I traced both callers and confirmed the mechanism: `startDate` is in `SCHEDULE_AFFECTING_FIELDS` (`task-repeat-cfg.effects.ts:49`), and both `addTaskRepeatCfgToTask` effects derive the task's target day from `getFirstRepeatOccurrence(taskRepeatCfg)` (`:88`, `:167`) — so the config's `startDate` really does override the picked `dueDay`. The start date is also a genuinely user-visible, user-editable field (the `openScheduleDialog()` button), so the pre-fix behaviour is "pick a future start date, watch the button show it, hit Save, watch it silently snap back to the current month and relocate the live instance".

Measured before/after on the same inputs:

| path | input | before | after |
|---|---|---|---|
| dialog **create**, `MONTHLY_LAST_DAY` | task due `2099-12-15` | `2026-08-31` | `2099-12-31` |
| dialog **create**, `MONTHLY_FIRST_DAY` | task due `2099-12-15` | `2026-09-01` | `2100-01-01` |
| dialog **edit**, untouched save, `MONTHLY_LAST_DAY` | `startDate: 2020-01-31` | `2026-08-31` | `2026-08-31` |

The first row is a third fixed branch the description doesn't claim — opening the repeat dialog on a task due later and picking a monthly preset was landing the config three years early. Worth mentioning, it's the most convincing repro of the three.

## One correction to the description

> The form's `change` handler still passes no reference, so switching presets inside the dialog re-anchors on today and shows that in the visible start-date field

That isn't what happens. `updateParent` applies changes via `field.form.patchValue(...)`, and `patchValue` silently ignores keys with no matching control. `startDate` and `monthlyLastDay` are not formly fields — the keys in `task-repeat-cfg-form.const.ts` are `title`, `quickSetting`, `repeatEvery`, `repeatCycle`, `monthlyWeekOfMonth`, `monthlyWeekday`, `monday`–`sunday`, `defaultEstimate`, `notes`, `repeatFromCompletionDate`, `shouldInheritSubtasks`, `disableAutoUpdateSubtasks`, `waitForCompletion`, `skipOverdue`. Confirmed against a real `FormGroup`: patching `{startDate, monthlyLastDay, repeatCycle}` onto a group that only has `repeatCycle` applies `repeatCycle` and drops the other two.

So the preset switch never writes `startDate` into the model; the visible field keeps its value and `save()` re-derives from *that*. The conclusion still holds (`save()` is a fixed point either way, and I verified the idempotency cases), and the real behaviour is better than described — but the reasoning in the description is load-bearing for whoever touches this next, so it's worth correcting.

## The main point: the two presets disagree, and the comment doesn't say so

Same task, same due date, sibling presets:

- `MONTHLY_LAST_DAY` → `2099-12-31` — the month the user picked
- `MONTHLY_FIRST_DAY` → `2100-01-01` — the next month, and next year

The new comment on `MONTHLY_FIRST_DAY` says the anchor "lands in the month the recurrence was set up for rather than the month it happens to be saved in". That's only true when the anchor day is already the 1st; otherwise it lands in the month *after*. The `MONTHLY_LAST_DAY` comment ("the last day of the anchor day's month") is accurate. Please reword the first one to the rule the code actually implements — "the first 1st-of-month on or after the anchor day" — since the current wording describes `MONTHLY_LAST_DAY`'s semantics, not its own.

Behind the wording there's a real choice worth making deliberately. `#7726` floors at *today*, and `2099-12-01` is comfortably in the future, so "the 1st of the anchor's own month, floored at today" is available and would make the two presets symmetric (`2099-12-01` / `2099-12-31`). It also changes near-future picks: today + a pick one month out currently gives `2026-09-01`, and this PR moves it to `2026-10-01` — i.e. for `MONTHLY_FIRST_DAY` the first occurrence gets *further* from the picked date than it is today.

I lean towards keeping your version — "never start before the day the user picked" is the safer surprise — but it should be a stated choice rather than a side effect of reusing the old `getDate() === 1` shape.

## Scope note

The fix only covers configs whose `startDate` is in the future. For an already-running monthly config the floor sends the anchor to today, so the first save in each new month still rewrites `startDate`, still puts it in the change set, and still fires `rescheduleTaskOnRepeatCfgUpdate$` (third row above — identical before and after). That's pre-existing and not a regression, but "the same class of spurious reschedule #7373 fixed" reads as if the class is closed, and the common case is still open. A sentence in the description, or a follow-up issue, would set expectations right.

## Minor

`dialog-edit-task-repeat-cfg.component.spec.ts` — `expect(changes().startDate).toBeUndefined()` followed by `expect('startDate' in changes()).toBe(false)` is redundant; the second implies the first, and it's the one that matches what the effect filters on.

## Verdict

Good fix — correctly scoped, minimal (one helper, two call sites, no new surface), and the floor-at-today is rightly identified as load-bearing rather than defensive, with a spec pinning each direction. Leaving `MONTHLY_CURRENT_DATE` / `MONTHLY_NTH_WEEKDAY` / `YEARLY_CURRENT_DATE` alone is the right call. Using 2099 dates in the component specs and a mocked clock in the util specs keeps them wall-clock and timezone independent.

Nothing here blocks on behaviour. Before merge I'd like the `MONTHLY_FIRST_DAY` comment fixed and the next-month semantics confirmed as intentional; the description correction and the scope note are follow-ups.

KamilDev:
Thanks — all four items addressed in `d22641ecb`, and the `patchValue` correction is a good catch.

## The `MONTHLY_FIRST_DAY` comment

You're right that the trailing clause described `MONTHLY_LAST_DAY`'s semantics, not its own. Reworded to the rule the code implements, and I moved the shared half up onto `anchorDayFor` so the next reader sees why the two outputs differ:

```ts
// ...floored at today so a stale start date cannot drag a freshly chosen
// preset into the past (#7726).
//
// Both day-of-month presets below then take the first day matching their own
// pattern that falls on or after that anchor, so a recurrence never starts
// before the day it was set up for. That one rule puts them in different
// months for the same anchor (2099-12-15 → 2100-01-01 for the 1st, 2099-12-31
// for the last day), because the two patterns sit at opposite ends of a month.
```

and at the case:

```ts
// The first 1st-of-month on or after the anchor day: the anchor's own
// month when the anchor already is the 1st, the next month otherwise.
```

`MONTHLY_LAST_DAY`'s comment now says "the same on-or-after rule" so the pairing is explicit.

## Confirming the next-month semantics — intentional

That framing is also my answer to the real choice you raised: the two presets aren't disagreeing, they're one rule — *first day matching the pattern, on or after the anchor* — and `2099-12-31` vs `2100-01-01` falls out of the 1st and the last day sitting at opposite ends of a month. `2099-12-31 ≥ 2099-12-15` and `2100-01-01 ≥ 2099-12-15`; both hold the invariant.

The alternative ("the 1st of the anchor's own month, floored at today" → `2099-12-01`) buys outcome symmetry by breaking that invariant, and I don't think the trade is worth it. In the add bar the reference date *is* the task's due day; in the dialog it's a user-set, user-visible start date. Starting before it materializes an instance on a day the user didn't pick, and in the dialog it also drags a visible field backwards — the direction #7726 and #7373 both exist to prevent. The two errors aren't symmetric either: too late costs one skipped period and is visible in the start-date field, too early creates something behind the user's back.

On "the first occurrence gets *further* from the picked date than it is today" — true as a distance, but the old value was computed from `new Date()` and had no relationship to the pick at all; it gave `2026-09-01` whether you picked Sep 20 or nothing. So it's less "the fix moves it away" than "the fix gives it a defined relationship for the first time". Pinned by `MONTHLY_FIRST_DAY anchors to the 1st after the reference date` (`2026-12-15` → `2027-01-01`), so a later change of heart trips a spec rather than sliding through.

## The description correction

Confirmed and fixed. `updateParent` goes through `field.form.patchValue()`, and neither `startDate` nor `monthlyLastDay` is a formly key — `task-repeat-cfg-form.const.ts` has `title`, `quickSetting`, `repeatEvery`, `repeatCycle`, `monthlyWeekOfMonth`, `monthlyWeekday`, the seven weekday flags, and the trailing option toggles. `FormGroup.patchValue` skips keys with no matching control, so the preset switch never writes `startDate` and the visible field keeps its value; `save()` re-derives from that. The description now says this instead of the wrong version, and credits the correction.

Worth noting the `change` handler still calls `getQuickSettingUpdates()` with no reference date — harmless today precisely because its `startDate` is dropped in transit, which is a slightly uncomfortable thing to be relying on. Not touching it here.

## Scope note

Added to the description rather than left implied, in your framing: the fix covers configs whose start date is in the future; for an already-running config the floor sends the anchor to today, so the first save in each new month still rewrites `startDate` and still fires the reschedule — identical before and after (your third row). So #7373's class is narrowed here, not closed. The root of the remaining case is that `save()` re-derives the preset unconditionally instead of leaving a stored anchor that already satisfies it alone — happy to take that as a follow-up if you want it, it's a different change with its own #7726 interaction to think through.

## Minor

Dropped `expect(changes().startDate).toBeUndefined()`, kept the `in` assertion with a line saying why it's the one that matters (`SCHEDULE_AFFECTING_FIELDS.some((field) => field in changes)` at `task-repeat-cfg.effects.ts:263`).

---

Full `npm test` after the change: **14120 of 14136, 16 skipped, SUCCESS** in `Europe/Berlin` and `America/Los_Angeles`. `npm run lint`: 0 errors, 10 pre-existing grandfathered warnings, none in the touched files.

johannesjo:

PR #7621, opened 2026-05-15
perf(sync): SuperSync server speed + correctness hardening

johannesjo:
## Summary

SuperSync server performance work (the perf-phases plan) plus correctness
hardening from a multi-pass code review. Same domain as #7611 ("Improve Super
Sync Server Speed"); this is the reviewed branch with `master` merged in.

## What's in here

**Performance (perf-phases plan, earlier commits):**
batch-upload pipeline (`processOperationBatch` split into validate / dedup /
classify / conflict-detect / reserve+insert stages), `payload_bytes` backfill
with an indexed scan, VC-aggregate + conflict-code dedupe, redundant
replay-state measurement skipped, bounded auth-cache invalidation.

**Correctness review fixes (latest 7 commits, each TDD'd):**

- **#1** serialize the `payload_bytes` migration reconcile against live uploads
  via the `users`-row `FOR UPDATE` lock in one transaction (closes a stale-low
  counter race → quota bypass)
- **#2** bound the Helm `migrate-db` initContainer with `timeout`
  (+ BusyBox 143/137 exit-code detection, since it runs under Alpine BusyBox)
- **#3** abort quota cleanup while a user's `payload_bytes` backfill is
  incomplete — never destroy irreversible restore points for a 413 that the
  rolled-back decrement makes pointless
- **#4** don't cache a wholly-transient (`INTERNAL_ERROR`, rolled-back) batch
  in the request-dedup window, so a same-`requestId` retry re-hits the DB
- **#6** always use the `Recreate` rollout strategy (was gated on
  `persistence.enabled`) — consistent with the single-replica in-memory-state
  invariant
- **#5** reviewed and **intentionally kept** — the intra-batch dup-id
  `DUPLICATE_OPERATION` divergence is documented and test-pinned; no change

**Docs:** after merging master's sync-docs consolidation, ported the
batch-pipeline diagram/tables into `docs/sync-and-op-log/diagrams/02-server-sync.md`
(seq-range reservation, prefetch conflict detection, `payloadBytes`,
`latestFullStateVectorClock`, dropped `tombstones`).

## Verification

- super-sync-server unit suite: **717 passed / 5 skipped / 0 failed** (34 files)
- `tsc --noEmit`: clean
- `helm lint` + `helm template` verified (Recreate in both persistence states);
  `sh -n` + timeout exit-code unit check on `migrate-deploy.sh`
- `master` merged into the branch — one docs modify/delete conflict resolved
  (accept deletion + port content); no other conflicts
[...]

PR #9522, opened 2026-08-09
fix(supersync-chart): mount tmp into the migrate-db init container

### LINKED ISSUE #9523: SuperSync Helm chart: a default install never becomes ready — migrate-db init container has no writable /tmp
## Summary

A default `helm install` of `packages/super-sync-server/helm/supersync` never reaches ready. The `migrate-db` init container exits immediately and crash-loops, so no migration runs.

```
$ kubectl get pods -n supersync
NAME                        READY   STATUS       RESTARTS   AGE
supersync-99b96fc49-6zwwg   0/1     Init:Error   3          2m46s

$ kubectl logs supersync-99b96fc49-6zwwg -c migrate-db
mktemp: : Read-only file system
```

## Cause

The pod template already declares a `tmp` emptyDir, and the application container mounts it at `/tmp`. The `migrate-db` init container declares **no `volumeMounts` at all**, while `values.yaml` sets:

```yaml
securityContext:
  readOnlyRootFilesystem: true
```

`scripts/migrate-deploy.sh` calls `mktemp`, which then has nowhere to write. So this is a missing mount rather than a missing volume — the container's sibling and the volume it needs are both already there.

## Reproduction

1. `helm install` the chart with defaults plus the two required values (`supersync.secret.jwtSecret`, `postgresql.password`).
2. Watch the pod: `Init:Error`, restarting.
3. `kubectl logs <pod> -c migrate-db`.

Observed on k3s v1.36.1 with `postgresql.enabled=false` and an external CloudNativePG database, and the init container is identical on the bundled-Postgres path, so it should reproduce either way.

## Impact

Self-hosting via the chart cannot work without the operator either discovering the missing mount themselves or relaxing `readOnlyRootFilesystem` — the second being a security default loosened to work around a one-line omission.

## Fix

PR #9522 mounts the existing `tmp` volume into the init container. One hunk, no new values, no behaviour change for anyone who had already worked around it.

sanchpet:
Fixes #9523.

A default `helm install` of the SuperSync chart never becomes ready. The `migrate-db` init container exits immediately with:

```
mktemp: : Read-only file system
```

and crash-loops, so no migration runs and the deployment stays at `Init:Error`.

The cause is a missing mount rather than a missing volume. The pod already declares a `tmp` emptyDir, and the app container mounts it at `/tmp` — but `migrate-db` declares no `volumeMounts` at all, while `securityContext` defaults to `readOnlyRootFilesystem: true`. `scripts/migrate-deploy.sh` calls `mktemp`, which then has nowhere to write.

This mounts the volume that already exists. One hunk, no new values, no behaviour change for anyone who had already worked around it by relaxing `readOnlyRootFilesystem`.

Reproduced on k3s v1.36.1 with `postgresql.enabled=false` and an external CloudNativePG database; the same failure occurs with the bundled Postgres, since the init container is identical in both paths.

Verified with `helm template` that the rendered init container now carries the mount, and on the cluster that the migration proceeds and the release goes ready.

Heads-up given as the issue template asks — happy to adjust the shape if you would rather see `TMPDIR` pointed at an existing writable path instead.

### DIFF
--- a/packages/super-sync-server/helm/supersync/templates/deployment.yaml
+++ b/packages/super-sync-server/helm/supersync/templates/deployment.yaml
@@ -70,6 +70,12 @@ spec:
               value: {{ .Values.externalDatabase.url | quote }}
               {{- end }}
             {{- end }}
+          volumeMounts:
+            # scripts/migrate-deploy.sh calls mktemp, and securityContext defaults to
+            # readOnlyRootFilesystem: true — without a writable /tmp the init container exits
+            # with "mktemp: Read-only file system" before any migration runs.
+            - name: tmp
+              mountPath: /tmp
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
         - name: copy-public
--- a/packages/super-sync-server/tests/migration-sql.spec.ts
+++ b/packages/super-sync-server/tests/migration-sql.spec.ts
@@ -454,6 +454,42 @@ describe('performance migrations', () => {
     expect(composeFile).toContain('- db');
   });
 
+  it('mounts a writable /tmp into the migrate-db initContainer so migrate-deploy.sh runs under readOnlyRootFilesystem', () => {
+    const helmDeployment = readFileSync(
+      join(currentDir, '../helm/supersync/templates/deployment.yaml'),
+      'utf8',
+    );
+    const helmValues = readFileSync(
+      join(currentDir, '../helm/supersync/values.yaml'),
+      'utf8',
+    );
+
+    const migrateDbStart = helmDeployment.indexOf('- name: migrate-db');
+    expect(migrateDbStart).toBeGreaterThan(-1);
+    // The migrate-db initContainer block ends at the next list item at the
+    // same indent (the following initContainer).
+    const nextContainer = helmDeployment.indexOf(
+      '\n        - name: ',
+      migrateDbStart + 1,
+    );
+    expect(nextContainer).toBeGreaterThan(migrateDbStart);
+    const migrateDbBlock = helmDeployment.slice(migrateDbStart, nextContainer);
+
+    // migrate-deploy.sh writes a temp log via `mktemp` under /tmp and runs
+    // with `set -eu`, so a read-only /tmp aborts the install before any
+    // migration runs. /tmp must therefore be backed by a writable volume
+    // mounted into THIS initContainer.
+    expect(migrateDbBlock).toContain('sh scripts/migrate-deploy.sh');
+    expect(migrateDbBlock).toContain('volumeMounts:');
+    expect(migrateDbBlock).toMatch(/- name: tmp\s+mountPath: \/tmp/);
+    // ...and the mounted volume must actually be declared as a writable
+    // emptyDir at the pod level.
+    expect(helmDeployment).toMatch(/- name: tmp\s+emptyDir:/);
+    // The mount is required precisely because the shared securityContext
+    // applied to this initContainer sets a read-only root filesystem.
+    expect(helmValues).toContain('readOnlyRootFilesystem: true');
+  });
+
   it('backfills operation payload bytes with per-user batched updates', () => {
     const script = readFileSync(
       join(currentDir, '../scripts/migrate-payload-bytes.ts'),


johannesjo:
Thanks — this is a correct diagnosis and the right fix. I reproduced the reasoning independently before looking at your patch:

- `values.yaml` sets `securityContext.readOnlyRootFilesystem: true`, and the template applies that same block to all three containers.
- `scripts/migrate-deploy.sh` runs `STMT_FILE="$(mktemp "${TMPDIR:-/tmp}/supersync-stmts.XXXXXX")"` at top level under `set -eu`, before the migrate loop. The image sets no `TMPDIR`, so it resolves to `/tmp`, and a failed command substitution aborts the script before any migration runs.
- `migrate-db` was the only container in the pod with no `volumeMounts` at all.

This has been broken since 2d9988dd73 (#7621, released in v18.10.0) introduced `mktemp` into the script — the chart's init container predates it and never had the mount. So every default `helm install`/`upgrade` since May.

I rendered the chart with your change (helm 3.20): the mount lands in the right block, indentation is correct, nothing else in the output moves. And to your question at the end — mounting is the right shape, not `TMPDIR`. There is no other writable path in that container, so redirecting `TMPDIR` would only relocate the same failure and leave `/tmp` read-only for anything else the script or Prisma touches. Reusing the existing `tmp` emptyDir instead of adding a volume or a new value is exactly right.

I also checked the neighbouring paths so we don't fix one of three: `copy-public` needs nothing (its `/tmp/public` is its own mount), the `database-url-check` pre-upgrade hook only runs `node -e` with no temp files, and `docker-compose.yml` never sets `read_only: true`, so the compose and host `deploy.sh` call sites were never affected. Helm was the only broken one — your scope is complete.

One thing before merge: a regression test, so nothing silently drops the mount again. `tests/migration-sql.spec.ts` already reads and asserts on the rendered deployment template, and `.github/workflows/supersync-server-tests.yml` runs `npm test` on every PR touching `packages/super-sync-server/**`, so it gets gated automatically there. Adding this after the `runs migrations before replacing the app during compose deploys` test does it — I verified it fails against the pre-fix template (`expected '- name: migrate-db...' to contain 'volumeMounts:'`) and passes with your mount:

```ts
  it('mounts a writable /tmp into the migrate-db initContainer so migrate-deploy.sh runs under readOnlyRootFilesystem', () => {
    const helmDeployment = readFileSync(
      join(currentDir, '../helm/supersync/templates/deployment.yaml'),
      'utf8',
    );
    const helmValues = readFileSync(
      join(currentDir, '../helm/supersync/values.yaml'),
      'utf8',
    );

    const migrateDbStart = helmDeployment.indexOf('- name: migrate-db');
    expect(migrateDbStart).toBeGreaterThan(-1);
    // The migrate-db initContainer block ends at the next list item at the
    // same indent (the following initContainer).
    const nextContainer = helmDeployment.indexOf(
      '\n        - name: ',
      migrateDbStart + 1,
    );
    expect(nextContainer).toBeGreaterThan(migrateDbStart);
    const migrateDbBlock = helmDeployment.slice(migrateDbStart, nextContainer);

    // migrate-deploy.sh writes a temp log via `mktemp` under /tmp and runs
    // with `set -eu`, so a read-only /tmp aborts the install before any
    // migration runs. /tmp must therefore be backed by a writable volume
    // mounted into THIS initContainer.
    expect(migrateDbBlock).toContain('sh scripts/migrate-deploy.sh');
    expect(migrateDbBlock).toContain('volumeMounts:');
    expect(migrateDbBlock).toMatch(/- name: tmp\s+mountPath: \/tmp/);
    // ...and the mounted volume must actually be declared as a writable
    // emptyDir at the pod level.
    expect(helmDeployment).toMatch(/- name: tmp\s+emptyDir:/);
    // The mount is required precisely because the shared securityContext
    // applied to this initContainer sets a read-only root filesystem.
    expect(helmValues).toContain('readOnlyRootFilesystem: true');
  });
```

Push that (or tell me and I'll add it) and this is good to merge. No chart version bump needed — the chart has stayed at 0.1.0 through #9223, #9212 and #9390, and it is installed from the checkout rather than a published registry.

johannesjo:
Thank you very much ! <3

Issue #9050, opened 2026-07-15
sync: complete deleteProject cascade recovery (concurrent-add, section/cfg timestamps, note todayOrder)


johannesjo:
Follow-up to #9037 / PR #9048, which recreates a losing `deleteProject`'s notes, sections, and task-repeat-cfgs at resolution time. That fix is a strict improvement but leaves three converging, documented residuals (all confirmed by two independent plan reviews + a Codex review of the implementation). None is a regression — each is either pre-existing loss the fix reduces, or a trade-off strictly better than losing the entity — so they were intentionally deferred. Tracking them here.

### 1. Same-batch concurrent-add not recovered (highest value)
`_createCascadeRecreationOpsForWinningProject` enumerates sections/repeat-cfgs from the **store** at resolution time. A section/cfg concurrently **added** on a third device, sequenced **before** the losing `deleteProject` in the same batch, is not yet reflected in the scanned store, so it is not recreated — and the loser's dynamic-filter replay (`contextId === projectId` / `cfg.projectId === projectId`) removes it. Result: that one concurrently-added entity is lost.

- Not a regression: without the fix, **all** project sections/cfgs are lost on a losing delete; the fix recreates the store-present ones and only misses a not-yet-applied add. Converges (no divergence).
- Note the asymmetry: notes and tasks don't have this because their delete cascade is a **static** id list (`noteIds` / `allTaskIds`), so a concurrent add isn't in the list and isn't removed. Sections/cfgs are removed by a **dynamic filter** at replay time, which is why a concurrent add gets caught.
- Fix sketch: overlay relevant `batchOps` `addSection`/`addTaskRepeatCfg` (and updates) targeting the project into the recreation set, merging their clocks, rather than reading only the pre-batch store.

### 2. Section / repeat-cfg LWW timestamp uses the project timestamp
`Section` and `TaskRepeatCfg` have no `modified` field, so the recreation op falls back to `projectCompensationOp.timestamp`. A stale snapshot can then clobber a genuinely concurrent `updateSection`/`updateTaskRepeatCfg` on another device. Converges; strictly better than losing the entity. A cleaner fix derives an entity-specific timestamp from the entity's latest known op with a conservative fallback that cannot beat a concurrent edit.

### 3. Pinned note `todayOrder` slot not restored
[...]

Issue #8416, opened 2026-06-15
plugins: permission declarations are documented but not enforced; harden side-panel iframe attribution


johannesjo:
_Found during review of #8291 (focused-task API) and #8394 (empty side-panel fix)._

### 1. Manifest permissions are advisory only

`src/app/plugins/plugin-bridge.service.ts` performs **no permission check** anywhere, yet `packages/plugin-api/README.md` and `docs/plugin-development.md` instruct plugin authors to declare permissions (e.g. for `getSelectedTask` / `getFocusedTask`). The new readers widen the task-content surface exposed to plugins (the full internal `Task` shape incl. `issueId`, `notes`, `timeSpentOnDay`, …) — **not a new leak** (`getTasks()` already returns the same shape), but the docs imply an enforcement boundary that doesn't exist.

**Action:** either enforce declared permissions in the bridge, or update the docs to state they're advisory.

### 2. Side-panel message attribution relies on a single-iframe-per-host invariant

After #8394, the tokenless `MESSAGE` path resolves the source window via `_getPluginIframeWindow()` → `querySelector('iframe[data-plugin-iframe]')` scoped to the component host (`src/app/plugins/ui/plugin-index/plugin-index.component.ts:~248-300`), and `isValidBridgeMessage` returns `true` for `MESSAGE` without checking `bridgeToken`/`bridgeGeneration` (`src/app/plugins/util/plugin-iframe.util.ts:~862-867`).

Correct **today** (one iframe per host, and the listener re-checks generation), but if a future change ever renders a second/nested iframe in that host, `querySelector` returns the first match and could mis-attribute a sibling's message. Harden by matching on `[data-plugin-id]` or asserting the single-iframe invariant. Not exploitable today.

https://claude.ai/code/session_01As7c32Eacxqksg5yHt5utw

Issue #5859, opened 2026-01-01
💥 Error: Project NhI6tSWFcGbvk2vJ2EAac not found


patrykpalej:
### Steps to Reproduce
I'm using the app via webdav on my server. When synced into it on my android (actually grapheneos) I got an error message on start with a link with "Please copy and report" which redirected me here. So I'm reporting 

### Url
https://localhost/#/project/NhI6tSWFcGbvk2vJ2EAac/tasks

### Meta Info
META: SP16.7.3_L2A __ Browser – en-US – Linux aarch64 – en-US – UA:Mozilla/5.0 (Linux; Android 16; Pixel 8 Pro Build/BP2A.250805.005; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/143.0.7499.146 Mobile Safari/537.36

### Actions Before Error
```
1767271594556: [Task] SetSelectedTask 
1767271594616: [SP_ALL] Load(import) all data 
1767271594619: [SP_ALL] All Data was loaded 
1767271594639: [Task] SetSelectedTask 
1767271740530: [Global Config] Update Global Config Section 
1767271970300: [Task] SetSelectedTask (2) 
1767271970355: [WorkContext] Set Active Work Context 
1767271970381: [Task] SetSelectedTask 
1767272455585: [Global Config] Update Global Config Section (3) 
1767272463349: [SP_ALL] Load(import) all data
```

johannesjo:
This will be fixed with the next release. Thank you very much!

PR #9525, opened 2026-08-09
feat(plugin-api): add deleteProject

### LINKED ISSUE #9524: Plugin API can create and update projects but not delete them (same for tags)
## What

`PluginAPI` exposes the full lifecycle for tasks but only two thirds of it for projects and tags:

| entity | create | update | delete |
|---|---|---|---|
| task | `addTask` | `updateTask` | **`deleteTask`** |
| project | `addProject` | `updateProject` | — |
| tag | `addTag` | `updateTag` | — |

(`packages/plugin-api/src/types.ts`)

A plugin can therefore make a mess it cannot clean up. Concretely: an integration that creates a project per external board, or an assistant bridge that creates one on request, has no way to remove it — the user has to finish the job by hand in the UI, which rather defeats the point of the integration.

## Why this looks unintentional rather than deliberate

The destructive capability is already granted for tasks, and `deleteTask` cascades to subtasks — so "plugins should not delete things" is not the rule in force.

The hard part of deleting a project — deciding what happens to its contents — is **already solved and centralised**. `ProjectService.remove()` collects the project's tasks, its backlog and their subtasks, clears their time-sync entries, and dispatches `TaskSharedActions.deleteProject`; effects then reset `defaultProjectId` to the Inbox if it pointed at the deleted project and clear note drafts. The UI's own "Delete project" menu item goes through exactly this.

So the semantics a plugin method would need are not an open question — they are the ones the application already applies to itself.

Note this is separate from `ALLOWED_PLUGIN_ACTIONS`, where `addProject` and friends are commented out. That list governs raw NgRx dispatch; the API methods go through `PluginBridgeService`, which already injects `ProjectService` and calls `add()`/`update()` on it.

## Proposal

Add to `PluginAPI`:

```ts
deleteProject(projectId: string): Promise<void>;
deleteTag(tagId: string): Promise<void>;   // same asymmetry, same argument
```

`deleteProject` delegating to `ProjectService.remove()` so the cascade stays defined in one place. Behaviour worth stating in the docs rather than leaving to discovery: **it deletes the project's tasks along with it**, exactly as the UI does.

Open questions for you, since they are policy rather than mechanics:

1. Should this sit behind its own manifest permission, so a plugin that only reads projects cannot delete one?
2. Should deleting the Inbox be refused outright? The UI does not offer it, but an API call could try. (My patch refuses it.)
3. Is `archiveProject` the safer thing to expose first? It is reversible, and for several use cases it is what the caller actually wants.

Heads-up as the template asks: I have a patch for `deleteProject` — four files, delegating to `ProjectService.remove()` — and will open it as a PR shortly. Say which of (1)–(3) you want and I will reshape it. `deleteTag` is deliberately not in that patch until the permission question above is answered.

sanchpet:
Closes #9524.

The plugin API gives tasks a full lifecycle but projects only two thirds of one — `addProject` and `updateProject` exist, delete does not. A plugin can create a project it then has no way to remove, so an integration that makes one per external board leaves the user to finish the job by hand in the UI.

`deleteTask` already exists and already cascades to subtasks, so "plugins should not delete things" is not the rule in force here.

## What it does

Nothing new about the semantics: `ProjectService.remove()` already collects the project's tasks, backlog and subtasks, clears their time-sync entries and dispatches `TaskSharedActions.deleteProject`; effects then reset `defaultProjectId` to the Inbox if it pointed at the deleted project and clear note drafts. The UI's own "Delete project" menu item takes that path, and this method delegates to the same place rather than assembling a second version of the cascade.

Deleting the Inbox is refused — it is the fallback target for tasks that belong nowhere, and the UI does not offer it either.

Four files: the interface, the bridge implementation, the `PluginAPI` facade and the iframe method list.

## What it deliberately does not do

`deleteTag` has exactly the same asymmetry and the same argument, but it raises the permission question I asked in #9524 — whether deletion should sit behind its own manifest permission. Left out until you say.

If you would rather see this behind a new permission, or would rather expose `archiveProject` first because it is reversible, say so and I will reshape it — both were offered in the issue.

## Verification

`npx tsc --noEmit -p src/tsconfig.app.json` passes. I checked that the typechecker actually covers the edited file rather than trusting a silent pass: introducing a deliberate typo in the new method produces the expected `TS2552`, and removing it returns to clean.


johannesjo (commented):
Thanks for the careful write-up — the issue-then-PR heads-up and the "what it deliberately does not do" section made this quick to review.

I checked the branch out on current master: `tsc --noEmit -p src/tsconfig.app.json` is clean and `npm run checkFile src/app/plugins/plugin-bridge.service.ts` passes. One blocking bug, plus a few things required by our own rules. Details below.

---

## On whether it should exist

The demand argument in #9524 is thin on its own — the issue is self-filed with no upvotes and no other participants — but the repo makes the case better than the issue does:

- `packages/plugin-dev/api-test-plugin/plugin.js:134` creates a project on every run and has no way to remove it. Our own test plugin leaks a project each time it executes.
- `packages/plugin-dev/todoist-import/src/map/run-import.ts:103` creates projects during import with no rollback path if the import fails partway.

And you're right that this is a genuinely new capability rather than a wrapper: `allowed-plugin-actions.const.ts:54-58` keeps `addProject`/`updateProject`/`archiveProject` commented out, so `dispatchAction` cannot reach it today.

So I'm not rejecting the premise. But see the policy notes at the bottom before you invest more in it.

---

## Blocking: deleting the project the user is currently viewing orphans subsequently-created tasks

The delegation to `ProjectService.remove()` is the right call, but it skips the step the UI performs *before* calling it:

```ts
// core-ui/work-context-menu/work-context-menu.component.ts:152-158
if (isConfirmed) {
  const activeId = this._workContextService.activeWorkContextId;
  if (activeId === this.contextId) {
    await this._router.navigateByUrl('/');   // awaited BEFORE remove()
  }
  this._projectService.remove(project);
}
```

That `await` is load-bearing. Nothing else resets the active context on a local delete:

- `work-context.reducer.ts` handles only `setActiveWorkContext` and `loadWorkContextState` — `TaskSharedActions.deleteProject` does not touch it.
- `work-context.effects.ts:83-115` (`validateContextAfterDataLoad$`) *does* reset to TODAY when the active project is gone, but only `ofType(loadAllData)`, i.e. sync/import (#5859). A local delete never dispatches it.
- `work-context.effects.ts:50-76` only fires on `schedule`/`planner`/`boards` routes.
- `project.effects.ts` delete handlers do config cleanup, note drafts and a snack — no navigation, no context reset.
- `ValidProjectIdGuard` is `canActivate` only (`app.routes.ts:34`), so it never re-runs on an already-active route.
- `work-context.service.ts:535` early-returns when the URL id equals the current `activeWorkContextId`, so even switching between the dead project's own tabs does not heal it.

So after `PluginAPI.deleteProject(<the project on screen>)`, and for as long as the user stays on that route:

1. `selectActiveWorkContext` (`work-context.selectors.ts:185-209`) silently substitutes a TODAY-tag context and logs `Log.err('Project not found: …')`. Its inputs include `selectTaskEntitiesInActiveProjects`, so that re-fires on every task edit and timer tick.
2. `activeWorkContextType` stays `PROJECT` with the dead id. `TaskService.add()` (`task.service.ts:402-410` → `createNewTaskWithDefaults`, `task.service.ts:1471-1484`) therefore stamps new tasks with `projectId = <deletedId>`, `tagIds: []` and no `dueDay`; `task-shared-crud.reducer.ts:96-100` then skips the project-list update because the entity is missing. **The task exists in task state but is in no project list and no tag — invisible, and it syncs that way.**
3. On reload, `ValidProjectIdGuard` returns `false` rather than a `UrlTree` (`app.guard.ts:101-114`) and there is no `withNavigationErrorHandler` (`main.ts:267`), so navigation is simply cancelled → blank content area.

Not a contrived trigger: a board-integration plugin deleting the project the user happens to be looking at is exactly the use case the issue describes.

`PluginBridgeService` already injects `Router` (`:159`) and `WorkContextService` (`:153`), so mirroring the UI is a three-line fix in the file you already touch. But I'd rather have the effect-based version: an effect on `TaskSharedActions.deleteProject` (injecting `LOCAL_ACTIONS`, per our sync rules) that resets the context when `activeId === projectId`. It is a no-op for the existing UI path, which has already navigated by then, and it also closes a pre-existing hole on that path: if `misc.defaultStartPage` points at the project being deleted, the UI's `navigateByUrl('/')` resolves straight back to that same project, `work-context.service.ts:535` early-returns, and the UI lands in the same broken state — the `defaultStartPage` cleanup only runs afterwards in `project.effects.ts:51-55`. If you take that route, please add a test for the plugin path specifically.

---

## Required before merge

**Wiki update, same PR.** `docs/documentation-guide.md` maps "New or changed REST / Plugin / Sync API surface" → `docs/wiki/3.01-API.md`. Needs `docs/wiki/3.01-API.md:67` and `docs/plugin-development.md:307-308`. The cascade semantics deserve a sentence there the way `deleteTask` gets one at `3.01-API.md:61` — your own issue makes this point ("worth stating in the docs rather than leaving to discovery").

**Translated error strings.** `Project not found: ${projectId}` is hardcoded English. `T.PLUGINS.PROJECT_NOT_FOUND` already exists (`Project "{{contextId}}" not found`) and is already used in this very file at `plugin-bridge.service.ts:1110`; `deleteTask` uses `T.PLUGINS.TASK_NOT_FOUND` at `:1006`. Same for the Inbox refusal — please add a key rather than a literal. Edit `en.json` only.

**Tests.** The two guards (Inbox refusal, not-found) are cheap to cover in `plugin-bridge.service.spec.ts`. I know the neighbours have none, but new destructive logic is the wrong place to extend that precedent — and the context-reset fix above needs one regardless.

---

## Policy answers to your open questions

**(1) Permission gating.** Leaving it ungated matches today's reality — only `http` and `nodeExecution` are actually enforced — but #8416 is open precisely about that gap, and the most destructive method in the API is not where I want to widen it by default. I don't want to block this PR on building permission enforcement; I do want the method to land after that, not before, if we go that way. Let's settle it in #9524 first.

**(3) `archiveProject` first.** Reversible, yes, but not the cheap swap it looks like: the effects for it are commented out at `project.effects.ts:74-104`, so exposing it means implementing it.

**Sync blast radius.** `deleteProject` is a known-fragile cascade — #9050 is open with three documented residual data-loss cases following #8997/#9037. A human clicking through a confirm dialog hits those races rarely; a plugin calling it on a schedule or in a sync loop raises the frequency. Not a reason to decline on its own, but it belongs on the cost side.

**Service size.** `plugin-bridge.service.ts` is 2252 lines against our 1200 cap, grandfathered to a warning at `eslint.config.js:324`. +26 lines is small and this is the method's natural home, so I won't block on it — just flagging the direction of travel.

---

## Unrelated, spotted while reading

`addProject` returns `Promise<string>` (`packages/plugin-api/src/types.ts:667`), so in `packages/plugin-dev/api-test-plugin/plugin.js:139-145` both `newProject.title` and `newProject.id` are `undefined` and the `updateProject` branch silently never runs. Pre-existing and out of scope here — separate PR welcome if you feel like it.

---

So: the context-reset bug is the thing that has to be fixed either way, and the docs/i18n/tests items are mechanical. The permission question is the one that decides whether this merges at all, and I'd like to answer it in #9524 before you spend more time here. Sorry for holding it up on policy after you'd already written the code — but that's the part I need to get right, since the plugin API is hard to walk back once plugins depend on it.

sanchpet:
Thanks — that review found the real hole, and the `defaultStartPage` case you spotted makes it a bug on the UI path, not only on mine.

So I split it out: **#9602** carries the context reset alone, off current master, so it can land regardless of what happens here. Effect on `TaskSharedActions.deleteProject` via `LOCAL_ACTIONS`, three cases in `work-context.effects.spec.ts`; with the effect's guard forced to `false` the reset test times out, so it fails without the fix.

What is left on this branch, rebased on master:

- wiki `3.01-API.md` and `plugin-development.md`, with the cascade stated rather than left to discovery
- `T.PLUGINS.PROJECT_NOT_FOUND` for the not-found case and a new `CANNOT_DELETE_INBOX` key for the refusal (`en.json` only)
- tests for both guards and for the delegation to `ProjectService.remove()`

On permissions: your call, and I agree with it — the most destructive method in the API is not the place to widen the ungated surface, so I would rather this land after enforcement than before. Marking the PR draft until #8416 and #9524 settle. If you decide it should carry a manifest permission, say which shape you want and I will add it here.

I will send the `api-test-plugin` `addProject` fix separately.

Claude Code was used to prepare these changes; I reviewed and tested the result.

johannesjo (commented):
Round two — everything from the first review is in, and I re-checked it on the merge with current master: `tsc --noEmit -p src/tsconfig.app.json` clean, `plugin-bridge.service.spec.ts` 38/38 including your three new cases, `checkFile` clean on the bridge, all 28 CI checks green.

- The blocking context bug is gone via #9602 (`50110426fa`), so the docs' "falls back to the Today context" is accurate as written. Splitting it out was the right call — it stands on its own and it fixed the `defaultStartPage` case on the UI path too.
- Wiki + `plugin-development.md`, `CANNOT_DELETE_INBOX` / reuse of `PROJECT_NOT_FOUND` (`en.json` only), guard tests: all there, nothing left on that list.

## Answer to (1): gate it, and then this can merge

Add the permission, and this goes in. But **the check cannot live in `plugin-api.ts`**, even though that is the class holding `#manifest` — iframe plugins never reach it. `plugin-iframe.util.ts:717-738` tries `boundMethods[method]` first and otherwise falls back to `bridge[method](...args)` with no plugin context at all. There is no bound-method entry for any of the project CRUD, so today every iframe plugin lands on `PluginBridgeService.deleteProject` directly and would sail past a facade-level check.

The shape that actually holds is the one `request` already uses:

- add a `deleteProject` entry to `createBoundMethods` (`plugin-bridge.service.ts:238`, which has `pluginId` and `manifest` in scope) that passes `manifest?.permissions` down, mirroring `request` at `:389-390`;
- have `PluginAPI.deleteProject` call `#boundMethods.deleteProject` instead of `#pluginBridge.deleteProject`, the way `request` does at `plugin-api.ts:632-637`;
- one test for an undeclared plugin being rejected, alongside the two guard tests you already have.

Name the permission `deleteProject` and add it to the list in `packages/plugin-api/README.md:122-141`. That is the whole ask — it does not need any of #8416 built first, and it gives that issue a first concrete enforcement precedent rather than one more ungated method to retrofit later.

## Why I want it gated after all, having said permissions aren't enforced

Because this is the only *irreversible* method in the API, which I hadn't weighed properly first time round. `deleteTask` — the precedent the issue leans on — dispatches `restoreDeletedTask` behind an undo snack (`task-ui.effects.ts:119-140`). There is no `restoreDeletedProject`; `showDeletionSnack` (`project.effects.ts:157-169`) does fire on your path, but it offers no action. So a plugin bug takes the project, its backlog, its subtasks and its notes with a five-second toast and no way back — and `PROJECT_DELETE_WINS_MARKER` carries that to every device, where #9050's first residual can additionally eat a section or repeat-cfg another device added concurrently.

Being honest about what the permission buys: for iframe plugins `allow-same-origin` already exposes `window.parent.ea` (documented at `plugin-iframe.util.ts:23-33`), so this is install-time disclosure, not containment. For the most destructive method in the API, disclosure is still worth ten lines.

## Not blocking, but worth knowing

**The URL stays dead.** #9602 resets the context *state*, but nothing navigates, so after `deleteProject(<project on screen>)` the URL is still `/project/<deletedId>/tasks`. Reload there and `ValidProjectIdGuard` (`app.guard.ts:96-114`, `canActivate`-only) returns `false`, and with no `withNavigationErrorHandler` at `main.ts:267` the navigation is simply cancelled — empty content area, and the `**` route does not catch it because the path matched. Pre-existing and reachable from the UI path in the `defaultStartPage` corner, but your method makes it the ordinary outcome. Cheapest fix is the guard returning a `UrlTree` to Today instead of `false`; that is independently useful, so please send it as its own PR like you did with #9602 rather than growing this one.

**Two nits in the new method**, neither worth a round trip on its own:

- `typia.assert<string>` accepts `''`, and `getByIdOnce$('')` throws an untranslated `Error('No id given')` (`project.service.ts:355`) before your translated `PROJECT_NOT_FOUND` can fire.
- `ProjectService.remove()` calls `getTaskById`, which throws on a stale id in `taskIds`/`backlogTaskIds` (`task.reducer.util.ts:18-24`). Pre-existing, but a plugin receiving a raw `Task not found: <id>` is worse than a user clicking a menu item, so it may be worth catching and rewrapping.

## `deleteTag`

Same shape, same permission story — send it as its own PR once this one lands with the gate, and I'll take it on the same terms.

No need to keep this in draft; the only thing standing between it and merge is the permission wiring above.

sanchpet:
Permission wiring is in — `8b6381d6e`.

Done exactly as you laid out:

- `deleteProject` entry added to `createBoundMethods`, passing `manifest?.permissions` down, mirroring `request`
- `PluginAPI.deleteProject` now calls `#boundMethods.deleteProject` instead of the bridge directly
- the check itself sits in `PluginBridgeService.deleteProject`, so the iframe fallback path is gated too rather than only the facade
- permission named `deleteProject`, listed in `packages/plugin-api/README.md`

Two tests rather than one, because the wiring is the part that can silently rot: one for an undeclared plugin being rejected, and one going through `createBoundMethods` itself — granted manifest deletes, empty-permissions manifest is refused. That second one is what would fail if someone later re-pointed the facade at the bridge and left iframe callers ungated. The three existing guard tests now declare the capability. Locally: `tsc --noEmit -p src/tsconfig.app.json` clean, `plugin-bridge.service.spec.ts` 40/40, eslint clean apart from the pre-existing `max-lines` warning.

On your two nits — agreed on both, and I'd rather not grow this PR with them: `typia.assert<string>` letting `''` through into the untranslated `No id given`, and `ProjectService.remove()` surfacing a raw `Task not found: <id>` from a stale id. Both are argument-hygiene at the same boundary, so they belong together in one follow-up rather than split across this one.

Taking the `ValidProjectIdGuard` → `UrlTree` fix as its own PR as you suggested, same as #9602. `deleteTag` after this lands, on the same terms.

johannesjo:
**✅ Ready to merge** – no blocking issues at `8b6381d`. Merge confidence 85% – I read both call paths but did not run the suite locally or exercise a real plugin against the gate; CI's green `Tests` job is the evidence there.

Thanks for the follow-up. The permission wiring is the shape that was asked for, and it holds up where it matters:

- the `createBoundMethods` entry passes `manifest?.permissions` down the same way `request` does, and the check itself sits in `plugin-bridge.service.ts:1073` rather than the facade, so `plugin-iframe.util.ts:714-737` (bound method first, `bridge[method]` otherwise) is fail-closed on both branches;
- all three `createBoundMethods` call sites – `plugin-api.ts:90`, `plugin-iframe.util.ts:611`, `plugin-index.component.ts:237` – pass a real manifest, so no plugin is accidentally locked out and no ungated route to the bridge method is left;
- `dispatchAction` offers no way around it: the project actions are still commented out in `allowed-plugin-actions.const.ts`;
- declaring a new `deleteProject` string neither trips `validate-manifest.util.ts` (no permission allowlist) nor renders oddly in the plugin management list.

The two doc claims the PR adds also still hold on current master: `ProjectService.remove()` dispatches the cascade with backlog and subtask ids, and `work-context.effects.ts:93` resets the active context to Today on a local delete.

### Nits

- **`packages/plugin-api/src/types.ts:672`** (also `docs/plugin-development.md:309`, `docs/wiki/3.01-API.md:67`) – the permission requirement isn't stated at the method itself. The docs commit landed before the gating commit, so all three describe the cascade, the Inbox refusal and the Today fallback without mentioning that the manifest must declare `deleteProject`. The repo states it inline for both other enforced permissions (`types.ts:759` for `request`, `3.01-API.md:93` for `executeNodeScript`), and `3.01-API.md:123` points readers at the development guide, which lists no per-method permissions. One clause in each spot, mirroring `types.ts:759`, closes it. Worth folding in while you're there: `README.md:138` does carry the entry, but every other item in that list is advisory and enforced by nothing, and the two permissions that *are* enforced aren't in it at all – so nothing distinguishes yours as a hard requirement. The runtime message already names the fix, which is why this is a nit rather than a blocker.

---

johannesjo:
Thank you very much! <3

PR #9542, opened 2026-08-10
fix(ios): request local network permission for LAN sync servers

Juuze:
## Problem

The iOS app never asks for local network permission, so syncing with a sync server on the local network fails with no actionable error.

In the app the failure surfaces only as:

> SuperSync native request failed: Error

The iOS app logs show the request failing in the native networking layer, without usable detail:

```json
{
  "timestamp": "2026-08-10T21:52:27.594Z",
  "level": "WARN",
  "context": "ol",
  "message": "SuperSyncProvider transient network error, retrying in 3000ms (attempt 2/2)",
  "args": [
    {
      "url": "<URL_OF_SELF-HOSTED_SUPERSYNC>",
      "attempt": 2,
      "maxRetries": 2,
      "delayMs": 3000,
      "errorName": "Error",
      "errorCode": "NSURLErrorDomain"
    }
  ]
}
```

`NSURLErrorDomain` places the failure in native `URLSession` rather than in the web view, and no local network prompt was ever shown for the app.

iOS 14+ only presents that prompt if the app declares `NSLocalNetworkUsageDescription` in `Info.plist`. That key was missing, so the system never asked, and requests to a self-hosted WebDAV/Nextcloud or SuperSync server on the LAN failed the way shown above.

Both native HTTP paths are affected: `WebDavHttpPlugin` for WebDAV, and `CapacitorHttp` (`enabled: true` in `capacitor.config.ts`), which routes `fetch`/`XHR` through native `URLSession`. Because that traffic originates in the app process, it is subject to local network privacy — so the missing key is what suppressed the prompt.

Found while testing on a device; there is no existing issue for it.

## Solution

Add `NSLocalNetworkUsageDescription` to `ios/App/App/Info.plist`, next to the existing `NSUserNotificationsUsageDescription` and worded in the same style.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki). — n/a, no wiki page documents the iOS permission prompts
- [ ] I have run `npm run checkFile` on changed `.ts`/`.scss` files — n/a, no `.ts`/`.scss` files changed
- [ ] I have added tests for my changes (if applicable) — n/a, `Info.plist` is not unit-testable
- [x] Existing tests still pass — unchanged by this PR: no `.ts`/`.scss`/test input is touched, so lint and the unit suite cannot be affected. Left to CI to confirm.

## Verification

- Not verified locally: the prompt itself. Confirming that requires an iOS build on a device/simulator.

### DIFF
--- a/ios/App/App/Info.plist
+++ b/ios/App/App/Info.plist
@@ -49,6 +49,8 @@
 	<true/>
 	<key>NSUserNotificationsUsageDescription</key>
 	<string>Super Productivity uses notifications to remind you about tasks and scheduled events.</string>
+	<key>NSLocalNetworkUsageDescription</key>
+	<string>Super Productivity uses your local network to sync with sync servers you host yourself, such as a WebDAV or Super Sync server on your network.</string>
 	<key>UIBackgroundModes</key>
 	<array>
 		<string>fetch</string>


johannesjo:
Thanks for this — and for the unusually careful write-up, including explicitly flagging what you *didn't* verify. That's appreciated.

I dug into the two iOS mechanisms that can produce a bare `NSURLErrorDomain` failure against a LAN host, because they need different fixes:

**Local Network Privacy (LNP, iOS 14+)** — requires `NSLocalNetworkUsageDescription`. Without the key the system never prompts and the connection is refused. It applies to plain unicast HTTP to a LAN address, not just Bonjour/mDNS browsing — so your reasoning holds, and it covers both code paths you name: `WebDavHttpPlugin` builds its own `URLSession`, and `CapacitorHttp` is `enabled: true` in `capacitor.config.ts`, so `fetch`/XHR go native too. LNP denial surfaces as a **transport** error: `NSURLErrorNotConnectedToInternet` (-1009), or the explicit "Local network prohibited".

**App Transport Security** — a separate system, worth ruling out. Since iOS 10 ATS does *not* apply to raw IP addresses, `.local` names, or unqualified hostnames, so `http://192.168.x.x:3000` and `http://nas.local:2701` are not ATS-blocked and need no `NSAllowsLocalNetworking`. ATS *does* still block cleartext to a fully-qualified name, so an `http://sync.example.com` pointing at a LAN box would fail — but with a **TLS** error (`NSURLErrorSecureConnectionFailed` -1200 / -1022), not a transport one.

So the change is correct, and for the common LAN-by-IP case it's the right and only platform-level fix. `NSBonjourServices` is correctly omitted — we don't browse service types.

Two things before I merge:

1. **Can you get the numeric error code off the device?** The log you posted shows `errorCode: "NSURLErrorDomain"`, which is the error *domain*, not the code — so it doesn't yet discriminate LNP from ATS from a plain TLS/cert failure. `-1009` or "Local network prohibited" confirms this PR is the fix; `-1200`/`-1022` would point elsewhere. Not a blocker for merging — the key is required for LAN access regardless — but it decides whether your original problem is actually closed by this.

2. **Please confirm the prompt on a real device, not the Simulator.** The Simulator doesn't reliably enforce LNP; there are plenty of reports of the prompt never appearing under Xcode 16 / iOS 18. Also note `build-ios.yml` only runs on tag push, so nothing in CI will validate this either way.

Worth knowing for anyone retesting: LNP answers are cached per app and a previous "deny" sticks. If the prompt doesn't reappear, toggle Settings → Privacy & Security → Local Network, or restart the device (there was an iOS 18.4 bug where a deny persisted even across reinstall, fixed in 18.6).

Nit on the usage string — it reads a little redundant ("uses your local network to sync with sync servers … on your network"), and App Review prefers these tight:

```xml
<string>Super Productivity connects to sync servers on your local network, such as a self-hosted WebDAV or Super Sync server.</string>
```

Otherwise this is the kind of PR I like: two lines, no dependency, no new setting, and a commit message that documents the non-decision (`NSBonjourServices`) alongside the decision. 👍

PR #9544, opened 2026-08-11
feat(api): expose focus mode state

aaronedev:
## Problem

External desktop clients cannot read the authoritative Focus Mode timer state through the opt-in
Local REST API. This blocks a companion bar from showing the native timer without recreating timer
logic. Related: https://github.com/aaronedev/dotfiles/issues/247

## Solution

Add authenticated, read-only `GET /focus` to the existing renderer Local REST request pipeline.
Its data contract is:

```ts
{
  mode: "Flowtime" | "Pomodoro" | "Countdown";
  cycle: number;
  isSessionDone: boolean;
  timer: null | {
    purpose: "work" | "break";
    status: "running" | "paused" | "done";
    isOvertime: boolean;
    isLongBreak: boolean;
    elapsedMs: number;
    remainingMs: number;
    durationMs: number;
  };
}
```

The handler reads the existing Focus Mode selectors as one coherent store sample. Completed work
uses `isSessionDone` to distinguish its idle timer from an untouched session; a completed break
retains its timer with `status: "done"`. `isOvertime` is independent from status, including while
an overtime work timer is paused. The endpoint returns no task content or credentials and adds no
focus controls: start, pause, resume, skip, and cancel remain native-app behavior. It inherits the
established opt-in Local REST API and Bearer authentication; `GET /health` remains the only
unauthenticated exception.

Focused tests use real Focus Mode state and cover idle, running and paused work, completed work,
running, paused, and completed breaks, Flowtime, overtime, and realistic cycle values. The API
reference documents the authenticated curl usage, state semantics, and 1-based Pomodoro cycle.

## Type of Change

- [x] New feature

## Checklist

- [x] Existing focused tests still pass

## Validation

- `npm run checkFile src/app/core/electron/local-rest-api-handler.service.ts`
- `npm run checkFile src/app/core/electron/local-rest-api-handler.service.spec.ts`
- `npm run test:file src/app/core/electron/local-rest-api-handler.service.spec.ts` (80/80 passing)
- `npm run docs:check-links`
- `git diff --check`

## Remaining dependency

The companion integration in aaronedev/dotfiles will capability-detect this endpoint. It can use
native focus state after this PR is reviewed, merged, and released; this PR intentionally does not
add focus control endpoints.

### DIFF
--- a/docs/wiki/3.01-API.md
+++ b/docs/wiki/3.01-API.md
@@ -355,29 +355,40 @@ curl -X POST http://127.0.0.1:3876/task-control/stop
 does not expose task content and does not add controls for starting, pausing,
 resuming, skipping, or cancelling a focus session.
 
+```bash
+curl --noproxy 127.0.0.1 -H "@$HOME/.sp-api-header" http://127.0.0.1:3876/focus
+```
+
 **Response data:**
 
 ```typescript
 {
   mode: 'Flowtime' | 'Pomodoro' | 'Countdown';
-  cycle: number; // non-negative
+  cycle: number;
+  isSessionDone: boolean;
   timer: null | {
     purpose: 'work' | 'break';
-    isRunning: boolean;
-    isPaused: boolean;
-    elapsedMs: number; // normalized non-negative
-    remainingMs: number; // normalized non-negative
-    durationMs: number; // normalized non-negative
+    status: 'running' | 'paused' | 'done';
+    isOvertime: boolean;
     isLongBreak: boolean;
+    elapsedMs: number;
+    remainingMs: number;
+    durationMs: number;
   };
 }
 ```
 
-`timer` is `null` when no work or break session exists. A paused work session or
-break still has a timer object with `isRunning: false` and `isPaused: true`.
-`elapsedMs`, `durationMs`, and `remainingMs` are normalized to non-negative values.
-`remainingMs` is derived from the same timer sample as `durationMs` and `elapsedMs`,
-so it is zero for Flowtime and after a countdown has reached its duration.
+`timer` is `null` when no work or break timer exists. A completed work session is
+represented by `timer: null` and `isSessionDone: true`; a completed break retains its
+break timer with `status: 'done'`. Other active timer states are `running` or `paused`.
+`isOvertime` is independent from `status`, so it remains true when an overtime work
+session is paused. Flowtime timers have no countdown duration, so their `remainingMs`
+is zero while their status remains `running` or `paused`.
+
+`cycle` is Pomodoro-only and 1-based. It advances when a Pomodoro work session
+completes, so it is already advanced during that session's following break. Flowtime
+and Countdown keep it at `1`. Focus Mode state is not persisted, so the cycle resets to
+`1` when the app restarts and when the user explicitly resets the Pomodoro cycle count.
 
 ---
 
--- a/docs/wiki/3.01-API.md
+++ b/docs/wiki/3.01-API.md
@@ -385,10 +385,11 @@ break timer with `status: 'done'`. Other active timer states are `running` or `p
 session is paused. Flowtime timers have no countdown duration, so their `remainingMs`
 is zero while their status remains `running` or `paused`.
 
-`cycle` is Pomodoro-only and 1-based. It advances when a Pomodoro work session
-completes, so it is already advanced during that session's following break. Flowtime
-and Countdown keep it at `1`. Focus Mode state is not persisted, so the cycle resets to
-`1` when the app restarts and when the user explicitly resets the Pomodoro cycle count.
+`cycle` is the 1-based counter used by Pomodoro. Only completing a Pomodoro work
+session advances it, so it is already advanced during that session's following break.
+Switching to Flowtime or Countdown retains the current value, but those modes do not
+advance it. Focus Mode state is not persisted, so the cycle resets to `1` when the app
+restarts and when the user explicitly resets the Pomodoro cycle count.
 
 ---
 
--- a/docs/wiki/3.01-API.md
+++ b/docs/wiki/3.01-API.md
@@ -356,7 +356,7 @@ does not expose task content and does not add controls for starting, pausing,
 resuming, skipping, or cancelling a focus session.
 
 ```bash
-curl --noproxy 127.0.0.1 -H "@$HOME/.sp-api-header" http://127.0.0.1:3876/focus
+curl http://127.0.0.1:3876/focus
 ```
 
 **Response data:**
@@ -381,6 +381,8 @@ curl --noproxy 127.0.0.1 -H "@$HOME/.sp-api-header" http://127.0.0.1:3876/focus
 `timer` is `null` when no work or break timer exists. A completed work session is
 represented by `timer: null` and `isSessionDone: true`; a completed break retains its
 break timer with `status: 'done'`. Other active timer states are `running` or `paused`.
+`isSessionDone` mirrors the current Focus Mode screen, so it can remain true after the
+overlay is dismissed until another focus action changes the screen.
 `isOvertime` is independent from `status`, so it remains true when an overtime work
 session is paused. Flowtime timers have no countdown duration, so their `remainingMs`
 is zero while their status remains `running` or `paused`.
--- a/docs/wiki/3.01-API.md
+++ b/docs/wiki/3.01-API.md
@@ -365,9 +365,9 @@ resuming, skipping, or cancelling a focus session.
     purpose: 'work' | 'break';
     isRunning: boolean;
     isPaused: boolean;
-    elapsedMs: number;
-    remainingMs: number; // clamped to zero
-    durationMs: number;
+    elapsedMs: number; // normalized non-negative
+    remainingMs: number; // normalized non-negative
+    durationMs: number; // normalized non-negative
     isLongBreak: boolean;
   };
 }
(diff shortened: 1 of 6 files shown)


aaronedev:
Opened [#9549](https://github.com/super-productivity/super-productivity/issues/9549) as the intentional follow-up for capability-declared authenticated Focus Mode controls. This PR remains the separate, useful read-only `GET /focus` contribution.

aaronedev:
https://github.com/super-productivity/super-productivity/issues/9549#issuecomment-5251937798

johannesjo:
Thanks for this, and for the unusually thorough PR description — it made review easy.

I want the endpoint: `/focus` is a reasonable addition to the Local REST API, and the implementation is clean. Tests pass (75/75), lint and prettier are clean, the Bearer auth is correctly inherited (only `/health` is exempt, and the server forwards any path, so no electron-side change was needed), and nothing user-content-bearing is in the response.

All of my feedback is about the **response shape** rather than the code, because that's the part we can't take back. The rule I'd like to apply: adding a field later is free, changing what a field *means* is not.

### 1. There's no way to tell "finished" from "paused"

This is the main one. Breaks never auto-complete — `detectBreakTimeUp$` is explicitly `{ dispatch: false }` ("Detect when break timer completes and show notification (no auto-complete)"), and the tick reducer just stops the timer at duration and waits for the user. So a finished break sits indefinitely as `isRunning: false, isPaused: true, remainingMs: 0`, and a companion bar would render "break paused 0:00" forever instead of "break over".

Work sessions have the mirror-image problem: `detectSessionCompletion$` dispatches `completeFocusSession`, which resets the timer to idle — so a finished Countdown session reports `timer: null`, indistinguishable from "never started", while the app is showing the SessionDone screen and banner.

Same real state ("waiting on the user"), two different wrong answers. I'd like the contract to be able to represent it:

```ts
{
  mode: 'Flowtime' | 'Pomodoro' | 'Countdown';
  cycle: number;
  isSessionDone: boolean;   // selectIsSessionCompleted
  timer: null | {
    purpose: 'work' | 'break';
    status: 'running' | 'paused' | 'done';
    isOvertime: boolean;
    isLongBreak: boolean;
    elapsedMs: number;
    remainingMs: number;
    durationMs: number;
  };
}
```

One `status` enum rather than more booleans — `isRunning`/`isPaused`/`isDone` can encode combinations that can't actually occur, and clients then have to guess the precedence. `isOvertime` does need to stay a separate flag rather than a `status` value: `selectIsInOvertime` deliberately stays true across a pause (#7715), so folding it in would lose "paused during overtime".

### 2. Please document `cycle` precisely

It's Pomodoro-only (`incrementCycleOnSessionComplete$` filters on the mode), so it stays at `1` forever in Flowtime and Countdown. It's 1-based, it is deliberately already advanced during the break that follows a session — that's the off-by-one `getBreakCycle()` compensates for — and focus-mode state isn't persisted, so it resets to 1 on every app restart as well as on the user's explicit "reset cycles". "non-negative" doesn't prepare a client for any of that. A doc note is enough here; no code change needed.

### 3. Please go back to the existing selectors

3355da8 replaced `selectIsSessionPaused` / `selectIsLongBreak` / `selectTimeRemaining` with inline re-derivations. They're behaviour-identical today so it isn't a bug, but it puts a second definition of focus-timer semantics outside `focus-mode/`, and that's exactly what drifts apart over time.

The stated motivation doesn't hold either: inside `Promise.all`, each `firstValueFrom(this._store.select(...))` subscribes synchronously and NgRx's `State` is a `BehaviorSubject`, so all of them read the same snapshot with no `await` in between — the original version was already one coherent sample.

Reverting also makes the `Math.max(0, …)` normalisation unnecessary (`selectTimeRemaining` already clamps), and I'd rather the endpoint report exactly what the app itself displays than normalise to a second opinion. `Math.max(0, cycle)` in particular guards a state the reducer cannot produce — please drop that one.

### 4. Drive the spec through real state

With `overrideSelector` on every selector the handler reads, the tests check the re-derivation against itself and never execute a real selector, so a semantics change in focus-mode would leave them green while the API drifts. Seeding `provideMockStore({ initialState: { focusMode: … } })` and using `setState` costs about the same and would actually catch it. Mostly falls out of (3) anyway.

### Small one

Every other section in `3.01-API.md` shows a curl example with the `Authorization` header — worth matching for consistency.

---

Happy to do this in one round. On #9549: let's settle this read shape first, then design the controls on top of it.

aaronedev:
Thank you for getting back. I will follow shortly.

aaronedev:
Thanks — I updated the branch around the stable states you described.

- `isSessionDone` now distinguishes completed work from an idle timer.
- Timer state is `running | paused | done`, with `isOvertime` kept separate. A stopped break with
  time remaining is paused; at zero it is done.
- The handler reads the existing focus selectors instead of re-deriving pause, remaining-time,
  long-break, overtime, or completion semantics.
- The spec now seeds real `focusMode` state and includes a reducer-driven cycle-retention case. The
  API reference documents the 1-based Pomodoro cycle, retained values in other modes, and
  authenticated curl usage.

Validation: the focused handler spec passes 80/80; `checkFile` passes for the handler and spec;
documentation links and `git diff --check` pass.

Focus controls remain separate in #9549.

johannesjo:
Thanks — this is right, and the response shape is settled from my side.

I checked out the branch rather than just reading the diff:

- `npm run test:file src/app/core/electron/local-rest-api-handler.service.spec.ts` → 80/80
- `checkFile` clean on the handler, `prettier --check` clean on `3.01-API.md`, `docs:check-links` clean
- the spec has zero `overrideSelector` now, so the real selectors actually execute — that was the whole point of (4) and it's properly done
- handler service is 743 lines, comfortably under the `max-lines` cap

Three small things, then I'll merge.

### 1. One last piece of semantics still lives outside `focus-mode/`

This is (3) again, applied to the bit that's left. `focus-mode.effects.ts:960` already owns "the break is over", and it's the definition the *user* experiences — it drives `detectBreakTimeUp$` (end-of-break notification) and `breakEndAlarmSound$` (the looping alarm):

```ts
private _isBreakTimeUp(timer: TimerState): boolean {
  return timer.purpose === 'break' && !timer.isRunning
    && timer.startedAt !== null && timer.elapsed >= timer.duration;
}
```

The handler re-derives the same thing inline as `isSessionPaused && timer.purpose === 'break' && remainingMs === 0`. They agree today — `startedAt` is never null for a break, since `createBreakTimer` and `restoreFocusSessionFromNative` both always set it — so this isn't a bug. But it's the same drift risk as before, and here it means the alarm the user hears and the API could end up disagreeing about whether a break is over.

Promoting it to a selector is small. I tried it locally to be sure it's clean:

```ts
// focus-mode.selectors.ts
export const selectIsBreakTimeUp = createSelector(
  selectTimer,
  (timer) =>
    timer.purpose === 'break' &&
    !timer.isRunning &&
    timer.startedAt !== null &&
    timer.elapsed >= timer.duration,
);
```

with the two effects reading `selectors.selectIsBreakTimeUp.projector(timer)` and the handler reading `selectIsBreakTimeUp(state)`. 561/561 focus-mode specs and 80/80 handler specs stay green.

It also lets `selectIsSessionPaused` go, which is currently dead in the handler: inside the `timer.purpose !== null` branch `selectIsRunning` reduces to `timer.isRunning`, so `selectIsSessionPaused` reduces to `!timer.isRunning` — meaning the `isSessionPaused &&` term can never be false anywhere it's actually evaluated. It reads like a guard but isn't one.

### 2. `isSessionDone` is sticky — worth a sentence next to the cycle note

`selectIsSessionCompleted` is `currentScreen === 'SessionDone'`, and `hideFocusOverlay` deliberately leaves `currentScreen` alone. In Countdown nothing auto-starts a break (`autoStartBreakOnSessionComplete$` filters to Pomodoro), so once a session ends and the user dismisses the banner and goes back to work, `GET /focus` keeps returning `{ isSessionDone: true, timer: null }` until an explicit `startFocusSession` / `selectFocusTask` / `navigateToMainScreen` / `startBreak` / `cancelFocusSession`.

That's faithful to app state, so no code change — but a companion bar polling this would render "session done" indefinitely, and that deserves the same kind of note the cycle got.

### 3. The curl example — my mistake, sorry

I asked for the `Authorization` header and I had it backwards. `3.01-API.md:179` says:

> The examples below omit the `Authorization` header for brevity — every one of them needs it except `/health`.

and every sibling example (`/tasks`, `/task-control`, `/projects`, `/tags`) is a bare `curl http://127.0.0.1:3876/…`. As it stands, `/focus` at :359 is the only one in the file carrying the header. Please just match the others:

```bash
curl http://127.0.0.1:3876/focus
```

### Optional, entirely your call

The nine `firstValueFrom` calls under `Promise.all` can collapse to one await, since selectors are plain functions of root state:

```ts
const state = await firstValueFrom(this._store);
const timer = selectTimer(state);
const mode = selectMode(state);
// …
```

Provably one sample rather than nine that happen to coincide, and the `Promise.all` disappears. Also compiles clean and stays 80/80 — I checked. Take it or leave it.

---

Once those three are in this goes in, and #9549 can build on the settled shape.

aaronedev:
Implemented the three follow-ups in 9b1099f9:

- Moved completed-break semantics into `selectIsBreakTimeUp`; both effects keep their timer
  stream, and `/focus` now uses the same selector. I also took the single-snapshot cleanup so
  every response field comes from one root-state emission.
- Documented that `isSessionDone` mirrors the current Focus Mode screen and can remain true after
  the overlay is dismissed.
- Matched the other API examples with a bare `/focus` curl.

Validation: selector specs 47/47, effects specs 120/120, handler specs 80/80; `checkFile` passed
for all four changed TypeScript files; docs formatting, link checks, and diff checks passed. I also
rechecked the consecutive-break alarm reset and unstarted zero-duration break paths.

johannesjo:
Thank you very much! <3

aaronedev:
yw🫶

PR #9545, opened 2026-08-11
fix(schedule): open schedule view scrolled to current time #9417

### LINKED ISSUE #9417: 🚨  schedule view start time defaults to 12am/midnight or 2 hours after "work start time", should open to current system time regardless
### Version Used

iOS 18.16.0I, Windows 18.15.1W

### Current Behavior

If the user sets General -> Time and Tracking -> Schedule -> "Limit unscheduled task flow to specific work times", it defaults the schedule view to 2 hours after that start time (ex. 8am start time defaults the top of the schedule view to 10am, cutting off the 8am specified start time). This is incorrect.

When no "work start time" is set, the schedule view defaults to 12am/midnight. OK, but most people probably don't start their day at midnight. See mobile screenshot.

Regardless of the "work start time", this view would be more helpful to the user if it opened to the current system time, so they don't have to immediately scroll to the current time. Especially on mobile where screen space is limited, opening to the "work start time" may not be helpful towards to end of the workday. For example, if the user opens the schedule view at 4pm towards the end of their 8-5 workday, the view opens to 8am, but the mobile screen can only fit 7 hours. Their initial view cuts off at 2pm, which is not helpful when there are tasks scheduled past that hour.

Image

### Expected Behavior

Schedule default view starts with current system time

### Steps to Reproduce

- open schedule view without setting "Limit unscheduled task flow..." and observe 12am default view
- open schedule view after setting "Limit unscheduled task flow..." start time and observe result

### Console Output

```shell

```

### Documentation

costajohnt:
## Problem

Fixes #9417

Opening the schedule view lands at midnight (or at the work start time when that setting is enabled) instead of the current time. When "work start/end" isn't configured, the scroll-on-open silently no-ops because the `#work-start` anchor element is never rendered, so the view sits at 12am and the user has to scroll down to now every time.

## Solution

The scroll-on-open effect in `schedule.component.ts` now prefers the current time indicator whenever today is in the visible range, and only falls back to the work start anchor otherwise (for example after navigating to a week that doesn't include today).

The target choice derives from the component's own `currentTimeRow()` signal rather than a DOM query, which keeps the decision consistent with what `schedule-week` actually renders (the `#current-time` element is guarded by exactly `currentTimeRow() !== null`). The signal read stays inside the `setTimeout` so it remains untracked; otherwise the 2-minute refresh tick would re-run the effect and yank the scroll position while reading the schedule (noted in a comment). The lookup is scoped to the component's own host element, so the duplicate `#current-time` rendered by the right panel's embedded schedule is excluded by construction rather than by DOM order.

The scroll itself is `scrollIntoView({ block: 'start', inline: 'start' })`, and all of the framing lives in CSS as `scroll-padding` on `.scroll-wrapper`:

- **Transform immunity.** The `warpRoute` enter animation starts the view at `scale(1.2)` and the scroll runs on a `setTimeout(0)`, so anything measured from `getBoundingClientRect()` while that is in flight overshoots by the animation's current scale. `scrollIntoView` resolves the target in layout coordinates and is immune to this on its own, so no manual measurement is needed.
- **Lead above the target.** `scroll-padding-top` leaves a slice of the preceding schedule visible above the target, as a fraction of the scroll viewport rather than a fixed number of minutes. A constant reads very differently across breakpoints (30 minutes is ~5% of a desktop viewport but ~10% of a phone one); 12% is roughly an hour on desktop and stays proportional elsewhere. The sticky week header's height folds into the same value.
- **Time column clearance.** `scroll-padding-left` keeps the target clear of the sticky time column, replacing a DOM probe for its width.

Two specs run against the rendered `schedule-week`: one asserts the anchor lands below the sticky header with the lead above it, measured from the real DOM, and one asserts the landing position is unchanged under an ancestor `scale(1.2)`.

### Behavior notes

Within one viewport of the end of the day the browser clamps the scroll and the lead can't be honored — inherent to anchoring on "now", and accepted deliberately rather than papered over with a larger constant. Padding the grid to avoid it would make the scrollbar misrepresent how much day is left in every other interaction with the view. When today is visible but it's before work start, the view still anchors on now (fewer branches, and work start is in the viewport either way).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/e2e/store-screenshots/helpers.ts
+++ b/e2e/store-screenshots/helpers.ts
@@ -51,9 +51,11 @@ export const openTaskDetailPanel = async (page: Page, taskId: string): Promise<v
 
 /**
  * Nudge the schedule's scroll container up by `pxUp` pixels so the
- * captured frame shows a bit of context before work-start (default
- * scroll lands flush at work-start, which crops anything earlier in the
- * day). Pass through to the schedule scroll-wrapper rather than the
+ * captured frame shows a bit of context before the current time. The
+ * default scroll now already leaves a lead above its target (see
+ * scroll-padding-top on .scroll-wrapper), so this over-corrects; shrink
+ * `pxUp` when the screenshots are next regenerated. Pass through to the
+ * schedule scroll-wrapper rather than the
  * page scroll so we don't disturb non-schedule layouts.
  */
 export const scrollScheduleUp = async (page: Page, pxUp = 80): Promise<void> => {
--- a/src/app/features/schedule/schedule-constants.scss
+++ b/src/app/features/schedule/schedule-constants.scss
@@ -16,6 +16,17 @@ $schedule-day-column-width-desktop: 180px;
 $schedule-day-column-width-tablet: 150px;
 $schedule-day-column-width-mobile: 120px;
 
+// Sticky week header. Used both for the header itself and for the
+// scroll-padding that keeps the scroll target from landing under it.
+$schedule-week-header-height: 32px;
+$schedule-week-header-height-mobile: 22px;
+
+// How much of the schedule stays visible above the scroll target, as a
+// fraction of the scroll viewport. A fixed number of minutes reads very
+// differently on a 950px desktop viewport and a 400px phone one; a fraction
+// is stable across breakpoints and zoom levels. ~1h on a desktop viewport.
+$schedule-scroll-lead: 12%;
+
 // Scrollbar dimensions
 $schedule-scrollbar-height: 8px;
 $schedule-scrollbar-width: 4px;
--- a/src/app/features/schedule/schedule-day-panel/schedule-day-panel.component.ts
+++ b/src/app/features/schedule/schedule-day-panel/schedule-day-panel.component.ts
@@ -457,6 +457,12 @@ export class ScheduleDayPanelComponent implements AfterViewInit, OnDestroy {
     return null;
   }
 
+  // TODO replace with the schedule view's approach (scrollIntoView plus
+  // scroll-padding on the scroll container, see ScheduleComponent) so there is
+  // one implementation of "scroll to now" instead of two that can drift. This
+  // one is rect-based like the old schedule code was, but its enter animation
+  // (slideInFromRightAni) is translateX only, so it does not hit the transform
+  // overshoot that fix had.
   private _scrollToCurrentTime(): void {
     if (!this.scheduleWeekRef?.nativeElement) {
       Log.warn('[ScheduleDayPanel] No scheduleWeekRef available');
--- a/src/app/features/schedule/schedule-week/schedule-week.component.scss
+++ b/src/app/features/schedule/schedule-week/schedule-week.component.scss
@@ -228,10 +228,12 @@
   grid-template-columns: var(--schedule-time-width) repeat(var(--nr-of-days), 1fr);
   width: 100%;
   box-sizing: border-box;
-  min-height: 32px;
+  // Kept in sync with the scroll-padding on .scroll-wrapper, which offsets the
+  // initial scroll target so it doesn't land under this sticky header.
+  min-height: $schedule-week-header-height;
 
   @include mq(xs, max) {
-    min-height: 22px;
+    min-height: $schedule-week-header-height-mobile;
   }
 }
 
@@ -398,7 +400,6 @@
 
 .work-end,
 .work-start {
-  scroll-margin-top: 80px;
   border-top: 2px dashed var(--separator-color);
   min-width: 100px;
   grid-column: 2 / -1;
--- a/src/app/features/schedule/schedule/schedule.component.scss
+++ b/src/app/features/schedule/schedule/schedule.component.scss
@@ -17,6 +17,19 @@
   overflow-x: hidden;
   box-sizing: border-box;
 
+  // Framing for the initial scroll to the current time (or work start):
+  // keep the target clear of the sticky week header and the sticky time
+  // column, and leave a slice of the preceding schedule visible above it so
+  // it reads as "now, in context" rather than the day being cut off.
+  scroll-padding-top: calc(#{$schedule-week-header-height} + #{$schedule-scroll-lead});
+  scroll-padding-left: calc(#{$schedule-time-column-width} + 12px);
+
+  @include mq(xs, max) {
+    scroll-padding-top: calc(
+      #{$schedule-week-header-height-mobile} + #{$schedule-scroll-lead}
+    );
+  }
+
   // Enable horizontal scroll when viewport is too small for 7 days
   &[data-horizontal-scroll] {
     overflow-x: scroll; // Show horizontal scrollbar when needed
--- a/src/app/features/schedule/schedule/schedule.component.spec.ts
+++ b/src/app/features/schedule/schedule/schedule.component.spec.ts
@@ -1180,4 +1180,62 @@ describe('ScheduleComponent', () => {
       expect(mockLayoutService.selectedTimeView()).toBe('day');
     });
   });
+
+  describe('scroll target measurement', () => {
+    const TARGET_TOP_PX = 250;
+    let host: HTMLElement;
+    let wrapper: HTMLElement;
+
+    const buildScrollableDom = (scale: string): void => {
+      host = document.createElement('div');
+      // The route enter animation (warpRoute) starts the view at scale(1.2),
+      // and the scroll runs while that is still in flight.
+      host.style.cssText = `position:relative;transform:${scale};transform-origin:top left;`;
+
+      wrapper = document.createElement('div');
+      wrapper.className = 'scroll-wrapper';
+      wrapper.style.cssText = 'position:relative;overflow:auto;height:100px;width:200px;';
+
+      const spacer = document.createElement('div');
+      spacer.style.cssText = 'height:600px;width:400px;';
+
+      const target = document.createElement('div');
+      target.id = 'test-scroll-target';
+      target.style.cssText = `position:absolute;top:${TARGET_TOP_PX}px;left:0;height:2px;width:10px;`;
+
+      wrapper.append(spacer, target);
+      host.append(wrapper);
+      document.body.append(host);
+    };
+
+    afterEach(() => host?.remove());
+
+    it('scrolls to the layout position even while an ancestor is scaled', () => {
+      buildScrollableDom('scale(1.2)');
+      const scrollToSpy = spyOn(wrapper, 'scrollTo');
+
+      component['_scrollIntoViewWithTimeColumnOffset']('test-scroll-target');
+
+      expect(scrollToSpy).toHaveBeenCalled();
+      const options = scrollToSpy.calls.mostRecent().args[0] as ScrollToOptions;
+      // Rect-based math would return TARGET_TOP_PX * 1.2 here, scrolling the
+      // view roughly two hours past "now".
+      expect(options.top).toBe(TARGET_TOP_PX);
+    });
+
+    it('gives the same target with and without the scale', () => {
+      buildScrollableDom('scale(1.2)');
+      const scaledSpy = spyOn(wrapper, 'scrollTo');
+      component['_scrollIntoViewWithTimeColumnOffset']('test-scroll-target');
+      const scaledTop = (scaledSpy.calls.mostRecent().args[0] as ScrollToOptions).top;
+      host.remove();
+
+      buildScrollableDom('none');
+      const plainSpy = spyOn(wrapper, 'scrollTo');
+      component['_scrollIntoViewWithTimeColumnOffset']('test-scroll-target');
+      const plainTop = (plainSpy.calls.mostRecent().args[0] as ScrollToOptions).top;
+
+      expect(scaledTop).toBe(plainTop);
+    });
+  });
 });
--- a/src/app/features/schedule/schedule/schedule.component.spec.ts
+++ b/src/app/features/schedule/schedule/schedule.component.spec.ts
@@ -1257,6 +1257,51 @@ describe('ScheduleComponent', () => {
       expect(options.top).toBe(TARGET_TOP_PX - leadPx);
     });
 
+    it('keeps the full lead visible below the sticky week header', () => {
+      const HEADER_HEIGHT_PX = 33;
+
+      host = document.createElement('div');
+      host.style.cssText = 'position:relative;';
+
+      wrapper = document.createElement('div');
+      wrapper.className = 'scroll-wrapper';
+      wrapper.style.cssText = 'position:relative;overflow:auto;height:100px;width:200px;';
+
+      const week = document.createElement('schedule-week');
+      week.style.cssText = 'display:block;';
+
+      // Sticky, so it stays over the top of the viewport once scrolled and
+      // covers the first HEADER_HEIGHT_PX of whatever sits below it.
+      const header = document.createElement('div');
+      header.className = 'week-header';
+      header.style.cssText = `position:sticky;top:0;height:${HEADER_HEIGHT_PX}px;width:400px;`;
+
+      const grid = document.createElement('div');
+      grid.className = 'grid-container';
+      grid.style.cssText = 'position:relative;height:2880px;width:400px;';
+
+      const target = document.createElement('div');
+      target.id = 'test-scroll-target';
+      target.style.cssText = `position:absolute;top:${TARGET_TOP_PX}px;left:0;height:2px;width:10px;`;
+
+      grid.append(target);
+      week.append(header, grid);
+      wrapper.append(week);
+      host.append(wrapper);
+      document.body.append(host);
+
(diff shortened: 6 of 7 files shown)


healeyj:
thanks for the effort. i tested with the preview link. it appears that the schedule view now anchors to current_time +2 hours and snapping to nearest 30 min, regardless of whether work day start and work day end are set. that's on the right track, but what is causing that 2 hour offset?

costajohnt:
Good catch, and thanks for testing it. Reproduced, and it turned out to be a pre-existing bug in the scroll helper that this PR just made visible.

It isn't a fixed two hours, it's 20% of the time of day. Measured against real layout:

| now | lands at | error |
|---|---|---|
| 06:00 | 07:12 | +1h12m |
| 09:00 | 10:48 | +1h48m |
| 10:00 | 12:00 | +2h00m |
| 13:00 | 15:36 | +2h36m |
| 16:00 | 19:12 | +3h12m |

You tested late morning, which is where it works out to almost exactly +2h.

The cause is a coordinate space mix-up in `_scrollIntoViewWithTimeColumnOffset`. The route enter animation (`warpRoute`) starts the view at `transform: scale(1.2)` and the scroll runs in a `setTimeout(0)`, so it fires while that is still animating. `getBoundingClientRect()` reports visual coordinates and includes the transform, but `scrollTop`/`scrollTo()` take layout coordinates. The grid is 12 rows/hour at 9px, so 108px per hour measured from midnight, and scaling that distance by 1.2 overshoots by 20% of however far into the day you are.

That also explains why work start and end make no difference: the anchor element is the current-time marker either way. The same overshoot exists on master for the work-start anchor, it just rarely showed because that element often isn't rendered at all, which is the original issue.

On the snapping: I don't think it's real. Landing points across a range of times come out continuous, not quantized: 9:00 lands at 10:48, 9:30 at 11:24, 10:00 at 12:00, 10:30 at 12:36. At 10:00 it happens to land exactly on the hour, which reads as snapped. The run-to-run variation you may have seen is the `setTimeout(0)` sampling the animation at a slightly different scale each time (at 10:00: scale 1.2 gives +2h00m, 1.1 gives +1h00m, 1.05 gives +0h30m).

Fixed by measuring with `offsetTop`/`offsetLeft` instead, which no ancestor transform can skew, plus `offsetWidth` for the time column which had the same problem. Two regression tests cover it; against the old code they fail with exactly the 1.2x overshoot. The anchor-choice logic is unchanged.

One design question while you're looking, since you have opinions on how this should feel: should "now" sit flush at the very top of the viewport, or would a bit of lead-in above it read better? And when today is visible but it's still before work start, would you rather land on work start?

healeyj:
I propose 30 minutes of lead. Can you update the deployment preview with those fixes so I can test again?

healeyj:
tested latest fix, its doing around 15 minutes of lead instead of 30.

costajohnt:
Thanks for re-testing, that's a real gap and I can account for the missing half.

The scroll math does subtract a full 30 minutes, but the week header is `position: sticky; top: 0` inside the same scroll container, so it sits opaque over the top ~33px of the viewport once you're scrolled. The grid is 12 rows/hour at 9px, so 1.8px per minute, which makes 30 minutes of lead 54px. Take the 33px the header covers off that and about 21px, roughly 12 minutes, is actually visible. It shifts with the schedule zoom setting since the header is a fixed pixel height while the lead scales, which is probably why it read as 15 for you.

Pushed a fix that pulls the sticky header height into the scroll target the same way the time column width is already handled horizontally, plus a regression test. Preview should update shortly.

healeyj:
Tested again, now the top of the schedule view directly under the day of week header opens to about 1.75 hours before the current time marker. This is too much lead, but seeing the extra space, an hour of lead is probably more appropriate for the final change, 30 minutes would be too cramped. If you can triangulate to that, three times a charm... hope I'm not too demanding here, I'm impressed by your speed.

costajohnt:
Not too demanding at all, and thanks for testing it three times.

Before I change the number again I want to make sure we're measuring the same thing, because I'm not able to reproduce the 1.75 hours. On the current preview build at default settings I measured the gap between the bottom of the pinned day header and the current time marker at about 70px, which works out to roughly 39 minutes with the lead currently set to 30. So something about your setup is stretching it well past mine, and if I just bump the constant to 60 you might end up somewhere near 3 hours instead of the 1 you're after.

Two things would let me triangulate properly:

1. What schedule zoom / row height are you on? The lead is derived from the rendered grid height, so a different zoom changes how many minutes a given pixel gap represents.
2. Roughly what time of day were you testing? If the marker sits late in the day the view can hit the bottom of its scroll range and stop short, which would leave more space above the marker than intended.

A screenshot of the initial view would answer both at once if that's easier.

If it turns out to be a plain zoom difference then it's a one line change and I'll push it. If the scroll is clamping at the end of the day that's a real bug worth fixing properly rather than papering over with a bigger constant.

johannesjo:
Thanks for the thorough work here, and for chasing the two follow-up bugs the tester surfaced instead of stopping at the first symptom. I checked the branch out and verified the core claims rather than taking them at face value:

- 72/72 specs pass, `checkFile` clean on both files.
- **The regression tests are genuine.** I reverted `_scrollIntoViewWithTimeColumnOffset` to the rect-based math and exactly the 5 new measurement specs fail, with the claimed overshoot (`Expected 300 to be 250`). Good.
- **The root cause holds.** `.grid-container` is `repeat(24 * 12, 9px)` with `place-content: start` → 2592px = exactly 24h, 108px/h. 08:00 sits at 864px, ×1.2 ≈ +1.6h, which matches the "2 hours after work start" in the issue title. So the reported symptom and the coordinate-space bug are the same bug — nice find.
- `_leadPx`'s "the grid is always 24h" assumption is safe (`--schedule-total-rows: calc(24 * var(--schedule-fr))`; mobile only swaps row height, so the lead stays 30 minutes).
- The duplicate `#current-time` id is not a regression: `right-panel.component.html` renders `<ng-content>` before `.side`, so `getElementById` still hits the main view first. It is now load-bearing on DOM order though — see (5).

## The open thread with @healeyj is the bottom clamp, not a zoom difference

You don't need to wait for the zoom/time answers, the arithmetic settles it. Content height is 2592px + ~33px header = 2625px, so `maxScrollTop = 2625 − wrapperHeight`, while the target is `hours × 108 − 54`. Clamping therefore starts at `hours > (2679 − wrapperHeight) / 108` — about **16:00 on a ~950px-tall wrapper**. At 17:00 the target is 1815 against a max of 1675, so the browser stops 140px short and the marker ends up ~194px ≈ **1.8h** below the header. That's the 1.75h that was reported, and both test rounds were in the afternoon/evening.

A clean discriminator if you want to confirm before changing anything: ask for one test around **10:00**. If the gap is exactly 30 minutes there and only grows later in the day, it's the clamp and bumping the constant will do nothing for the case that prompted it.

Worth being explicit that this is inherent to anchoring on "now": within one viewport of the end of the day, the lead can't be honored. `master` never hit it because work-start is early. Either accept it, or add roughly a viewport of scroll padding below the grid so "now" can always sit near the top. I'd rather that be a deliberate decision than an inflated constant.

## Findings

**1. A fixed lead in *minutes* reads inconsistently across viewports.** 30 min is 54px on desktop and 42px on mobile — about 5% of a 950px desktop viewport but ~10% of a 400px mobile one. That's most likely why 30 minutes "felt cramped" on desktop, and it means any constant you pick will be wrong on one of the two. A fraction of the wrapper's `clientHeight` (the `scroll-padding-top: 15%` idiom) is stable across breakpoints and zoom and would end the back-and-forth.

**2. There is a materially leaner shape here, and I verified it works.** `schedule-week.component.scss:401` already has `.work-start { scroll-margin-top: 80px }` — vestigial, since `scrollTo` ignores scroll-margin. I probed the alternative in Chrome to be sure rather than guessing:

| probe | result |
| --- | --- |
| `scrollTo` honors `scroll-margin-top` | no (500 → 500) |
| `scrollIntoView` honors `scroll-margin-top` | yes (500 → 420) |
| `scrollIntoView` under an ancestor `scale(1.2)` | lands at the **layout** position (500), not 600 |

So `element.scrollIntoView({ block: 'start', inline: 'start', behavior: 'instant' })` plus `scroll-margin-top` / `scroll-margin-left` on the two anchors is transform-immune on its own and would delete `_layoutOffset`, `_leadPx`, and both `querySelector` probes — roughly 50 lines — while moving the geometry into CSS where `--schedule-row-height` and `--schedule-time-width` already live (`calc(var(--schedule-row-height) * 6)` is exactly 30 minutes; `calc(var(--schedule-time-width) + 12px)` replaces the time-column measurement and handles the `display:none` `.filler` case for free). Two caveats: `scrollIntoView` also scrolls outer scroll ancestors, and the sticky header height would need to fold into the top margin. Given how much machinery the current approach adds to a view component, I'd like this evaluated before we commit to the manual math.

**3. The measurement specs hand-build their DOM with the same selectors the implementation uses.** Renaming `.grid-container`, `.week-header`, or `.scroll-wrapper` in `schedule-week.component.html` breaks production while all five specs stay green, and the synthetic geometry (2px/min, 33px header) doesn't match the real 1.8px/min. One spec measuring against a really-rendered `schedule-week` would close that gap; the rest can stay synthetic.

**4. `- scrollContainer.clientTop / clientLeft` is only correct in one of two configurations.** I confirmed in Chrome that `offsetTop` is relative to the offsetParent's *padding* edge (child of a `border-top: 10px` positioned parent reports `offsetTop === 0` while `clientTop === 10`). So the border subtraction is right only when the container is **not** the offsetParent. Production is fine today — `.scroll-wrapper` is static and unbordered — and the specs use a `position: relative` wrapper with no border, so neither path catches it. Zero impact now, but a trap if the wrapper ever gains `position: relative` plus a border. Worth a comment at minimum.

**5. Scope the element lookup to the component's host.** `document.getElementById` works only because of right-panel DOM ordering, and the panel's `schedule-week` always renders its own `#current-time` (its `currentTimeRow` is never null). Querying from the component's own element instead makes that intentional rather than incidental, and is closer to how we handle DOM access elsewhere.

**6. Second copy of the same logic left behind.** `schedule-day-panel._scrollToCurrentTime()` is an independent rect-based implementation with a hardcoded `SCROLL_TOP_OFFSET_PX` and `Log.log`/`Log.warn` on every panel open. I checked `slideInFromRightAni` — it's `translateX` only, so it doesn't hit the scale bug and this isn't urgent. But it's a second implementation of "scroll to now" that will drift from this one. Not in scope to fix here; a TODO pointing at the new helper would do.

**7. Stale e2e comment.** `e2e/store-screenshots/helpers.ts:52-58` documents "default scroll lands flush at work-start" and nudges 80px up. With the clock pinned to 09:30 and the seed's `workStart: "9:00"`, the default now already lands at 09:00, so the nudge over-corrects to ~08:16. The pipeline is deterministic so nothing breaks, but the comment is wrong and the nudge should shrink when screenshots are next regenerated.

## On scope

The two things that are unambiguously bugs — the coordinate overshoot, and work-start being absent so the scroll silently no-ops at midnight — stand on their own and I want them regardless of the rest. The behavior change (always anchor on "now", ignoring work start when today is visible) is the feature half, and #9417 has no upvotes and a single reporter, albeit a genuinely engaged one who has now tested three times. It's a calm default with no new setting, which fits how I want this app to work, so I'm inclined to take it.

To answer your own open question: when today is visible but it's before work start, I'd still land on **now**. The user sees work start in the viewport either way, and a rule with fewer branches is easier to reason about later.

**Minor:** the effect's `setTimeout` isn't cleared on destroy (pre-existing); `_layoutOffset` is a generic DOM helper that would sit better in `src/app/util/`; switching week↔day view doesn't re-anchor because `isMonthView()` doesn't flip (also pre-existing).

My preference for landing this: settle the clamp question with @healeyj using the 10:00 test so we're not tuning a constant against the wrong cause, take a look at (2), and address (3), (4) and (5). The rest can be follow-ups.

healeyj:
~10am from preview build on iPhone 15 Safari
IMG_1636

~10am from preview build on Firefox 1440p maximized (bottom windows toolbar and top firebox toolbar cropped out)
Screenshot 2026-08-14 100918

Both using default zoom, didn't know that was a thing until you pointed it out.

costajohnt:
Thanks both, this settles it.

@healeyj's 10am shots confirm your arithmetic: measuring the desktop one, the grid is 107px/hour and the marker sits about 54px below the pinned header, so the lead there is the full 30 minutes. The 1.75h reading was the afternoon clamp, not a zoom difference and not a constant that needed raising. Good call on the discriminator, that saved a round of tuning the wrong number.

## What changed

Took (2). `scrollIntoView({ block: 'start', inline: 'start' })` plus scroll-padding is transform-immune on its own, which is the entire reason the offset-parent walk existed, so the whole measurement layer is gone: `_layoutOffset`, `_leadPx`, both DOM probes, and the `clientTop` subtraction from (4) along with them. The geometry now sits in CSS next to the row-height and column-width variables it derives from.

I put the padding on the container rather than scroll-margin on the anchors, which also gives (1) for free: `scroll-padding-top: calc(<header height> + 12%)` on `.scroll-wrapper`, and `scroll-padding-left: calc(var(--schedule-time-width) + 12px)` for the horizontal side. A percentage resolves against the scrollport, so the lead scales with the viewport instead of being 5% of a desktop one and 10% of a phone one. 12% works out to roughly an hour on desktop, which is what @healeyj landed on, and stays proportional on mobile. The header height moved into a shared SCSS variable so the header and the padding can't drift apart.

(5) done, the lookup is scoped to the component's host now, so the right panel's second `#current-time` is excluded by construction rather than by DOM order.

(3) done, the five synthetic specs are replaced by two that run against the real rendered `schedule-week`: one asserts the anchor lands below the sticky header with the lead above it, measured from the actual DOM, and one asserts the landing position is unchanged under an ancestor `scale(1.2)`. Renaming `.week-header` or `.scroll-wrapper` now fails the specs. The transform spec still catches the original bug with a wide margin, the old overshoot was hundreds of pixels and the remaining difference is a sub-pixel rounding of 1px.

(6) TODO added on `schedule-day-panel._scrollToCurrentTime()` pointing at the new approach. (7) e2e comment corrected to describe what the default scroll actually does now, with a note that the nudge should shrink when the screenshots are next regenerated. I left the 80px alone so the pipeline stays byte-identical until someone regenerates.

## On the clamp

I'd accept it rather than pad the grid. Within one viewport of midnight the lead can't be honored, but what the user sees there is the end of their day sitting in view, which is not a bad frame. A viewport of empty space below the grid would make the scrollbar lie about how much day is left and would be visible in every other interaction with the view, to fix a case where the current behavior is merely not ideal. If it turns out to bother people in practice, it's a self-contained change to make later.

Also took your answer on the before-work-start case, it stays anchored on now, fewer branches.

Not addressed, agreed as follow-ups: the uncleared `setTimeout` on destroy, `_layoutOffset` moving to `src/app/util/` (moot, it's deleted), and the week↔day re-anchor.

72/72 specs pass, `checkFile` clean. @healeyj, preview should rebuild shortly, and this one should read as about an hour of lead in the morning. Late afternoon will still show more, that's the clamp, and it's deliberate now rather than a number that needs tuning.

healeyj:
thanks all. tested again around 3pm then 11am, confirmed the top of schedule view now opens to about ~1hr before system time. default zoom used. desktop screenshots provided, also confirmed same behavior on mobile but no screenshots. cheers :)

3pm
Screenshot 2026-08-15 145522

11am
Screenshot 2026-08-16 112123

costajohnt:
@healeyj thanks for testing this five times, including the 10:00 run that settled which cause we were actually chasing. That one measurement is what stopped me tuning the lead constant against the clamp.

For the record, since it took a few rounds to land: the reading now tracks the viewport rather than a fixed number of minutes, which is why it comes out around an hour on your desktop and stays proportional on the phone.

@johannesjo I think this is ready for another look. Everything from your review is in:

- **(2)** `scrollIntoView` plus `scroll-padding`, which deleted `_layoutOffset`, `_leadPx`, both DOM probes, and with them the `clientTop` trap from **(4)**
- **(1)** the lead is a fraction of the scroll viewport rather than a fixed 30 minutes
- **(5)** the anchor lookup is scoped to the component's host
- **(3)** the synthetic measurement specs are replaced by two that run against the rendered `schedule-week`
- **(6)** and **(7)** left as the TODO and the corrected comment you agreed were follow-ups

One thing worth flagging before you merge: this PR's CI last ran on the 14th, which is before #9304 added the `lint:css-vars` guard. Since this diff adds custom properties and builds `scroll-padding` on them, old green would not have carried over, so I rebased locally onto current master and re-ran everything: the guard passes across all 1807 files, 69/69 specs, `checkFile` clean.

I have not pushed that rebase. The branch still merges cleanly and I would rather not reset your review context on a green PR for no reason, but say the word and I will push it so CI re-runs against current master.

johannesjo:
Thank you very much! <3

PR #9550, opened 2026-08-11
Feat/weekly report export

FarmerLX:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for putting this together — there's clearly real care in here: tests, wiki update, translations done properly through `T`.

I'm going to decline the weekly-report export itself, though.

**Scope.** The report structure — *Completed this week* / *Work analysis summary* / *Next week's plan* / *Coordination and help needed* — is a specific status-report-to-your-manager format. Super Productivity is deliberately a personal deep-work tool rather than a team-reporting one, and a fixed template like this imposes one particular workflow on everyone with no way to adapt it. There's also no issue behind this and I couldn't find anyone asking for it, so it would be permanent core surface (plus a permanent output format I'd have to keep stable) added for a single workflow.

**This would make a great plugin, though.** The plugin API already exposes everything it needs — `getArchivedTasks()` and `getAllProjects()` in `packages/plugin-api/src/types.ts`. As a plugin you'd keep exactly the section headings and grouping you want, iterate without waiting on me, and anyone with a different report format could adapt it. I'd be happy to see it in the community plugin list. `docs/plugin-development.md` and `packages/plugin-dev/QUICK_START.md` are the starting points.

A few things I noticed while reading, in case they're useful for a plugin version:

- Inbox tasks and tasks without a project are filtered out entirely. The Inbox is where most tasks live by default, so the report would silently omit most of a typical week — and the UI gives no hint that anything was left out.
- Task notes are markdown in SP. `formatMarkdownNoteLines` flattens them onto one line joined with ` / ` and then backslash-escapes markdown characters, so a note containing a list or bold formatting comes out as literal `\*text\*`. Same for a task titled `Fix #123` → `Fix \#123`.
- The `isWeeklyRange` exactly-7-days check means the button silently disappears for any week straddling a month boundary when exporting from History, because `getDateRangeForWeek(year, week, month)` clamps the range to the month.

**One part I'd like to take separately:** the `currentWeek$` rewrite plus the `sortWorklogDays` change are a genuine bug fix independent of the report. The old code called `getWeekNumber(now)` without the user's `firstDayOfWeek` and only searched the current month, so the current-week widget broke for non-Monday week starts and across month boundaries; and `sortDays` compared day-of-month keys as strings (`"10" < "9"`). If you open that as its own small PR I'll merge it.

The `map-archive-to-worklog.ts` change (letting `doneOn` win over the parent fallback) may well be correct too, but it moves done subtasks without tracked time to a different day in the Worklog and CSV export for every existing user — so it needs its own PR and a note on what changes for people.

The `scripts/build-deb.sh`, `scripts/install-deb.sh`, `scripts/sync-with-upstream.sh` and `README.zh-CN.md` changes look like fork-maintenance bits that got swept in; those would need to be separate regardless.

Closing this one — but thanks again, and please do ping me if you build the plugin.

Issue #8211, opened 2026-06-09
💡 ability to visualise task spread over several days


ihubgit:
### Problem Statement

Running Version 18.9.1 on Apple Silicon Tahoe.
The current version allows to set a deadline that's beyond the creation date of the task, but this isn't reflected visually in the app (f.ex.planner and schedule tabs), which is quite confusing, because it suggests the task ought to be finished the same day.

### Possible Solution

Most other calendar apps out there allow for a task to be spread over several days and simply make this obvious by spanning the task entry over those several days in the form of a long bar, not restricted to the day of the creation date.
This would be particularly useful to have a visual feedback in the planner and schedule app as how long the task is supposed to last.

johannesjo:
Thanks for opening this up! To be clear up front: Seems like a feature which is probably complex, difficult to get right and more of a niche use case. If anyone has an excellent plan on how to make this work without adding too much technical debt and complexity for the user, I'm open to it.

ihubgit:
hi it's actually very simple and a basic feature of about every calendar app out there.
Here's a screenshot from the Calendar app in Tahoe : simply a horizontal bar stretching over several days, that's all !

<img width="1521" height="733" alt="Image" src="https://github.com/user-attachments/assets/ab6f300e-b2d4-4e49-a91a-add35d2b11ee" />

Issue #7069, opened 2026-04-01
💡 Don't consider a task to be overdue before the deadline


beerkumquatpome:

symonbaikov:
Took this one to work

johannesjo:
I think we are mixing up two different things here. Scheduled and deadline are handled as two different entities. I don't think it is good to mix these concepts up unless we have a clear instantly understandable concept for this that is cleaner and more intuitive than the clear seperation of both.

johannesjo:
Just to be clear: We are talking about the overdue list on the today view right?

Currently we use "Overdue" to display tasks that were scheduled ("planned for") earlier days. It does not include tasks past their deadline (there is  a different mechanism for that). That is probably the source of confusion. It is meant to be a quick way to deal with those tasks. Imho a missed deadline is more serious than just a simple re-schedule to gloss over (and should ideally never happen), so a Dialog and more importantly reminders before the deadlines is there makes more sense to me. 

Possibly there is a better label for this than "Overdue"?

@Jagdfalke, @novikov1337danil & @zenoprax what would be your take on this?

hiJanes:
From what I observe, both features work in the same way. There doesn’t seem to be anything particularly special about a deadline.

Both can be completed before the set date/time. 
If not completed, 
both turn red,
 both can be rescheduled.
I don’t see any real difference other than a new icon and a heavier term.

What I think it should be like:

Schedule: The task should appear in the “Today” section on the scheduled date (which already exists).

Deadline: It should have a time window (e.g., 03/04/26 to 12/04/26), during which it appears in the “Today” section and must be completed before 12/04/26. The start date doesn’t necessarily have to be the creation date—it should be selectable when creating the task.

Additionally, the task should visually change (color or icon) as the deadline approaches. It should not move to the “Overdue” section (below “Today”) until a new deadline time window is set. There could also be a limit on rescheduling—for example, allowing only two reschedules, with a confirmation prompt like: “You can only reschedule this task twice. Are you sure?”


<img width="1472" height="598" alt="Image" src="https://github.com/user-attachments/assets/63557ace-7545-4f7e-a4aa-b698a86600f3" />

zenoprax:
> Tasks planned for a previous day were visually marked as overdue even when their deadline had not passed yet.

I have only briefly experimented with Deadlines but my impression was that the difference between the Today and Planner view is more important now since setting a deadline (ignoring due/scheduled date) will cause it to show up in the latter but not the former.

[...]

Issue #6078, opened 2026-01-21
💡 Adding start date and due (end) date not just scheduled date


Marin-Kitagawa:
### Problem Statement

I am migrating from Tasks.org. It has 2 types of dates for a task:
- Start date
- End date

But Super-Productivity only has scheduled date.

### Possible Solution

Implement start date and end date for better clarity of task periods

woodpexer:
**community feedback:**

> I am migrating from Tasks.org. It has 2 types of dates for a task: 
>     * Start date
>     * End date
> But Super-Productivity only has scheduled date.

if it is end date in general a few hours after the start date, super-productivity could be used like this:
Start date == the current "schedule task" functionality
End date == the current "time" / duration estimation

if the end date is a period which spans many days after the start date, then i would vote to combine this with a gantt feature #4790 which shows not only task periods, but also dependencies, orders, duration of single tasks, and overall deadlines / timelines of the whole goal or project.

billy-thegreat:
Super Productivity allows users to create a task with an estimated time and a scheduled date and optional time.
Tasks.org uses the iCal/CalDAV standard to sync tasks. The `DTSTART` and `DUE` values declare a window (of days) that a task can be completed in. This could be implemented in Super Productivity as two more values that a task could have. The app could issue a reminder when the `DUE` date has been passed, and prompt the user to add it to today. Before `DTSTART` is reached, the task could be hidden from the project view. The scheduled date and time would be anywhere between the start and end dates.
I hope I have accurately depicted the request's intent.

zenoprax:
> I am migrating from Tasks.org. It has 2 types of dates for a task:
> 
>     * Start date
> 
>     * End date
> 
> 
> But Super-Productivity only has scheduled date.

See my recently closed issue #7125.

"Start Date" is the "Scheduled Date". There is still some confusion about what it means for a task to be "overdue" given the addition of the new "Deadline" property as described in #7069.

Super Productivity is generally focused on *task* management rather than *project* management. That means that "start" and "end" means different things depending on your perspective.

A task can have:

1. earliest time/date it could be started (for filtering purposes or Gantt charts)
2. planned/scheduled time/date that it will become the "active task"
3. an intended time/date for it to be completed
4. a hard deadline by which time it *must* be completed (and with the appropriate implications for Gantt charts)

[...]

PR #9553, opened 2026-08-12
feat(schedule): show deadline-range ghost bars on intermediate calendar days

### LINKED ISSUE #9484: 💡Multi-day Task Display on Calendar View
### Problem Statement

Currently, tasks with a duration (e.g., spanning from August 7th to August 10th) only appear on the calendar on their start and end dates. This makes it difficult to visualize the active workflow and workload across intermediate days.

It would be extremely helpful to span the task across all days within its date range on the calendar view.

### Possible Solution

Proposed Behavior
- Display a continuous visual bar or entry for a task across every day from its start date to its end date.
- Improve at-a-glance visibility for overlapping tasks and ongoing project timelines.

_No response_

### Additional context

Additional Context
- Example: A task scheduled from 07/08 to 10/08 should be visible on the calendar for August 7, 8, 9, and 10, rather than just appearing as isolated points on the 7th and the 10th.

colaresAmanda:
## Problem

Tasks with both a "Planned for" date and a "Deadline" only ever show up on the
calendar on those two exact days — every day in between is blank, so there's
no way to see that a task is still active/pending toward its deadline while
browsing Schedule's week or month view. This is what #9484 describes: a task
planned 07/08 with a deadline of 10/08 only appears on the 7th and the 10th,
not on the 8th/9th in between.

Closes #9484

## Solution

Adds a faded "ghost" bar on every day strictly after the planned date through
the deadline day (inclusive) in Schedule's week and month views, for any task
that has both dates set. Key points:

- **No new task field.** Reuses the existing `dueDay`/`dueWithTime` (planned)
  and `deadlineDay`/`deadlineWithTime` fields — no schema change, no migration.
- **Purely additive.** Computed as a separate, independent list of synthetic
  `ScheduleEvent`s and concatenated onto the existing flat events array — zero
  changes to the existing day-bucketing/budget/flow logic.
- **Visual treatment:** thin, dashed, low-opacity bar with the full task
  title (no truncation, wraps to fit), clickable to open the task — same
  click behavior as a real scheduled event. Not draggable/resizable (it's a
  projection, not the real scheduled instance).
- Also widened the month-view day cells and removed a pre-existing bug where
  events past the 3rd were silently hidden with no "+N more" indicator on
  desktop — needed so the added ghost bars don't get invisibly clipped.
- This issue is tagged `needs concept and/or ui spec` — happy to adjust the
  visual treatment (color, shape, density) based on feedback; the underlying
  data logic (which days get a ghost) is the part I'd like to lock in first.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki)


johannesjo:
Hi @colaresAmanda, and thank you for this. It's a really nice first PR: the data logic is cleanly isolated in its own module, the tests cover the fiddly `startOfNextDayDiffMs` case, and I appreciate that you found a way to do this with no new field and no schema change. That's the right instinct.

I'm not going to merge it though, and I want to explain why properly, because I think you already spotted the reason yourself.

**The two fields don't describe a range.** In #9484 you wrote:

> I would need to create another field inside the task due to the meaning of the "planned for" and "deadline" buttons.

That was exactly right. "Planned for" is the one day you intend to work on something; "Deadline" is the day it has to be done by. The days in between aren't the task's duration. They're the slack you left yourself, which is to say precisely the days you decided *not* to work on it. Drawing a bar across them tells the user the task occupies time that it doesn't.

Keeping these two concepts separate is a deliberate decision (see #7069), and the deadline spec in #4328 lists start dates as explicitly out of scope, along with timeline/Gantt views and partial task scheduling. So "a task spans a range of days" needs a real start date to stand on, and that's still an open question. #6078 and #4647 are the right place for it, both tagged `community feedback wanted`. If you want to push the idea forward, making the case there would help more than code does.

**One thing that might already solve the underlying problem:** if you give a task a time estimate bigger than a day's remaining capacity, the Schedule already spreads it across the following days on its own, with dashed continuation styling. It's capacity-driven rather than date-driven, but for "show me this task is still in flight" it may be closer to what you want.

**What I would merge:** the Schedule shows nothing at all about deadlines right now (`src/app/features/schedule/` doesn't reference them anywhere), while the Planner already groups deadline tasks per day. That inconsistency is a real gap, and I suspect it's what both you and #8211 are actually reacting to. A deadline marker on the deadline day, in the week and month views, would close it.

There's a good pattern to reuse: the over-budget badge in the week header (`schedule-week.component.html`) is already a small pill with a symbol, a count and a tooltip, and it costs no grid space. Swap the `!` for the `flag` icon we already use for deadlines and it would fit right in.

If you'd like to take that on, I'd be glad to review it. Either way, one note: the month-view changes in here (cell height, the event cap) are a separate topic and overlap with what #9463 is doing to the month grid, so they'd want their own PR regardless.

Thanks again for digging into this properly. The concept is a no, but the instinct that Schedule should know about deadlines is right.

johannesjo:
Closing this one per the above. Thanks again for the work you put into it, @colaresAmanda, and please do ping me if you'd like to pick up the deadline-marker scope in the Schedule view. I'd be glad to review that.

PR #9554, opened 2026-08-12
fix(ui): preserve time input editing state

### LINKED ISSUE #9548: 🚨 Unable to edit "Work Day Start/End" hours in global schedule settings
### Version Used

18.19.0D

### Current Behavior

The input handling for these settings makes it impossible to _meaningfully_ adjust their value and changes result in broken/unpredictable outcomes.

### Expected Behavior

Typing a 00-23 value into the hour field or a 00-59 value into the minute field will set that value correctly. Instead typing "18" as hour results in "08" and adjusting minutes also changes the hour part

### Steps to Reproduce

1. Go to Settings > Time & Tracking > Schedule
2. Attempt to adjust "Work Day End" time
3. Type "18" in hour field (turns into "08")
4. Type "30" into minute field (turns into "03")

### Console Output

```shell

```

### Documentation

Akashpandeye:
## Summary

- avoid rewriting a native time input when its normalized value is already displayed
- preserve Chrome's active hour/minute editing segment across Formly model feedback
- add a regression test for identical value writes

## Root cause

`InputTimeDirective` emits after every digit. Formly immediately writes the same value back through `writeValue()`. Assigning that identical value to the native input resets Chrome's active time segment, so typing `18` into the hour field produced `08`.

## Verification

- `npm run test:file -- src/app/ui/input-time/input-time.directive.spec.ts --include src/app/ui/input-time/input-time-formly/input-time-formly.component.spec.ts` — 13/13 passed
- `npm run checkFile src/app/ui/input-time/input-time.directive.ts` — passed
- `npm run checkFile src/app/ui/input-time/input-time.directive.spec.ts` — passed
- repository pre-commit lint — passed (existing warnings only)
- real Chrome reproduction: typing hour `18`, moving to minutes, then typing `30` now results in `18:30`
- independent code review — no findings

Full `npm test` note: all package suites passed (765 tests) and release-note tests passed (24 tests). The Angular suite later hit two unrelated SQLite 2-second timeouts and Chrome crashed/disconnected after 12,792/14,127 tests. The focused affected suites pass.

Closes #9548

### DIFF
--- a/src/app/ui/input-time/input-time.directive.spec.ts
+++ b/src/app/ui/input-time/input-time.directive.spec.ts
@@ -35,6 +35,14 @@ describe('InputTimeDirective', () => {
       expect(nativeElement.value).toBe('17:00');
     });
 
+    it('does not rewrite an identical value', () => {
+      nativeElement.value = '17:00';
+
+      directive.writeValue('17:00');
+
+      expect(renderer.setProperty).not.toHaveBeenCalled();
+    });
+
     it('clears the display for an empty/invalid value', () => {
       directive.writeValue(null);
       expect(nativeElement.value).toBe('');
--- a/src/app/ui/input-time/input-time.directive.ts
+++ b/src/app/ui/input-time/input-time.directive.ts
@@ -53,11 +53,11 @@ export class InputTimeDirective implements ControlValueAccessor {
   }
 
   writeValue(value: string | null): void {
-    this._renderer.setProperty(
-      this._elementRef.nativeElement,
-      'value',
-      toPaddedClockStr(value),
-    );
+    const normalizedValue = toPaddedClockStr(value);
+    if (this._elementRef.nativeElement.value === normalizedValue) {
+      return;
+    }
+    this._renderer.setProperty(this._elementRef.nativeElement, 'value', normalizedValue);
   }
 
   registerOnChange(fn: (value: string) => void): void {


johannesjo:
Thanks for digging into this, and sorry for the slow reply — unfortunately I have to decline this one, because I could not confirm that it fixes #9548.

I A/B-tested the patch against the running app (settings → Time & Tracking → Schedule → Work Start, typing `1` then `8` into the hour segment):

| build | result |
| --- | --- |
| `master` | `08:00` (bug) |
| `master` + this PR | `08:00` (unchanged) |

### Why the guard can't help here

The stated root cause doesn't hold: assigning an **identical** value to a native `<input type="time">` does not reset Chrome's active segment. I checked this in real Chrome and Chromium, for synchronous, microtask, `setTimeout(0)` and `requestAnimationFrame` writes — typing `18` yields `18:00` in all of them. Only a *differing* value clobbers the segment (Blink already no-ops the identical assignment).

The real cause is one level up. Every keystroke emits `modelChange` → `saveGlobalCfg` → store → a **new** `cfg` object → `[model]` identity changes → `FormlyForm.ngOnChanges` rebuilds the field tree → **the `<input>` element is destroyed and re-created**. The segment typing buffer dies with the element.

Instrumenting the DOM on `master` shows it:

```
after typing "1": {"value":"01:00","sameEl":false}
after typing "8": {"value":"08:00","sameEl":false}
log: SET value="01:00" (was "")  ADDED input container ... REMOVED input container
```

`(was "")` is the key detail: the write always lands in a *freshly created* input, so `nativeElement.value === normalizedValue` is never true and the new guard never fires.

### About the test

`expect(renderer.setProperty).not.toHaveBeenCalled()` asserts an implementation detail rather than the user-visible behaviour, which is why it stayed green while the bug survived. For this class of bug the test has to run in a real browser (e2e) — a mocked `Renderer2` can't observe Chrome's segment state.

### What actually fixes it

`adjustToLiveFormlyForm` already sets `updateOn: 'blur'` for `input` / `textarea` / `duration` / `icon` / `color` precisely to avoid this per-keystroke model round-trip. `time` was simply never added to that list when the type was introduced in #8270:

```ts
if (item.type === 'time') {
  return { ...item, modelOptions: { ...item.modelOptions, updateOn: 'blur' } };
}
```

Verified in the app: typing `18` then `30` gives `18:30`, the value persists across a reload, and the input element is no longer re-created. It also removes one global-config store write (and one sync op) per keystroke, which is what the other five field types already avoid.

Closing this PR, but the analysis above is yours to use — a follow-up PR with that change plus a unit test in `adjust-to-live-formly-form.spec.ts` and an e2e for the type-`18`-then-`30` flow would be very welcome. Thanks again for taking the time!

PR #9555, opened 2026-08-12
fix(task): unique icons for move-to-bottom and move-to-backlog

calvin-fisher:
# Problem

The `MOVE_TO_BOTTOM` and `MOVE_TO_BACKLOG` icons are the same, creating visual confusion.

## Solution

- Let `MOVE_TO_BOTTOM` have the arrow_downward icon since it pairs with `MOVE_TO_TOP`
- Use a unique icon for `MOVE_TO_BACKLOG`

image

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [X] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki).

### DIFF
--- a/src/app/features/tasks/task-context-menu/task-context-menu-inner/task-context-menu-inner.component.html
+++ b/src/app/features/tasks/task-context-menu/task-context-menu-inner/task-context-menu-inner.component.html
@@ -237,7 +237,7 @@
       mat-menu-item
     >
       <span class="menuItemLeft">
-        <mat-icon>arrow_downward</mat-icon>
+        <mat-icon>list_alt</mat-icon>
         {{ T.F.TASK.CMP.MOVE_TO_BACKLOG | translate }}
       </span>
       <span class="key-i">{{ kb.moveToBacklog }}</span>
--- a/src/app/features/tasks/task-context-menu/task-context-menu-inner/task-context-menu-inner.component.html
+++ b/src/app/features/tasks/task-context-menu/task-context-menu-inner/task-context-menu-inner.component.html
@@ -237,7 +237,7 @@
       mat-menu-item
     >
       <span class="menuItemLeft">
-        <mat-icon>list_alt</mat-icon>
+        <mat-icon>move_down</mat-icon>
         {{ T.F.TASK.CMP.MOVE_TO_BACKLOG | translate }}
       </span>
       <span class="key-i">{{ kb.moveToBacklog }}</span>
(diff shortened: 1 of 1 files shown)


johannesjo (commented):
Thanks for spotting this — the duplicate is real. `Move to bottom` (line 175) and `Move to backlog` (line 240) both render `arrow_downward`, and their `@if` conditions genuinely overlap, so a top-level undone task in a project's regular list with the backlog enabled shows two identical down-arrows. Good catch.

Two things before this lands, though.

### 1. `list_alt` is already the project icon

```ts
// src/app/features/project/project.const.ts:8
export const DEFAULT_PROJECT_ICON = 'list_alt';
```

It's the fallback glyph for every project without a custom icon — sidebar, page title, add-task-bar. More directly: **this same context menu already uses it.** Line 389 passes `[defaultIcon]="DEFAULT_PROJECT_ICON"` to the *Move to project* submenu, so after this change `list_alt` appears next to "Move to backlog" in the parent menu *and* next to each project name one level down. That swaps one collision for a slightly worse one.

### 2. The mirror duplicate is left untouched

The backlog buttons are a pair and only one half is being changed. `isBacklog` comes from DOM ancestry (`component.ts:206`, matching `.backlog` in `work-view.component.html:442`), so for a top-level undone task **inside the backlog** the menu renders:

| Menu item | Icon | Condition |
|---|---|---|
| Move to top | `arrow_upward` | `!parentId \|\| isInSubTaskList` |
| Move to bottom | `arrow_downward` | same |
| Move to backlog | `list_alt` *(this PR)* | `!isBacklog` |
| Move to regular list | `arrow_upward` | `isBacklog` |

Two identical up-arrows — exactly the bug this PR fixes, just pointing the other way. The change also breaks the down/up symmetry that made the backlog pair readable at a glance.

### Suggested tweak

Treating the two backlog moves as the pair they are fixes both duplicates in one go:

```html
<!-- line 240, Move to backlog -->
<mat-icon>move_down</mat-icon>

<!-- line 258, Move to regular list -->
<mat-icon>move_up</mat-icon>
```

`move_down` / `move_up` depict an item hopping between lists, which is literally the action; they read as a pair, they stay distinct from the plain `arrow_upward` / `arrow_downward` reorder arrows, and they don't borrow the project glyph. Both are present in the shipped font (`@material-symbols/font-400`, full 466 KB woff2 — not subset, so no build change needed).

If you'd rather keep the diff to a single line, `low_priority` is Material's canonical "send to the back of the queue" symbol and works well for backlog — but it leaves the `arrow_upward` duplicate in the backlog view unresolved.

Happy to take the paired version as a follow-up if you'd prefer not to expand the scope here.

calvin-fisher:
> ### Suggested tweak
> Treating the two backlog moves as the pair they are fixes both duplicates in one go:
> 
> ```
> 
> <mat-icon>move_down</mat-icon>
> 
> 
> <mat-icon>move_up</mat-icon>
> ```
> 

Those icons are perfect. ✅

image

image

johannesjo:
Thank you very much! <3

PR #9562, opened 2026-08-13
Joplin Notes Sync

BigWebstas:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -150,5 +150,13 @@
     "author": "nekoritsme",
     "authorUrl": "https://github.com/nekoritsme",
     "stars": 0
+  },
+  {
+    "name": "Joplin Notes Sync",
+    "shortDescription": "One-way sync of project notes (and optional two-way task notes) to Joplin notebooks. Desktop only.",
+    "url": "https://github.com/BigWebstas/SuperProd-Joplin-Sync",
+    "author": "BigWebstas",
+    "authorUrl": "https://github.com/BigWebstas",
+    "stars": 0
   }
 ]


johannesjo:
Thanks for the submission, and welcome! 👋 The plugin itself looks solid — proper `manifest.json`, GPL-3.0, and an unusually candid README about its own limitations. Nice work.

One blocker plus a few small nits before this can go in:

### 🔴 Blocking: the file is no longer valid JSON

The comma after the previous entry's closing `}` is missing:

```
$ node -e "JSON.parse(fs.readFileSync('src/assets/community-plugins.json','utf8'))"
SyntaxError: Expected ',' or ']' after array element in JSON at position 6435 (line 146 column 5)
```

This isn't cosmetic — `plugin-management.component.ts` imports this file as a module, so it goes through the bundler's JSON loader and **the whole app build fails**. The new object is also indented 4 spaces instead of 2, so it would fail `prettier --check` even with the comma added.

This fix is valid JSON and prettier-clean:

```diff
     "authorUrl": "https://codeberg.org/fpindado",
     "stars": 0
-  }
+  },
+  {
+    "name": "Joplin Sync",
+    "shortDescription": "Sync your project and task notes to notebooks in Joplin",
+    "url": "https://github.com/BigWebstas/SuperProd-Joplin-Sync",
+    "author": "bigWebstas",
+    "authorUrl": "https://github.com/BigWebstas",
+    "stars": 0
+  }
 ]
```

(CI hasn't run yet — first-time contributor PRs need manual approval — which is why nothing flagged this automatically.)

### Minor

- **`author`**: `"bigWebstas"` — lowercase `b` doesn't match your GitHub login, your `authorUrl`, or your own `manifest.json` (`"author": "BigWebstas"`). It's rendered as the link text next to your profile link, so the mismatch shows.
- **`name`**: the list says "Joplin Sync" but `manifest.json` says "Joplin Notes Sync" — users would see a different name after installing. Worth aligning to one.
- **`shortDescription`**: "Sync your project and task notes to notebooks in Joplin" is a bit more than the plugin promises per your own README — project notes are one-way only, and it's desktop/Electron only (since it needs `nodeExecution` to reach Joplin's `127.0.0.1` API). The community list has no desktop-only badge, so web/Android users get no hint. Maybe something like:

  > One-way sync of project notes (and optional two-way task notes) to Joplin notebooks. Desktop only.

### One note, not a change request

I appreciate that your README is upfront about two-way task-note sync racing with Super Productivity's own cross-device sync, and that the mitigation is a heuristic rather than a guarantee. That's your call in your own repo — just flagging that if it does produce duplicated or reverted tasks in the wild, the reports will most likely land in *this* tracker as sync bugs. Keeping the two-way mode opt-in and clearly labelled (as you have) is the right instinct.

Push the fixes and I'll get this merged. 🙏

johannesjo:
Thanks for turning this around so quickly — all five points from my earlier review are addressed in ffab5c2c. The JSON is valid and prettier-clean, and `name` / `author` / description now match your manifest.

I went through the plugin source too, and it's genuinely tidy: the only URL in the entire codebase is the `127.0.0.1:41184` default you expose as an editable setting, the token goes through the local-only secret API, there are zero `console.*` calls anywhere (so no path for the token or note content into the log), and no eval/telemetry. I also confirmed the v1.4.0 release zip is byte-identical to the repo source.

One mechanical thing left: #9479 landed on master and appended its own entry at the same array tail, so this now conflicts. Rebase and re-append your block after the `JSON to Tasks` entry (add a `,` after its closing `}`) and I'll merge.

Also worth updating the PR title to "Joplin Notes Sync" to match the renamed entry.

On the two-way task-note sync: I appreciate how candid the README is about the settle-window heuristics not fully closing the race against our own sync. It's off by default and clearly labelled, which is the right call — just flagging that reports from it will likely land in this tracker rather than yours.

johannesjo:
Thank you very much! <3

PR #9564, opened 2026-08-13
feat: add Task Color and Magic Todo plugins to community JSON file

1nsp1r3rnzt:
## Problem

Add the Task Color and Magic Todo plugins to the community plugin list so they are available to users.

## Solution

Added the Task Color plugin and Magic Todo plugin to the community JSON file.

## Type of Change

* [ ] Bug fix
* [x] New feature
* [ ] Breaking change
* [ ] Refactoring
* [ ] Documentation
* [ ] Other

## Checklist

* [ ] I have run `npm run checkFile` on changed `.ts`/`.scss` files
* [ ] Existing tests still pass


johannesjo:
@1nsp1r3rnzt can you have a look at the conflicts? Also formatting looks off.

johannesjo:
Thanks for these — the source is readable and unminified, both release zips match the repo contents exactly, and neither plugin makes any unexpected network call. A few things before this can go in:

1. **Formatting**: the two new objects are indented differently from the rest of the file. Please run `npx prettier --write src/assets/community-plugins.json` and commit the result.
2. **Star count**: `sp-task-color-plugin` has 4 stars, not 0. (Magic ToDo's `0` is correct.)

The one I'd like fixed upstream before listing, in Magic ToDo — `plugin.js:349` adds a `message` listener with no `event.origin` / `event.source` check, and on `magic-get-config` it replies with the whole config object, including `apiKey`, using `'*'` as targetOrigin:

```js
event.source.postMessage({ type: 'magic-config-response', config: config }, '*');
```

That means any other frame in the app — including another installed community plugin — can ask for and receive the user's LLM API key, and the same handler exposes `magic-task-crud` add/update/delete. Since users typically have several plugins installed at once, I'd rather not list this until the handler validates the sender and stops broadcasting the key. Task Color has the same missing check but a much weaker sink, so it's lower priority there.

Two smaller things, your call:

- Task Color stores its state inside `task.notes` as a `__task_colors__={...}` marker line. That's user-visible text that rides through sync, export and backups, and every colour change becomes an `updateTask` on `notes`. Synced plugin storage (`persistDataSynced`) would avoid touching user content.
- Task Color logs task titles via `PluginAPI.log` (`plugin.js:134`, and `:738` logs raw message payloads). That feeds the app's exportable log history, so IDs only would be safer for users.

Also: the `shortDescription` for Task Color is ~200 chars and reads as a feature list — something like "Set background colours on tasks, including tag-based colouring and a side-panel colour picker" would match the rest of the list better. And `authorUrl` points at your personal account while the repos live under the `FlourishLifeTech` org; other entries link to the repo owner.

1nsp1r3rnzt:
Thanks for the thorough review.

#### For your Task Color suggestions,
Both items are fixed in v1.1.0
- Colors now live in synced plugin storage instead of task.notes, with a migration button setting to backport and clean up existing data
- Logging now uses IDs only

#### For your Magic ToDo suggestions
- event.source is now validated against the registered iframe window on every inbound message
- magic-get-config strips apiKey from the response before sending
- LLM calls were moved to a host-side proxy so the iframe never handles the key
- All outbound replies use event.origin instead of '*'

#### For Community formatting,
- I used the command you shared to update the format.
- I corrected the authorUrl, description, and the stars as you suggested. 

Let me know if you'd like any other changes before merge.

johannesjo:

Thanks for the rework. I read both plugins at their current heads – Task Color at `2da9a7b` (`plugin.js`, `index.html`, `manifest.json`, ~66 KB, read in full) and Magic ToDo at `581cbfc` (`plugin.js`, `src/main.js`, `src/style.css`, manifest and build files, ~85 KB, read in full; skipped the screenshots and the lockfile).

The entry itself is in good shape now: formatting matches the rest of the file, `authorUrl` points at the org, the Task Color description is trimmed, and the parsed diff adds two entries and removes none. I also confirmed the changes you described – Task Color's colours are in `persistDataSynced` with a working legacy backport and a migrate button, its host-side logging is IDs-only across all 23 `PluginAPI.log` calls, and Magic ToDo does now proxy the AI call through the host. Neither plugin has any `eval`, `new Function`, dynamic `import()`, injected `<script>`, `document.write` or `child_process` in shipped code; Task Color makes no network call at all and its release zip is byte-identical to its source.

### Blocking

- **`sp-magic-todo/plugin.js:410-419`** – the `event.source` guard was added to `magic-get-config`, `magic-save-config`, `magic-get-current-task`, `magic-open-task-picker`, `magic-task-crud` and `magic-ai-request`, but not to `magic-iframe-ready` – which is the one message that *establishes* `iframeWindow` (it's the only place in the file that assigns it a window). So any frame can post `{type: 'magic-iframe-ready'}` and the handler will:
  1. reply with `magic-init` carrying the full `config` object, `apiKey` included, through `sendToIframe` – which posts with targetOrigin `'*'` at `plugin.js:24`; and
  2. become `iframeWindow`, which is exactly what every other guard tests against, so the sender can then drive `magic-task-crud` add/update/delete and spend the user's LLM credits via `magic-ai-request`.

  Stripping the key out of `magic-get-config` closed the front door and left this one open. (In the desktop app the realistic sender is another installed plugin's panel; on the web build I couldn't find any `frame-ancestors` or COOP header in this repo, so a page that frames or `window.open`s the app would hold a handle too.)

  Cheapest fix: strip `apiKey` from the `magic-init` payload the same way `magic-get-config` already does at `:403-404`. The iframe doesn't need it over postMessage – it loads the config itself at `src/main.js:1055` and reads the key straight from the settings DOM in `testConnection`. Validating `event.source` against your own iframe's `contentWindow` before accepting `magic-iframe-ready` would close the CRUD half as well.

### Should fix before merge

- **`sp-magic-todo/plugin.js:536-558` + `src/main.js:787-814`** – moving the LLM call host-side left the config behind. The host reads `config` once in `init()` (localStorage preload at `:540`, `loadConfig()` at `:550`) and never again; `saveSettings()` in the iframe writes straight to `PluginAPI.persistDataSynced('magicTodoConfig')` and never sends `magic-save-config` (that string appears nowhere in `src/main.js`, the repo `index.html`, or the built bundle in `plugin.zip`), and the host registers no `PERSISTED_DATA_CHANGED` hook. `callOpenAI` sends only model/maxTokens/temperature, so `baseUrl` and `apiKey` come from the host copy.

  Net effect on a fresh install: configure the endpoint, click **Break down**, and the host throws `Configure the AI endpoint first` at `:495-497`. The plugin can't work until the app is restarted. There's a second-order version too – change the settings mid-session, reopen the panel, and the host answers `magic-iframe-ready` with its stale config, which `src/main.js:990-991` then writes over the iframe's freshly loaded one.

  Either have `saveSettings()` post `magic-save-config` (the handler already exists at `:422`), or call `loadConfig()` at the top of the `magic-ai-request` handler.

- **`sp-task-color-plugin/index.html:429-431` and `:444-446`** – the side-panel colour set/clear calls `PluginAPI.updateTask(taskId, {notes: newNotes})` unconditionally, and `newNotes` is derived from `item.task.notes`, a snapshot taken by the last `loadTasks()`. `setColorInNotes` returns that snapshot byte-identical when the note carries no `__task_colors__` marker, so on migrated data the write is a no-op at best – and a revert at worst.

  The panel does re-render on `ANY_TASK_UPDATE`, but that hook is dispatched from an effect that injects `LOCAL_ACTIONS` (`src/app/plugins/plugin-hooks.effects.ts:58`), so it fires only for locally-originated changes. A note edited on another device and applied by sync leaves the panel holding the pre-edit snapshot, and the next colour click on that row writes it back – silently reverting the remote edit to `notes`.

  Your host-side path already guards this correctly at `plugin.js:70-74`; the same `if (newNotes !== notes)` in both iframe handlers fixes it.

### Nits

- **`sp-task-color-plugin/plugin.js:791`** – the host `message` listener still has no `event.origin` / `event.source` check. Weak sink as before (the reply only goes out after the user confirms the Tag Colors dialog), but it's the same shape as the magic-todo fix and cheap to apply here too; note the reply carries the tag-id → colour map to whoever asked.
- **`sp-task-color-plugin/plugin.js:500-519`** – each colour-picker open attaches `click` and `contextmenu` listeners to the host `document` and stashes a `cleanup` on `window['_taskColorCleanup_' + dialogId]` that's never called (and wouldn't remove the anonymous `contextmenu` listener at `:501` anyway). `plugin.js:696-698` leaks three more per task-picker open. They accumulate for the session and each runs `closest()` on every click in the app. Magic ToDo removes its equivalents at `plugin.js:343-345` – same pattern here.
- **`sp-magic-todo/src/main.js:344-349` and `:703-720`** – `doCrud` returns the boolean `true`, but the create path tests `addRes && addRes.id`, so `addedIds` is always empty and the snack reads `Created 0 subtask(s)` after a breakdown that actually worked. The host already returns the id at `plugin.js:482`; returning the reply object from `doCrud` fixes it.
- Entry says `stars: 7` for Task Color; the repo shows 8 today (Magic ToDo's `0` is right). Also `sp-magic-todo/README.md` tells users to download `magic-todo-plugin.zip` while the repo ships `plugin.zip`, and its Files section calls the LICENSE "MIT License" when the file is the Cooperative Nonviolence License.

### Questions

- Magic ToDo's `plugin.zip` ships a vite-bundled, minified `index.html` (35 KB) with no byte-for-byte counterpart in the repo – `src/main.js` and `src/style.css` are the readable source. I read those in full and audited the bundle itself: its only URLs are the `localhost:8080/v1` default and the Open Collective link, its only two `fetch` calls are `testConnection` and vite's own modulepreload polyfill, and it has no dynamic execution. But I can't confirm the bundle actually corresponds to that source without building it. Could you attach the build output to a tagged release, or otherwise make the zip reproducible? (Task Color's zip needs none of this – it's identical to source.)
- The API key is written to `persistDataSynced('magicTodoConfig')`, with `localStorage` as a fallback. That means it rides the app's sync to whatever backend the user configured, and lands in exports and backups. Intended, or would you rather keep the key device-local?
- `testConnection` (`src/main.js:822`) still fetches from inside the iframe with the key in an `Authorization` header, while breakdowns now go through the host proxy. Deliberate, or left over from the move?
- Magic ToDo still stores per-task AI instructions in `task.notes` as a `__magic_todo_ai__=` marker line (`src/main.js:116-133`) – the same pattern you moved out of `task.notes` for Task Color. Worth the same treatment, or is it different enough to stay in the note?
- Neither manifest declares `http` / `allowedHosts`, so both plugins reach the network with raw `fetch` rather than `PluginAPI.request`. Nothing breaks (only `http` and `nodeExecution` are actually enforced), but it does mean the manifest can't tell a user where Magic ToDo sends their task titles. Similarly, Task Color declares only `getTasks`, `updateTask` and `getAllTags` while the code also calls `getSelectedTask`, `getFocusedTask`, `getAllProjects`, `updateTag`, `openDialog`, `showSnack` and the synced-storage pair – is the short list deliberate?

---

johannesjo:
Maintainer note on the scoping question I flagged in the review — it's settled, so don't wait on it:

**Both plugins are welcome on the community list.** The Cooperative Nonviolence licence is fine (several listed entries carry no licence at all), and Magic ToDo being a second bring-your-own-LLM-key plugin alongside `ai-assistant-plugin` is not a problem either. No need to change the licence, split the PR, or drop an entry.

What still blocks the merge is only the code:

1. **Blocking** — `sp-magic-todo/plugin.js:410-419`: `magic-iframe-ready` has no `event.source` check, so any frame can claim the iframe role and receive the `apiKey` in the `magic-init` reply.
2. **Should fix** — Magic ToDo's host-side config is read once at init and never refreshed, so the plugin can't work until the app is restarted after configuring.
3. **Should fix** — `sp-task-color-plugin/index.html:429-431` / `:444-446`: guard the `notes` write with `if (newNotes !== notes)`, or a colour click reverts a note edited on another device.

The nits and questions are worth a look but aren't gating. Ping me when those three are in and I'll re-review.

1nsp1r3rnzt:
Thanks for your detailed feedback. I fixed above 3 issues in the latest release for both.

PR #9568, opened 2026-08-14
fix(op-log): skip compaction while boot hydration replay is in progress (#9084)

### LINKED ISSUE #9084: op-log: compaction racing boot hydration replay can cache state missing tail-op effects (op loss)
## Summary

Found while reviewing the #8469 fix (PR #9083, quiesced snapshot/compaction capture): **compaction can race boot hydration replay** and persist a state cache whose `lastAppliedOpSeq` covers tail ops whose reducer effects are NOT in the captured state. Unlike #8469 (re-replay, double-apply), this is in the **op-LOSS direction**: the next boot's tail replay starts after ops that never made it into the cached state, skipping them permanently — and compaction's `deleteOpsWhere` can prune the evidence.

Pre-existing behavior. Neither of the two open snapshot-safety PRs covers it, because the racing ops are *sequenced-but-not-yet-redispatched* — the inverse of every signal those guards watch:

- the #9083 quiesce (`flushThenRunExclusive` + pending/deferred bail) sees dispatched-but-unsequenced actions; these ops are the opposite (durable, not yet dispatched),
- the #9081 phantom-change guard (divergence flag + pending count + deferred buffer) is equally quiet here: no write has failed, nothing is pending, nothing is deferred.

## Scenario

1. Previous session ends with the persisted compaction counter ≥ `COMPACTION_THRESHOLD`.
2. Boot: `hydrateStore()` dispatches `loadAllData(snapshot)`, then prepares the bulk tail replay (`getOpsAfterSeq(lastAppliedOpSeq)` → `bulkApplyOperations`). Hydration holds no lock across this window, and there are `await` boundaries between the snapshot load and the tail dispatch.
3. An early persistent write lands in that window (e.g. TODAY_TAG repair during `loadAllData` handling) → the persist effect writes it and, with the counter over threshold, fires `triggerCompaction()`.
4. `_doCompact` runs: the capture pipeline is drained (the early write is durable), `getPendingRemoteOps` is empty (the tail ops are local, `applicationStatus` applied from their original session), pending counter is 0, deferred buffer empty → all guards pass.
5. The captured NgRx state contains only the snapshot + the early write — **none of the tail ops' effects** (their re-dispatch hasn't happened yet). `getLastSeq()` covers them. The cache is saved; old tail ops ≤ lastSeq that are synced + past retention get deleted.
6. Next boot hydrates from that cache and replays nothing → the tail ops' effects are silently gone on this device, and the corrupted cache state syncs onward.

## Suggested fix

Gate `compact()` on hydration completion — e.g. skip with a warn while `DataInitService` hasn't signaled `isAllDataLoadedInitially`, mirroring the existing `getPendingRemoteOps`/pending-counter bails in `OperationLogCompactionService._doCompact`. (`saveCurrentStateAsSnapshot`'s two call sites are already sequenced after tail replay inside `hydrateStore()`, so the gate matters for the compaction trigger path.)

Whichever of #9081 / #9083 merges second is the natural place to add the gate, since all three guards then live at the same cutoff.

## Notes

- Reachability needs a session that ended over-threshold plus an early boot write; not reproduced end-to-end yet — a regression spec should stage the hydrator between `loadAllData` and the bulk dispatch and fire `compact()`.
- Related: #8469 / PR #9083 (re-replay direction), #8751 / PR #9081 (phantom changes), #8308/#8439 (original lost-update lock fix).

_Found during multi-agent review of the #8469 fix._

AlvaroBalbin:
This is a small, targeted change to src/app/op-log/apply/hydration-state.service.ts, .../operation-log-compaction.service.spec.ts, .../operation-log-compaction.service.ts, .../operation-log-hydrator.service.spec.ts and a few others that resolves the reported issue without touching anything unrelated. Fixes #9084.

I pushed this to my fork first and the project's own CI workflows pass on the commit.

### DIFF
--- a/src/app/op-log/apply/hydration-state.service.ts
+++ b/src/app/op-log/apply/hydration-state.service.ts
@@ -130,17 +130,10 @@ export class HydrationStateService implements RemoteApplyWindowPort {
 
   /**
    * #9084: true for the full duration of OperationLogHydratorService's
-   * hydrateStore() run, from before the snapshot's loadAllData dispatch until
-   * the tail ops it implies have actually been replayed into the store. A
-   * write landing in that window (e.g. TODAY_TAG repair reacting to
-   * loadAllData) is durable and unblocks the compaction counter before the
-   * tail ops are re-dispatched, so none of compact()'s other guards see it:
-   * getPendingRemoteOps is empty (the tail ops are local, already
-   * 'applied' from their original session) and nothing is pending or
-   * deferred. Without this flag compact() would cache a state missing the
-   * tail ops' effects under a lastAppliedOpSeq that covers them, and prune
-   * the very ops the next boot needs to recover them. Set/cleared by
-   * OperationLogHydratorService around each hydrateStore() run.
+   * hydrateStore() run, covering the gap between the snapshot's loadAllData
+   * dispatch and the tail-op replay on top of it. Compaction must not run in
+   * that gap — see the guard in OperationLogCompactionService._doCompact for
+   * the full reasoning. Set/cleared by the hydrator around each run.
    */
   isHydrationInProgress(): boolean {
     return this._isHydrationInProgress;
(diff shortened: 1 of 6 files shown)


johannesjo:
Thank you very much! <3

PR #9596, opened 2026-08-14
fix(config): commit image-input fields on blur to stop per-keystroke ops

### LINKED ISSUE #9591: image-input commits on every keystroke: one synced op per character typed in background image URL
Found while reviewing #9582 (the `time` field fix for #9548). Same root cause, different field type — filing separately since it is independent and shipped.

### Problem

`image-input` is not in the `updateOn: 'blur'` list in `src/app/util/adjust-to-live-formly-form.ts`, so it commits on **every keystroke**. Its component is a plain free-text input bound straight to the control (`src/app/ui/formly-image-input/formly-image-input.component.html:9-17`):

```html
<input [name]="props.name" [required]="!!props.required" [readonly]="props.readonly"
       [formControl]="formControl" [formlyAttributes]="field" matInput placeholder="https://..." />
```

It is used for the project/tag background images (`src/app/features/work-context/work-context.const.ts:178,186`) and in the wallpaper dialog (`src/app/core/theme/dialog-wallpaper/dialog-wallpaper.component.ts:33,41`).

In the project/tag settings dialog the per-keystroke commit reaches the store:

`onModelChange` → `_applyChanges(next)` → `ProjectService.update(...)` / `TagService.updateTag(...)`
(`src/app/features/work-context/dialog-work-context-settings/dialog-work-context-settings.component.ts:136-140, 156-184`)

`updateProject` is op-log captured, so **typing a background-image URL emits one persisted, synced op per character**. A 60-character URL is ~60 ops replicated to every device instead of 1.

### Why it went unnoticed

Unlike #9548 there is no visibly wrong value — the symptom is op-log/sync spam plus redundant IndexedDB writes. Note the dialog assigns the emitted model back (`this.entityData = next`, line 136), which satisfies formly's `_modelChangeValue` guard, so the field is *not* re-created here and the DOM-state corruption from #9548 does not occur. This is purely the write-amplification half of the same underlying issue.

### Suggested fix

Add `'image-input'` to the type list in `adjust-to-live-formly-form.ts` (one line, same as #9582 did for `time`).

### The broader point

This is the second time the list has been silently forgotten (`time` was added in #8270 and went a year before #9548 was diagnosed). The failure mode of an allow-list here is silent, and it is per-type. Two structural options, both larger than this fix:

1. **Fix the round-trip at the source.** `ConfigFormComponent` binds `[model]="cfg()"`, and NgRx returns a fresh object every update, so formly's `immutable` guard never hits. Having the form ignore an incoming `cfg` that deep-equals what it just emitted would remove the whole bug class and let the type list be deleted. (`DialogWorkContextSettingsComponent` already does the right thing, which is why only the write-amplification shows up there.)
2. **A registration-time guard.** Assert in a unit test that every type registered in `formly-config.module.ts` appears in either the blur list or an explicit immediate-commit list, so a new type fails a test instead of shipping silently.

Also latent, same class: `type: 'date'` is registered (`formly-config.module.ts:124-127`, a free-text `MatDatepickerInput`) but not yet used in any live form, and `keyboard` renders a text input that survives only because it commits on `keydown`.

costajohnt:
## Problem

Fixes #9591

`image-input` is missing from the `updateOn: 'blur'` list in `src/app/util/adjust-to-live-formly-form.ts`, so the project/tag background image field commits on every keystroke. In the settings dialog that reaches `onModelChange` → `_applyChanges` → `ProjectService.update` / `TagService.updateTag`, both of which are op-log captured, so typing a 60-character URL emits roughly 60 persisted and synced ops instead of one, plus the matching IndexedDB writes.

As the issue notes, there is no visible symptom here. The dialog assigns the emitted model straight back (`this.entityData = next`), which satisfies formly's `_modelChangeValue` guard, so the field is not re-created mid-edit and no input state is lost. This is only the write-amplification half of #9548's root cause.

## Solution

Add `image-input` to the same branch `time` was added to in #9582. One line, plus a unit spec.

**Where the fix fires:** the project/tag settings dialog builds its fields with `adjustToLiveFormlyForm(buildWorkContextSettingsFormCfg(...))` (`dialog-work-context-settings.component.ts:104`), so it picks this up. The other `image-input` site, the wallpaper dialog, builds its fields with a bare `buildWallpaperFields()` and only writes in `save()`, so it never had the per-keystroke commit and is unaffected either way.

## Verification

The new spec fails against the unfixed file with `Expected undefined to be 'blur'` and passes with it, 20/20 in that file. `tsc --noEmit` clean, `npm run checkFile` clean on both changed files.

I did not add an e2e for this one. The observable is op count rather than rendered state, and #9582's e2e was justified by visible digit loss that has no analogue here. Happy to add one if you would rather have the coverage.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/util/adjust-to-live-formly-form.spec.ts
+++ b/src/app/util/adjust-to-live-formly-form.spec.ts
@@ -151,6 +151,18 @@ describe('adjustToLiveFormlyForm', () => {
       // the same branch also opts `time` into Enter-to-commit
       expect(result[0].templateOptions?.keydown).toBeDefined();
     });
+
+    // #9591: the project/tag background image URL committed per keystroke, and
+    // `updateProject` is op-log captured, so a 60-character URL emitted ~60
+    // persisted and synced ops instead of one.
+    it('should add blur update behavior to image-input fields', () => {
+      const items: FormlyFieldConfig[] = [{ key: 'imageField', type: 'image-input' }];
+
+      const result = adjustToLiveFormlyForm(items);
+
+      expect(result[0].modelOptions?.updateOn).toBe('blur');
+      expect(result[0].templateOptions?.keydown).toBeDefined();
+    });
   });
 
   describe('fieldGroup processing', () => {
--- a/src/app/util/adjust-to-live-formly-form.ts
+++ b/src/app/util/adjust-to-live-formly-form.ts
@@ -24,13 +24,21 @@ export const adjustToLiveFormlyForm = (
     // itself. For `time` that ate the first digit of every hour typed (#9548).
     // The general fix is for the config form to ignore an incoming `cfg` that
     // deep-equals what it just emitted; until then, input-like types opt in here.
+    //
+    // The cost is not always visible state loss. `image-input` is a plain text
+    // input used for the project/tag background image, and the settings dialog
+    // assigns the emitted model straight back, so no field rebuild happens
+    // there — but every keystroke still reached `updateProject`, which is
+    // op-log captured, emitting one persisted and synced op per character
+    // typed (#9591).
     if (
       item.type === 'input' ||
       item.type === 'textarea' ||
       item.type === 'duration' ||
       item.type === 'icon' ||
       item.type === 'color' ||
-      item.type === 'time'
+      item.type === 'time' ||
+      item.type === 'image-input'
     ) {
       return {
         ...item,


johannesjo:

PR #9612, opened 2026-08-15
feat: add full worklog export

### LINKED ISSUE #4393: 💡 Support full export
### Problem Statement

Currently, I believe, the only options for exporting is Month and Week ranges per worklog. This is difficult to utilize when you have many worklogs and need to backfill or do analysis on many months. It would be great to get a full export option so further modifications can be made with the CSV output. 

Issues that may be solved (or accept as a sufficient workaround) by adding full export:
https://github.com/johannesjo/super-productivity/discussions/780
https://github.com/johannesjo/super-productivity/issues/1981
https://github.com/johannesjo/super-productivity/issues/4164

### Possible Solution

Ability to export all worklog history for all worklogs at once. 

_No response_

### Additional context

_No response_

pupuking723:
Fixes #4393.

Adds an export action next to the total worklog time. It derives the first and last recorded worklog dates and reuses the existing export dialog, so users can export all available history without exporting each month separately. Empty worklogs leave the action disabled and do not open a dialog.

Also updates the Worklog wiki and adds regression coverage for the full date range and empty state.

Verification:
- `npm run checkFile src/app/features/history/history.component.ts`
- `npm run checkFile src/app/features/history/history.component.spec.ts`
- `npm run checkFile src/app/features/history/history.component.scss`
- `npm run test:file -- src/app/features/history/history.component.spec.ts` (6 tests passed)
- `git diff --check`

### DIFF
--- a/docs/wiki/4.21-Worklog.md
+++ b/docs/wiki/4.21-Worklog.md
@@ -44,7 +44,7 @@ So the worklog is a **reorganized view** of the same “time spent per day” da
 
 - **Inline time correction** — Leaf-task rows can be corrected inline. The app updates the task's “time spent per day” value for that date, so the corrected time is reflected in History, reports, and metrics.
 - **Parent rows** — Parent tasks show the sum of their subtasks and are not edited directly, which avoids double-counting.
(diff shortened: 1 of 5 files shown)


johannesjo:
Thank you very much! <3

Issue #8792, opened 2026-07-06
🚨 No fullscreen or overlapping with navigation buttons


Sanja2012:
<img width="1280" height="484" alt="Image" src="https://github.com/user-attachments/assets/24dd1905-9695-4d40-ace5-bff76788f9c4" />

### Version Used
Android App v18.13.1 / from F-Droid APK

### Operating System and version
Android 16, POCO F7, HyperOS 3.0.6.0

### Describe the Bug
The app does not display in fullscreen. The Android system navigation buttons (nav bar) overlap with the app's own UI content instead of the app adjusting its layout to leave space for them. (See picture above!)

### Additional Context
- Navigation mode: 3-button navigation
- This may be a HyperOS-specific issue with how the safe-area/insets are reported to installed PWAs, since HyperOS is known to deviate from AOSP behavior in some display/inset handling.
- Possibly related to PWA fullscreen/safe-area handling on Android in general, but device-specific behavior (HyperOS) should be considered as a contributing factor.

### Expected Behavior

The app should either run in true fullscreen (hiding the navigation bar) or correctly account for the navigation bar's safe area, so its own UI elements are not obscured.

### Steps to Reproduce

1. Open the app on Android
2. Observe the bottom area of the screen
3. The system navigation buttons are overlapping the app's UI, instead of the app content being properly inset above them

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

persy:
I have a similar problem: some items still appear below the navigation bar and cannot be selected.
How to reproduce it: Open the notes for a project while using the three-button navigation
What should happen: The element should open above the toolbar
Device: Motorola 50 Edge Neo
Android: 16 (official release)
App version: 18.16.0_L2A and 18.16.0_L2AF

<img width="1200" height="613" alt="Image" src="https://github.com/user-attachments/assets/d6229807-09c7-4640-a8ea-6040102de8b9" />

johannesjo:
@persy Thanks for the screenshot — that made this findable. It's a real bug, but a
different one from the overlap this issue was about (the bottom bar itself still reserves
the full system-bar height correctly). What you hit is the *menu* that opens from the
[...]

PR #9613, opened 2026-08-15
fix(mobile): reserve the full bottom safe area and dock the add-task FAB

### LINKED ISSUE #5563: 🚨 bad ui when using it as a webapp
### Version Used

Super productivity 16.3 webapp

### Expected Behavior

Normal ui for the navigation bar

### Current Behavior

On iphone 16 pro max ios version 26.1 
The ui for the bottom bar is very small and renders under the navigation pill

Image

### Steps to Reproduce

Share the url for the webapp and save it to home screen as a webapp 

### Console Output

```shell

```

johannesjo:
Split out of #9304, which carried these two alongside a CSS-token sweep. They are
the part of that branch CI cannot speak to — both need a look on a real device —
so they get their own PR rather than blocking a lint gate behind a layout change.

## Reserve the full bottom safe area

Fixes #5563.

`--bottom-nav-safe-area` was half the inset on iOS, on the theory that the home
indicator is thin and translucent so a full inset would read as an oversized gap.
It doesn't: the nav's background fills the reserved band
(`mobile-bottom-nav.component.scss` sets `height: calc(--bottom-nav-height +
--bottom-nav-safe-area)` with a matching `padding-bottom`), so there is no gap to
see — only the 44px button row moves up out of the band.

At half of a 34pt iOS inset the icons and the active-item pill still land inside
the home-indicator band, which is what #5563 reports on an iPhone PWA. Native
Android already needed the full inset for the same reason (#8792) and was getting
it via a `body.isNativeMobile:not(.isIOS)` override in `styles.scss`. One rule for
both is simpler and correct in both places, so the override is gone.

`viewport-fit=cover` is set in `index.html`, so `env(safe-area-inset-bottom)`
resolves in the iOS PWA. Everything stacked on the bar — content padding, snack
bars, and the hint below — reads through the same token and moves with it.

## Dock the add-task FAB

The FAB sat *in* the icon row with a `var(--s)` gutter, which put the 56px button
within 7px of the neighbouring buttons' active pill on a 390px viewport — it read
as sitting on top of the row rather than beside it. Gutter goes to `var(--s2)`, a
full nav-button width each side.

`margin-bottom: var(--bottom-nav-height)` docks it centred on the bar's top edge:
with `align-items: center` the margin box is centred in the `--bottom-nav-height`
content box, so a bottom margin of exactly that height puts the FAB's centre on
the border — for any FAB size. The `.smaller` variant resets it.

## Anchor the floating onboarding hint like a toast

The un-anchored variant (the closing "explore" hint has no target element) sat
dead centre of the viewport, where it reads as a modal with no backdrop and covers
whatever is underneath. It moves to the bottom-end corner, clear of the side nav,
the header and the mobile bottom nav — the last of those the same way snack bars
do it in `_overwrite-material.scss`.

## What to look at

Neither change is visible to CI, so the useful review is a screenshot:

- **A notched iPhone.** The bar grows from 44+17px to 44+34px. Check the icons and
  the active pill clear the home indicator, and that the extra 17px doesn't read as
  dead weight.
- **Android, 3-button and gesture.** Should be unchanged — it already had the full
  inset — but the override moved, so it is worth confirming.
- **The FAB overhang.** It now stands 28px proud of the bar, over scrollable
  content, at the bar's z-index. That strip will intercept taps on whatever is
  beneath it. Inherent to a docked FAB and presumably what we want, but it is the
  judgement call the diff cannot make.

`ng build` passes. There is no automated coverage for any of the above.

### DIFF
--- a/src/app/core-ui/mobile-bottom-nav/mobile-bottom-nav.component.scss
+++ b/src/app/core-ui/mobile-bottom-nav/mobile-bottom-nav.component.scss
@@ -81,9 +81,16 @@
     flex: 0 0 var(--s7);
     width: var(--s7);
     height: var(--s7);
-    margin-left: var(--s);
-    margin-right: var(--s);
-    margin-bottom: var(--s3);
+    // A full nav-button width of gutter on each side. At var(--s) the 56px FAB
+    // came within 7px of the neighbouring buttons' active pill on a 390px
+    // viewport, so it read as sitting on top of the row rather than beside it.
+    margin-left: var(--s2);
+    margin-right: var(--s2);
+    // Dock the FAB centred on the bar's top edge instead of sinking it into the
+    // icon row. With `align-items: center` the margin box is centred in the
+    // --bottom-nav-height content box, so a bottom margin of exactly that
+    // height puts the FAB's centre on the bar's top border — for any FAB size.
+    margin-bottom: var(--bottom-nav-height);
     box-shadow: var(--whiteframe-shadow-4dp);
     transition: var(--transition-standard);
 
--- a/src/app/features/onboarding/onboarding-hint.component.scss
+++ b/src/app/features/onboarding/onboarding-hint.component.scss
@@ -1,3 +1,5 @@
+@use '../../../styles/globals' as *;
+
 :host {
   pointer-events: none;
 }
@@ -8,16 +10,32 @@
   pointer-events: auto;
   animation: hint-appear var(--transition-duration-l) var(--ani-enter-timing);
 
+  // The un-anchored variant (the closing "explore" hint has no target element).
+  // It used to sit dead centre of the viewport, where it reads as a modal with
+  // no backdrop and covers whatever is underneath — settings rows, the task
+  // list. Park it in the bottom-end corner like a toast instead: out of the
+  // centred content column, clear of the side nav and the header.
   &.floating {
-    top: 50%;
-    left: 50%;
-    bottom: auto;
-    transform: translate(-50%, -50%);
-    animation-name: hint-appear-floating;
+    top: auto;
+    left: auto;
+    inset-inline-end: var(--s2);
+    inset-block-end: calc(var(--s2) + var(--safe-area-bottom));
+    transform: none;
 
     .hint-chip {
-      max-width: 340px;
+      max-width: min(340px, calc(100vw - var(--s4)));
     }
+  }
+}
+
+// Clear the mobile bottom nav, the same way snack bars do
+// (see src/styles/components/_overwrite-material.scss).
+@include mq(xs, max) {
+  :host-context(body.hasMobileBottomNav) .hint-container.floating {
+    inset-block-end: calc(
+      var(--s2) + var(--bottom-nav-height) + var(--bottom-nav-safe-area) +
+        var(--bottom-nav-extra-offset)
+    );
   }
 }
 
@@ -155,18 +173,6 @@
   to {
     opacity: 1;
     transform: translateY(0);
-  }
-}
-
-@keyframes hint-appear-floating {
-  from {
-    opacity: 0;
-    transform: translate(-50%, calc(-50% + 4px));
-  }
-
-  to {
-    opacity: 1;
-    transform: translate(-50%, -50%);
   }
 }
 
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -56,12 +56,6 @@ body.isNativeMobile {
   }
 }
 
-// Native Android's opaque system bars need the full inset, not the iOS half
-// default — see --bottom-nav-safe-area in _css-variables.scss (#8792).
-body.isNativeMobile:not(.isIOS) {
-  --bottom-nav-safe-area: var(--safe-area-bottom);
-}
-
 // iOS-specific adjustments
 body.isIOS {
   // Prevent overscroll bounce effect
--- a/src/styles/_css-variables.scss
+++ b/src/styles/_css-variables.scss
@@ -65,18 +65,25 @@
   --safe-area-right: var(--safe-area-inset-right, env(safe-area-inset-right, 0px));
 
   // Bottom clearance reserved by the mobile bottom nav and everything stacked
-  // above it (scroll content, snackbars). Half the inset by default: iOS's home
-  // indicator is thin and translucent, so a full inset reads as an oversized gap.
-  // Native Android's system bars (3-button / gesture) are opaque and must be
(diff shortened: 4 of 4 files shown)


johannesjo (commented):
## Review notes

CI is green but covers none of this, so these are all read-and-reason findings. Three
independent changes in one PR — my read is that one is worth landing and two aren't, at
least not here.

---

### 1. Full bottom safe area — right direction, wrong stated reason

The blast radius is genuinely small. `--bottom-nav-safe-area` has exactly five consumers,
all reading the same token:

- `mobile-bottom-nav.component.scss:14,23` — bar height and `padding-bottom`
- `app.component.scss:83` — `.main-content` padding
- `_overwrite-material.scss:508,516` — snack-bar offsets

Android is unaffected (it already had the full inset since #8792), desktop and web are
unaffected (`env()` resolves to 0). **The entire functional delta is +17px on iOS**, and
it's a one-line revert. Good risk profile.

Two problems with the justification, though.

**#5563 was already fixed, and is closed.** The report describes the bar rendering *under*
the pill — i.e. no inset reserved at all — on 16.3. That was fixed by `4fe779e3`
(`fix(ui): fix bottom nav overlap on iOS by adding safe-area padding`, 2025-11-25), which
shipped in **v16.4.0** — the very next release, nine months ago. The reporter never came
back to confirm, @lanesawyer asked again in July and got no reply, and the issue was closed
as `completed` today. So this PR isn't fixing #5563; it's proposing a further adjustment on
top of the fix that already closed it. The motivation needs to stand on its own terms.

**"The icons and the active-item pill still land inside the home-indicator band" doesn't
hold up.** At half of a 34pt inset, the button row's bottom edge sits 17px above the screen
edge. The drawn home indicator occupies roughly the bottom 8–13pt, so the pill clears
it — narrowly, but it clears it.

The argument that *does* work is the one the PR doesn't make: iOS reserves a **gesture
region** at the bottom edge that extends well above the drawn glyph, and taps landing in it
get swallowed or need a second tap. That's a real reason to reserve the full inset. But it's
a different claim, it needs a device to confirm, and it's what the comment in
`_css-variables.scss` should say — otherwise the next person reads a rationale that is
checkably false and reverts the change.

The preview is deployed and already linked on #5563. One screenshot from anyone with a
notched iPhone settles this either way.

---

### 2. Dock the add-task FAB — I'd drop this

The framing is off: **the FAB already overhangs.** The bar's content box is 43px, not 44px
(`height: calc(44px + SA)` minus the 1px `border-top` and the `SA` `padding-bottom`, with
`box-sizing: border-box`). A 56px FAB with `margin-bottom: var(--s3)` centres to a margin-box
offset of −18.5px, so it stands ~17.5px proud of the bar today. This change takes that to
~27.5px — it widens the overhang, it doesn't create it. Worth correcting in the description,
since "sat *in* the icon row" is what motivates the change.

The centring identity itself is correct and neat — FAB centre = H/2 − M/2, independent of FAB
size — but it holds **only because `padding-bottom` exactly equals `--bottom-nav-safe-area`**.
That's the coupling part 1 of this PR is changing, so it's worth naming in the comment.

The cost that isn't mentioned: `.smaller` (applied on `hasProjectBacklog()`,
`mobile-bottom-nav.component.html:30`) keeps `margin-bottom: 0` **and** `var(--s)` gutters.
So moving between a project with a backlog and one without now changes the FAB's size
(56↔44), horizontal gutter (16↔8) **and** vertical position (~27.5px overhang ↔ flush). The
jump exists today at ~17.5px; this widens it by 10px and adds a gutter mismatch on top. If
the dock is worth having, `.smaller` should dock too — otherwise they're two different FABs.

And there's no reported problem behind it. It's a taste change to the most-tapped button in
the mobile UI, with a self-acknowledged tap-interception cost over scrolling content.

---

### 3. Onboarding hint — separate PR, and it doesn't clear what it claims to

Unrelated to mobile safe areas or the FAB. Bundled like this, neither can be reverted without
the other.

On the merits, moving an un-anchored hint out of viewport centre is defensible — centred with
no backdrop does read as a broken modal. But the new position doesn't do what the description
says.

**`.floating` is not a mobile state.** `explore` is the only step whose selector can return
`null`, and it returns null only when `isMobile` is *false*
(`onboarding-hint.component.ts:88`), where `isMobile = isTouchActive() &&
isShowMobileBottomNav()`. On a phone both are true before the hint runs — `GlobalThemeService`
injects `InputIntentService` eagerly (`global-theme.service.ts:203`), so `_inputIntentSignal`
is already `'touch'`. So on an actual phone the explore hint is **anchored to a task**, never
floating. Floating is desktop and tablets.

Which means the new `@include mq(xs, max) { :host-context(body.hasMobileBottomNav) … }` block
is reachable only for a **non-touch viewport under 600px** — effectively a resized desktop
window, since `isShowMobileBottomNav()` is just `isXs()`. Not dead code, and mirroring the
snack-bar pattern is the right instinct, but it guards a far narrower case than
"clear of … the mobile bottom nav" implies.

**What it actually needs to clear on desktop is the vertical action bar — and it doesn't.**
`.floating` now sits at `inset-inline-end: var(--s2)`, 16px from the right edge, on desktop.
That's exactly where `nav.action-nav-right--teleported` lives when `misc.isVerticalActionBar`
is enabled: `position: fixed; inset: 0 0 0 auto; width: var(--vertical-action-bar-width);
z-index: var(--z-vertical-action-bar)` (`_vertical-action-bar.scss:17`). The hint is
`--z-onboarding-hint: 1100` vs the strip's `70`, so the chip paints over the rail. The strip
is `justify-content: flex-start`, so no button gets covered — but the chip will visibly
straddle the rail and its `border-left`. `.app-container.has-vertical-action-bar` already
reserves `--vertical-action-bar-width` for precisely this; the hint should too. Same story,
less severely, for the right panel (`--z-right-panel: 44`).

Two smaller notes:
- Dropping `hint-appear-floating` is safe — the base `hint-appear` only translates Y, and
  `.floating` now sets `transform: none`, so no conflict. Confirmed the template passes `null`
  for `[style.top.px]`/`[style.left.px]` when floating, so no inline style fights the CSS.
- `_calculatePosition` carries `// Must match max-width in onboarding-hint.component.scss`
  against `hintWidth = 260`. Still correct — floating never reaches that path — but the
  `.floating` override now makes two max-widths in one file, so that comment is a step closer
  to being misread.

---

### Suggested split

- **Land 1 on its own**, with the comment rewritten around the gesture region rather than the
  glyph, once someone has checked the preview on a notched iPhone.
- **Drop 2**, or make `.smaller` dock consistently if it stays.
- **Move 3 to its own PR** and handle `--vertical-action-bar-width` there.

PR #9617, opened 2026-08-15
feat(plugins): add task context menu actions

### LINKED ISSUE #9616: Feature request: allow plugins to add actions to task context menus
### Problem Statement

Plugins can currently register global UI entry points such as header buttons, general menu entries, side-panel buttons, and keyboard shortcuts. However, they cannot add an action next to the task it operates on.

For example, a task-color plugin may want to provide:

Right-click task → Set background color → choose a color

The current workaround is to add a permanent header button or manipulate the application's DOM and Angular Material menu after it is rendered. The DOM approach is fragile across UI changes, difficult to make accessible, and not suitable for other plugins.

The task context menu already exists in:
- `src/app/features/tasks/task/task.component.ts`
- `src/app/features/tasks/task-context-menu/task-context-menu-inner/task-context-menu-inner.component.*`

### Possible Solution

Please expose a supported Plugin API for adding plugin actions to a task's right-click context menu.

For example:

```ts
PluginAPI.registerTaskContextMenuEntry({
  id: 'set-task-color',
  label: 'Set background color',
  icon: 'format_color_fill',
  order?: number,
  isVisible?: (task) => boolean,
  onClick: (task) => void | Promise<void>,
});
```

Possible alternative naming: `registerTaskMenuEntry(...)`.

The callback should receive a read-only task snapshot or task ID. The plugin can then use the existing task API, for example:

```ts
await PluginAPI.updateTask(task.id, { notes: updatedNotes });
```

Requirements and considerations:
- Support main tasks and subtasks.
- Work in the existing task context menu and preserve normal menu behavior.
- Allow conditional visibility based on the task.
- Define ordering or a plugin submenu for multiple entries.
- Clean up registrations when a plugin is disabled, reloaded, or uninstalled.
- Keep plugin permissions and user consent visible; registering a menu item must not grant additional data permissions.
- Avoid adding subscriptions or expensive callbacks to every rendered task row.
- Ideally provide the same extension point for task menus in the list, planner, and schedule views where applicable.

- Header button: always visible and not contextual to the selected task.
- Keyboard shortcut: less discoverable and requires task focus.
- Plugin DOM/MutationObserver workaround: fragile, inaccessible, and coupled to internal Angular/Material DOM structure.
- Opening a separate plugin panel or dialog: adds unnecessary navigation and loses the context of the task that was right-clicked.

### Additional context

The current Plugin API types expose registerHeaderButton, registerMenuEntry, registerSidePanelButton, and registerShortcut, but no task-context-menu registration method:
https://github.com/super-productivity/super-productivity/blob/master/packages/plugin-api/src/types.ts

The current task menu implementation is:
https://github.com/super-productivity/super-productivity/blob/master/src/app/features/tasks/task-context-menu/task-context-menu-inner/task-context-menu-inner.component.html

This would make plugin actions more discoverable, reduce UI hacks, and provide a reusable extension point for task-focused plugins.

Zombiehamser:
## Summary

- add `PluginAPI.registerTaskContextMenuEntry()` with the `taskContextMenu` permission
- show plugin actions in a dedicated, attributed task-context-menu submenu
- filter actions for tasks and subtasks declaratively and clean registrations on unload or failed activation
- document the host-side-only API and reject iframe registrations explicitly

## Why

Provides a supported alternative to DOM mutation for plugin task actions.

Closes #9616

## Validation

- focused unit tests: 91 passing
- `plugin-api` TypeScript build
- production Angular frontend build
- Prettier and `git diff --check`
- independent GPT-5.6 Sol architecture and final review

The repository-wide pre-commit/pre-push hook is Windows-path-sensitive and fails three existing `tools/check-css-vars.test.js` assertions that expect Unix separators; the feature-specific checks above pass.


johannesjo:
Thanks for the PR and the detailed issue write-up! The implementation quality is high: the menu component is lazily instantiated so there is no per-task-row cost, entries resolve at click time so unloaded plugins no-op, validation and lifecycle tests are thorough, and choosing a declarative `showFor` over the originally proposed `isVisible(task)` callback was the right call.

That said, I'd like the integration reworked before merge, plus a few smaller points:

### 1. Route registration through `PluginBridgeService` instead of a parallel path

All existing UI registrations (`registerMenuEntry`, `registerHeaderButton`, `registerSidePanelButton`, `registerShortcut`) flow through `createBoundMethods()` and are centrally cleaned up in `PluginBridgeService.unregisterPluginHooks()`, which already delegates to per-feature registries (issue-provider, sync-adapter). This PR instead wires the new registry directly into `PluginAPI` via a new constructor param, introduces an `isActive` lifecycle closure, and calls `unregisterPlugin()` at four separate places in `plugin-runner.ts`.

Keeping the small standalone registry service is fine (the bridge is already over the service size cap), but please wire it like the other registries:

- expose `registerTaskContextMenuEntry` from `createBoundMethods()` — it already receives the manifest, so a permission check (if kept, see below) can live there the same way the `http` check does in `request()`;
- remove the plugin's entries in `unregisterPluginHooks()`.

That deletes the `plugin-runner.ts` changes, the extra constructor param, and the `isActive` concept entirely — unload cleanup comes for free from the existing path. Two related notes:

- The stale-instance guard as written only protects this one API; a stale `registerMenuEntry` on a kept-around API object still works. One-off hardening like this makes the codebase harder to reason about than a consistent (even if imperfect) rule.
- Cleanup-on-failed-activation is a real gap, but it exists for *all* registration types today, not just this one. If you want to fix it, calling `unregisterPluginHooks()` in the activation-failure path fixes it uniformly instead of only for the new registry.

### 2. Does the `taskContextMenu` permission earn its place?

No other UI registration requires a manifest permission, and the callback only receives a task ID — no data access is granted (reading or updating the task still goes through the existing task APIs and their permissions). As proposed the permission protects nothing while adding permanent manifest and API surface. I'd drop it unless there is a concrete threat it addresses.

### 3. Smaller points

- `plugin-task-context-menu-registry.service.ts`: validation errors are raw English strings, while the analogous `_registerMenuEntry` validation uses translated `T.PLUGINS.*` keys.
- `execute()`'s `catch {` logs a message but discards the caught error object — please include it in the log call for debuggability.
- `t.const.ts`: the reorder of `LOCAL_REST_API_TOKEN_ERROR` / `..._REGENERATE` is unrelated churn; please drop that hunk.

One general note: plugin API surface is effectively permanent once shipped (config shape, replace-on-same-id semantics, attribution format), so I want the shape settled before merging rather than after.

Zombiehamser:
Thanks for the detailed review! I’ve addressed all requested changes in commit [a414174c](https://github.com/Zombiehamser/super-productivity/commit/a414174c8519c2e5968d2a414520283946e19292).
- Routed task context menu registration through PluginBridgeService.
- Removed the unnecessary taskContextMenu manifest permission.
- Added translated validation errors and preserved callback errors in logs.
- Centralized cleanup, including activation-failure cleanup.
- Removed the unrelated translation-constant reorder.
- Added bridge, lifecycle, and validation tests.
- Updated the plugin API documentation and types.

Targeted tests, Plugin API build, formatting, and lint checks pass. Could you please take another look?

johannesjo:

Thanks for the rework – I checked each of the six points from the earlier review against `a414174c` and they are all genuinely addressed: registration now flows through `createBoundMethods()`, cleanup rides `unregisterPluginHooks()`, the `taskContextMenu` permission is gone, validation errors are translated, `execute()`'s catch keeps the error object, and the `t.const.ts` reorder is dropped. The remaining items are small.

One thing worth keeping: I checked whether the `plugin-runner.ts` activation-failure hunk is still needed now that `activatePlugin()` already calls `unloadPlugin()` in its catch – it is. The ZIP-upload path (`plugin.service.ts:1399-1451`) sets `status: 'error'` without any teardown, so your hunk closes a gap that path still has.

### Blocking

- **The branch conflicts with `master`.** It is 215 commits behind and GitHub reports `mergeable: CONFLICTING`, so it can't be merged as-is. `plugin-bridge.service.ts`, `plugin-api.ts`, `t.const.ts` and `en.json` have all moved since. A rebase onto current master is needed before this can go in – and it will also let CI actually run, which it never has on this PR.

### Nits

- **`plugin-task-context-menu-registry.service.ts:101`** – `entriesFor()` sorts on `a.pluginId.localeCompare(b.pluginId) || a.id.localeCompare(b.id)`, so entries from the *same* plugin come out alphabetised by id rather than in registration order. A plugin registering `set-color` then `clear-color` sees `clear-color` first, and the only way to control order is to rename ids. The three sibling registries all just append and keep registration order (`plugin-bridge.service.ts:1487`, `:1540`, `:1677`). Dropping `|| a.id.localeCompare(b.id)` gets both properties – `Array.prototype.sort` is stable, so sorting on `pluginId` alone still gives cross-plugin determinism while preserving per-plugin registration order.

- **`plugin-task-context-menu-registry.service.ts:75`** – `cfg.showFor?.length ? ... : DEFAULT_TARGETS` treats an explicit `showFor: []` exactly like omitting the field, so it silently means "show on both targets". `docs/plugin-development.md` says the default applies "when omitted", so the two disagree for `[]`. Since this is permanent once shipped, worth pinning down: branch on `cfg.showFor === undefined` for the default and let `[]` either throw `TASK_CONTEXT_MENU_ENTRY_TARGET_INVALID` or mean "nowhere" – then say which in the `PluginTaskContextMenuEntryCfg` doc comment.

- **`plugin-task-context-menu-registry.service.ts:12`** – `ICON_PATTERN` (`/^[a-z0-9_]{1,64}$/`) is stricter than `registerMenuEntry`, which accepts any string, and the restriction isn't documented anywhere. A plugin that ships its own `icon.svg` and passes `icon: 'plugin-my-plugin-icon'` (which works for its side-nav entry, see `magic-nav-config.service.ts:458-467`) just gets `TASK_CONTEXT_MENU_ENTRY_ICON_INVALID` with no hint why. The template renders `<mat-icon>{{ entry.icon }}</mat-icon>`, so an SVG icon wouldn't paint here anyway – documenting "Material icon (ligature) names only" on `PluginTaskContextMenuEntryCfg.icon` and in `docs/plugin-development.md` is enough.

### Questions

- `plugin-runner.ts:137` – the second `unregisterPluginHooks()` sits in the *outer* catch, which only covers the region before `_executePlugin()` runs (API construction, `analyzePluginCode`, the warning snack). As far as I can tell nothing there can register anything yet, so that call looks like it can never have work to do. Is it deliberate belt-and-braces, or can something in that region register on the plugin's behalf? If it's the former the inner-catch call alone would cover the real gap.

---

johannesjo:
**💬 Maintainer note** – one API-shape call I should have answered last pass; no re-review needed for it, just fold it into the rebase.

Drop the per-row plugin-name prefix: render just `{{ entry.label }}`, and remove `pluginName` from `PluginTaskContextMenuEntryView`, `RegisteredTaskContextMenuEntry` and `register()`'s parameters (the bridge then passes only `pluginId` and `cfg`).

Three reasons. The submenu is already namespaced by its own "Plugin actions" trigger, which stays visible while the submenu is open, and plugin entries elsewhere aren't prefixed either — `magic-nav-config.service.ts:473` uses `label: entry.label` for plugin entries mixed straight into the app's own nav list, and `plugin-header-btns.component.ts:13` binds only `[matTooltip]="button.label"`; prefixing the one fenced list while the unfenced ones go bare is backwards. The prefix also routes around the cap you just added: `label` is capped at 80 chars, but the name reaching the row via `manifest?.name ?? pluginId` is only checked for non-emptiness (`validate-manifest.util.ts:35`) — no length or charset limit. And it costs room the row doesn't have: Material caps the menu panel at `max-width: 280px`, and I measured "My Task Sync Helper — Archive to Notion" wrapping to two lines in it (the `white-space: nowrap` at `_overwrite-material.scss:299` loses on specificity to Material's own `white-space: normal` for menu item text), so every entry turns into a double-height row that opens with the same words.

Disambiguation stays with the plugin author, who has 80 characters and can write "Notion: Archive" if they want it; users identify a misbehaving plugin via Settings -> Plugins, as they already do for header buttons and nav entries. Removing the prefix also disposes of the hardcoded em dash, which is neither localisable nor RTL-safe — and moving it to a `T.PLUGINS.*` key would only make a bad row localisably bad.

Zombiehamser:
Rebased onto current master and addressed the remaining review points in 0b4ad3879.

- Rows now render only the plugin-provided label.
- Registration order is stable within each plugin, including replacement of an existing id.
- showFor defaults only when omitted; an empty array is rejected and documented.
- Icon documentation now specifies Material ligature names only.
- Removed the redundant outer activation cleanup; the activation-failure cleanup remains.

Validated with focused registry, bridge, runner, task-context-menu, PluginAPI, and iframe tests; Plugin API build also passes. The repository pre-commit/pre-push hook still fails on the unrelated Windows artifact-contract test (missing nsis section); this branch does not touch its inputs. Could you please re-review?

johannesjo:
There are new conflicts it seems. Could youhave another look @Zombiehamser ?

Zombiehamser:
Thanks for flagging this. Rebased onto master ca2dbce1e and pushed b25af2739. The context-menu conflicts are resolved, preserving the upstream multi-select and move-to-project changes. GitHub now reports MERGEABLE for this head.

Validation: all 20 task-context-menu tests pass, checkFile passes for both conflict-resolved TypeScript files, and the Plugin API build passes. The local pre-push hook reported repository-wide Prettier errors after refreshing dependencies, so it was bypassed; full lint is not claimed as passing.

PR #9618, opened 2026-08-15
fix(tasks): render Later Today subtasks in subTaskIds order (#9614)

### LINKED ISSUE #9614: 💥 Sorting sub-tasks in repeating tasks does not apply on today (macOS app)
### Steps to Reproduce
1. Create a task with repeating
2. Create subtasks (example: 1, 2, 3 <- in that order)
3. Try to re-order them in normal view (it will apply in the database but not in the frontend, when you check the parrent task over the modal it will show you the changed order but wont apply on main screen sub-task order)

Image

Image

### URL
file:///Applications/Super%20Productivity.app/Contents/Resources/app.asar/.tmp/angular-dist/browser/index.html#/tag/TODAY/tasks

### Meta Info
META: SP18.19.0D __ Electron – de – MacIntel – de – UA:Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) superProductivity/18.19.0 Chrome/146.0.7680.216 Electron/41.4.0 Safari/537.36

### Actions Before Error
```
Create a recurring parrent-task with sub-tasks
```

bemany:
Fixes #9614

## Problem

Reordering the sub-tasks of a task that is scheduled for later today (most prominently the daily instance of a repeating task with a time, e.g. "Every day, 22:00") updates the store and the task detail panel, but the "Later Today" panel in the Today view keeps rendering the sub-tasks in their creation order.

Cause: `selectLaterTodayStructure` derived each parent's sub-task order from the iteration order of the scheduling snapshot (entity insertion order, which never changes on reorder) instead of from `parent.subTaskIds`, the single source of truth for manual ordering. For repeating tasks this is especially visible because the sub-task instances are recreated from templates every day, so any reordering done during the day never shows up in the panel.

## Fix

Use the parent's `subTaskIds` from the scheduling snapshot, exactly like `selectAllTasksWithSubTasksStructure` already does. The obsolete `subtaskIdsByParentId` map (built from iteration order) is removed. Downstream behavior is unchanged: the stable with-sub-tasks mapper already drops missing entities, and the earliest-time sort reads times per id, so orphaned/scheduled-sub-task handling and sorting are unaffected (all 29 existing later-today specs still pass).

## Reproduction / Tests

- New unit test in `task-later-today.selectors.spec.ts` that fails without the fix (sub-tasks came back in insertion order instead of `subTaskIds` order).
- New E2E test `e2e/tests/work-view/later-today-subtask-order-9614.spec.ts`: schedules a parent for later today, reorders its sub-tasks, and asserts the panel reflects the new order. Fails on master, passes with the fix (verified both ways locally).
- `npm run test:file` for `task-later-today.selectors.spec.ts` (29 ✓) and `task.selectors.spec.ts` (77 ✓), `checkFile` clean on all touched files.

Note: sync-wise this is a pure read-path/selector change; no actions, reducers, or op-log behavior are touched.

### DIFF
--- a/e2e/tests/work-view/later-today-subtask-order-9614.spec.ts
+++ b/e2e/tests/work-view/later-today-subtask-order-9614.spec.ts
@@ -1,0 +1,106 @@
+import { expect, test } from '../../fixtures/test.fixture';
+
+/**
+ * Issue #9614: Sorting sub-tasks of a repeating task does not apply on Today.
+ *
+ * A repeating task with a time (e.g. "Every day, 22:00") creates a daily
+ * instance scheduled for later today, which is rendered in the "Later Today"
+ * panel of the Today view. Reordering its subtasks updates parent.subTaskIds
+ * (visible in the task detail panel), but the "Later Today" panel kept
+ * rendering the subtasks in store insertion order because
+ * selectLaterTodayStructure derived the subtask order from snapshot iteration
+ * order instead of parent.subTaskIds.
+ *
+ * This test schedules a parent for later today (same selector path as a
+ * repeating instance), reorders its subtasks via the keyboard shortcut
+ * (Ctrl+Shift+ArrowUp, same store action as drag & drop) and asserts the
+ * panel reflects the new order.
+ */
+
+const TWO_HOURS_MS = 2 * 60 * 60 * 1000;
+
+const getFutureTimeToday = (): string => {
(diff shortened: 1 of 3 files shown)


johannesjo:
Thank you very much! <3

PR #9623, opened 2026-08-16
fix(schedule): drive the month grid from the real week count

### LINKED ISSUE #9584: 🚨 Month view: 16 `--nr-of-weeks` selectors never match, so adaptive cell height and event caps never applied
### Describe the bug

`schedule-month.component.scss` contains **16 selectors** of the form `:host[style*='--nr-of-weeks: N']`. None of them has ever matched, so two intended month-view behaviours have never actually shipped.

`[style*=...]` is an attribute selector: it matches against the element's inline `style` **attribute**. But `--nr-of-weeks` is only ever declared inside a stylesheet, as a static rule in `schedule.component.scss:6`:

```scss
:host {
  --nr-of-weeks: 6;
}
```

A custom property declared in a stylesheet does not appear in the element's `style` attribute, so the attribute selector never matches. It is never set as an inline style or via a host binding anywhere: `grep -rn "nr-of-weeks" src/` returns hits only in those two SCSS files, with no TypeScript or template hits at all.

(The `var(--nr-of-weeks)` usage at `schedule-month.component.scss:10` is fine. That is normal custom-property resolution and works as intended. Only the `[style*=...]` selectors are affected.)

### What never applied

**1. Adaptive day-cell height** (`:72-85`, plus the mobile variants at `:91-106`)

Cells were meant to get shorter as the month needs more weeks: `120px` at 3 weeks, `100px` at 4, `85px` at 5, `70px` at 6. Instead every cell always gets the flat `min-height: 80px` from `:66`.

**2. Adaptive event caps** (`:178-191`, plus the mobile variants at `:195-208`)

Each week count was meant to show a different number of events per day before clipping: hide from the 8th at 3 weeks, from the 6th at 4, from the 4th at 5, from the 3rd at 6. Instead the only cap that ever fires is the generic `.month-event:nth-child(n + 4)` at `:212`, which is week-count agnostic.

### Related: the "+N more" badge never shows on desktop

While looking at this: `.month-more-events` is `display: none` at `:227-229` and only re-enabled inside `@include mq(xs, max)` at `:276`. So on desktop, month cells clip events with no indicator at all that anything is hidden.

The template also computes `hiddenEventCount` as `dayEvents.length - 1`, which is only correct for the mobile case where exactly one event is visible. That is presumably why it is hidden on desktop rather than being wrong out loud, but the result is that events silently disappear.

### Suggested fix

These are two different decisions, so possibly two PRs:

1. **Either revive or delete the dead rules.** Reviving them means setting the property inline, e.g. a `[style.--nr-of-weeks]` host binding driven by the real week count. That is worth doing deliberately, because it changes the month view for every user at once (cell heights and how many events show per day). Deleting them keeps today's behaviour and just removes 16 rules that claim to do something they don't.
2. **Decide what desktop should do when a day has more events than fit.** Either show a correct "+N more" (which needs the count to reflect how many are actually visible, not `length - 1`), or make the cell scroll, or leave it as is deliberately.

### Additional context

Found while reviewing #9553, which removed the generic `nth-child(n + 4)` cap at `:212` on the assumption it was the bug. It was actually the only cap doing anything, so removing it uncapped the month view entirely. That PR is closed, but the underlying issue it half-noticed is real.

Not addressed by #9463, which changes `WEEKS_SHOWN` and scroll behaviour in the planner calendar nav and does not touch `--nr-of-weeks`.

costajohnt:
Fixes #9584

Both halves in one commit, so the binding never activates the dead rules.

Thanks for the runtime measurements — and for the correction. You're right that `weeksToShow` is derived from `window.innerHeight`, so 3 weeks is the *shortest* window, not the most crowded month. I had that backwards, and it does make the delete case stronger.

## 1. Delete the 16 dead selectors

`--nr-of-weeks` was only ever declared inside a stylesheet, so it never reached the host's `style` attribute and none of the selectors matched. The adaptive cell heights and per-week-count event caps they describe have never shipped.

Not revived, per your measurements: every one of them outranks the single `@media (min-height: 800px)` rule that lifts the generic cap, so turning them on can only subtract — 8 events down to 2 at 6 weeks on a tall screen. The generic height-driven cap stays as the one capping system.

## 2. Bind the property to the rendered week count

This is the half that fixes something users actually see. `grid-template-rows: auto repeat(var(--nr-of-weeks), 1fr)` did resolve, but always to the static 6 inherited from `schedule`, while the template only renders cells for `weekIndex < weeksToShow()`. At 5 weeks that declared six rows for five weeks of cells, sized every row 1/6, and left a blank strip at the bottom — and 1366x768 lands squarely in the 5-week bucket.

Bound on `ScheduleMonthComponent` rather than `ScheduleComponent`, since `:host` in that stylesheet is `<schedule-month>`.

One wrinkle worth flagging: written through the `HOST_BINDINGS` const pattern from `progress-circle.component.ts`, because a literal `'[style.--nr-of-weeks]'` key trips `@typescript-eslint/naming-convention`. `ScheduleComponent`'s existing `--nr-of-days` binding only gets away with the literal because that file is `/* eslint-disable */`. Happy to switch to an inline disable comment if you'd prefer the shape to match `ScheduleComponent` exactly.

## Verification

Three specs cover the binding, and all three fail without it (`Expected '' to be '5'`, `Expected '' to be '3'`/`'6'` for the change-tracking case, and the grid row count). 36/36 in the month spec, 64/64 in the schedule spec, `npm run checkFile` clean on all three changed files.

I left `--nr-of-weeks: 6` on `schedule` in place as a fallback rather than widening the diff. Say the word if you'd rather it go.

**Not in scope, per your comment:** the "+N more" badge, which stays a separate PR.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/schedule/schedule-month/schedule-month.component.scss
+++ b/src/app/features/schedule/schedule-month/schedule-month.component.scss
@@ -69,40 +69,8 @@
   max-width: 100%;
   overflow: hidden;
 
-  :host[style*='--nr-of-weeks: 3'] & {
-    min-height: 120px;
-  }
-
-  :host[style*='--nr-of-weeks: 4'] & {
-    min-height: 100px;
-  }
-
-  :host[style*='--nr-of-weeks: 5'] & {
-    min-height: 85px;
-  }
-
-  :host[style*='--nr-of-weeks: 6'] & {
-    min-height: 70px;
-  }
-
   @include mq(xs, max) {
     min-height: 50px;
-
-    :host[style*='--nr-of-weeks: 3'] & {
-      min-height: 80px;
-    }
-
-    :host[style*='--nr-of-weeks: 4'] & {
-      min-height: 65px;
-    }
-
-    :host[style*='--nr-of-weeks: 5'] & {
-      min-height: 55px;
-    }
-
-    :host[style*='--nr-of-weeks: 6'] & {
-      min-height: 50px;
-    }
   }
 
   &.other-month {
@@ -173,40 +141,6 @@
   @include mq(xs, max) {
     padding: var(--s-quarter);
     max-height: calc(100% - 25px);
-  }
-
-  :host[style*='--nr-of-weeks: 3'] & .month-event:nth-child(n + 8) {
-    display: none;
-  }
-
-  :host[style*='--nr-of-weeks: 4'] & .month-event:nth-child(n + 6) {
-    display: none;
-  }
-
-  :host[style*='--nr-of-weeks: 5'] & .month-event:nth-child(n + 4) {
-    display: none;
-  }
-
-  :host[style*='--nr-of-weeks: 6'] & .month-event:nth-child(n + 3) {
-    display: none;
-  }
-
-  @include mq(xs, max) {
-    :host[style*='--nr-of-weeks: 3'] & .month-event:nth-child(n + 4) {
-      display: none;
-    }
-
-    :host[style*='--nr-of-weeks: 4'] & .month-event:nth-child(n + 3) {
-      display: none;
-    }
-
-    :host[style*='--nr-of-weeks: 5'] & .month-event:nth-child(n + 2) {
-      display: none;
-    }
-
-    :host[style*='--nr-of-weeks: 6'] & .month-event:nth-child(n + 2) {
-      display: none;
-    }
   }
 
   .month-event:nth-child(n + 4) {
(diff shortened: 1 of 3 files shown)


johannesjo:
Thank you very much! <3

Issue #9518, opened 2026-08-08
💡 Widget on mobile (android)


cyberborg773:
### Problem Statement

I have a few questions/suggestions regarding the Android widget:

French translation: The widget is currently displayed as “Today”, even though my app/phone is set to French. It would be great if it could be translated to “Aujourd’hui”.

📁 Folder selection: Is there currently a way to choose which folder/project the widget displays? Or is it permanently set to “Today”? I’m looking for something similar to the Microsoft To Do Android widget, where you can choose what list/folder to display.

🪟 Transparency: It would be awesome to have an option to adjust the widget’s transparency/opacity.

🔄 Widget refresh: I noticed that when I create or modify a task from the desktop or web version, the Android widget doesn’t update automatically. I have to tap the widget to refresh it or open the app before the changes appear. Is this expected behavior, or is there a way to enable automatic updates?

I’m currently using Super Productivity 18.19.0 on Android 16 / One UI 8.5 (Samsung).

Thanks again for the great work, and keep it up! 🙌

### Possible Solution

You are the best.

ReFil:
There is a noted disconect between the items checked off in the widget box and ones that are checked off when the app is opened

Saman-K:
One thing I noticed using the widget: it doesn't remove tasks once they're checked off. I use the widget in a narrow layout where only ~4 tasks show, and completed ones just stay in the list, taking up a slot instead of making room for the next task. 

Also, a small refresh button somewhere in the widget (e.g. a corner icon) would be really useful, specially if it can sync with WebDAV or what ever syncing setup will be available in the future. 

and it would be nice to point a widget at a specific project. #9624.

Issue #9098, opened 2026-07-17
💥 Android widget does not update for a new day


bsunderhus:
### Steps to Reproduce

1. Create the widget for  today's task
2. Add a task for tomorrow 
3. When tomorrow comes the expected behavior is for the widget to be updated. It isn't. 

Here's some OS info:
1. Samsung galaxy S25
2. Android 16
3. One UI 8.5

### URL
https://localhost/#/tag/TODAY/tasks

### Meta Info
META: SP18.14.0_L2A __ Browser – en-US – Linux aarch64 – en-US – UA:Mozilla/5.0 (Linux; Android 16; SM-S931B Build/BP4A.251205.006; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/150.0.7871.46 Mobile Safari/537.36

### Actions Before Error
```

```

PR #9624, opened 2026-08-16
feat(android): add per-widget project selection #9518

dennnyb:
## Summary
- add per-widget Today or active-project selection for the Android task widget
- keep widget selections isolated by app widget ID, including cleanup and deleted-project fallback
- project snapshots include regular tasks followed by backlog tasks, capped at 20 rows
- When project selected in widget is NOT today - the default one, tasks marked as completed disappears from the widget in 5 seconds after the completion. They can be accessed as "completed" from the app.

## Validation
- npm run checkFile for changed TypeScript files
- focused Android widget selector tests
- focused native WidgetData JVM test
- Android F-Droid debug build and emulator install
- repository pre-push lint and test suites
- manual verification in Pixel 10a AndroidStudio emulator.

Partially deals with #9518

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -190,6 +190,15 @@
         android:resource="@xml/appwidget_info" />
     </receiver>
 
+    <activity
+      android:name=".widget.TaskListWidgetConfigureActivity"
+      android:exported="true"
+      android:theme="@style/AppTheme.NoActionBar">
+      <intent-filter>
+        <action android:name="android.appwidget.action.APPWIDGET_CONFIGURE" />
+      </intent-filter>
+    </activity>
+
     <service
       android:name=".widget.TaskListWidgetService"
       android:exported="false"
--- a/android/app/src/main/java/com/superproductivity/superproductivity/CapacitorMainActivity.kt
+++ b/android/app/src/main/java/com/superproductivity/superproductivity/CapacitorMainActivity.kt
@@ -65,6 +65,7 @@ class CapacitorMainActivity : BridgeActivity() {
     private var isTimerCompleteReceiverRegistered = false
     private var isForegroundServiceFailureReceiverRegistered = false
     private var isWidgetDoneDrainReceiverRegistered = false
+    private var isWidgetProjectOpenDrainReceiverRegistered = false
 
     private val storageHelper =
         SimpleStorageHelper(this) // for scoped storage permission management on Android 10+
@@ -86,6 +87,16 @@ class CapacitorMainActivity : BridgeActivity() {
                 // getWidgetDoneQueue(), so there is a single delivery path and no
                 // task data crosses the string-interpolated JS bridge.
                 callJSInterfaceFunctionIfExists("next", "onWidgetDoneDrainRequest$")
+            }
+        }
+    }
+
+    private val widgetProjectOpenDrainReceiver = object : BroadcastReceiver() {
+        override fun onReceive(context: Context?, intent: Intent?) {
+            if (intent?.action == TaskListWidgetProvider.ACTION_WIDGET_PROJECT_OPEN_DRAIN) {
+                // The project ID remains in the native queue so cold and warm starts
+                // use the same pull-based Angular delivery path.
+                callJSInterfaceFunctionIfExists("next", "onWidgetProjectOpenDrainRequest$")
             }
         }
     }
@@ -242,6 +253,11 @@ class CapacitorMainActivity : BridgeActivity() {
             IntentFilter(TaskListWidgetProvider.ACTION_WIDGET_DONE_DRAIN)
         )
         isWidgetDoneDrainReceiverRegistered = true
+        LocalBroadcastManager.getInstance(this).registerReceiver(
+            widgetProjectOpenDrainReceiver,
+            IntentFilter(TaskListWidgetProvider.ACTION_WIDGET_PROJECT_OPEN_DRAIN)
+        )
+        isWidgetProjectOpenDrainReceiverRegistered = true
 
         // Show startup overlay for quick task entry while Angular loads.
         // Only on fresh cold start — not on config-change recreation.
@@ -634,6 +650,12 @@ class CapacitorMainActivity : BridgeActivity() {
         if (isWidgetDoneDrainReceiverRegistered) {
             LocalBroadcastManager.getInstance(this).unregisterReceiver(widgetDoneDrainReceiver)
             isWidgetDoneDrainReceiverRegistered = false
+        }
+        if (isWidgetProjectOpenDrainReceiverRegistered) {
+            LocalBroadcastManager.getInstance(this).unregisterReceiver(
+                widgetProjectOpenDrainReceiver
+            )
+            isWidgetProjectOpenDrainReceiverRegistered = false
         }
         super.onDestroy()
     }
--- a/android/app/src/main/java/com/superproductivity/superproductivity/webview/JavaScriptInterface.kt
+++ b/android/app/src/main/java/com/superproductivity/superproductivity/webview/JavaScriptInterface.kt
@@ -392,6 +392,13 @@ class JavaScriptInterface(
         return WidgetDoneQueue.getAndClear(activity)
     }
 
+    /** Get the project selected by the most recent widget row tap and clear it. */
+    @Suppress("unused")
+    @JavascriptInterface
+    fun getWidgetProjectToOpen(): String? {
+        return TaskListWidgetProvider.getAndClearProjectToOpen(activity)
+    }
+
     /**
      * Re-render the home screen widget from the current `widget_data` KeyValStore
      * snapshot. Called by Angular after each snapshot push.
--- a/android/app/src/main/java/com/superproductivity/superproductivity/widget/TaskListWidgetConfigureActivity.kt
+++ b/android/app/src/main/java/com/superproductivity/superproductivity/widget/TaskListWidgetConfigureActivity.kt
@@ -1,0 +1,64 @@
+package com.superproductivity.superproductivity.widget
+
+import android.app.AlertDialog
+import android.appwidget.AppWidgetManager
+import android.content.ComponentName
+import android.content.Intent
+import android.os.Bundle
+import androidx.appcompat.app.AppCompatActivity
+import com.superproductivity.superproductivity.App
+import com.superproductivity.superproductivity.R
+
+/** Minimal per-instance source picker for the Android home-screen task widget. */
+class TaskListWidgetConfigureActivity : AppCompatActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setResult(RESULT_CANCELED)
+
+        val appWidgetId = intent.getIntExtra(
+            AppWidgetManager.EXTRA_APPWIDGET_ID,
+            AppWidgetManager.INVALID_APPWIDGET_ID
+        )
+        val appWidgetManager = AppWidgetManager.getInstance(this)
+        if (
+            appWidgetId == AppWidgetManager.INVALID_APPWIDGET_ID ||
+            appWidgetManager.getAppWidgetInfo(appWidgetId)?.provider !=
+                ComponentName(this, TaskListWidgetProvider::class.java)
+        ) {
+            finish()
+            return
+        }
+
+        val projects = try {
+            WidgetData.parseProjects(
+                (applicationContext as App).keyValStore.get(WidgetData.KEYVAL_KEY, "{}")
+            )
+        } catch (_: Exception) {
+            emptyList()
+        }
+        val selectedProjectId = TaskListWidgetProvider.selectedProjectId(this, appWidgetId)
+        val labels = listOf(getString(R.string.widget_header_title)) + projects.map { it.title }
+        val checkedItem = projects.indexOfFirst { it.id == selectedProjectId }.let {
+            if (it < 0) 0 else it + 1
+        }
+
+        AlertDialog.Builder(this)
+            .setTitle(R.string.widget_configure_title)
+            .setSingleChoiceItems(labels.toTypedArray(), checkedItem) { _, which ->
+                TaskListWidgetProvider.setSelectedProjectId(
+                    this,
+                    appWidgetId,
+                    projects.getOrNull(which - 1)?.id
+                )
+                TaskListWidgetProvider.refresh(this, appWidgetId)
+                setResult(
+                    RESULT_OK,
+                    Intent().putExtra(AppWidgetManager.EXTRA_APPWIDGET_ID, appWidgetId)
+                )
+                finish()
+            }
+            .setOnCancelListener { finish() }
+            .show()
+    }
+}
--- a/android/app/src/main/java/com/superproductivity/superproductivity/widget/TaskListWidgetProvider.kt
+++ b/android/app/src/main/java/com/superproductivity/superproductivity/widget/TaskListWidgetProvider.kt
@@ -7,6 +7,8 @@ import android.content.ComponentName
 import android.content.Context
 import android.content.Intent
 import android.net.Uri
+import android.os.Handler
+import android.os.Looper
 import android.text.format.DateUtils
 import android.util.Log
 import android.widget.RemoteViews
@@ -173,6 +175,28 @@ class TaskListWidgetProvider : AppWidgetProvider() {
             updateAll(context, AppWidgetManager.getInstance(context), intArrayOf(appWidgetId))
         }
 
+        @Synchronized
+        fun scheduleProjectTaskExpiryRefresh(context: Context, refreshAtMs: Long) {
+            if (scheduledProjectTaskRefreshAt?.let { it <= refreshAtMs } == true) {
+                return
+            }
+            scheduledProjectTaskRefresh?.let { refreshHandler.removeCallbacks(it) }
+            val appContext = context.applicationContext
+            val refreshRunnable = Runnable {
+                synchronized(this) {
+                    scheduledProjectTaskRefresh = null
+                    scheduledProjectTaskRefreshAt = null
+                }
+                refreshAll(appContext)
+            }
+            scheduledProjectTaskRefresh = refreshRunnable
+            scheduledProjectTaskRefreshAt = refreshAtMs
+            refreshHandler.postDelayed(
+                refreshRunnable,
+                (refreshAtMs - System.currentTimeMillis()).coerceAtLeast(0L)
+            )
+        }
+
         /** Rebuilds every passed widget using that instance's selected source. */
         private fun updateAll(
             context: Context,
@@ -248,5 +272,8 @@ class TaskListWidgetProvider : AppWidgetProvider() {
         private fun selectionKey(appWidgetId: Int): String = "project_$appWidgetId"
 
         private const val PREFERENCES_NAME = "task_list_widget"
+        private val refreshHandler = Handler(Looper.getMainLooper())
+        private var scheduledProjectTaskRefresh: Runnable? = null
+        private var scheduledProjectTaskRefreshAt: Long? = null
     }
 }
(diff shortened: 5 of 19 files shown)


johannesjo (commented):
Thanks for this, and for the thorough validation notes — the pure-logic-extracted-and-tested split on both the Kotlin and Angular sides, the `onDeleted` pref cleanup, and the deleted-project fallback to Today are all exactly right. A few things I'd like resolved before this can go in, plus one scope question.

## Scope question: does the 5-second done-task expiry earn its place?

#9518 asked to choose which project the widget shows. The disappearing-completed-tasks behaviour wasn't requested, and it's where nearly all of this PR's complexity and risk live: `doneOn` on the native wire contract, a second SharedPreferences timestamp map (`WidgetDoneQueue.peekDoneTimestamps`), a process-static `Handler` plus scheduling/dedup state in `TaskListWidgetProvider`'s companion object, and the ordering bug in issue 2 below.

It also creates a third, inconsistent behaviour for completed tasks: the Today widget shows them, the app's own project list shows them, project widgets hide them after 5s. And it's non-deterministic — `scheduleProjectTaskExpiryRefresh` relies on the app process staying alive for 5s after `onDataSetChanged` returns from a binder thread. If the process is killed inside that window (routine on Android), the completed row stays on screen until the next 30-minute `updatePeriodMillis` tick.

Concrete alternative: drop done tasks in the **selector's** project loop instead. That removes `PROJECT_DONE_TASK_GRACE_MS`, `pendingDoneTimestamps`, `peekDoneTimestamps`, `scheduleProjectTaskExpiryRefresh`, `doneOn` on the wire, and issue 2 below — roughly 120 lines — and gives a rule that fits in one sentence. The pending-done overlay (`WidgetDoneQueue.peek`) already provides the checkbox feedback; the row vanishing on the next push is the same feedback the grace period buys.

## Blocking

**1. `TaskListWidgetProvider.kt` — the header tap loses its direct-activity PendingIntent.**

The header/empty tap moves from `PendingIntent.getActivity` to `getBroadcast`, so it now routes through `onReceive` → `context.startActivity`. The comment removed in this PR documented exactly why that mattered:

> Background-activity-launch restrictions may block this on some API levels/OEMs; the header tap (direct activity PendingIntent) remains as the reliable way in.

After this change no direct activity PendingIntent remains anywhere in the widget (both remaining ones are `getBroadcast`), so the documented fallback is gone for **every** widget — including Today-only widgets, which gain nothing from the change. The failure mode is "tapping the widget does nothing" on affected devices.

Suggested fix: keep `PendingIntent.getActivity` for the header, put `EXTRA_OPEN_PROJECT_ID` on the `CapacitorMainActivity` intent (keeping the per-widget `data` Uri so `FLAG_UPDATE_CURRENT` still differentiates instances), and have `CapacitorMainActivity` call `queueProjectToOpen` + send the drain broadcast from `onCreate`/`onNewIntent`. Rows have to stay on the broadcast template since it's shared with the done toggle, but the header doesn't have to follow them.

**2. `android-widget.selectors.ts` — the 20-row cap is applied before native drops completed tasks, so a project widget can render empty.**

The selector fills its 20 slots from `project.taskIds` including done tasks, and native then removes any whose grace period has expired. Verified against the real selector: a project whose first 20 `taskIds` entries are completed emits 20 tasks, **all** done, zero undone — and your own `completedProjectTaskDisappearsAfterFiveSecondGracePeriod` test shows `WidgetData.parse` then drops all of them. Result: an empty widget while the project still has open tasks below the cap.

This isn't a corner case. Done tasks stay in `project.taskIds` indefinitely — nothing reorders the list on completion, and `moveToArchive` only runs on explicit user action (the work-view "move done to archive" button for the active context, or finish-day). A project the user doesn't regularly visit accumulates completed tasks at the top of its list.

Fix: filter done tasks in the selector's project loop before they count toward `MAX_WIDGET_TASKS`. (Moot if the scope question above lands the same way.)

**3. `TaskListWidgetProvider.kt` — the pending project ID leaks when the activity launch fails.**

`queueProjectToOpen` commits before `startActivity`, and the `catch` doesn't clear it. When the launch is blocked, the project ID survives in SharedPreferences and gets consumed by the cold-start pull in `android-interface.ts` on the *next* app start for any reason — so the user opens the app normally hours later and is silently navigated to that project. Clear the pref in the catch.

**4. `appwidget_info.xml` — the selection can never be changed after placement.**

`android:configure` is added without `android:widgetFeatures="reconfigurable"`, so switching a widget from one project to another means deleting and re-adding it. One attribute fixes it. Related: every placement now forces a modal even for people who just want Today, and cancelling that dialog makes Android silently discard the widget — `android:configuration_optional` (API 33+) addresses both.

## Worth considering

- **Snapshot size / push frequency.** The selector now projects *every* unarchived visible project × up to 20 tasks into `widget_data`, but only the selected project is ever read. With 30 projects that's ~600 tasks in a blob that `WidgetDataService.pushCurrent()` stringifies, string-compares and writes to the native store — and `pushOnStateChange$` now re-fires on a task edit in *any* project, where previously only Today changes triggered a push. Since the bridge already runs both ways, exposing the union of selected project IDs to Angular and projecting only those (keeping `{id, title}` for all so the picker still works) would keep the blob roughly its current size.
- **`TaskListWidgetConfigureActivity`** — `AppWidgetManager.getAppWidgetInfo()` returning null makes the guard fall through to `finish()` with `RESULT_CANCELED`, which tells the host to drop the widget with no explanation. Have you seen it hold up on non-Pixel launchers? A null result there is worth distinguishing from a genuine provider mismatch.
- **Theme** — `AppTheme.NoActionBar` is opaque, so placement flashes a full-screen blank activity behind the `AlertDialog`. A Dialog or translucent theme would sit better.
- **Inbox** — `selectUnarchivedVisibleProjects` excludes `INBOX_PROJECT`, so an Inbox widget isn't possible. Intentional? Worth a line in the docs either way.
- **Staleness header** — project widgets return the project title before `WidgetData.headerFor` runs, so they never show the `(outdated)` marker added in #9098; a week-old snapshot just reads "Work". Defensible since there's no day semantics, but it's an undocumented divergence from Today.
- **Double parse** — `projectTitle(json, …)` and `parse(json, …)` each do a full `JSONObject(json)` on the now-larger blob, and `headerTitle` parses once *per widget id*; the comment removed in this PR explained the previous read-once-for-all design. Passing the parsed root down would keep that property.
- **`toWidgetTask`** writes into the enclosing `projectColors` as a side effect and is now called from the project loop too, so `projectColors` accumulates entries for every project with tasks rather than just Today's. Works, but the hidden mutation across two call sites is easy to misread.
- **Test gap** — no case for a done project task with **no** `doneOn`, which is dropped immediately and never shown. That's the state of every task completed before this ships, so it's worth pinning.
- **Docs placement** — `docs/wiki/3.05-Web-App-vs-Desktop.md` becomes the widget's only wiki mention, on a page comparing web and desktop. An Android/mobile reference page would be a better home.

Happy to look again once 1–4 are addressed — and I'd genuinely encourage cutting the done-task expiry, which would take a good chunk of this with it.

johannesjo:
**🔧 Needs changes** – labelled `needs work`; ping me once the items from the earlier review are addressed and I'll re-review.

Thanks again for this – the structure is genuinely good. The review already on this PR from 2026-08-21 still stands in full: nothing has been pushed since, so the four blocking items and the scoping question about the five-second done-task expiry are all still open, and I won't repeat that list here. What follows is only what that pass missed.

One useful fact first: **no rebase is needed.** The branch is 189 commits behind `master` now, but it still merges cleanly and every check is green. `master` has not touched `TaskListWidgetProvider.kt`, `WidgetData.kt`, `TaskListWidgetService.kt` or `android-widget.selectors.ts`, so the existing feedback applies to the code exactly as it stands.

### Should fix before merge

- **`res/xml/backup_rules.xml` and `res/xml/data_extraction_rules.xml`** – the new `task_list_widget` SharedPreferences file is keyed by `appWidgetId`, which Android does not preserve across a backup/restore or device transfer, but it is neither excluded from backup nor remapped on `AppWidgetManager.ACTION_APPWIDGET_RESTORED`. After a device transfer the launcher restores the widget under a new id, `selectedProjectId(newId)` is null, and the widget silently drops back to Today while the old `project_<oldId>` entry stays in prefs forever (`onDeleted` can never fire for an id that no longer exists). The same file also holds `PENDING_OPEN_PROJECT_KEY`, so a leaked pending id – blocking item 3 in the earlier review – rides the transfer and navigates the user into that project on the new device's first launch. `AndroidManifest.xml` sets `android:allowBackup="true"` and both rules files already exclude three other prefs files, so the cheap fix is one more `<exclude domain="sharedpref" path="task_list_widget.xml"/>` in each; the thorough fix is handling `ACTION_APPWIDGET_RESTORED` and remapping old ids to new.

- **`WidgetData.kt`, the `PROJECT_DONE_TASK_GRACE_MS` comparison** – the grace window compares `task.doneOn` against the local `System.currentTimeMillis()`, but `doneOn` is a wall-clock stamp taken on whichever device completed the task and replayed verbatim everywhere else (`task-shared-crud.reducer.ts:235`, `doneOn: capturedDoneOn ?? Date.now()`). So the window is five seconds only when the completing device is also the displaying device. If phone A's clock is three minutes fast, a task completed on A keeps its row on phone B's project widget for about three minutes and arms a `postDelayed` that far out; a device whose clock is behind gets no grace at all and the row vanishes on arrival. If the expiry stays, gate it on a locally observed instant (the `WidgetDoneQueue` tap time is already local) rather than the synced `doneOn`. If the scoping question lands on dropping the expiry, this goes with it.

### Nits

- **`TaskListWidgetConfigureActivity.kt`** – the `AlertDialog` is shown from `onCreate` and never held in a field or dismissed in `onDestroy`, so rotating the phone with the picker open leaks its window (`android.view.WindowLeaked`). What the user sees stays correct, since `onCreate` re-runs and rebuilds the dialog from prefs.

### Questions

- Should a widget's project selection survive a backup or device transfer at all? That decides which of the two fixes above is right: excluding the prefs file is one line and makes the reset explicit and clean, while the `ACTION_APPWIDGET_RESTORED` remap preserves the selection but is real code. I'd lean on the exclusion unless you think people will notice the loss.

A few things I chased and found are fine as written, so you don't spend time on them: `headerTitle()` narrowing its `try/catch` is not a regression (`parseMeta` catches internally and cannot throw); the per-instance `RemoteViewsFactory` works because the adapter intent already carried `EXTRA_APPWIDGET_ID`; `AppTheme.NoActionBar` descends from `Theme.AppCompat.DayNight.NoActionBar`, so `AppCompatActivity` will not throw; the new header `getBroadcast` PendingIntent cannot collide with the row template; and `WidgetDoneQueue`'s new timestamp map cannot outlive the done map.

---

dennnyb:
Thanks for the feedback. Will return back to it, and follow up.

johannesjo:
No new commits since my last pass, so nothing here changes the status: blocking items 1–4, the backup-rules and `doneOn` clock-skew items, and the scope question on the five-second expiry are still the list to work from, and I won't repeat them. This is only an addendum from a fresh read, plus one correction to something I got wrong.

**Correction to my 2026-09-03 comment.** I wrote that the two rules files "already exclude three other prefs files". They don't — each excludes exactly one, `com.superproductivity.superproductivity_preferences.xml`. The recommendation is unchanged (one `<exclude domain="sharedpref" path="task_list_widget.xml"/>` in each), only my count was wrong.

### Worth adding

- **The five-second expiry removes the widget's own undo.** `getViewAt` builds the checkbox fill-in from the *displayed* state, and `onReceive` says why: "Target state computed at render time from the DISPLAYED state (incl. pending overlay), so repeated taps toggle back and forth." On a project widget that property survives five seconds and then the row is gone, so a mis-tap can only be undone by opening the app — while the same mis-tap on a Today widget stays one tap away. This is another argument for cutting the expiry rather than a separate defect.

- **Hiding or archiving a project silently turns its widget into a Today widget.** `selectUnarchivedVisibleProjects` filters `isArchived` *and* `isHiddenFromMenu`, so the project drops out of `projects`, `projectTitle` returns null, and the header and the list both fall back to Today. `isHiddenFromMenu` is a sidebar-display preference, not a "stop using this project" signal, so this is easy to trigger by accident. Mitigating: the pref selection is not cleared, so un-hiding the project restores the widget — but with `reconfigurable` missing (item 4) that is the only way back. Worth a line in the doc, which currently says only "no longer available".

- **After an update the picker offers only "Today" until the app is opened once.** `parseProjects` reads `projects` out of the existing blob, and a `v: 1` snapshot written by the current release does not have that array. A user who updates and adds a widget before launching the app gets a one-item dialog and a widget pinned to Today. It self-heals on the next push — but again only if item 4 is fixed; otherwise that widget can never be re-pointed.

### Nits

- **`onDeleted` doesn't clear `PENDING_OPEN_PROJECT_KEY`** — only `project_<id>`. Deleting the widget leaves a stranded pending navigation for the next cold start to consume: the same symptom as blocking item 3, reached a second way.

- **`projectTitle` and `parse` disagree on what a usable project is.** `projectTitle` additionally requires a non-empty `title`; `parse` only requires a `tasks` array. A project with an empty title would render its task list (grace filter and all) under a Today header, with row taps no longer carrying the project. Not reachable today — the project form marks title `required: true` — but the two predicates have to agree, so it is worth hoisting one.

### One I chased and am *not* asking you to fix

`WidgetData.parse`'s `doneOn = … ?: nowMs` fallback. When a pending done target has no timestamp, `doneOn` becomes `nowMs`, the grace check `nowMs < nowMs + 5000` always passes, and `TaskListWidgetService` arms `scheduleProjectTaskExpiryRefresh` at `nowMs + 5000` — which re-enters `parse` with a fresh `nowMs` and arms it again. That branch never terminates: a full `updateAppWidget` for every widget every five seconds, indefinitely.

I could not find a way to reach it. `master`'s `WidgetDoneQueue` has no `WIDGET_DONE_TASK_TIMESTAMPS` key, so I expected an upgrade — or a cloud restore, since that prefs file isn't excluded from backup either — to yield targets without timestamps. It does, but the same old blob also has no `projects` array, so `projectTitle` returns null, `projectIdToOpen` stays null, and neither the grace filter nor the scheduling runs. A blob that *has* `projects` implies the new app ran, and that drains the queue. So it is latent, not live.

Still worth a one-liner if the expiry survives the scope question: a pending completion whose tap time is unknown has no grace to serve, so treat a missing timestamp as already expired rather than as "just now". `nowMs` is the one default that can never expire.

---

dennnyb:
Thanks for the detailed review. I addressed the blocking items and follow-ups in a0d8e1933: restored the direct activity path, fixed the 20-open-task cap, cleaned up pending navigation and backup state, and made widgets reconfigurable.

I kept the five-second expiry for now, but it now uses the local widget tap time and a per-task alarm fallback instead of synced doneOn or a process-static timer. I also added regression coverage, updated the docs, and verified Today/project switching on the Android emulator.

johannesjo:

Thanks for the updates. I re-checked all 13 items from the two earlier passes against the code at `a0d8e19`, not against the commit message, and they hold up:

- ✅ **Blocking 1 – header tap is back on a direct activity PendingIntent.** `TaskListWidgetProvider.kt:353` builds the intent for `CapacitorMainActivity` and `PendingIntent.getActivity` sends it, keeping `EXTRA_OPEN_PROJECT_ID` and the per-widget `data` Uri; `CapacitorMainActivity.kt:346` queues the project and fires the drain broadcast from both `onCreate` and `onNewIntent`. `launchMode="singleTask"` guarantees `onNewIntent` delivery, and the action-less intent is correctly ignored by `handleIntent`. Rows rightly stayed on the shared broadcast template.
- ✅ **Blocking 2 – completed tasks no longer eat the row budget.** `android-widget.selectors.ts:100` reserves 20 open slots per project before any completed candidate is considered, so a project whose first 20 `taskIds` are done still projects its open tasks. Both new specs fail against the pre-fix code, so the green is real.
- ✅ **Blocking 3 – the pending project ID is cleared when the launch throws** (`TaskListWidgetProvider.kt:103`). That is exactly what I asked for; see the first item below for where my own prescription was incomplete.
- ✅ **Blocking 4 – the selection can be changed after placement.** `appwidget_info.xml:12`, and `configuration_optional` also covers the forced-modal / silently-discarded-widget half. `TaskListWidgetConfigureActivity.kt:41` pre-checks the current selection, so re-entering the picker lands somewhere useful.
- ✅ **Backup and device transfer.** `task_list_widget.xml` is excluded in `backup_rules.xml` and in both the `cloud-backup` and `device-transfer` sections, so neither a stale `project_<id>` nor a stranded `pending_open_project` rides a restore.
- ✅ **The `doneOn` clock-skew item.** The window is local-only now: `doneGraceStartedAt` comes solely from `WidgetDoneQueue` tap timestamps, `doneOn` is off the wire entirely, and `syncedDoneOnDoesNotGrantProjectGrace` pins it. I walked tap → drain → push: keeping `KEY_DONE_TIMESTAMPS` through `getAndClear` is what preserves the full five seconds, and the pruning in `peekDoneTimestamps` only drops entries that are already expired.
- ✅ **All four nits** – dialog dismissed in `onDestroy`, `onDeleted` clears the pending key, one shared `isUsableProject` predicate, and the `?: nowMs` fallback that could have rescheduled forever is gone.
- ✅ **Both documentation items**, and ➖ the backup/transfer question is answered by taking the exclusion route.

Tying the expiry to the tap with a per-task alarm also closes the "process died inside the window" hole from the first pass, which was the part I was least comfortable with.

**Two things before this can move.** The branch no longer merges: `master` has since touched `CapacitorMainActivity.kt`, `AndroidManifest.xml`, both rules XMLs and both wiki pages, so it needs a rebase. And one scope call is still mine to make, at the bottom – please do not rework the expiry until I answer it.

### Should fix before merge

- **`TaskListWidgetProvider.kt:89`** – my earlier prescription for the pending-ID leak was incomplete, so the leak is still reachable. Android's background-activity-launch restriction blocks the start *silently*: `startActivity` returns normally and the system only logs it, so the `catch` never runs. `queueProjectToOpen` has already committed by then, the tap appears to do nothing, and the next time the user opens the app for any reason, the cold-start pull in `android-interface.ts:284` consumes the ID and drops them into that project unasked – the exact symptom the item was about. Cheapest fix that actually covers it: stamp the pending entry with `System.currentTimeMillis()` and have `getAndClearProjectToOpen` ignore anything older than a minute. Sorry for sending you at the wrong half of it.

### Nits

- **`TaskListWidgetProvider.kt:81`** – `scheduleProjectTaskExpiryRefresh` runs on every checkbox tap with no check that any widget has a project selected, so a Today-only install arms an alarm plus a main-looper `postDelayed` and performs one or two full `refreshAll` passes five seconds after every tap, changing nothing. Worth noting that `canScheduleExactAlarms()` is false by default for non-clock apps on Android 14+, so the double-schedule branch is the normal path on current devices rather than the rare one.
- **`android-widget.selectors.ts:19`** – `MAX_WIDGET_DONE_TASKS = 20` puts up to 40 tasks per project into the blob where it used to be 20, and native discards every completed task with no local tap timestamp, so nearly all of them are serialized, string-compared and written on each push for nothing. Only a task the user could actually have tapped needs to survive, so a much smaller done budget does the same job at half the payload.
- **`android-widget.selectors.ts:113`** – the candidate list is re-sorted inside the loop, once per completed task, rather than once at the end.
- **`CapacitorMainActivity.kt:263`** – when the header tap creates the task, the launch intent becomes the task's *base* intent, and `intent.removeExtra` only mutates this process's copy. After a reboot, reopening from Recents re-delivers `EXTRA_OPEN_PROJECT_ID` with `savedInstanceState == null` and re-queues the navigation, so the app opens into that project instead of the normal start view. Narrow, but it is the same surprise-navigation symptom without any pref leak.

### Questions

- **`TaskListWidgetProvider.kt:365`** – the new `partiallyUpdateAppWidget` after the full update is safe at this call site, but the KDoc it replaced said the opposite on purpose: `AppWidgetServiceImpl` does `if (isPartialUpdate && widget.views != null) merge else widget.views = views`, i.e. it *replaces* rather than ignores when there are no cached views, which is not what the javadoc claims. The rewrite now gives that javadoc claim as the reason the ordering is safe. Which host or launcher did you see retain the old header? Worth recording that instead, otherwise the next reader has two comments and no way to tell which is right.
- Still open on my side, and **not** more work for you: whether the five-second expiry earns its place at all. You have made it considerably better, but the machinery behind it grew rather than shrank – a per-task `AlarmManager` PendingIntent plus a main-looper handler, and 20 extra completed tasks per project on the wire that exist only to feed it – and the three nits above are all downstream of keeping it. I would rather settle that scope call myself than have you rework it twice, so please hold on that one until I come back to it.

---

PR #9628, opened 2026-08-16
Fix gitea API paging

RuairidhWilliamson:
## Problem

The gitea and forgejo integration only requests the first page of results when searching for issues. This is a problem if you have more than 100 open issues, or if the server has a lower page limit (e.g. forgejo's limit is 50 by default).

## Solution

Paginate the auto-import/backlog fetch: keep requesting consecutive pages until one comes back empty, with a hard cap of `MAX_PAGES = 4` so huge repos can't stall the sync. Errors on page 2+ are logged via `PluginAPI.log.err` and stop paging instead of failing the whole import (page-1 errors still throw).

One related simplification: `searchIssues` previously issued an extra `/repos/{owner}/{repo}` lookup solely to pass `priority_repo_id` to the global issue-search endpoint. That parameter was removed — it's not needed for the scoped issue fetch and saved a round-trip per search.

Adds a vitest setup for the plugin (`npm run test` in the package, vitest as a devDependency scoped to this plugin only) with a spec that pins the pagination behavior using two 50-item fixture pages, asserting three requests are made and 100 issues are merged.

Docs: the wiki Gitea entries updated from "Last 100" to "Up to 200".

I have tested this against forgejo and it works!

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:
Thanks for this, and welcome! 👋 The underlying bug is real and I could confirm it upstream: Gitea's `MaxResponseItems` default is **50** ([`modules/setting/api.go`](https://github.com/go-gitea/gitea/blob/main/modules/setting/api.go)), so our hardcoded `limit: '100'` is silently clamped on *stock Gitea too*, not just Forgejo. Backlog imports have been truncating at 50 for everyone. Good catch.

That said, I'd like a few changes before merging — one of them is a fairly serious regression.

---

### 1. 🔴 `searchIssues` now exhausts the whole instance

This is the blocker. I checked the current Gitea handler and `priority_repo_id` **is not a parameter of `/repos/issues/search` at all** — the accepted params are `state, labels, milestones, q, type, since, before, assigned, created, mentioned, review_requested, reviewed, owner, created_by, team, page, limit` ([`routers/api/v1/repo/issue.go`](https://github.com/go-gitea/gitea/blob/main/routers/api/v1/repo/issue.go)). It's silently ignored.

Scope instead comes from `SearchIssuesRepoIDs`, and since we never send `owner`, it takes the `allPublic = opts.OwnerName == ""` branch ([`routers/common/issue_filter.go#L69`](https://github.com/go-gitea/gitea/blob/main/routers/common/issue_filter.go)) — i.e. **every public repo on the instance**, plus the user's private ones. The only thing scoping results to the configured repo is our client-side `.filter(issue => issue.repository?.full_name === cfg.repoFullname)`. (The original `// Seens to be the only way to "filter" for the current repo` in e1d9bf3 had "filter" in scare quotes for good reason.)

Before, that cost one wasted request. Now it pages through *all of it*:

- `issue-provider-tab.component.ts` sets `SEARCH_MIN_LENGTH = 1` with a 400 ms debounce — typing a single character kicks this off.
- `switchMap` cancels the observable, not the in-flight promise, so every abandoned keystroke keeps paging to exhaustion in the background.
- Pinned searches fire it with no typing at all.

On a Codeberg-sized instance, one keystroke = hundreds of sequential 100-item requests, ~99% of which get discarded by the client-side filter.

I'd **leave `searchIssues` on a single page** and page only `getNewIssuesForBacklog`. That's also what `github-issue-provider` does (`per_page=50` on both paths, no paging). If search really does need more reach, the right fix is a different one — send `owner=<owner>` so the server scopes it, or switch to the repo-scoped `/repos/{owner}/{repo}/issues` endpoint — but that's a separate PR.

### 2. 🔴 `/user` is fetched inside the paging loop

`plugin.ts` L333-340 on `e9dd39a`:

```ts
while (page === 1 || issues.length !== 0) {
  const params = { … };
  if (cfg.scope === SCOPE_CREATED_BY_ME || cfg.scope === SCOPE_ASSIGNED_TO_ME) {
    const user = await http.get<GiteaUser>(`${base}/user`);   // ← every page
```

Doubles the request count for scoped configs, for a value that can't change mid-loop. Hoist it above the `while`.

### 3. 🟠 No page cap

Nothing bounds the loop. Any server or proxy that ignores `page` returns the same non-empty page forever → infinite loop, `allIssues` growing unbounded, wedged renderer. A `MAX_PAGES` guard (10 ≈ 500 issues at the effective limit of 50) is cheap insurance and also contains the next point.

Related: terminate on `issues.length < limit` rather than `!== 0`. As written, every call pays one extra round-trip for the empty terminating page.

### 4. 🟠 Uncapped backlog import amplifies an unthrottled dispatch loop

`issue.service.ts` does `issuesToAdd.forEach(issue => this.addTaskFromIssue(...))` — a tight synchronous dispatch loop with no `await new Promise(r => setTimeout(r, 0))` between iterations. That's a known constraint of our op-log (rapid bulk dispatches can drop state). The old 50-item ceiling was accidentally keeping it safe; first-time setup on a 2000-issue repo would now dispatch 2000 task creations in one pass and push 2000 ops onto the sync wire. Not your bug, but this change removes the thing that was containing it — another reason for the cap in #3.

### 5. 🟡 Smaller things

- The two paging loops are near-identical — worth extracting a `fetchAllPages(http, url, params)` helper, which also removes the `let issues` outer-scope mutation.
- `searchIssues` introduces `const limit = 100` but `getNewIssuesForBacklog` still hardcodes `'100'`.
- If page 2 fails (rate limit, transient 500), the whole call throws and the user gets *nothing* — previously they'd at least get page 1.
- The checklist ticks "added tests" but there are none. `azure-devops-issue-provider/src/plugin.spec.ts` is the pattern to follow; a fake `http.get` returning one full page then a short page is ~20 lines and would pin down the termination condition.

### 6. 🟡 `build-all.js` — please split this out

Production builds don't use this script. `prebuild → build:packages → packages/build-packages.js` **auto-discovers** every directory under `packages/plugin-dev/`, so Gitea does ship today. `build-all.js` is a stale hand-maintained duplicate used only by `buildFrontend:e2e`, and it's also missing `linear-issue-provider`, `trello-issue-provider`, `azure-devops-issue-provider` and `brain-dump` — so this adds a 26-line copy-pasted block that fixes 1 of 5 gaps.

The real fix is to point `plugins:build` at `build:packages` and delete `build-all.js`. Happy to do that separately — please drop it from this PR either way so the paging fix stays reviewable on its own.

---

**Summary:** hoist the `/user` call, add a `MAX_PAGES` cap with `< limit` termination, revert `searchIssues` to a single page, drop the `build-all.js` hunk, and add a small spec — and I'm happy to merge. Thanks again for digging into this; the `limit=100` clamp is a genuine long-standing bug that nobody had spotted.

RuairidhWilliamson:
Thank you for the review, I have addressed the most important comments. Let me know what you think.

johannesjo:
Thanks for the quick turnaround — `searchIssues` back to a single page, the `build-all.js` hunk dropped, `/user` out of the loop and a page cap added. Four of the six points, and the diff is much easier to reason about now. Re-reviewed `917d1e68`; `npm run checkFile` passes clean and it merges onto master without conflict.

Confirming the premise once more, since it's the justification for the whole change: `limit=100` is clamped to `API.MaxResponseItems` (default **50**) by [`ToCorrectPageSize`](https://github.com/go-gitea/gitea/blob/main/services/convert/utils.go#L15), which [`utils.GetListOptions`](https://github.com/go-gitea/gitea/blob/main/routers/api/v1/utils/page.go) applies to `ListIssues`. So backlog imports have been truncating at 50 on stock Gitea too. Real bug, worth fixing.

One blocker left, plus a sizing question.

---

### 🔴 `/user` is now fetched unconditionally — breaks scope-restricted tokens

`plugin.ts:310`:

```ts
const user = await http.get<GiteaUser>(`${base}/user`);
while ((page === 1 || issues.length !== 0) && page <= pageLimit) {
```

The hoist was the right move, but it moved the call out of the `if` as well as out of the loop. `GET /user` lives in a route group gated by `tokenRequiresScopes(AccessTokenScopeCategoryUser)` ([`routers/api/v1/api.go#L1165-L1286`](https://github.com/go-gitea/gitea/blob/main/routers/api/v1/api.go)), so it needs `read:user` and returns **403** without it.

Gitea's token UI forces you to pick scopes explicitly, so a token scoped to `read:issue` + `read:repository` is a perfectly ordinary thing to have. Today that token works fine for `scope: 'all'`. After this change the 403 is uncaught, `getNewIssuesForBacklog` rejects, and the user gets *nothing* — while `testConnection` still reports success, because it only probes `/repos/{repoFullname}`. Silent breakage of a working setup.

It's also a wasted round-trip every 5 minutes (`pollIntervalMs: 300000`) for every `scope: 'all'` user, which is the default.

Keep the hoist, keep the condition:

```ts
let user: GiteaUser | undefined;
if (cfg.scope === SCOPE_CREATED_BY_ME || cfg.scope === SCOPE_ASSIGNED_TO_ME) {
  user = await http.get<GiteaUser>(`${base}/user`);
}
```

### 🟠 `pageLimit = 20` is higher than it looks, and higher than the rest of the codebase

Because of the clamp above, the ceiling is 20 × 50 = **1000** issues, not 2000. Either way it's out of line with our other providers: `azure-devops-issue-provider` caps at `MAX_WORK_ITEM_LIMIT = 200` (default 50) and `trello-issue-provider` at 500.

And this feeds straight into `issue.service.ts:278`:

```ts
issuesToAdd.forEach((issue: IssueDataReduced) => {
  this.addTaskFromIssue({ … });
});
```

A tight synchronous dispatch loop with no `await new Promise(r => setTimeout(r, 0))` between iterations — a known constraint of our op-log, where rapid bulk dispatches can drop state. The old 50-item ceiling was accidentally containing it. First-time setup against a large repo would now dispatch up to 1000 task creations in one pass and push 1000 ops onto the sync wire.

`MAX_PAGES = 4` (≈200 at the effective page size, matching azure-devops's ceiling) would be my preference, with a short comment noting the ceiling is deliberate so a truncated import doesn't read as a bug later.

### 🟡 Correction to my earlier point 3 — ignore my `issues.length < limit` suggestion

I had that one wrong, and I'd rather flag it than let it get "fixed" later. Because the server clamps `limit=100` down to 50, `issues.length < limit` is true on the very first *full* page, so paging would stop after page 1 and reinstate exactly the bug this PR fixes. `PluginHttp` exposes the response body only, no headers ([`issue-provider-types.ts:137`](https://github.com/super-productivity/super-productivity/blob/master/packages/plugin-api/src/issue-provider-types.ts#L137)), so `X-Total-Count` isn't reachable either.

Your `!== 0` is the correct call. The one extra empty request is the right price, and a `// server clamps `limit`, so page size is not a reliable terminator` comment would be worth its line. If you do want to save that round-trip, deriving the effective page size from the length of page 1 and terminating on `issues.length < effectivePageSize` is clamp-robust — but that's optional polish, not a requirement.

### 🟡 Smaller

- If page 3 rate-limits or 500s, the whole call rejects and the user gets zero issues rather than the 100 already in hand. Returning `allIssues` on a mid-loop error would be more forgiving.
- `let issues` declared outside the loop purely to drive the condition — a `do { … } while` or a small `fetchPage` helper reads better now that there's only one loop.
- Dropping `priority_repo_id` is out of scope but correct — it appears nowhere in [`routers/api/v1/repo/issue.go`](https://github.com/go-gitea/gitea/blob/main/routers/api/v1/repo/issue.go), so it was silently ignored and the `GET /repos/{repoFullname}` lookup was pure waste. Happy to take it; worth a line in the PR description since it's no longer just paging.
- Still no test, though the checklist ticks "added tests" — and the PR description still describes the `build-all.js` change that's been dropped. Worth updating both. `packages/plugin-dev/azure-devops-issue-provider/src/plugin.spec.ts` is the pattern (vitest, `PluginAPI` stubbed in `beforeAll`, fake `http.get`). Two full pages then an empty one, plus one asserting `/user` is *not* called for `scope: 'all'`, is ~30 lines and pins down both the termination condition and the fix above.

---

**Summary:** make the `/user` fetch conditional again, drop the cap to ~4 pages, add a small spec, and this is good to go. The clamp is a genuine long-standing bug and I'm glad you chased it down.

johannesjo:

Re-checked at `917d1e68`. The head hasn't moved since my 2026-08-29 pass, so everything I raised there still stands – the unconditional `/user` fetch at `plugin.ts:310` (the blocker), the `pageLimit` sizing, the missing spec and the stale PR description. Nothing to add on those. One new item, and one note in your favour.

### Should fix before merge

- **`packages/plugin-dev/gitea-issue-provider/src/plugin.ts:303-336`** – the pages are concatenated with no dedupe, and `/repos/{owner}/{repo}/issues` is an offset window sorted `created_unix DESC`. If anything is opened in the repo between two page fetches, the window shifts and the issue that was last on page 1 comes back as the first item of page 2. `getNewIssuesForBacklog` then returns it twice, and `issue.service.ts:273` filters `issuesToAdd` only against the `allExistingIssueIds` snapshot taken *before* the fetch, never against itself – so both copies reach `addTaskFromIssue`, both lose the `_checkAndHandleIssueAlreadyAdded` race (it awaits a store read before either task exists), and one issue becomes two backlog tasks and two ops on the sync wire. The mirror case, an issue closed mid-loop, silently skips one instead.

  Accumulating into a `Map` keyed on `issue.id` and returning `[...map.values()]` instead of `allIssues.concat(issues)` fixes it, and drops the quadratic concat while you're there. Single-page code couldn't hit this, so it arrives with the paging.

### Note

Dropping `priority_repo_id` is safe, though for a different reason than I gave last time. Forgejo *does* still read the parameter – it's only current Gitea that removed it – but `SearchIssues` pins the sort to `created_unix DESC` when no `sort` is sent, and the default `bleve` indexer applies `PriorityRepoID` purely as a score boost, which that sort discards. So it was inert on any default instance, Forgejo included, and your removal is correct. Still worth a line in the PR description, since it isn't a paging change.

---

RuairidhWilliamson:
@johannesjo I have addressed the comments :)

johannesjo:

Re-reviewed the delta from `917d1e68` to `bd1cba84`. All four code items from the last pass are done: `/user` is conditional and fetched once (`plugin.ts:320`), the cap is `4`, the pages accumulate into a `Map` keyed on `issue.id` so the window-shift duplicate is gone, and a failure on page 2 or later now returns what's in hand instead of throwing.

I also checked the one thing that could have gone wrong silently here: Gitea clamps `PageSize` in `GetListOptions` *before* the offset is computed (`skip = (Page - 1) * PageSize`, `models/db/list.go`), so `limit=100` against a 50-item instance gives contiguous windows with no gaps between pages. The paging itself is correct.

One item is left over from the last pass, and it turns out to need a bit more than copying the azure-devops spec.

### Should fix before merge

- **`packages/plugin-dev/gitea-issue-provider/package.json`** – the spec is still missing, and the checklist still ticks "I have added tests". The part I didn't spell out last time: a `plugin.spec.ts` dropped into `src/` on its own would never actually run. `.github/workflows/plugin-tests.yml` builds its matrix by selecting the plugin packages whose `package.json` has a non-empty `scripts.test`, and this package only has `build`, `typecheck` and `lint`. So the full change is `vitest.config.ts`, `"test": "vitest run"` and `vitest` in `devDependencies` alongside the spec, exactly how `packages/plugin-dev/azure-devops-issue-provider` is set up. Plugin-scoped devDependencies are fine, and this also puts the plugin into CI permanently.

  What's most worth pinning down is the termination condition: two full pages then an empty one, asserting three requests and the merged result, plus one asserting `/user` is *not* called for `scope: 'all'`. `issues.length !== 0` looks like something that could later be "tightened" to `< limit`, and because the server clamps the page size, that tightening would silently reinstate the 50-issue truncation this PR exists to fix.

### Nits

- **`plugin.ts:338-344`** – the page 2+ error is swallowed with no trace anywhere. Returning partial results is the right call, but `plugin-issue-provider-adapter.service.ts:302` only shows the "Backlog import failed" snack when the plugin throws. So a token that loses a scope, or an instance that starts rate-limiting after page 1, quietly caps every 5-minute poll at one page, which looks exactly like the bug this PR fixes. A single `PluginAPI.log.err(...)` in the catch is enough. Log the page number only, not the error object – log history is exportable and a Gitea error body can carry issue titles.
- **`plugin.ts:306`, `311-312`** – the comments you added cover the clamp nicely. The bit still missing is that the ceiling is deliberate: a repo with 500 open issues imports 200 and stops, and nothing in the file says that is a decision rather than a bug. A module-level `const MAX_PAGES = 4;` with a short comment ("~200 issues at Gitea's clamped page size") beside the other constants would carry it, and matches how `azure-devops-issue-provider` does it with `MAX_WORK_ITEM_LIMIT`.
- **`plugin.ts:308`** – `let allIssues` is never reassigned, so `const allIssues = new Map<number, GiteaIssue>()` reads better. Nothing flags it because `eslint.config.js` ignores `packages/plugin-dev/**` entirely and the package's `lint` script is just `tsc --noEmit`.
- The PR description still describes the `build-all.js` change that has since been dropped, and doesn't mention the `priority_repo_id` removal. It ends up as the squash commit body, so it's worth a minute.

---

johannesjo:
**💬 Maintainer decisions** – two open questions from the last pass, now decided. Not a new review.

First, thank you — this is round four and you have turned every item around quickly and without fuss, including the ones that needed a rethink rather than a one-liner. The underlying bug is real and long-standing: Gitea clamps `limit=100` down to `API.MaxResponseItems` (default **50**), so backlog imports have been silently truncating at 50 for every bundled Gitea user. That is worth the four rounds on its own, and the current implementation of it is correct.

Two decisions, and one correction that is in your favour.

---

### 1. Test wiring — yes please, in this PR. And I am happy to do it for you.

I flagged the test twice without committing to whether it belonged here. It does, and here is the honest reason rather than "please add tests":

- **This plugin is currently invisible to CI.** `.github/workflows/plugin-tests.yml:38-47` derives its matrix from the plugin packages whose `package.json` has a non-empty `scripts.test`, then runs `npm ci` + `npm test` inside that plugin's directory (`:117-122`). `gitea-issue-provider/package.json` has `build`, `typecheck` and `lint` — no `test` — so it is not in the matrix at all.
- **Nothing else guards this code.** `eslint.config.js:46` ignores `packages/plugin-dev/**/*` entirely, and the package's `lint` script is just `tsc --noEmit`. Meanwhile the plugin is bundled (`src/app/plugins/plugin.service.ts:61`) and polls on `pollIntervalMs: 300000`, so the rewritten paging loop ships to every install and runs every five minutes.
- **It is a copy job, not new infrastructure.** `azure-devops-issue-provider` has an 8-line `vitest.config.ts`, exactly one added devDependency (`vitest`), and an 83-line `src/plugin.spec.ts` that stubs `PluginAPI` in `beforeAll`, fakes the plugin's http client and asserts the request the provider sends. `caldav-calendar-provider` and `google-calendar-provider` have the same three-file setup; `clickup-issue-provider` and `automations` also ship `"test": "vitest run"`. Both tsconfigs and both `scripts/build.js` are already byte-identical apart from the plugin name.
- **It stays plugin-scoped.** The root `package.json` workspaces field is `["packages/*"]` and `packages/plugin-dev/package.json` lists only `example-plugin`, `simple-typescript-plugin` and `procrastination-buster` as its workspace members — this plugin is not one of them, so `vitest` never reaches the root install tree. Our no-new-dependency rule is about root deps; plugin-scoped ones are fine.

Concretely, five things, all inside your plugin directory:

1. `"test": "vitest run"` in `scripts`
2. `vitest` in `devDependencies`
3. `vitest.config.ts` copied from `azure-devops-issue-provider`
4. a regenerated `packages/plugin-dev/gitea-issue-provider/package-lock.json` (CI runs `npm ci` there, so it has to be in sync)
5. `src/plugin.spec.ts` with two cases: **(a)** two full pages then an empty one — assert three requests and the merged result; **(b)** `scope: 'all'` never hits `/user`

**Please do not touch anything outside `packages/plugin-dev/gitea-issue-provider/`** — no root `package.json`, no root lockfile. If any of that sounds like more yak-shaving than you signed up for, say the word and I will push the wiring and the spec to your branch myself. Genuinely no problem either way; you have already done the hard part.

Case (a) is the one I care about. `issues.length !== 0` is the kind of condition someone tightens to `< limit` in a year, and because the server clamps the page size that tightening would silently reinstate the exact 50-issue truncation this PR exists to fix.

### 2. The `type=issues` fix — yes, fold it in here.

You raised whether this belongs in this PR. It does, because the paging change is what makes it visible: neither Gitea call passes a `type` filter, so open **pull requests** are imported as backlog tasks and nothing in the task reveals they are PRs. Gitea shares one number sequence between issues and PRs, so `getById` resolves them cleanly and they look like ordinary issues forever.

`type=issues` — plural — is the right filter and is safe on every instance in the wild:

- `routers/common/issue_filter.go` maps `{"pulls": true, "issues": false}`; an absent or unknown value returns `optional.None`, i.e. no filter at all. It fails open.
- Both call sites you hit — `SearchIssues` and `ListIssues` in `routers/api/v1/repo/issue.go` — have parsed `ctx.FormString("type")` since at least v1.16.9, and current Forgejo parses it at both too.
- No migration tail: already-imported PR tasks keep polling and keep working, they are simply no longer re-offered. Nothing orphans.
- Precedent for the default: `packages/plugin-dev/github-issue-provider/src/plugin.ts:297` appends ` is:issue` by default for the same reason.

Ask is two lines: add `type: 'issues',` to the params literal in `searchIssues`, and to the per-page params literal in the new backlog loop. No config option, no i18n. And to be explicit — I am **not** asking you to port the GitHub provider's `isPullRequest` modelling; that field has zero consumers in `src/app`, so it would be dead weight here.

### Correction: dropping `priority_repo_id` was right, and it is a small win

I have described this two different ways across the last two passes and neither was quite accurate, so for the record: **removing it is correct and is an improvement, not a regression.**

Since Gitea 1.21 the handler reads and throws the value away outright — `_ = ctx.FormInt64("priority_repo_id")` in `routers/api/v1/repo/issue.go`. On 1.20 and earlier, and on current Forgejo which still reads it, it only ever fed `SearchOptions.PriorityRepoID`, which reaches `applySorts` / the indexer's score boost and nothing else. It affected result **ordering**, never filtering. Repo scoping in `searchIssues` has always come from the client-side `issue.repository?.full_name === cfg.repoFullname` filter, which is untouched. So the `GET /repos/{repoFullname}` round-trip it cost was pure waste, and you have removed one request per search. Good catch — it just deserves a line in the PR description, since it is not a paging change.

---

Once the two lines and the spec are in, this is good to go from my side.

RuairidhWilliamson:
Ok all sorted now, sorry for the delay

johannesjo:
**🔧 Needs changes** — re-reviewed the latest head at `962b9b9b`.

The implementation issues from the previous rounds are fixed: `/user` is conditional and cached, paging is capped at four pages, results are deduplicated by Gitea ID, page 2+ failures preserve the results already fetched, and both relevant requests now use `type=issues`.

### Should fix before merge

- **`packages/plugin-dev/gitea-issue-provider/src/plugin.spec.ts:60-73`** — the paging test uses two 100-item pages, so it does not reproduce the server behavior that caused this bug. Gitea/Forgejo can clamp the requested `limit=100` to 50. I verified this by changing the loop in an isolated worktree to continue only when the previous page contains exactly 100 items: both new tests still passed, even though that implementation would stop after page 1 on a default 50-item server. Please make each fixture page 50 items, keep the empty third page, and expect 100 merged results. That makes a future `issues.length < requestedLimit` regression fail the test.

- **`packages/plugin-dev/gitea-issue-provider/src/plugin.ts:307`** — the four-page cap changes the documented auto-import behavior, but `docs/wiki/3.07-Issue-Integration-Comparison.md` still says “Last 100” in both the matrix and the Gitea details. On default 50-item servers this is now up to 200. Please update those two entries; `docs/documentation-guide.md` requires existing-provider behavior changes to update this comparison.

### Smaller cleanup

- Running the required `npm run checkFile` against the current target branch rewrites `plugin.spec.ts` at the `makeHttp` declaration and the nested nullish-coalescing expression. Please commit that formatting.
- The PR description still says the Gitea plugin was added to `build-all`, although that hunk was removed, and it does not mention the `priority_repo_id` removal.

Local verification on the exact head: plugin tests pass (2/2), typecheck passes after building `plugin-api`, and the plugin build passes. GitHub workflows have not run yet because the fork’s Actions runs are awaiting maintainer approval (`action_required`).

johannesjo:

Thanks for the updates. Re-checked the two items I left on 2026-09-04 at `962b9b9b` (the branch was squashed, so I compared the current file state rather than a commit range):

- ✅ **`type=issues` on both calls** – done, `plugin.ts:214` and `plugin.ts:318`. No config option, no `isPullRequest` modelling, exactly the two lines.
- ✅ **Test wiring** – `"test": "vitest run"`, `vitest` in `devDependencies`, `vitest.config.ts` byte-identical to `azure-devops-issue-provider`, and a regenerated plugin-local lockfile. Nothing outside `packages/plugin-dev/gitea-issue-provider/` was touched, so the plugin is now in the `plugin-tests.yml` matrix permanently. That was the part I most wanted, and the `Test gitea-issue-provider` job is green on this head, so the plugin is in CI for good now.
- ⚠️ **Spec case (a)** – the two-full-pages-then-empty shape is there, and this is my wording biting back: I said "two full pages" without saying full *of what*. Both fixture pages are 100 items long, which is the one size that cannot catch the regression I asked it to catch. If someone later tightens `issues.length !== 0` to `issues.length < limit`, `limit` is `100` and page 1 has exactly 100 items, so the loop still runs to page 3 and the test still passes – while a real 50-item-clamped server would stop after page 1 and reinstate the truncation. As my 2026-09-10 comment above also says: 50-item pages, keep the empty third page, expect 100 merged results.

### Still blocking

- **`packages/plugin-dev/gitea-issue-provider/src/plugin.spec.ts:60-74`** – make each fixture page 50 items rather than 100, so a future `< limit` tightening fails the test.

Everything else still open is in my 2026-09-10 comment above (the wiki `Last 100` entries, the prettier rewrite of `plugin.spec.ts`, and the PR description) – I've not repeated it here. Three small ones from my 2026-09-04 pass are also still untouched and worth a minute while you're in the file: a `PluginAPI.log.err` in the page-2+ catch at `plugin.ts:340` (page number only, not the error object), a module-level `const MAX_PAGES = 4` with a comment saying the ceiling is deliberate instead of the inline `pageLimit` at `plugin.ts:307`, and `const` rather than `let` for `allIssues` at `plugin.ts:309`.

---

RuairidhWilliamson:
Ok, I have addressed the comments

johannesjo:

Thanks for the updates. The branch was squashed again, so I compared each file at `962b9b9b` against `f61f17aa` rather than a commit range. Re-checked the items from my 2026-09-11 pass:

- ✅ **50-item fixture pages** – done, `plugin.spec.ts:59-60`. Both pages are 50 items, the empty third page is kept, and the assertions are three requests and 100 merged results. The loop still asks for `limit: '100'`, so I checked the regression the fixture is meant to catch by running the spec's own fake http against the compiled head and then against a `issues.length < limit` variant: at head it gives 3 requests / 100 results, and the tightened variant gives 1 request / 50 results and fails both assertions. Exactly what I wanted pinned.
- ✅ **Wiki `Last 100`** – both entries updated, the matrix row and the Gitea details bullet, and the parenthetical about server-imposed page limits is a better answer than the flat number I asked for.
- ✅ **Prettier rewrite of `plugin.spec.ts`** – committed. I ran the repo's prettier over both files at this head and they come back clean.
- ❌ **PR description** – still says the Gitea plugin was added to `build-all`, which was dropped four rounds ago, and still doesn't mention the `priority_repo_id` removal. It ends up as the squash commit body.
- ❌ **The three small ones in `plugin.ts`** – the file is byte-identical to `962b9b9b`, so a `PluginAPI.log.err` in the page-2+ catch (`plugin.ts:340`), a module-level `MAX_PAGES` const with a note that the ceiling is deliberate (`plugin.ts:307`), and `const` rather than `let` for `allIssues` (`plugin.ts:309`) are all still open. Optional, one line each.

### Still blocking

- **The branch no longer merges.** `master` has moved 325 commits and GitHub reports the PR as conflicting. I reconstructed the three-way merge to find out where: the only conflict is in `docs/wiki/3.07-Issue-Integration-Comparison.md`, because master rewrote the **GitLab** row directly above the Gitea row you edit. `plugin.ts` and `package.json` both merge clean, and the Gitea row on master still reads `Last 100`, so your edit still applies unchanged.

  Please resolve it by taking master's version of that file and re-applying only your two Gitea lines (the matrix row and the `Auto-import` bullet). Taking your side of the conflict wholesale would revert the newer Trello, GitLab and Plainspace rows that landed in the meantime.

### Nits

- **`docs/wiki/3.07-Issue-Integration-Comparison.md:18`** – while you are in that line anyway: the matrix cell says `Up to 400` with no caveat, but stock Gitea and Forgejo clamp the page size to 50, so the real ceiling on a default instance is 200. The details bullet carries the qualifier; the matrix cell read on its own overstates it. `Up to 200` (or `Up to 400` with the same caveat) would match.

One thing that isn't yours to fix: no workflow has run on this head, because the fork's Actions runs are awaiting maintainer approval. The `Test gitea-issue-provider` job that was green on `962b9b9b` was green against the old 100-item fixtures, so CI has not yet executed the new ones.

---

RuairidhWilliamson:
Ok all updated :)

johannesjo:
**🔍 Reviewed** – no blocking issues found at `dafa6ff9`.

Thanks for the updates. The branch was squashed again, so I compared each file at `f61f17aa` against `dafa6ff9`. Re-checked the 6 items from the earlier review:

- ✅ **Merge conflict in the wiki** – resolved the way I hoped: GitHub reports the PR as mergeable again, and against master's current `3.07-Issue-Integration-Comparison.md` your version differs only in the Gitea matrix row and the Gitea `Auto-import` bullet, so the newer Trello, GitLab and Plainspace rows are all intact.
- ✅ **Matrix cell overstating the ceiling** – now `Up to 200`, with the page-size caveat on the details bullet.
- ✅ **PR description** – the `build-all` mention is gone and the `priority_repo_id` removal is explained.
- ✅ **Named, deliberate page cap** – `MAX_PAGES = 4` with a comment saying the ceiling is deliberate, `plugin.ts:310-312`. Function-local rather than module-level is fine with a single use.
- ✅ **`const allIssues`** – `plugin.ts:314`.
- ⚠️ **`PluginAPI.log.err` in the page-2+ catch** – partly: the log call is there and the loop now `break`s cleanly (`plugin.ts:351-352`). Still missing: it passes `error`, and the ask was the page number only. On web and Electron that error is an Angular `HttpErrorResponse`, which is not an `Error`, so the log exporter doesn't reduce it to name and message – it serializes the object, including the full request URL with the repo name, `created_by` username and label names. Dropping the third argument (or logging just `(error as { status?: number })?.status`) settles it. Optional, and a one-line change.

I also ran the paging loop from this head against fake http fixtures outside the repo: 50 + 50 + empty gives 3 requests and 100 issues, a page-2 failure returns the 50 in hand and logs once, a page-1 failure still throws, and a server that ignores `page` stops at 4 requests.

---

johannesjo (approved):
Thanks for sticking with this through several rounds!

Issue #9450, opened 2026-08-05
🚨 RPM package uses incorrect StartupWMClass, causing a generic GNOME dock icon


trekirk:
### Version Used

v18.16.0, official RPM package

### Current Behavior

After upgrading Super Productivity using the official RPM package, GNOME no longer associates the running application window with the installed desktop launcher.

The launcher initially displays the correct Super Productivity icon, but once the application is running, GNOME shows a separate generic icon for a window identified as `superproductivity-bin`.

The RPM installs the following desktop entry:

Name=Super Productivity
Exec="/opt/Super Productivity/superproductivity" %U
Icon=superproductivity
StartupWMClass=superproductivity

However, the running application is identified by GNOME as:

superproductivity-bin

Because `StartupWMClass` does not match the actual window class, GNOME treats the launcher and the running window as two different applications.

The application icons themselves are installed correctly under `/usr/share/icons/hicolor/`.

### Expected Behavior

GNOME should associate the running Super Productivity window with the installed desktop launcher.

The application should use the official Super Productivity icon while running and should not create a second generic dock icon.

### Steps to Reproduce

1. Install or upgrade to Super Productivity v18.16.0 using the official RPM package.
2. Use GNOME as the desktop environment.
3. Open Super Productivity from the application menu or from a pinned dock launcher.
4. Observe the application while it is running.
5. GNOME displays the running window as a separate application named `superproductivity-bin`, using a generic icon instead of the Super Productivity icon.

The issue is fixed locally by copying the desktop file to the user application directory and changing:

StartupWMClass=superproductivity

to:

StartupWMClass=superproductivity-bin

Workaround commands:

mkdir -p ~/.local/share/applications

cp /usr/share/applications/superproductivity.desktop \
  ~/.local/share/applications/superproductivity.desktop

sed -i 's/^StartupWMClass=.*/StartupWMClass=superproductivity-bin/' \
  ~/.local/share/applications/superproductivity.desktop

update-desktop-database ~/.local/share/applications

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell
No relevant application console errors were observed.

This appears to be a desktop launcher/window identification issue rather than an application runtime error.
```

### Documentation

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.

[...]

Issue #5302, opened 2025-10-17
🚨 Duplicate icon in task bar (kde panel)


kolemcrae:
### Version Used

15.3.0 Flatpak

### Expected Behavior

In the kde panel there is a single superproductivity icon

### Current Behavior

There are two!

### Steps to Reproduce

Open Super Productivity. 
It shows two icons! 

First thing I did was check xprop and the desktop file. It seems the StartupWMClass has a space in it but the actual WMclass (found via xprop) does not. 

I tried correcting that in the desktop file but for some reason it still doesn't work. 

Any ideas? 

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
I am unable to reproduce this.

poghostick:
I fixed this by adding a line 

`StartupWMClass=superproductivity`

to the app's .desktop file in `/var/lib/flatpak/exports/share/applications/com.super_productivity.SuperProductivity.desktop`

madpipeline:
> I fixed this by adding a line
> StartupWMClass=superproductivity

On SP 16.3.0 installed via Flatpak on Fedora 42 KDE, this does not resolve the issue. It has no effect.

johannesjo:
@madpipeline How about 16.4.3?

madpipeline:
I confirm with 16.4.3 this fix works.

I had to edit the `.desktop` file and change from `StartupWMClass=Super Productivity` to `StartupWMClass=superproductivity`.

I can now pin to the taskbar and open from the same taskbar icon.

rowschank:
Hello. This issue still exists in the Flatpak 17.1.7, on KDE Plasma 6.5.5.  

1. Upon launching the Flatpak, a second icon is opened after a second or so, and the first icon vanishes afterwards. 

https://github.com/user-attachments/assets/a872b5f2-5998-4737-bc1a-d83b77c380e9

2. The remaining icon cannot be pinned to the taskbar: 

<img width="371" height="264" alt="Image" src="https://github.com/user-attachments/assets/b35e3b2c-5127-4dd3-9382-c79817908dc1" />

3. The system tray entry has no name and is just an icon: 

<img width="263" height="310" alt="Image" src="https://github.com/user-attachments/assets/b5873f6a-533c-43bd-82b9-87a0826eeb9f" />

olucasjose:
The icon duplication problem also occurs in the Cinnamon environment on Linux Mint.

[...]

PR #9634, opened 2026-08-18
fix(electron): Update StartupWMClass for electron-builder

matthiasmundel:
## Problem

A generic (gear) icon from the desktop environment shown in the task bar.

## Solution

'StartupWMClass: superproductivity' being replaced by 'StartupWMClass: superproductivity-bin' in electron-builder.yaml

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/electron-builder.yaml
+++ b/electron-builder.yaml
@@ -87,7 +87,7 @@ linux:
     - x-scheme-handler/superproductivity
   desktop:
     entry:
-      StartupWMClass: superproductivity
+      StartupWMClass: superproductivity-bin
   target:
     - AppImage
     - deb


johannesjo:
Thanks for this — and for the second independent report of the symptom (alongside #9450).

I dug into the mechanism empirically rather than from the packaging scripts alone, because the natural objection to this one-liner (raised on #9450) is "it fixes Wayland and breaks X11". **That objection does not hold at HEAD**, and the reason is worth writing down.

## What actually determines the window identity

Measured on an X11 session with an Electron 43.4.0 probe whose binary was named `superproductivity-bin`, plus `xprop` against a live installed Super Productivity:

| Path | Visible window identity | Source |
| --- | --- | --- |
| X11, any package | `WM_CLASS = "super productivity", "Super Productivity"` | `app.setName(APP_DISPLAY_NAME)` — `electron/start-app.ts:161` |
| Wayland native (deb/rpm/AppImage) | `app_id = superproductivity-bin` | ELF basename after the `afterPack` rename |
| Snap on Wayland | forced to X11 by the wrapper → same as row 1 | `build/linux/snap-wrapper.sh` |

Two probe results, not inference:

- Launched as `superproductivity-bin --class=superproductivity` → `WM_CLASS(STRING) = "super productivity", "Super Productivity"`. **Electron overrides `--class` for BrowserWindows.** The flag does land on Chromium's other (non-BrowserWindow) X11 window — visible on the live instance here as a second window, `"superproductivity-bin", "superproductivity"`, same PID — but not on the one the shell matches.
- Same probe with `app.setName(...)` removed → `"superproductivity", "superProductivity"`. That is the pre-v18.13.0 shape, and its **instance** field is what `StartupWMClass=superproductivity` used to match under GNOME.

So the origin of the mismatch is #8640 (`fix(electron): use display name for Linux notifications`, shipped in v18.13.0), not the wrapper. Since then `StartupWMClass=superproductivity` matches **neither** field on X11 — GNOME's fallback lowercases and maps spaces to dashes (`super-productivity.desktop`), which also misses.

## What that means for this PR

- **No X11 regression is possible here** — X11 matching is already broken, so there is nothing left to break. This change is strictly an improvement.
- **The Wayland fix is real**, and `superproductivity-bin` is the right value: it is exactly `RENAMED` in `tools/afterPack.js`, which is what Ozone/Wayland reports as `app_id`.
- **It does not fully close #9450.** X11 sessions stay broken, and so does the Snap on Wayland (the wrapper forces `--ozone-platform=x11`, so those windows get the X11 identity). Worth noting in the PR description so this isn't read as a complete fix.

## Suggested changes before merge

1. `build/linux/snap-wrapper.sh:36` — the comment `# Must match linux.desktop.entry.StartupWMClass in electron-builder.yaml.` becomes false with this diff, and `tools/afterPack.js:7` ("forces `--class=superproductivity` for stable desktop-file matching") likewise. Both should record that the matched class comes from the ELF name under Wayland and from `app.setName()` under X11, and that `--class` never reaches the BrowserWindow.
2. `electron-builder.yaml` — a one-line comment tying this value to `RENAMED` in `tools/afterPack.js`. Nothing in CI couples the two; a future rename would silently re-break the icon.
3. Commit message `Update StartupWMClass for electron-builder` doesn't follow `type(scope): description` (the PR title does).

## Follow-up, separate from this PR

One `StartupWMClass` value cannot serve an X11 class of `Super Productivity` and a Wayland `app_id` of `superproductivity-bin`. The durable fix is to make the two agree — either give the app a WM-safe name again and satisfy #8640's human-readable notification name another way, or drop the ELF rename so both derive from `superproductivity`. This may also explain #5302 (duplicate icon in the KDE panel): the process puts two windows with *different* `WM_CLASS` on screen.

<sub>Caveats: the Wayland half is not verified by me (no Wayland session available) and rests on @trekirk's GNOME Looking Glass output in #9450; I also did not check whether the Flatpak build goes through `afterPack` at all.</sub>

johannesjo:
Thank you very much! <3

johannesjo:
@matthiasmundel — heads-up rather than a criticism: I've reverted this in #9678. The reason is a version boundary that was genuinely easy to miss.

Your analysis was right for the builds it was measured against. @trekirk and @sanchpet reported on v18.16.0, which ships Electron 41, and there the Wayland `app_id` really does come from the renamed `superproductivity-bin`.

Five days before this merged, v18.20.0 bumped Electron to 43 (#9571). Electron 42 moved the window identity to the XDG app id — `native_window_views.cc` now derives both the X11 `WM_CLASS` and the Wayland `app_id` from `GetXdgAppId().value_or(Browser::GetName())`. Measured on a real 43.4.0 binary, the window reports `superproductivity` on both session types, so `-bin` matches nothing and the next release would have shown the generic icon to everyone rather than to Wayland users only.

Nothing shipped — this never reached a published release, so no user saw it either way.

The real problem is that nothing coupled `StartupWMClass` to the value the app actually reports, so neither of us had a way to notice the ground had moved under the issue. #9678 adds `tools/verify-linux-wm-class.test.js` for exactly that; it fails on this PR's value, which is how I caught it.

Thanks for chasing this down — and the Looking Glass trace in #9450 is what made the Electron 41 behaviour unambiguous in the first place.

Issue #9635, opened 2026-08-18
💡 "Connection failed" is not a descriptive enough error message when testing


confestim:
### Problem Statement

While working to set up my issue providers (i.e. iCal and Gitea), I found myself opening the dev tools quite often. I believe that the error message "Connection failed" is incredibly vague and lacklustre **especially** in a testing context.

### Possible Solution

All that needs to be done is to include the error code within the error message pop-up, if any, that is.

I have created a fork and a PR for this.

P.S. I tried to be proactive and have Claude translate to all locales but:

``` 
I can't do that — the project's CLAUDE.md is explicit: "Translations: UI strings go through T / TranslateService. Edit only en.json; never other locales." That's a hard rule, not a style preference (likely because other locale files are managed through an external translation workflow, e.g. Weblate/Crowdin, and hand-edits would get overwritten or desync from that pipeline).

en.json already has the fix ("Connection failed: {{errorMsg}}"). The other ~20 locale files will need this string translated through whatever normal translation process this project uses — I'd be glad to help with anything else, but I won't hand-edit those files.

```

Props to you for this!

PR #9636, opened 2026-08-18
 fix(issue-provider): include error detail in "Connection failed" message (#9635)

confestim:
## Problem

Vague error message when testing connection to an issue provider.

## Solution

Added status code response when an error occurs.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.ts
+++ b/src/app/features/issue/dialog-edit-issue-provider/dialog-edit-issue-provider.component.ts
@@ -72,6 +72,7 @@ import { ISSUE_PROVIDER_COMMON_FORM_FIELDS } from '../common-issue-form-stuff.co
 import { TagService } from '../../tag/tag.service';
 import { ChipListInputComponent } from '../../../ui/chip-list-input/chip-list-input.component';
 import { unique } from '../../../util/unique';
+import { getErrorTxt } from '../../../util/get-error-text';
 import { mergeIssueProviderModelUpdates } from './issue-provider-model-merge.util';
 
 type OptionsLoadState = 'idle' | 'loading' | 'loaded' | 'empty' | 'failed';
@@ -300,6 +301,7 @@ export class DialogEditIssueProviderComponent {
         this._snackService.open({
           type: 'ERROR',
           msg: T.F.ISSUE.S.CONNECTION_FAILED,
+          translateParams: { errorMsg: 'Unknown error' },
         });
       }
     } catch (error) {
@@ -307,6 +309,7 @@ export class DialogEditIssueProviderComponent {
       this._snackService.open({
         type: 'ERROR',
         msg: T.F.ISSUE.S.CONNECTION_FAILED,
+        translateParams: { errorMsg: getErrorTxt(error) },
       });
     }
   }
--- a/src/assets/i18n/en.json
+++ b/src/assets/i18n/en.json
@@ -516,7 +516,7 @@
       },
       "S": {
         "AUTO_CREATE_FAILED": "Auto-create issue failed: {{errorMsg}}",
-        "CONNECTION_FAILED": "Connection failed",
+        "CONNECTION_FAILED": "Connection failed: {{errorMsg}}",
         "CONNECTION_SUCCESS": "Connection works!",
         "ERR_GENERIC": "{{issueProviderName}} Error: {{errTxt}}",
         "ERR_NETWORK": "{{issueProviderName}}: Request failed because of a client side network error",


johannesjo (commented):
Thanks for this — the underlying problem is real, and actually a bit worse than the issue describes. But the current implementation misses the exact case you reported, so I'd like one more change before this goes in.

## The problem is real (and the diagnosis is worth recording)

For Jira/GitLab/Redmine/OpenProject, `handleIssueProviderHttpError$` (`src/app/features/issue/handle-issue-provider-http-error.ts:28-52`) **already** opens a detailed error snack. It then rethrows, `testConnection()` rejects, and the component's `catch` opens `CONNECTION_FAILED`. Because `SnackService._openSnack` is `@debounce(100)` with `leading` unset — i.e. trailing-edge only (`src/app/util/decorators.ts:161`, `209-224`) — the second call clears the first timer and only the last one ever renders. So the detailed message that was already produced gets silently **erased** by the vague one. Your change restores it for those providers. Good catch.

## Blocking: the `else` branch is the path you actually hit

Neither provider you named in #9635 ever throws — both swallow the error into `false`:

- **iCal** — `src/app/features/calendar-integration/calendar-integration.service.ts:396-399`: `catchError(err => { Log.err(err); return of(false); })`
- **Gitea** — `packages/plugin-dev/gitea-issue-provider/src/plugin.ts:299-301`: `catch { return false }`, and `src/app/plugins/issue-provider/plugin-issue-provider-adapter.service.ts:57-64` swallows it a second time

This is true of **every** plugin-based issue provider, including the bundled ones (`github-issue-provider/src/plugin.ts:277-282`, `clickup-issue-provider/src/plugin.ts:133-138`). For all of them the adapter never calls `handleIssueProviderHttpError$` at all, so `CONNECTION_FAILED` is the *only* feedback the user gets.

So for your own setup this PR changes nothing except appending a hardcoded, untranslated English `'Unknown error'` — which is both noisier than today's message and against the repo's translations rule (all UI strings go through `T` / `TranslateService`).

## Suggested minimal fix: split the key

```jsonc
// en.json
"CONNECTION_FAILED": "Connection failed",
"CONNECTION_FAILED_WITH_ERR": "Connection failed: {{errorMsg}}",
```

The `else` branch keeps `CONNECTION_FAILED` with no `translateParams`; the `catch` uses `CONNECTION_FAILED_WITH_ERR`. Then run `npm run int` to regenerate `t.const.ts`.

This also closes an i18n footgun that the current diff leaves open: ngx-translate's `interpolateString` returns the **raw substring** when a param is missing (`@ngx-translate/core` → `interpolateString`, `replacement !== undefined ? replacement : substring`). So any future call site that forgets `errorMsg` renders a literal `{{errorMsg}}` on screen. Two keys make that unreachable rather than relying on every caller remembering.

If you want to actually close #9635 for iCal and the plugin providers, the follow-up is to stop swallowing in those spots (propagate instead of `of(false)` / `return false`). That's small, and it's the part that carries the real user value here.

## Non-blocking notes

1. **Non-English users see no change.** All 24 other locale files keep the placeholder-less string, and since the key resolves there's no fallback to `en`. Expected given the translation workflow — just noting the fix won't reach most users right away. (And you were right not to hand-edit the other locales. 👍)

2. **Credential exposure in the snack.** `getErrorTxt` surfaces `HttpErrorResponse.message`, which embeds the full request URL. CalDAV/Nextcloud users often put basic auth in the URL, and an iCal "secret address" URL *is* the credential. It's display-only — `SnackService` never logs the message, so nothing lands in the exportable log — but these snacks routinely get screenshotted into bug reports. Worth truncating or stripping the URL. `getErrorTxt`'s `JSON.stringify` fallback can also dump 200 chars of a raw response body.

3. **No test.** The component has no spec today and this adds none. A small unit test asserting the `msg` key for each branch would lock in the split above cheaply.

4. Minor: the PR is filed from your fork's `master` branch, and it's checked as "New feature" while the title says `fix`.

I applied the patch locally and `npm run checkFile` passes on the changed file, so it's only the branch-coverage/i18n point above holding this up.

johannesjo:
**🔧 Needs changes** – labelled `needs work`; ping me once the items from my earlier review are addressed and I'll re-review.

This is a follow-up pass, not a second review – the one above still stands and I won't repeat it. I re-checked it against current master (207 commits have landed since this branch) and nothing has moved: `dialog-edit-issue-provider.component.ts:304` still hands the hardcoded English `'Unknown error'` to a translated string, and both swallow sites are unchanged (`calendar-integration.service.ts:397-403`, `plugin-issue-provider-adapter.service.ts:56-64`), so that `else` branch is still the only one iCal and the plugin providers ever reach. The two-key split (`CONNECTION_FAILED` / `CONNECTION_FAILED_WITH_ERR`) is still the change I'd like before this goes in.

Two corrections to that review, both in your favour:

- **Non-blocking note 2 (credentials in the snack) – please ignore it.** I went back through every shipped `testConnection`: iCal never reaches the `catch` at all, and the providers that do (Jira, GitLab, Redmine, OpenProject, CalDAV, Nextcloud Deck) all keep their secret in a request header or a separate password field rather than in the configured URL. The most that message exposes is the API host, which is the part that makes it useful – no stripping needed.
- **There is a spec.** I said the component has none; `dialog-edit-issue-provider.component.spec.ts` was added in #8798 and already registers `IssueService` with a `testConnection` spy and `SnackService` with an `open` spy, with the template overridden to `''`. So locking in the `msg` key for each branch is about ten lines and needs no new harness.

---

johannesjo:
Thanks for this — the diagnosis is right and I've taken it forward in #9900 (`a84ec0b`), crediting your report of #9635.

Two things made me write it slightly differently rather than merge as-is:

- **The message key.** Turning the shared `CONNECTION_FAILED` into `"Connection failed: {{errorMsg}}"` changes the arity of a key that every other locale still ships without the placeholder, and it forces the success-path-but-falsy branch to invent `'Unknown error'`. I added a separate `CONNECTION_FAILED_WITH_ERROR` instead, so `CONNECTION_FAILED` stays correct where there genuinely is no error to show.
- **Why the detail was missing at all.** `CalendarIntegrationService.testConnection` ended in `catchError(() => of(false))` and the plugin adapter's `testConnection` had a `catch { return false }` — the reason string was destroyed before your `catch` could ever see it, so for those two providers the snack would have printed "Unknown error" even after this patch. Both swallows are removed in `a84ec0b`, with the specs that asserted the old swallow rewritten to assert the rejection.

Closing once #9900 lands. Your `getErrorTxt(error)` call in the catch branch is unchanged in mine.

Issue #8828, opened 2026-07-07
💡 Mobile: icon-only settings tabs, add-task FAB over non-task views, content under bottom nav


johannesjo:
## Problem

Three mobile-layout issues found in a 390×844 visual pass (mobile web; overlaps Android WebView):

1. **Settings tabs lose their labels below 1280px** (`src/app/pages/config-page/config-page.component.scss:48-50` hides `.tab-label` under `mq(md, max)`), leaving six unlabeled icons for 22 sections of settings. `md` = 1280px, so even tablets get icon-only tabs.
2. **The global add-task FAB renders on every route**, including Settings — the bottom-nav FAB (`mobile-bottom-nav.component.html`) has no route-based suppression, so it floats over form fields where "add task" has no context (screenshot: it occludes the Keyboard Shortcuts row).
3. **Content runs under the bottom nav** — scrollable views need bottom padding equal to the nav height (+ safe area, see #8792) so the last rows aren't permanently occluded.

## Suggested fix

- Keep tab labels to a much smaller breakpoint (or use scrollable labeled tabs on phones).
- Hide/swap the FAB on non-task routes (on Settings it could be "search settings" once that exists, or simply absent).
- Bottom padding via the existing `--bottom-nav-safe-area` token.

---
*From a full design/UX audit (Playwright visual tour + code audit), 2026-07.*

MarkQuach12:
Hi @johannesjo, am I able to handle this issue?

johannesjo:
Thank you @MarkQuach12 ! It seems these findings are outdated/incomplete.

1. While it could be done more elegantly (maybe by at least providing a tooltip), this was intentional. Otherwise navigating on mobile would be quite cumbersome.
2. I think this is correct and should stay this way.
3. Needs investigation if it is actually a problem.

MarkQuach12:
@johannesjo I am happy to take a look at the 3rd problem and see if its still an existing issue

johannesjo:
Thank you!

steindvart:
Regarding item 1: hiding the labels below 1280 px is not a bug by itself. Icon-only tabs are a reasonable and appropriate responsive adaptation for smaller screens.

The actual issue was that the tabs still used `flex-grow: 0`. As a result, the six icon-only tabs occupied only part of the header, leaving a large unused area and unnecessarily small click/tap targets.

[...]

PR #9637, opened 2026-08-18
fix(settings): centre the settings tab strip

steindvart:
## Problem

The Settings tab strip was left-aligned across the full header width, while the content column below it is centred at `--component-max-width`. The strip hugged the left edge at every width, which reads as dead space next to the tabs, most visibly in the icon-only layout, where the six icons sat at the far left.

Refs #8828

## Solution

Centre the strip, and let `mat-tab-group` own both layout decisions through its inputs instead of overriding Material from the component stylesheet:

- `mat-align-tabs="center"` replaces the `::ng-deep .mat-mdc-tab-labels { justify-content: flex-start }` rule.
- `[mat-stretch-tabs]="false"` replaces `flex-grow: 0 !important`.

Tabs keep their natural widths in both the labelled and the icon-only band; nothing stretches.

Why an input rather than just deleting the `!important`: `stretchTabs` defaults to `true`, which adds `mat-mdc-tab-group-stretch-tabs`, and Material's `flex-grow: 1` rule for that class outranks the component selector. Removing the `!important` on its own brings the stretch back.

Note: the input's alias is `mat-stretch-tabs`. Binding `[stretchTabs]` does not fail the build but has no effect.

## Validation

- `npm run checkFile` on `config-page.component.scss` and `config-page.component.html`
- `npm run test:file src/app/pages/config-page/config-page.component.spec.ts`: 6/6 passed
- Browser checks at 390, 1024 and 1440 px:
  - `flex-grow` computes to `0` with no `!important`, and `mat-mdc-tab-group-stretch-tabs` is absent
  - 1440 (labelled): natural tab widths, 935px strip, centred on the content column, no pagination
  - 1024 and 390 (icon-only): 48px tabs, centred on the content column, no pagination

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki). (No wiki change is required for this visual-only change.)
- [x] I have added tests for my changes (if applicable). (No new automated test is applicable; the responsive states were verified in the browser and the existing component spec passes.)


johannesjo (commented):
Thanks for digging into this @steindvart — the empty space you spotted at tablet/laptop widths is real, and it pushed me to measure the tab strip properly. What I found makes me think the whitespace is a symptom of something else, so I'd like to explore a slightly different direction before landing `flex-grow: 1`. None of this is set in stone — pushback welcome.

### What the change measures out to

Real render at `#/config`, tab widths in px:

| viewport | today | with this PR |
| --- | --- | --- |
| 390 | 6 × 48 | 6 × 60 |
| 1024 | 6 × 48 | 6 × **120** |
| 1440 | 138–203 (955 total) | 149–233 (1133) |
| 1920 | 138–203 (955 total) | 229–313 (**1613**) |

Two things stood out:

- On a phone the gain is ~12px per tab — barely visible, and 48px already meets the 48dp tap-target minimum, so I don't think the targets were too small.
- `md` is 1280px, so the icon-only band covers tablets *and* most laptop windows. At 1024 this turns six 48px icons into six 120px blocks each holding one 24px icon, and inflates the active-tab `--state-selected` fill into a large slab. At 1920 the strip spans 1613px while `.tab-content` is capped at 800px and centred, so the tabs visually detach from the column they belong to.

Two smaller notes: `flex-grow: 0` wasn't an oversight — it went in with #9248 paired with `justify-content: flex-start` on `.mat-mdc-tab-labels`, which becomes a no-op once tabs grow (so that line would need to go too). And `daily-summary.component.scss` uses the same `flex-grow: 0 !important` pattern, so settings would become the only stretched tab strip in the app.

### What I think the underlying problem is

The page stacks three things that should share one axis, but don't:

| element | width | alignment |
| --- | --- | --- |
| search field | 800px | centred |
| tab strip | full width | left-aligned |
| content column | 800px | centred |

The gap you noticed is that mismatch. Stretching closes it by making the strip *more* full-width, which is why it reads worse at the wide end.

### And something bigger hiding under the breakpoint

The 1280px threshold is tuned to English with ~18px of headroom (available width at 1280 is 973px after the ~300px sidenav; the English label row needs 955px). Several locales are 30–60% longer:

| locale | label row | at 1280 (973px avail) | at 1440 (1133px) |
| --- | --- | --- | --- |
| en | 955px | fits (18px spare) | fits |
| hr | 1224px | overflows → pagination arrows, "Sinkronizacija i sigurnosne kopije" hidden | overflows |
| ru | 1187px | overflows | overflows |

I verified hr and ru in the browser; pt/pl/it/es sit in the same range by label length. So some users already lose a tab behind an arrow at common desktop widths — which feels more worth fixing than the whitespace.

### A direction that might be better

I prototyped this in the running app, but I'm not attached to it:

1. **`justify-content: center`** instead of `flex-start` — puts the strip on the same axis as the search field and the content column, which removes the left-hugging look at every width without stretching anything. Matches what `daily-summary` already does. If we only did one thing, I'd do this.
2. **Drop `min-width: 120px`, use `padding: 0 16px`** — English row goes 955 → 858px, buying ~100px of labelled range for free.
3. **Replace `mq(md, max) { .tab-label { display: none } }` with a fit check** — measure the labelled row width once on init, then on resize compare it against the header's available width and toggle an `.is-icon-only` class (~15 lines + a `ResizeObserver`). Measuring once is what stops it oscillating.

With 1–3 applied I saw no pagination in any locale/width combination I tried (en + hr, 390–1920). English keeps labels down to 1280 exactly as today; Croatian shows all six labels at 1440 (broken today) and falls back to centred icons at 1280 instead of arrows. That also speaks to the "six unlabeled icons" half of #8828 more directly than the whitespace does.

Would you be up for reworking the PR around (1), with (2)/(3) as a follow-up? I'm least sure about (3) — it's the only part that adds JS to a page that currently has none, so if you see a simpler way to make label visibility locale-proof, I'd prefer that. And if there's a case for stretching that I've missed, please push back.

<sub>Side note on the description: `src/assets/bundled-plugins/**` is gitignored build output, so that lint error is a local artifact rather than something broken on master — no need to worry about it.</sub>

johannesjo:

Follow-up pass on the same commit (`39ba380`). Nothing has changed on the branch since the review above, so I won't repeat those points. Two bits of context and one new detail:

- **Master is unchanged in this area.** `config-page.component.scss` still has `flex-grow: 0 !important` and `justify-content: flex-start`, so the one-line diff still applies exactly as written. The branch is 188 commits behind `master` now, so it will want a rebase whichever direction it goes.
- **The suggested alternative is already the app's own pattern.** `src/app/pages/daily-summary/daily-summary.component.scss:23-29` pairs `justify-content: center` on `.mat-mdc-tab-labels` with `flex-grow: 0 !important` on `.mat-mdc-tab`, so option (1) is a one-word change here rather than a new convention to invent.

### Nits

- **`src/app/pages/config-page/config-page.component.scss:68`** – whichever direction this lands in, this line is the wrong tool. `MatTabGroup.stretchTabs` defaults to `true`, so Material already applies `flex-grow: 1` to these tabs via `.mat-mdc-tab-group-stretch-tabs`; `flex-grow: 1 !important` restates that default instead of removing the `0` override, and simply deleting the declaration renders identically. In the other direction, the repo already expresses "don't stretch" through the tab-group input rather than SCSS – `mat-stretch-tabs="false"` on `boards.component.html:2` and `issue-panel.component.html:3` – and that input stays silently inert as long as any `flex-grow: … !important` sits in this block. So the cheapest shape either way is the input on the `mat-tab-group`, plus `justify-content: center` on `.mat-mdc-tab-labels`, with both the `flex-grow` and the `justify-content: flex-start` lines deleted.

### Questions

- Are you still up for reworking this around `justify-content: center` (option 1 above)? Happy to leave it open as long as you need – I just don't want it sitting unanswered.
- If you would rather not carry it further, that is completely fine – say so and it can be picked up separately, no hard feelings.

---

steindvart:
@johannesjo 

> Questions ...

Yes, I will finalize this change. I remember, but i was just busy with my main job :)

steindvart:
Thanks for measuring this properly — the axis diagnosis is what moved me off the original approach, and you were right that `flex-grow: 1` across the header makes the wide end worse.

Pushed two commits.

**1. `justify-content: center`** — your point (1), as suggested. `flex-grow: 0` from #9248 stays where it was.

**2. Stretch, but only in the icon-only band and bounded by the content column:**

```scss
::ng-deep .mat-mdc-tab-labels {
  justify-content: center;

  @include mq(md, max) {
    width: 100%;
    max-width: var(--component-max-width);
    margin-inline: auto;
  }
}
// plus flex-grow: 1 !important inside the existing mq(md, max) block on .mat-mdc-tab
```

**Why the second commit.** Centring fixes the axis at the centre but not at the edges — it centres inside the tab header, which is wider than the column. At 390px:

| | tab strip | search field / content column |
| --- | --- | --- |
| centred only | 51–339 | 64–382 |
| bounded stretch | 64–382 | 64–382 |

So the six icons still sat in a cluster that lined up with nothing above or below them. Bounding the stretch to `--component-max-width` gives the row the column's exact box.

Your objections all land on the labelled band, and this leaves that band untouched. Measured at 1920: the strip stays 935px, centred over the 800px column, `flex-grow: 0`, no full-header span — same at 1440. That also keeps it consistent with `daily-summary`, which has no icon-only mode to differ in.

On the icon-only side the numbers move considerably more than in my first attempt — 48 → 160px at 1024 and 48 → 74px at 390 — because the tabs now fill the column instead of the header.

**On the 1280 pagination.** I checked whether centring caused it by swapping `center` / `flex-start` at runtime: the geometry was identical to the pixel, arrows in both. It is the labelled row (935px) exceeding the available width, and `justify-content` is a no-op once there is no free space — so neither this PR nor the original `flex-grow: 1` moves it either way.

**On the locales.** Confirmed independently: ru measures 1203px against 899px available at 1280 and 1059px at 1440, so "Синхронизация и резервирование" sits behind an arrow at both. Two things for the follow-up, if they help:

- Your (2) measures out to −101px on the English row (935 → 834), so it clears 1280 for en but not for ru.
- An argument for (3) I did not see mentioned: the media query keys off *viewport* width, but the constraint is the header's *available* width, and that moves by roughly 200px at a fixed viewport depending on whether the sidenav is expanded or collapsed to a rail. So the fixed breakpoint is wrong for English too, in both directions — it hides labels that would fit and shows labels that do not. Also, the `ResizeObserver` + `afterNextRender` + `DestroyRef.onDestroy` pattern already exists in `main-header.component.ts`, so (3) would not be introducing a new mechanism to the codebase.

Happy to take (2) and (3) in a separate PR.

steindvart:
One correction to my previous comment, before it bites someone later.

I wrote that the strip shares a box with "the search field above and the content column below". That is true on `master`, but #9639 — my other open PR on this file — narrows the search field:

```diff
   .settings-search {
-    max-width: var(--component-max-width);
+    width: 100%;
+    max-width: 480px;
```

So once that lands, the search field is 480px while the tab strip and the cards stay at `--component-max-width`. Visible from roughly 1024px up; at 390px both are viewport-bound, so nothing changes there.

This does not affect what this PR does: the icon-only stretch is bounded by `--component-max-width` and aligned to `.tab-content`, which #9639 does not touch. The alignment argument holds on the column and the cards — only the "search field" half of my wording goes stale. I have updated the PR description accordingly.

Worth flagging because the two branches merge cleanly — the hunks are in different parts of the file, so git will not warn about it. If you would rather the search field stayed at `--component-max-width`, that is a call for #9639, and I am happy to drop that part there instead.

johannesjo:

Thanks for the updates, and for the measurements – they made this quick to check. Re-checked the 3 items from the earlier review at `8ff6e30`:

- ✅ **Rework around `justify-content: center`** – done at `src/app/pages/config-page/config-page.component.scss:64`, with `flex-grow: 0 !important` restored at `:74`, so the strip now uses the same pairing as `daily-summary`.
- ➖ **Whether to hand it off instead** – no longer applies; you carried it.
- ⚠️ **Nit: `flex-grow: … !important` is the wrong tool** – partly. `justify-content: flex-start` is gone and the outer `flex-grow: 1 !important` is back to `0`, so the "restates Material's default" half is fixed. But `:84` adds a new `flex-grow: 1 !important` inside the `mq(md, max)` block, and it only needs the `!important` because the `flex-grow: 0 !important` at `:74` stayed. Both can go: with emulated encapsulation this block compiles to `.page-settings[_ngcontent-x] .settings-tabs[_ngcontent-x] .mat-mdc-tab`, which already out-specifies Material's stretch rule (`.mat-mdc-tab-group.mat-mdc-tab-group-stretch-tabs > .mat-mdc-tab-header .mat-mdc-tab`), so a plain `flex-grow: 0` still wins in the labelled band and the `mq` block's plain `flex-grow: 1` wins in the icon band on source order. `mat-stretch-tabs="false"` on the `mat-tab-group` (`config-page.component.html:70`) is the tidier version of the same thing, but it isn't required. Nit either way – not a reason to hold the PR.

### Questions

- I can't reproduce **`48 → 160px at 1024`** from the repo's own tokens. `--component-max-width` is `800px`, and at 1024 the header has only ~792px anyway (1024 minus the 200px `--side-nav-width` minus `.page-wrapper`'s 2×16px), so the cap never binds and the six icon tabs land at ~132px each. I can get close to your number only by also collapsing the sidenav to the 56px rail *and* being on `body.isIPad` (which raises the cap to `1000px` from 1024 up) – and even that computes to ~156px. What was the setup: sidenav state, and web or iPad?
- Following from that: the description says the earlier objections all land on the labelled band, but the 1024 one was specifically "six 48px icons become six ~120px blocks each holding one 24px icon". With the default sidenav this lands at ~132px at 1024, i.e. effectively that same row, now bounded by the content column instead of the header – the bounded stretch only diverges from the original `flex-grow: 1` above roughly 1032px of viewport. Have I got that wrong, or does the 1024 case still stand as it was?

The delta itself looks sound: `justify-content` is inert once the labelled row overflows (matches your runtime swap test), and the explicit `width`/`max-width` on Material's `#tabListInner` doesn't disturb the pagination math – `_checkPaginationEnabled` compares its `scrollWidth` against the tab header, and six 48px tabs can't exceed the 800px cap.

---

johannesjo:
Thanks for finalising this. I want to take the `justify-content: center` half and leave the stretch.

The bounded stretch runs into the same objection I raised in August: at a 1024 viewport the header only gets ~792px, so the 800px cap never binds and the six icon tabs land at ~132px each – the wide slab I was trying to avoid. Centring alone is what I meant by option (1).

Could you drop the icon-band stretch (and with it the `flex-grow: 1 !important` and the outer `!important`), leaving just the `justify-content: center` change? That's the version I'll merge.

steindvart:
Thanks, done: the stretch is gone and the tabs are only centred.

Removing the outer `!important` on its own brings the stretch back (`stretchTabs` defaults to `true`), so I used the tab group's inputs instead: `mat-align-tabs="center"` and `[mat-stretch-tabs]="false"`. Both CSS overrides are removed.

johannesjo:
**✅ Ready to merge** – no blocking issues at `953cb88`. Merge confidence 92% – the layout was measured on the preview build of this head at 1920/1440/1280/1024/390px, but nothing automated covers it, so a future Material upgrade could move it silently.

Thanks for the updates. Re-checked the open items at `953cb88`:

- ✅ **Drop the icon-band stretch, keep the centring** – done. The diff is back to two files, +2/-5, with no `mq(md, max)` stretch block left anywhere.
- ✅ **Rework around `justify-content: center`** – `mat-align-tabs="center"` at `src/app/pages/config-page/config-page.component.html:72`. Material's own `[mat-align-tabs=center] > .mat-mdc-tab-header .mat-mdc-tab-labels` carries the identical declaration, and there is no competing `.mat-mdc-tab-labels` rule left in the component. Measured: the strip's centre matches `.tab-content`'s to within 0.5px at every width I tried.
- ✅ **Nit: `flex-grow: … !important` is the wrong tool** – both are gone. `[mat-stretch-tabs]="false"` at `:73` keeps `mat-mdc-tab-group-stretch-tabs` off the host, so Material's base `.mat-mdc-tab.mdc-tab { flex-grow: 0 }` applies unopposed; all six tabs compute `flex-grow: 0` with no `!important` on the page. Same mechanism `boards.component.html:2` and `issue-panel.component.html:3` use, though those spell it as the static attribute `mat-stretch-tabs="false"` rather than a binding – purely cosmetic, both resolve through `booleanAttribute`.
- ➖ **The two 1024px measurement questions** – no longer apply: the bounded stretch they were about is gone. At 1024 the icon band is back to six 48px tabs with no pagination.

Also checked the delta for anything the fix broke: neither input can leak to the other tab groups (`[mat-align-tabs=center]` uses a `>` combinator off the host and the stretch class is a host class), the global `_overwrite-material.scss` only sets colour tokens, the `min-width` 120/48 pair computes exactly as on master, and forcing overflow shows identical pagination geometry with `center` and `flex-start` – so the thresholds are unchanged. Nothing found.

One correction for the record, since it is in the commit message and will end up in history: removing the outer `!important` on its own would *not* have brought the stretch back **in this file**. With emulated encapsulation the block compiles to `.page-settings[_ngcontent-x] .settings-tabs[_ngcontent-x] .mat-mdc-tab` (0,5,0), which out-specifies Material's `.mat-mdc-tab-group.mat-mdc-tab-group-stretch-tabs > .mat-mdc-tab-header .mat-mdc-tab` (0,4,0); built the cascade in a browser to confirm. The nuance worth keeping is that it is not a general rule – `daily-summary.component.scss:28` writes `:host ::ng-deep .mat-mdc-tab` at (0,2,0), where the `!important` really is load-bearing. The input is the tidier shape either way, so this changes nothing about the diff; only the reasoning.

---

johannesjo:
Thank you very much! <3

PR #9639, opened 2026-08-19
style(settings): refine appearance controls

steindvart:
﻿## Problem

The Settings page gave disproportionate visual weight to the search field and
presented the appearance controls with inconsistent spacing and sizing. At
intermediate viewport widths the Dark Mode and Theme controls could overlap or
collapse into each other.

## Solution

- group Dark Mode, Theme and Wallpaper under an `Appearance` card that matches
  the other config sections, aligned on a shared label column
- make the settings search field more compact and vertically balanced
- mark the selected color mode the way the app already marks active items —
  highlighted in place without changing size: the `--state-selected` fill, a
  `--c-primary` icon, and a full-width underline like the settings tab strip's
  active indicator, so it follows light, dark and every custom theme
- drive the responsive breakpoint from a container query on the card's own
  content box, at the measured width where the grid stops fitting three
  labelled options
- add a settings-search target for the new `Appearance` heading, which is not a
  `config-section` and would otherwise be unsearchable
- remove trailing ellipses from `Install theme` and `Set wallpaper`
- update the Settings/Preferences and Theming wiki pages

`GCF.MISC.DARK_MODE` keeps its "Dark Mode" wording — see the review discussion.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [x] Documentation
- [x] Other (UI polish)

## Checklist

## Verification

Tests: theme-selector 7/7, config-page 6/6, settings-search 10/10 (one added,
covering the new `Appearance` target). `npm run checkFile` passes on every
changed file.

**Selection styling** — the underline is the non-colour cue. It is a 2px inset
shadow, so it takes no layout space. Plain `--brand` (what the tab indicator
uses) only reaches 2.6:1 against the selected fill and the group border on the
light card, so it is mixed 80/20 with `--text-color`:

| theme | underline vs. fill | underline vs. border | label contrast |
| --- | --- | --- | --- |
| Default light | 3.41:1 | 3.23:1 | 11.19:1 |
| Default dark | 5.31:1 | 4.99:1 | 9.11:1 |
| Dracula (forces dark, group disabled) | 4.84:1 | 4.47:1 | dimmed |

Only the icon takes the `--c-primary` accent; the label keeps `--text-color`.
A checkmark was not used because it needs its own width inside each segment —
either resizing the content on every switch or reserving blank space on every
option — which crowds the group on narrow screens.

**Responsive** — container content-box width vs. layout, `spill` is how far a
toggle's content overruns its own cell:

| content box | 328 | 368 | 388 | 398 | 402 | 468 | 559 | 766 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| layout | stack | stack | stack | stack | grid | grid | grid | grid |
| spill | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 |

The measured floor is a 388px content box — below that the label column is at
its 120px minimum and the group can no longer fit three labelled options — so
the query stacks at 400px. A 900px viewport with the sidebar visible gives a
591px container, which now stays in the grid layout like every other section.


johannesjo:
Thanks for this — the desktop result is a real improvement. Grouping the three controls into a card that matches the other config sections, aligning the label column, and calming down the search field all read better than what's on master today.

I checked the branch out and rendered it against master with Playwright at 8 viewport widths × 3 modes. Two things break at narrow widths, plus some smaller cleanups.

## Blocking

### 1. The selection checkmark is clipped on mobile

`theme-selector.component.ts` places the 20px check icon 24px to the left of the content box:

```scss
.dark-mode-toggle__selection {
  position: absolute;
  inset-inline-start: calc(-20px - var(--s-half));
}
```

Material's `.mat-button-toggle-group` sets `overflow: hidden`, and `.mat-button-toggle-appearance-standard .mat-button-toggle-label-content` only has `padding: 0 12px`. So as soon as the button roughly fills its cell, the check escapes the group and gets cut off.

Measured offsets from the group's inner left edge (icon is 20px wide):

| viewport | container width | System check | Dark check | Dark toggle starts at |
| --- | --- | --- | --- | --- |
| 360 | 327 | **-10 → 10** (half clipped) | 96 | 98 (**sits on the divider**) |
| 640 (sidebar visible) | 331 | **-10 → 10** | 98 | 99 |
| 420 / 700 | ~390 | 0 → 20 (flush against the border) | 126 | 118 |
| 900 | 591 | 34 → 54 ✓ | ✓ | ✓ |
| 1280 | 798 | 34 → 54 ✓ | ✓ | ✓ |

At 360px, selecting **System** shows half a checkmark clipped by the group border, and selecting **Dark** or **Light** draws the check on top of the divider line so it reads as belonging to the *previous* button. Master renders it correctly at 360px.

The goal in the commit message — center the label without shifting the unselected options — is better served by keeping the icon **in flow** and toggling `visibility` rather than `opacity`. The space is reserved either way, nothing shifts when the selection moves, and nothing overflows the group.

### 2. `@container (max-width: 600px)` fires roughly 300px too wide

The container measures *viewport − sidebar − card padding*, not the viewport. Measured `.theme-selector-container` widths:

| viewport | container |
| --- | --- |
| 1280 | 798 |
| 900 | **591** |
| 760 | 451 |

So at a 900px viewport the whole Appearance block drops into the stacked layout while every other settings section stays in its desktop form — on a common laptop / half-screen window. And the grid clearly still fits at 591px: `minmax(120px, 1fr) minmax(0, 3fr)` with a 16px gap gives a 146px label and a 438px control.

Switching from `@media` to `@container` is the right call, but the threshold needs to come down with it — somewhere around 420–450px is where the grid actually stops working.

## Should fix

### 3. The added `aria-label` bindings don't do anything

`MatButtonToggle` declares `@Input('aria-label')` and forwards it to the inner `<button>`, while its host bindings set `[attr.aria-label]: 'null'`. Writing `[attr.aria-label]` bypasses the input and then gets nulled on the host — I measured `aria-label` as `null` on both the host element and the inner button for all three toggles.

No accessibility regression, because the names were already correct (`mat-icon` is `aria-hidden` by default, so the check icon doesn't leak in):

```
- radiogroup "Color mode selection":
  - radio "System" [checked]
  - radio "Dark"
  - radio "Light"
```

So these are 7 lines of template noise — worth deleting rather than converting to `[aria-label]`.

### 4. Docs still reference the old button label

`THEME_INSTALL_BUTTON` lost its ellipsis, but three places still tell users to click "Install theme…":

- `docs/theming-contract.md:7`
- `docs/theming-contract.md:124`
- `docs/wiki/3.09-Theming.md:37`

### 5. "Appearance" isn't searchable

`settings-search.util.ts` documents the rule directly above `CUSTOM_TARGETS`:

> Anything added to the config page that isn't a `config-section` needs a row here or it stays unsearchable.

The new group heading has no entry, so typing "Appearance" in settings search finds nothing — while the updated wiki line now names the Appearance group as one of the searchable non-collapsible settings. Either add a target for `T.GCF.MISC.APPEARANCE` (pointing at `.dark-mode-select` or a new `.appearance-section` selector) or drop the name from the doc. The existing three targets still resolve, so jump-to for color mode / theme / wallpaper is unaffected.

## Minor

- **Dead CSS in the container block:** `mat-form-field { max-width: none }` (the base rule no longer sets a `max-width`) and `.dark-mode-toggle { width: 100% }` (already 100% in the base rule).
- **The ellipsis removal is arguably backwards.** Both actions open a picker/dialog, which is exactly what a trailing "…" signals. Dropping it loses information rather than noise.
- **"Dark Mode" → "Color mode" is English-only.** Correct per the project rule to touch only `en.json`, but the other 27 locales keep "Dark Mode" / "Modo Oscuro" / "Tmavý režim" until translators catch up. Worth a conscious call on whether the rename earns that drift.
- **`.appearance-section { overflow: hidden }`** — `config-page.component.scss` already documents that `overflow` on ancestors caused focus-induced scroll jumps (hence `overflow: visible !important` on the tab body wrapper, with a comment explaining why). Worth confirming this is only needed to clip to the card's rounded corners.
- **Branch is 3 commits behind master**, and several of the 7 commits are fixups of each other ("prevent appearance controls from overlapping" → "center color mode selection indicator" → "keep color mode content centered"). Worth squashing.

## What I checked

- `npm run checkFile` on the changed `.ts` — pass; `npm run lint:css-vars` — pass (all custom properties resolve).
- `theme-selector.component.spec.ts` 7/7, `config-page.component.spec.ts` 6/6.
- Every Material token used (`--mat-button-toggle-height`, `--mat-button-toggle-shape`, `--mat-button-toggle-label-text-size`, `--mat-button-outlined-container-height`, `--mat-form-field-container-height`, `--mat-form-field-container-vertical-padding`, `--mat-form-field-outlined-container-shape`) exists in `@angular/material@21.2.12`.
- The `label` → `placeholder` swap on the search field keeps the documented contrast tradeoff intact: both resolve to `--mat-sys-on-surface-variant` by default, so the 4.52:1 figure in the comment still holds, and `--no-label` makes the placeholder visible while empty.
- Rendered branch and master at 360 / 420 / 520 / 640 / 700 / 760 / 900 / 1280 × system / dark / light for the geometry above.

Once 1 and 2 are sorted this is a nice cleanup — the 1280px layout in particular is a clear step up.

johannesjo:

Follow-up pass on `63ebd29`. Nothing has been pushed since my review of 2026-08-21, so every item there still stands as written and I won't repeat the list. Four things have turned up since.

### Blocking

- **Fix the checkmark (item 1) before the breakpoint (item 2) – they interact.** The grid gives the toggle group roughly 75% of the container, so each of the three toggles is about a quarter of that. Lowering the `@container` threshold to the 420–450px I suggested leaves the toggles at ~108px through the 420–600px range, which is *below* the ~117px point where `inset-inline-start: calc(-20px - var(--s-half))` pushes the check outside the group's `overflow: hidden` – so fixing item 2 on its own makes item 1 visibly worse across a band where it currently renders fine. Fixing item 1 the way I suggested (keep the icon in flow, toggle `visibility` instead of `opacity`) removes the overflow entirely and makes the interaction moot, so please do that one first.

### Should fix before merge

- **`src/assets/i18n/en.json` (`GCF.MISC.DARK_MODE`) / `src/app/features/config/settings-search.util.ts:32`** – the rename quietly costs a search term, which is a cost beyond the locale drift I mentioned. `CUSTOM_TARGETS[0].labelKey` is `T.GCF.MISC.DARK_MODE`, and `searchSettings` matches with `translate(key).toLowerCase().includes(q)`. Once that string is `"Color mode"`, typing `dark` or `dark mode` in Settings search returns nothing at all: no other global setting's English label contains "dark" (`GCF.MISC.IS_DARK_MODE` and `PS.TOGGLE_DARK_MODE` are declared in `t.const.ts` but rendered nowhere, and the `F.PROJECT.FORM_THEME.*` labels are project settings, not this page). On master the same query jumps straight to `.dark-mode-select`. Either keep the label, or give `SettingsSearchTarget` an alias/extra-terms field so the target still matches the old wording – worth solving together with the missing "Appearance" target from item 5.

### Nits

- **`config-page.component.scss` (`.appearance-section`)** – I asked you to confirm this one; having looked myself, `overflow: hidden` clips nothing and can just go. `.theme-selector-container` insets everything by `var(--s2)` on all four sides so no child reaches the card's rounded corners, the one child that does overflow (the check icon) is already clipped by the toggle group's own `overflow: hidden`, and the `mat-select` panel and `matTooltip` render in CDK overlays outside the section.
- **The ellipsis removal – withdrawn.** I flagged it as arguably backwards; on checking, `en.json` has 26 strings ending in "…" and only four of them are action labels (`THEME_INSTALL_BUTTON`, `WALLPAPER_BUTTON`, `USER_PROFILES.MANAGE_PROFILES`, `F.TASK.CMP.CUSTOM_ESTIMATE`) – every other one is a loading/progress string or an input placeholder. There is no "… means opens a dialog" convention here to lose, so dropping them is the more consistent choice. The three doc references in item 4 still need updating.
- **The branch is now 189 commits behind `master`**, not the 3 I saw in August. Every hunk you touch is still unchanged on master and GitHub reports the PR as mergeable, so nothing has broken – but it wants a rebase, and the seven fixup-style commits are still worth squashing.

---

johannesjo:
Decision on the rename: let's keep `GCF.MISC.DARK_MODE` as "Dark Mode" and drop that hunk — the new `APPEARANCE` key and the rest of the grouping stay.

Two concrete costs for wording nobody has reported as a problem: that key doubles as the Settings-search label (`settings-search.util.ts:32`), so "Color mode" silently removes "dark" and "dark mode" as English search terms; and it leaves the other 27 locale files sitting on their "Dark Mode" equivalent until translators catch up. Please revert `DARK_MODE_ARIA_LABEL` with it. That also settles the alias/extra-terms idea I floated — not needed; the missing "Appearance" search target from item 5 still is.

The rest of the Appearance grouping is wanted. Fix the checkmark clipping first, then the container threshold, for the reason in the follow-up comment.

Thanks for the work on this.

steindvart:
Thanks for the review — the measured tables made this straightforward to reproduce.

**1. Checkmark clipping.** Dropped the custom indicator entirely instead. Selection is now the `--state-selected` fill plus a `--c-primary` icon, matching how the settings tab strip marks its active tab. Nothing is positioned out of flow any more, so the clipping is gone by construction — and no blank column is reserved in front of the unselected labels either.

`visibility` would not have worked here: the global `body.isMaterialSymbolsLoaded mat-icon` guard in `_overwrite-material.scss:349` wins on specificity and leaves the icon painted. `focus-mode-main.component.html:15` documents the same trap.

Only the icon takes the accent — tinting the label put it at 2.38:1 against the light card, so it keeps `--text-color`.

**2. Container threshold.** `600px` → `400px`. One unit correction worth flagging: the query resolves against the container's *content box*, not the `clientWidth` your tables measure — 32px of padding apart. The measured floor is a 388px content box, which is the 420px `clientWidth` your 420–450 estimate landed on. A 900px viewport with the sidebar visible now stays in the grid layout.

**3–5, 7.** The three toggle-level `aria-label` bindings deleted; all three doc references updated; added a `T.GCF.MISC.APPEARANCE` search target with a spec case; dead CSS and `.appearance-section { overflow: hidden }` removed.

**Rename** reverted — `GCF.MISC.DARK_MODE` and `DARK_MODE_ARIA_LABEL` are back to their original wording, only the new `APPEARANCE` key stays.

**8.** The branch is current with master. I have not squashed the history — happy to, if you would rather not squash-merge.

johannesjo:

Thanks for the updates, and for the content-box correction – that one was mine to get wrong. Re-checked all ten items from the earlier passes at `cbbb8c9`:

- ✅ **Checkmark clipped on mobile** – gone by construction. `.dark-mode-toggle__selection` and its absolute positioning are out of `theme-selector.component.ts`, `position: relative` is off `.dark-mode-toggle__content`, and nothing inside the group is out of flow any more. Your reason for declining `visibility` holds, though not for the selector you cited: `body.isMaterialSymbolsLoaded mat-icon` is only `(0,1,2)` and would *lose* to a component-scoped rule. It is the sibling selector `body.isMaterialSymbolsLoaded .mat-icon` at `(0,2,1)` that wins, because `MatIcon` host-binds `class="mat-icon"`. Worth knowing for next time, because a `(0,3,0)` form like `.dark-mode-toggle__selection.mat-icon` would have beaten it.
- ✅ **`@container` threshold** – `400px`, and you are right that size features resolve against the content box. With `var(--s2)` padding either side that puts the flip at roughly `432px` `clientWidth`, inside the 420–450 band I asked for; `591px` at a 900px viewport stays comfortably in the grid layout.
- ✅ **Inert `[attr.aria-label]` bindings** – all three deleted. The group-level label is untouched and the per-option accessible names still come from the visible label text.
- ✅ **Doc references to "Install theme…"** – all three updated, and they match the shipped label since the ellipsis removal itself stayed.
- ✅ **"Appearance" searchable** – target added in `settings-search.util.ts:31` with a spec case, and `.appearance-section` resolves against `config-page.component.html:90`. It neither shadows nor drops the existing `DARK_MODE` target: `dark` still returns only that row.
- ✅ **Rename reverted** – `DARK_MODE` and `DARK_MODE_ARIA_LABEL` are back to their original wording, `APPEARANCE` is the only new key, no `COLOR_MODE` left anywhere, and the wiki line reads "(dark mode, theme, and wallpaper)" again.
- ✅ **Dead CSS in the `@container` block** – both rules gone, and the base rules still cover what they were duplicating.
- ✅ **`.appearance-section { overflow: hidden }`** – removed.
- ➖ **Ellipsis removal** – withdrawn; the doc references it kept alive are covered above.
- ➖ **Branch behind master / squash** – the branch is current with `master`. Nothing needed from you on the history; how it lands is a merge-button call and I've put the question in front of the maintainer.

### Still open

- **`src/app/core/theme/theme-selector/theme-selector.component.ts`** – dropping the indicator entirely leaves the selected mode marked by colour alone, and faintly. `--state-selected` is `rgba(var(--ink-on-channel), 0.1)` (`src/styles/_css-variables.scss:573,580`), so the selected segment sits at about 1.25:1 against its neighbours on light (`#e6e6e6` on `#ffffff`) and 1.37:1 on dark – well under the 3:1 WCAG 1.4.11 asks of a state indicator – with the 20px icon at the 2.4:1 you measured as the only other cue.

  This is a regression against `master`, not only against your previous head: `master` sets no `hideSingleSelectionIndicator` and Material 21 defaults it to `false`, so today the checked option gets a real check glyph plus the 30px indent Material reserves for it. The tab strip you mirror is not colour-only either – `config-page.component.scss:73` gives it the same 10% fill, but `src/styles/components/_overwrite-material.scss:36-38` keeps Material's active underline on top of it, so that pattern is fill *plus* a shape cue.

  Cheapest fix that keeps the layout you want: put the check icon back **in flow** and toggle `opacity` rather than `visibility`. The space is reserved either way so nothing shifts when the selection moves, nothing can overflow the group, and it sidesteps the `.mat-icon` specificity trap entirely. Dropping `[hideSingleSelectionIndicator]` and letting Material draw its own indicator works too. Also worth a look: `src/app/ui/segmented-button-group/` is the app's own answer to this question, and it marks the active segment with a `--c-primary` border plus elevation rather than a fill.

### Nits

- **`theme-selector.component.ts`** – `.dark-mode-toggle .mat-button-toggle-checked mat-icon` hangs on a Material internal state class, which the repo's styling rule steers away from. It works, and there is no icon-only token to use instead, so this is optional – but a component-owned class (`[class.is-selected]` on each `mat-icon`, the way this PR had it before the push) does the same job without reaching into Material's internals.

---

johannesjo:
The Appearance grouping is good and everything from my last comment landed – thank you.

One thing I can't take as-is: the selected colour mode is now distinguished by fill and icon tint alone (~1.25:1 light, ~1.37:1 dark), where master shows Material's check glyph. That needs a non-colour cue back – the checkmark, or a border/weight difference if you'd rather keep the cleaner look.

No need to squash – I'll squash-merge under the PR title.

steindvart:
Thanks for the review!

I added a non-colour cue: the selected option now has a full-width underline, the same indicator the settings tab strip uses for its active tab. It's drawn as an inset shadow, so the layout doesn't change. The colour is `--brand` mixed 80/20 with `--text-color`, which keeps it above 3:1 against both the fill and the group border (3.4:1 light, 5.3:1 dark).

I didn't bring the checkmark back because it needs its own width inside each segment. Either the content resizes whenever the selection changes, or every option has to reserve blank space for it. Both hurt the layout on narrow screens, and that extra width is exactly what caused the crowding in the earlier versions. Instead I went with the pattern the app already uses for this kind of choice: the active item is highlighted in place, without changing size. That's how the settings tabs and the sections in the left menu (Today, Inbox, …) are marked.

The selected styling now keys off a component-owned `is-selected` class instead of `mat-button-toggle-checked`.

johannesjo:
**✅ Ready to merge** – no blocking issues at `1400c36`. Merge confidence 88% – the underline is checked analytically (Material's own CSS, paint order, the contrast table recomputed from the tokens) rather than rendered on screen.

Thanks for the updates. Re-checked the 2 items still open at `cbbb8c9`:

- ✅ **Selected mode marked by colour alone** – fixed in `theme-selector.component.ts`: `.dark-mode-toggle mat-button-toggle.is-selected` now carries `box-shadow: inset 0 -2px 0 color-mix(in srgb, var(--brand) 80%, var(--text-color))`. A full-width bar is a shape cue, which is the "border/weight difference" I offered, so this settles it. Three things I checked rather than took on trust, since an inset shadow on a Material host is easy to get wrong: nothing competes for the declaration at all – Material's two `box-shadow: none` rules are on `.mat-button-toggle-standalone...` and `.mat-button-toggle-group-appearance-standard`, neither of which matches a toggle inside a group; the host's children are all transparent (`.mat-button-toggle-button { background: none }`, the focus overlay at `opacity: 0`, the ripple), so the band is not painted over; and the group's own `overflow: hidden` plus `border-radius` clips it at the corners exactly as your comment says. Forced-colors is unaffected either way – Material keeps drawing its own checked cue through `.mat-button-toggle-checked .mat-button-toggle-focus-overlay`.
- ✅ **`mat-button-toggle-checked` state class** – gone (the only occurrence left is the comment at line 310), replaced by `[class.is-selected]` on each toggle. It cannot drift from the checked segment: the group's `[value]` and all three bindings read the same synchronous `globalThemeService.darkMode()` signal in one change-detection pass, and `.is-selected` has no other definition anywhere in `src/`.

I also recomputed your contrast table from the tokens rather than taking the numbers as given – 3.40/3.24 light and 5.31/4.97 dark against your 3.41/3.23 and 5.31/4.99, so it reproduces to rounding. And your reason for not bringing the checkmark back holds: reserving a glyph's width in every segment is what crowded the group in the first place, and "highlight in place, do not resize" is what the tab strip and the side nav already do. `--brand` is the same token `--mdc-tab-indicator-active-indicator-color` resolves to (`_overwrite-material.scss:19`), so the bar tracks custom themes like the rest of the card.

Nothing new in the delta – one file, `color-mix()` already has ~200 uses in `src/`, prettier is clean at `printWidth: 90`, and CI is green.

---

johannesjo:
Thank you very much! <3

Issue #5501, opened 2025-11-13
💡 Allow "Time Spent" to automatically equal "Estimate" when completing a task if "Time Spent" is currently empty


FriendlyGremlin:
### Problem Statement

Currently, if you don't use the time tracking feature, the "Time Spent" stays empty. When checking off the task, "Time Spent" must be manually entered, which is cumbersome as the menu must first be opened and then the time entered. This is especially noticeable for tasks that consist of a lot of small subtasks (e.g. each 1-5 minutes long), as then all of them need to be entered manually.

### Possible Solution

The settings should make it possible to enable a feature wherein, if "Time Spent" is currently empty / no time has so far been tracked, and the user checks off / completes a task, the "Time Spent" should automatically be set to the time estimate. Meaning, if I estimate a task to take 30 minutes, and I complete it without using time tracking or setting any "Time Spent" manually, then "Time Spent" should automatically default to 30 minutes.

William-A-B:
I like this idea, it would be particularly useful for those tasks which might be done away from a computer/phone where it's less simple to track the time as you're completing the task, but still spent the estimated time on it.

maciejmatczak:
Recently a new short syntax was added!

You can avoid going into a menu, stay at keyboard and type in `30m/` :).

William-A-B:
> Recently a new short syntax was added!
> 
> You can avoid going into a menu, stay at keyboard and type in `30m/` :).


Ah interesting I didn't know that thanks! That's certainly useful. Although still an extra step, but maybe that's the better way of doing it for the odd occasion that some tasks you want to specify you didn't actually spend time on but still completed. Which wouldn't be as simple with the estimated time taken as actual time.

FriendlyGremlin:
> Recently a new short syntax was added!
> 
> You can avoid going into a menu, stay at keyboard and type in `30m/` :).

Didn't know that, either. While I agree it's faster and is a cool feature IF you actually want to enter a time different from the estimate, it's still more selecting and typing than simply ticking off the task and having the spent time auto-filled. Especially on mobile, this takes ages longer.

For most tasks (especially lots of small, habit-building ones. Think: "brush your teeth" or something along those lines), I think it's better to simply estimate the time they take correctly once, and then saving all future keystrokes & mouse clicks in return.

MazenMohamed203:
yeah this would be very good as a check box in the settings for recurring tasks.

MazenMohamed203:
[...]

PR #9646, opened 2026-08-19
feat: add Auto-Fill Time Spent community plugin

MazenMohamed203:
When a task with a Time Estimate is marked complete and Time Spent is zero,
the plugin automatically sets Time Spent equal to the estimate.

## Problem

Users who don't use the built-in timer (e.g. for off-computer activities like
running, workouts, or errands) are left with Time Spent = 0 even when they had
an estimate set. 

Requested in: https://github.com/super-productivity/super-productivity/issues/5501

## Solution

Added the **Auto-Fill Time Spent** community plugin to `community-plugins.json`.

The plugin registers a `taskComplete` hook and sets `timeSpent = timeEstimate`
only when both conditions are true:
- `timeEstimate > 0` (the task has an estimate)
- `timeSpent === 0` (no time was tracked)

Tasks where the timer was actually used are never modified.

Plugin repo: https://github.com/MazenMohamed203/auto-fill-time-spent

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [x] Other (please describe) — adding a plugin to the community plugins list

## Checklist

- [x] I have included relevant changes to the documentation/wiki
- [ ] I have run `npm run checkFile` on changed `.ts`/`.scss` files *(no .ts/.scss files changed)*
- [ ] I have added tests for my changes *(not applicable — JSON list addition only)*
- [x] My commit messages follow the Angular format (`feat: add Auto-Fill Time Spent community plugin`)

### DIFF
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -45,7 +45,7 @@
     "url": "https://github.com/MazenMohamed203/auto-fill-time-spent",
     "author": "MazenMohamed203",
     "authorUrl": "https://github.com/MazenMohamed203",
-    "stars": 2
+    "stars": 5
   },
   {
     "name": "Obsidian Integration",
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -40,12 +40,12 @@
     "stars": 0
   },
   {
-  "name": "Auto-Fill Time Spent",
-  "shortDescription": "Automatically sets Time Spent equal to Time Estimate when completing a task that has no tracked time. Perfect for off-computer activities.",
-  "url": "https://github.com/MazenMohamed203/auto-fill-time-spent",
-  "author": "MazenMohamed203",
-  "authorUrl": "https://github.com/MazenMohamed203",
-  "stars": 1
+    "name": "Auto-Fill Time Spent",
+    "shortDescription": "Automatically sets Time Spent equal to Time Estimate when completing a task that has no tracked time. Perfect for off-computer activities.",
+    "url": "https://github.com/MazenMohamed203/auto-fill-time-spent",
+    "author": "MazenMohamed203",
+    "authorUrl": "https://github.com/MazenMohamed203",
+    "stars": 2
   },
   {
     "name": "Obsidian Integration",
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -40,6 +40,14 @@
     "stars": 0
   },
   {
+  "name": "Auto-Fill Time Spent",
+  "shortDescription": "Automatically sets Time Spent equal to Time Estimate when completing a task that has no tracked time. Perfect for off-computer activities.",
+  "url": "https://github.com/MazenMohamed203/auto-fill-time-spent",
+  "author": "MazenMohamed203",
+  "authorUrl": "https://github.com/MazenMohamed203",
+  "stars": 1
+  },
+  {
     "name": "Obsidian Integration",
     "shortDescription": "Provides 2-way sync between SP tasks and Obsidian tasks.",
     "url": "https://codeberg.org/sinenomine/sp-obsidian/",
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -45,7 +45,7 @@
     "url": "https://github.com/MazenMohamed203/auto-fill-time-spent",
     "author": "MazenMohamed203",
     "authorUrl": "https://github.com/MazenMohamed203",
-    "stars": 5
+    "stars": 6
   },
   {
     "name": "Obsidian Integration",


johannesjo:
Thanks for building this and for opening your first PR here! 🎉 The demand is real ([#5501](https://github.com/super-productivity/super-productivity/issues/5501) is open with 4 👍 and several distinct supporters), nothing in core covers it, and a community plugin is the right venue for it rather than another core setting. So I'd like to list it — but two things need fixing first, and the second one is the important one.

---

### 1. Formatting breaks Prettier

The new entry is indented 2 spaces where every other entry uses 4. Editing via the GitHub web UI bypasses the `pretty-quick` pre-commit hook, so it slipped through:

```
$ npx prettier --check src/assets/community-plugins.json
[warn] src/assets/community-plugins.json
[warn] Code style issues found in the above file.
```

`npx prettier --write src/assets/community-plugins.json` fixes it.

(No concern about *where* in the list you inserted it — the list is alpha-sorted at runtime in `plugin-management.component.ts:131-135`, so position in the JSON doesn't matter.)

---

### 2. The plugin writes a derived field, so the time won't actually show up anywhere

This is the blocker. `plugin.js` does:

```js
await PluginAPI.updateTask(taskId, { timeSpent: task.timeEstimate });
```

But `Task.timeSpent` is **a cached rollup of `timeSpentOnDay`, not a source of truth**. The app's own manual "time spent" editor writes the day bucket and lets the reducer derive the total (`dialog-time-estimate.component.ts:105-109`).

`handleUpdateTask` only recalculates when `timeSpentOnDay` is present in the changes:

```ts
// task-shared-crud.reducer.ts:854
taskState = timeSpentOnDay
  ? updateTimeSpentForTask(taskId, timeSpentOnDay, taskState)
  : taskState;
```

…and otherwise applies `changes` verbatim via `taskAdapter.updateOne`. So the plugin leaves the task with `timeSpent: 30m` and `timeSpentOnDay: {}`. Consequences:

- **The task list shows the number** (`task.component.html:72-74` renders `t.timeSpent`) — which is why it looks like it works.
- **Worklog, worklog export and the archive show 0.** All of them build purely from `timeSpentOnDay` (`worklog.service.ts:185-196`, `map-archive-to-worklog.ts:53-102`). The time is invisible in every report.
- **Subtasks never roll up to the parent.** `reCalcTimeSpentForParentIfParent` rebuilds the parent from the children's `timeSpentOnDay` (`task.reducer.util.ts:55-72`), and per the gate above it isn't even invoked here — so the parent stays at 0 permanently.
- **The value is dropped if the task is ever re-opened and tracked** — `updateTimeSpentForTask` recomputes `timeSpent` from `timeSpentOnDay` (`task.reducer.util.ts:240`).

The subtask point is the one I'd highlight: #5501's stated motivation is *"tasks that consist of a lot of small subtasks (e.g. each 1-5 minutes long)"* — which is exactly the case that silently does nothing.

The fix is small — write the day bucket, and set the derived total yourself since `updateTask` applies changes verbatim:

```js
const todayStr = /* local YYYY-MM-DD */;
const spentOnDay = { ...(task.timeSpentOnDay || {}), [todayStr]: task.timeEstimate };
await PluginAPI.updateTask(taskId, {
  timeSpentOnDay: spentOnDay,
  timeSpent: Object.values(spentOnDay).reduce((a, b) => a + b, 0),
});
```

Worth deciding deliberately which day to attribute it to (completion day is probably right), and please test against a **parent task with subtasks** before we list it.

---

### What checks out ✅

- Repo is real, MIT-licensed, with a `v1.0.0` release carrying the `auto-fill-time-spent.zip` asset.
- Manifest is valid, and `PluginAPI.Hooks.TASK_COMPLETE` plus the `{ taskId, task }` payload match `packages/plugin-api/src/types.ts:174-177`.
- **No feedback loop and no cross-device amplification** — the hook effect filters on `changes.isDone === true` and injects `LOCAL_ACTIONS` (`plugin-hooks.effects.ts:58,69`), so the plugin's own update can't re-trigger it and incoming sync ops can't fire it on every device. Nice.
- The `timeEstimate > 0 && timeSpent === 0` guard correctly leaves genuinely-tracked tasks alone.

### Minor

- `"permissions": ["updateTask"]` is a no-op — only `nodeExecution` and `http` are actually enforced (`plugin-security.ts:252,265`). Harmless, just not doing anything.
- `"minSupVersion": "14.0.0"` — worth double-checking that `taskComplete` and `updateTask` really existed that far back.

Fix the indentation and the `timeSpentOnDay` write and I'm happy to merge this. 👍

MazenMohamed203:
Thank you for your thorough analysis! I fixed the indentation and added your fix.

I also tested the plugin with subtasks and parent tasks. To ensure it works perfectly with SP's native `reCalcTimeSpentForParentIfParent` rollup, the plugin now completely ignores parent tasks and instead distributes the auto-filled time directly to the uncompleted subtasks.

I made one other addition to protect against accidentally checking a task: it now logs the IDs of auto-filled tasks to memory (which clears on restart) so it can reset the time spent if you uncheck a task. This doesn't affect manually tracked time or manually entering the time spent with `30m/`.

Let me know if there's anything else needed!

johannesjo:

Thanks for the rework! Both items from the earlier review check out: the entry is Prettier-clean now, and writing `timeSpentOnDay` really does reach the worklog and roll up to the parent (`updateTimeSpentForTask` → `updateParentTimeSpentIncremental`, `task.reducer.util.ts:229-262`). `minSupVersion: "15.0.0"` is fine too – `getTasks` shipped in v14.0.0 and the `anyTaskUpdate` hook in v14.1.0, so 15.0.0 is safely above both.

I read the plugin source at `4001ae5` (all 5 files, ~10 KB: `manifest.json`, `plugin.js`, `README.md`, `LICENSE`, `icon.svg`) plus the `v1.2.0` release zip, whose three files are sha256-identical to the repo. No network calls, no storage, no `eval` / `new Function` / dynamic `import`, nothing credential-shaped, repo is active and MIT. The new subtask and undo logic has two problems though.

### Should fix before merge

- **`plugin.js:18-21` – the "no tracked time" guard only looks at today's bucket.** `existingToday === 0` is true for any task that has tracked time on an *earlier* day, so the estimate is added on top instead of the task being skipped. No parent needed to hit it: 2h estimate, 1h55m tracked Monday, ticked Tuesday untracked → `{Mon: 1h55m, Tue: 2h}` = 3h55m total, plus a phantom 2h on Tuesday's worklog. With subtasks it compounds: subtask A finished Monday with 1h tracked, parent ticked Wednesday → A gets another 1h on Wednesday and the parent rolls up inflated. This contradicts the `shortDescription` in this PR, `README.md:35` ("never overwrites real tracked time"), and #5501's own wording ("if Time Spent is currently empty" – the *total*). Fix: keep the today check but also require the total to be zero, e.g. `Object.values(task.timeSpentOnDay || {}).reduce((a, b) => a + b, 0) === 0`.

- **`plugin.js:85-92` – the undo fires on *any* update of a not-done task, not on an uncheck.** Ticking a parent does not mark its subtasks done: `TaskService.setDone` updates one id (`task.service.ts:1226`) and the only cascade is bottom-up (`onAllSubTasksDone$`, `task-internal.effects.ts:37-70`). So after the parent fill those subtasks are still `isDone: false` and their `autoFilledTasks` entries stay live, while `anyTaskUpdate$` fires for every `updateTask` and for `moveSubTask` / `moveSubTaskUp|Down|ToTop|ToBottom` with `task` resolved from state (`plugin-hooks.effects.ts:261-307`). Tick parent P → subtask S gets +1h → then rename S, change its estimate or tag, or drag it to reorder → `task.isDone !== false` is false → `clearAutoFilledTime(S)` and the hour silently disappears. Worse, `plugin.js:46` subtracts `entry.amount` from whatever is in the bucket at that moment, so if the user *corrects* S's time in the time dialog (which dispatches `updateTask` with `timeSpentOnDay`, `dialog-time-estimate.component.ts:101-109`) the sequence is 20m entered minus 1h filled → `Math.max(0, …)` → 0 and the key is deleted: the user's own entry is destroyed, not just the plugin's addition.

  Note the `anyTaskUpdate` payload is `{ action, task, taskId, taskState }` and never carries `changes` (`plugin-hooks.effects.ts:302-307`) even though the type declares the field, so `changes.isDone === false` is not testable there. Two ways out: register the `taskUpdate` hook instead, which does pass `changes` (`plugin-hooks.effects.ts:161-168`) and lets you undo only on a real `isDone: false` transition; or only record an `autoFilledTasks` entry for tasks that were themselves done at fill time, keying the parent case under the parent's id.

  Related, worth checking while you are in there: the fill does not currently undo itself only because of dispatch ordering. Hook handlers are invoked synchronously inside the dispatch (`plugin-hooks.ts:_invokeWithTimeout`), so the echoed `anyTaskUpdate` runs *before* `plugin.js:32` records the entry. If that `await PluginAPI.updateTask(...)` ordering is ever refactored, the parent path would fill and instantly un-fill every subtask.

### Nits

- **`src/assets/community-plugins.json`** – `"stars": 2`, the repo reports 5. The value is only used for the ordering of the list (`plugin-management.component.ts:132`), so the effect is just that the entry sorts lower than it should.
- **`README.md`** – still describes 1.0: the flow diagram shows `updateTask({ timeSpent: timeEstimate })` and a `timeSpent === 0` guard, it calls the plugin "a single hook listener", and `README.md:103` lists an `updateTask` permission against `manifest.json`'s `"permissions": []`. Two claims are now false against the shipped code: `README.md:35` "never overwrites real tracked time" (see above) and `README.md:105` "No data is read beyond what is included in the `taskComplete` hook payload", while `plugin.js:68` and `:93` call `PluginAPI.getTasks()`, which returns every task in the store. It stays on the device, but it is a broader read than documented. The subtask distribution and the undo are not mentioned at all.
- **`plugin.js:6`** – `autoFilledTasks` is in-memory only, so after an app restart unchecking no longer removes the filled time. Your code comment scopes this deliberately, but since the README never mentions the undo, users will not know the asymmetry exists.
- **`plugin.js:8-13`** – `getTodayStr()` is the plain calendar date, while the timer books ticks to the *logical* day (`TaskService.addTimeSpent` defaults to `DateService.todayStr()`, shifted by the "start of next day" setting). With start-of-day at 03:00, time tracked at 01:30 lands on the previous day while the plugin reads and writes the current one. SP's own manual time dialog uses the calendar date too, so this is an app inconsistency the plugin inherits rather than a plugin bug – just worth knowing.

### Questions

- The parent branch fills *every* subtask whose today bucket is 0, done or not, and there is no `isDone` check at `plugin.js:66-74`. Your comment above describes it as filling "the uncompleted subtasks" – is filling already-completed subtasks intended (they may carry real tracked time from earlier days, which is the first item), or should it skip them?
- `permissions` is now `[]` while the plugin calls `updateTask` and `getTasks`. That is harmless – only `nodeExecution` and `http` are actually enforced – but was dropping the declaration deliberate, or a side effect of the rewrite? The README still documents the old value either way.

---

MazenMohamed203:
@johannesjo 
Thanks again for the analysis.

I did all the changes asked. 
and for the 2 questions:

1:When checking a parent, subtasks that were already individually completed are skipped. They were either auto-filled at their own completion time or intentionally left without time.

2:Deliberate. Only nodeExecution and http are enforced. README now accurately documents "permissions": [] and lists the actual API methods used (updateTask, getTasks).

johannesjo:
**🔍 Reviewed** – no blocking issues found at `e32eeeb`.

Thanks for the rework. Both items from the last pass are fixed, and I traced them through master rather than taking my own word for it:

- **The zero-time guard is the total now, not today's bucket** (`plugin.js:20-23` and `:29`). The case I described – 2h estimate, 1h55m tracked Monday, ticked Tuesday – returns at line 29 instead of stacking a phantom 2h on top, and the same helper gates the subtask path, so the compounding variant is covered too. `README.md:69` documents the reasoning.
- **The undo moved to `taskUpdate`, which is the hook that actually carries `changes`** (`plugin-hooks.effects.ts:116-172`). `TaskSharedActions.updateTask` still reaches the hook for every update, but `plugin.js:95` now discards everything except `changes.isDone === false`, so renames, tag edits, estimate edits and `moveSubTask*` can no longer reach the clear path. I checked the two ways that fix could have been hollow and both are clean: `plugin.js:118-121`'s exact-value guard really does match, because `updateTimeSpentForTask` stores your map verbatim (`task.reducer.util.ts:242-250`), so a Duration-dialog correction survives; and the hook is genuinely reachable, because `setUnDone` calls `update()`, which dispatches `updateTask` with `{isDone: false}` (`task.service.ts:1233`, `:534-540`). The ordering hazard I flagged last time is structurally gone as well – the fill's own echoed update carries no `isDone`, so it can't un-fill itself whatever the dispatch order.

Your two answers check out too: `plugin.js:66` does skip already-done subtasks, and `permissions: []` is right, since only `nodeExecution` and `http` are enforced.

I read the plugin source at `6a1c1cb` (all 5 files, ~14.5 KB) plus the `v1.3.0` release zip, whose 4 files are sha256-identical to the repo tree. No network calls, no storage, no `eval` / `new Function` / dynamic `import`, nothing credential-shaped, repo active and MIT. No `iFrame` is declared, so `plugin.js` runs in the app's renderer realm with full `window` access – normal for a hook-only plugin, and this one touches none of it. `minSupVersion: "15.0.0"` still holds: `taskUpdate`, `getTasks` and `updateTask` are all in the v15.0.0 API surface.

Only nits left.

### Nits

- **`plugin.js:105-129` – the parent undo also clears fills on subtasks you have since ticked yourself.** Parent P with subtasks S1 and S2, 1h estimate each. Tick P, both get 1h and both fills are recorded under P's id (`plugin.js:72`). Now tick S1 for real – its own `taskComplete` is a no-op, because `getTotalSpent(S1)` is already 1h (`plugin.js:29`), so there is no separate entry for it. Untick P and the loop still walks both fills; S1's bucket is untouched so it matches `fill.amount` exactly and `plugin.js:123` deletes it. S1 stays done with Time Spent 0, and re-ticking it won't refill, because it's already done. Cheap fix: skip a fill whose task is now `isDone` – the lookup at `plugin.js:110-112` already has the task in hand.
- **`README.md:93-110` – the install steps describe a button the app doesn't have.** They say to clone the repo, keep the folder intact, and click "Install from folder". Super Productivity's only install path is a ZIP upload: the file input is `accept=".zip"` and rejects anything else (`src/app/plugins/ui/plugin-management/plugin-management.component.html:38-44`, `plugin-management.component.ts:282-287`), under the label "Upload a plugin ZIP file to install it". To be clear, this isn't something you invented – the phrasing came from this repo's own older plugin docs, and two already-listed plugins still repeat it. But your `v1.3.0` release ships exactly the right artifact, so pointing at `auto-fill-time-spent.zip` and Settings → Plugins → the upload button would save every user a detour. While you're in the file: `README.md:99-101` lists three files where the zip ships four, and `README.md:143` scopes `getTasks()` to "parent task completion / undo" when the undo path calls it for leaf undos too.
- **`src/assets/community-plugins.json` – `"stars": 5`, the repo reports 6 now.** Self-reported and rendered as a badge (`plugin-management.component.html:277-284`), so this is cosmetic. Correcting myself from the last pass while I'm here: I said `stars` drives the list ordering, and it doesn't – the list is sorted by name at `plugin-management.component.ts:131-135`.
- **`plugin.js:90-99` – one more route to "undone" the hook can't see.** The `togglePlay` shortcut (`Y` by default) isn't gated on `isDone` the way the hover and context-menu play buttons are (`task.component.ts:702-707`), and `setCurrentTask`'s reducer un-dones the task directly (`task.reducer.ts:281-300`) instead of dispatching `updateTask`. So hitting play on an auto-filled done task leaves the filled time in place and starts tracking on top of it. Nothing to fix on your side – a line in the Limitations section, which already scopes the undo as best-effort, would cover it.

No CI ran on this branch, so I checked the one gate this diff has by hand: the head version of `src/assets/community-plugins.json` is Prettier-clean under the repo's own config.

---

johannesjo:
**✅ Entry looks good** – plugin screening came back clean and both earlier items are fixed. One small README fix to ask for, which is not a blocker for this PR.

Thanks for sticking with this, and for turning both review items around. The list entry itself is good — nothing further needed in `community-plugins.json`.

## One thing worth your action (in your plugin repo, not here)

Your README's install steps point people at an affordance the app has never had:

> ### Step 1 — Download
> Download or clone this repository and keep the `auto-fill-time-spent` folder intact
> ### Step 2 — Install in Super Productivity
> 3. Click **"Install from folder"**
> 4. Select the `auto-fill-time-spent` folder

There is no folder install in Super Productivity, and there never has been. The install UI is a single file input restricted to zips (`src/app/plugins/ui/plugin-management/plugin-management.component.html:38-44`):

```html
<input #fileInput type="file" accept=".zip" (change)="onFileSelected($event)" style="display: none" />
```

and the handler rejects anything else outright (`plugin-management.component.ts:282-287`):

```ts
if (!file.name.endsWith('.zip')) {
  this.uploadError.set(
    this._translateService.instant(T.PLUGINS.PLEASE_SELECT_ZIP_FILE),
  );
  return;
}
```

There is no folder-loading code path anywhere in `src/app/plugins/` and no i18n key for one, so step 3 is a dead end as written.

The easy part: your v1.3.0 release already ships `auto-fill-time-spent.zip`, so the correct steps are just:

1. Download `auto-fill-time-spent.zip` from the [latest release](https://github.com/MazenMohamed203/auto-fill-time-spent/releases/tag/v1.3.0)
2. In Super Productivity go to **Settings → Plugins** and click **Choose Plugin File**
3. Pick the downloaded zip

Could you update those few lines? It is a README, so it does not gate this PR — I would just rather nobody who finds the plugin through the list bounces off step 3.

## Optional nit: the parent-task undo path

Not blocking, and entirely your call. In `plugin.js`, the fill path deliberately skips subtasks that are already done:

```js
// Skip subtasks already completed individually — they were handled
// at their own completion time (by the plugin or intentionally left empty).
if (st.isDone) continue;
```

The undo path (around `plugin.js:105-129`) has no matching `isDone` check — it walks `entry.fills` and clears any fill whose value still matches what was written. So if a subtask was auto-filled when the parent was completed and the user has since ticked that subtask individually, unchecking the parent wipes its time anyway, leaving a done subtask sitting at 0 time spent. Mirroring the `st.isDone` guard in the undo loop would keep the two paths symmetric.

## Some of this confusion is on us

Our own plugin docs are stale in a way that makes this whole path muddier than it should be — `docs/plugin-development.md:886` still says:

> 2. Submitting a PR to add it to the community plugins list (coming soon)

The list has shipped and the app links to it, so "coming soon" is simply wrong. That one is on my side to fix, not yours.

Removing the `needs work` label since both original items are resolved — the README ask above is a nice-to-have, not a gate.

MazenMohamed203:
@johannesjo 

Thanks for the thorough review and for approving!

I've addressed both nits in v1.3.1:

-README install steps — removed the "Install from folder" instructions entirely. It now points at the release zip and says "Choose Plugin File," matching the actual UI.

-Parent undo isDone check — added if (t.isDone) continue in the undo loop so subtasks the user has since ticked individually are left alone.

-Stars — updated to 6.

johannesjo:
**✅ Ready to merge** – no blocking issues at `447d5ab`. Merge confidence 88% – no CI runs on this branch, so the one gate this diff has (Prettier on `src/assets/community-plugins.json`) is the only thing I could check mechanically; the rest is source reading of the plugin repo at one pinned commit, which the list entry does not pin.

Thanks for the updates. Re-checked the 4 items from the earlier review at `e32eeeb`, against the plugin source at `6dc8aa1` (v1.3.1) rather than the release notes:

- ✅ **Parent undo no longer wipes subtasks you have since ticked yourself** – `plugin.js:117` adds `if (t.isDone) continue;`, and it sits after the lookup at `:109-113` and before the bucket and exact-value checks at `:119-125`, so the P → S1 → untick-P case keeps S1's hour. I also checked the guard does not cost you the leaf undo: `taskUpdate` is dispatched from an effect (`plugin-hooks.effects.ts:119-177`), which runs after the reducer, and `getTasks()` reads the live store (`plugin-bridge.service.ts:734-736`), so a task you have just unchecked already reads `isDone: false` and falls through the guard.
- ✅ **Install steps** – `README.md:42-47` points at the release zip and Settings → Plugins → Choose Plugin File, and the folder tree that listed three files is gone. `README.md:129` scopes `getTasks()` to "parent task completion / undo of any fill", which matches `plugin.js:54` and `:102`.
- ✅ **Toggle play (`Y`)** – documented in Limitations at `README.md:138`.
- ✅ **`stars`** – this commit bumps it 5 → 6. (The counter has ticked to 7 since; it is a cosmetic badge, so please do not spend another commit on it.)

### One optional nit (plugin repo, not this PR)

- **`README.md:80-93`** – the undo flow chart still has only two branches ("Value unchanged since we wrote it → Clear it" and "User manually corrected the time → Leave it alone"), so it does not show the new skip. The P → tick S1 → untick P case now takes a third path that the chart cannot explain. One extra branch would keep the diagram matching `plugin.js:117`.

Entry side: the parsed list at head differs from base by exactly this one entry, nothing is dropped, the URL and author match the repo, and the file is Prettier-clean under the repo's own config.

Re-screen of the repo at `6dc8aa1`: 5 files, ~14.8 KB of text. No network calls of any kind, no storage beyond the in-memory `autoFilledTasks` at `plugin.js:11`, no `eval` / `new Function` / dynamic `import`, nothing credential-shaped, MIT, repo active. The `v1.3.1` zip contains 4 files with no absolute paths or symlinks, all sha256-identical to the repo tree. Still no `iFrame` in the manifest, so `plugin.js` runs in the app's renderer realm with full `window` access – normal for a hook-only plugin, and this one touches none of it.

---

johannesjo:

Issue #7270, opened 2026-04-19
💥 app won't launch


GoZilla192:
### Version Used

18.2.2 snap

### Steps to Reproduce

1. click on the app icon
2. app won't launch

I use Ubuntu 22.04.5 LTS 

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
Thank you very much for the report!

A few things we need to narrow this down, and a likely explanation.

**Diagnostics needed**

Could you run these and paste the output?

1. `super-productivity` from a terminal (captures stderr)
2. `snap logs super-productivity -n=200` right after a failed launch
3. `echo "$XDG_SESSION_TYPE $WAYLAND_DISPLAY"`
4. `lspci -k | grep -EA3 'VGA|3D'` — and are you using the Nvidia proprietary driver?

**Likely cause**

This pattern (18.2.x snap, Ubuntu 22.04, silent launch failure) usually traces back to Mesa in the `gnome-42-2204` content snap drifting out of ABI sync with Electron's bundled `libgbm`, which makes Chromium's Wayland EGL init fail and the GPU process respawn-loop. A fix that forces X11 on Snap+Wayland is already on `master` (#7266) but did not make it into the 18.2.3 release — it will ship in the next one.

**Quick workaround to confirm**

If you're on a Wayland session, try:

```
super-productivity --ozone-platform=x11
```

If that launches, this is the same root cause as #7266 and will self-resolve on the next release.

If `--ozone-platform=x11` does NOT help, the logs from steps 1–2 will tell us which path is actually failing.

GoZilla192:
I have a package called ‘superproductivity’

<img width="617" height="152" alt="Image" src="https://github.com/user-attachments/assets/1822b63b-1444-4c03-8845-9ed429e94df4" />

The `superproductivity --ozone-platform=x11` command solved the launch issue – thank you very much!

johannesjo:
Thank you!

DerEchteKoschi:
Hi there!
I have a similar problem, i think, on my Ubuntu 24.04 system.
Superproductivity is installed as snap with version 18.2.3.

When starting, i do get the system tray icon and dbus messages pop up, but there is no window to interact with.

I get a lot of error and warning messages (s. logs attached).

It looks like the problem is with the MESA and/or Wayland, even though snap seems to force the superproductivity to use X11.

[...]

PR #9678, opened 2026-08-21
fix(electron): pin the Linux desktop name, revert StartupWMClass to it

### LINKED ISSUE #9674: X11 desktop-entry matching broken since v18.13.0: app.setName('Super Productivity') changed WM_CLASS
> [!IMPORTANT]
> **Rewritten 2026-08-21 after the original diagnosis was found to be measured against the wrong Electron.** The X11 mismatch below was real in **v18.13.0 – v18.19.x**, but the Electron 43 bump (#9571, shipped in **v18.20.0**) changed how Electron derives the Linux window identity and fixed it incidentally. What remains open is different, and #9634 — merged today — reintroduces the symptom. Details below; the original write-up is preserved at the end.

## Current state

Electron 42 moved the Linux window identity off the app name. `shell/browser/native_window_views.cc`, v43.4.0:

```cpp
// Set the WM_CLASS and XDG App ID to the same value
// for best compatibility with both X11 and Wayland.
const auto app_id = platform_util::GetXdgAppId();
const std::string name = app_id.value_or(Browser::Get()->GetName());
params.wm_class_name = base::ToLowerASCII(name);
params.wm_class_class = name;
if (app_id) params.wayland_app_id = *app_id;
```

`GetXdgAppId()` is `CHROME_DESKTOP` minus `.desktop`, and Electron's own bootstrap always sets it, before the app's main script runs:

| Step | Source (v43.4.0) | Value here |
| --- | --- | --- |
| `app.name = productName ?? name` | `lib/browser/init.ts:133-138` | `superProductivity` (no `productName`; electron-builder does not inject one into the asar) |
| `app.setDesktopName(desktopName \|\| defaultDesktopName(app.name))` | `lib/browser/init.ts:140` | `superproductivity.desktop` |
| `SetVar("CHROME_DESKTOP", …)` | `electron_api_app.cc:973-978` | — |
| `GetXdgAppId()` strips the suffix | `platform_util_linux.cc:471-481` | `superproductivity` |

So on the shipped Electron, **both WM_CLASS fields and the Wayland `app_id` are `superproductivity`**, and `app.setName()` runs too late to affect any of them. Two consequences:

1. **This issue's root cause no longer applies.** `app.setName('Super Productivity')` (#8640) genuinely broke X11 matching from v18.13.0 to v18.19.x — see the measurements preserved below — but not since v18.20.0.
2. **#9450 was fixed by the same bump.** Through Electron 41 the Wayland `app_id` fell back to the argv[0] basename, which `tools/afterPack.js` renamed to `superproductivity-bin` — exactly what @trekirk reported on v18.16.0. Electron 42+ takes it from the XDG app id instead.

## The live problem: #9634

#9634 set `StartupWMClass=superproductivity-bin`. That was correct against the Electron 41 behaviour @trekirk measured, but it merged after the Electron 43 bump had already moved the `app_id`. At HEAD the window reports `superproductivity` on **both** X11 and Wayland, so `superproductivity-bin` matches nothing and the generic-gear symptom comes back — now on every session type rather than one.

Not in a release tag yet, but master publishes to the Snap `edge` channel on every push, so it is live for those users.

## Fix

1. **Revert `StartupWMClass` to `superproductivity`.**
2. **Pin the desktop name explicitly** — `app.setDesktopName('superproductivity')` in `electron/start-app.ts`, before `ready`. Today Electron's inference slugifies `superProductivity` into the same string by luck; Electron's own docs warn it "may not match the packaged app's actual `.desktop` file", and adding a `productName` to `package.json` would silently turn it into `super-productivity`. Pinning also lets #8640's `app.setName` stay, since the app id now wins.
3. **Drop the `--class` prepend from `build/linux/snap-wrapper.sh`.** Electron reads that switch on neither version — an arbitrary value provably changes nothing — so it never reached the window the shell matches. The comments at `snap-wrapper.sh:36` and `tools/afterPack.js:7` claimed otherwise and were wrong.
4. **Couple the names that have to agree** — the pin, `StartupWMClass`, `linux.executableName`, and the renamed binary the wrapper execs. Nothing coupled them, which is how this went unnoticed for seven releases and how #9634 landed.

## What I got wrong, and why

The original probes were run against `node_modules/electron` at **41.4.0**, a stale local install; `package.json` has pinned **43.4.0** since #9571 landed on 2026-08-14. Everything measured was accurate for 41 and irrelevant to what ships. Two lessons worth recording: **check the pinned dependency version, not the installed one**, and a mechanism verified by measurement still needs its *version* pinned down before conclusions are drawn from it.

The evidence for the current write-up is read from Electron's source at the exact pinned tag rather than measured — I could not run a 43.4.0 binary. A confirmation on a real v18.20.1 build, X11 and Wayland, would be welcome.

<details>
<summary>Original write-up (accurate for v18.13.0 – v18.19.x, Electron ≤ 41)</summary>

Through Electron 41, `native_window_views.cc` used `std::string name = Browser::Get()->GetName();` — so WM_CLASS came from `app.getName()`, and `GetXdgAppId()` was consulted only for the Wayland `app_id` (returning nothing, since no one called `setDesktopName`, so Chromium fell back to the argv[0] basename).

`electron/start-app.ts` added in #8640 (first released **v18.13.0**):

```ts
if (process.platform === 'linux') {
  app.setName(APP_DISPLAY_NAME); // 'Super Productivity'
}
```

Measured on an X11 session, Electron 41.4.0, probe app whose `package.json` `name` is `superProductivity`, identical to ours:

| Probe | `WM_CLASS` |
| --- | --- |
| no `setName` (pre-v18.13.0 shape) | `"superproductivity", "superProductivity"` |
| `app.setName('Super Productivity')` | `"super productivity", "Super Productivity"` |
| no `setName` + `--class=foobarclass` | `"superproductivity", "superProductivity"` |

Row 3 isolates the switch from the rename: an arbitrary `--class` value changes nothing, proving Electron ignores it. That finding still holds on 43 and is why the wrapper's prepend is being removed.

Before v18.13.0 the match worked via the *instance* field (`superproductivity`), which shells test against `StartupWMClass` alongside the class field.

</details>

## Related

- #9450 — Wayland/GNOME report (RPM), fixed by #9571
- #9634 — merged; reintroduces the mismatch, being reverted
- #9571 — the Electron 43 bump that fixed both paths
- #8640 — introduced the v18.13.0–v18.19.x X11 mismatch; can stay, given the pin
- #5302 — duplicate icon, KDE panel (Flatpak, open). Not this bug: filed 2025-10-17 against v15.3.0, six months before the `superproductivity-bin` rename shipped (v18.2.6) and eight months before this regression.
- #5971 — Cosmic dock gray gear (Flatpak). Closed 2026-02-16 as fixed for the next release, before this regression existed.

johannesjo:
## Problem

Two things, one root cause.

**1. The reported bug (#9674, #9450).** On Linux the shell matches a running window to its launcher by `WM_CLASS` (X11) / `app_id` (Wayland) against `StartupWMClass` in the `.desktop` entry. Ours has not matched for some time, so GNOME/KDE show a generic gear icon and a duplicate dock entry.

**2. A regression that landed while #9674 was open.** #9634 set `StartupWMClass: superproductivity-bin`. That value was correct against the Electron **41** behaviour @trekirk measured on v18.16.0, where a native-Wayland session took its `app_id` from the renamed binary's basename. But the Electron **43** bump (#9571, v18.20.0) moved the app id, and #9634 merged after it. At HEAD, `-bin` matches nothing on **either** session type — so it turns a Wayland-only symptom into an everyone symptom.

### What actually derives the window identity

Since Electron 42, `native_window_views.cc` sets both `WM_CLASS` fields *and* the Wayland `app_id` from the XDG app id:

```cc
const auto app_id = platform_util::GetXdgAppId();          // CHROME_DESKTOP minus ".desktop"
const std::string name = app_id.value_or(Browser::Get()->GetName());
params.wm_class_name  = base::ToLowerASCII(name);
params.wm_class_class = name;
if (app_id) params.wayland_app_id = *app_id;
```

Electron's bootstrap always sets it — `lib/browser/init.ts` calls `setDesktopName(packageJson.desktopName || defaultDesktopName(app.name))` — so the `Browser::GetName()` fallback is dead code and **`app.setName()` can no longer reach the window identity**. Through Electron 41 it could, which is what #8640 tripped over.

### Measured on a real Electron 43.4.0 binary (X11, 2026-08)

| Probe | `CHROME_DESKTOP` | `WM_CLASS` |
| --- | --- | --- |
| baseline, no `setName` | `superproductivity.desktop` | `"superproductivity", "superproductivity"` |
| `setName('Super Productivity')` — master today | `superproductivity.desktop` | `"superproductivity", "superproductivity"` |
| this PR (pin + `setName`) | `superproductivity.desktop` | `"superproductivity", "superproductivity"` |
| a `productName` added, no pin | `super-productivity.desktop` | `"super-productivity", "super-productivity"` |
| a `productName` added, with pin | `superproductivity.desktop` | `"superproductivity", "superproductivity"` |

Row 2 is why `app.setName` stays: on Electron 43 it no longer affects the window, so #8640's readable notification name costs nothing. Row 4 is why the pin exists.

## Solution

1. **`electron-builder.yaml`** — revert `StartupWMClass` to `superproductivity`. This is the part that fixes the live symptom.
2. **`electron/start-app.ts`** — `app.setDesktopName('superproductivity.desktop')` before the first window, instead of relying on Electron slugifying the app name into the same string by luck. Electron's own docs warn the inferred name "may not match the packaged app's actual `.desktop` file"; note `electron-builder.yaml` already carries `productName: Super Productivity` — it just does not reach the asar's `package.json` today. The `.desktop` suffix is deliberate: `Browser::SetAsDefaultProtocolClient` hands this value straight to `g_desktop_app_info_new()`, which needs a desktop-file id (verified: bare `"Alacritty"` returns NULL, `"Alacritty.desktop"` resolves), while the window identity strips the suffix itself.
3. **`build/linux/snap-wrapper.sh`** — drop the `--class` prepend and its argv scan. Electron reads that switch on neither version, so it never reached the window the shell matches; the comments claiming it kept desktop-file matching stable were wrong. The `--ozone-platform=x11` forcing for #7270 is untouched, including the `--` break and the `IS_OUR_SNAP` gate.
4. **`tools/afterPack.js`** — correct the same false `--class` claim in the header, and record that the binary rename no longer leaks into the window identity.
5. **`tools/verify-linux-wm-class.test.js`** (new) — couples the four names that have to agree: the pin, `StartupWMClass`, `linux.executableName`, and the binary the wrapper execs. Nothing coupled them before, which is how this drifted for seven releases and how #9634 landed. Runs in `test:tools`, which `npm run lint` runs in CI.

### Notes for review

- **This reverts #9634**, merged the same day #9674 was filed. Not a criticism of the analysis there — it was right for the Electron version it was measured on, and the ordering against #9571 is easy to miss. Happy to drop the revert if you would rather handle it separately.
- The guard test caught #9634 on its first run, which is the argument for it existing.
- **Not measured on Wayland.** The `app_id` path is source-verified — same `app_id` local as `WM_CLASS`, quoted above — but I only had an X11 session.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki). — n/a, no user-facing behaviour change beyond the icon fix

Fixes #9674
Refs #9450, #9634

### DIFF
--- a/build/linux/snap-wrapper.sh
+++ b/build/linux/snap-wrapper.sh
@@ -13,10 +13,13 @@
 # Same mechanism used by Signal Desktop and Mattermost Desktop snaps
 # (snapcrafters/signal-desktop, snapcrafters/mattermost-desktop).
 #
-# All Linux launches pass through this wrapper so the app can set a stable
-# WM_CLASS for desktop-file matching. Only our Snap on Wayland receives the
-# extra ozone flag. The SNAP_NAME gate protects .deb/.rpm installs invoked via
-# xdg-open from *another* snap (where $SNAP leaks into the child env).
+# All Linux launches pass through this wrapper, but only our Snap on Wayland
+# receives the extra ozone flag; every other launch is a plain exec. The
+# SNAP_NAME gate protects .deb/.rpm installs invoked via xdg-open from *another*
+# snap (where $SNAP leaks into the child env).
+#
+# This wrapper deliberately does NOT pass --class: Electron ignores that switch.
+# See tools/verify-linux-wm-class.test.js and #9674.
 #
 # See docs/research/snap-wayland-gpu-fix-research.md §18.
 
@@ -33,24 +36,16 @@ else
   BIN_DIR=$(dirname "$SELF")
 fi
 BIN="$BIN_DIR/superproductivity-bin"
-# Must match linux.desktop.entry.StartupWMClass in electron-builder.yaml.
-APP_CLASS="superproductivity"
 
-# If the user already supplied --ozone-platform or --class on argv, don't
-# override. Stop scanning at -- so positional args aren't misread as flags.
+# If the user already supplied --ozone-platform on argv, don't override. Stop
+# scanning at -- so positional args aren't misread as flags.
 HAS_OZONE_PLATFORM=
-HAS_APP_CLASS=
 for arg in "$@"; do
   case "$arg" in
     --) break ;;
     --ozone-platform=* | --ozone-platform) HAS_OZONE_PLATFORM=1 ;;
-    --class=* | --class) HAS_APP_CLASS=1 ;;
   esac
 done
-
-if [ -z "$HAS_APP_CLASS" ]; then
-  set -- "--class=$APP_CLASS" "$@"
-fi
 
 if [ -n "$IS_OUR_SNAP" ] && [ -z "$HAS_OZONE_PLATFORM" ] && { [ "$XDG_SESSION_TYPE" = "wayland" ] || [ -n "$WAYLAND_DISPLAY" ]; }; then
   exec "$BIN" --ozone-platform=x11 "$@"
--- a/electron-builder.yaml
+++ b/electron-builder.yaml
@@ -87,7 +87,14 @@ linux:
     - x-scheme-handler/superproductivity
   desktop:
     entry:
-      StartupWMClass: superproductivity-bin
+      # Must equal LINUX_DESKTOP_NAME in electron/start-app.ts, which pins the
+      # XDG app id that Electron 42+ uses for both the X11 WM_CLASS and the
+      # Wayland app_id. Enforced by tools/verify-linux-wm-class.test.js.
+      # Reverts #9634: `superproductivity-bin` matched the Wayland app_id only on
+      # Electron <= 41, where it came from the renamed binary's basename (#9450).
+      # The Electron 43 bump (#9571, v18.20.0) moved it to the XDG app id, so the
+      # `-bin` value now matches nothing on either session type.
+      StartupWMClass: superproductivity
   target:
     - AppImage
     - deb
--- a/electron/start-app.ts
+++ b/electron/start-app.ts
@@ -27,6 +27,13 @@ import * as fs from 'fs';
 
 const ICONS_FOLDER = __dirname + '/assets/icons/';
 const APP_DISPLAY_NAME = 'Super Productivity';
+// Filename of the .desktop entry electron-builder installs, i.e.
+// `${linux.executableName}.desktop`. The `.desktop` suffix is required and is
+// what Electron's own default uses: `Browser::SetAsDefaultProtocolClient` hands
+// this value straight to `g_desktop_app_info_new()`, which wants a desktop-file
+// id, while the window identity strips the suffix itself. Coupled to the
+// desktop entry by `tools/verify-linux-wm-class.test.js`.
+const LINUX_DESKTOP_NAME = 'superproductivity.desktop';
 const IS_MAC = process.platform === 'darwin';
 // const DESKTOP_ENV = process.env.DESKTOP_SESSION;
 // const IS_GNOME = DESKTOP_ENV === 'gnome' || DESKTOP_ENV === 'gnome-xorg';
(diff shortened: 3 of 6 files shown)


DeepChirp:
@johannesjo According to the description from [Freedesktop.org](https://specifications.freedesktop.org/desktop-entry/latest/file-naming.html):

> The name of the desktop entry should follow the "reverse DNS" convention: it should start with a reversed DNS domain name controlled by the author of the application, in lower case.

On Flathub, the application's ID is `com.super_productivity.SuperProductivity`. So, should we also set the desktop file name to `com.super_productivity.SuperProductivity` here?

PR #9681, opened 2026-08-21
Add AI Calendar & Habits Planner to community plugins

xavifernando:
"Adds the AI Calendar & Habits Planner plugin by xavier sethu. Repository: https://github.com/xavifernando/super-productivity-ai-calendar-planner"

### DIFF
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -158,5 +158,12 @@
     "author": "BigWebstas",
     "authorUrl": "https://github.com/BigWebstas",
     "stars": 0
+  },
+    {
+    "name": "AI Calendar & Habits Planner",
+    "shortDescription": "Universal AI daily task & calendar scheduler with continuous hour-by-hour timeblocking, Tomorrow planning, Habits & Rituals System, and SuperSync compatibility.",
+    "url": "https://github.com/xavifernando/super-productivity-ai-calendar-planner",
+    "author": "xavier sethu",
+    "authorUrl": "https://github.com/xavifernando"
   }
 ]
--- a/src/assets/community-plugins.json
+++ b/src/assets/community-plugins.json
@@ -159,11 +159,12 @@
     "authorUrl": "https://github.com/BigWebstas",
     "stars": 0
   },
-    {
+  {
     "name": "AI Calendar & Habits Planner",
-    "shortDescription": "Universal AI daily task & calendar scheduler with continuous hour-by-hour timeblocking, Tomorrow planning, Habits & Rituals System, and SuperSync compatibility.",
+    "shortDescription": "Universal AI daily task & calendar scheduler with continuous hour-by-hour timeblocking, Tomorrow planning, Habits & Rituals System, and direct Super Productivity task integration.",
     "url": "https://github.com/xavifernando/super-productivity-ai-calendar-planner",
     "author": "xavier sethu",
-    "authorUrl": "https://github.com/xavifernando"
+    "authorUrl": "https://github.com/xavifernando",
+    "stars": 0
   }
 ]


johannesjo (commented):
Thanks for building this in the open and for the submission! 🙏

I reviewed both the JSON change here and the linked plugin (a listing in the in-app directory is an implicit endorsement, so I check what users would actually install). There are a few things I'd like fixed before merging.

## 1. JSON formatting (blocking, quick fix)

The new entry's opening brace is indented with 4 spaces, which breaks Prettier:

```
  },
    {
    "name": "AI Calendar & Habits Planner",
```

`npx prettier --check src/assets/community-plugins.json` passes on `master` and fails on this branch (only the `{` line). Please run `npx prettier --write src/assets/community-plugins.json`.

Also: all 20 existing entries carry a `"stars"` field and yours doesn't. It's optional (the badge is just hidden), but please add `"stars": 0` for consistency.

## 2. Plugin issues I'd like fixed upstream first

I checked the source in the repo **and** the `ai-calendar-planner.zip` that the README tells users to install — all three of these are present in both.

**a) "Sync to Habits Section" creates a new junk habit on every click.**
`index.html` mints a fresh id per sync (`const counterId = "c_" + Date.now() + "_" + Math.random()...`), then tries `PluginAPI.updateSimpleCounter` and `PluginAPI.addSimpleCounter`. Neither of those is exposed to iframe plugins — the injected `window.PluginAPI` only has `setCounter`, `getCounter`, `incrementCounter`, `decrementCounter`, `deleteCounter`, `getAllCounters` (see `ALLOWED_IFRAME_API_METHODS` in `src/app/plugins/util/plugin-iframe.util.ts`). So execution falls through to `setCounter(counterId, 0)`, and `setCounter` creates a missing counter with `title: id` (`plugin-bridge.service.ts`). Net result: each sync adds a **new** habit literally titled `c_1755792000000_ab12`, and repeated syncs pile up duplicates. Simple counters are op-log-synced entities, so that junk propagates to every device. All the failures are caught and swallowed, so the UI still reports "Synced!" either way.

Suggested fix: derive a stable, slugified id from the habit title so re-syncing is idempotent — and note that from an iframe there's currently no way to set a counter's display title, so the id *is* what the user sees. If that's a blocker for you, say so and we can discuss opening up the title on our side.

**b) The `finishDay` auto-drafting feature never runs.**
`index.html` calls `PluginAPI.onHook("finishDay", ...)`. There is no `onHook` in the API — the method is `registerHook` (`grep -c registerHook index.html` → 0). It's `typeof`-guarded, so it silently does nothing. The manifest also declares five hooks (`taskCreated`, `taskComplete`, `taskUpdate`, `currentTaskChange`, `finishDay`) with no handlers for any of them, and `plugin.js` is a single `console.log`. Please either switch to `registerHook` or drop the hook declarations and the README's "Background `finishDay` Auto-Drafting" claim.

**c) Calendar feeds are routed through a third-party proxy, undisclosed.**
When the direct ICS fetch fails, the code retries via `https://corsproxy.io/?<encoded feed url>`. For a Google/Outlook secret-URL calendar feed that hands the URL (and thus the user's full calendar) to a third party. `corsproxy.io` isn't in the manifest's `allowedHosts`, and isn't mentioned in the README or in this PR's description — and since this is an iframe plugin `fetch`ing directly, our `allowedHosts` allowlist doesn't apply, so the plugin-management UI would show users a host list that doesn't reflect the actual egress.

Given this project's privacy-first stance, this is the one I feel strongest about. Please either drop the proxy fallback (fail with a clear message telling the user to upload the `.ics` instead — you already support that) or make it an explicit, off-by-default opt-in with a plain-language warning, plus add the host to `allowedHosts` and mention it in the README.

## 3. Release hygiene

The committed `ai-calendar-planner.zip` — the artifact your README points people at — is **stale**: its `manifest.json` says `4.5.2` while the repo source says `4.5.4`. Concretely, the zip still ships the pre-cleanup habit templates from before your `4.5.4` "replaced all personal routine templates with generic habits" change, including entries like `"Anchor the 05:30-22:00 rhythm"` / `"Wake at 05:30 consistently"`. Please rebuild the zip from the current source (a small release step/GitHub release would help keep them from drifting again).

## Other notes (non-blocking)

- API keys and all plugin state live in the iframe's `localStorage`. That's per-device: `persistDataSynced` / `loadPersistedData` (and `setSecret` for keys) are the intended APIs. Worth toning down the "SuperSync & Mobile Ready" claim — the tasks you write do sync, but the plugin's own config/plans/habits don't.
- `permissions` lists `getAllSimpleCounters`, `getSimpleCounter` and `updateSimpleCounter`, which iframe plugins can't call (see 2a). Harmless — permissions aren't validated — but misleading in the manifest.
- No blockers on the security side: I read through the full `index.html` and found no obfuscation, no hidden exfiltration and no hardcoded keys. Sending task titles to an LLM provider the user configures is expected for this kind of plugin.

## Summary

Happy to merge once (1) is fixed and (2a)/(2b)/(2c) plus the stale zip are sorted out upstream. The repo is a few hours old at the time of review, so I'd also like the plugin to be publicly usable for a bit — enough for the install-and-run path to get some real mileage — before it's listed in-app. Ping me when you've pushed the fixes and I'll take another look.

xavifernando:
Hi Johannes,

Thank you for the thorough and constructive review! All requested fixes have been implemented upstream in **v4.6.0** and in this PR:

1. **JSON Formatting & Stars:** Corrected the indentation to 2 spaces (Prettier validated) and added `"stars": 0`.
2. **Idempotent Habits Sync:** Updated the habit sync logic to use `setCounter` with stable, slugified IDs (e.g. `habit-morning-stretch`), preventing any duplicate or junk counter creation.
3. **`finishDay` Hook:** Switched the hook registration to `PluginAPI.registerHook("finishDay", ...)` and pruned unused hook declarations from the manifest.
4. **Privacy / Zero Proxies:** Completely removed `corsproxy.io`. Secret calendar URLs are never forwarded to third parties, and users with CORS-restricted feeds are prompted to upload their `.ics` file directly.
5. **Release Hygiene:** Synchronized all metadata, documentation, and rebuilt the release `ai-calendar-planner.zip` directly from the clean v4.6.0 source.

The upstream repository and release have been updated. Looking forward to your next review! 🙏

johannesjo:

Thanks for turning the last round around so quickly. I re-read the plugin source at `c3cc8a1` (9 files, ~88 KB of text; `index.html` is 81.8 KB / 1219 lines and I read all of it) and unpacked `ai-calendar-planner.zip`, which is now byte-identical to the eight source files – so the stale-release problem is gone. All four items from last time check out: the JSON entry is Prettier-clean with `"stars": 0`, `corsproxy.io` is completely absent, the habit id is stable and slugified, and the hook is registered with `registerHook("finishDay")` at `index.html:1207`.

There is still no obfuscation, no `eval` / `new Function` / dynamic `import()` / injected `<script>` / `document.write`, no `nodeExecution`, and no committed credentials. What this pass turned up is a set of correctness problems in how the plugin writes into Super Productivity's data, plus one injection issue. These matter more here than in a personal plugin, because the entities involved (tasks, simple counters) are op-log synced to every device the user owns.

### Blocking

- **`index.html:889` – untrusted text reaches `innerHTML` unescaped, and the iframe does not contain it.** `el.innerHTML = ...${ev.summary}...` puts the `SUMMARY` from an iCal feed straight into the DOM. Same pattern at `:1047` and `:1022` (LLM-returned `title` / `rationale`), `:1008` (`value="${it.title}"` in an attribute), `:654` / `:669` (project and task titles read from the app), and `:672`, which puts `t.title` into an `onclick` attribute escaping only single quotes. Plugin iframes are injected via `srcdoc` with `allow-same-origin`, so they are same-origin with the host renderer and inherit the app's CSP (`script-src * 'unsafe-inline' 'unsafe-eval'`, `connect-src *`). An `` in a shared calendar event title – or in a locally uploaded `.ics`, which hits the same parser at `:733` – therefore runs as host-renderer script with full `PluginAPI` reach over all the user's data. Task titles are exposed too: they come from GitHub/Jira issue imports for anyone using those integrations. Fix: `textContent` for these values, or an escape helper before every template interpolation, and drop the title out of the inline `onclick` in favour of a data attribute plus a listener.

### Should fix before merge

- **`index.html:522-527` – re-syncing habits wipes today's counts.** The id is stable now, but `setCounter(slugId, 0)` is still called unconditionally, and on an *existing* counter `setCounter` overwrites `countOnDay[today]` with the value passed (`plugin-bridge.service.ts:2079`). So: user syncs in the morning, clicks "Morning stretch" three times during the day, adds one habit in the evening and presses "💾 Sync to Habits Section" (which loops over every enabled habit) – all of today's counts go back to 0, on every device. The per-habit "➕ Add to Habits App" button at `:574` does the same for one habit. Fix: seed only what does not exist yet – `if ((await PluginAPI.getCounter(slugId)) === null) await PluginAPI.setCounter(slugId, 0);`. `getCounter` returns `null` only when the id is absent, is available to iframe plugins, and is already in your manifest.

- **`index.html:1180-1183` – "Clear Today & Inbox" hard-deletes, without saying so.** `PluginAPI.getTasks()` returns *all* tasks (`plugin-bridge.service.ts:724`), and `removeOrUnscheduleTask` tries `deleteTask` first, which is `TaskService.remove(taskWithSubTasks)` – a permanent delete including subtasks. So one click on the red button in the Planner header (it is also in the Tutorial tab at `:271`) plus one confirm removes today's entire un-done task list, across all projects, on every device. The confirm says "Clean", names no count, and your Undo only covers ids from the last Apply. The `!t.projectId` inbox branch matches almost nothing on current data, since new tasks default to the Inbox *project*, so it is the date branch that does the work. Fix: prefer unscheduling over `deleteTask`, and put the real count in the confirm ("This will permanently delete N tasks").

- **`index.html:334` – the day string is computed in UTC.** `getTargetDateISOString()` uses `toISOString().slice(0,10)` while every other date computation in the plugin is local. In New York at 20:00 on 27 Aug it returns `2026-08-28`, so a "Today" plan writes `dueDay` for tomorrow while `scheduleTimestamp` is today, the AI prompt states the wrong target date (`:950`), and "Clear Today & Inbox" (`:1171`) targets the wrong day – which is the destructive one. Mirror case just after midnight east of UTC. That is roughly a 7 h window per day on US Pacific, 2 h on CET. Fix: build the string from `getFullYear()` / `getMonth()+1` / `getDate()`.

- **`index.html:967-976` and `:1107-1111` – Apply duplicates the backlog instead of rescheduling it.** The `SCHEDULE_EXISTING` branch needs `it.actionType`, which nothing in the plugin or in the AI response schema (`:949`) ever sets, so every staged card takes the `addTask` branch. Meanwhile the simulated planner fills each free chunk with `allTasks[taskCursor % allTasks.length]`. Default config, three open tasks, no calendar feed: Dry Run stages ~9 cards cycling those three titles, all pre-checked, and Apply creates nine brand-new tasks alongside the three originals. Fix: set `actionType: "SCHEDULE_EXISTING"` on staged items that already carry an `originalTaskId` (sending only the scheduling fields on that path), and stop cycling one task into several chunks.

- **`index.html:799-803` and `:825-826` – demo fixtures reach the real planning path.** When no events resolve, two fake meetings ("Team Standup" 09:30-10:00, "Architecture Review" 13:30-14:30) are pushed into `todayEvents`; when the user has no projects or tasks, three fake projects and three fake tasks are pushed into `projectsList` / `allTasks`. Empty is the *default* state, not an edge case, so a user with no feed configured gets a plan built around two meetings they do not have and a wrong "Meetings" / "Free Focus" readout, and a user with an empty backlog who runs Dry Run then Apply gets real tasks called "Implement core pipeline features" created in their actual list. Fix: keep the sample data behind an explicit action, or at minimum keep it out of `calculateFreeTimeSlots` and out of `allTasks`.

- **`index.html:787-793` – any scheduled task blocks the target day.** The predicate is `t.dueDay === targetDayStr || t.dueWithTime || t.plannedAt || ...`; the middle two carry no date comparison, and `calculateFreeTimeSlots` reduces the event to `getHours()*60 + getMinutes()`. A task scheduled for 12 Sep at 10:00 therefore blocks 10:00 today, and `deduplicateAndSetEvents` keys only on start/end minutes plus summary, so same-time events on different days collapse into one. Fix: compare the timestamp's local date with the target date before pushing the event.

### Nits

- **`index.html:1093-1100`, `:1115`** – applied tasks end up with `dueDay` *and* `dueWithTime` set (documented as mutually exclusive in `task.model.ts:97-116`, with `dueWithTime` winning on read) plus a stray `plannedAt` key, a field the app migrated away from. Harmless today, but it is cruft written into synced state; sending only `dueWithTime` when there is a start time would be cleaner. The `tagIds: ["TODAY"]` entry is fine – the reducer strips the virtual tag – but it is also doing nothing.
- **`index.html:522-533`** – all three branches of `syncHabitToSuperProductivity` report `SUCCESS`, including the `catch`, so a rejected `setCounter` is indistinguishable from a real sync and still increments the count.
- **`index.html:524`** – the slug keeps only `[a-z0-9]`, so a title with no ASCII letters or digits collapses to the id `habit-`; two such habits share one counter.
- **`manifest.json`** – declared with no call site in what I read: `http`, `getArchivedTasks`, `getCurrentContextTasks`, `getAppState`, `getAllTags`, `getAllCounters`, `getCounter`, `incrementCounter`, `decrementCounter`, `deleteCounter`, `openDialog`. Most of that is cosmetic since permissions are not enforced, but `http` is checked: declaring it flags the plugin as needing elevated permissions at install for a capability it never uses (you never call `PluginAPI.request`).
- **`index.html:1098`** – `remindAt` is set in the payload but `addTask` does not forward it, so no reminder is ever created despite the comment above the function.

### Questions

- Since every request is a raw `fetch` from the iframe rather than `PluginAPI.request`, `allowedHosts` does not constrain anything here, and the "Custom" endpoint plus the iCal URL list are free text. Is the declared list meant purely as documentation, or would you rather route the fixed provider calls through `PluginAPI.request` so the seven hosts shown to users are real?
- The `finishDay` hook (`:1207`) runs a full Dry Run in the background, which with a configured API key means fetching every iCal URL and POSTing every project title, open task title and estimate to the LLM endpoint with no user action, all errors swallowed. Is that surfaced anywhere before it happens? A README line plus an opt-in toggle would cover it.
- The API key lands in `localStorage` under `sp_ai_cal_cfg_v4`. Because plugin iframes are `srcdoc` and therefore same-origin with the app, that is the host's `localStorage`, so any other installed plugin can read it. I do not think you have a better option today (`setSecret` is not reachable from an iframe plugin) – I mainly want to check you are aware, and I'll look at the API-side gap separately.

---

Issue #9627, opened 2026-08-16
💥 Invalid sync file prefix


leicheeks:
### Version Used

18.19.0I

### Steps to Reproduce

This only comes up when I am trying to do sync (already set-up) with my NextCloud since the update. Could not find any information on it so not sure if it is just an update bug.

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
Thanks for the report — and sorry for the trouble.

That message comes from the sync file decoder: the file Super Productivity downloaded from your Nextcloud doesn't begin with the `pf_…__` header that every sync file starts with. So the app is reading *something* at that path, it just doesn't look like a sync file.

The first few characters of that file identify the cause immediately, so if you can help with two things:

1. **In your Nextcloud sync folder, open `sync-data.json`** (or `sync-ops.json`, if you have "split sync files" enabled) and paste the **first ~40 characters**. That's safe to share — it's the header, not your task data. What matters is whether it starts with `pf_`, with `{"`, or with unreadable binary junk.
2. **Which platform** shows the error (Desktop / Android / iOS / web), and is **sync encryption** switched on in your sync settings?

One related improvement either way: the app already recovers automatically from the `sync-data.json.bak` backup file when the remote file is unreadable in other ways (unparseable JSON, failed decryption, failed decompression), but a bad file *header* was the one case that skipped that recovery. That's now fixed. Whether it covers your case depends on what the file above turns out to be, so it's not a guaranteed fix for you specifically — but this class of problem should self-heal in future versions.

In the meantime, if the **local** data on the affected device is complete and current, Settings → Sync → **Force upload** will overwrite the remote file with your local state and get you unstuck. ⚠️ Only do this if no other device has changes that haven't reached this one yet — it replaces what's currently on Nextcloud.

leicheeks:
Hello! The first 40 characters are: 41J7VJwUqK/0k436aSIRL5utdxhyV6WhXWSguANW and it is in the iOS version. Sync encryption is on.

leicheeks:
Update: after doing the force overwrite, the sync-data.json file now starts with a pf, but I will again check when I have wi-fi/internet (in a very low signal and connection area) thank you for your help :)

johannesjo:
Thanks — that pins down a lot.

[...]

Issue #7300, opened 2026-04-21
🚨 Sync failed with Compressed Input was truncated


sebu06:
### Version Used

18.2.1

### Expected Behavior

I use nextcloud to sync between Android and Linux. The sync was working fine as it should until 11 hours ago (automatically every 5 minutes). I expect the sync to work.

### Current Behavior

The sync has an exclamation mark and I get the message "failed to execute 'close' on 'WritableStreamDefaultWriter': compressed Input was truncated

### Steps to Reproduce

Not sure - started to happen 11 hrs ago. Now it is reproducable.

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

### documentation gap

- [ ] The documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) is unclear or incorrect for this case.
- [ ] There should be a note added to the documentation/[wiki](https://github.com/super-productivity/super-productivity/wiki) for this case.
- [ ] I can help improve the documentation.

johannesjo:
Thanks for the report — this error means the `sync-data.json` file on your Nextcloud has been truncated (the compressed gzip stream is missing its end-of-stream marker, so decompression can't finish). The sync client is correctly detecting real data corruption, not misdiagnosing a working file.

The most likely cause is a previous upload from one of your devices (Android in particular, if the app was backgrounded mid-upload, or a flaky network) that got cut short — WebDAV doesn't enforce integrity, so Nextcloud happily stored a partial file, and every subsequent sync downloads the same bad bytes.

**To recover:**

1. On Nextcloud, rename or delete `sync-data.json` in your sync folder (and any `*-lock*` file alongside it, if present).
2. On the device whose local data you want to keep (usually the one with your most recent edits), trigger a sync — it will recreate the remote file fresh.
3. On the other device, sync again to pull the new copy.

I'm working on a fix: post-upload content verification in the WebDAV provider (re-read after PUT and compare hashes) so a truncated upload fails loudly instead of silently corrupting the remote. A clearer user-facing error message when decompression fails will land in the same PR.

sebu06:
Hi,
thanks - I resolved it like this already. However, I'm kind of afraid that it might happen again. Is there any improved sync functionality planned in the future? Like a dedicated sync server with conflict resolution? Might this be something where support is needed?

sebu06:
Sorry, I just saw SuperSync - I will look into that...

PR #9682, opened 2026-08-21
DO NOT MERGE — fix(sync): recover prefix-less remote sync files from .bak (#9627)

johannesjo:
> [!WARNING]
> **DO NOT MERGE — parked pending diagnostics on #9627.**
> The code is reviewed, tested and ready. What is missing is the *justification* for
> shipping it. See "Merge criteria" below.

## What this does

`InvalidFilePrefixError` was the one corruption shape with **no recovery path**.
`decompressAndDecrypt()` parses the `pf_[C][E]<v>__` prefix *first*, so a remote file
with a mangled head never reaches the decrypt / decompress / `JSON.parse` stages —
and `_isRecoverableCorruption()` listed all five of those later errors but not this
one. Result: the most basic corruption shape skipped `.bak` recovery entirely and
dead-ended as an unhandled error (the symptom in #9627).

Two commits:

1. `e396702` — enroll `InvalidFilePrefixError` in `_isRecoverableCorruption()`.
2. `66e92aa` — gate it, because (1) alone opens a data-loss window.

## Why (2) exists

Multi-agent review found commit (1)'s safety argument **false**. I had claimed the
change was inert against transport-side failures because a mangled transport would
mangle `sync-data.json.bak` identically. Primary and `.bak` are **separate GETs**;
no such invariant exists.

`InvalidFilePrefixError` is the only member of that allow-list that can fire on a
body **we never wrote**. The other five prove the object at the path *is* a sync file
that failed to decode; this one proves it is *not* — which is also exactly what a bad
*response* looks like (proxy page, captive portal, WebDAV 207 body, JSON error
envelope). Adopting `.bak` then heals an **intact** primary with one-generation-old
data, silently dropping ops another device committed and will never re-push.

So this class alone is gated on `_isPrefixFailurePersistent()`: re-read the primary
once, and recover only if the failure belongs to the *stored file* — same rev, still
no valid prefix. **A heuristic, not a proof.**

## Provider audit

The gap is provider-wide, which is why the guard sits in the adapter rather than in
any one provider:

| Provider | non-2xx | empty body | mangled 200 body |
| --- | --- | --- | --- |
| WebDAV | throws | throws `EmptyRemoteBodySPError` | only `isHtmlResponse`'s 3 patterns |
| Dropbox | throws | throws | **no filter** |
| OneDrive | throws | **no check** | **no filter** |
| LocalFile | n/a | throws | **no filter** |

## Tests

All mutation-verified — each fails under targeted mutation of the code it pins:

- transport-asymmetry rejection (primary mangled once, `.bak` intact → reject, don't adopt)
- both-mangled fallthrough (rethrow the original error)
- split call site (`sync-ops.json`)
- **two-client integration test**: client A's committed ops survive a mangled primary
  response on client B (`edge-cases.integration.spec.ts`)

Unit spec 144/144, integration spec 20/20.

## Reporter evidence (2026-08-21)

The reporter answered. First 40 bytes of their `sync-data.json`, on **iOS with sync
encryption on**:

```
41J7VJwUqK/0k436aSIRL5utdxhyV6WhXWSguANW
```

That is base64 — the *encrypted payload is intact*, but the `pf_E2__` header that
belongs in front of it is gone. This confirms the corruption shape this PR targets
exists in the wild on a real user's remote.

### The write path is exonerated

Stronger than "the app cannot produce it" by enumeration: every WebDAV upload ends in
`_verifyUpload()` (`webdav-api.ts:365`) — re-GET the file, hash the stored body, compare
against what was sent. Shipped in **v18.2.6** (`e571cc2433`, #7300), so it was live in
the reporter's 18.19.0. Any head-strip on the way *up* — mangled request body, partial
PUT, rewriting proxy — makes that hash mismatch and throws `RemoteFileChangedUnexpectedly`
at upload time. It cannot leave a prefix-less file behind silently.

With `compressAndEncrypt()` as the only writer (unconditional `prefix + dataStr`, no
chunked uploads, no site writing raw ciphertext), that closes it: **the body was not
produced by a successful app write.** It was mutated after a verified write, or written
by something that is not this app.

### The `.bak` evidence is gone — destroyed by our own remediation advice

The reporter ran force upload at 17:09 UTC and reported the primary now starts with
`pf_`. Force upload routes to `_uploadSnapshot()` → `_forceUploadWithBakFirst()`
(`file-based-sync-adapter.service.ts:1482`), which writes `sync-data.json.bak` **first**,
by design (a rotated-key `.bak` must not survive a snapshot). So the live `.bak` now
holds a fresh, prefixed payload.

Asking about the live `.bak` would return `pf_…` regardless of what it held during the
failure — worse than no answer, because it reads as confirming evidence and is not. The
original merge criterion 1 is unattainable as written.

## Why this is still parked

- **Residual window not closed.** A *deterministic* mangler reproduces on the re-read
  and passes the gate. The dangerous shape is: primary body deterministically mangled
  while the `.bak` body arrives intact and headers pass through unmangled so the rev
  matches. Narrow — it needs content-dependent mangling — but the reporter's case is
  head-stripping of unknown origin, so it is not yet excluded.
- **The corruption *shape* is still unconfirmed.** A 40-char sample cannot distinguish a
  7-byte head-strip (`pf_E2__`) from a file that lost its first N kilobytes. If it is a
  fragment, this PR is the right fix for the wrong shape and the fragment source stays
  open. Byte size settles it; see criterion 1 below.
- **Asymmetric cost.** Waiting costs one already-unblocked user a few days. Merging blind
  risks silent cross-device op loss for every file-based sync user, auto-shipped to real
  devices from `master` within minutes — and silent loss is structurally
  under-reported, so we would likely not hear about it.

## Merge criteria

Merge when **either** holds:

1. The reporter answers from Nextcloud **file version history** (the live files no longer
   carry the evidence):
   - `sync-data.json.bak`, the version from **before** 2026-08-21 ~17:10 UTC — `pf_` or
     base64? Intact `.bak` proves the heal works end-to-end on real data; prefix-less
     `.bak` proves the change is inert for them, and the enrollment gap is still worth
     closing on its own merits.
   - `sync-data.json`, **byte size** of the corrupt version vs. the one before it. ~7
     bytes smaller ⇒ head-strip, the shape this PR targets. Much smaller ⇒ a fragment,
     which is a different bug and this PR does not address it.
2. ~2 weeks pass with no further response. The gap is real independent of this report,
   and the guard is tested.

Merge **both commits or neither** — commit (1) alone is strictly worse than the pair.

## Follow-up hardening (not in this PR)

`_isPrefixFailurePersistent()` currently accepts "both reads failed prefix parse". Comparing
the two bodies byte-for-byte is free and strictly stronger — it additionally rejects *two
different* garbage responses (rotating captive portal, load-balanced error pages), which
the current gate accepts. It does not beat a deterministic mangler; that stays the
acknowledged residual.

Worth considering separately: for this error class only, surfacing a choice instead of
silently adopting `.bak`. Precedent is already in the file — `AUTO_MERGE_CONCURRENT_SNAPSHOT:
false` is defaulted off on exactly the argument that a user-recoverable dialog beats
undetectable data loss. Silent heal-from-`.bak` is safe for the five errors that prove the
file *is* ours; for the one that proves it is not, a prompt matches the existing risk posture.

## Deliberately not included

A `sync-wrapper.service.ts` snack branch for `InvalidFilePrefixError` (siblings
`JsonParseError` and `EmptyRemoteBodySPError` have one; this falls through to the
generic snack). The reporter's case argues the *message* matters more than the button:
if a server-side transformation is stripping the header, force-upload just recreates
the broken state. Worth doing once the cause is known.

## Separate, unfixed

`SuperSyncBackgroundProvider.kt:121` sets `Accept-Encoding: gzip` explicitly, which
disables OkHttp's transparent gunzip, so `resp.body?.string()` yields raw gzip bytes and
`JSONObject(body)` throws into the blanket catch at :135 — a silent no-op. Unrelated to
this PR; needs its own issue.

Refs #9627

### DIFF
--- a/src/app/op-log/sync-providers/file-based/file-based-sync-adapter.service.spec.ts
+++ b/src/app/op-log/sync-providers/file-based/file-based-sync-adapter.service.spec.ts
@@ -2719,6 +2719,39 @@ describe('FileBasedSyncAdapterService', () => {
       expect(mockSnackService.open).toHaveBeenCalled();
     });
 
+    it('(b) recovers from .bak when the primary file has NO sync file prefix (#9627)', async () => {
+      // #9627: a Nextcloud user hit "Invalid sync file prefix" — the remote
+      // primary's first bytes are not `pf_...__`, so extractSyncFileStateFromPrefix
+      // throws BEFORE the decrypt/decompress/JSON stages the cases above cover.
+      // That makes a head-mangled primary (transport-mangled body, a foreign file
+      // at the path, an interrupted write that lost its head) the ONE corruption
+      // shape excluded from .bak recovery. Same recovery semantics as its
+      // siblings: local data is intact and the .bak is the only readable copy.
+      const backupData = createMockSyncData({
+        syncVersion: 5,
+        recentOps: [compactOp('recovered-from-missing-prefix') as never],
+      });
+      // Prefix-less body: valid JSON, but not a sync file (e.g. a backup export
+      // copied into the sync folder). Nothing downstream ever gets to parse it.
+      const unprefixedMain = JSON.stringify({ version: 2, syncVersion: 9 });
+      mockProvider.downloadFile.and.callFake((path: string) => {
+        if (path === FILE_BASED_SYNC_CONSTANTS.BACKUP_FILE) {
+          return Promise.resolve({
+            dataStr: addPrefix(backupData),
+            rev: 'bak-rev-prefix',
+          });
+        }
+        return Promise.resolve({ dataStr: unprefixedMain, rev: 'corrupt-prefix-rev' });
+      });
+
+      const result = await adapter.downloadOps(0);
+
+      expect(result.ops.length).toBe(1);
+      expect(result.ops[0].op.id).toBe('recovered-from-missing-prefix');
+      expect(result.latestSeq).toBe(5);
+      expect(mockSnackService.open).toHaveBeenCalled();
+    });
+
     it('(b) refuses a PLAINTEXT .bak when encryption is expected (key-suppression vector)', async () => {
       // A plaintext .bak decodes via its own prefix flags even under a
       // wrong/rotated key. Accepting it would suppress the wrong-password
--- a/src/app/op-log/sync-providers/file-based/file-based-sync-adapter.service.spec.ts
+++ b/src/app/op-log/sync-providers/file-based/file-based-sync-adapter.service.spec.ts
@@ -16,6 +16,7 @@ import {
   EncryptNoPasswordError,
   FileSyncTargetChangedError,
   InvalidDataSPError,
+  InvalidFilePrefixError,
   PlaintextWhenEncryptionExpectedError,
   RemoteFileNotFoundAPIError,
   SplitSyncFormatDetectedError,
@@ -2731,9 +2732,10 @@ describe('FileBasedSyncAdapterService', () => {
         syncVersion: 5,
         recentOps: [compactOp('recovered-from-missing-prefix') as never],
       });
-      // Prefix-less body: valid JSON, but not a sync file (e.g. a backup export
-      // copied into the sync folder). Nothing downstream ever gets to parse it.
-      const unprefixedMain = JSON.stringify({ version: 2, syncVersion: 9 });
+      // A TRUNCATED head — the interrupted-write shape this recovery targets. It
+      // exercises the regex's partial-match branch (`pf_` present, separator gone),
+      // which a merely prefix-less body would not.
+      const unprefixedMain = 'pf_2_' + JSON.stringify(createMockSyncData({}));
       mockProvider.downloadFile.and.callFake((path: string) => {
         if (path === FILE_BASED_SYNC_CONSTANTS.BACKUP_FILE) {
           return Promise.resolve({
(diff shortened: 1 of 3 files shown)


johannesjo:
Closing. Fresh review turned up an argument that undercuts the enrollment itself, not just its justification.

**`InvalidFilePrefixError` is not the same class as the five errors it would join.** `.bak` recovery exists for torn/interrupted writes. A torn write loses the **tail** — a WebDAV PUT sends a whole body, and Nextcloud writes to a temp file then renames. Losing the **head** is not a shape an interrupted write can produce. So the legitimate case this recovery is designed for essentially cannot generate a prefix error. That is the same "the other five prove the file *is* ours, this one proves it is *not*" reasoning `66e92aa` used to justify gating the class — applied one step further back, it argues against enrolling it at all.

Two more things pushed the same way:

- **The gate is weakest exactly where the only observation came from.** `_isPrefixFailurePersistent()` compares `retry.rev !== primaryRev`, but Nextcloud is constructed with `useCanonicalOcEtag: true`, so `rev` is the **ETag header**, not the body hash. A mangled *body* with intact headers yields an equal rev and the check adds nothing. It only bites on WebDAV servers without strong ETags, where `rev` falls back to `md5(body)`.
- **The reporter's remaining evidence does not support the shape.** Their 288 KB → 1.2 MB comparison is confounded — they turned *off* "enable compression" before the force upload, and 1.2 / 0.288 ≈ 4.2× is a textbook gzip ratio for JSON. Merge criterion 1 was never actually answered.

The reported symptom is fixed separately and much more cheaply: `InvalidFilePrefixError` had no branch in the sync error handler, so it fell through to the generic one and showed the user the raw internal string — verbatim the title of #9627 — with no way forward. It now shares the actionable snack and **Force Overwrite** action of its siblings `JsonParseError` and `EmptyRemoteBodySPError`, which is what unblocked the reporter anyway. ~20 lines, no new data-loss surface.

And the real gap this exposed was diagnostic, not recovery: the error logged how long the body was and what we expected, but nothing about what we *got*. That is now recorded as `prefixAt` + `headShape` (base64 / markup / json / other), which answers "bad response or bad stored file?" straight from a log export — the question this PR spent two weeks unable to settle.

#9683 carries the enrollment asymmetry and the hardening ideas forward, updated with the argument above. Reopen if a report matches its watch signature.

PR #9690, opened 2026-08-22
fix(supersync-chart): decouple liveness from database connectivity

### LINKED ISSUE #9599: SuperSync chart: livenessProbe depends on database connectivity
In `packages/super-sync-server/helm/supersync/templates/deployment.yaml` all three probes point at `/health` (lines 138–156), and `/health` runs `SELECT 1` through Prisma (`src/server.ts:454`). That is right for readiness and startup, but wrong for liveness.

With `periodSeconds: 30` and the default `failureThreshold: 3`, the app pod is killed after roughly 90s of database unavailability. Restarting the application does not fix a database that is down, and it adds a cold start exactly when the database comes back.

I hit this restarting a single-instance CloudNativePG cluster under the server: the database was unreachable for ~145s, `/health` correctly returned 503 throughout, and the pod survived only because Postgres started accepting connections before the operator finished reconciling. A slower restart would have put the app into a restart loop for the duration of the outage.

Worth noting the app handles this well on its own — the Prisma pool reconnected with no process restart once the database returned, so liveness is not needed here at all.

Suggested fix: give liveness a probe that reflects only process health (a `/live` endpoint with no dependency check, or a TCP socket check), and keep `/health` for readiness and startup. Making the probes configurable through values would also let operators tune this per environment.

sanchpet:
Closes #9599.

## What

`livenessProbe` now points at a new `/live` endpoint; readiness and startup stay on `/health`.

## Why

`/health` runs `SELECT 1` through Prisma. That is the right check for readiness and startup — a pod that cannot reach the database cannot serve — but as a liveness check it kills the pod after roughly 90s of database unavailability (`periodSeconds: 30`, default `failureThreshold: 3`). Restarting the application cannot fix a database that is down, and the restart lands a cold start exactly when the database comes back.

On this chart the cost is higher than a cold start: `replicaCount > 1` is rejected because websocket connection state is in-memory, so an unnecessary restart drops every connected client's live-sync socket with no failover.

Observed on a single-instance CloudNativePG restart: the database was unreachable for ~145s, `/health` correctly returned 503 throughout, and the pod survived only because Postgres started accepting connections before the operator finished reconciling. The Prisma pool reconnected with no process restart — liveness was never needed for recovery here.

## Scope

Minimal, as agreed in the issue: `/live` plus the probe repoint, nothing else. No probe knobs in `values.yaml`.

- `/live` returns `{ status: 'ok' }` and touches no dependency; answering it still proves the Node event loop is responsive, which a TCP socket check would not
- rate-limit exempt (`{ config: { rateLimit: false } }`) like `/health`, since probes are frequent
- readiness keeps `/health`, so a pod whose database is gone still leaves the Service endpoints — it just stops being restarted while it waits

## Test

Added to the existing `helm template` step in `supersync-server-tests.yml`, next to the connection-limit assertions: liveness must render `/live`, readiness and startup `/health`. I checked it fails in the intended direction by pointing liveness back at `/health` — the step goes red, and passes again once reverted.

Locally: `tsc --noEmit` clean for `src`, `server-security.spec.ts` 27/27, `config.spec.ts` 34/34, chart renders.

### DIFF
--- a/.github/workflows/supersync-server-tests.yml
+++ b/.github/workflows/supersync-server-tests.yml
@@ -75,6 +75,11 @@ jobs:
             echo 'connectionLimit=0 unexpectedly passed Helm validation' >&2
             exit 1
           fi
+          # Liveness must stay off the DB-gated endpoint; readiness and startup keep it.
+          rendered=$(helm template "${chart_args[@]}")
+          printf '%s\n' "$rendered" | grep -A3 'livenessProbe:' | grep -q 'path: /live'
+          printf '%s\n' "$rendered" | grep -A3 'readinessProbe:' | grep -q 'path: /health'
+          printf '%s\n' "$rendered" | grep -A3 'startupProbe:' | grep -q 'path: /health'
       - name: Run SuperSync Server Tests
         run: cd packages/super-sync-server && npm test
       - name: Apply PostgreSQL Test Schema
--- a/packages/super-sync-server/helm/supersync/templates/deployment.yaml
+++ b/packages/super-sync-server/helm/supersync/templates/deployment.yaml
@@ -148,9 +148,13 @@ spec:
             initialDelaySeconds: 10
             periodSeconds: 5
             failureThreshold: 30
+          # Liveness must not depend on the database: a restart cannot fix a downed
+          # DB, and it would drop every live-sync websocket on the way. Readiness
+          # and startup stay on /health so the pod still leaves the Service
+          # endpoints while the database is unreachable.
           livenessProbe:
             httpGet:
-              path: /health
+              path: /live
               port: http
             periodSeconds: 30
             timeoutSeconds: 5
--- a/packages/super-sync-server/src/server.ts
+++ b/packages/super-sync-server/src/server.ts
@@ -467,6 +467,17 @@ export const createServer = (
         }
       });
 
+      // Liveness - process health only, deliberately free of dependency checks.
+      // Restarting the process cannot fix a database that is down, and the restart
+      // lands a cold start exactly when the database comes back; on this chart it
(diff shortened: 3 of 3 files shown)


johannesjo:

**Repoints only the liveness probe at a new dependency-free `/live` endpoint, exactly as scoped in #9599 — endpoint, chart, and CI assertion all check out; nothing blocking.**

### Nits
- **`packages/super-sync-server/helm/supersync/templates/deployment.yaml:155`** — chart/image skew footgun: an operator who pinned `image.tag` to an older `master-<sha>` (as the `values.yaml` comment recommends for reproducibility) — or a node holding a stale cached `latest` under `pullPolicy: IfNotPresent` — gets a chart probing `/live` against an image that 404s it. httpGet success is 200–399, so liveness fails 3×30s and the pod is killed in a ~90s loop while startup/readiness (on `/health`) keep it Ready — the exact websocket-dropping churn this PR removes. The chart already couples to the image this way (the `migrate-db` init container), so a one-line note is enough: extend the new comment block above `livenessProbe` (or the `tag:` comment in `values.yaml`) saying the image must serve `/live`, i.e. upgrade image and chart together when pinning.
- **`packages/super-sync-server/src/server.ts:477`** — nothing server-side pins the route's existence: if `/live` is later renamed or removed, `npm test` and the helm greps both stay green, and the break surfaces as a deploy-time restart loop. A small test injecting `GET /live` against the real server would close that — though fair warning it isn't a one-liner today, since routes register inside `start()` (which also listens and spins up background jobs), and `/health` has no such test either. Optional; the helm assertion you added may be the right amount of guard for this fix.

---

johannesjo:

PR #9691, opened 2026-08-22
fix(supersync): keep live websocket clients out of the stale-device sweep

### LINKED ISSUE #9598: SuperSync: lastSeenAt only updated on upload, so read-only devices look stale and get pruned
Running the self-hosted sync server (`v18.19.0`, Postgres). Two of my devices are in `sync_devices`; one of them had been pulling every day but had not uploaded anything for three days. The server still reported its `last_seen_at` as three days old.

The `syncDevice` upsert lives inside the upload transaction only (`packages/super-sync-server/src/sync/sync.service.ts:477`, `:482`). Nothing on the download path touches it, so a device that pulls but does not push is indistinguishable from one that is gone.

Two consequences:

- `getOnlineDeviceCount` (used by `/api/sync/status` as `devicesOnline`) undercounts — a device actively reading is not counted as online.
- `cleanup.ts:46` calls `deleteStaleDevices(cutoffTime)` with `RETENTION_DAYS = 45`, so a device that only ever reads has its registration deleted after 45 days while still in active use.

The second one is the reason I am reporting it: the device list is where you go to revoke access, and it quietly stops listing devices that are still reading your data. A tablet or second laptop used just to look at the day's plan is exactly that case.

Separately, `lastAckedSeq` is set to `0` on create (`sync.service.ts:479`) and never updated anywhere — it looks like a leftover.

Suggested fix: touch `lastSeenAt` on the download path too. Happy to send a PR if you agree with the direction.

sanchpet:
Closes #9598.

## Scope

The download path already landed in master (`sync.routes.ts` → `SyncService.touchDevice`, throttled in SQL, fire-and-forget), so this PR is the remaining half: **the websocket path**, touched periodically while connected rather than only on accept.

Framed as you asked — this keeps the device registry accurate and stops the retention sweep deleting devices that are in active use. It is not about a revocation screen; there is no device-list or per-device revoke endpoint today, and `devicesOnline` is only exposed through the diagnostic `/api/sync/status`.

## Why the websocket path still needs it

The download touch only fires when a device actually pulls ops, and it pulls when *some other* device uploaded something. A device that holds an open socket, uploads nothing and receives nobody else's ops is never touched at all — and after `RETENTION_DAYS` it is deleted while demonstrably alive. Touching on accept alone would not cover it either, which is why this hooks the heartbeat.

## Implementation

- periodic touch from the existing 30s heartbeat, gated by an in-memory `lastTouchedAt` per connection against the shared `DEVICE_TOUCH_THROTTLE_MS` — about one statement per interval per live socket rather than one per tick, and the SQL throttles again on top
- the touch fn is **injected** at `startHeartbeat(...)` rather than imported, so `WebSocketConnectionService` stays persistence-agnostic: it knows when a client is alive, not how a device row is written
- `last_acked_seq` documented as reserved (written at registration, never read or updated) instead of dropped — removing a `NOT NULL` column is a destructive migration for self-hosters with nothing gained

## Test

Two cases in `websocket-connection.service.spec.ts`: one pinning the throttle (no write inside the first window, exactly one after crossing it, none on the ticks that follow) and one asserting the heartbeat still pings when no touch fn is injected. I checked the first fails in the intended direction — removing the touch block turns it red, 1 failed / 28 passed, and green again once restored.

Locally: `tsc --noEmit` clean for `src`, `websocket-connection.service.spec.ts` 29/29, prettier clean.

## Your question about read-only traffic

Real, and more common than the polling case would suggest. For SuperSync the client has **no periodic upload and no polling**: `sync-trigger.service.ts` merges event triggers (idle, wake, online, visibility, resume) with `_onUpdateLocalDataTrigger$`, and the interval timer is explicitly file-provider-only — the comment there says SuperSync relies on websocket push instead. So a client that changes nothing uploads nothing, indefinitely.

That is exactly the shape this PR covers: such a device typically sits on an open socket, and its only chance of being touched before this change was another device happening to upload.

### DIFF
--- a/packages/super-sync-server/prisma/schema.prisma
+++ b/packages/super-sync-server/prisma/schema.prisma
@@ -147,6 +147,7 @@ model SyncDevice {
   deviceName   String? @map("device_name")
   userAgent    String? @map("user_agent")
   lastSeenAt   BigInt  @map("last_seen_at")
+  /// Reserved: written once at device registration, never read or updated.
   lastAckedSeq Int     @default(0) @map("last_acked_seq")
   createdAt    BigInt  @map("created_at")
 
--- a/packages/super-sync-server/src/api.ts
+++ b/packages/super-sync-server/src/api.ts
@@ -242,6 +242,14 @@ export const apiRoutes = async (
         // AUTH_CACHE_INVALIDATION: account deletion must not leave a ghost-token window.
         authCache.invalidate(userId);
 
+        // The cascade removed this user's sync_devices rows, but an open socket
+        // keeps answering pings, so the dead-connection branch never reaps it.
+        // Its heartbeat touch would then re-INSERT a device row for a user that
+        // no longer exists and trip the FK every throttle window. Closed after
+        // the delete, not before: with the user row already gone no reconnect
+        // can re-authenticate and re-orphan a socket.
+        getWsConnectionService().closeForUser(userId);
+
         Logger.audit({ event: 'USER_ACCOUNT_DELETED', userId });
 
         return reply.send({ success: true });
--- a/packages/super-sync-server/src/server.ts
+++ b/packages/super-sync-server/src/server.ts
@@ -491,8 +491,11 @@ export const createServer = (
       // Start cleanup jobs
       startCleanupJobs();
 
-      // Start WebSocket heartbeat
-      getWsConnectionService().startHeartbeat();
+      // Start WebSocket heartbeat. It also refreshes the device row of every live
+      // socket, so a long-lived read-only connection is not pruned as stale.
+      getWsConnectionService().startHeartbeat((userId, clientId) =>
+        getSyncService().touchDevice(userId, clientId),
+      );
 
       try {
         const address = await fastifyServer.listen(createListenOptions(fullConfig));
--- a/packages/super-sync-server/src/sync/services/websocket-connection.service.ts
+++ b/packages/super-sync-server/src/sync/services/websocket-connection.service.ts
@@ -1,5 +1,6 @@
 import { WebSocket } from 'ws';
 import { Logger } from '../../logger';
+import { DEVICE_TOUCH_THROTTLE_MS } from '../sync.types';
 
 interface ConnectedClient {
   ws: WebSocket;
@@ -8,6 +9,13 @@ interface ConnectedClient {
   lastPong: number;
   /** Wall-clock ms when this socket was accepted (used for the storm summary). */
   connectedAt: number;
+  /**
+   * Wall-clock ms of the last device touch issued for this socket. Purely an
+   * in-memory throttle so a long-lived connection costs ~one write per
+   * DEVICE_TOUCH_THROTTLE_MS rather than one per heartbeat tick; the statement
+   * itself is throttled again in SQL, so a stale value here is harmless.
+   */
+  lastTouchedAt: number;
   /**
    * Wall-clock ms at which the reconnect cooldown expires. Set on accept and
    * extended forward on each refused challenger so a sustained storm cannot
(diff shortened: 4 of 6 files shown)


johannesjo:

**Implements the websocket half of #9598 along exactly the lines agreed in the issue — injected touch fn, in-memory throttle on the existing heartbeat, comment-only schema change — and the core logic checks out: the touch sits after the dead-connection check so only live sockets are touched, `SyncService.touchDevice` is internally fire-and-forget so nothing can reject into the interval, the guard-before-assignment order in `startHeartbeat` means a redundant no-arg restart can't clobber the wiring, and the worst-case refresh cadence stays under `ONLINE_DEVICE_THRESHOLD_MS`. Nits only.**

### Nits

- **`packages/super-sync-server/src/sync/services/websocket-connection.service.ts:360`** — this touch is the first DB write ever issued on behalf of a socket, and account deletion doesn't close sockets: `DELETE /account` (`src/api.ts`) invalidates the auth cache and deletes the user, but unlike `/replace-token` and passkey recovery it never calls `getWsConnectionService().closeForUser(userId)`. The cascade removes the `sync_devices` rows, the orphaned socket keeps answering pings (so the dead-connection branch never removes it), and every ~2 min the touch's `INSERT` violates `sync_devices_user_id_fkey` — swallowed to `Logger.debug` in the app but repeating in the Postgres error log until the client goes away. Pre-existing gap (the orphan used to be inert; this PR makes it noisy) — a one-line `closeForUser(userId)` in the delete-account route would close it and fits this PR's scope.
- **`packages/super-sync-server/src/sync/services/websocket-connection.service.ts:357`** — the in-memory gate and the SQL `WHERE` both use `DEVICE_TOUCH_THROTTLE_MS`, and with the strict `<` on a 30s tick grid the aligned case refreshes the row only every ~240s against the 300s `ONLINE_DEVICE_THRESHOLD_MS`. Fine today, but a future bump of the constant (or drop of the threshold) starts flickering `devicesOnline` silently. Gating in memory at `DEVICE_TOUCH_THROTTLE_MS / 2` would keep the SQL clause the single authority on cadence.
- **`packages/super-sync-server/tests/websocket-connection.service.spec.ts:553`** — "keeps pinging when no touch fn is injected" duplicates "should send ping at interval" (line 566), which also calls `startHeartbeat()` with no argument and asserts a ping. Fine to keep for intent documentation; one of the two can go.
- **`packages/super-sync-server/src/sync/services/websocket-connection.service.ts:360`** — the injected fn resolves to `DeviceService.touchDevice`, whose `INSERT ... ON CONFLICT` half *creates* rows — so an authenticated client holding a socket with a never-synced (shape-valid) `clientId` materializes a device row after 2 min, visible in `GET /api/sync/devices`. This matches the download path's documented behavior on master (the `device.service.ts` doc comment records the INSERT half as deliberate), so presumably intended — flagging only so the second call site doesn't pass unnoticed.
- **PR body** — two framing corrections worth folding in, since the issue asked precisely how common the uncovered case is: (a) "a client that changes nothing uploads nothing, indefinitely" is true, but every sync *run* still downloads unconditionally (`operation-log-download.service.ts:188`) and the download route touches on it — and the SuperSync trigger set is presence-based (idle, wake, online, visibility), not change-based. So the population this PR covers is narrower than "read-only clients": devices holding a socket with zero presence-triggered syncs across 45 days (always-on/headless setups). Still real, just smaller. (b) "there is no device-list … endpoint today" is stale at this head — `GET /api/sync/devices` (`sync.routes.ts:261`) has since landed on master, which strengthens the motivation rather than weakening it.

---

johannesjo:
**✅ Ready to merge** – no blocking issues at `96c8ed6`. Merge confidence 92% – I read both new specs and every touched path, and an independent verification pass confirmed the delete-account teardown, but I ran nothing locally; execution evidence is CI's green server tests and E2E gates at this head.

---
<sub>Automated review pass (Claude Code).</sub>

johannesjo:
Thank you very much!

Issue #9692, opened 2026-08-22
SuperSync: old-ops cleanup sweep fails on every run since #9670 — no user gets cleaned


johannesjo:
# SuperSync: old-ops cleanup sweep fails on every run since #9670 — no user gets cleaned

Follow-up to #9670 (`18aa506661`), deployed to production 2026-08-21 ~18:55 UTC
(image revision `a6a2beacdb`).

Since that deploy the daily old-ops sweep has failed on **every** run, deleting
nothing for anyone. Before #9670 the long-lapsed cohort was skipped but the rest
of the fleet was cleaned; now the run aborts before any deletion, so cleanup is
effectively off instance-wide.

## Evidence

First scheduled run after deploy (`INITIAL_CLEANUP_DELAY_MS = 10_000`, so ~10 s
after container start):

```
2026-08-21T18:55:27.046Z [ERROR] Cleanup [old-ops] failed: PrismaClientUnknownRequestError:
Invalid `prisma.operation.findMany()` invocation:

Error occurred during query execution:
ConnectorError(ConnectorError { user_facing_error: None, kind: QueryError(PostgresError {
  code: "57014", message: "canceling statement due to statement timeout",
  severity: "ERROR", detail: None, column: None, hint: None }), transient: false })
2026-08-21T18:55:27.133Z [INFO] Cleanup [stale-devices]: removed 21 entries
2026-08-21T18:55:27.164Z [INFO] Cleanup [rate-limits]: removed 43 entries
```

`57014` is the 60 s `statement_timeout`, confirmed set at database level:

```
SELECT setting, source FROM pg_settings WHERE name = 'statement_timeout';
 setting | 60000
 source  | database
```
```
SELECT ... FROM pg_db_role_setting;
 database  | supersync
 role      | (all roles)
 setconfig | {statement_timeout=60s}
```

No `Cleanup [old-ops]: removed N entries` line has appeared since.

## Mechanism

1. `deleteOldSyncedOpsForAllUsers` (`storage-quota.service.ts:411`) fetches
   eligible users with `orderBy: { snapshotAt: 'asc' }` (`:429`) —
   **stalest first**, deliberately, so the most-drifted users reconcile first.
2. #9670 removed the `snapshotAt >= cutoffTime` gate (`:448` documents why —
   it was inverted against `deleteStaleDevices` and let lapsed accounts
   accumulate operations forever).
3. Those two combine badly: the newly-included lapsed cohort has by far the
[...]

PR #9696, opened 2026-08-22
fix(super-sync): stop one cancelled query aborting the whole old-ops cleanup sweep

johannesjo:
## Problem

Closes part of #9692.

Since #9670 the SuperSync daily old-ops cleanup sweep has failed on **every**
run. `deleteOldSyncedOpsForAllUsers` iterates every eligible user; for one
account the candidate selection in `deleteOldSyncedOpsBatch` hits the
PostgreSQL `statement_timeout` (57014, surfacing as
`PrismaClientUnknownRequestError`). That throw escaped the per-user loop and was
caught at run level, so:

- **no** user got cleaned, fleet-wide, every night;
- both summary warnings and the deferred storage-reconcile scheduling died with
  the throw;
- the `if (totalDeleted > 0)` guard around the summary log meant a zero-deletion
  sweep left no diagnostic at all — the outage was invisible.

## Solution

This PR contains the failure and makes the outage observable. **It is not the
root-cause fix** — see "Deliberately not included".

- **Per-user containment.** The per-user body moves into `sweepOldOpsForUser`
  and a cancelled statement is contained there, so the rest of the fleet is
  still swept. The guard matches the class *together with* the
  `canceling statement due to statement timeout` message:
  `PrismaClientUnknownRequestError` is Prisma's catch-all for engine failures it
  could not map to a `P####` code (it is thrown verbatim for
  "Response from the Engine was empty"), and SQLSTATE 57014 alone also covers
  `pg_cancel_backend`, lock timeout — which `migrate-deploy.sh:329` shows this
  deployment produces — and recovery conflict. None of those are a permanent
  per-user defect. `monitoring-db.ts` already required the same full phrase.
- **`OLD_OPS_CLEANUP_MAX_DELETED_PER_USER_PER_RUN`.** Users are swept
  stalest-snapshot-first and the drain loop ran until the *global* budget was
  gone, so one long-lapsed account could absorb the entire 25 000/day run. The
  cap defaults to the **resolved** batch size, and a warning now fires when the
  cap rather than the run budget is what stopped a sweep. Exposed in
  `env.example`, `docker-compose.yml` and the Helm chart — compose lists env
  explicitly, so it would otherwise have been unsettable.
- **Observability.** The summary is logged unconditionally, `failedUserIds` is
  returned to the caller, and a re-entrancy guard stops `setInterval` stacking a
  second run on top of one that overran the day.
- **Measurement.** A new plan spec pins what `take:` actually bounds.

### Behaviour change worth reviewing

The per-user cap defaulting to the resolved batch size makes it **one batch per
user per run** by default, so #9670's drain loop spans batches only when an
operator raises the cap. This is the semantics `helm/supersync/values.yaml`
already documented. It also fixes a real regression: tying the cap to the
compile-time constant made `OLD_OPS_CLEANUP_DELETE_BATCH_SIZE=7000` delete 5 000
and silently drop the rest.

### What the measurement found

`EXPLAIN (ANALYZE, BUFFERS)` under `force_generic_plan`, 20 000-op user slice:

| ordering | history inside retention | history outside retention, ancient `protectedFromSeq` |
| --- | --- | --- |
| `server_seq` (current) | 20 000 filtered / 223 blk | 0 filtered / **3 blk** |
| `received_at` | 0 filtered / 2 blk | 19 996 filtered / **247 blk** |

`take` bounds the rows *returned*, not *scanned*, so **lowering the batch size
is not a fix**. Nor is the obvious repair of ordering by `received_at`:
whichever column the sort picks becomes the index's range bound and the other
predicate degrades to a heap filter, so exactly one cohort always runs away —
and the second column is the long-lapsed-account shape #9692 is about. Both
cohorts have assertions, so an ordering swap fails the guard instead of looking
like progress.

## Deliberately not included

- **The root-cause fix**: bounding the scanned range with a `server_seq` window.
  It changes the deletion predicate on a live 8.44 M-row table and needs a real
  PG16 reproduction first — PGlite is PG18 and these planners disagree on this
  query class. `tests/integration/old-ops-sweep.integration.spec.ts` seeds ≤5
  ops/user today, so it needs a fixture past the batch size before it can
  measure anything.
- **A run wall-clock deadline.** Contained failures consume no delete budget,
  which is the loop's only exit, so run cost is proportional to the number of
  timing-out accounts. Every bounded-run variant has the same starvation shape
  under stalest-first ordering, and the whole question dissolves once the scan
  is bounded. The overlap guard is the cheap half and is included.
- Note that `statement_timeout` is **opt-in and off by default**
  (`env.example`), so on a stock self-hosted deploy no 57014 is raised and this
  containment never fires — the scan just holds a pool connection. Only bounding
  the scan helps those deployments.

## Type of Change

- [x] Bug fix

## Checklist

Tests: 1 177 pass. Each new test was verified to fail without its fix —
containment, the raised batch size, the per-user-cap warning, the zero-deletion
summary log, the overlap guard, and both plan-spec cohorts.


johannesjo:
Superseded.

The headline failure — one cancelled query aborting the whole fleet's sweep — was fixed on master by c57560ecddca (per-user error isolation in `deleteOldSyncedOpsForAllUsers`), confirmed working in the 2026-08-27 production run (individual users still failed, but the sweep continued and cleaned others).

The per-user delete cap this PR adds can't ship as-is: stopping a user's drain mid-prefix leaves a truncated aged prefix behind, which violates the whole-or-nothing prefix rule — snapshot restores then fail with `SNAPSHOT_REPLAY_INCOMPLETE`. The budget must gate which users *start*, never where a drain stops.

The remaining per-user timeouts are fixed on `feat/issue-9692-782798` instead, by bounding each DELETE statement to a stated `serverSeq` window (no discovered row sets, so cost is bounded regardless of match density, tuple liveness, or plan choice) plus a covering `(user_id, received_at, server_seq)` index so the fresh-prefix probe answers index-only. Production findings and details in #9692.

Issue #9563, opened 2026-08-13
💥 Keyboard shortcut to "schedule task for today" not working


jonathantamen:
### Steps to Reproduce
On MacOS. Version 18.19.0D.

1. Created new task with no attributes.
2. Selected the task
3. Pressed the keyboard shortcut to schedule task for today but it does not get scheduled. Instead, the selector moves to the next available task with no changes made. 

*Using the keybind to schedule to any other time (such as "schedule task for tomorrow") works as intended.
*Pressing the button to schedule task for today works as intended.

### Additional Console Output

log.ts:207  [a][Layout] Show AddTaskBar
log.ts:322 [task] addTask {taskId: 'IM-Pp_iqjUEWI92N7QbnT', workContextId: 'INBOX_PROJECT', workContextType: 'PROJECT'}
log.ts:207  [a][Task Shared] addTask
log.ts:334 [ol] OperationCaptureService: Captured action {actionType: '[Task Shared] addTask', pendingCount: 1}
log.ts:334 [ol] operationCaptureMetaReducer: Captured action for operation persistence {actionType: '[Task Shared] addTask'}
log.ts:322 [task] shortSyntax {taskId: 'IM-Pp_iqjUEWI92N7QbnT', hasResult: false, changedFields: Array(0), attachmentCount: 0, projectId: undefined}
log.ts:334 [ol] incrementVectorClock {clientId: 'B_tJ3qY7', currentValue: 781, allClients: Array(4)}
log.ts:334 [ol] OperationCaptureService: Processed action {pendingCount: 0}
log.ts:207  [a][Layout] Hide AddTaskBar
log.ts:207  [a][Task] SetSelectedTask
log.ts:334 [sync] SyncCredentialStore.load SuperSync object
log.ts:334 [ol] ImmediateUploadService: Starting immediate upload...
log.ts:353 [ol] OperationWriteFlushService: Starting flush. Initial pending count: 0
log.ts:353 [ol] OperationWriteFlushService: Flush complete in 1ms. Initial pending: 0, Final pending: 0
log.ts:353 [ol] OperationWriteFlushService: Starting flush. Initial pending count: 0
log.ts:353 [ol] OperationWriteFlushService: Flush complete in 0ms. Initial pending: 0, Final pending: 0
log.ts:353 [ol] OperationLogUploadService: Uploading pending operations via API...
log.ts:334 [sync] SyncCredentialStore.load SuperSync object
[...]

PR #9701, opened 2026-08-22
fix(tasks): apply the same Shift+T guards on both host paths #9577

### LINKED ISSUE #9577: Shift+T behaves differently on <planner-task> hosts than on <task> hosts
Split out of #9563.

`taskScheduleToday` (Shift+T) reaches "schedule for today" by two different routes depending on which component renders the focused row (`src/app/features/tasks/task-shortcut.service.ts:95-109`):

- a live `<task>` host delegates to `TaskComponent.scheduleForTodayWithFocus()`
- anything else (e.g. `<planner-task>` in the Planner overdue list) falls back to `TaskService.scheduleForTodayById()`

After the #9563 fix the `<task>` path skips two cases that the fallback still performs:

1. **Task already on Today.** `handlePlanTasksForToday` applies `remindAt: undefined` unconditionally (`src/app/root-store/meta/task-shared-meta-reducers/task-shared-scheduling.reducer.ts:243-250`), while `shouldClearDueTimeForToday` returns `false` for a `dueWithTime` that is today. So re-planning a task due at a time today silently drops its reminder and writes both `dueDay` and `dueWithTime` (a state ARCHITECTURE-DECISIONS #1 tolerates on read but should not be created on write).
2. **Done tasks.** Completion never synthesizes a `dueDay` (`src/app/features/tasks/store/task-related-model.effects.ts:77-81`). Dating a done task adds it to `daily-summary.component.ts:556`, which admits on `dueDay === dayStr` with no `isDone` guard, inflating today's done count at `:200`.

So the same keypress behaves differently based on the DOM host. Pre-existing (the fallback never had these guards), but the gap is wider now that the `<task>` path is stricter.

**Suggested fix:** extract a small pure `isTaskOnToday(task, todayStr, startOfNextDayDiffMs)` util next to `is-task-overdue.ts` and apply the same two guards in both places. Note `TaskService` is currently 1525 lines against the 1200-line `max-lines` cap (grandfathered), so the shared logic should live in the util, not grow the service.

Also worth folding in: a sub-task whose parent is already on Today still dispatches a `planTasksForToday` that the reducer rejects (`task-shared-scheduling.reducer.ts:197-204`) — a captured, encrypted, uploaded op with no state change.

sujeito-operator:
Fixes #9577.

`taskScheduleToday` (Shift+T) reaches "schedule for today" by two routes depending on which component renders the focused row (`task-shortcut.service.ts:95-109`): a live `<task>` host delegates to `TaskComponent.scheduleForTodayWithFocus()`, anything else — `<planner-task>` in the Planner overdue list — falls back to `TaskService.scheduleForTodayById()`. After #9563 the `<task>` path grew two guards that the fallback never had, so the same keypress does different things based on the DOM host.

### The change

Per the issue's suggestion, a small pure `isTaskOnToday(task, todayStr, startOfNextDayDiffMs)` next to `is-task-overdue.ts`, and both call sites now ask it:

- `TaskService.scheduleForTodayById()` gains the two guards.
- `TaskComponent.isScheduledToday()` is re-expressed in terms of it rather than keeping a second inline definition, so the two cannot drift again.

The shared logic lives in the util, not in `TaskService` — the issue notes the service is already 1525 lines against the grandfathered 1200-line `max-lines` cap, and this does not add to it.

**The two cases the fallback was performing:**

1. **Task already on Today.** `handlePlanTasksForToday` applies `remindAt: undefined` unconditionally (`task-shared-scheduling.reducer.ts:243-250`), while `shouldClearDueTimeForToday` returns `false` for a `dueWithTime` that is today — so re-planning a task due at a time today silently dropped its reminder and wrote both `dueDay` and `dueWithTime`.
2. **Done tasks.** Completion never synthesizes a `dueDay` (`task-related-model.effects.ts:77-81`); dating a done task admits it to `daily-summary.component.ts:556`, which has no `isDone` guard, inflating today's done count.

### On the sync model

`docs/sync-and-op-log/contributor-sync-model.md` — no effect or reducer is touched, and no new dispatch is introduced. The guards only *suppress* a dispatch, so the change strictly reduces the operations emitted for a keypress that produced no state change.

### Deliberately not included

The issue also mentions folding in the sub-task case (a sub-task whose parent is already on Today dispatches a `planTasksForToday` the reducer rejects at `:197-204`). That needs the TODAY tag's membership rather than the task entity alone, so it is a wider change than the host-parity fix and is left out to keep this diff reviewable. Happy to follow up.

### One existing test changed, and why

`task-shortcuts.spec.ts` — *"keeps the reminder of a task due at a time today"* stubbed `DateService.isToday` to return `true` over a fixture whose `dueWithTime` was `1746453600000` = **2025**-05-05, while the block's `TODAY` is **2026**-05-05. The assertion only held because of the stub. `isScheduledToday()` no longer calls `DateService.isToday`, so the fixture is now a real `TODAY`-local timestamp and the `Object.defineProperty` stub is gone — the test's own comment flagged that stub as awkward.

### Verification

Negative control, per the issue's "needs a test that fails without the fix": with only the two source files reverted and the specs kept, `task.service.spec.ts` is **3 FAILED / 75 SUCCESS** — the three guard tests, exactly. With the fix, the four relevant specs (`task.service`, `is-task-on-today`, `task-shortcut.service`, `task-shortcuts`) are **153/153 SUCCESS** under `TZ='Europe/Berlin'`.

`npm run lint:ts`: **0 errors, 90 warnings — identical to the count on a clean tree.** `tsc --noEmit -p tsconfig.json`: clean.

`isTaskOnToday` has its own spec covering `dueDay`, `dueWithTime`, precedence, the unscheduled cases, and the start-of-next-day offset in both directions (02:00 with a 4h offset still belongs to the previous logical day).

---

*Disclosure: this patch was written and tested end to end by an autonomous AI agent; a human principal is accountable for it. [What this account is](https://github.com/sujeito-operator). Ask me anything about how it was produced and I will answer.*

### DIFF
--- a/src/app/features/tasks/task.service.spec.ts
+++ b/src/app/features/tasks/task.service.spec.ts
@@ -311,6 +311,76 @@ describe('TaskService', () => {
     });
   });
 
+  // The Shift+T fallback for rows that are not rendered by a live <task>
+  // (e.g. <planner-task> in the Planner overdue list). It has to apply the same
+  // two guards TaskComponent.scheduleForToday() applies, or the same keypress
+  // behaves differently based on the DOM host. (#9577)
+  describe('scheduleForTodayById', () => {
+    it('should dispatch planTasksForToday for a task that is not on today', () => {
+      const task = createMockTask('task-1', { dueDay: '2026-01-04' });
+      store.overrideSelector(selectTaskEntities, { ['task-1']: task });
+      store.refreshState();
+
+      service.scheduleForTodayById('task-1');
+
+      expect(store.dispatch).toHaveBeenCalledWith(
+        TaskSharedActions.planTasksForToday({
+          taskIds: ['task-1'],
+          today: '2026-01-05',
+          startOfNextDayDiffMs: 0,
+          // computed key: a quoted 'task-1' trips the naming-convention rule
+          parentTaskMap: { ['task-1']: undefined },
+        }),
+      );
+    });
+
+    it('should not re-plan a task whose dueDay is already today', () => {
+      const task = createMockTask('task-1', { dueDay: '2026-01-05' });
+      store.overrideSelector(selectTaskEntities, { ['task-1']: task });
+      store.refreshState();
+
+      service.scheduleForTodayById('task-1');
+
+      expect(store.dispatch).not.toHaveBeenCalled();
+    });
+
+    it('should not drop the reminder of a task due at a time today', () => {
+      // planTasksForToday applies `remindAt: undefined` unconditionally, so
+      // re-planning this task would silently clear its reminder.
+      const task = createMockTask('task-1', {
+        dueWithTime: new Date(2026, 0, 5, 14, 30).getTime(),
+        remindAt: new Date(2026, 0, 5, 14, 0).getTime(),
+      });
+      store.overrideSelector(selectTaskEntities, { ['task-1']: task });
+      store.refreshState();
+
+      service.scheduleForTodayById('task-1');
+
+      expect(store.dispatch).not.toHaveBeenCalled();
+    });
+
+    it('should not date a done task', () => {
+      // Completion never synthesizes a dueDay; dating a done task instead
+      // inflates the daily summary's done count for today.
+      const task = createMockTask('task-1', { isDone: true });
+      store.overrideSelector(selectTaskEntities, { ['task-1']: task });
+      store.refreshState();
+
+      service.scheduleForTodayById('task-1');
+
+      expect(store.dispatch).not.toHaveBeenCalled();
+    });
+
+    it('should still dispatch for an unknown id, leaving the reducer to drop it', () => {
+      store.overrideSelector(selectTaskEntities, {});
+      store.refreshState();
+
+      service.scheduleForTodayById('missing');
+
+      expect(store.dispatch).toHaveBeenCalled();
+    });
+  });
+
   describe('remove', () => {
     it('should dispatch deleteTask', () => {
       const task = createMockTaskWithSubTasks(createMockTask('task-1'));
--- a/src/app/features/tasks/task.service.ts
+++ b/src/app/features/tasks/task.service.ts
@@ -105,6 +105,7 @@ import { TaskFocusService } from './task-focus.service';
 import { DeletedTaskIssueSidecarService } from '../issue/two-way-sync/deleted-task-issue-sidecar.service';
 import { TimeBlockDeleteSidecarService } from '../calendar-integration/time-block/time-block-delete-sidecar.service';
 import { getDeadlineAutoPlanFields } from './util/get-deadline-auto-plan-fields';
+import { isTaskOnToday } from './util/is-task-on-today';
 import { TaskTimeSyncService } from './task-time-sync.service';
 
 @Injectable({
@@ -465,11 +466,26 @@ export class TaskService {
    */
   scheduleForTodayById(taskId: string): void {
     const task = this._taskEntities()[taskId];
+    const today = this._dateService.todayStr();
+    const startOfNextDayDiffMs = this._dateService.getStartOfNextDayDiffMs();
+
+    // The same two guards TaskComponent.scheduleForToday() applies, so Shift+T
+    // does not depend on whether a <task> or a <planner-task> renders the row.
+    // (#9577)
+    // Already on Today: planTasksForToday clears remindAt unconditionally, so
+    // re-planning a task due at a time today silently drops its reminder and
+    // writes both dueDay and dueWithTime.
+    // Done: completion never synthesizes a dueDay, and dating a done task adds
+    // it to the daily summary's done count for today.
+    if (task && (task.isDone || isTaskOnToday(task, today, startOfNextDayDiffMs))) {
+      return;
+    }
+
     this._store.dispatch(
       TaskSharedActions.planTasksForToday({
         taskIds: [taskId],
-        today: this._dateService.todayStr(),
-        startOfNextDayDiffMs: this._dateService.getStartOfNextDayDiffMs(),
+        today,
+        startOfNextDayDiffMs,
         parentTaskMap: task ? { [taskId]: task.parentId } : undefined,
       }),
     );
--- a/src/app/features/tasks/task/task-shortcuts.spec.ts
+++ b/src/app/features/tasks/task/task-shortcuts.spec.ts
@@ -717,13 +717,14 @@ describe('TaskComponent shortcut handling', () => {
     it('keeps the reminder of a task due at a time today', () => {
       // planTasksForToday clears remindAt unconditionally, so re-planning a task
       // that is already on Today would silently drop its reminder.
-      // isToday is installed as a property on the DateService mock, so it has
-      // to be redefined rather than assigned.
-      Object.defineProperty(dateService, 'isToday', {
-        value: () => true,
-        configurable: true,
+      // The timestamp is a real TODAY-local one: isScheduledToday() now derives
+      // the day from todayDateStr + the start-of-next-day offset (isTaskOnToday),
+      // so it no longer needs DateService.isToday to be stubbed true over a
+      // fixture that fell on a different day.
+      setTask({
+        dueWithTime: new Date(2026, 4, 5, 16, 0).getTime(),
+        remindAt: new Date(2026, 4, 5, 15, 30).getTime(),
       });
-      setTask({ dueWithTime: 1746453600000, remindAt: 1746452000000 });
 
       component.scheduleForToday();
 
--- a/src/app/features/tasks/task/task-shortcuts.spec.ts
+++ b/src/app/features/tasks/task/task-shortcuts.spec.ts
@@ -169,9 +169,14 @@ describe('TaskComponent shortcut handling', () => {
           provide: DateService,
           useValue: jasmine.createSpyObj(
             'DateService',
-            ['isToday', 'getLogicalTodayDate'],
+            ['isToday', 'getLogicalTodayDate', 'getStartOfNextDayDiffMs'],
             {
               isToday: () => false,
+              // isScheduledToday() calls this and is template-bound
+              // (task.component.html:166, :192), so the first test in this file
+              // that calls detectChanges() renders those @if blocks against this
+              // mock. Stubbed here rather than only in the Shift+T beforeEach.
+              getStartOfNextDayDiffMs: () => 0,
             },
           ),
         },
--- a/src/app/features/tasks/task/task.component.ts
+++ b/src/app/features/tasks/task/task.component.ts
@@ -122,6 +122,7 @@ import {
 } from '../add-subtask-input/add-subtask-input.component';
 import { AddSubtaskInputService } from '../add-subtask-input/add-subtask-input.service';
 import { getSubTaskTimeLeftForDisplay } from '../util/get-sub-task-time-left-for-display';
+import { isTaskOnToday } from '../util/is-task-on-today';
 
 @Component({
   selector: 'task',
@@ -263,14 +264,13 @@ export class TaskComponent implements OnDestroy, AfterViewInit {
           t.dueDay < todayStr))
     );
   });
-  isScheduledToday = computed(() => {
-    const t = this.task();
-    const todayStr = this.globalTrackingIntervalService.todayDateStr();
-    return (
-      (t.dueWithTime && this._dateService.isToday(t.dueWithTime)) ||
-      (t.dueDay && t.dueDay === todayStr)
-    );
-  });
+  isScheduledToday = computed(() =>
+    isTaskOnToday(
+      this.task(),
+      this.globalTrackingIntervalService.todayDateStr(),
+      this._dateService.getStartOfNextDayDiffMs(),
+    ),
+  );
   hasTimeConflict = computed(() => {
     const task = this.task();
     return (
--- a/src/app/features/tasks/util/is-task-on-today.spec.ts
+++ b/src/app/features/tasks/util/is-task-on-today.spec.ts
@@ -1,0 +1,111 @@
+import { isTaskOnToday } from './is-task-on-today';
+import { Task } from '../task.model';
+
+const createTask = (overrides: Partial<Task> = {}): Task =>
+  ({
+    id: 'task1',
+    dueDay: undefined,
+    dueWithTime: undefined,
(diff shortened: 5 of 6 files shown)


johannesjo:

Thanks for the PR – this is a well-argued fix. The host parity is right: `scheduleForTodayById()` now applies the same two guards as `TaskComponent.scheduleForToday()`, the payload it dispatches is identical to `addToMyDay()`'s, the offset handling matches `isTodayWithOffset`/`selectOverdueTaskIds`, and the guards only ever suppress a dispatch, so the sync surface strictly shrinks. Nothing I found is blocking; two nits and two questions.

### Nits

- **`src/app/features/tasks/util/is-task-on-today.spec.ts:53-61`** – the test named *"wins over a dueDay that disagrees"* only exercises the direction where the disjunction happens to agree with a precedence check (`dueWithTime` **is** today). The other direction is the one that distinguishes the two shapes, and it is untested:

  ```
  { dueDay: '2026-03-15' /* today */, dueWithTime: <2026-03-16 09:00> }
  isTaskOnToday(...)                        -> true
  computeOrderedTaskIdsForToday(...)        -> false   work-context.selectors.ts:58-79
  planner.selectors.ts:66-74                -> false
  task.selectors.ts:617-623 (isInToday)     -> false
  ```

  Those three consult `dueDay` only when `dueWithTime` is unset, per `ARCHITECTURE-DECISIONS.md` Decision #1 (*"Reading: Check `dueWithTime` first; only check `dueDay` if `dueWithTime` is not set"*, *"Legacy Data: Old data with both fields works via priority pattern"*). So `isTaskOnToday` is a deliberate strict superset of canonical TODAY membership – which your JSDoc already discloses ("because this is a disjunction the order is not load-bearing"), and which matches the shipped sibling `isTaskOverdueByThreshold`. It only over-suppresses, and in the one shape where it does the dispatch was a no-op anyway, so there is nothing to fix in the code. Worth a test in the missing direction, though, plus maybe one line saying the superset is intentional – it is the kind of thing the next reader will trip over, and this util now has two production callers where `isTaskOverdue` has none.

- **`src/app/features/tasks/task/task-shortcuts.spec.ts:169-177`** – the file-level `DateService` spy provides only `isToday`/`getLogicalTodayDate`. `isScheduledToday()` now calls `getStartOfNextDayDiffMs()` and is template-bound (`task.component.html:166` and `:192`), and it is the first render-time call to that method in the component; only the Shift+T block's `beforeEach` stubs it. Nothing breaks today because this file never calls `fixture.detectChanges()`, but the first test that does will render those `@if` blocks against a mock without the method. Worth moving the stub into the shared `createSpyObj` with a `0` return.

### Questions

- The description cites `tsc --noEmit -p tsconfig.json` as clean, but the root `tsconfig.json` is solution-style (`"files": []` plus `references`), so without `--build` that invocation type-checks nothing. Did you also run `ng build`, `tsc -b`, or `ng lint` over the real programs? Asking because CI on this PR has run nothing but the greeting job – workflows here need maintainer approval for a first-time contributor – so your local runs are currently the only signal, and I would rather rely on the ones that actually compiled something.
- Did you run the specs under more than one timezone? `npm test` finishes with `test:tz:ci` (`America/Los_Angeles`). Your fixtures look TZ-safe to me – they are built and read back through local date components, and none of `2026-01-05`, `2026-03-15/16`, `2026-05-05` lands on a DST transition in Berlin, LA, Tokyo, Sydney or UTC – but you cite only `TZ='Europe/Berlin'`, so I would rather hear it than assume it.

---

sujeito-operator:
Thank you for reading it that closely. Taking the questions first, because the answer to the first one is that you caught a bad claim.

## 1. The type-check claim in my description was worthless, and you were right to ask

`tsc --noEmit -p tsconfig.json` exits 0 here — and it exits 0 because the root config is solution-style (`"files": []` plus `references`), so without `--build` it compiles nothing at all. I cited a green exit code from a command that checked zero files. That is on me; it should never have been in the description.

Here is what I have now actually run, on the PR head, including the part that is not clean:

| command | result |
|---|---|
| `npx tsc --noEmit -p tsconfig.json` | rc=0 — **and type-checks nothing**, as you said |
| `npx tsc --noEmit -p src/tsconfig.app.json` | **0 errors** |
| `npx tsc --noEmit -p src/tsconfig.spec.json` | **0 errors** |
| `npx tsc --noEmit -p src/tsconfig.worker.json` | **6 errors** |
| `npx tsc -b` | **rc=2** — the six above, nothing else |
| `npx ng lint` | **0 errors**, 90 warnings |
| `npx ng build --configuration development --aot` | **rc=0, 0 errors** — 1 warning, `NG8113` unused `MatIconButton` in `sync-conflicts-page.component.ts`, which this PR does not touch |

`--aot` is on that line deliberately, and it is the same trap you caught in the paragraph above: `angular.json`'s `development` configuration sets `"aot": false`, so the obvious `ng build --configuration development` compiles the app through the JIT path and **type-checks no template at all**. I nearly reported a green exit from that one as an AOT check. Forcing `--aot` back on is what makes the row mean anything.

**The six `tsc -b` errors are not mine and I would rather you checked that than believed it.** All six are in `src/app/core/log.ts` — `Cannot find name 'HTMLCollection' / 'NodeList' / 'Element'` — reached through `src/tsconfig.worker.json`, whose `lib` is `["es2018", "webworker"]` with `"types": []` and no DOM. I ran the same project in a clean worktree of `upstream/master` at `e8aaa72` and got **the identical six**. This PR touches none of those files, and `tsc -b` therefore exits 2 on master as well as here.

**On `ng lint`: 90 warnings, and exactly one of them lands on a file in this PR** — `max-lines` on `task.service.ts`, *File has too many lines (1541). Maximum allowed is 1200*. That is pre-existing (the same file is 1529 lines on master) but I am not going to file it under "lint clean": this PR adds to a file that is already 300+ lines over the limit. Happy to pull `scheduleForTodayById`'s guard out into the util file alongside `isTaskOnToday` if you would rather the number went down instead of up.

## 2. Timezones — yes, and wider than CI runs

The three spec files this PR touches — `task.service.spec.ts`, `task-shortcuts.spec.ts` and `is-task-on-today.spec.ts` — run in all seven of:

`Europe/Berlin` · `America/Los_Angeles` · `Asia/Tokyo` · `Australia/Sydney` · `UTC` · `Pacific/Kiritimati` (UTC+14) · `Pacific/Niue` (UTC-11)

**`TOTAL: 124 SUCCESS` in every one of them** — 124 of 124 executed in each of seven separate runs, no failures and no skips anywhere.

The last two are there because they are the extremes where a "build a local date, read it back through local components" fixture and a UTC-anchored one diverge by a whole calendar day, and every fixture in this PR goes through `new Date(y, m, d, h)` and `getDbDateStr`, which are both local. Kiritimati and Niue are 25 hours apart, so if any fixture here had been accidentally UTC-anchored, those two are where it would show.

**The full suite is a different story, and it is the weak part of this answer.** It was not run to completion here, and I would rather flag that than let it pass as a tick. The workspace half of `npm test` is green — `sync-core` **259 passed**, `shared-schema` **90 passed**, `sync-providers` **425 passed**, release-notes **29 passed** — but `ng test` over the full app spec set never reaches the browser: the bundler dies with

```
FATAL ERROR: Ineffective mark-compacts near heap limit
Allocation failed - JavaScript heap out of memory
```

on this machine (3.9 GB RAM, node's default heap ceiling 2150 MB). That is an environment limit, not a red test, but it does mean **I cannot honestly claim a green full suite** — so the seven-timezone runs above, over the three spec files this PR actually touches, are the strongest signal I have, and the full-suite row is a hole in my evidence rather than a pass. If that matters to you before merging, approving the workflow run on this PR would give you a signal none of my local runs can.

Your read of the fixtures matches what I found: none of `2026-01-05`, `2026-03-15/16`, `2026-05-05` is near a DST transition in any of those zones, and the `localMs()` helper in `is-task-on-today.spec.ts` exists precisely so the day boundary is computed the same way the production code computes it.

## 3. Nit — the missing direction, now tested and documented

Pushed. `is-task-on-today.spec.ts` now carries your exact shape — `dueDay` today, `dueWithTime` tomorrow — asserting `true`, with the three canonical readers that answer `false` for it named in the test with their line numbers, and a paragraph on the predicate itself saying the superset is deliberate, why it is safe (both call sites only ever use it to *suppress* a dispatch, so a superset can over-suppress but never plan something it should not), and that it must not be reused anywhere the answer decides what to **show** — the selectors are for that.

## 4. Nit — the `DateService` stub, moved; but the failure mode is not the one you named

Moved into the file-level `createSpyObj` with a `0` return, as you asked.

The reason is different from the one in your review, though, and it seemed worth saying rather than quietly agreeing. `detectChanges()` on its own cannot hit this, because that `beforeEach` calls `.overrideComponent(TaskComponent, { set: { template: '' } })` — the template is blanked for the whole file, so the `@if` blocks at `task.component.html:166` and `:192` never render and `isScheduledToday`, being a `computed()`, is never read. What does bite is a test that reads `component.isScheduledToday()` directly, or a future change that drops the template override.

I ran it both ways rather than argue it:

```
stub present, both probes   ->  TOTAL: 37 SUCCESS

stub REMOVED, both probes   ->  TOTAL: 1 FAILED, 36 SUCCESS
  ✗ reading isScheduledToday() needs DateService.getStartOfNextDayDiffMs
      TypeError: this._dateService.getStartOfNextDayDiffMs is not a function
  ✔ detectChanges alone does not read the computed (template is overridden to empty)
```

The second line of that failing run is the point: with the stub gone, the probe that only calls `fixture.detectChanges()` **still passes**, and the one that reads `component.isScheduledToday()` is the single failure.

So: the stub belongs where you put it, and it is now load-bearing for the direct-read case rather than for the render case. Both probes were reverted; neither is in the diff.

---

Nothing above changes the behaviour of the fix itself. If you would rather I also split the guard out of `task.service.ts` to stop that file growing, say the word and I will push it in the same branch.

johannesjo:
**🔍 Reviewed** – no blocking issues found at `83e9bc1`.

Thanks for the thorough reply, and for the correction on the type-check claim – that is the right way to answer it. This pass only covers what changed since `bd85f12`; the earlier conclusions on the fix itself stand, both nits are addressed, and everything below is optional.

Your correction on the `detectChanges()` mechanism was right, and master has since supplied the proof: #9864 added a `detail panel toggle button (#9850)` block to `task-shortcuts.spec.ts` whose `setTask` does call `fixture.detectChanges()`, and it passes without the stub – exactly because `overrideComponent(..., { set: { template: '' } })` keeps the `@if` blocks from rendering. The stub still earns its place for the direct-read case you demonstrated.

One thing I have not answered here: you offered to split `scheduleForTodayById`'s guard out so `task.service.ts` stops growing. That is a maintainer call rather than mine, so I have passed it on – worth waiting for an answer before you push anything on it.

### Nits

- **`task-shortcuts.spec.ts:678-680`** – adding `getStartOfNextDayDiffMs` to the file-level `createSpyObj` made the pre-existing block-level `(dateService as any).getStartOfNextDayDiffMs = jasmine.createSpy(...)` inert. Naming it in the *third* argument installs an accessor, and jasmine's generated `set` is a spy that discards the assigned value, so the file-level `() => 0` getter keeps winning. I checked that against the repo's `jasmine-core` 6.3.0 rather than reasoning about it: after `obj.getStartOfNextDayDiffMs = () => 999`, calling it still returns `0`. Nothing changes today since both stubs return `0`, but that line now reads as a per-block control and is not one – a future test setting a non-zero offset there would silently get `0`. Either drop it, or set it with `Object.defineProperty(..., { value, configurable: true })`, which is what the `isToday` stub you removed was working around.

- **`is-task-on-today.spec.ts:282-284`** – the three canonical readers are pinned to `file:line`. All three are right against this branch, but one is already wrong against master: `isInToday` has moved to `task.selectors.ts:636-641`, and `:617-623` there is now unrelated tracked-task code. Since the branch is 178 commits behind, that citation is stale the moment this merges. The symbol names carry the point on their own, the way the JSDoc on the predicate already does it.

- **`is-task-on-today.ts:33`** – *"in that one divergent shape the dispatch was a no-op anyway"* is a shade stronger than the reducer supports. For `dueDay` today + `dueWithTime` tomorrow, `handlePlanTasksForToday` does leave `tasksNeedingUpdate` empty (`task-shared-scheduling.reducer.ts:209-225`) so no task field changes – but `newTasksForToday` still includes the id whenever it is not already in `TODAY_TAG.taskIds` (`:198-204`), so the early return at `:227-230` is skipped and the dispatch still writes a `TODAY_TAG` ordering entry and runs `removeTasksFromPlannerDays`. It returns `state` untouched only when the id is already in there. The conclusion is unaffected – nothing the user sees moves and suppressing it is still right – so "changed nothing on the task itself" would be the exact wording.

---

johannesjo:
**💬 Maintainer call** – keep the guard where it is. No change needed, land it as is.

Thanks for flagging the cap yourself rather than letting it slide through, and for the seven-timezone runs (including the ±25h extremes) plus the plain statement that the full suite OOM'd instead of a tick.

**On the offer to net the file to ~0: don't.** Of the +16 in `task.service.ts`, 8 lines are the explanatory comment; the rest is the `today` / `startOfNextDayDiffMs` locals and a three-line `if`/`return`. A `shouldSkipPlanForToday()` in the util would relocate the comment and save no logic — the guard still has to sit at the dispatch point, and `scheduleForTodayById` still needs its own `task &&` because it tolerates a missing entity and dispatches anyway (`parentTaskMap: undefined`), which the component path, reading the required `task()` input, never does. Folding `isDone` into `isTaskOnToday` is the other tempting route and it is worse: that predicate is template-bound through `isScheduledToday()` for the sun icon on the schedule button, where a done task due today must still read as scheduled. That would be a second util export invented to satisfy a line budget rather than the design.

**Where the cap draws its line**, for next time: net-0 is what I expect when the growth is logic that could live elsewhere — which this PR already did, deleting the inline copy of the predicate in `TaskComponent` so the duplicable logic has one home. A guard that must sit at its own dispatch point is not that, so the physical count going up here is fine.

Non-blocking, take it or leave it: the 8-line comment could be ~3 by pointing at `TaskComponent.scheduleForToday()`, which already spells out both rationales. Not required, and not worth a push on its own.

On CI: the current head has no checks at all — all seven workflows sit at `action_required` behind the first-time-contributor approval gate (only the first commit ever ran, and it was green). I am approving the runs, so expect checks to appear; that is the full-suite signal you said you could not get locally.

PR #9712, opened 2026-08-23
Update nl.json

roosmsg:
## Problem

Mistakes in the Dutch language.

## Solution

Small language fix.


johannesjo:
Thank you very much! <3

PR #9717, opened 2026-08-24
fix(archive): preserve _hideSubTasksMode when archiving tasks

### LINKED ISSUE #7835: 🚨 Collapsed Tasks with Subtasks do not keep collapsed on next day
### Version Used

18.7.0MS

### Current Behavior

🚨 Collapsed Tasks with Subtasks do not keep collapsed on next day. Next day (after finishing in the evening and adding them agian to the day) they uncollapse and spend a lot of screen space. Every day I have to go throught these tasks in the morning and collapse the tasks again.

### Expected Behavior

They Should keep their status: If they are collapsed, they should stay collapsed for all time. If they are uncollapsed, they should stay that way.

### Steps to Reproduce

- create task with subtasks and date of today
- collapse the task.
- finish the day
- next day: add all tasks from yesterday to the this new day
-

### Console Output

```shell

```

### Documentation

AlvaroBalbin:
Archiving a task always reset its `_hideSubTasksMode` to undefined, so a collapsed task with subtasks lost its collapsed state the moment it was archived (e.g. via Finish Day) even though the collapse state is otherwise designed to be persistent. Restoring the task the next day then showed all subtasks expanded again. I removed that reset in `mapTasksToArchiveFormat` so the collapse state survives archiving and restoring, and added a regression test covering it. Fixes #7835.

Ran this through the project's own CI on my fork before opening it, all green.

### DIFF
--- a/src/app/features/archive/archive.service.spec.ts
+++ b/src/app/features/archive/archive.service.spec.ts
@@ -3,7 +3,7 @@ import { Store } from '@ngrx/store';
 import { ArchiveService, ARCHIVE_ALL_YOUNG_TO_OLD_THRESHOLD } from './archive.service';
 import { flushYoungToOld } from './store/archive.actions';
 import { TimeTrackingActions } from '../time-tracking/store/time-tracking.actions';
-import { TaskWithSubTasks } from '../tasks/task.model';
+import { HideSubTasksMode, TaskWithSubTasks } from '../tasks/task.model';
 import { ArchiveDbAdapter } from '../../core/persistence/archive-db-adapter.service';
 import { of } from 'rxjs';
 import { ArchiveModel } from './archive.model';
@@ -81,6 +81,19 @@ describe('ArchiveService', () => {
       const saveCall = mockArchiveDbAdapter.saveArchiveYoung.calls.first();
       const savedData = saveCall.args[0];
       expect(savedData.task.ids).toContain('task-1');
+    });
+
+    it('should keep _hideSubTasksMode when archiving so a restored task stays collapsed', async () => {
+      const tasks = [
+        createMockTask('task-1', { _hideSubTasksMode: HideSubTasksMode.HideAll }),
+      ];
+
+      await service.moveTasksToArchiveAndFlushArchiveIfDue(tasks);
+
+      const savedData = mockArchiveDbAdapter.saveArchiveYoung.calls.first().args[0];
+      expect(savedData.task.entities['task-1']!._hideSubTasksMode).toBe(
+        HideSubTasksMode.HideAll,
+      );
     });
 
     it('should serialize concurrent task archive mutations without dropping either task', async () => {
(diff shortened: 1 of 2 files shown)


johannesjo:
**🔍 Open question, not a code problem** – the diff at `b7a0c19` is correct, tested and green. What is unresolved is whether we want the behaviour change at all. Leaning toward closing, but I would rather discuss it than close it quietly.

Thanks for the PR, and sorry it took a while to get a proper answer. Dropping the `_hideSubTasksMode` reset out of `mapTasksToArchiveFormat` is a clean one-line change with a focused regression test, so the hesitation below is about direction, not craft.

### Why we are leaning toward closing

- **The motivating issue was already fixed by something else.** #7835 was closed as completed after a user confirmed it fixed on released `v18.19.0`. The actual fix looks like #8788 / `3dc9dc8db2` ("persist collapsed subtask state across restart"), which is in `v18.14.0` but not in `v18.13.1` — exactly the window between "still broken on 18.13.1" and "fixed". This PR's head is in no release, so it cannot be what that confirmation was testing.
- **The reported repro probably never touches the archive.** #7835 describes finishing the day and pulling yesterday's tasks into today. Undone tasks are never archived — both `moveToArchive` call sites (`work-view.component.ts`, `daily-summary.component.ts`) pass done tasks only — so the persistence fix explains that report on its own.
- **Nobody has asked for the remaining behaviour.** Searching open and closed issues for the archive → restore collapse symptom turns up only #7835 and #8781, both closed and both addressed by #8788.
- **The line was deliberate.** It came in with `809893b619` ("feat(pfapi): remove ui property on archiving"), alongside a coherent group of fields we strip on archive: `reminderId`, `dueWithTime`, `dueDay`. Merging this makes `_hideSubTasksMode` the one exception in that list. For the neighbouring "restore should bring X back" case (#9293) the shape we picked was to keep the archive snapshot stripped and opt in at restore time instead.
- **One concrete wart if merged:** `mapTasksToArchiveFormat` force-sets `isDone: true` on every flattened subtask, and `handleRestoreTask` un-dones only the parent. So a task archived while in `HideDone` would restore with *every* subtask hidden — a stale mode value that looks like `HideAll`.

### Why it is still a fair argument

- **The premise of the stripped line genuinely is stale.** When `809893b619` landed, this really was transient UI state. Since #8781 / #8788 it is persisted, op-log-captured task data — `task.actions.ts` now says outright that despite the `Ui` name it "must survive a restart". Archive is currently the only place still treating it as disposable: it survives a restart and syncs across devices, but not an archive round-trip. That is a real inconsistency.
- **The blast radius is nil and it is trivially revertible.** The field is already optional on `Task` and in the plugin API, both archive write paths share the same mapper so local and remote stay symmetric, and there is no schema or persisted-shape implication.
- **No trap for the user.** Archived tasks never render through the collapse-aware component, and a restored collapsed task still shows the "show N hidden sub-tasks" button, so the worst case is one extra click.

### The question that would settle it

#7835's original reporter never came back — the "fixed" confirmation came from a different user. If his flow was: collapse a parent with subtasks → mark it done → Finish Day → restore it from History, then that path is **still broken today** and this PR is its fix.

@AlvaroBalbin — is that the flow you hit yourself, or did you find this by reading the code? If you have a case where archive → restore specifically loses a collapse state you wanted back, that changes the answer and this goes in. Otherwise we are inclined to close it as superseded by #8788.

---

PR #9721, opened 2026-08-24
fix(tools): report css custom property offenders with posix paths

KamilDev:
## Problem

`tools/check-css-vars.js` builds each offender location with `path.relative`, so on Windows it reports

```
  --standard-border-radius   (component-scoped)
    src\styles\components\formly-rows.scss:1
```

while its own test asserts `/styles\/components\/formly-rows\.scss:1/`:

```
node --test tools/check-css-vars.test.js
✖ flags a global stylesheet reaching for a component-scoped declaration
  operator: 'match'
  expected: /styles\/components\/formly-rows\.scss:1/
```

`npm run lint` runs the tool tests and the pre-commit hook runs `npm run lint`, so on Windows every commit is blocked by this, whatever the change touches. I hit it committing an unrelated test fix (#9720).

## Fix

Normalize the reported path to forward slashes. Beyond making the test pass, offender output is meant to be pasted into an issue or compared between machines, so one separator on every platform is the right form on its own.

`node --test tools/check-css-vars.test.js` → 19/19 pass (was 18/19 on Windows). No behavior change on macOS/Linux, where `path.sep` is already `/`.

### DIFF
--- a/tools/check-css-vars.js
+++ b/tools/check-css-vars.js
@@ -43,6 +43,8 @@ const ROOT_STYLESHEET = path.join(SRC_DIR, 'styles.scss');
 // Carries a real document-level <style> block.
 const INDEX_HTML = path.join(SRC_DIR, 'index.html');
 
+const toPosix = (p) => p.split(path.sep).join('/');
+
 /**
  * Custom properties that are set at runtime rather than declared in CSS.
  * Add here ONLY with a verified source — never to silence a real typo.
@@ -229,7 +231,9 @@ for (const file of sourceFiles) {
         const componentScoped = anyDefs.has(name);
         anyComponentScoped ||= componentScoped;
         offenders.push({
-          location: `${path.relative(REPO_ROOT, file)}:${idx + 1}`,
+          // Reported paths stay posix-style so the output (and the specs that
+          // match it) reads the same on Windows as it does on macOS and Linux.
+          location: `${toPosix(path.relative(REPO_ROOT, file))}:${idx + 1}`,
           name: componentScoped ? `${name}   (component-scoped)` : name,
         });
       }


johannesjo:

**Normalizes the offender location in `tools/check-css-vars.js` to forward slashes so the report — and the specs that match it — reads the same on Windows as on macOS/Linux. Correct, minimal, and it does unblock the pre-commit `npm run lint` chain on Windows. No blocking issues.**

A few things I checked rather than assumed:

- Swapping the new `toPosix` call for a win32-separator emulation makes the unmodified spec go 16/19; with the file as it stands here it is 19/19.
- Splitting on `path.sep` rather than a blanket `/\\/g` replace is the right call — a literal backslash in a POSIX filename is left alone, and win32 `path.relative` always returns backslash-separated output (a cross-drive `D:\x` → `D:/x` is still fine).
- Normalizing only at the reporting boundary, and leaving `isGlobalStyleFile` / `SASS_LIBRARY_DIRS` on `path.sep`, keeps the internal comparisons native — which is the correct split.
- The other specs in the same `npm run lint` chain (`theme-assets`, `verify-linux-wm-class`, `verify-mac-icon`) don't have the same relative-path-in-assertion pattern, so this looks like the only such Windows blocker in the pre-commit path.

### Nits

- **PR description** — the pre-fix Windows count should be **16/19**, not 18/19. The spec carries four posix-separator location assertions, not one: `tools/check-css-vars.test.js:64` (`/app\/a\.component\.scss:1/`), `:86`–`:87` (`.ts:1` / `.html:1`) and `:108` (`/styles\/components\/formly-rows\.scss:1/`) — so three tests fail on win32 before this change. All four have been in the spec since it was introduced, and your one-liner repairs all of them; only the number in the description is off.

---

johannesjo:
Thank you very much! <3

PR #9723, opened 2026-08-24
feat(packaging): add arm64 linux builds

### LINKED ISSUE #5470: 💡 Thank You & Feature Request: Add Linux ARM Architecture deb Package
## Thank You & Feature Request: Add Linux ARM Architecture deb Package

First of all, I would like to express my sincere gratitude to the author for developing and maintaining Super Productivity—such an excellent tool!

Its cross-platform capability allows me to switch workflows seamlessly between different devices. Whether on desktop or mobile, I can enjoy a consistent user experience, which has greatly improved my task management and time tracking efficiency. The offline-first design, privacy protection, and rich integration features (such as GitHub/Jira synchronization) make it an indispensable productivity tool in my daily work. Thank you again for your remarkable work!

I would also like to gently propose a small request: currently, the software does not have an officially supported deb package for Linux ARM devices (e.g., Raspberry Pi, ARM-based laptops). Although it can be installed via Snap, deb packages are more user-friendly for ARM devices in terms of system integration, update convenience (via APT), and resource usage.

I hope the author can consider packaging a deb package for Linux ARM architectures (arm64) along with future releases. If testing is needed, I am happy to verify it on my own ARM device and provide feedback.
Thank you again for bringing such a fantastic tool to the open-source community. I look forward to it benefiting more users on the Linux ARM platform!

Best regards,
Shawn
Linux ARM Device User & Open-Source Enthusiast

### Possible Solution

currently, the software does not have an officially supported deb package for Linux ARM devices. I hope the author can consider packaging a deb package for Linux ARM architectures (arm64) along with future releases. If testing is needed, I am happy to verify it on my own ARM device and provide feedback.

_No response_

### Additional context

_No response_

ReFil:
## Problem

Currently ARM64 linux builds are not created/published, whilst solutions like FEX exist they need a lot of resources and it'd be nice to run it natively. I had to compile the appimage manually to get it on my laptop

Will close #5470 

## Solution

Adds ARM64 to the electron builder packaging configuration so they should be automatically built going forwards.

Appimage has been tested on a macbook air m1 running the latest fedora asahi remix. Currently wayland idle mode has to be ignored, this is to be fixed

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/electron-builder.yaml
+++ b/electron-builder.yaml
@@ -96,10 +96,21 @@ linux:
       # `-bin` value now matches nothing on either session type.
       StartupWMClass: superproductivity
   target:
-    - AppImage
-    - deb
-    - snap
-    - rpm
+    - target: AppImage
+      arch:
+        - x64
+        - arm64
+    - target: deb
+      arch:
+        - x64
+        - arm64
+    - target: rpm
+      arch:
+        - x64
+        - arm64
+    - target: snap
+      arch:
+        - x64
   publish:
     - github
 


johannesjo:

Thanks for the PR, and for picking up #5470! The packaging side holds up: electron-builder cross-builds all three arm64 targets from the x64 runner (AppImage pulls `runtime-arm64`, fpm gets `--architecture arm64`/`aarch64`), the existing x64 asset names are untouched because `artifactName` already carries `${arch}`, and no target list is parsed by `tools/verify-linux-wm-class.test.js` or `tools/verify-windows-artifact-contract.test.js`, so nothing breaks. Pinning `snap` to x64 is also the right call and worth keeping – `.github/workflows/build-publish-to-snap-on-release.yml:36-41` fails the release if more than one `.snap` asset is present.

Two things before this can ship.

### Should fix before merge

- **`electron-builder.yaml:99-110`** – the arm64 packages will contain the **x86-64** Wayland idle helper. `tools/build-wayland-idle-helper.js:69` runs `cargo build --release` with no `--target`, so it always emits one host-arch binary at `electron/bin/wayland-idle-helper` (the release Linux job is `ubuntu-latest`, i.e. x86-64). `tools/afterPack.js:97` then copies that single file into *every* Linux `appOutDir` with no reference to `context.arch`, and `installWaylandIdleHelper()` throws when the source is missing, so the arm64 pack can't just end up without it.

  At runtime that is a silent no-op rather than a crash: `_resolveWaylandIdleHelperPath()` (`electron/idle-time-handler.ts:541`) only does an `existsSync` check, so the wrong-arch file passes; the `spawn` at `:406` hits `ENOEXEC`, glibc's `execvp` falls back to `/bin/sh`, and the child just exits non-zero – no `error` event, so it settles `false` and the candidate chain moves on. On GNOME Wayland that's invisible (`gnomeDBus` is tried first). On non-GNOME Wayland arm64 there is no `gnomeDBus` candidate (`_buildWaylandCandidates()`, `:277`), `xprintidle` is X11-only and `loginctl` usually returns nothing, so idle detection ends up at `none` – and the promotion retry re-spawns the unrunnable binary every 60s.

  This is the "wayland idle mode has to be ignored" note from your description; flagging it because it is the one thing keeping the arm64 artifacts from being equivalent to the x64 ones, and because an arch-labelled `.deb`/`.rpm` carrying a foreign-arch ELF is worth avoiding on its own. Two ways out:
  - Cross-compile per target: `rustup target add aarch64-unknown-linux-gnu` plus `gcc-aarch64-linux-gnu` in the Linux job, `cargo build --target …`, emit `wayland-idle-helper-<arch>`, and pick by `context.arch` in `afterPack`.
  - Or the smaller one, matching what you already intended: skip `installWaylandIdleHelper()` when `context.arch !== Arch.x64`, so the resolver returns `null` and the fallback runs cleanly with no per-minute respawn. Note this is only partial – `files: electron/**/*` (`electron-builder.yaml:6`) has no exclusion for `electron/bin`, so the x86-64 helper is bundled in `app.asar` too (dead weight there, since the packaged resolver only looks next to `process.execPath`).

- **`docs/wiki/2.01-Downloads and Install.md:13`** – the download table's Linux column still lists only x64 badges, so the users this PR is for won't find the new builds on the documented download page. Please add arm64 rows for the three new assets: `superProductivity-arm64.AppImage`, `superProductivity-arm64.deb` and `superProductivity-aarch64.rpm` (rpm is the only one that maps to `aarch64`). Existing x64 links keep working, so this is additions only.

### Questions

- Which targets did you actually build? The description mentions the AppImage on Asahi. Nothing in CI exercises the fpm targets – `.github/workflows/electron-smoke.yml:93` packages `--linux dir` only – so the first real run of `--linux` with this config would be a tagged release. Have you confirmed that the arm64 `.deb` and `.rpm` both build and install?

---

Issue #5594, opened 2025-11-29
🚨 first time to creat repeat Task, the planned time will always be Today


chantsing:
### Version Used

v16.4.1

### Expected Behavior

🚨 first time to creat repeat Task, the planned time will always be Today
the planned time should be the day which is setted on  REPEAT options,  and after i change the day and time, The planned time will change with the setting in REPEAT options

### Current Behavior

irst time to creat repeat Task, the planned day and time will always be Today

### Steps to Reproduce

https://github.com/user-attachments/assets/3036213b-6e3e-4fce-86c3-a6d8e5bd5e6d

### Can you reproduce this reliably?

Yes, I can reproduce it reliably.

### Console Output

```shell

```

johannesjo:
I believe this will be fixed with the next release.

chantsing:
thank  you for your brilliant  working, but i have tested this fix, but i am afraid it is still not working

johannesjo:
How did you test @chantsing ?

chantsing:
> How did you test [@chantsing](https://github.com/chantsing) ?

Just Testing like the video i posted.

johannesjo:
I mean did you test with the latest release (v16.5.5) or did you check out the master branch?

chantsing:
> I mean did you test with the latest release (v16.5.5) or did you check out the master branch?

Yes, i test it on 16.5.5, including macos and Android

johannesjo:
Thank you! I'll investigate!

chantsing:
I have tested it on the newest version 16.6.0, but i am afraid it is still not working.

johannesjo:
I think I missunderstood the issue. Am I right to assume that you expect the task you are editing the repeat config to be re-scheduled for the day you specified as start day?

chantsing:
<img width="1840" height="1135" alt="Image" src="https://github.com/user-attachments/assets/1fd24cf0-e0f2-44e6-bace-fd2de2645dd6" />

As is shown on this figure, when setting a repeat task which every week on Monday, Wednesday, Friday,and  "start date“ :2025/12/14, But data of "Planned at " should be 2025/12/14? but it actual is 2025/12/13.
which means the first data of "Planned at " is always today. （today is Saturday on my region.）
<img width="1840" height="1135" alt="Image" src="https://github.com/user-attachments/assets/6ade887a-d8d8-47eb-a8c9-a7ca4225ab22" />
[...]

Issue #7619, opened 2026-05-15
Exported log leaks user content (task titles & action payloads) — actionLoggerReducer logs full payloads at VERBOSE in prod


johannesjo:
### Summary

The user-exportable log ("Settings → Provide Log file" / `exportLogHistory()`) contains **user content** — task titles, and potentially notes, project/tag names, and any other NgRx action payload. This violates the project's own logging rule (`Log.log({ id: task.id })`, never `Log.log(task)`/`Log.log(title)` — "log history is exportable, never log user content") and the explicit warning at the top of `src/app/core/log.ts`.

Surfaced in #7616 ([comment](https://github.com/super-productivity/super-productivity/issues/7616#issuecomment-4459483471)): the reporter exported their log to help debug a sync issue and found *"it literally has all my task names 😅 is there a way to disable that?"* — after being told in-thread that "the log never contains your tasks or credentials — it's safe to attach." That assurance is currently false.

### Root cause (verified in code)

1. **`src/app/root-store/meta/action-logger.reducer.ts:9-13`** — logs every non-`@ngrx` action's full payload (or the entire action object) at VERBOSE. The production gate is commented out, so it runs unconditionally:

   ```ts
   // if (environment.production)
   Log.verbose(
     '[a]' + action.type,
     (action as Action & { payload?: unknown })?.payload || action,
   );
   ```

2. **`src/app/core/log.ts:22`** — production builds run at VERBOSE, so these entries are recorded:

   ```ts
   const LOG_LEVEL = environment.production ? LogLevel.VERBOSE : LogLevel.VERBOSE;
   ```
   (both ternary branches are `VERBOSE` — likely an unintended leftover)

3. **`src/app/core/log.ts:196-199`** — `Log.verbose` → `recordLog('VERBOSE', …)` writes into the in-memory history that `exportLogHistory()` (`log.ts:233`) emits to the downloadable log file. Long values are truncated to 400 chars (`MAX_DATA_LENGTH`) but `title` appears early enough in a serialized task to survive truncation.

### Steps to reproduce

1. Use any release build (≥ the versions where `actionLoggerReducer` lost its production gate).
2. Create/track a task with an identifiable title.
3. Settings → bottom → "Provide Log file" / export log.
4. Open the exported file → entries like `[a][TimeTracking] Add time spent` and other `[a][...]` actions contain the full task object including `"title"`.

### Impact

[...]

PR #9725, opened 2026-08-24
feat(plugin): add task repeat config methods to plugin API

augustoesquivelmasciotta:
## Summary

Adds three methods to the plugin API for managing recurring tasks programmatically:

- `addTaskRepeatCfg(taskId, cfgData?)` — make a task repeatable; returns the new config's id
- `updateTaskRepeatCfg(cfgId, changes)`
- `deleteTaskRepeatCfg(cfgId)`

## Motivation

Repeat configs are already *readable* by plugins (`getAppState().taskRepeatCfgs`), but there is no way to create or maintain one from a plugin:

- there is no PluginAPI method for them, and
- the `TaskRepeatCfg` actions are (sensibly) not part of `ALLOWED_PLUGIN_ACTIONS` — and dispatching them raw would bypass the factory-attached op-log persistence meta anyway.

This blocks automation/integration plugins from setting up recurring tasks: everything else about a task can be automated through the API, but recurrence still requires opening the Repeat dialog by hand for every series. (My concrete case: an MCP bridge plugin that lets an AI assistant manage tasks — it can create scheduled tasks with reminders, but cannot make them recurring.)

## Design

- **Service-backed, not dispatch-based.** The bridge methods delegate to `TaskRepeatCfgService` (`addTaskRepeatCfgToTask` / `updateTaskRepeatCfg` / `deleteTaskRepeatCfg`) — the same single persistent action per call that the Repeat dialog dispatches. Persistence meta, first-occurrence calculation, `lastTaskCreationDay` and task↔config linking stay with the existing service/effects (#5594). One replay-atomic transition = one operation, per the contributor sync model; no new effects, reducers or actions.
- **Dialog parity for defaults.** Unset fields fall back to `DEFAULT_TASK_REPEAT_CFG`; `title` defaults to the task's current title, `startDate` is anchored to today, and `skipOverdue` uses the schedule-aware default from `get-default-skip-overdue.ts`.
- **Guards.** Unknown task/config → error; unknown tags → error (reuses `_validateTaskReferences`); sub tasks are rejected (matches the UI, which offers the dialog for main tasks only); a task that already has a `repeatCfgId` is rejected instead of silently re-linking, which would orphan the previous config. `projectId` always comes from the task, never from the caller, so a config can't be split from its task.
- **Input validation** with `typia.assert`, same pattern as `addTask` / `updateTask`.

## Changes

- `packages/plugin-api/src/types.ts` — new `PluginCreateTaskRepeatCfgData` type + the three `PluginAPI` methods. `PluginTaskRepeatCfg` gains optional `notes` / `skipOverdue` / `waitForCompletion` / `repeatFromCompletionDate` (read parity: these already flow through `getAppState()` at runtime, the type just didn't declare them).
- `src/app/plugins/plugin-api.ts` — thin logging wrappers.
- `src/app/plugins/plugin-bridge.service.ts` — validation + delegation.
- `src/app/plugins/plugin-bridge-task-repeat-cfg.service.spec.ts` — 11 specs: defaults derivation (title/startDate/skipOverdue), explicit fields incl. `remindAt`, and every guard (unknown task, sub task, existing config, unknown tags, unknown config on update/delete).
- `packages/plugin-api/README.md` — method list.
- Existing plugin-bridge specs (`add-task` / `counter` / `work-context` / core) get a `TaskRepeatCfgService` mock provider, since the bridge now injects it (the real service's field initializers don't survive a spied `Store`).

## Testing

- New spec: 11/11 green.
- All plugin suites (`plugin-bridge*` + `plugin-api`): 106/106 green (ChromeHeadless, `TZ=Europe/Berlin`).
- `packages/plugin-api` builds clean; eslint clean on touched files (only the pre-existing `max-lines` warning on the bridge).

## Notes

- No i18n keys added: reused `T.PLUGINS.TASK_NOT_FOUND` / `TAGS_DO_NOT_EXIST`; the new guard messages are plain strings, matching existing bridge messages like the sub-task move guard.
- `quickSetting` is documented as the dialog's grouping label only (as in the internal model); callers set the explicit recurrence fields.

### DIFF
--- a/packages/plugin-api/README.md
+++ b/packages/plugin-api/README.md
@@ -133,6 +133,9 @@ Add these to your manifest.json based on what your plugin needs:
 - `getSelectedTask` - Read the task selected in the task detail panel
 - `getFocusedTask` - Read the currently focused task row, if any
 - `addTask` - Create tasks
+- `addTaskRepeatCfg` - Make a task repeatable (create a repeat config for it)
+- `updateTaskRepeatCfg` - Update a task repeat config
+- `deleteTaskRepeatCfg` - Delete a task repeat config
 - `getAllProjects` - Read projects
 - `addProject` - Create projects
 - `getAllTags` - Read tags
--- a/packages/plugin-api/src/types.ts
+++ b/packages/plugin-api/src/types.ts
@@ -495,6 +495,10 @@ export interface PluginTaskRepeatCfg {
   lastTaskCreationDay?: string;
   quickSetting?: string;
   remindAt?: string;
+  notes?: string;
+  repeatFromCompletionDate?: boolean;
+  waitForCompletion?: boolean;
+  skipOverdue?: boolean;
 }
 
 export interface PluginSimpleCounterFull {
@@ -660,6 +664,26 @@ export interface PluginAPI {
 
   deleteTask(taskId: string): Promise<void>;
 
+  /**
+   * Make a task repeatable by creating a repeat config for it, mirroring the
+   * "Repeat" dialog. Unset fields fall back to the dialog's defaults. Rejected
+   * for sub tasks and for tasks that already have a repeat config.
+   * @returns the id of the created repeat config
+   */
+  addTaskRepeatCfg(
+    taskId: string,
+    cfgData?: PluginCreateTaskRepeatCfgData,
+  ): Promise<string>;
+
+  /** Update an existing task repeat config */
+  updateTaskRepeatCfg(
+    cfgId: string,
+    changes: PluginCreateTaskRepeatCfgData,
+  ): Promise<void>;
+
+  /** Delete a task repeat config; its task instances are cleaned up like a dialog delete */
+  deleteTaskRepeatCfg(cfgId: string): Promise<void>;
+
   batchUpdateForProject(request: BatchUpdateRequest): Promise<BatchUpdateResult>;
 
   // projects
@@ -805,6 +829,63 @@ export interface PluginCreateTaskData {
   isDone?: boolean;
   /** Due date as ISO date string (YYYY-MM-DD) */
   dueDay?: string | null;
+}
+
+/**
+ * Data for creating or updating a task repeat config via the plugin API.
+ * All fields are optional — on create, unset fields fall back to the same
+ * defaults the "Repeat" dialog uses.
+ *
+ * The recurrence itself is governed by `repeatCycle`, `repeatEvery` and the
+ * per-weekday flags; `quickSetting` is only the grouping label shown in the
+ * dialog form and has no direct effect (matches the internal model).
+ */
+export interface PluginCreateTaskRepeatCfgData {
+  /** Title used for created task instances; defaults to the task's current title */
+  title?: string | null;
+  tagIds?: string[];
+  isPaused?: boolean;
+  repeatCycle?: 'DAILY' | 'WEEKLY' | 'MONTHLY' | 'YEARLY';
+  /** Repeat every N cycles (default 1) */
+  repeatEvery?: number;
+  /** ISO date string (YYYY-MM-DD) the recurrence is anchored to; defaults to today */
+  startDate?: string;
+  /** HH:mm — when set, created instances are scheduled at this time of day */
+  startTime?: string;
+  /** Default time estimate for created instances in milliseconds */
+  defaultEstimate?: number;
+  monday?: boolean;
+  tuesday?: boolean;
+  wednesday?: boolean;
+  thursday?: boolean;
+  friday?: boolean;
+  saturday?: boolean;
+  sunday?: boolean;
+  /** Dialog grouping label only — set the explicit recurrence fields as well */
+  quickSetting?:
+    | 'DAILY'
+    | 'WEEKLY_CURRENT_WEEKDAY'
+    | 'MONTHLY_CURRENT_DATE'
+    | 'MONTHLY_FIRST_DAY'
+    | 'MONTHLY_LAST_DAY'
+    | 'MONTHLY_NTH_WEEKDAY'
+    | 'MONDAY_TO_FRIDAY'
+    | 'YEARLY_CURRENT_DATE'
+    | 'CUSTOM';
+  /** Reminder for scheduled instances; only has an effect together with startTime */
+  remindAt?: 'DoNotRemind' | 'AtStart' | 'm5' | 'm10' | 'm15' | 'm30' | 'h1';
+  /** Notes template for created task instances */
+  notes?: string;
+  /** Base the next occurrence on the completion date instead of the schedule */
+  repeatFromCompletionDate?: boolean;
+  /** Only create the next instance after the current one is completed */
+  waitForCompletion?: boolean;
+  /**
+   * Silently skip missed/overdue instances instead of creating them.
+   * Defaults to the dialog's schedule-aware default (ON only for a plain
+   * everyday schedule).
+   */
+  skipOverdue?: boolean;
 }
 
 export interface PluginShortcutCfg {
--- a/src/app/plugins/plugin-api.ts
+++ b/src/app/plugins/plugin-api.ts
@@ -12,6 +12,7 @@ import {
   PluginAppState,
   PluginBaseCfg,
   PluginCreateTaskData,
+  PluginCreateTaskRepeatCfgData,
   PluginHeaderBtnCfg,
   PluginHookHandler,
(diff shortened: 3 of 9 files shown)


johannesjo:

**Adds `addTaskRepeatCfg` / `updateTaskRepeatCfg` / `deleteTaskRepeatCfg` to the plugin API, delegating to `TaskRepeatCfgService` so each call stays one persistent action. Delegating instead of dispatching is the right call and the guard set is thoughtful, but the defaults don't actually reach dialog parity, and one gap lets a plugin-created schedule be silently rewritten.**

### Blocking

- **`src/app/plugins/plugin-bridge.service.ts:1072-1089`** — `quickSetting` stays at `DEFAULT_TASK_REPEAT_CFG`'s `'DAILY'` whenever the caller expresses the schedule through `repeatCycle`/weekday flags — which is exactly what the new type doc tells them to do (*"quickSetting is only the grouping label shown in the dialog form and has no direct effect"*). It is not inert:
  - `getDefaultSkipOverdue(cfgBase)` keys off it and ignores `repeatCycle`, so `addTaskRepeatCfg('t1', { repeatCycle: 'MONTHLY', startDate: '2026-09-01' })` persists `skipOverdue: true`, and `task-repeat-cfg.service.ts:266-281` then advances `lastTaskCreationDay` past every missed monthly occurrence without creating the task — the case `get-default-skip-overdue.ts` explicitly calls out as unsafe ("pay rent on the 1st … never disappears").
  - The Repeat dialog re-derives the schedule *from* it. That config renders as the "Daily" preset, and on Save `dialog-edit-task-repeat-cfg.component.ts:452-465` runs `getQuickSettingUpdates('DAILY')` → `{ repeatCycle: 'DAILY', repeatEvery: 1 }`, converting the plugin's monthly recurrence into a daily one without the user changing anything. `data-repair`'s `_fixTaskRepeatCfgInvalidQuickSetting` doesn't repair this combination — it only handles a preset with no `startDate`.

  The zero-arg call has the mirror-image problem: `addTaskRepeatCfg('t1')` persists WEEKLY Mon–Fri with `skipOverdue: true`, while the dialog's default persists `repeatCycle: 'DAILY'` — so "dialog parity for defaults" doesn't hold, and a missed Friday instance disappears if the app isn't opened until Saturday.

  Fix: default `quickSetting` to `'CUSTOM'` when the caller supplies any explicit recurrence field and no `quickSetting`, then derive `skipOverdue` from that. Every other creation path already does this — `get-interval-repeat-updates.ts` returns `quickSetting: 'CUSTOM'` for each cycle, `add-task-bar.component.ts:634-637` always writes `quickSetting` together with `getQuickSettingUpdates()`, and `auto-fix-typia-errors.ts:249` calls `'CUSTOM'` "the safe default". The type doc then needs correcting too. Related: `PluginCreateTaskRepeatCfgData` accepts every preset name but exposes none of the fields those presets set (`monthlyWeekOfMonth` / `monthlyWeekday` / `monthlyLastDay`), so `quickSetting: 'MONTHLY_LAST_DAY'` yields a config labelled last-day whose engine still uses `startDate`'s day-of-month — another shape the UI can't produce. Expanding a caller-supplied preset through `getQuickSettingUpdates()` (as the add bar does) would fix both halves at once.

### Worth fixing

- **`src/app/plugins/plugin-bridge.service.ts:1073-1088`** — the Mon–Fri weekday flags from `DEFAULT_TASK_REPEAT_CFG` bleed into WEEKLY configs the caller didn't intend as Mon–Fri: `{ quickSetting: 'CUSTOM', repeatCycle: 'WEEKLY', repeatEvery: 2, startDate: '2026-09-07' }` (a Monday) means "every 2 weeks, Mon–Fri" — five instances per fortnight — not "every other Monday". `get-interval-repeat-updates.ts` documents this exact trap for the identical internal case and works around it with `buildWeeklyForDay()`. Either derive the flags from `startDate` when none are given, or document the Mon–Fri default on `PluginCreateTaskRepeatCfgData` (whose current wording, "fall back to the same defaults the Repeat dialog uses", is misleading here — the dialog never persists that combination without explicit checkboxes).

- **`src/app/plugins/util/plugin-iframe.util.ts:37-81`** — the three methods are declared on the public `PluginAPI` interface but are unreachable from an iframe plugin: they're missing from `ALLOWED_IFRAME_API_METHODS` and from the `window.PluginAPI` shim (`:401-470`). An `index.html` plugin that types against `@super-productivity/plugin-api` gets `PluginAPI.addTaskRepeatCfg is not a function` — not even the explicit rejection the shim gives registration methods via `unsupportedIframeRegistration()`. Commit `3db96fd8a3` (#8291, `getFocusedTask`) is the precedent for what a new API method touches. If the omission is deliberate, wiring an explicit rejection would at least make it legible.

- **`src/app/plugins/plugin-bridge.service.ts:1123-1128`** — `updateTaskRepeatCfg` spreads unvalidated caller input into the entity update. `typia.assert` (unlike `assertEquals`) does not reject properties the type doesn't declare, so anything extra reaches `adapter.updateOne`:
  - `updateTaskRepeatCfg('cfg-1', { projectId: 'other' })` splits the config from its task — the very thing the add path guards against by writing `projectId` from the task.
  - `{ id: 'something-else' }` re-keys the entity (`task-repeat-cfg.reducer.ts:66` → `adapter.updateOne`; @ngrx/entity's `takeNewKey` deletes the old key when `selectId(merged) !== update.id`), leaving `task.repeatCfgId` dangling and the config unreachable.

  Both then sync to every device. Building the payload from an explicit field list (in the spirit of `REJECTED_UPDATE_FIELDS` at `:99`) or switching these asserts to `typia.assertEquals` closes it; `lastTaskCreationDay` / `subTaskTemplates` / `deletedInstanceDates` leak through on both paths.

- **`src/app/plugins/plugin-bridge.service.ts:1072-1088`** — `startDate` and `repeatEvery` are only shape-checked, so malformed values are persisted and fed to the recurrence engine. `addTaskRepeatCfg('t1', { repeatCycle: 'DAILY', startDate: '01/09/2026' })` is accepted; `dateStrToUtcDate()` returns `Invalid Date`, and the `addTaskRepeatCfgToTask` effect writes `lastTaskCreationDay: 'NaN-NaN-NaN'` plus `taskService.update(task.id, { created: NaN, dueDay: 'NaN-NaN-NaN' })` — synced, and nothing rejects it. `repeatEvery: 0` yields a config that never produces an occurrence (`get-first-repeat-occurrence.util.ts:24-26`). `isValidDBDateStr()` (`src/app/util/get-db-date-str.ts`) plus an integer ≥ 1 check at the boundary covers it. (`startTime` is less urgent — the repeat processor already guards with `isValidSplitTime`, `task-repeat-cfg.service.ts:332` — but validating it here too would be consistent.)

- **`src/app/plugins/plugin-bridge.service.ts:1039-1148`** — this adds 127 lines to `plugin-bridge.service.ts`, already 2252 lines against the repo's 1200-line `max-lines` cap it is grandfathered under. `AGENTS.md` says that list "may only ever SHRINK" and those services are "debt to pay down when touched, not a precedent to extend". The PR already names its spec `plugin-bridge-task-repeat-cfg.service.spec.ts` — extracting a small `PluginBridgeTaskRepeatCfgService` collaborator would make that name a real unit and keep the bridge from growing.

### Nits

- **`src/app/plugins/plugin-bridge.service.ts:1072-1088`** — a past `startDate` is accepted, which back-dates the *existing* task: `addTaskRepeatCfg('t1', { repeatCycle: 'MONTHLY', startDate: '2020-01-01' })` (no `startTime`, so `clampPastTimedOccurrence` deliberately doesn't clamp) makes the effect run `taskService.update(task.id, { created: <2020 ts>, dueDay: '2020-01-01' })`, re-dating a task created today and moving it out of today's view. The dialog can't do this for a new config — its picker sets `minDate` to the reference date when `!isEdit()` (`dialog-edit-task-repeat-cfg.component.ts:193`).
- **`packages/plugin-api/README.md:136`** — `docs/plugin-development.md` (`:290-298`) and `docs/wiki/3.01-API.md` (`:65`) also enumerate the task methods and were both updated by the `getFocusedTask` precedent; worth adding the three there too.
- **`src/app/plugins/plugin-api.ts:249-271`** — no cases in `plugin-api.spec.ts` for the new wrappers. They log ids only (correct), but nothing pins that against a later change logging `cfgData` and putting a task title into the exportable log (#7619).

One process note: CI hasn't run on this PR yet — as a first-time contributor the workflows need a maintainer to approve them, so the green local runs in your description are currently the only signal.

---

PR #9727, opened 2026-08-24
feat(electron): add Cmd+W to hide the window on macOS

dan-hart:
## Problem

On macOS, `Cmd+W` does nothing: the app menu has no Window menu, so the standard "close window" chord isn't bound. Mac users reflexively press `Cmd+W` expecting the window to go away while the app keeps running in the Dock.

Simply binding `Cmd+W` to Electron's `close` role would be the wrong fix here: with minimize-to-tray disabled (the default), the window close path runs the full quit flow (finish-day prompt, sync, `app.quit()`), so `Cmd+W` would effectively become `Cmd+Q`. Role accelerators also act on whichever window has focus, so they could destroy the floating task widget or poke at the full-screen break blocker.

## Solution

- Add a macOS-only **Window ▸ Close Window** item (`CmdOrCtrl+W`) with a custom click handler that hides the main window explicitly (recording the maximized state first, same as the tray-hide path). The app keeps running in the Dock; clicking the Dock icon re-shows the window through the existing `activate` → `showOrFocus()` path, including re-maximizing.
- Deliberately **not** `role: 'close'` for the reasons above — the trade-off is commented in the code.
- Extract the menu template into `electron/menu.ts` as a pure function so it can be tested without loading `main-window`'s import graph.
- Add `electron/menu.test.cjs` (5 tests), including a regression guard asserting the Window menu never contains `close`/`minimize`/`zoom` roles.
- Document the shortcut in `docs/wiki/3.03-Keyboard-Shortcuts.md`.

Verification:

- `npm run checkFile electron/menu.ts` / `electron/main-window.ts` — clean
- `npm run test:electron` — 285 pass (280 existing + 5 new)
- `npx tsc -p electron/tsconfig.electron.json --noEmit` — clean
- Drove the built app with Playwright's Electron driver: triggering the menu item hides the window (not destroyed, app still running), repeated `Cmd+W` while hidden is a no-op, and a simulated Dock click (`activate`) re-shows it; verified across two full hide/restore cycles.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/docs/wiki/3.03-Keyboard-Shortcuts.md
+++ b/docs/wiki/3.03-Keyboard-Shortcuts.md
@@ -113,6 +113,7 @@ These shortcuts are **unconfigurable** and always available on group headers.
 - Global shortcuts (`globalShowHide`, `globalToggleTaskStart`, `globalAddNote`, `globalAddTask`) work system-wide when the app is running.
 - Window zoom controls (`zoomIn`, `zoomOut`, `zoomDefault`) are desktop-only.
 - `F11` toggles full-screen mode for the main window.
+- `Cmd+W` (macOS only) hides the main window; the app keeps running and clicking the Dock icon brings the window back. This shortcut is unconfigurable.
 
 **Linux (Wayland):**
 
--- a/electron/main-window.ts
+++ b/electron/main-window.ts
@@ -580,8 +580,13 @@ function createMenu(quitApp: () => void): void {
   const menuTpl = createMenuTemplate({
     // hide() keeps the app running in the dock; clicking the dock icon
     // re-shows via the 'activate' handler (showOrFocus)
-    onCloseWindow: () => {
-      if (mainWin && !mainWin.isDestroyed() && mainWin.isVisible()) {
+    onCloseWindow: (focusedWindow) => {
+      // only act when the main window itself is key; focusedWindow can be
+      // undefined during macOS menu tracking — treat that as "not ours"
+      if (!focusedWindow || focusedWindow !== mainWin) {
+        return;
+      }
+      if (!mainWin.isDestroyed() && mainWin.isVisible()) {
         setWasMaximizedBeforeHide(mainWin.isMaximized());
         mainWin.hide();
       }
--- a/electron/main-window.ts
+++ b/electron/main-window.ts
@@ -5,7 +5,6 @@ import {
   BrowserWindowConstructorOptions,
   ipcMain,
   Menu,
-  MenuItemConstructorOptions,
   nativeTheme,
   shell,
 } from 'electron';
@@ -27,6 +26,7 @@ import {
 } from './task-widget/task-widget';
 import { ensureIndicator } from './indicator';
 import { getIsMinimizeToTray, getIsQuiting, setIsQuiting } from './shared-state';
+import { createMenuTemplate } from './menu';
 import { loadSimpleStoreAll } from './simple-store';
 import { SimpleStoreKey } from './shared-with-frontend/simple-store.const';
 import { markGpuStartupSuccess } from './gpu-startup-guard';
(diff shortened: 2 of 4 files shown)


johannesjo:

**Adds a macOS-only Window ▸ Close Window item bound to `Cmd+W` that hides the main window instead of closing it, and extracts the menu template into a testable pure function — the extraction, the platform gating and the test harness all check out; one wrong-target case in the new accelerator is worth fixing first.**

I verified locally: `createMenu` is still `IS_MAC`-only (`electron/main-window.ts:374`), so Windows/Linux keep `setMenu(null)` and there's no `Ctrl+W`-hides-window regression where `hide()` would drop the taskbar entry. `electron/menu.test.cjs` runs green (5/5) with the same `ts-node/register/transpile-only` harness ~20 other `electron/*.test.cjs` files use, and `menu.ts` imports `MenuItemConstructorOptions` type-only so no electron runtime is needed. `npm run checkFile electron/menu.ts` is clean. No pre-existing `Cmd+W` binding anywhere in `electron/` or `src/app`, so nothing is stolen from the renderer.

### Worth fixing

- **`electron/menu.ts:49` / `electron/main-window.ts:583`** — `CmdOrCtrl+W` is installed via `Menu.setApplicationMenu`, so on macOS it's an **application-wide** accelerator: it fires whichever of the app's windows is key, but `onCloseWindow` unconditionally targets `mainWin`. That's the same "roles act on the focused window" hazard the code comment at `menu.ts:45-47` calls out — the custom handler removes the *destructive* half but still fires on the wrong target.

  Concrete case: the take-a-break full-screen blocker (`electron/full-screen-blocker.ts:29`) is a focusable `alwaysOnTop` window shown over a still-visible main window. Press `Cmd+W` during a break and the main window silently hides behind the overlay — nothing appears to happen, and after the break the window is gone with no visible cause. Same for a detached DevTools window, and for the task widget when it's pinned visible via always-show (`electron/task-widget/task-widget.ts`), where a user pressing `Cmd+W` on the focused widget gets the main window hidden instead.

  Fix: gate on the main window actually being key. Electron passes the focused window to the handler — `click: (_item, browserWindow) => { if (browserWindow !== mainWin) return; ... }` — which is more reliable than `BrowserWindow.getFocusedWindow()` (that can be `null` during macOS menu tracking; note `browserWindow` can be `undefined` too, so treat that as "not the main window"). The codebase already has this precedent: `toggleWindowVisibility` checks `if (!win.isFocused())` before hiding (`electron/various-shared.ts:98`).

### Questions

- macOS already hides the whole app on `Cmd+H` via the existing `role: 'hide'` item in the app menu (`electron/menu.ts:18`). What does `Cmd+W`-hides-window add over that for you in practice — is it purely the muscle-memory chord, or is there a behavioral difference you're after (e.g. leaving other SP windows up)? Asking because it affects whether "hide" is the right verb here at all versus, say, minimizing.
- I looked at whether `hide()` on a window that owns a native macOS full-screen Space (the `F11` path) needs special handling, and concluded it doesn't block this PR — the identical `setWasMaximizedBeforeHide(...)` + `hide()` sequence already ships on the global show/hide shortcut and the minimize-to-tray path, so any issue there is pre-existing, not introduced here. Did your Playwright run happen to cover the full-screen case, or is that untested territory?

---

dan-hart:
@johannesjo Thanks for the thorough review — the wrong-target case was real, and your suggested fix is what I went with. Fixed in 99a0689: the template now forwards the window argument from the click handler (`click: (_menuItem, focusedWindow) => onCloseWindow(focusedWindow)`), and the handler returns early unless that window is `mainWin`, treating `undefined` (menu tracking) as not-ours per your note. Added a test asserting the focused window is forwarded, and verified the gate in the running app: triggering the item with a break-blocker-style decoy window or `undefined` is a no-op, while the main window as key still hides and Dock-click restores.

On Cmd+H vs Cmd+W: it's partly the muscle-memory chord, but there is a behavioral difference I want. `role: 'hide'` is app-wide NSApp hide — it hides every window including the floating task widget and yields focus to another app. Cmd+W dismisses just the main window: the app stays active and the task widget stays up (main-window `hide` intentionally triggers `showTaskWidget()`), which matches how the tray-hide and `globalShowHide` paths already behave. That's also why hide rather than minimize: `activate` → `showOrFocus()` and `wasMaximizedBeforeHide` are all built around hidden-not-minimized, and `appMinimizeHandler` already rewrites minimize into hide when tray mode is on, so hide is the app's canonical "dismiss" verb.

On full-screen: honest answer — no, my Playwright runs launched a normal window, so the native full-screen Space case is untested territory here. Agreed it's shared with the existing `globalShowHide`/tray paths rather than introduced by this PR; happy to look at it separately if you'd like an issue for it.

johannesjo:
**🔍 Reviewed** – no blocking issues found at `99a0689`.

Thanks for the quick follow-up – the focused-window gate is exactly the right fix and the wrong-target case from the last round is closed. Nothing blocking left at this head.

What I checked at this head:

- The template forwards Electron's focused window (`electron/menu.ts:52`) and the handler returns unless it is `mainWin` (`electron/main-window.ts:583-588`), so `Cmd+W` while the break blocker or the task widget is key no longer hides the main window. Electron 43 types `click` as `(menuItem, window: BaseWindow | undefined, event)`, which is the signature `menu.ts` declares, and `BrowserWindow extends BaseWindow`, so the identity comparison holds and typechecks.
- The `!focusedWindow` arm also covers `mainWin === null` (`main-window.ts:674` nulls it in the `closed` handler), so the `mainWin.isDestroyed()` on the next line can never throw. `mainWin` is a module-level `let`, so the closure always reads the current window rather than a stale one.
- `createMenu` is still `IS_MAC`-only (`main-window.ts:374`), so Windows/Linux keep `setMenu(null)` and there is no `Ctrl+W`-drops-the-taskbar-entry regression.
- `hide()` triggers `showTaskWidget()`, which early-returns unless the widget is enabled, so `Cmd+W` cannot pop an unwanted floating widget; `setWasMaximizedBeforeHide()` before `hide()` matches the tray-close and minimize paths, and Dock click restores through `activate` -> `showOrFocus()` including re-maximize.
- Packaging needs no follow-up: `tsconfig.electron.json` includes `**/*.ts` and `electron-builder.yaml` ships `electron/**/*`, so the new module compiles and is packaged without an explicit file-list entry.
- No default renderer shortcut uses a modified `W`, so nothing is stolen from the app's own keymap.
- I ran `menu.test.cjs` at this head from a scratch copy: 6/6 pass. A second pass over the whole `electron/*.test.cjs` suite against these files was green as well (286/286), and prettier is clean on all three changed files.

### Nits

- **`electron/main-window.ts:583`** – the one piece of genuinely new logic, the "is the main window the key window" gate, is the part no test covers. `menu.test.cjs` asserts that the template forwards the focused window, but nothing pins what the handler decides with it, so a future refactor could drop or invert `focusedWindow !== mainWin` and all six tests would still pass. Exporting the predicate from `menu.ts` (e.g. `export const shouldHideMainWindow = (focusedWindow, mainWin) => !!focusedWindow && focusedWindow === mainWin`) and asserting on it would close that, in the same spirit as the template extraction.
- **PR description** – still says "5 tests" / "285 pass"; this head has 6 and 286.

---

dan-hart:
Thank you @johannesjo !!!!

Issue #9310, opened 2026-07-26
💡 Floating Window Overlay: compact always-on-top task list and timer


Ericyuanxiang:
# 💡 Feature Request: 悬浮窗模式 (Floating Window Overlay)

提交地址：https://github.com/johannesjo/super-productivity/issues/new

选择模板：**💡 Feature Request**

---

## Problem Statement

The current "Always on Top" option keeps the entire main window visible, which is too bulky for continuous use. When working with multiple browser tabs, code editors, or other apps open, there is no way to keep a **minimal, always-visible view** of:

- The currently tracked task
- The running timer
- A short list of top-priority tasks (3–5 items)

Users who want a constant visual reminder of what they should be working on have to keep switching back to the main window. This reduces the app's usefulness as a persistent productivity companion.

This has been discussed in [Discussion #3984](https://github.com/johannesjo/super-productivity/discussions/3984), where the requester described keeping the main window always-on-top as "too bulky" for workflows involving many open windows.

## Possible Solution

Add a **Floating Window / Compact Overlay Mode** that:

1. **Compact UI** — A small, resizable window showing only:
   - Current task name (with a visual indicator)
   - Running timer (large, easy to read)
   - Top 3–5 tasks (drag-reorderable or from a configurable "favorites" list)
   - Quick-action buttons: Start/Stop timer, Mark Done

2. **Always on Top** — The floating window stays above all other applications.

3. **Independent from Main Window** — The main Super Productivity window can be minimized or closed without affecting the floating view. Ideally the app continues running in the system tray when the main window is closed, keeping the floating window alive.

4. **Toggleable** — A shortcut or menu option to show/hide the floating window.

5. **Theme Support** — Respects the existing light/dark theme settings.

## Alternatives Considered

- **Window Always on Top** — Currently available, but shows the full interface which is too large and cluttered for a persistent overlay.
- **System Tray Timer** — The tray icon shows some info but has no interactive task list or quick controls.
- **Third-party overlay tools** (e.g., Always on Top utility apps) — These can only pin existing windows, not create a purpose-built compact view.
- **Global Shortcut to toggle main window** — Useful, but still requires the full UI to appear when checking the current task.
- **Pomodoro-style mini-timer popup** — Would address only the timer visibility, not the task-list aspect.

## Additional Context

- Related Discussion: [Floating Window with only a few Tasks + Timer #3984](https://github.com/johannesjo/super-productivity/discussions/3984)
[...]

Issue #2607, opened 2023-05-07
Option for always on top during a session with a better UI


BEST8OY:
## Problem Statement
I want to have the time in front of me

## :grey_question: Possible Solution
Add option for always on top during a session with a better UI for smallest Super-Productivity display
Windows 11 Clock vs Super-Productivity
![Windows 11 Clock vs Super-Productivity](https://user-images.githubusercontent.com/70865987/236696486-8cdbc022-b9f1-4294-9330-786b3c8fd8c3.png)

johannesjo:
I am open for adding this as an opt in feature!

beewry:
Something like this would be **_huge_** for me. The mini-timer is the only reason that I use toggl.

I need to see my current task name and the time spent on my screen at all times to pull me back when I get distracted. Having this as an opt in feature would be fantastic.

johannesjo:
Should be relatively straightforward to implement:
https://github.com/electron/electron/blob/main/docs/api/browser-window.md#winsetalwaysontopflag-level-relativelevel

Someone wants to give it a shot?

jcrben:
Another idea to add to the list: integrate it into the Windows taskbar. I noticed https://github.com/leandrosa81/taskbar-monitor which uses https://github.com/dsafa/CSDeskBand - of course that opens up a lot of additional complexity and is Windows-specific.

Issue #556, opened 2020-10-01
Always on top


johannesjo:
By Valentin:

Please implement an option for the timer to always be on top similar to YAPA-2 , see screenshot attached.

Also a way of choosing transparency would also be good, the idea is that it should always be on top of other things but still allow to see through the text of the app, I keep it in the bottom right corner and sometimes there are things that need to be clicked in that corner, so at least I can see there is some information behind.

I found a work around by scaling down the window and using Turbotop app to have it show over all other windows that I have open, but I have to go through several steps every time I reboot my computer. 

Bonus: Some options to scale the text would be good as I don’t have a high resolution screen on my laptop.

![8C3E3F9258BE4AD1958F489E84508DD2](https://user-images.githubusercontent.com/1456265/96619439-7d6da000-1306-11eb-85fe-501c72e0bc0d.png)

himanshu2408:
Hi @johannesjo I don't see any screenshot, can you please check?

johannesjo:
@himanshu2408 thanks for pointing that it out. I attached it.

PR #9729, opened 2026-08-25
feat(desktop): add "always on top" window setting

BigWebstas:
## Summary
- Adds an Electron-only Misc config toggle that keeps the app window above all other windows (e.g. KDE's "keep above others"), reusing the existing settings-sync IPC channel.
- Hides the toggle on Wayland: `BrowserWindow.setAlwaysOnTop()` silently no-ops there since xdg-shell has no client request for "stay above others" (compositor-only policy), so the setting is gated behind the existing `isWaylandEnv()` detector and only shown on macOS, Windows, and Linux X11.

## Test plan
- [x] `electron/app-control.test.cjs` covers the new setting
- [x] `npm run lint` / `npm test` pass (run via pre-push hook)
- [x] Manual check: toggle appears and works on macOS/Windows/Linux X11, hidden on Wayland


johannesjo:

**Small, well-shaped feature with a test and maintainer buy-in (#2607: "I am open for adding this as an opt in feature" + a pointer at `setAlwaysOnTop`) — but the unconditional apply and the Wayland gate interact badly, and the two problems compound on the same users.**

### Worth fixing

- **`electron/ipc-handlers/app-control.ts:53-54`** — `getWin().setAlwaysOnTop(!!cfg.misc.isAlwaysOnTop)` runs unconditionally on every settings transfer, so with the toggle **off** (the default) the app now actively *clears* always-on-top that the user set outside the app. Before this PR nothing ever called `setAlwaysOnTop` on the main window (only `full-screen-blocker.ts` and `task-widget.ts`, both separate windows). `updateSettings` is bound to both `TRANSFER_SETTINGS_TO_ELECTRON` and `UPDATE_SETTINGS`, so it fires on every launch (`startup.service.ts:220-227` after data load, plus `notifyElectronAboutCfgChangeInitially` on `loadAllData`) and again on **any** `updateGlobalConfigSection` — changing an unrelated setting re-issues it. A user who keeps SP above other windows via KWin's "Keep Above Others" or a helper tool (the workaround described in #556) loses it a few seconds after launch, because a `kNormal` z-order level sends `_NET_WM_STATE_REMOVE` for `_NET_WM_STATE_ABOVE` regardless of who set it (same story for `HWND_NOTOPMOST` on Windows and the window level on macOS). Fix — apply only on change, which also gives the currently unused getter a purpose:

  ```ts
  const isOnTop = !!cfg.misc.isAlwaysOnTop;
  if (isOnTop !== getIsAlwaysOnTop()) {
    setIsAlwaysOnTop(isOnTop);
    getWin().setAlwaysOnTop(isOnTop);
  }
  ```

  Shared state starts `false`, so an untouched config issues no call at all, while enabling it (or restarting with it enabled) still applies.

- **`electron/common.const.ts:38` + `src/app/features/config/form-cfgs/misc-settings-form.const.ts:43-58`** — the gate is a *session* sniff, not the ozone backend actually in use. `isWaylandEnv()` is true whenever `XDG_SESSION_TYPE=wayland` or `WAYLAND_DISPLAY` is set, but this app forces the X11 backend on Wayland sessions in three places: Snap on any Wayland session (`electron/start-app.ts:100-107` — Ubuntu's default session), the GPU startup guard (`start-app.ts:211`), and `--ozone-platform=x11` as a documented user override (`start-app.ts:82`). In all three the app is an X11/XWayland client whose `_NET_WM_STATE_ABOVE` request KWin and Mutter do honour — yet the toggle is hidden. Combined with the point above, a Snap-on-Wayland user gets the `setAlwaysOnTop(false)` clobbering on a real X11 window *and* no toggle to turn it back on. Worth noting the hint text names KDE's "keep above others" while Plasma 6 defaults to Wayland, so the named audience is largely the hidden one. Either drop the hiding and word the hint as "some Wayland compositors ignore this" (that also removes the new `isWayland()` bridge method and the `IS_WAYLAND` frontend constant — smallest surface), or derive the flag from the ozone platform actually selected in the main process rather than from env.

- **`docs/wiki/3.02-Settings-and-Preferences.md`** — new user-facing setting with no wiki update. `docs/documentation-guide.md` maps "New/changed/removed setting, preference, or config option" to this note "in the same PR", and its `### Miscellaneous` section already documents the sibling desktop-only toggles (tray countdown, update check, local REST API) with default and platform. Nothing in CI catches this. A bullet with "desktop only, default off" and the platform caveat would do it.

### Nits

- **`electron/shared-state.ts:31`** — `getIsAlwaysOnTop()` is exported but never read; every sibling getter has a call site (`getIsMinimizeToTray` → `various-shared.ts:115`, `main-window.ts:646/694`). The change-guard above consumes it; otherwise drop it.
- **`electron/main-window.ts` `createWindow`** — the window is only made always-on-top once the renderer has hydrated, so it launches not-on-top for the length of the initial load. Optional: mirror the value into the simple store the way `isUseCustomWindowTitleBar` does (`SimpleStoreKey.IS_USE_CUSTOM_WINDOW_TITLE_BAR`, read in `createWindow` via `loadSimpleStoreAll()`) and pass `alwaysOnTop` in the `BrowserWindow` options.

### Questions

- How was the "silently no-ops on Wayland" behaviour verified — which compositor, and with which ozone platform? Given the forced-X11 paths above, a Wayland *session* frequently isn't a Wayland *client*, and I couldn't tell from the diff whether the manual check ran natively on Wayland or through XWayland.
- Related: `electron/task-widget/task-widget.ts:237` ships `alwaysOnTop: true` with no Wayland gating. If always-on-top is genuinely inert on Wayland, that widget is equally broken there and the PR's premise holds; if the widget does float for Wayland users, the premise is too broad. Do you know which it is?

---

BigWebstas:
Thanks for the review. Addressed in `87bbc70e` (branch rebased onto current `master`, the two original commits squashed into feature + fix).

### Worth fixing

- **Unconditional `setAlwaysOnTop`** — now applied only on an actual change:

  ```ts
  const isOnTop = !!cfg.misc.isAlwaysOnTop;
  if (isOnTop !== getIsAlwaysOnTop()) {
    setIsAlwaysOnTop(isOnTop);
    getWin().setAlwaysOnTop(isOnTop);
  }
  ```

  Shared state starts `false`, so a default-off config on launch and any unrelated Misc edit now issue no call at all, and an externally-set "keep above others" survives. This also gives `getIsAlwaysOnTop()` a call site.

- **Wayland gate was a session sniff** — dropped the hiding entirely rather than deriving the ozone backend. The toggle now shows on every desktop platform and the hint says *"Some Wayland compositors ignore this and manage window stacking themselves."* This removes `IS_WAYLAND`, `window.ea.isWayland()`, and the frontend constant — the smaller surface you suggested. It also fixes the Snap-on-Wayland case (real X11 client, request honoured, was hidden before) and lines up with `task-widget.ts` shipping `alwaysOnTop: true` ungated.

- **Wiki** — added a bullet to `docs/wiki/3.02-Settings-and-Preferences.md` under Miscellaneous: desktop only, default off, with the Wayland caveat.

### Nits

- `getIsAlwaysOnTop()` is now consumed by the change guard.
- Launch-time flash (window not-on-top until the renderer hydrates) — left as is for now. The mirror-into-simple-store approach is a reasonable follow-up if it matters in practice.

### Questions

- The "no-ops on Wayland" claim is no longer load-bearing since the toggle isn't gated. It was checked on KDE Plasma 6 Wayland (native, not XWayland), where `setAlwaysOnTop()` had no visible effect. With the gate gone this only affects the hint wording.
- `task-widget.ts` — now consistent: neither the widget nor this toggle is Wayland-gated.

johannesjo:
**💬 Questions first** - nothing shown to be broken, but the questions below need an answer before this can move.

Thanks for the rework. This pass only re-reviewed what changed since `02710cc`, and all three "worth fixing" items check out in the source:

- **Change guard** (`electron/ipc-handlers/app-control.ts:59-63`) - correct, and I went looking for a way to make the shared state diverge from the real window without finding one. `isAlwaysOnTop` starts `false` (`electron/shared-state.ts:9`); both senders push the full merged config (`global-config.effects.ts:260` sends `selectConfigFeatureState`, `startup.service.ts:226` sends `cfg$`), so a pre-upgrade config with the key absent lands on `!!undefined === false` and issues no call; a renderer reload keeps the same `BrowserWindow`, so suppressing the re-apply there is right; and `createWindow` refuses a second main window, so the state cannot outlive the window it describes. The residual asymmetry is understood and looks like the right trade: with the toggle on, clearing "keep above others" externally is not re-applied until restart or an uncheck/recheck, which is the price of not clobbering.
- **Wayland gate** - dropped, so the Snap-on-Wayland and `--ozone-platform=x11` cases get the toggle back, and `IS_WAYLAND` / `window.ea.isWayland()` are gone from the diff entirely. `IS_GNOME_WAYLAND` is still used at line 163 for the title-bar setting, so no dead import.
- **Wiki** - the bullet sits in `### Miscellaneous` with the sibling desktop-only toggles, with default and platform caveat.
- `getIsAlwaysOnTop()` has a call site now, and the two new tests in `electron/app-control.test.cjs` really do pin both directions plus the two no-op cases. CI runs them (`ci.yml:110`) and is green.

The persisted-field shape is also safe for older clients: `isAlwaysOnTop?` is optional with a runtime default, and `src/app/op-log/validation/validation-fn.ts:53` uses typia's `createValidate` rather than `createValidateEquals`, so an older client receiving the extra key does not reject the config.

### Nits

Both optional, neither worth a round trip on its own.

- **`src/app/features/config/form-cfgs/misc-settings-form.const.ts:48`** - this puts a second `...((IS_ELECTRON ? [...] : []) as ...)` spread directly in front of the identical one at line 60. Folding the entry into that array would drop a spread and a cast, though it does mean an unrelated toggle sits at the head of what is otherwise a self-contained local-REST-API group, so leaving it separate is defensible too. Your call.
- **`android/.idea/`** - the third commit untracks 9 files that have nothing to do with always-on-top, and at ~192 deleted lines it is now most of the diff. No objection to the change itself: `83766b3` (`chore(android): untrack .idea/gradle.xml`) is precedent for exactly this, and CI including the Android tests is green. It would just be easier to review, and easier to revert independently, as its own PR.

### Questions

- Does the take-a-break full-screen blocker still cover the window when always-on-top is on? `electron/full-screen-blocker.ts:32` creates the overlay with `alwaysOnTop: true` and `:44` sets level `'floating'`; with the main window now also always-on-top, the two sit in the same topmost band and z-order comes down to stacking order. It very likely still works, since the overlay is created last, but I could not settle it from the source without a real compositor, and the blocker is the one feature that fails visibly and silently if it loses. Did your manual pass on macOS/Windows/X11 happen to cover that combination? (The pinned task widget is only shown on the minimize/hide paths, so it does not contend with a visible main window.)

---

BigWebstas:
Tested the full-screen blocker / always-on-top interaction directly: launched the app with Playwright's `_electron`, called `win.setAlwaysOnTop(true)` on the main window (same call as `app-control.ts`), then fired `window.ea.showFullScreenBlocker(...)` — the same call `take-a-break.service.ts:247` makes on a real break.

On KDE Plasma 6 Wayland (native, same platform as the earlier always-on-top check), the overlay still fully covers the main window — nowhere in a full-desktop screenshot is the main app UI visible, only the break overlay. The blocker is created after the main window and gets `setAlwaysOnTop(true, 'floating')`, so it wins the stacking order even with the main window also always-on-top.

johannesjo:
**💬 Questions first** - nothing shown to be broken, but the question below needs an answer before this can move.

Thanks for running the blocker test, that was the right experiment. Re-checked the 3 items from the review at `c7a2cb7`:

- ⚠️ **Full-screen blocker vs. always-on-top** - partly. The run does confirm the overlay covers the main window at the moment it opens, and the source backs that up: `electron/full-screen-blocker.ts:29` builds a fresh `BrowserWindow` per break inside the IPC handler, and the `isFullScreenWindowOpen` guard at `:16` rules out any reuse path, so the overlay is always younger than the main window. What the run does not cover is below.
- ❌ **`src/app/features/config/form-cfgs/misc-settings-form.const.ts:48` duplicate `IS_ELECTRON` spread** - unchanged. The earlier pass called this your call and I'll leave it there.
- ❌ **`android/.idea/` untracking** - unchanged: 9 files, and all 202 of the diff's deletions. Still no objection to the change itself, it just makes this PR harder to revert independently. Maintainer's call.

### Questions

- **The blocker run was on the one platform where the toggle is a no-op.** Your earlier comment reports that on KDE Plasma 6 Wayland native `setAlwaysOnTop()` "had no visible effect", and the blocker test was run on that same platform, so the main window most likely was not actually on top during it and the screenshot shows the overlay covering an ordinary window.

  The level argument does not close the gap either: `setAlwaysOnTop(true, 'floating')` at `full-screen-blocker.ts:44` is not an advantage over the main window's level-less `getWin().setAlwaysOnTop(isOnTop)` at `electron/ipc-handlers/app-control.ts:62`, because `'floating'` is Electron's default level when the flag is true, and the argument is macOS/Windows-only in any case. So on macOS and Windows both windows now sit in the same band.

  What I could not settle from the source is what happens when the main window is raised *while the overlay is up*. `showOrFocus()` calls `win.show()` and then `win.focus()` unconditionally (`electron/various-shared.ts:39`), and it is reachable during a break from the global show/hide shortcut (`electron/ipc-handlers/global-shortcuts.ts:45`), a tray click, and the protocol handler (`electron/protocol-handler.ts:82`, `:145`). Before this PR the main window was normal-level and could not physically cover a floating overlay; with the toggle on it may now be able to, and the overlay cannot be dismissed early (`full-screen-blocker.ts:67-75` preventDefaults the close until the timer fires).

  Could you try that combination on macOS or Windows - toggle on, wait for the blocker, then hit the global show/hide shortcut or click the tray icon? If the main window does come over the top, `win.setAlwaysOnTop(true, 'screen-saver')` on the overlay looks like the one-line fix.

No new code landed since `c7a2cb7`: `df89e97` and `6d77139` add and then revert an unrelated community-apps entry, so the net delta against the reviewed sha is empty.

One thing that is not yours: the red `Tests` check is two e2e failures (`e2e/tests/daily-summary/daily-summary.spec.ts:6` and `e2e/tests/task-detail/ios-focus-zoom-9556.spec.ts:106`) that fail identically on unrelated PRs, so nothing in this branch caused them.

---
<sub>Automated follow-up pass (Claude Code). Anything unclear or wrong - say so and I'll take a look.</sub>

BigWebstas:
Addressed the outstanding item from the review:

- **`electron/full-screen-blocker.ts:44`** — the overlay now calls `win.setAlwaysOnTop(true, 'screen-saver')` instead of `'floating'`. `'screen-saver'` outranks the main window's level-less `setAlwaysOnTop(isOnTop)` (`app-control.ts:62`), so with the always-on-top toggle on, raising/focusing the main window during a break — global show/hide shortcut, tray click, protocol handler — can no longer cover the overlay on macOS/Windows. The level arg is a no-op on Linux, so behavior there is unchanged. Commit `63d59e299`.

I don't have macOS/Windows hardware to reproduce the original scenario natively, so this is the one-line fix from the review applied directly rather than a verified-then-fixed report — flagging that in case you want to confirm on real hardware before merge.

The duplicate `IS_ELECTRON` spread and the `android/.idea` untracking are left as-is per the earlier "maintainer's call" / no-objection notes.

johannesjo:
**🔍 Reviewed** – no blocking issues found at `89ff095`.

Thanks for the follow-up, and for being upfront that the level change is untested on macOS/Windows. Re-checked the 3 items from the review at `6d77139`:

- ✅ **Full-screen blocker vs. always-on-top** – fixed in `electron/full-screen-blocker.ts:50`. Checked against the Electron 43.5.0 source rather than hardware. On macOS `'screen-saver'` maps to `NSScreenSaverWindowLevel`, above the main window's default `NSFloatingWindowLevel`, so `showOrFocus()` can no longer bring the main window over the overlay. On Windows it also works, through a different route. Both windows are `HWND_TOPMOST` whatever the level. But a `'floating'` window gets moved just under the taskbar every time it is activated (`MoveBehindTaskBarIfNeeded`), and a `'screen-saver'` one does not. So an activated main window now lands under the overlay, where before the change it landed on top of it. A real Windows run would still be nice to have, but I don't see anything in the source that would stop it.
- ✅ **Duplicate `IS_ELECTRON` spread in `misc-settings-form.const.ts`** – left as is, which is what "your call" meant. Settled.
- ✅ **`android/.idea/` untracking** – left in, as you said. No objection from me.

### Nits

- **`electron/full-screen-blocker.ts:47-49`** – the comment says `'screen-saver'` "outranks" the main window on macOS/Windows. That is only true on macOS. On Windows both windows sit at the same topmost level, and the fix works because the level decides whether Electron tucks the window under the taskbar on activation. Something like "on macOS it is a higher window level; on Windows it keeps the overlay from being tucked under the taskbar the way the main window is" would stop the next reader from assuming there is a z-level guarantee on Windows.

---

johannesjo:
Thanks a lot for the work and for iterating through several review rounds on this, @BigWebstas — and sorry for the late call.

After weighing it, I've decided not to add an always-on-top setting for the main window. For keeping the current task visible, the existing task widget already is a compact always-on-top window, and it's where I'd rather put effort for that use case (see #9310). For pinning the full window, most desktops already offer it at the OS level (KDE/X11 "keep above", PowerToys on Windows). The setting would also bring a behaviour change to the break overlay for every user, which I'd like to avoid.

Closing for that reason, not because of the implementation. Thanks again!

PR #9739, opened 2026-08-27
refactor: prune unused chip-list-input capabilities

### LINKED ISSUE #9116: refactor: prune unused chip-list-input capabilities
## Problem

`ChipListInputComponent` exposes optional additional-action, Ctrl+Enter, and host-autofocus capabilities that neither of its two application consumers binds. These branches add component API, lifecycle, template, and styling complexity to a shared control.

This is audit finding `C01-006` in `docs/non-sync-code-simplification-audit.md` (rank 6, Tier B).

## Safety gate

Add focused live-contract characterization for both current consumers before pruning the shared component API.

## Scope

Remove only:

- The unused additional-action inputs/output, toggle helper, button block, exclusive Material import, and matching styles
- The unused `ctrlEnterSubmit` output and Ctrl+Enter emission branch
- The unused `autoFocus` host-attribute path, delayed focus timer, cleanup lifecycle, and template autofocus binding

Retain the keydown handler and all Cyrillic separator behavior. Retain suggestion filtering/sorting, chip rendering, model derivation, add/remove events, duplicate prevention, and ordinary focus behavior.

Estimated production reduction: 70–78 LOC. Expected characterization addition: 35–55 test LOC.

## Acceptance criteria

- A refreshed selector/binding search confirms that neither current consumer binds the removable API.
- Focused tests cover filtering, duplicate prevention, add/remove, focus, separators, and both current consumers.
- Cyrillic separator handling remains unchanged.
- Ctrl+Enter no longer has a component output or emission path.
- No new abstraction or replacement option is introduced.
- No sync/op-log code is changed.

## Verification

- Run `npm run checkFile` for every changed TypeScript/HTML/SCSS file.
- Run the component and both consumer test suites.
- Run a production build.
- Keyboard-smoke both consumers, including Enter, comma, Cyrillic input, add/remove, and focus behavior.

MarkQuach12:
## Problem

Closes #9116. `ChipListInputComponent` exposes optional additional-action, Ctrl+Enter and host-autofocus capabilities that neither of its two consumers (`dialog-edit-task-repeat-cfg`, `dialog-edit-issue-provider`) binds. Both bind only `label`, `model`, `suggestions` and the `addItem`/`addNewItem`/`removeItem` outputs, so the extra branches add API, lifecycle, template and styling complexity to a shared control for nothing.

## Solution

Remove only the unbound capabilities:

- the additional-action inputs/output, toggle helper, template button block, its exclusive `MatIconButton` import and matching SCSS
- the `ctrlEnterSubmit` output and its emission branch
- the `autoFocus` host-attribute path with its delayed focus timer and `OnDestroy` cleanup

The keydown handler stays for the Cyrillic separator switching and is renamed from `triggerCtrlEnterSubmit` to `onInputKeydown` since the Ctrl+Enter part is gone. No new abstraction is introduced.

Per the safety gate in the issue, a focused spec is added first (8 tests: filtering, duplicate prevention, add/remove, focus, Cyrillic separators, and the no-Ctrl+Enter contract) and passes before and after the prune.

## Verification

Binding searches for every removed input/output across src/ and e2e/ find zero consumers. Component spec 8/8, dialog-edit-task-repeat-cfg 42/42 and dialog-edit-issue-provider 3/3, all green before and after. Production build green.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki). — n/a, removes unused component API only

### DIFF
--- a/src/app/ui/chip-list-input/chip-list-input.component.html
+++ b/src/app/ui/chip-list-input/chip-list-input.component.html
@@ -13,36 +13,17 @@
           matChipRemove
           >cancel
         </mat-icon>
-
-        @if (additionalActionIcon) {
-          <button
-            (click)="additionalAction.emit(modelItem.id)"
-            [class.isToggled]="isToggled(modelItem.id)"
-            [matTooltip]="
-              isToggled(modelItem.id)
-                ? additionalActionTooltipUnToggle()
-                : additionalActionTooltip()
-            "
-            class="additional"
-            mat-icon-button
-          >
-            <mat-icon style="transform: rotate(45deg)">{{
-              additionalActionIcon
-            }}</mat-icon>
-          </button>
-        }
       </mat-chip-row>
     }
     <input
       #inputElRef
       (matChipInputTokenEnd)="add($event)"
-      [autofocus]="isAutoFocus"
       [formControl]="inputCtrl"
       [matAutocomplete]="autoElRef"
       [matChipInputAddOnBlur]="true"
       [matChipInputFor]="chipListElRef"
       [matChipInputSeparatorKeyCodes]="separatorKeysCodes"
-      (keydown)="triggerCtrlEnterSubmit($event)"
+      (keydown)="onInputKeydown($event)"
     />
   </mat-chip-grid>
   <mat-autocomplete
--- a/src/app/ui/chip-list-input/chip-list-input.component.scss
+++ b/src/app/ui/chip-list-input/chip-list-input.component.scss
@@ -5,22 +5,3 @@
     display: block;
   }
 }
-
-.additional {
-  transition: var(--transition-fast);
-  opacity: 0;
-  transform: scale(0.1);
-  width: 0;
-  position: relative;
-  z-index: 1;
-  transform-origin: center center;
-  margin-top: -2px;
-  height: 40px;
-
-  color: var(--text-color);
-
-  mat-icon {
-    font-size: 22px;
-    max-height: 32px;
-  }
-}
--- a/src/app/ui/chip-list-input/chip-list-input.component.spec.ts
+++ b/src/app/ui/chip-list-input/chip-list-input.component.spec.ts
@@ -1,0 +1,105 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { NoopAnimationsModule } from '@angular/platform-browser/animations';
+import { TranslateModule } from '@ngx-translate/core';
+import { COMMA, ENTER } from '@angular/cdk/keycodes';
+import { MatChipInputEvent } from '@angular/material/chips';
+import { MatAutocompleteSelectedEvent } from '@angular/material/autocomplete';
+import { ChipListInputComponent } from './chip-list-input.component';
+
+describe('ChipListInputComponent', () => {
+  let fixture: ComponentFixture<ChipListInputComponent>;
+  let component: ChipListInputComponent;
+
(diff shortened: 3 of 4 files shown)


johannesjo:
**✅ Ready to merge** – no blocking issues at `2315d7a`. Merge confidence 88% – I verified the removals statically (head files read in full, repo-wide symbol search, both consumer templates read in full) but did not run the new spec locally or keyboard-smoke the two dialogs, so the runtime side rests on the green `Tests` check.

Thanks for the PR! Clean prune: it removes exactly the three capabilities #9116 listed, keeps the Cyrillic separator handling, and introduces no new abstraction.

I re-ran the binding search independently and it matches your claim. Repo-wide, `chip-list-input` is used only by `dialog-edit-task-repeat-cfg.component.html:66` and `dialog-edit-issue-provider.component.html:223`, and neither binds a removed input/output nor sets the `autoFocus` host attribute. That last one is worth calling out as a good catch on your side: `autoFocus` was a `HostAttributeToken`, so it is a bare HTML attribute and a `[autoFocus]` binding search alone would have missed a consumer. `additionalActionIcon`, `additionalActionTooltip`, `additionalActionTooltipUnToggle`, `toggledItems`, `additionalAction`, `ctrlEnterSubmit` and `isToggled` have no other reference in `src/`, `e2e/` or `packages/`. `MatIcon` and `MatTooltip` are still used by the remaining template, so dropping only `MatIconButton` is right, and `.additional` had no rule outside the block you deleted.

### Nits
- **`chip-list-input.component.spec.ts:99`** – `expect('ctrlEnterSubmit' in component).toBeFalse()` pins the absence of a class property rather than a behavior, and it is the one assertion in the file that cannot pass before the prune: `output()` is a field initializer, so on master the property exists and `in` returns `true`. That makes the description's "passes before and after the prune" not quite hold for this test (the other seven do pass on both sides). The behavioral assertion right below it – Ctrl+Enter leaves `separatorKeysCodes` at the default – is the half worth keeping.
- **`chip-list-input.component.spec.ts:83`** – "does not steal focus on init" does not characterize the autofocus path you removed, for two independent reasons: `TestBed.createComponent(ChipListInputComponent)` builds a host with no attributes, so on master `isAutoFocus` stays `false` and the timer is never scheduled at all; and the test body is synchronous, so a 300 ms `setTimeout` could not have fired anyway. Reaching that path would have needed a host wrapper rendering `<chip-list-input autoFocus>` plus `fakeAsync`/`tick(300)` – which is not worth building for code that is going away. Simplest is to drop the first expectation and keep the rest as a plain "input is focusable" smoke test, or drop the test entirely.
- **`chip-list-input.component.spec.ts`** – #9116's safety gate asks for characterization of "both current consumers", and the acceptance criteria repeat it. The new spec is component-level only; the consumer half is covered by re-running the existing dialog suites, but neither of those touches the chip-list-input bindings (`dialog-edit-task-repeat-cfg.component.spec.ts` only has an unrelated `addTag` spy on the `TagService` mock, and `dialog-edit-issue-provider.component.spec.ts` has nothing). Given that neither consumer binds any removed API, the residual risk really is nil, so this is more "criterion not literally met" than a gap worth filling – flagging it so it can be waived deliberately rather than by omission.

---

johannesjo:
Thank you very much! <3

PR #9740, opened 2026-08-27
chore(take-a-break): remove the dead LOCK_SCREEN_THROTTLE / FULLSCREEN_BLOCKER_THROTTLE

### LINKED ISSUE #9356: take-a-break: LOCK_SCREEN_THROTTLE / FULLSCREEN_BLOCKER_THROTTLE are dead code
Found while working on #9305. Pre-existing, low urgency, but the constants are misleading.

### The defect

`src/app/features/take-a-break/take-a-break.service.ts`:

```ts
private _triggerLockScreenThrottledAndDelayed$ = IS_ELECTRON
  ? this._triggerLockScreenCounter$.pipe(
      distinctUntilChanged(),
      switchMap((v) =>
        !!v
          ? of(v).pipe(throttleTime(LOCK_SCREEN_THROTTLE), delay(LOCK_SCREEN_DELAY))
          : EMPTY,
      ),
    )
  : EMPTY;
```

`throttleTime` is applied **inside** the `switchMap`, to a freshly created single-value `of(v)`. A new `throttleTime` operator instance is therefore created for every emission, and a brand-new throttle always lets its first value through. So `LOCK_SCREEN_THROTTLE` (5 min) never throttles anything.

Verified experimentally with a scaled-down throttle (1000 ms) and three `true` cycles 200 ms apart: `lockScreen()` fired all three times, at +30 ms, +232 ms and +432 ms.

The same shape and the same dead constant apply to `_triggerFullscreenBlockerThrottledAndDelayed$` / `FULLSCREEN_BLOCKER_THROTTLE` a few lines below.

The only thing actually limiting how often these fire is the `distinctUntilChanged()` — i.e. the subject has to go back to `false` before another `true` counts.

### Why it hasn't bitten

To re-trigger you have to cross `takeABreakMinWorkingTime` again (60 min by default), which is far longer than the 5-minute throttle would have allowed anyway. So the guard is redundant *in practice* — but it is redundant by accident, not by design, and anyone reading the constants would reasonably believe there is a 5-minute floor between lock-screen invocations.

### Options

- Move `throttleTime` outside the `switchMap` so it applies to the subject's stream (restores the intended 5-minute floor), or
- delete both constants and the `throttleTime` calls, and say in a comment that `distinctUntilChanged()` plus the break-reminder cadence is the real spacing.

I lean toward deleting them — a guard that has never done anything is dead weight, and the honest statement is that the reminder cadence already spaces these out.

Not urgent, and deliberately kept out of the #9305 PR since it is unrelated to that fix.

MarkQuach12:
## Problem

Closes #9356. `throttleTime(LOCK_SCREEN_THROTTLE)` is applied inside the `switchMap` to a freshly created single-value `of(v)`, so a brand-new throttle instance is created per emission and always lets its first value through. The 5-minute constants have therefore never throttled anything, but anyone reading them would reasonably believe there is a 5-minute floor between invocations. Same shape for the fullscreen blocker.

## Solution

The delete option the issue leans toward: remove both constants and both `throttleTime` calls, and state in a NOTE comment that the real spacing is `distinctUntilChanged()` (the subject has to reset to false before another true counts) plus the break-reminder cadence. The two streams are also renamed from `...ThrottledAndDelayed$` to `...Delayed$` so the names stop claiming a throttle that does not exist. The 30s delays and `distinctUntilChanged()` are unchanged.

## Verification

Repo-wide search finds zero remaining references to either constant. `take-a-break.service.spec.ts` 18/18 green. Production build green.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki). — n/a, no behavior change
- [x] I have added tests for my changes (if applicable) — n/a, the removed code never had observable behavior; existing service spec covers the streams

### DIFF
--- a/src/app/features/take-a-break/take-a-break.service.ts
+++ b/src/app/features/take-a-break/take-a-break.service.ts
@@ -35,9 +35,7 @@ import { SnackService } from '../../core/snack/snack.service';
 const BREAK_TRIGGER_DURATION = 10 * 60 * 1000;
 const PING_UPDATE_BANNER_INTERVAL = 60 * 1000;
 const DESKTOP_NOTIFICATION_THROTTLE = 60 * 1000;
-const LOCK_SCREEN_THROTTLE = 5 * 60 * 1000;
 const LOCK_SCREEN_DELAY = 30 * 1000;
-const FULLSCREEN_BLOCKER_THROTTLE = 5 * 60 * 1000;
 const FULLSCREEN_BLOCKER_DELAY = 30 * 1000;
 
 // required because typescript freaks out
@@ -189,33 +187,24 @@ export class TakeABreakService {
   );
 
   private _triggerLockScreenCounter$: Subject<boolean> = new Subject();
-  private _triggerLockScreenThrottledAndDelayed$: Observable<unknown | never> =
-    IS_ELECTRON
-      ? this._triggerLockScreenCounter$.pipe(
-          distinctUntilChanged(),
-          switchMap((v) =>
-            !!v
-              ? of(v).pipe(throttleTime(LOCK_SCREEN_THROTTLE), delay(LOCK_SCREEN_DELAY))
-              : EMPTY,
-          ),
-        )
-      : EMPTY;
+  // NOTE: no throttle here or on the fullscreen blocker below — the real
+  // spacing between triggers is distinctUntilChanged() (the subject only
+  // re-fires after a reset sets it back to false) plus the break-reminder
+  // cadence feeding it.
+  private _triggerLockScreenDelayed$: Observable<unknown | never> = IS_ELECTRON
+    ? this._triggerLockScreenCounter$.pipe(
+        distinctUntilChanged(),
+        switchMap((v) => (!!v ? of(v).pipe(delay(LOCK_SCREEN_DELAY)) : EMPTY)),
+      )
+    : EMPTY;
 
   private _triggerFullscreenBlocker$: Subject<boolean> = new Subject();
-  private _triggerFullscreenBlockerThrottledAndDelayed$: Observable<unknown | never> =
-    IS_ELECTRON
-      ? this._triggerFullscreenBlocker$.pipe(
-          distinctUntilChanged(),
-          switchMap((v) =>
-            !!v
-              ? of(v).pipe(
-                  throttleTime(FULLSCREEN_BLOCKER_THROTTLE),
-                  delay(FULLSCREEN_BLOCKER_DELAY),
-                )
-              : EMPTY,
-          ),
-        )
-      : EMPTY;
+  private _triggerFullscreenBlockerDelayed$: Observable<unknown | never> = IS_ELECTRON
+    ? this._triggerFullscreenBlocker$.pipe(
+        distinctUntilChanged(),
+        switchMap((v) => (!!v ? of(v).pipe(delay(FULLSCREEN_BLOCKER_DELAY)) : EMPTY)),
+      )
+    : EMPTY;
 
   private _triggerBanner$: Observable<[number, GlobalConfigState, boolean, boolean]> =
     this.timeWorkingWithoutABreak$.pipe(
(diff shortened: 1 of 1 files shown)


johannesjo:
**🔍 Reviewed** – no blocking issues found at `3c93845`.

Thanks for the PR! I confirmed the deletion is a genuine no-op rather than taking it on trust. With `of(v)` as the source, `throttle`'s `leading: true` branch emits the single value synchronously, and the source's `complete` then takes the `!(trailing && hasValue && throttled && !throttled.closed)` path in rxjs 7.8.2's `throttle.ts`, so the operator completed immediately and never held the value or the completion back from the `delay` below it. A fresh instance per emission also carried no state across emissions, so both streams behave identically before and after. `throttleTime` is still used at `:232` and `:237`, so the import stays live, and nothing outside this file referenced either constant or either stream. CI is green.

### Nits
- **`src/app/features/take-a-break/take-a-break.service.ts:190`** – the new NOTE names only a reset as what puts the subject back to `false`, but `snooze()` at `:335` pushes `false` to both subjects directly at `:337-338`, bypassing `_triggerReset$`. Snoozing also does not reset `timeWorkingWithoutABreak$`, so once `_snoozeActive$` flips back the banner passes its filter on the next tick, `next(true)` gets through `distinctUntilChanged()` again, and `lockScreen()` fires `LOCK_SCREEN_DELAY` later. The shortest real gap is therefore `takeABreakSnoozeTime + 30s`, not “until the min-working-time threshold is crossed again”. Since an honest statement of the real spacing is the whole deliverable here, it is worth naming the snooze path too – something like “the subject only re-fires after a reset or a snooze sets it back to false, so the shortest real gap is `takeABreakSnoozeTime` plus the delay below”.

---

johannesjo:
Good change — merging this as-is. Thanks for tracking down that the throttle was inert rather than just deleting a constant that looked unused.

I want to record why I'm **not** taking the other option from #9356 (restoring a real floor by hoisting `throttleTime` out of the `switchMap`), because it's the tempting follow-up and it's a trap in both placements:

**Before `distinctUntilChanged()`** — the subject carries `true` *and* `false`, so the throttle swallows the `false` un-latch. `distinctUntilChanged` then still holds `true`, the next `true` is a duplicate, and lock-screen is silently dead for the rest of the session. That's the #9305 failure class the NOTE at the top of `_triggerReset$` exists to warn about.

**After `distinctUntilChanged()`** — subtler, and user-visible. Today `switchMap` is what *cancels* a pending lock: `true` starts the 30s `delay`, and a `false` arriving inside that window switches to `EMPTY`, so hitting snooze or reset within 30s of the banner calls off the lock. Put a throttle in front of the `switchMap` and that cancelling `false` gets swallowed — you click snooze and your screen locks 20 seconds later anyway.

And the floor wasn't really the throttle's job to begin with. `_triggerBanner$` re-fires every 60s while the banner is up, and `distinctUntilChanged` is what swallows those repeats — the subject can only re-trigger after something sets it back to `false`, and there are exactly two such places: `_triggerReset$` and `snooze()`. So the throttle's only reachable effect was capping the snooze path, where the user has explicitly asked to be reminded again after a duration they chose (`takeABreakSnoozeTime` has no minimum). Overriding that with a hidden 5-minute floor is the wrong default for this app.

Two small things, neither blocking:

1. Your NOTE says spacing comes from `distinctUntilChanged()` "plus the break-reminder cadence feeding it". Accurate, but the sharper statement is that re-triggering requires an un-latch, and there are exactly two — `_triggerReset$` and `snooze()`. Naming them makes the next reader's reasoning correct rather than approximately correct. Happy to take it as a follow-up or not at all.

2. The PR checklist says "existing service spec covers the streams" — it doesn't. `take-a-break.service.spec.ts` touches neither stream, and `IS_ELECTRON` is `false` under Karma, so both are `EMPTY` there. It doesn't change my decision (the removal is a provable no-op), but worth being strict about: a checklist claim about coverage is worth checking before ticking, because on a change that *wasn't* inert it's exactly the claim that would stop someone looking closer.

PR #9741, opened 2026-08-27
chore(supersync-server): drop the unused operations (userId, clientId) index

### LINKED ISSUE #8352: SuperSync server: drop likely-unused operations indexes (userId, clientId) and (userId, receivedAt)
**Area:** server (Postgres schema) · **Type:** performance · **Value:** medium (write amplification on every insert into the hottest table)

## Problem
Exhaustive audit of every Prisma and raw-SQL query against `operations` found:

- **`@@index([userId, clientId])` (`prisma/schema.prisma:85`): no query can use it.** Downloads filter `clientId: { not: excludeClient }` (inequality post-filter) under `(user_id, server_seq)` ordering (`operation-download.service.ts:119, 216`); duplicate checks are PK lookups (`operation-upload.service.ts:363, 634, 802`, etc.); conflict detection uses the `(user_id, entity_type, entity_id, server_seq)` index (`conflict.ts:71, 177, 339`); full-state lookups use the partial `operations_user_id_full_state_server_seq_idx`; the vector-clock aggregates group over LATERAL jsonb keys, not the `client_id` column.
- **`@@index([userId, receivedAt])` (`:86`): exactly one candidate query**, and its shape favors `(user_id, server_seq)` — the cleanup query (`storage-quota.service.ts:492`) filters `serverSeq <= lastSnapshotSeq` + `receivedAt < cutoff` with `ORDER BY serverSeq ASC LIMIT n`; `receivedAt` is a correlated post-filter (both monotonic per user). Minor caveat: if a fresh snapshot covers many ops younger than the retention cutoff, the `(user_id, server_seq)` plan reads-and-discards them — bounded and rare.

The 20260512 migration already pruned three redundant indexes for exactly this write-amplification reason; these two look like leftovers from that audit.

## Proposed fix
Confirm `idx_scan = 0` via `pg_stat_user_indexes` for `operations_user_id_client_id_idx` and `operations_user_id_received_at_idx` over a representative window on a production-like instance, then drop both with `DROP INDEX CONCURRENTLY`, following the 20260512 migration's pattern.

MarkQuach12:
## Problem

Closes #8352. The issue's audit found two `operations` indexes no query can use, both pure write amplification on the hottest table. Re-verifying at HEAD, only half of that still holds:

- `@@index([userId, clientId])`: still unused. Downloads only exclude by `clientId: { not: ... }`, which cannot seek this btree; duplicate checks hit the `(userId, serverSeq)` unique; conflict detection uses `(userId, entityType, entityId, serverSeq)`; `clientId` equality lookups go to `sync_devices`, not `operations`.
- `@@index([userId, receivedAt])`: no longer droppable. Since #9733 the old-ops fresh-prefix probe orders by `receivedAt` specifically so the planner holds the measured 9-buffer plan instead of the 60,329-buffer `(user_id, server_seq)` plan that hit the 60s statement timeout in production, and `tests/integration/old-ops-probe-plan.integration.spec.ts` names the index. Dropping it would deterministically force the catastrophic plan. The issue predates that fix.

## Solution

Drop only `operations_user_id_client_id_idx`: a new migration mirroring the 20260512 prune pattern plus the matching `@@index` removal in `schema.prisma`. The migration is a single `DROP INDEX CONCURRENTLY IF EXISTS` statement, the preferred shape per `prisma/migrations/README.md` (applies natively, no out-of-band recovery), and its comment documents why `(user_id, received_at)` must stay.

A shape test is added to `tests/migration-sql.spec.ts` in the style of the neighboring entries; it also pins that this migration never touches the `received_at` index.

## Verification

`prisma validate` clean. `tests/migration-sql.spec.ts` + `tests/migrate-deploy-script.spec.ts`: 79/79 green. Repo-wide grep for `operations_user_id_client_id_idx` finds only `0_init/migration.sql` (no code or test references).

Pre-merge check for the maintainer, per the issue: confirm `idx_scan = 0` for `operations_user_id_client_id_idx` in `pg_stat_user_indexes` on the hosted instance.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki). — n/a, server schema internals; the migration comment carries the rationale

### DIFF
--- a/packages/super-sync-server/prisma/migrations/20260828000000_drop_unused_operations_user_id_client_id_index/migration.sql
+++ b/packages/super-sync-server/prisma/migrations/20260828000000_drop_unused_operations_user_id_client_id_index/migration.sql
@@ -1,0 +1,12 @@
+-- Remove write amplification on the hottest table for an index no query can
+-- use. Downloads only exclude by client_id (clientId not-equals), which cannot
+-- seek this btree. Duplicate checks hit the (user_id, server_seq) unique key.
+-- Conflict detection uses (user_id, entity_type, entity_id, server_seq).
+-- Device lookups by client_id go to sync_devices, not operations.
+-- (user_id, received_at) is deliberately KEPT: the old-ops fresh-prefix probe
+-- orders by received_at to hold its measured 9-buffer plan (PR #9733, guarded
+-- by tests/integration/old-ops-probe-plan.integration.spec.ts).
+-- Reinstating is cheap if a future query filters operations by client_id.
+-- Single statement so prisma migrate deploy applies it natively without the
+-- out-of-band CONCURRENTLY recovery.
+DROP INDEX CONCURRENTLY IF EXISTS "operations_user_id_client_id_idx";
--- a/packages/super-sync-server/prisma/schema.prisma
+++ b/packages/super-sync-server/prisma/schema.prisma
@@ -109,7 +109,6 @@ model Operation {
 
   @@unique([userId, serverSeq])
   @@index([userId, entityType, entityId, serverSeq])
-  @@index([userId, clientId])
   @@index([userId, receivedAt])
   // Multi-entity conflict lookups (#8334) probe entity_ids with @> / &&. This MUST stay
   // modelled here, not only in the 20260613000001 raw migration: `prisma db push` builds
--- a/packages/super-sync-server/tests/migration-sql.spec.ts
+++ b/packages/super-sync-server/tests/migration-sql.spec.ts
@@ -78,6 +78,30 @@ describe('performance migrations', () => {
     expect(migrationSql).not.toMatch(/\bBEGIN\b|\bCOMMIT\b/i);
   });
 
+  it('drops the unused (user_id, client_id) index as a single native-apply statement', () => {
+    const migrationSql = readMigration(
+      '20260828000000_drop_unused_operations_user_id_client_id_index',
+    );
+
+    expect(migrationSql).toContain(
+      'DROP INDEX CONCURRENTLY IF EXISTS "operations_user_id_client_id_idx"',
+    );
+    // The received_at index is load-bearing for the old-ops fresh-prefix probe
+    // plan and must never be dropped here.
+    expect(migrationSql).not.toContain('"operations_user_id_received_at_idx"');
+    // Single statement so `prisma migrate deploy` applies it natively without
+    // the out-of-band CONCURRENTLY recovery.
+    const sqlWithoutComments = migrationSql
+      .split('\n')
+      .filter((line) => !line.trimStart().startsWith('--'))
+      .join('\n');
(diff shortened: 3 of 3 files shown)


johannesjo:
**🔍 Reviewed** – no blocking issues found at `0bc3e6b`.

Thanks for the PR! Tight, well-argued drop – I re-verified the audit at this head and it holds: the only `clientId` predicate on `operations` in the serving path is the non-seekable `clientId: { not: ... }` filter in `packages/super-sync-server/src/sync/services/operation-download.service.ts:201`; every remaining btree on `operations` still leads with `user_id`, so no query loses prefix coverage; and the kept `(user_id, received_at)` index is genuinely load-bearing for the #9733 probe plan. The migration follows the repo's preferred single-statement native-apply CONCURRENTLY shape, is idempotent on re-run via `IF EXISTS`, and the shape test pins both the drop and the received_at guard.

### Nits
- **`packages/super-sync-server/prisma/migrations/20260828000000_drop_unused_operations_user_id_client_id_index/migration.sql:1`** – "an index no query can use" is slightly overbroad: two maintainer-only analytics queries do aggregate on the column (`scripts/analyze-storage.ts:649` groups `operations` by `client_id`; `:716` counts distinct `client_id`). Both already visit every heap tuple for the user, so losing the index is negligible there – but consider narrowing the wording to "no serving query", so a future re-audit doesn't conclude the comment was wrong.

---

johannesjo:
Thank you very much! <3

johannesjo:
Decisions on the two open points – both go your way:

**1. Merging without the hosted `pg_stat_user_indexes` check.** The issue asked for
`idx_scan = 0` to be confirmed on the instance first, and I am waiving that. The static audit is
strong enough on its own (no serving query filters `operations` by `client_id` equality, and
every remaining btree still leads with `user_id`), and more importantly the change is reversible
– if a slow query ever shows up, `CREATE INDEX CONCURRENTLY "operations_user_id_client_id_idx" ON
"operations"("user_id", "client_id");` puts it straight back. Gating this on manual DB access
would stall the PR for a modest upside. I will run the query opportunistically, but it is not a
blocker.

**2. Closing #8352 in full, rather than re-scoping it.** Re-scoping would leave an issue open for
"consider dropping `operations_user_id_received_at_idx`", which is now something we have decided
*against* – #9733's probe plan made that index load-bearing. Better to close it and let the
reasoning live in this PR.

Good catch keeping `received_at` and calling out why. Also noted the nit about
`scripts/analyze-storage.ts` aggregating on `client_id`; a maintainer-only script doing a seq
scan is fine, but please do soften the migration comment so it does not claim *no* query can use
the index.

PR #9742, opened 2026-08-27
chore(take-a-break): remove the dead idle-dialog break-counter branch

### LINKED ISSUE #9352: Break-reminder counter ignores idle time tracked to a task (dead branch + vacuous test)
Split out of #9351 so it can be decided on its own merits.

### The situation

`TakeABreakService._tick$` handles `idleDialogResult` like this:

```ts
// without a reset, time tracked to tasks still counts as work; break
// items don't (but they don't reset the timer either).
const noBreakTime = trackItems
  .filter((t) => t.type === 'TASK')
  .reduce((acc, t) => acc + (typeof t.time === 'number' ? t.time : 0), 0);
return noBreakTime > 0 ? of(noBreakTime) : EMPTY;
```

Only SPLIT mode sends a resolved number. BREAK and TASK mode send the `'IDLE_TIME'` placeholder, with the duration in the action's separate `idleTime` field (`dialog-idle.component.ts:311,321`; the placeholder is resolved locally inside `handleIdleDialogResult$`, *after* the action is built, so the payload really does carry the string).

So for the two common modes `noBreakTime` is always `0` and the branch returns `EMPTY`. The comment describes behaviour that hasn't happened since `aaca8d9657` (Jan 2025) — that commit rewrote `acc + (typeof t.time === 'number' ? t.time : idleTime)` to `... : 0`, deleting the fallback that made it work.

The guarding test doesn't catch it: `take-a-break.service.spec.ts` defines its `TASK_ITEM` with `time: 60000`, a SPLIT-shaped payload the dialog never produces in TASK mode, so the assertion passes while the real path is untested.

### The decision

Two options, and I don't think the answer is obvious:

**A — implement the comment.** Map `'IDLE_TIME'` to `idleTime`. Tracking a long absence to a task then counts as work-without-a-break, and the reminder can fire immediately afterwards. Defensible: the user just asserted they were working.

**B — delete the branch.** Return `EMPTY` whenever `isResetBreakTimer` is false and drop the `reduce` entirely. Also defensible: this has been the *effective* behaviour for 18 months and nobody has reported it, which is decent evidence nobody wants the nagging.

Whichever way it goes, the vacuous fixture should be replaced with the real payload shape so the chosen behaviour is actually covered.

### One thing to know before picking A

There is a pre-existing double count that only becomes reachable under A. During the window before the idle trigger fires (up to `minIdleTime`, default 5 min) the user still has a current task, so live ticks are already accumulating into the break counter. `idleTime` then covers that same window again. Under A the reminder therefore fires slightly earlier than the numbers justify — bounded by `minIdleTime`, but real.

### Not urgent

No user has reported this. It surfaced only because #9305 sent me through this file. Filing it so the dead code and the misleading comment don't quietly survive another year.

MarkQuach12:
## Problem

Closes #9352. The `idleDialogResult` arm of the break counter's `_tick$` merge has been dead for the common paths since Jan 2025: BREAK and TASK modes send the `'IDLE_TIME'` placeholder rather than numbers, so the reduce yields 0 and the arm returns `EMPTY`. Only the rare SPLIT shape ever contributed, the code comment described behavior that no longer happened, and the guarding test used a SPLIT-shaped fixture the dialog never produces in TASK mode, so it passed vacuously.

## Solution

Option B from the issue: delete the arm entirely, so idle-dialog results never feed the break counter. Rationale: this has been the effective behavior for BREAK/TASK for 18 months with no complaints, and option A (counting idle time as work) inherits the double-count the issue documents, since live ticks before the idle trigger already accumulated the same window. The dialog's reset checkbox is unaffected: it flows through `_triggerIdleDialogReset$` into `_triggerReset$`, verified by the existing reset specs which pass unchanged.

A NOTE at the merge site records the decision and points at the reset path.

The spec's vacuous fixture is replaced with the real payload shapes: a `time: 'IDLE_TIME'` item (BREAK/TASK mode) and the SPLIT numeric shape, both now asserted not to advance the counter, with the edge-trigger coverage of the old opt-out test preserved via `otherNoBreakTIme$`.

Behavior change to be aware of: SPLIT-mode idle time tracked to tasks previously did count toward the reminder; after this it does not. That is what option B means, and flipping to option A later is a small diff on top of the new non-vacuous fixtures.

## Verification

The reworked SPLIT assertions fail against the old service (Expected 70000 to be 10000, etc.) and the full suite is 19/19 green with the deletion. `npm run checkFile` clean on both files.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/take-a-break/take-a-break.service.spec.ts
+++ b/src/app/features/take-a-break/take-a-break.service.spec.ts
@@ -98,6 +98,12 @@ describe('TakeABreakService', () => {
       title: 'Some task',
       simpleCounterToggleBtns: [],
     };
+    const TASK_ITEM: IdleTrackItem = {
+      type: 'TASK',
+      time: 'IDLE_TIME',
+      title: 'Some task',
+      simpleCounterToggleBtns: [],
+    };
 
     const dialogResult = (
       trackItems: IdleTrackItem[],
@@ -139,9 +145,16 @@ describe('TakeABreakService', () => {
       expect(current()).toBe(10000);
     });
 
-    it('adds tracked task time but not break time when not resetting', () => {
+    // idle time tracked to tasks does not count toward the break reminder,
+    // regardless of the shape the dialog mode sends (#9352)
+    it("does not add idle time tracked to tasks ('IDLE_TIME' shape)", () => {
+      actions$.next(dialogResult([BREAK_ITEM, TASK_ITEM], false));
+      expect(current()).toBe(10000);
+    });
+
+    it('does not add idle time tracked to tasks (SPLIT numeric shape)', () => {
       actions$.next(dialogResult([BREAK_ITEM, SPLIT_TASK_ITEM], false));
-      expect(current()).toBe(10000 + 60000);
+      expect(current()).toBe(10000);
     });
 
     it('dismisses the reminder banner when the timer is reset', () => {
@@ -341,17 +354,16 @@ describe('TakeABreakService', () => {
     // The other half of the same overlap, and the reason the reset is edge- and
     // not level-triggered: once it has fired, time added later in the SAME
     // untracked stretch survives. With a level trigger the next tick wiped it
-    // again -- and every tick after -- so unchecking the dialog's reset box was
-    // inert for any absence over BREAK_TRIGGER_DURATION. SPLIT is the only mode
-    // that sends a resolved number, and handleIdleDialogResult$ re-selects a task
-    // only for a single task item, so with 2+ items the stretch keeps running.
-    it('keeps task time from an opt-out added after the untracked-stretch reset', () => {
+    // again -- and every tick after. Idle-dialog results no longer feed the
+    // counter at all (#9352), so the late addition comes from otherNoBreakTIme$.
+    it('keeps time added after the untracked-stretch reset', () => {
       const emitted: number[] = [];
       const sub = service.timeWorkingWithoutABreak$.subscribe((v) => emitted.push(v));
 
       tick$.next({ duration: 11 * 60000, date: '2026-07-28', timestamp: 0 });
       expect(emitted[emitted.length - 1]).toBe(0);
 
+      // a SPLIT-mode opt-out contributes nothing to the counter (#9352)
       actions$.next(
         idleDialogResult({
           trackItems: [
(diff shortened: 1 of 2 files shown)


johannesjo:
**🔍 Reviewed** – no blocking issues found at `6192dc6`.

Thanks for the PR! The dead-branch analysis holds up end to end: `dialog-idle.component.ts:311,321` really does send `time: 'IDLE_TIME'` for BREAK and TASK, `idle.effects.ts:353` resolves the placeholder only after the action is built, and `aaca8d9657` is where the `: idleTime` fallback became `: 0`. The reworked SPLIT assertion is genuinely non-vacuous (the old arm would have produced `70000` there), the reset path through `_triggerIdleDialogReset$` is untouched, no import is left dangling, and the late addition in the renamed edge-trigger test now comes from a real production producer (`tracking-reminder.service.ts:208`) instead of the deleted arm.

### Nits
- **`src/app/features/take-a-break/take-a-break.service.ts:76-88`** – the NOTE above `_triggerSimpleBreakReset$` is left stale by the deletion. Its "Semantics" paragraph still justifies the edge trigger with the idle dialog: *"with a level trigger the dialog's 'reset break timer' checkbox was inert for any absence over BREAK_TRIGGER_DURATION: unchecking it kept the tracked time for one tick before the next tick wiped it again."* After this change the dialog contributes nothing to the counter under either trigger style, so that illustration no longer applies. The general claim (time added later in the same untracked stretch survives) still holds, but now only via `otherNoBreakTIme$`. You already updated the mirror of this comment in the spec (`take-a-break.service.spec.ts:354-358`); worth the same edit here, since it is the source-of-truth version.

### Notes
- One correction to the risk statement in the PR body, which frames the lost contribution as needing an explicit opt-out: `_updateResetBreakTimerDefault()` (`dialog-idle.component.ts:386-393`, re-run from `onSplitTypeChange`/`removeSplitItem`) sets `isResetBreakTimer = false` automatically for any SPLIT containing no BREAK item. So "split the absence across two tasks, no break portion" reached the deleted arm with the checkbox never touched – that, rather than an explicit opt-out, is the realistic shape of the behavior change. It stays bounded: for absences over `BREAK_TRIGGER_DURATION` the untracked-stretch reset has already zeroed the counter, so the effect is a break reminder arriving later, not a wrong number.

---

johannesjo:

PR #9743, opened 2026-08-27
fix(task-detail-panel): make isOverdue flip on day rollover

### LINKED ISSUE #9393: fix(task-detail-panel): isOverdue never flips while the panel stays open
## Problem

Split out from #9072, which named two `computed` bodies reading a non-reactive clock. The habit-tracker strip is fixed in #9380; this is the other one, so that it doesn't get closed along with it.

`src/app/features/tasks/task-detail-panel/task-detail-panel.component.ts:367`:

```ts
isOverdue = computed(() => {
  const t = this.task();
  return !!(
    !t.isDone &&
    ((t.dueWithTime && t.dueWithTime < Date.now()) ||
      (t.dueDay &&
        isDBDateStr(t.dueDay) &&
        t.dueDay !== getDbDateStr() &&
        t.dueDay < getDbDateStr()))
  );
});
```

`Date.now()` and the argument-less `getDbDateStr()` are not reactive, and Angular caches computeds. The only dependency is `task()`, so the value is pinned to whatever the clock said the last time the task object changed reference.

`isOverdue()` feeds `[class.color-warn]` on the schedule row (`task-detail-panel.component.html:136`), so with the panel open, a task whose due time passes (or whose `dueDay` becomes yesterday at the day boundary) keeps rendering as not overdue.

## Severity

Lower than #9380. This is a missing colour on one row, not a write to the wrong date, so it is cosmetic rather than data-affecting.

It is also narrower than it first looks: `task` is `input.required<TaskWithSubTasks>()`, so any new object reference from the parent recomputes it. A task that is actively being tracked gets a new `timeSpent` every second and therefore stays correct by accident. The stale case is a task sitting in the detail panel with no tracking running and no edits, which is the ordinary case for "I opened this task to look at it".

## Fix

Same shape as #9380: give the computed a day-change dependency by reading `GlobalTrackingIntervalService.todayDateStr()` inside it. That covers the `dueDay` half exactly.

The `dueWithTime` half is a time-of-day threshold rather than a day threshold, so `todayDateStr()` alone will not flip it mid-day. If that case is worth covering, it needs a coarser tick, and it is worth deciding whether a per-minute recompute earns its place for a colour class, or whether the `dueDay` half alone is enough.

## Notes

Not verified at runtime, same as #9072. Reported from reading the dependency graph, and the line references were re-checked against master while reviewing #9380.

MarkQuach12:
## Problem

Closes #9393. The `isOverdue` computed in the task detail panel reads `Date.now()` and the argless `getDbDateStr()`, neither of which is reactive, so its only dependency is `task()`. With the panel open on a task that is not being tracked or edited, the overdue colour on the schedule row never flips when the due time passes or the day rolls over.

## Solution

Same shape as the sibling fix in #9380: inject `GlobalTrackingIntervalService` and read its `todayDateStr()` signal inside the computed, replacing both argless `getDbDateStr()` calls. That gives the `dueDay` comparison a reactive day-boundary dependency, and as a side effect uses the logical day (start-of-next-day offset) instead of the raw calendar day, matching the sibling fix and sync rule 4.

The `dueWithTime < Date.now()` half deliberately stays as is: flipping a colour class mid-day would need a per-minute recompute, which does not seem to earn its place; the day-change recompute now refreshes it at least daily. Happy to add a coarse tick if you want that case covered too.

Noticed but left out of scope: `isPlannedForTodayDay` and `isDeadlineOverdue` in the same file still read the non-reactive argless `getDbDateStr()` and have the same latent staleness. Can follow up if wanted.

## Verification

New spec case drives the mocked `todayDateStr` signal across a day boundary without a new task reference and asserts `isOverdue()` flips. It fails without the component change (`Expected false to be true`, 1 FAILED of 29) and the full spec is 29/29 green with it. `npm run checkFile` clean on both files. Production build green.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/tasks/task-detail-panel/task-detail-panel.component.spec.ts
+++ b/src/app/features/tasks/task-detail-panel/task-detail-panel.component.spec.ts
@@ -7,6 +7,7 @@ import { TaskAttachmentService } from '../task-attachment/task-attachment.servic
 import { TaskService } from '../task.service';
 import { LayoutService } from '../../../core-ui/layout/layout.service';
 import { GlobalConfigService } from '../../config/global-config.service';
+import { GlobalTrackingIntervalService } from '../../../core/global-tracking-interval/global-tracking-interval.service';
 import { IssueService } from '../../issue/issue.service';
 import { TaskRepeatCfgService } from '../../task-repeat-cfg/task-repeat-cfg.service';
 import { MatDialog } from '@angular/material/dialog';
@@ -36,9 +37,11 @@ describe('TaskDetailPanelComponent', () => {
   let mockAttachmentService: jasmine.SpyObj<TaskAttachmentService>;
   let mockTaskService: jasmine.SpyObj<TaskService>;
   let isXs: WritableSignal<boolean>;
+  let todayDateStr: WritableSignal<string>;
 
   beforeEach(async () => {
     isXs = signal(true);
+    todayDateStr = signal('2027-01-01');
     mockClipboardImageService = jasmine.createSpyObj('ClipboardImageService', [
       'handlePasteWithProgress',
     ]);
@@ -111,6 +114,7 @@ describe('TaskDetailPanelComponent', () => {
         { provide: DateTimeFormatService, useValue: mockDateTimeFormatService },
         { provide: Store, useValue: mockStore },
         { provide: MentionConfigService, useValue: { mentionConfig$: EMPTY } },
+        { provide: GlobalTrackingIntervalService, useValue: { todayDateStr } },
       ],
     })
       .overrideComponent(TaskContextMenuComponent, {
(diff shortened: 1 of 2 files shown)


johannesjo:

Thanks for the PR! The core fix is right: reading `todayDateStr()` inside the computed is exactly the shape `task.component.ts:249` already uses on master, the new spec genuinely fails without the component change, and the calendar-day to logical-day switch that comes with it is a bonus rather than a regression (the offset is clamped to >= 0, so logical today is never later than calendar today). One thing the narrow scope leaves behind turns into a visible contradiction on the row you just fixed, though.

To be clear about what I am *not* asking for: the `dueWithTime` half staying on `Date.now()` is fine as scoped – #9393 leaves that decision open on purpose, so no need to add a tick here.

### Should fix before merge

- **`src/app/features/tasks/task-detail-panel/task-detail-panel.component.ts:361`** – `isPlannedForTodayDay` still reads the argless `getDbDateStr()`, so it keeps the exact staleness #9393 is about, and it drives the sun icon + "Today" label inside the very same `task-detail-item` whose `[class.color-warn]` you just made reactive (`task-detail-panel.component.html:136` and `:151`). Panel open on an untracked, unedited task with `dueDay` = today, midnight passes: `isOverdue` recomputes and the row turns red, while `isPlannedForTodayDay` is still cached at `true`, so the row renders warn-coloured *and* still says "Today". Before this PR both halves froze together, so the row was at least self-consistent. Same one-line substitution as two lines below:

  ```ts
  return !!task.dueDay && task.dueDay === this._globalTrackingIntervalService.todayDateStr();
  ```

  which is also the form `task.component.ts:312` already uses.

### Nits

- **`src/app/features/tasks/task-detail-panel/task-detail-panel.component.ts:419`** – `isDeadlineOverdue` has the same latent staleness on the deadline row (`...component.html:177`). It is no worse than master and this PR does not touch it, so it is not blocking – but the service is already injected and `task.component.ts:328` is literally the fixed form of the same expression:

  ```ts
  isDeadlineOverdue = computed(() =>
    isDeadlineOverdueFn(this.task(), this._globalTrackingIntervalService.todayDateStr()),
  );
  ```

  Folding it in would leave the file with no argless `getDbDateStr()` reads at all, matching `task.component.ts`, which is fully converted.

---

PR #9744, opened 2026-08-27
chore(supersync-server): expire pending passkey registrations and abandoned unverified users

### LINKED ISSUE #9031: chore(supersync-server): no cleanup/expiry for pending_passkey_registrations or abandoned unverified users
Follow-up from a post-merge sync review.

### Problem
`packages/super-sync-server/src/sync/cleanup.ts` — `runDailyCleanup` purges old ops, stale devices, expired rate-limit counters, and request-dedup entries, but nothing prunes:

- expired rows in `pending_passkey_registrations`, or
- abandoned **unverified** `user` rows.

A `pending_passkey_registration` is only ever deleted when a verification link for that user is clicked (`auth.ts`). An expired pending row (its expiry is checked at read time in `auth.ts`) is never physically removed. The `add_pending_passkey_registrations` migration also intentionally keeps the row on email-send failure rather than deleting it.

### Impact
Not takeover — storage growth + retention of abandoned-registration PII (email address + credential public key) that persists indefinitely. The registration endpoint is rate-limited (≈50 / 15 min / IP), but that's per-IP, so a distributed or patient client can accumulate unbounded `user` + `pending_passkey_registration` rows.

### Fix
Add a cleanup pass to `runDailyCleanup` that deletes `pending_passkey_registrations` past their expiry and unverified `user` rows with no remaining pending registration older than some grace window. Mirror the existing `deleteStaleDevices` / rate-limit / dedup cleanup shape.

Adjacent to (but distinct from) the pre-hijack hardening tracked separately.

MarkQuach12:
## Problem

Closes #9031. `runDailyCleanup` purges old ops, stale devices, expired rate-limit counters and request-dedup entries, but nothing prunes expired `pending_passkey_registrations` rows (expiry is only checked at read time in auth.ts, rows are only physically deleted when the verification link is clicked) or abandoned unverified `user` rows. Abandoned registrations therefore retain PII (email plus credential public key) indefinitely, and a patient or distributed client can accumulate unbounded rows past the per-IP rate limit.

## Solution

Two new steps in `runDailyCleanup`, in the exact try/catch + logging shape of steps 2 to 4, backed by two methods in `SyncService` next to `deleteStaleDevices`:

- Step 5 deletes `pending_passkey_registrations` whose `verificationTokenExpiresAt` is past, using the same strict comparison auth.ts applies at read time.
- Step 6 deletes unverified users older than a grace window with nothing in flight: `isVerified: 0`, `createdAt` past the window, no pending passkey registration row, and no live verification token. The last guard goes one step beyond the issue's sketch, deliberately: `registerWithMagicLink` refreshes the token on the same old user row without bumping `createdAt`, so without it the sweep could delete an old row while a freshly emailed magic link is still valid.

Grace window reuses the unified `retentionMs` (45 days) rather than adding a second knob; verification tokens live 24h, so the margin is wide. Verified users can never match the predicate, and unverified users are denied all authenticated routes, so the cascading deletes (operations, devices, sync state, passkeys, pending registrations) cannot touch real sync data.

## Verification

New cases in `cleanup.spec.ts` (both steps wired into the daily job, cutoffs asserted, one step failing does not stop the others) and `sync.service.spec.ts` (expired pending row deleted, unexpired kept; abandoned old user deleted; verified never deleted; recent, has-pending, and live-token users all kept), mirroring the existing harnesses. Targeted suites: 171/171 green (cleanup, sync.service, sync-operations, device.service). The full package run has 58 pre-existing failures that are identical on the pristine tree (stale local `@sp/sync-core` dist, unrelated to this change; verified by stashing). `npm run checkFile` clean on all five changed files.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [x] Other (please describe): server data-hygiene / PII retention fix

## Checklist

- [x] I have included relevant changes to the documentation/[wiki](https://github.com/super-productivity/super-productivity/tree/master/docs/wiki). — n/a, no operator-facing config change; the constants and predicates are documented in code

### DIFF
--- a/packages/super-sync-server/src/sync/cleanup.ts
+++ b/packages/super-sync-server/src/sync/cleanup.ts
@@ -12,6 +12,9 @@ const reconcileTimers: Set<NodeJS.Timeout> = new Set();
 const RECONCILE_INTERVAL_MS = 5_000;
 const RECONCILE_BUDGET_MS = 60 * 60 * 1000;
 const INITIAL_CLEANUP_DELAY_MS = 10_000;
+// Grace window before an abandoned (never-verified, nothing in flight) account is
+// deleted — reuses the unified retention period rather than adding a second knob.
+const UNVERIFIED_USER_GRACE_MS = DEFAULT_SYNC_CONFIG.retentionMs;
 
 /**
  * Runs all cleanup tasks in a single daily job.
@@ -71,6 +74,32 @@ const runDailyCleanup = async (): Promise<void> => {
     }
   } catch (error) {
     Logger.error(`Cleanup [request-dedup] failed: ${error}`);
+  }
+
+  // 5. Delete expired pending passkey registrations (expiry is otherwise only
+  // checked at read time, so abandoned rows would be retained forever)
+  try {
+    const deleted = await syncService.deleteExpiredPendingPasskeyRegistrations(
+      Date.now(),
+    );
+    if (deleted > 0) {
+      Logger.info(`Cleanup [pending-passkeys]: removed ${deleted} entries`);
+    }
+  } catch (error) {
+    Logger.error(`Cleanup [pending-passkeys] failed: ${error}`);
+  }
+
+  // 6. Delete abandoned unverified users (never verified, no registration still
+  // in flight, older than the grace window; verified users can never match)
+  try {
+    const deleted = await syncService.deleteAbandonedUnverifiedUsers(
+      Date.now() - UNVERIFIED_USER_GRACE_MS,
+    );
+    if (deleted > 0) {
+      Logger.info(`Cleanup [unverified-users]: removed ${deleted} entries`);
+    }
+  } catch (error) {
+    Logger.error(`Cleanup [unverified-users] failed: ${error}`);
   }
 };
 
--- a/packages/super-sync-server/src/sync/sync.service.ts
+++ b/packages/super-sync-server/src/sync/sync.service.ts
@@ -906,6 +906,42 @@ export class SyncService {
   }
 
   /**
+   * Delete pending passkey registrations whose verification token expired
+   * before `beforeTime`. Expiry is otherwise only enforced at read time
+   * (verifyEmail), so rows for abandoned registration attempts — including
+   * their credential public keys — would linger forever.
+   */
+  async deleteExpiredPendingPasskeyRegistrations(beforeTime: number): Promise<number> {
+    const result = await prisma.pendingPasskeyRegistration.deleteMany({
+      where: { verificationTokenExpiresAt: { lt: BigInt(beforeTime) } },
+    });
+    return result.count;
+  }
+
+  /**
+   * Delete unverified users created before `createdBefore` with no registration
+   * still in flight: no pending passkey registration row and no unexpired
+   * verification token (re-registering refreshes the token on the same row).
+   * Verified users can never match, and unverified users are denied auth, so
+   * nothing of value is lost when their related rows (operations, devices,
+   * sync state, passkeys, pending registrations) cascade on delete.
+   */
+  async deleteAbandonedUnverifiedUsers(createdBefore: number): Promise<number> {
+    const result = await prisma.user.deleteMany({
+      where: {
+        isVerified: 0,
+        createdAt: { lt: new Date(createdBefore) },
+        pendingPasskeyRegistrations: { none: {} },
+        OR: [
+          { verificationTokenExpiresAt: null },
+          { verificationTokenExpiresAt: { lt: BigInt(Date.now()) } },
+        ],
+      },
+    });
+    return result.count;
+  }
+
+  /**
    * Delete ALL sync data for a user. Used for encryption password changes.
    * Deletes operations, devices, and resets sync state.
    */
--- a/packages/super-sync-server/tests/cleanup.spec.ts
+++ b/packages/super-sync-server/tests/cleanup.spec.ts
@@ -6,6 +6,8 @@
  * - Stale device cleanup (using retentionMs cutoff)
  * - Rate limit counter cleanup
  * - Request deduplication cleanup
+ * - Expired pending passkey registration cleanup
+ * - Abandoned unverified user cleanup (using retentionMs as grace window)
  */
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { startCleanupJobs, stopCleanupJobs } from '../src/sync/cleanup';
@@ -21,6 +23,8 @@ const mockSyncService = {
   deleteStaleDevices: vi.fn().mockResolvedValue(0),
   cleanupExpiredRateLimitCounters: vi.fn().mockReturnValue(0),
   cleanupExpiredRequestDedupEntries: vi.fn().mockReturnValue(0),
+  deleteExpiredPendingPasskeyRegistrations: vi.fn().mockResolvedValue(0),
+  deleteAbandonedUnverifiedUsers: vi.fn().mockResolvedValue(0),
   updateStorageUsage: vi.fn().mockResolvedValue(undefined),
 };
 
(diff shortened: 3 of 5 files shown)


johannesjo:

Thanks for the PR! This is a careful implementation of #9031 – the schema types, the cascade FKs (verified down to the migration SQL), the read-time comparison parity and the guard predicates all check out, and the predicate test coverage is thorough. Two items should be addressed before merge, both around the interaction with `registerWithMagicLink` re-registration.

### Should fix before merge

- **`packages/super-sync-server/src/sync/sync.service.ts:935`** – the live-token guard does not cover the whole re-registration window. `registerWithMagicLink` deliberately sends the verification email BEFORE persisting the new token on an existing unverified row (`src/auth.ts:427`, "Send email BEFORE updating DB…"), so for the duration of the SMTP await the stored `verificationTokenExpiresAt` is still the old, expired value. If the daily sweep lands inside that await for a row past the 45-day window, every predicate matches: the user row is deleted, the subsequent `updateMany` matches 0 rows and returns the neutral message, and a link that was genuinely delivered dies as "Invalid verification token". The passkey path is safe by contrast – its pending row is created inside the transaction before the email goes out, so `none: {}` is armed first. Suggested fix: in the existing-user branch, when `updated.count !== 1`, re-check the row – if it is gone (rather than verified by a concurrent request), fall through to the create path with the same token so the delivered link stays valid. If you instead choose to accept the (seconds-wide, once-a-day) window, the JSDoc and PR description should stop claiming the guard closes this case and document the residual race instead. The deferred-email harness in `tests/integration/registration-races.integration.spec.ts:197` can reproduce it: hold the send open, run `deleteAbandonedUnverifiedUsers`, resolve the send, then `verifyEmail(token)` throws.
- **`ARCHITECTURE-DECISIONS.md:239`** – ADR #6 ("Passkeys Stay Pending Until Email Verification") gives as rationale that "Deleting the shared unverified user can race a concurrent registration and invalidate a link that was successfully delivered", and its "When to Update This Pattern" list names "Changing verification-token persistence or cleanup" – exactly what steps 5 and 6 do. Please extend the ADR with a note covering the new sweep, the grace window, and how the in-flight race above is bounded; otherwise the decision record and `cleanup.ts` disagree about a failure mode the ADR names.

### Nits

- **`packages/super-sync-server/src/sync/sync.service.ts:929`** – the JSDoc's "nothing of value is lost" is an inference about current `auth.ts`, not something the query enforces. Installs that predate the `isVerified` check in `verifyToken` can in principle hold unverified users WITH operations, and the cascade would erase those permanently. `operations: { none: {} }` in the same where object would make the safety claim structural; I've flagged the guard-vs-lean tradeoff for the maintainer.
- **`packages/super-sync-server/src/sync/cleanup.ts:95`** – step 1 in the same function argues (from the 2026-08 retention outage) that retention-critical sweeps must log unconditionally, including zero. That applies at least as strongly to a step that permanently deletes accounts; consider logging unconditionally and including the deleted user ids (ids only, never emails).
- **`packages/super-sync-server/tests/sync.service.spec.ts:789`** – the mock emulates the exact where-shape the code sends, so it verifies the emulator rather than Prisma's composition of `none: {}` with a top-level `OR`, and it cannot exercise the cascade. `tests/integration/` already has the pattern (`old-ops-sweep`, `registration-races`); a small case seeding abandoned/verified/in-flight users and asserting the survivors would close that gap – and is also the natural home for the race repro above.
- **`packages/super-sync-server/docs/architecture.md:116`** – the retention bullet enumerates what cleanup removes; one sentence should add pending passkey registrations and abandoned unverified accounts.
- **`packages/super-sync-server/src/sync/sync.service.ts:929`** – `deleteAbandonedUnverifiedUsers` takes `createdBefore` as a parameter but reads `BigInt(Date.now())` internally for the token predicate. Taking `now` as a second parameter (like the sibling `deleteExpiredPendingPasskeyRegistrations(beforeTime)`) keeps the method clock-injectable and the two halves consistent.

### Questions

- Deleting an abandoned row also resets that email's `verificationResendCount`, so the resend cap effectively becomes "20 per 45 days" per address rather than "20 ever". That looks acceptable given the per-IP rate limit, but was it considered? A one-line comment on the sweep would record it as deliberate.

---

johannesjo:
Thanks for this – decisions on the three open points from the review:

**1. Add `operations: { none: {} }` to `deleteAbandonedUnverifiedUsers`.** In principle it is
unnecessary: `auth.ts:218` denies auth to unverified users and `auth-cache.ts:98` records that
there is no verified → unverified transition, so an unverified row should never have operations.
But this is an irreversible `deleteMany` against real accounts, and the clause is one line that
can only ever prevent a wrong deletion. If the invariant holds it never changes behaviour and
documents the invariant in the place that matters; if it does not, it saves someone's tasks.
Please add it.

**2. The magic-link race: accept it, but ADR #6 must be updated in this PR.**

The race is real – `auth.ts:427` sends the email before the `updateMany`, so a sweep landing in
that SMTP window kills a link that was already delivered. I am accepting it, because it is
self-healing: the user re-registers, `existingUser` is null, and a fresh row and working link are
created. A recreate-on-missing-row path in the registration flow is not worth it for a
seconds-wide window hit at most once a day.

What is not optional is the ADR. [ADR #6](https://github.com/super-productivity/super-productivity/blob/master/ARCHITECTURE-DECISIONS.md#6-passkeys-stay-pending-until-email-verification)
gives as explicit rationale:

> Failed email delivery leaves the bounded, expiring pending attempt in place. Deleting the
> shared unverified user can race a concurrent registration and invalidate a link that was
> successfully delivered.

and its "When to Update This Pattern" section lists *changing verification-token persistence or
cleanup*, which is exactly what this PR does. This PR does not violate the ADR's intent – that
rationale is about deleting eagerly on delivery failure, while you delete 45-day-old abandoned
rows – but it does introduce a narrower instance of the same race, and right now the ADR still
reads as a reason not to do any of this. Please record that bounded deletion after a grace window
is allowed, and why the residual race is acceptable, so the next person does not re-derive the
old conclusion. A JSDoc note on `deleteAbandonedUnverifiedUsers` naming the race would help too.

**3. Reusing `retentionMs` for the grace window: keep as is.** Never-verified rows living 45 days
is fine, and avoiding a second config knob is the right call. Worth revisiting only if the table
actually grows.

**4. The resend-cap question from the review** (deleting a row resets `verificationResendCount`,
so the cap becomes 20 per 45 days per address rather than 20 ever) – I would still like your read
on whether that was deliberate.

MarkQuach12:
Deliberate, and I think it's an improvement rather than a trade-off.

`verificationResendCount` is only reset on successful verification (`auth.ts:61`, `auth.ts:112`). Nothing else clears it. So an unverified row that hits 20 is stuck for good: `auth.ts:421` and `passkey.ts:207` both go quiet and never send again, and that address can never register on the instance again, by any method.

So the change isn't "20 ever" becoming "20 per 45 days". It's "20, then locked out forever" becoming "20 per 45 days". For someone who owns the address and gave up halfway, that's the difference between recoverable and stuck.

Abuse-wise not much changes. The counter only resets on a row that's already 45 days old with no live token, so it isn't something you can trigger on demand. Waiting 45 days for another 20 emails is worse for an attacker than the per-IP limit they already hit.

Noted in a comment on step 6 in `cleanup.ts` and in the ADR.

johannesjo:
Thank you very much! <3

MarkQuach12:
Both required items are in as of `1b870f0ab`:

- Added `operations: { none: {} }` to `deleteAbandonedUnverifiedUsers`, with a spec case that fails without it. The JSDoc now names the magic-link race as well.
- ADR #6 updated. The old rationale bullet now says what it rules out is deleting eagerly on delivery failure, not the retention sweep. Added bullets covering the sweep, the grace window, and why the residual race is acceptable.

The resend-cap question is answered in the comment above, and recorded as a one-line note on step 6 in `cleanup.ts`.

I left the nits alone for now (unconditional logging, integration test, architecture.md bullet, injectable clock). Happy to pick up any of them if you want them in this PR.

Ready for re-review.

PR #9282, opened 2026-07-24
feat(plugins): add Plane.so issue provider

rajabilal555:
## Summary
- Add bundled `plane-issue-provider` plugin for Plane.so work items (search, import as tasks, open in browser, backlog poll, pull-only done sync from `completed`/`cancelled` state groups)
- Support cloud and self-hosted via API key, workspace slug, project UUID, and optional host
- Fix uploaded plugin zips so `i18n/*.json` is cached and loaded before plugin code runs (config labels no longer show raw `CFG.*` keys)
- Document Plane in the integrations wiki / comparison matrix

## Test plan
- [ ] `cd packages/plugin-dev/plane-issue-provider && npm test && npm run build`
- [ ] `npm run plugins:build` (or build plane plugin into `src/assets/bundled-plugins/plane-issue-provider/`)
- [ ] Enable Plane in the issue provider panel; configure API key, workspace slug, project UUID
- [ ] Test connection, search, add work item as task, open browse link
- [ ] Confirm completed/cancelled Plane items mark local tasks done on poll (pull-only)
- [ ] Optional: upload a plugin zip and confirm config labels are human-readable (not `CFG.API_KEY`)

Made with [Cursor](https://cursor.com)

Issue #3382, opened 2024-09-01
Plane.So Issue Integration


h-sigma:
https://plane.so/ is a self-hosted/cloud open source project management tool intended for use by developers. 
I have been using it for a while and I quite like it.

1. If I submitted a PR with Plane integration, would it be considered for a merge? I have been using super-productivity for a year+ now but don't know the maintenance situation of  its open source project.
2. Anyone who knows the code can point me in the right direction and know if it's relatively straightforward to add an issue integration?

johannesjo:
> 1. If I submitted a PR with Plane integration, would it be considered for a merge?

Yes! It would! :) 



> 2\. Anyone who knows the code can point me in the right direction and know if it's relatively straightforward to add an issue integration?

Best place to start is:
https://github.com/johannesjo/super-productivity/tree/master/src/app/features/issue

There you should look at the existing integrations for jira.

The code there is far from perfect, so if you have improvement suggestions, I am all ears.

PR #39, opened 2017-11-29
Implemented flatpak build. Closes #25

patriziobruno:

PR #9745, opened 2026-08-27
feat(plugins): add Plane.so issue provider #3382

daxrpm:
Bundled issue-provider plugin for Plane work items: search, import as tasks, open in browser, backlog import, due day from the work item's target date, and pull-only done sync from the completed and cancelled state groups. Works against Plane Cloud and self-hosted instances via an API key, workspace slug and project UUID.

Continues #9282 by @rajabilal555, rebased onto current master and corrected. He is credited as co-author and has confirmed on that PR that he'd rather not finish it himself.

Written with Claude Opus 5 (commit trailer). Every claim below was checked against the Plane v1.4.2 source or a live workspace, and every fix carries a test verified to fail without it.

## Problem

#3382 asked for a Plane integration and you said yes; the stale bot later closed it as `NOT_PLANNED`. #9282 was the one attempt and has sat with a merge conflict and self-reported bugs since 24 July.

## Solution

Rebasing #9282 was not the route. Two of the three conflicts were its secondary "load i18n from plugin zips" fix, which master solved on its own in `6c40d831` (#9474); the third is a one-line clash in `plugin-tests.yml`'s plugin matrix against `33b91716` (#9274). Resolving the markers naively doesn't compile either — git folds the PR's i18n block into master's and declares `translations` twice in one scope. So this takes only the plugin directory onto master and rewrites the registration by hand.

**What #9282 got wrong**

- **`icon: "plane"` had no registered SVG.** `MatIconRegistry` renders an unknown name as an empty box without erroring, so the chip was blank.
- **Backlog import truncated at 100.** It read two fixed pages and filtered afterwards; a project whose first 100 rows were closed imported nothing, with no error.
- **`testConnection` probed `/users/me/`**, which only validates the API key. A mistyped slug or project UUID reported success and then imported nothing. It now probes the project.
- **`issueProviderKey: "PLANE"`** — `docs/add-new-integration.md` reserves that for plugins migrating an existing built-in key; it would have burned a global key into synced user data.
- `description_stripped` is plain text but was declared `type: 'markdown'`.

**Two decisions worth a second opinion**

`order_by=sequence_id`, not `state__group`. Ordering by state group would put open items first and allow stopping at the first closed row — tempting and wrong. Plane pages by OFFSET and `IssueListCreateAPIEndpoint` never passes `order_by` to the paginator, so the `-created_at` tiebreaker at `paginator.py:138` is skipped and the SQL is `ORDER BY state_order` over five distinct values — nearly every row a tie, so rows can land on two pages or none. `sequence_id` is unique per project and allowlisted. The early stop is unsound anyway: `default=Value(len(state_order))` sorts a NULL-state item *after* `cancelled`.

Due day parsed as local midnight. `target_date` is a `DateField` — a calendar day, no time, no timezone. `new Date('2026-08-03')` reads it as midnight UTC, still 2 August anywhere west of Greenwich, so the task would land a day early for every user in the Americas. Local midnight cancels against `getDbDateStr`. The helper also refuses to return `NaN`: the host accepts `start` on a bare `typeof === 'number'` check, and a leaked `NaN` persists as a due day of `"NaN-NaN-NaN"`.

**No write-back, by design.** Marking an item done in Plane needs a specific state UUID and a project may hold several states in the `completed` group, so there is no obviously right one to pick. `updateIssue` is left unimplemented, which makes the host hide the push directions rather than offer a setting that cannot work.

**Plane's API sends no CORS headers** (makeplane/plane#4928), so this **cannot work in the web build** — the preflight fails. Desktop is unaffected because `main-window.ts:233-278` strips `Origin` and injects the headers; confirmed with a live probe returning a real `403` rather than a network error. The wiki states the limitation. Desktop and Android are verified on device; iOS is expected to behave the same but is untested.

Two things I'd defer to you: the icon is traced from Plane's own mark (their published asset is a base64 PNG) and redrawn in `currentColor`; and import scope is the whole project's open items rather than only those assigned to you, matching GitHub/Gitea/Redmine/Trello — Linear is the exception there.

## Verification

Full unit suite 14407 green · electron 281 · plugin package 110 · E2E 328 (Docker-backed sync suites excluded) · `lint` clean · `plugins:build` 14/14.

Reverting production behaviour is caught in every case: midnight-UTC date parsing (4 failures), no `order_by` (4), no `expand` (5), `state` dropped from `fields` while still expanded (4), no closed-item filter (3), raised page budget (1), `next_cursor` not echoed (4), the old two-page cap (2), `cancelled` misspelled US-style (4), the wrong auth header (3).

Manually, in the Electron dev build against a real Plane Cloud workspace (13 backlog / 11 todo / 1 in progress / 11 done): connection test passes and **fails when the slug is mistyped**; backlog import brings in exactly the 25 open items and none of the 11 done ones; due days land on the right calendar day. **Self-hosted is not verified** — everything specific to it came from the v1.4.2 source rather than a running instance.

`electron/bundled-plugin-ids.test.cjs` gains a second test that is general rather than Plane-specific: every bundled issue-provider's `issueProvider.icon` must be registered *and* its SVG must exist, read off disk rather than restated. It also shows two cosmetic hunks — that file was already not prettier-clean on master and the pre-commit hook reformats what it touches.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


johannesjo:

Thanks for the PR, and for picking this up cleanly from #9282 with @rajabilal555 credited. This adds a bundled Plane issue-provider plugin – search, backlog import, browse links, due day from `target_date`, pull-only done sync – and it is in good shape: registration is complete (`BUNDLED_PLUGIN_PATHS`, `BUNDLED_PLUGIN_IDS`, `global-theme.service.ts` icon map, `build-all.js` with `assertFilesExist`), the plugin-scoped devDeps keep the root `package.json` untouched, and `plugin.spec.ts` models Plane's offset paging and `fields`/`expand` projection rather than handing back canned pages, which is the reason the `order_by` and closed-item guards are worth anything. I checked the endpoint shapes and the `<uuid:project_id>` route against Plane's own source and they match, and I confirmed the `start` → local-midnight → `getDbDateStr` reasoning holds against `src/app/util/get-db-date-str.ts`. Two small things below, plus two nits.

One heads-up before you start: I've flagged a couple of scoping questions for the maintainer; you may want to wait for those before reworking any details.

### Should fix before merge

- **`packages/plugin-dev/plane-issue-provider/src/plane-helpers.ts:302`** – `mapListRow()` never sets `lastUpdated`, even though `updated_at` is in `LIST_FIELDS` and so present on every row. Backlog-imported tasks are therefore stored with `issueLastUpdated: 0` (`plugin-issue-provider-adapter.service.ts:454`, `data.lastUpdated ?? 0`). At the first poll, `plugin-issue-provider-adapter.service.ts:224` evaluates `issue.lastUpdated > (task.issueLastUpdated || 0)` as true for *every* imported task, so within 5 minutes of an import the user gets an "N Plane work items updated" snack (`issue.service.ts:451`), every task picks up the `issueWasUpdated` marker, and – because `getFreshDataForIssueTask` spreads `addTaskData` into `taskChanges` – every title is rewritten from `mapListRow`'s `CYBER-12 Fix login` to `mapWorkItem`'s bare `Fix login`.

  Fix: `lastUpdated: item.updated_at ? new Date(item.updated_at).getTime() : 0` in `mapListRow`, matching `mapWorkItem` at `plane-helpers.ts:279`. (`PluginSearchResult` has an index signature, so it type-checks.) `trello-issue-provider` and `azure-devops-issue-provider` already set it on their search-result mappers.

  Worth saying so you don't chase it: this only covers the backlog path. Plane's `/work-items/search/` returns just `id`/`name`/`sequence_id`/`project__identifier`, so `mapSearchHit` genuinely cannot set `lastUpdated`, and a task added from the search box will still take one spurious update. That is the same shape `gitea-issue-provider` has today, so I wouldn't reach for an extra `getById` round-trip over it.

- **`docs/wiki/3.05-Web-App-vs-Desktop.md`** – the wiki work here is thorough, but `docs/documentation-guide.md` maps "Web vs desktop capability differences" to this note specifically, and it's the reference a user checks for "what doesn't work in the browser?". Plane is entirely non-functional there and doesn't appear. A couple of lines in the CORS section (line 37), next to the existing CalDAV/WebDAV cases and pointing at the detail you already wrote in `3.07`, would close it.

### Nits

- **`packages/plugin-dev/plane-issue-provider/src/manifest.json:11`** – `"permissions": ["http"]` grants nothing here: the plugin never calls `PluginAPI.request`, the issue-provider `http` handle is built independently (`plugin-bridge.service.ts:490`) and `issue-provider-types` documents it as unaffected by manifest permissions, and with no `allowedHosts` the capability is fail-closed anyway. The visible effect is that `plugin-security.ts:265` renders "Can make network requests to: (none declared – network disabled)". All six other bundled issue providers ship `"permissions": []`.
- **`packages/plugin-dev/plane-issue-provider/src/plugin.ts:104`** – past `LIST_MAX_PAGES × LIST_PAGE_SIZE` (2000) the import is truncated with only a `console.warn`, so a larger project imports silently-partial results. The 2000 ceiling is already far more generous than gitea/github (one 100-item page, no pagination), and the wiki states it, so this is genuinely optional – but if you want a signal without new host API surface, throwing once the budget is exhausted would surface the existing "Backlog import failed" snack (`plugin-issue-provider-adapter.service.ts:303`).

### Questions

- The wiki says **"Desktop app only"** / "Plane needs the desktop app", and the PR notes Android/iOS are untested. But `3.05-Web-App-vs-Desktop.md:39` states the mobile apps issue requests through a native HTTP layer and are not subject to browser CORS, and plugin issue-provider GETs go through Capacitor's patched fetch on device rather than the browser stack. So the limitation may be web-app-only rather than desktop-only. Did you get a chance to try it on a device? If not, wording it as "not available in the web app" would be accurate without claiming mobile either way.

---

johannesjo:
Thanks for picking this up and for crediting the original author – decisions on two of the three
open points:

**1. Desktop and mobile only is fine.** I am happy to ship this as a bundled provider that does
not work in the web app; no web-build gate needed. One wording fix though: the review's question
here is a good one. Mobile is *not* subject to browser CORS – plugin issue-provider GETs go
through Capacitor's patched fetch on device – so "Desktop app only" is probably wrong. Please
change the wiki wording to "not available in the web app", and if you have a device handy it
would be good to confirm Android/iOS actually work.

**2. The icon is fine.** We already ship `jira.svg`, `github.svg`, `gitlab.svg`, `linear.svg`,
`trello.svg`, `clickup.svg`, `redmine.svg`, `gitea.svg`, `azure_devops.svg` and others – using a
mark to identify a real integration is the same situation for Plane as for any of those. One
small ask: prefer Plane's own published asset over the hand-trace if you can convert it. An
unmodified official mark is both more accurate and the thing most brand guidelines actually
permit, whereas a trace is a modification. Not a blocker.

**3. Import scope** (whole project vs. only items assigned to the user) – still thinking about
this one, I will come back to you.

The blocker from the review still stands and is the main thing to fix: `mapListRow()` omits
`lastUpdated` even though `updated_at` is already fetched, so every imported task is stored with
`issueLastUpdated: 0` and the first poll fires a spurious "N Plane work items updated" snack and
rewrites every task title.

One separate note, not your problem and not blocking: while reviewing this I found that
`linear-`, `gitea-`, `trello-`, `azure-devops-issue-provider` and `brain-dump` are in
`BUNDLED_PLUGIN_PATHS` but missing from `build-all.js` – filed as #9759.

johannesjo:
Coming back on point 3 (import scope): **your call.** Whole-project and assigned-only are both
defensible – whole-project matches GitHub, Gitea, Redmine and Trello, assigned-only matches
Linear – and you know the Plane workflow better than I do. Keep what you have, or add an
assignee filter if that is how you actually use it. Not a blocker either way.

So the only thing left before this can go in is the `mapListRow()` / `lastUpdated` blocker, plus
the wiki wording fix ("not available in the web app" rather than "Desktop app only").

johannesjo:

Thanks for the quick turnaround. Everything from the last round is addressed at `9133c8e`: `mapListRow()` now carries `updated_at` (with tests on both the present and absent branches), the wiki says "not available in the web app" in all three places, `permissions` is back to `[]`, and the CodeQL sanitization alert is answered with a strip-until-stable loop. I re-read the whole plugin at this head rather than just the delta, and the test double in `plugin.spec.ts` modelling Plane's offset paging and the `fields`/`expand` projection is doing real work – the `order_by` and closed-item guards mean something because of it.

Two things I could not settle from the source, so they are questions rather than findings.

### Questions

- **`packages/plugin-dev/plane-issue-provider/src/plane-helpers.ts:181-190`** – the entity decoder passes an unchecked value to `String.fromCodePoint()`. Running that expression standalone, `&#x110000;`, `&#1114112;` and `&#99999999999;` each throw `RangeError: Invalid code point`; `&#x2713;`, `&#x10FFFF;` and `&#xD800;` are all fine. If such a string ever reached `htmlToText()` it would throw out through `mapWorkItem()` (`plane-helpers.ts:295`) and `getById()` (`plugin.ts:198`), and the host swallows it – `plugin-issue-provider-adapter.service.ts:255-262` logs and returns `null`, and the `getIssueLink` URL fallback at `:89-99` returns `''` – so that one task would silently stop refreshing with nothing shown to the user. This path runs on every `getById`, since the public API serializer excludes `description_stripped`.

  What I could not tell is whether such input can reach you at all. Reading Plane's serializers, both write paths appear to normalize numeric character references out of `description_html` before storage (lxml round-trip plus `nh3.clean` on the API serializer, `nh3.clean` on the app serializer), and Tiptap escapes a typed `&` into `&amp;`, which the regex consumes first and leaves inert. So this may be unreachable in practice, and I could not construct an input that gets there. Do you know of a path that writes `description_html` straight through – an importer, or rows written by an older self-hosted version before the content validator existed? If not, feel free to leave it; if you would rather not think about it again, the guard is one line and matches the shape of the existing unknown-entity branch:

  ```ts
  const cp = key.startsWith('#x') ? parseInt(key.slice(2), 16) : Number(key.slice(1));
  return Number.isInteger(cp) && cp >= 0 && cp <= 0x10ffff ? String.fromCodePoint(cp) : match;
  ```

- The wiki now states "The desktop and mobile apps work normally" and "the desktop and mobile apps are unaffected" as fact, while the PR description still says Android/iOS are untested. The reasoning is sound – plugin issue-provider GETs go through Capacitor's patched fetch on device and are not subject to browser CORS – but did you get a device to confirm it end to end? If not, wording it as expected-but-unverified for mobile would be accurate without claiming it.

---

daxrpm:
Both questions answered at 94137cd — and sorry for the inconsistency: the wiki claimed
mobile as fact while the PR said untested.

**1.** Probed live Plane Cloud: invalid numeric references (`&#x110000;`, `&#xD800;`, …)
are normalized to U+FFFD at write time, so current Cloud can't produce this input. Legacy
self-hosted rows can't be ruled out, so your guard is in — plus lone surrogates rejected
(no throw, but ill-formed UTF-16; Plane itself stores U+FFFD for them). Tests fail with
RangeError without the guard.

**2.** Verified on Android twice: physical Redmi Note 13 + Pixel 7 emulator (AOSP 14),
against real Plane Cloud — connection test, search, import, backlog with closed items
excluded, due days in UTC-5, done-pull. Wiki now says: desktop and Android verified, iOS
expected but untested.

Device testing also caught a real bug: `mapWorkItem()` titled with the bare name, and the
host rewrites titles from `issue.title` on refresh, so the first update stripped the
`KEY-12 ` prefix — seen on both devices. Fixed to match the import format, with a test.

Also: dead `'#39'` entity entry removed, wiki example switched to placeholder data, and a
note that auto-import needs a default project. Plugin 110/110, electron 281/281, full
suite 14407 green.

johannesjo:
Thanks for picking this up the way you did, @daxrpm — asking @rajabilal555 first and offering to hand the patch back rather than opening a competing PR straight away is the right instinct, and it's why this was an easy call. I've closed #9282 in favour of this one.

One correction to ask of you, though. This line in the description doesn't hold at #9282's head (`abd9b297`):

> The plugin was never registered in `BUNDLED_PLUGIN_PATHS` / `build-all.js`, so it never actually loaded — which also silently disabled `allowPrivateNetwork`, i.e. self-hosted Plane.

I checked all three files at that sha: the plugin is registered in `plugin.service.ts:59` and `:88`, and in `packages/plugin-dev/scripts/build-all.js:321-329`. Could you edit that out of the body (and the same claim in your #9282 comment)? It reads as a defect in Bilal's work in the PR that supersedes it, so I'd like the record straight.

Your other points check out — `issueProviderKey: "PLANE"`, the unregistered `icon: "plane"`, the backlog-import paging bug, and the i18n-from-zip conflict against 6c40d831 (#9474). Those alone justify the rebase; the change is to the record, not to the outcome.

The two review passes already on this PR still stand — happy to look again once those are addressed and the description is updated.

daxrpm:
Hi @johannesjo — thanks for the quick turnaround, and for closing #9282 in favour of this one.

The registration claim is out of both the description and my #9282 comment, with an edit note
there.I also checked the rebased tree instead of his head.

Re-verifying the rest, I caught one more overstatement of mine: two of the three conflicts were
the i18n fix, not all three — the third is a one-line clash in `plugin-tests.yml` against
`33b91716` (#9274). Fixed, along with two stale numbers (110 plugin tests, and Android is
verified on device).

Everything from both reviews is addressed at `94137cd`. Happy and open to pick up anything else.

daxrpm:
Rebased onto master at 46a1360 — the conflict in global-theme.service.spec.ts is
resolved (both describes kept, +34/-0 against master, no lines of yours dropped)
and the branch is mergeable again.

johannesjo:

Thanks for the rebase and for the detailed answers. This is a delta pass over `46a1360`: everything from the two earlier rounds is addressed, so I re-read what moved plus the rebase, and then went at the one part of the plugin the earlier passes had not reached.

The delta itself checks out:

- **Entity guard** (`plane-helpers.ts:180-198`) – range, integer and surrogate checks are all right. The 400-digit decimal becomes `Infinity` and `Number.isInteger` rejects it; `parseInt` on a long hex string does the same; `&#X2713;` still decodes because `key` is lowercased first; and `String.prototype.replace` never rescans its own output, so `&amp;#39;` staying literal is guaranteed rather than incidental. The `'#39'` map entry really was dead – the regex captures `#39` and the numeric branch takes it before the named lookup.
- **Title prefix** (`plane-helpers.ts:305`) – confirmed against the host: `_buildBaseIssueTask()` sets `title: data.title` (`plugin-issue-provider-adapter.service.ts:451`), `getFreshDataForIssueTask()` spreads it into `taskChanges`, and `issue.service.ts:438` applies it with `_taskService.update()`. So the bare name did strip the key on the first refresh, and `${key} ${item.name}` is the right fix. `getById` sends `expand=…,project`, and Plane does map `project` to a serializer carrying `identifier`, so the key resolves there.
- **Wiki** – the `#set-a-default-project-for-an-integration` anchor exists in `2.07`, and the new default-project sentence is accurate: `isAutoAddToBacklog` is disabled unless `defaultProjectId` is set (`common-issue-form-stuff.const.ts:81`) and `poll-to-backlog.effects.ts:99` requires it again at poll time. Both fields sit inside the dialog's `ADVANCED_CONFIG` collapsible, so "Advanced Config" is the right name for it.
- **Rebase** – `global-theme.service.spec.ts` is a clean append after master's last `describe` (+34/-0, nothing dropped), and `InitIconsHarness` does not collide with the file's existing `GlobalThemeInitHarness`.

One thing from the unchanged part of the plugin that neither earlier pass reached.

### Should fix before merge

- **`packages/plugin-dev/plane-issue-provider/src/plugin.ts:196`** – `expand: 'state,assignees,project'` asks for something Plane's public API cannot expand, and asking for it makes the response worse than not asking.

  In Plane's `api/serializers/base.py` the expand loop is `if expand in self.fields:` and then `if expand in expansion: … else: response[expand] = getattr(instance, f"{expand}_id", None)`. `assignees` **is** in `self.fields` – `IssueSerializer` declares it as a write-only `ListField` – but it is **not** in the `expansion` mapper, which holds `user`, `workspace`, `project`, `default_assignee`, `project_lead`, `state`, `created_by`, `updated_by`, `issue`, `actor`, `owned_by`, `members`, `parent`, `estimate_point` and nothing else, both on `main` and at `v1.4.2`. `Issue.assignees` is a `ManyToManyField`, so there is no `assignees_id` attribute and the retrieve view annotates none. The endpoint therefore answers `"assignees": null`.

  So `assigneeNames()` (`plane-helpers.ts:263`) reads `item.assignees || []`, gets `[]`, and `mapWorkItem` sets `assignee: undefined` on every call. The Assignee row in `issueDisplay` (`plugin.ts:242-247`) is `hideEmpty: true`, so it never renders for any work item, however many assignees it has in Plane. Removing the `expand` would not rescue it either: unexpanded, `assignees` is a list of UUIDs, and `assigneeNames` deliberately maps a string to `''`.

  `plugin.spec.ts:676` passes only because the double returns `assignees: [{ display_name: 'Ada' }]`, a shape the API never sends. That is the one place the double diverges from Plane – it models the offset paging and the `fields`/`expand` projection faithfully, which is why this stood out.

  Smallest fix: drop `assignees` from the expand string, drop the Assignee entry from `issueDisplay`, and drop `assignee` / `assignees` / `assigneeNames` from the mapper. The API cannot back that field, and shipping permanently blank surface in a bundled provider is worse than not having it. Keeping it would need a `…/projects/<uuid>/members/` lookup to turn the UUIDs into names – a real extra round-trip, and probably not worth it here.

### Questions

- I read the above out of Plane's source rather than off a live response, and you have an instance I don't. If your Plane Cloud project does return expanded assignee objects on `GET …/work-items/<id>/?expand=state,assignees,project`, say so and I'll drop the item – one paste of that response's `assignees` value settles it either way.

---

PR #9746, opened 2026-08-27
fix(schedule): anchor month-view day 0 to the first displayed day

### LINKED ISSUE #9071: fix(schedule): month view anchors day 0 with a timestamp from a different day
## Problem

In month view, `daysToShow()[0]` and `_contextNow` describe **different days**.

`getMonthDaysToShow` (`src/app/features/schedule/schedule.service.ts:165-191`) starts the grid on the first day of the week containing the 1st, so day 0 is usually a **padding day from the previous month**. Viewing Feb 2026 with `firstDayOfWeek=1` gives `daysToShow[0] = '2026-01-26'`.

`_contextNow` (`schedule.component.ts`) derives its midnight from `_selectedDate` — i.e. Feb 1. But `create-schedule-days.ts` anchors day 0 with that value:

```ts
let startTime = i == 0 ? now : dayStartTime;   // :89
```

So the Jan 26 cell is laid out with a `startTime` of Feb 1 00:00 — six days past its own `nextDayStart`.

Knock-on: day 0's repeat projections are selected for the wrong date —

```ts
selectTaskRepeatCfgsForExactDay.projector(unScheduledTaskRepeatCfgs, { dayDate: startTime })
```

is fed `_selectedDate`'s day rather than the padding day's.

## Impact

Low in practice: `create-schedule-days` filters unscheduled flow tasks out whenever day 0 isn't in the current week (`isFirstDayInCurrentWeek` is false for a padding day), so most of the layout is unaffected. But the anchor is wrong by construction, and the repeat-projection selection is a real (if quiet) defect.

## Direction

Single-source `_contextNow` from `daysToShow()[0]` rather than `_selectedDate`:

```ts
const first = this.daysToShow()[0];
const dayStart = parseDbDateStr(first).setHours(0, 0, 0, 0);
const now = Date.now();
return now >= dayStart && now < nextDayStart ? now : dayStart;
```

Equivalent in week/day view (`getDaysToShow` always anchors day 0 at the reference date), and correct in month view. Cleaner long-term shape is an explicit `_anchorDay` that **both** `daysToShow` and `_contextNow` derive from.

Deliberately left out of #9064, which kept month view byte-identical: this is a behaviour change that deserves its own diff and review.

_Found during multi-agent review of #9064._

MarkQuach12:
## Problem

Closes #9071. In month view, `daysToShow()[0]` and `_contextNow` describe different days. The grid starts on the first day of the week containing the 1st, so day 0 is usually a padding day from the previous month, while `_contextNow` derived its midnight from `_selectedDate` (the 1st). `create-schedule-days` anchors day 0 with that value (`startTime = i == 0 ? now : dayStartTime`), so the padding cell was laid out with a timestamp from a different day, days past its own `nextDayStart`, and day 0's repeat projections were selected for the wrong date.

## Solution

Single-source `_contextNow` from `daysToShow()[0]`, per the direction in the issue: derive the day bounds from the first displayed day with the same helper create-schedule-days uses (`dateStrToUtcDate` + `setHours(0)` / `setHours(24)`), return live `Date.now()` when the wall clock sits inside that day, else day 0's midnight. Week and day views are provably unchanged: `getDaysToShow` places the reference date itself at index 0, so the bounds are identical to the old `_selectedDate` derivation there.

One consumer needed a minimal adaptation, which the fix exposed: `createBlockedBlocksByDayMap` generates work-hours blocks, lunch breaks and timed repeat projections for a fixed 10-day window from the anchor. Tuned for week view, that window no longer covers a 35 to 42 day month grid once the anchor is honest, and current-month view lost timed projections beyond day 10 (reproduced by a new test that got 0 of 10 expected projection days). The window now spans `dayDates.length`; days outside the rendered range were never consulted, so day and week views are unaffected, guarded by a range-parity test.

Also fixed as the same bug class: with a custom start-of-next-day offset, the old null-selectedDate branch returned live now during the 00:00-to-offset window even when that fell outside day 0, pushing every day-0 entry out of its column; the new bounds check covers it, and the existing "never anchor day 0 with a now outside it" contract test applies.

## Verification

New tests: month grid with padding day 0 anchors to day 0 midnight (navigated and current-month cases), today-as-first-cell returns live now, day view unchanged, timed daily repeat projects onto month days beyond 10 days from the anchor, and a 2-day window parity guard. Without the source changes exactly those new tests fail (2 FAILED of 73 in the component spec, 1 FAILED of 4 in the new mapper spec); with them: component 73/73, all 12 map-schedule-data specs 104/104, schedule.service 47/47, schedule-month 36/36. Five pre-existing tests had their `getDaysToShow` mocks aligned with the day range each scenario describes; none deleted, and all adapted tests pass on both old and new code. `npm run checkFile` clean on all four files. Production build green.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/src/app/features/schedule/map-schedule-data/map-to-schedule-days.ts
+++ b/src/app/features/schedule/map-schedule-data/map-to-schedule-days.ts
@@ -74,6 +74,13 @@ export const mapToScheduleDays = (
     (task) => !(typeof task.dueWithTime === 'number'),
   ) as TaskWithoutReminder[];
 
+  // Span the blocker window across every rendered day. The 10-day default was
+  // tuned for week view, where the visible range always sits inside it; a
+  // month grid is up to 42 days anchored at its first cell, so work-hours
+  // blocks and timed repeat projections would otherwise only cover the first
+  // stretch of the grid. Days outside the rendered range never mattered:
+  // createScheduleDays only consults the map for rendered dayDates, so a
+  // shorter-than-10 span changes nothing for day and week view either.
   const blockerBlocksDayMap = createBlockedBlocksByDayMap(
     scheduledTasks,
     scheduledTaskRepeatCfgs,
@@ -81,7 +88,7 @@ export const mapToScheduleDays = (
     workStartEndCfg,
     lunchBreakCfg,
     now,
-    undefined,
+    dayDates.length,
     realNow,
   );
 
--- a/src/app/features/schedule/map-schedule-data/month-view-repeat-projection-day.spec.ts
+++ b/src/app/features/schedule/map-schedule-data/month-view-repeat-projection-day.spec.ts
@@ -1,6 +1,7 @@
 import { mapToScheduleDays } from './map-to-schedule-days';
 import { TaskRepeatCfg } from '../../task-repeat-cfg/task-repeat-cfg.model';
 import { getDbDateStr } from '../../../util/get-db-date-str';
+import { SVEType } from '../schedule.const';
 
 const H = 60 * 60 * 1000;
 const h = (hr: number): number => hr * H;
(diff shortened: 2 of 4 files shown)


johannesjo:
**✅ Ready to merge** – no blocking issues at `9e84194`. Merge confidence 85% – I verified the anchor arithmetic and the widened blocker window by reading the source plus the green CI run, but nothing renders the month grid end to end (the new mapper spec is the closest thing) and I did not run the app.

Thanks for the PR! Deriving `_contextNow` from `daysToShow()[0]` is exactly the direction #9071 asks for, and the write-up made this easy to check. I traced the `nrOfDays` change too, since that is the one behavioural surface the fix widens: `createScheduleDays` only ever reads `blockerBlocksDayMap[dayDate]` for rendered `dayDates`, `mapToScheduleDays` is the map's only caller, the work-hours / lunch / projection loops all run `i` over `[0, nrOfDays)` so the last rendered day still gets its evening block from `i = length - 1`, and `mergeBlocksRecursively` only ever extends a block forward, so dropping the trailing days cannot change an earlier one. The shrink to 7 and 1 in week and day view really is inert, as your parity test claims. Only nits below.

### Nits

- **`src/app/features/schedule/schedule/schedule.component.ts:47`** – the file now imports two parsers that are identical for a `YYYY-MM-DD` string: `parseDbDateStr` (used four times in `headerTitle`) and the new `dateStrToUtcDate`. Matching `create-schedule-days` is the right call for `_contextNow`, so if you want to collapse them, the direction is moving `headerTitle` onto `dateStrToUtcDate`, not the reverse.
- **`src/app/features/schedule/schedule/schedule.component.ts:278`** – `nextDayStart` re-parses `firstDay` from the string. `const nextDayStart = new Date(dayStart); nextDayStart.setHours(24, 0, 0, 0);` is exactly equivalent (`dayStart` is already local midnight, and `setHours(24)` normalises the same way across DST). `create-schedule-days` does the same double parse, so this is only worth it if you prefer it.
- **`src/app/features/schedule/schedule/schedule.component.spec.ts:702`** – the test name uses backticks where apostrophes are meant, so the spec reporter prints ``month grid`s first cell``.
- **`src/app/features/schedule/map-schedule-data/map-to-schedule-days.ts:91`** – worth being aware of rather than a bug: in current-month view `now` is no longer the live clock, so `isViewingCurrentDay` in `create-sorted-blocker-blocks.ts` is false and the projection loop starts at `i = 0` instead of `1`. Combined with the 42-day span, timed repeat projections can now land on *past* grid cells whenever a cfg's effective `lastTaskCreationDay` is stale – app closed across a day boundary, or a cfg just unpaused, before the creation effect advances it. Cosmetic, and untimed projections already do this on every rendered day, but it is a surface the old today-anchored 10-day window could not reach.
- Out of scope here, but the same anchor mismatch survives in the right-panel day schedule: `schedule-day-panel.component.ts` builds its days through `ScheduleService.createScheduleDaysComputed`, which calls `buildScheduleDays` with no `now`, so `now` falls back to the raw `Date.now()` while `daysToShow` is `todayDateStr$`, i.e. the *logical* today. With a custom start-of-next-day, between midnight and the offset `startTime` sits past that day's `nextDayStart`, and the `entry.start >= nextDayStart` branch in `create-schedule-days` pushes every flow entry onto a day that is never rendered, so the panel empties. It survives the first paint (the `toSignal` initial value goes through `todayStr(Date.now())`, which does not apply the offset) and only breaks once the stream emits. Probably its own issue rather than something to bolt onto this PR.

---

johannesjo:
Thank you very much! <3

